Issue #820, opened 2026-07-13
feat(markov): CTMM simulation + prediction — Gillespie path generator, occupancy π(t), sdtab (Phase 5 follow-up)


roninsightrx:
## Summary

Add **simulation and prediction** for the continuous-time Markov (CTMM) endpoint (`[markov_model]`, #759). Phase 5 (#819) shipped the **fit path** only; `simulate()` / `predict()` do not yet produce CTMM output. This is the deferred Phase-5 simulation deliverable.

Covers the time-homogeneous generator (built once per subject); a drug-driven `Q(t)` simulation rides on Phase 6 (#817).

## Scope

Per the plan (`plans/tte-survival-markov.md` §8.8.2 / §8.8.4 / §8.8.5):

- **Gillespie / Doob path generator** (`src/markov/simulate.rs`): for a per-subject generator `Q`, draw the latent continuous-time state path — sojourn `τ ~ Exp(−q_ss)` in the current state `s`, then jump to state `k ≠ s` with probability `q_sk / (−q_ss)` — up to the `[simulation] horizon`.
- **Observation on a grid**: record the simulated state at each `[simulation] obs_schedule` time (§8.8.4), emitting `SimOutcome::Category { state }` (the variant already exists) — map the generator index back to the declared DV code.
- **Prediction**: state-occupancy vector `π(t) = π₀ · expm(Q·t)` → `Prediction::CatProbs { probs }` (reuses the Phase-5 `matrix_exp`; no new numerics). Needs the `Prediction` enum, which does not exist yet (§8.8.1) — coordinate with the categorical (#760) prediction work, which needs the same enum.
- **Initial state / occupancy**: `init` occupancy vector (parsed but rejected in #819) drives the path's starting state and `π₀`; decide fixed vs. data-first-state vs. estimated (the #759 open question).
- **Diagnostics (sdtab, §8.8.5)**: CTMM row = occupancy `π(t)` at the EBE η + a state-prediction residual.

## Validation

Per the epic's simulation-estimation (SSE) convention (§8.8.8, §14.11): simulate a path from a known `Q`, refit, and recover the intensities (degenerate/round-trip oracle). A VPC-style check (simulated vs. observed state-occupancy proportions over time) is the natural external comparator; NONMEM has no feedback/path simulation here, so the anchor is the SSE round-trip plus the closed-form `π(t) = π₀ expm(Qt)` for the prediction side.

## Out of scope

[...]

PR #773, opened 2026-07-09
feat(io): add discrete-state/count observation plumbing (Phase 4.0) [refs #760 #759]

TeunP:
## Why

Phase 4.0 of the TTE/Markov plan (`plans/tte-survival-markov.md` §19) is the shared foundation that lets the **categorical track (C, #760)** and the **Markov track (D, #759)** build in parallel. It carves the discrete-state / count observation **plumbing** out of the front of Phase 4 so the Markov phases (4c/5/6) depend only on this thin slice — not on the whole categorical distribution phase.

No user-facing behaviour changes here: this is record types, reader routing, and output shapes only. The likelihood math, DSL blocks, and `simulate()`/`predict()` producers land in Tracks C/D.

## What changed

- **`types.rs`** — ungate `ObsRecord` and `Subject.obs_records` to **default-on**, keeping the `Event` variant `#[cfg(feature = "survival")]` (the same mixed-`cfg` shape `SimOutcome` already uses). Add `ObsRecord::{DiscreteState, Count}` and `SimOutcome::{Category, Count}` (+ `continuous_value()` arms).
- **`io/datareader.rs`** — new `ObsRouting { tte, discrete, count }` replaces the single `tte_cmts` set, giving 3 routing targets. Integer-DV routing for discrete-state / count CMTs with a non-integer, negative, and `u32`-overflow guard (mirrors the TTE integer-code rule, #192), plus a disjointness guard (`validate()`). Test-facing `read_nonmem_csv_filtered_routed` entry (no parser populates the discrete/count sets in Phase 4.0). The TTE wrappers keep their `tte_cmts` signature and wrap via `ObsRouting::tte_only()`, so `api::read_population_for` is untouched.
- **Ungate cascade** (the cost of making the plumbing default-on): 130 mechanical `#[cfg(survival)]` strips on empty `obs_records` field-inits, and **19 refutable-pattern fixes** — adding enum variants made `let ObsRecord::Event {..} = record;` refutable in survival code (`survival/mod.rs`, `stats/likelihood.rs`, `api.rs`) and tests (`tte_smoke.rs`, `tte_convergence.rs`). Each became `let-else { continue; }` (production loops skip non-Event records) or `panic!` (test assertions); one non-exhaustive `SimOutcome` match got a Category/Count arm. Behaviour-identical today — nothing produces non-Event records yet — and exactly the coexistence Tracks C/D will need.

## Alternatives considered

[...]

PR #896, opened 2026-07-20
feat(pk): dose any compartment on the analytical engine; superposition and event-driven paths agree [closes #375]

TeunP:
## Why

`#375` reported that an infusion into a compartment the analytical closed forms cannot deliver
into `panic!`s from inside the event-driven prediction walk. It was reachable from ordinary
NONMEM-format data — a positive `RATE` into `CMT=3` of a `two_cpt_oral` model, on a subject that
also has a time-varying covariate, an `EVID=3/4` reset, or IOV.

Fixing it surfaced two larger problems underneath, both silent:

**The three analytical paths disagreed on the same dose.**

| path | unroutable infusion | out-of-range bolus |
|---|---|---|
| dose superposition | ignores `cmt`, computes it as central | ignores `cmt`, silent |
| event-driven walk | `panic!` | `panic!` |
| `sens` gradient walk | `None => {}` — silently dropped | silently dropped |

The gradient one is the sharp edge: FOCE/FOCEI would have differentiated a *different dosing
history* than it predicted.

**Superposition never read `dose.cmt` at all.** `single_dose_concentration` branches only on
`is_infusion()` and picks the closed form from the *model*, so it placed every bolus in
compartment 1 and every infusion into central, whatever the data said. Which answer you got
depended only on whether the subject happened to carry a TV covariate / reset / IOV, since those
route to the walk, which places doses correctly. NONMEM confirms the walk and contradicts
superposition — by up to two orders of magnitude:

| case | NONMEM | walk | superposition (before) |
|---|---|---|---|
| `three_cpt_iv` bolus → periph-2 | 0.015194 | **0.015194** | 1.67266 |
| `two_cpt_iv` bolus → periph | 0.068016 | **0.068016** | 1.706286 |
| `one_cpt_oral` bolus → central | 1.8097 | **1.809675** | 1.19324 |

## What changed

**1. Reject what genuinely has no target.** `check_dose_compartments` (`src/api/validation.rs`),
a sibling of `check_modeled_dose_rates`, analytical-engine only. Infusion → the compartment must
have a rate channel; bolus → `cmt` must not exceed the state count. Reported once per
`(kind, compartment)` with subject/time context. Wired into `check_model_data` (so `fit()`
returns `Err`), the `assert_*` twin on `predict`/`simulate`/`suggest_start`, and the `Result`
form on the adaptive chokepoint, `run_covariance`, and `run_sir`.

[...]

Issue #891, opened 2026-07-17
Inner EBE loop converges to a sub-optimal local mode on multimodal individual posteriors (weakly-identified random effects)


roninsightrx:
## Summary

The inner EBE (empirical Bayes) loop can converge to a **sub-optimal local mode** when a subject's individual posterior is **multimodal** — which happens for weakly-identified random effects on nonlinear readouts. The returned η̂ is then not the global posterior mode, which biases that subject's FOCEI objective contribution (and its `H̃`) and makes EBE-based diagnostics unreliable for the affected subjects.

## Evidence

Found via a per-subject decomposition of the FOCEI objective against NONMEM (`METHOD=1 INTER`) on the 2-cpt fluconazole model with saturable protein binding (`ferx-testdata/fluconazole_radboudumc`), evaluated at the model's initial estimates (no outer optimization):

- **28 of 31 subjects match NONMEM's per-subject OBJ to < 0.005** — so ferx's per-subject FOCEI formula and sensitivities are correct.
- **3 subjects diverge**, all with a poorly-identified `V1` (ferx already warns 48% ETA_V1 shrinkage). ferx's inner loop lands in a different local mode than the global one:

  | subject | ferx η(CL, V1, BMAX) | global mode (NONMEM) | ferx OBJ | global OBJ |
  |---|---|---|---|---|
  | 14 | (−1.54, **+0.70**, +0.43) | (−1.27, **−1.93**, +0.44) | 92.1 | **70.2** |
  | 22 | (−1.17, **+0.37**, +0.58) | (−1.23, **−2.07**, +0.66) | 104.8 | **64.5** |
  | 26 | (−0.46, **+0.67**, +0.77) | (−1.46, −2.02, +0.78) | **22.1** | 33.6 |

  For subjects 14 and 22 ferx's mode has a **higher** individual objective than the global mode — i.e. the inner loop missed the better mode. (For 26, ferx's mode is genuinely *lower* than the one NONMEM reports — a further illustration that these posteriors are multimodal and mode selection is optimizer-path-dependent.)

- The mismatch is **not** inner-loop precision: init OFV is identical at `inner_tol` 1e-5 / 1e-10 / 1e-12, with `inner_restarts = 3`, and with a Nelder-Mead inner. The warm/cold BFGS descent deterministically lands in the nearest local mode; tightening tolerance just converges more precisely to the *same* wrong mode.

## Impact

[...]

PR #900, opened 2026-07-20
feat(categorical): binary endpoint simulate + predict + sdtab (Track C, Phase 4 Slice 1b) [closes #897]

### LINKED ISSUE #897: feat(categorical): binary endpoint simulate + predict (Track C, Phase 4 Slice 1b)
## Summary

**Slice 1b** of the Binary / logistic endpoint (Track C, Phase 4): `simulate()`, `predict()`
and diagnostics. Slice 1a (fit) shipped via #807 / #809 and closed #760; the simulation and
prediction half was always the second slice
(`plans/categorical-endpoint-phase4.md` §9) but was left without its own tracking issue when
#760 closed.

## The gap

A `[binary_model]` endpoint is **fit-only**. Its observations are `ObsRecord::DiscreteState`
rows in `obs_records`, and the Gaussian emitter walks `obs_times` — so the discrete records are
never visited and **`simulate()` / `predict()` return zero rows for them, silently and without
an error**. On the committed example (`examples/binary_logistic.ferx` +
`data/binary_logistic.csv`) that is 180 observations in, 0 rows out.

This is asymmetric with CTMM, which fails loud on `simulate()` (`api/simulate.rs` chokepoint
assert). Binary — a shipped, documented endpoint with a NONMEM/`glm` anchor — fails quiet.

Consequences: no VPC, no simulate-side validation, and in particular **no SSE round trip** —
which matters because every existing binary test scores *observed* data and would pass
unchanged if the sampler drew from the wrong probability (`plans/tte-survival-markov.md`
§8.8.8).

## Scope

Per `plans/categorical-endpoint-phase4.md` §3 Seam 4 / §5:

- [ ] **`Prediction` enum** (`plans/tte-survival-markov.md` §8.8.1) — a categorical prediction is
  a probability vector, which `PredictionResult { pred: f64 }` cannot carry. Does not exist
  today. **Shared prerequisite**: #820 (CTMM occupancy prediction) needs the same `CatProbs`
  variant, and its issue body points at #760 to coordinate — a pointer that now dangles into a
  closed issue.
- [ ] **`SimOutcome::Category` producer.** The variant has existed since #773 (Phase 4.0) with
  **zero producers**. Fixed grid, sampler only: `p = link⁻¹(lp@η)`, one uniform draw per record —
  no horizon, no event location, no Gillespie (binary is the easy row of §8.8.2).
- [ ] **`predict_categorical()`** returning `P(Y=k)` per record, alongside `predict_survival`.
  `predict()`'s own signature must stay untouched — the R wrapper binds it.
- [ ] **`--simulate` (synthetic design) support** — binary template rows + write-back of the
  drawn states into the output dataset, mirroring the TTE templates added in #522.
- [ ] **SSE round trip** (Tier 3): simulate from a known θ, refit, recover it.
- [ ] Docs: `docs/estimation/categorical.qmd` currently states the SSE guard "is added with the
  categorical simulation slice" — that promise is this issue.

## Deferred (not this slice)

- **sdtab / per-endpoint diagnostics.** `{model}-sdtab.csv` is Gaussian-shaped
  (IPRED/IWRES/CWRES) and emits no row for any non-Gaussian endpoint — TTE and CTMM included.
  Giving every non-Gaussian endpoint a diagnostics block (§8.8.5: martingale / deviance /
  Pearson) is its own piece of work, not binary-specific.
- Ordinal / Poisson / negative-binomial simulation — later slices, and Poisson additionally
  needs an `lgamma` builtin, which the expression evaluator does not have.
- VPC tooling (category-proportion bands) — downstream, likely ferx-r.

## Validation

The R `glm` / NONMEM `F_FLAG=1` anchors already committed under
`tests/reference/binary_logistic/` cover the fit side and are unaffected. The simulate side is
anchored by the **SSE round trip**, per the epic's license-free convention (§8.8.8, §14.11) —
NONMEM has no comparable "draw then recover" object to anchor against as a single artifact.

Refs: `plans/categorical-endpoint-phase4.md` §3 Seam 4, §5, §9; `plans/tte-survival-markov.md`
§8.8.1/§8.8.2/§8.8.8. Slice 1a: #760 (#807, #809). Shares the `Prediction` enum with #820.

TeunP:
## Why

A `[binary_model]` endpoint shipped **fit-only**. Its observations are `ObsRecord::DiscreteState`
rows in `obs_records`, and the Gaussian emitter walks `obs_times` — so the discrete records were
never visited and **`simulate()` and `predict()` both returned zero rows for them, silently and
without an error.**

Measured on the committed example (`examples/binary_logistic.ferx` + `data/binary_logistic.csv`):

```
subjects=60 obs_records=180 obs_times=0
simulate() returned 0 rows
predict() returned 0 rows
```

180 observations in, 0 rows out, on a shipped endpoint that already carries an R `glm` and a
NONMEM `F_FLAG=1` anchor. This is also asymmetric with CTMM, which fails *loud* on `simulate()`.

The consequence beyond the missing output: no simulate-side validation at all, and in particular
**no SSE round trip** — which matters because every existing binary test scores *observed* data
and would pass unchanged if the sampler drew from the wrong probability.

Closes #897.

## What changed

- **`Prediction` enum + `EndpointPredictionResult`** (`types.rs`, plan §8.8.1). A categorical
  prediction is a probability vector; `PredictionResult { pred: f64 }` cannot carry one. Only
  variants with a live producer are declared (`Continuous`, `CatProbs`) — the plan also sketches
  `Survival` and `Rate`, but TTE prediction already ships as the richer
  `SurvivalPredictionResult`, and `Rate` waits on the count slice, so adding either now would be
  an untestable dead arm.
- **`simulate_binary`** — the first producer of `SimOutcome::Category`, which has existed
  unproduced since #773. Fixed grid, `p = link⁻¹(lp@η)`, one `Open01` draw per record on the
  shared RNG stream. Emitted alongside `survival::simulate_tte` at the same seam, so a mixed
  PK + binary model emits both.
- **`predict_categorical()`** — new entry point alongside `predict_survival`. `predict()`'s
  signature is deliberately untouched (the R wrapper binds it) and remains the Gaussian
  predictor.
- **`--simulate` (synthetic design) support** — binary template rows, plus write-back of the
  drawn states into the output dataset. The write-back is an **in-place merge**, not the
  wholesale replace the TTE arm uses, so it composes with existing records instead of dropping
  them.
- **Fix: `predict()` on a CTMM model** returned no rows silently. The `simulate()` guard's
  message already *claimed* to cover `predict()`, but it sits in the simulate chokepoint only.
  Now fail-loud, pointing at #820.
- **sdtab diagnostics** (§8.8.5). One row per binary record: `DV` (observed 0/1), `PRED`
  (`P(Y=1)` at η = 0), `IPRED` (at the EBE η), `IWRES` (standardized Pearson residual). `CWRES`
  is blank — the conditional weighted residual is defined through the Gaussian residual-variance
  model, which a Bernoulli outcome does not have. Carried on a new
  `SubjectResult::discrete_rows` (additive, `skip_serializing_if` empty, so existing `FitResult`
  JSON is byte-identical).

### The sdtab append is deliberately not written the obvious way

Each sdtab column is built by its own independent pass over `subjects × ipred.len()`, and
`write_sdtab_csv` indexes every column by row — so a pass that failed to extend for the new rows
would **panic at write time**. Rather than edit all five passes and leave the same trap for the
next column added, `append_discrete_rows` extends *whatever columns exist*, keyed by name: known
names get a value, everything else gets NaN, which the writer already renders as an empty cell. A
column added later is therefore blank-by-default on discrete rows instead of truncating the table.

Also fixed while here: a binary-only fit has **zero** Gaussian rows, so the `CMT` column was
dropped by the existing multi-CMT heuristic — losing the only field identifying which endpoint a
row came from.

### The design decision worth reviewing

The likelihood, the sampler and the predictor now share **one** linear-predictor evaluator
(`for_each_binary_lp`), so they cannot disagree about which records belong to the endpoint, what
a `TIME` term resolves to, or what `lp` a record carries. Sampler/likelihood drift is the exact
failure the SSE exists to catch; sharing the evaluator makes that class of drift
*unrepresentable* rather than merely tested-for. The stable-`softplus` NLL form is preserved —
only `lp` is shared, not `p`, so no numerical stability is traded away.

## Alternatives considered

- **Extending `PredictionResult` with an optional probability vector** instead of a new enum:
  rejected — it is ferx-r-visible and would widen a struct every Gaussian caller allocates.
- **Declaring all four planned `Prediction` variants up front**: rejected — two would have no
  producer and no test, and the coverage gate would read them as untested new lines.

## Cross-repo dependency

No R change is required to merge. `Prediction` / `predict_categorical` are new *additive* surface
the R wrapper can pick up in a follow-up; nothing it currently binds changed shape.

## Breaking changes
- [x] Public Rust API (`api.rs` / `types.rs`) — **additive only**: new `Prediction`,
  `EndpointPredictionResult`, `DiscreteObsDiagnostic`, `predict_categorical`, and a
  `SubjectResult::discrete_rows` field. No existing signature or field changed, and the new field
  is `#[serde(default, skip_serializing_if = "Vec::is_empty")]` so serialized `FitResult` output
  for every existing model is byte-identical.
- [ ] None

## Numerical validation

- [x] Compared against: prior ferx behaviour (the fit side is untouched and its anchors still
  pass), plus a **simulate → fit recovery (SSE)** round trip for the new sampler.

The existing R `glm` / NONMEM `F_FLAG=1` anchors (`tests/reference/binary_logistic/`, OFV
213.5955) are unaffected. To be precise rather than claim "no fit-path change": `binary_data_term`
is refactored onto the shared evaluator, so the fit path *is* touched — but only in how it iterates
records, not in what it computes, and the hand-computed Tier-1 NLL tests plus the `glm` OFV anchor
both pin that. Everything else added here runs post-fit or in `simulate()`/`predict_categorical()`. Per the epic's convention (§8.8.8,
§14.11), the *simulate* side is anchored by SSE rather than NONMEM: there is no comparable
"draw then recover" object in NONMEM to anchor against as a single artifact.

```
# SSE: simulate 20 replicates x 180 records from a known theta, refit, recover
truth      TH0 = -0.6    THX = 1.1    THT = 0.4
recovered  within 0.15 (~3 SE at n = 3600)   -> PASS
```

**The SSE was mutation-checked rather than trusted.** Inverting the Bernoulli comparison
(`u < p` → `u > p`) makes it fail as intended:

```
theta[0] recovered 0.6675233672130765, generated from -0.6 (diff 1.2675233672130766)
```

## Performance
- [x] No significant impact expected — no fit-path code changed; the new work runs only in
  `simulate()` / `predict_categorical()`.

## Tests

- [x] Unit test(s) added in the same module — 10 Tier-1 in `src/categorical/mod.rs`: link
  stability and saturation at `±∞`, **sampler↔likelihood agreement**, draw calibration and
  exactness at degenerate `p`, Pearson residual, `Prediction::prob` accessor.
- [x] Regression test added that fails without this fix —
  `binary_simulate_emits_one_category_row_per_record` (the zero-rows bug), and
  `ctmm_predict_panics` for the CTMM guard.

19 Tier-2 in `tests/categorical_smoke.rs` (row counts, seed reproducibility + genuine variation,
normalized probabilities, documented Gaussian-`predict()`-is-empty behaviour, sdtab contents, and
the all-columns-equal-length invariant + a real `write_sdtab_csv` round trip to disk, plus
multi-endpoint draw determinism and the NaN-predictor skip added after review). Tier-3 SSE in
`tests/categorical_convergence.rs`.

An [xhigh review](https://github.com/FeRx-NLME/ferx-core/pull/900#issuecomment-5030815797) found
12 issues in the first two commits — three of them serious, including seeded draws that were not
reproducible across processes (`HashMap` iteration order) and a joint TTE + binary `--simulate`
that silently dropped every binary row. All are fixed in `9a880cc4`; the comment documents each
with its failure scenario. Counts above are post-fix and post-merge with main.

Coverage note: the gate runs `--features ci,survival` and skips Tier 3, so the sampler and the
predict path are each hit by fast Tier-1/2 tests, not left to the SSE alone.

```
--lib (ci,survival)        2787 passed, 0 failed
markov_smoke (ci,markov)     28 passed, 0 failed
tte_smoke (ci,survival)      86 passed, 0 failed
categorical_smoke            19 passed, 0 failed
categorical_convergence SSE   1 passed (slow-tests)
```

- [x] Example added to `examples/` or inline below — reuses the existing
  `examples/binary_logistic.ferx`; no new model file needed.

<details>
<summary>Example</summary>

```r
# simulate: one 0/1 draw per binary record, ipred carries p
sim <- simulate(model, data, params, n_sim = 100)

# predict: a probability vector per record, not a scalar
pred <- predict_categorical(model, data, params)
#> CatProbs { probs: [0.5, 0.5] }   at theta = 0
```

A `[simulation]` block driving a binary endpoint needs `times`, not a `horizon` — binary
outcomes are observed on the fixed grid. That is now a clear error rather than an empty result.

</details>

## Docs
- [x] `docs/` updated for user-visible changes — `docs/estimation/categorical.qmd` gains a
  simulation/prediction section and an SSE-validation section. It previously *promised* the SSE
  guard "is added with the categorical simulation slice"; this PR is that slice.
- [x] New pages linked in `docs/_quarto.yml` — no new pages; existing ones already linked.

## Checklist
- [x] `cargo clippy` — **no new warnings from this PR.** To be precise rather than tick a box:
  the repo has a pre-existing clippy backlog, and three warnings land in files I touched
  (`run.rs:710` unused-`mut` — a `markov`-feature-gate artifact; `simulate.rs:533`
  too-many-arguments and `:798` complex-type, both on `emit_subject_rows`, whose signature this
  PR does not change — it only adds a call inside the body). All three are on `main` already.
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, no `sens/` or `pk/`
  code touched; the binary endpoint routes to FD by design and this PR does not change that.

### Example execution
- [x] No example execution step needed beyond the above — the binary example is exercised
  directly by the Tier-2 `fit_from_files` test and the Tier-3 SSE, both run locally and green.

## Reviewer hints

**Focus here:**
1. `for_each_binary_lp` in `src/categorical/mod.rs` — the shared-evaluator refactor. It changes
   `binary_data_term`'s internals (fit path) even though the fit *result* is unchanged; the
   Tier-1 hand-computed NLL tests and the `glm` anchor both pin that.
2. The `--simulate` write-back in `src/api/run.rs`. Rows are paired per CMT in emission order.
   A short list (records skipped for a NaN predictor) leaves later templates untouched rather
   than shifting every row onto the wrong record — worth a second pair of eyes.
3. Whether `Prediction` is the right shape to hand to ferx-r, since #820 will reuse `CatProbs`
   for CTMM occupancy and that fixes the contract.

**Skimmable:** the Tier-2 assertions, the docs, and the CHANGELOG.

## Open questions

- **This PR unblocks #820** (CTMM simulate/predict), which needs exactly this `Prediction` enum —
  the dependency runs that way, not the reverse. Worth confirming `CatProbs { probs: Vec<f64> }`
  suits occupancy π(t) before it ships to R and the shape is fixed.
- **TTE / CTMM sdtab diagnostics are deliberately out of scope — filed as #901.** Binary's rows
  land here, and the append mechanism above is generic enough that TTE/CTMM rows drop in without
  touching the column passes. But the *residuals* are not a copy-paste: martingale/deviance has
  four dispatch paths (`{Analytic, OdeAccumulated} × {Single, Forward, Reset}`), and under
  **clock-reset RTTE** the cumulative hazard is evaluated per *gap* — so a `δ − H(t)` written for
  the single-event case is silently wrong there, in a way that looks entirely plausible on a plot.
  Left truncation, interval censoring (no `δ`), and the deviance formula's `0·ln(0)` traps each
  need their own handling. #901 scopes it single-event first, RTTE second.


TeunP:
## xhigh review — 12 findings, all resolved in `9a880cc4`

Reviewed with independent adversarial readers rather than self-assessment. Three findings were serious; two reviewers converged on the first with process-level proof.

### High

| Finding | Resolution |
|---|---|
| **Non-deterministic draws.** `simulate_binary` / `predict_binary` / `binary_diagnostics` consumed the RNG while iterating `model.endpoints` — a `HashMap`, randomized per process. With two binary endpoints the *same seed* gave different outcomes in different processes; sdtab row order varied run to run. | New `binary_endpoints()` helper over the already-sorted `binary_cmts()`, now the only path to a binary endpoint. The PR was inconsistent with itself: the `--simulate` template builder already used `binary_cmts()`; the sampler three lines away did not. |
| **Joint TTE + binary `--simulate` dropped every binary row.** The TTE write-back assigned `obs_records = events` wholesale, destroying the binary templates created moments earlier — reintroducing the exact silent-no-rows bug this slice exists to fix. | Retains non-`Event` records and splices the events in. |
| **sdtab TIME was the internal clock.** Discrete rows reported the occasion-shifted engine timeline while Gaussian rows report raw input TIME, so the two could not be joined; a reset subject emitted times absent from the input. Three doc comments asserted the opposite. | `ObsRecord::DiscreteState` now carries `raw_time` alongside `time`, as Gaussian rows always have. The predictor still evaluates on the internal clock; only reporting uses raw. |

### Medium / low

- `predict_categorical` was the one public entry point with **no guards at all** — a time-varying covariate on the linear predictor was silently frozen at baseline (#741's family). Now asserts like `predict()` and `fit()`.
- The Pearson residual computed `p*(1-p)`, which cancels to exactly zero once `inv_logit` saturates (`lp ≳ 36.7`), blanking the **largest** residuals — and only on the positive tail. Now takes `q = inv_logit(-lp)` independently; a new Tier-1 test pins tail symmetry out to `lp = 50`.
- Simulate write-back keyed by a per-CMT queue, which misaligned every later row after a skipped record — the opposite of what its own comment claimed. Now keys by `(cmt, raw_time)`.
- `predict_binary` missing the state clamp its two siblings apply; `N_OBS` showing a misleading `0` on discrete rows (blank now); `binary_diagnostics` walking records twice; `.fitrx` comment claiming a recompute that does not exist.
- Test hygiene: fixed temp filename → `tempfile::tempdir()`.

### One thing worth reading carefully

I introduced a bug while fixing one and caught it: the first single-pass `binary_diagnostics` evaluated PRED under `with_model_time(raw_time)` while IPRED used the internal clock — which would have made PRED wrong for precisely the reset-occasion subjects the third finding is about. The record struct now carries both clocks explicitly.

And both new Tier-2 tests initially **passed while asserting nothing** — they carried `return` escape hatches. A probe showed the DSL evaluator guards every arithmetic route to NaN (`log(0)`, `0/0`, `sqrt(-1)`, `TIME/TIME`), so the reachable path is a NaN *covariate*. Rewritten around that, hatches removed.

**Limitation stated plainly:** the ascending-CMT-order assertion catches a HashMap regression only probabilistically (~50% per run with two endpoints), since within one process the map's order is stable. An in-process test cannot reproduce cross-process hash randomization. The real guard is structural — `binary_endpoints()` is the only path, and it says so.

### Validation

```
fmt                                          clean
cargo check --all-targets  ci / ci,markov / ci,survival,slow-tests   clean
--lib (ci,survival)                          2747 passed, 0 failed
categorical_smoke                              19 passed, 0 failed
markov_smoke                                   28 passed, 0 failed
tte_smoke                                      86 passed, 0 failed
```

TeunP:
## Round-2 xhigh review — 1 HIGH + 2 MEDIUM, fixed in `d785a14a`

A second adversarial pass (fresh lenses: audit-the-fixes, and cross-cutting/merge integration) found that **round 1's fixes introduced two new defects**, and that the headline feature was broken in a way neither round-1 review nor green CI caught.

### HIGH — binary-only `--simulate` could not work at all

Reproduced, not inferred:

```
$ ferx binonly.ferx --simulate
Loaded 4 subjects, 24 observations
Error: [error_model] has no entry for observed compartment(s) 1
```

The synthetic subject materialised `observations` / `obs_cmts` on CMT 1 unconditionally, while this PR's own new error *requires* a binary model to supply `times`. Supplying them fabricated phantom Gaussian rows that validation then rejected. **No workaround existed** — adding an `[error_model]` stops the model being binary-only. TTE + binary was hit too: a working TTE `--simulate` broke the moment a binary endpoint was added.

Fixed by building the Gaussian grid only when the model has a continuous endpoint. **Mutation-verified**: reverting the fix fails the new test.

Worth recording how it got through. Codecov flagged `src/api/run.rs` at **10%**. I correctly diagnosed the *number* as a premature grade — the fast coverage job runs `--features ci`, without `survival`, so these lines can't execute there — and then used that to stop thinking about `run.rs` coverage entirely. The explanation was right; the conclusion drawn from it was wrong. Nothing tested `run_model_simulate` with a binary model, and that is exactly where the bug was.

### MEDIUM — both introduced by round 1's own fixes

| Defect | Detail |
|---|---|
| **Duplicate `times` collapsed onto one draw** | Round 1 replaced a per-CMT queue with a map keyed by `(cmt, raw_time)` to fix a misalignment bug. `times = [0.5, 0.5]` is legal and un-deduped, so both records shared one state and a Bernoulli draw was silently discarded. Now a `VecDeque` per key — duplicates each get their own draw, identity keying still survives a skip. |
| **A skipped record was fit as observed data** | `simulate_binary` skips a NaN-predictor record; its template kept placeholder `state = 0`, which passed `validate_binary_states` and was then scored by the Bernoulli likelihood as a genuine observed failure. Round 1 traded a misalignment bug for a **data-fabrication** bug. Un-drawn templates are now removed, not stamped. |

### Test quality

The determinism test's primary assertion was a **tautology** — `assert_eq!(first, run())` in one process, against one `HashMap` instance, cannot fail whether or not the fix is present. Only the ordering assert had power, at ~50% with two endpoints. Rewritten against **five** endpoints, so a hash-order regression escapes with probability 1/120.

### Claims corrected rather than left standing

- CHANGELOG said "`predict()` itself is unchanged" — false, it gained a CTMM panic, which the same CHANGELOG discloses two bullets down.
- "the simulated states are written back to the output dataset" — misleading; `run_model_simulate` returns an in-memory `Population`, no file is written.
- Docs said the likelihood, sampler and predictor "cannot disagree" — overstated. They deliberately differ at `lp = ±∞`: the sampler treats it as a valid degenerate `p`, the likelihood repels the optimizer. Now documented.
- Added the missing `### Changed` entry for `ObsRecord`'s source-breaking new field.

### Also

`ObsRecord::Count` gained `raw_time` so the Poisson slice cannot reintroduce the internal-clock reporting bug. The "no per-call allocation" comment was false after round 1 and now explains why the allocation is deliberate (HashMap-order summation would make the OFV ULP-nondeterministic across processes).

**Clean on re-review:** the merge with main (no dropped hunks; #891 multistart and #896 compartment-dosing interactions verified safe), `raw_time` threading, retain+splice ordering, the Pearson fix, and `with_model_time` nesting (the guard drops before the callback runs).

**Filed, not fixed here:** FeRx-NLME/ferx-r#271 — ferx-r maps every simulated row through `continuous_value()`, so binary rows return `NA` with no `CMT` column to distinguish them. The feature is unusable from R until the glue is updated.

```
fmt                                                    clean
cargo check --all-targets  ci / ci,markov / ci,survival,slow-tests   clean
--lib                                     2787 passed, 0 failed
categorical_smoke                           21 passed, 0 failed
markov_smoke                                28 passed, 0 failed
tte_smoke                                   86 passed, 0 failed
```

ronkeizer:
## Code review — PR #900 (medium effort)

8 finder angles + direct verification. Findings ranked most-severe first. The refactor onto `for_each_binary_lp` and the SSE round-trip are solid; issues cluster where the new binary machinery meets the *other* discrete endpoint (CTMM) and the `--simulate` write-back edge cases.

### 1. 🔴 New `write_sdtab_csv` guard is a regression for live CTMM (and Count) fits — `src/io/output.rs:1322`

The guard fires on `pop_has_discrete && !result_has_discrete`. But:
- `markov = ["survival"]`, so a CTMM fit compiles this `#[cfg(feature = "survival")]` block.
- CTMM observations are `ObsRecord::DiscreteState` (`src/markov/endpoint.rs`), so `pop_has_discrete = true`.
- `binary_diagnostics` only populates `discrete_rows` for `EndpointLikelihood::Binary` (`src/categorical/mod.rs:414`, `else { continue }`), so a CTMM fit produces **zero** `discrete_rows` → `result_has_discrete = false`.

So a normal, non-restored CTMM fit always trips the guard. `write_sdtab_csv` returns `Err` and the CLI (`src/bin/run_model.rs:189`) prints `Warning: failed to write sdtab` and writes **no** sdtab — with a message blaming a `.fitrx` checkpoint restore that never happened. Before this PR the guard didn't exist and CTMM fits wrote an sdtab. A mixed Gaussian+CTMM model loses its valid Gaussian rows too. No markov test calls `write_sdtab_csv`, so it's uncaught.

**Fix:** scope `pop_has_discrete` to the CMTs an endpoint *should* produce diagnostics for (Binary), or exclude CTMM/Count CMTs from the check — distinguish "diagnostics were dropped in round-trip" from "this endpoint kind never produces them yet."

### 2. 🟠 All-NaN-predictor subject leaves fabricated placeholder `state=0` rows in `--simulate` — `src/api/run.rs:561`

The comment right above (lines 545–550) states a draw-less template must be *removed*, not left at its placeholder, because a `state = 0` placeholder passes `validate_binary_states` and is then scored as a genuine observed failure — "fabricated data." But the removing `retain_mut` is nested inside `if !drawn.is_empty()`. When a subject gets **zero** draws (every binary record NaN-skipped — e.g. an `lp` referencing a covariate the synthetic `[simulation]` subjects don't carry, since they're built with `covariates: HashMap::new()`), `drawn` is empty, the block is skipped, and all placeholder `0` rows survive and are fit as observed zeros — exactly the outcome the comment says the removal prevents. Partial-NaN subjects are handled; only the all-NaN subject leaks.

**Fix:** drop the `if !drawn.is_empty()` guard and always run `retain_mut` (empty `drawn` → all draw-less rows correctly dropped).

### 3. 🟠 `predict()` `assert!`-panics on any model carrying a CTMM endpoint, including mixed Gaussian+CTMM — `src/api/predict.rs:77`

The fail-loud intent is reasonable for CTMM-only, but `assert!(!model.has_ctmm(), ...)` is unconditional and a hard panic:
- A **mixed** continuous + CTMM model previously returned its Gaussian predictions (silently omitting CTMM); it now panics the whole call, dropping predictions it can legitimately produce.
- `nca_sweep` init calls `predict()` (`src/suggest_start/sweep.rs:19`), so `--inits-from-nca=nca_sweep` on a model with a CTMM endpoint panics *during initialization*, before fitting.
- It's a `panic!`, not `Err` — `predict()` is public API the R wrapper binds; unwinding across FFI is worse than a returned error.

**Fix:** return `Err` rather than `assert!`, and only bail when CTMM is the *only* endpoint (skip CTMM CMTs otherwise) so mixed models keep working.

### 4. 🟡 Binary write-back `retain_mut` drops `DiscreteState` rows on *any* non-binary CMT (latent) — `src/api/run.rs:562`

`retain_mut` matches every `DiscreteState` record, then looks up `(cmt, raw_time)` in `drawn`. A CTMM `DiscreteState` row misses and is **dropped**. Unreachable today only because CTMM `--simulate` is rejected upstream — but it becomes active data-loss the day #820 lands, the same silent-empty class this PR exists to kill. Scope the `retain_mut` match to binary CMTs now while it's cheap.

### 5. 🟡 `binary_endpoints()` heap-allocs + sorts on the FOCEI hot path — `src/categorical/mod.rs`

`discrete_subject_nll` now routes through `binary_endpoints(model)` → `model.binary_cmts()`, which allocates a fresh `Vec<usize>` and `sort_unstable`s it on every call. It runs once per subject per outer iteration *and* across the FD-Hessian stencil (`stats/likelihood.rs`, O(n_eta) evals/subject/iter) — thousands-to-millions of tiny allocs over a fit where the pre-PR `binary_data_term` allocated zero. Determinism (the stated reason for sorting) doesn't require per-call allocation: compute the sorted CMT list once (cache on `CompiledModel`, or hoist above the closure) and iterate the slice.

### 6. 🔵 `inv_logit` duplicates the private `nn::sigmoid` — `src/categorical/mod.rs:39`

Byte-for-byte identical stable logistic (`if x>=0 { 1/(1+e^-x) } else { e/(1+e) }`) already exists at `src/nn/mod.rs:151`. Two copies drift on any future tail-bound fix. Make `nn::sigmoid` `pub(crate)` and call it. (Note: not the same as `PkNum::inv_logit` in `sens/num.rs`, which is the naive non-stable form.)

### 7. 🔵 `append_discrete_rows` keys columns by runtime name string — `src/io/output.rs:1268`

Columns are filled via `match name.as_str() { "IPRED" => …, _ => NaN }`. A renamed/localized Gaussian column silently fills NaN (blank) on discrete rows with no compile-time check; a user `[output]` column colliding with a reserved name (`PRED`, `DV`, …) gets the discrete diagnostic value instead of a blank. Low-likelihood but no type guards it. The deeper form is a per-row record the writer iterates once, rather than N independent column passes plus a string-keyed catch-up.

---
*Angles A/B/C (correctness) independently converged on #1; #5–7 are cleanup. Conventions check found no clear CLAUDE.md violation (the diagnostics are deterministic functions of a NONMEM/glm-anchored p, and the epic permits SSE anchoring for the simulate side).*

TeunP:
@ronkeizer — thanks, this was the review that mattered. Two of your three severe findings were regressions **this PR introduced**, one of them about an hour before you posted. All addressed in `0e8b736c`.

### 1 🔴 sdtab guard breaks live CTMM — **removed, not narrowed**

You're right and the diagnosis is exact: CTMM records are `DiscreteState`, only binary populates `discrete_rows`, so every healthy CTMM fit tripped it and the CLI wrote no sdtab while blaming a restore that never happened.

I tried your suggested fix first — scope `pop_has_discrete` to binary CMTs — and it **can't be done there**. `write_sdtab_csv` receives only `(result, population, path)`, and `FitResult` carries no endpoint metadata, so the function cannot distinguish "diagnostics were dropped in round-trip" from "this endpoint kind never produces them". Rescuing it means adding public API for a stopgap whose real fix is #911.

So I withdrew the guard entirely. It traded a shipped, NONMEM-anchored feature for a warning about a case #911 already tracks — a bad trade. Added `ctmm_fit_still_writes_an_sdtab` to `markov_smoke`, and noted on #911 that the fail-loud variant needs endpoint metadata at the sdtab boundary and should land *with* the round-trip rather than before it.

Your point that no markov test called `write_sdtab_csv` is the real lesson — verified, zero occurrences at the previous HEAD. That's why neither of my review rounds nor CI saw it.

### 2 🟠 all-NaN subject keeps placeholders — fixed

Correct, and worse than it looks: the comment directly above the `retain_mut` *promises* draw-less templates are removed precisely so a `state = 0` placeholder isn't scored as a genuine observed failure. Nesting it under `if !drawn.is_empty()` meant the all-NaN subject got exactly the fabricated data the comment ruled out. Now unconditional, as you suggested.

### 3 🟠 `predict()` panics on mixed Gaussian + CTMM — fixed

Also a regression of mine. Now fires only when there is no continuous endpoint — i.e. when the call would return an empty vec, which is the confusing case the assert was for. A mixed model gets its Gaussian rows, exactly as it does with a binary endpoint.

### 4 🟡 `retain_mut` would drop CTMM rows — fixed now, as you advised

Scoped to binary CMTs. Cheap today, silent data loss the day #820 lands.

### 5 🟡 hot-path allocation — fixed

`discrete_subject_nll` now takes a single-endpoint fast path straight out of the map (every real model today), so the sorted `Vec` is built only in the multi-endpoint case where deterministic summation order actually buys something. You were right that determinism doesn't require per-call allocation; I'd defended it as a necessary cost and it wasn't.

### 6 🔵 duplicate `inv_logit` — **declined, with a reason**

`nn` and `survival` are independent cargo features (`src/lib.rs:14` / `:4`), so `categorical` cannot call `nn::sigmoid` — neither module is guaranteed to be compiled alongside the other. Sharing it needs an ungated math home, which means touching `nn` in a categorical PR. I cross-referenced both copies with an explicit "change one, change the other" note so the drift you're worried about is at least discoverable. Happy to do the shared-home move as a cleanup if you'd rather.

### 7 🔵 string-keyed column fill — deferred to #901

Agreed on the deeper form. It's the same restructure as the row-contiguity question (all Gaussian rows, then all discrete), which is only decidable once TTE/CTMM rows exist and there are three families competing for row order — noted on #901 so it's decided once rather than three times.

---

```
fmt                                             clean
cargo check --all-targets  ci / ci,markov / ci,survival,slow-tests   clean
--lib                              2787 passed, 0 failed
categorical_smoke                     22 passed, 0 failed
markov_smoke                          29 passed, 0 failed   (+ctmm sdtab regression)
tte_smoke                             86 passed, 0 failed
```

TeunP:
Re-review requested from @roninsightrx — noting that the review above was posted from @ronkeizer, which isn't a collaborator on this repo, so a formal review request can't be routed to it (`gh` accepts the handle and silently drops it). Tagging both so it lands either way.

Specifically worth a second look, since these are fixes-to-findings and that activity has produced regressions twice in this PR already:

- the sdtab guard **removal** (rather than the narrowing you suggested) — I could not make it correct at that boundary, reasoning in the reply above; #911 now carries the constraint
- the unconditional `retain_mut` — it now deletes any binary-CMT `DiscreteState` record that received no draw, so it is worth confirming there's no legitimate path where a template should survive without one
- the `predict()` assert's new condition (`!has_ctmm() || sigma non-empty`) — using "has a continuous endpoint" as the proxy for "has something to return"

roninsightrx:
## Re-review — fix commit `0e8b736c`

Thanks @TeunP — went through the fix commit at high effort (workflow-backed, independent verify per finding). **Two of the five substantive fixes landed clean; three attempts don't fully hold.** Details below, ranked.

### ✅ Fixed correctly
- **All-NaN placeholder rows** (`run.rs`) — `if !drawn.is_empty()` gate removed, `retain_mut` now unconditional. Draw-less subjects no longer keep fabricated `state=0` rows. 👍
- **write-back CMT scope** (`run.rs:579`) — `if !binary_cmts.contains(cmt) { return true; }` correctly stops the binary write-back from deleting CTMM `DiscreteState` rows once #820 lands. 👍
- Declining the `inv_logit`/`nn::sigmoid` merge (independent cargo features) and deferring the string-keyed sdtab column fill to #901 are both reasonable, and the cross-reference comment is the right stopgap.

### ⚠️ 1. sdtab guard removed wholesale reopens the fitrx-restore gap — `src/io/output.rs:1304`
Removing the guard fixes the false rejection of live CTMM fits, but it also drops the protection for its *original* target: a **mixed Gaussian+binary** fit restored from a `.fitrx` checkpoint (`io::fitrx::parse_subjects` rebuilds `discrete_rows: Vec::new()`) now writes an sdtab with all Gaussian rows present and **every binary row silently missing** — no error. A binary-only restored fit writes a header-only table (the `cols.is_empty()` check doesn't fire — columns exist, they're just zero-length).

The commit argues scoping needs endpoint metadata that `write_sdtab_csv(result, population, path)` doesn't carry. Agreed that per-CMT scoping is out of reach here — but the record-variant test was the wrong key anyway. The clean discriminator is **"was this result restored from a checkpoint?"**, not "does it have discrete records": that separates CTMM-live (not restored → allow) from binary-restored (restored → reject) exactly, with no endpoint metadata. If `FitResult` (or the restore path) can carry a `restored_from_checkpoint` flag, guard on that. Deferring the full fix to #911 is fine, but a mixed-fit restore silently dropping binary rows is worth at least that lightweight guard now.

### ⚠️ 2. `predict()` CTMM guard uses a fragile sigma proxy + drops rows with no warning — `src/api/predict.rs:79`
`!model.has_ctmm() || !model.default_params.sigma.values.is_empty()` uses "sigma non-empty" as a stand-in for "has a continuous endpoint." Two gaps:
- **Imprecise proxy:** a CTMM-only model that declares a Gaussian residual-error block (non-empty `sigma`) but no Gaussian *structural* endpoint skips the assert and returns an **empty vec** silently — the very case the assert exists to catch. Check the endpoint set directly (`model.endpoints.values().any(|e| matches!(e, EndpointLikelihood::Gaussian{..}))`) instead of inferring it from sigma.
- **Silent omission:** for a mixed model, CTMM records are now dropped from the prediction table with nothing pushed to `FitResult.warnings` — the repo convention (CLAUDE.md "Warning and Error Conventions") is to collect non-fatal issues there. A caller expecting occupancy rows gets a short table and no signal. Push a warning when CTMM CMTs are omitted.

### ❌ 3. The FOCEI hot-path allocation is not actually removed — `src/categorical/mod.rs:226`
The fast path opens with `let cmts = model.binary_cmts();` — and `binary_cmts()` does `.collect()` + `sort_unstable()` (`types.rs:3102`), i.e. it **allocates and sorts a `Vec` on every call**, which is the exact allocation the comment says it avoids ("no reason to build the sorted CMT list, so read it straight out of the map"). Net effect:
- **Single binary endpoint** (every real model today): one alloc+sort per call — unchanged from before, so the optimization is a **no-op**.
- **Multi binary endpoint**: `[only]` fails, control falls through to `for (cmt, endpoint) in binary_endpoints(model)`, and `binary_endpoints` *also* calls `binary_cmts()` (`mod.rs:139`) — so now **two** allocs+sorts per call, a **pessimization**.

`discrete_subject_nll` runs per subject per FOCEI inner iteration and per FD-Hessian stencil point, so this is the hot path. To actually remove the alloc: detect the single-endpoint case without collecting (e.g. iterate `model.endpoints` and short-circuit, or count `Binary` endpoints), or cache the sorted CMT list once on `CompiledModel`.

### 🔵 4. Nit — dead re-destructure — `src/categorical/mod.rs:229`
`if let Some(endpoint @ EndpointLikelihood::Binary { .. })` already proves the variant, then it's re-matched with `let EndpointLikelihood::Binary { link, lp_fn, .. } = endpoint else { unreachable!() }`. Bind `link`/`lp_fn` in the outer pattern and drop the `unreachable!` block.

---
*#2 and #4 from the prior round are resolved. #1/#3/#5 above are the attempts that need another pass.*

TeunP:
@roninsightrx — thorough, thank you. You were right on all three, and #3 was the one I'd defended as a fix when it wasn't. Fixed in `000f8d37`; I verified each against the code this time rather than reasoning about it.

### ❌ 3. Hot-path allocation — you're exactly right, it was a no-op / pessimization

`let cmts = model.binary_cmts()` allocates + sorts on every call — the precise cost the comment claimed to avoid. Single endpoint: one alloc, unchanged. Multi: two allocs (the fallback calls `binary_cmts()` again). I wrote a comment asserting an optimization the next line contradicted.

`discrete_subject_nll` now probes `model.endpoints` directly with two `next()` calls — no allocation for the single-endpoint case (every real model), and the sorted `binary_endpoints` walk only when there are genuinely ≥2 binary endpoints, where the deterministic order actually earns its keep. The dead `unreachable!()` re-destructure (your #4) is gone with it.

### ⚠️ 2. `predict()` sigma proxy — fixed, but not with the endpoint-set check

Your diagnosis is right; the specific fix you suggested isn't available in this codebase, though. `model.endpoints.values().any(Gaussian)` would be **false for every plain PK model** — I checked, and `EndpointLikelihood::Gaussian` is never inserted into `endpoints` for a PK model (the only production inserts are TTE / binary / CTMM; the Gaussian variant is constructed only in tests). That check would reject every healthy mixed PK + CTMM model at the assert.

The actual determinant of whether `predict()` returns rows is whether any subject has continuous observations — `obs_times`. That's what I guarded on. Your counterexample (CTMM-only + `[error_model]` → non-empty sigma) is now handled correctly: its `obs_times` are empty, so the assert fires.

On the **silent-omission** half: `predict()` returns `Vec<PredictionResult>` and has **no `warnings` channel** — the repo convention you cite (collect into `FitResult.warnings`) can't apply because `predict()` produces no `FitResult`. Pushing a warning would need a signature change, which is ferx-r-visible. I'd rather surface that than invent a sink — do you want the signature widened (e.g. return warnings alongside), or is the mixed-model behaviour (Gaussian rows returned, CTMM absent, consistent with how binary behaves) acceptable as-is? Leaving it for your call rather than guessing.

### ⚠️ 1. sdtab guard — restored, keyed on the flag you identified

Exactly your point: the discriminator is "was this restored from a checkpoint", not the record variant. Added `FitResult::restored_from_checkpoint` (`#[serde(skip)]`, so no `.fitrx` / JSON / ferx-r format change; set only in `fitrx::load_fit`, `false` everywhere else) and the guard keys on it. Live CTMM writes sdtab again (`markov_smoke::ctmm_fit_still_writes_an_sdtab`, the regression test that was missing); a restored binary fit is refused (`categorical_smoke::sdtab_refuses_a_restored_fit_missing_its_discrete_rows`). #911 removes the guard once diagnostics round-trip.

### Declined, with reasons

- **`inv_logit` / `nn::sigmoid` merge** — can't call across the two: `nn` and `survival` are independent cargo features (`lib.rs:14` / `:4`), so neither module is guaranteed compiled alongside the other. Needs an ungated math home, which is a `nn`-touching cleanup. Cross-referenced both copies with a "change one, change the other" note as the stopgap. Happy to do the shared-home move separately if you want it.
- **String-keyed sdtab column fill** — agreed on the deeper form; it's the same restructure as the row-contiguity question (Gaussian rows then discrete rows), deferred to #901 to decide once TTE/CTMM rows exist rather than three times.

### One honest caveat on local validation

`--lib` shows **2847 passed, 1 failed** locally — `test_slsqp_moves_on_mu_referenced_two_cpt_oral_cov`. I verified it is **not** from this PR: it fails identically (`SLSQP OFV = -1132.16`) at the previous green-CI commit `0e8b736c`, the test file is unchanged from `main`, and CI runs `--lib` on `ubuntu-latest` where SLSQP reaches the expected ≈-1182. It's a macOS-vs-Linux SLSQP convergence divergence (the test's own comment cites the Linux number), unrelated to these changes. Flagging it rather than quietly reporting "all green" — it does not affect the CI result, but it is a real local failure and you should know I saw it.

```
categorical_smoke   23 passed, 0 failed   (+restored-guard, +multistart-stability)
markov_smoke        29 passed, 0 failed   (+ctmm sdtab regression)
--lib             2847 passed, 1 failed   (the pre-existing macOS SLSQP divergence above)
cargo check --all-targets  ci / ci,markov / ci,survival,slow-tests   clean
```

TeunP:
Follow-up to the re-review — one gap I caught in my own fix, plus your open question.

### A gap the re-review didn't flag, but I found reviewing my own fix (`539124db`)

The restore guard from `000f8d37` has a test (`sdtab_refuses_a_restored_fit_missing_its_discrete_rows`) — but that test sets `restored_from_checkpoint = true` **by hand**. So it exercises the guard, and never checks that `load_fit` actually sets the flag. If the wiring broke (wrong literal, a bypassed path), a genuinely restored fit would slip straight through the guard and no test would notice — the "guard tested, wiring untested" split that's bitten this PR more than once.

Added `restored_flag_is_set_only_by_load_fit` in `fitrx.rs`: a real save → load round trip asserting the flag is `false` on a freshly built result and `true` after `load_fit`. The full chain (`load_fit` sets it → `write_sdtab_csv` refuses on it) is now covered end to end, not just the last link.

### Your open question — the `predict()` warning for omitted CTMM rows

You asked for a warning when a mixed model drops its CTMM rows. I looked at wiring it and it needs a decision from you, because `predict()` returns `Vec<PredictionResult>` — there is **no `warnings` channel** on that path (unlike `fit()`, which has `FitResult.warnings`). So the CLAUDE.md "collect non-fatal issues into warnings" convention has nowhere to land here without changing the signature, which is ferx-r-visible.

Three options, your call:
1. **Leave it** — a mixed model returns Gaussian rows and silently omits CTMM, which is exactly how a mixed model with a *binary* endpoint already behaves (`predict()` is the Gaussian predictor; non-Gaussian endpoints have their own predictors, `predict_survival` / `predict_categorical`, and CTMM's is #820). Consistent, no new surface.
2. **Widen the signature** to return `(Vec<PredictionResult>, Vec<String>)` or similar — the honest place for the warning, but a breaking change for ferx-r.
3. **Defer to #820**, where the CTMM occupancy predictor lands and the "you called the wrong predictor" guidance belongs.

I'd lean (1)+(3): the behaviour is already consistent with binary, and #820 is where CTMM prediction — and any guidance about it — properly belongs. But it's your call; I didn't want to either invent a warnings sink or silently drop the request.

### Honest status

I'm deliberately not saying "it's good now" — this PR's pattern has been that each fix round surfaces something, and self-review has missed what your reviews caught every time. The bug class has narrowed (new-code correctness → fix correctness → now a test-coverage gap I caught myself), which is the right direction, but I'd value one more focused pass from you on the `000f8d37` + `539124db` delta specifically rather than treating it as done.

```
categorical_smoke   23 passed, 0 failed
markov_smoke        29 passed, 0 failed
fitrx wiring test    1 passed  (load_fit sets the flag)
--lib             2848 passed, 1 failed — the pre-existing macOS-only SLSQP
                  divergence (fails identically at 0e8b736c; green on Linux CI)
cargo check --all-targets  ci / ci,markov / ci,survival,slow-tests   clean
```

TeunP:
## Self-review — xhigh (independent line read + 3 adversarial verifiers + build/test)

Reviewed the whole diff line-by-line, ran three independent adversarial passes (simulate/write-back, sdtab/serde, tests/conventions), and built + ran the fast tiers on `ci,survival,markov`. Not reporting CI-green as correctness — every claim below was traced to code or reproduced.

**Verdict:** the production logic is sound; the shared `for_each_binary_lp` evaluator, the `raw_time` split, the RNG ordering and the sdtab column invariant all check out. Two real defects surfaced — both fixed in this push — and four test-adequacy holes that let plausible mutations ship green — all closed.

### Verified clean
- `raw_time` genuinely carries the **user** clock: datareader sets `raw_time = time` *before* `time += time_offset`, so reset-occasion rows join back to the input; internal `time` drives the predictor. Not a no-op.
- `--simulate` write-back pairing: identity-keyed (`(cmt, raw_time)`) `VecDeque` FIFO aligns the k-th draw to the k-th duplicate-time row; `to_bits()` is robust where `==` isn't.
- RNG determinism: `binary_endpoints` sorts (`binary_cmts()`), `simulate_binary` runs last and consumes **zero** randoms when no binary endpoint exists → existing seeded Gaussian/TTE streams byte-identical.
- sdtab: every column extended uniformly by `append_discrete_rows`; ID/TIME expressions identical to the Gaussian pass; CMT forced on via `any_discrete`. Column-length invariant holds for binary-only, mixed, and restored fits.
- TV-covariate guard **does** cover `Binary { lp_covariates }` (validation.rs) — `predict_categorical`'s guard is real.
- serde: `discrete_rows` (`skip_serializing_if`) + `restored_from_checkpoint` (`skip`) keep existing `FitResult` JSON byte-identical.

### Defects found and fixed (this push)

| # | Sev | Location | Problem | Fix |
|---|-----|----------|---------|-----|
| 1 | High (latent) | `io/output.rs` `write_sdtab_csv` | The `restored_from_checkpoint` refusal keyed on `pop_has_discrete` — **always false on a real restore**. `load_fit` re-reads the bundle through `read_nonmem_csv_mapped`, which hard-codes `ObsRouting::default()` (empty discrete set), so binary DVs come back as *Gaussian* observations and no `DiscreteState` survives. The guard never fired on the path it targets, and its test passed only on a hand-built population `load_fit` can't produce (real trigger: an external `load_fit → write_sdtab_csv`, e.g. ferx-r). | Key on the restored **model's** declared binary endpoints (`model_text` → `parse_full_model` → `binary_cmts()`), which is exactly "would a live fit have produced `discrete_rows`?". CTMM (no binary endpoint) is correctly still allowed. Test rewritten to exercise the real path (pop with no discrete rows) + a no-false-positive case. |
| 2 | Low–Med | `api/run.rs` `run_model_simulate` | A binary-only `--simulate` whose logit is NaN for a **whole** subject emits zero sim rows → `sims_by_id.get(id)` misses → the per-subject `continue` skips the placeholder cleanup → every template `state = 0` reaches `fit()` as **fabricated observed data**. Narrow trigger: the logit must evaluate NaN with no covariate (synthetic subjects have none) — reachable via unguarded `Op::Pow`, e.g. `TH0 ^ 0.5` at `TH0 < 0` (`powf` of a negative base). Verified: pre-fix that model survives 12 placeholders into a converged fit; post-fix, 0. The code comment claimed the cleanup "runs unconditionally"; the outer `continue` gated it. | Bind an empty `sims` instead of `continue`-ing, so the drop reaches the zero-draw subject. Regression test added. Comment corrected. |

### Test-adequacy gaps closed

| Gap | Why it mattered | New test |
|-----|-----------------|----------|
| Draw direction `u < p` only pinned by the **slow-gated** SSE; the Tier-1 test asserts on a *local re-implementation*, not production `simulate_binary`. Inverting the production draw shipped green on the PR job. | The PR's whole thesis is "sampler drift is what the SSE catches" — but it doesn't run per PR. | `production_binary_draw_frequency_tracks_p` drives the real path at p=0.8; inverted → ~0.2 → fails fast. |
| `raw_time ≠ time` (the marquee "Changed" field) tested **nowhere** — every fixture sets them equal, so a `time`↔`raw_time` swap in `for_each_binary_lp` passes the whole suite. | The reset-occasion behavior the field exists for was unverified. | `binary_predict_uses_internal_time_and_reports_raw_time` (internal 4.0, raw 1.0). |
| No mixed Gaussian+binary through `simulate()`. | The dual-emit seam in `emit_subject_rows` was untested. | `mixed_gaussian_and_binary_simulate_emits_both`. |
| sdtab PRED only range-checked (`∈ [0,1]`); a PRED computed at the EBE instead of η=0 stays in range and passes. | PRED (η=0) vs IPRED (EBE) not distinguished. | `binary_sdtab_pred_is_population_typical_not_ebe` matches PRED against `predict_categorical` (η=0 by contract), row-for-row. |

### Left as-is (noted, low)
- `predict_categorical` on a NaN predictor returns `[NaN, NaN]` silently, where `simulate_binary` skips+warns. Defensible (predict is a deterministic transform, no draw to fabricate) — flagging, not changing.

---
Fixes pushed as `d0bb7adf`. Fast tiers green on `ci,survival,markov`: lib categorical 11, categorical_smoke 28, markov_smoke 29, tte_smoke 86 — 0 failed; rustfmt clean.

PR #903, opened 2026-07-20
fix(estimation): honest SAEM MH mixing metric + FREM/iiv_on_ruv safeguards [closes #895]

### LINKED ISSUE #895: SAEM: 0% MH acceptance and omega/sigma explosion in IIV-on-RUV FREM models
## Summary

SAEM estimation on a FREM model with `iiv_on_ruv` produces 0% Metropolis-Hastings acceptance across all iterations, causing `omega_RUV` to explode to ~49 (vs true value 0.28) and `sigma` to hit an internal ceiling at e^5 = 148.4 (SD scale). The same model estimates correctly with IMP and IMPMAP. Fixing sigma at the known value recovers the correct `omega_RUV`, confirming the root cause is an uncontrolled omega-sigma tradeoff in the SAEM M-step when the sampler isn't mixing.

## Environment

- ferx 0.2.0
- R 4.6.0 (macOS, Apple Silicon)

## Reproducer

Model: FREM with BLOCK(11) for PK + 8 covariates, plus separate `omega ETA_RUV` and `iiv_on_ruv = ETA_RUV`. 12 total ETAs. Full omega diagonal ranges from 0.046 to 622 (13,500x range).

```r
library(ferx)

# SAEM on IIV-on-RUV FREM — omega_RUV explodes
fit <- ferx_fit(
  model    = "models/frem_iivruv.ferx",
  data     = "models/frem_iivruv_data.csv",
  method   = "saem",
  settings = list(
    n_exploration   = 1000L,
    n_convergence   = 1000L,
    n_mh_steps      = 50L,
    omega_burnin    = 20L,
    seed            = 12345L,
    optimizer_trace = TRUE
  ),
  covariance = FALSE
)

# omega_RUV = 49.14 (should be ~0.28)
# sigma^2 = 22026  (should be ~0.03)
diag(fit$omega)[1]   # ETA_RUV
fit$sigma[1]^2        # SIGMA_RUV

# Trace shows 0% MH acceptance at every iteration
tr <- ferx_trace(fit)
all(tr$mh_accept_rate == 0)  # TRUE
```

### 1. MH acceptance rate is 0% for all 2000 iterations

The Metropolis-Hastings sampler never accepts a single proposal. This was confirmed via `optimizer_trace`:

```
iter   mh_accept_rate
1      0
10     0
100    0
500    0
1000   0
2000   0
```

With 0% acceptance, ETAs are never updated from their initial values during the E-step. The M-step then operates on stale/degenerate sufficient statistics, and the omega-sigma non-identifiability drives both parameters to their respective ceilings.

### 2. Explosion timeline (from trace)

omega_RUV and sigma are stable through the burnin period, then diverge during exploration:

```
iter   omega_RUV   sigma (SD)   phase
1      0.2978      0.1843       explore    (= initial values, within burnin)
20     0.2978      0.1991       explore    (burnin ends here)
50     0.3400      0.2606       explore    (drift begins)
100    0.5675      0.4009       explore
200    1.9370      0.7497       explore    (7x target)
500    49.1787     148.4132     explore    (at ceiling: ~49 and e^5)
1000   49.7678     148.4132     explore
1500   49.1558     148.4131     converge   (locked in)
2000   49.1367     148.4132     converge
```

Sigma saturates at exactly e^5 = 148.413 (SD), i.e. sigma^2 = e^10 = 22026. This appears to be an internal bound.

Meanwhile, PK parameters (TVCL, TVV, TVMAT) and PK omega diagonals are approximately correct, suggesting they are less sensitive to the broken sampling.

### 3. Fixing sigma recovers omega_RUV

When sigma is fixed at the known NONMEM value, the non-identifiability is broken and omega_RUV converges correctly:

```r
# Model modified: sigma SIGMA_RUV ~ 0.1738 (sd) FIX
# Result: omega_RUV = 0.2827 (target: 0.279)
```

This confirms the root cause is the omega_RUV x sigma tradeoff, not a problem with the SAEM algorithm itself.

### 4. Tuning knobs have no effect

All of the following produce identical results (omega_RUV ~ 49, sigma ~ 148):

| Setting                        | omega_RUV | sigma^2  |
|--------------------------------|-----------|----------|
| Baseline                       | 49.14     | 22026    |
| omega_burnin = 100             | 48.76     | 22026    |
| omega_burnin = 200             | 48.42     | 22026    |
| adapt_interval = 25            | 48.76     | 22026    |
| n_mh_steps = 75                | 49.06     | 22026    |
| n_convergence = 1500           | 49.06     | 22026    |
| n_leapfrog = 5 (HMC)          | 49.14     | 22026    |
| omega_RUV init = 0.01          | 49.19     | 22026    |
| **sigma FIXED at NM value**    | **0.28**  | **0.03** |

The HMC result is byte-identical to baseline MH, suggesting `n_leapfrog` may be silently ignored when `iiv_on_ruv` is active.

## Root cause analysis

Two separate issues compound here:

### Issue A: 0% MH acceptance in high-dimensional FREM

The 12-dimensional ETA space with extreme scale differences (omega diagonal: 0.046 to 622) makes MH proposals ineffective. With a single joint proposal covariance, the sampler can't simultaneously propose reasonable moves for ETAs spanning 4 orders of magnitude.

This likely affects all FREM SAEM runs (not just iiv_on_ruv models), though the consequences are less visible when there's no weakly-identified parameter to exploit the broken sampling.

### Issue B: Omega-sigma non-identifiability in M-step

The IIV-on-RUV parameterization creates a ridge: `residual variance ~ sigma^2 * exp(2 * ETA_RUV)`, so the marginal residual is a function of `sigma^2 * exp(omega_RUV)`. When the E-step isn't mixing (0% acceptance), the M-step sufficient statistics are biased, and there's nothing preventing omega_RUV and sigma from walking along the ridge in opposite directions.

NONMEM's SAEM handles this case correctly (omega_RUV converges to 0.28 with free sigma). The difference may be in NONMEM's use of block-specific proposal covariances, simulated annealing, or constrained M-step updates.

## Suggested improvements

In rough priority order:

1. **Fix MH proposal scaling for FREM-scale omegas.** Block-diagonal proposals (separate proposals per omega block) or adaptive scaling per ETA dimension would address the fundamental 0% acceptance issue. This would benefit all FREM SAEM runs, not just IIV-on-RUV.

2. **Add omega upper bounds.** Syntax like `omega ETA_RUV ~ 0.30 (0.0, 5.0)` with explicit lower/upper bounds would let users constrain weakly-identified omega elements. This is a general-purpose safeguard.

3. **Regularize the sigma M-step** for IIV-on-RUV models. A penalty or constraint preventing sigma from growing beyond some multiple of its initial value during SAEM would break the positive feedback loop.

4. **Warn when MH acceptance is persistently 0%.** A runtime warning after the burnin period (e.g., "MH acceptance rate is <1% after 100 iterations — SAEM results may be unreliable") would help users diagnose this without needing `optimizer_trace`.

## Reprex files

The full reprex (model files, data, R scripts, trace data, and diagnostic figures) is available on request. Key files:

- `models/frem_iivruv.ferx` — the FREM model with iiv_on_ruv
- `models/frem_iivruv_data.csv` — FREM-formatted dataset (475 subjects)
- `saem_diagnostics.R` — diagnostic script (trace extraction, all experiments)
- `saem_tuning.R` — tuning knob sweep (4 configs)

roninsightrx:
## Why

Issue #895 reported SAEM on a FREM model with `iiv_on_ruv` showing 0% Metropolis-Hastings acceptance across all iterations, with `omega_RUV` exploding to ~49 (true ~0.28) and `sigma` pinned at its e⁵ ≈ 148 SD ceiling; the same model estimated correctly with IMP/IMPMAP, and fixing sigma recovered `omega_RUV`.

Reproducing on current `main` (a faithful warfarin FREM + `iiv_on_ruv` fit) shows **the explosion no longer happens** — the componentwise MH kernel added after the reporter's 0.2.0 already mixes these models. But the reporter's diagnostic was genuinely misleading and two latent hazards remained:

- The block kernel is **exactly 0%** for FREM-scale Ω (the near-deterministic covariate ETAs — pinned by their pseudo-observations to a posterior SD of ≈√EPSCOV ≪ √Ω_jj — reject every joint proposal). The trace's `mh_accept_rate` reported *only the block kernel*, so it read 0% even though the chain mixed fine via the componentwise sweep. That block-only 0% is exactly what the reporter observed.
- The `iiv_on_ruv` residual variance is `σ²·exp(2·η_RUV)`, so σ and ω_RUV are only jointly identified along a ridge. A genuinely poorly-mixing run can ride σ up to its e⁵ ceiling.

## What changed

All in `src/estimation/saem.rs`:

1. **Combined acceptance reporting** — `mh_accept_rate` (optimizer trace + verbose banner) is now the block + componentwise rate, the honest E-step mixing diagnostic.
2. **FREM-aware block proposal** — each FREM covariate coordinate is damped in the joint block move by `min(1, √EPSCOV/√Ω_jj)`, so block acceptance recovers from 0% and the joint move again explores the correlated PK block. Non-FREM models multiply by exactly `1.0` → byte-for-byte unchanged.
3. **Mixing warning** — when the combined post-burn-in acceptance stays below 1%, SAEM appends a `FitResult.warnings` entry (sampled ETAs never moved ⇒ Ω/σ unreliable).
4. **σ × ω_RUV ridge cap** — for `iiv_on_ruv` models, each free RUV-scaled residual σ's growth is capped (~20× its starting SD) as a **post-M-step clamp**, not an NLopt bound, so a well-posed fit whose σ stays below the cap reproduces the un-capped trajectory bit-for-bit. A run that ends pinned against the cap is warned as weakly identified. FREM EPSCOV (always FIX) and any FIXed σ are skipped.

## Alternatives considered

- **Omega bounds DSL** (issue suggestion 2) — deferred; large cross-optimizer surface.
- **Deeper optimizer-level mixing rewrite** (suggestion 1) — the componentwise kernel already carries correctness; a single-lever proposal-scaling change risks regressing well-tuned fits. The scoped FREM block damping is the targeted win.
- **σ cap via NLopt bound** — rejected: changing the bound perturbs NLopt's search path even when the optimum is interior (~0.3 OFV drift observed on a healthy fit). The post-M-step clamp is a true no-op below the cap.

## Cross-repo dependency

No public Rust API or `FitResult` field change.

## Breaking changes
- [x] None

## Numerical validation
- [x] Compared against prior ferx commit (current `main`)
- [x] Reference values:

Non-FREM `iiv_on_ruv` model (wide-Ω, `--simulate`), before vs after — **identical** (×1.0 multiplier, σ cap non-binding):

```
# before / reference (main)
OFV 2153.4319   ETA_RUV 0.292958   PROP_ERR 0.165132
# after
OFV 2153.4319   ETA_RUV 0.292958   PROP_ERR 0.165132
```

FREM + `iiv_on_ruv` (warfarin, SAEM): stays bounded (omega_RUV O(1), σ ≪ ceiling), no mixing warning; trace `mh_accept_rate` now ~5–20% (was 0.0000, block-only).

## Performance
- [x] No significant impact expected.

## Tests
- [x] Unit tests added (`cargo test --lib` passes — 2636 tests): `compute_ruv_sigma_caps`, `saem_mixing_warning`, `mh_steps` block scaling (`None` ≡ all-ones; zero scale freezes that coordinate).
- [x] Regression test added: Tier-3 slow `frem_iiv_on_ruv_saem_stays_bounded`.

- [x] Not applicable — surfaces through existing trace column + warnings.

## Docs
- [x] `docs/estimation/saem.qmd` updated
- [x] `CHANGELOG.md` `[Unreleased]` → Fixed entry added
- [x] No new `_quarto.yml` pages

## Example execution
- [x] No example execution step needed.

## Checklist
- [x] `cargo clippy` clean on new lines
- [x] No `Dual2`/`PkNum` sensitivity path touched
- [x] No `Cargo.toml` version bump

## Reviewer hints

- The σ cap is a **post-M-step clamp** (`ruv_sigma_caps` + clamp loop), not an NLopt bound — that keeps healthy fits bit-identical.
- `blk_eta_scale` is `None` for non-FREM models and `×1.0` for ordinary ETAs, so nothing off the FREM path changes.
- Combined `mh_accept_rate` is computed per-iteration from `iter_acc`/`iter_prop`, independent of the adaptation-reset counters.

## Open questions

- σ growth cap is e³ (~20×). Open to a different multiple if reviewers prefer.


ronkeizer:
### Update: validated against the real FREM reprex + added the ω_RUV backstop

After the initial commit I ran the **actual** #895 reprex (the 475-subject FREM workshop model, not just the synthetic warfarin FREM), which **does** still misbehave on current `main`. Full bisection:

- Likelihood machinery is correct (ω_RUV fixed ≈ 0 → OFV matches plain FREM / NONMEM).
- With both σ and ω_RUV free, the SAEM M-step walks them **up together** in an unstable feedback loop (ω_RUV 0.30 → 9.5); FIXing σ **or** ω_RUV recovers ω_RUV = 0.29 (NONMEM 0.279).
- FOCEI/IMP fail too, so the joint identifiability is not SAEM-specific.

New commit (`640e8dd7`) adds the **ω_RUV growth cap** (symmetric to the σ cap, correlation-preserving so a block Ω stays PD) — bounding ω_RUV at ~6 instead of ~49 on the reprex — and makes both cap-binding warnings recommend **FIXing σ** (the validated recovery path). Both caps remain pure no-op backstops on a well-posed fit (the simple `iiv_on_ruv` model is reproduced bit-for-bit).

This PR is therefore the **diagnostics + damage-control + guidance** scope. The full cure (stable joint σ/ω_RUV SAEM updating; the post-fit FOCE-OFV fragility; the FREM generator emitting ETA_RUV as a separate diagonal Ω) is tracked in #904.

ronkeizer:
### Root cause found and fixed — this is now a real cure, not just damage control (`e375e837`)

Instrumenting the reprex per-iteration showed the actual mechanism: **`mean(η_RUV)` drifts** (to −0.87 over 130 iters). `Y = f + EPS·exp(η_RUV)` has no typical-value θ, so — unlike mu-referenced structural etas, whose mean folds into their TVP each M-step — η_RUV's mean is never absorbed. Since `ω_RUV = mean(η_RUV²)`, that drift injects a spurious `mean²` (>½ of the inflated ω_RUV) and pumps the runaway.

**Fix:** re-center η_RUV to zero mean each SAEM iteration, absorbing the shift into σ (`η_RUV −= mean; σ ×= exp(mean)`). Residual variance is exactly unchanged (likelihood untouched); `E[η_RUV]=0` is restored so the Ω M-step measures the true variance. σ plays the residual-scale typical-value role that the TVP plays for structural etas.

Reprex, both starts now converge (NONMEM: ω_RUV 0.279, σ 0.177, OFV 6852):

| σ start | ω_RUV | σ | OFV |
|---|---|---|---|
| 0.032 (30× low) | 0.264 | 0.209 | 7706 |
| 0.177 (correct) | 0.280 | 0.203 | 7734 |

Previously ω_RUV ran to ~9–49 with OFV 4e8+. The σ/ω_RUV growth caps from the prior commit remain as no-op backstops. The slow FREM regression is tightened to assert ω_RUV and σ stay < 1.0 with no instability warning. Remaining FOCEI/IMP-standalone divergence + the ferx-r FREM generator (ETA_RUV in-block vs NONMEM's separate diagonal) are tracked in #904.

src/estimation/saem.rs
```diff
@@ -1115,6 +1154,149 @@ fn combined_additive_sigma_at_floor(model: &CompiledModel, params: &ModelParamet
...
+    if n == 0 || !absorb_sigma.iter().any(|&a| a) {
+        return 0.0;
+    }
+    let mean = etas.iter().map(|e| e[kr]).sum::<f64>() / n as f64;
+    for e in etas.iter_mut() {
+        e[kr] -= mean;
+    }
+    for k_s in 0..log_sigma.len() {
+        if absorb_sigma.get(k_s).copied().unwrap_or(false) {
+            log_sigma[k_s] += mean;
```
ronkeizer, line 1221: **Correctness: re-centering breaks its stated residual-variance invariance for a combined error model with a FIXed component.**

`ruv_scale = exp(2·η_RUV)` multiplies the *whole* residual variance `R = σ_add² + (σ_prop·f)²` (`likelihood.rs:641`). Re-centering shifts `η_RUV -= mean` (scaling all of `R` by `exp(-2·mean)`) but only scales the *absorbing* (free, non-FREM) sigmas by `exp(mean)`. When one combined component is FIXed (e.g. additive fixed, proportional free), only the free one is scaled, so the invariance is violated.

Concrete: add=0.1 FIXed, prop=0.2 free, mean=0.3 → the additive contribution goes `0.1² → 0.1²·e^{-0.6}` (−45%) every iteration. The docstring/docs claim "each subject's residual variance is exactly unchanged" — false here. This silently perturbs the objective and biases the σ ↔ ω_RUV split, the very thing #903/#904 stabilize. No parser/validation guard blocks `iiv_on_ruv` + combined + one FIX. The Tier-1 test only exercises `absorb_sigma = [true]`, so it misses this.

Fix: skip re-centering (and the σ cap) when any RUV-scaled sigma component is FIXed/partially-fixed, or validate-reject that config.

src/estimation/saem.rs
```diff
@@ -1115,6 +1154,149 @@ fn combined_additive_sigma_at_floor(model: &CompiledModel, params: &ModelParamet
...
+    if !has_ruv_eta {
+        return vec![None; n];
+    }
+    (0..n)
+        .map(|i| {
+            if sigma_fixed.get(i).copied().unwrap_or(false) || Some(i) == frem_cov_sigma {
+                return None;
+            }
+            let upper = log_sigma_upper.get(i).copied().unwrap_or(f64::INFINITY);
+            Some((log_sigma_init[i] + SAEM_RUV_SIGMA_LN_GROWTH).min(upper))
```
ronkeizer, line 1185: **Robustness: caps anchored to the *initial* guess clamp a well-posed fit and emit a false warning.**

`cap = min(log σ₀ + 3, upper)` ≈ 20× the *init*, and the clamp at `:2328` is unconditional. If a user sets a rough low init (e.g. init 0.01, true 0.3), the M-step σ is clamped at `0.01·e³ ≈ 0.2 < 0.3` on a perfectly well-posed run — corrupting the estimate and firing the "ridge weakly identified / not trustworthy" warning even though the ridge is fine; only the starting guess was low. Same for the ω_RUV cap. There is no lower guard against a low starting value.

src/estimation/saem.rs
```diff
@@ -1889,6 +2176,13 @@ pub fn run_saem(
                 SAEM_OMEGA_DIAG_FLOOR,
             );
 
+            // #895: cap the iiv_on_ruv Ω variance so a runaway ridge can't inflate
+            // ω_RUV without bound (the original report saw ~49). No-op for a
+            // well-posed fit; a correlation-preserving rescale keeps Ω PD.
+            if let (Some(k), Some(log_cap)) = (model.residual_error_eta, ruv_omega_cap) {
+                apply_ruv_omega_cap(&mut state.omega_mat, k, log_cap);
```
ronkeizer, line 2183: **Niche correctness: this rescales a FIXed off-diagonal covariance that was just restored.**

FIXed rows/cols are restored from the template at `:2156-2165`, then this rescales *all* RUV off-diagonals by `s = √(v_cap/v_old)`, including `Ω[k,j]` where `j` is a FIXed eta. If the RUV eta sits in a `block_omega` with a FIXed partner, the user-declared fixed covariance silently changes whenever the cap binds — violating "FIXed entries left exactly as declared." Only fires when the cap binds (rare post-fix), so niche, but worth guarding the off-diagonal loop to skip FIXed `j`.

src/estimation/saem.rs
```diff
@@ -2220,6 +2527,61 @@ pub fn run_saem(
...
+    // #895: warn when a free RUV-scaled residual σ ended pinned against the
+    // iiv_on_ruv growth cap. That signals the σ × ω_RUV ridge is poorly
+    // identified from the data alone; the cap kept σ bounded but the split
+    // between σ and ω_RUV should not be trusted — fix one of them (e.g. σ at a
+    // known value) or drop the IIV on the residual error.
+    for (i, cap) in ruv_sigma_caps.iter().enumerate() {
+        if let Some(cap) = cap {
+            if final_params.sigma.values[i].max(1e-300).ln() >= cap - 1e-6 {
+                warnings.push(format!(
+                    "SAEM residual sigma '{}' hit its iiv_on_ruv growth cap (≈{:.4}); the \
```
ronkeizer, line 2548: **Cosmetic: this warning misattributes the user's own upper bound to the iiv_on_ruv cap.**

When `log_sigma_upper < log σ₀ + 3`, `cap == upper`. A σ that legitimately converges to its user-set upper bound then trips `final σ.ln() >= cap - 1e-6` and emits "hit its iiv_on_ruv growth cap … estimates are not trustworthy", blaming the RUV safeguard for the user's own bound. No numeric harm, but the message is misleading — distinguish `cap == upper` from the growth-driven cap.

src/estimation/saem.rs
roninsightrx: Fixed in #982: added a `ruv_recenter_allowed` gate — re-centring now runs only when every non-FREM residual σ is free, else it's skipped and the growth caps back it up. Also decoupled the absorbing-σ set from the growth-cap set. Regression test `ruv_recenter_allowed_only_when_all_residual_sigma_free`.

src/estimation/saem.rs
roninsightrx: Fixed in #982: both caps are now re-anchored to the data-informed σ/ω_RUV reached by end of exploration (`reanchor_ruv_sigma_caps`/`reanchor_ruv_omega_cap`, loosen-only), so a too-low init no longer spuriously clamps or false-warns a well-posed fit.

src/estimation/saem.rs
roninsightrx: Fixed in #982: `apply_ruv_omega_cap` now takes `omega_fixed` and skips FIXed off-diagonal partners, so a user-declared fixed covariance is left exactly as restored. Test `apply_ruv_omega_cap_leaves_fixed_offdiagonal_untouched`.

src/estimation/saem.rs
roninsightrx: Fixed in #982: `compute_ruv_sigma_caps` carries no growth cap when the user's own upper bound is already tighter (NLopt enforces it), so a σ at its own bound is no longer mis-flagged. Test `ruv_sigma_caps_defer_to_tighter_user_upper_bound`.

PR #906, opened 2026-07-21
fix(estimation): guarded multi-start inner EBE for weakly-identified etas [closes #891]

### LINKED ISSUE #891: Inner EBE loop converges to a sub-optimal local mode on multimodal individual posteriors (weakly-identified random effects)

roninsightrx:
## Why

The inner EBE loop converges to a **sub-optimal local mode** when a subject's individual posterior is **multimodal** — which happens for weakly-identified random effects on nonlinear readouts (#891). The returned η̂ is not the global posterior mode, biasing that subject's FOCEI objective and making EBE diagnostics (η̂, IPRED, IWRES, shrinkage) unreliable, silently.

A guarded multi-start inner EBE (`inner_restarts`, default 1) already existed, but two gaps left #891 unaddressed:

1. **Trigger too narrow.** It re-seeded only subjects with system resets or time-varying covariates. The weakly-identified case #891 targets — a poorly-identified `V1` (~48 % shrinkage) in a saturable-clearance fluconazole model — has neither, so those subjects were never scanned.
2. **Never fired on the eval-only path.** The multi-start is gated on `eta_init.is_none()`, but the `maxiter = 0` (NONMEM `MAXEVAL=0`) evaluation — the exact scenario #891's evidence is drawn from — passed `Some(zeros)`, which the gate does not recognise as a cold start. In practice the multi-start fired only at the final diagnostic EBE pass of a real fit.

## What changed

- **`estimation/inner_optimizer.rs`** — new `weakly_identified_coords`: a two-point central second difference of the inner objective per coordinate at the converged mode. Because `obj` carries the prior term `½·ηᵀΩ⁻¹η`, that difference estimates the posterior curvature `H_ii = data_info_i + (Ω⁻¹)_ii`; a coordinate is flagged when the data adds less curvature than the prior already carries (`H_ii < 2·(Ω⁻¹)_ii`, ≈ conditional shrinkage ≳ 0.3). Reset/TV subjects still scan every random effect (unchanged); every other subject scans only its flagged coordinates. Fixed effects are excluded via a variance floor (`from_matrix` regularises a zero variance to a 1e-8 eigenvalue, so an exact-zero test never survives).
- **`estimation/outer_optimizer.rs`** — the eval-only path (`evaluate_at_initial_params`) now passes `None` instead of `Some(zeros)`. Both seed the inner search at η = 0, but `None` is the honest cold start that enables the guarded multi-start. An eval-only run has no warm EBE history, so this is also semantically correct.

The multi-start fires only at genuine cold starts (eval-only + the final diagnostic pass), never per outer evaluation, so the **fit trajectory and runtime are unchanged**. Well-identified subjects pay only the `2·n_eta` probe evaluations and stay **bit-identical** to `inner_restarts = 0`.

## Alternatives considered

- *Scan all subjects on every cold start* — simplest, but pays a full multi-start on well-identified subjects. The curvature probe restricts the expensive seed solves to coordinates where a missed mode is actually plausible.
- *Use per-subject shrinkage as the trigger* — shrinkage is a post-fit population aggregate, unavailable at solve time (chicken/egg). The per-coordinate curvature is the solve-time analogue.

## Breaking changes
- [x] None

## Numerical validation

`ferx-testdata/fluconazole_radboudumc` (`fluconfinal.ferx`, FOCEI, `block_sigma`, 31 subjects, `maxiter = 0`):

```
inner_restarts = 0 : init OFV 1184.8066   (single-start baseline)
inner_restarts = 1 : init OFV 1122.5965   <- recovers global modes
inner_restarts = 2 : init OFV 1122.5965
```

Exactly reproduces the issue's proof-of-concept (1184.8 → 1122.6). Subjects 14 & 22 recover NONMEM's `V1 ≈ −2` modes; a `±2·sd` seed (`restarts = 1`) already suffices. NONMEM `METHOD=1 INTER` anchors this init evaluation.

- [x] Compared against: NONMEM (per-subject decomposition, issue #891) and the reverted prototype in the issue
- [x] Reference values above

## Performance
- [x] No significant impact expected — cold-start only; well-identified subjects pay only a `2·n_eta` objective probe; the outer optimisation keeps its warm EBEs, so the fit trajectory is unchanged.

## Tests
- [x] Unit tests added (`cargo test --lib` passes — 2641 pass, 0 fail):
  - `weakly_identified_coords_flags_flat_not_sharp`
  - `weakly_identified_coords_skips_fixed_and_nonfinite`
  - `inner_restarts_bit_identical_on_wellidentified_subject` (no-regression)
- The fluconazole end-to-end anchor lives in the PR (its data is in the sibling `ferx-testdata`, not the repo), per repo convention for external-data validation.

## Docs
- [x] `docs/model-file/fit-options.qmd` — `inner_restarts` semantics updated for the weakly-identified trigger
- [x] `CHANGELOG.md` `[Unreleased] → Fixed` entry added
- [x] `FitOptions::inner_restarts` doc-comment updated

## Reviewer hints

- The curvature threshold (`H_ii < 2·(Ω⁻¹)_ii`) and the fixed-effect variance floor (`1e-7`, above the `from_matrix` `1e-8` regularisation) are the two tunables — see `weakly_identified_coords`.
- The `outer_optimizer.rs` one-liner (`Some(&cold_etas)` → `None`) is what makes the eval-only path actually exercise the multi-start; the existing `outer_maxiter_zero_is_eval_only_and_optimizer_independent` test still passes.

## Open questions

- Scope was **trigger-only** by request. #891's third bullet — a **diagnostic warning** when the multi-start finds a materially lower mode — is deferred to a follow-up (would plumb a `multistart_gain` field through `EbeResult` → `InnerLoopStats` → `FitResult.warnings`).


roninsightrx:
Thanks @copilot — both addressed in 8d542863.

**1. `weakly_identified_coords` doc-comment (inner_optimizer.rs).** You're right that the doc conflated two distinct degeneracy paths. Clarified: a non-finite objective *at the mode* (`nll`) disables the probe entirely and returns all-`false` (no coordinate scanned), whereas a non-finite/non-positive *per-coordinate curvature* flags that one coordinate as weakly identified so a pathological direction is scanned. The code was already correct; only the comment was ambiguous.

**2. "bit-identical" no-regression test.** Tightened to exact `==` on both `eta` and `nll` (was `abs() < 1e-9` / `1e-12` slack). Bitwise identity is in fact the real contract, not just closeness: for a unimodal subject no alternate seed beats the improvement guard (`cand_nll + 1e-9 < nll`), so `eta`/`nll` are never reassigned and the values returned are the very same f64s the base solve produced; the two `find_ebe` calls are deterministic (pure objective, no RNG). Test passes with strict equality.

Issue #912, opened 2026-07-21
refactor(types): make DoseEvent::cmt private to compiler-enforce the CMT=0→1 convention (accessor landed in #913)


TeunP:
## Problem

`CMT` is 1-based and `CMT=0` is NONMEM's *default dose compartment*, i.e. compartment 1. After #899 every engine agrees on that — but the agreement is held together by **roughly a dozen independent `saturating_sub(1)` call sites**, not by construction. Nothing stops the next dose-application site from getting it wrong, and the default failure is silent.

This is a maintainability issue, not a live bug. Filed because the evidence that it recurs is concrete.

## Evidence that it recurs

Before #899 the codebase carried **three different spellings** of the same conversion, which is how `CMT=0` ended up meaning four different things depending on which driver a subject took:

- `let cmt_idx = dose.cmt - 1;` — underflowed on `CMT=0` (debug panic / release silent drop)
- `let cmt_idx = dose.cmt.saturating_sub(1);` — correct
- `if dose.cmt > 0 { let cmt = dose.cmt - 1; … }` / `if d.cmt >= 1 { … }` — skipped the dose

During #899 itself I fixed the sites a grep surfaced, and **still missed three** in `src/sens/ode_provider.rs` (`:5139`, `:5289`, `:5755`). They were caught only because a `Dual2`-vs-FD parity test happened to cover one of them — it failed with `left = 0.0, right = 7.7975846378641815`. The other two are route-onset and saltation jets: they would not have zeroed a prediction, they would have silently dropped a **gradient term** while the value still looked right.

A fourth site (`DoseAttrMap::indexed_slot`) was found later still, by review rather than by test: `CMT=0` missed the compartment-indexed lookup and fell back to the bare `F` slot, so a model declaring `F1 = 0.6` delivered the dose at `F = 1.0`.

Four misses in one focused pass on a convention the author was actively thinking about is the argument.

## Why it is easy to get wrong

Both wrong forms read as reasonable at the call site:

```rust
let cmt_idx = d.cmt - 1;        // looks like ordinary 1-based → 0-based
if d.cmt >= 1 { /* … */ }       // looks like a sensible bounds guard
```

Neither is obviously incorrect without knowing that `CMT=0` is meaningful rather than malformed. Two review comments in `sens/` had in fact entrenched the opposite reading (#642 review #1, #449 review #8 — the latter citing a datareader guarantee that does not exist), so the wrong mental model was written into the code as justification.

## Suggested direction

[...]

Issue #910, opened 2026-07-21
fix(pk): a dose CMT valid on a transit/IG model is out of range on its ODE twin and is silently dropped


TeunP:
## Problem

The absorption ODE twin does not preserve the analytical model's `CMT` numbering, so a dose compartment that is **valid on a `transit`/`igd` model** can be **out of range on the twin that actually solves it** — where it is silently dropped by the twin's `if cmt_idx < n { … }` bounds guard.

This is the same failure class as #899 and #375 (a dose that names a compartment the engine cannot deliver into is ignored rather than rejected), reached by a different mechanism: not an out-of-range `CMT` in the dataset, but a `CMT` that changes meaning when the subject is rerouted.

> **Status: found by code trace while implementing #899, not reproduced by a run.** The reasoning below is cited to file:line, but the first step of any fix should be a failing test that pins the actual numbers. It is filed rather than fixed because #899's validation cannot see the twin (see "Why #899 does not catch it").

## The numbering shifts

The twin's source is reconstructed at `src/parser/model_parser.rs:7784-7791`, and `n_states = states.len()` at `src/parser/model_parser.rs:5028`:

- 1-cpt absorption model → `ode(obs_cmt=central, states=[central])` → **`n_states = 1`**
- 2-cpt absorption model → `ode(obs_cmt=central, states=[central, periph])` → **`n_states = 2`**

The **depot is dropped**: absorption becomes the forcing term `transit(n, mtt)` / `igd(mat, cv2)` (`src/parser/model_parser.rs:7719-7723`), not a state.

Against the analytical topologies:

| model | analytical | twin |
|---|---|---|
| `one_cpt_transit`, `one_cpt_ig` | `n_states: 2`, `["depot", "central"]` (`src/pk/topology.rs:126-131`, `:137-143`) | `n_states: 1`, `["central"]` |
| `two_cpt_transit`, `two_cpt_ig` | `n_states: 3`, `["depot", "central", "peripheral"]` (`src/pk/topology.rs:165-171`, `:174-180`) | `n_states: 2`, `["central", "periph"]` |

So every compartment shifts down by one across the reroute:

- **`CMT=1`** (analytical depot) → twin state 0 (`central`), which the absorption forcing already occupies. `input_rate_consumes_cmt` skips the bolus branch for it and the mass arrives as `R_in` over time — **accidentally correct**, which is why this has gone unnoticed.
[...]

PR #642, opened 2026-07-01
feat(ode): close SS + lagtime / SS + rate-under-F / SS + modeled-dose FD gradient cells [#486]

hiddevandebeek:
feat(ode): close SS + lagtime / SS + rate-under-F / SS + modeled-dose FD gradient cells [#486]

## Why
Issue #486 tracks the remaining finite-difference (FD) *gradient* cells in ferx's
analytic-sensitivity coverage. Three steady-state (`SS=1`) ODE dosing sub-cases still
fell back to FD even though SS bolus/infusion dosing itself has been analytic since
#473: an SS dose combined with (a) an estimated lagtime, (b) a rate-defined infusion
under bioavailability `F ≠ 1`, and (d) a modeled-duration/rate dose (`RATE=-1`/`-2`).
This is PR3 of a 5-PR plan (`plans/analytic-gradient-completion/`) to close the
remaining Easy–Medium FD cells; it is fully isolated in `equilibrate_ss_state_g` and
develops/merges independently of the sibling PRs.

## What changed
All three sub-cases are closed by two additive mechanisms in `src/sens/ode_provider.rs`:

1. **(d) modeled-dose × SS + (b) rate-under-F × SS** — `equilibrate_ss_state_g` now
   threads the caller's mode-aware `inf_eff` jet (rate, window) — the same jet the non-SS
   event-driven walk already uses for a modeled or `F`-scaled infusion — into its
   per-cycle active/quiet split, and injects the identical rate-off saltation the main
   walk's `K_INF_END` uses at each cycle's window boundary. This closes both sub-cases
   with one change, since both feed through the same `inf_eff` jet; the
   `has_rate_defined_ss_infusion_under_f` gate helper is now dead code and removed.
2. **(a) SS + estimated lagtime** — a new `K_SS_SEED` timeline event seeds the pre-arrival
   window `[t_dose, t_dose+lag)` from the previous interval's steady-state tail, via a new
   `ss_state_at_phase_g` (the dual mirror of production's `ss_state_at_phase`) advancing
   the trough forward by `phase = II − lag`. A new `extend_flow_by_dual_duration_g`
   helper Taylor-extends a state by a dual-valued (not just event-time) duration —
   needed because `phase`, and the active/quiet split inside it, are durations whose
   *length* (not a forcing) carries `lag`'s jet. The SS dose's own arrival keeps the
   **unmodified** general lagtime saltation (no special-casing needed): the trough's
[...]

PR #449, opened 2026-06-22
feat(sens): analytic gradients for time-varying covariates — ODE walk [#439] + analytical inner [#447]

hiddevandebeek:
## Why

Advances #439 (the ODE analytic-sensitivity Tier-1 sweep): arms **analytic
FOCE/FOCEI gradients for time-varying covariates on ODE models**. A model whose
covariates change over time (per-event `WT`, `CRCL`, …) previously fell back to
the finite-difference gradient on every ODE subject; now the **bolus** TV-cov case
gets the exact analytic outer gradient and the light `Dual1` inner η-gradient.

This is the `(θ,η)`-seeding **restructure** the remaining Tier-1 items need: the
static ODE provider seeds the dual on the *PK parameters* and computes `∂pk/∂(θ,η)`
once — valid only when `pk` is constant along the trajectory. TV-cov makes `pk`
vary per covariate segment, so a single `∂pk/∂η` can't represent it. The fix
mirrors the **analytical** TV-cov path (`run_obs_tvcov`): seed the dual on the
underlying `(θ,η)` (`M = n_theta + n_eta ≤ MAX_ODE_AXES`) and re-evaluate `pk` per
event, so the dual carries `∂state/∂(θ,η)` directly (no chain).

This PR also **closes #447** (the analytical-PK counterpart): the light `Dual1`
inner EBE gradient declined TV-cov / oral-infusion subjects and reverted to FD even
though the analytical *outer* already served them — now fixed by the same pattern
(see "Analytical inner (#447)" below).

> **Stacked on #446** (base `feat/ode-percmt-readout-439`). Rebases onto `main`
> once #438 → #446 land.

## What changed

- **`integrate_tvcov_g<T: PkNum>`** — a bolus event-driven walk over the dual state,
  the ODE analogue of the analytical `event_driven_sens_g`. It reproduces
  production's `ode_predictions_event_driven`: a merged dose+obs timeline (dose
  sorts before a co-timed obs), each segment `[cur_t, t_event]` integrated with the
  params evaluated **at** `t_event` (NONMEM end-of-interval), boluses applied after
  the segment, state captured at each observation. Generic over `T`, so it serves
  both the `Dual2<M>` outer and `Dual1<n_eta>` inner.
- **`seed_pk_dual2<M>` / `seed_pk_dual1<N>`** — per-event flat PK-slot duals seeded
  on `(θ,η)` / `η` at a covariate snapshot (the ODE counterpart of `run_obs_tvcov`'s
[...]

PR #913, opened 2026-07-21
fix(ode): reject an out-of-range dose CMT and unify CMT=0 on the ODE engine [closes #899]

### LINKED ISSUE #899: fix(ode): an out-of-range dose CMT is silently dropped on the ODE engine (analytical side fixed in #375)
## Problem

On the ODE engine, a dose whose `CMT` is past the end of the declared state vector is **silently dropped**. Every bolus-application site is an unguarded `if cmt_idx < n { … }` with no `else`:

| site | context |
|---|---|
| `src/ode/predictions.rs:2192` | `u[cmt_idx] += dose_f_bio[i] * dose.amt;` |
| `src/ode/predictions.rs:3537` | `u[cmt_idx] += dose_f[i] * d.amt;` |
| `src/ode/predictions.rs:4066` | `u[cmt_idx] += ode.dose_attr_map.f_bio(..) * d.amt;` |

Nothing upstream rejects it either: the datareader has no model, and `check_dose_compartments` (added in #375) returns early for ODE models by design — it validates against `PkModel` topology, which an `[odes]` model does not have.

So a typo'd `CMT` in an ODE dataset produces a fit that converges, reports a finite OFV, and quietly ignores some or all of the dosing. No error, no warning.

## Why now

This is the **same failure class** PR #896 removed on the analytical side, where the three paths disagreed about an unroutable dose (one panicked, one mis-routed, one silently dropped it). That PR fixed the analytical engine and explicitly left the ODE engine untouched, because the fix needs a different input — the ODE state count, not `PkModel` topology — and half-doing it would have been worse than naming it.

The docstring on `check_dose_compartments` (`src/api/validation.rs`) currently says so in as many words, so this issue is the tracked half of that statement.

## The mechanism already exists

The adaptive path solved exactly this. `reject_unsupported_dose_compartment` (`src/ode/predictions.rs:2335`) takes the ODE spec and the state count and returns a typed error:

```rust
if cmt > n_states {
    return Err(format!(
        "decision {decision_index}: dose into compartment {cmt} but the model has \
         {n_states} state(s)"
    ));
}
```

Both adaptive dose-materialisation sites call it (`:3021`, `:3064`), so controller-emitted doses are already covered. It is only the **static** dataset path that isn't.

## Suggested direction

Validate up front rather than at the application site, mirroring what #375 did for the analytical engine:

- Extend `check_dose_compartments` to handle ODE models instead of returning early: for `model.ode_spec`, check `dose.cmt` against the declared state count (`ode_spec.n_states` / the `states=[…]` list) rather than `PkModel` topology.
- Reuse the existing `E_DOSE_CMT_OUT_OF_RANGE` code and message shape, naming the declared state names so the error is actionable ("model declares states [depot, central, periph]").
- The three `if cmt_idx < n` guards then become defensive, and should say so — matching how the analytical walk's guards are now written.

Points needing a decision:

1. **`CMT=0`.** On the analytical engine this is NONMEM's default dose compartment and is accepted for a bolus. Confirm what the ODE engine does with it today before deciding whether to keep the same convention.
2. **Input-rate compartments.** `input_rate_consumes_cmt` already excludes a compartment fed by a built-in absorption function from bolus application (deliberately — it receives `R_in` over time instead). That is a *legitimate* skip and must not be turned into an error; only the out-of-range case should be.
3. Whether an out-of-range `CMT` on an **observation** row has the same problem, or is caught elsewhere.

## Validation

A regression test that fails without the fix (an ODE model + a dose into `CMT = n_states + 1`, asserting an error rather than a silently-reduced prediction), plus a positive control per declared compartment so the check isn't over-broad. No NONMEM anchor needed — this rejects invalid input, it does not change any numerical path.

## References

- Named as out of scope in PR #896 / #375 and in the `check_dose_compartments` docstring.
- Existing typed guard for the adaptive path: `src/ode/predictions.rs:2335`.
- Analytical-side equivalent: `check_dose_compartments`, `src/api/validation.rs`.

TeunP:
## Why

On the ODE engine every dose-application site was an unguarded `if cmt_idx < n { … }` with no `else`, and nothing upstream rejected an out-of-range `CMT`: the datareader has no model, and `check_dose_compartments` (#375) returned early for ODE models by design — it validates against `PkModel` topology, which an `[odes]` model does not have.

So a typo'd `CMT` produced a fit that **converged and reported a finite OFV having ignored the dose entirely**. No error, no warning. Confirmed by running it — 2-state model, single bolus `AMT=100`:

```
[CMT=3] preds        = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[CMT=3] no-dose      = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[CMT=3] control CMT1 = [5.6147, 8.6507, 11.1168, 11.7698, 11.1260, 8.5794]
```

`fit()` returned `Ok` with `ofv = 5999999999834.21` and five warnings, none mentioning the dose; the valid control fit at `ofv = 11.68`.

**`CMT=0` was worse than the issue reported.** #899 framed the ODE engine as internally consistent and merely disagreeing with the analytical engine. It disagreed with *itself* — four behaviours, and which one you got depended on the dataset:

| site | `CMT=0` did |
|---|---|
| `ode_predictions_with_extra_breaks_and_stats` (plain datasets) | bare `- 1` → **underflow**: debug panic `attempt to subtract with overflow`; release wrapped to `usize::MAX`, failed the bound, **silently dropped** |
| `ode_predictions_event_driven`, `adaptive_frozen_replay_tv`, `ss_state_at_phase` | `saturating_sub(1)` → compartment 1 |
| `equilibrate_ss_state` (+ its `sens/` twin) | bailed out → *single-dose* curve instead of the steady state |
| `gated_infusions`, `_with_states`, `apply_segment_boundary`, the `sens/` dose walk | skipped the dose |

A plain dataset panicked; add a time-varying covariate and the same dataset silently worked, dosing compartment 1. That dataset-dependence is why it survived.

## What changed

Two halves, mirroring what #375 did analytically.

**1. Validate up front.** `check_dose_compartments` no longer exempts ODE models. The **range** rule now runs for both engines, reading `OdeSpec::n_states` and naming the declared `state_names`; the **infusable-subset** rule stays analytical-only (an ODE dose is delivered by indexing the state vector, so every declared state accepts both a bolus and a zero-order input). The one part of the routing rule that does carry over is `CMT=0` on an **infusion** — the default dose compartment is defined for a bolus but not for a zero-order input, and the analytical engine already rejects it.

**2. Unify `CMT=0` → compartment 1** across all eleven conversion sites, including the `sens/` gradient twins. This matches the analytical engine, NONMEM's `DEFDOSE`, and what three of the ODE drivers already did.

**3. …including compartment-indexed dose attributes.** `DoseAttrMap::indexed_slot` normalises `cmt.max(1)`, so a dose written `CMT=0` reads `F1` / `ALAG1` rather than missing the indexed map. Without it a model declaring `F1 = 0.6` and no bare `F` left `PK_IDX_F` at `PkParams::default()`'s **1.0**, delivering the dose at full amount with no error — the last site where `CMT=0` still meant something other than compartment 1. Analytical models are unaffected: per-compartment `F{cmt}`/`ALAG{cmt}` are ODE-only (`model_parser.rs:1590` rejects an unbound one on a `pk(...)` model outright), so their indexed map holds only `D{cmt}`/`R{cmt}` — and a `CMT=0` modeled-`RATE` dose is an infusion by `DoseEvent::is_infusion`, which this PR rejects on both engines.

Three smaller things found while doing it:

- **`predict_survival` was the only member of the `predict`/`simulate` family without `assert_dose_compartments_supported`.** Added — an unroutable dose otherwise silently changes the exposure a joint PK-TTE hazard reads.
- **The `population_uses_analytical_pk` bail-out had to be scoped to analytical models.** It is sound when the `pk` line is a pure-TTE model's placeholder, but an `[odes]` block is never a placeholder: a hazard reading an ODE state has the system integrated for it even with no PK observation in the dataset. Ungated, a TTE-only + ODE-PK model would have skipped the whole check.
- **Two `sens/` comments asserted a datareader guarantee that does not exist** (#642 review #1, #449 review #8: "a malformed `CMT=0` … the datareader rejects it upstream"). It does not — only a *missing* `CMT` column defaults to 1; an explicitly written `CMT=0` passes through. Corrected in place.

## Alternatives considered

**`CMT=0` = reject, instead of = compartment 1.** Defensible — an `[odes]` block names every compartment explicitly, so arguably there is no default to fall back on. Rejected because three of the ODE production paths already implement compartment 1, and because it would turn a dataset that produces *correct* numbers today (ODE + TV covariate + `CMT=0`) into a hard error. "Silently wrong → correct" is a better change than "quietly correct → rejected". Discussed on the issue.

**`panic!` in the now-defensive guards**, matching `pk/event_driven.rs:719` and `sens/propagate.rs:935`. Planned, then reversed: those analytical functions are `pub(crate)` behind an always-validating `api` layer, whereas the ODE drivers are `pub fn`s re-exported through `pub mod ode` and reachable with no validation at all. They stay bounds checks, with comments saying why.

## Cross-repo dependency

No public Rust API signature changes; `check_dose_compartments` is `pub(crate)`.

## Breaking changes
- [x] None

Behaviour changes for datasets that were previously silently wrong: an ODE dose past the declared states is now an error rather than ignored, and an ODE `CMT=0` bolus now doses compartment 1 rather than nothing (or panicking). An ODE `CMT=0` **infusion** is now rejected, matching the analytical engine since #375.

## Numerical validation
- [x] Compared against: prior ferx commit / the analytical engine

No NONMEM run needed, for two reasons:

1. The out-of-range rejection is invalid-input handling — it changes no numerical path.
2. The `CMT=0` remap *is* a numerical change, but it is anchored transitively rather than from scratch: the analytical engine's `CMT=0` behaviour is already NONMEM-validated (#375), and `tests/analytical_ode_equivalence.rs` pins the ODE engine to the analytical one. The new `bolus_cmt0` and `steady_state_cmt0` cases run that comparison across all six model families.

## Performance

The range rule is O(doses), once per public call, on a path that already walked every dose.

## Tests

| tier | where | what |
|---|---|---|
| 1 | `src/api/tests/dose_compartment_tests.rs` | 5 cases: out-of-range rejected, message names the declared states + `block: odes`, every declared state accepted (positive control), the `CMT=0` bolus/infusion split, the zero-amount split |
| 1 | `src/ode/predictions_tests.rs` | `ode_with_states_ignores_out_of_range_cmt` re-pointed at a genuinely out-of-range `CMT` (it previously conflated `CMT=0` with out-of-range), plus a new `CMT=0` = compartment 1 case |
| 1 | `src/sens/ode_provider_tests.rs` | two `Dual2`-vs-FD parity tests (bolus + steady state), each also asserting bit-equality with the `CMT=1` twin |
| 1 | `src/types_tests.rs` | `CMT=0` resolves `F1`/`ALAG1`: pins the concrete failure (`F1 = 0.6` must not deliver at `1.0`), asserts value **and slot** agree with `CMT=1` (so the sens walk differentiates the right parameter), plus a not-over-broad control — a map indexing only compartment 2 still falls through to the bare slot |
| 2 | `tests/dose_compartment_routing.rs` | 6 cases at the public boundary: `fit()` errors, `check_model_data` reports it, `predict()` panics, positive control per state, `CMT=0` predicts identically to `CMT=1`, and the observation-`CMT` confirmation |
| 2 | `tests/analytical_ode_equivalence.rs` | `bolus_cmt0` + `steady_state_cmt0` across all six families |

**The parity test earned its keep.** After my first pass at the `sens/` sites, `ode_provider_cmt_zero_bolus_matches_the_default_compartment_and_fd` failed with `left = 0.0, right = 7.7975846378641815` — the analytic gradient path still never applied the dose. Three more sites in `sens/ode_provider.rs` had the same `d.cmt >= 1` gate. Two of them (`:5139`, `:5289`) are route-onset and saltation jets: they would not have zeroed the value, they would have silently dropped a **gradient term** while the prediction still looked right. Worth noting that the six-family equivalence sweep passed *before* that fix, because production was already correct — only the `Dual2`-vs-FD oracle saw it.

Local runs, all green:

```
cargo +nightly fmt --check                          OK
cargo check --all-targets  (ci / ci,survival / ci,markov)   0 / 0 / 0
cargo test --lib --features ci                      2689 passed, 0 failed
cargo test --lib --features ci,survival             2802 passed, 0 failed, 18 ignored
cargo test --test dose_compartment_routing          16 passed, 0 failed
cargo test --test analytical_ode_equivalence         7 passed, 0 failed
```

## Docs
- [x] `docs/` updated for user-visible changes — new "`CMT` on `[odes]` models" section in `docs/data-format.qmd`

Two `Fixed` entries, plus an amendment to the #375 entry, which recorded the ODE/analytical `SS` + `CMT=0` disagreement as still open. It no longer is.

## Checklist
- [x] `cargo clippy` clean
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `equilibrate_ss_state_g` and the dual dose walk changed, both covered by the new `Dual2`-vs-FD parity tests

## Docs & examples
- [x] No example execution step needed (fix with no user-visible API surface; `docs/` change is prose)

## Reviewer hints

**Focus on `src/sens/ode_provider.rs`.** It carries five of the eleven conversion sites and is where the first pass was wrong. The distinction that matters: sites applying a **bolus** or an **SS trough** now map `CMT=0` → index 0; sites filtering **infusions** keep their `d.cmt >= 1` guard, because a `CMT=0` infusion is rejected up front. If a bolus site still reads `d.cmt >= 1`, that is a bug.

**`src/api/validation.rs`** — the ODE branch takes the range rule and the `CMT=0`-infusion rule but skips the infusable-subset rule. Check that split reads correctly.

Skimmable: `CHANGELOG.md`, `docs/data-format.qmd`, and the comment-only corrections in `src/ode/predictions.rs`.

## Follow-ups filed

1. **#912 — the convention has no single home.** `CMT=0` = compartment 1 is now enforced by ~a dozen independent `saturating_sub(1)` call sites rather than by construction, and both wrong forms (`d.cmt - 1`, `if d.cmt >= 1`) read as reasonable at the call site. The evidence that it recurs is this PR itself: three spellings existed beforehand, I missed three sites during it, and a fourth (`indexed_slot`) surfaced by review rather than by test. #912 proposes a single named `DoseEvent::cmt_idx()` accessor, and explicitly defers the invasive version — making `.cmt` private, the only thing that actually *prevents* the mistake — until it recurs again.

2. **#910 is filed off the back of this**: the absorption ODE twin drops the depot, so its `n_states` is one lower and every `CMT` shifts down — `CMT=2` on a `one_cpt_transit` passes validation and is out of range on the twin that actually solves it. Same silent-drop class, different mechanism, and this PR's validation cannot see it (the primary carries `ode_spec: None` and the twin is built lazily inside `effective_for`). Found by code trace, **not reproduced** — flagged as such in the issue.

Closes #899.


TeunP:
## Adversarial review round — four more CMT=0 gaps fixed

Ran an xhigh multi-agent review of this PR (8 dimension finders → 3 independent skeptics per finding, defaulting to *refute* → completeness critic). It found that the CMT=0 unification was incomplete: the convention was still open-coded across ~a dozen sites and several had drifted. Ten findings survived verification, collapsing to four distinct defects. Each is now fixed **with a mutation-verified regression test** — I reverted every fix and confirmed the matching test fails before restoring it.

| # | site | defect (CMT=0) | mutation proof |
|---|---|---|---|
| **A** | `sens/ode_provider.rs` `equilibrate_ss_input_rate_state_g` | SS-into-absorption **gradient** returned a zero trough while the value path equilibrated → silent value≠gradient FOCEI error. The *fifth* CMT=0 site. | grad `0.51` vs value `10.49` when reverted |
| **B** | `predictions.rs` / `ode_provider.rs` / `types.rs` (`f.cmt + 1 == d.cmt`) | on the **event-driven** driver a CMT=0 dose into a `zero_order` compartment delivered **no mass** — bolus suppressed, no window opened | pred `0.0` vs control `0.634` when reverted |
| **C** | `validation.rs` SS gates | the `E_ABSORPTION_SS_ZERO_ORDER` / `E_ABSORPTION_SS_LAG` rejections keyed on raw `d.cmt`, so CMT=0 **bypassed** them into B's silent drop | `rejected: []` when reverted |
| **D** | `adaptive.rs` | stale "no-op for ODE" comment that #899 makes false | comment |

### The mutation check earned its keep twice

My **first** B test used a plain dataset and **passed even with the bug present** — because the plain driver opens the zero-order window through a different `saturating_sub` path, while the bug lives only on the event-driven driver (reset / TV covariate / IOV). A green-but-vacuous test. Reverting the fix and watching the test *not* fail is what exposed it; I traced the two paths with a throwaway probe, confirmed the drop is event-driven-only, and rewrote the test to force that driver. Worth stating plainly since a passing suite would otherwise have hidden it.

### Sustainability — this is why "issues keep popping up"

The root cause across A/B/C is the one #912 names: `CMT=0 ≡ compartment 1` was enforced by ~a dozen independent `cmt - 1` / `cmt >= 1` / `f.cmt + 1 == cmt` idioms that disagreed. Rather than add four more scattered `.max(1)` sites, I pulled #912's fix forward: the convention now has a single named home — `DoseEvent::cmt_idx()` (0-based) and `cmt_1based()` (1-based, CMT=0→1) — and the fixes route through it. **#912 is resolved by this PR** rather than deferred.

### Findings the review *refuted* (kept for the record)

Four findings were killed by the skeptics, correctly: three "no test for X" observations (a missing test is not a runtime defect — though A/B's missing tests are exactly why those bugs existed, so I added them), and a gradient-parity finding on the zero-order twin that was refuted on value==gradient grounds but whose *value* half (B) was the real bug.

### Verification

All local, green: `cargo fmt --check`, `cargo test --lib --features ci`, `--test dose_compartment_routing`, `--test analytical_ode_equivalence`. New tests: SS-first_order CMT=0 Dual2-vs-FD parity (A), event-driven zero_order CMT=0 value equality (B), SS-zero_order CMT=0 rejection (C), and a `cmt_idx`/`cmt_1based` accessor unit test.

Neither #915 (analytical SS exact-solve) nor #900 (binary endpoint) touches any CMT=0 site — checked. #915 shares two files with this PR in the SS-absorption region, so whichever lands second gets a semantic re-check on the merge build; no textual conflict on the functions changed here.

TeunP:
## Code review (xhigh, adversarial) — reject out-of-range dose CMT + unify CMT=0

Traced all twelve `.cmt` conversion sites, the `sens/` gradient twins, and the `check_dose_compartments` restructure line-by-line — no happy-path glancing. The core work is **correct and well-tested**. One real gap surfaced, now fixed on this branch (`42a60140`).

### Findings

| # | Sev | Location | Issue |
|---|-----|----------|-------|
| B1 | **Medium** | `api/validation.rs:1036` | `check_absorption_closed_form_support` rejected `CMT=0` on a closed-form transit/IG model as a *"non-depot compartment"* — but `CMT=0` **is** the depot (compartment 1). Directly contradicted this PR's *"CMT=0 = compartment 1 everywhere / last site"* claim. **Fixed + tested (`42a60140`).** |
| O1 | note | `types.rs` / `CHANGELOG.md` | *"the last site where CMT=0 meant something other than compartment 1"* / *"every site resolves CMT=0 to compartment 1"* were overstated — B1 was another site, and the adaptive controller deliberately treats `CMT=0` as an *error*. Corrected in the B1 commit. |
| O2 | note | `sens/ode_provider.rs:5143/5294` | The route-onset & saltation jet fixes (`d.cmt>=1` → `saturating_sub`) are correct by inspection but have no *targeted* test exercising a **lagged / modeled-rate `CMT=0` event-driven** dose — the added parity tests use plain bolus + SS. A coverage nicety, not a defect. |

### B1 — confirmed, fixed, verified

Reproduced on `one_cpt_transit`: `fit()` CMT=1 → `Ok`, CMT=2 → `Err` (correct), **CMT=0 → `Err`** / `predict()` **panics** with *"does not support dosing into a non-depot compartment (CMT=0)"*. The check keyed on the raw `dose.cmt != 1`; the data reader passes an explicit `CMT=0` through unchanged (only a *missing* CMT column defaults to 1), so it reached this gate and a valid dataset was rejected.

Accepting it is sound because `CMT=0` behaves exactly like `CMT=1` on **both** paths — the closed form folds every dose through absorption regardless of `cmt`, and on the ODE twin `input_rate_consumes_cmt(twin, 0) == input_rate_consumes_cmt(twin, 1)` (both `saturating_sub(1)==0`). Verified bit-equal on each:

```
closed-form transit  CMT=0 preds == CMT=1  [1.0949, 1.4833, 0.9944]  equal
ODE-twin  (TV-cov)   CMT=0 preds == CMT=1  [1.0938, 1.4533, 0.9202]  bit-equal
CMT=2 still rejected on both paths
```

Fix: `dose.cmt != 1` → `dose.cmt_1based() != 1` (routes through this PR's own accessor; `CMT=0→1` accepted, `CMT>=2` still rejected). Plus two mutation-verified regression tests (`transit_…` / `ig_…_cmt_zero_is_the_default_depot_not_rejected`), a CHANGELOG entry, and the O1 corrections.

Verification: lib `2692 passed / 0 failed`; transit `18/0`, ig `49/0`, dose_compartment_routing `7/0`, analytical_ode_equivalence `18/0`; both new tests fail when the guard is reverted (mutation-confirmed); `fmt --check` clean.

### Traced and clean (no defect)

- **`predictions.rs:3058`** adaptive `u[cmt - 1]` — safe: `action.validate()` at :3028 rejects `cmt==0` up front (controller dosing deliberately requires an explicit CMT; a reasonable interface asymmetry vs. data-file `CMT=0`).
- **`predictions.rs:1172`** `gated_infusions` bare `- 1` — after the `cmt==0` early-return; infusion `CMT=0` rejected up front.
- **`predictions.rs:2688`** `m.cmt==0` — a *monitor* (observation) target, not a dose; correctly no default-compartment semantics.
- **`input_rate_consumes_cmt`** normalizes `CMT=0` via `saturating_sub(1)` consistently, so the `CMT=0`-into-`zero_order` fix is coherent across every driver.
- **`equilibrate_ss_input_rate_state_g`** guard `dose.cmt_idx() >= n` ≡ old `cmt>n` minus the wrong `CMT=0` bail.
- **`predict_survival`** guard is a genuine no-op for pure-TTE (early-return via `ode_spec.is_none() && !population_uses_analytical_pk`).
- Range rule / `CMT=0`-infusion rule / dedup-by-`(is_infusion,cmt)` in `check_dose_compartments` all correct.

Follow-ups #910 (transit-twin CMT shift) and #912 (accessor home) are correctly filed; the B1 fix does **not** depend on #910 (verified twin `CMT=0`==`CMT=1` directly).

**Verdict:** solid PR; the CMT=0 unification is now genuinely complete across both engines after `42a60140`.

🤖 Reviewed with an xhigh adversarial pass by [Claude Code](https://claude.com/claude-code)

Issue #914, opened 2026-07-22
perf(ode): solve the ODE bolus/infusion steady state exactly, and warn when it can't


TeunP:
## Problem

`#908` made the **analytical** event-driven walk solve the periodic steady state exactly as `u_ss = (I − M)⁻¹·b`. The **ODE** engine's ordinary bolus/infusion steady state still expands a truncated pulse train, and — like the analytical walk before #908 — does so **silently**.

Two gaps remain, both in `equilibrate_ss_state` (`src/ode/predictions.rs`) and its dual twin `equilibrate_ss_state_g` (`src/sens/ode_provider.rs`):

1. **Truncation.** The bolus/infusion branch loops up to `SS_EQUILIBRATION_CYCLES = 50` cycles of `(apply dose; integrate II)`. The residual is `≈ exp(−50·λ_slow·II)` — negligible for typical PK, but tens of percent for a slow disposition, exactly as measured for the analytical walk in #908.

2. **Silence.** `note_ss_nonconvergence_if_capped` is wired only into the ODE **input-rate** (absorption-kernel) branch (`src/ode/predictions.rs`, the `input_rate_consumes_cmt` path). A capped ordinary bolus/infusion equilibration emits no warning. #908 wired this into the analytical fallback; the ODE bolus/infusion path was left out of scope.

## The exact solve already exists and applies

For a **linear** ODE disposition the periodic steady state is the same affine fixed point the input-rate branch and the analytical walk already use: `crate::dosing::periodic_ss_fixed_point_g` (moved there in #908; originally #835). The ordinary bolus case is `advance_forced` = "add the bolus, integrate one `II`" and `advance_unforced` = "integrate one `II` with no dose"; the infusion case wraps the active-window RHS as the existing code already does. The linearity self-check declines a genuinely **nonlinear** RHS (Michaelis–Menten, etc.), which correctly falls back to the capped train — now with the warning attached.

## Suggested scope

- Route `equilibrate_ss_state`'s bolus and infusion branches through `periodic_ss_fixed_point_g` first, with the solver's own `reltol`/`abstol` (not `0` — unlike the analytical walk, the RK45 propagation carries real solver noise, so the existing `256·reltol` linearity tolerance is correct here).
- Mirror it in the dual twin `equilibrate_ss_state_g` so FOCE/FOCEI differentiates the exact steady state, exactly as #908 did for the analytical pair. This is the harder half — it carries the modeled-window / lagtime / rate-off saltation duals.
[...]

Issue #867, opened 2026-07-16
SS-absorption predict/simulate: 50-cycle equilibration under-converges for nonlinear heavy-accumulation models


TeunP:
### Summary

For a **steady-state (`SS=1`) dose into a built-in absorption compartment** with a **nonlinear** disposition (e.g. Michaelis–Menten) that **accumulates heavily** (elimination half-life ≫ dosing interval `II`), `predict()` / `simulate()` return an SS trough materially below the true periodic steady state.

The closed-form fast path (`equilibrate_ss_input_rate_fixed_point`) correctly *declines* a nonlinear disposition, so equilibration falls to the explicit pulse-train iteration (`equilibrate_ss_state`), which is capped at `SS_EQUILIBRATION_CYCLES = 50`. When the per-cycle carryover ratio `J → 1`, 50 cycles is far short of convergence and the returned trough is silently too low.

### This is pre-existing — not a #835 regression

This behaviour predates #835 and is unchanged by it. #835 only made the **linear** closed-form fast path actually fire, which is *exact* for any accumulation (the closed form `u_ss = (I − M)⁻¹·b` sums the infinite periodic history directly). The nonlinear path has always used the capped iteration (introduced with SS-into-absorption, #719 / #834). Filing separately so it isn't lost.

### Evidence

Probe: 1-cpt, `first_order(ka=2)` absorption into MM elimination `d/dt central = R_in − Vmax·central/(Km+central)`, `amt=1000`, `II=8`, default `ode_reltol=1e-4`. Ground truth = a 2500-dose explicit run-in observed at the same interval phase.

| Vmax | Km   | J (est) | equil cycles | predicted trough | true trough | rel. error |
|------|------|---------|--------------|------------------|-------------|-----------|
| 130  | 400  | 0.90    | 50 (capped)  | 3954             | 6351        | 38% low   |
| 130  | 1000 | 0.96    | 50 (capped)  | 7095             | 16515       | 57% low   |
| 130  | 5000 | 0.99    | 50 (capped)  | 16240            | 77575       | 79% low   |

Low-accumulation cases are accurate (e.g. `Vmax=150, Km=100`, J≈0.14 → 0% error), so this only bites slowly-accumulating saturable models.

### Impact

SS-absorption `predict()` / `simulate()` for slowly-accumulating **saturable** drugs (phenytoin-like). The **linear** case is exact; bolus / infusion into a *plain* compartment under SS is unaffected (separate equilibration).

### Fix directions

[...]

Issue #835, opened 2026-07-13
SS-absorption fitting: analytic dual sensitivities for the linear SS fixed-point (FOCEI parity)


TeunP:
Follow-up to #719 gap 1 (#834).

Gap 1 landed SS dosing into a built-in absorption compartment for `predict()`/`simulate()` (fast, NONMEM-anchored) and for `fit()` (converges). The SS trough is a closed-form linear fixed point `u_ss = (I − M)⁻¹·b` (`ode::predictions::equilibrate_ss_input_rate_fixed_point`), so predictions are cheap. But an SS-absorption subject's **sensitivities** are still declined to the finite-difference fallback (`ode_tvcov_supported` / `ode_iov_subject_supported` return `false` for `has_ss && !ode.input_rate.is_empty()`), so fitting a large SS-absorption dataset is slower than an analytic ODE model.

### Scope
Carry the fixed point over `Dual1`/`Dual2` — the implicit-function derivative `du_ss/dp = (I − M)⁻¹·(dM·u_ss + db)` (and its second order for the FOCEI outer Laplace Hessian) — in `sens/ode_provider.rs`, then flip the two SS×input-rate gate declines to admit the smooth density kernels. Validate with the existing dual-vs-FD harness (`check_vs_production` / `check_hessian_vs_fd_of_grad`) and re-fit the #719 NONMEM anchor to full-speed FOCEI. This is the SS analogue of #486 design B (modeled-dose + SS), already flagged there as a follow-up.

### Out of scope
- SS into a `zero_order()` window (periodic-window equilibration, `E_ABSORPTION_SS_ZERO_ORDER`).
- SS + absorption lagtime (`E_ABSORPTION_SS_LAG`).

Issue #375, opened 2026-06-16
fix(pk): event-driven path panics on infusion into an oral depot (cmt 1) instead of a graceful error


TeunP:
## Context

Follow-up from the review of #350 / #351.

PR #350 added support for infusion into the **central** compartment of oral
models on the event-driven analytical path. Infusion into an oral model's
**depot** (`cmt 1`) — and oral **peripheral** infusion — remain unsupported,
and the event-driven path handles "unsupported" by `panic!`-ing rather than
returning an error.

## Problem

`event_driven_predictions()` returns `Vec<f64>` — no error channel — so
`propagate_with_bounds()` `panic!`s when it meets an infusion it can't route.

`src/pk/event_driven.rs`, the infusion-routing match in `propagate_with_bounds`
(~L672):

```rust
_ => panic!(
    "event-driven PK: infusion into compartment {} not supported \
     for model {:?}. ...",
    d.cmt, pk_model
),
```

This is **reachable from user data**: any oral model with a `RATE>0` dose into
`cmt 1` (or an oral peripheral cmt) that *also* has a time-varying covariate, an
EVID=3/4 reset, or IOV is routed to the event-driven path and aborts the
process. A library should surface this as a propagated `Err` the caller (CLI /
ferx-r) can report — not a panic that crashes the R session. (A sibling
out-of-range dose-compartment `panic!` ~L523 shares the same limitation.)

## Note on current behaviour

The two analytical paths disagree on oral-depot infusion: the superposition path
silently bypasses the depot to central, while the event-driven path panics.
Whatever the canonical semantics (graceful error, or NONMEM-style
zero-order-into-depot), both paths should agree.

## Suggested direction

Prefer rejecting (or defining) oral-depot / oral-peripheral infusion **upfront**
at the dose-validation layer — the same place coded/negative `RATE` is already
rejected (#324) — so the prediction hot path stays panic-free and both
analytical paths behave identically. Alternatively, thread a `Result` out of the
prediction API.

## References
- PR #350 / #351 (the central-infusion fix this follows up)
- #350 "Open questions" scoped depot-infusion semantics out deliberately.

PR #915, opened 2026-07-22
fix(pk): solve the analytical steady state exactly instead of truncating it [closes #908]

### LINKED ISSUE #908: perf(pk): solve the analytical steady state exactly instead of a truncated 50-cycle pulse train
## Problem

The analytical event-driven walk equilibrates a steady-state dose by running a **truncated pulse train** — up to `SS_EQUILIBRATION_CYCLES = 50` cycles with an early stop at `SS_EQUILIBRATION_TOL` (`src/dosing.rs`). The residual tail is `≈ exp(−50·λ_slow·II)`, where `λ_slow` is the slowest disposition rate constant.

For fast PK that is far below precision. For slow PK it is not, and the early stop never fires there (it needs the increment to be negligible, which is exactly what a slow mode prevents). Measured relative difference from the exact periodic steady state:

| model | `II` | shift |
|---|---|---|
| `one_cpt_oral`, CL 0.1, V 50 (t½ ≈ 350 h) | 12 | **3.0e-1** |
| `two_cpt_iv`, Q 0.5, V2 500 | 12 | **8.9e-2** |
| `three_cpt_iv`, Q3 0.5, V3 400 | 12 | 7.6e-2 |
| `three_cpt_iv` (the repo's anchor parameters) | 6 | 3.8e-2 |
| `three_cpt_iv` (same) | 12 | 6.0e-3 |

A 30 % difference on a one-compartment model is not an edge case — that is an ordinary long-half-life drug.

Two things make it worse than the numbers suggest:

1. **It is silent.** `equilibrate_ss_state_event_driven` (`src/pk/event_driven.rs`) calls `record_ss_equilibration_cycles` but never `note_ss_nonconvergence_if_capped`, whose only call site is the ODE path (`src/ode/predictions.rs`, added by #867/#874). A capped analytical equilibration produces no warning.
2. **The same dataset can move between representations.** Dose superposition uses the *exact* periodic closed form and is unaffected. A subject is routed to the pulse-train walk by a time-varying covariate, an `EVID=3/4` reset, IOV, or — since #375 — a dose into a non-default compartment. So adding one unrelated dose row can shift an `SS` prediction by several percent.

## The exact solution already exists in this repo

For a **linear** disposition the periodic steady state has a closed form: one cycle is an affine map `u ↦ M·u + b`, so the fixed point is `u_ss = (I − M)⁻¹·b`. The ODE engine already does exactly this (`src/ode/predictions.rs`, via `crate::sens::linsolve::solve_linear_system_g`, added by #834/#835), including a dual version for gradients.

The analytical models' disposition is *always* linear. Porting that solve to `equilibrate_ss_state_event_driven` and its dual twin `equilibrate_ss_g` (`src/sens/propagate.rs`) would make the walk exact, remove the truncation entirely, and make it agree with dose superposition by construction rather than to a tolerance. It would also be **faster** — one small linear solve instead of up to 50 propagator cycles.

## Suggested scope

- Replace the pulse-train loop in `equilibrate_ss_state_event_driven` and `equilibrate_ss_g` with the `(I − M)⁻¹·b` solve, reusing `sens::linsolve` as the ODE path does.
- `M` and `b` come from one cycle of the existing propagators, so no new closed form is needed — this is a re-use, not a derivation.
- Keep the pulse train as a fallback for any non-linear case (there should be none on the analytical models; assert it).
- Oracle: the superposition path's exact SS closed form — e.g. `one_cpt_iv_bolus_ss_g` returns `(amt/V)·e^{−k·t}/(1 − e^{−k·II})` (`src/sens/one_cpt.rs`). Walk and superposition must then agree to f64 precision, not to a truncation tolerance, on every model including the slow cases above. Plus a `Dual2`-vs-FD parity test for the gradient twin, per CLAUDE.md.
- The existing NONMEM `ss_cmt0` / `ss_cmt1` anchors (`nonmem_anchor/dose_cmt_ss_cmt*.ctl`) stay as the cross-tool check.

## Interim

The behaviour is documented in `CHANGELOG.md` under #375 and in `docs/model-file/steady-state.qmd` ("Why `N = 50`?"), with the governing rule and measured examples. Documenting it was the stopgap; this issue is the fix.

Separately and more cheaply: wiring `note_ss_nonconvergence_if_capped` into the analytical walk would at least make the truncation loud while the exact solve is pending.

## References

- Surfaced by the adversarial review of #907 (follow-up to #375 / #896).
- Truncation constants: `SS_EQUILIBRATION_CYCLES`, `SS_EQUILIBRATION_TOL` — `src/dosing.rs`.
- Exact linear SS solve to re-use: `src/ode/predictions.rs` + `src/sens/linsolve.rs` (#834/#835).
- The unwired warning: `note_ss_nonconvergence_if_capped` (#867/#874).

TeunP:
## Why

The analytical event-driven walk equilibrated an `SS=1` dose by iterating a **truncated pulse train** — up to `SS_EQUILIBRATION_CYCLES = 50` cycles of *(apply dose; propagate one `II`)*, with an early stop at `SS_EQUILIBRATION_TOL`. The leftover is `≈ exp(−50·λ_slow·II)`.

For fast PK that is far below precision. For slow PK it is not — and the early stop never fires there, because it needs the per-cycle increment to be negligible, which is exactly what a slow mode prevents. Worse, it was **silent**: `note_ss_nonconvergence_if_capped` had only ever been wired into the ODE path.

Three things made this more than a corner case:

1. Routing to the walk is per *subject*, and is triggered by a time-varying covariate, an `EVID=3/4` reset, IOV, or — since #375 — a dose into a non-default compartment. So adding one unrelated dose row could move an `SS` prediction by tens of percent.
2. Dose superposition, which serves everything else, uses the *exact* periodic closed form. The same dataset in two representations disagreed.
3. The gradient walk truncated in lockstep with the value walk by construction (`ss_dual_cycle_should_stop` exists for exactly that), so FOCE/FOCEI differentiated a truncated steady state.

Filed as #908 out of the adversarial review of #907, which documented the caveat as a stopgap. This is the fix.

## What changed

One dosing cycle of these models is an **affine map** `u ↦ M·u + b`: the analytical disposition is always linear (constant-coefficient), and the dose enters additively. So the periodic steady state has a closed form,

```
u_ss = (I − M)⁻¹·b
```

`M`'s columns (the homogeneous one-cycle response of each unit basis state) and `b` (the forced response from a zero state) both come from one cycle of the **same propagators the walk already runs** — this is a re-use, not a new formula.

**No new solver either.** `equilibrate_ss_input_rate_fixed_point_g` already owned exactly this algebra for the ODE input-rate path (#835). It was moved verbatim to `crate::dosing` as `periodic_ss_fixed_point_g` — `dosing.rs` is the module that exists so `pk/` and `sens/` need not depend upward on `ode/` — and the analytical callers pass `reltol = abstol = 0`, which degenerates its solver-noise-scaled linearity tolerance to the `1e-9·scale` rounding floor that exact propagators want. First commit is pure code motion; 197 SS tests unchanged.

Both walks flip in the **same commit**, deliberately. Converting only the value walk would leave FOCE/FOCEI differentiating a steady state it never predicted — the exact value/gradient mismatch that stalls the outer line search. Over a dual `T` the same solve returns the implicit-function derivative `∂u_ss/∂p = (I − M)⁻¹·(∂b/∂p − ∂M/∂p·u_ss)` straight out of `solve_linear_system_g`, with no hand-assembled `dM`/`db` — a strictly better object than the derivative of a truncated recursion.

The pulse train survives only for a **singular `I − M`** — a zero disposition rate constant, i.e. a compartment that never empties. No periodic steady state exists there, and that path now raises the #867 non-convergence warning instead of returning a silently truncated state.

## Alternatives considered

**Documenting it** — that was #907, and it is what #908 exists to replace.

**Wiring only the warning** into the analytical walk (the issue's "separately and more cheaply" option). It would have made the truncation loud without fixing it, and the exact solve is *faster*, so there was no accuracy/cost trade to make.

**Also converting the ODE bolus/infusion path** in this PR, which would let the caveat be deleted outright rather than narrowed. Measured first: the three tests that compare the analytical walk against its ODE twin have 5.9–7.5 orders of headroom on those fixtures (residual 8.2e-12 and 2.8e-14 against `1e-6` tolerances), so analytical-only does not disturb them. Kept out to hold the blast radius down; the ODE dual twin carries the modeled-window / lag / saltation complexity. Worth its own issue.

## Numerical validation

**Oracle: the dose-superposition SS closed forms** — the exact periodic solution, which #908 names as the oracle. New test `ss_equilibration_matches_the_exact_periodic_closed_form` compares the walk's trough against them on every walk-supported model:

| model | rel. difference |
|---|---|
| 1cpt IV bolus (fast) | **0.0 — bit-identical** |
| 1cpt oral (fast) | **0.0 — bit-identical** |
| 1cpt IV bolus (slow, CL 0.1) | 2.2e-16 |
| 1cpt IV infusion | 1.7e-16 |
| 1cpt oral (slow) | 2.2e-16 |
| 2cpt IV bolus (slow, Q 0.5 V2 500) | 3.2e-16 |
| 2cpt IV infusion | 7.6e-16 |
| 2cpt oral | 2.0e-16 |
| 3cpt IV bolus (slow, cond ≈ 86) | 4.2e-13 |
| 3cpt oral | 2.6e-14 |

Agreement is at f64 precision rather than to a tolerance, so the two representations of one dataset now agree **by construction**.

**Error removed** (measured, walk trough vs the closed form):

| model | `II` | was |
|---|---|---|
| `one_cpt_oral` (CL 0.1, V 50 — t½ ≈ 350 h) | 12 | **3.0e-1** |
| `two_cpt_iv` (Q 0.5, V2 500) — central | 12 | 1.0e-1 |
| `two_cpt_iv` (Q 0.5, V2 500) — peripheral **amount** | 12 | **5.8e-1** |
| `three_cpt_iv` (Q3 0.5, V3 400) — central | 12 | 8.2e-2 |
| `three_cpt_iv` (Q3 0.5, V3 400) — third-cpt **amount** | 12 | **5.2e-1** |
| `three_cpt_iv` (the repo's anchor params) | 24 | 2.3e-3 |

Compartment *amounts* — what `[derived]` and per-compartment sdtab columns report — were hit considerably harder than concentrations. #908's table quotes concentrations only.

**NONMEM 7.6.0, slow regime — the anchor that actually exercises the fix.** The existing `ss_cmt0` / `ss_cmt1` anchors sit at `CL = 5` (`λ·II = 1.2`), where the old truncation left `e^{−60} ≈ 1e-26`; they pass identically before and after and cannot see the bug. New `ss_slow_advan1` / `ss_slow_advan2` (`ADVAN1`/`ADVAN2 TRANS2`, `CL = 0.1`, t½ ≈ 347 h, `SS=1 II=12`) put NONMEM in the affected regime — its `ADVAN` steady state is the exact periodic closed form, `PRED(0) = 84.337333295 = (100/50)/(1 − e^{−0.024})` to every printed digit, where the pre-fix walk read **29.4 % low**. `slow_accumulation_ss_matches_nonmem` pins both the walk and the default dispatch to the IV and oral anchors at `1e-9`. Control streams + datasets + verbatim `results/*.{lst,tab}` are committed; every PRED was **reproduced first-hand via `nmfe76`** on NONMEM 7.6.0 at zero relative deviation. (The old fast anchors still pass too, unchanged.)

**Two `f64`-parity tests passed unchanged** at their existing `1e-12` / `1e-10` tolerances, so no relaxation was needed. Note that the two 3-cpt SS cases in `event_walk_g_3cpt_matches_production_f64` were previously pinning a value **2.3e-3 away from the truth** to 1e-10 precision — both walks truncated identically, so the test was green. They now pin the right number.

## Performance

- [x] Faster — `n + 4` propagator calls plus one `n×n` solve (`n ≤ 4`) replace up to 50 propagations.

The 50-cycle cap was reached **routinely**, not only on pathological inputs — 6 of 10 measured cases, including both ODE-twin fixtures and both 3-cpt parity cases. (Hitting the cap is not the same as being inaccurate: on the fast fixtures the residual there is still ~8e-12.)

Benchmark (same worktree, sequential, test *execution* time only, excludes compilation):

```
197 lib tests matching `ss_`:   54.89 s (before)  →  20.00 s (after)
```

## Tests

**Verified by mutation, not assumed.** Forcing the value walk back to the pulse train (`n_states` → `0` at the call site):

- the `cycles == 1` assertion fires on the *first, fastest* case — so a silent fall back to the train cannot pass unnoticed even where the value would still be right;
- with that assertion disabled, the value check fails on the slow 1-cpt row at **57.54 vs 82.34 — a 30.1 % miss**, reproducing #908's headline figure from the test suite.

Forcing only the **dual** walk back (the dangerous half — this is the gap #907 was written to close) fails three existing tests: `event_walk_g_3cpt_matches_production_f64`, `provider_modeled_duration_ss_iov_kappa_coupled_matches_fd_of_predict_iov`, and `ss_cmt_zero_gradient_equilibrates_like_the_default_compartment`. The value/gradient lockstep is machine-enforced, in both directions.

New coverage:

- `ss_equilibration_matches_the_exact_periodic_closed_form` — the oracle table above, plus `cycles == 1` so the exact path cannot stop being taken silently.
- `ss_equilibration_falls_back_when_no_steady_state_exists` — both routes to a singular `I − M` (`CL = 0`, where `ke = 0` makes the propagator the identity; `Q = 0`, where the eigen solve declines), asserting the capped train runs *and* that the non-convergence warning now fires from the analytical walk. Reachable from a Tier-1 test only because `PkParams` is built directly, bypassing model validation.
- `ss_ill_conditioned_gradient_matches_fd_of_production` — the new derivative where `cond(I − M) ≈ 57`, three orders worse-conditioned than any existing SS gradient test.

Tolerances tightened where the truncation had been the limiting error: `2e-3 → 1e-12` (×2) in `event_driven_applies_oral_ss_infusion_into_central` — a nine-order tightening — and `1e-7 → 1e-12` (×4) across the SS-vs-closed-form tests.

**One test rolls its own finite differences instead of calling `check_full_provider_vs_fd`, and the reason is worth a look.** At `cond ≈ 57` the mixed second derivative is a near-total cancellation: `∂²f/∂η_CL∂η_V ≈ −1.96e-2` against `f ≈ 6.25e2`, so the 4-point stencil subtracts values agreeing to ~12 digits and its roundoff floor `≈ eps·f/h²` swamps the answer. Measured against the analytic value:

| `h` | 1e-2 | 3e-3 | 1e-3 | 3e-4 | **1e-4** | 3e-5 | 1e-5 |
| rel. err | 8.2e-5 | **3.5e-7** | 4.8e-5 | 5.7e-4 | **5.1e-3** | 7.3e-2 | 7.2e-1 |

Truncation dominates left, roundoff right, and the shared harness's `h = 1e-4` sits on the wrong side of the minimum. The dominant element `∂²f/∂η_CL² = 6.2007e2` shows the same V-shape with its floor at 1.2e-8. **The FD reference is what fails there, not the analytic derivative** — so that test pins second derivatives at the measured optimum `h = 3e-3` and leaves the well-conditioned first derivatives at the usual step. The table is in the test's docstring.

## Cross-repo dependency

No public API surface changes; `periodic_ss_fixed_point_g` is `pub(crate)`.

## Breaking changes

- [x] None — but see the CHANGELOG note: **`SS=1` results from earlier versions on this path are worth regenerating.** No API or file-format change; the numbers simply become correct.

## Docs

- [x] `CHANGELOG.md` `[Unreleased]` — new `Fixed` entry; the #375 caveat block it replaces is **deleted**, since that caveat was entirely about this path.
- [x] `docs/model-file/steady-state.qmd` — "Why `N = 50`?" split: the analytical walk now gets its own section deriving the exact fixed point, and the `N = 50` discussion is scoped to the ODE path, which still expands the train. Cost table and the supported-paths table updated. A callout tells users with older `SS=1` results to regenerate them.

## Checklist

- [x] `cargo clippy` clean
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `equilibrate_ss_g` is generic over `PkNum` throughout; the solve rides `solve_linear_system_g`, which is the same generic body the ODE path uses.

## Reviewer hints

Focus on **`periodic_ss_fixed_point_g`'s two callers**, specifically what goes in the *forced* half versus the *unforced* half. The modeled-window duals (`RATE=-1/-2`) belong to `b` only: `M` is the homogeneous monodromy and carries `∂P/∂(θ,η)` through the dual disposition params, but the infusion window is inhomogeneous forcing. Threading it into the unforced call would corrupt the propagator, and the value walk would not notice.

Second: the closures share **one** `EigenCacheG` through a `RefCell` (the helper takes `Fn`, not `FnMut`). Without that, the eigen solve would run `n + 4` times instead of once and the change would be slower, not faster.

Skimmable: the CHANGELOG and docs prose, and the mechanical tolerance tightenings.

## Follow-up filed

The ODE bolus/infusion steady state (`ode/predictions.rs`, `sens/ode_provider.rs`) still expands the capped pulse train, and — like the analytical walk before this PR — does so **silently**: `note_ss_nonconvergence_if_capped` is wired only into its input-rate branch. The same helper applies (closures become `solve_ode` over one `II`), which would cost `n + 4` ODE solves instead of 50 and let the caveat go entirely. Tracked as **#914**. Only after that does the #375/#908 caveat disappear outright and the analytical↔ODE SS twin tests tighten from tolerance to exactness.


TeunP:
## Review — exact analytic SS `(I−M)⁻¹·b` (xhigh, adversarial)

**Verdict: numerically sound, no correctness blockers. One real doc error (wrong version label). Ship after that.**

Not CI-trust — built PR head `a648c0e9` in an isolated worktree (default features, so the analytic `Dual2` path is live, not FD-only `ci`) and ran everything below first-hand.

### Verified independently
- All 5 new tests pass locally: the 4 lib tests + `slow_accumulation_ss_matches_nonmem`.
- **Mutation reproduced.** Forcing the value walk back to the pulse train fails `ss_equilibration_matches_the_exact_periodic_closed_form` instantly on the first case (`cycles: left 24, right 1`). The `cycles == 1` guard is load-bearing, not decorative.
- **NONMEM anchor is real.** Committed `ss_slow_advan1.tab` has `PRED(0) = 8.4337333295E+01`, which is `(100/50)/(1 − e^{−0.024})` exact; the test's `NONMEM_SLOW_IV`/`_ORAL` arrays are the t=1/4/8 obs rows, transcribed correctly.

### Core correctness — the flagged risk, confirmed right
The one thing that would silently corrupt gradients is the forced/unforced split. Checked both walks line-by-line:
- Bolus jump (`pk.f * amt`, F-dual) and the infusion window + its `RATE=-1/-2` window duals (`synthetic_inf`) enter **`b` only** (forced closure). Unforced passes `&[]` for `inf_dual`, so `M` is the pure homogeneous monodromy carrying `∂P/∂(θ,η)`. ✓
- The forced closure's per-cycle op is **byte-identical** to the old train iteration → the affine map is the exact linearization → the fixed point is exact, not approximate.
- Dual carries the implicit-function derivative through generic `solve_linear_system_g`; no hand-assembled `dM`/`db`. Value + gradient flip in the same commit; lockstep enforced by `event_walk_g_3cpt_matches_production_f64` and the `CMT=0` test.
- ODE caller keeps `reltol/abstol`; analytical passes `0,0` → `vtol` degenerates to the `1e-9·scale` floor. `EigenCacheG` shared via `RefCell`, closures called sequentially → no double-borrow. `z0` subtraction is a no-op for the analytical (always 0) path.

### Findings

| # | Sev | Location | Issue | Fix |
|---|-----|----------|-------|-----|
| 1 | **Low (doc)** | `docs/model-file/steady-state.qmd:201` (callout header) | Reads **"## Before v0.2.0 this path truncated"** — false. `[0.2.0]` shipped 2026-07-03 *with* the truncation (the #375 blast-radius widening is itself under `[Unreleased]`, after 0.2.0), and `Cargo.toml` is still `v0.2.0`. This fix is under `[Unreleased]` → next release. | "Through v0.2.0", or reference the unreleased/next version — not "before". |
| 2 | Info | `src/sens/propagate.rs` dual fallback | Omits the #867 `note_ss_nonconvergence_if_capped` that the value fallback fires (`src/pk/event_driven.rs:473`). Deliberate (avoids a double-warn), but implicit — relies on the value walk always co-running. No bug today. | One-line comment: value path owns the warning; dual stays silent to avoid duplicates. |
| 3 | Info (won't-fix) | `Q=0` singular fallback | Returns a physically-wrong state (sum of boluses, no central elimination) because `two_cpt_eigen_g` declines wholesale. Unreachable via validation (log-params never hit exactly 0); the warning fires; honestly documented. | none — note only. |

Finding 1 is the only actionable item.

### Checked and dismissed
- Near-singular-but-nonzero `I − M` (slow drug): the exact solve handles it — that's the fix. `ss_ill_conditioned_gradient_matches_fd_of_production` (cond ≈ 57) validates the derivative, with an honest FD-minimum table justifying the custom `h = 3e-3` (the shared harness's `h = 1e-4` sits past the stencil's roundoff wall — the FD *reference* fails there, not the analytic derivative).
- No stale references to the moved `equilibrate_ss_input_rate_fixed_point_g`; the 3 doc links are updated.
- `record_ss_equilibration_cycles(1)` sentinel: only test-side readers, no production threshold logic keys off the cycle count.

<sub>🤖 xhigh review via Claude Code.</sub>

TeunP:
**Addressed in `ed5b08c3`:**

- **Finding 1** (doc, actionable) — fixed. Callout header `docs/model-file/steady-state.qmd:201` now reads **"Through v0.2.0 this path truncated"**. Agreed: the truncation shipped *in* 0.2.0 and this fix is `[Unreleased]`, so "before" was wrong.
- **Finding 2** (info) — added the explicit note at the dual fallback (`src/sens/propagate.rs`): the value walk owns the #867 cap warning and the two walks co-run per subject, so warning here too would double-emit.
- **Finding 3** (info, won't-fix) — no change, as flagged. `Q=0` singular fallback is unreachable via validation (log-params never hit exactly 0), the warning fires, and it's documented.

Comment/doc only; no behavioural change. Thanks for the first-hand verification (mutation repro + NONMEM `PRED(0)` check).

Issue #703, opened 2026-07-06
Different results using different number of threads


roninsightrx:
I'm getting different results when using 4 threads vs 15 threads, for a a model that I'm running with FOCEI. Getting lower OFV with 15 threads.

Issue #367, opened 2026-06-15
epic(estimation): NONMEM-class analytic outer gradient — bundle inner-loop, AD, and optimizer-default work


hiddevandebeek:
## North star

A **NONMEM-class FOCE/FOCEI estimator whose outer (population) gradient is analytic — not finite-differenced** — fast enough and reproducible enough to be the default optimizer everywhere. This epic bundles the inner-loop, outer-gradient, AD-strategy, and optimizer-default threads that all serve that one goal, and sequences them.

## Blueprint: Almquist, Leander & Jirstrand 2015 (JPKPD)

*"Using sensitivity equations for computing gradients of the FOCE and FOCEI approximations to the population likelihood"* — [PMC4432110](https://pmc.ncbi.nlm.nih.gov/articles/PMC4432110/). This paper is the reference design for the whole program. Key results, mapped to ferx:

**The exact FOCE/FOCEI outer gradient (their Eq. 23):**
```
d logL_F/dθ_m = Σ_i [ ∂l_i/∂θ_m|_η̂ + ½ η̂ᵀΩ⁻¹(∂Ω/∂θ_m)Ω⁻¹η̂ − ½ tr(Ω⁻¹ ∂Ω/∂θ_m) ]
                − ½ Σ_i tr( H_i⁻¹(η̂) · dH_i(η̂)/dθ_m )
```
splits into two pieces, and the split is the plan:

1. **Data-likelihood term — `dη̂/dθ` cancels (Eq. 27).** Because `∂l_i/∂η = 0` at the inner optimum, the implicit `(∂l_i/∂η)(dη̂/dθ)` term vanishes; this dominant part of the outer gradient is **exact with η̂ held fixed**. → This is precisely ferx's current fixed-EBE gradient. The paper proves that's correct, not an approximation.
2. **`log|H|` Hessian-determinant term — `dη̂/dθ` does NOT cancel.** `dH_i(η̂)/dθ` needs the EBE response explicitly (Eq. 46):
   ```
   dη̂/dθ = −(d²l_i/dη²)⁻¹ · (d²l_i/dη∂θ)
   ```
   i.e. **second-order sensitivities** (`d²x/dη²`, `d²x/dη∂θ`).

[...]

PR #917, opened 2026-07-22
test(iov): assert bit-identical OFV for DSL-vs-column occasion fit

TeunP:
## Why

`test_iov_occasion_dsl_matches_column_fit` was reported flaky under heavy CPU load: the `(r_col.ofv - r_dsl.ofv).abs() < 1e-6` assertion failed once in a full parallel `cargo test --lib` run, while passing in isolation and on a re-run. The working hypothesis was rayon work-steal reduction-order non-determinism amplified by derivative-free BOBYQA.

Investigation showed the premise no longer holds:
- The FOCE OFV reduction is **already order-stable** — `foce_population_nll{,_iov}` collect per-subject NLLs into an ordered `Vec` then sum serially (#703), so it is thread-count-independent.
- The inner EBE loop (`run_inner_loop_warm`) is order-preserving, and `find_ebe` keys on occasion **structure** (group indices via `split_obs_by_occasion`'s first-appearance order), not label values.
- So the two fits — occasions from `iov_column` `{1,2}` vs the `time(3.5)` DSL rule `{0,1}`, identical partition — are **bit-identical**, not merely close.

Measured: `diff = 0.0` (both OFVs `0x4062a9f411bee591`) across **145 fits** — 25 reps x {1, 8} threads, plus **120/120 under 12-hog CPU load**. The reported failure did not reproduce on current code (likely a stale pre-#703 binary or a misattribution). The `< 1e-6` tolerance was comparing two values that are exactly equal, and loosening it is the one change that would let a *real* future divergence pass silently.

## What changed

Replaced the tolerance assertion with exact bit-equality plus a finite guard:

```rust
assert!(r_col.ofv.is_finite() && r_dsl.ofv.is_finite(), …);
assert_eq!(r_col.ofv.to_bits(), r_dsl.ofv.to_bits(), …);
```

Added a comment explaining why exact equality is correct here and instructing maintainers **not** to loosen it (a failure would mean a real regression — broken partition equivalence or nondeterminism leaking into estimation). No production code touched; the fit result itself is unchanged.

## Alternatives considered

- **Loosen tolerance to ~1e-3** (the originally-proposed fix): rejected — weakens a guarantee that holds exactly (`diff = 0.0`); a genuine col!=dsl regression up to ~1e-3 would pass silently.
- **Make the per-subject reduction order-stable** (the other proposed fix): already implemented in #703 — nothing left to do.
- **Assert partition equivalence directly instead of via two full fits**: redundant — the derivation is already unit-tested (`apply_iov_occasion_*` tests) and the end-to-end numeric equivalence is a stronger, still-non-flaky claim.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable (test-only; the fit and its OFV are unchanged — this only tightens how the test compares two already-computed OFVs)

Determinism evidence: `diff = 0.0` across 145 fits (25 reps x {1,8} threads; 120/120 under 12-hog CPU load).

## Performance
- [x] No significant impact expected (test-only)

## Tests
- [x] The modified test **is** the regression guard — it now fails on any real col/dsl divergence instead of only on a >1e-6 one. `cargo test --lib --features ci api::iov_integration::test_iov_occasion_dsl_matches_column_fit` passes.
- [x] No new test file needed (this hardens an existing one)

- [x] Not applicable (internal test change, no API surface)

## Docs
- [x] No user-visible change; no `CHANGELOG.md` entry (test-only, per CLAUDE.md)

## Checklist
- [ ] `cargo clippy` clean — **not run** (trivial `to_bits`/`assert_eq!` edit; rustfmt `--check` clean, target test green)
- [x] `Dual2` sensitivity path — N/A (test-only, not gradient-reachable)
- [x] `Cargo.toml` version bump — N/A (non-breaking)

## Reviewer hints

The one judgment call: is exact bit-equality too strict? Argument that it is not — col and dsl run **in the same process on the same CPU in the same build**, so any libm/FMA quirk hits both identically; the assert compares them *to each other*, never to a golden constant, so it is platform-robust (holds on x86 CI and arm64/Rosetta dev alike). It can only fail on (a) a partition-derivation regression or (b) nondeterminism entering estimation — both real bugs. Full rationale is in the added code comment and the commit body.

## Open questions

- Was the original 2026-07-22 failure on a pre-#703 binary? I could not reproduce it on current `main`; treating it as resolved-unreproducible. Shout if you want a tracking issue instead.

### DIFF
--- a/src/api/tests/iov_integration.rs
+++ b/src/api/tests/iov_integration.rs
@@ -694,9 +694,41 @@ fn test_iov_occasion_dsl_matches_column_fit() {
         r_dsl.ofv,
         r_col.ofv,
     );
+    // The same bit-identity thesis extends to the per-occasion kappa EBEs: same
+    // subject data, same theta/Omega trajectory, same first-occ->g0/second-occ->g1
+    // group map, so every kappa component must match bit-for-bit. This is a
+    // *stricter* guard than the OFV check — a summed OFV can round-trip to the
+    // same bits while a per-kappa ULP diverges — so pin it exactly too.
     assert_eq!(r_col.ebe_kappas.len(), r_dsl.ebe_kappas.len());
-    for (a, b) in r_col.ebe_kappas.iter().zip(&r_dsl.ebe_kappas) {
-        assert_eq!(a.len(), b.len(), "same number of occasions per subject");
+    for (subj, (a, b)) in r_col.ebe_kappas.iter().zip(&r_dsl.ebe_kappas).enumerate() {
+        assert_eq!(
+            a.len(),
+            b.len(),
+            "subject {} must have the same number of occasions",
+            subj,
+        );
+        for (k, (ka, kb)) in a.iter().zip(b).enumerate() {
+            assert_eq!(
+                ka.len(),
+                kb.len(),
+                "subject {} occasion {} must have the same kappa dimension",
+                subj,
+                k,
+            );
+            for (j, (va, vb)) in ka.iter().zip(kb.iter()).enumerate() {
+                assert_eq!(
+                    va.to_bits(),
+                    vb.to_bits(),
+                    "subject {} occasion {} kappa[{}]: DSL EBE {} must be \
+                     bit-identical to column EBE {}",
+                    subj,
+                    k,
+                    j,
+                    vb,
+                    va,
+                );
+            }
+        }
     }
 }
 
--- a/src/api/tests/iov_integration.rs
+++ b/src/api/tests/iov_integration.rs
@@ -662,11 +662,37 @@ fn test_iov_occasion_dsl_matches_column_fit() {
     let r_dsl = fit(&model, &pop, &model.default_params, &opts_dsl)
         .expect("DSL-occasion fit should succeed");
 
+    // The two fits share model + population and differ only in how each
+    // subject's occasion partition is derived; both produce the identical
+    // partition (relabelled 0/1 vs 1/2 — see above), and the FOCE pipeline is
+    // bit-reproducible regardless of the thread schedule: the per-subject
+    // objective reduces in a fixed subject order (collect-then-serial-sum,
+    // #703) and the inner EBE loop is order-preserving. So the two OFVs are
+    // *bit-identical*, not merely close.
+    //
+    // Assert that exactly. col and dsl run in the same process on the same CPU,
+    // so this compares two values that shared every FP op sequence — it is a
+    // 0-vs-0 check that cannot be perturbed by parallel-reduction order or CPU
+    // load (the class of flakiness a loose tolerance would only paper over).
+    // If it ever fails, do NOT loosen it: a real regression has broken
+    // partition equivalence (independently pinned by the `apply_iov_occasion_*`
+    // derivation tests above) or leaked nondeterminism into estimation. The
+    // sibling `test_iov_ode_fit_matches_analytical_twin` compares two *different*
+    // numeric engines (ODE vs closed form) and rightly uses a basin tolerance;
+    // this test does not.
     assert!(
-        (r_col.ofv - r_dsl.ofv).abs() < 1e-6,
-        "DSL occasion rule OFV {} must match column OFV {}",
+        r_col.ofv.is_finite() && r_dsl.ofv.is_finite(),
+        "both OFVs must be finite (col={}, dsl={})",
+        r_col.ofv,
         r_dsl.ofv,
-        r_col.ofv
+    );
+    assert_eq!(
+        r_col.ofv.to_bits(),
+        r_dsl.ofv.to_bits(),
+        "DSL-occasion OFV {} must be bit-identical to column OFV {} \
+         (same partition + deterministic pipeline)",
+        r_dsl.ofv,
+        r_col.ofv,
     );
     assert_eq!(r_col.ebe_kappas.len(), r_dsl.ebe_kappas.len());
     for (a, b) in r_col.ebe_kappas.iter().zip(&r_dsl.ebe_kappas) {


TeunP:
## Review — `test(iov): assert bit-identical OFV for DSL-vs-column occasion fit`

**Scope:** 1 file, +29/−3, test-only. Swaps the `(r_col.ofv - r_dsl.ofv).abs() < 1e-6` tolerance in `test_iov_occasion_dsl_matches_column_fit` for a finite-guard + `to_bits()` exact-equality pair plus an explanatory comment. No production code touched.

### Thesis is correct — verified independently, not taken on faith

The claim is that the column `{1,2}` fit and the `time(3.5)` DSL `{0,1}` fit produce a **bit-identical** OFV, so `1e-6` was comparing two exactly-equal values and loosening it would only mask a real future regression. Traced the whole chain at code level rather than trusting the PR narrative:

- **Occasion label value never enters arithmetic.** `split_obs_by_occasion` (`likelihood.rs:2209`) groups by **first-appearance order**, carrying the occasion id only as a `HashMap` key (ordering comes from the deterministic `occ_order` vec, not map iteration). `iov_occ_to_k` maps occ→enumerate index; `iov_combined_effect` (`likelihood.rs:2268`) slices kappa by the **group index `g`**, never the label. So col `{1,2}` and dsl `{0,1}` both map first-occ→g0, second-occ→g1 — the labels are a pure grouping key, perturbing zero FP ops.
- **Same group count K.** Fixture: col obs `[1,1,1,2,2,2]` + `dose_occ [1,2]` → K=2 (`iov_integration.rs:97`). dsl `time(3.5)` → obs `[0,0,0,1,1,1]` + `dose_occ [0,1]` → K=2 (derivation test `:608`). Identical stacked-η dimension `[η_bsv, κ_g0, κ_g1]`; no appended dose-only axes on either side.
- **#703 serial sum confirmed.** `foce_population_nll` and `_iov` (`likelihood.rs:2065`, `:2107`) both `par_iter().map(...).collect::<Vec>()` then `per_subj.iter().sum()` — thread-count-independent, as claimed.
- **Comment references are real.** Sibling `test_iov_ode_fit_matches_analytical_twin` exists (`:259`); the `apply_iov_occasion_*` derivation tests exist (`:597+`). Message arg order in both new asserts is correct.

Because `OFV_col(θ) == OFV_dsl(θ)` bit-for-bit at *every* θ, the two BOBYQA trajectories from the same start coincide, so the *final* OFVs are bit-identical — not just the per-eval ones. The tightening compares the two fits **to each other** (same process, CPU, build), never to a golden constant, so it is platform-robust. Agreed with the reasoning.

**Empirically confirmed:** `cargo test --lib --features ci api::iov_integration::test_iov_occasion_dsl_matches_column_fit` → `1 passed` (3.06s). The `assert_eq!(to_bits, to_bits)` holds, so the two OFVs really are bit-identical, not merely close. Cold build is warning-clean on the touched file (the one `n_subj` warning is pre-existing in `outer_optimizer.rs`, outside this diff).

### Findings

1. **[Medium — incomplete tightening] `ebe_kappas` values still unchecked.** Lines 671–674 assert only per-subject *lengths* (`a.len() == b.len()`). The PR's own bit-identity thesis implies the kappa EBEs are bit-identical too (same subject data, same θ/Ω trajectory, same group→axis map). Kappa bit-equality is a **stricter** guard than OFV bit-equality — a summed OFV can round-trip to identical bits while a per-kappa ULP diverges. Since the test is being hardened on exactly that thesis, tighten the inner loop to `assert_eq!(a[k].to_bits(), b[k].to_bits())`. Safe (values are deterministic) and completes the intent. Not a merge blocker.

2. **[Nit] Signed-zero / inf pedantry.** `to_bits()` treats `+0.0`/`−0.0` (and the two inf signs) as unequal. Irrelevant for an OFV (a NLL sum is never exactly ±0), and the finite-guard runs *first* and rejects NaN/inf before the bit-compare — the right design: it's what catches a both-NaN / both-inf case that bit-equality alone would silently pass. No action.

3. **[Process] clippy.** Checklist admits it wasn't run; the diff introduces no lint-triggering constructs (`assert!`/`assert_eq!`/`.to_bits()`/`.is_finite()` + a comment). Ran locally to close the gap.

### Verdict

Correct, well-reasoned, and the rationale survives adversarial tracing rather than resting on the PR narrative. The diff itself has no defects; empirical run is green. One worthwhile strengthening (finding 1). **Recommend: apply finding 1, then merge.**

TeunP:
### Clippy addendum

Followed up on the checklist's un-run clippy so the record is complete:

- **CI clippy is green, and this diff is clippy-clean.** CI runs `cargo clippy --no-default-features --features ci,markov` (`.github/workflows/ci.yml:192`) — no `--tests`, so it lints lib + bins only. This PR touches a test file and adds no lib code, so it introduces no lint surface CI checks.
- **Running with `--tests` surfaces pre-existing debt unrelated to this PR:** 4 `approx_constant` errors (deny-by-default correctness lint) from the literal `3.14` used as a V2 volume in `#[cfg(test)]` fixtures in `src/sens/two_cpt_explicit.rs` (landed with #367), plus ~367 style warnings across the tree. CI never compiles `#[cfg(test)]` code, so none of this reaches the CI clippy job — it's latent, not a gate failure.

**No action for #917.** Flagging only so "clippy not run" isn't left ambiguous. If the CI clippy job ever gains `--tests`, those fixtures need `#[allow(clippy::approx_constant)]` or `3.14`→`3.15` first — orthogonal to this change.

TeunP:
Addressed in `f1063578`.

**Finding 1 — fixed.** Walked `ebe_kappas` down to each scalar component; assert `to_bits()` equality per subject/occasion/component (was per-subject length only). Verified: the test passes with the stricter guard, so the kappa EBEs are bit-identical, not merely close — confirms the thesis at the stronger level you raised (a summed OFV can round-trip to identical bits while a per-kappa ULP diverges; here it doesn't).

**Finding 2 — no change**, agreed: the finite-guard runs first, so the bit-compare never sees NaN/±inf.

**Finding 3 — my diff is clippy-clean.** `cargo clippy --lib --tests --features ci` flags nothing on `iov_integration.rs`. It does surface 4 pre-existing `approximate_constant` errors on the `3.14` test fixtures in `src/sens/two_cpt_explicit.rs` (on `main` since `add7f17b`); CI's Clippy job omits `--tests`, so it never lints those `#[cfg(test)]` modules. Out of scope for this test-only PR — filed separately.

Issue #419, opened 2026-06-19
Reconcile bioavailability F with rate-defined infusions (RATE>0, RATE=-1): NONMEM scales duration, ferx scales rate


TeunP:
## Summary

ferx applies bioavailability `F` to an infusion by scaling the **rate**
(`PkParams::bioavailable_rate = F·rate`, `src/types.rs`) over the data/modeled
duration (`src/pk/event_driven.rs` infuses over `dose.duration`, unscaled). This
matches NONMEM for a **duration-defined** infusion (`RATE=-2` → `D{cmt}`: rate
`= F·AMT/D`, duration `D` fixed), but **diverges** for a **rate-defined**
infusion — both `RATE>0` (data) and `RATE=-1` → `R{cmt}` (#324) — where NONMEM
keeps the rate at `R` and scales the **duration** to `F·AMT/R`.

Total exposure (`F·AMT`) agrees either way; only the infusion **shape** differs,
and only when `F ≠ 1`.

## Why this surfaced

#324 added modeled infusion rate (`RATE=-1` → `R{cmt}`, #418). By design
`resolve_rate` makes `RATE=-1 R=r` resolve to exactly the explicit `RATE=r`
infusion, so it inherits ferx's rate-scaling. At `F=1` (the usual case for
IV/SC infusions, and the NONMEM-anchored case) there is no divergence. The
`F≠1` behaviour is currently **pinned and documented**, not silently shipped:

- `tests/modeled_rate.rs::modeled_rate_under_f_matches_duration_equivalent`
  characterises it (`RATE=-1 R=20` + F ≡ `RATE=-2 D=5` + F in ferx);
- `docs/src/data-format.md`, `faq.md`, and `model-file/ode-models.md` note the
  caveat.

## Proposed fix

Make `F` application mode-aware: for **rate-defined** infusions (`RATE>0` and
`RATE=-1`), keep the rate and scale the duration to `F·AMT/R`; leave `RATE=-2`
(duration-defined) scaling the rate as today.

Scope notes:
- This **changes existing `RATE>0` + F behaviour**, so it needs its own NONMEM
  anchor (an ADVAN with `$PK F1<1` and a `RATE>0` infusion) and a CHANGELOG
  **Changed** entry.
- It touches the shared infusion-application path (`event_driven.rs`, the SS
  equilibration, the superposition path, and the AD twin), so it must keep the
  `RATE=-1 ≡ explicit RATE=r` invariant intact (both move together).
[...]

PR #919, opened 2026-07-23
refactor(types): make DoseEvent::cmt private to compiler-enforce the CMT=0 convention [closes #912]

### LINKED ISSUE #912: refactor(types): make DoseEvent::cmt private to compiler-enforce the CMT=0→1 convention (accessor landed in #913)

TeunP:
## Why
`CMT` is 1-based and `CMT=0` is NONMEM's default dose compartment (compartment 1). Since #899 every engine agrees — but the agreement was held together by ~a dozen independent `saturating_sub(1)` call sites, not by construction. Nothing stopped the next dose-application site from open-coding `cmt - 1` (underflows on `CMT=0`) or `if cmt >= 1` (drops `CMT=0`), and the default failure is silent. #899's review found **four** such misses in one focused pass; #913 added `cmt_idx()`/`cmt_1based()` as the named home (step 1) but left the raw field `pub`, so the mistake stayed writable.

This is step 2 of #912: make the field private so the raw value cannot be decremented ad hoc — a bare `.cmt - 1` no longer compiles outside `types.rs`.

## What changed
- `DoseEvent::cmt` is now **private**. Three read paths:
  - `cmt_idx()` — 0-based state index (`CMT=0`/`1` → 0). Existed (#913).
  - `cmt_1based()` — 1-based compartment (`CMT=0` → 1). Existed (#913).
  - `cmt_raw()` — **new**; the literal authored value. The escape hatch for the genuine raw readers: `CMT=0` detection, diagnostic echo, and callees that do their own `0`→1 mapping (`DoseAttrMap::indexed_slot` does `cmt.max(1)` internally, so passing the raw value is bit-preserving).
- `synthetic_cycle_dose()` — **new** `pub(crate)` helper. Replaces the two `..dose.clone()` struct-update spreads in `event_driven.rs`; a private field forbids `{ ..base }` outside the module (E0451).
- **93 read sites** across 16 files migrated, compiler-driven (making the field private lights up exactly the `DoseEvent.cmt` reads via E0616 — the correct filter, since grep can't distinguish `d.cmt` from `f.cmt` on `InputRateForcing`). Purely mechanical, behaviour-preserving:
  - `.cmt - 1` / `.saturating_sub(1)` → `cmt_idx()` (34)
  - `if cmt==0 {1} else {cmt}` → `cmt_1based()` (1)
  - every other raw read → `cmt_raw()` (58, returns the field verbatim)

## Alternatives considered
- **Newtype `Cmt(usize)`** with `.idx()`/`.one_based()`/`.raw()`: keeps the field `pub` (struct-literal construction still works) but the inner value stays decrementable and every read site changes anyway. The issue prescribes private-field + accessors.
- **Normalise `cmt`→1 in the datareader / `resolve_subject_doses`**: deliberately rejected in #912 — it would destroy the information needed to reject a `CMT=0` *infusion* (which has no default compartment), and `resolve_subject_doses` isn't universal (`sens/ode_provider` builds its own dual walk), so it would create a new path-dependent split.

## Cross-repo dependency

ferx-r never reads or constructs `DoseEvent.cmt` (its only `.cmt` is `r.cmt` on an output row). No ferx-r change required.

## Breaking changes
- [x] Public Rust API (`types.rs`) — `DoseEvent::cmt` field is no longer `pub`.

<details>

Read the compartment via `cmt_idx()` (0-based state index), `cmt_1based()` (1-based, `CMT=0`→1), or `cmt_raw()` (literal authored value). Construct via `DoseEvent::new` / `DoseEvent::modeled` (unchanged). No in-tree or ferx-r consumer touched the field directly.
</details>

## Numerical validation
- [x] Not applicable — behaviour-preserving refactor. Every migrated read returns the identical value it did before (`cmt_raw()` ≡ the field; `cmt_idx()` ≡ the prior `saturating_sub(1)`; `cmt_1based()` ≡ the prior `if cmt==0 {1} else {cmt}`). The existing suite is the oracle (#912's stated validation).

```
lib tests (--features ci):          2698 passed; 0 failed; 18 ignored
analytical_ode_equivalence:            7 passed; 0 failed
nonmem_dose_compartment_anchor:       18 passed; 0 failed
dose_compartment_routing:             12 passed; 0 failed
compiles: --features ci  AND  --features "survival markov nn slow-tests"
```

## Performance
- [x] No significant impact expected — the accessors are `#[inline]` and identical to the code they replace.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes) — extended `dose_cmt_accessors_*` with `cmt_raw()`, added `synthetic_cycle_dose_preserves_shape_and_clears_steady_state`.
- Behaviour-preserving migration; the #899 `CMT=0` suite (value + gradient + end-to-end on both engines) is the regression backstop.

## Docs
- [x] `CHANGELOG.md` — N/A (internal refactor, no user-facing behaviour change).
- [x] No user-visible change.

## Checklist
- [x] `cargo clippy` clean (no new warnings in changed files).
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the `sens/` migration is accessor-only; all `sens_outer_gradient` Dual2-vs-FD parity tests pass.
- [ ] `Cargo.toml` version bump — see Open questions.

## Reviewer hints
- The only interesting file is `types.rs` (field + three accessors + `synthetic_cycle_dose`). Everything else is a mechanical `.cmt` → accessor swap.
- The one judgment per site is which accessor: `cmt_idx` (0-based index), `cmt_1based` (1-based compare), `cmt_raw` (literal / callee maps internally). Applied locally from the surrounding operator; the compiler + the CMT=0 suite verify it.
- `cmt_raw()` at slot-lookup sites (`indexed_slot`, `f_bio`, `lagtime`, `lag_slot`) is deliberate: those callees do `cmt.max(1)` themselves, so passing the raw value is exactly the pre-refactor behaviour.

## Open questions
- Version bump? The field-visibility change is technically breaking to the public Rust API, but no in-tree or ferx-r consumer reads it. Pre-1.0 + ferx-r pins by commit, so I left `Cargo.toml` untouched — happy to bump the minor if you'd rather flag it.

### DIFF
--- a/src/api/output_columns.rs
+++ b/src/api/output_columns.rs
@@ -221,7 +221,7 @@ pub(crate) fn compute_extra_output_columns(
                     match &model.ode_spec {
                         Some(ode) => ode
                             .dose_attr_map
-                            .lagtime(subject.doses[d].cmt, &pk_d.values),
+                            .lagtime(subject.doses[d].cmt_raw(), &pk_d.values),
                         None => pk_d.lagtime(),
                     }
                 })
--- a/src/api/validation.rs
+++ b/src/api/validation.rs
@@ -406,9 +406,11 @@ pub(crate) fn check_absorption_dosing(
     // (`check_absorption_closed_form_support`, which rejects `dose.ss && dose.is_infusion()` with no
     // `II` condition), so the same subject errors on both paths instead of one erroring and the
     // other silently mis-serving.
-    // `cmt_raw()` here (not `cmt_1based()` like the SS-*bolus* gates above) is the one
-    // raw-vs-1based choice in this file where it could matter — a `CMT=0` misses these
-    // 1-based `f.cmt + 1` sets. It is safe: both checks are `is_infusion()`-gated, and a
+    // `cmt_raw()` here (not `cmt_1based()` like the SS-*bolus* gates above): a `CMT=0`
+    // would miss these 1-based `f.cmt + 1` sets, so unlike most raw reads the accessor
+    // choice is load-bearing (the only other place it bites is `has_lagtime_on_cmt` below,
+    // where it is immaterial — that path is analytical-only and the indexed-lag branch is
+    // dead behind `ode_spec.is_some()`). It is safe: both checks are `is_infusion()`-gated, and a
     // meaningful (`AMT>0`) `CMT=0` infusion is already a hard error from
     // `check_dose_compartments` (`E_DOSE_CMT_NOT_INFUSABLE` — `CMT=0` is a default *bolus*
     // compartment, undefined for a zero-order input), so it never reaches a fit regardless
--- a/src/api/validation.rs
+++ b/src/api/validation.rs
@@ -406,6 +406,15 @@ pub(crate) fn check_absorption_dosing(
     // (`check_absorption_closed_form_support`, which rejects `dose.ss && dose.is_infusion()` with no
     // `II` condition), so the same subject errors on both paths instead of one erroring and the
     // other silently mis-serving.
+    // `cmt_raw()` here (not `cmt_1based()` like the SS-*bolus* gates above) is the one
+    // raw-vs-1based choice in this file where it could matter — a `CMT=0` misses these
+    // 1-based `f.cmt + 1` sets. It is safe: both checks are `is_infusion()`-gated, and a
+    // meaningful (`AMT>0`) `CMT=0` infusion is already a hard error from
+    // `check_dose_compartments` (`E_DOSE_CMT_NOT_INFUSABLE` — `CMT=0` is a default *bolus*
+    // compartment, undefined for a zero-order input), so it never reaches a fit regardless
+    // of whether this gate also flags it; an `AMT=0` `CMT=0` infusion is inert. So
+    // `contains(&d.cmt_raw())` cannot miss a reachable case. (Behaviour-preserving: this
+    // was the pre-#912 raw `d.cmt`.)
     let has_ss_infusion = population.subjects.iter().any(|s| {
         s.doses
             .iter()
--- a/src/api/validation.rs
+++ b/src/api/validation.rs
@@ -409,12 +409,12 @@ pub(crate) fn check_absorption_dosing(
     let has_ss_infusion = population.subjects.iter().any(|s| {
         s.doses
             .iter()
-            .any(|d| d.ss && d.is_infusion() && cmts.contains(&d.cmt))
+            .any(|d| d.ss && d.is_infusion() && cmts.contains(&d.cmt_raw()))
     });
     let has_infusion_zero_order = population.subjects.iter().any(|s| {
         s.doses
             .iter()
-            .any(|d| d.is_infusion() && zero_order_cmts.contains(&d.cmt))
+            .any(|d| d.is_infusion() && zero_order_cmts.contains(&d.cmt_raw()))
     });
     if has_ss_infusion {
         diags.push(
@@ -586,10 +586,10 @@ pub(crate) fn check_modeled_dose_rates(
                     (DoseAttr::Rate, "-1", "rate", "E_MODELED_RATE_NO_PARAM", "R")
                 }
             };
-            if !reported.insert((attr, dose.cmt)) {
+            if !reported.insert((attr, dose.cmt_raw())) {
                 continue;
             }
-            let cmt = dose.cmt;
+            let cmt = dose.cmt_raw();
             // A coded-`RATE` dose into compartment `cmt` requires a matching
             // `D{cmt}`/`R{cmt}` parameter so `resolve_rate` has a slot to read — for
             // BOTH engines. `active_dose_attr_map()` returns the engine-correct map
@@ -762,7 +762,7 @@ pub(crate) fn check_dose_compartments(
     for subject in &population.subjects {
         for dose in &subject.doses {
             let is_infusion = dose.is_infusion();
-            let cmt = dose.cmt;
+            let cmt = dose.cmt_raw();
             // **Range rule — every dose, whatever its amount.** Both walks
             // bound-check the compartment *before* the amount is read
             // (`event_driven::…impl`'s `if cmt_idx < n_states { … } else { panic! }`
@@ -1055,7 +1055,8 @@ pub(crate) fn check_absorption_closed_form_support(
                      into a disposition compartment, so the two paths would disagree. Dose the \
                      depot (CMT=1), or use an ODE absorption model for direct central/peripheral \
                      dosing.",
-                    subject.id, dose.cmt
+                    subject.id,
+                    dose.cmt_raw()
                 ));
             }
             if dose.ss && dose.is_infusion() {
@@ -1080,14 +1081,16 @@ pub(crate) fn check_absorption_closed_form_support(
                 // twin can't yet serve: a twin-less form, or a model with an absorption lagtime
                 // (the twin's SS+lagtime pre-arrival seed is still bolus-only — same scope as the
                 // ODE-path `E_ABSORPTION_SS_LAG`). `effective_for` performs the reroute.
-                if model.absorption_ode_equivalent.is_none() || model.has_lagtime_on_cmt(dose.cmt) {
+                if model.absorption_ode_equivalent.is_none()
+                    || model.has_lagtime_on_cmt(dose.cmt_raw())
+                {
                     return Some(format!(
                         "{name} does not support steady-state (SS) doses in this form (subject \
                          {}): SS reroutes to the ODE absorption twin, but this model has no twin \
                          {}. Use a non-SS multiple-dose schedule, or write the model as an ODE \
                          transit()/igd() forcing in [odes].",
                         subject.id,
-                        if model.has_lagtime_on_cmt(dose.cmt) {
+                        if model.has_lagtime_on_cmt(dose.cmt_raw()) {
                             "for the SS + lagtime combination (a follow-up)"
                         } else {
                             "(an unrecognised closed form)"
@@ -2089,7 +2092,7 @@ pub fn check_model_data_warnings(
         // but this warnings pass must stay panic-free if run on such a model rather
         // than hit `resolve_rate`'s slot `.expect`.
         let attr_map = model.active_dose_attr_map();
-        if attr_map.indexed_slot(attr, d.cmt).is_some() {
+        if attr_map.indexed_slot(attr, d.cmt_raw()).is_some() {
             // Initial-estimate warning pass: typical values at t=0.
             let pk = (model.pk_param_fn)(&init_params.theta, &zero_eta, &s.covariates, 0.0);
             d.resolve_rate(attr_map, &pk.values).duration
@@ -2177,15 +2180,15 @@ pub fn check_model_data_warnings(
                     RateMode::ModeledDuration => (DoseAttr::Duration, DoseEvent::DURATION_FLOOR),
                     RateMode::ModeledRate => (DoseAttr::Rate, DoseEvent::RATE_FLOOR),
                 };
-                if let Some(slot) = attr_map.indexed_slot(attr, d.cmt) {
+                if let Some(slot) = attr_map.indexed_slot(attr, d.cmt_raw()) {
                     let pk = pk_at_init.get_or_insert_with(|| {
                         // Initial-estimate warning pass: typical values at t=0.
                         (model.pk_param_fn)(&init_params.theta, &zero_eta, &s.covariates, 0.0)
                     });
                     if pk.values[slot] <= floor {
                         match attr {
-                            DoseAttr::Duration => nonpos_dur.insert(d.cmt),
-                            DoseAttr::Rate => nonpos_rate.insert(d.cmt),
+                            DoseAttr::Duration => nonpos_dur.insert(d.cmt_raw()),
+                            DoseAttr::Rate => nonpos_rate.insert(d.cmt_raw()),
                             DoseAttr::F | DoseAttr::Lag => unreachable!("only D/R reach here"),
                         };
                     }
--- a/src/frem/mod.rs
+++ b/src/frem/mod.rs
@@ -322,7 +322,7 @@ pub fn transform_dataset_for_frem(
                 ".".to_string(),
                 "1".to_string(), // EVID
                 format!("{}", dose.amt),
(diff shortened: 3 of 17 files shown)


TeunP:
## Review — `DoseEvent::cmt` private (#912 step 2)

**Verdict: correct and mergeable.** Ran an adversarial site-by-site pass (no happy-path assumption); could not find a behavior-changing site. The "behavior-preserving" claim holds *by construction*. Minor hardening notes only — none blocking.

### Correctness — verified, not trusted
Accessor defs confirmed on the branch, so each swap class is provably identity:

| Def | Value | Replaces |
|---|---|---|
| `cmt_idx()` | `self.cmt.saturating_sub(1)` | `.saturating_sub(1)` (~34) — **exactly equal** |
| `cmt_1based()` | `self.cmt.max(1)` | `if cmt==0 {1} else {cmt}` (1) — **exactly equal** |
| `cmt_raw()` | `self.cmt` | bare `.cmt` (58) — **verbatim** |

Checks that could have broken it, all clear:

- **The two plain `.cmt - 1` sites** (`gated_infusions`, `ode_provider` saltation ~4541) — `.cmt - 1` ≠ `saturating_sub(1)` at `CMT=0`. Both guarded by an **immediately-preceding** `cmt_raw()==0`/`<1` early-return, so `cmt>=1` is guaranteed at the subtraction → identical, and the underflow can no longer be reintroduced by deleting the guard.
- **Slot-lookup callees fed `cmt_raw()`** (`indexed_slot`, `f_bio`, `lagtime`, `lag_slot`, and `f_bio_slot` which *delegates* to `indexed_slot`) all do `cmt.max(1)` internally → `CMT=0` resolves to compartment 1, same as the old raw pass. `f_bio_slot` was not in the PR's stated max(1) list but is covered transitively.
- **`dose_channel(cmt_raw())`** does `checked_sub(1)` → `CMT=0` → `None` → the "unsupported compartment" panic arm. Behavior-preserving *and* semantically correct: `cmt_1based()` here would have silently rerouted a `CMT=0` infusion to compartment 1 instead of rejecting it. Good discrimination.
- **`input_rate_consumes_cmt(cmt_raw())`** does its own `saturating_sub(1)` — `cmt_raw()` vs `cmt_1based()` collapse to the same index for `CMT=0`; immaterial.
- **`synthetic_cycle_dose()`** is byte-identical to the two `DoseEvent{ time:0, ss:false, ii:0, ..dose.clone() }` spreads it replaces.
- **Completeness is compiler-guaranteed**: private field ⇒ a bare `.cmt` read won't compile. Remaining `DoseEvent.cmt` field reads on the branch = **0**; the only struct-literal constructions were the 2 converted spreads (datareader + all else go through `::new`/`::modeled`). A partial migration cannot compile.
- **ferx-r**: sole `.cmt` is `r.cmt` on an output row, not `DoseEvent`. Cross-repo claim confirmed.

### Findings (minor, non-blocking)

1. **`cmt_raw()` is a `pub` escape hatch that only *partially* enforces.** `d.cmt_raw() - 1` / `d.cmt_raw().saturating_sub(1)` re-open the exact underflow class the PR closes, and the compiler no longer stops them — the enforcement is "harder + greppable," not airtight. Cheap hardening: drop to `pub(crate)` (no external consumer, incl. ferx-r, reads it, unlike the old field) and/or a doc steer toward `cmt_idx()`. 58/93 sites route through the hatch, though nearly all are genuinely-raw (diagnostics, CSV round-trip, set-membership, constructor args).

2. **Set-membership sites are the one load-bearing accessor choice.** `cmts.contains(&d.cmt_raw())` / `zero_order_cmts.contains(&d.cmt_raw())` in `validation.rs` — a `CMT=0` would miss a 1-based set (`cmt_1based()` would not). Verified non-issue: both `is_infusion()`-gated, `CMT=0` infusion rejected upstream, so unreachable. Pre-existing. But unlike the other 57 raw sites the choice actually matters here — a `// CMT=0 unreachable: infusion-gated` note would earn its keep.

3. **Pre-existing asymmetry, out of scope, flag for follow-up:** `propagate.rs::equilibrate_ss_g` still builds its synthetic SS dose via `DoseEvent::new(0.0, amt, cmt_raw(), dose.rate, false, 0.0)` (recomputes `duration = amt/rate`), while the value path uses `synthetic_cycle_dose()` (preserves the #419 F-reshaped `(rate,duration)`). Under `F≠1` on an SS infusion these differ — possible value-vs-gradient mismatch. **Not a regression** (both predate this PR; only `cmt`→`cmt_raw()` changed here), but since #919 touches both, worth a dedicated check.

4. **Test gap (trivial):** `synthetic_cycle_dose_preserves_shape_and_clears_steady_state` covers `rate_mode` only transitively via `is_infusion()`. One `assert_eq!(cyc.rate_mode, dose.rate_mode)` nails it.

### Style / conventions
- Accessor docs are excellent — each explains the `CMT=0` semantics and *why* the field is private.
- Version bump: agree with **no bump**. Pre-1.0, ferx-r pins by commit and doesn't read the field.
- CHANGELOG omission correct (internal refactor).

**Bottom line:** clean structural hardening, correctly executed, existing #899 `CMT=0` suite as oracle. Ship it; optionally fold in finding #1 (`pub(crate)` on `cmt_raw`) before merge and file #3 as a follow-up.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

TeunP:
Thanks for the adversarial pass. Addressed the findings in `f2dbd706`:

- **F1 (`cmt_raw()` visibility)** — dropped to `pub(crate)`. One correction to the premise: there *was* one external-crate reader, the `gen_iiv_anchor` integration test; it now uses `cmt_1based()`, which is the resolved 1-based compartment a NONMEM dose row should carry anyway (DEFDOSE `0`→1). Footgun is now fully crate-internal.
- **F2 (set-membership at `validation.rs` 412/417)** — added the note. Verified the unreachability myself rather than trusting it: a meaningful (`AMT>0`) `CMT=0` infusion is a hard error from `check_dose_compartments` (`E_DOSE_CMT_NOT_INFUSABLE`, 837/876), and an `AMT=0` one is inert — so `contains(&cmt_raw())` can't miss a reachable case.
- **F4 (test)** — `synthetic_cycle_dose` now asserts `rate_mode`/`infusion_def` survive, on a *modeled* dose so it's non-vacuous.
- **F3 (SS-infusion F-window asymmetry)** — confirmed real and pre-existing; filed as #920. Key detail: the existing `F≠1` SS-infusion parity test covers the **ODE** twin (`equilibrate_ss_state_g`), not the analytic `equilibrate_ss_g`, so the analytic path is genuinely unverified under `F≠1`. Left out of this PR as you suggested.

Version bump: kept none, per your agreement.

TeunP:
Folded #920 in (`cc796f1f`) — it's a non-bug. A rate-defined SS infusion under `F ≠ 1` declines to FD before `equilibrate_ss_g` runs (the `has_bioavailability() && has_rate_defined_infusion()` gate doesn't special-case SS), so its `DoseEvent::new` synthetic window is never used with a reshaped window. Added `ss_rate_defined_infusion_under_f_declines_to_fd` to lock it in (bolus stays analytic; F-infusion declines both providers; F-free twin stays analytic — non-vacuous). Closing #920.

TeunP:
## Review (xhigh, two independent passes)

Reviewed at xhigh with an own-code adversarial pass **and** an independent fresh-eyes reviewer (which compiled under 4 feature sets and ran the new tests). **Verdict: correct, behaviour-preserving, mergeable — no high or medium issue.**

**Why it's safe by construction.** The three accessors agree for `CMT ≥ 1` and differ *only* at `CMT=0`, so the only regression path is a wrong accessor at a site actually reached with `CMT=0`. The compiler kills the other two failure modes (a missed `DoseEvent.cmt` → E0616; a wrong-type swap → no such method). Every load-bearing site was traced through its surrounding code:

| Class | Result |
|---|---|
| 35 `cmt_idx()` ≡ `saturating_sub(1)` | identical; the 2 bare `.cmt-1` sites are guarded (`cmt_raw()==0`/`<1` early-return), so `CMT=0` never reaches the subtraction |
| 4 `cmt_1based()` ≡ `max(1)` / `if 0{1}else` | identical |
| 69 `cmt_raw()` = verbatim field | callees (`indexed_slot`/`f_bio`/`lagtime`/`lag_slot`/`f_bio_slot`/`input_rate_consumes_cmt`/`dose_channel`) map `0→1` internally; `==0`/`<1`/`>=1` and set-membership gates are literal-preserving |
| validation `contains(&cmt_raw())` (412/417) | safe — a meaningful (`AMT>0`) `CMT=0` infusion already hard-errors upstream (`E_DOSE_CMT_NOT_INFUSABLE`); `AMT=0` is inert |
| `synthetic_cycle_dose()` | byte-identical to both replaced `{ ..dose.clone() }` spreads |
| new tests | non-vacuous (reshape + modeled-dose discriminators; bolus control + F-free discriminator) |

**Findings (both LOW) + resolution:**
1. **validation.rs gate comment overstated** ("the one raw-vs-1based choice… where it could matter") — `has_lagtime_on_cmt` at ~1090 is another raw-pass, merely immaterial (analytical-only, indexed-lag branch dead behind `ode_spec.is_some()`). **Fixed** in `199d0dd9` (comment-only).
2. **frem/mod.rs writes `cmt_raw()` while gen_iiv_anchor writes `cmt_1based()`** — **correct as-is, no change.** frem is a faithful passthrough of the source dose record (it writes `time`/`amt`/`cmt` raw; only RATE is reconstructed for NONMEM's `-1`/`-2` sentinels), so `cmt_raw()` is the right choice — resolving `CMT=0→1` there would break that faithfulness. gen_iiv_anchor differs because `cmt_raw()` is `pub(crate)`; it uses `cmt_1based()`, and the warfarin data is all `CMT=1`, so its output is unchanged.

**Conventions / perf / security:** tier-1 tests in the sibling `#[path]` files ✓; no CHANGELOG (internal refactor) ✓; accessors `#[inline]`, zero perf impact ✓; no security surface.

**Local verification:** full suite (lib + all integration targets, `--features "survival markov nn"`) — **3592 passed, 0 failed, 121 ignored**.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

Issue #351, opened 2026-06-15
fix(pk): event-driven analytical path silently drops infusion input on oral models


TeunP:
## Summary

The event-driven analytical PK path silently drops infusion (zero-order, `RATE>0`) input on **oral** models, producing ~0 concentration instead of the correct curve.

## Details

The oral propagators `propagate_one/two/three_cpt_oral` in `src/pk/event_driven.rs` take no input-rate argument, so a depot-bypassing infusion into the central compartment (`cmt 2`) has its rate — `rate_central`, correctly computed by the infusion-routing match for `(OneCptOral, 2) => rate_central` — silently discarded by the propagator.

**Impact:** any oral model receiving an infusion dose returns ~0 from the event-driven path, which is used for subjects with time-varying covariates, EVID=3/4 resets, and all IOV models. The superposition path (no-covariate subjects) handles it correctly, so the same model gives drastically different predictions for no-TV versus TV/IOV subjects — silently, with no error.

Surfaced during the #327 (bioavailability on IV/infusion) follow-up; distinct root cause.

## Scope note

Infusion into an oral **depot** (`cmt 1`) is a separate, ambiguous case (the superposition path bypasses to central; NONMEM does zero-order-into-depot) and remains an explicit error — out of scope here.

## Fix

#350 — threads the central input rate into the oral propagators via linear superposition (reuses the tested IV propagators from a zero initial state). The autodiff twin needs no change: the inner optimizer already routes oral+infusion subjects to FD gradients.

PR #350, opened 2026-06-15
fix(pk): apply infusion input to oral models on the event-driven path [closes #351]

TeunP:
## Why

The event-driven oral propagators (`propagate_one/two/three_cpt_oral`) took no input-rate argument, so a depot-bypassing infusion (`RATE>0` into the central compartment, `cmt 2`) had its rate **silently dropped** — giving ~0 concentration for any oral + infusion subject routed through the event-driven path (time-varying covariates, EVID=3/4 resets, or IOV), while no-covariate subjects (superposition path) got the correct curve.

The infusion-routing match already computed the central input rate for oral models (`(OneCptOral, 2) => rate_central`), but the oral propagators discarded it. Found during the #327 (bioavailability on IV/infusion) follow-up; distinct root cause.

## What changed

Thread a `rate_central` argument into the three oral propagators and add the constant central input by **linear superposition** — reuse the (tested) IV propagators from a zero initial state to isolate the pure forced response, then add it to the central (+ peripheral) compartments. The depot is untouched (depot-bypass convention, matching the superposition path).

The autodiff twin needs no change: the inner optimizer already routes oral + infusion subjects to FD gradients (the AD oral kernels are bolus-only), and that FD path now derives from the corrected objective. The guard's comment is updated to reflect it.

## Alternatives considered

- **Rewrite the AD oral propagators to carry the infusion.** Rejected: needs a cmt-routing remap (oral `cmt 2` is central, but the IV propagator treats `cmt 2` as peripheral), and it's unverifiable locally (`autodiff`-gated, no Enzyme).
- **Add an FD-fallback guard for oral + infusion in the AD path.** The inner optimizer already has exactly this guard, so no AD change was needed.
- **A bespoke oral-infusion closed form per propagator.** Rejected in favour of reusing the tested IV propagators (linear superposition).

## Cross-repo dependency

| Repo | PR | Status | Must merge |
|------|----|--------|------------|
| ferx (R) | — | not needed | — |

## Breaking changes
- [x] None

Oral + infusion models now produce correct (non-zero) event-driven predictions; no API or format change.

## Numerical validation
- [x] Compared against: NONMEM (closed form) and the superposition reference
- New test `event_driven_applies_oral_infusion_into_central`: for all three oral families, `event_driven_predictions == predict_concentration` (the superposition reference — a depot-bypassing IV-into-central infusion, i.e. the NONMEM result) at `F=1`, where it was ~0 before; and predictions scale linearly with `F`. Self-sufficient (independent of #327).

[...]

Issue #327, opened 2026-06-14
refactor(pk): retire (or guard) dead predict_iov_option_a — it silently drops F for IV/infusion if reactivated


TeunP:
## Background

`predict_iov_option_a` ([`src/pk/mod.rs:390`](https://github.com/FeRx-NLME/ferx-core/blob/main/src/pk/mod.rs)) is the legacy "Option-A" per-occasion analytical-superposition IOV path. Its own docstring says it is "retained only as a fallback for models that support neither the ODE nor the event-driven analytical path."

It builds infusion/IV-bolus concentrations via `one_cpt_infusion(dose, tau, cl, v)` / `one_cpt_iv_bolus(...)` (and the `*_ss` variants) — **none of which take `f_bio`**. So along this path, **bioavailability `F` is silently dropped for IV bolus and infusion dosing** (only the oral arms pass `p.f_bio()`).

## Why it is not a live bug today

The dispatch in `predict_iov` is:

```
ODE model                       -> ode_predictions_event_driven
supports_event_driven(pk_model) -> event_driven_predictions   <-- all 6 PkModel variants
else                            -> predict_iov_option_a        <-- unreachable
```

`supports_event_driven` returns `true` for **all six** `PkModel` variants (`OneCptIv/Oral`, `TwoCptIv/Oral`, `ThreeCptIv/Oral`), so the `else` branch is currently unreachable. The active event-driven path applies `F` correctly to both dose forms (bolus amount `f_bio()*amt`; infusion `f_bio*rate` with duration preserved — matches NONMEM), verified in `src/pk/event_driven.rs::{f_bioavailability_scales_bolus_and_infusion_linearly, lagtime_shifts_bolus_and_infusion_in_time}`.

## Risk / trigger

This is a latent landmine adjacent to active roadmap work. If a **7th** `PkModel` is added — e.g. via the built-in absorption models (#322) — and it is *not* listed in `supports_event_driven`, `predict_iov_option_a` becomes live and silently produces `F`× -wrong concentrations for any IV/infusion dosing, with no error. Silent-wrong is exactly the failure mode #324 just removed from the datareader.

## Proposed fix (pick one)

[...]

PR #921, opened 2026-07-23
test(pk): NONMEM .ctl anchor for oral central-infusion (event-driven) [closes #376]

### LINKED ISSUE #376: test(pk): dedicated NONMEM .ctl anchor for oral central-infusion (event-driven), currently anchored transitively
## Context

Follow-up from the review of #350 / #351.

PR #350 fixed the event-driven analytical path to apply a central (depot-bypass)
infusion on oral models. Its numerical validation is **transitive**: the new
tests assert `event_driven_predictions == predict_concentration` (the
superposition reference), and the superposition reference for a depot-bypass
infusion is the IV-into-central closed form, which is separately NONMEM-anchored.
Correctness therefore chains to NONMEM through the superposition path rather than
via a direct comparison.

## Why

CLAUDE.md's testing policy asks every behaviour that produces numerical results
to be validated against equivalent NONMEM output. The transitive chain is sound,
but a dedicated `.ctl`-backed anchor — an oral model (e.g. 1-cpt `ADVAN2`) with a
`cmt 2` infusion (`RATE>0`, depot bypass) and an estimated `F` — compared
directly to NONMEM `PRED/IPRED` would satisfy the rule more strictly and guard
the F-on-infusion interaction (#327/#349) on a real model.

## Scope
- Add a small oral + central-infusion dataset and `.ctl`, run NONMEM, and add the
  comparison to the test suite or the relevant docs page (as done for other
  estimators/error models).
- Cover F≠1 (the cross-path consistency the event-driven and superposition paths
  now share).

## References
- PR #350 / #351
- Related anchor-rigor issue: #284

TeunP:
Closes #376.

## Why

Follow-up from the review of #350 / #351 (issue #376). PR #350 fixed the
event-driven analytical path to apply a central (depot-bypass) infusion on oral
models, but its numerical validation is **transitive**: the #350 test only
asserts `event_driven_predictions == predict_concentration` (the superposition
reference), which in turn chains to NONMEM through the IV-into-central closed
form.

The gap this leaves is real and specific:

- `tests/nonmem_dose_compartment_anchor.rs` / the `dose_cmt_*` anchors cover oral
  infusion into the **peripheral** compartment only (ADVAN4 `CMT=3`, ADVAN12
  `CMT=3/4`) — **never the central `CMT=2`** of an oral model, which is exactly
  the #350 depot-bypass case.
- F-on-infusion (the #419 duration-scaling) was NONMEM-anchored only for a
  **cmt1 IV** infusion (`bioavailability_infusion.ctl`, ADVAN1) — never for F on
  an oral model's central infusion (the #327/#349 interaction #376 asks to
  guard).

## What changed

Test-only. Adds a dedicated `.ctl`-backed anchor for an oral model's central
(`CMT=2`) infusion, directly against NONMEM `PRED`, covering F=1 **and** F≠1:

- **Control streams + data** (NONMEM 7.6.0, `$ESTIMATION MAXEVAL=0`,
  `$TABLE ... FORMAT=,1PE17.10`): `nonmem_anchor/oral_central_inf_advan{2,4}_f{1,06}.ctl`
  (1-cpt ADVAN2 and 2-cpt ADVAN4, `RATE=25` into `CMT=2`, `F=1` and `F2=0.6`),
  their shared datasets, and verbatim outputs under
  `nonmem_anchor/results/oral_central_inf_*.{lst,tab}`.
- **`tests/oral_central_infusion_nonmem_anchor.rs`** (5 tests). Like the sibling
  dose-compartment anchor, every case asserts the **event-driven walk** (forced
  by a varying `WT` column the model never reads) *and* the **default
  superposition dispatch** against NONMEM `PRED`, and against each other — so a
  path that dropped the central rate (the #350 bug) or applied `F` on only one
  path fails here where the F=1-only self-check could not. The 1-cpt cases
  additionally assert ferx against the exact ADVAN1 **duration-scaled closed
  form** — a license-free oracle that reproduces the committed NONMEM digits
  independently of NONMEM's solver.

The obs grid (`t = 1, 3, 6, 10, 16 h`) straddles both infusion ends (`4 h` at
`F=1`, `2.4 h` at `F=0.6`), so the same times sit in the rising phase for one `F`
and the decay phase for the other — a rate/duration mix-up on one side cannot
pass both.

## Alternatives considered

- **Reuse the existing IV/peripheral anchors via the reduction** (oral `CMT=2`
  infusion with an empty depot ≡ 1-cpt IV `cmt1` infusion). Rejected as the
  primary evidence — #376 explicitly asks for a *dedicated* ADVAN2 run rather
  than trusting the reduction. NONMEM was run fresh; it *confirms* the reduction
  (`t=1` PRED `0.47581 = 5·(1−e^{−0.1})`, plateau `rate/(V·ke)=5`) rather than it
  being assumed.
- **Extend `nonmem_dose_compartment_anchor.rs`.** That harness is bit-pinned and
  carries no `F`; a separate file matching the repo's `*_nonmem_anchor.rs`
  convention keeps the F-threading isolated.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation
- [x] Compared against: NONMEM 7.6.0 (control streams + verbatim `.lst`/`.tab`
  committed under `nonmem_anchor/`)
- [x] Reference values (NONMEM `PRED`, `FORMAT=,1PE17.10`):

```
# ADVAN2 (1-cpt oral) CMT=2 infusion, obs t = 1, 3, 6, 10, 16 h
F=1  : 0.47581290982  1.2959088966  1.3495955849  0.90466097461  0.49648846959
F2=.6: 0.47581290982  1.0047315645  0.74432344989 0.49893492919  0.27382129479
# ADVAN4 (2-cpt oral) CMT=2 infusion
F=1  : 0.46222223658  1.1951957279  1.0956488990  0.62303307734  0.31054165440
F2=.6: 0.46222223658  0.90910738405 0.57799283210 0.33374887600  0.17144103343
```

ferx matches all four on both analytical paths at `rel < 1e-9`. The F=1 and
F=0.6 curves share their `t=1` value (rate held, duration scaled — #419), and
the 1-cpt values reproduce the duration-scaled ADVAN1 closed form. The anchor
was mutation-tested (perturbing one reference constant fails both the NONMEM
assertion and the independent closed-form oracle).

## Performance
- [x] No significant impact expected (test-only)

## Tests
- [x] Regression test added — `tests/oral_central_infusion_nonmem_anchor.rs`
  (5 tests, all pass). Fails without the #350 fix (the event-driven walk
  returned ~0 for an oral central infusion).

## Docs
- [x] `CHANGELOG.md` — N/A (test-only; no user-visible change, per CLAUDE.md)
- [x] No user-visible change; `nonmem_anchor/README.md` anchor catalogue row added

## Reviewer hints

- The walk-vs-superposition split is exercised through the public `predict()`
  via the WT-covariate trick, identical to `nonmem_dose_compartment_anchor.rs`.
- ferx analytical models carry a single global `PK_IDX_F` applied to whichever
  compartment the dose targets (per-compartment `F{cmt}` is ODE-only,
  `types.rs:552`), so the NONMEM match puts the bioavailability on the dosed
  compartment (`F2`, not `F1` — `F1` is irrelevant with an empty depot).

## Open questions

None.


TeunP:
## Self-review (xhigh, adversarial — no happy paths)

Reviewed with fresh adversarial eyes; verified the load-bearing claims by reading production code rather than trusting the green test.

### Crux claims — verified, not assumed

| # | Claim under attack | Verdict | Evidence |
|---|--------------------|---------|----------|
| A | NONMEM constants transcribed correctly | ✅ | 20/20 byte-match the committed `.tab` |
| B | `[event-driven walk]` genuinely runs the walk (not superposition twice) | ✅ | `read_nonmem_csv(_, None, None)` auto-detects `WT` as a covariate (`datareader.rs:737`); `WT` varies → `is_tv` (`:1466`) → `obs_covariates` populated → `Subject::has_tv_covariates()` true → `compute_predictions_with_tv` (`pk/mod.rs:2126`) routes analytical + has_tv to `event_driven_predictions` (the #350 code). `plain_preds` (no `WT`) → superposition. Genuinely two distinct paths. |
| C | Analytical path, not a silent ODE fallback | ✅ | No `[odes]` block → `ode_spec` is `None` → analytical branches only |
| D | `F` duration-scales on cmt2 (not rate) | ✅ | NONMEM confirms; obs grid straddles both `F`-ends, so a rate/duration mix-up cannot pass. `F=1` and `F=0.6` share their `t=1` value (rate held) |
| E | Depot-bypass reduction (oral cmt2 ≡ 1-cpt IV cmt1) | ✅ | NONMEM run fresh, not assumed; `t=1` PRED = `5·(1−e^{−0.1})` exactly (plateau `rate/(V·ke)=5`) |

### Finding + fix

- **rigor — tolerance slack.** `NONMEM_TOL` was `1e-9` with no measured basis. The **measured** ferx-vs-NONMEM gap is ~2e-11 (max 1.98e-11 on ADVAN4, the last printed digit of the 11-sig-fig `$TABLE` reference). Probing: `1e-11` fails, `1e-10` passes all five. **Fixed** → `NONMEM_TOL = 1e-10` (single source of truth, reused by the closed-form-oracle check), removing the ~50× slack a subtle regression could hide in, with ~5× headroom over the floor to avoid last-digit flaking. Commit `9d3922dd`.

### Known limitation (not a blocker)

Walk == superposition for a central infusion, so if the `WT` → walk routing ever silently broke, `walk_preds` would fall back to superposition and the test would still pass. This is the same accepted limitation as the sibling `nonmem_dose_compartment_anchor.rs` — inherent to the `WT`-trick (no cmt2-infusion scenario has the two paths legitimately disagree). Mitigation is the code-read proof (B), not test self-verification.

### Other

- Mirrors the sibling anchor's structure and the repo's `*_nonmem_anchor.rs` convention. `cargo fmt` / `clippy` clean. Test-only → no CHANGELOG (correct per CLAUDE.md). No perf/security surface.
- Mutation-verified: a +2.1e-5 perturbation of one reference constant fails two independent assertions (the NONMEM anchor and the closed-form oracle).

**Verdict: sound.** No correctness defects; one rigor slack found and fixed.

PR #922, opened 2026-07-23
perf(ode): solve the ODE bolus/infusion steady state exactly, and warn when it can't [closes #914]

### LINKED ISSUE #914: perf(ode): solve the ODE bolus/infusion steady state exactly, and warn when it can't

TeunP:
## Why

`#908` made the **analytical** event-driven walk solve the periodic steady state exactly as `u_ss = (I − M)⁻¹·b`. The **ODE** engine's ordinary bolus/infusion SS still expanded a truncated pulse train (up to `SS_EQUILIBRATION_CYCLES = 50`), leaving a residual `≈ exp(−50·λ_slow·II)` — negligible for typical PK but tens of percent for a slow disposition, and **silently**. Two gaps, both in `equilibrate_ss_state` and its dual twin `equilibrate_ss_state_g`:

1. **Truncation** — the bolus/infusion branch never used the exact fixed point the analytical walk already had.
2. **Silence** — `note_ss_nonconvergence_if_capped` was wired only into the ODE input-rate branch; a capped ordinary bolus/infusion equilibration emitted no warning.

Closes #914.

## What changed

- **`equilibrate_ss_state` (`src/ode/predictions.rs`)** — a linear disposition now routes its bolus and infusion branches through `crate::dosing::periodic_ss_fixed_point_g` (the same exact solve #908/#835 use), returning `(I − M)⁻¹·b` directly. `advance_forced` integrates one cycle with the dose (bolus pulse, or active-infusion + quiet window); `advance_unforced` integrates one `II` of disposition alone (the propagator `M` — the infusion's active/quiet windows share the homogeneous propagator, so a full-`II` decay reconstructs it and the window length enters only through `b`). The exact solve uses the tightened `ss_equilibration_opts` tolerance and passes it to the linearity self-check.
- A genuinely nonlinear RHS (Michaelis–Menten, …) fails the affine self-check → the capped pulse train, now with `note_ss_nonconvergence_if_capped` wired in (**gap 2**).
- **`equilibrate_ss_state_g` (`src/sens/ode_provider.rs`)** — the dual twin mirrors this so FOCE/FOCEI differentiates the exact steady state. The infusion `advance_forced` carries the moving active/quiet boundary's `inf_window` jet through the same `inject_rate_saltation` the fallback loop uses (`dtinf = inf_window − t_inf_val`, zero value part → the value map stays linear while the window sensitivity rides `b`). The nonlinear fallback loops are unchanged; the value walk owns the #867 warning (the twin must not double-count it).

## Alternatives considered

Routing the nonlinear bolus fallback through Anderson acceleration (as the input-rate branch does for a nonlinear disposition that admits SS) was out of scope per the issue — the ordinary bolus/infusion path keeps the capped train + warning for a nonlinear RHS. The exact solve covers the (overwhelmingly common) linear case.

## Breaking changes
- [x] None — `equilibrate_ss_state`/`equilibrate_ss_state_g` are private; `periodic_ss_fixed_point_g` was already `pub(crate)`. No public API surface. ferx-r unaffected (no pin bump).

## Numerical validation
- **Oracle** — the analytical closed forms, exactly as #908. A linear ODE model's exact SS now matches its analytical twin to integration tolerance (tightened `ss_peripheral_{infusion,bolus}_matches_the_ode_twin` from `1e-6` → `1e-8`; the 1-cpt trough matches the hand-derived geometric-series SS `F·amt·e^{−ke·II}/(1 − e^{−ke·II})` in `ss_linear_disposition_uses_exact_fixed_point`).
- **`Dual2`-vs-FD parity** — `ode_provider_ss_linear_bolus_uses_exact_solve` (bolus) and `ode_provider_ss_rate_under_f_uses_exact_solve` (rate-defined infusion under `F`, exercising the window-jet saltation) check value + `∂/∂η` + `∂/∂θ` + both Hessian blocks against production + FD.
- **Nonlinear fallback preserved** — `ss_nonlinear_bolus_with_steady_state_uses_fallback` (MM disposition that admits SS → correct trough via the fallback, no warning), `ss_nonlinear_over_capacity_bolus_caps_and_warns` (mean input > Vmax → 50-cycle cap + #867 warning), and `ode_provider_ss_nonlinear_fallback_early_stop_matches_full_budget` (the dual early stop is now nonlinear-only; gradient-faithful vs a forced-full-budget run).
- Existing NONMEM SS anchors (`ss_lagtime_nonmem`, `nonmem_dose_compartment_anchor`, `oral_peripheral_infusion`) stay as the cross-tool check.

## Performance
- [x] Faster — the linear case replaces ~50 RK45 one-cycle integrations with `~n+2` integrations + one `n×n` solve.

## Tests
- [x] Unit tests added (`cargo test --lib` passes) — see Numerical validation.
- [x] Regression: the truncation was silent; the new tests fail against the old truncated train.

## Docs
- [x] `docs/model-file/steady-state.qmd` — the "ODE path" section, the `N=50` section, the path/cost tables, and the analytic-gradient section now describe the exact solve + nonlinear fallback; the `#914` caveat is removed.
- [x] `CHANGELOG.md` `[Unreleased]` → Fixed.

## Reviewer hints
The subtle half is the dual infusion `advance_forced`: the `inject_rate_saltation` adds the velocity **jump** (`r·e·dtinf`), which keeps the one-cycle map exactly affine in `u0` with a window-independent `M = e^{A·II}` — so the window sensitivity lands entirely in `b` and `periodic_ss_fixed_point_g` assembles the exact `∂u_ss/∂(θ,η)`. Confirmed by the FD-parity tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,15 @@ section of the SDLC for the versioning policy).
   for any external code that constructs or exhaustively destructures these variants.
 
 ### Fixed
+- **Steady-state (`SS=1`) dosing on `[odes]` models is now exact for a linear disposition, and
+  warns instead of silently truncating on a nonlinear one** (#914). An ordinary ODE bolus or
+  infusion SS dose previously equilibrated by expanding a pulse train capped at 50 cycles, which
+  under-reported the steady state by tens of percent for a slow disposition (the same truncation
+  #908 removed from the analytical engine) — silently. A linear disposition now solves the exact
+  periodic fixed point `(I − M)⁻¹·b` directly (value and analytic FOCE/FOCEI gradient), so slow PK
+  is exact rather than low; a genuinely nonlinear RHS (e.g. Michaelis–Menten) still falls back to
+  the capped iteration but now surfaces a non-convergence warning when the cap is reached without
+  settling. Also faster: the linear case replaces ~50 RK45 cycles with a handful.
 - **`predict()` on a CTMM (`[markov_model]`) model now fails loud instead of silently returning
   no rows** (#759). The equivalent `simulate()` guard already existed and its message claimed to
   cover `predict()`, but it only ran on the simulate path. State-occupancy prediction `π(t)` is
--- a/docs/model-file/steady-state.qmd
+++ b/docs/model-file/steady-state.qmd
@@ -41,7 +41,7 @@ Every prediction path in ferx-core honours `SS=1`:
 | ------------------------------------------------- | ------------------------------------------------------------------------- | -------------------------------------------------- |
 | Analytical (1-/2-/3-cpt, no TV covariates)        | `predict_concentration` in `src/pk/mod.rs`                                | Closed-form geometric-series                       |
 | Analytical (1-/2-/3-cpt, time-varying covariates) | `event_driven_predictions_with_schedule` in `src/pk/event_driven.rs`      | Exact linear fixed point `(I − M)⁻¹·b`             |
-| ODE (`[odes]`-block models)                       | `ode_predictions` / `ode_predictions_event_driven` in `src/ode/predictions.rs` | Numerical pulse expansion via the RK45 solver |
+| ODE (`[odes]`-block models)                       | `ode_predictions` / `ode_predictions_event_driven` in `src/ode/predictions.rs` | Exact linear fixed point `(I − M)⁻¹·b`; nonlinear RHS falls back to numerical pulse expansion |
 
 Both kinds of paths start from the same underlying identity — the
 choice between them is a question of whether the geometric series has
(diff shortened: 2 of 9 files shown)


TeunP:
## Adversarial review (xhigh) — exact ODE steady-state solve

**Method.** Three independent reviewers — one per surface (f64 `equilibrate_ss_state`, the dual-sensitivity `equilibrate_ss_state_g`, and test adequacy) — each instructed to hunt for real correctness bugs with no happy-path rationalization, plus a line-level read. Every finding was verified independently rather than accepted on a reviewer's word; passing CI is **not** treated as evidence of correctness.

**What the PR does.** A linear ODE disposition's `SS=1` bolus/infusion now solves the exact periodic fixed point `u_ss = (I − M)⁻¹·b` (via the shared `dosing::periodic_ss_fixed_point_g`) instead of a 50-cycle pulse train that silently truncated tens of percent low on a slow disposition — the same fix #908 made on the analytical engine. The dual twin mirrors it so FOCE/FOCEI differentiates the exact steady state. A genuinely nonlinear RHS still falls back to the capped train, now with the #867 non-convergence warning wired into the ordinary bolus/infusion path (previously input-rate-only).

### Verdict on the production code: correct

- **No silent wrong value is reachable.** `periodic_ss_fixed_point_g` validates the *computed* `u_ss` against a fresh `advance_forced` call (a two-cycle fixed-point check), so even a wrong M/b decomposition cannot self-certify — it fails the check and falls back to the iteration. A genuinely nonlinear map's residual is `O(scale)` (~1e8·reltol), eight orders above the `vtol ≈ 256·reltol·scale` bound; a linear map's residual carries no `1/(1−ρ)` amplification, so heavy accumulation (ρ→1) does not falsely reject it.
- **Value = gradient.** `solve_ode_g` runs its adaptive step accept/reject on the value parts only, so the dual walk takes byte-identical steps to the f64 walk; the saltation adds exactly zero to the value (`dtinf.val() = 0`). FOCEI differentiates the same trough it predicts.
- **`∂u_ss/∂window` is exact.** For the (gated) autonomous linear disposition the one-cycle map is affine in `u0` with `M = e^{A·II}` window-*independent* (`e^{A(II−w)}·e^{Aw} = e^{A·II}`), so the window sensitivity rides `b` alone. `inject_rate_saltation` injects the velocity **jump** `r·e·dtinf` (the Leibniz boundary term), which then quiet-propagates — matching `∂y/∂w = Quiet_prop·rate·e_cmt` by direct computation, with the active window's interior term captured by integrating to the nominal bound. The δ²-Hessian term is exact at θ₀ (the perturbation `δ` has value 0, so the value-only `jdotg_value` is correct).
- **Autonomy is enforced, not assumed.** A TIME/TAFD/TAD-reading RHS — which would break the cycle-recurrence assumption — is routed to FD by the support gate (`ode_provider_ss_tad_dependent_rhs_routes_to_fd`). An SS infusion into a built-in-absorption compartment is rejected upstream (`E_ABSORPTION_SS_INFUSION`), so the new plain-`+rate` branch is unreachable for those.
- Edge cases (`ii ≤ 0`, `cmt ≥ n`, overlapping `t_inf > ii`, `t_inf = ii` zero-quiet-window, integration failure) all degrade to the correct zero-trough / fallback path.

### Issues found — both test-quality, both fixed in `8138978d`

1. **Weak cycle bound (`src/sens/ode_provider_tests.rs`).** `ode_provider_ss_nonlinear_fallback_early_stop_matches_full_budget` asserted `(1..full_cycles).contains(&early_cycles)`, but the half-open range *includes 1* — so it would pass vacuously if a regression let the MM model take the single-cycle exact solve instead of the fallback (the stated intent is ">1"). Only redundantly backstopped by the `full_cycles == 50` check. Tightened to `2..full_cycles`, matching the sibling `ss_nonlinear_bolus_with_steady_state_uses_fallback` and the existing `2..` convention.

2. **No PR-level truncation-sensitive infusion oracle.** The slow-disposition value oracle only covered the **bolus** (`ss_linear_disposition_uses_exact_fixed_point`, `ke·II ≈ 0.024`, ~30% gap). Every PR-level *infusion* test sat at `ke·II ≥ 1.2`, where a 50-cycle truncation residual is `exp(−60)` — invisible, so those tests pass on *both* the old truncated train and the new exact solve and cannot catch an infusion truncation-tail bug (the slow-infusion twin lives in `tests/`, which runs nightly only). Added `ode_ss_slow_infusion_matches_analytical_ss` (`ke·II ≈ 0.03`, ~22% gap), anchoring the infusion exact solve to the independent `one_cpt_infusion_ss` closed form — it fails against the pre-#914 truncated train.

### Non-blocking observations (recorded, no change)

- **Warning-branch assertion** in `ss_nonlinear_over_capacity_bolus_caps_and_warns` matches only the shared message prefix. This is deliberate and correct — the capped drift can estimate ρ either side of 1, so pinning a specific branch would be flaky (the sibling `ss_input_rate_no_steady_state_warns` matches the prefix for the same reason).
- **Pre-existing, out of scope:** the overlapping-infusion guard tests the F-scaled `t_inf` while the upstream `W_STEADY_STATE_INFUSION` warning tests the unscaled `duration`; a rate-defined infusion with `F > 1` and `duration ≤ II < F·duration` returns zeros without the warning. Narrow, predates #914 (`F > 1` is unusual). Not addressed here to keep scope tight.

Re-validated after the fixes: full `--lib` suite, the two touched tests, and the tightened integration twins remain green.

PR #924, opened 2026-07-24
fix(pk): pure-TTE population declines the predictor instead of panicking on a CMT≥2 dose [closes #905]

### LINKED ISSUE #905: fix(pk): a pure-TTE population is exempt from dose-compartment validation but still reroutes to the walk, which panics
## Problem

`check_dose_compartments` exempts a **pure-TTE / pure-discrete population** from dose-compartment validation (`population_uses_analytical_pk`, `src/api/validation.rs`). The exemption is correct in itself: such a model still needs a `[structural_model]` line, and the idiom is a placeholder `one_cpt_iv` with dummy `CL`/`V`, whose 1-compartment topology would reject perfectly good dose records aimed at a TTE endpoint's `CMT`.

But `dose_needs_event_walk` (`src/pk/mod.rs`) does **not** know about that exemption. It reroutes any bolus with `cmt != 1` onto the event-driven walk, whose out-of-range guard is a `panic!`. So on an exempt population the check declines to validate and the walk then aborts the process on the very dose it declined to look at.

Reproduced by an adversarial-review agent during PR #896:

```
pure_tte_subject_with_a_high_cmt_dose_still_predicts
  -> panicked: dose into compartment 2 but model has 1 states
```

## Reachability

Narrow, but real:

- Via `read_population_for`, TTE rows are routed to `obs_records`, leaving `obs_times` empty, so `event_driven_predictions_with_schedule_impl` early-returns on `n_obs == 0` and the panic is never reached. **This is the normal path and it is safe.**
- Via `read_nonmem_csv` (no endpoint routing), the rows land in `obs_times` and the walk runs. That is the lower-level public API, and it is what the exemption's own regression test uses.

So the exemption's safety currently rests on a guard in a different module that it never references — worth pinning either way.

## Why it was not fixed in #896

PR #896's review fixed three confirmed defects in this area (a wrong-OFV analytic readout path, the `AMT=0` range-check hole, and a `[derived]` grid path left on the old predicate). This one was left deliberately: the obvious fixes each have a real cost and the choice is a design call, not a mechanical one.

- **Range-check exempt populations anyway** — rejects legitimate TTE data, since the placeholder has `n_states = 1` and the TTE endpoint's `CMT` is routinely ≥ 2. This is exactly what the exemption exists to prevent.
- **Teach `dose_needs_event_walk` about the exemption** — it takes `(PkModel, &Subject)` and has no access to the model's endpoints or to the population, so this means changing its signature and every call site.
- **Make the walk's out-of-range guard non-fatal** — reintroduces the silent dose-dropping that #375 removed.

A fourth option is probably the right one: have the *predictor* decline to run at all for a population with no Gaussian observation, rather than having the *validator* special-case it — i.e. move the exemption from "skip the check" to "skip the prediction". That is a larger change than a bug fix should carry.

## Validation

A regression test asserting that a pure-TTE population with a `CMT ≥ 2` dose either predicts or returns an actionable error — but never panics — reachable through both `read_nonmem_csv` and `read_population_for`. No NONMEM anchor needed: this is about an error channel, not a numerical path.

## References

- Surfaced by the adversarial review of PR #896 (see the review comment on that PR).
- `population_uses_analytical_pk` + `check_dose_compartments`: `src/api/validation.rs`.
- `dose_needs_event_walk`: `src/pk/mod.rs`.
- The `n_obs == 0` early return the exemption implicitly relies on: `src/pk/event_driven.rs`.
- Same family as #898 (panic-vs-`Result` on the non-`fit` entry points).

TeunP:
## Why

A pure-TTE / pure-discrete population aborts the process on a `CMT ≥ 2` dose. Such a
population is **exempt** from dose-compartment validation (`check_dose_compartments`
via `population_uses_analytical_pk`) — its `[structural_model]` is a placeholder
`one_cpt_iv` with dummy `CL`/`V`, and its TTE endpoint's `CMT` is routinely ≥ 2, so
range-checking the dose against the 1-compartment placeholder would reject perfectly
good data. But the **predictor** did not know about that exemption:
`dose_needs_event_walk` still reroutes any `cmt != 1` bolus onto the event-driven
walk, whose out-of-range guard is a `panic!`. So the validator declined to check a
dose that the walk then aborted on (#905).

Reproduced on `main` (adversarial review of PR #896):

```
pk_model = OneCptIv;  obs_times = [5.0];  obs_cmts = [2]
panicked at src/pk/event_driven.rs:785 — "dose into compartment 2 but model has 1 states"
```

Reachable when the population is loaded model-blind (`read_nonmem_csv`, no endpoint
routing → TTE rows land in `obs_times`, so `n_obs > 0` and the walk runs). The
endpoint-routed loader (`read_population_for`, what `fit_from_files` uses) was safe by
accident — it routes TTE rows into `obs_records`, leaving `obs_times` empty, so the
walk's `n_obs == 0` early-return fired. The exemption's safety rested on a guard in a
different module it never referenced.

## What changed

**The predictor now declines in lockstep with the exemption** (issue option 4). One
shared per-subject predicate owns the invariant:

- **`pk::subject_feeds_analytical_pk(model, subject)`** — true iff the subject has a
  Gaussian observation (an obs whose CMT is not a registered non-Gaussian endpoint) or
  a pk-only prediction time. `validation::population_uses_analytical_pk` is now
  `subjects.any(...)` of it — single source of truth.
- **Value gate** — `compute_predictions_with_tv_into_with_schedule` (the one funnel
  behind predict / simulate / likelihood / sdtab / postfit) returns `vec![NaN; n_obs]`
  when `ode_spec.is_none() && !subject_feeds_analytical_pk`, before reaching the walk.
- **Gradient gate** — `sens::provider::subject_routes_to_event_walk` returns `false`
  for the same condition, so FOCE/FOCEI never enters the walk's `sens` twin
  (`propagate::active_rates_g` / the bolus `cmt_idx` guard, both of which `panic!` the
  same way). Load-bearing, not cosmetic: on a model-blind `fit()` the gradient step
  hits the twin panic independently of the value path.
- **IOV twins** — IOV models (`n_kappa > 0`) route through *separate* funnels
  (`pk::predict_iov` for value; `subject_sensitivities_iov` /
  `subject_eta_grad_iov_analytical` for the gradient), which do not consult
  `subject_routes_to_event_walk`. The same decline is added to all three, keyed on the
  same predicate. (The routed loader already declines these via the existing
  `subject_has_survival_records` gate; this only fires on the model-blind loader.)
- **Residual skip** — the Gaussian residual loops in `individual_nll_into_with_schedule`
  and `obs_nll_subject_from_preds` (the SAEM/IS twin) skip an observation whose CMT is a
  non-Gaussian endpoint (no-op on the routed path; keeps a model-blind `NaN` pred from
  poisoning the inner NLL).
- The `check_dose_compartments` exemption comment now states the safety explicitly.

**Why the per-subject gate is safe within `ode_spec.is_none()`:** nothing but a
Gaussian observation consumes the analytical predictor there. A hazard that reads a
concentration is written `hazard = <expr>`, which the parser turns into a synthesised
`__chz` ODE state (`prescan_ode_hazards`) → forces `ode_spec = Some`; every hazard left
on an analytical model is a closed-form `HazardSpec::Analytic` over `(θ, η, cov)`. A
binary/categorical linear predictor is `LinearPredictorFn = Fn(θ, η, cov) -> f64` — no
concentration. So a subject with no Gaussian obs feeds the predictor nothing.

## Alternatives considered

- **Range-check exempt populations anyway** — rejects legitimate TTE data (the
  placeholder has `n_states = 1`, the endpoint's `CMT` ≥ 2). The exemption exists to
  prevent exactly this.
- **Teach `dose_needs_event_walk` the exemption** — it takes `(PkModel, &Subject)`; wiring
  the population's endpoints through it means a signature change across 6 call sites, and
  is the wrong layer.
- **Make the walk's guard non-fatal** — reintroduces the silent dose-dropping #375 removed.
- **Route non-Gaussian obs out of `obs_times` on the in-memory path** (make model-blind ==
  file-load) — stronger (also fixes the latent Gaussian-misclassification), but needs
  reconstructing `ObsRecord`s and is larger than this bug warrants. Deferred.

## Scope / known limitation

The fix guarantees **panic-free on every path**, and **correct on the endpoint-routed
path**. A model-blind `fit()` of a pure-TTE population still cannot fit the TTE data (it
never reached `obs_records`), so its outer FOCEI OFV is a degenerate `NaN` and the
covariance step declines cleanly — no abort. A correct TTE fit requires
`read_population_for` / `fit_from_files`. This is inherent to the model-blind reader, not
a regression.

## Cross-repo dependency

No `pub` API signature changed (`subject_feeds_analytical_pk` is `pub(crate)`;
`compute_predictions_with_tv*` bodies only). ferx-r unaffected, no pin bump.

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable — this is an error-channel fix, not a numerical path (per the issue:
  "No NONMEM anchor needed"). The routed path's numbers are unchanged (the gate returns the
  same empty `preds` the `n_obs == 0` early-return already produced).

## Performance
- [x] No significant impact expected — one `ode_spec.is_none()` + predicate check per
  prediction call; the predicate short-circuits on `endpoints.is_empty()` (every plain PK
  model), so non-survival builds and ordinary PK models pay a single bool.

## Tests
- [x] Unit test (`sens/provider_tests.rs::tte_only_subject_declines_the_event_walk_gradient`)
  pins the non-IOV gradient gate directly.
- [x] Regression tests that fail without this fix (`tests/dose_compartment_routing.rs`):
  pure-TTE `CMT ≥ 2` dose × predict / simulate / fit, through **both** `read_nonmem_csv`
  and `read_population_for`; the same for the **IOV** funnel (`simulate` + `fit` with `OCC`);
  plus a **positive control** (a Gaussian-bearing subject on a TTE model still predicts
  finite — the gate must not over-fire).
- [x] All new tests **mutation-verified**: reverting each gate makes its test panic —
  value `event_driven.rs:785`, non-IOV gradient (unit-assert flips), IOV value
  `event_driven.rs:785`, IOV gradient `propagate.rs:1408`.
- Full lib suite `--features survival`: **2823 / 0**. Integration
  `dose_compartment_routing --features survival`: **28 / 0**. No-`survival`
  `cargo check --tests` clean. Existing exemption tests stay green.

The IOV funnel gap (item above) was found by an adversarial (xhigh) review of this PR
and fixed in the same change.

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` → `Fixed` entry added (#905).
- [x] No new DSL / fit option / output field → no `docs/` page change.

## Reviewer hints

The load-bearing claim is *"on an analytical model nothing but a Gaussian observation
consumes the predictor"* — see `subject_feeds_analytical_pk`'s doc. If that held only for
hazards but not for binary/CTMM linear predictors, the value gate would silently kill a
conc-driven discrete endpoint; it holds because `LinearPredictorFn` has no concentration
argument and a conc-driven hazard/intensity forces `ode_spec = Some` (out of the gate's
domain). The gradient gate mirrors the value gate exactly — they must agree or FOCE would
differentiate a different dosing history than it predicted.

## Open questions

Whether to also do the heavier "route non-Gaussian obs out of `obs_times` on the in-memory
path" so a model-blind `fit()` becomes *correct* (not just panic-free). Deferred here as
out of scope for the crash fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,14 @@ section of the SDLC for the versioning policy).
   no rows** (#759). The equivalent `simulate()` guard already existed and its message claimed to
   cover `predict()`, but it only ran on the simulate path. State-occupancy prediction `π(t)` is
   still to come (#820).
+- **A pure-TTE / pure-discrete population no longer panics on a `CMT ≥ 2` dose** (#905). Such a
+  population is exempt from dose-compartment validation — its `pk` line is a placeholder and the
+  TTE endpoint's `CMT` is routinely ≥ 2 — but the predictor still rerouted the dose onto the
+  event-driven walk, which aborted the process on the very dose the validator had declined to
+  check. The analytical predictor now declines in lockstep with the exemption: a subject with no
+  Gaussian observation returns `NaN` without entering the walk (value and FOCE/FOCEI gradient
+  paths alike), so `predict()`, `simulate()` and `fit()` stay panic-free on such a population
+  through both the endpoint-routed and model-blind loaders.
 
 ### Performance
 - **Closed-form modified-release absorption** (#860). A static multi-route absorption model
--- a/src/api/validation.rs
+++ b/src/api/validation.rs
@@ -641,37 +641,37 @@ pub(crate) fn check_modeled_dose_rates(
 /// placeholder's 1-compartment topology would reject a working model over dose
 /// records no PK code ever reads.
 ///
-/// Deliberately a **population**-level predicate, not a per-subject one. "This
-/// subject has no Gaussian observation" does **not** imply the PK path is
-/// skipped for it: the event-driven walk is driven by the subject's event
-/// schedule (doses and `EVID=2` times), and the FOCE/FOCEI sensitivity provider
-/// runs per subject regardless. A per-subject exemption would let a TTE-only
-/// subject *of a joint PK-TTE model* carry an unroutable infusion straight back
-/// into the panics this check exists to prevent — a routine pattern (an early
-/// dropout, or a TTE-only arm, that still has dose records).
+/// Deliberately a **population**-level gate on *whether the dose-compartment check
+/// runs at all*, not a per-subject one. If **any** subject asks the analytical
+/// predictor for a value, the check validates **every** dose in the dataset — so a
+/// TTE-only subject *of a joint PK-TTE model* (an early dropout, or a TTE-only arm,
+/// that still has dose records) has its unroutable infusion caught up front, rather
+/// than carried into the event-driven walk's `panic!`. Making *validation* per
+/// subject would reopen exactly that.
 ///
-/// Non-Gaussian endpoints are registered per CMT in `CompiledModel::endpoints`,
-/// so an observation CMT absent from that map is Gaussian. Without the
-/// `survival` feature no non-Gaussian endpoint can be parsed, so every
-/// observation is Gaussian by construction and this is always `true`.
+/// This is distinct from — and complementary to — the per-subject predictor gate
+/// (`subject_feeds_analytical_pk`, #905): validation is population-scoped so a live
+/// population is fully checked, while the predictor declines the walk per subject for
+/// any subject whose output is unused. In a live population the check already rejects
+/// a bad dose, so the predictor gate never has to; in an exempt population the check
+/// is skipped and the predictor gate is what keeps the walk unreached.
+///
+/// The per-subject truth (which observation CMTs are Gaussian, and the
+/// no-`survival`-feature short circuit) lives in [`crate::pk::subject_feeds_analytical_pk`];
+/// this is `any(...)` of it.
 fn population_uses_analytical_pk(model: &CompiledModel, population: &Population) -> bool {
-    #[cfg(feature = "survival")]
-    {
-        if model.endpoints.is_empty() {
-            return true;
-        }
-        population.subjects.iter().any(|s| {
-            !s.pk_only_times.is_empty()
-                || s.obs_cmts
-                    .iter()
-                    .any(|cmt| !model.endpoints.contains_key(cmt))
-        })
-    }
-    #[cfg(not(feature = "survival"))]
-    {
-        let _ = (model, population);
-        true
-    }
+    // Single source of truth with the predictor gate: `subject_feeds_analytical_pk`
+    // is the per-subject predicate, and this is `any(...)` of it. The two compose —
+    // this population-level test decides whether the dose-compartment check runs at
+    // all (a *live* population validates every dose, including a TTE-only subject's),
+    // while the per-subject gate declines the walk for a subject the analytical
+    // predictor would feed nothing (#905). See `subject_feeds_analytical_pk` for why
+    // "no Gaussian observation" implies "predictor output unused" on an analytical
+    // model.
+    population
+        .subjects
+        .iter()
+        .any(|s| crate::pk::subject_feeds_analytical_pk(model, s))
 }
 
 /// Every dose must name a compartment the **analytical** engine can actually
@@ -761,6 +761,18 @@ pub(crate) fn check_dose_compartments(
     // Nothing in this dataset asks the analytical predictor for a value — the
     // `pk` line is a pure-TTE / pure-discrete model's placeholder. See
     // `population_uses_analytical_pk` for why this is a population-level test.
+    //
+    // **This exemption is safe because the predictor declines in lockstep.** Skipping
+    // the dose-compartment check would be a landmine on its own: `dose_needs_event_walk`
+    // still reroutes a `CMT ≥ 2` dose onto the event-driven walk, whose out-of-range
+    // guard is a `panic!`. That was #905 — the check declined to validate a dose the
+    // walk then aborted on. The fix is the shared predicate: the same
+    // `subject_feeds_analytical_pk` that makes this population exempt also makes
+    // `compute_predictions_with_tv_*` (value) and `subject_routes_to_event_walk`
+    // (gradient) return a NaN/decline for every one of its subjects, so the walk is
+    // never reached. The exemption's safety no longer rests on the far-away
+    // `n_obs == 0` early return in a different module; it is the negation, per subject,
+    // of this very predicate.
     //
     // **Analytical models only.** An `[odes]` block is never a placeholder: a
     // hazard that reads an ODE state has the system integrated for it even when
--- a/src/pk/mod.rs
+++ b/src/pk/mod.rs
@@ -861,6 +861,17 @@ pub fn predict_iov(
     kappas: &[Vec<f64>],
 ) -> Vec<f64> {
     use std::collections::HashMap;
+    // #905: the IOV value funnel is a structural peer of
+    // `compute_predictions_with_tv_into_with_schedule` and needs the same declination —
+    // an analytical subject with no Gaussian observation feeds the predictor nothing,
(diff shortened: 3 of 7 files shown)


TeunP:
## Adversarial review (xhigh) of this change

Ran an xhigh self-review — 3 independent trace agents + a line-level read — to confirm no path reaches the event-driven walk unguarded.

### Found + fixed one HIGH issue in-place

IOV models (`n_kappa > 0`) route through **separate** funnels — `predict_iov` (value) and `subject_sensitivities_iov` / `subject_eta_grad_iov_analytical` (gradient) — none of which consulted `subject_routes_to_event_walk`. An analytical IOV model with a closed-form-`family` TTE hazard (an allowed combination — only ODE-`hazard`+IOV is rejected), loaded model-blind, still panicked on a `CMT ≥ 2` dose:

- **value** — `src/pk/event_driven.rs:785`, via `simulate` / SAEM / IMP / postfit
- **gradient** — `src/sens/propagate.rs:1408`, via the IOV inner analytic gradient on `fit`

Fixed by declining in all three IOV funnels on the same predicate. **No regression:** the endpoint-routed loader already declines these via the existing `subject_has_survival_records` gate, so the new gate fires only on the model-blind loader.

### Verified sound (no change needed)

- **Non-IOV walk completeness** — every production predictor (`predict`/`simulate`/`fit`/`postfit`/`npde`/`gauss_newton`/`saem`/`agq`/`importance_sampling`) routes through the gated funnel; `model_preds` is `#[cfg(test)]`; `adaptive` is ODE-only.
- **No over-fire** — the gate cannot decline a subject that carries a Gaussian observation (`obs_cmts.any(∉ endpoints)`); a concentration-reading endpoint forces `ode_spec = Some` and leaves the gate's domain.
- **No under-fire** — a *live* population still range-checks every dose (`population_uses_analytical_pk = any(subject_feeds_analytical_pk)` ↔ the per-subject predictor gate).
- **Dense-residual path** unreachable for an exempt subject — `block_sigma + endpoint` is rejected up front (`E_BLOCK_SIGMA_TTE_UNSUPPORTED`).

### Also added

- A **positive-control test** — a Gaussian-bearing subject on a TTE model still predicts finite (a one-CMT typo in the predicate would otherwise slip through undetected; the decline tests all assert `NaN`).
- A symmetric skip in `obs_nll_subject_from_preds` (the SAEM/IS twin of the FOCE inner skip).
- A doc fix.

### Validation & data

This is an **error-channel fix, not a numerical one**, so there is no NONMEM / nlmixr2 anchor — and none is needed (the issue states this). The supported (endpoint-routed) path's numbers are **unchanged**: the gate returns the same empty `preds` the walk's `n_obs == 0` early-return already produced. Concrete evidence instead:

- **Mutation** (each gate reverted → its test panics at the named site, confirming it is load-bearing):
  - non-IOV value gate → `predict` panics `src/pk/event_driven.rs:785`
  - non-IOV gradient gate → unit assertion flips (`subject_routes_to_event_walk` returns `true`)
  - IOV value gate → `simulate` panics `src/pk/event_driven.rs:785`
  - IOV gradient gate → `fit` panics `src/sens/propagate.rs:1408`
- **Suite** — full lib `--features survival` **2823 / 0**; integration `dose_compartment_routing --features survival` **28 / 0**; no-`survival` `cargo check --tests` clean.
- **CI** — green on the reviewed head, including `codecov/patch` (new lines covered), Survival (TTE), Markov (CTMM), Clippy, Format.

**Known limitation (by design):** a model-blind `read_nonmem_csv` + `fit()` of such a population stays degenerate (`NaN` OFV, covariance step declines cleanly, no abort) — a correct TTE fit needs `read_population_for` / `fit_from_files`.

Issue #410, opened 2026-06-19
Arm user-ODE analytic η/θ sensitivities (Option A, Phase 5)


hiddevandebeek:
## Arm the user-ODE analytic η/θ sensitivities (Option A, Phase 5)

The analytic-sensitivity rollout (#367) shipped scoped to the analytical PK models. The **ODE sensitivity engine is already built and tested** in `src/sens/ode_provider.rs` but ships gated off behind `const ODE_SENS_ENABLED = false` (`src/sens/provider.rs`); ODE models currently take the prior gradient-free-outer / FD-inner path. This issue tracks arming it.

### Already in place
- `ode_provider.rs` (~950 lines): augmented RK45 carrying state + `∂state/∂(θ,η)` jets over `Dual2`, the `[individual_parameters]` chain over `Dual2`, returning the same `SubjectSens` shape the analytical path produces.
- Outer-gradient wiring (`population_gradient_sens` dispatch), gated by `ODE_SENS_ENABLED`.
- 6 FD-validation tests (`ode_provider_*_matches_production`): 2-cpt, oral + `F`, infusion, non-zero init, EVID reset, covariate — passing.
- Scope: RHS-program models, `ObsCmt`/simple Form-C readout, bolus + finite infusion, `F`, resets, covariates, ≤12 params / ≤16 dual axes.

### To arm (outer analytic η/θ gradient — the main win)
- [ ] Flip `ODE_SENS_ENABLED = true`.
- [ ] **Harden the observation/time matching** in `ode_provider.rs` (`obs_map` is keyed on `t.to_bits()`): a value-equal-but-bit-different break time (`±0.0`, different arithmetic) misses. Replace exact-bit keying with index/tolerance matching. (The 6 tests pass only because their times align bit-exactly.)
- [ ] **NONMEM cross-check** for an ODE model's estimates/SE (CLAUDE.md requirement for numerical behaviour).
- [ ] **Performance pass / right-size N** — the augmented RK45 is ~O(N) costlier per step; confirm acceptable and the dual-width cap (`MAX_ODE_AXES`).

### Follow-ups (out of the minimal arming)
- Light `Dual1` inner-EBE ODE provider (the inner loop stays on FD initially, like the LTBS path).
- Steady-state dosing for ODE (dual SS-equilibration), IOV, SDE sensitivities, `input_rate` absorption, lagtime, expression scaling — all currently excluded by `ode_analytical_supported`.

[...]

Issue #378, opened 2026-06-16
estimation: ODE↔analytical FOCE *marginal* OFV diverges (3-cpt IV, post-#330); equivalence test only covers fixed-η NLL


TeunP:
## Context

Surfaced by the ferx-r pin bump (FeRx-NLME/ferx-r#170, pin → `1bdd9105`). ferx-r's
`test-ode-analytical-equivalence.R` compares the **FOCE marginal OFV** (a
`maxiter=0` fit — inner η̂ solve + Laplace `log|H̃|`) of each ODE-form example
against its analytical sibling. After the bump, `three_cpt_iv_ode` vs analytical
`three_cpt_iv` (proportional error, FOCE) diverges, and the magnitude is
**platform/build-sensitive**:

| Platform / build              | `abs(ode_ofv − an_ofv)` | band |
|-------------------------------|-------------------------|------|
| Linux x86_64, release (CI)    | **18.3**                | 1.0  |
| macOS, dev build (local)      | 1.06                    | 1.0  |

On macOS-dev, `two_cpt_iv` (0.52) and `two_cpt_oral_cov` (1.8) also exceed their
0.5 bands but **pass** on Linux-release — i.e. the marginal OFV for these stiff
pairs is fragile across platform/optimisation. The `three_cpt_iv_ode` example
already uses `ode_reltol = 1e-10` / `ode_abstol = 1e-12`, so this is **not** a
solver-tolerance issue.

## Why it's the FOCE *marginal* path, not structural

`tests/analytical_ode_equivalence.rs` asserts ODE↔analytical agree on (a) pointwise
PRED and (b) the **fixed-η** population NLL (`individual_nll`, `OFV_RTOL = 2e-3`) —
both still pass on `1bdd9105`. So the structural model and per-point likelihood are
fine; the divergence is only in the FOCE **marginal** objective (the inner η̂ solve
plus the Laplace `log|H̃|` determinant), which that test does not exercise.

The change that touched exactly this machinery is **#330** (`c10c906b`: tighten
`inner_tol` to `1e-5`, evaluate the FOCE residual variance at `f(η=0)`, add the
`log|H̃|` EBE-response term). The steep 3-cpt IV proportional-error curve amplifies
the difference. ferx-r `main` (previous pin, `7664845`) is green on Linux-release,
so the bump is what moved `three_cpt_iv` from < 1.0 to 18.3 there.

## Ask

[...]

PR #927, opened 2026-07-24
fix(estimation): report the analytic inner gradient for in-scope ODE models [closes #926]

### LINKED ISSUE #926: report: `gradient_method_inner` mislabels in-scope ODE models as "finite differences"
### Summary

`build_info::gradient_method_inner` reports **"finite differences"** for an in-scope
**ODE** model, even though that model's inner EBE solve actually runs the exact analytic
`Dual1` ODE η-gradient (#410). It is a **reporting-only** defect — the diagnostic label is
wrong; estimates, OFV, and the gradients themselves are unaffected — but it is misleading:
it is what led the #378 issue text and the ferx-r `test-ode-analytical-equivalence.R`
comments to state "the ODE form uses finite differences", which is false.

### Where

`src/build_info.rs` — `gradient_method_inner` (currently ~line 66) decides `Analytic` vs
`FiniteDifferences` from:

```rust
let analytic = analytic_inner_grad_supported_model(model)          // closed-form only
    || (iov_sens_supported(model)                                 // IOV inner branch
        && model.default_params.omega_iov.is_some()
        && !analytic_inner_common_bail(model));
```

`analytic_inner_grad_supported_model` (`inner_optimizer.rs`) resolves to
`sens::provider::analytical_supported`, whose core (`analytical_supported_core`,
`provider.rs`) requires `model.ode_spec.is_none()` — so it is **false for every ODE
model**. There is a disjunct for the **IOV** inner route but **none for the plain
(non-IOV) ODE inner route**, so an in-scope ODE non-IOV model falls through to
`FiniteDifferences`.

### Actual behaviour

The live per-subject route `analytic_inner_grad_supported` (`inner_optimizer.rs`) has an
explicit `if model.ode_spec.is_some()` branch that runs the light `Dual1` ODE inner
provider (`ode_inner_grad_supported`, #410), subject only to the global escape hatches
(`gradient = fd`, SDE, magnitude × `block_sigma`, forced-FD). So these models run
**analytic**, and the report should say so.

- Outer report (`gradient_method_outer`) is already correct — it reads `sens_supported`,
  which includes the ODE path.
- IOV ODE inner is already correct via the IOV branch above.
- **Non-IOV ODE inner is the lone straggler.**

### Impact

Diagnostic/label only. No effect on estimates, objective, gradients, or any file output's
numerics. Priority is low; the cost is user confusion (as seen in #378).

### Fix direction

Add a non-IOV ODE disjunct mirroring the existing IOV one, e.g. an
`ode_inner_grad_supported_model(model) && !analytic_inner_common_bail(model)` term, where
the new model-level predicate single-sources the ODE master switch
(`ODE_SENS_ENABLED && ode_analytical_supported`). It must respect the same scope the live
route does so it never over-claims analytic where subjects actually FD:

- `ode_analytical_supported` requires `n_kappa == 0`, which makes
  `analytic_inner_common_bail`'s LTBS×IOV clause vacuous, so the guard reduces to exactly
  the live ODE branch's escape hatches — plain LTBS ODE stays in scope (correct: the
  `Dual1` walk shares `solve_ode_g` with the objective, #474).
- Do **not** widen it to a bare `sens_supported` disjunct — that reintroduces the CTMM
  "vacuous `tv_fn`" misreport trap that `analytic_inner_grad_supported_model` guards
  against.
- Leave `analytic_inner_grad_supported_model` itself untouched (it is also consumed by
  `fd_fallback_warning`; changing it would regress the warning).

Add a report-equals-route agreement unit test (report `Analytic` ⟺ the live route takes
the analytic branch) so the label can't drift from the dispatch again as scope grows.

Surfaced during #378. Reporting-only, so no NONMEM anchor applies.

TeunP:
Closes #926.

## Why

The [#378](https://github.com/FeRx-NLME/ferx-core/issues/378) investigation surfaced a second, cosmetic defect (deferred at the time): `build_info::gradient_method_inner` **mislabels** the inner-loop gradient method for `[odes]` models.

`fit()` already runs the exact analytic `Dual1` ODE inner η-gradient (`ode_inner_grad_supported`, #410) for an in-scope ODE model — as it does for closed-form and IOV models — and the **outer**-loop report already recognizes ODE (via `sens_supported`). But the **inner** report derived its answer from `analytic_inner_grad_supported_model` → the closed-form-only `analytical_supported`, which is `false` for every ODE model (ODE models have no `tv_fn`). So every in-scope ODE non-IOV model was reported as `"finite differences"` in the fit banner and `{model}-fit.yaml`, contradicting both the outer report and the route actually run. This mislabel is called out as defect #2 in the #378 diagnosis and misled the original investigation (and a ferx-r comment).

## What changed

Add an ODE non-IOV disjunct to `gradient_method_inner`, mirroring the existing IOV branch:

```rust
let analytic = analytic_inner_grad_supported_model(model)                    // closed-form + CTMM
    || (ode_inner_grad_supported_model(model)                               // ODE non-IOV  ← new
        && !analytic_inner_common_bail(model))
    || (iov_sens_supported(model) && omega_iov.is_some()                    // IOV
        && !analytic_inner_common_bail(model));
```

- New model-level predicate `sens::provider::ode_inner_grad_supported_model` = `ODE_SENS_ENABLED && ode_analytical_supported(model)` — the model-level sibling of the per-subject `ode_inner_grad_supported`, single-sourcing the `ODE_SENS_ENABLED` master switch.
- Gated by the shared `analytic_inner_common_bail` escape hatches (`gradient = fd`, SDE, magnitude × `block_sigma`). Because `ode_analytical_supported` requires `n_kappa == 0`, the common bail's LTBS×IOV clause `(log_transform && n_kappa > 0)` is **vacuous** here, so the disjunct reduces to *exactly* the model-level gates the live per-subject ODE branch (`analytic_inner_grad_supported`) applies. Plain LTBS ODE stays in scope, matching the live route.
- The closed-form and IOV branches, and `analytic_inner_grad_supported_model` itself, are **untouched** — so `fd_fallback_warning` is unchanged.

Reporting only — **no** estimate, OFV, or diagnostic changes.

## Alternatives considered

- **Fold ODE into `analytic_inner_grad_supported_model`** — rejected: that predicate also feeds `fd_fallback_warning`, so making it true for ODE would change the FD-fallback warning logic (a behavior change beyond reporting).
- **Use the broad `sens_supported` disjunct** (which already includes `ode_analytical_supported`) — rejected: `sens_supported` reintroduces the CTMM vacuous-`tv_fn` trap that `analytic_inner_grad_supported_model` was written to avoid (an endpoint-only CTMM model is given a vacuous `tv_fn` and would be waved through). A dedicated ODE-only predicate sidesteps it.

## Cross-repo dependency

`ode_inner_grad_supported_model` is `pub(crate)`; the only public surface is the **string value** of `FitResult.gradient_method_inner`, which ferx-r reads as an opaque display string. No schema/API change, no pin bump.

## Breaking changes
- [x] None

> The reported **value** of `gradient_method_inner` changes for in-scope ODE models (`"finite differences"` → `"analytic (Dual2)"`), but the field's name/type/shape is unchanged — ferx-r consumes it as a display string.

## Numerical validation
- [x] Not applicable (reporting label — no estimates, OFV, or diagnostics change)

## Performance
- [x] No significant impact expected (one extra model-level predicate call in a reporting function)

## Tests

New tests (all **mutation-verified** — dropping the ODE disjunct flips the report to `FiniteDifferences` while the live route stays analytic):
- `build_info::inner_in_scope_ode_returns_analytic` — an in-scope ODE model now reports `Analytic`.
- `build_info::inner_forced_fd_ode_returns_fd` — `gradient = fd` still reports FD (escape-hatch composition).
- `build_info::inner_out_of_scope_ode_model_returns_fd` — an ODE model with no `rhs_program` stays FD (renamed from `inner_ode_model_returns_fd`, with a clarifying comment).
- `estimation::inner_optimizer::in_scope_ode_reports_and_takes_the_analytic_inner_route` — pins **report == route** (`gradient_method_inner` vs the per-subject `analytic_inner_grad_supported`), exactly like the CTMM agreement tests.

Full lib suite **2706/0** (18 slow-ignored); `--features "survival markov"` reporting subset **23/0**; `rustfmt` clean.

- [x] Not applicable (internal reporting fix, no new API surface)

> Effect: for an in-scope `[odes]` fit the banner / `-fit.yaml` line changes from `gradient_method_inner: finite differences` to `gradient_method_inner: analytic (Dual2)`.

## Docs
- [x] No user-visible change *(beyond the corrected report string)* — the docs already state in-scope ODE inner is analytic (`docs/estimation/foce.qmd`, since #410); `docs/output.qmd` describes the field generically. `CHANGELOG.md` `[Unreleased] → Fixed` entry added.

## Checklist
- [x] `cargo clippy` clean (no new warnings in the touched files)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A (boolean predicate + reporting only)
- [x] `Cargo.toml` version bumped if breaking — N/A (not breaking)

## Reviewer hints

The load-bearing argument is the **equivalence** in `gradient_method_inner`: `ode_inner_grad_supported_model(model) && !analytic_inner_common_bail(model)` must equal the model-level decision of the live per-subject ODE branch in `analytic_inner_grad_supported`. It does because `ode_analytical_supported` requires `n_kappa == 0`, which makes `analytic_inner_common_bail`'s only ODE-irrelevant clause (`log_transform && n_kappa > 0`) vacuous — leaving exactly `{escape-hatch, gradient=fd, SDE, magnitude×block_sigma}`, the live branch's gates. Also worth confirming: the deliberate non-touch of `analytic_inner_grad_supported_model` (and hence `fd_fallback_warning`), and that the all-FD-population ODE case matches the already-accepted ODE-IOV reporting contract.

## Open questions

None. Tracked by #926 (this PR closes it); #378 is the closed originating investigation the defect surfaced in, referenced in the `CHANGELOG` entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,14 +71,17 @@ section of the SDLC for the versioning policy).
   units, platform-sensitive) that surfaced after the `inner_tol` tightening in #330. FREM keeps
   its cold-restart re-centering unchanged.
 - **The reported inner-loop gradient method for `[odes]` models is no longer mislabeled "finite
-  differences"** (#378). `fit()` already runs the exact analytic inner η-gradient for an in-scope
-  ODE model (as it does for closed-form and IOV models), but the reported `gradient_method_inner`
-  — shown in the fit banner and `{model}-fit.yaml` — was derived from a closed-form-only predicate
-  and always read "finite differences" for ODE, even while the outer-loop report correctly read
-  "analytic". It now reports the analytic method for an in-scope ODE model, matching the outer
-  report and the route actually run; genuinely finite-differenced subjects (steady-state, oral
-  infusion, modeled-duration doses) are still surfaced by the route banner and the FD-fallback
-  warning. Reporting only — no estimate, OFV, or diagnostic changes.
+  differences"** (#926, follow-up to #378). `fit()` already runs the exact analytic inner
+  η-gradient for an in-scope ODE model (as it does for closed-form and IOV models), but the
+  reported `gradient_method_inner` — shown in the fit banner and `{model}-fit.yaml` — was derived
+  from a closed-form-only predicate and always read "finite differences" for ODE, even while the
+  outer-loop report correctly read "analytic". It now reports the analytic method for an in-scope
+  ODE model, matching the outer report and the route actually run. The report and the
+  FD-fallback warning now share one predicate, so an in-scope ODE model whose every subject is
+  genuinely finite-differenced (oral infusion into a built-in absorption compartment, or a
+  rate-defined infusion under `F ≠ 1`) is still surfaced by the route banner and the warning
+  rather than silently labeled analytic. Reporting only — no estimate, OFV, or diagnostic
+  changes.
 - **Steady-state (`SS=1`) dosing on `[odes]` models is now exact for a linear disposition, and
   warns instead of silently truncating on a nonlinear one** (#914). An ordinary ODE bolus or
   infusion SS dose previously equilibrated by expanding a pulse train capped at 50 cycles, which
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,15 @@ section of the SDLC for the versioning policy).
   the same EBE. This removes the ODE↔analytical FOCE **marginal** OFV divergence (up to ~18 OFV
   units, platform-sensitive) that surfaced after the `inner_tol` tightening in #330. FREM keeps
   its cold-restart re-centering unchanged.
+- **The reported inner-loop gradient method for `[odes]` models is no longer mislabeled "finite
+  differences"** (#378). `fit()` already runs the exact analytic inner η-gradient for an in-scope
+  ODE model (as it does for closed-form and IOV models), but the reported `gradient_method_inner`
+  — shown in the fit banner and `{model}-fit.yaml` — was derived from a closed-form-only predicate
+  and always read "finite differences" for ODE, even while the outer-loop report correctly read
+  "analytic". It now reports the analytic method for an in-scope ODE model, matching the outer
+  report and the route actually run; genuinely finite-differenced subjects (steady-state, oral
+  infusion, modeled-duration doses) are still surfaced by the route banner and the FD-fallback
+  warning. Reporting only — no estimate, OFV, or diagnostic changes.
 - **Steady-state (`SS=1`) dosing on `[odes]` models is now exact for a linear disposition, and
   warns instead of silently truncating on a nonlinear one** (#914). An ordinary ODE bolus or
   infusion SS dose previously equilibrated by expanding a pulse train capped at 50 cycles, which
--- a/src/build_info.rs
+++ b/src/build_info.rs
@@ -54,52 +54,23 @@ impl GradientMethodKind {
 
 /// Gradient method used in the inner (per-subject EBE) loop.
 ///
-/// Coarse model-level mirror of the per-subject route in
-/// [`estimation::inner_optimizer::analytic_inner_grad_supported`]. The analytic
-/// η-gradient runs on one of three in-scope routes; this reports their union:
-///   - **closed-form** (`analytic_inner_grad_supported_model`): an analytical PK
-///     model (`tv_fn` populated) in the provider's scope — plain LTBS is served,
-///     LTBS + η-dependent `ExpressionScale` is not;
-///   - **ODE** (`ode_inner_grad_supported_model`): the light `Dual1` ODE walk (#410).
-///     The closed-form predicate misses it (ODE models have no `tv_fn`), so it is
-///     reported explicitly — the inner analog of how `sens_supported` already lets
-///     [`gradient_method_outer`] recognize ODE (#378 task B);
-///   - **IOV** (`iov_sens_supported` + `omega_iov`): the stacked-η IOV walk, which
-///     `analytic_inner_grad_supported_model` also misses (it requires `n_kappa == 0`).
-/// Each route is gated by the shared escape hatches (`gradient = fd`, SDE, …) via
-/// `analytic_inner_common_bail`. Otherwise FD.
+/// Best-case, model-level mirror of the per-subject route
+/// `estimation::inner_optimizer::analytic_inner_grad_supported` runs. The reported method is
+/// the union of the three in-scope inner routes — closed-form / CTMM, the light `Dual1` ODE
+/// walk (#410), and the stacked-η IOV walk — computed by the shared
+/// `inner_reports_analytic_model` predicate that `fd_fallback_warning` reads too, so the
+/// reported label and the FD-fallback warning cannot drift (#378 task B / #926). Otherwise
+/// FD. Before #926 this read a closed-form-only predicate and mislabeled every in-scope ODE
+/// model "finite differences".
 ///
-/// This is **best-case, model-level**: per-subject fallbacks — **TV-cov + LTBS**,
-/// where the event-driven inner walk declines LTBS so every such subject runs FD, and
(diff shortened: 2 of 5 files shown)


TeunP:
## Adversarial review — 3 independent reviewers + verification

Reviewed at `xhigh` with three independent adversarial subagents (correctness, convention/drift, and test-quality lenses), each reading the diff blind and prompted to **break** the fix, then verified every finding against the code by hand. CI-green was **not** treated as correctness evidence.

### Core fix is sound

All five correctness attacks on the branch-B equivalence found no defect. The claim that `ode_inner_grad_supported_model(model) && !analytic_inner_common_bail(model)` equals the live per-subject ODE branch's model-level gate holds: `analytic_inner_common_bail`'s only ODE-irrelevant clause (`log_transform && n_kappa > 0`) is **vacuous** because `ode_analytical_supported` forces `n_kappa == 0`; the branch is disjoint from the closed-form (`ode_spec` present) and IOV (`n_kappa == 0`) branches; and best-case truthfulness holds (whenever `ode_analytical_supported`, a plain-bolus subject is served by the static walk, and any lagtime/TV-cov/SS model routes to the event-driven walk).

### Findings — one real defect + doc errors, all fixed in `8a4f1b1f`

| Sev | Finding | Resolution |
|-----|---------|------------|
| **Med** | `fd_fallback_warning` still read the closed-form-only `analytic_inner_grad_supported_model`, so an in-scope ODE model whose **whole population** runs the FD inner (e.g. oral infusion into a built-in absorption compartment) persisted `analytic (Dual2)` to `fit.yaml` with **no** reconciling warning — contradicting the per-subject route banner and breaking the report↔warning coupling the comment at `inner_optimizer.rs` asserts. Both reviewers converged on this. | Extracted a single `inner_reports_analytic_model` predicate (closed-form \| ODE \| IOV) read by **both** `gradient_method_inner` and `fd_fallback_warning`, so the label and the warning can't drift. All-FD in-scope ODE now warns. Regression test added, **mutation-verified**. |
| **High** | `build_info` doc + CHANGELOG listed **steady-state** and **modeled-duration** doses as genuinely-FD ODE subjects — they are **analytic** on the ODE event-driven walk (`ode_tvcov_supported` admits `has_ss`/`has_modeled_dose`; #914/#530), contradicting the #914 entry two lines below. | Corrected to the real FD cases: oral infusion into a built-in absorption compartment, and a rate-defined infusion under `F ≠ 1`. |
| **Med** | `build_info` doc claimed the closed-form inner does **not** serve LTBS × `ExpressionScale` — it does (test `ltbs_plus_expression_scale_inner_matches_outer`). `analytic_inner_common_bail`'s own doc header was self-contradictory (header said "stays a common bail", body says "is NOT a bail"). | Both corrected; only LTBS × IOV stays a bail. |
| **Med** | CHANGELOG referenced the closed **#378** (a duplicate of the entry directly above it) instead of the issue this PR closes. | → `#926` (follow-up to #378). |
| **Low** | `provider.rs` doc glossed that the live ODE branch hand-inlines a *subset* of the bail (coinciding only at `n_kappa == 0`); an intra-doc link pointed at a private fn; a two-build test loop was vacuous (`gradient_method_inner` ignores the build arg); ODE + LTBS report was untested. | Reworded; link → code span; loop dropped; `inner_in_scope_ode_ltbs_returns_analytic` added. |

### Verification

- **Mutation-verified**: dropping the ODE disjunct from the shared `inner_reports_analytic_model` fails **4** tests — both report tests, the LTBS report, **and** the all-FD warning test — proving the shared predicate drives both consumers.
- Full lib suite **2708 / 0** (18 slow-ignored); `--features "survival markov"` subset **20 / 0**; `rustfmt` + `clippy` clean on the touched files.

Nothing outstanding from the review. The fix is reporting-only — no estimate, OFV, or diagnostic changes.

Issue #700, opened 2026-07-06
feat(sim): adaptive dosing — time-varying covariate support via per-event PK recompute [#391]


TeunP:
Follow-up to the adaptive-dosing safety guards in #699 (which **reject** time-varying-covariate subjects rather than run them silently wrong). This issue is the real **support**.

## Problem
The reactive driver (`run_adaptive_population` / `ode_predictions_adaptive_impl`) resolves each subject's PK **once** from the baseline (t=0) covariate snapshot — `api.rs`: `(model.pk_param_fn)(&theta, &eta_slice, &subject.covariates, 0.0)` — and threads that single `pk_params_flat` through every integration segment. A covariate that changes over the horizon therefore drives CL/V/KA off its t=0 value for the whole run, corrupting predictions, monitored signals, and dose decisions. #699 guards this with a typed error.

## Proposed support
Route the adaptive integrator through **per-event PK recomputation**, as the normal path does for `has_tv_covariates()` subjects via `ode_predictions_event_driven` / `compute_event_pk_params_into` (per-event `pk_at_dose` / `pk_at_obs` snapshots). Additionally, LOCF-interpolate the covariate to the decision time when a decision falls between observation rows (today the driver falls back to the baseline map unless the decision coincides with an obs row).

Remove the `has_tv_covariates()` branch from `reject_unsupported_adaptive` once the per-event path is wired and anchored.

## Validation (per CLAUDE.md — no NONMEM for adaptive)
- Degenerate oracle: a non-reactive controller over a TV-covariate subject reproduces `simulate()` on the equivalent static regimen.
- Reproduce an mrgsolve dynamic-dosing example that carries a time-varying covariate.

Tracked under #391.

PR #699, opened 2026-07-06
fix(sim): reject IOV/TV-cov/reset/SDE in adaptive dosing instead of silent wrong answers [#391]

TeunP:
## Why

An adversarial audit of the adaptive-dosing S1/S2 engine (#391) found four model/data
combinations that ran **silently wrong** — wrong numbers, no error, no warning, and
invisible to the frozen-replay verifier (which replays the same crippled engine). The
reactive driver integrates each subject from a single baseline (t=0) PK snapshot,
between-subject-variability only (kappas held at zero), and never re-derives PK, applies a
reset, or carries process noise:

- **IOV** (`kappa`) — all kappas forced to 0 (`api.rs`: `eta_slice.resize(n_eta + n_kappa, 0.0)`); biased predictions, reactive decisions, and dose ledger.
- **Time-varying covariates** — PK resolved once at t=0, so a covariate changing over the horizon drives CL/V/KA off its baseline value for the whole run.
- **System resets** (EVID=3/4) — the driver never applies the reset; compartment state silently fails to zero.
- **SDE `[diffusion]`** — the deterministic integrator drops the process-noise term.

This violates the module's own "never a silent wrong answer" contract.

> **Scope note — this PR only makes these cases fail *loudly*. It does not add support for
> any of them.** Actually running IOV / time-varying-covariate / reset / SDE models correctly
> under adaptive dosing is deferred future work, each tracked as a dedicated #391 follow-up
> issue — see [Future work](#future-work--real-support-not-in-this-pr) below.

## What changed

A single guard `reject_unsupported_adaptive(model, population)` at the
`run_adaptive_population` chokepoint (both public entry points — `simulate_adaptive` and
`simulate_adaptive_from_spec` — funnel through it) turns each of the four into a **typed
error**, mirroring the `n_kappa > 0` guards `fit()` already applies (IMPMAP / trust-region).
The low-level `ode_predictions_adaptive_impl` and its readout-covariate unit test are left
untouched: the guard is production-layer, so the internal readout-snapshot primitive that
future per-event PK support will build on is preserved.

Also in this PR:
- **Coverage backfill** — `adaptive_bolus_f_scaling_matches_static`: the bolus `u += F·amt` emit path was only ever exercised at F=1 (the infusion-F seam had a test; the bolus one did not).
[...]

Issue #391, opened 2026-06-17
epic(sim): adaptive / feedback dosing simulations (state-reactive regimens)


TeunP:
## Summary
Add the ability to simulate dosing regimens where future doses depend on the
simulated state — TDM target-attainment, oncology toxicity-driven dose
reduction, biomarker titration, target-controlled infusion — and to do so across
**multiple separately-fit models** (PK + PD + safety) using each model's **final
estimated parameters and uncertainty**, with **built-in verification** of the
event bookkeeping. Today every dose is fixed in `subject.doses` before
integration; `simulate()` is a pure forward pass with no seam for state-reactive
dosing and no notion of composing fitted models.

## Motivation
Adaptive/dynamic dosing is the standard tool for dose & schedule selection with
narrow-therapeutic-window drugs (Hooijmaijers et al., CPT:PSP 2023). That
framework is inherently multi-model — PK drives a PD/biomarker model which drives
a safety model whose state triggers dose modification — and those models are
typically fit *separately* (sequential PK → PD → safety). mrgsolve supports the
reactive-dosing part via modeled events (`self.mevent`), the `evtools` plugin, or
an R-loop — all hand-coded C++ with manual EVID/event-log and float-time
bookkeeping, none offering a declarative titrate/hold/reduce/stop abstraction
(mrgsolve's own docs note its examples are "scheduled, not feedback"), and none
offering composition of fitted artifacts or self-verification of the bookkeeping.
ferx can do all of this declaratively, with no per-model compile step and the
engine owning the timeline.

## Gap in ferx today
`ode_predictions` (`src/ode/predictions.rs`) builds a `break_times` timeline from
a fixed dose list and integrates segment-by-segment carrying state `u` forward.
There is no point where model logic reads `u` and injects a dose. That segment
loop is the natural feedback seam. There is also no model-composition / linking
concept — each `CompiledModel` is integrated independently.

---

## Part A — Reactive dosing controller (three layers, progressive disclosure)

**1. Declarative `[adaptive_dosing]` block** — the differentiator. Observe a
(optionally assay-noised) signal at a decision schedule; first-matching-rule
ladder. Covers TDM / toxicity / titration with no imperative code:

```
[adaptive_dosing]
  observe          = central / V          # signal: any expr over states/params (reuses read_observable)
  with_assay_error = true                 # titrate on the DV (noisy assay), not IPRED — TDM realism
  at               = every 24 from 24 to 336
  start_dose       = 100
[...]

PR #928, opened 2026-07-25
feat(sim): honor system-reset events (EVID=3) in adaptive dosing [closes #716]

### LINKED ISSUE #716: feat(sim): adaptive dosing — honor system-reset events (EVID=3/4) [#391]
Follow-up to the adaptive-dosing safety guards in #699 (which **reject** subjects carrying a system reset rather than run them with the reset silently ignored). This issue is the real **support**.

## Problem
The reactive driver (`ode_predictions_adaptive_impl`) builds its integration break-points from decisions, doses, and observation times but **never reads `subject.reset_times`** (verified: zero references across the impl). An EVID=3/4 reset is therefore silently dropped — the compartment state fails to zero at the reset time, and every downstream prediction, monitored signal, and dose decision is wrong. Worse, the frozen-schedule replay verifier is **also** blind to resets (both engines ignore them), so the safety net passes a wrong result. This defeats #391 Part E; the audit rated it **HIGH**. #699 guards `subject.has_resets()` with a typed error.

## Proposed support
- Fold `reset_times` into the driver's break-point set and zero all compartment amounts when a reset boundary is crossed, mirroring the state-propagating event-driven analytical / ODE path that already honors resets.
- Make the frozen-replay verifier reset-aware so it can actually catch a divergence in this case (today it is blind, so it cannot backstop the feature).

Remove the `has_resets()` branch from `reject_unsupported_adaptive` once the reset-aware path is wired and anchored.

- Degenerate oracle: a non-reactive controller over a subject with a mid-horizon reset reproduces `simulate()` on the equivalent static regimen (which honors the reset), bit-for-bit.
- Frozen-schedule replay bit-equality, after the verifier is made reset-aware.

Tracked under #391.

TeunP:
## Why

Adaptive/feedback-dosing simulation (`simulate_adaptive` / `simulate_adaptive_from_spec`, epic #391) **rejected** any base subject carrying a system reset (EVID=3/4) with a typed error — the reactive driver never applied the reset, so leaving it in would silently fail to zero the compartments. Real TDM / MIPD workflows are episodic (a reset starts a fresh course), so this was a genuine gap, not just a papered-over corner. This is the *support* follow-up to the #699 rejection guard; the audit rated it **HIGH**. Tracked under epic #391.

## What changed

The reactive driver and **both** frozen-replay engines now honor an EVID=3 reset carried by the base subject, mirroring the static event-driven engine (`ode_predictions_event_driven`):

- **`integrate_segment` gains a `reset_floor` param** (all 6 call sites updated). Infusions / zero-order windows opened before the most-recent reset are turned off. Reset-free callers pass `f64::NEG_INFINITY`, so their active set is unchanged — and the prediction dispatcher already routes reset subjects off the plain `ode_predictions` path, so those callers are byte-identical.
- **Three reset-application sites** — the reactive driver (`ode_predictions_adaptive_impl`), the shared constant-path replay engine (`ode_predictions_with_extra_breaks_and_stats`), and the per-event replay (`adaptive_frozen_replay_tv`) — add reset times to their break set, zero the compartments (re-seeding any `init(state)=expr` at the LOCF PK) at each reset break, and thread the running `reset_floor` into `integrate_segment`. The reset is applied **before** the decision hook / observation, so it sorts ahead of a dose or obs at the same instant (Reset < Dose < Obs). This makes the **frozen-replay verifier reset-aware**, so the reset is validated each run (#391 Part E) rather than passing a wrong result.
- **`auc_target` on a reset subject is rejected** — but for a subtler reason than the TV/IOV deferral. `ode_dense_solve_states` *is* reset-aware (it breaks at each reset and re-seeds the state), so windows entirely before or after a reset integrate exactly. The gap is the trapezoid: `adaptive_window_signal_aucs` integrates each inter-decision window on a *uniform* grid, and a reset is a state discontinuity at an arbitrary time, so the one window straddling it would need pre-/post-reset grid nodes (rt⁻, rt⁺) to integrate the jump correctly. Rather than report that one window's AUC biased low, defer. Every other adaptive output is reset-aware.
- **EVID=4 (reset+dose)** stays rejected by the dose-free-base precondition (it carries a dose), so only pure EVID=3 resets reach the adaptive path.
- The `has_resets()` branch is removed from `reject_unsupported_adaptive`.

## Alternatives considered

- **Exact-bit reset detection + a survival guard (`Err`)** — matches the existing #700 tv-covariate guard, but the non-`Result` replay engine (`ode_predictions_with_extra_breaks_and_stats` returns `Vec<f64>`) could only `panic!` on a sub-1e-15 grid collision. Chose **tolerance detection** (`RESET_MATCH_TOL = 1e-12`): resets are coarse episode boundaries, so a tolerance match cannot alias two distinct resets, and a reset merged into a sub-1e-15 neighbour by the break dedup is still applied at that representative break — no panic, no fragility. (Decisions / observations keep exact-bit lookups — they key `HashMap`s — hence the retained #700 guard, extended here to the constant path so a reset can't shadow a decision's exact-bit lookup.)
- **Supporting `auc_target` on reset subjects now** — the dense solver already applies resets, so the only missing piece is placing grid nodes at each reset instant (pre-/post-reset) in `adaptive_window_signal_aucs` so the straddling window's trapezoid resolves the discontinuity. Deferred as a focused follow-up (reject `auc_target`+reset for now) to keep this PR's scope on the driver/replay reset semantics.

## Cross-repo dependency

No `pub` API change (`reject_unsupported_adaptive` is `pub(crate)`; `simulate_adaptive` / `AdaptiveSimulateOptions` unchanged), so no ferx-r pin bump is required. The behavior reaches R users on the next routine pin bump; an optional `NEWS.md` line could note it.

## Breaking changes
- [x] None

## Numerical validation

Per **CLAUDE.md / epic #391**, feedback dosing has no NONMEM equivalent, so it is validated against the engine's own trusted paths and mrgsolve — not NONMEM.

- [x] Compared against: the event-driven engine via `predict()` (which routes reset subjects to `ode_predictions_event_driven`), on the ledger-realized regimen.
- Degenerate oracle (`adaptive_reset_matches_static_predict`): a fixed-dose controller over a dose-free base subject with a mid-horizon EVID=3 reset reproduces `predict()` on the same realized doses + reset, across a bolus regimen. Cross-engine tolerance (dense reactive driver vs event-driven `predict()`, ~1e-6 relative); a dropped/mis-applied reset moves a prediction by O(dose).
- Reset-floor on a controller-issued infusion (`adaptive_reset_turns_off_spanning_infusion_matches_static_predict`): an infusion spanning the reset is turned off from the reset on, matching `predict()`.
- Positive control (`adaptive_reset_zeros_state_positive_control`): reset vs no-reset diverge at a post-reset observation (0 vs ~18), proving the zeroing is real (the oracle is not vacuous).
- **IOV × reset oracle** (`adaptive_iov_reset_matches_predict_iov_and_zeros_state`): exercises the third reset site — `adaptive_frozen_replay_tv` — which the constant-model oracles above do not reach (only an IOV / TV-covariate subject drives `event_pk = Some`). A mid-horizon reset on a real IOV model reproduces `predict_iov` (an **independent** third engine, routed to the reset-aware event-driven walker) on the realized doses + reset, with a no-reset positive control at t=48. This both validates the IOV replay reset path and closes the patch-coverage gap the reviewer flagged.
- **Collision guard** (`adaptive_reset_decision_collision_is_rejected`): a decision within 1e-15 of a reset (adjacent doubles at t=1) is a typed error, not a silently dropped decision.
- The default-on reset-aware frozen-replay verifier runs in every oracle above; its `Ok` is part of each assertion.
- **Mutation-verified**: disabling the driver's state-zeroing fails the oracle tests — via the verifier catching `reactive≈18` vs `static=0`, confirming both the tests and the verifier's reset-awareness have teeth.

## Performance
- [x] No significant impact expected — reset-free subjects take an empty `reset_times` fast path (`NEG_INFINITY` floor, no extra breaks); byte-identical to before.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 2714 pass / 0 fail)
- [x] Regression test added that fails without this fix (the positive control + oracles; mutation-verified)

- [x] Not applicable as a standalone file — the change is a simulation-API behavior (a dataset EVID=3 row combined with an `[adaptive_dosing]` block now runs instead of erroring). Documented in `docs/model-file/adaptive-dosing.qmd`.

## Docs
- [x] `docs/` updated — `docs/model-file/adaptive-dosing.qmd` "Current scope and limits" now documents reset support (and the `auc_target`+reset deferral).
- [x] `CHANGELOG.md` `[Unreleased]` entry added (Added).

## Checklist
- [x] `cargo clippy` clean — CI-exact `cargo clippy --no-default-features --features ci,markov` finishes with no errors and no new warnings on the changed lines.
- [x] `Dual2` sensitivity path unaffected — `integrate_segment` is on the `f64` simulate/predict path, not a `*_g<T: PkNum>` gradient kernel; no gradient-reachable code changed.
- [x] `Cargo.toml` version bump — not needed (non-breaking).

## Reviewer hints

- The three reset-application sites must agree on ordering (**Reset < Dose < Obs**) and on the reset PK snapshot (LOCF `pk_readout` in the driver, `pk_params_flat` on the constant replay, `last_pk` in `adaptive_frozen_replay_tv`). The frozen-replay verifier cross-checks the driver against the replay each run, so a mismatch surfaces loudly.
- `RESET_MATCH_TOL` tolerance detection vs exact-bit (see *Alternatives*) — the constant-path decision-survival guard is the exact-bit backstop for decisions.
- The `auc_target`+reset rejection is a deliberate deferral: `ode_dense_solve_states` under the AUC pass *is* reset-aware, but its uniform per-window trapezoid grid can't resolve the reset discontinuity in the straddling window without pre-/post-reset nodes. (Not the frozen-snapshot reason of the TV/IOV deferral.)

## Open questions

- Reset-aware `auc_target` is deferred. The dense solver already applies resets; the follow-up is only to place pre-/post-reset grid nodes in `adaptive_window_signal_aucs` so the straddling window's trapezoid resolves the discontinuity. Fine to leave for a focused follow-up?


TeunP:
## Review — EVID=3 reset in adaptive dosing (xhigh)

Reviewed the full diff against the surrounding code, **empirically probed** the one claim everything hinges on, audited the parallel metric funnels, and cross-checked with an independent reviewer. The shipped **behavior is correct** — reset zeroing, the infusion floor, the `Reset < Dose < Obs` ordering, and the EVID=4 rejection all hold. Three issues below (being fixed in a follow-up commit on this branch).

### Findings

| # | Sev | Location | Problem | Fix |
|---|-----|----------|---------|-----|
| **F1** | 🟠 MED | `api/adaptive.rs` auc-guard comment + error + PR "Alternatives" | **Rationale is factually wrong.** It says the AUC pass "seeds the state once and **never zeros it at a reset** … integrated as if the reset never happened", and that supporting it needs "making `ode_dense_solve_states` reset-aware". The opposite is true: `adaptive_window_signal_aucs` clones the base subject (keeps `reset_times`) and calls `ode_dense_solve_states`, which **is** reset-aware — `build_segment_break_times` adds resets and `apply_segment_boundary` re-seeds (since `3f58c7d3`). Probe: state after a t=5 reset is `t=6 → 0.0, t=10 → 0.0`. The rejection is *safe* (errors, no bad number), but the reason is false. The genuine residual concern is only the uniform per-window trapezoid grid not resolving the reset discontinuity in the single straddling window (windows before/after are exact). | Correct the rationale to the real reason (keep the deferral — doing it right needs pre/post-reset grid nodes, genuinely larger). |
| **F2** | 🟠 MED | `predictions.rs` `adaptive_frozen_replay_tv` reset block | **Third reset site untested → `codecov/patch` RED (86.57%, 9 missing).** That block runs only when `event_pk = Some` (tv-cov / IOV); every new test uses the constant `ODE_NO_IIV` model, so it routes to the *constant* replay. `subject_needs_per_event_pk` does not flag resets, so reset+IOV / reset+tv-cov **is allowed** yet unverified — including the `last_pk` snapshot at a reset landing on a non-decision break. | Add a reset+IOV oracle vs `predict_iov` (reset-aware) — covers the lines and validates driver↔tv-replay. |
| **F3** | 🟡 LOW | `predictions.rs:2959` collision guard | Driver's decision↔reset collision `Err` branch is untested (part of the 9 missing). *(The analogous obs-shadowing does not exist — observations are recorded via per-segment `saveat` + exact match on the solver's returned point, not a break-boundary lookup, so the decision guard is correctly scoped. Verified.)* | One test hitting the `Err`. |

### Verified correct (checked, not assumed)

- **Floor `start >= reset_floor`**: a dose issued *at* the reset instant is kept, one started before is dropped — correct for `Reset < Dose`, consistent across `active_infusions` / `active_zero_order_inputs` / `wrap_rhs_with_forcings`.
- **EVID=4 safely rejected**: it populates `subject.doses` (datareader) → the dose-free-base precondition (`predictions.rs:2750`) rejects it before the reset path.
- **Ordering `Reset < Dose < Obs`** at all three sites (reset applied at the top of the loop iteration).
- **All 6 `integrate_segment` call sites** updated; reset-free callers pass `NEG_INFINITY` → byte-identical.
- **Parallel-funnel audit**: `target_window` is ledger-derived (reset-aware); `auc_target` is the only separate solve (F1). `|| s.has_resets()` is **necessary**, not redundant (`subject_needs_per_event_pk` doesn't flag resets).
- **CI**: Check / Clippy / Format / Test / Survival / Markov + the Coverage *job* green; only the `codecov/patch` *gate* is red (= F2/F3).

Behavior is sound; F1 (false rationale) and F2 (green the patch gate + verify reset+IOV) are worth fixing before merge. Fixes incoming on this branch.

Issue #932, opened 2026-07-25
feat(sim): adaptive base regimen — support system resets / EVID=4 [#702]


TeunP:
Follow-up to #702. A base regimen combined with a **system reset** — EVID=3, or an EVID=4 reset+dose row — is a typed error today (the combo guard in `ode_predictions_adaptive_impl`).

## Support
Compose #716's reset-aware machinery (state-zeroing + the `reset_floor` that turns off infusions opened before a reset) with #702's base-dose seeding:
- A base infusion spanning a reset must be turned off at the reset (the `reset_floor` already does this for controller infusions; extend it to base infusions).
- An EVID=4 (reset+dose) row must reach the adaptive path — today its dose trips the base-regimen combo guard before the reset is considered.

## Validation (per CLAUDE.md — no NONMEM for adaptive)
- Degenerate oracle: base regimen + mid-horizon EVID=3 reset + controller reproduces `predict()` on the realized regimen carrying the same reset.
- Positive control: a base infusion spanning the reset is turned off (diverges from a no-reset run), so the oracle isn't vacuous.

Tracked under #702.

Issue #931, opened 2026-07-25
feat(sim): adaptive base regimen — support inter-occasion variability [#702]


TeunP:
Follow-up to #702. A base regimen combined with **inter-occasion variability** (`kappa`) is a typed error today (the combo guard in `ode_predictions_adaptive_impl`).

## Support
Per-occasion κ bookkeeping for the pre-scheduled doses (occasion = decision window, the #701 convention), plus the IOV frozen-replay path carrying the base doses' `F`. Builds directly on the base × time-varying-covariate follow-up, since the IOV replay shares `adaptive_frozen_replay_tv` — land that first.

## Validation (per CLAUDE.md — no NONMEM for adaptive)
- Full-stack `predict_iov` reconstruction oracle (à la #701) with a loading dose: reconstruct the exact per-occasion κ and confirm the base + reactive trajectory matches.
- Extend the `tests/reference/vanco_iov_mrgsolve/` anchor with a loading dose.

Tracked under #702.

Issue #702, opened 2026-07-06
feat(sim): adaptive dosing — allow a pre-scheduled base regimen (loading dose) [#391]


TeunP:
Follow-up under #391 — noted in the `ode_predictions_adaptive_impl` docstring as "augmenting pre-scheduled doses is a later step", and reinforced by the dose-free-base guard hardened in #699.

## Problem
The reactive driver requires a **dose-free base subject** (`ode_predictions_adaptive_impl` rejects any non-empty `subject.doses`). But real TDM / MIPD workflows start from a **loading / maintenance regimen** and layer adaptive decisions on top (e.g. start vancomycin at a fixed mg/kg, then titrate on measured levels). Today that base regimen is a typed error, so the whole regimen must be controller-driven.

## Proposed support
Allow the base subject to carry pre-scheduled doses (and steady-state), integrate them, and augment with controller decisions at the decision schedule. The shadow-dose machinery already exists; the change is to seed it from `subject.doses` instead of requiring it empty, keeping the break-time timeline and TAD/TAFD anchors correct across pre-scheduled + injected doses.

## Validation (per CLAUDE.md — no NONMEM for adaptive)
- Degenerate oracle: a fixed base regimen with a controller that issues no doses reproduces `simulate()` on that regimen bit-for-bit.
- Reproduce an mrgsolve TDM example that combines a loading regimen with reactive titration.

Tracked under #391.

PR #937, opened 2026-07-26
feat(sim): pre-scheduled base regimen under time-varying covariates in adaptive dosing [closes #930]

### LINKED ISSUE #930: feat(sim): adaptive base regimen — support time-varying covariates [#702]
Follow-up to #702, which shipped pre-scheduled base regimens (loading / maintenance doses) in adaptive dosing on **constant-covariate, non-reset** models. A base regimen combined with a **time-varying covariate** is currently a typed error (the combo guard in `ode_predictions_adaptive_impl`).

## Why it's a separate slice
On the TV path (#700) the driver threads a per-segment `event_pk` snapshot, and the frozen-replay verifier's TV engine (`adaptive_frozen_replay_tv`) reads each dose's bioavailability `F` **from the realized ledger only**. A pre-scheduled base dose isn't in the ledger, so the TV replay has no `F` for it — which is why #702 scoped this combo out rather than ship it half-verified.

## Support
- Seed the base regimen into the driver's `shadow` on the TV path (the driver's base-dose seeding from #702 is already TV-agnostic; the per-segment PK recompute is #700's).
- Extend `adaptive_frozen_replay_tv` to carry the base doses' `F` alongside the ledger's `f_applied`, so the replay integrates the base regimen with the same `F` the driver used.
- Drop the `tv` branch from the base-regimen combo guard once wired + anchored.

- Degenerate oracle: base regimen + TV covariate + no-op controller reproduces the event-driven engine on that regimen.
- Extend an mrgsolve renal-decline anchor (à la `tests/reference/vanco_renal_mrgsolve/`) with a loading dose.

Tracked under #702.

TeunP:
## Why

Follow-up to #702, which shipped pre-scheduled base regimens (loading / maintenance doses) in adaptive-dosing simulation on **constant-covariate** models only. A base regimen combined with a **time-varying covariate** (declining renal function, a `TIME`-driven parameter) was a typed error — #702 scoped it out rather than ship it half-verified, because the TV frozen-replay engine (`adaptive_frozen_replay_tv`) read each dose's `F` from the realized ledger alone, and a pre-scheduled base dose isn't in the ledger. This lifts that restriction for plain bolus / infusion base doses — the real MIPD workflow of a loading dose given while renal function declines.

## What changed

`ode_predictions_adaptive_impl` (`src/ode/predictions.rs`):

- **Combo guard** drops the `tv` branch — a base regimen is now accepted with a time-varying covariate. Base × IOV (#931) and base × system-reset (#932) stay rejected.
- **Per-dose F.** On the TV path each base dose's bioavailability `F` is resolved from its **own** covariate snapshot (`event_pk.dose[k]`, the covariate active at the dose's administration time), symmetric with a controller-issued dose whose `F` is fixed from the LOCF PK at injection — instead of the frozen t=0 `pk_params_flat` the constant path uses. The dose is then integrated under the per-segment PK (#700).
- **TV sub-guard.** A base dose that is steady-state, lagged, into a built-in input-rate (transit / zero-order absorption) compartment, or carries a modeled (coded) `RATE` is a typed error under a time-varying covariate — its per-dose SS / lag / input-rate / rate-resolution bookkeeping needs threading through the hand-rolled TV replay (a #930 follow-up). The default-on frozen-replay verifier is the backstop.
- **Verifier** (`verify_adaptive_frozen_replay`): the TV branch builds `dose_f` = base doses' `F` (recomputed from each dose's own covariate snapshot, identically to the driver) ++ ledger `f_applied`, so the replay integrates the base regimen with the same `F` the driver used.

Constant-covariate and dose-free paths are byte-identical (221 adaptive + 148 ODE-prediction unit tests unchanged).

## Alternatives considered

Freezing base `F` at the t=0 baseline (the `pk_params_flat` the constant path uses) was simpler but silently wrong when `F` depends on the time-varying covariate — a mutation test pins exactly this (a stale t=0 `F` injects 800 units where the covariate-at-dose `F` injects 480). Resolving `F` per-dose from `event_pk.dose[k]` (already built by `compute_event_pk_params_into`, parallel to the base doses) mirrors how controller doses already work.

## Cross-repo dependency

No public Rust API change (both entry points already accepted a base subject; only the accepted covariate combination widened).

## Breaking changes
- [x] None

## Numerical validation

Per CLAUDE.md, adaptive/feedback dosing uses **mrgsolve** (NONMEM has no native feedback dosing), degenerate oracles, and the frozen-replay verifier — not NONMEM.

- **Independent closed-form oracles** (Tier-1): a loading bolus decaying under a declining CL; a base dose whose `F` drifts with the covariate — **mutation-verified** (the covariate-at-dose `F` injects 480 units; a stale t=0 `F` injects 800, and the test fails on that mutation); and a base **infusion** under per-segment CL.
- **Reactive** base + titration under a time-varying covariate passes the default-on frozen-replay verifier (driver vs `adaptive_frozen_replay_tv`, bit-exact); scope rejections for SS / lagged base doses.
- **mrgsolve renal-decline + loading anchor** (`tests/reference/vanco_renal_loading_mrgsolve/`, mrgsolve 1.7.2): a 1500 mg loading bolus rides on the subject, CRCL declines 120 → 41 mL/min, and the controller titrates daily maintenance. ferx reproduces the ladder **dose-for-dose** and every trough within a small cross-solver tolerance (RK45 vs LSODA), reactive in both directions (3 increases, 3 decreases).

```
# mrgsolve reference ladder (day, CRCL, trough mg/L, dose mg) — ferx matches
0  100  5.647391   937.5        increase
2   72  8.813241  1464.84375    increase
4   54 16.320448  1098.6328125  decrease
7   41 14.528939   617.98095703 reissue
```

- [x] Compared against: independent mrgsolve run + hand-computed closed forms

## Performance
- [x] No significant impact expected (constant/dose-free paths byte-identical; the TV base-dose work is O(n_base) F lookups)

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 2732 pass / 0 fail)
- [x] Regression test added that fails without this fix (per-dose-F mutation test; scope rejections)

- [x] Example added: `examples/adaptive_vanco_renal_loading.ferx` (loading dose + declining CRCL + titration)

## Docs
- [x] `docs/model-file/adaptive-dosing.qmd` updated (base-regimen scope)

## Checklist
- [x] `cargo clippy` clean (0 new warnings)
- [x] Functions on the `Dual2` sensitivity path stay differentiable (not touched — simulation-only path)
- [x] `Cargo.toml` version bumped if breaking (N/A — non-breaking)


TeunP:
## Code review — xhigh

Reviewed the full diff at `xhigh` depth (no happy-path passes), tracing the load-bearing invariants to source rather than trusting CI-green.

### Correctness — the invariant that could bite

The change introduces a new 4-way parallel index in `ode_predictions_adaptive_impl`: `resolved_base.doses[k]` ↔ `subject.doses[k]` ↔ `base_lagtimes[k]` ↔ `ev.dose[k]`. A wrong-but-consistent misalignment here would silently mis-integrate a delivered dose, so I traced each list to ground:

- **`resolve_subject_doses` never adds / drops / reorders / re-compartments** — it only swaps `rate` / `duration` / `rate_mode` in place (`dosing.rs:57-64`); `cmt` and `ss` survive `resolve_rate`'s `..self.clone()` spread. And the `!is_fixed()` guard rejects the *only* input that would take the owned/mutating branch, so every input that reaches the F-recompute is `Cow::Borrowed` — i.e. `resolved_base.doses` **is** `subject.doses` (object identity). Alignment is unconditional.
- **`event_pk.dose[k]`** is built by a straight `for k in 0..subject.doses.len()` push (no sort / dedup / filter), each entry using `dose_cov(k)` at `doses[k].time` — the correct dose-administration-time snapshot. Length `== subject.doses.len()` on the read path; the one empty-`dose` constructor is the IOV one, fenced off by the base×IOV reject upstream.
- **No panic:** `tv == event_pk.is_some()` verbatim, so the `.expect()` is unreachable; `f_bio` reads via a bounds-checked `.get(slot)`.
- **Verifier symmetry:** the replay recomputes base F from the *same* `ev.dose[k]`, so any snapshot / index divergence surfaces as a replay `Err`, never a silent wrong answer. (The verifier only proves driver == replay; absolute F is pinned independently by the closed-form oracle `F = 0.48 → 480`. That split is correct and both halves are present.)
- Re-derived all three closed-form oracles independently: loading `[1000, 301.19, 90.72]` at k=0.05; per-dose F `480 → 43.55` at k=0.1; infusion `475.81` then k=0.06 decay. ✓

No correctness, alignment, length, or panic hazard.

### Findings — all addressed in 256eafa1

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | low | Stale comment: the base-regimen break collection now runs on the time-varying path too (`if n_base > 0`), but the comment said "constant path only … no per-event-PK interplay". Code correct, comment wrong. | Comment corrected to describe the TV path (base infusion-end breaks from the covariate-resolved F). |
| 2 | low | 2 of 4 TV-scope-guard arms untested — `!is_fixed()` (modeled / coded-RATE base) and `input_rate_consumes_cmt` (transit / zero-order base). Only `ss` + `lag` had reject tests. | Added `adaptive_base_modeled_rate_dose_with_tv_covariate_is_rejected` and `adaptive_base_input_rate_dose_with_tv_covariate_is_rejected` (fixtures `ODE_TV_MRATE` / `ODE_TV_ABSORB`), each mutation-verified to pin its own arm. |
| 3 | nit | `decay_day` defined but never called in the mrgsolve reference generator. | Removed. |

Mutation note for #2: dropping the `!is_fixed()` arm makes a coded-RATE base dose slip the guard and then **panic** (`is_real_infusion: unresolved modeled dose`); dropping the input-rate arm lets that dose reach a later controller guard with a different message. Both tests fail loudly without their arm — non-vacuous, and the guard demonstrably converts a downstream panic into a clean typed error.

### Remaining (optional, non-blocking)

- No closed-form test of covariate-dependent F on an *infusion* (bolus F is pinned; the infusion oracle runs F=1). The apply-path is shared with the constant path, so low risk.
- The per-dose-F oracle can't distinguish the `dose_cov` snapshot from `obs_cov` at t=24 (both CRCL=60); the critical mutation (t=0 baseline) *is* pinned.

**Verdict:** correctness is sound on the axes that could silently corrupt a delivered dose; the three findings were comment / test-coverage / cleanup and are fixed in 256eafa1.

Issue #748, opened 2026-07-08
Adaptive-dosing frozen-replay verifier is blind to precomputed shared inputs (decision_pk / eta_occ / event_pk)


TeunP:
## Context

The adaptive/feedback-dosing frozen-replay verifier (`verify_adaptive_frozen_replay` → `adaptive_frozen_replay_tv`) is the default-on safety net for `simulate_adaptive{,_from_spec}`: it re-runs the realized dose ledger through the static engine and asserts bit-alignment with the reactive driver.

## The gap

The verifier receives the **same precomputed snapshot arrays** the driver used — `decision_pk`, `eta_occ`, and `event_pk` — all built once in `run_adaptive_population` (`src/api.rs`). It validates that the driver and the replay **consume** those snapshots identically; it does **not** validate that the snapshots are **correct**. So any error in how a snapshot is *built* (wrong covariate fed to `pk_param_fn`, wrong occasion→κ mapping, wrong occasion assignment) is applied identically by both sides and passes bit-exact.

This is not hypothetical: the #701 self-review found exactly this class of bug — `decision_pk[g]` built with the frozen `t=0` covariate instead of LOCF for an off-record time-varying-covariate + IOV decision (fixed in cbc8d146 on PR #739, guarded by a targeted regression test). The verifier could not catch it; only a hand-written regression test could.

## Why it matters

Every precomputed shared input is a blind spot. Today they are backstopped piecemeal by (a) unit-tested single-source helpers (`locf_decision_cov`, `occasion_of`), (b) the `predict_iov` reconstruction oracle (for κ), and (c) targeted regression tests. That is per-bug, not systemic — a new adaptive feature can reintroduce a fixed defect in a new build site and slip past the default-on verifier again.

## Proposed direction

Make the frozen-replay verifier **independently re-derive** each decision's PK/covariate (and ideally the occasion κ) rather than trusting the passed arrays — e.g. thread `pk_param_fn` (or the `CompiledModel`) into the replay so it recomputes `decision_pk[g]` from `(theta, eta_occ[g], its-own-LOCF-covariate, t_g)`. Then a wrong precomputed snapshot makes driver ≠ replay, so the default-on verifier fails automatically for every model — not just cases a regression test happens to cover.

Requires a small layering change (the driver/replay currently operate over `OdeSpec`/`PkParams` and do not hold `pk_param_fn`). Scoped as a follow-up to #701.

_Found during #701 (adaptive IOV) self-review; see PR #739._

Issue #701, opened 2026-07-06
feat(sim): adaptive dosing — inter-occasion variability (IOV) support [#391]


TeunP:
Follow-up to the adaptive-dosing safety guards in #699 (which **reject** IOV models rather than run them silently with kappas held at zero). This issue is the real **support**.

## Problem
`run_adaptive_population` forces all kappas to zero: `eta_slice.resize(n_eta + model.n_kappa, 0.0)` ("the reactive driver is BSV-only in this slice"). An IOV model therefore runs adaptive dosing with inter-occasion variability entirely suppressed — biased predictions, reactive decisions, and dose ledger. Separately, controller-injected doses are pushed to `shadow.doses` with **no occasion index**, and `shadow.occasions` / `dose_occasions` are never maintained, so there is no occasion→κ mapping for dynamically-added doses. #699 guards `n_kappa > 0` with a typed error.

## Proposed support
- Draw κ per occasion (as the fit / simulate paths do via `iov_occasion_groups` + `subject.dose_occasions`).
- Add occasion bookkeeping for controller-injected doses — assign each realized dose an occasion index at injection time.
- Apply per-occasion κ inside the per-event PK recompute.

Depends on the per-event PK infrastructure from the time-varying-covariate support issue. Remove the `n_kappa > 0` branch from `reject_unsupported_adaptive` once wired and anchored.

## Validation (per CLAUDE.md — no NONMEM for adaptive)
- Degenerate oracle over an IOV model + an mrgsolve dynamic-dosing anchor.

Tracked under #391.

PR #938, opened 2026-07-26
feat(sim): support inter-occasion variability with a pre-scheduled base regimen in adaptive dosing [closes #931]

### LINKED ISSUE #931: feat(sim): adaptive base regimen — support inter-occasion variability [#702]

TeunP:
## Why

Follow-up to #930 on the adaptive-dosing base-regimen ladder (#700 TV → #701 IOV → #716 reset → #702 base → #930 base×TV → **#931 base×IOV** → #932 base×reset). A pre-scheduled base regimen (a loading / maintenance dose, #702) combined with **inter-occasion variability** (`kappa`) was a typed error. This is the real MIPD/TDM setup — start on a fixed loading dose and titrate on measured levels — for a model whose clearance (or F, or any individual parameter) jumps per occasion.

## What changed

**Root cause (not "drop the guard"):** `compute_event_pk_params_iov` left `event_pk.dose` **empty** — an assumption from when the adaptive base subject was always dose-free. Three consumers read `event_pk.dose[k]` for a base dose's `F`: the driver F-block, the frozen-replay verifier, and the #748 snapshot check. So base×IOV would have panicked / failed there.

- **Fill `event_pk.dose` per-occasion** (`src/pk/mod.rs`): each base dose carries the PK of the occasion (decision window, #701) active at its administration time, so `f_bio` reads its `F` under that occasion's κ (`f_bio` reads `values[PK_IDX_F]`; `pk_param_fn` folds κ into it). Dose-free subjects → empty → byte-identical to the pre-#931 path.
- **Combo guard** (`src/ode/predictions.rs`): `(iov || has_resets)` → `has_resets` only. base×reset (#932) stays rejected; the driver F-block and verifier then work unchanged (they already read `event_pk.dose[k]`, gated on `event_pk.is_some()`, true under IOV). Scope restrictions (SS / lag / built-in input-rate / modeled-`RATE` base doses) unchanged.
- **#748 snapshot check** (`src/api/adaptive.rs`): added an independent per-dose re-derivation. The driver **and** the frozen-replay verifier both read `event_pk.dose[k]`, so a build-loop error there is applied symmetrically and the replay cannot catch it — this check is the backstop. It also closes the same latent blind spot for #930's base×TV (its dose snapshots were length-checked but never independently re-derived).

Loading dose administered before the first decision → baseline window (κ = 0), per the #701 convention.

## Alternatives considered

Computing the base-dose `F` directly in the driver via `decision_pk` + occasion, instead of through `event_pk.dose`. Rejected: it duplicates logic across the driver and the verifier and needs a parallel verifier change, whereas filling `event_pk.dose` is the single source of truth both already consume (the exact shape #930 used for base×TV).

## Cross-repo dependency

No public API change (`compute_event_pk_params_iov` is `pub(crate)`; `EventPkParams.dose` field pre-existed).

## Breaking changes
- [x] None

## Numerical validation

Per CLAUDE.md, adaptive/feedback dosing does not use NONMEM (it has no native feedback dosing). Validated with independent oracles + mrgsolve:

- [x] Compared against: **mrgsolve 1.7.2** (`tests/reference/vanco_iov_loading_mrgsolve/`) — dose-for-dose, and the independent **`predict_iov`** engine.
- [x] Reference values:

```
# vanco loading (1500 mg) + per-occasion-κ titration, seed 20260726, 8 daily decisions
# ferx (RK45)  vs  mrgsolve (LSODA), per-decision (trough mg/L, dose mg):
dec0 t=24  6.8441 / 6.8440   937.5
dec3 t=96  9.5078 / 9.5076   1831.0547
dec4 t=120 18.4355/ 18.4354  1373.2910   (decrease)
dec7 t=192 11.9592/ 11.9589  1716.6138
# max trough |Δ| ~3e-4 (< SIGNAL_TOL 2e-3); realized doses identical (exact f64 titration)
```

Oracles (in-crate, default-on where noted):
- `adaptive_base_loading_under_iov_matches_predict_iov` — reactive driver == `predict_iov` on the reconstructed per-occasion κ, @1e-9.
- `adaptive_base_dose_f_under_iov_matches_closed_form` — closed form `F = TVF·exp(κ₁)`, **mutation-verified** (a frozen-κ=0 build gives 500 vs 135.79; the both-sides-baseline mutation fools #748 + frozen-replay, so only this closed-form oracle catches it).
- `adaptive_base_loading_plus_titration_under_iov_passes_verifier` — off-grid decisions, default frozen-replay bit-exact.
- `catches_corrupted_dose_snapshot` — the new #748 dose re-derivation rejects a corrupted snapshot.

## Performance
- [x] No significant impact expected (dose-free path byte-identical; base×IOV fills one extra per-dose PK snapshot, same as base×TV).

## Tests
- [x] Unit test(s) added (`cargo test --lib` passes — **2739 pass / 0 fail**)
- [x] Regression test added that fails without this fix (the flipped `adaptive_base_regimen_with_iov_is_rejected` → support oracles; the mutation-verified F-drift oracle)

- [x] Example added to `examples/adaptive_vanco_iov_loading.ferx`

## Docs
- [x] `docs/model-file/adaptive-dosing.qmd` updated (base-regimen scope now includes IOV; IOV section cross-references the base regimen)
- [x] `CHANGELOG.md` `[Unreleased]` entry added (+ #702 / #930 entries reconciled so the block is internally consistent)

## Checklist
- [x] `cargo clippy` clean (exit 0; no new warnings from the changed lines)
- [x] `Dual2` sensitivity path untouched (adaptive simulation only)
- [x] No version bump (non-breaking)

### Example execution
- [x] The example is an `[adaptive_dosing]` model driven by `simulate_adaptive_from_spec` (not the plain CLI `--data` path). Its parse + a seeded reactive run are exercised by the fast, non-slow-gated guard `vanco_iov_loading_example_parses_and_pins_the_scenario`; the dose-for-dose mrgsolve cross-check runs nightly under `--features slow-tests`.

## Reviewer hints

- The load-bearing change is `compute_event_pk_params_iov` filling `.dose`; everything downstream (driver F-block, verifier, #748 check) already consumed `event_pk.dose[k]` and needed only the guard/message generalization.
- The `predict_iov` oracle has a real subtlety documented in the test: `predict_iov` indexes `kappas` by `iov_occasion_groups` order (obs-occasion order, dose-only occasions appended last), **not** occasion id — so the oracle's observation grid must span occasions 0..N in order for the dose's occasion to land at the matching group index.
- The mrgsolve R loop is `renal_loading`'s, with `cl_at` built from the per-occasion κ instead of the covariate; the loading bolus decays over (0, 24] under occasion 0's CL (end-of-interval), not the baseline κ=0.

## Open questions

None.


TeunP:
## Code review — xhigh (self-review + 2 independent adversarial agents)

**Method.** Independent line-level trace of the full diff, plus two adversarial subagents run in parallel — one hunting for a silent wrong-answer, one mutation-testing whether the new oracles actually bite (introducing production bugs and confirming which test catches each). CI-green was **not** treated as evidence of correctness.

**Verdict: no correctness bug.** The base × IOV mechanism is sound and triangulated by three near-independent engines — the driver `==` its frozen-replay static engine (bit-exact), the driver `==` `predict_iov` for valid configs (@1e-9), and the driver `==` mrgsolve 1.7.2 (dose-for-dose). All four headline oracles are mutation-verified to bite. The findings are documentation staleness, one test-rigor tightening, and coverage gaps — all resolved in the follow-up commit.

**Verified correct (not taken on faith):**
- `iov ⟹ tv` always holds, so the base-F block runs under IOV; there is no `iov && !tv` production path.
- `event_pk.dose[k] ↔ resolved_base.doses[k]` index parallelism holds (`resolve_subject_doses` is 1:1, never adds/drops/reorders).
- The occasion-corrected `F` reaches **both** delivery paths — the bolus jump (`injected_f`) and the infusion (`active_infusions`) — not computed-and-discarded.
- base × IOV × reset stays rejected (an EVID=4 row populates both `doses` and `reset_times`).
- `auc_target` × IOV is unconditionally rejected, so the frozen-snapshot AUC pass can never run under IOV.
- The dose-free path is byte-identical (empty loop, guard skipped) — consistent with the 2739 unrelated tests unchanged.

**Findings + resolutions:**

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| F1 | low | A base dose *before the first decision* resolves IOV-on-`F` at baseline κ=0 (untested boundary; its decay still uses occasion-0 CL — an F-vs-decay split). Convention-dependent, not a bug. | Confirmed baseline is the *consistent* convention (obs / EVID=2 already use baseline before the first decision); pinned it with `adaptive_base_dose_f_before_first_decision_uses_baseline_kappa` and documented it. |
| F2 | low | Stale `simulate_adaptive` docstring still said base × IOV/TV is a typed error. | Fixed. |
| F3 | low | Stale precondition comment ("only the constant, non-reset path reaches here … single frozen `pk_params_flat` governs both"). | Fixed. |
| W1 | low | Test B's non-vacuity guards (`> 1e-6`) were ~800× looser than the closed-form band (~8e-4) — non-vacuity rested on the pinned seed, not the guard. | Tied both guards to the band (10× margin). |
| E1 | med | Base **infusion** under IOV was claimed-supported but had zero coverage (all four IOV tests used boluses). | Added `adaptive_base_infusion_under_iov_matches_predict_iov` (an F-reshaped infusion validated vs the independent engine). |
| E2/E3 | low-med | Multi-dose / later-occasion base doses were untested (`n_base = 1` everywhere; the k ≥ 2 index-alignment trap unexercised). | Added `adaptive_base_multiple_doses_across_occasions_pass_verifiers`. |

**Subtlety surfaced while adding E3 (documented, not a bug).** The driver assigns a *segment's* occasion from the last obs / EVID=2 record it crossed (`segment_occ_at` — the #701 decision-window convention: the clearance in effect *during* the segment, the physically faithful choice for real-time feedback). `predict_iov` uses the #104 OCC end-of-interval rule (a segment ending at a *dose* takes that dose's occasion). The two agree when every segment ends at an observation (Test A, the mrgsolve anchor), but differ for a base dose at an occasion boundary with no coincident observation — so `predict_iov` is not a valid oracle there. Now documented in `compute_event_pk_params_iov` (a second `predict_iov` divergence alongside the already-documented pk-only one); the multi-dose test validates via the frozen-replay + #748 verifiers instead.

**Acknowledged, no action:**
- **W2 (low)** — the mrgsolve "dose-for-dose" check is partly self-referential (`REF_LADDER` doses are ferx's own realized ladder), but the troughs *are* cross-checked vs LSODA at `SIGNAL_TOL = 2e-3` (~7× the observed ~3e-4 gap; troughs span 4.35–18.44 mg/L, so a dropped / κ=0-frozen base regimen moves them far past it).
- **E4 (low)** — only the SS reject arm is tested under IOV; lag / input-rate / modeled-`RATE` rely on the shared `if tv && n_base > 0` guard (TV/IOV-agnostic), which the SS-under-IOV test confirms the IOV path routes into.
- **E5 (low)** — base dose after the last decision under IOV: shared logic with the shipped #702/#930 TV analogue.

**Evidence:** full `cargo test --lib` green (2742 pass, 0 fail); the mrgsolve anchor green under `--features slow-tests`; the F-drift and #748 dose-check oracles mutation-verified. Follow-up commit: `ba4ffe0c`.

PR #939, opened 2026-07-26
feat(sim): compose a pre-scheduled base regimen with system resets (EVID=3/4) in adaptive dosing [closes #932]

### LINKED ISSUE #932: feat(sim): adaptive base regimen — support system resets / EVID=4 [#702]

TeunP:
## Why

Follow-up to #702. A pre-scheduled **base regimen** combined with a **system reset**
(EVID=3, or an EVID=4 reset+dose row) was a typed error in `ode_predictions_adaptive_impl`
— the base-regimen combo guard rejected the pair before the reset was ever considered. This
composes #716's reset-aware machinery with #702's base-dose seeding so a subject can start on
a fixed loading/maintenance regimen, carry a mid-horizon reset (a washout, a new course), and
still be titrated by the controller — the real TDM/MIPD case. Tracked under #702 / #391.

## What changed

The core turned out to be a **single guard line**: the reactive driver, the reset-aware
static verifier (`ode_predictions_with_extra_breaks`), and the reference event-driven engine
already compose base doses with resets — base doses live in `shadow.doses`, so `active_infusions`
already gates a base infusion by the `reset_floor`, and all three engines apply `Reset < Dose`
at a shared instant. The combo guard was the sole over-conservative barrier.

- Narrowed the guard from `!subject.doses.is_empty() && subject.has_resets()` to
  `… && tv`: base × reset is now supported on the **constant-covariate** path; under a
  time-varying covariate or IOV (`event_pk`/`eta_occ = Some`) it stays a typed error.
- **EVID=4** (reset+dose) records *both* a `reset_times` entry and a `doses` entry, with no
  marker distinguishing its dose from a #702 base dose — so it previously tripped the same
  guard. It now flows through as a base dose landing at the reset instant (zeroed first, then
  applied), matching `predict()`.

## Alternatives considered

- **Enable TV/IOV base × reset in this PR too.** The per-event-PK replay (`adaptive_frozen_replay_tv`)
  *is* reset-aware, so it nearly works — but its composition with a base regimen across a reset
  is not yet oracle-verified against the reference. Deferred as a follow-up, matching the
  one-axis-per-PR discipline of #930/#931.
- **Reject SS base × reset up front.** Gated it on the oracle instead; the SS × reset oracle
  passes (both engines treat an SS dose as re-establishing its own equilibrium), so it is in
  scope rather than rejected.

## Cross-repo dependency

No public Rust API change — `simulate_adaptive` / `simulate_adaptive_from_spec` signatures are
unchanged; only the set of inputs they accept without erroring widened.

## Breaking changes
- [x] None

## Numerical validation

Per CLAUDE.md, adaptive/feedback dosing does **not** anchor against NONMEM (NONMEM has no native
feedback dosing). Validated instead with degenerate oracles + the default-on frozen-schedule
replay verifier (the internal dose-bookkeeping anchor), all in `src/api/tests/adaptive_sim_tests.rs`:

- **Degenerate oracle** — base 500 mg loading + mid-horizon EVID=3 reset + fixed controller
  reproduces `predict()` (which routes reset subjects to the event-driven walker) on the realized
  (base ∪ ledger) regimen carrying the same reset.
- **Positive control** — a base *infusion* spanning the reset is turned off at the reset (t=18
  reads ~0 vs ≫0 without the reset), so the oracle is not two engines agreeing on an un-reset run.
- **EVID=4 oracle** — a reset+dose row reaches the path and its dose lands *after* its own reset.
- **SS base × reset oracle** — a steady-state base dose wiped by a mid-horizon reset still matches.
- **Boundary** — base × reset under IOV is still rejected with a typed error.

- [x] Compared against: prior ferx engine (`predict()` / event-driven) + frozen-replay verifier
- [x] Full lib suite: `cargo test --lib` → **2746 passed, 0 failed, 18 ignored**

## Performance
- [x] No significant impact expected (one added boolean term in a precondition)

## Tests
- [x] Regression test added that fails without this fix (the two prior `…_is_rejected` tests were
  converted to oracles; they fail against the un-narrowed guard)

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added; also corrected #716's now-stale "EVID=4 still
  rejected" note. `simulate_adaptive` rustdoc updated.
- [x] `docs/model-file/adaptive-dosing.qmd` updated — the *Pre-scheduled base regimens* and
  *System resets (EVID=3)* bullets now describe base × reset composing on the constant-covariate
  path (EVID=3/4), with time-varying-covariate / IOV base × reset still a typed error. (An earlier
  revision of this PR wrongly claimed there was no dedicated page.)

## Checklist
- [x] `cargo clippy` clean (only pre-existing `src/bin/generate_data.rs` dev-bin warnings)
- [x] Not on the `Dual2` sensitivity path (adaptive driver is `f64`-only)
- [x] No `Cargo.toml` version bump (non-breaking)

- [x] Not a standalone file — programmatic API. Sketch:

```rust
// Base subject: a 500 mg loading dose at t=0 AND a system reset at t=36.
let mut base = subject_with_doses(vec![DoseEvent::new(0.0, 500.0, 1, 0.0, false, 0.0)]);
base.reset_times = vec![36.0];                     // EVID=3 washout mid-horizon
let opts = AdaptiveSimulateOptions { decision_times: vec![24.0, 48.0], ..Default::default() };
// Previously: Err("… base regimen … not yet supported together with system resets …").
// Now: the reset zeros the state, the controller keeps titrating, the frozen-replay verifier passes.
let res = simulate_adaptive(&model, &pop, &params, 1, controller, &opts)?;
```

### Example execution
- [x] No example execution step needed (programmatic adaptive API; no `.ferx` / CLI / data change)

## Reviewer hints

- The whole production change is the guard at `ode_predictions_adaptive_impl` (`src/ode/predictions.rs`)
  — narrowed to `&& tv`. Everything else is tests + docs.
- Worth a look: the `&& tv` boundary (IOV implies `tv`, so both TV-covariate and IOV are caught —
  see `adaptive_base_regimen_with_reset_under_iov_is_rejected`), and the SS × reset gate decision.

## Open questions

- Fast-follow to lift base × reset under a time-varying covariate / IOV? The replay is already
  reset-aware; it needs the base+reset oracle on the per-event-PK path before the guard drops `&& tv`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,8 +71,23 @@ section of the SDLC for the versioning policy).
   `init(state)=expr`) and turns off controller-issued infusions opened before it, exactly
   as `predict()` / `simulate()` do, and the default-on frozen-replay verifier is
   reset-aware so the reset is validated each run. Previously a reset subject was rejected
-  with a typed error. (An EVID=4 reset+dose row is still rejected by the dose-free-base
-  rule, so only pure EVID=3 resets reach the adaptive path.)
+  with a typed error. (Initially only a pure EVID=3 reset on a **dose-free** base subject;
+  #932 below lifts that to a reset combined with a base regimen, including an EVID=4
+  reset+dose row.)
+- **Base regimen combined with a system reset (EVID=3 / EVID=4) in adaptive-dosing
+  simulation** (#932). `simulate_adaptive()` / `simulate_adaptive_from_spec()` now accept a
+  base subject that carries BOTH a pre-scheduled loading / maintenance regimen (#702) AND a
+  system reset — composing #716's reset machinery with #702's base-dose seeding on the
+  constant-covariate path. The reset zeros the compartments and turns off a base infusion
+  opened before it (the reset floor, previously applied only to controller-issued infusions),
+  and an EVID=4 reset+dose row's dose now reaches the adaptive path — landing after its own
+  reset (Reset < Dose) — instead of tripping the base-regimen guard. Validated by a degenerate
+  oracle (base regimen + mid-horizon EVID=3 + controller reproduces `predict()` on the realized
+  regimen carrying the same reset), a positive control (a base infusion spanning the reset is
+  turned off, so the oracle is not vacuous), an EVID=4 oracle, and a steady-state base × reset
+  oracle; the default-on frozen-replay verifier (reset- and base-aware) checks every run. Lifts
+  the #702/#716 `base × reset` restriction. Base × reset UNDER a time-varying covariate (#930)
+  or IOV (#931) remains a typed error — a #932 follow-up — never a silent mis-integration.
 - **Warning when a `combined` error model's additive initial estimate is negligibly
   small** (#847). A pre-fit check (`W_ADDITIVE_INIT_SCALE`) flags an additive SD start
   below 1% of the observation scale (median `|DV|`) on that endpoint. A near-zero
--- a/docs/model-file/adaptive-dosing.qmd
+++ b/docs/model-file/adaptive-dosing.qmd
@@ -360,11 +360,15 @@ slow-gated `tests/adaptive_vanco_anchor.rs` (nightly + on push to `main`).
   bioavailability `F` is resolved from its own snapshot — the covariate active at, and (under IOV)
   the occasion κ of, the dose's administration time (symmetric with a controller dose, whose `F` is
   fixed at injection) — and which is then integrated under the per-segment PK, matching independent
-  mrgsolve loading-dose runs (renal-decline #930, per-occasion κ #931). Still a typed error, never a
-  silent mis-integration of a delivered dose: a base regimen combined with system resets (EVID=3/4,
-  #932), and — under a time-varying covariate or IOV — a steady-state, lagged, built-in input-rate,
-  or modeled-`RATE` base dose (a #930 / #931 follow-up whose per-dose SS / lag / input-rate /
-  rate-resolution bookkeeping the frozen-replay engine does not yet reproduce).
+  mrgsolve loading-dose runs (renal-decline #930, per-occasion κ #931). A base regimen also
+  **composes with system resets** (EVID=3/4) on the constant-covariate path (since #932): the reset
+  zeros the state and turns off a base infusion opened before it, and an EVID=4 reset+dose row's dose
+  lands after its own reset — see *System resets* below. Still a typed error, never a silent
+  mis-integration of a delivered dose: a base regimen combined with a system reset **under a
+  time-varying covariate or IOV** (a #932 follow-up), and — under a time-varying covariate or IOV — a
+  steady-state, lagged, built-in input-rate, or modeled-`RATE` base dose (a #930 / #931 follow-up
+  whose per-dose SS / lag / input-rate / rate-resolution bookkeeping the frozen-replay engine does
+  not yet reproduce).
 - **Time-varying covariates are supported** — the driver recomputes PK per
   event/segment from the covariate active in that segment (the NONMEM
   end-of-interval convention that `predict()` / `simulate()` use), so a covariate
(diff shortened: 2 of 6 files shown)


TeunP:
## Self-review (xhigh) — resolved in `c5cf1027`

Line-by-line read of the diff plus an independent adversarial trace across all three reset-aware engines (reactive driver, dense static `ode_predictions_with_extra_breaks_and_stats`, and the event-driven walker `predict()` routes reset subjects to). Mutation-tested the oracles: reverting `&& tv` fails all five constant-path tests while the two rejection tests still pass — they have teeth.

### Verdict — the guard change is numerically correct

On the constant path a base × reset subject gets `event_pk = None` (`subject_needs_per_event_pk` excludes `has_resets`), so `tv == false` and it runs; under a time-varying covariate or IOV `event_pk = Some`, so it is rejected. All three engines share every reset primitive (`reset_floor` gating in `active_infusions`, `reseed_prescheduled_states_at`, `apply_prescheduled_boluses_at`, `initial_state`) and apply `Reset < Dose < Obs` over aligned break-times. Traced sound: base infusion spanning a reset (turned off, `start < reset_floor`), EVID=4 reset+dose coincident (dose lands on the zeroed state), SS dose landing exactly on a reset (`copy_from_slice` overwrite), reset coincident with a decision (collision guard fails loudly), non-zero `init(state)` reset.

### Fixed in `c5cf1027`

| Sev | Location | Issue |
|-----|----------|-------|
| Med | `docs/model-file/adaptive-dosing.qmd` | The *Pre-scheduled base regimens* and *System resets (EVID=3)* bullets still said base × reset (incl. an EVID=4 reset+dose row) is rejected — contradicting the shipped feature. Rewrote both; corrected the PR body's wrong "no dedicated page" claim. |
| Low | `predictions.rs` (reset-times break comment) | "Only pure EVID=3 resets reach here … base × reset is a follow-up" — falsified; an EVID=4 row + base now reaches this path. |
| Low | `predictions.rs` (SS-reseed rationale) | "a base × reset combo is rejected above so no reset intervenes" — a reset now can intervene; rewrote to explain the reseed stays correct (it overwrites with the SS equilibrium regardless of prior state). |
| Low | `predictions.rs` (`resolved_base` note) | "only a base × reset subject is rejected upstream" — now only under a time-varying covariate / IOV. |
| Low | `predictions.rs` (`verify_adaptive_frozen_replay` doc) | "dose-free subject … only its doses are replaced" — stale since #702 (the base regimen survives on the subject; the ledger is appended). |
| Test-gap | `adaptive_sim_tests.rs` | Only the IOV half of the `&& tv` boundary was pinned; a `&& tv` → `&& iov` regression would silently accept a time-varying-covariate base × reset subject. Added `adaptive_base_regimen_with_reset_under_tv_covariate_is_rejected`; tightened both reject assertions to `constant-covariate path only` (the old `constant-covariate` substring matched the previous message too, so it did not discriminate). |

`cargo test --lib`: **2747 passed, 0 failed, 18 ignored**.

### Out of scope — a pre-existing bug found during review, filed separately

`adaptive_window_signal_aucs` (`predictions.rs`) rebuilds each decision window's static subject from the **ledger only** (`sub.doses = ledger…`), so a constant-covariate subject carrying a **base regimen (#702) together with an `auc_target`** has its base (loading / maintenance) doses **dropped from the window AUC** — the target-attainment metric then underestimates exposure. The `auc_target` guard (`adaptive.rs:414`) rejects TV / IOV / reset subjects but **not** a plain constant-covariate base-regimen subject, so it is reachable. This predates #932 (the AUC pass was written assuming a dose-free subject; the #702 base-regimen work did not update it) and is unrelated to base × reset — left untouched here to keep this PR focused, flagged for a separate fix.

No public Rust API change; no ferx-r follow-up.

PR #942, opened 2026-07-27
fix(sim): integrate the pre-scheduled base regimen in the adaptive auc_target window-AUC pass [closes #940]

### LINKED ISSUE #940: adaptive: base regimen dropped from auc_target window AUC (exposure underestimated)
## Summary

In adaptive-dosing simulation, a subject that carries a pre-scheduled **base regimen** (#702 — a loading / maintenance dose on `subject.doses`) **together with an `auc_target`** has its base doses **omitted** from the per-decision-window AUC that drives target-attainment dosing. The exposure metric therefore underestimates AUC, which can make the controller over-dose.

Found during review of #939 (base × system reset). Pre-existing — predates #932; introduced when the #702 base-regimen work added `subject.doses` support without updating the AUC path. Unrelated to system resets.

## Root cause

`adaptive_window_signal_aucs` (`src/ode/predictions.rs`) rebuilds each window's static subject from the controller **ledger only**:

```rust
let mut sub = base_subject.clone();
sub.doses = ledger
    .iter()
    .filter(|e| e.time <= a + 1e-9)
    .map(|e| DoseEvent::new(e.time, e.amt, e.cmt, e.rate, false, 0.0))
    .collect();          // <-- REPLACES doses; a pre-scheduled base regimen is dropped
```

`sub.doses` is overwritten, so any base regimen on `base_subject.doses` is discarded for the AUC integration. The function's doc comment still describes `base_subject` as "the dose-free subject the run was driven from" — it was written pre-#702, assuming a fully controller-driven subject, where `sub.doses = ledger` is correct.

## Why it's reachable (no guard)

The only `auc_target` precondition guard (`src/api/adaptive.rs:414`) rejects **time-varying-covariate / TIME-in-PK / IOV / reset** subjects:

```rust
if auc_target.is_some()
    && (model.n_kappa > 0
        || population.subjects.iter().any(|s|
            crate::pk::subject_needs_per_event_pk(model, s) || s.has_resets()))
{ return Err(...) }
```

A plain **constant-covariate** subject carrying a base regimen is **not** rejected, so it reaches `adaptive_window_signal_aucs` and hits the drop.

Contrast `verify_adaptive_frozen_replay` (same file), which correctly **appends** the ledger to the base doses (`static_subject.doses.extend(...)`) — the pattern the AUC pass should follow.

## Fix direction

Fold the base regimen into each window subject alongside the filtered ledger — keep `base_subject`'s pre-scheduled `subject.doses` at/before the window's left edge, then append the ledger doses — preserving base-dose attributes (SS/II, lagtime, rate/infusion) that the `ledger → DoseEvent::new(..., false, 0.0)` reconstruction discards.

## Validation (adaptive dosing → no NONMEM, per CLAUDE.md)

- **Degenerate oracle**: a constant-covariate subject with a base loading dose + `auc_target` — window AUC matches `predict()` / the static engine on the (base ∪ ledger) regimen.
- **Positive control**: the base dose materially changes the window AUC (so the test is non-vacuous).
- **Revert check**: the oracle fails without the fix.

## Also update

- `adaptive_window_signal_aucs` doc comment ("dose-free subject").
- `docs/model-file/adaptive-dosing.qmd` `auc_target` bullet, if it implies base regimens compose with `auc_target`.

---
Found in #939; origin #702; family #391.

TeunP:
Closes #940.

## Why
`adaptive_window_signal_aucs` — the metrics-only signal-AUC pass behind `auc_target_attainment` (#391 S2.5b) — set its per-window static subject's doses to the controller **ledger only**, dropping any pre-scheduled base regimen (#702 — a loading / maintenance dose on `subject.doses`). On a base-regimen run combined with `auc_target`, the reported exposure — and hence `auc_target_attainment` — was biased low. Reachable on plain constant-covariate subjects: the `auc_target` guard (`api/adaptive.rs:411`) rejects only TV-cov / TIME-in-PK / IOV / reset, so a base-regimen subject slips through. `auc_target` is metrics-only (does not feed the controller), so the symptom is a misreported metric, not a control-loop error. Found during review of #939; out of scope there.

## What changed
- `adaptive_window_signal_aucs`: compose each window's static subject as base ∪ ledger, mirroring `verify_adaptive_frozen_replay` — base doses kept via `clone` (preserving `SS`/`II`/lag/`rate_mode`/infusion attributes a `DoseEvent::new` rebuild would strip), the ledger appended, then `retain(|d| d.time < b)`. `< b` also keeps a base maintenance dose landing strictly inside a window `(a, b)`; byte-identical to the old `<= a` for controller doses (which sit on decision-grid edges), so the pure-controller path is unchanged.
- Refreshed the pass's doc comments (window convention + the stale "dose-free subject" line, both pre-#702).
- Docs: `docs/model-file/adaptive-dosing.qmd` — the `auc_target` row and the base-regimen section now state the exposure metric integrates the base regimen.

## Alternatives considered
`<= a` (as first scoped in the issue) keeps only base doses at/before the window's left edge — correct for a loading dose, but still silently drops a base maintenance dose landing inside a window, a smaller variant of the same bug. `< b` fixes both at no cost on the controller path. The one residual is a ≈ ½·Δsignal·(window/panels) trapezoid node-placement bias for an *instantaneous bolus into the monitored compartment* landing mid-window (≈ 0.9 % in the test); zero for an infusion / absorption base dose (vancomycin, the canonical `auc_target` case, is infused → continuous), and far smaller than dropping the dose.

## Cross-repo dependency

(`adaptive_window_signal_aucs` is `pub(crate)`; no public API change.)

## Breaking changes
- [x] None

## Numerical validation
Adaptive dosing uses no NONMEM anchor (CLAUDE.md exception). Two Tier-1 degenerate-oracle tests, mutation-verified:
- `adaptive_window_signal_aucs_includes_base_loading_dose`: base IV loading bolus before the window; window AUC matched to the closed form to `1e-4`; positive control vs the dose-free subject (base adds ≈ 18 %).
- `adaptive_window_signal_aucs_includes_mid_window_base_dose`: base bolus strictly inside the window; AUC matched to closed form (2 % band — documented mid-window bolus node bias) and asserted ≫ ledger-only (ratio ≈ 3.3), which a `<= a` filter would collapse to 1.

Mutation results: reverting the fix (base dropped) fails both; flipping `< b`→`<= a` fails the mid-window test only, the loading test still passes.
- [x] Not NONMEM/nlmixr2 (adaptive-dosing exception)

## Performance
- [x] No significant impact expected (an extra `extend`+`retain` per window; the pass is already O(m²) and runs post-run only).

## Tests
- [x] Unit test(s) added (`cargo test --lib`: 2744 passed, 0 failed)
- [x] Regression test added that fails without this fix (mutation-verified)

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` Fixed entry added

## Checklist
- [x] `cargo clippy` clean on changed lines (CI invocation; the 4 `approx_constant` errors under `--tests` are pre-existing test code, off CI's clippy path)

## Reviewer hints
Focus on the window composition in `adaptive_window_signal_aucs`. SS / infusion base doses are covered by construction — cloning routes them through the same resolve / break-timeline / SS machinery as `predict()`; the tests use boluses for tight closed forms. The `< b` vs `<= a` choice is deliberate (see Alternatives).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,14 @@ section of the SDLC for the versioning policy).
   for any external code that constructs or exhaustively destructures these variants.
 
 ### Fixed
+- **Adaptive-dosing `auc_target` exposure metric now integrates the pre-scheduled base regimen**
+  (#940). The signal-AUC pass behind `auc_target_attainment` (#391 S2.5b) previously scored each
+  decision window on the controller's realized doses only, dropping any pre-scheduled base regimen
+  (#702 — a loading / maintenance dose), so on a base-regimen run the reported exposure — and hence
+  `auc_target_attainment` — was biased low. Each window now integrates the base regimen (a loading
+  dose before the window and a maintenance dose landing inside it) alongside the realized ledger,
+  matching `predict()` / `simulate()` on the combined regimen. Constant-covariate subjects only, as
+  before (time-varying-covariate / IOV / reset `auc_target` runs remain rejected).
 - **FOCE inner EBE recovery no longer discards a near-optimal partial for a worse fallback**
   (#378). When a subject's individual objective is multimodal (e.g. a 3-cpt IV proportional
   model with six IIV etas conditioned on ten points), the closed-form inner BFGS can reach the
--- a/docs/model-file/adaptive-dosing.qmd
+++ b/docs/model-file/adaptive-dosing.qmd
@@ -53,7 +53,7 @@ with the run, never with a fitted model's parameters.
 | `confirm` | | Debounce: act only after this many *consecutive* matches of the same rule (default `1` = act on the first breach). |
 | `levels` | | Discrete titration ladder `[d1, d2, …]` (oncology). With `levels`, bare `increase`/`decrease` step one rung; without it, use `increase N%` / `decrease N%`. |
 | `target_window` | | Therapeutic band `[low, high]` for the monitored signal, used **only** to report the `pct_time_in_window` outcome metric — it does **not** influence dosing (the `when` rules do). `high` may be `inf` for a one-sided "at or above `low`" target. Omit it to leave `pct_time_in_window` unreported, rather than guessing a band from the rule thresholds. |
-| `auc_target` | | Exposure band `[low, high]` for the **area under the monitored signal** over each inter-decision window (e.g. vancomycin AUC₂₄ when decisions are daily), used **only** to report the `auc_target_attainment` outcome metric — like `target_window` it does **not** influence dosing. `low` must be `≥ 0`; `high` may be `inf`. Declaring it turns on a signal-AUC pass that re-integrates the realized doses on a dense grid; omit it to skip that pass and leave `auc_target_attainment` unreported. |
+| `auc_target` | | Exposure band `[low, high]` for the **area under the monitored signal** over each inter-decision window (e.g. vancomycin AUC₂₄ when decisions are daily), used **only** to report the `auc_target_attainment` outcome metric — like `target_window` it does **not** influence dosing. `low` must be `≥ 0`; `high` may be `inf`. Declaring it turns on a signal-AUC pass that re-integrates any pre-scheduled base regimen (#702) together with the realized controller doses on a dense grid; omit it to skip that pass and leave `auc_target_attainment` unreported. |
 | `with_assay_error` | signal¹ | Titrate on the **noised measurement** of a model output (named by `assay_cmt`) instead of a latent expression (default `false`). The signal's value comes from that output's `[scaling]` readout and its σ from that output's `[error_model]`, so the two are always on the same scale — there is no re-typed `observe` expression to mis-scale. Mutually exclusive with `observe`. |
 | `assay_cmt` | | The 1-based compartment of the model output measured under `with_assay_error` — its `[scaling]` readout is the signal value and its `[error_model]` σ the noise. **Required** with `with_assay_error`. |
 
@@ -345,7 +345,12 @@ slow-gated `tests/adaptive_vanco_anchor.rs` (nightly + on push to `main`).
   through the same dose-resolution, break-timeline, and steady-state machinery as
   `predict()` / `simulate()`, and they appear in the controller's dose `history`; the
   frozen-replay verifier rebuilds the base regimen alongside the realized ledger, so the
-  pre-scheduled + reactive bookkeeping is checked bit-for-bit each run.
+  pre-scheduled + reactive bookkeeping is checked bit-for-bit each run. The `auc_target`
+  exposure metric likewise integrates the base regimen: each window's AUC covers the base
+  doses — a loading dose before the first window and a maintenance dose landing inside a
+  window alike — not just the controller's, so `auc_target_attainment` is not biased low on a
+  base-regimen run (constant-covariate; the time-varying / IOV / reset `auc_target` caveats
+  below are unchanged).
   A base dose that **shares a time with a decision is observed pre-dose** (the trough),
   symmetric with the controller's own doses — the controller reads the level *before* the
   scheduled dose lands, then that dose is applied (#933); for a steady-state base dose the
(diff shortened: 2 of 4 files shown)


TeunP:
## Adversarial self-review (xhigh) — two independent passes

Reviewed at `b64f8219`. Two passes: an independent reviewer pass + a manual line read, cross-checked. CI-green was **not** treated as correctness.

**Verdict: correct.** One doc inconsistency found and fixed during review (below); no functional issues.

### Finding (resolved)

| Sev | Location | Problem | Fix |
|-----|----------|---------|-----|
| Low (doc) | `src/ode/predictions.rs:4185` | The inline lead-in comment still read "carrying **only the doses at or before the window's left edge `a`**" — the old `<= a` semantics — contradicting both the new `time < b` code and the detailed comment three lines below it ("any base maintenance dose strictly inside `(a, b)` are kept"). | Rewrote to "the base regimen and ledger doses … before the right edge `b`". Comment-only; folded into `b64f8219`. |

### Correctness — verified

- **The fix.** `sub.doses = ledger…` (which *replaced* the subject's doses) → `sub.doses.extend(ledger…); sub.doses.retain(|d| d.time < b - 1e-9)`. Base doses now survive the composition; mirrors `verify_adaptive_frozen_replay`.
- **Filter.** `d.time < b - 1e-9` excludes the dose *at* `b` (the next window's left-edge dose, which would fold a post-dose jump into this window's right endpoint) while keeping doses ≤ `a` (state setup) and any base maintenance dose strictly inside `(a, b)`. For controller doses — which land on decision-grid edges — `< b` is byte-identical to the old `<= a`, so the pure-controller path is unchanged (the two pre-existing AUC tests still pass).
- **Base-dose attributes.** Base doses are kept via `clone`, preserving `SS`/`II`/lagtime/`rate_mode`/infusion; the `DoseEvent::new(…, false, 0.0)` rebuild (which resets those) is applied only to ledger/controller doses, which are plain boluses.
- **Unsorted doses.** `sub.doses` becomes `base ++ ledger`, which need not be time-ordered. `ode_dense_solve_states` sorts its break timeline (`build_segment_break_times`) and applies doses by *time match* in `apply_segment_boundary` (not by Vec index); `dose_lagtimes`/`dose_f_bio` are rebuilt parallel to the same `sub.doses`. Order-independent — and empirically confirmed: the mid-window test's `sub.doses = [base@12, ledger@0]` is out of order and still yields the correct AUC.
- **Single PK snapshot** stays valid: this pass is reachable only for constant-covariate subjects (TV-cov / TIME-in-PK / IOV / reset are rejected upstream at `api/adaptive.rs:411`), so base doses read the same `pk_params_flat`.

### Tests

- Two Tier-1 degenerate-oracle tests. Closed forms verified numerically. Tolerances: `1e-4` for the smooth-decay loading-dose window; `2e-2` for the mid-window bolus (documented ≈ 0.94 % trapezoid node bias, since an instantaneous bolus into the monitored compartment jumps the signal off-node).
- Non-vacuous: both assert a material AUC increase from the base dose (`> 1.1×` and `> 2×` vs ledger-only).
- Mutation-verified: reverting the fix (base dropped) fails both tests; flipping `< b` → `<= a` fails the mid-window test only (the loading test still passes, since its base dose is at/before `a`). This is exactly the discrimination each test is meant to provide.
- Changed lines are covered.

### Conventions / performance / security

- **Conventions:** Tier-1 unit tests in the sibling `*_tests.rs`; the adaptive-dosing "no NONMEM" exception is honored (degenerate oracle); CHANGELOG `[Unreleased]` entry + docs updated (`adaptive-dosing.qmd`).
- **Performance:** one extra `extend`+`retain` per window — negligible against the already-`O(m²)` per-window dense solve, and the pass runs post-run only (never in the fit/inner loop).
- **Security:** none — internal metric computation, no I/O or untrusted input.

### Non-blocking follow-ups

- SS / lagged / infusion base doses are covered *by construction* (clone → the same `ode_dense_solve_states` machinery as `predict()`) but are not directly unit-tested in this pass; an SS-base-dose oracle would be a cheap add.
- The mid-window instantaneous-bolus node bias (≈ 0.9 %) is inherent to the uniform per-window trapezoid grid; a node-at-dose refinement is the same follow-up as the `O(m²)`→`O(m)` resume-solve already noted in the function docs.

_Two-pass review (independent reviewer + manual). The one issue found was fixed in-branch._

PR #677, opened 2026-07-02
feat(estimation): analytic gradient for magnitude×iiv_on_ruv×IOV and LTBS×IOV outer [#486]

hiddevandebeek:
## Why

Two finite-difference *gradient* cells tracked in #486 were flagged **EASY** by the post-#673 live-gate audit (one gate-flip + a parity test each). This closes both, extending analytic coverage to models that previously fell back to per-subject FD on the population gradient:

1. **Custom / time-varying residual-error magnitude × `iiv_on_ruv` under IOV.** #673 shipped this for the non-IOV path but *gated it to non-IOV* pending validation of the κ-augmented residual-eta assembly. The assembly turns out to be dimension-generic, so IOV composes for free.
2. **LTBS (log-transform-both-sides) × IOV — outer θ/Ω/σ gradient.** LTBS was analytic on the non-IOV outer/inner (#665/#673) but declined entirely under IOV.

A third audit candidate — analytic Form-C readout + `[initial_conditions]` baseline — was investigated and **deferred**: it is not a gate-flip (init must be folded into the concentration *before* the readout, and the init impulse runs at a different dual width than `run_obs`), so it is genuinely HARD, not EASY.

## What changed

**1. Custom magnitude × `iiv_on_ruv` under IOV (`sens_outer_gradient.rs`)**
- Dropped the `ruv.is_some() && model.n_kappa > 0` sub-clause from `prepare_stacked`'s magnitude bail. The residual-eta assembly (`rr` in the BSV block; `a[l]`, `m_vec[rr]`, `prep.w[j][rr]` loop over all stacked axes incl. κ) is dimension-generic, so the magnitude direct-θ terms compose with occasion random effects with no new code. The magnitude × M3-censored bail is kept.

**2. LTBS × IOV — outer only (`provider.rs`, `inner_optimizer.rs`)**
- Relaxed the `log_transform` bail in `iov_analytical_supported` (`has_lagtime` still declines).
- Wired the post-walk `ln(f)` jet into `subject_sensitivities_iov` (`apply_ltbs_transform_outer`, applied LAST — after the in-walk scale quotient / Form-C readout / init impulse — reproducing production's scale-then-log order `ln(f/s)`; `predict_iov` logs last, verified).
[...]

PR #673, opened 2026-07-02
feat(sens): LTBS analytic inner subsets + residual-magnitude × iiv_on_ruv outer [#486]

hiddevandebeek:
## Why
Two of the remaining #486 Tier-1 analytic-gradient cells: closing them removes finite-difference fallbacks for common model shapes so those fits get the exact (and, at higher η-dimension, faster) analytic gradient. Continuation of #665 (plain-LTBS inner) and #644/#659 (residual-magnitude outer).

> **Scope note:** this is the *tractable half* of the Tier-1 batch. On implementation, **magnitude × M3** and the **Form-C readout variants** (per-CMT / init / cross-endpoint R) turned out to need materially more machinery than initially scoped (threading the censored-row magnitude through the EBE response + a FOCEI `dd2_dtheta` `log|H̃|` branch; per-CMT dispatch through four dual paths). They are deferred to focused follow-up PRs rather than shipped half-done in the gradient/covariance domain.

## What changed
- **LTBS analytic *inner* EBE gradient — remaining closed-form subsets.** Extends the shared `g = ln(f)` inner jet (new `apply_ltbs_transform_inner`, the inner twin of `apply_ltbs_transform_outer`) to LTBS combined with an η-dependent `ExpressionScale obs_scale`, with **time-varying covariates** (the event-driven inner walk), and with a `TIME`-built-in structural parameter. Applied **last**, after any scale quotient (reproducing `ln(f/s)`), mirroring the outer. LTBS × IOV still routes to FD via `iov_analytical_supported`'s own `!log_transform` gate. Removed the now-obsolete model-level LTBS+ExpressionScale inner bail (`analytic_inner_common_bail`) and the TV-cov / TIME inner declines.
- **Custom / time-varying residual-error magnitude × `iiv_on_ruv`** — exact analytic FOCEI outer (θ/Ω/σ) gradient. `theta_block` now adds the residual-eta `c̃`-column direct-θ terms (`m_vec[rr] += ε²/R²·Rθ` and `∂(d/R)/∂θ·w[rr]`), mirroring the existing `sigma_block` terms. Model-level gate relaxed (dropped `residual_error_eta.is_none()` for magnitude models). Made the test harness `precise_ebe_ruv` magnitude-aware (it was using the plain, unscaled variance).

## Alternatives considered
Splitting into one focused PR per subsystem (my initial recommendation). Bundled the two validated cells here per request; deferred the two that proved Hard.

## Cross-repo dependency
| Repo | PR | Status | Must merge |
[...]

PR #631, opened 2026-06-30
feat(parser): analytic FOCEI sensitivities for direct θ/η in a Form-C readout (#486)

hiddevandebeek:
## Why

On the ODE path, a Form-C readout `y = <expr>` is served with exact analytic FOCE/FOCEI sensitivities when `expr` references states, individual parameters, constants, and covariates (#410/#439/#540). The remaining gap (#486 status matrix, "Non-simple readout — direct θ/η"): a readout that references a parameter **directly** — a bare `THETA(i)`/`ETA(k)`, e.g. an additive baseline `y = central/V1 + TVBASE` or a multiplicative readout scale `y = central/V1 * (1 + ETA_SCALE)` — dropped the subject to finite differences, because the readout assembled no `∂y/∂θ` / `∂y/∂η` term.

This closes that gap: a direct θ/η in a Form-C readout is now analytic on every walk (static, mixed-order, TV-cov, inner EBE, IOV).

## What changed

A direct `THETA(i)`/`ETA(k)` in a readout **is** semantically an individual parameter equal to that θ/η (NONMEM's `MU_i = THETA(i)`). So instead of bolting a bespoke θ/η term onto each of the five analytic walks, the parser **desugars** it:

- At parse time, each bare `THETA(i)`/`ETA(k)` in a `[scaling] y = …` readout is rewritten into a synthetic individual parameter (`__ferx_ro_th{i} = THETA(i)` / `…eta{k} = ETA(k)`); the readout AST is rewritten to reference that parameter (`PushTheta`/`PushEta` → `PushVar`).
- The readout then carries no bare θ/η ops, so it is `dual_evaluable`, and its `∂y/∂θ`/`∂y/∂η` (and the 2nd-order `∂²y/∂η∂θ`, `∂²y/∂η²` blocks) ride the **already-validated** individual-parameter sensitivity chain — identically across all five walks, with zero new sensitivity-layer code.
- The synthetic parameters are hidden (`__ferx_ro_*` prefix) from EBE/sdtab/fit output and the unused-parameter census; the f64 prediction value is unchanged.

An `ETA(k)` reference reuses the existing random effect (no new omega). A readout referencing a neural-network output stays on the FD fallback.

## Alternatives considered

[...]

PR #637, opened 2026-07-01
feat(sens): analytic TIME event-time built-in sensitivities (#486, #610)

hiddevandebeek:
## Why

PR #610 added the `TIME`/`time` event-time built-in inside `[individual_parameters]`
(NONMEM-style `$PK IF (TIME.GE.45) CL = …` time-dependent switches), but a model that
**uses** `TIME` in a structural parameter routed its FOCE/FOCEI sensitivity gradient to
finite differences. This threads the per-event time into the analytic `Dual1`/`Dual2`
kernels so those models get exact analytic gradients (faster and non-noisy), on every
estimation path. Tracks the #486 sensitivity matrix.

`TIME` is data (an event time): `∂param/∂TIME` is not a parameter sensitivity. Like a
time-varying covariate, it makes the parameter piecewise/time-varying and is threaded as
a per-event constant, so the fix mirrors the existing TV-covariate constant-threading.

## What changed

- **Per-event time seeding.** A `ModelTimeGuard` is set at each per-event seed seam so
  `Op::PushTime` resolves to that event's time in the dual walk, and the event time is
  passed to `pk_param_fn` for the f64 value:
  - closed-form non-IOV `run_obs_tvcov` / `run_obs_grad_tvcov`;
  - closed-form IOV `build_iov_sources` (value **and** `CombinedDerivs` together);
  - ODE `seed_pk_dual1/2` (+ `seed_tvcov_snapshots` folds the event time into its dedup
    key only when `uses_time_builtin`, so non-`TIME` TV-cov dedup is unchanged);
  - ODE IOV `run_subject_iov` / `run_subject_iov_eta` via the shared `seed_iov_events`
    (a new `per_event` flag forces per-event seeding under `uses_time`).
- **Routing.** `TIME` models route through the per-event walk (even with no TV
  covariates) via `uses_time_builtin` in `analytical_supported`,
  `subject_sensitivities_impl` / `subject_eta_grad_impl`, the TV-walk triggers,
  `analytic_inner_grad_supported`, `iov_analytical_supported`, `ode_analytical_supported`,
  `ode_subject_supported` (declines `TIME` → static walk can't hold params constant),
  `ode_tvcov_supported`, and `ode_iov_supported`.
- **ExpressionScale on the event-driven walk.** The shared, provider-agnostic
  `apply_expression_scale_outer` / `_inner_dispatch` (previously called only on the
  dose-superposition / static paths) is now applied post-walk on the event-driven walks
[...]

PR #950, opened 2026-07-28
feat(sens): analytic LTBS inner under IOV + direct-theta/eta Form C readout on the closed-form engine [#486]

hiddevandebeek:
## Why

Two remaining finite-difference cells from the #486 matrix, both on the closed-form engine and both hit by ordinary everyday models rather than corner cases:

1. **LTBS × IOV inner.** `log(DV) ~ additive(...)` combined with `[iov]` is a mainstream NONMEM idiom. Its population gradient has been analytic since #677, but the inner EBE gradient stayed on finite differences because the first-order IOV walk carried no `ln` jet. This was the **last deliberate inner/outer scope split** in the matrix — the two loops differentiated the same subject by different means, which is exactly the asymmetry the provider's scope rule exists to prevent.

2. **Closed-form Form C readout referencing a θ/η directly.** A readout like `y = central/V * TVSCALE + ETA_BASE` — a baseline or scale factor that is not an `[individual_parameters]` entry — has been analytic on the **ODE** engine since #631, which desugars each bare reference into a hidden individual parameter. The closed-form engine simply never ran that pass (`collect_readout_theta_eta_synth` was gated on `is_ode`), so the bare `PushTheta`/`PushEta` ops survived into the readout bytecode, `dual_evaluable` stayed false, and every such subject dropped to FD on both loops.

Neither changes any estimate: analytic and FD gradients locate the same optimum. What changes is that these fits stop paying `~2·n_eta` predictions per inner step, and — for (1) — that the EBE and the population gradient now provably differentiate the same function.

## What changed

**LTBS × IOV inner.** `run_obs_iov_eta` applies the shared `apply_ltbs_transform_inner` last, after the per-occasion scale quotient — mirroring exactly where `subject_sensitivities_iov` applies `apply_ltbs_transform_outer`. The transform is a per-observation scalar quotient (`∂g/∂x = (∂f/∂x)/f`), identical on every stacked axis, so the occasion κ columns carry it exactly as the BSV η columns do: no per-occasion jet, no dimension-specific handling. `analytic_inner_common_bail` therefore drops its `log_transform && n_kappa > 0` clause.

**Direct-θ/η readout on the closed-form engine.** The desugaring itself is engine-agnostic; what differed is *where the synthetic parameter's PK slot comes from*. An ODE model routes every individual parameter through `ode_param_slots`, known at the collection point. An analytical model draws from the spare region `allocate_readout_extra_slots` owns — and neither `pk_model` nor `pk_param_map` exists that early in the parse. So the analytical append is **deferred to that allocator**, which already has the information and already merges its slots into `pk_var_slots`. Sizing policy is preserved on both sides: text-referenced readout parameters keep priority and their pre-existing hard error on overflow (the user named them); synthetic θ/η stay best-effort, so a model that parsed before this change — where a direct θ/η took no slot at all — cannot start failing, and instead keeps the FD fallback with the existing parse warning.

### A scope hole that had to be ruled out, not assumed

`analytic_inner_common_bail` is shared by the closed-form **and** ODE IOV inner routes, so dropping a clause from it could have opened an analytic inner for a model whose outer still declines. It does not: `ode_iov_supported` declines `log_transform` independently (`ode_provider.rs:962` — that walk applies LTBS in PK-parameter space *before* the η/θ/κ chain, so production's scale-then-log order cannot be rebuilt post-walk), so ODE IOV × LTBS still routes to FD on **both** loops via the support gate. The test that previously pinned this through the bail now pins it through the support gate explicitly, rather than leaving it implicit.

## Alternatives considered

For the readout desugaring, the obvious move was to flip the `is_ode` gate and allocate slots in place. That does not work — the slot pool is not known at that point in the parse — and forcing it would have meant either duplicating the analytical slot-allocation rule at the collection site (two sources of truth for one layout) or moving the ODE append later too (churning a path that is working). Deferring only the analytical half keeps each engine appending where its slot map is actually known, and both still land before `build_pk_param_fn`, which is the real constraint the original comment names.

## Cross-repo dependency

No public Rust API change; nothing in `FitResult`/sdtab moves.

## Breaking changes
- [x] None

## Numerical validation

**Predictions and estimates are unchanged by construction** — this moves the *gradient route*, not the objective. So the meaningful validation is gradient parity against finite differences of the very objective each loop minimises, plus a negative control proving each test is load-bearing.

- [x] Compared against: central FD of the production predictor / objective, at base commit `54a7d449`
- [x] Reference values:

```
LTBS x IOV inner gradient vs central FD of individual_nll_iov
  (2 fixtures: plain LTBS x IOV, and LTBS x IOV x ExpressionScale)
  every stacked [eta_bsv, kappa] coordinate: max_relative = 1e-4   PASS

  negative control - transform call reverted, same test:
    analytic -5699.434247988035  vs  FD -73.01212026401447    FAIL (as intended)

Direct-theta/eta closed-form readout vs central FD of the readout-aware predictor
  value + all first/second eta/theta derivatives (check_full_provider_vs_fd)     PASS
  d y/d TVSCALE and d y/d ETA_SC both non-zero (silent-zero-jet guard)           PASS
  light inner eta-gradient vs full provider eta block: max_relative = 1e-9       PASS

  negative control - `is_ode` gate restored, same test:
    "a direct-theta/eta closed-form readout is desugared and must stay analytic" FAIL (as intended)
```

No NONMEM anchor is included because there is no new numerical object to anchor: both cells replace a finite-difference approximation of an already-NONMEM-anchored objective with its exact derivative. The FD reference *is* the incumbent behaviour.

## Performance
- [x] Faster — the inner EBE loop for the affected models goes from `~2·n_eta` predictions per gradient to one provider evaluation. Not separately benchmarked; this is the same mechanism and the same order of saving as every prior #486 cell.

## Tests

Two Tier-1 FD-parity tests, each with a verified negative control (above). The LTBS one runs two fixtures because the *ordering* is the part that silently goes wrong: applying the `ln` jet before the scale quotient instead of after still yields a plausible gradient — it is just the gradient of `ln(f)/s` rather than `ln(f/s)` — and only FD of the real objective catches it. The readout test asserts **non-zero** `∂y/∂θ` and `∂y/∂η` specifically, because a synthetic parameter that never reaches `pk_var_slots` reads as a silent zero jet while every other derivative still matches FD.

Full suite: **2751 passed, 1 failed**. The failure is `ode::predictions::tests::ss_input_rate_over_capacity_deep_saturation_declines`, which **fails identically on clean `54a7d449`** (same value, `18626040.316590052`) — pre-existing on main and unrelated to this PR. Flagging it rather than folding a fix in here.

<details>

A closed-form readout naming a theta and an eta directly — analytic on both loops as of this PR, FD before it:

```toml
[parameters]
  theta TVCL(0.2, 0.001, 10.0)
  theta TVV(10.0, 0.1, 500.0)
  theta TVSCALE(1.5, 0.01, 100.0)
  omega ETA_CL ~ 0.09
  omega ETA_V  ~ 0.04
  omega ETA_SC ~ 0.05
  sigma PROP_ERR ~ 0.02 (sd)
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV  * exp(ETA_V)
[structural_model]
  pk one_cpt_iv(cl=CL, v=V)
[scaling]
  y = central / V * TVSCALE + ETA_SC
[error_model]
  DV ~ proportional(PROP_ERR)
```

`TVSCALE` and `ETA_SC` reach the prediction only through the readout. Before this PR their sensitivities came from finite differences (the whole subject dropped to FD); now they ride the individual-parameter sensitivity chain, and the reported inner/outer gradient method is analytic.

</details>

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added (user-facing change), or N/A

`docs/estimation/foce.qmd` — the LTBS × IOV sentence said the inner "stays on finite differences"; now states both loops are analytic on closed-form models and that the ODE case still falls back, with the reason. `docs/model-file/scaling.qmd` — the analytical Form C row listed "a direct `THETA`/`ETA` ... reference" among the FD fallbacks; that entry is replaced with the desugaring description and the (still real) slot-overflow fallback.

## Checklist
- [x] `cargo clippy` clean — CI's exact gate (`cargo clippy --no-default-features --features ci,markov`) emits warnings only, all pre-existing in `src/bin/` dev tooling

### ferx-site
- [ ] New example `.ferx` file added to `examples/`
- [ ] New data file added to `data/`

No new DSL syntax, fit option, example file, or data file — both cells change the gradient route for syntax that already parses.

### ferx-book

- [x] No example execution step needed (no user-visible behaviour change beyond gradient route; no `examples/` file affected)

## Reviewer hints

**Focus on two things.**

First, `run_obs_iov_eta`'s new call site — specifically that it sits *after* the `ScalarScale` / `ExpressionScale` quotients and before `Some(out)`. That ordering is the whole correctness argument, and it is the one thing a reader cannot verify by inspection alone, which is why the second test fixture exists.

Second, the deferred append in `model_parser.rs`. Worth confirming that appending to `indiv_var_names` / `indiv_stmts` at the `allocate_readout_extra_slots` call site is still upstream of every consumer — `build_pk_param_fn` (immediately below), `pk_indices` (built later), and `parse_scaling_block`'s rewrite (later still). I traced each; a second pair of eyes on that ordering is the highest-value review.

Skimmable: the comment churn. Several comments across `provider.rs`, `inner_optimizer.rs`, `provider_tests.rs` and `build_info.rs` asserted "LTBS × IOV inner stays on FD"; a few also claimed LTBS had "no inner counterpart", which was already stale before this PR (#673 added `apply_ltbs_transform_inner`). Those are corrected in passing.

## Open questions

The two commits are independent and touch disjoint code — happy to split into two PRs if that reviews better.

The slot-overflow path for synthetic readout parameters is exercised by construction but not by a dedicated test: reaching it needs a PK model whose spare slots are already consumed (a 3-cpt oral with modeled-dose parameters *and* text-referenced readout parameters). Worth adding if you want that branch pinned rather than reasoned about.


hiddevandebeek:
## Code review

This PR does two things — it enables the analytic `ln(f)` inner gradient for closed-form LTBS × IOV, and it extends the Form C direct-`THETA`/`ETA` readout desugaring to the closed-form engine. Most of the findings below are loose ends in one of those two changes. Ten findings survived an independent verification pass; the six most consequential are correctness issues.

### Correctness

**1. `uses_closed_form_ltbs_inner` was not updated alongside the removed bail** — `src/estimation/inner_optimizer.rs:1624`, `src/types.rs:3123`

Removing `|| (model.log_transform && model.n_kappa > 0)` from `analytic_inner_common_bail` turns on the analytic `ln(f)` closed-form inner EBE gradient for LTBS × IOV, but `CompiledModel::uses_closed_form_ltbs_inner()` still hard-codes `n_kappa == 0` — and its doc-comment cites the now-deleted bail as its justification ("LTBS + IOV routes the inner loop to FD, so neither needs the tightened fit/covariance tolerances"; repeated at `src/types.rs:5898`). That flag is the sole input to `FitOptions::effective_inner_tol` (tightens to `LTBS_FIT_INNER_TOL` = 1e-6) and `effective_cov_inner_tol` (tightens covariance EBE reconvergence to `LTBS_COV_INNER_TOL` = 1e-8, consumed at `src/estimation/covariance.rs:526`). `types.rs` is not touched by this PR.

Concretely: fit a closed-form model with `log(DV) ~ additive(ADD_ERR)` plus `iov_column = OCC` — exactly the fixture the new `analytic_iov_inner_grad_matches_fd_of_nll_closed_form_ltbs` test builds — and request covariance. The EBEs are now found with the analytic ln-jet gradient, which carries the documented provider-vs-`compute_predictions` gap that the `ln` wrap amplifies, while both the fit and covariance steps stay at the loose default `inner_tol` (1e-5). The result is standard errors inflated by the same mechanism measured at roughly 65% on warfarin theta SEs in the non-IOV case, with unchanged point estimates, so nothing looks wrong. Pre-PR this could not occur, because the bail forced the FD inner.

Either extend `uses_closed_form_ltbs_inner` to `ode_spec.is_none() && log_transform` (subject to the IOV block-Omega over-convergence caveat its own doc raises), or keep a deliberate IOV carve-out — but the two predicates should not be left out of sync.

**2. Accepted readout synthetics can exceed the 9-wide `const_dispatch!` cap** — `src/parser/model_parser.rs:12136`

The accept loop grows `indiv_param_program.pk_slots()` with no check against the dispatch-table width that both closed-form sens providers enforce. Consider a closed-form `two_cpt_oral` mapping `cl,v,q,v2,ka,f,lag` (7 program slots) with a readout such as `[scaling] y = central/V + BMAX*C/(KD+C) + THETA(6) + ETA(4)`: both synthetics are pushed, so `prog.pk_slots().len()` reaches 11. `analytical_supported_core` only checks that every slot maps via `slot_to_dim` — it never bounds the count — and `analytic_readout_dual_supported` is now true because the desugaring made the readout dual-evaluable, so `analytical_supported` returns true and `build_info::gradient_method` / `gradient_method_inner` report "analytic". But `const_dispatch!(slots.len(); 1,…,9)` at `src/sens/provider.rs:5031` (`subject_sensitivities`) and `:3949` (`subject_eta_grad`) has no arm for 11 and falls through to `None`, so both providers decline every subject and the fit runs entirely on finite differences at roughly 10× the cost, while the persisted label says analytic. This is the "claim analytic then FD every subject" misreport that `analytic_readout_dual_supported`'s own doc comment says the gate exists to prevent (#637).

**3. `__ferx_ro_*` names leak into the user-facing mu-referencing warning** — `src/parser/model_parser.rs:1699`

Appending the synthetics on the closed-form engine exposes them to the mu-referencing check in `api/fit.rs`, which — unlike the four sibling consumers of the same list — does not filter synthetic readout params.

Using the PR's own new fixture `ONECPT_IV_DIRECT_THETA_ETA_READOUT` (`[scaling] y = central / V * TVSCALE + ETA_SC`): the deferred append puts `__ferx_ro_eta2` into `indiv_var_names`/`indiv_stmts`, giving `indiv_param_names = [CL, V, __ferx_ro_th2, __ferx_ro_eta2]` and `eta_map = [0, 1, -1, 2]`. `detect_mu_refs` matches no pattern for a bare `Eta(k)`, so `model.mu_refs` has no entry, and `api/fit.rs:657-676` — which skips only `eta_idx < 0` — pushes it. `FitResult.warnings` then reads: *"individual parameter(s) not mu-referenced: `__ferx_ro_eta2`. This can strongly affect convergence; prefer forms such as `CL = TVCL * exp(ETA_CL)`"*. The user is asked to fix a parameter that does not exist in their model file and cannot act on the advice. The four sibling sites (`src/api/output_columns.rs:110`, `src/api/validation.rs:2945` and `:2955`, `src/parser/model_parser.rs:2908`) all guard with `is_synthetic_readout_param`; this one should too.

**4. A synthetic `Custom` `EtaParamInfo` row reaches `FitResult` and the R metadata layer** — `src/parser/model_parser.rs:1702`

For the same fixture, `classify_indiv_params` now sees the synthetic statement `__ferx_ro_eta2 = ETA(2)`. A bare `Expression::Eta(k)` matches none of `classify_expr`'s patterns, so the `else` arm pushes `EtaParamInfo { eta_name: "ETA_SC", param_type: Custom, linked_theta: None, individual_param_name: "__ferx_ro_eta2" }`. `api/fit.rs:1975` clones `model.eta_param_info` verbatim with no synthetic filter, so ferx-r's eta/parameter table gains a row whose individual-parameter name is a parser-internal token and whose `param_type` is `Custom` rather than the model's actual parameterisation.

There is a second-order effect worth checking: `resolve_param_derivs` (`src/sens/provider.rs:1125`) and `subject_eta_grad` (`:3929`) both fall back to `None` — i.e. FD for the whole model — whenever any `eta_param_info` entry is not `LogNormal` and the program path is unavailable, so this synthetic `Custom` entry can also poison that lognormal fallback for closed-form models that would otherwise take the `pk·sel` chain.

**5. The shared FD-fallback warning quotes the ODE slot count on the analytical engine** — `src/parser/model_parser.rs:12151`

`readout_synth_fd_note` is now shared between the ODE and analytical sizing sites, but its text hard-codes `MAX_PK_PARAMS` (128) as "the {}-slot layout". That is only correct at the ODE site; the analytical allocator draws from `0..=PK_IDX_MTT` minus `consumes_pk_slot`/modeled slots, i.e. at most about 11 and typically 4–7.

A user fitting an analytical `pk three_cpt_oral(...)` whose readout references more distinct bare `THETA`/`ETA` than the spare region holds (3-cpt-oral consumes CL, V, Q, V2, KA, Q3, V3 plus reserved F/lagtime, leaving only slots 9 and 10) drops silently to finite differences and is told: *"needs 2 more PK slot(s) than the 128-slot layout has free … free up 2 individual-parameter slot(s)"*. The figure is wrong by an order of magnitude for the engine that produced the warning, and the remediation cannot be acted on.

**6. The doc comment and `#[allow]` now bind to `ReadoutExtraSlots`, not to the allocator** — `src/parser/model_parser.rs:12049`

The new `ReadoutExtraSlots` struct was inserted between `allocate_readout_extra_slots`'s doc comment plus `#[allow(clippy::too_many_arguments)]` and the function itself. Two consequences: the allocator — which the same hunk grew from 8 to 9 parameters — is now un-suppressed and will trip `clippy::too_many_arguments` on the CI Clippy job (`cargo clippy --no-default-features --features ci,markov`), a warning that does not exist on `main`; and `cargo doc` renders the allocator's ~19-line doc block as the documentation for `ReadoutExtraSlots`, leaving `allocate_readout_extra_slots` undocumented.

### Observations

**7. Stale comments at the live IOV inner-dispatch site** — `src/estimation/inner_optimizer.rs:1170`, `:1747`

Line 1160 still lists LTBS among the common bails, and line 1170 asserts "LTBS still routes IOV to FD via `analytic_inner_common_bail` (`log_transform`)" — immediately above the `let analytic_iov_inner = … && !analytic_inner_common_bail(model)` that now evaluates to `true` for LTBS × IOV. Line 1747 separately describes the removed `log_transform && n_kappa > 0` clause as present but "vacuous". The dispatch site is the most authoritative place a maintainer looks for "what routes to FD"; leaving it asserting the deleted invariant invites either a wrong scope matrix in a follow-up or a re-added conflicting bail. The PR already corrected six comments of this class (`provider.rs:978`/`1329`/`2562`, `build_info.rs`, and the two test files), so these read as omissions rather than deliberate.

**8. The ODE bail list is now an exact duplicate of `analytic_inner_common_bail`** — `src/estimation/inner_optimizer.rs:1846`

The only thing that justified the hand-inlined copy in `analytic_inner_grad_supported`'s ODE branch was the `log_transform && n_kappa > 0` clause that made the shared bail a strict superset. With that clause deleted the two lists coincide exactly — as the diff's own new comment at `src/sens/provider.rs:976` states. Leaving both in the tree means the next escape hatch added to `analytic_inner_common_bail` (the documented single source of truth) will silently not apply to ODE inner subjects, which will keep claiming an analytic inner gradient the provider then declines — the inner/outer route split this PR set out to close. Lines 1846-1859 could become `if analytic_inner_common_bail(model) { return false; }`.

**9. The desugaring pass is split into two halves roughly 370 lines apart** — `src/parser/model_parser.rs:1698`

The same decision — do the synthetic readout params fit, and if not drop them all and warn — is now written twice: at lines 1330-1342 against `ode_free_slot_count`, and inside `allocate_readout_extra_slots` at 12126-12140 against `free.split_off`. `is_ode` is tested at three sites (1330, 1698, 12071) that must stay mutually consistent. A future change to the policy (for example accepting a partial set rather than dropping all) applied to only one branch would make the two engines disagree on whether the same readout is analytic, producing different gradients for one model expressed two ways, with no test pinning the pair. A single `assign_readout_synth_slots(engine_pool) -> (accepted, note)` called by both engines, with one append site, would remove the class.

**10. The new analytical FD-fallback branch is untested** — `src/parser/model_parser.rs:12133`

The `synth_params.len() > remaining.len()` branch, the `fd_note` field, and the `readout_synth_params = readout_alloc.accepted_synths` empty-vec path are all changed lines with no covering test (nothing asserts the note text or exercises a full-slot readout model), which puts the Codecov `patch` status at risk under the ≥90% diff-coverage rule. Functionally it also means it has not been verified that dropping the synthetics on the analytical engine actually leaves `dual_evaluable = false` — if it did not, such a model would take the analytic path with the bare `PushTheta`/`PushEta` ops and produce a zero jet for those axes rather than falling back to FD.

---

*Reviewed at high effort with a multi-agent finder + independent verification pass; 6 of 25 candidate findings were refuted and dropped.*

hiddevandebeek:
## Response to review

Thank you — this was a genuinely useful review. I verified every premise against the code before acting; all ten hold. Eight are addressed in `0617aba4`, one is declined with reasoning, and one is folded into another fix.

Three findings turned out to be **pre-existing on `main` rather than introduced here**. I have fixed them anyway, because this PR roughly doubles their exposure, but the attribution matters for the changelog and for anyone bisecting:

* **#3 and #4** — the ODE engine has appended these same `__ferx_ro_*` synthetics since #631, and `classify_indiv_params` (`model_parser.rs:2095`) runs *after* both append sites. So both leaks already reach `FitResult` and the mu-referencing warning for ODE models today; extending the desugaring to the closed-form engine doubles the affected model set. The filters are therefore applied at the shared sites, and the new test runs **both** engines.
* **#2** — the >9 overflow is reachable on `main` through text-referenced readout parameters alone (`allocate_readout_extra_slots` admits up to `free.len()` of them, and `analytical_supported_core` never counted). #486's synthetics add a second route to the same latent bug.

---

### Correctness

**1 — `uses_closed_form_ltbs_inner` out of sync.** Confirmed and fixed. The `n_kappa == 0` carve-out was justified *solely* by the bail this PR deleted, so it should have moved in the same commit. Dropped it: the predicate is now `log_transform && ode_spec.is_none()`, which is exactly "runs the analytic `ln(f)` closed-form inner gradient" — the property both tolerances actually key off.

I considered the conservative alternative you offered (a deliberate IOV carve-out, citing the doc's own IOV block-Ω over-convergence caveat) and rejected it. That caveat argues against *blanket*-tightening every model; it is not an argument for excluding a model that demonstrably runs the gradient the tightening exists for. The #654 case it names is M3 + IOV, which is not LTBS and is unaffected either way. Leaving the carve-out in would have meant shipping the inflated-SE path you describe.

The two predicates are now pinned together — `analytic_iov_inner_grad_matches_fd_of_nll_closed_form_ltbs` asserts `uses_closed_form_ltbs_inner()` immediately after `!analytic_inner_common_bail()`, so they cannot drift apart again silently. Negative control: restoring `&& self.n_kappa == 0` fails that assertion.

This changes standard errors for closed-form LTBS + IOV fits, so it has a `### Fixed` changelog entry and a docs update rather than being folded in silently.

**2 — dispatch-width overflow.** Confirmed. Added `closed_form_slot_count_supported` to `analytical_supported_core`, with `MAX_CLOSED_FORM_SLOTS = 9` documented as needing to move in step with the two `const_dispatch!` tables.

Two notes on scope. I bounded the **program** slot list rather than also the `lognormal_param_derivatives` fallback's `model.pk_indices`: `pk_indices` is parallel to every `[individual_parameters]` assignment (intermediates included), so bounding it risked newly declining models that work today, and that path is untouched by this PR. And I chose the honesty gate over widening the tables to 13. Widening would convert these fits from honest-FD to genuinely-analytic, which is better for users — but it is four more monomorphisations of `run_obs`/`run_obs_grad`, a behaviour improvement beyond this PR's scope, and it should be its own change with its own build-time measurement. Worth filing.

Your reproduction needed adjusting, incidentally: `two_cpt_oral` leaves only 4 spare slots (`consumes_pk_slot` reserves `F` and `LAGTIME` unconditionally on top of `required_pk_params`), so that fixture reaches 8, not 11. The test uses `three_cpt_oral` — 8 structural slots, and the two #486 synthetics take the only two free ones (`N`, `MTT`) — reaching 10. That routes the regression through this PR's own mechanism rather than the pre-existing one.

**3 — `__ferx_ro_*` in the mu-referencing warning.** Confirmed and filtered, matching the four sibling consumers you identified. Scope is narrower than the finding implies — `saem_non_mu_referenced_individual_params_warning` fires only when the method chain contains SAEM (`fit.rs:722`) — but the warning text is exactly as unactionable as described when it does.

**4 — synthetic `Custom` `EtaParamInfo`.** Confirmed, including the second-order effect you flagged as worth checking: a single non-`LogNormal` entry does flip both `resolve_param_derivs` (`provider.rs:1125`) and `subject_eta_grad` (`:3929`) to `None`. Fixed at the source instead of at the consumers — `classify_indiv_params` now skips synthetic assignments — which covers `FitResult`, the fallback poisoning, and any future consumer in one place. Verified no positional indexing of `eta_param_info` anywhere; every access is by name or a total predicate. `__ferx_ro_th{i} = THETA(i)` was already inert (no eta ⇒ no classification path matches), so only the eta half changed behaviour.

**5 — wrong slot count in the shared warning.** Confirmed. `readout_synth_fd_note` now takes the calling engine's actual pool. The test asserts the exact text a 2-free-slot analytical model produces; the negative control (passing `MAX_PK_PARAMS`) reproduces your "128-slot layout" wording verbatim.

**6 — doc comment and `#[allow]` re-parented.** Confirmed, and slightly worse than described: the region on `main` already contained an orphaned doc block for `collect_readout_theta_eta_synth` sitting above the allocator's, so inserting the struct left it carrying ~27 lines documenting two *other* functions. Moved the struct above the whole block, restored the `#[allow]` to the function, and returned the orphaned paragraph to `collect_readout_theta_eta_synth`. Confirmed against CI's exact gate (`cargo clippy --no-default-features --features ci,markov`): no `too_many_arguments` warning at the allocator; the two remaining in that file are `build_pk_param_fn` and `check_unused_parameters`, both pre-existing.

### Observations

**7 — stale comments at the dispatch site.** Fixed, and you are right that these were omissions rather than deliberate. Line 1160's bail list was also wrong in the other direction (it still listed "IIV on residual error", removed some time ago), so it has been rewritten against the current bail rather than patched.

**8 — duplicated ODE bail list.** Fixed as suggested: the ODE branch is now `if analytic_inner_common_bail(model) { return false; }`. The reasoning in the finding is exactly right — the inlined copy's only justification was the clause this PR deleted, and leaving it would have set up the next escape hatch to silently miss ODE inner subjects.

**9 — the split desugaring pass. Declined, with reasoning.** The duplication is real, but the two halves are not the same decision applied twice: they size against genuinely different pools (`ode_free_slot_count` over the 128-slot layout vs. the differentiable spare region, which is ~2–11 and depends on `pk_model`), and the analytical half structurally *cannot* run at the ODE half's site, because `pk_model`/`pk_param_map` do not exist yet. A unified `assign_readout_synth_slots(engine_pool)` would still need both call sites and both appends; it would centralise the accept/drop *policy* while leaving the pool computation and the append timing where they are. That is worth doing, but it is a refactor of a path this PR is already changing substantially, and I would rather it not ride along unreviewed.

What I did instead is close the specific risk you named — "no test pinning the pair". `synthetic_readout_params_stay_out_of_user_facing_output` runs the same readout through both engines and asserts both desugar it, so a policy change applied to only one branch now fails a test. The drop path cannot be pinned symmetrically (the pools differ by construction), which is the honest limit of that mitigation.

**10 — untested analytical fallback branch.** Fixed, and thank you for naming the specific unverified assumption. `readout_synth_overflow_drops_to_fd_with_engine_correct_note` asserts all three things: the accepted set is empty, the note quotes the right pool, and — the one that actually needed checking — that dropping the synthetics really does leave `dual_evaluable = false`. It does; `analytical_supported` reports FD. Had it not, those axes would have returned a silent zero jet on the analytic path rather than falling back, which is the worse failure.

---

### Verification

Four new tests, each with a verified negative control (all four reverts were applied together and each test failed on its own assertion, with the expected message):

```
readout_slot_overflow_reports_fd_not_analytic
  gate:  !analytical_supported                                              PASS
  route: subject_sensitivities / subject_eta_grad both None                 PASS
  neg:   gate removed -> "a model wider than the closed-form dispatch
         tables must report FD, not analytic"                               FAIL (as intended)

readout_synth_overflow_drops_to_fd_with_engine_correct_note
  no __ferx_ro_* appended; !analytical_supported; note text                 PASS
  neg:   MAX_PK_PARAMS restored -> "needs 1 more PK slot(s) than the
         128-slot layout has free"                                          FAIL (as intended)

synthetic_readout_params_stay_out_of_user_facing_output  (both engines)
  no synthetic in eta_param_info; none in the mu-ref warning                PASS
  neg:   filters removed -> "[closed-form] a parser-internal parameter
         must not reach FitResult.eta_param_info"                           FAIL (as intended)

analytic_iov_inner_grad_matches_fd_of_nll_closed_form_ltbs  (+ new assertion)
  uses_closed_form_ltbs_inner() on both fixtures                            PASS
  neg:   n_kappa == 0 restored -> "plain LTBS: a model whose inner runs
         the analytic ln-jet must also take the tightened LTBS
         fit/covariance tolerances"                                         FAIL (as intended)
```

Full suite: **2754 passed, 1 failed** — still only `ode::predictions::tests::ss_input_rate_over_capacity_deep_saturation_declines`, which fails identically on clean `54a7d449` (same value, `18626040.316590052`). Nothing in the suite asserted the old `n_kappa == 0` behaviour or the old inlined ODE bail list, which is itself worth noting: both were unpinned.

Clippy on CI's exact gate: warnings only, all pre-existing.

Changelog gained one `### Performance` entry (the honest `gradient_method` reporting) and three `### Fixed` entries (the LTBS + IOV standard errors, the synthetic-parameter leaks, and the FD-note slot budget). `docs/estimation/foce.qmd`, `docs/model-file/fit-options.qmd`, and `docs/model-file/scaling.qmd` are updated for the tolerance and reporting changes.

### Follow-ups worth filing

* Widening the closed-form dispatch tables from 9 to 13, converting the finding-2 models from honest-FD to analytic.
* The `assign_readout_synth_slots` unification from finding 9.
* The `lognormal_param_derivatives` fallback's own unbounded `pk_indices` width, which finding 2's fix deliberately does not touch.

Issue #570, opened 2026-06-27
perf(estimation): joint PK-TTE fit re-solves the augmented ODE twice per NLL eval (share the Gaussian+CHZ solve)


TeunP:
## Observation

A full joint PK-TTE FOCEI fit (ODE-accumulated hazard, landed in #564 / PR #567) is slow: a
toy model of **8 subjects, n_eta=1, maxiter=3** took **~320 s**. It is *not* a per-call
pathology — a single NLL evaluation is ~13 ms; the cost is **call-count × a redundant double
ODE solve**, which a real (run-to-convergence, larger-N) fit multiplies into something
impractical.

## Diagnosis

For a joint subject, the augmented PK + cumulative-hazard ODE system is integrated **twice per
NLL evaluation**:

1. The Gaussian path solves the ODE to read the observable compartment (`model_predictions`
   → `ode_predictions*`), and
2. `tte_ode_nll` (`src/stats/likelihood.rs`) solves the **same** augmented system again via
   `ode_dense_solve_states` to read `H(t) = u[chz_state]` (and `h(t)` from the RHS) at the
   event/censor times.

This double-solve happens on **every** inner EBE evaluation (BFGS) and is repeated again for
**every FD-Hessian perturbation** of η in the FOCEI outer step (`data_term_hessian_fd` calls the
TTE closure ~`2·n(n+1)` times). Multiplied across subjects × outer iterations × the outer
optimizer's own function evals, the redundant integrations dominate.

Secondary cost: the Gaussian analytic-sensitivity (`Dual2`) path now integrates the augmented
system **including** `__chz` (and the `exp` hazard), even though CHZ never feeds back into the PK
compartments — so its sensitivity rows are computed but unused by the Gaussian `h_matrix`.

## Fix direction

- **Share one integration.** Integrate the augmented system once per (subject, η), requesting
  state output at the **union** of the Gaussian obs times and the TTE event/censor/entry times,
  and read both the observable (ipreds) and `CHZ`/`h` from that single solve — instead of
  `model_predictions` and `tte_ode_nll` each solving independently. `ode_predictions_with_states`
  already returns full state vectors at obs times; the seam is merging the TTE times into its
  `saveat` and threading the states back to both consumers.
- **Drop CHZ from the Gaussian sensitivity solve** (or skip its sensitivity rows): CHZ has no
  feedback into PK, so `∂central/∂η` is unchanged whether or not `__chz` is in the `Dual2` system.
[...]

PR #952, opened 2026-07-29
feat(ode): stiff + high-order ODE steppers behind a shared Stepper abstraction [closes #387]

### LINKED ISSUE #387: perf(ode): instrument solver regime; add ODE integrator choices by use case
## Motivation

ferx currently has one ODE integrator: explicit Dormand-Prince RK45 (`ode/solver.rs`). That is a real limitation, but there are two different performance regimes to keep separate:

1. **Tight-tolerance accuracy-limited fits** such as the Savic `transit()` NONMEM anchor (#385, `ode_reltol = ode_abstol = 1e-9`, NONMEM `TOL=9`). Here RK45 is not mainly stability-throttled; it is taking many accepted small steps to satisfy local error. A lower-order stabilized method such as ROCK2 is unlikely to fix this anchor. The method class that matches NONMEM's runtime is a high-order adaptive-order multistep solver (Adams/BDF/LSODA-style).
2. **Loose-tolerance stiff/stability-limited models** such as TMDD, diffusion-like systems, and some input-rate models at default tolerances. ROCK2 or Rosenbrock-style methods may help there, but this is a different motivation from #385.

The issue remains about broadening the ODE engine beyond RK45, but solver choice must be driven by measured regime, not by assuming all slow ODE fits are stability-limited.

## Required first step: instrument before choosing a solver

Add/run RK45 instrumentation on representative models, especially the #385 transit fit at `ode_*tol = 1e-9`, and report:

- accepted steps
- rejected steps
- accepted min-step-clamped steps
- wall-clock and RHS evaluation counts where available

Interpretation gate:

- **accepted-step-dominated** at tight tolerance → prioritize high-order adaptive multistep (Adams/BDF/LSODA-style) for #385.
- **rejection/min-step/stability dominated** at loose tolerance → ROCK2/Rosenbrock are viable targets for those models.

## Solver plan by use case

1. **Instrumentation and benchmark harness**
   - Collect accepted/rejected/min-step-clamped counts in `ode/solver.rs`.
   - Run on the Savic transit anchor at `1e-9` and at default tolerances.
   - Run at least one loose-tolerance stiff candidate (TMDD/diffusion/input-rate model).

2. **Prototype the #385-relevant path before committing ROCK2 as the headline fix**
   - Prototype a high-order adaptive multistep route: Adams/BDF/LSODA-style, or a narrow BDF/Rodas prototype sufficient to test the transit anchor.
   - Compare wall-clock and step regime at `ode_reltol = ode_abstol = 1e-9` against current RK45 and NONMEM.

3. **ROCK2 for loose-tolerance stability-limited cases**
   - Add behind `[fit_options] ode_solver = rk45 | rock2` only after instrumentation shows the intended regime.
   - Position it as a no-Jacobian stabilized explicit option for real-negative-spectrum systems, not as the expected #385 fix.

4. **Rosenbrock/Rodas fallback**
   - General-purpose stiff fallback for cases ROCK2 cannot cover, but do not describe it as minimal until the Dual2/linear-solve cost is resolved.
   - FD Jacobian is cheap for f64 prediction/simulation, but estimation has additional AD constraints below.

## AD and estimation constraints

The ODE path is **not** FD-only in estimation. FOCEI/inner gradients for supported ODE models go through `solve_ode_g` over `Dual2` (`sens/ode_provider.rs`; `ode/solver.rs`), carrying `∂u/∂p` and `∂²u/∂p²` analytically through the stepper.

Any new solver usable in estimation needs both:

- an f64 path for `simulate()`/`predict()` and fallback likelihood work
- a `Dual2`-generic sibling for the analytic sensitivity path, matching the current `solve_ode` / `solve_ode_g` split

ROCK2 therefore requires more than just an f64 implementation. Rosenbrock/Rodas is especially sensitive here: pushing `Dual2` through Jacobian construction and dense LU every step means differentiating a linear solve, with second-derivative cost and correctness concerns.

## Step-control cleanliness constraints

The current RK45 controller is intentionally a memoryless I-controller. A Gustafsson PI controller was tried and rejected because history-dependent accept/reject decisions raised the differential noise floor and stalled FOCEI.

New solvers must preserve this property for estimation:

- step-size and accept/reject decisions are value-based only (`.val()` for dual numbers)
- derivative components must flow through a fixed step sequence
- spectral-radius sniffs or stiffness detectors must also be value-only
- avoid history-dependent controller state unless it is proven clean for the gradient route

## Acceptance criteria

- Instrumentation output is available for the #385 transit anchor at `1e-9` and demonstrates whether the run is accepted-step-dominated or stability/rejection-limited.
- The issue/PR description reports the measured regime before selecting ROCK2, Rosenbrock, or multistep as the next implementation.
- Any solver exposed to estimation has an explicit `Dual2` story and tests against the scalar path.
- Documentation distinguishes ROCK2's likely loose-tolerance stiff-model value from the tight-tolerance #385 multistep motivation.

hiddevandebeek:
## Why

ferx has had exactly one ODE integrator: explicit Dormand-Prince RK45. On a **stiff** system an
explicit method is *stability*-limited rather than accuracy-limited — it shrinks the step to stay
stable, not to meet `ode_reltol` — so the fit either crawls or silently exhausts `ode_max_steps`
and freeze-pads the rest of the segment. That regime covers a real slice of PK/PD modelling: fast
reversible binding / TMDD-style quasi-equilibrium, Michaelis-Menten with `KM` far below observed
concentrations, long transit chains, QSP cascades.

Every peer engine either ships an auto-switching stiff integrator as its default (NONMEM ADVAN13 =
LSODA, mrgsolve DLSODA, Pumas `AutoVern7(Rodas5P)`) or makes a stiff solver a one-keyword switch
(Monolix `odeType = stiff`, Torsten `pmx_solve_bdf`). ferx was the only one with no stiff path at
all.

Closes #387 — the measured-regime gate that issue sets is now on record (see
*Numerical validation*), and its remaining track is spun out (see *Open questions*).

## What changed

**`[fit_options] ode_method = rk45 | vern7 | rosenbrock23 | rodas4 | rodas5p`**, default `rk45`.

Four new methods on **two independent axes** — because a slow ODE fit has two possible causes
and they want opposite fixes.

**Axis 1 — stability.** Three **linearly implicit Rosenbrock** steppers: one Jacobian, one LU
factorization and a fixed number of back-solves per step, with *no Newton iteration*.

- `rosenbrock23` — Shampine & Reichelt's `ode23s`, 3 stages, order 2(3), L-stable.
- `rodas4` — Hairer & Wanner, 6 stages, order 4(3), L-stable and stiffly accurate.
- `rodas5p` — Steinebach (*BIT* 63:27, 2023), 8 stages, order 5(4).

**Axis 2 — order.** `vern7`, Verner's 7(6) explicit pair (10 stages), for fits that are
*accuracy*-limited rather than stability-limited. There, step count scales as `tol^(−1/p)` and
a stiff method buys stability the fit already has; order is what pays. Measured below on the
very anchor that motivated #387.

**Why Rosenbrock and not BDF/ESDIRK/Radau.** ferx's integrator has to run twice over — at `f64`
for predictions and at `Dual2` for the analytic sensitivities — from one generic body. Because a
Rosenbrock step has no Newton iteration, it is a straight-line sequence of `+ − × ÷`, so
instantiating it at `Dual2` differentiates the discrete step exactly, with no
convergence-dependent semantics. A Newton-based method would need either the implicit-function
correction or a fully converged inner solve before its derivatives were even meaningful. For
PK-sized systems (2–20 states) the LU is also cheap enough that BDF's "one solve per step"
advantage buys nothing, and BDF loses L-stability above order 2. This is the `Dual2` story #387
asks for.

**Architecture: one `Stepper` abstraction, shared drivers.** The first cut of this PR wired the
method in at one call site and missed the one that matters — `predict()`'s walk calls
`solve_ode_dense` directly, so `ode_method` was silently ignored there (caught by the new
end-to-end test: a stiff model came back 1.5 % wrong). Rather than chase call sites, `solver.rs`
now defines a `Stepper` trait — attempt a step, score it, evaluate *your own continuous
extension* inside it — and everything else is written once, generically, over both the stepper
and the state scalar `T: PkNum`:

- `integrate_dense_g` — the single adaptive loop behind `solve_ode`, `solve_ode_dense`,
  `solve_ode_g` and the new `solve_ode_g_dense`. Step control, `saveat` clamping, `min_dt`
  force-accept, divergence break, soft sampling and statistics are all method-agnostic.
- `solve_ode_until_threshold` — event-time root-finding, now bisecting the stepper's own
  interpolant instead of a hard-coded RK45 Hermite cubic.
- `make_stepper` is the only place a method name selects an implementation.

Consequences: **every feature works with every method** — non-Gaussian endpoints (TTE /
categorical / CTMM), TTE event-time simulation, adaptive/feedback dosing, `[output]` state
columns, and the analytic-sensitivity path — with no per-method or per-feature wiring, and a
method added later inherits the whole set. The Rosenbrock methods carry real continuous
extensions for this (Rodas4/Rodas5P from their `H` blocks, `ode23s` from Shampine's quadratic),
so in-step readouts never perturb the trajectory they read from (#570's invariant, now for all
methods).

The duplicated generic RK45 loop in `solve_ode_g` is gone; it is the same driver at `T = Dual2`.

**Supporting changes**
- `sens/linsolve.rs`: the existing generic Gaussian elimination is split into reusable
  `lu_factor_in_place_g` / `lu_solve_in_place_g` so one factorization serves all stages of a step
  (`O(n³)` once, `O(n²)` per stage). `solve_linear_system_g` delegates and is **bit-identical** to
  before — verified against the old body on 4 000 random systems including ill-conditioned ones.
- `PkNum` gains a `'static` supertrait bound (every implementor is a plain value type) so a
  stepper can live behind a trait object.

## Alternatives considered

- **`diffsol`** (MIT, BDF + SDIRK, generic `Scalar` trait) — rejected as a dependency: its
  matrix layer is nalgebra/faer, whose LU requires `ComplexField`/entity traits that `Dual2`
  cannot satisfy, so it could not serve the sensitivity path. Useful as an f64 cross-check.
- **Sundials/CVODE bindings** — f64-only across the C ABI, same blocker, plus a C dependency.
- **Tsit5** — implemented, measured, and removed. Same 7 stages and FSAL structure as the
  default Dormand-Prince tableau with a smaller truncation-error constant, so it looked like a
  free upgrade; on the transit anchor at `1e-9` it needed 3 720 accepted steps against RK45's
  3 940 (~6 %) and was *slower* in wall-clock, because `Rk45Stepper` is hand-unrolled while the
  generic tableau stepper walks index loops. Vern7's 64 % step reduction clears that overhead;
  6 % does not. The measurement and the route to reviving it — express RK45 itself as an
  `ErkTableau` and retire the unrolled copy, accepting a re-anchoring of the default path — are
  recorded in `ode/explicit_rk.rs` so the work is not silently redone.
- **Serving in-step readouts by integrating *to* them** (rxode2/LSODA's event-table model) —
  implemented first, then replaced. It works and needs no interpolant, but it makes requesting a
  readout change the step sequence, which breaks #570's bit-identity invariant and the
  adaptive-dosing replay verifier. Real continuous extensions cost ~100 lines and avoid both.

## Cross-repo dependency

`ode_method` needs no ferx-r change to *work*: `ferx_fit(settings = ...)` forwards every key
straight to `ferx_core::parser::model_parser::apply_fit_option` (no allowlist of its own) and
then calls `sync_ode_solver_opts`, so `ferx_fit(settings = list(ode_method = "rodas5p"))`
already reaches the integrator. ferx-r also never constructs `OdeSolverOptions` directly, so
the new field does not break its build. The only follow-up is documentation — the roxygen
`settings` list in `R/ferx_fit.R` documents `ode_reltol`/`ode_abstol`/`ode_max_steps` and should
gain `ode_method`.

## Breaking changes
- [ ] None

<details>

Two additive public-API changes; neither affects `.ferx` files, `FitResult`, or sdtab:

1. `OdeSolverOptions` gains a `method` field. Construction via `..Default::default()` (the
   pattern used everywhere in-tree) is unaffected; an exhaustive struct literal needs the new
   field or a `..Default::default()` tail.
2. `FitOptions` gains `ode_method`, defaulting to `OdeMethod::Rk45`.
3. `PkNum` gains a `'static` supertrait bound. Every existing implementor (`f64`, `Dual1`,
   `Dual2`, `DualMixed`) already satisfies it.

</details>

## Numerical validation

- [x] Compared against: **prior ferx behaviour** (RK45 path is bit-identical) and internally
      against the RK45 engine given enough step budget to be right
- [x] Reference values:

```
Measured convergence order (fixed-step, nonlinear non-autonomous problem with exact solution
y = (sin t, t²); a mistyped tableau digit shows up here as order loss, which is what pins the
transcribed coefficients):

  rosenbrock23   observed order 2.00   (design 2)
  rodas4         observed order 3.71   (design 4, approached from below on this problem)
  rodas5p        observed order 4.45   (design 5)

Stiff binding system (k_on 1e4, k_off 1e3, k_el 0.15 — TMDD/quasi-equilibrium shape), 24 h:

  rk45,     default budget : exhausts all 10 000 steps, never reaches t = 24,
                             >1e-3 relative error at the final observation (silent freeze-pad)
  rk45,     5e6 steps      : reference, matches the analytic quasi-equilibrium value 0.4304
  rodas5p / rodas4 /
  rosenbrock23, default    : < 5 000 steps, no min-step clamps, agree with the reference to 1e-4
```

RK45 bit-identity through the refactor is asserted by the full lib suite (2 700+ tests) plus a
dedicated test that the scalar and generic drivers agree bit-for-bit at `T = f64`.

**#387's measured-regime gate** (`savic_transit_solver_regime_across_ode_methods`, run with
`--nocapture`; the harness is permanent so the numbers stay reproducible). Step counts below
are deterministic and reproduce exactly run to run; the `ms` column is the median of five runs
of a `--profile ci-test` (optimized) build on one machine and moves by tens of percent between
runs — read the step counts, and take the timings as ordering only:

```
=== Savic transit anchor @ 1e-9 (NONMEM TOL=9) ===
method           attempt    accept    reject   clamped     ms (median of 5)
Rk45                4060      3940       120         0        3.9
Vern7               1560      1420       140         0        1.7
Rosenbrock23       85300     85180       120         0       62.4
Rodas4             16320     16200       120         0       17.6
Rodas5P             4240      4060       180         0        5.8

=== Savic transit anchor @ defaults ===
method           attempt    accept    reject   clamped     ms (median of 5)
Rk45                 780       740        40         0        0.5
Vern7                580       520        60         0        0.7
Rosenbrock23        3300      3220        80         0        2.2
Rodas4              1560      1480        80         0        1.7
Rodas5P              780       740        40         0        1.1
```

Reading — this is the whole #387 decision, made from data:

- The #385 anchor is **accuracy-limited, not stability-limited**: 97 % of steps accepted, zero
  min-step clamps, for every method.
- **The stiff methods do not rescue it.** `rodas5p` takes the *same* step count as RK45 and is
  ~1.5× slower, because each step buys a Jacobian and a factorization the fit does not need;
  `rodas4` and `rosenbrock23` need 4× and 22× the steps, ordered exactly as their convergence
  orders predict.
- **Higher order does rescue it.** `vern7` takes **2.8× fewer steps and is ~2.3× faster**.
- The crossover is real and confirms `rk45` should stay the default: at *default* tolerances
  `vern7` is ~1.4× slower (10 stages vs 6 with FSAL), because there is no accuracy pressure for
  its extra order to relieve.

That answers #387's gate in the issue's own terms — and rules out ROCK2 for #385, since a
low-order stabilized explicit method is the wrong axis twice over.

**Not yet done: a NONMEM anchor for a stiff model** (ADVAN13). The stiff methods are currently
validated against the NONMEM-anchored RK45 engine, not against NONMEM directly — which is why
`docs/maturity.qmd` lists the feature as **experimental**.

## Performance
- [x] No significant impact expected — *for the default path*: `rk45` is bit-identical, so an
      existing fit is unchanged.
- Stiff methods trade `n + 1` extra right-hand-side evaluations (finite-difference Jacobian and
  `∂f/∂t`) plus one `O(n³)` factorization per step for stability. On a non-stiff 1–3 compartment
  model RK45 stays faster; the stiff methods win only where stiffness, not accuracy, caps the
  step. Documented as such.

## Tests

17 unit tests in `src/ode/rosenbrock_tests.rs` + 2 integration tests in `tests/stiff_ode_method.rs`
+ 2 new `linsolve` tests. The load-bearing ones:

- measured convergence order per method (pins the transcribed tableaus);
- `ode23s` error-estimator order (pins the stage-3 `h·γ·f_t` scaling — the `h·f_t` variant some
  implementations use is measurably an `O(h²)` estimate of an `O(h³)` error);
- stiff system RK45 cannot solve on its default budget;
- `Dual2` value part matches the scalar solve, and `Dual2` gradients match FD of the scalar solve;
- scalar and generic drivers bit-identical at `f64`;
- interpolant continuity (`Θ = 0 → y₀`, `Θ = 1 → y₁` exactly) and accuracy vs a solve landing on
  the same times;
- soft sampling does not perturb the hard saves, for all four methods;
- event-time root-finding works for all four methods;
- LU reuse across right-hand sides matches per-RHS elimination (this one caught a real bug — see
  *Reviewer hints*).

## Example

```toml
[structural_model]
  ode(obs_cmt=central, states=[central, bound])
[odes]
  d/dt(central) = -(CL/V) * central - KON * central + KOFF * bound
  d/dt(bound)   =  KON * central - KOFF * bound
[fit_options]
  ode_method = rodas5p
  ode_reltol = 1e-9
  ode_abstol = 1e-11
```

With `KON = 1e4`, `KOFF = 1e3` the Jacobian carries an eigenvalue near `−1.1e4`; `rk45` cannot
reach `t = 24 h` inside `ode_max_steps`, `rodas5p` does it in a few hundred steps.

## Docs
- [x] `docs/` updated for user-visible changes (`model-file/fit-options.qmd`,
      `model-file/ode-models.qmd` § Stiff systems, `index.qmd`, `maturity.qmd`)

## Checklist
- [x] `cargo clippy` clean (no new diagnostics in changed files)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the stiff steppers *are* on
      that path; `solve_ode_g` runs the same stepper at `Dual2`, with tests against the scalar path

## Reviewer hints

Focus here:

1. **`sens/linsolve.rs` — the permutation order.** `lu_solve_in_place_g` applies the full row
   permutation *before* any elimination (LAPACK `dgetrs` order), because `lu_factor_in_place_g`
   swaps whole rows and therefore stores `L` in permuted order. My first version interleaved the
   swaps and was wrong for any pivot at a column other than the first — invisible at `n = 2` and
   invisible on the Rosenbrock `W` (diagonally dominant at small `h`, so it rarely pivots at all).
   The residual assertion in `lu_reuse_matches_per_rhs_elimination` is what caught it.
2. **`ode/solver.rs` — the `Stepper` split.** The RK45 Butcher arithmetic moved into
   `Rk45Stepper::attempt` with its exact associations preserved (`dt·(B31·k1 + B32·k2)`, not
   `(dt·B31)·k1`), which is what keeps `T = f64` bit-identical. Worth a look that nothing drifted.
3. **`ode/rosenbrock.rs` — the tableau blocks.** Transcribed from `OrdinaryDiffEq.jl`; do not
   eyeball the digits, the order tests are the check.

Skimmable: docs, CHANGELOG, the parser/`FitOptions` plumbing.

## Open questions

1. **`vern7`'s in-step readouts are 3rd-order.** Verner's own continuous extension needs three
   extra stages and a coefficient block larger than the method; `vern7` therefore interpolates
   with the cubic Hermite (the same class RK45 uses). Steps are 7th-order accurate but reads
   *between* steps are not, so a model that leans on in-step readouts (joint PK-TTE hazard, CTMM
   occupancy, adaptive monitors) at very tight tolerance is better served by `rk45` or `rodas5p`.
   Documented, and the parity test carries a per-method interpolant bound that makes the trade
   visible rather than hidden. Transcribing the real Verner interpolant is a clean follow-up.
2. **`vern9` is now nearly free** if anyone wants `1e-12`-class work — same file, another
   tableau, and it inherits every consumer through `Stepper`. Not added because nothing in-tree
   needs that band yet.
3. **No auto-switching.** `ode_method` is a manual choice; ferx does not detect stiffness. When
   `J` is assembled (as it is for Rosenbrock) a Gershgorin bound is nearly free and more robust
   than the stage-difference heuristic, and `OdeSolverStats::min_step_clamped_steps` is already a
   usable stiffness signal. Worth a follow-up issue?
4. **Two pre-existing test failures on my box**, both unrelated and both verified as such:
   `ss_input_rate_over_capacity_deep_saturation_declines` (A/B confirmed — fails identically with
   the untouched base linear solver) and `test_iov_occasion_dsl_matches_column_fit` (fails only
   under the parallel harness, passes in isolation; every change here is a no-op when
   `method == Rk45`). Do these reproduce in CI?


hiddevandebeek:
## Correctness audit of the transcribed tableaus

Every coefficient in this PR was copied by hand from `OrdinaryDiffEq.jl`, which is the part of
the change least protected by the existing test suite: a wrong digit compiles, runs, and
produces plausible concentrations. The tableaus have now been checked along three independent
lines, chosen because each is blind to the others' failure modes.

**1. Elementwise diff against the source.** Not a value-set diff — position by position, since
a value-set comparison cannot see a coefficient placed in the wrong slot.

| Tableau | Checked | Result |
|---|---|---|
| Rodas4 | `A` (6×6), `C` (6×5), `alpha`, `d`, `H` (2×6) | exact match |
| Rodas5P | `A` (8×8), `C` (8×7), `alpha`, `d`, `H` (3×8) | exact match |
| Vern7 | `a`, `b`, `b_tilde`, `c` (58 coefficients) | exact match |

**2. Structural identities, computed from ferx's constants alone.** This owes nothing to any
source, and catches the permutation errors line 1 cannot. Rosenbrock methods are published as
`(alpha, Gamma, b)` but implemented as `A = alpha·Gamma^-1`, `C = diag(1/gamma) − Gamma^-1`;
recovering `Gamma` by forward substitution must reproduce `sum_j alpha_ij = c_i` and
`sum_j Gamma_ij = d_i`. Both hold to ~1e-15 (1.4e-15 Rodas4, 4.9e-15 Rodas5P), which binds five
independently transcribed arrays to one another. For Vern7 the classical order conditions hold
through `sum b_i c_i^6 = 1/7` (5e-13), plus the tree condition `sum b_i a_ij c_j = 1/6` and
`sum b_tilde_i = 0`. This is now a permanent test
(`rosenbrock_tableaus_satisfy_their_defining_identities`), not a scratch script.

**3. Behavioural.** Measured convergence orders Ros23 1.99 / Rodas4 3.64 / Rodas5P 4.19 /
Vern7 7.69; measured *interpolant* orders 4.86 / 4.31 / 3.41 / 4.71; and a three-way check on
the stiff binding system where the closed-form matrix exponential, rxode2 5.1.2
(liblsoda/lsoda/dop853) and ferx agree to ~1e-10.

### Test gaps this closed

Three tests were passing without testing much, and are fixed in 3dd3c0d4:

- The dense-output check asserted only endpoint continuity. `Theta=0 -> y0` and `Theta=1 -> y1`
  hold structurally for *any* `H`, so it would have passed with the Rodas dense-output
  coefficients deleted. Replaced with a measured interpolant order.
- `tests/stiff_ode_method.rs` anchored the stiff methods on ferx's own RK45 and nothing else.
  It now anchors on a closed-form matrix exponential, and checks the RK45 reference against it
  first.
- Vern7's order band was `[5.5, 7.8]` around a measured 7.69 — wide enough to pass on a method
  that had lost two orders. Tightened to `[6.5, 8.2]`.

### Timings corrected

The originally published numbers came from a debug build. They have been replaced throughout
(PR body, `CHANGELOG.md`, `docs/model-file/ode-models.qmd`) with medians of five runs of a
`--profile ci-test` build. The conclusions are unchanged and `vern7` is better than first
reported — 2.3× faster at `TOL=9`-equivalent accuracy rather than 1.4×, with `rodas5p` 1.5×
slower rather than 2.6×. The docs now state explicitly that the step counts are the
reproducible measurement and the milliseconds indicate ordering only.

### Still outstanding

No NONMEM anchor for a stiff model (ADVAN13). The stiff methods are validated against the
NONMEM-anchored RK45 engine and against closed forms, not against NONMEM directly. This is the
one gap in the validation and is unchanged by the above.

TeunP:
@hiddevandebeek : need an advan13 anchor ?

hiddevandebeek:
## Code review — `Stepper` abstraction, Rosenbrock methods, Vern7

Reviewed the full diff (`e8605a5..3dd3c0d`, 19 files). The abstraction itself is clean and the tableau constants check out; the issues below are the ones I would want resolved before merge. Two are blocking.

### Blocking

**1. `tests/stiff_ode_method.rs` does not compile — `exact_conc` is undefined** (`tests/stiff_ode_method.rs:109`, `:129`)

`cargo check --test stiff_ode_method` at the PR head fails with two `error[E0425]: cannot find function 'exact_conc' in this scope`. The analytic reference for the stiff 2-state binding system was never written, and a repo-wide grep finds no definition. Since CI runs `cargo check --tests`, this is a hard red build, and — more importantly — neither of the two new stiff tests has ever executed. They are the only end-to-end evidence that `ode_method` actually reaches the integrator and that the Rosenbrock steppers reproduce the correct trajectory.

**2. Rosenbrock stall freeze-pads the remaining save points while reporting a clean stats block** (`src/ode/solver.rs:546`)

When a Rosenbrock attempt is unusable (singular or NaN `W`) at `min_dt`, `integrate_dense_g` records a *rejected* step and breaks out of the driver. The remaining `saveat` points are then filled with the last state (lines 613–620), so the prediction is finite but wrong. The RK45 divergence path, by contrast, force-accepts at `min_dt` and increments `min_step_clamped_steps` on every clamp, eventually tripping `MAX_CONSECUTIVE_MIN_STEP_CLAMPS`.

The consequence is that the one diagnostic the new "Which regime am I in?" section of `docs/model-file/ode-models.qmd` tells users to read stays at zero for exactly the failure it is meant to surface. `tests/transit_nonmem_anchor.rs` also asserts `min_step_clamped_steps == 0` for every method as proof that no method is stability-throttled, so that assertion cannot distinguish "healthy" from "stalled and frozen" on the Rosenbrock path.

### Correctness

**3. The CHANGELOG bit-identity claim does not hold on the generic/`Dual2` path** (`src/ode/solver.rs:393`)

The f64 driver preserves the original floating-point association (its comment says so explicitly), but the generic one does not. The deleted `comb()` closure accumulated stages term-by-term as `((u + k1*(dt*B31)) + k2*(dt*B32))`; the shared `Rk45Stepper::attempt` computes `u + dt*(B31*k1 + B32*k2)`. Every caller of `solve_ode_g` — `src/sens/ode_provider.rs` has roughly 14 call sites, reached whenever `gradient_method = auto` puts an ODE model on the analytic path — therefore integrates a last-bit-different trajectory. Given the chaotic outer line search, the converged theta can differ well beyond the trailing digits, while the release notes state that existing fits are bit-identical. Either restore the association in the generic stepper or soften the claim to cover only the f64 path.

**4. `ode_method` reaches the CTMM occupancy solve, where the state count is `n²`** (`src/ode/solver.rs:90`)

`sync_ode_solver_opts` stamps `method` onto `ode.solver_opts`, and `src/markov/endpoint.rs:413` forwards those same options to `ctmm_inhomogeneous_transition_with_opts`, which integrates a row-major flattening of `dP/dt = P·Q(t)` — `s²` states for an `s`-state chain. A CTMM model with `ode_method = rodas4` therefore builds an `s²×s²` finite-difference Jacobian: for a 5-state chain, 26 extra RHS evaluations plus a 25×25 LU per step attempt, inside the per-transition loop of every likelihood evaluation. The occupancy RHS is linear and non-stiff, so this buys nothing. The documented cost model (`n + 1` extra evaluations and one `O(n³)` factorization, with `n` read as the `[odes]` state count) does not describe this case.

**5. The `ode_method` parse-error message omits `vern7`** (`src/parser/model_parser.rs:6084-6085`)

The message enumerates `rk45/rosenbrock23/rodas4/rodas5p`. `OdeMethod::parse` accepts ten tokens, including `vern7`/`verner7` plus the `dopri5`, `ros23`, `ode23s` and `rodas5` aliases. A user who mistypes the documented `vern7` is told it is not a valid value at all.

**6. `vern7` has no maturity entry** (`docs/maturity.qmd:62`)

The new row covers the Rosenbrock methods only. With no row of its own, `vern7` reads as covered by the existing `ODE models (Dormand-Prince RK45)` **beta** entry, when it is a brand-new un-anchored tableau with a documented interpolant-order mismatch for in-step readouts — the same reason its stiff siblings were marked experimental.

**7. Contradictory `vern7` benchmark numbers across three pages** (`docs/model-file/fit-options.qmd:35`)

The `[fit_options]` reference row says "~1.4× faster" at `1e-9` and "~1.7× slower" at defaults. `docs/model-file/ode-models.qmd` (3.9 ms vs 1.7 ms) and `CHANGELOG.md` both report ~2.3× faster and ~1.4× slower from the same measurement. The two figures appear to have been swapped; the canonical reference page currently understates the benefit and overstates the cost.

### Performance

**8. The Jacobian is rebuilt on every rejected retry** (`src/ode/rosenbrock.rs:545`)

`RosStepper::attempt` calls `build_and_factor_w` unconditionally, and that spends `1 + n + 1` RHS evaluations on `f0`, the `n` Jacobian columns and `f_t`. After a rejection only `dt` has changed — `u` and `t` are identical, so `J`, `f0` and `f_t` are bit-identical to what was just computed, and only the `1/(γh)` diagonal and the LU genuinely need redoing. `rodas.f` and `OrdinaryDiffEq.jl` both cache `J` across rejections. Gating the Jacobian and `f_t` build on a `(u, t)`-changed check (or a `jac_valid` flag cleared in `on_accept`) leaves only the `O(n³)` refactorization per retry, and would improve the wall-clock figures currently quoted in the docs.

**9. Non-FSAL Vern7 discards the `f_end` it just paid for** (`src/ode/explicit_rk.rs:357`)

With `dense_required` set, `attempt` evaluates `f_end = rhs(u_new, t + dt)` at line 331; `on_accept` then sets `have_k1 = false`, so the next `attempt` re-evaluates the RHS at exactly that state and time. Every accepted step on the dense path — joint PK-TTE hazard readouts, CTMM occupancy, `[output]` state columns, adaptive-dosing monitors, and all of `solve_ode_until_threshold`, which forces `set_dense_required(true)` — costs 11 evaluations instead of 10. Moving `f_end` into `ks[0]` and keeping `have_k1 = true` on the `dense_required` branch removes the 10% overhead.

### Observation

**10. The catch-all `stiff =>` arm defers exhaustiveness to a runtime panic** (`src/ode/solver.rs:305`)

`make_stepper`'s final arm is an irrefutable binding rather than a variant arm, so the next `OdeMethod` added (the `explicit_rk` module docs invite exactly this) compiles cleanly, routes into `RosStepper::new`, and hits `unreachable!("{other:?} does not use the Rosenbrock stepper")` in a worker thread mid-fit. The same pattern recurs in `method_gamma`, `method_err_exp` and `RosStepper::attempt`. Listing the three Rosenbrock variants explicitly would turn each of those into a compile error instead.

PR #665, opened 2026-07-01
feat(sens): analytic outer gradient for LTBS + time-varying covariates [#486]

hiddevandebeek:
## Why

The gate audit that motivated #663 flagged **LTBS + time-varying covariates** as an arbitrary asymmetry: the dose-superposition path serves LTBS **outer-analytic / inner-FD**, but `tvcov_analytical_supported` declined any `log_transform` model outright, sending *both* loops to FD for an LTBS + TV-cov subject. LTBS combined with covariates is a common workflow, so this left those fits on the slow FD outer gradient for no technical reason — the TV-cov walk emits the same raw prediction jet, so the outer can apply the identical `ln(f)` transform.

## What changed

- **Extracted** the inline superposition `ln(f)` jet transform (`g = ln f`, `∂g/∂x = ∂f/∂x·inv`, `∂²g/∂x∂y = inv·∂²f/∂x∂y − inv²·∂f/∂x·∂f/∂y`) into a shared `apply_ltbs_transform_outer(&mut SubjectSens, log_transform)` so the two outer paths can't drift.
- **Removed** the `|| model.log_transform` bail from `tvcov_analytical_supported`, and apply the helper **last** in `subject_sensitivities_tvcov` — after any `ScalarScale`/`ExpressionScale` quotient — reproducing production's scale-then-log order `ln(f/s)`. So LTBS composes with a TV covariate *and* an `ExpressionScale` `obs_scale`.
- **Inner stays FD**: `subject_eta_grad_tvcov` declines LTBS (belt-and-suspenders for direct callers); in the pipeline the inner is already FD for LTBS via `analytic_inner_common_bail` (covariance-Hessian stability — the deliberate design).
- **ODE path untouched** — it already applies `ln(f)` in-walk via `apply_output_transform`, so LTBS + TV-cov is already analytic there.

## Numerical validation
- [x] Compared against central FD of the log-scale production predictor (`compute_predictions_with_tv`): value, `∂f/∂η`, `∂f/∂θ`, both 2nd-order blocks — for plain LTBS + TV-cov **and** LTBS + `ExpressionScale obs_scale = 1000/V` + TV-cov (`ltbs_tvcov_outer_matches_production`).
- [x] Inner-declines assertion (`subject_eta_grad(...).is_none()` while `subject_sensitivities(...).is_some()`).
- Full library suite: **2106 passed, 0 failed** — the existing superposition LTBS tests (`provider_ltbs_matches_production`) confirm the helper extraction is behaviour-preserving.

[...]

Issue #436, opened 2026-06-20
feat(estimation): exact analytical FOCE/FOCEI covariance Hessian via Almquist sensitivity equations (replace FD)


hiddevandebeek:
## Summary

Compute the FOCE/FOCEI covariance R-matrix (Hessian of `−2logL_F`) **analytically** via the Almquist–Leander–Jirstrand sensitivity-equation framework, replacing the current finite-difference covariance step for in-scope analytical PK models. This is the natural successor to the exact analytical **gradient** shipped in #367/#381 — extending the same machinery one derivative further.

**This is an enhancement, not a bug fix.** #432 verified the current FD covariance is *correct* (it reproduces NONMEM `MATRIX=R` to <1% at the same point). An exact analytical Hessian would land at the same values — its value is **robustness, speed, and exactness**, not a different number.

## Motivation

The covariance step currently builds the Hessian by finite differences:
- default (`covariance_ofv_hessian = true`): a reconverged-OFV **central second difference** — `1/h²` noise amplification and `O(n²)` reconverged OFV evaluations;
- the `use_analytical` path: central FD of the analytic gradient — `1/h`, but still FD.

The `1/h²` path is **FD-fragile on flat/high-shrinkage directions**: e.g. at `fd_hessian_step = 2e-3` on warfarin block-omega the diagonal second difference suffers catastrophic cancellation and SE(ω²KA) blows up to ~1017 (fine at neighbouring steps). The in-code comment already flags it: *"may be over-optimistic. Try tuning fd_hessian_step"* (`src/api.rs`). An exact analytical Hessian removes the FD entirely — no step to tune, no cancellation regime, a single analytic pass instead of `O(n²)` reconverged fits.

## Approach (reuse Almquist + extend to 3rd order)

Almquist (2015) gives the exact **gradient**
`d(logL_F)/dθ = Σᵢ [ dlᵢ(η̂ᵢ)/dθ − ½ tr(H̃ᵢ⁻¹ dH̃ᵢ/dθ) ]`
[...]

Issue #893, opened 2026-07-17
CLI: show OFV before covariance step


roninsightrx:
In SAEM estimation, when running from CLI, can we compute and report the OFV before we start the covariance matrix. Sometimes based on OFV we already can cancel the covariance step.

PR #844, opened 2026-07-14
fix(estimation): FREM analytic gradients differentiate the wrong likelihood; AGQ/Laplace scope parity [#251]

hiddevandebeek:
## Why

Two things. The second was found while doing the first, and is the more serious.

### 1. AGQ/Laplace's analytic score didn't reach FOCE/FOCEI's scope (#251)

Its residual chain was hand-rolled for the Gaussian case, so `analytic_score_supported` carried **six exclusions on top of** the FOCE/FOCEI gate — M3-BLOQ, correlated residuals (`block_sigma`), FREM, `iiv_on_ruv`, custom σ magnitude, and LTBS. All six have been analytic for FOCE/FOCEI since #486; under AGQ they silently fell back to a finite-differenced score.

### 2. FREM's analytic gradients differentiate the wrong likelihood

`individual_nll` does not score a FREM covariate pseudo-observation (`FREMTYPE > 0`) against the PK model at all — its prediction is `theta[i] + eta[j]` and its variance is the dedicated `EPSCOV` error:

```rust
// stats/likelihood.rs
let frem_pred  = theta[theta_idx] + eta[eta_idx];
let frem_v     = (sigma_values[fc.covariate_sigma_index]).powi(2);
data_ll += resid * resid / frem_v + frem_v.ln();
```

But the sensitivity providers returned the ordinary **PK jet** for those rows, and the gradient assemblies read the ordinary residual variance. **Both** loops were affected — the outer gradient (FOCE/FOCEI minimising one objective while differentiating another) and the inner gradient (EBEs converging to the mode of the wrong likelihood).

This isn't a new class of bug here. The FREM-aware residual override (`build_frem_r_override`) was added **for SAEM**, and `overwrite_frem_pseudo_obs_rows` already stamps the exact `{0,1}` Jacobian into the FOCE H-matrix "so the analytic branch does not silently drop the correction the FD path performs". SAEM, importance sampling and the CWRES path all apply it. The two analytic gradients were simply never given it.

It stayed latent because FREM is conventionally fit with `method = saem`, which uses neither gradient — **and the covariate-ω ground-truth test runs SAEM too.**

## What changed

[...]

PR #953, opened 2026-07-29
feat(estimation): exact analytic FOCE/FOCEI covariance R-matrix via FD-of-Dual2 third order [#436]

hiddevandebeek:
## Why

Covariance has been the last fully-finite-difference frontier in ferx — #486 calls it *"the one every user pays on every fit"*. The R-matrix is a second difference of the **reconverged** objective: `~2·n_free²` evaluations, each re-solving every subject's inner loop, error amplifying as `ε/h²`, and an `fd_hessian_step` to tune.

The exact object needs one derivative more than the outer gradient (Almquist 2015 gives `∂F/∂x`; the observed information is `∂²F/∂x²`, third order in `f`). That is the subject of *"From exact gradients to exact standard errors: third-order sensitivity equations for the FOCE and FOCEI population likelihood"* (van de Beek et al., bioRxiv 2026), and of `feat/436-analytic-cov-hessian` — which implements it with a `Dual3` type and is 749 commits behind `main`.

This lands the same result **without `Dual3`**.

## What changed

**Third-order sensitivities by finite-differencing the exact `Dual2` jet.** `subject_sensitivities_cov` differences the second-order block along each `(θ, η)` axis:

```
∂³f/∂a∂b∂c = [ D2(x + h·e_c)[a,b] − D2(x − h·e_c)[a,b] ] / 2h
```

Because `Dual2` carries every axis simultaneously, differencing along one axis and reading the *whole* exact second-order block yields every third-order slice containing that axis — so one uniform loop fills all four tensors (`∂²f/∂θ²`, `∂³f/∂η³`, `∂³f/∂η²∂θ`, `∂³f/∂η∂θ²`) in `2N+1` calls to the existing provider per subject (`N = n_theta + n_eta`). No new kernel and no new dual type. The EBEs are reconverged **once**, at the converged point, at `cov_inner_tol` — the assembly's envelope and EBE-response terms are derived under stationarity, so it needs the modes converged to the covariance step's own tolerance rather than the fit's (review finding 8). That single solve replaces the `~2·n_free²` the stencil performs; it is also the solve `base_ofv` was already doing and discarding, so it costs nothing new.

FOCE costs twice this: it needs a second sweep at η=0 for the Sheiner–Beal linearisation point, so `2·(2N+1)` provider calls per subject. The speed figure below is the FOCEI one.

Step is `ε^(1/3)`, the textbook central-difference optimum. That optimum is only *available* because the differenced quantity is exact to machine precision; the usual reason to avoid this trick — differencing something that is itself a finite difference — does not apply to a `Dual2` jet.

| | evaluations | error amplification | inner re-solves |
|---|---|---|---|
| FD-of-OFV (current) | `~2·n_free²` reconverged objectives | `ε/h²` | every point |
| analytic (this PR) | `2N+1` provider calls / subject (FOCE: `2·(2N+1)`) | `ε/h` | one, at the converged point |

**Assembly** is ported from `feat/436-analytic-cov-hessian` — `sens_cov_hessian.rs` (the Almquist M2/M3 natural Hessian, the FOCE Sheiner–Beal variant, `pack_natural_hessian`) — and adapted to current `main`: `prepare` gained an `eta_hat` argument, `mixed_eta_theta` gained `ruv`, and `censored` moved from `Prep` onto `et[j]` as the single source. The branch's `dual3.rs` (513 lines), its `num.rs` impls and its `ode_provider.rs` plumbing are **not** ported — the FD route makes them unnecessary, which is also what kept a 749-commit rebase tractable, since those were the files that would have conflicted worst.

## Alternatives considered

**Rebasing the `Dual3` branch as-is.** Exact to machine precision rather than to `~1e-8`, but it means carrying a third-order dual type through every closed form and the ODE provider, and rebasing ~1160 lines of the most-churned files in the crate across 749 commits. The FD route reaches the same assembly with an approximation confined to the third derivative of `f` alone.

**Finite-differencing the existing analytic outer gradient.** Cheaper still (`O(n_free)` gradient evaluations, no new math at all, and the FOCEI gradient is already validated exact by `theta/omega/sigma_gradient_matches_reconverged_fd`). Not pursued here because the assembly is the more accurate object and already existed. Worth revisiting as a fallback for models this PR's scope excludes.

## Cross-repo dependency

New `[fit_options]` key `analytic_cov_hessian`; no `FitResult` / sdtab field change and no public Rust API change.

## Breaking changes
- [ ] Public Rust API
- [x] None

**Standard errors on in-scope models will change**, by the difference between an exact Hessian and a finite-difference approximation of it. That is the point of the PR, and the accurate direction — but it is a real numerical change, not a refactor.

## Numerical validation

- [x] Compared against: the reconverged FD-of-OFV stencil this replaces, at base commit `d9c6e63a`
- [x] Reference values:

```
=== #436 covariance Hessian — speed & accuracy (warfarin 1-cpt, dim=7, debug) ===
accuracy : max |analytic - reconverged-FD| = 8.17e-5   (analytic noise-free; no fd_hessian_step)
speed/subj: analytic 15.415 ms   FD-of-gradient 46.054 ms   -> 3.0x faster
```

The ported assembly's own 13 tests pass unchanged, including packed- and natural-Hessian comparisons against reconverged **precise-EBE** finite differences on both diagonal and block-Ω models, and the FOCE variants.

The four third-order tensors are pinned **three independent ways**, because the real risk is index layout — a transposed slice gives plausible finite numbers and would surface only as subtly wrong SEs:

1. **Total symmetry** of `∂³f/∂η³` under all six permutations (catches any stride error on its own).
2. **An independent route** to `∂³f/∂η²∂θ` — recomputed by differencing `∂²f/∂η∂θ` along `η` instead of `∂²f/∂η²` along `θ`, touching different jet blocks and different axes.
3. **A pure-`f64` anchor** — a triple central difference of `compute_predictions_with_tv`, the production predictor, with no duals involved at all.

Plus an assertion that the cov wrapper leaves the base jet bit-identical.

No NONMEM anchor: this computes an already-NONMEM-anchored quantity by a different route, and is validated against the incumbent stencil directly.

## Performance
- [x] Faster — 3.0× per subject on warfarin (debug), and `O(N)` rather than `O(n_free²)` in evaluations, with no inner re-solves. The gap widens with parameter count.

## Tests

`third_order_sensitivities_layouts_and_values` (Tier 1) covers the three checks above. The ported `sens_cov_hessian` test module (13 tests) covers the assembly.

`cargo test --lib`: **2768 passed, 1 failed** — `ode::predictions::tests::ss_input_rate_over_capacity_deep_saturation_declines`, which reproduces identically on clean `main`. Pre-existing, flagged rather than folded in.

- [x] Not applicable (existing fits get a better R-matrix; the new key only forces the old route)

## Docs
- [x] No new pages (existing pages amended)

Written once the scope boundary settled, which was the stated reason for deferring them. `docs/model-file/fit-options.qmd` gains an `analytic_cov_hessian` row carrying the full exclusion list, and notes that `fd_hessian_step` is inert when the analytic route serves the fit. `docs/estimation/foce.qmd`'s covariance section now describes **both** routes rather than asserting finite differences, including why the `×2` OFV scaling applies to the analytic sum and why both routes reconverge at `cov_inner_tol`.

## Checklist
- [x] `cargo clippy` clean — CI's gate (`--no-default-features --features ci,markov`) reports no errors

### ferx-site / ferx-book
- [x] New `[fit_options]` key (`analytic_cov_hessian`) → needs a `ferx-site/model-dsl/` line once the docs above are written

### Example execution
- [x] No example execution step needed (no user-visible behaviour change beyond SE accuracy)

## Reviewer hints

**Focus on the scope gate** in `subject_sensitivities_cov`. The review found five model classes it wrongly admitted (AGQ/Laplace, FREM, `Selected` error specs, non-Gaussian endpoints, and the parameter-dependent flip-flop reroute to an ODE twin), all now closed and all the same root cause: the gate had been written from scratch rather than derived from the exclusions `analytic_outer_gradient_available` and `pop_nll_opts` already encode. It is now aligned with those, and says so — but *aligned with* is not *proved complete*, and this remains the load-bearing safety property. It is *not* redundant with the provider returning `Some`: `subject_sensitivities` has grown well past this assembly's derivation — LTBS since #665/#673, expression scaling, Form-C readouts, IOV, the event-driven walk. Handing the Gaussian-only assembly one of those jets would not fail; it would return a plausible **wrong** Hessian, i.e. wrong standard errors with no symptom.

This is not hypothetical — the branch's own `analytic_cov_hessian_gates_out_of_scope` caught it during the port: LTBS stopped declining the moment the `Dual3` provider was swapped for the FD wrapper, because the underlying second-order provider happily serves it. Worth checking the gate list is complete rather than merely non-empty.

Second, `analytic_cov_hessian` in `covariance.rs` is **all-or-nothing** across subjects by design. A Hessian half-assembled from an analytic block and a finite-difference block would be neither, and the resulting SEs would be silently per-subject method-dependent.

Skimmable: `sens_cov_hessian.rs` is a port, not new work — the diff against `feat/436-analytic-cov-hessian` is the three API adaptations listed in the commit message.

## Open questions

**Where should the scope boundary sit for a first landing?** Currently: plain analytical Gaussian, non-IOV, no LTBS / scaling / readout / `iiv_on_ruv` / correlated or custom-magnitude residual / TV-cov / M3 — plus, after the review, no FREM, no covariate-`Selected` error spec, no non-Gaussian endpoint, no `method = laplace`/`agq`, and not `gradient = fd`. That is narrow — a lot of real models fall outside it and keep the FD stencil. The assembly already contains M3 machinery (`subject_cov_hessian_m3_natural`), so censoring may be closer than the gate implies; I kept the branch's own conservative boundary rather than widening it unvalidated.

**FOCEI-AGQ is reachable and currently excluded.** `agq_nodes().is_some()` is coarser than the mathematics requires. FOCEI-AGQ anchors on the Gauss-Newton `H̃`, built from *first* derivatives of `f`, so `∂²H̃/∂x²` needs third-order sensitivities — exactly what this PR computes. `method = laplace` is genuinely different: its exact anchor carries `∂²f/∂η²`, so its second derivative needs **fourth** order. Splitting the clause (to `hessian_anchor() == Exact`) once the grid-response second derivative exists would cover FOCEI-AGQ without new sensitivity machinery.

**Should this default to on?** It currently does (`analytic_cov_hessian = true`), which means every in-scope fit's SEs change on merge. The alternative is default-off for one release with an opt-in, then flip. I have no strong view; it depends how much you want the two routes compared on real fits first.

**`∂²f/∂θ²` is taken from the FD sweep**, though the `Dual2` walk already computes it exactly and `obs_sens_from_dual2` discards the θ-θ block (the same "computed and thrown away" pattern as `h_inner` in #951). Keeping it exact is free arithmetic but means touching the hot-path scatter; I left it uniform and noted it in the doc comment.


hiddevandebeek:
## Correction: three defects found by self-audit after opening this PR

None of these were caught by a failing test, which is itself the finding — the suite validated the mathematics and never exercised the wiring.

### 1. Every standard error was inflated by √2 (41.5%) — `3ddc8c1b`

`subject_packed_cov_hessian` returns `∂²Fᵢ/∂x²`, but `compute_covariance` consumes `∂²OFV/∂x²` with `OFV = 2·Σᵢ Fᵢ` (see `population_gradient_sens`'s `grad[k] += 2.0 * gi[k]`), and its `covariance = 2·H⁻¹` assumes that convention. The per-subject Hessians were summed unscaled — exactly half the right matrix.

No symptom: the result stayed symmetric, positive-definite and plausibly sized. The factor-of-2 class of #209.

**Why nothing caught it.** Every other test in the module validates the assembly against a finite difference of *its own gradient*, which is scale-blind. No test reached `compute_covariance` through the analytic path at all. Added `analytic_cov_matches_the_fd_stencil_through_compute_covariance`, which compares standard errors end-to-end against the FD stencil and fails at `4.15e-1` without the fix.

### 2. The cost claim in this PR body was wrong — `065e805c`

The table below says "no inner re-solve". That was false. `select_fd_step` probes `2·n_free` perturbed points, **each reconverging every subject's inner loop**, purely to size a finite-difference step the analytic route never uses — and it ran *before* the analytic attempt. So the analytic path was paying `2·n_free` reconverged population objectives for nothing: most of the cost this PR exists to remove.

The attempt is now made first and the step selection skipped when it succeeds. `base_ofv` is still evaluated once as a finiteness guard, which is cheap and worth keeping.

The `3.0× per subject` figure remains a microbenchmark of the per-subject assembly (`cov_hessian_speed_and_accuracy_report`) and still does **not** measure `compute_covariance` end-to-end. Treat it as a lower bound on the assembly, not a measured fit-level speedup.

### 3. The new end-to-end test could have passed vacuously — `065e805c`

Had the scope gate declined, `compute_covariance` would have fallen back to the same FD stencil and the test would have compared FD against itself — passing while proving nothing. The negative control happened to prove otherwise (removing the `×2` moved the result), but nothing *asserted* it. Now asserted at the source, so a future narrowing of the gate fails loudly instead of silently emptying the test.

---

### Still unverified, for the reviewer

- **FIX / structurally-zero parameters.** The FD stencil only fills `free_idx`; the analytic path writes the full `n×n`. Downstream zeroes FIX rows/cols and symmetrises, so I believe this is fine, but I have not tested a model with a FIX parameter through the analytic path.
- **The scope gate's completeness** — still the thing I would most want checked. It is asserted positively and narrowly, but "narrow" is not the same as "correct", and the LTBS hole (found only because the ported branch shipped its own gate test) shows the failure mode is silent wrong numbers rather than a decline.
- **The IOV pre-check** in `analytic_cov_hessian` is redundant with `subject_sensitivities_cov`'s own `n_kappa > 0` gate, and its `dose_occasions` condition is the wrong shape. Harmless but should be deleted or fixed.

hiddevandebeek:
## Code review — analytic third-order covariance R-matrix

This PR makes the analytic third-order covariance R-matrix the default path. The core concern is that the new scope gate is materially narrower than the objective it now replaces: several model classes pass every clause of the gate, receive a Hessian assembled from the wrong likelihood, and produce finite, positive-definite, plausible-looking standard errors with no warning. Because `analytic_cov_hessian` defaults to `true`, these are silent regressions relative to the FD stencil, which handled all of them correctly.

Findings are listed most-severe first.

### Correctness

**1. The analytic path never consults `options.agq_nodes()` — AGQ and Laplace fits receive a FOCE/FOCEI Hessian** (`src/estimation/covariance.rs:692`)

`compute_covariance` builds its FD stencil through `pop_nll_opts`, which dispatches to `agq_population_nll` whenever `options.agq_nodes()` is `Some` — that is, for `method = laplace` (any `n_agq >= 1`) and for `method = focei, n_agq > 1`. The documentation on `pop_nll_opts` is explicit about why: "**Every** production site that needs 'the objective for *this* fit' must call this ... An AGQ fit whose covariance step differenced the *FOCE* objective would report standard errors for a likelihood it never optimised."

The new branch bypasses that helper entirely. It dispatches on `options.interaction` alone and calls `subject_packed_cov_hessian`, which assembles the FOCEI marginal `Φ + ½log|H̃|` under the Gauss-Newton anchor. A `method = laplace` fit of a plain closed-form Gaussian model (warfarin one-compartment oral, diagonal Ω, proportional error) sits squarely inside the new scope gate and now reports SEs from the FOCEI H̃-anchored marginal rather than the exact-anchor marginal it minimised. With `n_agq > 1`, the entire quadrature correction is absent from the reported information matrix. Suggest gating on `options.agq_nodes().is_none()`.

**2. No FREM exclusion in the scope gate; the assembly re-derives residual variance without `build_frem_r_override`** (`src/sens/provider.rs:4231`, `src/estimation/sens_cov_hessian.rs:176`)

`prepare_stacked` (`sens_outer_gradient.rs:932-1067`) builds `build_frem_r_override(model.frem_config, &subject.fremtype, sigma)` and substitutes `(r, d, d2)` on rows with `fremtype > 0`, also suppressing `mult_row`/`ruv` there; `foce_subject_nll_standard` does the same via `frem_r_override`. The new module never consults `frem_config` — `inner_eta_responses` (`:549`), `subject_cov_hessian_m3_natural` (`:757`), `m3_sigma_derivs` (`:407`), `sigma_derivs` (`:197`) and `foce0_sigma` (`:975`) all call `model.error_spec.variance_at(cmt, f, sigma)` / `dvar_df` directly.

Nothing in the gate at `provider.rs:4231-4244` excludes a FREM model, and a FREM run on a closed-form one-compartment PK model with diagonal Ω passes every clause. Every covariate pseudo-observation row is then scored with `(σ_prop·f)²` from the PK error model instead of `EPSCOV²`, so the SEs for the FREM covariate ω block — the entire purpose of a FREM run — are wrong by whatever ratio those two variances differ. This is the same failure class PR #844 fixed for the gradients.

**3. Error-model lookups key on `subject.obs_cmts[j]` instead of `ErrorSpec::obs_keys(subject)`** (`src/estimation/sens_cov_hessian.rs:190`)

`ErrorSpec::obs_keys` (`types.rs:1712`) returns the CMT column for `Single`/`PerCmt`, but evaluates the covariate selector per observation for `ErrorSpec::Selected` (#669). `sens_outer_gradient.rs:471-476` states the constraint directly: "per-observation endpoint keys must come from the covariate selector (`obs_keys`), not the raw CMT column — a `Selected` spec keys endpoints by branch index, decoupled from `obs_cmts` (typically all-1 on an analytical single-endpoint model). Passing `obs_cmts` here would score every row against the wrong branch's sigma."

The new module does exactly that at `:190` (`sigma_derivs`), `:405` (`m3_sigma_derivs`), `:547` (`inner_eta_responses`), `:755` (`subject_cov_hessian_m3_natural`), and `:1063` / `:1291` (both FOCE assemblies). Nothing in the scope gate excludes a `Selected` spec, so for a model with e.g. `DV ~ proportional(SIG_ADULT) when AGE>=18 else proportional(SIG_PED)`, `prep.et` is branch-correct while the module's own `ErrD2` and σ-derivative scalars use branch 1's σ throughout — a mismatched Hessian and silently wrong σ and θ standard errors on every paediatric row.

**4. The scope gate omits `model.has_non_gaussian()`** (`src/sens/provider.rs:4231`)

`analytic_outer_gradient_available` bails on `model.has_non_gaussian()` because `foce_subject_nll` folds the non-Gaussian data term *and* its finite-difference η-Hessian into `hrh`/`log|H̃|` (`likelihood.rs:940-1010`), which the `sens/` Gaussian assembly cannot express. The new gate lists LTBS, IOV, scaling, Form-C, `analytical_init`, `iiv_on_ruv`, correlated and custom-magnitude residuals — but not `has_non_gaussian()`. A model with closed-form PK plus a `[tte_model]` or `[binary_model]` endpoint (`ode_spec.is_none()`, `pk_model = one_cpt_oral`) passes every clause, and `subject_packed_cov_hessian` returns a Hessian that omits the survival/discrete information and its `log|H̃|` curvature entirely — over-optimistic SEs with no fallback to the stencil that would have captured them.

**5. Transit/IG subjects rerouted to their ODE twin escape the intended "no ODE" bail** (`src/sens/provider.rs:4240`)

The gate's only routing check is `subject_routes_to_event_walk`, which returns `false` immediately for `*_transit`/`*_ig` models (`supports_event_driven` omits them). However, `subject_sensitivities` calls `effective_model_for_eval`, which reroutes such a subject to `absorption_ode_equivalent` both when it carries time-varying covariates/`TIME` (`effective_for`) and whenever `absorption_flip_flop_at(model, subject, theta, eta)` holds. A transit or IG subject in the flip-flop regime is therefore served by the ODE provider, and `subject_sensitivities_cov` central-differences that adaptive-step RK45 jet with `h = ε^(1/3)·(1+|x|) ≈ 6e-6`. The step-size controller changes its step sequence discontinuously under a perturbation that small, so the difference divides integrator noise (~`ode_reltol`) by `1.2e-5` and the third-order tensors return as noise. The justification given for the `ε^(1/3)` step is that "the differenced quantity is exact to machine precision", which does not hold on this path, and nothing detects the violation. Both the PR description and CHANGELOG list ODE as out of scope.

**6. Duplicate `obs_records` field in the test fixture breaks `--features survival`** (`src/estimation/sens_cov_hessian.rs:1826`)

`Subject::obs_records` is declared unconditionally in this PR (`types.rs:833`, "Compiled unconditionally (Phase 4.0)"). The literal at `:1825-1827` supplies `obs_records: vec![]`, then a `#[cfg(feature = "survival")]`-attributed `obs_records: vec![]` again. The default build strips the second line and compiles; `cargo test --lib --features survival` fails with E0062, "field `obs_records` specified more than once", taking down the entire `ferx-core` lib test target rather than just this module. This is a stale copy of the pre-Phase-4.0 pattern — deleting lines 1826-1827 resolves it.

**7. `analytic_cov_hessian` is in `framework_keys()` but has no arm in `apply_fit_option`** (`src/types.rs:6031`)

The field's own documentation offers the escape hatch: "Set `false` to force finite differences." A user who writes `[fit_options] analytic_cov_hessian = false` (or `ferx_fit(settings = list(analytic_cov_hessian = FALSE))`) falls through `apply_fit_option` (`model_parser.rs:5982-6432`) to `_ => return Ok(false)`, and the block parser turns that into a hard error: ``[fit_options]: unknown key `analytic_cov_hessian` ``. The fit aborts at parse time, and there is no route from a `.ferx` file or the R wrapper back to the finite-difference covariance that was the only path before this PR — the flag is reachable from Rust only. No test enumerates `framework_keys()` against `apply_fit_option`, so CI does not catch the gap. Given that the flag is the documented mitigation for findings 1-5, wiring it up seems worth doing before merge.

**8. `cov_inner_tol` becomes a silent no-op on the analytic path** (`src/estimation/covariance.rs:693`)

The FD stencil routes every objective evaluation, including the base point, through `reconverge_point`, which re-runs `find_ebe` at `options.effective_cov_inner_tol(...)` — the stated rationale being that "the second-difference-of-OFV R-matrix is far more sensitive to EBE precision than the fit ... any model can opt in" (`covariance.rs:577-582`). `analytic_cov_hessian` receives `eta_hats` (the modes converged during the fit at `inner_tol`) and passes them straight to `subject_packed_cov_hessian`, whose M2 envelope term `−M_ξᵀH⁻¹M_ζ` and the `inner_eta_responses` chain are both derived under the stationarity assumption `∂lᵢ/∂η|_η̂ = 0`. A user who fits at a loose `inner_tol` and then sets `cov_inner_tol = 1e-10` for trustworthy SEs now gets bit-identical results to the loose run, with the residual inner gradient contaminating the Hessian and no diagnostic.

**9. `gradient = fd` no longer suppresses analytic sensitivities in the covariance step** (`src/estimation/covariance.rs:692`)

`analytic_outer_gradient_available` (`provider.rs:942`) opens with `!matches!(model.gradient_method, GradientMethod::Fd)`, so a user who hits a bad analytic-sensitivity result sets `gradient = fd` and reverts every θ/Ω/σ derivative to finite differences. The new covariance branch is guarded only by `options.analytic_cov_hessian` and the `subject_sensitivities_cov` gate, neither of which inspects `model.gradient_method`. Such a user still receives a covariance R-matrix built from third-order finite differences of the same `Dual2` jets they opted out of, unless they separately discover the new key.

**10. The analytic Hessian loop has no cancellation checkpoint** (`src/estimation/covariance.rs:510`)

The FD stencil checked `crate::cancel::is_cancelled` inside its `par_iter` on every perturbed point and returned `CovarianceStepResult::Unusable(COV_CANCELLED_MSG)` promptly. The analytic loop iterates every subject with `2·(n_theta+n_eta)+1` provider evaluations plus an O(n_eta³·n_obs) M3 assembly each, with no check anywhere on the path, and for the default `covariance_method = r` there is no later checkpoint either. This matters for the SAEM path in particular, where `saem.rs` deliberately prints the OFV before the covariance step so a CLI user can "interrupt (Ctrl-C) before paying for the covariance matrix" (#893) — that affordance is now inoperative, and the step reports Success rather than cancelled.

### Summary

Findings 1-5 share a root cause: the scope gate for the new path was not derived from the set of exclusions that `analytic_outer_gradient_available` and `pop_nll_opts` already encode. Aligning it with those (adding `agq_nodes().is_none()`, `frem_config.is_none()`, `!has_non_gaussian()`, a `Selected`-spec exclusion or a switch to `obs_keys`, and a routing check that accounts for `effective_model_for_eval`) would close all five. Finding 7 is worth prioritising regardless, since the documented escape hatch is the stated mitigation for the rest.

Lower-priority items not detailed above: duplicated helper functions between the new module and `sens_outer_gradient`, a dead `foce_sb_fixed_natural` marked `#[allow(dead_code)]`, and the absence of a NONMEM comparison in the validation set.

hiddevandebeek:
## Review response — all ten findings confirmed and fixed

I verified every premise against the code before touching anything. All ten hold, including the
two I most wanted to be wrong (1 and 2). Findings 1–5 are, as the summary says, one mistake made
five times: I wrote the scope gate from scratch instead of deriving it from the exclusions
`analytic_outer_gradient_available` and `pop_nll_opts` had already enumerated. The gate now says
so in a comment, so the next person widening it knows where the list came from.

### Correctness

**1. AGQ / Laplace — fixed.** `agq_nodes().is_some()` now declines. The check sits in
`analytic_cov_hessian` (where `options` is in scope) rather than in the provider gate, since it
is a property of the fit, not the model. You are right that this was the worst of the ten: a
plain `method = laplace` warfarin fit is squarely in scope, and it would have reported
H̃-anchored FOCEI standard errors for the exact-anchor marginal it actually minimised — the
precise failure `pop_nll_opts`'s doc comment warns about, committed by the one site that
bypassed it.

**2. FREM — fixed.** `frem_config.is_some()` now declines. I had not made the connection to
#844, and should have: it is the same defect in the same shape one derivative up. The assembly
calls `variance_at`/`dvar_df` at five sites with no `build_frem_r_override`, so every covariate
pseudo-observation row would have been scored with `(σ_prop·f)²` instead of `EPSCOV²`.

**3. `Selected` error spec — declined rather than rewired.** `ErrorSpec::Selected` now bails at
the gate. I chose the exclusion over the `obs_keys` switch you offered as the alternative:
threading `obs_keys` through six sites is the better end state, but it is a change I could not
have validated in this PR — there is no `Selected` fixture in the module and building one
correctly is the same work as the fix. Declining is provably right today; the switch can land
with a fixture behind it. Flagged in the follow-up list below.

**4. `has_non_gaussian()` — fixed.** Added to the gate, mirroring
`analytic_outer_gradient_available`'s clause and for the identical reason. Note the clause is
not exercisable from a default build (`has_non_gaussian()` is unconditionally `false` without
`--features survival`), which I have said in the test rather than leaving it looking covered.

**5. Flip-flop reroute — fixed, and this was the one I would not have found.** You are right
that `subject_routes_to_event_walk` is not the whole routing story: `effective_model_for_eval`
reroutes a transit/IG subject to its ODE twin on a *parameter-dependent* predicate that no
model-level gate can evaluate. The gate now checks that the effective model is still closed-form
— at the base point **and at all `2N` perturbed points**, because a subject sitting on the
flip-flop boundary can be closed-form at `x` and rerouted at `x ± h`. Your diagnosis of why this
matters is exactly right and is now the comment: the `ε^(1/3)` step is justified *only* by the
differenced quantity being exact, and on an adaptive-step RK45 solution that premise is false.

**6. `--features survival` — fixed.** Stale pre-Phase-4.0 duplicate; deleted. Verified with
`cargo check --lib --tests --features survival`, which now compiles. This is a CI gap as much as
a code one — the default build cannot see it.

**7. Unreachable fit option — fixed, plus the missing backstop.** Added the `apply_fit_option`
arm. More usefully, added
`every_advertised_fit_option_key_has_an_apply_fit_option_arm`, which enumerates
`framework_keys()` and `method_specific_keys()` for all ten methods and asserts each key is
dispatched. Nothing tied those two lists together before, which is why the documented escape
hatch for findings 1–5 was itself a hard parse error.

**8. `cov_inner_tol` — fixed.** The analytic path now consumes EBEs reconverged at
`cov_inner_tol`, not the fit's modes. Since `base_ofv` already reconverged at `x_hat` and threw
the modes away, this reuses that solve rather than adding one — net zero cost, and it removes a
duplicated population inner solve on the FD path too.

**9. `gradient = fd` — fixed.** Now the first clause of the gate, matching
`analytic_outer_gradient_available`.

**10. Cancellation — fixed.** The per-subject loop checks `is_cancelled`. Returning `None` alone
would have been worse than useless (it drops into the *more* expensive FD stencil), so the
caller re-checks the flag and reports `COV_CANCELLED_MSG` instead of falling through.

### Lower-priority items

- **`foce_sb_fixed_natural`** — reachable only from this module's tests, so it is now
  `#[cfg(test)]` rather than `#[allow(dead_code)]`. A future production caller disappearing is
  now a compile error instead of a silent allow. (My first grep missed the `allow` because it is
  written `#[allow(clippy::type_complexity, dead_code)]` — you were right.)
- **Docs** — written now that the scope boundary is settled, which was the stated reason for
  deferring them: `docs/model-file/fit-options.qmd` gains an `analytic_cov_hessian` row with the
  full exclusion list, and `docs/estimation/foce.qmd`'s covariance section now describes both
  routes instead of asserting finite differences. The CHANGELOG scope list is updated with the
  five new exclusions.
- **Duplicated helpers with `sens_outer_gradient`** — acknowledged, not addressed. Consolidating
  them is a diff across the port boundary and I would rather it not ride in the same PR as five
  correctness fixes.
- **No NONMEM comparison** — this computes an already-NONMEM-anchored quantity by a different
  route, and is validated against the incumbent stencil end-to-end. I do not think a fresh
  NONMEM run adds evidence here, but say so if you disagree and I will run one.

### Tests

`analytic_cov_hessian_gates_out_of_scope` now asserts the decline for `gradient = fd`, FREM, and
`Selected`, and re-asserts that the model is back **in** scope once every flip is undone — so the
assertions cannot all be passing for some unrelated reason. Findings 4 and 5 are the two that
resist a unit test: one is feature-gated out of the default build, the other is
parameter-dependent. Both are called out in the test body rather than left looking covered.

`cargo test --lib`: 2770 passed, 1 failed — `ss_input_rate_over_capacity_deep_saturation_declines`,
which reproduces identically on clean `main`. `cargo check --lib --tests --features survival`:
clean. `cargo clippy --no-default-features --features ci,markov` (CI's gate): no errors.

### A note on finding 1's scope, and on FOCE

`agq_nodes().is_some()` is **coarser than the mathematics requires**, and deliberately so for
this PR.

FOCEI-AGQ (`n_agq > 1`) anchors on the Gauss-Newton `H̃ = Ω⁻¹ + Σⱼ pⱼaⱼaⱼᵀ`, which is built from
*first* derivatives of `f`. So `∂²H̃/∂x²` needs `∂²a/∂x²` — third order in `f`, exactly what this
PR computes. The node map `b_k = η̂ + √2·L·z_k` and its `log|H̃|` scaling differentiate through
the same objects, and #951 already established that the grid response separates cleanly with
`∂nll/∂b` independent of `k`. An analytic Hessian for FOCEI-AGQ is therefore reachable with what
is already here, and the exclusion is a scope decision rather than an impossibility.

`method = laplace` is genuinely different: its exact anchor `H = Ω⁻¹ + Σⱼ(L″aⱼaⱼᵀ + L′Aⱼ)`
contains `A = ∂²f/∂η²`, so its second derivative needs **fourth**-order sensitivities. The same
order asymmetry that separates the FOCEI and Laplace outer Hessians.

Declining both today is correct — nothing assembles the quadrature correction's second
derivative yet — but the clause is one line so splitting it later is trivial: relax to
`hessian_anchor() == HessianAnchor::Exact` once the grid-response second derivative exists.

**FOCE** is served, and was from the start — `subject_packed_cov_hessian_foce` on the
Sheiner–Beal marginal, dispatched on `!options.interaction`, with four tests (frozen and full
natural Hessian, diagonal and block Ω). None of these fixes narrowed it. Worth recording that it
costs **twice** what FOCEI does: it needs a second full third-order sweep at η=0 for the
linearisation point, so `2·(2N+1)` provider calls per subject. The `3.0×` figure in the PR body
is the FOCEI number.

### What I would still like checked

The scope gate is now aligned with two predicates that had already enumerated this territory,
which is a much better basis than the one I started from — but "aligned with" is not "proved
complete", and the failure mode is still silently wrong numbers rather than a decline. If there
is a third predicate encoding an exclusion neither of those two carries, it will not be in this
gate either.

I have also not tested a FIX parameter through the analytic path (the FD stencil fills only
`free_idx`; the analytic path writes the full `n×n`, and downstream zeroes and symmetrises).

PR #956, opened 2026-07-30
fix(fit-options): advertise cov_inner_tol as a framework covariance key; guard the reverse invariant

hiddevandebeek:
## Why

`cov_inner_tol` had a working `apply_fit_option` arm and a `fit-options.qmd` entry while appearing in **neither** advertised key list, so `unsupported_keys_warnings` told the user that a value the parser had just applied would be ignored:

```
fit option `cov_inner_tol` is not used by method `FOCEI` and will be ignored.
```

The existing guard walks advertised key → parser arm. It catches a key documented but not wired; it cannot catch the reverse, which is what shipped here.

Review of the first cut (see the review comment on this PR) found the fix behaviourally right but applied in a way that re-arms the same trap, plus six smaller issues. This PR is the fix and that follow-up.

## What changed

**The advertisement.** `cov_inner_tol` joins its five siblings — `covariance`, `covariance_method`, `covariance_fallback`, `analytic_cov_hessian`, `fd_hessian_step` — in `framework_keys()`, and the six are named as a group by a new `covariance_keys()`. This is the `ode_solver_keys()` pattern, for the same reason: framework-level with respect to the *method*, conditional on something else.

The first cut instead copied the key into nine `method_specific_keys` arms. That is inconsistent (none of the five siblings is per-method) and it re-arms the bug: `method_specific_keys` is an exhaustive `match`, so a new `EstimationMethod` variant is compiler-forced to get an arm, but nothing forces `cov_inner_tol` into it. The next covariance-capable method reproduces this exact false warning.

**The Bayes carve-out becomes a rule over the group.** Bayes reports posterior credible intervals and disables the covariance and SIR steps, so all six keys really are inert there. That is now said once, for the group:

```
fit option `cov_inner_tol` configures the post-fit covariance step, but this fit
ends in Bayesian estimation, which reports posterior credible intervals instead
of Hessian standard errors and runs no covariance step, so it has no effect.
```

Withholding a key from `method_specific_keys` produced the *wrong* diagnostic instead: "not used by method `Bayes`" is the text a **misspelled** key gets, and it lists every method-specific key as the suggested alternative. It also covered one key of six — `covariance_method = rsr` with `method = bayes` was, and would have stayed, silent.

**One implementation of "which stage runs the covariance step."** `FitOptions::covariance_stage()` calls the same `is_last_estimating_stage` that `fit_inner` calls, via a new `FitOptions::eval_only_methods()` that `fit_inner` now uses too. So the diagnostic and the run cannot disagree: `[bayes, focei]` runs the step and draws no notice; `[focei, bayes]` does not; a trailing `imp_eval_only` stage does not take ownership of it.

**Validation.** `cov_inner_tol` now parses with `parse_pos_finite`, like its sibling `fd_hessian_step`. `cov_inner_tol = -1` used to parse, flow through `effective_cov_inner_tol` into `find_ebe`'s convergence test that no subject can satisfy, and make every covariance-step EBE reconvergence burn the full `inner_maxiter` — with the SEs coming out of unconverged modes and nothing said about it.

**Guards** (all mutation-tested, see Tests):

- `every_method_advertises_the_whole_covariance_key_group` asserts **per method** over all ten variants. This is the property the reverse scraper cannot express: it unions the ten arms, so membership in any one satisfies it, and deleting the key from eight of nine arms left it green.
- The reverse scraper (`every_apply_fit_option_arm_is_advertised_or_deliberately_exempt`) now fails loudly on an indent-8 line that looks like match-arm material but does not parse as a head. There is no `rustfmt.toml`, so `max_width = 100`, and a wrapped alias chain splits into two lines neither of which survived the filter — the first has no `=>`, the second does not start with `"` — silently dropping both keys. The `> 80` floor cannot see one key of 120 go. Measured before writing the assert: 208 non-blank lines at that indent, 114 contain `=>`, all but the `_` wildcard parse, none begins with `|`. Also removes the dead `let … else` on `split().next()`, which is always `Some`.
- `EXEMPT` drops `method`/`methods`. They have no `apply_fit_option` arm at all (the function's own doc says the list-chain syntax stays in the block parser), so exempting them only pre-authorises a future arm to skip the guard.

**Docs.** The `global_maxeval` default is `0` → auto `30 * (n_params + 1)`, not `200 * (n_params + 1)`; the stale figure was also still in the source comment twelve lines above the code that computes it. `covariance_fallback` handles a non-PD covariance Hessian whether `R` came from the analytic R-matrix (#436) or the FD stencil — and the seven other statements still asserting "FD Hessian" for that same object are swept, so the page a corrected row links to no longer contradicts it. The two mentions that quote the engine's emitted warning verbatim are left alone. A new `foce.qmd` section documents which stage runs the covariance step and when none does.

## Alternatives considered

**Keep the nine per-method entries and add a per-method guard.** Rejected: the guard would pin a convention that is wrong on its own terms. `cov_inner_tol`'s applicability is decided by whether the covariance step runs, never by the estimator, so a per-method list is a fact about six keys × N methods maintained by hand — 6N chances to forget one, and forgetting one is this bug.

**Move the key to `framework_keys()` and drop the Bayes signal entirely** (matching the five siblings, none of which warns under Bayes). Rejected: the signal is correct and worth keeping; only its *spelling* was wrong. The notice generalises it to all six keys instead of losing it.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#___ | not yet open — docs-only follow-up | after |

No `pub` API changed, so ferx-r builds against its current pin either way and **no `Cargo.lock` bump is required to merge this**. The follow-up is that `R/ferx_fit.R:244` currently documents this bug as a known wart — "the engine may emit a spurious 'not used by method ... will be ignored' warning for this key; the value *is* applied" — which becomes false here, plus the two new behaviours to document. A lock bump is needed only if that PR adds a regression test, since ferx-r CI builds the pinned ferx-core.

Checked and *not* affected on the R side: `covariance` is `RESERVED` (a dedicated `ferx_fit()` argument) and is set directly on `opts` without reaching `user_set_keys`, so `ferx_fit(method = "bayes", covariance = FALSE)` does not start emitting the new notice — only keys passed via `settings` can. And `classify_warning` routes a message to `CovarianceFailed` only on "covariance step failed" / "covariance failed" / ("covariance step" **and** "not positive definite"); the notice matches none, so it lands in `general` like today's key warnings and does not draw covariance-failure remediation advice in `ferx_get_warnings()`.

## Breaking changes

- [x] None

Public API baseline is unchanged — `covariance_keys()`, `eval_only_methods()` and `covariance_stage()` are all `pub(crate)`.

## Numerical validation

No estimate, OFV, residual or diagnostic changes: this is fit-option plumbing, a warning, a parse-time bound and documentation, and every numerical path is untouched. The one behavioural edge — `cov_inner_tol = 0` / negative / `NaN` — previously produced silently unconverged covariance-step EBEs and is now rejected at parse time.

## Performance

## Tests

- [x] Regression test added — fails without the fix

Tier 1 throughout; no fit is driven.

| Test | Pins |
|---|---|
| `test_cov_inner_tol_does_not_warn` | all nine covariance-step methods apply the value and stay silent |
| `test_cov_inner_tol_reports_no_effect_for_bayes` | the value is still applied; the notice is the covariance-step one, not "not used by method" |
| `test_cov_inner_tol_rejects_non_positive_and_non_finite` | `0`, `-1e-11`, `nan`, `inf` rejected with a named error |
| `covariance_keys_are_a_subset_of_framework_keys` | the two lists cannot drift |
| `every_method_advertises_the_whole_covariance_key_group` | per method, over all ten variants |
| `bayes_chain_warns_that_covariance_keys_have_no_effect` | the notice text in full — it reaches the CLI and fit YAML verbatim |
| `bayes_chain_warns_once_per_distinct_covariance_key` | one notice per distinct key, no whitespace runs |
| `covariance_key_notice_follows_the_last_estimating_stage` | `[bayes, focei]` vs `[focei, bayes]` straddle the predicate |
| `eval_only_trailing_stage_leaves_the_covariance_step_on_bayes` | `imp_eval_only` does not take ownership |
| `every_apply_fit_option_arm_is_advertised_or_deliberately_exempt` | the reverse invariant, now loud on an unparsed arm head |

**Mutation-tested — six mutations, each reddening its own test, each reverted after:**

| Mutation | Test that died |
|---|---|
| `cov_inner_tol` out of `framework_keys()` | subset + per-method group + no-warn (3 red) |
| predicate → "chain contains Bayes" | `covariance_key_notice_follows_the_last_estimating_stage` |
| `imp_eval_only` ignored in `eval_only_methods` | `eval_only_trailing_stage_leaves_the_covariance_step_on_bayes` |
| an arm head wrapped rustfmt-style across two lines | `every_apply_fit_option_arm_is_advertised_or_deliberately_exempt` |
| `parse_pos_finite` → `parse_f64` | `test_cov_inner_tol_rejects_non_positive_and_non_finite` |
| stray double space in the notice | both `bayes_chain_*` (2 red) |

## Example

<details>
<summary>Example</summary>

```toml
[fit_options]
  method = focei
  cov_inner_tol = 1e-11
```

Before — the value is applied, and the fit warns anyway:

```
fit option `cov_inner_tol` is not used by method `FOCEI` and will be ignored.
Method-specific options for `FOCEI`: ...
```

After — no warning. Under `method = bayes`, where the covariance step genuinely does not run:

```
fit option `cov_inner_tol` configures the post-fit covariance step, but this fit
ends in Bayesian estimation, which reports posterior credible intervals instead
of Hessian standard errors and runs no covariance step, so it has no effect.
```

And a value no EBE can converge to is refused rather than silently exhausting `inner_maxiter`:

```
Error: cov_inner_tol must be a positive finite value, got 0
```

</details>

## Docs

- [x] `docs/` updated for user-visible changes — `model-file/fit-options.qmd`, `estimation/foce.qmd` (new section on which stage runs the covariance step), plus the covariance-Hessian wording sweep across `estimation/sir.qmd`, `estimation/parameterization.qmd`, `api/simulation.qmd`

No new pages, so `docs/_quarto.yml` is unchanged. `docs-lint` is green: the `### Covariance and standard errors` leaf section sits at 4939 of the 5000-character R1 limit, which is why the Bayes explanation lives in `foce.qmd` rather than inline there.

## Checklist

- [x] `tools/preflight.sh` green — fmt, 5 `cargo check` feature sets, clippy (`--all-targets`), rustdoc, docs, public-API baseline

`cargo test --lib` was green locally at 4289 passed / 0 failed before the merge with `main`. The post-merge rerun was OOM-killed by an unrelated concurrent build on the same box, so CI's `Tests + coverage` is the authority for the merged tree; the local `check` matrix (which compiles the test targets) is green on it. The merge conflicted only in `CHANGELOG.md` (both entries kept), and `main` added no `apply_fit_option` arm, no key-list entry and no `EstimationMethod` variant, so it cannot interact with the guards here.

## Docs & examples

- [x] Not applicable: `cov_inner_tol` is an existing `[fit_options]` key, its `docs/` entry updated in place — no new DSL syntax, example `.ferx`, data file or estimator
- [x] No example execution needed — no numerical behaviour changes, so no `examples/*.ferx` output moves


hiddevandebeek:

I verified the substance of the fix first, and it holds up:

- Every method this PR advertises `cov_inner_tol` for really does reach `covariance::run_covariance_step` → `compute_covariance`, which reads `options.effective_cov_inner_tol(...)` (`covariance.rs:453,773`). FOCE/FOCEI/Laplace via `outer_optimizer.rs:484/2412/2863`, GN + hybrid via `gauss_newton.rs:463/583`, SAEM via `saem.rs:3496`, IMP **and** IMPMAP via the shared `run_mcem`/`run_mcem_mixture` (`impmap.rs:1359/2325`), VI via `vi/run.rs:1045`.
- The Bayes carve-out matches behaviour: `fit.rs:1482` clears `run_covariance_step` for the Bayes stage, and `bayes.rs` never calls the step.
- The reverse guard does what it claims. I re-implemented the scraper and ran it against three source pairs: on current `main` it reports exactly `["cov_inner_tol"]` (i.e. it reddens on the bug), and on the PR — and on the PR's `types.rs` merged against current `main`'s `model_parser.rs` — it reports `[]`. 120 arm keys extracted either way, so the `> 80` floor is not near-tripping and the guard will not spuriously fail after merge.
- `global_maxeval`'s auto default really is `30 * (n as u32 + 1)` (`outer_optimizer.rs:1215-1219`), so the doc correction is right.
- Both routes (analytic `#436` R-matrix and the FD stencil) fill the same `hess` before the non-PD/`|λ|`-rectify branch (`covariance.rs:946-1010`), so the `covariance_fallback` re-wording is accurate.
- Docs-lint R1: the `### Covariance and standard errors` leaf section goes 4881 → ~4910 chars against the 5000 cap. Passes, with ~90 chars of headroom.

Findings below, roughly in order of how much I'd want them addressed.

---

### 1. `cov_inner_tol` should be a `framework_key`, not repeated across 9 method arms — medium
`src/types.rs:8151` (and the eight sibling additions)

Every other covariance-only key already lives in `framework_keys()`: `covariance`, `covariance_method`, `covariance_fallback`, `analytic_cov_hessian`, `fd_hessian_step` (`types.rs:8083-8087`), plus `sir` and the `sir_*` family. None of them warns under Bayes today, even though `fit.rs:1482` makes all of them exactly as inert there as `cov_inner_tol` is (it clears `sir` on the same two lines). So after this PR one of six covariance keys warns for Bayes and five stay silent — the carve-out is real but it is applied inconsistently, and a user who sets `covariance_method = rsr method = bayes` still gets nothing.

The more concrete cost is the maintenance trap, which is the same shape as the bug being fixed. `method_specific_keys` is an exhaustive `match`, so a new `EstimationMethod` variant is compiler-forced to get an arm — but nothing forces `cov_inner_tol` into that arm. The PR description already records having to do this by hand once ("Include the newer VI estimator in the forward and reverse key-list coverage"). Concrete regression: a future `EstimationMethod::Its` (or whatever) routes through `run_covariance_step` like every other estimating method, its author copies the SAEM arm minus a line, and `method = its` + `cov_inner_tol = 1e-11` prints "fit option `cov_inner_tol` is not used by method `ITS` and will be ignored" for a value the covariance step just applied. Neither guard fails — see finding 2.

One line in `framework_keys()` gets the correct behaviour for every current and future method, and if the Bayes distinction is worth keeping it belongs in `unsupported_keys_warnings` as an explicit "covariance keys are inert when the chain ends in Bayes" rule covering all six, not as an omission from nine lists.

### 2. The reverse guard is a union over methods, so it is strictly weaker than the bug it targets — medium
`src/parser/model_parser_tests.rs:4495-4590`

`advertised` is built by `extend`-ing all ten `method_specific_keys(m)` into one set, so membership in *any* arm satisfies it. Mutation test: delete `"cov_inner_tol"` from any eight of the nine arms this PR adds it to and `every_apply_fit_option_arm_is_advertised_or_deliberately_exempt` stays green. Only `test_cov_inner_tol_does_not_warn` reddens, and only because *its* method list happens to be hardcoded and complete right now. Per CLAUDE.md's "ask what the smallest edit that removes the fix is, and which test dies" — for a *new* method the honest answer is "none".

If finding 1 isn't taken, the guard that would actually close this is per-method rather than union: for each `EstimationMethod`, assert the covariance-key set is present iff that method runs the covariance step. That is a property the union form cannot express.

### 3. Under-matched arm heads are silently dropped; the `> 80` floor cannot see one — medium
`src/parser/model_parser_tests.rs:4534-4568`

The comment correctly identifies under-matching as the silent failure mode, but `arm_keys.len() > 80` only catches *wholesale* failure (120 keys today, so 39 could vanish unnoticed). The realistic way one key vanishes is rustfmt: there is no `rustfmt.toml`, so `max_width = 100`, and an arm head at indent 8 longer than that wraps to

```rust
        "some_long_key_name"
        | "some_other_long_key_alias" => {
```

Neither line survives the filter — the first has no `=>`, the second starts with `|` not `"` — so both keys disappear from `arm_keys` and an unadvertised key sails through. The existing head `"impmap_defensive_alpha" | "imp_defensive_alpha" => {` is already ~61 columns, so a three-way alias chain gets there.

Cheap fix that is exact today: make the skip loud. Every indent-8 line in the body that contains `=>` currently *does* parse as a literal head (I checked — zero rejects), so

```rust
assert!(looks_like_head, "unparsed arm head at KEY_ARM_INDENT: {t:?} — the scraper would silently drop this key");
```

plus an assert that no indent-8 line starts with `|` or is a bare `"..."` with no `=>`, turns a future rustfmt wrap into a red test instead of a silent hole.

Minor, same function: `let Some(head) = t.split("=>").next() else { continue; }` — `split(_).next()` on a non-empty `&str` is always `Some`, so that `else` arm is dead. `t.split("=>").next().unwrap_or(t)` or just indexing `[0]` reads more honestly.

### 4. `EXEMPT` lists two keys that have no `apply_fit_option` arm — low
`src/parser/model_parser_tests.rs:4502`

`apply_fit_option`'s own doc says "Does NOT handle `method` (which has list-chain syntax) — that stays in the block parser" (`model_parser.rs:7500`), and the scrape confirms it: `method`/`methods` are not among the 120 extracted keys. The comment's stated reason for exempting them ("select the chain itself") is therefore describing a code path that isn't in the function being scraped. Harmless today, but it pre-authorises a future `"method" => …` arm to skip the guard. `ignore` / `accept` / `ignore_subjects` are genuinely exempt — I confirmed all three `return Ok(true)` before the `user_set_keys.push` at `model_parser.rs:8181`, so they can never trigger the warning. Suggest dropping `method`/`methods` and keeping the comment for the three that are real.

### 5. The `global_maxeval` correction leaves the same stale number in the source — low
`docs/model-file/fit-options.qmd:89` → `src/estimation/outer_optimizer.rs:1642`

The doc row is now right, but the comment at the call site still reads:

```rust
// full inner-loop pass; budget is `global_maxeval` (default
// `200 * (n_params + 1)` when 0).
```

which is the exact figure this PR is correcting, twelve lines above the code that computes `30 * (n as u32 + 1)`. Worth fixing in the same PR so the next person reading the source doesn't re-introduce the doc error.

### 6. The "not FD-specific" correction leaves the adjacent statements contradicting it — low
`docs/model-file/fit-options.qmd:120`, `src/types.rs:6501`

After this change the `covariance_fallback` row says "the covariance Hessian … whether `R` is analytic or finite-difference", while:

- the row directly above it (`fit-options.qmd:119`, `covariance_method`) still says "All methods require a positive-definite **FD** Hessian";
- `docs/estimation/foce.qmd:431` — the page the corrected row links to, in the `covariance_fallback` section itself — still says "the **FD Hessian** can have every free-block eigenvalue ≤ 0 … Setting `covariance_fallback = sir`";
- `src/types.rs:5291` (`CovarianceStatus`) still documents "**FD Hessian** was non-PD; SIR was run as a fallback and succeeded".

Same sentence, same claim, same wrongness — a reader who follows the link lands on the version this PR just corrected. Either sweep the four or leave all four for a dedicated pass, but shipping one corrected and three stale is the worst of the three options.

### 7. `cov_inner_tol` accepts `0`, negatives and `NaN` — low, pre-existing
`src/parser/model_parser.rs:7666`

```rust
"cov_inner_tol" => opts.cov_inner_tol = Some(parse_f64("cov_inner_tol")?),
```

`parse_f64`, not `parse_pos_finite`. `cov_inner_tol = -1` parses, flows through `effective_cov_inner_tol` into `find_ebe`'s convergence test, and no subject ever converges — every covariance-step EBE reconvergence burns the full `inner_maxiter` and the SEs come out of unconverged modes, with no diagnostic. Its closest sibling `fd_hessian_step` (also covariance-only, also a positive float) uses `parse_pos_finite` at `model_parser.rs:7727`. Flagging it as pre-existing and consistent with `inner_tol` (`:7664`, same gap) — but this PR is what promotes `cov_inner_tol` to a first-class, nine-estimator advertised option, so it is a natural moment to validate it like `fd_hessian_step`.

---

Nothing here blocks the fix — the behaviour change is correct and the docs corrections are accurate. Findings 1 and 2 are the ones I'd want resolved before merge, because together they mean the class of bug being fixed can recur with every guard green.

hiddevandebeek:
## Review findings — resolution

| # | Finding | Resolution |
|---|---|---|
| 1 | `cov_inner_tol` duplicated into 9 method arms while its five siblings are `framework_keys()` | Taken. Moved to `framework_keys()`; the six are grouped by a new `covariance_keys()`. The Bayes distinction is kept but as a rule over the group in `unsupported_keys_warnings` — "configures the post-fit covariance step … has no effect" — so `covariance_method = rsr` under `method = bayes` is no longer silent, and the diagnostic is no longer the one a misspelled key gets. `covariance_stage()` decides it by calling the same `is_last_estimating_stage` `fit_inner` calls, via a new `FitOptions::eval_only_methods()` that `fit_inner` now uses too. |
| 2 | Reverse guard unions the arms, so it is strictly weaker than the bug | Taken, as the per-method assertion you described: `every_method_advertises_the_whole_covariance_key_group` over all ten variants. With finding 1 the concrete regression is gone anyway, but the guard now names the method if a covariance key ever drifts back into per-method lists. |
| 3 | Under-matched arm heads dropped silently; `> 80` floor cannot see one | Taken. An indent-8 line that looks like arm material but does not parse as a head now fails loudly. Exact today, measured rather than assumed: 208 non-blank lines at that indent, 114 contain `=>`, all but `_ => return Ok(false),` parse, none begins with `\|`. Verified by mutation — wrapping `"conddist" \| "saem_conddist" => …` across two lines reddens it with `"\"conddist\""` named. Dead `let … else` gone. |
| 4 | `EXEMPT` lists `method`/`methods`, which have no arm | Taken, dropped. The scrape confirms they are not among the 120 keys, and the guard stays green without them. |
| 5 | Stale `200 * (n_params + 1)` still in the source comment | Taken, `outer_optimizer.rs:1642`. |
| 6 | One corrected statement, three stale | Taken, and widened: the three named plus four more of the same class (`foce.qmd` covariance-step-failure text, `sir.qmd`, `parameterization.qmd`, `api/simulation.qmd`). The two that quote the engine's emitted warning verbatim are deliberately left. |
| 7 | `cov_inner_tol` accepts `0` / negative / `NaN` | Taken. Now `parse_pos_finite`, like `fd_hessian_step`. Left `inner_tol` alone — same gap, but changing it is a behaviour change for existing models and belongs in its own PR. |

Each new assertion was mutation-tested (six mutations, each reddening its own test and reverted after); the table is in the PR description.

One thing your review's docs-lint note did not survive contact: the `### Covariance and standard errors` section had ~90 characters of headroom, not enough for the Bayes explanation, which lives in a new `foce.qmd` section instead. The `fit-options.qmd` row gains only "Must be positive and finite." and the section lands at 4939 of 5000.

Issue #484, opened 2026-06-24
Custom residual error magnitude models, e.g. for time-varying RUV


roninsightrx:
especially time-dependent residual error magnitude, like:
```
PROP = THETA(2)
IF(TIME>24) PROP = THETA(2) * THETA(3)
```
but also other custom models that depend on a theta

roninsightrx:
## Plan: custom / time-varying residual-error magnitude

Investigated. Today residual variance is a pure function of `(error_model, f, sigma)` — `V = Σ(loading_k·σ_k)²`, loading ∈ {1, f}. Nothing time- or theta-dependent can enter, so the NONMEM idiom

```nonmem
PROP = THETA(2)
IF(TIME.GT.24) PROP = THETA(2) * THETA(3)
W = SQRT(THETA(3)**2 + PROP**2 * IPRED**2)
```

is inexpressible. (`busulfan_shukla` shows the same class of gap for a *structural* param: its within-subject `DAY1` TIME>24 CL effect had to be smuggled into `[odes]` because `[individual_parameters]` are subject-static — there's no equivalent escape hatch for the error model.)

### Target syntax

Each sigma argument in `[error_model]` may be an **expression** over thetas / covariates / `TIME`/`TAD`/`TAFD` / individual params, reusing the `[scaling]` expression evaluator. A bare sigma name keeps today's exact behaviour.

```ferx
[parameters]
  sigma PROP_ERR ~ 0.1 (sd)
  sigma ADD_ERR  ~ 22.0 (sd)
  theta RUV_LATE(1.5, 0.0, 10.0)

[error_model]
  DV ~ combined(PROP_ERR * (if (TIME > 24.0) RUV_LATE else 1.0), ADD_ERR)
```

Semantics — the expression is a per-observation **multiplier on that sigma's loading**:

```
V_i = Σ_k (loading_k,i · m_k(TIME_i, cov_i, θ) · σ_k)²   (+ block_sigma cross terms)
```

Bare name ⇒ `m_k ≡ 1`, so existing models stay bit-identical.

### Design (minimal-churn)

[...]

PR #963, opened 2026-08-04
fix(stats): make residual-variance derivatives floor-aware [closes #958]

### LINKED ISSUE #958: bug(estimation): analytic (Dual2) FOCEI objective diverges from FD when a parameter-dependent init() meets a sqrt binding quadratic
The analytic (`Dual2`) FOCEI objective and the FD objective disagree **at the same parameter values** when a model combines two things:

1. a state whose initial condition is an expression in a model parameter — `init(<state>) = <parameter>`, and
2. a `sqrt` binding quadratic in the ODE block.

Either one alone is fine. Together they diverge. The objective is not only what the optimizer minimises, it is what ΔOFV/ΔAIC model selection compares, so a gradient-path-dependent objective changes which model wins with nothing in the output indicating a problem.

cc @hiddevandebeek @roninsightrx

Versions: `ferx` 0.2.0 (core 0.2.0), R 4.5.1, macOS.

## Reproducer

Two states, six thetas, two etas, six subjects, noise-free synthetic data. `optimizer = bobyqa` is **pinned** in every run: without that, changing `gradient` also changes the optimizer, because `optimizer = auto` resolves to NLopt L-BFGS when an analytic gradient is available and BOBYQA otherwise (`types.rs:5564`), and the comparison would be confounded.

Evaluation 1 is the starting point, so the objective there is a property of the model and the initial estimates — it cannot depend on how the gradient is computed.

| ODE form | `init(RTOT)` | `gradient = fd` | `gradient = auto` | auto / fd |
|---|---|---|---|---|
| Michaelis-Menten | `20` | −57.85903 | −57.85889 | 1.000000 |
| Michaelis-Menten | `RBASE` | −42.15390 | −42.15373 | 1.000000 |
| binding quadratic | `20` | 329.59377 | 329.59383 | 1.000000 |
| **binding quadratic** | **`RBASE`** | **324.81958** | **325.90558** | **1.003343** |

Three cells agree to ~1e-6. The fourth is off by 0.33 % — a thousand times the agreement floor of the other three, so this is not solver noise.

The two ODE blocks differ only in how free drug is obtained:

```
# Michaelis-Menten — agrees
  ct = CENT / VC
  fb = ct / (KM + ct)
  d/dt(CENT) = -KEL * CENT - KINT * fb * RTOT * VC
  d/dt(RTOT) =  KSYN - KDEG * RTOT - KINT * fb * RTOT

# binding quadratic — disagrees, but only with a parameter-dependent init
  ct = CENT / VC
  bb = ct - RTOT - KM
  cf = 0.5 * (bb + sqrt(bb*bb + 4*KM*ct))
  fb = cf / (KM + cf)
  d/dt(CENT) = -KEL * cf * VC - KINT * fb * RTOT * VC
  d/dt(RTOT) =  KSYN - KDEG * RTOT - KINT * fb * RTOT
```

### `repro.ferx`

Swap `init(RTOT) = RBASE` for `init(RTOT) = 20` and `gradient = auto` for `fd` to walk the four cells.

```
[parameters]
  theta KEL(0.05, 0.001, 10)
  theta VC(5, 0.5, 100)
  theta RBASE(20, 1, 400)
  theta KINT(0.2, 0.001, 20)
  theta KDEG(0.02, 0.0001, 5)
  theta KM(10, 0.1, 1000)

  omega IIV_KEL   ~ 0.09
  omega IIV_RBASE ~ 0.09

  sigma PROP_DRUG ~ 0.15 (sd)
  sigma PROP_TGT  ~ 0.15 (sd)

[individual_parameters]
  KEL   = KEL * exp(IIV_KEL)
  RBASE = RBASE * exp(IIV_RBASE)
  VC    = VC
  KINT  = KINT
  KDEG  = KDEG
  KM    = KM
  KSYN  = RBASE * KDEG

[structural_model]
  ode(states=[CENT, RTOT])

[odes]
  init(RTOT) = RBASE
  ct = CENT / VC
  bb = ct - RTOT - KM
  cf = 0.5 * (bb + sqrt(bb*bb + 4*KM*ct))
  fb = cf / (KM + cf)
  d/dt(CENT) = -KEL * cf * VC - KINT * fb * RTOT * VC
  d/dt(RTOT) =  KSYN - KDEG * RTOT - KINT * fb * RTOT

[scaling]
  y[CMT=1] = CENT / VC
  y[CMT=2] = RTOT

[error_model]
  CMT=1: DV ~ proportional(PROP_DRUG)
  CMT=2: DV ~ proportional(PROP_TGT)

[fit_options]
  method     = focei
  maxiter    = 2
  ode_reltol = 1e-9
  ode_abstol = 1e-9
  gradient   = auto
  optimizer  = bobyqa
  covariance = false
```

### `repro.csv`

102 rows: six subjects at three dose levels, drug in `CMT=1` and total target in `CMT=2`, sampled at 1/3/7/14/28/56/84/120. Generated noise-free at the nominal values with `rxode2`:

```r
library(rxode2); library(data.table)
KEL <- 0.05; VC <- 5; RBASE <- 20; KINT <- 0.2; KDEG <- 0.02; KM <- 10
m <- rxode2({
  ct = CENT / VC
  fb = ct / (KM + ct)
  d/dt(CENT) = -KEL * CENT - KINT * fb * RTOT * VC
  d/dt(RTOT) =  KSYN - KDEG * RTOT - KINT * fb * RTOT
})
tms <- c(1, 3, 7, 14, 28, 56, 84, 120)
d <- rbindlist(lapply(1:6, function(i) {
  amt <- c(100, 100, 300, 300, 900, 900)[i]
  ev  <- data.frame(id = 1L, time = c(0, tms),
                    amt = c(amt, rep(NA_real_, length(tms))),
                    evid = c(1L, rep(0L, length(tms))), cmt = "CENT")
  s <- as.data.table(rxSolve(m, c(KEL=KEL, VC=VC, KINT=KINT, KDEG=KDEG, KM=KM,
                                  KSYN = RBASE*KDEG),
                             ev, inits = c(CENT = 0, RTOT = RBASE),
                             atol = 1e-10, rtol = 1e-10))[time %in% tms]
  rbind(data.table(ID=i, TIME=0,   AMT=amt, EVID=1L, MDV=1L, CMT=1L, DV=0),
        data.table(ID=i, TIME=tms, AMT=0,   EVID=0L, MDV=0L, CMT=1L, DV=round(s$CENT/VC, 4)),
        data.table(ID=i, TIME=tms, AMT=0,   EVID=0L, MDV=0L, CMT=2L, DV=round(s$RTOT, 4)))
}))
fwrite(d, "repro.csv")
```

First rows:

```
ID,TIME,AMT,EVID,MDV,CMT,DV
1,0,100,1,1,1,0
1,1,0,0,0,1,16.6612
1,3,0,0,0,1,11.5968
1,7,0,0,0,1,5.6143
```

**One trap worth flagging for anyone shrinking this further.** An earlier version of this reproducer used `KDEG = 0.5` with samples out to 24, so `d/dt(RTOT) = KDEG*(RBASE − RTOT)` relaxed to `RBASE` within a couple of time units. The trajectory forgets its initial condition long before the first sample, the sensitivity with respect to that initial condition is ≈ 0, and both paths agree trivially. The target has to still be moving across the sampling window — in the data above it falls to 11 % of baseline.

## Where it gets large

Found on a bigger model of the same class: two-compartment, a target state initialised at a baseline parameter carrying IIV, drug and total target observed. Bisecting that model, `optimizer = bobyqa` pinned throughout and ODE tolerances at 1e-9, first-evaluation objective:

| variant | fd | auto | auto / fd |
|---|---|---|---|
| `init(RTOT) = <constant>` | 7 878.578 | 7 883.926 | 1.0007 |
| no `init()` at all | 1.281542e8 | 1.288076e8 | 1.0051 |
| `init(RTOT) = RBASE`, **no IIV** on RBASE | 41 658.74 | 8 515.532 | **0.2044** |
| `init(RTOT) = RBASE`, IIV on RBASE | 70 943.44 | 1.540199e7 | **217.10** |

So a θ-dependent init already diverges by ~5× with no random effect on that θ, and adding IIV takes it to ~217×. Both the θ-seed and the η-seed of the initial state look wrong, the η one worse.

The sub-percent differences in the two control rows are what one would expect benignly: the FOCEI objective depends on each subject's inner EBE solution, the inner loop consumes the gradient, so the two paths can land on marginally different η̂ at a poorly-fitting starting point.

## Why this looks like a scope gap rather than a wrong formula

`CLAUDE.md:94` states the invariant and the required fallback:

> Every change to an analytic sensitivity, gradient, marginal, or likelihood path requires a `Dual2`-vs-FD parity test. […] A wrong sensitivity compiles and runs silently — there is no second copy of the formula to disagree with it. […] If a model is outside the analytic scope it must route to FD via the support predicates (`sens_supported` / `analytic_inner_grad_supported_model`); unit-test that routing so a scope gap fails loudly to FD instead of silently returning a wrong gradient.

#439 (`ODE analytic sensitivities: remaining scope`) tiers the known out-of-scope features — TV-covariates, IOV, estimated lagtime, steady state, `input_rate`, SDE, expression `obs_scale`. **Initial conditions do not appear anywhere in it.** So `init(<state>) = <parameter>` appears to be neither armed nor listed as routing to FD; it takes the analytic path and returns a different objective.

Suggested order of work:

1. Add `init()` expressions that reference θ or η to the support predicate, so these models route to FD — the loud-fallback behaviour #439's plan describes for everything else out of scope.
2. Add a `Dual2`-vs-FD parity test with a parameter-dependent `init()` **and** a `sqrt` in the ODE block to the `check_full_provider_vs_fd` harness. The four-cell design above is offered for it: three cells must agree, and testing only the smooth-RHS cell would have passed.
3. Then seed `∂state(0)/∂θ` and `∂state(0)/∂η` in the sensitivity walk, and the fallback from (1) can be removed.

## Consequences seen downstream

On the larger model, `gradient = auto` — the default — produced a fit reporting:

- `converged: FALSE`, "Outer optimization did not converge"
- standard errors implying 2.4–10.9 %RSE on the affected parameters
- point estimates 27–43 % away from NONMEM 7.6 and nlmixr2 on three parameters where **those two agree with each other to within 0.4 %**

`gradient = fd` on the same model does not show that displacement. Since `auto` is the default and resolves to the analytic path, a user hits this without opting into anything.

## Two smaller things found alongside

**1. `gradient = ad` is accepted by the parser and rejected by the engine.** The parser's own error text still advertises it (`model_parser.rs:6266`, "expected 'auto', 'ad', or 'fd'"), but a fit fails with:

```
Fit error: gradient_method = ad is no longer supported: the Enzyme
automatic-differentiation path was retired in favour of the analytic `Dual2`
sensitivities. Set gradient_method = auto ... or fd.
```

Probably a leftover of the #428 scrub — the parse-time message could name only `auto`/`fd`, or `ad` could be accepted as an alias for the analytic path.

**2. The FOCEI polish phase of `method = gn_hybrid` appears to make almost no progress.** On the larger model, with `verbose = true`:

- the GN phase printed all 328 iterations and converged (OFV 2444.891795, `delta=1.13e-7`, trust radius → 0)
- the FOCEI polish then printed **one** evaluation, at the same objective, and produced no second evaluation in the following ~927 s before the run was stopped

The same model and settings under `method = focei` from a cold start average ~5.4 s per evaluation, so the polish is at least two orders of magnitude slower per evaluation. Stated as a lower bound: I did not wait for a second evaluation, so I cannot report a rate — only that none arrived in ~15 minutes with verbose printing on.

roninsightrx:
## Why

`gradient = auto` (analytic `Dual2` inner/outer gradients) and `gradient = fd` produced **different FOCE/FOCEI objectives for the same model at the same estimates** (#958). On the issue's reproducer the first-evaluation OFV was auto `325.90558` vs fd `324.81958` — a 0.33% gap, ~1000× the agreement floor of the control cells. Because the objective is what ΔOFV/ΔAIC model selection compares, this makes model choice gradient-path-dependent with nothing in the output flagging it. `auto` is the default, so users hit it without opting in.

Root cause: the residual-variance floor `MIN_VARIANCE = 1e-12` (which clamps `(f·σ)²` so a vanishing prediction cannot produce a zero variance) is applied in `variance_at`/`residual_variance` but **not** in the derivative accessors `ErrorSpec::dvar_df` / `d2var_df2` (and their `_scaled` variants). Those returned the raw `2·f·σ²` / `2·σ²` even where the variance is clamped constant in `f` (true derivative 0).

On a proportional-error row whose prediction is driven to ~0 (drug fully eliminated at a late sample), the objective's variance is floored and locally constant, so `∂R/∂f = 0`. The analytic inner-EBE gradient used the raw nonzero slope → its per-observation coefficient diverged from FD of `individual_nll` → the EBE search landed on a different mode → different FOCEI objective.

**The issue misdiagnosed this** as a θ-dependent `init()` + `sqrt` sensitivity scope gap. The ODE prediction sensitivities `∂f/∂η` are correct throughout (verified vs FD across an η grid); init sensitivities are already seeded (`init_fd_derivs`). The `init(θ)` + `sqrt` combination is only the *trigger* that pushes a drug row below the floor with non-negligible `∂f/∂η`. Bisection with a per-observation FD confirmed exactly **one** row (the eliminated-drug tail sample) carried the entire gradient error.

## What changed

`dvar_df` / `d2var_df2` (and `dvar_df_scaled` / `d2var_df2_scaled`) return `0` when `variance_at(cmt, f, sigma) <= MIN_VARIANCE` — i.e. exactly where the raw variance was clamped, so `∂v/∂f = ∂²v/∂f² = 0`. `d2var_df2` gained an `f` parameter to test the floor; every call site (`sens_cov_hessian`, `sens_outer_gradient`, `gauss_newton`) already had the prediction in hand. Above the floor the accessors are unchanged (bit-identical to `2·f·σ²` / `2·σ²`), so no non-floored fit moves.

## Alternatives considered

The issue suggested routing θ/η-dependent `init()` to FD via a support predicate. Rejected: the init sensitivities are already correct and are not the cause, so an FD fallback would mask the real bug **and** needlessly disable the (correct) analytic gradient for every parameter-dependent-init model. Fixing the floor-unaware derivative is the general, minimal fix — it also covers any proportional-error model with a near-zero prediction, ODE or closed-form.

## Numerical validation

- [x] Compared against: `gradient = fd` on the same model (the FD objective is the anchor — it already sees the clamped variance) and prior ferx commit.

```
# issue reproducer, first-evaluation OFV
# before
gradient=auto : 325.905579
gradient=fd   : 324.819579     (auto/fd = 1.003343)
# after
gradient=auto : 324.808689
gradient=fd   : 324.808711     (agree to ~7e-8)
```

Per-subject inner EBEs also match between paths after the fix (they diverged before — e.g. subject 3 EBE `[0.517, -0.101]` under auto vs `[0.522, -0.152]` under fd).

## Breaking changes
- [x] None

## Performance
- [x] No significant impact expected (the added `variance_at` call in the derivative accessors is a few flops on a non-hot path relative to ODE integration).

## Tests
- [x] Unit test added (`cargo test --lib` passes — 2801 pass, 0 fail)

Both new tests fail on the pre-fix code (verified by neutralizing the floor gate):
- `stats::residual_error::tests::dvar_df_and_d2var_df2_are_floor_aware` — pins the derivatives to 0 at a floored prediction and to central FD of `variance_at` above the floor.
- `estimation::inner_optimizer::tests::analytic_inner_gradient_ode_init_theta_sqrt_matches_fd` — the exact reported scenario (two-state binding ODE, `init(RTOT) = RBASE`, `sqrt` quadratic, per-CMT proportional error); analytic inner η-gradient must match FD of `individual_nll` at a non-zero η that floors the drug tail sample.

## Docs
- [x] No user-visible API change (behavioural correctness fix)

## Checklist
- [x] `Dual2` sensitivity path stays differentiable — the change is in the residual-variance derivative accessors, not the `*_g<T: PkNum>` kernels; analytic ↔ FD parity is now *restored*, not broken

## Not in scope

The issue's two "smaller things" are left as follow-ups (unrelated to the root cause): (1) `gradient = ad` accepted by the parser but rejected by the engine — the parse-time message still advertises `ad`; (2) `gn_hybrid`'s FOCEI polish phase appearing near-stalled. Happy to open separate issues/PRs.

## Reviewer hints

Focus on `src/stats/residual_error.rs` — the two floor gates and the `d2var_df2` signature change. The call-site edits are mechanical (thread the already-available `f` into `d2var_df2`); the value at each site is the same prediction already passed to the adjacent `variance_at`/`dvar_df`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,6 +159,17 @@ section of the SDLC for the versioning policy).
   for any external code that constructs or exhaustively destructures these variants.
 
 ### Fixed
+- **Gradient-path-dependent FOCE/FOCEI objective on proportional-error models with a
+  near-zero prediction** (#958). The residual-variance floor (`MIN_VARIANCE`, which clamps
+  `(f·σ)²` so a vanishing prediction cannot produce a zero variance) was applied to the
+  variance but **not** to its analytic derivatives `∂R/∂f` / `∂²R/∂f²`. On a row whose
+  prediction is driven to ~0 (e.g. drug fully eliminated at a late sample) the variance is
+  clamped and locally constant in `f`, so its true derivatives are 0, but the accessors
+  returned the raw `2·f·σ²` / `2·σ²`. That made the analytic inner-EBE gradient disagree with
+  the finite-difference objective it minimises, so `gradient = auto` (analytic) and
+  `gradient = fd` could converge to different empirical-Bayes modes and report different
+  objective values (hence different ΔOFV/ΔAIC) for the same model at the same estimates. The
+  variance-derivative accessors are now floor-aware, restoring analytic ↔ FD agreement.
 - **Standard errors for closed-form LTBS models under IOV** (#486). The tighter inner-EBE
   tolerances that log-transform-both-sides models take for the fit and covariance steps
   (`LTBS_FIT_INNER_TOL` / `LTBS_COV_INNER_TOL`, #665) were previously skipped whenever the
--- a/src/estimation/gauss_newton.rs
+++ b/src/estimation/gauss_newton.rs
@@ -1151,8 +1151,9 @@ fn subject_nll_pop_grad_analytical_laplace_cached(
     let d_vec: Vec<f64> = (0..n_obs)
         .map(|j| error_spec.dvar_df(err_keys[j], ipreds[j], sigma_values))
         .collect();
-    // ∂²R/∂f² per observation. f-independent for additive/proportional/combined
-    // (0 for additive, 2·σ_prop² otherwise), but the *per-CMT* value can differ
+    // ∂²R/∂f² per observation. Constant for additive/proportional/combined
+    // (0 for additive, 2·σ_prop² otherwise) except where the variance floor
+    // clamps it to 0 (#958), but the *per-CMT* value can differ
     // across observations: e.g. an Emax PK/PD model with proportional error on
     // PK (CMT=2) and additive on PD (CMT=3) needs `d2 = 2·σ_prop²` at PK obs
     // and `d2 = 0` at PD obs. Dispatch through `error_spec.d2var_df2` so the
@@ -1160,7 +1161,7 @@ fn subject_nll_pop_grad_analytical_laplace_cached(
     // the cmt argument and returns the scalar value uniformly. The β_j chain
     // at line ~1095 then reads `d2_vec[j]` per obs.
     let d2_vec: Vec<f64> = (0..n_obs)
-        .map(|j| error_spec.d2var_df2(err_keys[j], sigma_values))
+        .map(|j| error_spec.d2var_df2(err_keys[j], ipreds[j], sigma_values))
         .collect();
 
     // Conditional Hessian H̃ = a'·diag(1/R)·a + ½·c̃'·c̃ + Ω⁻¹.
--- a/src/estimation/inner_optimizer_tests.rs
+++ b/src/estimation/inner_optimizer_tests.rs
@@ -464,6 +464,95 @@ fn analytic_inner_gradient_m3_matches_fd_on_warfarin_bloq() {
     }
 }
 
+/// #958, as reported: a two-state target-binding ODE with a **parameter-
+/// dependent initial condition** (`init(RTOT) = RBASE`) and a `sqrt` binding
+/// quadratic in the RHS, observed at two endpoints under proportional error.
+/// The analytic ODE inner η-gradient must match a central FD of `individual_nll`
+/// at a non-zero η that pushes the drug endpoint's late sample into the
+/// variance floor. This is the exact scenario the issue bisected to; the raw
+/// four-cell design isolates it to `init(θ)` + `sqrt`, but the mechanism is the
+/// floor-unaware `∂R/∂f`, so the reproduction is the analytic-vs-FD inner
+/// gradient here (the prediction sensitivities themselves were never wrong).
+#[test]
+fn analytic_inner_gradient_ode_init_theta_sqrt_matches_fd() {
+    use std::cell::RefCell;
+    let model = crate::parser::model_parser::parse_model_string(
+        "[parameters]\n  theta KEL(0.05,0.001,10)\n  theta VC(5,0.5,100)\n  theta RBASE(20,1,400)\n  theta KINT(0.2,0.001,20)\n  theta KDEG(0.02,0.0001,5)\n  theta KM(10,0.1,1000)\n  omega IIV_KEL   ~ 0.09\n  omega IIV_RBASE ~ 0.09\n  sigma PROP_DRUG ~ 0.15 (sd)\n  sigma PROP_TGT  ~ 0.15 (sd)\n[individual_parameters]\n  KEL   = KEL * exp(IIV_KEL)\n  RBASE = RBASE * exp(IIV_RBASE)\n  VC    = VC\n  KINT  = KINT\n  KDEG  = KDEG\n  KM    = KM\n  KSYN  = RBASE * KDEG\n[structural_model]\n  ode(states=[CENT, RTOT])\n[odes]\n  init(RTOT) = RBASE\n  ct = CENT / VC\n  bb = ct - RTOT - KM\n  cf = 0.5 * (bb + sqrt(bb*bb + 4*KM*ct))\n  fb = cf / (KM + cf)\n  d/dt(CENT) = -KEL * cf * VC - KINT * fb * RTOT * VC\n  d/dt(RTOT) =  KSYN - KDEG * RTOT - KINT * fb * RTOT\n[scaling]\n  y[CMT=1] = CENT / VC\n  y[CMT=2] = RTOT\n[error_model]\n  CMT=1: DV ~ proportional(PROP_DRUG)\n  CMT=2: DV ~ proportional(PROP_TGT)\n[fit_options]\n  method     = focei\n  ode_reltol = 1e-9\n  ode_abstol = 1e-9\n",
+    )
+    .expect("parse binding-quadratic ODE with parameter-dependent init");
+
+    // Drug (CMT 1) at 4 times incl. a far-tail sample where drug ~ 0 (floored
+    // proportional variance), then total target (CMT 2) at the same times.
+    let tms = [1.0, 7.0, 28.0, 120.0];
+    let mut obs_times = Vec::new();
+    let mut obs_cmts = Vec::new();
+    for &t in &tms {
+        obs_times.push(t);
+        obs_cmts.push(1usize);
+    }
+    for &t in &tms {
+        obs_times.push(t);
+        obs_cmts.push(2usize);
+    }
+    let n = obs_times.len();
+    let subject = Subject {
+        id: "1".into(),
+        doses: vec![DoseEvent::new(0.0, 300.0, 1, 0.0, false, 0.0)],
+        obs_times,
+        obs_raw_times: Vec::new(),
+        // Rough noise-free-ish values; exact magnitudes are immaterial — the
+        // analytic gradient is checked against FD of the objective on the SAME
+        // data, so the assertion is a self-consistency (gradient-path) check.
+        observations: vec![
+            30.0, 5.0, 0.5, 0.0, // drug
+            16.0, 10.0, 6.0, 15.0, // target
+        ],
+        obs_cmts,
+        covariates: HashMap::new(),
+        dose_covariates: Vec::new(),
+        obs_covariates: Vec::new(),
+        pk_only_times: Vec::new(),
+        pk_only_covariates: Vec::new(),
+        reset_times: Vec::new(),
+        cens: vec![0; n],
+        occasions: vec![1; n],
+        obs_l2: Vec::new(),
+        dose_occasions: Vec::new(),
+        fremtype: Vec::new(),
+        obs_records: vec![],
+    };
+
+    let theta = &model.default_params.theta;
+    let omega = &model.default_params.omega;
+    let sigma = &model.default_params.sigma.values;
+    let eta = vec![0.5, -0.15];
+
+    // Precondition: the drug tail sample really does hit the variance floor.
+    let preds = crate::pk::compute_predictions_with_tv(&model, &subject, theta, &eta);
+    assert!(
+        model.residual_variance_at(1, preds[3], sigma) <= 1e-12,
+        "drug tail prediction {} must drive its proportional variance to the floor",
+        preds[3]
+    );
+
+    let analytic = analytic_eta_nll_gradient(&model, &subject, theta, &eta, omega, sigma)
+        .expect("analytic ODE inner gradient must be supported");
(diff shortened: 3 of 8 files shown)


ronkeizer:
## Code review (high-effort, workflow-backed) — 1 correctness defect (2 facets) + 2 cleanups

The floor-awareness fix is correct for the bare-σ path but leaves the **custom RUV-magnitude (#484)** path exposed to the exact bug it set out to close.

### 🔴 Correctness — scaled derivatives floor on the *unscaled* variance

`dvar_df_scaled` (`src/stats/residual_error.rs:357`) and `d2var_df2_scaled` (`:421`) delegate to `dvar_df`/`d2var_df2`, whose floor gate tests `variance_at(cmt, f, sigma)` — the **unscaled** variance. But `residual_rd`/`residual_rd2` pair these derivatives with `variance_at_scaled(cmt, f, sigma, &[], m)` (multiplier folded in *before* the floor). Whenever the magnitude multiplier `m ≠ 1`, the derivative's floor boundary sits at a different `f` than the variance's:

- For `f` in the band `1e-6/(m·σ) < f ≤ 1e-6/σ` with `m > 1`, the paired variance `(f·m·σ)² > MIN_VARIANCE` (objective sees a live, non-constant variance) but the derivative gates on `(f·σ)² ≤ MIN_VARIANCE` and returns **0**.
- `residual_rd2` then feeds a zeroed analytic `∂R/∂f` into the inner-EBE gradient while `individual_nll`'s FD objective sees the real derivative → `gradient=auto` and `gradient=fd` can again converge to **different EBE modes / OFV** on a magnitude model. That is the gradient-path-dependence #958 was meant to eliminate — still present on the scaled path.

**Facet 2 (`:421`, covariance/SEs):** `d2var_df2_scaled` inherits the same unscaled gate → the M3 covariance Hessian contracts a spurious non-zero curvature (`2·m²·σ² ≠ 0` where the scaled variance is floored constant), biasing the reported standard errors for that endpoint.

**Fix:** gate the `_scaled` accessors on the same scaled variance they pair with — `variance_at_scaled(cmt, f, sigma, &[], mult) ≤ MIN_VARIANCE` — computing the raw (unfloored) slope times `m²`.

### 🟡 Test coverage — floored scaled path has no FD-parity test

New regression tests exercise only the `mult = 1` unscaled path. The floored custom-magnitude path (`dvar_df_scaled`/`d2var_df2_scaled`, `residual_rd2` `Some(m)` branch) is never asserted against FD — so the bug above passes CI silently, violating the CLAUDE.md rule that every analytic-sensitivity change carry a Dual2/analytic-vs-FD parity test *exercising the floored path*.

### 🟡 Perf — `variance_at` recomputed 2–3× per observation

At the covariance/gradient hot-path call sites (`m3_sigma_derivs`, `inner_eta_responses`, `subject_cov_hessian_m3_natural`, `foce_sb_fixed_natural`, `subject_cov_hessian_foce_natural`), `r = variance_at(...)` is already computed, then `dvar_df` and `d2var_df2` each recompute it internally via the new floor gate — 3× the `variance_at` evals per observation. Cheap per call (no alloc) but avoidable across every obs × subject × outer iteration; thread the computed `r` (or an `is_floored` flag) into the derivative accessors.

_Fixing the correctness defect + adding the parity test now._

Issue #970, opened 2026-08-15
Split ferx-core into a Cargo workspace: scoping for #969 lever 3


roninsightrx:
Scoping document for **lever 3 of #969** ("split the crate"), which #969 asked to be scoped separately if lever 1 (`[profile.ci-test]` knobs) does not deliver.

#969 measured that ~70 % of every PR job is compiling `ferx-core` itself, twice. This issue is the "if (1) does not deliver" branch, written up front so the decision is made on numbers rather than appetite. Measured against `main` @ 8a711c0f, with #968 still open.

Two findings below change #969's framing and are the parts worth reading first: **§0/§7** (the split's payoff is entirely warm-cache reuse, and the caching mechanism has a known failure mode that can be probed in a day *before* any refactor) and **§8** (#969's lever 2 was dismissed on a reading that misses the largest available lever).

---

## 0. What a split actually buys, and what it does not

This is the part that decides whether the refactor is worth several weeks.

**Measured today** (`src/`, 224 472 lines over 289 files):

| Region | Lines | Notes |
|---|---:|---|
| `estimation/` | 49 958 | biggest single module |
| `sens/` | 38 017 | 17 478 of it is `provider_tests.rs` + `ode_provider_tests.rs` |
| `parser/` | 30 920 | `model_parser.rs` 16 677 + its tests 13 622 |
| `api/` | 23 202 | incl. `api/tests/` 12 087 |
| `ode/` | 17 072 | |
| `pk/` | 15 933 | |
| `io/` | 12 137 | |
| `stats/` | 8 703 | |
| `types.rs` (+ tests) | 8 627 | |
| `survival/` | 4 132 | feature-gated |
| `sim/`, `frem/`, `suggest_start/`, `markov/`, `nn/`, misc | ~11 800 | |
| **of which `*_tests.rs` sibling files** | **63 999** | plus `api/tests/` 12 087 |

3 093 `#[test]` functions live in `src/`; 128 files in `tests/`.

**What the split buys:**

1. **Warm-cache PR builds.** Today *any* one-character change to `src/` invalidates the entire ~18 min build, because the compilation unit is the whole crate. After a split, a PR that only touches `estimation/` reuses cached artifacts for `ferx-ir`, `ferx-types`, `ferx-pk`, `ferx-parser` — roughly 60 % of the tree. **This is the whole value proposition, and it is contingent on §7 (artifact reuse actually working in CI). If §7 does not hold, do not do this refactor.**
[...]

Issue #969, opened 2026-08-15
CI wall clock is compile-bound on the ferx-core lib: measure the ci-test profile trade-off


roninsightrx:
## Problem

Per-PR CI wall clock is ~22 min, and it is **compile-bound on a single crate**. #968 took the topology wins that were available (four ~27m jobs to two, and the endpoints job from 122 test binaries down to 8), landing wall clock at ~22m. What remains is not something job topology can fix: the critical path is now one crate compiling twice.

Measured **warm-cache** on run [31891690228](https://github.com/FeRx-NLME/ferx-core/actions/runs/31891690228) (an earlier cold-cache run gave misleadingly high numbers; see the correction on #968):

| Job | compile | run | total |
|---|---|---|---|
| Tests + coverage (core) | 14m11s | 7m24s | 22m |
| Tests + coverage (TTE/CTMM endpoints) | 12m43s | 2m56s | 16m |

**Caveat on every number in this issue: n=1 per configuration, and the noise is large.** The core job's scope did not change in #968 — it gained a test binary, if anything — yet it moved 27m to 22m across two runs. That ~5m swing on identical work is run-to-run variance on GitHub's shared runners and is the noise floor here. Any experiment run against this issue needs repeated runs, not a single before/after.

**Most of each job is compiling `ferx-core` itself, twice** (~64% of the core job, ~79% of the endpoints job). Any `cargo test` invocation that names both a lib target and an integration target builds the crate in two flavours — once with `--cfg test` for the lib harness, once as a plain dependency for the integration binaries — and at `ci-test` (which `inherits = "release"`, so `opt-level = 3`, `codegen-units = 16`) each pass is roughly 9–11 min on a 2-core runner.

The integration-test binaries are a real but secondary cost. Warm-cache against warm-cache, the endpoints job dropping 114 of them saved ~5m26s of compile (18m09s to 12m43s) and ~6m41s of run (9m37s to 2m56s).

> An earlier revision of this issue claimed the binaries were "nearly free". That came from comparing the cold-cache endpoints job against the warm-cache markov baseline — an invalid comparison that hid ~5m of dependency-graph build in the endpoints figure. Corrected here and on #968.

[...]

PR #534, opened 2026-06-26
feat(estimation): analytic ExpressionScale obs_scale gradient (inner + ODE) [#486]

hiddevandebeek:
## Why

Status issue **#486** tracks which scenarios the analytic forward-sensitivity path covers. One remaining FD holdout was the **η-dependent `ExpressionScale` `obs_scale`** divisor (e.g. `obs_scale = 1000 / V` with `V` carrying IIV):

- On the **analytical** 1/2/3-cpt path the *outer* gradient already differentiated the scale program, but the per-subject **inner** EBE loop reverted to finite differences — the light first-order provider didn't carry the quotient rule.
- On the **ODE** path *both* loops fell back to FD (the only analytic route was a Form-C readout `y = state/V`).

This PR closes both gaps so the divisor scale is fully analytic, removing the FD steps (faster inner loop, and the reported gradient method reads "analytic" instead of FD).

> While scoping this against `origin/main` I noticed the #486 matrix footnote¹ ("Analytical IOV inner is FD") is **stale** — IOV inner is already analytic on `main` (`iov_analytical_supported` + `find_ebe_iov` → `analytic_eta_nll_gradient_iov` / `subject_eta_grad_iov`, shipped via #466/#439). Not touched here; flagging for a matrix refresh.

## What changed

The scale quotient rule `∂(f/s)/∂x = (∂f/∂x)·s⁻¹ − f·(∂s/∂x)·s⁻²` (`x ∈ {η, θ}`) is applied on the final `(θ, η)`-space prediction jet, evaluated once per subject over the differentiable scale program:

- **Analytical inner** — new `apply_expression_scale_inner` (η-block only) in the light provider, wired after the init impulse / before LTBS. `ExpressionScale` dropped from `analytic_inner_common_bail`; IOV (`iov_analytical_supported` requires `ScalingSpec::None`) and ODE keep declining it through their own gates.
- **ODE outer + inner** — the closed-form dispatch is extracted into provider-agnostic `apply_expression_scale_outer` / `apply_expression_scale_inner_dispatch` (single source of truth) and reused by the ODE provider on its `SubjectSens` / `Vec<ObsGrad>`. `ode_analytical_supported` now admits `ExpressionScale` (axis-bounded, `!log_transform`); `ode_tvcov_supported` excludes it so it's served on the **static** walk only.

[...]

PR #466, opened 2026-06-23
feat(estimation): exact analytic FOCE/FOCEI gradients for IOV models [#439]

hiddevandebeek:
## Summary

Extends the analytic sensitivity path to **inter-occasion variability (IOV)**, over the stacked random-effects vector `[η_bsv, κ₁..κ_K]` with the block-diagonal prior `Ω_bsv ⊕ K·Ω_iov`. Both the **ODE** (`[odes]`) and **closed-form** analytical IOV paths now have exact analytic gradients on **both** the inner EBE and outer (θ/Ω/σ) loops — previously these fell back to finite differences (the ODE path on both loops; the closed-form path on the inner loop).

| | Outer gradient | Inner EBE gradient |
|---|---|---|
| Closed-form IOV | analytic (pre-existing) | **analytic (new)** |
| ODE IOV | **analytic (new)** | **analytic (new)** |

This is part of the #439 tier plan (Tier 1, IOV/κ).

## What changed

**ODE IOV (`src/sens/ode_provider.rs`)** — previously finite-difference on both loops:
- **Outer gradient**: a `Dual2` event-driven walk seeded with per-occasion κ axes — the *same* TV-cov walk (`integrate_tvcov_readout`) fed per-occasion parameters. Production's `predict_iov` mirrors this by calling the same `ode_predictions_event_driven`, so the dual walk matches the production predictor to ~1e-9 (the parity already established for TV-cov). Covers 1-/2-/3-cpt IV and oral, bolus dosing, with and without time-varying covariates.
- **Inner gradient**: a light first-order `Dual1` stacked-η walk, plus an analytic BSV H-matrix.
- A shared per-subject gate (`ode_iov_subject_supported`) and a shared event-mapping helper (`seed_iov_events`) keep the inner (`Dual1`) and outer (`Dual2`) on matched scope.

**Closed-form IOV (`src/sens/provider.rs`)**:
- The inner EBE gradient is now an exact analytic `Dual1` stacked-η walk (`run_obs_iov_eta` / `subject_eta_grad_iov_analytical`), where it was finite-difference. The source-building (per-subject gate, occasion split, per-event `CombinedDerivs`) is factored into `build_iov_sources`, shared with the outer `Dual2` walk.

**Wiring**:
[...]

PR #974, opened 2026-08-16
perf(sens): bucket the ODE IOV dual-width dispatch ladder [closes #971]

### LINKED ISSUE #971: Bucket the Dual1<N> dispatch ladders: 96 const-generic widths are 38.6% of the crate's LLVM IR
## Problem

`cargo llvm-lines` on the lib (`--no-default-features --features ci --profile ci-test`) reports **17 423 997 LLVM lines across 340 426 function copies**, with a completely flat distribution — the single largest function is 0.1 %. The cost is not any one hot function, it is **monomorphisation over const-generic dual widths**.

* `sens::` symbols account for **90.1 %** of all LLVM lines in the crate.
* `Dual1<N>` is instantiated for **every integer width N = 1…96** — 96 distinct widths, **6 728 028 lines = 38.6 % of the entire crate's IR**.
* By first path segment: `sens` 59.1 % / 209 870 copies, `ode` 14.9 %, `parser` 7.9 %, `types` 5.0 %, `pk` 4.4 %, and — for contrast — `estimation` (the largest module by line count, 49 958 lines) just **1.9 %**.

This matters because per-PR CI wall clock is compile-bound (#969) and, per the `-Ztime-passes` breakdown posted there, **~93 % of the lib compile is LLVM** (`LLVM_thinlto` 312 s + `LLVM_passes` 76 s of a 404 s total) while the front end is ~1 %. Fewer, larger monomorphic copies is therefore the most direct attack on the actual cost centre — and unlike the profile knobs in #969 it has no runtime downside for the common case.

## Where the 96 widths come from

`src/sens/ode_provider.rs:89`:

```rust
/// Largest stacked `(θ, η_bsv, κ_1..κ_K)` axis count for ODE IOV subjects.
pub(crate) const MAX_ODE_IOV_AXES: usize = 96;
```

and the ladder that enumerates it:

```rust
macro_rules! dispatch_ode_iov_axes {
    ($dim:expr, $worker:ident, $($arg:expr),+ $(,)?) => {
        match $dim {
            1 => $worker::<1>($($arg),+),
            2 => $worker::<2>($($arg),+),
            ...
            96 => $worker::<96>($($arg),+),
```

Two call sites (`src/sens/ode_provider.rs:2593`, `:2620`) expand it over `run_subject_iov::<M>`, and each instantiation drags the whole ODE-integration + sensitivity stack through at `Dual1<M>`. High-occasion IOV subjects are wide only in the kappa stack, so the tail widths (say 40–96) exist to serve a handful of real models while each one costs a full copy of the machinery.

The same shape, at smaller scale, appears in the `disp!` / `dispatch_tv!` tables bounded by `MAX_ODE_AXES = 24` (four tables) and `dispatch_init_impulse!` bounded by `MAX_SCALE_AXES = 24`.

## Proposal

**Bucket the dispatch ladders**: round the runtime axis count up to the next bucket and pass the logical width as a runtime argument, leaving the extra lanes zero.

```
exact:  1..=12          (common case, keeps today's codegen exactly)
bucket: 16, 24, 32, 48, 64, 96
```

96 instantiations → ~18. If the reduction in IR is anywhere near proportional for that ladder, it removes a large fraction of the 38.6 %.

Constraints to respect:

* **Padding costs runtime**, proportional to the gap between the true width and the bucket — a 33-axis subject would run 48 lanes. That is why the small widths stay exact: they are the common case, and the tail is where instantiation cost is paid for models that are already expensive. The bucket ladder is a tuning parameter; measure before settling it.
* The dual arithmetic must ignore padded lanes semantically (they stay 0, so they already do), but any code that reads `N` as "the number of live axes" must switch to the runtime width.
* The existing compile-time tripwires (`const _: () = assert!(MAX_ODE_IOV_AXES == 96, …)`, and the `MAX_ODE_AXES <= MAX_SCALE_AXES` coupling from the #534 audit) exist to stop a widened cap from silently reaching a `_ => None` FD fallback or a silent `_ => {}` no-op. Bucketing must keep an equivalent guard: a subject wider than the largest bucket must still route to FD **loudly**, never silently.

## Validation (per `CLAUDE.md`)

This touches an analytic sensitivity path, so the `Dual2`-vs-FD parity rule applies in full:

* Existing `*_g_dual_matches_fd` per-kernel checks and the `check_full_provider_vs_fd` harness must stay green — they are the guard that a bucketed width still produces the exact same `∂f/∂η` / `∂f/∂θ`.
* Add a unit test pinning bucket selection (width 33 → bucket 48, width 96 → 96, width 97 → FD fallback with the fallback actually observable, not silent).
* Add a Tier-1 test asserting the sensitivities from a bucketed instantiation are bit-identical to the exact-width instantiation for a width that is not a bucket boundary.

## Acceptance

* `cargo llvm-lines --lib --no-default-features --features ci --profile ci-test` before/after, reporting total lines, total copies, and the `Dual1<N>` share.
* `RUSTFLAGS="-Ztime-passes"` lib build before/after (the `total` and `LLVM_*` lines).
* Runtime check on the IOV path: `iov_convergence` (Tier 3) before/after, best-of-3, plus a wide-kappa subject if one exists in the fixtures — the padding cost must be shown, not assumed.
* Numbers posted here and cross-linked from #969.

## Context

Surfaced by the Phase 0 measurement gate for #969's lever 3 (crate split); full data on #970, which this supersedes as the useful outcome of that investigation. Baseline measurements were taken on a 15-core arm64 macOS box with `RUSTC_WRAPPER=` and `CARGO_INCREMENTAL=0` — shares transfer to the 2-core CI runner, absolute times do not.

roninsightrx:
## Why

Per-PR CI wall clock is compile-bound (#969), and per `-Ztime-passes` ~93 % of the lib compile is LLVM. `cargo llvm-lines` (#970/#971) traced that cost to monomorphisation over const-generic dual widths rather than to any hot function: the ODE IOV dispatch ladder instantiated `run_subject_iov` / `run_subject_iov_eta` — and with them the whole ODE-integration + sensitivity stack — once for **every** stacked axis count `1..=MAX_ODE_IOV_AXES = 96`. Widths `13..=96` alone were **62 % of the crate's LLVM IR**.

Closes #971.

## What changed

Round the runtime axis count up to the next monomorphised width and leave the extra lanes zero.

Padding is semantically inert on this path, which is why it is safe rather than merely plausible:

* every seeder already guards its axis writes (`seed_pk_dual2_iov` / `seed_pk_dual1_iov` skip `ax >= M`, `eval_scale_dual*` seeds only `dim < M`);
* every readout indexes by the **runtime** `n_theta` / `n_stacked`, never by `N`;
* the dual arithmetic is elementwise over lanes — no cross-lane reduction — so a zero lane stays zero and cannot perturb a live one;
* the adaptive ODE step control uses the **value-only** RMS error norm (`Stepper::attempt`), so padded lanes cannot change step sizes.

New `src/sens/widths.rs` holds `bucket_for` / `buckets_well_formed` / `slices_eq` (all `const fn`), and `ODE_IOV_WIDTH_BUCKETS` in `ode_provider.rs` is the ladder:

```
exact:  1..=24
bucket: 28, 32, 40, 48, 56, 64, 80, 96      (96 instantiations -> 32)
```

The split follows the measured per-width cost, not a guess:

* **`1..=24` exact.** Those widths are already instantiated by the `MAX_ODE_AXES` / `MAX_SCALE_AXES = 24` ladders, so the IOV ladder shares their monomorphisations and each width costs only **~16 k** extra LLVM lines. Exactness is nearly free here, and this range covers the ordinary IOV model — so the common case pays **zero** runtime padding.
* **`> 24` bucketed.** Past that cap the IOV ladder is the sole user, at **~115 k lines per width** (~1.2 % of the post-change crate each). The step is chosen to hold the `O(M²)` outer padding penalty at ~1.5× rather than the ~2.1× a coarser `16/24/32/48/64/96` ladder measured (numbers below).

### Tripwires

The old `const _: () = assert!(MAX_ODE_IOV_AXES == 96, …)` literal check is replaced by a strictly stronger compile-time pair, keeping the #438 / #466 / #534 convention that a scope gap must fail **loudly**:

* `buckets_well_formed(&ODE_IOV_WIDTH_BUCKETS, MAX_ODE_IOV_AXES)` — the ladder must be strictly ascending and end exactly at the cap `ode_iov_subject_supported` enforces, so no in-cap width can fall through to the dispatch `_ => None` and silently reach FD;
* `slices_eq(&[…arms…], &ODE_IOV_WIDTH_BUCKETS)` inside the dispatch macro — the macro's literal arm list and the ladder const cannot desync.

A subject wider than the last bucket is still declined by the per-subject gate before any dispatch, exactly as before.

## Alternatives considered

* **The coarse ladder the issue proposed** (`1..=12` exact, then `16, 24, 32, 48, 64, 96`, 18 widths). Measured: **−47.1 %** IR (vs −43.5 % here) but the outer walk cost up to **3.0×** at the boundary just past 16 — i.e. right where real IOV models live. Trading 3.6 points of compile for that was the wrong side of the trade; hence the exact-through-24 split.
* **Crate split** (#969 lever 3) — measured and rejected as a NO-GO in #970; this supersedes it as the useful outcome of that investigation.

## Cross-repo dependency

No `pub` API changed (`widths` is `pub(crate)`, `ODE_IOV_WIDTH_BUCKETS` is `pub(crate)`).

## Breaking changes
- [x] None

## Numerical validation

Gradients are **bit-identical**, not merely close — this is the `Dual2`-vs-FD path, so that is the standard the CLAUDE.md rule sets.

- [x] Compared against prior ferx commit `c56ade0c`
- [x] Existing `*_g_dual_matches_fd` per-kernel checks and the `check_full_provider_vs_fd` harness stay green (full `cargo test --lib`: **2822 passed, 0 failed**)
- [x] New `ode_iov_bucketed_dispatch_is_bit_identical_to_the_exact_width` compares a subject whose widths are *not* bucket boundaries (`m_dim = 27 → 28`, `n_stacked = 25 → 28`) against the exact-width instantiation the pre-#971 ladder ran, on `f`, `∂f/∂η`, `∂f/∂θ`, `∂²f/∂η²`, `∂²f/∂η∂θ` — via `to_bits()`, so even `+0.0` vs `−0.0` would fail
- [x] `iov_convergence` (Tier 3) unchanged: 11 passed, 11.71 s → 10.51 s (best of 3)

## Performance

Measured on a 15-core arm64 macOS box, `RUSTC_WRAPPER=`, `CARGO_INCREMENTAL=0`, `--lib --no-default-features --features ci --profile ci-test`.

**`cargo llvm-lines`**

| | before | after |
|---|---|---|
| total lines | 17,421,539 | **9,846,640 (−43.5 %)** |
| total copies | 340,378 | **189,179 (−44.4 %)** |
| distinct const-generic widths | 99 | 41 |
| `sens::` share | 10,303,251 (59.1 %) | 5,583,666 (56.7 %) |

**`RUSTFLAGS="-Ztime-passes"` (lib only, deps warm)**

| pass | before | after |
|---|---|---|
| `total` | 376.7 s | **127.5 s (−66.1 %, 2.95×)** |
| `LLVM_thinlto` | 294.4 s | **85.0 s (−71.1 %)** |
| `LLVM_passes` | 66.0 s | **31.4 s (−52.4 %)** |

**Runtime** — the padding cost is shown, not assumed:

- [x] Faster (compile) / mixed (runtime), justified below.

| check | before | after |
|---|---|---|
| `iov_convergence` (closed-form IOV, Tier 3) | 11.71 s | 10.51 s |
| pna-scale ODE IOV gradient, `m_dim = 90 → 96` | 0.94 s | 1.10 s (**+17 %**, vs the predicted `(96/90)² = 1.14×`) |

Direct sweep of `ode_subject_sensitivities_iov` (1-cpt IV ODE, κ on CL; `m_dim = 2θ + 1η + k·κ`; best of 5, `ci-test` profile) — the worst measured step-up across a bucket boundary is **~1.5–1.6×**, and everything at `m_dim ≤ 24` pays nothing:

```
 k  m_dim bucket  outer_ms  inner_ms       k  m_dim bucket  outer_ms  inner_ms
 9    12    12      0.23      0.03        37    40    40     12.13      0.29
10    13    13      0.30      0.03        38    41    48     24.57      0.29
13    16    16      0.54      0.05        45    48    48     30.67      0.34
14    17    17      0.71      0.05        46    49    56     45.62      0.34
21    24    24      2.44      0.10        53    56    56     50.88      0.43
22    25    28      3.74      0.10        54    57    64     69.55      0.45
25    28    28      4.29      0.16        61    64    64     78.02      0.56
26    29    32      5.96      0.15        62    65    80    115.26      0.54
29    32    32      6.47      0.20        77    80    80    149.96      0.79
30    33    40     10.32      0.21        78    81    96    232.18      0.79
33    36    40     11.21      0.26        93    96    96    280.35      1.47
34    37    40     11.51      0.27
```

For contrast, the coarse `16/24/32/48/64/96` ladder on the same fixture: `m_dim 16 → 17` cost **0.55 → 1.64 ms (3.0×)** and `32 → 33` cost **6.13 → 16.48 ms (2.7×)`.

## Tests
- [x] Unit tests added in the same module (`cargo test --lib` passes: 2822 passed, 0 failed)
- [x] `src/sens/widths.rs`: `bucket_for` round-up / decline / idempotence, `buckets_well_formed` accept + reject, `slices_eq`
- [x] `ode_iov_width_buckets_round_up_and_decline_past_the_cap` — pins the real ladder (`33 → 40`, `96 → 96`, `97 → None`, `0 → None`), asserts every in-cap width has a bucket, and asserts **no** round-up exceeds ~1.55× the exact `M×M` work
- [x] `ode_iov_bucketed_dispatch_is_bit_identical_to_the_exact_width` — the bit-identity teeth described above, outer `Dual2` and inner `Dual1`, plus the public entry points
- [x] `ode_iov_subject_past_the_last_bucket_declines_to_fd` — a 97-axis subject is declined **at the gate** (observably `None` from `ode_iov_subject_supported` and both entry points), never silently through the ladder

- [x] Not applicable (internal — no new API surface, no DSL change)

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added (under `Performance` — the runtime characteristics of wide-IOV ODE fits change)
- [x] No other user-visible change (no new fit option, DSL syntax, or output field)

## Checklist
- [x] `cargo clippy` clean (no new warnings; `cargo fmt --check` clean)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the walk itself is untouched; only the width selected for it changed, and bit-identity is asserted
- [x] No version bump needed (non-breaking)

## Reviewer hints

Focus on two things:

1. **Is padding really inert?** The four reasons are listed under *What changed*; the load-bearing ones are the `ax < M` seeder guards and the value-only ODE error norm. The bit-identity test is the empirical check, but the argument should stand on its own.
2. **The tripwire swap.** The old assert pinned a literal `96`; the new pair pins *the ladder's last entry to the cap* and *the macro arms to the ladder*. Convince yourself no edit to either can produce an in-cap width that reaches `_ => None`.

Skimmable: the mechanical `dispatch_ode_iov_axes!` rewrite (the 96-arm `match` became a generated one) and the `call_at_width!` helper, which exists only because `macro_rules!` cannot nest two repetitions of different lengths.

## Open questions

The remaining `13..=24` mass (~189 k lines per width, ~23 % of the post-change crate) belongs to the `MAX_ODE_AXES` / `MAX_SCALE_AXES = 24` `disp!` / `dispatch_tv!` / `dispatch_init_impulse!` ladders, which this PR deliberately leaves alone: unlike the IOV workers they carry live-width assumptions (`debug_assert_eq!(N, n_eta)` in `seed_pk_dual1`, `debug_assert_eq!(M, n_theta + n_eta)` in `seed_pk_dual2`) that bucketing would have to relax first. Worth a follow-up issue, or fold into #969?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,18 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Performance
+- **The ODE inter-occasion-variability (IOV) analytic sensitivity path now compiles from a
+  bucketed set of dual widths instead of one specialisation per stacked axis count**, cutting
+  the crate's generated LLVM IR by 43 % (17.4 M → 9.8 M lines) and the lib's `-Ztime-passes`
+  total by ~3.2× locally — the direct attack on the compile-bound CI wall clock tracked in
+  #969. Gradients are unchanged: a stacked width that is not a bucket boundary is padded with
+  zero lanes and returns bit-identical `∂f/∂η` / `∂f/∂θ` to the exact-width walk. Widths up to
+  24 axes — the ordinary IOV model — are still specialised exactly, so they pay no runtime
+  cost at all; wider subjects run up to ~1.5× more work in the outer walk for the padded
+  lanes, and the 96-axis cap (past which a subject falls back to finite differences) is
+  unchanged (#971).
+
 ### Fixed
 - **The default (`Auto` → analytic-gradient NLopt L-BFGS) optimizer no longer stalls at the
   initial estimates on its first step.** From the identity initial Hessian the opening search
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ section of the SDLC for the versioning policy).
 - **The ODE inter-occasion-variability (IOV) analytic sensitivity path now compiles from a
   bucketed set of dual widths instead of one specialisation per stacked axis count**, cutting
   the crate's generated LLVM IR by 43 % (17.4 M → 9.8 M lines) and the lib's `-Ztime-passes`
-  total by ~3.2× locally — the direct attack on the compile-bound CI wall clock tracked in
+  total by ~2.95× locally (376.7 s → 127.5 s) — the direct attack on the compile-bound CI wall clock tracked in
   #969. Gradients are unchanged: a stacked width that is not a bucket boundary is padded with
   zero lanes and returns bit-identical `∂f/∂η` / `∂f/∂θ` to the exact-width walk. Widths up to
   24 axes — the ordinary IOV model — are still specialised exactly, so they pay no runtime
--- a/src/sens/mod.rs
+++ b/src/sens/mod.rs
@@ -30,3 +30,4 @@ pub mod three_cpt;
 pub mod three_cpt_explicit;
 pub mod two_cpt;
 pub mod two_cpt_explicit;
+pub(crate) mod widths;
--- a/src/sens/ode_provider.rs
+++ b/src/sens/ode_provider.rs
@@ -148,7 +148,10 @@ const _: () = assert!(
 ///   `32/48/64/96` ladder costs.
 ///
 /// The ladder is a tuning parameter: widen it where padding cost bites, narrow it where
-/// compile time does.
+/// compile time does. When you retune it, edit the literal arm list in
+/// [`dispatch_ode_iov_axes`] to match — `macro_rules!` can't iterate a const, so the two are
+/// kept in lockstep by the `slices_eq` compile-time assert in `dispatch_iov_widths!` rather
+/// than shared from one source.
 pub(crate) const ODE_IOV_WIDTH_BUCKETS: [usize; 32] = [
     1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 28, 32,
     40, 48, 56, 64, 80, 96,
--- a/src/sens/ode_provider.rs
+++ b/src/sens/ode_provider.rs
@@ -123,112 +123,93 @@ const _: () = assert!(
      MAX_SCALE_AXES (and its dispatch_init_impulse! table) to at least MAX_ODE_AXES."
 );
 
+/// Monomorphised `(Dual1, Dual2)` widths for the ODE IOV dispatch ladder (#971).
+///
+/// Enumerating every width `1..=MAX_ODE_IOV_AXES` instantiated the entire
+/// ODE-integration and sensitivity stack 96 times per worker — 62 % of the crate's LLVM
+/// IR sat on widths `13..=96` alone, and ~93 % of the lib compile is LLVM (#969/#970). So the ladder is
+/// **bucketed**: the runtime axis count is rounded up to the next width here and the extra
+/// lanes are left zero (see [`crate::sens::widths`] for why padding is semantically inert —
+/// every seeder guards its axis writes with `ax < N` and every readout indexes by the
+/// runtime `n_theta` / `n_stacked`, never by `N`).
+///
+/// Padding is not free at *runtime*: the outer walk is `Dual2<M>` with an `M×M` Hessian per
+/// value, so rounding `M` up costs about `(bucket/exact)²` (measured, #971 — the inner
+/// `Dual1<N>` walk is `O(N)` and two orders of magnitude cheaper either way). The ladder is
+/// therefore split where the two costs actually sit:
+///
+/// * **`1..=24` exact.** These widths are already instantiated by the `MAX_ODE_AXES` /
+///   `MAX_SCALE_AXES = 24` ladders, so the IOV ladder shares their monomorphisations and
+///   each extra width here costs only ~16 k LLVM lines (~0.2 % of the crate). Exactness is
+///   nearly free, and this range covers the ordinary IOV model.
+/// * **`> 24` bucketed.** Past the shared cap the IOV ladder is the sole user, at ~115 k
+///   lines per width (~1.2 % of the crate each), so the tail rounds — with a step chosen to
+///   hold the `O(M²)` padding penalty at ≤ ~1.5× rather than the ~2.1× a coarser
+///   `32/48/64/96` ladder costs.
+///
+/// The ladder is a tuning parameter: widen it where padding cost bites, narrow it where
+/// compile time does.
+pub(crate) const ODE_IOV_WIDTH_BUCKETS: [usize; 32] = [
+    1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 28, 32,
+    40, 48, 56, 64, 80, 96,
+];
+
+// Tripwire, replacing the old `MAX_ODE_IOV_AXES == 96` literal check: the ladder must be
+// strictly ascending and reach exactly the cap `ode_iov_subject_supported` enforces.
+// Otherwise a subject inside the gate but past the last bucket would hit the ladder's
+// `_ => None` arm and fall **silently** back to FD instead of being declined loudly by the
+// gate (the #438 / #466 / #534 convention, carried over to the bucketed form).
 const _: () = assert!(
-    MAX_ODE_IOV_AXES == 96,
-    "MAX_ODE_IOV_AXES changed: update dispatch_ode_iov_axes! to enumerate the same range"
+    crate::sens::widths::buckets_well_formed(&ODE_IOV_WIDTH_BUCKETS, MAX_ODE_IOV_AXES),
+    "ODE_IOV_WIDTH_BUCKETS must be strictly ascending and end at MAX_ODE_IOV_AXES: the \
+     dispatch ladder has to cover the whole range ode_iov_subject_supported admits, or an \
+     in-scope subject silently falls back to FD"
 );
 
-macro_rules! dispatch_ode_iov_axes {
-    ($dim:expr, $worker:ident, $($arg:expr),+ $(,)?) => {
-        match $dim {
-            1 => $worker::<1>($($arg),+),
-            2 => $worker::<2>($($arg),+),
-            3 => $worker::<3>($($arg),+),
-            4 => $worker::<4>($($arg),+),
-            5 => $worker::<5>($($arg),+),
-            6 => $worker::<6>($($arg),+),
-            7 => $worker::<7>($($arg),+),
-            8 => $worker::<8>($($arg),+),
-            9 => $worker::<9>($($arg),+),
-            10 => $worker::<10>($($arg),+),
-            11 => $worker::<11>($($arg),+),
-            12 => $worker::<12>($($arg),+),
-            13 => $worker::<13>($($arg),+),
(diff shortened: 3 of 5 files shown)


roninsightrx:
## Code review (high effort) — mechanism sound, no correctness bug found

Reviewed the full diff plus the enclosing IOV workers, seeders, readouts, dispatch macro, and tripwires across 8 angles.

**The padding-is-inert claim holds under inspection**, not just the bit-identity test:
- Every IOV seeder guards its writes — `seed_pk_dual2_iov` uses `n_theta.min(M)`, `if ax >= M { continue }`, and `bx < M`; `seed_pk_dual1_iov` the `Dual1` analogue.
- Every readout indexes by the **runtime** `n_theta` / `n_stacked` (`run_subject_iov`, `apply_scale_quotient_row`, `eval_scale_dual`), never by `N`.
- The IOV workers carry **no** `debug_assert_eq!(M, …)` — those live only in the non-IOV `seed_pk_dual2` / `seed_pk_dual1`, which this PR correctly leaves at exact width.
- `buckets_well_formed` + `slices_eq` are strictly stronger than the deleted `MAX_ODE_IOV_AXES == 96` literal: they pin the ladder's last entry to the cap and the macro arms to the const, so no in-cap width can reach `_ => None`. The cap stays 96 and no other consumer depended on the old literal assert.

The `m_dim 27 → 28` bit-identity test (ODE-integrated, `to_bits`) empirically nails the load-bearing value-only-step-norm assumption. LGTM on correctness.

### Minor (docs/cleanup) — fixed in 989294a… (pushed to this branch)

1. **`ode_provider_tests.rs`** — the contract doc on `ode_iov_width_buckets_round_up_and_decline_past_the_cap` said *"exact through 12"*, but the ladder (and the test body's `for w in 1..=24`) is exact through **24**. Leftover from the coarse-ladder alternative. → fixed.
2. **`CHANGELOG.md`** — entry claimed *"~3.2× locally"*; the PR's own `-Ztime-passes` table is 376.7 s → 127.5 s = **2.95×**. → corrected to match the measured number.
3. **`ODE_IOV_WIDTH_BUCKETS`** — the 32-width list is written twice (const + inline in `dispatch_ode_iov_axes!`), kept in lockstep only by the `slices_eq` assert. Not a bug (build fails on desync), but added a pointer comment on the const so a future retune knows to edit both.

No blocking issues.

🤖 via /code-review (Claude Opus 4.8)

Issue #264, opened 2026-06-11
Test: end-to-end SIR fallback (FailedNonPd → run_sir_core → SirFallback)


TeunP:
### Follow-up from #245

PR #245 added the `covariance_fallback = sir` path: when the FD Hessian is non-PD at convergence, `compute_covariance` returns `CovarianceStepResult::FailedNonPd { reason, fallback_proposal }`, and `api.rs` then runs `run_sir_core` with that proposal and reports `covariance_status = sir_fallback`.

Every branch in that chain is unit-tested in isolation:
- `build_non_pd_fallback_proposal` (proposal construction, PD-ness, condition-number cap) — `outer_optimizer.rs`
- `resolve_covariance_status` (status precedence) — `api.rs`
- `select_fd_step` (adaptive step) — `outer_optimizer.rs`
- Tier-2 wiring test asserts the fallback stays **inert** on a well-identified model — `tests/covariance_fallback.rs`

**Gap:** there is no end-to-end test that actually drives `compute_covariance → FailedNonPd → run_sir_core → SirFallback` and asserts the resulting status + SIR credible intervals. The Tier-2 test documents why: a real PK fit can't be made deterministically non-PD without a contrived concave model.

### Proposed work
Add a synthetic concave-model (or otherwise reliably non-PD-at-convergence) fixture and a Tier-2 test that:
1. Fits with `covariance_fallback = sir`.
2. Asserts `covariance_status == SirFallback`.
3. Asserts `sir_ci_theta` / `sir_ci_omega` / `sir_ci_sigma` are populated and finite.
4. Asserts the non-PD warning (`reason`) is present in `warnings`.

This closes the one integration gap noted in the #245 review (https://github.com/FeRx-NLME/ferx-core/pull/245#issuecomment-4680266142). Non-blocking; the isolated branches are all covered.

TeunP:
## Re-scoped after digging into the code — the literal end-to-end isn't achievable

Implemented in #283. Investigating the real paths showed the original plan (drive the whole chain through `fit()`) is infeasible, so I closed the gap at the seam instead.

**Why a real non-PD `fit()` can't be produced:**
1. **The optimizer won't stop on a concave point.** BOBYQA's budget is `outer_maxiter·(n+1) + 40·(n+1)` (`outer_optimizer.rs`), so even `outer_maxiter = 0` runs a fixed `40·(n+1)` warmup and walks to the PD minimum.
[...]

PR #975, opened 2026-08-17
fix(api): arm the non-PD SIR fallback under `sir = true` [closes #972]

### LINKED ISSUE #972: SIR requested via `sir = true` is silently skipped when the covariance step fails
## Summary

When a fit requests SIR via `sir = true` but the covariance step **fails** (non-PD FD Hessian → no `covariance_matrix`), SIR is silently skipped with only a warning. The user's explicit request for SIR does **not** arm the non-PD SIR fallback, even though that fallback exists precisely for the failed-covariance case. To actually get SIR when covariance fails, the user must instead set a *different, separate* option — `covariance_fallback = sir`. This is surprising: `sir = true` and `covariance_fallback = sir` are wired to independent code paths.

## Current behavior

Two independent SIR triggers:

1. **`sir = true`** — standard path. Hard-requires `result.covariance_matrix`. On covariance failure it skips and pushes:
   > "SIR requested but covariance matrix is not available. Enable covariance = true in [fit_options]."
   (`src/api/fit.rs:1664-1691`)

2. **`covariance_fallback = sir`** — the *only* trigger for SIR-on-covariance-failure. The gate `should_run_sir_fallback` (`src/api/postfit.rs:380`) checks `options.covariance_fallback == CovarianceFallback::Sir` and **ignores `options.sir` entirely**. It seeds `run_sir_core` from the rectified-|eigenvalue| Hessian proposal (`build_non_pd_fallback_proposal`, `src/estimation/covariance.rs:168`) rather than an inverted covariance.

`covariance_fallback` defaults to `None` (`src/types.rs:5467`).

### Consequence

A user who sets only `sir = true`, hits a non-PD Hessian, gets **no SIR** and a warning that points at `covariance = true` — not at the mechanism (`covariance_fallback = sir`) that would actually give them SIR intervals from the rectified Hessian. The capability exists but is unreachable from the option the user naturally reached for.

## Proposed fix

Pick one (in rough order of preference):

- **A.** When `sir = true` and the covariance step failed, automatically arm the non-PD SIR fallback (as if `covariance_fallback = sir`), provided `compute_covariance` produced a usable fallback proposal. Most closely matches user intent ("I asked for SIR; run it off whatever proposal you can build").
- **B.** Keep the paths separate but make the skip warning actionable: point the user at `covariance_fallback = sir` instead of `covariance = true`.
- **C.** Both — arm the fallback under `sir = true`, and improve the warning for the case where no proposal could be built at all.

## Notes / scope

- The fallback still needs a usable Hessian eigendecomposition to build the proposal; if the Hessian itself is degenerate (NaN / all-zero block), no proposal is produced and SIR genuinely cannot run — the warning should say so.
- Standalone `run_sir` (`src/estimation/run_sir.rs:192`) also hard-requires `fit.covariance_matrix` and cannot seed off the fallback proposal. Consider whether it should accept the fallback proposal too, or at least give a clearer error. (Possibly a follow-up.)
- Docs to update: `docs/model-file/fit-options.qmd` (SIR + `covariance_fallback` interaction).
- Add a Tier-1 unit test on the `sir = true` + failed-cov gating (the `should_run_sir_fallback` / `resolve_sir_fallback` helpers are already unit-testable without driving a real non-PD fit — see #264).

roninsightrx:
## Why

`sir = true` and `covariance_fallback = sir` were wired to two independent code paths (#972).

When the FD Hessian came back non-positive-definite, a user who had asked for SIR got **no intervals at all** — just a warning pointing at `covariance = true` (an option they usually already had on). Meanwhile `compute_covariance` had *already built* the `|eigenvalue|`-rectified fallback proposal for exactly that case; reaching it required knowing about a second, separately-named option whose docs describe it as a covariance-step setting rather than a SIR one.

So the capability existed but was unreachable from the option the user naturally reached for, and the warning pointed away from the mechanism that would have helped.

This implements **Option C** from the issue: arm the fallback under `sir = true`, *and* fix the warning for the case where no proposal can be built at all.

## What changed

**`src/api/postfit.rs`**

- `should_run_sir_fallback` gains a `sir_requested` parameter; the opt-in condition becomes `covariance_fallback == sir || sir`. The other three gate conditions are untouched: no real covariance, no normal SIR result already, and a proposal actually present. So `sir = true` + non-PD Hessian now draws from the rectified-|λ| proposal and the fit reports `covariance_status: sir_fallback`.
- New `sir_unavailable_warning(sir_requested, covariance_requested, has_covariance_matrix, has_fallback_proposal, sir_ran)` — a pure function returning the warning only when SIR was requested and genuinely could not run, distinguishing the two real causes:
  - covariance step never run → *enable `covariance = true`* (the old message, now only where it applies);
  - covariance step ran but the FD Hessian's eigendecomposition returned no finite eigenvalues (NaN/Inf) → nothing to draw from, and no option would change that.
  
  It returns `None` when a proposal *was* built, because in that case the fallback ran and `resolve_sir_fallback` already pushed its own `"SIR fallback failed: …"` — no two messages stacked on one failure.

**`src/api/fit.rs`**

- The `sir = true` + no-covariance branch no longer warns inline; that was firing *before* the fallback had its chance. The warning is emitted after both SIR paths resolve, and is skipped when the fit was cancelled.

`covariance_fallback = sir` keeps a distinct meaning: the fallback *without* asking for SIR on a successful covariance step.

### Subtlety worth a reviewer's eye

The new message names both the covariance step and the Hessian, and `classify_warning` (`src/types.rs`) tests `"covariance step failed"` and `"degenerate"` **before** `"sir requested"` in its else-if chain. Either token would have misrouted a SIR warning to `covariance_failed` (Critical) / `optimizer_health`. The message avoids both tokens, the constraint is commented at the message site, and the routing is pinned by adding the message to the classifier's `classify_warning_roundtrips_every_engine_message` table.

## Alternatives considered

Option B alone (keep the paths separate, just make the warning point at `covariance_fallback = sir`) leaves the user a manual step to reach a capability they already asked for. Option A alone leaves the genuinely-unreachable case (degenerate Hessian, no proposal) still reporting the misleading `covariance = true` advice. C is both, and is what the issue lists as preferred.

Not done here, flagged in the issue as a possible follow-up: standalone `run_sir` (`src/estimation/run_sir.rs`) still hard-requires `fit.covariance_matrix`, and `run_covariance()` discards the fallback proposal (`src/estimation/run_covariance.rs:274`), so a post-hoc SIR fallback is unreachable from that entry point.

## Cross-repo dependency

No public Rust API change; `CovarianceStatus` values and `FitResult` fields are unchanged.

## Breaking changes
- [x] None

Behaviour change (not breaking, but worth naming): a fit with `sir = true`, `covariance = true` and a non-PD Hessian previously reported `covariance_status: failed` with no SIR fields; it now reports `covariance_status: sir_fallback` with SIR CIs. A user who set `covariance_fallback = none` *and* `sir = true` now gets the fallback; that is the intent of the issue (an explicit SIR request outranks the default-valued fallback key).

## Numerical validation
- [x] Not applicable — no numerical path is touched. The fallback proposal (`build_non_pd_fallback_proposal`) and `run_sir_core` are unchanged; this PR only changes *whether* they are reached and *what the warning says*. Every existing fit with a successful covariance step is bit-identical.

## Performance

One case does more work than before by design: `sir = true` with a non-PD Hessian now runs a SIR pass instead of skipping. That is the requested behaviour.

## Tests

Tier 1, in `src/api/tests/tests_sir_fallback.rs` and `src/types_tests.rs`:

- `sir_requested_alone_arms_the_non_pd_fallback` — the regression test: `should_run_sir_fallback(fallback_is_sir = false, sir_requested = true, …)` fires, and each of the other three conditions still gates it.
- `sir_fallback_gate_fires_only_when_all_conditions_hold` / `…_blocked_by_each_condition` — updated for the new arity, still one-deviation-per-assert.
- `sir_unavailable_warning_is_silent_when_sir_is_not_stranded` — the four silent cases, including "fallback fired and failed" (no stacked warning).
- `sir_unavailable_warning_points_at_covariance_when_the_step_never_ran`.
- `sir_unavailable_warning_reports_a_degenerate_hessian_when_the_step_ran` — asserts the message does **not** say `covariance = true`.
- The new message added to `classify_warning_roundtrips_every_engine_message`, pinning it to `sir` / Warning rather than `covariance_failed` / Critical.

These follow #264's precedent: the gate and the warning are pure functions, unit-testable without driving a real fit to a non-PD Hessian (which the optimizer's fixed warmup budget cannot reliably reach, and which a degenerate fixture cannot reliably survive inside SIR).

Full run: `cargo test --lib` — 2814+ pass, 0 fail. `cargo check --tests` clean.

The user-facing shape is just:

```toml
[fit_options]
covariance = true
sir = true
```

which, on a non-PD Hessian, now yields SIR CIs and `covariance_status: sir_fallback` instead of a skip.

## Docs

`docs/model-file/fit-options.qmd`: new "When the covariance step fails" subsection under SIR explaining the automatic fallback and the two remaining skip cases, plus a note on the `covariance_fallback` row that `sir = true` arms the same fallback on its own. No new page, so `_quarto.yml` is unchanged.

## Checklist
- [x] `cargo clippy` clean (no new warnings; pre-existing ones untouched)

### ferx-site
- [x] No new DSL syntax, no new `[fit_options]` key (both keys already documented), no new example or data file.

### ferx-book
- [x] No new estimator / DSL feature / fit option.

### Example execution
- [x] No example execution step needed — no user-visible change to any example's output (no example uses `sir = true` with a failing covariance step).

## Reviewer hints

Focus on two things:

1. **The gate's new disjunction** (`src/api/postfit.rs`, `should_run_sir_fallback`) — specifically whether `sir = true` overriding an explicit `covariance_fallback = none` is the intent. The issue's Option A says yes; it is called out under Breaking changes above.
2. **The warning's wording constraint** — the `classify_warning` ordering trap described under "What changed". If you reword the message, keep `"covariance step failed"`, `"covariance failed"` and `"degenerate"` out of it, or the classifier test will catch you.

The rest (test arity updates, docs, changelog) can be skimmed.

## Open questions

Should `run_covariance()` also retain the fallback proposal so a post-hoc covariance run can reach SIR the same way? That is the `src/estimation/run_covariance.rs:274` `sir_fallback_proposal: _` discard, and it is the remaining place where the two paths are still asymmetric. Left out of this PR as the issue's flagged follow-up — happy to fold it in here if you'd rather it land together.


roninsightrx:

Reviewed the diff vs `main`. The gate change (`should_run_sir_fallback`), the warning routing, and the test arity all check out. The `classify_warning` ordering trap is genuinely avoided: the new message contains neither `"covariance step failed"`/`"covariance failed"` nor `"degenerate"`, and `"covariance step"` is only fatal when paired with `"not positive definite"`.

Three findings, all addressed in the follow-up commit on this branch.

### 1. `src/api/postfit.rs` — medium: the "no proposal" message asserts a cause it cannot know

The message hard-asserts a single cause (*"its eigendecomposition did not produce finite eigenvalues"*), but `compute_covariance` returns `Unusable` — and therefore no `sir_fallback_proposal` — from at least four other paths:

- invalid `fd_hessian_step` (`covariance.rs:564`)
- non-finite base OFV (`:702`)
- ill-conditioned / zero-diagonal FD stencil — `"Hessian has ill-conditioned entries … zero diagonal — flat objective"` (`:944`)
- singular score cross-product under `covariance_method = s` / `rsr` (`:1025`)

Concrete case: a fit with one weakly-identified theta whose FD diagonal is ~0 fails at `:944`, and the user is told the eigendecomposition returned NaN/Inf. Wrong cause, and the actionable hint (raise `fd_hessian_step`, drop the flat parameter) is lost. The `s`/`rsr` case is worse — `R⁻¹` was computed fine and only `S` was singular, so SIR *could* in principle have run off `R⁻¹`, yet the message claims no proposal is buildable.

**Fix applied:** the message now says no usable SIR proposal could be built and points at the covariance-step warning that carries the real cause, rather than naming one. The test that pinned the old wording now also asserts the message does *not* contain `"eigen"`.

### 2. `src/api/postfit.rs` / `src/api/fit.rs` — low: a Bayesian fit is told to enable an option it already has on

`covariance_requested` was passed as `options.run_covariance_step && result.bayes.is_none()` (`fit.rs:1717`), so a Bayes fit run with `covariance = true` *and* `sir = true` took the `!covariance_requested` arm and got *"Enable covariance = true in [fit_options]"* — advice for an option the user had already set, i.e. exactly the failure mode this PR set out to remove. Suppressing the covariance step for Bayes is deliberate, but reusing that flag as "covariance was never requested" makes the SIR warning misleading.

**Fix applied:** `bayes_fit` is now a separate parameter and gets its own message (posterior credible intervals replace the Hessian-based covariance SIR would draw from), with a test and a `classify_warning` round-trip entry.

### 3. `docs/model-file/fit-options.qmd` — low: "in only two cases" is inaccurate

Same reason as finding 1 — any other `Unusable` covariance failure also skips SIR with no proposal — and it omitted cancellation (both `resolve_sir_fallback` and the new warning block bail on `is_cancelled`).

**Fix applied:** reworded to "when the covariance step produces neither a covariance matrix nor a fallback proposal", enumerating the real causes and noting the Bayesian and cancelled cases. The CHANGELOG entry carried the same overclaim and was corrected to match.

### Not a finding — worth flagging

`sir = true` now overrides an explicit `covariance_fallback = none` (the disjunction in `should_run_sir_fallback`). The PR body flags this as intended per issue #972 Option A and the docs row was updated, so I have not filed it — but it is the one behavioural change a user could be surprised by, since the `none` row still reads "leaves the covariance step as failed" before the new caveat sentence.

roninsightrx:
## Review — 3 findings, all fixed in 63158553

Gate change, warning routing and the test arity updates check out. The `classify_warning` ordering trap called out in the PR description is genuinely avoided: the new messages contain neither `"covariance step failed"` / `"covariance failed"` nor `"degenerate"`, and the `"covariance step"` arm is compound with `"not positive definite"`, which is absent. `sir = true` overriding `covariance_fallback = none` is intentional per #972 Option A and documented — not filed as a finding.

### 1. `src/api/postfit.rs:446` — medium: the message asserts a cause it cannot know

The "no proposal" message hard-asserted a single cause (*"its eigendecomposition did not produce finite eigenvalues"*), but `compute_covariance` returns `Unusable` — i.e. no `sir_fallback_proposal` — from at least five paths:

| `src/estimation/covariance.rs` | cause |
|---|---|
| :564 | `fd_hessian_step` not positive/finite |
| :702 | non-finite base OFV at convergence |
| :944 | `Hessian has ill-conditioned entries` — flat diagonal or non-finite FD stencil |
| :991 | eigendecomposition returned no finite eigenvalues (the one the message named) |
| :1025 | singular score cross-product `S` under `covariance_method = s`/`rsr` |

Concrete misbehaviour: a fit with one weakly-identified theta whose FD diagonal is ≈0 fails at :944 and the user is told the eigendecomposition diverged — wrong cause, and the actionable hint (tune `fd_hessian_step`, or drop the flat parameter) is lost. The `s`/`rsr` case is worse: `invert_psd_with_floor` **succeeded** there, so a perfectly good `R⁻¹` existed and only `S` was singular, yet the message claims no proposal could be built from the Hessian.

**Fixed**: the generic message now reads *"…no usable SIR proposal could be built from it, so SIR could not run — see the covariance warning above for the cause"* and points at the covariance-step warning that carries the real cause, instead of naming one. Regression assert added that the message contains no `"eigen"`.

### 2. `src/api/postfit.rs:430` — low: Bayesian fits get exactly the advice #972 removed

`covariance_requested` was passed as `options.run_covariance_step && result.bayes.is_none()` (`fit.rs:1717`). A Bayes fit with `covariance = true` and `sir = true` therefore takes the `!covariance_requested` arm and is told *"Enable covariance = true in [fit_options]"* — an option the user already set, and one that would not help anyway (`fit.rs:1194` disables the covariance/SIR steps for `EstimationMethod::Bayes` unconditionally). That is the same useless advice this PR set out to delete, just relocated. Note `options.sir` (not `stage_opts.sir`) drives the block at `fit.rs:1664`, so the path is reachable.

**Fixed**: `covariance_requested` is now the raw `options.run_covariance_step`, `bayes_fit` is a separate argument, and Bayesian fits get their own message — *"SIR requested but not run: Bayesian estimation reports posterior credible intervals instead of a Hessian-based covariance, which SIR would have to draw from."* Pinned in `classify_warning_roundtrips_every_engine_message` as `sir`/Warning.

### 3. `docs/model-file/fit-options.qmd:252` — low: "only two cases" is not the real condition

Same root as finding 1 — any other `Unusable` covariance failure also skips SIR with no proposal — and it omits cancellation (both `resolve_sir_fallback` and the new warning block bail on `is_cancelled`, so a cancelled fit skips SIR *without* the promised warning).

**Fixed**: the paragraph now states the real condition (neither a covariance matrix nor a fallback proposal), lists the failure modes, names the Bayesian case, and notes that a cancelled fit skips SIR without an extra warning.

---

`cargo test --lib` green (2832 tests, incl. 4 `sir_unavailable_warning` cases and the classifier round-trip), `cargo clippy --lib --tests` adds no new warnings, `cargo fmt --check` clean.

Re the open question: keeping `run_covariance()`'s proposal discard (`run_covariance.rs:274`) as a follow-up is the right call — folding it in here would widen the diff past what the classifier/warning change needs.

Issue #258, opened 2026-06-11
Missing DV on a scored observation row is silently read as DV=0 (no MDV=1 inference)


roninsightrx:
## Problem
A row with `EVID=0`, a **missing** `DV` (`.`, `NA`, or blank), and **no `MDV` column** (or `MDV=0`) is silently scored as a real **`DV = 0.0`** observation in the likelihood.

The observation DV is parsed with `parse_f64` (`src/io/datareader.rs:665`), which does `s.parse().unwrap_or(0.0)` and does **not** use the missing-aware `is_missing_cell` check that `parse_f64_or_nan` (`:674`) and `parse_evid` (`:690`) use. So `parse_f64(".") == 0.0`.

Because the observation arm only gates on `evid == 0 && mdv == 0` (`src/io/datareader.rs:1078`), a missing DV is **not** inferred as `MDV=1` — it must be set explicitly by the user. NONMEM users usually do set `MDV=1`, so this is often benign, but when they don't it is a silent data-corruption risk: phantom zero observations pull estimates and inflate residuals with no warning.

## Current behavior
- `EVID=0`, `MDV=1` → correctly excluded from the OFV. ✅
- `EVID=0`, `DV=.`, no/zero `MDV` → **scored as `DV=0.0`**. ❌ (this issue)

## Proposed fix
1. In the observation arm, treat an `EVID=0` row whose `DV` cell is missing (`is_missing_cell`: `.` / `NA` / `NaN` / blank) as `MDV=1` — i.e. skip it from the likelihood (do not push a `0.0` observation).
2. Emit a warning (new code, e.g. `W_MISSING_DV`) summarizing how many such rows were auto-skipped, surfaced in `FitResult.warnings` and `ferx check` (data path), consistent with the existing data-quality warnings (`W_ADDL_MISSING_II`, etc.). Classify it under the `data_quality` category in `classify_warning`.

## Acceptance criteria
- [ ] An `EVID=0` row with missing `DV` and no `MDV=1` is excluded from the OFV (no phantom 0.0 observation).
- [ ] A single summary warning (`W_MISSING_DV`) reports the count of auto-skipped rows; appears in fit warnings and `ferx check --data`.
- [ ] Documented in the check-report code registry (`src/diagnostics.rs` + `docs/src/file-formats/check-report.md`) and the data-format docs.
[...]

PR #981, opened 2026-08-18
feat(io): read a missing DV as a design point when simulating [closes #957]

### LINKED ISSUE #957: A missing DV should not exclude a row from simulation
Versions: `ferx-core` @ `54a7d44` (read), `ferx` 0.2.0 (tested), R 4.5.1.

`src/io/datareader.rs:1808-1830` skips an `EVID=0` record whose `DV` cell is `.`, `NA` or blank:

```rust
} else if evid == 0 && mdv == 0 {
    // Observation record
    let dv = parse_f64(row.get(dv_col).map(|s| s.as_str()).unwrap_or("0"));
    ...
    // A missing DV cell (`.` / `NA` / blank) means "no observation":
    // `parse_f64` coerced it to `0.0` above, so without this guard a
    // forgotten MDV would inject a phantom scored row. NONMEM convention
    // marks such rows MDV=1; treat a forgotten one the same way -- skip it
    // and count it for the single W_MISSING_DV summary (#258).
```

**For fitting this is right and I am not asking for it to change.** A record with no observation should not be scored, and #258's reasoning about a forgotten `MDV` is sound.

The question is whether it should also apply on the **simulation** path, where `DV` is the output rather than an input.

## Why it is a problem there

Simulating from a design — dosing plus sampling times, no observations yet — is the canonical use of `ferx_simulate()`. The natural way to write that template is `DV = .` at every sampling time, which is exactly how NONMEM writes "no observation here" and what `$SIMULATION` accepts. Today that returns zero simulated rows:

```r
tmpl <- data.frame(
  ID = 1, TIME = c(0, 0.25, 1, 4), DV = ".",
  EVID = c(1L, 0L, 0L, 0L), AMT = c(1.0, ".", ".", "."),
  CMT = 1L, MDV = c(1L, 0L, 0L, 0L)
)
# ferx_simulate(...) -> 0 rows
```

Substituting any number for `DV` — a value that is then overwritten by the simulated one — makes it work. Requiring a placeholder in the column you are about to generate is backwards, and it silently produces an empty result for the most natural input there is.

This is not hypothetical: it is what broke `ferxtranslate`'s concordance-data generator (`data-raw/generate_concordance_data.R`), which had been building `DV = "."` templates. The symptom was an empty frame and then an unrelated recycling error in downstream R code.

## What I am suggesting

Scope the skip to the paths where "not an observation" is the right reading:

1. **Fitting / estimation** — keep the current behaviour exactly as is.
2. **Simulation** — treat a missing `DV` on an `EVID=0`, `MDV=0` record as "simulate here", since the DV is being produced rather than consumed. The `MDV=1` exclusion stays: that is the user explicitly saying the record is not an observation, and it is unambiguous.

If that split is unattractive, the fallback is to make the current behaviour impossible to miss rather than merely warned about: on the simulate path, a run where *every* observation record was skipped for a missing DV is almost certainly a template, and returning zero rows there could be an error rather than a `W_MISSING_DV` line.

## Related

`W_MISSING_DV` is raised correctly here, but it never reaches the caller of `ferx_simulate()` — `ferx_fit()` has a warnings channel and `ferx_simulate()` does not. Filed separately as FeRx-NLME/ferx-r#283, since the fix belongs on that side. Together they are why this reads as a silent failure rather than a documented restriction: the engine says the right thing and nobody hears it.

roninsightrx:
## Why

`src/io/datareader.rs` skips an `EVID=0`, `MDV=0` record whose `DV` cell is `.` / `NA` / blank, and counts it for `W_MISSING_DV` (#258). That is the right reading when the `DV` is an **input**: a forgotten `MDV` must never inject a phantom scored `DV=0` into the likelihood.

It is the wrong reading when the `DV` is the **output**. Simulating from a design — dosing plus sampling times, `DV = .` everywhere because the values are what the run is about to produce — returned **zero rows**. Substituting any number into the column that is then overwritten made it work, which is backwards: the most natural template there is silently produced an empty result.

That template is exactly what NONMEM's `$SIMULATION` accepts, and it is what broke `ferxtranslate`'s concordance-data generator (`data-raw/generate_concordance_data.R`): an empty frame, then an unrelated recycling error downstream.

Closes #957.

## What changed

The split the issue asks for, option 1 (fitting unchanged, simulation rereads the row):

- **`MissingDvPolicy`** (`Skip` | `KeepAsDesign`), carried on `ObsRouting` — the struct that already rides the whole read path describing how an `EVID=0` row becomes an observation record. `Skip` is the `Default` and the policy on every fitting entry point, so fit behaviour is bit-identical.
- **`KeepAsDesign`** keeps the row as a design point: the sampling time is real, only the observation has not been generated yet. The retained row carries a placeholder DV — `NaN` for a Gaussian endpoint (never mistakable for a measured `0.0`; every downstream `|DV|` consumer already guards on `is_finite`), `0` for an integer-coded discrete-state / count endpoint (`NaN` is not a state index). The simulated value replaces it.
- **`MDV=1` still excludes the record on both paths.** That is the user explicitly saying the row is not an observation, and it is unambiguous. Only the *forgotten* `MDV` is reread.
- A present-but-invalid DV is still a hard error under either policy (`checked_integer_dv` is skipped only when the cell is actually missing).
- **`api::read_population_for_simulation()`** is the new public entry point. It shares its body with `read_population_for()` and differs only in the policy passed.

Supporting refactor — what lets a non-default policy reach every covariate / filter / routing combination without a flag on each wrapper:

- The six `read_nonmem_csv_*` wrappers now delegate to one **`read_nonmem_csv_routed`**, which builds the covariate read set (declared union, or the explicit lenient list) and augments it with any `[data_selection]`-referenced column. The union / augmentation logic existed in four near-copies and is now written once.
- `read_population_for`'s six-arm match collapses to a single call. Behaviour is unchanged: the previous "Gaussian-only fast path" differed from the TTE path only in passing an empty routing, which is what `tte_and_discrete` produces for a Gaussian-only model.
- `read_nonmem_csv_filtered_tte` / `read_nonmem_csv_with_covariates_tte` existed solely for `read_population_for` and are now dead, so they are deleted; the coverage test that drove them directly drives the general reader instead (and additionally asserts the covariate-table presence rule).

## Alternatives considered

- **A flag threaded through all six wrappers** — same reach, but a new parameter on each `pub(crate)` reader plus ~30 mechanical test-call-site edits, and the duplicated union/augmentation logic stays duplicated.
- **The issue's fallback** (keep the skip; make "every observation row was skipped" a hard error on the simulate path) — narrower: it converts a silent empty result into a loud one but still refuses the idiomatic template. The split delivers what the user meant to write. Not implemented, since doing both would make the error unreachable on the path that now succeeds.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#287 (Draft until this merges) | open | after |

This PR is the engine capability. `ferx_simulate()` reaches it only once the R glue's four simulate entry points (`ferx_rust_simulate`, `_from_fit`, `_adaptive`, `_with_uncertainty`) call `read_population_for_simulation` instead of `read_population_for`, which needs the `ferx-core` commit in `ferx-r/src/rust/Cargo.lock` bumped first (the `[patch]` only applies to local builds). Nothing here breaks ferx-r as it stands — `read_population_for` is untouched.

## Breaking changes
- [x] None

`read_population_for` keeps its signature and its behaviour; the new function is purely additive.

## Numerical validation
- [x] Not applicable (no estimation / PK / stats path is touched — this is a data-reading policy, and the fitting policy is unchanged)

The external anchor for the *new* behaviour is the convention itself: `DV = .` on a sampling record is what NONMEM's `$SIMULATION` accepts as "simulate here", and `MDV=1` is what NONMEM uses for "not an observation" — this change makes ferx read both the same way. Fit-side reading is byte-identical to before (`Skip` is the default, asserted against the same CSV in `skip_policy_still_drops_every_row_of_the_same_template`).

## Performance

One extra `bool` comparison per observation row; the wrapper collapse removes a `Vec<String>` clone on the declared-covariates path.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3047 pass, 0 fail)

Tier 1, `src/io/datareader_tests.rs` (both policies asserted against the *same* CSV, so a regression in either shows up as a diff between them):

- `keep_as_design_retains_missing_dv_rows_as_sampling_times` — all three sampling times survive, DVs are `NaN`, no `W_MISSING_DV`.
- `skip_policy_still_drops_every_row_of_the_same_template` — the fitting reading of that identical file, unchanged.
- `keep_as_design_still_excludes_mdv1_rows`
- `keep_as_design_leaves_present_dvs_alone` — a valued template reads identically under both policies.
- `keep_as_design_places_integer_endpoint_rows_with_a_zero_placeholder` — discrete-state / count design rows, and the same rows still skipped + warned under `Skip`.
- `keep_as_design_still_rejects_an_out_of_range_integer_dv`

Tier 1, new `src/api/tests/simulation_template_tests.rs`:

- `simulation_reader_keeps_a_dv_dot_template_the_fit_reader_drops` — the API-level split.
- `simulating_a_dv_dot_template_emits_one_row_per_sampling_time` — **the issue's exact failure**: `simulate()` on a `DV = .` template returns 3 rows with finite values instead of 0 rows.

<details>
<summary>Example</summary>

```csv
ID,TIME,DV,EVID,AMT,CMT,MDV
1,0,.,1,100,1,1
1,0.25,.,0,.,1,0
1,1,.,0,.,1,0
1,4,.,0,.,1,0
```

```rust
let (population, _covariates) =
    read_population_for_simulation(&model, &None, "design.csv", None, None, None, &[])?;
let sims = simulate(&model, &population, &model.default_params, 1000);
// 3 simulated records per replicate — one per sampling time.
// Read with read_population_for() instead: 0 records, plus a W_MISSING_DV warning.
```

</details>

## Docs
- [x] `docs/` updated for user-visible changes — `docs/data-format.qmd` (the missing-DV note now states the simulation reading) and `docs/api/simulation.qmd` (new `read_population_for_simulation()` section)
- [x] New pages linked in `docs/_quarto.yml` — N/A, no new pages

## Checklist
- [x] `cargo clippy` clean — no new findings in the changed files (the four pre-existing `approx_constant` errors in `src/sens/two_cpt_explicit.rs` are untouched and unrelated)
- [x] `Cargo.toml` version bumped if breaking — N/A, additive

### ferx-site
- [x] No new `.ferx` DSL syntax or `[fit_options]` key, no new example model, no new data file

### ferx-book

### Example execution
- [x] No example execution step needed — no `.ferx` file, CLI path, or example dataset behaviour changes; the CLI's `--simulate` uses the synthetic `[simulation]` block and never reads a CSV

## Reviewer hints

Focus on two things:

1. **The placeholder choice** (`src/io/datareader.rs`, the two `dv_missing` arms). `NaN` for Gaussian is deliberate over `0.0` so a design point can never be read as a measured zero; the population it produces is for simulation only, and `read_population_for_simulation`'s doc says not to hand it to `fit()`. Integer endpoints get `0` because `NaN` is not a state index.
2. **The wrapper collapse** (`read_nonmem_csv_routed` and `read_population_for`) — the only place fit behaviour could have moved. The parity argument is that the deleted Gaussian-only arm and the TTE arm differed solely in the routing sets, which are empty for a Gaussian-only model either way.

The rest is mechanical.

## Open questions

None blocking. One deliberate non-decision: whether `simulate()` should *also* refuse a population with zero observation rows, as the issue's fallback suggests. Left out — with the split in place, the case that motivated it now succeeds, and a hard error there would fire on legitimately dose-only designs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,16 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Added
+- **A missing `DV` no longer empties a simulation.** Simulating from a design — dosing plus
+  sampling times, with `DV = .` because the values are what the run is about to produce — used
+  to return zero rows: every `EVID=0` row with a missing `DV` was skipped as a forgotten
+  `MDV=1` (#258), which is the right reading only when the `DV` is an input. The new
+  `read_population_for_simulation()` reads such a row as a **design point** instead, so the
+  natural template simulates as written and no placeholder number is needed in the column
+  about to be overwritten. Fitting is unchanged, and `MDV=1` still excludes a record on both
+  paths (#957).
+
 ### Performance
 - **The ODE inter-occasion-variability (IOV) analytic sensitivity path now compiles from a
   bucketed set of dual widths instead of one specialisation per stacked axis count**, cutting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,9 @@ section of the SDLC for the versioning policy).
   `read_population_for_simulation()` reads such a row as a **design point** instead, so the
   natural template simulates as written and no placeholder number is needed in the column
   about to be overwritten. Fitting is unchanged, and `MDV=1` still excludes a record on both
-  paths (#957).
+  paths (#957). Kept design rows are reported as `W_DESIGN_DV`, the simulation-side counterpart
+  of `W_MISSING_DV`, so a simulation run off an *observed* dataset makes its extra rows visible
+  rather than silently carrying more rows than a fit of the same data.
 
 ### Performance
 - **The ODE inter-occasion-variability (IOV) analytic sensitivity path now compiles from a
@@ -42,6 +44,13 @@ section of the SDLC for the versioning policy).
   unchanged (#971).
 
 ### Fixed
+- **A design population is now rejected by `fit()` instead of being fitted to its placeholders.**
+  The population returned by `read_population_for_simulation()` carries a `NaN` placeholder for
+  each not-yet-generated observation, and nothing checked observations for finiteness: with
+  log-transformed-both-sides on natural-scale data the placeholder was floored to a finite,
+  extreme value and the fit ran to completion on fabricated data with no warning. Such a
+  population now fails up front with `E_NONFINITE_DV`, and propensity-score-matched simulation
+  reports the real cause instead of a misleading "EBE did not converge" (#957).
 - **The default (`Auto` → analytic-gradient NLopt L-BFGS) optimizer no longer stalls at the
   initial estimates on its first step.** From the identity initial Hessian the opening search
   direction is `−∇`, and on models where the scaled gradient is large at the start (e.g.
--- a/docs/api/simulation.qmd
+++ b/docs/api/simulation.qmd
@@ -35,6 +35,51 @@ for sim in &sims[..5] {
 }
 ```
 
+## `read_population_for_simulation()`
+
+Read a NONMEM-format CSV that will be **simulated from** rather than fitted.
+
+```rust
+pub fn read_population_for_simulation(
+    model: &CompiledModel,
+    covariate_decls: &Option<Vec<CovariateDecl>>,
+    data_path: &str,
+    fallback_columns: Option<&[&str]>,
+    iov_column: Option<&str>,
+    filter: Option<&SelectionFilter>,
+    column_map: &[(String, String)],
+) -> Result<(Population, Option<CovariateTable>), String>
+```
+
+Identical to `read_population_for()` — the reader every fitting entry point uses
+— except in how it reads a **missing `DV`** on a scored row (`EVID=0`,
+`MDV=0`). For a fit, a missing `DV` means "no observation here", so the row is
+skipped and counted for the `W_MISSING_DV` warning. For a simulation the `DV` is
+the *output*, so the same row is a **design point**: a sampling time whose value
+has not been generated yet.
+
+That makes the natural design template simulate as written:
+
+```csv
+ID,TIME,DV,EVID,AMT,CMT,MDV
+1,0,.,1,100,1,1
+1,0.25,.,0,.,1,0
+1,1,.,0,.,1,0
+1,4,.,0,.,1,0
+```
+
+```rust
+let (population, _covariates) =
+    read_population_for_simulation(&model, &None, "design.csv", None, None, None, &[])?;
+let sims = simulate(&model, &population, &model.default_params, 1000);
+```
+
+Read with `read_population_for()` the same file yields zero simulated rows. Rows
+marked `MDV=1` are excluded on both paths. The retained rows carry a placeholder
+`DV` (`NaN`, or `0` for an integer-coded endpoint) that the simulated value
+replaces, so the returned population is for simulation only — do not pass it to
+`fit()`.
+
 ## `simulate_with_seed()`
 
 Same as `simulate()` but with a fixed random seed for reproducibility.
--- a/docs/api/simulation.qmd
+++ b/docs/api/simulation.qmd
@@ -75,10 +75,19 @@ let sims = simulate(&model, &population, &model.default_params, 1000);
 ```
 
 Read with `read_population_for()` the same file yields zero simulated rows. Rows
-marked `MDV=1` are excluded on both paths. The retained rows carry a placeholder
-`DV` (`NaN`, or `0` for an integer-coded endpoint) that the simulated value
-replaces, so the returned population is for simulation only — do not pass it to
-`fit()`.
+marked `MDV=1` are excluded on both paths. Retained rows are reported once as a
+`W_DESIGN_DV` warning on the returned population, so a simulation run off an
+*observed* dataset makes visible that it kept rows a fit of the same data would
+have skipped.
+
+The retained rows carry a placeholder `DV` (`NaN`, or the endpoint's first
+declared state code for an integer-coded endpoint) that the simulated value
+replaces, so the returned population is **for simulation only** — do not pass it
+to `fit()`. That is enforced, not merely advised: a population carrying
+non-finite observations is rejected by `fit()` (and reported by `ferx check`) as
+`E_NONFINITE_DV`, and propensity-score-matched simulation
+(`SimulateOptions::match_method`) rejects it too, since a design template has
+nothing to compute posthoc etas from.
 
 ## `simulate_with_seed()`
 
--- a/docs/data-format.qmd
+++ b/docs/data-format.qmd
@@ -41,6 +41,15 @@ ferx-core reads data in NONMEM-compatible CSV format. This is the standard forma
 > the row is skipped rather than scored as `DV=0` — and emits a single
 > `W_MISSING_DV` warning reporting how many rows were skipped. Set `MDV=1`
 > explicitly to silence it, or fix the data if the missing values are an error.
+>
+> **Simulation reads it the other way round.** When the dataset is a *design*
+> being simulated from, the `DV` is the column about to be produced, so
+> `DV = .` at every sampling time means "simulate here" rather than "nothing to
+> score here". `simulate()` fed a population read by
+> `read_population_for_simulation()` keeps those rows and emits one simulated
+> record per sampling time; no placeholder number is needed in the `DV` column.
+> `MDV=1` still excludes the record on both paths — that is the explicit
+> statement that the row is not an observation.
 
 ## Inferring doses without an `EVID` column
 
--- a/docs/data-format.qmd
+++ b/docs/data-format.qmd
@@ -48,8 +48,11 @@ ferx-core reads data in NONMEM-compatible CSV format. This is the standard forma
 > score here". `simulate()` fed a population read by
 > `read_population_for_simulation()` keeps those rows and emits one simulated
 > record per sampling time; no placeholder number is needed in the `DV` column.
+> The kept rows are reported once as `W_DESIGN_DV`, so a simulation run off an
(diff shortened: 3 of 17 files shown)


roninsightrx:
## Code review — `5ab917c0`

Reviewed the datareader missing-DV policy split, the wrapper collapse, docs/changelog, and the mechanical import removals.

**Wrapper collapse verified.** Checked `read_nonmem_csv_routed` line by line against the six deleted arms: `declared_union` + `augment_with_filter` ordering, `decls`-vs-`covariate_columns` precedence, `ObsRouting::tte_and_discrete(∅, ∅) == default`, and `Option<CovariateTable>` presence all match the old behaviour. No dead references to the two deleted readers remain. No bug there.

Four findings, all on the same theme: the NaN / `0` placeholder is assumed to be a loud failure, but every path that could notice it is either absent or silently launders it.

### MEDIUM — `src/io/datareader.rs:1998` — `W_MISSING_DV` disappears on the simulate path

Under `KeepAsDesign` nothing counts the reread rows, so no warning is emitted in either direction.

Scenario: a VPC simulated from the *observed* dataset (the common case once ferx-r routes all four simulate entry points to `read_population_for_simulation`). Rows where the user genuinely forgot `MDV=1` are excluded by `fit()` but kept by the simulate reader — so the simulated dataset carries more rows than the fitted one, at times that were never scored, with no diagnostic.

Suggest keeping a counter for kept design rows and emitting an informational warning (e.g. `W_DESIGN_DV`) so the row-count divergence is visible.

### MEDIUM — `src/api/run.rs:793` — the "do not pass this to `fit()`" contract is documentation-only

Nothing enforces it, and the NaN placeholder is not the loud failure the PR assumes:

- `check_model_data` / `validate_*` never test observations for finiteness — `src/api/validation.rs` has an `is_finite` check only on TIME (line 1235).
- On the LTBS case-2 path, `log_transform_observations` (`src/api/run.rs:47-52`) silently launders the placeholder: `*v <= 0.0` is false for NaN so it is not counted, and `NaN.max(LTBS_FLOOR)` returns `LTBS_FLOOR` (Rust's `f64::max` returns the non-NaN operand). Every design point becomes a finite, extreme "observed" value `ln(LTBS_FLOOR)`, and the fit runs to completion on fabricated data with no warning.
- Integer endpoints are worse: the `0` placeholder is a legal state/count, so a mis-routed simulation population scores as all-zero states.

A cheap guard — reject non-finite / placeholder observations at `fit()` entry, or tag the `Population` as simulation-only — would make the misuse loud.

### MEDIUM/LOW — `src/api/simulate.rs:435` — propensity matching's precondition no longer catches a design template

`find(|s| s.observations.is_empty())` used to reject a `DV = .` design template with a clear message, because the reader dropped every row. It now passes — the rows exist, their DVs are NaN — and `run_inner_loop_warm` computes posthoc EBEs from a NaN objective. Either the non-finite-eta guard at line 461 fires with a misleading "its EBE did not converge" message, or (the likelier path, if the inner optimizer returns the finite starting eta on a NaN objective) matching proceeds on all-zero etas and silently produces an arbitrary assignment.

Worth an explicit finiteness check on `observations` in that precondition block.

### LOW — `src/io/datareader.rs:1965` — integer placeholder bypasses the range check

`checked_integer_dv` is skipped for design rows, so the placeholder `state: 0` can be an out-of-range code for endpoints whose declared `state_codes` are 1-based (`collect_state_obs`, `src/markov/endpoint.rs:68`, returns `Err(bad_code)` → `SUBJECT_SENTINEL_NLL`; `validate_ctmm_states` would reject it). Not reachable from `simulate()` today (CTMM simulation is rejected at `src/api/simulate.rs:829`), so this only bites via the unguarded fit misuse above — but the placeholder would be safer as the endpoint's first declared state code rather than a hard `0`.

roninsightrx:
All four addressed in `ab87eec6`.

**MEDIUM `datareader.rs` — no warning under `KeepAsDesign`.** The counter now runs under both policies and the summary picks its wording from `routing.missing_dv`: `W_MISSING_DV` for skipped rows, `W_DESIGN_DV` for kept ones, so the row-count divergence between a fit and a simulation of the same file is visible from either side. Documented in `check-report.qmd`, `data-format.qmd`, and `api/simulation.qmd`.

**MEDIUM `run.rs` — documentation-only contract.** Now enforced: `check_finite_observations` runs first in `check_model_data_rule`, so `fit()` stops with `E_NONFINITE_DV` (and `ferx check` reports it) before the placeholders reach the likelihood. The diagnostic's suggestion names `read_population_for` as the fix. Separately, `log_transform_observations` now skips non-finite values instead of flooring them, so the LTBS case-2 laundering path is closed even if the guard is ever bypassed — and a `NaN` no longer counts as a non-positive DV.

**MEDIUM/LOW `simulate.rs` — propensity precondition.** Added a finiteness check next to the emptiness one, with a message that names the real cause ("a `DV = .` design template carries NaN placeholders") rather than letting it surface as an EBE convergence failure downstream. The `simulate_with_options_diag` doc comment was updated to match.

**LOW `datareader.rs` — integer placeholder.** `ObsRouting` gained a `design_states: HashMap<usize, usize>` map; `read_population_for_policy` populates it from each CTMM endpoint's first declared `state_codes` entry, so a `1`/`2`-coded table gets a decodable placeholder. Unregistered CMTs keep `0` (Binary is `{0,1}`, counts start at 0), so it is opt-in per endpoint.

Tests (all Tier 1): two in `datareader_tests.rs` (`W_DESIGN_DV` count, registered-vs-default placeholder) and four in `api/tests/simulation_template_tests.rs` (warning on the real reader, `E_NONFINITE_DV` on a design population with the fit-read population staying clean, `log_transform_observations` NaN handling, propensity rejection with matching-off still simulating). `cargo test --lib` 2841 pass; `--features markov` 3013 pass; `cargo check --tests`, `fmt`, and `clippy` clean.

Issue #985, opened 2026-08-18
Mixture models: other estimators (SAEM/IMP/Bayes) + IOV (#983 follow-up)


roninsightrx:
## Context

Follow-up to #983 (which itself follows #977 / #980, the mixture-model /
`$MIXTURE` work). PR #984 landed the bounded items of #983 — standard errors /
covariance (Phase 6, NONMEM `$COVARIANCE MATRIX=R` cross-check to < 1 %), the
`p(k)` mixing form, and `.fitrx` round-trip of `MIXEST`/`PMIX_*`.

This issue tracks the two **large** items #983 deferred. Both currently error
clearly at `fit()` rather than silently mis-fitting (`src/api/fit.rs`), so this is
new capability, not a bug fix.

### 1. Other estimators for mixtures
Only FOCE/FOCEI estimate a mixture today; `fit()` rejects the rest. Wire, in rough
order of value:

- **SAEM** — the E-step must sample the latent class indicator alongside `η`
  (per-subject class draw from the current posterior `p_ik`, then the usual η
  MCMC within the drawn class). The M-step updates the mixing coefficients from
  the sampled class frequencies.
- **IMP / IMPMAP** importance sampling — the proposal and weights must marginalise
  over the class (or sample it), mirroring the FOCEI mixture marginal
  `L_i = Σ_k p_ik L_ik`.
- **Bayes (HMC)** — the class indicator becomes a discrete latent variable;
  marginalise it out of the per-subject likelihood (Rao-Blackwellised, as in the
  FOCEI objective) rather than sampling it, so HMC still sees a continuous target.

Each is effectively its own project; they can land in separate PRs.

### 2. Inter-occasion variability under a mixture
`fit()` rejects IOV + mixture together. The per-(subject × class) inner solve
would need to carry per-occasion `κ` for each class, and the covariance/packing
layout (`estimation::mixture`, `estimation::parameterization`) would need the IOV
segment to coexist with the per-class Ω/Σ override tail. Note the current mixture
covariance code assumes no IOV (`mixture_gradient` places the override segment
immediately after σ) — that assumption must be revisited here.

## Validation
Extend the NONMEM anchor family (`tests/nonmem/mixture_iv*`) as each item lands:

- **SAEM** → NONMEM `METHOD=SAEM` on the mixture anchor.
- **IMP / IMPMAP** → NONMEM `METHOD=IMP` / `IMPMAP`.
[...]

Issue #983, opened 2026-08-18
Mixture models: SEs/covariance, other estimators, IOV, p(k) form (#977 follow-up)


roninsightrx:
## Context

Follow-up to #977 (mixture models / `$MIXTURE`). PR #980 landed Phases 1–5:
parser + data model + validation, per-class Ω/Σ packing, FOCE/FOCEI mixture
objective, analytic posterior-weighted outer gradient (+ gradient optimizers),
and the per-subject `MIXEST` / `PMIX_1..PMIX_K` outputs with a NONMEM 7.5.1
cross-check (ferx OFV 301.600 vs NONMEM 301.582; MIXEST 30/30 subjects agree).

This issue tracks what PR #980 deliberately left out — everything currently
errors clearly rather than silently mis-fitting.

### 1. Standard errors / covariance step (Phase 6 — highest priority)
Mixture fits skip the covariance step entirely, so there are no SEs / RSEs / a
covariance matrix. Needed:
- The FD-of-OFV Hessian must be built on the **mixture** OFV
  (`estimation::mixture::mixture_ofv`), not the single-population objective the
  covariance step currently assumes.
- `packed_param_label` (`estimation/covariance.rs:47`) mislabels the per-class
  Ω/Σ override segment and the mixing-logit thetas as `packed[N]` — give them
  real names (`ETA_MIX{k}` / `SIGMA_MIX{k}`, mixing thetas by their theta name).
- Decide the SE convention for the mixing fraction (report on the logit scale
  and/or delta-method back to `p(k)`).

### 2. Other estimators for mixtures
Currently only FOCE/FOCEI. `fit()` rejects the rest. Wire, in rough order:
- **SAEM** (the E-step must sample the latent class alongside η).
- **IMP / IMPMAP** importance sampling.
- **Bayes** (HMC) — the class indicator becomes a discrete latent variable.

### 3. Inter-occasion variability under a mixture
`fit()` rejects IOV + mixture together. The per-(subject × class) inner solve
would need to carry per-occasion κ.

### 4. `p(k)` probability mixing form
`p(k) = …` is parsed and validated but not evaluated — only `logit(k)` is wired
into `eval_mixing_log_probs` / `mixing_logp_grad`. Either evaluate it (softmax
is skipped; probabilities used directly, renormalised) or reject it at parse
time until it is.

### 5. `.fitrx` checkpoint round-trip of PMIX/MIXEST
[...]

PR #986, opened 2026-08-18
feat(estimation): inter-occasion variability under a mixture model [#985]

roninsightrx:
## Why

`fit()` rejected a `[mixture]` model that also carried inter-occasion variability (a `kappa` term). This was one of the two large items #983 deferred to #985; the other estimators (SAEM/IMP/Bayes) remain separate follow-ups. IOV + mixture is a natural combination (subpopulations that also vary occasion-to-occasion), and NONMEM supports it via `$OMEGA BLOCK(1) … SAME` under `$MIX`.

## What changed

The packing groundwork was already IOV-aware — `pack/unpack_params`, `packed_len`, `packed_fixed_mask`, and `omega_structural_zero_mask` all lay out `[theta, omega, sigma, omega_iov, mixture_overrides]` — and `find_ebe` already routes to the IOV inner solve when `n_kappa > 0` (each class's `class_params` clone keeps the shared base `omega_iov`). Three gaps remained:

1. **`mixture_ofv`** called the plain `foce_subject_nll`, silently dropping the per-occasion κ̂ the inner solve produced. It now calls `foce_subject_nll_iov` (with `ebe.kappas` + the class-shared `omega_iov`) when `n_kappa > 0`, so the κ-augmented marginal enters each class likelihood `L_ik`. Non-IOV mixtures keep the byte-identical legacy call.
2. **`mixture_gradient`**'s override-tail base `ov_base` omitted the IOV segment length (`nt + n_omega + n_sigma`), colliding with the κ slots — the exact collision #985 flagged. It now adds `n_iov`, and returns `None` (→ central-FD outer gradient via `mixture_gradient_fd`) when `omega_iov` is present, since the per-class analytic sensitivity provider does not yet emit κ-slot gradients.
3. Dropped the IOV-rejection guard at `fit.rs`.

IOV is **class-shared** (it lives in the base parameter block, ahead of any per-class Ω/Σ override tail), matching NONMEM `$OMEGA BLOCK(1) … SAME`. The covariance step is unchanged — it already builds its FD Hessian on the K-fold mixture objective, and the packed layout was already IOV-aware.

## Alternatives considered

Emitting an analytic κ-slot outer gradient for the mixture (rather than FD fallback) — deferred: the per-class sensitivity provider would need κ columns per class, and the FD outer gradient is correct and cheap enough. Left `ov_base` layout-correct (`+ n_iov`) so an analytic κ path can be added later without touching the offsets.

## Cross-repo dependency

No public API or output-field change.

## Breaking changes
- [x] None

## Numerical validation
- [x] Compared against: **NONMEM 7.5.1** (`tests/nonmem/mixture_iv_iov.ctl`, licensed pmx container)
- [x] Reference values:

```
# NONMEM (fixed IIV/IOV Ω + Σ, METHOD=1 INTER)   ferx
OFV (no const)  470.814                          470.930
TVCL1           1.057                            1.069
TVCL2           3.330                            3.275
V               10.259                           10.253
p(1)            0.526                            0.506
MIXEST          — (sdtab)                        30/30 subjects agree
```

Standard errors are **not** cross-checked for this model: NONMEM's covariance step terminates with rounding errors and substitutes the R matrix (`R MATRIX SUBSTITUTED: YES`, no printable SE block / `.cov`) — a known NONMEM fragility for mixture + IOV covariance. ferx's covariance step completes. The mixture SE machinery is anchored to `$COVARIANCE MATRIX=R` in the non-IOV `mixture_nonmem.rs`.

## Performance
- [x] No significant impact expected (IOV mixture pays the usual K× inner-solve cost; non-IOV path unchanged).

## Tests
- [x] Unit tests added (`cargo test --lib` passes, 2890 total):
  - `iov_mixture_fit_runs_and_emits_posteriors` — fit() returns Ok, finite OFV, K=2 posteriors
  - `iov_mixture_ofv_differs_from_no_iov_marginal` — the IOV marginal actually enters the objective
  - `iov_mixture_gradient_falls_back_to_fd` — analytic gradient returns None; FD gradient full-length + finite
  - `iov_mixture_covariance_step_runs` — covariance step succeeds over the full packed layout
- [x] Slow-gated NONMEM cross-check: `tests/mixture_iov_nonmem.rs`

- [x] Inline below

<details>
<summary>Example</summary>

```toml
[parameters]
  theta TVCL1(1.0, 0.01, 100.0)
  theta TVCL2(3.0, 0.01, 100.0)
  theta TVV(10.0, 0.1, 1000.0)
  theta P1(0.5, 0.001, 0.999)
  omega ETA_CL ~ 0.05
  kappa KAPPA_CL ~ 0.02       # inter-occasion variability on CL
  sigma EPS ~ 0.04

[mixture]
  nsub = 2
  p(1) = P1

[individual_parameters]
  CL = if (MIXNUM == 1) TVCL1 * exp(ETA_CL + KAPPA_CL) else TVCL2 * exp(ETA_CL + KAPPA_CL)
  V  = TVV

[structural_model]
  pk one_cpt_iv(cl=CL, v=V)

[error_model]
  DV ~ proportional(EPS)

[fit_options]
  iov_column = OCC
```
</details>

## Docs
- [x] `docs/estimation/mixture.qmd` — new "Inter-occasion variability under a mixture" section + NONMEM table; removed IOV from "Not yet supported"

## Checklist
- [x] `cargo clippy` clean (no new warnings on changed lines)
- [x] `Dual2` sensitivity path untouched (IOV mixture routes to FD outer gradient; inner κ sensitivities were already validated for non-mixture IOV)

## Reviewer hints

Focus on `mixture_ofv`'s IOV branch (`src/estimation/mixture.rs`) and the `ov_base` / `None`-gate in `mixture_gradient`. The claim that IOV is class-shared rests on `class_params` cloning the base `omega_iov` unchanged — worth confirming that's the intended semantics (it matches `$OMEGA … SAME`). The NONMEM covariance-step failure is documented in the test's module comment.

## Open questions

Per-class IOV Ω overrides (`omega_iov(k)`) are out of scope — IOV is currently class-shared. Flag if a per-class IOV variance is wanted; it would need a new field on `MixtureParams` and override-tail plumbing.


roninsightrx:
## Code review (medium) — findings + fixes

Reviewed `aee7cab4` (IOV under a mixture). Four findings; the correctness/robustness three are fixed in `6fb11b99`, the perf-only one is deferred with rationale.

### 1. ⛔ Postfit discarded per-subject κ EBEs — **fixed**
`MixtureEval` carried no kappas, so the final mixture branch (`outer_optimizer.rs:1903`) returned `Vec::new()` for `result.kappas`. The removed `fit.rs` guard now lets an IOV mixture reach postfit, so `compute_subject_results` hit the `kappas.is_empty()` branches: sdtab IPRED/IWRES/CWRES and per-subject OFV were computed at **κ = 0** (IOV ignored), κ shrinkage was never computed (`fit.rs:1599` false), and `.fitrx` `ebe_kappas` was empty. Total OFV was correct (from `foce_subject_nll_iov`), but the per-subject diagnostics — the user-visible outputs of #985 — were wrong, and the new test only checked pmix/mixest/finite-OFV so it passed anyway.

**Fix:** `MixtureEval` now carries `mixest_kappas` (winning class's per-occasion κ̂ per subject); the final inner loop threads it into `FitResult.kappas`. Empty inner `Vec` for a non-IOV mixture, so non-IOV behaviour is byte-identical.

### 2. 🧹 Dead `n_iov` in `mixture_gradient` — **fixed**
The function returns `None` for any `omega_iov.is_some()` model above line 297, so `n_iov` was provably 0 on every reachable path. Removed; `ov_base = nt + n_omega + n_sigma`.

### 3. ⚠️ Gating-predicate divergence — **fixed**
`mixture_ofv` branched on `model.n_kappa > 0` while `mixture_gradient` gates on `params.omega_iov.is_some()`; a present-`n_kappa`-but-absent-`omega_iov` model would hit `.expect("IOV model (n_kappa > 0) has omega_iov")` and panic in-library. `mixture_ofv` now keys the branch on `cp.omega_iov` (the exact field the nll unwraps, same predicate as the gradient), so the two paths can't disagree on scope and no unwrap can fire.

### 4. 🐌 Warm-start carries η but not κ — **deferred (perf only)**
`mixest_etas`/`cached_etas_by_class` store the BSV η only; `find_ebe_iov`'s warm path fills η slots and cold-restarts κ from 0 each outer eval, costing extra inner iterations on IOV mixtures. Correctness of the objective at fixed params is unaffected. The clean fix needs a warm-κ argument threaded through `find_ebe`/`find_ebe_iov` (a hot shared fn with 35 call sites) plus a parallel per-class κ warm cache — disproportionate for a few extra inner iters on one narrow model class. Left as a follow-up.

### Test
`iov_mixture_postfit_carries_kappa_ebes` (Tier-2) asserts a fit populates `ebe_kappas` (one entry per subject, 2 occasions × 1 κ each, ≥1 non-zero) and a populated `shrinkage_kappa` — both empty before the fix. Full `mixture`/`iov` lib suites green; no new clippy warnings.

roninsightrx:
## Copilot review — addressed (`bc4fcebf`)

Both comments applied:

1. **`iov_mixture_covariance_step_runs` empty kappas slice** — now passes `&eval.mixest_kappas` (the real MIXEST-class per-occasion κ̂) instead of `&[]`, so the test drives the full IOV packed layout end-to-end and won't silently under-exercise if the covariance method later consumes κ. Passes.

2. **`#986 regression` comment** — the fix is part of the #985 feature (found in this PR's review); comment changed to `#985 regression (PR #986 review)` to match the file's other IOV markers.

All `iov_mixture` lib tests green.

PR #991, opened 2026-08-19
feat(estimation): Bayesian estimation for mixture models [#985]

roninsightrx:
> **Stacked on #987** (SAEM for mixtures). Base is `worktree-985-saem-mixture`; review/merge after #987. The only shared change is the mixture method-gate in `fit.rs`.

## Why

Issue #985's third estimator sub-item: **Bayesian (full-MCMC) estimation** for `[mixture]` models. Until now `fit()` rejected `method = bayes` for a mixture. (SAEM landed in #987; IOV in #986. IMP/IMPMAP remains.)

## What changed

The latent class is **Rao-Blackwellised** — marginalised out of every Gibbs block rather than sampled — so each subject's likelihood contribution is the K-class marginal `−log Σ_k p_ik·exp(−nll_ik)`, a smooth continuous target (the issue's prescribed approach).

The single lever is `subject_nll` (the one per-subject NLL every Gibbs block already calls): making it return the class-marginal (via `combine_subject` over per-class `individual_nll` under a `MixtureClassGuard`) makes the whole sampler mixture-aware at once:

- **η block** — samples against the class-marginal posterior via a new `mh_steps_mixture` (HMC is disabled for mixtures: its Dual2 gradient is single-class);
- **(θ, σ) block** — the componentwise/joint Metropolis already sums `subject_nll`, so the mixing thetas (constant *or* covariate-dependent logit mixing) are sampled with no extra machinery — they enter only through the marginal;
- **Ω block** — the conjugate inverse-Wishart draw from `S = Σ ηη'` gives the class-shared Ω.

Ω/σ are class-shared; per-class overrides (`omega(k)`/`sigma(k)`) are rejected with a clear error. The post-loop pass routes through `mixture_ofv`, so OFV, per-subject `MIXEST`/`PMIX`, and EBEs are the K-fold marginal values at the posterior mean.

## Alternatives considered

Sampling the class indicator (as SAEM does) was the alternative; the issue specifies Rao-Blackwellisation for Bayes so HMC/MH see a continuous target, and it turns out to collapse to a one-line change (`subject_nll` → marginal) rather than a per-class M-step restructure.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

- [x] Recovers the mixture MLE under diffuse priors:

```
                 ferx Bayes (post. mean)   FOCEI MLE (NONMEM mixture_iv.ctl)
TVCL1            0.94                       0.981
TVCL2            2.78                       2.842
V                10.0                       9.98
p(1)             0.47                       0.472
OFV (marginal)   301.8                      301.58
```

- A direct NONMEM `METHOD=BAYES` reference is **not** used: NONMEM's BAYES sampler aborts in burn-in on this model (`$MIX` + FIXed `$OMEGA`/`$SIGMA` — it insists on Gibbs-sampling Ω, which is FIXed), a known NONMEM fragility (cf. the covariance-step failure on IOV mixtures). The shared optimum the Bayes posterior recovers is itself NONMEM-anchored by the FOCEI run, and the identical mixture-marginal machinery is anchored directly to NONMEM SAEM in #987.

## Tests
- [x] Unit tests (`cargo test --lib` passes): `mixture_subject_nll_is_marginal_lse` (marginal == hand-computed log-sum-exp), `bayes_mixture_fit_populates_mixest` (short fit, MIXEST set), `bayes_rejects_per_class_overrides`.
- [x] Tier-3 slow: `bayes_mixture_recovers_optimum` in `tests/mixture_nonmem.rs`.

## Docs
- [x] `docs/estimation/mixture.qmd` — new "Bayesian estimation under a mixture" section

## Checklist
- [x] `cargo clippy` clean (new code warning-free)
- [x] No `Dual2`/`sens` path touched (HMC gradient is disabled for mixtures, not modified)

## Reviewer hints

- The whole feature is `subject_nll`'s mixture branch + `mh_steps_mixture` + the post-loop `mixture_ofv` routing; the rest of `bayes.rs` is untouched. The `mu-ref θ Gibbs` block needs no change: a class-switched typical value cannot be paired to one η, so `full_mu_ref` is already false for these models and all thetas (incl. the mixing thetas) flow through the marginal-aware Metropolis block.

## Open questions
- IMP/IMPMAP for mixtures is the last remaining #985 sub-item.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,17 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
+- **Bayesian estimation for mixture models (#985).** `[mixture]` models can now be fit with
+  `method = bayes`. The latent class is Rao-Blackwellised — marginalised out of every Gibbs block
+  rather than sampled — so each subject's likelihood contribution is the K-class marginal
+  `−log Σ_k p_ik·exp(−nll_ik)` and the sampler keeps a smooth continuous target. The η block samples
+  against the class-marginal posterior (HMC is disabled for mixtures; its analytic gradient is
+  single-class), the (θ,σ) block samples the mixing thetas (constant or covariate logit) with no extra
+  machinery, and Ω is drawn from its class-shared conjugate conditional. Reported OFV, per-subject
+  `MIXEST`/`PMIX`, and EBEs are the K-fold marginal values at the posterior mean. Per-class Ω/σ
+  overrides are rejected (Ω/σ are class-shared). Validated by recovering the mixture MLE under diffuse
+  priors (a direct NONMEM `METHOD=BAYES` reference is impractical — its sampler aborts on `$MIX` with
+  FIXed Ω/Σ).
 - **SAEM estimation for mixture models (#985).** `[mixture]` models can now be fit with
   `method = saem`, not only FOCE/FOCEI. The E-step samples the latent class per subject (from the
   current posterior `PMIX_i`) and runs the η-MCMC within the drawn class; the M-step estimates the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,11 @@ section of the SDLC for the versioning policy).
   unchanged (#971).
 
 ### Fixed
+- **Per-subject diagnostics now honour `MIXEST` in a mixture fit (#985).** `IPRED`, `PRED`,
+  `IWRES`, `CWRES`, `EBE_OFV`, and `[derived]`/`[output]` columns were computed with `MIXNUM`
+  pinned to class 1 for every subject, even for subjects the fit assigned to another class — so a
+  class-2 subject's sdtab row paired its class-2 EBEs with class-1 typical values. They are now
+  evaluated in each subject's fitted class. Affects FOCE/FOCEI, SAEM, and Bayes mixture fits.
 - **Mixture covariance-step and diagnostics correctness (#984, follow-up to #983).** Five fixes to
   the mixture SE/covariance and checkpoint paths: (1) the covariance step now reconverges its per-class EBEs at
   `cov_inner_tol`, not the fit's `inner_tol`, so a loose fit followed by a tight `cov_inner_tol` for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,10 @@ section of the SDLC for the versioning policy).
   `MIXEST`/`PMIX`, and EBEs are the K-fold marginal values at the posterior mean. Per-class Ω/σ
   overrides are rejected (Ω/σ are class-shared). Validated by recovering the mixture MLE under diffuse
   priors (a direct NONMEM `METHOD=BAYES` reference is impractical — its sampler aborts on `$MIX` with
-  FIXed Ω/Σ).
+  FIXed Ω/Σ). Inter-occasion variability is supported: the κ block samples against the same
+  class-marginal target. A run that asks for HMC (`saem_n_leapfrog > 0`) on a mixture now warns that
+  the Metropolis-Hastings η kernel was used instead, and a high R̂ on a mixture warns about label
+  switching, which would make the reported posterior mean an average across class labels.
 - **SAEM estimation for mixture models (#985).** `[mixture]` models can now be fit with
   `method = saem`, not only FOCE/FOCEI. The E-step samples the latent class per subject (from the
   current posterior `PMIX_i`) and runs the η-MCMC within the drawn class; the M-step estimates the
--- a/docs/estimation/mixture.qmd
+++ b/docs/estimation/mixture.qmd
@@ -218,11 +218,47 @@ posterior sits almost exactly on the class boundary, which NONMEM SAEM and ferx
 SAEM split marginally differently. Covariate-dependent mixing under SAEM is
 exercised end-to-end in `tests/mixture_nonmem.rs`.
 
+## Bayesian estimation under a mixture
+
+Mixtures can also be fit with full-MCMC Bayesian estimation (`method = bayes`,
+#985). The latent class is handled **Rao-Blackwellised** — marginalised out
+rather than sampled — so the sampler still sees a smooth continuous target: each
+subject's likelihood contribution is the K-class marginal
+$-\log \sum_k p_{ik}\,\exp(-\mathrm{nll}_{ik})$. Because every Gibbs block
+evaluates the per-subject likelihood through this one marginal, the whole
+Gibbs-within-MH sampler becomes mixture-aware in one place:
+
+- the **η block** samples $\eta_i$ against the class-marginal posterior (a
+  `chol(Ω)`-preconditioned random walk; HMC is disabled for mixtures because its
+  analytic gradient is single-class);
+- the **(θ, σ) block** — which also carries the mixing thetas — samples them
+  against the marginal, so the mixing coefficients (constant *or*
+  covariate-dependent logit mixing) are sampled with no extra machinery;
+- the **Ω block** draws the (class-shared) Ω from its conjugate full conditional.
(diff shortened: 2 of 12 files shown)


roninsightrx:
## Code review — Bayes mixture implementation (medium effort)

Reviewed the Bayes path specifically. One HIGH finding (a real sampling bug), plus a few smaller ones. **All actionable findings are fixed in the commit above**; one is a pre-existing issue outside this diff.

---

### 🔴 HIGH — the κ block broke the Rao-Blackwellisation (fixed)

`bayes.rs` routed the IOV κ Gibbs block through the SAEM kernel `mh_kappa_steps`, which scores its proposal with `individual_nll_iov` — *single-class*, evaluated with the thread-local `MIXNUM` at its default of 1. Its `nll_current` argument is `nll[i]`, which the η block had just set to the **K-class marginal**. So the acceptance test compared a marginal current value against a class-1 proposal: an offset of roughly `−log p_i1`, biasing acceptance and making κ target `p(κ | class 1)` instead of the class-marginal posterior. It also wrote a single-class value back into `nll[i]` (repaired only by the pre-block-3 refresh, so the θ block survived).

This is reachable today: `fit.rs` allows `method = bayes` on a mixture and #986 allows mixture + IOV, so any `[mixture]` model with `OCC` + `kappa` hits it. No test covered the combination.

**Fix:** new `mh_kappa_steps_mixture`, which evaluates the proposal through `subject_nll` so both sides of the ratio are the same marginal target every other Gibbs block uses. Regression test `mixture_kappa_kernel_targets_the_marginal` pins it two ways — a zero-scale move must return the marginal baseline untouched (the single-class kernel returns a value differing by `−log p_i1`), and after a live move the returned NLL must equal `subject_nll` at the κ actually retained. Added `bayes_iov_mixture_fit_runs` as an end-to-end smoke test.

### 🟠 MEDIUM — no label-switching guard on the class-specific θ posterior mean (fixed: warning)

Class labels are identified only by the `POP_PRIOR_SD = 10` normal prior centred on the user's inits — effectively diffuse on the log scale. If a chain relabels `TVCL1 ↔ TVCL2`, the reported posterior *mean* averages across modes and estimates neither class. The PR's own numbers are consistent with partial mode-mixing: TVCL1 0.94 vs MLE 0.981 and TVCL2 2.78 vs 2.842, both pulled *toward each other*, while the class-symmetric V (10.0 vs 9.98) and p1 (0.47 vs 0.472) are exact. The 10% tolerance in the slow test hides it.

**Fix:** a mixture-specific warning when max split-R̂ exceeds the convergence threshold, naming label switching and pointing at better-separated inits; documented in `mixture.qmd`, including the direction the observed means are pulled. An ordering constraint would be the stronger fix but is a behavioural change worth its own issue.

### 🟠 MEDIUM — "the `$THETA` bounds act as uniform priors" is not true (fixed: docs)

The (θ, σ) block (componentwise and joint) sets `theta[idx] = u_new.exp()` with no clamp to `theta_lower`/`theta_upper` and no rejection outside them — only the mu-ref conjugate move clamps. The actual prior is `N(u0, 10²)` centred on the initial value. Two consequences: the stated validation rationale in `tests/mixture_nonmem.rs` was wrong, and `MIXL(0.0, -10.0, 10.0)` can wander outside its declared bounds.

**Fix:** corrected the claim in the test doc comment and documented the real prior in `mixture.qmd`. Note the prior being centred on the inits also weakens "diffuse priors ⇒ recovers the MLE" as an anchor, since the inits sit near the MLE — worth keeping in mind for the validation story. *Enforcing* declared bounds in the (θ, σ) block is pre-existing Bayes behaviour and a separate change.

### 🟡 LOW — HMC silently disabled for mixtures (fixed)

A user who sets `saem_n_leapfrog > 0` got the MH kernel with no notice, and `n_divergent` was reported as 0, which reads as "HMC ran cleanly". Every other capability downgrade in this function pushes a warning. Added one (test: `bayes_mixture_warns_when_hmc_requested`).

### 🟡 LOW — the per-class-override rejection fired too late in a method chain (fixed)

`method = [focei, bayes]` on a model with `omega(k)`/`sigma(k)` ran the entire FOCEI stage to convergence and only then errored. Moved the check into the `fit()` mixture gate, which already walks `options.method_chain()` before any stage runs; `bayes.rs` keeps its own check defensively. Test extended.

### 🟡 LOW — `meval.ebe_stats` was discarded (fixed)

The mixture branch dropped the `InnerLoopStats` from `mixture_ofv` and hardcoded `ebe_convergence_warnings: 0` / `total_ebe_fallbacks: 0` (the non-mixture branch dropped its stats too). Per-class EBE non-convergence at the posterior mean is likelier here — K multimodal inner solves, cold-started — and the same failures produce warnings on the FOCEI mixture path. Now plumbed through for both branches.

### 🟡 LOW — the reported mixture OFV depends on `options.interaction` (fixed: docs)

`mixture_ofv` picks FOCE vs FOCEI Laplace from `options.interaction`. For a scalar `method = bayes` that keeps its `true` default, but the chain form sets `opts.interaction = (chain.last() == FoceI)` — so `method = [saem, bayes]` reports a **non-interaction** marginal OFV while `method = bayes` reports the interaction one, same model and data. Documented, with the caveat that Bayes OFV now means two different things (Laplace marginal for a mixture, posterior-mean joint NLL ×2 otherwise).

---

### Not fixed — out of this diff's scope

**🟠 `compute_subject_results` ignores MIXEST (`src/api/postfit.rs:553`).** It is called with no `MixtureClassGuard`, so IPRED/IWRES/CWRES for every subject are computed with `MIXNUM == 1`, even for subjects the fit assigned `MIXEST = 2`. The docs state "EBEs are the K-fold marginal values" — the η̂ do come from the MIXEST class, but the predictions built from them do not. Pre-existing and affects FOCEI/SAEM mixtures equally, so it wants its own issue rather than a fix here. Happy to file it.

**🟡 Two heap allocations per `subject_nll` call.** `eval_mixing_log_probs` returns a fresh `Vec` and `nll` is a fresh `vec![0.0; k]`, on a function called `n_subjects × (n_pop + 1) × sweeps × chains` times, each doing K full likelihood evaluations. `logp` depends only on θ and baseline covariates, so it could be hoisted per subject per θ-move; the `nll` buffer could live in a scratch struct alongside `EventPkParams`. Worth doing if Bayes mixture runtime becomes a complaint.

---

Verified after the fixes: `cargo test --lib` — 2907 passed, 0 failed; `cargo check --tests` clean; `cargo fmt` applied.

roninsightrx:
Follow-up: merged latest `main` (which now carries the SAEM-mixture work from #987) into this branch and retargeted the PR base from `worktree-985-saem-mixture` to `main`. One conflict, in `docs/estimation/mixture.qmd` — both sides added a bullet to the *Not yet supported* list (main: a theta shared between the mixing expression and a structural typical value under SAEM; this branch: per-class Ω/σ overrides under Bayes). Resolved by keeping both. `cargo test --lib` green after the merge.

Also fixed the `postfit.rs` MIXEST issue from the review rather than deferring it: `compute_subject_results` ran with no `MixtureClassGuard`, so IPRED/PRED/IWRES/CWRES/EBE_OFV were computed at `MIXNUM = 1` for every subject — a subject the fit assigned to class 2 got its class-2 EBEs paired with class-1 typical values. `compute_extra_output_columns` had the same gap for `[derived]`/`[output]` expressions branching on `MIXNUM`. Both now take the fitted classes from `mixture_posteriors` and evaluate each subject under its own guard; non-mixture callers pass `None`. New Tier-1 tests in `src/api/tests/tests_mixture_postfit.rs` pin IPRED/PRED to the class-2 closed form and the derived column to the class-2 CL. Since this was pre-existing, it fixes FOCE/FOCEI and SAEM mixture output too — changelog entry under **Fixed**.

`cargo test --lib`: 2919 passed, 0 failed.

Issue #528, opened 2026-06-25
IMP importance-sampling diverges on analytical [initial_conditions] models (ESS collapse on baseline subjects)


roninsightrx:
## Summary

A non-IOV analytical model with an `[initial_conditions]` baseline **diverges in the IMP (`$EST METHOD=IMP`) phase** of a `method = [saem, imp]` fit: the SAEM phase converges normally, then IMP's importance weights collapse (`median ESS/K → 0.00`) and the importance-weighted M-step walks θ to the parameter bounds (`OFV → 1e35`, V→upper, CL→lower). The same model fits cleanly in NONMEM (`run14`, SAEM→IMP) and converges in ferx FOCEI.

Found while benchmarking the 6-thioguanine `run14` model (analytical translation, `ferx-testdata/thioguanine_mmc/run14_analytical_exact.ferx`).

## Reproduction

Analytical translation of NONMEM `run14` (1-cpt oral, `init(central) = CONC0 * V`, `log_additive` + `iiv_on_ruv`, `method = [saem, imp]`). Data: 31 subjects, 7 carry a pre-dose baseline (`CONC0 > 0`).

| run | OFV | V | CL | result |
|---|---|---|---|---|
| NONMEM SAEM→IMP | **−249.23** | 270.5 | 110.4 | ✓ |
| ferx FOCEI (analytical) | −242.08 | 288.9 | 114.1 | ✓ |
| ferx SAEM-only | −207.6 | 266.9 | 246 | ✓ stable |
| ferx SAEM→IMP **+init** | **1e35** | →1e6 | →1e-6 | ✗ diverges |
| ferx SAEM→IMP **−init** (drop `[initial_conditions]`) | −233.07 | 255.8 | 143.9 | ✓ converges |

The clean isolation: **drop the `[initial_conditions]` block and SAEM→IMP converges; add it back and IMP diverges.** SAEM-only (no IMP) is stable with init.

IMP trace (with init):
```
iter 1: -2logL(IS) = -232.27  (low-ESS subjects: 4)
iter 2: -2logL(IS) = -172.86  (low-ESS subjects: 16)   <- ESS collapsing
... median ESS/K = 0.00 ...   -> θ pinned to bounds, OFV 1e35
```
No-init trace recovers (`low-ESS subjects: 2 → 5 → 1 → 0`, OFV stabilizes at −244).

## What it is NOT (ruled out)

[...]

PR #992, opened 2026-08-19
feat(estimation): IMP/IMPMAP estimation + objective evaluation for mixtures [#985]

roninsightrx:
> **Stacked on #991** (Bayes) → #987 (SAEM). Base is `worktree-985-bayes-mixture`; review/merge after those. The shared change is the mixture method-gate in `fit.rs`.

## Why

The last of #985's estimator sub-items: **IMP / IMPMAP** for `[mixture]` models — both **estimation** and class-marginal **objective evaluation**. (SAEM #987, Bayes #991, IOV #986.) With this, the full #985 estimator set is complete.

## What changed

Two capabilities:

### 1. Class-marginal objective evaluation (`imp_eval_only`, NONMEM `IMP EONLY=1`)
Per subject and class, run the class MAP inner solve (under a `MixtureClassGuard`), importance-sample the class-conditional marginal `L_ik`, and combine as FOCEI does: `-2 log L = -2 Σ_i log Σ_k p_ik L_ik`. New `run_importance_sampling_mixture`. This is the Monte-Carlo marginal a chain reports (`method = [saem, imp]`, `imp_eval_only = true`).

### 2. Estimating IMP / IMPMAP (class-partitioned MCEM)
`run_mcem` delegates a mixture to new `run_mcem_mixture`. Each iteration importance-samples η **within each class**, forms the responsibilities `PMIX_ik ∝ p_ik L_ik`, then runs responsibility-weighted M-steps:
- mixing coefficients from `PMIX` (reusing SAEM's mixing M-step with `r̄ = PMIX`);
- class-shared Ω `= Σ_i Σ_k PMIX_ik·η̂η̂ᵀ_ik / N`;
- θ/σ via `theta_sigma_weighted_mstep_mixture` — the responsibility- and IS-weighted obs M-step, each class's samples scored under its own `MIXNUM` guard, so class-switched typical values are recovered from their own class.

Final OFV / `MIXEST` / `PMIX` / EBEs come from `mixture_ofv` at the estimates; the covariance step runs on the mixture objective. Ω/σ are class-shared (per-class overrides rejected); IOV / FREM / SDE are rejected (as for the single-population MCEM).

The per-subject/per-class E-step parallelises over subjects with the class guard entered **inside** each worker (subject-parallel, thread-local correct).

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

- [x] **Objective evaluation** vs NONMEM `METHOD=IMP EONLY=1` (`mixture_iv_saem.ctl`, 2nd `$EST`): ferx `-2 log L = 300.82 ± 0.06` vs NONMEM `300.87` — a 0.05-unit match, within the MC SE.
- [x] **Estimation** — IMPMAP recovers the mixture MLE:

```
                ferx IMPMAP    FOCEI MLE (NONMEM mixture_iv.ctl)
TVCL1           1.04           0.98
TVCL2           2.58           2.84
V               9.96           9.98
p(1)            0.47           0.472
```

IMP is the noisiest of the mixture estimators — the class-switched typical value cannot be mu-referenced, so θ is estimated by the weighted M-step alone (the code already warns about this for non-mu-ref models), which converges more slowly than SAEM's MCMC and needs more iterations. For a well-identified point estimate prefer SAEM/FOCEI; use IMP for the marginal likelihood.

## Tests
- [x] Unit (`cargo test --lib`): objective — `mixture_is_marginal_is_finite_and_below_class_worst`, `mixture_is_rejects_iov`; estimation — `impmap_mixture_short_run_populates_posteriors`, `imp_mixture_short_run_ok`, `estimating_imp_rejects_per_class_overrides`; updated `fit_on_mixture_model_rejects_unsupported_method` to `FoceGn`.
- [x] Tier-3 slow (`tests/mixture_nonmem.rs`): `imp_mixture_marginal_matches_nonmem`, `impmap_estimating_mixture_recovers_optimum`.

## Docs
- [x] `docs/estimation/mixture.qmd` — "IMP objective evaluation" + "Estimating IMP / IMPMAP" sections

## Checklist
- [x] `cargo clippy` clean (new code warning-free)
- [x] No `Dual2`/`sens` path touched

## Reviewer hints
- Two entry points: `run_importance_sampling_mixture` (objective) and `run_mcem_mixture` (estimation) + `theta_sigma_weighted_mstep_mixture`. Both mirror `mixture_ofv`'s per-class structure but parallelise over subjects with the guard inside the worker. Since Ω/σ are class-shared, the per-class params are identical — only the `MIXNUM` guard differentiates classes, which is why the E-step builds one `params_iter` and relies on the guard.

## Open questions
- A class-aware mu-reference shift would sharpen IMP's θ estimates (it's the one thing SAEM/IMP lack for class-switched typical values), but the η→θ pairing is class-dependent and not exposed by the parser — a possible follow-up. This completes the #985 estimator set (SAEM, Bayes, IMP objective + estimation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,15 +20,16 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
-- **IMP objective evaluation for mixture models (#985).** Importance sampling now evaluates the
-  class-marginal likelihood `−2 Σ log Σ_k p_ik L_ik` for a `[mixture]` model (`method = imp` with
-  `imp_eval_only`, NONMEM `METHOD=IMP EONLY=1`): per subject and class it runs the class MAP solve and
-  importance-samples the class-conditional marginal, then combines them by log-sum-exp — the true
-  Monte-Carlo marginal for AIC/BIC and NONMEM comparison. The usual use is a chain, e.g.
-  `method = [saem, imp], imp_eval_only = true`. Matches NONMEM to well under a Monte-Carlo SE
-  (`−2 log L = 300.82 ± 0.06` vs NONMEM `300.87` on the two-class anchor). *Estimating* IMP/IMPMAP
-  (parameter updates) for a mixture is not yet class-partitioned and is rejected with a clear pointer
-  to this path; IOV under the mixture IMP objective is likewise rejected.
+- **IMP / IMPMAP estimation and objective evaluation for mixture models (#985).** Importance sampling
+  now both **estimates** a `[mixture]` model (`method = imp` / `impmap`) and **evaluates** its
+  class-marginal likelihood `−2 Σ log Σ_k p_ik L_ik` (`imp_eval_only`, NONMEM `METHOD=IMP EONLY=1`).
+  Estimation runs a class-partitioned MCEM: per subject and class it importance-samples η under the
+  `MIXNUM` guard, forms the responsibilities `PMIX_ik ∝ p_ik L_ik`, and runs responsibility-weighted
+  M-steps for the mixing coefficients, the class-shared Ω, and the class-switched typical values.
+  Objective evaluation combines the per-class IS marginals by log-sum-exp and matches NONMEM to well
+  under a Monte-Carlo SE (`−2 log L = 300.82 ± 0.06` vs NONMEM `300.87` on the two-class anchor).
+  Ω/σ are class-shared (per-class overrides rejected); IOV/FREM/SDE are not supported under the
+  mixture IMP paths.
 - **Bayesian estimation for mixture models (#985).** `[mixture]` models can now be fit with
   `method = bayes`. The latent class is Rao-Blackwellised — marginalised out of every Gibbs block
   rather than sampled — so each subject's likelihood contribution is the K-class marginal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,12 @@ section of the SDLC for the versioning policy).
   Objective evaluation combines the per-class IS marginals by log-sum-exp and matches NONMEM to well
   under a Monte-Carlo SE (`−2 log L = 300.82 ± 0.06` vs NONMEM `300.87` on the two-class anchor).
   Ω/σ are class-shared (per-class overrides rejected); IOV/FREM/SDE are not supported under the
-  mixture IMP paths.
+  mixture IMP paths, and a theta shared between the mixing expression and a structural typical value
+  is rejected (as under SAEM). The per-class draws use the same ISCALE pilot search and `impmap_mceta`
+  multi-start MAP as the single-population MCEM; objective evaluation reports a real per-subject ESS
+  (the worst ESS among the classes a subject loads on), so `imp_low_ess_threshold` and the
+  proposal-collapse warning apply to mixtures. `impmap_trace` and `impmap_auto` / `imp_auto` are not
+  wired for mixtures and now warn instead of being silently ignored.
 - **Bayesian estimation for mixture models (#985).** `[mixture]` models can now be fit with
   `method = bayes`. The latent class is Rao-Blackwellised — marginalised out of every Gibbs block
   rather than sampled — so each subject's likelihood contribution is the K-class marginal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
+- **IMP objective evaluation for mixture models (#985).** Importance sampling now evaluates the
+  class-marginal likelihood `−2 Σ log Σ_k p_ik L_ik` for a `[mixture]` model (`method = imp` with
+  `imp_eval_only`, NONMEM `METHOD=IMP EONLY=1`): per subject and class it runs the class MAP solve and
+  importance-samples the class-conditional marginal, then combines them by log-sum-exp — the true
+  Monte-Carlo marginal for AIC/BIC and NONMEM comparison. The usual use is a chain, e.g.
+  `method = [saem, imp], imp_eval_only = true`. Matches NONMEM to well under a Monte-Carlo SE
+  (`−2 log L = 300.82 ± 0.06` vs NONMEM `300.87` on the two-class anchor). *Estimating* IMP/IMPMAP
+  (parameter updates) for a mixture is not yet class-partitioned and is rejected with a clear pointer
+  to this path; IOV under the mixture IMP objective is likewise rejected.
 - **Bayesian estimation for mixture models (#985).** `[mixture]` models can now be fit with
   `method = bayes`. The latent class is Rao-Blackwellised — marginalised out of every Gibbs block
   rather than sampled — so each subject's likelihood contribution is the K-class marginal
--- a/docs/estimation/mixture.qmd
+++ b/docs/estimation/mixture.qmd
@@ -319,7 +319,11 @@ $$
 
 This is the Monte-Carlo marginal likelihood — the number to compare against
 NONMEM's `#OBJV` for `IMP EONLY` — and it is what a chained fit reports on
-`FitResult.importance_sampling`. The usual workflow is to estimate with SAEM (or
+`FitResult.importance_sampling`. The reported per-subject ESS is the worst ESS
+among the classes that subject actually loads on (classes with
+$\mathrm{PMIX}_{ik}\approx 0$ are excluded, since a poor proposal there barely
+enters the marginal), so `imp_low_ess_threshold` flags mixture subjects the same
+way it does single-population ones. The usual workflow is to estimate with SAEM (or
 FOCE/FOCEI) and append an evaluation stage:
 
 ```toml
@@ -330,7 +334,8 @@ FOCE/FOCEI) and append an evaluation stage:
 
 On `tests/nonmem/mixture_iv.csv` this recovers the NONMEM `IMP EONLY` objective to
 well under a Monte-Carlo standard error: ferx `−2 log L = 300.82 ± 0.06` vs NONMEM
-`300.87` (`mixture_iv_saem.ctl`'s second `$EST`).
+`300.87` (`mixture_iv_saem.ctl`'s second `$EST`), with `ess_min/K = 0.82` and
+`ess_med/K = 0.91`.
 
 ### Estimating IMP / IMPMAP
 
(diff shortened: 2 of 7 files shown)


roninsightrx:
## Code review — IMP/IMPMAP mixtures (#985)

Reviewed `worktree-985-imp-mixture` against `origin/main`. The overall shape (per-class IS with PMIX-weighted marginal, delta-method MC SE, weighted M-step) looks right. Ten findings below; the five under "Medium" are ones I'd want addressed before merge.

### Medium

**1. `mixing_structural_overlap` check is missing — `src/estimation/impmap.rs:1504`**

`SaemMixture::build` is called, but unlike `run_saem` (`saem.rs:1926`) the mixture MCEM never runs `mixing_structural_overlap`. `mstep_mixing`'s own doc says it *assumes* the mixing thetas are dedicated, and that `run_saem` rejects the shared-theta case up front. Here a theta that drives both `logit(1)` and a structural typical value is estimated by `theta_sigma_weighted_mstep_mixture` (line 1668) and then silently overwritten by `mstep_mixing` (1694) — the structural estimate is discarded and the fit is wrong with no error. `docs/estimation/mixture.qmd` also still describes this rejection as SAEM-only.

**2. Per-subject ISCALE pilot search is dropped in the mixture E-step — `src/estimation/impmap.rs:1619`**

`subject_is_draws` is called with a hardcoded `iscale = 1.0`, whereas the single-population MCEM E-step runs `find_optimal_iscale` first (`impmap.rs:~858`). That pilot is what rescues subjects whose posterior Hessian is a poor curvature estimate, and is documented in #528/#961 as the guard against IS weight collapse driving the weighted M-step to the bounds. With it off, a mixture MCEM has a materially higher chance of exactly the collapse-to-bounds failure mode that `objective_converged` was added to catch. (Eval-only pinning 1.0 is intentional; this is the *estimating* path.)

**3. ESS diagnostics are hardcoded as perfect — `src/estimation/importance_sampling.rs:643`**

`run_importance_sampling_mixture` returns `ess_min: 1.0`, `ess_median: 1.0`, `low_ess_subjects: Vec::new()`, and sets each subject's `ess_fraction: 1.0` (line 616); `imp_low_ess_threshold` is explicitly discarded (line 633). So a run in which every per-class proposal collapsed reports ideal sampling: `fit.rs`'s "ESS = 0 (proposal collapse)" warning (`fit.rs:~1366`) can never fire for a mixture, and the fit YAML reports `ess_min = 1.0`. The delta-method variance does inflate `mc_standard_error`, but the user-facing ESS fields actively assert the opposite of the truth. Suggest surfacing the PMIX-weighted min per-class ESS, or a sentinel/`None`.

**4. The mu-ref / thetas-without-ETA warnings never fire for mixtures — `src/estimation/impmap.rs:1379`**

The delegation `if model.mixture.is_some() { return run_mcem_mixture(...) }` sits at `impmap.rs:399`, *before* `run_mcem` builds `mu_ref_pairs` and pushes the "no mu-referenced parameters" and `non_fixed_thetas_without_eta` warnings (`impmap.rs:608`). `run_mcem_mixture` starts an empty `warnings` vec and never adds them. The PR description says "the code already warns about this for non-mu-ref models" — it doesn't for this path, and this is precisely the path where every θ goes through the weighted M-step alone (the PR's own accuracy caveat, TVCL2 2.58 vs 2.84). Users get no warning that θ may be unreliable.

**5. Cancellation is dropped — `src/estimation/impmap.rs:1748`**

The single-population `run_mcem` ends its loop with `if is_cancelled(cancel) { return Err("cancelled by user") }` (`impmap.rs:1080`). The mixture version has the two in-loop `break`s (1531, 1645) but no post-loop check, so a Ctrl-C mid-fit falls straight through to the full K-class `mixture_ofv` pass *and* `run_covariance_step` (the most expensive part of the run) before returning `Ok` with parameters from a truncated MCEM. `fit_inner`'s post-chain check turns it into an `Err` eventually, but only after paying the whole cost; direct callers of `run_impmap` / `run_imp` (tests, the R glue) get a silently bogus `Ok`.

### Low

**6. No per-subject cancel poll in the mixture E-step — `src/estimation/impmap.rs:1573`**

`run_mcem` polls inside its `par_iter` worker (`impmap.rs:752`) precisely because one E-step is minutes on a large dataset. The mixture E-step is `n_classes×` more expensive and polls only before/after, so a cancel set mid-E-step isn't observed until the whole subject × class sweep finishes.

**7. Missing `log_det` finiteness guard — `src/estimation/importance_sampling.rs:509`**

`run_importance_sampling` errors with "IS: Ω log-determinant is not finite" before sampling (`~205`). The mixture entry point checks only that `params.mixture` is `Some` and `n_kappa == 0`; a class Ω with non-finite `log_det` flows into `subject_is_estimate` and yields a silently garbage (NaN/±inf) marginal instead of a clear error.

**8. Mixing thetas are handed to the θ/σ BOBYQA as free coordinates — `src/estimation/impmap.rs:1668`**

`theta_sigma_weighted_mstep_mixture` optimises all `n_theta` entries, but the mixing thetas (e.g. `MIXL`) never enter `obs_nll_subject_into` — SAEM explicitly relies on the θ/σ M-step "leaving them untouched" (`saem.rs:2703`). Here they inflate the BOBYQA dimension (`maxeval = maxiter*(n+1)`), add degenerate directions to the interpolation model, and can leave `theta_cur[j]` perturbed at an arbitrary point in the bounds, from which `mstep_mixing` gets only 4–8 evals to recover. Pinning `lower[j] = upper[j] = log_theta[j]` for `mix.mixing_theta_idx` before the call would be exact and cheaper.

**9. `prev_draws` retained unconditionally, including for IMPMAP — `src/estimation/impmap.rs:1746`**

`run_mcem` gates this on `recenter == ProposalRecenter::SampleMoments` specifically to avoid holding `K·n_subjects` samples on the MAP path. The mixture version keeps `K·n_subjects·n_classes` full sample sets every iteration, yet the MAP path only ever reads `pd[c][i].mean` (line 1580). Storing just the per-class means for `Map` would cut peak memory roughly `K×`.

**10. Stale comment / error text — `src/api/fit.rs:230`**

The comment says "the *estimating* IMP/IMPMAP MCEM rejects mixtures in `run_mcem` with a clear message" — this PR is exactly what makes it estimate them. The reworded error also advertises only "IMP objective evaluation" while `Impmap` is now admitted and estimating IMP/IMPMAP works. Minor, same area: `run_mcem_mixture` silently ignores `mceta` (`impmap.rs:1799`), `collect_trace` (`impmap_trace: None`, 1856) and the `auto` sample ramp, so those user-set options are no-ops for mixtures without a warning.

roninsightrx:
All ten review findings above are fixed in 86498e5f.

| # | Finding | Fix |
|---|---|---|
| 1 | `mixing_structural_overlap` missing | `run_mcem_mixture` now runs the same check `run_saem` does and errors with the equivalent message |
| 2 | ISCALE pilot dropped | `find_optimal_iscale` runs per subject **per class**, using `iscale_min`/`iscale_max` |
| 3 | ESS hardcoded to 1.0 | per-subject ESS = worst ESS among classes with `PMIX > MIX_ESS_PMIX_FLOOR` (1e-3); `ess_min`/`ess_median`/`low_ess_subjects` and `imp_low_ess_threshold` all real, so the `fit.rs` proposal-collapse warning can fire |
| 4 | mu-ref / no-ETA warnings never fired | both pushed from `run_mcem_mixture`, with the mixing thetas excluded from the no-ETA list (`mstep_mixing` owns them) |
| 5 | cancellation dropped | post-loop `is_cancelled` → `Err`, before `mixture_ofv` and `run_covariance_step` |
| 6 | no per-subject cancel poll | added at the top of the E-step worker, returning `SubjectDraws::cancelled` placeholders |
| 7 | missing `log_det` guard | base Ω and every class Ω checked before sampling |
| 8 | mixing thetas free in the θ/σ BOBYQA | pinned via `lower[j] = upper[j] = log_theta[j]` each iteration |
| 9 | `prev_draws` retained on the MAP path | retained only for `SampleMoments`; MAP keeps a new `prev_means` (per-class weighted means only), ~K× less peak memory |
| 10 | stale comment/error text; `mceta`/`collect_trace`/`auto` silent no-ops | text corrected; `impmap_mceta` **wired** for mixtures through a new per-subject `find_ebe_multistart` helper factored out of `run_map_multistart`; `impmap_trace` and `auto` now warn |

### Effect on the anchor

Fixes 2 and 8 change the mixture MCEM's numeric path, in the right direction:

| | TVCL1 | TVCL2 | TVV | p(1) |
|---|---|---|---|---|
| FOCEI optimum | 0.98 | 2.84 | 9.98 | 0.472 |
| IMPMAP before | 1.04 | 2.58 | 9.96 | 0.47 |
| IMPMAP after | **0.998** | 2.597 | **10.005** | 0.465 |

`TVCL2` remains the coordinate the weighted M-step resolves least well — that is the accuracy caveat the new warning now tells users about. The `IMP EONLY` objective is unchanged (the ESS fix is passive): `300.8196 ± 0.0573` vs NONMEM `300.8707`, now reporting `ess_min/K = 0.816`, `ess_med/K = 0.909`.

### Tests

Eight new Tier-1 tests: shared-theta rejection, the two warning sets, cancel → `Err`, `find_ebe_multistart` (no-Cholesky identity with `find_ebe`, and never-worse-than-baseline under restarts), pinned-bound M-step, real-ESS-honours-threshold, and non-finite class log-det. `cargo test --lib` 2932 passed / 0 failed; both slow NONMEM mixture anchors re-run and green; `fmt` and `check --tests` clean.

Docs (`docs/estimation/mixture.qmd`) and the changelog entry updated for the new rejection, the ESS semantics, the wired ISCALE/MCETA, the unsupported `impmap_trace`/`auto`, and the revised anchor numbers.

roninsightrx:
Addressed the Copilot review finding in 8cd3441a.

**`run_importance_sampling_mixture` missing `n_eta == 0` / SDE guards** — valid. The mixture MCEM path (`run_mcem_mixture`) already carried both, but the objective-evaluation entry point did not, so an `n_eta = 0` mixture would return a meaningless marginal and an SDE mixture a silently biased one (the IS obs-NLL path doesn't thread EKF process noise). Both now refuse up front with the same messages `run_importance_sampling` uses.

Regression tests added: `mixture_is_rejects_no_random_effects`, `mixture_is_rejects_sde` (`cargo test --lib mixture_is_` — 6 passed). `cargo fmt` / `cargo clippy` clean on the changed file.

Issue #1000, opened 2026-08-20
bug(trust_region): reports `Converged: YES` when the optimizer terminated on `MaxIters`


TeunP:
## Summary

`optimizer = trust_region` reports `Converged: YES` whenever argmin's executor
returns `Ok(_)` — including when it terminated because it hit `outer_maxiter`.
A fit that simply ran out of iterations is therefore reported to the user as a
converged fit, with standard errors computed at the non-stationary point.

## Where

`src/estimation/trust_region.rs:348`

```rust
let (converged, mut best_x) = match result {
    Ok(res) => {
        ...
        (true, vec)      // <-- unconditional: any Ok(_) is "converged"
    }
    Err(e) => {
        warnings.push(format!("Trust-region did not converge: {}", e));
        (false, x0.clone())
    }
};
```

argmin's `Executor::run()` returns `Ok` for *every* normal termination, so
`TerminationStatus::Terminated(TerminationReason::MaxIters)` lands in the `Ok`
arm alongside a genuine solver-tolerance stop. The `Err` arm only catches an
actual solver *error*, which is why the existing warning never fires on a
budget-exhausted run.

## Reproduce

`examples/one_cpt_iv.ferx` with its two omegas deleted (a fixed-effects-only
model — the shape from #989, which makes this trivially easy to hit because
there is no inner EBE loop to absorb a poor `sigma` start), plus
`optimizer = trust_region`, against the repo's own `data/one_cpt_iv.csv`:

| `maxiter` | reported | OFV |
|---|---|---|
| 5 | `Converged: YES` | 11325.8283 |
| 20 | `Converged: YES` | 11179.1274 |
| 300 | `Converged: YES` | 8504.4349 |
| 2000 | `Converged: YES` | −269.6370 |

The true optimum is −269.6370 (FOCE, FOCEI, Laplace, `gn_hybrid`, and every
NLopt outer optimizer all reach it from the same start; NONMEM `$OMEGA 0 FIX`
anchors it at −269.63700440). The first three rows are off by 8 000–11 000 OFV
units and are all reported as converged. At `maxiter = 300` the run also emits
`%RSE` of 6121% / 9632% on the thetas — the SEs are the only hint anything is
wrong, and they are the kind of thing a user attributes to the model rather than
to the optimizer.

## Scope

Not specific to zero-Omega models — it applies to any `trust_region` fit that
exhausts `outer_maxiter`. The fixed-effects case is just the cheapest
[...]

PR #1002, opened 2026-08-20
feat(parser): allow a model with no random effects (empty Omega) on continuous endpoints [closes #989]

### LINKED ISSUE #989: feat(parser): allow a model with no random effects (empty Ω) on continuous endpoints — the 0×0 path already ships for binary/TTE
## Summary

A model with no random effects is rejected at parse time for any continuous
(residual-error) endpoint:

```
Error parsing model: No omega parameters defined
```

The 0×0 Ω path this needs **already exists and already ships** — it is simply
gated to `[event_model]` / `[binary_model]` endpoints. `examples/binary_logistic.ferx`
declares **zero omegas and zero sigmas**, runs `method = focei`, and is a shipped
example; `examples/ctmm_2state.ferx` is a second zero-omega FOCEI example. So the
estimator handles `n_eta = 0` today. Only continuous endpoints cannot reach it.

## Reproduce

Shipped example, shipped data, `ferx 0.3.0 (core 0.3.0)`, rustc 1.98.0-nightly:

```bash
sed -e '/omega ETA_/d' -e 's/ \* exp(ETA_CL)//' -e 's/ \* exp(ETA_V)//' \
    examples/one_cpt_iv.ferx > one_cpt_iv_noeta.ferx
```

```r
ferx_fit("one_cpt_iv_noeta.ferx", data = "data/one_cpt_iv.csv")
#> Error parsing model: No omega parameters defined
```

## Where

`src/parser/model_parser.rs:1842` gates the empty-Ω branch on the endpoint type:

```rust
let omega = if eta_names_bsv.is_empty() && (has_event_model || has_binary_model_block) {
    // 0×0 Omega — `from_matrix` handles the empty matrix (cholesky of a
    // 0-dim matrix is trivial, log|Ω| = 0); `build_omega_fixed` below
    // already returns an empty `Vec` for an empty eta list.
    OmegaMatrix::from_diagonal(&[], Vec::new())
} else {
    build_omega_matrix(&omegas, &block_omegas, &eta_names_bsv)?
};
```

and `build_omega_matrix` (`src/parser/model_parser.rs:10453`) is the unconditional
`n == 0` rejection. The comment at the gate states the empty-matrix maths is
already handled, and `build_omega_fixed` already returns an empty `Vec`.

## Why it matters

**Modelling.** Naive-pooled fitting, fixed-effects-only models, single-subject
fits, and any reduced model where a variance component has collapsed to the
boundary and should be removed rather than fixed near zero.

**Diagnostics — the case that hit this.** Localising a Dual2-vs-FD objective
discrepancy required removing the inner empirical-Bayes problem entirely, so that
both gradient paths must evaluate the same objective at fixed θ. That is a
one-line model edit and it is currently impossible; the workaround is to scale Ω
toward zero, which shrinks the inner problem without removing it and so cannot
prove the negative.

## Needs evaluation before it is implemented

Filing this as the finding plus open questions rather than a prescription — the
gate is one condition, but the surfaces behind it need checking:

1. **Is σ separable from Ω here?** They are independent: `src/api/run.rs:279`
   already records that "a pure-TTE model (no error model, hence no sigma)"
   is a supported shape, and `binary_logistic.ferx` carries neither. For a
   **continuous** endpoint σ defines the likelihood and must stay required; the
   ask is only that Ω become optional. Worth stating explicitly in whatever
   validation replaces the current check, because "no random effects" and "no
   residual error" are separate capabilities and TTE is the case that needs both.
2. **Does the FOCEI objective reduce correctly at `n_eta = 0`?** It should
   collapse to the plain ML/least-squares objective with no inner optimisation
   and no `log|Ω|` term. `binary_logistic` suggests the path works, but that is a
   non-Gaussian endpoint — the Gaussian residual path has not been exercised this
   way.
3. **What do the η-dependent outputs do when there are no ηs?** Shrinkage, EBE
   tables, `IWRES`/`CWRES` construction, the Ω block of the covariance step, and
   the `.fitrx` round-trip all assume at least one η somewhere. Each needs to
   either produce an empty section or be omitted — not divide by zero.
4. **Which estimators are in scope?** FOCEI is the one demonstrated. SAEM, IMP
   and the Bayes path all sample ηs and may need separate handling or an explicit
   "not applicable without random effects" error.

## Suggested acceptance criteria

- `examples/one_cpt_iv.ferx` with its two omegas deleted fits, and its θ
  estimates match a NONMEM `$EST METHOD=0 POSTHOC` naive-pooled run on the same
  data (or an equivalent reference).
- A continuous zero-Ω model and a TTE zero-Ω/zero-σ model are both covered by
  tests, so the two capabilities cannot regress independently.
- A model that declares an `[error_model]` but no `sigma` still errors, with a
  message that distinguishes it from the Ω case.

TeunP:
## Why

A continuous (residual-error) model was rejected at parse time:

```
Error parsing model: No omega parameters defined
```

The 0×0 Ω path already shipped — but it was gated on the *endpoint type*, not on
anything about the residual likelihood. `examples/pktte_joint.ferx` with its one
omega deleted carries a live `[error_model]`, Gaussian residual rows, and
`n_eta = 0`, and **fits on `main` today** (OFV 74.5193). The same PK model
without the TTE endpoint did not parse. So the Gaussian zero-Ω objective was
already reachable and exercised; only the gate stood between it and a plain PK
model.

Motivating uses: naive-pooled fitting, fixed-effects-only models, single-subject
fits, reduced models where a variance component collapsed to the boundary and
should be removed rather than fixed near zero — and (the case in #989)
localising a Dual2-vs-FD objective discrepancy by removing the inner problem
entirely, which is impossible when you can only scale Ω toward zero.

Closes #989.

## What changed

**Enablement.** `build_omega_matrix` returns the empty matrix for an empty eta
list; the endpoint-type gate and its now-dead `has_event_model` local are gone.
TTE, binary and continuous endpoints now share one code path instead of two
(net −21/+12 lines at the call site).

**Only Ω becomes optional.** `sigma` and `[error_model]` are still required for a
continuous endpoint — "no random effects" and "no residual error" are separate
capabilities, and both existing errors are unchanged and separately pinned.

**Guardrails.** The removed rejection was also the only thing catching a
*forgotten* omega block. Unresolved identifiers fall through to covariate lookup
(documented behaviour), so deleting an `omega` line while leaving `exp(ETA_CL)`
in place silently turned a random effect into a covariate — and
`ferx check model.ferx` reported `ok — no errors` for it. Three new diagnostics:

| Code | Severity | When |
|---|---|---|
| `E_ETA_NOT_DECLARED` | error | Unresolved `ETA…`/`KAPPA…` name, data present and lacking the column. Pushed **ahead of** `E_MISSING_COVARIATE` so `first_error` — what `fit()` reports — names the real fix. |
| `W_ETA_NOT_DECLARED` | warning | Same shape, no `--data`. Warning because a genuine NONMEM-exported `ETA_CL` column cannot be ruled out. |
| `E_SAEM_NO_RANDOM_EFFECTS` | error | `method = saem` at `n_eta = 0`, anywhere in a method chain. |

`E_MISSING_COVARIATE`'s message is byte-for-byte unchanged for the names that
remain, and pinned by a regression test.

**Behaviour table:**

| | before | after |
|---|---|---|
| `ferx check` (omega block deleted, no data) | `ok — no errors` | `warning[W_ETA_NOT_DECLARED]` |
| `ferx check … --data` | `E_MISSING_COVARIATE` | `error[E_ETA_NOT_DECLARED]` |
| `fit()` | "covariate(s) not found in data" | "no `omega` / `kappa` declaration defines them… `omega ETA_CL ~ 0.09`" |
| `ferx check` on a genuinely pooled model | — | clean, 0 warnings |
| `method = saem`, `n_eta = 0` | ran, reported converged, OFV 0.038 high | rejected up front |

## Alternatives considered

**Keeping the gate and adding a `[fit_options]` opt-in.** Rejected — NONMEM
parity is that you simply omit the block, and an opt-in flag would be a second
way to say what deleting the omega lines already says.

**A parse-time error on any `ETA`-prefixed undeclared name.** Rejected — reading
a NONMEM-exported `ETA_CL` column as a genuine covariate is legitimate, and
without a dataset ferx cannot tell the two apart. Hence error-with-data /
warning-without.

**Rejecting SAEM vs. letting it run.** It does run, reports `Converged: YES`,
and lands at −269.599 against FOCEI's −269.637 — the SA gain sequence stops
short because there is no latent variable to average. A quietly-wrong answer is
worse than an error, and `imp`/`impmap`/`bayes` already reject `n_eta = 0`.
Checked in `check_model_options` rather than at run time so `ferx check` catches
it and `methods = [saem, focei]` fails before the SAEM stage burns time.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#290 | open (issue, not PR) | after |

ferx-r#290 is a **follow-up, not a blocker**: `ferx_covariance()` / `ferx_sir()`
`stop()` on an ETA-less fit because their EBE-warm-start guards read "no ETA
columns" as a corrupt fit object. `ferx_fit()` itself is unaffected (its
covariance step runs inside Rust). No `Cargo.lock` bump needed — this PR adds no
`pub` API.

Also verified **not** broken on the R side: `ferx_estimates()`. `ferx_fit.R:1925`
already normalises an empty Ω to `matrix(numeric(0), 0L, 0L)`, so `dim()` is
`c(0,0)` not `NULL` and the `matrix(om, 1L, 1L)` fallback is skipped. Had it been
reachable it would have emitted a phantom `OMEGA(1,1) = NA` row.

## Breaking changes
- [ ] Public Rust API
- [x] **None** — strictly widens what parses. The only behaviour change to a
      previously-working model is `method = saem` at `n_eta = 0`, which was only
      reachable for TTE/binary models and produced an approximate answer.

## Numerical validation

NONMEM 7.6.0, `examples/one_cpt_iv_pooled.ferx` transcribed to ADVAN1 TRANS2
against `data/one_cpt_iv.csv`, committed at
`nonmem_anchor/results/zero_omega_pooled.{ctl,lst,ext,cov}`.

- [x] Compared against: **NONMEM 7.6.0** (`$OMEGA 0 FIX`, `$EST METHOD=0 MAXEVAL=9999 SIGDIGITS=4`)

| Quantity | ferx | NONMEM | Δ |
|---|---|---|---|
| OFV @ NONMEM's optimum | −269.63700436625965 | −269.63700440359776 | **3.7e-8** |
| OFV @ initial params | 11374.7713 | 11374.771336890 | 4e-8 |
| TVCL (converged) | 4.840820 | 4.84070 | 2.5e-5 rel |
| TVV | 52.834133 | 52.8324 | 3.3e-5 rel |
| SIGMA (variance) | 0.0166360 | 0.0166323 | 2.2e-4 rel |
| SE TVCL (`rsr`) | 0.166312 | 0.166298 | 8e-5 rel |
| SE TVV (`rsr`) | 1.760499 | 1.76021 | 1.6e-4 rel |
| SE SIGMA, SD scale (`rsr`) | 0.018243 | 0.0182395 | 2e-4 rel |

Two conventions worth flagging for the reviewer:

1. **`$OMEGA 0 FIX`, not an omitted `$OMEGA`.** Deleting `$OMEGA` does *not* give
   a fixed-effects population fit — NM-TRAN emits `(WARNING 1) NM-TRAN INFERS
   THAT THE DATA ARE SINGLE-SUBJECT`, drops the ID grouping, and on this dataset
   aborts with `TIME DATA ITEM IS LESS THAN PREVIOUS TIME DATA ITEM` at records
   11, 21, …, 91. A degenerate ETA with variance fixed at 0 keeps NONMEM in
   population mode with per-subject dose bookkeeping.
2. **SEs are ferx's under `covariance_method = rsr`.** NONMEM's `$COVARIANCE`
   default is the RSR sandwich; ferx's is `r`. Comparing the two defaults gives
   SEs differing by ~2×, which is the documented convention gap, not a bug.

Independent cross-checks: `Σ IWRES² = 89.999998` on 90 observations (the exact
σ-score stationarity condition `Σ IWRES² = n_obs`), and `OFV = Σ IWRES² + Σ log Var`
reproduces the reported value by hand from the sdtab. NONMEM's *with-constant*
figure (−104.228068) matches the hand computation `+ n·log 2π` (−104.228133).

Estimator sweep on the zero-Ω model — `foce`, `focei`, `laplace`, `gn_hybrid`
and every NLopt outer optimizer (`bobyqa`, `nlopt_lbfgs`, `slsqp`, `mma`) all
reach −269.6370; the ODE engine −269.6379; `gradient = fd` −269.6262.

## Performance
- [x] No significant impact expected — one branch removed from a parse-time path.

## Tests

21 new, **all mutation-verified** (each new-behaviour assertion was confirmed to
fail when the corresponding production change is reverted).

- **Tier 1 (15)** — `src/api/tests/zero_omega_tests.rs`, `src/parser/model_parser_tests.rs`:
  empty-Ω construction (incl. `log_det == 0` and `omega_packed_len(0) == 0`);
  end-to-end parse of a zero-Ω continuous model; σ and `[error_model]` still
  required, with messages that don't mention Ω; kappa-without-BSV builds a 1×1
  Ω_IOV while Ω stays 0×0; the `is_random_effect_shaped` truth table; the
  diagnostic split and its ordering; `E_MISSING_COVARIATE` byte-identity; the
  SAEM guard incl. mid-chain, plus negative controls for `foce`/`focei`/`laplace`.
- **Tier 2 (4)** — `tests/zero_omega_fit.rs`, `outer_maxiter = 3`, no convergence
  loop: the fit returns `Ok` with a 0×0 Ω and empty η-shrinkage; `PRED ≡ IPRED`
  and `CWRES ≡ IWRES` **bit-for-bit** (the cheapest proof the conditional step
  really collapsed rather than carrying a phantom η); SAEM rejected; and a
  `survival`-gated TTE companion with neither Ω nor σ, so the Gaussian and
  survival zero-Ω paths cannot regress independently (#989 AC 2).
- **Tier 3 (2)** — `tests/zero_omega_nonmem_anchor.rs`: eval-at-NONMEM's-optimum
  is **ungated** (evaluation-only, returns immediately) so the anchor registers
  PR coverage; full-convergence comparison is `slow-tests` gated.

`cargo test --lib --features ci`: **2961 passed, 0 failed** (2946 on `main` + 15).

## Example
- [x] `examples/one_cpt_iv_pooled.ferx` — `examples/one_cpt_iv.ferx` with both
  omegas and both `exp(ETA_*)` terms removed. Carries the `$OMEGA 0 FIX` caveat
  and the anchor values in its header.

## Docs
- [x] `docs/model-file/parameters.qmd` — "No omega at all — fixed-effects-only (naive-pooled) fits"
- [x] `docs/model-file/individual-parameters.qmd` — "Forgetting an omega declaration", under Covariate Detection where the fallthrough is documented
- [x] `docs/faq.qmd` — "Fixed-effects-only fits vs NONMEM": the `$OMEGA 0 FIX` correction with NM-TRAN's actual output, the anchor table, and both convention traps
- [x] `docs/estimation/saem.qmd` — why `n_eta = 0` is rejected, with the −269.599 evidence
- [x] `docs/estimation/gauss-newton.qmd` — fixed-effects σ-start sensitivity
- [x] `docs/file-formats/check-report.qmd` — the three new codes (that page promises stable codes to tooling consumers)
- [x] `CHANGELOG.md` `[Unreleased]` → Added

## Checklist
- [x] `cargo clippy --no-default-features --features ci,markov` — no new warnings (my line uses `chain.contains`; the two remaining `iter().any` hits at `validation.rs:1789` predate this PR)
- [x] `cargo check --tests --features ci` clean
- [x] Dual2 sensitivity path untouched — no `sens/` or `*_g<T: PkNum>` code in this diff

## Reviewer hints

**Focus on the guardrails, not the enablement.** The one-line gate removal is
mechanical and the anchor pins it hard. The judgement calls are all in
`src/api/validation.rs`:

- **The `ETA…`/`KAPPA…` prefix heuristic.** Naming-convention-based, which is
  uncomfortable, but there is no reserved prefix in the DSL to key on. The
  narrowing that makes it safe: it only fires for names that are *already*
  errors (unresolved), so a real `ETA_CL` data column resolves and never reaches
  it. `ETHNIC` is in the test fixtures as the deliberate near-miss.
- **Diagnostic ordering.** `E_ETA_NOT_DECLARED` must precede
  `E_MISSING_COVARIATE` because `fit()` surfaces `first_error` only. Pinned by
  test and mutation-verified.
- **Error-with-data vs. warning-without.** Deliberate, to avoid a false positive
  on a NONMEM-exported ETA column.

**Verified unreachable, but worth a second pair of eyes:**
`build_omega_matrix` has a second caller at `model_parser.rs:1929` that builds
the **kappa/IOV** matrix. Relaxing `n == 0` could in principle have turned an IOV
error into a silent `omega_iov: Some(0×0)`, and `omega_iov.is_some()` is what
several consumers read as "IOV is active". It is guarded: the call sits inside
the `else` of `if kappa_info.diagonal.is_empty() && kappa_info.block.is_empty()`,
and both push sites (`10308`, `10392`) append to `kappa_names_ordered` in the
same block as the spec push. A test pins Ω 0×0 alongside Ω_IOV 1×1.

**Skimmable:** the docs, and the anchor artefacts under `nonmem_anchor/results/`.

## Open questions

1. **Anchor tolerance.** The eval test uses `1e-3`, not the repo's usual `0.5`.
   Deliberate: a 1% `TVCL` perturbation moves this OFV by only **0.458**, so the
   conventional band would pass a 1%-wrong fit. The fixed-effects objective is
   flat enough near the optimum that 0.5 is not discriminating here. Reasoning is
   in the test; flagging in case you'd rather stay with house convention.
2. **`optimizer = trust_region` reports `Converged: YES` when it terminated on
   `MaxIters`** — filed separately as #1000, pre-existing and not specific to
   this PR (it affects η models too). It surfaced here because `n_eta = 0` makes
   it trivial to hit: with no inner loop to absorb a poor σ start, TR reports
   converged at OFV 8504 vs the true −269.637. Not fixed here to keep this PR
   scoped; say the word if you'd rather fold it in.
3. **`examples/one_cpt_iv_pooled.ferx` mirroring.** The template asks for new
   examples to be mirrored into ferx-site and ferx-book. Not done — separate
   repos, and I'd rather you decide whether this example earns a site page.


TeunP:
## Review — xhigh, adversarial pass

Every claim below was executed against a local `--features ci` build of `085119d8`, not read off the diff. **2 blockers, 1 high, 2 medium, 6 low.** The core of the change — the parser gate removal and the `n_eta = 0` reduction — holds up; the problems are in CI wiring, patch coverage, and doc claims that the shipped example contradicts.

---

## BLOCKER 1 — CI is red: endpoint-coverage guard fails

`tests/ci_workflow_endpoint_coverage.rs:114`, reproduced locally:

```
these tests/*.rs use a `survival`/`markov` feature cfg but are NOT built by the
`Tests + coverage (TTE/CTMM endpoints)` job in .github/workflows/ci.yml: ["zero_omega_fit"].
```

`tests/zero_omega_fit.rs` carries a `#[cfg(feature = "survival")]` test, so it must be listed in that job. Fix: add `--test zero_omega_fit` to the `cargo llvm-cov` invocation at `.github/workflows/ci.yml:118-125`.

Worth noting *why* this escaped the pre-push check: only `cargo test --lib` and `cargo check --tests` were run. `cargo check --tests` compiles integration tests but never executes them, so a `tests/*.rs` guard test cannot fail locally under it.

## BLOCKER 2 — `codecov/patch` 76.00% vs the 90% gate

18 lines missing, all in `src/api/validation.rs`:

- **The entire `W_ETA_NOT_DECLARED` block** (`validation.rs:2968-2985`). `grep -rn W_ETA_NOT_DECLARED src/ tests/` returns the emit site and the docs table — no test anywhere.
- **`if plain.is_empty() { return diags; }`** (`validation.rs:95`) — the eta-only early return. This is the *actual bug-report shape* (omega deleted, nothing else missing); every test in `zero_omega_tests.rs` mixes in a `WT`, so the most common real path is the untested one.
- The plural arms of `undeclared_random_effect_message` (`"random effects"` / `"them"`).

---

## HIGH — `gn` is documented as applicable at `n_eta = 0`; it returns garbage on the example this PR ships

`docs/model-file/parameters.qmd` states:

> **Which estimators apply.** `foce`, `focei`, `laplace`, `gn` and `gn_hybrid` all reduce correctly and agree to the last printed digit.

All five run on `examples/one_cpt_iv_pooled.ferx` + `data/one_cpt_iv.csv`:

| method | Converged | OFV | TVCL | SE TVCL |
|---|---|---|---|---|
| `foce` / `focei` / `laplace` / `gn_hybrid` | YES | −269.6370 | 4.8408 | — |
| **`gn`** | **NO** | **8670.0458** | **4.1258** | 286.4 (RSE 6942%) |

The claim is false for `gn`. The hedge two paragraphs down names `examples/one_cpt_iv.ferx` "with its omegas removed" — but that file *is* `one_cpt_iv_pooled.ferx`, same `sigma PROP_ERR ~ 0.02 (sd)` start. So the failing case is the flagship example, and the doc's own escape hatch points at it under a different name.

The deeper issue is an inconsistent bar. SAEM is **hard-rejected** for landing 0.04 OFV units short (−269.5986 vs −269.6370, per the rationale comment at `validation.rs:1762`). `gn` silently returns a result **8940 OFV units** off with 6942% RSE and no diagnostic beyond `Converged: NO`. Either drop `gn` from the applicable list and emit a warning at `n_eta = 0`, or at minimum correct the sentence.

## MEDIUM — the new guard covers SAEM only; `imp` / `impmap` / `bayes` have identical semantics

All three already reject `n_eta = 0` at run time — `importance_sampling.rs:113`, `importance_sampling.rs:532`, `bayes.rs:452` — for exactly the reason the PR gives for SAEM. But only SAEM got a check-time diagnostic:

```
$ ferx check p_imp.ferx --json      # method = imp, n_eta = 0
{ "valid": true, "model": "p_imp", "diagnostics": [] }
```

The PR's own justification — *"check it here so `ferx check` catches it and a `methods = [saem, focei]` chain fails up front rather than mid-run"* — applies verbatim to `methods = [focei, imp]`, which today runs the whole FOCEI stage before dying.

`docs/estimation/saem.qmd` compounds it:

> The check runs in the same validation pass as `ferx check` … `imp`, `impmap` and `bayes` reject `n_eta = 0` for the same reason.

Adjacent sentences, and a reader will take the second to inherit the first. It doesn't.

## MEDIUM — IOV-only (`n_eta = 0`, `n_kappa > 0`) is newly enabled with a parse test only

`test_parse_no_bsv_omega_with_kappa_is_iov_only` pins that it parses, and its own comment says the shape "was also blocked by the old rejection". So this is a **new model class** with numeric output and no fit-level test and no oracle.

It does work — I ran it end to end on `data/warfarin_iov.csv`: converges in 82 iterations, `KAPPA_CL = 0.634427 (CV% 79.7, SE 0.208107)`, per-occasion kappa shrinkage computed, covariance clean, and no false `W_ETA_NOT_DECLARED` on the declared kappa.

And the degenerate oracle passes. Force every row to a single occasion; an IOV-only model is then mathematically identical to a BSV-only model with the same variance:

| model | OFV (`maxiter = 0`, same start) |
|---|---|
| `omega ETA_CL ~ 0.09`, no kappa | 552.9797 |
| `kappa KAPPA_CL ~ 0.09`, `iov_column = OCC` (all 1) | 552.9797 |

That is ~20 lines as a Tier-2 test and it belongs in this PR — CLAUDE.md requires a test for a new capability, and for a family with no NONMEM equivalent the degenerate oracle is the sanctioned substitute.

---

## LOW

1. **`docs/model-file/parameters.qmd`: "No `OMEGA` section is printed" is false.** `src/io/output.rs:176` prints `--- OMEGA Estimates ---` unconditionally, so a zero-Ω fit shows the header with an empty body. The sibling `--- OMEGA ---` writer at `output.rs:645` *is* guarded on `n_eta > 0`. Guard `:176` the same way and the doc becomes true.

2. **`docs/estimation/saem.qmd` broken sentence:** *"Run anyway, it terminates on its own criterion slightly short of the optimum"*.

3. **`docs/model-file/individual-parameters.qmd` quotes the error non-verbatim.** Real CLI output is `error[E_ETA_NOT_DECLARED] parameters:1: …` and ends with "A model with no random effects at all is valid (#989): to make it fixed-effects-only, drop the `exp(ETA_CL)` term as well as the omega line." The doc silently drops both the `:1` and the trailing sentence, with no ellipsis.

4. **`src/api/adaptive.rs:1106` holds a second, independent copy** of the `"covariate(s) not found in data (case-sensitive)"` message that did not get the eta split. `adaptive.rs:1085` calls `check_covariates` first, so this is the narrower `[adaptive_dosing]`-declared-covariate check — but the two now disagree on how a forgotten omega reads.

5. **`is_random_effect_shaped` prefix heuristic and `KAPPA`.** `KAPPA` is a genuine covariate name in myeloma/oncology PK (serum free light chains). A model using a `KAPPA` covariate absent from the data now reads "declare a random effect", and `ferx check` without `--data` warns on every run. The message text does carry the escape hatch ("or — if this is meant to be a data column — add that column to the data file"), so it degrades gracefully; recording it rather than blocking on it. No repo dataset has an `ETA*`/`KAPPA*` column, so nothing in-tree regresses.

6. **YAML/JSON asymmetry.** `{model}-fit.yaml` emits a bare `omega:` (parses as null) where it was previously a map; `{model}-fit.json` is clean (`{"rows":0,"cols":0,"data":[]}`).

---

## Verified good — things I tried to break and could not

- **The `rsr` anchor reproduces the doc table exactly.** `covariance_method = rsr` → SE TVCL **0.166312**, SE TVV **1.760499** vs NONMEM `1.66298E-01` / `1.76021E+00`. The `r`-vs-`rsr` factor-of-two note in `faq.qmd` is correct and the default-`r` SEs are 0.077597 / 0.943539 as that note implies.
- Full fit lands −269.6370 / TVCL 4.840820 / TVV 52.834133; covariance step computes at `n_eta = 0` without panicking; sdtab carries no ETA columns; `shrinkage_eta` empty.
- `--simulate` on a fixed-effects-only model works (simulate + refit round-trips).
- The guardrail fires correctly on all three surfaces: no data → `W_ETA_NOT_DECLARED` (`valid: true`); `--data` → `E_ETA_NOT_DECLARED` (`valid: false`); `fit()` → same text, and it genuinely is the `first_error`.
- **The second `build_omega_matrix` caller is genuinely unreachable with an empty list.** `model_parser.rs:1929` sits inside the `else` of `if kappa_info.diagonal.is_empty() && kappa_info.block.is_empty()`, and both push sites append to `kappa_names_ordered` in the same block as the spec push. The reviewer hint in the PR body is accurate.
- **No divide-by-zero at `n_eta = 0`.** `saem.rs:1651` and `saem_conddist.rs:88` both compute `(n_mh_steps / n_eta).max(2)` but are gated `if n_eta >= 2`.
- `check_model_options` is invoked once with the full `options` at `fit.rs:948`, *before* the `for (stage_idx, &method) in chain.iter()` loop at `fit.rs:1198` — the "fails up front on a chain" claim holds.
- `has_binary_model_block` survives the removal of `has_event_model` (still consumed by `is_endpoint_only`, `model_parser.rs:1117`), and `is_endpoint_only` is unaffected by the change.

---

Blockers 1 and 2 must land before merge. Recommend folding in the `gn` doc correction, the IOV-only degenerate-oracle test, and the `output.rs:176` header guard while the branch is open; the `imp`/`impmap`/`bayes` check-time guard is a clean follow-up rather than scope for this PR.

TeunP:
## Review findings addressed — `128e6ef0`

Both blockers fixed, plus the HIGH and the two MEDIUMs. What changed:

**Blocker 1 — endpoint-coverage guard.** `--test zero_omega_fit` added to the `Tests + coverage (TTE/CTMM endpoints)` job. `cargo test --test ci_workflow_endpoint_coverage` now passes locally.

**Blocker 2 — patch coverage.** Four new validation tests cover the three uncovered paths: the whole `W_ETA_NOT_DECLARED` block in `validate_model_file`, the `data_path.is_some()` side (which must *not* double-report the finding as warning **and** error), the `plain.is_empty()` early return in `check_covariates` — the eta-only case, which is the actual shape of the #989 report and the one path every existing test happened to avoid — and the plural arms of `undeclared_random_effect_message`. All four mutation-verified: flipping `data_path.is_none()`, replacing the `plain.is_empty()` guard with `if false`, and singularising the plural arm each fail exactly the intended test and nothing else (11 passed / 4 failed, the right 4).

**HIGH — `gn`.** `parameters.qmd` and `gauss-newton.qmd` corrected. `gn` is now documented as **not recommended** at `n_eta = 0`, with the measured numbers (stops at 8670, `Converged: NO`, ~6900% RSE, vs −269.6370 for the other four). `gauss-newton.qmd` also named the wrong example file and omitted the terminal OFV; both fixed, and the `0.13 (sd)` start is recorded as reaching −269.6322 — right basin, still 0.005 short of FOCEI. Left the estimator itself alone: adding a `W_GN_NO_RANDOM_EFFECTS` warning is a behaviour change worth its own PR.

**MEDIUM — `imp` / `impmap` / `bayes`.** Documented accurately in `saem.qmd` rather than fixed in code: they refuse `n_eta = 0` at run time, `ferx check` reports such a model as valid, and a `methods = [focei, imp]` chain still runs its FOCEI stage first. Extending the check-time guard to them means a new diagnostic code, so it belongs in a follow-up rather than here.

**MEDIUM — IOV without BSV.** `iov_without_bsv_reduces_to_bsv_on_a_single_occasion` added as a Tier-2 degenerate oracle.

> **Correction to my review above:** I wrote that the two objectives agree "bit-for-bit, not to a tolerance." That is wrong, and it cannot be true. The IOV marginal builds its kappa columns of `H` by central differences (`foce_subject_nll_iov`, `EPS = 1e-6`) where the BSV path takes its eta columns from the shared Jacobian, so the two `log|H̃|` terms differ in the last few ULP of the stencil. Measured delta **1.3e-7** on ~553. The test asserts `< 1e-6` and says why.
>
> The band still makes it an oracle rather than a smoke test: perturbing the IOV prior to 0.10 moves the delta to **8.4e-2**, five orders of magnitude above the band — so a silently-dropped kappa penalty (worth ~24 OFV units here) cannot pass. Mutation-verified.

**LOW 1 — empty `OMEGA` section.** `output.rs:176` now guarded on `n_eta > 0`, mirroring the YAML writer's `--- OMEGA ---`. Verified three ways: suppressed at `n_eta = 0`, unchanged for a model with etas, and the `KAPPA (IOV)` section still prints for an IOV-only model.

**LOW 2, 3** — `saem.qmd` sentence and the truncated `E_ETA_NOT_DECLARED` quote in `individual-parameters.qmd` both fixed.

### Not addressed, deliberately

- **LOW 4** — the second copy of the covariate message at `adaptive.rs:1106` is untouched. It is the `[adaptive_dosing]`-declared-covariate check, a different call site, and giving it the eta split is scope creep here.
- **LOW 5** — the `KAPPA`-as-a-real-covariate false positive (myeloma free light chains) stands. The message already carries the escape hatch, and narrowing the heuristic would cost the guardrail more than the false positive costs.
- **LOW 6** — the YAML/JSON asymmetry (`omega:` → null vs `{"rows":0,...}`) stands; JSON is the programmatic surface and it is correct.

Local: `cargo test --lib --features ci` **2966 passed / 0 failed**; the three zero-omega binaries and the endpoint guard pass; `cargo fmt --check` clean; no new clippy findings.

Issue #995, opened 2026-08-19
Reserved names live in seven scattered sets, three of them function-local: export them per scope instead of making consumers guess


TeunP:
ferx reserves names in at least six places. Three of those sets are **function-local Rust consts**, none is exported, one is not a set at all but a regex-shaped predicate, and the same name carries different status in different sets. Consumers therefore mirror them by hand, and the mirror drifts silently.

This is the generalisation of #993, #994 and FeRx-NLME/translator#17. Filing separately so those stay small and closable -- each is a concrete bug with a concrete fix, and each is also an independent instance of this.

## The inventory

| set | location | scope | form |
|---|---|---|---|
| `RESERVED_PK_SLOTS` | `types.rs` | `[individual_parameters]` | `F`, `LAGTIME` |
| `DoseAttr::from_indexed_name` | `types.rs` | `[individual_parameters]` | **a predicate, not a list**: `^(f\|d\|r\|alag\|lagtime)[0-9]+$` |
| `RESERVED_ODE_NAMES` | `model_parser.rs:9061` | `[odes]` RHS | `TIME`, `T`, `TAFD`, `TAD`, `MACHEPS` |
| `BUILTIN_SPECIALS` | `model_parser.rs:3517` | expression parse | the above plus `IPRED`, `PRED`, `DV` |
| `RUV_FORBIDDEN_NAMES` | `model_parser.rs` | residual-magnitude expr | `IPRED`, `PRED`, `DV`, `TAD`, `TAFD`, `IRES`, `IWRES`, `CWRES` -- `TIME` explicitly **allowed** |
| `DATA_BLOCK_ROLES` | `model_parser.rs` | `[data]` | `id`, `time`, `dv`, `evid`, `amt`, `cmt`, `rate`, `mdv`, `ii`, `ss`, `cens`, `addl`, `tentry`, `fremtype` |

Plus the `init()` scope, which is a seventh and is not a named constant at all -- it is whatever `collect_undefined_vars` happens not to flag, which is exactly the bug in #994.

`RESERVED_ODE_NAMES` and `BUILTIN_SPECIALS` are declared **inside function bodies**. They cannot be referenced by anything, including other parts of ferx-core.

## `TIME` alone has four different statuses

- `[odes]` RHS: reserved and meaningful.
- `[odes] init()`: accepted, and always exactly zero -- #994.
- residual-magnitude expressions: explicitly allowed, and called out in a comment as "the one allowed built-in".
[...]

Issue #767, opened 2026-07-09
Parser diagnostics: collect + order errors instead of first-error-wins (F{n}/ALAG{n} reject pre-empts #308/#309)


TeunP:
## Problem

`parse_full_model` is **first-error-wins** — it returns `Result<_, String>` and bails on the first problem it hits. Because validation passes run in a fixed code order rather than a fundamentality order, a lower-level diagnostic can be masked by a higher-level one raised earlier in the function.

Concrete case surfaced in the #725 review (finding #4): the analytical `F{n}`/`ALAG{n}` reserved-name reject runs inside the analytical dose-attribute loop (`src/parser/model_parser.rs` ~L1555), which executes **before** `build_pk_param_fn` (undefined-reference guard, #308) and the required-PK-param check (#309). So a model with **both** an undefined covariate **and** an unmapped `F1`:

```
[individual_parameters]
  CL = TVCL * exp(ETA_CL) * UNDEFINED_COV   # the real, fundamental problem
  F1 = TVF1                                  # unmapped reserved name
[structural_model]
  pk one_cpt_oral(cl=CL, v=V, ka=KA)
```

reports the `F1` error first and hides the undefined `UNDEFINED_COV` — the user fixes `F1`, re-runs, and only *then* sees the more fundamental error. Two round-trips for one broken model.

This is **not a correctness bug** — every one of these is genuinely an error and the messages are individually correct/actionable. It's a diagnostic-quality/ordering gap.

## Options

**A — minimal reorder.** Move the `F{n}`/`ALAG{n}` reserved-name rejection to run *after* the undefined-reference / required-param validation, so the most fundamental error reports first. Caveat: the reject is currently entangled with the `D{cmt}`/`R{cmt}` routing loop, which *must* run before `build_pk_param_fn` (it populates `analytical_modeled_slots`). So this means splitting the F/lag rejection into a separate pass after `build_pk_param_fn`, not just moving a line.

[...]

Issue #710, opened 2026-07-07
[individual_parameters] silently reads a forward-referenced name as 0 instead of erroring


roninsightrx:
## Summary

`[individual_parameters]` statements are evaluated in file order with a plain running variable map. A statement that references a name declared *later* in the same block doesn't error — it silently resolves the not-yet-defined name to `0.0` (the same "undefined variable reads 0.0" bytecode fallback used elsewhere), producing a wrong-but-plausible-looking model with no warning from `ferx check` or at fit time.

## Reproduction

Translating `ferx-testdata/pembrolizumab_radboudumc/ferx/model066.ferx` (a time-varying-clearance ODE model, see #708) to an equivalent closed-form 2-cpt model, with:

```
[individual_parameters]
  CL   = TVCL * (LBWT / 52.5)^THETA_LEANCL * exp(IMAX*TIME^HILL/(TI50^HILL+TIME^HILL)) * exp(ETA_CL)
  V1   = TVV1 * (LBWT / 52.5)^THETA_LEANV * exp(ETA_V1)
  Q    = TVQ  * (LBWT / 52.5)^THETA_LEANCL
  V2   = TVV2 * (LBWT / 52.5)^THETA_LEANV * exp(ETA_V2)
  IMAX = TVIMAX
  TI50 = TVTI50 * exp(ETA_TI50)
  HILL = TVHILL
```

`CL` references `IMAX`, `TI50`, `HILL` — all declared *after* it. Verified with the fit itself: running with `TVIMAX = -0.3` and `TVIMAX = -5.0` (all else equal, `maxiter = 0`) gives **bit-identical OFV** (17200.8264 both times) — `IMAX` (and likewise `TI50`/`HILL`) has zero effect on predictions, because `CL`'s formula reads them as `0` regardless of their actual theta/omega value, collapsing `exp(IMAX*TIME^HILL/(TI50^HILL+TIME^HILL))` to `exp(0) = 1` always.

Reordering the block so `IMAX`/`TI50`/`HILL` are declared before `CL` fixes it — OFV then correctly differs between the two `TVIMAX` values (16581.58 vs 15771.92).

`ferx check model066.ferx --data ...` reports `ok: no errors (0 warning(s))` on the broken (original order) file — no diagnostic at all.

## Expected behavior

[...]

Issue #315, opened 2026-06-13
Parser: ODE individual parameter declared but never used is silently dead (no warning) — #309 parallel


TeunP:
## Summary

#310 added a "computed but never used" warning for individual parameters, but it is scoped to **analytical** models (`if !pk_param_map.is_empty()` — `pk_param_map` is empty for ODE). An ODE `[individual_parameters]` entry that is declared but never referenced in the `[odes]` RHS / `[scaling]` / `[derived]` / `[output]` (and is not `F`/`lagtime`, which auto-route) is routed to a free slot by `ode_param_slots`, computed every evaluation, but **never read** — silently inert, with no warning.

`check_unused_parameters` only inspects thetas/etas/sigmas, and the theta/eta feeding the dead parameter counts as "used" the moment it appears in the `[individual_parameters]` declaration — so neither the parameter nor its feeding random effect is reported.

Surfaced in the self-review of #310.

## Reproduction

```
[parameters]
  theta TVKE(0.5)
  omega ETA_KE ~ 0.09
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  KE = TVKE * exp(ETA_KE)   # never referenced in [odes]
[odes]
  d/dt(central) = -(CL/V) * central
```

`KE` and `ETA_KE` are dead; the `ETA_KE` omega element is unidentifiable. No diagnostic. The analytically-equivalent mistake (oral model declaring `F` but forgetting `f=F`) **is** flagged by #310.

## Proposed fix

Extend the computed-but-unused census (`src/parser/model_parser.rs`, the `if !pk_param_map.is_empty()` block) to ODE models, **excluding** engine-applied slots — i.e. parameters that `ode_param_slots` routes to `PK_IDX_F`/`PK_IDX_LAGTIME` (named `f`/`lagtime`/`alag`), which are applied to the dose without appearing in the RHS. The census already tokenizes `[odes]` (it is an unnamed block), so a parameter referenced in the RHS correctly counts as used.

## Acceptance

- An ODE param declared but never referenced (and not `f`/`lagtime`) → "computed but never used" warning.
- `F`/`lagtime` declared (and engine-applied) on an ODE model → not flagged.
- All existing ODE `examples/*.ferx` parse with no new warnings.
- Regression test.

## Relation

ODE counterpart of the #309 "computed but never used" warning added in #310.

PR #1003, opened 2026-08-20
fix(parser): reject a dose attribute that is applied by the engine and read by the model [closes #993]

### LINKED ISSUE #993: Diagnose a bare `F` that is both assigned and read: it is applied as bioavailability as well, silently
> **Answered, and now a request rather than a question.** It *is* applied: a model
> that assigns `F` and also reads it double-counts, silently, and renaming the
> parameter changes every prediction by exactly its own value. Measured in
> [this comment](#issuecomment-5345822443). What remains open is the ask below --
> should the engine diagnose it.

## The original question (answered: yes)

A model that assigns `F` in `[individual_parameters]` and uses it as an ordinary parameter validates clean, with no diagnostic. `F` is the reserved bioavailability slot. **Is that value being applied as dose bioavailability at solve time?**

It is. The model below silently computes something other than what it reads.

## Reproduction

Measured against ferx 0.3.0 (`ferx-r@7889719`).

```
[parameters]
  theta TVCL(2.0, 0.0, 1e15)
  theta TVV(10.0, 0.0, 1e15)
  omega ETA_CL ~ 0.09
  sigma EPS1 ~ 0.1 (sd)

[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  F  = CL/V          # intended as an elimination rate constant, not bioavailability

[structural_model]
  ode(obs_cmt=CENT, states=[CENT])

[odes]
  d/dt(CENT) = -F * CENT

[error_model]
  DV ~ proportional(EPS1)
```

```
$ ferx_model_validate("ftest.ferx")
Result: VALID
```

No warning, no note. Renaming `F` to `KEL` changes nothing about validation, so validation cannot distinguish the two intents.

**ANSWERED** -- see [the measurement below](#issuecomment-5345822443). The solver does feed it into `PK_IDX_F`: renaming that parameter from `F` to `KE` changes every prediction by exactly `CL/V`, and both spellings validate with zero errors and zero warnings. The question this issue now asks is only whether that should be diagnosed.

## Why the name matters more than a typical reservation

`F` is a single letter and one of the most common variable names in pharmacometrics. In NONMEM it is the **model prediction** -- `Y = F * (1 + EPS(1))` is the standard proportional error line, and `IPRED = F` the standard prediction assignment. In ferx the same spelling in `[individual_parameters]` means **bioavailability**. Two different concepts, one letter, and no diagnostic separating them.

Concretely, in the translator's bundled NONMEM corpus **10 of 12 models reference `F`**, and `pk_1cmt_oral.mod` contains both meanings at once. `nonmem2rx` hands us:

```r
rxf.rxddta1. <- 1
f(ABS) <- rxf.rxddta1.     # NONMEM F1 -- bioavailability
f <- CENTRAL/scale2        # NONMEM F  -- the prediction
y <- f * (1 + eps1)
```

`LAGTIME` has the same structure but is a far less likely collision, so this is really about `F`.

## What the engine already does elsewhere

ferx-core rejects a name using a reserved *prefix* with a clear, actionable message (`model_parser.rs:1325-1335`):

> parameter name `{bad}` uses the reserved `{prefix}` prefix, which ferx reserves for {reserved_for} (#486). Rename it.

and the source comments around `model_parser.rs:1222` describe the risk explicitly:

> silently hijacking the reserved F/lagtime slots, aliasing the CL slot in pk_indices, or exhausting the ...

So the failure mode is understood in the code; it just does not appear to be diagnosed for the bare `F`/`LAGTIME` names the way it is for reserved prefixes.

## Possible remedies, in preference order

1. **Diagnose the ambiguity.** If a model assigns `F` (or `LAGTIME`) and *also* references it from `[odes]` or `[error_model]`, that is almost certainly not bioavailability -- warn, naming both readings. This is the cheapest change and catches the realistic case.
2. **Make the intent explicit.** A distinct spelling for the reserved slot (`BIOAV`, or a `[dosing]` block) leaves `F` free for ordinary use. Breaking, so only worth it if (1) proves insufficient.
3. **Document and close.** If `F` is only consumed when the structural model actually has a dosing target and is otherwise inert, say so in the format docs and this becomes a documentation fix.

The answer determines how `ferxtranslate` translates NONMEM `F1`: see FeRx-NLME/translator#16, which this no longer blocks. The `init()` scope rule has the same shape -- enforced here, stated nowhere authoritative, learnable only by probing: FeRx-NLME/ferx-core#994. The compartment-indexed sibling of this problem -- an ordinary parameter named `F1`/`D1`/`R1` becoming a dose attribute -- is FeRx-NLME/translator#17, demonstrated end to end there.

## Not urgent

No shipped model is known to be affected. The translator does not currently emit `F` at all -- the NONMEM prediction is inlined before it reaches `[individual_parameters]`, and `F1` is not yet translated. This is being filed before that work starts, not after it broke something.

TeunP:
## Why

Closes #993.

`F` is the reserved bioavailability name. A model that declares it in `[individual_parameters]` **and** reads it back on the prediction path gets it applied twice — once by the engine at the dose event, once where it is read — and nothing said so. Both spellings of the same model validated with zero errors and zero warnings while differing by exactly `CL/V`:

```
 t    named F      named KE     ratio
 0.5   9.512294    95.122942   0.100000
24.0   0.907249     9.072490   0.100000
```

Rather than reason from the source, I swept every candidate name against the engine — each declared as a plain rate constant read in the `[odes]` RHS, control being the same model renamed to `ZQ`:

```
     name   PRED(t=0.5)  vs control   verdict
       KA     95.122942    1.000000   inert
    Q/V2/Q3/V3            1.000000   inert
    N/MTT/MAT/CV2         1.000000   inert
       S1/S2              1.000000   inert
      KEL     95.122942    1.000000   inert
        F      9.512294    0.100000   *** IMPLICITLY CONSUMED ***
       F1      9.512294    0.100000   *** IMPLICITLY CONSUMED ***
  LAGTIME     96.078944    1.010050   *** IMPLICITLY CONSUMED ***
     ALAG     96.078944    1.010050   *** IMPLICITLY CONSUMED ***
    ALAG1     96.078944    1.010050   *** IMPLICITLY CONSUMED ***
 LAGTIME1     96.078944    1.010050   *** IMPLICITLY CONSUMED ***
       D1     95.122942    1.000000   inert
       R1     95.122942    1.000000   inert
```

`0.100000` is exactly `F`; `1.010050` is exactly `exp(LAGTIME·ke)`. So the defect is not one name but two families — bioavailability and dose lag, bare and compartment-indexed. `N`/`MTT`/`MAT`/`CV2` hold canonical PK slots but are only read through an explicit `transit(n=N, …)` / `igd(…)` arg mapping; `S{n}` is not routed at all.

`docs/model-file/ode-models.qmd` has said "do not multiply by `F` anywhere in the right-hand side" since the dose-entry migration, and carries a note that **earlier ferx versions required** `F` in the flux. Those models compute `F²` today and nothing catches them. That is the case for an error over a warning: the prose has been in place the whole time and did not enforce anything.

## Anchored on NONMEM 7.6.0

Everything above is ferx measuring ferx — a model reading its own `F` against the same model renamed. Self-consistent, but both sides run on the same engine. Two `ADVAN13` control streams differing in **exactly one `$DES` line**, `$THETA` all `FIX`, `MAXEVAL=0`, `F1 = 0.5`, close that gap:

| | `$DES` line for the central compartment |
|---|---|
| **A** — modern | `DADT(2) =  KA*A(1) - K*A(2)` |
| **B** — legacy, the convention the ODE docs' migration note describes | `DADT(2) =  F1*KA*A(1) - K*A(2)` |

```
   t   NONMEM-A     NONMEM-B     B/A
 0.5   0.513070     0.256530     0.499990
 1.0   0.730400     0.365200     0.500000
 2.0   0.823870     0.411930     0.499994
 4.0   0.715540     0.357770     0.500000
 8.0   0.481420     0.240710     0.500000
12.0   0.322710     0.161350     0.499985
24.0   0.097198     0.048599     0.500000
```

`B/A = 0.5 = F1` at every point (max deviation `1.5e-5`, the 5-significant-figure precision of NONMEM's `$TABLE`), so the legacy form really does apply `F` twice. **And NONMEM reports nothing** — run B completes and prints an objective function value, `646.654` against A's `656.358`. ferx being stricter here is a deliberate divergence, not a shared behaviour, which is what the migration note and the `ferxtranslate` guidance rest on.

The ferx equivalent of **A** reproduces NONMEM-A to `8.5e-6` relative — below the reference's own table precision — and the ferx equivalent of **B** does not parse. Both halves are pinned in `tests/dose_attr_double_use_nonmem_anchor.rs`; control streams, data, `.lst` and `.tab` outputs are git-tracked under `nonmem_anchor/`.

## What changed

A dose attribute is rejected when the engine applies it **and** the model reads it on the prediction path.

- **`src/types.rs`** — `DoseAttrConsumption::of(name, slot)` classifies a name into `{attribute, EveryDose | CodedRateOnly, compartment}`. Exported and documented rather than another function-local const, so #995's generator can read it; carries the scope and consequence that issue asks for per name. `DoseAttr::noun()` / `applied_as()` sit beside it so a message can't name one attribute and describe another.
- **`[odes]` rejection in `build_ode_spec`** — walks the resolved `stmts_owned` AST, next to the existing covariate-in-RHS rejection and the case-collision guards. AST-level rather than block text: comments are already stripped by `extract_blocks`, `d/dt(X)` and `Assign` left-hand sides are not reads, and `ode_template`-generated equations plus the injected joint-PK-TTE `d/dt(__chz_n)` hazard lines are all in the tree by that point.
- **`[scaling]` rejection** — same walk over both `y` and `obs_scale`, after `parse_scaling_block` so a malformed block reports its own syntax error first.
- **`[odes] init(...)` rejection** — `init(state) = <expr>` directives are split out of the block *before* the derivative statements are parsed, so they need their own walk. Without one the rule had a hole the exact size of its own remedy: a user told to drop `F` from the RHS could move it into the initial condition and get the same silent double application, from the block the error names.
- **`[adaptive_dosing] observe` rejection** — `observe` is compiled by the same `build_y_output_fn` as a Form-C `y` readout, so it reaches individual parameters. It is also the highest-stakes of the four: `observe` is what the `when` rules compare against, so a double-applied `F` biases the **titration decision** and every dose the controller then emits, not just a reported number.
- **Two absorption-ODE-twin declines** — the twin re-emits the model's blocks into a source that *is* an ODE model, so an ODE-scoped check can reject a twin whose analytical primary was accepted, and `get_or_build`'s `.expect()` turns that into a mid-fit panic. `[adaptive_dosing]` is now dropped from the twin, and a `f=`/`lagtime=` parameter that also fills a disposition role declines it — both matching the function's standing "decline rather than panic at eval time" policy. Filed the general hazard as #1008.
- **Shared collectors** — `collect_indiv_param_reads` / `record_coded_rate_reads` replace three near-identical inline loops. The boundary that matters is "an expression compiled against the individual parameters", not a block heading; the duplicated collector is what made two of the four paths easy to miss.
- **`D{n}`/`R{n}` in `check_modeled_dose_rates`** — see below.
- **`E_DOSE_ATTR_DOUBLE_USE`** in `parse_error_to_diagnostic`, so `ferxtranslate` / ferx-r can act on it rather than parsing prose.

### Scope is wider than the issue title, in one direction and narrower in two

**Wider — `D{n}`/`R{n}`, but on a different gate.** They are inert on ordinary data and consulted only under coded `RATE`. Re-running the sweep against `RATE=-1` data, every model *except* `R1` hard-errors with the existing `E_MODELED_RATE_NO_PARAM`; the hole is only `R{n}` present **and** read. Measured driving the infusion while simultaneously being read in the RHS, matching the analytic zero-order solution to 6 s.f.:

```
   t=  0.50   R1=20:  9.754115    R1=50: 24.385288
   expected           9.754115           24.385288
```

This cannot be a parse-time error — `R1` = "response 1" is an ordinary PD name and such a model is correct whenever the data never codes `RATE`, which `types.rs:479-488` already states as deliberate policy. So it is checked against the dataset, beside `E_MODELED_RATE_NO_PARAM`. The stale comment there claiming the collision is "a documented reserved-name note in `docs/`, not a runtime check" is updated.

**Narrower — two of the four blocks proposed on the issue are wrong.** Both measured:

- `[error_model]` referencing `F` **already hard-errors**: residual-magnitude expressions never see individual parameters (`build_ruv_magnitude` takes thetas/etas/sigmas/covariates), so `F` resolves as a covariate and the existing undeclared-covariate check catches it. Nothing added.
- `[derived]` / `[output]` reads are **legitimate** — a model with `[derived] AUCF = F * 100 / CL` and `[output] F` emits correct columns (`F = 0.500000`, `AUCF = 6.028641`). Rejecting there would also panic the absorption ODE twin, whose `get_or_build` `.expect()`s a re-parse of a source that re-emits `[derived]` verbatim.

Analytical models stay out of scope: they bind F/lag through an explicit `pk(..., f=F)` mapping, so a second use is stated rather than silent.

## Alternatives considered

**Warning instead of error.** Rejected: the docs already declare the case invalid, and the `F²` migration note shows prose does not enforce. ferx's policy for this exact shape is already `Err` — a covariate in the ODE RHS is rejected at parse time with the reason spelled out as "producing silently-wrong dynamics", as are forward references (#710), duplicate `d/dt`, and reserved `TIME`/`TAD`.

**A flat list of forbidden names.** Rejected for the reason #995 gives: `F` *is* the correct spelling for bioavailability. The name is fine; the name read as something else in the same model is the defect.

**Text-level census over the block source** rather than the AST. The existing dead-param census works that way deliberately, and its over-count is safe because it only ever makes a parameter look *used*. Here an over-count is a false rejection, so the AST is the right level.

## Cross-repo dependency

No ferx-r change required: I checked every `.ferx` on `origin/main` and none reads a reserved name outside its own LHS. New public items on `types.rs` are additive.

## Breaking changes
- [ ] None

<details>

A model that folds `F` into the absorption flux — the pre-dose-entry convention, e.g. `d/dt(central) = F * KA * depot / V - …` — now fails to parse instead of quietly computing `F²`. Drop `F` from the right-hand side. If the parameter was never bioavailability, rename it; the name is what routes it.

Measured blast radius is zero: no `.ferx` under `examples/`, no inline model in `src/` or `tests/`, and no ferx-r model on `origin/main` is in this class — the latter swept by running all 68 models through this branch's `ferx check` binary, not by grep.

**ferx is now stricter than NONMEM**, measured (see the anchor section): NONMEM allows a `$PK` `F1` in `$DES` and quietly computes `F²`, with no diagnostic. A mechanically translated control stream can therefore newly fail to parse even though it ran in NONMEM — relevant to anyone converting control streams, `ferxtranslate` included. The remedy is the same: drop the `F` from the translated RHS rather than transcribe `$DES` literally.

One genuine false-positive class, accepted: a **dose-free** PD model using a bare `F` as a fraction in the RHS parses today (only `W_NO_DOSES` fires) and is now rejected. The remedy is the rename the docs already prescribe, and the message says so.

</details>

## Numerical validation
- [x] Compared against: **NONMEM 7.6.0** — see the anchor section above
- [x] Reference values: `nonmem_anchor/results/dose_attr_double_use_{A,B}.tab`. ferx reproduces the correct convention to `8.5e-6` relative and rejects the one NONMEM computes as `F²`. The PR emits no new numbers of its own — it rejects models that previously produced wrong ones — but the *diagnosis* is now anchored externally rather than self-measured. All shipped examples still validate (the only failures are the three that fail identically on `main`: `ctmm_2state` needs `markov`, `warfarin_dcm` needs `nn`, `warfarin_scaled` uses the retired `gradient_method = ad`).
- [ ] Not applicable

## Performance
- [x] No significant impact expected — one AST walk per `[odes]`/`[scaling]` block at parse time; nothing on the fit hot path.

## Tests
- [ ] No new tests needed

Sixteen Tier-1 tests in `src/parser/model_parser_tests.rs` beside the `#315` carve-out they mirror, plus Tier-2 tests in `tests/modeled_rate.rs` (the data-gated `D{n}`/`R{n}` half, both engines), `tests/check_command.rs` (the `E_DOSE_ATTR_DOUBLE_USE` mapping) and `tests/dose_attr_double_use_nonmem_anchor.rs` (the NONMEM anchor).

**Every new test was mutation-tested** — four mutations, each disabling one property the tests claim to guard, each required to make a named test fail:

| mutation | disables | caught by |
|---|---|---|
| A | the parser check ever firing | the 3 rejection tests |
| B | the `EveryDose` gate, so `D{n}`/`R{n}` reject at parse time | `ordinary_names_read_in_odes_rhs_are_accepted` |
| C | case-insensitive collection of reads | the case test, and only that one |
| D | the data-gated `D{n}`/`R{n}` branch | `modeled_rate_param_also_read_in_rhs_…` |
| E | the `init_specs` walk | `dose_attr_read_in_odes_init_is_rejected` |
| F | the `n_states` bound on indexed attributes | `out_of_range_indexed_dose_attr_reports_the_compartment_error` |
| G | the `parse_error_to_diagnostic` sentinel | `dose_attr_double_use_is_reported_as_its_own_code` |
| H | the `[scaling]` `D{n}`/`R{n}` recording loop | `modeled_rate_param_read_in_scaling_…` |
| I | the `[adaptive_dosing]` rejection | `dose_attr_read_in_adaptive_observe_is_rejected` |
| J | the `[adaptive_dosing]` recording call | `coded_rate_param_read_in_adaptive_observe_is_recorded` |
| K | swallowing the `observe` parse error | `malformed_adaptive_observe_now_fails_at_parse_time` |
| L | the `adaptive_dosing` twin decline | `adaptive_observe_does_not_reach_the_absorption_twin` |
| M | the disposition-role twin decline | `dose_attr_param_reused_as_a_disposition_role_declines_the_twin` |
| N | `F` removed from the anchor model (no bioavailability at all) | the NONMEM anchor — ferx goes to exactly `2x` NONMEM-A |

My first attempt at C was a dud — I mutated the probe side, but both sides normalise, so it still matched and the test passed. Re-aimed at the collection site, where the risk actually is.

## Example
- [ ] Example added to `examples/`
- [x] Inline below

<details>
<summary>Example</summary>

```toml
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  F  = CL / V          # intended as an elimination rate constant

[odes]
  d/dt(central) = -F * central
```

```
error[E_DOSE_ATTR_DOUBLE_USE]: [odes]: `F` is this model's bioavailability — the
engine already scales each dose amount by it — but it is also read in [odes], so
the value is applied twice: once at the dose, once where you read it (#993). If
`F` is meant to be bioavailability, remove it from [odes]; if it is meant to be an
ordinary parameter, rename it — `F` is a reserved dose-attribute name.
```

</details>

## Docs
- [x] New pages linked in `docs/_quarto.yml` — N/A, no new pages
- [x] `CHANGELOG.md` `[Unreleased]` entry added, under `Changed`
- [ ] No user-visible change

`ode-models.qmd` gains a "Reading a dose attribute is an error, not a warning" section stating the diagnostic and the `EveryDose` / `CodedRateOnly` split; `individual-parameters.qmd` gains a callout on the reserved-name table. Cross-reference anchors were verified by rendering the page through pandoc and reading back the generated ids — one was wrong on the first pass and is fixed.

## Checklist
- [x] `cargo clippy` — **zero** warnings attributable to the five files this PR changes (verified by filtering `--all-targets --message-format short` output to those paths). Note for accuracy: the tree is *not* clippy-clean overall on `--features ci,survival` — `src/sens/two_cpt_explicit.rs:477,527,528,579` trip the deny-by-default `clippy::approx_constant`, and 411 warnings exist repo-wide. That file is untouched here, so those are pre-existing and out of scope.
- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched; this is parse-time only

## Reviewer hints

Focus on **scope boundaries** — the risk here is a false rejection breaking a correct model, not a missed detection:

1. `DoseAttrConsumption::of` — the `EveryDose` / `CodedRateOnly` split is what keeps `R1`-as-a-rate-constant working. Mutation B is the guard.
2. The `[derived]`/`[output]` exclusion, and why including it would panic the absorption twin.
3. `ode_slot_map` is parallel to the individual-parameter list only as a **prefix** — the `__ferx_pktime_` desugaring appends names after the slot map is built. Both lookups use `get(i)`; a real dose attribute is always a user name, hence inside the prefix.

Skimmable: the docs and CHANGELOG prose, and the message-formatting helpers.

## Review trail

Three adversarial passes, each of which found something the previous had missed. Recorded because the pattern is informative: every time the check was extended to a new surface, the new surface was fine and something *adjacent* broke.

| pass | found |
|---|---|
| xhigh, whole PR | the `[odes] init(...)` hole; the untested `E_DOSE_ATTR_DOUBLE_USE` mapping; the untested `[scaling]` recording loop; out-of-range `F{n}` reporting the wrong defect |
| self-check | `[adaptive_dosing] observe` — a fourth prediction path, and the one that biases dosing decisions |
| medium, last commit | the twin panic via re-emitted `[adaptive_dosing]`; then, probing that, a second twin panic via a dose-attribute parameter reused as a disposition role |

Filed rather than fixed here, all assigned:

- **#1004** — the analytical engine keeps the same silent double use. Out of the agreed scope, lands on the default engine, and the severity is genuinely open (`build_obs_scale_spec` already treats the analogous volume double-divide as a *warning*). Needs its own blast-radius sweep first.
- **#1005** — `[scaling]` is parsed a second time to collect the reads, with a narrower `ParseCtx` than the real parse. Unreachable today only because of the ODE case-collision guard.
- **#1008** — `get_or_build`'s `.expect()` turns *any* twin parse failure into a mid-fit panic. Both known instances are fixed here; the mechanism is still armed for the next ODE-only check anyone adds.

## Open questions

`build_ode_spec` is the natural home for the `[odes]` check, but parse errors are first-error-wins (#767), so this can pre-empt other diagnostics on a model with several problems. It sits with its existing peers rather than ahead of them, so ordering among ODE-block errors is unchanged — but it is a real cost that #767 is the systemic fix for. The `[adaptive_dosing]` walk runs before `[scaling]` is parsed, so an `observe` error now pre-empts a `[scaling]` syntax error; same class, same fix.


TeunP:
## Review (xhigh, adversarial self-review)

Read every hunk plus the enclosing functions, then probed the parts I could not settle by reading. Nine findings — **seven fixed in `43dd2aaa`**, two deliberately left with reasons below.

### Fixed

**1. `init(state) = …` escaped the check entirely.** *(correctness)*

`build_ode_spec` splits `init(...)` directives out into `init_specs` *before* `stmts_owned` is parsed, and the read-set only walked `stmts_owned`. So this parsed clean — measured, not inferred:

```
[odes]
  init(central) = F * 100.0
```

That is a hole the exact size of the rule's own remedy: a user told "remove `F` from `[odes]`" could move it into the initial condition and get the identical silent double application, from the same block the error names. `init_specs` is now walked too.

**2. Nothing tested the `E_DOSE_ATTR_DOUBLE_USE` mapping.** *(test-coverage)*

The whole argument for minting a code instead of leaving it `E_PARSE` is that `ferxtranslate` / ferx-r can act on it without matching prose — but the mapping keys off the substring `"reserved dose-attribute name"`, and nothing asserted the code. Reword the message and every parse-time case silently degrades to `E_PARSE` with the suite green. Added a Tier-2 test in `tests/check_command.rs`, which also pins `block == None` (the reason `.with_block()` was deliberately omitted).

**3. The `[scaling]` `D{n}`/`R{n}` recording loop was untested on both engines.** *(test-coverage)*

`tests/modeled_rate.rs` reads `R1` in the *RHS*, which `build_ode_spec` records — the `[scaling]` loop marks nothing there. It is the **only** recorder an analytical model has. I confirmed it works (ODE `y = central/V + 0.0*R1` and analytical `obs_scale = V + 0.0*R1` both produce the diagnostic under `RATE=-1`), but deleting the loop, or pointing `active_dose_attr_map_mut()` at the wrong engine's map, failed nothing. Now covered on both engines, plus the no-coded-`RATE` silence.

**4. Out-of-range `F{n}` reported the wrong defect.** *(correctness)*

The check ran ahead of the `cmt > n_states` validation, so `F5` on a one-state model was told to rename itself when the real problem is that compartment 5 does not exist. `check_dose_attr_double_use` now takes `n_states` and declines out-of-range indices, leaving them to the precise error twenty lines later.

**5–7.** `E_DOSE_ATTR_DOUBLE_USE` added to the code table in `docs/file-formats/check-report.qmd` — the page the "consumers can act on the code" argument depends on. `parse_error_to_diagnostic`'s doc comment no longer claims two recognised shapes. The `individual-parameters.qmd` cross-reference is labelled with its actual target instead of "Bioavailability". Docs and CHANGELOG now say `init(...)` counts.

All four behaviours are mutation-tested — each mutation caught by exactly the test that claims to guard it, and only that test:

| mutation | caught by |
|---|---|
| drop the `init_specs` walk | `dose_attr_read_in_odes_init_is_rejected` |
| drop the `n_states` guard | `out_of_range_indexed_dose_attr_reports_the_compartment_error` |
| break the `parse_error_to_diagnostic` sentinel | `dose_attr_double_use_is_reported_as_its_own_code` |
| disable the `[scaling]` recording loop | `modeled_rate_param_read_in_scaling_is_rejected_on_both_engines` |

Full suite after: lib 3075 passed / 0 failed, every integration binary green, rustfmt clean, zero clippy warnings on the changed lines (the four `approx_constant` errors are pre-existing in the untouched `src/sens/two_cpt_explicit.rs`).

### Left, with reasons — both filed as follow-ups

**8. The analytical engine keeps the same silent double use — and a new test now pins it as correct.** *(scope; filed as #1004)*

`analytical_f_mapping_is_not_a_double_use` argues an explicit `pk(..., f=F)` makes the second use "stated rather than silent". That does not hold for a readout that never mentions the volume: `pk one_cpt_iv(cl=CL, v=V, f=F)` with `[scaling] obs_scale = 50.0 * F` applies `F` at the dose and again in the readout, with no diagnostic at all. The example the test happens to use (`obs_scale = V * F`) trips the pre-existing volume warning in `build_obs_scale_spec` — which is itself the codebase's own precedent that this shape is worth flagging, one severity below what this PR now does for ODE models.

Outside the agreed scope (#993 + `Dn`/`Rn`), and it lands on the **default** engine — #1003 shipped the ODE half only after measuring a blast radius of zero across `examples/`, `src/`, `tests/` and ferx-r `origin/main`, and that sweep has not been run for analytical models. The severity is also genuinely open: `build_obs_scale_spec` already treats the analogous volume double-divide as a *warning*, so the codebase's own precedent points one notch below what this PR does for ODE. Filed as #1004 with the sweep as step 1.

**9. `[scaling]` is parsed a second time to collect the reads.** *(simplification; filed as #1005)*

The re-walk re-runs `split_scaling_entry` / `parse_scaling_key` / `parse_scalar_expression` with a **narrower** `ParseCtx` than the real parse (no state names in `defined_vars`), so a name that is both a state and an individual parameter resolves to `Covariate` here and `Variable` there. Only the ODE case-collision guard makes that unreachable today, and nothing ties the two parses together. The clean fix — `parse_scaling_block` returning the referenced names — is a signature change to a 300-line, 12-argument function, a pure refactor with no behaviour change: not worth the churn inside a fix PR, but a real seam. Filed as #1005.

### Checked and found sound

- **The absorption ODE twin cannot be made to panic by the new error.** I probed `v=F1`, `mtt=ALAG1` and `mtt=D1` on `one_cpt_transit`; #725's unmapped-dose-attribute rejection fires first in every case, so the twin never builds a source the new check could reject. `transit_twin_with_reserved_f_name_still_builds` is the right instinct and the exposure is genuinely closed.
- `visit_expr_nodes` covers every recursive `Expression` variant, so no read hides inside a `Conditional` or a `UnaryFn` argument.
- `check_modeled_dose_rates`'s `reported` set guards both branches — N coded rows still yield one diagnostic, not N.
- Ordering is safe on both sides: `build_ode_spec` runs long before `[scaling]`, so `active_dose_attr_map_mut()` reaches the ODE map; `model.dose_attr_map` is populated before the scaling loop on the analytical side. The `ode_slot_map` prefix-parallelism claim in the reviewer hints is accurate — both synthetic families (`__ferx_ro_*`, `__ferx_pktime_*`) are append-only.

Issue #1011, opened 2026-08-20
SAEM: fixed-effect-only theta (no ETA) drifts in the un-damped numerical M-step — FREM TVFRD1 0.039 vs 0.39


roninsightrx:
## Summary

SAEM's numerical M-step for a **fixed-effect-only theta** (one with no ETA anywhere, so
not mu-referenced) is biased, not merely slow. On a FREM `iiv_on_ruv` model it drives the
absorption fraction `TVFRD1` from its 0.383 start to **0.039**, while IMP (0.311), IMPMAP
(0.318) and NONMEM IMP (0.394) agree — and it drags `TVV` +6% and `TVMAT` +9% along with it.

Restarting SAEM *at* the IMPMAP solution does not rescue it: it walks the parameter back
down to 0.065. So this is not a starting-value or burn-in artifact — the M-step actively
pushes the parameter away from the marginal optimum.

#1004 adds the user-facing advisory for this case (SAEM previously said nothing; IMP/IMPMAP
have warned since #406). This issue tracks fixing the estimator.

## Reproducer

`~/git/ferx/frem/frem_reprex_v030` (`frem_iivruv.ferx` + `frem_iivruv_data.csv`, 475 subjects,
12 ETAs, BLOCK(11) + `omega ETA_RUV` + `iiv_on_ruv`). `TVFRD1` splits absorption between a
zero-order duration and a first-order rate, and carries no ETA:

```
theta TVFRD1(0.38267823715853594, 0.01, 0.99)
...
MAT = TVMAT * exp(ETA_MAT)
D1  = MAT * (1.0 - TVFRD1)      # modeled duration, RATE=-2 in the data
KA  = 1.0 / (MAT * TVFRD1)
```

`method = saem`, `n_exploration = 1000`, `n_convergence = 1000`, `n_mh_steps = 50`,
`omega_burnin = 100`, `adapt_interval = 25`, `seed = 12345`. ~55 s per run.

## Measurements

| Run | TVCL | TVV | TVMAT | TVFRD1 |
|---|---|---|---|---|
| NONMEM IMP (reference) | 7.060 | 133.8 | 2.680 | 0.394 |
| ferx IMP | 7.068 | 135.5 | 2.751 | 0.311 |
| ferx IMPMAP | 7.046 | 134.2 | 2.780 | 0.318 |
| ferx SAEM | 7.169 | 143.2 | 3.020 | **0.039** |
[...]

Issue #996, opened 2026-08-19
SAEM/IMP mixtures: class-aware mu-reference shift for MIXNUM-switched typical values (#985 follow-up)


roninsightrx:
## Context

Follow-up to #985 / PR #992 (IMP/IMPMAP for mixtures). Every stochastic mixture
estimator currently estimates the `MIXNUM`-switched typical value by a full
numerical M-step, because the closed-form mu-reference θ move assumes **one θ per
η** and a mixture pairs one η with *K* class θ's.

`docs/estimation/mixture.qmd` states the user-visible cost:

> IMP is the noisiest of the mixture estimators — the class-switched typical value
> cannot be mu-referenced, so θ is estimated by the weighted M-step alone, which
> converges more slowly than SAEM's MCMC and typically needs more iterations.

Concretely, on the two-class anchor estimating IMPMAP recovers `TVCL1 ≈ 1.04`,
`TVCL2 ≈ 2.58` against the FOCEI optimum `0.98` / `2.84` — the class θ's are the
worst-recovered quantities in the fit, and they are exactly the ones a mu-ref
shift would move analytically.

## Current state — three independent layers say "no"

1. **Detector never fires.** `detect_pattern` (`src/parser/model_parser.rs:474`)
   has no `Expression::If` arm, and the canonical ferx mixture form is a ternary
   if-*expression*:

   ```
   CL = if (MIXNUM == 1) TVCL1 * exp(ETA_CL) else TVCL2 * exp(ETA_CL)
   ```

   `collect_mul_anchors` finds no `Expression::Theta` anchor (the branch arms are
   the anchors), so `CompiledModel.mu_refs` is **empty** for a mixture regardless
   of anything downstream. Note this is also *silent*: the
   "Mu-referencing disabled for conditional parameter(s)" warning
   (`model_parser.rs:2644`) keys off `Statement::If` branches, not an if-expression.
2. **SAEM gate.** `use_closed_form_mstep = options.mu_referencing &&
   !mu_ref_pairs.is_empty() && saem_mix.is_none()` (`src/estimation/saem.rs:1961`),
   with the rationale at `saem.rs:1903-1907`:
   > a `MIXNUM`-switched typical value pairs one η with several class thetas, so the
   > closed-form `log_theta += γ·mean(η)` update (which assumes one theta per η) is
   > ill-defined.
3. **IMP gate.** Same disable at `src/estimation/impmap.rs:1499-1503`; the mixture
[...]

Issue #918, opened 2026-07-22
SAEM: Add support for logit-normal MU-referencing for bounded (0,1) parameters


JordanBrooks33:
## Summary
SAEM fails to correctly estimate bounded (0,1) parameters (e.g., bioavailability fractions) when IIV-on-RUV is present. The auto-MU-referencing detection doesn't recognize the logit-normal form `param = inv_logit(theta + eta)`, causing biased M-step optimization.

## Reproduction

**Model**: FREM with IIV-on-RUV and a fraction parameter FRD1 ∈ (0,1)

```
# Logit parameterization with ETA
theta LOGIT_FRD1(-0.477, -10.0, 10.0)
omega ETA_FRD1 ~ 0.01  # in block omega

FRD1 = 1.0 / (1.0 + exp(-(LOGIT_FRD1 + ETA_FRD1)))
```

**Results** (same model, same data):

| Method | FRD1 | Error vs Reference |
|--------|------|-------------------|
| IMP | 0.321 | reference |
| IMPMAP | 0.322 | reference |
| **SAEM** | **0.139** | **-57%** |

SAEM warning: `individual parameter(s) not mu-referenced: FRD1`

## Root Cause Analysis

ferx auto-MU-referencing recognizes:
- `CL = TVCL * exp(ETA_CL)` ✓
- `COV = TV + ETA` ✓

But does **not** recognize:
- `FRD1 = inv_logit(LOGIT_FRD1 + ETA_FRD1)` ✗

This causes LOGIT_FRD1 to be optimized via BOBYQA in the M-step rather than analytically. When IIV-on-RUV is present, the conditional samples are noisy, leading to biased gradients.

## What We Tried

| Approach | FRD1 Result | Other Params |
|----------|-------------|--------------|
| Default auto-MU-ref | 0.22 (-31%) | OK |
| `mu_referencing=FALSE` | 0.35 (+9%) | TVMAT: -60% |
| Manual MU syntax | 0.32 (-1%) | TVMAT: -86% |
| Full block omega with ETA_FRD1 | 0.14 (-57%) | OK |

**Key finding**: `mu_referencing=FALSE` helps FRD1 but breaks TVMAT (which needs exp MU-ref). Manual MU syntax (`MU_CL = log(TVCL); CL = exp(MU_CL + ETA_CL)`) is not recognized.

## Requested Enhancement

1. **Add logit-normal MU-referencing support**:
   ```
[...]

Issue #406, opened 2026-06-18
IMP/IMPMAP diverge on high-dimensional FREM: inner MAP solver returns a wrong joint mode


roninsightrx:
## Summary

IMP and IMPMAP diverge on high-dimensional FREM models because the inner-loop
EBE/MAP solver (`inner_optimizer::find_ebe`) fails to converge the joint
posterior mode in the extreme multi-scale regime FREM produces. The
importance-sampling proposal is centered on this broken mode, so every subject
collapses to ~0 effective sample size and `−2logL(IS)` explodes to ~1e11–1e13
instead of converging.

NONMEM converges cleanly on the equivalent model
(`ferx-testdata/frem_workshop/HO1/nm/run6.mod`): IMPMAP OFV **6887**, IMP eval
**6869**.

This refines and partly supersedes the earlier hand analysis in
`ferx-testdata/frem_jordan/ISSUE_IMPMAP_HIGH_DIM_FREM.md`, which attributed the
failure to IS proposal *width*. The proposal width is a real but **secondary**
defect — the primary cause is a **wrong MAP mode**.

## Reproduction

- Model: `ferx-testdata/frem_workshop/HO1/ferx/run1_frem.ferx` (1-cpt oral,
  `log_additive`, 475 subjects, 12 ETAs → BLOCK(12) omega: CL/V/MAT + 9
  covariate pseudo-obs etas).
- Data: `data/dataset3_study7_frem.csv`.
- Run with `method = imp` (or `impmap`):

```
iter 1: -2logL(IS) = 1.07e11   (low-ESS subjects: 475 / 475)
iter 2: -2logL(IS) = 1.90e12   (low-ESS subjects: 475 / 475)
...     wanders 1e11 .. 1e13, never converges
```

## Root cause — broken MAP mode

Instrumented the obs-NLL at the returned mode (subject 256, at the initial
parameters):

```
obs_nll @ eta_hat = 1.343e8          (a sane value is O(10))
eta_hat = [CL 3.81, V 2.93, MAT -3.61, FOOD 0.01, WT 4.65, HT 11.05,
           AGE 14.64, RACE2 5.84, RACE3 4.47, BMI 4.13, CRCL -43.54, NCI 0.80]
```

PK etas are log-scale (`CL = TVCL·exp(η)`), so `η_CL = 3.8` ⇒ CL is **45×** —
nonsense. The mode is not a posterior mode at all.

[...]

PR #1012, opened 2026-08-20
feat(estimation): SAEM warns when an estimated theta carries no ETA

roninsightrx:
## Why

A colleague's FREM `iiv_on_ruv` reprex (`~/git/ferx/frem/frem_reprex_v030`, 475 subjects,
12 ETAs) still mis-estimates the absorption fraction `TVFRD1` under SAEM on v0.3.0. The
reprex attributes it to #918 (logit-normal mu-referencing). That is a misdiagnosis:
`TVFRD1` has **no ETA and no logit link**, so there is no mu-ref pattern for #918 to detect.

It is instead the #406 class — a **fixed-effect-only theta**. Such a theta is not
mu-referenced, so it never receives the γ-damped closed-form `log θ += γ·mean(η)` update.
Its only channel is the η-frozen numerical M-step (`theta_sigma_mstep_light`), whose result
is assigned undamped (`log_theta = theta_new;` — `gamma` goes to the mu-ref shift and
`gamma_omega` to Ω, neither reaches here). It therefore re-maximises the conditional
observation likelihood against a *single* MCMC η draw every iteration: MCEM with K = 1, not
a stochastic-approximation update on an averaged sufficient statistic. It drifts.

SAEM said nothing about this. Its existing advisory
(`saem_non_mu_referenced_individual_params_warning`, `src/api/fit.rs`) walks
`indiv_param_names` and skips any parameter with `eta_idx < 0`, so it structurally covers
only "has an ETA that could not be mu-referenced" and cannot see "has no ETA at all".
IMP/IMPMAP have warned about the latter since #406.

The estimator fix is tracked separately in #1011. This PR is the advisory, so the failure
stops being silent.

## What changed

`run_saem` now calls `non_fixed_thetas_without_eta` — the same helper IMP/IMPMAP use — and
pushes a SAEM-labelled advisory naming the offending thetas. Placed right after
`use_closed_form_mstep` is decided, so the mixture case can pass the class thetas whose
class-aware shift is actually active (matching the contract that helper documents).

Plus a `docs/estimation/saem.qmd` callout with the measurements and the two working
remedies, and a `CHANGELOG.md` `[Unreleased]` entry.

Warning-only. Estimates on the reprex are bit-identical before and after
(`TVFRD1 = 0.038803`, `TVV = 143.156643`, `TVMAT = 3.019761`).

## Alternatives considered

Fixing the estimator here rather than warning. Rejected for this PR: the candidate fixes
(SA-damping the assignment, maximising over a bank of recent η draws, protecting the channel
during burn-in, auto-attaching a dummy ETA) all change numerical results across every SAEM
model and need their own validation sweep. They are written up in #1011. The advisory is
independent of whichever one lands and is worth having regardless — the user currently gets
a wrong answer with no diagnostic at all.

Extending the existing `saem_non_mu_referenced_individual_params_warning` instead of adding
a second one. Rejected: that function is keyed on `indiv_param_names` × `eta_map` and is
about *ETA forms* ferx could not pattern-match; a theta with no ETA is not in its domain and
the remedies differ (rewrite the ETA form vs. add an ETA / hold FIX / change estimator).
The two are complementary and non-overlapping.

## Cross-repo dependency

No public Rust API change; the warning flows through the existing `FitResult.warnings`
channel that ferx-r already renders.

## Breaking changes

- [x] None

## Numerical validation

- [x] Compared against: NONMEM IMP, ferx IMP, ferx IMPMAP, and prior ferx commit
- [x] Reference values:

Reprex: `frem_iivruv.ferx` + `frem_iivruv_data.csv`, `method = saem`,
`n_exploration = 1000`, `n_convergence = 1000`, `n_mh_steps = 50`, `omega_burnin = 100`,
`adapt_interval = 25`, `seed = 12345`.

```
run                                     TVCL     TVV   TVMAT   TVFRD1
NONMEM IMP (reference)                 7.060   133.8   2.680    0.394
ferx IMP                               7.068   135.5   2.751    0.311
ferx IMPMAP                            7.046   134.2   2.780    0.318
ferx SAEM                              7.169   143.2   3.020    0.039   <- the defect
ferx SAEM started at the IMPMAP fit     7.046   136.4   3.191    0.065
ferx SAEM + ETA_FRD1 (omega^2 = 0.01)   7.181   142.5   2.646    0.313   <- remedy 1
ferx SAEM, TVFRD1 FIX at 0.394          7.025   131.9   2.770        -   <- remedy 2
```

Restarting SAEM at the IMPMAP solution still walks `TVFRD1` back down, so this is not a
starting-value problem. `TVV` (+6%) and `TVMAT` (+9%) are collateral — the FIX row puts all
three within 3% of NONMEM, confirming `TVFRD1` is the single defect.

From `ferx_trace()`, `TVFRD1` leaves its 0.383 start on the *first* iteration and decays
monotonically through the window where MH acceptance is 0.09–0.13 (targets are 0.40/0.44)
and Ω is still held by `omega_burnin`. The mu-referenced `TVMAT` does not move in that same
window, because its only update is `γ·mean(η)` and `mean(η) ≈ 0` while the chain is stuck.

**Before/after this PR:** identical estimates, one added warning line. The advisory is
warning-only by construction — it reads `model.mu_refs` and `theta_fixed` and pushes a
string; no optimizer state is touched.

## Performance

One `HashSet` build over `mu_refs` at fit setup.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3026 passed)

`saem_warns_when_an_estimated_theta_has_no_eta` — a single-population 1-cpt IV model with
`V = TVV` (no ETA) next to a mu-referenced `CL`; asserts the advisory fires, is
SAEM-labelled, names `TVV`, and does *not* name the mu-referenced `TVCL`.

`saem_no_eta_advisory_silent_when_every_theta_is_mu_referenced` — same model with
`V = TVV * exp(ETA_V)`; asserts silence, so the advisory keeps its signal.

Both run a 4-exploration + 2-convergence SAEM fit (Tier 1, sub-second), following the
`saem_mixture_*` pattern already in the module.

## Example

- [x] Not applicable (no new API surface — an added diagnostic on an existing channel)

The advisory as emitted on the reprex:

```
SAEM: estimated parameter(s) [TVFRD1] have NO associated ETA, so they are not
mu-referenced and are moved only by the η-frozen numerical M-step. That channel
re-maximises against a single MCMC η draw with no SA damping and can drift far from the
marginal optimum, taking correlated typical values with it. STRONGLY add an ETA to each
(e.g. `P = TVP * exp(ETA_P)` with a small, optionally FIX, omega — ferx applies
mu-referencing automatically), or hold the parameter FIX, or use FOCEI/IMPMAP.
```

## Docs

- [x] `docs/` updated for user-visible changes — new callout in `docs/estimation/saem.qmd`
      beside the existing non-mu-referenced warning, with the comparison table and both remedies
- [x] New pages linked in `docs/_quarto.yml` — N/A, existing page

## Checklist

- [x] `cargo clippy` clean (no new warnings; the pre-existing `saem.rs` hits at lines
      829/870/1414/1438 are untouched)
- [x] `Cargo.toml` version bump not needed (additive, non-breaking)

### ferx-site

### ferx-book
- [x] No new estimator or fit option

### Example execution
- [x] Rebuilt ferx-core: `cargo build --release`, reran the reprex through the CLI — advisory
      fires, estimates unchanged to the last printed digit
- [x] No `examples/*.ferx` affected

## Reviewer hints

The substance is the 30-line block in `src/estimation/saem.rs` after
`use_closed_form_mstep`. Two things worth checking:

1. **`class_mu_ref_thetas` for mixtures.** `non_fixed_thetas_without_eta` documents that
   callers must pass only the class thetas whose class-aware shift is *actually active*, so
   a class theta that fell back to the numerical M-step is correctly flagged. SAEM
   deliberately excludes MIXNUM-switched thetas from the closed form (#996, hard-class
   bias), so they will now be named by this advisory too. I believe that is right — they are
   exactly in the situation the advisory describes — but it is the one place where the new
   warning changes what a mixture user sees.
2. **Overlap with the existing advisory.** Confirm the two really are disjoint:
   `saem_non_mu_referenced_individual_params_warning` `continue`s on `eta_idx < 0`, which is
   precisely the set this one covers.

The docs table and the `#1011` write-up can be skimmed — they are the same numbers as the
Numerical validation section above.

## Open questions

Should the mixture case be quieter? See reviewer hint 1 — a `MIXNUM`-switched theta already
gets the #996 "estimated by the numerical M-step" advisory, so it will now carry two
overlapping messages. I left both because they say different things (why it is numerical vs.
what to do about it), but happy to suppress this one when the #996 one already fired.


roninsightrx:
## Code review (medium) — `origin/main...HEAD`

Merged latest `main` into this branch first (`22098276..777c403a`, docs-only, clean). `cargo check --tests` passes.

Note this PR is **stacked**: the diff against `main` carries both #1014's M-step damping and #1012's no-ETA advisory, so the review covers both.

### Medium

**1. `src/estimation/saem.rs:3276` — σ is damped on a θ-keyed gate.**

`damp_mstep(&mut log_sigma, &sigma_new, gamma_theta)` uses `numerically_estimated_theta`, a predicate that says nothing about σ.

On the θ side the blanket application is benign — a pinned or FIXed θ comes back from NLopt with `new == cur`, so the blend is a no-op and the effective damped set is exactly the free non-anchor θ (which is the intended #1011 target). σ has no such protection: it is genuinely re-maximised by NLopt every iteration. So an all-mu-referenced model leaves σ on the undamped assignment, and adding one unrelated covariate coefficient silently switches σ onto a 0.03-capped SA update. Either σ needs its own gate, or it should stay undamped.

**2. `src/estimation/saem.rs:2480` — the advisory fires on thetas where "add an ETA" is wrong advice.**

`non_fixed_thetas_without_eta` flags any free θ not in `model.mu_refs`. On the module's own `mix996_model` that emits `[TVCL1, TVCL2, MIXL]`. `MIXL` is the mixing-probability theta, updated by `mstep_mixing` (the separate logistic M-step at :3426), not by the numerical θ M-step — both the diagnosis and the remedy are wrong for it. Same shape for covariate coefficients (`WTCL` in `CL = TVCL*(WT/70)^WTCL*exp(ETA_CL)`), `F1`, `ALAG`: they are genuinely fixed-effect-only, but "give it an ETA" is not the remedy a user should follow.

Suggest excluding `mix.mixing_theta_idx` outright, and softening the remedy wording for coefficient-type thetas. (This is your Open Question 1 — my read is the `MIXNUM` class thetas *should* stay named, but `MIXL` should not.)

**3. Blast radius vs. anchors — the damping gate trips for most covariate models, which the validation doesn't cover.**

`damps_numerical_mstep` is `.any()` over free non-anchor θ, so a routine allometric model trips it on the exponent alone, and no "has no effect" warning fires to tell the user. That is defensible (`WTCL` *is* the #1011 defect class), but the numerical anchors here are one FREM reprex plus one mixture control.

Compounding it: `theta_sigma_mstep_light` runs NLopt warm-started at `maxiter = 3` (exploration) / `5` (convergence), so `theta_new` is already a partial step, not the `argmax` the Robbins-Monro justification assumes. Scaling that by 0.03 — and in convergence by `1/(k−k1)` — shrinks an already-truncated step ~33×; for a legitimately identified non-mu-referenced θ starting far from its optimum, that risks stalling.

Please run the gated slow SAEM convergence tests before merge: `tests/saem_block_omega_collapse.rs`, `tests/iov_convergence.rs`, `tests/frem_warfarin.rs`.

### Low

**4. `src/estimation/saem.rs:182` — `mstep_sa_step` has a cliff at `cap == 1.0`.** `cap` is documented as an exploration-phase cap, and convergence returns `gamma` regardless of `cap` — except `cap >= 1.0` short-circuits both phases to `1.0`. So `mstep_damping = 0.999` gives full convergence-phase SA damping and `1.0` gives none: a user tuning down from "off" gets a discontinuity, not a gradient.

**5. `src/estimation/saem.rs:211` — `sanitize_mstep_damping` inverts intent for `+INFINITY`.** `!cap.is_finite()` is tested before the `> 1.0` branch, so a programmatic caller passing `f64::INFINITY` (naturally read as "no damping") gets `MSTEP_SA_MAX_STEP = 0.03` — maximum damping. `INFINITY` should fall into the `> 1.0 → 1.0` ("off") branch; only NaN needs the default.

**6. `src/estimation/saem.rs:2557` — the "no effect" reason text is wrong on one path it covers.** It says "every estimated theta in this model is mu-referenced … or is FIX", but the branch also fires under `mu_referencing = false` — `theta_is_mu_ref_anchor_mask` is a structural property and ignores the option, so anchors mask as anchors even though nothing is actually mu-referenced and the numerical M-step *does* move them. (Your own `mstep_damping_is_inert_when_mu_referencing_is_off_but_every_theta_is_anchored` exercises exactly that.) The skip is intentional and documented at :140; only the user-facing wording needs to say "mu-referenceable" / mention `mu_referencing = false`.

**7. `src/estimation/saem.rs:186` — the first convergence iteration is undamped.** At `k = k1 + 1`, `gamma = 1/(k−k1) = 1.0`, so the whole exploration-accumulated θ/σ is overwritten by a single-draw maximiser — the exact assignment this change exists to avoid. It mirrors the pre-existing `gamma_omega` schedule so it is probably intentional; worth a sentence in the doc comment.

roninsightrx:
## Review findings addressed — `984f52e0`

All seven, plus the validation run. One correction to my own finding #1 below.

**#1 — σ damped on a θ-keyed gate.** Not an oversight after all: `theta_sigma_mstep_light` is a *single* NLopt problem over the concatenated `[θ; σ]` vector, so `theta_new` and `sigma_new` are one joint maximiser and must share one γ. Damping only θ would leave σ jumping to the value optimal for the θ the M-step wanted while θ is held back — σ absorbing exactly the misfit the damping just stopped θ from fixing. So the behaviour stands; the defect was that it read as accidental. `gamma_theta` → `gamma_mstep`, with the joint-solve rationale in `mstep_sa_step`, `damp_mstep`, the SAEM page and the fit-options table. Also documented what makes the blanket application honest on the θ side: a pinned or FIXed dim comes back `new == cur`, so the damped set is already exactly the free un-anchored θ.

**#2 — advisory names thetas where "add an ETA" is wrong.** Fixed, and it answers your Open Question 1 the *opposite* way to what I first suggested — suppress. `MIXL` and the `MIXNUM`-switched class thetas are both filtered out now:

- Mixing coefficients: never moved by the numerical θ/σ M-step, and the parser *forbids* a mixing expression from depending on an η, so both halves of the message are false for them. `run_mcem_mixture` has filtered the same set since #992.
- `MIXNUM`-switched anchors and identity-scale-dropped log-mu-refs: they carry an η in the model text and already get their own message (`mix_switched_skipped` / `dropped_identity`), which says everything true and actionable. IMP/IMPMAP make the same split via `shift_disabled`; SAEM already pushes the equivalent, so here it is a filter rather than a second partition. Under `mu_referencing = false` those messages don't fire, so the thetas correctly stay in scope.

Remedy text now splits by what the parameter is: a typical value → mu-referenceable form; a covariate coefficient / allometric exponent / structural constant → `FIX`, or cross-check against FOCEI/IMPMAP.

**#3 — blast radius vs. anchors.** Ran the gated suites on the merged branch: `frem_warfarin` (14), `iov_convergence` (11), `mixture_iov_nonmem`, `saem_block_omega_collapse`, `saem_combined_error`, `saem_omega_burnin` — **29 passed, 0 failed**, NONMEM mixture anchor included. The truncated-M-step concern (`maxiter = 3` × γ) doesn't show up in any of them, so I'm treating it as answered rather than pursuing it further.

**#4 / #7 — the `1.0` cliff and the undamped first convergence iteration.** Both left as-is behaviourally, both now documented where a reader hits them. `1.0` is a *sentinel*, not a cap value; the option is discontinuous there because `cap` only ever governs exploration and "off" has to restore the whole pre-#1011 trajectory to be a useful escape hatch (your existing test pins why: lifting only the exploration cap left the reprex at 0.047, not 0.039). The first convergence iteration is undamped at `γ = 1/(k−k1) = 1`, mirroring `gamma_omega` exactly.

**#5 — `sanitize_mstep_damping(+∞)`.** Real bug, fixed. `!cap.is_finite()` was tested before `> 1.0`, so `f64::INFINITY` — which reads as "no damping" — got `MSTEP_SA_MAX_STEP`, the *tightest* damping. Now only NaN takes the default; `+∞` falls through to the `1.0` off value.

**#6 — the "no effect" reason text.** Fixed. The gate reads a structural property of the model text, so it stays true under `mu_referencing = false` even though nothing is mu-referenced and the numerical M-step does move every theta. Skipping the damping there is deliberate (and the docs now say so explicitly, as out-of-scope rather than as no-op), but the message no longer claims the thetas are mu-referenced — it names that case.

### Tests

- `saem_no_eta_advisory_skips_mixing_and_already_explained_class_thetas` — the mixture advisory must skip `MIXL`/`TVCL1`/`TVCL2` *and* the #996 message must still fire (suppress the wrong message, not the right one).
- `saem_no_eta_advisory_still_fires_for_a_real_no_eta_theta_in_a_mixture` — filter keeps its signal: a genuine no-ETA theta in the same mixture is still named, and named alone.
- `mstep_damping_is_inert_when_mu_referencing_is_off_but_every_theta_is_anchored` — tightened to assert the *reason*, not just that some warning fired.
- `sanitize_mstep_damping_clamps_out_of_range_values` — `+∞` moves from the fallback group to the off group.

`cargo test --lib` 3036 passed / 0 failed. `cargo fmt` clean, no new clippy hits in `saem.rs` (the four pre-existing ones are untouched).

Docs and `CHANGELOG.md` updated for every user-visible piece: the σ rationale, the `1.0` sentinel, the reworded remedy, the two advisory exclusions, and the `+∞` clamp.

Issue #619, opened 2026-06-30
SAEM: covariate effect on IIV parameter collapses to bound when typical value isn't log-linear in one theta (non-mu-referenceable)


roninsightrx:
## Summary

SAEM cannot estimate a covariate effect on an IIV parameter when the parameter's typical value is **not log-linear in a single theta**. The covariate slope confounds with the random effect and collapses to its bound, inflating OFV.

Surfaced on the fluconazole problem (ADVAN3/TRANS4 translation of `fluconfinal014.mod`).

## Reproduction

Model CL line (additive renal gradient, matching NONMEM):

```
CL = (TVCL_NR + (CKDCRCYS - 90.0) * THETA_CLCKD) * exp(ETA_CL)
```

`method = saem`, seed 123. Result vs NONMEM FOCE:

| Param        | SAEM (this model) | NONMEM |
|--------------|-------------------|--------|
| THETA_CLCKD  | **0.000000** (pinned at lower bound) | 1.41 |
| TVCL_NR      | 86.7              | 122    |
| ETA_CL ω     | 0.439 (inflated)  | 0.543  |
| **OFV**      | **1214**          | **734.6** |

The renal covariate — the model's dominant covariate — is effectively dropped.

## Root cause

`detect_pattern` (`src/parser/model_parser.rs:426`) only marks an ETA as mu-referenceable when its typical-value factor is log-linear in a **single** theta (`THETA*exp(ETA)`, `exp(log(THETA)+ETA)`, `THETA*exp(ETA)*const`, additive `THETA+ETA`). The CL factor `(TVCL_NR + (CKDCRCYS-90)*THETA_CLCKD)` carries **two** thetas → `collect_mul_anchors` returns 2 anchors → pattern rejected → ETA_CL is not mu-referenced (run warning confirms: `mu-ref: ETA_BMAX, ETA_V1`, no ETA_CL).

Failure chain:
1. ETA_CL falls to the non-mu-ref M-step (`theta_sigma_mstep_light`, 3–5 derivative-free BOBYQA iters), which minimizes obs-NLL with the **sampled etas held fixed**.
2. The MH sampler lets ETA_CL absorb the between-subject CL variation, including the renal-covariate-correlated part.
3. With etas frozen, the M-step sees ~zero gradient on THETA_CLCKD → slides to the lower bound `0` and sticks.
4. Covariate dropped → TVCL_NR biased low, ETA_CL inflated, OFV up.

[...]

PR #1014, opened 2026-08-21
fix(estimation): SA-damp the SAEM numerical theta/sigma M-step [closes #1011]

### LINKED ISSUE #1011: SAEM: fixed-effect-only theta (no ETA) drifts in the un-damped numerical M-step — FREM TVFRD1 0.039 vs 0.39

roninsightrx:
Closes #1011.

## Why

SAEM drove the absorption fraction `TVFRD1` of a FREM `iiv_on_ruv` model to
**0.039** while ferx IMP (0.311), ferx IMPMAP (0.318) and NONMEM IMP (0.394)
agreed, dragging `TVV` +6% and `TVMAT` +9% with it. `TVFRD1` carries no ETA, so
it is not mu-referenced.

The marginal likelihood is not ambiguous about this. Profiling −2logL over
`TVFRD1` (all other parameters `FIX`ed at the IMPMAP solution, evaluated with
IMPMAP's MAP-centred proposals) is smooth, unimodal, and minimised at ≈ 0.29:

```
TVFRD1   0.02    0.0388  0.06    0.10    0.15    0.20    0.25    0.3184  0.394   0.50    0.65
-2logL   6985.9  6980.3  6974.5  6966.0  6958.0  6951.9  6948.8  6948.6  6951.9  6963.0  6992.1
```

SAEM sat 32 units up a monotone slope with a clear downhill gradient, and its
joint solution evaluated 26 ± 4 units worse than IMPMAP's. So this was not a
competing local optimum — SAEM was stopping where the likelihood was still
falling.

**Root cause.** The numerical θ/σ M-step assigned NLopt's maximiser outright:

```rust
log_theta = theta_new;
log_sigma = sigma_new;
```

That re-maximises the η-frozen conditional observation likelihood against a
*single* MCMC η draw every iteration. It is `argmax` of one draw, not the
stochastic-approximation average of `E[argmax]`, so it carries a Monte-Carlo bias
that does not decay with iteration count. A log-mu-referenced θ never sees this —
its closed-form `log θ += γ·mean(η)` update *is* an exact Robbins-Monro average —
and Ω was already protected by a per-iteration SA cap. The θ channel was the one
left exposed, and `OMEGA_SA_MAX_STEP`'s own comment says so:

> Maximum per-iteration stochastic-approximation step for the Ω sufficient
> statistic *during the exploration phase*. **The θ/σ M-step uses the full γ (1.0
> in exploration)**, but Ω is averaged at no more than this rate so a single
> un-equilibrated MCMC draw cannot overwrite a correlated Ω […]

## What changed

The M-step result is blended in as a Robbins-Monro step
`θ ← θ + γ_θ·(θ* − θ)` instead of assigned, with `γ_θ` capped at
`MSTEP_SA_MAX_STEP = 0.03` during exploration and following the full decaying
`γ = 1/(k−k1)` in convergence — the θ-side counterpart of the existing Ω cap.

The cap is the **`mstep_damping`** `[fit_options]` key (default `0.03`, range
`(0, 1]`, alias `saem_mstep_damping`):

```
[fit_options]
  method        = saem
  mstep_damping = 0.01   # smaller damps harder; 1.0 turns it off
```

```
mstep_damping=<unset>  TVFRD1 = 0.290096  TVMAT = 2.686277
mstep_damping=0.03     TVFRD1 = 0.290096  TVMAT = 2.686277
mstep_damping=1.0      TVFRD1 = 0.038803  TVMAT = 3.019761   <- exactly the pre-fix numbers
mstep_damping=0.01     TVFRD1 = 0.354514  TVMAT = 2.879758
```

`1.0` is a true escape hatch back to the previous behaviour, bit-for-bit. Setting
the key on a model it cannot affect (all-mu-referenced/`FIX`, or a mixture) warns
rather than being silently ignored.

Exposing it caught a bug in the schedule: `mstep_sa_step` applied the cap only
during exploration and returned the raw `γ` in convergence, so `1.0` still damped
at `γ = 1/(k−k1)` there and left the reprex at `TVFRD1` 0.047 rather than its true
undamped 0.038803 — i.e. "off" was not off. It now short-circuits on `cap >= 1.0`
in both phases, pinned by a regression test that names the wrong value.

Two gates keep the blast radius to the models that actually have the defect
(`damps_numerical_mstep`):

* **Only when NLopt has a θ to estimate.** A θ that is `FIX`ed or pinned out by
  the closed-form mu-ref shift is returned unchanged, so a fit whose every θ is
  one of those is byte-identical to before, σ included.
* **Not for mixtures.** See *Alternatives* — this one was found by the slow tests.

## Alternatives considered

**Averaging the M-step objective over the last K η draws** (the other direction
named in #1011). Implemented and measured: `TVFRD1` reaches 0.056 at K = 5, 0.061
at K = 20, 0.081 at K = 50, and 0.130 at K = 150 — still climbing, at 15× the wall
time (835 s vs 55 s). Consecutive SAEM draws are heavily autocorrelated, so a
K-window buys few independent samples. Damping the iterate accumulates over every
iteration instead and reached 0.290 at no extra cost, so the bank was dropped.

**Excluding mixtures** was not a guess — the Tier-3 anchors caught it. A
`MIXNUM`-switched typical value goes through this same numerical M-step (#996
routes it there deliberately), but it is solving a different problem: the class
typical values must *separate* from a common start, and the class assignments only
settle once they have. Damping that excursion stalls it — a 0.03 cap left
`TVCL1 = 1.145` against NONMEM's 1.002 on `tests/nonmem/mixture_iv_saem`, and
dragged the chained IMP marginal test down with it. Whether the #1011 bias also
affects mixture class θ (it plausibly does) needs its own schedule and its own
anchor, so it is left alone rather than half-fixed.

**Ruled out along the way**, recorded so they are not re-tried:

* *Sampler mixing.* Raising `n_mh_steps` 50 → 300 made it **worse** (0.0101, at
  the lower bound), and the chain's η_RUV spread is correct — sd 0.5206 against
  its own ω_RUV sd 0.5171, with M-step weights spanning ~[0.02, 20] and
  ESS ≈ 195/475, comparable to IMPMAP. Not an under-mixing problem.
* *σ.* Fixing σ at the IMPMAP value made it worse (0.010), so the #895 σ growth
  cap is not implicated either.
* *The LTBS prediction floor.* The M-step floors `f = preds[j].max(1e-12)` while
  the E-step does not, which under LTBS would clamp legitimately-negative
  log-scale predictions. Patched and tested: byte-identical, the floor never binds
  here (those negative IPREDs are `ln(LTBS_FLOOR)` and FREM covariate rows).
* *`n_leapfrog` being ignored.* It is not — HMC is deliberately disabled for
  `iiv_on_ruv` (the Dual2 kernel has no `exp(2·η_ruv)` rule) and says so in a
  warning. No bug.

## Cross-repo dependency

No public API change.

## Breaking changes

- [x] None

Estimates change for SAEM fits that have a free non-mu-referenced theta — that is
the fix. Fits whose thetas are all mu-referenced or `FIX`, and all mixture fits,
are unchanged.

## Numerical validation

- [x] Compared against: NONMEM IMP, ferx IMP, ferx IMPMAP, prior ferx commit
- [x] Reference values:

FREM `iiv_on_ruv` reprex (475 subjects, 12 ETAs), `method = saem`,
`n_exploration = 1000`, `n_convergence = 1000`, `n_mh_steps = 50`,
`omega_burnin = 100`, `adapt_interval = 25`, `seed = 12345`:

```
                        TVCL      TVV     TVMAT   TVFRD1   sigma
before (undamped)      7.169    143.2     3.020    0.039   0.213
after  (this PR)       7.123    139.5     2.686    0.290   0.170
NONMEM IMP             7.060    133.8     2.680    0.394   0.177
ferx IMP               7.068    135.5     2.751    0.311      --
ferx IMPMAP            7.046    134.2     2.780    0.318   0.183
marginal -2logL optimum for TVFRD1: ~0.29
```

`TVMAT` and σ now land essentially on NONMEM (2.686 vs 2.680; 0.170 vs 0.177).
`TVFRD1` sits on ferx's own marginal optimum; the residual gap to NONMEM's 0.394
is a separate, smaller matter and the advisory from #1012 still fires.

Exploration-cap sweep behind the choice of 0.03:

```
cap    TVCL      TVV        TVMAT     TVFRD1    sigma
off    7.169    143.157     3.020     0.0388    0.213
0.1    7.134    139.247     2.952     0.1697    0.172
0.03   7.123    139.536     2.686     0.2901    0.170
0.01   6.975    129.927     2.880     0.3545    0.169
0.003  7.090    136.862     2.708     0.3574    0.170
```

0.03 lands on the marginal optimum; tighter caps overshoot toward NONMEM's 0.394
but pull `TVV` off (129.9 vs 133.8).

Healthy-model check — `examples/warfarin_saem.ferx` is unchanged within noise:
OFV −286.0038 → −285.8255, thetas within 1%.

## Performance

One blend per M-step. The reprex runs in the same ~55 s before and after.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3029)

`damp_mstep_is_a_robbins_monro_blend` — pins the blend, that `γ >= 1.0` assigns
bit-for-bit (the property the gates rely on), that `γ = 0` freezes, and that a
pinned dim where `new == cur` is a no-op at every γ.

`mstep_sa_step_caps_exploration_and_frees_convergence` — the schedule: off when
there is no θ to estimate, capped in exploration, full decaying γ in convergence,
and `MSTEP_SA_MAX_STEP <= OMEGA_SA_MAX_STEP`.

`test_mstep_damping_round_trips_and_validates` / `test_mstep_damping_under_focei_warns`
— the option under both spellings, the `(0, 1]` range (1.0 accepted, 0 / negative
/ >1 rejected, and a rejected value not clobbering the previous one), and the
unsupported-key warning under FOCEI.

`damps_numerical_mstep_only_when_nlopt_has_a_theta` — the gate across all eight
cases: free/unpinned, all-pinned, all-`FIX`, unpinned-but-`FIX`, no mu-ref at all,
σ-only (no thetas), and both mixture arms.

Slow tests were run in full (`--features slow-tests`, lib **and** `--tests`) —
that is what caught the mixture regression above.

## Example

- [x] Not applicable (no new API surface — an internal M-step schedule)

## Docs

- [x] `docs/estimation/saem.qmd` — the #1012 callout is updated with the fix, the
      before/after table, the mixture exclusion, and why the η-bank alternative lost
- [x] New pages linked in `docs/_quarto.yml` — N/A, existing page
- [x] `CHANGELOG.md` `[Unreleased]` entry added (under `Fixed`)

## Checklist

### ferx-site / ferx-book

### Example execution
- [x] Rebuilt ferx-core and reran the reprex + `warfarin_saem` through the CLI
- [x] No `examples/*.ferx` needed changing

## Reviewer hints

Three small functions carry the change — `damps_numerical_mstep` (the gate),
`mstep_sa_step` (the schedule), `damp_mstep` (the blend) — plus the three
call sites that now go through `damp_mstep` instead of assigning.

Worth a look:

1. **`MSTEP_SA_MAX_STEP = 0.03` is calibrated on one model.** The sweep above is
   the whole basis for the value. A second anchor with a fixed-effect-only theta
   would firm it up; I did not have one.
2. **σ is damped alongside θ.** σ is always in this M-step, so under the gate it
   is damped whenever a free numerical θ exists. It moved 0.213 → 0.170 (toward
   NONMEM's 0.177) here, but it is coupled to the θ gate rather than judged on its
   own — deliberate, since the two are fit jointly, but it is a real coupling.
3. **The mixture veto is a whole-model veto**, not per-theta. A mixture with a
   fixed-effect-only theta unrelated to the class structure keeps the old
   behaviour for that theta too.

## Open questions

None outstanding. (An earlier revision kept the cap as an internal constant; it is
now the `mstep_damping` fit option, which is what surfaced the `1.0`-is-not-off
bug above.)

Stacked on #1012 (the advisory) — review that one first; this PR's diff is the second commit.


roninsightrx:
## Review follow-ups applied — 483ea59a

All six findings from the `/code-review medium` pass are fixed on this branch.

### 1. `mstep_damping` was validated only in the parser (medium)

`run_saem` used `options.saem_mstep_damping.unwrap_or(MSTEP_SA_MAX_STEP)` raw. `FitOptions` is public, so a Rust caller — the ferx-r glue, a test, any embedder — could hand it `Some(-0.5)`, which makes `damp_mstep` step θ and σ *away* from the M-step maximiser on every iteration, or `Some(0.0)`, which freezes both for the whole exploration phase. Silent in both cases.

Added `sanitize_mstep_damping`: non-finite or `<= 0.0` falls back to the calibrated default (not to a hair above zero — a near-zero cap is itself a frozen fit), `> 1.0` clamps to `1.0` (which is what any γ above 1 already meant), in-range passes through. The substitution is reported in `FitResult.warnings`. Unit test over the ranges, plus a fit-level test that drives `Some(-0.5)` through `api::fit` and asserts the warning names both the rejected value and the substitute.

### 2. The gate was "does NLopt have any free θ" (medium)

That is much wider than the shape #1011 was measured on. Where the closed-form mu-ref shift does not run — `mu_referencing = false`, identity-packed θ, a θ disqualified by a negative lower bound — `pinned` is empty, so *every* free theta was capped at 3% per M-step, including thetas that carry an ETA and are not the biased channel. At the default `n_exploration = 250` that is ~85 damped M-steps on a configuration none of the #1011 anchors cover.

`damps_numerical_mstep` now also requires the θ to carry **no ETA** — the same condition the no-ETA advisory warns on. `theta_has_eta_mask` is extracted out of `non_fixed_thetas_without_eta` so the gate and the advisory read one definition rather than two.

Effect on the anchors: the #1011 reprex is unchanged (`TVFRD1` carries no ETA, so it still damps, same γ, same 0.290). Fits whose free thetas all carry an ETA revert to the pre-#1011 assignment. New Tier-1 test fits a `mu_referencing = false` model whose every θ has an ETA and asserts `mstep_damping` reports no effect.

### 3–6. Docs and message fixes

- `damp_mstep`'s doc comment had been absorbed into `damps_numerical_mstep`'s (no blank line between the blocks), leaving `damp_mstep` undocumented and the gate's rustdoc opening with two paragraphs about a different function. Split, and the gate got its own doc for the new condition.
- `docs/model-file/fit-options.qmd` called `mstep_damping` the counterpart of `omega_burnin`. The counterpart is the per-iteration SA step cap on Ω, not the burn-in freeze — every other description in the PR says the cap. Fixed.
- The same row now states that `1.0` disables the damping in the convergence phase as well as exploration, since the option is otherwise described as an exploration-phase cap.
- The parser error hardcoded `` `mstep_damping` `` even when the user wrote `saem_mstep_damping`, and did not echo the offending value. Now uses `key` and appends `, got {v}`, matching `parse_f64_min` / `parse_unit_interval`. Test covers both.

### One extra, found while verifying finding 2

The CHANGELOG entry claimed ``warfarin_saem`` shifts OFV −286.00 → −285.83. It does not. All three of its thetas are log-mu-referenced, so `pinned` covers every θ and the gate was already off there before this change. An A/B run of the same model under both revisions (`mstep_damping = 0.5`, short schedule) gives identical estimates and the "has no effect" warning under each. The claim is corrected to state the fit is unchanged, and the entry now describes the narrowed gate and the clamp.

### Verification

`cargo fmt --check` clean · `cargo test --lib` 3034 passed, 0 failed · `cargo check --tests` clean · no new clippy warnings.

roninsightrx:
## Copilot review addressed — b0fa3f77

All three comments are the same point, and it is correct: the mask the #1011 gate consumes is mu-reference **detection**, not a scan of ETA usage, so documenting it as "carries an ETA" overpromises. An η attached in a form the parser cannot pair with a single θ — an additive `X = TVX + ETA_X`, or a covariate model whose typical value is not log-linear in one θ (#619) — reads as un-anchored and is damped.

**The behaviour is unchanged, and is the side to err on.** Such a θ has no closed-form shift either, so the numerical M-step really is its only mover — which is exactly the biased channel #1011 is about. Damping it is the intended outcome, and it matches what this branch did before the gate was narrowed, so there is no blast-radius change hiding behind the wording.

What was wrong was the naming, so that is what changed:

- `theta_has_eta_mask` → `theta_is_mu_ref_anchor_mask`, with its doc now stating the limit of the predicate outright: `true` means the parser paired this θ with an η it can move by a closed-form shift (the `TVX * exp(ETA_X)` and logit forms, plus an active class anchor), and an η in an unrecognised form reads `false`. It also records why that is right for both consumers, and that the cost is the #406 advisory's "no associated ETA" wording being stricter than the predicate — where the remedy it prints ("attach an ETA", i.e. put it in a mu-referenceable form) is the same either way.
- `damps_numerical_mstep`'s parameter is `theta_is_mu_ref_anchor`, and its rustdoc leads with the detection caveat rather than with "carries no ETA".
- The runtime "no effect" warning now says *every estimated theta is mu-referenced or `FIX`*.
- `docs/model-file/fit-options.qmd` says "Ignored (with a warning) when every estimated theta is mu-referenced or `FIX`" — directly the third comment.
- `docs/estimation/saem.qmd` and the CHANGELOG entry both switch to mu-referencing wording and spell out the additive / non-log-linear-covariate case explicitly.
- Test names and comments follow (`damps_numerical_mstep_only_for_a_free_unanchored_theta`, `mstep_damping_is_inert_when_mu_referencing_is_off_but_every_theta_is_anchored`).

Verification unchanged: `cargo fmt --check` clean · `cargo test --lib` 3034 passed, 0 failed · `cargo check --tests` clean.

PR #1015, opened 2026-08-21
refactor(estimation): extract fixed-eta observation-NLL gradients from saem

dominic-irx:
## Why

`obs_nll_subject_into` / `obs_nll_subject_grad` / `obs_nll_sum` and friends — the fixed-η
observation likelihood and its `∂/∂(θ, σ)` — lived inside `saem.rs`, reachable only by SAEM. Two
separate pieces of work need exactly that evaluator: the analytic `[covariate_nn]` weight
gradients (FeRx-NLME/ferx-core#TBD, DCM line) and variational inference (VI line), whose ELBO
data term *is* the fixed-η observation NLL.

Extracting it is a prerequisite for both, so it is proposed on its own rather than buried in
either.

## What changed

`src/estimation/fixed_eta_gradient.rs` (+ its sibling `_tests.rs`) now hold those functions;
`saem.rs` calls them. Pure code movement — no signature, no numerics, no behaviour change.

## Alternatives considered

Leaving them in `saem.rs` and having the other consumers `use` them from there. Rejected because
it makes `saem` a de-facto utility module for estimators that have nothing to do with stochastic
approximation, and because the SAEM file is already ~4k lines.

## Cross-repo dependency

## Breaking changes
- [x] None

The extracted items are `pub(crate)`.

## Numerical validation
- [ ] Reference values

Code movement with no numerical change. The 4 relocated FD-parity tests
(`obs_nll_subject_grad_matches_obs_nll_sum_fd`, `..._iov_matches_fd`, `..._per_cmt_matches_fd`,
`..._block_sigma_cross_endpoint_matches_fd`) move with the code and still pass, which is the
check that the move preserved the gradients.

## Performance
- [ ] Faster / Slower

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes) — the existing tests move
      alongside the code
- [ ] Regression test added that fails without this fix
- [ ] No new tests needed

`cargo test --lib`: **3024 passed, 0 failed**. `cargo test --lib saem`: 94 passed.

## Docs
- [ ] `CHANGELOG.md` `[Unreleased]` entry added, **or N/A (internal/refactor/CI)** ← this one
- [x] No user-visible change

## Checklist
- [x] `cargo clippy` clean (no new findings in the moved code)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — not touched; this is the
      fixed-η path, not `sens/`

### ferx-site

### ferx-book
- [x] No estimator, DSL feature, or fit option change

### Example execution

## Reviewer hints

The diff is large but mechanical: ~1050 lines leave `saem.rs` and reappear in
`fixed_eta_gradient.rs`. The thing worth actually reading is that nothing changed on the way —
`git show --stat` plus a spot check of one function should be enough.

One deliberate, non-mechanical detail: this branch does **not** add `vi: None` to `saem.rs`'s
`OuterResult` initializer. An earlier draft of this commit did, which made it depend on a field
the VI branch introduces and left both unbuildable in isolation. That line now lives with the
commit that adds the field.

## Open questions

None.


dominic-irx:

roninsightrx:
## Code review (high effort) — and a rebase that was required to land it

**Verification first.** I reconstructed the move mechanically rather than trusting the diff: every removed line from `saem.rs` matched the added lines in `fixed_eta_gradient.rs` byte-for-byte, modulo three `fn` → `pub(crate) fn` bumps, the new module doc header, and two banner-comment lines. The relocated tests were identical modulo rustfmt. `cargo check --lib --tests` was clean with default features and with `markov,nn`.

So the "pure movement, no behaviour change" claim was true **against the branch's original base** — and that turned out to be exactly the problem.

---

### 🔴 HIGH — the extracted file was a pre-#1029 snapshot; resolving the conflict the obvious way would have silently reverted the residual-magnitude feature for SAEM

The branch forked at `ae6c208c`. Since then `main` landed `46f6f2ea` (*apply the residual magnitude in every estimator*, #484/#1029), which modified **all three** moved functions:

| function | main `saem.rs` | lines changed by #1029 |
|---|---|---|
| `obs_nll_subject_into_iov` | 601 | ~14 |
| `obs_nll_subject_grad_iov` | 685 | ~63 |
| `obs_nll_subject_grad` | 1204 | ~66 |

The extracted file had **zero** occurrences of `ruv_obs_mult`. The drift was `residual_variance_at` vs `residual_variance_at_scaled`, `dvar_df` vs `dvar_df_scaled`, `dvar_dlogsigma` vs `dvar_dlogsigma_scaled`, plus the whole `gaussian_obs_nll_scaled` θ-FD branch that carries the direct θ→variance channel.

The trap is in how the conflict presents. GitHub reported `CONFLICTING`, but the conflicts land **only in `saem.rs`** — the new file is an *add*, so it conflicts with nothing. Resolving by "drop the conflicted `saem.rs` hunks, keep the new file" compiles green and reverts the feature: any model with `weight = <expr>` or a θ-dependent magnitude would get an M-step objective **and gradient** computed against the unscaled variance → biased θ/σ, with no test failing to say so.

**Fixed** by rebasing onto current `main` and re-extracting the three functions from *main's* version. `main` had also landed #1031 (sample-size-weighted IOV kappa) and the #1012 M-step damping fixes on the same functions; taking main's blocks wholesale picks all of them up.

### 🟠 MEDIUM — two FD-parity tests added on `main` after the fork were left behind

`main`'s `saem.rs` gained `check_saem_mstep_matches_fd` (6654) and its two callers `obs_nll_subject_grad_weighted_error_matches_fd` (6754) and `obs_nll_subject_grad_theta_dependent_magnitude_matches_fd` (6788), all exercising the moved `obs_nll_subject_grad`. Leaving them in `saem.rs` would split the FD-parity suite for one function across two files, against both this PR's stated goal and the CLAUDE.md rule that a gradient change ships with its parity test.

**Fixed** — they move alongside the code. The sibling now holds all six tests.

### 🟡 LOW — broken intra-doc link

The module doc pointed at ``[`crate::estimation::vi`]``; no `vi` module exists yet. Harmless for plain `cargo doc` (the module is `pub(crate)`), but `--document-private-items` flags it, and the doc asserted a consumer that isn't in the tree.

**Fixed** — prose kept, link markup dropped, and the "both callers share one copy" line corrected while SAEM is still the only caller. The link returns with the VI commit, same reasoning this PR applies to the deferred `vi: None` field.

### 🟡 LOW (not changed) — `obs_nll_sum` widened to `pub(crate)` for a test

Its only non-`saem` consumer is the FD reference in `fixed_eta_gradient_tests.rs`. Production visibility widened crate-wide for a test-only need; a `#[cfg(test)] pub(crate)` or a `test_helpers` home would keep the surface honest. Left as-is — noting it rather than churning the diff.

---

### What's on the branch now

Re-extracted against `main` (`82e0e8f7`). The move is still provably pure — mechanically verified, not asserted:

- extracted production code is **byte-identical** to `main`'s `saem.rs` lines 590–889 and 1187–1409, modulo the three `pub(crate)` bumps
- the `saem.rs` diff against `main` is pure deletion plus exactly six added lines: the `fixed_eta_gradient` import, the dropped now-unused `obs_nll_subject_from_preds` import, the replacement banner, and `pub(crate) fn obs_nll_sum`
- `cargo fmt`'s only delta was the same two call-site reflows the original commit described

Checks: `cargo check --lib --tests` clean (default and `--features markov,nn`); all 6 moved tests pass; 86 `saem` unit tests pass; full `cargo test --lib` → **3277 passed, 0 failed**; `cargo doc --document-private-items` reports nothing for this module. `cargo clippy` shows 4 pre-existing errors, all `src/sens/two_cpt_explicit.rs` PI-constant, untouched by this diff.

PR #1016, opened 2026-08-21
feat(nn): [covariate_nn] correctness and usability fixes for deep compartment models

dominic-irx:
> **Draft: stacked on #1015.** This branch currently contains #1015's commit as its base. Once
> #1015 merges, GitHub will drop it from this diff automatically. Review the five `nn` / `sens`
> commits; the `refactor(estimation)` one belongs to #1015.

## Why

Five defects and gaps in `[covariate_nn]` (deep compartment models), all found while fitting a
busulfan-shaped DCM, plus one performance path. Two of them changed the answer silently, which is
the reason this is a PR rather than a backlog item:

- **Time-varying NN inputs were frozen at baseline.** The network reads per-event covariates like
  every other consumer, but its input names are declared in the block rather than in an
  expression, so nothing registered them as *referenced*. The fit pipeline then pruned their
  trajectories as irrelevant and the network saw one constant value per subject for the whole
  record — no error, no warning, a time-varying covariate simply having no effect.
- **Every output-layer bias started at 0.** With a `softplus` head that starts *every* PK
  parameter at `softplus(0) = 0.693` — a 0.69 L volume regardless of what the parameter means.
  On a 60-subject busulfan-shaped DCM with identical data and seed, the bare head converged to
  `−2 log L` 2033.6 with the clearance decline understated 43%, and **reported `converged: true`**.

## What changed

| Commit | |
|---|---|
| `fix(parser)` | register `[covariate_nn]` inputs as referenced covariates |
| `test(nn)` | pin rejection of NN inputs the data does not carry |
| `feat(nn)` | `center` / `scale` — per-input normalization |
| `feat(nn)` | analytic weight gradients in the fixed-eta path |
| `feat(nn)` | `init` — start the network at plausible values |
| `perf(sens)` | DCM+IOV eta-gradients analytically rather than by finite differences |

`center` / `scale` exist because raw covariates are badly scaled for a network: `WT ≈ 70`
saturates a `tanh` layer at initialization, where its derivative is ~0 and the layer is nearly
blind to its input. On a two-covariate DCM, unnormalized inputs pushed weights to ~1e11 and left
the residual error 15× too high; standardized inputs kept every weight inside `[-1.3, 1.6]`.

All of it sits behind the existing `nn` cargo feature — default builds are unaffected.

## Alternatives considered

For `center` / `scale`: estimating the normalization from the data rather than declaring it.
Rejected on two grounds — `predict()` must apply the same transform the fit used, and recomputing
statistics on new data would silently change the model; and a declared constant keeps the model
file a complete description of the transform, as `(WT/70)^0.75` already is.

For `init`: emitting the value approximately (biases only). Rejected in favour of zeroing the
output-layer weight block alongside the biases, so the network emits exactly `init` for every
subject at iteration 0 whatever its covariates, and those weights still take gradient from the
first step.

## Cross-repo dependency

No public Rust API change; `[covariate_nn]` gains two optional keys in the model file.

## Breaking changes
- [ ] `.ferx` model file format — `center` / `scale` / `init` are **optional and default to the
      identity / previous behaviour**, so existing models are unchanged
- [ ] Public Rust API
- [x] None

## Numerical validation
- [ ] Warfarin example converges and estimates are within tolerance of prior run — warfarin has
      no `[covariate_nn]`; not the relevant fixture
- [x] Compared against: **prior ferx commit** on a busulfan-shaped DCM (same data, same seed)
- [x] Reference values:

```
# init, 60-subject busulfan-shaped DCM, identical data and seed
bare softplus head   -2 log L 2033.6   clearance decline understated 43%   converged: true
init = [1.0, 10.0]   -2 log L 1061.0   decline recovered

# center / scale, two-covariate DCM
unnormalized inputs  weights ~1e11      residual error 15x too high
standardized inputs  weights within [-1.3, 1.6]
```

- [ ] Not applicable

No NONMEM comparison: NONMEM has no neural-network structural model, so there is no equivalent
run to anchor against. The anchor used instead is a prior ferx commit on the same data and seed,
plus the FD-parity tests below.

## Performance
- [x] Faster — `perf(sens)`: DCM+IOV eta-gradients are served analytically instead of by finite
      differences. Not separately benchmarked; the win is `O(1)` model evaluations per subject
      instead of `O(n_eta + K·n_kappa)` central differences.
- [ ] No significant impact expected / Slower

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib --features nn` passes)
- [x] Regression test added that fails without this fix — the referenced-covariate fix is pinned
      by a test asserting the NN's inputs survive `prune_irrelevant_tv_covariates`
- [ ] No new tests needed

`cargo test --lib --features nn`: **3084 passed, 0 failed**. `cargo test --lib` (default
features): 3024 passed, 0 failed.

The analytic weight gradients carry `Dual2`-vs-FD parity tests in
`src/estimation/nn_theta_gradient_tests.rs`, per the CLAUDE.md rule that a wrong sensitivity
compiles and runs silently.

- [ ] Not applicable

<details>
<summary>Example</summary>

```toml
[covariate_nn]
  inputs = [WT, ZCOV]
  center = [70.0, 0.0]     # per-input normalization: the network sees (x - center) / scale
  scale  = [20.0, 1.0]
  outputs = [CL, V]
  init = [1.0, 10.0]       # one starting value per output, on the parameter's own scale
```

Without `init`, a `softplus` head starts every output at `softplus(0) = 0.693` — see the
reference values above for what that costs.

</details>

## Docs
- [x] `docs/` updated for user-visible changes — `docs/model-file/covariate-nn.qmd`
- [ ] New pages linked in `docs/_quarto.yml` — no new pages
- [ ] No user-visible change

**Known gap:** two small hunks (10 + 8 lines) documenting DCM behaviour in
`docs/estimation/vi.qmd` are *not* in this PR. That page is created by the VI branch, so those
lines cannot live on either branch alone. They will follow as a small docs PR once both are on
`main`.

## Checklist
- [x] `cargo clippy` clean
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `sens/provider.rs` is
      touched by the `perf` commit and its FD-parity tests pass

### ferx-site
- [x] New `[covariate_nn]` keys (`center` / `scale` / `init`) → **needs a ferx-site update**; not
      yet opened
- [x] No new example `.ferx` file or data file

### ferx-book
- [x] New DSL keys → **may need a book chapter update**; not yet opened

### Example execution
- [ ] Rebuilt ferx-core / ferx-r, ran affected examples, rendered site and book pages
- [x] **Not done in this PR.** No `examples/*.ferx` uses `[covariate_nn]` (it is feature-gated and
      off by default), so there is no affected example to run. Flagging rather than silently
      checking the box.

## Reviewer hints

Focus on `fix(parser): register [covariate_nn] inputs as referenced covariates` — it is the
smallest commit and the one that was silently changing fits. The rest are additive and
feature-gated.

`src/parser/model_parser_tests.rs` will look alarming in the diff: this branch was rebased onto
73 commits of upstream, and upstream's `[mixture]` tests were appended at the same end-of-file
position as these NN tests. Both are kept; nothing was replaced.

## Open questions

- Should `center` / `scale` warn when left at the identity on an input whose magnitude is far
  from 1? The failure mode is quiet and the fix is mechanical, so a diagnostic may be worth more
  than the docs note.


dominic-irx:

roninsightrx:
## Review — high effort

Reviewed `95ac28e3..2d31b11a` (the five `nn` / `sens` commits plus the changelog fix), ~2.7k lines. I rebased the branch onto `main` first (#1015's merged form differs from the commit that was on this branch), so line references are against the pushed head.

The core math checks out. The `∂NLL/∂w = Σₖ (∂NLL/∂zₖ)(∂zₖ/∂wⱼ)` decomposition is sound, the pre-activation bias seed genuinely avoids the `0/0` on a saturated head, the `±sign` one-sided quotients compose into a central difference at the caller's own step, `bytecode_is_dynamic` correctly marks `PushNnOutput` dynamic so the cov-static fold cannot silently freeze an NN slot, and the empty `dtheta` / `d2eta` blocks from the η-only builder really are never read by `run_obs_iov_eta`.

Seven findings, roughly in severity order.

### 1. `src/sens/provider.rs:1375` — the relaxed predicate is not wired into the gate that actually routes the inner loop

`iov_analytical_eta_supported` is consulted only inside `subject_eta_grad_iov_analytical` (`provider.rs:1977`). But every path that *calls* `subject_eta_grad_iov` gates first on `iov_sens_supported`, which for a closed-form model is `iov_analytical_supported` — the **strict** predicate:

- `inner_optimizer.rs:1230` (`analytic_iov_inner`)
- `inner_optimizer.rs:186` (`iov_inner_subject_route`, which also feeds `resolve_gradient_method_for_reporting`)

For a closed-form DCM+IOV model those still return `false`, so the inner EBE gradient stays on FD and `n_fd_subjects` still counts every subject. The CHANGELOG's "`n_fd_subjects` 60 → 0, 336 s → 162 s" is not reproducible from this branch through the inner loop.

The one in-repo caller that does reach the new path is `agq.rs:1301`, which calls `analytic_eta_nll_gradient_iov` without the `iov_sens_supported` gate — so AGQ silently switches route while `iov_inner_subject_route` reports the same subject as FD. That is exactly the route/report drift the #637 guards exist to prevent.

Also worth noting: the relaxation is not NN-specific. Any model whose `prog.n_theta_axis() != model.n_theta` newly takes the analytic η-only path via AGQ.

Either widen `iov_sens_supported` (or add an η-only sibling and use it at the two inner-loop gates), or scale back the CHANGELOG claim to what the branch actually delivers.

### 2. `src/parser/model_parser.rs:14310` — `ModelNnGuard::enter_for` swallows a `forward_raw` failure

```rust
nn.mapper.forward_raw(w, covariates).unwrap_or_default()
```

yields `vec![]` for that block. `Op::PushNnOutput` then falls through to `unwrap_or_else(|| { debug_assert!(false, ...); 0.0 })`, so in a **release** build every NN output reads as `0.0` and the η-gradient is taken against a network that emits nothing — no error, no warning. Production's own `pk_param_fn` (line ~14070) `.expect()`s the identical call with a message saying it can only be a wiring bug.

This is the failure mode the guard's own doc comment says it exists to prevent. Either propagate `None` (fall back to FD) or panic consistently with the production path.

### 3. `docs/model-file/covariate-nn.qmd:255` — link to a page that does not exist

`[variational inference](../estimation/vi.qmd)` — `docs/estimation/vi.qmd` is not in the repo (the VI branch creates it; confirmed absent from `docs/estimation/`). Quarto resolves `.qmd` cross-references against the project and errors on an unresolvable target, so `Deploy Docs` will fail on push to `main`. The commit immediately before this branch (`e3fad47d docs(estimation): drop the module link to the unlanded vi module`) removed exactly this class of link. The PR body already flags two other `vi.qmd` hunks as deliberately withheld, so this one reads as an oversight.

### 4. `src/nn/mod.rs:708` — a non-finite `center` is reported as a `scale` error on an input named "center"

```rust
if center.iter().any(|c| !c.is_finite()) {
    return Err(NnError::InvalidScale { input: "center".to_string(), value: f64::NAN });
}
```

formats as ``scale` for input 'center' must be finite and non-zero; got NaN``. A user who writes `center = [inf, 0]` is told the wrong key, a fabricated input name, and a value they never typed. Mirror the `scale` loop: report the offending index's real input name and the actual value, under a `center`-specific variant.

### 5. `src/parser/model_parser.rs:17017` — per-call clone of the ambient NN outputs in the hottest evaluator

`eval_statements_g` now does `with_nn_outputs(|o| o.to_vec())` on every invocation, including each recursive descent into an `If` body, on a path the FOCE inner loop hits millions of times. With a live guard that is a `Vec<Vec<f64>>` deep clone (one heap allocation per block plus the outer vec) per statement-list evaluation; with no NN it still costs a `RefCell` borrow and a `Vec` construction. Hoist the fetch to the non-recursive entry points, or borrow at the two `eval_bytecode_g` / `eval_condition_indexed` call sites, rather than cloning per frame.

### 6. `src/parser/model_parser.rs:14284` — dead-code warning on the default feature set

`ModelNnGuard::enter` is called only from the `#[cfg(feature = "nn")]` arm of `enter_for`, so a default build warns — verified with `cargo check --lib`:

```
warning: associated function `enter` is never used
  --> src/parser/model_parser.rs:14284:19
```

The checklist's "clippy clean" presumably came from a `--features nn` run. Gate `enter` on the feature, or fold it into `enter_for`.

### 7. `src/types.rs:5287` — `NeuralNetworkInfo` does not record `center` / `scale` / `init`

The struct carries shape, activations, weight count and offset, and input/output names. `docs/model-file/covariate-nn.qmd:207` states ferx-r round-trips this losslessly via `fit.json` inside the `.fitrx` archive. Now that normalization materially changes what the network computes, a consumer reconstructing the network from the fit output sees an identity transform: the reported weights are meaningless without the `(x − center)/scale` they were fitted under, and nothing in the fit YAML says the inputs were standardized. Add the vectors to `NeuralNetworkInfo` and the YAML summary in the same PR that introduces the keys.

### Checked and cleared

For the record, these looked suspicious and are fine: the `theta_inits` indices used by `init` match the `W`-then-`B` per-layer flat layout and `offsets[l-1]`; NN weight thetas are unbounded so `theta_packs_log` is false and zeroing the output weight block cannot produce `ln(0)`; `static_nn_input_map`'s `Option<f64>` comparison correctly declines the mixed present/absent case; `Softplus::invert`'s piecewise branches round-trip through `apply`'s three-way split within the tests' tolerance; `NamedMlpMapper` has no serde derives, so the new fields break no on-disk format.

### Sync note

The branch was rebased onto `main` and force-pushed. Three conflicts were resolved: the `model_parser_tests.rs` end-of-file collision with upstream's `[mixture]` tests (both kept, 587 added / 0 removed), `fixed_eta_gradient.rs` where the #484/#1029 magnitude branch had to move inside the new `signed_theta_deriv` closure, and `model_parser.rs` where `ModelNnGuard` met `main`'s `pub(crate) current_model_time()`. One commit was added on top: the three CHANGELOG entries had been filed inside the released `## [0.3.0]` section (pre-existing on the branch, not a rebase artefact) with `init` duplicated verbatim — moved to `[Unreleased]`, deduplicated, text unchanged.

`cargo test --lib --features nn`: 3337 passed, 0 failed. `cargo test --lib`: 3277 passed, 0 failed. `cargo clippy --features nn` and `cargo fmt --check` clean.

roninsightrx (changes requested):
## Follow-up review

One blocking correctness finding remains after the previous review fixes.

### [P1] Decline the shortcut when an NN weight has a direct expression path

`NnGradPlan::build` marks every coordinate in each NN weight block as covered, relying on the invariant that NN weights reach the likelihood only through the network output. That invariant is not enforced: `model_parser.rs` extends `theta_names` with the generated `W_…` / `B_…` names before parsing `[individual_parameters]`, so this is accepted:

```ferx
CL = TYPICAL_PK.CL * exp(ETA_CL) + B_TYPICAL_PK_2_1
```

On a static-covariate DCM the plan still builds. A local regression probe against central FD produced:

```
theta[1]: hybrid = -1.59675490e-1
          central = -6.37932981e-2
          relative error = 6.00e-1
```

For a directly referenced output bias, its extra direct derivative is folded into `dNLL/dz` and then incorrectly spread through `jz` to the whole block. For a directly referenced non-bias weight, the direct term is omitted entirely because only output biases are perturbed. This silently gives SAEM/IMP/VI a wrong fixed-eta theta gradient.

Please either reject direct references to generated NN theta names at parse time, or record that dependency and make `NnGradPlan::build` decline to the existing per-theta FD path. Add a central-FD regression test for the accepted syntax.

roninsightrx:
Fixed in 6291451b.

**The finding is right, and I reproduced it exactly** — with the guard disabled, the new regression test prints `theta[1]: got -1.59675490e-1, central FD -6.37932981e-2`, digit for digit the probe in the review.

Both halves of the premise check out in the code:

- `theta_names` really is extended with the generated `W_…` / `B_…` names at `model_parser.rs:1394-1398`, and `[individual_parameters]` parses at ~1442 — after. So the syntax resolves.
- `NnGradPlan::build` marks the *entire* weight run covered (`for c in covered.iter_mut().take(end).skip(nn.weights_offset)`) while only `bias_theta_idx` is finite-differenced, and declines for exactly four things — no NN block, time-varying inputs, an out-of-range run, a failed Jacobian. Nothing about direct references.

**Took the decline branch, not parse-time rejection.** CLAUDE.md's rule is that a scope gap routes to FD through a support predicate rather than erroring, and it is the rule this module's own doc cites for the time-varying case — so the model still fits, just on the per-theta path. It also keeps the central-FD regression on genuinely *accepted* syntax, which parse-time rejection would have made impossible.

Three notes on how:

1. **Detection sits at the single point in `parse_atom` where an identifier becomes a theta index**, so a direct reference is caught from any block — `[scaling]`, `[odes]`, `[derived]` — not only from the one a dedicated walker remembers to visit.
2. **The flag rides in `CompiledModel::parse_warnings`**, following the `W_ABSORPTION_TWIN_DECLINED` precedent, rather than a new public field: `CompiledModel` is not `#[non_exhaustive]`, so a field would be a source break for external struct literals.
3. **A latent second error, fixed too:** the comment above `build_pk_param_fn` asserted that NN-weight theta "aren't visible to user expressions". They are — and that belief is presumably why the invariant was left unguarded. Corrected in place.

Tests: the syntax parses and is recorded; a plain DCM model is **not** flagged (or the shortcut would never run at all); the plan declines for a bias *and* a non-bias reference — the latter being the variant where the term is dropped rather than misattributed; and the theta gradient matches central FD for both. I verified the last one fails without the fix rather than trusting that it passes.

`cargo test --lib --features nn`: 3411 passed. `cargo test --lib`: 3333 passed. Clippy clean on the touched regions.

PR #1017, opened 2026-08-21
feat(estimation): variational inference (`method = vi`), with its validation

dominic-irx:
> **Draft: stacked on #1015.** This branch contains #1015's commit as its base; GitHub will drop
> it from this diff once #1015 merges. 29 commits, developed as one line — happy to split further
> if that helps review.

## Why

A third way to marginalize the random effects, alongside profiling them out at their mode
(FOCE/Laplace) and sampling them (SAEM/IMP). VI fits a tractable posterior `q(η)` per subject and
optimizes its parameters jointly with θ/Ω/Σ by maximizing the evidence lower bound, with no inner
loop. It suits models whose fixed-effects part is highly flexible (deep compartment models), cases
where FOCE's inner optimization is unstable, and anyone who wants a per-subject posterior
*covariance* without paying for a Hessian.

Following Janssen et al. (2024), with two departures: the KL term is taken in closed form rather
than by Monte Carlo, and Ω is then updated by its exact maximizer rather than by gradient descent,
which removes the Ω instability that paper reports.

## What changed

`method = vi`, plus `vi_family`, `vi_kl`, `vi_omega_update`, `vi_sigma_update`, `vi_iters`,
`vi_mc_samples`, `vi_lr`, `vi_avg_last`, `vi_eta_grad`, `vi_final_ofv`, `vi_seed`. IOV
(`Ω_iov`, stacked prior, two-block closed-form Ω) is supported end to end. `FitResult$vi` carries
the per-subject posterior, and a `method = vi` fit writes an `eta_posterior` block to the fit YAML.

Also here, and the reason the branch is long: three validation harnesses and the record they
produced (`VI_VALIDATION.md`), an `agq_eval_only` mode used as the arbiter, and two defects those
harnesses found and fixed.

## Alternatives considered

- **Sampled KL vs closed form.** Both are implemented (`vi_kl`); the closed form is the default
  because it removes the KL's Monte-Carlo noise entirely. `vi_kl = mc` is retained to reproduce
  the published behaviour and because a family with no closed-form KL will need it.
- **Gradient-stepped Ω vs its exact maximizer.** Both implemented (`vi_omega_update`). The closed
  form is default; Janssen et al. report Ω fluctuating badly under gradient descent (their Fig. 3)
  and the closed form removes that by construction rather than by tuning.
- **Adam-stepped σ vs its exact maximizer** (`vi_sigma_update`, new here). Closed form is default
  for exactness and consistency with Ω — **but see the honest note under Numerical validation: it
  is not what closes the warfarin gap.**

## Cross-repo dependency
| ferx (R) | — | not needed for this PR | — |

`FitResult` gains a `vi` field. ferx-r reads `FitResult` but the field is additive and optional;
surfacing `fit$vi` in R is a follow-up, and `ferx-r`'s `Cargo.lock` will need the usual bump via
`tools/update-ferx-core-lock.sh` before its CI sees this.

## Breaking changes
- [ ] `.ferx` model file format — additive: new `method` value and new optional `[fit_options]` keys
- [x] `FitResult` / sdtab fields changed — **additive**: `FitResult.vi: Option<ViResult>`, and a
      new `vi:` block in the fit YAML. Nothing existing renamed or removed.
- [x] Public Rust API (`types.rs`) — new `ViResult`, `ViFamily`, `ViKl`, `ViOmegaUpdate`,
      `ViSigmaUpdate`, `ViFinalOfv` types and `FitOptions` fields
- [ ] None

## Numerical validation
- [x] Warfarin example converges and estimates are within tolerance
- [x] Compared against: **ferx AGQ (`n_agq = 9`), ferx FOCEI, nlmixr2 7.0 FOCEI, nlmixr2 `emvi`,
      and per-subject NUTS (numpyro)**
- [x] Reference values — warfarin, 10 subjects, 110 obs, 1-cpt oral, lognormal η on CL/V/KA,
      proportional error:

```
                              TVCL       TVV      TVKA     sigma        OFV
ferx AGQ (n_agq = 9)      0.132687   7.73746   0.81090  0.010565   -285.977
ferx FOCEI                0.132695   7.73771   0.81080  0.010565   -286.004
nlmixr2 FOCEI             0.132738   7.73851   0.82849  0.010574   -285.947
ferx VI (mc = 128)        0.132692   7.73795   0.81097  0.011234   -285.425
nlmixr2 emvi (20k iters)  0.132625   7.73894   0.81124  0.010369   -284.755
```

The FOCEI pair also pins the objective *convention*: ferx omits `½·n_obs·log 2π` and nlmixr2 has
`adjObf` defaulting to TRUE; the measured offset is **zero** (a mismatch would show as
`110·log 2π ≈ 202`).

**Per-subject posterior against NUTS** (4 chains × 20 000 draws, 0 divergences, population
parameters FIXed at the AGQ estimate so the comparison is about `q` alone):

```
median over 10 subjects        eta CL    eta V   eta KA
|VI mean - NUTS mean|          1.6e-5   2.1e-5   2.7e-5
VI var / NUTS var               1.002    0.999    1.002
max |VI corr - NUTS corr|      0.0042   0.0062   0.0077
```

- [ ] Not applicable

**Three things I would rather state than have a reviewer find:**

1. **`vi_mc_samples = 8` (the default) does not converge on warfarin.** It returns σ 34% high with
   an OFV 11.3 units short of AGQ, reporting `converged: true`. The convergence rule stops when
   the ELBO's drift falls below its own Monte-Carlo noise floor, and at 8 draws it does so while
   real drift remains. 32 draws → −284.6, 128 → −285.4. Documented in `vi.qmd`; the **default is
   unchanged** pending a check against the DCM models it was presumably tuned for.
2. **The `vi_sigma_update` closed form is exact but is not what fixes that.** At 128 draws the two
   σ routes agree to 0.3 OFV with `adam` marginally ahead. It is adopted for exactness and
   consistency with Ω, not for a measured improvement. An earlier draft of this branch attributed
   the warfarin gap to a σ-specific step-size pathology; that attribution was wrong and the
   correction is recorded in `VI_VALIDATION.md` §4.11a.
3. **The documented 20–25% posterior-variance understatement does not reproduce here**, and this
   PR corrects `vi.qmd` to say so. Measured: 0.2% on warfarin, 4% thinned to two observations a
   subject. The reason is measurable rather than lucky — the Laplace covariance matches NUTS to
   0.1%, so the true posterior *is* Gaussian and a Gaussian `q` has nothing to get wrong. The
   20–25% figure comes from deep compartment models; the page now scopes it to that regime.

## Performance
- [x] No significant impact expected on existing estimators — VI is a new code path; the shared
      `fixed_eta_gradient` evaluator (#1015) is unchanged in behaviour
- [ ] Faster / Slower

## Tests
- [x] Regression test added that fails without this fix — several, notably
      `vi_recovers_the_agq_solution_on_warfarin` (Tier 3, `--features slow-tests`), which fits AGQ
      *in the test* and asserts both σ routes reach it
- [ ] No new tests needed

`cargo test --lib`: **3129 passed, 0 failed**. `cargo test --features slow-tests --test vi`:
12 passed, 0 failed.

Includes the `Dual2`-vs-FD parity the CLAUDE.md rule requires for anything on the gradient path:
`neg_elbo_gradient_matches_central_fd` covers the whole packed vector, and the σ M-step is pinned
against the vanishing σ-gradient from three starting values — exactly, not to a tolerance, because
the draws are seeded on `(seed, iter, subject, sample)` and do not depend on σ.

- [ ] Not applicable

<details>
<summary>Example</summary>

```toml
[fit_options]
  method = vi
  vi_family = full_rank
  vi_mc_samples = 128      # 8 (the default) does not converge on this model -- see vi.qmd
  vi_final_ofv = laplace   # otherwise `ofv` is NaN: the ELBO is a bound, not an OFV
```

```
OFV: -285.4251                     # Laplace objective at the VI estimate; AGQ gives -285.977
  PROP_ERR = 0.011234              # AGQ / FOCEI give 0.010565
neg_two_elbo: -285.242259          # a bound on -2 log L, not comparable with an OFV
elbo_tightness_ratio: 1.007        # ~1 is healthy; >25 warns
```

</details>

## Docs
- [x] `docs/` updated — `docs/estimation/vi.qmd` (substantially), `docs/model-file/fit-options.qmd`
- [ ] No user-visible change

## Checklist
- [x] `cargo clippy` clean
- [x] Functions on the `Dual2` sensitivity path stay differentiable — VI consumes `sens/` rather
      than modifying it, and the ELBO gradient carries its own FD parity test
- [ ] `Cargo.toml` version bumped if breaking — additive only

### ferx-site
- [x] New `method = vi` and `vi_*` `[fit_options]` keys → **needs a ferx-site update**; not yet opened
- [x] No new example `.ferx` file or data file

### ferx-book
- [x] New estimator → **needs a book chapter**; not yet opened

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core / rendered site and book pages
- [x] Rebuilt ferx-core and ran the VI models via the CLI (`tools/vi-emvi-comparison/`,
      `tools/vi-nuts-anchor/`), which is how the tables above were produced
- [x] No `examples/*.ferx` uses `method = vi`, so there is no affected example page to render.
      ferx-r was **not** rebuilt against this branch — flagging rather than checking the box.

## Reviewer hints

**Where to focus.** `src/estimation/vi/elbo.rs` (the objective and its gradients) and
`src/estimation/vi/run.rs` (the optimizer loop, the two closed-form M-steps, and the convergence
rule). Everything in `tools/` is out-of-tree validation harness and can be skimmed.

**What is subtle.** The convergence rule has two criteria and *either* is sufficient. One commit
here fixes a case where that was wrong: with every population parameter FIXed — a legitimate
request, it is how you read per-subject posteriors at a known estimate — the parameter-stability
test compared a constant vector against itself, reported settled immediately, and stopped the fit
after 500 iterations with `elbo_tightness_ratio: 78` and `−2·ELBO = +2026` on a model that reaches
`−283`. It now runs 6250 iterations to `−282.6` at a ratio of 1.4.

**`VI_VALIDATION.md` is committed with its retractions intact**, deliberately. Its first run drew
a conclusion that two implementations agreeing on a ~10 OFV gap made that gap "a property of VI
rather than a defect in either"; it was two defects, and the section stands as written with the
correction beside it. The standing lesson is in §4.11a and it shaped the harnesses: co-validation
localizes a disagreement but cannot orient one.

## Open questions

- **Should `vi_mc_samples` default rise from 8?** It is demonstrably too low on warfarin. I have
  not changed it because I cannot check it against the deep compartment models it was presumably
  tuned for, where draws are expensive.
- **Should the parameter-stability convergence criterion see `φ`?** The fix here only disables it
  when it is *vacuous* (nothing can move). The broader version — folding `φ` into the stability
  measure — would be stricter in general, but `φ` is high-dimensional and per-subject, so a max
  over it could stop the criterion from ever firing in the flat-parameter-space case it exists for.
- One defect found by this work is **not** fixed here and is outside VI: ferx's FOCEI does not
  honour `block_omega` structural zeros (it estimates the full lower triangle and reports
  `n_parameters = 10` where an honoured mixed Ω is 8). Recorded in `VI_VALIDATION.md` §4.13. Happy
  to open it as an issue.


dominic-irx:

roninsightrx (changes requested):
Requesting changes for the following correctness and integration issues:

1. **[P1] The documented VI → AGQ evaluator chain is rejected.** In `src/api/validation.rs:2246`, `n_agq > 1` is rejected whenever the chain contains `Vi`, even when the same chain ends in `Laplace` with `agq_eval_only = true`. That makes the recommended `methods = [vi, laplace]`, `n_agq = 21`, `agq_eval_only = true` path impossible. Please make this check stage-aware (reject `n_agq` only when no quadrature stage consumes it) and add a VI+Laplace regression test.

2. **[P1] The default configuration is known to return a materially wrong fit as converged.** `FitOptions::default()` sets `vi_mc_samples = 8` (`src/types.rs:6199`), while the PR description records that this produces sigma 34% high and an OFV 11.3 units short on warfarin with `converged: true`. A new estimator should not silently certify its canonical example at a known-bad default. Please either raise the default to a validated value or strengthen convergence/diagnostics so this run cannot be reported as successful.

3. **[P1] Parameter-only early stopping can still terminate while the variational posterior is moving.** `run.rs:617-650` measures stability only on packed population parameters `x`; the all-FIXed guard handles the fully vacuous case, but if even one population coordinate is free and settles before `phi`, `param_settled` alone schedules a stop. Since either criterion is sufficient, a still-changing per-subject posterior can be returned as converged. Include `phi` in this criterion (with a suitable aggregate), or require an independent objective/posterior check before it can certify convergence.

4. **[P2] `FitResult.vi` becomes stale after a later estimating stage.** `src/api/fit.rs:1562-1564` saves the last VI result and `fit.rs:2195` attaches it unconditionally to the final fit. For `methods = [vi, focei]`, the final theta/Omega/sigma and subject diagnostics come from FOCEI, but `vi.eta_means`, `eta_covs`, ELBO, and tightness still describe the pre-FOCEI parameter point, with no marker that they are historical. Preserve VI output only across trailing evaluation-only stages, or attach the parameter snapshot/stage identity needed to interpret it.

5. **[P2] Correlated residual models are rejected before the documented Adam fallback can run.** `closed_form_sigma_support` explicitly returns a fallback reason for `residual_correlations`, and the public option docs say this configuration falls back to Adam, but the allow-list in `src/api/validation.rs:2554-2573` omits `EstimationMethod::Vi`, so every `block_sigma` + VI fit fails validation. Either add VI to the supported non-IOV dense-residual path with a regression test, or remove the contradictory fallback claim.

6. **[P2] The new YAML block does not safely serialize subject IDs.** `src/io/output.rs:2131` writes `id: {id}` without quoting/escaping, unlike the nearby importance-sampling block. Valid string IDs such as `001`, `a: b`, `#x`, or values containing quotes can change YAML type or make the file invalid, defeating the promise that the posterior is keyed by the original subject ID. Serialize the ID as a proper YAML string and test a special-character/numeric-looking ID.

7. **[P2] The user-facing fit-options page still documents the old implementation.** `docs/model-file/fit-options.qmd:233-274` says IOV is unsupported, VI always runs its full budget, and the defaults are 1000 iterations / 3 draws / 0.05 learning rate. The code and VI page implement IOV, early stopping, and defaults of 25000 / 8 / 0.02. Please update this table in the same PR so users do not configure the estimator from obsolete behavior.

Validation performed: `cargo test --lib estimation::vi` (101 passed, 1 slow test skipped) and `cargo test --test agq agq_eval_only` (5 passed). The existing AGQ tests cover standalone and FOCEI chains, but not the advertised VI+Laplace combination that finding 1 identifies.

roninsightrx:
Merged `main` into the branch and addressed all seven review findings. Two commits: `89fbb7fa` (merge) and `9e375c0e` (fixes).

## Merge

Two conflicts, both additive:

- `docs/_quarto.yml` — kept both sidebar entries (`estimation/vi.qmd` and main's new `estimation/parameterization.qmd`).
- `CHANGELOG.md` — main's bullets continue the existing `### Performance` section, so they were placed first; this branch's new `### Added` / `### Changed` / `### Fixed` sections follow.

Every code file auto-merged. `origin/main` is now an ancestor of the PR head.

## Findings

**1 (P1) — the VI → AGQ evaluator chain is rejected.** Fixed. `n_agq` is a *chain-wide* option, not a per-stage one, so a chain pairing VI with a quadrature stage legitimately carries `n_agq > 1` — the grid belongs to the Laplace stage and VI ignores it. The check now fires only when no stage consumes the option (`laplace` at any `n_agq`, `focei` at `n_agq > 1`), i.e. on a VI-only chain, where setting it really is a mistake worth reporting. Regression test `n_agq_is_accepted_when_a_quadrature_stage_consumes_it` covers `[vi, laplace]` with and without `agq_eval_only`, `[vi, focei]`, the VI-only rejection, and `n_agq = 1`.

**2 (P1) — the default configuration returns a materially wrong fit as converged.** Fixed by strengthening the diagnostic rather than by raising the default, since the default was deliberately deferred pending a check against the deep-compartment models it was tuned for.

The failure is specific: `trace_has_settled` stops the run when the ELBO's drift is no longer distinguishable from Monte-Carlo noise, and at a low `vi_mc_samples` that becomes true while real drift remains. The drift is *below* the per-window noise that test measures against, so no single window comparison can see it — that is the definition of the noise floor. What a noise-limited stop and a genuine plateau differ in is not the size of each step but its **sign**: noise alternates, real drift does not.

So `trace_still_drifting` splits the trace tail into 10 block means and counts how many consecutive steps move the same direction. That extracts a trend far below the amplitude of one step, at no cost (the trace is already in hand), and is deliberately one-sided and conservative — it needs 80% of steps to agree, which a fair coin clears on 9 steps under 5% of the time. When it fires, `converged` is set to `false` and a warning names `vi_mc_samples` and the measured recovery point (`32`). The parameter criterion is exempt: it exists for flat directions the objective can never settle on, where the trend is real but irrelevant.

Unit test `systematic_drift_is_detected_below_the_settling_test_noise_floor` pins all of it, including the assertion that the synthetic trace *is* below the settling test's noise floor — without which the test would not mean anything.

**3 (P1) — parameter-only early stopping can terminate while the posterior is moving.** Fixed. The criterion now judges `φ` alongside `x`, on the same window and the same tolerance, extracted into `estimates_have_settled` so it is unit-testable. `max_relative_change` is already a max over coordinates, so the max of the two calls is the same question asked over the concatenated vector.

The all-FIXed guard is kept even though the `φ` half now independently defeats that case: an `x` that cannot move must never *contribute* evidence of convergence, whatever `φ` happens to be doing on a given window. Test: `stability_requires_both_the_population_vector_and_the_posterior_to_be_still`.

**4 (P2) — `FitResult.vi` becomes stale after a later estimating stage.** Took the second of the two options offered — marking rather than dropping. Dropping would discard the block exactly on `methods = [vi, imp]` / `[vi, laplace]`, which is the recommended way to finish a VI fit, and the variational covariance is the only per-subject covariance in the result.

`ViResult` gains `superseded_by: Option<String>`, set in `fit_inner` (the only place that can see the rest of the chain — `run_vi` is handed a blanked per-stage option set). It uses the same `is_last_estimating_stage` predicate the loop uses, so trailing `imp_eval_only` / `agq_eval_only` evaluators do not count as superseding. It is emitted in the fit YAML only when set, so an ordinary VI fit's output is unchanged, and a warning says the same in prose. Test `test_vi_block_is_marked_when_a_later_stage_re_estimates` covers all three cases: VI alone, VI + evaluator, VI + FOCEI.

**5 (P2) — correlated residual models rejected before the Adam fallback can run.** Fixed by adding VI to the allow-list, with a test, rather than by removing the fallback claim — the fallback claim is correct. VI's data term is `obs_nll_subject_grad`, whose non-IOV path routes a dense `R` through the same full-FD fallback FOCE and SAEM use (`fd_all` when `residual_correlations` is non-empty), so the objective and its θ/σ gradient are right there. Only the *closed-form* σ maximizer is unavailable, and `closed_form_sigma_support` already declines it with a recorded reason and falls back to Adam. `block_sigma` + IOV stays rejected separately by `E_BLOCK_SIGMA_IOV_UNSUPPORTED`, so the IOV path is never reached.

New test `test_vi_accepts_block_sigma_cross_endpoint_fit` mirrors the existing SAEM and IMP cases in `iov_integration.rs`: check-time acceptance, a finite `−2·ELBO` from the dense-`R` data term, and the σ fallback warning actually firing.

**6 (P2) — subject IDs not safely serialized.** Fixed, routed through the existing `yaml_quote` used for the environment block. Test `vi_posterior_ids_are_quoted_and_escaped` covers `001`, `a: b`, `#x`, `true`, and an embedded quote, and asserts one row per subject so an ID swallowing the next would fail.

**7 (P2) — fit-options page documents the old implementation.** Updated. `vi_iters` 1000 → `25000` and reframed as a ceiling with early stopping (including what settling is judged on and which fields report the outcome); `vi_mc_samples` 3 → `8`, with the noise-floor explanation and the measured warfarin numbers; `vi_lr` 0.05 → `0.02`. The method section's "IOV and non-Gaussian endpoints are not supported" is now split: IOV *is* supported (`q` spans the stacked `[η, κ₁ … κ_K]` vector, per-occasion means under `vi$kappa_means`), non-Gaussian endpoints are not, and the model is rejected rather than silently mis-fitted.

## Validation

`cargo test --lib`: **3639 passed, 0 failed**, 23 ignored. `cargo check --tests` and `cargo clippy --lib` clean. Six new tests across the four fix sites, plus VI added to the existing `block_sigma` allow-list case.

One pre-existing item, unrelated to this PR: `cargo clippy --all-targets` reports 4 `approx_constant` errors in `src/sens/two_cpt_explicit.rs` (test fixtures using `3.14`). They are present on `origin/main` and untouched here.

Issue #1023, opened 2026-08-21
CI clippy never lints test code; `cargo clippy --all-targets` fails on main (4 false-positive approx_constant denials)


TeunP:
Noticed while running the local verification sweep for #1001. Not caused by that branch — reproduced on `main`.

## The gap

The clippy job lints the library target only:

```yaml
# .github/workflows/ci.yml:152
run: cargo clippy --no-default-features --features ci,markov
```

No `--all-targets` and no `--tests`, so **no `#[cfg(test)]` code and no `tests/*.rs` integration target is ever linted**. Given how much of this repo's logic lives in test modules — the `Dual2`-vs-FD parity harnesses, the NONMEM anchor comparisons, the sibling `*_tests.rs` files — that is a large unlinted surface.

## What it has been hiding

`cargo clippy --all-targets` fails outright on `main` today. `approx_constant` is `deny`-by-default, so these are errors, not warnings:

```
error: approximate value of `f{32, 64}::consts::PI` found
   --> src/sens/two_cpt_explicit.rs:477:39
    |
477 |             (1000.0, 1.0, 4.41, 15.5, 3.14, 29.3), // 2-cpt NONMEM-fit-ish
    |                                       ^^^^
```

Four hits, all inside the `#[cfg(test)] mod tests` that begins at `src/sens/two_cpt_explicit.rs:447` — lines 477, 527, 528 and 579. Last touched in `add7f17b`.

**The lint is wrong here.** In every one of those tuples `3.14` is the inter-compartmental clearance `Q` of a "NONMEM-fit-ish" 2-compartment parameter set, sitting alongside `CL = 4.41`, `V1 = 15.5`, `V2 = 29.3`. It is a plausible PK number that happens to share three digits with π. So the fix is not to edit four literals — it is one `#[allow(clippy::approx_constant)]` on that test module, with a comment saying why.

This is worth recording precisely because the lint is a false positive: whoever turns `--all-targets` on will meet four confident-looking errors about π in a pharmacokinetics file and may "fix" real parameter values to silence them.

## Why it matters beyond these four

`cargo clippy` with no further arguments is what a contributor runs locally, and `.github/PULL_REQUEST_TEMPLATE.md` asks them to tick:

> - [ ] `cargo clippy` clean

[...]

Issue #1022, opened 2026-08-21
[error_model] silently keeps only the first of several plain `DV ~ ...` lines


TeunP:
Found while fixing #1001. Same family — a single-endpoint `[error_model]` accepting
something it does not honour — but a distinct mechanism, so filed separately rather
than folded into that PR.

## What happens

A `[error_model]` block with more than one plain `DV ~ ...` line parses, keeps the
**first**, and silently discards the rest. No error, no warning.

```
[error_model]
  DV ~ proportional(S_SMALL)
  DV ~ additive(S_SMALL)
```

Measured on the #1001 branch binary against a 6-subject / 24-observation dataset,
`maxiter = 0` (so the OFV is a deterministic function of the model text):

| `[error_model]` | OFV |
|---|---|
| `DV ~ proportional(S_SMALL)` alone | −6.8397 |
| `DV ~ proportional(S_SMALL)` **then** `DV ~ additive(S_SMALL)` | −6.8397 |

Identical — the `additive` line contributed nothing and said nothing.

## Where

`parse_error_model`, `src/parser/model_parser.rs`:

```rust
match singles.into_iter().next() {
    Some((model, names, flags)) => {
        Ok((ParsedErrorModel::Single(model, names), flags, iiv_on_ruv))
    }
    None => Err("No error model found in [error_model] block".to_string()),
}
```

`singles` accumulates every plain `DV ~ ...` line in the block; `.next()` takes the
first and drops the tail.

## The bar is already set elsewhere in the same function

Two neighbouring cases in `parse_error_model` are explicit about exactly this:

- Duplicate per-CMT entries → `"[error_model] has more than one entry for CMT={}"`.
- Mixing a plain line with `CMT=N:` lines → `"[error_model] mixes a plain `DV ~ ...`
  line with `CMT=N:` per-compartment lines; use one style or the other"` (verified —
  this one is *not* silent).
- A second `iiv_on_ruv = ...` → `"[error_model] has more than one `iiv_on_ruv` entry"`.

So the block already rejects every other kind of duplicate. The plain-line case is
the one gap, and it is the one a user is most likely to hit by editing a model in
place — commenting out an error model by writing a new line above it, or pasting a
replacement without deleting the original.

## Suggested fix

Reject a second plain `DV ~ ...` line, matching the `iiv_on_ruv` wording:

```
[error_model] has more than one plain `DV ~ ...` line; a single-endpoint error
[...]

PR #1025, opened 2026-08-21
fix(parser): reject a single-endpoint [error_model] with mis-ordered sigmas [closes #1001]

### LINKED ISSUE #1001: [error_model] single-endpoint sigma names are validated then ignored; roles come from declaration order
## Summary

A single-endpoint `[error_model]` accepts a sigma name, checks that it exists,
and then ignores it. The sigmas are consumed **positionally from the declaration
order** in `[parameters]`, so `DV ~ proportional(S_SMALL)` silently uses
`S_BIG`'s variance whenever `S_BIG` is declared first.

The fit converges and returns numbers. There is no diagnostic at any level -- the
model validates, and `ferx_model_validate()` with data reports no finding.

The per-CMT and covariate-selected paths do the opposite and bind strictly by
name, so the same file means two different things depending on which
`[error_model]` form it uses.

Reproduced on `main` @ `8432ebe3` with `target/debug/ferx`, and on the released
`ferx-r` 0.3.0 (`7889719`).

## Reproducer

Four models, identical except for the `sigma` declarations and the name written
in `[error_model]`. `maxiter = 0` holds the thetas at their initials, so the OFV
is a deterministic function of the model text.

`a_big_first_names_small.ferx`:

```
[parameters]
  theta TVKE(0.1, 0.001, 1.0)
  theta TVV(10.0, 1.0, 100.0)

  omega ETA1 ~ 0.09

  sigma S_BIG ~ 2.0 (sd)
  sigma S_SMALL ~ 0.05 (sd)

[individual_parameters]
  KE = TVKE * exp(ETA1)
  V = TVV

[structural_model]
  ode(obs_cmt=CENT, states=[CENT])

[odes]
  d/dt(CENT) = -KE * CENT

[error_model]
  DV ~ proportional(S_SMALL)

[scaling]
  obs_scale = V

[fit_options]
  method = focei
  maxiter = 0
  covariance = false
```

`data.csv` -- 6 subjects, one 100 mg bolus, four observations each:

```
ID,TIME,AMT,EVID,MDV,CMT,DV
1,0,100,1,1,1,0
1,0.5,0,0,0,1,9.5
1,2,0,0,0,1,8.2
1,8,0,0,0,1,4.5
1,24,0,0,0,1,0.9
```
(repeat for `ID` 2..6; any dataset with a handful of observations shows it)

```bash
ferx a_big_first_names_small.ferx --data data.csv
```

### Results

| # | `sigma` declaration order | `[error_model]` | OFV |
|---|---|---|---|
| a | `S_BIG` then `S_SMALL` | `proportional(S_SMALL)` | **103.217050** |
| b | `S_BIG` then `S_SMALL` | `proportional(S_BIG)` | 103.217050 |
| c | `S_SMALL` then `S_BIG` | `proportional(S_SMALL)` | **-41.979370** |
| d | `S_SMALL` alone | `proportional(S_SMALL)` | -41.979370 |

- **a vs b**: changing the name in `[error_model]` changes nothing. The name is
  not read.
- **a vs c**: the same `[error_model]` line, differing only in the order of two
  `sigma` declarations, gives a completely different fit.
- **c vs d**: `-41.979` is what `proportional(S_SMALL)` actually means. Case (a)
  is off by that much while naming `S_SMALL`.

The same happens with `combined(A, B)`: transposing the two arguments changes
nothing, and transposing the two `sigma` declarations changes the fit
(103.209 -> 36.181 on a two-sigma variant of the model above).

## The two paths disagree

Same model, two sigmas 40x apart, each `[error_model]` form probed twice -- once
swapping the declaration order, once swapping the names in `[error_model]`:

| `[error_model]` form | swap declaration order | swap names in `[error_model]` |
|---|---|---|
| single-endpoint `proportional(...)` / `combined(...)` | **103.209 -> 36.181** | no change |
| per-CMT `CMT=n: DV ~ ...` | no change (409.550) | **409.550 -> 3508.421** |
| covariate-selected `if (COV ...) { ... }` | no change (409.549) | **409.549 -> 3508.395** |

Exactly complementary. The name-swap column is the control: without it, "the
declaration order does not matter" would be equally consistent with the names
also being ignored on those paths.

## Where

`build_error_spec`, `src/parser/model_parser.rs:11327`:

```rust
ParsedErrorModel::Single(model, args) => {
    // Single-endpoint sigmas are consumed positionally from the global
    // sigma vector. Each argument must scale exactly one declared sigma
    // (a bare name or a magnitude expression, #484); `arg_sigma_name`
    // both resolves that sigma and rejects an unknown/typo'd reference.
    for arg in &args {
        arg_sigma_name(arg, sigma_names)?;
    }
    Ok((model, ErrorSpec::Single(model)))
}
```

`args` is validated and then dropped -- `ErrorSpec::Single(model)` carries no
name. Twenty lines below, `ParsedErrorModel::PerCmt` does the opposite:

```rust
let idx = sigma_names.iter().position(|s| s == nm).ok_or_else(|| { ... })?;
```

So the behaviour is deliberate and documented in the comment; what seems
unintended is that the syntax invites a name and then silently disregards it.

## Why this is worth fixing rather than documenting

The failure is silent at every level a user can check. The model validates, the
fit converges, and the reported SEs and diagnostics are all internally
consistent -- there is nothing to notice. A model with an unused `$SIGMA` ahead
of the one it references is enough to trigger it; two epsilons are not required.

It also cannot be worked around by writing the names correctly, which is the
natural assumption. The only workaround is to order the `sigma` declarations to
match the roles, which is not stated anywhere and is not what the syntax
suggests.

## Suggested fixes, in order of preference

1. **Bind by name on the single-endpoint path**, as `PerCmt` already does.
   `ErrorSpec::Single` would need to carry the resolved sigma indices rather than
   just the `ErrorModel`. This makes the three forms agree and makes the written
   name mean what it says.

2. **Reject a mismatch.** If binding by name is a larger change than it looks,
   error at parse when the argument order does not match the declaration order --
   `[error_model] combined(S_SMALL, S_BIG) does not match the sigma declaration
   order (S_BIG, S_SMALL); single-endpoint sigmas are consumed positionally`.
   Loud and cheap, and it cannot silently mis-fit.

3. **At minimum, warn.** A `W_` diagnostic when the referenced names are not the
   first `n_sigma()` declared, in declaration order. Weakest option -- it leaves
   a correct-looking file that means something else -- but strictly better than
   today.

Option 2 would be a breaking change for any existing model that relies on the
current behaviour while naming the sigmas in a different order; those models are
currently getting a different fit from the one they appear to describe, so
breaking them is arguably the point.

## Context

Found while fixing FeRx-NLME/translator#6. The translator now orders its emitted
`sigma` declarations to match the error roles
(FeRx-NLME/translator#22) so its own output is correct either way, but that
protects only files this package generates -- a hand-written or hand-edited model
has no such protection.

TeunP:
## Why

A single-endpoint `[error_model]` consumes its sigmas **positionally** from the
`[parameters]` declaration order. The names written in the arguments were checked for
existence and then discarded, so `DV ~ proportional(S_SMALL)` bound `S_BIG` whenever
`S_BIG` happened to be declared first — and `combined(A, B)` ignored a transposition of
its two arguments entirely.

Nothing surfaced it. The model validated, the fit converged, and every reported SE and
diagnostic was internally consistent, so there was nothing for a user to notice. The
only workaround was to order the `sigma` declarations to match the roles, which is not
stated anywhere and is not what the syntax suggests. The per-CMT and covariate-selected
forms bind strictly **by name**, so the same file meant two different things depending
on which `[error_model]` form it used.

The unused-parameter warning made it worse rather than better: it keys off the names in
`[error_model]`, so in the reproducer it reported `S_BIG` — the sigma actually driving
the fit — as "declared but not referenced", and said nothing about `S_SMALL`, the one
being ignored.

Reported in #1001, found while fixing FeRx-NLME/translator#6.

## What changed

`build_error_spec`'s `Single` arm now requires each argument to name the sigma that
actually occupies its slot, rejecting a mismatch with `E_SIGMA_ORDER_MISMATCH`.

This **generalises a check that already existed**: `validate_block_sigma_single_error_order`
enforced exactly this rule, but only when the model also declared a `block_sigma`. That
function is deleted and its call site removed — keeping it would have left two authorities
for the same rule, free to drift, and its loop was already redundant with the
`arg_sigma_name` resolution in `build_error_spec`. One resolution rule now covers the whole
positional path, `block_sigma` included.

Two distinct failures, deliberately not merged into one message:

- **Order mismatch** — argument *i* resolves to a declared sigma that is not
  `sigma_names[i]`. `E_SIGMA_ORDER_MISMATCH`; remedy is to reorder either list.
- **Too few distinct sigmas** — reachable only by repeating a name (`combined(S_A, S_A)`
  with one sigma declared), since the argument count is already pinned to the model's
  `n_sigma()` upstream and every argument must resolve to a declared sigma. That is a
  count problem, not an ordering one, so it reports the count and does **not** suggest a
  reordering that cannot help. Plain `E_PARSE`.

### Why reject rather than bind by name

Binding by name was the issue's first preference and I evaluated it. `ErrorSpec::Single`
carries no indices — every consumer (`sigma_loadings`, `sigma_loading_slopes`,
`residual_variance`, `sigma_types`, `dvar_dlogsigma`, `combined_additive_sigma_indices`)
reads the *leading* slots, and the #484 residual-magnitude multipliers are stored per
argument slot and applied by flat sigma index, an alignment that holds today only because
`Single` is positional. Threading indices through all of that to support an ordering the
positional path cannot otherwise express is a large change for no expressive gain: the
same model can be written correctly by moving one line. Rejecting also keeps `Single`
cheap and keeps one rule for the whole positional path.

A useful side effect: the #484 alignment above stops being a coincidence and becomes an
invariant the parser enforces.

## Alternatives considered

Warning rather than error (the issue's option 3) — rejected. A warning leaves a
correct-looking file that means something else, which is the failure mode being fixed. The
blast-radius sweep (below) found zero affected models, so the cost of erroring is nil and
the cost of warning is a permanently ambiguous DSL.

## Cross-repo dependency

No `pub` API change. All 69 `.ferx` models on ferx-r `origin/main` were swept and none is
affected; the change reaches ferx-r at its next routine lock bump.

## Breaking changes

- [ ] Public Rust API
- [ ] None

<details>

A model whose single-endpoint `[error_model]` names its sigmas in an order other than the
`[parameters]` declaration order now fails to parse. Such a model was already getting a
different fit from the one it appeared to describe. Fix by reordering either list so they
agree:

```
[parameters]
  sigma PROP_ERR ~ 0.04 (sd)     # slot 1 -> proportional
  sigma ADD_ERR  ~ 0.1  (sd)     # slot 2 -> additive

[error_model]
```

**Blast radius, measured — zero affected models.** Swept by running `ferx check` over every
model and grepping for the new code, not by inspection:

| Corpus | Models | Newly rejected |
|---|---:|---:|
| `examples/*.ferx` | 101 | 0 |
| ferx-r `origin/main` `**/*.ferx` | 69 | 0 |
| inline models in `src/` + `tests/` | — | 0 (full suite green) |

The one shape that looked at risk is `examples/warfarin_frem.ferx`, which declares
`EPSCOV` after `PROP_ERR` and consumes it via `frem_sigma` rather than `[error_model]`.
That is in-order and unaffected; it has a regression test.

</details>

## Numerical validation

- [x] Reference values:

Two control streams differing in exactly one line
(`nonmem_anchor/sigma_order_{small,big}.ctl`, `ADVAN1 TRANS1`, `$THETA` all `FIX`,
`$OMEGA 0.09 FIX`, `MAXEVAL=0 METHOD=1 INTER`), against ferx at `inner_tol = 1e-10`,
`ode_reltol = 1e-10`, `ode_abstol = 1e-12`:

```
  sigma (SD)   NONMEM 7.6.0     ferx           |diff|
  0.05        -6.83893110      -6.838931      < 1e-6
  2.00       103.82896422     103.828964      < 1e-6
  --------------------------------------------------
  delta       110.66789532     110.667895
```

So the silent misbind was worth **110.67 OFV units** on 24 observations, and both
endpoints of that interval are independently confirmed. At the default `inner_tol = 1e-5`
the gap to NONMEM is ~8e-4 on *both* rows and closes together when the tolerance is
tightened — that is EBE noise, not disagreement.

Nothing previously pinned `Single`-path sigma consumption against NONMEM at all, so these
anchors are new coverage regardless of this fix.

NONMEM cannot express the bug: `$ERROR` indexes `EPS(1)`/`EPS(2)` positionally, with no
name to disagree with. That is *why* the fix is a rejection — there is no reference
behaviour to re-bind to, only a spelling that promises something the engine does not do.

Reproducer, before → after (6 subjects, 24 obs, `maxiter = 0`):

```
                                             before        after
a  decl [S_BIG, S_SMALL], proportional(S_SMALL)   103.8281   E_SIGMA_ORDER_MISMATCH
b  decl [S_BIG, S_SMALL], proportional(S_BIG)     103.8281   103.8281   (unchanged)
c  decl [S_SMALL, S_BIG], proportional(S_SMALL)    -6.8397    -6.8397   (unchanged)
d  decl [S_SMALL],        proportional(S_SMALL)    -6.8397    -6.8397   (unchanged)
```

`a` and `b` were the same fit before this PR — that is the bug. `c == d` exactly, which is
the trailing-unused-sigma invariant, now pinned against NONMEM as well.

## Performance

- [x] No significant impact expected — one extra string compare per `[error_model]`
  argument (≤2), at parse time.

## Tests

- [x] Unit test(s) added (`cargo test --lib` passes — 3131 passed, 0 failed as of `a1bff88d`)

Every new test was run against the unfixed tree first and confirmed red (6 new + 1 updated
red, 2 controls green), then green after the fix. The two sentinel-substring couplings were
additionally mutation-verified: breaking the `parse_error_to_diagnostic` sentinel drops the
code to `E_PARSE` and fails the `check_command` test; perturbing a NONMEM constant by 1e-4
fails all three anchor comparisons.

Tier 1 — `src/parser/model_parser_tests.rs` (9 new, 1 updated):
wrong name; correct order parses (control); transposed `combined`; repeated arg with two
sigmas declared and with one; trailing unused sigma parses and keeps its existing warning;
the FREM spelling (`frem_sigma`, a separate test); #484 magnitude expression; LTBS spelling;
and the pre-existing `block_sigma` order test re-pointed at the generalised message.

Tier 1 — `src/api/tests/block_name_diagnostic_tests.rs` (2 new, added in review): the
`E_SIGMA_ORDER_MISMATCH` mapping and the deliberate `E_PARSE` split for a repeated argument,
both driven by the parser's own error string rather than a pasted literal.

Tier 2 — `tests/check_command.rs` (1 new): `ferx check` reports `E_SIGMA_ORDER_MISMATCH`
with no doubled block prefix, mirroring the #993 sentinel pin.

Tier 2 — `tests/sigma_order_nonmem_anchor.rs` (5 new): the three NONMEM comparisons above,
the rejection, and `default_tolerances_stay_close_to_nonmem`. `maxiter = 0`, no convergence
loop, so no `slow-tests` gate.

Tier 2 — `tests/per_route_lag.rs` (1 new, added in review): the shipped example binds its
proportional and additive sigmas by **role**, not merely in a parseable order.

> **Note on tiering.** An earlier draft of this section assumed Tier-2 files are only
> compile-checked per PR. They are not: `.github/workflows/ci.yml:259` runs
> `cargo llvm-cov --tests …`, so `tests/*.rs` both run and count toward the patch gate on
> every PR. Only `slow-tests`-gated (Tier-3) tests are deferred to nightly. `CLAUDE.md`'s
> Tests section is stale on this point.

## Example

- [x] Not applicable — this removes an accepted spelling rather than adding one. The
  accepted form is unchanged and documented in `docs/model-file/error-model.qmd`.

## Docs

- [x] `docs/` updated — new **Sigma order** section in `docs/model-file/error-model.qmd`
  (rule, rejected example, why it is an error not a re-binding, the trailing-sigma
  carve-out); cross-reference from `docs/model-file/parameters.qmd`'s sigma-role table,
  which previously said "first sigma / second sigma" without saying *which* first;
  `E_SIGMA_ORDER_MISMATCH` row in `docs/file-formats/check-report.qmd`.
- [x] No new pages, so no `docs/_quarto.yml` change.
- [x] `CHANGELOG.md` `[Unreleased]` entry added under `Changed`.

## Checklist

- [x] `cargo clippy --no-default-features --features ci,markov` (the CI invocation) clean — warnings
      only, and none in the changed regions of `model_parser.rs` / `validation.rs`
- [x] `cargo check --tests` clean on `ci`, `ci,survival,slow-tests`, `ci,markov`
- [ ] Not on the `Dual2` sensitivity path — parse-time only
- [ ] No version bump (pre-1.0 breaking parse change, batched with the other `[Unreleased]`
      breaking entries)

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → `ferx-site/model-dsl/` page updated or PR opened
      — no new syntax; this **removes** an accepted spelling. The rule is documented in
      `docs/model-file/error-model.qmd#sigma-order` and cross-referenced from `parameters.qmd`.
- [x] New example `.ferx` file added to `examples/` → matching entry in ferx-site examples and ferx-book
      — no new example. `examples/per_route_lag_absorption.ferx` is **modified** (see below); it is
      already listed.
- [x] New data file added to `data/` → mirrored to `ferx-r/inst/examples/data/` and referenced in site/book
      — none added. `nonmem_anchor/sigma_order.csv` is anchor input, not a user-facing dataset.

### ferx-book
- [x] New estimator, DSL feature, or fit option → relevant book chapter updated or PR opened
      — no new feature. The breaking-change note is in `CHANGELOG.md`; the rule and both remedies
      are in `error-model.qmd`.

- [x] All affected ferx-site example `.qmd` pages render cleanly — none affected
- [x] All affected ferx-book chapters render cleanly — none affected

> **One example is affected and its `--simulate` run has *not* been re-executed.**
> `examples/per_route_lag_absorption.ferx` declared `sigma ADD_ERR` before `sigma PROP_ERR` while
> writing `combined(ADD_ERR, PROP_ERR)`. Those are consumed positionally, so 0.02 was the
> *proportional* coefficient and 0.10 the *additive* SD — the inverse of the file's own header, both
> inline comments, and the `--simulate` parameters it advertises. Both lists were transposed
> together, so the new order check cannot see it: it compares names to slots, never names to roles.
> The declarations and arguments are now in role order, pinned by
> `shipped_per_route_lag_example_binds_its_residual_components_by_role`.
>
> This **changes the numerical meaning of a shipped example**. The binding is verified by test, but
> `cargo run --release -- examples/per_route_lag_absorption.ferx --simulate` (a 500-iteration fit)
> has not been run against the corrected file. Worth doing before merge.

## Reviewer hints

Focus on `build_error_spec`'s `Single` arm. Since review it has two guards, in this order:
a **repeat** check against the sigmas earlier arguments already claimed, then the slot
comparison. The order is load-bearing — without it `combined(S_A, S_A)` with two sigmas
declared lands in the mismatch arm and is told to reorder two lists that no permutation can
bring into agreement, *and* gets the reorder-shaped `E_SIGMA_ORDER_MISMATCH` that a consumer
would act on. Both spellings of the repeat now report the same defect with the same code.

The `None` arm the first draft asked reviewers to scrutinise is gone (`a1bff88d`): once the
repeat is caught first, every resolved name is distinct and declared, so by argument `idx` at
least `idx + 1` sigmas exist and `sigma_names.get(idx)` is always `Some`. An arm no input can
reach is one no test can cover — it was 8 permanently-missed lines and the entire
`codecov/patch` shortfall. It is folded into the mismatch arm, which still degrades to a
message rather than an `unreachable!()` panic.

Both remedies in the message are deliberate and both matter: reordering the *arguments* of a
`combined` model swaps which sigma is proportional (a different model, ~32 OFV units on the
repo's own data), and reordering `block_sigma` *names* without permuting its lower triangle
swaps the two variances and parses clean (~28 OFV units against the committed anchor).

Second: confirm deleting `validate_block_sigma_single_error_order` lost nothing. Its logic
was a strict subset (same loop, gated on `block_sigmas` non-empty), and no error can fire
between its old call site and `build_error_spec` — the intervening lines are pure captures
— so error ordering is unchanged. The updated `block_sigma` test pins that the case is
still rejected.

Skimmable: the docs and CHANGELOG prose, and the anchor `.ctl`/`.lst` files.

## Open questions

None blocking. Two things found in passing and filed separately rather than folded in:

- **#1022** — `[error_model]` silently keeps only the **first** of several plain `DV ~ ...`
  lines, while every other duplicate in the same block (per-CMT entries, `iiv_on_ruv`, mixing
  plain with `CMT=`) is rejected explicitly. Verified on this branch: a second
  `DV ~ additive(...)` line changes nothing and warns nothing. Same "accepted but not
  honoured" family as #1001, different mechanism.
- **#1023** — CI's clippy job lints the library target only, so `#[cfg(test)]` code is never
  linted; `cargo clippy --all-targets` consequently fails on `main` today with four
  false-positive `approx_constant` denials in `src/sens/two_cpt_explicit.rs`'s test module
  (`3.14` there is an inter-compartmental clearance `Q`, not π). Pre-existing, unrelated to
  this branch, and left alone here.


TeunP:
## Review — `xhigh` (10 finder angles + adversarial verify)

**The rule is right, the anchor is solid, and the blast-radius claim holds — I re-measured all
three rather than taking them on trust. What is wrong is the *remediation*: two of the three fixes
this PR tells a user to apply silently produce a different model, and one class of rejection is
misdiagnosed and handed the wrong structured code. All fixed on the branch, together with two
adjacent bugs the review turned up.**

### Re-measured, not taken on trust

| Claim | Result |
|---|---|
| the two NONMEM constants | `-6.8389311000825916` / `103.82896421994307` match `nonmem_anchor/results/sigma_order_{small,big}.ext` **exactly** |
| "zero affected models" | 119 in-repo `.ferx` + 69 on `ferx-r@origin/main` swept by resolving each argument against the declared order → **0** newly rejected |
| "no error can fire between the deleted call site and `build_error_spec`" | true — `model_parser.rs:2064–2081` is three infallible captures, no `?`. Error ordering is unchanged and the deleted guard is a strict subset |
| `block_sigma` entries land in true `[parameters]` order | true (`model_parser.rs:11304`, same single pass as plain `sigma` lines) — so "the `[parameters]` declaration order" in the message is accurate for them |
| the `None` arm's reachability argument | true as written: `args.len() == n_sigma()` is pinned at `:12218` and duplicate declarations are rejected at `:1367` |
| `ErrorSpec::Single` reads the leading slots | true — `sigma_loadings` yields `(0, f)` / `(1, 1.0)`, so the enforced alignment is the right one |
| the 15 new tests | all pass |
| the `E_SIGMA_ORDER_MISMATCH` sentinel | intact across its `\`-continuation in the source, and unreachable by any earlier arm in `parse_error_to_diagnostic` |

One correction to a premise in the PR description: **Tier-2 tests do run on PRs.** `ci.yml:259` is
`cargo llvm-cov --tests --no-default-features --features ci`, and the step's own comment says
"`--tests` so the integration tests both run and count". `CLAUDE.md`'s Tests section ("compile-checked
on every PR … run nightly") is stale, not this PR. So `tests/sigma_order_nonmem_anchor.rs` and the
`check_command.rs` sentinel pin both register per-PR coverage.

---

### Blocking — the remediation is wrong in three ways

**1. `(or reorder the arguments)` re-creates #1001.** `model_parser.rs:12355`

For `combined`, transposing the arguments swaps which sigma is the proportional coefficient. It is
not a different spelling of the same model; it is a different model. Measured on
`data/correlated_residual_combined.csv`, starting from a rejected file:

| remedy | result |
|---|---|
| reorder the **declarations** | OFV 14.6210, `PROP_ERR = 0.20` proportional — the model the user wrote |
| reorder the **arguments** | OFV 46.4040, `ADD_ERR = 1.00` proportional — **31.8 OFV units away**, and exactly the pre-#1001 binding |

The parenthetical is also meaningless for `proportional`/`additive` (one argument), so `combined`
is the only case where it applies and the only case where it is dangerous. `E_SIGMA_ORDER_MISMATCH`
exists precisely so `ferxtranslate`/ferx-r can apply the fix mechanically — an auto-fixer that picks
the shorter rewrite silently changes the model. Same wording in `error-model.qmd:106`,
`check-report.qmd:73` and `CHANGELOG.md:36`.

**2. "reorder the sigma declarations" silently corrupts a `block_sigma` model.** `model_parser.rs:12355` × `:11304`

The lower triangle is positional too (`sigmas.push(SigmaSpec { name: names[row], value: values[val_idx].sqrt() })`),
so reordering the names without permuting it swaps the two variances and parses clean. Measured on
this repo's own NONMEM-anchored `examples/correlated_residual_combined.ferx` (OFV 18.7274): transpose
it the way a NONMEM port would, then apply the message's advice literally →

```
block_sigma (PROP_ERR, ADD_ERR) = [1.00, 0.10, 0.04] FIX
```

→ `ferx check` reports `ok: no errors (0 warning(s))`, and the fit gives `PROP_ERR = 1.000000
(CV% 100.0)`, **OFV 46.7557 vs the anchored 18.7274** — a 28-unit silent mis-specification produced
by following the fix instructions. The matching rewrite is `[0.04, 0.10, 1.00]`, which nothing says.
`block_sigma` is exactly the case the deleted `validate_block_sigma_single_error_order` used to own.

**3. A repeated argument is misdiagnosed, and the two spellings of the same defect get different
codes.** `model_parser.rs:12346`

The `None` arm reasons, correctly, *"That is a count problem, not an ordering one — say so, and do
not suggest a reordering that cannot help."* But it only fires when `sigma_names` runs out. With two
sigmas declared, the repeat lands in the `Some(slot)` arm:

```
sigma S_A ~ 0.05 (sd)
sigma S_B ~ 1.0  (sd)
DV ~ combined(S_A, S_A)
```
→ *"argument 2 names sigma 'S_A' … supplies 'S_B' … Reorder the sigma declarations … (or reorder the
arguments)."* No declaration order puts `S_A` in both slots (duplicates are rejected at `:1371`) and
transposing identical arguments is a no-op — **both** prescribed fixes are impossible. And the coding
is inverted: this unfollowable message gets the machine-actionable `E_SIGMA_ORDER_MISMATCH`, while
the *clear* count message deliberately omits the sentinel and degrades to `E_PARSE`. The PR's own
`single_error_model_repeated_sigma_arg_errs` pins the wrong behaviour, asserting on `S_B` — a sigma
the user never wrote.

---

### Two adjacent bugs this review turned up (both fixed)

**4. A trailing sigma is not inert on the FOCE score.** `gauss_newton.rs:688`

`dr_diag_d_log_sigma`'s `Combined` arm answers every `sigma_k >= 1` with the *additive* sigma's
derivative, where its `Additive` and `Proportional` siblings correctly return `0.0`. The caller loops
`for ks in 0..n_sigma` over the **whole flat sigma vector**, so a third, unreferenced sigma — legal,
and documented by this PR as "a later slot that the error model never loads" — picks up
`∂R/∂log σ = 2·σ_add² ≠ 0` for a parameter that never enters `R`. That is a spurious row and column in
`subject_nll_pop_grad`, which is also the score cross-product behind `covariance = s` / `rsr`
(`covariance.rs:468`) and `trust_region.rs:112`: **wrong sandwich standard errors for every parameter,
silently**, on `method = foce` / `gn` / `gn_hybrid`. Pre-existing — but this PR is what makes the shape
documented, tested and supported.

**5. `examples/per_route_lag_absorption.ferx` has its two residual components the wrong way round.**

```
sigma ADD_ERR  ~ 0.02 (sd)       # additive residual …
sigma PROP_ERR ~ 0.10 (sd)       # proportional residual (~10% CV)
DV ~ combined(ADD_ERR, PROP_ERR)
```

Slot 0 is the proportional coefficient, so this file simulates and fits with a **2% proportional term
and a 0.10 mg/L additive floor** — the inverse of its own header (`sigma combined (0.02 additive +
0.10 proportional)`), both inline comments, and the `--simulate` parameters it advertises. It is
self-consistent, so nothing fails; `tests/per_route_lag.rs:637` only asserts it parses, and
`absorption.qmd:931` points readers at it as the `--simulate` demo.

This is the limit of the new rule, and worth stating in the PR: **it compares names to slots, never
names to roles**, so a file with *both* lists transposed is invisible to it — and to the sweep. That
is exactly what a user does when they write the declarations in the order the arguments read. A
role-oriented sweep over all 188 models finds this one file and no other.

---

### Smaller (all fixed)

| # | Where | What |
|---|---|---|
| 6 | `model_parser_tests.rs:5392` | the retargeted `block_sigma` assertion (`"consumed positionally" && "PROP_ERR"`) passes for **both** the argument-1 and argument-2 messages, so an off-by-one in the new `enumerate` loop still passes |
| 7 | `error-model.qmd:118` | the docs bless the "declared but not referenced … will not affect predictions" warning for the FREM shape — but `likelihood.rs:567` reads `sigma_values[covariate_sigma_index]`, so for `frem_sigma` that clause is false |
| 8 | `diagnostics.rs:36` | module doc says *"Add new codes here and document them in `check-report.qmd`"*; only the `.qmd` was updated |
| 9 | `model_parser.rs:14031` | the hoist left a byte-identical `Regex::new(r"[A-Za-z_]\w*")` in `used_sigma_names`, on the same parse path 16 lines earlier and scanning a strictly wider set — ~1.07 ms release / ~5.25 ms debug per parse, more than the two new statics save |
| 10 | `sigma_order_nonmem_anchor.rs:181` | delta asserted at `1e-5` while composing two measurements each toleranced at `1e-5` → worst case `2e-5`; flaky by construction |
| 11 | `sigma_order_nonmem_anchor.rs:222` | `default_tolerances_…` re-inlines the whole model to vary three `[fit_options]` lines — the drift the file's own `DATA` comment argues against, and its `1e-2` band is wide enough to hide it |
| 12 | `sigma_order_nonmem_anchor.rs:174` | second `fit()` duplicates `single_sigma_matches_nonmem`'s call verbatim (model literal included) |
| 13 | `model_parser_tests.rs:5430` | comment quotes `103.217 / −41.979` — issue #1001's dataset, contradicting both the PR body (`103.8281 / −6.8397`) and the anchor, on a *third* model that never fits data. Also `:140` numbers slots 0-based where the messages and docs use 1-based |
| 14 | `model_parser_tests.rs:5521` | `single_error_model_frem_shape_parses` hand-writes the model, so it does **not** pin `generate_frem_model`'s append-last ordering as its comment claims. The real pin is `test_generate_frem_model_preserves_scaling_block`, which parses generated text — the ordering *is* covered, just not by the test that says so |
| 15 | `error-model.qmd:120` | adding a `CMT=1:` prefix (or an `if/else` selector, which is not ODE-gated) turns the rejection into a clean by-name binding. `frem_sigma`, `[mixture] sigma(k)`, block_sigma off-diagonals and per-CMT/`Selected` all bind by name — four of five mechanisms — so the escape hatch deserves naming in the docs |

---

### Not findings — checked and clear

`arg_sigma_name` rejects multi-sigma expressions, so no "first match" hole; the `#484` per-slot
multipliers really are argument-indexed and the new check is what makes that alignment an invariant;
`sigma_names` carries no duplicates and is never reordered after parse; `PerCmt`/`Selected` bind by
name as claimed; the hoisted regexes are byte-identical and `^…$` anchors per-haystack; `temp_model`
is process+counter unique; `nonmem_anchor/results/*.{lst,ext}` is the documented committed
convention; the `E_SIGMA_ORDER_MISMATCH` sentinel cannot be swallowed by an earlier arm in
`parse_error_to_diagnostic`.

On **reject vs. bind-by-name**: I checked the cost estimate rather than accepting it.
`ErrorSpec::Single(ErrorModel)` has 78 non-test references and only 2 production construction sites,
and `residual_error.rs` duplicates the `PerCmt` body ten times for `Single` — so giving `Single` an
`EndpointError` is genuinely smaller than the PR argues (net-negative production lines) and this repo
has done that shape of compiler-driven migration before (#912/#919). **I still agree with rejecting
here**: NONMEM has no name to disagree with, so there is no reference behaviour to re-bind to. But
the honest framing is "this is a scope decision", not "threading indices is too large" — worth a
follow-up issue rather than a paragraph in the PR body.


---

### Fixed on the branch — `97fb662a`

```
 CHANGELOG.md                                 | 21 +-
 docs/file-formats/check-report.qmd           |  2 +-
 docs/model-file/error-model.qmd              | 31 +-
 examples/per_route_lag_absorption.ferx       |  8 +-
 src/api/tests/block_name_diagnostic_tests.rs | 74 +++-
 src/diagnostics.rs                           |  1 +
 src/estimation/gauss_newton.rs               | 54 ++-
 src/frem/mod.rs                              |  9 +-
 src/parser/model_parser.rs                   | 91 ++++--
 src/parser/model_parser_tests.rs             | 66 ++-
 tests/per_route_lag.rs                       | 38 ++
 tests/sigma_order_nonmem_anchor.rs           | 92 ++--
 12 files changed, 389 insertions(+), 98 deletions(-)
```

- **The order message** now names the declaration-side rewrite as *the* fix, says to permute a
  `block_sigma`'s lower triangle with it, and states what the argument-side rewrite costs. Same in
  `error-model.qmd` (with a `callout-warning` and a worked `block_sigma` example),
  `check-report.qmd` and the CHANGELOG.
- **The repeat** is caught by a `claimed` set *before* the slot comparison, so both spellings report
  the same defect with the same code and neither prescribes a reordering. The `None` arm is now
  unreachable — kept as an error rather than `unreachable!()`, so a future relaxation degrades to a
  message instead of a panic in the parser.
- **`dr_diag_d_log_sigma`** gets an `else if sigma_k == 1 { … } else { 0.0 }`, matching its two
  siblings, plus `combined_ignores_sigmas_past_its_own_slots` pinning all three arms together.
- **The example** is in role order, with `shipped_per_route_lag_example_binds_its_residual_components_by_role`
  pinning the roles rather than the spelling (parsing alone cannot catch a double transposition).
- **`parse_error_to_diagnostic`** gets Tier-1 tests in its existing sibling module, driven by the
  parser's own error string rather than a pasted literal — the sentinel is split across a `\`
  continuation at the emitting site, so a copy could agree with itself while disagreeing with what
  is emitted.
- Plus: the `block_sigma` assertion now pins the argument index; the anchor's delta band `1e-5 → 2e-5`;
  `default_tolerances_…` built through `model_with_options` instead of re-inlining the model;
  `ofv()` carries the sibling anchors' `is_finite` guard; the FREM comment no longer claims a pin it
  does not make (and `frem/mod.rs` documents the EPSCOV-last constraint where it is *created*);
  stale OFVs and the 0-based "slot" comment corrected; `E_SIGMA_ORDER_MISMATCH` added to
  `src/diagnostics.rs`'s registry.

**Verification.** `cargo test --lib --features ci` → **3131 passed, 0 failed**; anchor 5/5;
`check_command` 21/21; `per_route_lag` 2/2. `cargo fmt --check` clean, `cargo check --tests` clean on
`ci,survival,slow-tests`. The one remaining lib warning (`src/api/run.rs:867`, `unused_mut`) is
pre-existing on this branch and untouched.

Every new and changed assertion was **mutation-verified** — four independent mutations, one build,
disjoint attribution:

| mutation | went red |
|---|---|
| `else if sigma_k == 1` → bare `else` | `combined_ignores_sigmas_past_its_own_slots` |
| example roles transposed back | `shipped_per_route_lag_example_binds_its_residual_components_by_role` |
| "lower triangle" clause dropped | `test_parse_full_model_block_sigma_single_endpoint_order_mismatch_errs` |
| repeat guard defeated | `single_error_model_repeated_sigma_arg_errs`, `…_without_a_second_sigma_errs`, `repeated_sigma_argument_is_not_coded_as_an_order_mismatch` |

Nine controls stayed green under all four, including `shipped_per_route_lag_example_parses` — which
is the point of the new role test: the parse test cannot see a double transposition.

### Left for you

- **The PR description is missing the template's `## Docs & examples` section**
  (`### ferx-site` / `### ferx-book` / `### Example execution`) — the body goes `## Checklist` →
  `## Reviewer hints`. `CLAUDE.md` says *"always read `.github/PULL_REQUEST_TEMPLATE.md` and fill
  every section"*, and it is the section that matters most here: this is a breaking DSL change, and
  it now also touches a shipped `examples/*.ferx`. I did not edit your description.
- **Test counts in `## Tests`**: 9 new `#[test]` in `model_parser_tests.rs` (not 8 — the enumeration
  merges `trailing_unused_sigma_parses` and `frem_shape_parses` into one bullet), and 5 in
  `sigma_order_nonmem_anchor.rs` (not 4 — `default_tolerances_stay_close_to_nonmem` is uncounted).
- **`src/diagnostics.rs`'s registry has drifted badly** — ~65 emitted codes are missing from it
  (`E_DOSE_ATTR_DOUBLE_USE` from #993, the whole `E_ABSORPTION_*` and `E_BLOCK_SIGMA_*` families, …).
  I added only this PR's row; the table needs generating or re-labelling as partial. Worth its own
  issue.
- **`ferxtranslate`** generates `.ferx` from `$SIGMA` + `$ERROR`, both positional with no name to
  disagree. A translator that emits `[parameters]` in `$SIGMA` order and `combined(<prop eps>, <add eps>)`
  in role order will emit newly-rejected models whenever the two disagree. No evidence of it in this
  repo's corpus, but it is the one consumer that builds model text mechanically from a positional
  source — and it reaches this change transitively (ferx-r `Cargo.lock` → its own pin), with no knob
  on their side.

Issue #814, opened 2026-07-12
ODE-twin reroute doesn't inherit call-time ODE tolerances or the inner ODE convergence-stall (surfaced by #719)


TeunP:
## Summary

The absorption **ODE twin** (`absorption_ode_equivalent`, the `transit()`/`igd()` forcing model that `CompiledModel::effective_for` swaps in for TV-covariate / `TIME` / — as of #719 — IOV subjects) is a **separate `CompiledModel` reconstructed from parse-time source text**. Two consequences of that separateness were surfaced by the #719 code review (PR #813); both are **pre-existing** (they already affected the TV-covariate / flip-flop reroute) but are now hit far more often because every IOV subject on a closed-form transit/IG model routes to the twin.

## 1. Call-time ODE tolerances don't reach the twin

`absorption_ode_equivalent_source` re-emits only the model-file `[fit_options]` block. A caller that sets `ode_reltol` / `ode_abstol` / `ode_max_steps` at **call time** (via `FitOptions`, e.g. from ferx-r `ferx_fit(settings=list(ode_reltol=1e-10))`) has that override merged into the runtime `FitOptions` and applied to the *primary* model — but the primary closed-form model is analytic, so `sync_ode_solver_opts` is a no-op on it, and nothing pushes the call-time setting onto the twin's `OdeSpec`. On the first `effective_for`, the twin is parsed from source that (for a typical transit author) omits ODE tolerances entirely, so **every rerouted subject integrates at the parse-default tolerance (~1e-6), silently ignoring the requested accuracy** for the whole fit/predict.

- Impact: a user asking for tight ODE accuracy on a transit/IG + IOV (or TV-cov) fit silently doesn't get it. (The committed #719 anchors avoid this only because their model DSL carries `ode_reltol=1e-9` explicitly in `[fit_options]`.)
- Fix direction: after building the twin (`get_or_build`), push the live `FitOptions` ODE settings onto its `OdeSpec` (a `sync_ode_solver_opts` on the effective model), or carry the runtime ODE options into the twin at reroute time.

## 2. The inner EBE loop's ODE convergence-stall reads the un-rerouted model

[...]

Issue #735, opened 2026-07-07
Flip-flop transit with lagtime / f / user-[odes] has no ODE twin (still returns 0)


TeunP:
### Context

PR #733 auto-routes a flip-flop transit closed form (`ke ≥ KTR`, or `α ≥ KTR` for 2-cpt) to its ODE `transit()` twin, so those models predict and fit correctly instead of returning a zero profile.

The reroute needs a twin, and the desugar that builds it (`transit_ode_equivalent_source`, `src/parser/model_parser.rs:6234`) **declines** when the closed-form transit model also carries a `lagtime=` / `f=` mapping or a user `[odes]` block (it validates only `cl,v,n,mtt` / `cl,v1,q,v2,n,mtt`). So a flip-flop `pk one_cpt_transit(cl,v,n,mtt, lagtime=LAG)` (etc.) has **no twin to route to** and keeps returning `0`.

### Current behaviour (safe, not silent)

Those cases keep the closed form's zero and emit the **actionable** `W_TRANSIT_FLIP_FLOP` warning ("rewrite it as an explicit ODE transit model"). So they fail loud, not silently — covered by `transit_flip_flop_without_twin_warns_actionable` in `tests/transit_analytic_equivalence.rs`.

### Enhancement

Extend `transit_ode_equivalent_source` to also build a twin for `lagtime=` / `f=` transit models (apply the lag / bioavailability in the twin's dosing / `[odes]`, as the ODE `transit()` path already supports both), so those flip-flop models auto-route too instead of requiring a manual ODE rewrite. User `[odes]` blocks stay out of scope (no unique desugar).

_Follow-up to #733._

PR #1027, opened 2026-08-21
fix(parser): build the absorption ODE twin at parse time and decline on failure [closes #1008]

### LINKED ISSUE #1008: Absorption ODE twin: get_or_build's .expect() turns any twin parse failure into a mid-fit panic
Follow-up from the review of #1003 (which closes #993). Not a live bug today — both known instances are fixed in that PR — but the mechanism that produced them is still armed.

## The trap

`AbsorptionOdeEquivalent::get_or_build` (`src/types.rs`) turns any parse failure of the reconstructed twin source into a panic:

```rust
let mut built = crate::parser::model_parser::parse_model_string(&self.source)
    .expect("internal: absorption ODE equivalent failed to build");
```

The comment justifies it: "The source is reconstructed from already-validated model blocks, so parsing it is infallible in practice; a failure is an internal reconstruction bug." That reasoning holds only while the twin's parse can never disagree with the primary's. It can, for a structural reason:

**the primary is an analytical model and the twin is an ODE model.** Any parse check that is ODE-scoped therefore runs on the twin and not on the primary. A model the primary accepts can produce a twin the parser rejects, and the rejection surfaces as a panic — mid-fit, on the plain `predict`/`fit` path, since `effective_for` builds the twin lazily for TV-covariate / `TIME` / IOV subjects.

## Evidence: #1003 armed it twice in one commit

Both reproduced, both now fixed in that PR, both found only because the reviewer went looking:

1. **`[adaptive_dosing]` re-emitted into the twin.** `absorption_ode_equivalent_source` declines when `[odes]`, `[scaling]` or `[initial_conditions]` are present — those are exactly the blocks whose ODE-only checks could bite — but `[adaptive_dosing]` was not on that list and was re-emitted verbatim. A plain transit model with `observe = central / (V * F)` parsed, then panicked on twin build. Fixed by adding `adaptive_dosing` to the skip list.

2. **A dose-attribute parameter reused as a disposition role.** `pk one_cpt_transit(cl=CL, v=F, n=N, mtt=MTT, f=F)` passes the #735 shadow guard (`F` *is* the intended `f=` mapping) but the generated twin emits `d/dt(central) = … − (CL/F) * central` and `obs_scale = F`, which the twin's own parse rejects as a #993 double use. Fixed by declining the twin when the `f=`/`lagtime=` parameter also fills a disposition role.

The function already anticipates this class in one place — the `f=V1` slot collision dry-runs `ode_param_slots` and declines, with the comment "rather than panicking at eval time" — but it anticipates it *one check at a time*. Three separate guards now exist for three separate ways the twin can fail to parse, and each was added after someone tripped it.

## What to consider

The general fix is to make the twin structurally incapable of panicking: dry-run `parse_model_string(&src)` at the end of `absorption_ode_equivalent_source` and return `None` on error, so a twin that would not parse is simply never offered and the model stays closed-form.

The cost is real and is why #1003 did not just do it: the lazy `OnceLock` exists precisely so "a fit whose subjects never hit the fallback pays no extra parse or allocation" (`AbsorptionOdeEquivalent`'s own doc). A dry-run parse forfeits that for every transit / IG model. Options, roughly in increasing ambition:

1. **Dry-run and discard.** Simplest; one extra full parse per transit/IG model at parse time. Probably milliseconds, but measure before claiming that.
2. **Dry-run and keep.** Build eagerly and store the `CompiledModel` instead of the source, converting "lazy + can panic" into "eager + can decline". Same single parse, no wasted work for models that *do* use the twin; costs the parse for models that never reroute.
3. **Make the failure non-fatal at use time.** Have `effective_for` fall back to the primary when the twin cannot build. Rejected on its face — silently predicting with the closed form for a subject that needs the ODE fallback is exactly the wrong-number-without-a-diagnostic failure the twin exists to prevent — but recorded so the option is not re-litigated.

Option 1 or 2. Whichever is chosen, the acceptance test is the same: construct a model whose twin fails to parse and assert `absorption_ode_equivalent.is_none()` rather than a panic. `dose_attr_param_reused_as_a_disposition_role_declines_the_twin` and `adaptive_observe_does_not_reach_the_absorption_twin` (both added in #1003) are the existing shape to follow.

## Why this is worth doing rather than leaving

Every future ODE-scoped parse check inherits the trap, and the failure mode is the worst kind: it does not show up in the parse tests, it does not show up for the model author, and it appears as an internal panic in someone else's fit only once a subject has a time-varying covariate. The three point guards that exist are load-bearing and undocumented as a set — nothing tells the next person adding an ODE-only check that they have to think about the twin at all.

## Related

- #1003 / #993 — where both instances were found and fixed.
- #735 — the shadow guard and the `f=V1` slot-collision dry-run, the existing precedent for "decline rather than panic".
- #1004, #1005 — the other two follow-ups from the same review.

TeunP:
## Why

`AbsorptionOdeEquivalent::get_or_build` compiled the reconstructed twin source lazily and `.expect()`ed the parse, on the reasoning that "the source is reconstructed from already-validated model blocks, so parsing it is infallible in practice". That reasoning is structurally wrong, and #1008 is the writeup: **the primary is an analytical model and the twin is an ODE model**, so every ODE-scoped parse check runs on the twin and not on the primary. A model this parser accepts can reconstruct into a twin the parser rejects — and the rejection surfaced as an internal panic mid-`fit`/`predict`, the first time a TV-covariate / `TIME` / IOV / SS / infusion subject rerouted.

Three point guards existed for three separate doors, each added after someone tripped it (`f=V1` slot collision, #735; `[adaptive_dosing]` re-emission and a dose-attribute param in a disposition role, both #1003). While measuring for this PR I found a **fourth, live**: an individual parameter named after one of the twin's own states.

```
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V = TVV
  N = TVN
  MTT = TVMTT
  CENTRAL = TVX          # meaningless to the analytical primary — it has no state namespace

[structural_model]
  pk one_cpt_transit(cl=CL, v=V, n=N, mtt=MTT)
```

```
internal: absorption ODE equivalent failed to build: "[odes]: name `CENTRAL` collides with a
previously-declared state (case-insensitive); state, individual-parameter, and ODE-block
intermediate names must all be distinct."
```

Reproduces in three shapes (stray parameter, `v=CENTRAL` disposition role, 2-cpt `PERIPH`). None of the three guards look at state names — which is the point: the set of ODE-only checks is open-ended, so guarding them one at a time loses.

## What changed

**Build the twin eagerly at parse time and decline on failure** (issue's option 2).

- `AbsorptionOdeEquivalent` now holds a `Box<CompiledModel>` instead of `String` + `OnceLock`. `build(&str) -> Result<Self, String>`; `built()` replaces `get_or_build()`; `sync_solver_opts` stamps directly instead of recording an override for a later build (`TwinSolverOpts` is gone).
- The parser builds it at the attach site — after the primary's own blocks have parsed, so a broken primary reports its own error first. On failure it attaches `None` and pushes a `W_ABSORPTION_TWIN_DECLINED` parse warning **carrying the twin parser's own message**, so the reason is readable without a debugger. `check_model_file` maps it to that code (alongside the two existing `W_DERIVED_*` precedents), and `classify_warning` gives it a `WarningCode::AbsorptionTwinDeclined` (`absorption_twin_declined`) rather than the `general` bucket, so the fit-path warning is branchable too — a warning nobody can filter is only half-delivered.
- The three point guards are kept and their comments updated. They now decline *before* reconstructing a source they know is unusable, so those cases stay specifically named rather than falling through to the generic warning.

Declining is not silent-wrong, which is what makes the trade safe: without a twin the model stays closed-form, and every subject that would have needed the fallback is rejected **up front** — `check_absorption_closed_form_support` for TV-covariates / IOV / SS / infusion / resets (an `Err` from `fit()`, a panic with an actionable message from `predict()`/`simulate()`), `check_absorption_flip_flop_no_twin` for the flip-flop regime. Both already keyed on `absorption_ode_equivalent.is_none()`, so a declined twin lands on exactly the path the existing guards' declines land on.

## Alternatives considered

- **Option 1, dry-run and discard** — strictly dominated by option 2: it pays the twin parse, throws the result away, then pays it *again* on the first reroute. Two parses where option 2 pays one.
- **Option 3, non-fatal at use time** (fall back to the primary when the twin can't build) — rejected in the issue and not revisited: silently predicting with the closed form for a subject that needs the ODE fallback is precisely the wrong-number-without-a-diagnostic failure the twin exists to prevent.
- **Dissolving the wrapper** into a bare `Option<Box<CompiledModel>>` on `CompiledModel` — the type now holds one field, but it names the concept and anchors the doc comment that explains this whole trap. Kept.

## Cross-repo dependency

No `pub` API change: `AbsorptionOdeEquivalent`'s methods are `pub(crate)`, and the struct is only ever reached through `CompiledModel`.

## Breaking changes

- [x] None

## Numerical validation

- [x] Not applicable — no numerical path changes. The twin is byte-identical to what the lazy path built; only *when* it is built, and what happens when it cannot be, changed. The closed-form↔ODE equivalence suites (`tests/transit_analytic_equivalence.rs`, `tests/ig_analytic_equivalence.rs`) pin the twin's numbers and are unchanged.

## Performance

- [x] Slower — justified: one extra model parse (**~12 ms**) per transit/IG model, at parse time.

Measured on the `ci-test` profile (release-opt, no LTO), 200 iterations after warm-up:

| model | primary parse | twin parse | twin source | ratio |
|---|---|---|---|---|
| `examples/one_cpt_transit.ferx` | 11982 µs | 12332 µs | 723 B | 103 % |
| `examples/two_cpt_transit.ferx` | 11887 µs | 12341 µs | 904 B | 104 % |
| `examples/one_cpt_ig.ferx` | 12085 µs | 12187 µs | 725 B | 101 % |
| `examples/warfarin.ferx` (no twin) | 12000 µs | — | — | — |

Two things make the ratio less alarming than it looks: the cost is essentially **fixed**, not proportional to the model (twin-less warfarin parses in the same ~12 ms as a 723-byte twin source), and the lazy path paid the same ~12 ms anyway — just later, on the first subject that rerouted. A fit whose subjects never reroute now pays 12 ms it previously avoided.

The **ratio** is the robust number here; treat the absolute µs as order-of-magnitude only. A re-measurement while the machine was running four concurrent cargo builds put the same parse at 41 ms — a 3.4× swing from load alone. Primary and twin were always timed in the same run, so the 101–104 % ratio is unaffected.

## Tests

- [x] Unit tests added (`cargo test --lib`)
- [x] Regression test that fails without this fix

Every new assertion was mutation-checked against a predicted failure set written down first (three mutations behind a temporary env switch, since the machine could not afford three rebuilds; scaffolding removed after):

| mutation | must fail | result |
|---|---|---|
| twin build always returns `Err` | five "twin present" assertions + the message-content assertion | 6 failed, as predicted |
| build failure declines silently | `state_named_…_with_a_warning` only | 1 failed, as predicted |
| #993 disposition-role guard disabled | `dose_attr_param_reused_…` only | 1 failed, as predicted |

The interesting cell is `dose_attr_param_reused_…` **passing** under the always-fail mutation: the #993 guard declines before any twin source is reconstructed, so the forced failure is never reached. Predicted in advance; had it failed, my model of the guard ordering would have been wrong.

`state_named_parameter_declines_the_absorption_twin_with_a_warning` — the new live instance, over all three shapes; asserts `is_none()` **and** that the warning carries the twin parser's reason naming the colliding state. Without this change it panics rather than failing.

`an_ordinary_transit_model_keeps_its_twin_and_warns_about_nothing` — the positive control, so the decline path cannot pass by declining everything.

`transit_twin_with_reserved_f_name_still_builds` strengthened: it used to force the lazy build and assert only "didn't panic", which is now vacuous — it asserts the twin is present and is a working ODE model.

`dose_attr_param_reused_as_a_disposition_role_declines_the_twin` gained an assertion that a guard-recognised decline does **not** fall through to the generic warning, so the guards stay load-bearing rather than becoming dead code shadowed by the backstop.

`classify_warning_absorption_twin_declined_is_warning` pins the taxonomy mapping; the parser test asserts the same mapping on the *produced* message, so the code cannot drift away from the text carrying it.

That second assertion earned its keep immediately: it failed on first run with `left: "flip_flop"`. The decline message *lists* the flip-flop reroute among the features the model has just lost, and `classify_warning`'s `lower.contains("flip-flop regime")` arm claimed it first — so a declined twin would have been reported to users as a flip-flop subject. The new arm is now ordered ahead of it, with a comment saying why (an explicit `W_` token beats a prose match).

- [x] Inline below — no new API surface, but the behaviour change is user-visible, so here is the model that used to panic.

<details>
<summary>A model that panicked, and now warns</summary>

```toml
[individual_parameters]
  CL  = TVCL * exp(ETA_CL)
  V   = TVV
  N   = TVN
  MTT = TVMTT
  CENTRAL = TVX          # collides with the twin's own state name

[structural_model]
  pk one_cpt_transit(cl=CL, v=V, n=N, mtt=MTT)
```

Before: parses and fits fine — until the first subject with a time-varying covariate, at which point `fit()` panics with `internal: absorption ODE equivalent failed to build`.

After: parses, fits, and warns.

```
This absorption model's ODE equivalent could not be built, so the model stays closed-form:
[odes]: name `CENTRAL` collides with a previously-declared state (case-insensitive); …
Subjects needing the ODE fallback (time-varying covariates, a `TIME`-dependent parameter,
IOV, steady-state or infusion doses, or the flip-flop regime) will be rejected with an
explicit error instead of silently rerouting (W_ABSORPTION_TWIN_DECLINED, #1008).
```

A subject that actually needs the fallback then gets the ordinary twin-less rejection, naming the subject and the unsupported feature. Renaming `CENTRAL` restores the twin.

</details>

## Docs

- [x] `docs/model-file/absorption.qmd` — a callout on what a model loses when its twin declines, and how to get it back
- [x] `docs/model-file/iov.qmd` — the twin-less cause list gained the third cause
- [x] `docs/file-formats/check-report.qmd` + the `src/diagnostics.rs` code table — `W_ABSORPTION_TWIN_DECLINED`
- [x] `docs/warnings.qmd` — the `absorption_twin_declined` `WarningCode` row

## Checklist

- [x] `cargo clippy` clean
- [x] No `Dual2` sensitivity-path code touched — this is parse-time wiring; the twin's own ODE/sensitivity path is byte-identical to before

### ferx-site
- [x] No new DSL syntax or `[fit_options]` key
- [x] No new example `.ferx` file, no new data file

### ferx-book

### Example execution
- [x] No example-execution step needed — no user-visible change to any shipped `examples/*.ferx` model. Every example that carries a twin still builds one (asserted by `an_ordinary_transit_model_keeps_its_twin_and_warns_about_nothing` and the strengthened `transit_twin_with_reserved_f_name_still_builds`); the new warning fires only for models that would previously have panicked.

## Reviewer hints

The load-bearing question is whether *declining* is safe everywhere a twin used to be assumed. Five sites read `absorption_ode_equivalent.is_some()` as a scope decision — `api/validation.rs` (×3), `api/postfit.rs`, `estimation/inner_optimizer.rs`, `sens/provider.rs` — and all five take the twin-less branch, which is the reject-or-warn path, not a silent numerical fallback. That is the thing to check independently.

Second: the twin build is now inside `parse_full_model`, so it must not recurse. It cannot — the twin source's `[structural_model]` is `ode(...)`, and `absorption_ode_equivalent_source` matches only `pk one_cpt_transit` / `two_cpt_transit` / `one_cpt_ig` / `two_cpt_ig` — but it is worth confirming.

Third: build placement. The source is still reconstructed early (unchanged bytes); only the *compile* moved to the attach site. Note that the parser's own `sync_ode_solver_opts` runs **before** that attach point, and did so before this change too — the twin gets its tolerances from its own re-emitted `[fit_options]`, and the call-time (#814) override path stamps a twin that is by then attached. `call_time_ode_tolerances_reach_the_absorption_twin` covers both directions.

## Open questions

The per-parse cost is essentially **fixed** rather than proportional to the model, which is odd, so I measured where it goes before speculating in this PR. My hypothesis was per-call `Regex::new` (`model_parser.rs` has 30 sites, most not behind `LazyLock`; `absorption_ode_equivalent_source` alone compiles 4). **That is refuted:** those 4 compile in 942 µs, **2.3 %** of one parse — so even if all 30 ran on every parse it would be well under a fifth of it. The fixed cost is somewhere else and I have not attributed it. Recording the negative result so nobody re-runs the same guess; not filing an issue on an unattributed number.


TeunP:
## Review — xhigh, 10 finder angles + sweep

The core trade is right, and I checked it independently rather than taking the reviewer hints on faith.

**Is declining safe everywhere a twin was assumed?** Yes. `effective_for` reroutes on exactly five conditions — `compiled_model_uses_time_builtin`, `has_tv_covariates`, `n_kappa > 0`, `has_periodic_ss_dose`, `any(is_infusion)` — and `check_absorption_closed_form_support` rejects each of them under `absorption_ode_equivalent.is_none()` (its `dose.ss` arm is a superset of `has_periodic_ss_dose`). The flip-flop reroute in `effective_model_for_eval` is covered by `check_absorption_flip_flop_no_twin` at η = 0 and by `absorption_flip_flop_ebe_warning` post-fit. The other two `is_some()` readers flip *safely*: `is_ode_iov` (inner_optimizer.rs:540) only downgrades a Nelder-Mead-skip heuristic on a combination `fit()` already rejects, and `iov_sens_supported` falls through to `iov_analytical_supported`, which excludes transit/IG by `pk_model` — so it returns `false` and routes to FD, not to a wrong analytic gradient.

**No recursion**: confirmed — the twin's `[structural_model]` is `ode(...)`, so `absorption_ode_equivalent_source` returns at the four `pk …` regexes.

**No new I/O**: a `[data]` block is stored as a path string at parse time, so the eager build adds no file reads.

**`WarningCode` is `#[non_exhaustive]`**, so the new variant really is non-breaking — the "no version bump" call is correct.

Nine findings, most severe first.

---

### 1. A declined twin gets a rejection message that asserts a cause list it does not belong to — and the real reason is dropped on exactly that path (correctness)

`src/api/validation.rs:1801`, plus the five `is_none()` arms at 1235 / 1250 / 1264 / 1330 / 1345.

The PR's safety argument is "a subject that needs the fallback is rejected up front with an actionable message". It *is* rejected — but the message describes a different model:

- flip-flop: *"This model has no ODE twin to fall back on (a user `[odes]` / `[scaling]` / `[initial_conditions]` block, or a parameter whose name collides with a reserved `f`/`lagtime` slot, declines the desugar…)"* — none of which is true for `CENTRAL = TVX`.
- SS / infusion: *"this model has no twin **(an unrecognised closed form)**"* — it is a perfectly recognised closed form.
- IOV / `TIME`: *"this form is outside the automatic ODE-equivalent rewrite"* — it isn't; the rewrite ran and its output was rejected.

And the reason that *would* be actionable never reaches the user on this path: `fit()` reads `model.parse_warnings` at `fit.rs:997`, but `check_absorption_closed_form_support` returns `Err(e)` at `fit.rs:276`, long before. `predict()` / `simulate()` panic with the assert message. The only entry point that shows the decline reason is `ferx check`. So the user is told to rewrite the model as an explicit `[odes] transit()` forcing when the fix is to rename one parameter.

Fix: carry the decline reason into those messages.

### 2. `classify_warning`'s token arm sits below four prose arms, while the message it matches interpolates arbitrary text (correctness)

`src/types.rs:4585`.

The message is `"…stays closed-form: {e} Subjects needing…"`, where `{e}` is *whatever the twin parser said*. The chain tests `"did not converge"`, `"covariance step failed"` / `"covariance failed"`, `"covariance step regularized"` / `"off-diagonal fd stencil"`, and `"ill-conditioned"` / `"condition number"` before reaching the new arm — so a twin error carrying any of those substrings classifies as e.g. `condition_number` at **Critical**.

This is the same failure the PR already tripped once (`flip_flop` claimed the message) and fixed by ordering against *that one arm*. The general form of the fix is the one its own comment argues for — an explicit `W_` token beats every prose match — so move the arm to the head of the chain.

### 3. The load-bearing safety claim has no test (test-coverage)

`src/parser/model_parser_tests.rs:2044`.

All six new / strengthened assertions stop at `absorption_ode_equivalent.is_none()` plus warning text. Nothing exercises the actual claim: declined twin + a subject that needs the fallback ⇒ rejected, not panicking and not silently closed-form. The decline is safe *because* the guards key on `is_none()`; that coupling is exactly what a future reroute condition added to `effective_for` without a matching guard arm would break, and the suite would stay green.

### 4. Four `Regex::new` per call, now called twice per parse (efficiency)

`src/parser/model_parser.rs:8255`.

The twin's own `parse_full_model` reaches `absorption_ode_equivalent_source`, compiles `transit_one` / `transit_two` / `ig_one` / `ig_two`, matches none, and returns `None`. Per this PR's own measurement that is 942 µs of pure regex compilation, now paid twice per transit / IG model. `DIFFUSION_LINE_RE` (model_parser.rs:12) is the `static … LazyLock<Regex>` pattern already in the file.

(Aside on the open question: this does not explain the ~12 ms fixed cost either — it is ~8 % of one parse, not the bulk of it.)

### 5. The twin's error is spliced mid-sentence (correctness)

`src/parser/model_parser.rs:3095`. `"…closed-form: {e} Subjects needing…"` assumes `e` ends in a period. Most parser errors do not, so the common rendering runs two sentences together. The tests miss it because the one error they use happens to end in `.`.

### 6. The field doc still describes the lazy twin (conventions)

`src/types.rs:3175`. `CompiledModel::absorption_ode_equivalent` — the doc a reader hits first — still says *"built lazily so it reuses the whole ODE prediction/sensitivity path unchanged and costs nothing for a fit whose subjects never need it"*, and lists `None` as "not a closed-form absorption model, or a form outside the desugar's scope". Both are now false. A reader trusting it concludes `None` can only be a scope limit — the exact inference #1008 exists to correct.

### 7. The twin-less cause enumerations were not updated (conventions)

`src/pk/mod.rs:120` (doc) and `:134` (the `let Some(twin) = … else` comment); the same closed list at `src/api/validation.rs:1230`, `:1245`, `:1261`. The PR's thesis is that this cause set is open-ended; leaving closed enumerations in place re-arms the reasoning error for the next reader.

### 8. Both branches of the tolerance test now run one code path (simplification)

`src/types_tests.rs:2291`. The branches existed to separate "override applied inside `get_or_build`" from "`built.get_mut()` stamp on an already-built twin". With `TwinSolverOpts` and the `OnceLock` gone there is one path, so "Sync-then-read" and "Read-then-sync" assert the same call twice — and read as covering an ordering distinction that no longer exists.

### 9. The docs' silent-decline list omits the #735 guards (conventions)

`docs/model-file/absorption.qmd:284`. The parenthetical names the user-block case and the #993 double use, but not the reserved-`f`/`lagtime` shadow guard (model_parser.rs:8390) or the `ode_param_slots` collision dry-run (:8425) — both of which decline with no warning at all. A user hunting for why their transit model rejects IOV will not find their case.

---

Not findings, recorded so they are not re-derived: the parse-time `sync_ode_solver_opts` at model_parser.rs:2672 running before the attach site is unchanged behaviour (the twin took its tolerances from its own re-emitted `[fit_options]` before this PR too); discarding the twin's own `parse_warnings` is right (they would be near-duplicates of the primary's); and `Box<CompiledModel>` is strictly *less* restrictive for auto-traits than `OnceLock<Box<CompiledModel>>`, so there is no `Sync` regression.

TeunP:
## Review follow-up — all nine addressed in `9aad6824`

Fixes for every finding in [the review above](https://github.com/FeRx-NLME/ferx-core/pull/1027#issuecomment-5379638991) are pushed. What changed, in the review's order:

**1. Declined-twin rejections asserted a false cause list.** The decline message is now built and read back through one pair of functions in `types.rs` — `absorption_twin_declined_warning` / `absorption_twin_decline_reason` — sharing a single marker constant, so the twin parser's reason round-trips verbatim. `api/validation.rs` gained `twin_less_cause` (the short parenthetical: `(its twin was built and declined, #1008)` vs `(an unrecognised closed form)`) and `twin_decline_clause` (the reason itself), threaded through all six twin-less rejection sites including the flip-flop one. A declined-twin TV-covariate model now gets:

> `one_cpt_transit` does not support within-subject time-varying covariates (subject TV1): … and this model has no ODE twin **(its twin was built and declined, #1008)** to reroute to. Use an ODE absorption model. **Note: this model is not outside the rewrite's scope — its twin was built and then rejected by its own parse: `[odes]`: name `CENTRAL` collides with a previously-declared state … Fixing that restores the twin and this feature, and is usually easier than the rewrite suggested above.**

On the reason's storage: it stays in `parse_warnings` rather than a new `CompiledModel` field, because `CompiledModel` is not `#[non_exhaustive]` — a public field would be a breaking change for struct-literal construction, which is out of proportion for a diagnostic. That makes the `W_ABSORPTION_TWIN_DECLINED` token load-bearing, so writer and reader now sit next to each other with the contract documented.

**2. `classify_warning` arm ordering.** Moved to the head of the chain, ahead of every prose arm rather than just ahead of the one that collided. `classify_warning_twin_decline_beats_prose_arms_in_the_reason` pins four reasons that each contain a substring a later arm matches (`ill-conditioned`, `did not converge`, `covariance failed`, `condition number`).

**3. The safety claim is now tested.** Each of the three declined shapes asserts that `check_absorption_closed_form_support` *rejects* a TV-covariate subject and that the rejection quotes the colliding name — reached directly rather than through `fit()`, so it stays Tier 1. `an_ordinary_transit_model_keeps_its_twin_and_warns_about_nothing` gained the negative control: a twin-carrying model must accept the same subject, so the positive assertion cannot pass by rejecting everything.

**4. Regex hoisting.** The four patterns are `static … LazyLock<Regex>`, matching `DIFFUSION_LINE_RE`.

**5. Sentence splice.** The reason is trimmed and given a terminator when it lacks one; both cases pinned.

**6/7. Stale docs.** The `absorption_ode_equivalent` field doc is rewritten (eager build, three-cause `None` list, pointer to the reason accessor, explicit "do not treat `None` as outside the desugar's scope"). The closed cause enumerations in `pk/mod.rs` and three `api/validation.rs` comments are replaced by a pointer to it.

**8. Tolerance test.** The second branch now re-syncs with different tolerances and pins second-sync-wins, which the first branch does not cover.

**9. Docs.** The silent-decline list gained the #735 reserved-slot shadow and slot-collision guards.

### Verification

Every new assertion was mutation-checked against a failure set written down before running. All six matched exactly:

| mutation | must fail | result |
|---|---|---|
| decline reason unavailable (`absorption_twin_decline_reason` → `None`) | the rejection assertion + the round-trip test | 2, as predicted |
| decline arm moved back below the prose arms | the prose-collision test only | 1, as predicted |
| terminator normalisation dropped | the round-trip test only | 1, as predicted |
| TV-cov arm drops its `is_none()` guard | the negative control only | 1, as predicted |
| `twin_less_cause` always generic | the "not unrecognised" assertion only | 1, as predicted |
| `twin_decline_clause` empty | the "quotes the reason" assertion only | 1, as predicted |

The interesting cell is the last two both landing on the *same* test but different assertions — which is why the rejection is asserted on content (the colliding name) and on the absence of "an unrecognised closed form" separately, rather than on the wrapper wording. Scaffolding was a single env-gated switch, removed after; `grep` confirms it is gone.

`cargo test --lib --features ci`: 3126 passed, 0 failed, 19 ignored. `cargo fmt --check` clean. Clippy warning count unchanged at 264 before and after, so no new lint.

### Two things left for the author

- **The PR body is now partly stale.** The Tests section lists the original three-mutation table, and the Reviewer-hints line saying a declined twin gets "the ordinary twin-less rejection, naming the subject and the unsupported feature" is exactly what finding 1 corrected — it now also names the cause and quotes the reason.
- **CHANGELOG placement, deliberately not touched.** `## [Unreleased]` carries duplicate `### Changed` / `### Added` / `### Fixed` sections and this PR's entry sits in the lower `### Fixed` while recent entries land in the upper one. `origin/main` has the same duplication, so it is pre-existing rather than introduced here; moving the entry would conflict with the other in-flight branches for a cosmetic gain. Flagging rather than fixing.

TeunP:
## Verification update — re-run with `survival` on

Correcting the scope of my previous comment's numbers. It cited `--features ci`, which is narrower than it looks: `ci = []` is a no-op feature and `survival` is a *separate* default-off gate, so those runs compiled **none** of the `#[cfg(feature = "survival")]` code — including 42 gated blocks in the four files this PR touches, three of them in `api/validation.rs` immediately after `check_absorption_closed_form_support`.

Re-run at `9aad6824` with both gates on:

| check | result |
|---|---|
| `cargo check --tests --features ci,survival` | exit 0 |
| `cargo test --lib --features ci,survival` | **3246 passed, 0 failed, 19 ignored** |
| `cargo clippy --lib --features ci,survival` | 273 warning sites |
| `cargo fmt --all -- --check` | clean |

3126 → 3246 confirms 120 survival-gated tests genuinely executed this time.

`cargo check --tests` also closes a gap I had left open in the first pass: the integration tests under `tests/` had never been compiled against these changes. They compile.

On clippy I did not want to trust a bare count, so I intersected all 273 warning locations against the 353 lines this commit changed. Exactly one overlap — `model_parser.rs:8291`, the pre-existing `this block may be rewritten with the ? operator` lint on the `if let … else if let` chain. It sat at `:8293` before the `LazyLock` statics shifted it down two lines, and the `ig_two` → `IG_TWO` rename is what put it on a changed line. Not introduced here; **no new lints**.

Unrelated, noted in passing since the survival build surfaced it: `unused_mut` at `src/api/run.rs:867`. Came in with `ab87eec6` (#957) on main, not a file this PR touches, so left alone.

Patch coverage against the `codecov/patch` gate is still measuring locally (CI-equivalent invocation: `cargo llvm-cov --tests --no-default-features --features ci --profile ci-fast`); I will add it here when it lands.

### Still not verified, so the picture stays honest

- **Tier 3 slow tests** — notably `tests/transit_analytic_equivalence.rs` and `ig_analytic_equivalence.rs`, the suites that would actually *prove* this PR's "the twin is byte-identical, only *when* it is built changed" claim. Nightly by design. The reasoning holds (same reconstructed source, same parse) but I have not executed the proof.
- CI has not yet reported on `9aad6824`, and no second human has reviewed the fixes.

TeunP:
## Patch coverage — 96.70%, and a thin margin found and fixed on the way

Measured locally with the CI-equivalent invocation (`ci.yml:259`):
`cargo llvm-cov --tests --no-fail-fast --no-default-features --features ci --profile ci-fast --lcov`.

Scored over the **whole PR diff vs `origin/main`** (16 files, 895 insertions) — the range the `codecov/patch` status uses — by intersecting the `lcov` `DA:` records against the post-image lines of `origin/main...HEAD`:

```
file                                   exe   cov     pct   missed
src/api/validation.rs                   41    38   92.7%   1302, 1303, 1845
src/parser/model_parser.rs              17    17  100.0%
src/pk/mod.rs                            1     1  100.0%
src/sens/provider.rs                     1     1  100.0%
src/types.rs                            31    31  100.0%

PATCH: 88/91 = 96.70%   (gate: >= 90%)
```

**The margin was the real finding.** At `9aad6824` this was **90.16%** — over the gate by 0.16 points, i.e. one further uncovered line from failing. All six uncovered lines were in `api/validation.rs`, and all six were the rejection messages the review fix had just rewritten: the `TIME` arm (1302–1303), the `no_twin_cause` slot in the SS message (1393), the infusion message (1407–1408), and the flip-flop decline cause (1845).

That was not just a number. My tests covered the TV-covariate rejection only — but a declined twin costs the model its **SS and infusion** reroutes (#719) by exactly the same mechanism, through their own message arms, and an author who hit one of those first was covered by nothing. `3649196c` drives all three declined shapes through an SS-dose and an infusion subject as well, asserting each rejection names the decline cause and quotes the twin parser's reason.

Both new populations are built by mutating the TV-covariate one, so each assertion first pins *which* arm fired (`steady-state (SS) doses` / `infusion doses`) — without that they would pass just as happily on the TV-covariate message and prove nothing. The coverage re-measurement is the independent confirmation: 1393 / 1407 / 1408 flipped from missed to covered, which happens only if the intended arms really ran.

Remaining uncovered: 1302–1303 (the `TIME`-dependent-parameter rejection) and 1845 (the flip-flop decline cause). Both need a fourth and fifth model shape — a declined twin that also uses the `TIME` built-in, and one that is also in the flip-flop regime at η = 0. Reachable, but past the point of proportionate return at 96.7%; flagging rather than adding.

### Tier 2 also ran

`cargo llvm-cov --tests` builds and runs the integration tests, so this pass exercised `tests/*.rs` as well — **136 test binaries, 6821 passed, 0 failed, 148 ignored** (the ignored ones are the `slow-tests`-gated Tier 3 fits). That closes the "Tier 2 never ran" caveat from my previous comment.

Tier 3 remains un-run, by design — including the closed-form↔ODE equivalence suites that would *prove* the "twin is byte-identical" claim rather than argue it. Those are nightly.

### Current state

`3649196c`. `cargo test --lib --features ci`: 3126 passed, 0 failed. `--features ci,survival`: 3246 passed, 0 failed. `cargo check --tests --features ci,survival`: exit 0. `cargo fmt --check` clean. No new clippy warnings.

PR #1033, opened 2026-08-21
feat(api): guard imp/impmap/bayes and gn at n_eta = 0 in ferx check [closes #1007, closes #1006]

### LINKED ISSUE #1006: warn(gn): Gauss-Newton returns a wrong answer at n_eta = 0 with no diagnostic (follow-up to #989)
Follow-up from the review on #1002 (#989).

## What happens

`method = gn` on a fixed-effects-only model (`n_eta = 0`) returns a badly wrong answer and reports nothing beyond `Converged: NO`. On `examples/one_cpt_iv_pooled.ferx` + `data/one_cpt_iv.csv`, with the example's own `sigma PROP_ERR ~ 0.02 (sd)` start:

| method | Converged | OFV | TVCL | SE TVCL |
|---|---|---|---|---|
| `foce` / `focei` / `laplace` / `gn_hybrid` | YES | **−269.6370** | 4.8408 | — |
| `gn` | NO | **8670.0458** | 4.1258 | 286.4 (RSE 6942%) |

8940 OFV units off. From a `0.13 (sd)` start `gn` does converge, to −269.6322 — the right basin, still 0.005 short of FOCEI.

## Why

BHHH's outer-product Hessian over-estimates curvature far from the optimum. In a mixed-effects fit the inner EBE loop absorbs a mis-specified `sigma` and keeps the outer iterate near the basin; at `n_eta = 0` there is no inner loop, so the objective starts orders of magnitude above the optimum (11374 here) and the LM-damped step collapses.

## Why it's worth a guard

The bar is currently inconsistent. #1002 **hard-rejects** `method = saem` at `n_eta = 0` (`E_SAEM_NO_RANDOM_EFFECTS`) because SAEM lands ~0.04 OFV units short of the optimum. `gn` lands 8940 units off and ships the result silently.

## Suggested fix

A `W_GN_NO_RANDOM_EFFECTS` warning from `check_model_options` when `chain` contains `FoceGn` and `model.n_eta == 0`, pointing at `gn_hybrid` / `focei`. Warning rather than error, since `gn` *does* reach the optimum from a good `sigma` start — the failure is start-sensitivity, not an invalid objective.

Worth considering alongside: a post-fit warning whenever a GN run at `n_eta = 0` ends `Converged: NO`, which is the state that actually predicts the bad answer.

## Current state

Documented, not guarded — `docs/model-file/parameters.qmd` ("`gn` is not recommended at `n_eta = 0`") and `docs/estimation/gauss-newton.qmd#fixed-effects-only-models` both carry the measured numbers as of #1002.

## Test

Tier-2, evaluation-free: assert `check_model_options` emits the code for `gn` / `gn_hybrid`-free chains at `n_eta = 0` and stays silent at `n_eta > 0`. The convergence behaviour itself is already pinned in the docs; a slow-gated test asserting `gn` misses the optimum would be pinning a bug rather than a contract.

### LINKED ISSUE #1007: check: imp / impmap / bayes reject n_eta = 0 only at run time, not in ferx check (follow-up to #989)
Follow-up from the review on #1002 (#989).

## What happens

`imp`, `impmap` and `bayes` all refuse a model with no random effects, but only **at run time**:

- `src/estimation/importance_sampling.rs:113` — `"Importance sampling requires at least one random effect…"`
- `src/estimation/importance_sampling.rs:532` — same, mixture entry point
- `src/estimation/bayes.rs:452` — `"Bayesian estimation requires at least one random effect (eta)"`

So `ferx check` passes a model it knows cannot fit:

```console
$ ferx check p_imp.ferx --json      # method = imp, n_eta = 0
{ "valid": true, "model": "p_imp", "diagnostics": [] }
```

And a chain pays for the whole preceding stage before failing:

```
methods = [focei, imp]      # runs FOCEI to convergence, then errors
```

## Why it matters

#1002 added `E_SAEM_NO_RANDOM_EFFECTS` in `check_model_options` for exactly this situation, with exactly this rationale:

> check it *here* so `ferx check` catches it and a `methods = [saem, focei]` chain fails up front rather than mid-run

That argument applies verbatim to the other three, and they were left out only to keep #1002 scoped. The result is an inconsistency a user will hit: `saem` at `n_eta = 0` fails instantly with a diagnostic code, `imp` at `n_eta = 0` fails after the fit has run.

## Suggested fix

Extend the guard in `check_model_options` (`src/api/validation.rs:1771`) to cover `Imp`, `Impmap` and `Bayes`. The existing code is `E_SAEM_NO_RANDOM_EFFECTS`, which is documented as stable in `docs/file-formats/check-report.qmd`, so it should not be renamed — either keep it for `saem` and add a sibling `E_METHOD_NO_RANDOM_EFFECTS` for the other three, or add the sibling for all four and leave the old code emitting alongside for `saem`. Worth a moment's thought about which reads better for tooling consumers.

The run-time errors should stay as the backstop for direct `fit()` callers who bypass `check_model_options`.

## Also update

`docs/estimation/saem.qmd` currently spells the split out explicitly (added in #1002):

> `imp`, `impmap` and `bayes` refuse `n_eta = 0` for the same reason, but they do so **at run time rather than in `ferx check`**…

That paragraph becomes wrong once this lands, and `docs/file-formats/check-report.qmd` needs the new code if one is added.

## Test

Tier-1 in `src/api/tests/zero_omega_tests.rs`, alongside the existing SAEM guard tests: each of the three rejected at `n_eta = 0`, each allowed at `n_eta > 0`, and the chain case (`methods = [focei, imp]`) rejected up front.

TeunP:
## Why

Two follow-ups from the review on #1002 (#989), both closing the same gap: `ferx check` reported a model as valid that the engine already knows it cannot fit well.

**#1007** — `imp`, `impmap` and `bayes` all refuse `n_eta = 0`, but only **at run time**. So `ferx check` passed the model, and `methods = [focei, imp]` ran the entire FOCEI stage to convergence before erroring. #1002 added `E_SAEM_NO_RANDOM_EFFECTS` in `check_model_options` for exactly this situation, with exactly this rationale; the other three were left out only to keep #1002 scoped. The result was an inconsistency a user hits directly: `saem` at `n_eta = 0` fails instantly with a diagnostic code, `imp` at `n_eta = 0` fails after the fit has run.

**#1006** — `method = gn` on a fixed-effects-only model returns a badly wrong answer with no diagnostic beyond `Converged: NO`. On `examples/one_cpt_iv_pooled.ferx` + `data/one_cpt_iv.csv`, from the example's own `sigma PROP_ERR ~ 0.02 (sd)` start:

| method | Converged | OFV |
|---|---|---|
| `foce` / `focei` / `laplace` / `gn_hybrid` | YES | **−269.6370** |
| `gn` | NO | **8670.0458** |

8940 OFV units off. BHHH's outer-product Hessian over-estimates curvature far from the optimum; at `n_eta = 0` there is no inner EBE loop to absorb a mis-specified `sigma`, so the objective starts orders of magnitude above the optimum and the LM-damped step collapses. The bar was inconsistent: #1002 hard-rejects `saem` at `n_eta = 0` because it lands ~0.04 OFV units short, while `gn` lands 8940 units off and shipped the result.

## What changed

**#1007** — `check_model_options` now rejects `Imp` / `Impmap` / `Bayes` at `n_eta = 0` under a new `E_METHOD_NO_RANDOM_EFFECTS`, firing anywhere in a method chain. `E_SAEM_NO_RANDOM_EFFECTS` is documented as stable in `docs/file-formats/check-report.qmd`, so it keeps its code and the three additions share one new sibling code (the second of the two options the issue offered). The run-time errors stay as the backstop for direct `fit()` callers who bypass `check_model_options`.

> One correction to the issue text: it lists three run-time guard sites. There are **four** — `src/estimation/impmap.rs:490` has its own, reached by the estimating `imp` / `impmap` stages. All four are left in place.

**#1006** — `check_model_options` emits `W_GN_NO_RANDOM_EFFECTS` for `chain.contains(FoceGn) && n_eta == 0`, pointing at `gn_hybrid` / `focei`. A **warning**, not an error, since `gn` does reach the optimum from a good `sigma` start — the failure is start-sensitivity, not an invalid objective. This also takes up the issue's "worth considering alongside": a pure-GN run that ends unconverged at `n_eta = 0` now pushes a matching post-fit warning from `gauss_newton.rs`, since `Converged: NO` is the state that actually predicts the bad answer. `gn_hybrid` is exempt from both — its FOCEI polish recovers the optimum.

## Alternatives considered

For #1007, the issue offered two shapes: rename/extend the existing code, or add a sibling. Renaming is out — `E_SAEM_NO_RANDOM_EFFECTS` is documented as stable and tooling may match on it. Emitting the new sibling for all four (with the old code alongside for `saem`) would mean `saem` produces two diagnostics for one problem, which reads worse for a consumer counting errors. So: `saem` keeps its code, the three new methods share the sibling.

For #1006, an error rather than a warning was considered and rejected for the reason the issue gives — `gn` reaches the optimum from a good start, so hard-rejecting it would block a working configuration.

## Cross-repo dependency

No public Rust API change; the new diagnostics flow through the existing `CheckReport` / `FitResult.warnings` channels the R wrapper already surfaces.

## Breaking changes

- [x] None

A model that was previously *accepted* by `ferx check` and then failed mid-fit is now rejected up front. That is the point of #1007, and the fit-time outcome is unchanged (still an error, just earlier).

## Numerical validation

- [x] Not applicable (validation / diagnostics only — no numerical path touched)

The measured numbers motivating the `gn` warning are already committed in `docs/model-file/parameters.qmd` and `docs/estimation/gauss-newton.qmd` as of #1002, and are quoted above.

## Performance

A `methods = [focei, imp]` chain at `n_eta = 0` now fails before running FOCEI instead of after — strictly less work in the case it applies to.

## Tests

Tier-1, in `src/api/tests/zero_omega_tests.rs` alongside the existing SAEM guard tests:

- each of `imp` / `impmap` / `bayes` rejected at `n_eta = 0`, with the message naming the offending method and *not* reusing the stable SAEM code
- the chain case (`methods = [focei, imp]`) rejected up front
- all three allowed at `n_eta > 0`
- `gn` warns at `n_eta = 0`, with severity asserted as `Warning` (so `ferx check` stays `valid: true`)
- negative controls: `gn_hybrid` at `n_eta = 0` and `gn` at `n_eta > 0` must not warn

Plus Tier-1 in `src/estimation/gauss_newton.rs` for the post-fit warning: an unconverged pure-GN run on `examples/one_cpt_iv_pooled.ferx` carries it, while `gn_hybrid` on the same budget and a mixed-effects `gn` run do not.

Every new test was mutation-verified — disabling either guard, and making the GN post-fit warning fire unconditionally or never, each fails the expected tests and nothing else.

## Example

- [x] Not applicable (diagnostic surface only; no new API or DSL)

```console
$ ferx check p_imp.ferx --json      # method = imp, n_eta = 0
# before: { "valid": true, "model": "p_imp", "diagnostics": [] }
# after:  E_METHOD_NO_RANDOM_EFFECTS — method = imp requires at least one random effect (n_eta = 0)...
```

## Docs

- `docs/file-formats/check-report.qmd` — rows for both new codes.
- `docs/estimation/saem.qmd` — the paragraph stating that `imp`/`impmap`/`bayes` refuse `n_eta = 0` *at run time rather than in `ferx check`* is now false and has been rewritten (the issue flagged this explicitly).
- `docs/model-file/parameters.qmd` and `docs/estimation/gauss-newton.qmd` — record that the `gn` combination is now flagged rather than left to the reader.

## Checklist

- [x] `cargo clippy` clean (4 warnings in the tree, all on pre-existing lines, none in this diff)
- [x] Not on the `Dual2` sensitivity path

### ferx-site / ferx-book
- [x] No new DSL syntax, `[fit_options]` key, example, or data file — no site/book change needed.

### Example execution
- [x] No example execution step needed (diagnostics only; no example output changes)

## Reviewer hints

The substance is two blocks in `src/api/validation.rs:1824-1884` and one in `src/estimation/gauss_newton.rs`. Worth checking:

- that `E_SAEM_NO_RANDOM_EFFECTS` is genuinely untouched (tests assert the new methods do not emit it, and that `saem` does not emit the new one);
- the `gn_hybrid` exemption, which appears in both places and must agree — `check_model_options` keys off `chain.contains(&FoceGn)` while the post-fit warning keys off `!matches!(options.method, FoceGnHybrid)`, because during a hybrid run the GN phase sees `method = FoceGnHybrid`;
- one deliberate side effect, called out in the code comment: an `imp_eval_only` chain at `n_eta = 0` previously downgraded the IMP failure to a post-fit warning and still returned a `FitResult`. It is now rejected up front instead. Requesting an IMP objective on a model with nothing to integrate over is a spec contradiction, so failing before the fit reads better than fitting and warning — but it is a behaviour change and worth a second opinion.

## Open questions

None blocking. The `E_METHOD_NO_RANDOM_EFFECTS` naming choice (one shared code for the three methods, `saem` keeping its own) is the one judgement call the issue explicitly left open — happy to switch to a per-method code if that reads better for tooling consumers.


TeunP:
## Review — xhigh

Read the diff hunk-by-hunk plus the enclosing functions (`check_model_options`, `run_foce_gn`, `fit_inner`'s stage loop), traced the callers, and ran the new tests locally. The two guards are the right shape and the `gn_hybrid` exemption reasoning holds up — I reproduced it: on `examples/one_cpt_iv_pooled.ferx` the hybrid's polish goes `11356.5058 → −269.6370` from the same 2-iteration GN phase that the pure-`gn` control leaves unconverged.

Twelve findings. Three are substantive; the rest are housekeeping. I pushed fixes for nine of them to this branch (`7c2d9324`) — details under each.

### Substantive

**1. `methods = [gn, focei]` gets both #1006 diagnostics, although the FOCEI stage polishes it** — `src/estimation/gauss_newton.rs:378`, `src/api/validation.rs:1938`

The exemption is stated as "`gn_hybrid` is exempt, since its FOCEI polish recovers the optimum". A hand-written `methods = [gn, focei]` chain *is* that polish, spelled out, and gets neither exemption:

- check time: `chain.contains(&EstimationMethod::FoceGn)` fires on any chain containing `gn`;
- post-fit: `fit.rs` sets `stage_opts.method = method` and `stage_opts.methods = Vec::new()` before each stage, so during the GN stage `options.method` is `FoceGn` (not `FoceGnHybrid`) *and the chain is gone*. The warning is pushed, and `fit.rs` accumulates every stage's warnings into the final `FitResult` — so a chain that converges to the true optimum still ships "`[FOCE-GN]` … the result may be badly wrong, not just imprecise".

Fixed. The rule is now "`gn` is the **last estimating stage**", which subsumes both spellings, via the existing `api::postfit::is_last_estimating_stage` (so a trailing eval-only IMP still counts `gn` as last — covered by a test). The post-fit half can't see the chain from inside `run_foce_gn`, so the warning text is now a named const (`GN_ZERO_ETA_NONCONVERGENCE_WARNING`) and `fit_inner` drops that exact string from a non-terminal GN stage through `keep_gn_zero_eta_warning`, which is unit-tested on its own.

**2. `W_GN_NO_RANDOM_EFFECTS` never reaches `fit()`** — `src/api/validation.rs:1940`

It's the first warning-severity diagnostic `check_model_options` has ever produced (checked: no `Diagnostic::warning` anywhere in that function on `main`), and `fit_inner` consumes the function through `first_error(...)`, which discards non-errors. So `ferx check` reports the warning and `fit()` — and the R wrapper — silently do not. The post-fit backstop only fires on `!converged`, so a `gn` run at `n_eta = 0` that *does* report converged gets no diagnostic at all from the fit path. And every future warning added to this function would inherit the same silent drop.

Fixed: `fit_inner` now pushes the warning-severity option diagnostics into `accumulated_warnings`, exactly the way `check_model_data_warnings` is already handled one line above. (It classifies as `general`/`Warning` via `classify_warning`; the post-fit one classifies as `convergence`/`Critical` on "did not converge", which is right.)

**3. The `imp_eval_only` behaviour change is untested and unlogged** — `src/api/tests/zero_omega_tests.rs`, `CHANGELOG.md`

The reviewer hints flag it and ask for a second opinion — I agree with the call (requesting an IMP objective with nothing to integrate over is a spec contradiction; failing before the fit is better than fitting and warning). But `check_model_options_rejects_imp_anywhere_in_a_chain` leaves `imp_eval_only` at its `false` default, so nothing pins the new behaviour: re-exempting the eval-only path — the obvious way to answer your own open question later — passes every test. And a call that returned `Ok(FitResult)` now returns `Err`, which the CHANGELOG entry doesn't mention while "Breaking changes: None" is ticked.

Fixed: added `check_model_options_rejects_eval_only_imp_without_random_effects` (mutation-verified) and a sentence to the CHANGELOG entry.

### Housekeeping

**4. New codes missing from the `src/diagnostics.rs` registry.** That module's doc says: *"Codes are stable identifiers (prefix `E_` for errors, `W_` for warnings). Add new codes here and document them in `docs/file-formats/check-report.qmd`."* The `.qmd` half was done, the registry half wasn't (`E_SAEM_NO_RANDOM_EFFECTS` from #1002 is missing too). **Fixed** — all three rows added.

**5. The SAEM block and the new loop are the same shape twice.** **Fixed** — one table of `(method, name, code, rationale)`; `saem` keeps `E_SAEM_NO_RANDOM_EFFECTS`, the three additions share the sibling code, and both messages come out byte-identical to before (the per-method sentence is the only text that varies). Adding a fifth method to the family is now one row.

**6. No coverage for the singular `method = imp` form.** Every new test drives `methods = [...]`, i.e. `method_chain()`'s non-empty branch; the `method = imp` spelling a user actually writes takes the other branch. **Fixed** — added a test that asserts `opts.methods.is_empty()` first, then checks all three methods (mutation-verified: swapping `chain.contains` for `options.methods.contains` fails only this test).

**7. Altitude — "requires ≥1 random effect" now lives in six places.** The check plus five run-time guards: `importance_sampling.rs:113` and `:532`, `impmap.rs:488` and `:1490`, `bayes.rs:452`. (Incidentally: the PR body says the issue lists three run-time sites and there are actually four — there are five.) **Partially addressed**: the check side is one table now and the comment names all five, but the deeper fix — an `EstimationMethod::requires_random_effects()` predicate that the run-time guards also consume, so a new integrating method can't be half-wired — I left alone as out of scope for a diagnostics PR. Worth a follow-up issue.

**8. Docs.** `check-report.qmd` and `gauss-newton.qmd` said "`method = gn` at `n_eta = 0`" / "`gn_hybrid` is exempt from both"; both now state the last-estimating-stage rule and name the `[gn, focei]` case. **Fixed.**

### Not fixed — flagging only

**9. `"Gauss-Newton: max iterations reached without convergence"` is pushed on every non-converged exit**, including the two `break`s that already pushed their own reason. A trust-radius collapse at iteration 5 of 100 now reports three warnings — "trust radius collapsed", "max iterations reached", and the new #1006 line — one of which contradicts the other two. Pre-existing (`gauss_newton.rs:369`), but the PR adds a third message to the same block. I left it: `types_tests.rs:729` maps that exact string to `Critical`/`convergence`, so making the push conditional would silently downgrade a collapsed run's reported severity, and that deserves its own change.

**10. `outer_maxiter: 2, // guarantees non-convergence from the example's poor 0.02 (sd) start`** — the budget guarantees it, not the start: the only normal convergence branch is `rel_change < 1e-6 && iter > 3`, so two iterations cannot converge from *any* start. The test verifies the warning **gate**, not the BHHH-collapse regime #1006 is about. **Comment and assertion message corrected**; the test is otherwise fine as a gate test.

**11. One problem, several diagnostics.** `methods = [saem, imp]` at `n_eta = 0` emits `E_SAEM_NO_RANDOM_EFFECTS` *and* `E_METHOD_NO_RANDOM_EFFECTS`; `[imp, impmap, bayes]` emits three copies of the latter. That's the outcome "Alternatives considered" rejects for `saem`. Defensible as-is (three methods, three problems) — left alone, but if a consumer counts errors it's worth knowing.

**12. On the naming question in "Open questions":** one shared `E_METHOD_NO_RANDOM_EFFECTS` reads right to me. The method name is in the message and the block is `fit_options`; a per-method code would give tooling nothing it can't get from parsing that, at the cost of three more stable identifiers to keep forever.

### Verification

`cargo test --lib --features ci` green (3131 passed, 19 ignored, 0 failed), `cargo fmt` clean, and all three CI compile gates pass (`cargo check --tests --no-default-features` with `ci`, `ci,survival,slow-tests`, `ci,markov`). `cargo clippy --no-default-features --features ci,markov` reports nothing on any changed line. Every new test mutation-verified individually: reverting the gn gate to `chain.contains`, weakening it to `chain.last()`, re-exempting eval-only IMP, making `keep_gn_zero_eta_warning` unconditional, and pointing the guard at `options.methods` each fail exactly the intended test and nothing else.

TeunP:
### Follow-up: the review's claims are now measured, not reasoned

Re-checked my own review before signing off, because two of its statements were derived from reading code rather than from running it. Both are now pinned by tests in `773fa1c9` (`tests/zero_eta_method_guards.rs`, Tier 2, 0.24 s):

- **"the check-time warning now reaches `fit()`"** and **"a `[gn, focei]` chain no longer claims the result may be badly wrong"** were unit-tested at the level of the *decision functions* (`check_model_options`, `keep_gn_zero_eta_warning`). Nothing exercised the wiring in between — which is exactly where both can be lost, since `first_error` drops warnings and `run_foce_gn` cannot see the chain. Now driven through `fit()` end to end.
- **"both messages come out byte-identical"** was asserted from the string concatenation, and no test pinned it: the existing SAEM test only checks `contains("at least one random effect")`, so folding the two guards into one table could have handed `saem` the generic rationale with nobody noticing. Both messages are now asserted in full.

Measured output for a standalone `gn` at `n_eta = 0` (2-iteration budget):

```
* method = gn on a model with no random effects (n_eta = 0) is start-sensitive: … Prefer method = gn_hybrid or focei …
* Gauss-Newton: max iterations reached without convergence
* Gauss-Newton did not converge on a model with no random effects (n_eta = 0). … the result may be badly wrong … Use method = gn_hybrid or focei …
```

and for `methods = [gn, focei]`: only the middle line survives, which is the intended outcome — the stage really did stop early, but nothing claims the *result* is wrong.

**One thing worth your call.** That first and third line both end in "use `gn_hybrid` or `focei`", so a standalone `gn` run at `n_eta = 0` now says the same thing twice. That is a consequence of my fix for the dropped-warning gap, not of your original PR. I left both: they have different triggers (one is configuration advice available before the fit, the other is an outcome), and over-warning on a configuration the docs already call "not recommended" is the safe side. Trimming the post-fit message to just the convergence claim, and letting the check-time one carry the remedy, would also be defensible — say the word and I'll do it.

Also checked, since neither was covered by the PR's cross-repo section: no ferx-r code or test matches on these codes or messages, and the one `expect_length` on warnings there is for simulation warnings. `ferx-r`'s CI builds from its pinned `Cargo.lock` rev anyway, so nothing downstream moves until that is bumped.

Not fixed, unchanged from the review above: the `"max iterations reached"` mis-attribution on trust-radius-collapse exits (needs its own change — `types_tests.rs` pins that string to `Critical`/`convergence`), the `EstimationMethod::requires_random_effects()` refactor across the five run-time guards, and the multiple-diagnostics-for-one-problem case. I did not independently mutation-verify the six tests that came with the original PR — I took that claim at face value.

TeunP:
### Remedy trimmed to one place (`35b07ebc`)

Taking the option I flagged above: the two #1006 diagnostics are now split by role rather than both ending in the same advice.

- `W_GN_NO_RANDOM_EFFECTS` carries the **whole** remedy — including "or improve the sigma start", which the post-fit message previously owned alone, so nothing is lost in the trim.
- The post-fit warning carries only the **outcome**: this run did not converge, and in this regime that means the result may be badly wrong rather than merely imprecise.

A standalone `gn` run at `n_eta = 0` now reads:

```
* method = gn on a model with no random effects (n_eta = 0) is start-sensitive: … Prefer
  method = gn_hybrid or focei for a fixed-effects-only model, or improve the sigma start.
* Gauss-Newton: max iterations reached without convergence
* Gauss-Newton did not converge on a model with no random effects (n_eta = 0). BHHH
  curvature is unreliable far from the optimum in this regime, so the result may be badly
  wrong, not just imprecise.
```

Inside `fit()` the two always travel together, so the advice is never lost: the post-fit warning needs `n_eta = 0` and a non-hybrid, non-polished `gn` stage, which is precisely when the check raises `W_GN_NO_RANDOM_EFFECTS`. The one case that does lose it is a caller reaching `run_foce_gn` directly and bypassing `check_model_options` — noted on the const rather than papered over.

Pinned from **both** sides by `the_gn_remedy_is_stated_exactly_once`, so the duplication cannot creep back and the advice cannot quietly vanish either: restoring the remedy to the post-fit message (two mentions) and dropping it from the check (zero) each fail that test, and nothing else. `docs/estimation/gauss-newton.qmd` records the split.

Issue #864, opened 2026-07-16
Outer-optimizer conditioning rework (cold-start basin selection on ill-conditioned models)


roninsightrx:
Standalone effort split out from #847. Goal: let ferx reach the NONMEM basin from cold start on ill-conditioned models (fluconazole: OFV 734.6, Q 16.8) without multistart or NONMEM-quality inits — while regressing **no** existing fit. This is a robustness/optimization improvement, **not** a correctness fix: the objective is confirmed correct (ferx reaches 734.67 from NONMEM's final estimates) and the block_sigma gradient (#847) is validated.

## Problem statement (evidence-based)

On ill-conditioned models the outer optimizer **stalls at a non-stationary point**: the line search backtracks to ~0 while the packed gradient is still large. Fluconazole cold start: OFV 762 (Q collapses 11→4.7), frozen with `grad:TVQ≈+5.9`, `grad:TVBMAX≈+137`. NONMEM navigates the identical surface to 734.6 / Q 16.8.

Confirmed by trace + controlled runs:

| Lever | Result | Conclusion |
|---|---|---|
| default (analytic grad) | 762, stalls | baseline |
| `gradient_method = fd` | 1043, frozen Q=10 | **not** the gradient — reconverged-FD is worse |
| `inner_tol = 1e-10` | 861, still stalls | **not** gradient noise — clean grad still stalls |
| multistart (12) | 738, ρ=0.933 | Q basins near-equal-OFV (738@Q4.85 vs 734.6@Q16.8, Δ3.4) |
| NONMEM inits | 734.67 | objective is correct |

So (i) the objective is right, (ii) the stall is the **optimizer** (gradient-method- and inner_tol-independent), (iii) Q is genuinely weakly identified (NONMEM condition number 1046).

## Root causes (four coupled defects — must be fixed together)

1. **Log-transform gradient trap.** Sign-constrained θ are packed as `log θ` (`parameterization.rs:theta_packs_log`), so `g_logθ = θ·∂OFV/∂θ → 0` as θ→0. Collapsed params can't climb back, and the gradient-norm convergence test is falsely satisfied. NONMEM uses linear `θ/θ_init` (constant Jacobian).
2. **Weak line search.** The hand-rolled `optimize_bfgs` loop is Armijo-only; it accepts steps where the dominant BMAX/KD descent drags coupled Q into the wrong basin, and gives up at non-stationary points. A strong-Wolfe (curvature-condition) search is required.
[...]

Issue #469, opened 2026-06-23
estimation(FOCEI): nonlinear-frailty ω² reads ~17% above NONMEM/nlmixr2 on identical data — inner-Hessian accuracy lead (spin-off of #440)


TeunP:
### Summary

On the **same** TTE Weibull frailty dataset (`tests/reference/tte_weibull/`, 100 subjects, η on log-shape), ferx FOCEI estimates the frailty variance **ω² = 0.204**, while two independently-implemented Laplace engines agree well below it: **NONMEM LAPLACIAN 0.175 ≈ nlmixr2 FOCEI 0.173**. Everything else matches across all three to ≤0.5% / identical −2LL. Because all three are the same method (FOCEI/Laplace) on identical data, the ~17% gap is an **implementation** difference, *not* the shared approximation bias tracked in #440 — and it's a concrete, reproducible target for tightening the optimizer.

### Evidence (3-tool, identical files)

| | ferx FOCEI | nlmixr2 FOCEI | NONMEM LAPLACIAN |
|---|---|---|---|
| **Weibull** scale | 21.87 | 21.87 | 21.90 |
| shape | 2.214 | 2.20 | 2.210 |
| **ω² (log-shape)** | **0.204** | **0.173** | **0.175** |
| −2LL | 639.99 | 639.99 | 639.98 |
| **Exponential** ω² (log-rate, *linear* frailty) | 0.290 | 0.293 | 0.286 |

Two diagnostics localize it:

- **It's specific to the *nonlinear* frailty.** The exponential (linear frailty on log-rate) ω² agrees across all three to ~2% (0.286–0.293); the discrepancy only appears for the Weibull *shape* (a nonlinear hazard parameter).
- **The likelihood is flat in ω² over that range** (−2LL differs by ~0.01 from 0.175→0.204), so the data can't itself prefer one value. Practical impact on this dataset is negligible — the value here is as a **controlled probe of inner-Hessian accuracy**, not a user-facing wrong number.

### Why this is separate from #440

[...]

Issue #274, opened 2026-06-12
FOCEI covariance (non-IOV) omits the log|H̃| EBE-response curvature Δ; inconsistent with the IOV stencil


hiddevandebeek:
## Summary

The non-IOV FOCEI covariance step (`compute_covariance`, central FD of the analytic population gradient) omits one curvature term that the IOV stencil (scalar second-difference of the reconverged OFV) captures: the `log|H̃|` **EBE-response cross-curvature**

```
Δ = d/dθ [ ½ · ∂log|H̃|/∂η · dη̂/dθ ].
```

`Δ` is non-zero whenever the residual variance depends on η (proportional/combined error) and grows with the proportional error magnitude. The result is a small but systematic, σ_prop-growing discrepancy between the non-IOV and IOV stencils on FOCEI-INTER models, and a corresponding drift of the non-IOV SEs away from the true marginal Hessian `∇²(−2logL)`.

This is **not** a missing `log|H̃|` — the determinant is present in both the objective and the analytic gradient. It is specifically the EBE-response term dropped by the envelope-theorem shortcut.

## Where the term is and is not handled

The FOCEI per-subject marginal is the Almquist (2015) Laplace form
`data_ll(η̂) + η̂'Ω⁻¹η̂ + log|Ω| + log|H̃|`, with `H̃ = a'diag(1/R)a + ½·c̃'c̃ + Ω⁻¹`
(`a = ∂f/∂η`, `c̃ = (∂R/∂η)/R`). See `stats::likelihood::foce_subject_nll_interaction`
(the `log_det_htilde` term).

- **Objective (`ofv` closure / `pop_nll`)**: includes `log|H̃|`. ✓
- **Analytic gradient (`gauss_newton::subject_nll_pop_grad_analytical_laplace`)**: differentiates `log|H̃|` w.r.t. θ/σ/Ω analytically, through the `R`/σ/Ω channels (the `q_j = a_j'H̃⁻¹a_j` reservoir, the `β_j`/`γ_j` coefficients, and `Ω⁻¹H̃⁻¹Ω⁻¹`). ✓ It holds `a` and `η̂` fixed within a single evaluation, then reconvergence recomputes them at each FD node.
- **Dropped term**: the gradient invokes the envelope theorem, which zeros only the *inner* objective `∂/∂η[data_ll + η'Ω⁻¹η] = 0` — **not** `log|H̃|`. The true total gradient therefore carries an extra term:

```
[...]

Issue #997, opened 2026-08-19
Defaults and reporting: stiff-solver default changes the answer, convergence unverifiable under BOBYQA, optimizer/gradient coupling


TeunP:
## Summary

Ran ferx 0.3.0 alongside NONMEM 7.6 on the same stiff fast-binding model and
dataset, fitting each with matched settings. ferx's diagnostics found things
NONMEM does not report at all — but three **defaults** silently produced a
worse or unverifiable answer, and one diagnostic is unavailable exactly when it
is most needed.

Filing as one issue because the four are related: each is a case where the fit
*looks* fine and the output does not say otherwise.

Everything below reproduces on ferx's own bundled example or on a synthetic
model + dataset included in this issue. No external data is required.

---

## 1. `converged = TRUE` is unverifiable under a derivative-free optimizer

`final_gradient` is populated for L-BFGS and **NULL for BOBYQA** — that is,
absent precisely for the optimizer where a stall is most likely and there is no
gradient to check against.

Reproduces on the bundled `mm_oral` example, no extra data:

```r
library(ferx)
x <- ferx_example("mm_oral")
file.copy(x$model, "base.ferx"); file.copy(x$data, "d.csv")
L <- readLines("base.ferx"); i <- grep("^\\[fit_options\\]", L)
writeLines(append(L, c("  optimizer = lbfgs",  "  gradient = auto"), i), "lbfgs.ferx")
writeLines(append(L, c("  optimizer = bobyqa", "  gradient = auto"), i), "bobyqa.ferx")
for (f in c("lbfgs.ferx", "bobyqa.ferx")) {
  fit <- ferx_fit(f, data = "d.csv")
  cat(f, "converged =", fit$converged, " OFV =", round(fit$ofv, 4),
      " final_gradient =", if (length(fit$final_gradient)) "present" else "NULL", "\n")
}
```

Observed:

```
lbfgs.ferx   converged = TRUE  OFV = -453.3985  final_gradient = present
bobyqa.ferx  converged = TRUE  OFV = -452.4083  final_gradient = NULL
```

Both report success; BOBYQA is 0.99 OFV worse and offers nothing to check the
claim with. On a larger problem the gap is much bigger — see §2.

**Suggestion:** compute a finite-difference gradient at the solution for
*reporting* when the optimizer does not supply one, or state in the fit object
that convergence is unverified.

## 2. A `never_moved` check is free and would have caught a real stall

[...]

PR #1034, opened 2026-08-21
fix(estimation): stop reporting trust_region as converged on MaxIters [closes #1000]

### LINKED ISSUE #1000: bug(trust_region): reports `Converged: YES` when the optimizer terminated on `MaxIters`

TeunP:
## Why

`optimizer = trust_region` reported `Converged: YES` whenever argmin's executor returned `Ok(_)` — including when it terminated because it hit `outer_maxiter`. A fit that simply ran out of iterations was reported to the user as converged, with standard errors computed at a non-stationary point.

On `examples/one_cpt_iv.ferx` with its omegas deleted (the #989 fixed-effects shape, which makes this trivially easy to hit) against `data/one_cpt_iv.csv`:

| `maxiter` | reported | OFV |
|---|---|---|
| 5 | `Converged: YES` | 11325.8283 |
| 300 | `Converged: YES` | 8504.4349 |
| 2000 | `Converged: YES` | −269.6370 |

The true optimum is −269.6370 (NONMEM `$OMEGA 0 FIX` anchors it at −269.63700440). The first two rows are 8 000–11 000 OFV units off and were both reported as converged; the `%RSE` of 6121% / 9632% was the only hint, and that is the kind of thing a user attributes to their model rather than to the optimizer.

Not specific to zero-Omega models — it applies to any `trust_region` fit that exhausts `outer_maxiter`.

## What changed

**The suggested fix in the issue — inspect the termination status instead of the `Result` discriminant — is necessary but not sufficient, and on its own would make things worse.** argmin's `TrustRegion::terminate` returns `NotTerminated` unconditionally (`argmin-0.11.0/src/solver/trustregion/trustregion_method.rs:297`), and the default `terminate_internal` only adds `MaxItersReached` and `TargetCostReached`. We never set a target cost, so **`MaxItersReached` is the only reachable termination reason**. Mapping it to "not converged" without giving the solver a way to stop on its own would report `Converged: NO` for every trust-region fit, including the correct one.

So this PR adds the missing stopping rule first, then maps the status honestly:

1. **`TrustRegionWithStop`** wraps the solver, delegating `init`/`next_iter` and adding the `terminate` argmin lacks. It stops once the run has gone 20 consecutive iterations with no measurable progress — no objective improvement beyond `outer_ftol` (default `1e-6`, relative) *and* no parameter movement beyond `outer_xtol` (default `1e-4`, relative). An `ever_progressed` guard keeps a trust region that rejects from its very first step from reading as converged. The executor checks `terminate` *before* the `MaxIters` test, so a run that settles on its last permitted iteration is still reported converged.
2. **`classify_termination`** maps status → `(converged, warning)`. Only solver-side stops count: `SolverConverged` and `TargetCostReached`. `MaxItersReached` reports `Converged: NO` with the warning the issue asks for, naming the budget and the gradient norm the fit stopped at. Interrupt / timeout / solver bail-out likewise report not-converged with a reason.
3. Two smaller fixes in the same path: the trust-region optimizer now populates **`final_gradient`** (it always returned `None`, unlike the FOCE and GN paths), and a settled run whose final OFV is not finite is not reported as converged (`ofv_fixed` clamps a blown-up objective to a `1e20` sentinel, which the no-progress rule could otherwise settle on).

Result on the issue's own reproduction:

| `maxiter` | OFV | before | after |
|---|---|---|---|
| 5 | 11325.8283 | `Converged: YES` | **`Converged: NO`** + warning |
| 300 | 8504.4349 | `Converged: YES` | **`Converged: NO`** + warning |
| 2000 | −269.6370 | `Converged: YES` (ran all 2000) | `Converged: YES` at **iteration 1920** |

Stopping when it settles also saves work: the warfarin fit returns at iteration 59 rather than grinding out the remaining budget.

## Alternatives considered

**A gradient-stationarity stop — tried first, and rejected on measurement.** It is the natural criterion (this optimizer already computes the exact analytic outer gradient, so testing it is free) and it worked beautifully on analytic models. It fails on ODE models, because the outer gradient is evaluated at **fixed** EBEs:

| model | point | ‖∂OFV/∂x‖ |
|---|---|---|
| `one_cpt_iv_pooled` (n_eta = 0, analytic) | optimum, OFV −269.6370 | 3.3e−5 |
| `warfarin` (n_eta = 2, analytic) | optimum, OFV −280.3640 | 4.2e−2 |
| `one_cpt_iv_ode` | **optimum**, OFV −387.0737 | **43.2** |
| `one_cpt_iv_ode` | **worse** point, OFV −386.3451 | **4.0** |

The last two rows are the killer: the gradient norm is **non-monotone in solution quality**, so any threshold between 4 and 43 calls the worse point converged and the better one not. (Each "optimum" above is the point the default optimizer converges to and reports `YES` for, measured under identical options.) Objective and step size are the noise-robust measures, and they are what the NLopt outer optimizers already stop on — hence `outer_ftol` / `outer_xtol`.

**Reporting the status without a stopping rule** — see above: every fit would read `Converged: NO`.

**Deriving convergence post-hoc from the returned point** rather than in-loop: same criterion problem, and it would keep burning the full budget on every fit.

## Cross-repo dependency

No public API change. `FitResult.converged` / `.warnings` / `.final_gradient` are existing fields; the R wrapper surfaces them already.

## Breaking changes

- [x] None

Behaviourally, some `trust_region` fits that previously self-reported `Converged: YES` now report `NO`. That is the fix, not a regression — but anything downstream keying off `converged` for this optimizer will see the change.

## Numerical validation

- [x] Compared against: NONMEM (`$OMEGA 0 FIX` anchor) and the default outer optimizer on the same options

Estimates are unchanged where the fit already converged — the parameter path is untouched; only the stopping point and the reported verdict change. On the zero-Ω anchor, the converged run lands on the same OFV as before and on the NONMEM value:

```
# reference: NONMEM $OMEGA 0 FIX
OFV -269.63700440    TVCL 4.84082

# before (maxiter 2000, ran the full budget)
OFV -269.6370   TVCL 4.840820   Converged: YES

# after (stops at iteration 1920)
OFV -269.6370   TVCL 4.840706   Converged: YES
```

TVCL differs in the 5th decimal because the run now stops 80 iterations earlier, on a tail that was still creeping by ~2e−5 OFV per iteration; both agree with NONMEM to 4 significant figures.

Warfarin under `foce`: `trust_region` reaches OFV −280.3640, the same value `nlopt_lbfgs` converges to, and now reports it at iteration 59.

## Performance

- [x] Faster — `trust_region` fits that settle now stop when they settle instead of running to `outer_maxiter`. Warfarin (`foce`): 5000 iterations → **59**. `one_cpt_iv_pooled` (maxiter 2000): 2000 → **1920**. `one_cpt_iv_ode`: 3000 → **73**.

The added per-iteration cost is one comparison plus a step-norm over the packed vector. The one-off gradient evaluation at the returned point is per *fit*, not per iteration, and is a cache hit whenever the last evaluated point is the one returned.

## Tests

Tier-1, in `src/estimation/trust_region.rs`:

- `classify_termination_only_calls_solver_side_stops_converged` — the mapping itself, which is the whole defect: `MaxItersReached` must not read as convergence and must name the budget.
- `made_progress_reads_objective_and_step` — both limbs of the progress predicate, including a rejected step (neither moves), a cost increase, and a NaN objective.
- `note_iteration_*` (three tests) — the stall tracker: fires on exactly the right iteration, never fires without prior progress, and resets its window on progress.
- `trust_region_maxiter_exhaustion_is_not_converged` — an actual fit with `outer_maxiter = 2`, asserting `!converged`, the warning text, and that `final_gradient` is now populated.

Tier-3 (slow-gated), in `tests/new_optimizers.rs`: `trust_region_reports_converged_when_it_settles` — the converged direction end to end, checked against the OFV the `auto` optimizer reaches under the same options rather than a hardcoded constant.

All mutation-verified: restoring the `MaxIters ⇒ converged` mapping, disabling the stopping rule, making it fire immediately, and dropping the step limb of the progress test each fail the expected tests and nothing else.

**Audit** — the issue asks to check the other argmin-driven optimizers for the same `Ok(_) => true` shape. There are none: `Executor::new` appears exactly once in the tree, here. The Nelder-Mead inner fallback is our own implementation and returns its own convergence flag.

## Example

- [x] Not applicable (no new API surface; behaviour of an existing `optimizer` value)

## Docs

`docs/estimation/optimizers.qmd` gains a **Trust-region convergence reporting** section: the stopping rule and its tolerances, the before/after table, why it is not a gradient test (with the ODE numbers), and an explicit note on what "converged" does and does not mean here.

## Reviewer hints

Focus on `src/estimation/trust_region.rs`. The two things worth real scrutiny:

1. **`TRUST_REGION_NO_PROGRESS_ITERS = 20`.** A trust region shrinks its radius after each rejected step, so a short run of rejections is ordinary mid-descent behaviour and must not read as convergence; the count has to sit above that transient. Measured settling points are 59 (warfarin), 73 (ODE) and 1920 (pooled) iterations, so 20 is not close to either edge — but it is a constant chosen from four models, and a second opinion on it is welcome.
2. **What `converged` now means.** It means *this optimizer stopped making progress* — the same thing an NLopt `xtol`/`ftol` stop means — not that the point is a verified optimum. A trust-region fit can settle above what another optimizer reaches, and the two cases are locally indistinguishable: on warfarin under `foce` the settled point *is* the optimum `nlopt_lbfgs` finds, while under `focei` the same fit settles 4.1 OFV units above it (−281.88 vs −286.00). Both look identical from inside the solver — frozen objective, every step rejected — so no local criterion separates them. That shortfall is #864, which #1000 explicitly scopes out; it is documented in `optimizers.qmd` and recorded in the Tier-3 test's rationale, and I will add the measurement to #864 separately.

`classify_termination` and the stall tracker are pure functions, tested directly; the wrapper around argmin is a thin delegation.

## Open questions

Whether `converged` should be *further* qualified for the stall case — e.g. reporting converged but attaching an advisory when every step in the settling window was rejected. I left it out because the two stall shapes above are indistinguishable, so the advisory would fire on the correct warfarin `foce` fit too. Happy to add it if you would rather have the false positive than the silence.

---

## Update after review (`2aaaf3dd` … `b6b1c8da`)

Everything above still describes the change, with three amendments from the xhigh review. Nothing above has been edited; read this section as the correction.

### 1. The FOCEI shortfall was a truncated gradient, not #864

`FoceiProblem::gradient` summed the fixed-η̂ score `2·Σ gᵢ` and dropped the FOCEI `log|H̃|` EBE-response term `tᵢ` (#274/#289) that `gauss_newton::build_gn_system` already adds — `grad(OFV) = 2·Σ (gᵢ + tᵢ)`. That comment names the consequence in advance: *"stalling above it (warfarin FOCEI -276.6 → -286.0)"*. It is now fixed, and the reviewer-hints observation about `focei` settling 4.1 OFV units short goes with it:

| model | method | optimizer | OFV before | OFV now | iterations |
|---|---|---|---|---|---|
| `warfarin` | `focei` | `trust_region` | **−281.88** | **−286.0042** | 113 |
| `warfarin` | `focei` | `nlopt_lbfgs` | −286.0042 | −286.0042 | 54 |
| `warfarin` | `foce` | `trust_region` | −280.3640 | −280.3640 | 59 |
| `one_cpt_iv_pooled` | `focei` | `trust_region` | −269.6370 | −269.6370 | 1920 |

`trust_region` under `focei` now reaches the same optimum as `nlopt_lbfgs` — agreeing to 4 decimals on the OFV and 4–5 significant figures on every theta (TVCL 0.132691 vs 0.132695, TVV 7.737391 vs 7.737705, TVKA 0.810779 vs 0.810797). FOCE and the zero-Ω NONMEM anchor are bit-stable, which is what the theory predicts: `tᵢ` is identically zero without interaction and without random effects. The anchor still lands on −269.6370 / TVCL 4.840706 at iteration 1920 against NONMEM's −269.63700440.

Pinned by a Tier-1 parity test (`gradient_matches_central_fd_of_the_marginal_cost`) against central FD of the EBE-resolved objective, both interaction modes. Mutation-verified — reverting gives analytic 131.61 vs FD 90.98 on the TVCL coordinate, a 45 % error.

### 2. "Alternatives considered" — the gradient-stationarity numbers were taken on that gradient

`one_cpt_iv_ode.ferx` is `method = focei` with `proportional` error, so `tᵢ ≠ 0` there: the ‖g‖ = 43.2-at-the-optimum / 4.0-at-a-worse-point non-monotonicity in the table above was measured on the incomplete gradient. The decision to stop on objective + step size rather than gradient norm still stands — that is what `bobyqa` does and it is robust to EBE noise either way — but the ODE evidence for it should be re-measured before being cited again.

### 3. Tolerances, and two verdicts that were structurally wrong

- `outer_ftol` is resolved through `resolve_outer_ftol` rather than a hardcoded `1e-6`, so a pure-TTE fit gets the `1e-8` #469 tightening instead of stopping 100× looser than every other optimizer.
- `outer_ftol` / `outer_xtol` were documented as `bobyqa`-only; `docs/model-file/fit-options.qmd` now says they govern `trust_region` too.
- The stopping rule needs 21 outer iterations, so `maxiter ≤ 20` could never report converged — including the `maxiter = 5` row in the table above, which is structural rather than a verdict on the fit. `TRUST_REGION_NO_PROGRESS_ITERS` is unchanged at 20; the warning now says so.
- A run that rejects every step from iteration 0 (`rho = 0/0 = NaN`) now exits immediately against its *starting values* instead of spending the whole budget frozen and then advising "raise maxiter".
- `final_ofv.is_finite()` could not reject the ~1e20 divergence sentinel it was written for — the sentinel is finite. `DIVERGENCE_OFV` / `ofv_is_valid` now live in `estimation::outer_optimizer`, shared with `multistart_prefers`.
- `final_gradient` is taken after `clamp_to_bounds`, and `n_iterations` reports the outer iterations actually run (it was hardcoded `0`, printed directly beside the budget warning).

### 4. New rejection: `trust_region` × a quadrature stage (`E_OPTIMIZER_AGQ`)

`method = laplace` (any `n_agq`) and `method = focei, n_agq > 1` score the quadrature marginal through `pop_nll_opts`, but only `optimize_nlopt` and `optimize_bfgs` route the matching gradient. The trust region has no `agq_nodes()` branch, so it descended on the FOCE/Laplace closed form while reporting quadrature OFVs — converging silently to the FOCE optimum, exactly as `FitOptions::agq_nodes` warns. Rejected at check time; full support is [#1047](https://github.com/FeRx-NLME/ferx-core/issues/1047).

### 5. Multi-start behaviour change

`multistart_prefers` prefers a converged candidate over a non-converged one regardless of OFV. That key was inert while every `trust_region` start claimed convergence; it is now live, so a settled start can be selected over one that reached a lower OFV before running out of budget. Documented in `optimizers.qmd` and the CHANGELOG; the ranking itself is unchanged.

### Verification

`cargo test --lib --no-default-features --features ci`: 3131 passed, 0 failed. `cargo check --tests` green under `ci`, `ci,survival,slow-tests`, and `ci,markov`. `cargo clippy --no-default-features --features ci,markov` clean. Every new test mutation-verified.

**Not covered by a test:** the clamp-before-gradient ordering fix — no observable hook short of a contrived model that walks a theta out of bounds.


TeunP:
## Review — xhigh

The core diagnosis is right and the "status alone is necessary but not sufficient" framing is the correct call: mapping `MaxItersReached` without adding a stopping rule would have reported `Converged: NO` for every run. `classify_termination` and the stall tracker being pure and directly tested is the right shape.

I verified the argmin claims against `argmin-0.11.0` rather than taking them on trust: `TrustRegion::terminate` does return `NotTerminated` unconditionally (`trustregion_method.rs:295`), `Solver::terminate_internal` does call `terminate` *before* the `MaxIters` test (`core/solver.rs:96-107`), and on a rejected step `next_iter` returns `state.param(param).cost(self.fxk)` — identical param and cost — so the frozen-state stall detection is sound. Executor calls `terminate_internal` exactly once per loop with `update()` already applied.

14 findings below. **1, 2 and 4 are the ones I would not merge without.**

---

### 1. `final_gradient` is the FOCEI-incomplete gradient — and it is what the "gradient stationarity doesn't work" measurement was taken on

`src/estimation/trust_region.rs:588` reports `prob.gradient(&vec)`, i.e. `FoceiProblem::gradient` = `2·Σ gᵢ`. The Gauss-Newton path computes the same quantity as **`2·Σ (gᵢ + tᵢ)`** (`gauss_newton.rs:1870-1882`), where `tᵢ` is the `log|H̃|` EBE-response term. Its comment is explicit:

> For FOCEI (interaction) the fixed-η̂ analytic gradient drops the `log|H̃|` EBE-response term `tᵢ` (the #274/#289 Δ, envelope theorem zeros the inner objective but not `log|H̃|`). Adding it back gives the gradient optimizer the full marginal gradient, so it reaches the true minimum **instead of stalling above it (warfarin FOCEI -276.6 → -286.0)**.

`subject_nll_pop_grad` (which `FoceiProblem::gradient` calls) returns `gᵢ` only; `subject_nll_pop_grad_with_cache` returns the cache carrying `gn_theta_correction`. Both are `pub(crate)`.

Three consequences:

- **The new user-visible number is wrong under `focei`.** `final_gradient` and the `Final ‖∂OFV/∂x‖ = …` appended to the warning are non-zero at the true optimum for any interaction fit. That is a new surface this PR introduces, and it is the number a user is being pointed at to judge stationarity.
- **The reviewer-hints stall is this, not #864.** You measured warfarin `focei` settling at −281.88 vs −286.00. That is the same signature the GN comment records for the missing `tᵢ` (−276.6 → −286.0), on the same model, with the same estimator. Attributing it to "two stall shapes are locally indistinguishable" is one layer too shallow — the trust region is descending on an incomplete gradient, so of course it stops above the minimum.
- **It undermines the "Alternatives considered" table.** `one_cpt_iv_ode.ferx` is `method = focei` with `proportional` error, so `tᵢ ≠ 0` there. The ‖g‖ = 43.2-at-the-optimum / 4.0-at-a-worse-point non-monotonicity is measured on a gradient that is *known to be incomplete for that model*. The rejection of a gradient-stationarity stop may still be right, but as it stands it is not established.

I am not asking you to change the stopping rule — no-progress is defensible on its own merits. I am asking that the *gradient the optimizer descends on and now reports* be the complete one, which is a one-line switch to `subject_nll_pop_grad_with_cache`.

### 2. `TRUST_REGION_DEFAULT_FTOL = 1e-6` bypasses `resolve_outer_ftol` — re-opens #469 for `trust_region`

`trust_region.rs:330-332`:

> Relative objective tolerance for the no-progress test, used when `[fit_options] outer_ftol` is unset. **Matches the NLopt outer default (see `resolve_outer_ftol`).**

It does not. `resolve_outer_ftol(is_non_gaussian, is_ode, override)` (`outer_optimizer.rs:1213`) returns **`1e-8`** for a pure non-Gaussian (TTE / Phase-4 categorical) non-ODE model and `1e-6` otherwise. Its own docstring is the #469 rationale: at `1e-6` a Weibull shape-frailty reads ω² 0.204 against the NONMEM/nlmixr2 consensus 0.175; `1e-8` lands 0.176.

So a TTE fit under `optimizer = trust_region` stops 100× looser than the same fit under every other optimizer, and now reports `Converged: YES` while doing it. Call `resolve_outer_ftol(model.has_non_gaussian(), model.is_ode_based(), options.outer_ftol)` — the function already exists and already encodes the `is_ode` floor you would otherwise have to rediscover.

### 3. `outer_maxiter ≤ 20` can never report converged

The rule needs one progress transition plus `TRUST_REGION_NO_PROGRESS_ITERS = 20` no-progress transitions. `terminate` is called at iters `0..=max_iters`, i.e. `max_iters` transitions (the first call only seeds `prev`). So `SolverConverged` is unreachable for `maxiter ≤ 20`, and reachable at `maxiter = 21` only if the very first transition is the only progress the run ever makes.

A screening fit at `maxiter = 10` that genuinely settles at iteration 3 reports `Converged: NO` + "raise maxiter" — the inverse of #1000, with a remedy that does nothing until the budget clears 21. Note the docs table's own `maxiter = 5` row is structurally `NO` regardless of the fit. Worth either a floor (`min(20, max_iters/2)`-style) or an explicit documented caveat; a test for a small-budget converged fit would have caught it.

### 4. Multi-start now prefers a converged-but-worse `trust_region` start

`multistart_prefers` (`api/fit.rs:168-176`) is `(!b_conv && c_conv) || (b_conv == c_conv && c_ofv < b_ofv)` — **a converged candidate beats a non-converged one regardless of OFV**, within the valid class. Before this PR every `trust_region` start reported `converged = true`, so that limb was inert and selection was purely by OFV. It is now live.

`n_starts = 4`: start 0 settles in a shallow minimum at OFV −200 (`converged`), start 2 is still descending toward −280 when it hits the budget (`not converged`) → multi-start now returns −200 where it used to return −280.

That is a change to *which fit is returned*, not just to a flag. The "Breaking changes" section says only that the reported flag changes. It needs a line, and ideally a test.

### 5. `final_ofv.is_finite()` does not catch the sentinel it says it catches

`trust_region.rs:636-640`:

> the trial objective is clamped to a large sentinel when it blows up, so the no-progress rule can otherwise "settle" on that sentinel.

`1e20` is finite. The guard only rejects a recomputed `inf`/`NaN`, and `final_ofv` is a *cold* recompute (`run_inner_loop_warm(…, None, …)`), which can land on large-but-finite garbage that passes. `multistart_prefers` already defines the cutoff for exactly this — `DIVERGENCE_OFV = 1e14` (`api/fit.rs:169`) — and its docstring describes the same 1e20 sentinel mechanism. Reuse it.

### 6. `final_gradient` is evaluated before `clamp_to_bounds(&mut best_x, …)`

`prob.gradient(&vec)` at line 588 runs on the raw `get_best_param()`; `best_x` is clamped at line 606, before `unpack_params`, the covariance step and `packed_estimate`. If the unconstrained trust region walks a theta past its bound, `final_gradient` and the quoted ‖g‖ belong to a different point than `params`, `ofv` and `covariance_matrix`. Clamp first, then take the gradient.

### 7. `n_iterations: 0` (line 670)

Still hardcoded, two lines below a `res.state().get_iter()` you now read for the verbose trace. `io/output.rs:90` prints `Iterations: {}` in the same block as the new warning, so a budget-exhausted fit prints `Iterations: 0` directly above `reached outer_maxiter (500)`. The iteration count is now the headline diagnostic for this fix ("settles at 59", "stops at 1920") and it is the one number the user cannot see.

### 8. `docs/model-file/fit-options.qmd` not updated (CLAUDE.md)

CLAUDE.md: *"Any user-visible feature (new fit option, new estimator, new file-format directive, behavioural change) must update the relevant page — typically … `docs/model-file/fit-options.qmd` for `[fit_options]` keys."*

`outer_xtol` (line 25) still reads *"Only affects `bobyqa` … gradient-based optimizers use their own fixed internal tolerances"*, and `outer_ftol` (line 26) still scopes itself to *"the derivative-free `bobyqa` outer optimizer"*. Both are now false — this PR makes them govern `trust_region`.

### 9. A run that never progresses burns the whole budget

`ever_progressed` correctly refuses to call a never-started run converged, but there is no matching early exit. A fit started at (or numerically on) the optimum gives `rho = reduction_ratio(fxk, fxkpk, mk0, mkpk)` = 0/0 = NaN, `NaN > eta` is false, every step is rejected, param and cost frozen from iteration 0 — so the run pays all 500 default iterations of inner-loop + AD-gradient + Hessian to arrive at `Converged: NO` and "raise maxiter", which is not the problem. `stalled_iters >= N && !ever_progressed` → stop with a distinct reason is one branch.

### 10. `classify_termination`'s docstring contradicts the design (line 480)

> `SolverConverged` (**our stationary-gradient criterion**)

It is a no-progress criterion; line 313 in the same file says so explicitly and the PR body rejects a gradient stop outright. This is exactly the confusion the long `TRUST_REGION_NO_PROGRESS_ITERS` rationale exists to prevent.

### 11. The warning names an option key that does not exist (line 491)

`"reached outer_maxiter ({max_iters}) … raise maxiter"` — `outer_maxiter` is the internal `FitOptions` field; the only key the parser accepts is `maxiter` (`model_parser.rs:6539`), which is also what the remedy clause correctly says. One sentence, two names, one of them unusable.

### 12. Tier-3 anchor is two-sided (`tests/new_optimizers.rs:468`)

`(result.ofv - reference.ofv).abs() < 0.05` fails if `trust_region` reaches a *better* OFV than `auto`. The stated intent — "must land at the same optimum, not somewhere short of it" — is one-sided. Fixing finding 1 is precisely the change that would trip this and read as a regression.

### 13. The post-fit gradient never hits the cache on the converged path

> free when the last evaluated point is the one returned

On `SolverConverged` the run stops *because* its last steps were rejected, so `CostFunction::cost` has already overwritten `grad_cache` with the rejected trial point — the lookup misses every time in the successful case. Bounded (per fit, not per iteration), but the comment states the opposite of what happens. argmin calls `problem.gradient(&new_param)` on every acceptance, so caching the last accepted point's gradient would make it genuinely free.

### 14. `l2_norm` shadows its argument (line 337)

`fn l2_norm(v: &[f64])` then `.map(|v| v * v)` — `v` rebinds from slice to element. No runtime effect; rename to `x`.

---

### On the two things you asked about

**`TRUST_REGION_NO_PROGRESS_ITERS = 20`.** The framing "20 stalled iterations" undersells what the rule actually tests. Because progress is an OR (`Δf > ftol` **or** `‖Δx‖ > xtol`) and a rejected step moves neither, 20 consecutive no-progress iterations means 20 consecutive *rejections*, i.e. the radius has shrunk by ~`0.25²⁰ ≈ 1e-12` from the last accepted step. That is a much stronger and more defensible criterion than the count alone suggests, and I would say so in the constant's docstring — it also answers "why not 5, why not 100" better than the four measured settling points do. The number is fine. See finding 3 for its one hard edge.

**What `converged` means.** Agreed with your reading and with leaving the advisory out — but the `focei` example you use to argue the two shapes are indistinguishable is finding 1, not #864. Once the gradient carries `tᵢ`, it is worth re-measuring whether they still look the same from inside the solver.

TeunP:
## Review follow-ups applied

All 14 findings addressed. Commit: `2aaaf3dd`. Full `--features ci` lib suite green (3131 passed, 0 failed); `cargo check --tests` green under `ci`, `ci,survival,slow-tests` and `ci,markov`; CI clippy command clean.

| # | Finding | Outcome |
|---|---------|---------|
| 1 | `final_gradient` / descent direction drops the FOCEI `tᵢ` | **fixed** — `2·Σ(gᵢ + tᵢ)` |
| 2 | `TRUST_REGION_DEFAULT_FTOL` bypasses `resolve_outer_ftol` | **fixed** |
| 3 | `maxiter ≤ 20` can never converge | **fixed** (message, not the constant — see below) |
| 4 | Multi-start prefers converged-but-worse | **documented** (CHANGELOG + `optimizers.qmd`) |
| 5 | `is_finite()` misses the 1e20 sentinel | **fixed** — shared `ofv_is_valid` |
| 6 | Gradient taken before `clamp_to_bounds` | **fixed** — clamp moved first |
| 7 | `n_iterations: 0` | **fixed** |
| 8 | `fit-options.qmd` says the tolerances are bobyqa-only | **fixed** |
| 9 | Never-progressing run burns the whole budget | **fixed** — dedicated early exit |
| 10 | `classify_termination` docstring says "stationary-gradient" | **fixed** |
| 11 | Warning names `outer_maxiter` | **fixed** — names `maxiter` |
| 12 | Two-sided Tier-3 OFV anchor | **fixed** — one-sided |
| 13 | Post-fit gradient cache comment is wrong | **fixed** — comment corrected |
| 14 | `l2_norm` shadows its argument | **fixed** |

---

### 1 — the FOCEI gradient, with a number

`FoceiProblem::gradient` now returns `2·Σ (gᵢ + tᵢ)`, reusing
`subject_nll_pop_grad_with_cache` + `LaplaceGradCache::gn_theta_correction` — the
same pair `build_gn_system` uses. `GradCache` carries `tᵢ` alongside `gᵢ` rather
than folded in, because the BHHH Hessian must stay on the raw score.

New Tier-1 test `gradient_matches_central_fd_of_the_marginal_cost` pins
`Gradient::gradient` against central FD of `CostFunction::cost` (which re-solves
the EBEs, so the FD *is* the marginal derivative), on warfarin, for both
`interaction = false` and `true`. Each probe builds a fresh `FoceiProblem` so
`cost` is a pure function of `x` with no warm-start carry-over.

Mutation-verified — reverting to `2·Σ gᵢ`:

```
interaction=true coord 0: analytic 131.60916522936705 vs central FD 90.9834560445688
```

That is a 45 % error on the `TVCL` coordinate of the vector the trust region
descends on. FOCE and additive-error fits are unchanged bit-for-bit (`tᵢ ≡ 0`),
so the Tier-3 `foce` test and the zero-Ω anchor are untouched; `focei` fits
should now reach the OFV `nlopt_lbfgs` does, which I think retires the −281.88 vs
−286.00 observation in the reviewer hints from #864.

### 3 — I did **not** change `TRUST_REGION_NO_PROGRESS_ITERS`

You measured 20 from four models and asked for a second opinion; my opinion is
that it is right, and quietly shrinking a convergence criterion in a review pass
would partly re-open #1000. What I changed instead is the *message*: when
`maxiter` is below 21 the warning now says settling cannot be demonstrated at
that budget, rather than telling the user to improve their starting values for a
run that was never allowed to settle. The constant's docstring now also states
what the count actually tests — 20 consecutive *rejections*, i.e. radius shrunk
by ≈`0.25²⁰`.

If you would rather have an adaptive window (`clamp(maxiter/4, 5, 20)`), say so
and I will swap it in — it is a two-line change and the tests are already
parameterised on the constant.

### 9 — the early exit

`stalled_without_ever_progressing()` terminates with a `SolverExit` whose reason
`classify_termination` maps to a warning about the *starting values*. A fit that
rejects every step from iteration 0 (`rho = 0/0 = NaN`, so `rho > eta` is always
false) used to pay all 500 default iterations of inner-loop + AD-gradient +
Hessian to arrive at "raise maxiter". This is close to the `never_moved` /
`stalled_at_init` check asked for in #997 §2, scoped to this optimizer.

### 5 — one cutoff, two callers

`DIVERGENCE_OFV` / `ofv_is_valid` moved out of `multistart_prefers`'s body into
`estimation::outer_optimizer` and is now used by both. The unit test asserts the
thing the old guard missed:

```rust
assert!(1e20_f64.is_finite());
assert!(!ofv_is_valid(1e20));
```

### 4 — not fixed, documented

`multistart_prefers`'s "prefer converged" key is deliberate policy, so I did not
touch the ranking. Both the CHANGELOG entry and `optimizers.qmd` now say that the
key was inert while every `trust_region` start claimed convergence and is now
live, with the advice to give `trust_region` multi-starts enough `maxiter` for the
good start to settle.

---

### Separate bug found while reviewing — not fixed here

`optimizer = trust_region` pairs an **AGQ objective** with a **FOCE gradient**:
`ofv_fixed` → `pop_nll_opts` → `agq::agq_population_nll` when `agq_nodes()` is
`Some`, while `gradient`/`hessian` have no `agq_nodes()` branch at all. The NLopt
path guards this (`outer_optimizer.rs:3129` routes to
`agq::agq_population_gradient`), and both `FitOptions::agq_nodes` and that call
site document the failure in terms:

> an outer loop minimising the AGQ objective while fed the analytic *FOCE*
> gradient would converge, silently, to the wrong parameters.

So `method = laplace` or `method = focei, n_agq > 1` with `optimizer =
trust_region` converges to the FOCE optimum while reporting AGQ OFVs — and after
this PR it will report `Converged: YES` for it. `gauss_newton.rs` has the
mirror-image gap (uses `pop_nll`, so it silently ignores a requested `laplace`).
Orthogonal to #1000; I have a draft issue ready to file if you want it.

### Not covered by a test

Finding 6 (clamp-before-gradient) is an ordering fix with no observable hook
short of a contrived model that walks a theta out of bounds — the fix is in, the
regression test is not.

TeunP:
## AGQ × `trust_region`: guarded here, full support tracked separately

Commit `4805646e`.

**Correction to my previous comment first.** I wrote that `gauss_newton.rs` has "the mirror-image gap (uses `pop_nll`, so it silently ignores a requested `laplace`)". That is wrong. `agq_nodes()` keys off `options.method`, and `gn` / `gn_hybrid` *are* methods — so `laplace` can never be requested alongside them, and GN's objective and gradient agree with each other. Only the trust region is affected, and only because it is an `optimizer` rather than a `method`, so it composes with `laplace` / `focei, n_agq > 1`.

I also checked `optimize_bfgs` before writing the guard: it calls `population_gradient` (`outer_optimizer.rs:2329`), which is the AGQ-aware path. So the exposure is exactly `optimizer = trust_region` × a quadrature stage, and nothing else.

### What landed

Rejection, not support — `E_OPTIMIZER_AGQ` in `check_model_options`, next to the existing `E_OPTIMIZER_IOV`:

```
optimizer = trust_region does not support the quadrature objective (method = laplace).
Its gradient is the FOCE/Laplace closed form, so it would descend on a different function
than the one being scored and converge silently to the FOCE optimum while reporting
quadrature OFVs. Use optimizer = bobyqa, slsqp, lbfgs, nlopt_lbfgs, mma, or bfgs, which
route the quadrature gradient.
```

It reuses the `laplace_stage` / `focei_quadrature` chain predicates the quadrature-cap block already computes, so a chain whose *later* stage is the quadrature one is caught too. Plain FOCEI (`n_agq = 1`) is untouched. Registered in `diagnostics.rs` and `docs/file-formats/check-report.qmd`, with a CHANGELOG entry.

Tier-2 test in `tests/agq.rs` covers: `laplace` at `n_agq` 1 and 3, `focei` at 3, plain `focei` staying allowed, all seven AGQ-aware optimizers staying allowed, and the chained-stage case. Mutation-verified — dropping either limb of the guard fails it.

### Why not fix it properly here

Routing `FoceiProblem::gradient` to `agq::agq_population_gradient` is the easy half. The hard half is the BHHH Hessian: `4·Σ gᵢgᵢᵀ` needs **per-subject** scores, and `agq_population_gradient` returns a population gradient only. So a real fix needs either a new per-subject quadrature-score API or a different Hessian for the quadrature case, plus an `analytic_gradient_available` scope gate and an FD fallback. That is a feature, not a review follow-up, and it does not belong on top of #1000.

Happy to open it as an issue assigned to @TeunP — say the word and I will file it with the above as the body.

Full `--features ci` lib suite still green: 3131 passed, 0 failed.

TeunP:
Filed as #1047, assigned @TeunP, and linked from the guard comment and CHANGELOG (`c3ec167d`).

The issue body includes a "worth deciding first" section: `trust_region` is not the `auto` pick for anything, quadrature cost is dominated by the `n_agq^d` grid rather than outer-iteration count, and the AGQ-aware optimizers already cover the use case — so "reject clearly" may be the right permanent answer and #1047 closes as documentation.

TeunP:
## Measured, not asserted

I said in an earlier comment that FOCEI fits "should now reach the OFV `nlopt_lbfgs` does". That was an inference from the FD parity test, not a measurement, so I ran the fits. It holds:

| model | method | optimizer | OFV | iterations |
|---|---|---|---|---|
| `warfarin` | `focei` | `trust_region` | **−286.0042** | 113 |
| `warfarin` | `focei` | `nlopt_lbfgs` | −286.0042 | 54 |
| `warfarin` | `foce` | `trust_region` | −280.3640 | 59 |
| `one_cpt_iv_pooled` | `focei` | `trust_region` | −269.6370 | 1920 |

The 4.1 OFV `focei` gap in the reviewer hints (−281.88 vs −286.00) is gone — the two optimizers now agree to 4 decimals on the OFV and 4–5 significant figures on every theta:

```
                trust_region      nlopt_lbfgs
TVCL              0.132691         0.132695
TVV               7.737391         7.737705
TVKA              0.810779         0.810797
```

FOCE is bit-stable at −280.3640 / iteration 59, and the zero-Ω anchor is bit-stable at −269.6370 / TVCL 4.840706 / iteration 1920 against NONMEM's −269.63700440 — both as the theory predicts, since `tᵢ` is identically zero without interaction and without random effects. So the fix is confined to exactly the case it was meant for.

`Iterations: 1920` now prints where the anchor used to report `Iterations: 0`.

## Also closed since the last comment

- **`Solver::terminate` had no test** (`b6b1c8da`). The tracker tests pinned `note_iteration` and `stalled_without_ever_progressing` in isolation, but the branch turning a never-started run into the `TRUST_REGION_NO_INITIAL_PROGRESS` `SolverExit` was reachable only from a real fit. `terminate_maps_the_tracker_onto_argmin_statuses` now drives the trait method with synthetic `IterState`s and pins all three verdicts, feeding the frozen one through `classify_termination` to check the starting-values warning comes out.
- **The PR description is updated** with an "Update after review" section carrying the table above, the correction that the `one_cpt_iv_ode` gradient-norm numbers in "Alternatives considered" were taken on the incomplete gradient, and the `maxiter = 5` row being structurally `NO` rather than a verdict on that fit. Nothing you wrote was edited.

## State

`cargo test --lib --features ci`: **3132 passed, 0 failed**. `cargo check --tests` green under `ci`, `ci,survival,slow-tests`, `ci,markov`. `cargo clippy --features ci,markov` clean. Every new test mutation-verified.

One thing still uncovered, stated plainly: the clamp-before-gradient ordering fix has no regression test — there is no observable hook short of a contrived model that walks a theta out of its bounds.

Issue #1026, opened 2026-08-21
A branch-local parameter bound by pk(..., f=F) doses with the parameter while [scaling] reads a same-named data column — one name, two bindings, silently


TeunP:
Found while implementing #1004 (the analytical half of #993), checking whether the new check had a blind spot. It does, and the blind spot turns out to be a live bug on its own — wider than dose attributes.

## Summary

An individual parameter assigned **only inside an `if` with no `else`** is bound and applied correctly by `pk(..., f=F)`, but the *same name* read in `[scaling]` resolves to nothing and yields **`NaN` predictions**, with a clean parse and an empty `parse_warnings`.

## Measured

ferx at `worktree-worktree-1004-analytical-dose-attr` (behaviour is not new — nothing in #1004 touches this path). One subject, `WT = 80`, `predict` at fixed θ, observations at t = 1, 4, 12:

```ferx
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  KA = TVKA
  if (WT > 70.0) {
    F = TVF          # branch-local: no `else`
  }

[structural_model]
  pk one_cpt_oral(cl=CL, v=V, ka=KA, f=F)
```

| Variant | Predictions |
|---|---|
| **A** branch-local `F`, no `[scaling]` | `[0.7304006334509247, 0.7155442434203282, 0.3227080678738096]` |
| **B** unconditional `F` (`F = TVF` at top level), no `[scaling]` | `[0.7304006334509247, 0.7155442434203282, 0.3227080678738096]` |
| **C** no `F` declared at all | clean parse error (below) |
| **D** branch-local `F` **+** `obs_scale = 1.0 / F` | **`[NaN, NaN, NaN]`** |
| **E** unconditional `F` + `obs_scale = 1.0 / F` | rejected — `E_DOSE_ATTR_DOUBLE_USE` (#1004) |

**A and B are bit-identical**, so the branch-local parameter is bound to the dose route and applied exactly as an unconditional one would be. The binding works.

For **D**, the model parses clean and reports nothing:

```
indiv_param_names     = ["CL", "V", "KA"]     <- F absent
referenced_covariates = ["WT"]                <- F absent here too
parse_warnings        = []
```

Because the name is in neither list, the `E_MISSING_COVARIATE` guard from #765 cannot fire either — there is no name registered for it to find missing. Every prediction is `NaN`.

## Not specific to dose attributes

[...]

Issue #1024, opened 2026-08-21
nonmem_anchor/: 9 anchor families are indexed nowhere; the only README scopes itself to absorption models


TeunP:
Noticed while adding the analytical NONMEM anchor for #1004.

## What is there now

`nonmem_anchor/` holds **44** control streams. The only index is `nonmem_anchor/README.md`, whose title and opening line scope it to one feature family:

> `# NONMEM anchor — built-in absorption models (transit / inverse-Gaussian)`
> Reference NONMEM runs for the built-in absorption input-rate functions, to satisfy the CLAUDE.md "compare with NONMEM output" rule

Its table indexes 15 streams by name plus two glob rows (`dose_cmt_*.ctl` (13), `oral_central_inf_advan{2,4}_f{1,06}.ctl` (4)), so 32 of the 44 are reachable from it.

## The gap

The remaining **9 anchor families / 11 files are indexed nowhere** — not in `README.md`, not in `RUN.md`:

| Control stream(s) | What it anchors |
|---|---|
| `dose_attr_double_use_{A,B}.ctl` | #993 — dose attribute read on the prediction path (ODE) |
| `analytical_dose_attr_double_use_{A,B}.ctl` | #1004 — the same on the analytical engine |
| `correlated_residual_combined.ctl` | correlated combined residual error |
| `iov_anchor.ctl`, `iov_scaling_anchor.ctl` | IOV, and IOV × `[scaling]` |
| `transit_iov.ctl` | transit absorption × IOV |
| `flip_flop_transit.ctl` | flip-flop transit regime |
| `ig_indomain.ctl`, `ig_indomain_sim.ctl` | in-domain inverse-Gaussian |
| `ss_mm.ctl` | steady state × Michaelis-Menten |

These are not absorption input-rate anchors, so they are not merely *missing* from the table — they fall outside the scope the file declares for itself. Meanwhile the table has already outgrown that scope: it carries a **Dose compartment** row (`ADVAN1/2/3/4/11/12`) and an **Oral central infusion** row, neither of which is a built-in absorption input-rate function. So the file is being used as the general catalogue while still describing itself as the absorption one.

## Why it is worth fixing

The reason to commit these artifacts at all is so the next person can tell what already has an external anchor without re-running NONMEM. That only works if they can find them. Today, discovering that #993 has a committed NONMEM pair requires already knowing the test file name — the artifacts are effectively private to whoever added them.

[...]

PR #1035, opened 2026-08-21
fix(parser): reject a dose attribute mapped by pk(..., f=F) and also read on the analytical prediction path [closes #1004]

### LINKED ISSUE #1004: Analytical engine: a dose attribute mapped via pk(..., f=F) and also read in [scaling] is applied twice, silently — gates ferxtranslate#16
> **Priority: this gates ferxtranslate#16.** Raised from "follow-up, severity open" after
> the ferxtranslate maintainer measured their side (see comments).
>
> That consumer emits analytical models routinely and is **unexposed only by accident**:
> `.infer_pk_macro()` draws pk argument values from `CL, V, V1, V2, V3, Q, Q2, Q3, KA`,
> none of which can match the dose-attribute grammar, so it never emits `f=` / `alag=`
> at all. The combination this issue is about is *unreachable* from its output — not
> guarded against. Their `.deconflict_dose_attr_names()` renamer cannot help, because
> renaming is the wrong repair when the binding is legitimate: `F` really is the
> bioavailability, and the defect is the second read.
>
> **`translator#16` is the change that makes it constructible for the first time** — it
> begins emitting `f=` / `alag=` deliberately. On that day there is no engine check
> behind it on analytical models. So landing this is not a convenience that saves
> duplicated work downstream; it decides whether #16 ships with a backstop.
>
> Their stated plan: if this lands before #16 starts, they scope #16 against the real
> engine behaviour instead of writing their own check.

---

Follow-up from the review of #1003 (which closes #993).

#993 made it a hard error for an **ODE** model to declare a dose attribute the engine applies at the dose event and *also* read it on the prediction path (`[odes]` RHS, an `init(...)` seed, or the `[scaling]` readout). The analytical engine was deliberately left out of that PR's scope. The gap is real and, on the more commonly used engine, silent.

## The gap

```
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  F  = TVF

[structural_model]
  pk one_cpt_iv(cl=CL, v=V, f=F)

[scaling]
  obs_scale = 50.0 * F
```

`F` scales the dose amount at the dose event *and* divides the readout. No error, no warning. The equivalent ODE model is now rejected with `E_DOSE_ATTR_DOUBLE_USE`.

## Why the PR's rationale does not cover it

The scope note in #1003 argues that an analytical model binds bioavailability through an explicit `pk(..., f=F)` mapping, so a second use is "stated rather than silent". That holds for the reader who is looking at both blocks at once; it does not hold as a diagnostic. Nothing in the model says the value is applied twice, which is the same thing that was true of the ODE case before #993.

`parse_full_model`'s test `analytical_f_mapping_is_not_a_double_use` currently pins the accepting behaviour, using `obs_scale = V * F` as its example. That particular spelling happens to trip the pre-existing volume warning in `build_obs_scale_spec` ("a built-in analytical `pk` block already outputs concentration … this `obs_scale` divides by `V` again on top"), which masks how quiet the general case is. Replace `V` with a literal or any non-volume expression and there is no diagnostic at all.

## The open design question: error or warning?

Not a mechanical port of the ODE check. The codebase already has a precedent pointing the other way:

- `build_obs_scale_spec` treats the exactly-analogous **volume** double-divide on an analytical model as a **warning**, on the grounds that a deliberate additional transform is legitimate.
- #993 treats the bioavailability double-use on an ODE model as an **error**.

There is also a NONMEM-idiom argument on both sides. `S1 = V/F` is the standard apparent-clearance convention *when `F1` is not separately defined* — the `CL/F`, `V/F` parameterisation. Defining `F1` **and** `S1 = V/F1` is double-counting in NONMEM too, so an error is defensible; but the two spellings look similar enough that a warning may be the safer landing for the analytical engine.

## Scope / blast radius

Unmeasured, and that is the first task. #1003 shipped the ODE half only after confirming zero affected models across `examples/`, `src/`, `tests/`, and ferx-r `origin/main`. The analytical engine is the default for most shipped models, so the same sweep has to run before picking a severity.

## Suggested work

1. Sweep `examples/`, inline models in `src/`/`tests/`, and ferx-r `origin/main` for an analytical model that maps `f=`/`lagtime=` and also references that parameter in `[scaling]`.
2. Decide error vs. warning on the evidence from (1) plus the `build_obs_scale_spec` precedent.
3. Reuse `DoseAttrConsumption::of` (`src/types.rs`) — it already classifies by name, and the analytical path passes `slot = None`, which is exactly the case it documents.
4. Re-point `analytical_f_mapping_is_not_a_double_use` at whatever the decision is, and add a case whose readout does **not** mention the volume, so the volume warning cannot mask the result.
5. Extend the `docs/model-file/individual-parameters.qmd` callout and `docs/file-formats/check-report.qmd`'s `E_DOSE_ATTR_DOUBLE_USE` row, which currently scope the parse-time half to ODE models.

## Related

- #993 / #1003 — the ODE half.
- #995 — the single reserved-name list; `DoseAttrConsumption::of` is the intended home for this classification.

TeunP:
## Why

Closes #1004 — the analytical half of #993.

#993 (PR #1003) made it a parse error for an **ODE** model to declare a dose attribute the engine applies at the dose event and *also* read it on a prediction path. Analytical models were deliberately left out, on the argument that an explicit `pk(..., f=F)` mapping makes a second use "stated rather than silent".

That argument does not survive contact with the model file. Nothing in it says the value is applied twice — which is precisely the silence #993 closed on the other engine, and this is the **default** engine for most shipped models:

```ferx
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  F  = TVF

[structural_model]
  pk one_cpt_oral(cl=CL, v=V, ka=KA, f=F)

[scaling]
  obs_scale = 1.0 / F      # F applied at the dose AND here
```

Measured on `main` (`predict` at fixed θ, `F = 0.5`): every prediction is scaled by exactly `F` relative to the same model without the readout — `B/A = 0.500000` at all seven observation times, no error, no warning.

The masking effect the issue predicted is real too: the test that pinned the accepting behaviour used `obs_scale = V * F`, whose `V` trips the pre-existing volume double-divide **warning** in `build_obs_scale_spec`. Replace `V` with a literal and there is no diagnostic at all. The new tests deliberately avoid `V` in the readout.

**This gates ferxtranslate#16.** That consumer emits analytical models routinely and is unexposed only by accident: `.infer_pk_macro()` draws pk argument values from `CL, V, V1, V2, V3, Q, Q2, Q3, KA`, none of which can match the dose-attribute grammar, so it never emits `f=`/`alag=` at all. `translator#16` is the change that starts emitting them deliberately — and their `.deconflict_dose_attr_names()` renamer is the wrong tool for this shape, since the binding is legitimate (`F` really *is* the bioavailability; the defect is the second read).

### Scope sweep (the issue's task 1)

Ran before picking a severity, matching what #1003 did for the ODE half. Looked for an analytical model that maps `f=`/`lagtime=`/`alag=` and also references that parameter in `[scaling]`, `[initial_conditions]`, or `[adaptive_dosing] observe`:

| Sweep target | Affected models |
|---|---|
| `examples/*.ferx` | **0** |
| inline models in `src/**/*.rs` (per string literal) | **1** — `adaptive_observe_does_not_reach_the_absorption_twin`, a #993 regression fixture, updated here |
| inline models in `tests/*.rs` | **0** |
| ferx-r `origin/main` (57 files carrying a `pk ...(...)` line: `inst/examples/models/*.ferx`, `examples/models/*.ferx`, `R/*.R`, `inst/examples/*.R`, `README.md`, `NEWS.md`) | **0** |

### Error, not warning

The issue framed this as open, with `build_obs_scale_spec`'s volume warning as the precedent pointing the other way. Error, for three reasons:

1. **The volume precedent does not transfer.** That one warns because a second division by `V` *can* be a deliberate additional transform — a documented, supported feature (`examples/scaling_expression.ferx`). There is no reading under which applying bioavailability twice is intended.
2. **The sweep found nothing to break** (table above), so the strict landing costs no existing model.
3. **The NONMEM-idiom argument cuts the same way.** `S1 = V/F` is the standard apparent-volume convention *when `F1` is not separately defined*. Defining `F1` **and** dividing by it double-counts in NONMEM too — NONMEM just never says so (anchored below). A warning would leave the default engine quieter than the ODE engine about an identical defect.

## What changed

`check_dose_attr_double_use` now runs on **both** engines. The only difference is what it is handed as the slot map:

- ODE: `ode_slot_map` — the *name* is what routes a bare `F`/`LAGTIME` onto the dose (unchanged, #993).
- Analytical: new `analytical_dose_attr_slot_map`, built from the `pk(...)` role map — the *mapping* is what binds it. Resolution mirrors `build_pk_param_fn`'s (exact name, then the legacy lowercase compat lookup), so the check and the code that actually wires the value can't disagree. Non-mapped parameters get a `usize::MAX` sentinel, which falls through to name-based recognition exactly as unrelated ODE slots do — so an analytical `D{n}`/`R{n}` still classifies as `CodedRateOnly` and stays data-gated.

Four analytical prediction-path surfaces are covered, mirroring the ODE four:

| ODE (#993) | Analytical (this PR) |
|---|---|
| `[odes]` RHS | — (no such block) |
| `[odes] init(...)` seed | `[initial_conditions] init(...)` |
| `[scaling]` `y`/`obs_scale` | same |
| `[adaptive_dosing] observe` | same |

The `[initial_conditions]` surface is included for the same reason #993 had to walk `init(...)`: without it the rule has a hole the size of its own remedy — a user told to drop `F` from `[scaling]` can move it into `init(central) = F * 100`. That block also now records `D{n}`/`R{n}` reads for the data-gated half, matching `[scaling]`.

**The remediation clause is engine-specific**, because the repair is: on ODE models "rename it — `F` is a reserved dose-attribute name"; on analytical models "remove the `f=F` mapping from the `pk(...)` call — the mapping is what makes `F` a dose attribute". Renaming an analytical model's parameter fixes nothing, since the mapping follows it. `parse_error_to_diagnostic` therefore keys `E_DOSE_ATTR_DOUBLE_USE` off both clauses, each pinned by its own test in `tests/check_command.rs`.

Unchanged: a parameter merely *named* `F` that no `pk(...)` argument maps is an ordinary parameter on the analytical engine and stays silent (this is what ferxtranslate's nlmixr2 sources produce); `[derived]`/`[output]` reporting reads stay silent; the volume double-divide stays a warning.

## Alternatives considered

**Warning instead of error** — rejected, see "Error, not warning" above. **Reusing the ODE `ode_slot_map` path** — impossible: it is empty for analytical models, which is exactly why the gap existed.

## Cross-repo dependency

| ferx (R) | _not yet open_ | needed after merge | after |

No `pub` API change, so nothing forces a lock bump — but ferx-r's `NEWS.md` currently states, in a **Breaking changes** bullet, that "on an analytical `pk ...` model the same double use is still accepted silently — do not rely on the engine to catch it there", pointing at #1004. That sentence becomes wrong the moment this merges, so a follow-up ferx-r PR (lock bump to this commit + NEWS correction) is needed after.

## Breaking changes
- [ ] Public Rust API
- [ ] None

<details>

An analytical model that maps a dose attribute **and** reads it back now fails to parse:

```ferx
[structural_model]
  pk one_cpt_oral(cl=CL, v=V, ka=KA, f=F)
[scaling]
  obs_scale = V / F          # now E_DOSE_ATTR_DOUBLE_USE
```

Fix by dropping whichever half was not meant:

- `F` really is bioavailability → remove it from the readout (`obs_scale = V`, or drop `[scaling]` entirely — a `pk` block already outputs concentration).
- `F` is an ordinary parameter → **remove the `f=F` argument** from the `pk(...)` call. Renaming does not help here; the mapping follows the parameter.

Unaffected: the ordinary `CL/F`, `V/F` apparent-parameter convention, which does *not* map `f=`. A parameter named `F` that nothing maps is an ordinary parameter.

</details>

## Numerical validation

NONMEM 7.6.0, run locally. Two `ADVAN2 TRANS2` control streams — NONMEM's own **analytical** library routine, no `$DES` — differing in exactly one `$PK` line, `$THETA` all `FIX`, `MAXEVAL=0`, `$OMEGA 0 FIX` (`nonmem_anchor/analytical_dose_attr_double_use_{A,B}.ctl`, outputs under `nonmem_anchor/results/`):

- **A** — `F1` at the dose only: `S2 = V`
- **B** — `F1` at the dose *and* in the scale: `S2 = V/F1`

`S2` is NONMEM's readout divisor — the exact counterpart of ferx's `obs_scale`.

```
   t   NONMEM-A     NONMEM-B     B/A
 0.5   0.513070     0.256530     0.500000
 1.0   0.730400     0.365200     0.500000
 2.0   0.823870     0.411930     0.500000
 4.0   0.715540     0.357770     0.500000
 8.0   0.481420     0.240710     0.500000
12.0   0.322710     0.161350     0.500000
24.0   0.097198     0.048599     0.500000
```

`B/A = 0.5 = F1` at every point, and **NONMEM reports nothing**: B completes with `#OBJV = 646.654` against A's `656.358`, and the two `.lst` warning blocks are byte-identical (the same NM-TRAN population/ETA boilerplate). So ferx is deliberately stricter than NONMEM here, exactly as on the ODE half.

ferx's own analytical engine reproduces **NONMEM-A** to 6e-6 relative (`pk one_cpt_oral(cl=CL, v=V, ka=KA, f=F)`), and its accepted-vs-doubled ratio is `B/A = 0.500000` at every time — the same factor, measured on `main` before the fix. Run A also re-derives the #993 `NONMEM_A` table from a *different* NONMEM routine (`ADVAN2` closed form vs `ADVAN13` integrator), so the two anchors cross-check each other.

- [x] Full suite green: 3595 tests, 0 failures (`--lib --tests`, `--features ci`).
- [ ] Warfarin example converges (N/A — parse-time rejection, no estimation change)
- [x] Compared against: NONMEM 7.6.0
- [x] Reference values: above; committed in `nonmem_anchor/results/analytical_dose_attr_double_use_{A,B}.{tab,lst}`

## Performance
- [x] No significant impact expected — one extra pass over the individual-parameter list per parse.

## Tests
- [x] Unit tests added (`src/parser/model_parser_tests.rs`): `[scaling]` (three readouts, none mentioning `V`, so the volume warning cannot mask the result), the lag half in both `lagtime=`/`alag=` spellings, `[initial_conditions]`, `[adaptive_dosing] observe`, plus two scope-floor tests — an unmapped parameter named `F` stays ordinary, and `f=FBIO` makes `FBIO` the attribute while a separate `F` stays ordinary.
- [x] Diagnostic-code test for the analytical sentinel (`tests/check_command.rs`).
- [x] NONMEM anchor extended with the analytical pair (`tests/dose_attr_double_use_nonmem_anchor.rs`).
- [x] Regression test that fails without the fix — every test above passes only with it (mutation-checked individually).
- [x] `analytical_f_mapping_is_not_a_double_use` re-pointed at the new contract, as the issue asked.

## Example
- [x] Not applicable — this removes a silently-wrong model, it does not add API surface.

## Docs
- [x] `docs/model-file/individual-parameters.qmd` — callout extended to the analytical engine, with the mapping-not-name distinction and the NONMEM divergence.
- [x] `docs/model-file/scaling.qmd` — new callout: bioavailability/lag is an **error** here, unlike the volume double-divide **warning** directly above it (the precedent the issue raised).
- [x] `docs/model-file/ode-models.qmd`, `docs/model-file/adaptive-dosing.qmd` — scope notes updated (both said "ODE only").
- [x] `docs/file-formats/check-report.qmd` — `E_DOSE_ATTR_DOUBLE_USE` row rewritten for both engines.
- [x] `CHANGELOG.md` `[Unreleased]` entry added; the #993 entry's "as does an analytical model's explicit `pk(..., f=F)` mapping" clause corrected, since it now states the opposite of the shipped behaviour.

## Checklist
- [x] `cargo clippy` clean — CI's invocation (`--no-default-features --features ci,markov`) exits 0. Verified mechanically that **no clippy diagnostic lands on a line this PR adds**, by cross-referencing clippy's locations against the added-line ranges from `git diff -U0`. (Adding `--tests`, which CI does not, surfaces 4 pre-existing `approx_constant` errors in `src/sens/two_cpt_explicit.rs` — a file this PR does not touch, with the offending `3.14` PK-parameter fixtures byte-identical on `origin/main`.)
- [x] Sensitivity path untouched (parse-time only)
- [ ] `Cargo.toml` version bump — not done; the `[Unreleased]` section already carries the #993 break for the same release.

## Reviewer hints

Focus on `analytical_dose_attr_slot_map`: it must agree with `build_pk_param_fn`'s binding resolution or the check and the wiring disagree. Note the `usize::MAX` sentinel is load-bearing — it makes `DoseAttrConsumption::of` fall through to name-based recognition, which is what keeps analytical `D{n}`/`R{n}` on the data-gated path.

Also worth a look: the absorption ODE twin. It declines outright on `[odes]`/`[scaling]`/`[initial_conditions]`, and drops `[adaptive_dosing]`, so none of the new rejections can reach `get_or_build`'s `.expect()` (the #1008 hazard class). `adaptive_observe_does_not_reach_the_absorption_twin` was the one fixture the sweep found; it asserted the primary *accepts* a dose-attr-reading `observe`, which is now exactly what is rejected, so it is re-pointed — the block-drop it guards is now unreachable-by-construction and the test says so.

Skimmable: the doc edits.

## Follow-ups filed while working on this

- **#1026** — a branch-local individual parameter (assigned only inside an `if` with no `else`) *is* bound and applied by `pk(..., f=F)`, but reading that name in `[scaling]` yields `NaN` predictions with a clean parse and no warnings; it is registered in neither `indiv_param_names` nor `referenced_covariates`, so #765's `E_MISSING_COVARIATE` guard cannot fire either. Measured, and **not** dose-attribute-specific (an ordinary parameter behaves identically). It is also the one shape this PR's check cannot see, since the check iterates `indiv_param_names` — so that model and the one rejected here are the same defect with different outcomes.
- **#1024** — 9 NONMEM anchor families are indexed nowhere; `nonmem_anchor/README.md` scopes itself to absorption models while being used as the general catalogue.
- **#1005** — commented: this PR adds a *third* instance of the re-parse pattern that issue describes, so the fix should retire `collect_indiv_param_reads` rather than leave three callers.


TeunP:
## Review — xhigh

Reviewed the diff plus the enclosing functions, and measured every claim below on this branch (`0011be65`) rather than reading it off the code. The `[scaling]` and `[adaptive_dosing]` halves are right and the NONMEM anchor is solid (`analytical_dose_attr_double_use_A.tab` reproduces the #993 `NONMEM_A` table digit-for-digit from a different routine, which is a genuine cross-check). Two blocking issues, both measured.

---

### 1. The `[initial_conditions]` arm rejects correct models — the engine never applies `F` there

`src/pk/mod.rs:414`, on `analytical_init_concentration`:

> Reuses the generic single-dose closed forms with `T = f64` and **`F = 1`** (an initial condition is not an absorbed dose, so bioavailability does not apply).

`analytical_init_concentration_g` passes `F = 1` for every compartment, and no lag is applied to the impulse either. So on the analytical engine an init amount is **not** a dose-attribute-carrying quantity, and reading `F` in an init expression applies it exactly **once**:

```ferx
[structural_model]
  pk one_cpt_oral(cl=CL, v=V, ka=KA, f=F)
[initial_conditions]
  init(depot) = F * 500        # bioavailable residue of a pre-study 500 mg dose
```

Measured on this branch — rejected:

```
[initial_conditions]: `F` is this model's bioavailability — the engine already scales
each dose amount by it — but it is also read in [initial_conditions], so the value is
applied twice: once at the dose, once where you read it (#1004). …
```

The message is factually wrong for this block, and so is `docs/model-file/initial-conditions.qmd`'s new sentence.

The rationale in the code comment does not survive either:

> without it the rule has a hole the size of its own remedy — a user told to drop `F` from `[scaling]` could move it into `init(central) = F * 100`

That does not reproduce the `[scaling]` defect. `obs_scale` multiplies **every** prediction, including the F-scaled dose contribution, which is what makes it a doubling. An init is a single additive term the engine never touches with `F`. Moving `F` from `[scaling]` into `init(...)` changes the model to something else entirely; it does not launder the same double count.

Same root cause, second symptom: the site also calls `record_coded_rate_reads` (line 2915), so `init(central) = D1 * RATE0` marks `D1` as a prediction-path read and `check_modeled_dose_rates` errors as soon as the dataset carries a `RATE=-2` row — a *data-gated* failure for a model that is correct.

**Suggested fix:** drop the `[initial_conditions]` arm (both the check and the `record_coded_rate_reads` call), revert the doc sentence, and replace the rejection test with an acceptance test that pins the engine's `F = 1` init contract. Worth noting the ODE `[odes] init(state) = <expr>` half of #993 has the same flaw — verified, `init(central) = F * 100.0` is rejected there too — but that is pre-existing and can be a follow-up.

---

### 2. The absorption-twin regression guard is now vacuous — mutation-verified

`adaptive_observe_does_not_reach_the_absorption_twin`'s half-two swaps in `observe = central / V`. The original test's teeth were that the twin **is** an ODE model, so a re-emitted `observe = central / (V * F)` trips `check_dose_attr_double_use` and `get_or_build()`'s `.expect()` panics. With the dose attribute gone there is nothing for the twin's parse to object to.

Mutation check — deleted `"adaptive_dosing"` from the twin's skip list at `model_parser.rs:8481` and re-ran:

```
test parser::model_parser::tests::adaptive_observe_does_not_reach_the_absorption_twin ... ok
```

Green with the block-drop removed. The comment "a twin that builds at all is the proof the block was dropped" no longer holds.

**Suggested fix:** assert the drop directly. `AbsorptionOdeEquivalent` already owns the reconstructed source; a `pub(crate) fn source(&self) -> &str` lets the test assert `!source.contains("[adaptive_dosing]")`, which is non-vacuous and independent of what the controller happens to read.

Related: the twin builder's own comment (`model_parser.rs:8464`) still says `[adaptive_dosing]` is dropped because a re-emitted `observe` "hits `check_dose_attr_double_use` — which the analytical primary is deliberately out of scope for". This PR inverts that premise. The crash reason is now dead (the primary rejects first — verified on a `one_cpt_transit` model), leaving only the stale-spec reason.

---

### 3. Nondeterministic diagnostic when one parameter fills both roles

`analytical_dose_attr_slot_map` iterates `pk_param_map` (a `HashMap`) unsorted, and the message's role lookup (`find`) never filters on `attr`. With `pk one_cpt_oral(cl=CL, v=V, ka=KA, f=X, lagtime=X)` + `obs_scale = 2.0 * X`, eight parses of the identical source **in one process** alternated between:

```
`X` is this model's absorption lag …    remove the `f=X` mapping …
`X` is this model's bioavailability … remove the `lagtime=X` mapping …
```

Both are self-contradictory (noun from the last-written slot, role from the first `find` hit), and neither is reproducible. `build_pk_param_fn` sorts the same map 3 900 lines down for exactly this reason:

> Iterate in sorted key order so a model with several bad bindings always reports the same one (HashMap iteration order is otherwise arbitrary).

**Fix:** sort the entries in `analytical_dose_attr_slot_map`, and filter the role lookup by the slot that matches `attr`.

---

### 4. The "Unreachable" role fallback is reachable, and quotes text the file does not contain

The slot map binds via exact match **or** `value.to_lowercase()` (mirroring `build_pk_param_fn`'s compat lookup), but the role lookup only matches `value.as_str() == name`. So the lowercase-bound case falls into the `unwrap_or_else` branch commented "Unreachable".

Measured — `[individual_parameters] tlag = TVLAG` + `pk one_cpt_oral(..., alag=TLAG)` + `obs_scale = 2.0 * tlag`:

```
… remove the `lagtime=tlag` mapping from the `pk(...)` call …
```

The file says `alag=TLAG`. Both halves are wrong — the canonical `lagtime` instead of the user's `alag` (the exact drift the doc comment claims to prevent), and the declaration's casing instead of the mapping's. Quote `{role}={value}` straight off the map entry.

---

### Smaller

| | |
|---|---|
| `model_parser.rs:2740` | Comment above `scaling_reads` still says "only the rejection is ODE-scoped" — contradicted by the split 30 lines below. The sibling comments at 2534 and 2762 were rewritten; this one was missed. |
| `api/validation.rs:3002` | "the message already opens with `[odes]: ` / `[scaling]: `" — now also `[initial_conditions]: ` and `[adaptive_dosing]: `. |
| `model_parser.rs:2775` | The six-line engine-split is copy-pasted at three call sites, and `usize::MAX` does double duty as a slot sentinel *and* an unbounded `n_states`. `Option<usize>` + one wrapper collapses all three and names both intents. |
| `model_parser.rs:9078` | `value.to_lowercase()` is inside the `position` closure — one allocation per parameter scanned. Hoist it. |
| tests | Nothing pins that an analytical `[derived]`/`[output]` read stays silent, though the PR body and CHANGELOG both assert it. It does hold (verified), it is just unpinned; `dose_attr_read_only_for_reporting_is_accepted` is an `ode(...)` model. |
| `model_parser.rs:9068` | The check iterates `indiv_param_names`, so a branch-local (`if`-scoped) mapped `F` escapes it while `build_pk_param_fn` still binds it — the same defect being a hard error or fully silent depending on indentation. Filed as #1026; worth stating in the doc comment rather than leaving implied. |

TeunP:
## Fixes pushed — `462006df`

11 of the 13 review points applied. Full `--lib` suite green (3127 passed, 0 failed), `check_command` + `dose_attr_double_use_nonmem_anchor` green, `cargo clippy --no-default-features --features ci,markov` exits 0 with no new diagnostic, `cargo check --tests` clean, `cargo fmt --check` clean.

### 1 — `[initial_conditions]` arm removed

The rejection and its `record_coded_rate_reads` call are gone; the block now carries a note saying why it is deliberately not a check site. `analytical_dose_attr_read_in_initial_conditions_is_rejected` became `..._is_accepted` and covers `init(central)`, `init(depot)`, and the lag half. `docs/model-file/initial-conditions.qmd`, `individual-parameters.qmd`, `ode-models.qmd`, `check-report.qmd`, and the CHANGELOG entry all say the opposite of what they said.

### 2 — twin block-drop guard made real

Added `AbsorptionOdeEquivalent::source()` (`#[cfg(test)]`) and asserted the reconstructed source carries no `[adaptive_dosing]`. Re-ran the same mutation — `"adaptive_dosing"` deleted from the skip list at `model_parser.rs:8463`:

```
test parser::model_parser::tests::adaptive_observe_does_not_reach_the_absorption_twin ... FAILED
```

Was green under that mutation before, fails now.

### 3 + 4 — diagnostic is deterministic and quotes the file

`analytical_dose_attr_slot_map` iterates sorted entries; the message's role lookup moved into `analytical_role_binding`, which filters on `attr` and resolves through the same exact/lowercase rule that assigned the slot, returning the `role=value` pair verbatim. `f=X, lagtime=X` now yields one stable message whose noun and quoted role agree, and `alag=TLAG` binding a declared `tlag` says `remove the \`alag=TLAG\` mapping`. Two new tests, both mutation-verified against the pre-fix behaviour.

### Smaller

`to_lowercase()` hoisted out of the `position` closure; `analytical_dose_attr_read_only_for_reporting_is_accepted` added; the three stranded comments (`scaling_reads`, twin builder, `parse_error_to_diagnostic`) refreshed; the branch-local scope gap now stated on `check_dose_attr_double_use` rather than left implied.

### Left for you

- **The engine-split triplication.** Removing the init arm cuts it from three call sites to two. Collapsing the pair behind one wrapper with `n_states: Option<usize>` — which would also stop `usize::MAX` meaning two opposite things — is a signature change beyond a review fix, so I left it.
- **#1026 (branch-local mapped dose attribute).** Not closed; it is the registration bug on that issue, and fixing it there closes this hole with it. Now documented on the check.
- **The ODE `[odes] init(state) = <expr>` half of #993** has the same false positive this PR's init arm had — verified, `init(central) = F * 100.0` is rejected there too, and the ODE engine does not apply `F` to a state seed either. Pre-existing, untouched here; worth its own issue.

TeunP:
Filed the ODE half as #1046 — `[odes] init(state) = F * 100` hits the same false positive, since `OdeSpec::initial_state` seeds the raw expression value and dose attributes resolve only at dose events. Sweep there is also 0 affected models (one test fixture pins the wrong behaviour), so it lands as a widening, not a break. Not touched here.

Issue #1032, opened 2026-08-21
Epic: model-based meta-analysis (MBMA) support


roninsightrx:
Umbrella issue for making FeRx good at **model-based meta-analysis** (MBMA), collected from running two published analyses end to end against `ferx-core` 0.2.0 / `ferx-r` 0.2.0.9000.

## Context

FeRx **already does MBMA**, and does it well. On the naproxen/osteoarthritis case study (18 trials, 36 arms) it reproduces the published analysis to three significant figures:

| | FeRx | `nlmixr2` | Monolix (published) | Boucher NONMEM |
|---|---|---|---|---|
| naproxen effect on Emax | **0.792** (SE 0.065) | 0.793 (SE 0.065) | 0.792 (SE 0.064) | 0.79 |
| ET50 placebo / naproxen (wk) | 0.686 / 0.213 | 0.690 / 0.213 | 0.698 / 0.216 | 0.69 / 0.21 |
| ω(E0) | 0.622 | 0.622 | 0.627 | 0.62 |
| ω(Emax) | 0.738 | 0.739 | 0.763 | 0.74 |

It also does the whole job on the bounded-endpoint case study: between-study variability, between-treatment-arm variability as weighted IOV, SIR, and trial simulation.

> **Correction (2026-08-21).** An earlier version of this issue said `nlmixr2` has no inter-occasion variability and therefore could not do the second case study. **That is wrong** — `nlmixr2` supports occasion-level random effects natively with `eta ~ var | OCC`, and fitting the canakinumab model that way recovers BTAV γ = 2.023 (SD) against the published 2.005, BSV 0.273 against 0.253, and residual 2.457 against 2.466. My earlier numbers were measured on a BSV-only model, which was my own modelling limitation and not a property of the engine. Nothing else in this issue depends on that claim; FeRx and `nlmixr2` are both viable for both case studies.

**So none of what follows is about capability.** It is about two things: a few places where the engine is silently wrong on inputs that are normal in MBMA and rare elsewhere, and the fact that the DSL has no vocabulary for the one operation that defines this analysis class.

## The theme worth noticing

The three bugs already filed are all triggered by things that are **routine in MBMA and unusual elsewhere**:

| Issue | Trigger | Normal in MBMA because |
|---|---|---|
| #1020 | a signed `[scaling]` readout | the standard bounded-endpoint transform is a weighted **logit**, negative for any arm below 50% response |
[...]

PR #1037, opened 2026-08-21
fix(estimation): make SIR survive a rank-deficient proposal covariance [closes #1021]

### LINKED ISSUE #1021: ferx_sir: fails with 'All SIR samples had invalid weights' when FIXed parameters make the proposal covariance singular
## Summary

`ferx_sir()` fails with `All SIR samples had invalid weights` whenever the proposal covariance has **two or more effectively-null directions**.

The underlying cause is that `FIX`ed parameters are carried in `fit$cov_matrix` with a variance of exactly zero and are then sampled from. A fixed parameter has no uncertainty to sample; including it makes the proposal singular by construction. One such direction happens to be survivable; two are not — and a second `FIX`, or a near-flat ridge in the likelihood, supplies the second.

## Reprex

Bundled example only. Fixes a theta, a sigma, then both:

```r
library(ferx)
ex <- ferx_example("warfarin")
m  <- readLines(ex$model)

patch <- function(fix_theta, fix_sigma) {
  x <- m
  if (fix_theta) x[grep("theta TVKA",     x)] <- "  theta TVKA(1.0, FIX)"
  if (fix_sigma) x[grep("sigma PROP_ERR", x)] <- "  sigma PROP_ERR ~ 0.1 (sd) FIX"
  f <- tempfile(fileext = ".ferx"); writeLines(x, f); f
}

for (cfg in list(c(FALSE,FALSE), c(TRUE,FALSE), c(FALSE,TRUE), c(TRUE,TRUE))) {
  fit <- ferx_fit(patch(cfg[1], cfg[2]), ex$data, verbose = FALSE)
  ev  <- eigen(fit$cov_matrix, only.values = TRUE)$values
  r   <- try(ferx_sir(fit, sir_samples = 200L, sir_resamples = 100L), silent = TRUE)
  cat(sprintf("theta FIX=%-5s sigma FIX=%-5s  null dirs=%d  SIR: %s\n",
      cfg[1], cfg[2], sum(abs(ev) < 1e-12),
      if (inherits(r, "try-error")) conditionMessage(attr(r, "condition")) else "OK"))
}
```

```
theta FIX=FALSE sigma FIX=FALSE  null dirs=0  SIR: OK
theta FIX=TRUE  sigma FIX=FALSE  null dirs=1  SIR: OK
theta FIX=FALSE sigma FIX=TRUE   null dirs=1  SIR: OK
theta FIX=TRUE  sigma FIX=TRUE   null dirs=2  SIR: ferx_sir: All SIR samples had invalid weights
```

## A ridge substitutes for the second FIX

The second null direction does not have to come from a `FIX`. On a model-based meta-analysis with one fixed sigma plus a known likelihood ridge (two parameters that trade off, determining only their sum):

| Model | null dirs | ridge | 2nd smallest eigenvalue | SIR |
|---|---|---|---|---|
| MBMA as written (σ fixed, ridge present) | 1 | yes | 7.94e-4 | **fails** |
| same, ridge parameter removed from the model | 1 | no | 8.18e-4 | OK |

The two second-smallest eigenvalues are within 3% of each other, so this is not a conditioning threshold that could be papered over by inflating the proposal — it is the combination of an exactly-null direction with a nearly-null one.

## Why this bites MBMA specifically

Every model-based meta-analysis in the study-as-subject convention fixes the residual variance:

```
sigma ADD_ERR ~ 1.0 (variance) FIX
```

That is not a modelling choice that could be dropped — the observation is `DV = y / SE` and the prediction is divided by the same `SE`, so fixing σ to 1 *is* the inverse-variance weighting scheme. Every such model therefore starts with one null direction before anything else goes wrong, and `ferx_sir()` is exactly the tool one would reach for when the asymptotic covariance is doubtful, which on these models it usually is.

`fit$cov_matrix` for such a fit:

```r
diag(fit$cov_matrix)
#>   TVE0 TVEMAX TVET50 B_E0_FLARE B_EMAX_FLARE B_EMAX_TRT B_LET50_TRT ETA_E0 ETA_EMAX ADD_ERR
#> 0.0024 0.0987 0.0284     0.0994       0.1434     0.0042      0.0727 0.0289   0.0300  0.0000
```

`ADD_ERR` is fixed and contributes an exact zero. Estimating it instead makes SIR run — but it dismantles the weighting, so it is a diagnostic rather than a workaround.

## Suggested fix

Exclude `FIX`ed parameters from the SIR proposal and from the resampled draws — hold them at their fixed values and sample only the free block of the covariance. That is correct on its own terms (a fixed parameter has no posterior to sample) and it removes one null direction for free, which is enough to make the common MBMA case work.

Two smaller things that would help independently:

- The error is uninformative. `All SIR samples had invalid weights` gives no hint that the proposal was singular. Reporting the rank deficiency, and naming the offending parameters, would turn a dead end into a diagnosis.
- Consider warning at fit time when `cov_matrix` is rank-deficient for a reason other than `FIX` — that is the ridge case, and it is worth surfacing regardless of SIR.

## Environment

```
ferx-r     0.2.0.9000
ferx-core  0.2.0
R          4.5.3, aarch64-apple-darwin
```

roninsightrx:
## Why

`ferx_sir()` / `sir = true` failed with `All SIR samples had invalid weights` on models whose
covariance matrix has a direction the data do not identify — reported in #1021, and the common
case for model-based meta-analysis (#1032), where fixing the residual variance *is* the
inverse-variance weighting scheme and cannot be dropped.

The issue's proposed cause — `FIX`ed parameters being sampled from — is **not** the live one:
SIR has restricted the proposal to the free block since b0f2085f (in v0.2.0), and both the
`ferx_sir()` R path and the inline path carry the FIX flags correctly. The actual mechanism is
downstream of that:

`compute_covariance` floors the FD Hessian's near-zero eigenvalues before inverting it
(`invert_psd_with_floor`, floor = `λ_max · 1e-10`). A non-identified direction therefore comes
back in `covariance_matrix` with variance ≈ `1/floor` — 1e7 and up, i.e. a proposal standard
deviation of **thousands** in packed (log) space. Every draw along that direction lands outside
the packed parameter bounds, is rejected, and the run dies with no clue why. That is exactly what
the fit's own `Covariance step regularized: eigenvalue floor applied to FD Hessian (3 of 6
free-block eigenvalues clipped ...)` warning is telling you, one step earlier.

## What changed

**1. The free block is conditioned before it is sampled** (`condition_free_proposal`, `sir.rs`):

- **Cap:** each eigen-direction's variance is capped so `sqrt(λ)·|v_k| ≤ (upper_k − lower_k)/4` —
  ±2 proposal standard deviations stay inside the packed bounds of every coordinate it loads on.
  The cap is deliberately loose: it bites only a direction whose proposal is wider than the box the
  parameter is declared to live in, so a legitimately wide uncertainty is untouched.
- **Floor:** near-null directions (a likelihood ridge surviving the FIX exclusion — the "second
  null direction" in the issue) are floored instead of being fatal. The floor is anchored on the
  **capped** spectrum, not the raw one: anchored on the raw spectrum a single exploded direction
  (1e8) puts the relative floor at ~1e-2 and reports every well-determined direction (~1e-4) as
  null, inflating real variances by orders of magnitude. There is a unit test pinning this.
- Sampling and the Student-t density both use the conditioned covariance (its `log|C|` comes from
  the modified eigenvalues), so the importance weights stay self-consistent.

**2. Both repairs are reported**, as `SIR:` fit warnings naming the parameters, e.g.

```
SIR: proposal was shrunk in 3 direction(s) so draws stay inside the parameter bounds
[TVKA -0.74, ETA_KA -0.60, ETA_V +0.28 (sd 6.15e3 → 2.89e0); ...]. Those directions come from
eigenvalue-floored (non-identified) curvature in the covariance step; the SIR CIs along them
understate the true uncertainty.
```

**3. The dead-end error is now a diagnosis.** When every sample really is rejected, the message
reports the rejection tally by cause, the coordinates whose bounds were overshot, and the
proposal's rank deficiency:

```
All 8 SIR samples had invalid weights (rejected: 8 out of bounds, 0 non-positive
theta/sigma/omega, 0 non-finite OFV, 0 cancelled). Bound first overshot by: TVCL (8) (count of
samples, first offending coordinate only). Proposal covariance is rank-deficient beyond the
FIX-ed parameters: 2 direction(s) carry no uncertainty [PROP_ERR -1.00; TVV -1.00]. ... Fix or
drop one parameter from each listed combination, or re-fit a model the data can identify.
```

The issue's third suggestion — warn at fit time when the covariance is rank-deficient for a reason
other than FIX — is already implemented: that is the `Covariance step regularized: eigenvalue floor
applied to FD Hessian ...` warning, which fires whenever free-block eigenvalues were clipped. The
docs now connect the two.

## Alternatives considered

- **Condition-number capping** (as `build_non_pd_fallback_proposal` does for the non-PD path):
  doesn't rescue this. The exploded direction *is* `λ_max`, so capping `λ_max/λ_min` at 1e8 still
  leaves a proposal sd in the hundreds. The bound-derived cap is the one that maps onto why the
  samples are rejected.
- **Uniform down-scaling of the whole proposal**: preserves correlation structure but shrinks
  well-determined directions too. Per-direction capping leaves identified directions bit-identical.
- **Dropping null/exploded directions from the proposal entirely**: equivalent to fixing those
  combinations at their ML values, with no warning surface and a discontinuous change in the
  Student-t dimensionality. Flooring + capping degrades continuously and is reportable.

## Cross-repo dependency

`SirResult` gains a `warnings: Vec<String>` field; nothing in ferx-r constructs `SirResult`, so
nothing breaks. But `ferx_rust_sir()` builds its return list by hand and has no `warnings` slot, so
R users of `ferx_sir()` get the improved **error** (it is thrown) but not the new **warnings**. A
small follow-up in ferx-r should add `warnings = new_fit.warnings` to that list and merge it into
the fit — happy to open it.

## Breaking changes
- [x] None — `SirResult` is additive; no `.ferx`, sdtab, or `FitResult` field changes.

## Numerical validation
- [x] Compared against: prior ferx commit `ae6c208c`

Bundled warfarin + `data/warfarin.csv`, `sir_samples = 200`, `sir_resamples = 100`, matching the
issue's reprex (a theta FIXed, a sigma FIXed, then both). Healthy fits are **bit-identical**
before and after — the cap never fires on them:

```
                                   before (ae6c208c)      after
theta FIX=F sigma FIX=F   ESS      53.9                   53.9
theta FIX=T sigma FIX=F   ESS      87.9                   87.9
theta FIX=F sigma FIX=T          FAILED "All SIR samples  ESS 1.0 + shrinkage warning
                                  had invalid weights"    (3 of 6 free-block eigenvalues
                                                           were floored in the cov step)
theta FIX=T sigma FIX=T   ESS      6.6                    6.6
```

The `sigma FIX` case is a deliberately misspecified fit (PROP_ERR pinned far from its ML value),
which is what makes its Hessian non-PD; ESS 1.0 is the honest answer for a proposal that carries
no usable information in three directions, and the warning says so rather than the run dying.

No NONMEM anchor applies: this is SIR proposal conditioning (a sampler-robustness property), not a
new estimator or likelihood term. NONMEM's `$COVARIANCE` has no equivalent object — its SIR is a
PsN wrapper with its own proposal handling — and the numbers on all healthy fits are unchanged.

## Performance
- [x] No significant impact expected — one extra symmetric eigendecomposition of the free block
  (n_packed ≤ ~50) per SIR run, against M inner-loop optimizations.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3034 passed)

`estimation::sir::tests` (7 new): healthy block passes through untouched (`L·Lᵀ` and `log|C|`
exact); an explosive direction is capped and named; a ridge is floored and both its parameters
named; two null directions survive; an exploded eigenvalue does not swamp the floor; both error
paths; and the failure-message content.

`estimation::run_sir::tests::run_sir_survives_an_explosive_proposal_direction` is the end-to-end
regression: fit warfarin, add 1e8 to one covariance diagonal (keeps the matrix PSD — a covariance a
real fit could hand us), and assert SIR completes and warns. Verified to fail without the cap with
exactly the pre-fix symptom.

- [x] Not applicable (fix; no new API surface or DSL syntax)

## Docs
- [x] `docs/` updated — new "Rank-deficient and non-identified proposals" section in
  `docs/estimation/sir.qmd` (existing page, no `_quarto.yml` change), covering FIX handling, the
  ridge and explosive-direction cases, the warnings, and the improved error.

## Checklist
- [x] `cargo clippy` clean for the touched code (no new warnings introduced; the crate's
  pre-existing warning count is unchanged)
- [x] No `Dual2` / sensitivity-path code touched
- [x] `Cargo.toml` version unchanged (non-breaking)

## Reviewer hints

The heart is `condition_free_proposal` in `src/estimation/sir.rs` — two passes (cap, then floor)
and why the order matters. `PROPOSAL_BOUND_SIGMAS = 4.0` is the one tuned constant: it decides how
wide a proposal direction must be before it is treated as garbage rather than as uncertainty. The
rest (rejection tally, message builder, warning plumbing at the three `run_sir_core` call sites) is
mechanical.

## Open questions

- A fit that now *completes* with ESS ≈ 1 still prints CIs that mean very little. Should a low-ESS
  warning (say ESS < 10, or ESS < 1% of `sir_samples`) be added? It is outside this issue's scope
  but is the natural companion to these warnings.
- `simulate_with_uncertainty(method = "asymptotic")` samples the same covariance and will exhaust
  its retry budget on such a fit. Its error already names the cause ("covariance may be
  ill-conditioned or near a bound"); reusing this conditioning there would make it work instead.
  Left out deliberately as scope.


roninsightrx:
## Code review — `fix-1021-sir-rank-deficient-proposal` (effort: medium)

The core idea (eigen-cap + eigen-floor the free block before sampling, then report it) is sound, and the weight math stays self-consistent — `log_det_proposal` cancels between `log_q_k` and `log_q_hat`, so the changed determinant cannot bias weights. The `SampleOutcome` tally, the `.position()` rewrite of the bounds check, and the `coord_names`/`free_names` index domains all line up correctly. Findings below are about the cap criterion and reporting, not the plumbing.

### Medium

**1. `src/estimation/sir.rs:435` — the bound cap assumes x̂ is centred in the box**

The half-width is `(upper[i] − lower[i]) / 4`, i.e. it assumes the proposal centre x̂ sits in the middle of the box. It does not. Concrete case: a theta declared `(0, 100)` with ML estimate 0.5 packs to bounds `[ln(1e-10), ln(100)] = [-23.0, 4.6]` (width 27.6, half-width 6.9) with x̂ = −0.69. Distance to the *upper* bound is 5.3, but the cap permits ±13.8. An exploded direction loading on that theta is "capped" and still puts the large majority of draws above the upper bound — the exact `All SIR samples had invalid weights` failure this PR is fixing, now with a warning claiming the proposal was made bound-safe.

Fix: use `min(x̂_k − lower_k, upper_k − x̂_k)` (times `2 / PROPOSAL_BOUND_SIGMAS`), which requires threading `x_hat[free_idx]` into `condition_free_proposal`.

### Low

**2. `src/estimation/sir.rs:185` — the "±2 sd inside the bounds" guarantee overstates what the cap buys**

Two further reasons: (a) the cap is applied per eigen-direction, but coordinate `k`'s marginal variance is `Σ_i λ'_i v_ki²`, so with several capped directions the realised per-coordinate sd can exceed the half-width by up to √n; (b) the proposal is Student-t with `sir_df = 5`, whose variance is `ν/(ν−2) = 1.67 ×` the scale matrix and whose tails give `P(|t₅| > 2) ≈ 10%` per coordinate. Even a perfectly centred x̂ therefore rejects a noticeable fraction of draws along a capped direction. Either tighten `PROPOSAL_BOUND_SIGMAS` for the t-inflation, or drop the "±2 sd stay inside" phrasing from the doc comment.

**3. `src/estimation/sir.rs:435` — cap is much tighter for omega/sigma than the comment implies**

The claim that the cap "bites only a direction wider than the box the parameter is declared to live in" is weak for omega/sigma: `compute_bounds` gives omega log-Cholesky diagonals `[-6, 6]` and sigma `[-8, 5]`, so their half-widths are only 3.0 and 3.25. A collapsing omega (a common outcome) legitimately has log-Cholesky uncertainty of that order, so its direction gets shrunk, its CI silently narrowed, and the user is told the direction comes from "eigenvalue-floored (non-identified) curvature" — a misdiagnosis for a merely imprecise omega. Consider gating the cap on a much larger overshoot (e.g. cap only when `sqrt(λ)·|v_k| > c · half_width` for `c ≫ 1`) so it targets the 1e3–1e4 sd signature it is aimed at.

**4. `CHANGELOG.md:191` — ±3 vs ±2 mismatch**

Changelog says "capped so ±3 standard deviations stay inside the packed bounds"; the code (`PROPOSAL_BOUND_SIGMAS = 4.0` ⇒ ±2), the doc comment, and `docs/estimation/sir.qmd` all say ±2. Fix the changelog.

**5. `src/estimation/run_sir.rs:207` — SIR warnings accumulate on repeated passes**

`out = fit.clone()` then unconditionally appends `SIR: …`. If the input `fit` was produced with `sir = true` (which already pushed the same `SIR:` warnings in `fit_inner`), or if a user pipes one `run_sir` output into another, identical warning text accumulates in `FitResult.warnings` and prints once per pass. Dedupe against existing warnings, or strip prior `SIR:` entries from the clone.

**6. `src/estimation/sir.rs:134` — `describe_direction` can degrade to a contentless message**

Loadings are filtered at `|v| ≥ 0.15` before the fallback. For a model with many free coordinates, a degenerate direction spread across ~50 parameters has every loading below 0.15, so the warning becomes `1 direction(s) carry no uncertainty [no dominant parameter]` — no actionable content, and the follow-on advice "Fix or drop one parameter from each listed combination" has nothing to point at. Falling back to the top-3 loadings regardless of magnitude would keep the message useful.

**7. `src/estimation/sir.rs:229` — healthy fits are no longer bit-reproducible**

Even a strictly-PD, well-conditioned free block is now round-tripped through `symmetric_eigen` → `V Λ Vᵀ` → `cholesky` rather than taking `sub_cov.cholesky()` directly. The reconstruction differs from the input at ~1e-16 relative (the new healthy-block test asserts 1e-12, not bit-equality), which perturbs every drawn sample and therefore the inner-loop EBEs and weights. SIR output for healthy fits is not bit-identical to pre-PR versions, contrary to the PR description. Cheap fix: short-circuit to the direct Cholesky when no eigenvalue was capped or floored.

---

roninsightrx:
Cross-repo follow-up opened: FeRx-NLME/ferx-r#302 (draft).

It adds the missing `warnings` slot to `ferx_rust_sir()`'s return list and folds the messages onto the fit (`fit$warnings` + `sir`-category rows in `fit$warnings_structured`), so `ferx_sir()` users actually see the proposal diagnostics this PR adds — previously they died at the FFI boundary. It also gives the `sir` warning category message-specific guidance, since the old "raise sir_samples" advice is wrong for a direction the data do not inform.

That PR must merge **after** this one, followed by a `tools/update-ferx-core-lock.sh` bump — its end-to-end test asserts a warning only this branch emits.

Also pushed here: `f9d80383`, which routes the new `SIR:` / `SIR fallback:` messages to the `sir` warning category. They matched none of the classifier's existing SIR phrasings, so on the inline `sir = true` path they were landing in `general` with no category guidance; the shrinkage message also name-drops "the covariance step", so it needed an explicit prefix match to stay out of the covariance categories.

roninsightrx:
## Review follow-up — all 7 findings addressed in e712e0c6

**1 (medium) — cap assumed x̂ centred in the box.** Fixed. New `proposal_sd_caps()` in `sir.rs` measures room as `min(x̂ − lower, upper − x̂)` and caps at `room · 2 / PROPOSAL_BOUND_SIGMAS`, so the ±2 sd guarantee holds against the *nearer* bound. The `(0, 100)` theta at `0.5` now caps at ~2.6 rather than 13.8. Room is floored at `PROPOSAL_MIN_ROOM_FRAC` (1%) of the box width so an estimate sitting exactly on a bound still yields a PD proposal instead of aborting with "no positive eigenvalue" — a regression the naive `min()` would otherwise have introduced.

**2 (low) — "±2 sd inside the bounds" overstated.** Partly fixed, partly documented. The Student-t inflation is now compensated in the cap: `sd_cap` is divided by `sqrt(ν/(ν−2))`, so ±2 sd is a statement about the *draws*, not the scale matrix (`sir_df ≤ 2` falls back to the raw scale rather than dividing by NaN). The per-direction-vs-per-coordinate accumulation is not compensated — a global rescale would shrink the identified directions too — but with the trigger gate below only a handful of directions are capped at all, and both `PROPOSAL_BOUND_SIGMAS`'s doc comment and `docs/estimation/sir.qmd` now say so explicitly instead of claiming an exact guarantee.

**3 (low) — cap too aggressive for omega/sigma.** Fixed. New `CAP_TRIGGER_FACTOR = 10.0`: a direction is only shrunk once it overshoots its cap by 10× in sd (100× in variance). A collapsing-but-real omega (sd ~3 against a cap of ~1.5) is now left alone; the 1e3–1e4 overshoot the cap exists for still trips it. Covered by `condition_free_proposal_leaves_a_legitimately_wide_direction_alone` and `..._still_caps_past_the_trigger`.

**4 (low) — changelog ±3 vs ±2.** Fixed, and reworded to say the room is measured to the nearer bound.

**5 (low) — SIR warnings accumulate.** Fixed. Extracted `push_sir_warnings()` in `run_sir.rs`, which skips lines already present. Handles both the `sir = true` input fit and `run_sir`-piped-into-`run_sir`; idempotence is asserted in `push_sir_warnings_does_not_duplicate`. The `run_sir` doc comment claiming warnings are left unchanged was stale since the original #1021 commit — corrected.

**6 (low) — `describe_direction` contentless message.** Fixed. Loadings below `DIRECTION_LOADING_MIN` are still suppressed when a dominant one exists; when *none* clears the threshold it now names the top three regardless of magnitude. Both branches tested (100-coordinate spread direction, and a 0.99/0.1/0.05 direction that must not gain the noise).

**7 (low) — healthy fits no longer bit-reproducible.** Fixed. When no eigenvalue was capped or floored, `condition_free_proposal` factors the symmetrised input directly and skips the `V Λ Vᵀ` round trip; the reconstruction path is kept as the fallback. `condition_free_proposal_is_bit_identical_on_a_healthy_block` asserts `to_bits()` equality against `sub_cov.cholesky()` on a 3×3 block, so the PR description's bit-identical claim now holds.

### Verification

`cargo test --lib` — 3047 passed, 0 failed. `cargo fmt --check` and `cargo clippy --lib` clean on the touched files.

Unrelated pre-existing issue, not from this PR: `cargo clippy --tests` fails on four `clippy::approx_constant` denials in `src/sens/two_cpt_explicit.rs` (`3.14` used as a PK parameter value in test fixtures). That file is byte-identical to `origin/main`, so it is a main-branch problem, not a blocker here.

---

roninsightrx:
Review addressed — all seven findings, in `e712e0c6`.

Note on provenance: I worked these findings in parallel with another session and we landed on the same design (nearest-bound room, overshoot hysteresis, direct Cholesky for an untouched spectrum, loading fallback, warning dedupe, changelog fix). `e712e0c6` was pushed first and goes further in two places, so I dropped my duplicate rather than merging two versions of the same change:

- it compensates the **Student-t scale inflation** (`sqrt(nu/(nu-2))`), so the ±2 sd budget is about the *draws* rather than the Cholesky scale — that is finding 2's substantive half, which I had only handled by softening the doc wording;
- it **floors the usable room** at 1% of the box width, so an estimate sitting exactly on a bound still yields a PD proposal instead of tripping "no positive eigenvalue". My version clamped the room to zero and leaned on the eigen floor, which is the weaker answer.

Its trigger is 10× where mine was 8×; no reason to prefer mine.

Verified at `e712e0c6` (warfarin, 200/100, the four configs from #1021's reprex):

```
theta FIX=F sigma FIX=F   ESS 53.9     unchanged
theta FIX=T sigma FIX=F   ESS 87.9     unchanged
theta FIX=F sigma FIX=T   ESS  1.0     + shrink warning (was: "All SIR samples had invalid weights")
theta FIX=T sigma FIX=T   ESS  6.6     unchanged
```

The healthy numbers are now bit-identical *by construction*, not just numerically indistinguishable — finding 7's short-circuit means a block that needs no conditioning is factored directly. Shrink targets moved from sd 2.89/3.58/3.60 to 1.84/2.03/2.37, which is the t-inflation compensation plus the nearest-bound room doing their work. `cargo test --lib`: 3047 passed.

On finding 3 specifically: the omega risk was real, and worth restating for the record — an omega log-Cholesky diagonal is bounded `[-6, 6]`, so a collapsing-but-real omega had packed uncertainty of the same order as its safe width and was being shrunk *and* labelled non-identified. The 10× gate separates that from the 1e3–1e4 sd signature the cap is actually for.

Also pushed downstream: FeRx-NLME/ferx-r@101fade updates the `sir` warning guidance text in ferx-r#302, since "wider than the box the parameter is declared to live in" no longer describes the criterion.

Issue #650, opened 2026-07-01
feat(scaling): allow full 'y = <expr>' output readout (Form C) on analytical PK models for flexible multi-DV residual errors


roninsightrx:
## Summary

Allow a full, arbitrary output-readout expression — `[scaling] y = <expr>` (Form C) — on **analytical** PK models, so residual errors can be attached to flexibly-defined DV types (parent/metabolite, free-vs-total drug, protein-binding corrections). Today Form C is **ODE-only**; analytical models can only post-scale the built-in concentration by a *divisor* (`obs_scale`, Form B), which cannot express additive nonlinear readout terms.

## Motivating example

A user wants a total-vs-free concentration readout on an analytical 1-cpt model, switched by a data covariate `FREE`, with a saturable-binding (albumin/BMAX) correction:

```
[scaling]
  if (FREE == 0) {
    y = central / V1 + BMAX * ALBBMAX * (central / V1) / (KD + central / V1)
  } else {
    y = central / V1
  }
```

This is a legitimate parent/metabolite-style multi-DV pattern (a nonlinear binding correction applied to the total-drug rows only). It cannot be written today for an analytical model. Example for this is the fluconazole PK model.

## Why it doesn't work today

1. **Form C is rejected on analytical models.** `[scaling] y = <expr>` requires an ODE model — `src/parser/model_parser.rs:5622-5627`:
   > `[scaling]: 'y = <expr>' (Form C) requires an ODE model — use 'obs_scale = <expr>' for analytical PK`
2. **`obs_scale` is a divisor, not a readout.** Analytical models only get Form B `obs_scale = <expr>`, applied divisively as `pred_scaled = pred_raw / scale` (`src/types.rs:2058-2069`). The binding term `+ BMAX*ALBBMAX*C/(KD+C)` is **additive** and cannot be expressed as a divisor.
3. **Selection is keyed on CMT, not an arbitrary covariate.** Multi-DV is supported via `ScalingSpec::PerCmt` / `ErrorSpec::PerCmt` (keyed on the data `CMT` column; there is no `DVID` concept). The example above keys on covariate `FREE`, and even the per-CMT `y[CMT=N]` readout is ODE-only.

Note: conditional logic itself is **not** the blocker — `parse_scalar_expression` already supports `if/else`, and the residual-*magnitude* multiplier (#484) accepts `if/else` expressions. The gap is specifically the analytical-model ban on a full `y = <expr>` readout.

## Proposed direction

[...]

PR #1038, opened 2026-08-21
fix(ode): stop clamping negative Form C [scaling] readouts to zero [closes #1020]

### LINKED ISSUE #1020: ODE Form C readout: negative [scaling] predictions silently clamped to zero
## Summary

`ode_predictions` clamps the **final, post-readout** prediction vector at zero. That is right for a concentration, but it is applied unconditionally — including to a Form C `[scaling]` `y[CMT=N] = <expression>` readout, which is an arbitrary user expression and is often legitimately negative.

Any negative part of such a readout silently becomes `0`. No warning, no error; the fit converges and returns plausible-looking parameters.

## Reprex

`negpred.ferx` — the readout is a straight line crossing zero:

```
model negpred
[parameters]
  theta TVA(1.0, 0.1, 10.0)
  omega ETA_A ~ 0.01
  sigma ADD_ERR ~ 1.0 (sd)
[individual_parameters]
  A = TVA * exp(ETA_A)
[structural_model]
  ode(states=[clock])
[odes]
  d/dt(clock) = 1
[scaling]
  y[CMT=1] = A * (clock - 5)
[error_model]
  DV ~ additive(ADD_ERR)
[fit_options]
  method = focei
  gradient = fd
  covariance = false
```

`negpred.csv`:

```
ID,TIME,DV,EVID,AMT,CMT,MDV
1,0,0,1,0,1,1
1,2,-3,0,0,1,0
1,4,-1,0,0,1,0
1,6,1,0,0,1,0
1,10,5,0,0,1,0
```

```r
library(ferx)
ferx_predict("negpred.ferx", "negpred.csv")
#>   ID TIME PRED   expected
#> 1  1    2    0         -3
#> 2  1    4    0         -1
#> 3  1    6    1          1
#> 4  1   10    5          5
```

At `TVA = 1` the readout is `clock - 5`. The two negative points are returned as `0`; the positive ones are exact.

## Cause

`src/ode/predictions.rs`:

```rust
    // Clamp negative predictions to zero (ODE solver overshoot guard).
    // NaN intentionally NOT clamped ...
    clamp_negative_predictions(&mut predictions);      // L2501
```

```rust
fn clamp_negative_predictions(predictions: &mut [f64]) {   // L1592
    for p in predictions.iter_mut() {
        if *p < 0.0 { *p = 0.0; }
    }
}
```

This runs on the final vector, i.e. **after** `OdeReadout::PerCmt` / `Single` has evaluated the `[scaling]` expression (`L1534`). The overshoot guard is a statement about a *state* (an amount or concentration cannot go below zero), but it is being enforced on the *readout*, which the code itself documents as arbitrary:

```rust
    /// Both Form C variants (`Single` and `PerCmt`) call arbitrary ...   // L1818
```

## Why it matters

Any signed readout is affected: change from baseline, a z-score, a difference from a comparator, an effect on a signed scale — and logits.

Concretely, it silently breaks the standard model-based meta-analysis of a bounded endpoint. The published construction (Boucher & Bennetts; Bracis et al., *CPT:PSP* 2026) transforms a responder proportion as

```
DV       = sqrt(N) * logit(observed proportion)
y[CMT=1] = sqrt(N) * logit(predicted proportion)
```

which is negative for every arm with a response rate below 50% — most of a real dataset. Fitting the published canakinumab/RA ACR20 model against its published Monolix estimates:

| | ferx | Monolix (published) |
|---|---|---|
| T50 (weeks) | 2.67 | 4.08 |
| BSV ω (SD) | 0.080 | 0.253 |
| residual σ (SD) | 3.03 | 2.47 |
| log-OR canakinumab | 1.111 | 1.104 |

Every placebo-arm prediction comes back as exactly `0`. σ inflates to absorb the mismatch and ω collapses. The drug effects survive because they are identified largely by the arms that do exceed 50%.

The companion continuous-endpoint model (WOMAC pain / SE, strictly positive) reproduces the same published analysis to three digits, which localises it to the sign of the readout rather than anything else in the model.

## Suggested fix

Apply the clamp only where the non-negativity is a physical property of the quantity — a bare state/concentration readout (`obs_scale`, analytical PK, generated `ode_template`) — and leave Form C `[scaling]` expressions unclamped. The NaN behaviour documented in the same comment should be kept either way.

If the clamp is load-bearing for the "missing `PerCmt` entry surfaces loudly" case mentioned in the comment, that check could be made explicit rather than riding on the sign.

A per-model opt-out (`[fit_options] clamp_predictions = false`) would be a workable interim, but the default is the problem: the current behaviour is silent, and it is not detectable from the fit output.

## Environment

```
ferx-r     0.2.0.9000
ferx-core  0.2.0
R          4.5.3, aarch64-apple-darwin
```

roninsightrx:
## Why

Issue #1020. `ode_predictions` clamped the **final, post-readout** prediction vector at zero. That is right for a concentration, but it was applied unconditionally — including to a Form C `[scaling]` readout (`y = <expr>` / `y[CMT=N] = <expr>`), which is an arbitrary user expression and is often legitimately negative.

Any negative part of such a readout silently became `0`: no warning, no error, and nothing in the fit output to detect it. The fit converged with plausible-looking parameters, the residual σ inflated to absorb the mismatch, and the between-subject variance collapsed.

The reporter hit this on the standard model-based meta-analysis of a bounded endpoint (Boucher & Bennetts; Bracis et al., *CPT:PSP* 2026), where the responder proportion is transformed as `sqrt(N) * logit(p)` — negative for every arm below 50%, i.e. most of a real dataset. Other signed readouts affected: change from baseline, a difference from a comparator, a z-score, an effect on a signed scale.

The overshoot guard is a statement about a **state** (an amount or concentration cannot go below zero, so a negative value is RK overshoot). It was being enforced on the **readout**, which the code itself documents as arbitrary. The analytical Form C path (#650) already clamps the concentration fed *into* the readout and leaves the readout's output alone, so the two engines disagreed.

## What changed

- New `OdeReadout::clamps_negative()` — true only for the bare-state `ObsCmt` readout, false for both Form C variants (`Single`, `PerCmt`).
- `clamp_negative_predictions` takes the readout and is a no-op unless it clamps. That covers every driver sharing the epilogue: dense (`ode_predictions*`), dense-with-states, adaptive (`ode_predictions_adaptive_impl`), and the frozen-schedule replay.
- The event-driven walk's inline per-observation clamp takes the same gate.
- `sens/ode_provider.rs::resolve_obs_readout` (the `Dual2`/`Dual1` readout site) takes the **same** gate, so the analytic gradient still equals FD of the f64 predictor — a clamped jet carries zero derivatives, an unclamped one does not, so a gate that drifted between them would be a silent wrong-gradient bug.
- `NaN` behaviour is unchanged everywhere: it survives `< 0.0` per IEEE 754, so a missing `PerCmt` entry or a bad scale still surfaces as a NaN objective. The "missing entry surfaces loudly" case never rode on the sign — `OdeReadout::eval` returns `NaN` for a map miss.
- The `ObsCmt` clamp is untouched.

Not changed (deliberate, flagged for follow-up): the adaptive-dosing **assay** clamp (`(latent + √var·eps).max(0.0)` in the monitor readout) is a separate, documented statement about an assay — "an assay cannot read below zero" — not the solver guard. A signed Form C readout used as a feedback-dosing signal would still be clamped there; worth a separate issue if anyone needs it.

## Alternatives considered

A per-model opt-out (`[fit_options] clamp_predictions = false`), as the issue suggests as an interim. Rejected: the default is the problem, and the readout type already tells us exactly when the guard is meaningful. No new option, no migration.

## Cross-repo dependency

No public API change; `OdeReadout::clamps_negative` is additive.

## Breaking changes
- [x] None

Behaviour changes for models with a Form C readout that goes negative — which is the fix. A model whose readout is non-negative everywhere is bit-identical.

## Numerical validation

- [x] Compared against: **NONMEM 7.5.1**

`nonmem_anchor/signed_readout.ctl` — `ADVAN13` one-compartment IV bolus written as `$DES`, read out as a change from baseline, all θ `FIX`, `$OMEGA 0 FIX`, `MAXEVAL=0`:

```
$DES    DADT(1) = -K*A(1)
$ERROR  IPRED = A(1)/V - BASE     ; CL=4, V=12, BASE=2
```

| t (h) | NONMEM `IPRED` | ferx (this PR) | ferx (before) |
|---|---|---|---|
| 0.25 | 5.6670 | 5.667037 | 5.667037 |
| 1 | 3.9711 | 3.971094 | 3.971094 |
| 4 | 0.19664 | 0.196643 | 0.196643 |
| 12 | −1.8474 | −1.847370 | **0** |
| 24 | −1.9972 | −1.997204 | **0** |

NONMEM applies no non-negativity guard to `$ERROR` output. Every point now agrees to within the reference table's own 5-significant-figure rounding (worst case 1.4e-5 relative). Pinned in `tests/signed_readout_nonmem_anchor.rs` (Tier 2 — `predict` at fixed parameters, no convergence loop).

The issue's own reprex, via the CLI:

```
ID,TIME,DV,PRED,IPRED,...
1,2,-3,-2.999970,-3.000000
1,4,-1,-0.999990,-1.000000
1,6, 1, 0.999990, 1.000000
1,10,5, 4.999951, 5.000000
```

(was `0, 0, 1, 5`).

`Dual2`-vs-FD parity: `ode_provider_signed_form_c_readout_not_clamped_and_matches_production` runs the full provider harness (`check_vs_production` + `check_hessian_vs_fd_of_grad` + `check_inner_outer_eta_parity`) on a fixture whose readout straddles zero, and asserts the fixture is genuinely negative first so the check can't pass vacuously with both sides clamped.

## Performance
- [x] No significant impact expected — one enum match per prediction vector (and it now skips a pass over the vector for Form C models).

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3028 passed)

New:
- `only_bare_state_readouts_clamp_negatives` — the gate's truth table.
- `form_c_per_cmt_readout_keeps_negative_predictions` — the issue's reprex as a Tier-1 spec (clock system, readout `A*(clock-5)`), exact values.
- `form_c_single_readout_keeps_negative_predictions_on_all_drivers` — dense, dense-with-states, and event-driven drivers.
- `ode_provider_signed_form_c_readout_not_clamped_and_matches_production` — analytic/FD parity on a signed readout.
- `tests/signed_readout_nonmem_anchor.rs` — the NONMEM anchor above.

All four fail with the gate flipped back to the old behaviour (verified by temporarily restoring it).

Kept green: `test_ode_predictions_still_clamps_negatives` (the `ObsCmt` clamp) and `test_ode_predictions_propagates_nan_from_readout` (the NaN tripwire).

- [x] Not applicable (fix with no new API surface) — the DSL is unchanged; `[scaling] y[CMT=1] = sqrt(N) * log(P / (1 - P))` simply works now.

## Docs
- [x] `docs/` updated — new "Signed readouts (since #1020)" section in `docs/model-file/scaling.qmd`, with the NONMEM comparison table and a note that a difference-from-baseline readout wants tight `ode_reltol`/`ode_abstol` (the subtraction amplifies solver error).

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the dual readout site's gate mirrors the f64 one exactly, covered by the parity test above.

## Docs & examples
- [x] No example execution step needed (no new DSL syntax, no new fit option, no new example model).

## Reviewer hints

The load-bearing bit is that the f64 gate (`src/ode/predictions.rs`) and the dual gate (`src/sens/ode_provider.rs:resolve_obs_readout`) must stay identical — that's what keeps analytic gradients equal to FD. Everything else is mechanical.

Worth a second look: whether the adaptive-dosing assay clamp (`predictions.rs`, the `AssayNoise` branch) should follow. I left it — it is a documented statement about an assay reading, not the solver guard — but a signed readout used as a controller signal is clamped there today.

## Open questions

None.


roninsightrx:
## Review — no correctness bug in the core change

Verified locally on `worktree-1020-formc-negative-readout`:

- `cargo test --lib` — 3028 pass; `cargo check --tests` clean;
  `tests/signed_readout_nonmem_anchor.rs` passes.
- The NONMEM anchor is a real 7.5.1 run and self-consistent — at `t = 0`,
  `IPRED = 100/12 - 2 = 6.3333` matches the `.tab`.
- The `clamps_negative()` gate is applied at **every** clamp site: no
  ungated caller of `clamp_negative_predictions` or of the inline clamps
  remains, and both dual-readout sites (`integrate_subject_duals`,
  `integrate_tvcov_readout`) route through the single `resolve_obs_readout`.
- Out-of-scope-by-construction, confirmed: the SDE/EKF clamps are unreachable
  for Form C (the parser rejects that combination), hazard and binary endpoints
  do not ride the readout, and squared residual variance is sign-safe.

Three low-severity findings; two are now handled.

---

### 1. `docs/model-file/scaling.qmd:116` — doc was wrong about Form B *(fixed in 1fffbd51)*

The page said the clamp "is not applied to Forms B or C". That is wrong for
Form B. Forms A and B (`obs_scale`) leave the readout as
`OdeReadout::ObsCmt` — the divisor is applied *downstream* of the readout — so
`clamps_negative()` is `true` and the clamp still runs on the compartment state
before the divide. A reader following the old wording would expect an
`obs_scale = <expr>` model's negative prediction to survive; it does not.

Corrected to "not applied to Form C", plus a paragraph spelling out why A and B
still clamp.

### 2. `src/ode/predictions.rs:1597` — one engine divergence the PR text says is closed is still open *(not fixed; non-blocking)*

The PR description says the analytic and ODE Form C paths "now agree". They
agree on the readout output, but not on the state feeding it: the analytic Form
C path clamps the *concentration fed into* the readout, while the ODE path now
clamps nothing at all.

For the canonical `[scaling] y = central / V`, an RK overshoot at a late trough
returns a small negative IPRED on the ODE engine and exactly `0` on the
analytic twin. Numerically negligible — residual variance goes as `f²` — so this
is not a merge blocker. But if you want true parity, the overshoot guard belongs
on the state entering the readout, not on the readout's output. Worth a line in
the PR description either way, so the "the two paths now agree" claim isn't
read more strongly than it holds.

### 3. `src/ode/predictions.rs:3414` — adaptive-dosing monitor sign asymmetry *(filed as #1039)*

With this fix, an `ObserveMode::Ipred` monitor sees a signed Form C readout
while an `ObserveMode::Dv` monitor on the same model still gets `.max(0.0)`.
Same model, two monitor modes, opposite sign semantics — a controller
thresholding a change-from-baseline signal silently reads `0` for every negative
sample. The PR flags this as deliberate follow-up, which is the right call for
scope; filed as #1039 so it isn't lost.

PR #303, opened 2026-06-13
chore(ci): Enzyme autodiff CI — toolchain image + AD-vs-FD consistency job [#281]

TeunP:
## Why

ferx-core's **default** Cargo build enables Enzyme autodiff (`default = ["autodiff"]`) — the nominal primary build — yet the autodiff inner-gradient path is exercised **nowhere** in CI: every job builds `--features ci` (finite differences) on plain `nightly`. This is the FD-only coverage gap that let the whole #278 bug class ship undetected, and it means the regression guard #280 added (`tests/autodiff_fd_consistency.rs`, gated `#![cfg(feature = "autodiff")]`) currently runs nowhere automatically. Closes the structural half of #281.

## What changed

Three CI-only files (no product code, no public API):

1. **`.github/docker/enzyme-toolchain.Dockerfile`** — a lean `rustc`+LLVM+Enzyme image, a fork of `../ferx-r/Dockerfile`'s proven `enzyme-builder` stage minus the R/RStudio layers (`llvm.targets=X86` trim for a CI image).
2. **`.github/workflows/enzyme-toolchain-image.yml`** — builds that image **once** on a GitHub-hosted runner (with `jlumbroso/free-disk-space`) and pushes it to `ghcr.io/ferx-nlme/enzyme-toolchain`; manual + monthly + on-recipe-change.
3. **`.github/workflows/autodiff.yml`** — pulls that image and runs `autodiff_fd_consistency` under the fat-LTO `release` profile (Enzyme requires fat LTO for cross-crate correctness).

## Resolved design decisions

- **Image owner = ferx-core.** The image exists to test *ferx-core's* autodiff code; the test and the code under test both live here, so owning the image here keeps this repo's CI self-contained with no cross-repo CI coupling. The recipe tracks ferx-r's `enzyme-builder` stage (documented in the Dockerfile header).
- **Cadence mirrors `slow-tests.yml`:** `schedule` (daily) + `workflow_dispatch` + `push: main` on AD paths. **No per-PR trigger** — the fat-LTO Enzyme compile is heavy; a per-PR run would go red on fork PRs (read-only `GITHUB_TOKEN` can't pull the image until it is public) and add noise before the image is proven. Daily + push-to-main still catches AD regressions within a day and on every main merge touching AD code. Promotion to a required PR-path check is a noted future step once the job is proven stable.

## Bootstrap sequence (after merge — needs org admin)

[...]

PR #1041, opened 2026-08-21
fix(parser): reject unknown [block] names instead of silently ignoring them [closes #1040]

### LINKED ISSUE #1040: Parser silently ignores unknown [block] names — add E_UNKNOWN_BLOCK
Reported downstream while evaluating ferx as a third PopPK engine alongside NONMEM and nlmixr2
(InsightRX/pharma_ai_llm#1100). Copied here as the upstream tracking issue.

## Problem

An unrecognised `[block]` header produces **no diagnostic at all**. `extract_blocks`
(`src/parser/model_parser.rs:10042`) collects every `[name]` header it sees into a
`HashMap`, and the rest of `compile_model` reads blocks by *name lookup* —
`blocks.get("individual_parameters")`, `blocks.get("fit_options")`, … — rather than by an
exhaustive match. A block whose name nothing looks up is simply never read.

So a misspelled optional block is silently dropped:

```
[fit_option]          # note: singular
  method     = focei
  maxiter    = 500
  covariance = true
```

`ferx check` reports `valid: true`. The fit then runs with the *default* method and
iteration cap and **no covariance step** — it returns without standard errors, and
nothing says why. Same shape for `[scaling]`, `[output]`, `[derived]`, `[covariates]`,
`[simulation]`, …: each silently drops the behaviour it was meant to configure.

Unknown *keys* inside a known block are already hard errors that enumerate the valid set:

```
[event_model]: unknown key `foo` — valid keys: cmt, family, scale, rate,
               shape, alpha, gamma, loghr, hazard, type, clock
```

Keys are closed-world; block names are open-world. This is the one failure mode in the
validation surface that is **silent rather than loud**, and it yields a plausible-looking
fit built on ignored input.

Why it matters more than a typical typo class: ferx's grammar is new enough that LLM
authoring tools have essentially no training data for it, so invented and misspelled
block names are the *highest-probability* authoring error, not an edge case.

## Ask

New registry entry, following the existing `docs/file-formats/check-report.qmd` convention:

| Code | Severity | Meaning |
|---|---|---|
| `E_UNKNOWN_BLOCK` | error | A `[block]` header is not a recognised block name. |

- Message in the `[event_model]` style — offending name plus the valid set, plus a
  did-you-mean when there is a near match.
- `block` / `line` populated from the header (`ExtractedBlocks::block_lines` already
  records the 1-based header line, so this is the granularity we already have).
- Emitted at **parse time**, so it reaches both `ferx check` (batch, via
  `parse_error_to_diagnostic`) and `fit()` (via `first_error`) with no separate path.
- Single source of truth for the block registry, exported `pub` so `ferx-r` and
  `ferxtranslate` can read it from the engine instead of duplicating it.

**Severity: `error`, not `warning`** — a silently-dropped block changes the numerical
result, and warnings keep `valid: true` per the check-report doc. The counter-argument is
forward compatibility (an older binary reading a file that uses a newer block would
hard-fail), but a version gate is the better answer there than ignoring unknown names.

Blocks the parser reads today (the registry to close over):

`parameters`, `structural_model`, `error_model`, `individual_parameters`, `odes`,
`scaling`, `initial_conditions`, `derived`, `output`, `covariates`, `fit_options`,
`data`, `data_selection`, `simulation`, `adaptive_dosing`, `diffusion`, `mixture`,
`event_model`, `binary_model`, `markov_model`, `covariate_nn`.

Note the feature-gated ones (`event_model` / `binary_model` under `survival`,
`markov_model` under `markov`, `covariate_nn` under `nn`) must stay **known** in a
default build — otherwise a valid model would be rejected as unknown by a binary built
without the feature. `[covariate_nn]` already errors with `E_NN_FEATURE_DISABLED` in that
case; the survival/markov blocks are currently silently ignored, which is the same
silent-drop class and worth closing in the same pass.

## Paired `ferx-r` fix (follow-up, tracked separately)

`ferx_model_validate()` computes an `unknown` section list and prints `[unknown section]`,
but leaves it out of the returned status:

```r
ok <- isTRUE(rust_result$ok) && length(missing) == 0L   # `unknown` not included
```

so `res$ok` is `TRUE` for a model with an unrecognised block. That R-side list is also a
duplicate of the Rust registry and **has already drifted** — `optional_sections`
(`R/ferx_model_validate.R:80`) omits `covariates`, `event_model`, `binary_model`,
`markov_model`, `data_selection`, `adaptive_dosing`, and `simulation`, all of which the
Rust parser supports and all of which appear in ferx's own bundled examples. Today
`ferx_model_validate(ferx_example("two_cpt_oral_cov")$model)` prints
`covariates [unknown section]`.

Fix: fold `unknown` into `ok`, and source the block registry from the engine.

## Not in scope (noted so they are not re-derived)

- **Block-level line numbers.** `line` points at the `[block]` header, not the offending
  token — already documented as a deferred enhancement.
- **Inconsistent enumeration.** ~15 of 73 `unknown`/`invalid` errors list the valid
  alternatives. `[event_model]` is the good pattern; `[derived] integral(): unknown
  keyword argument` is the clearest gap.

roninsightrx:
## Why

Reported downstream (InsightRX/pharma_ai_llm#1100) while evaluating ferx as a third PopPK
engine alongside NONMEM and nlmixr2, and copied here as #1040.

An unrecognised `[block]` header produced **no diagnostic at all**. `extract_blocks` collects
every `[name]` header it sees into a map, and the rest of `compile_model` reads blocks by *name
lookup* — `blocks.get("individual_parameters")`, `blocks.get("fit_options")`, … — never by an
exhaustive match. A block whose name nothing looks up is simply never read.

So a misspelled optional block was silently dropped:

```
[fit_option]          # note: singular
  method     = focei
  maxiter    = 500
  covariance = true
```

`ferx check` reported `valid: true`. The fit then ran with the *default* method and iteration
cap and **no covariance step** — returning without standard errors, and nothing said why. Same
shape for `[scalings]`, `[outputs]`, `[derived]`, `[covariates]`, `[simulation]`: each silently
dropped the behaviour it was meant to configure.

Unknown *keys* inside a known block were already hard errors that enumerate the valid set
(`[event_model]: unknown key 'foo' — valid keys: cmt, family, scale, …`). Keys were
closed-world; block names were open-world. This was the one failure mode in the validation
surface that is silent rather than loud, and the one that yields a plausible-looking fit built
on ignored input.

It matters more than a typical typo class because the `.ferx` grammar is new enough that LLM
authoring tools have essentially no training data for it, so invented and misspelled block
names are the highest-probability authoring error, not an edge case.

## What changed

`BLOCK_REGISTRY` in `parser/model_parser.rs` is now the closed-world list of block names, each
with the header form it accepts (`[name]`, `[name INSTANCE]`, or either) and the cargo feature
it needs. `extract_blocks` records every header it sees as `(type, instance, line)` and
validates the lot in one pass before returning, so three silent-drop shapes are now errors:

| Code | Meaning |
|---|---|
| `E_UNKNOWN_BLOCK` | header is not a recognised block name |
| `E_DEPRECATED_BLOCK` | header *was* ferx syntax and is no longer read — currently `[initial_values]` |
| `E_BLOCK_INSTANCE_NAME` | instance name given where none is taken (`[fit_options DOSE]`), or omitted where one is required (`[covariate_nn]`) |
| `E_BLOCK_FEATURE_DISABLED` | recognised block needs a feature this binary lacks — `[event_model]` / `[binary_model]` without `--features survival`, `[markov_model]` without `--features markov` |

The message follows the `[event_model]` convention — every offender, the valid set, and a
did-you-mean for a near match (prefix relation or edit distance ≤ 2):

```
$ ferx check typo.ferx
error[E_UNKNOWN_BLOCK] fit_option:35: Unknown block `[fit_option]` — did you mean
  `[fit_options]`. Valid blocks: adaptive_dosing, covariate_nn, covariates, data, data_selection,
  derived, diffusion, error_model, fit_options, individual_parameters, initial_conditions,
  mixture, odes, output, parameters, scaling, simulation, structural_model.
    help: did you mean `[fit_options]`?
invalid: typo — 1 error(s), 0 warning(s)
```

`[initial_values]` is the one deprecated name that matters in practice. It was ferx's own
spelling for initial estimates before they moved inline into `[parameters]`; the parser stopped
reading it and, because unknown names were dropped in silence, never said so. Three of ferx-r's
checked-in example models still carry one, and no valid name is within suggesting distance
(`initial_conditions` is well past edit distance 2), so it gets a remediation instead of a bare
"unknown block":

```
error[E_DEPRECATED_BLOCK] initial_values:43: Deprecated block `[initial_values]` is no longer
  read: initial estimates are declared inline in `[parameters]` (`theta NAME(init, lower,
  upper)`, `omega NAME ~ variance`); the block has not been read for several releases — delete it.
```

Emitted at parse time, so it reaches `ferx check` (batch, via `parse_error_to_diagnostic`) and
`fit()` (via `first_error`) with no separate path. Since a parse failure is terminal — there is
no `ParsedModel` to read `block_lines` from — the check report recovers `block` and `line` from
the message text; the mapping is pinned by unit tests on both sides. When the message names
exactly one header the line is lifted into the structured `line` field and removed from the
message, since the renderer already prints `block:line` ahead of it; with several headers each
keeps its own parenthetical and no `line` is attached, because promoting the first would
mislocate the rest.

The registry is exported as `known_block_names()` (re-exported at the crate root) so `ferx-r`
and `ferxtranslate` can read the list from the engine instead of keeping their own copy. It is
filtered by the features the binary carries, so a default build does not advertise
`[event_model]` and then refuse it — the registry still *recognises* the name in every build,
which is what keeps that refusal a specific "build with `--features survival`" rather than a bare
unknown. The duplicated R-side list in `ferx_model_validate.R` has already drifted — its
`optional_sections` omits `covariates`, `event_model`, `binary_model`, `markov_model`,
`data_selection`, `adaptive_dosing` and `simulation`, so
`ferx_model_validate(ferx_example("two_cpt_oral_cov")$model)` prints `covariates [unknown section]`
today.

Two existing tests needed updating, both because they relied on the old permissiveness:

- `test_extract_blocks_recognizes_named_block_form` used an invented `[some_named_block FOO]`;
  it now uses `[covariate_nn FOO]`, a real named-only block.
- `test_parse_all_example_ferx_files` skipped a feature-gated example only when it had no
  `[structural_model]`. That nuance existed because a joint PK + CTMM file like
  `examples/ctmm_pd_2state.ferx` used to parse in a default build — as a *PK-only model*, with
  the endpoint silently discarded. The skip is now simply "this build lacks the feature the file
  needs", and the rejection itself is pinned by
  `test_feature_gated_block_rejected_without_its_feature`.

## Alternatives considered

**`warning` rather than `error`.** Rejected: a silently-dropped block changes the numerical
result, and warnings keep `valid: true` per the check-report doc, which is exactly the state
that was misleading. The counter-argument is forward compatibility — an older binary reading a
file that uses a newer block now hard-fails — but a version gate is the better answer there
than ignoring unknown names.

**Leaving the feature-gated blocks alone.** They are *known* names, so `E_UNKNOWN_BLOCK` would
not have covered them, and they fail in exactly the same way: a default build reading a
`[markov_model]` file quietly produced a model with no CTMM endpoint. Same class, same pass.

**A closed-world check in each consumer** (the shape the downstream report first considered).
Rejected there and here: it puts the check at the wrong layer and means maintaining another
copy of a registry that has already drifted once.

## Cross-repo dependency
| ferx (R) | FeRx-NLME/ferx-r#303 | open (draft) | after |

The paired `ferx-r` fix is FeRx-NLME/ferx-r#303, held as a draft until this merges and its
`src/rust/Cargo.lock` is bumped to a commit carrying `known_block_names()`. There,
`ferx_model_validate()` computed an `unknown` section list and printed `[unknown section]` but
left it out of the returned status (`ok <- isTRUE(rust_result$ok) && length(missing) == 0L`), so
`res$ok` was `TRUE` for a model with an unrecognised block; #303 folds `unknown` into `ok` and
takes the section list from `known_block_names()` instead of its own drifted copy.

Nothing in this PR breaks the current ferx-r build — `known_block_names` is purely additive, and
ferx-r's own bundled examples use only registry names. The one file class that does break is a
`.ferx` carrying `[initial_values]`, a block ferx-core has never read: #303 strips it from the
three legacy files under ferx-r's `examples/models/`.

## Breaking changes
- [ ] Public Rust API
- [ ] None

<details>

A model file that parsed before and carried an unrecognised `[block]` header now fails to
parse. In every such case the block was being ignored, so the fix is either to correct the name
(the error names the near match) or to delete the block. Nothing that was actually being read
changes behaviour.

The one case worth calling out: a file with `[event_model]` / `[binary_model]` /
`[markov_model]` built **without** the matching cargo feature used to parse as a Gaussian-only
model with the endpoint discarded. It now errors, naming the feature to build with.

</details>

## Numerical validation
- [x] Not applicable — parse-time validation only. No estimation, PK, or stats path is touched;
  every model that parsed *and was fully read* before parses identically now (`cargo test --lib`
  green on default, `--features survival`, `--features markov`, `--features nn`).

## Performance
- [x] No significant impact expected — one pass over the block headers already collected, at
  parse time.

## Tests

Tier 1, `src/parser/model_parser_tests.rs`: unknown name rejected with line + valid set +
did-you-mean; all offenders reported together; no-near-match case carries no hint; repeated
header reported once; uppercase `[FIT_OPTIONS]` still valid; instance name on an unnamed-only
block; named-only block without one; feature-gated block rejected without / accepted with its
feature; `nearest_block_name` thresholds; registry sorted and unique.

Tier 1, `src/api/tests/block_name_diagnostic_tests.rs` (new sibling): the
`parse_error_to_diagnostic` mapping — code, `block`, `line` and `suggestion` for each shape,
`[covariate_nn]` keeping its dedicated `E_NN_FEATURE_DISABLED`, and everything else still
falling through to `E_PARSE`.

Tier 2, `tests/check_command.rs`: `ferx check` end-to-end on a `[fit_option]` typo, and the
same rejection reaching the parser directly.

<details>
<summary>Example</summary>

```toml
[parameters]
  theta TVCL(0.2, 0.001, 10.0)
  theta TVV(10.0, 0.1, 500.0)
  sigma PROP_ERR ~ 0.02

[individual_parameters]
  CL = TVCL
  V  = TVV

[structural_model]
  pk one_cpt_iv(cl=CL, v=V)

[error_model]
  DV ~ proportional(PROP_ERR)

[fit_option]            # typo: singular
  method = focei
```

```
error[E_UNKNOWN_BLOCK] fit_option:16: Unknown block `[fit_option]` (line 16) — did you mean
  `[fit_options]`. Valid blocks: adaptive_dosing, binary_model, covariate_nn, covariates, data,
  data_selection, derived, diffusion, error_model, event_model, fit_options,
  individual_parameters, initial_conditions, markov_model, mixture, odes, output, parameters,
  scaling, simulation, structural_model.
    help: did you mean `[fit_options]`?
invalid: model — 1 error(s), 0 warning(s)
```

Before this PR: `ok: model — no errors (0 warning(s))`, and the fit ran with the default method
and no covariance step.

</details>

## Docs
- [x] `docs/` updated for user-visible changes — `docs/file-formats/check-report.qmd` gains the
  four codes; `docs/model-file/index.qmd`'s block overview is completed to the full registry
  (it listed 8 of 21) with the closed-world rule and the `[initial_values]` retirement spelled
  out; `docs/model-file/neural-networks.qmd` flags `[dynamics_nn]` as not accepted yet — it is
  shown there as syntax, and with closed-world names such a file is now rejected outright rather
  than parsed with the block ignored.
- [x] No new pages, so `_quarto.yml` is unchanged

## Checklist
- [x] `cargo clippy` clean (no new warnings; the pre-existing ones in `sir.rs` / `validation.rs`
  are untouched)
- [x] No `Dual2` / sensitivity code touched

## Docs & examples
- [x] No new DSL syntax, `[fit_options]` key, example file, or data file — nothing to mirror to
  ferx-site or ferx-book.
- [x] No example execution needed: no example changed, and all of `examples/` still parses
  (`test_parse_all_example_ferx_files`, on default, `survival` and `markov` builds).

## Reviewer hints

Focus on `BLOCK_REGISTRY` — is anything missing or mis-formed? The list was derived by grepping
every `blocks.get(…)` / `extracted.named.get(…)` in `model_parser.rs`; if a block is read
somewhere I missed, it is now rejected as unknown. `test_parse_all_example_ferx_files` covers
everything the bundled examples use, which is all 21 minus `data` and `mixture`.

Second: the `BlockForm` calls. `covariate_nn` is `Named` (only read out of the `named` map),
`event_model` / `binary_model` / `markov_model` are `Either` (both maps are read), everything
else is `Unnamed`.

Third: recovering `block` / `line` from the message text in `parse_error_to_diagnostic` is
unlovely but is what the terminal-parse-error path forces. Both ends are pinned by tests, so a
reworded message fails loudly rather than silently downgrading to `E_PARSE`.

## Open questions

Forward compatibility: an older binary now hard-fails on a file using a block a newer version
added. The issue's position — and mine — is that a version gate is the right answer to that,
not tolerating unknown names, but if we want the gate first this PR should wait for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,19 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Changed
+- **An unrecognised `[block]` name is now an error (#1040).** Blocks were read by name lookup, so a
+  header the parser did not know was never read and never reported: a misspelled `[fit_option]` left
+  `ferx check` saying `valid: true` while the fit ran with the default method, the default iteration
+  cap and **no covariance step** — returning without standard errors and no indication why. The same
+  went for `[scalings]`, `[outputs]`, `[derived]`, `[covariates]` and friends. Block names are now
+  closed-world, like the keys inside a block already were: an unknown header is `E_UNKNOWN_BLOCK`,
+  listing every offender with its line, the full valid set, and a did-you-mean for a near match. Two
+  neighbouring silent drops go with it — an instance name where none is taken (`[fit_options DOSE]`)
+  or missing where one is required (`[covariate_nn]`) is `E_BLOCK_INSTANCE_NAME`, and a block whose
+  cargo feature this binary lacks (`[event_model]` without `--features survival`, `[markov_model]`
+  without `--features markov`) is `E_BLOCK_FEATURE_DISABLED` instead of being parsed away. The
+  recognised block names are exported as `known_block_names()` so wrappers can read the list from the
+  engine rather than keeping their own copy.
 - **A dose attribute that is also read by the model is now an error (#993).** `F`,
   `LAGTIME`/`ALAG` and the compartment-indexed `F{n}`/`ALAG{n}`/`LAGTIME{n}` are applied by the
   engine **at the dose event**. A model that declares one and *also* references it in `[odes]`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,9 +30,13 @@ section of the SDLC for the versioning policy).
   neighbouring silent drops go with it — an instance name where none is taken (`[fit_options DOSE]`)
   or missing where one is required (`[covariate_nn]`) is `E_BLOCK_INSTANCE_NAME`, and a block whose
   cargo feature this binary lacks (`[event_model]` without `--features survival`, `[markov_model]`
-  without `--features markov`) is `E_BLOCK_FEATURE_DISABLED` instead of being parsed away. The
-  recognised block names are exported as `known_block_names()` so wrappers can read the list from the
-  engine rather than keeping their own copy.
+  without `--features markov`) is `E_BLOCK_FEATURE_DISABLED` instead of being parsed away, and
+  `[initial_values]` — ferx's own former spelling for initial estimates, unread since they moved
+  inline into `[parameters]` — is `E_DEPRECATED_BLOCK`, naming the replacement rather than offering
+  a did-you-mean that does not exist. The recognised block names are exported as
+  `known_block_names()` so wrappers can read the list from the engine rather than keeping their own
+  copy; it reflects the features the binary was built with, so it never advertises a name the same
+  binary would refuse.
 - **A dose attribute that is also read by the model is now an error (#993).** `F`,
   `LAGTIME`/`ALAG` and the compartment-indexed `F{n}`/`ALAG{n}`/`LAGTIME{n}` are applied by the
   engine **at the dose event**. A model that declares one and *also* references it in `[odes]`
--- a/docs/file-formats/check-report.qmd
+++ b/docs/file-formats/check-report.qmd
@@ -54,6 +54,9 @@ Optional fields are omitted entirely when absent (not emitted as `null`).
 |------|----------|---------|
 | `E_PARSE` | error | The model file failed to parse (catch-all for parser errors). |
 | `E_MISSING_BLOCK` | error | A required `[block]` is absent. `block` names which one. |
+| `E_UNKNOWN_BLOCK` | error | A `[block]` header is not a recognised block name. The message lists every offending header with its line, the full valid set, and a did-you-mean when there is a near match; `suggestion` carries the near match on its own. See [#1040](https://github.com/FeRx-NLME/ferx-core/issues/1040). |
+| `E_BLOCK_INSTANCE_NAME` | error | A `[block NAME]` instance name is given for a block that takes none (e.g. `[fit_options DOSE]`), or omitted for one that requires it (`[covariate_nn NAME]`). |
+| `E_BLOCK_FEATURE_DISABLED` | error | A recognised block needs a cargo feature this binary was not built with — `[event_model]` / `[binary_model]` need `--features survival`, `[markov_model]` needs `--features markov`. (`[covariate_nn]` has its own `E_NN_FEATURE_DISABLED`.) |
 | `E_NN_FEATURE_DISABLED` | error | A `[covariate_nn]` block requires building with `--features nn`. |
 | `E_MISSING_COVARIATE` | error | The model references a covariate not present in the data (case-sensitive). |
 | `E_ETA_NOT_DECLARED` | error | An unresolved identifier reads as a random effect (`ETA…` / `KAPPA…`) but no `omega` / `kappa` declaration defines it, and the data carries no such column. Usually a deleted `omega` line whose `exp(ETA_…)` term was left behind. |
--- a/docs/file-formats/check-report.qmd
+++ b/docs/file-formats/check-report.qmd
@@ -54,7 +54,8 @@ Optional fields are omitted entirely when absent (not emitted as `null`).
 |------|----------|---------|
 | `E_PARSE` | error | The model file failed to parse (catch-all for parser errors). |
 | `E_MISSING_BLOCK` | error | A required `[block]` is absent. `block` names which one. |
-| `E_UNKNOWN_BLOCK` | error | A `[block]` header is not a recognised block name. The message lists every offending header with its line, the full valid set, and a did-you-mean when there is a near match; `suggestion` carries the near match on its own. See [#1040](https://github.com/FeRx-NLME/ferx-core/issues/1040). |
+| `E_UNKNOWN_BLOCK` | error | A `[block]` header is not a recognised block name. The message lists every offending header with its line, the valid set **this build accepts**, and a did-you-mean when there is a near match; `suggestion` carries the near match on its own. See [#1040](https://github.com/FeRx-NLME/ferx-core/issues/1040). |
+| `E_DEPRECATED_BLOCK` | error | A `[block]` that *was* ferx syntax and is no longer read — currently `[initial_values]`, whose contents moved inline into `[parameters]`. Reported separately from `E_UNKNOWN_BLOCK` so the message can name the replacement instead of offering a did-you-mean that does not exist. |
 | `E_BLOCK_INSTANCE_NAME` | error | A `[block NAME]` instance name is given for a block that takes none (e.g. `[fit_options DOSE]`), or omitted for one that requires it (`[covariate_nn NAME]`). |
 | `E_BLOCK_FEATURE_DISABLED` | error | A recognised block needs a cargo feature this binary was not built with — `[event_model]` / `[binary_model]` need `--features survival`, `[markov_model]` needs `--features markov`. (`[covariate_nn]` has its own `E_NN_FEATURE_DISABLED`.) |
 | `E_NN_FEATURE_DISABLED` | error | A `[covariate_nn]` block requires building with `--features nn`. |
--- a/docs/model-file/index.qmd
+++ b/docs/model-file/index.qmd
@@ -37,7 +37,15 @@ covariance step while `ferx check` still reported `valid: true`.
 The same applies to the instance-name form: `[covariate_nn NAME]` requires a
 name, and every other block refuses one (`E_BLOCK_INSTANCE_NAME`). A block whose
 cargo feature this binary lacks is refused too, rather than quietly dropped
-(`E_BLOCK_FEATURE_DISABLED`).
+(`E_BLOCK_FEATURE_DISABLED`); the valid-set the error message enumerates is
+likewise the set *this* build accepts, so a default build does not advertise
+`[event_model]`.
+
+`[initial_values]` is **not** a block. Initial estimates are declared inline in
+`[parameters]` (`theta NAME(init, lower, upper)`, `omega NAME ~ variance`); the
+separate block predates that and has not been read for several releases. A file
+still carrying one is now rejected with `E_DEPRECATED_BLOCK` naming the
+replacement — delete the block.
 
 ## Minimal Example
 
--- a/docs/model-file/index.qmd
+++ b/docs/model-file/index.qmd
@@ -11,9 +11,33 @@ ferx-core models are defined in `.ferx` files using a declarative DSL. Each file
 | `[structural_model]` | Yes | Specify the PK model (analytical or ODE) |
 | `[error_model]` | Yes | Define the residual error model |
 | `[odes]` | If ODE | ODE right-hand-side equations |
+| `[adaptive_dosing]` | No | Feedback / TDM dose controller for `simulate()` |
+| `[covariates]` | No | Declare which data columns are covariates (and how they interpolate) |
+| `[covariate_nn NAME]` | No | Neural-network covariate model (needs `--features nn`) |
 | `[data]` | No | Point at the dataset CSV (`$DATA` equivalent); overridden by an explicit CLI/R path |
+| `[data_selection]` | No | Row/subject filters applied to the dataset |
+| `[derived]` | No | Post-hoc derived quantities written to sdtab |
+| `[diffusion]` | No | SDE diffusion terms |
 | `[fit_options]` | No | Configure estimation method and optimizer |
+| `[initial_conditions]` | No | Non-zero compartment initial amounts |
+| `[mixture]` | No | Mixture (subpopulation) model |
+| `[output]` | No | Extra sdtab output columns |
+| `[scaling]` | No | Compartment scaling / observation readouts |
 | `[simulation]` | No | Define a simulation trial design |
+| `[binary_model]` | No | Binary / logistic endpoint (needs `--features survival`) |
+| `[event_model]` | No | Time-to-event endpoint (needs `--features survival`) |
+| `[markov_model]` | No | Continuous-time Markov endpoint (needs `--features markov`) |
+
+Block names are **closed-world**: a header that is not in this table is an error
+(`E_UNKNOWN_BLOCK`), reported with its line number and a did-you-mean when there
+is a near match. A misspelled optional block used to be ignored in silence — a
+`[fit_option]` typo left the fit running with the default method and no
+covariance step while `ferx check` still reported `valid: true`.
+
+The same applies to the instance-name form: `[covariate_nn NAME]` requires a
+name, and every other block refuses one (`E_BLOCK_INSTANCE_NAME`). A block whose
+cargo feature this binary lacks is refused too, rather than quietly dropped
+(`E_BLOCK_FEATURE_DISABLED`).
 
 ## Minimal Example
 
--- a/docs/model-file/neural-networks.qmd
+++ b/docs/model-file/neural-networks.qmd
@@ -69,6 +69,16 @@ DOI [10.1002/psp4.12808](https://doi.org/10.1002/psp4.12808).
 
 ### `[dynamics_nn]` — Low-dimensional Neural ODEs
 
+::: {.callout-important}
+## Not accepted yet — design sketch only
+
+`[dynamics_nn]` is **Phase B** and the parser does not read it. Since block
+names became closed-world, a file containing one is rejected outright with
+`E_UNKNOWN_BLOCK` rather than parsed with the block ignored (see the
+[check report](../file-formats/check-report.qmd)). The syntax below shows the
+intended shape; it is not usable in this release.
+:::
+
 A small neural network appears on the **right-hand side of an ODE**,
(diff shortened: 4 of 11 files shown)


roninsightrx:
## Code review — PR #1041 (medium effort)

Reviewed on a throwaway worktree of `fix-1040-unknown-block` (10 files); `cargo check --lib` clean on the branch.

**Registry completeness check** (the PR's own reviewer hint): I cross-checked `BLOCK_REGISTRY` against every `blocks.get(…)` / `extracted.named.get(…)` in `model_parser.rs`, against every `[header]` in `examples/**.ferx`, `docs/**`, and `tests/`, and against the feature gates in `Cargo.toml`. The registry is complete and correctly formed for this repo — with the one exception below. The header forms (`Named` for `covariate_nn`, `Either` for the three endpoint blocks) match how the maps are read. `frem` uses its own `parse_blocks` and is unaffected.

### Findings

**medium — `src/parser/model_parser.rs:10058`: `[initial_values]` is missing from the registry**

`[initial_values]` is a deprecated ferx block that the parser silently ignored *on purpose*, and it is still live downstream:

- `ferx-r/R/ferx_model_validate.R:80` lists it in `optional_sections`
- `ferx-r/CLAUDE.md:66` documents it as a section
- older `ferx_model()` templates emitted it
- three checked-in models still carry it: `ferx-r/examples/models/{warfarin,mm_oral,two_cpt_iv}.ferx`

After this PR every such file hard-fails. And `nearest_block_name("initial_values")` returns `None` (edit distance to `initial_conditions` is well past 2, and there's no prefix relation), so the user gets a bare "Unknown block" with no did-you-mean — for a block that was ferx's own syntax. Worth an explicit deprecated entry that either errors with "removed in vX; delete the block" or is accepted-and-ignored, rather than falling into the generic unknown bucket.

**low — `src/api/validation.rs:2953`: `E_BLOCK_FEATURE_DISABLED` is an unguarded fallback**

The ``Block `[`` branch is an if/else with `E_BLOCK_FEATURE_DISABLED` as the *else*: any error string starting with ``Block `[`` that lacks the substring `"instance name"` gets labelled feature-disabled. The function's own doc says every shape is matched on a sentinel pinned by a test; this one isn't. Suggest matching on the actual sentinel (``requires building ferx-core with `--features``) and falling through to `E_PARSE` otherwise, so a future ``Block `[…]`` message can't be mis-coded.

**low — `src/parser/model_parser.rs:10167`: unknown-block message drops the instance name**

`[foo BAR]` is reported as ``Unknown block `[foo]` (line N)`` — text the user cannot grep for in their own file. Worse, a file containing both `[foo]` and `[foo BAR]` produces two entries that both read ``[foo]``, distinguishable only by line number. Include the instance name when present.

**low — `src/api/validation.rs:2971`: line number is double-printed**

`attach` sets both `block` and `line` while the message already embeds them, so the rendered check report reads:

```
fit_option:16: Unknown block `[fit_option]` (line 16) …
```

(as shown in the PR body). That's exactly the double-print the neighbouring #993 branch comment explicitly avoids by *not* calling `with_block`. Either drop `(line N)` from the parser message, or don't attach.

**low — `src/parser/model_parser.rs:10089`: `known_block_names()` is build-independent**

It lists `event_model` / `binary_model` / `markov_model` even in a default build. So the "Valid blocks:" enumeration advertises names this binary will then reject with `E_BLOCK_FEATURE_DISABLED` — and `ferx-r`, told to source its list from here, will call `[markov_model]` valid against a core that refuses it. Consider filtering by `block_feature_enabled`, or exposing the feature name alongside the block name.

**low — `docs/model-file/neural-networks.qmd:77`: documented block no longer parses**

The page shows a `[dynamics_nn ka_nn]` block as syntax. With block names now closed-world, that file fails with `E_UNKNOWN_BLOCK` and no did-you-mean. Since this PR is what turns it into a hard error, the page should say the block isn't accepted yet — right now that's only implied by a "Pending (Phase B)" cell in a later table.

### No issues found in

`edit_distance`, the `check_block_names` dedup, the header-form matching, or the `block_and_line` message parsing. The `covariate_nn` / `E_NN_FEATURE_DISABLED` precedence is preserved correctly.

roninsightrx:
All six addressed in fb9c21ea. Green on default, `--features survival`, `--features markov`,
`--features nn`; `cargo test --tests` clean; no new clippy warnings.

**medium — `[initial_values]` missing from the registry.** Right, and worse than the review
suggests: it is not merely deprecated, it has *never* been read by ferx-core — `grep -rn
initial_values src/` on `main` returns nothing, so every file carrying one has been silently
dropping it since inline inits landed.

Added `DEPRECATED_BLOCKS`, a table separate from `BLOCK_REGISTRY` (a deprecated name is neither
valid — it must not appear in the "valid blocks" enumeration — nor unknown), checked ahead of the
unknown bucket and reported as a new `E_DEPRECATED_BLOCK`:

```
error[E_DEPRECATED_BLOCK] initial_values:43: Deprecated block `[initial_values]` is no longer
  read: initial estimates are declared inline in `[parameters]` (`theta NAME(init, lower,
  upper)`, `omega NAME ~ variance`); the block has not been read for several releases — delete it.
```

Chose error-with-remediation over accept-and-ignore: accepting it silently is the exact failure
mode this PR exists to close, and the remediation is mechanical enough that a consumer can act on
the code. The three ferx-r example models are fixed in FeRx-NLME/ferx-r#303, which also drops the
`initial_values` entry from `optional_sections` and the section list in its docs.

**low — `E_BLOCK_FEATURE_DISABLED` unguarded fallback.** Fixed. It now matches on
`requires building ferx-core with \`--features` and anything else under the ``Block `[`` prefix
falls through to `E_PARSE`, with a test (`other_block_prefixed_errors_fall_through_to_e_parse`)
pinning that.

**low — unknown-block message drops the instance name.** Fixed both halves: the header is echoed
as written (`` `[foo BAR]` ``), and the dedup key is now `(type, instance)` rather than
`(type, is_named)`, so `[foo BAR]` and `[foo BAZ]` are two findings instead of one entry that
names only the first.

**low — line number double-printed.** Fixed, keeping both consumers whole. The line is lifted
into the structured `line` field **and removed from the message** when the message names exactly
one header — the unambiguous case, which is what the PR body showed. With several headers each
keeps its own parenthetical and no `line` is attached at all: promoting the first would have
pointed the structured location at one finding while the message described several. So:

```
error[E_UNKNOWN_BLOCK] fit_option:35: Unknown block `[fit_option]` — did you mean `[fit_options]`. …
```

The alternative (drop the attach, per the #993 precedent) was rejected because the structured
`block`/`line` is the thing the originating issue asked for — it is what lets an agent act on the
report without matching prose.

**low — `known_block_names()` is build-independent.** Fixed: it now filters on
`block_feature_enabled`, so a default build neither advertises `[event_model]` nor tells ferx-r
that `[markov_model]` is valid. The registry still *recognises* those names in every build, which
is what keeps the rejection a specific `E_BLOCK_FEATURE_DISABLED` naming the feature rather than
a bare unknown. Pinned by `test_known_block_names_track_the_build_features`, which asserts
membership `== cfg!(feature = "…")` in both directions. Doc'd on the function and in
check-report.qmd, since the result is now build-dependent and a consumer has to read it from the
same binary it parses with.

**low — `docs/model-file/neural-networks.qmd` documented block no longer parses.** Fixed with a
callout at the top of the `[dynamics_nn]` section saying it is Phase B, the parser does not read
it, and a file containing one is now rejected with `E_UNKNOWN_BLOCK`. The other two `dynamics_nn`
mentions (covariate-nn.qmd's roadmap line, the `ExtractedBlocks` doc comment) are prose about a
future block, not syntax, so they stand.

Coverage: added tests for the `block_feature_enabled` unknown-feature arm and for a malformed
`(line …)` parenthetical, which were the two lines codecov flagged.

roninsightrx:
Correction to my reply above: I said `[initial_values]` had "never" been read. Wrong — it *was* read, and was removed from the engine in e5e934d; my grep only proved it is absent from current `main`. ferx-r's own NEWS records the removal, and notes that the parser has silently ignored leftover blocks ever since. That silent-ignore window is why three of its example models still carry one, so the finding and the fix stand; only my history was off. The committed text (`DEPRECATED_BLOCKS`, check-report.qmd, model-file/index.qmd) says "has not been read for several releases" and was always correct.

Issue #540, opened 2026-06-26
Analytic FOCE/FOCEI gradients for Form C readouts referencing covariates


roninsightrx:
## Summary

Form C ODE readouts (`[scaling] y = <expr>`) that reference a **covariate** currently fall back to the finite-difference (FD) gradient, even under `gradient = auto`. They should get the exact analytic `Dual2`/`Dual1` gradient like simple Form C readouts (`y = central/V1`) do since #410/#439.

## Motivating model

`fluconazole_radboudumc` (2-cpt IV with saturable protein binding; translation of NONMEM `fluconfinal014.mod`). Its readout converts free → total concentration and references the `FREE` assay-type covariate:

```
y = central / V1 + (1.0 - FREE) * BMAX * ALBBMAX * (central / V1) / (KD + central / V1)
```

`central` is a state; `V1`, `BMAX`, `ALBBMAX`, `KD` are individual parameters; **`FREE` is a covariate**. The model is forced to `gradient = fd` today purely because of that covariate reference.

## Root cause

`ode_analytical_supported` (`src/sens/ode_provider.rs:100`) admits a `Single` Form C readout only when `OdeOutputProgram::is_simple()` holds, and `is_simple` (`src/parser/model_parser.rs:4620`) is false if the bytecode contains **any** `PushTheta`/`PushEta`/`PushCov`/`PushNnOutput`. The dual evaluator `eval_output_g` (`model_parser.rs:9895`) passes **empty** θ/η/cov slices to the bytecode VM, so a `PushCov` would read `0.0` — hence covariates must be excluded from the dual path.

## Fix (this issue: covariates)

In the ODE sensitivity provider the state is integrated as `Dual2<N>` seeded on the **individual-parameter** axes; a covariate is a **constant** in that basis (it carries no parameter derivative — `PushCov` lifts an f64 to `T::from_f64`). So a covariate in the readout needs only its *value* threaded through, with no change to the η/θ chain or the FOCEI Hessian:

1. Carry `cov_names` on `OdeOutputProgram` and have `eval_output_g` build the covariate slice and pass it to `eval_bytecode_g` (mirrors the f64 `out_fn` and the existing `ScaleDerivProgram` for `obs_scale = <expr>`, which already differentiates over covariates).
[...]

PR #1042, opened 2026-08-21
fix(parser): resolve TIME in a [scaling] Form C readout, and stop undefined names in [scaling] evaluating to 0 [closes #1028]

### LINKED ISSUE #1028: [scaling]: undefined identifiers silently evaluate to 0 (and TIME is not resolvable there)

## Summary

An identifier in `[scaling]` that resolves to nothing is accepted and silently evaluates to **0**. The model validates, fits, converges and reports plausible parameters — for a structural model that is not the one written.

This is the same class as #994 (`init()` accepts `TIME`, always zero, silently), applied to the one expression block that still has no undefined-reference guard.

## Reprex

```
model undef
[parameters]
  theta TVA(1.0, 0.1, 10.0)
  omega ETA_A ~ 0.01
  sigma ADD_ERR ~ 1.0 (sd)
[individual_parameters]
  A = TVA * exp(ETA_A)
[structural_model]
  ode(states=[clock])
[odes]
  d/dt(clock) = 1
[scaling]
  y[CMT=1] = A * TOTALLY_UNDEFINED_NAME
[error_model]
  DV ~ additive(ADD_ERR)
[fit_options]
  method = focei
  gradient = fd
  covariance = false
```

```r
ferx_model_validate("undef.ferx")$ok
#> TRUE

ferx_predict("undef.ferx", "any.csv")
#>   ID TIME PRED
#> 1  1    0    0
#> 2  1   10    0
#> 3  1   20    0
```

## The case that actually bites

`TIME` is available in `[odes]` right-hand sides (see the shipped `warfarin_ode_time.ferx`) but **not** in `[scaling]`. So the natural way to write a response-versus-time readout —

```
[scaling]
  y[CMT=1] = EMAX * TIME / (TIME + T50)
```

— makes the entire structural prediction identically zero, with no diagnostic. I hit this writing a model-based meta-analysis, concluded the opposite of the truth from the resulting fit, and only caught it by comparing predictions against a published reference.

For response-versus-time models with no dosing (the whole MBMA class) there is no compartment to carry time, so the current workaround is a dummy integrating state:

```
[structural_model]
  ode(states=[clock])
[odes]
  d/dt(clock) = 1
```

which works, but `clock` starts at the subject's **first record**, not at t = 0 — a second silent trap for any dataset without a `TIME = 0` row.

## Suggested fix

The machinery already exists. `collect_undefined_vars()` in `src/parser/model_parser.rs` does exactly this job for `[odes]`, and its own doc comment states the rationale:

```rust
/// Used to reject undefined references in the `[odes]` block
/// before they silently corrupt the dynamics (issue #314).
```

Run the same collector over every `y[CMT=N]` and `obs_scale` expression, against the union of individual parameters, ODE states, covariate columns and builtins; reject at parse time naming the offending identifier.

Two things worth doing in the same pass:

1. **Make `TIME` / `T` / `TAFD` / `TAD` resolvable in `[scaling]`.** They already are in `[odes]`, the asymmetry is not documented anywhere, and it removes the reason to reach for the `clock` workaround at all.
2. #995 (reserved names live in seven scattered sets) is the obvious prerequisite for doing this without adding an eighth.

## Related

- #994 — same silent-zero class, `init()` scope
- #995 — reserved-name sets, needed to define "defined" consistently
- #1005 — `[scaling]` is already parsed a second time for readout name references; likely the natural hook
- #314 — the `[odes]` guard this would reuse
- #767 — parser diagnostics: collect rather than first-error-wins
- #811 — a `$PRED`-style compartment-free structural model would remove the `clock` workaround entirely

## Environment

```
ferx-core 0.2.0 / ferx-r 0.2.0.9000, R 4.5.3 aarch64-apple-darwin
```

roninsightrx:
## Why

Closes #1028.

Two silent-zero failures in `[scaling]`, both of which produce a model that
validates, fits, converges, and reports plausible parameters for a structural
model nobody wrote.

**1. `TIME` in a Form C readout was always 0.** `y = <expr>` referencing the
`TIME` built-in parsed fine, but the readout runs *after* the integrator drops
its `ModelTimeGuard`, so `Op::PushTime` read the thread-local's `0.0` default.
The natural way to write a response-versus-time readout —

```ferx
[scaling]
  y[CMT=1] = EMAX * TIME / (TIME + T50)
```

— therefore collapsed the whole time-dependent term with no diagnostic. The
reporter hit this writing a model-based meta-analysis and "concluded the opposite
of the truth from the resulting fit", catching it only by comparing against a
published reference. The analytic Form C path (#650) had already been given the
guard; the ODE path never was, so this was an undocumented asymmetry between two
halves of the same feature.

**2. An undefined identifier in `[scaling]` reached the predictor as 0.** The
parser classifies any name it cannot bind to a theta / eta / individual parameter
/ state as a **covariate**, i.e. a required data column — which is the correct
design, since the parser genuinely cannot tell a typo from a column the dataset
carries. But two links in that chain were missing: `build_obs_scale_spec` never
registered its covariate leaves at all, and `predict()` ran no covariate check.
So `ferx_predict()` on the issue's reprex returned `PRED = 0` at every row.

### `TIME` / `T` in `[scaling]`

- `OdeReadout::eval` takes the observation `time` and enters `ModelTimeGuard`
  for its two Form C arms. That makes it the single seam shared by the ODE
  predictor (`read_observable`) and the analytic path
  (`pk::apply_analytic_readout`) — whose own outer `with_model_time` wrapper is
  now redundant and dropped, so the two cannot drift again. The `ObsCmt` arm
  (the overwhelmingly common built-in readout) skips the guard entirely.
- `sens::ode_provider::resolve_obs_readout` enters the matching guard, so the
  `Dual2`/`Dual1` walks differentiate the same expression FD evaluates. `TIME`
  is η/θ-constant, so it folds in as `k(t)` — the guard changes the value, not
  the derivative basis.
- The reactive driver enters it around the `[adaptive_dosing] observe` closure
  too: `observe` compiles through the same `build_y_output_fn`, but the driver
  calls the closure directly rather than through `OdeReadout::eval`.
- `T` / `t` — the `[odes]` spelling of the same built-in — folds into
  `Expression::Time` in `[scaling]`, so the two blocks agree on the name. The
  fold targets `Covariate` leaves only, so a state or individual parameter
  genuinely named `T` still wins.
- `obs_scale` referencing `TIME` (or `T`) is now a **parse error**. The divisor
  is subject-static — `apply_scaling` evaluates it once at `t = 0` and divides
  the whole prediction vector — so `TIME` there could only ever read 0. The
  error names Form C as the place for a time-dependent readout.

### Undefined-name guard

- `build_obs_scale_spec` collects its covariate leaves and returns them;
  `parse_scaling_block` pools them with the Form C `y` covariates so **both**
  halves of the block reach `register_referenced_covariates`.
- `predict()` gains `assert_covariates_present`, matching the `assert_*` fail-loud
  convention its sibling preconditions already use. `fit()` and `simulate()`
  already refused this; only `predict()` did not.

`TAFD` / `TAD` are deliberately **not** made built-ins in `[scaling]`: both are
standard NONMEM dataset column names, and reserving them would break every model
that legitimately reads such a column. They stay covariate references, which the
guard above now makes fail loudly when the column is absent. This is the one part
of the issue's suggested fix I did not take; see *Alternatives considered*.

## Alternatives considered

**Reject unbound identifiers at parse time** (the issue's literal suggestion,
reusing `collect_undefined_vars` from the `[odes]` guard). This does not work for
`[scaling]`: unlike an `[odes]` RHS — which parses with `fallback_covariate =
false` and genuinely has a closed name scope — `[scaling]` expressions may
reference covariates, and the parser has no dataset to check them against. A
parse-time rejection would have broken every covariate-scaled readout in the
wild. The existing register-then-check-against-data mechanism is the right
shape; it just had two holes in it.

**Reserve `TAFD` / `TAD` in `[scaling]` too.** Rejected: see above. A follow-up
issue can add them as built-ins if the name collision is judged acceptable, but
it is a breaking change for real datasets and does not belong in a silent-zero
fix.

## Cross-repo dependency

No `pub` API signature changed (`OdeReadout::eval` and `read_observable` are
`pub(crate)` / private).

## Breaking changes

- [ ] Public Rust API
- [ ] None

<details>

Two narrow breaks, both of which only affect models that were already silently
wrong:

1. **`obs_scale = TIME` / `obs_scale = T` is now a parse error.** The divisor is
   subject-static, so it was reading `TIME = 0` — the fix is to move the
   time-dependent term into a Form C `y = <expr>` readout, where `TIME` resolves
   per observation.
2. **A `[scaling]` expression referencing a data column named `T` now reads the
   model-time built-in.** `[odes]` has always reserved that name; `[scaling]` now
   matches. Rename the column, or reference it under a different name.

Not a break, but worth calling out: a model whose `obs_scale` references a
covariate the data does not carry now fails with `E_MISSING_COVARIATE` instead of
dividing by the covariate map's `0.0` default. Same for `predict()` on any model
with a missing covariate column.

Also worth noting for anyone who used the workaround: the dummy integrating
state —

```ferx
[structural_model]
  ode(states=[clock])
[odes]
  d/dt(clock) = 1
```

— is no longer needed, and is better dropped. `clock` starts at the subject's
**first record**, not at `t = 0`, so it silently disagrees with `TIME` on any
dataset with no `TIME = 0` row.

</details>

## Numerical validation

- [x] Reference values:

New anchor `nonmem_anchor/scaling_time_readout.{ctl,csv}` — `ADVAN13 TOL=9`,
`MAXEVAL=0`, every `$THETA` `FIX`, `$OMEGA 0 FIX`, so it is a pure evaluation at
known parameters with no optimiser on either side. ferx's `[scaling] y = <expr>`
and NONMEM's `$ERROR IPRED = ...` are the same object — the structural prediction
handed to the residual error model, evaluated once per observation record — and
NONMEM has always had `TIME` in scope there, so the expression translates line
for line.

```
$ERROR
  IPRED = A(1)/V + BETA*TIME/(TIME + T50)
```

```ferx
[scaling]
  y = central / V + BETA * TIME / (TIME + T50)
```

100 mg IV bolus, `CL = 5`, `V = 50`, `BETA = 3`, `T50 = 4`:

```
   t   NONMEM IPRED   A(1)/V alone   BETA·t/(t+T50)
 0.5   2.2358         1.9025         0.3333
 1.0   2.4097         1.8097         0.6000
 2.0   2.6375         1.6375         1.0000
 4.0   2.8406         1.3406         1.5000
 8.0   2.8987         0.8987         2.0000
12.0   2.8524         0.6024         2.2500
24.0   2.7529         0.1814         2.5714
```

ferx reproduces the `IPRED` column to **<1e-4 relative** (NONMEM's table carries
five significant figures). The middle column is what the pre-fix readout produced
— 1.18x off at the first sample, 15x by 24 h — so the anchor discriminates the
bug by three orders of magnitude more than the tolerance.
`the_nonmem_anchor_discriminates_a_stale_time_of_zero` pins that separation, so
the anchor cannot quietly stop discriminating if the reference is ever
regenerated.

Committed outputs: `nonmem_anchor/results/scaling_time_readout.{lst,tab}`;
catalogue entry + run recipe added to `nonmem_anchor/README.md`.

**Simulate-and-refit recovery** (`examples/scaling_time_readout.ferx --simulate`,
30 subjects, seed 1) — a second, independent oracle: both parameters that live
*only* in the time term come back, which they could not if `TIME` were 0.

```
param     estimate   truth
TVCL      4.9139     5.0
TVV      50.5053    50.0
TVEMAX    2.9914     3.0     <- the TIME term's asymptote
TVT50     4.0442     4.0     <- and its half-maximal time
```

FOCEI, covariance computed, 0.4 s wall.

## Performance

The guard is a thread-local `Cell::replace` plus a restore on drop, entered once
per observation and only on the Form C arms — the `ObsCmt` readout (every model
without a `[scaling] y`) is untouched. It is negligible against the bytecode eval
it wraps.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 3074
      passed, 0 failed)

13 new tests. Every one was confirmed to **fail** with the guard neutered to
`enter(0.0)` / against `origin/main` behaviour:

| Test | Tier | What it pins |
|------|------|--------------|
| `sens::ode_provider::…::ode_provider_form_c_time_readout_evaluated_at_obs_time` | 1 | `∂y/∂TVBETA == t` on the ODE dual walk, plus FD-of-production parity |
| `sens::ode_provider::…::ode_form_c_time_readout_prediction_uses_obs_time` | 1 | the f64 predictor itself — differencing at two `BETA` recovers exactly `BETA·t` |
| `api::adaptive_sim_tests::from_spec_observe_reads_the_time_builtin_at_the_decision` | 1 | `observe = TIME` sees each decision's time; the dose ladder discriminates a stale 0 |
| `parser::…::scaling_obs_scale_registers_covariate_references` | 1 | `obs_scale` covariate leaves become required data columns |
| `parser::…::scaling_obs_scale_per_cmt_registers_covariate_references` | 1 | …for the per-CMT form too |
| `parser::…::scaling_obs_scale_indiv_param_is_not_a_covariate` | 1 | negative control — no over-registering |
| `parser::…::scaling_obs_scale_rejects_the_time_builtin` | 1 | all four spellings rejected, error names Form C |
| `parser::…::scaling_y_readout_accepts_the_time_builtin` | 1 | `TIME` in Form C is a built-in, not a column |
| `parser::…::scaling_y_readout_accepts_the_t_time_alias` | 1 | `T`/`t`/`TIME`/`time` all compile to `Op::PushTime` |
| `parser::…::scaling_y_readout_t_alias_does_not_shadow_a_bound_name` | 1 | a bound `T` wins over the alias |
| `parser::…::scaling_y_readout_folds_the_t_alias_under_every_node_kind` | 1 | the fold recurses through `Conditional` / `Condition::And` / `UnaryFn` / `Power`, not just top-level operands |
| `api::scaling_undefined_tests::predict_rejects_an_undefined_scaling_identifier` | 1 | the issue's reprex, through `predict()` |
| `api::scaling_undefined_tests::predict_accepts_a_scaling_identifier_the_data_carries` | 1 | positive control — the guard keys on the data, not the name |
| `tests/scaling_time_readout_nonmem_anchor.rs` (2 tests) | 2 | the NONMEM anchor + its discrimination check |

Two pre-existing analytic-path tests (`form_c_time_readout_evaluated_at_obs_time`,
`iov_form_c_time_readout_evaluated_at_obs_time`) also fail with the guard
neutered, which confirms `OdeReadout::eval` really is now the single seam for
both paths.

## Example

- [x] Example added to `examples/scaling_time_readout.ferx`

```ferx
[structural_model]
  ode(states=[central])            # no obs_cmt — Form C below provides the readout

[odes]
  d/dt(central) = -(CL / V) * central

[scaling]
  y = central / V + EMAX * TIME / (TIME + T50)
```

`cargo run --release -- examples/scaling_time_readout.ferx --simulate` — runs
clean and recovers the truths (table above).

## Docs

  - `docs/model-file/scaling.qmd` — new "`TIME` in a Form C readout" and
    "Undefined names are data columns, not zeros" sections; the Form B section
    now states the `obs_scale` restriction and why.
  - `docs/model-file/adaptive-dosing.qmd` — `observe` inherits the built-in.
  - `docs/examples/scaling.qmd` — pointer to the new example + anchor.
- [x] No new pages, so `docs/_quarto.yml` is unchanged.
- [x] `CHANGELOG.md` `[Unreleased] → Fixed` entry added.

## Checklist

- [x] `cargo clippy --all-targets` clean (the 5 remaining warnings are
      pre-existing in `sir.rs` / `validation.rs:1847+`, untouched by this PR)
- [x] `Dual2` sensitivity path stays differentiable — `resolve_obs_readout` is
      generic over `PkNum` and the guard sits outside the dual arithmetic;
      `Op::PushTime` was already handled by `eval_bytecode_g` as `k(t)`.
- [ ] `Cargo.toml` version bumped — not needed (unreleased `0.3.0`, and the
      `[Unreleased]` entry carries the two narrow breaks).

## Reviewer hints

Focus on:

- **`src/ode/predictions.rs` — the six `read_observable` call sites.** Each had
  to supply the right time. Four are observation reads (`subject.obs_times[idx]`
  or the `obs_map`-keyed `t_start`), two are adaptive-controller reads (the
  decision time `t_start`, and the AUC grid point `grid[i]`). Worth checking I
  picked the right clock at each.
- **`src/parser/model_parser.rs::rewrite_scaling_time_alias`** — the `T` fold's
  precedence rule (`Covariate` leaves only, so a bound name wins).
- The `visit_expr_nodes_mut` / `visit_condition_nodes_mut` pair is mechanical —
  the mutable twin of the existing `visit_expr_nodes` walker, same arms.

Can be skimmed: the docs and CHANGELOG prose, and the NONMEM `.lst`.

## Open questions

- **`TAFD` / `TAD` in `[scaling]`.** Left as covariate references on purpose
  (they are common dataset column names, so reserving them would break real
  models), which does leave the asymmetry with `[odes]` — where they *are*
  reserved and a same-named column is unreachable — in place. Happy to open a
  follow-up if you'd rather have them as built-ins with a documented break.
- **`ferx_model_validate()` with no `--data` still returns `ok = TRUE`** on the
  issue's reprex, since the required-column check is data-dependent by nature.
  I considered a data-absent warning listing every required covariate (there is
  precedent — `W_ETA_NOT_DECLARED`), but it would fire on every ordinary
  covariate model. `ferx check --data` reports `E_MISSING_COVARIATE`. Say the
  word if you want the noisier version.

---


roninsightrx:
## Code review — `2880946b`

The core change is sound: the ODE Form C readout now runs under a per-observation `ModelTimeGuard` (the analytic side already did), `obs_scale` covariate leaves are finally registered as required data columns, and `predict()` gains the covariate check that `fit()`/`simulate()` already had. Guard placement covers every readout seam I could find (`record_observations`, event-driven, adaptive decision hook, window AUC grid, dual walk), and `visit_expr_nodes_mut` covers every recursive `Expression` variant.

Five findings below.

### 1. `src/pk/modified_release.rs:275` — medium

The MR closed-form fast path bypasses the readout entirely (`mr_predictions` → `mr_observable_g` = `m·amount`) and admits a readout only via a linearity probe: `y0 = readout(0) ≈ 0`.

A now-supported Form C readout with an additive time term (`y = central/V + BETA*TIME/(TIME+T50)`) evaluates `y0` under the ambient model-time thread-local — `0.0` at that point — so `y0 == 0`, the probe passes, and the whole TIME term is silently dropped for any MR-scoped subject (input-rate forcings, ≤2 states, bolus doses). `predict`/`simulate`/FOCE values then disagree with the ODE path this PR just fixed.

Worse, scope membership becomes dependent on ambient thread-local state: if `mr_scope` were ever reached under a guard with t≠0, `y0 ≠ 0` and the same model would decline instead.

Suggest declining in `identify_disposition`/`recover_disp_params_g` when the readout program contains `Op::PushTime`.

### 2. `src/parser/model_parser.rs:7366` (applied at `:7420`) — medium

`rewrite_scaling_time_alias` rewrites *every* `Covariate("T")`/`Covariate("t")` leaf in a Form C readout to `Expression::Time`.

Form C `y` covariates have been registered as required data columns since #540, so a dataset with a real column named `T` (or `t`) and a readout like `y = central/V * T` worked correctly before this PR and now silently reads model time instead of the column — no error, wrong predictions. That is the exact silent-wrong class this PR is fixing, reintroduced for one name.

Suggest only folding `T` when the model/data does not carry that covariate, or at minimum erroring on the collision instead of silently preferring the alias.

### 3. `src/ode/predictions.rs:1603` (also `:4653`, `src/sens/ode_provider.rs:1741`) — low

The readout's `TIME` is seeded from `subject.obs_times`, the internal monotonic timeline. For subjects with stacked reset occasions whose data TIME restarts, that differs from `obs_raw_times`, which `ModelParameters::ruv_obs_mult` (`src/types.rs:3751`) explicitly documents as "the raw data-file time, matching what NONMEM's `$ERROR` sees".

Since this PR frames Form C as the `$ERROR` twin (and anchors it against NONMEM), those two `TIME`s should agree; today a reset-stacked dataset gets one convention in the readout and the other in the residual-magnitude model.

### 4. `src/api/predict.rs:63` — low

`assert_covariates_present` keys purely on `population.covariate_names`. An in-memory `Population` whose `Subject.covariates` maps are populated but whose `covariate_names` was left empty (a legal, previously-working programmatic construction — `predict()` resolves covariates from the subject map, not the name list) now panics.

`simulate()` has the same shape, so it is at least consistent, but it is a public-API behaviour change worth a CHANGELOG note, or a fallback that also accepts a name every subject's map carries.

### 5. `src/parser/model_parser.rs:7420` — low

The alias fold covers `T`/`t` only, while `[odes]` also reserves `TAFD`, `TAD`, and `MACHEPS` (`RESERVED_ODE_NAMES`, `:9701`). The new docs/example state that `[scaling]` spells the built-ins the way `[odes]` does, but `y = ... TAD ...` still falls through to a covariate and now fails with a misleading `Model references covariate(s) not found in data: TAD`. Either fold them too, or make the error name them as unsupported-in-`[scaling]` built-ins.

### Checked and cleared

Guard nesting/restore semantics (`ModelTimeGuard` is RAII + thread-local, rayon-safe); `obs_map` keys are `subject.obs_times`, so the `t_obs` lookups match the integrator grid exactly; adaptive `t_start` uses are keyed off the same map; `scaling_covariates` registration happens at the same seam as the pre-existing `y_covariates` one, so `CovariateIdx` resolution ordering is unaffected; analytic providers (`apply_readout_jet`, `eval_readout_at_obs`) already carried the matching guard; `ReadoutSnapshots::Static` caches only the param jet, not the readout value.

roninsightrx:
All five addressed in `3c73439a`.

**1. MR fast path (`src/pk/modified_release.rs`) — fixed as suggested.** New `OdeOutputProgram::reads_time_builtin()` (scans the bytecode for `Op::PushTime`); `recover_disp_params_g` declines on it, and `mr_scope` declines too so scope membership can never depend on the ambient thread-local. Regression test `declines_time_dependent_readout` pins all three MR entry points (`mr_scope` / `mr_predictions` / `mr_subject_sensitivities`) declining, with a baseline assertion that the same model *without* the time term is still admitted, and a check that the ODE path now serving the model carries the term exactly (`Δ = 0.5·t`).

**2. `T` alias vs. a real `T` column — fixed.** The fold now loses to a `[covariates]` declaration — the same precedence a bound state or individual parameter already had — and pushes a `parse_warnings` note whenever it *does* fire, naming both escapes (spell it `TIME`, or declare the column). So the substitution is never silent, and the previously-working `T`-column model has an explicit way to keep working. `[scaling]` peeks the declarations rather than hoisting the block parse, so a malformed `[covariates]` still reports at its own site in the existing order. Tests: `scaling_y_readout_t_alias_loses_to_a_declared_covariate`, `scaling_y_readout_t_alias_fold_warns` (including that the unambiguous `TIME` spelling does not warn).

**3. Readout `TIME` convention — fixed.** New `Subject::readout_time(j)` (raw with `obs_times` fallback) is now the single source at all five readout seams: `record_observations`, the event-driven walk, `apply_analytic_readout`, and both dual walks (`sens::provider::apply_readout_jet` / `eval_readout_at_obs`, `sens::ode_provider::resolve_obs_readout`). The integrator keeps `obs_times`, exactly as `$DES`'s clock is not `$ERROR`'s `TIME`. Test `form_c_readout_time_is_the_raw_data_clock` uses a readout that returns the model-time thread-local, so the prediction vector *is* the sequence of `TIME` values seen — asserted against a reset-stacked `obs_raw_times`, plus the no-raw-times fallback.

**4. `predict()` covariate check — fixed.** `check_covariates` now also accepts a name every subject's `covariates` map carries, so a programmatically-built `Population` with an empty `covariate_names` keeps working (the predictors resolve from the subject map anyway). No change for CSV-read populations, and the reprex still fails loudly. Test `predict_accepts_a_covariate_carried_only_by_the_subject_maps`. CHANGELOG notes the `predict()` check and this allowance.

**5. `TAFD` / `TAD` / `MACHEPS` — fixed via the message, not a rejection.** I started on the fold/reject route and then hit `docs/model-file/scaling.qmd`: keeping them as ordinary covariates in `[scaling]` is deliberate and documented, precisely so a dataset that really carries a `TAD` column can use it — rejecting them would have broken that. So the fix is the second option you offered: `E_MISSING_COVARIATE` now appends a sentence naming the `[odes]`-only scope when one of those names is among the missing, and is byte-for-byte unchanged otherwise (the historical-text pin still passes). Tests: `check_covariates_names_the_odes_scope_for_a_missing_builtin` and `check_covariates_hint_is_absent_for_an_ordinary_covariate`; `scaling_odes_only_builtins_stay_covariates` pins that they keep registering as required data columns.

Docs (`docs/model-file/scaling.qmd`) gained the raw-clock statement, an "if your dataset has a column named `T`" subsection, and the scope note on the missing-`TAD` error. CHANGELOG entry extended.

`cargo test --lib` 3083 passed / 0 failed, `cargo check --tests` clean, the NONMEM anchor still passes, no new clippy warnings on the touched files.

PR #1043, opened 2026-08-21
fix(ode): gate the adaptive DV monitor assay clamp on the readout [closes #1039]

### LINKED ISSUE #1039: Adaptive-dosing DV monitor still clamps negative Form C readouts to zero (ipred/dv sign asymmetry)
## Summary

After #1020 / #1038, an adaptive-dosing monitor reading the same Form C
`[scaling]` model gets **opposite sign semantics depending on its
`ObserveMode`**:

- `ObserveMode::Ipred` — reads the readout through the (now signed) prediction
  path, so a negative Form C readout is returned as-is.
- `ObserveMode::Dv` — adds assay noise and then applies `.max(0.0)`
  (`src/ode/predictions.rs:3414`, "Edge (b): an assay cannot read below zero"),
  so every negative sample collapses to exactly `0`.

## Why it matters

The `.max(0.0)` on the DV path was written when *every* readout was a
compartment amount / concentration, where "an assay cannot read below zero" is
true. It is not true of a Form C readout, which is an arbitrary expression:
a change from baseline, a difference from comparator, a z-score, or the
`sqrt(N) * logit(p)` transform used in model-based meta-analysis.

Concretely: a controller thresholding a change-from-baseline signal
(`y = central / V - BASE`) silently reads `0` for every sample where the
subject is below baseline — exactly the region such a controller is usually
designed to react to. There is no warning; the controller just sees a floored
signal and makes the wrong dose decision.

The asymmetry is the sharp edge: the *same model* with `mode = ipred` behaves
correctly and with `mode = dv` does not.

## Where

- `src/ode/predictions.rs:3414` — the DV-path clamp.
- `src/ode/predictions.rs:1859` — `OdeReadout::clamps_negative()`, the existing
  gate #1020 introduced. The DV monitor path does not consult it.

## Proposed fix

Gate the DV clamp on the same predicate the prediction path now uses:

```rust
let noised = latent + var.sqrt() * eps;
if ode.readout.clamps_negative() { noised.max(0.0) } else { noised }
```

That keeps the assay floor exactly where it is physically true (bare-state /
`obs_scale` readouts, Forms A and B) and drops it for Form C, matching the
prediction path. Add a Tier-1 test that a Form C monitor in `dv` mode returns a
negative sample, alongside the existing
`adaptive_dv_clamps_negative_at_zero` (`src/ode/predictions_tests.rs:2126`),
which should be kept and re-scoped to a bare-state readout.

## Validation

Per the epic #391 exception in `CLAUDE.md`, this family is not NONMEM-anchored.
A degenerate-oracle check suffices: with `sigma -> 0`, a `dv` monitor on a Form
C model must equal the `ipred` monitor on the same model, sample for sample —
including negative samples. That equality is what the current clamp breaks.

## Context

Split out of the #1038 review (fix for #1020). Deliberately deferred there: the
prediction/likelihood path was the scoped change, and the monitor path is
adaptive-dosing surface (epic #391) with its own validation rules.

roninsightrx:
## Why

Closes #1039.

After #1020 / #1038 a Form C `[scaling]` readout keeps its negative part on the
prediction path — but an adaptive-dosing monitor reading the *same model* still
got opposite sign semantics depending on its `ObserveMode`:

- `ObserveMode::Ipred` — reads the (now signed) prediction path, negative readout
  returned as-is.
- `ObserveMode::Dv` — added the assay residual draw and then applied `.max(0.0)`
  unconditionally (`src/ode/predictions.rs`, "Edge (b): an assay cannot read below
  zero"), so every negative sample collapsed to exactly `0`.

That floor was written when every monitored readout was a compartment amount /
concentration, where non-negativity is physical. It is not true of a Form C
readout, which is an arbitrary expression: a change from baseline, a difference
from comparator, a z-score, or the `sqrt(N) * logit(p)` transform used in
model-based meta-analysis.

Concretely: a controller thresholding `y = central / V - BASE` silently read `0`
for every sample below baseline — exactly the region such a controller exists to
react to — with no warning, and dosed on the floored signal. The asymmetry was the
sharp edge: the same model behaved correctly under `mode = ipred` and not under
`mode = dv`.

## What changed

Gate the DV clamp on `OdeReadout::clamps_negative()` — the same predicate the
prediction path (`clamp_negative_predictions`, the event-driven inline clamp) and
the `sens` providers already consult:

```rust
let noised = latent + var.sqrt() * eps;
if ode.readout.clamps_negative() { noised.max(0.0) } else { noised }
```

The assay floor therefore stays exactly where it is physically true — bare-state
(`obs_cmt`) readouts and Forms A/B, which keep the default bare-state readout —
and drops for both Form C variants, matching the prediction path.

## Alternatives considered

Clamping the *latent* value instead of the noised one: same defect, and it would
diverge from the prediction path, which clamps neither for Form C. Warning on a
negative DV sample: noise on a legitimately-signed signal makes negatives the
normal case, so the warning would fire on every well-behaved run.

## Cross-repo dependency

No `pub` API change: the gate is internal to `ode_predictions_adaptive_impl`.

## Breaking changes
- [x] None

Behaviour change is confined to Form C models with a `dv` monitor, where the
previous output (`0`) was the bug.

## Numerical validation

Per the epic #391 exception in `CLAUDE.md`, adaptive / feedback dosing is **not**
NONMEM-anchored — NONMEM has no controller reading simulated state. Validated with
the **degenerate oracle** that rule prescribes:

- `adaptive_dv_zero_variance_equals_ipred_on_signed_form_c` — with `sigma -> 0`, a
  `dv` monitor on a signed Form C model (`clock - 5`) equals the `ipred` monitor
  decision for decision across the sign change: `[-5, -3, 1, 5]` at
  t = `[0, 2, 6, 10]`, negative samples included. This equality is exactly what the
  ungated clamp broke (it returned `[0, 0, 1, 5]`).

- [x] Not applicable (adaptive-dosing family — degenerate oracle above per #391)

## Performance
- [x] No significant impact expected — one predicate call (a match on an enum
  discriminant) per DV monitor per decision.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3068 passed, 0 failed)

Tier 1, in `src/ode/predictions_tests.rs`:

- `adaptive_dv_form_c_monitor_keeps_negative_sample` — a noised (`sd = 0.1`) Form C
  `dv` monitor returns a negative, genuinely perturbed sample. Returns `0.0` without
  the fix.
- `adaptive_dv_zero_variance_equals_ipred_on_signed_form_c` — the degenerate oracle
  above.
- `adaptive_dv_clamps_negative_at_zero` — **kept**; it already exercises a bare-state
  (`OdeReadout::ObsCmt`) readout, and its comment is re-scoped to say so explicitly
  and point at the Form C exemption. It still pins the floor where it is real.

The shape of the affected model, for reference:

```toml
[scaling]
  y[CMT=1] = central / V - BASE     # change from baseline: signed by construction

[adaptive_dosing]
  monitor CHANGE = cmt 1, mode = dv  # now sees the negative arm; previously read 0
```

## Docs
- [x] `docs/` updated for user-visible changes — `docs/model-file/adaptive-dosing.qmd`
  (the "noised reading below zero is clamped to 0" sentence now states the readout
  gate and the `sigma -> 0` equality) and `docs/model-file/scaling.qmd` (the signed-
  readout section now covers the monitor path too).
- [x] `CHANGELOG.md` `[Unreleased]` entry added under `### Fixed`.

## Checklist
- [x] `cargo clippy` clean — no new findings. (`--lib --tests` has 4 pre-existing
  `approx_constant` denies in `src/sens/two_cpt_explicit.rs`, untouched by this PR.)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — not touched;
  `sens/ode_provider.rs` already gates its clamp on the same predicate (#1020).

### ferx-site
- [x] No new DSL syntax, `[fit_options]` key, example, or data file.

### ferx-book
- [x] No new estimator, DSL feature, or fit option.

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` model uses a Form C
  readout with a `dv` monitor, so no example output changes.

## Reviewer hints

Focus on the one gated expression in `src/ode/predictions.rs` (the `ObserveMode::Dv`
arm) and on which test now owns which semantic: the bare-state floor
(`adaptive_dv_clamps_negative_at_zero`) versus the Form C exemption (the two new
tests). Worth confirming the claim in the surrounding comment still holds — under
`Dv` the declarative path compiles no `observe` expression, so `latent` is the
model's own readout and σ comes from the same output, which is what makes
"gate on the model's readout" the right predicate here rather than something
monitor-local. The docs and changelog edits can be skimmed.

## Open questions

The asymmetry in the *other* direction is left as-is and is not a bug: on a
bare-state readout an `ipred` monitor can still return a negative solver-overshoot
value while a `dv` monitor floors it at `0`. Clamping the `ipred` monitor would be
a separate change to the latent readout path, out of scope for #1039.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,18 @@ section of the SDLC for the versioning policy).
   with no diagnostic from NONMEM (`nonmem_anchor/dose_attr_double_use_{A,B}.ctl`).
 
 ### Fixed
+- **An adaptive-dosing `dv` monitor no longer floors a negative Form C `[scaling]` readout at zero
+  (#1039).** The assay floor on the `ObserveMode::Dv` path ("an assay cannot read below zero") was
+  written when every monitored readout was a compartment amount or concentration, and was applied
+  unconditionally after the residual draw. A Form C `y = <expr>` readout is an arbitrary
+  expression — a change from baseline, a difference from a comparator, a z-score, the
+  `sqrt(N) * logit(p)` transform — so the *same model* read correctly under `mode = ipred` and
+  came back as exactly `0` under `mode = dv` for every negative sample, silently: a controller
+  thresholding a change-from-baseline signal saw `0` over precisely the region it was written to
+  react to, and dosed accordingly. The floor is now gated on the same predicate as the prediction
+  path (#1020), so it applies only to the bare-state readout and to Forms A/B, which keep it. With
+  `sigma → 0` a `dv` monitor again reproduces the `ipred` monitor sample for sample, negative
+  samples included.
 - **SAEM no longer lets a fixed-effect-only theta drift away from the marginal optimum (#1011).** The
   numerical θ/σ M-step assigned NLopt's maximiser outright, re-maximising against a *single* MCMC η
   draw each iteration — `argmax` of one draw rather than the stochastic-approximation average of
--- a/docs/model-file/adaptive-dosing.qmd
+++ b/docs/model-file/adaptive-dosing.qmd
@@ -182,8 +182,9 @@ expression may be legitimately signed (a change from baseline, a z-score, a
 difference from comparator), so its negative samples reach the controller
 unchanged, exactly as they do under `Ipred`. See
 [Scaling](scaling.qmd#signed-readouts-since-1020). With `sigma → 0` a `Dv`
-monitor therefore reproduces the `Ipred` monitor sample for sample on *any*
-readout, negative samples included.
+monitor therefore reproduces the `Ipred` monitor sample for sample on a Form C
+readout, negative samples included; on the bare-state readout the floor still
+applies, so the two agree only where the latent value is non-negative.
 
 ## Example
 
--- a/docs/model-file/adaptive-dosing.qmd
+++ b/docs/model-file/adaptive-dosing.qmd
@@ -175,7 +175,15 @@ draws (under a fixed `seed`):
   draws, so the frozen-replay verifier stays exact.
 
 A `Dv` monitor on a compartment with no `[error_model]` is a typed error (never a
-fabricated σ), and a noised reading below zero is clamped to 0.
+fabricated σ), and a noised reading below zero is clamped to 0 — but **only on the
+default bare-state readout**, where non-negativity is a physical property of the
+quantity. Since #1039 a Form C `[scaling] y = <expr>` readout is exempt: the
+expression may be legitimately signed (a change from baseline, a z-score, a
+difference from comparator), so its negative samples reach the controller
+unchanged, exactly as they do under `Ipred`. See
+[Scaling](scaling.qmd#signed-readouts-since-1020). With `sigma → 0` a `Dv`
+monitor therefore reproduces the `Ipred` monitor sample for sample on *any*
+readout, negative samples included.
 
 ## Example
 
--- a/docs/model-file/scaling.qmd
+++ b/docs/model-file/scaling.qmd
@@ -115,6 +115,12 @@ PK concentration), where non-negativity is a physical property of the quantity
 and a negative value can only be integration overshoot. It is not applied to
 Form C.
 
+The same gate applies to the adaptive-dosing monitor path: a `Dv` monitor's
+assay floor (`max(0)` after the residual draw) fires only on the bare-state
+readout, so a Form C signal a controller thresholds is not silently floored at
+`0` over its whole negative region (#1039). See
+[Adaptive dosing](adaptive-dosing.qmd).
+
 Forms A and B (`obs_scale = <expr>` / `obs_scale[CMT=n] = <expr>`) keep the
 default bare-state readout — the divisor is applied *downstream* of it — so the
 clamp still guards the compartment state they read. Only Form C (`y = <expr>`)
--- a/src/ode/predictions.rs
+++ b/src/ode/predictions.rs
@@ -3409,9 +3409,23 @@ pub(crate) fn ode_predictions_adaptive_impl(
                             // finiteness guard. Value-pathology (a NaN/∞ in `sigma`, a
                             // diverged IPRED) is whole-sim garbage-in, out of scope here.
                             let eps = assay_standard_normal(a.base_seed, decision_index, &m.name);
+                            let noised = latent + var.sqrt() * eps;
                             // Edge (b): an assay cannot read below zero; clamp the
                             // noised value at 0 (BLQ-blinding is deferred to Part F).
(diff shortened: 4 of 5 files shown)


roninsightrx:

Reviewed the gate on the adaptive `ObserveMode::Dv` assay clamp, plus tests/docs/changelog.

### Findings (1)

**`docs/model-file/adaptive-dosing.qmd:184` — minor, docs correctness.**

The new sentence claims:

> With `sigma → 0` a `Dv` monitor therefore reproduces the `Ipred` monitor sample for sample on *any* readout, negative samples included.

That is false for the bare-state (`OdeReadout::ObsCmt`) readout — which is exactly the readout that still clamps. Concrete case: a bare-state model where RK overshoot leaves `state[idx] = -1e-9` at a decision time. `read_observable` does not clamp (the prediction-recording clamp lives at `src/ode/predictions.rs:4658`, not on the monitor path), so the `Ipred` monitor returns `-1e-9` while the `Dv` monitor with `sigma → 0` returns `noised.max(0.0) = 0`. The PR's own "Open questions" section acknowledges this asymmetry and calls it out of scope, so the doc contradicts the shipped behaviour.

Suggested fix: scope the claim to Form C readouts. The same sentence in `CHANGELOG.md` reads as Form-C-scoped from its surrounding paragraph and is fine as-is.

### Verification done

- The gate predicate is the right one for both entry paths: `compile_adaptive` (`src/sim/adaptive_control.rs:311-326`) sets `observe = None` for `with_assay_error`, and the programmatic path (`src/api/adaptive.rs:252-259`) always sets `observe: None` — so under `Dv` the `latent` really is `read_observable` over `ode.readout`, making the gate on `ode.readout.clamps_negative()` sound. `PerCmt` is all-Form-C by construction (`OdeReadout::eval`), so the whole-readout gate cannot mis-fire per compartment.
- `assay_standard_normal` has exactly one call site — no second un-gated DV clamp elsewhere.
- `cargo test --lib adaptive_dv` — 8/8 pass, including the two new tests and the retained bare-state floor test.
- The doc anchor `scaling.qmd#signed-readouts-since-1020` resolves to the real heading at `docs/model-file/scaling.qmd:100`.
- The extracted `noised` binding is behaviour-preserving for the clamped branch; `adaptive_dv_clamps_negative_at_zero` still pins the floor where it is physical; `adaptive_dv_form_c_monitor_keeps_negative_sample` genuinely fails without the fix (latent `-5`, sd `0.1`, so no draw can reach `0`).

Nothing else.

PR #1044, opened 2026-08-21
feat(dsl): first-class residual weighting `weight = <expr>`, applied in every estimator [closes #1029]

### LINKED ISSUE #1029: feat(dsl): first-class residual weighting -- weight = COL on the error model
Part of the MBMA umbrella issue (linked below). This is the single highest-value item in it.

## The problem

Inverse-variance weighting of the observations *is* what makes a meta-analysis a meta-analysis. In the study-as-subject construction that every published MBMA uses (NONMEM, Monolix, `nlme`), it has to be hand-built in **three separate places that must agree**:

```r
# 1. in R -- transform the observation
DV = WP / WPSE
```

```
# 2. in [scaling] -- divide the prediction by the same quantity
y[CMT=1] = (E0 - EMAX * clock / (clock + ET50)) / WPSE

# 3. in [parameters] -- fix the residual so the weight is exactly 1/SE^2
sigma ADD_ERR ~ 1.0 (variance) FIX
```

Three places, one idea. Divide in only one of them and the model still fits — it just answers a different question.

## The second cost: diagnostics become unreadable

Because `DV` now lives in weighted units, **every diagnostic FeRx ships is in the wrong units**: `sdtab`, the VPC, obs-vs-pred, residual-vs-time. Users have to multiply observed and simulated values back by `WPSE` before plotting anything. That is precisely the manual back-transformation step the Monolix tutorial has to document, and it is pure overhead — the information needed to undo it is right there in the data.

## Proposal

Declare the weight once, on the error model:

```
[error_model]
  DV ~ additive(ADD_ERR) weight = WPSE
```

Engine behaviour:

- divide residual and prediction by `weight` internally, so the objective is unchanged from the hand-built version;
- leave `DV` on the natural scale in the data;
- report `PRED`, `IPRED`, `CWRES`, `sdtab` and the VPC **on the natural scale**, with no back-transformation;
- fix `sigma` to 1 implicitly, or error if the user also declares it free — that contradiction is currently a silent modelling bug.

## It generalises to the bounded-endpoint case

The other standard MBMA transform is a weighted logit for a responder proportion:

```
DV       = sqrt(N) * logit(p_observed)
y[CMT=1] = sqrt(N) * logit(p_predicted)
```

which is the same feature with `weight = 1/sqrt(NARM)` on a logit readout. Note that here `sigma` must be **estimated** rather than fixed, because the `p(1-p)` factor was never divided out and σ absorbs it — the published canakinumab fit puts it at 2.466, and `1/sqrt(p(1-p))` at p ≈ 0.4 is about 2.04, which is most of that. With `weight =` as an explicit declaration, "is sigma fixed here?" becomes a property the engine can reason about instead of a trap.

## Why this is worth doing

It is not keystroke-saving. It removes the largest remaining source of silent error in the MBMA workflow (inconsistent division across three sites), and it turns FeRx's diagnostics from "correct but in units nobody can read" into "usable as shipped".

In a straight comparison of MBMA tooling this is the feature that would put FeRx clearly ahead of the Monolix workflow rather than level with it. Everything else in the umbrella is either a bug fix or a convenience; this one changes what the tool is like to use.

## Related

- #1021 — `ferx_sir()` currently fails when `FIX`ed parameters make the proposal singular, which every study-as-subject MBMA triggers *because* of the fixed σ this proposal would manage
- #811 — compartment-free `$PRED` structural model, the other big MBMA-motivated DSL item

## Environment

```
Verified against the published naproxen (18 trials) and canakinumab (9 studies) analyses.
```

roninsightrx:
## Why

Closes #1029 (part of the MBMA umbrella #1032).

Inverse-variance weighting of the observations *is* what makes a meta-analysis a meta-analysis. In the study-as-subject construction every published MBMA uses, it has to be hand-built in **three separate places that must agree**:

```r
DV = WP / WPSE                                            # 1. in R
```
```
y[CMT=1] = (E0 - EMAX*clock/(clock+ET50)) / WPSE          # 2. in [scaling]
sigma ADD_ERR ~ 1.0 (variance) FIX                        # 3. in [parameters]
```

Three places, one idea. Divide in only two of them and the model still fits — it just answers a different question. And because `DV` then lives in weighted units, **every diagnostic ferx ships** (`sdtab`, VPC, obs-vs-pred, residual-vs-time) comes out in units nobody can read, so users hand-multiply everything back by `WPSE` before plotting. That is exactly the manual back-transformation step the Monolix tutorial has to document, and the information needed to undo it is right there in the data.

### 1. The DSL: declare the weight once, on the error model

```
[error_model]
  DV ~ additive(ADD_ERR) weight = WPSE
```

`weight = W` means **"score as if `DV` and the prediction were both divided by `W`"**. Writing $y' = y/W$, $f' = f/W$, the natural-scale residual variance is

$$\text{Var}(y-f) = W^2 \cdot \text{Var}(f/W)$$

so the **additive** loading picks up a factor `W` and the **proportional** loading is untouched — $W\cdot(f/W)=f$, because a common scale factor cancels out of a constant-CV error:

| Error model | Unweighted | With `weight = W` |
|---|---|---|
| `additive(A)` | $\sigma_A^2$ | $W^2\sigma_A^2$ |
| `proportional(P)` | $(f\sigma_P)^2$ | unchanged → **rejected** |
| `combined(P, A)` | $(f\sigma_P)^2 + \sigma_A^2$ | $(f\sigma_P)^2 + W^2\sigma_A^2$ |

That identity is what makes the front end small: the modifier is *exactly* a per-observation residual magnitude (#484) on the additive sigma slot, so it **desugars into that channel** and reuses its expression compiler, `Dual1` θ-derivative program, and diagnostics wiring. No new engine math on the FOCE/FOCEI path, no new `CompiledModel` state, and `DV`, `PRED`, `IPRED`, `CWRES`, `sdtab` and the VPC all stay on the natural scale with no back-transformation.

**Deliberate deviation from the issue's proposal:** σ is **not** implicitly fixed. The issue asks for "fix sigma to 1 implicitly, or error if the user also declares it free", but its own second example needs σ *estimated* — the weighted logit of a responder proportion, where σ absorbs the $p(1-p)$ factor that was never divided out ($1/\sqrt{p(1-p)}\approx 2.04$ at $p\approx0.4$, most of the published 2.466). Both conventions are real, so `weight =` stays orthogonal to σ and the docs spell out which case wants which. The "three sites must agree" bug it was meant to prevent is gone anyway — there is only one site now. *(Confirmed with @roninsightrx.)*

### 2. Every estimator now applies it

The first cut inherited #484's gate: magnitudes were rejected for anything but `foce`/`focei`. The reason was real — SAEM's M-step, the Gauss-Newton BHHH gradients, the importance-sampling likelihood and proposal precision, and the IOV individual likelihood (SAEM's E-step, the Bayes MH target, the IOV IS weights) all read the residual variance through call sites that never applied the multiplier. Allowing them would have fitted a silently mis-specified error model.

All of them now thread it, and **the gate is gone**. FOCE, FOCEI, Laplace/AGQ, SAEM, IMP, IMPMAP, GN, GN-hybrid and Bayes score the same weighted likelihood, so `method =` chooses the algorithm rather than the model.

| Path | What was reading an unscaled variance |
|---|---|
| `likelihood::individual_nll_iov` | Bayes MH target, SAEM E-step, IOV IS weights |
| `saem` | `obs_nll_subject_into_iov`; both M-step gradient paths (IOV + non-IOV): `V`, `∂V/∂f`, `∂V/∂log σ` |
| `importance_sampling` | the IOV draw loop's obs NLL; both proposal-precision builders (diagonal + `block_sigma` dense) |
| `gauss_newton` | `R`, `∂R/∂f`, `∂²R/∂f²`, `∂R/∂log σ` in both closed forms (Almquist Laplace + Sheiner–Beal) |

**Gradients.** A magnitude built from covariates and `TIME` alone — every `weight =` model — is a per-observation constant, so scaling the variance and its derivatives leaves the analytic gradients exact. A **θ-dependent** magnitude adds a direct `∂R/∂θ` channel that a chain rule through the prediction alone omits:

- **SAEM's M-step** now forward-differences the whole observation likelihood when a magnitude is active, capturing both channels. It costs nothing extra — the prediction solve per θ was already being paid, and re-evaluating the magnitude bytecode is negligible beside it. Magnitude-free models keep the exact legacy chain and are numerically unchanged.
- **Gauss-Newton's** closed forms have no place to put that term (it enters `log|H̃|` as well as the data term), so a θ-dependent magnitude routes those subjects to GN's FD fallback — which differences the magnitude-aware marginal and is therefore still exact, just slower. This follows the precedent the FOCE/FOCEI path already sets for its out-of-scope combinations.

New shared primitives, all in `residual_error.rs` (the single owner of this math): `dvar_dlogsigma_scaled`, `compute_r_diag_maybe_scaled`, and `gaussian_obs_nll_scaled` — the last so SAEM can evaluate the same NLL formula at a perturbed point instead of duplicating it. `RuvMagnitude` gains a `theta_dependent` flag (`#[non_exhaustive]`, set from the θ set `build_ruv_magnitude` already collects).

### 3. Three pre-existing #484 bugs found and fixed

All three would have made `weight =` silently wrong:

- **A covariate referenced *only* from an `[error_model]` expression was not registered as model-referenced**, so `prune_irrelevant_tv_covariates` dropped the per-record snapshots for any subject whose only time-varying column was that one — and every observation was then scored with the subject's *first* value of it, with no diagnostic. Caught by the end-to-end smoke run below (IWRES used `WPSE=0.5` for all three rows of subject 1).
- **A zero / negative / non-finite magnitude is now rejected** (`E_RUV_MAGNITUDE_NONPOSITIVE`). The variance is proportional to its square, so a zero magnitude — an MBMA arm with no reported standard error, whose covariate evaluates to `0` — collapsed that row's variance onto the `1e-12` floor and let one observation own the whole objective, while reporting a converged fit.
- **The Sheiner–Beal `dr` chain rule hard-coded `2·σ_p²·pred·dpred`** in its `Combined` branch, dropping the proportional slot's magnitude factor. Caught by the new GN test once it was strengthened to carry a magnitude on *both* slots — with a weight alone `m_prop ≡ 1` and the scaled and bare slopes coincide, which is exactly the sort of vacuous pass the second assertion in each new test now rules out.

## Alternatives considered

- **Divide `DV` and every prediction at the boundary** (the literal reading: transform at load like LTBS does, back-transform at output). Gives a bit-identical OFV, but touches every prediction path including the `Dual2` sensitivity kernels, and puts the natural-scale reporting burden on the output layer. Rejected: much larger blast radius for a constant-factor difference.
- **Scale the variance uniformly by `W²`** on all slots. Simpler, but wrong for proportional/combined — it weights the constant-CV component in the opposite direction from what "divide both sides" means.
- **Subtract the $\sum 2\ln W_j$ Jacobian to match a hand-built OFV exactly.** Would require touching ~8 scattered data-term accumulation sites; missing one silently desynchronises the objective between estimators. Not worth it for a constant that cancels out of every comparison (see below).
- **Hand-adding a direct-`∂R/∂θ` term to the GN closed forms.** It enters `log|H̃|` too, so it means recomputing the whole `H̃` assembly per θ. Routing to the existing magnitude-aware FD fallback is the precedent the FOCE/FOCEI path already sets and costs nothing for the θ-free case that actually matters here.

## Cross-repo dependency

No `pub` API change beyond additive items (`RuvMagnitude::theta_dependent` on a `#[non_exhaustive]` struct, `CompiledModel::has_theta_dependent_ruv_magnitude`). New DSL syntax; one removed diagnostic code (`E_RUV_MAGNITUDE_METHOD_UNSUPPORTED`, which only ever rejected).

## Breaking changes
- [x] None

Behaviour that previously **errored** now runs (magnitudes under SAEM/IMP/GN/Bayes). Existing magnitude-free fits are numerically unchanged — every new branch is gated on `ruv_mult.is_some()` / `has_custom_ruv_magnitude()`, and the bare-sigma association is preserved rather than merged into the scaled one.

## Numerical validation

**Degenerate oracle (Tier 1, `stats::likelihood`).** The weighted objective is pinned against the hand-built construction it replaces — same rows, `DV/W` and `PRED/W` pre-divided, scored by an unweighted additive model on the same σ:

```
nll_weighted == nll_hand_built + Σ_j ln(W_j)     (relative epsilon 1e-12)
```

This is the exact internal analogue of a NONMEM dose-bookkeeping anchor: the thing `weight =` claims to reproduce, reproduced bit-for-bit up to the documented data-only Jacobian.

**Objective-function offset.** ferx scores on the natural scale (that is what keeps the diagnostics readable), so its OFV differs from a hand-built `DV/W` run by $\sum_j 2\ln W_j$. That term is data-only: identical for every model fitted to the same rows with the same weight column, so it cancels out of every ΔOFV, LRT and AIC comparison. Only the absolute number differs from a published figure. Documented, and pinned by the test above.

**Analytic-vs-FD gradient parity (Tier 1).**

| Harness | Fixture | Checks |
|---|---|---|
| `estimation::sens_outer_gradient` | `combined(...) weight = 1/sqrt(NARM)` | analytic `subject_theta_gradient` / `subject_sigma_gradient` vs Richardson-reconverged FD of the marginal NLL, plus the packed gradient |
| `estimation::saem` | θ-free weight, and a θ-dependent magnitude | M-step `(nll, grad)` vs `obs_nll_sum` and its forward difference |
| `estimation::gauss_newton` | magnitude on **both** slots; Laplace and Sheiner–Beal; θ-dependent via the FD fallback | `subject_nll_pop_grad` vs central FD of `subject_nll_at` |

The θ-dependent SAEM test **fails without the direct-θ fix** — the gradient for the magnitude-only θ reads `0.000000e0` against a true `5.669090e-1`. That is the precise silent-wrong-gradient failure this closes.

**Cross-estimator agreement (Tier 2, `tests/weighted_error_all_methods.rs`).** All nine estimators fit a weighted model and return a finite OFV; and each one scores a varied weight column differently from a constant one, so a method that quietly ignored the multiplier fails rather than passing vacuously.

**End-to-end.** A 3-subject weighted fit via the CLI, checking IWRES row by row against `(DV − IPRED)/(WPSE·σ)`:

```
ID,TIME,DV,PRED,IPRED,CWRES,IWRES,...
1,1,9.100000,9.642129,9.576220,-1.014078,-0.952440   # (9.1-9.576220)/0.5 = -0.952440  ✓
1,4,6.200000,6.555028,6.377628,-0.375415,-0.222035   # (6.2-6.377628)/0.8 = -0.222035  ✓
1,8,4.100000,3.918412,3.709192, 0.135654, 0.325673   # (4.1-3.709192)/1.2 =  0.325673  ✓
```

`DV`, `PRED`, `IPRED` all on the natural scale, no back-transformation. (Before the TV-covariate fix, all three rows divided by `0.5`.)

**No NONMEM comparison.** There is no NONMEM run to anchor against as a single object here: NONMEM has no `weight =`, and its equivalent *is* the hand-built three-site construction — which is precisely what the degenerate oracle compares against, exactly, in-process. Anchoring against a NONMEM run of the hand-built model would test the hand-built model, not this feature.

- [x] Compared against: the hand-built weighted construction (exact, Tier-1 oracle) + Richardson FD + cross-estimator agreement
- [ ] Warfarin example converges — N/A; no shared estimation path changes for a magnitude-free model

## Performance
- [x] No significant impact expected. Every new branch is gated on an active magnitude, so bare-sigma models take the same code path as before. For a magnitude-active model: SAEM's M-step θ loop adds one magnitude-bytecode evaluation per θ per subject (negligible beside the prediction solve it already does), IMP hoists the multiplier out of its draw loop (computed once per subject, not once per sample), and a θ-dependent magnitude under GN drops to the FD fallback — slower, and the only case where it is.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib --all-features` passes: 3309 tests)

31 new tests:

| Where | What |
|---|---|
| `parser::model_parser` (×16) | grammar (`weight = WPSE`, `weight = 1/sqrt(NARM)` with its own parens); additive-slot targeting; combined leaves the proportional slot bare; composition with an existing #484 magnitude; rejection on proportional-only / per-CMT / inside `if/else` / empty RHS / η-dependence / undeclared covariate; `split_weight_modifier` unit tests (a `WEIGHT` covariate *inside* the sigma parens, `weight ==`, `weighted =`, no statement before it); covariate registration + TV-snapshot survival |
| `stats::likelihood` (×2) | degenerate oracle vs the hand-built construction; `individual_nll_iov` vs `individual_nll` under a within-subject-varying weight |
| `estimation::sens_outer_gradient` (×1) | analytic θ/σ gradient vs reconverged FD, weighted combined model |
| `estimation::saem` (×2) | M-step nll + gradient vs `obs_nll_sum` and FD, θ-free and θ-dependent |
| `estimation::gauss_newton` (×3) | Laplace, Sheiner–Beal, and FD-fallback routing |
| `estimation::importance_sampling` (×1) | proposal precision vs a hand-built `Ω⁻¹ + Jᵀ R_scaled⁻¹ J` |
| `api::validation` (×5) | `E_RUV_MAGNITUDE_NONPOSITIVE`: positive passes; zero / negative / non-finite rejected; no-magnitude model not checked |
| `tests/weighted_error_all_methods.rs` (×2, Tier 2) | all nine estimators fit; each scores varied ≠ constant weights |

- [x] Inline below

```
[parameters]
  theta TVCL(0.2, 0.01, 10.0)
  theta TVV(10.0, 1.0, 100.0)
  omega ETA_CL ~ 0.09
  sigma ADD_ERR ~ 1.0 (variance) FIX     # inverse-variance weighting: sigma is the unit

[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV

[structural_model]
  pk one_cpt_iv(cl=CL, v=V)

[error_model]
  DV ~ additive(ADD_ERR) weight = WPSE   # WPSE = this arm's reported standard error

[covariates]
  WPSE continuous
```

```
--- SIGMA Estimates (additive) ---
  ADD_ERR [FIX]        = 1.000000  SE = ---
```

with `sdtab` in natural units (excerpt above).

## Docs
- [x] `docs/` updated — new "Residual weighting (`weight = ...`)" section in `docs/model-file/error-model.qmd` (semantics + the variance table, the three-site construction it replaces, the is-sigma-fixed question with both MBMA cases, the OFV constant, and the restriction list). The existing magnitude section now documents estimator coverage, the direct-θ channel, and `E_RUV_MAGNITUDE_NONPOSITIVE`.
- [x] `CHANGELOG.md` `[Unreleased]` — two `Added` entries (the DSL, the all-estimator support), two `Fixed` entries for the #484 bugs.
- [x] No new pages, so no `_quarto.yml` change.

## Checklist
- [x] `cargo clippy --all-targets --all-features` clean (the 4 pre-existing `approximate value of PI` errors in `src/sens/two_cpt_explicit.rs` are untouched by this PR)
- [x] `Dual2` sensitivity path untouched — the weight is η-independent and rides the existing `RuvMagnitude` channel
- [ ] `Cargo.toml` version bump — not needed (additive, non-breaking)

## Reviewer hints

- **The one line of real math** is the additive-slot targeting in `parse_error_model`: `Combined`'s argument order is `(proportional, additive)`, so the weighted slot is index 1. If that identity ($W\cdot\text{loading}(f/W)$ = `W` on the constant loading, `f` on the proportional one) is wrong, everything else is wrong.
- **`split_weight_modifier`** does a depth-0 scan rather than a regex, because a weight expression carries its own parentheses (`1/sqrt(NARM)`) and a magnitude expression could contain a covariate literally named `WEIGHT`. Four unit tests pin those cases.
- **The estimator work is mechanical except for two judgement calls**, both worth a look: SAEM's M-step switching to a full forward difference when a magnitude is active (correct and free, but it *is* a different numerical path for those models), and GN routing θ-dependent magnitudes to FD rather than growing a direct-`∂R/∂θ` term.
- **Worth a second look:** whether rejecting `weight =` on a purely proportional model is the right call versus warning. I think erroring is right — the modifier would provably do nothing there — but it is a judgement call.
- Skimmable: `git show --stat` is mostly tests and docs; the desugaring means there is no new variance formula on the FOCE/FOCEI path to review.

## Open questions

1. **Should σ be auto-declared?** A `weight =` model whose additive sigma is not declared in `[parameters]` could implicitly get `σ = 1 FIX`, making the MBMA model two lines shorter. Left out (it needs parameter-vector injection and reports a σ the user never wrote), but it is the closest remaining thing to the issue's "fix sigma implicitly" bullet.
2. **`block_sigma` + magnitude under SAEM/IMP.** Correlated residual error already routes to its own FD paths, so it is correct, but I have not added a magnitude × `block_sigma` cross test. Worth a follow-up issue if that combination is expected in practice.

https://claude.ai/code/session_01SHMoKdDEyT1qLvywcECnyf


roninsightrx:
1. Yes, should make implicit assumption that \sigma = 1.
2. Yes, should be threaded through all estimation methods.

Issue #1051, opened 2026-08-23
diagnostics: the E_*/W_* code registry has drifted — 62 of 98 emitted codes are unregistered, and nothing enforces the rule


TeunP:
## Summary

`src/diagnostics.rs`'s module doc states a rule:

> Codes are stable identifiers (prefix `E_` for errors, `W_` for warnings).
> **Add new codes here and document them in `docs/file-formats/check-report.qmd`.**

Both halves have drifted far enough that neither table can be trusted as an authority.

Measured on `worktree-worktree-1001-sigma-order` (same result on `main` modulo this branch's one row):

| | count |
|---|---:|
| distinct `E_*` / `W_*` codes actually emitted in `src/**/*.rs` | **98** |
| rows in `src/diagnostics.rs`'s registry table | 29 |
| rows in `docs/file-formats/check-report.qmd` | 36 |
| emitted but **missing from the in-code registry** | **62** |
| emitted but **missing from the docs table** | **55** |
| registered but never emitted | 0 |

Reproduce:

```bash
git grep -hoE '"(E|W)_[A-Z0-9_]+"' -- 'src/**/*.rs' | tr -d '"' | sort -u > /tmp/emitted
grep -oE '`(E|W)_[A-Z0-9_]+`' src/diagnostics.rs | tr -d '`' | sort -u > /tmp/registered
comm -23 /tmp/emitted /tmp/registered | grep -vE '^(E|W)_[A-Z]$'   # single letters are test fixtures
```

Whole families are absent from both: `E_ABSORPTION_*` (7), `E_BLOCK_SIGMA_*` (7), `E_AGQ_*`,
`E_BINARY_*`, `E_CTMM_*`, `E_DOSE_CMT_*`, `E_TRANSIT_*`, `E_MODELED_{RATE,DURATION}_*`,
`W_DERIVED_CMT_*`, and more. The two tables have also drifted **from each other** — eight codes
including `E_DOSE_ATTR_DOUBLE_USE` (#993), `E_NONFINITE_DV` and `E_SAEM_NO_RANDOM_EFFECTS` are in
the docs table but not in the in-code registry.

## Why it matters

The module doc points external consumers at the registry explicitly — *"so external callers (coding
agents, language bindings) can act on validation output programmatically instead of regex-matching
prose."* A consumer that enumerates the table and treats anything else as unknown silently
mishandles 62 of the 98 codes `ferx check` can emit. Nothing enforces the rule, so every PR that
[...]

PR #1055, opened 2026-08-24
fix(parser): accept a dose attribute read by an [odes] init(...) seed [closes #1046]

### LINKED ISSUE #1046: [odes] init(state) reading F/LAGTIME is wrongly rejected as a dose-attribute double use — the seed applies it once
## Summary

#993 folds `[odes] init(state) = <expr>` reads into the same read-set as the `[odes]` RHS, so an initial condition that references `F` / `LAGTIME` / `ALAG` is rejected with `E_DOSE_ATTR_DOUBLE_USE`. It is not a double use: the ODE engine seeds the state with the raw expression value and applies dose attributes only at dose events, so the attribute is applied **once**.

This is the ODE twin of the defect found reviewing #1035 (the analytical half of the same rule), where the `[initial_conditions]` arm was removed for exactly this reason before merge.

## Measured

```ferx
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  F  = TVF

[structural_model]

[odes]
  init(central) = F * 100.0
```

```
[odes]: `F` is this model's bioavailability — the engine already scales each dose
amount by it — but it is also read in [odes], so the value is applied twice: once
at the dose, once where you read it (#993). If `F` is meant to be bioavailability,
remove it from [odes]; if it is meant to be an ordinary parameter, rename it —
`F` is a reserved dose-attribute name.
```

## Why it is not a double use

`OdeSpec::initial_state` (`src/ode/predictions.rs:1820`) is the whole of the seeding path:

```rust
pub fn initial_state(&self, params: &[f64]) -> Vec<f64> {
    match &self.init_fn {
        Some(f) => f(params),
        None => vec![0.0; self.n_states],
    }
}
```

The `init(...)` expression value goes straight into the state vector. `F` and lag are resolved through `DoseAttrMap` at dose events (`resolve_f` / `resolve_lag`), which the seed never passes through. So the dose term carries `F` once and the init term carries `F` once — **different quantities**, not the same one applied twice.

`init(central) = F * 100` is a meaningful model: the bioavailable residue of a pre-study 100 mg dose. It is currently unwritable without renaming the parameter, which is exactly the wrong remedy for a model whose `F` really is bioavailability.

### The "hole the size of its own remedy" argument does not hold

`build_ode_spec` justifies pulling init reads in with:

> a user told to drop `F` from the RHS could move it into `init(central) = F * AMT` and get the same silent double application from the same block.

They do not get the same double application. `d/dt(central) = ... F ...` folds `F` into the flux, so every gram that ever entered the compartment gets multiplied by `F` a second time — that is the genuine `F²` defect #993 was written for, and it must stay rejected. An initial condition is one additive quantity beside the dose-driven solution; scaling it by `F` scales that quantity and nothing else.

The analytical engine states the same contract explicitly (`src/pk/mod.rs:414`, `analytical_init_concentration_g`):

> Reuses the generic single-dose closed forms with `T = f64` and **`F = 1`** (an initial condition is not an absorbed dose, so bioavailability does not apply).

The ODE engine behaves identically; it just has no comment saying so.

## Same defect on lag

`init(central) = LAGTIME * 100` is rejected on the same path, and the seed is deposited at the record start with no time shift — so lag is likewise applied once.

## Scope sweep

| Sweep target | Affected models |
|---|---|
| `examples/*.ferx` | **0** (only `adaptive_platelet_ladder.ferx` and `one_cpt_oral_baseline.ferx` use `init(`, neither reads a dose attribute) |
| inline models in `src/**/*.rs` | **1** — `dose_attr_read_in_odes_init_is_rejected` (`src/parser/model_parser_tests.rs:1783`), the fixture that pins the wrong behaviour |
| inline models in `tests/*.rs` | **0** |

So no shipped model breaks either way; the fix is a widening, not a break.

## Suggested fix

`build_ode_spec` currently merges both sources into one set:

```rust
visit_stmt_nodes(&stmts_owned, &mut collect_reads);
for (_, init_expr) in &init_specs {
    visit_expr_nodes(init_expr, &mut collect_reads);
}
check_dose_attr_double_use(&indiv_names_owned, &indiv_slots_owned, &rhs_reads, ...)?;
```

So this is not a one-line deletion — the init walk has to be dropped from `rhs_reads`, not the call. Steps:

1. Stop folding `init_specs` into `rhs_reads`; check only the RHS.
2. Leave the `D{n}`/`R{n}` prediction-path recording off init too, for the same reason (an init amount consults no dose attribute at all, modeled duration/rate included). Worth checking whether `[odes]` init currently feeds `mark_prediction_path_read`; the analytical side did and it was removed in #1035.
3. Flip `dose_attr_read_in_odes_init_is_rejected` to an acceptance test covering both `F` and `LAGTIME`, citing `initial_state`'s contract.
4. Add the "not a check site" note to `build_ode_spec`, mirroring the one now at the analytical `[initial_conditions]` block.
5. Docs: `docs/model-file/ode-models.qmd` (the "Reading a dose attribute is an error" section) needs an init carve-out, and `docs/file-formats/check-report.qmd`'s `E_DOSE_ATTR_DOUBLE_USE` row should stop implying init is covered on either engine.

## Notes

- The `[odes]` **RHS** half of #993 is correct and stays.
- Behaviour is unchanged for every model in the repo (sweep above), so this can land as a plain `Fixed` CHANGELOG entry rather than a breaking one — it only *accepts* models that previously failed.
- Related: #1004 / #1035 (analytical half, where this arm was caught in review and removed), #993 (the original rule).

TeunP:
## Why

#993 folds `[odes] init(state) = <expr>` reads into the same read-set as the `[odes]` RHS, so an initial condition that references `F` / `LAGTIME` / `ALAG` is rejected with `E_DOSE_ATTR_DOUBLE_USE`. **It is not a double use.**

`OdeSpec::initial_state` (`src/ode/predictions.rs`) is the whole of the seeding path — it writes the raw expression value into the state vector. `F` and lag are resolved through `DoseAttrMap` at dose events (`resolve_f` / `resolve_lag`), which the seed never passes through. So the dose term carries `F` once and the init term carries `F` once, on two **different** quantities.

`init(central) = F * 100` — the bioavailable residue of a pre-study 100 mg dose — is a meaningful model. It was unwritable, and the diagnostic advised the one repair that is always wrong for it: renaming the parameter whose meaning *is* bioavailability.

This is the ODE twin of the defect found reviewing #1035, where the analytical `[initial_conditions]` arm was removed for exactly this reason before merge. The two engines now agree.

The `[odes]` **RHS** half of #993 is correct and stays: `d/dt(central) = … F …` folds `F` into the flux, so every gram that ever entered the compartment is multiplied by it a second time — the genuine `F²` defect.

## What changed

Stopped folding `init_specs` into `rhs_reads` in `build_ode_spec` (3 lines), and replaced the comment that justified it.

The fix had to drop the **init walk** rather than skip the check: that same `rhs_reads` set also drives `mark_prediction_path_read` further down, whose marks `check_modeled_dose_rates` turns into a data-gated error once a `RATE=-2` dose lands on the compartment. Skipping only the parse-time check would have left a `D{n}` init read recorded, so a correct model would still have been rejected as soon as its dataset happened to carry a coded RATE.

## Alternatives considered

Keeping the rejection and telling users to rename. Rejected: the whole point of the reserved name is that bioavailability legitimately *is* `F`, so renaming is not a repair, it is a workaround that makes the model lie about itself. The measurement below shows the reference implementation does not treat it as a double use either.

## Cross-repo dependency

No public Rust API change and no `.ferx` format change: this only *accepts* models that previously failed to parse.

## Breaking changes
- [x] **None** — pure widening. Verified no shipped model changes behaviour: the only two `examples/*.ferx` using `init(` read `CIRC0` and `CONC0 * V` (no dose attribute), and ferx-r's `origin/main` bundles no `.ferx` with an `init(`.

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0** (`ADVAN13 TOL=9`, `MAXEVAL=0`, all `$THETA` `FIX`, `$OMEGA 0 FIX`)
- [x] Reference values:

Four new control streams. Each **pair** holds the seeded *value* fixed and varies only whether the name is a dose attribute — `A_0(1) = F1*100` versus an ordinary `FSEED` of equal value — which isolates the one question under test: does an initial condition treat `F1`/`ALAG1` as an attribute, or as a number?

```
   t   F-A / F-B    LAG-A / LAG-B
 0.0   1.00000      1.40000
 0.5   0.95123      1.33170
 1.0   1.90480      1.26680     <- dose record; F: +50, LAG: no jump yet
 1.5   1.81190      1.20500
 2.0   1.72360      3.08710
 4.0   1.41110      2.52750
 8.0   0.94591      1.69420
```

Three facts, each independently readable off that table:

1. **The seed is the raw value.** `IPRED(0) = 1.0 = (F1·100)/V = 50/50`. Had NONMEM applied bioavailability to the seed it would read `25/50 = 0.5`. For lag, `IPRED(0) = 1.4 = 70/50`, deposited at t=0 **unshifted** — a lagged seed would leave the compartment empty until t=0.7.
2. **The dose still gets the attribute, exactly once.** The F run jumps by `50 = F1·100` at t=1; the LAG run shows *no* jump at t=1 or t=1.5 because the dose lands at t=1.7, which is where the t=2 value comes from.
3. **A and B are byte-identical** (`.tab` files compare equal; `#OBJV` 575.124 for the F pair, 579.151 for the lag pair). The reference engine treats a dose-attribute name in an initial condition exactly as it treats an ordinary parameter of the same value — precisely ferx's contract.

ferx reproduces both tables to `< 1e-4` relative (`tests/odes_init_dose_attr_nonmem_anchor.rs`).

> **NONMEM detail worth recording.** NONMEM deposits `A_0(n)` at the subject's **first data record**, not at t=0. An earlier draft of these streams had its first record at t=0.5 and the seed landed there undecayed — which would have made the ferx comparison mismatch for a reason unrelated to this issue. The committed dataset carries an explicit t=0 observation so NONMEM's convention coincides with ferx's (`initial_state` at record start), and the seed is readable directly as `IPRED(0) = seed/V`.

## Performance
- [x] No significant impact expected — strictly less work at parse time (one walk removed).

## Tests

- [x] Unit test(s) added in the same module — `cargo test --lib --features ci`: **3208 passed, 0 failed, 19 ignored**, measured on this branch *merged with* `origin/main` @ `6c766518` (which moved after the PR was opened), together with `cargo check --tests`, `cargo fmt --check` and `cargo clippy --all-targets` (no new diagnostics)

**8 of the 10 new tests fail** with the fix reverted — re-measured at the review-round head, not just at the first commit, by re-adding the deleted walk, re-running, and restoring (`md5` compared against a pre-mutation snapshot). The two that pass in both states do so **by design**, and are marked below: the overshoot guard pins that the RHS rejection survives, and the census test pins the behaviour of an *ordinary* name, which this change does not touch.

| Test | Pins |
|---|---|
| `dose_attr_read_in_odes_init_is_accepted` | all six spellings (`F`, `LAGTIME`, `ALAG`, `F1`, `ALAG1`, `LAGTIME1`) parse, and produce no "computed but never used" warning |
| `dose_attr_read_in_both_odes_init_and_rhs_is_still_rejected` | overshoot guard — the RHS walk is intact |
| `coded_rate_param_read_in_odes_init_is_not_recorded` | a `D1` init read records no prediction-path mark; a `D1` **RHS** read still does |
| `ferx_matches_nonmem_when_an_init_seed_reads_bioavailability` | ferx vs NONMEM, F |
| `ferx_matches_nonmem_when_an_init_seed_reads_the_absorption_lag` | ferx vs NONMEM, lag |
| `seeding_from_a_dose_attribute_equals_seeding_from_an_ordinary_parameter` | bit-identical twin (stricter than the 1e-4 anchor bound) |
| `an_init_seed_reading_a_dose_attribute_survives_a_system_reset` | the same, across an `EVID=3` re-seed |
| `analytical_and_ode_init_agree_when_the_seed_reads_bioavailability` | cross-engine: #1035's carve-out and this one give the same answer |
| `ode_provider_init_reading_a_dose_attribute_matches_fd` | `Dual2`/`Dual1`-vs-FD parity for the newly reachable `∂init/∂F` and `∂init/∂lag` (see the checklist below) |
| `an_init_read_counts_as_a_use_for_the_never_used_census` | an init read of an **ordinary** name counts as a use — *passes in both states by design* |

- [ ] Not applicable

<details>
<summary>Example</summary>

```toml
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV
  F  = TVF

[structural_model]

[odes]
  init(central) = F * 100.0        # bioavailable residue of a pre-study 100 mg dose
```

With `F = 0.5`, `CL = 5`, `V = 50` the seed is `50` and `PRED(0) = 1.0`. Doses in the dataset are still scaled by `F` exactly once. Previously this was a parse error.

</details>

## Docs
- [x] `docs/` updated for user-visible changes — `docs/model-file/ode-models.qmd` (the "Reading a dose attribute is an error" section claimed init was covered; the section lead, the verbatim message sample and a new callout now carry the carve-out and the NONMEM result), `docs/model-file/individual-parameters.qmd` (the reserved-names callout still named the `init(...)` seed as a rejected read *and* prescribed renaming — the one repair #1046 exists to eliminate; raised as #1057 in review) and `docs/file-formats/check-report.qmd` (`E_DOSE_ATTR_DOUBLE_USE` row states the exemption)
- [ ] New pages linked in `docs/_quarto.yml` — no new pages
- [x] `CHANGELOG.md` `[Unreleased]` entry — the #993 entry's own text claimed init seeds were rejected, which is now false. #993 has not shipped in a tagged release, so that entry is **corrected in place** (with the exemption and the NONMEM result) rather than contradicted by a separate `Fixed` line.
- [ ] No user-visible change

## Checklist
- [x] `cargo clippy --all-targets` clean — measured, not argued: no new diagnostics, and `check_dose_attr_double_use` is not among the `too many arguments` reports (the seventh parameter sits at the threshold, not over it)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — **parity test added** (`src/sens/ode_provider_tests.rs::ode_provider_init_reading_a_dose_attribute_matches_fd`).
  An earlier revision of this checklist said no parity test was required because no kernel is touched. That was true of the *diff* and false of the *reachable model set*, which is what CLAUDE.md's rule is about: `∂init/∂F` and `∂init/∂lag` could not be reached before this change, so the seed derivative had never had to compose with the `F`-scaled dose bolus or the lagtime saltation. Raised as #1059 in review and fixed here — three ODE fixtures (`init` reads `F`, reads `LAGTIME`, reads an ordinary parameter as control), each with IIV on the attribute, run through `check_vs_production`, `check_inner_outer_eta_parity` and `check_hessian_vs_production_fd`, plus non-vacuity asserts that `max|∂f/∂θ|` and `max|∂f/∂η|` are actually non-zero (otherwise the parity would agree on an identically-zero derivative and pin nothing).
  The fixtures pin `ode_reltol = 1e-10` / `ode_abstol = 1e-12`, which is **load-bearing**: at default tolerances the FD *reference* is under-resolved and a correct model misses the harness's `2e-4` bound. It passed at defaults on macOS, so this would have been invisible locally and surfaced only in CI (cf. the documented macOS-vs-Linux flip in `check_full_provider_vs_fd_with_step`).

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key — no new syntax; an existing construct stops being rejected
- [ ] New example `.ferx` file — none added
- [ ] New data file — none added

### ferx-book

### Example execution
- [x] **No example execution step needed** — no `examples/*.ferx` is affected. Verified by sweep: the only two using `init(` seed from `CIRC0` and `CONC0 * V`, neither a dose attribute. Since the change only accepts previously-failing models, no currently-passing example can shift.

## Reviewer hints

Focus on `src/parser/model_parser.rs` — the deletion is 3 lines; the rest of that hunk is the replaced rationale.

The subtle part is **why the deletion, not a skipped check**: `rhs_reads` has two consumers, the parse-time `check_dose_attr_double_use` and the `mark_prediction_path_read` loop ~330 lines later. `coded_rate_param_read_in_odes_init_is_not_recorded` is the test that pins the second one, in both directions.

The rest can be skimmed — anchors, docs, CHANGELOG.

Worth knowing: the reset test's non-vacuity guard initially asserted the wrong thing. The post-reset value at t=1.5 is *bit-identical* to t=0.5, because both sit 0.5 h after an identical seed. That coincidence is why it compares against a no-reset arm rather than checking monotonicity within one trajectory.

## Review round

Reviewed at `xhigh`, which produced #1057 (docs), #1058 (diagnostic wording) and #1059 (parity) — all three fixed in `6c27cb99` — plus #1060, a pre-existing wrong-gradient bug that is **not** caused by and not fixed by this PR.

#1058 is worth reading as a design point. With the init walk gone, only an **RHS** read is a double use, but the message still said "read in `[odes]`" / "remove it from `[odes]`" — so on a model that reads the attribute in *both* places, a user following the message literally would delete the (legal) init read and get the identical error back. `check_dose_attr_double_use` now takes a separate `read_site` (`"the [odes] RHS"`) for the two "where" clauses, while `block` stays the `"{block}:"` prefix that `parse_error_to_diagnostic` and the `expect_parse_err` assertions key on.

The review also caught **two of the original tests being vacuous**, both reproduced before fixing:

- The "computed but never used" assertion could not fail. The census exempts all six dose-attribute names via `RESERVED_PK_SLOTS` / `from_indexed_name` *before* it reads block text, so it passed with the `init(...)` line deleted entirely. The real test needs an **ordinary** name, with a delete-the-line arm as its own non-vacuity check.
- The reset test's `assert_ne!` also passes for a reset that *zeroed* the compartment without re-seeding (`0.0 != 0.8607`), and the A/B twin agrees on two zeros. Measured `a[1] == a[2]` bitwise — that equality is the actual discriminator and is now asserted alongside.

## Open questions

`E_DOSE_ATTR_DOUBLE_USE` is still absent from the in-code registry in `src/diagnostics.rs` (it is in the docs table only) — that is #1051's tracked drift across 61 codes and is deliberately not touched here.

One finding from the review is **deliberately left unfixed**, because fixing it now would document a bug. `docs/estimation/foce.qmd` states that `init(...)` combined with an estimated lagtime "still falls back to FD"; measured, a time-varying-covariate + `init(...)` + `LAGTIME` model reports `ode_tvcov_supported = true`, i.e. analytic. But that is exactly the configuration #1060 shows returns a **wrong** analytic gradient (300× on `∂f/∂η`). So the page describes the intended behaviour and the code does not honour it — and if #1060 is fixed by restoring the FD fallback, the page becomes correct again without an edit. Tracked there rather than papered over here. (Only the lagtime clause of that sentence was measured, not its other five.)


src/parser/model_parser.rs
```diff
@@ -10013,21 +10013,35 @@ fn build_ode_spec(
...
     let mut collect_reads = |e: &Expression| {
         if let Expression::Variable(name) = e {
             rhs_reads.insert(name.to_ascii_uppercase());
         }
     };
     visit_stmt_nodes(&stmts_owned, &mut collect_reads);
-    for (_, init_expr) in &init_specs {
-        visit_expr_nodes(init_expr, &mut collect_reads);
-    }
     check_dose_attr_double_use(
```
TeunP, line 10057: **The parse-time diagnostic now over-reports.** With the init walk gone, only an **RHS** read is a double use — but the message this call emits still says the read is "in `[odes]`" and tells the user to "remove it from `[odes]`" (`check_dose_attr_double_use`, model_parser.rs:9083-9089, `block = "[odes]"`).

For the model your own overshoot test builds (`init(central) = F * 100` **and** `d/dt(central) = … F …`) a user who follows that advice literally and deletes the *init* read still gets the same error — while `docs/model-file/ode-models.qmd` now tells them the init read was fine all along.

The data-gated half of the very same diagnostic is already precise about this: `src/api/validation.rs:823` says "`also read on the model's prediction path ([odes] RHS or [scaling])`". The two halves of `E_DOSE_ATTR_DOUBLE_USE` now disagree with each other.

Worth narrowing the `[odes]` wording to name the RHS. Note the `block` argument is also the message's `"{block}:"` prefix and `expect_parse_err` asserts `err.contains("[odes]:")`, so the fix probably wants a separate "where" clause rather than a changed `block` string.

src/parser/model_parser_tests.rs
```diff
@@ -1773,40 +1773,126 @@ fn dose_attr_read_in_odes_rhs_is_matched_case_insensitively() {
...
+        };
+        // And it must not merely parse — the census must not then call the parameter
+        // dead. An `init(...)` line is block text the token census counts, so a
+        // reachable model produces no "computed but never used" advice either.
+        assert!(
+            parsed
+                .model
+                .parse_warnings
+                .iter()
+                .all(|w| !w.contains("never used")),
```
TeunP: **This assertion is vacuous — it cannot fail for any of the six names.** I verified empirically: I parsed all six models with the `init(...)` line *deleted entirely* and every one still produces zero "never used" warnings.

```
NOINIT F: never-used warnings = []
NOINIT LAGTIME: never-used warnings = []
NOINIT ALAG: never-used warnings = []
NOINIT F1: never-used warnings = []
NOINIT ALAG1: never-used warnings = []
NOINIT LAGTIME1: never-used warnings = []
```

The census exempts every one of them before it can look at block text (`model_parser.rs:3430-3441`): `F`/`LAGTIME`/`ALAG` via `ode_slot_map … RESERVED_PK_SLOTS`, and `F1`/`ALAG1`/`LAGTIME1` via `DoseAttr::from_indexed_name(name).is_some()`.

So the comment above it — "An `init(...)` line is block text the token census counts, so a reachable model produces no 'computed but never used' advice either" — attributes the pass to the wrong mechanism, and a regression that stopped counting init lines would not be caught here.

If the intent is to test the census, the seeded name has to be one the exemption does *not* cover (an ordinary `FSEED`, say, read only from the init). If the intent is just "parsing this model is clean", say that and drop the census claim.

tests/odes_init_dose_attr_nonmem_anchor.rs
```diff
@@ -0,0 +1,289 @@
...
+        // t = 0.5 sits 0.5 h after the original seed, so both read
+        // `seed·e^(−k/2)/V` and are bit-identical. That coincidence is *why* this
+        // needs the no-reset arm rather than a monotonicity check on `a` alone.
+        let plain = run(CSV_NO_RESET, attr);
+        assert_eq!(
+            plain.len(),
+            4,
+            "{attr}: four observations without the reset"
+        );
+        assert_ne!(
```
TeunP, line 261: **The non-vacuity guard doesn't discriminate the thing it names.** `assert_ne!(a[2], plain[2])` passes for *any* post-reset value that isn't uninterrupted decay — including an `EVID=3` that zeroes the compartment and never re-seeds (`0.0 != 0.8607`). The A/B twin above would pass on that too, since two zeros are bit-identical. So a regression that dropped the re-seed leaves every assertion in this test green.

The exact discriminator is already named in the comment right above and is one line away. Measured on this branch:

```
RESET F preds = [1.0, 0.951229424500799, 0.951229424500799, 0.9048374180361209]
a[1] == a[2] (bitwise) -> true
```

t = 1.5 sits 0.5 h after the re-seed exactly as t = 0.5 sits 0.5 h after the original seed, so a genuine re-seed makes them bit-identical — while a zeroing reset, or no reset at all, cannot. Adding

```rust
assert_eq!(
    a[2].to_bits(),
    a[1].to_bits(),
    "{attr}: t=1.5 after the re-seed ({}) must equal t=0.5 after the original seed ({})",
    a[2],
    a[1]
);
```

alongside the existing `assert_ne!` pins re-seeding in both directions.

tests/odes_init_dose_attr_nonmem_anchor.rs
```diff
@@ -0,0 +1,289 @@
...
+                .expect("the model parses")
+                .model;
+            predict(&m, &pop, &m.default_params)
+                .into_iter()
+                .map(|p| p.pred)
+                .collect()
+        };
+        let a = run(CSV_RESET, attr);
+        let b = run(CSV_RESET, "SEED");
+        assert_eq!(a.len(), 4, "{attr}: four observations across the reset");
```
TeunP, line 225: `a.len()` is asserted but `b.len()` is not, and the loop below is `a.iter().zip(b.iter())` — a zero-length `b` would skip every comparison and the test would still pass. Its sibling at :278 gets this right with `assert_eq!(a.len(), b.len(), …)` before zipping.

```suggestion
        assert_eq!(a.len(), 4, "{attr}: four observations across the reset");
        assert_eq!(b.len(), a.len(), "{attr}: same number of predictions");
```

tests/odes_init_dose_attr_nonmem_anchor.rs
```diff
@@ -0,0 +1,289 @@
...
+            "{attr}: t=1.5 with the reset ({}) must differ from uninterrupted decay \
+             ({}) — otherwise EVID=3 never re-seeded and the twin above is vacuous",
+            a[2],
+            plain[2]
+        );
+    }
+}
+
+#[test]
+fn seeding_from_a_dose_attribute_equals_seeding_from_an_ordinary_parameter() {
```
TeunP, line 273: This test asserts only that arm A equals arm B — there is no value anchor in it — so it stays green if both arms return all-zeros or all-NaN (`NaN.to_bits() == NaN.to_bits()`). The docstring calls it "the strongest statement of the contract", and it would be, with one anchored value.

The file already has the constants: adding `assert_matches(&a, want, …)` (or even just `assert!(a[0] > 0.0)`) at the top of each loop iteration makes it self-standing rather than dependent on the two anchor tests happening to cover the same models.

tests/odes_init_dose_attr_nonmem_anchor.rs
```diff
@@ -0,0 +1,289 @@
...
+    // at record start *and* to re-seed after an `EVID=3/4` reset, so accepting the
+    // read means the seed expression is now evaluated on the reset path too. Nothing
+    // about a reset consults `DoseAttrMap`, so the attribute must stay an ordinary
+    // value there as well — pinned the same way, against the twin that seeds from an
+    // ordinary parameter of equal value.
+    //
+    // Without this the reset path would be an untested inference rather than a
+    // measured one; it is the only place the accepted expression runs more than once.
+    for (attr, value) in [("F", 0.5), ("LAGTIME", 0.7)] {
+        let run = |csv: &str, init_reads: &str| -> Vec<f64> {
```
TeunP, line 220: This closure re-implements `preds()` above verbatim except for the CSV — temp file, `read_nonmem_csv`, `parse_full_model`, `predict`, map to `pred`. Parameterizing the existing helper removes the copy:

```rust
fn preds_on(csv: &str, src: &str) -> Vec<f64> { /* current preds() body, with `csv` */ }
fn preds(src: &str) -> Vec<f64> { preds_on(CSV, src) }
```

Then the loop body is `preds_on(CSV_RESET, &model(attr, value, attr))`, and the two copies cannot drift (they already differ in the `expect` strings).

docs/model-file/ode-models.qmd
```diff
@@ -348,11 +348,40 @@ remove it from [odes]; if it is meant to be an ordinary parameter, rename it —
 ```
 
 The rule follows the **expressions compiled against your individual parameters**,
-not the block headings: every read inside `[odes]` counts, including an
-`init(state) = …` seed, and so does the `[adaptive_dosing] observe` signal. That
-last one matters most — `observe` is what the `when` rules compare against, so a
-double-applied `F` there biases the titration decision itself and every dose the
-controller goes on to emit, not just a reported number.
+not the block headings: every read in the `[odes]` **RHS** counts, and so does the
```
TeunP, line 353: This narrowing to "the `[odes]` **RHS**" is right, but the section's own lead sentence 13 lines up (`:338`, just outside the diff) still says the unqualified version:

> Declaring `F` *and* referencing it in `[odes]` or `[scaling]` is rejected at parse time with `E_DOSE_ATTR_DOUBLE_USE`

A reader who stops at the heading + first paragraph gets the pre-#1046 rule. Worth the same "`[odes]` RHS" qualifier there, since the callout that corrects it sits below the fold.

TeunP (commented):
## Review — `xhigh`

Built the branch and ran the affected suites: `cargo test --lib dose_attr` (22 passed) and `cargo test --test odes_init_dose_attr_nonmem_anchor --test analytical_init_equivalence` (10 passed). I also re-derived all four NONMEM tables by hand from `CL=5, V=50, k=0.1` — F arm: seed 50, `+50` at t=1, giving `1.0 / 0.95123 / 1.90484 / 1.81194 / 1.72360 / 1.41114 / 0.94591`; lag arm: seed 70 deposited unshifted at t=0, dose landing at t=1.7, giving `1.4 / 1.33172 / 1.26677 / 1.20490 / 3.08712 / 2.52745 / 1.69424`. Both reproduce the committed `.tab` files to the last printed digit, and each `.lst` echoes the `$PK` of the `.ctl` it sits beside.

**The central argument holds.** I traced every `f_bio` / `lagtime` consumer: all of them are dose-application sites (`u[cmt] += f_bio * dose.amt`, `dose.bioavailable_infusion(f_bio)`, the event-time shift). Nothing on the seeding path touches `DoseAttrMap`, so the carve-out is right — and deleting the walk rather than skipping the check is the correct shape, since `rhs_reads` also drives `mark_prediction_path_read`.

One point the description undersells: the hole the deleted comment worried about — `init(central) = F * AMT` — was never expressible. `init_defined` admits only declared states, individual parameters and `MACHEPS`, so `AMT` in an init has always been a parse error for an unrelated reason. The original justification was weaker than it read.

Findings below, most severe first. Two are on lines outside the diff, so they're here rather than inline.

---

### 1. A third docs page still carries the claim this PR removes — `docs/model-file/individual-parameters.qmd:327`

The reserved-names callout says:

> …they never appear in a correct `[odes]` block (RHS or `init(...)` seed), `[scaling]` readout, or `[adaptive_dosing] observe` signal. Declaring one *and* referencing it on the prediction path applies it twice and is rejected with `E_DOSE_ATTR_DOUBLE_USE`… If you need an ordinary parameter, rename it: the name is what routes it.

That parenthetical is the exact claim `ode-models.qmd` and `check-report.qmd` were updated for, and the page goes on to prescribe the repair this PR exists to eliminate. It now also contradicts itself: 17 lines down (`:344`) the same callout documents `init(depot) = F * 500` applying `F` exactly once on the analytical engine. This is the page a user lands on when they look up what `F` means, so it's the likeliest of the three to be read.

### 2. No `Dual2`-vs-FD parity test for the newly reachable gradient class

Before this PR, no model could have `∂init/∂F ≠ 0` or `∂init/∂lag ≠ 0` — the parser rejected them. The analytic ODE provider seeds `init_state` by central FD of `init_fn` over `model.pk_indices` (`src/sens/ode_provider.rs:1377`, called from 1917 / 2023 / 2129) and *separately* carries `∂/∂F` on the dose bolus and `∂/∂lag` on the event-time saltation. Those two contributions have never had to compose before.

I checked, and **the composition is correct** — I wrote three throwaway fixtures against `check_full_provider_vs_fd` (plain bolus, `[fit_options] ode_reltol = 1e-10, ode_abstol = 1e-12`, `η = [0.1, -0.05, 0.07]`):

```
init(central) = F * 100         -> analytic, parity PASS
init(central) = LAGTIME * 100   -> analytic, parity PASS
init(central) = FSEED * 100     -> analytic, parity PASS   (control)
```

So this is a coverage gap, not a defect. But CLAUDE.md is explicit that it's the gap that matters:

> Every change to an analytic sensitivity, gradient, marginal, or likelihood path requires a `Dual2`-vs-FD parity test. … If a model is outside the analytic scope it must route to FD via the support predicates … unit-test that routing so a scope gap fails loudly to FD instead of silently returning a wrong gradient.

The checklist's *"not touched — this is a parse-time check-set change only, no kernel or provider is involved"* is true of the **code** and not of the **reachable model set**. Every `init(...)` fixture in `src/sens/provider_tests.rs` today seeds from an ordinary parameter (`TVC0 * V`, `BASE`, `1000.0 / V`). Since the fixtures above already pass, adding one is cheap and pins a composition that currently nothing does.

*Aside, pre-existing and not caused by this PR:* `docs/estimation/foce.qmd:34` states that "`init(...)` combined with a reset / **estimated lagtime** / finite infusion / input-rate forcing / steady-state / modeled-dose still falls back to FD". The lag fixture above reports `ode_analytical_supported = true` and is served analytically, so that sentence looks stale for at least the lagtime case. It becomes more visible now that a lag-seeded init is writable.

### 3. The parse-time diagnostic now over-reports — see inline on `src/parser/model_parser.rs:10045`

### 4–8. See inline comments on `model_parser_tests.rs`, `odes_init_dose_attr_nonmem_anchor.rs` and `ode-models.qmd`

Two of them are assertions that cannot currently fail (verified empirically, details inline); the rest are a missing length check, a missing value anchor, a duplicated helper, and one unqualified sentence.

---

Nothing here blocks the change — the deletion is correct and the reasoning is sound. #1 and #3 are user-facing and worth fixing before merge; the rest are a line or two each.

TeunP:
Follow-up to my review above — here is everything I actually ran against this branch, so the claims in those comments are checkable rather than asserted. Plus one out-of-scope finding that came out of it and explicitly **should not block this PR**.

## 1. Test suites on the branch

```
cargo test --lib dose_attr
```
22 passed, 0 failed. The ones this PR touches:

```
parser::model_parser::tests::dose_attr_read_in_odes_init_is_accepted ... ok
parser::model_parser::tests::dose_attr_read_in_both_odes_init_and_rhs_is_still_rejected ... ok
parser::model_parser::tests::dose_attr_read_in_odes_rhs_is_rejected ... ok
parser::model_parser::tests::dose_attr_read_in_odes_rhs_is_matched_case_insensitively ... ok
parser::model_parser::tests::dose_attr_read_in_scaling_is_rejected ... ok
parser::model_parser::tests::dose_attr_read_in_adaptive_observe_is_rejected ... ok
parser::model_parser::tests::dose_attr_read_only_for_reporting_is_accepted ... ok
```

Note `coded_rate_param_read_in_odes_init_is_not_recorded` does **not** contain the substring `dose_attr`, so it is **not** in those 22. Flagging in case you lean on that filter locally too — CI runs the whole suite, so this is about local spot-checks, not the branch. It matches `odes_init`, and all three of this PR's new parser tests pass under that filter:

```
cargo test --lib odes_init

test parser::model_parser::tests::dose_attr_read_in_odes_init_is_accepted ... ok
test parser::model_parser::tests::coded_rate_param_read_in_odes_init_is_not_recorded ... ok
test parser::model_parser::tests::dose_attr_read_in_both_odes_init_and_rhs_is_still_rejected ... ok

test result: ok. 3 passed; 0 failed
```

```
cargo test --test odes_init_dose_attr_nonmem_anchor --test analytical_init_equivalence
```
10 passed, 0 failed:

```
ferx_matches_nonmem_when_an_init_seed_reads_bioavailability ... ok
ferx_matches_nonmem_when_an_init_seed_reads_the_absorption_lag ... ok
seeding_from_a_dose_attribute_equals_seeding_from_an_ordinary_parameter ... ok
an_init_seed_reading_a_dose_attribute_survives_a_system_reset ... ok
analytical_and_ode_init_agree_when_the_seed_reads_bioavailability ... ok
analytical_central_init_matches_ode_init ... ok
analytical_2cpt_central_init_matches_ode_init ... ok
analytical_3cpt_central_init_matches_ode_init ... ok
analytical_init_zeroed_after_reset ... ok
predict_iov_carries_analytical_init ... ok
```

## 2. The NONMEM anchors, re-derived by hand

I did not take the committed `.tab` files on trust. With `CL = 5`, `V = 50`, `k = 0.1`:

**F arm** — seed `F1·100 = 50`, dose `+F1·100 = +50` at `t = 1`:

| t | closed form | committed |
|---|---|---|
| 0 | `50/50` | 1.0000 |
| 0.5 | `50·e^-0.05/50` = 0.951229 | 9.5123E-01 |
| 1 | `(50·e^-0.1 + 50)/50` = 1.904837 | 1.9048E+00 |
| 1.5 | `95.2419·e^-0.05/50` = 1.811937 | 1.8119E+00 |
| 2 | `95.2419·e^-0.1/50` = 1.723600 | 1.7236E+00 |
| 4 | `95.2419·e^-0.3/50` = 1.411143 | 1.4111E+00 |
| 8 | `95.2419·e^-0.7/50` = 0.945909 | 9.4591E-01 |

**LAG arm** — seed `ALAG1·100 = 70` deposited unshifted at `t = 0`, dose `+100` landing at `t = 1.7` (F = 1, no `F1` declared):

| t | closed form | committed |
|---|---|---|
| 0 | `70/50` | 1.4000 |
| 0.5 | 1.331715 | 1.3317E+00 |
| 1 | 1.266767 | 1.2668E+00 |
| 1.5 | 1.204901 | 1.2050E+00 |
| 2 | `(70·e^-0.17 + 100)·e^-0.03/50` = 3.087117 | 3.0871E+00 |
| 4 | 2.527450 | 2.5275E+00 |
| 8 | 1.694239 | 1.6942E+00 |

Every digit reproduces. I also checked each `results/*.lst` echoes the `$PK` of the `.ctl` committed beside it, and that the A/B pairs differ only in the seeded name (`diff` is 3 hunks: `$PROBLEM`, the `A_0(1)` line, the `$THETA` comment).

## 3. The two "assertion cannot fail" findings — measured, not inferred

**`model_parser_tests.rs:1838`.** I parsed all six models with the `init(...)` line **deleted entirely**:

```
NOINIT F:         never-used warnings = []
NOINIT LAGTIME:   never-used warnings = []
NOINIT ALAG:      never-used warnings = []
NOINIT F1:        never-used warnings = []
NOINIT ALAG1:     never-used warnings = []
NOINIT LAGTIME1:  never-used warnings = []
```

Zero warnings with no init read at all, so the assertion passes either way. The census exempts all six before it looks at block text (`model_parser.rs:3430-3441`).

**`odes_init_dose_attr_nonmem_anchor.rs:254`.** Predictions across the `EVID=3` reset:

```
RESET F preds = [1.0, 0.951229424500799, 0.951229424500799, 0.9048374180361209]
a[1] == a[2] (bitwise) -> true
```

`a[1] == a[2]` exactly, as the comment predicts. That is the assertion that would pin re-seeding; `assert_ne!(a[2], plain[2])` also passes if the reset merely zeroed the compartment.

## 4. The gradient class this PR newly makes reachable

`∂init/∂F` and `∂init/∂lag` could not be non-zero before #1046. I checked the composition rather than just flagging it — three fixtures through `check_full_provider_vs_fd`, plain bolus, `θ = [10, 50, x]`, `η = [0.1, -0.05, 0.07]`:

| seed | route | parity |
|---|---|---|
| `init(central) = F * 100` | analytic | **PASS** |
| `init(central) = LAGTIME * 100` | analytic | **PASS** |
| `init(central) = FSEED * 100` (control) | analytic | **PASS** |

So this PR is correct on that axis — it is a missing-test issue only (#1059), not a defect.

**One trap for whoever adds the fixture:** at the default `ode_reltol = 1e-4` my first attempt failed at a margin of `2.14e-4` against the harness's `2e-4` bound on an otherwise-correct model. That is the FD reference being under-resolved, not the kernel. Every existing ODE parity fixture sets `ode_reltol = 1e-10` / `ode_abstol = 1e-12`; do the same or you will chase a ghost.

## 5. #1060 is unrelated and should not block this merge

While checking the `docs/estimation/foce.qmd` aside from my review I found a genuine pre-existing bug and filed it as #1060: on an **ODE** model, a subject with **time-varying covariates** and **IIV on an estimated lagtime** is served as analytic but its `∂f/∂η` is wrong by ~300× and `∂²f/∂η²` by ~23×.

**It does not touch this PR:**

- It reproduces with **no `init(...)` at all** (that is case C in the issue) — the trigger is TV-cov × ODE × η-on-lagtime.
- It reproduces on `origin/main` semantics; nothing in this diff is in the path. This PR changes a parse-time check set only.
- The three fixtures in §4 above — including the `LAGTIME`-seeded one — all pass, because they are plain-bolus subjects with no time-varying covariates.

`init(...)` only *amplifies* the error once you are already in the broken cell (588× vs 23×), which is how I stumbled into it from here. That is a coincidence of where I was looking, not a dependency.

Merge this on its own merits. #1060 wants its own fix and its own parity fixture.

## Filed from this review

| | |
|---|---|
| #1057 | docs: carve-out not propagated to `individual-parameters.qmd` (+ `ode-models.qmd:338`) |
| #1058 | `E_DOSE_ATTR_DOUBLE_USE` parse-time message over-reports for `[odes]` |
| #1059 | no `Dual2`-vs-FD parity test for a dose-attribute-seeded init |
| #1060 | **unrelated** — ODE + TV-cov + IIV-on-lagtime wrong analytic η-gradient |

#1057 and #1058 are this PR's own drift and are cheap to fold in here if you'd rather not carry them separately.

PR #1056, opened 2026-08-24
feat(parser): [scaling] intermediates, continuation lines, and 2-arg min/max [closes #1030]

### LINKED ISSUE #1030: feat(dsl): [scaling] intermediates, continuation lines, and 2-arg min/max

## Summary

Three individually-defensible restrictions collide and force model files that are effectively unmaintainable. Fixing any one of them relieves most of the pressure; fixing intermediates in `[scaling]` relieves nearly all of it.

- [ ] Named intermediates in `[scaling]`
- [ ] Continuation lines in `[individual_parameters]`
- [ ] Two-argument `min()` / `max()`

## What it produces today

This is the readout for a standard bounded-endpoint MBMA — a weighted logit of an Emax time course with the published `[0.01, 0.99]` clamp. One line, ~200 characters, with the same sub-expression written **four times** because there is nowhere to name it:

```
[scaling]
  y[CMT=1] = sqrt(NARM) * log((if (EMAX * clock / (clock + T50) > 0.01) EMAX * clock / (clock + T50) else 0.01) / (1 - (if (EMAX * clock / (clock + T50) > 0.01) EMAX * clock / (clock + T50) else 0.01)))
```

The published Mlxtran for the identical model is three readable lines:

```
ACR20    = Emax*(t/(t+T50))
ACR20sat = min(max(ACR20,0.01),0.99)
pred     = logit(ACR20sat)*sqrt(Narm)
```

No reviewer can check the first version, and no author will revisit it.

## The three restrictions

| Restriction | Error today | Wanted |
|---|---|---|
| No named intermediates in `[scaling]` | ``[scaling]: unknown key `ACR20` `` | local assignments, as `[individual_parameters]` already allows |
| No continuation lines in `[individual_parameters]` | ``Expected an assignment, an `if` block, or `d/dt(...)`, got Plus`` | a leading `+` (or `\`) continues the previous expression |
| No two-argument `min` / `max` | `Missing closing parenthesis for function max` | `min(a, b)` / `max(a, b)` |

The continuation-line one bites on any additive-on-logit covariate model, where the natural layout is one covariate per line:

```
LEMAX = LEMAX0
        + log(OR_ABATA) * ABATA
        + log(OR_ADALI) * ADALI
        + log(OR_CANAK) * CANAK
        + ETA_EMAX
        + KAPPA_EMAX / sqrt(NARM)
```

which must currently be collapsed onto one line.

## Note on the `min`/`max` diagnostic

`max(ACR20, 0.01)` reports **`Missing closing parenthesis for function max`**, which sends the reader looking for a bracket bug that does not exist. Even if two-argument forms stay unsupported, the message should say the arity is wrong. `if (…) … else …` does work and is the current substitute, at the cost of duplicating the guarded expression.

## Suggested priority

Intermediates in `[scaling]`. That is where readout logic naturally lives, it makes the clamp expressible once, and it subsumes most of the need for the other two.

## Related

- #811 — a compartment-free `$PRED` structural model would give readout logic a proper home
- #1028 — `[scaling]` undefined-identifier guard; intermediates would need to participate in the same "defined" set

## Environment

```
```

roninsightrx:
## Why

Closes #1030 (part of the MBMA umbrella #1032).

Three individually-defensible restrictions collided and produced model files that
are effectively unreviewable. The standard bounded-endpoint MBMA readout — a
weighted logit of an Emax time course with the published `[0.01, 0.99]` clamp —
came out as **one ~200-character line with the same sub-expression written four
times**, because `[scaling]` had nowhere to name it:

```ferx
[scaling]
  y[CMT=1] = sqrt(NARM) * log((if (EMAX * clock / (clock + T50) > 0.01) EMAX * clock / (clock + T50) else 0.01) / (1 - (if (EMAX * clock / (clock + T50) > 0.01) EMAX * clock / (clock + T50) else 0.01)))
```

The published Mlxtran for the identical model is three readable lines. It now
reads that way in ferx too:

```ferx
[scaling]
  ACR20    = EMAX * TIME / (TIME + T50)
  ACR20SAT = min(max(ACR20, 0.01), 0.99)
  y        = log(ACR20SAT / (1 - ACR20SAT)) * sqrt(NARM)
```

## What changed

**1. Named intermediates in `[scaling]`.** Any line whose key is not
`obs_scale` / `y` declares a local binding the entries below it can reference,
following the same define-above-use-below rule `[individual_parameters]` already
uses.

The implementation choice that matters: intermediates are **inlined into the
entry's AST**, immediately after that entry parses and before anything else looks
at it. There is no new `Expression` variant and no new evaluation stage, so every
downstream consumer sees exactly the expression the user would otherwise have
written by hand — bytecode compilation, the `Dual2` sensitivity programs, the
#1028 required-data-column scan, the #993/#1004 dose-attribute double-use
rejection, the `T`-alias fold, the #486 direct-θ/η desugaring, and the #650
readout slot allocator all keep working with no changes to their logic. The three
pre-scans that run *before* `parse_scaling_block` inline through the same shared
helper, so they can't drift on which lines are entries and which are bindings.
The cost of inlining is that an intermediate used twice is evaluated twice, which
is precisely what hand-expansion costs today.

Rejected, each with its own message: forward and self references (which makes
cycles unrepresentable rather than a recursion guard to get right), a name that is
already a theta / eta / individual parameter / compartment (a reference would
resolve to that, never to the binding), duplicates, reserved names (`TIME`, `T`,
`MIXNUM`), and `[CMT=N]` subscripts.

One rejection is worth calling out separately: **an intermediate no entry reads is
an error.** Before this PR every key other than `obs_scale` / `y` was rejected
outright, and that is what caught a misspelt `obs_scal = V`. Now that any key is
legal syntax, that typo would silently disable scaling instead — so the same
protection is re-established from the other side, with the error naming both
readings.

**2. Operator continuation lines** in `[individual_parameters]`, `[odes]`,
`[scaling]`, `[derived]`, and `[initial_conditions]`. Start the continued line with
an operator, or end the previous one with it:

```ferx
LEMAX = LEMAX0
        + log(OR_ABATA) * ABATA
        + log(OR_ADALI) * ADALI
        + ETA_EMAX
```

Both spellings are unambiguous because no statement in these blocks can *begin*
with a binary operator, so this strictly widens the accepted grammar rather than
reinterpreting anything that parsed before. Applied at block extraction, so the
token-stream consumers (`[individual_parameters]`, `[odes]`) and the line-oriented
one (`[scaling]`) cannot disagree about what a line is.

**3. Two-argument `min(a, b)` / `max(a, b)`,** available everywhere the DSL parses
expressions. They desugar to `Expression::Conditional` — the `if (a >= b) a else b`
form they replace — which is already evaluated, bytecode-compiled,
`Dual2`-differentiated and index-resolved throughout the pipeline. So they get all
of that for free and cannot drift from the hand-written equivalent.

Two diagnostics come with it. `max(a, b)` used to report **`Missing closing
parenthesis for function max`**, sending the reader after a bracket bug that does
not exist; arity errors now say what is actually wrong. And a one-argument
`max(x)`, previously the silent identity, is now rejected.

## Alternatives considered

- **A two-argument AST node for `min`/`max`** instead of desugaring. Rejected:
  ~30 match sites (eval, bytecode, differentiation, index resolution, the
  `Dual2` walk) would each need a new arm, and a missed one is a silently wrong
  gradient. The desugaring has one source of truth.
- **Textual substitution of intermediates** before parsing. Rejected: error
  messages would quote the expanded text, and identifier-boundary handling is a
  hazard with no upside over AST substitution.
- **A separate evaluation stage for intermediates** (compile them to their own
  program, evaluate, feed into the entry). Rejected for this phase: it would need
  the sensitivity chain, the covariate scan, and the dose-attribute scan each
  taught about a new stage, for a saving (re-evaluating a shared sub-expression)
  that nobody gets today either.
- **Trailing `\` for continuation.** The issue offered it as an alternative;
  leading/trailing operators need no new lexical token and match the layout the
  issue itself shows.

## Cross-repo dependency

No `pub` API change in practice: `build_y_output_fn` gained a parameter, but it is
reached only from within this crate (`sim::adaptive_control`), which is updated
here.

## Breaking changes

- [ ] Public Rust API
- [ ] None

<details>

Two narrow behaviours change, both of which only affect model files that were
already doing something inert or wrong:

1. **A one-argument `min(x)` / `max(x)` is now a parse error.** It used to be the
   silent identity (the parser has no known-function list, so an unrecognised
   one-argument call returns its argument unchanged). Any model relying on that
   was not clamping anything.

2. **A `[scaling]` key that is neither `obs_scale`/`y` nor read by any entry is
   rejected as an unused intermediate,** with a different message than the old
   `unknown key`. Same set of files rejected, different wording — except that such
   a key *is* now accepted when an entry reads it, which is the feature.

</details>

## Numerical validation

- [x] Not applicable in the NONMEM-anchor sense — this is DSL surface only and
  introduces no new numerical path.

The equivalence is pinned directly instead, which is the stronger statement here:
`test_scaling_intermediates_match_hand_expanded_readout` compiles the
intermediate-based readout and the hand-expanded one-liner it replaces and asserts
the two output functions agree **bit-for-bit** (`to_bits()`) at four points
spanning both sides of the clamp. Since the hand-expanded form is the one already
anchored (Form C's NONMEM anchor in `docs/model-file/scaling.qmd`), the named form
inherits that anchor exactly.

`test_scaling_intermediate_keeps_the_analytic_readout_path` pins the other half:
a readout that was analytic when written inline must not drop to
finite-difference sensitivities once it is named. It was verified non-vacuous by
temporarily disabling the pre-scan inlining, which makes it fail.

## Performance

- [x] No significant impact expected — parse-time only. An intermediate used
  twice is evaluated twice at runtime, identical to hand-expansion.

## Tests

- [x] Unit tests added in the same module (`cargo test --lib` passes: 3190 passed, 0 failed)
- [x] Regression tests that fail without the change

New Tier-1 tests:

| Test | Pins |
|---|---|
| `test_scaling_intermediates_match_hand_expanded_readout` | bit-identity with the hand-expanded readout |
| `test_scaling_intermediates_two_sided_clamp` | `min`/`max` compose into the `[0.01, 0.99]` clamp, biting on both sides |
| `test_scaling_intermediate_covariate_is_registered` | a covariate reached only through an intermediate is a required data column (#1028) |
| `test_scaling_intermediate_in_obs_scale` | intermediates work on the Form B half too |
| `test_scaling_intermediate_keeps_the_analytic_readout_path` | #486 / #650 pre-scans see through intermediates |
| `analytical_dose_attr_read_through_scaling_intermediate_is_rejected` | #1004 is not laundered by naming the read |
| `test_scaling_intermediate_cannot_smuggle_time_into_obs_scale` | #1028's `obs_scale`-is-subject-static rejection still fires |
| `test_scaling_intermediate_{forward,self}_reference_rejected`, `_duplicate_`, `_shadowing_`, `_reserved_name_` | the rejection surface |
| `test_parse_scaling_unknown_key_errors` (updated), `..._with_cmt_subscript_errors` | the typo guard survives the syntax widening |
| `test_join_continuation_lines` | the joiner, including the leading-operator-on-first-line no-op |
| `test_individual_parameters_operator_continuation_lines` | the issue's covariate model, evaluated |
| `test_scaling_continuation_lines` | continuation on the line-oriented `[scaling]` path |
| `test_initial_conditions_continuation_lines` | continuation on the `[initial_conditions]` path, pinned on the init *amount* (a silently-dropped continuation would still parse) |
| `test_min_max_two_args_in_individual_parameters` | two-arg `min`/`max` evaluate, and the clamp bites |
| `test_min_max_arity_diagnostics` | both new arity messages |

- [x] Inline below (no standalone `examples/` file: the MBMA model has no committed
  dataset to run it against, so a file there would not be exercised)

<details>
<summary>Example</summary>

```ferx
[scaling]
  ACR20    = EMAX * TIME / (TIME + T50)
  ACR20SAT = min(max(ACR20, 0.01), 0.99)
  y        = log(ACR20SAT / (1 - ACR20SAT)) * sqrt(NARM)
```

```ferx
[individual_parameters]
  LEMAX = LEMAX0
          + log(OR_ABATA) * ABATA
          + log(OR_ADALI) * ADALI
          + ETA_EMAX
          + KAPPA_EMAX / sqrt(NARM)
```

</details>

## Docs

- [x] `docs/model-file/scaling.qmd` — new *Named intermediates* section (rules,
  the inlining consequence, `min`/`max`)
- [x] `docs/model-file/individual-parameters.qmd` — new *Continuation Lines*
  section, `min`/`max` in the function table
- [x] No new pages, so `docs/_quarto.yml` unchanged
- [x] `CHANGELOG.md` `[Unreleased]` → `Added`

## Checklist

- [x] `cargo clippy --no-default-features --features ci,markov` clean — no new warnings (the pre-existing `approx_constant` errors that appear under `--all-targets` are in `sens/two_cpt_explicit.rs` test data and are not linted by CI)
- [x] `cargo check --tests` clean
- [x] No `Dual2`-path formula touched — `min`/`max` reuse `Expression::Conditional`, which is already differentiated

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` or dataset touched.

## Reviewer hints

Focus on **`inline_scaling_intermediates` and its call sites**. That is the whole
correctness argument: if a scan over `[scaling]` lines does not inline, it sees the
un-expanded readout and its check silently under-fires. The sites are
`build_obs_scale_spec`, `build_y_output_fn`, `collect_readout_theta_eta_synth`
(#486), `allocate_readout_extra_slots` (#650), and `collect_indiv_param_reads` for
the #993/#1004 read scan. One further scan needs nothing: the `downstream_refs`
scan in `parse_model_file` matches raw block text, so it already sees names inside
intermediate lines.

`join_continuation_lines` is worth a second look for the claim that it only widens
the grammar: it joins when the current line starts with `+ - * / ^`, or the previous
ends with one of those plus `,` or `=`. The argument is that no statement in a
continuation block can start with a binary operator, so any such line was
previously a parse error.

Skimmable: the tests, the docs, and the `CHANGELOG` entry.

## Open questions

- The unused-intermediate rejection is an **error**, not a warning. I chose the
  error because the alternative silently disables `obs_scale` on a typo, but it is
  the one place a user could reasonably want laxer behaviour (e.g. keeping a
  commented-out readout's intermediate around). Confirmed as-is by Ron.

*(Resolved: `initial_conditions` is now in `CONTINUATION_BLOCKS` — its grammar is
`init(NAME) = <expr>`, one per line, so a line opening with a binary operator is
unreachable there too.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,6 +228,24 @@ section of the SDLC for the versioning policy).
   stopped theta from fixing.
 
 ### Added
+- **`[scaling]` accepts named intermediates, expressions can be split across lines, and `min`/`max`
+  take two arguments (#1030).** Three restrictions that individually looked defensible combined to
+  make a standard bounded-endpoint readout unmaintainable: a model-based meta-analysis logit-Emax
+  readout with the published `[0.01, 0.99]` clamp came out as one ~200-character line with the same
+  sub-expression written four times, because there was nowhere to name it. Now: (1) any `[scaling]`
+  line whose key is not `obs_scale` / `y` declares a **named intermediate** usable by the entries
+  below it, following the same define-above-use-below rule as `[individual_parameters]` —
+  intermediates are inlined, so covariates reached only through one are still required data columns
+  and the dose-attribute double-use rejection still sees them; (2) a long expression may be
+  **continued across lines** by starting the continued line with an operator or ending the previous
+  one with it, in `[individual_parameters]`, `[odes]`, `[scaling]`, and `[derived]`; (3) **`min(a, b)`
+  and `max(a, b)`** are available everywhere the DSL parses expressions, desugaring to the inline
+  conditional so they differentiate and compile exactly like the hand-written `if (a >= b) a else b`.
+  The three-line readout now reads the way the published Mlxtran does. Two smaller diagnostics come
+  with it: `max(a, b)` used to report `Missing closing parenthesis for function max`, which sent the
+  reader after a bracket bug that did not exist — arity errors now say so — and a `[scaling]` key
+  that is neither `obs_scale`/`y` nor read by any entry is rejected, so a misspelt `obs_scal = V`
+  still fails loudly instead of silently disabling scaling.
 - **SAEM now warns when an estimated theta carries no ETA at all.** A fixed-effect-only theta is not
   mu-referenced, so it never gets the γ-damped closed-form `log θ += γ·mean(η)` update and is moved
   only by the η-frozen numerical M-step — which re-maximises against a *single* MCMC η draw with no
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,7 +238,8 @@ section of the SDLC for the versioning policy).
   intermediates are inlined, so covariates reached only through one are still required data columns
   and the dose-attribute double-use rejection still sees them; (2) a long expression may be
   **continued across lines** by starting the continued line with an operator or ending the previous
-  one with it, in `[individual_parameters]`, `[odes]`, `[scaling]`, and `[derived]`; (3) **`min(a, b)`
+  one with it, in `[individual_parameters]`, `[odes]`, `[scaling]`, `[derived]`, and
+  `[initial_conditions]`; (3) **`min(a, b)`
   and `max(a, b)`** are available everywhere the DSL parses expressions, desugaring to the inline
   conditional so they differentiate and compile exactly like the hand-written `if (a >= b) a else b`.
   The three-line readout now reads the way the published Mlxtran does. Two smaller diagnostics come
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,10 @@ section of the SDLC for the versioning policy).
   with no diagnostic from NONMEM (`nonmem_anchor/dose_attr_double_use_{A,B}.ctl`).
 
 ### Fixed
+- **A `[derived]` row aggregate followed by more expression is now an error instead of silently
+  dropping the rest (#1030 review).** `CMAX = max(IPRED) * 2` parsed as the aggregate and discarded
+  the `* 2`, reporting the unscaled maximum. There is no aggregate form that continues into a larger
+  expression, so it now says so rather than returning a wrong number.
 - **A sigma declared past the ones a `combined` `[error_model]` names no longer perturbs the FOCE
   score or the `s` / `rsr` standard errors (#1001 review).** `dr_diag_d_log_sigma`'s `Combined` arm
   answered every `sigma_k >= 1` with the *additive* sigma's derivative, where its `additive` and
@@ -241,7 +245,10 @@ section of the SDLC for the versioning policy).
   one with it, in `[individual_parameters]`, `[odes]`, `[scaling]`, `[derived]`, and
   `[initial_conditions]`; (3) **`min(a, b)`
   and `max(a, b)`** are available everywhere the DSL parses expressions, desugaring to the inline
-  conditional so they differentiate and compile exactly like the hand-written `if (a >= b) a else b`.
+  conditional so they differentiate and compile exactly like the hand-written `if (a >= b) a else b`
+  — each argument appears twice in the desugared tree, so clamp a named intermediate rather than a
+  long expression. In `[derived]`, where `min`/`max` also name the row aggregate, the two are told
+  apart by the second argument: a comparison is a row filter, anything else is the numeric clamp.
   The three-line readout now reads the way the published Mlxtran does. Two smaller diagnostics come
   with it: `max(a, b)` used to report `Missing closing parenthesis for function max`, which sent the
   reader after a bracket bug that did not exist — arity errors now say so — and a `[scaling]` key
--- a/docs/model-file/individual-parameters.qmd
+++ b/docs/model-file/individual-parameters.qmd
@@ -79,8 +79,9 @@ This is the natural layout for an additive-on-logit covariate model — one
 covariate per line — which previously had to be collapsed onto a single line
 ([#1030](https://github.com/FeRx-NLME/ferx-core/issues/1030)). A trailing `=`
 continues too, so `LEMAX =` on its own line followed by an indented expression is
-accepted. The same applies in `[odes]`, `[scaling]`, and `[derived]`; newlines
-*inside* `(...)` or `[...]` have always been ignored and still are.
+accepted. The same applies in `[odes]`, `[scaling]`, `[derived]`, and
+`[initial_conditions]`; newlines *inside* `(...)` or `[...]` have always been
+ignored and still are.
 
 ## Conditional Logic (`if` / `else`)
 
--- a/docs/model-file/individual-parameters.qmd
+++ b/docs/model-file/individual-parameters.qmd
@@ -45,10 +45,42 @@ name from quietly reading as zero and collapsing the formula.
 | `log()`, `ln()` | `log(TVCL)` |
 | `sqrt()` | `sqrt(WT)` |
 | `abs()` | `abs(ETA_CL)` |
+| `min(a, b)`, `max(a, b)` | `max(TVCL * WT / 70, 0.1)` |
 | Parentheses | `TVCL * (WT/70)^0.75` |
 | Comparisons (in `if` conditions) | `WT > 70`, `SEX == 1`, `AGE != 0` |
 | Logical (in `if` conditions) | `&&` (and), `\|\|` (or), `!` (not) |
 | Inline conditional | `if (SEX == 1) TVCL * 1.5 else TVCL` |
+
+`min` / `max` take **exactly two** arguments and desugar to the inline
+conditional (`max(a, b)` → `if (a >= b) a else b`), so they differentiate and
+compile identically to the hand-written form. A one-argument `max(x)` is a parse
+error rather than the silent identity it used to be
+([#1030](https://github.com/FeRx-NLME/ferx-core/issues/1030)).
+
+## Continuation Lines
+
+A long expression may be split across several physical lines. Two spellings work,
+and both are unambiguous because no statement can *begin* with a binary operator:
+put the operator at the start of the continued line, or at the end of the line
+being continued.
+
+```
+[individual_parameters]
+  LEMAX = LEMAX0
+          + log(OR_ABATA) * ABATA
+          + log(OR_ADALI) * ADALI
+          + ETA_EMAX
+
+  KEL   = CL /
+          V
+```
+
+This is the natural layout for an additive-on-logit covariate model — one
+covariate per line — which previously had to be collapsed onto a single line
+([#1030](https://github.com/FeRx-NLME/ferx-core/issues/1030)). A trailing `=`
+continues too, so `LEMAX =` on its own line followed by an indented expression is
+accepted. The same applies in `[odes]`, `[scaling]`, and `[derived]`; newlines
+*inside* `(...)` or `[...]` have always been ignored and still are.
 
 ## Conditional Logic (`if` / `else`)
 
--- a/docs/model-file/scaling.qmd
+++ b/docs/model-file/scaling.qmd
@@ -335,6 +335,62 @@ bit-for-bit-equal predictions on the analytic and ODE paths at matched
 parameters — the analytic-vs-ODE equivalence is checked by
 `analytic_form_c_matches_ode_form_c_binding_readout`.
 
+## Named intermediates (since #1030)
+
+Any line in `[scaling]` whose key is **not** `obs_scale` or `y` declares a
+**named intermediate** — a local binding the `obs_scale` / `y` entries below it
+can reference, exactly as `[individual_parameters]` already allows. This is what
+makes a bounded-endpoint readout writable once instead of once per occurrence.
+
+The standard model-based meta-analysis readout — a weighted logit of an Emax time
+course with the published `[0.01, 0.99]` clamp — reads as three lines:
+
+```ferx
+[scaling]
+  ACR20    = EMAX * TIME / (TIME + T50)
+  ACR20SAT = min(max(ACR20, 0.01), 0.99)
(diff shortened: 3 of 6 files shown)


roninsightrx:

Five findings. Ordered roughly by user impact.

---

**1. Docs claim two-arg `min`/`max` work in `[derived]`; at statement top level they don't.**
`docs/model-file/scaling.qmd:381`, `CHANGELOG.md:238`

`parse_derived_block` (`src/parser/model_parser.rs:4045`) intercepts `max|min|tmax|integral` when they are `tokens[0]` and treats the second argument as a **row-filter condition**, not a value. So `X = max(EMAX, 0.01)` in `[derived]` routes to `parse_derived_cond("0.01")` and fails — while the *same* call nested one level deep (`X = 1 * max(EMAX, 0.01)`) silently means the new numeric form. Same spelling, two meanings, position-dependent.

Fix the doc/changelog claim, and ideally make the top-level derived path fall back to the numeric form when the second argument contains no comparison (mirroring the existing `tokens_contain_comparison`).

---

**2. `min`/`max` desugaring duplicates each operand into both the condition and a branch — the stated perf characterisation is off.**
`src/parser/model_parser.rs:18509`

Every argument appears twice in the resulting AST (once in `Condition::Compare`, once in the taken branch), and the duplication is multiplicative through nesting *and* through intermediate inlining. The PR's own flagship example —

```ferx
ACR20SAT = min(max(ACR20, 0.01), 0.99)
y        = log(ACR20SAT / (1 - ACR20SAT)) * sqrt(NARM)
```

— expands `EMAX * TIME / (TIME + T50)` to **8 copies** in the bytecode and in the `Dual2` sensitivity program, not the "an intermediate used twice is evaluated twice" the PR description and CHANGELOG state. Numerically correct; a three-deep clamp is 8–16×. Worth stating accurately.

---

**3. Intermediate name-clash check misses declared covariates and the eval-time built-ins.**
`src/parser/model_parser.rs:7911`

The check covers theta / eta / individual parameter / compartment, but not `declared_covariates` (in scope at the call site) nor `TAD` / `TAFD` / `MACHEPS`. So `[scaling] WT = 70` silently rebinds every `WT` leaf in the `obs_scale` / `y` entries to the literal **and** removes `WT` from the required-data-column set — the exact opposite of the "No shadowing … the binding would be silently dead" rule the new docs state. Either reject it or document it as deliberate shadowing.

---

**4. Define-above-use-below is enforced for intermediate→intermediate but not for entries.**
`src/parser/model_parser.rs:14212` (`inline_scaling_intermediates`)

Entries are substituted against **all** intermediates regardless of source order, so

```ferx
[scaling]
  y   = FOO * 2
  FOO = 3
```

parses and works, even though forward references between intermediates are rejected and both `scaling.qmd` and the doc comment promise "define above, use below". The rule is enforced on one side only.

---

**5. Dead-binding guard is token-based while substitution is AST-based, so a binding can silently vanish.**
`src/parser/model_parser.rs:14243` (`check_scaling_intermediates_used`)

`check_scaling_intermediates_used` scans `source_identifiers`; `inline_scaling_intermediates` walks the AST. A name appearing in a position the substituter cannot reach therefore counts as "used" and the binding disappears with no error. Concretely, with a `[covariate_nn FOO]` block present, `[scaling] FOO = 2` plus `y = FOO.OUT1` resolves `FOO.OUT1` to `Expression::NnOutput` before intermediates are consulted: the guard sees `FOO` in the token stream and passes, the intermediate is never inlined and never reported. Same for `y = FOO(A1)`, which parses as an unknown `UnaryFn` (identity). Low likelihood, but it is precisely the silent-typo case the guard exists to prevent.

---

**Checked and found clean:** `visit_expr_nodes{,_mut}` and `resolve_expr_indices` both descend into `Condition::Compare`, so intermediates inside a `min`/`max` guard are substituted and index-resolved; all five inlining call sites (`build_obs_scale_spec`, `build_y_output_fn`, `collect_readout_theta_eta_synth`, `allocate_readout_extra_slots`, `collect_indiv_param_reads`) skip non-entry lines correctly and pass `&[]` where there is no `[scaling]` scope (`compile_observe`, `[adaptive_dosing] observe`); `parse_scalar_expression` rejects trailing tokens, so a continuation join that swallows a following statement fails loudly rather than silently dropping it; and `join_continuation_lines` ordering relative to `join_bracketed_lines` and the `ode_template` block merge is safe.

roninsightrx:
All five findings fixed in `39268f11`.

| # | Finding | Fix |
|---|---------|-----|
| 1 | Two-arg `min`/`max` broken at `[derived]` statement top level | A second argument with no comparison operator cannot be a row filter → parse the whole statement as an expression. Doc claim is now true. Also: trailing tokens after a row aggregate's `)` were silently dropped (`max(IPRED) * 2` reported the unscaled maximum) — now an error, with a `Fixed` CHANGELOG entry. |
| 2 | Stated perf characterisation off | Corrected in `scaling.qmd` and `CHANGELOG.md`: each `min`/`max` operand appears twice, multiplying through nesting and inlining — a two-sided clamp read from two places expands its value eight times. |
| 3 | Name-clash check missed declared covariates + `TAD`/`TAFD`/`MACHEPS` | Both added. The covariate case is called out in the docs as the mirror image: there the *binding* would win and drop the data column from the required-column set. |
| 4 | Define-above-use-below enforced on one side only | Entries now obey it too — an `obs_scale`/`y` line may only read an intermediate declared above it. |
| 5 | Token-based guard vs AST-based substitution | `source_identifiers` now skips an identifier followed by `(` or `.`, so a name the substituter cannot reach is reported unused instead of vanishing. |

Seven Tier-1 regression tests added, each failing without its fix:

- `parse_derived_two_arg_max_at_statement_top_level` (top level and nested agree, and the clamp bites)
- `parse_derived_aggregate_filter_still_parses` (the comparison second argument is untouched)
- `parse_derived_aggregate_rejects_trailing_tokens`
- `test_scaling_intermediate_eval_builtin_name_rejected`
- `test_scaling_intermediate_covariate_shadowing_rejected`
- `test_scaling_entry_forward_reference_rejected`
- `test_scaling_intermediate_in_unreachable_position_is_reported`

`cargo test --lib`: 3198 passed, 0 failed. `cargo fmt --check`, `cargo check --tests` clean; no new clippy warnings.

PR #1061, opened 2026-08-24
fix(parser): reject TIME in init() and publish the init scope rule [closes #994]

### LINKED ISSUE #994: `init()` accepts `TIME` (and `time`, not `Time`) -- always zero, and contradicting its own diagnostic
`init()` accepts `TIME`. Its own error message says it should not, and `TIME` at init time is always exactly zero -- so the expression is accepted, contributes nothing, and says nothing.

Found while a consumer (`ferxtranslate`) was reverse-engineering the init scope rule by probing the engine one name at a time, because the error message is the only place the rule is written down.

## The contradiction

The diagnostic:

```
error[E_PARSE]: [odes] init(central): references undefined name(s): TAD. An init
expression may only reference declared states (0 at init time), individual
parameters, or the MACHEPS constant (defined: BL, CL, V, central).
```

Measured on a 1-cpt ODE model, varying **only** the init RHS (`BL` an individual parameter, `central` the declared state):

| init RHS | result |
|---|---|
| `MACHEPS` | ok |
| `TIME` | **ok -- accepted, and absent from the message** |
| `time` | **ok** |
| `Time` | E_PARSE |
| `central` | ok |
| `BL` | ok |
| `3.0` | ok |
| `T` | E_PARSE |
| `TAFD` | E_PARSE |
| `TAD` | E_PARSE |

Reproduced independently on two builds (installed 0.3.0 and a `target/debug/ferx` from a June-complete tree) and two different models.

## It is silent, not merely undocumented

`TIME` at init evaluates to exactly zero, always:

```
init(central) = TIME + 50   ->  PRED [0.904837, 0.818731, 0.67032, 0.44933, 0.090723]
init(central) = 0 + 50      ->  PRED [0.904837, 0.818731, 0.67032, 0.44933, 0.090723]
                                max |diff| = 0.0

init(central) = TIME        ->  PRED all exactly 0.0
```

So there is no reading under which `TIME` is legitimately in scope: model time at init is zero by definition, and the expression can only ever mislead the person who wrote it. A model that says `init(central) = TIME * SOMETHING` initialises to zero, validates clean, and fits.

## Why it happens: representation, not scope

The undefined-name check exempts exactly one name by hand (`model_parser.rs`, the `init` branch):

```rust
undef.retain(|n| !n.eq_ignore_ascii_case("MACHEPS"));
```

`TIME` is never in `undef` to begin with. A bare `TIME` parses to `Expression::Time` (`model_parser.rs:3215`, "A bare `TIME` parses to `Expression::Time`") -- a dedicated AST node, not a variable -- so `collect_undefined_vars` cannot see it. `TAFD` and `TAD` go through the variable path (`ctx.tafd` / `ctx.tad`), are absent from the init context, and are correctly flagged.

All three sit in the same list:

```rust
const RESERVED_ODE_NAMES: &[&str] = &["TIME", "T", "TAFD", "TAD", "MACHEPS"];
```

`init()` accepts one of them and rejects two, purely on how each is represented in the AST. That is not a decision anybody made.

**The casing is the tell.** `MACHEPS` is deliberately case-insensitive, and says so in a comment. `TIME` accepts exactly `TIME` and `time` and rejects `Time` -- the signature of a token match, not of a scope rule. A name that was meant to be in scope would not behave that way.

## Both init surfaces are affected

Not confined to `[odes] init()`. The analytical path takes it too:

```
[structural_model]
  pk one_cpt_iv(cl=CL, v=V)

[initial_conditions]
  init(central) = TIME + 50     ->  ok, no errors
```

## Why this matters to a consumer

The rule is **enforced** in ferx-core and **stated** nowhere authoritative, so every consumer mirrors it by guesswork and the mirror drifts silently when the engine changes. Two consumers reading the same engine get different answers depending on whether they believed the diagnostic or tested it.

`ferxtranslate` hit the sharp edge of this for real. Its scope guard exists precisely so it never emits something the engine will reject; it mirrored the wider `[odes]` reserved-name list, so a NONMEM `A_0(n) = TAD` passed the guard, emitted `init(CENT) = TAD`, and the engine refused the file. Loud, caught, fixed.

Had the source said `A_0(n) = TIME`, the same guard would have passed, the engine would have **accepted** it, and the compartment would have silently initialised to zero with a green test suite. The loud case was the lucky one.

## Which is it, and the ask

Two possibilities, and the evidence points hard at the first:

- **(a) `TIME` should be rejected** -- the message is right, the parser leaks. Supported by: `TIME` is always zero at init, so it carries no value; the casing asymmetry shows it was never scoped deliberately; `TAFD`/`TAD`, its siblings in the same reserved list, are rejected.
- **(b) `TIME` is legitimately in scope** -- then the message needs it added, and `TAFD`/`TAD`/`Time` need explaining, and someone needs to say what a value that is always zero is for.

**Minimum fix:** make the parser and the message agree, whichever way you decide.

**Durable fix, and the part worth more than this bug:** expose the accepted-name set so consumers do not hardcode a copy of it. Today the only way to learn the rule is to probe the engine name by name, which is how this was found. `ferxtranslate`'s own contributor guide says it must never generate syntax the parser rejects; the corollary is that it should not have to guess what the parser accepts.

## Related

- #993 -- a bare `F` in `[individual_parameters]` silently applied as bioavailability. Same family: engine behaviour knowable only by experiment. Different surface and different fix, hence a separate issue.

Happy to add reviewers -- assigning myself for now.

roninsightrx:
## Why

Issue #994. `init()` accepted `TIME` on both init surfaces while its own diagnostic said it should not, and `TIME` at an initial condition is always exactly zero — so the expression was accepted, contributed nothing, and said nothing:

```
init(central) = TIME + 50   ->  PRED [0.904837, 0.818731, 0.67032, 0.44933, 0.090723]
init(central) = 0 + 50      ->  identical, max |diff| = 0.0
init(central) = TIME        ->  PRED all exactly 0.0
```

The split was **representational, not a scope decision**. A bare `TIME` parses to `Expression::Time`, a dedicated AST node, so `collect_undefined_vars` — which walks `Expression::Variable` — could never see it. `TAFD` / `TAD`, its siblings in the same `RESERVED_ODE_NAMES` list, go through the variable path and were correctly flagged. The casing was the tell: `TIME` accepted exactly `TIME` and `time` and rejected `Time` — a token match, not a scope rule.

This bit a consumer for real: `ferxtranslate` mirrors the scope rule by guesswork because the diagnostic is the only place it is written down. A NONMEM `A_0(n) = TAD` was caught loudly; `A_0(n) = TIME` would have been accepted and silently zeroed the compartment with a green test suite.

## What changed

Option **(a)** from the issue for `TIME` — the message was right, the parser leaked — and option **(b)** for `MIXNUM`, per @TeunP's follow-up comment.

1. **`TIME` rejected on both init surfaces.** `expr_references_time_node()` (a node-level check, deliberately *not* the `T`/`t`-alias-folding `expr_references_time_builtin` — the alias fold never runs on an init RHS) guards `[odes] init(...)` and `[initial_conditions] init(...)`. The error points at `d/dt(...)` (ODE) / `[scaling]` Form C (analytical), where `TIME` resolves per event.

2. **`MIXNUM` kept and documented.** The enum has exactly two payload-free variants, `Time` and `MixNum`, and both leaked for the same structural reason — but `MIXNUM` is *not* inert: it resolves through `current_mixture_class()` to the subject's class, so a class-switched baseline does real work (verified in the issue comment: `init(central) = MIXNUM * 100` is bit-identical to the class-2 control, posterior 1.0). A blanket "reject every built-in the walker cannot see" would have deleted a working feature. It is now named in the diagnostic, which listed neither built-in.

3. **The scope rule is published** as `INIT_SCOPE_BUILTINS` / `INIT_REJECTED_BUILTINS`, re-exported from the crate root next to `known_block_names()` — the same arrangement #1040 set up for block names, so a generator sources the rule from the same binary it will parse with. The `[odes] init(...)` diagnostic is rendered from those lists, so message and guard cannot drift apart again.

New message:

```
[odes] init(response): references the `TIME` built-in, but an initial condition is
evaluated at the time origin — `TIME` there is always exactly 0, so the expression
can only ever read as its t = 0 value. Write that value directly, or move the time
dependence into the `d/dt(...)` RHS, where `TIME` resolves to the integrator's
current time. See issue #994.
```

and the undefined-name message now ends `... or the MACHEPS / MIXNUM built-ins (defined: ...). The time built-ins (TIME, T, TAFD, TAD) are out of scope — an initial condition is evaluated at the time origin.`

## Alternatives considered

- **Make `TIME` legitimately in scope (option b).** Rejected: it carries no information at t = 0, so the only models it can produce are ones that mislead their author.
- **Reject every built-in invisible to `collect_undefined_vars`.** The natural implementation, and it would have taken `MIXNUM` with it.
- **Reuse `expr_references_time_builtin` (#1028's `obs_scale` guard).** It also matches the `T`/`t` alias via `Expression::Covariate`. On an ODE init RHS `T` is already caught as an undefined name, and on the analytical surface `T` parses as a covariate leaf — reusing it there would have started rejecting a legitimately declared `T` data column. Hence the narrower node-level check.

## Cross-repo dependency

No ferx-r change is required to merge. **Follow-up, not blocking:** `ferx-r` may want to surface `INIT_SCOPE_BUILTINS` / `INIT_REJECTED_BUILTINS` the way it surfaces `known_block_names()` (#1040), which is what would let `ferxtranslate` retire its hardcoded copy of the rule. Reaching `ferxtranslate` needs the usual two bumps in order (ferx-r `src/rust/Cargo.lock`, then the downstream ferx-r pin).

## Breaking changes

- [ ] Public Rust API — *additive only* (two new `pub const`s; nothing removed or changed)
- [ ] None

<details>

A model with `TIME` (or `time`) in an `init(...)` expression now fails to parse instead of silently reading `0`. Every such model was already getting the t = 0 reading, so the fix is mechanical: substitute the value the expression had at t = 0.

```ferx
# before — accepted, TIME reads 0
[odes]
  init(central) = TIME + 50

# after
[odes]
  init(central) = 50
```

If the intent was a genuinely time-dependent quantity, it belongs in `d/dt(...)` (ODE) or in a `[scaling]` Form C readout (`y = <expr>`), where `TIME` resolves per observation.

</details>

## Numerical validation

- [x] Not applicable (parse-time rejection — no numerical path is touched; the accepted set is unchanged apart from `TIME`, whose only reachable value was `0`)

Per CLAUDE.md's "every new feature requires a comparison with NONMEM output": there is no NONMEM object here. NONMEM's `$PK A_0(n)` has no time argument at all — the analogue of this bug does not exist there — so the anchor is the issue's own measurement, that `init = TIME + 50` and `init = 0 + 50` produce bit-identical predictions (max |diff| = 0.0), which is what makes rejection lossless.

## Performance

- [x] No significant impact expected (one AST walk per `init` line, at parse time)

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 3209 passed)

Tier 1, in `src/parser/model_parser_tests.rs`:

- `test_init_time_builtin_rejected` — `TIME`, `TIME + 50`, `time * KIN`, `2 * (TIME + KIN)` all rejected, error names the built-in and cites #994.
- `test_init_time_alias_and_siblings_still_rejected` — `T` / `TAFD` / `TAD` (driven off `INIT_REJECTED_BUILTINS`) stay rejected; the new guard sits in front of the undefined-name check and must not shadow it.
- `test_init_undefined_name_message_lists_the_real_scope` — the diagnostic names every entry of both const lists, so the message cannot drift from the guard.
- `test_init_mixnum_accepted_in_mixture_model` — `init(response) = MIXNUM * 100` parses and seeds `100.0` at the class-1 default.
- `test_analytical_init_time_builtin_rejected` — the `[initial_conditions]` surface takes the same rejection.

## Example

- [x] Not applicable (a rejection, not new API surface) — the CLI text is shown under **What changed**.

## Docs

  - `docs/model-file/ode-models.qmd` — new **What an `init(...)` expression may reference** section (`#init-scope`): a table of the closed scope (states → 0, individual parameters, covariates on the analytical surface, `MACHEPS`, `MIXNUM`; time built-ins rejected), the before/after note, and a pointer to the exported const lists. This is the authoritative statement the issue asked for.
  - `docs/model-file/initial-conditions.qmd` — new Limitations bullet, linking to that section.
- [x] `CHANGELOG.md` `[Unreleased]` entry added (one under `Fixed`, one under `Added`)

## Checklist

- [x] `cargo clippy` clean (warning count unchanged vs. base)
- [x] Sensitivity path untouched — this is parse-time validation only
- [ ] `Cargo.toml` version bumped if breaking (deferred to release, per the SDLC)

## Docs & examples

- [x] No new DSL syntax, `[fit_options]` key, example `.ferx`, or data file → no ferx-site / ferx-book change needed
- [x] No example execution step needed (no example uses `TIME` in an `init`; checked across `examples/`, `tests/`, `docs/`)

## Reviewer hints

- The substantive decision is **`TIME` out, `MIXNUM` in** — worth agreeing with before the mechanics. They leak identically; they behave oppositely.
- `expr_references_time_node` vs. the existing `expr_references_time_builtin` is deliberate (see Alternatives) — that's the subtle bit.
- The rest is message wiring and docs and can be skimmed.

## Open questions

- Should `MIXNUM` be listed on the `[initial_conditions]` (analytical) surface docs too? It evaluates through the same thread-local there, but a class-switched analytical baseline has no test in-tree; I documented the scope on the shared page rather than claiming per-surface behaviour that isn't pinned.
- #995 catalogues the case-rule inconsistency across built-ins (`MACHEPS` case-insensitive, `TIME` exactly two spellings, `MIXNUM` case-insensitive, dose attributes lowercased). This PR does not change any casing rule — `Time` stays rejected — so that stays #995's call.


roninsightrx:
## Review

Self-review of this PR before it goes to a human reviewer. The `TIME` rejection itself holds up on both surfaces; one finding is blocking, because it is the same class of error the issue is about — a scope rule stated more confidently than the engine implements it.

---

### 🔴 Blocking — `INIT_SCOPE_BUILTINS` / `INIT_REJECTED_BUILTINS` describe the `[odes]` surface, but claim both

`src/parser/model_parser.rs:14022` and `:14033`

The two consts are documented as the init scope rule, with `INIT_REJECTED_BUILTINS` saying explicitly *"on either the `[odes] init(...)` or the `[initial_conditions] init(...)` surface"*. That is true of `TIME` (both surfaces now reject it) and of `MIXNUM` (an `Expression::MixNum` node, so it resolves through `current_mixture_class()` on both). It is **not** true of the other three names, and the divergence is a deliberate pre-existing rule, not an accident:

`src/api/validation.rs:132`

```rust
// `TAFD` / `TAD` / `MACHEPS` are solver-injected built-ins *inside `[odes]`* and
// ordinary covariates everywhere else — including `[scaling]`, which is
// deliberate (`docs/model-file/scaling.qmd`) so a dataset that really carries a
// `TAD` column can use it.
```

So on the analytical `[initial_conditions]` surface, which parses with `ParseCtx::new` (unknown identifier → `Expression::Covariate`):

- `init(central) = TAD * 10` **parses today** — I checked on this branch; it registers `TAD` as a required data column, exactly as `[scaling]` does. `INIT_REJECTED_BUILTINS` says it is rejected there.
- `init(central) = MACHEPS` is **not** machine epsilon there. `MapEnv::resolve` (`model_parser.rs:14738`) special-cases `MACHEPS` only under `Expression::Variable`; on this surface it is a `Covariate`, so it becomes a required data column and fails with `E_MISSING_COVARIATE`. `INIT_SCOPE_BUILTINS` says it is in scope.

A consumer that mirrors these lists — which is the entire point of exporting them — would over-reject `TAD` and under-reject `MACHEPS` on analytical models. The exported rule has to be at least as accurate as the diagnostic it replaces.

Same error in the docs (`docs/model-file/ode-models.qmd`, the new `#init-scope` section): the table is introduced with *"the scope is closed, and the same on both init surfaces"*, which is what I need to withdraw.

The rule as actually implemented:

| | `[odes] init(...)` | `[initial_conditions] init(...)` |
|---|---|---|
| declared states | yes, bound to `0` | n/a |
| individual parameters | yes | yes (plus thetas / etas) |
| covariates | no — go through `[individual_parameters]` | yes, by name |
| `MACHEPS` | yes, any casing | no — ordinary covariate |
| `MIXNUM` | yes, any casing | yes, any casing |
| `TIME` / `time` | **rejected** (this PR) | **rejected** (this PR) |
| `T`, `TAFD`, `TAD` | rejected (undefined-name) | ordinary covariates |
| `KAPPA_*` | n/a | rejected (#107) |

Two ways out, and I'd take the first:

1. Scope the names to what they describe — `ODE_INIT_SCOPE_BUILTINS` / `ODE_INIT_REJECTED_BUILTINS` — and add the analytical surface's rule as its own statement rather than folding it into one list. Costs nothing; the consts are new in this PR, so there is no compatibility question.
2. Keep the names and rewrite both doc comments per-surface. Cheaper diff, but a flat list named for both surfaces will be read as applying to both, which is how this went wrong the first time.

### 🟡 Test asserts too little to catch the regression it guards

`src/parser/model_parser_tests.rs` — `test_init_time_alias_and_siblings_still_rejected`

The test exists to prove the new `TIME` node guard, which now sits *in front of* the undefined-name check, does not shadow it. But it only asserts `err.contains("init(response)")`, which both errors satisfy. If the `TIME` guard ever widened to match `T`/`TAFD`/`TAD`, the test would still pass while silently reporting the wrong diagnostic. Assert on `"undefined name(s)"` so the test pins *which* check fired.

### 🟢 Guard ordering reports `TIME` before undefined names

`src/parser/model_parser.rs:9754`

`init(response) = TIME + BASE` reports only the `TIME` problem; the undefined `BASE` surfaces on the next parse. Both messages are actionable and the fix order is obvious, so I'd leave it — noting it so a reviewer doesn't have to rediscover it.

### 🟢 Casing asymmetry survives, deliberately

`Time` is still rejected while `TIME`/`time` are, now, rejected for a different reason — so the three spellings finally agree on the *outcome*, by accident of the guard. The underlying inconsistency (`MACHEPS` and `MIXNUM` case-insensitive, `TIME` exactly two spellings) is #995's call and is untouched here.

---

Everything else — the node-level guard vs. `expr_references_time_builtin`, the `MIXNUM` keep, the migration note, the changelog split — I'd stand behind as written. Blocking item is the exported rule's accuracy; happy to push the rename + per-surface docs before review.

roninsightrx:
All four review items addressed in `6bdce61b`, on top of the `main` merge that landed here (#1046 / #1030) — full `cargo test --lib` green afterwards (3241 passed), clippy warning count unchanged vs. base.

### 🔴 Exported rule scoped to the surface it describes

Renamed to `ODE_INIT_SCOPE_BUILTINS` / `ODE_INIT_REJECTED_BUILTINS`, with the divergence stated in both doc comments, in the docs (the `#init-scope` table is now two columns, one per surface), and in the `[initial_conditions]` limitation bullet — which previously claimed `T`/`TAFD`/`TAD` were rejected there and now says the opposite, correctly.

The claim is now **checked, not asserted**. New Tier-1 test `test_analytical_init_clock_names_are_ordinary_covariates` parses `init(central) = {T,TAFD,TAD,MACHEPS} * 10` on an analytical model and asserts each lands in `referenced_covariates` — i.e. that they really are data columns on that surface. It passes, confirming the finding.

### 🔴/🟢 One diagnostic for every clock, every spelling

Went further than the review asked, since the fix made it nearly free. The node guard and the undefined-name split now converge on a single message, so all six spellings (`TIME`, `time`, `Time`, `T`, `TAFD`, `TAD`) get the explanation instead of splitting three ways on representation. The casing tell the issue identified is gone from this surface — `Time` no longer reports as a plain undefined name.

### 🟢 Both problems from one parse

```
$ ferx t2.ferx   # init(response) = Time * TAD + BASE
Error: [odes] init(response): references the time built-in(s): TAD, Time. An initial
condition is evaluated at the time origin, so a clock there is always exactly 0 and the
expression can only ever read as its t = 0 value. Write that value directly, or move the
time dependence into the `d/dt(...)` RHS, where `TIME` resolves to the integrator's
current time. It also references undefined name(s): BASE. See issue #994.
```

Two clocks and an undefined name, one parse. Pinned by `test_init_reports_clock_and_undefined_name_together`.

### 🟡 Test assertions pin which diagnostic fired

`test_init_time_alias_and_siblings_still_rejected` → `test_init_time_clocks_rejected_in_every_casing`: iterates all four clocks × three casings and asserts the clock diagnostic specifically, so a widened guard reporting the wrong message fails instead of passing. Same tightening on the `TIME` test.

### Changelog

Second `Fixed` entry for the unified diagnostics; the `Added` entry now carries the const rename and states the per-surface difference rather than implying one shared rule.

Ready for review.

PR #1062, opened 2026-08-24
feat(parser): sample-size-weighted IOV `kappa K ~ γ² weight = N` (BTAV) [closes #1031]

### LINKED ISSUE #1031: feat(dsl): sample-size-weighted IOV -- kappa ... weight = N (BTAV)

## Summary

Between-treatment-arm variability (BTAV) — the arm-level random effect in every published longitudinal MBMA — is always distributed

$$\eta^{(BTAV)}_{ik} \sim N\!\left(0, \frac{\gamma^2}{N_{ik}}\right)$$

i.e. it scales with the number of subjects in that arm. FeRx can express it, but only by hand, inside a structural expression:

```
LEMAX = LEMAX0 + log(OR_ABATA) * ABATA + ETA_EMAX + KAPPA_EMAX / sqrt(NARM)
```

## Why that placement is unfortunate

It works — I verified it recovers γ = 2.036 against the published 2.005 on the canakinumab dataset — but it puts a **variance-structure decision in the middle of a structural equation**, where:

- `/ NARM` instead of `/ sqrt(NARM)` is an easy slip that produces a plausible wrong answer rather than an error;
- `ferx_model_inspect()` reports `KAPPA_EMAX` as an ordinary IOV term, with no indication it is weighted;
- the reported estimate is a variance on an unweighted scale, so comparing it against a published γ needs a `sqrt()` and a mental note about arm size.

## Proposal

Declare it where the rest of the variance structure is declared:

```
kappa KAPPA_EMAX ~ 2.0 (sd) weight = NARM
```

meaning `kappa_ik ~ N(0, omega_iov / N_ik)`. The engine applies the scaling; the structural expression goes back to reading `... + ETA_EMAX + KAPPA_EMAX`.

Benefits beyond tidiness: `ferx_model_inspect()` and the fit printout can report both the raw γ and the effective SD at a typical arm size (γ/√N ≈ 0.14 at N = 200 in the published fit), which is the number readers actually need — a raw γ of 2.0 on a logit scale looks alarming until it is divided.

## Relationship to #1029

This is the same idea as `weight =` on the error model, applied to a random effect rather than the residual. If both land they should share syntax and validation, and probably an implementation.

## Environment

```
Verified against the published canakinumab RA analysis (gamma = 2.0045).
```

roninsightrx:
## Why

Between-treatment-arm variability (BTAV) — the arm-level random effect in every published longitudinal MBMA — is always distributed

$$\kappa^{(BTAV)}_{ik} \sim N\!\left(0, \frac{\gamma^2}{N_{ik}}\right)$$

i.e. it scales with the number of subjects in that arm: a 400-subject arm's mean wanders a quarter as far as a 25-subject arm's. ferx could express it only by hand, inside a structural expression:

```
LEMAX = LEMAX0 + log(OR_ABATA) * ABATA + ETA_EMAX + KAPPA_EMAX / sqrt(NARM)
```

That works — it recovers γ = 2.036 against the published 2.005 on the canakinumab dataset (#1031) — but it puts a **variance-structure decision in the middle of a structural equation**, where `/ NARM` instead of `/ sqrt(NARM)` produces a plausible wrong answer rather than an error, nothing marks `KAPPA_EMAX` as weighted, and the divisor has to be repeated in every expression the kappa enters.

Closes #1031. Sibling of #1029 (`weight =` on the residual); the issue asked the two to share syntax, validation and implementation, and they do.

## What changed

Declare it where the rest of the variance structure is declared:

```
kappa KAPPA_EMAX ~ 2.0 (sd) weight = NARM
```

meaning `κ_ik ~ N(0, Ω_IOV / N_ik)`, with the structural expression back to `... + ETA_EMAX + KAPPA_EMAX`.

**The engine applies it by reparameterisation, not by scaling the prior.** Every reference to a weighted kappa in `[individual_parameters]` is rewritten `KAPPA` → `KAPPA / sqrt(W)` at parse time. With `κ ~ N(0, γ²)` the rewritten effect is distributed exactly `N(0, γ²/W)`, so the declared model *is* the hand-written one it replaces. Three things fall out of that:

- every estimator, analytic sensitivity kernel and diagnostic sees a model shape it already supports — no per-estimator plumbing, no new FD fallbacks;
- the estimated Ω_IOV stays the **unweighted γ²**, the quantity a published analysis quotes;
- κ̂ EBEs — and their shrinkage, which compares `Var(κ̂)` against Ω_IOV — stay on Ω_IOV's own scale.

The rewrite runs before covariate collection and before the pk/tv closures are built, so a covariate named only by a weight expression is registered like any other (the #484/#1029 time-varying-snapshot trap) with no extra plumbing.

Guards, each closing a silent-wrong-answer path:

- the weight may reference **covariates, thetas and `TIME`** — never a random effect (that would make Var(κ) a function of κ);
- a weighted kappa may be referenced **only in `[individual_parameters]`** — elsewhere it would read as the unweighted κ, so the parser rejects it by name across every other block;
- `block_kappa` cannot carry a weight (the divisor would have to apply to the covariances for the matrix to stay valid) — message says to declare a diagonal `kappa`;
- `weight =` on any other `[parameters]` declaration is **rejected rather than dropped**: the declaration regexes are unanchored, so the modifier would otherwise sit past the end of a successful match and vanish — exactly the failure mode this feature exists to remove;
- a zero / negative / non-finite weight at any observation **or dose** record is `E_KAPPA_WEIGHT_NONPOSITIVE` up front (the blank arm-size cell in an MBMA dataset) instead of a division by zero mid-fit; a weight that moves *within* an occasion warns (`W_KAPPA_WEIGHT_VARIES_WITHIN_OCCASION`) — κ is drawn once per occasion, so `Ω_IOV/W` is ambiguous, but a dropout-adjusted N may be intended.

Mu-reference detection runs on a **pre-rewrite snapshot** for weighted models only (byte-identical detection for every existing model): the matcher pattern-matches `exp(Eta(a) + Eta(b))`, and wrapping a kappa in a division could mask the *BSV* eta sharing that `exp(...)`, silently costing SAEM/IMP their closed-form θ step.

Reporting: the fit prints the effective SD at the median arm next to γ², and `FitResult` carries `kappa_weights` (source text) and `kappa_weight_typical` (median arm size) through the YAML and the `.fitrx` bundle.

```
--- KAPPA (IOV) Estimates ---
  KAPPA_CL             = 1.840240  (CV% = 135.7)  SE = N/A
                         weight = NARM  →  SD = 0.0678 at NARM = 400.0000 (κ ~ N(0, KAPPA_CL/NARM))
```

## Alternatives considered

**Scale the prior instead of the effect** — carry a per-occasion `Ω_IOV/W_ik` through the inner-loop quadratic form and log-det, the FOCE Hessian, SAEM/IMP/Bayes samplers, shrinkage and the covariance step. That is the literal reading of the declaration, and it touches a dozen call sites, each an independent chance to miss one and get a *silently* mis-specified variance. It also reports EBEs on the weighted scale, so shrinkage needs its own weighting to stay meaningful.

**Scale the κ values where they enter each predictor** — one edit per value funnel (`pk::predict_iov`, both sens providers, the likelihood, SAEM, Bayes), plus a matching factor on every `∂/∂κ` seed. Same multiplication of sites, with the derivative seeds as an extra silent-failure surface.

The parse-time rewrite is one site, and its correctness is a statement about a distribution rather than about a dozen call sites agreeing.

## Cross-repo dependency

`FitResult` gained two fields, and `ferx-r/src/rust/src/lib.rs` builds `FitResult` literals in three places (~1749, ~3379, ~3736), so a ferx-r follow-up must add `kappa_weights: Vec::new(), kappa_weight_typical: Vec::new()`. Local ferx-r builds pick this up via the `[patch]` immediately; ferx-r CI only after its `Cargo.lock` is bumped. That PR can also surface the weight and the effective SD in `ferx_model_inspect()`, which is the reporting half of the issue.

## Breaking changes
- [ ] None

<details>

No `.ferx` migration: `weight =` is new syntax and every existing model parses unchanged. (A model that already carried a stray `weight = ...` on a `[parameters]` line was silently ignoring it and is now an error — that model was not doing what it looked like.)

Rust API: `CompiledModel` gains `kappa_weights: Vec<Option<KappaWeight>>`; `FitResult` gains `kappa_weights: Vec<Option<String>>` and `kappa_weight_typical: Vec<Option<f64>>`. Struct-literal constructors need the new fields (`Vec::new()` for a model with no weighted kappa). New public items: `KappaWeight`, `KappaWeightFn`, `CompiledModel::has_weighted_kappa`.

Serialised output is unchanged for every model without a weighted kappa: both `FitResult` fields are left empty and are `skip_serializing_if = "Vec::is_empty"`, and the `.fitrx` IOV wire fields are `#[serde(default)]` so older bundles still load.

</details>

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0** (`pmx` container), control stream + outputs committed at `tests/nonmem/kappa_weight_iov.{ctl,lst,ext}`, data at `tests/nonmem/kappa_weight_iov.csv` (`data/warfarin_iov.csv` + an arm-size column: `NARM` = 400 on occasion 1, 25 on occasion 2 — a 16× spread in N, so 4× in κ's effective SD).
- [x] Reference values:

NONMEM has no `weight =` on an `$OMEGA`, so its equivalent model writes the scaling into `$PK` by hand — which is exactly what the declaration desugars into. Converged FOCE from the same starting values:

```
# NONMEM 7.6.0                         # ferx (weight = NARM)
OFV (without constant)  207.455        205.683
TVCL                    0.31727        0.32464
TVV                     8.4018         8.4445
TVKA                    2.6180         2.4467
omega^2(CL)             0.59721        0.61281
omega^2(V)              0.012400       0.012779
omega^2(KA)             1.0571         0.89725
gamma^2 (Omega_IOV)     2.0013         1.8402
sigma^2 (proportional)  0.037588       0.040884
```

ferx reaches a **lower** objective from the same start, on a surface where κ is weakly identified (it enters divided by √400 and √25).

The residual difference is the pre-existing cross-engine FOCE/IOV gap, **not** the weighting — evaluated at *identical fixed* parameters it is smaller with the weight than without:

| Parameters | Model | NONMEM | ferx | Δ |
|---|---|---|---|---|
| inits, γ² = 0.01 | weighted `κ/√NARM` | 466.779 | 466.720 | 0.06 |
| inits, γ² = 0.01 | plain `κ` | 332.265 | 331.620 | 0.65 |
| NM MLE, γ² = 2.00 | weighted `κ/√NARM` | 207.455 | 205.943 | 1.51 |
| NM MLE, γ² = 2.00 | plain `κ` | 238.889 | 238.524 | 0.37 |

The gap tracks how strongly κ enters the marginal, not whether it is weighted.

The declaration itself is exact, and that is pinned directly rather than inferred: `weight = NARM` and the hand-written `KAPPA_CL / sqrt(NARM)` score **bit-identically** (≤ 1e-8 relative) under FOCE, FOCEI, Laplace and SAEM, and the analytic `∂f/∂κ` matches finite differences of the production `predict_iov` to tolerance, per occasion, with the per-occasion `1/√W` factor asserted explicitly. (IMP/IMPMAP are absent from the estimator list because they reject IOV outright — a pre-existing limitation, untouched here.)

## Performance
- [x] No significant impact expected — the rewrite is parse-time; the fitted model is the same expression tree a user would have written by hand.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3232 tests)

Tier 1:
- `parser::model_parser` (15): the `κ/√W` equality against the hand-written model, per-record weight reads, Ω_IOV left on the unweighted scale, covariate registration, BSV mu-ref preservation, and every rejection (η in the weight, empty RHS, non-kappa declaration, `block_kappa`, reference outside `[individual_parameters]`, plus the unweighted-kappa control that must stay unrestricted).
- `api::validation` / `api::postfit` (10): non-positive / non-finite weight on an observation and on a dose record, the within-occasion warning firing exactly once, the median-over-occasions typical arm, and the no-op paths for an unweighted model.
- `sens::provider` (2): full provider vs FD of `predict_iov` over `[η_bsv, κ_g0, κ_g1]` on a weighted model, and `∂f/∂κ` equal to the unweighted model's scaled by `1/√W` per occasion.
- `io::output`, `io::fitrx`: the printed weight line (both arms) and the `.fitrx` round-trip.

Tier 2 (`tests/weighted_kappa_iov.rs`): declared-vs-hand-written objective equality under every IOV-capable estimator, the weight actually moving the objective, and the reported weight + typical arm size.

Tier 3 (`tests/kappa_weight_nonmem.rs`, `slow-tests`): the NONMEM anchor above.

- [x] Example added to `examples/` or inline below — inline; the runnable anchor model lives at `tests/nonmem/kappa_weight_iov.ferx`.

<details>
<summary>Example</summary>

```toml
[parameters]
  kappa KAPPA_EMAX ~ 2.0 (sd) weight = NARM

[individual_parameters]
  LEMAX = LEMAX0 + log(OR_ABATA) * ABATA + ETA_EMAX + KAPPA_EMAX

[covariates]
  NARM continuous     # number of subjects in this arm
```

```
--- KAPPA (IOV) Estimates ---
  KAPPA_EMAX           = 4.145382  SE = 0.412000
                         weight = NARM  →  SD = 0.1440 at NARM = 200.0000 (κ ~ N(0, KAPPA_EMAX/NARM))
```

</details>

## Docs
- [x] `docs/` updated for user-visible changes — new "Sample-size-weighted IOV" section in `docs/model-file/iov.qmd` (syntax, why declare rather than write, what is reported, rules and diagnostics, and the NONMEM validation tables), the `kappa` syntax line in `docs/model-file/parameters.qmd`, and a cross-link from the residual `weight =` section in `docs/model-file/error-model.qmd`.
- [x] New pages linked in `docs/_quarto.yml` — N/A, no new page (sections on existing pages; anchors verified against a local `quarto render`, which was not committed).

## Checklist
- [x] `cargo clippy` clean — no new warnings from any file this PR touches.
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the rewrite emits ordinary `BinOp(Div)` / `UnaryFn("sqrt")` nodes that the existing symbolic differentiator and bytecode compiler already handle; the provider-vs-FD parity test above is the proof.

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → `ferx-site/model-dsl/` page updated or PR opened — **not done**: `weight =` on a kappa needs a site page update, to follow with the ferx-r PR.
- [x] New example `.ferx` file added to `examples/` → N/A (no new `examples/` model; the anchor model is a test fixture).
- [x] New data file added to `data/` → N/A (the anchor dataset lives under `tests/nonmem/`, like every other NONMEM anchor).

### ferx-book
- [ ] New estimator, DSL feature, or fit option → relevant book chapter updated or PR opened — to follow.

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core — deliberately deferred; see Cross-repo dependency.
- [x] All affected `examples/*.ferx` run cleanly via CLI — no `examples/` model is affected; the weighted anchor was run end-to-end through the CLI (`cargo run --release --bin ferx -- tests/nonmem/kappa_weight_iov.ferx --data tests/nonmem/kappa_weight_iov.csv`) and produced the estimates in the table above.
- [ ] All affected ferx-site example `.qmd` pages render cleanly — none affected.

## Reviewer hints

The load-bearing decision is in `parse_full_model`, right after `indiv_stmts` is parsed: `rewrite_weighted_kappas` plus the two things around it — the mu-ref snapshot and the "not outside `[individual_parameters]`" scan. If the rewrite is right, everything downstream is the model that was already supported, which is why there is no per-estimator code in this diff.

Worth a look:

- `rewrite_weighted_kappas` deliberately does **not** use `visit_expr_nodes_mut`: that walker recurses into a replacement node, and replacing `Eta(k)` with an expression still containing `Eta(k)` would re-fire forever. The hand-rolled walk descends only into nodes it did not create.
- The block scan is textual (by kappa name, uppercased) because the compiled form differs per block while the name does not. `[parameters]` and `[individual_parameters]` are excluded; everything else, named blocks included, is scanned.
- `split_weight_modifier` now runs over *every* `[parameters]` line before the declaration regexes. Whole-word + single-`=` matching means `theta WEIGHT(70)` and `weight == 2` are untouched, and comments are already stripped upstream; a weight on a non-kappa line raises rather than disappears.

Skimmable: the ~15 mechanical `kappa_weights: Vec::new(),` additions to `CompiledModel` / `FitResult` struct literals in test helpers, and the committed NONMEM `.lst`/`.ext`.

## Open questions

The within-occasion-variation check is a **warning**, not an error. κ is drawn once per occasion, so a weight moving underneath it makes `Ω_IOV/W` genuinely ambiguous — but a dropout-adjusted N is a plausible thing to want, and the model still evaluates (each record is scaled by its own weight). Happy to promote it to an error if you'd rather the ambiguity be unrepresentable.


roninsightrx:
## Code review — sample-size-weighted IOV (`kappa K ~ γ² weight = W`)

Reviewed the full diff (parser rewrite, validation, postfit, output, fitrx, types, tests).

The core design is sound: desugaring `KAPPA` → `KAPPA / sqrt(W)` inside `[individual_parameters]` means every downstream consumer (sens, SAEM, IMP, shrinkage, output) sees an ordinary kappa with no special-casing. The `Expression`/`Statement` walk in `rewrite_weighted_kappas` covers every child-bearing variant, the eta-slot mapping (`n_eta + k`) matches `names_ordered`, and the rewrite lands before covariate collection and before every consumer of `indiv_stmts`.

Seven findings below.

### Medium

**1. `src/parser/model_parser.rs:1590` — mu-ref detection runs on the pre-rewrite snapshot**

`mu_ref_stmts` is snapshotted *before* `rewrite_weighted_kappas`, so it registers mu-refs that the actual (rewritten) model does not satisfy — which breaks the PR's own "declared form == hand-written form" invariant.

Concretely, `CL = TVCL * exp(KAPPA_CL)` with `weight = NARM` detects a **kappa** mu-ref on the snapshot, while the hand-written twin `exp(KAPPA_CL / sqrt(NARM))` detects none. `estimation/bayes.rs:423` then rejects the declared model (`kappa mu-references are not yet supported`) while accepting the hand-written one.

The rationale inverts in order-dependent `find_exp_eta_in_mul` cases: for `CL = TVCL * exp(KAPPA_CL) * exp(ETA_CL)` the snapshot returns the *kappa* index (left-to-right `.or_else`), so the BSV `ETA_CL` mu-ref that the post-rewrite statements would have found is lost, and SAEM/IMP drop to the numerical M-step.

Detecting on the rewritten statements is correct for every case I could construct — `(a, b) => a.or(b)` already tolerates the non-bare kappa operand.

**2. `src/api/validation.rs:296` — weight is only checked at the initial theta**

`check_kappa_weights` evaluates the weight at `model.default_params.theta` only, but the parser deliberately permits thetas in the weight expression ("may depend only on covariates, thetas and TIME"). A weight such as `weight = TVN * NARM` that is positive at the initial θ but crosses zero during the outer optimization yields `sqrt` clamped to 0 → `κ/0` → an infinite individual parameter, with no diagnostic — exactly the silent failure this check exists to prevent.

Either restrict the weight to θ-free expressions, or re-check/guard at eval time.

**3. `src/api/simulate.rs:381` (and `:1011`) — the weight check never runs on the simulation path**

`simulate()` runs only `check_covariates`, never `check_model_data_rule`, so `E_KAPPA_WEIGHT_NONPOSITIVE` never fires when simulating. A blank or zero `NARM` cell in a simulation dataset — the exact MBMA scenario the error was written for — produces infinite parameters and NaN/Inf simulated DV instead of the up-front error.

### Low

**4. `src/parser/model_parser.rs:1518` — an unreferenced weighted kappa is silently inert**

A weighted kappa never referenced in `[individual_parameters]` gets no rewrite. Two consequences: (a) the weight is silently ignored while `print_results` still prints the `→ SD = γ/√N` line, implying the scaling was applied; (b) a covariate named *only* by the weight expression never reaches `referenced_covariates`, so the datareader does not read the column, `obs_cov` lacks it, `eval_expression` returns `0.0`, and `check_kappa_weights` reports a misleading `E_KAPPA_WEIGHT_NONPOSITIVE` ("weight evaluates to 0 … missing or non-numeric cell") for a column that *is* present in the CSV.

**5. `src/api/postfit.rs:772` — doc comment detached from its function**

`kappa_weight_typicals` was inserted between `compute_kappa_shrinkage`'s doc comment and its `fn`, so the shrinkage docs (`1 - sqrt(mean(κ̂²)) / …`, "Returns NaN for a given kappa when …") now document `kappa_weight_typicals`, and `compute_kappa_shrinkage` is undocumented. Move the three orphaned `///` lines back down.

**6. `src/api/validation.rs:411` / `:2627` — redundant full-dataset pass**

`check_kappa_weights` is called twice (again at 2627 for the warning half), each doing a full walk over every observation *and* every dose of every subject, invoking a boxed closure that allocates a fresh `HashMap` per record (`model_parser.rs:2472`). On a large MBMA dataset that is a redundant second full pass; returning both outputs from one walk (or caching) avoids it.

**7. `src/parser/model_parser.rs:1562` — error names the block instance, not the block type**

In the "weighted kappa outside `[individual_parameters]`" scan, `named_blocks` yields `(instance_name, lines)` from `extracted.named.values().flat_map(|m| m.iter())`, so the error prints e.g. ``[armA] references `KAPPA_EMAX` `` — the block instance name rather than the block type (`event_model armA`). Confusing when the instance name doesn't resemble a block name.

Issue #811, opened 2026-07-12
Feature:  compartment-free custom structural model (`$PRED`-equivalent)


Robterheine:
**Type:** Feature request (new capability, not a bug)
**Scope:** `src/parser/model_parser.rs` (structural model parsing + compiled-model dispatch), `src/types.rs` (`CompiledModel`/`PkModel`), estimation call sites that dispatch on structural model kind. Motivated by model-based meta-analysis (MBMA) workflows, but applies to any use case needing a closed-form equation with no ODE/compartments.

## Motivation

`[structural_model]` today only accepts two things:

1. A whitelisted PK compartment template — `pk one_cpt_oral(cl=CL, v=V, ka=KA)` etc. Verified at `src/parser/model_parser.rs:3893` (`parse_structural_model`), which regex-matches `pk\s+(\w+)\(...\)` and maps the name to one of nine hardcoded `PkModel` variants (`src/types.rs:549-559`: `OneCptIvBolus` … `ThreeCptInfusion`), erroring `"Unknown PK model: {}"` for anything else.
2. A full ODE system — `ode(states=[...])` with a `[odes]` block of `d/dt(state) = expr` equations, parsed separately (per the model-file docs under `docs/src/model-file/ode-models.md`).

There is a `[scaling] y = <expr>` readout (Form C) that lets you write an arbitrary output expression — but it's a post-transform *on top of* ODE state variables, and the parser enforces that an ODE model is present to use it. There is no way to declare a structural model that is just a single closed-form equation with **zero compartments** — the NONMEM `$PRED`-block equivalent (e.g. an Emax dose-response, a log-linear dose/exposure relationship, or a published meta-regression equation combining covariates and fitted parameters with no underlying compartmental PK).

This blocks model-based meta-analysis (MBMA) style modelling, where the "structural model" is frequently just a regression-like equation with covariates pulled from the literature dataset, not a PK compartment model at all.

## What already exists that this can be built on

This is a smaller lift than it might look, because the pieces already exist — mostly independently, in different subsystems that just haven't been connected for this use case:

[...]

PR #1063, opened 2026-08-24
feat(parser): compartment-free `$PRED`-equivalent structural model [#811]

roninsightrx:
Design was posted for feedback in #811 first; this is that design implemented.

## Why

`[structural_model]` accepted exactly two things: a whitelisted `pk NAME(...)` closed form, or an `ode(...)` system with an `[odes]` block. A model whose prediction is simply an equation — no compartments, no dosing — could not be written at all.

That is the shape of every model-based meta-analysis structural model surveyed in #811 and #1032: an Emax time-course on a WOMAC pain score, a logit-Emax on an ACR20 responder proportion, a dose-response curve. It is also the NONMEM `$PRED` block, which had no ferx equivalent.

The workaround users write today (noted in #1032) is a compartment nobody uses, driven by `d/dt(clock) = 1`, with the real equation hidden in a `[scaling] y = <expr>` readout. This deletes that state.

### The form

```
[structural_model]
  EFF = EMAX * TIME / (ET50 + TIME)
  y   = E0 - EFF
```

A block with no `pk` / `ode` / `ode_template` line and at least one `NAME = <expr>` line declares its prediction directly.

**No new keyword**, deliberately: every other assignment-bearing block in the DSL (`[individual_parameters]`, `[odes]`, `[scaling]`, `[derived]`) is a bare `NAME = expr` list, so a marker here would be the only one — and `y = E0 - EMAX * TIME / (ET50 + TIME)` already says what it is.

### Implementation: desugaring, not a third readout

The equation lines are **moved into the `[scaling]` block** at parse time and the structural block is replaced by an internal marker — the same rewrite `apply_ode_template` performs for `ode_template`. Everything the Form C pipeline already does then applies with no second implementation: named intermediates and continuation lines (#1030/#1056), the `TIME` built-in and undefined-name handling (#1028) — an undeclared name is a required **data column**, which is how an MBMA equation reaches the per-row covariates it regresses on — covariate registration, the direct-θ/η desugaring (#486), and the `OdeOutputProgram` sensitivity program.

Diagnostics from that pipeline are retargeted at `[structural_model]`, so no message names a block the user never wrote.

### The kind is derived, not stored

An `AnalyticReadout` with an **empty `state_names`** *is* the marker, and only the parser can build one (`#[non_exhaustive]`; every compartment-bearing readout carries at least `central`). `CompiledModel::is_algebraic()` reads that, so it cannot drift out of step with the readout it describes — and no `CompiledModel` literal anywhere in the tree needed touching.

### Parameter layout follows the ODE path

An analytical Form C readout squeezes its parameters into the spare region `allocate_readout_extra_slots` owns — at most 11 slots, typically 4–7. A compartment-free model has no closed form, so every individual parameter is assigned from the full `MAX_PK_PARAMS` layout via `ode_param_slots`, exactly as an ODE model's are. `build_pk_param_fn` already keys its ODE-vs-analytical branch on an empty `pk_param_map`, so it needed no change.

This is load-bearing: @billdenney's unstructured-placebo strategy would hit an 11-parameter wall immediately. There is a test at 40. (His follow-up — *hundreds* of such parameters — is out of scope here and filed as #1064 / #1065.)

### Prediction

Short-circuits both funnels (`compute_predictions_with_tv_into_with_schedule`, `predict_iov`): no integration, no dose resolution, no closed form.

**Under IOV the readout is evaluated per occasion**, not once from the BSV η. Every other analytic Form C readout is BSV-only because the *concentration* carries the per-occasion κ and the readout merely maps it. A compartment-free model has no concentration, so the same convention would have silently dropped κ from the prediction entirely — which matters directly for the BTAV models in #1031.

### Analytic gradients

A compartment-free model is the degenerate case of the Form C readout jet, and the easiest one in `sens/`: `∂y/∂θ = (∂y/∂p)·(∂p/∂θ)`, with no state jet, no dose walk and no integrator. Both orders come out of one dual evaluation per observation, so the outer walk yields the η-η Hessian and the η-θ cross block FOCEI needs without a second pass.

The new `sens/algebraic.rs` reuses the ODE provider's seeders and the shared `obs_sens_from_dual2` scatter rather than growing a branch inside the 5.5k-line closed-form provider. Three details carry the FD parity, each mirroring a specific line of `pk::apply_analytic_readout`:

1. the per-observation-vs-static snapshot cadence,
2. the two different clocks — parameters on the observation time, the readout on the raw data-file time (#1028),
3. the output transform (LTBS), through the shared generic so the floor-then-log derivative cannot drift from the f64 path.

Declines — at the *gate*, not at a dispatch fallthrough — for IOV, per-CMT readouts, non-dual-evaluable readouts, and past the 24-axis cap it shares with `MAX_ODE_AXES`.

### The thing that must not be fudged

`analytical_supported_core` matches on concrete `PkModel` variants, and a compartment-free model carries a **placeholder** that the match would happily admit. Every closed-form sensitivity walk would then evaluate a solution the model does not have, against doses it does not carry. It declines explicitly, before the match — the route/report drift #637 exists to prevent.

The rest of the `.pk_model` audit is complete. Four readers needed a guard; the others are already inert (they early-return on transit/IG, or iterate doses this model does not have):

- `suggest_start::inits_from_nca` would have run non-compartmental analysis on a dose-response curve and "estimated" a clearance from it.
- `compute_predictions_with_states` and the `[derived]` grid in `output_columns` would have reported all-zero one-compartment amounts as this model's state; they now return empty → NaN, the convention every other out-of-scope case uses.
- `analytical_compartment_names` advertised `["central"]`, so a bare `central` in `[derived]` resolved to a compartment reference that reads NaN forever instead of being reported as the undefined name it is.
- the `EventSchedule` caches in `inner_optimizer` / `bayes` built a schedule for a walk the predictor never takes — and time-varying covariates are the norm here, so that was the common path, not an edge case.

`W_NO_DOSES` is dropped for such a model at the fit boundary: the reader is model-blind and flags a dose-free dataset as a probable missing `AMT` column, which is every MBMA fit.

### Rejections, by name

`[odes]`, `[initial_conditions]`, `[diffusion]` and `[scaling]` alongside a compartment-free model; `obs_scale` (there is no built-in prediction to divide); a compartment amount in the equation (`y = central / V` says "you forgot the `pk` line", rather than falling through to an undefined identifier or a silent covariate lookup).

### Block tightening (behaviour change)

Implicit detection forces a line classifier, which closes two latent silent-wrong-model bugs:

- the `pk` matcher was **unanchored**, so `zpk one_cpt_iv(cl=CL, v=V)` parsed as a valid one-compartment IV model;
- `parse_structural_model` returned on the **first** `pk` match and **ignored every other line**, so a stray or mistyped line was discarded without a word.

Every line is now classified as one of the four accepted forms; an unrecognized line, a second disposition line, or a mix of a compartment model with equation lines is an error naming the offending line. Same "reject, don't silently default" class as #261 / #1028 / #1040. All 3259 existing lib tests pass unchanged, so nothing in-tree relied on the leniency.

## Alternatives considered

- **`algebraic PRED = <expr>` keyword, as the issue proposed.** Single-line only, inconsistent with `[scaling]`'s `y`, and gives up intermediates — which is what makes a bounded-endpoint (weighted-logit) readout maintainable rather than a 200-character line.
- **A degenerate zero-state ODE.** Maximal reuse, but keeps the integrator, the solver diagnostics and the dose-routing machinery for a model that has none of them.
- **A new `StructuralModelKind` enum on `CompiledModel`.** ~29 struct-literal edits and a second source of truth next to the readout's own state list.
- **FD gradients in v1** (the issue's own scoping). Rejected once it was clear the zero-state case is the *easiest* sensitivity, not the hardest — and that a standalone module avoids touching the closed-form provider at all.

## Cross-repo dependency

No public API signature changed; `CompiledModel::is_algebraic()` is additive. `data/emax_timecourse.csv` is a new example dataset — mirroring it into `ferx-r/inst/examples/data/` is a follow-up, not a merge blocker.

## Breaking changes

- [ ] Public Rust API
- [ ] None

<details>

A `[structural_model]` block containing lines other than its one disposition directive is now a parse error. Such a file was already being read as something other than what it said — the extra lines were silently discarded, and a mistyped `pk` line still matched. The error names the offending line and lists the accepted forms.

</details>

## Numerical validation

- [x] Compared against: **NONMEM 7.5.1 `$PRED`** (`nonmem_anchor/algebraic_emax.{ctl,csv}`, results in `nonmem_anchor/results/`), the `d/dt(clock) = 1` twin this replaces, and the hand-evaluated closed form
- [x] Reference values:

```
30 subjects x 8 dose-free timepoints; both engines converged independently
from the same starting estimates.

                ferx          NONMEM 7.5.1
  OFV           62.5385       62.5384884
  TVE0          10.203867     10.2040
  TVEMAX        5.917646      5.91754
  TVET50        2.099437      2.09949
  omega^2(E0)   0.092357      0.0924622
  sigma (SD)    0.480452      0.480452
  SE TVE0       0.572367      0.572542
  SE TVEMAX     0.101099      0.101121
  SE TVET50     0.144743      0.144760
  SE omega^2    0.023942      0.0239849
  SE sigma      0.023437      0.0234361
```

~1e-4 relative on every estimate and standard error, 5 decimal places on the objective. `$PRED` is the exact analogue, so the control stream is a direct translation rather than an approximation of one. SEs are compared against `$COV MATRIX=R` — ferx's covariance is a pure `R⁻¹`, and NONMEM's default sandwich would not be comparable. They are also the part that exercises the second-order blocks of the analytic jet, so a wrong `∂²f/∂η²` or `∂²f/∂η∂θ` shows up there even when the estimates agree.

The **anchored model is the shipped example** (`examples/emax_timecourse.ferx`), and the two copies of the dataset are pinned byte-equal by the test — so what a user runs from `examples/` is the thing NONMEM was compared against.

Also validated by known-truth recovery (the data was simulated from θ = (10, 6, 2), ω² = 0.09, σ = 0.5) and by equivalence with the dummy-ODE twin.

## Performance

A compartment-free model skips integration entirely and takes the analytic gradient. Existing models are unaffected — the added work on their path is one line-classification pass per parse. No integrator is monomorphised behind the new dual ladders, so the compile-cost sensitivity in #971 is untouched.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3259 passed, 0 failed)

**Tier 1, parser** (`src/parser/model_parser_tests.rs`, 13 tests): parses without `pk`/`ode`; named intermediates; requires a `y` entry; rejects `obs_scale`, compartment references, `[scaling]`, and each compartment-only block; free names register as covariates; the mistyped-`pk` and stray-line regressions; mixed and duplicated disposition lines; 40 readout parameters.

**Tier 1, sensitivities** (`src/sens/algebraic_tests.rs`, 14 tests): every block of the jet against central FDs of the production predictor — with intermediates, a covariate, time-varying covariates, a nonlinear readout, and LTBS; the inner `Dual1` walk against the outer jet; every dispatch arm up to the cap; and both directions of the scope gate (analytic reported *and* taken; IOV / past-cap declining at the gate rather than at a fallthrough).

**Tier 2** (`tests/algebraic_ode_equivalence.rs`): dummy-ODE twin equivalence, hand-computed equation, bounded fit smoke.

**Tier 3** (`tests/algebraic_pred_nonmem_anchor.rs`, slow-gated): the NONMEM `$PRED` anchor and known-truth recovery.

## Example

- [x] Example added to `examples/emax_timecourse.ferx` (+ `data/emax_timecourse.csv`)

```
cargo run --release --bin ferx -- examples/emax_timecourse.ferx --data data/emax_timecourse.csv
```

Runs clean, converges, covariance step succeeds, no warnings.

## Docs

- [x] `docs/` updated — a new *Compartment-Free Models (the `$PRED` equivalent)* section in `docs/model-file/structural-model.qmd` (the form, the grammar it shares with Form C, the rejected-blocks table, the worked example with its NONMEM translation, the anchor table, and the gradient scope), plus a callout in `scaling.qmd` pointing Form C users at it
- [x] `CHANGELOG.md` `[Unreleased]` entry added (one under Added, one under Changed for the block tightening)

## Checklist

- [x] `cargo clippy` clean for this diff (the 4 pre-existing `approximate value of PI` errors in `src/sens/two_cpt_explicit.rs` are untouched)
- [x] Functions on the `Dual2` sensitivity path stay differentiable, and the new kernel carries its `Dual2`-vs-FD parity tests

### ferx-site
- [ ] New `.ferx` DSL syntax → `ferx-site/model-dsl/` page — **follow-up**
- [ ] New example `.ferx` + data file → matching entries in ferx-site examples and ferx-book — **follow-up**

### Example execution
- [x] Rebuilt ferx-core (`cargo build --release`) and ran `examples/emax_timecourse.ferx` via the CLI

## Reviewer hints

Focus on:

1. **`sens/algebraic.rs`'s "Matching production exactly" section** and the three details it names. Those are what the FD-parity tests actually verify, and the two-different-clocks one is the easiest to get wrong by reading only `apply_analytic_readout`'s signature.
2. **`analytical_supported_core`'s early decline** (`src/sens/provider.rs`) and the four audited `.pk_model` guards. Without them a placeholder is dispatched on and every walk evaluates the wrong solution.
3. **The per-occasion IOV readout** in `predict_iov`. It mirrors the `[scaling]` per-occasion loop directly above it, including the `subject.occasions.is_empty()` guard (not `occ_groups`, for the dose-only-occasion reason documented there).
4. **The block tightening.** The only part that can break an existing file. Both closed bugs were silent, so anything it breaks was already being misread — but a second opinion on the classifier's four forms is worth having.

## Open questions

Carried from the design discussion in #811, and now narrowed by what landed:

1. **κ / IOV gradients fall back to FD.** Predictions are per-occasion correct; only the analytic gradient declines. Acceptable for a first release, given every BTAV model (#1031) carries a κ? Lifting it means per-occasion dual seeding in the same module — contained, but not free.
2. **The 24-axis cap** is shared with `MAX_ODE_AXES` because the inner walk's `∂p/∂η` comes from `param_derivatives_at_cov`, whose own table is bounded by it. Raising it is cheaper here than there (no integrator behind these ladders), but the two must move together. @billdenney's hundreds-of-parameters case is filed separately as #1064 (sparsity) / #1065 (profiling) and is not what this cap is about.
3. **The naproxen/OA case study** as a shipped example — #1032 has the datasets. Follow-up, or worth adding before release?


roninsightrx:
Pushed two follow-ups from the open questions, which are now closed.

### 1. IOV gradients are analytic (`7d422fab`)

They were falling back to FD whenever the model carried a `kappa` — which is the case MBMA actually needs, since a between-treatment-arm variance component (#1031) is an arm-level random effect and most published models have one.

The IOV jet turns out to be the ordinary readout jet seeded on the stacked `[η_bsv, κ_0…κ_{K−1}]` axes: no walk to run, no carryover to carry. Such a model has no concentration, so κ reaches the prediction only through the individual parameters the equation reads — which is exactly why `predict_iov` already evaluates the readout per occasion for it, and why these walks seed each observation's **own** occasion block rather than dropping the κ axes the way every other analytic readout does. A BSV-only jet would report `∂y/∂κ = 0` for a model whose entire between-arm structure lives there, and every κ would sit at its initial value while the fit reported an analytic gradient.

Widths are bucketed against `ODE_IOV_WIDTH_BUCKETS` (cap 96), because the stacked count grows with the occasion count — a study with many arms is exactly that — and monomorphising 96 widths is the compile-cost trap #971 measured. The per-occasion derivative source stays under the ordinary 24-axis table, checked separately in `supported_iov` so a model past it declines at the gate rather than at a dispatch fallthrough.

One trap worth flagging for review: under IOV the individual-parameter program is built over the **combined** `[η_bsv, κ]` vector, so its η axis is `n_eta + n_kappa`, not `n_eta`. Checking it against `n_eta` — which is what the BSV walks correctly do — silently declines every IOV model. Same convention as `iov_analytical_supported`.

### 2. The published MBMA case study is now an example and a test (`7d422fab`)

Naproxen vs placebo in osteoarthritis — Case Study 1 of Bracis et al., *CPT:PSP* 2026;15:e70158, 18 trials, 36 arms, longitudinal WOMAC pain. It is the model class this feature exists for, and it had to be written as `d/dt(clock) = 1` until now.

| | ferx | Monolix (published) |
|---|---|---|
| naproxen effect on Emax | 0.7930 (SE 0.0646) | 0.792 (SE 0.064) |
| ET50, placebo (weeks) | 0.6901 | 0.698 |
| ET50, naproxen (weeks) | 0.2132 | 0.216 |
| ω(E0), SD | 0.6215 | 0.627 |
| ω(Emax), SD | 0.7387 | 0.763 |

The structural model is now three lines:

```
[structural_model]
  EFF = EMAX * TIME / (TIME + ET50)
  y   = (E0 - EFF) / WPSE
```

The `/ WPSE` is not a unit conversion — it is the weighting scheme, together with `sigma ADD_ERR ~ 1.0 (variance) FIX`. `DV` is the arm mean divided by its reported SE, the prediction is divided by the same SE, so each arm enters the likelihood with weight `1/SE²`. The test asserts sigma is actually fixed: unfix it and the whole thing silently degrades to an ordinary homoscedastic fit that still converges and still looks reasonable.

It also fits the dummy-clock workaround to convergence alongside and requires the two to agree — the migration guarantee for models already written the old way.

### A note on how the IOV equivalence is tested

My first attempt compared two *fits* of the IOV pair at `maxiter = 5` and they disagreed (OFV −147.9 vs −110.3). That was the test's fault, not the code's: the compartment-free model now takes the analytic gradient while the dummy-ODE twin does not, so five iterations from the same start land in different places. The equivalence claim is about the *prediction*, so it is now compared as `predict_iov` output at a fixed parameter point, with an explicit guard that κ actually moves the prediction — otherwise the test could pass by both engines dropping κ and agreeing on the wrong answer.

### Remaining

Only the axis-cap question is still open, and it is not really about this PR: @billdenney's hundreds-of-parameters case is filed as #1064 (factor/indexed thetas + per-subject sparsity) and #1065 (profiling the nuisance block out).

Verified: 3265 lib tests default / 3478 all-features, 0 failures; `cargo check --all-features --tests`; clippy clean for the diff; all slow-gated tests pass.

roninsightrx:
Merged `main` in and resolved the conflicts (`135c42f3`); the branch is mergeable again.

`main` had moved 15 commits ahead, including **#1031 (sample-size-weighted IOV)** and the #1046 / #994 parser fixes.

### The conflicts

**`CHANGELOG.md`** — two `### Added` entries at the same insertion point. Kept both.

**`src/parser/model_parser_tests.rs`** — both sides appended new tests. Worth recording how I got this wrong first, because the failure mode is quiet: I resolved it with `git merge-file --union`, on the assumption that both sides were clean appends at a shared point. They are not — the conflict regions interleave, so the union spliced main's new tests into the middle of this branch's function bodies. **`cargo check --lib` passed** (the test module is `cfg(test)`, so it was never compiled) and only `cargo test --lib` caught it, as unclosed delimiters ~600 lines from the actual damage.

Redone as *main's version of the file* + *this branch's appended block*, after verifying against the merge base that this branch's diff really is a single insert at EOF. Every function from both sides is present, with no duplicates; the one name that disappears is `dose_attr_read_in_odes_init_is_rejected`, which #1046 deliberately renamed to `..._is_accepted` and which this branch never touched.

### #1031 and this PR meet at the same model

That merge is more than a textual one: #1031 is the between-treatment-arm variance component, and this PR is the compartment-free structural model — a longitudinal MBMA wants both at once, and until now neither existed.

A weighted kappa is applied by rewriting every reference as `K / sqrt(W)`, so it reaches the per-occasion jet added in `7d422fab` as an ordinary — but now covariate-dependent — individual parameter, and that jet has to carry the `1/√W` factor into `∂y/∂κ`. `ad6ac4e7` asserts the composition rather than assuming it: FD parity against `predict_iov` with deliberately **unequal** arm sizes (400 vs 25), since equal sizes would make the factor a constant and hide a wrong per-occasion scaling. It passes. The compartment-free docs section now cross-references the weighted-kappa declaration too — it is the usual BTAV form, and a reader who has just been told IOV is analytic here will want it.

So the full MBMA shape is now writable end to end:

```
[parameters]
  kappa KAPPA_EMAX ~ 2.0 (sd) weight = NARM      # BTAV, #1031

[structural_model]                                # #811
  EFF = EMAX * TIME / (TIME + ET50)
  y   = (E0 - EFF) / WPSE
```

### Verified after the merge

3313 lib tests pass (this branch's 3265 plus main's additions), 0 failures; all slow-gated tests pass (`algebraic_pred_nonmem_anchor`, `algebraic_ode_equivalence`, `mbma_naproxen_case_study`); `cargo check --all-features --tests`; clippy clean for the diff.

CI: `Check`, `Clippy`, `Format` and CodeFactor green; the two `Tests + coverage` jobs are still running.

roninsightrx:
Review findings addressed in `65d096aa`. All five were real; two of them were bugs I would not have found from the tests I had written.

### 1. A bare theta/eta in the equation silently lost its analytic gradient — `model_parser.rs`

Confirmed and fixed. The guard on the #486 synthetic-acceptance step tested `is_ode`, which is `false` for a compartment-free model even though such a model appends its synthetics on the **ODE-layout** path. The analytical allocator returns an empty `accepted_synths` for it (it has no spare-slot squeezing to do), and that empty list then overwrote the collected synthetics — so the synthetic was appended and consumed a PK slot, while `parse_scaling_block` received nothing to rewrite. `dual_evaluable` stayed false and the model fell to FD behind a warning blaming the readout's shape. Now keyed on `uses_ode_param_layout`.

Worth recording how nearly I mis-cleared this: my first reproduction wrote the reference as `y = ... + THETA(2) * TIME + ETA(1)` and **passed**, which for a few minutes looked like the finding was wrong. It was my test — a direct θ/η in a readout is referenced *by name* (`TVSLOPE`, `ETA_E0`), and `THETA(2)` was parsing as something else entirely. With the right spelling the assertion fails exactly as described. The regression test now also pins the synthetic *count* at one per reference, which is what would catch the mirror-image bug of appending them twice.

### 2. A declared parameter named like a compartment was rejected as a compartment — `model_parser.rs`

Confirmed and fixed, with the misleading message the review predicted:

```
[structural_model] y: `central` is a compartment amount, but this model declares no
compartments … or define `central` in [individual_parameters].
```

— on a model whose `[individual_parameters]` block defines `central`. Names the user declared are now removed from the forbidden set before the check. A real state still takes precedence in the expression parser's scope, so an actual compartment reference is unaffected.

I widened this slightly beyond the compartment-free case: the same false positive existed on the closed-form path, where an individual parameter named `peripheral` was rejected for the same reason. One condition covers both, and leaving the closed-form half broken would have been an odd place to stop.

### 3. `ferx check` and `fit()` disagreed on `W_NO_DOSES` — `api/validation.rs`

Confirmed. `fit()` filtered it; `ferx check` re-emits `population.warnings` wholesale and did not — so every MBMA model got "check that the dataset has an AMT column" from one command and silence from the other, which is precisely the agreement that block's comment says it exists to keep.

Rather than adding a second copy of the condition, both now call `reader_warning_suppressed`. Pinned in both directions: a compartment-free model must **not** be warned, and a compartment model on the same dose-free data still **must** be — the suppression is scoped to models that cannot be dosed, not to dose-free data.

### 4. Stale scope comment — `sens/algebraic.rs`

Correct, and my fault: I updated the module docs when the IOV walks landed and missed the inline comment in `programs()`. Reworded to say which predicate does serve κ and why the two are deliberately separate.

### 5. Dataset provenance and an unrunnable script — `data/`

Both fair. The script now takes the source path as an argument (verified to reproduce the committed CSV byte-for-byte), and `data/mbma_naproxen.README.md` records the citation, the licence position, the column semantics and the weighting scheme. The rows are published *aggregate* summary statistics rather than individual-level data, but they are someone else's work and the repo should say so.

### Verification

3315 lib tests default / 3529 all-features, 0 failures; `cargo check --all-features --tests`; clippy clean for the diff; all slow-gated tests pass. Findings 1 and 2 have Tier-1 tests that fail without their fixes; finding 3 has two Tier-2 `ferx check` tests.

Thanks — the first finding in particular was a silent-FD regression sitting inside the feature's own headline claim.

TeunP:
@roninsightrx @billdenney — input on the three open questions below (axis cap, IOV gradients, the naproxen case study), plus a gap analysis of what I think is left between here and being the best tool for MBMA at the end.


Went through #1063's design against the tree and re-read the thread against the MBMA methodology it's serving. The trace is accurate — I re-verified every code claim (the unanchored `pk` regex at `model_parser.rs:12228`, the first-match return at `:12285`, the `state_names.is_empty()` early return at `pk/mod.rs:255`, the `0..=PK_IDX_MTT` confinement at `:14749`, `analytical_supported_core`'s concrete-variant match at `sens/provider.rs:516`) and they all hold. Deriving the kind from an empty state list rather than storing a flag, and desugaring into Form C rather than adding a third readout, are both the right calls: they're the two decisions that keep this from becoming a fourth parallel funnel.

Answers to the three questions first, then the part I think matters more — what's actually left between here and "best tool for MBMA".

## Q2 — axis cap: keep 24, and decouple the constant

Keep 24. Your own #1064 analysis is the reason: an unstructured-placebo model is sparse, so with `ID` = study a subject touches ~8 of the 800 thetas and the model is an 8-axis-per-subject model. The cap was never what was binding on Bill's case. Raising it to 96 would buy a model that has more than 24 *simultaneously active* axes per subject **and** no sparsity structure to exploit — I can't construct that model from anything in the thread.

Against that, raising it isn't free anywhere else: `MAX_ODE_AXES` carries a `const_assert` requiring four `disp!` / `dispatch_tv!` ladders to widen in lockstep (`sens/provider.rs:98-104`), and `Dual2<M>` is O(M²), so the widest instantiation at 96 is ~16× the work of 24 and monomorphises 96 arms across four tables. Compile time and binary size land on every model, not just the compartment-free ones.

Two things I'd do instead:

1. **Give the algebraic path its own constant** — `MAX_ALGEBRAIC_AXES` initialised to `MAX_ODE_AXES`, with a comment that it may not exceed it (because the inner `∂p/∂η` comes from `param_derivatives_at_cov`, whose table is bounded there). Today it's one line and changes nothing. Later it's the difference between "raise the algebraic cap" and "raise the ODE cap too, and pay for the integrator ladder we didn't need". The coupling should be a deliberate assignment, not an inherited constraint.
2. **Measure the decline rather than asserting it.** "FD above the cap is not catastrophic for a scalar expression per row" is very likely true and is currently unmeasured. One benchmark at ~30 axes in the PR description settles it, and if it turns out to be worse than expected, that's the argument for 96 arriving with evidence attached. (We've twice shipped something whose cost was asserted and not measured; cheap to not do again.)

Also: when the cap declines, the warning should name the axis count and the cap, so a user whose fit just got slower can see why.

## Q1 — κ/IOV to FD: yes, ship it, with three conditions

Ship it. At MBMA sizes the FD cost doesn't bind: a study is a subject, arms per study are 2–4, so κ width per subject is single digits and the population is tens of subjects. Holding the PR that unblocks *writing* these models at all, for a gradient-path optimisation that isn't on the critical path at the target scale, is the wrong trade.

Three conditions, and the second is the one I actually care about:

1. Test the routing in both directions (κ=0 → analytic, κ>0 → FD) **and** assert `build_info` reports FD — the failure mode #637 exists for is the report drifting from the route, not the route itself.

2. **The per-occasion readout needs the twin oracle, not just a unit test.** You flagged it as a design decision, and I'd go further: it's the one place where compartment-free deliberately *diverges* from every other analytic readout. Which raises a question I couldn't settle by reading — if a Form-C `[scaling] y = E0 + KAPPA_X` on an analytical PK model evaluates its readout parameters BSV-only (because the concentration is what carries κ), then the same expression written compartment-free now includes κ and the two forms give different numbers for what looks to a user like the same model. If that's right, it wants a documented note at minimum, and it makes the κ-carrying twin equivalence test load-bearing rather than nice-to-have. Please run the dummy-clock twin *with* a κ before this leaves draft — if the twin disagrees, that disagreement is the finding.

3. File the follow-up now and reference it from the CHANGELOG line, so "IOV routes to FD" is discoverable by someone who hits it rather than being thread archaeology.

One caution on the FD path with MBMA data specifically: `weight = SE` with σ FIXed at 1 makes the individual objective sharply curved, and precisely-weighted large-N arms make it more so. Worth one FD-vs-analytic gradient comparison on an actual weighted model (not just an unweighted Emax) before calling the fallback correct — FD step sizing on that objective is exactly the shape that bit us in #317.

## Q3 — naproxen: split it, but don't ship without an end-to-end number

Follow-up for the shipped example. The PR is already ~4,950 additions across 32 files; adding a case study, its dataset and its docs makes it unreviewable, and the docs/example follow-ups are already carved out in the checklist.

But it shouldn't leave draft with only a synthetic Emax anchor either, and it doesn't have to, because **the naproxen validation already exists**: #1032's table has FeRx reproducing the published analysis to three significant figures via the `d/dt(clock) = 1` workaround. So the cheapest strong evidence is to re-run that same model compartment-free and put the two columns side by side in the PR description. No new files, one CLI run, and it converts the epic's existing number into a migration guarantee on the exact model users have today.

Then commit to the follow-up rather than leaving it as "maybe" — #1032 notes that #1019, #1020 and #1021 were all found by running these case studies by hand, and all three would have been caught by having one in CI. A shipped MBMA example is regression infrastructure, not a doc nicety.

### Three validation additions that cost almost nothing

- **A hand-computed OFV, Tier 1.** With `n_eta = 0` (which #989 made a supported fit), a compartment-free model's −2LL is a closed-form sum over rows. Pin it against a hand-computed value. That anchors the whole stack — readout, error model, weighting, objective — without needing NONMEM to be installed, and it's the one test that can't be wrong for the same reason the code is wrong.
- **A degenerate-weight oracle.** `weight = 1` must be bit-identical to no `weight =` on this path. Catches wiring errors between #1044 and the new readout for free.
- **Name what the twin does *not* prove.** The clock twin compares a readout reading a *state* (`clock`, carried through the ODE jet) against one reading the `TIME` builtin (`Op::PushTime`, ∂ = 0). The values must agree; the gradient paths are structurally different, so a green twin is not evidence about the new sensitivity kernel — `Dual2`-vs-FD parity is the only oracle for that half. Worth a sentence in the PR so a reviewer doesn't read the twin as covering both. (And the twin needs a tight tolerance — the ODE side integrates, so anything looser than ~1e-10 relative hides real disagreement.)

## The part I think matters more: what's left to be the best MBMA tool

Framing the gaps in three tiers, because they're won differently.

**Tier 1 — expressiveness. Nearly done.** `weight =` (#1029) merged, intermediates and continuations (#1030) merged, undefined-name rejection (#1028) merged, compartment-free here, weighted κ (#1031) open, `factor()` + sparsity (#1064) and profiling (#1065) filed. After #1031 every construction in the published literature I can find is writable without a workaround. This tier reaches parity, not leadership.

**Tier 2 — statistical structures nobody has first-class. This is where "best" is actually won.** Two items, and both are cheaper in ferx than they look:

1. **Within-arm residual correlation across timepoints — CS, AR(1), and continuous-time AR.** A study-arm's mean at successive timepoints comes from the same overlapping patients, so the residuals are correlated. Boucher & Bennetts Part II devotes a section to it and is explicit about the cost of ignoring it (over-precise estimates, and studies with many timepoints getting more weight than they should); Bill asked for AR(1) *and* a continuous-time AR, with weighting, and called out that unstructured fixed effects and correlated residuals are the two ways this gets solved in practice — which means #1064 and this item are alternatives to each other, not independent asks.

   Here's the part that makes this the highest-leverage item on the list: **the hard half already exists.** ferx has a dense correlated-`R` path — `compute_r_matrix_with_correlations`, its scaled twin, the `dr/df` and `d²r/df²` derivative builders, and `compute_iwres_with_correlations` — and it is *in the objective*, not just in diagnostics (`estimation/inner_optimizer.rs:2306`, `estimation/sens_outer_gradient.rs:479`). What's missing is only the *source* of the off-diagonal structure: today it comes from correlated `block_sigma` pairs across **distinct** sigma slots (parent/metabolite-shaped), plus a `(time, occasion)` replicate fallback. There is no way to say "these rows are the same arm at different times, correlate them with one ρ (or ρ^Δt) on the same sigma". That's a new grouping-and-structure source feeding machinery that already exists and is already differentiated — a much smaller lift than "implement correlated residuals", and it's the single biggest statistical differentiator left. It also has to compose with #1044: the correlation applies to the *weighted* residual, i.e. `R = W·C·W`, which is precisely Bill's "would also need to handle weighting".

2. **Nested variability levels with per-level weights** (Teun's question, and Bill's "arbitrarily deep random effects... weighting at all the levels... IOV as the degenerate case of weight = 1"). Where the field is: NONMEM has `$LEVEL` (since 7.3, reworked in 7.5) — super-ID data items with arbitrary nesting depth, site-within-country as the worked example, η shared *above* subject; FOCE requires `SLOW` and `MATRIX=R`, and there are real centering subtleties (`LEVCENTER=0`, `FNLETA=0`). Notably, **NONMEM has no per-level weighting** — the BTAV γ²/N-at-arm-level, unweighted-at-study-level asymmetry that every longitudinal MBMA needs is exactly what no mainstream tool declares. So this isn't catch-up, it's open ground, and #1031 is already the first level of it.

   Architecturally ferx has a cheaper route than NONMEM's: generalise the occasion mechanism to *N nested levels below ID*, each declared as `(grouping column, weight expression)`, with IIV as the degenerate top level — Bill's own framing. The user re-bases `ID` at the coarsest grouping (sponsor, or study), and per-subject likelihood independence survives, so the entire outer/inner FOCE structure is untouched; the cost is EBE dimension per subject, which is nothing at MBMA sizes. NONMEM's above-ID sharing breaks that independence and would be deep surgery for the same modelling power. Worth a design-explore issue on #1032, sequenced after #1031 — explicitly not part of #1063.

**Tier 3 — workflow. Where "best" is *felt*, and mostly ferx-r work.**

- **Contrast-scale everything** (ferx-r#297). An MBMA answers a treatment *difference*; every diagnostic on arm means shows the thing that isn't the answer. Contrast VPC, and SEs/intervals on contrasts rather than on arm parameters.
- **Data construction** (ferx-r#298). Arm-level literature tables are what people actually have; deriving and validating the weight column is where the silent errors enter — which is the same argument that made #1029 worth doing.
- **Heterogeneity and prediction intervals.** τ² and a prediction interval for a *future* study, built on ferxgui's existing uncertainty propagation. Boucher Part II explicitly advises against Q/I² for longitudinal MBMA, so skip those and go straight to the τ²-based interval.
- **Missing SD/SE handling** with an audit trail of which rows fell back — named as a recurring friction point in Part II, and cheap.
- **Small-study-count inference.** MBMA populations are 8–40 "subjects", where asymptotic SEs off an FD Hessian are optimistic. We already have `s` / `rsr` and SIR; what's missing is a documented recommendation for this analysis class rather than a knob.
- **Simulation semantics.** Simulating a future trial needs an arm size to scale both the residual and the BTAV — if `weight =` reads a data column, what does `simulate()` do when the user is inventing arms that aren't in the data? That needs an explicit answer, and the prediction-interval item depends on it.
- **Binary / categorical arm-level endpoints** (#760, PR #900). The canakinumab ACR20 case is half of the two published case studies — worth tracking as an MBMA dependency rather than a separate endpoint feature.
- **Model-based network meta-analysis.** Deferred in #1032 and rightly so, but it's the thing that would make ferx *the* tool rather than *a* tool — and it becomes tractable off the back of #1064, since treatment is a factor.

## One point on #1064's rank-deficiency question

You asked Bill whether to drop a reference level, use sum-to-zero, or require no intercept. My view: make it explicit in the syntax as you proposed, and default to **sum-to-zero when the factor is declared as a nuisance/profiled block** (#1065) — nobody reads those levels, the grand mean stays interpretable, and the nuisance block stays symmetric and better conditioned. Reference-level is what a regression user expects when they *do* want to read the coefficients, so it should be available, just not the default for the unstructured-placebo case.

More important, and I think missed so far: **the collinearity isn't only with a fixed intercept.** In a real MBMA the study-level effect is a *random* effect, so `factor(STUDY, TIME)` plus an η on the same additive scale is over-parameterised too — that study's η is the mean of its own timepoint levels. Whatever rank check `factor()` grows has to consider the random-effects structure, not just fixed intercepts, or it'll pass a model that's unidentified in exactly the configuration this feature exists to serve.

---

*Nothing above blocks #1063. Q1 and Q3 are "ship, with a test"; Q2 is "keep it and add one constant".*

roninsightrx:
@TeunP — worked through all of it. Two of your asks were already in the branch, one of them found a **real pre-existing bug**, and the "measure it rather than assert it" one came back with the opposite of what we both assumed. Changes in `e1cf629b`.

## Q1(2) — the per-occasion readout: you were right, and it is worse than a divergence

You asked whether a Form C `y = ... KAPPA_X` on an analytical PK model evaluates BSV-only, so the same expression written compartment-free gives different numbers. It does — and the compartment path is not merely *different*, it is **wrong**. Filed as #1079.

`pk::apply_analytic_readout` is called from `predict_iov` with `eta_bsv`, so any individual parameter the readout reads is evaluated at κ = 0 while the concentration it is handed carries the occasion's κ. Against the hand-written ODE twin, on `y = central / V + BASE` with `BASE = TVBASE * exp(KAPPA_B)`:

| t | analytical Form C | ODE twin | rel |
|---|---|---|---|
| 1  | 6.236181504 | 7.533624045 | 1.7e-1 |
| 25 | 6.315446946 | 5.656087038 | 1.2e-1 |

12–20% on the prediction. Hand-checked: both engines agree on the *concentration* to solver tolerance, and the entire difference is `BASE(κ=0.5) = 3.297` vs `BASE(κ=0) = 2`. The ODE twin is right — an individual parameter carrying a κ is per-occasion by definition, and `$ERROR` is evaluated per record with that record's occasion.

Two things worth recording about why it survived #650/#655:

- **The obvious test cannot show it.** `y = central / V` with κ on `V` agrees with its twin to 1e-13, because the predictor reconstructs the amount as `conc × V(κ=0)` and the readout divides by the same `V(κ=0)` — they cancel exactly. You need a readout where the κ-carrying parameter does *not* cancel: an additive baseline, a second analyte, a bounded transform. (That cancellation is load-bearing today, and is the thing to be careful of when fixing this.)
- **FD would not have caught it either.** `eval_readout_at_obs` seeds the readout parameters BSV-only precisely to *match* the f64 path, so the gradient is the exact derivative of the wrong function and `Dual2`-vs-FD parity passes. Only a cross-engine oracle finds it.

Not a regression from #1063 — the compartment-free path is the correct one, and the contrast is what made it visible. Per Ron's call it is a separate PR: the fix changes numbers for existing compartment models and needs the sensitivity half moved in lockstep.

## Q1 — IOV: already analytic, and the κ-carrying twin is in

Both of your first two conditions landed before your review. Since `7d422fab` the compartment-free IOV path is **analytic**, not FD: the jet is the ordinary readout jet seeded on the stacked `[η_bsv, κ_0…κ_{K−1}]` axes — no walk, no carryover, because there is no concentration. Routing is tested in both directions, and `build_info` is now asserted directly (`e1cf629b`) for all three cases: BSV analytic, IOV analytic, past-cap FD.

`algebraic_iov_matches_the_dummy_ode_twin` is the κ-carrying twin you asked for. It compares `predict_iov` output at a fixed parameter point rather than two fits — an earlier fit-based version disagreed, and that was the test's fault, not the code's: the two routes take different trajectories from the same start. It also asserts κ **moves** the prediction, so it cannot pass by both engines dropping it.

Your `weight = SE` / σ-FIXed caution is now a test rather than an argument: `Dual2`-vs-FD parity on the weighted MBMA objective, with per-row SEs spanning an order of magnitude (a constant SE reduces to a rescaling and hides a per-row error). Every other parity test used proportional error.

## Q2 — axis cap: the measurement inverts the premise

The dedicated constant was already there. The measurement was not, and it is the interesting part. Per gradient evaluation, compartment-free, release build:

| axes | analytic (`Dual2`) | FD of the predictor | ratio |
|---|---|---|---|
| 5  | 5.6 µs   | 4.7 µs  | 1.19× |
| 14 | 28.1 µs  | 10.8 µs | 2.61× |
| 24 | 141.8 µs | 26.6 µs | **5.33×** |

Analytic grows 25× across a 4.8× axis increase (quadratic — `Dual2<M>` is O(M²)); FD grows 5.6× (linear). **The analytic jet is slower than FD at every axis count measured, and 5× slower at the cap.**

The mechanism is that this is the opposite of the ODE case: there, one integration dominates any number of dual lanes, so computing all `M` derivatives in one pass is an enormous win. Here the differentiated thing is a single scalar expression per row — nothing to amortise against, so the O(M²) lane arithmetic is pure overhead.

So: keep 24, for a stronger reason than compile time. Raising it to 96 would make these models *slower*. The cap is not a cliff to push back — past it, FD is the better route anyway. Recorded next to the constant, since "raise the cap" is exactly the wrong instinct to have later, and the crossover question ("should these models prefer FD *below* the cap too?") is filed as #1084 rather than decided here — it is a change of default, and the honest FD comparison needs objective evaluations, not predictor evaluations.

Your other cap request is in: past the cap the model now says so at parse, naming both its axis count and the cap.

## Q3 — naproxen: already shipped, and it is the anchored model

Shipped rather than deferred, and I think it earns its place: the example *is* the model the NONMEM anchor runs, with the two dataset copies pinned byte-equal, so what a user runs from `examples/` is what was validated.

| | ferx | Monolix (published) |
|---|---|---|
| naproxen on Emax | 0.7930 (SE 0.0646) | 0.792 (SE 0.064) |
| ET50 placebo / naproxen | 0.6901 / 0.2132 | 0.698 / 0.216 |
| ω(E0) / ω(Emax), SD | 0.6215 / 0.7387 | 0.627 / 0.763 |

Plus the dummy-clock workaround fitted to convergence alongside and required to agree — the migration guarantee on the exact model users have today. Your regression-infrastructure argument is why I would rather not defer it: #1019/#1020/#1021 were all found by running these by hand.

Your three validation additions are all in:

- **Hand-computed OFV at `n_eta = 0`.** Anchors readout + error model + objective against arithmetic. It also caught something worth pinning: the objective drops the `ln(2π)` constant (NONMEM's convention). An OFV that gained or lost `n·ln(2π)` would converge to identical estimates and look entirely plausible, while every LRT against a NONMEM run would be off by an observation-count-dependent term.
- **Degenerate-weight oracle.** `weight = 1` bit-identical to no weight.
- **What the twin does not prove.** Now stated in the file: the two sides reach the same value by structurally different routes — a state carried through the ODE jet vs `Op::PushTime`, whose derivative is identically zero — so a green twin says nothing about the sensitivity kernel. Tolerance tightened 1e-6 → **1e-10** as you asked; it holds comfortably.

## Tier 2 — filed

- **#1081** — within-arm residual correlation (CS / AR(1) / continuous-time AR), with your "the hard half already exists" framing front and centre: the dense correlated-`R` path is in the objective, and what is missing is only the *source* of the off-diagonal structure. Includes the `R = W·C·W` composition with the weighting, the PD bound on CS, and Bill's point that this and #1064 are alternatives to each other rather than independent asks.
- **#1082** — nested variability levels with a weight per level, with the `$LEVEL` survey and your architectural argument (re-base `ID` at the coarsest grouping, keep per-subject independence, pay only in EBE dimension). Sequenced after #1031.
- **#1083** — `simulate()` semantics for `weight = COL` when the arm is invented, noting the prediction-interval dependency.

## On #1064's rank deficiency

Your point that the collinearity is not only with a *fixed* intercept is the one I had missed, and it is the case that matters: with a random study effect, `factor(STUDY, TIME)` plus an η on the same additive scale is over-parameterised, since that study's η is the mean of its own timepoint levels. Added to #1064 — whatever rank check `factor()` grows has to see the random-effects structure, or it will pass exactly the configuration the feature exists to serve. Sum-to-zero as the default for a profiled nuisance block, reference-level available, is what I will spec.

---

Verified after all of this: 3319 lib tests default / all-features clean, 0 failures; all slow-gated tests pass; `cargo check --all-features --tests`; clippy clean for the diff.

Issue #978, opened 2026-08-17
epic(ode): a-priori stiffness probe + finite-guarded auto-switch (ode_method = auto)


roninsightrx:
## Summary

Add `ode_method = auto`: a two-layer stiffness handler that picks the ODE solver automatically instead of forcing the user to know whether a model is stiff.

1. **A-priori probe (cheap, validated)** — one init-state Jacobian eigenvalue check picks the *starting* solver before any integration.
2. **Runtime detector + finite-guard** — catches emergent (post-dose) stiffness and parameter drift the start can't predict, and never escalates into a *worse* solver.

This is ferx's answer to "why don't we have an LSODA / NONMEM ADVAN13" — see the design note at the bottom for why a composite auto-switch beats reimplementing LSODA.

Related: #959 (max-steps freeze-pad → silently wrong on stiff models), #708 (Dual-sens hangs, no bail on min-step clamps), #520 (loose tolerances → bad covariance on ODE models), #518 (warn on `ode_*` keys for non-ODE models). Prior art: #387, #445 (added the Rosenbrock stiff solvers).

## Motivation — measured across the ferx-testdata library

Ran all 15 ODE models × all 5 solvers (rk45, vern7 explicit · rodas4, rodas5p, rosenbrock23 implicit), fixed 3-iteration budget, 300 s cap:

- **Stiffness is model-specific, and it defeats the whole explicit family.** Both explicit solvers stall on the 3 cyclophosphamide models — `cyclo_final` completes *one* ODE solve in 300 s under rk45/vern7; all 3 implicit solvers finish in 0–10 s (~18× on `cyclo_initial`).
- **Every implicit solver diverges — on a different model.** `rodas4` 💥 `tmdd_crib` (OFV 9.0e7), `rodas5p` 💥 `tmdd_cr` (8.7e5), `rosenbrock23` 💥 `tmdd_crib` (1.3e9), where explicit stays clean (~5000). Failures are **disjoint** → a backup stiff method recovers. **No implicit solver is a safe unconditional default.**
- **Accuracy gap**: even where explicit is faster, implicit often reaches a lower OFV at equal budget (`tmdd_cr` ΔOFV −125) — ties into #520.

## Part A — a-priori probe (validated, ~5 µs)

You cannot decide from the dataset (stiffness is a Jacobian property, not a schedule). You *can* evaluate `J = ∂f/∂u` once at the initial state and read its eigenvalues. Tested on all 15 models:

[...]

Issue #959, opened 2026-08-02
bug(ode): exhausting ode_max_steps freeze-pads the remaining save points and returns normally — silently wrong predictions on stiff models


TeunP:
When an integration segment exhausts `ode_max_steps`, the solver **freeze-pads every remaining save point with the last state** and returns normally. No error, no warning, no flag on the result. The predictions are finite, plausible-looking, and wrong.

On a stiff full-TMDD model at the **default** `ode_max_steps = 10000` this produces a 12 % error in the concentration profile — and identical predicted concentrations at different times, which is self-evidently impossible for a decaying curve.

cc @hiddevandebeek @roninsightrx

Versions: `ferx` 0.2.0 (core 0.2.0), R 4.5.1, macOS.

## The symptom, from ferx alone

Same model and data, only `ode_max_steps` changed:

| `ode_max_steps` | wall | PRED at t=0.25 | PRED at t=0.50 |
|---|---|---|---|
| **10000 (default)** | 0.28 s | **112.516634** | **112.516634** |
| 100000 | 2.36 s | 110.579958 | 99.111694 |
| 2000000 | 40.30 s | 110.579958 | 99.111694 |

At the default the two times return **the same number to seven figures**: the trajectory stopped advancing and the remaining points were padded. `ferx_predict()` returned a normal data.frame in every case; nothing in the output distinguishes row 1 from rows 2-3.

Cross-checked against `rxode2` (liblsoda, which switches to BDF when the system goes stiff) at `atol = rtol = 1e-12`:

```
rxode2:  t0.25 = 110.580000   t0.50 = 99.111690
```

so rows 2-3 are right to ~4e-7 and row 1 is wrong by 12 %.

## It degrades silently with stiffness

Sweeping `KON` with `KOFF = KON * KD` at fixed `KD = 20` — this leaves the binding **equilibrium** untouched and changes only the timescale separation, so the true solution barely moves. All at default `ode_max_steps`, compared against rxode2 at 1e-12:

| KON | KOFF | max rel. error | median rel. error | ferx wall |
|---|---|---|---|---|
| 0.1 | 2 | 2.6e-04 | 5.8e-12 | 0.050 s |
| 1 | 20 | 1.0e-03 | 3.7e-12 | 0.094 s |
| 10 | 200 | 2.7e+05 | 1.9e-11 | 0.156 s |
| 100 | 2000 | 3.1e+08 | **2.1e-01** | 0.138 s |
[...]

PR #1066, opened 2026-08-24
feat(ode)!: a-priori stiffness probe, `ode_method = auto`, now the default [#978]

roninsightrx:
## Why

Closes the first part of #978. Stiffness defeats the explicit solver family on some PK/PD
models — the cyclophosphamide models in `ferx-testdata` are the standing example — and today the
only remedy is for the user to already know their model is stiff and name `rodas4` themselves. Get
it wrong in the other direction and a Rosenbrock method is pure overhead, or worse: every implicit
stepper in this library has been measured diverging on *some* model where the explicit family
stayed clean. So "just default to a stiff method" is not available either.

`ode_method = auto` decides from the model instead of from the user's prior knowledge.

## What changed

Before each integration, build the Jacobian `J = ∂f/∂u` of the `[odes]` right-hand side at the
state that integration is about to start from and read its fastest mode, `max |Re λ(J)|`. Above
`30` per time unit the segment starts on a stiff Rosenbrock method (`rodas4`, or `rodas5p` at
`ode_reltol ≤ 1e-8`); otherwise it keeps `rk45`. Naming any other `ode_method` still forces that
method exactly — `auto` is the only value that decides anything.

**`auto` is the default** (per the decision on the open questions below). A model that names no
stepper is probed; `ode_method = rk45` restores the previous behaviour exactly. `OdeMethod::default()`
was being read inside the guard as "the explicit stepper to retreat to", which silently inverts
once the enum default is `Auto` — the fallback is now the separately-named
`OdeMethod::EXPLICIT_FALLBACK`.

The resolution lives at the three `make_stepper` call sites (`solve_ode_dense`,
`solve_ode_g_dense`, `solve_ode_until_threshold`) rather than being threaded through the
prediction and sensitivity layers. Two properties fall out of that placement rather than being
engineered on top of it:

- **The probe sees each segment's own initial state.** For a binding model the fast eigenvalue is
  carried by `KON · central` and is identically zero before any drug is present, so the declared
  initial condition is the one state at which such a model looks *least* stiff. Because the
  predictor already splits a subject at every dose, the probe sees the post-dose Jacobian on every
  segment after the first, with no extra plumbing — and it re-decides as the optimizer moves θ.
- **`f64` and `Dual2` resolve identically.** Both paths probe their own integration and agree
  because the probe reads `.val()` only, exactly like every other comparison a driver makes. The
  analytic gradient therefore differentiates the trajectory the predictor reports.

**The guard.** An escalation that returns a non-finite trajectory, *or that clamps at `min_dt`*, is
discarded and the segment re-solved explicitly. The second trigger is the one that matters and it
is not what the issue specified: measured on a system that blows up in finite time, the escalated
`rodas4` solve came back as a **finite** `4.8e282` with 10 min-step clamps — the freeze-pad of #959
— so a non-finite test alone would have passed the silently-wrong answer straight through. Scoped
to `auto`; a named method is honoured as asked, failed result included.

Two counters on `OdeSolverStats` record segments escalated and escalations rejected.

## Alternatives considered

- **Implement LSODA / ADVAN13.** Rejected in the issue for reasons that still hold: BDF is
  Newton-based and multistep, so it breaks the `Dual2`-through-the-same-stepper architecture, and
  PK dosing forces an order-1 restart at every dose.
- **Probe once per subject at the declared initial condition**, as the issue proposed. This is
  what the measurement below overturned — see *Numerical validation*.
- **A dimensionless discriminator** (`λ_max · span`) instead of an absolute rate. Scale-free, but
  it makes a long segment on a benign model read stiff, and it discards the calibration the issue
  had already validated. Kept the absolute threshold and documented the time-unit caveat instead.

## Cross-repo dependency

No `pub` API removed or changed; `OdeMethod::Auto` and two `OdeSolverStats` fields are additive.

## Breaking changes
- [x] None *to the format or the API* — but see the behavioural note below.

No `pub` API is removed or changed: `OdeMethod::Auto`, `OdeMethod::EXPLICIT_FALLBACK` and two
`OdeSolverStats` fields are additive, and every `.ferx` file parses exactly as before.

**Behavioural, and deliberate:** a *stiff* ODE model that names no `ode_method` now integrates on
a different (and better-conditioned) stepper, so its predictions and OFV change — on the
cyclophosphamide model, from a fit that spends 827 s without finishing an optimizer iteration to
one that converges in 0.7 s at 0.013 OFV units from NONMEM. Non-stiff models keep the explicit
stepper and their existing numbers. `ode_method = rk45` pins the old behaviour.

## Numerical validation

**The measurement that changed the design.** The issue's Part A table reads
`|Re λ|max` at each model's *declared initial condition* and reports the TMDD parameterizations as
uniformly non-stiff (`≤ 12.4`) — from which it concludes the probe "avoids the divergence trap"
because those models would never escalate. @TeunP's review questioned exactly this and asked for
the number at both evaluation points. Measured per segment, where the probe actually runs:

| model family | `λ_max` across segments | verdict at threshold 30 |
|---|---|---|
| cyclophosphamide | 20 – 234 | stiff on every segment |
| TMDD SAD (`cr`, `crib`, `ib`) | 3.7 – 126 | **straddles the threshold within one model** (240 of 580 segments escalate) |
| TMDD SAD (`qss`, `mm`, `linear`) | 1.8 – 2.1 | non-stiff |
| busulfan, pembrolizumab | 0.1 – 4.6 | non-stiff |

So the safety property the issue relied on does not survive per-segment evaluation: TMDD *does*
escalate, on its high-dose subjects. That is the correct call — those segments really are stiff —
but it means the guard carries correctness rather than being a backstop, which is what @TeunP
predicted. It is also why the guard's stall trigger was added.

**Prediction equivalence** (one pass, default tolerances, all methods):

```
cyclo_final     rk45 sum=2.4531595078e6   rodas4 sum=2.4531653506e6   auto sum=2.4531653506e6
tmdd_sad_cr     rk45 sum=1.2002684700e5   rodas4 sum=1.2002672439e5   auto sum=1.2002675404e5
tmdd_sad_crib   rk45 sum=1.2001372581e5   rodas4 sum=1.2001360484e5   auto sum=1.2001363208e5
```

`auto` reproduces `rodas4` exactly on cyclophosphamide (all segments escalate) and lands between
the two on TMDD (some segments escalate), agreeing to 7 digits throughout. No non-finite
predictions under any method.

**NONMEM anchor — cyclophosphamide `run2_final`, FOCEI, 55 subjects / 226 observations:**

```
NONMEM FOCEI (nm/cyc_focei.lst)   OFV 3241.7206   theta 4.29  120    25.3  0.683 0.931 0.406 0.866
ferx  rodas4 (converged, 0.4 s)   OFV 3241.7050   theta 4.291 119.52 25.34 1.011 0.931 0.406 0.865
ferx  auto   (converged, 0.7 s)   OFV 3241.7078   theta 4.291 119.55 25.34 1.037 0.932 0.406 0.865
ferx  rk45                        OFV 3243.8278   theta 4.287 119.89 25.37 1.005 0.931 0.407 0.871   (827 s, ONE capped iteration, not converged)
```

ΔOFV vs NONMEM is 0.013 for `auto`. θ4 differs from NONMEM (1.04 vs 0.68) under *both* ferx
methods, so it is a pre-existing property of this fit rather than anything `auto` introduces.

**Fit-level speed** (FOCEI, fixed 4-outer-iteration budget except where noted):

| model | `rk45` | `auto` |
|---|---|---|
| cyclophosphamide `run2_final`, to convergence | 827 s for one capped iteration (not converged) | **0.7 s**, OFV 3241.708 |
| TMDD SAD `cr` | 372.4 s, OFV 5018.5 | **2.3 s**, OFV 4888.4 |
| TMDD SAD `crib` | 33.1 s, OFV 4885.0 | **2.2 s**, OFV 4893.0 |

Read the wall-clock column first. At a *fixed iteration budget* the two solvers sit at different
points on the optimizer's path, so the OFV column there mixes accuracy with how far each got —
`crib` lands slightly higher and `cr` 130 units lower, and neither is an accuracy claim. The
converged cyclophosphamide row is the like-for-like comparison.

## Performance
- [x] Faster — `tmdd_sad_cr` FOCEI at a 4-iteration budget: **372.4 s → 2.3 s**; cyclophosphamide
      `run2_final` to convergence: 827 s for one capped iteration (not converged) → **0.7 s**.

Cost when nothing escalates: `n + 1` right-hand-side evaluations plus one `O(n³)` eigensolve per
segment — about one solver step, against segments that take tens to thousands. Zero-length segments
skip the probe entirely. `auto` is opt-in, so no existing fit changes.

## Tests
- [x] Unit tests added in the same module (`cargo test --lib` passes: 3293 tests, 0 failures)

Tier-1 (`src/ode/stiffness_tests.rs`, 15 tests): the probe against the exact spectrum of a
diagonal system; stiff/non-stiff separation; that it reads the fastest *mode* and not state
magnitude; that it declines to classify a non-finite or empty system; the pre-dose-vs-post-dose
evaluation-point property; that a named method is never second-guessed; the tolerance-dependent
stiff choice; **that `f64` and `Dual1` resolve identically at four states** (the property the
gradient depends on); the guard re-solving a stalled escalation; and that a named method never
moves the new counters.

`Dual2`-vs-FD parity under the switched method (CLAUDE.md's sensitivity rule):
`sens::ode_provider::tests::ode_provider_matches_fd_under_the_auto_stiff_switch`, with a
non-vacuity assertion that the fixture really escalates.

Tier-2 (`tests/ode_method_auto.rs`): the fit option reaching the solver, `auto` reproducing
`rodas4`'s predictions with a non-vacuity check that it did not silently stay on `rk45`, the
estimation path returning a finite OFV, and the parse diagnostic listing `auto`.

Tier-3 (gated): `auto` and `rodas5p` converging to the same optimum at `ode_reltol = 1e-10`.

## Example

```toml
[fit_options]
  ode_method = auto
```

On the cyclophosphamide model above that is the whole diff, and it takes the fit from
827 s for one capped iteration (not converged) to 0.7 s at OFV 3241.708 (NONMEM: 3241.721).

## Docs
- [x] `docs/` updated — `model-file/ode-models.qmd` (new "Letting ferx pick the stepper" section
      plus the method table), `model-file/fit-options.qmd`, `maturity.qmd`

## Checklist
- [x] `cargo clippy` clean
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the probe is generic over
      `PkNum` and reads `.val()` only, so it adds no branch the derivative rides

## Reviewer hints

1. **`integrate_resolved_g`** (`src/ode/solver.rs`) — the whole feature's control flow, including
   the guard and how counters are merged across the two solves.
2. **The guard's stall trigger.** It is a deliberate widening of what the issue specified, on the
   evidence in *Numerical validation*. The trade-off: a stiff method that clamps even once now
   costs a second, explicit solve. Measured on cyclophosphamide and TMDD there are no spurious
   fallbacks (`auto` matches `rodas4` exactly on cyclo and stays fast), but this is the knob most
   likely to need tuning against a model I have not run.
3. **The threshold is dimensional.** `30` is a rate in the model's own time unit, calibrated for
   the hour-based PK convention. Documented, but it is a real limitation of the design and not
   something the code can detect.

## Open questions — resolved

1. **Scope.** Parts B and C are filed as #1080, which also carries @TeunP's point 1 (the
   `min_step_clamped_steps` signal is blind to the fast-binding failure mode, so anything built
   on it inherits the blind spot) and the two premises from #978 that did not survive
   measurement. This PR stays Part A.
2. **Default.** `auto` is the default as of `5dce418b`. Test fallout and what it revealed is in
   [this comment](https://github.com/FeRx-NLME/ferx-core/pull/1066#issuecomment-5413377436).
3. **Maturity.** The Rosenbrock steppers are promoted experimental → beta, since the default path
   is not a place to ship components marked experimental. This is the one judgment call in the
   PR that is a claim rather than a code change — flagging it for an explicit yes/no.


roninsightrx:
## Open questions — resolved

Decisions from @roninsightrx on the three open questions in the description:

1. **Scope.** Parts B and C get their own issue rather than riding this PR. Filing it; this PR
   stays Part A.
2. **Default.** `auto` becomes the default **now**, not after a soak as opt-in. Implementing in
   this PR — see below for what that touches.
3. **@TeunP's point 1** (the `min_step_clamped_steps` signal is blind to the fast-binding failure
   mode, where the explicit solver returns quickly with a 40 %-wrong answer) goes into the Part
   B/C issue, since anything built there on that signal inherits the blind spot.

### What making `auto` the default changes

Beyond flipping `#[default]`, three things need care and are worth a reviewer's attention:

- **`OdeMethod::default()` is currently used inside the guard to mean "the explicit fallback".**
  With `Auto` as the enum default that reading silently inverts, so the fallback becomes a named
  constant instead of `default()`.
- **Every ODE model now pays the probe**, and stiff ones now escalate without being asked. That is
  the intent, but it makes the escalation path load-bearing for models nobody opted in for — the
  guard is what stands behind that.
- **The Rosenbrock steppers move onto the default path** while `docs/maturity.qmd` lists them as
  *experimental*. Leaving that as-is would mean shipping experimental components by default, so
  they are promoted to *beta* on the strength of this PR's validation (NONMEM anchor +
  equivalence tests). Flagging it explicitly because it is a maturity claim, not a code change —
  say the word if you would rather they stay experimental and `auto` be the only beta row.

Numbers in the description were measured with `auto` opt-in; they are unchanged by the default
flip, since a model that names a method still gets it. I will re-run the suite and report any
result that moves because a model now escalates by default.

roninsightrx:
## Default flipped — `auto` is now the default (`5dce418b`)

All three decisions actioned. Parts B and C are now #1080, which also carries @TeunP's point 1.

### The one thing that needed care

`OdeMethod::default()` was being read inside the guard as "the explicit stepper to retreat to".
Those were the same value until this commit and are not the same *concept* — left alone, the
guard would have fallen back to `auto`, i.e. to the escalation it exists to undo. The fallback is
now `OdeMethod::EXPLICIT_FALLBACK`, documented as deliberately distinct from `default()`.

### Test fallout, and what it says

Six tests asserted the old default. Every one turned out to be *about the explicit driver* rather
than about the default, so each now names `rk45` with the reason recorded — "add the method the
test used to get for free" is otherwise indistinguishable from papering over a regression. The
interesting one:

- `solve_ode_with_stats_counts_min_step_clamped_accepts` uses `du = -100·u` with `min_dt = 1.0`.
  Under the new default the probe escalates (λ = 100) *and* the clamp then trips the guard into a
  second explicit solve — doubling every counter the test asserts on. That is the guard working,
  observed by accident.
- `tests/stiff_ode_method.rs` had a test pinning the premise "RK45 on the default budget is
  materially wrong on this system". It now needs to name `rk45`, so I added the companion:
  `the_default_rescues_the_system_that_defeats_the_explicit_method` — same model, same
  tolerances, same step budget, no `ode_method` at all, and it lands on the brute-force reference
  where naming `rk45` still does not. The two bracket the change.

Full suite green: **3293 lib tests, 142 integration test binaries, 0 failures**, no numerical
anchor moved. `cargo clippy` clean.

### Maturity

I promoted the Rosenbrock steppers experimental → beta, as flagged in the previous comment: the
default path is not a place to ship components marked experimental. The promotion rests on this
PR's validation (NONMEM FOCEI anchor, 3241.708 vs 3241.721; prediction equivalence across all
five methods). Say the word if you would rather they stay experimental — that would argue for
reverting the default flip too, since the two go together.

### What an existing user feels

- A **stiff** model's predictions change, because a better-conditioned integrator now produces
  them. This is the intended effect and the reason for the flip.
- A **non-stiff** model keeps the explicit stepper and its existing numbers, paying one Jacobian
  and one eigensolve per segment for the check.
- A model that **names** a method is completely unaffected.
- `ode_method = rk45` restores the old behaviour exactly.

Documented under `### Changed` in the changelog rather than only under `### Added`, since the
default flip is the part an existing user notices.

roninsightrx:
Rosenbrock → beta confirmed, thanks — already in `5dce418b`, no change needed.

## `codecov/patch` failure fixed (`2f517bfa`)

The gate caught something real rather than a threshold quibble: **`solve_ode_until_threshold`'s entire `auto` branch had no test**, retry included. That is the driver a joint PK-TTE fit locates event times on, and the branch that stops a diverged stiff solve from being laundered into a censored subject — the last place to leave untested. My Tier-1/2 tests all went through the dense drivers and never touched it.

Three tests added, each pinning a contract:

- `until_threshold_escalates_a_stiff_system_and_finds_the_crossing` — stiff hazard system where RK45 exhausts its step budget *before* the crossing and returns `Failed`, while `auto` finds it. Threshold is 4 rather than 2 on purpose: at 2, RK45 reaches the crossing on its last few steps and the contrast disappears (found by writing it at 2 first and watching it pass for the wrong reason).
- `until_threshold_retries_a_failed_escalation_on_the_explicit_method` — counts RHS calls to prove the retry ran. Both methods fail on that system, which is the point: what is under test is that the guard re-runs, not that the re-run rescues it.
- `make_stepper_falls_back_to_the_explicit_stepper_for_auto` — pins that the defensive arm really is RK45 (same error norm, proposed state, controller exponent).

Plus `the_rosenbrock_stepper_refuses_auto_loudly` for the spelled-out non-Rosenbrock arms `Auto` joined.

**Uncovered added lines: 18 → 2.** What remains is one `unreachable!` arm inside `attempt()`, reachable only from a hand-built stepper the constructor refuses to produce, and a panic branch inside a test's own match. Neither is worth contorting a test to reach.

Suite green: 3297 lib tests, 142 integration binaries, clippy and fmt clean.

roninsightrx:
## Review response

Eleven findings. **Seven fixed**, one downgraded on measurement, three deferred to #1080 with reasons. Taking them in order of what actually changed.

---

### Confirmed reachable bug — the event-time guard missed stalls (`solver.rs:871`)

**Correct, and the most valuable finding here.** `solve_ode_until_threshold` retried only on
`ThresholdCrossing::Failed`, but that driver force-accepts at `min_dt` and keeps going — so a
stalled stiff solve returns an ordinary-looking `Crossed`/`ReachedEnd` that no inspection of the
outcome can distinguish from a healthy one. On the TTE path that is a wrong simulated event time,
or a censoring laundered out of a diverged solve. My own comment above the retry called it "not
optional" while the code only handled the loud half.

I did not want to take this on argument, so the regression test asserts the precondition first:

```rust
assert!(matches!(named, ThresholdCrossing::Crossed(_)),
    "precondition: the stalled stiff solve must still *report success*");
```

That assertion **passes** — `rodas4` on the fixture really does return `Crossed` off a trajectory
it never integrated. Reachable, not theoretical.

Fix: `until_threshold_with` now returns `(outcome, min_step_clamped)`, and the guard retries on
`Failed` **or** a clamp, re-solving from the entry state. Test:
`until_threshold_retries_an_escalation_that_stalled_but_still_reported_success`, which counts
right-hand-side calls to make the re-solve observable (the monitor is `d/dt(chz) = 1`, exactly
linear, so it sails through the stall — that is the whole point of the fixture).

### Unbounded eigensolve (`stiffness.rs:158`) — correct

`DMatrix::complex_eigenvalues` calls `do_decompose(..., max_niter = 0, ...).unwrap()`, and
`max_niter = 0` *disables* the cap rather than setting it to zero. So the documented
"non-convergence → `None`" path did not exist; a Jacobian it could not reduce would spin in a
worker thread. Now `Schur::try_new` with an explicit 1000-iteration cap, so the documented
behaviour is the real behaviour. A probe is a hint, and a hint is never worth hanging for.

### `rodas5p` band mismatch (`stiffness.rs:108`) — correct, fixed

`auto` cut at `1e-8` while `OdeMethod`'s docs send a hand-choosing user to `rodas5p` at `1e-9`.
Moved to `1e-9`; the unit test now pins all three boundary points and says *why* (it is a
consistency requirement between the probe and the prose, not a free choice).

### Doc fixes — all correct

- **`EXPLICIT_FALLBACK` rustdoc contradicted itself.** My own bug, and an instructive one: the
  blanket `s.replace("OdeMethod::default()", "OdeMethod::EXPLICIT_FALLBACK")` I used to retarget
  the call sites also rewrote the sentence explaining the distinction, inverting it. Rewritten.
- **`FitOptions::ode_method` still documented `Rk45` as the default.** Fixed.
- **`fit-options.qmd` said "`rk45` (default)" in the body** while the Default column said `auto`,
  and still carried the stale "6× faster" figure. Both fixed.
- **`auto_stiff_segments` over-promised.** It is a floor, not a total —
  `solve_ode_until_threshold` takes no stats, so TTE-path escalations are not counted. Said so.

### Fast-but-not-stiff escalates (`stiffness.rs:130`) — correct, and my claim was wrong

The discriminator reads how *fast* the fastest mode is, not how *separated* the modes are.
Textbook stiffness is a ratio; this is not. A transit chain with `ktr = 50` has every eigenvalue
at −50, is not stiff, and escalates. A minute-clock model reads 60× larger and escalates
wholesale. `|Re λ|` rather than `Re λ` widens it further — a *growing* mode counts, which is why
the PR's own guard test escalates a blow-up.

The CHANGELOG claim "nothing changes where nothing is stiff" was therefore wrong, and I have
corrected it rather than leaving it standing. Both limitations are now documented in the module
docs and in `ode-models.qmd`.

I did **not** change the discriminator, and the reason is that the obvious alternatives are worse
in ways the data supports: a ratio `λ_max/λ_min` is undefined the moment any mode sits at zero,
which is every compartment before its first dose; normalizing by segment length makes a long
segment on a benign model read stiff. The escalations this causes are *slower, not wrong* — the
guard bounds the damage, and `ode_method = rk45` opts out. Tracked in #1080.

---

### Downgraded on measurement — the HIGH finding (`stiffness.rs:183`)

The mechanism is real and correctly reasoned: `resolve_method` re-decides every evaluation, so a
segment near the threshold could flip across an FD θ-perturbation, and the PI-controller note at
`solver.rs:604` is the right precedent. I checked it against the per-segment `λ_max` distribution
for the model cited as worst case (TMDD, 240/580 escalating):

| window around the threshold | segments |
|---|---|
| ±1e-6 | 0 |
| ±0.01 | 0 |
| ±1.0 | 0 |
| ±3.0 | 40 |

Nearest below 30: **19.10**. Nearest above: **31.46**. Nothing within 1.46 units — about 5 % —
of the boundary, and the values cluster by dose level rather than spreading densely. An FD step
of 1e-6 relative moves `λ_max` by ~3e-5 absolute, five orders of magnitude short of flipping
anything here.

Two further points against acting now:

1. **The trajectory is already a discontinuous function of θ**, through the step controller's
   accept/reject decisions — and that discontinuity is *larger* (tolerance-level, default
   `reltol = 1e-4`) and far more frequent than a method flip (`rk45` vs `rodas4` agree to ~1e-6
   relative on the measured models). The method switch is not a new class of noise; it is a
   smaller instance of noise the FD paths already ride.
2. **Hysteresis, the suggested mitigation, would be actively wrong here.** It requires
   remembering the previous decision, which makes the choice depend on evaluation *order* — and
   the `f64` prediction and the `Dual2` sensitivity solve do not share an order. That trades a
   rare threshold flip for a systematic disagreement between the gradient and the trajectory it
   is supposed to differentiate, which is the one invariant this design rests on.

So: not a blocker, no code change, and the constraint is now written down. Tracked in #1080 with
the measurement, since a model that *does* park a segment on the threshold would see it.

### Deferred to #1080

- **`finite_g` is val-only, so a finite `Dual2` with NaN jets passes.** Plausible and worth
  closing, but I have no repro, and the fix has a real cost: rejecting on the `Dual2`
  instantiation but not on `f64` makes the two paths choose different steppers, which is the
  invariant above. Wants a reproducing case first, not a speculative guard.
- **The explicit re-solve's result is returned unchecked.** True. The alternative — returning the
  stiff garbage instead — is not better, so this needs a real "both failed" signal rather than a
  third solve.
- **The rejection is not surfaced to users.** Already #1080 item 2, and this sharpens it: the
  field doc's advice is not actionable while no production path reports solver statistics. A
  `FitResult.warnings` entry is the repo-conventional answer.

---

**Suite after the fixes** (`49f3afb4`): 3298 lib tests and 142 integration binaries, 0 failures, 0 panics; clippy clean.

PR #1067, opened 2026-08-25
fix(ci): repair the slow-tests push-leg path filter [closes #949]

### LINKED ISSUE #949: fix(ci): slow-tests path filter names the removed src/api.rs and omits src/sim/** [#391]
`.github/workflows/slow-tests.yml:20` filters the `push: branches: [main]` trigger on `src/api.rs`. That file no longer exists — the API was split into the `src/api/` directory (`mod.rs`, `adaptive.rs`, `fit.rs`, …), so the pattern never matches. `src/sim/**` is not in the filter at all.

Consequence for adaptive dosing: all eight mrgsolve anchors are `slow-tests`-gated, so a PR runs only their parse/pin counterparts (`ci.yml:278-279` — fast tier, `--features ci`). A merge to `main` touching `src/api/adaptive.rs`, `src/sim/adaptive.rs`, or `src/sim/adaptive_control.rs` triggers nothing, and the anchors wait for the nightly cron. `src/ode/predictions.rs` *is* in the filter, which is why this has gone unnoticed — most adaptive work touches the driver.

`docs/model-file/adaptive-dosing.qmd:298` and `:332` state the anchors run "nightly and on push to `main`"; the push half is currently false for adaptive-only changes.

## Fix

- `src/api.rs` → `src/api/**`; add `src/sim/**`.
- Audit the rest of the list against the current tree for the same class of rot.
- Either correct the docs claim or leave it once the filter is right.

TeunP:
## Why

Closes #949.

The `slow-tests.yml` push trigger is supposed to re-run the heavy fit suite when a push to `main` touches code the fits exercise. Its path filter had rotted, and each stale or missing entry makes a push silently skip the suite — no failure, no signal, just no run.

The rot named in the issue was `src/api.rs`: the file was split into `src/api/`, the glob went dead, and an api-only change stopped firing the push leg. Reviewing that fix surfaced the same failure in eight more places, so the scope grew from "repair three entries" to "stop the filter being the kind of thing that rots".

**Measured, over the last 400 first-parent commits on `main`:**

| filter | commits that fire the suite |
|---|---|
| original 7-entry list | 292 |
| a corrected per-module list | 359 |
| this PR | 364 |
| neither | 37 |

So the original filter silently skipped **72 of 400** pushes that should have run the suite. The flip side is the honest cost of fixing that: the suite now fires on 91% of pushes to `main` rather than 73%, i.e. **+72 heavy runs per 400**. Those 72 are exactly the fit-affecting commits that were being skipped, so the cost is the point — but it is a real ~25% increase in slow-suite runs and should be a knowing choice, not a side effect. Curating by module recovers most of that but still costs a standing judgement call; going to `src/**` costs **4 extra runs in 400** (1% — one of them a `Cargo.lock` bump that ought to run anyway) and removes the failure mode entirely.

## What changed

**1. The filter is now one mechanical rule** — re-run when a push touches the code compiled into this job's test binaries, the data those tests read, or the definition of the job itself:

```yaml
- '.github/workflows/slow-tests.yml'
- 'Cargo.lock'
- 'Cargo.toml'
- 'data/**'
- 'examples/**'
- 'nonmem_anchor/**'
- 'src/**'
- 'tests/**'
```

Beyond the three entries in the issue, this closes gaps that were never listed and would have kept rotting:

- `src/ode/**` — the list named three `src/ode/*.rs` files and had **already re-rotted**: `rosenbrock.rs` (rodas4/rodas5p/rosenbrock23) and `explicit_rk.rs` (vern7) were added afterwards and never listed. Only the dispatch in `solver.rs` was covered, not the step bodies.
- `src/pk/**` — six slow-gated anchors exercise the closed forms directly.
- `src/stats/**` — was narrowed to `likelihood.rs`, omitting `residual_error.rs`, the sole owner of the residual-variance math under every OFV/CWRES.
- `src/survival/**` — the run step already adds `--features survival` *for the sole purpose* of executing `tests/tte_convergence.rs`, so the job paid to run TTE fits that no survival change could trigger.
- `src/frem/**`, `src/markov/**`, `src/nn/**`, `src/suggest_start/**`, `src/dosing.rs`, `src/io/**` — each with its own slow-gated fit or on every fit's dose/data path.
- **`examples/**` and `nonmem_anchor/**`** — 103 and 274 tracked files, read by the anchors at run time (`examples/adaptive_vanco_auc.ferx`, `Path::new("nonmem_anchor/transit_iov.csv")`). Neither was in any version of the list, and `src/**` does not cover them.
- **`Cargo.lock`** — a bare `cargo update` bumping nlopt, nalgebra or rand moves optimizer behaviour and RNG streams under every stochastic fit.

**2. `tests/slow_test_path_filter.rs` pins the filter in both directions**, so this cannot rot again:

- every tracked file under an input root the suite reads must be matched by the filter — this is what catches a *narrowed* filter, the general form of #949;
- every pattern in the filter must match at least one tracked file — this catches a *dead* glob, the specific form.

The input roots are **derived, not listed**: compile inputs are mechanical (`cargo test` builds `src/` and `tests/`), and runtime inputs are recovered from the fixture path literals in the test sources themselves. Add a test that reads `fixtures/foo.csv` and the guard fails until `fixtures/**` is in the filter — nobody has to remember. It also implements GitHub's filter-pattern semantics properly (`**` crosses `/`, `*` does not, `!` exclusions with last-match-wins).

## Alternatives considered

**Keep a curated per-module list.** This is what the first two rounds of this PR did. Rejected on measurement: it fires on 359 of 400 commits versus 363 for `src/**`, so it buys 4 skipped runs (1%) in exchange for the exact failure mode the PR exists to remove. The original rejection of `src/**` was on CI cost; that cost is not there.

**A shell guard (`.github/scripts/check-slow-test-paths.sh`) plus a dedicated CI job.** Written, then removed. Three reasons: it could only check glob *existence*, so a filter narrowed to `src/estimation/**` passed it green; its hand-rolled YAML parser had two verified false positives (a valid `!src/bin/**` exclusion, and a `branches:` block placed after `paths:`); and it duplicated a mechanism the repo already has — `tests/ci_workflow_endpoint_coverage.rs` pins a `ci.yml` invariant as a Tier-2 Rust test. Following that precedent costs no extra CI job, since `tests/` already builds on every PR.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

## Performance

Measured above: 364 of 400 first-parent commits fire the suite, against 292 under the original filter — +72 heavy runs per 400 pushes. Versus a curated per-module list (359) the difference is 5, so almost all of the increase is recovering silently-skipped runs rather than over-triggering.

## Tests
- [x] Regression test added that fails without this fix — `tests/slow_test_path_filter.rs` (Tier 2)

Mutation-tested, six mutations:

| mutation | expected | result |
|---|---|---|
| filter narrowed to `src/estimation/**` only | fail | fails (the case a shell existence-check could not catch) |
| `examples/**` removed | fail | fails |
| `src/**` → `src/api.rs` (the original #949 rot) | fail | fails, both tests |
| valid `!src/bin/**` exclusion added | fail | fails, naming `src/bin/generate_data.rs` — `cargo test` builds bin targets, so excluding them does leave them uncovered |
| `branches:` block moved after `paths:` | pass | passes (a false positive in the shell version) |
| second trigger with its own `paths:` added | pass | passes, `docs/**` does not leak in |

## Docs
- [x] No user-visible change

`CHANGELOG.md` deliberately not touched — CI-only, per the repo's changelog rule ("A PR that only touches internal refactors, tests, or CI does not need an entry").

## Checklist
- [x] `cargo clippy` clean — zero warnings from the new file under `--features ci --tests`
- [x] Functions on the `Dual2` sensitivity path stay differentiable (no production Rust changed)

## Docs & examples
- [x] No example execution step needed (CI-only change; no `.ferx` or `data/` file added)

## Reviewer hints

The interesting file is `tests/slow_test_path_filter.rs`, not the YAML. Two things worth a second pair of eyes:

1. **The derivation of input roots** (`required_inputs` / `runtime_input_roots`). It is the part that makes the guard self-maintaining, and it is also the part that can be wrong quietly. Dot-prefixed roots are deliberately excluded: CI-guard tests open `.github/workflows/*.yml` to assert on them, and without that rule the guard demanded `.github/**` in the filter — which would fire the 60-minute suite on every Dependabot edit. That is a rule about where fixtures live, not an exemption, but it is the judgement call in the file.
2. **`glob_matches`** implements GitHub's pattern semantics by hand. `github_glob_semantics` covers `**` vs `*` across `/`, `?`, exact files, and exclusion ordering, but it is a reimplementation of someone else's matcher and worth reading against [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#patterns-to-match-file-paths).

## Open questions

None outstanding. The earlier open question — whether to add a periodic check that listed paths still exist — is answered in the affirmative and implemented here, in the stronger set-coverage form rather than mere existence.


TeunP:
## Review — medium effort, no happy paths

The three changes are each **correct and justified**, and I verified all three against the tree rather than taking the description's word for it:

- `src/api.rs` is indeed gone; `src/api/` exists and `src/api/tests/{iov_integration,sde_integration,tests_sir_fallback}.rs` carry slow-gated fits, so the push leg has genuinely not covered them since the split.
- `src/sens/provider_tests.rs` carries a slow-gated test — `src/sens/**` is load-bearing, not speculative.
- `src/sim/adaptive.rs` is on the real adaptive path (`src/api/adaptive.rs:18` does `use crate::sim::adaptive::{…}`), and `tests/adaptive_vanco_anchor.rs` / `tests/adaptive_platelet_anchor.rs` are slow-gated. Correct.

The change is also strictly widening, as claimed. So: no bug *in* the diff.

My findings are all against the **completeness** of the list — which is the thing this PR is for, and which you explicitly asked about in the reviewer hints. Severity is bounded by the daily cron (a miss is a delay to the next 06:00 run, not a permanent hole), but that delay is exactly the signal #949 was filed to restore.

### Findings

**1. `src/ode/` is enumerated file-by-file and has already re-rotted.** (line 24)
`src/ode/rosenbrock.rs` and `src/ode/explicit_rk.rs` were both added after this list was written and are absent from it. Both hold user-selectable steppers reachable from `[fit_options] ode_method` — the *dispatch* is in `solver.rs:147`, which is listed, but the step bodies are not. `explicit_rk.rs`'s own module header names `tests/transit_nonmem_anchor.rs` (slow-gated) as the anchor it exists to serve. A push changing the Vern7 tableau or a Rosenbrock step matches nothing. This is the identical failure class to the `src/api.rs` rot, sitting in the hunk you're already editing. → `src/ode/**`.

**2. `src/pk/**` — the evidence you asked for is affirmative.** (line 20)
You left it out pending verification. Slow-gated tests that exercise `src/pk/` closed forms directly: `tests/transit_analytic_equivalence.rs` (3), `tests/ig_analytic_nonmem_anchor.rs` (2), `tests/parallel_mixed_nonmem_anchor.rs` (2), `tests/ss_absorption_nonmem_anchor.rs`, `tests/zero_order_absorption.rs`, `tests/per_route_lag_nonmem_anchor.rs`. Not a guess — it belongs in this PR.

**3. `src/dosing.rs` missing.** (line 20)
Per CLAUDE.md the single shared home of dose resolution + SS equilibration; imported by `src/pk/event_driven.rs`, `src/sens/propagate.rs`, `src/ode/predictions.rs`, `src/api/simulate.rs`. `tests/ss_absorption_nonmem_anchor.rs` (slow) turns on `SS_EQUILIBRATION_CYCLES` directly.

**4. `src/stats/` narrowed to `likelihood.rs`.** (line 38)
`src/stats/residual_error.rs` is per CLAUDE.md the sole owner of the residual-variance-and-derivatives math — the dense `R`, IWRES/CWRES, and the `variance_at*` / `dvar_df*` dispatch under every OFV. Slow-gated consumers: `tests/dense_residual_convergence.rs`, `tests/saem_combined_error.rs`, `tests/npde_validation.rs`. → `src/stats/**`.

**5. `src/survival/**` missing — and the job pays for it anyway.** (line 20)
The run step's own comment says `survival` is in the feature list *so that* `tests/tte_convergence.rs` (gated on both `survival` and `slow-tests`) executes here. So this workflow deliberately builds and runs TTE convergence fits, but a push touching `src/survival/parametric.rs` — the hazard closed forms those fits estimate — never fires the leg it was configured to serve.

**6. Four endpoint families absent entirely.** (line 20)
`src/markov/**` (`tests/markov_nonmem.rs`, 2 slow), `src/frem/**` (`tests/frem_warfarin.rs`, 6), `src/nn/**` (`tests/nn_fit_convergence.rs`, 1), `src/suggest_start/**` (`tests/suggest_start_comparison.rs`, 1). The CTMM matrix-exponential or the FREM covariate expansion can break its own NONMEM anchor with no push-leg signal.

**7. Nothing outside `src/` can trigger the job.** (line 15)
Three other inputs determine what runs and what it's compared against:
- **`data/**`** — 85 files under `tests/` read `data/*.csv` fixtures by literal path. Regenerate a fixture and every anchored slow test's numbers move, silently.
- **`tests/**`** — the 69 slow-gated files themselves, plus the shared `tests/common/mod.rs` harness. Land a new slow test on `main` and it doesn't run until the cron.
- **`Cargo.toml`** — defines both the `ci-test` profile this job builds with *and* the `slow-tests` / `survival` features that un-`#[ignore]` the fits.

### On your open question

> Should this be paired with a periodic check that every path in the list still exists?

Yes — and finding 1 is the argument for it: the list re-rotted **while this PR was being written to de-rot it**. A stale glob and a correct glob are visually identical, which is precisely why review didn't catch either. Worth doing here rather than separately.

Two notes on shape, if you add it:
- Check it as part of the **fast** PR job, not inside `slow-tests.yml` — this workflow only runs when the filter already matched, so a self-check here can never fire on the stale entry that caused the skip.
- Existence is the weaker half. The stronger check is the one that would have caught findings 2–7: assert that every top-level `src/` module is either listed or on an explicit allowlist of "cannot move a fit". That turns adding a module into a forced decision instead of a silent omission.

### Minor

`src/sens/**` and `src/sim/**` are inserted before `src/parser/**`, breaking the list's alphabetical order. In a list whose stated virtue is being reviewable by eye, that's worth keeping tidy.

TeunP:
## Findings resolved — pushed `66bcf442`

All seven, plus the open question.

**Path filter** — now matched by directory rather than enumerated by file, which is the property that keeps it from rotting again:

| Finding | Resolution |
|---|---|
| 1. `src/ode/` re-rotted | `src/ode/**` replaces the three named files, picking up `explicit_rk.rs` (vern7) and `rosenbrock.rs` |
| 2. `src/pk/**` | added — the six slow anchors are the evidence |
| 3. `src/dosing.rs` | added |
| 4. `src/stats/` narrowed | `src/stats/**` replaces `likelihood.rs`, picking up `residual_error.rs` |
| 5. `src/survival/**` | added |
| 6. markov / nn / frem / suggest_start | all four added |
| 7. non-`src/` inputs | `Cargo.toml`, `data/**`, `tests/**` added |

Also added `src/io/**` (the dataset reader every fit is fed from) and `src/types.rs` (`CompiledModel` / `Population` / `FitOptions`). List is alphabetical again, and every entry keeps a rationale comment — YAML-parsed to confirm the trailing comments don't leak into the globs: 19 entries, exactly as intended.

**Open question — yes, done here.** `.github/scripts/check-slow-test-paths.sh` fails when:

1. a listed glob matches no tracked file (existence — the `src/api.rs` failure), and
2. a new top-level `src/` module is neither listed nor on an explicit "cannot move a fit" allowlist (completeness — the findings 2–7 failure).

It runs as a `Slow-test path filter` job in **ci.yml**, not in slow-tests.yml, for the reason in my review: slow-tests.yml only runs once its filter has already matched, so a self-check there can never fire on the stale entry that caused the skip. No toolchain needed — checkout plus a shell script, so it costs nothing on the PR critical path. Runnable locally:

```bash
.github/scripts/check-slow-test-paths.sh
```

Mutation-tested rather than assumed — three mutations, each caught, and the unmutated tree passes:

- reintroduce `src/api.rs` → fails on both checks
- delete `src/pk/**` → fails the completeness check
- rename the `paths:` key → fails with an explicit "parsed zero globs" error rather than silently passing

That last one matters: a parser that quietly matched nothing would make the guard green against *any* filter, so it's an explicit hard failure.

**One judgement call worth flagging.** The allowlist currently excludes `src/categorical/` — it has no Tier-3 fit today, so listing it would be padding. But that is a fact about the tree right now, not a property, and the guard is what makes it safe: the day someone adds a slow categorical fit, adding `src/categorical/` to the module tree is not what fires the check — it's already allowlisted. If you'd rather not carry that footgun, the alternative is to drop it from the allowlist and add `src/categorical/**` to the filter now, at the cost of one unnecessary suite run per categorical push. Happy to switch if you prefer the conservative side.

CI cost: the filter is materially wider than before, so more pushes to `main` will run the 5–60 min suite — that is the intended trade, but it's a bigger step up than the original diff implied.

TeunP:
## Withdrawing the judgement call — `bd0ec31f`

I hedged on the allowlist instead of settling it. Settled now, by measurement, and the answer goes against the list I pushed an hour ago.

### The cost argument is empirically false in this repo

The original PR rejected `src/**` on CI cost. I kept a curated list on the same reasoning. Over the **last 252 first-parent commits on `main`**:

| | commits that fire the suite |
|---|---|
| curated per-module filter | 208 |
| `src/**` | **208** |
| neither (docs, CI, CHANGELOG) | 44 under both |

**Zero difference.** Not "small" — zero. Every commit that touches an allowlisted-only file also touches a listed module, so the allowlist has never once prevented a run. And no commit in that window touched only `src/bin/`, so even excluding the bin targets buys nothing.

(The first pass of this measurement reported 7 commits where the list "saved" a run. All 7 were `src/api.rs` — the rot itself, before the split. Correcting for that gave the zero above.)

### So the curated list was the worst of both

It bought no CI time, and carried two standing failure modes — **both of which had already fired in this tree**:

1. A named path that is deleted, renamed, or split leaves a dead glob, indistinguishable from a live one in review. `src/api.rs` (#949) — and `src/ode/` had re-rotted the same way while this PR was open.
2. A new module is a silent omission until someone remembers. That is what left `src/pk`, `src/survival`, `residual_error.rs`, `frem`, `markov`, `nn`, `suggest_start`, `dosing.rs` uncovered.

`src/**` cannot fail either way. Delete a module and it matches fewer files, never zero.

### What the filter is now

One rule — the code compiled into this job's test binaries, the data those tests read, and the definition of the job itself:

```yaml
- '.github/workflows/slow-tests.yml'   # the feature set / cargo command this job runs
- 'Cargo.lock'                         # a nlopt/nalgebra/rand bump moves fit numbers and RNG streams
- 'Cargo.toml'                         # the ci-test profile + slow-tests / survival features
- 'data/**'                            # fixture CSVs the NONMEM anchors read by literal path
- 'src/**'                             # everything linked into the fits
- 'tests/**'                           # the gated fits themselves + tests/common
```

`Cargo.lock` is new and no version of the list had it — a bare `cargo update` bumping nlopt, nalgebra, or rand moves optimizer behaviour and RNG streams under every stochastic fit. That is a real hole, independent of this argument.

### Guard script shrinks

Allowlist and the "is every module covered?" pass are **deleted**, not kept-and-tested. With `src/**` there is nothing left to be incomplete about, and the allowlist was itself the standing judgement call the guard exists to remove — keeping it would have been the guard sanctioning its own failure mode.

It keeps the existence check, which still has teeth: rename `data/` or `tests/`, or typo the workflow's self-reference, and a glob dies silently. Re-mutation-tested against the new shape — renaming `data/**`, typoing the self-path, and renaming the `paths:` key each fail; unmutated tree passes.

### Answering the thing I should have answered the first time

`src/categorical/` isn't a judgement call any more, because there are no judgement calls left. Nothing to defend, nothing to re-litigate when a module changes character, and no way for the next `src/api.rs` to happen.

TeunP:
## Correction + rework — `44235458`

### First, a correction to the numbers I posted above

[My earlier comment](https://github.com/FeRx-NLME/ferx-core/pull/1067#issuecomment-5407145778) claimed "252 first-parent commits ... exactly the same 208 commits ... zero difference". **That measurement was broken and those numbers are not reproducible.**

The classifier used `git show --name-only`, which emits an *empty combined diff* for a clean merge commit. 207 of the 400 first-parent entries on `main` are merges, and the script's `[ -z "$files" ] && continue` silently dropped every one — so the sample excluded this repo's dominant integration path, PR merges, which is exactly where multi-file changes land.

Recomputed with the first-parent diff (`git diff --name-only <sha>^1 <sha>`) over the full 400:

| filter | commits that fire the suite |
|---|---|
| original 7-entry list | **292** |
| a corrected per-module list | 359 |
| this PR | 363 |
| neither | 37 |

The delta between curating and `src/**` is **4 in 400 (1%)**, not zero — one `Cargo.lock`-only bump, one `src/build_info.rs`, two `src/bin/run_model.rs`. The conclusion holds (and one of those four is a dep bump that should run anyway), but the honest headline is the one I never computed: **the original filter silently skipped 67 of 400 pushes**. That is the number that justifies this PR.

### Two more input gaps, found by deriving instead of asserting

While rebuilding the guard to derive its expectations from the tree, it immediately failed on two roots that were in **no** version of this filter, and that `src/**` does not cover:

- **`examples/**`** — 103 tracked files. The adaptive anchors load `examples/adaptive_vanco_auc.ferx` and seven siblings.
- **`nonmem_anchor/**`** — 274 tracked files, read as `Path::new("nonmem_anchor/transit_iov.csv")` and `.../results/*.phi`.

Both are now in the filter. Neither would have been found by inspection; the derivation found them on first run.

### The shell guard is gone

Replaced by `tests/slow_test_path_filter.rs`. Three reasons, all of them defects in the version I pushed earlier:

1. **It could not catch the thing it was for.** Existence-only. Narrow the filter to `- 'src/estimation/**'` and it passed green while coverage collapsed.
2. **Its hand-rolled YAML parser had two verified false positives** — a valid `!src/bin/**` exclusion reported as a dead glob, and a `branches:` block placed after `paths:` whose `'main'` entry got validated as a path.
3. **It duplicated a mechanism this repo already has.** `tests/ci_workflow_endpoint_coverage.rs` pins a `ci.yml` invariant as a Tier-2 Rust test with a set equality and a broken-detector self-check. Following that precedent costs no extra CI job (the dedicated `Slow-test path filter` job is removed), and the bash version was the only `.sh` in the repo, in a `.github/scripts/` directory invented for it.

The new test asserts coverage **in both directions**: every tracked file under an input root the suite reads must be matched, *and* every pattern must match a tracked file. Input roots are derived — compile inputs mechanically (`cargo test` builds `src/` and `tests/`), runtime inputs from the fixture path literals in the test sources — so a new module or fixture directory fails the guard until it is listed. Nobody has to remember.

Mutation-tested, six mutations:

| mutation | result |
|---|---|
| filter narrowed to `src/estimation/**` only | **fails** — the case the shell guard passed |
| `examples/**` removed | fails |
| `src/**` → `src/api.rs` (the original rot) | fails, both tests |
| valid `!src/bin/**` exclusion added | fails, naming `src/bin/generate_data.rs` |
| `branches:` moved after `paths:` | passes (was a false positive) |
| second trigger with its own `paths:` | passes, `docs/**` does not leak |

On the fourth row: I expected a pass and the test was right, not me. `cargo test` builds bin targets, so excluding `src/bin/**` genuinely does leave four compiled inputs uncovered, and the diagnostic names them. Left as-is.

### Also fixed

- The `ci.yml` job comment describing an allowlist and module-coverage check that had already been deleted — moot now, the job is gone.
- One judgement call remains and is flagged in the reviewer hints rather than buried: dot-prefixed roots are excluded from the derived inputs, because CI-guard tests open `.github/workflows/*.yml` to assert on them and without that rule the guard demanded `.github/**` in the filter — which would fire the 60-minute suite on every Dependabot edit.
- PR description rewritten; it still said "four lines", "one file", "`src/**` rejected", and "no new tests needed".

Issue #1, opened 2026-05-13
Expose resolved gradient method in FitResult


TeunP:
## Problem

`FitResult` does not currently return the gradient method that was actually used during fitting. When the user requests `gradient = \"auto\"`, the engine resolves it at runtime to `ad` or `fd` depending on whether the autodiff feature is compiled in and whether the model has an analytical PK path (ODE models fall back to FD even when AD is available). The R layer currently stores only the *requested* method, which is labelled as such in the output but leaves users unable to confirm which method was actually used.

## Proposed change

Add a `gradient_used` field (string or enum variant label) to `FitResult`, set it after the `Auto` branch resolves to a concrete method. Expose it through `fit_result_to_list` so the R layer can display `Gradient (used): ad` alongside `Gradient (requested): auto`.

## Context

- The resolution logic lives in `ferx_rust_fit` in `ferx/src/rust/src/lib.rs` — `Auto` is set on `opts.gradient_method` but the concrete choice is made inside the engine.
- ODE models silently fall back from AD to FD even when `gradient = "ad"` is requested (documented in `ferx_fit()` roxygen). A `gradient_used` field would make this fallback visible without users needing to read warnings.

Issue #1070, opened 2026-08-25
bug(sens): TAD/TAFD are lifted as f64 constants, dropping the lagtime term for a TAD-dependent ODE RHS with IIV on the lag


TeunP:
Both ODE sensitivity walks hand the `TAD` / `TAFD` built-ins to the RHS program as **`f64` constants**, so `∂TAD/∂ALAG = −1` never enters the sensitivity chain. For a model whose `[odes]` right-hand side references `TAD` *and* carries an estimated lagtime with IIV, the analytic `∂f/∂η_lag` is therefore missing a term.

Found by code reading while fixing #1060 — not measured. The #1060 reproducer has a `TAD`-free RHS (`-(CL/V) * central`), so it is untouched by this.

## Mechanism

`eval_rhs_anchored` (`src/sens/ode_provider.rs:3333`) computes

```
tafd = t − first_dose_time
tad  = t − last_dose_eff
```

as plain `f64` and passes them to `OdeRhsProgram::eval_rhs_g`, which writes them into the variable table as constants (`src/parser/model_parser.rs:16509`, and its own comment says so: "constants w.r.t. the parameters, lifted as such").

But `last_dose_eff` *is* lag-dependent on the event-driven walk — it is set to the dose's **lagged arrival** (`ode_provider.rs:5240`, `last_dose_eff.max(t_event)` where `t_event = d.time + lag_val(idx)`). So for a subject whose lag carries an η, `TAD` at any point after that dose depends on `η_lag` with derivative `−1`, and a RHS term like `d/dt(central) = -(CL/V) * central * (1 + THETA_T * TAD)` picks up a contribution the walk drops.

The static walk has the same lifting but no lag-dependent anchor (it declines lagtime subjects outright, `ode_provider.rs:505`), so this is specific to the event-driven walk.

## Scope

Narrow: needs all of an `[odes]` RHS reading `TAD`/`TAFD`, an estimated lagtime, and IIV on that lagtime. There is an existing parity fixture for a TAD-dependent RHS on the TV-cov walk (`ode_provider_tvcov_tad_dependent_rhs_matches_production`, `ode_provider_tests.rs`), but it carries no lagtime, so the combination is uncovered.

## What to do

1. First, measure. Extend the existing TAD fixture with `LAGTIME = TVP * exp(ETA_P)` and check `check_vs_production` + `check_hessian_vs_production_fd`. If it passes, the reasoning is wrong and this closes.
[...]

Issue #1069, opened 2026-08-25
bug(sens): input-rate onset saltations inject the rate jump but not the field jump across a moving boundary


TeunP:
The two **input-rate onset** saltations on the ODE event-driven walk inject only the forcing's rate jump across the moving boundary, not the jump in the RHS field itself. When a time-varying covariate changes across the onset *and* the compartment already holds drug, the omitted term is a real first-order gradient error, and the Jacobian in the second-order term is taken from one side only.

Found by sweeping the `K_DOSE` branch while fixing #1060; not measured, so this is a code-reading finding rather than a reproduction.

## Sites

Both in `src/sens/ode_provider.rs`:

- `:5026` — the input-rate onset at a lagged dose arrival (`igd` / `transit` / `first_order` / `weibull` / `zero_order` switching on at `t_dose + ALAG`).
- `:5321` — the per-route onset `K_ROUTE_ONSET` (#859), at `t_dose + ALAG_cmt + lag_route`.

Both call `inject_rate_saltation`, whose contract (doc at `:3409`) is explicit that it assumes the state is continuous **and the field is the same on both sides**: "only `ẋ` jumps by the forcing `Δr`", with `J⁻ = J⁺ = J`. That assumption is what #1060 turned out to violate for the bolus arrival, and what #653 had already handled for the rate-*off* boundary via `general_rate_off_saltation`.

#880 moved the onset's `Δr` and `frac` to the post-arrival snapshot (`onset_params`, `:4962`), which fixed the *rate* half. The *field* half — `(f(u; P_pre) − f(u; P_post))·δlag` — has no term at all, and the `J` fed to the δ² coefficient is still single-sided.

## When it bites

The missing first-order term is proportional to the state at the onset, so it vanishes for the common case of a first dose into an empty compartment. It is non-zero when:

- the model has an `init(...)` baseline, or
- the dose is a later one with drug still present,

**and** a covariate record separates the two sides of the onset.

That is the same shape as the lagged-infusion defect fixed alongside #1060, where the analogous term measured 0.2% on a 70→76 weight step and 2.2% on 70→140, scaling with the field contrast. I would expect the same order here.

## Suggested fix

[...]

Issue #1068, opened 2026-08-25
bug(sens): a lagged arrival landing exactly on a covariate-changing record gives a 1% eta-gradient error (ODE event-driven walk)


TeunP:
On the ODE event-driven walk, a lagged dose arrival that lands **exactly on a covariate-changing record** yields an `∂f/∂η_lag` that is off by ~1% on every observation after it. The value stays exact, so nothing in the output looks wrong.

Found while fixing #1060 and deliberately left out of that PR: it is a different defect class (measure-zero, continuous objective) and the fix is a semantics decision rather than a bug fix.

## Why it happens

The arrival `τ(η) = t_dose + ALAG` is a moving boundary, and the walk assigns PK snapshots by the end-of-interval convention: the segment *ending* at `τ` takes the dose record's snapshot, the segment it *opens* takes the next record's. #1060 fixed the case where those two differ and `τ` sits strictly between records.

When `τ` coincides with a record at `t_rec`, the assignment **flips** as `τ` sweeps through it:

- `τ = t_rec − ε`: the stretch `(t_rec − ε, t_rec)` is governed by the record's field.
- `τ = t_rec + ε`: that same stretch is governed by the dose row's field.

As `ε → 0` the flipped stretch shrinks to nothing, so the **value is continuous** — but the two one-sided derivatives differ by a finite amount. The analytic walk returns one side; a central finite difference straddles the coincidence and returns the average of both. Half the jump is the observed error.

## Measured

Model and subject are the `#1060` fixtures now in `src/sens/ode_provider_tests.rs` (`ONECPT_IV_LAG_CROSS_ODE`, `lag_crossing_subject`), at `θ = [10, 50, 0.75, 0.5]`, `η = [0.1, −0.05, 0]` — so `ALAG = 0.5` exactly and the arrival is exactly `1.5`, the record where `WT` steps 70 → 74. Measured on the branch that already carries the #1060 fix, i.e. this is what survives it.

| obs | t | value (analytic vs production) | `∂f/∂η_lag` analytic | central FD | rel |
|---|---|---|---|---|---|
| 2 | 1.5 | `2.102542192752` = `2.102542192752` | `0.254676664` | `-1051270.97` | — |
| 3 | 2.0 | agree | `0.225071852` | `0.227345366` | **1.0e-2** |
| 4 | 4.0 | agree | `0.134657326` | `0.136017538` | **1.0e-2** |
[...]

PR #1071, opened 2026-08-25
fix(sens): evaluate a moving dose boundary's two sides on their own covariate snapshots [closes #1060]

### LINKED ISSUE #1060: bug(sens): ODE + time-varying covariates + IIV on lagtime returns a wrong analytic eta-gradient (grad off 300x, Hessian 23x) while reporting analytic
On an **ODE** model, a subject with **time-varying covariates** and an **estimated lagtime carrying IIV** is served by `subject_sensitivities` as analytic, but its `∂f/∂η` and `∂²f/∂η²` are wrong by 2–3 orders of magnitude. The value itself is exact, so a fit converges quietly to the wrong place rather than failing loudly.

Found while chasing an unrelated aside during the #1055 review. **Not caused by #1055** — it reproduces with no `init(...)` at all.

## Measurement

Assert-margin against central FD of `compute_predictions_with_tv` (the same reference `check_full_provider_vs_fd` uses), with the harness's own bounds — `max_relative` 2e-4 on the gradient, 3e-3 on the Hessian, plus the matching absolute epsilons. **Margin ≤ 1.0 = the harness assert passes.**

| case (ODE, 1-cpt IV bolus) | value | grad | H@1e-2 | H@3e-3 | H@1e-3 | H@3e-4 | H@1e-4 |
| **A** TVcov + init + LAGTIME(IIV) | 0.00e0 | **1.99e3** | **5.88e2** | **5.88e2** | **5.88e2** | **5.88e2** | **5.88e2** |
| B TVcov + init, no lag | 0.00e0 | 7.75e-1 | 5.29e-1 | 4.76e-2 | 7.49e-3 | 7.57e-3 | 7.57e-3 |
| **C** TVcov + LAGTIME(IIV), no init | 0.00e0 | **2.99e2** | **2.29e1** | **2.29e1** | **2.29e1** | **2.29e1** | **2.29e1** |
| D init + LAGTIME(IIV), no TVcov | 0.00e0 | 7.75e-1 | 1.34e-1 | 1.21e-2 | 7.49e-3 | 7.57e-3 | 7.57e-3 |
| E TVcov + LAGTIME, **no IIV on lag** | 0.00e0 | 7.75e-1 | 3.97e-1 | 3.57e-2 | 7.49e-3 | 7.57e-3 | 7.57e-3 |
| F const WT (static walk) + init + LAGTIME(IIV) | 0.00e0 | 7.75e-1 | 1.34e-1 | 1.21e-2 | 7.49e-3 | 7.57e-3 | 7.57e-3 |

### Why this is not FD conditioning

Three independent reasons:

1. **`value` margin is exactly `0.00e0` in every row.** The analytic walk's `obs.f` and the FD reference's predictor agree bit-for-bit, so both sides are differentiating the *same* function. Only the derivatives disagree.
2. **The failing rows are dead flat across a 100× span of `h`** (1e-2 → 1e-4), while every passing row converges monotonically (`5.29e-1 → 7.57e-3`). Per the `check_full_provider_vs_fd_with_step` docstring, "the FD converges *to* the analytic value as `h` is refined — the analytic kernel is the accurate party". Here it converges to something else.
3. **The first-order gradient fails too** (1.99e3 / 2.99e2), not just the second-order block, so this is not the known ill-conditioning of the 4-point Hessian stencil.

### Harness validation

Before trusting any of the above I ran the repo's own `lagtime_tvcov_walk_matches_fd_of_production` fixture, subject, θ and η — which is green in CI — through the identical measurement:

| h | value | grad | hess |
|---|---|---|---|
| 1e-2 | 3.38e-7 | 3.49e-4 | 4.14e-1 |
| 1e-3 | 3.38e-7 | 3.49e-4 | 4.14e-3 |
| 1e-4 | 3.38e-7 | 3.49e-4 | 3.38e-3 |

All ≤ 1.0 — the harness reproduces the CI pass, so the apparatus is sound.

## Localization

Comparing rows: **C fails, E passes, D and F pass.** The defect needs *all three* of:

- an **ODE** model (the closed-form twin is covered and green — that is the `lagtime_tvcov_walk_matches_fd_of_production` fixture above, `pk one_cpt_oral(..., lagtime=LAGTIME)`),
- a subject on the **event-driven TV-covariate walk** (F, with a constant covariate, takes the static walk and passes),
- **η on the lagtime** (E, with `LAGTIME = TVP` and no `ETA`, passes).

That points at the lagtime's η-jet not being carried through the event-time saltation on the **ODE** TV-cov walk specifically. `init(...)` is not required; it only amplifies the error (588× vs 23×) by adding pre-arrival signal.

Closest relative is #880 (closed) — "Analytic Hessian wrong for IIV on an absorption lag with a rate-on input-rate forcing (`inject_rate_saltation` δ² term)" — same IIV-on-a-lag saltation family, different trigger. Worth checking whether this is the same δ² term on a different path.

## Repro

```toml
[parameters]
  theta TVCL(10.0, 1.0, 100.0)
  theta TVV(50.0, 5.0, 500.0)
  theta THETA_WT(0.75, 0.01, 2.0)
  theta TVP(0.7, 0.05, 5.0)
  omega ETA_CL ~ 0.09
  omega ETA_V  ~ 0.09
  omega ETA_P  ~ 0.04
  sigma PROP_ERR ~ 0.04
[individual_parameters]
  CL = TVCL * (WT/70)^THETA_WT * exp(ETA_CL)
  V  = TVV  * exp(ETA_V)
  LAGTIME = TVP * exp(ETA_P)
[structural_model]
[odes]
[covariates]
  WT continuous
[scaling]
  obs_scale = V
[error_model]
  DV ~ proportional(PROP_ERR)
[fit_options]
  ode_reltol = 1e-10
  ode_abstol = 1e-12
```

Subject — one bolus, `AMT=100` into `CMT=1` at `t = 1` (`dose_covariates` `WT = 70`); observations at `t = 0, 0.5, 1.5, 2, 4, 8` with `WT = 70, 72, 74, 76, 80, 85`; `covariates` `WT = 70`. Evaluate at `θ = [10, 50, 0.75, 0.7]`, `η = [0.1, -0.05, 0.07]`, so the lag is `0.7·e^0.07 ≈ 0.751` and the dose arrives at `≈ 1.751` — between the `t = 1.5` and `t = 2.0` observations, i.e. the moving boundary sits inside a covariate segment change.

Then `check_full_provider_vs_fd(&model, &subject, &theta, &eta)` fails on `df_deta`.

## Impact

`subject_sensitivities` returns `Some` here, so this is on the production FOCEI path (`subject_sensitivities_impl` → `ode_subject_sensitivities` for any `ode_spec.is_some()` model in scope) — not a defensive-only wrong answer. A fit of an ODE model with time-varying covariates and IIV on an absorption lag gets a wrong inner EBE gradient and a wrong outer θ/Ω/σ gradient, with no diagnostic. The prediction is exact, so nothing looks wrong in the output.

If the intended scope was for this cell to fall back to FD, then the support predicate is the bug rather than the kernel — either way the CLAUDE.md rule applies:

> If a model is outside the analytic scope it must route to FD via the support predicates; unit-test that routing so a scope gap fails loudly to FD instead of silently returning a wrong gradient.

There is currently no ODE + TV-cov + lagtime parity fixture in `src/sens/provider_tests.rs` — the covered `lagtime_tvcov_walk_matches_fd_of_production` cell is the closed-form one.

## Docs note

`docs/estimation/foce.qmd:34` currently claims ODE models are analytic for bolus dosing "with or without **time-varying covariates** … and now also covers an **estimated lagtime**". Whichever way this is fixed, that sentence needs to match.

TeunP:
## Why

Closes #1060.

On an ODE model with time-varying covariates and IIV on the lagtime, `subject_sensitivities` returned an `∂f/∂η` that was wrong by up to 300× (Hessian 23×) **while reporting analytic**. The prediction was exact, so a FOCE/FOCEI fit converged quietly to the wrong optimum with nothing in the output to look at — the silent-wrong-number class this repo treats as most severe.

A dose under an estimated lagtime lands at a moving boundary `τ(η) = t_dose + ALAG`, and the walk injects the resulting jump as a saltation. Segment params follow the NONMEM end-of-interval convention (`predictions.rs:4585`, `Kind::Dose => pk_at_dose[idx]`): the segment **ending** at `τ` runs under the dose record's covariate snapshot, the segment `τ` **opens** runs under the next record's. The injection evaluated *both* sides on the dose record's, so whenever those two records carried different covariates the post-arrival velocity `g(x⁺)`, its Jacobian `J⁺`, and the cross term `J⁺·ẋ⁻` all used the wrong parameters.

The error lands entirely in the derivative — `δlag` is jet-only (value 0), which is exactly why the value stayed bit-exact — and scales with the ratio of the two covariate-scaled fields.

The bolus arrival was the last saltation in the walk without a post-side snapshot lookahead. The rate-on onset got one in #880 (whose comment names this failure mode almost verbatim: "under a TV covariate crossing the onset the dose snapshot diverges (a several-percent gradient error)"), and the rate-off boundary got the two-sided treatment in #653.

## What changed

**The bolus arrival** (`ode_provider.rs`, `K_DOSE` estimated-lagtime branch). `g(x⁺)`, `J⁺` and the cross term now evaluate on the arrival segment's snapshot; `g(x⁻)` and `J⁻` stay on the dose record's. The `coef2` formula and the injection line are untouched — the block comment already stated the correct two-sided math (`ẋ̇± = J(x±)·g(x±)`), only the wiring disagreed with it.

**The lagged-infusion rate-on**, same instant, carried the matching defect and is fixed with it. That path used the closed-form `inject_rate_saltation`, whose contract explicitly assumes the field is identical on both sides (state continuous, `J⁻ = J⁺`, only the rate jumps) — so it has no term at all for the field difference and evaluates its Jacobian one-sided. It now always takes `general_rate_off_saltation`, the sign-mirror of the rate-*off* handling built in #653.

**Both velocities carry every forcing that straddles the boundary.** `ẍ = J·(g + f)` for a concurrently-active infusion, zero-order window or pointwise `R_in`, so leaving `f` out of `g±` drops `½(J⁻−J⁺)·f` from the curvature. That cancelled exactly while both sides read one snapshot, which is why it survived every earlier fixture; making the sides differ exposes it. Membership is **strict** — a forcing that *toggles* at the instant injects its own saltation at that same event, and the two compose to the exact one-sided form only if the boundary leaves it out. Straddling versus toggling is the discriminator, not frozen versus co-moving: a co-moving window that merely spans the arrival is grid-constant on both sides and must be counted.

**The five post-boundary snapshot lookaheads become one helper**, returning the first event strictly later than the boundary — which is what the walk itself integrates against, since a co-timed event opens a zero-length segment. The single exception is a co-timed sibling **dose**, which must still be returned so that two doses arriving together telescope.

A constant-covariate model with no concurrent forcings is bit-identical throughout: `boundary_velocity` with an empty forcing set reduces to the bare `eval_rhs_anchored` the branch called before, and the lookahead falls back to the same snapshot. The one deliberate exception is the lagged-infusion rate-on, which now takes the general two-sided form even when the snapshots agree; `general_rate_off_saltation` reduces to `inject_rate_saltation` **analytically** in that case, not bit-for-bit, so such a fit can move in the last few ULP.

## Numerical validation

**NONMEM 7.6.0 anchor** (`nonmem_anchor/tvcov_lag_saltation.ctl`). The FOCEI objective builds its `h` matrix from this same analytic `∂f/∂η` (`inner_optimizer.rs`), so a wrong sensitivity moves the **reported OFV**, not only the search path — `#OBJV` under `MAXEVAL=0 POSTHOC` is therefore a valid and sharp anchor:

| | OFV | vs NONMEM |
|---|---|---|
| NONMEM 7.6.0 `#OBJV` | `-302.91470812769074` | — |
| ferx, this PR | `-302.914703` | **5e-6** |
| ferx, base | `-310.431699` | **7.52** |

All three EBEs reproduce to every printed digit on all eight subjects. (An earlier revision of this description claimed no NONMEM run could anchor the change because predictions are bit-exact; that is true of `IPRED` and wrong about the objective.)

Red-phase measurement, on the issue's own reproducer:

```
∂f/∂η_lag   analytic 0.34487735   FD reference 0.36681859
relative miss                     0.059815
1 − (70/76)^0.75                  0.0598143
```

Five significant figures. That is the arithmetic signature of evaluating the post-arrival velocity under the dose row's `WT = 70` instead of the arrival segment's `WT = 76` — so the fixture pins the mechanism, not merely the symptom. Value margin was `0.00e0` throughout, before and after.

On the infusion sibling (which needs an `init(...)` baseline to make the state non-zero at rate-on, the only condition under which the omitted field term is non-zero): `∂f/∂η` off **0.2%** on a 70→76 weight step and **2.2%** on 70→140, scaling with the field contrast exactly as the mechanism predicts.

## Tests

Thirteen new fixtures. Each is verified non-vacuous by reverting the specific change it guards and observing it go red — the numbers below are those red-phase measurements.

Six of them came out of the review round:

| fixture | what it pins | red-phase |
|---|---|---|
| `..._arrival_straddles_frozen_infusion_...` | a straddling forcing belongs in `v±` | `∂²f/∂η_LAG²` `0.87941` vs `0.89639` |
| `..._arrival_straddles_comoving_infusion_...` | …**including** when it co-moves — this is the case that rules out "co-moving ⇒ exclude" | `-0.65279` vs `-0.48247` |
| `..._arrival_straddles_input_rate_...` | a snapshot-dependent `R_in` makes the omission first-order | gradient `0.75665` vs `0.79632` |
| `..._arrival_straddles_covariate_free_input_rate_is_exact` | control for the above; green on both sides by design | — |
| `..._iov_lagtime_rate_on_across_occasion_boundary_...` | value-only snapshot equality drops a κ jet | `∂²f/∂η_LAG∂κ` `+0.05207` vs `-0.01478` (sign flip) |
| `..._co_timed_doses_telescope_across_covariate_change` | a co-timed sibling dose must stay in the lookahead | gradient `1.1222` vs `0.8954` |

and the original seven:

| fixture | what it pins |
|---|---|
| `..._arrival_crosses_covariate_...` | #1060 row C — the issue's model and subject verbatim |
| `..._init_arrival_crosses_covariate_...` | row A, three-way weight ramp so pre/post/prev are three distinct snapshots — catches a "fix" that moved `g(x⁻)` too |
| `..._arrival_before_covariate_change_...` | control: identical except the arrival segment's record carries the dose row's weight. Green before **and** after |
| `..._two_doses_split_segments_...` | one no-crossing arrival and one crossing arrival in a single walk, with residual drug at the second |
| `..._indexed_alag_...` | the `ALAG1` spelling (#369), which resolves through the dose-attribute map |
| `..._infusion_rate_on_crosses_covariate_...` | the sibling path |
| `ode_lagtime_tvcov_arrival_crosses_covariate_matches_fd_of_production` | the same geometry under the strict harness in `provider_tests.rs` (2e-4 / 3e-3 rather than the ODE module's 2e-3), asserting analytic service first so the comparison cannot be FD-vs-FD |

`ode_provider_lagtime_tvcov_matches_production` also gains a `check_hessian_vs_production_fd` call. Its Hessian block was self-FD — differentiating the analytic gradient by itself — which is structurally blind to an error that is consistent across η, i.e. to this one. That is the coverage gap that let a green fixture with η-on-lag **and** TV covariates sit next to this bug: its single dose is at `t = 0` and arrives before any covariate update, so its arrival never crossed one.

Mutation check: reverting only the snapshot threading re-reddens all five crossing fixtures while the equal-snapshot control and both closed-form tests stay green.

`cargo test --lib --no-default-features --features ci`: **3290 passed, 0 failed**. The full `--features ci` suite over every test binary: green locally, 0 failures anywhere. CI green on `a4a981da` across all seven checks, `codecov/patch` included.

## Docs

`docs/estimation/foce.qmd` claimed an `init(...)` baseline combined with a reset / estimated lagtime / infusion / input-rate forcing / steady state / modeled dose "still falls back to FD". That has been false since #486 — the gate admits all of them and green tests assert it. The sentence was left stale on purpose during #1055, because correcting it then would have documented a feature that returned wrong gradients. It now describes something that works, so it is corrected here, along with a statement of which snapshot each side of a moving boundary uses and an explicit enumeration of the approximations that remain.

## Deliberately not in this PR — tracked by #1078

The five saltation defects left open by this work have their own epic, **#1078** — *exact event-time saltations at every moving boundary on the ODE event-driven walk*. It carries the dependency order (they are not independent: #1076 restructures the sites #1075 would otherwise patch one by one, and #1076/#1068 share the `Option` plumbing), the shared acceptance test, and the three conventions this PR introduced that are easy to get backwards when touching `v±` at any site.

Three of those five are approximations that survive **in this cell**. All three are now named in `foce.qmd` rather than left silent, and each is filed with the measurement that would otherwise have to be redone:

- **#1068** — an arrival landing **exactly** on a covariate-changing record returns a one-sided derivative where the two-sided one does not exist (~1 % on downstream observations; value continuous and exact). Measure-zero, and the fix is a semantics decision that needs `integrate_tvcov_readout` to return `Option` again. The lookahead unification here does make the returned value the genuine one-sided limit of the branch the walk's own ordering uses, which it previously was not.
- **#1076** — two *moving* boundaries colliding at one instant each inject the whole covariate field jump, counting it twice: **5.9 % on the gradient**. Pre-existing and bit-identical either side of this PR, but structurally reachable rather than measure-zero (an infusion ending exactly when the next dose is given).
- **#1075** — `ẍ = ∂f/∂t + J·f` for an explicitly time-dependent field, and a pointwise `R_in(tad)` is one; the `½(∂f⁻/∂t − ∂f⁺/∂t)` term is not carried. 6.4 % on the worst `∂²f/∂η_LAG²` element of the input-rate fixture, identically zero when the kernel reads no covariate (which the control fixture demonstrates). Needs a `∂R_in/∂tad` primitive that exists only at `tad = 0` today and applies to all four saltation sites. The fixture asserts a bound so it cannot grow silently.

The other two epic members are **#1069** (the input-rate onset saltations still inject the rate jump but not the field jump — the last closed-form sites) and **#1072** (the remaining value-only snapshot comparisons at the two rate-off boundaries, which is where the κ-jet defect fixed here still lives).

Also filed from the same sweep, and deliberately **outside** #1078:

- **#1070** — `TAD`/`TAFD` lifted as `f64` constants, dropping `∂TAD/∂lag = −1`. Now measured at **1.5–64 %** on a model `ode_tvcov_supported` admits, wrong at *every* observation including ones that precede any boundary — so it is the segment integration, not a saltation. No longer a code-reading hunch. Out of the epic because the fix is a dual-`tad` signature change across both walks, far wider than saltations; it blocks *measurement* rather than arithmetic (it is why review finding #3's TAD anchor shipped here on construction alone).
- **#1073** (yours) — the lagged dose record's covariate snapshot stretched to the arrival, **14.9 OFV** against NONMEM. A production *value* divergence, the most severe item in the area, and a **precondition** for three of the epic's five: moving the timeline break changes where and whether the field jumps at all. Recommended order overall is #1073 → #1070 → #1078.
- **#1077** — a `CMT=0` infusion is delivered by the production predictor and skipped by the analytic walk, another **value** divergence. No edges in either direction.

## Cross-repo dependency

No `pub` API change; the fix is internal to the sensitivity walk.

## Breaking changes
- [x] None

Behaviour changes for affected fits — an ODE model with time-varying covariates and IIV on a lagtime whose arrivals cross a covariate change will now converge somewhere different, because it was previously following a wrong gradient. That is the point of the fix, and the CHANGELOG entry says so in user-facing terms.

## Performance

The bolus path swaps a slice pointer; the extra `Dual1` param vector and the extra prepared-forcing set are both shared when the two snapshots agree, so a constant-covariate model allocates nothing new. The lagged-infusion rate-on takes about four extra RHS point-evaluations per arrival — it no longer has a fast path, because the predicate that guarded it was value-only and dropped κ jets under IOV. That is noise beside the hundreds of RHS evaluations a segment spends integrating between two events. The five separate lookahead scans become one, and it no longer runs at all for a lagtime-free dose.

## Tests

## Docs

## Checklist
- [x] `cargo clippy` clean (no new warnings in the edited region; the 261 pre-existing lib warnings are #1023)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — both edits are inside `*_g<T: PkNum>` generic code and were verified through the `Dual2` (outer), `Dual1` (inner) and IOV monomorphizations by the parity + inner/outer tests

## Reviewer hints

The load-bearing question is **which snapshot belongs to which side**, and it is easy to get backwards. The pre-side uses the *dose row's* snapshot rather than the previous record's — that is not an oversight, it is what production does (`predictions.rs:4585`, `Kind::Dose => pk_at_dose[idx]`, for the segment *ending* at the event). `init_arrival_crosses` (three distinct snapshots) is the fixture that fails if that is misread in either direction.

Second: the unified lookahead returns the first **strictly-later** event, with **one exception** — a co-timed sibling dose. Both halves are load-bearing and pull in opposite directions. Strictly-later is right because a co-timed event opens a zero-length segment the walk never integrates; the dose exception is right because two doses arriving together telescope only if the first's post side is the second's pre side. Getting either half wrong is a silent multi-percent gradient error, and each has its own fixture (`co_timed_doses_telescope_across_covariate_change` for the exception).

Third: the membership rule in `boundary_velocity`. `strict` at the `K_DOSE` boundaries, inclusive at the rate-offs, and the doc comment carries the composition algebra that decides which is which. `ode_provider_bolus_concurrent_infusion_lagtime_hessian_matches_fd` (co-timed) and `..._arrival_straddles_comoving_infusion_...` (straddling) are the pair that pins it — they are both co-moving and they need opposite treatment, which is the thing that makes this easy to get wrong.

## Open questions

None outstanding. The earlier question here — whether the `pk_snapshot_equal` fast path belonged on the infusion path but not the bolus — was resolved by measurement: it belongs on neither, because value-equality drops κ jets at an IOV occasion boundary (`∂²f/∂η_LAG∂κ` came out with the wrong sign). The two pre-existing rate-off uses are #1072.


TeunP:
Self-review pass over this diff (independent line read, not a re-run of CI). Two findings; one was mine and is fixed in `2603d765`, the other is filed.

**1. Fixed — the lookahead ran for models that cannot use it.** I hoisted `post_params` to the top of the `K_DOSE` branch so the bolus arrival and the lagged-infusion rate-on could share it. That put its forward timeline scan on *every* dose event, including models with no lagtime — where both consumers are unreachable and there is no moving boundary to inject a saltation at. For a subject with many doses that is an O(doses × events) walk added to every inner-EBE and outer-gradient evaluation, paid by TV-covariate models that get nothing back. Both consumers sit under `has_lagtime`, so the scan is now gated on it and otherwise returns the dose row's snapshot — the same value the loop's own fallback arm gives. Suite still 3284/0.

**2. Filed as #1072 — `pk_snapshot_equal` is value-only, which is unsound under IOV.** The infusion fast path here asks that predicate whether the field jumps. Its own doc scopes the argument to the *non-IOV* PK map, and correctly so: under IOV two snapshots can carry identical covariate values and different κ jets, so `(v⁻ − v⁺)` has zero value but a non-zero jet, and the closed form drops it. IOV does reach this kernel (`run_subject_iov` → `integrate_tvcov_readout` at `:2875`).

I kept the fast path rather than fixing it here, for two reasons. It is **pre-existing** — #653 asks the same predicate at the rate-off boundary (`:5489`) and has the same exposure — so this PR extends an existing gap to a second boundary rather than creating a class. And removing it is not free: the closed form is one `J·g` eval against roughly five for the general form, paid on every lagged infusion in every evaluation including all the constant-covariate ones, which is exactly what #653's review optimised away. #1072 covers both sites and argues for making the predicate jet-aware, which keeps the fast path firing for every non-IOV model where the doc's argument holds exactly.

This also resolves the asymmetry I raised in the PR description's open question: the bolus path has no fast path (so it is unaffected), the infusion path does (so it is exposed). That was a deliberate choice on the bolus side and an inherited one on the infusion side, and #1072 is where it gets settled.

Worth knowing for the merge itself: now that #1067 has landed, this will be the first PR to fire the slow-tests push leg via the new `src/sens/**` entry. It cannot move any heavy fit — the four slow-gated tests carrying a lagtime (`first_order_alag_nonmem_anchor`, both `gen_transit_multidose_*`, `iov_convergence`) have no covariate column at all in their datasets, so the two snapshots are equal and this change is bit-identical for them. Any red in that run is main's existing #1053.

TeunP:
## Review — `xhigh`, adversarial

Reviewed at `2603d765`. Built PR head and base (`95ac28e3`) in throwaway worktrees and ran the suite both sides. The seven new fixtures pass and they do pin the mechanism — the `1 − (70/76)^0.75` signature is the real thing, and the `init_arrival_crosses` three-snapshot fixture does rule out a fix that moved `g(x⁻)` too. The two reviewer hints check out (details at the bottom).

The fix is right and the diagnosis is right. Two things on top of that, both measured rather than argued:

* **It anchors against NONMEM, and it lands.** `MAXEVAL=0 POSTHOC` on a crossing dataset: `#OBJV −302.91470813` vs this PR's `−302.914703` (5e-6) and base's `−310.431699` (**7.52 off**), with all three EBEs reproduced to every printed digit on all eight subjects. The PR's premise for skipping NONMEM is wrong — the FOCEI objective takes its `∂f/∂η` from the *analytic* provider (`inner_optimizer.rs:1097`), so this bug was moving the reported OFV, not just the search path. Full run in **#10**; control stream and data ready to commit.
* **It is incomplete at second order.** Reproducer in **#1**.

Eleven findings below, ranked; #1 is the only one I'd call blocking.

---

### 1. `coef2` still drops the `½(J⁻−J⁺)·f` curvature of concurrent forcings — **measured**

`src/sens/ode_provider.rs:5264` / `:5278` / `:5339`

The bolus branch builds `g_minus`/`g_plus` from the bare `eval_rhs_anchored`, not from `boundary_velocity`, so concurrently-active forcings (another dose's infusion, a zero-order window, a pointwise `R_in`) are absent from `ẋ̈±`. That omission **cancelled exactly while `J⁻ = J⁺`** — which is why it was invisible before. This PR makes the two Jacobians differ, and the cancellation goes with it:

```
coef2 = ½J⁻v⁻ + ½J⁺v⁺ − J⁺v⁻      with v± = g± + f
      = [½J⁻g⁻ + ½J⁺g⁺ − J⁺g⁻] + ½(J⁻−J⁺)·f
                                    ^^^^^^^^^^^^ dropped
```

First order is safe: `f` is identical on both sides for infusion / zero-order forcings, so it cancels in `(g⁻−g⁺)`. The loss is `δlag²`-only, i.e. `∂²f/∂η²` — the FOCEI Hessian and the covariance step, not the gradient.

Measured on the fixture below, `∂²f/∂η_lag²` vs double-FD of `compute_predictions_with_tv`:

| | analytic | production FD | miss |
|---|---|---|---|
| PR head `2603d765` | `0.87941` | `0.89639` | **1.9 %** |
| base `95ac28e3` | `0.30034` | `0.59657` | 50 % |

So this PR improves the element 26× and leaves 1.9 % on the table. `check_vs_production` (gradient) passes on head — consistent with the term being second-order only.

<details>
<summary>Reproducer — append to <code>src/sens/ode_provider_tests.rs</code>, fails on <code>2603d765</code></summary>

```rust
const PROBE_TWO_CMT_LAG_INF_CROSS: &str = r#"
[parameters]
  theta TVCL(10.0, 1.0, 100.0)
  theta TVV(50.0, 5.0, 500.0)
  theta THETA_WT(0.75, 0.01, 2.0)
  theta TVP(0.7, 0.05, 5.0)
  theta TVK21(0.4, 0.01, 5.0)
  omega ETA_CL ~ 0.09
  omega ETA_V  ~ 0.09
  omega ETA_P  ~ 0.04
  sigma PROP_ERR ~ 0.04
[individual_parameters]
  CL = TVCL * (WT/70)^THETA_WT * exp(ETA_CL)
  V  = TVV * exp(ETA_V)
  K21 = TVK21
  ALAG1 = TVP * exp(ETA_P)
[structural_model]
  ode(obs_cmt=central, states=[central, periph])
[odes]
  d/dt(central) = -(CL/V) * central + K21 * periph
  d/dt(periph)  = -(CL/V) * periph
[covariates]
  WT continuous
[scaling]
  obs_scale = V
[error_model]
  DV ~ proportional(PROP_ERR)
[fit_options]
  ode_reltol = 1e-10
  ode_abstol = 1e-12
"#;

#[test]
fn probe_bolus_lag_crossing_with_concurrent_infusion() {
    let model = parse_model_string(PROBE_TWO_CMT_LAG_INF_CROSS).expect("parse");
    let mut subject = bolus_subject(&[0.0, 0.5, 1.5, 2.0, 4.0, 8.0]);
    // dose 0: infusion into `periph` (cmt 2), t=0, amt 400 @ rate 100 -> window [0, 4].
    //         `ALAG1` only, so cmt 2 falls back to an all-zero PK_IDX_LAGTIME: this
    //         forcing is FROZEN in absolute time, it does NOT co-move with the bolus lag.
    // dose 1: bolus into `central` (cmt 1), t=1, ALAG1 -> arrival ~1.751, inside the
    //         segment the t=2 record (WT 140) governs.
    subject.doses = vec![
        DoseEvent::new(0.0, 400.0, 2, 100.0, false, 0.0),
        DoseEvent::new(1.0, 100.0, 1, 0.0, false, 0.0),
    ];
    let wt = |w: f64| HashMap::from([("WT".to_string(), w)]);
    subject.dose_covariates = vec![wt(70.0), wt(70.0)];
    subject.obs_covariates = vec![
        wt(70.0), wt(70.0), wt(70.0), wt(140.0), wt(150.0), wt(160.0),
    ];
    subject.covariates.insert("WT".to_string(), 70.0);
    assert!(subject.has_tv_covariates());
    assert!(ode_tvcov_supported(&model, &subject), "must be analytic");
    let theta = vec![10.0, 50.0, 0.75, 0.7, 0.4];
    let eta = vec![0.1, -0.05, 0.07];
    check_vs_production(&model, &subject, &theta, &eta);          // passes on head
    check_hessian_vs_production_fd(&model, &subject, &theta, &eta); // fails on head
}
```

```
cargo test --lib --no-default-features --features ci probe_bolus_lag
```
</details>

**Why the one-line fix is wrong.** Routing `g±` through `boundary_velocity` (state `u_minus`/`u`, params `params`/`post_params`, preps `prepared_forcings`/`prep_for(post_params)`, anchors `pre_anchor`/`t_event`) makes the probe green — and reddens `ode_provider_bolus_concurrent_infusion_lagtime_hessian_matches_fd`, whose comment states the governing rule:

> the forcing is continuous across the bolus event and **does not shift with the bolus lag**, so adding it to the event-time saltation makes `d²f/dη_LAG²` diverge from FD here.

Both fixtures are correct. They differ in whether the concurrent forcing **co-moves with the boundary**: in the existing test both doses ride the same bare `LAGTIME`; in my probe the infusion sits on a lag-free compartment. So the real fix is a per-forcing test — a forcing whose window start carries the *same* lag jet as this boundary is co-moving and stays out of `v±`; one that is frozen in absolute time goes in. Everything needed is already at hand in the loop (`lag_val(k)`, `dose_lag_slot[k]`, `dose_lagtimes_dual`).

---

### 2. Mirror of #1 — the new rate-on branch feeds **co-moving** forcings into `v⁻`/`v⁺`

`src/sens/ode_provider.rs:5170`–`5186`

Same rule, opposite error. `boundary_velocity` adds every infusion/zero-order window active at the instant to both sides. Two lagged infusions sharing one bare `LAGTIME`, one spanning the other's rate-on, with a covariate change across it: the sibling co-moves, cancels at first order, and contributes a spurious `½(J⁻−J⁺)·f` to `coef2`. Not measured — filed as the symmetric case so whatever predicate closes #1 covers both sites.

---

### 3. New rate-on anchors the pre side at TAD = 0

`src/sens/ode_provider.rs:5174` (and `5196`, the shared `anchor` into `general_rate_off_saltation`)

`v_minus` is evaluated with `last_dose = t_event`, so a `TAD`-referencing RHS sees TAD = 0 on the pre side. Production integrates the segment *ending* at the arrival with `last_dose_eff` = the **previous** dose's arrival (`predictions.rs`, `last_dose_eff_ed` filtered at `cur_t`). This is a first-order term — `(v⁻−v⁺)·δlag` — so it is a gradient error, not curvature.

The sibling bolus branch 40 lines below gets it right:

```rust
let pre_anchor = if last_dose_eff.is_finite() { last_dose_eff } else { t_event };
```

and `K_INF_END` passes `last_dose_eff`. Only this new site uses one anchor for both sides. Note `general_rate_off_saltation` takes a single `anchor` for all three `jdotg_value` calls, so a full fix needs a pre/post anchor pair there (`jg_minus` on the pre anchor, `jg_plus`/`jg_cross` on `t_event`).

**Not measurable end-to-end today**: #1070 dominates. A TAD-RHS probe on this geometry misses 13 % at an observation that sits *before* the boundary, so the boundary term is unobservable until #1070 lands. Patching only the anchor changed the result by exactly 0.

---

### 4. `pk_snapshot_equal` is value-only; IOV κ jets slip through it

`src/sens/ode_provider.rs:5147`, predicate at `:3578`

```rust
a.iter().zip(b.iter()).all(|(x, y)| x.val() == y.val())
```

Its doc-comment is explicit that the "values ⇒ jets" argument is the **non-IOV** deterministic PK map. Under IOV, `seed_iov_events` seeds `pk_at_dose[d]` per occasion group, so two records in different groups with identical covariates and `κ̂ = 0` compare **equal with different jets** — and `κ̂ = 0` is the inner-BFGS cold start and the first outer iteration.

The Open-questions section names this hazard as the reason not to add the check to the bolus path, then uses it on the infusion path. The reason given there ("the two branches are genuinely different amounts of work") is a cost argument; it does not answer the correctness objection, which applies here verbatim. Effect is second-order only (`δlag` has value 0, so the jet-only field difference lands in the `Dual2` block, not the `Dual1` inner gradient) — but it is the same class of silent-wrong-number the PR is fixing.

Pre-existing at `K_INF_END`/`K_ZO_END` (#653). Worth one issue covering all three sites; a jet-aware comparison, or gating on `n_stacked > n_eta`, closes it.

---

### 5. The lookahead is now the **fifth** copy, and one copy is dead

`src/sens/ode_provider.rs:4888` (new) vs `:5000` (`onset_params`, #880), `:5385` (`K_ROUTE_ONSET`), `:5539` (`K_INF_END`), `:5678` (`K_ZO_END`)

`post_params` is hoisted to the top of `K_DOSE`, but the input-rate arm still computes its own `onset_params` 90 lines later. So a lagtime + absorption model — the *common* lagtime case — pays two forward scans per dose and discards one.

They already disagree. `post_params` skips a co-timed `K_ROUTE_ONSET`; `onset_params` does not, and falls through its `_` arm to the dose row instead. Reachable with a route-lagged forcing whose `route_lag` resolves to 0.

The five copies vary along two axes — skip set, and `_`-arm fallback (`&pk_at_dose[idx]` vs `last_params`) — which is exactly the shape that rots. One helper:

```rust
fn post_boundary_params<'a, T>(
    tl: &[(f64, u8, usize)], p: usize, t_event: f64, fallback: &'a [T], …
) -> &'a [T]
```

and let the input-rate arm read the hoisted value. The skip of `K_ROUTE_ONSET` is correct for all five (a route onset carries no params of its own and opens a zero-length segment), so unifying is a strict improvement rather than a behaviour bet.

---

### 6. Co-timed **record** events are not skipped

`src/sens/ode_provider.rs:4895`

The skip set covers `K_INF_END`/`K_ZO_END`/`K_ROUTE_ONSET` at the same instant but not a co-timed `K_OBS`/`K_PKONLY`. An arrival landing exactly on a record returns *that* record's snapshot, although the segment it opens is zero-length and the field governing the flow after the arrival is the **next** event's.

#1068 describes this cell as "the walk returns the one-sided derivative of the branch its own event ordering uses". The value it actually returns is neither one-sided limit. Worth a sentence in #1068 so whoever picks it up knows the current behaviour is a third thing, not a defensible one-sided choice.

---

### 7. `post_params` computed for doses that never read it

`src/sens/ode_provider.rs:4888`

Runs for every `K_DOSE` under `has_lagtime`, including (a) doses taking the input-rate arm, which ignores it (see #5), and (b) doses whose own resolved lag is zero — a model declaring only `ALAG2` scans on every cmt-1 dose whose `lag_slot` falls back to an all-zero `PK_IDX_LAGTIME`, where the saltation multiplies by a zero `dlag`. The `!has_lagtime` arm is dead for the same reason: both consumers sit inside `has_lagtime`. Folding #5 removes most of this.

---

### 8. `post_d1` allocated even when both sides share a snapshot

`src/sens/ode_provider.rs:5292`–`5297`

In the no-crossing case `post_params` **is** `params`, so `post_d1` is a byte-identical copy of `pre_d1` — one extra heap `Vec` per dose per provider evaluation, on the inner-EBE hot path, in a module that elsewhere goes out of its way to avoid exactly this (`let saveat = [t_event]`, #449 review #14). A `std::ptr::eq` guard reusing `pre_d1` covers it.

---

### 9. Test coverage: the two-sided branches are never exercised **with a forcing present**

All seven new fixtures run with an empty forcing set — no concurrent infusion, no zero-order window, no built-in `input_rate`, no IOV. `ONECPT_LAG_INIT_CROSS_ODE` has no absorption block and the infusion fixture is the subject's only dose, so `prep_post` is always empty and `boundary_velocity` reduces to `eval_rhs_anchored`. That is precisely why a green suite sits next to #1.

Three fixtures worth adding: (a) the probe above; (b) the same geometry with a built-in `input_rate` on another compartment, which puts a *field-dependent* forcing into `v±` and makes the difference first-order rather than curvature-only; (c) IOV × lagtime × TV-cov crossing at `κ = 0`, to guard #4.

---

### 10. NONMEM anchor — **run it, it lands**

CLAUDE.md asks for a NONMEM comparison for any behaviour producing numerical results. The PR declines on the grounds that predictions are bit-identical and only derivatives move. The first half is true; the conclusion is not, and the reason is worth pinning down because it also raises the stakes of the fix.

`src/estimation/inner_optimizer.rs:1097` builds the FOCEI `h_matrix` — the `∂f/∂η` inside `log|H'ΩH + R|` — from **`subject_eta_jacobian`, the analytic provider**, whenever `analytic_inner_grad_supported`. `compute_jacobian_fd` is only the fallback. So a wrong `∂f/∂η` lands directly in the reported **objective function value**, not merely in a search direction. `#OBJV` is therefore a valid anchor, and a sharp one.

Ran it. `MAXEVAL=0 POSTHOC`, no convergence loop, θ fixed at the values ferx starts from — one 0.05 s NONMEM evaluation:

| | OFV | vs NONMEM |
|---|---|---|
| NONMEM 7.6.0 `#OBJV` | `-302.91470812769074` | — |
| ferx @ `2603d765` (this PR) | `-302.914703` | **5e-6** |
| ferx @ `95ac28e3` (base) | `-310.431699` | **7.52** |

ETA3 (the lagtime EBE), all eight subjects:

| ID | NONMEM | PR head | base | base − NM |
|----|--------|---------|------|-----------|
| 1 |  0.279085 |  0.279085 |  0.278995 | −9.0e−05 |
| 2 | −0.330314 | −0.330314 | −0.321935 | **+8.4e−03** |
| 3 |  0.132899 |  0.132899 |  0.132849 | −5.0e−05 |
| 4 | −0.286676 | −0.286676 | −0.284953 | +1.7e−03 |
| 5 |  0.202363 |  0.202363 |  0.202173 | −1.9e−04 |
| 6 | −0.123614 | −0.123614 | −0.121924 | +1.7e−03 |
| 7 |  0.294623 |  0.294623 |  0.294581 | −4.2e−05 |
| 8 | −0.073444 | −0.073444 | −0.071413 | +2.0e−03 |

This PR reproduces NONMEM to every printed digit, on all three etas, on all eight subjects. `ETA1`/`ETA2` match equally (omitted for width). It is the standard ferx anchor shape — `outer_maxiter: 0`, `|ofv − NM_OBJV| < 0.5` — and the base commit fails that assertion by 15×.

**The dataset geometry.** Oral 1-cpt `depot → central`, ADVAN13/`$DES`, dose into `depot` at `t = 0` on a `WT = 70` row, first record at `t = 1` carrying `WT = 140`; `ALAG1 = TVP·exp(ETA3)` with `TVP = 0.7`, `ω²_lag = 0.04`, so all eight arrivals fall in `(0.52, 0.95)` — strictly inside the crossed interval, and all on a *first* dose, so `g(x⁻) = 0` and **only** the post-side snapshot is under test. That is #1060 in isolation.

One deliberate choice worth calling out: `WT` sits on `KA` as well as on `CL`. With a plain `KA` the depot compartment's own velocity carries no covariate, the post-arrival field does not jump, and the defect does not bite — which is exactly what the new test file already says about its own oral variant. Putting the covariate where the dose lands is what makes the cell load-bearing.

<details>
<summary><code>model.ctl</code></summary>

```
$PROBLEM #1060 anchor: ODE + time-varying covariate + IIV lagtime, arrival crossing
$INPUT ID TIME DV AMT EVID CMT MDV WT
$DATA data.csv IGNORE=@

$SUBROUTINE ADVAN13 TOL=9

$MODEL
  COMP=(DEPOT,DEFDOSE)
  COMP=(CENTRAL,DEFOBS)

$PK
  CL    = THETA(1)*(WT/70)**THETA(4)*EXP(ETA(1))
  V     = THETA(2)*EXP(ETA(2))
  KA    = THETA(3)*(WT/70)**THETA(5)
  ALAG1 = THETA(6)*EXP(ETA(3))
  S2    = V

$DES
  DADT(1) = -KA*A(1)
  DADT(2) =  KA*A(1) - (CL/V)*A(2)

$ERROR
  IPRED = F
  Y     = IPRED*(1+EPS(1))

$THETA
  (0, 10.0)  ; 1 TVCL
  (0, 50.0)  ; 2 TVV
  (0, 1.0)   ; 3 TVKA
  (0, 0.75)  ; 4 WTEXP_CL
  (0, 0.75)  ; 5 WTEXP_KA
  (0, 0.7)   ; 6 TVP lagtime

$OMEGA
  0.09  ; ETA_CL
  0.09  ; ETA_V
  0.04  ; ETA_LAG

$SIGMA
  0.0025 ; PROP variance (sd 0.05)

$ESTIMATION METHOD=1 INTERACTION MAXEVAL=0 POSTHOC PRINT=1 NOABORT
$TABLE ID TIME WT DV IPRED CL V KA ALAG1 ETA1 ETA2 ETA3
       ONEHEADER NOPRINT FILE=ebe.tab
```

The matching `model.ferx` is the same six θ / three ω / one σ with `ode(obs_cmt=central, states=[depot, central])`, `obs_scale = V`, `maxiter = 0`, `covariance = false`. Eight subjects × eight records; DV simulated from the same structural model under the end-of-interval convention, no RNG dependence beyond a fixed seed. Run folder (ctl, data, ferx model, `.lst`/`.ext`/`.phi`/`ebe.tab`, RUN.md) is ready to drop in as `nonmem_anchor/tvcov_lag_saltation/` + `nonmem_anchor/results/` alongside `dose_attr_double_use_{A,B}.ctl` — say the word and I'll push it.

**Consequence for the PR description.** "Values are unchanged and bit-exact; only derivatives move" is right about `IPRED` and wrong about the OFV: base misses NONMEM's objective by 7.52 units on this dataset. Worth correcting there and in the changelog entry, because it changes what a user should expect — an affected fit was reporting a wrong OFV, not merely walking a wrong path to the same one.

Note the finding-#1 residual is a *second*-order term (`d2f_deta2` → `sens_outer_gradient.rs`), so it does not show up in this MAXEVAL=0 OFV; it needs a gradient-consuming outer method (`gn`, `gn_hybrid`, `trust_region`, SLSQP/L-BFGS/MMA).

---

### 11. `CMT=0` infusion: both velocities off by `Δr`

`src/sens/ode_provider.rs:5178`

`v_minus[cmt_idx] -= inf_eff[idx].0` runs unconditionally, but `boundary_velocity` only *added* that rate under `d.cmt_raw() >= 1`. For a `CMT=0` infusion the subtraction removes a rate that was never added: the **difference** stays right (so the gradient survives) but both `J⁻·v⁻` and `J⁺·v⁺` are evaluated at velocities off by `Δr`, corrupting `coef2`. The surrounding comments say this is unreachable from a validated call but explicitly *not* rejected by the datareader. Mirroring `boundary_velocity`'s own `cmt_raw() >= 1` guard on the subtraction is a one-liner.

---

### Reviewer hints — both check out

**The `_ => &pk_at_dose[idx]` fallback is correct.** `last_params = &pk_at_dose[idx]` is assigned at the end of the branch, before any later event is processed, so every consumer of the fallback sees the value the comment claims. The only kinds that can sort at or after `K_DOSE` and reach `_` are `K_RESET` (sorts strictly earlier, kind 0) and later-time `K_INF_END`/`K_ZO_END`/`K_ROUTE_ONSET` — all of which take `last_params` at the head of the loop. Right, and right for the stated reason.

**Co-timed multi-dose telescopes.** Worth recording, because it looks like a double-count and isn't. Two lagged doses arriving together: dose *k*'s lookahead lands on its **sibling's** `pk_at_dose`, and the sibling's own pre side is that same snapshot, so the field terms cancel pairwise:

```
[g(x,p₀) − g(x+Δ₁,p₁)] + [g(x+Δ₁,p₁) − g(x+Δ₁+Δ₂,p_post)]  =  g(x,p₀) − g(x+Δ₁+Δ₂,p_post)
```

Holds for bolus+bolus and bolus+infusion in either order, and holds even when the two dose rows carry different covariates. No cohorting needed here, unlike `K_ZO_END`.

**The asymmetry in Open questions — keep it, different reason.** Don't add `pk_snapshot_equal` to the bolus path. But the κ-jet objection you raised against it is *also* true on the infusion path where you did use it (#4), so the thing to revisit is that use, not the omission.

---

### Repro environment

```
git fetch origin worktree-fix-1060-tvcov-lag-saltation
cargo test --lib --no-default-features --features ci lagtime_tvcov   # 10 passed on head
```

Base `95ac28e3` and head `2603d765` both built and run. On base the probe in #1 fails at the *gradient* (`∂f/∂η_lag` `0.33275` vs `0.69868`) — the #1060 bug itself — which is why the residual second-order gap was invisible until this PR landed the first-order fix.

NONMEM 7.6.0 via `nonmemdocker:V0.1`, `nmfe76`, run folder `~/NONMEMruns/tvcov-lag-saltation/`.

TeunP:
## Follow-up — I ran the NONMEM anchor for the **multi-dose** case, and it does not match

My review comment above anchored the single-dose geometry: NONMEM `#OBJV -302.91470813` vs this PR `-302.914703`, EBEs to every printed digit. That cell is clean.

It is also the cell that cannot see the pre side. Every dose in it is a *first* dose, so `g(x⁻) = 0` and only the post-side snapshot is under test. So I built the multi-dose twin — second dose at `t = 6` whose lagged arrival lands at `≈ 6.7` with residual drug present, dose row carrying `WT = 150` and the next record (`t = 7`) carrying `WT = 75`.

It fails, and **not on the gradient — on `PRED`.**

### Measurement

Four datasets, identical model, identical `$EST METHOD=1 INTERACTION MAXEVAL=0 POSTHOC`. Subject 1, `PRED` at `t = 7`:

| dataset | ferx `2603d765` | NONMEM 7.6.0 |
|---|---|---|
| **A** single dose | *(matches everywhere)* | ✓ `#OBJV` Δ = 5e-6 |
| **B** multi-dose, dose row `WT 150`, next record `WT 75` | `0.816054` | `0.846890` |
| **C** = B + an extra record at `t = 6.5` carrying `WT 75` | `0.886124` | `0.846890` ← *unchanged from B* |
| **D** = B but dose row `WT 75` (= next record) | `0.896650` | `0.896650` ✓ |

`#OBJV`, same four:

| | ferx | NONMEM | Δ |
|---|---|---|---|
| A single dose | `-302.914703` | `-302.91470813` | **5e-6** |
| B multi-dose | `-474.660106` | `-459.77259187` | **14.89** |
| D control | `-415.644111` | `-415.64410902` | **2e-6** |

**D is the isolation.** It differs from B in exactly one number — the `WT` on the dose row — and it flips ferx from 14.9 OFV out to an exact match. So the disagreement is entirely about which covariate snapshot governs the interval `[dose record, lagged arrival]`.

**C pins the direction.** Inserting a record between the dose row and the arrival moved ferx by `+0.070` and moved NONMEM by **nothing**. NONMEM was already using the later record's `$PK` across that whole stretch.

### What the two engines actually do

* **NONMEM** breaks on **record** times only. `$PK` is evaluated per record; the advance `[5, 6]` runs under the `t = 6` dose record's `$PK`, and the lagged dose is injected *inside* the next record's advance, under the `t = 7` record's `$PK`. The lagged arrival is an event, not a PK breakpoint.
* **ferx** does not break at the dose record's own time at all. `Kind::Dose` sits at `d.time + lag` (`ode/predictions.rs`, and the same in `integrate_tvcov_g`'s timeline: `tl.push((d.time + lag_val(k), K_DOSE, k))`), so the dose row's snapshot is stretched from the previous record all the way to `t_dose + lag`.

So ferx runs `[t_dose, t_dose + lag]` on the **dose row's** covariates where NONMEM runs it on the **next record's**. The two agree only when those two records carry the same covariate values — which is why every existing fixture passes, and why single-dose passes even when they differ (the state is zero over that interval, so it cannot matter).

The fix is one timeline break: push the dose *record* time as its own event, `[prev, t_dose]` under the dose row's snapshot and `[t_dose, arrival]` under the next record's.

### What this means for this PR

Nothing here says the PR is wrong. It makes the analytic walk faithfully reproduce ferx production, and against production FD it does. The single-dose anchor confirms production itself matches NONMEM wherever the pre-arrival interval carries no drug.

But it does reframe the reviewer hint:

> The pre-side uses the *dose row's* snapshot rather than the previous record's — that is not an oversight, it is what production does (`predictions.rs:4585`, `Kind::Dose => pk_at_dose[idx]`, for the segment *ending* at the event).

That is an accurate description of ferx production, and production is the right thing for the gradient to match. It is not what NONMEM does. Under NONMEM's convention there is **no field jump at the arrival at all** unless a record falls strictly between the dose row and the arrival — so the two-sided boundary the PR builds is modelling a discontinuity whose position is itself the thing that diverges.

Concretely, the `docs/estimation/foce.qmd` sentence this PR adds —

> the two sides of the boundary are evaluated on their own snapshots — the segment ending at the arrival on the dose record's, the segment it opens on the next record's, matching the end-of-interval convention the predictor uses (#1060)

— is true of the predictor and false of NONMEM. Worth softening to name the predictor rather than implying the convention is NONMEM's, until the production side is settled.

Suggest this lands as its own issue against `ode/predictions.rs` rather than growing #1071's scope; the fix there will then want the saltation revisited, because moving the break changes where (and whether) the field jumps.

### Reproducing

Run folders, all four: `~/NONMEMruns/tvcov-lag-{saltation,saltation-multidose,md-control,md-shrink}/`, each with `model.ctl`, `data.csv`, the matching `model.ferx`, and `.lst`/`.ext`/`.phi`/`ebe.tab`. Model is oral 1-cpt `depot → central`, ADVAN13/`$DES`, `ALAG1 = TVP·exp(ETA3)`, `WT` allometric on **both** `CL` and `KA` — the `KA` term matters, without a covariate on the dose compartment's own velocity the post-arrival field does not jump and neither this nor #1060 reproduces.

ferx side: `run_model_with_data(model.ferx, data.csv)` with `maxiter = 0`, `covariance = false`, analytic gradient.

TeunP:
**Correction to my follow-up above — disregard the `foce.qmd` suggestion.**

I wrote that the new sentence

> the two sides of the boundary are evaluated on their own snapshots — the segment ending at the arrival on the dose record's, the segment it opens on the next record's, matching the end-of-interval convention the predictor uses (#1060)

"is true of the predictor and false of NONMEM. Worth softening to name the predictor rather than implying the convention is NONMEM's."

It already names the predictor. I misread it as a NONMEM-equivalence claim. The sentence is accurate as written — nothing to change here, and no docs edit is needed in this PR.

The real documentation gap is a different one and does not belong in this PR: nothing tells a user that a **multi-dose** model with an estimated lagtime and time-varying covariates diverges from NONMEM in *value* (#1073). That is a known-limitation note, and it should land with the fix rather than as a hedge attached to a correct sentence — a caveat added now would be rewritten anyway, since fixing #1073 collapses the two-sided treatment this sentence describes. Noted on #1073 instead.

Everything else in the two reviews stands.

TeunP:
## Response to the `xhigh` review — all eleven findings, plus two the response itself turned up

Five commits on top of `fce3e39e`. Every claim below is measured; where I disagree with the review I say so and give the number that settles it.

### The blocking one (#1), and why the proposed predicate is wrong

Reproduced exactly: your probe gives `∂²f/∂η_LAG²` `0.87941` vs `0.89639`, 1.9 %, gradient clean. Your algebra is right — `coef2` drops `½(J⁻−J⁺)·f`, and it cancelled only while `J⁻ = J⁺`.

The proposed fix rule is not. You concluded from `ode_provider_bolus_concurrent_infusion_lagtime_hessian_matches_fd` that the discriminator is whether the forcing **co-moves with the boundary**, and that a co-moving one stays out of `v±`. That is one hypothesis consistent with the two fixtures you had; here is the third geometry that separates it from the alternative — a **straddling co-moving** forcing, both doses on one bare `LAGTIME`, window `[lag, 4+lag]`, arrival `1+lag` strictly inside:

| geometry | `∂²f/∂η_LAG²` at `fce3e39e` | production FD | verdict |
|---|---|---|---|
| straddling **frozen** (your probe) | `0.87941` | `0.89639` | must be IN (1.9 %) |
| straddling **co-moving** (new) | `-0.65279` | `-0.48247` | must be **IN** (35 %) |

Under "co-moving ⇒ out" the second row would have to be exact already. It is 35 % out. So the rule is not frozen-vs-co-moving; it is **straddling vs toggling**:

- a forcing that **straddles** the instant is grid-constant on both sides and belongs in `v±`, whether or not it co-moves;
- a forcing that **toggles** at the instant injects its own saltation at that same event, and the two compose to the exact one-sided form only if this one leaves it out.

That is derivable rather than fitted. Writing `S` for the straddling set and `r` for a co-starting sibling's rate, the arrival contributes `(g⁻−g⁺)δ + [½J⁻(g⁻+S) + ½J⁺(g⁺+S) − J⁺(g⁻+S)]δ²` and the sibling `−rδ + ½J⁺r·δ²`, summing to exactly `(v⁻−v⁺)δ + (½J⁻v⁻ + ½J⁺v⁺ − J⁺v⁻)δ²` for the true one-sided `v⁻ = g⁻+S`, `v⁺ = g⁺+S+r`. Counting `r` on both sides double-counts it at first order and adds a spurious `½(J⁻−J⁺)r` at second — which is what your existing MM fixture was detecting. Both fixtures were right; the reading of them was the thing to fix.

Implemented as a `strict` flag on `boundary_velocity`. The rate-off sites keep inclusive membership (their own `(v⁻−v⁺)` bookkeeping depends on it); the `K_DOSE` boundaries take strict.

### #2 — same rule, and it is not the co-moving one either

The rate-on now uses strict membership and puts its own rate on the post side alone, rather than counting it on both and subtracting from the pre side. Same reasoning, mirrored. A co-moving sibling that merely spans the rate-on stays in, for the reason above.

### #11 — folded into #2

The unconditional `v_minus[cmt_idx] -= inf_eff[idx].0` is gone: with strict membership this infusion is out of both sides and added back to the post side, so there is nothing to subtract. The whole rate-on saltation is now skipped for `cmt_raw() < 1`, matching the walk, which delivers no rate for such a dose.

Reading that through turned up a bigger, separate thing: **production `active_infusions` has no compartment guard at all**, so a `CMT=0` infusion is delivered by `compute_predictions_with_tv` into state 0 and skipped by the analytic walk. That is a **value** divergence, not a derivative one. Filed as #1077.

### #4 — reproduced, and worse than flagged: the element comes out with the wrong sign

You called it second-order-only and worth an issue. It is second-order-only, and it is a sign flip. Fixture: `κ` on `CL`, `LAGTIME` with IIV, a prior bolus leaving residual drug, a lagged infusion arriving between the last occasion-1 record and the first occasion-2 record, `κ̂ = 0`.

| element (t = 8) | analytic | double-FD of `predict_iov` | rel |
|---|---|---|---|
| `∂²f/∂η_LAG∂κ_g0` | `+0.05207426` | `−0.01477625` | **1.28, sign flip** |
| `∂²f/∂η_LAG∂κ_g1` | `−0.12254723` | `−0.05569672` | 0.55 |
| everything else | — | — | ≤ 3e-7 |

Two things about that geometry are load-bearing and the first cost me a cycle: without the prior bolus the state is zero at the arrival, the whole term vanishes, and the fixture passes vacuously — which is exactly what my first version did. And the occasion boundary sits on an observation record, because `predict_iov` is not a valid oracle at a boundary without one (#931).

You are right that the cost argument in Open questions does not answer the correctness objection. The site is **this PR's**, so this was a regression I introduced, not a pre-existing exposure I widened — and it no longer consults `pk_snapshot_equal` at all. Four extra RHS point-evaluations per lagged infusion, against the hundreds a segment spends integrating between two events. The two pre-existing rate-off uses stay with #1072, which now carries this measurement and a concrete shape for the jet-aware predicate.

### #3 — right by construction, and I can show it is unmeasurable

Confirmed against `predictions.rs` directly (`last_dose_eff_ed`, filtered at `cur_t + 1e-12`), not by analogy with the bolus branch. `general_rate_off_saltation` now takes a pre/post anchor pair; the two rate-off sites pass the same value twice, so they are byte-identical.

Your "changed the result by exactly 0" reading is right about the conclusion and understates the cause. I built the TAD probe: 1-cpt with `-(CL/V)*central*(1 + 0.3*TAD)`, `LAGTIME` with IIV, `WT` on `CL` time-varying, records after the first arrival. `ode_tvcov_supported` returns **true**.

| t | `∂f/∂η_CL` | `∂f/∂η_V` | `∂f/∂η_LAG` rel |
|---|---|---|---|
| 0.8 | 7e-9 | 1e-10 | **1.5e-2** |
| 1.5 | 1e-9 | 3e-11 | **1.9e-1** |
| 4.0 | 8e-11 | 6e-10 | **3.3e-1** |
| 8.0 | 1e-11 | 8e-10 | **6.2e-1** |

Every non-lag axis exact to 1e-9; the lag axis wrong at **every** observation including t = 0.8, which precedes both boundaries and can therefore only be the segment integration itself. #1070 is not latent — it is a 1.5–64 % silently-wrong gradient on a model the gate admits, and it swamps anything the anchor could show. The anchor ships with a comment saying exactly that; #1070 now carries the table and a suggestion to route the combination to FD in the interim.

### #5, #6, #7, #8 — done, with one correction you will want to check

The five lookaheads are one `post_snapshot` helper. It returns the first event **strictly** later than the boundary, which is what the walk itself integrates against (`if t_event > cur_t` — every co-timed event opens a zero-length segment), so the kind-by-kind skip sets fall out and #6 comes with it. That also makes the #1068 sentence in `foce.qmd` true for the first time: the walk now really does return the one-sided derivative of the branch its own event ordering uses, rather than the third thing you identified. Re-measured on #1068's geometry, the 1 % is unchanged in size and has flipped to the other side of FD, which is what a branch switch should look like. Noted there.

**The correction.** "Strictly later" is wrong for one case, and it is the case your own reviewer hint documented. A co-timed sibling **dose** must still be returned, because the telescoping you worked out depends on dose *k*'s post side being dose *k+1*'s pre side. Skipping it leaves an uncancelled `g(x+Δ₁,p₁) − g(x+Δ₁,p_post)`: **25 %** on the gradient (`1.1222` vs `0.8954`) of a fixture with two co-timed lagged doses arriving past a covariate change. Green on the pre-review kernel, red with the over-general skip, green with the exception — so the fixture pins the exception, not the general rule. No existing co-timed-dose fixture carries time-varying covariates, where the leftover is identically zero, which is why 3290 lib tests did not catch it. Caught in self-review, and the reason your hint was worth writing down.

#7: the input-rate arm reads the shared lookahead instead of its own; the binding is now a closure so a lagtime-free dose pays no scan and there is no dead `!has_lagtime` arm. #8: the post-side `Dual1` params are shared when the two snapshots agree by value, and the prepared forcings when they are the same object (`prepare_dual` reads the full duals, so value equality would not license that one).

### #9 — six fixtures

Straddling frozen infusion; straddling co-moving infusion; straddling built-in `input_rate` with the covariate on `KA`; a covariate-free control for it; IOV × lagtime across an occasion boundary; and the co-timed-doses telescoping guard. Four of the six are red on `fce3e39e`; the control is green on both sides by design.

### #10 — you were right, and the changelog said the wrong thing

Anchor commit taken as-is. The `[Unreleased]` entry claimed "predictions were unaffected", which is true of `IPRED` and misleading about the OFV for exactly the reason you give (`inner_optimizer.rs` builds `h_matrix` from the analytic `∂f/∂η`). Rewritten to say the objective missed NONMEM by 7.52 units before the fix and 5e-6 after.

### What I did not fix, and why

The input-rate fixture is exact on the gradient and **not** on the second order — 6.4 % on the worst `∂²f/∂η_LAG²` element. That is a distinct term, not a leftover of #1: for an explicitly time-dependent field `ẍ = ∂f/∂t + J·f`, and a pointwise `R_in(tad)` is exactly that, so the derivation carries a `½(∂f⁻/∂t − ∂f⁺/∂t)·δ²` term alongside the Jacobian ones. It is a **difference**, so it is identically zero whenever the two sides share a snapshot — which is every fixture that existed before this PR. The covariate-free control isolates it: same forcing, same geometry, `KA = TVKA`, both orders exact.

Fixing it needs `∂R_in/∂tad` at arbitrary `tad` (only `rate_dtad_at_zero` exists) and applies to all four saltation sites, so it is #1075 rather than half-landed here. The fixture asserts a bound instead of exactness so it cannot grow silently, and the acceptance test for #1075 is swapping that bound for `check_hessian_vs_production_fd`.

Also measured and filed, both pre-existing and bit-identical either side of this PR:

- **#1076** — two moving boundaries colliding at one instant each inject the whole covariate field jump, so it is counted twice: **5.9 % on the gradient** for an infusion whose window ends exactly at the next lagged dose's arrival (`dur == t_bolus − t_inf`, an ordinary regimen shape, not a coincidence of continuous values).
- **#1077** — the `CMT=0` infusion value divergence above.

Your #1073 stands untouched; nothing here changes what production does.

### Verification

`cargo test --lib --no-default-features --features ci`: 3290 passed, 0 failed. Full `--features ci` suite across all binaries: green. `foce.qmd` now names the straddling-forcing fix and enumerates the three remaining approximations with their issue numbers, instead of implying the cell is exact.

Issue #1073, opened 2026-08-25
bug(ode): a lagged dose record's covariate snapshot is stretched to the lagged arrival, diverging from NONMEM by 14.9 OFV


TeunP:
## Summary

`ode/predictions.rs` never breaks the event timeline at a **dose record's own time**. `Kind::Dose` is pushed at `d.time + lag`, so under an estimated lagtime the dose row's covariate snapshot governs the flow from the *previous* record all the way to the lagged arrival. NONMEM breaks `$PK` on record times and injects the lagged dose inside the **next** record's advance.

The two therefore disagree on which disposition parameters govern `[t_dose, t_dose + lag]` whenever a lagged dose record carries different time-varying covariates from the record after it — and there is residual drug at the arrival.

Measured at **14.89 OFV units**, on `PRED`, not on the gradient.

This is not a sensitivity bug. It is not #1060 / #1068 / #1069 / #1070 / #1072. Found while reviewing #1071 (which is correct as scoped — see [Relationship to #1071](#relationship-to-1071)).

## Measurement

Anchor set committed on the #1071 branch: `nonmem_anchor/tvcov_lag_saltation{,_multidose,_md_control,_md_shrink}.{ctl,csv}` + `nonmem_anchor/results/`. Oral 1-cpt `depot → central`, `ADVAN13 TOL=9`, `ALAG1 = TVP·exp(ETA3)`, `WT` allometric on **both** `CL` and `KA`, `$EST METHOD=1 INTERACTION MAXEVAL=0 POSTHOC` — one evaluation at fixed θ on each side, no convergence loop.

| | ferx `2603d765` | NONMEM 7.6.0 `#OBJV` | Δ |
|---|---|---|---|
| **A** single dose | `-302.914703` | `-302.91470812769074` | 5e-6 ✓ |
| **D** control | `-415.644111` | `-415.64410902342507` | 2e-6 ✓ |
| **B** multi-dose | `-474.660106` | `-459.77259187257408` | **14.89** ✗ |

Subject 1, `PRED` at `t = 7`:

| | ferx | NONMEM |
|---|---|---|
| **B** multi-dose, dose row `WT 150`, next record `WT 75` | `0.816054` | `0.846890` |
| **C** = B + an extra record at `t = 6.5` carrying `WT 75` | `0.886124` | `0.846890` ← *unchanged from B* |
| **D** = B but dose row `WT 75` (= next record) | `0.896650` | `0.896650` |

[...]

PR #1074, opened 2026-08-25
docs(claude): require every oracle to exercise every side of the object under test

TeunP:
## Why

#1060 shipped with a **green** single-dose NONMEM anchor sitting next to a **14.9-OFV** multi-dose divergence (#1073) that no fixture in the repo could see.

The reason is structural, not an oversight. Every dose in a single-dose dataset is a *first* dose, so the state is zero before the arrival, `g(x⁻) = 0`, and any error on the **incoming** side of a dose event cancels instead of showing up. The anchor was correct, tight (Δ = 5e-6), and blind to half the object it was anchoring.

The same degeneracy has bitten this repo before in other shapes, each rediscovered by measuring:

* an **oral** fixture whose dose lands in a compartment carrying no covariate, so the post-arrival field never jumps and the defect leaks a couple of percent instead of 300× (`sens/ode_provider_tests.rs` records this for #1060);
* a **rate-on** boundary with no `init(...)` baseline, so the state is zero exactly where the omitted term multiplies it (#1071's own infusion fixture needs `init` to bite);
* a fixture asserted against ferx's **own predictor**, which agrees with a wrong answer by construction whenever both paths share the convention under test — the reason #1073 needed an external reference to surface at all.
* a **cross-engine** twin, which is not automatically a second reference either: it only sees a defect that lives *downstream* of the point where the engines part. #1079's κ = 0 Form C readout was catchable by an ODE twin because each engine applies the readout itself; the per-occasion snapshot feeding `ALAG`/`F`/`D{n}`/`R{n}` is not, because it is built in `predict_iov` *before* the `ode_spec` branch and handed to both arms. Added after review of #1090 surfaced it.

`CLAUDE.md` already requires a NONMEM comparison. It does not say the comparison has to be non-degenerate, and that is the part that failed.

## What changed

One paragraph in `CLAUDE.md`, placed with the existing NONMEM-comparison rule in the Tests section. It states the requirement (pair every dose-event anchor with a multi-dose case landing on residual drug), lists the four specific degeneracies above to check for, and records the vary-one-input-at-a-time habit that localised #1073 — the B/D pair in `nonmem_anchor/tvcov_lag_saltation*` differs by exactly one number, which is what pinned the defect to a single interval.

The rule's lead names the three oracle kinds it covers — NONMEM anchor, ODE twin, FD parity — rather than NONMEM anchors alone. Two of the four degeneracies involve no NONMEM at all, and the bolded lead is how this file gets indexed: scoped to NONMEM anchors it routed readers away from a Tier-1 `Dual2`-vs-FD test, which is exactly where #1079 hid. (The PR title still says "NONMEM anchors" and now under-describes the diff; the scope is as stated here.)

One further change in the **adjacent** `Dual2`-vs-FD paragraph, which required asserting against central finite differences of the `f64` production predictor without noting that this is the very "asserted against ferx's own predictor" degeneracy listed above — two paragraphs in contradiction with no pointer between them. It now names the limit: FD parity pins the derivative against the value path, not the value path itself. #1079 is the worked case, where `eval_readout_at_obs` seeded `group = None` precisely to match the `f64` path and parity passed against the exact derivative of the wrong function.

## Alternatives considered

**Leaving the cross-engine shape out.** It is the one a reader is most likely to get wrong, because two engines *look* like two references. Skipping it would leave the rule stating a weaker claim than the evidence supports.

**"Add a multi-dose case" as the whole rule.** Too narrow. Multi-dose is one instance of the general failure — a fixture that silently degenerates so that the thing under test is multiplied by zero. The oral-compartment and missing-`init` cases are the same failure and would not be caught by a dose-count rule.

**Putting it in the PR template instead.** The template is a per-PR checklist; this is a standing property of the anchor corpus and belongs where the NONMEM rule already lives.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

The measurement motivating the rule is in #1073 and in the anchor set on #1071's branch (`nonmem_anchor/tvcov_lag_saltation*`, four control streams, NONMEM 7.6.0).

## Performance

## Tests
- [x] No new tests needed (why: `CLAUDE.md` prose only — no code changes)

## Docs
- [x] No user-visible change

`CHANGELOG.md` N/A — internal contributor guidance, not user-facing.

## Checklist
- [x] `cargo clippy` clean (no code changed)
- [x] Functions on the `Dual2` sensitivity path stay differentiable (N/A — no code changed)

### ferx-site
- [x] N/A — no DSL syntax, `[fit_options]` key, example, or data file added

### ferx-book
- [x] N/A — no estimator, DSL feature, or fit option added

### Example execution
- [x] N/A — no code changed, no example affected


TeunP:
@roninsightrx : can you have a look - think we need single dose and multiple dose anchoring as a rule

TeunP:
Read the rule as it now stands (three commits), including the clause I added myself in `6c011678`. Two findings, one nit already fixed.

Nothing here is code, so the green `Check`/`Clippy`/`Format` jobs say the file parses as Markdown and nothing else. They are not evidence about the rule.

## 1. The bolded lead sentence under-scopes the paragraph

> **A NONMEM anchor must keep every side of the object under test non-degenerate.**

Two of the four degeneracies listed underneath are not about NONMEM anchors: "asserted against ferx's own predictor" and the cross-engine twin. Both are oracles, neither involves NONMEM.

This matters because of how the file is actually read — bolded leads are the index. A reader who files this under "NONMEM anchor rules" will not apply it to a Tier-1 `Dual2`-vs-FD test, and a Tier-1 `Dual2`-vs-FD test is precisely where #1079 hid for two releases. The rule would have caught that bug and the heading routes it away from it.

Suggest widening the lead to the general object and keeping NONMEM as the leading instance rather than the category — something like "An oracle must keep every side of the object under test non-degenerate", with the single-dose case still first because it is the most common. Not editing this one myself since it is your sentence and the framing is a judgement call.

## 2. The next paragraph prescribes the degenerate oracle this one warns about

Directly below, the `Dual2`-vs-FD rule says:

> assert it against central finite differences of the `T = f64` production predictor

That is bullet three of the new paragraph, verbatim — a fixture asserted against ferx's own predictor. And #1079 is the worked proof that it goes wrong: `eval_readout_at_obs` seeded `group = None` *precisely to match* the f64 path, so FD parity passed against the exact derivative of the wrong function.

So the file now contains two adjacent paragraphs, one warning against what the other requires, with no pointer between them. A contributor following the second in good faith writes the blind test, and the first paragraph never fires because it reads as being about NONMEM anchors (see finding 1 — these compound).

One clause in the `Dual2` paragraph closes it: FD parity pins the *derivative* against the value path, not the value path itself, so it is only an oracle for the gradient — see the non-degeneracy rule above for what it cannot see. Worth doing in this PR while the context is loaded.

## 3. Nit, self-inflicted — fixed in `dfd49e60`

My clause spelled `kappa` where the file uses Unicode Greek everywhere else (η, θ, Σ; κ appeared zero times before it).

## What I verified rather than assumed

The branch-point claim in the new clause: `dose_params` is built in `predict_iov` before the `if let Some(ref ode) = model.ode_spec` branch and both arms receive the same slice, so neither engine rebuilds it — that is what makes an analytic-vs-ODE comparison blind to an error in `combined_for(occ)`. The readout contrast holds for the stated reason: `apply_analytic_readout` and `integrate_tvcov_readout` are genuinely separate per-engine implementations.

## Otherwise

The paragraph is long for a file loaded into every session, but the content is load-bearing and I would not trim it blind. If it does need to come down, the #1060/#1073 sentence is the most compressible — it links out to the full story anyway.

PR #1086, opened 2026-08-25
fix(io): re-source the naproxen MBMA fixture to its primary publication [closes #1085]

### LINKED ISSUE #1085: license(data): naproxen MBMA fixture is a derivative of an ND-licensed work — re-source to Boucher & Bennetts Part II (CC BY-NC) and label it non-MIT
**Type:** Licensing / provenance correction (with a validation upside)
**Scope:** `data/mbma_naproxen.csv`, `data/mbma_naproxen.README.md`, `data/mbma_naproxen_source.py`, `tests/mbma_naproxen_case_study.rs`, `examples/mbma_naproxen.ferx`, `Cargo.toml`, root `LICENSE`/`README.md`

## The problem

`data/mbma_naproxen.csv` shipped in #1063 as the fixture for the MBMA case study. Its README attributes it to Bracis et al., *CPT:PSP* 2026;15:e70158, and notes the paper is CC BY-NC-ND. Both clauses are a problem for this repo, and the **ND** one is being violated right now.

The article's own permissions block, verbatim:

> This is an open access article under the terms of the http://creativecommons.org/licenses/by-nc-nd/4.0/ License, which permits use and distribution in any medium, provided the original work is properly cited, the use is **non-commercial** and **no modifications or adaptations are made**.

(Confirmed independently by Europe PMC's license field on `PMC12896392`: `cc by-nc-nd`.)

1. **ND is violated.** `mbma_naproxen_source.py` renames columns, recodes `TRT`, drops `WP` and emits a ferx-layout CSV. That is an adaptation, and ND forbids distributing adaptations at all — not merely without attribution. The derived file is committed to a public repo.
2. **NC conflicts with the repo's MIT grant.** `LICENSE` grants commercial use over the tree; that file cannot carry it.

## The finding that makes this cheap to fix

**The dataset is not Bracis's.** Their paper says so plainly:

> "The dataset [11] includes summary-level data from 18 placebo-controlled longitudinal trials of naproxen in OA using WOMAC pain scores."

Reference [11] is **Boucher M, Bennetts M. Many Flavors of Model-Based Meta-Analysis: Part II — Modeling Summary Level Longitudinal Responses. *CPT:PSP* 2018;7:288–297** (`10.1002/psp4.12299`, `PMC5980518`) — the analysis Bracis et al. explicitly replicate in MonolixSuite. Bracis's supplementary CSV is itself a transformed copy of Boucher & Bennetts' data.

Boucher & Bennetts Part II is **CC BY-NC — no ND**:

> This is an open access article under the terms of the http://creativecommons.org/licenses/by-nc/4.0/ License, which permits use, distribution and reproduction in any medium, provided the original work is properly cited and is not used for commercial purposes.

Its supplementary file `PSP4-7-288-s007.csv` (columns `STUD,NFLG,TIME,NTRT,FLARE,WP,WPSE`, 122 data rows) is the same data we ship. Verified on row 1: `WP` 5.3, `WPSE` 0.110312188, `NTRT` 217 — our committed row is `DV` 48.045462 = 5.3 / 0.110312188, `WPSE` 0.110312188, `NIND` 217.

So re-sourcing to the actual origin **removes the ND violation outright** and is more accurate scholarship besides. NC remains, and is handled by labeling rather than removal (decision below).

## Decision

**Keep the fixture in-tree, clearly labeled as not-MIT.** The reason is not convenience: re-sourcing turns it into an anchor against the *original* publication, which reports this exact model fitted in **three independent engines** — and ferx already agrees with all of them (see below). That is a stronger validation asset than the Monolix comparison alone, and it is worth carrying a per-file license exception for.

Recorded so it is not re-litigated later: a CC BY-NC file in an otherwise-MIT repo means downstream commercial users are not granted rights over *that file*. It is a test fixture, never linked into the binary or any published artifact, and the exception is explicit — but the constraint is real and this issue is where it is written down.

### 1. Re-point the derivation at the primary source

`data/mbma_naproxen_source.py` currently consumes Bracis's `naproxen_data_transformed_se.csv`, which carries a precomputed `transWP`. Boucher & Bennetts' `PSP4-7-288-s007.csv` does not, so the script computes `DV = WP / WPSE` itself — one extra line, everything else is the same relabeling (`STUD → ID`, `NTRT → NIND`, arm/treatment coding from `NFLG`).

The regenerated CSV must be **byte-identical** to the committed one. If it is not, the difference is the finding and gets resolved before anything else here lands.

### 2. Label it

- `data/mbma_naproxen.LICENSE` — CC BY-NC 4.0 text plus the attribution line, sitting next to the data.
- `data/mbma_naproxen.README.md` — rewrite the Source section: primary source Boucher & Bennetts Part II (CC BY-NC), with Bracis et al. cited as the tutorial that replicated the analysis and as the source of the Monolix comparison values, not of the data.
- Root `LICENSE` / `README.md` — a short "everything is MIT except `data/mbma_naproxen.*`, which is CC BY-NC 4.0" note. Someone auditing the tree must be able to find this without reading `data/`.
- `examples/mbma_naproxen.ferx` header and `tests/mbma_naproxen_case_study.rs` module docs — both currently name only Bracis; both cite the primary source.

### 3. Keep it out of the published crate

`Cargo.toml` has no `include`/`exclude`, so `cargo package` ships everything git-tracked — including `data/`. Add an `exclude` for `data/mbma_naproxen.*` so the crate published to crates.io is uniformly MIT. The slow-gated test that reads it does not run from a packaged crate.

### 4. ferx-r stays blocked

FeRx-NLME/ferx-r#312 mirrors the *synthetic* `emax_timecourse` example only, and explicitly holds the naproxen one. That stays true: NC data has no business in a CRAN-bound package. Cross-reference this issue there.

## The upside: a three-engine anchor we are currently not claiming

Boucher & Bennetts Part II Table 2 reports this model fitted in **BUGS, NONMEM, and R (`nlme`)**. ferx's shipped fit already reproduces the NONMEM column:

| Parameter | BUGS | NONMEM | R (nlme) | ferx (today) |
|---|---|---|---|---|
| E0 (nonflare) | 5.22 (0.28) | 5.20 (0.13) | 5.20 (0.27) | — |
| ΔE0 (flare) | 0.93 (0.33) | 0.96 (0.25) | 0.96 (0.32) | — |
| Emax_p (nonflare) | −1.14 (0.41) | −1.16 (0.24) | −1.15 (0.32) | — |
| ΔEmax_p (flare) | −0.86 (0.47) | −0.82 (0.09) | −0.82 (0.39) | — |
| ΔEmax_n | −0.79 (0.06) | −0.79 (0.09) | −0.79 (0.07) | 0.7930 (0.0646) |
| ET50_p (weeks) | 0.67 | 0.69 | 0.67 | 0.6901 |
| ET50_n (weeks) | 0.19 | 0.21 | 0.20 | 0.2132 |
| τ1 (ω E0, SD) | 0.86 | 0.62 | 0.62 | 0.6215 |
| τ2 (ω Emax, SD) | 0.71 | 0.74 | 0.74 | 0.7387 |

(Sign on ΔEmax_n is a parameterisation difference — Boucher & Bennetts write Emax as a negative effect on the pain score, the shipped model as a positive reduction. The dashed rows are simply not asserted yet, not disagreements.)

So extend `mbma_naproxen_reproduces_the_published_analysis` to pin against the **NONMEM column of Table 2** as well as the Monolix values it already uses, and assert the four currently-unasserted parameters. This is worth more than a second copy of the same claim: the Monolix numbers come from a tutorial replication, whereas Table 2 is the primary analysis, and the three engines disagreeing among themselves (τ1: BUGS 0.86 vs 0.62) tells us how tight the tolerance can honestly be.

Also worth capturing from the same paper, since the shipped model depends on it and the sentence is the reason `weight =` (#1029) exists at all:

> "As the weights were based on the observed SEs, σ was fixed to 1 in the modeling. However, in R, using the NLME function, it was not possible to fix σ to 1."

(@billdenney noted in #811 that recent `nlme` versions have since lifted this.)

## Done when

1. `data/mbma_naproxen.csv` regenerates byte-identically from `PSP4-7-288-s007.csv`, and no file in the tree is a derivative of an ND-licensed work.
2. The per-file CC BY-NC exception is discoverable from the root `LICENSE`, the data README, and a `LICENSE` file beside the data.
3. `cargo package` ships no non-MIT file.
4. The case-study test pins the Boucher & Bennetts Table 2 NONMEM column in addition to the Monolix values, including the four parameters not currently asserted.
5. FeRx-NLME/ferx-r#312 cross-references this issue for why the naproxen example is held.

roninsightrx:
## Why

`data/mbma_naproxen.csv` shipped in #1063 attributed to Bracis et al., *CPT:PSP* 2026;15:e70158, which is **CC BY-NC-ND**. Two problems, one of them live:

1. **ND was being violated.** `mbma_naproxen_source.py` renames columns, recodes the treatment flag, drops `WP` and emits a ferx-layout CSV. That is an adaptation, and ND forbids *distributing* adaptations at all — not merely without attribution. The derived file was committed to a public repo.
2. **NC conflicts with the repo's MIT grant**, which promises commercial use over the whole tree.

The fix is cheap because **the dataset is not Bracis's**. Their paper says so: "The dataset [11] includes summary-level data from 18 placebo-controlled longitudinal trials of naproxen in OA using WOMAC pain scores," where [11] is Boucher M & Bennetts M, *Many Flavors of Model-Based Meta-Analysis: Part II — Modeling Summary Level Longitudinal Responses*, *CPT:PSP* 2018;7:288–297 ([doi:10.1002/psp4.12299](https://doi.org/10.1002/psp4.12299), `PMC5980518`) — the analysis Bracis et al. replicate in MonolixSuite. That article is **CC BY-NC, no ND** (verified against the Europe PMC full text: the only `creativecommons.org` URL in it is `licenses/by-nc/4.0/`).

Re-sourcing removes the ND violation outright, is more accurate scholarship, and — see below — upgrades the fixture from a one-engine comparison to a three-engine anchor. Closes #1085.

## What changed

**Re-pointed the derivation at the primary source.** `data/mbma_naproxen_source.py` now consumes `PSP4-7-288-s007.csv` (`STUD,NFLG,TIME,NTRT,FLARE,WP,WPSE`). It computes `DV = WP / WPSE` itself, because that file has no precomputed ratio; everything else is the same mechanical relabelling (`STUD → ID`, `NFLG → ARM`/`TRT`, `NTRT → NIND`).

**Labelled it.** New `data/mbma_naproxen.LICENSE` (full CC BY-NC 4.0 text, the attribution line, and what the exception means in practice); `data/mbma_naproxen.README.md` Source section rewritten to the primary source, with Bracis et al. cited as the tutorial replication and as the origin of the Monolix comparison values, not of the data; a short "everything is MIT except `data/mbma_naproxen.*`" note in the root `LICENSE` and `README.md`, so an auditor finds it without reading `data/`; `examples/mbma_naproxen.ferx` and the test module docs re-cited. One stale Bracis attribution in an existing `[Unreleased]` CHANGELOG entry (the #1029/#1031 entry) corrected in passing.

**Kept it out of the published crate.** `Cargo.toml` had no `include`/`exclude`, so `cargo package` shipped everything git-tracked. `exclude = ["data/mbma_naproxen.*"]` now covers the CSV, the data README and the data LICENSE. Verified:

```
$ cargo package --list | grep -i 'mbma\|naproxen'
data/mbma_naproxen_source.py
examples/mbma_naproxen.ferx
tests/mbma_naproxen_case_study.rs
```

All three are our own MIT files. The slow-gated test that reads the CSV does not run from a packaged crate.

### The regenerated CSV is not byte-identical — one digit, and it is explained

#1085 asked for byte-identity and said any difference is a finding to resolve first. There is one, in one of 122 rows:

```diff
-3,1,2,15.897229,1,1,0.188083087,121,0
+3,1,2,15.897230,1,1,0.188083087,121,0
```

It is a double-rounding artifact in the intermediate we no longer use. Bracis et al.'s `transWP` column stores `WP / WPSE` to seven decimals; for study 3, naproxen, week 2 that is `15.8972295`, whose float64 is `15.89722949999…`, which formats down at six decimals. The exact ratio is `2.99 / 0.188083087 = 15.897229504…`, which correctly rounds **up**:

```python
>>> f"{15.8972295:.6f}"          # via the tutorial's rounded intermediate
'15.897229'
>>> f"{2.99 / 0.188083087:.6f}"  # from the primary source's WP and WPSE
'15.897230'
```

So the new value is the more accurate one, and computing the ratio from `WP` and `WPSE` avoids the intermediate entirely. Nothing in the fit moves — same OFV to four decimals, same estimates to four decimals (below).

The data README now records the upstream file's size and `sha256` alongside the derived file's, so a future re-derivation can be checked rather than eyeballed.

### The upside: a three-engine anchor, now claimed

Boucher & Bennetts Table 2 reports this model fitted in **BUGS, NONMEM and R (`nlme`)**. Table pulled from the article's full text (not transcribed by hand), and ferx reproduces every point estimate in it:

| Parameter | BUGS | NONMEM | R (`nlme`) | Monolix | ferx |
|---|---|---|---|---|---|
| E0 (nonflare)      | 5.22 (0.28)  | 5.20 (0.13)  | 5.20 (0.27)  | — | **5.1972** (0.2578) |
| ΔE0 (flare)        | 0.93 (0.33)  | 0.96 (0.25)  | 0.96 (0.32)  | — | **0.9629** (0.3154) |
| Emax_p (nonflare)  | −1.14 (0.41) | −1.16 (0.24) | −1.15 (0.32) | — | **1.1580** (0.3147) |
| ΔEmax_p (flare)    | −0.86 (0.47) | −0.82 (0.09) | −0.82 (0.39) | — | **0.8197** (0.3792) |
| ΔEmax_n (naproxen) | −0.79 (0.06) | −0.79 (0.09) | −0.79 (0.07) | 0.792 (0.064) | **0.7930** (0.0646) |
| ln(ET50_p)         | −0.40 (0.17) | −0.37 (0.20) | −0.40 (0.17) | — | **−0.3709** |
| ln(ΔET50_n)        | −1.24 (0.31) | −1.17 (0.20) | −1.20 (0.29) | — | **−1.1747** (0.2712) |
| ET50_p (weeks)     | 0.67 | 0.69 | 0.67 | 0.698 | **0.6901** |
| ET50_n (weeks)     | 0.19 | 0.21 | 0.20 | 0.216 | **0.2132** |
| τ1 (ω E0, SD)      | 0.86 | 0.62 | 0.62 | 0.627 | **0.6215** |
| τ2 (ω Emax, SD)    | 0.71 | 0.74 | 0.74 | 0.763 | **0.7387** |

`mbma_naproxen_reproduces_the_published_analysis` now pins that whole NONMEM column, including the four parameters the Monolix comparison left unasserted, in addition to the Monolix values it already checked.

Two conventions, both confirmed against the paper's own equations rather than inferred:

- **Sign of the Emax family.** Eq. (1) is `Y = E0 + Emax·t/(ET50+t)` with `Emax` negative (pain falls); the shipped model writes `y = E0 − EMAX·t/(t+ET50)` with `EMAX` positive. Every Emax-family ferx estimate is therefore the negation of the published one — a parameterisation difference, not a disagreement.
- **ET50 is multiplicative in the log.** Table 2's `ET50_n` row is `ET50_p · exp(ln ΔET50_n)` = `0.69 · exp(−1.17)` = 0.21, matching `TVET50 * exp(B_LET50_TRT)`, despite Eq. (4) being written additively.

**Only point estimates are pinned.** The three engines agree on the estimates to the last published digit but disagree substantially on the standard errors — τ1 is 0.86 in BUGS against 0.62 in the other two, and the SE of ΔEmax_p is 0.47 / 0.09 / 0.39 across BUGS / NONMEM / `nlme`. A tolerance wide enough to span that would assert nothing. ferx's SEs are additionally pure `R⁻¹`, i.e. NONMEM's `MATRIX=R` rather than its default sandwich, which is a second reason not to read that column as one number. The one SE all four engines agree on is ΔEmax_n (0.06–0.09), and that one is asserted.

Tolerance is 0.01 absolute: Table 2 reports two decimals, so ±0.005 is pure rounding, and the worst observed residual across all eleven rows is 0.005 (ln ΔET50_n: −1.1747 vs −1.17).

## Alternatives considered

**Drop the fixture and the case study.** Cheapest way to a uniformly-MIT tree, and it was the wrong trade: it would discard the only real-data MBMA validation in the repo, and the re-sourcing makes that validation *stronger* (three engines, primary analysis) rather than merely legal. Carrying one labelled per-file exception for a test fixture that is never linked into the binary and never published in the crate is the cheaper cost.

**Keep Bracis as the source and add attribution.** Does not work — ND forbids distributing the adaptation regardless of attribution.

**Reproduce the artifact digit to preserve byte-identity.** Would mean reintroducing a dependency on the ND file's rounded intermediate in order to preserve a value that is measurably less accurate. Rejected.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#312 | not needed | — |

FeRx-NLME/ferx-r#312 mirrors the *synthetic* `emax_timecourse` example only and explicitly holds the naproxen one; NC data has no business in a CRAN-bound package. That stays true after this PR — the exception is now documented rather than latent, which is the outcome #312 was waiting on. Cross-referenced there.

## Breaking changes
- [x] None

## Numerical validation

- [x] Compared against: **Boucher & Bennetts Table 2 (BUGS / NONMEM / `nlme`)**, the Bracis et al. Monolix replication, and the prior ferx run on the pre-regeneration CSV
- [x] Reference values:

The two committed slow tests pass:

```
running 2 tests
test mbma_naproxen_reproduces_the_published_analysis ... ok
test mbma_naproxen_compartment_free_matches_the_dummy_clock_workaround ... ok

test result: ok. 2 passed; 0 failed
```

Full fit on the regenerated data — identical to the pre-regeneration run at every digit reported, which is the check that the one-digit CSV change is inert:

```
--- THETA Estimates ---
Parameter            Estimate           SE       %RSE
----------------------------------------------------
TVE0                 5.197214     0.257778        5.0
TVEMAX               1.157997     0.314661       27.2
TVET50               0.690095     0.115447       16.7
B_E0_FLARE           0.962909     0.315409       32.8
B_EMAX_FLARE         0.819738     0.379194       46.3
B_EMAX_TRT           0.792967     0.064577        8.1
B_LET50_TRT         -1.174657     0.271207       23.1

--- OMEGA Estimates ---
  ETA_E0               = 0.386233  (CV% = 62.1)  SE = 0.131123
  ETA_EMAX             = 0.545723  (CV% = 73.9)  SE = 0.189083

--- SIGMA Estimates (additive) ---
  ADD_ERR [FIX]        = 1.000000

OFV: 261.8149
```

Against Table 2's NONMEM column: `5.1972` vs 5.20, `0.9629` vs 0.96, `1.1580` vs 1.16, `0.8197` vs 0.82, `0.7930` vs 0.79, `ln(0.690095) = −0.3709` vs −0.37, `−1.1747` vs −1.17, `0.6901` vs 0.69, `0.690095·exp(−1.174657) = 0.2132` vs 0.21, `√0.386233 = 0.6215` vs 0.62, `√0.545723 = 0.7387` vs 0.74.

## Performance

## Tests
- [x] Regression test added that fails without this fix — `mbma_naproxen_reproduces_the_published_analysis` now pins the eleven Table 2 point estimates, so a future re-derivation that silently changes the data fails against the primary publication rather than against a value we transcribed from our own previous run
- [x] `cargo test --release --features slow-tests --test mbma_naproxen_case_study` passes (2/2)

The only Rust change is in `tests/`, which is coverage-ignored by role per `codecov.yml`.

- [x] Not applicable (licensing / provenance correction; `examples/mbma_naproxen.ferx` changes only in its header comment)

## Docs
- [x] No other user-visible change — `docs/model-file/error-model.qmd` mentions "the naproxen-style analysis" but attributes nothing, so nothing there needed re-citing

## Checklist
- [x] `cargo clippy --tests --features slow-tests` clean on the touched file (pre-existing warnings elsewhere unchanged)
- [x] `cargo fmt --all --check` clean
- [x] No `Dual2` sensitivity path touched

### ferx-site / ferx-book
- [x] No new DSL syntax, `[fit_options]` key, estimator or example file — nothing to mirror. The one data-file rule that would normally apply ("new data file added to `data/` → mirrored to `ferx-r/inst/examples/data/`") is exactly what FeRx-NLME/ferx-r#312 holds, and this PR is the reason it stays held.

### Example execution
- [x] `examples/mbma_naproxen.ferx` runs cleanly via CLI against the regenerated data (output above)
- [x] No site/book page affected

## Reviewer hints

Two places worth actual attention:

1. **The one-digit CSV change.** `git diff data/mbma_naproxen.csv` is a single line. The question to satisfy yourself on is whether the new digit is the correct one; the two-line Python above is the whole argument.
2. **The tolerance choice in the new assertions.** `NM_TOL = 0.01` against two-decimal published values. The alternative — asserting the SE column too — is deliberately not done, and the module docs say why. If you disagree, that's the thing to push on.

Everything else is prose: licence text, README, root notes, `Cargo.toml` `exclude`.

The provenance claims are checkable without a subscription. The licence: `curl -s https://www.ebi.ac.uk/europepmc/webservices/rest/PMC5980518/fullTextXML | grep -o 'creativecommons[^<"]*'`. The data: the same endpoint's `/supplementaryFiles` returns a zip containing `PSP4-7-288-s007.csv` (PMC's own file endpoint is behind a proof-of-work challenge and returns HTML to `curl`).

## Open questions

None blocking. One judgement call recorded for the record rather than for debate: a CC BY-NC file in an otherwise-MIT repo means downstream commercial users are not granted rights over *that file*. It is a test fixture, never linked into the binary, never in the published crate, and the exception is explicit in four places — but the constraint is real, and #1085 is where the decision to accept it is written down.


roninsightrx:

Reviewed `a42bd102..HEAD` — the naproxen re-sourcing commits. The `src/ode/**` files in `main...HEAD` belong to the already-merged #1066 and are out of scope here.

### Verified positively

- The committed CSV round-trips exactly through the new script: I reconstructed a synthetic `PSP4-7-288-s007.csv` from `WP = DV·WPSE` and the script emitted a byte-identical `data/mbma_naproxen.csv`.
- Its sha256 matches the hash pinned in the README.
- Shape is 122 rows / 18 studies / 36 arms, with `ARM == TRT` on every row — so `ARM <- NFLG` is consistent with the old `ARM` column.
- The one changed digit is correct: `2.99 / 0.188083087 = 15.897229504…` → `15.897230`.
- `cargo check --tests` is clean, and no stale Bracis attribution is left anywhere as the data source.

### Findings

**1. `data/mbma_naproxen_source.py:47` — medium: the new shape checks don't restore the guarantee the comment claims**

The comment says the shape checks restore the loud failure the old `TRT == "Naproxen"` test gave. They don't. `NFLG` is written into both `ARM` and `TRT` without ever inspecting what `1` means. A re-issued supplementary file that recoded the flag (1 = placebo) passes `EXPECTED_COLUMNS`, `EXPECTED_ROWS` and `EXPECTED_FLAGS` unchanged and silently emits an inverted `TRT` column — the naproxen effect on Emax then estimates with the wrong sign, and nothing in the pipeline notices.

The README already states the invariant that would catch this ("if a re-derivation does not reproduce the second hash, the difference is a finding"). Make the script enforce it: sha256 the output and compare, or add a semantic check (naproxen arms' `WP` below placebo at the last timepoint).

**2. `data/mbma_naproxen_source.py:101` — low: the 9-decimal `WPSE` reformat is unchecked**

`WPSE` is reformatted to 9 decimals while `DV` is computed from the full-precision `float(r["WPSE"])`. The docstring asserts 9 dp is exact for every published value, but nothing checks it. If any upstream `WPSE` carried a 10th significant decimal, the committed `DV` would have been divided by a different number than the `WPSE` the model divides the prediction by — silently perturbing the `1/WPSE²` weighting the whole analysis rests on.

One line fixes it:

```python
if float(f"{wpse:.9f}") != wpse:
    raise SystemExit(f"WPSE {wpse!r} is not exact at 9 decimals")
```

**3. `data/mbma_naproxen_source.py:79` — low: validation strips, the writer does not**

Validation strips (`r["NFLG"].strip()`) but the writer does not — `r["NFLG"]`, `r["STUD"]`, `r["TIME"]`, `r["FLARE"]`, `r["NTRT"]` are written raw. A supplementary CSV exported with `", 1"`-style padding passes the flag check and then emits leading spaces into the fixture, which `read_nonmem_csv` will reject (or mis-parse) rather than the script failing at derivation time. Strip on write too.

**4. `Cargo.toml:16` — low: the excluded fixture leaves a test that panics in the packaged crate**

`exclude = ["data/mbma_naproxen.*"]` drops the fixture from the packaged crate, but `tests/mbma_naproxen_case_study.rs` still ships in it and hardcodes `data/mbma_naproxen.csv` (`tests/mbma_naproxen_case_study.rs:89`). Anyone running `cargo test --features slow-tests` against the published or vendored crate gets a panic on a missing file instead of a skip. Either exclude the test file too, or have it skip with a clear message when `DATA` does not exist.

roninsightrx:
## Review findings addressed — 02685745

All four, plus verification of each.

**1 (medium) — `NFLG` is now read for what it *means*, not just its shape.**
`check_naproxen_arms_improve` compares each study's two arms at their last shared
time point. All 18 published trials have the `NFLG == 1` arm lower there, so the
guard requires a two-thirds majority — a wide margin against real data, and zero
against a recoded flag. Verified by re-running the script on a synthetic
supplementary file with 0 and 1 swapped:

```
error: only 0 of 18 studies have the NFLG==1 arm below the NFLG==0 arm at their
last shared time. NFLG does not mean 1 = naproxen here, so ARM and TRT would be
inverted and the naproxen effect on Emax would estimate with the wrong sign.
```

The README's sha256 invariant is now enforced rather than stated: the script
hashes the text it is about to write and refuses to write on a mismatch.
`--allow-new-hash` acknowledges a deliberate refresh (and the README says to
record the new hash when you use it). Verified by perturbing one `WP` by 0.01 —
rejected with both hashes printed, and accepted, with the new hash flagged
`(NEW)` in the summary line, under the flag.

The two guards are deliberately not redundant: the hash is what you switch off
to refresh, and the semantic check still has to hold when you do.

**2 (low) — `WPSE` is verified exact at 9 decimals.** `if float(f"{wpse:.9f}") !=
wpse` before deriving, naming the offending study and time. Verified by appending
a 10th digit to one value:

```
error: WPSE '0.1369244655' (study 1, time 12) is not exact at 9 decimals. DV
would be divided by a different value than the WPSE written beside it,
perturbing the 1/WPSE^2 weighting.
```

**3 (low) — every written field is stripped**, and the `float()` parses too.
Verified by padding every field of every row of the synthetic source with a
leading space: output is byte-identical to the committed fixture instead of
carrying the padding through.

**4 (low) — the slow test is excluded from the packaged crate** alongside the
fixture it hardcodes, so a published or vendored crate has no test rather than a
panic on a missing file. `cargo package --list` now shows neither
`data/mbma_naproxen.csv` nor `tests/mbma_naproxen_case_study.rs`.

### Regression check

The committed fixture is unchanged, and still round-trips byte-identically: I
reconstructed a synthetic `PSP4-7-288-s007.csv` from `WP = DV·WPSE`, ran the
rewritten script, and `cmp` is clean against `data/mbma_naproxen.csv` — so all of
the above is added guard-rail, not a change to the data. `cargo check --tests`
passes.

`CHANGELOG.md` and `data/mbma_naproxen.README.md` are updated to describe the
enforced invariants and the widened exclude.

Issue #1080, opened 2026-08-25
epic(ode): runtime stiffness detection — early abort, escalation for named methods, and step-level AutoSwitch


roninsightrx:
Follow-up to #978, which is now Part A only (a-priori Jacobian eigenvalue probe, shipped in
#1066 and **the default** for ODE models). This issue carries the two runtime layers that were
left out, plus a blind spot in the signal both of them were specified to use.

Read #978 for the original design and the 5-solver sweep that motivates it; this issue records
what changed in the process of building Part A, because two of its assumptions did not survive
measurement.

## What Part A already covers

Worth stating up front, because it shrinks Part B:

- **The `auto` guard already catches the silent freeze-pad.** #1066 discards an escalated solve
  and re-runs it explicitly when it returns non-finite **or clamps at `min_dt`**. That is #959's
  failure mode, and it is covered for `auto` (i.e. for the default) today.
- **Per-segment probing already tracks parameter drift.** The probe re-runs at whatever θ/η the
  current likelihood evaluation carries, so "the fit walks into a stiff region" is handled
  without a runtime detector.

What is *not* covered: a user who names a stiff method by hand still gets the freeze-pad with no
guard, and nothing bails out early — a stalled explicit solve still grinds to `max_steps`.

## Part B — early abort + escalation for explicitly-named methods

Original plan in #978. Reduced scope in light of the above:

1. **`stiff_abort_after: Option<u32>` in `OdeSolverOptions`** — stop `integrate_dense_g` once
   `min_step_clamped_steps` crosses a budget instead of grinding to `max_steps`. Addresses #708
   and the *cost* half of #959. Independent of `auto`.
2. **Surface the guard's decisions.** `OdeSolverStats` gained `auto_stiff_segments` /
   `auto_stiff_rejected` in #1066, but no fit or CLI path reports solver statistics at all today,
   so a user cannot see that their model escalated, or that an escalation was rejected. A
   non-zero rejection count is the single most actionable ODE diagnostic there is — "the probe
   was right that this is stiff and wrong about which method could integrate it" — and right now
   it is invisible outside the test suite.
3. **Per-subject latching.** #978 Phase 3. Now a pure optimisation rather than a correctness
   measure: measure first, since the probe is ~one solver step per segment and may not be worth
   the cache.

## Part C — step-level `AutoSwitch`

Unchanged from #978: switch the stepper *in place* inside `integrate_dense_g` rather than
[...]

Issue #708, opened 2026-07-06
Analytic Dual-sensitivity gradient hangs on time-varying-CL ODE (TIME^HILL): single-subject straggler, no bail on finite min-step clamps


roninsightrx:
## Summary

A FOCEI fit with the default analytic gradient (`gradient = auto` → Dual2 → L-BFGS) **effectively hangs** on the pembrolizumab `model066` (2-cpt IV with an Emax-on-time clearance multiplier `CL(t) = CLI·exp(IMAX·TIME^HILL/(TI50^HILL + TIME^HILL))` implemented as a user ODE). The process is not deadlocked — it is one CPU core grinding on a single pathological subject's **analytic sensitivity ODE** integration, while the other cores sit idle waiting on the per-subject `par_iter`.

Switching to `gradient = fd` (→ BOBYQA, value-only ODE) converges the same model in ~30 s. So the value-only dynamics are fine; only the Dual1/Dual2 sensitivity path is pathological.

## Reproduction

Model: `ferx-testdata/pembrolizumab_radboudumc/ferx/model066.ferx` (303 subjects, 1231 obs), default `[fit_options]`:

```
[fit_options]
  method     = focei
  maxiter    = 4000
  covariance = true
```

Run with default cores. Optimizer prints `gradient: analytic (Dual2)` → `Starting NLopt Lbfgs`, reaches `Eval 1: OFV = 6787.6`, then sits. `ps` shows **99.9 % CPU (one core)** for 15+ minutes with no further output.

## Diagnosis (from `sample` on the live process)

- Main thread parked in the rayon `install` latch; workers idle in `wait_until_cold`; **one** worker busy.
- Busy stack (leaf-most, by sample count):
  ```
  population_gradient_sens_mixed
    → sens::ode_provider::integrate_tvcov_g<Dual1>
    → ode::solver::solve_ode_g<Dual1>
    → parser::model_parser::eval_bytecode          (ODE RHS interpreted)
  estimation::inner_optimizer::find_ebe
    → inner_minimize_with_grad / nelder_mead_minimize / argmin_inner_fallback
    → ode::solver::solve_ode_dense
  ```
- Single-subject straggler: `par_iter` over subjects, one subject's sensitivity solve never completes in reasonable time, so 14 cores idle.

## Root cause

[...]

PR #1087, opened 2026-08-25
feat(ode): report solver decisions, add a stiff-abort budget, cheapen the probe [#1080]

roninsightrx:
## Why

Part A of #1080 (the a-priori stiffness probe, #1066) shipped `ode_method = auto` as the default. Three things it left open:

1. **A stalled explicit solve still grinds to `max_steps`** (#708, and the *cost* half of #959). Naming a stiff method by hand gets no guard and no bail-out.
2. **The `auto` guard's decisions are invisible.** `OdeSolverStats` gained `auto_stiff_segments` / `auto_stiff_rejected` in #1066, but **no production path reports solver statistics at all** — `ode_predictions_with_solver_stats` has no production caller. So a model could escalate, have the escalation *rejected*, and be re-solved explicitly with no trace outside the test suite. A rejection is the single most actionable ODE diagnostic there is ("the probe was right that this is stiff and wrong about which method could integrate it"), and only the user can act on it.
3. **Per-subject latching** — #978 Phase 3, which #1080 downgraded to "measure first; the probe may not be worth the cache".

The issue's design constraint is honoured throughout: the **a-priori probe is the only layer that sees the fast-binding family** (@TeunP's sweep — 40 % median error with *zero* clamps and no stall), so nothing here is allowed to weaken per-segment probing.

## What changed

**1. `ode_stiff_abort_after` (#708).** New `[fit_options]` key + `OdeSolverOptions::stiff_abort_after: Option<u32>`: abandon a segment once that many of its steps have clamped at `min_dt` instead of grinding to `ode_max_steps`.

- **Off by default, deliberately.** An abort freeze-pads the segment's remaining output times, so it trades a slow-but-eventually-integrated segment for a padded one — a change to the objective, not only to its cost. `None` is bit-identical to the pre-#1080 driver (pinned by a test).
- The **threshold driver** (TTE / adaptive dosing) reports an abort as `Failed` rather than padding: its output *is* the answer (an event time, or a censoring), and a padded crossing would be laundered into the likelihood as though it had been integrated.
- Composes with `auto`: an escalation that trips the budget is discarded by the existing guard exactly as any other stalled escalation is (the guard's trigger is `min_step_clamped_steps > 0`, which an abort has by construction), so the budget makes a *failed* escalation cheaper without making the caller keep its freeze-padded trajectory.
- Reads `.val()` like every other driver decision, so `f64` and `Dual2` abort on the same step (tested).

**2. The fit now says what the solver did.** The post-fit per-subject prediction pass runs inside a thread-local `SolverStatsScope`, and its counters become an `ode_solver` warning (new `WarningCode`, classified and documented):

- `Warning` — steps clamped at `min_dt`, an escalation discarded and re-solved explicitly, or segments cut short by `ode_stiff_abort_after`.
- `Info` — `auto` escalated and it worked. Routine on a stiff model (TMDD `cr`: 240/580 segments) and not a problem, but a decision the user never asked for and could not otherwise see.
- Counters ride along in the `details` payload.

Using the **ordinary predictor** under a scope, rather than giving `ode_predictions_with_solver_stats` a production caller, is the point: the counters then describe the real dispatch (TV covariates, resets, the event-driven walker, IOV) at the final estimates instead of a diagnostic re-creation of it, and it costs one thread-local read per segment. Scope is honest about its reach — it sees the trajectories the fit *reports*, not every parameter value the optimizer passed through; that limitation is documented on the function.

**3. Per-subject latching: measured, rejected, and replaced with something sound.**

The cost is real — on a 3-state binding model the probe is 0.56 µs against a 1.75 µs benign segment (33 % overhead; 19 % on a realistic observation grid). But a per-subject cache is **unsound**: stiffness is a property of *where the model is* (TMDD `cr`/`crib` span 3.7–126 within one dataset), and a binding model's fast mode is identically zero before its first dose — latching the first segment's "non-stiff" verdict would defeat exactly the layer the issue says must be protected.

So instead of caching the verdict, the probe got cheaper. `resolve_method` now tries **Gershgorin's bound** before the eigensolve: every eigenvalue lies in a disc centred on a diagonal entry, so `max_i (|a_ii| + Σ_{j≠i}|a_ij|)` bounds `|λ|max` and therefore `|Re λ|max`. A bound below the threshold *proves* no eigenvalue reaches it — one-sided in exactly the direction the escalation test (`λ_max ≥ threshold`) can short-circuit. `O(n²)` adds replace the `O(n³)` Schur iteration and its `DMatrix` allocation on the segments that cannot be stiff, which is most of them.

Measured (`measure_probe_cost_against_segment_cost`, an `#[ignore]`d harness committed so the numbers are reproducible):

| | probe | `auto` segment | pinned `rk45` segment | `auto` overhead |
|---|---|---|---|---|
| benign, bare segment — before | 0.56 µs | 2.32 µs | 1.75 µs | **33 %** |
| benign, bare segment — after | **0.055 µs** | 1.80 µs | 1.75 µs | **3.1 %** |
| benign, 7-point obs grid — before | | 3.62 µs | 3.05 µs | **18.9 %** |
| benign, 7-point obs grid — after | | 3.13 µs | 3.08 µs | **1.3 %** |
| stiff (kon = 100) | 0.56 µs | 13.7 µs | 1028 µs | probe = 3.6 % of the segment |

No decision changes: the fast path is tested against the exact eigensolve across five decades of the binding rate at three states each, and the exact `max_abs_re_eigenvalue` is kept (now `#[cfg(test)]`) as the definition it must agree with.

## Alternatives considered

- **Threading `Option<&mut OdeSolverStats>` from `fit()` down to the drivers** for item 2 — would touch every prediction dispatcher (`compute_predictions_*`, the event-driven walker, the SS equilibrator, the MR reroute) to serve a diagnostic production never asks for, and would still only cover the paths someone remembered to thread.
- **Making `stiff_abort_after` on by default** — rejected: it changes the objective (freeze-padding), and #1066's guard already covers the *silently-wrong* half for `auto`. The default stays "grind, and now say so".
- **Per-subject latching (item 3)** — measured and rejected on soundness, above.
- **A step-rejection-rate signal** (the issue's suggestion for the fast-binding family) — not built here. It needs measuring on @TeunP's sweep first, and it belongs with Part C / the better-discriminator problem rather than bolted onto this.

## Cross-repo dependency

`FitOptions` gained a field with a `Default`, and the new warning is an additive `WarningCode` variant; the R wrapper reads `warnings_structured` generically and needs no change to keep working. A follow-up to expose `ode_stiff_abort_after` in `ferx_fit(settings = …)` is optional, not required.

## Breaking changes
- [x] None

`OdeSolverOptions` / `OdeSolverStats` gained fields (both are `pub` structs constructed with `..Default::default()` in-tree); `WarningCode` is `#[non_exhaustive]`.

## Numerical validation

Default-path numerics are **unchanged by construction** — the abort is opt-in and the Gershgorin bound only short-circuits the direction it can prove — and that is pinned rather than asserted:

- `tests/transit_nonmem_anchor.rs` (the #387 solver-decision gate) reproduces its documented step counts **exactly** on this branch, for every method:

```
=== Savic transit anchor @ 1e-9 (NONMEM TOL=9) ===
method           attempt    accept    reject   clamped
Rk45                4060      3940       120         0
Vern7               1560      1420       140         0
Rosenbrock23       85300     85180       120         0
Rodas4             16340     16200       140         0
Rodas5P             4240      4060       180         0
```

- `the_gershgorin_fast_path_agrees_with_the_exact_eigensolve` — identical verdicts to the exact eigensolve across `kon ∈ [1e-3, 1e3]` × three states.
- `stiff_abort_after_leaves_a_clean_solve_bit_identical` — bit-for-bit equality (`to_bits`) on a healthy solve with a budget of 1.
- Full `cargo test --lib` green (3371 tests), including the `Dual2`-vs-FD parity suites that would catch any change to a sensitivity path.

**No NONMEM anchor applies to the new behaviour itself**: NONMEM has no min-step-clamp budget and no solver-diagnostics warning, and the default path's numbers are unchanged (the anchor above is the comparison). The abort's semantics are pinned by unit tests instead.

CLI evidence for item 2 (stiff two-state model, `data/one_cpt_iv.csv`):

```
--- Warnings ---
  * W_ODE_SOLVER_DIAGNOSTICS: ode_method = auto escalated 30 integration segment(s) to a stiff
    stepper at the final estimates; every other segment used rk45, no escalation was rejected,
    and no step clamped at the minimum step size. …
```

and the same model with `ode_method = rk45` pinned: **53.6 s** to OFV −310.24 within a 20-iteration budget, versus **0.21 s** to OFV −322.38 under `auto`. A non-stiff example (`examples/mm_iv.ferx`) emits no `ode_solver` warning, so the diagnostic does not become noise on every ODE fit.

## Performance
- [x] Faster — `auto`'s per-segment overhead on a non-stiff 3-state model: **33 % → 3.1 %** (bare segment), **18.9 % → 1.3 %** (7-point observation grid); probe 0.56 µs → 0.055 µs. Stiff segments are unaffected (they still pay the eigensolve, correctly). `ode_stiff_abort_after` bounds a stalled segment's cost when set.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3371)

New Tier-1 tests: abort fires at the budget / is a no-op above the clamp count / is bit-identical on a clean solve / aborts identically at `Dual2`; the threshold driver fails instead of padding; the stats scope collects, nests, restores, and tees rather than steals; an aborted escalation is still rejected by the `auto` guard; the Gershgorin bound never undercuts the exact spectrum and never changes a verdict; the warning's three shapes (`None` / `Info` / `Warning`) plus its `classify_warning` round-trip; and two end-to-end `fit()` tests — a stiff ODE model reports its escalations, a benign one reports nothing.

<details>
<summary>Example</summary>

```toml
[fit_options]
  # Give up on a segment after 25 of its steps have clamped at the minimum step
  # size, instead of grinding on to ode_max_steps. Off by default.
  ode_stiff_abort_after = 25
```

```
--- Warnings ---
  * W_ODE_SOLVER_DIAGNOSTICS: the ODE solver did not integrate cleanly at the final estimates
    (ode_method = auto): 12 step(s) clamped at the minimum step size …; 3 of 240 stiff
    escalation(s) chosen by ode_method = auto were discarded as unusable and re-solved with
    rk45 — the stiffness probe was right that those segments are stiff and wrong that the stiff
    method it picked could integrate them … naming ode_method = rodas5p (or rosenbrock23)
    explicitly is the next thing to try.
```

</details>

## Docs
- [x] `docs/` updated for user-visible changes — `model-file/fit-options.qmd` (the new key), `model-file/ode-models.qmd` ("Where those counts show up" + "Bounding what a stalled segment costs"; the `auto` guard section now points at the report), `warnings.qmd` (the `ode_solver` code and its `details` payload)
- [x] `CHANGELOG.md` `[Unreleased]` entry added (Added ×2, Performance ×1; the #978 entry's "a fit does not print solver statistics today" clause is now amended, since it ships in the same release)

## Checklist
- [x] `cargo clippy` clean (no new warnings; the pre-existing ones in `sir.rs` / `validation.rs` / `fit.rs` and the dead store at `solver.rs:1031` are untouched)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the abort is a `.val()`-only decision, tested for `f64`/`Dual2` identity
- [x] No version bump needed (additive)

## Docs & examples
- [x] New `[fit_options]` key → `docs/model-file/fit-options.qmd` updated (a ferx-site PR would be the mirror if that page is duplicated there)
- [x] No new example `.ferx` or data files
- [x] Rebuilt ferx-core: `cargo build --release`; ran `examples/mm_iv.ferx` (clean, no new warning) and a stiff two-state model (warning as quoted above)
- [ ] ferx-r rebuild — not needed, no public-API change the wrapper consumes

## Reviewer hints

- `src/ode/solver.rs`: the abort check sits *after* the accept block so a clamped step's own `saveat` values are still saved; `min_step_clamps` is counted locally rather than read off `stats` so the budget behaves the same whether or not the caller asked for counters.
- `integrate_resolved_g` was split into a tee wrapper + `_inner`; the guard logic inside `_inner` is untouched.
- `src/ode/stiffness.rs`: the one-sidedness of the Gershgorin bound is the whole argument — worth checking that the early return can only ever choose the *explicit* method.
- `src/api/fit.rs`: the scope is scoped to `compute_subject_results` only, and that loop is sequential (the thread-local sees all of it).

## Open questions

- The `W_ODE_SOLVER_DIAGNOSTICS:` token is user-visible in the CLI warning list, matching the `W_DERIVED_CMT_*` / `W_ABSORPTION_TWIN_DECLINED` convention. Happy to strip it to a prose opener if the convention is considered legacy.
- The `Info`-severity escalation note fires on every fit where `auto` escalates. That is the "you could not see this" half of item 2, but it is also a line on every TMDD fit — say the word if it should be gated behind `verbose` instead.

https://claude.ai/code/session_01Juy7XdkXCs4BMSuycDjfNt


roninsightrx:
## Review follow-up — all six findings fixed in `4abf643`

Every finding reproduced against the code as written; none were false positives. Fixes, in the order reported:

**1. Spurious abort on a segment that finished (`solver.rs`, medium).** Confirmed: the check sat at the bottom of the loop body with no `t < tf` guard, so a segment whose *last* step clamped was counted as truncated even though it reached `tf` with every `saveat` saved. Now `t < tf - 1e-15 && …`, with the reason stated at the call site — the counter's whole job is to distinguish "cheap" from "cheap because it stopped integrating", and a false positive there is the one failure it must not have. Regression test: `a_clamp_on_the_last_step_is_not_an_abort` (5 clamps, budget 5, segment ends exactly on the fifth → `stiff_aborted_segments == 0`).

**2. Threshold driver discarded a crossing it had already found (`solver.rs`, medium).** Confirmed and the more serious of the two: the budget check was inside the clamp block, ahead of the crossing test, so a step that both clamped and bracketed the crossing returned `Failed` — and `solve_ode_until_threshold` escalates that to a hard error after the explicit re-solve aborts the same way. Moved below the accept block, mirroring the dense driver's existing "check after the saves" rule. Regression test: `the_threshold_driver_keeps_a_crossing_found_on_a_clamped_step`.

**3. Rejected escalations claimed freeze-padding (`solver.rs` + `postfit.rs`, medium).** Confirmed, and it was structural rather than incidental: `stalled = min_step_clamped_steps > 0` *is* the rejection trigger, so **every** rejected escalation carried the false claim. Took the suggested route — new `OdeSolverStats::discarded_clamped_steps` records clamps taken inside an attempt the guard threw away, the warning's freeze-padding clause uses `min_step_clamped_steps - discarded_clamped_steps`, and the rejection clause now says explicitly that those steps are excluded and why. The double-count is fixed the same way: a budgeted abort inside a discarded attempt is no longer counted at all, since `stiff_aborted_segments` now means "truncated **result**", not "truncated attempt" (documented on the field). `details` gained `discarded_clamped_steps` and `kept_clamped_steps`.

**4. `Info` note promoted to `Warning` on re-classification (`types.rs`, low).** Confirmed — in-process is fine (`rebuild_warnings_structured` prefers the native entry), but anything re-classifying from the plain `warnings` strings, e.g. a fit YAML read back, saw only the shared token. The escalation note now carries its own `W_ODE_SOLVER_ESCALATION_NOTE` token with an `Info` classifier arm. Test asserts the severity survives the round trip.

**5. Twin-only models were outside the scope gate (`fit.rs`, low).** Confirmed, and worth more than "low" in one respect: `sync_ode_solver_opts` deliberately carries the tolerances *and* the new `stiff_abort_after` onto the absorption twin, so those subjects could abort with nothing reported at all. Gate is now `integrates_odes(model)` = `ode_spec.is_some() || absorption_ode_equivalent.is_some()`, extracted so it is testable; test pins a closed-form `one_cpt_transit` (no `ode_spec`, twin present) as in scope.

**6. Docs understated the trade (low).** Agreed and generalised: because the budget lives on `OdeSpec::solver_opts`, it bites on *every* likelihood evaluation, which makes the objective discontinuous in θ (a segment clamping `N-1` times at θ and `N` at θ+h truncates in a different place) — stalling the line search and corrupting the covariance step's FD Hessian. `fit-options.qmd`, `ode-models.qmd` and the changelog now say that a fit whose `ode_solver` warning reports aborts is a *diagnosis*, not an estimate, and to turn the budget off once a stiff method fixes the grinding.

On the three minor notes: `false` is now documented alongside `0` / `off` / `none` (kept because an R `FALSE` arrives as that string; `true` has no sensible meaning for a count). `until_threshold_with`'s work staying outside `SolverStatsScope` is real — the event-time driver takes no stats at all, as `auto_stiff_segments`' own doc already records — and is left for the Part C / step-rejection-rate work rather than widened here. The two `fit()` tests are Tier-1 by the same precedent as `src/api/tests/iov_integration.rs`; both return after one outer iteration and the pair runs in 0.55 s.

`cargo test --lib` green (3377), `cargo check --tests` clean, no new clippy warnings.

PR #1088, opened 2026-08-25
feat(api): let a [simulation] design state the covariates of the arms it invents [closes #1083]

### LINKED ISSUE #1083: design(sim): what does simulate() do with weight = COL when the arm is invented?
**Type:** Design question → then a fix
**Scope:** `src/api/simulate.rs`, `src/sim/`, and whatever `[simulation]` grows to express it.
**Raised by:** @TeunP on #1063. Blocks the τ²-based prediction interval on the MBMA epic #1032. Follows #1031.

## The question

`kappa K ~ γ² weight = NARM` (#1031) and `DV ~ additive(S) weight = WPSE` (#1029)
both read a **data column**. `fit()` always has one — the arm exists, its size and
its reported SE are in the dataset.

`simulate()` frequently does not. The point of simulating a trial is to invent
arms that are not in the data: *"what would a 300-subject arm of this design look
like?"* There is no row to read `NARM` or `WPSE` from, because the arm does not
exist yet.

Today's behaviour needs establishing (a missing covariate reads `0.0` in some
paths, which for a weight would be a division by zero or a silent collapse), and
then deciding.

## Why it matters beyond tidiness

1. **Trial simulation is a primary MBMA deliverable.** Dose/duration/comparator
   selection is on the list of what MBMA is *for* (Upreti & Venkatakrishnan). A
   simulate path that cannot express "an arm of size N" cannot answer the question
   the analysis was run to answer.
2. **The prediction interval depends on it.** A prediction interval for a *future*
   study needs both parameter uncertainty and τ², and the BTAV contribution to a
   future arm is `γ²/N` for the N you are proposing — so "what N?" is not an
   implementation detail, it is the input.
3. **Silence here is expensive.** A weight that quietly reads as 1 gives a
   simulated arm the variability of a single-subject arm; one that reads as 0
   gives it none, or a division by zero. Both produce plots.

## Options

- **Require it in the simulation design.** `[simulation]` gains per-arm weight
  values, and simulating without one for a weighted κ is an error naming the
  column. Explicit, and it makes the modeller state the design — which is what
  they are doing anyway.
- **Carry the weight expression and evaluate it against the simulated row.** Works
  when the weight is a function of columns the simulation already defines; still
  needs an answer when it is not.
- **Fall back to the fitted data's typical value**, i.e. the median arm size,
  reported loudly. Convenient, and defensible for a quick look, but it silently
  answers a different question than the one asked.

Recommendation: the first, with the third available explicitly rather than as a
default. A simulated trial's arm sizes are part of the design being evaluated, so
requiring them is not a burden — it is the question.

## Definition of done

1. Establish and document what happens today for a weighted κ and a weighted
   residual under `simulate()`, including a missing weight column.
2. A stated, tested convention, with a loud error rather than a silent default
   where the weight cannot be resolved.
3. Round-trip: simulating with the fitted data's own weights must reproduce the
   fitted model's variability structure (the degenerate oracle for whatever
   mechanism is chosen).
4. Documented in `docs/model-file/simulation.qmd` alongside the weight declarations
   themselves.

roninsightrx:
## Why

`kappa K ~ γ² weight = NARM` (#1031) and `DV ~ additive(S) weight = WPSE` (#1029) both read a data column. `fit()` always has one — the arm exists, and its size and its reported standard error are columns in the dataset. `simulate()` frequently does not: the point of simulating a trial is to propose arms that are **not** in the data, and an arm that does not exist yet has no row. #1083 asked what happens today and what the convention should be.

The investigation (posted [on the issue](https://github.com/FeRx-NLME/ferx-core/issues/1083#issuecomment-5418309064)) found two defects.

**1. `[simulation]` could not express a covariate at all.** `SimulationSpec::covariates` was a field the parser always wrote empty (`model_parser.rs:6104`) and nothing ever read; `run_model_simulate` built every synthetic subject with `covariates: HashMap::new()`. A weighted model reaching `--simulate` was rejected with:

```
Error: simulation failed: Model references covariate(s) not found in data
(case-sensitive): NARM. Available covariate columns: (none).
```

— a message that names no fix, on a path that has no data file. Meanwhile `docs/model-file/simulation.qmd:164` documented the feature as though it worked ("Covariates can be specified per subject in the `[simulation]` block"), which was false.

**2. The weight guards had uneven coverage across the simulate entry points.** Each ran its own hand-picked subset of the model-vs-data checks:

| entry point | `check_covariates` | `check_kappa_weights` | `check_residual_magnitude` |
|---|---|---|---|
| `simulate_with_options_diag` (what ferx-r calls) | ✅ | ✅ | ❌ |
| `simulate_with_uncertainty` | ✅ | ✅ | ❌ |
| `simulate_adaptive_*` | ✅ | ❌ | ❌ |
| `simulate` / `simulate_with_seed` | ❌ | ❌ | ❌ |
| `fit()` | ✅ | ✅ | ✅ |

That mattered more than it looks, because **`BinOp::Div` returns `0.0` when its divisor underflows** (`model_parser.rs:16145`, bytecode twin at `:16749`). An unresolved weight therefore does not blow up:

- `κ/√0` evaluates to **zero**, so the arm quietly loses its between-arm variability — no `NaN`, no `inf`, nothing to trip over;
- a zeroed residual weight collapses the *additive* loading, so the simulated observation comes back equal to its own IPRED. Reproduced on `main`:

```
id=1 time=1 ipred=8.2104679996  value=21.5042801990   <- NARM=100, noisy
id=2 time=1 ipred=7.3464335261  value=7.3464328618    <- NARM=0, noise gone
```

Both produce plausible-looking plots. The guards were the entire defence, and two of them were missing where it counted.

## What changed

**`[simulation] covariate NAME = ...`** — a scalar (broadcast to every subject) or a list (one value per subject):

```
[simulation]
  n_subjects = 3
  times      = [1, 2, 4, 8]
  covariate NARM = [400, 200, 50]
```

Validated against the `[covariates]` block when one exists, against `n_subjects` (a short list is an error, never a recycled or zero-filled tail), for duplicates, and for finiteness. Keys stay order-independent — the length rule runs after the block is read, so a `covariate` line may precede its `n_subjects`. The arm is matched on the `covariate` *word*, not the prefix, so a future `covariates = ...` key is still reported as unknown rather than read as a covariate named `s`.

**Required, not defaulted.** A model that references a covariate the design does not supply is refused, in the terms this path has:

```
[simulation]: the model references covariate(s) `NARM` that the simulation design does
not supply. A simulated trial invents arms that are in no dataset, so their covariates —
an arm size behind `weight = N`, a reported standard error behind `weight = SE`, a body
weight — are part of the design and must be stated: add `covariate NARM = <value>` (or
`= [v1, ...]`, one per subject) to [simulation].
```

**One shared check list.** `check_simulation_data` = covariates + kappa weights + residual magnitude, called from every simulate entry point. The two Vec-returning ones cannot signal, so they enforce it as a panic at the same chokepoint `validate_iov_simulatable` already uses. Ordered covariates-first so existing callers' first error is unchanged.

## Alternatives considered

The issue listed three options and recommended the first; the investigation agreed and the implementation follows it.

- **Carry the weight expression and evaluate it against the simulated row** — works only when the weight is a function of columns the simulation already defines, and still needs an answer when it is not. It does not remove the question, it relocates it.
- **Fall back to the fitted data's typical (median) arm size, reported loudly** — silently answers a different question than the one asked. It is still available, explicitly: simulate against the fitted dataset (`--data`, or pass the fitted `Population` to `simulate()`) and the weights come from the rows they were fitted from. That is documented in `simulation.qmd`.

On the guard side, the alternative was to add `check_residual_magnitude` to each entry point individually. Bundling is what stops the subsets from drifting apart again — the drift is the defect, not any one missing call.

## Cross-repo dependency

No public Rust signature changed. ferx-r calls `simulate_with_options_diag` / `simulate_with_uncertainty`, both of which keep their signatures; they now reject a case they previously mis-simulated.

## Breaking changes
- [ ] Public Rust API
- [x] None

`[simulation] covariate NAME = ...` is new syntax, not changed syntax — no existing model file means anything different. The guard additions turn previously-silent wrong output into an error, which is the point of the fix; no working model is affected (all 22 bundled `[simulation]` examples were re-run, see below).

## Numerical validation
- [x] Not applicable — this adds a DSL statement and restores guards that already existed on `fit()`. No estimator, likelihood, gradient or PK solution is touched.

The correctness claims that *are* numerical are pinned as degenerate oracles rather than against NONMEM (NONMEM has no equivalent single object here — it is the simulate side of an already-anchored `fit()` construction, and #1031/#1029 carry the NONMEM anchors for the scoring itself):

- `weight ≡ 1` simulates **bit-identically** to the same model with no `weight =` modifier, on both families (`a_weight_of_one_is_bit_identical_to_no_weight`, `an_arm_size_of_one_simulates_bit_identically_to_an_unweighted_kappa`). Both assert non-vacuity — the κ one needed a dose added to its fixture, since without one every IPRED is `0.0` and the oracle would compare zeros.
- The weight reaches the draw in the right direction: a 4× residual weight widens the deviation from IPRED ~4×; larger arms shrink the between-arm spread of IPRED, which is `κ ~ N(0, γ²/N)` doing what it says.

## Performance

The chokepoint check repeats per uncertainty draw. Both expensive halves are gated on the model actually declaring a weight (`has_weighted_kappa` / `has_custom_ruv_magnitude`), false for every model that does not use one; where true, the pass is linear in the observations the draw is about to simulate anyway. Noted in the comment at the call site.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3371, and 3544 with `--features "survival markov"`)

New coverage:

- **Parser** (`model_parser_tests.rs`): scalar broadcast, per-subject list, length rule checked under either key order, undeclared-name rejection against `[covariates]`, duplicates, bad values, nameless `covariate =`, and the `covariates = ...` prefix trap.
- **Design resolution** (`tests/simulation_design_covariates_tests.rs`, new): per-subject values land on the right subject; a design omitting a referenced weight is refused by name with the fix in the message; an unweighted model needs nothing; an extra covariate the model ignores is allowed.
- **Guards** — these fail on `main`: `simulate_rejects_a_zero_weight_the_way_fit_does`, `the_vec_returning_entry_point_panics_on_a_zero_weight` (both families).
- **Oracles**: as described under Numerical validation.

- [x] Example added inline below (and in `docs/model-file/simulation.qmd`)

<details>
<summary>Weighted residual, end-to-end via <code>--simulate</code></summary>

```toml
[parameters]
  theta TVE0(10.0, 0.1, 100.0)
  theta TVEMAX(6.0, 0.1, 100.0)
  theta TVET50(2.0, 0.01, 100.0)
  omega ETA_E0 ~ 0.09
  sigma ADD ~ 1.0 (sd)

[individual_parameters]
  E0   = TVE0 * exp(ETA_E0)
  EMAX = TVEMAX
  ET50 = TVET50

[structural_model]
  y = E0 - EMAX * TIME / (ET50 + TIME)

[error_model]
  DV ~ additive(ADD) weight = WPSE

[covariates]
  WPSE continuous

[simulation]
  n_subjects = 5
  times      = [1.0, 2.0, 4.0, 8.0]
  seed       = 7
  covariate WPSE = [0.5, 0.7, 1.0, 1.4, 2.0]
```

```
Model: mbma_ruv_ok
Simulating 5 subjects (seed=7)...
Loaded 5 subjects, 20 observations
...
Final OFV = 30.929362
```

Dropping the `covariate WPSE` line now gives the `[simulation]:` message quoted above, instead of "not found in data".

</details>

## Docs
- [x] `docs/` updated — `simulation.qmd` (new syntax in the table and the code block; the false "Covariates in Simulation" paragraph replaced with the real behaviour plus a `#sim-covariates-required` section on why neither silent default is safe), cross-linked from `iov.qmd` and `error-model.qmd` at each `weight =` section
- [x] New pages linked in `docs/_quarto.yml` — N/A, no new page
- [x] `CHANGELOG.md` `[Unreleased]` entry added (one under Added, one under Fixed)

## Checklist
- [x] `cargo clippy` clean (no new warnings on touched code; pre-existing `approx_constant` in `sens/two_cpt_explicit.rs` and `manual_contains` in `validation.rs:2263+` are untouched)
- [x] `Cargo.toml` version bump — not needed (additive)

### Example execution
- [x] All 22 bundled `examples/*.ferx` carrying a `[simulation]` block were re-run through `--simulate` against this branch; none newly hits the design-covariate gate.
- [x] `cargo check --tests` and `cargo check --tests --features "survival markov"` clean.

## Reviewer hints

Worth a close look:

- **`parse_simulation_block`'s new match arm** (`model_parser.rs`). The `covariate`-word match rather than a `starts_with` prefix is deliberate and has its own test; without it `covariates = ...` would parse as a covariate named `s`.
- **Where the `[simulation]` parse moved** (`parse_full_model`). It now runs *after* the `[covariates]` peek so it can validate declared names, but still *before* `[adaptive_dosing]`, so first-error ordering is unchanged. The peek uses `.ok()` and cannot itself error, so nothing was reordered into it.
- **The panic at `simulate_inner_with_draw`.** Failing loud is the right call here specifically because the failure it catches is silent — but it is a panic on a shared path, so it deserves scrutiny. It matches the existing `validate_iov_simulatable` precedent two lines above.

Skimmable: the docs prose, and the mechanical `parse_sim_block` test-helper rename (11 call sites gained a second argument via a one-line wrapper).

## Open questions

**A weighted κ still cannot complete `--simulate` end-to-end, for a reason orthogonal to this PR.** Synthetic `[simulation]` subjects carry no occasion labels, so *any* κ model — weighted or not — simulates fine and then fails at the subsequent fit with `Model declares kappa (IOV) parameters but no occasion labels were found`. Verified as pre-existing on `main` with an unweighted κ.

I did not fix it here because deciding what an occasion *means* for a synthetic subject is a real design call (arm-as-occasion for MBMA, dose-as-occasion for PK IOV) that belongs with #1031/#391, not with this issue. The library path — a real `Population` whose occasions come from the design or the data, which is how ferx-r does trial simulation — works, and is what the new κ oracles exercise. Happy to open a follow-up issue for the `--simulate` occasion gap if you agree it should be one.

Closes #1083.

https://claude.ai/code/session_014Ezut1SKUEWZ9YBEf1jbj4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
+- **`[simulation]` can now state the covariates of the arms it invents (#1083).** `covariate NAME = <value>`
+  — a scalar for every subject, or `= [v1, v2, ...]` with one value per subject — gives the synthetic
+  subjects a covariate value, which a `[simulation]` design previously had no way to express at all.
+  This is what makes trial simulation work for a model-based meta-analysis: `kappa K ~ γ² weight = NARM`
+  and `DV ~ additive(S) weight = WPSE` both read a data column, and the arm a simulation is proposing
+  has no row to read it from. A model that references a covariate the design does not supply is now
+  refused by name, with the fix in the message, rather than reported as a column "not found in data" on
+  a path that has no data file. See `docs/model-file/simulation.qmd`.
+
 - **`ode_method = auto` picks the ODE stepper for you, and is now the default (#978).** An a-priori stiffness probe builds
   the Jacobian `J = ∂f/∂u` of the `[odes]` right-hand side at the state each integration segment
   starts from and reads its fastest mode, `max |Re λ(J)|`; a system above `30` per time unit starts
@@ -175,6 +184,18 @@ section of the SDLC for the versioning policy).
   with no diagnostic from NONMEM (`nonmem_anchor/dose_attr_double_use_{A,B}.ctl`).
 
 ### Fixed
+- **`simulate()` silently removed an arm's residual noise when its `weight = <expr>` column was blank
+  or zero (#1083).** `fit()` has rejected a non-positive residual magnitude since #1029, but no
+  `simulate()` entry point ran that check — each ran its own subset of the model-vs-data checks and
+  this one was in none of them. Because the modifier multiplies the *additive* loading, a zero weight
+  did not blow up: it removed that arm's residual variability entirely, and the simulated observation
+  came back equal to its own IPRED — finite, plottable, and wrong. The weighted-kappa check had the
+  same uneven coverage: present on `simulate_with_options*` but absent from `simulate` /
+  `simulate_with_seed` and from the adaptive-dosing path, where `κ/√W` with a zero `W` evaluates to
+  zero rather than to infinity, so the arm quietly lost its between-arm variability with no `NaN` to
+  trip over. Every simulate entry point now runs one shared list of checks; the two that return a
+  bare `Vec` and cannot signal enforce it as a panic, matching the existing IOV precondition.
+
 - **An estimated lagtime whose lagged dose arrival crossed a time-varying covariate change got a
   silently wrong analytic gradient on ODE models (#1060).** The dose lands at a moving boundary
   `t + ALAG`, and the walk injects the resulting jump as a saltation. The segment ending at that
--- a/docs/model-file/error-model.qmd
+++ b/docs/model-file/error-model.qmd
@@ -351,6 +351,18 @@ The same modifier rides an IOV `kappa` declaration, where it weights the
 `κ ~ N(0, Ω_IOV / N)` — the arm-level random effect of a longitudinal MBMA. See
 [Sample-size-weighted IOV](iov.qmd#sample-size-weighted-iov-weight-).
 
+### Simulating an arm that is not in the data
+
+The weight is a data column. `fit()` always has one — the arm exists, and its
+reported standard error is a column — but a simulated trial frequently does not,
+because the point of simulating one is to propose arms that are not in the data.
+So a `[simulation]` design must state the weight with `covariate WPSE = [...]`,
+and one that omits it is refused rather than defaulted: the modifier scales the
+*additive* loading, so a weight silently read as zero does not blow up — it
+removes the arm's residual noise entirely and the simulated observation equals its
+own prediction. See
+[Covariates in simulation](simulation.qmd#sim-covariates-required).
+
 ## Log-transform-both-sides (LTBS)
 
 For data whose residual error is *multiplicative* (a constant CV across the
--- a/docs/model-file/iov.qmd
+++ b/docs/model-file/iov.qmd
@@ -197,6 +197,16 @@ This is the random-effect sibling of the residual
 [`weight = ...`](error-model.qmd#residual-weighting-weight-...) modifier, and the
 two share syntax and validation.
 
+### Simulating an arm that is not in the data
+
+The weight is a data column, and `fit()` always has one — the arm exists, and its
+size is a column. A simulated trial frequently does not: the point of simulating
+one is to propose arms that are *not* in the data. Arm sizes are therefore part of
+a `[simulation]` design and must be stated with `covariate NARM = [...]`; a design
+that omits one is refused rather than defaulted. See
+[Covariates in simulation](simulation.qmd#sim-covariates-required) for why neither
+silent default is safe.
+
 ## Model File Setup
 
 ```
--- a/docs/model-file/simulation.qmd
+++ b/docs/model-file/simulation.qmd
@@ -14,6 +14,7 @@ The optional `[simulation]` block defines a virtual clinical trial design for ge
   seed       = SEED
   times      = [t1, t2, t3, ...]
   horizon    = T          # TTE only — administrative censoring time
+  covariate NAME = VALUE  # or = [v1, v2, ...], one per subject
 ```
 
 | Key | Aliases | Description |
@@ -24,6 +25,7 @@ The optional `[simulation]` block defines a virtual clinical trial design for ge
 | `seed` | | Random seed for reproducible simulations |
 | `times` | | Continuous observation time points (Gaussian endpoints) |
 | `horizon` | | Administrative censoring time for TTE endpoints (finite, `> 0`) |
+| `covariate NAME` | | Value of covariate `NAME` for the simulated subjects — a scalar, or one value per subject (see [Covariates in simulation](#covariates-in-simulation)) |
 
 A block must define **at least one of `times` or `horizon`**, matched to the
 model's endpoints:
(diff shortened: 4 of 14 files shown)


roninsightrx:
## Code review — `feat(api): let a [simulation] design state the covariates of the arms it invents`

Reviewed commit `ff1ba783` (14 files, +806/−38) at effort level `high`. The core design is sound — the shared `check_simulation_data` bundle and the required-covariate convention are the right calls. Six issues:

### 1. `src/api/run.rs:277` — the new gate bypasses `check_covariates`'s name classification, producing actively wrong advice

`simulation_design_covariates` rejects *every* unsupplied `model.referenced_covariates` name with "add `covariate NAME = <value>` to [simulation]". But `check_covariates` deliberately partitions that same list: eta-shaped names (`ETA*`/`KAPPA*`) get `E_ETA_NOT_DECLARED` with "declare `omega ETA_CL ~ <variance>`" (#989), and `TAD`/`TAFD`/`MACHEPS` get the ODE-built-in scope hint (#1028).

Since #989 an undeclared eta is only a **parse warning** (`model_parser.rs:3824`), so a `[simulation]` model with a typo'd `ETA_CLL` now reaches this gate first (`run.rs:416`, before `simulate_with_options_diag` at `:519`) and is told to declare a random effect as a design covariate. A user who follows that advice gets a model that runs with the eta frozen at a constant — silently zero IIV, the same class of failure this PR exists to remove.

Route eta-shaped and built-in-shaped names to the existing messages before emitting the design-covariate one.

### 2. `src/api/adaptive.rs:210` — `simulate_adaptive` (the programmatic public entry) still runs none of the checks

Only `simulate_adaptive_from_spec` was updated (`adaptive.rs:1088`). `simulate_adaptive` has no `check_covariates` / `check_kappa_weights` / `check_residual_magnitude` at all, before or after this PR, and `run_adaptive_population` does not funnel through `simulate_inner_with_draw`'s panic chokepoint.

So the PR's stated invariant ("one shared list, called from every simulate entry point") is not achieved: a caller of `simulate_adaptive` with a `weight = WPSE` model and a blank/zero weight column still gets an assay draw with the residual noise silently removed. The PR body's coverage table also lists `simulate_adaptive_*` as already having `check_covariates`, which is only true for one of the two.

### 3. `src/api/simulate.rs:920` — the per-draw check evaluates at `default_params.theta`, not the drawn `params`

`check_kappa_weights` and `check_residual_magnitude` both evaluate their expressions at `model.default_params.theta`; `params` is in scope here (it is passed to `validate_iov_simulatable` four lines above). Two consequences:

- Under `simulate_with_uncertainty`, a weight or magnitude expression that depends on θ (both `weight.eval` and `ruv_magnitude.eval_obs` take `theta`) can go non-positive on a *draw* while the guard, pinned to the point estimate, passes — the exact silent collapse this PR fixes survives on the uncertainty path.
- Because the check is draw-invariant, repeating it once per draw can never report anything the up-front call at `:1035` didn't.

Relatedly, the perf comment ("both expensive halves are gated on `has_weighted_kappa` / `has_custom_ruv_magnitude`") omits `check_covariates`, which is ungated and, when `population.covariate_names` is empty (the in-memory-`Population` case, `validation.rs:83`), scans every subject for every referenced covariate on every draw.

### 4. `src/parser/model_parser.rs:6170` — a one-element *list* silently broadcasts, contradicting the documented rule

`match values.len() { 1 => values.resize(n_subjects, values[0]), … }` does not distinguish the scalar form from `= [70]`. With `n_subjects = 3`, `covariate WT = [70]` fills all three subjects with 70 instead of erroring — while `simulation.qmd` states "A list must have exactly `n_subjects` entries — a short one is a parse error, not a silently recycled or zero-filled tail", and the PR body repeats that guarantee.

A user who writes `[400]` meaning "arm 1 only" gets every arm at 400 with no diagnostic. Track whether the value came from a bracketed list and only broadcast the bare-scalar form.

### 5. `src/api/run.rs:304` — unchecked index `vals[i]`

The `n_subjects`-vs-length invariant is enforced only in `parse_simulation_block`; `simulation_design_covariates` re-derives `per_subject` by indexing `vals[i]` for `i in 0..n_subjects` with no guard. `SimulationSpec` is a public struct with public fields, and the function is `pub(crate)`, so any future in-crate caller that builds a spec without going through the parser gets an index-out-of-bounds panic instead of an error. A `vals.get(i).copied()` with an explicit `Err` (or a debug assert) costs nothing.

### 6. `src/api/mod.rs:18` — the `#[allow(unused_imports)]` is applied to the whole re-export list

It is needed only for `check_residual_magnitude` (now used in non-test builds solely through `check_simulation_data` inside `validation.rs`), but it sits on the entire `pub(crate) use validation::{…}` block, suppressing dead-re-export warnings for ~20 names. Any of those going unused later will be invisible. Scoping it to the one item (e.g. a separate `use` line) keeps the signal.

PR #769, opened 2026-07-09
perf(pk): skip redundant base-pass scaling in predict_iov when occasions present

TeunP:
## Why
Follow-up to the post-merge review of #723 (which fixed occasion-less `[scaling]`
on the IOV `simulate()` path). That fix added an unconditional "base pass"
`apply_scaling` over **all** predictions before the per-occasion scaling loop in
`predict_iov`. When a subject's observations carry occasion labels, every
observation is owned by exactly one occasion group (`subject.occasions` is
parallel to the obs grid — fully populated or empty, never partial), so the
per-occasion loop overwrites the base pass entirely. The base pass is therefore
pure redundant work whenever the observations carry occasions — and `predict_iov`
runs on the estimation hot path (IMP / SAEM / FOCEI inner loops, plus the FD
`_plus`/`_minus` pair in the inner optimizer), so on IOV + `ExpressionScale`
models it costs one extra `pk_param_fn` evaluation + `build_obs_scale_array`
allocation per subject-eval, per iteration.

## What changed
Guard the base pass behind **`if subject.occasions.is_empty()`** — the condition
that no observation carries an occasion label:
- no observation occasions (occasion-less `simulate()`, or a subject whose only
  occasion labels are dose-side) → run the base pass with the BSV+κ=0 vector
  (the case #723 added it for);
- observations carry occasions → skip straight to the per-occasion loop, which
  already determines every prediction.

Also skips the redundant `apply_scaling` (+ clone) for dose-only occasion groups
inside the per-occasion loop (they own no observations, so the pass writes
nothing), and drops an unused `raw` clone on the occasion-less path.

**Review correction (2nd commit).** The guard originally keyed on
`occ_groups.is_empty()`. But `iov_occasion_groups` appends *dose-only* occasions
(a dose in an occasion with no co-timed sample) as groups with an empty obs-index
list, so `occ_groups` can be non-empty while no observation is owned by any group.
A subject with dose occasions but no observation occasions would then take the
per-occasion branch and have its predictions returned **UNSCALED** — the #723 bug.
It is unreachable through the current data readers (they gate observation- and
dose-occasion labels on the same column; the #757 DSL rule ties `occasions` to
`obs_times`), so output stayed bit-identical, but the guard now keys on the real
condition (`subject.occasions`), matching the documented invariant.

## Alternatives considered
"Guard on `occ_groups.is_empty()`" — the original form; rejected on review because
`occ_groups` includes dose-only occasion groups that own no observations, so it is
[...]

Issue #655, opened 2026-07-01
feat(scaling): analytic FOCEI/FOCE gradient for Form C readout on IOV subjects


roninsightrx:
## Summary

Follow-up to #650 (analytic Form C `[scaling] y = <expr>` on analytical PK models).

Form C readouts are differentiated **analytically** on the static dose-superposition path
and the time-varying-covariate / oral-infusion event-walk path. **IOV subjects** (`kappa`
declarations, `n_kappa > 0`) currently fall back to the finite-difference gradient when a
Form C readout is present. The prediction is exact (the readout is applied in the f64
IOV predictor, `pk::compute_predictions_iov` → `apply_analytic_readout`), and the parser
emits a warning — but the FOCEI/FOCE gradient for those subjects is FD, not analytic.

## Why it's deferred

The IOV analytic sensitivity walk (`subject_sensitivities_iov` /
`subject_eta_grad_iov_analytical` in `src/sens/provider.rs`) runs the event-driven `Dual2` /
`Dual1` walk over a **stacked** `[η_bsv, κ_group0, …]` axis layout, seeding per-occasion
PK-parameter duals. Routing the compartment-amount state through the readout program there
is analogous to what `run_obs_tvcov` already does, but the stacked-axis seeding and the
per-occasion `CombinedDerivs` sources add enough surface that #650 gated IOV to FD rather
than land it under-tested. The readout is BSV-only (a `kappa` reference is rejected at
parse, #107), so the readout output itself is occasion-independent given the per-occasion
concentration — the missing piece is threading the amount jet through the IOV walk.

## Where FD is currently forced

- `src/sens/provider.rs` — `subject_sensitivities_iov` and `subject_eta_grad_iov_analytical`
  both `return None` when `model.analytic_readout.is_some()` (search for the `#650` gate
  comment "the IOV dual walk does not yet route the amount state through the readout").

## Proposed direction

Mirror the `run_obs_tvcov` / `run_obs_grad_tvcov` injection (#650) inside the IOV walk:
build the per-observation PK-slot dual vector from the walk's per-occasion `PkDual`, form
the central amount `concentration × V`, and evaluate `OdeOutputProgram::eval_output_g` over
the stacked `Dual2` / `Dual1` type — extracting the readout's `(value, grad, hess)` in the
stacked-axis basis, exactly as the concentration is read today. Gate to FD only when a
non-structural readout parameter overflows the structural `PkDual` slots (as the TV-cov path
[...]

Issue #107, opened 2026-05-26
IOV: residual NONMEM gap — simultaneous cross-occasion event ordering + ODE Form-C κ readout


roninsightrx:
## Summary

After #106 replaced the Option-A IOV prediction with a continuous, per-occasion-aware path (`pk::predict_iov`), ferx's IOV likelihood matches NONMEM much more closely (gap at NONMEM's fixed MLE on warfarin IOV went ~37 → ~17 OFV units). Two narrow, related residuals remain — both are cases where per-occasion κ is not fully threaded into a calculation. This issue tracks closing them; it supersedes the carryover portion of #104.

## 1. Simultaneous cross-occasion event ordering (the ~17-unit residual)

When an occasion-ending **observation** and the next occasion's **dose** share a timestamp, ferx's event sort processes the dose before the obs (`Dose < Obs` tie-break in `src/pk/event_driven.rs::kind_order`). Under the end-of-interval parameter convention, the interval *preceding* that timestamp then decays with the **new** occasion's clearance, whereas NONMEM processes records in **data order** (obs row first → the interval uses the *ending* occasion's clearance).

Warfarin IOV hits this exactly: occasion-1's last obs and occasion-2's dose are both at `t = 120`, so the `[96, 120]` interval decays with occasion-2's CL in ferx vs occasion-1's in NONMEM. Because that interval sets the occasion-1→2 hand-off state, the error propagates through all of occasion 2 — accounting for the bulk of the residual ~17 units.

**Fix direction.** Honour data order for simultaneous events. This needs the original per-record sequence retained for ties — currently doses and observations live in separate arrays (`Subject::doses`, `Subject::obs_times`) with no global record index, so the `(time, kind)` sort can't reconstruct input order. Plan: thread a global record-order index from `io/datareader.rs` onto each event and use it as the tie-break (falling back to `kind_order` only when sequence is unavailable). Must not change behaviour for the common case (data already ordered dose-before-obs at a shared time) or the intentional `Reset < Dose` ordering.

**Acceptance.** ferx's FOCEI objective at NONMEM's fixed MLE matches NONMEM to within cross-engine tolerance (the ~0.5-unit band the non-IOV NONMEM cross-checks meet); tighten/retire the residual band in `tests/warfarin_iov_nonmem.rs` (currently `[8, 28]`).

## 2. ODE Form C output reads κ = 0

[...]

PR #1090, opened 2026-08-26
fix(pk): evaluate the analytic Form C readout per occasion under IOV [closes #1079]

### LINKED ISSUE #1079: bug(pk): analytic Form C readout evaluates kappa-carrying individual parameters at kappa=0 under IOV — silently wrong predictions
**Type:** Bug (silently wrong predictions and objective)
**Scope:** `src/pk/mod.rs` (`apply_analytic_readout` / `predict_iov`), `src/sens/provider.rs` (`eval_readout_at_obs`, `run_obs_iov`, `run_obs_iov_eta`), `src/sens/ode_provider.rs` (IOV readout seeding).
**Affects:** an **analytical** (closed-form `pk ...`) model with IOV and a `[scaling] y = <expr>` Form C readout (#650/#655) whose expression reads an individual parameter that carries a `kappa`.
**Found by:** @TeunP, reviewing #1063 — the question was whether the compartment-free readout's per-occasion evaluation *diverges* from the compartment path. It does, and the compartment path is the wrong one.

## The bug

`pk::apply_analytic_readout` builds the readout's individual parameters from
`(model.pk_param_fn)(theta, eta, ...)`, and `predict_iov` calls it with **`eta_bsv`** — the between-subject η with **κ held at 0**. The concentration it is handed *does* carry the occasion's κ (the event-driven walk uses per-occasion PK params), but every individual parameter the readout itself reads is evaluated as though κ were zero.

So a readout that reads a κ-carrying parameter silently uses that parameter's typical value in every occasion.

The existing doc comment says the readout is "BSV-only — a κ reference is rejected at parse", and that is true of an *explicit* `KAPPA_*` in the expression. It does not cover the ordinary case: a κ that enters through an individual parameter, which nothing rejects.

## Reproduction

```
[parameters]
  theta TVCL(3.0, 0.01, 100.0)
  theta TVV(20.0, 1.0, 500.0)
  theta TVBASE(2.0, 0.01, 100.0)
  omega ETA_CL ~ 0.09
  kappa KAPPA_B ~ 0.04
  sigma PROP_ERR ~ 0.04 (sd)

[individual_parameters]
  CL   = TVCL * exp(ETA_CL)
  V    = TVV
  BASE = TVBASE * exp(KAPPA_B)

[structural_model]
  pk one_cpt_iv(cl=CL, v=V)      # <- the ODE twin below replaces only this

[scaling]
  y = central / V + BASE

[error_model]
  DV ~ proportional(PROP_ERR)

[fit_options]
  iov_column = OCC
```

Two 100 mg doses (t = 0, 24), observations at t = 1, 3, 25, 27, occasions 1/1/2/2,
`eta_bsv = [0.1]`, `kappas = [[0.5], [-0.4]]`. Against the hand-written ODE twin
(`ode(states=[central])`, `d/dt(central) = -(CL/V)*central`, same `[scaling]`,
`ode_reltol = 1e-12`):

| t | analytical Form C | ODE twin | rel. diff |
|---|---|---|---|
| 1  | 6.236181504 | 7.533624045 | 1.7e-1 |
| 3  | 5.040770689 | 6.338213230 | 2.0e-1 |
| 25 | 6.315446946 | 5.656087038 | 1.2e-1 |
| 27 | 5.097668164 | 4.438308256 | 1.5e-1 |

12–20% on the prediction.

## Which one is right

The ODE twin. Hand-check at t = 1: both engines agree on the concentration
(4.2366 vs 4.2362, solver tolerance), and the difference is entirely in `BASE` —
the twin uses `BASE(κ = 0.5) = 2·e^0.5 = 3.297`, the analytical path uses
`BASE(κ = 0) = 2`. Same at t = 25 with `BASE(−0.4) = 1.341` vs `2`.

An individual parameter carrying a κ *is* per-occasion by definition, and NONMEM's
`$ERROR` — which the Form C readout is the analogue of — is evaluated per record
with that record's occasion. The ODE path already does this (the readout is
applied inside the integrator with the per-observation parameter snapshot), and so
does the compartment-free path added in #1063. The analytical closed-form path is
the only one that does not.

## Why it survived #650 / #655

The natural readout to test is `y = central / V`, and it **cannot** show the bug
even with κ on `V`: the predictor reconstructs the amount as `conc × V(κ=0)` and
the readout then divides by the same `V(κ=0)`, so the two cancel exactly and the
result is the concentration, which was already correct. (Verified: that model
agrees with its ODE twin to 1e-13.) The bug needs a readout where the κ-carrying
parameter does not cancel — an additive baseline, a second analyte, a bounded
transform.

## Why finite differences do not catch it

The analytic sensitivity path is consistent with the wrong value path:
`eval_readout_at_obs` seeds the readout parameters from the BSV-only snapshot
(`group = None`) precisely to match `apply_analytic_readout(..., eta_bsv, ...)`.
So the gradient is the exact derivative of the wrong function, and a
`Dual2`-vs-FD parity test passes. The only oracle that catches it is a
cross-engine one — the ODE twin, or a hand-computed value.

## Fix

1. **f64 path** — evaluate the readout per occasion in `predict_iov` when the model
   has `n_kappa > 0`, using that occasion's `[η_bsv, κ]`. #1063 already does exactly
   this for the compartment-free case (`pk/mod.rs`, guarded on
   `subject.occasions.is_empty()` rather than on `occ_groups`, for the
   dose-only-occasion reason documented there) — the same loop generalises.
2. **Sensitivities** — seed the readout parameter duals with `group = Some(g)`
   instead of `None` in `run_obs_iov` / `run_obs_iov_eta`, and the ODE IOV
   equivalent. #1063's `sens/algebraic.rs` has the stacked-axis seeding written out
   for the zero-state case; the compartment case is the same layout with the
   concentration jet still carrying its own κ.
3. **Do not double-count.** `central = conc × V` must keep using the *same* `V` the
   concentration was computed with, or the amount reconstruction breaks. This is
   why the `y = central / V` case currently looks right, and it is the thing to be
   careful about: the cancellation is load-bearing today.

## Definition of done

1. The reproduction above agrees with its ODE twin to solver tolerance.
2. A regression test with a **non-cancelling** readout (an additive κ-carrying
   baseline), on both the closed-form and ODE engines, plus the `y = central / V`
   case pinned so the amount reconstruction is not broken while fixing this.
3. `Dual2`-vs-FD parity re-run after the seeding change — it passes today against
   the wrong function, so it must be re-established against the right one.
4. A CHANGELOG `Fixed` entry saying plainly that affected fits were wrong, since
   anyone who ran such a model has an objective to re-check.

## Note

Not a regression from #1063 — the compartment-free path there is correct, and the
contrast is what made this visible. #1063 documents its own per-occasion behaviour
and links here.

roninsightrx:
## Why

Closes #1079. An **analytical** (closed-form `pk ...`) model with IOV and a `[scaling] y = <expr>` Form C readout evaluated the readout's individual parameters at **κ = 0**.

`pk::apply_analytic_readout` builds the readout's individual parameters from `(model.pk_param_fn)(theta, eta, …)`, and `predict_iov` called it once with `eta_bsv` — the between-subject η with κ held at 0. The concentration it was handed *did* carry the occasion's κ (the event-driven walk uses per-occasion PK params), but every individual parameter the readout itself read was frozen at its typical value. A readout reading a κ-carrying parameter — an additive baseline, a second analyte, a bounded transform — was therefore silently wrong, and so was the objective built from it.

The reproduction from the issue (`BASE = TVBASE * exp(KAPPA_B)`, `y = central / V + BASE`), against its hand-written ODE twin:

| t | analytic (before) | ODE twin | rel. diff |
|---|---|---|---|
| 1  | 6.236181504 | 7.533624045 | 1.7e-1 |
| 3  | 5.040770689 | 6.338213230 | 2.0e-1 |
| 25 | 6.315446946 | 5.656087038 | 1.2e-1 |
| 27 | 5.097668164 | 4.438308256 | 1.5e-1 |

The ODE twin is the correct one: both engines agree on the concentration to solver tolerance, and the whole difference is `BASE(κ=0.5) = 3.297` vs `BASE(κ=0) = 2`.

Two things kept this hidden:

* **The natural readout cannot show it.** `y = central / V` reconstructs the amount as `conc × V(κ=0)` and then divides by the same `V(κ=0)`, so the two cancel exactly — with or without a κ on `V`. The bug needs a readout where the κ-carrying parameter does not cancel.
* **FD parity passes.** `eval_readout_at_obs` seeded the readout parameter duals `group = None` *precisely to match* the `f64` path, so the analytic gradient was the exact derivative of the wrong function. Only a cross-engine oracle catches it.

Not a regression from #1063 — that PR's compartment-free path is correct, and the contrast is what made this visible (found by @TeunP reviewing #1063).

## What changed

1. **`f64` path (`src/pk/mod.rs`).** The per-occasion readout loop #1063 added for the compartment-free case now covers every model with `n_kappa > 0`: each occasion group's observations get `apply_analytic_readout(..., [η_bsv, κ_g], ...)`. Guarded on `subject.occasions.is_empty()` rather than `occ_groups` for the documented dose-only-occasion reason (#769). The amount reconstruction stays consistent — `conc × V` uses the *same* occasion `V` the concentration was computed with — so the load-bearing `y = central / V` cancellation is preserved.

2. **Sensitivities (`src/sens/provider.rs`).** `eval_readout_at_obs` seeds the readout PK params from the observation's own IOV event source `(pk, cd, Some(g))` instead of a κ = 0 snapshot, so `run_obs_iov` / `run_obs_iov_eta` differentiate the corrected prediction. That source turns out to be *exactly* the snapshot the readout needs (same occasion, same covariate/`TIME` snapshot, same per-event-vs-per-group gate as `apply_analytic_readout`'s own `per_obs`), so `ReadoutSnapshots` and its per-observation `CombinedDerivs` evaluations are **deleted** rather than re-seeded.

3. **ODE IOV needed no change.** `integrate_tvcov_readout` already resolves each observation's readout against `pk_at_obs[j]`, which is group-seeded. Verified, and now pinned by a cross-engine sensitivity test.

4. **Blast radius: the readout was the lone outlier.** Everything else a κ reaches through an individual parameter was always occasion-correct, because it takes a different route. `predict_iov` builds `dose_params` per dose from *that dose's own occasion* (`combined_for(subject.dose_occasions[d])`), and both engines read the dose attributes straight out of that snapshot rather than rebuilding them — analytic via `pk/event_driven.rs` (`pk_at_dose.iter().map(|p| p.lagtime())`, `bioavailable_doses(subject, pk_at_dose)`), ODE via `ode/predictions.rs` (`resolve_subject_doses_with(subject, &ode.dose_attr_map, |k| &pk_at_dose[k].values)`). So `ALAG` / `F` / `D{n}` / `R{n}` under IOV never had this bug. One call site was wrong, and only that one. (Checked by @TeunP below.)

The parse-time rejection of a **direct** `KAPPA_*` reference in a readout expression (#107) is unchanged — the `[scaling]` eta scope is BSV-only. Its rationale comment is corrected: a κ reaching the readout through an `[individual_parameters]` entry is fine and occasion-correct.

## Alternatives considered

Keeping the κ = 0 convention and documenting it: rejected. It contradicts the ODE engine, the compartment-free path, and NONMEM's per-record `$ERROR` — three independent definitions of the same object — and it fails silently.

Building fresh per-occasion readout snapshots (the shape the issue's fix sketch implies): correct but wasteful. The event sources are already the right snapshots, and reusing them removes a `CombinedDerivs` evaluation per observation *and* makes it structurally impossible for the readout's seeding to drift from the concentration's.

## Cross-repo dependency

No public API change; the behaviour change is in the numbers.

## Breaking changes
- [x] None

Not a source-breaking change, but **numerically breaking for affected fits** — see the CHANGELOG entry: anyone who ran an analytical IOV model whose Form C readout reads a κ-carrying parameter has an objective to re-check.

## Numerical validation

Cross-engine, which is the only oracle that catches this class (see Why). The Form C ODE readout is itself NONMEM-anchored via the free/total protein-binding model in `docs/model-file/scaling.qmd`, so agreement with the ODE twin inherits that anchor — this is the same validation route `analytic_form_c_matches_ode_form_c_binding_readout` (#650) already uses for the non-IOV case.

```
# before — analytic vs its ODE twin, y = central/V + BASE, BASE carrying KAPPA_B
t=1:  6.236181503710916  vs  7.533624045111317   (1.7e-1)

# after
t=1:  agrees to 1e-8 relative; y − central/V == 2·e^0.5 exactly (occasion 1)
                                            == 2·e^−0.4        (occasion 2)
```

- [x] Compared against: the ODE engine (NONMEM-anchored readout path), and hand-computed `BASE(κ_occ)`
- [x] `cargo test --lib`: 3358 passed, 0 failed

## Performance

- [x] Faster on the gradient path, marginally slower on the `f64` path.

The IOV analytic gradient loses one `CombinedDerivs` evaluation per observation (the deleted κ = 0 readout snapshot). `predict_iov` gains `k_groups` readout passes over the observation grid instead of one — the price of the correct answer, and only for models that have a readout *and* κ.

## Tests

- [x] Unit tests added (`cargo test --lib` passes)
- [x] Regression tests that fail without this fix

`src/pk/mod.rs`:
* `analytic_form_c_iov_readout_uses_occasion_kappa_matching_ode_twin` — the issue's reproduction against its ODE twin, plus the decomposition (`y − central/V` must equal `BASE` at *that observation's* κ, and must not collapse to the typical value). Fails before the fix with exactly the issue's numbers (`6.236181503710916` vs `7.533624045111317`).
* `analytic_form_c_iov_identity_readout_cancels_kappa_on_v` — the load-bearing cancellation pinned with a κ on `V` itself.

`src/sens/provider_tests.rs`:
* `iov_form_c_kappa_baseline_readout_provider_matches_fd` — Dual2-vs-FD parity re-established against the corrected function, plus the per-group κ-axis assignment (each observation's `∂y/∂κ` lands on its own occasion's stacked axis and leaks nothing onto the other). Fails if the walk keeps the old `group = None` seeding while the `f64` path is fixed (verified: `0.0` vs `2.10`).
* `iov_form_c_kappa_baseline_readout_tvcov_matches_fd` — same on a TV-covariate subject (the per-event source branch).
* `iov_form_c_kappa_baseline_readout_matches_ode_twin_sens` — closed-form vs ODE-twin value **and** gradient, the sensitivity-level cross-engine oracle.

Existing readout × IOV tests are unchanged numerically (their readout parameters are κ-less); only their comments, which documented the old convention, are updated.

- [x] Not applicable — no new API surface. The model in the issue is the example, and it now predicts correctly.

## Docs
- [x] `docs/model-file/scaling.qmd` — the "under IOV the readout parameters are BSV-only" sentence replaced with the per-occasion rule.
- [x] `docs/model-file/iov.qmd` — the `KAPPA_*` limitation bullet now says what *is* supported (κ via an individual parameter, occasion-correct on both engines) alongside what is rejected.
- [x] `CHANGELOG.md` `[Unreleased]` → `Fixed`, stating plainly that affected fits were wrong.

## Checklist
- [x] `cargo clippy` clean (the 4 `approx_constant` errors in `src/sens/two_cpt_explicit.rs` are pre-existing and untouched by this PR)
- [x] `Dual2` path stays differentiable — the readout jet is seeded from the same `CombinedDerivs` source as the concentration
- [x] No version bump needed

## Reviewer hints

The subtle part is **fix note 3 in the issue**: `central = conc × V` must keep using the same `V` the concentration was computed with. It does — both now come from the observation's occasion source — which is why `y = central / V` still cancels. That cancellation is asserted to `1e-14` rather than bit-for-bit: the multiply/divide round-trip differs by ~1 ULP between `V(κ)` and `V(0)`, so bit-equality was luck, not invariant.

Skimmable: the deletion of `ReadoutSnapshots` (mechanical), and the comment-only updates to the existing `#655` tests.

## Open questions

A **direct** `KAPPA_*` reference in a readout is still rejected at parse (#107). On the closed-form path it would now actually resolve, since the readout is handed `[η_bsv, κ]` — lifting the rejection is a small feature, but it needs the `[scaling]` eta scope widened on both engines, so it is left out of a bug fix. Worth its own issue?


TeunP:
Checked the blast radius on the sibling paths, since the root cause here is a single call site handed `eta_bsv` — the natural question is whether the same κ = 0 convention leaked into the other things a κ reaches through an individual parameter.

It did not. Dose attributes take a different route and were always occasion-correct. `predict_iov` builds `dose_params` per dose from **that dose's own occasion**:

```rust
// src/pk/mod.rs
let occ = subject.dose_occasions.get(d).copied().unwrap_or(0);
pk_params_at_time(model, theta, &combined_for(occ), subject.dose_cov(d), subject.doses[d].time)
```

and both engines read the attributes straight out of that snapshot rather than rebuilding them:

* analytic — `pk/event_driven.rs`: `pk_at_dose.iter().map(|p| p.lagtime())`, and `bioavailable_doses(subject, pk_at_dose)` for `F`
* ODE — `ode/predictions.rs`: `resolve_subject_doses_with(subject, &ode.dose_attr_map, |k| &pk_at_dose[k].values)` for `D{n}`/`R{n}`

So `ALAG`/`F`/`D{n}`/`R{n}` under IOV never had this bug, and the readout really was the lone outlier. Worth a sentence under "What changed" — anyone arriving from #1079 will ask, and "one call site, and only that one" is the reassuring part.

One observation, not a change request.

The PR's own lesson is that FD parity could not see this, because `eval_readout_at_obs` seeded `group = None` *precisely to match* the f64 path. The corollary is sharper than it first looks: **two engines are not automatically two references.** A cross-engine oracle only sees a defect that lives downstream of the point where the engines part.

The readout qualifies — each engine applies it itself (`apply_analytic_readout` vs `integrate_tvcov_readout`), which is why the ODE twin caught this. The per-occasion snapshot above does not: it is built in `predict_iov` *before* the `ode_spec` branch and handed to both arms, so an analytic-vs-ODE comparison would inherit an error in `combined_for(occ)` and agree on the wrong answer. A real limit on what the twin can certify, independent of this PR.

Related: nothing currently pairs IOV with lag under a cross-engine oracle — `ode_lagtime_full_sens_matches_analytical_twin` has lag and `n_kappa = 0`, `iov_analytical_expr_scale_equals_ode_twin` has IOV and no lag. Per the paragraph above, adding one would buy less than it appears to, which is itself the point.

General form of this is on #1074, where the anchor non-degeneracy rule lives.

roninsightrx:
Thanks — both taken.

**Blast radius.** Verified your read and added it to "What changed" as item 4. `predict_iov` builds `dose_params` from each dose's own occasion, and neither engine rebuilds the attributes from it: `pk/event_driven.rs:591` (`pk_at_dose.iter().map(|p| p.lagtime())`) and `:598` (`bioavailable_doses(subject, pk_at_dose)`), `ode/predictions.rs:4406` (`resolve_subject_doses_with(..., |k| &pk_at_dose[k].values)`). "One call site, and only that one" is indeed the reassuring part, and it is the readout's *own* parameter evaluation that made it the outlier — it is the only thing that rebuilt a snapshot instead of reading one.

**The branch-point limit.** Agreed, and it is sharper than the FD-parity lesson: FD parity fails to be independent because the two sides *are* the same path, while a cross-engine twin fails to be independent only over the prefix the engines share. The readout sits after the branch, `combined_for(occ)` before it — so the twin certifies exactly the suffix and nothing above it. That is a limit on this PR's own oracles too, worth stating plainly: the ODE twin here certifies that the readout is applied at the right occasion's parameters, not that `iov_occasion_groups` / `combined_for` assign the right occasion in the first place. That assignment is pinned separately (the `iov_occasion_groups` unit tests, the dose-only-occasion regression), by construction rather than by twin.

I see #1074 already carries the general form, including the bullet added off this review — so nothing further from me there, and no code change from either point.

On the IOV × lag gap: concur it buys less than it looks, for the reason you give — lag enters through `dose_lagtimes`, built before the branch and handed to both arms, so a twin would agree on a wrong lag. Leaving it unwritten rather than adding an oracle that certifies less than its name suggests.

PR #1093, opened 2026-08-26
feat(parser): three-argument clamp(x, lo, hi) in the model DSL [closes #1092]

### LINKED ISSUE #1092: feat(dsl): 3-arg clamp(x, lo, hi)
## Request

Add a three-argument `clamp(x, lo, hi)` to the model DSL, as a follow-up to the two-argument `min`/`max` added in #1030.

## Why

Bounding a readout into an open interval is a routine step for any bounded endpoint — a logit-scale response, a fraction, a probability — and the only way to write it today is the nested form:

```
  SAT   = min(max(ACR20, 0.01), 0.99)
```

This reads poorly at the point where it matters most: the reader has to unpick the nesting to see which bound is which, and the argument order flips between the inner and outer call. `clamp(ACR20, 0.01, 0.99)` states the intent directly. It is also the name Rust itself uses (`f64::clamp`), and the parser already calls `v.clamp(1e-15, 1.0 - 1e-15)` internally inside the `logit` evaluator (`src/parser/model_parser.rs:16186`), so the name is already the house idiom.

## Current behaviour

`clamp(x, lo, hi)` is a parse error. `parse_primary` only admits a second argument for `min`/`max` (`src/parser/model_parser.rs:19684`), so a third argument produces:

```
function `clamp` takes 1 argument, but a `,` was found — only `min(a, b)` and `max(a, b)` take two.
```

Worth noting the adjacent trap, which is an argument for adding this properly rather than letting people improvise: an unrecognised one-argument function name is **not** an error. `parse_primary` turns any `name(arg)` into `Expression::UnaryFn(name, arg)` (`model_parser.rs:19733`), and every consumer falls through to identity for a name it does not know — `eval_expr` at `model_parser.rs:16189` (`_ => v`), the bytecode compiler at `model_parser.rs:16615` (leaves the argument on the stack), and `differentiate_with_chain` at `model_parser.rs:18783` (returns `da`). So a single-argument `clamp(x)` — a plausible typo while reaching for this feature — silently becomes a no-op that fits, converges and gives wrong numbers.

## Suggested implementation

Desugar at the parse site, exactly as `min`/`max` already do, rather than adding a three-argument AST node. `Expression::Conditional` is already evaluated, bytecode-compiled, `Dual2`-differentiated and index-resolved throughout the pipeline, so a nested-conditional desugar inherits all of it and cannot drift from the hand-written form it replaces:

```
clamp(x, lo, hi)  ->  if (x <= lo) lo else (if (x >= hi) hi else x)
```

Concretely, in `parse_primary` (`src/parser/model_parser.rs:19684`):

- extend `is_min_max` to a three-way `is_min_max_clamp`, admitting a third comma-separated argument for `clamp` only;
- build `Conditional(Compare(x, Le, lo), lo, Conditional(Compare(x, Ge, hi), hi, x))`;
- keep the existing arity diagnostics honest — `clamp` with one or two arguments must be a hard error naming the three-argument form, not a fall-through to `UnaryFn`.

Two details worth deciding rather than inheriting:

1. **Argument duplication.** As with `min`/`max`, `x` appears three times in the desugared tree. Same cost as writing the conditional by hand, and the `min`/`max` comment already tells the reader to name a complex guarded expression with an intermediate — but with three occurrences it is worth restating in the docs.
2. **`lo > hi`.** The nested-conditional form silently returns `lo`. A parse-time check when both bounds are literals, and no runtime check otherwise, seems the right balance; alternatively reject only the literal case and document the rest.

## Also needed

- `[derived]` disambiguation: `min`/`max` there also name the row aggregate and the two are told apart by arity (`model_parser.rs:4404`). `clamp` has no aggregate meaning, so it should be unambiguous — worth a test asserting that.
- Tests mirroring the `min`/`max` coverage: parse, eval, bytecode-vs-AST equivalence (`model_parser_tests.rs` has a `bc_vs_ast` helper), and derivative correctness at and across both bounds.
- CHANGELOG entry alongside the #1030 note.
- Docs wherever the two-argument `min`/`max` form is described.

## Context

Came up writing the MBMA book's Case Study 2 chapter, where `min(max(ACR20, 0.01), 0.99)` guards the logit transform of an ACR20 rate before it is fitted. Three occurrences in the manuscript today. Not blocking — the nested form is correct, just unpleasant to read in a text meant to teach the DSL.

roninsightrx:
## Why

Closes #1092.

Bounding a readout into an open interval — a logit-scale response, a fraction, a probability — could only be written as the nested form:

```
  SAT = min(max(ACR20, 0.01), 0.99)
```

which reads poorly exactly where it matters: the reader has to unpick the nesting to see which bound is which, and the argument order flips between the inner and outer call. `clamp(ACR20, 0.01, 0.99)` says it directly. It is also the name Rust uses (`f64::clamp`) and the name the parser already uses internally (`v.clamp(1e-15, 1.0 - 1e-15)` inside the `logit` evaluator), so it is already the house idiom.

The motivating case is the MBMA book's Case Study 2 chapter, which guards the logit transform of an ACR20 rate this way in three places.

There is also a trap worth closing while here. An unrecognised one-argument function name was **not** an error: `parse_atom` turned any `name(arg)` into `Expression::UnaryFn(name, arg)`, and every consumer falls through to identity for a name it does not know — `eval_expr` (`_ => v`), the bytecode compiler (leaves the argument on the stack), `differentiate_with_chain` (returns `da`). So `clamp(x)` — a plausible typo while reaching for this feature — silently became a no-op that fits, converges and gives wrong numbers.

## What changed

**Desugared at the parse site**, exactly as the two-argument `min` / `max` from #1030 are, rather than adding a three-argument AST node:

```
clamp(x, lo, hi)  →  if (x <= lo) lo else if (x >= hi) hi else x
```

`Expression::Conditional` is already evaluated, bytecode-compiled, `Dual2`-differentiated and index-resolved throughout the pipeline, so `clamp` inherits all of it and cannot drift from the nested form it replaces. It works in every expression the DSL parses — `[individual_parameters]`, `[scaling]`, `[odes]`, `[derived]`.

Two details the issue asked to decide rather than inherit:

1. **Argument duplication.** `x` lands in the tree three times, `lo` and `hi` twice each — one copy of `x` fewer than the nested `min(max(...))` it replaces. Same advice as `min`/`max`: clamp a *named* value, not a long expression. Restated in the docs, where the "eight copies" figure for the nested form already lives.
2. **`lo > hi`.** The desugaring would return `lo` for every `x`, silently. With **both bounds literal** that can only be a typo, so it is a parse error. Bounds that are not both literals are left to run — a runtime ordering test would be paid on every evaluation of every clamp.

**Arity diagnostics** are kept honest rather than falling through to `UnaryFn`:

- `clamp` with one, two or four arguments → hard error naming `clamp(x, lo, hi)`.
- A third argument to `min` / `max` now points at `clamp` rather than reporting a missing `)`.
- The existing one-argument-with-a-comma message now names all three forms.

**`[derived]` disambiguation.** `min` / `max` there also name the row aggregate and are told apart by the second argument (a comparison is a row filter). `clamp` has no aggregate meaning and is not in the aggregate dispatch list, so it parses as a plain expression at statement top level and nested alike. Asserted by a test.

## Alternatives considered

A dedicated three-argument AST node (`Expression::Clamp`) would need adding to the evaluator, the bytecode compiler, `differentiate_with_chain`, the index resolver and the generic `PkNum` VM — five places that can disagree with each other, for a form that is exactly two nested conditionals. Desugaring has no such surface.

A **runtime** `lo > hi` check was rejected: it costs an ordering test on every evaluation of every clamp on the hot path, to catch a mistake that is a typo in a model file.

## Cross-repo dependency

No `pub` API change — this is parser-internal DSL surface.

## Breaking changes
- [x] None

`clamp` is new syntax, purely additive. The one behaviour that changes for existing files is that a one-argument `clamp(x)` is now rejected instead of being the silent identity — that was the bug, and any model relying on it was already producing wrong numbers.

## Numerical validation

The comparator here is not NONMEM but the exact form `clamp` replaces: `clamp` is *defined* as the nested `min(max(x, lo), hi)`, so the anchor is bit-equality with it, pinned by test rather than by a fit.

```
# clamp_matches_nested_min_max_everywhere
# clamp(X, 0.01, 0.99) vs min(max(X, 0.01), 0.99), bit-for-bit,
# at x = -1, 0, 0.005, 0.01, 0.5, 0.99, 1.0, 2.0  — including *on* both
# bounds, where the branch taken decides which of two equal values is
# returned.  Also checked against Rust's own f64::clamp at each point.
```

Derivatives are checked against central finite differences of the `f64` evaluator in all three regimes (below `lo`, inside, above `hi`) with the probe steps kept clear of the kinks, and the convention *on* the bounds — where FD cannot speak — is pinned exactly: `x <= lo` and `x >= hi` take the bound branch, so ∂/∂x is 0 on both bounds and 1 strictly between them.

## Performance

Parse-time only. The compiled tree is one conditional shallower than the `min(max(...))` it replaces and holds one fewer copy of `x`.

## Tests

Eight Tier-1 tests in `src/parser/model_parser_tests.rs`:

| Test | Pins |
|---|---|
| `clamp_matches_nested_min_max_everywhere` | bit-equality with `min(max(..))` and with `f64::clamp`, on and around both bounds |
| `clamp_bytecode_matches_ast` | bytecode VM ≡ AST evaluator, literal *and* variable bounds |
| `clamp_dual2_matches_fd_in_every_regime` | `Dual2` value/grad/Hessian vs central FD, all three regimes + a variable lower bound |
| `clamp_derivative_convention_on_the_bounds` | ∂/∂x = 0 on each bound, 1 strictly inside — so a later `<`/`<=` edit fails |
| `test_clamp_in_individual_parameters` | end to end through `parse_full_model` → `pk_param_fn`, both bounds biting |
| `test_clamp_arity_diagnostics` | 1/2/4-argument `clamp` rejected; 3-argument `min` points at `clamp` |
| `test_clamp_rejects_inverted_literal_bounds` | inverted literals rejected; equal bounds legal; non-literal bounds unchecked |
| `parse_derived_clamp_is_unambiguous` | `[derived]` top level ≡ nested, trailing operators honoured |

`cargo test --lib`: 3403 passed, 0 failed.

Inline below rather than a new `examples/` file — this is one expression form, not a model.

<details>

```toml
[individual_parameters]
  # before
  SAT = min(max(ACR20, 0.01), 0.99)
  # after
  SAT = clamp(ACR20, 0.01, 0.99)
```

```
# identical predictions, identical gradients — clamp is defined as the
# nested form, and the tests pin the two bit-for-bit.

# the arity trap that is now an error rather than a silent no-op:
  SAT = clamp(ACR20)
  → `clamp` takes exactly three arguments: `clamp(x, lo, hi)`.

# bounds the wrong way round, which used to return 0.99 for every ACR20:
  SAT = clamp(ACR20, 0.99, 0.01)
  → `clamp(x, 0.99, 0.01)` has its bounds inverted — the lower bound must
    not be above the upper bound.
```

</details>

## Docs
- [ ] No user-visible change

Both pages that describe the two-argument `min` / `max`:

- `docs/model-file/individual-parameters.qmd` — operator table row + the arity/desugaring paragraph.
- `docs/model-file/scaling.qmd` — section retitled, the duplication note extended with `clamp`'s three copies, and the `[derived]` paragraph noting `clamp` has no aggregate spelling.

No new page, so nothing to add to `_quarto.yml`.

## Checklist
- [x] `cargo clippy` clean

`cargo clippy --lib --tests` reports no new findings; the four pre-existing `approximate value of PI` errors in `src/sens/two_cpt_explicit.rs` (test fixtures using `3.14` as a parameter value) are untouched by this PR.

Nothing in `sens/` or `pk/` is edited. The desugared tree is `Expression::Conditional`, which the generic `PkNum` VM already handles — `clamp_dual2_matches_fd_in_every_regime` and `clamp_derivative_convention_on_the_bounds` exercise that path directly.

### ferx-site

This is new DSL syntax, so a ferx-site `model-dsl/` follow-up is warranted — flagged under Open questions rather than claimed here.

### ferx-book

The MBMA book's Case Study 2 chapter is the motivating caller (three occurrences of the nested form) and can be simplified once this lands. Not blocking — the nested form stays correct.

No `examples/*.ferx` uses `clamp` — the syntax is new, and nothing existing changes meaning.

## Reviewer hints

Focus on `src/parser/model_parser.rs` in `parse_atom`, the `(is_min_max || is_clamp)` branch. Two things are subtle:

- **Branch order and the `<=` / `>=` choice.** `x <= lo` is tested first, so with `lo == hi` the lower bound wins, and on either bound the *bound* is returned rather than `x`. Numerically identical either way, but it fixes which side of the kink the derivative comes from — `clamp_derivative_convention_on_the_bounds` pins it.
- **The clone pattern.** `arg` appears three times in the tree, so two clones and one move; `arg2` / `arg3` twice each. Worth a glance that no clone was dropped, since a missed one would be a borrow error rather than a wrong tree.

The rest — tests, docs, changelog — can be skimmed.

## Open questions

Should the inverted-bounds check extend to a *partially* literal case, e.g. `clamp(x, 1.0, SOME_THETA)` where the θ's upper init is below `1.0`? It is checkable at parse time from the bounds declaration, but it reaches into `[parameters]` from the expression parser and would fire on models that are fine at the initial estimate. Left out deliberately; happy to add it if you'd rather be strict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,9 @@ section of the SDLC for the versioning policy).
   conditional, so it differentiates and compiles exactly like the nested form. `clamp` with any
   arity other than three is a parse error — a one-argument `clamp(x)` would otherwise have been
   read as the silent identity — and literal bounds given the wrong way round (`clamp(x, 0.9, 0.1)`)
-  are rejected rather than quietly returning `0.9` for every `x`.
+  are rejected rather than quietly returning a bound for every `x`. One input is deliberately not
+  the nested form: `NaN` fails both bound tests, so `clamp` propagates it into the objective
+  function, where `min(max(x, lo), hi)` would have pinned it to `lo` and let a wrong number fit.
 - **`[simulation]` can now state the covariates of the arms it invents (#1083).** `covariate NAME = <value>`
   — a scalar for every subject, or `= [v1, v2, ...]` with one value per subject — gives the synthetic
   subjects a covariate value, which a `[simulation]` design previously had no way to express at all.
@@ -235,6 +237,11 @@ section of the SDLC for the versioning policy).
   with no diagnostic from NONMEM (`nonmem_anchor/dose_attr_double_use_{A,B}.ctl`).
 
 ### Fixed
+- **A third argument to a `[derived]` row aggregate was dropped in silence (#1092).**
+  `CMAX = max(IPRED, TIME > 0, 99)` computed `max(IPRED, TIME > 0)` — the aggregate form reads only
+  the value and the optional row filter, and never looked at what followed. It is now a parse error
+  that names `clamp(x, lo, hi)`, which is what the extra argument usually means. The same mistake
+  written inside a larger expression was already rejected.
 - **`simulate()` silently removed an arm's residual noise when its `weight = <expr>` column was blank
   or zero (#1083).** `fit()` has rejected a non-positive residual magnitude since #1029, but no
   `simulate()` entry point ran that check — each ran its own subset of the model-vs-data checks and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
+- **Three-argument `clamp(x, lo, hi)` in the model DSL (#1092).** Bounding a readout into an
+  interval no longer has to be written as the nested `min(max(x, lo), hi)`, whose argument order
+  flips between the inner and outer call. Available in every expression the DSL parses
+  (`[individual_parameters]`, `[scaling]`, `[odes]`, `[derived]`); it desugars to the inline
+  conditional, so it differentiates and compiles exactly like the nested form. `clamp` with any
+  arity other than three is a parse error — a one-argument `clamp(x)` would otherwise have been
+  read as the silent identity — and literal bounds given the wrong way round (`clamp(x, 0.9, 0.1)`)
+  are rejected rather than quietly returning `0.9` for every `x`.
 - **`[simulation]` can now state the covariates of the arms it invents (#1083).** `covariate NAME = <value>`
   — a scalar for every subject, or `= [v1, v2, ...]` with one value per subject — gives the synthetic
   subjects a covariate value, which a `[simulation]` design previously had no way to express at all.
--- a/docs/model-file/individual-parameters.qmd
+++ b/docs/model-file/individual-parameters.qmd
@@ -46,6 +46,7 @@ name from quietly reading as zero and collapsing the formula.
 | `sqrt()` | `sqrt(WT)` |
 | `abs()` | `abs(ETA_CL)` |
 | `min(a, b)`, `max(a, b)` | `max(TVCL * WT / 70, 0.1)` |
+| `clamp(x, lo, hi)` | `clamp(ACR20, 0.01, 0.99)` |
 | Parentheses | `TVCL * (WT/70)^0.75` |
 | Comparisons (in `if` conditions) | `WT > 70`, `SEX == 1`, `AGE != 0` |
 | Logical (in `if` conditions) | `&&` (and), `\|\|` (or), `!` (not) |
@@ -56,6 +57,16 @@ conditional (`max(a, b)` → `if (a >= b) a else b`), so they differentiate and
 compile identically to the hand-written form. A one-argument `max(x)` is a parse
 error rather than the silent identity it used to be
 ([#1030](https://github.com/FeRx-NLME/ferx-core/issues/1030)).
+
+`clamp(x, lo, hi)` takes **exactly three** arguments and is the two-sided bound:
+`clamp(ACR20, 0.01, 0.99)` says what `min(max(ACR20, 0.01), 0.99)` says, without
+the argument order flipping between the inner and outer call. It desugars to the
+same nested conditional — `if (x <= lo) lo else if (x >= hi) hi else x` — so on
+each bound the *bound* is returned and the derivative there is zero. Any other
+arity is a parse error, and literal bounds the wrong way round
+(`clamp(x, 0.9, 0.1)`) are rejected rather than silently returning `0.9` for
+every `x`; bounds that are not both literals are not checked
+([#1092](https://github.com/FeRx-NLME/ferx-core/issues/1092)).
 
 ## Continuation Lines
 
--- a/docs/model-file/scaling.qmd
+++ b/docs/model-file/scaling.qmd
@@ -399,7 +399,7 @@ above), the dose-attribute double-use rejection sees through them, and the
 analytic-sensitivity path is unaffected. The one consequence of inlining is that
 an intermediate used twice is *evaluated* twice — which is exactly what writing
 it out by hand costs. Note that `min` / `max` count as two uses each (see below),
-so a two-sided clamp on a named value expands that value eight times — or three
+so a two-sided clamp on a named value expands that value four times — or three
 times written as `clamp(x, lo, hi)`.
 
 ### `min(a, b)`, `max(a, b)` and `clamp(x, lo, hi)`
@@ -423,8 +423,19 @@ On each bound the *bound* is returned, so the derivative there is zero. It puts 
 in the tree three times rather than the nested form's four, but the advice above is
 unchanged — clamp a named value, not a long expression. Any arity other than three
 is a parse error, and literal bounds the wrong way round (`clamp(x, 0.9, 0.1)`) are
-rejected rather than silently returning `0.9` for every `x`; bounds that are not both
-literals are not checked.
+rejected rather than silently returning a bound for every `x`; bounds that are not
+both literals are not checked.
+
+`clamp` differs from the nested pair in one place, deliberately: on a `NaN` input.
+`NaN` fails both `x <= lo` and `x >= hi`, so `clamp` falls through to `x` and the
+`NaN` propagates — into the prediction, and from there into the objective
+function, where it is caught. The nested `min(max(x, lo), hi)` instead returns
+`lo`, because `max`'s `x >= lo` is false for `NaN` and the `else` branch is the
+bound. So if you rewrite an existing `min`/`max` pair as `clamp` around a
+quantity that can go `NaN` — a `log` or `sqrt` of something that dips negative
+mid-search, say — that model was being quietly pinned to `lo`, and will now
+surface the `NaN` instead. That is the intended direction: a silently bounded
+`NaN` is a wrong number that fits.
 
 In `[derived]`, `min` / `max` also name the row aggregate (`max(IPRED, TIME > 0)` is
 "the largest `IPRED` over rows where `TIME > 0`"). The two are told apart by the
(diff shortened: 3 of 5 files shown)


roninsightrx:
## Review (medium effort)

Read the parser hunk against its surroundings (`parse_atom`, the tokenizer's unary-minus folding, `parse_derived_block`, `check_dose_attr_double_use`).

**No blocking bug.** The core desugaring is correct: `p` shadowing through the three `parse_add_sub` calls is right; the RParen/comma arity checks cover the 1/2/4-argument forms; negative literal bounds still reach the inverted-bounds check (the tokenizer folds `-5` after a comma into `Token::Number(-5.0)`, so `clamp(x, 5, -5)` *is* rejected); the `x`-appears-three-times duplication doesn't trip the dose-attribute double-use check (it's set-based); and `clamp` falls through cleanly to the `PerRow` path in `[derived]`.

Three low-severity findings, all now fixed in 9c7768d6.

### 1. `clamp` and `min(max(..))` diverge on `NaN` — the equivalence claim is unqualified

`max(x, lo)` desugars to `if (x >= lo) x else lo`, which returns `lo` for a `NaN` `x`; the outer `min` keeps it. The new desugar's `x <= lo` and `x >= hi` are *both* false for `NaN`, so it falls through to `x` and propagates the `NaN`.

That's the better behaviour — a model with `min(max(f(X), lo), hi)` around a readout that can go `NaN` (a `log`/`sqrt` of something that dips negative mid-search) was being silently pinned to `lo`, i.e. a wrong number that fits, where the `NaN` would have been caught at the OFV. The problem was that the commit message ("cannot drift from the nested form it replaces"), the docs, and the test docstring ("must agree everywhere") all asserted an equivalence that doesn't hold, and `clamp_matches_nested_min_max_everywhere` only probed finite `x`.

Fixed by making the claim true to what the code does rather than changing the code: the desugar comment and `docs/model-file/scaling.qmd` now state the `NaN` convention and why it's the wanted direction, the test is renamed `clamp_matches_nested_min_max_on_every_finite_x`, and a new `clamp_propagates_nan_where_nested_pins_to_lo` pins the divergence (AST and bytecode) as a decision rather than an accident. Note the same branch choice already made the derivative 0 *on* each bound where the nested form passes 1 through — that one was deliberate and pinned, so the two conventions are now cross-referenced from each other.

### 2. The two expansion counts in `scaling.qmd` contradict each other

`docs/model-file/scaling.qmd:402` said a two-sided nested clamp "expands that value eight times — or three times written as `clamp(x, lo, hi)`", while the new paragraph below said clamp "puts `x` in the tree three times rather than the nested form's four". Counting the AST, nested `min(max(x, lo), hi)` has `x` four times and `clamp` three, so the "eight" is wrong (eight is that clamp read from *two* places, which is what the section below actually says). The PR's appended "or three times" is what made the two numbers directly comparable and the inconsistency visible. Corrected to four.

### 3. The new "a third `,` points at `clamp`" guard doesn't reach `[derived]` statement-top-level aggregates

`src/parser/model_parser.rs` — in `[derived]`, `CMAX = max(IPRED, TIME > 0, 99)` takes the aggregate branch (`args.len() != 2`), reads `args[0]`/`args[1]` and drops `args[2]` silently; `close_idx + 1 == tokens.len()` still holds, so the trailing-token check right above it doesn't fire either. Pre-existing, but the PR's stated contract — "a third argument to `min`/`max` now points at `clamp`" — was only true in the `parse_atom` path.

Fixed with an `args.len() > 2` rejection in the aggregate branch, naming `clamp(x, lo, hi)` for `min`/`max` and not for `tmax`, plus `parse_derived_aggregate_rejects_a_third_argument`. Changelog entry added under `Fixed`.

Also corrected, while in the same lines: the inverted-literal-bounds comment, docs sentence and test docstring all said `clamp(x, 0.9, 0.1)` "would quietly return `0.9` for every `x`". It wouldn't — the desugar returns `lo` below the crossing and `hi` above it, never `x`. (The rejection itself is right.)

`cargo test --lib` — 3405 passed, 0 failed. `cargo fmt --check` clean.

PR #1094, opened 2026-08-26
feat(ode): switch the ODE stepper in place when a segment turns stiff [#1080]

roninsightrx:
## Why

Part C of #1080 — the step-level `AutoSwitch` layer.

`ode_method = auto` (Part A, #1066) probes the Jacobian at each segment's **entry** state. For any model whose stiffness is carried by a binding term, that is the state it looks least stiff at: both factors of `KON · C · R` are zero when a dose lands. A depot-absorption binding model reads `|Re λ|max = 10.2` on entry and `7.6e3` an hour later — and `auto` keeps the explicit stepper for the whole span.

That is not merely a slow segment. Measured on that model over 24 h at default tolerances, against a `rodas5p` reference at `reltol = abstol = 1e-12`:

| method | median rel err | max rel err | attempted steps | min-`dt` clamps |
|---|---|---|---|---|
| `rk45` (what `auto` picks today) | **1.43** | 10.6 | 10 000 (budget exhausted) | 0 |
| `rodas4` | 6.8e-8 | 3.2e-6 | 79 | 0 |

**The runtime signal the issue proposed does not see this.** #1080 suggested the step-rejection rate as the trigger for Part C (since `min_step_clamped_steps` is known blind here). Measured across five decades of `KON` on the fast-binding sweep from the #978 review, it is blind in the same direction — the broken cases have a *lower* rejection rate than the healthy ones:

| KON | \|Re λ\|max | median rel err | attempted | rejection rate | clamps |
|---|---|---|---|---|---|
| 0.1 | 11 | 3.6e-7 | 66 | 0.030 | 0 |
| 1 | 110 | 1.0e-7 | 275 | 0.058 | 0 |
| 10 | 1101 | 3.8e-8 | 2408 | 0.051 | 0 |
| 100 | 11010 | **1.7** | 10 000 | **0.026** | 0 |
| 1000 | 110100 | **3.1** | 10 000 | **0.025** | 0 |

So no step-history counter can trigger the switch. The only thing that sees this family is the a-priori discriminator — which is exactly what #1080 said should be an explicit design constraint on anything built here.

Both measurement harnesses are committed as `#[ignore]` tests (`measure_runtime_stiffness_signals_on_the_binding_sweep`, `measure_the_mid_segment_stiffness_a_start_probe_cannot_see`), so the premise can be re-measured rather than re-argued.

## What changed

`auto` re-runs the probe at the state the segment has reached, every 25 accepted steps, and swaps the stepper **in place** when the verdict changes — the segment keeps everything it has integrated and carries on with the other method, rather than restarting or re-solving.

| | median rel err | attempted steps | wall |
|---|---|---|---|
| `auto` before | 1.43 | 10 000 | 1.7 ms |
| `auto` after | 6.7e-8 | 90 | 0.1 ms |
| pinned `rodas4` | 6.8e-8 | 79 | 0.1 ms |

Details:

- **Cadence.** 25 accepted steps. A probe is `n + 1` RHS evaluations plus an `O(n²)` Gershgorin bound (the `O(n³)` eigensolve only runs when that bound cannot rule stiffness out — Part B), so it amortizes to ~1% of a segment. A benign segment finishes in fewer than 25 steps and is **never re-probed**: it pays nothing and its trajectory stays bit-identical to a pinned `rk45` run (pinned by test). Switches per segment are capped at 8.
- **`T`-consistency.** The re-probe reads `PkNum::val` only, like every other comparison a driver makes, so the `f64` prediction and the `Dual1`/`Dual2` sensitivity solve switch at the same accepted step — the analytic FOCE/FOCEI gradient keeps differentiating the trajectory the predictor reports. Tested by asserting identical `OdeSolverStats` between the two.
- **Guard.** A mid-segment escalation *is* an escalation: it joins `auto_stiff_segments` and is guarded exactly like a start-of-segment one — non-finite or clamped ⇒ discard and re-solve with the explicit stepper **pinned** (pinning is also what stops the fallback escalating again). A segment that never left the explicit stepper is deliberately *not* guarded: there the re-solve is the same solve, and discarding would just buy a second copy of it (pinned by a step-count test).
- **Event-time driver.** `solve_ode_until_threshold` switches on the same rule. It needs it more than the dense path: its output *is* the likelihood contribution, so a badly integrated span moves an event time or launders a censoring rather than costing one prediction.
- **New counter** `auto_switched_segments` (one per segment, however many times it changes its mind), reported through the Part B `ode_solver` warning and its `details` payload.
- **New key** `[fit_options] ode_auto_switch` (default `true`, ignored under a named `ode_method`). `false` restores one method per segment, chosen at its start — bit-for-bit (pinned by test).

## Alternatives considered

- **Step-rejection rate / accepted-steps-per-output-point as the trigger** — the issue's own candidate. Rejected on measurement: see the sweep table above. Rejection rate is anti-correlated with the failure; steps-per-output-point moves in the right direction but does not separate (the accurate `KON = 10` case sits at 95 steps/point).
- **Restart the segment on the other stepper** instead of switching in place — throws away the work already done and re-integrates the benign first half, which is what Part C exists not to do.
- **Hysteresis band around the threshold** instead of a switch cap — would make `auto`'s mid-segment verdict disagree with its start-of-segment verdict, i.e. two thresholds for one decision. A count cap bounds the same worst case without splitting the discriminator.

## Cross-repo dependency

No public Rust signature changed; `FitOptions` gained a field with a `Default` (`ode_auto_switch: true`).

## Breaking changes
- [x] None

`ode_method = auto` is the default, so this changes the trajectory of any ODE segment whose stiffness appears mid-way — that is the point of the change, and it moves those segments *toward* the tight-tolerance reference. Non-stiff segments are bit-identical.

## Numerical validation
- [x] Compared against: a tight-tolerance reference solve (`rodas5p`, `reltol = abstol = 1e-12`, cross-checked against a tight `rk45`; the two agree to 1e-13) and against pinned `rodas4`
- [x] Reference values:

```
depot-absorption binding model, KON=100, 24 h, hourly saves, default tolerances
central compartment, relative error vs the tight reference:

# before                                        # after
rk45   median 1.43   max 1.06e1  10000 steps    auto  median 6.74e-8  max 3.12e-6   90 steps
auto   median 1.43   max 1.06e1  10000 steps    (pinned rodas4: 6.75e-8 / 3.16e-6, 79 steps)
```

- [x] Not applicable for a NONMEM anchor: NONMEM has no equivalent knob (LSODA switches internally and does not expose the decision), so the anchor here is a tolerance-refined reference solve of the same system — the standard used for the rest of #1080. The existing NONMEM-anchored ODE fits (cyclophosphamide OFV 3241.708 vs NONMEM FOCEI 3241.721, the TMDD rows) are unchanged: those escalate at the segment start, where nothing about this PR applies.

## Performance
- [x] Faster where it fires — depot-binding segment: 10 000 steps / 1.7 ms → 90 steps / 0.1 ms (17×), while going from 143% median error to 7e-8.
- [x] No impact where it does not: a benign segment finishes before the first re-probe, so it never pays for one, and its output is bit-identical to pinned `rk45`.

## Tests
- [x] Unit tests added (`cargo test --lib` passes, 3389 tests)
  - `auto_switches_in_place_when_a_segment_turns_stiff_mid_way` — including the premise assertions (explicit is wrong here, *and* clamps nothing), so the trigger's justification fails loudly if it ever stops holding
  - `disabling_the_switch_reproduces_the_segment_start_decision_bit_for_bit`
  - `a_benign_segment_is_never_switched` (bit-identical to pinned `rk45`)
  - `the_switch_happens_at_the_same_step_at_f64_and_at_a_dual`
  - `a_stalled_mid_segment_switch_is_discarded_and_re_solved_explicitly`
  - `an_unswitched_explicit_segment_is_not_re_solved`
  - `the_event_time_driver_switches_mid_span`
  - parser: `test_apply_fit_option_ode_auto_switch`, `test_ode_auto_switch_reaches_ode_spec`
  - warning: `a_mid_segment_switch_is_reported_on_the_escalation_note`, `a_switch_is_reported_alongside_an_unclean_solve`
- [x] Two `#[ignore]` measurement harnesses committed alongside them

- [x] Inline below

```toml
[fit_options]
  # default; set false to go back to one stepper per segment, chosen at its start
  ode_auto_switch = true
```

The fit's `ode_solver` warning then says so:

```
W_ODE_SOLVER_ESCALATION_NOTE: ode_method = auto escalated 12 integration segment(s) to a
stiff stepper at the final estimates; ... 9 of them changed stepper part-way through the
segment, because a mid-segment re-probe disagreed with the verdict the segment started on
(ode_auto_switch).
```

## Docs
- [x] `docs/model-file/fit-options.qmd` — `ode_auto_switch` row
- [x] `docs/model-file/ode-models.qmd` — new "When a segment turns stiff halfway through" section, plus the counter list and the `ode_solver` paragraph
- [x] `docs/warnings.qmd` — the `ode_solver` row now mentions the switch clause

## Checklist
- [x] `cargo clippy` clean (no new warnings; the pre-existing `type_complexity` / `neg_cmp_op_on_partial_ord` lints in `solver.rs` are untouched)
- [x] `Dual2` path stays differentiable — the switch decision reads values only, and the `f64`/dual step sequences are asserted identical
- [x] No version bump needed

### ferx-site
- [x] New `[fit_options]` key → `docs/model-file/fit-options.qmd` updated here; ferx-site page needs the same row if it mirrors this table

### Example execution
- [x] `cargo run --release --bin ferx -- examples/bioavailability_ode.ferx --data data/bioavailability_oral.csv` runs clean (OFV −489.4712, no `ode_solver` warning — a non-stiff ODE model, unchanged)

## Reviewer hints

Focus on:

1. **`integrate_dense_g`** — the re-probe block at the end of the accepted-step branch. The placement is load-bearing: it runs after the step is committed *and* after its interpolated soft reads are taken, so a switch cannot disturb the trajectory behind it or the values already saved off it.
2. **`integrate_resolved_g_inner`** — the guard now keys on "did this run ever put a stiff stepper in play" (`ran_stiff`) rather than "did the segment start stiff". The `!ran_stiff ⇒ usable` arm is what keeps an ordinary clamped explicit segment from being solved twice.
3. **Counter attribution** — `auto_stiff_segments` counts a mid-segment escalation once per segment and never double-counts a segment that started stiff or that switched down and back up; `escalation_counted` carries that.

Skimmable: the docs and the measurement harnesses.

## Open questions

- The switch is **bidirectional** (stiff → explicit as well), capped at 8 changes per segment. That is the honest reading of `AutoVern7(Rodas5P())`, and it lets a post-binding tail drop the `O(n³)` cost, but the measured cases here only ever switch upward. Happy to restrict it to one direction if you would rather not carry the down-switch untested by a real model.
- Cadence is a constant (25 accepted steps), not a fit option. It could be exposed, but it is a cost/latency knob with no obvious user-facing meaning — better left as an implementation detail unless a model turns up that needs a different one.

https://claude.ai/code/session_01Juy7XdkXCs4BMSuycDjfNt


roninsightrx:
## Review — `feat(ode): switch the ODE stepper in place when a segment turns stiff`

Reviewed `44677d53`: `src/ode/solver.rs`, `src/types.rs`, `src/parser/model_parser.rs`, `src/api/postfit.rs`, tests, docs.

The core design is sound — the driver owns the stepper, re-probes on an accepted-step cadence, and the guard is keyed on `ran_stiff`. The `T`-consistency argument holds too: `resolve_method` → `fd_jacobian` read `.val()` only, so `f64` and `Dual*` instantiations switch at the same step.

Seven findings; the first two are real defects.

### 1. `src/ode/solver.rs:1614` — high. Pre-switch clamps discard a successful switch.

`stalled = attempt.min_step_clamped_steps > 0` counts clamps taken by the **explicit** stepper *before* the mid-segment switch, but `usable` then discards the whole attempt because `ran_stiff` is true.

Scenario: an `auto` segment starts explicit, the fast mode appears, `dt` collapses and one step force-accepts at `min_dt` around step ~20; the re-probe at step 25 switches to `rodas4` and the rest integrates cleanly to `7e-8`. The good switched trajectory is thrown away and the segment is re-solved with **pinned `rk45`** — i.e. exactly the 143 %-error result this feature exists to avoid — while `auto_stiff_rejected` tells the user the stiff method failed when it never did.

Fix: snapshot the clamp counter at the switch so only post-switch clamps feed `stalled`. (The docs at `fit-options.qmd:38` and `ode-models.qmd` already describe the intended behaviour — the guard fires if the **stiff half** clamps — so this is the code disagreeing with the spec.)

### 2. `src/ode/solver.rs:1125` — medium. Same bug in the event-time driver.

`until_threshold_with` sets `min_step_clamped = true` (~line 1252) during the explicit phase and never clears it across the switch, so `!run.ran_stiff || (… && !run.min_step_clamped)` discards a successfully-switched crossing search and re-runs it `pinned_explicit` from `u_entry`. Here the consequence is a moved event time or a laundered censoring — worse than in the dense path, per the PR's own reasoning.

### 3. `src/api/postfit.rs:1466` — medium, user-visible. Collapsed line-continuations in `switched_clause`.

The literal emits 14-space runs: `…because a␣×14 mid-segment re-probe disagreed with the verdict the segment started on␣×14 (ode_auto_switch).` The new tests only assert `msg.contains("9 of them changed stepper")`, so nothing catches it.

Separately, in the **Warning** branch (line 1546) the clause is appended after "…parameter estimates that produce these dynamics.", where `parts` may only mention clamps — "9 of them" then has no antecedent and reads as "9 of the clamped steps". Give that branch its own wording, e.g. "9 segment(s) changed stepper part-way through".

### 4. `src/ode/solver.rs:932` (and `:1341`) — low. Abort budget not reset on switch.

The `ode_stiff_abort_after` budget (`min_step_clamps` / `clamps`) survives the stepper swap. A segment that spends `budget − 1` clamps on the explicit stepper and then switches gets a nearly-exhausted budget for the stiff phase, so the stiff attempt is aborted and freeze-padded for a stability limit that belonged to the method it replaced. Same fix as #1.

### 5. `src/ode/solver.rs:885` — medium (design / docs). New θ-discontinuity surface, on by default.

The re-probe adds up to `MAX_AUTO_SWITCHES_PER_SEGMENT` extra threshold tests on `|Re λ|max` per segment, each a step-discontinuity in the solution as a function of θ — and the two sides differ by ~143 % on the motivating model, not by solver tolerance.

`fit-options.qmd` explicitly warns that `ode_stiff_abort_after` "makes the objective *discontinuous* in θ … corrupts the finite-difference Hessian the covariance step builds", and therefore ships it **off**. `ode_auto_switch` multiplies the same discontinuity surface by up to 8 decision points per segment and ships **on**.

Concrete: `compute_covariance` perturbs θ by ~1 % relative; a subject with a segment hovering at `|Re λ|max ≈ 30` at a re-probe switches at θ and not at θ+h, and that subject's OFV contribution jumps far above the FD noise floor → wrong SEs. At minimum this deserves the same caveat paragraph the abort budget got.

### 6. `docs/model-file/ode-models.qmd:392` — low (also `src/ode/solver.rs:83`, `fit-options.qmd:38`).

"A benign segment pays nothing: it finishes in fewer than 25 steps, so it is never re-probed" is not true of ordinary segments. The PR's own `a_benign_segment_is_never_switched` integrates 24 h with 24 output points at default tolerances — necessarily ≫ 25 accepted steps, so it *is* re-probed repeatedly (the trajectory stays bit-identical, but the cost claim doesn't hold). Reword to "a segment shorter than 25 steps is never re-probed; longer benign segments pay ~1 probe per 25 steps."

### 7. `src/ode/stiffness_tests.rs:1058` — low. Test doesn't cover the arm it claims to.

`an_unswitched_explicit_segment_is_not_re_solved` sets `auto_switch: false`, which takes the early `return run(method, opts, stats)` in `integrate_resolved_g_inner` — i.e. it pins the *pre-Part-C* path, not the new `!ran_stiff ⇒ usable` arm the doc comment describes. To cover that arm it needs `auto_switch: true` on a clamping segment whose verdict never flips (e.g. `du/dt = u²` over a span short enough that fewer than 25 steps are accepted).

roninsightrx:
Pushed `89a75291` — all seven findings addressed.

| # | Fix |
|---|---|
| 1 | New `OdeSolverStats::stiff_min_step_clamped_steps` attributes each clamp to the stepper that took it; the dense guard's `stalled` test reads that instead of the segment total. |
| 2 | Same in `until_threshold_with`: `min_step_clamped` → `stiff_min_step_clamped`, set only while a stiff stepper is active. |
| 3 | Continuations escaped in `switched_clause`; the Warning branch got its own wording ("N segment(s) changed stepper part-way through"). |
| 4 | `ode_stiff_abort_after` budget reset on a switch, in both drivers. |
| 5 | θ-discontinuity / FD-Hessian caveat added to `fit-options.qmd` and `ode-models.qmd`. |
| 6 | Benign-segment cost claim corrected in the docs, `solver.rs`, and the changelog entry. |
| 7 | `an_unswitched_explicit_segment_is_not_re_solved` now runs with `auto_switch: true`, over a fast rotation whose `Re λ = 0` no re-probe can escalate — so it pins the `!ran_stiff ⇒ usable` arm rather than the pre-Part-C early return. |

Three new tests, each verified to fail against the pre-fix code and pass after:

- `explicit_clamps_before_a_switch_do_not_discard_the_stiff_half`
- `explicit_clamps_before_a_switch_do_not_discard_a_crossing`
- `the_abort_budget_starts_again_when_the_stepper_is_swapped`

plus exact-sentence and no-double-space assertions on both `switched_clause` tests.

One note on the fixture: its stiff phase displaces the state from equilibrium by only `1e-6`. That is deliberate. A larger transient at the switch clamps on *both* steppers, and a stiff half that clamps is a genuine stall the guard is right to discard — so the small perturbation is what separates "the explicit half clamped" from "the stiff half stalled" in a single fixture.

`cargo test --lib`: 3409 passed, 0 failed. `cargo check --tests` clean; clippy adds no new warnings.

roninsightrx:
## Review - latest head 89a75291

Findings:

1. Low - docs/warnings.qmd:90 documents the ode_solver details payload without the two new keys this PR now emits from src/api/postfit.rs:1499 and src/api/postfit.rs:1503. The warning prose says mid-segment switches are reported, but API consumers reading the details table will not know to look for stiff_min_step_clamped_steps or auto_switched_segments. Please add both keys to the ode_solver details row.

2. Low - src/ode/solver.rs:174 still says Auto is only a starting decision and the method is fixed for the duration of one segment. With ode_auto_switch defaulting true, that public rustdoc is now wrong for default Auto. Please update the enum docs to describe the start probe plus optional mid-segment re-probes, or point readers to OdeSolverOptions::auto_switch.

Verification:

- cargo test --lib (3409 passed, 22 ignored)
- cargo check --tests
- gh pr checks: Check, Format, Clippy, core coverage, TTE/CTMM coverage, codecov/patch all passing; slow-tests skipped by workflow.

roninsightrx:
## Re-review — `da8ad7c2`

Checked both findings from the review of `89a75291` and the fixes for them. Both findings were real, and both fixes are correct.

**1. `docs/warnings.qmd:90`** — confirmed. The `ode_solver` `details` row listed 11 keys against the 13 `postfit.rs` emits: `auto_switched_segments` was missing from the original PR commit, `stiff_min_step_clamped_steps` from mine. I updated the JSON payload without updating its published contract. The new row matches `src/api/postfit.rs:1493-1505` key-for-key and in the same order.

**2. `src/ode/solver.rs:174`** — confirmed, and the more consequential of the two. `OdeMethod::Auto`'s rustdoc still read "not a mid-step switch: the method is still fixed for the duration of one segment", which `ode_auto_switch` defaulting to `true` makes flatly wrong on the default path. Stale from the feature commit; I did not grep the enum doc. The replacement is accurate, including the claim that both modes keep dense output, event root-finding and `Dual2` sensitivities — the swap happens between accepted steps, after that step's interpolated reads are taken, and `the_switch_happens_at_the_same_step_at_f64_and_at_a_dual` pins the `Dual` half of it.

**Also checked**

- Swept the tree for other surviving start-only claims. The remaining hits (`types.rs:5495`, `fit-options.qmd:38`, `CHANGELOG.md:72`, `solver.rs:282`) all describe what `ode_auto_switch = false` does, which is correct.
- The new intra-doc link `OdeSolverOptions::auto_switch` resolves. The rustdoc errors that remain in `solver.rs` (`solve_ode_dense`, `Stepper`, `AUTO_SWITCH_PROBE_INTERVAL` → private items) all pre-date this PR and are not worth widening scope for here.
- `cargo test --lib ode::`: 259 passed, 0 failed.

No further findings — this looks ready to merge from my side.

PR #1095, opened 2026-08-26
feat(parser): theta level blocks for unstructured effects [closes #1064]

### LINKED ISSUE #1064: feat(dsl): factor / indexed theta vectors, and the per-subject parameter sparsity that makes hundreds of them estimable
**Type:** Feature (DSL + estimation machinery)
**Scope:** `src/parser/model_parser.rs` (`[parameters]` / `[individual_parameters]`), `src/types.rs` (`ModelParameters`, theta naming/packing), `src/sens/` (per-subject axis sets), `src/estimation/` (outer gradient scoping, covariance), `src/io/output.rs` (compact estimate reporting).
**Motivated by:** @billdenney on #811 — model-based meta-analysis with an unstructured placebo effect. Part of the MBMA epic #1032; independent of, and useful beyond, the compartment-free structural model (#811 / #1063).

## The ask, in the user's words

> I can often hit hundreds of unstructured-placebo parameters in an MBMA model. (Think 100 studies \* 8 time points each — that's common and not the highest count I've seen.)
>
> One key issue is making it easy to define those parameters. Typing 800 values is no fun, and I've written multiple "replicate all these values for me" types of preprocessors for models in the past.

An **unstructured placebo model** gives every timepoint in every study its own fixed effect, so that no parametric placebo time-course can bias the drug effect. It is a standard MBMA strategy (Boucher & Bennetts Part II lists it, alongside correlated residual error, as one of the two accepted fixes for within-study correlation). The individual estimates are a nuisance — the point is what they *absorb*.

## Why this is not just a declaration-syntax problem

Declaring 800 thetas by hand is the visible complaint, but a model that merely *parses* with 800 thetas will not fit. Verified against the current tree:

| Wall | Where | At n_theta = 800 |
|---|---|---|
| Declaration | 800 `theta` lines | the stated complaint |
| Parameter slots | `PkParams` is a fixed `[f64; MAX_PK_PARAMS]`, `MAX_PK_PARAMS = 128` (`src/types.rs:390`) | 800 separate individual parameters do not fit the layout at all (`ode_param_slots` errors) |
| Optimizer | `Optimizer::Auto` resolves to **BOBYQA** whenever no analytic outer gradient is available (`src/types.rs:6179`) | BOBYQA's interpolation set is (n+1)(n+2)/2 = **321,201** points |
| Analytic gradient | `Dual2<M>` is O(M²) per operation; the dispatch ladders cap at `MAX_ODE_AXES = 24` (96 for ODE+IOV) | M = 800 → 640,000 lanes per operation |
| Covariance | default `MATRIX=R` is an FD-of-OFV Hessian that **reconverges every subject's EBEs at each stencil point** (`src/estimation/covariance.rs:631`) | n(n+1)/2 ≈ **320,400** reconverged OFV evaluations |

So this issue is two things at once: a way to *write* the parameters, and the machinery that makes them *estimable*.

## The structural fact that makes it tractable

An unstructured effect is **indexed**: for observation row *r*, exactly one of the 800 parameters is active — the one for `(study(r), time(r))`. Two consequences:

**1. The design is sparse.** In the usual MBMA data mapping (`ID` = study, treatment arm = occasion / BTAV per #1031), a subject touches exactly *its own* 8 thetas. So:

- the **per-subject** analytic dual width is ~8 + the shared structural parameters — comfortably inside the existing 24-axis ladder. The 800-parameter model is an 8-axis-per-subject model; nothing needs widening to 800.
- an FD outer gradient scoped the same way costs about 8 full-population passes rather than 800, because perturbing θ_k only moves the subjects that touch it.
- the score cross-product and the Hessian are block-sparse.

**2. The effect is linear** and governs a disjoint set of rows, which is what makes the follow-up issue (profiling it out entirely) possible. Out of scope here.

### Part A — factor / indexed theta declarations

Data-driven form (the one that answers "replicate these for me"):

```
[parameters]
  theta PLACEBO ~ factor(STUDY, TIME)(0.0, -10.0, 10.0)
```

One theta per **observed** combination of the named data columns, discovered when the data is bound; the `(init, lower, upper)` triple broadcasts to every level. Level names are recorded so estimates report as `PLACEBO[STUDY=7,TIME=4]`, and the level → index map is stored on the fit so a later `predict()` / `simulate()` binds identically.

Explicit form, the primitive underneath, for users who already have an index column:

```
[parameters]
  theta PLACEBO[800](0.0, -10.0, 10.0)
[individual_parameters]
  PL = PLACEBO[PLA_IDX]           # PLA_IDX is a data column, 1-based
```

Either way `PL` is **one** individual parameter occupying **one** `PkParams` slot — a gather, not 800 slots. That removes the layout wall outright.

`ferx-r`'s planned `ferx_mbma_data()` (FeRx-NLME/ferx-r#298) is the natural place to build the index column for the explicit form.

**Identifiability must be explicit.** `factor(STUDY, TIME)` alongside any intercept is rank-deficient — the levels sum to the intercept. Options, one of which must be chosen rather than defaulted silently: drop a reference level (`factor(..., ref = first)`), impose sum-to-zero, or require the model to declare no intercept. Getting this wrong produces a flat direction that the optimizer wanders along and a singular covariance step, which is exactly the class of silent failure this codebase treats as a bug.

### Part B — per-subject active-parameter sets

A gather has exact, statically-known sparsity: ∂PL/∂θ_k = 1 if k = idx else 0. Register, per subject, which thetas it touches, then use that set to drive:

- the analytic sensitivity dual width (per subject, not per model),
- FD outer-gradient scoping — perturb θ_k, re-evaluate only the subjects in its active set,
- the block structure of the score cross-product and the Hessian.

This is the part that turns A from writable into fittable, and it is a **generic** capability: any model where a parameter is read by a subset of subjects benefits, not only MBMA.

### Part C — guardrails that must land with it

- **Optimizer.** BOBYQA must refuse, or auto-switch, above some dimension rather than silently attempting a 321k-point interpolation set. Related to the defaults discussion in #997.
- **Covariance.** The FD-of-OFV R matrix is not viable at this size. `covariance_method = s` (cross-product, already implemented as `CovarianceMethod::CrossProduct`) is the workable path; the default needs to route there, or say clearly why it will not run.
- **Output.** 800 estimates need a compact table in the sdtab / fit YAML, not 800 top-level keys. Nuisance parameters should be reportable separately from the structural ones.

## What a user can do today, and how much this saves

Because a compartment-free model's objective (#811) is one expression per row, the naive path is not hopeless: with `optimizer = nlopt_lbfgs` and `covariance_method = s`, a dense FD gradient over 800 thetas is ~1600 full-population passes per iteration, dominated by the inner EBE loop — order minutes, not hours. Part B takes that to ~8 passes. So A alone is already usable, and B is the difference between "tolerable" and "unremarkable".

## Definition of done

1. `theta NAME ~ factor(COL, ...)` expands to one theta per observed level at data-bind time, with broadcast init/bounds, recorded level names, and an explicit identifiability convention.
2. `theta NAME[N]` + `NAME[IDX_COL]` gather parses and evaluates, occupying one parameter slot.
3. Per-subject active-parameter sets are computed and consumed by the outer gradient (FD and analytic) and by the covariance step; a unit test asserts the gradient is bit-comparable to the dense computation on a small model, and that the analytic dual width tracks the per-subject set rather than `n_theta`.
4. A model with several hundred factor levels fits end to end in a test (Tier 3, `slow-tests` gated), with a sane optimizer/covariance route and no silent BOBYQA attempt.
5. Estimates for factor levels appear in a compact form in both output paths, and the level → index map round-trips through `predict()` / `simulate()`.
6. `docs/model-file/parameters.qmd` documents both forms, the identifiability convention, and the guidance on optimizer/covariance at scale.
7. Validation: the effect of an unstructured placebo model on the drug-effect estimate is compared against the equivalent NONMEM `$PRED` model with hand-written `THETA(n)` levels, on a dataset small enough to write both.

roninsightrx:
## Why

From @billdenney on #811, via the MBMA epic #1032:

> I can often hit hundreds of unstructured-placebo parameters in an MBMA model. (Think 100 studies \* 8 time points each — that's common and not the highest count I've seen.)
>
> One key issue is making it easy to define those parameters. Typing 800 values is no fun, and I've written multiple "replicate all these values for me" types of preprocessors for models in the past.

An **unstructured placebo model** gives every timepoint in every study its own fixed effect, so no parametric placebo time-course can bias the drug effect. It is a standard MBMA strategy. The individual estimates are a nuisance — the point is what they *absorb*.

Declaring 800 thetas by hand is the visible complaint. The deeper one is that a model which merely *parses* with 800 thetas will not fit: `PkParams` is a fixed `[f64; MAX_PK_PARAMS]`, so 800 separate individual parameters do not fit the slot layout at all, and the default optimizer and covariance step are both tuned for the handful-of-parameters models everything else in this engine is.

## What changed

The structural fact that makes it tractable: an unstructured effect is **indexed**. For observation row *r*, exactly one of the 800 thetas is active — the one for `(study(r), time(r))`. So it is a **gather**, not 800 parameters: one `PkParams` slot, one contiguous θ block, one index per record. That removes the layout wall outright.

**Part A — two declaration forms.**

The explicit primitive:

```
[parameters]
  theta PLACEBO[800](0.0, -10.0, 10.0)
[individual_parameters]
  PL = PLACEBO[PLA_IDX]          # PLA_IDX is a 1-based data column
```

and the data-driven form, which answers "replicate these for me":

```
[parameters]
  theta PLACEBO[STUDY, TIME](0.0, -10.0, 10.0)
[individual_parameters]
  CL = TVCL * exp(ETA_CL) + PLACEBO     # index implicit
```

a level block declares one theta per **observed** combination, so an unbalanced design does not acquire cells nobody measured, and levels are named after their combination (`PLACEBO[STUDY=7,TIME=4]`).

**Binding.** A level block's level count is a property of the *dataset*, so the file entry points read the data, discover the combinations, and re-parse with the counts in hand (milliseconds). That keeps every compiled closure built once against the real θ vector rather than rebuilt in place afterwards. The index reaches the gather as a synthesized per-record covariate (`__level_NAME`), so the existing time-varying-covariate plumbing carries it and there is no parallel path — and when the index is constant within a subject that machinery is left alone entirely. `fit()` on a model that never got bound refuses and names both ways out.

**Identifiability.** Per @TeunP's comment on the issue, this cannot be decided by scanning for a fixed intercept: `[STUDY, TIME]` is rank-deficient against a **random** effect at the same or coarser grouping too — that study's η *is* the mean of its own timepoint levels — and that configuration is the model this feature exists to serve. The check reads both sides:

- *model side*: does some `[individual_parameters]` assignment combine a read of the block with an η? Taint propagates through intermediate assignments, so the idiomatic `BASE = TVCL + PLACEBO` / `CL = BASE * exp(ETA_CL)` is recognised.
- *data side*: do the block's leading columns partition the subjects?

Both true → sum-to-zero **within** each leading-column group, leaving that group's η to carry its mean. Otherwise global sum-to-zero. `contrast = sum_to_zero | sum_to_zero_within | ref | none` overrides; a choice that leaves a nested group's mean free while an η also carries it is a **hard error**. Sum-to-zero is implemented as contrast coding, so the packed optimizer needs no constraint machinery.

**Part C — guardrails.**

- **Pre-fit index check.** An out-of-range, non-integral, or missing gather index is named before the fit starts. The evaluator's own guard is `NaN` — deliberately not `0.0`, since `x/0` already underflows to `0.0` here — but a NaN prediction reports as "the fit diverged", not "your `PLA_IDX` column is 0-based".
- **Optimizer.** `Optimizer::Auto` had no dimension awareness. Above 64 free coordinates it now resolves to `nlopt_lbfgs` rather than BOBYQA, whose interpolation model grows quadratically in the parameter count. An explicit `optimizer = bobyqa` is honoured, with a warning.
- **Covariance.** Above 100 free coordinates a *defaulted* `covariance_method` routes to the score cross-product; an explicit `covariance_method = r` is honoured and told what the `n(n+1)/2` reconverged stencil costs.
- **Output.** A block over 20 levels leaves the estimate table for a compact summary, with its independently estimated contrast coefficients written to the fit YAML as a one-line-per-coefficient `theta_blocks:` list rather than hundreds of top-level keys. Note what that does *not* include: a sum-to-zero or reference level is **derived**, has no θ of its own, and so appears in neither. The complete label set — dependent levels included — is on `theta_level_map()`; the YAML block is labelled `n_free_coefficients` so it cannot be misread as the level count.

**Naming.** The construct was first spelled `theta PLACEBO ~ factor(STUDY, TIME)(...)`. Two problems with that, both fixed here. `factor` risks reading as an R *type coercion* — that PLACEBO is estimated as a categorical thing rather than as N continuous fixed effects. And `~` means "is distributed as" in every other place the DSL uses it (90 occurrences across omega/kappa/sigma and `DV ~ proportional(...)`), so putting a coercion-shaped word on its right made the misreading almost inevitable. The brackets already mean "index into this block" at every reference site, so reusing them removes the keyword and the `~` overload at once, and makes declaration and use match. `by(...)` was rejected because in a mixed-effects tool it reads as random-effect grouping and so suggests an η; `unstructured(...)` because it collides with unstructured covariance / `BLOCK` omega.

**Out of scope, follow-ups to file:** Part B (per-subject active-θ sets scoping the outer gradient and covariance — a speed win, not a correctness fix), the Tier-3 several-hundred-level convergence test, and the NONMEM `$PRED` anchor.

## Alternatives considered

- **Rebuilding `pk_param_fn` (and its siblings) in place after binding**, instead of re-parsing. Re-parsing costs milliseconds and touches nothing; the in-place rebuild would have needed the individual-parameter ASTs retained on `CompiledModel` and every derived closure regenerated in lockstep.
- **Expanding a level block into N individual parameters**, instead of a gather. This is the design the issue rules out: `MAX_PK_PARAMS = 128`, so it does not reach 800 at all.
- **Enforcing sum-to-zero as an optimizer constraint** rather than contrast coding. Contrast coding is exact, needs no packing changes, and makes the dependent level's derivative fall out of the same symbolic machinery.
- **Excluding gather models from the analytic sensitivity scope** (what I planned). Turned out unnecessary — see below.

## Cross-repo dependency

`bind_theta_levels` and `theta_level_map` are newly `pub` at the crate root. `ferx-r` needs them only if it wants to bind in-memory populations; the file entry points it already calls bind on their own. `ferx_mbma_data()` (FeRx-NLME/ferx-r#298) is the natural place to build the index column for the explicit form.

## Breaking changes
- [x] Public Rust API (`api.rs` / `types.rs`) — **additive**: `CompiledModel::theta_blocks()` and `::free_packed_dim()` (methods, not fields), `FitOptions.covariance_method_set`, `ThetaBlocks` / `LevelBlockDecl` / `LevelContrast` / `LevelBinding`, `bind_theta_levels`, `theta_level_map`. The block metadata deliberately lives behind the existing opaque `IndivParamPartials` rather than as a new `CompiledModel` field: `CompiledModel` is public and not `#[non_exhaustive]`, so a new required field would break every external struct literal — a constraint the repo already records at `src/types.rs`. The first revision of this PR got that wrong and is corrected in `0e414c16`.
- [ ] `.ferx` model file format — additive syntax only; every existing model parses unchanged.
- [ ] None

## Numerical validation

- [x] Full `cargo test` green after merging `main` — **4097 passed, 0 failed**.
- [x] **`Dual2`-vs-FD parity** for the new gather kernel, per CLAUDE.md — `sens::provider::tests::theta_gather_sens`, on both a subject-constant and a per-observation index, through the existing `check_full_provider_vs_fd` harness.
- [x] **Degenerate oracle** — a one-level block under `contrast = none` is bit-identical to the plain scalar `theta` it degenerates to.
- [x] **Exact sum-to-zero** — levels sum to `0` to `1e-12`, globally and within each group.
- [ ] NONMEM `$PRED` anchor with hand-written `THETA(n)` levels — deferred to the follow-up alongside the Tier-3 convergence test.

Two things the tests corrected against the issue's assumptions, both worth reading:

1. **A gather is analytically differentiable.** `∂f/∂θ_k` through it is exact under `Dual2` — the level index is integer-valued data, so its jet is empty and taking `.val()` loses nothing. A *small* level-block model therefore keeps the exact analytic gradient rather than falling back to FD; only the axis count gates it. The planned scope exclusion was dropped.
2. **Past the `1..=MAX_SCALE_AXES` dual ladder the provider declines cleanly** and the subject finite-differences — it does not truncate to the first 24 axes. Pinned by a test, since a truncation would be a wrong gradient nothing else in the tree could contradict.

And one premise of my own I had to correct: a gather is sparse in the **individual parameter**, not in the prediction. Once a gathered parameter feeds a compartment model, a later observation depends through the state on the levels that applied before it. The compartment-free (`$PRED`-shaped) MBMA model this serves has no such memory, which is why the design is sparse *there* — and that is the assumption Part B will rest on.

## Performance
- [x] No significant impact expected. A model with no vector/level block short-circuits on `ThetaBlocks::is_empty()` at every consumer, the parse-time table is a thread-local that stays empty, and no existing bytecode gains an op.

## Tests
- [x] Unit tests in the parser (11 — declaration, gather eval, NaN policy, literal folding, bare-reference rejection), `api::levels` (22 — binding, level discovery, both contrasts, the identifiability errors, index-column synthesis), `api::validation` (7 — the pre-fit index check), `types` (10 — the scale guards), `sens::provider` (5 — the FD parity and sparsity pins).
- [x] Tier 2 in `tests/theta_level_blocks.rs` (10) — the full file path: binding, unbalanced designs, both index errors, `predict()` on a bound model and its refusal of an unbound one, binding determinism, and `--simulate`.
- [x] **The `--simulate` test found a real bug** and is the reason it was worth writing rather than reasoning that the binding "should" work: the #1083 design-covariate check demanded `__level_PLACEBO` be stated in `[simulation]`, which made every level-block model unsimulatable and named a reserved-prefix column as the fix. Fixed in the last commit.

## Example

```
[parameters]
  theta TVCL(2.0, 0.001, 20.0)
  theta PLACEBO[STUDY, TIME](0.0, -5.0, 5.0)
  theta TVV(10.0, 0.1, 500.0)
  omega ETA_CL ~ 0.04
  sigma PROP_ERR ~ 0.05

[individual_parameters]
  CL = TVCL * exp(ETA_CL) + PLACEBO
  V  = TVV

[structural_model]
  pk one_cpt_iv(cl=CL, v=V)

[error_model]
  DV ~ proportional(PROP_ERR)
```

With two studies sampled at TIME 1/4/12, and an η at the study level, this binds to six levels under **within-study** sum-to-zero — four free θ:

```
--- THETA Estimates ---
Parameter            Estimate           SE      %RSE
TVCL                 2.041237     ...
PLACEBO[STUDY=1,TIME=1]   ...
PLACEBO[STUDY=1,TIME=4]   ...
PLACEBO[STUDY=2,TIME=1]   ...
PLACEBO[STUDY=2,TIME=4]   ...
TVV                 10.117... 
```

`PLACEBO[STUDY=1,TIME=12]` and `PLACEBO[STUDY=2,TIME=12]` are each their study's sum-to-zero contrast, so they carry no θ of their own. At 800 levels the block leaves the table for:

```
--- THETA BLOCKS ---
PLACEBO  798 levels   min -1.8412  median 0.0134  max 1.9077
```

## Docs
- [x] `docs/model-file/parameters.qmd` — new "Theta vectors and level blocks" section: both forms, the identifiability table and its rationale, and the optimizer/covariance guidance at scale.
- [x] `CHANGELOG.md` `[Unreleased]` — one `Added` and three `Changed` entries.
- [ ] New pages linked in `docs/_quarto.yml` — N/A, the section lives on an existing page.

## Checklist
- [x] `cargo clippy` clean (no new warnings in the touched files)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the gather is implemented once in the generic evaluator and once in the `PkNum` bytecode path, with the FD parity test above

## Review round 1

Six findings, all legitimate, fixed in `0e414c16`. Three were things this PR got wrong rather than incomplete:

| # | Finding | Resolution |
|---|---|---|
| 1 | The auto within-group contrast test only required the leading columns to be *constant within* a subject, so `[ARM, TIME]` with several subjects per arm chose `sum_to_zero_within` and removed a real degree of freedom — those subjects' ηs cannot carry an arm mean. | The test now requires the leading tuple to identify subjects **one-to-one**; shared groups take global sum-to-zero. |
| 2 | `ferx check` never bound data-driven blocks, so it reported `E_MISSING_COVARIATE` for the reserved `__level_PLACEBO` and disagreed with `fit()`. | `validate_model_file` binds before the data-dependent checks; binding failures surface as `E_THETA_LEVEL_BINDING`. Regression test pins the reproduction. |
| 3 | The derived contrast level ignored the declared init/bounds/FIX — a fixed nonzero init made it `−(N−1)·init`, not the broadcast value. | A constrained block now requires `init = 0` and bounds spanning 0; the docs define bounds as applying to the free coefficients. `contrast = none` keeps full broadcast semantics. |
| 4 | `free_packed_dim` counted a dense triangular Ω regardless of FIX flags, structural zeros, or diagonal structure — 14 fixed diagonal ηs read as 108 free coordinates, silently flipping `optimizer = auto` and the default covariance estimator on unrelated models. | Now counts the exact `packed_fixed_mask` / `omega_structural_zero_mask` coordinates the optimizer itself uses. |
| 5 | Adding `CompiledModel.theta_blocks` was a public source break, not additive. | Metadata moved behind `IndivParamPartials`; exposed as `CompiledModel::theta_blocks()`. |
| 6 | Compact output reported free coefficients as the level count and omitted derived levels, contradicting this PR's own claim. | Output relabelled to `n_free_coefficients` / `coefficients`; complete labels retained on `theta_level_map()`. |

**Finding 6 is addressed by correction, not by restoration** — worth a reviewer's eye. The derived level's *value* is still absent from the fit YAML; it is reconstructible (minus the sum of the block's coefficients under sum-to-zero, exactly 0 under `ref`) and its label is on `theta_level_map()`, but if you want `n_levels` and the derived value in the YAML itself, say so and I will add them.

## Reviewer hints

Focus on:

- **`src/api/levels.rs`** — `resolve_contrast` / `assign_groups`. This is where the identifiability decision is made, and it is the part with real modelling judgement in it. The rule is stated in the module docs and in `docs/model-file/parameters.qmd`; if you disagree with the default, that is the place to say so.
- **`build_level_rules`** (`model_parser.rs`) — the level → θ map. The contiguity requirement on contrast groups is asserted rather than assumed.
- **`eval_gather` / `gather_g`** — two copies of the same lookup, deliberately (`f64` and `PkNum`). They must agree; the FD parity test is what holds them together.

Skimmable: the `Expression::ThetaGather` arms added to the AST walkers. The compiler found every exhaustive one; the real risk was the walkers with a `_ =>` catch-all, which compile fine while silently skipping the gather's index expression — those are the ones worth a second look. (An earlier revision also touched 28 hand-built `CompiledModel` fixtures; that churn is gone now that the metadata sits behind `IndivParamPartials`.)

## Open questions

1. **The `Auto` contrast rule.** I default to within-group sum-to-zero only when the block shares a scale with an η *and* the leading columns partition the subjects. The η test matters: without one, within-group sum-to-zero would force every study's mean placebo effect equal, which is a modelling assumption rather than a normalization. But the "shares a scale" detection is a taint walk over `[individual_parameters]`, which will miss a model that combines the block and the η somewhere else (`[scaling]`, say). Such a model gets global sum-to-zero and no error. Is the explicit `contrast =` override enough of an escape hatch, or should an ambiguous case refuse?

2. **Dose and EVID=2 rows under a a level block block** take the level of the most recent observation at or before them (LOCF), since a level is a property of an *observation*. That only matters for a model whose gathered parameter also drives dosing dynamics — not the unstructured-placebo case, which reads it in the prediction. Flagging it because it is a convention, not a derivation.

https://claude.ai/code/session_014Ezut1SKUEWZ9YBEf1jbj4


roninsightrx:
Review findings:

1. **[P1] The automatic within-group contrast test is too weak and changes identifiable models.** In `src/api/levels.rs:194-214`, `leading_partitions_subjects` only checks that each subject has a constant leading-column key. That is true for any baseline grouping covariate, even when many subjects share the same value. For example, with `theta EFFECT[ARM, TIME]` and multiple subjects per arm, a subject-level eta does not carry the arm mean; global sum-to-zero is sufficient. The current test nevertheless chooses `sum_to_zero_within`, removes one fixed-effect degree of freedom per arm, and rejects an explicit global contrast as unidentified. The data-side condition needs to establish that the leading tuple identifies the eta unit (for the current engine, effectively a one-to-one relation with subject), not merely that it is constant within subject.

2. **[P1] `ferx check` never binds data-driven blocks.** `validate_model_file` reads and validates with the first, unbound parse at `src/api/validation.rs:3754-3795`; unlike the fit paths, it never calls `bind_theta_levels`. A focused CLI reproduction with a valid `theta PLACEBO[STUDY, TIME]` model and data returned `E_MISSING_COVARIATE` for the reserved synthetic column `__level_PLACEBO`. It also cannot run binding-time checks such as contrast resolution and the single-level rejection, so check and fit disagree. Bind the mutable parsed model and population before the data-dependent checks.

3. **[P1] Derived contrast levels do not honor the declared init/bounds/FIX contract.** `build_level_rules` encodes the dependent level as `NegSum` at `src/parser/model_parser.rs:12808-12830`, while only the free coordinates receive the declaration's init, bounds, and fixed flag. With 800 levels bounded to `[-10, 10]`, the dependent value can reach `[-7990, 7990]`; with a nonzero fixed init, it becomes `-(N-1) * init`, not the broadcast fixed value. This contradicts the documented declaration semantics and can evaluate the model outside user bounds. The contrast parameterization needs to enforce level-space bounds, or the DSL must reject or define incompatible init, bounds, and FIX combinations explicitly.

4. **[P2] `free_packed_dim` counts a different space from the optimizer.** At `src/types.rs:3985-3991`, BSV and IOV axes are combined into one dense triangular Omega, regardless of diagonal structure, FIX flags, structural-zero masks, or separate BSV/IOV blocks; mixture override coordinates are omitted. As a simple extreme, 14 fixed diagonal etas plus two free thetas and one sigma has three free coordinates, but this returns 108. The result silently switches `optimizer = auto` and, above 100, changes the default covariance estimator. This should count `pack_params` coordinates after `packed_fixed_mask` and `omega_structural_zero_mask`, which are the masks the actual optimizer and covariance code use.

5. **[P1] Adding `CompiledModel::theta_blocks` is a public Rust source break, not additive.** `CompiledModel` is public and not `#[non_exhaustive]`; adding the required field at `src/types.rs:3020-3023` breaks every external struct literal. The repository already records this exact constraint at `src/types.rs:3348-3351`. The PR description labels the API change additive and leaves the version and cross-repo migration unchecked, but the in-repo fixture churn demonstrates the break. Either avoid the field or treat this as a breaking release and land the dependent migration.

6. **[P2] Compact output reports free coordinates as the level count and omits derived levels.** `compact_theta_blocks` and YAML emission infer blocks solely from `FitResult.theta_names` at `src/io/output.rs:88-111` and `src/io/output.rs:1878-1955`. Dependent sum-to-zero or reference levels have no theta name, so an 800-level global block is reported as 799 levels, and a 100-study by 8-time block with within-study coding is reported as 700, not 800. Those dependent values are also absent from `theta_blocks.levels`, despite the PR stating every per-level value remains in YAML. The bound labels and rules need to survive into output so derived values can be reconstructed and counted.

Verification performed:

- `cargo test --lib theta_ -- --nocapture`: 95 passed.
- `cargo test --test theta_level_blocks -- --nocapture`: 10 passed.
- Focused `ferx check --json` reproduction confirmed the false `E_MISSING_COVARIATE: __level_PLACEBO`.
- PR is currently reported as conflicting with `main`; Codecov reports 82.92% patch coverage, below the repository's 90% gate.

roninsightrx:
Addressed all six review findings in 0e414c16:

1. Auto within-group contrast now requires the leading tuple to identify subjects one-to-one; shared groups use global sum-to-zero.
2. `validate_model_file` now binds level blocks before data-dependent checks and reports `E_THETA_LEVEL_BINDING` directly.
3. Constrained blocks reject nonzero broadcast initializers and bounds excluding zero; docs now define bounds as applying to free contrast coefficients, while `contrast = none` retains broadcast init/FIX semantics.
4. Scale guards now count the exact packed free mask, including diagonal/mixed BSV, separate IOV, FIX/structural-zero entries, and mixture overrides.
5. Level metadata moved behind the existing opaque `indiv_param_partials` field and is exposed through `CompiledModel::theta_blocks()`, removing the public struct-literal break.
6. Complete resolved labels, including dependent levels, are retained by `theta_level_map`; compact console/YAML output now accurately calls its stored entries free coefficients rather than misreporting them as all levels.

Verification:
- `cargo test --lib`: 3457 passed, 20 ignored
- `cargo test --test theta_level_blocks`: 12 passed
- `cargo check --tests`
- `cargo clippy --lib` completes with the branch baseline warnings

roninsightrx:
Audited the six findings and the fixes in `0e414c16` independently — read each hunk rather than taking the reply at face value, and re-ran the suite myself.

## The findings hold up

All six are legitimate. Three are things this PR got **wrong**, not merely incomplete, and they are worth naming as such:

- **#1** I explicitly considered the multi-subject-per-group case while writing the rule and dismissed it ("partition is enough"). That was wrong. Several subjects per arm means their ηs cannot carry the arm mean — η is mean-zero under Ω by construction, so an arm-level fixed offset is genuinely identified. Forcing within-arm sum-to-zero deletes a real degree of freedom and then rejects the *correct* global contrast as unidentified. The one-to-one test is right.
- **#5** I labelled a new public `CompiledModel` field "additive". It isn't, `src/types.rs` already records the constraint, and the 28-fixture churn was the break announcing itself — I read it as mechanical noise.
- **#3** The docs claimed the `(init, lower, upper)` triple "broadcasts to every level". False for the derived level under any constraint.

## The fixes verify

| Finding | Verified |
|---|---|
| 1 | `leading_identifies_subjects` adds the no-two-subjects-share-a-key check; a subject with no observations returns `false` (conservative → global). Test: `a_leading_group_shared_by_subjects_uses_global_sum_to_zero`. |
| 2 | `validate_model_file` binds before the data-dependent checks and surfaces `E_THETA_LEVEL_BINDING`. `check_binds_level_blocks_before_data_validation` pins the exact `__level_PLACEBO` reproduction. Binding is a no-op returning `Ok` for models with no blocks, so the `else` arm still runs for everything else. |
| 3 | `validate_constrained_level_init` rejects nonzero init and bounds excluding 0. Note this also disposes of the `FIX` half of the finding: with `init = 0` forced, a fixed block's derived level is `-0 = 0`, consistent. |
| 4 | Now counts `packed_fixed_mask` / `omega_structural_zero_mask` — the masks estimation itself uses. Correct fix. |
| 5 | Metadata behind `IndivParamPartials` (already opaque, `pub(crate)` fields, constructed externally via `::empty()`), exposed as `CompiledModel::theta_blocks()`. Removes the break rather than documenting it. |
| 6 | Output relabelled; see below. |

Independent full run on `0e414c16`: **4082 passed, 0 failed**.

## Two things I would not close silently

**1. Finding 6 is fixed by correction, not by restoration.** The review asked for the bound labels and rules to "survive into output so derived values can be reconstructed and counted". What landed instead relabels the output to `n_free_coefficients` / `coefficients` and keeps the complete labels on `theta_level_map()`. That does fully fix the *defect* — the false claim and the misleading label — and the derived value is reconstructible (minus the sum of the block's coefficients under sum-to-zero, exactly `0` under `ref`).

What it does not do is put `n_levels` or the derived value in the fit YAML. Doing that needs the level metadata on `FitResult`, which has 31 struct-literal sites and no `Default` — i.e. a new public field and the same class of break finding #5 just objected to, this time on a struct the PR template does contemplate changing (with a ferx-r PR). **That is a call for you, not for me to make unilaterally in the same PR where the lesson was the opposite.** Say the word and I will add the field, tick the `FitResult` box, and open the ferx-r side.

**2. `free_packed_dim_matches_separate_packed_masks` re-derives the implementation** and compares it to itself; the load-bearing assertion is the `== 7` pin. I added `free_packed_dim_ignores_fixed_omega_coordinates` in `04fdf066` — the review's own case, 14 fixed diagonal ηs plus two free θ and one free σ, asserting **3** (the old formula gives **108**) and that neither scale guard fires. That is the test that would have caught the finding.

## Also fixed

- **The `CONFLICTING` merge state**, which the review flagged and the fix commit did not address. Merged `origin/main` in `c71164fa` — only `CHANGELOG.md` collided (this entry against #1092's `clamp`); source auto-merged. Full suite after the merge: **4097 passed, 0 failed**.
- **Three stale claims in the PR description** that the fix commit invalidated and left standing: "every per-level estimate still lands in the fit YAML", `CompiledModel.theta_blocks` listed as a public field, and the 28-fixture churn in the reviewer hints. All corrected, and the description now carries a Review round 1 table.

## Still open, deliberately

`contrast = ref` has two limits, now documented rather than implied: the reference is always the first level in sort order (no `ref = <LEVEL>`), and it is a single reference for the whole block rather than one per group — only `sum_to_zero_within` splits into groups. A regression user reading per-study baselines probably wants the per-group form. Happy to add either; neither is a defect in what is here.

PR #1101, opened 2026-08-27
feat(api): retain block_sigma correlations in FitResult [closes #1100]

### LINKED ISSUE #1100: FitResult omits block_sigma residual correlations, so a correlated-residual fit cannot be round-tripped
## Summary

`FitResult` carries `sigma` / `sigma_names` (the estimated diagonal SDs) but not the `block_sigma` correlations. Those live on `CompiledModel.residual_correlations` and never reach the result, so a consumer holding a `FitResult` cannot reconstruct the residual covariance the fit was evaluated against, and cannot write the fitted error model back into a model file.

Not a correctness bug — the fits themselves are right. It is a completeness gap in the result type.

## What is actually going on

Worth stating precisely, because it is easy to get wrong (I did, initially). `block_sigma` off-diagonals are **not** estimated parameters that are being dropped. `build_residual_correlations` (`src/parser/model_parser.rs:13335`) converts the declared covariance into a correlation at parse time:

```rust
let rho = value / (vi.sqrt() * vj.sqrt());
...
out.push(ResidualCorrelation { sigma_i: i, sigma_j: j, rho });
```

and `residual_correlations` is thereafter **read-only** — every non-test reference in `src/` consumes it; nothing mutates it during estimation. This is deliberate and documented in `docs/model-file/error-model.qmd`:

> The diagonal sigma SDs are estimated unless marked `FIX`; the off-diagonal covariance is always converted to a fixed correlation.

So `FitResult.sigma` containing only the diagonal is correct. The gap is that the *fixed* ρ is also absent, and ρ is required to reconstruct the residual covariance:

```
Cov(εᵢ, εⱼ) = ρᵢⱼ · σᵢ · σⱼ      with σ fitted, ρ declared
```

A consumer has the σ's and not the ρ's, so it has neither the matrix nor a way to rebuild the `block_sigma` declaration.

## Why it is worth closing

- **Round-tripping a fit is impossible for these models.** Writing estimates back into a model file as new initial values — for a refit, a warm start, a post-hoc likelihood evaluation, or a bootstrap — works for every other parameter class and silently cannot work here. Found while benchmarking #1017: `fluconazole_radboudumc` was the one model in a nine-model suite that could not be scored post-hoc, purely for this reason. (It is recoverable by re-reading ρ out of the `.ferx` source, which is what I ended up doing — but that means the result type is not self-describing, and any tool operating on `FitResult` alone is stuck.)
- **`ferx-r` cannot present the residual covariance.** It can show the σ diagonal and nothing about the correlation structure the user declared, even though that structure is part of what the fit assumed.
- **It is the only parameter block with this property.** `omega` ships as a full `DMatrix` including off-diagonals; `omega_iov` likewise. `sigma` is a bare `Vec<f64>` with the correlation structure stripped.

## Suggested shape

Additive, no breaking change: carry the correlations through to the result, e.g.

```rust
/// Residual-error correlations in force for this fit, copied from the model.
/// Fixed by declaration (see `block_sigma`), not estimated — present so the
/// residual covariance `rho_ij * sigma_i * sigma_j` is reconstructible from a
/// `FitResult` alone.
#[serde(default, skip_serializing_if = "Vec::is_empty")]
pub residual_correlations: Vec<ResidualCorrelation>,
```

`ResidualCorrelation` is already a plain `{ sigma_i, sigma_j, rho }` struct in `types.rs:1460`. `skip_serializing_if` keeps existing fit YAML/JSON byte-identical for the diagonal-σ models that are the overwhelming majority. A matching `block_sigma:` entry in the fit YAML would make the emitted file re-runnable.

If the preference is to expose the assembled covariance rather than the raw ρ's, that works equally well for the round-trip use case — the ρ list is just the smaller and more directly re-declarable form.

## Minor, same area

- `docs/model-file/error-model.qmd` says the correlation is "always converted to a fixed correlation", but the `simulate()` paragraph ~20 lines below refers to "a **fitted** or fixed `block_sigma` off-diagonal". The word *fitted* contradicts the paragraph above and is what led me to assume these were estimated parameters. Worth a one-word fix.
- `build_residual_correlations` ends each loop iteration with `let _fixed = block.fixed;` (`model_parser.rs:13396`). It is a genuine no-op — the `FIX` flag is applied to the diagonal `SigmaSpec`s at `model_parser.rs:12823`, and ρ is unconditionally fixed regardless — so the binding looks like a leftover and reads as though a case is being dropped. Deleting it (or replacing it with a comment saying why there is nothing to do) would save the next reader the trace I just did.

## Environment

Found on branch `pr1017-vi` at `9e375c0e` while benchmarking #1017 across `ferx-testdata`; the gap is independent of that PR and applies to `foce` / `focei` / `saem` / `imp` equally.

roninsightrx:
## Why

`FitResult` retained the fitted diagonal sigma SDs but dropped the fixed correlations declared by
`block_sigma`. A consumer holding only the result therefore could not reconstruct the residual
covariance, write fitted error parameters back into a model, or present the correlation structure.

Closes #1100.

## What changed

- Add `FitResult.residual_correlations` with backward-compatible serde defaults and omission for
  diagonal residual models.
- Copy the model correlations into every live fit result and preserve them through JSON and `.fitrx`
  save/load.
- Emit named `block_sigma` pairs in fit YAML with both the declared correlation and the covariance
  reconstructed from the fitted sigma SDs.
- Validate correlation indices and values when loading `.fitrx` bundles.
- Remove the redundant `BlockSigmaSpec.fixed` state/no-op; the parser already applies `FIX` to the
  diagonal `SigmaSpec` entries.
- Correct the documentation's misleading reference to a "fitted" off-diagonal and document the new
  result/output field.

## Alternatives considered

Storing the fully assembled covariance matrix was considered. Retaining the raw fixed correlations
is smaller, matches `CompiledModel`, and is directly usable when rewriting a `block_sigma`
declaration. The human YAML also reports the assembled covariance for convenience.

## Cross-repo dependency
| ferx (R) | — | **required follow-up** | after this merges |

**Correction (review):** the earlier "not needed" row was wrong. `residual_correlations` is a
*required* public field, and `ferx-r/src/rust/src/lib.rs` builds `FitResult` with three
**exhaustive** struct literals (lines 1660, 3278, 3635 — no `..Default::default()`). Because
`ferx-r/src/rust/.cargo/config.toml` patches in the local `../ferx-core`, a local
`cd ../ferx-r && R CMD INSTALL .` fails with `E0063: missing field residual_correlations` as soon as
this branch is checked out beside it. Serde defaulting does not help struct-literal construction.

The follow-up is ordered, not optional (per `CLAUDE.md`):

1. Merge this PR.
2. ferx-r PR: add `residual_correlations: Vec::new()` to the three literals **and** bump the
   ferx-core pin via `ferx-r/tools/update-ferx-core-lock.sh` (never a bare `cargo update`).
   Both must be in the same PR — the field does not exist on ferx-core `main`, so the literal
   change cannot compile in ferx-r CI before the lock bump.
3. Any downstream repo (e.g. `ferxtranslate`) bumps its ferx-r pin.

Local ferx-r builds are broken between step 1 and step 2 for anyone with both checkouts.

## Breaking changes
- [ ] None

<details>

Serialized JSON and `.fitrx` data are backward compatible: missing correlations default to an empty
vector, and the field is omitted for diagonal-sigma models. External Rust code that constructs a
`FitResult` with a struct literal must add `residual_correlations: Vec::new()` (or the applicable
correlations).

</details>

## Numerical validation
- [x] Compared against: NONMEM 7.5.1 and the existing correlated-residual anchor

```
# existing reference: data/correlated_residual_combined.csv
block_sigma = [0.04, 0.10, 1.00] FIX
rho = 0.10 / (sqrt(0.04) * sqrt(1.00)) = 0.5
NONMEM METHOD=1 MAXEVAL=0 OFV: 18.722087

# after
ferx initial-evaluation OFV: unchanged by this result-only wiring
FitResult rho: 0.5
fit YAML covariance: 0.5 * 0.2 * 1.0 = 0.100000
```

## Performance

## Tests

<details>

```toml
[parameters]
  block_sigma (PROP_ERR, ADD_ERR) = [0.04, 0.10, 1.00] FIX

[error_model]
```

```yaml
block_sigma:
  ADD_ERR__PROP_ERR:
    covariance: 0.100000        # sigma-scale rho*sigma_i*sigma_j, NOT an entry of R
    covariance_fixed: true      # both SDs were declared FIX; false for a bare block_sigma
    correlation: 0.500000
    correlation_fixed: true     # always true: the off-diagonal is a fixed correlation
```

</details>

## Docs
- [x] New pages linked in `docs/_quarto.yml` — N/A; no new page was added
- [ ] No user-visible change

## Checklist
- [ ] `cargo clippy` clean — repository-wide clippy currently stops on four pre-existing
  `approx_constant` errors in `src/sens/two_cpt_explicit.rs`; no changed file produced an error
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A; no sensitivity or PK
  formula changed
- [ ] `Cargo.toml` version bumped if breaking — no serialized-format version bump is required for
  this additive, defaulted field

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key — N/A; neither changed
- [x] New example `.ferx` file added — N/A; no example file added
- [x] New data file added — N/A; no data file added

### ferx-book
- [x] New estimator, DSL feature, or fit option — N/A; none added

- [x] All affected documentation pages render cleanly: `quarto render docs`
- [x] No external site/book example execution needed; no DSL, estimator, data, or example changed

## Reviewer hints

The important wiring is the copy in `api/fit.rs`, the serde behavior on `FitResult`, and the
`SigmaWire` `.fitrx` round-trip. The YAML covariance deliberately uses the fitted sigma SDs while
retaining the declared fixed rho. Diagonal-sigma JSON and `.fitrx` output remain unchanged because
the empty vector is skipped.

Verification performed:

- `cargo check --tests`
- `cargo test --lib` — 3,530 passed, 22 ignored, 0 failed
- `cargo test --test fit_result_residual_correlations`
- focused JSON, `.fitrx`, parser, and YAML tests
- `quarto render docs`

## Open questions

None for the core change. ferx-r can expose `residual_correlations` in a follow-up after this merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,10 @@ section of the SDLC for the versioning policy).
 ### Added
 - **Correlated-residual fits now retain their fixed `block_sigma` correlations (#1100).**
   `FitResult.residual_correlations`, JSON, `.fitrx`, and fit YAML now carry the declared
-  correlations, so consumers can reconstruct the fitted residual covariance without re-reading
-  the model source; diagonal-sigma output remains unchanged.
+  correlations, so consumers can reconstruct the sigma-scale residual covariance without re-reading
+  the model source; diagonal-sigma output remains unchanged. The fit YAML's `block_sigma` section
+  reports `correlation_fixed` (always true) separately from `covariance_fixed`, which is true only
+  when both sigma SDs were declared `FIX`.
 - **Theta level blocks — hundreds of fixed effects from one declaration (#1064).**
   `theta PLACEBO[800](0.0, -10.0, 10.0)` declares 800 thetas sharing one init/bounds triple, read back
   by a *gather* — `PL = PLACEBO[PLA_IDX]`, where `PLA_IDX` is a 1-based data column. The data-driven
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
+- **Correlated-residual fits now retain their fixed `block_sigma` correlations (#1100).**
+  `FitResult.residual_correlations`, JSON, `.fitrx`, and fit YAML now carry the declared
+  correlations, so consumers can reconstruct the fitted residual covariance without re-reading
+  the model source; diagonal-sigma output remains unchanged.
 - **Theta level blocks — hundreds of fixed effects from one declaration (#1064).**
   `theta PLACEBO[800](0.0, -10.0, 10.0)` declares 800 thetas sharing one init/bounds triple, read back
   by a *gather* — `PL = PLACEBO[PLA_IDX]`, where `PLA_IDX` is a 1-based data column. The data-driven
--- a/docs/api/types.qmd
+++ b/docs/api/types.qmd
@@ -147,9 +147,18 @@ pub struct FitResult {
 
 For a model declared with `block_sigma`, `residual_correlations` contains the
 fixed off-diagonal correlations used by the fit. Together with the fitted
-sigma SDs, each residual covariance is reconstructible as
-`rho * sigma[sigma_i] * sigma[sigma_j]`. The field is empty for diagonal
+sigma SDs they reconstruct the **sigma-scale** covariance
+`Cov(EPS_i, EPS_j) = rho * sigma[sigma_i] * sigma[sigma_j]` — the same quantity
+NONMEM reports in a `$SIGMA BLOCK`. The field is empty for diagonal
 residual-error models.
+
+Note that this is *not* the entry of a subject's residual covariance matrix
+`R`. Each observation applies its own sigma *loadings* first, so the `R` term
+pairing observations `j` and `k` is `c_j * c_k * rho * sigma_i * sigma_j`,
+where `c` is `1` for an additive component and the individual prediction `f`
+for a proportional one. A consumer rebuilding `R` for, say, a paired
+total/unbound model must include the `f_total * f_unbound` factor; see
+[Error models](../model-file/error-model.qmd) for the full formula.
 
 ## SubjectResult
 
--- a/docs/api/types.qmd
+++ b/docs/api/types.qmd
@@ -129,6 +129,8 @@ pub struct FitResult {
     pub theta_names: Vec<String>,
     pub omega: DMatrix<f64>,
     pub sigma: Vec<f64>,
+    pub sigma_names: Vec<String>,
+    pub residual_correlations: Vec<ResidualCorrelation>,
     pub covariance_matrix: Option<DMatrix<f64>>,
     pub se_theta: Option<Vec<f64>>,
     pub se_omega: Option<Vec<f64>>,
@@ -142,6 +144,12 @@ pub struct FitResult {
     pub warnings: Vec<String>,
 }
 ```
+
+For a model declared with `block_sigma`, `residual_correlations` contains the
+fixed off-diagonal correlations used by the fit. Together with the fitted
+sigma SDs, each residual covariance is reconstructible as
+`rho * sigma[sigma_i] * sigma[sigma_j]`. The field is empty for diagonal
+residual-error models.
 
 ## SubjectResult
 
--- a/docs/model-file/error-model.qmd
+++ b/docs/model-file/error-model.qmd
@@ -701,7 +701,27 @@ identical exclusion in the likelihood) and stay independent.
 The fixed correlations are copied to `FitResult.residual_correlations`, so the
 residual covariance assumed by a fit remains available without re-reading the
 model source. The human-readable fit YAML also emits a `block_sigma` section
-with each named pair's fitted covariance and declared correlation.
+with each named pair's covariance and declared correlation:
+
+```yaml
+block_sigma:
+  ADD_ERR__PROP_ERR:
+    covariance: 0.100000
+    covariance_fixed: false
+    correlation: 0.500000
+    correlation_fixed: true
+```
+
+`covariance` is the **sigma-scale** covariance `rho * sigma_i * sigma_j`, the
+same quantity NONMEM prints for a `$SIGMA BLOCK` — not an entry of a subject's
+residual covariance matrix `R`. Each observation applies its own sigma loadings
+first, so the `R` term is `c_j * c_k * rho * sigma_i * sigma_j` with `c = 1`
+for an additive component and `c = f` (the individual prediction) for a
+proportional one — the `f_total * f_unbound * Cov(EPS_total, EPS_unbound)` form
+above. `correlation_fixed` is always `true` (the off-diagonal is converted to a
+fixed correlation); `covariance_fixed` is `true` only when both sigma SDs in
+the pair were declared `FIX`, since otherwise the SDs — and hence the
+covariance — moved during estimation.
 
 Under `method = focei` the off-diagonal covariance enters the interaction
 correction as well as the data term: the Almquist 2015 conditional Hessian
--- a/docs/model-file/error-model.qmd
+++ b/docs/model-file/error-model.qmd
@@ -691,12 +691,17 @@ include the process-noise inflation.
 `simulate()` draws the full residual vector for each subject's paired
 observations (same subject time and occasion) from the dense covariance `R`
 built by `compute_r_matrix_with_correlations` — the same matrix FOCE/FOCEI/SAEM/
-`imp` evaluate the likelihood against — so a fitted or fixed `block_sigma`
+`imp` evaluate the likelihood against — so a fixed `block_sigma`
 off-diagonal is reproduced in simulated total/unbound (or cross-branch, for a
 covariate-selected model) rows, and a VPC or posterior-predictive check of the
 correlated endpoints recovers the specified correlation (#672). FREM covariate
 pseudo-observations are excluded from the correlated draw (mirroring the
 identical exclusion in the likelihood) and stay independent.
+
+The fixed correlations are copied to `FitResult.residual_correlations`, so the
+residual covariance assumed by a fit remains available without re-reading the
+model source. The human-readable fit YAML also emits a `block_sigma` section
+with each named pair's fitted covariance and declared correlation.
 
 Under `method = focei` the off-diagonal covariance enters the interaction
 correction as well as the data term: the Almquist 2015 conditional Hessian
--- a/src/api/fit.rs
+++ b/src/api/fit.rs
@@ -2132,6 +2132,7 @@ fn fit_inner(
         omega: result.params.omega.matrix.clone(),
         sigma: result.params.sigma.values.clone(),
         sigma_names: result.params.sigma.names.clone(),
+        residual_correlations: model.residual_correlations.clone(),
         error_model: model.error_model,
         covariance_matrix: result.covariance_matrix,
         // The optimizer's exact packed vector (FOCE/FOCEI paths), so a later
--- a/src/api/tests/simulate_with_uncertainty_tests.rs
+++ b/src/api/tests/simulate_with_uncertainty_tests.rs
@@ -401,6 +401,7 @@ fn synthetic_fit(template: &ModelParameters) -> FitResult {
         omega: template.omega.matrix.clone(),
         sigma: template.sigma.values.clone(),
         sigma_names: template.sigma.names.clone(),
+        residual_correlations: Vec::new(),
         error_model: ErrorModel::Proportional,
         covariance_matrix: Some(cov),
         se_theta: None,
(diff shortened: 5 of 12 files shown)


roninsightrx:
## Code review — `feat(api): retain block_sigma correlations in FitResult`

Reviewed at medium effort.

**Verified OK:** `.fitrx` validation runs before `wire_to_fit_result` (`fitrx.rs:1638`); wire structs are not `deny_unknown_fields` (forward-compatible); `JSON_SCHEMA_VERSION` correctly not bumped (purely additive); sigma is on the SD scale so `rho·σᵢ·σⱼ` is dimensionally right; `BlockSigmaSpec.fixed` removal is safe (the local `fixed` binding still drives `SigmaSpec.fixed`); the new integration test's API signatures and data file all exist.

### Findings

**1. `src/types.rs:4935` — HIGH. Adding a required public field to `FitResult` breaks the sibling R wrapper, contrary to the PR's "ferx (R) — not needed" row.**

`../ferx-r/src/rust/src/lib.rs` builds `FitResult` with three *exhaustive* struct literals (lines 1660, 3278, 3635 — no `..Default::default()`). Because `ferx-r/src/rust/.cargo/config.toml` patches in the local `../ferx-core`, every local `R CMD INSTALL .` of ferx-r fails with `E0063: missing field residual_correlations` the moment this lands, and ferx-r CI fails as soon as its `Cargo.lock` is bumped. The migration note in the PR body already says this ("external Rust code … must add `residual_correlations: Vec::new()`") — the cross-repo table contradicts it. A matching ferx-r PR is required per CLAUDE.md.

**2. `src/io/output.rs:2150` — MEDIUM. `fixed: true` is hard-coded next to a covariance that is not fixed when the block is declared without `FIX`.**

`block_sigma (A,B) = [...]` without `FIX` is legal (`docs/model-file/error-model.qmd:671`: "The diagonal sigma SDs are estimated unless marked `FIX`"), so the emitted `covariance: rho·σ̂ᵢ·σ̂ⱼ` is a fitted quantity that moved during estimation, yet the YAML labels the whole entry `fixed: true`. Compounding it, this PR deletes `BlockSigmaSpec.fixed` (`model_parser.rs`), removing the only record of whether the block carried `FIX`, so the writer cannot report it correctly. Use `result.sigma_fixed[i] && result.sigma_fixed[j]` for the covariance, or move the flag under `correlation`, where it is true.

**3. `src/io/output.rs:2147` (and `docs/api/types.qmd:148`) — MEDIUM. The reported "covariance" is the sigma-scale covariance, not the residual covariance entering `R`.**

For any proportional component the runtime cross term is loading-scaled: `cross_observation_covariance` (`stats/residual_error.rs:766-777`) multiplies `rho·σᵢ·σⱼ` by each row's loading coefficient (`f` for proportional) — exactly the `f_total * f_unbound * Cov(...)` form documented at `error-model.qmd:679`. A consumer following the new doc sentence ("each residual covariance is reconstructible as `rho * sigma[i] * sigma[j]`") to rebuild `R` for, say, a total/unbound `block_sigma` will be off by `f_i·f_j`. Either rename the YAML key to `sigma_covariance` / note the ipred scaling, or state the loading-scaled formula in the docs.

**4. `src/io/output.rs:3922` — LOW. The new YAML unit test pins an index ordering the parser can never emit, so it cannot catch an `i`/`j` swap in the writer.**

`build_residual_correlations` (`model_parser.rs:13285-13292`) always pushes `sigma_i` = the block's *row* name and `sigma_j` = the *col* name, i.e. `sigma_i > sigma_j` in global sigma order. The test sets `sigma_i: 0, sigma_j: 1` and asserts the key `PROP_ERR__ADD_ERR`, whereas a real fit of the documented model emits `ADD_ERR__PROP_ERR` (as the PR body's own example shows). The only test covering the real ordering lives in `tests/fit_result_residual_correlations.rs`, which per CLAUDE.md is nightly-only — so no per-PR job ever sees the actually-emitted key.

**5. `src/io/fitrx.rs:1628` — LOW. The new `.fitrx` correlation guard accepts `|rho| == 1.0`.**

A bundle carrying `rho = ±1` for a genuinely paired sigma loads cleanly and yields a singular subject `R` (the 2×2 block is exactly rank-deficient), producing a NaN/Inf OFV on a later evaluation instead of a clear load-time error. The parser has the same inclusive bound (`(-1.0..=1.0).contains(&rho)`), so this only mirrors an existing gap — but since the check is being introduced here, `abs() >= 1.0` would be the safer bound. The loop also does not reject duplicate `(i, j)` pairs, which would double-count the cross term in `cross_observation_covariance` and emit duplicate YAML keys.

---

roninsightrx:
## Review findings addressed — 61f5b18

All five findings from the review above are resolved.

### 1. HIGH — `FitResult` field breaks ferx-r (PR body said "not needed")

The finding is correct and the cross-repo table was wrong. `residual_correlations` is a *required* public field, and `ferx-r/src/rust/src/lib.rs` builds `FitResult` with three **exhaustive** struct literals (1660, 3278, 3635 — no `..Default::default()`), so the local `[patch]` in `ferx-r/src/rust/.cargo/config.toml` makes `R CMD INSTALL .` fail with `E0063: missing field residual_correlations`. Serde defaulting does not help struct-literal construction.

No code fix here — this one is genuinely cross-repo. **PR body corrected** to mark the ferx-r change as a required, ordered follow-up:

1. merge this PR;
2. one ferx-r PR that adds `residual_correlations: Vec::new()` to the three literals **and** bumps the pin via `ferx-r/tools/update-ferx-core-lock.sh` — both in the same PR, since the field does not exist on ferx-core `main` and the literal change cannot compile in ferx-r CI before the lock bump;
3. downstream repos bump their ferx-r pin.

I deliberately did *not* open the ferx-r PR yet: it cannot compile until this merges.

### 2. MEDIUM — `fixed: true` hard-coded next to a non-fixed covariance

`src/io/output.rs`. A bare `block_sigma (...) = [...]` estimates its diagonal SDs, so the covariance those SDs scale moves during the fit. The entry now separates the two:

```yaml
block_sigma:
  ADD_ERR__PROP_ERR:
    covariance: 0.100000
    covariance_fixed: false     # derived from sigma_fixed[i] && sigma_fixed[j]
    correlation: 0.500000
    correlation_fixed: true     # always true — the off-diagonal is a fixed correlation
```

`BlockSigmaSpec.fixed` stays deleted: `FIX` on a block already marks every sigma in it fixed (`test_parse_block_sigma_fix_marks_sigmas_fixed`), so `result.sigma_fixed` carries the information the writer needs. Two tests cover it — the `FIX` and non-`FIX` cases.

### 3. MEDIUM — reported "covariance" is sigma-scale, not the residual covariance in `R`

Kept the `covariance` key: under a `block_sigma:` heading it is the sigma-block covariance, matching what NONMEM prints for `$SIGMA BLOCK`. Fixed the docs instead, which were the part that would mislead. `docs/api/types.qmd` and `docs/model-file/error-model.qmd` now say the value is the **sigma-scale** `rho * sigma_i * sigma_j` and *not* an entry of `R`, and give the loading-scaled runtime form `c_j * c_k * rho * sigma_i * sigma_j` (`c = 1` additive, `c = f` proportional) with the explicit warning that rebuilding `R` for a paired total/unbound model needs the `f_total * f_unbound` factor.

### 4. LOW — YAML test pinned an ordering the parser cannot emit

`build_residual_correlations` always sets `sigma_i` = the block's *row* name, so `block_sigma (PROP_ERR, ADD_ERR)` yields `(1, 0)` / key `ADD_ERR__PROP_ERR`. The unit test pinned `(0, 1)` / `PROP_ERR__ADD_ERR` and so could not catch an i/j swap. It now uses the real ordering, with a comment saying why, plus the second test from finding 2. Both are `--lib`, so per-PR coverage no longer depends on the nightly-only integration test.

### 5. LOW — `.fitrx` guard accepted `|rho| == 1`, and duplicate pairs

Both tightened, in the loader **and** the parser (the parser had the same inclusive bound):

- `|rho| >= 1.0` is now rejected. One wrinkle worth flagging: the parser's original `rho = value / (sqrt(vi) * sqrt(vj))` rounds an exactly-singular block (`[0.04, 0.04, 0.04]`) down to `rho = 0.9999999999999998`, which slips past a naive `|rho| < 1` test — my first attempt failed on exactly that. The check now compares on the covariance scale with a single sqrt (`value.abs() < (vi * vj).sqrt()`), which is both exact for that case and slightly more accurate for the retained `rho`.
- Duplicate `(i, j)` pairs (unordered) are rejected on `.fitrx` load, since they would double-count the cross term in `cross_observation_covariance` and emit duplicate YAML keys.

Tests: parse-time unit-rho rejection for both signs, and `.fitrx` rejection of `rho = ±1`, `NaN`, and a duplicate pair, plus a positive control asserting a well-formed pair still loads.

### Verification

- `cargo fmt --check` clean
- `cargo test --lib` — 3532 passed, 22 ignored, 0 failed
- `cargo test --test fit_result_residual_correlations` — 1 passed
- `cargo check --tests` clean; the two remaining warnings (`recorded` never used, an `unused_mut` in `model_parser_tests.rs`) are present on `main` unchanged
- `cargo clippy --lib --tests` produced no new errors on the changed files

`CHANGELOG.md` updated for the new `covariance_fixed` / `correlation_fixed` keys.

---

Issue #1098, opened 2026-08-27
vi: the bad-basin detector warns but does not demote `converged`, unlike the adjacent noise-floor detector


roninsightrx:
## Summary

`FitResult.converged` comes back `true` on a VI fit that landed 1594 OFV units from the optimum with σ welded to its runaway bound. The condition **is already detected** — `elbo_tightness()` flags it and emits a good warning — but that branch is the only one of the two quality checks in `run_vi` that leaves `converged` untouched.

This is a reporting bug, not an estimation one. The estimation half is #1097.

## The asymmetry

Both checks live in `src/estimation/vi/run.rs`, ~40 lines apart, and both answer "did this fit actually get anywhere". They disagree on what to do about it:

`trace_still_drifting` → **demotes** (`src/estimation/vi/run.rs:867`):

```rust
if converged && !param_settled && trace_still_drifting(&trace, settle_window(n_iters_run)) {
    converged = false;
    noise_floor_stop = true;
    warnings.push(...);
}
```

`tightness.is_implausible()` → **warns only** (`src/estimation/vi/run.rs:911`):

```rust
if tightness.is_implausible() {
    warnings.push(format!(
        "VI: the ELBO is not a usable bound at this estimate — ... The optimizer has almost \
         certainly settled in a bad basin: `converged` reflects a flat objective, not a \
         good one. ...",
    ));
}
```

The comment above it states the intent exactly right:

> Is the bound we are about to report actually tight? A stuck optimizer produces a flat trace and stable parameters — indistinguishable, to the convergence test, from a converged one. **This is the check that tells them apart, and it is the difference between reporting a bad fit as successful and saying so.**

It tells them apart and then reports the bad fit as successful anyway. The warning string has to spend a clause explaining that `converged` does not mean what it says, which is the tell.

## Reproduction

The warfarin case from #1097 — testdata `warfarin_pk` model, `method = vi`, unmodified starting values:

```
converged:              true
elbo_tightness_ratio:   10004.67
sigma (PROP_ERR):       145.837        (FOCEI: 0.010565)
AGQ −2 log L:           +1307.919      (FOCEI: −285.977)
n_iterations:           1500           (of a 25000 ceiling)
```

The warning fires correctly. `converged` still says `true`.

## Why this matters beyond cosmetics

[...]

PR #1102, opened 2026-08-27
fix(api): warn on hidden parameter guards [closes #1099]

### LINKED ISSUE #1099: Boundary warning covers theta only; sigma and Omega can finish pinned to their internal runaway guards with nothing reported
The boundary warning covers θ only; σ and Ω can finish welded to their internal runaway guards with nothing reported

## Summary

`boundary_estimate_warning` tells a user when a **θ** estimate finished on a bound. There is no equivalent for **σ** or **Ω**. That is backwards relative to how visible the bounds are: θ's bounds are the ones the user typed into the model file and can see, while σ's and Ω's are internal guards the user never declared and cannot inspect.

A fit can therefore return σ pinned to `exp(5) = 148.413159` and report nothing about it. Split out of #1098 as an estimator-agnostic concern; VI (#1097) is just the first method observed to trip it.

## The gap

`boundary_estimates` (`src/api/postfit.rs:1071`) iterates `params.theta` and nothing else:

```rust
fn boundary_estimates(params: &ModelParameters) -> Vec<(String, f64, f64, &'static str)> {
    let mut hits = Vec::new();
    for i in 0..params.theta.len() {
        if params.theta_fixed.get(i).copied().unwrap_or(false) { continue; }
        let est = params.theta[i];
        if let Some((side, bound)) =
            theta_boundary_side(est, params.theta_lower[i], params.theta_upper[i])
        { ... }
    }
    hits
}
```

Note it already handles the subtlety that matters here — it skips FIXed entries. That exclusion is not optional, because `compute_bounds` pins FIXed parameters *by construction* (`src/estimation/parameterization.rs:345`):

> Parameters marked FIX are given `lower == upper == packed_value` […]

so a naive "is it on a bound" sweep would flag every FIXed θ, ω and σ in the suite. `thioguanine_mmc` alone has three FIXed θ and one FIXed ω.

Meanwhile σ and Ω get bounds nobody asked for (`src/estimation/parameterization.rs:388`):

```rust
// Sigma bounds (log-transformed)
for _ in 0..n_sigma {
    lower.push(-8.0); // exp(-8) ≈ 3e-4
    upper.push(5.0);  // exp(5) ≈ 148
}
```

and, for the Ω Cholesky diagonal, `[-6.0, 6.0]`. These are runaway guards, and they work — they are why a diverging fit returns a finite number instead of an overflow. The complaint is only that hitting one is currently indistinguishable, in `FitResult`, from converging to an interior optimum.

## Observed

From the #1017 benchmark, `warfarin_pk` under `method = vi` at `vi_mc_samples = 128`:

```
sigma (PROP_ERR):  148.413159   == exp(5), the compute_bounds ceiling, to all printed digits
converged:         true
warnings:          nothing about sigma or any bound
```

FOCEI on the same data and start gives σ = 0.010565. The estimate is off by a factor of 14 000 and sits exactly on a guard the model file never mentions.

## Why it is worth a general fix

- **The value is not a fit result.** `exp(5)` is an implementation constant. Reporting it in the σ column, with no annotation, presents an artifact of the guard as an estimate.
- **It is the cheapest possible bad-fit detector.** Equality against a known constant — no tolerance tuning, no statistics, no per-method knowledge.
- **The bounds are invisible to the user.** A θ on a bound is at least explicable by reading the model file. `σ = 148.41` explains nothing; a user cannot even discover where 148.41 came from without reading `parameterization.rs`.
- **Consumers are already wired for it.** `warnings_structured` carries severity and category, and `ferx-r` surfaces warnings — a new `WarningCode` needs no new plumbing.

## Suggested shape

Extend the existing fit-end warning rather than adding a parallel mechanism:

1. Generalise `boundary_estimates` to also sweep the packed σ and Ω coordinates against the same `PackedBounds` that `clamp_to_bounds` uses, comparing **in packed space** (log for σ, log-Cholesky-diagonal for Ω) so the comparison is against the literal `5.0` / `6.0` rather than a round-tripped decimal.
2. **Skip FIXed entries**, exactly as the θ sweep already does, for the `lower == upper` reason above.
3. Distinguish the two cases in the message, because the remedies differ: a **user-declared** θ bound means "widen your bound or reconsider the covariate"; an **internal guard** on σ/Ω means "this fit diverged" and there is no bound for the user to widen.
4. Give the guard case its own `WarningCode` (`W_PARAM_AT_RUNAWAY_GUARD` or similar) so R-side and CI consumers can branch on it without string-matching.

Whether hitting an internal guard should additionally force `converged = false` is the open question — I lean yes, but it changes behaviour for every estimator, so it is worth deciding separately from the warning. #1098 argues that case for VI specifically, where the fit is demonstrably not converged.

## Scope

Across the nine-model benchmark, only VI ever hit an internal guard; FOCEI and SAEM stayed interior on every model, including ones where they landed on a clearly worse optimum than the winner. So this is a latent gap that VI surfaced, not a regression — and the guard is worth having regardless of which method trips it next.

Related: #1098 (the `converged` flag on the same fit), #1097 (why VI diverges there), #1017.

roninsightrx:
## Why

Free parameter estimates could finish on ferx's hidden packed-space guards while the fit reported no appropriate warning. This made an implementation safety constant such as `exp(5) = 148.413159` look like an ordinary fitted sigma. Theta bounds were diagnosed separately, but that path also mislabeled ferx's implicit `1e-10` / `1e9` theta caps as user-declared bounds.

Closes #1099.

## What changed

- Add the typed `parameter_at_runaway_guard` warning code.
- Inspect packed coordinates using the same `pack_params`, `compute_bounds`, coordinate labels, reporting values, and FIX mask used by optimization.
- Accept only a few ULPs of packed-space arithmetic noise, covering exact scaled-space clamps after unscaling without classifying values merely near a wide safety rail.
- Route implicit theta cap hits to the new warning while keeping genuine user-declared theta bounds on `boundary_estimate`.
- Cover base Omega/Sigma, Omega-IOV, and mixture overrides while excluding every FIX'd coordinate.
- Give lower hits collapse-to-zero guidance and upper hits runaway-estimate guidance, including both remedies when a fit hits both sides.
- Include reporting-scale and packed-space values in structured warning details.

This does not change the fitted values or the `converged` flag.

## Alternatives considered

- Reusing `boundary_estimate` for every parameter would give hidden safety guards the wrong remediation ("relax the bound") and would prevent consumers from distinguishing a declared constraint from a diverged coordinate.
- Comparing reporting-scale values would round-trip Omega through variance/covariance output. Comparing the optimizer coordinate directly tests the literal `-8/5`, `-6/6`, and `-10/10` guards.
- A range-relative proximity tolerance was rejected because these are broad safety rails, not inferential boundaries. The implemented tolerance is only four machine epsilons in packed space, enough for scaled clamp/unscale round-off.

## Cross-repo dependency

ferx-r already surfaces structured warning codes generically; no glue or lockfile update is required.

## Breaking changes

- [x] None

The new `WarningCode` variant is additive, and the public enum is already `#[non_exhaustive]`.

<details>

Not applicable.

</details>

## Numerical validation

```text
Not applicable: this post-fit diagnostic does not alter estimation, fitted values, OFV, or convergence.
```

- [x] Not applicable (diagnostic-only change; numerical results are unchanged)

## Performance

- [x] No significant impact expected — one linear packed-vector scan at fit end

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 3,531 passed, 22 ignored)
- [x] Regression tests added that fail without this fix: sigma and Omega guard hits, FIX exclusion, scaled-space ULP recovery, exact-vs-near behavior, lower/upper/mixed guidance, hidden theta caps versus user bounds, warning classification, and stable serde token

Additional checks:

- `cargo check --tests` passes.
- `cargo fmt --all -- --check` passes.
- `quarto render docs` passes.
- `cargo clippy --lib` passes (with the repository's existing warnings).
- `cargo clippy --lib --tests` reaches four pre-existing `clippy::approx_constant` denials in `src/sens/two_cpt_explicit.rs`; it reports no new error from this diff.

- [x] Example included inline below

<details>
<summary>Example (structured output)</summary>

When a free `PROP_ERR` sigma finishes at the internal packed upper guard `5.0`, `warnings_structured` contains:

```json
{
  "severity": "Warning",
  "category": "parameter_at_runaway_guard",
  "message": "Internal optimizer parameter guard reached by estimate(s): PROP_ERR (estimate 148.4132; packed coordinate 5.0000 at upper guard). ...",
  "source_method": null,
  "details": {
    "guard_space": "packed",
    "parameters": [{
      "parameter": "PROP_ERR",
      "estimate": 148.4131591025766,
      "packed_estimate": 5.0,
      "packed_guard": 5.0,
      "side": "upper"
    }]
  }
}
```

</details>

## Docs

- [x] `docs/` updated for user-visible changes (`warnings.qmd` and `estimation/parameterization.qmd`)
- [x] No new page was added, so `docs/_quarto.yml` is unchanged; generated `docs/_site/` remains ignored
- [ ] No user-visible change

## Checklist

- [ ] `cargo clippy` clean — blocked by four pre-existing `approx_constant` denials outside this diff, noted above
- [x] Sensitivity-path differentiability is unaffected; no `sens/` or generic PK code changed
- [x] `Cargo.toml` version bump not required; the API addition is non-breaking

### ferx-site

- [x] No new `.ferx` syntax or `[fit_options]` key; external model-DSL docs are not affected
- [x] No new data file

### ferx-book

- [x] No new estimator, DSL feature, or fit option; book update is not applicable

- [ ] All affected `examples/*.ferx` run cleanly via CLI: `cargo run --release -- examples/<model>.ferx --data data/<data>.csv`
- [x] No example execution step needed (fit-end diagnostic only; no estimator, DSL, data, or example behavior changed)

## Reviewer hints

The subtle point is the split between spaces: guard proximity is checked within four machine epsilons of the literal packed bounds, while the warning also reports the ordinary natural-scale value. Please focus on `runaway_guard_estimates` using the optimizer's existing layout/mask helpers, implicit-theta-cap routing, and side-specific guidance. Genuine user-declared theta bounds deliberately remain a separate warning category.

## Open questions

None for this PR. Whether an internal guard hit should also force `converged = false` remains separate behavior covered by #1098.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,11 +310,13 @@ section of the SDLC for the versioning policy).
   with no diagnostic from NONMEM (`nonmem_anchor/dose_attr_double_use_{A,B}.ctl`).
 
 ### Fixed
-- **Hidden Omega/Sigma runaway guards now produce a fit warning (#1099).** A free
-  Omega, Omega-IOV, Sigma, or mixture override pinned to an internal packed-space
-  safety limit emits the typed `parameter_at_runaway_guard` warning; FIX'd
-  coordinates remain excluded, and user-declared Theta bounds continue to use
-  the separate `boundary_estimate` warning.
+- **Hidden parameter guards now produce a fit warning (#1099).** A free Theta at
+  the implicit `1e-10` / `1e9` cap, or an Omega, Omega-IOV, Sigma, or mixture
+  override pinned to an internal packed-space safety limit, emits the typed
+  `parameter_at_runaway_guard` warning. FIX'd coordinates remain excluded;
+  user-declared Theta bounds continue to use the separate `boundary_estimate`
+  warning. Lower hits are identified as collapse toward zero, while upper hits
+  are identified as runaway estimates.
 - **A closed-form model with IOV and a `[scaling] y = <expr>` readout evaluated the readout's
   individual parameters at `kappa = 0` — predictions and the objective were silently wrong
   (#1079).** The readout is the analogue of NONMEM's `$ERROR` and is evaluated per record, so an
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,6 +310,11 @@ section of the SDLC for the versioning policy).
   with no diagnostic from NONMEM (`nonmem_anchor/dose_attr_double_use_{A,B}.ctl`).
 
 ### Fixed
+- **Hidden Omega/Sigma runaway guards now produce a fit warning (#1099).** A free
+  Omega, Omega-IOV, Sigma, or mixture override pinned to an internal packed-space
+  safety limit emits the typed `parameter_at_runaway_guard` warning; FIX'd
+  coordinates remain excluded, and user-declared Theta bounds continue to use
+  the separate `boundary_estimate` warning.
 - **A closed-form model with IOV and a `[scaling] y = <expr>` readout evaluated the readout's
   individual parameters at `kappa = 0` — predictions and the objective were silently wrong
   (#1079).** The readout is the analogue of NONMEM's `$ERROR` and is evaluated per record, so an
--- a/docs/estimation/parameterization.qmd
+++ b/docs/estimation/parameterization.qmd
@@ -80,6 +80,13 @@ That is how every bound-respecting optimizer (NLopt SLSQP / L-BFGS / MMA, the
 built-in BFGS, the Gauss-Newton step clamp) holds them fixed — there is no
 separate free-subspace vector.
 
+The Omega and Sigma limits above are internal runaway guards, not model-file
+bounds. If a free estimate reaches one exactly, the fit emits the structured
+`parameter_at_runaway_guard` warning. FIX'd coordinates are excluded because
+their equal bounds are intentional. The check is made in packed space, so it
+compares directly with the literal guard even though fit output reports Omega
+on the variance/covariance scale and Sigma on its stored SD scale.
+
 ## Where the packed space surfaces
 
 It is not purely internal; it shows up in several user-visible places.
--- a/docs/estimation/parameterization.qmd
+++ b/docs/estimation/parameterization.qmd
@@ -80,12 +80,15 @@ That is how every bound-respecting optimizer (NLopt SLSQP / L-BFGS / MMA, the
 built-in BFGS, the Gauss-Newton step clamp) holds them fixed — there is no
 separate free-subspace vector.
 
-The Omega and Sigma limits above are internal runaway guards, not model-file
-bounds. If a free estimate reaches one exactly, the fit emits the structured
+The Omega and Sigma limits above, plus the `1e-10` / `1e9` caps substituted for
+a log-packed Theta range that extends beyond them, are internal guards rather than
+model-file bounds. If a free estimate reaches one (within the few ULPs that
+optimizer scaling can introduce), the fit emits the structured
 `parameter_at_runaway_guard` warning. FIX'd coordinates are excluded because
 their equal bounds are intentional. The check is made in packed space, so it
 compares directly with the literal guard even though fit output reports Omega
-on the variance/covariance scale and Sigma on its stored SD scale.
+on the variance/covariance scale and Sigma on its stored SD scale. A lower hit
+means the component collapsed toward zero; an upper hit means it ran away.
 
 ## Where the packed space surfaces
 
--- a/docs/warnings.qmd
+++ b/docs/warnings.qmd
@@ -54,6 +54,7 @@ Each `WarningCode` token, what it flags, and a typical response.
 | `eps_shrinkage` | Warning | Residual (EPS) shrinkage high / negative. | Sparse data for the error model — simplify it. |
 | `eta_shrinkage` | Warning | One or more ETA shrinkages exceed ~30%. | Data poorly inform that IIV — consider removing it on the affected parameter(s). |
 | `boundary_estimate` | Warning | One or more THETA estimates are pinned to an optimizer bound. | Non-identifiability or a too-tight bound — inspect; relax the bound or simplify. |
+| `parameter_at_runaway_guard` | Warning | One or more free OMEGA / SIGMA coordinates are pinned to an internal optimizer safety guard. | The affected value is not an interior optimum — do not treat it as a reliable estimate; revisit the model, data, initial estimates, or method. |
 | `inflated_rse` | Warning | One or more THETA estimates have RSE > ~50%. | Imprecisely estimated — often over-parameterization; consider simplifying. |
 | `high_correlation` | Warning | One or more THETA (fixed-effect) pairs have \|correlation\| ≥ 0.95. | Over-parameterization / non-identifiability — fix or remove one of each pair. |
 | `data_quality` | Warning | Dataset issue (missing DV, ADDL/II, non-positive DV, …). | Fix the dataset before trusting the fit. |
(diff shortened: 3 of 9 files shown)


roninsightrx:
## Code review (medium)

Checked the layout alignment between `pack_params` / `compute_bounds` / `packed_fixed_mask` / `coordinate_names` / `coordinate_values` coordinate-by-coordinate (theta, Ω lower-triangle, sigma, Ω_IOV, mixture overrides). The walk is consistent, the FIX exclusion is doubly safe (mask filter *and* `lower >= upper`), and structural-zero off-diagonals pack to `0.0`, so they can't false-hit the ±10 rail. I also verified numerically that `ln(exp(x))` and `ln(sqrt(exp(x)²))` round-trip bit-exactly at all six guards (−8/5, −6/6, −10/10), so the unpack→repack path does not by itself defeat the equality test.

Three findings.

### 1. `src/api/postfit.rs:1113` — exact equality misses real guard hits whenever parameter scaling is active

The outer optimizer works in scaled space: `lower_s = bounds.lower[i] / scale[i]` (`outer_optimizer.rs:1471`) and the final vector is recovered as `x[i] = xs[i] * scale[i]` (lines 2073 / 2604) with **no re-clamp to `bounds`** afterwards. `(u / s) * s == u` fails for roughly 9% of scale factors, so a coordinate that NLopt pinned exactly on the scaled rail unscales to ~1 ULP off the guard, `packed_guard_side` returns `None`, and the warning silently never fires.

Concrete case: `optimizer = lbfgs` (`ParameterScaling::Auto` → `Abs` for Bfgs / Lbfgs / NloptLbfgs / Slsqp, `resolve_scaling:1192`), sigma driven to the `-8.0` floor with `scale[i] = ln(sigma_init)`, e.g. `(-8.0 / 2.302585092994046) * 2.302585092994046 != -8.0`.

The default BOBYQA path is unaffected (scale = 1.0), which is likely why the tests — which call `runaway_guard_warning` directly on hand-built params — don't see it.

Fix: either clamp the unscaled final vector to `bounds` before `unpack_params`, or compare with a ULP / tiny-relative tolerance instead of `==`.

### 2. `src/api/postfit.rs:1244` — the message and remediation only describe the *upper* rail, but lower-guard hits are the common case and mean the opposite thing

A free coordinate landing on `-8.0` (sigma → `3.35e-4`) or `-6.0` (Ω diagonal → variance `6.1e-6`) is a parameter *collapsing to zero*, not a runaway. That happens routinely — the additive component of a combined error model on concentration data, or an IIV the data can't support — and those users now get "Internal optimizer runaway guard reached … The fit reached an implementation safety limit … do not treat the affected value(s) as reliable estimates."

The right advice for a lower hit is "remove this random effect / error component", not "revisit initial estimates or method". Suggest branching the message text on `hit.side`; the `docs/warnings.qmd` row has the same single-sided framing.

### 3. `src/api/postfit.rs:1144` — theta is excluded on the premise that theta bounds are user-declared, but ferx's theta bounds are only partly user-declared

`compute_bounds` applies its own implementation caps to theta: `theta_lower.max(1e-10).ln()` and `theta_upper.min(1e9).ln()` (`parameterization.rs:361-363`), and the parser defaults an undeclared bound to `f64::INFINITY` (`model_parser.rs:2363/2376`). So a theta with no user-declared upper bound that runs away lands on the hidden `ln(1e9)` rail and is reported as `boundary_estimate` with the remediation "relax the bound" — a bound the user never wrote and cannot relax.

That is exactly the class of hidden guard #1099 is about, and it stays mis-categorised. Either route the `INFINITY` / `0.0`-defaulted theta rails to the new code, or note the gap explicitly.

roninsightrx:
Addressed all three findings in 52205e65.

1. **Scaled-space rail hits:** replaced strict equality with a four-machine-epsilon packed-space comparison. This covers the concrete `(-8 / ln(10)) * ln(10)` unscale case while a value `1e-12` inside the rail remains unflagged. Added a regression test using that exact scale factor.
2. **Lower versus upper remediation:** warning text now branches by side. Lower-only hits say the component collapsed toward zero and suggest removing/simplifying it; upper-only hits describe a runaway estimate; mixed hits include both remedies. Docs and tests cover all three cases.
3. **Hidden theta caps:** log-packed theta hits on ferx's implicit `1e-10` / `1e9` caps now route to `parameter_at_runaway_guard`. Genuine finite user-declared theta bounds remain on `boundary_estimate`, and FIX'd theta caps remain excluded. Added regression coverage for lower cap, upper cap, user bound, and FIX exclusion.

Validation after the fixes:

- `cargo test --lib`: 3,531 passed, 22 ignored
- `cargo clippy --lib`: passed with existing repository warnings
- `quarto render docs`: passed

The PR title/body and changelog/docs have also been updated for the expanded hidden-theta scope. `converged` behavior remains unchanged and stays with #1098.

PR #1103, opened 2026-08-27
fix(estimation): demote VI bad-basin convergence [closes #1098]

### LINKED ISSUE #1098: vi: the bad-basin detector warns but does not demote `converged`, unlike the adjacent noise-floor detector

roninsightrx:
> [!IMPORTANT]
> Stacked on #1017, whose current head is this branch's parent commit. Keep this PR draft until #1017 merges; the diff against `main` will then collapse to commit `2f5716b4`.

## Why

VI already detects an implausibly loose final ELBO and warns that the optimizer almost certainly settled in a bad basin. Despite that diagnosis, the fit still returned `converged: true`, so programmatic consumers could retain a fit known to be unusable. This closes #1098.

## What changed

- Demote `converged` whenever the final ELBO tightness diagnostic is implausible.
- Suppress the generic “increase `vi_iters`” warning for this failure mode because more iterations are not the relevant remedy.
- Emit `W_VI_BAD_BASIN` and classify it as the critical, machine-readable `vi_bad_basin` warning code.
- Add focused regression tests for demotion, warning selection, severity/category, and the healthy-control path.
- Document the warning code and add the `[Unreleased]` changelog entry.

## Alternatives considered

Classifying the existing prose as ordinary `convergence` would make the boolean accurate but would lose the distinction between an exhausted iteration budget and a fit trapped in a bad basin. A dedicated warning code preserves actionable machine-readable semantics.

## Cross-repo dependency
| ferx (R) | follow-up not yet open | required after core lands | after |

The core change is additive and ferx-r already transports unknown structured warning categories. A follow-up should add `vi_bad_basin` to ferx-r's guidance vocabulary when its ferx-core lock is bumped.

## Breaking changes
- [ ] None

Additive, non-breaking API change: `WarningCode` is `#[non_exhaustive]`, and this adds the `ViBadBasin` variant / `vi_bad_basin` serialization token.

<details>

No migration is required. Consumers should continue handling unknown warning codes generically.

</details>

## Numerical validation

```text
Not applicable: this changes result reporting after estimation and does not alter estimates,
the ELBO, OFV, residuals, or diagnostics.
```

## Performance

The ELBO tightness calculation already ran; the patch only changes handling of its result.

## Tests

Verification:

- `cargo fmt --check`
- `cargo test --lib`: 3,641 passed; 23 ignored
- `cargo check --tests`
- `cargo clippy --lib` completes with the branch's existing warnings
- `cargo clippy --all-targets` remains blocked by four pre-existing `clippy::approx_constant` errors in `src/sens/two_cpt_explicit.rs`

<details>

```text
# Before
converged: true
warning: VI: the ELBO is not a usable bound at this estimate ...

# After
converged: false
warnings_structured:
  - severity: Critical
    category: vi_bad_basin
    message: "W_VI_BAD_BASIN: VI: the ELBO is not a usable bound at this estimate ..."
```

</details>

## Docs
- [ ] No user-visible change

## Checklist
- [ ] `cargo clippy` clean

No gradient-reachable code is changed, and the additive `#[non_exhaustive]` warning variant is not breaking, so no version bump is needed.

### ferx-site

No DSL, example, or data changes.

### ferx-book

No new estimator, DSL feature, or fit option.

- [ ] All affected `examples/*.ferx` run cleanly via CLI: `cargo run --release -- examples/<model>.ferx --data data/<data>.csv`
- [ ] No example execution step needed (internal refactor / docs-only / no user-visible change)

No example execution step applies: this is a reporting-only fix exercised by the unit regression tests above; it changes no model, estimator calculation, DSL, or example.

## Reviewer hints

Review commit `2f5716b4` while #1017 is still open. The key policy is in `run.rs`: the tightness check now runs before the generic non-convergence warning, demotes the boolean, and records a dedicated stop reason so the generic advice is skipped.

## Open questions

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Fixed
+- **VI's bad-basin detector now reports `converged: false` (#1098).** An implausibly loose
+  final ELBO already identified a fit trapped far from a usable variational approximation, but
+  the result still exposed `converged: true` to programmatic consumers. Such fits are now demoted,
+  omit the misleading "increase `vi_iters`" warning, and carry the critical structured warning
+  code `vi_bad_basin`.
 - **`methods = [vi, laplace]` with `n_agq > 1` is no longer rejected (#1017).** `n_agq` is a
   chain-wide option, so the documented VI readout — `methods = [vi, laplace]`,
   `agq_eval_only = true`, which turns VI's ELBO lower bound into a real `−2 log L` — carries it
--- a/docs/warnings.qmd
+++ b/docs/warnings.qmd
@@ -45,6 +45,7 @@ Each `WarningCode` token, what it flags, and a typical response.
 | `covariance_step` | Info | Covariance-step informational note (e.g. evaluation cost). | None. |
 | `condition_number` | Critical | Ill-conditioned covariance / high condition number. | Over-parameterized — drop a random effect or covariate. |
 | `optimizer_health` | Warning | Trust-radius collapse / degeneracy. | Try a different optimizer or re-scale parameters. |
+| `vi_bad_basin` | Critical | VI's final ELBO tightness check found that a flat objective is an unusable variational bound, not convergence. | Reject the fit; use better starting values or initialize VI from a fitted point with `method = [focei, vi]`. |
 | `dw_autocorrelation` | Warning | IWRES autocorrelation (Durbin–Watson out of range). | Revisit the residual-error / structural model. |
 | `eta_normality` | Warning | ETA departs from normality. | Consider a mixture or transform. |
 | `experimental` | Warning | An experimental feature (SDE, neural-network) was used. | Treat results with caution; validate. |
--- a/src/estimation/vi/run.rs
+++ b/src/estimation/vi/run.rs
@@ -43,7 +43,7 @@ use crate::types::{
 use super::adam::{averaging_start, AdamConfig, AdamState, PolyakAverager};
 use super::elbo::{
     closed_form_omega, closed_form_sigma, closed_form_sigma_support, population_neg_elbo,
-    stacked_prior, unsupported_data_term_reason, ElboConfig, Families, PackedLayout,
+    stacked_prior, unsupported_data_term_reason, ElboConfig, ElboTightness, Families, PackedLayout,
 };
 use super::family::{FullRank, MeanField, VariationalFamily};
 
@@ -317,6 +317,40 @@ pub(crate) fn kl_fallback_warning(
              vi_mc_samples, or set vi_kl = mc to silence this."
         )
     })
+}
+
+/// Demote a fit whose final ELBO diagnostic says the optimizer stopped in a bad basin.
+///
+/// Kept separate from [`ElboTightness::is_implausible`] because that method diagnoses the
+/// bound, while this one owns the user-visible consequence: a flat objective in the wrong
+/// basin is not convergence.
+fn bad_basin_warning(converged: &mut bool, tightness: &ElboTightness) -> Option<String> {
+    if !tightness.is_implausible() {
+        return None;
+    }
+
+    *converged = false;
+    Some(format!(
+        "W_VI_BAD_BASIN: VI: the ELBO is not a usable bound at this estimate — the data term \
+         sits {:.0}x further above its value at the variational means than a posterior-shaped \
+         q would put it ({:.3e} against an expected {:.3e}). The optimizer has almost \
+         certainly settled in a bad basin. Reported converged: false. Re-run from better \
+         starting values (for a [covariate_nn] model, declare `init` so the network starts at \
+         plausible parameter values), or chain `method = [focei, vi]` to start VI from a \
+         fitted point.",
+        tightness.ratio(),
+        tightness.excess,
+        tightness.expected
+    ))
+}
+
+/// Whether a generic iteration-budget warning adds information after quality checks.
+fn needs_iteration_budget_warning(
+    converged: bool,
+    noise_floor_stop: bool,
+    bad_basin_stop: bool,
+) -> bool {
+    !converged && !noise_floor_stop && !bad_basin_stop
 }
 
 fn build_family(kind: ViFamily, d: usize) -> Box<dyn VariationalFamily> {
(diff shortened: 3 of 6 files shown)


roninsightrx:
## Code review (medium effort) — commit `2f5716b4`

The move itself is sound: `converged` isn't read between the old and new positions of the tightness block, and the new `classify_warning` arm is first, so no other arm shadows `w_vi_bad_basin`. Three findings.

### 1. Suppressing the iteration-budget advice on a loose bound can hide the right remedy — `src/estimation/vi/run.rs:940`

`needs_iteration_budget_warning(converged, noise_floor_stop, bad_basin_stop)` drops the "increase `vi_iters`" warning whenever the tightness ratio exceeds 25. But a run that simply hit the `vi_iters` ceiling far from the optimum *also* has a loose bound — this repo's own test comment at `src/estimation/vi/run_tests.rs:1163` cites a stopped-short warfarin run with `elbo_tightness_ratio: 78`. Such a fit now gets only "settled in a bad basin — re-run from better starting values", while `docs/estimation/vi.qmd:613` tells users the opposite: "If a VI fit reports non-convergence with an `elbo_tightness_ratio` far from 1, raise `vi_iters` before concluding anything."

The two conditions are not mutually exclusive. Suggest suppressing only when `converged` was still true *before* the demotion — i.e. when the iteration budget was not the binding stop.

### 2. Zero-random-effect models are demoted to `converged: false` with a Critical warning — `src/estimation/vi/run.rs:928`

`ElboTightness::ratio()` returns `f64::INFINITY` when `expected == 0.0` (`src/estimation/vi/elbo.rs:1218`), and `expected` is `Σ ½(n_eta + K·n_kappa)` (`elbo.rs:1295`), which is exactly 0 when `model.n_eta == 0` with no IOV. `run_vi` rejects only an empty population and non-Gaussian endpoints (`run.rs:482-488`) — not a model with no ETAs — so such a fit reaches `is_implausible() == true` with `excess ≈ 0` (finite, so the `is_finite()` guard doesn't catch it).

Before this commit that was a spurious warning; now it flips `converged` to false and emits `W_VI_BAD_BASIN` (Critical) on a perfectly healthy fit. Suggest guarding `is_implausible` with `expected > 0.0`, or returning `ratio = 1.0` for the zero-dimension case.

### 3. Docs left contradicting the new behaviour — `docs/estimation/vi.qmd`

Only `docs/warnings.qmd` was updated. Still stale:

- `vi.qmd:396-398` — "`converged` answers 'did the objective stop moving', which a stuck optimizer satisfies just as well as a successful one".
- `vi.qmd:391` — "a stuck optimizer also reports a covariance".
- `vi.qmd:405` — "Above 25 the fit emits a warning" (it now also demotes `converged`).
- `vi.qmd:411-413` — presents the DCM example reading 306.6 "while the fit reported `converged: true`", which after this change reports `converged: false`.
- `docs/model-file/covariate-nn.qmd:248` — same narrative in the table discussion.

PR #1104, opened 2026-08-27
fix(estimation): reject false VI convergence in bad basins [closes #1097]

### LINKED ISSUE #1097: vi: diverges on warfarin from ordinary starting values, pins sigma to exp(5) bound, reports converged
`method = vi` diverges on the warfarin PK model from ordinary starting values, pins σ to its runaway bound, and reports `converged: true`.

Found while benchmarking #1017 against FOCEI/SAEM across the `ferx-testdata` suite. Warfarin is the PR's own showcase model, and it reproduces there — but only from the starting values in `ferx-testdata/warfarin_pk/ferx/run1_focei.ferx`, which are further from the optimum than the ones in the PR description and `docs/estimation/vi.qmd`.

## Repro

Self-contained; data is `ferx-testdata/warfarin_pk/data/warfarin.csv` (10 subjects, 110 obs). This is the testdata model verbatim, with only `method` changed:

```
[parameters]
  theta TVCL(0.3, 0.001, 10.0)
  theta TVV(2, 0.1, 500.0)
  theta TVKA(1.0, 0.01, 50.0)

  omega ETA_CL ~ 0.1
  omega ETA_V  ~ 0.1
  omega ETA_KA ~ 0.1

  sigma PROP_ERR ~ 0.01 (sd)

[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV  * exp(ETA_V)
  KA = TVKA * exp(ETA_KA)

[structural_model]

[error_model]
  DV ~ proportional(PROP_ERR)

[fit_options]
  method       = vi
  vi_final_ofv = laplace
  covariance   = false
```

```
ferx warfarin.ferx --data warfarin.csv
```

## Observed vs expected

| | TVCL | TVV | TVKA | σ (PROP_ERR) | AGQ −2 log L |
|---|---|---|---|---|---|
| FOCEI | 0.132695 | 7.7377 | 0.81080 | 0.010565 | **−285.977** |
| SAEM | 0.132678 | 7.7363 | 0.81051 | 0.010331 | −285.896 |
| **VI** | 0.227873 | **2.6418** | 0.74477 | **145.837** | **+1307.919** |

A gap of **1594 units**. FOCEI and SAEM both converge correctly from this same start in under a second; VI stops after 1500 iterations (of a 25000 ceiling).

σ = 145.8 is not a plausible proportional-error SD, and at `vi_mc_samples = 128` it lands on **exactly `148.413159 = exp(5)`** — the σ upper bound from `compute_bounds` (`src/estimation/parameterization.rs`, `upper.push(5.0)`). The bound is doing its job; nothing reports that a parameter finished welded to it.

The `elbo_tightness_ratio` guard **does** fire here (1.0e4) with a well-written warning naming the bad basin, so this is not silent. But `converged` is still `true`, which is what a programmatic consumer of `FitResult` reads.

## Isolation

One factor varied per row, all VI, common AGQ −2 log L:

| variant | AGQ −2 log L | σ | tightness |
|---|---|---|---|
| far start, σ as `(sd)` | +1307.919 | 145.837 | 1.0e4 |
| far start, σ as variance | +1307.919 | 145.837 | 1.0e4 |
| far start, `vi_mc_samples = 128` | +1310.087 | **148.413159 = exp(5)** | 1.3e6 |
| far start, `vi_lr = 0.002` | — | **148.413159 = exp(5)** | 1.6e6 |
| only TVV moved to optimum | +583.132 | 3.926 | 23.8 |
| at-optimum start | −276.697 | 0.013728 | 0.972 |
| at-optimum, `vi_mc_samples = 128` | −285.512 | 0.011177 | 0.999 |
| at-optimum, `vi_mc_samples = 512` | −285.675 | 0.011062 | — |
| *FOCEI reference* | *−285.977* | *0.010565* | — |

Reading:

1. **The `(sd)`-vs-variance declaration is irrelevant** — those two runs are bit-identical. Not a parsing/scale bug.
2. **Start distance is the entire effect**, and it degrades smoothly rather than falling off a cliff — correcting only `TVV` gives a partial failure at σ = 3.93.
3. **σ init scale is not the driver** — starting at `sigma PROP_ERR ~ 0.5 (sd)` still ends at 140.6.

## The documented remedy does not apply here

`docs/estimation/vi.qmd` says:

> **If a VI fit lands well short of a FOCEI or AGQ fit of the same data, raise `vi_mc_samples` first.** More iterations will not help; the rule has already decided more epochs at this noise level cannot resolve the drift, and it is right.

From the bad basin that is not true: 8 → 128 draws does not help at all (it makes σ worse, reaching the bound exactly), and `vi_lr = 0.002` does not either. The advice is accurate **inside** a good basin, where the bottom half of the table shows it working exactly as described (−276.7 → −285.5 → −285.7). The docs currently promise it unconditionally.

Also confirmed on this model: chaining `method = [focei, vi]` is not a rescue. Starting from the converged FOCEI point, VI moves **downhill**, −285.977 → −276.697 (σ 0.0106 → 0.0137). The PR is candid about this for warfarin's σ; the benchmark found it generalises (cyclophosphamide: 3236.9 → 3269.0, a 32-unit loss from a good start).

## Suggested direction

Separable, roughly in order of value:

1. **Do not report `converged: true` when a σ or Ω coordinate is sitting on its bound.** Cheap, mechanical, independent of any tuning work, and converts this from "wrong answer labelled converged" to "wrong answer labelled as such". A dedicated warning code would let R-side consumers act on it.
2. **Qualify the `vi_mc_samples` guidance** in `docs/estimation/vi.qmd` to say it is the right knob once `q` is in the right basin, and to point at `elbo_tightness_ratio` for telling the two situations apart.
3. **Investigate why σ runs away at all** on a single-proportional-error model, where `vi_sigma_update = closed_form` is active and σ is supposedly the exact conditional maximizer each iteration. That is the part I did not get to the bottom of: an exact maximizer reaching `exp(5)` suggests the data term feeding it is being evaluated under a `q` that has already collapsed, rather than σ itself being mis-stepped.

Worth noting for triage that cyclophosphamide fails through σ too, from the *opposite* update route (`vi_sigma_update` falls back to Adam there — per-CMT error models). A σ-focused diagnostic may cover both.

## Environment

Branch `pr1017-vi` at `9e375c0e`, release build, macOS arm64 (Darwin 25.5.0), 8 rayon threads. Full benchmark across nine testdata models, method comparison, and the harness are written up separately.

Related: #276 (the feature request), #1017 (the implementation).

roninsightrx:
> [!IMPORTANT]
> This PR is stacked on #1017 (`vi-estimation-clean`) and is intentionally a draft until that PR merges. The current diff against `main` includes #1017; the reviewable change is commit `016a9883`.

## Why

Closes #1097.

VI can settle in a start-dependent bad basin on the warfarin model, drive the proportional-error sigma to its internal `exp(5)` runaway bound, and still report `converged: true`. Programmatic consumers therefore receive a successful status for an estimate that is roughly 1,594 objective units worse than the FOCEI reference.

The investigation found that the closed-form sigma update is behaving correctly conditional on the initial variational distribution: the far start predicts nearly zero late concentrations, so `E_q[(y-f)^2/f^2]` is enormous and its exact conditional maximizer is a very large sigma. That variance then flattens the data gradient before q can leave the bad basin. The existing `elbo_tightness_ratio` correctly diagnoses the result (`~1.0e4`) but previously did not override the trace-settling flag.

## What changed

- Treat `elbo_tightness_ratio > 25` as a fit-end convergence failure, not merely a warning.
- Detect free Omega/sigma coordinates pinned to ferx's internal runaway guards; fixed and structural-zero coordinates are excluded, and user-declared theta bounds keep their existing diagnostic.
- Emit both `boundary_estimate` and `convergence` warning categories for an internal Omega/sigma bound hit so callers can act on either condition.
- Document the distinction between a near-optimum Monte Carlo noise floor (increase `vi_mc_samples`) and a start-dependent bad basin (improve population starts or use FOCEI/SAEM).
- Add regression coverage and a changelog entry.

This deliberately does not relabel ELBO as an OFV or alter the VI update equations. It makes the returned status match the health information the estimator already computes.

## Alternatives considered

I evaluated tighter initial q covariance, delaying the closed-form sigma update by 500 and 5,000 iterations, mode/Laplace-shaped q initialization, damped sigma updates, lower learning rates, higher draw counts, and a full 25,000-iteration run. None provided a robust rescue from the far start; a damped full-budget run still ended near sigma 4.7 with a poor objective. Because the conditional sigma update is mathematically consistent with the bad initial q, tuning it would hide rather than fix the failure mode. Fit-end health gates are deterministic, cheap, and directly address the false-success contract.

## Cross-repo dependency

No public field or warning enum representation changes; only the value of `converged` and emitted warning categories are corrected for unhealthy VI fits.

## Breaking changes

- [x] None

<details>

Not breaking. Consumers should already treat `converged: false` as an unusable fit; this patch makes that flag truthful in the bad-basin and internal-bound cases.

</details>

## Numerical validation

- [ ] Warfarin example converges and estimates are within tolerance of prior run — the issue reproduction is intentionally expected **not** to converge.
- [x] Compared against: issue #1097 results, FOCEI reference, and prior ferx commit `9e375c0e`
- [x] Reference values:

```
# before (far-start warfarin VI)
TVCL 0.227873; TVV 2.641826; TVKA 0.744768; sigma 145.837003
Laplace -2 log L 1322.746; elbo_tightness_ratio ~10005; converged true

# reference (FOCEI from issue #1097)
TVCL 0.132695; TVV 7.7377; TVKA 0.81080; sigma 0.010565
-2 log L -285.977

# after
Numerical estimates are deliberately unchanged; elbo_tightness_ratio ~10005;
converged false, with convergence and boundary_estimate warnings.
```

The fix is classification/diagnostic correctness. It does not claim that the far-start variational optimization now reaches the FOCEI basin.

## Performance

The new work is one linear scan over the packed population vector at fit end.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 3,641 passed, 23 ignored)

Also verified `cargo check --tests`, the exact issue reproduction, and `quarto render docs`.

<details>

```toml
# No new syntax or option.
```

```
# A bad-basin VI fit now reports: converged: false
```

</details>

## Docs

- [ ] New pages linked in `docs/_quarto.yml` — no new page added
- [ ] No user-visible change

## Checklist

- [x] `cargo clippy` passes (repository baseline warnings remain)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — sensitivity/PK code is untouched

`cargo clippy --all-targets` additionally exposes four pre-existing `clippy::approx_constant` errors in test literals in `src/sens/two_cpt_explicit.rs`; none are on this diff.

### ferx-site

- [ ] New `.ferx` DSL syntax or `[fit_options]` key — not applicable
- [ ] New example `.ferx` file — not applicable
- [ ] New data file — not applicable

### ferx-book

- [ ] New estimator, DSL feature, or fit option — not applicable

- [ ] Rebuilt ferx-r against updated ferx-core — no cross-repo/API change
- [x] All affected `examples/*.ferx` run cleanly via CLI — no example files changed; exact issue model reproduced directly
- [x] All affected ferx-site example `.qmd` pages render cleanly — core docs rendered with Quarto; no site page changed
- [ ] All affected ferx-book chapters render cleanly — no book chapter changed

## Reviewer hints

The key review points are `variability_bound_hits` and `apply_convergence_health_gates` in `src/estimation/vi/run.rs`:

- the tightness threshold is the existing `ElboTightness::IMPLAUSIBLE_RATIO` (25), not a new tuning parameter;
- theta coordinates are skipped because declared theta bounds can represent valid constrained optima;
- fixed and structural-zero Omega/sigma slots are skipped because their immobility is intentional;
- boundary and convergence warnings remain separate so `classify_warning` exposes both stable categories.

The larger apparent diff is inherited from #1017. Review commit `016a9883`; docs and regression tests make up most of its line count.

## Open questions

None for this correctness fix. A more robust VI initializer could be explored separately; the experiments above did not identify a reliable tuning-only remedy.


roninsightrx:
Closing this as superseded — both halves of the change have since landed on `main` by other routes.

**Half A — demote `converged` when `elbo_tightness_ratio` is implausible.**
Landed as #1103 (`fix-1098-vi-bad-basin`). `main` now has the tightness gate at `src/estimation/vi/run.rs:328`, emitting `W_VI_BAD_BASIN` classified as `WarningCode::ViBadBasin`. This PR's diff still shows it replacing the old warn-only block only because the branch's merge base predates #1103; rebased, it is a duplicate that conflicts.

**Half B — Ω/σ coordinates pinned to an internal runaway guard.**
Landed earlier as #1099 (`4636c1cf`, 2026-08-26), and in a more general form: `runaway_guard_estimates` / `runaway_guard_warning` in `src/api/postfit.rs:1171`, wired into every fit at `src/api/fit.rs:2123`.

| | this PR (`variability_bound_hits`) | #1099 on `main` |
|---|---|---|
| scope | VI only, Ω/σ coordinates | all estimators, every packed coordinate incl. internal THETA caps |
| detection | relative tolerance 1e-3 of the bound range | `packed_guard_eq` on the packed coordinate |
| output | two prose warnings | typed `ParameterAtRunawayGuard` + structured `details` (parameter / estimate / packed_estimate / packed_guard / side) |

#1112 (VI gradient clipping) is orthogonal: it removes the *cause* that drove σ to the `exp(5)` guard on warfarin, so the gate rarely fires there now, but it neither adds nor removes detection.

**What this PR contributed that is genuinely still missing:** #1099 *warns* on a runaway-guard hit but does not demote `converged`. That gap is real, but it is not VI-specific — it belongs in `api/fit.rs` / `api/postfit.rs` for every estimator. Filed as #1118.

Worth recording for anyone revisiting #1097: this PR's gates would **not** have caught the remaining false success #1112 turned up — the `TVV = 7` start reports tightness 5.58 (below the 25 threshold, so gate A stays silent) with σ = 5.62 (interior; the guard is `exp(5)` = 148.4, so gate B stays silent). The backstop for that shape of failure is a median/MAD `trace_has_settled`, filed as #1119.

PR #1108, opened 2026-08-27
fix(ode): report failed auto fallbacks [#1080]

roninsightrx:
## Why

Follow-up to #1080 and the review notes carried there. When `ode_method = auto` rejected an
unusable stiff escalation, it returned the pinned-RK45 re-solve unconditionally. If that fallback
also exhausted `ode_max_steps`, stopped early, or returned non-finite saved values, production
diagnostics still described the retry as though it had repaired the segment.

This matters especially for the max-step-exhaustion shape documented in #1080: it can freeze-pad
a finite trajectory without ever clamping at `min_dt`, so the existing clamp counters cannot see
it.

## What changed

- Count every dense integration attempt that stops before its requested segment end.
- Attribute unfinished stiff attempts discarded by the `auto` guard separately from unfinished
  trajectories actually returned to callers.
- Check the explicit retry and increment `auto_fallback_failed` when it is unfinished or returns
  non-finite saved values.
- Surface total/discarded/kept unfinished counts and failed fallbacks in the structured
  `ode_solver` warning.
- Document that the old "explicit answer, twice the cost" bound applies only when the fallback
  completes successfully.

The returned trajectory remains deterministic: when both attempts fail, ferx keeps the pinned
explicit result rather than guessing at a third method, but now says plainly that neither attempt
was usable.

## Alternatives considered

- Return the rejected stiff result: worse, because it is the result the guard already proved
  unusable.
- Try a third solver automatically: rejected because solver ordering would be speculative and
  would add another discontinuous method choice to the objective.
- Infer failure only from clamp counts: insufficient because `ode_max_steps` can be exhausted
  with zero min-step clamps.

## Cross-repo dependency

No `FitResult`, serialized schema, or ferx-r glue changes are required.

## Breaking changes
- [x] None

<details>

Not applicable.

</details>

## Numerical validation

```text
Not applicable: this changes solver diagnostics only. The selected and returned trajectories are
unchanged; the previously unchecked explicit result is still returned when both attempts fail.
```

## Performance

The extra result check runs only after an already-rejected escalation and reuses the retry's
solver statistics.

## Tests

Validation run:

- `cargo test --lib`: 3,540 passed, 22 ignored
- `cargo test --lib ode::`: 260 passed, 3 ignored
- `cargo test --lib api::ode_solver_diagnostics_tests`: 13 passed
- `cargo check --tests`: passed
- `cargo clippy --lib`: passed with pre-existing warnings
- `quarto render docs`: passed

`cargo clippy --all-targets --all-features` remains blocked by five pre-existing
`clippy::approx_constant` errors in `src/sens/two_cpt_explicit.rs`; none is in this diff.

- [x] Not applicable (diagnostic fix with no new model-file surface)

<details>

```toml
# No new .ferx syntax.
```

```text
# The existing ode_solver warning now includes auto_fallback_failed and unfinished-segment counts.
```

</details>

## Docs
- [ ] New pages linked in `docs/_quarto.yml` (no new page added)
- [ ] No user-visible change

## Checklist
- [ ] `cargo clippy` clean — repository-wide all-target clippy has the unrelated pre-existing
  `approx_constant` failures noted above; `cargo clippy --lib` completes successfully
- [x] Functions on the `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay differentiable
  (the generic solver path and full library suite pass; this PR deliberately retains the guard's
  value-only scalar/Dual2 decision symmetry)

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key → not applicable; no new syntax or key
- [ ] New example `.ferx` file added to `examples/` → not applicable; no example added
- [ ] New data file added to `data/` → not applicable; no data added

### ferx-book
- [ ] New estimator, DSL feature, or fit option → not applicable

- [ ] All affected `examples/*.ferx` run cleanly via CLI: no examples affected
- [x] All affected ferx-site example `.qmd` pages render cleanly: full `quarto render docs` passed
- [ ] All affected ferx-book chapters render cleanly: no book chapters affected
- [x] No example execution step needed (diagnostic-only fix; no model/example input changed)

## Reviewer hints

The subtle invariant is attribution. `unfinished_segments` counts work actually attempted, including
the discarded stiff pass. `discarded_unfinished_segments` identifies the subset whose trajectory
did not reach the caller, so the warning reports their difference as `kept_unfinished_segments`.
This mirrors the existing `min_step_clamped_steps` / `discarded_clamped_steps` split and prevents a
rejected attempt from being mislabeled as freeze-padding in the returned result.

Please focus on `integrate_resolved_g_inner`'s separately-scored fallback and the subtraction in
`ode_solver_diagnostics_warning`.

## Open questions

None for this PR. The separate #1080 concern about finite `Dual2` values with non-finite derivative
jets remains intentionally open pending the reproducing system requested in the issue discussion.


roninsightrx:
## Code review — `fix(ode): report failed auto fallbacks` (#1108)

Reviewed the full diff (`CHANGELOG.md`, `docs/model-file/ode-models.qmd`, `docs/warnings.qmd`, `src/api/postfit.rs`, `src/api/tests/ode_solver_diagnostics_tests.rs`, `src/ode/solver.rs`, `src/ode/stiffness_tests.rs`) plus the surrounding solver drivers.

The accounting itself checks out: `unfinished_segments` / `discarded_unfinished_segments` are set once per `integrate_dense_g` call, `merge` destructures exhaustively so nothing was dropped, `saturating_sub` can't underflow (`discarded == attempt.unfinished` by assignment), and the guard's value-only `usable` decision is untouched, so `f64`/`Dual2` stepper symmetry is preserved. The remaining issues are attribution and scope.

### Medium

**`src/api/postfit.rs:1751` — the new `unfinished` clause double-reports segments the `aborted` clause already reports.**
The `stiff_abort_after` branch in `integrate_dense_g` is gated on `t < tf - 1e-15` and then breaks, so *every* `stiff_aborted_segments` increment is by construction also an `unfinished_segments` increment (and, when the attempt isn't discarded, a `kept_unfinished_segments` one). With `ode_stiff_abort_after = 5` and 3 aborts the warning reads "…3 returned segment(s) stopped before their requested end time…; 3 segment(s) were abandoned early by ode_stiff_abort_after = 5…", and the details payload shows `kept_unfinished_segments: 3` beside `stiff_aborted_segments: 3` with no documented relationship — a consumer summing damaged segments gets 6. The same overlap exists with the min-dt stall break (`record_min_step_failure` + `break`), whose `clamped` clause already states the tail was freeze-padded. The only genuinely new signal is `max_steps` exhaustion; consider subtracting `stiff_aborted_segments` from the reported count, or documenting the overlap.

**`src/ode/solver.rs:1713` — `usable` was not widened with `attempt.unfinished_segments`.**
The exact shape the PR's "Why" section calls out is still returned unguarded. A stiff escalation that exhausts `ode_max_steps` with zero min-dt clamps and finite freeze-padded output gives `stalled == false` and finite `out`, so `usable == true`, no fallback runs, `auto_stiff_rejected == 0` and `auto_fallback_failed == 0` — only `kept_unfinished_segments == 1`. The warning then emits the generic unfinished clause with none of the actionable "name `rodas5p`" advice, and the user cannot tell this from an unfinished *explicit* solve, whose remedy is the opposite (raise `max_steps` / loosen tolerance vs. pin a different stiff method).

### Low

**`src/ode/solver.rs:1745` — asymmetric failure criteria.**
The stiff attempt is rejected on *any* stiff min-dt clamp (`stalled`), but the fallback is only scored failed when it is unfinished or non-finite. A pinned-explicit re-solve that clamps at `min_dt` thousands of times yet happens to reach `tf` leaves `auto_fallback_failed == 0`, so the `rejected` clause still tells the user the guard repaired the segment; the only trace is `kept_clamped_steps`. Worth either matching the criteria or documenting the intent.

**`src/api/postfit.rs:1744` — dangling antecedent risk.**
The fallback clause opens with "of those explicit re-solves", whose antecedent is the `rejected` clause pushed above it. That holds only because `auto_fallback_failed` is set exclusively alongside `auto_stiff_rejected`; nothing enforces the invariant, and `parts.join("; ")` would produce a dangling "N of those explicit re-solves also failed" if a future stats source ever had `fallback_failed > 0` with `rejected == 0`.

**`docs/model-file/ode-models.qmd:328` — counter miscount and misplaced scope.**
The paragraph says "Two failure counters" and then introduces three (`unfinished_segments`, `discarded_unfinished_segments`, `auto_fallback_failed`), with a fourth derived key (`kept_unfinished_segments`) in the payload. It also sits inside the "Three counters record what `auto` did" section, but `unfinished_segments` is recorded under named methods too — which is exactly what the new `an_unfinished_named_solve_is_warned_without_an_auto_rejection` test pins. A user on `ode_method = rk45` who sees the key in `details` has no doc coverage for it.

**`src/ode/solver.rs:404` — missing "floor, not a total" caveat.**
Neither `unfinished_segments` nor `auto_fallback_failed` carries the caveat that `auto_stiff_segments` has. `solve_ode_until_threshold` takes no `OdeSolverStats` at all and does its own escalation/retry, so for a TTE / adaptive-dosing model whose likelihood is dominated by the event-time driver these counters are structurally zero — yet the new user-facing docs say "Do not treat a fit with a non-zero value as a usable result", which invites reading a zero as an all-clear.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1112, opened 2026-08-28
fix(estimation): clip VI gradients so Adam cannot freeze [closes #1097]

### LINKED ISSUE #1097: vi: diverges on warfarin from ordinary starting values, pins sigma to exp(5) bound, reports converged

roninsightrx:
## Why

Closes #1097.

`method = vi` on warfarin from the `ferx-testdata/warfarin_pk` initial estimates (`TVCL 0.3, TVV 2, TVKA 1.0, σ 0.01` — the same values NONMEM's control stream uses) did not fit badly. It **froze**.

Under a proportional error model `TVV = 2` makes late predictions ≈ 0, so `(y−f)²/(σf)²` explodes: iteration 0 evaluates at `−2·ELBO = 1.5e14`. Adam's second moment `v` absorbs that gradient. Because `v` averages *squares* and decays at `beta2 = 0.999`, it lands at ~`1e25`, giving `√v̂ ≈ 1e12`; once the fit recovers and ordinary gradients are ~`1e3`, every step is `lr·1e3/1e12 ≈ 1e-9·lr`, and shedding twelve orders of magnitude at 0.999/iteration takes tens of thousands of iterations. The optimizer stops moving, and since a frozen trace is a flat trace, the settling rule fires at the earliest iteration it is allowed to — 1500 of a 25 000 ceiling — reporting `σ` welded to its internal `exp(5)` guard, ~1594 objective units short of FOCEI.

**It was neither of the two things it looked like**, and both were checked before writing any code:

- **Not a bad basin.** Profiling `−2·ELBO` against `TVV` with everything else FIXed (q-only fits, early stopping disabled) is *monotone* from `TVV = 2` to `TVV = 20`: `410014 → −79.6`. There is no local minimum near the 2.475 the fit reported — the optimizer was walking away from a downhill direction.
- **Not `σ`.** The closed-form `σ` M-step is the exact conditional maximizer given `q` and was behaving correctly; a `q` predicting near-zero concentrations genuinely implies an enormous proportional `σ`. Pinning `σ` at the FOCEI value from the same start still left `TVV` at 2.83 after 25 000 iterations. `σ = exp(5)` was the symptom.

This answers #1097's "suggested direction" item 3, which #1103 (the `converged: false` demotion) deliberately left open.

## What changed

- `AdamConfig` gains `grad_clip`, a global-L2 gradient clip applied inside `AdamState::step`. If `‖g‖₂` exceeds it the whole vector is rescaled to that length — **direction is preserved exactly**, only magnitude is capped — so `v` is bounded by `grad_clip²`. Applied inline, so no copy of the gradient is made.
- New `[fit_options]` key `vi_grad_clip`, default `1e4`; `0` restores the previous behaviour.
- Clipping the **norm** rather than each coordinate is deliberate: a per-coordinate `clamp` shortens the largest components relative to the rest and points the step somewhere the gradient does not.
- Non-finite coordinates are excluded from the norm, matching `step`'s existing treatment of them as zero — otherwise one overflowing subject would drive the norm to infinity and scale every other coordinate to zero.

## Alternatives considered

- **`beta2 = 0.9`** instead. Also fully recovers the reference (`−284.04`, tightness 1.000) and is the confirming experiment for the diagnosis, but it changes Adam's character on every model to cure a transient. Clipping targets the failure directly and is a no-op once a fit is healthy.
- **Adaptive / AutoClip-style threshold** (percentile of the gradient's own recent history). Rejected on measurement: it is *worse than no clipping* here (`−2·ELBO` 2.4e9 at p=0.9). The early history **is** the pathology — when the first ten gradients are all ~`1e14` the threshold sits at `1e14` and nothing is ever clipped. Breaking the freeze requires an external scale.
- **Tuning the existing knobs.** `vi_mc_samples = 128`, `vi_lr` at 0.002 or 0.1, `vi_iters = 200000`, tighter `Ω` inits: all still land at `σ = exp(5)`. `vi_iters` is inert because once frozen the trace really has plateaued.
- **Chaining `[focei, vi]`** works (−284.51) but defeats the purpose of a standalone estimator on a 10-subject, 110-observation model.

## Cross-repo dependency

New `FitOptions` field with a default; no field is removed or renamed, and no output schema changes.

## Breaking changes
- [ ] None

<details>

`FitOptions` gains `vi_grad_clip: f64`. Callers using `..FitOptions::default()` (every call site in-tree, and the documented pattern) need no change; only code constructing `FitOptions` field-by-field must add it. `AdamConfig` likewise gains `grad_clip: f64` — internal, but `pub`.

VI numerical results change for fits that were previously frozen. That is the point of the PR, and such fits were reporting `σ` at an internal guard. Fits that already converged are unchanged (see below). `vi_grad_clip = 0` reproduces the old trajectory exactly.

</details>

## Numerical validation

- [x] Compared against: **NONMEM 7.5.1 FOCEI**, `ferx-testdata/warfarin_pk/nm/run1_focei.lst` — same data (`data/warfarin.csv` is byte-identical to that folder's copy) and the *same initial estimates*
- [x] Reference values:

```
# NONMEM FOCEI (nm/run1_focei.ext, final iteration)
TVCL 1.32695e-1  TVV 7.73770  TVKA 8.10795e-1  SIGMA(1,1) 1.11621e-4 (sd 0.010565)
OMEGA 2.85923e-2 / 9.59216e-3 / 3.35872e-1     OFV -286.004219

# before (vi_grad_clip = 0)
TVCL 0.243099  TVV 2.475030  TVKA 0.793892  sigma 148.413159  = exp(5), the runaway guard
-2*ELBO 3.6e7  elbo_tightness_ratio 1205200  stopped at iteration 1500 of 25000

# after (vi_grad_clip = 1e4)
TVCL 0.132693  TVV 7.737865  TVKA 0.810998  sigma 0.011617
-2*ELBO -284.612  elbo_tightness_ratio 0.994  converged, 10375 iterations
```

**Seed stability** — baseline fails on all four; clipped, all four land on the anchor:

| `vi_seed` | before: σ | after: TVCL | after: TVV | after: σ | after: −2·ELBO |
|---|---|---|---|---|---|
| default | 148.413159 | 0.132693 | 7.737865 | 0.011617 | −284.612 |
| 1 | 148.413159 | 0.132692 | 7.737893 | 0.011601 | −284.637 |
| 7 | 148.413159 | 0.132694 | 7.737868 | 0.011595 | −284.226 |
| 42 | 148.413159 | 0.132693 | 7.737910 | 0.011630 | −284.799 |
| 12345 | 148.413159 | 0.132693 | 7.737979 | 0.011609 | −284.179 |

**Threshold insensitivity** — every value works, only `0` fails, because Adam is scale-invariant in steady state:

| `vi_grad_clip` | 0 | 1 | 10 | 100 | 1e3 | 1e4 | 1e5 |
| −2·ELBO | 3.6e7 | −279.83 | −272.65 | −279.84 | −283.56 | **−284.61** | −284.59 |
| tightness | 1.2e6 | 1.012 | 1.014 | 1.014 | 0.996 | 0.994 | 0.995 |

**No regression on models that already converged** (`ferx-testdata`, VI on each):

| model | FOCEI ref | baseline | `vi_grad_clip = 1e4` |
|---|---|---|---|
| `propofol_schnider` | −3038.84 | −3050.291, 2500 it | −3050.291, 2500 it |
| `vancomycin_uvm` | 996.69 | 1001.295, 2000 it | 1001.295, 2000 it |

Identical to seven significant figures, as expected for a clip that binds uniformly or not at all.

Also fixes a silent false success this investigation turned up separately: starting the same model at `TVV = 7` previously returned `converged: true` with `elbo_tightness_ratio 5.58` — **below the bad-basin threshold of 25, so #1103's gate does not fire** — at `σ = 5.62` against a true `0.0106`, 530× off. With clipping that start gives `−283.41`, `σ = 0.0119`.

## Performance

One extra L2 norm per Adam step — a single pass over a vector the step already traverses. On the models above, iteration counts are unchanged (propofol, vanco) or the fit now converges where it previously ran to a premature stop.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: **3668 passed, 0 failed**)

Tier 1 (`src/estimation/vi/adam_tests.rs`, 5 new):
- `grad_clip_scale_caps_norm_and_preserves_direction` — 3-4-5 geometry; inside/on/outside the ball, and the component ratio is unchanged (the property a per-coordinate clamp destroys)
- `grad_clip_scale_disabled_is_identity`, `grad_clip_scale_ignores_non_finite_coordinates`
- `grad_clip_prevents_second_moment_poisoning` — the mechanism on synthetic gradients: a `1e14` spike then ordinary `1e3` ones. Measured over the steady state, because `m` is poisoned too but sheds eleven orders of magnitude in ~240 iterations while `v` needs ~5e4 to shed twenty-two — that asymmetry *is* the bug. Unclipped moves `<1e-6` over the window; clipped, `>0.1`.
- `grad_clip_is_a_no_op_when_it_binds_uniformly` — the scale-invariance claim the default rests on

Tier 1 (`run_tests.rs`, 2): `gradient_clipping_is_on_by_default` (with the measurement table in its doc comment, following `default_mc_samples_is_the_validated_value`), `vi_grad_clip_reaches_the_adam_config`.

Tier 1 (`model_parser_tests.rs`, 1): `test_apply_fit_option_vi_grad_clip` — parses, accepts `0`, rejects negative and malformed.

Tier 3 (`tests/warfarin_vi_nonmem.rs`): `ferx_vi_recovers_nonmem_from_the_testdata_start`, gated behind `slow-tests`, asserting against the NONMEM column already pinned at the top of that file. **Verified it passes** (`cargo test --release --features slow-tests`, 2.28 s) and it is the assertion that fails without this change. It exercises what the unit tests cannot: a real ELBO gradient large enough to poison `v`, from the actual predictor and residual model rather than a hand-written spike.

- [x] Not applicable (new fit option, documented; no new example model)

<details>

```toml
[fit_options]
  method        = vi
  vi_grad_clip  = 1e4   # the default; 0 restores the pre-fix behaviour
```

```
# with the ferx-testdata warfarin initial estimates (TVV = 2)
TVCL 0.132693   TVV 7.737865   TVKA 0.810998   PROP_ERR 0.011617
-2*ELBO -284.612   elbo_tightness_ratio 0.994   converged: true
```

</details>

## Docs

`docs/estimation/vi.qmd` gains a **Gradient clipping, and the freeze it prevents** section (`{#grad-clip}`) covering the mechanism, the NONMEM comparison, the two things it was *not*, why the threshold barely matters, and why an adaptive threshold fails. Both options tables updated.

Also **qualified an existing claim** the same investigation showed to be wrong. `vi.qmd` said unconditionally that a VI fit short of FOCEI/AGQ should have `vi_mc_samples` raised first, "more iterations will not help". #1097 already flagged this as false from a bad start; it is now scoped to the noise-floor case (tightness near 1) and points at gradient clipping for the case where tightness is in the hundreds or thousands.

## Checklist

- [x] `cargo clippy` clean — no new warnings on the changed files. `--all-targets` still surfaces 4 pre-existing `clippy::approx_constant` errors in `src/sens/two_cpt_explicit.rs` test literals (the same ones noted on #1104); none are on this diff.
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `sens/` and `pk/` untouched
- [ ] `Cargo.toml` version bumped if breaking — the `FitOptions` addition is source-compatible for `..Default::default()` callers

### ferx-site
- [x] New `[fit_options]` key — `vi_grad_clip` documented in `docs/model-file/fit-options.qmd` and `docs/estimation/vi.qmd`
- [ ] New example `.ferx` file — not applicable
- [ ] New data file — not applicable

### ferx-book
- [x] New fit option — the VI chapter should pick up `vi_grad_clip`; flagging for a follow-up there

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core — no cross-repo API change
- [x] All affected `examples/*.ferx` run cleanly via CLI — no example file changed; the #1097 reproduction and three `ferx-testdata` models run directly
- [ ] All affected ferx-site example `.qmd` pages render cleanly — no site page changed
- [ ] All affected ferx-book chapters render cleanly — no chapter changed

## Reviewer hints

The whole change is `grad_clip_scale` and the one line in `AdamState::step` that applies it (`src/estimation/vi/adam.rs`); everything else is plumbing, tests, and docs.

Worth focusing on:
- **The default of `1e4`.** It is an absolute constant, which normally deserves suspicion. The argument that it is safe is the threshold-insensitivity table plus Adam's steady-state scale invariance — a clip that always binds is a uniform rescale and leaves `m̂/√v̂` alone. The failure mode to think about is a much larger dataset where gradients are routinely above `1e4`: there the clip binds every iteration, which is normalized-gradient Adam. I validated on models up to ~2000 observations; I did not validate on one with ~1e5 observations.
- **Excluding non-finite entries from the norm** (`grad_clip_scale`) — this has to agree with `step`'s existing "treat as zero" rule or one overflowing subject silences every coordinate.
- The two things ruled out before the fix (monotone `TVV` profile; `σ` pinned) are the load-bearing claims for why this is the *right* fix rather than one that happens to work.

## Open questions

1. **Should `beta2` drop from 0.999 as well?** It independently fixes this and converges 13–30% faster on propofol and vancomycin_uvm. I left it alone because it is a broader behavioural change than this fix needs, but the speedup is real and might be worth its own PR.

2. **Two follow-ups this investigation surfaced that I have deliberately not bundled:**
   - The `TVV = 7` false success above (`converged: true`, tightness 5.58, `σ` 530× off) is fixed by this PR, but nothing *detects* it — a fit in that state with clipping disabled would still be certified. #1103's gate is at 25.
   - `trace_has_settled` uses mean/variance, so a heavy-tailed trace produces a huge `se` and any drift hides inside it. Every unhealthy run stops at the earliest arithmetically possible iteration. Swapping to median/MAD leaves healthy runs unchanged and correctly reports "not settled" on the failing trace at every checkpoint. Worth doing as a backstop, but it is not what fixes the estimates and would muddy this PR.

3. **`ferx-testdata/thioguanine_mmc` currently cannot be run at all** — its model needs `lndv` (only in `nm_6tg_20250106_clean.csv`) and `CONC0` (only in `nm_6tg_20250106_ferx.csv`). Unrelated to VI; filing separately unless someone knows the intended file.


roninsightrx:
Two findings before this is ready:

1. **[P1] Compute the clipping norm without overflowing** (`src/estimation/vi/adam.rs:68`). The norm currently squares each finite coordinate before summing. For any finite `|g| > sqrt(f64::MAX)` (or enough merely large coordinates), `g * g`/the sum becomes `inf`; the `norm.is_finite()` guard then returns scale `1.0`, so clipping is silently disabled. `AdamState::step` squares the same unscaled value again, sets `v = inf`, and that coordinate is frozen permanently. This is the exact failure mode the new guard is meant to prevent. Please use an overflow-safe scaled sum-of-squares (BLAS `dnrm2` style) and add a regression with an extreme finite vector such as `[f64::MAX, 1.0]`, asserting a finite nonzero scale and a finite clipped vector whose norm is capped.

2. **[P2] Remove or correct the claimed iteration-count speedup** (`docs/estimation/vi.qmd:414`). This says clipping makes `propofol_schnider` and `vancomycin_uvm` settle 13–30% sooner, but the PR's numerical-validation and performance sections say clipping leaves their iteration counts unchanged (2500 and 2000). The 13–30% speedup is attributed in the PR body to the separate `beta2 = 0.9` experiment, which this patch deliberately does not include, so the user-facing docs currently claim a benefit this change did not produce.

I otherwise found the option plumbing and regression coverage well targeted. Locally, `cargo test --lib grad_clip`, `cargo check --tests`, and the new slow warfarin regression all pass.

roninsightrx:
Both fixed in b3666136 — thanks, the P1 was a real hole.

**P1 — overflow-safe norm.** You were right that it defeats itself: `Σ g²` returns `inf` for a finite gradient, the finiteness guard hands back scale `1.0`, and `step` then squares the same unscaled value into `v = inf`. Clipping disabled precisely where it is load-bearing.

Switched to the BLAS `dnrm2` form — factor out the largest magnitude, accumulate `Σ (g/max)²` which is in `[1, n]` by construction. The full norm `max · r` is never formed, since that is the quantity that overflows: the comparison is arranged as `max > clip / r` and the scale as `(clip / r) / max`, so both sides only ever divide.

The same rewrite fixes the mirror case you did not mention — `Σ g²` flushes to **zero** below `√f64::MIN_POSITIVE`, reporting a zero norm for a tiny gradient. Covered too.

Two regressions added, and I confirmed both **fail against the previous naive norm** before keeping them:
- `grad_clip_scale_survives_gradients_that_overflow_a_naive_norm` — your suggested `[f64::MAX, 1.0]` plus `[f64::MAX, f64::MAX]`, `[1e160; 8]` (each square fine, the sum overflows), and a mixed non-finite vector. Asserts a finite non-zero scale `< 1`, a finite clipped vector whose norm is the clip, and — the part that matters — that `step` still moves the coordinate on the following iteration rather than freezing it.
- `grad_clip_scale_survives_gradients_that_underflow_a_naive_norm`.

**P2 — corrected.** The claim was wrong, and wrong in the way you identified: 13–30% was the `beta2 = 0.9` arm, which this patch does not include. Measured iteration counts with clipping are unchanged at 2500 (propofol) and 2000 (vancomycin_uvm), and the docs now say that. Only occurrence in the tree; the PR body already stated it correctly.

Verification after the change: `cargo test --lib` 3670 passed / 0 failed, `cargo fmt --check` clean, no new clippy warnings on the diff, and the slow NONMEM anchor `ferx_vi_recovers_nonmem_from_the_testdata_start` still passes — so the rewritten norm did not perturb the fit itself.

Issue #1121, opened 2026-08-29
bug(ode): an SS dose with a lagtime is equilibrated at the arrival, not seeded at the dose record — 7.67 OFV under time-varying covariates


TeunP:
## Summary

An `SS=1` dose carrying a **lagtime**, in a subject with **time-varying covariates**, is
propagated wrongly by the event-driven ODE walk: the periodic trough is equilibrated *at the
lagged arrival* under the dose row's covariates, instead of being seeded at the dose **record**
and propagated to the arrival under the record that governs that interval.

Measured against NONMEM 7.6.0, **7.67 OFV** on an eight-subject anchor.

The dense predictor does this correctly, and so does the analytic sensitivity twin. Only the
time-varying-covariate production path is missing the step.

## Measurement

`nonmem_anchor/dose_form_lag_ss.{ctl,csv}`, oral `depot → central`, `ADVAN13 TOL=9` / ferx
`[odes]`, `SS=1` with `II=12` at `t=0`, `ALAG1 = TVP·exp(ETA3)`, `WT` allometric on `CL` and `KA`,
`MAXEVAL=0` on both sides.

Two one-variable controls localise it — each ingredient is exact on its own:

| case | ferx | NONMEM 7.6.0 | Δ |
|---|---:|---:|---:|
| SS + lagtime, **flat WT** (`dose_form_lag_ss_flatwt`) | `2968.202993` | `2968.2029933781705` | ~1e-7 |
| SS + TV covariate, **no lagtime** (`dose_form_lag_ss_nolag`) | `1700.878933` | `1700.8789328578407` | ~1e-7 |
| SS + lagtime + TV covariate (`dose_form_lag_ss`) | `-382.381669` | `-390.0469773085871` | **7.67** |

Only the interaction is wrong. Subject 1's predictions are systematically **high**, worst at the
troughs — the signature of too little elimination across the pre-arrival window:

```text
 TIME     ferx      NONMEM     ratio
    1   0.976052   0.341450    2.86
    2   1.494753   1.411200    1.06
    4   0.866182   0.727490    1.19
    8   0.211986   0.112950    1.88
   11   0.073251   0.027542    2.66
   13   0.565788   0.165700    3.41
```

The U-shape is an additive excess carried from the trough: it dominates at both ends and is
diluted in the middle where the new dose's absorption is largest.

## Cause

`ode_predictions_event_driven` has no SS pre-arrival seed.

Both of the other implementations do:

[...]

Issue #1110, opened 2026-08-28
bug(ode): an [odes] RHS reading TAD together with a lagtime returns NaN predictions — and what TAD means before the first arrival is undecided


TeunP:
An `[odes]` model whose right-hand side reads the `TAD` built-in returns `NaN` for
every prediction as soon as the model also carries a lagtime. This is the plain
form of such a model — no time-varying covariates, no steady state, no IOV — and
there is no test covering the combination.

Found while building the FD gate for #1070; it is a separate defect (a **value**
NaN, not a wrong derivative) and it blocks that issue's NONMEM anchor.

## Reproduction

1-cpt IV, `LAGTIME = TVLAG * exp(ETA_LAG)`, `d/dt(central) = -(CL/V) * central * (1.0 + 0.3 * TAD)`,
bolus at `t = 0`, `lag ≈ 0.5`, observations at **2 / 4 / 8 / 24** — all long after the
arrival, no record anywhere near it:

```
predictions            = [NaN, NaN, NaN, NaN]
same model, no LAGTIME = [86.617, 70.212, 37.812, 0.224]
```

A zero coefficient does not save it: `0.0 * TAD` still yields `NaN`, because
`0.0 * NaN = NaN`.

## Mechanism

`tad_anchor` (`ode/predictions.rs:1641`) filters doses by
`d.time + lag <= t_start + 1e-12` and folds to `NEG_INFINITY` when none qualifies,
which the caller converts to `NaN`:

```rust
let last_dose_eff = subject.doses.iter().enumerate()
    .filter(|(i, d)| d.time + dose_lagtimes[*i] <= t_start + 1e-12)
    ...
```

The dense predictor (`ode_predictions`) begins its first segment at the dose
**record** time, so with a lagtime no dose has yet arrived and `t_start` sees an
empty filter. The injected `TAD` is `NaN`, the RHS evaluates to `NaN`, and the
trajectory is poisoned from there on.

The event-driven predictor pushes doses onto its timeline at `d.time + lag`
(`predictions.rs:4518`), so it starts *at* the arrival and escapes this — which is
why the defect only appears on the plain path. Production routes to event-driven
only for resets or time-varying covariates (`pk/mod.rs:2117`), so the simplest
model is the one that breaks.

## Related parity gap

For a no-covariate lagtime subject the analytic **sensitivity** walk takes the
event-driven route (`ode_tvcov_supported` admits on `model.has_lagtime()` alone)
[...]

Issue #1078, opened 2026-08-25
epic(sens): exact event-time saltations at every moving boundary on the ODE event-driven walk


TeunP:
Tracker for the five open defects in one mechanism: the **event-time saltations** the ODE event-driven walk injects at a moving boundary. Split out of #1060's review round so the ordering and the shared acceptance test live somewhere that does not scroll away. #486 stays the status *matrix*; this is the work *plan*.

## The mechanism

Under an estimated lagtime (or a modeled `D{cmt}`/`R{cmt}`) a dose's arrival `τ(η) = t_dose + ALAG`, an infusion's window edges, a zero-order window's edges and a per-route absorption onset are all **moving boundaries** — their position carries a jet. The RK45 timeline is `f64`, so each one needs an explicit saltation:

```
u += (v⁻ − v⁺)·δ + (½J⁻v⁻ + ½J⁺v⁺ − J⁺v⁻)·δ²
```

Every issue below is one term of that being wrong, missing, or double-counted. They are all **silent**: `δ` is jet-only (value 0), so predictions stay bit-exact and only the gradient/Hessian move — and the FOCEI objective builds its `h` matrix from that same analytic `∂f/∂η`, so a wrong term lands in the **reported OFV**.

## Acceptance

`docs/estimation/foce.qmd` currently reads:

> Three approximations remain in this cell, all narrow and all second-order or measure-zero: … (#1068) … (#1076) … (#1075).

This epic closes when that sentence can be deleted and the moving-boundary paragraph states exactness without qualification.

## Issues, in order

| # | issue | error | order | blocked by |
|---|---|---|---|---|
| 1 | **#1072** — value-only `pk_snapshot_equal` at `K_INF_END`/`K_ZO_END` | Hessian sign flip under IOV at `κ̂ = 0` | second | — |
| 2 | **#1069** — input-rate onset saltations carry the rate jump, not the field jump | unmeasured | first | — |
| 3 | **#1076** — two colliding moving boundaries each inject the whole field jump | **5.9 % gradient** | first | #1073, #1069 |
| 4 | **#1075** — `ẍ = ∂f/∂t + J·f`; the explicit time-derivative term is uncarried | **6.4 % Hessian** | second | #1076 |
| 5 | **#1068** — arrival landing exactly on a covariate-changing record | ~1 % gradient | first | #1073, #1076 |

### Why that order, edge by edge

[...]

PR #1123, opened 2026-08-29
fix(ode): govern each segment by the record that terminates it [closes #1073]

### LINKED ISSUE #1073: bug(ode): a lagged dose record's covariate snapshot is stretched to the lagged arrival, diverging from NONMEM by 14.9 OFV

TeunP:
## Why

A lagged dose's covariate snapshot was stretched forward to its arrival. ferx broke the
integration timeline only at `d.time + ALAG`, never at the dose row's own time, so the dose row's
covariate / IOV snapshot governed everything up to the arrival.

NONMEM evaluates `$PK` at every data record and then advances *to* that record, so the interval
containing a lagged arrival belongs to the record that **terminates** it. On the committed
multi-dose anchor (`tvcov_lag_saltation_multidose`) this was worth **14.89 OFV**.

Two corrections to the issue as filed, both established by reading the code before starting:

- **The closed-form engine had the same defect.** The issue scoped itself to
  `ode/predictions.rs`; it is four sites, not two.
- **`K_INF_END` is the wrong fix template.** The issue cites it as the precedent for "not a
  record", but it consumes `last_pk` — the *previous* record — while the fix needs the *next* one.
  Copying it reproduces the bug with the sign flipped.

## What changed

Each dose now contributes two timeline events: a `DoseRecord` at `d.time`, the only place it
supplies parameters, and the arrival, which is a state jump supplying none. A non-record event
takes the snapshot of the next record ahead, resolved in one backward scan.

Both events are pushed **unconditionally**. Skipping the record when `ALAG = 0` is the obvious
optimisation and it is wrong in the opposite direction: the arrival is no longer a parameter
source, so the segment ending at `d.time` would look forward *past* the dose. There is a
regression test for exactly that, and it fails under that mutation.

Dose *attributes* — `F`, `ALAG`, and the SS equilibration — are repinned to the dose row
explicitly. Before the split they were the same object as the segment snapshot and the
distinction did not show.

**The rule is not specific to arrivals.** Applying it to the other non-record boundaries — an
infusion end, a zero-order cutoff, a per-route onset falling strictly between two records — fixed
a second divergence of the same family in the ODE walk, measured against NONMEM 7.6.0 at **4.2 %**
on the predictions after the boundary and **23.5 OFV**. The closed-form walk already treated
infusion edges as sub-interval bounds inside one record interval, so the two production engines
now agree where they used to differ. Widening rather than special-casing was a measured decision,
not a guess — see *Alternatives*.

Fixed in **five** engines — both production predictors, both analytic sensitivity twins, and
the reactive adaptive-dosing walk — so FOCE/FOCEI, SAEM, VI, HMC and `simulate_adaptive` move
together. In the twins this retires the co-timed-sibling-dose exception in `post_snapshot` (both
sides of an arrival now read the same record, so the telescoping is trivial) and changes the
pre-side snapshot at the `K_INF_END` / `K_ZO_END` rate-offs from the previous record to the
enclosing one.

The fifth engine was found in review. `ode_predictions_adaptive_impl` resolved each segment
with an LOCF carry-forward and emitted no break at a plain dose row, so it could not express
the rule even in principle — which broke #391's **degenerate oracle** (a controller re-emitting
a fixed regimen must equal `simulate()` on that regimen) the moment production moved. Measured
on a base infusion whose window ends between two records under a changing covariate:

```
  t=6   adaptive IPRED  748.559090   vs   predict()  841.644811     11.1 %
```

It is fixed on both halves — the rate boundaries this PR moved *and* the dose row, which was a
pre-existing gap — because leaving one half converted would make that walk internally
inconsistent. A side effect worth naming: the resolution is now break-insertion invariant, so a
**holding** decision no longer perturbs the trajectory. `readout_pk` at a break's left boundary
stays on LOCF deliberately: that answers "the covariate *at* this instant" for a decision
readout and an injected dose's F, and looking forward there would let a controller read a
covariate that has not been recorded yet.

Keeping that promise took a second fix, found by an adversarial re-read rather than by a
failing test. Both adaptive walks advanced their LOCF carry with `last_pk = seg_pk` after
every segment — correct while the resolution returned `last_pk` unchanged at a non-record
break, which is what its comment said, but the #1073 resolution looks *forward* there. So the
carry began moving onto records the walk had not reached, and a decision landing on a
non-record instant read a covariate that had not been recorded yet. Measured on a model whose
`F` reads a covariate stepping 100 → 50 at `t = 24`, with a decision at `t = 12`: the injected
dose took `F = 0.4` instead of `0.8` — a factor of two in delivered dose, carried through every
later observation. The carry now advances only at an actual record, which also stops an
EVID=3/4 reset (not a record) from disturbing it, matching every other engine.

### One rule, one place

The resolution itself is `dosing::governing_record_indices(n, is_record) -> Vec<Option<usize>>`,
called by all four static engines. It resolves **indices**, not values, so no
`PkDual<Dual2<N>>` (~4.7 KB at `N = 8`) is materialised per event on the FOCEI gradient path.
Before this the four walks split two-and-two on the trailing-tail rule — the ODE pair kept the
previous record, the closed-form pair took the dose row — which is unobservable today only
because nothing follows such a boundary. They now share one answer: **keep the last record that
ran**.

### A NaN this introduced, and fixed

The new `(d.time, arrival]` segment starts before any dose has arrived, so the TAD anchor folded
to `NEG_INFINITY` and the guard injected `NaN`, which multiplies into the state (`0.0 * NaN`) and
poisons every prediction of an `[odes]` RHS reading `TAD` — a finite fit becoming the `1e20`
sentinel. Caught by a peer session working #1070; verified here by code read and reproduction
before acting on it.

The anchor now falls back to the subject's **first arrival**. Two properties are load-bearing and
both are mutation-tested: *finite* (restoring `NaN` gives `[NaN, NaN, NaN]`), and
*segment-invariant* — anchoring at the segment start instead restarts `TAD` at zero at each record
inside the pre-arrival window, and two subjects identical but for one extra sample there diverged
by 4.2e-4 at *every* later time. A prediction must not move because someone took an extra sample.

The dense predictor's `tad_anchor` moved onto the same fallback (review finding 2). Leaving it
on `NaN` would have made two subjects of the same model and the same data shape behave
differently — `compute_predictions_ode` selects between the two predictors on
`subject.has_resets()`, so the one carrying an `EVID=3/4` row would fit while its neighbour hit
the `1e20` sentinel. The analytic twin's dose-free anchor moved from `NEG_INFINITY` to `NaN` for
the same reason: `+∞` there would differentiate a different function than the value path
predicts.

This does not settle #1110, which owns what `TAD` *means* before a dose has arrived; it only
guarantees the value is finite, mesh-independent, and identical across both predictors
meanwhile. Related and **not** touched here: the sdtab `TAD` column (`io/output.rs`) is
computed independently and is lag-blind (`d.time`, not `d.time + ALAG`), so it disagrees with
the `[odes]` builtin by `ALAG`. Pre-existing, and #1110's subject.

## Alternatives considered

**Special-casing the arrival and leaving the other non-record boundaries alone.** Rejected on
measurement: a dedicated NONMEM probe (infusion end strictly between two records, `WT` step on the
later record, residual drug present) showed the ODE walk 4.2 % off while the closed-form walk was
bit-exact against NONMEM. Leaving that would keep the two production engines disagreeing with each
other, which `analytical_ode_equivalence` would surface later as an unrelated-looking failure.

**Splitting into two PRs (ODE, then closed form).** Rejected: the twins must move with their own
production engine or the `Dual2`-vs-FD gate compares a stale derivative, and VI joined the
analytic-provider consumer list in #1017, so a deferred twin now moves more estimators than it did
when the plan was written.

## Cross-repo dependency

No public Rust API change; `EventKind` gained a variant but it is crate-internal.

## Breaking changes
- [ ] Public Rust API
- [x] None (no API change — but **numerical results move**, see below)

<details>

This **moves converged fits** for any model combining a lagtime with time-varying covariates or
IOV, and for any ODE model whose infusion or zero-order window ends between records under a
changing covariate. Estimates from before this change are not comparable — re-run rather than
diffing. Models with constant covariates and no IOV are unaffected.

</details>

## Numerical validation

Every number below is NONMEM 7.6.0 via the InsightRX Docker image, `MAXEVAL=0` on both sides at
the same fixed θ.

**The issue's own anchor**

```
                        before        after         NONMEM 7.6.0
B  multi-dose crossing  -474.660106   -459.772592   -459.77259187257408   Δ 14.89 -> ~1e-5
A  single-dose control  -302.914703   -302.914703   -302.91470812769074   unchanged
D  agreeing-rows ctrl   -415.644111   -415.644111   -415.64410902342507   unchanged
```

Subject 1's `PRED` at `t = 7`, which is what localises it:

```
         ferx before   ferx after    NONMEM
  B        0.816054     0.846892     0.846890
  C        0.886124     0.846892     0.846890   <- C was NONMEM-invariant; ferx was not
  D        0.896650     0.896650     0.896650
```

C is the stronger half and is now asserted as an **invariance**: inserting a record inside
`[dose row, arrival]` carrying the same covariate as the record already governing the interval
must change nothing. That holds for a reason rather than by arithmetic coincidence, so it would
still fail a fix that happened to land on B's number.

**New dosing-form anchors** (`tests/dose_form_lag_nonmem_anchor.rs`)

```
  lagged infusion, window end between records   -436.381586  vs  -436.381586377188   ok
  EVID=4 reset carrying a lagged dose           -455.383750  vs  -455.383750284224   ok
  SS + lagtime, flat WT                         2968.202993  vs  2968.202993378171   ok
  SS + TV covariate, no lagtime                 1700.878933  vs  1700.878932857841   ok
```

**A gap this uncovered — filed as #1121, not fixed here**

SS + lagtime + TV covariate together sit **7.67 OFV** out. Each ingredient is exact alone (the two
SS controls above), so it is the interaction. Cause is structural and outside this issue: the
event-driven walk has no SS pre-arrival seed, which the dense predictor and the analytic twin both
have. Measured at the pre-#1073 baseline `6c9af38b` the same triple sat **10.91** out — so this
change *improves* it by 3.24 without closing it. The stream and its NONMEM result ship committed
so whoever picks it up starts from the measurement.

- [x] Compared against: NONMEM 7.6.0, and prior ferx commit `6c9af38b` for the baseline above

## Performance

One extra timeline event per dose and one backward scan over the timeline; both are O(events)
against segment integrations that dominate by orders of magnitude.

Two things got **faster** than before this PR, both from review:

- The closed-form schedule no longer builds propagation bounds for zero-length intervals. Every
  dose contributes a `DoseRecord → Dose` pair that is zero-length whenever `ALAG = 0`, and each
  cost an O(n_doses) scan plus an alloc, a sort and a dedup — making schedule construction
  O(n_doses²). Both walks skip these intervals anyway.
- The closed-form twin resolves record **indices** rather than materialising a
  `Vec<Option<PkDual<T>>>`. At `N = 8` that was ~4.7 KB per event, ~190 KB per subject per
  gradient evaluation in the FOCEI inner loop — and `cacheable_schedule` refuses lagtime models,
  so it was rebuilt on every call for exactly the models this PR is about.

## Tests

- [x] Unit tests added in the same module (`cargo test --lib --features ci` — **3698 passed, 0 failed**)

Nineteen new unit tests, each asserting the **segment's parameters** through a hand-computed
closed form or against an independent engine, rather than an aggregated objective — an OFV moves
for either half of the fix and isolates neither. All are mutation-tested; the mutation that
breaks each is recorded here.

**ODE production** (`src/ode/predictions_tests.rs`)

| test | mutation that must break it |
|---|---|
| `lagged_arrival_segment_runs_on_the_next_record_not_the_dose_row` | revert the lookahead |
| `a_dose_record_governs_the_advance_into_its_own_time` | sort `DoseRecord` after `Obs` |
| `a_zero_lagtime_dose_still_takes_its_own_row_for_the_incoming_segment` | skip the push when `lag == 0` |
| `an_infusion_end_between_records_runs_on_the_terminating_record` | revert the lookahead |
| `a_non_record_event_past_the_final_record_does_not_poison_the_walk` | revert the lookahead |
| `a_tad_reading_rhs_survives_the_pre_arrival_segment_of_a_lagged_first_dose` | restore the `NaN` anchor |
| `the_pre_arrival_tad_anchor_does_not_depend_on_the_sampling_mesh` | anchor at the segment start |

**The shared resolver** (`src/dosing.rs`) — five tests pinning "a record governs itself", "a
non-record takes the next record ahead", "the trailing tail keeps the last record that ran", the
adjacent-record and leading-non-record cases, and the no-record-at-all case.

The trailing rule lives here because it is **not** observable through predictions — a segment
ending after the final record precedes no observation by construction. That is why the test that
used to claim it (`…falls_back_to_the_previous_one`) had no mutation; it is renamed to the
robustness check it actually is, its *observable* half made non-degenerate, and it now carries a
note telling the next reader not to "strengthen" it by asserting a number no fallback can move.

**Closed-form production** (`src/pk/event_driven.rs`) — `ed_lagged_arrival_…`,
`ed_a_zero_lagtime_dose_…` and `ed_a_record_inside_the_dose_to_arrival_window_does_not_move_the_prediction`
(the invariance half, the same property anchor C pins against NONMEM).

**Both twins** — value-path-vs-production equivalence, which is the oracle `Dual2`-vs-FD parity
cannot supply: FD perturbs a twin's own value path, so a twin resolving a segment to a different
record moves both sides together and agrees on the wrong answer.

- `dual_walk_value_and_grad_match_production_across_a_lagged_arrival` runs
  `event_driven_sens_with_doses_g::<Dual2<1>>` against `event_driven_predictions`, seeding the
  jet on the **last** record's `CL` so the derivative pins *which* record the
  `(dose row, arrival]` segment resolved to.
- `ode_provider_stacked_non_record_boundaries_under_tvcov_match_production` puts a bolus arrival
  **and** a moving infusion end inside one record interval.

**Adaptive** — `adaptive_base_infusion_ending_between_records_matches_static_predict`, the
degenerate oracle for the geometry above, and
`adaptive_locf_carry_does_not_advance_past_a_non_record_break`, which pins the carry through
an injected dose's `F`. The frozen-replay verifier cannot see either: the driver and
`adaptive_frozen_replay_tv` share the resolution *and* the carry, so they agree with each other
whichever they pick.

### Mutations that initially survived

Three, all recorded because they are the failure mode this PR is about — a test that cannot fail
is not evidence.

1. Giving `DoseRecord` the same rank as `Obs` changed nothing: the stable sort plus insertion
   order kept it first anyway. Only a strictly-after rank pins it.
2. The first ODE-twin fixture put the infusion into `periph`, which `ALAG1` does not lag. Its
   rate-off is then a **fixed** boundary, `d_off` carries no jet, and the whole saltation is
   inert — reverting `pre_params` at `K_INF_END` to `last_params` (the 23.5-OFV mutation) passed
   all 160 `ode_provider` tests. The infusion must ride the same `ALAG1` as the bolus.
3. `post_snapshot`'s "continue past a non-record" needs **two** stacked non-records: with the
   arrival's successor already a record, the scan returns it whether or not it can skip.
   Mutating `_ => continue` to `_ => return None` survived the full suite until the fixture
   became `1.5 obs │ 1.751 arrival │ 1.901 infusion end │ 2.0 obs`.
4. A first attempt at the adaptive reset case added a short-circuit to the resolver and a test
   claiming to pin it. The mutation survived, because the segment *ending* at a reset is
   discarded — the reset zeros the state before any readout — so its resolution cannot reach a
   prediction. Both the short-circuit and that test were dropped; the reasoning is written down
   where the resolver is defined, so the apparent omission is not "fixed" again later. The
   property that *is* observable is the LOCF carry, and the replacement test fails under the
   mutation that restores the leak.

## Docs

- [x] `docs/` updated — `docs/estimation/foce.qmd` (the two-sided-boundary paragraph asserted the
      old convention explicitly and is now wrong; corrected, with the surviving co-timed-record
      case and the #1078 follow-ups restated), a new `docs/faq.qmd` entry with a worked
      example and the null test, and a new `### TAD before the first dose arrives` section in
      `docs/model-file/ode-models.qmd` for the behaviour change above
- [x] `CHANGELOG.md` `[Unreleased]` entry added, with the migration note

## Reviewer hints

Focus on:

1. **`post_snapshot` in `sens/ode_provider.rs`.** It is rewritten, not ported. The
   co-timed-sibling-dose exception it used to carry is deliberately dropped, and the scan now
   continues *past* non-records instead of stopping at one. The argument for both is in the
   comment; if the argument is wrong the gradient is wrong in a way FD parity cannot see, because
   FD perturbs the twin's own value path and both sides move together.
2. **The pre-side snapshots at the saltation sites.** Several `&pk_at_dose[idx]` arguments were
   *shadowing* no-ops before the split and are now genuinely different objects. The one at what is
   now `dose_params` was mixing a velocity from one snapshot with forcings from another.
3. **Sort ranks.** `DoseRecord` sits after `Reset` and before `Obs`; the datareader's existing
   one-ULP nudge for a co-timed observation (`io/datareader.rs`) is what carries file order, which
   I had wrongly assumed was broken until I measured it.

Skimmable: the anchor `.ctl`/`.csv` files and the generator, which are mechanical.

## Open questions

None blocking. Two things deliberately left alone and filed rather than half-fixed:

- **`Kind::Reset` still takes `last_pk`.** NONMEM runs `$PK` at an EVID=3/4 row, so this is a
  latent divergence of the same family. Fixing it inside this PR would have meant changing reset
  semantics under cover of a covariate-snapshot fix.
- **#1121** (SS pre-arrival seed), measured above.


TeunP (commented):
## Review — `bf4d8b36`

Read the four engines line by line against the record convention this establishes, plus the
enclosing functions of every hunk. `cargo test --lib --features ci` is green here (3665 passed,
0 failed, 23 ignored, 372s) — noting up front that this proves less than it looks, for the reason
in **3** below.

The core rule is right and the anchor set is well built. The two SS controls that isolate #1121
to the interaction, and anchor C asserted as an invariance rather than a number, are the strongest
part of this. What follows is what I think should change before it lands.

---

### Correctness

**1. The adaptive per-event TV walk is a fifth engine, still on the old convention.**
[`predictions.rs:3722`](src/ode/predictions.rs#L3722) resolves each segment with `segment_pk_at`,
which maps only obs / pk-only records and otherwise carries the **previous** record forward — and
[`build_segment_break_times:5331`](src/ode/predictions.rs#L5331) emits no break at a plain dose row
at all, so the walk cannot represent the new rule even in principle.

The arrival-side divergence there was pre-existing. The one at **infusion ends, zero-order cutoffs
and route onsets is new**: production moved to the terminating record, the adaptive walk still
takes the previous one. That breaks the degenerate oracle #391 is validated with — a controller
re-emitting a fixed regimen must equal `simulate()` on that regimen — for any ODE model with a TV
covariate and an infusion window ending between records. Not mentioned in the engine inventory and
not filed.

**2. `TAD` before the first arrival now differs between the two production ODE predictors.**
The event-driven walk falls back to the subject's first arrival
([`predictions.rs:4735`](src/ode/predictions.rs#L4735)); the dense predictor's `tad_anchor`
([`:1641`](src/ode/predictions.rs#L1641)) still answers `NaN`.

`compute_predictions_ode` picks between them on `subject.has_resets()`. So for an `[odes]` RHS
reading `TAD` with a lagged first dose, a subject **with** an EVID=3/4 row gets finite predictions
and its neighbour **without** one gets `0.0 * NaN` → the 1e20 sentinel. Same model, same data
shape. This is the cross-engine disagreement the *Alternatives* section rejects widening away
from, and it is a user-visible behaviour change with no CHANGELOG or docs entry.

**3. Three of the four engines carry no test that runs on this PR.**
All seven new unit tests call `ode_predictions_event_driven` only, and every new integration test
is `slow-tests`-gated — CI on `bf4d8b36` reports that job as `skipping`. So the rewritten
[`post_snapshot`](src/sens/ode_provider.rs#L4766), `next_record_params`, the `pre_params` swaps at
`K_INF_END`/`K_ZO_END`, the `arrival_post_params` fallback change, `next_record_pk_g`, and the
whole record split in `pk/event_driven.rs` ship untested here.

Reviewer hint 1 says a wrong `post_snapshot` is invisible to FD parity because FD perturbs the
twin's own value path. That is exactly right, and it is why the green FD tests in my run
(`ode_provider_matches_fd_under_the_auto_stiff_switch`, `tvcov_provider_matches_fd_of_production`,
the `iov_m3_ode_*_matches_reconverged_fd` set) are not evidence about these lines. The oracle that
*would* see it is a **twin-value-path vs production equivalence** check under a TV covariate + a
lagged dose — the twin's value path must equal `ode_predictions_event_driven` exactly, and a
mis-resolved segment breaks that. Same shape for `sens/propagate.rs` against `pk/event_driven.rs`.
Cheap, Tier-1, and it closes the gap the Codecov patch gate will flag anyway.

**4. The trailing-lookahead fallback is not the same rule in all four engines.**
[`event_driven.rs:795`](src/pk/event_driven.rs#L795) and
[`propagate.rs:1406`](src/sens/propagate.rs#L1406) fall back to `pk_for(ev)` — the **dose row** for
a `Dose` — while [`predictions.rs:4678`](src/ode/predictions.rs#L4678) and
[`ode_provider.rs:4868`](src/sens/ode_provider.rs#L4868) fall back to `last_pk` / `last_params`,
the **previous record**. The comment at `ode_provider.rs:4868` claims the two agree.

Unobservable today only because nothing follows such a boundary. Latent the moment a trailing
record, a reset, or the `with_states` second pass reads that state.

**5. The comment at the `K_DOSE` saltation still asserts the pre-#1073 rule.**
[`ode_provider.rs:5009`](src/sens/ode_provider.rs#L5009): *"The segment ending here uses the dose
row's own snapshot (the `params` binding above…)"*. `params` is now the next record's. The
fallback sentence three lines down was updated; this one was not — and it is the sentence a future
reader consults before touching these saltations. Restoring `&pk_at_dose[idx]` on its authority
would put the stretch back in the gradient only, where FD parity cannot see it and the value
anchors stay green.

**6. Dose-free subject: production `TAD` is `NaN`, the twin's is `+∞`.**
`first_arrival_ed.unwrap_or(f64::NAN)` vs the twin keeping `NEG_INFINITY`
([`ode_provider.rs:4799`](src/sens/ode_provider.rs#L4799)). Both lines were rewritten here and the
comment names the difference as intentional, but it means the twin differentiates a different
function than the value path predicts — invisible to FD for the usual reason. Reachable for a
placebo arm or a baseline-only subject in a joint model.

---

### Efficiency

**7. `next_record_pk_g` materialises a `PkDual` per event on the `Dual2` hot path.**
[`propagate.rs:1395`](src/sens/propagate.rs#L1395). `PkDual<Dual2<N>>` is 8 × (1 + N + N²) f64 —
~4.7 KB at N = 8, so ~190 KB of heap per call for a 40-event subject, once per subject per gradient
evaluation in the FOCEI inner loop. `cacheable_schedule` refuses lagtime models, so the schedule is
rebuilt every call for exactly the models this PR is about.
[`ode_provider.rs:4839`](src/sens/ode_provider.rs#L4839) already solves this the cheap way —
`Vec<Option<usize>>` plus a resolver. Same trick here, resolving through the existing `pk_for_g`.

**8. Bounds are computed eagerly for the new zero-length intervals.**
[`event_driven.rs:261`](src/pk/event_driven.rs#L261) builds `bounds_per_interval` for every
`windows(2)` pair, including the `DoseRecord → Dose` pair each dose contributes when `ALAG = 0` —
an O(n_doses) scan, a `Vec` alloc, a sort and a dedup that the walk then skips via
`if ev.time > cur_t`. That term is O(n_doses²), not the O(events) the Performance section claims.
A `w[1].time > w[0].time` guard removes it.

---

### Tests

**9. `a_non_record_event_past_the_final_record_falls_back_to_the_previous_one` is degenerate.**
[`predictions_tests.rs:7373`](src/ode/predictions_tests.rs#L7373) — every snapshot in the fixture
is `pk_one(cl_a, v)`, so no choice of fallback moves the asserted number. Which is why the mutation
table records "(robustness + exact value)" rather than a mutation: there isn't one. Either give the
final observation a different `CL` so the choice is observable, or rename it to the NaN/panic smoke
test it actually is.

---

### Cleanup

**10.** [`ode_provider.rs:5586`](src/sens/ode_provider.rs#L5586) — `let _ = dose_idx;` is dead;
`dose_idx` is still read three lines later. The comment above it also still says *"Fall back to the
dose snapshot if no later record carries params"*, contradicting the `unwrap_or(last_params)` it
documents.

**11.** [`ode_provider.rs:5406`](src/sens/ode_provider.rs#L5406) and `:5308` — with `last_dose_eff`
seeded to the first arrival, `is_finite()` is always true at a `K_DOSE`, so both `else { t_event }`
branches are unreachable and the comment still says *"On the first dose `last_dose_eff` is
`NEG_INFINITY`"*. Behaviourally a no-op only because the seed equals `t_event` at the first arrival
— worth stating that explicitly rather than leaving a guard whose invariant is undocumented.

**12.** [`predictions.rs:4480`](src/ode/predictions.rs#L4480) — `PartialEq, Eq` on `Kind` is unused;
`is_record` uses `matches!`. Dropping it keeps `==` from bypassing the predicate.

**13.** [`predictions_tests.rs:7402`](src/ode/predictions_tests.rs#L7402) — redundant parens around
the assigned value; `unused_parens` fires on every `cargo test --lib`.

---

### Checked and fine

Recording these so they aren't re-derived:

- The `ALAG = 0` bit-identity claim holds in all four walks — `DoseRecord` sorts first, the
  zero-length segment is skipped by `t_event > cur_t`, and the `Dose` arm no longer reads `pk_now`.
- `next_record_*` resolves a record to itself: the reverse scan sets `seen` before `acc[i]`.
- Sort ranks agree with production for infusion end and zero-order end.
- `post_snapshot` skips co-timed events by `INFUSION_EPS` while `next_record_params` skips by index.
  I could not construct a geometry where the two disagree — a record co-timed with an arrival does
  terminate the incoming segment and is not the post side — so this looks correct, but it is the
  #1068 coincidence and the asymmetry is worth a note in the code.
- Anchor `.ctl` / `.ferx` pairs match on θ, Ω, Σ and structure; both `nolag` controls differ by
  exactly one line.

TeunP:
## Review addressed — all 13, plus a bug the fixes themselves introduced

Every finding is fixed in code; nothing is deferred to a follow-up issue. Two of them
(**1** and **3**) turned out to be larger than they read. Four mutations survived a first
draft — three of the new tests, and one "fix" that turned out to be inert — and an
adversarial re-read afterwards found a **real bug I had just introduced**. All of that is
recorded below rather than quietly corrected, because it is the same failure mode the
review is about: work that looks done because nothing failed.

---

### 1 — the adaptive walk was a fifth engine on the old convention. Measured, then fixed.

Right, and worse than "latent": I built the discriminating fixture before touching anything
and the divergence is live. Base infusion, `rate = 200` over `amt = 1000`, so the window
`[0, 5]` ends between the `t = 0` record (`CRCL = 100`) and the `t = 6` record
(`CRCL = 50`); holding controller, so the degenerate oracle applies:

```
  t=6   adaptive IPRED  748.559090   vs   predict()  841.644811     11.1 %
```

Fixed rather than filed, and fixed on **both** halves rather than only the one this PR
introduced. Stopping at the rate boundaries would have left the walk internally
incoherent — rate-offs on the new rule, dose rows on the old — which is worse for the next
reader than either end.

- `governing_segment_pk_at` / `governing_segment_occ_at` replace the `segment_pk_at`
  LOCF carry at `t_end`, resolving a non-record boundary to the next record ahead.
- Base **dose rows** join the record index (production's `Kind::DoseRecord`), with a
  matching break time on the `tv` path only — the constant path keeps its byte-identical
  canaries, since there every snapshot is the same object anyway.
- `readout_pk` at `t_start` is deliberately left on LOCF. That one answers "the covariate
  **at** this instant" for a decision readout and an injected dose's F, where looking
  forward would let a controller read a covariate that has not been recorded yet.

A side effect worth naming: the resolution is now break-insertion invariant, so a
**holding** decision no longer perturbs the trajectory. That is what makes the oracle hold
at all, and it was not true before.

`adaptive_base_infusion_ending_between_records_matches_static_predict` pins it. Note the
frozen-replay verifier cannot: the driver and `adaptive_frozen_replay_tv` share the
resolution, so they agree with each other whichever record they pick. Only the static
engine is independent.

### 2 — `TAD` split between the two production ODE predictors

Fixed by moving the dense predictor onto the same rule, not by reverting the walk: the
event-driven fallback is there because the NaN poisons the state, so the fix has to go
forwards. `tad_anchor` now falls back to the subject's earliest lagged arrival and returns
`NaN` only for a genuinely dose-free subject.

You were also right that this is user-visible and undocumented. Added: a `CHANGELOG`
bullet, and a `### TAD before the first dose arrives` section in
`docs/model-file/ode-models.qmd` stating the sign convention, the mesh-independence
guarantee, and that #1110 still owns the semantics.

One thing I did **not** touch, flagging rather than fixing: the sdtab `TAD` column
(`io/output.rs`) is computed independently and is **lag-blind** — it uses `d.time`, not
`d.time + ALAG`. So under a lagtime the sdtab column and the `[odes]` builtin disagree by
`ALAG`. Pre-existing, unrelated to the record convention, and squarely #1110's subject.

### 3 — three engines shipping untested. The two mutations that survived.

Fixed, and the fixing is the interesting part. Twelve Tier-1 tests, all mutation-verified.
Two of the first-draft fixtures were **inert**, in exactly the way the review warns about:

**(a) A frozen infusion window makes the whole saltation inert.** My first ODE-twin fixture
put the infusion into `periph`, which `ALAG1` does not lag. Its rate-off is then a fixed
boundary, `d_off` carries no jet, and reverting `pre_params` at `K_INF_END` to
`last_params` — the exact 23.5-OFV mutation — passed all 160 `ode_provider` tests. The
infusion has to ride the same `ALAG1` as the bolus.

**(b) `post_snapshot`'s "continue past a non-record" needs two stacked non-records.** With
the arrival's successor already a record, the scan returns it whether or not it can skip.
Mutating `_ => continue` to `_ => return None` survived the full suite. The final fixture
stacks them inside one record interval:

```text
  1.5  obs (WT 70)  │  1.751 bolus arrival  │  1.901 infusion end  │  2.0 obs (WT 140)
```

Both mutations now fail on `df_deta` (`0.5647` vs FD `0.9872` for (b)).

The oracle is the one you named: `check_vs_production` compares the twin's own **value
path** against `compute_predictions_with_tv` before it compares derivatives against FD *of
that engine*. Same shape for the closed-form twin —
`dual_walk_value_and_grad_match_production_across_a_lagged_arrival` runs
`event_driven_sens_with_doses_g::<Dual2<1>>` against `event_driven_predictions`, seeding
the jet on the **last** record's `CL` so the derivative pins *which* record the
`(dose row, arrival]` segment resolved to. Reverting to `pk_for_g` fails it on the value
assertion.

### 4 + 7 + 9 — one rule, one place

These three collapsed into a single change. The resolution is now
`dosing::governing_record_indices(n, is_record) -> Vec<Option<usize>>`, called by all four
engines, resolving **indices** rather than values — which is also **7**: no
`PkDual<Dual2<N>>` is materialised per event on the FOCEI gradient path.

The trailing rule is settled as **"keep the last record that ran."** That is what the two
ODE engines already did, and it is the more faithful analogue — after the final record
there is no further `$PK` evaluation. The closed-form pair moved onto it.

**9** follows from that. You are right that there is no mutation for the old test: a
segment ending after the final record precedes no observation, so predictions cannot see
the choice. So I did both things you offered — renamed it to
`a_non_record_event_past_the_final_record_does_not_poison_the_walk`, made its *observable*
half non-degenerate (the two records now carry different `CL`, and it asserts the exact
value with a separation guard), and moved the trailing rule to where it **is** observable:
five unit tests on the resolver itself. The test now carries an explicit "do not
strengthen this by asserting a number no fallback can move" note, so the next reader does
not re-derive this.

### 5, 6, 10, 11, 12, 13

- **5** — comment rewritten; it now says `params` is the next record ahead and names
  restoring `&pk_at_dose[idx]` as the regression it would be.
- **6** — the twin's dose-free anchor is `NaN`, matching production. (`is_finite()` and
  `f64::max` behave identically for `NaN` and `NEG_INFINITY`, so the change is confined to
  the `eval_rhs_anchored` readout, which is the point.)
- **10** — `let _ = dose_idx;` gone, comment corrected to state the `last_params` fallback
  and why it agrees with the resolver's trailing rule.
- **11** — kept the guards, added `debug_assert!` and a comment stating the invariant
  explicitly: the seed makes them dead, and at the first arrival the seed *is* `t_event`,
  so they were a no-op before too. Reverting the seed now fails loudly instead of silently
  injecting `TAD = NaN`.
- **12** — `PartialEq, Eq` dropped from `Kind`.
- **13** — parens removed.

### 8 — the O(n_doses²) schedule build

Fixed with the `w[1].time > w[0].time` guard. An empty `Vec` allocates nothing and its
`windows(2)` is empty, so were it ever read it would correctly leave the state untouched
over a zero-length interval. Both walks skip these by the same `ev.time > cur_t` test, and
`cur_t` always equals the previous event's time, so the guard is exactly the walk's own
condition.

### What the adversarial pass afterwards found

Re-reading the fixes rather than the original code turned up one **real bug I had just
introduced**, and one fix of mine that was inert.

**The LOCF carry leaked into the future.** Both adaptive walks did `last_pk = seg_pk`
after every segment, with a comment saying it was a no-op at a non-record break —
true of `segment_pk_at`, false of the forward-looking resolution that replaced it. So
the carry moved onto records the walk had not reached, and `readout_pk` — deliberately
LOCF so a controller cannot read an unrecorded covariate — read the future. On a model
whose `F` reads a covariate stepping 100 → 50 at `t = 24`, with a decision at `t = 12`:
the injected dose took `F = 0.4` instead of `0.8`. The carry now advances only at an
actual record.

**And one that was not a bug.** I first "fixed" the reset case by short-circuiting the
resolver, on the reasoning that every other engine has an explicit
`Kind::Reset => last_pk` arm. The mutation survived: the segment *ending* at a reset is
discarded — the reset zeros the state before any readout — so which record governs it
cannot reach a prediction. I removed the short-circuit and the test that claimed to pin
it, and wrote the reasoning down where the resolver is defined so the apparent omission
does not get "fixed" again. The LOCF change is what actually makes a reset harmless
here, since a reset is not a record.

Also from that pass: the two `debug_assert!`s on the TAD seed are now conditioned on
`t_event` being finite. The seed is a `min` over the arrivals and `f64::min` ignores
`NaN`, so an optimizer excursion into a NaN lagtime would have panicked a debug-build
fit rather than degrading.

### Merge with #1070

`main` moved while this was in flight, and #1070 lands in the same TAD machinery. Both
conflicts (CHANGELOG, `foce.qmd`) resolved as unions — main's FD-gate paragraphs kept,
this branch's #1073 rewrite replacing main's stale #1060 moving-boundary text.

One semantic interaction, checked rather than assumed: #1070's anchors document "the
plain (dense) predictor returns NaN for a TAD-reading RHS under a lagtime" and force
the event-driven route to avoid it. That sentence is false as of finding **2** above, so
it is corrected in `tests/tad_lag_nonmem_anchor.rs`, both `tad_lag_*_fit.ferx` headers
and the `tad_gate_subject` doc comment. The `.ctl`/`.lst` pair is left alone — a `.ctl`
comment is echoed into its committed `.lst`, and editing one without re-running NONMEM
would desync them. The routing forcing is kept: it pins the engine the measurement was
made on. The anchor's numbers are unaffected, and measured so rather than argued — its
first observation is at `t = 2`, and while the integrator does traverse `(0, ALAG]`, the
compartment is empty there so the `TAD` term multiplies zero.

### Your "checked and fine" list

Kept, with one addition you asked for: the `post_snapshot`-skips-by-`INFUSION_EPS` vs
`next_record_*`-skips-by-index asymmetry is now noted in the code rather than left to be
re-derived.

---

**Verification** (on the merged tree): `cargo test --lib --features ci` — **3698 passed,
0 failed** (was 3677). All four `dose_form_lag`, all four `tvcov_lag_saltation`, all six
`ss_absorption` and both of #1070's `tad_lag` anchors green. `cargo check --tests` and
`cargo fmt --check` clean.

TeunP:
## Correction and two more gaps, from a mutation sweep over the fixes themselves

I claimed the review was fully addressed. Mutation-testing my *own* changes — not just the
new tests — found that three of them were unpinned, and that one of the thirteen findings
was not a real defect. Correcting the record.

### Finding 6 is a non-issue, and I said otherwise

I reported the twin's dose-free `NEG_INFINITY` as a genuine divergence — "the twin
differentiates a different function than the value path predicts". **That is wrong**, and
so was the code comment I wrote asserting it.

Every consumer of `last_dose_eff` gates on `is_finite()`, which is false for `NaN` and for
`NEG_INFINITY` alike. `eval_rhs_anchored` (`ode_provider.rs:3395`) injects `TAD = NaN`
either way; both saltation guards take their fallback either way; and the `max` at an
arrival is unreachable without a dose — and returns `t_event` for both regardless, since
`f64::max` ignores `NaN`. Reverting `f64::NAN` to `f64::NEG_INFINITY` passes all 447
`sens::` tests, and that is not a coverage gap: the two are observationally identical.

I have kept the `f64::NAN`, because writing the same literal production writes means a
future reader comparing the two sites does not have to redo this analysis. But the comment
now says it is a readability alignment and records why `NEG_INFINITY` would have been
equivalent, so the claim is not re-derived as fact later.

### Two of my fixes shipped unpinned

Both were the failure mode of your finding **3**, in my own work.

**The dense `tad_anchor` fix (finding 2) had no test.** Reverting it to `NaN` passed all
3699 lib tests — a user-visible behaviour change with nothing holding it. Now pinned by
`both_production_ode_predictors_agree_on_tad_before_the_first_arrival`, which calls both
production predictors directly on one reset-free subject with an observation inside the
`(dose row, arrival]` window. That is the real shape of the finding: the two are selected
per subject on `has_resets()` (`pk/mod.rs:2117`), so a divergence is a user-visible fork,
not a tolerance question. Under the mutation it reports `[NaN, NaN, NaN]`.

**The IOV occasion resolution had no test.** Reverting `governing_segment_occ_at` to the
pre-#1073 LOCF `segment_occ_at` passed all 220 adaptive tests. Now pinned by
`the_adaptive_segment_pk_and_occasion_resolve_to_the_same_record`, asserted at the
resolvers rather than through a simulation: `event_pk.obs[j]` already embeds occasion `j`'s
κ in its PK values, so if the two resolvers land on different records the segment
integrates PK carrying one window's κ against an `eta` carrying another's — incoherent
whichever window is "right", and invisible unless a record and a non-record break sit in
different decision windows.

### Where that leaves the mutation ledger

Every decision point this PR changes now has a mutation that kills a test:

| change | mutation |
|---|---|
| closed-form production lookahead | `pk_for` (dose row) |
| the `DoseRecord` push at `lag == 0` | skip it |
| closed-form twin lookahead | `pk_for_g` |
| ODE production lookahead | `last_pk` |
| `K_INF_END` pre-side | `last_params` |
| `post_snapshot` continue-past | `_ => return None` |
| the shared trailing rule | asserted directly on the resolver |
| adaptive segment resolution | `segment_pk_at` |
| adaptive LOCF carry | `last_pk = seg_pk` |
| adaptive occasion resolution | fall back to `last_occ` |
| dense `tad_anchor` | `f64::NAN` |

The one change with no entry is finding **6**, for the reason above — it has no observable
behaviour to mutate.

Issue #15, opened 2026-05-16
lagtime: SS=1 steady-state pulse train is not shifted by lagtime


roninsightrx:
## Summary

Discovered while landing [PR #12](https://github.com/FeRx-NLME/ferx-core/pull/12) (lagtime feature). When a model declares `LAGTIME` and a subject has steady-state (`SS=1`) doses, the SS pulse train is treated as **unshifted** — only the post-SS continuation is delayed by lagtime. This is silently wrong for those subjects.

`fit_inner` in `src/api.rs` pushes a warning into `FitResult.warnings` so the user knows, but the actual predictions are still biased.

## Impact

- Realistic NONMEM-style maintenance-dosing datasets (loading dose at `SS=1` to reach steady state, then probe) combined with a delayed-onset oral PK model will fit biased estimates.
- Severity scales with `lagtime / II` — a 0.5h lagtime on a q24h SS regimen is small; a 4h lagtime on a q12h SS regimen is substantial.
- No silent failure mode (the warning is loud) but the wrong-prediction outcome is silent.

## Where the bug lives

The SS handling is in the analytical superposition path inside `single_dose_concentration` / the per-model `*_ss` formulas (and analogues in `src/pk/event_driven.rs` and `src/ode/predictions.rs`). The lagtime shift in [src/pk/mod.rs:123](src/pk/mod.rs#L123) (`predict_concentration`) is applied to `dose.time` once — it does not propagate into the SS pulse-train accumulation that happens inside `single_dose_concentration` for `dose.ss = true` records.

## Suggested approach

NONMEM `ALAG` shifts each pulse in the SS train. To match: inside the SS branch of `single_dose_concentration` (and the corresponding event-driven / ODE branches), shift the per-pulse arrival time of each accumulated pulse by `pk_params.lagtime()`. The closed-form SS expressions need a derivation pass because shifting the pulse train changes the steady-state convolution sum.

Could potentially be expressed as: SS conc at `t` with lagtime `L` = SS conc at `t - L` without lagtime, provided `t > L`. Worth checking against NONMEM directly on a canonical example.

## Acceptance

- [ ] SS+lagtime predictions match NONMEM `ALAG1` on a 1-cpt oral SS example to numerical precision.
- [ ] Coverage on the analytical, event-driven, and ODE paths.
- [ ] Warning emitted from [src/api.rs:402](src/api.rs#L402) removed or reduced to an INFO-level note.

roninsightrx:
[...]

PR #1125, opened 2026-08-29
fix(ode): seed a lagged steady-state dose at its record, not its arrival [closes #1121]

### LINKED ISSUE #1121: bug(ode): an SS dose with a lagtime is equilibrated at the arrival, not seeded at the dose record — 7.67 OFV under time-varying covariates

TeunP:
## Why

An `SS=1` dose carrying a lagtime, in a subject with time-varying covariates, was propagated
wrongly: the periodic trough was equilibrated **at the lagged arrival**, computed throughout under
the dose row's covariate snapshot. NONMEM loads the compartments at the dose **record** (phase
`II − ALAG`) and advances to the arrival under the record that *terminates* that interval — the
#1073 convention. With flat covariates the two are the same number, because propagating from phase
`II − ALAG` for `ALAG` hours returns exactly to the trough, which is why this stayed invisible.

Measured on `nonmem_anchor/dose_form_lag_ss` (8 subjects, `MAXEVAL=0` at identical parameters):
**7.665 OFV**. Closes #1121.

**Stacked on #1123** (#1073) — the anchor datasets and `tests/dose_form_lag_nonmem_anchor.rs` only
exist on that branch. Do not merge before it.

### Three claims in the issue body are wrong, and two of them changed the fix

| issue says | measured |
|---|---|
| "the dense predictor does this correctly" | `reseed_prescheduled_states_at` seeds at the record **and then overwrites at the arrival**. It is correct only because the dense path has a single snapshot, so overwrite ≡ propagate. Copying its pattern verbatim would have been a no-op. |
| "the analytic sensitivity twin does this correctly" | **False — identical defect.** `ode_provider.rs` `K_DOSE` did `u = ss_trough_cache[idx].take()`, discarding everything `K_SS_SEED` and the walk had just propagated. |
| the issue's per-time table ("ratio 2.86 … 3.41") | It compares ferx **`PRED`** against NONMEM **`IPRED`**, conflating the state defect with EBE drift. Like for like at η = 0 the error is **2.1 %** at `t = 1` and **0.8 %** at the `t = 11` trough — small per point, worth 7.665 OFV once `ETA_LAG` absorbs it over 8 × 11 observations under a 5 % proportional error model. |

The issue also scoped the fix to the ODE walk. Measured, the **closed-form** engine returned
`-382.381669` on the same model — bit-identical to the pre-fix ODE answer. Fixing only the ODE walk
would have shipped two production engines 7.665 OFV apart, so both are fixed here.

## What changed

One shared predicate, `ode::predictions::ss_seeded_at_record(dose, lag)`
(`ss && ii > 0 && 0 < lag < ii`), read by **all four** walks so the seed and the arrival-time
equilibration stay exactly complementary — both firing loads the trough twice, neither leaves the
compartments empty.

- **`ode/predictions.rs`** — `Kind::DoseRecord` seeds `ss_state_at_phase(…, II − lag)` from the
  dose row's snapshot; `Kind::Dose` equilibrates only when not seeded.
- **`pk/event_driven.rs`** — same shape at `EventKind::DoseRecord`, and the post-hoc pre-arrival
  **prediction patch loop is deleted**. It overwrote `preds[j]` without ever touching the running
  state, and evaluated the whole pre-arrival decay under the dose row's snapshot.
- **`sens/ode_provider.rs`** — `K_DOSE` no longer overwrites `u` for a seeded dose; the
  `ss_trough_cache` (#642 review #4) has no reader left and is gone. Two compensations that existed
  *only* because the trough has `∂/∂lag = 0` are now scoped to **unseeded** SS doses: the
  `g_minus = 0` skip in the bolus saltation, and the hand-injected `−δlag` embedded jet in the
  infusion saltation. Once the state genuinely flows, the walk supplies that jet itself and leaving
  them in double-counts — but an unseeded SS dose (`lag == 0`, `lag >= II`) still meets the original
  premise, and `δlag` can be jet-nonzero where `lag` is value-zero, so that arm is reachable.
- **`sens/provider.rs`** — `ss_lagtime_walk_unsupported` **stays** (the closed-form *dual* still has
  no seed) with its rationale rewritten: it used to be justified by the overlay that no longer
  exists.

`predict_iov` gets a pre-arrival state as a side effect — it routes through the same walk and
previously returned zero for any observation between an `SS` dose's record and its arrival.

## Alternatives considered

Seeding *every* SS dose at the record with `ss_state_at_phase(…, II)` would unify the two branches,
but for `lag == 0` it integrates a full extra cycle and moves every existing non-lagged SS result by
solver error for no gain. `lag >= II` is excluded rather than approximated: phase `II − lag` is not
a phase, and flowing the bare trough forward across whole cycles skips their pulses — silently low
rather than obviously broken.

## Cross-repo dependency

No public API signature changed.

## Breaking changes

- [x] None

Numerical results move for models combining `SS=1`, a lagtime, and time-varying covariates or IOV —
re-run rather than diffing. Nothing else is affected.

## Numerical validation

NONMEM 7.6.0, `MAXEVAL=0` at the `.ferx` initial estimates. Four new runs committed under
`nonmem_anchor/results/`.

```
                                              before        after         NONMEM
E3   SS + lag + TV cov                   -382.381669   -390.046977   -390.046977308587
E3   same model, closed-form engine      -382.381669   -390.046977   (transfers via E3)
E3P  sample INSIDE the pre-arrival window        n/a   -380.654622   -380.654622895383
E3B  SS record mid-stream, not first             n/a   -476.729557   -476.729556897566
E5   SS INFUSION + lag + TV cov                  n/a   -458.896245   -458.896244660178

E2   lagged infusion            (control)  -436.381586   -436.381586   unchanged
E3a  SS + lag, flat WT          (control)   2968.202998   2968.202998   unchanged
E3b  SS + TV cov, no lag        (control)   1700.878939   1700.878939   unchanged
E4   EVID=4 reset + lag         (control)   -455.383751   -455.383751   unchanged
```

The algorithm was confirmed against NONMEM **by hand before any code was written**, from the 1-cpt
oral steady-state closed form at η = 0: `PRED(t=0) = 0.2868678` vs NONMEM `0.28687` (that row is the
seed and nothing else), and `PRED(t=1) = 0.955452` vs `0.95553` (seed → propagate → bolus →
propagate). The *buggy* value was predicted the same way at `0.976020`, against the `0.976052` the
engine actually produced.

## Performance

One `ss_state_at_phase` at the record replaces one `equilibrate_ss_state` at the arrival. The twin
drops one cached lookup and one `Vec` allocation per subject.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3679 passed, 0 failed)

**The EBE-free `PRED` oracle is the most useful thing here.** NONMEM's `$TABLE` carries `PRED`
(η = 0) and these runs are `MAXEVAL=0` at the `.ferx` inits, while `predict()` builds `zero_eta` and
walks the same dispatcher `fit()` uses. So the two sides evaluate the identical function at
identical parameters, **pointwise, with no empirical-Bayes step on either side**. Five such tests
run in ~1 s, are ungated, and carry the diff's coverage — the Tier-3 OFV anchors never run on a PR
and register none.

Layered so a partial failure localises rather than puzzles:

| test | what it isolates |
|---|---|
| `ss_state_at_phase_matches_nonmem_at_the_dose_record` | the **seed alone** — no walk, no propagation, no bolus. `t = 0` is `MDV=1`, so `predict()` cannot reach it. |
| `ferx_population_predictions_match_nonmem_*` (×5) | the whole chain, per observation, ~1e-4 relative |
| `ferx_matches_nonmem_*` (×5, slow) | the objective |
| `ferx_analytical_and_ode_engines_agree_*` | the two production engines against each other |
| `ode_provider_ss_lagtime_tvcov_matches_production` | production ↔ twin coupling |
| `ss_record_seed_declines_a_zero_lag_and_a_full_interval_lag` | the shared predicate's boundaries |

**On the coupling test — read its doc comment before trusting it as a regression test.** Both
engines carried the same defect, so a twin-vs-production comparison is green before *and* after a
correct joint fix. It is a **coupling guard, not an oracle**; NONMEM is the oracle. It was
mutation-tested: restoring only the twin's `K_DOSE` overwrite turns it red **on `obs.f`**, while the
flat-covariate `ode_provider_ss_lagtime{,_infusion}_…` tests go red only on their *derivative*
blocks. That split is why the cell had to exist — SS × lagtime × TV covariates was the one
combination with no test at all.

Steps had to land together: `inner_optimizer.rs:2322` builds the inner loop's IPREDs straight from
`sens.iter().map(|o| o.f)`, so fixing production alone would have driven the EBE search from the
twin's wrong predictions.

### Non-degeneracy

The original anchor's lagtimes are all in `[0.536, 0.951]` and its first observation is at `t = 1`,
so **no observation fell inside any subject's pre-arrival window** — it tested the propagation and
never the readout. E3P adds a sample at `t = 0.5` (inside the window for all eight subjects, and an
invariance check: the new record carries the `WT` already governing the interval it splits, so it
must not move anything else). E3B puts the `SS` row mid-stream with drug already present, where the
two conventions differ by a large factor rather than a few percent. E5 covers the SS **infusion**
branch of `ss_state_at_phase`, which crosses the rate-off boundary and had flat-covariate coverage
only.

## Example

## Docs

- [x] `docs/` updated — `faq.qmd` (a new section under the #1073 timing rule) and
      `model-file/lagtime.qmd` (SS × lagtime × TV covariates, and the unvalidated `lag >= II` case)

## Checklist

- [x] `cargo clippy` clean

## Reviewer hints

The subtle part is **`sens/ode_provider.rs`**, not the production seed. Three edits there, and the
last two are not in the issue: the `g_minus` skip and the `−δlag` injection are compensations for
`∂trough/∂lag = 0`, and their correctness now depends on *which* SS doses are seeded. Writing
`x⁻(lag) = Φ_lag(S(II − lag))`, the exact derivative is `g(x⁻) + Φ′·S′·(−1)`; the walk carries the
second term through the seed's phase jet and the saltation carries the first. Under flat covariates
they cancel, which is what makes the existing flat-covariate Hessian-vs-FD tests the live guard on
those two lines.

`nonmem_anchor/simulate_dose_form_lag_saltation_data.py` gained three emitters; the three existing
CSVs were verified byte-identical by md5 before and after regenerating.

Skimmable: the docs, the CHANGELOG, and the committed `.lst`/`.ext`/`.tab` outputs.

## Open questions

Three adjacent gaps found while doing this, none caused by it, none fixed here:

1. **`TAD` has no referent in the SS pre-arrival window.** The anchor filter is
   `d.time + lag <= cur_t`, so a lagged SS dose contributes nothing before its arrival — the state
   there is now physically real but `TAD` falls back to another dose or to the first arrival. The
   anchor models' RHS do not read `TAD`, so no test here can see it. Distinct from #1110: an SS dose
   *does* have a prior pulse, at `d.time + lag − II`, so the answer is determined rather than a
   convention.
2. **`lag >= II` is unguarded upstream.** Nothing rejects it; this PR declines to seed it and says
   so in the docs, but a validation error would be better.
3. **`propagate_g` has no `K_SS_SEED` analogue**, which is what keeps `ss_lagtime_walk_unsupported`
   in place. Adding one would let the gate be lifted.

Happy to file these as issues.


TeunP:
**CI note — this PR gets no automatic checks, by design of the workflow triggers.**

`ci.yml` is `pull_request: branches: [main]`, and this PR is stacked on
`worktree-fix-1073-lagged-dose-record` (#1123) rather than on `main`, so no workflow fires. That
base is deliberate: it keeps the diff to *this* issue's 12 files instead of replaying #1123's 38.

Two consequences worth stating rather than leaving to be discovered:

* CI was run manually instead, via `workflow_dispatch` on the branch —
  [run 33257255590](https://github.com/FeRx-NLME/ferx-core/actions/runs/33257255590). Note that a
  dispatch run is push-shaped, so anything gated on `github.event_name == 'pull_request'` (the
  Codecov *patch* status in particular) does not report. Patch coverage should be re-checked once
  this retargets.
* **Retarget this to `main` after #1123 merges**, at which point the normal PR checks attach and the
  diff narrows back to this issue's changes on its own.

Local verification in the meantime, all on this branch:

```
cargo test --lib                                    3680 passed, 0 failed, 23 ignored
cargo test --features slow-tests --test dose_form_lag_nonmem_anchor
                                                      15 passed, 0 failed
cargo check --tests --features survival,markov      clean
cargo fmt --all -- --check                          clean
cargo clippy --no-default-features --features ci,markov,nn   clean (CI's exact invocation)
```

One caveat on that last line: `cargo clippy --all-targets` reports 4 `approx_constant` errors in
`src/sens/two_cpt_explicit.rs` test data. They are pre-existing, in a file this PR does not touch,
and CI does not use that form.

TeunP:
## Review — xhigh, adversarial

Reviewed `ca2ea27e` against `bf4d8b36`. The seed convention is right and the E3/E3P/E3B evidence
is convincing. Two things measured on this branch are **regressions against the base**, and both
are invisible to every anchor in the PR. Reproductions below run in ~1 s each.

---

### 1. (blocking) An SS **infusion** whose previous cycle is still running at the record loses the residual rate

`ss_seeded_at_record` admits any `0 < lag < II`. `ss_state_at_phase`'s own doc comment states the
precondition it needs:

> For SS infusions this assumes `phase ≥ dose.duration` (the prior infusion has finished by
> `phase`), i.e. `lagtime ≤ II − dose.duration`

Nothing enforces it. When `lag > II − t_inf` the phase `II − lag` falls **inside** the previous
cycle's infusion window, so `active = phase.min(t_inf)` returns a state with the infusion still
flowing — and the record→arrival segment then integrates under `active_infusions`, which knows
nothing about a virtual previous-cycle window. The residual rate is dropped.

Measured, 1-cpt IV `[odes]`, `SS=1 AMT=600 RATE=100` (`t_inf = 6 h`), `II = 12`, one TV covariate
so the subject routes to `ode_predictions_event_driven`. Same steady state, two lagtimes,
compared at equal cycle phase:

```
                        phase 6.5      phase 9.0      phase 11.0
ALAG1 = 0.2  (11.8>6)  12.283345532   9.566279119   7.832206907
ALAG1 = 8.0  ( 4.0<6)  11.244655357   8.757346397   7.169908811
                        ---- 8.456e-2 relative, every point ----
```

On `bf4d8b36` the two rows are bit-identical (rel 1e-16).

Three further notes:

* **Both production engines.** `pk/event_driven.rs:835` reproduces the same 8.456e-2 via
  `ss_state_at_phase_event_driven`'s `bounds = [0.0, phase]` arm.
* **The twin agrees with the wrong answer.** `ss_state_at_phase_g` (`sens/ode_provider.rs:4127`)
  has the identical `phase_val.min(t_inf_val)` shape, so `check_vs_production` is green on it.
* **The static ODE path is untouched** and still returns `12.283345532` — so as of this branch the
  two production engines disagree by 8.5 % on one model, which is the exact failure mode the PR
  was written to end.

Fix is either direction: exclude `lag > II − t_inf` from `ss_seeded_at_record`, or carry the
previous cycle's residual window into the pre-arrival segment's forcing set. The first is a
one-line change and matches what the doc comment already claims.

<details><summary>repro</summary>

`tests/zz_probe.rs`, `cargo test --test zz_probe --no-default-features --features ci -- --nocapture`.
Model: `d/dt(central) = -(CL/V)*central`, `obs_scale = V`, `CL = TVCL*(WT/70)^WTEXP`, `ALAG1 = TVLAG`,
`ode_reltol 1e-11`. Data: `SS=1 II=12 AMT=600 RATE=100` at `t=0`, observations at `lag + {6.5, 9, 11}`,
all `WT=70`, plus one `EVID=2 WT=140` row at `t=500` — far past every observation, present only to
make `has_tv_covariates()` true and route the subject to the event-driven walk.

</details>

---

### 2. (blocking) `lag >= II` pre-arrival predictions regress from the SS tail to exactly `0.0`

The issue-#15 overlay being deleted covered `d.ss && d.ii > 0.0 && lag > 0.0` — including
`lag >= II`, where `ss_state_at_phase_event_driven` returns the bare trough for a non-positive
phase and the phase reconstruction for the last `II` hours. `ss_seeded_at_record` stops at
`lag < II`, and nothing replaces it, so the whole window falls back to the empty running state.

`pk one_cpt_iv(lagtime=ALAG1)`, `SS=1 AMT=600`, `II=12`, `ALAG1=15`, observations at `t=1` and
`t=10` (both inside `[0, 15)`):

```
              t=1            t=10           t=20 (post-arrival)
bf4d8b36     5.172153128    8.527438878    10.415437366
ca2ea27e     0.000000000    0.000000000    10.415437366
```

The alternatives section argues `lag >= II` is "silently low rather than obviously broken" — but
the shipped behaviour is *zero*, which under a proportional error model is the obviously-broken
one. It is also a **discontinuity in an estimated lagtime**: as `lag` crosses `II` during the outer
search, every pre-arrival readout jumps from a finite tail to `0`.

`ss_record_seed_declines_a_zero_lag_and_a_full_interval_lag` pins the predicate's boundary; nothing
pins the resulting prediction, which is why this got through.

`docs/model-file/lagtime.qmd` should say what actually happens rather than "unvalidated rather than
supported" — and open question 2 (a validation error) looks less like a nice-to-have now.

---

### Anchors missed

**E5 is degenerate on the branch it exists to cover.** `AMT 100 / RATE 100` gives `t_inf = 1 h`
against `II = 12` and `ALAG1 ≈ 0.7`, so the seed phase is `11.3` and `ss_state_at_phase` only ever
takes its `phase > t_inf` arm. The `phase ≤ t_inf` arm — `active = phase.min(t_inf)` plus
`extend_flow_by_dual_duration_g`, real code with no other coverage — is never entered. That is
precisely where finding 1 lives, and E5 is green. This is the rule the PR quotes at itself: an
oracle has to keep every side of the object non-degenerate. One extra subject (or dataset) with
`t_inf` a real fraction of `II` and `lag > II − t_inf` closes it.

**The closed-form engine has no ungated test.** All five pointwise `PRED` tests use `[odes]` models;
`dose_form_lag_oral_cf_fit.ferx` appears only in the `slow-tests`-gated engine-agreement test. So
the 53-line `pk/event_driven.rs` change — a *different* defect (the post-hoc overlay) with a
*different* fix — is validated on a PR by nothing but the pre-existing flat-covariate unit test.
The PR's own argument is that the EBE-free `PRED` oracle is cheap and ungated, and it applies here
verbatim:

```rust
assert_pred_matches_nonmem(
    "dose_form_lag_oral_cf_fit.ferx",
    "dose_form_lag_ss.csv",
    "results/dose_form_lag_ss.tab",
    1e-4,
);
```

That single line would have caught finding 1's closed-form half.

**No TV-covariate `Dual2`-vs-FD test for the twin's SS-infusion arm.**
`ode_provider_ss_lagtime_tvcov_matches_production` is an oral bolus, so the `−δlag` scoping change
at `sens/ode_provider.rs:5214` is guarded only by the flat-covariate
`ode_provider_ss_lagtime_infusion_matches_production`. The reviewer hints call that line the subtle
one and note the two terms cancel under flat covariates — which is the argument that a flat test
pins the cancellation, not the TV-covariate behaviour. An infusion copy of the new fixture with
`dose_covariates`/`obs_covariates` variation (so `check_vs_production`'s FD-of-production on
`df_deta`/`df_dtheta` is live) is a few hundred milliseconds.

**Nothing covers a reset inside the pre-arrival window.** With the arrival-side equilibration gone,
an `EVID=3/4` between the record and the arrival now zeroes the seeded trough and the arrival applies
only `F·AMT`; before, the arrival re-equilibrated. That may well be the NONMEM-correct reading — the
SS load happens at the record, so a later reset should wipe it — but it is a silent, undocumented
behaviour change, and the E4 control carries a lagged dose with no `SS=1`.

---

### Smaller

* **`sens/ode_provider.rs:5409` — `g_minus` for a seeded SS dose is evaluated at `TAD = 0`.**
  Removing the `is_ss_dose` skip makes `boundary_velocity` run, but `pre_anchor` falls back to
  `t_event` whenever `last_dose_eff` is `NEG_INFINITY` — which is the case whenever the SS dose is
  the subject's first, since the filter is `d.time + lag <= cur_t`. For an SS dose the anchor is
  *determined*: the prior pulse at `d.time + lag − II`, i.e. `TAD = II` at the arrival. Open question 1
  describes this as unobservable ("the anchor models' RHS do not read TAD"), but `ode_tvcov_supported`
  has no TAD gate, so a `TAD`-referencing `[odes]` model reaches this line and gets a wrong `∂f/∂lag`
  into FOCEI. The SS-aware `cur_t - elapsed.rem_euclid(d.ii)` branch two lines below
  (`ode/predictions.rs:4730`) already computes the right number — only the `<= cur_t` filter keeps it
  out of reach. Worth doing here rather than deferring, or gating loudly.

* **`sens/ode_provider.rs:4483` — `n_ss_seeds` still uses the pre-#1121 predicate**
  (`is_ss_dose(d) && lag_val(*k) > 0.0`), with a comment claiming it matches production's reservation.
  Harmless today (capacity hint only), but it is the one surviving copy of the old condition, in the
  file the PR calls the subtle one, and it contradicts the design statement that every site reads the
  shared predicate.

* **`ss_seeded_at_record` should live in `dosing.rs`.** CLAUDE.md names that module "SS-equilibration
  policy … the single home shared by pk/ode/sens", and this predicate decides *where* the equilibration
  happens for three cross-module consumers. As written, the analytic PK engine takes a dependency on
  `crate::ode::predictions` for a rule that has nothing to do with ODEs.

* **`sens/provider_tests.rs:9455` — the IOV non-degeneracy guard uses `theta[2]`**, while the model
  resolves `ALAG1 = TVLAG * exp(ETA_LAG [+ κ])`. The doc comment calls those two samples "the direct
  evidence that the IOV gap is closed rather than assumed to be"; if a later edit to `theta`/`stacked`
  drops the resolved lag below `0.3`, the test silently reverts to the post-arrival-only version it
  replaced and the guard still passes. Compute it from the same expression the walk uses, per occasion.

---

### What holds up

The record-vs-arrival convention, the E3/E3P/E3B/E5 geometry choices, and the honesty about the three
wrong claims in the issue body are all good. The EBE-free `PRED` oracle is the right idea and the
`compared == ferx.len()` guard against a silent zero-comparison pass is exactly the paranoia that
belongs in a helper like that. `ss_state_at_phase_matches_nonmem_at_the_dose_record` pinning the depot
leg in closed form rather than with a loose `> 0` is the right instinct too. The doc comment on
`ode_provider_ss_lagtime_tvcov_matches_production` telling the reader it is a coupling guard and not an
oracle is worth keeping as a pattern.

The two blocking items are both the same shape as what the PR set out to fix: a case where the
seeded path and the equilibrated path are *not* the same number, and no anchor stands on that side.

TeunP:
## Review round 2 — both blocking items fixed, plus a third defect they led to

Everything below was reproduced first and measured after. Two new NONMEM 7.6.0 runs
now anchor the cells the review named; the semantics they settle turned out to
contradict one of my own hypotheses, so this is measurement rather than argument.

### What NONMEM actually does (71 `PRED` rows, agreement to table precision)

| case | NONMEM 7.6.0 |
|---|---|
| `0 < ALAG < II` | seed at phase `II − ALAG`; the #1073 record rule applies inside the window |
| `ALAG > II − T_inf` (infusion) | the previous cycle's infusion **keeps running past the record**, stopping at `record + (T_inf − phase)` |
| `ALAG ≥ II` (bolus) | phase **clamped to 0**, not wrapped — the record carries the SS *peak* |
| `ALAG ≥ II` (infusion) | delivers `T_inf + (ALAG − II)` hours of drug — **not mass-balanced** |

The clamp is worth dwelling on because I had guessed wrong: wrapping the phase into
`[0, II)` is the natural-looking reading and gives `PRED(0) = 2.1815`, where NONMEM
gives `13.197 = (D/V)/(1 − e^{−k·II})`, the post-pulse peak. Clamping is also the only
choice that is **continuous in `ALAG`**, which matters directly for the review's point
about discontinuity: `ALAG` is routinely estimated, and wrapping would step every
pre-arrival prediction by `e^{−k·II}` the moment the optimiser walked it across `II`.

The last row is measured, not assumed: reconstructing the implied rate from NONMEM's
own `PRED` trajectory gives exactly `100.00` for 8 h on an `AMT = 600`, `RATE = 100`
dose — 800 mg. ferx does not reproduce that; `docs/model-file/lagtime.qmd` states the
divergence, and the four ferx routes are asserted to agree with each other there.

### 1 (blocking) — SS infusion residual: fixed

Reproduced at `4.05e-2` before the fix, with the static and event-driven routes
disagreeing on the same model. The previous cycle's window belongs to no `DoseEvent`,
so it is now carried explicitly: `crate::dosing::ss_residual_infusion_end` supplies the
edge, and each engine breaks and gates on it the same way it does its own infusion end.
The twin needed a real rate-off boundary for it too (it moves with `lag` and with the
window length exactly as the real end does), not just a break.

### 2 (blocking) — `ALAG ≥ II`: fixed

Now seeded at the clamped phase. `ss_seeded_at_record` drops its `lag < ii` bound and
`ss_seed_phase` clamps; the boundary test that pinned the old bound is replaced by one
that pins the phase arithmetic, and four new anchors pin the resulting predictions.

### 3 — the same cell, post-arrival, found while fixing 2

Not in the review, and pre-existing: re-equilibrating at the arrival is a shortcut for
propagating the seed there, and it is exact only while the flowed state *is* the trough
— i.e. `lag ≤ II`. Past that the two static predictors read **+4.3 %** high *after* the
arrival. Gated on `ss_arrival_is_trough`; `pk::predict_concentration` additionally
splits its collapsed `C_ss(t − t_eff)` into tail + one real arrival for that case (an
algebraic identity for `lag ≤ II`, so the ordinary path stays bit-identical).

### Result

Four routes × two geometries, all against the same NONMEM tables:

```text
                              ALAG > II - T_inf      ALAG >= II
  [odes]  event-driven            2.5e-5              1.9e-5
  closed form  event-driven       2.5e-5              1.9e-5
  [odes]  static                  2.9e-5              2.9e-5   (was 4.3e-2)
  analytical superposition        2.9e-5              2.9e-5   (was 4.3e-2)
```

`2e-5` is NONMEM's `$TABLE` print precision — these are exact. The two ferx engines
also agree with each other to `1e-10` on every dataset, including the unanchorable
`ALAG ≥ II` × infusion cell.

### Mutation-tested, because a green test proves nothing on its own

Nine new ungated tests; each mutation kills a targeted subset, and none kills all nine:

```text
  wrap the seed phase instead of clamping   -> the four ss_lag_ge_ii tests
  drop the residual infusion window         -> the three walking engines
  re-equilibrate at every arrival           -> the two static ss_lag_ge_ii tests
  drop the seed phase in the superposition  -> the two superposition tests
```

That the superposition survives the residual-window mutation is not a gap: it does not
walk, so it needs only the phase — recorded in the test module docs so a future reader
does not "fix" it.

### The rest of the review

* **E5 degenerate** — closed. The new infusion geometry puts `T_inf` at half the
  interval, so the `phase ≤ T_inf` arm is entered by two of three subjects. E5's doc now
  says what it cannot reach and where that lives.
* **Closed-form engine had no ungated test** — added, exactly as suggested, plus the
  pre-arrival variant.
* **No TV-covariate `Dual2`-vs-FD for the SS-infusion arm** — added, sized so
  `II − ALAG < T_inf` so the two `is_ss_dose`-scoped terms do not cancel.
* **Nothing covered a reset in the pre-arrival window** — added (see below).
* **`n_ss_seeds` used the old predicate** — fixed.
* **`ss_seeded_at_record` should live in `dosing.rs`** — moved, with `ss_seed_phase`,
  `ss_arrival_is_trough` and `ss_residual_infusion_end`. `dosing` is a
  leaf module, so `is_real_infusion` moved down with them rather than have `dosing` call
  up into `ode::predictions`.
* **IOV guard used `theta[2]`** — now resolved through the model's own parameter
  function per occasion, the way `predict_iov` does.
* **`g_minus` evaluated at `TAD = 0`** — investigated, **not fixed here**, and the
  reason is worth more than the fix would have been. I wrote the pre-arrival anchor
  (a seeded SS dose contributing from its record, at the pulse that produced the
  state there) and then backed it out, because two things make it not a local
  change: the dense predictor returns `NaN` where the walk falls back to the first
  arrival, so the two routes do not agree on the base case; and the twin keeps
  `last_dose_eff` as a running `max` over arrivals while production re-folds per
  segment — so the twin never wraps an SS anchor per cycle, and it cannot absorb a
  pre-arrival anchor at all, since that anchor is *earlier* than the first arrival
  and a max discards it. Fixing production alone would have widened a divergence
  instead of closing one. All three findings go to #1126 with the reproduction.

### Two things I found and did **not** fix here

Both are pre-existing and neither is caused by this issue. Filed as #1129 and as an
addendum on #1126.

* **A reset before a lagged infusion's arrival corrupts `∂f/∂η_LAG`.** Inserting an
  `EVID=3` reset that is a physical *no-op* (the state is already zero) flips the twin's
  gradient from `+7.14` to `−12.53` against FD, with flat covariates and a non-SS dose.
  The value is unaffected, so it is invisible in predictions and reaches FOCEI silently.
  The new reset test is deliberately a **bolus**, with a comment pointing at this, so it
  measures the #1121 behaviour instead of inheriting an unrelated failure.
* **`TAD` is `NaN` before a subject's first arrival**, and because it multiplies into
  the state the `NaN` reaches every later prediction for that subject. Not SS-specific —
  a plain lagged dose does the same — so it is the concrete failure mode of #1110's open
  question, and an addendum to #1126 rather than a new issue. Nothing in this PR claims
  to fix `TAD`, and there is deliberately no `TAD` test here: I could not verify one
  end-to-end, and an unverifiable behaviour change is exactly what this PR is arguing
  against.

One correction worth stating, since it nearly cost a real finding: I first read that
`NaN` as an artifact of my own probe (a `TAD` *integrator* has no periodic steady
state, so the SS solve is singular) and retracted it. Re-measured with `TAD`
modulating elimination instead — still `NaN`. The diagnosis of the degenerate model
was right; the retraction was not. A retraction is an assertion and carries the same
burden of proof as the claim it withdraws.

TeunP:
## Rebased onto `main` (#1123 landed), and one claim in my review response above was wrong

### Merge

Base retargeted from `worktree-fix-1073-lagged-dose-record` to `main`. Two textual
conflicts, both pure append-vs-append at a file tail, both sides kept:

* `src/dosing.rs` — #1073's `governing_record_tests` vs #1121's SS-seed vocabulary
  (plus the `is_real_infusion` that moved down out of `ode::predictions`).
* `src/sens/ode_provider_tests.rs` — #1070's TAD-gate section vs #1121's SS-infusion
  and reset twin tests. A doc comment describing an **oral** fixture that I had left
  stranded on the IV-infusion const back in `2e14f1f1` is dropped here — my own
  defect, caught reading the conflict rather than by any test.

The three walk files auto-merged, which on its own proves nothing: #1073 and #1121
rewrote the same event-driven walk. The interaction points were read rather than
assumed.

* #1073's shared `governing_record_indices` resolves every timeline slot to the
  record that *terminates* its segment. #1121 adds two **non-record** slots —
  production's residual-infusion `Kind::InfusionEnd` and the twin's `K_SS_INF_END` —
  and neither is admitted by `is_record` or by the twin's
  `K_DOSE_REC | K_SS_SEED | K_PKONLY | K_OBS` set, so both take the next record
  ahead, exactly as an ordinary infusion end does.
* `K_SS_SEED` **is** in the twin's record set and shares `pk_at_dose[idx]` with its
  dose's own `K_DOSE_REC` at the same instant, which is what the seed needs.
* `is_real_infusion` exists once, in `dosing.rs`; #1073 did not reintroduce it.

Measured after the merge, not assumed:

```text
  cargo test --lib                     3706 passed   0 failed
  dose_form_lag_nonmem_anchor            17 passed   0 failed
  ss_lagtime_edge_nonmem_anchor          11 passed   0 failed
  tad_lag_nonmem_anchor        (#1070)    2 passed   0 failed
  tvcov_lag_saltation_nonmem_anchor (#1073) 4 passed 0 failed
```

The last two are #1073's and #1070's own anchors, run deliberately: a regression I
caused in *their* territory should show up rather than be assumed away.

### The stale claim

Under "Two things I found and did **not** fix here" I wrote that `TAD` is `NaN`
before a subject's first arrival and that this is "not SS-specific — a plain lagged
dose does the same". That was true when I measured it, on #1123's branch before its
last two commits. **#1073 then fixed it**, and correctly: `tad_anchor` falls back to
the subject's earliest lagged arrival, and the twin seeds `last_dose_eff` the same
way. Re-measured on the merged tree:

```text
  plain lagged dose (ALAG1 = 3)   0  0  0  57.038106  43.460156  22.973573
```

All finite. So that bullet is withdrawn as written.

What survives is **steady-state-specific and has nothing to do with the lagtime**,
which is a different — and worse — finding than the one I filed:

```text
  SS=1, II=12, ALAG1=3    NaN NaN NaN NaN NaN NaN
  SS=1, II=12, ALAG1=0    NaN NaN NaN NaN NaN NaN   <- no lagtime, no pre-arrival window
```

`equilibrate_ss_state` hands the solver the bare `pk_params_flat`, so the RHS's
`params.get(MAX_PK_PARAMS + 1)` returns `None` and injects `NaN` into the
equilibration itself; the seed comes back `NaN` and every prediction for the subject
inherits it, post-arrival rows included. Full reproduction and mechanism on #1126.

Nothing in this PR claims to fix `TAD`, and there is still deliberately no `TAD`
test here — but the reason I gave for that has changed, so it is worth restating
accurately rather than leaving a wrong one on the record.

Issue #344, opened 2026-06-14
CI gap: ci-test profile disables debug-assertions, hiding debug_assert! guards from CI


TeunP:
## Summary
The CI test/coverage jobs build with `--profile ci-test`, which `inherits = "release"` (`Cargo.toml`) and therefore runs with **debug-assertions OFF**. As a result every `debug_assert!` in the codebase is a no-op in CI — including invariant guards meant to catch real bugs.

## How this surfaced
A pre-existing bug in `compute_max_stack` (`src/parser/model_parser.rs`) tripped its own `debug_assert!` for any expression containing a conditional (`if`/ternary), but **CI stayed green** because `ci-test` silences the assertion. It only failed under the default dev/test profile. The bug itself is fixed in #342, but the *class* of problem remains: any future off-by-one or invariant violation guarded by `debug_assert!` is invisible to CI.

Concretely, before #342 these four lib tests panicked under the dev profile yet passed in CI:
- `parser::model_parser::tests::bytecode_matches_ast_on_conditional_and_logic`
- `parser::model_parser::tests::test_inline_if_in_parameter_assignment`
- `parser::model_parser::tests::test_ode_rhs_defined_names_ok`
- `parser::model_parser::tests::test_parse_all_example_ferx_files`

Repro (dev profile): `cargo test --lib --no-default-features --features ci test_ode_rhs_defined_names_ok`.

## Proposed fix
**Add** a CI job that runs the suite with **debug-assertions enabled**, so these guards are exercised — keep the fast `ci-test` job as-is for speed. Options:
- A dedicated job running on the default `dev` profile (debug-assertions on), e.g. `cargo test --lib --no-default-features --features ci`; or
- A `ci-test`-like profile with `debug-assertions = true` used by a single job.

The `ci-test` profile exists for speed (release-level opt, no fat LTO — see the `Cargo.toml` comment), so don't flip it; add a separate job.

## Caveat / scoping
Turning debug-assertions on in CI may surface **other** latent `debug_assert!` failures elsewhere (this is the first we know of). Expect to triage those as part of enabling the job — which is exactly why this was kept out of #342 and filed separately.

Found while reviewing the absorption work (#322 / #343).

TeunP:
**Second consequence, measured on #1136: the disabled `debug-assertions` also make those lines permanently uncoverable, so they count against the `codecov/patch` gate.**

This issue frames the cost as "guards are hidden from CI" — they never execute, so a violated
[...]

PR #1130, opened 2026-08-29
fix(pk): route a model-time-reading [odes] RHS to the event-driven predictor [closes #1124]

### LINKED ISSUE #1110: bug(ode): an [odes] RHS reading TAD together with a lagtime returns NaN predictions — and what TAD means before the first arrival is undecided

### LINKED ISSUE #1124: bug(pk): the closed-form modified-release fast path silently drops a TAD-dependent RHS term — predictions 53% wrong
## Summary

A `[odes]` RHS that reads `TAD` is served by the **closed-form modified-release fast
path**, which never evaluates that RHS. The `TAD` term is silently dropped from the
**predictions themselves** — not approximated, absent. No error, no warning, no
`converged: false`.

Measured on a 1-cpt mixed `first_order` + `zero_order` model, the production predictor
returns numbers **bit-identical to the same model with the `TAD` term deleted**:

| t | production (`compute_predictions_with_tv`) | ODE twin (`ode_predictions`) | rel. error |
|---|---|---|---|
| 0.5 | 0.204264 | 0.203676 | 4.9e-4 |
| 1.3 | 0.956061 | 0.943412 | 6.5e-3 |
| 2.7 | 1.707419 | 1.560315 | 5.7e-2 |
| 3.6 | 1.606392 | 1.337585 | 1.2e-1 |
| 4.4 | 1.480698 | 1.106013 | 1.8e-1 |
| 8.2 | 0.957530 | 0.311414 | 4.9e-1 |
| 12.0 | 0.615780 | 0.052563 | **5.4e-1** |

At `t = 12` the reported concentration is **11.7× the true value**. The same model with
the `TAD` factor removed agrees with its ODE twin to **1.3e-9** at every time, so the
fixture is sound and the divergence is the `TAD` edit alone.

This is a **value** defect. It reaches `IPRED`, the OFV, residuals, and every downstream
diagnostic.

## Reproducer

```
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV * exp(ETA_V)
  FZO = TVFZO
  FZO1 = 1 - TVFZO
  KA = TVKA
  DUR = TVDUR
  LAG = TVLAG
[structural_model]
  ode(states=[central])
[odes]
  d/dt(central) = FZO1*first_order(ka=KA, lag=LAG) + FZO*zero_order(dur=DUR)
                  - CL/V*central*(1.0 + 0.3*TAD)
[scaling]
  y = central / V
```

Static subject, single 100 mg oral dose. **No lagtime and no IIV on a lag are required** —
this is independent of #1070.

## Root cause

`pk/modified_release.rs:111`:

```rust
fn probe_rhs_f64(spec: &OdeSpec, u: &[f64], p: &[f64], t: f64) -> Option<Vec<f64>> {
    ...
    // `tafd`/`tad` are constants w.r.t. the parameters; a supported (static)
    // disposition is autonomous so their value is irrelevant to the Jacobian.
    prog.eval_rhs_g::<f64>(u, p, t, t, 0.0, &mut du, &mut vars, &mut stack);
```

`identify_disposition` establishes time-invariance **behaviourally**, by probing the RHS
at `t` and `t + 1.0` and requiring the columns to match
(`modified_release.rs:180`). That probe passes `t` for `tafd` — so a `TAFD`-reading RHS
**is** caught — but hard-codes **`tad = 0.0`**. A `TAD`-reading RHS therefore evaluates
identically at both probe times and is admitted as a linear time-invariant system.

The comment's premise ("a supported (static) disposition is autonomous") is the
conclusion the probe is supposed to establish, so a `TAD` term makes it circular.

`-CL/V*central*(1 + 0.3*TAD)` is still *linear in the state*, so the linearity checks
pass too. The identified disposition is a plain 1-cpt with rate `CL/V`, and the `TAD`
factor is gone.

## Why nothing caught it

`pk/mod.rs:2281` puts `mr_predictions` **before** the `ode_predictions` twin, so
production and the MR fast path are the same code — comparing them agrees by
construction. The invariant the branch comment claims ("the admitted result reduces to
that twin to solver tolerance, `modified_release::tests::reduces_to_ode_*`") is exactly
what is violated, but no `reduces_to_ode_*` fixture reads `TAD`.

The guard `if has_tv || subject.has_resets() || uses_time` in front of that branch
excludes `TIME`-dependent models; `uses_time` is `model_uses_time_builtin`, which
inspects the individual-parameter program and does not see `TAD` in an RHS.

## Also bypasses #1070's FD gate

The same funnel is tried before the gated ODE provider on the sensitivity side
(`sens/provider.rs:3982`, `:5262`). Measured on the model above **plus** a compartment
`LAGTIME`:

```
ode_analytical_supported = false   (the #1070 gate declines, correctly)
mr_subject_eta_grad      = Some    (MR serves it anyway)
mr_subject_sensitivities = Some
```

So PR #1122's gate is bypassed for MR-eligible models. That is subsumed by this bug
rather than additive — such models are already wrong by the larger margin above — and the
two fixes compose in either order: once MR declines a `TAD`-reading RHS, the ODE provider
takes over and #1122's gate applies normally.

## Proposed fix

Decline in `mr_scope` when the RHS program reads `TAD`, **unconditionally on lagtime**
(row 3 above has no lagtime and is already 53% wrong). `parser::model_parser` already
exposes the predicate added for #1070 — `uses_dose_anchored_time_vars()`.

Cheaper alternative, and arguably the more honest one: pass a **varying** `tad` in
`probe_rhs_f64` (e.g. `t` and `t + 1.0` alongside `tafd`) so the existing
time-invariance check catches it the same way it already catches `TAFD`. That fixes the
class rather than the one spelling, and needs no new predicate.

Either way the model then routes to the ODE twin, which integrates the RHS as written.

## Validation

The ODE twin is the reference: the factor is in the model the user wrote, so unlike
#1110 there is no convention question here. A NONMEM anchor is expressible via
`MTIME`/`MTDIFF` (see `nonmem_anchor/tad_lag_*` and the three authoring traps recorded in
those control-stream headers).

Regression test must assert `reduces_to_ode_*` for a `TAD`-reading RHS — the gap that let
this through.

TeunP:
## Why

An `[odes]` right-hand side that reads `TAD` was served by the **closed-form modified-release
fast path**, which never evaluates that right-hand side. The `TAD` term was not approximated —
it was **absent from the predictions**. No error, no warning, `converged: true`.

Measured on the reproducer in #1124 (mixed `first_order` + `zero_order`, one central
compartment), production output was **bit-identical** to the same model with the term deleted,
at every observation:

| t | production | production, `TAD` factor deleted | ODE twin | rel |
|---|---|---|---|---|
| 0.5 | 0.097541 | 0.097541 | 0.097299 | 2.2e-4 |
| 4.4 | 1.509259 | 1.509259 | 1.204799 | 1.4e-1 |
| 12.0 | 0.718438 | 0.718438 | 0.089271 | **5.8e-1** (8.05×) |

This is a **value** defect: it reaches `IPRED`, the OFV, residuals, and every downstream
diagnostic.

Two things kept it invisible. `identify_disposition` establishes time-invariance
*behaviourally*, by probing the RHS at two times — but `probe_rhs_f64` pins `tad = 0.0`, so a
`TAD`-reading RHS evaluates identically at both probes and is admitted as time-invariant. And
the routing site tries the closed form **before** its ODE twin, so production and the closed
form are the same code there: `mr_vs_prod` was exactly `0.0`.

## What changed

Three changes, all load-bearing:

1. **`pk/mod.rs` — the routing guard (the actual fix).** Both production value entry points
   computed `uses_time` from `model_uses_time_builtin`, which inspects the individual-parameter
   program and **never the `[odes]` RHS**. `model_uses_time_anywhere` also asks the RHS, so
   these models take the **event-driven** predictor. Applied at
   `compute_predictions_with_tv_into_with_schedule` and `compute_predictions_with_states`.
2. **`mr_scope` declines the same predicate.** Required separately: `sens/provider.rs` reaches
   `mr_subject_sensitivities` / `mr_subject_eta_grad` without passing through `pk/mod.rs`
   routing at all. Without this the **value** would come from the ODE walk (with `TAD`) and the
   **gradient** from the closed form (without it) — worse than either alone.
3. **`OdeRhsProgram::reads_time_builtin()`** — a new `Op::PushTime` / `Expression::Time` walk.
   `stmts_read_slots` (hence `uses_time_vars`) structurally cannot see a bare `TIME`, which
   resolves through the model-time thread-local rather than a variable slot.

### The defect is wider than `TAD`

The two-time probe also cannot see anything in an **untaken `if` branch**. With the threshold
inside the observation window, all four spellings were wrong by **4.0e-1**:

| RHS | before | after |
|---|---|---|
| bare `TAD` | admitted (wrong) | declined |
| bare `TAFD` / `T` / `TIME`, `TAFD − TAD` | declined | declined |
| `if (TAD>6)` / `if (TAFD>6)` / `if (T>6)` / `if (TIME>6)` | **all admitted (wrong)** | all declined |

This is why the fix is not keyed on `uses_dose_anchored_time_vars()` (`TAD` only), which would
have left three of those four open.

### Also fixes #1110

The **dense** predictor anchors `TAD` on doses that have already arrived; under a lagtime none
has at the first segment's start, so it returned `NaN` for every observation. #1110's exact
reproducer goes `[NaN, NaN, NaN, NaN]` → `[4.561, 3.829, 2.254, 0.0245]`, and there is a
regression test for it.

Scope of that claim, precisely: the NaN is gone from the **two ordinary prediction entry
points**, which is where #1110 reproduced. The `tad_anchor` fold-to-`NEG_INFINITY` mechanism
itself is untouched, and `ode_predictions` is still called directly on the **SDE/EKF variance
path** (`stats/likelihood.rs:849`). I have not audited whether a `[diffusion]` model with a
`TAD`-reading RHS and a lagtime can still reach it there — that is worth checking, but it is
outside this PR, which does not touch SDE.

I have **not** closed #1110: that residual path plus its second half (what `TAD` should mean
*before* the first arrival, a convention question this does not settle) are both still open.

### The "runtime verify-gate" now exists

`modified_release.rs` cited a "runtime verify-gate" as the real safety net in three separate
comments, and `grep` returned only those three comments. Nothing implemented it. That is the
architectural reason #1124 shipped, above `probe_rhs_f64`.

`verify_against_ode_twin` re-validates every admitted subject against its integrated twin under
`#[cfg(debug_assertions)]`. It **asserts, never declines** — a gate that changed the route in
debug would make debug and release disagree about what ferx predicts. Mutation-tested: it fires
on #1124 at the *first* observation.

## Alternatives considered

**Varying `tad` in `probe_rhs_f64`** (the issue's "cheaper alternative, and arguably the more
honest one"). Measured — it is actively harmful. Passing `t` for both `tafd` and `tad` lets
`TAFD − TAD` cancel to a constant, flipping a correctly-**declined** RHS to wrongly
**admitted**, and branch-condition `TAD` still slips past. It trades one bug for another and
fixes neither branch case.

**Declining in `mr_scope` alone** (my first plan). Routes to the *dense* predictor, i.e. into
#1110: a silent 8× error becomes `NaN`. That is why the fix is at the routing guard.

**Giving `compute_predictions_with_states` an MR branch** to make the two entry points
consistent. This would destroy the only independent oracle in the system — pre-fix the two
disagreed by 1.9× on the same subject and nothing compared them. The asymmetry is deliberate;
there is now a test asserting it.

## Cross-repo dependency

No public API change (`model_uses_time_anywhere` and `reads_time_builtin` are `pub(crate)`).

## Breaking changes
- [x] None

Predictions **change** for models whose `[odes]` RHS reads model time — they were wrong before.
Models that do not read model time are bit-identical.

## Numerical validation
- [x] Compared against: **NONMEM 7.6.0** (`nonmem_anchor/mr_tad.ctl`, new)
- [x] Reference values:

```
# NONMEM (ADVAN13 TOL=9, $EST MAXEVAL=0 POSTHOC INTERACTION)
#OBJV = -234.478425749153

# ferx, before this PR (closed-form fast path)
OFV   = -12.4832        delta 222.0

# ferx, after
OFV   = -234.478426     delta 2.5e-7
```

The new anchor is oral 1-cpt with a single `first_order` forcing and `TAD` on elimination —
verified to be a shape `mr_scope` **did** admit, so it exercises the defective path. **Two
doses**: under a single dose `TAD == TAFD` and the anchor could not tell a `TAD` defect from a
`TAFD` one. The second dose lands on residual drug (22.90 against an eventual peak of 91.21),
so the incoming side of the event is live rather than cancelling against `g(x⁻) = 0`.

The `$DES` construction is **checked, not assumed** — authoring `TAD` in `$DES` is where the
`tad_lag` anchors went wrong twice. `mr_tad_det.ctl` (`$OMEGA 0 FIX`) against
`mr_tad_deterministic_truth.py` (independent stdlib RK4, no NONMEM, no ferx) agrees to
**1.7e-5**, which is `$TABLE` print precision.

## Performance

Affected models integrate instead of evaluating in closed form — but that set is exactly the
set that was wrong, plus the contrived case where a time variable provably cancels (`T − TAFD`
is the first dose time). Release builds are unchanged otherwise; the verify-gate compiles out.

Debug builds pay one twin integration per distinct (model, subject). The property asserted is
structural, not parameter-point-dependent, so once per pair is the right granularity.

Measured on `cargo test --lib --features ci`: **577 s unmemoized**, against 281 s / 470 s /
381 s memoized over three runs and a single 296 s run with no gate at all. Wall-clock on this
machine is too noisy to separate the memoized gate from no gate at all — I am not claiming it is
free. What the numbers do support is the memo itself: the unmemoized ~2× penalty sits well
outside that spread.

## Tests

Mutation-tested with both gates disabled — **5 of the new tests fail**:
`tad_reading_rhs_reduces_to_its_ode_twin`, `the_tad_term_is_live_in_the_predictions`,
`mr_scope_declines_every_model_time_spelling`, `mr_declines_what_the_ode_analytic_gate_declines`,
`a_tad_rhs_under_a_lagtime_predicts_finite_values`.

Also added:
- `mr_matches_the_states_entry_point` — the cross-entry-point oracle.
- `the_verify_gate_rejects_a_closed_form_that_misses_its_twin` / `..._accepts_...` /
  `..._stays_quiet_at_default_solver_tolerances`.
- `rerouting_a_bare_tafd_rhs_does_not_move_the_predictions` — the re-route is inert (6e-13) for
  models that were already correct.
- `tad_reading_mr_shaped_model_takes_a_route_that_matches_fd` — `Dual2`-vs-FD parity, since this
  reroutes a gradient path.
- `tests/mr_tad_nonmem_anchor.rs` — slow-gated, with a non-degeneracy assert that deleting the
  `TAD` factor moves the objective by >100 units.

`cargo test --lib --features ci`: **3690 passed, 0 failed**. `cargo test --tests --features
ci,survival`: all binaries green.

## Docs
- [x] `docs/` updated — new "A right-hand side that reads model time is always integrated"
      section in `docs/model-file/ode-models.qmd`, with a warning callout telling users with
      affected results from ≤0.3.0 to re-run; `docs/estimation/foce.qmd` cross-references #1124
      from the #1070 paragraph.
- [x] No new pages, so `docs/_quarto.yml` is unchanged.
- [x] `CHANGELOG.md` `[Unreleased]` entries added for #1124 and #1110.

## Checklist
- [x] `cargo clippy` clean (as CI runs it: `--no-default-features --features ci,markov,nn`)
- [x] `Dual2` path stays differentiable — no kernel touched; the change is routing only, and the
      rerouted gradient is FD-parity tested

## Reviewer hints

The subtle parts, in order:

1. **Why both gates.** Dropping either one leaves a value/gradient split. `mr_scope` alone also
   routes into #1110's `NaN`.
2. **The verify-gate's tolerance.** It scales as `max(1e-4, 500 × reltol)`, not a constant. A
   fixed `1e-4` false-positived on an ordinary multi-dose oral model — the zoo fixtures all pin
   `reltol = 1e-10`, so no unit test covered the default-tolerance regime; the NONMEM anchor
   caught it. Measured worst legitimate twin disagreement at defaults is 6.4e-3 (64× `reltol`,
   growing with dose count), so this leaves ~8× headroom against false positives and ~12×
   detection margin. A false positive panics a debug build, so the headroom is deliberately on
   that side.
3. **`compute_predictions_with_states` must keep having no MR branch.** It is the independent
   side of the cross-entry-point oracle.

Skimmable: the comment rewrites in `probe_rhs_f64` / `LINEARITY_TOL` / `identify_disposition`
(the old text asserted a safety net that did not exist), and the parser walk, which mirrors
`stmts_read_slots` line for line.

## Open questions

- #1110 is fixed in symptom but I left it open for its convention half. Close it, or split that
  into a fresh issue?
- The `WT` FIXED-zero-exponent covariate in `tad_lag_{A,B}` exists purely as a #1110 workaround
  and is now obsolete. I left those anchors untouched so this PR does not perturb a passing
  anchor — worth a follow-up to drop it?


TeunP:
## Review — xhigh, adversarial

Read the full diff at `456deaad` plus the enclosing functions on both sides of every routing site (`pk/mod.rs`, `pk/modified_release.rs`, `sens/ode_provider.rs`, `sens/provider.rs`, `estimation/inner_optimizer.rs`, `api/output_columns.rs`, `parser/model_parser.rs`), and re-ran the anchor's independent reference locally.

**The core fix is right.** Deciding model-time dependence *statically* rather than through `identify_disposition`'s two-time probe is the correct altitude — the "vary `tad` in `probe_rhs_f64`" alternative really is worse (`TAFD − TAD` cancels, and a branch condition is unreachable at two sample times either way), and the reasoning for gating at the routing site rather than only in `mr_scope` (otherwise you land in #1110's `NaN`) holds. What follows is what survived a hostile read.

### On the anchor — it holds up

I re-ran `mr_tad_deterministic_truth.py` against the committed `results/mr_tad_det.tab`: worst relative deviation **1.749e-05**, at `t = 11.50`. `$TABLE` prints 5 significant figures (`1.6245`, `2.6623`, `1.1448`, …), so print precision at `IPRED = 1.1448` is ±5e-5 absolute → 2.3e-5 in the script's `|a−b|/(1+|b|)` metric. **1.75e-5 sits under that bound**, so the "print precision, not either integrator" attribution is correct and the `$DES` construction is genuinely settled. (Nit: the script's own comment says "6 significant figures"; the table prints 5.)

I also probed the one thing that could have faked that number: `tad(t)` uses `d <= t + 1e-12`, so the RK4 `k4` stage at a segment's right endpoint can pick up the *post*-dose anchor. Patched it to a per-segment anchor and re-ran — the trajectory moves by **1.7e-13**. Not an issue; recording it so nobody has to re-derive it.

`#OBJV = -234.478425749153` in `results/mr_tad.ext` matches `NONMEM_OFV` exactly, no warnings beyond `NM-TRAN INFERS ... POPULATION`, `$OMEGA`/`$SIGMA` correctly mirror ferx's variance/sd conventions, and the two-dose + residual-drug design satisfies the non-degeneracy rule (`g(x⁻) ≠ 0` on the second dose). The one-state-ferx / DEPOT-NONMEM equivalence argument is sound. **The anchor is the strongest artifact in the PR.**

---

### Blocking-ish

**1. `sens/ode_provider.rs:809` — SS + a bare `TIME` RHS still escapes the non-autonomous gate.**

```rust
if has_ss && ode.rhs_program.as_ref().is_some_and(|p| p.uses_time_vars()) {
```

This PR's own new doc on `reads_time_builtin` says: *"Pair it with `uses_time_vars`; neither alone is sufficient."* This gate is exactly such a caller and was not paired. `d/dt(central) = -(CL/V)*central*(1 + 0.3*TIME)` with an `SS=1` dose and no `TIME` in `[individual_parameters]`: `compiled_model_uses_time_builtin` is false, `ode_subject_supported` passes, `ode_tvcov_supported` admits on `has_ss`, `uses_time_vars()` is false for bare `TIME` — so the subject reaches `equilibrate_ss_state_g`, which that gate's own comment says gives *"a wrong prediction **and** gradient"*. Same omission at `ode_provider.rs:2583` (IOV) and its attribution mirror `inner_optimizer.rs:274`. One clause each.

**2. `pk/mod.rs:47` — the widened predicate was applied to the value path only.**

The sites that make the *same* dense-vs-event-driven decision downstream still ask `compiled_model_uses_time_builtin`:

- `ode_tvcov_supported` (`ode_provider.rs:655`) — picks static vs event-driven **dual** walk;
- `api/output_columns.rs:602` — picks the `[derived]` grid-state engine.

For `d/dt(central) = -(CL/V)*central*(1 + 0.3*TAD)`, bolus doses, no lag / TV-cov / SS / modeled dose: the value now comes from `ode_predictions_event_driven`, the jet from the **static** dual walk, and `[derived]` states from the dense solve. That is the value/gradient split the PR body says `mr_scope` was widened to prevent, reintroduced one layer down. The new FD-parity test cannot see it — `MR_MIXED_1CPT_TAD` carries `first_order(ka=KA, lag=LAG)`, so `has_route_absorption_lag()` puts it on the event-driven dual walk where the two engines already agree.

**3. `modified_release.rs:779` — the memo defeats the gate's own premise.**

> *"a property of the **model and subject shape**, not of the parameter point"*

`mr_scope` contradicts that on four lines: `pk_param_fn(theta, eta, …)`, `identify_disposition(spec, &pk.values, 1.0)` (a numeric Jacobian probe *at that point*), `recover_disp_params_g` (`None` on a non-positive slope), and `route_flips(f, &dp, &pk.values)` (an explicitly parameter-dependent domain check). Keyed on `(spec, subject)` the gate validates **only the first (θ, η) a fit visits** — the initial estimates, where models are best-behaved. A closed form that degrades toward `ka ≈ ke` or a 2-cpt `α ≈ β` confluence is never re-checked. A coarse disposition/parameter bucket in the key, or an every-N-th re-verify, keeps most of the saving without the hole.

**4. `pk/mod.rs:2304` — the reroute is wider than the fixture that certifies it.**

`mr_scope` already declined `init_fn`, `dose_attr_map`, and `ss || is_infusion`, so those models were on the **dense** path and are now on the **event-driven** one. The two engines are known to disagree about where the timeline breaks around a lagged dose (#1073) and seed `init` from different snapshots (`initial_state(pk_params_flat)` at `t=0` vs `initial_state(&init_pk.values)` at the first record). `rerouting_a_bare_tafd_rhs_does_not_move_the_predictions` has no `init`, no SS, no `dose_attr_map`, no lag — so the docstring's absolute *"Widening this predicate never changes a model that was already correct"* rests on one plain-bolus fixture. Either narrow the claim or add the `init(...)` / SS cases.

---

### Tests

**5. `tad_reading_rhs_reduces_to_its_ode_twin` compares production against itself.** Post-fix `compute_predictions_with_tv` takes `uses_time == true` and calls `compute_event_pk_params_into` + `ode_predictions_event_driven`; `event_twin` builds the identical `EventPkParams` and calls the identical function. Every post-walk step is a no-op here (Form-C ODE readout ⇒ `model.scaling == None`, no analytical init, no LTBS), so the difference is **exactly `0.0`** and the `< 1e-6·(1+|b|)` tolerance is never exercised. This is the same by-construction agreement the PR body correctly diagnoses in `mr_vs_prod` — one layer over. Same for the twin half of `a_tad_rhs_under_a_lagtime_predicts_finite_values` (its finite/positive assert is real and does its job). The load-bearing oracles here are `the_tad_term_is_live_in_the_predictions` and the NONMEM anchor; these two only pin the route. Comparing against the *dense* `ode_predictions` — as `rerouting_a_bare_tafd_rhs_…` already does — would make them independent.

**6. `modified_release.rs:783` — the gate is a 5% net at default tolerances.** `(500·reltol).max(1e-4)` is `0.05` for any model that leaves `ode_reltol = 1e-4`. #1124 itself was 2.2e-4 at `t=0.5` and 1.4e-1 at `t=4.4`; the "~12× detection margin" is against the 5.8e-1 worst case only. The trade-off is defensible (a false positive panics a debug build) but the doc comment should say the net is coarse at defaults and tight only for models that pin `reltol`, so it is not later read as a general guarantee.

**7. `parser/model_parser.rs:18894` — the new walk's arms are essentially untested.** `mr_scope_declines_every_model_time_spelling` exercises `Op::PushTime` in a bytecode statement and `Condition::Compare`. Untested: `Condition::And` / `Or` / `Not`, `Expression::Power`, `UnaryFn`, `ThetaGather { idx }`, and `Expression::Time` via a plain `Assign`/`DiffEq`. A wrong `false` from any of them is a silently dropped term with no diagnostic — the exact #1124 failure mode. Direct Tier-1 tests on `stmts_read_time_builtin` (`if (TIME > 6 .and. X > 1)`, `.not.(TIME > 6)`, `KE = exp(0.1*TIME)`, `KE = 2.0^TIME`) are cheap and pin it structurally.

**8. `tests/mr_tad_nonmem_anchor.rs:127` — "pin the stepper" pins nothing.** `FitOptions` never sets `ode_method`, and `FitOptions::default()` is `OdeMethod::Auto` (`types.rs:6373`). The committed `nonmem_anchor/mr_tad_fit.ferx` *does* set `ode_method = rk45`, and no test reads that file — so twin and test disagree at birth with nothing to catch it. Add `ode_method: OdeMethod::Rk45`. (The comment is inherited verbatim from `tad_lag_nonmem_anchor.rs:126`, which has the same gap — worth fixing both.)

**9. `the_verify_gate_accepts_an_agreeing_closed_form` asserts `twin == twin`.** It passes the twin vector in as the closed form, so it holds for any tolerance down to `0.0` — including a `verify_tol = 0.0` that would panic on every real admitted model in debug. It does catch an unconditionally-panicking gate, which is its stated job; `the_verify_gate_stays_quiet_at_default_solver_tolerances` is the one carrying the real weight (it goes through `mr_predictions`). Worth saying so in the doc comment.

**10. `modified_release.rs:785` — the gate can go vacuous silently.** `twin.len() != mr.len()` returns, and every non-finite twin entry is `continue`d — with the memo key already consumed, so that `(spec, subject)` is never checked again and nothing records that it was skipped. The suppression reasoning is right (a broken twin is its own defect); make it observable, or don't insert the memo key when no comparison was actually performed.

---

### Smaller

**11. `modified_release.rs:647` — the predicate is duplicated.** `mr_scope` inlines `p.uses_time_vars() || p.reads_time_builtin()` while `pk::model_uses_time_anywhere` holds the same expression. `mr_scope` already calls `crate::pk::model_uses_time_builtin(model)` 48 lines earlier (`:599`) — swapping that one call for `model_uses_time_anywhere` subsumes both checks and deletes the block at `:640-649`. As written, a fifth spelling added later has to be wired into two places, and updating one reproduces the split this PR exists to close. (Same shape in the parser: `expr_reads_time_builtin` / `cond_reads_time_builtin` / `stmts_read_time_builtin` mirror `expr_reads_slots` / `cond_reads_slots` / `stmts_read_slots` arm for arm; a generic leaf predicate would carry both.)

**12. `docs/model-file/ode-models.qmd:239` — the re-run callout names only `TAD`.** The section body and the CHANGELOG both list all four spellings including the `if`-condition case (measured 4.0e-1), but the callout — the part a user skims — tells only `TAD` users to re-run. Someone with `if (TIME > 6.0) { KE = 3.0 }` reads it and concludes they are fine.

**13. `CHANGELOG.md:36` — the #1110 entry drops the scoping the PR body gives it.** The body is explicit that the `NaN` is gone from *the two ordinary prediction entry points* and that `ode_predictions` is still reached on the SDE/EKF variance path, unaudited. The user-facing entry says the `NaN` is fixed. Since #1110 is deliberately staying open, name the cell.

---

### Answers to the open questions

- **#1110** — keep it open, but retitle/re-scope it to the convention half plus the SDE/EKF residual path, and say in the issue that the two ordinary entry points are fixed here. Splitting a fresh issue for the convention question is fine too; what matters is that "returns NaN" no longer describes what is open.
- **The `WT` FIXED-zero workaround in `tad_lag_{A,B}`** — yes, worth a follow-up, and leaving them untouched here was the right call. Removing `THETA(5)` needs a fresh NONMEM run and a new committed OFV; that does not belong in a PR whose job is to change the routing.

Nothing above touches the `Dual2` kernels, and the FD-parity requirement is met for the cell the new test covers — my point 2 is that the *other* cell was never routed through one.

TeunP:
Fixed in `716cd5e5` + `f27c3761`, rebuilt on main in `09cf144b`, issue reference in `bd849fbe`. Eleven of thirteen applied; **2** is answered by measurement rather than a change, and one thing I found while checking **4** got settled mid-review by #1073 landing. Ten mutations, one per claim, each killing its target test.

Branch is merged up to main — it had gone `CONFLICTING` against #1123/#1125, which is why CI stopped reporting on it.

---

### 1 — SS gate. Fixed, and it is worse than the gate.

You were right that the gate was asked half-way, and right that its own test worked around the gap. Added `OdeRhsProgram::reads_model_time()` and switched all three sites (`ode_tvcov_supported`, `ode_iov_subject_supported`, `inner_optimizer`'s attribution mirror), with `ode_ss_bare_time_builtin_rhs_routes_to_fd` covering the spelling the existing test could not.

But before writing the clause I checked what declining actually buys, against an explicit **40-cycle pulse train** — the oracle with no SS machinery in it at all:

| RHS term | `uses_time_vars` | `reads_time_builtin` | `SS=1` at t=122 | pulse train | worst rel |
|---|---|---|---|---|---|
| *(autonomous — control)* | false | false | 10.027269 | 10.027269 | **7.6e-9** |
| `0.0*TAD` | true | false | **`NaN`** | 10.027269 | — |
| `0.03*TAD` | true | false | **`NaN`** | 8.890201 | — |
| `0.003*TAFD` | true | false | **`NaN`** | 5.122100 | — |
| `0.003*T` | true | false | 9.545089 | 8.015275 | **1.7e-1** |
| `0.003*TIME` | false | true | 9.545089 | 8.015275 | **1.7e-1** |

`T` was **already** declining and its value is 17% wrong anyway; `T` and `TIME` produce the identical number, so only the flags differed, never the runtime. And `0.0*TAD` — coefficient zero, contributes nothing — returns `NaN` where the pulse train gives the autonomous answer exactly.

So the gate closes a real asymmetry in the **gradient** route, and declining is still right (an analytic jet of a wrong value is worse than an FD one), but it does not make SS correct. The value defect is separate and larger than a routing PR — filed as #1139, with the pulse-train table and the reproducer, and referenced from each of the four sites.

### 2 — `ode_tvcov_supported:655` / `output_columns:602`. Not changed; measurement recorded there instead.

The structural point is right — the value path widened and these did not. I could not measure an effect. Against central FD of the production predictor, two bolus doses, `reltol = 1e-10`:

| RHS | value | `∂f/∂η` | `∂f/∂θ` |
|---|---|---|---|
| `0.3*TAD` | 8.9e-12 | 1.5e-9 | 3.0e-10 |
| `0.03*TIME` | 1.0e-12 | 6.4e-10 | 5.0e-11 |
| `if (TIME > 3) {…}` | 7.0e-9 | **4.6e-3** | 4.8e-4 |

The `if` row looked like your finding and is not. The RHS is discontinuous at `t = 3`, so the stepper places steps differently either side and the *value* is only good to ~1e-8; FD then amplifies that by `1/h`. Swept at `reltol = 1e-10`: `h = 1e-3 → 1.5e-6`, `1e-5 → 3.3e-5`, `1e-7 → 2.7e-2`. At `reltol = 1e-12` the same points are `1.2e-7`, `5.0e-6`, `5.4e-4` — a 100× improvement from tightening the *solver*, which a wrong jet would not show. The smooth-`TIME` control stays ~1e-9 across the whole sweep.

The reason it is empty in effect: the conditions under which the static and event-driven walks genuinely diverge — a lagtime, TV covariates, SS, resets, a modeled dose — are already in the trigger list, so such a subject never lands on the static walk. Left as is with the table and sweep in the comment, and a note to widen it only with a case that measures a divergence. Happy to change it if you have that case.

### 3 — Memo key. Fixed.

You are right that the comment contradicted `mr_scope` on four lines. The key now carries a coarse disposition bucket, `round(10·log₁₀ p)` per parameter — re-verifies on a >26% move, still collapses the thousands of evaluations inside one optimiser step. Two tests: `the_verify_gate_rechecks_after_the_disposition_moves` (5 → 15 on `CL` must re-check) and `the_disposition_bucket_is_coarse_enough_to_memoize` (a 0.1% nudge must not).

### 4 — Reroute claim. Narrowed, and the `init` case added.

`rerouting_an_init_seeded_rhs_does_not_move_the_predictions` measures 4.8e-11. The docstring now carries a measurement table instead of the absolute claim: `6e-13` bare-`TAFD`, `4.8e-11` `init`, and — since #1073 landed mid-review and made the dense path finite under a lagtime — `5.8e-12` and `5.3e-12` for the lagged `TAD` cases that previously had no comparable second engine. One case is still named as *not* comparable rather than asserted: SS is wrong on both engines per **1**, so an engine-vs-engine assert there would pin agreement on a wrong answer.

The `init` test also asserts `model_uses_time_anywhere(&model)` — without it the test passes under a *narrowed* predicate too, since production would then also be dense. That mutation survived a first run.

### 5 — Twin test. Fixed.

Now compares against the dense engine, plus an assert that the two do **not** agree bit-for-bit, so it cannot silently collapse into a self-comparison again. Mutating it back to `event_twin` kills it. The lagtime test's twin half keeps `event_twin` and now says why the comparison is a route pin rather than an oracle, and that its finite/positive assert is the load-bearing half — the independent references for that cell are the NONMEM `tad_lag_{A,B}` anchors and (post-#1073) the 5.8e-12 engine comparison in **4**.

### 6 / 9 / 10 — Verify gate. All three fixed.

The tolerance comment now states the net is `0.05` at the default `reltol` and that the ~12× margin is against #1124's worst observation only (5.8e-1), the same defect being 2.2e-4 at `t = 0.5`. The accept test says it is `twin == twin` and holds for any tolerance including `0.0`. A vacuous call releases its key.

On **10**, one correction: only the length-mismatch path is reachable from a caller — the `compared == 0` path needs the *twin* to be `NaN`, which the gate integrates itself. Both release; the length mismatch has a test, the other is covered by mutation.

### 7 — Parser arms. Fixed, and your list was reachable in a different place than it looks.

`resolve_variable_indices` lowers **every** `Assign`/`AssignIdx`/`DiffEq`/`DiffEqIdx` to `AssignBc`/`DiffEqBc`, so in a resolved `OdeRhsProgram` those four statement arms are unreachable and `expr_reads_time_builtin` is reached **only through `if` conditions**. My first draft put `exp(…TIME…)` in the derivative — it compiles to bytecode, and the `UnaryFn` mutation survived. Every expression case now sits inside a condition.

`ThetaGather` is not covered and the test says so: it comes only from a levelled theta reference (`NAME[idx]`), which has no route into an `[odes]` condition. Each case is asserted against a `TIME`-free twin of the same shape, so a blanket `true` fails as loudly as a blanket `false`.

### 8 — Stepper. Fixed in both files. All three anchors still pass, so `Auto` was selecting `Rk45` — now by construction rather than by luck.

### 11 / 12 / 13 — Fixed.

`mr_scope`'s two time gates collapse into one `model_uses_time_anywhere` call. The docs callout now names all four spellings with the `if`-condition example and both measured numbers. On **13** — see below; the entry needed more scoping than you asked for.

---

### One thing you did not ask about — and #1073 then settled it mid-review

While checking **4** I found the #1110 `NaN` was not fully gone, and that the lagtime was never the trigger. Isolated one variable at a time, on this branch before merging main:

| case | before #1073 | after |
|---|---|---|
| lag 0.5, obs 2/4/8/24 (all post-arrival) | finite | finite |
| lag 0.5, obs at 0.4 added | `[0.0, NaN, NaN, …]` | finite |
| lag 0.3, same obs list (0.4 now post-arrival) | finite | finite |
| **no lagtime at all**, dose at 1.0, obs at 0.4 | `[0.0, NaN, NaN, …]` | finite |
| autonomous RHS, lag 0.5, obs at 0.4 | finite | finite |

Any observation before the first dose *arrives* had no `TAD`; `tad_anchor` returned `NaN` "matching the sdtab convention" and it propagated through the state, losing that row **and every later one**. Right for a diagnostic column, wrong as an RHS input.

I shipped a test pinning that as a known gap, with a message saying that if it ever started passing, the convention had been decided and the docs must say so. Merging #1123/#1125 made it pass — **#1073 gave both predictors a pre-arrival anchor**. So the test is now flipped to assert the correct behaviour, keyed on **bit-identity** (adding a pre-arrival row must not move any later prediction) rather than a finite/positive check that a perturbed trajectory would also pass.

Three follow-on corrections from that, all in the merge commit:

- the CHANGELOG's #1110 entry drops the "for observations after the first dose arrives" scoping it had briefly carried;
- two source comments claiming the dense predictor returns `NaN` under a lagtime are corrected — it is finite now, and the two engines agree to **5.8e-12** (lag 0.5) and **5.3e-12** (lag 3.0), which are added to the routing predicate's table where it previously had to say the case was not comparable;
- only the `[diffusion]` / EKF variance path stays open under #1110. Re-measured post-merge, unchanged: `p_obs = NaN`, `ipred = 0` for any `TAD`-reading RHS, lagtime or not (#1131).

The steady-state finding in **1** was re-measured after #1121 too and is unchanged.

---

### Verification

On the merged tree: `--lib` 3725 passed / 0 failed. `--tests --features ci,survival` exit 0, 157 result lines, 0 failures. `cargo fmt --check` clean. Clippy as CI runs it: **0 warnings on any of the 1334 lines this branch adds** (65 in the same files are inherited from main — attributed line-by-line against `git diff origin/main`, not eyeballed). All three NONMEM anchors pass with `Rk45` now actually pinned.

Ten mutations, one per claim. Two of my own tests survived a first pass and had to be strengthened: the `init` reroute test passed under a *narrowed* predicate too (both engines dense ⇒ trivially equal) until it asserted the route, and the `UnaryFn` walk case tested nothing until it moved inside a condition.

Your independent re-run of `mr_tad_deterministic_truth.py` and the `d <= t + 1e-12` probe — thank you, both recorded in the script. I reproduce your **1.749e-05** exactly, and the "6 significant figures" comment is fixed (the table prints 5, as the README already said).

TeunP:
## Adversarial review — 10 angles, max effort

Ten independent finder angles over `origin/main...HEAD`, then verification of each candidate by reading or executing rather than by agreement between agents. Fifteen findings survived. The headline one is a **release-build abort**, reproduced rather than argued, and it turned out not to be this PR's fault — which changed the fix.

---

### 1 — Blocker: an `EVID=2` record aborts the run. Reproduced, and **pre-existing**.

`compute_event_pk_params_into`'s parameter-static arm filled `dose` and `obs` with the subject-static snapshot and left `pk_only` **empty**. Both event-driven walkers — `ode/predictions.rs:4695` and `pk/event_driven.rs:756` — open with the same three unconditional length asserts. `assert_eq!`, not `debug_assert_eq!`, so a release build aborts too.

```
model_uses_time_builtin  = false
model_uses_time_anywhere = true
has_tv_covariates=false  pk_only_times=[3.0]  pk_only_covariates.len()=0
subject_needs_per_event_pk = false
panicked at src/ode/predictions.rs:4695: assertion `left == right` failed
  left: 0   right: 1
```

My first read was that this PR caused it, by widening the routing predicate without widening `subject_needs_per_event_pk`. That was wrong, and worth stating plainly because it changed the fix. A second probe with an **autonomous** RHS and an `EVID=3` reset — `model_uses_time_anywhere = false`, so this PR's predicate is not involved and the reset clause is untouched by it — aborts identically. The hole has been reachable on `main` ever since resets were wired to that walker. #1124 widened the door, it did not build it.

Reaching it needs only a dataset with `EVID=2` rows plus a varying column the model does not reference: `prune_irrelevant_tv_covariates` clears `pk_only_covariates` and leaves `pk_only_times` populated, which is exactly the state where `subject_needs_per_event_pk` is false and `pk_only_times` is not.

So the fix is **not** to widen `subject_needs_per_event_pk` — that would materialise time-varying snapshots for models whose parameters genuinely do not vary, and would fix only one of the two routes. It is to make the static arm emit one snapshot per `pk_only` record, so the triple is structurally consistent. That arm decides whether snapshots *differ*, never how many there are. One change, both routes, both engines.

Pinned at two levels: a Tier-1 unit test on the materialiser (`event_pk_params_fills_pk_only_on_the_parameter_static_arm`) and an end-to-end test covering both routes (`an_evid2_record_does_not_abort_a_parameter_static_subject`), the reset arm asserting `!model_uses_time_anywhere` so the pair cannot later be misread as a #1124 regression. Both carry a `!subject_needs_per_event_pk` precondition — without it they pass via the other branch and prove nothing. Mutation-tested: reverting the fill fails both.

### 2 — The PR's own thesis, turned on itself: five gates asked the narrow predicate

`reads_model_time`'s doc comment says pairing the flags by hand per call site "is exactly how #1124's `TIME` gap survived". The PR then widened one routing site and left its siblings on `compiled_model_uses_time_builtin`, which inspects only the individual-parameter program. Five angles converged on the same lines independently.

**`stats/likelihood.rs:245`** — the joint PK-TTE shared solve. Its own comment states the contract it broke: *"It is only equivalent when the standalone prediction path takes that same snapshot route. Reject every case it routes elsewhere."* After the reroute, a joint model whose `[odes]` RHS reads model time still passed this gate, so the objective's Gaussian term came from `ode_predictions_and_chz` (dense, single `t=0` snapshot) while the reported IPRED came from the event-driven walk. The OFV was minimised on one engine and the diagnostics reported from another. Fixed.

**`api/output_columns.rs:602`** — the `[derived]` integral grid, whose comment claims consistency "with `compute_predictions_with_states`". That function was rerouted; this one was not, so a `[derived]` integral was silently computed on the dense engine instead of taking the NaN convention every neighbouring arm (IOV, TV, reset) uses for exactly this case. Fixed.

**`sens/ode_provider.rs:685`** — this one is a correction to my own earlier review response. I closed the previous round's finding 2 as "real in structure, empty in effect", on the argument that every condition under which the two walks diverge was already in the trigger list, and a measured FD table backing it. The argument had a hole and the table did not cover it: **`subject.has_resets()` is not in that trigger list**, though it *is* in the value path's condition. A reset subject on a non-autonomous RHS had its value integrated event-driven and its gradient taken on the static walk, which re-anchors `TAD` across the reset. The two-bolus fixtures behind that table contain no reset, so nothing measured it. The comment asserting otherwise was load-bearing — it was the stated reason not to widen. Widened, and the reasoning replaced with what actually holds.

### 3 — What I did not fix here, and why

**The `identify_disposition` probe is unsound on the state axis too.** It probes at `u = 0`, `eᵢ`, `2eᵢ` only, so a term that is a product of two distinct states is exactly zero at every probe point. `- KON*central*periph` on a 2-cpt model passes every check — mass conservation, coupling, sign — and a clean canonical 2-cpt is identified with the binding term dropped. That is #1124's exact failure shape with no time variable anywhere, so `reads_model_time()` is false and nothing here catches it. Same for a branch on the **state** (`if (central > 10) { KE = 3*CL/V }`), which the probe never enters — the PR's own `mr_scope_declines_every_model_time_spelling` tests that branch shape, but only with a time variable in it. This is pre-existing, it is a different oracle axis, and the fix (an off-axis probe at dose scale with a residual check, which would also work in release) is its own piece of work — filed as #1149.

**The verify gate protects less than its comments claimed.** It is `#[cfg(debug_assertions)]`, and every profile CI builds tests with inherits `release` — no `debug-assertions` override exists in `Cargo.toml`, `.cargo/`, or any workflow — so neither the gate nor its seven tests is compiled in CI, and users' release fits have no runtime net. That half is already tracked as #344, and it is worth noting how much this PR leaned on it: the tolerance comment cited the gate as what makes admitting a closed form safe. On top of that the memo key omits every absorption parameter (so the `ka ≈ ke` confluence its own doc named as the motivating case is precisely what it cannot see), omits `reltol` while `sync_ode_solver_opts` mutates it in place, and keys the subject by `id` alone, so a single-dose verdict is reused for a multi-dose regimen — the one case CLAUDE.md's non-degeneracy rule says must always be checked. The `OdeSpec` address is not a rare collision either: a parse → drop → reparse probe measured the same address 8/8. I have not changed the key here; I have removed the claims that overstated it, so nobody treats the gate as the reason admitting a closed form is safe. Filed as #1150.

**Adaptive.** The driver gates `event_pk` on the same narrow predicate, but that gate is genuinely about whether PK *parameters* vary, and for a time-free-parameter model they do not — passing snapshots there would be busywork. The real gap is that the #391 degenerate oracle compares exactly the driver against `predict()` and its only fixture has an autonomous RHS, so nothing exercises the pair after the reroute. That is a test to add in the adaptive epic, under its own validation rules — filed as #1151.

### 4 — #1139: the design question is decided by the code, not by preference

I opened #1139 saying the resolution was open — "for `TAD` a periodic steady state exists in principle, so rejecting forecloses the better fix". That framing was wrong in one specific way: it treated the four spellings as **one** decision. They are two.

A steady-state dose asks for the fixed point of the one-cycle map `u = P(u)`, so the only question is whether that map exists. `TAD` resets at each dose, so every cycle sees the identical field and `P` is well defined. `TAFD`/`T`/`TIME` read an absolute clock, so cycle *k* and *k+1* have different fields and there is no single `P`.

Measured against an explicit pulse train ending where the `SS=1` dose sits, growing the train backwards:

| RHS term | N=20 | N=40 | N=80 | rel change at N=80 |
|---|---|---|---|---|
| `0.0*TAD` | 10.027207 | 10.027269 | 10.027269 | **3.8e-11** |
| `0.03*TAD` | 8.890195 | 8.890201 | 8.890201 | **5.0e-13** |
| `0.003*TAFD` | 6.736388 | 5.175697 | 3.785408 | **0.27** |
| `0.003*T` / `TIME` | 8.015172 | 8.015274 | **150.60** | **17.8** |

`TAD` converges to 13 significant figures — a periodic steady state exists and ferx returns `NaN` for it. `TAFD` drifts 27% per doubling with no limit. `T`/`TIME` diverge outright once the train passes `t = −333.3`, where `1 + 0.003·t` changes sign and the decay becomes growth — a threshold predicted before the run and hit on schedule.

The `NaN` is plumbing, not a modelling limit. `model_parser.rs:11339` maps a missing `params[MAX_PK_PARAMS+1]` to `NaN`, and `equilibrate_ss_state` passes the bare `pk_params_flat`, which has no such slot. That is also why `0.0*TAD` returns `NaN`: the coefficient never gets to matter, because `0.0 × NaN = NaN`.

And the solver `TAD` needs is already written and already generic. `dosing.rs:189` `periodic_ss_fixed_point_g` builds `I − M` column by column from `advance_unforced(eᵢ) − z0` and self-checks by running one and two forced cycles — it tests **affineness in the state**, never time-invariance. `advance_unforced`/`advance_forced` already integrate the *cycle-local* window `(0, ii)`. A state-linear, time-varying RHS on a fixed cycle-local clock gives an affine one-cycle map, so `(I − M)⁻¹·b` is exact for it. The missing piece is one value: in cycle-local coordinates the last-dose anchor is `0.0`.

**Resolution: implement `TAD`, reject `TAFD`/`T`/`TIME`.** The rejection is not a scope limit a later PR lifts — no periodic steady state exists for those, so declining is the permanent right answer. Whatever lands must be anchored multi-dose: the point of `TAD` under SS is a later dose arriving with residual drug present, and `periodic_ss_fixed_point_g`'s self-check is an internal consistency test, not an external oracle. #1139 updated.

---

### Verification

Fixes are in `ada4ea8d`. **The production surface is four lines** — three in the `pk_only` fill, one at each of the three gate swaps; the other ~340 added lines are tests and documentation.

`--lib`: 3727 passed, 0 failed (+2 from the new regression tests). `--tests --features ci,survival`: exit 0, 157 result lines, 0 failures. `cargo fmt --check` clean. Clippy as CI runs it: **0 diagnostics on any of the 342 lines this change adds** — cross-referenced line by line against `git diff`, not eyeballed; the 305 crate-wide are pre-existing.

All three new tests mutation-tested: reverting the `pk_only` fill fails both materialiser tests, and reverting the `ode_provider` widening fails the routing test. The blocker itself was reproduced by execution before the fix and re-run after, on both routes.

One thing worth flagging about this review's own method: of the fifteen findings, the two that mattered most came from angles that *disagreed* with the framing I gave them. I told every finder the abort was this PR's regression; the removed-behaviour angle checked and found it was not, which is what redirected the fix from the routing gate to the materialiser. And the efficiency angle's measurement is the reason the reroute's cost is now written down instead of assumed to be zero.

Issue #1126, opened 2026-08-29
bug(ode): TAD has no referent inside a steady-state dose's pre-arrival window


TeunP:
## Summary

After #1121, an `SS=1` dose with a lagtime loads its periodic compartments at the dose
**record** and the walk carries that state to the lagged arrival. The state in
`[t_dose, t_dose + ALAG)` is therefore now physically real — it is the previous cycle's
decaying tail — but `TAD` in that window still has no referent pointing at it.

`ode_predictions_event_driven`'s anchor filter is

```rust
.filter(|(i, d)| d.time + dose_lagtimes[*i] <= cur_t + 1e-12)
```

so a lagged SS dose contributes nothing to the `TAD` anchor until its arrival. Inside the
pre-arrival window the anchor falls back to some *earlier* dose, or — for a first dose — to
`first_arrival_ed`. Neither is the right instant.

## The right answer here is determined, not conventional

This is **not** #1110. That issue asks what `TAD` means before a subject's genuinely first
arrival, where there is no prior dose to measure from and the answer is a convention. An `SS`
dose is the opposite case: the periodic fiction *does* have a prior pulse, at
`t_dose + ALAG − II`. So `TAD(t) = t − (t_dose + ALAG − II)` for `t` in the pre-arrival window,
and that is the same expression the post-arrival branch already computes — the existing

```rust
let elapsed = cur_t - (d.time + lag);
cur_t - elapsed.rem_euclid(d.ii)
```

already returns `t_dose + ALAG − II` for a *negative* `elapsed`, because `rem_euclid` wraps.
So the fix is likely to be relaxing the filter for SS doses to `d.time <= cur_t` and nothing
else — but that must be measured, not assumed.

## Why no test in #1121 could see it

Every anchor model in `nonmem_anchor/dose_form_lag_*` has an `[odes]` RHS that does not read
`TAD`, so the anchor value is identical either way. The defect is only reachable by a model
whose RHS references the `TAD` builtin.

## Suggested oracle

NONMEM's own `TAD` semantics for an SS record are not a clean external reference, so the
stronger test is internal and of the kind #1073's anchor C uses:

* **Segment invariance.** Inserting an extra observation inside the pre-arrival window that
  carries the same covariates as the record already governing it must not move any other
  prediction. Anchoring at `cur_t` instead makes `TAD` a sawtooth whose shape depends on the
  sampling mesh — the exact failure mode measured in #1073 (4.2e-4 at *every* later time from
[...]

PR #1136, opened 2026-08-30
fix(sens): thread TAD as a dual through the ODE sensitivity walks [closes #1070]

### LINKED ISSUE #1070: bug(sens): TAD/TAFD are lifted as f64 constants, dropping the lagtime term for a TAD-dependent ODE RHS with IIV on the lag

TeunP:
## Why

`TAD`'s anchor **is** the dose's lagged arrival (`d.time + ALAG`). But `eval_rhs_anchored` computed `tad = t − last_dose_eff` in `f64` and `eval_rhs_g` wrote it into the RHS variable table as a lifted constant, so `∂TAD/∂lag = −1` never entered the dual chain. Every analytic η/θ gradient of a `TAD`-reading `[odes]` RHS under an estimated lagtime was therefore wrong — everywhere the trajectory is integrated, not merely at a dose event. The *value* stayed exact, which is why no prediction test ever caught it, and FOCEI builds its `h` matrix from that same gradient, so the error reached the reported OFV.

[#1122](https://github.com/FeRx-NLME/ferx-core/pull/1122) shipped the interim fix: route the cell to finite differences. This is the real one — the term is now carried, and the FD detour is removed.

Re-measured at `df8bb8d6` with the interim gate bypassed, against FD of the production predictor:

| geometry | value | worst `∂f/∂η` | worst `∂f/∂θ` |
|---|---|---|---|
| `ALAG = θ·exp(η)` | 4.3e-16 | **6.63e-1** | **6.78e-1** |
| `ALAG = θ` only | 0 | 4.2e-9 | **6.76e-1** |
| `ALAG = 0.5` literal | 2.1e-16 | 1.1e-9 | 3.4e-9 |
| IOV, κ on `ALAG` | 0 | **1.98e-1** (vs FD of `predict_iov`) | — |

After the fix all four sit at the FD noise floor: **3.7e-9 / 6.6e-9 / 4.8e-9 / 8.9e-10**. Predictions are bit-identical throughout — this changes gradients only.

## What changed

`tad` is threaded as a dual instead of an `f64`:

- `OdeRhsProgram::eval_rhs_g` takes `tad: T` and writes it straight into the variable table. `tafd` stays `f64` — it anchors at the *unlagged* first dose record in both engines, so it genuinely carries no jet (measured exact).
- `eval_rhs_anchored` takes `last_dose_eff: T` and forms `TAD = T::from_f64(t) − last_dose_eff`.
- `integrate_tvcov_g` tracks the anchor as a dual: a new `arrival_dual(k) = T::from_f64(d.time) + lag_dual(k)` (the per-dose lag was already available as a dual — the walk was calling `.val()` on it), seeded at the subject's **first** arrival and advanced at each `K_DOSE` through a new `later_arrival` helper. `boundary_velocity` takes the dual too, so the one-sided velocities entering the saltation carry the anchor's jet — which can only show up in the Hessian, since they multiply a `δlag` whose value is zero.
- The `#1070` FD clauses come out of `ode_analytical_supported` and `ode_iov_supported`, the bail-reason string comes out of `inner_optimizer`, and the now-dead `OdeRhsProgram::uses_dose_anchored_time_vars` flag is removed with them.

Three anchors deliberately stay `f64`, each documented at its site:

| site | why |
|---|---|
| `jdotg_value` / `inject_rate_saltation` / `general_rate_off_saltation` | they return `Vec<f64>` and are multiplied by `δlag²`, so no jet of theirs is ever read — the coefficient is lifted whole. That is also *why* the second-order gap below survives this PR: making the anchor dual there cannot help while the result is a scalar |
| `integrate_g` (static walk) | folds the *unlagged* dose times, and `ode_subject_supported` declines an estimated lagtime outright — the term is absent, not dropped |
| the SS embedded-jet flow-back | SS + a `TAD`-reading RHS declines to FD through the broader `uses_time_vars` gate, and the SS anchor's own wrap convention in the pre-arrival window is unresolved ([#1126](https://github.com/FeRx-NLME/ferx-core/issues/1126)). Lifting a jet that cannot be validated would be worse than lifting none |

## Alternatives considered

Keeping the FD gate and narrowing it (a literal-constant lagtime is exact but was declined with the rest) — rejected: it needs a "does this lag carry a θ/η jet" predicate that does not exist, and it leaves the wrong chain in place rather than fixing it. This PR makes the question moot.

## Cross-repo dependency

No public API change; `gradient_method_inner` reporting for affected models flips from `finite differences` back to `analytic (Dual2)`, which ferx-r surfaces but does not branch on.

## Breaking changes

- [x] None

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0**, both committed anchors (`tests/tad_lag_nonmem_anchor.rs`)

| route | `tad_lag_A` | Δ | `tad_lag_B` | Δ |
|---|---|---|---|---|
| NONMEM | −271.990 | — | −169.317 | — |
| ferx analytic, `tad` lifted as `f64` | −272.1658 | 0.176 | −169.4240 | 0.107 |
| ferx FD (the interim #1122 route) | −271.9897 | 0.0003 | −169.3480 | 0.032 |
| **ferx analytic, `tad` threaded as a dual** | **−271.99037** | **0.0004** | **−169.34856** | **0.032** |

The analytic route now reproduces the FD route to the digits printed: two independent gradient paths agree, and both agree with NONMEM. The residual 0.032 on `B` belongs to the NONMEM control stream's `MTIME` construction, not to either ferx route — against an independent stdlib-RK4 reference ferx matches to 8.5e-6 while the control stream's table print resolves to 3.3e-5 (documented in that file's module docs).

The anchors' assertions were flipped accordingly: they now require the **analytic** route (FD is also right here, so only the reported route separates "the chain is correct" from "something quietly stopped claiming the model"), and the OFV tolerance is tightened `0.5 → 0.06` so it rejects both pre-fix deltas. Mutation-verified: reverting `*d = tad` fails both anchors at 0.175 / 0.109.

## Performance

- [x] Faster for affected models — they leave the FD fallback #1122 put them on, which costs `~2·n_eta+1` predictions per inner step.

  Every other model pays one extra dual subtract (`T::from_f64(t) − last_dose_eff`) per **RHS evaluation** — per RK stage, not per dose. `Sub` for `DualMixed` is `self + (-rhs)`, so with the `from_f64` zero-fill that is three jet-sized passes where the old code did one `f64` subtract. Benchmarked (standalone `DualMixed` replica, `rustc -O`, min-of-8):

  | width | f64 sub | dual sub | % of an RHS stage |
  |---|---|---|---|
  | `Dual2<24>` (`MAX_ODE_AXES`), light RHS | 128 ns | 787 ns | **+3.41%** |
  | `Dual2<48>` (IOV), light RHS | 994 ns | 4815 ns | **+4.04%** |
  | `Dual2<24>`, heavy 6-state RHS | 146 ns | 616 ns | +0.64% |

  So a few percent on a small `[odes]` block, well under one percent on a large one. It is also **avoidable**: `-last_dose_eff` is segment-invariant, so hoisting it out of the RHS closure and using the existing `impl Add<f64> for DualMixed` gives a bit-identical jet at +0.10–0.23%. Not done here — it needs a new `PkNum` trait method plus a hot-path change, which does not belong in this PR. Recorded on the issue instead.

  (An earlier version of this section called the cost "one extra dual add per dose" and cited lib-suite wall clock. Both were wrong: it is per RK stage, and the wall-clock numbers were contention from concurrent peer jobs, not signal.)

## Tests

- [x] Unit tests added (`cargo test --lib --features ci`: 3702 passed, 0 failed)
- [x] Regression tests that fail without the fix — every one below is mutation-verified

| test | geometry |
|---|---|
| `ode_tad_rhs_with_estimated_lagtime_is_analytic_and_exact` | `ALAG = θ·exp(η)`, bolus + lagged infusion, TV covariates |
| `ode_tad_rhs_with_theta_only_lagtime_is_analytic_and_exact` | θ-only lag — the **outer** axis, which no η touches |
| `ode_tad_rhs_with_literal_lagtime_is_analytic_and_exact` | control: zero true jet, so a *spurious* jet shows up here |
| `ode_iov_kappa_on_lagtime_tad_rhs_matches_fd_of_predict_iov` | κ on the lagtime, anchor switching across an occasion boundary |
| `ode_tad_rhs_prearrival_window_carries_the_lagtime_jet` | `init(...)` + records **before** the first arrival |
| `ode_tad_rhs_per_compartment_lags_anchor_on_their_own_axes` | `ALAG1`/`ALAG2` with separate θ and η — the anchor switches between arrivals carrying **different** jets |
| `tad_lagtime_keeps_the_analytic_outer_gradient_route` | the user-visible `sens_supported` / `analytic_outer_gradient_for_interaction` route |

All parity fixtures assert value + `∂f/∂η` + `∂f/∂θ` against FD of the production predictor, plus inner/outer η-parity. Second order is deliberately excluded — see below.

Retained unchanged as the neighbours that must not move: `ode_tafd_rhs_with_estimated_lagtime_stays_analytic`, `ode_tad_rhs_without_lagtime_stays_analytic`, `..._static_walk_stays_analytic`, `ode_tad_rhs_with_route_lag_only_stays_analytic`, `ode_ss_time_only_rhs_still_routes_to_fd`, `ode_provider_ss_tad_dependent_rhs_routes_to_fd`.

### Mutation results

| mutation | effect |
|---|---|
| `*d = tad` → `*d = T::from_f64(tad.val())` | 6.63e-1 / 6.78e-1 / 6.76e-1; IOV 1.98e-1; both NONMEM anchors fail at 0.175 / 0.109 |
| seed only: `T::from_f64(d.time) + lag_dual(k)` → `T::from_f64(d.time + lag_val(k))` | `∂f/∂η_ALAG` 11 % wrong at `t = 0.20`, rising to **56 %** at `t = 4` — the error is injected once in the pre-arrival window and carried multiplicatively, so it is *worse* after the arrival than before it |

## Docs

- [x] `docs/estimation/foce.qmd` — the `TAD × lagtime` paragraph now describes the analytic route
- [x] `CHANGELOG.md` `[Unreleased]` — #1122's entry is **replaced** rather than appended to, since it was never released and would otherwise contradict this one

## Checklist

- [x] `cargo clippy` clean on changed lines (the one remaining `dead_code` warning, `ThetaRefScope::recorded`, pre-dates this branch)
- [x] `Dual2` path stays differentiable — that is the entire change

## Reviewer hints

The load-bearing decisions are all about **which anchors need a jet and which do not**, and each is documented at its site. Worth a close read:

- `later_arrival` — the anchor is a `max` over arrivals, and a `max` of two differentiable functions has a kink where they cross. At an exact tie either one-sided derivative is a valid subgradient; this keeps the incumbent. It also mirrors `f64::max`'s NaN handling deliberately (a bare `>` comparison keeps a NaN incumbent instead of discarding it — unreachable today, which is exactly why it is handled rather than argued about).
- The **post-side** anchor at a `K_DOSE` boundary switched from `t_event` to `arrival_dual(idx)`. These agree on the value by construction (every `K_DOSE` timeline entry is built as `d.time + lag_val(k)`; an SS dose's extra break is a separate `K_SS_SEED` at the raw record time) and there is now a `debug_assert_eq!` pinning that.
- The **seed**. It is written once, outside the event loop, and with no `init(...)` the pre-arrival state is zero, so the term is vacuous on most fixtures. `ode_tad_rhs_prearrival_window_carries_the_lagtime_jet` is built specifically to make it live on both counts, and the mutation table above is what justifies its existence.

Skimmable: the three `equilibrate_*` sites (literal `0.0` → `T::from_f64(0.0)`), and the mechanical `.val()` coercions where an anchor feeds a value-only consumer.

## What this does NOT fix: second order

Worth reading before approving, because I added a Hessian assertion, watched it fail, and then removed it — deliberately.

First order is exact everywhere measured. `∂²f/∂η_ALAG²` is **not**, and the reason is a different defect: the `δlag²` saltation coefficient is built from `jdotg_value`, which returns `ẍ = J·g` — the **state** Jacobian only — while `ẍ = ∂f/∂t + J·f` and a `TAD`-reading RHS is non-autonomous. That is [#1075](https://github.com/FeRx-NLME/ferx-core/issues/1075), already open, whose body even names this exception ("the bare user RHS is autonomous *unless it reads `TIME`/`TAD`*").

**Correction from the review round.** An earlier version of this section said that residual "affects the FOCEI Hessian and the covariance step but not the objective or the gradient." That was wrong — I took it from the docs prose without reading the consumer. `d2f_deta2` is read by `sens_outer_gradient.rs` in `score_core`, `prepare_stacked` and `subject_packed_gradient_foce`, which *is* the analytic outer gradient handed to L-BFGS. `docs/estimation/foce.qmd` and the #1075 comment are corrected too.

It pre-dates this PR and this PR improves it, measured on the main fixture:

| state | worst `∂²f/∂η²` |
|---|---|
| before #1070 (`tad` lifted as `f64`) | **6.40e-1** |
| after this PR (`tad` threaded as a dual) | **2.23e-1** |
| after this PR + seeding the time axis in `jdotg_value` | 1.51e-1 |

Every neighbour is exact to 2e-7 — η-on-lag with the `TAD` coefficient zeroed, `TAD` with a θ-only lag, `TAD` with a literal lag — which localises it to the `δlag²` η-block, the only place `δlag` has an η component.

I tried the one-line fix that #1075 itself suggests (seed the anchor with `−1` in `jdotg_value`, so `TAD = t − anchor` carries `+1` on the same `ε` axis and one evaluation returns `J·g + ∂f/∂TAD`). It recovers a third of the gap and no more, so I **reverted it**: a partial fix to a primitive every saltation site shares makes the residual harder to attribute later, not easier. The measurement and that finding are posted to #1075.

Two consequences in this PR:

- The two multi-dose fixtures assert first order and carry an explicit comment saying second order is *not* asserted, with the numbers and the issue reference. Not a silent omission.
- `docs/estimation/foce.qmd` claimed both remaining approximations "are reachable only through that same co-timed geometry now". That is false for #1075 — a plain `TAD`-reading RHS with an η-lag reaches it with no straddling forcing and no colliding boundary — so the paragraph is corrected here.

## Open questions

- The SS embedded-jet flow-back keeps a constant anchor. That is correct *given* SS + `TAD` declines to FD, but it is a coupling between two gates rather than a local invariant. If [#1126](https://github.com/FeRx-NLME/ferx-core/issues/1126) resolves what `TAD` means inside an SS dose's pre-arrival window, that site should be revisited together with it.
- This PR does not touch the closed-form modified-release fast path, which is tried *before* the ODE provider at `sens/provider.rs` and has its own `TAD` defect ([#1124](https://github.com/FeRx-NLME/ferx-core/issues/1124), fixed by [#1130](https://github.com/FeRx-NLME/ferx-core/pull/1130)). The two are independent: `modified_release.rs`'s dual call site passes a constant `tad` at `t = 0` before and after this change.
- #1075's ordering in the #1078 epic has it blocked by #1076, on the argument that #1076 may merge the saltation sites first. That still holds for the co-timed sites, but the geometry above reaches the term through the plain `K_DOSE` arrival, which #1076 does not touch — so the bare-RHS half is separable if the ordering ever becomes inconvenient.

---

# Review round

A max-effort adversarial review (ten independent finder angles plus a gap sweep) was run against `0023ef31`+`35e7d187`. Fifteen findings; three were verified by executing code, and one candidate was refuted the same way. Everything below is fixed in this branch unless marked otherwise.

## The question the review forced, and the measurement that answers it

Removing the FD gate re-admits a cell whose second-order block is still 22% wrong, and that block reaches the outer gradient (above). Both NONMEM anchors run `outer_maxiter: 0`, so nothing here exercised the outer θ-gradient for the cell being re-opened. Two measurements now do.

**Is the objective route-dependent?** No. Three parameter vectors, each evaluated at `outer_maxiter = 0` under both routes:

| point | analytic | finite differences | Δ |
|---|---|---|---|
| analytic optimum | −278.17086221 | −278.16901333 | 0.0019 |
| FD + L-BFGS optimum | −273.25077291 | −273.24886364 | 0.0019 |
| FD + BOBYQA optimum | −277.64012629 | −277.63826147 | 0.0019 |

Identical offset at all three — inner-loop EBE noise, not a route effect. ferx's FOCEI builds `H̃` from `∂f/∂η` alone, so `d2f_deta2` enters the objective's *gradient* but not its *value*.

**Does the degraded curvature land somewhere worse?** No — it lands somewhere better. Full estimation runs (`outer_maxiter = 400`), which are comparable precisely because the objective is route-independent:

| route | OFV | `TVLAG` (simulated at 0.5) | `ω_LAG` (simulated at 0.02) |
|---|---|---|---|
| analytic + L-BFGS | **−279.216** | 0.001 (at bound) | 0.0568 |
| FD + L-BFGS | −273.829 | 0.5075 | 0.0200 |
| FD + BOBYQA | −278.606 | 0.4476 | 0.0064 |

The analytic route reaches a point 4.9 OFV units below where FD + L-BFGS stalls. The residual costs search-direction quality and contaminates standard errors; it does not move the objective, and on the one dataset where this is measurable it does not produce a worse optimum than the FD alternative.

## User-visible change this PR was not disclosing

`Optimizer::Auto` keys on `analytic_outer_gradient_for_interaction`, which this PR flips to `true` for the cell — so `gradient_method = auto` now resolves these models to **L-BFGS** where it previously chose BOBYQA. Same model, same data, different optimizer *and* a better gradient. Now documented in `CHANGELOG.md` and `docs/estimation/foce.qmd`, with `optimizer = bobyqa` named as the way back.

## Defects fixed

- **`debug_assert_eq!` panicked on a NaN lagtime.** `NaN != NaN`, so it aborted the exact optimizer excursion the two neighbouring `debug_assert!`s were written to tolerate — and `[profile.ci-test]` inherits `release`, so CI could never have caught it. Reproduced (`LAGTIME = (TVLAG − 2.0)^0.5`), fixed, and pinned by `ode_tad_rhs_nan_lagtime_degrades_rather_than_panicking`, mutation-verified.
- **Co-timed arrivals differentiated two different anchors.** The post-side velocities were handed `arrival_dual(idx)` while `later_arrival` kept the incumbent for the segment they open. The `K_DOSE` arm now folds **once** at the top and both consume the result, so they agree by construction. The review also showed this is reachable on an ordinary single-`LAGTIME` model, not just two `ALAG{n}` slots: `lag_dual(k)` reads a per-dose snapshot, so a covariate-, `TIME`- or κ-dependent lag gives two doses different jets.

## Test gaps closed

- `THETA_WT` was **not** `FIX` in `PREARRIVAL_INIT_TAD` (it borrowed the anchor's wording but not its `FIX`), so a live `∂f/∂THETA_WT` axis was being exercised. And the doc's routing rationale was wrong — `ode_subject_supported` declines `has_lagtime()`, so the lagtime routes the subject, not `WT`. That false belief is why `integrate_g` had no `TAD` coverage at all.
- The pre-arrival fixture asserted only value properties and never the axis it exists to pin; it now requires `∂f/∂η_LAG` to move both pre-arrival records.
- The IOV fixture exercised only the outer provider; the inner EBE route (a separate `Dual1` walk) was asserted nowhere after the deleted test. It now checks the route and inner/outer parity.
- The two NONMEM anchors shared one `OFV_TOLERANCE`; anchor A is now pinned 60× tighter (1e-3) while B keeps a bound appropriate to its known systematic `MTIME` offset.
- The anchors' flipped route assertion is model-level and cannot see a per-subject FD fallback — the assertion it replaced *could*, because model-level FD implies FD for every subject. A per-subject FD-fallback warning check is added alongside.

## Found, not fixed here

- **The static walk returns `NaN` before the first dose.** `integrate_g` has no pre-first-dose `TAD` fallback where production has had one since #1073: measured `[70.0, NaN, NaN, NaN]` against production `[70.0, 69.072, 157.027, 131.576]`. Pre-existing, needs neither a lagtime nor an `init(...)`, and posted to [#1110](https://github.com/FeRx-NLME/ferx-core/issues/1110#issuecomment-5474294572). This PR does not fix it, but it removes the comment it had added blessing that call site and the claim in `tad_gate_subject`'s doc that #1073 landed the fallback in "both twins".
- **The MR closed form may now claim `TAD` models.** `mr_subject_sensitivities` runs before the ODE provider and never re-checks `ode_analytical_supported`, whose scope this PR widens. Coordinate with [#1130](https://github.com/FeRx-NLME/ferx-core/pull/1130) / #1124.
- The perf hoist above.

## Refuted

`sqrt(TAD)` was flagged as producing an inf/NaN jet at a dose arrival, since `TAD` is now exactly `0.0` there with a live jet and `guard_floor(0.0)` uses `>=`. Measured: finite, values bit-identical to production. `DualMixed::sqrt` carries its **own** guard (`if !(self.value > 0.0)`, strictly greater) returning a zero-jet dual at zero. Two guards; only the outer one had been read.

---

# Review round 2 — verifying what round 1 asserted

Round 1 reported the co-timed-arrival finding as `PLAUSIBLE` and "fixed", and deferred the MR routing interaction as `PLAUSIBLE`. Neither had been executed. Both are now measured, and **one of them was wrong.**

## The co-timed tie-break was wrong, and so was round 1's diagnosis of it

`later_arrival` kept the **incumbent** at an exact tie, documented as "a free choice — pinned here so it is deterministic". That reasoning does not hold. The walk has already committed to one of the two perturbed configurations before the tie-break runs: the timeline is sorted `(time, kind)` by a **stable** `sort_by` with `K_DOSE` pushed in `subject.doses` order, so co-timed arrivals fire in ascending dose index — the ordering of the problem in which the *lower*-indexed dose arrives first. In that configuration the anchor must end as the **last-processed** dose's.

Measured on two doses whose lagged arrivals coincide bit-exactly (`ALAG1 = 1.5` at `t = 0`, `ALAG2 = 0.5` at `t = 1`, both `η = 0`), comparing the `η`-lag block of `d2f_deta2` at the tie against the two one-sided limits at `±1e-6`:

| tie-break | `max\|tie − δ⁻\|` | `max\|tie − δ⁺\|` |
|---|---|---|
| keep candidate (`>=`, adopted) | **6.6e-5** | 1.3e0 |
| keep incumbent (`>`, previous) | 1.5e0 | 8.0e-1 |

`δ⁻` is the limit whose event ordering the stable sort reproduces. Only the candidate rule lands on it; the incumbent rule sits between the two — a mixture that is not the derivative of any single consistent perturbation.

This is **second-order only**, which is why it hid: the boundary velocities' jets multiply a `δlag` whose *value* is zero, so they cannot reach `df_deta` at all. Both rules give identical first-order gradients, and a first-order probe shows nothing.

Pinned by `ode_tad_rhs_co_timed_arrivals_resolve_to_one_sided_limit`, which asserts the tie is bit-exact, that the two one-sided limits are distinguishable (so "matches one of them" is not vacuous), and that the tie matches `δ⁻` and not `δ⁺`. Mutation-verified: restoring `>` fails it at 1.482e0.

**Correction to round 1.** Round 1 diagnosed this as the post-side velocities using `arrival_dual(idx)` while `later_arrival` kept the incumbent for the segment, and "fixed" it by folding one `post_anchor` at the top of the `K_DOSE` arm. Wrong diagnosis. With the corrected tie-break those two expressions are identical in every reachable case — the timeline is sorted ascending, so the candidate always wins. Isolating the changes confirms it: `>=` alone reproduces the fix exactly; the hoist alone changes nothing. The hoist is kept as a **simplification** (one expression of "the anchor this arrival establishes", removing a triple recomputation and a tautological assert) and is now described that way.

## MR routing: measured, and it is not this PR's

Round 1 deferred this on the reasoning that MR never consults `ode_analytical_supported`. Measured by temporarily restoring an equivalent gate:

```
ode_analytical_supported                   = false
sens_supported                             = false
ode_subject_sensitivities (gated provider) = false   <- the gate works
subject_sensitivities                      = true    <- MR reaches the caller anyway
subject_eta_grad                           = true
```

So the MR fast path bypasses the gate entirely, and #1122's interim gate never protected that cell. Removing it here changes nothing for the modified-release path.

The underlying defect is real and severe, and is #1124: on an MR-eligible model with a **bare** compartment `LAGTIME` and a `TAD`-reading RHS, MR's answer is **50.6% off the ODE twin** on the value (37.8% on `∂f/∂η`). A bare `LAGTIME` leaves `dose_attr_map` **empty**, so `mr_scope`'s `!dose_attr_map.is_empty()` decline does not catch it, and `model_uses_time_builtin` inspects the individual-parameter program rather than the RHS. Posted to [#1124](https://github.com/FeRx-NLME/ferx-core/issues/1124#issuecomment-5479175928), flagged for #1130.

Worth recording: checking MR against `compute_predictions_with_tv` instead gives 8.9e-17 on the value — apparently perfect, and entirely circular, because `pk/mod.rs` tries `mr_predictions` first. The bit-exact agreement is the tell, not the evidence.

## Verification on the final code

- `cargo test --lib --features ci` — **3705 passed, 0 failed**
- `cargo test --tests --features ci,survival` — **155 binaries, 0 failures** (round 1 never completed this: it died on a full disk, an environment failure, not a code one)
- Both NONMEM anchors green under `--features slow-tests` at the tightened per-anchor tolerances
- `cargo check --tests`, `cargo fmt --check`, clippy clean on changed lines

---

# `codecov/patch` is red, deliberately — the 90% gate is unreachable on this diff

Recorded here rather than waved through, because the repo gates patch coverage at 90% with a 0%
threshold and this PR merges below it.

Measured by replicating the gate locally — same command CI runs
(`cargo llvm-cov --workspace --tests --no-default-features --features ferx-core/ci
--profile ci-fast --lcov`), then intersecting zero-hit lines with the diff. The replication
reproduces codecov's number exactly (**46/56 = 82.14%**), so the arithmetic below is the gate's
own, not an approximation.

The 10 uncovered lines are all in `src/sens/ode_provider.rs`, and they fall into three classes:

| lines | class | coverable |
|---|---|---|
| 5261, 5584, 5695 | `debug_assert!` **conditions** | **no** |
| 6131, 6132, 6279, 6280 | unreachable `pk_snapshot_equal` else-arms | **no** |
| 5486, 5591 | SS / non-finite-anchor fallbacks | yes, with work |

**The `debug_assert!` lines cannot be covered by any test.** `ci-fast` inherits `ci-test`
inherits `release`, and nothing in the profile chain sets `debug-assertions`, so the macro
compiles out — llvm-cov still emits a region for the condition, which then can never be hit.
This is systemic and pre-existing, not something this PR introduced: **all four** multi-line
`debug_assert!` conditions in the file are zero-hit, including line 2029, which is not in this
diff. It is the measurement gap #293 documents.

**The else-arms are unreachable by construction.** `params` is the first record at index `>= p`;
`post_snapshot` is the first record *strictly* later; every record co-timed with a rate-off sorts
before it, so both always resolve to the same record and `pk_snapshot_equal` is always true.
Verified three ways (analysis, two runtime probes, zero hits across the whole suite) and filed as #1160
separately — it is a pre-existing dead branch, not this change's.

So the ceiling is **49/56 = 87.5%**, below the 90% gate, and no test can close the gap. Merging
red with that stated, rather than reformatting `debug_assert!`s onto one line or deleting a
branch this PR did not create — either of which would clear the gate by gaming it.

## What was added while establishing this

Three Tier-1 tests (`df092562`), kept because they are worth having independent of coverage:

- `later_arrival_never_moves_the_anchor_backwards` — pins the fold's contract directly: an
  out-of-order candidate must not drag the anchor backwards, and a NaN candidate must not become
  it. This is the arm the `>` / `>=` mix-up fixed in `d7c58818` breaks.
- `ode_tad_rhs_infusion_end_inside_a_covariate_step_matches_production` and its `zero_order`
  twin — the rate-off boundaries under a TV covariate with a `TAD`-reading RHS. Both handlers
  share **one** anchor between pre- and post-boundary velocities, so a wrong anchor there moves
  `∂f/∂η` without moving `f`; the other TV-cov fixtures use a `t`-independent RHS and pinned
  that sharing only through a constant. Both carry the `check_vs_production` Dual2-vs-FD oracle
  plus inner/outer parity.

A first draft of the latter two claimed in their doc comments to exercise the general saltation
path. They do not, and the comments now state the measured scope instead.


TeunP:
## Adversarial self-review of `0023ef31`

Independent line read of the committed diff, after the fact — separate from the self-review that ran *during* construction (which caught the `later_arrival` NaN case, the missing `arrival_dual ≡ t_event` assert, and two stale comments). Several edits landed after that pass and had not been read back.

**No behavioural defect found.** Three findings, all mine, all claims that are wrong or under-specified — which is where the risk actually sits on a change whose behaviour is this heavily mutation-tested. Fixed in `35e7d187`.

### 1. `docs/estimation/foce.qmd` — cross-reference points the wrong way

The new paragraph ended "`TAD` combined with a **steady-state** dose is the one cell that still routes to FD, under the broader non-autonomous-RHS rule **below**." That rule is stated *above*, in the same paragraph ("only SS combined with a non-autonomous RHS (reads `TIME`/`TAFD`/`TAD`) stays on FD"). Now names the rule rather than pointing at it.

### 2. `later_arrival`'s tie-break was circular

The doc said it keeps "the earlier one in timeline order". At a tie, timeline order is precisely what is undetermined — so that sentence resolves nothing.

Checked rather than assumed: the timeline is sorted `a.0.partial_cmp(&b.0).then(a.1.cmp(&b.1))` — on `(time, kind)` only — and `sort_by` is **stable**, while `K_DOSE` entries are pushed in `subject.doses` order. So co-timed arrivals fire in ascending dose index and the incumbent is the lower-indexed dose. The doc now says that.

Worth being explicit about what this choice is: at an exact tie the anchor's derivative is genuinely one-sided and either branch is a valid subgradient, so this is a free choice being pinned for determinism, not a correctness claim. Production carries no jet there to disagree with.

### 3. The performance claim was wrong

The PR body said the cost is "one extra dual add per dose". It is one dual **subtract per RHS evaluation** — per RK stage, not per dose event.

Still negligible, but for a different reason than the one given: `eval_rhs_g` already opens every call with `vars.resize(n_vars_total, T::from_f64(0.0))`, writing `n_vars_total` duals of the same `1 + N + N²` size, so one extra `O(N²)` operation is roughly `1/n_vars_total` of the zeroing alone, before `eval_statements_g` is counted. The wall-clock numbers I originally quoted (268s → 273s) are worthless as evidence — three peer `cargo` jobs were competing for cores throughout, and a later run of the same suite took 291s. Corrected on the PR to lead with the structural argument.

## Checked and found sound

Recorded so a later reader does not re-derive them:

- **The exact-equality `debug_assert_eq!(arrival_dual(idx).val(), t_event)` is not 1-ULP-brittle.** Every dual `add` computes `value = self.value + rhs.value` in plain `f64` (`sens/jet.rs`, `sens/dual1.rs`), and `t_event` is built as `d.time + lag_val(k)` with `lag_val(k) = lag_dual(k).val()` — identical operations on identical operands.
- **`idx` at all four `arrival_dual` sites** is the timeline entry's payload (`let d = &subject.doses[idx]` at the head of the `K_DOSE` arm), so each resolves to that dose's own arrival, not a neighbour's.
- **The removed `iov_fd_reason` clause leaves no dangling binding.** Its `let eff = model.effective_for(subject)` was block-scoped, and the #814 block below re-declares it.
- **`tad_lag_B` is non-degenerate.** Its second dose at `t = 12` lands on 1.93 units of residual drug (`DV` at `t = 11.5`), and `t = 14` reads 6.15 against 5.26 for the fresh first dose — so the anchor switch is measured against live state, not from zero. That is the CLAUDE.md rule about a dose event's *incoming* side, and this anchor satisfies it.

## Do new NONMEM anchors need to be built? No — and for one cell, they cannot be

- The behaviour change that moves a **number a user sees** is the OFV, and `tad_lag_A` / `tad_lag_B` are exactly that comparison. They now assert the *analytic* route at a 6× tighter tolerance (0.5 → 0.06) and are mutation-verified to fail at the pre-fix deltas (0.175 / 0.109). `B` is the required multi-dose non-degenerate partner to `A`.
- The three new geometries — per-compartment `ALAG1`/`ALAG2`, IOV with κ on the lag, and the pre-arrival window under `init(...)` — are questions about the *shape of the gradient*, not about a new numerical result. CLAUDE.md's oracle for those is the `Dual2`-vs-FD parity test against the production predictor, which is what they are. None of them introduces an OFV convention NONMEM would arbitrate differently.
- The pre-arrival cell **cannot** be NONMEM-anchored even in principle: NONMEM has no `TAD` built-in in `$DES`, so the anchor there is whatever the control stream's author writes. Anchoring ferx's negative-TAD convention against a control stream I wrote would be circular — the same trap that made option (b) unusable during #1070's design (and the reason #1110 is a semantics decision rather than a bug). The value convention there is #1073's; this PR only differentiates it.

Issue #1132, opened 2026-08-30
test-ci: the adaptive mrgsolve anchors never run on a PR that changes the adaptive walk


TeunP:
## Problem

The eight adaptive mrgsolve anchors are `slow-tests`-gated:

| test file | what it anchors |
|---|---|
| `adaptive_vanco_anchor.rs` | constant-covariate AUC-TDM titration (#391 S2.5b) |
| `adaptive_vanco_renal_anchor.rs` | time-varying covariate (#700) |
| `adaptive_vanco_iov_anchor.rs` | per-occasion κ (#701) |
| `adaptive_vanco_renal_iov_anchor.rs` | covariate × occasion composition |
| `adaptive_vanco_loading_anchor.rs` | base regimen (#702) |
| `adaptive_vanco_renal_loading_anchor.rs` | base regimen × TV covariate (#930) |
| `adaptive_vanco_iov_loading_anchor.rs` | base regimen × IOV (#931) |
| `adaptive_platelet_anchor.rs` | reactive platelet ladder (#391 S2.5a) |

`slow-tests.yml` runs them **daily on a cron and on push to `main`** — i.e. after a merge, never
on the PR that would break them. CLAUDE.md names mrgsolve as *the* external comparator for the
adaptive/feedback-dosing family (NONMEM has no native feedback dosing), so for this engine those
eight are the only external oracle there is, and no PR ever consults it.

## How this bit

#1073 (PR #1123) changes the adaptive walk's segment PK/occasion resolution. Its full per-PR suite
was green — 3949 lib tests, 16 NONMEM anchors, `codecov/patch` — while **three of the eight
anchors were failing**:

```
vanco_renal      decision 4: ferx 7.123033 vs mrgsolve 7.095079  (|Δ| 0.027954 >= 0.02)
vanco_iov        decision 1: ferx 2.984472 vs mrgsolve 2.960637  (|Δ| 0.023835 >= 0.002)
vanco_renal_iov  decision 1: ferx 2.583398 vs mrgsolve 2.556064  (|Δ| 0.027334 >= 0.002)
```

They surfaced only because the anchors were run by hand. Had the PR merged, `slow-tests.yml` would
have gone red on `main` on the merge push, with the cause several commits back.

(The three failures turned out to be the *references* encoding the pre-#1073 convention rather
than a defect in the fix — corrected in `4fa73f05`. That does not change the CI point: a
convention change that invalidates three external references is exactly what a PR-time check is
[...]

Issue #1114, opened 2026-08-28
refactor(workspace): split into ferx-core / ferx-tools / ferx-cli, with a CI-enforced public API boundary


roninsightrx:
## Summary

Split the repo into a cargo **workspace** so model-development tooling —
bootstrap, stepwise covariate modelling, model search, cross-validation — can be
written in Rust without paying for it on every `ferx-core` rebuild, and without
adding a repository to the downstream pin chain. Single repo, strict API
separation enforced in CI.

Full plan: `plans/ferx-tools-split.md`. This issue is **P1** (mechanics + the
API gate); later phases are tracked separately.

```
ferx-core/                  # repo root = ferx-core package + workspace root
├─ Cargo.toml               # [package] ferx-core  +  [workspace] members
├─ src/                     # unchanged
├─ api/ferx-core-public-api.txt   # public-API baseline, diffed in CI
└─ crates/
   ├─ ferx-tools/           # depends on ferx-core
   └─ ferx-cli/             # depends on both; [[bin]] name = "ferx"
```

## Boundary rule

> **`ferx-core`** = one model, one dataset, one fit.
> **`ferx-tools`** = many fits, resampling, model-space search.

If it calls `fit()` more than once, it is a tool. Statistical kernels (npde,
CWRES, shrinkage, covariance/SE, SIR, FREM, VPC statistics) stay in core;
orchestration of repeated fits moves out.

## Motivation

**Compile isolation.** `ferx-core` is LLVM-bound — 93% LLVM / 1% frontend, with
`sens` alone ~90% of IR (#971). Tooling is orchestration code: loops over
`fit()`, non-generic, cheap. As a downstream crate it links the cached
`ferx-core` rlib, so editing `bootstrap.rs` does not rebuild 279k LOC. In-crate,
every tools edit pays the full lib compile plus relink.

This is the *opposite* case to #970's rejected split, which moved the hot generic
core (`sens`) out and would have duplicated monomorphization across a crate
boundary. This one stacks cold code on top and moves nothing.

**Enforced API boundary.** `ferx-tools` can only touch `pub` items, so tooling
goes through the same door R uses — which keeps the R layer thin by construction.

**Feature-matrix hygiene.** Core already has six features that unify across the
dependency graph. Tooling as a seventh multiplies the CI matrix; as a crate it
owns its own.

## Strict API separation — enforced, not conventional

The failure mode to guard against is not `ferx-tools` reaching into internals
[...]

PR #1137, opened 2026-08-30
fix(ode): re-seed init(...) at an EVID=3/4 reset from the reset row's own snapshot [closes #1133]

### LINKED ISSUE #1133: bug(ode): an EVID=3/4 reset re-seeds init(...) with the PREVIOUS record's covariate snapshot, not its own
Split out of #1073 / PR #1123, where this was found and deliberately left alone rather than
half-fixed under cover of a covariate-snapshot change. #1073 established the rule ferx now
follows everywhere else: **NONMEM runs `$PK` at every data record and ADVANs *to* it, so a
segment is governed by the record that terminates it.** An EVID=3/4 row is a data record. ferx
does not treat it as one.

## The site

`src/ode/predictions.rs` (the dense ODE predictor), two places:

```rust
let pk_now: PkParams = match kind {
    // A reset zeros the state below and, in ferx, is not treated as a
    // parameter source: the segment ending at it reuses the previous record.
    Kind::Reset => last_pk,                                    // :4947
    _ => governing_record[i].map_or(last_pk, &record_pk_at),
};
```

```rust
Kind::Reset => {
    // ...
    u = ode.initial_state(&last_pk.values);                    // :5148
    reset_floor = t_event;
}
```

## Half of it is inert; half of it is not

**The propagation half is unobservable.** The segment ending at the reset has its state zeroed (or
re-seeded) at the reset before any readout, so whichever record governs it cannot reach a
prediction. A "fix" there is untestable by construction — I tried one in #1123, the mutation
survived, and it was removed. Do not re-add it. The reasoning is written down at
`src/ode/predictions.rs:2788`.

**The re-seed half is observable, and this is the actual defect.** `initial_state` is a function of
the individual-parameter vector:

```rust
pub fn initial_state(&self, params: &[f64]) -> Vec<f64> {
    match &self.init_fn {
        Some(f) => f(params),
        None => vec![0.0; self.n_states],
    }
}
```

So an `[odes] init(CMT) = <expr>` whose expression depends on a covariate — or on a
covariate-driven individual parameter — is evaluated at a reset with **`last_pk`, the previous
record's snapshot**, not the reset row's. That re-seeded state feeds every downstream prediction;
nothing discards it.

NONMEM would run `$PK` at the EVID=3/4 row and re-seed from the parameters computed there. This is
the reset analogue of #1046/#1055, which established that `A_0` seeds at the **first record**.

## Reproduction shape

Needs all three, or it cannot show:

1. `[odes]` with `init(CMT) = expr` where `expr` reads a covariate (e.g. a baseline `BASE = f(WT)`),
2. an EVID=3 or EVID=4 row,
3. **the covariate changing at that reset row** relative to the preceding record.

Drop (3) and the two snapshots are equal and the bug is invisible — which is why the existing reset
tests (`reset_convergence.rs`, `reset_occasion_segmentation.rs`) do not catch it: they are
constant-covariate.

## Why the existing oracles are blind to it

- **`Dual2`-vs-FD parity cannot see it.** `src/sens/ode_provider.rs:4919` mirrors the production
  arm on purpose ("previous record's params, matching production's `Kind::Reset` arm"), so the twin
  differentiates the same convention the value path uses and both sides move together. This is the
  documented failure mode in CLAUDE.md, and it is how #1079 survived.
- **An analytic-vs-ODE twin cannot see it either.** The closed-form walk
  (`src/pk/event_driven.rs:788`) zeroes every compartment unconditionally and has no `init(...)`
  concept, so there is no second implementation to disagree.

That leaves **NONMEM** as the only oracle, which is what #1046 used for the first-record case.

## Suggested validation

Per CLAUDE.md's non-degeneracy rule, one dataset is not enough — the A/B twin pattern from #1046 /
#1055 fits directly:

- **A**: `init()` model, EVID=3 reset, covariate **steps at the reset row**.
- **B**: identical but the covariate is **flat** across the reset — must be unchanged by the fix,
  pinning that the anchor is measuring the reset snapshot and not something else.
- A multi-dose variant so the reset lands with residual drug rather than on an empty system.
- A `$EST METHOD=1 INTERACTION MAXEVAL=0 POSTHOC` run is enough: the FOCEI `h` matrix is the
  analytic Jacobian, so a wrong snapshot moves the reported OFV.

## Scope note

Only the `initial_state` call needs to change (resolve the reset row's own snapshot); the
`Kind::Reset => last_pk` arm feeding the discarded propagation can stay, with its existing comment
updated to say which half was the real one.

### LINKED ISSUE #1135: Analytical [initial_conditions] baseline is dropped at a reset instead of re-deposited (NONMEM re-applies A_0)
Found while building the NONMEM anchor for #1133. That issue is about the **ODE** path, and
the fix there is landing; this is its **closed-form** sibling, which is worse and needs its
own change.

## What NONMEM does

Measured on NONMEM 7.6.0 with `nonmem_anchor/reset_init_snapshot_B.ctl` (committed as part
of #1133's anchor). One compartment, `A_0(1) = 10*WT`, `WT` flat at 70, a 100 mg bolus at
`t = 4`, an EVID=3 reset at `t = 8`:

```text
   t     IPRED
 6.0     9.320824    <- residue of the t=4 dose
 8.0    14.000000    <- reset row: 700/50, the baseline RE-DEPOSITED
 9.0    12.667724
```

`A_0` is re-applied at the reset. The residue is discarded and the compartment restarts at
the baseline, then decays again. This is not specific to a covariate — arm B has a flat `WT`
— it is simply what a reset does.

## What ferx does

`src/pk/mod.rs`, `add_analytical_init_with`, says so outright:

```rust
/// A system reset (EVID=3/4) zeros every compartment, including the residual
/// baseline; nothing re-deposits the t=0 init afterwards. So the baseline only
/// contributes to observations strictly before the first reset — observations
/// at or after it see none of it.
```

```rust
let first_reset = subject.reset_times.iter().copied().fold(f64::INFINITY, f64::min);
...
if t < 0.0 || t >= first_reset {
    continue;
}
```

So for the same model expressed analytically, every observation at or after the first reset
loses the baseline entirely — `9.32 → 0` where NONMEM gives `9.32 → 14.0`. On a subject with
a baseline comparable to the dose contribution that is an O(1) error in the predictions, and
it moves the objective for any fit of such a dataset.

The two paths therefore disagree with each other as well as with NONMEM, which is the part
that makes it worth fixing rather than documenting: `[initial_conditions] init(central) =
CONC0 * V` and its `[odes] init(central) = CONC0 * V` transcription are advertised as
equivalent (`tests/analytical_init_equivalence.rs`, #521), and after #1133 they are not, on
any dataset with a reset.

## Shape of the fix

The overlay currently lays the baseline down once at the time origin and cuts it off at the
first reset. It needs to lay it down again at **each** reset:

- for each observation, find the latest reset at or before `t` (0 / the record start if
  none) — call it `t_r` — and add `a0(t_r) * kernel(t - t_r)`;
- `a0(t_r)` is evaluated at the **reset row's own covariate snapshot**, which #1133 just made
  available as `Subject::reset_cov(r)` / `EventPkParams::reset`, so the plumbing already
  exists;
- the IOV variant (`obs_decay_pk`) needs the same treatment for its decay kernel.

Multiple resets matter here: the current code takes `min` over `reset_times`, which is only
right because it is cutting off rather than re-depositing.

## Validation

The anchor already exists — this needs an analytic twin of the #1133 model
(`nonmem_anchor/reset_init_snapshot_fit.ferx`) compared against the same
`reset_init_snapshot_{A,B,C,D}.tab` numbers, plus an extension of
`tests/analytical_init_equivalence.rs` to a reset-carrying subject so the analytic and ODE
forms are pinned equal. Arm B (flat `WT`) is the one that isolates *re-deposit vs drop*; arms
A and C then check the snapshot, exactly as they do on the ODE side.

Non-degeneracy note: a single-reset, no-dose dataset cannot see this — with nothing in the
compartment the drop and the re-deposit differ only by the baseline itself. The committed
dataset already carries a dose before the reset so the discarded residue and the re-deposited
baseline are different numbers (`9.32` vs `14.0`).

Not blocking #1133: the ODE path is the one with the analytic-gradient twin and the adaptive
engines, and it is now anchored. This is the remaining half.

TeunP:
## Why

#1073 established the rule ferx now follows on every other boundary: **NONMEM runs `$PK` at every
data record and ADVANs *to* it, so a segment is governed by the record that terminates it.** An
EVID=3/4 row is a data record. ferx did not treat it as one — the reset arm re-seeded the state
with `last_pk`, the *previous* record's snapshot — so an `[odes] init(state) = <covariate-driven
expr>` restarted the episode on a stale covariate.

This was found during #1073 and split out deliberately rather than half-fixed under cover of a
covariate-snapshot change. It is a real user-visible divergence, not a latent one: with
`init(central) = 10*WT` and `WT` doubling at the reset, every post-reset prediction was a factor
of two out.

## What changed

The anchor came first — nine arms in total, `A` through `J`. The first four control streams
(`nonmem_anchor/reset_init_snapshot_{A,B,C,D}.ctl`) — `ADVAN13 TOL=9`, one compartment, all
`$THETA FIX`, `$OMEGA 0 FIX`, `MAXEVAL=0`, `A_0(1) = THETA(3)*WT`, with `CL` and `V` as plain
thetas so **`WT` reaches the prediction only through the seed**. A 100 mg bolus at `t = 4` puts
residual drug in the compartment when the reset arrives at `t = 8`, so the reset's incoming side
is live (CLAUDE.md's non-degeneracy rule).

| arm | `WT` at the reset row | `WT` on the next record | IPRED at `t = 8` | `#OBJV` |
|---|---|---|---|---|
| A | 140 | 140 | 28.000000 | 115.73262077584336 |
| B | 70 (flat) | 70 | 14.000000 | 56.990279701081718 |
| C | 140 | **200** | 28.000000 | 115.73262077584336 |
| D | 140 (EVID=4, +100 mg) | 200 | 30.000000 | 128.80319627590555 |

Three facts, each readable straight off the tables:

1. **NONMEM re-applies `A_0` at a reset.** Arm B seeds `700/50 = 14.0` where the decayed residue
   was `9.32` — a reset is not "zero and continue".
2. **It uses the reset row's own `$PK`.** Arm A reads `28.0 = (10·140)/50`; the previous record's
   `WT = 70` would have given `14.0`, which is what ferx produced.
3. **Not the next record ahead, either.** Arm C is bit-identical to A at all ten timepoints and to
   the last digit of `#OBJV`, despite `WT = 200` on every post-reset record.

Fact 3 is the one that shaped the fix. Reusing #1073's `governing_record` resolution — the obvious
move, since the reset is a record — would have passed arms A, B and D and failed only C. I
mutation-tested exactly that wrong fix; see **Tests** below.

### Five more arms

Review found each of the four above degenerate in some direction, and one of the new arms
overturned a design decision I had made without measuring. `.tab` **and** `.ext` for all nine are
committed under `nonmem_anchor/results/`, so every constant below is auditable against the file it
came from, without a NONMEM licence.

| arm | what it isolates | NONMEM reads | `#OBJV` |
|---|---|---|---|
| E | **two** resets (`WT` 140 at `t = 8`, 200 at `t = 12`) — the per-reset *index*, which A–D cannot see with one reset each | `28.0`, then `40.0` | 234.84197355103731 |
| F | `ETA(1)` inside `A_0`, with `$OMEGA 0.09` and `POSTHOC` — the objective and the gradient, not only the prediction | — | 71.223755303866326 |
| G | a standalone `EVID=3` at `t = 8` carrying `WT = 140` **and a separate** `EVID=1` at the same time carrying `WT = 90`. D cannot separate these: its reset and dose are one `EVID=4` row, so `reset_cov(r)` and `dose_cov(k)` are byte-identical maps and seeding from either passes | `28.0` at the reset, `30.0` once the dose lands — the dose row's snapshot would give `18.0`/`20.0` | 128.80319627590555 |
| H | the reset as the subject's **first** event, where no "the record before" rule has anything to resolve. `WT = 140` on the reset, `70` on the co-timed observation | `28.0` | 135.12319830791392 |
| J | `WT` **flat** — only the `OCC` column moves. The reset row carries `OCC = 2` between `OCC = 1` records, and `A_0` reads a THETA-selected occasion multiplier rather than an ETA, so the answer is deterministic and needs no EBE | `42.0` = `(10·70·3)/50`, occasion **2** — the reset row's own. The preceding record's occasion 1 gives `14.0` | 135.11165705456989 |

The reset row's covariates were not stored at all, so the change is:

* `Subject::reset_covariates`, populated by the reader (LOCF state is already refreshed at the top
  of the row loop, so it is the reset row's own snapshot) and riding the same sort permutation as
  `reset_times` — sorting the times alone would silently pair each reset with another reset's row;
* `EventPkParams::reset`, filled by `compute_event_pk_params_into`, `compute_event_pk_params_iov`
  and `predict_iov`;
* read in **all four** affected engines:
  * `ode_predictions_event_driven` — the dense predictor;
  * `sens/ode_provider.rs` — the `Dual2` twin. `init(...)` + reset has been **analytic** since #486
    (the "keeps it on FD" comment at the `integrate_tvcov_g` call site is stale and is corrected
    here), so leaving it on `last_params` would have left the FOCE/FOCEI gradient differentiating a
    different function than the one predicted;
  * `ode_predictions_adaptive_impl` and `adaptive_frozen_replay_tv` — both re-seeded from the
    decision-time LOCF carry. `pk_readout` stays LOCF: that is a statement about what a controller
    may read, not about the state a reset restores.

The `Kind::Reset => last_pk` arm in the segment-parameter match stays, with its comment corrected
to say why: the segment ending at a reset is discarded (the re-seed overwrites `u` before any
readout), so that value never leaves the loop. Keeping the arm explicit stops the reset falling
into the `_` branch and silently taking the next record ahead.

**The occasion rule changed once arm J was measured.** The first push evaluated a reset row's
snapshot under the *preceding record's* occasion — chosen to avoid regressing κ, and written into
the source as "tracked separately", which is to say decided against no evidence. Arm J is that
evidence, and it says the reset row's own `OCC` governs: `42.0` against `14.0`, a factor of three.
So `Subject::reset_occasions` now exists, parallel to `reset_times` and riding the same
permutation — the exact analogue of `reset_covariates`. The neighbour scan survives only as the
fallback for a `Subject` assembled in memory (a fixture, FREM, the R glue) that carries no `OCC`
at all.

## Alternatives considered

**Reuse `governing_record` (the next record ahead) instead of storing the reset row's covariates.**
Zero plumbing — no `Subject` field, no `EventPkParams` slot. Rejected on measurement: arm C shows
NONMEM does not do this, and on the natural LOCF dataset (arm A) the two agree, so it would have
shipped green and been wrong on any dataset where a covariate moves again immediately after a
reset.

**Push EVID=3/4 rows into `pk_only_times` so they arrive as ordinary records.** Cheaper, but
`kind_order` sorts `PkOnly` *after* `Reset`, so the re-seed would still have run on the old
snapshot — and it changes the segment structure for every reset-carrying dataset.

## Cross-repo dependency

Checked against `origin/main`: `ferx-r`'s Rust glue never constructs a `ferx_core::Subject`
literal (it only builds `SubjectResult`), so the new field is source-compatible there.

## Breaking changes
- [ ] None

<details>

Three public items grew:

* `Subject` gains a `reset_covariates: Vec<HashMap<String, f64>>` field. Any exhaustive struct
  literal must add it (`Vec::new()` is correct for a subject built in memory with constant
  covariates); `..Default::default()` callers are unaffected. Read it through the new
  `Subject::reset_cov(r)` accessor, which falls back to the subject-static map.
* `EventPkParams` gains a `reset: Vec<PkParams>` field, on the same terms.
* `ode_predictions_event_driven` and `ode_predictions_event_driven_with_states` take an eighth
  argument, `pk_at_reset: &[PkParams]`, asserted to be parallel to `subject.reset_times`.

</details>

## Numerical validation
- [x] Compared against: **NONMEM 7.6.0**, four control streams committed under `nonmem_anchor/`
- [x] Reference values:

```
# before (ferx, reset_init_snapshot.csv — WT steps 70 -> 140 at the reset row)
t=9   12.6677238525      <- seeded from the PREVIOUS record (WT=70)

# after
t=9   25.3354476910      <- NONMEM: 2.5335447691E+01
```

Every observation row of all four datasets is asserted against the NONMEM `IPRED` table to
`1e-7` relative in `tests/reset_init_snapshot_nonmem_anchor.rs` (Tier 2 — `predict` at fixed
parameters, no convergence loop, so it runs on every PR; #1132 records that the slow tier does
not).

## Performance

One extra `pk_param_fn` evaluation per reset row per subject, on datasets that have resets. The
IOV static seeding branch in `seed_iov_events` now shares one occasion-less `Dual2` seed between
the EVID=2 and reset vectors rather than evaluating it twice.

## Tests

Every new test was mutation-checked — the measured failure is quoted in each case.

| test | tier | kills |
|---|---|---|
| `reset_init_snapshot_nonmem_anchor.rs` (10 tests) | 2 | the fix reverted (`12.67` vs `25.34`, rel 5.0e-1) **and** the `governing_record` wrong fix (`36.19` vs `25.34`, arm C + the invariance test only) |
| `ode_provider_init_reset_midtimeline_reads_the_reset_rows_snapshot` | 1 | the `Dual2` twin left on `last_params` (`19.89` vs `44.76` = 2.25× = 90/40) |
| `adaptive_reset_reseeds_init_from_the_reset_rows_covariates` | 1 | the driver reverted (`548.81` vs `219.52`) **and**, separately, the frozen replay reverted (same Δ, roles swapped) |
| `iov_event_pk_seeds_each_reset_from_its_own_row_and_window` | 1 | wrong covariate row (`1.0` vs `0.5`) and baseline-κ (`0.8` vs `1.6`) |
| `test_reset_rows_captured_with_their_own_covariates` / `..._skipped_when_no_tv_covariates` | 1 | reader pairing and the no-TV fallback |

**On the oracles that could not see this.** Both existing sens twins for the reset arm turned out
to be degenerate, each for its own reason, and it is worth recording because each looks like it
should have caught it:

* `ode_provider_tvcov_init_reset_matches_production` is TV-cov with a mid-timeline reset, but its
  model is `V = TVV * exp(ETA_V)` — `WT` sits on `CL` only, so `init = BASE/V` is covariate-free
  and every candidate snapshot seeds the same number.
* `ode_provider_init_reset_dose_at_zero_matches_production` *does* put `WT` on `V`, but its reset
  is the first timeline event, where the reset row, `init_pk` and the dose@0 snapshot are all
  `WT = 60`.

Same on the adaptive side: `adaptive_reset_matches_static_predict` still passes with the driver's
fix reverted. The new tests break both degeneracies at once.

And the reason this needed NONMEM rather than an in-repo twin: `Dual2`-vs-FD cannot see it (the
twin mirrors the production arm deliberately, so FD moves both sides together — CLAUDE.md's
documented failure mode, and how #1079 survived), and the closed-form walk zeroes unconditionally
with no `init(...)` concept, so there is no second implementation to disagree.

- [x] Not applicable (behaviour fix, no new API surface)

## Docs
- [x] `docs/` updated for user-visible changes — `docs/model-file/covariates.qmd` gains a callout
      stating the rule, and a **stale claim is corrected there**: that page said an ODE model
      combining `init(...)` with an EVID 3/4 reset falls back to the finite-difference gradient.
      It has been analytic since #486.

## Checklist
- [x] `cargo clippy` clean (the four pre-existing `approx_constant` hits in
      `src/sens/two_cpt_explicit.rs` — PK values that happen to be `3.14` — are untouched)

## Reviewer hints

Read arm C first (`nonmem_anchor/reset_init_snapshot_split.csv` and the module header of
`tests/reset_init_snapshot_nonmem_anchor.rs`). It is the only dataset that separates "the reset
row's own snapshot" from "the next record ahead", and it is why this fix stores covariates instead
of reusing the resolution #1073 already has.

The subtle part is `sens/ode_provider.rs`: the twin has to move with production or the FOCEI
gradient differentiates a different function, and the parity harness that would normally catch that
cannot, because FD perturbs the twin's own value path. `check_vs_production` is what bites — it FDs
the *production* predictor — and beneath it the NONMEM anchor.

The rest is mechanical: ~180 `Subject` literals across the test tree gained one field.

## Open questions

The question that stood here — whether to read `OCC` on reset rows — is answered. Arm J measured
it, and the fix ships it.

What is left is narrower, and I would rather it were ruled on than filed unilaterally.
`reset_row_occasion`'s fallback — reached only when `reset_occasions` is empty — documents itself
as mirroring the walk's `init_pk`, and does not quite: `earliest_record_pk` scans observations and
`EVID=2` rows only, and lets the observation take a tie, while the fallback also scans doses,
first, so a dose takes it. A first-event reset with a dose in an occasion carrying no observations
would therefore seed under the dose's κ where the walk's own `init_pk` used the observation's —
two different κ for one seeding operation on one subject.

It is **unreachable from any data file**: the reader gates `occasions`, `dose_occasions` and
`reset_occasions` on one `occ_col.is_some()`, so with no `OCC` column every candidate is `None`
and the answer is κ = 0 either way. It needs a hand-built `Subject` that populates `occasions` but
not `reset_occasions`. Its own issue, or a fourth round on this branch?

Closes #1133. Filed, not fixed: #1135 — `add_analytical_init_with` drops the analytical
`[initial_conditions]` baseline at the first reset and never re-deposits it, so after this PR the
analytic and ODE transcriptions of the same model disagree on any reset-carrying dataset. Arm B is
already its anchor. Also filed: #1148 — an adaptive controller's decision-time covariate LOCF
still ignores reset rows. The one-line fix for it here split `locf_decision_cov` from
`segment_pk_at`, the mirror its own docstring names, handing a controller post-reset covariates
against pre-reset PK; it was backed out. The real fix needs resets in `AdaptiveRecordIndex`'s `at`
lookup but *not* its `governing` one, which must stay aligned with the dense engine's `is_record`.

## Public API (A6)

`api/ferx-core-public-api.txt` is regenerated with `tools/update-public-api.sh` — the same script
CI verifies with. Three added items and two changed signatures:

* **`Subject::reset_covariates`, `Subject::reset_occasions`** — every field of `Subject` is `pub`;
  it is the type a consumer assembles when building a population without going through the CSV
  reader. These are two more per-record parallel vectors alongside the existing
  `dose_covariates` / `obs_covariates` / `pk_only_covariates` and `dose_occasions` / `occasions`.
* **`Subject::reset_cov(r)`** — the reader for the first of those, mirroring the already-public
  `dose_cov` / `obs_cov` / `pk_only_cov`.
* **`EventPkParams::reset`** — the per-reset PK snapshot, alongside `dose` / `obs` / `pk_only`.
* **Breaking:** `ode_predictions_event_driven` and `ode_predictions_event_driven_with_states` take
  an eighth argument, `pk_at_reset: &[PkParams]`. A reset's seed cannot be derived from the other
  three arrays — that is the whole of #1133 — so it has to be passed.

Checked `ferx-r@origin/main` for the breaking pair: it calls neither function, references neither
`EventPkParams` nor `reset_times`, and builds no `Subject` literal, so no companion PR is needed.

## Verification

`origin/main` is merged in (the #1114 workspace split, among others) — clean, no conflicts.

| gate | result |
|---|---|
| `cargo test --features ci,survival --no-fail-fast` | 159 binaries, **4594 passed, 0 failed**, 166 ignored (the `slow-tests` tier) |
| `cargo test -p ferx-tools -p ferx-cli` | **158 passed, 0 failed** |
| `cargo check` × 5 feature sets (the `Check` job's own steps) | clean |
| `cargo clippy --features ci,markov,nn --all-targets` | no errors (local nightly `2026-08-29`; CI installs fresh) |
| `cargo fmt --all --check` | clean |
| `tools/update-public-api.sh --check` | matches the committed baseline |

The root-package count fell 4634 → 4594 across the merge. That is entirely #1114 moving the CLI
out of the root package — `src/bin/run_model.rs` (38 tests) and 11 of `tests/cli_binaries.rs` now
live in `crates/ferx-cli`, against `tests/prepare_run.rs` (4) and `tests/public_api_boundary.rs` (5)
arriving. 4634 − 38 − 11 + 4 + 5 = 4594, and the moved tests are covered by the member run above.


TeunP:
## Self-review at xhigh — 15 findings, all fixed

Ten independent finder angles over the diff plus a sweep, each candidate verified before it survived. The headline: **the PR as first pushed introduced two regressions of its own and shipped a core fix that no Tier-1 test could see.** Details below, most severe first.

### The fix traded a covariate error for a κ error

`predict_iov` seeded the reset row with `pk_only_combined` — κ = 0, the EVID=2 convention. But the code it replaced was `last_pk`, the *previous record's* `PkParams`, which carried **that occasion's κ**. So on an IOV model with κ on a parameter the seed reads, this PR moved the whole post-reset trajectory by `exp(κ)` — a factor of 2 at κ = ln 2, the same magnitude the anchor calls a bug on the covariate axis. It also disagreed with `compute_event_pk_params_iov` in the same commit, which used the decision window's κ: `simulate_adaptive()` and `predict()` would have returned different answers for one subject.

I had written that κ = 0 up as "a separate gap". It wasn't a gap, it was a regression, and the comment saying otherwise was mine.

Fixed with a shared `pk::reset_row_occasion` — the occasion of the last record strictly before the reset, which is exactly what `last_pk` held there — used by `predict_iov` and by both branches of the twin's `seed_iov_events`. Reading `OCC` on reset rows would make this exact rather than no-regression; that stays a follow-up.

### Nothing in Tier 1 could see the core fix

All five modified reset tests passed a uniform `vec![pk; n]` as `pk_at_reset`, making it bit-identical to `last_pk`. **Reverting `initial_state(&pk_at_reset[idx].values)` to `initial_state(&last_pk.values)` — the #1133 defect itself — passed the entire `--lib` suite.** Only the Tier-2 NONMEM anchor objected.

And every dataset in the repo had exactly one reset, so `pk_at_reset[idx]` was only ever `pk_at_reset[0]`: swapping one for the other also passed everything, including all four anchor arms.

### Two new anchor arms

| arm | what it pins | `#OBJV` |
|---|---|---|
| **E** `reset_init_snapshot_multi.csv` | two resets (WT 140 at t=8, 200 at t=12) — the per-reset **index** | 234.84197355103731 |
| **F** `reset_init_snapshot_F.ctl` | `ETA(1)` inside `A_0` under `$OMEGA 0.09`, `POSTHOC` — the **objective**, hence `∂init/∂η` | 71.223755303866326 |

F is the one that matters for the sensitivity path. Arms A–E are `$OMEGA 0 FIX` with an η-free `A_0`, so `∂init/∂η` is identically zero and a defect confined to the `Dual2` reset seed was caught by nothing outside this repo — and CLAUDE.md is explicit that a `Dual2`-vs-FD parity is an oracle for the gradient only. With an η in the seed, the FOCEI `h` matrix (the analytic Jacobian) moves the EBE and therefore `#OBJV`. NONMEM's own table shows the EBE genuinely moving: `IPRED(0) = 10.328` against `PRED(0) = 14.0`.

Both mutations now die: `[idx]` → `[0]` fails arm E (25.335 vs 36.193 at t=13) and the new Tier-1 `ode_reset_reseeds_each_reset_from_its_own_snapshot`.

Raw NONMEM output for all six arms is now committed under `nonmem_anchor/results/` — the convention `nonmem_anchor/.gitignore` states outright and that #1121 and #1073 both followed. Without it the 24 hard-coded constants were unauditable by anyone without a licence, which is the failure mode that produced the mirrored mrgsolve anchors in #1073.

### Four places the new state was not plumbed through

Each one silently reverts the fix, or splits two engines, on a dataset shape that reaches it:

- **`prune_irrelevant_tv_covariates`** neither scanned nor cleared `reset_covariates`. A covariate varying only on a reset row scanned as "no relevant TV", cleared the other three, and dropped the subject to the constant branch — the pre-#1133 answer. Worse, `reset_covariates` stayed populated, so `predict_iov` and the sensitivity seeders read the per-row map while FOCEI read the baseline: same subject, two OFVs.
- **`iov_walk_per_event`** (the twin of `subject_needs_per_event_pk`) and **`fit.rs`'s `needs_prune` gate** had the matching omissions.
- **`api/levels.rs`** never wrote the synthesized `__level_*` column into `reset_covariates`, so the reset seed read it as `0.0`. Newly introduced: pre-#1133 the reset read `last_pk`, built from obs/dose snapshots, which carried it.
- **`time_varying_covariate_names`** and **`static_nn_input_map`** both skipped the fourth vector.

### Two silent-failure paths and a same-time aliasing split

- Both adaptive engines used `ev.reset.get(r)` with a `None` fallback that **restored the pre-#1133 LOCF carry**, and the driver and its own frozen-replay verifier fell back to *different* quantities. Now both assert the length and index, matching `ode_predictions_event_driven`. (`.filter(|_| tv)` was a tautology — `tv = event_pk.is_some()` — and is gone.)
- **`check_event_pk_records`** — the #748 guard that exists *specifically* for snapshots the driver and the replay both read verbatim — never checked `event_pk.reset`. Its own docstring describes the scenario it was missing.
- `.any(...)` → `.position(...)` made the index load-bearing, and `position` takes the **first** match while the dense engine's last reset wins. Now `.rposition(...)`, which reproduces the dense semantics by construction.
- **`locf_decision_cov`** did not treat reset rows as a covariate source, so an adaptive decision after a reset read pre-reset covariates — the same defect family, one boundary over. The #748 verifier re-derives `decision_pk[g]` through that same function by design, so it mirrored the omission.

### Tests that passed for the wrong reason

- The IOV × reset twin ran with κ ≡ 0 at every occasion — blind to exactly the κ convention above. Now two distinct non-zero κ.
- Both sens reset twins asserted only twin-vs-production, which cannot see the two engines agreeing on a wrong snapshot. Both now also perturb *only* `reset_covariates` and require the prediction to move — arithmetic-free, and it catches the `reset_cov` fallback silently returning the subject-static map on both sides.
- `ODE_TV_INIT` declared no `[fit_options]`, so a `1e-3` assertion ran against a solver at `ode_reltol = 1e-4`. Tightened.

### Docs

`docs/model-file/initial-conditions.qmd` — the page a user reading about `init(...)` actually lands on — still said "System resets wipe the baseline… nothing re-deposits it", which arm B disproves, and contradicted itself twenty lines later. Now split by engine: `[odes]` re-deposits at the reset row's own covariates (anchored); analytical `[initial_conditions]` does not (#1135).

### Left as-is, deliberately

`apply_segment_boundary` re-seeds a reset from a single frozen snapshot, and I checked every caller: `ode_cumhaz_hazard` (the TTE likelihood), the `[derived]` grid path and `markov/endpoint` are **all frozen walks end to end**, so every candidate convention gives the same answer there. The one live gap is `ode_predictions_event_driven_with_states`' second pass on a TV subject, which uses the first observation's snapshot throughout — already inside `W_DERIVED_CMT_TV_ODE`'s scope, now noted explicitly at that warning. Calling this a fifth engine would overstate it.

The five pre-existing `clippy::approx_constant` errors are in `sens/two_cpt_explicit.rs` and `stats/special.rs`, untouched here — PK values that happen to be `3.14`.

TeunP:
## Phase-3 sweep + three more anchor arms — and one measurement that overturned a design decision

The xhigh review above ran ten finder angles but not the sweep pass (a fresh reviewer against the verified list, hunting only for gaps). That pass has now run. It found four more defects, **three of them in the commit that fixed the first fifteen**. Separately, building an anchor for a convention I had chosen *without* measuring showed the convention was wrong.

### NONMEM overturned the κ rule

The previous commit resolved an EVID=3/4 row's occasion from the **preceding record**. I picked that to avoid regressing κ and wrote, in the source, that reading `OCC` on reset rows "would make this exact rather than no-regression; that is tracked separately." That was a decision made against no evidence, so I built the evidence.

**Arm J** (`nonmem_anchor/reset_init_snapshot_J.ctl`) isolates the question. `WT` is flat, so nothing but the `OCC` column moves across the reset; `A_0` reads a THETA-selected occasion multiplier rather than an ETA, so the answer is deterministic and needs no EBE.

```text
   t     OCC     IPRED
 6.0      1      9.320824
 8.0      2     42.000000     <- (10·70·3)/50, occasion 2 — the reset row's OWN
 9.0      2     38.003172
```

`14.0` is what the preceding record's occasion 1 would give. A factor of three.

So `Subject::reset_occasions` now exists — parallel to `reset_times`, riding the same permutation, the exact analogue of `reset_covariates`. The neighbour scan survives only as the fallback for a `Subject` assembled in memory (FREM, the R glue, a fixture) that carries no `OCC` at all.

`reset_row_occasion` also now takes the reset's **index** rather than its time. Every caller iterates `0..reset_times.len()` and already has it; re-deriving it by a tolerance search would alias two resets inside that tolerance — the same class the previous commit fixed with `.rposition` in the adaptive engines. I found that one re-reading my own fix, not from the sweep.

### Two more degeneracies in the anchor set

**G** — arm D cannot separate "the reset row's snapshot" from "the co-timed dose's": its reset and dose are one EVID=4 row, so `reset_cov(r)` and `dose_cov(k)` are byte-identical maps and seeding from either passes. G splits them into a standalone EVID=3 at `t = 8` carrying `WT = 140` and a separate EVID=1 at the same time carrying `WT = 90`. NONMEM seeds `28.0`, then the dose lands for `30.0`; the dose row's snapshot would give `18.0`/`20.0`.

**H** — the reset as the subject's *first* event, where no "the record before" rule has anything to resolve. `WT = 140` on the reset, `70` on the co-timed observation: NONMEM reads `28.0`.

Nine arms now (A–H, J), with **`.tab` and `.ext` both** committed under `nonmem_anchor/results/` — the `IPRED` constants and the `#OBJV` values are each auditable against the file they came from, without a NONMEM licence.

One limit I wrote into the test rather than papering over: arm J's ferx twin reaches `OCC` as a **covariate**, because κ is estimated and a model file cannot fix it to a discriminating value. The occasion-**label** route is pinned by `pk::tests::iov_reset_takes_its_own_rows_occasion`, which drives `predict_iov` with explicit κ and asserts the 3× ratio. J anchors the rule against NONMEM; that test anchors ferx's second route to it.

### The sweep's four

- **`reset_row_occasion` never scanned `pk_only_times`.** On a TV-covariate dataset — exactly #1133's target class, where EVID=2 rows exist by construction — a reset whose preceding record was an EVID=2 row reached *past* it to an older record's occasion, reintroducing the κ regression that commit was written to prevent. EVID=2 rows now report κ = 0, which is what `last_pk` held there.
- **The same function broke exact-time ties toward the dose**, where the walk's `kind_order` (`DoseRecord < Obs`) leaves the *observation* in force. Now visited in timeline order.
- **`ode_tvcov_supported` never got the `reset_covariates` clause** its two siblings did, so a subject carrying only reset snapshots took the *static* dual walk while production took the per-event branch — a wrong analytic gradient, not an FD fallback, and invisible to `Dual2`-vs-FD because FD perturbs the twin's own value path.
- **`locf_decision_cov` — I had fixed half of a mirrored pair.** The previous commit made it reset-aware, which split it from `segment_pk_at`, the mirror its own docstring names ("share **one** covariate rule"). That hands a controller post-reset covariates against pre-reset PK parameters, and `verify_adaptive_snapshots` cannot catch it because it re-derives `dcov` through that very helper. **Backed out.** The real fix needs resets in `AdaptiveRecordIndex`'s `at` lookup but *not* its `governing` one — which must stay aligned with the dense engine's `is_record`, where a reset is deliberately excluded. Two halves of one index disagreeing on purpose is wider than the init-seed fix, so it is filed as #1148 with the reasoning, not smuggled in here.

### One more surviving mutation, closed

Anchor E pins the per-reset index on the *value* engine, but it runs through `predict` and never through `ode_subject_sensitivities` — so the twin's own `tl.push((rt, K_RESET, r))` and `pk_at_reset[idx]` were still only ever evaluated at index 0. Reverting either to a constant passed the entire unit suite. `ode_provider_reset_seeds_are_indexed_per_reset` closes it.

### Verification at this head

Full local suite at `540f98a1` (`--no-default-features --features ci,survival --no-fail-fast`, every target): **158 binaries, 4634 passed, 0 failed, 166 ignored** — the ignored are the `slow-tests`-gated convergence fits, which never run on a PR. Baseline at `94a11253` was 4629 passed; the +5 are the new tests above. `cargo fmt --all --check` clean.

Issue #293, opened 2026-06-13
Add a survival (and nn) coverage flag so feature-gated code is measured


TeunP:
## Problem

The coverage build measures the **FD `ci` surface only** — `coverage-pr` runs `--features ci`, the nightly `coverage` runs `--features ci,slow-tests`. Everything behind `#[cfg(feature = "survival")]` (TTE / survival / categorical endpoints) and `#[cfg(feature = "nn")]` (DCM / low-dim NODE) — including the survival-only blocks scattered through `datareader.rs`, `api.rs`, and `model_parser.rs` — is **not compiled** in that build, so Codecov reports it as **missed** even though the `survival` CI job actually exercises it.

Net effect: the reported number **undercounts** the true `ci+survival` surface, and it manufactures phantom "gaps" that no test can close (the coverage build can't reach them). This is the same root cause as `ad/dual.rs`, which #287 gated out of the report.

## Evidence

- While doing #286 backfill (#292), `datareader.rs`'s single biggest "gap" (lines ~1101–1195, 95 lines) turned out to be a `#[cfg(feature = "survival")]` TTE-routing block — uncoverable in the `ci` build, exercised by the `survival` job.
- `model_parser.rs`'s 607 "missed" lines include survival/nn validation branches that are likewise compiled out.
- Overall reported coverage is ~90.55% (no-LTO); the true `ci+survival` figure is higher.

## Proposal

Add a **`survival` coverage flag**:

- A coverage job (or extend an existing one) running
  `cargo llvm-cov --tests --no-default-features --features ci,survival --profile ci-test --lcov --output-path lcov.info`,
  uploaded to Codecov with `flags: survival`.
- Codecov then merges `fast` + `slow` + `survival` flags, so feature-gated lines are counted where they're exercised and the project number reflects the union.
- Use the `ci-test` profile (no fat LTO, from #290) to keep it cheap.

Optionally do the same for `nn` (a separate `nn` flag, or fold into the same job with `--features ci,survival,nn`).

## Considerations / open questions

- **Per-PR vs nightly.** The `survival` integration tests are Tier-2 (fast — see the `survival` job comment in `ci.yml`), so a per-PR `survival` flag is feasible and would make the PR-time number accurate. It adds one instrumented job; the `ci-test` profile keeps it within the PR envelope. Alternatively start nightly + carryforward, promote later.
[...]

PR #1138, opened 2026-08-30
refactor(api): split into ferx-core / ferx-tools / ferx-cli workspace, with a CI-enforced public API boundary [closes #1114]

### LINKED ISSUE #1114: refactor(workspace): split into ferx-core / ferx-tools / ferx-cli, with a CI-enforced public API boundary

roninsightrx:
## Why

Model-development tooling — bootstrap, stepwise covariate modelling, model
search, cross-validation — is orchestration code: loops over `fit()`,
non-generic, cheap to compile. `ferx-core` is the opposite: LLVM-bound, 93% LLVM
/ 1% frontend, with `sens` alone ~90% of IR (#971). Written in-crate, every
tools edit pays the full lib compile plus relink. As a downstream crate it links
the cached `ferx-core` rlib instead.

Second motive, and the one this PR actually spends its effort on: `ferx-tools`
can only touch `pub` items, so tooling goes through the same door ferx-r uses —
which keeps the R layer thin *by construction*. The failure mode to guard
against is not `ferx-tools` reaching into internals (cargo makes that
impossible); it is **`ferx-core` growing an ad-hoc `pub` in the same PR that
needs it**, which turns the boundary into paperwork.

This is the opposite case to #970's rejected split, which moved the hot generic
core out and would have duplicated monomorphization across a crate boundary.
This one stacks cold code on top and moves nothing.

## What changed

```
ferx-core/                       # repo root = ferx-core package + workspace root
├─ Cargo.toml                    # [package] ferx-core + [workspace] members
├─ api/ferx-core-public-api.txt  # public-API baseline, diffed in CI
└─ crates/
   ├─ ferx-tools/                # placeholder; depends on ferx-core
   └─ ferx-cli/                  # the moved binary; [[bin]] name = "ferx"
```

**The root stays the `ferx-core` package.** `ferx-r/src/rust/.cargo/config.toml`
patches `ferx-core = { path = "../../../ferx-core" }` — the repo *root*. Moving
the package under `crates/` would break every local ferx-r build, and break it
*silently*: the patch stops applying and ferx-r quietly builds `main` instead of
your branch.

**The `ferx` binary had to move.** `ferx-tools` depends on `ferx-core`, so a
binary in `ferx-core` that will call `bootstrap` would be a cycle. Installed
binary name and behaviour are unchanged.

**A2 — the enforcement.** `api/ferx-core-public-api.txt` is a committed
`cargo public-api` snapshot; the new `Public API baseline` job regenerates and
diffs it via `tools/update-public-api.sh --check`, so any widening fails until
the baseline is updated in the same PR. Doubles as semver protection for the
crates.io release and for ferx-r.

## Alternatives considered

**Separate repo.** Real advantages — never pays core's compile-bound CI (#969),
decouples release cadence, keeps tools' deps out of the `Cargo.lock` ferx-r
pins. The decisive cost is choreography: every core API change becomes core PR →
tools PR → ferx-r lock bump, and two API gaps (G1, G2) are already visible with
more to surface. Extraction later is cheap (`git filter-repo`, flip one `path`
dep to `git`); merging two repos back is not. The one thing a workspace
forfeits — *felt* API pressure — is bought back mechanically by A2.

**Five deviations from the issue's sketch**, each for a reason measured in this
PR rather than assumed:

1. **Baseline uses `ci,markov,nn`, not default features.** Those three cfgs each
   add public items, so a default-features baseline would let a `pub` added
   behind one of them through the gate unseen.

2. **`-sss`, not `-s`.** At `-s` the surface is 9190 lines of which ~3.6k are
   `impl core::marker::Freeze for …`, and adding one `pub struct` moves a dozen
   lines. A2 exists so a new public item is *one reviewable line*; `-sss` gives
   exactly that (4280 items).

3. **`#[doc(hidden)]` is banned by a test, not by the baseline.** Verified
   directly: adding `#[doc(hidden)] pub fn probe()` to `src/lib.rs` leaves
   `--check` **green**; the same `pub fn` without the attribute fails it with
   exactly one `+` line. `cargo public-api` omits hidden items, so A3 is a
   *working bypass* of the gate and cannot be enforced by the snapshot at all.
   `tests/public_api_boundary.rs` closes it, and also pins A1 — a
   `dev-dependency` on `ferx-cli` would compile fine, so cargo's cycle check is
   not sufficient there either.

4. **`coverage-pr` becomes `--workspace`, rather than a standalone member test
   job.** Verified — not assumed — that `cargo llvm-cov` does **not** go
   workspace-wide on its own in a root-package workspace: an unqualified run
   emits 102 `SF:` records, none under `crates/`. So root-scoping would have
   dropped the CLI's ~960 previously-measured lines from the `fast` flag,
   lowering project coverage for a pure file move *and* failing a later
   `crates/`-only PR at 0% on its own patch gate — the same trap #293 documents
   for the feature-gated surface. One instrumented build now runs *and* measures
   the members: no third cache key, no second `ferx-core` compile. This is the
   one existing job that changed; its display `name` is untouched, so branch
   protection is unaffected.

5. **`release.yml` needs no repoint.** It builds nothing — it only generates
   GitHub release notes from a tag.

**Feature flags belong to `ferx-core`**, so member-scoped and workspace-wide
commands write them package-qualified (`--features ferx-core/ci`); a bare
`--features ci` fails with *"none of the selected packages contains this
feature"*. Qualifying them also keeps the members linking the rlib the previous
CI step just built instead of forcing a second compile under a different feature
hash — which is what makes the two new `Check` / `Clippy` steps ~4s each.

**`resolver = "2"` preserves behaviour rather than changing it.** A standalone
edition-2021 package gets v2 implicitly, but inside a workspace the root
package's edition no longer implies it — cargo would warn and fall back to v1,
quietly changing how dev-/build-dependency features unify into the `ferx-core`
build.

## Cross-repo dependency

ferx-r is unaffected: the root package is still `ferx-core` at the same path, no
`pub` item changed, and the git dependency resolves by package name anywhere in
the repo. Verified by hand below. ferx-r can later add `ferx-tools` from the same
URL and rev — one extra `Cargo.lock` entry, **no new pin to bump**.

## Breaking changes
- [x] None *(for library and R consumers)*

One **build-command** change for source builds, no API or runtime change:

<details>

The root package is now the `ferx-core` **library** and carries no binary, so a
bare `cargo build --release` no longer produces `ferx`:

```bash
cargo build --release --workspace                      # or: -p ferx-cli
cargo run --release -p ferx-cli -- model.ferx --data data.csv
```

Workspace-wide commands need package-qualified features:
`--features ferx-core/ci`, not `--features ci`.

The installed binary is unchanged — same name, arguments, output. Consumers of
the `ferx-core` crate, ferx-r included, need no change.

</details>

## Numerical validation
- [x] Not applicable (workspace mechanics / CI only — no numerical code touched)

`crates/ferx-cli/src/main.rs` is a `git mv` of `src/bin/run_model.rs`; the only
edits are three test-fixture paths (`CARGO_MANIFEST_DIR` is now `crates/ferx-cli`)
and one clippy `needless_borrow` the new `--tests` lint step surfaced.

## Performance

CI cost is roughly flat: the two new `Check` / `Clippy` steps reuse the
`ferx-core` build the preceding step produced (~4s each locally), `coverage-pr`
gains the members inside its existing instrumented build, and the new
`Public API baseline` job is a rustdoc-JSON build (~10s locally after warm-up).
The payoff is future: `ferx-tools` edits will link the cached rlib instead of
recompiling 279k LOC.

## Tests
- [x] Unit test(s) added — `crates/ferx-tools` has one, routed through
      `ferx-core`'s *public* API so the placeholder still proves the one-way
      link it exists to establish
- [x] Regression test added that fails without this fix —
      `tests/public_api_boundary.rs` (`no_doc_hidden_escape_hatch` fails on the
      `#[doc(hidden)]` bypass the baseline cannot see;
      `ferx_core_does_not_depend_on_the_layers_above_it` pins A1)

`tests/cli_binaries.rs` was split: `CARGO_BIN_EXE_ferx` is only defined for test
targets of the package declaring the binary, so the `ferx` half moved to
`crates/ferx-cli/tests/cli_ferx.rs` and the `generate_data` half stayed.

**Acceptance evidence**

| Check | Result |
|---|---|
| Full workspace suite (`--workspace --tests --features ferx-core/ci`) | **159 test binaries, 0 failures**, exit 0 |
| Baseline job on an unmodified tree | green |
| Baseline job with a deliberate `pub fn` added | **fails**, `+pub fn ferx_core::__api_gate_probe()` — one line |
| Baseline job with `#[doc(hidden)] pub fn` added | **green** → hence the A3 test |
| `cargo build --release --workspace` | `target/release/ferx`, `check` and usage output identical |
| `cargo llvm-cov` scope in a root-package workspace | root-only: 102 `SF:`, 0 under `crates/` |
| `cargo package -p ferx-core` | `[workspace]` normalized away, `crates/` excluded, `api/` excluded |
| **ferx-r local patch (by hand — the failure is silent)** | resolves to this checkout (`source: None`); build emits **9 diagnostics identical** to the same build patched at `main` |

The ferx-r build fails on both sides for a **pre-existing** reason —
`kappa_weight_typical`, `kappa_weights`, `residual_correlations`,
`discrete_rows` drift — not from this split. Running the control against `main`
is what separates the two.

## Docs
- [x] `docs/` updated — `installation`, `cli`, `multistart`, `steady-state`,
      `absorption` for the `cargo run` path; new **SDLC §8.7 "The public-API
      gate"**; a *"If the `Public API baseline` job goes red"* section in
      `contributing.qmd`
- [x] New pages linked in `docs/_quarto.yml` — n/a, no new pages
- [x] `CLAUDE.md` / `AGENTS.md` gain a "Workspace layout and the public-API
      boundary" section; `PULL_REQUEST_TEMPLATE.md` and `.githooks/pre-commit`
      (`cargo fmt --all`) updated

## Checklist
- [x] `cargo clippy` clean (the one warning the new `--tests` step surfaced is fixed)
- [x] `Cargo.toml` version bumped if breaking — n/a

## Docs & examples
- [x] No new DSL syntax, `[fit_options]` key, example `.ferx`, or data file
- [x] Example execution: `cargo build --release --workspace` and a CLI smoke run
      exercised; no example output changes (the binary is the same code)

## Reviewer hints

Worth the attention:

- **`tools/update-public-api.sh`** — the pin comments are the substance. The
  script owns the dated nightly, the `cargo-public-api` version and the feature
  set, and CI *runs that same script*, so the generating and verifying commands
  cannot drift. The workflow greps the pinned date out of the script rather than
  repeating it.
- **`tests/public_api_boundary.rs`** — reads as style policing but is not; see
  deviation 3.
- **The `coverage-pr` diff** — the one existing job that changed (deviation 4).

Skimmable: the `cargo run --release -p ferx-cli` sweep across docs, and
`crates/ferx-cli/src/main.rs`, which is a pure move.

## Open questions

- **The pinned nightly is `nightly-2026-05-29`** — the toolchain this was
  generated on. `rust-toolchain.toml` says a bare `nightly`, which would make the
  baseline drift daily, so the job pins a date via `RUSTUP_TOOLCHAIN` (the same
  override the coverage jobs already use for `stable`). Bumping it is deliberate:
  change the date, rerun the script, commit the regenerated baseline. Happy to
  move the pin forward if you'd rather start from a recent nightly.
- **4280 public items** is itself a finding: the surface `ferx-tools` and ferx-r
  see is already very wide. A2 stops it growing silently but does not shrink it.
  Worth a follow-up pass?
- **`ferx-cli` is not published.** If `cargo install ferx-core` is ever meant to
  yield `ferx`, that becomes `cargo install ferx-cli` and the member needs
  publishing too — its `ferx-core` dep already carries a `version` alongside the
  `path` so that stays a one-line change.

Closes #1114.

https://claude.ai/code/session_01Jr9xeiwGejMj69y7cbt59V


roninsightrx:
## Codex review — blocking finding

The workspace split leaves several checked-in CLI invocations targeting the old root-package binary. In particular:

- `tools/vi-emvi-comparison/run.sh:74`
- `tools/vi-nuts-anchor/run.sh:42`
- the build instruction in `tools/vi-emvi-comparison/make-wide-residual-data.sh:29`

still use `cargo build --profile ci-fast --bin ferx`. At this PR head, that command fails immediately with:

```text
error: no bin target named `ferx` in default-run packages
help: available bin in `ferx-cli` package:
    ferx
```

Please update these executable validation harnesses to select `-p ferx-cli`. A repository-wide sweep is also needed for the shipped example headers and reference instructions that still use `cargo run ... -- <model>` or root-scoped `--bin ferx`; those commands now fail too. Feature-enabled examples should select `ferx-cli` and use package-qualified core features, e.g. `--features ferx-core/survival`.

Verification at `1cadc94a`: workspace formatting and compilation passed; `ferx-cli`, `ferx-tools`, CLI integration, and boundary tests passed; and `tools/update-public-api.sh --check` matched the committed baseline.

roninsightrx:
## Fixed in `022c1dc1` — and the finding was broader than the three files

Confirmed and fixed. The three executable harnesses now use `-p ferx-cli`:

| File | Now |
|---|---|
| `tools/vi-nuts-anchor/run.sh:42` | `cargo build --profile ci-fast -p ferx-cli` |
| `tools/vi-emvi-comparison/run.sh:74` | `cargo build --profile ci-fast -p ferx-cli` |
| `tools/vi-emvi-comparison/make-wide-residual-data.sh:29` | same, in the "build first" hint |

Verified that command lands the binary at `target/ci-fast/ferx` — the exact path
`make-wide-residual-data.sh` probes with `[ -x "$BIN" ]`.

The requested repo-wide sweep found **56 invocations across 52 files**, not just
the shipped example headers: `examples/*.ferx`, `nonmem_anchor/`, `plans/`,
`tests/reference/`, `examples/drafts/README.md`, and `SDLC.md`. Three shapes:

- `--bin ferx` → `-p ferx-cli`
- root-scoped `cargo run [--release] -- <model>` → `-p ferx-cli`
- feature-carrying runs also get the flag qualified, per your note —
  `--features survival|markov|nn` → `--features ferx-core/<f>`. Each verified
  to resolve (`cargo check -p ferx-cli --features ferx-core/<f>`, exit 0).

`--bin generate_data`, `--bin pktte_sim_anchor` and `--bin rtte_sim_anchor` are
left alone — those targets are still in `ferx-core`.

### The sweep is not the fix

Enumerating these by hand is what failed the first time, so the invariant is now
pinned in a test —
`checked_in_ferx_invocations_select_the_ferx_cli_package` in
`tests/public_api_boundary.rs`. It scans every `md`/`qmd`/`sh`/`ferx`/`R`/`rs`
file for a run/build that reaches the CLI without selecting its package, or a
`ferx-cli`-scoped command passing an unqualified core feature. (Needles are
built with `format!` so the file does not match its own doc comment — the trick
`ci_workflow_endpoint_coverage.rs` already uses.)

It earned its place immediately: on its first run it failed on **three
offenders that neither your list nor my sweep had enumerated** —
`examples/ctmm_2state.ferx`, `examples/ctmm_pd_2state.ferx`, and
`plans/dcm-and-low-dim-node.md:117`, all
`cargo run --release --features <f> -- …`, a shape absent from both hand-written
rule sets. That red-then-green is the regression evidence for this fix.

### Also caught: `SDLC.md`

Not in your list but broken by the same move, and the first pass missed it
entirely (I had updated `docs/development/sdlc.qmd`, which is a separate file).
Its build block said a bare `cargo build --release` → now `--workspace`, and its
CI table was stale on four counts: it now names the `Public API baseline` job,
`cargo fmt --all`, the `--workspace` coverage command, and the
package-qualified-features rule. `docs/cli.qmd`'s "Build the ferx binary" is
likewise `-p ferx-cli` now.

### Verification at `022c1dc1`

- `cargo build --profile ci-fast -p ferx-cli` → `target/ci-fast/ferx` ✅
- `cargo check -p ferx-cli --features ferx-core/{survival,nn,markov}` → exit 0 ✅
- `cargo fmt --all -- --check` clean; `tools/update-public-api.sh --check`
  matches the committed baseline ✅
- Full workspace suite: **159 test binaries, 0 failures** ✅

Thanks — the `tools/` scripts were the genuinely load-bearing miss; the rest
would have surfaced later as copy-paste failures with no obvious cause.

roninsightrx:
## Review — workspace split (`ferx-core` / `ferx-tools` / `ferx-cli`)

Reviewed the full diff, plus two empirical checks: `cargo check --unit-graph` for feature resolution, and a replay of the new path-scanning test over both this worktree and the main checkout.

The structure is right and the boundary guards are the good kind — they pin invariants the API baseline genuinely cannot see. Eight findings below, three of them substantive. **All eight are fixed in 9cd80019 on this branch**; each entry says what changed.

---

### 1. Members must set `default-features = false` — the "no second compile" claim is currently false

`crates/ferx-cli/Cargo.toml` and `crates/ferx-tools/Cargo.toml` depend on `ferx-core` without `default-features = false`. Verified with `--unit-graph`:

| step | resolved `ferx-core` features |
| --- | --- |
| `cargo check --tests --no-default-features --features ci` | `('ci',)` |
| `cargo check -p ferx-tools -p ferx-cli --no-default-features --features ferx-core/ci` | `('ci', 'default')` |

Different feature sets are different units, so the claim in `ci.yml` (and in CLAUDE.md and the PR body) that "the members link the already-built rlib instead of triggering a second `ferx-core` compile under a different feature hash" does not hold — both the `Check` and `Clippy` jobs rebuild all of `ferx-core` a second time.

The latent trap matters more than the minutes: `--no-default-features` is now silently defeated for `ferx-core` in every member-scoped and `--workspace` command, `coverage-pr` included. That is harmless only because `default = []` today. The manifest explicitly plans to promote `survival` (and via it `markov`) to default-on — the day that lands, `coverage-pr` measures a survival-enabled engine while its command line says `--no-default-features --features ferx-core/ci`.

**Fixed** — `default-features = false` on both member deps, with a comment saying why. Re-verified: the member step now resolves `('ci',)`.

### 2. `public-api` job silently uses an unpinned nightly, so its cache never hits

The job installs `nightly-2026-05-29` (read out of `tools/update-public-api.sh`, which is a nice touch), but every later step still sees the workflow-level `RUSTUP_TOOLCHAIN: nightly` from ci.yml:14. rustup auto-installs from that env override, so the job downloads a *second*, unpinned nightly and runs under it. `Swatinem/rust-cache` then derives the `ci-public-api` key from that toolchain's `rustc -vV` — which changes daily — so the key never repeats, the restore always misses, and the job does a cold rustdoc build of `ferx-core` on every run rather than the "~10s after warm-up" the PR estimates.

**Fixed** — the install step now appends `RUSTUP_TOOLCHAIN=$toolchain` to `$GITHUB_ENV`, before the cache step.

### 3. `scannable_files` walks the working tree, so the guard fails from the main checkout

`tests/public_api_boundary.rs` walks the real filesystem from `CARGO_MANIFEST_DIR`, skipping only `target`, `.git`, `_site`, `node_modules` — no dot-directory skip, no tracked-files filter. But `.gitignore` ignores `.claude/`, and this repo's documented worktree workflow parks sibling checkouts at `<repo>/.claude/worktrees/<branch>/`.

Replaying the exact scan logic from `/Users/ron/git/ferx/ferx-core` produces **4930 offenders**, dozens of them under `.claude/worktrees/<other-branch>/…`. After this merges, `cargo test` fails from the main checkout for any developer holding a pre-#1114 worktree — a failure they cannot fix by editing a tracked file.

The sibling guard `tests/slow_test_path_filter.rs` already solved this: it uses `git ls-files` and explicitly skips dot-prefixed roots, for exactly this reason.

**Fixed** — `scannable_files` now lists tracked files via `git ls-files -z`, matching the sibling guard. Only committed text can carry an instruction a reader would copy, so tracking status is the right filter.

### 4. `codecov.yml`: `tests/**` no longer reaches the relocated CLI tests

The `ignore` entry is root-anchored, so it misses `crates/ferx-cli/tests/cli_ferx.rs` (308 lines, essentially 100% executed). That file is now measured as product code — contradicting the ignore policy stated immediately above it, and quietly padding the number `project/floor_weekly` gates at 90%.

**Fixed** — added `crates/*/tests/**`, with a note on why the separate entry is needed.

### 5. The correct, most explicit invocation is reported as an offender

`names_root_bin` fires on any `--bin ferx` without checking whether the line also selects the package, so `cargo run --release -p ferx-cli --bin ferx -- model.ferx` — redundant but legal, and the clearest form a reader can write — trips the test. `unscoped_run` already excuses the package selector; `names_root_bin` should too.

**Fixed** — gated on the absence of `-p ferx-cli`. The classifier is also extracted from the tree scan into `is_broken_ferx_invocation` and pinned by a unit test, because a scan over today's tracked files can only show false *negatives* — it can never surface a false positive.

### 6. A1 guard reads comments as dependencies

The guard splits the manifest at the first `\n[package]` and rejects the substrings `ferx-tools` / `ferx-cli` anywhere in the remainder, comments included. It passes today only because the `[workspace]` block happens to sit above `[package]`. One explanatory comment placed below — the `exclude` block is an obvious spot for "crates/ is auto-excluded" — fails the test with a message about a forbidden dependency that does not exist.

**Fixed** — a quote-aware `strip_toml_comment` runs over the section first, with unit tests covering both the comment case and a `#` inside a quoted value.

### 7. `#[doc(hidden)]` scan is a raw substring match

It fires on the attribute appearing inside a doc comment or a string literal — and the crate docs are a likely place to *document* the ban, at which point the guard rejects its own explanation. It was also blind to `#[cfg_attr(…, doc(hidden))]`, which is the same bypass wearing a hat.

**Fixed** — matched in attribute position (trimmed line start), `cfg_attr` included. The remaining gap (an attribute broken across lines) is noted in the doc comment; rustfmt does not produce that for one this short.

### 8. `ferx-tools` version assertion rejects a pre-release

`assert_eq!(v.split('.').count(), 3)` assumes a bare `x.y.z`. `BUILD_INFO.ferx_version` is `env!("CARGO_PKG_VERSION")`, so `0.4.0-rc.1` — plausible given this PR's own crates.io-release discussion — yields four components and fails for a reason unrelated to what the test guards.

**Fixed** — asserts a non-empty, all-digit leading component instead.

---

### Verification

- `cargo test --no-default-features --features ci --test public_api_boundary` — 5 passed (2 new).
- `cargo test -p ferx-tools --no-default-features --features ferx-core/ci` — passed.
- `cargo clippy -p ferx-tools -p ferx-cli --tests --no-default-features --features ferx-core/ci` — clean, and finishes in 8s off the existing `ferx-core` rlib.
- `--unit-graph` re-checked for finding 1.

Pre-existing and untouched: `cargo clippy --workspace --tests` reports four `approx_constant` errors in `src/sens/two_cpt_explicit.rs` tests. CI's clippy step does not pass `--tests` to the root package, so it does not see them — worth a follow-up, not a blocker here.

PR #1141, opened 2026-08-31
feat(api): non-parametric bootstrap in ferx-tools, at PsN::bootstrap parity [closes #1140]

### LINKED ISSUE #1140: feat(tools): non-parametric bootstrap in ferx-tools, at PsN::bootstrap parity
## Summary

Implement the **non-parametric case bootstrap** in `ferx-tools` — resample subjects
with replacement, refit, and summarise the resulting parameter distribution into
bias, standard errors and confidence intervals. This is P3 of #1114 (the workspace
split), and the first real inhabitant of `ferx-tools`, which today holds one
placeholder `core_version()`.

Target: **feature parity with `PsN::bootstrap`** (user guide, PsN 5.7.1, revised
2018-03-02) for everything that is not NONMEM-specific plumbing, exposed as
`ferx bootstrap` and as a `ferx_tools::bootstrap` library API.

Bootstrap is the reason the boundary rule reads the way it does: *if it calls
`fit()` more than once, it is a tool.* This is 200+ fits of one model over
resampled data — pure orchestration, no new numerics in `ferx-core`.

## Why this and not just `$COVARIANCE`

ferx's SEs are the pure `R⁻¹` matrix (equivalent to NONMEM `$COVARIANCE
MATRIX=R`). Those are asymptotic and symmetric by construction. The bootstrap
gives an *empirical* uncertainty distribution that (a) does not assume normality,
(b) survives a non-PD or failed covariance step, and (c) reveals asymmetry in
poorly identified parameters — exactly the situation this repo keeps hitting
(#864 weakly-identified V1, #1084 fluconazole Q-collapse). It is also the standard
uncertainty deliverable in a population-PK report, so its absence is a real gap
against PsN.

---

## PsN option parity

Mapping every option in the guide. "Skip" = NONMEM-specific plumbing with no ferx
analogue.

| PsN option | ferx | Notes |
| --- | --- | --- |
| `-samples=N` | `--samples N` (default 200) | Guide: "roughly 200 model fits are necessary" |
| `-seed=N` | `--seed N` | See *Reproducibility* below — ferx is stricter than PsN here |
| `-threads=N` | `--threads N` | Parallel over replicates, not within a fit |
| `-sample_size=M` | `--sample-size M` | Default = number of subjects. Per-stratum form supported |
| `-stratify_on=COL` | `--stratify-on COL` | Whole-subject strata; error if a subject has >1 value |
| `-update_inits` / `-no-update_inits` | `--update-inits` (default on) | Replicates start from the base fit's final estimates |
| `-run_base_model` / `-no-` | `--run-base-model` (default on) | Base fit supplies bias reference + inits |
| `-skip_minimization_terminated` | `--skip-minimization-terminated` (default **on**) | `!FitResult.converged` |
| `-skip_estimate_near_boundary` | `--skip-estimate-near-boundary` (default **on**) | `WarningCode::BoundaryEstimate` in `warnings_structured` |
| `-skip_covariance_step_terminated` | `--skip-covariance-step-terminated` (default off) | Only meaningful with `--keep-covariance` |
| `-skip_with_covstep_warnings` | `--skip-with-covstep-warnings` (default off) | Ditto |
| `-keep_covariance` | `--keep-covariance` (default off) | Off ⇒ `run_covariance_step = false` per replicate. This is most of the speedup |
| `-keep_tables` | `--keep-tables` (default off) | Off ⇒ no per-replicate sdtab written |
| `-dofv` | `--dofv` | `outer_maxiter = 0` is already exact `MAXEVAL=0` semantics |
| `-summarize` | `--summarize` | Recompute results from an existing `raw_results.csv` under different exclusion criteria |
| `-bca` | *deferred* | Jackknife + BCa; guide itself flags it "very time-consuming". Follow-up |
| crash recovery (§7) | *partial* | `--summarize` covers the "too many samples filtered out" case. Resuming unfinished samples is a follow-up |
| `-allow_ignore_id` | skip | No ID-based `IGNORE`/`ACCEPT` in the ferx DSL |
| `-copy_data` | skip | ferx does not write a per-replicate data file unless asked |
| `-mceta=N` | skip | NONMEM `$ESTIMATION` option |
| `-rplots` | skip | Plotting belongs in `ferx-r` / xpose |
| `-keep_covariance` disk options, `-clean` | skip | No NONMEM run directory |

---

## Semantics that need deciding up front

These are the places where a faithful port is not the obvious implementation.

**1. Duplicate subjects must be made distinct.** A resample draws the same subject
more than once, and those copies have to enter the fit as *independent*
individuals — otherwise the likelihood is wrong. PsN handles this by renumbering
IDs in the written dataset. ferx keys subjects by `Subject.id` (a `String`), so
the copies get a deterministic suffix (`ID`, `ID#2`, `ID#3`, …) and the mapping is
recorded in the `included_individuals` / `sample_keys` outputs.

> **Carry over the PsN warning:** any model that reads `ID` as a covariate is
> invalidated by this. PsN documents the same hazard ("model code that relies on
> ID numbers will lead to errors"). Detect it — refuse the run if `ID` appears in
> the model's covariate set — rather than relying on the user to remember.

**2. Reproducibility is stricter than PsN's.** The guide admits a real defect:

> "The results of two runs will be different even if the seed is the same if the
> lst-file of the base model is present at the start of one run but not the
> other. Running the base model changes the state of the random number
> generator."

Do not port that. Derive each replicate's resample from a *per-replicate* seed
(`seed`, replicate index) so the drawn datasets depend only on `--seed`,
`--samples`, `--sample-size` and the strata — never on whether the base model ran,
on `--threads`, or on completion order. This is testable, and PsN's version is not.

**3. Stratification.** Strata come from a covariate that must take exactly one
value per subject; PsN errors otherwise and so should ferx, naming the offending
subject. Two `--sample-size` forms, both from the guide:

- a single number ⇒ strata get samples **in proportion to their size**, rounded
  (the guide notes the rounded parts need not sum to `sample_size`, except in the
  default case where it always does);
- an explicit per-stratum map, PsN syntax `--sample-size "1001=>12,1002=>24,1003=>10"`.

**4. Exclusion criteria are applied at summary time, not at fit time.** Every
replicate's result is written to `raw_results.csv` with its diagnostics; the
filters select rows when computing statistics. That is what makes `--summarize`
possible at all, and it matches PsN.

**5. Percentile CIs — the exact PsN estimator.** Weighted average at
`x_((n+1)p)`, verbatim from the guide:

```
n = number of observations + 1
p = percentile / 100
i = integer part of n*p
f = decimal part of n*p
percentile = (1-f)*x_i + f*x_(i+1)
```

with PsN's minimum-successful-sample rules: **19** for 5–95%, **39** for
2.5–97.5%, **199** for 0.5–99.5%, **1999** for 0.05–99.95%. Below the threshold
the interval is not reported (not silently extrapolated).

**6. `--keep-covariance` off is a semantic choice, not just speed.** With the
covariance step off, `standard.errors` in the raw results are absent for every
replicate — the bootstrap SE comes from the *spread of the estimates*, which is
the whole point. Only `--keep-covariance` populates the per-replicate SE columns
(and only then do the two `covstep` skip options mean anything). Make that
dependency an error, not a silent no-op.

---

## `ferx-core` API gaps

Deliberately small — almost everything needed is already public
(`fit`, `Population`/`Subject` with public fields, `FitResult`,
`parse_full_model_file`, `ParsedModel`, `bind_theta_levels`,
`read_nonmem_csv_with_covariates`, `configure_global_thread_pool`,
`FitOptions::{threads, verbose, run_covariance_step, outer_maxiter}`).

One genuine gap:

- **`api::run` has no "load, don't fit" entry point.** `run_model_with_data`
  parses the model, resolves the data path, applies `[data_selection]`, reads the
  population, binds `theta NAME[...]` level blocks and builds `ModelParameters` —
  then fits. A tool needs everything up to the fit. `build_init_params` and
  `build_selection_filter` are private, so today the only way to get a
  `ModelParameters` is to run a fit.

  Proposal: one public `prepare_run(model_path, data_path) -> PreparedRun`
  returning `{ parsed, population, init_params, covariate_table }`, with
  `run_model_with_data` reimplemented on top of it so there is one code path, not
  two. **A4 check:** this is equally useful from ferx-r (load a model + dataset to
  inspect or `predict()` without fitting), so it is not a workspace-only API.

  Both additions go in the baseline (`tools/update-public-api.sh`) in the
  implementing PR, per A6.

- **G2 (quiet inner fits) is already closed** — every `eprintln!` in the estimation
  path is behind `options.verbose`, and warnings go to `FitResult.warnings`. Worth
  a test pinning it, since 200 fits of per-iteration output is the failure mode.

- **G1 (thread-pool control) is already closed** — `configure_global_thread_pool`
  is public and `FitOptions::threads` scopes a fit-local pool. The bootstrap
  parallelises over *replicates* with a rayon pool and pins `FitOptions::threads =
  1` inside each, to avoid the N×N oversubscription this repo already hit in CI.

---

## Shape

```
crates/ferx-tools/src/bootstrap/
├─ mod.rs        BootstrapOptions, BootstrapResult, run_bootstrap()
├─ resample.rs   subject resampling: strata, sample sizes, per-replicate seeding
├─ summary.rs    bias, SEs, medians, percentile + SE-based CIs, diagnostic means
└─ output.rs     the CSV artefacts
```

```rust
pub fn run_bootstrap(
    prepared: &PreparedRun,
    options: &BootstrapOptions,
) -> Result<BootstrapResult, String>;
```

CLI:

```
ferx bootstrap run12.ferx --data warfarin.csv --samples 500 --seed 12345 --threads 5
ferx bootstrap run12.ferx --stratify-on STUDY --sample-size "1001=>12,1002=>24"
ferx bootstrap --summarize --directory boot_run12 --no-skip-minimization-terminated
```

This is the first `ferx` subcommand. `ferx model.ferx` (no subcommand) must keep
working unchanged — the dispatcher only treats the first argument as a subcommand
when it is a known verb and not an existing path. Whether that is the point to
adopt `clap` is an implementation call; #1114 explicitly left the door open, and
`ferx-cli` is where an arg-parsing dependency is allowed to live.

## Output files

PsN names, so an existing PsN user's tooling and habits transfer:

| File | Content |
| --- | --- |
| `raw_results.csv` | One row per fit, **first row = original dataset**; estimates, OFV, termination diagnostics |
| `bootstrap_results.csv` | means, bias, standard.errors, medians, percentile CIs, standard-error CIs, diagnostic means |
| `included_individuals1.csv` | Per replicate, the subject IDs drawn, in dataset order |
| `included_keys1.csv` | Per replicate, the internal 1..N ordinals |
| `sample_keys1.csv` | Per replicate, one count per original subject (times drawn) |
| `all_individuals1.csv` | Every subject of the original dataset, original order |
| `delta_ofv.csv` | `--dofv` only: bootstrap-data id, `ofv_bs,maxeval=0 − ofv_original` |

Row order is consistent across `raw_results` / `included_*` / `sample_keys`, as in
PsN, so row *j* means the same replicate in every file.

## Validation

Bootstrap has no single NONMEM run to anchor against — the comparator is
**PsN + NONMEM** (available here via the licensed `pmx` container), and it is
distributional, so the anchors have to be built from the parts that *are* exact:

1. **Degenerate oracle.** A "resample" that returns the original subject list
   must reproduce the base fit bit-for-bit. This pins the resampled-`Population`
   construction (ID suffixing, covariate carry-over, occasion/L2 columns) against
   the ordinary fit path.
2. **Frozen-seed replay.** For a fixed `--seed`, the drawn `sample_keys` are
   pinned in a Tier-1 unit test — and pinned to be independent of `--threads` and
   of whether the base model ran (see *Reproducibility*).
3. **Closed-form percentile CIs.** The PsN weighted-average estimator on
   hand-computed vectors, including the boundary cases at each minimum-N
   threshold (19/39/199/1999).
4. **Exact stratum composition.** Stratified draws must produce the stated counts
   per stratum, both proportional and explicit-map forms.
5. **PsN cross-check on warfarin** (Tier 3 / documented, not a CI gate): bootstrap
   SEs and 95% CIs from ferx vs PsN, and against ferx's own `$COVARIANCE`-style
   `R⁻¹` SEs. These agree to within Monte-Carlo error, not exactly; the comparison
   goes in `docs/` with the run details, per the "compare against NONMEM" rule.
6. **`--dofv` sanity.** With `--dofv`, evaluating the *original* fit's own
   parameter vector on the original data must give `Δofv = 0`.

## Scope of the first PR

In: everything in the parity table except `-bca` and unfinished-sample resume —
i.e. resampling (stratified and not), parallel replicate fits, `--update-inits`,
all four skip filters, summary statistics with PsN's CI estimator, the CSV
artefacts, `--dofv`, `--summarize`, the `ferx bootstrap` subcommand, a
`docs/tools/bootstrap.qmd` page, and a `CHANGELOG.md` entry.

Follow-ups: `-bca` (jackknife + acceleration), resuming a crashed run from
partial `raw_results`, and the `ferx-r` wrapper.

## Acceptance

- [ ] `ferx_tools::bootstrap::run_bootstrap` + `ferx bootstrap` subcommand; plain `ferx model.ferx` unaffected
- [ ] Stratified and unstratified resampling, both `--sample-size` forms, per-subject strata validated
- [ ] All four skip filters, applied at summary time, so `--summarize` can change them without refitting
- [ ] `bootstrap_results.csv` reproduces PsN's statistic set, with PsN's percentile estimator and minimum-N rules
- [ ] Resample stream depends only on the seed and the design — not on thread count, ordering, or whether the base model ran
- [ ] Replicate fits are quiet and the covariance step is off by default
- [ ] Model reading `ID` as a covariate is refused with a clear message
- [ ] `prepare_run` added to `api/ferx-core-public-api.txt` in the same PR, with `run_model_with_data` rebuilt on it
- [ ] Validation items 1–4 + 6 as tests; item 5 written up in docs
- [ ] `docs/` page added to `_quarto.yml`; `CHANGELOG.md` entry

Refs #1114.

roninsightrx:
## Why

ferx had no way to produce an empirical parameter-uncertainty distribution. The only
uncertainty on offer is the `R⁻¹` covariance step (equivalent to NONMEM `$COVARIANCE
MATRIX=R`), which is asymptotic and symmetric by construction, and which fails outright
when the matrix is non-PD. The non-parametric bootstrap is the standard uncertainty
deliverable in a population-PK report, and its absence was a real gap against PsN.

It is also P3 of #1114 and the first real inhabitant of `ferx-tools`, which until now held
one placeholder `core_version()`. Bootstrap is the boundary rule made concrete: *if it
calls `fit()` more than once, it is a tool.*

Closes #1140.

## What changed

`ferx_tools::bootstrap` + a `ferx bootstrap` subcommand. Resample whole subjects with
replacement, refit, and summarise the spread of the estimates.

**PsN parity** for everything that is not NONMEM run-directory plumbing:

| PsN | here |
|---|---|
| `-samples`, `-seed`, `-threads` | `--samples` (default 200), `--seed`, `--threads` |
| `-sample_size` | `--sample-size`, including the per-stratum `"1001=>12,1002=>24"` form |
| `-stratify_on` | `--stratify-on`, with PsN's "one value per individual" rule enforced |
| `-update_inits`, `-run_base_model` | `--no-update-inits`, `--no-run-base-model` |
| `-keep_covariance`, `-keep_tables` | `--keep-covariance`; no per-replicate sdtab is written |
| the four `-skip_*` filters | same four, same defaults (two on, two off) |
| `-dofv` | `--dofv` (`outer_maxiter = 0` is already exactly `MAXEVAL=0`) |
| `-summarize` | `--summarize` — re-reads `raw_results.csv`, refits nothing |
| output files | same names, same row-order guarantee across `raw_results` / `included_individuals` / `included_keys` / `sample_keys` |

Not implemented: `-bca` (needs a jackknife pass; the guide itself calls it "very
time-consuming") and resuming a crashed run's *unfinished* samples. `--summarize` covers
the "too many samples were filtered out" recovery case. `-allow_ignore_id`, `-copy_data`,
`-mceta`, `-rplots`, `-clean` have no ferx analogue.

**Three deliberate departures from PsN**, each because PsN's behaviour is a documented
defect rather than a design:

1. **Reproducibility.** Each replicate's draw is derived from the master seed *and its own
   index*, so the datasets depend only on `--seed` and the design. The PsN guide documents
   the opposite as a known wart — same seed, different results depending on whether the
   base model's `.lst` was present, because running it advances one shared RNG. Asserted
   here across `--threads 1` vs `4`.
2. **A model reading `ID` as a covariate is refused.** Resampling necessarily renames the
   duplicate copies of a subject (they must enter the fit as independent individuals), so
   such a model would silently see different values than in the base fit. PsN documents
   the identical hazard as a known bug and leaves it to the user to notice.
3. **`bootstrap_results.csv` is tidy** (one row per parameter) and the run-level
   diagnostics get their own file. PsN stacks statistic blocks sharing one header, which
   no data frame can read.

## Alternatives considered

**Where the ID renaming lives.** PsN renumbers IDs while writing each replicate's data
file. ferx keeps the original ID and appends `#2`, `#3`, … so a diagnostic can still be
traced back to the subject it came from, and `included_individuals1.csv` still writes the
*original* IDs so it agrees with `all_individuals1.csv`.

**Reading the stratification column.** First attempt used `Subject::covariates`, which is
wrong twice over: the column is usually a study identifier the model never declares (so it
is not on the subject at all), and a covariate that varies within a subject would be
silently accepted. It now reads the dataset directly, compares values as strings (so
non-numeric group labels work), and enforces the one-value-per-subject rule where the
offending subject can still be named.

**The normal quantile.** Started as a bisection on the CDF built from ferx-core's public
`erf`. It converges exactly — to the inverse of an `erf` whose own approximation error is
~1.5e-7, putting `z₀.₉₇₅` at 1.9599628 instead of 1.9599640. Inverting an approximation
buys the approximation's accuracy, not the bisection's. Replaced with Acklam's rational
approximation (~1.15e-9 relative), which is also exactly symmetric in the central branch.

**Not `clap`.** #1114 left the door open, but adding it would mean rewriting 968 lines of
hand-rolled parsing and its tests. `bootstrap` dispatches exactly like the existing `check`
and `summary` subcommands.

## Cross-repo dependency

`ferx-core`'s API only grew, so nothing in ferx-r breaks. A wrapper for `ferx bootstrap`
is a follow-up, and it will need the usual `Cargo.lock` bump to reach `prepare_run`.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated

The widening is one struct and two functions, and nothing else:

```
+pub struct ferx_core::PreparedRun            { parsed, population, init_params,
+                                               covariate_table, data_path,
+                                               data_path_warning, model_hash, data_hash }
+pub fn ferx_core::prepare_run(model_path, data_path) -> Result<PreparedRun, String>
+pub fn ferx_core::prepare_run_with_inits(model_path, data_path, inits_override) -> …
```

**Which caller needs it, and why the existing surface does not suffice** (A6): the
bootstrap needs everything `run_model_with_data` does *up to* the fit — data-path
resolution against the model's `[data]` block, the `[data_selection]` filter, the
covariate-aware read, `theta NAME[...]` level binding, and the initial `ModelParameters`.
`build_init_params` and `build_selection_filter` are private, so the only way to obtain a
`ModelParameters` was to run a fit. `run_model_with_data` is now implemented on top of
`prepare_run`, so there is one loading path rather than two — which matters more than the
convenience: a tool that loaded a model *slightly* differently from the CLI would produce
a bootstrap of a different model, silently.

**A4 (ferx-r reachability):** this is not a workspace-only API. "Load a model and its
dataset without fitting" is exactly what an R caller wants in order to inspect a
population or call `predict()`.

## Numerical validation

A bootstrap has no single NONMEM run to anchor against — the comparator is distributional.
So the anchors are built from the parts that *are* exact:

- [x] **Degenerate oracle.** A "resample" drawing every subject exactly once reproduces an
      ordinary fit **bit-for-bit** (`ofv.to_bits()`). This is what pins the replicate
      population against the engine: a dropped covariate map or a lost occasion column
      would change the objective and nothing else in the bootstrap would notice.
- [x] **Its complement**, because the identity case cannot see the duplication path:
      duplicating a subject grows `n_obs` by that subject's observations, gives the two
      copies identical EBEs and identical `ofv_contribution`, and the 2nd and 3rd copies
      move the objective by the *same* amount to 1e-9.
- [x] **Closed-form percentile CIs** against hand-computed values, plus all four of PsN's
      minimum-sample thresholds (19 / 39 / 199 / 1999) asserted from *both* sides. Those
      are not four rules — they all fall out of `(n+1)p ≥ 1`.
- [x] **Exact stratum composition** over repeated draws, on the real warfarin dataset given
      a synthetic `STUDY` column, plus the "one value per individual" refusal on real data.
- [x] **Seed reproducibility** across `--threads 1` vs `--threads 4`.
- [x] **Flat-vector round trip through a real model**: re-evaluating a fit's own estimates
      returns its OFV bit-for-bit — i.e. Δofv against itself is exactly 0, which is the
      `--dofv` sanity check.
- [x] Compared against: prior ferx commit's `R⁻¹` standard errors (below).

Warfarin, 200 samples, seed 12345 — six of seven parameters agree closely with the
covariance step, which is the reassuring result. The seventh is the case for the tool:

```
parameter               estimate   SE(R⁻¹)  SE(boot)   95% percentile     95% normal
TVCL                     0.13297   0.00663   0.00809   0.1185 – 0.1506   0.1171 – 0.1488
TVV                      7.73070   0.23406   0.24437   7.237  – 8.127    7.252  – 8.210
TVKA                     0.72521   0.12453   0.14048   0.4675 – 1.0641   0.4499 – 1.0006
OMEGA(ETA_CL,ETA_CL)     0.02859   0.01280   0.01046   0.0063 – 0.0474   0.0081 – 0.0491
OMEGA(ETA_V,ETA_V)       0.00958   0.00430   0.00374   0.0016 – 0.0158   0.0023 – 0.0169
OMEGA(ETA_KA,ETA_KA)     0.34896   0.16082   0.20343   0.1092 – 0.8622  -0.0498 – 0.7477
PROP_ERR                 0.01075   0.00094   0.00100   0.00885– 0.01259  0.00879– 0.01271
```

`OMEGA(ETA_KA)`: bootstrap SE 26% larger, strongly right-skewed (0.109–0.862 around
0.349), and the normal-approximation lower bound is **negative** — an impossible variance.
The percentile interval cannot go there, because every endpoint it reports is a value some
replicate actually produced. Written up in `docs/tools/bootstrap.qmd`.

`--dofv` on the same model returns all-positive differences, as it must: the original
fit's parameters minimise the objective on the original data.

## Performance

- [x] No significant impact expected — new code path only; nothing existing changed except
      `run_model_with_data`, which now calls the function its body became.

200 warfarin replicates in **2.5 s** on 8 threads (release). Replicate fits are quiet,
pinned to one thread each (the bootstrap parallelises over replicates, so nesting Rayon
pools would oversubscribe), skip the covariance step, skip SIR, and disable checkpointing —
without that last one, 200 fits would write and resume the *same* `{model}.tmp` and restore
each other's state.

## Tests

- [x] Unit tests in the same module (`cargo test --lib` passes)
- [x] Regression tests that fail without the fix — the oracles above

124 new tests: 55 unit (`ferx-tools`), 45 + 16 (`ferx-cli`), 8 Tier-2 end-to-end
(`crates/ferx-tools/tests/bootstrap_end_to_end.rs`), 4 for `prepare_run`
(`tests/prepare_run.rs`). All Tier-1/Tier-2: the end-to-end fits are capped at 1–2 outer
iterations or are pure evaluations, and the whole file runs in 2.2 s.

`cargo test --lib --no-default-features --features ci`: 3706 passed, 0 failed — the
`run_model_with_data` refactor is behaviour-preserving.

## Example

```bash
ferx bootstrap warfarin.ferx --data warfarin.csv --samples 200 --seed 12345 --threads 8
ferx bootstrap run12.ferx --stratify-on STUDY --sample-size "1001=>12,1002=>24,1003=>10"
ferx bootstrap --summarize --directory run12-bootstrap --no-skip-minimization-terminated
```

```
Model: warfarin
Data:  10 subjects, 110 observations from data/warfarin.csv
Bootstrap: 200 samples, seed 12345
Fitted 200 replicates in 2.5s

200 of 200 samples completed; 200 included in the statistics

parameter                original         mean         bias     SE(boot)       95% lo       95% hi
TVCL                      0.13297      0.13350      0.00053      0.00809      0.11852      0.15058
TVV                       7.73070      7.72536     -0.00534      0.24437      7.23688      8.12707
TVKA                      0.72521      0.73430      0.00909      0.14048      0.46745      1.06406
OMEGA(ETA_CL,ETA_CL)      0.02859      0.02595     -0.00265      0.01046      0.00630      0.04741
OMEGA(ETA_V,ETA_V)        0.00958      0.00848     -0.00110      0.00374      0.00163      0.01582
OMEGA(ETA_KA,ETA_KA)      0.34896      0.33303     -0.01594      0.20343      0.10923      0.86215
PROP_ERR                  0.01075      0.01067     -0.00008      0.00100      0.00885      0.01259

Wrote bootstrap results to warfarin-bootstrap
```

## Docs

- [x] `docs/tools/bootstrap.qmd` added, `docs/cli.qmd` updated
- [x] New page linked in `docs/_quarto.yml` (new "Tools" sidebar section)

## Checklist
- [x] `cargo clippy` clean (`-p ferx-tools -p ferx-cli --tests`)
- [ ] `Cargo.toml` version bump — not needed, additive only

## Docs & examples
- [x] No new `.ferx` DSL syntax or `[fit_options]` key, no new example model or data file
- [x] Rebuilt: `cargo build --release --workspace`
- [x] Ran the real binary against `examples/warfarin.ferx` + `data/warfarin.csv` for the
      table above, plus `--dofv`, `--summarize` and `--stratify-on`
- [ ] ferx-r rebuild — not needed, no ferx-r change in this PR

## Reviewer hints

Worth the attention:

- **`resample.rs`** — the draw is sorted by original ordinal on purpose. PsN builds a
  replicate by walking the original file and emitting each selected individual as many
  times as drawn, so dataset order and draw order coincide there; sorting reproduces that,
  which is what lets `included_individuals` and `included_keys` mean what the PsN guide
  says they mean. Statistically a no-op — a multiset has no order.
- **`summary::percentile`** — the `None` return is the whole minimum-sample-size rule.
- **`replicate_options`** in `bootstrap/mod.rs` — four settings that each look like
  tidiness and are not, especially `checkpoint = false`.
- **`src/api/run.rs`** — the refactor is a straight extraction; the only behavioural
  difference is that the two `eprintln!` lines moved into `run_model_with_data` so
  `prepare_run` is silent for a caller running it hundreds of times.

Skimmable: `output.rs` (CSV writers) and the test files.

## Open questions

- The `--summarize` path cannot recover `chi_square_df` from `raw_results.csv`, so it
  reads it back out of the existing `bootstrap_diagnostics.csv`. Alternative would be a
  small run-metadata file; that seemed like more machinery than the one value justifies.
- `-bca` and resuming unfinished samples are follow-ups. Worth separate issues now, or
  leave them in #1140's follow-up list?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,12 @@ section of the SDLC for the versioning policy).
   recompute results under different filters without refitting. Writes PsN-named CSV artefacts
   (`raw_results.csv`, `bootstrap_results.csv`, `included_individuals1.csv`, `sample_keys1.csv`, …).
   Unlike PsN, the drawn datasets depend only on the seed and the design — never on the thread
-  count, the completion order, or whether the base model was fitted first. A model that reads `ID`
-  as a covariate is refused rather than silently mis-fitted. See `docs/tools/bootstrap.qmd`.
+  count, the completion order, or whether the base model was fitted first. Bootstrapped parameters
+  are every estimated theta, Omega (free lower triangle), sigma, and IOV Omega, so a `kappa` model
+  gets bootstrap SEs and CIs for its inter-occasion variance and `--update-inits` / `--dofv` carry
+  it too. A model that reads `ID` as a covariate is refused rather than silently mis-fitted, and so
+  is a `[mixture]` model — its classes are identified only up to relabelling, so averaging across
+  replicates would mix them. See `docs/tools/bootstrap.qmd`.
 - **`prepare_run` / `prepare_run_with_inits` (#1140)** — public "load a model and its dataset, but
   do not fit" entry points returning a `PreparedRun`. `run_model_with_data` is now implemented on
   top of them, so a tool and the CLI cannot diverge in how a model is loaded.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,23 @@ section of the SDLC for the versioning policy).
   `tools/update-public-api.sh`. This doubles as semver protection for the crates.io release and
   for ferx-r. `#[doc(hidden)] pub` is banned, because `cargo public-api` omits such items and the
   attribute would otherwise be a silent bypass of the gate.
-- **A `ferx-tools` crate (#1114)** — the future home of multi-fit tooling (bootstrap, stepwise
-  covariate modelling, model search, cross-validation). Currently a placeholder; it can only reach
-  `ferx-core` through the same public API the R wrapper uses.
+- **A `ferx-tools` crate (#1114)** — the home of multi-fit tooling (bootstrap, stepwise
+  covariate modelling, model search, cross-validation). It can only reach `ferx-core` through the
+  same public API the R wrapper uses.
+- **`ferx bootstrap` — the non-parametric case bootstrap (#1140).** Resamples subjects with
+  replacement, refits the model to each replicate, and reports bias, standard errors and
+  confidence intervals from the spread of the estimates, at `PsN::bootstrap` feature parity:
+  `--samples`, `--seed`, `--threads`, `--sample-size` (including PsN's per-stratum
+  `"1001=>12,1002=>24"` form), `--stratify-on`, `--update-inits`, `--run-base-model`,
+  `--keep-covariance`, the four `--skip-*` exclusion filters, `--dofv`, and `--summarize` to
+  recompute results under different filters without refitting. Writes PsN-named CSV artefacts
+  (`raw_results.csv`, `bootstrap_results.csv`, `included_individuals1.csv`, `sample_keys1.csv`, …).
+  Unlike PsN, the drawn datasets depend only on the seed and the design — never on the thread
+  count, the completion order, or whether the base model was fitted first. A model that reads `ID`
+  as a covariate is refused rather than silently mis-fitted. See `docs/tools/bootstrap.qmd`.
+- **`prepare_run` / `prepare_run_with_inits` (#1140)** — public "load a model and its dataset, but
+  do not fit" entry points returning a `PreparedRun`. `run_model_with_data` is now implemented on
+  top of them, so a tool and the CLI cannot diverge in how a model is loaded.
 
 ### Fixed
 - **A steady-state dose with a lagtime is now seeded at the dose record, not equilibrated at the
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -262,6 +262,7 @@ name = "ferx-cli"
 version = "0.3.0"
 dependencies = [
  "ferx-core",
+ "ferx-tools",
  "serde_json",
  "tempfile",
 ]
@@ -293,7 +294,12 @@ dependencies = [
 name = "ferx-tools"
 version = "0.3.0"
 dependencies = [
+ "csv",
  "ferx-core",
+ "nalgebra",
+ "rand 0.10.2",
+ "rayon",
+ "tempfile",
 ]
 
 [[package]]
--- a/api/ferx-core-public-api.txt
+++ b/api/ferx-core-public-api.txt
@@ -17,6 +17,15 @@ pub struct ferx_core::api::PredictionResult
 pub ferx_core::api::PredictionResult::id: alloc::string::String
 pub ferx_core::api::PredictionResult::pred: f64
 pub ferx_core::api::PredictionResult::time: f64
+pub struct ferx_core::api::PreparedRun
+pub ferx_core::api::PreparedRun::covariate_table: core::option::Option<ferx_core::CovariateTable>
+pub ferx_core::api::PreparedRun::data_hash: core::option::Option<alloc::string::String>
+pub ferx_core::api::PreparedRun::data_path: alloc::string::String
+pub ferx_core::api::PreparedRun::data_path_warning: core::option::Option<alloc::string::String>
+pub ferx_core::api::PreparedRun::init_params: ferx_core::ModelParameters
+pub ferx_core::api::PreparedRun::model_hash: core::option::Option<alloc::string::String>
+pub ferx_core::api::PreparedRun::parsed: ferx_core::ParsedModel
+pub ferx_core::api::PreparedRun::population: ferx_core::Population
 pub struct ferx_core::api::SimulateOptions
 pub ferx_core::api::SimulateOptions::horizon: core::option::Option<f64>
 pub ferx_core::api::SimulateOptions::match_method: core::option::Option<ferx_core::propensity_match::MatchMethod>
@@ -60,6 +69,8 @@ pub fn ferx_core::api::fit_from_files(model_path: &str, data_path: core::option:
 pub fn ferx_core::api::predict(model: &ferx_core::CompiledModel, population: &ferx_core::Population, params: &ferx_core::ModelParameters) -> alloc::vec::Vec<ferx_core::PredictionResult>
 pub fn ferx_core::api::predict_categorical(model: &ferx_core::CompiledModel, population: &ferx_core::Population, params: &ferx_core::ModelParameters) -> alloc::vec::Vec<ferx_core::EndpointPredictionResult>
 pub fn ferx_core::api::predict_survival(model: &ferx_core::CompiledModel, population: &ferx_core::Population, params: &ferx_core::ModelParameters, time_grid: &[f64]) -> alloc::vec::Vec<ferx_core::SurvivalPredictionResult>
+pub fn ferx_core::api::prepare_run(model_path: &str, data_path: core::option::Option<&str>) -> core::result::Result<ferx_core::PreparedRun, alloc::string::String>
+pub fn ferx_core::api::prepare_run_with_inits(model_path: &str, data_path: core::option::Option<&str>, inits_override: core::option::Option<ferx_core::suggest_start::NcaInit>) -> core::result::Result<ferx_core::PreparedRun, alloc::string::String>
 pub fn ferx_core::api::read_population_for(model: &ferx_core::CompiledModel, covariate_decls: &core::option::Option<alloc::vec::Vec<ferx_core::CovariateDecl>>, data_path: &str, fallback_columns: core::option::Option<&[&str]>, iov_column: core::option::Option<&str>, filter: core::option::Option<&ferx_core::io::datareader::SelectionFilter>, column_map: &[(alloc::string::String, alloc::string::String)]) -> core::result::Result<(ferx_core::Population, core::option::Option<ferx_core::CovariateTable>), alloc::string::String>
 pub fn ferx_core::api::read_population_for_simulation(model: &ferx_core::CompiledModel, covariate_decls: &core::option::Option<alloc::vec::Vec<ferx_core::CovariateDecl>>, data_path: &str, fallback_columns: core::option::Option<&[&str]>, iov_column: core::option::Option<&str>, filter: core::option::Option<&ferx_core::io::datareader::SelectionFilter>, column_map: &[(alloc::string::String, alloc::string::String)]) -> core::result::Result<(ferx_core::Population, core::option::Option<ferx_core::CovariateTable>), alloc::string::String>
 pub fn ferx_core::api::resolve_data_path(model_data_path: core::option::Option<&str>, external_data_path: core::option::Option<&str>) -> core::result::Result<(alloc::string::String, core::option::Option<alloc::string::String>), alloc::string::String>
@@ -4070,6 +4081,15 @@ pub struct ferx_core::PredictionResult
 pub ferx_core::PredictionResult::id: alloc::string::String
 pub ferx_core::PredictionResult::pred: f64
 pub ferx_core::PredictionResult::time: f64
+pub struct ferx_core::PreparedRun
+pub ferx_core::PreparedRun::covariate_table: core::option::Option<ferx_core::CovariateTable>
+pub ferx_core::PreparedRun::data_hash: core::option::Option<alloc::string::String>
+pub ferx_core::PreparedRun::data_path: alloc::string::String
+pub ferx_core::PreparedRun::data_path_warning: core::option::Option<alloc::string::String>
+pub ferx_core::PreparedRun::init_params: ferx_core::ModelParameters
+pub ferx_core::PreparedRun::model_hash: core::option::Option<alloc::string::String>
+pub ferx_core::PreparedRun::parsed: ferx_core::ParsedModel
+pub ferx_core::PreparedRun::population: ferx_core::Population
 pub struct ferx_core::ResidualCorrelation
 pub ferx_core::ResidualCorrelation::rho: f64
 pub ferx_core::ResidualCorrelation::sigma_i: usize
@@ -4249,6 +4269,8 @@ pub fn ferx_core::predict(model: &ferx_core::CompiledModel, population: &ferx_co
 pub fn ferx_core::predict_categorical(model: &ferx_core::CompiledModel, population: &ferx_core::Population, params: &ferx_core::ModelParameters) -> alloc::vec::Vec<ferx_core::EndpointPredictionResult>
 pub fn ferx_core::predict_survival(model: &ferx_core::CompiledModel, population: &ferx_core::Population, params: &ferx_core::ModelParameters, time_grid: &[f64]) -> alloc::vec::Vec<ferx_core::SurvivalPredictionResult>
 pub fn ferx_core::prepare_frem(model_path: &std::path::Path, data_path: &std::path::Path, covariates: &[alloc::string::String], categorical_covariates: core::option::Option<&[alloc::string::String]>, output_model_path: core::option::Option<&std::path::Path>, output_data_path: core::option::Option<&std::path::Path>, missing_value: core::option::Option<f64>, fit_init: core::option::Option<&ferx_core::frem::FremFitInit>) -> core::result::Result<ferx_core::frem::FremPrepareResult, alloc::string::String>
+pub fn ferx_core::prepare_run(model_path: &str, data_path: core::option::Option<&str>) -> core::result::Result<ferx_core::PreparedRun, alloc::string::String>
+pub fn ferx_core::prepare_run_with_inits(model_path: &str, data_path: core::option::Option<&str>, inits_override: core::option::Option<ferx_core::suggest_start::NcaInit>) -> core::result::Result<ferx_core::PreparedRun, alloc::string::String>
 pub fn ferx_core::read_nonmem_csv(path: &std::path::Path, covariate_columns: core::option::Option<&[&str]>, iov_column: core::option::Option<&str>) -> core::result::Result<ferx_core::Population, alloc::string::String>
 pub fn ferx_core::read_nonmem_csv_with_covariates(path: &std::path::Path, decls: &[ferx_core::CovariateDecl], extra_columns: &[alloc::string::String], iov_column: core::option::Option<&str>) -> core::result::Result<(ferx_core::Population, ferx_core::CovariateTable), alloc::string::String>
 pub fn ferx_core::resolve_data_path(model_data_path: core::option::Option<&str>, external_data_path: core::option::Option<&str>) -> core::result::Result<(alloc::string::String, core::option::Option<alloc::string::String>), alloc::string::String>
--- a/crates/ferx-cli/Cargo.toml
+++ b/crates/ferx-cli/Cargo.toml
@@ -24,6 +24,7 @@ path = "src/main.rs"
 # silently defeated for `ferx-core` everywhere (harmless only while
 # `default = []`; `survival` is slated to become default-on).
 ferx-core = { path = "../..", version = "0.3.0", default-features = false }
+ferx-tools = { path = "../ferx-tools", version = "0.3.0" }
 serde_json = "1"
 
 [dev-dependencies]
--- a/crates/ferx-cli/src/bootstrap_cmd.rs
+++ b/crates/ferx-cli/src/bootstrap_cmd.rs
@@ -1,0 +1,395 @@
+//! `ferx bootstrap` — the non-parametric case bootstrap (#1140).
+//!
+//! Thin by construction (#1114 A5): parse flags, call
+//! [`ferx_tools::bootstrap::run_bootstrap`], print a table, pick an exit code.
+//! Every decision about resampling, filtering or statistics lives in
+//! `ferx-tools`.
+
+use std::path::PathBuf;
+
+use ferx_tools::bootstrap::{
+    resummarize, run_bootstrap, BootstrapOptions, BootstrapResult, SampleSize,
+};
+
+pub const BOOTSTRAP_USAGE: &str = "\
+Usage: ferx bootstrap <model.ferx> [--data <data.csv>] [options]
+       ferx bootstrap --summarize --directory <dir> [filter options]
+
+Resamples subjects with replacement, refits the model to each replicate, and
+reports bias, standard errors and confidence intervals from the spread of the
+estimates. Resampling is always over whole subjects.
+
+  --samples N          number of bootstrap datasets (default 200)
+  --seed N             master seed (default 1). Each replicate's draw is derived
+                       from this and its own index, so the datasets do not
+                       depend on --threads or on completion order.
+  --threads N          replicates to fit concurrently (each fit uses 1 thread)
+  --sample-size M      subjects per replicate (default: as many as the dataset).
+                       With --stratify-on, may instead be a per-stratum map:
+                       --sample-size \"1001=>12,1002=>24,1003=>10\"
+  --stratify-on COL    resample within strata defined by a dataset column. The
+                       column must take exactly one value per subject.
+  --directory DIR      where the CSV artefacts go (default {model}-bootstrap)
+  --ci LEVEL           two-sided confidence level in percent (default 95)
+
+  --no-update-inits    start replicates from the model file's initial estimates
+                       instead of the base fit's final ones
+  --no-run-base-model  skip the fit on the original dataset (disables bias, the
+                       standard-error intervals, --update-inits and --dofv)
+  --keep-covariance    run the covariance step for every replicate (slow; only
+                       needed for the two covstep filters below)
+  --dofv               evaluate each replicate's estimates on the original data
+                       with no estimation, and report the OFV difference
+
+  --no-skip-minimization-terminated    keep replicates that did not converge
+  --no-skip-estimate-near-boundary     keep replicates with an estimate on a bound
+  --skip-covariance-step-terminated    drop replicates whose covariance step failed
+  --skip-with-covstep-warnings         drop replicates with covariance warnings
+
+  --summarize          recompute the statistics from an existing run directory's
+                       raw_results.csv under different filters. Refits nothing.
+";
+
+fn flag(args: &[String], name: &str) -> bool {
+    args.iter().any(|a| a == name)
+}
+
+fn value<'a>(args: &'a [String], name: &str) -> Option<&'a String> {
+    args.iter()
+        .position(|a| a == name)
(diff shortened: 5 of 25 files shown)


roninsightrx (commented):
I found four correctness issues that should be fixed before merge.

1. **[P1] IOV parameters are silently excluded from the bootstrap** (`crates/ferx-tools/src/bootstrap/mod.rs:157-223`). `parameter_names` and `flatten_estimates` stop after theta / BSV Omega / sigma, and `params_from_estimates` merely clones the template's `omega_iov`. As a result, KAPPA gets no bootstrap SE or CI, `--update-inits` does not use the base fit's final KAPPA estimate, and `--dofv` does not evaluate the replicate's actual full parameter vector. I reproduced this with the shipped `examples/warfarin_iov.ferx`: the command succeeds but prints only TVCL/TVV/TVKA, the three BSV variances, and PROP_ERR; KAPPA_CL is absent. Mixture Omega/Sigma overrides have the analogous flat-vector gap. Please either include all fitted population parameters in the round trip or reject unsupported model classes explicitly; silently returning incomplete uncertainty is unsafe.

2. **[P2] `chi_square_df` undercounts block Omega parameters** (`crates/ferx-tools/src/bootstrap/mod.rs:668-674`). Counting `omega_fixed` counts etas, not the free lower-triangle covariance elements. A free 2x2 block contributes three estimated Omega parameters, but this function counts two. The test template already demonstrates the inconsistency: `parameter_names(&template(false))` has 6 entries while `count_estimated(&template(false))` returns 5. This writes the wrong chi-square reference for `--dofv`; IOV and mixture override coordinates are omitted too. Count the actual free packed coordinates (including free-mask/fixed semantics), not per-eta flags.

3. **[P2] `--keep-covariance` labels block-Omega SEs with the wrong parameters** (`crates/ferx-tools/src/bootstrap/mod.rs:455-459` and `503-509`). Parameter names/estimates are emitted in row-major lower-triangle order and omit structural zeros, whereas core documents `FitResult::se_omega` as the full column-major lower triangle for block Omega. For a 3x3 block, the third name is OMEGA(2,2), but the third raw SE is OMEGA(3,1); in a mixed block+diagonal Omega, structural-zero SE slots also shift every later value and can push sigma SEs out of alignment. Build the SE vector by looking up each named `(i,j)` (for example via `omega_se_at`) in exactly the same traversal used for names/estimates.

4. **[P2] `--summarize` bypasses option validation and can panic** (`crates/ferx-tools/src/bootstrap/mod.rs:588-624`). The fresh-run path validates the confidence level, but `resummarize` calls `summarize` directly. On a valid saved run, `ferx bootstrap --summarize --directory <dir> --ci 100` panics at `normal_quantile` and exits 101 instead of returning a user error. I reproduced this locally. Please share validation between fresh runs and re-summarization (and preferably keep invalid input from reaching an assertion in a library API).

Verification: `cargo test -p ferx-tools -p ferx-cli` passes (45 CLI unit, 16 CLI integration, 55 tools unit, and 8 bootstrap end-to-end tests). The missing coverage is specifically IOV/block-Omega end-to-end behavior and invalid re-summarization options.

roninsightrx:
All four confirmed and fixed in `2bf31c58`. Each was real, and each now has a regression test that fails without the fix. Thank you — findings 1 and 3 in particular were the kind that stay invisible: the run succeeds and quietly reports the wrong thing.

### 1 — IOV silently excluded (P1)

Confirmed, including the three downstream consequences you named. `ModelParameters::omega_iov` and `FitResult::omega_iov` were both simply not visited.

The flat vector now carries the free lower triangle of the IOV Omega as `OMEGA_IOV(KAPPA_CL,KAPPA_CL)`, **appended after sigma** so a model without IOV keeps exactly the columns it had. `--update-inits` and `--dofv` go through the same vector, so both now carry KAPPA. On `examples/warfarin_iov.ferx`:

```
parameter                        original         mean         bias     SE(boot)
...
OMEGA_IOV(KAPPA_CL,KAPPA_CL)      0.03904      0.04303      0.00399      0.01121
```

**Mixture models are refused rather than extended.** You are right that they have the analogous flat-vector gap, but that is the smaller half of the problem. A mixture's classes are identified only up to relabelling, so replicate *k*'s "class 1" need not be the original fit's class 1; averaging estimates across replicates without resolving the label switching gives a bias-and-interval table that is meaningless rather than merely noisy. That needs a relabelling rule this tool does not have, so `reject_mixture_model` says so and names the reason.

### 2 — `chi_square_df` undercounts block Omega (P2)

Confirmed, and the `parameter_names(&template(false))` = 6 vs `count_estimated` = 5 demonstration was exactly right. `count_estimated` now walks the same coordinate list as the names and filters on fixed-ness, with a block off-diagonal counting as fixed when either of its etas is (the flags are per-eta, so that is the only coherent mapping). IOV coordinates are included.

Worth stating why this one mattered more than its size suggests: the error was always in the direction of *too few* degrees of freedom, which makes the Δofv distribution compare favourably against a reference that is too small — a diagnostic failing in the one direction that does not raise an alarm.

### 3 — block-Omega SEs mislabelled (P2)

Confirmed. `se_omega` is column-major with structural zeros; the flat vector is row-major without them. Each SE is now looked up by its own `(i, j)` through `omega_se_at`, in the same traversal that produced the name.

One thing your description surfaced that needed a type change: `se_kappa` carries only the IOV **diagonal** variances, so an IOV covariance has no reported SE at all. `ReplicateResult::standard_errors` is therefore `Option<Vec<Option<f64>>>` — an absent SE is written as an empty CSV cell and read back as `None`, rather than encoded as a zero.

The regression test is `examples/warfarin_block_omega.ferx` (2×2 block over ETA_CL/ETA_V plus a standalone ETA_KA), which is your mixed block+diagonal case: two of the six triangle slots are structural zeros, so the old code produced a **ten**-element SE vector against **eight** names. `assert_eq!(se.len(), parameter_names.len())` is the decisive check.

### 4 — `--summarize` bypassed validation (P2)

Confirmed, exit 101 reproduced. Validation is now shared, but deliberately split in two:

- `BootstrapOptions::validate` — sample count and confidence level; called by both `run_bootstrap` and `resummarize`.
- `validate_for_run` — the covstep filters needing `--keep-covariance`; **fresh runs only**. Under `--summarize` the covariance diagnostics are already in `raw_results.csv`, so filtering on them is precisely the point and requires no covariance step now. Applying the fresh-run rule there would have rejected a legitimate re-summarization.

On your parenthetical: agreed, and `summarize` is `pub`, so it no longer trusts its caller either — an out-of-range level now yields `ci_standard_error: None` instead of reaching the assertion. `normal_quantile` keeps its domain assertion as a math primitive.

### The structural change behind 1–3

These were three symptoms of one thing: five separate hand-written walks over "theta, then the free lower triangle, then sigma" — for names, estimates, SEs, the rebuild into `ModelParameters`, and the parameter count. Two had already drifted. There is now one `Coord` enum and one `coordinates()` traversal that all five derive from, plus a test asserting they agree in length and order across diagonal, block, and IOV templates.

### Verification

`cargo test -p ferx-tools -p ferx-cli` — 158 passed (was 147): 57 CLI unit, 16 CLI integration, 65 tools unit, 20 end-to-end. `cargo test --lib --features ci` — 3706 passed. Clippy clean, public-API baseline unchanged (the fixes are all inside `ferx-tools`). Member line coverage 95.56%.

New tests, all failing without their fix: KAPPA present through names → estimates → summary → `raw_results.csv` → `--update-inits` round trip; `count_estimated == parameter_names().len()` for block and IOV templates; the block-Omega SE length/alignment check; mixture refusal; `resummarize` rejecting `--ci` at 0, 100 and beyond, plus the covstep filters still being *allowed* on the summarize path; and `summarize` returning no normal interval rather than panicking.

Docs (`docs/tools/bootstrap.qmd`) gained a "Which parameters are bootstrapped" section covering IOV, the empty-SE-cell convention, and the mixture refusal; `CHANGELOG.md` updated.

Issue #1145, opened 2026-08-31
decide: can a mixture model be bootstrapped? (label switching)


roninsightrx:
## Summary

Decide whether — and how — a `[mixture]` model (#977) can be bootstrapped, and either
implement it or write the refusal down as permanent.

#1141 refuses mixture models outright:

> the bootstrap does not support mixture models ([mixture] block). A mixture's classes
> are identified only up to relabelling, so a replicate's class 1 need not be the
> original fit's class 1, and averaging estimates across replicates would mix them.
> Bootstrapping a mixture needs a relabelling rule this tool does not have.

That is the right default — a wrong interval is worse than no interval — but it is a
placeholder for a decision, not the decision.

## The problem, precisely

A K-class mixture likelihood is invariant under permutation of the class labels: the
model with (class 1 = fast metabolisers, class 2 = slow) and the model with the labels
swapped have identical objective values. Which one an optimizer lands on depends on the
starting values and the data.

For a single fit that is harmless — the classes are read off the estimates. For a
bootstrap it is fatal in a way that produces *plausible numbers*:

- replicate *b* may converge to the permuted labelling;
- `mean(θ_class1)` then averages the fast metabolisers' CL over some replicates and the
  slow ones' over others;
- the resulting standard error is inflated by roughly the *between-class separation*,
  which is a property of the model, not a sampling uncertainty;
- and the percentile interval becomes bimodal, so its 2.5th percentile is a draw from
  the wrong class entirely.

Nothing about that failure looks like a failure. The run completes, every replicate
converges, and the table is wrong by an amount that grows with how well-separated the
classes are — i.e. it is worst exactly when the mixture is most justified.

`--update-inits` (on by default) mitigates but does not solve it: starting every
replicate from the base fit's estimates makes the original labelling the more likely
basin, but a resampled dataset can still cross over, and "usually right" is not a
property an uncertainty interval can be built on.

## Candidate relabelling rules

**(a) Order on a canonical statistic.** Sort classes by, say, the class-specific typical
CL, in every replicate and in the base fit. Cheap, deterministic, and the standard
identifiability constraint in the mixture literature. Fails when the ordering statistic
is not well separated — two classes with nearly equal CL that differ in V will be sorted
essentially at random, reintroducing the problem. Needs the user to say which parameter
orders the classes, or a documented default.

**(b) Match on per-subject class assignments.** `SubjectResult::mixest` (the most-likely
[...]

PR #1147, opened 2026-08-31
feat(tools): resume an interrupted bootstrap, reusing finished replicates [closes #1143]

### LINKED ISSUE #1143: feat(tools): resume an interrupted bootstrap, reusing finished replicates
## Summary

Let a bootstrap run that was interrupted pick up where it stopped, reusing the replicates
that already finished. This is the half of `PsN::bootstrap`'s recovery story (§7 of the
user guide) that #1141 did not implement.

`--summarize` already covers the *other* recovery case — "too many samples were filtered
out", where nothing needs refitting and only the exclusion criteria change. What is
missing is "the process was killed / the machine crashed / I want to rerun only the
samples that failed", which today means starting over from sample 1.

## Why this is cheap here, and better than PsN's version

Two properties the current design already has:

**The draws are a pure function of `(seed, index)`.** `resample::replicate_seed` derives
each replicate's RNG from the master seed and the replicate's own index, so replicate 57
of a resumed run draws *exactly* the dataset replicate 57 of the interrupted run would
have. A resumed run is therefore bit-identical to an uninterrupted one — not merely
statistically equivalent. PsN cannot promise that: its guide documents the RNG as shared
and order-dependent, so resuming there gives a different set of datasets than not
resuming would have.

**`raw_results.csv` already carries everything a replicate contributes** — estimates plus
the termination diagnostics the filters read. The file that makes `--summarize` possible
is the same file a resume needs to read.

So the work is not "design a checkpoint format". It is: write that file *incrementally*
rather than at the end, and skip the sample indices already in it.

## What has to change

**1. Incremental writes.** `output::write_all` currently runs once, after every fit is
done — so an interrupted run leaves nothing behind. Each replicate must be appended as it
completes. That means:

- open `raw_results.csv` with its header before the replicate loop and append per result;
- serialize the appends (the replicates run on a Rayon pool, so this needs a mutex or a
  results channel — the fits are seconds each and the write is microseconds, so
  contention is not a concern, but interleaved partial rows would be);
- flush per row, or accept losing the tail row to an unflushed buffer on a hard kill;
- write `included_individuals` / `included_keys` / `sample_keys` the same way, since
  #1141's row-order guarantee across those files must survive a resume.

**2. `--resume`.** Read the existing `raw_results.csv`, collect the sample indices
present, and fit only the missing ones. `output::read_raw_results` already parses the
file; it needs to tolerate a **truncated final row** (a hard kill mid-write) by dropping
it rather than erroring.

**3. What counts as "already done".** A row whose `error` column is non-empty is a
replicate that *ran and failed*. Two defensible answers, and the choice needs stating in
the docs rather than being implied: re-fit it (the failure may have been a transient
resource problem) or keep it (the failure is deterministic and refitting burns time to
reach the same place). PsN's answer is the second — its recovery instructions have the
user delete the `.lst` files of runs they want redone. A `--retry-failed` flag covering
both is probably the right shape.

**4. Consistency guards.** Resuming into a directory written by a *different* run is a
silent corruption: the estimates would be labelled with the wrong model's parameter
names. `prepare_run` already computes `model_hash` and `data_hash`; write them into a
small `bootstrap_run.json` alongside the CSVs together with `seed`, `samples`,
`sample_size` and `stratify_on`, and refuse a resume whose options or inputs do not
match, naming which one differs.

## Interaction with the rest

- **`--dofv`** runs after all replicates, over the whole set. A resumed run recomputes it
  for everything, which is correct and cheap (`outer_maxiter = 0`).
- **`--update-inits`** reads the base fit. A resumed run must not refit the base model if
  its row (`sample = 0`) is already in the file — otherwise the replicates start from a
  different point than the ones already on disk did.
- **`--bca`** (#TBD) has the same shape for its jackknife; `jackknife_results.csv` should
  be resumable the same way.

## Validation

- **Resume equals uninterrupted, bit-for-bit.** Run N samples; run the first K, truncate
  the file to those K rows, resume, and assert the resulting `raw_results.csv` is
  identical to the uninterrupted one. This is the whole claim, and per-replicate seeding
  is what makes it assertable rather than approximate.
- **A truncated final row is dropped, not fatal** — write a valid file, chop the last
  line mid-field, resume, and check that sample is refitted.
- **A mismatched directory is refused**, naming the field that differs (seed, sample
  size, model hash, data hash).
- **Nothing to do**: resuming a complete run refits zero replicates and rewrites the
  summary.

## Acceptance

- [ ] `raw_results.csv` and the three per-replicate files written incrementally
- [ ] `--resume` skips completed sample indices; `--retry-failed` re-runs errored ones
- [ ] `bootstrap_run.json` records the inputs and options; a mismatched resume is refused
- [ ] Base fit reused rather than repeated when its row is present
- [ ] Truncated trailing row tolerated
- [ ] Bit-for-bit resume-vs-uninterrupted test
- [ ] `docs/tools/bootstrap.qmd` gains a recovery section; `CHANGELOG.md`

Refs #1140, #1141.

roninsightrx:
## Why

`PsN::bootstrap`'s user guide documents two recovery situations. #1141 shipped one of them
(`--summarize`: "too many samples were filtered out" — nothing needs refitting, only the
exclusion criteria change). The other was missing: *the process was killed, the machine
crashed, I want to rerun only the samples that failed.* Today that means starting a
200-sample, hours-long run over from sample 1, because `output::write_all` ran once after
every fit was done — an interrupted run left nothing behind at all.

Two properties the #1140/#1141 design already had make this cheap, and make it better than
PsN's version:

- **The draws are a pure function of `(seed, index)`.** `resample::replicate_seed` derives
  each replicate's RNG from the master seed and the replicate's own index, so replicate 137
  of a resumed run draws *exactly* the dataset replicate 137 of the interrupted run would
  have. PsN cannot promise that — its guide documents the RNG as shared and
  order-dependent, so resuming there gives a different set of datasets than not resuming
  would have.
- **`raw_results.csv` already carries everything a replicate contributes** — estimates plus
  the termination diagnostics the filters read. The file that makes `--summarize` possible
  is the file a resume needs.

So this is not a new checkpoint format. It is: write that file incrementally rather than at
the end, and skip the sample indices already in it.

## What changed

**Incremental writes** (`bootstrap/journal.rs`, new). The four per-replicate files —
`raw_results.csv`, `included_individuals1.csv`, `included_keys1.csv`, `sample_keys1.csv` —
are appended as each replicate finishes, under one mutex per replicate so #1141's
row-alignment guarantee survives Rayon, and flushed per row so a hard kill loses nothing
that finished. Row order in the journal is *completion* order; the final `write_all`
rewrites all four in index order, so a completed run's artefacts stay byte-identical
however its replicates were scheduled. The row/header builders are shared with
`output::write_raw_results` rather than duplicated — a divergence there would put columns
out of step with their names, which is the one corruption a reader cannot detect.

**`--resume`.** Reads the existing `raw_results.csv`, collects the sample indices present,
and fits only the missing ones. Reopening the journal *rewrites* the kept rows rather than
appending onto whatever bytes were there, which is what makes a truncated trailing row
self-healing: appending onto a record cut mid-field would splice the fragment and the new
row into one corrupt record.

**Base fit reuse.** A resumed run does not refit the base model when its `sample = 0` row is
present. `--update-inits` starts every replicate from its estimates, and a second base fit
can land marginally elsewhere — refitting it would start the *new* replicates from a
different point than the ones already on disk, and the resumed run would stop equalling the
uninterrupted one.

**`--retry-failed`.** A row with a non-empty `error` column is a replicate that ran and
failed. The default is to keep it — PsN's answer, and the right one when the failure is
deterministic, which it usually is; paying for it again reaches the same place.
`--retry-failed` refits those instead, for a failure that was a transient resource problem
(OOM kill, full disk). The choice is stated in the docs rather than implied, including the
distinction from a merely *filtered* replicate, which `--summarize` handles.

**Consistency guards** (`bootstrap/manifest.rs`, new). Every run writes `bootstrap_run.json`
recording seed, samples, sample size, stratification column, `keep_covariance`/`dofv`, the
parameter names, and `prepare_run`'s SHA-256 model and data hashes. A `--resume` whose
inputs or options disagree is refused, naming the field:

```
--resume: `run12-bootstrap` was created with --seed 1, but this run asks for 42.
Resuming would mix two different runs' replicates into one raw_results.csv.
Use a fresh --directory, or drop the conflicting option.
```

Without it the resumed file would carry one model's estimates under another model's column
names — silent, because the file would still parse and the statistics would still compute.
`keep_covariance` and `dofv` are in there because they change the row's *width*, so
appending under a different setting would corrupt the file outright. The two hashes are
compared only when both sides have one (`prepare_run` leaves them `None` when a file cannot
be hashed); the parameter-name check is unconditional and catches the same class of mistake
when they are absent.

**Truncated-tail tolerance.** `read_raw_results` now drops a final record that is short,
unparseable, or unterminated (no trailing newline — the writer terminates every record it
finished, so a full-width last row without one is still a partial write). The sample is
simply refitted. A malformed row anywhere *else* is still an error: silently dropping one
would change what `--summarize` reports. Fit error text is flattened to one line on write,
so the "one row = one line" property the recovery rule leans on actually holds.

## Alternatives considered

**Sorting the journal by index instead of a final rewrite.** Would need the writer to buffer
out-of-order completions, which is exactly the state a crash destroys. Appending in
completion order and canonicalising at the end keeps the crash-visible file simple and the
finished file deterministic.

**Storing the base fit's estimates in the manifest instead of widening `raw_results.csv`'s
precision.** This started as the fix for the last remaining bit-for-bit gap (see below) and
was rejected: it would have left the *replicates'* estimates rounded, which `--dofv`
re-evaluates from, so the same gap would reappear under `--dofv`. Making the checkpoint
lossless fixes both.

**Refitting errored replicates by default.** Rejected for PsN parity and because it is the
worse default: a deterministic failure burns the same time to reach the same place. Exposed
as a flag instead.

## Cross-repo dependency

No `ferx-core` change at all — this is entirely `ferx-tools` + `ferx-cli`. The public-API
baseline is untouched.

## Breaking changes
- [x] None

One behavioural change worth calling out, though not breaking: **`raw_results.csv` is now
written at full round-trip float precision** (Rust's shortest-decimal-that-parses-back)
instead of a fixed ten decimals. That file is now the run's *checkpoint*: a resumed run
starts its replicates from the base-fit estimates read back out of it, and recomputes
`--dofv` from the replicates' own. Rounding perturbs those inputs in the eleventh digit, and
a fit started from a perturbed point lands somewhere slightly different — the resumed run
matched the uninterrupted one only to ~1e-9 rather than exactly. The statistics files
(`bootstrap_results.csv`, `bootstrap_diagnostics.csv`) keep their fixed ten decimals;
nothing is ever fitted from those.

Also: a *fresh* run now truncates an existing directory's per-replicate files at the start
rather than overwriting them at the end. Same net effect on success; on failure the old run
is gone either way once fitting begins. Documented.

## Numerical validation

- [x] Not applicable — no numerics changed. No estimator, likelihood, sensitivity or PK
  path is touched; this is file I/O and run orchestration.

The validation that matters here is the **bit-for-bit resume-vs-uninterrupted** test rather
than a NONMEM anchor — NONMEM has no bootstrap resume to anchor against, and the claim under
test is an equality between two ferx runs, which is exactly assertable:

```
a_resumed_run_equals_an_uninterrupted_one   (crates/ferx-tools/tests/bootstrap_end_to_end.rs)
  run 3 samples on warfarin        -> raw_results.csv (A)
  run 3 samples again, chop back to the base fit + sample 1, --resume
                                   -> raw_results.csv (B)
  assert A == B, cell for cell, `seconds` column excluded
  assert included_individuals1 / included_keys1 / sample_keys1 byte-identical
```

That is the whole claim of the issue, and per-replicate seeding plus base-fit reuse is what
makes it an equality rather than an approximation. It caught the precision problem described
under Breaking changes: before `fmt_exact`, samples 2 and 3 differed from the uninterrupted
run in the tenth decimal of every estimate and the tenth of the OFV.

## Performance

Four `write_record` + `flush` calls per replicate, against a fit measured in seconds. The
mutex is uncontended in practice for the same reason.

## Tests

Tier 1 (`--lib`, 91 tests in `ferx-tools`, up from 78):

- `bootstrap/manifest_tests.rs` — round trip; **every** manifest field refuses a resume and
  names itself (table-driven over all nine); an absent hash does *not* refuse.
- `bootstrap/journal_tests.rs` — headers; rows stay aligned across the four files when
  appended out of index order; the base fit writes no draw row (a phantom row there would
  offset every replicate after it); reopening rewrites cleanly over a file cut mid-field;
  the column shape follows `--keep-covariance`/`--dofv`; a multi-line error is flattened.
- `bootstrap/output_tests.rs` — a tail cut mid-field, a tail cut on a field boundary, and a
  full-width tail with no trailing newline are each dropped; a short row **in the middle** is
  an error; a header-only file reads as an empty run.
- `bootstrap/mod_tests.rs` — `load_resumable` skips completed indices and reuses the base
  row; a complete run refits nothing; a failure is kept by default and refitted under
  `--retry-failed`; a mismatched seed, disagreeing columns, a missing `raw_results.csv` and a
  missing manifest are each refused; `--no-run-base-model` drops the stored base fit;
  `--resume` without `--directory` and `--retry-failed` without `--resume` are refused.

Tier 2 (`crates/ferx-tools/tests/bootstrap_end_to_end.rs`, real warfarin model,
`outer_maxiter = 2`, no convergence loop — these run on every PR):

- `a_resumed_run_equals_an_uninterrupted_one` (above)
- `a_run_cut_mid_row_resumes_by_refitting_that_sample`
- `resuming_a_complete_run_refits_nothing`
- `resuming_a_directory_from_a_different_seed_is_refused`
- `a_run_records_what_it_was_created_from`

CLI (`crates/ferx-cli/src/bootstrap_cmd.rs`): flag parsing, a real `--resume` round trip
through `run()` including the mismatch exit code, and `--retry-failed` without `--resume`
exiting 1.

`cargo test --workspace --lib --tests` is green (3707 + 91 + the rest, 0 failures).

```bash
# The run dies at sample 137 of 200.
ferx bootstrap run12.ferx --data data.csv --samples 200 --directory run12-bootstrap

# Pick it up. Samples 1..136 are read back from raw_results.csv; 137..200 are fitted.
# The base model is not refitted.
ferx bootstrap run12.ferx --data data.csv --samples 200 --directory run12-bootstrap --resume

# Same, but also re-run the samples whose fit errored.
ferx bootstrap run12.ferx --data data.csv --samples 200 --directory run12-bootstrap \
  --resume --retry-failed
```

```
Resuming the run in run12-bootstrap
Fitted 64 replicates in 812.3s; reused 136 already on disk
```

## Docs

`docs/tools/bootstrap.qmd` gains **Recovering an interrupted run**: a table mapping the three
recovery situations to `--summarize` / `--resume` / `--resume --retry-failed`, why a resumed
run *is* the uninterrupted run (and why PsN cannot say that), the errored-vs-filtered
distinction, the mismatch refusal with its message, and a note on the precision change. The
output-files table gains `bootstrap_run.json` and a paragraph on which files are incremental.
No new page, so `_quarto.yml` is unchanged.

## Checklist
- [x] `cargo clippy` clean (no new warnings from the changed crates)
- [x] N/A — nothing on the `Dual2` path is touched
- [x] N/A — no version bump; no breaking change

### ferx-site / ferx-book
- [x] No new `.ferx` DSL syntax, `[fit_options]` key, example model or data file — this is
  two CLI flags on an existing tool, documented on its existing page.

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` or `data/` file is affected,
  and no ferx-r surface changes.

## Reviewer hints

Focus on three places:

1. **`journal.rs::create`** — the decision to *rewrite* the kept rows on reopen rather than
   append. That is the whole truncation story; if it appended, a file cut mid-field would be
   spliced into a corrupt record and the corruption would be invisible.
2. **`output.rs::read_raw_results`** — the drop-the-tail rule. It reads every record first
   because "may this bad record be dropped?" depends on whether anything follows it, which a
   streaming loop cannot know. Note the asymmetry: dropping is always *safe* for `--resume`
   (the sample is refitted) but not for `--summarize` (the statistics change silently), which
   is why only the tail is tolerated.
3. **`mod.rs::run_bootstrap`** ordering — draws before resume before journal before base fit.
   `load_resumable` must read `raw_results.csv` before `Journal::create` truncates it.

Skimmable: the CLI flag wiring, the docs prose, and `manifest.rs::check_compatible`, which is
nine mechanical field comparisons behind one message builder.

## Open questions

None blocking. One deliberate deferral: `--bca` (#TBD) will need its `jackknife_results.csv`
resumable the same way. The journal is written so that adding a fifth file is a field on
`Files` plus a row builder, but nothing here anticipates it beyond that, since the jackknife
does not exist yet.


crates/ferx-tools/src/bootstrap/manifest.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub stratify_on: Option<String>,
+    /// SHA-256 of the model file, as `ferx_core::PreparedRun` computed it.
+    pub model_hash: Option<String>,
+    /// SHA-256 of the dataset.
+    pub data_hash: Option<String>,
+    /// Recorded because it changes the *column shape* of `raw_results.csv`:
+    /// appending rows of a different width would corrupt the file.
+    pub keep_covariance: bool,
+    /// Recorded for the same reason — `--dofv` adds a column.
+    pub dofv: bool,
```
roninsightrx, line 40: [P1] Pin the initialization mode in the manifest

`update_inits` and `run_base_model` are absent from this manifest, but they determine `replicate_init` in `run_bootstrap`. For example, interrupt a default run and resume it with `--no-update-inits`: `check_compatible` accepts the directory, the reused rows were started from the stored base fit, and every missing row is started from the model-file estimates. The inverse transition has the same problem; changing `--no-run-base-model` can also toggle `update_inits` in the CLI. The resulting `raw_results.csv` silently mixes two fitting regimes, so it is no longer the uninterrupted run this feature promises. Please record and compare these options (or otherwise reject such changes while any samples remain), and add the mismatch regression cases.

CHANGELOG.md
```diff
@@ -54,6 +54,21 @@ section of the SDLC for the versioning policy).
...
+  the replicate's own index, a resumed run reproduces the uninterrupted one exactly — not merely
+  statistically, which is what PsN's shared, order-dependent RNG can offer. A replicate whose fit
+  *errored* is carried forward by default, as in PsN; `--retry-failed` refits those instead, for a
+  failure that was a transient resource problem. Every run now writes `bootstrap_run.json`
+  recording its seed, options and the hashes of the model and data files, and a `--resume` whose
+  inputs or options disagree is refused, naming the field. A trailing row cut off by a hard kill is
+  dropped and that sample refitted. `raw_results.csv` now carries full round-trip precision rather
+  than ten fixed decimals, because a resumed run's replicates start from the base-fit estimates read
+  back out of it; the statistics files are unchanged. See `docs/tools/bootstrap.qmd`.
   replicates would mix them; use SIR for uncertainty on a mixture model instead (#1145).
```
roninsightrx: [P2] Repair the displaced changelog text

The merge left the previous bootstrap bullet split: this dangling `replicates would mix them; use SIR...` sentence now appears after the new `--resume` entry, while lines 56–57 already end that earlier mixture discussion. Please restore the full sentence to the original bootstrap bullet and remove these duplicate dangling lines so the Unreleased changelog reads coherently.

roninsightrx:
Review complete: changes requested. The blocking issue is that the resume manifest does not pin `update_inits` / `run_base_model`, allowing an interrupted run to silently mix replicates started from the base fit with replicates started from the model-file estimates. I also left a small changelog merge-artifact comment.

Focused verification passed:
- `cargo test -p ferx-tools --lib bootstrap::` (90 passed)
- `cargo test -p ferx-cli bootstrap_cmd::` (22 passed)

roninsightrx:
Both findings fixed. Thanks — P1 was a real hole.

## P1 — pin the initialization mode (`0c223ab2`)

Confirmed and fixed. `update_inits` and `run_base_model` do determine `replicate_init`,
and neither was in the manifest, so the transition you describe passed every check and
produced a file whose reused replicates started at the base fit and whose refitted ones
started at the model file's estimates. Both halves are valid fits; the file holding them is
a bootstrap of neither, and nothing about it looks wrong — which is the class of defect the
manifest exists for.

`RunManifest` gains two fields:

- `run_base_model: bool` — pinned in its own right, because it also decides whether a
  `sample = 0` row belongs in the file at all.
- `replicate_inits: String` — `"base_fit"` or `"model_file"`.

The second is recorded as the one **effective** mode rather than as the raw `update_inits`
flag, because that is what `run_bootstrap` actually branches on:

```rust
let replicate_init = match (&original, options.update_inits) {
    (Some(base), true) => params_from_estimates(template, &base.estimates),
    _ => template.clone(),
};
```

`update_inits` has no effect without a base fit, so a caller that sets it alongside
`run_base_model = false` is asking for the model file's estimates just as plainly as one
that left it off. Comparing the two raw flags would refuse that resume, which changes
nothing — `describe_replicate_inits` collapses them to the mode instead, and
`the_recorded_initialization_mode_is_the_effective_one` pins all four combinations plus the
no-op-resume case.

Messages name the flag to change:

```
--resume: `run12-bootstrap` was created with replicates started from the base fit's
estimates (--update-inits), but this run asks for the model file's estimates
(--no-update-inits). Resuming would mix two different runs' replicates into one
raw_results.csv. Use a fresh --directory, or drop the conflicting option.
```

Regression cases added at both levels:

- `mod_tests.rs::a_resume_that_moves_the_replicate_starting_point_is_refused` — your exact
  scenario **and** its inverse (a run started from the model file, resumed asking for the
  base fit).
- `mod_tests.rs::a_resume_that_adds_or_drops_the_base_model_is_refused`.
- `manifest_tests.rs` — both new fields joined the table-driven "every field refuses by
  name" case list.
- `bootstrap_end_to_end.rs::resuming_with_a_different_replicate_starting_point_is_refused`
  — end to end through `run_bootstrap`, covering both flags.

One consequence worth flagging: the old
`a_resume_without_the_base_model_drops_the_stored_base_fit` test asserted the *tolerant*
behaviour you are asking me to reject, and it now fails by construction. It is replaced by
the refusal case above, plus
`a_run_without_a_base_model_drops_a_stored_base_fit` — a directory genuinely written
without a base model, whose matching manifest lets `load_resumable`'s
`.filter(|_| options.run_base_model)` still fire. That filter is now unreachable through
the CLI (the manifest refuses first) and is documented as defence for a library caller that
builds its own manifest.

Docs and the changelog entry updated to say the starting point is pinned and why.

## P2 — displaced changelog text (`8d5b3e23`)

Right, and it took a moment to see why: the artifact was not on my branch. `#1145` landed
on `main` after I branched and extended the *last sentence* of the bootstrap bullet —
exactly the line my new bullet is inserted after — so the artifact only appears once `main`
is merged in. You were reading the merged view; my tree read fine, which is why I checked
twice before believing you.

Worth noting: `origin/…-1143-bootstrap-resume` had already picked up a merge of `main`
(`e975e852`) that resolved this hunk the **wrong** way, keeping
`replicates would mix them. See docs/…` and dropping `#1145`'s `use SIR for uncertainty on
a mixture model instead` clause. So the artifact was live on the remote branch, not
hypothetical. Both merges are now resolved keeping `main`'s completed sentence with the new
bullet after it; `git diff origin/main...HEAD -- CHANGELOG.md` is a clean 16-line insertion
with nothing else touched.

## Verification

- `cargo test --workspace --lib --tests` — 0 failures
- `ferx-tools` lib: 94 (was 91); `bootstrap_end_to_end`: 26 (was 25)
- `cargo fmt --all --check` clean; no new clippy warnings from the changed crates

roninsightrx:
## Code review — `--resume` (medium effort)

Scope: `crates/ferx-tools/src/bootstrap/{journal,manifest,output,mod}.rs`, `crates/ferx-cli/src/bootstrap_cmd.rs`, docs. Reviewed this PR's own commits only (the `main...HEAD` range also carries merged-in `main` work).

The design holds up: draws precede resume precedes journal-create precedes base fit; `load_resumable` reads `raw_results.csv` before `Journal::create` truncates it; rewrite-on-reopen makes a mid-field tail self-healing; row alignment across the four files is restored on every reopen. The second commit already closed the `--update-inits` / `--run-base-model` manifest gap.

Three findings.

**1. `bootstrap/journal.rs:108` — medium. `create` truncates in place, so the recovery data lives only in memory during the rewrite.**

All four files are `truncate`d before the kept rows are rewritten. If `append_writer` fails on the 2nd–4th file (full disk, permissions, EMFILE), `create` returns `Err` with `raw_results.csv` already zeroed; a kill during the rewrite of e.g. 190 kept rows loses all 190 — the exact failure this feature exists to survive. Suggest writing the rewritten set to a sibling temp file and `rename`ing it into place (atomic on the same filesystem) instead of truncating in place.

**2. `bootstrap/journal.rs:126` — medium. Headers are written but never flushed, so `raw_results.csv` is 0 bytes until the first replicate finishes.**

The `raw_results.csv` header and the `sample_keys1.csv` subject-ID header go through `write_record` in `create`; `csv::Writer` buffers and the first `flush()` is in `append` (journal.rs:184). Kill the run during the base fit — often the longest single fit — and a later `--resume` does not reach the friendly "has no raw_results.csv, drop --resume" path (the file exists), it hard-errors out of `read_raw_results`:

```
does not look like a bootstrap raw_results file: expected column 1 to be `sample`, found `<missing>`
```

and the user has to delete the directory. A `flush()` after the two header writes fixes it.

**3. `docs/tools/bootstrap.qmd:205` — low. Recovery table row contradicts the manifest guard.**

"some samples failed for a reason you have since fixed → `--resume --retry-failed`": if the fix was to the model file or dataset, `model_hash`/`data_hash` refuses the resume outright. Only a fix *outside* the inputs is actually resumable — the prose two sections down says so, the table row does not. Suggest "…for a transient reason (OOM, full disk) that has since gone away".

PR #1152, opened 2026-08-31
fix(pk): check the disposition off the axes, not only along them [closes #1149]

### LINKED ISSUE #1149: bug(pk): identify_disposition's probe is unsound on the state axis — a bilinear term or a state-triggered branch is silently dropped from the closed form
## Summary

`identify_disposition` / `probe_rhs_f64` decide a **structural** property — is this system linear and time-invariant, so superposition can represent it? — by evaluating the `[odes]` RHS at a handful of points. #1124 fixed the *time* axis of that probe by gating on a syntactic predicate before it runs. The **state** axis is unsound in the same way and is still unguarded.

The probe evaluates only at `u = 0`, `u = eᵢ`, and `u = 2·eᵢ` — the coordinate axes, one state at a time, at a fixed amplitude of 1–2 model mass units. Three consequences, all with the same shape as #1124: a term is silently absent from the predictions, with no error, no warning, and `converged: true`.

## (a) A product of two distinct states is exactly zero at every probe point

```
[structural_model]
  ode(states=[central, periph])
[odes]
  d/dt(central) = first_order(ka=KA) - CL/V*central - K12*central + K21*periph - KON*central*periph
  d/dt(periph)  = K12*central - K21*periph
```

Probes at `(1,0)`, `(2,0)`, `(0,1)`, `(0,2)` all give `KON*central*periph = 0`. `f(0) = 0` passes, `k12`/`k21` both clear `COUPLING_EPS`, and the mass-conservation and sign checks pass. A clean canonical 2-cpt is identified and `Σ FRₖ·kernelₖ(t − LAGₖ)` is served with the binding term **dropped entirely**.

No time variable appears anywhere, so `reads_model_time()` is `false` and `mr_scope` admits it. This is the TMDD / target-binding shape, which is not exotic.

## (b) A branch on the *state* is never entered

`Statement::If` is fully supported in `[odes]`, and the probe only ever evaluates at `0`, `eᵢ`, `2eᵢ`:

```
[odes]
  if (central > 10) { KE = 3*CL/V } else { KE = CL/V }
  d/dt(central) = first_order(ka=KA) - KE*central
```

Every probe takes the `else` arm, a 1-cpt with `ke = CL/V` is identified, and the closed form under-predicts elimination 3× across the therapeutic range.

Note #1124's new test `mr_scope_declines_every_model_time_spelling` exercises exactly this branch shape — but only with a time variable inside it. The state version is a one-token edit away from the tests we ship.

## (c) Linearity is decided at a scale unrelated to the subject's amounts

`f(2·eᵢ) = 2·f(eᵢ)` is a homogeneity test at 1–2 *model mass units*, which has nothing to do with the doses the subject actually receives. Saturable elimination `- VM*central/(KM+central)` is therefore admitted whenever it is still linear at `u ∈ [1,2]`, which for a physiologically sized `VM/KM` means roughly `KM ≳ 1.7e6` model units — absurd in mg, entirely ordinary in ng or pmol, which is how biologic/TMDD models are written. Rewriting the same model in different units flips admission, so correctness depends on the user's choice of unit.

## Why this matters more than the individual cases

The oracle fails **open**: on an input it cannot characterise it constructs a wrong closed form and serves it. Compare `equilibrate_ss_input_rate`'s fixed-point self-check, which fails **safe** — it returns `None` and falls back to Anderson and then to an explicit pulse train. `mr_scope` returning `None` costs an ODE integration; admitting wrongly costs an 8× prediction error and a fit that converges to nonsense.

The current defence is a blacklist in front of the probe (`model_uses_time_anywhere`), which only ever covers the failure classes someone has already found. #1124 was one. The three above are three more.

The `verify_against_ode_twin` net does not cover this: it is `#[cfg(debug_assertions)]` (see #344, so it never runs in CI or in a user's fit), and at the shipped default `ode_reltol = 1e-4` its bound is 5% relative — (a) and (b) both slip that at ordinary parameter values.

## Suggested direction

Add an **off-axis, dose-scaled residual check** to the probe rather than another blacklist entry: one extra RHS evaluation at a mixed state whose amplitude comes from the subject's own dose (`max(amt)·F`), asserting `‖f(u) − A·u − b‖ ≤ tol·‖A·u‖` against the `A` just identified. That is O(1), works in release, and catches (a), (b) and (c) as one class instead of three special cases.

Worth pairing with a decision on whether `mr_scope` should fail safe in general — i.e. whether a model that cannot be *positively* shown affine should be declined rather than admitted.

Found during the #1130 review (#1124).

TeunP:
## Why

`identify_disposition` decides whether an `[odes]` right-hand side is a linear 1-/2-compartment disposition — and so whether the closed-form fast path may serve it, skipping integration entirely. It establishes linearity **behaviourally**, by evaluating the right-hand side at `u = 0`, `eᵢ` and `2eᵢ`.

Every one of those samples holds all *other* amounts at zero, and none exceeds 2. A term that only switches on away from those points is therefore invisible to every check the function runs, and the model is admitted to a path that never evaluates the right-hand side at all — so the term is **dropped from the predictions**, silently, with `converged: true`. That is #1124's failure shape with no time variable in it, which is why `mr_scope`'s model-time gate does not apply.

Three classes are reachable today:

- **a product of two distinct amounts** — `- KON*central*periph`, the target-binding/TMDD shape — is *exactly* zero whenever either amount is zero. `f(0) ≈ 0` passes, `f(2eᵢ) = 2f(eᵢ)` passes, both couplings clear `COUPLING_EPS`, the sign and mass-conservation patterns pass, and a clean canonical 2-cpt is identified with the binding term gone.
- **a branch on an amount** — `if (central > 10) { KE = 3*CL/V }` — is never entered, so the `else` arm is identified as the whole model.
- **saturable elimination with a large `KM`** is genuinely linear at amounts of 1–2, so it passes the homogeneity check while being saturated at the amounts a real dose produces. This one also made admission depend on the model's **units**: the same model written in ng rather than mg has a `KM` a million times larger, so it flipped from declined to admitted on a change of units alone.

Found while reviewing #1130 (#1124), filed as #1149.

## What changed

After the axis probes fix `A`, check the residual `‖f(u) − A·u‖` at mixed states across six decades of amount. `f(0) ≈ 0` is already established, so a genuinely linear disposition satisfies `f(u) = A·u` at every `u`.

Two mechanisms, deliberately separate — for a one-compartment model there is only one axis, so "off-axis" is vacuous there and the amplitude is doing all the work:

- **mixed states** (`n ≥ 2`, every amount non-zero at once) reach the state-product class, which no single-axis sample can;
- **amplitude** reaches anything whose behaviour changes with the amount — a threshold, or saturation.

The tolerance is `LINEARITY_TOL`, the same constant the on-axis homogeneity check uses, and for the same reason: this compares two *evaluations* of one right-hand side, with no integration in it, so for a linear system the residual is exact to rounding. It is **not** calibrated against solver error and must not be confused with `verify_against_ode_twin`'s bound, which compares two integrations and has to scale with `reltol`.

The ladder spans decades rather than adding one off-axis point because a threshold sits at whatever amount the model author wrote, and identification has no subject to derive a dose scale from. A linear right-hand side passes at every amplitude, so extra rungs cannot cause a false decline. Declining is the safe direction: the model routes to the ODE engine, which is slower and correct.

## Alternatives considered

**Pass the subject's dose scale in** (the shape #1149 suggested). `identify_disposition` is `pub` and takes `(spec, p, t)`; threading an amplitude through means either a signature change or a second public entry point. The ladder gets the same coverage without touching the public API, and is strictly broader — a dose-scaled probe still misses a threshold below the dose.

**Declining anything with an `if` in the right-hand side.** Far blunter: it would decline models whose branch is on a covariate or a parameter, which are perfectly linear in the amounts and are the common case for a class-switched baseline.

**Leaving it to `verify_against_ode_twin`.** It does catch this class — with the new check disabled, the gate fires on the bilinear model at the second observation of the grid below (`t = 2`, closed form 4.270073 vs twin 4.250366). But it is `#[cfg(debug_assertions)]`, and both profiles CI builds tests with — `ci-test` and `ci-fast` — inherit `release` (#344), so it runs in neither CI nor a user's fit. It is a development aid, not a guard.

## Numerical validation

Reference is the ODE engine, which the `mr_tad` / `tad_lag_{A,B}` NONMEM anchors already pin. Two doses (0 and 12 h), `KON = 0.002`, `ode_reltol = 1e-10`, measured with the new check disabled so the closed form is still admitted:

| t | closed form | ODE truth | rel. err |
|---|---|---|---|
| 0.5 | 2.583358 | 2.583093 | 1.0e-4 |
| 2 | 4.270073 | 4.250366 | 4.6e-3 |
| 8 | 2.926096 | 2.633272 | 1.1e-1 |
| 12 | 2.212366 | 1.775333 | 2.5e-1 |
| 24 | 3.249428 | 1.687115 | **9.3e-1** |

Worst 9.3e-1 at 24 h — **1.93× overprediction**, growing monotonically, at a modest binding rate. With the check in place the model declines and production returns the ODE values.

Re-measured after the rebase onto `main`; both value columns reproduce bit-for-bit. The error column previously read `|Δ|/(1+|ODE|)` (the codebase's assertion form) under a `rel` header, which made it disagree with the `1.93×` in the same sentence — it is now true relative error, `|Δ|/|ODE|`, and the two agree.

`--lib`: 3843 passed, 0 failed (3840 on `main` + this PR's 3). **No model in the existing fixture zoo was newly declined** — that is the load-bearing result for a change that can only make admission stricter. Concretely it is pinned by ~19 admission assertions: 9 `assert_reduces_to_ode` call sites (each `.expect("MR-supported")`), 8 `identify_disposition(...).is_some()` asserts, and 2 further `mr_predictions(...).expect(...)` sites. A newly-declined fixture returns `None` and fails at that `expect`.

## Performance

Identification goes from `3n + 1` right-hand-side evaluations to `3n + 5` — 7 to 11 for a two-compartment model. `MrDisposition`'s own doc already calls that probing a second-order cost (the fast path avoids ODE integration entirely) and notes it repeats on every value/gradient evaluation because the result is not yet memoised; this makes that pending memoisation worth slightly more. Models that newly decline pay the ODE path, which is the point.

## Tests

Three decline tests, each **paired with a control that must still be admitted**, so the assert pins the offending term rather than the model's shape, and each asserting **non-vacuity** — that the term really is invisible on the axes:

- bilinear: axis probes bit-identical to the `KON = 0` control;
- state branch: axis probes at `u = 1` and `u = 2` bit-identical to the unbranched control;
- large `KM`: the on-axis homogeneity residual is asserted to be *inside* `LINEARITY_TOL`, so the test cannot degenerate into a copy of the existing `KM = 5` case.

Mutation-tested: disabling the residual check fails all three.

## Docs

New section in `docs/model-file/ode-models.qmd`, parallel to the existing model-time one, with a warning callout for results from 0.3.0 and earlier. `CHANGELOG.md` `[Unreleased]` entry added.

The new callout is titled **"Nonlinear disposition: affects results from ferx 0.3.0 and earlier"** rather than repeating the model-time callout's title verbatim. This PR predates the docs-lint gate (#1163), and on rebase R3 caught the duplicate — two headings on one page generating the same id means the second is addressed as `…-1`. Disambiguating also tells a reader which of the two callouts they are in. Nothing links to either anchor, so no inbound reference breaks.

## Rebase

Originally branched on `worktree-fix-1124-mr-tad-drop` (#1130), which is now merged (`764784c8`). Rebased onto `main` and retargeted; the code diff is byte-identical, and the only content change is the docs heading above.

Rebasing also cleared this PR's `verify_against_ode_twin` failure on the `mr_tad` per-route-lag anchor. That was never this PR's defect: it was the tolerance artifact fixed in #1130 by `8cf01026` and `8528baab`, which the old base predated.

This is deliberately *not* part of #1130: that PR decides **routing** (which engine serves a model), this one changes **admission** (whether the closed form is valid at all). Different blast radii, and #1130 had already been through two review rounds — folding a new admission rule into it would have given the newest behaviour the least scrutiny.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,19 @@ section of the SDLC for the versioning policy).
   top of them, so a tool and the CLI cannot diverge in how a model is loaded.
 
 ### Fixed
+- **An `[odes]` right-hand side that is nonlinear in the compartment amounts is no longer
+  served by the closed-form fast path (#1149).** Linearity was established by evaluating the
+  right-hand side one compartment at a time with every other amount held at zero, so a term
+  that only switches on away from those points was invisible to it: a product of two amounts
+  (`- KON*central*periph`, the target-binding/TMDD shape) is exactly zero whenever either is
+  zero, and a branch testing an amount (`if (central > 10)`) was never entered, because the
+  largest amount probed was 2. Such models were admitted to the closed form, which does not
+  evaluate the right-hand side at all, so the term was **dropped from the predictions** —
+  the same silent failure as #1124, with no time variable involved. Identification now also
+  checks the right-hand side against the recovered linear system away from the axes, across
+  six decades of amount. Models that are linear in the amounts are unaffected and keep the
+  fast path, including those whose coefficients are arbitrary functions of `θ`, `η`, and
+  covariates.
 - **An `EVID=2` record no longer aborts a fit on a parameter-static subject (#1124 review).**
   Any subject routed to the event-driven walker while its PK parameters are constant — an
   `EVID=3/4` reset, or (since the change above) an `[odes]` right-hand side that reads model
--- a/docs/model-file/ode-models.qmd
+++ b/docs/model-file/ode-models.qmd
@@ -277,6 +277,49 @@ concentration by 12 h, and each of the four spellings read from an untaken `if` 
 wrong by 4.0e-1 relative. Models that do not mention model time in `[odes]` are unaffected.
 :::
 
+### A right-hand side that is nonlinear in the state is always integrated {#nonlinear-rhs}
+
+The same shortcut also needs the disposition to be **linear in the compartment amounts**, so
+it does not apply to a right-hand side where a state multiplies another state, or where a
+branch tests one:
+
+```
+[odes]
+  # nonlinear: integrated, never served by the closed form
+  d/dt(central) = first_order(ka=KA) - CL/V*central - Q/V*central + Q/V2*periph
+                  - KON*central*periph          # target binding (TMDD)
+  d/dt(periph)  = Q/V*central - Q/V2*periph
+```
+
+```
+[odes]
+  # state-triggered branch: also integrated
+  if (central > 10.0) { KE = 3.0*CL/V } else { KE = CL/V }
+  d/dt(central) = first_order(ka=KA) - KE*central
+```
+
+Saturable elimination (`- VMAX*central/(KM + central)`) has always been integrated for the
+same reason. As with model time, there is nothing to configure — the only consequence is
+that these models step the system rather than evaluating a closed form. If they are also
+stiff, see [Stiff systems](#stiff-systems) below.
+
+::: {.callout-warning}
+## Nonlinear disposition: affects results from ferx 0.3.0 and earlier
+
+Until [#1149](https://github.com/FeRx-NLME/ferx-core/issues/1149) ferx established linearity
+by evaluating the right-hand side one compartment at a time, with every *other* amount held
+at zero. A term that only switches on away from those points was therefore invisible:
+`KON*central*periph` is exactly zero whenever either amount is zero, and a threshold above
+the probe's amplitude was never crossed. Such a model was admitted to the closed-form path,
+which does not evaluate the right-hand side at all, so the term was **dropped from the
+predictions** — the same silent failure as the model-time case above, and with no time
+variable involved.
+
+Re-run any earlier result whose `[odes]` block multiplies two compartment amounts together,
+or branches on one. A right-hand side that is linear in the amounts is unaffected, including
+one whose *coefficients* are arbitrary functions of `θ`, `η`, and covariates.
+:::
+
 ### Stiff systems
 
 An explicit method like RK45 is *stability*-limited on a stiff system: it keeps shrinking
--- a/src/pk/modified_release.rs
+++ b/src/pk/modified_release.rs
@@ -113,6 +113,22 @@ const LINEARITY_TOL: f64 = 1e-7;
 /// two-compartment `k12`/`k21` pair from a one-way depot→central drain).
 const COUPLING_EPS: f64 = 1e-9;
 
+/// State amplitudes for [`identify_disposition`]'s off-axis residual check
+/// (#1149). Spans six decades because a state-triggered branch fires at whatever
+/// amount the model author wrote, and identification has no subject to derive a
+/// dose scale from. A genuinely linear RHS satisfies `f(u) = A·u` at every
+/// amplitude, so each extra rung costs one RHS evaluation and cannot cause a
+/// false decline.
+///
+/// Cost, stated rather than waved at: identification goes from `3n + 1` RHS
+/// evaluations to `3n + 5` — 7 to 11 for a two-compartment model, ~57% more
+/// probing. [`MrDisposition`]'s own doc already calls that probing a
+/// second-order cost (the fast path avoids ODE integration entirely) and notes
+/// it repeats on every value/gradient evaluation because the result is not yet
+/// memoised; this makes that pending memoisation worth a little more, and
+/// changes nothing about its correctness.
+const OFF_AXIS_PROBE_SCALES: [f64; 4] = [1.0, 1e2, 1e4, 1e6];
+
 /// Probe the forcing-free disposition RHS `du = f(u, p, t)` in `f64` at state
 /// `u`. Returns `None` when the spec has no compiled RHS program (hand-built
 /// specs / out of analytic scope). The forcings are applied by a separate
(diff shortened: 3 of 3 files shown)


TeunP:
## Review — max effort, no happy paths

Reviewed at `4e2892fa` (rebased onto `main` after #1130 merged as `764784c8`). Ten finder angles, adversarial verification of every candidate, plus a sweep pass. **No correctness defect survived verification in the production change.** Two real defects were found outside it — both in claims, not code — and both are now fixed. Details below, including the one candidate I raised and then had to withdraw.

### The production change holds up

The off-axis residual asks whether `f(u) = A·u` away from the axes, at four amplitudes spanning six decades. Two things make it safe rather than merely stricter:

- **`f(0) ≈ 0` is already established upstream**, by the `du0` guard at `modified_release.rs:200`. That matters more than it looks: the residual `f(u) − A·u` is only a linearity test *given* no affine term, so the guard is the precondition the new check silently depends on.
- **The comparison is exact to rounding.** It is two evaluations of the same RHS with no integration between them, so `LINEARITY_TOL` governs it the same way it governs the existing on-axis `f(2eᵢ) = 2f(eᵢ)` check. Nothing here is calibrated against solver error — which is the trap `verify_against_ode_twin` fell into and #1130 had to fix.

### A candidate I raised and then withdrew

I flagged a suspected **false decline** for turnover/indirect-response models — `d/dt(central) = kin − kout*central` has a constant production term, so `f(u) − A·u = kin ≠ 0` at every amplitude and the new loop would decline it.

**Refuted.** Such a model never reaches the off-axis loop: `f(0) = kin`, and the `du0` guard at `:200` returns `None` on `d.abs() > LINEARITY_TOL` first. The decline is pre-existing, not new. Recording it because the reasoning is the kind that looks conclusive and is not, and the guard that refutes it is 78 lines upstream of the code under review.

### Tests: mutation-verified, in both directions

All three decline tests fail when the residual loop is neutered (`&OFF_AXIS_PROBE_SCALES[..0]`), so none is vacuous. Each is paired with a control that must stay **admitted**, so the assert pins the offending term rather than the model's shape — and each asserts non-vacuity, i.e. that the term really is invisible on the axes. The `KM = 1e5` case is the strongest of the three: it asserts the on-axis homogeneity residual is *inside* `LINEARITY_TOL`, so it cannot silently degenerate into a copy of the existing `KM = 5` test.

### Two defects found, both in claims

**1. The validation table's error column was mislabelled.** It reported `|Δ|/(1+|ODE|)` — the codebase's assertion form — under a header reading `rel`. That made the summary sentence self-contradictory: "Worst 5.8e-1 at 24 h — **1.93× overprediction**" cannot both hold, because 1.93× *is* 9.3e-1 relative error, not 5.8e-1. Corrected to true relative error; the `1.93×` itself re-measured and confirmed.

**2. The docs page tripped R3 on rebase.** The new callout repeated the #1124 callout's title verbatim, so both generated `#affects-results-from-ferx-0.3.0-and-earlier` and the second was addressed as `…-1`. This PR predates the docs-lint gate (#1163), so it could not have been caught when written. Retitled; nothing links to either anchor.

### Re-measured, not inherited

Every number in the PR body predated ~100 commits of `main`, so all of it was re-run at `4e2892fa` rather than carried over:

- Both value columns of the divergence table reproduce **bit-for-bit** — closed form `2.583358 / 4.270073 / 2.926096 / 2.212366 / 3.249428`, ODE `2.583093 / 4.250366 / 2.633272 / 1.775333 / 1.687115`. Fold at 24 h = 1.9260.
- `--lib`: **3843 passed, 0 failed** (3840 on `main` + this PR's 3). The "no fixture newly declined" claim is the load-bearing one for a change that can only make admission stricter, and it is worth saying what actually pins it: ~19 admission assertions — 9 `assert_reduces_to_ode` call sites, 8 `identify_disposition(...).is_some()`, 2 further `expect("MR-supported")`. A newly-declined fixture returns `None` and fails at the `expect`.
- The `#[cfg(debug_assertions)]` argument was re-checked against `Cargo.toml`, not assumed: `ci-test` and `ci-fast` both `inherit = "release"`, so the twin gate runs in neither CI nor a user's fit. It is a development aid, and this check is the guard.

### Scope note

This PR does **not** fix the `mr_tad` per-route-lag twin divergence that made it red on its old base. That was the tolerance artifact fixed in #1130 (`8cf01026`, `8528baab`); the old base predated both. Saying so explicitly because the failure disappearing on rebase could otherwise be read as this PR having fixed it.

Issue #751, opened 2026-07-08
slow-tests red on main since 2026-07-02: #657 (SLSQP-fallback removal) exposed non-convergence that #639 depended on


TeunP:
## Summary

`slow-tests.yml` has been **red on `main` since 2026-07-02 12:06** — the same morning it was green at 11:23. **10 targets fail** on every run since (e.g. [run 28947981633](https://github.com/FeRx-NLME/ferx-core/actions/runs/28947981633)). These are **real regressions, not stale tests**: the affected test files were unchanged when they broke, and the fits reach the correct optimum but the `converged` flag and the covariance SEs come out wrong.

## Bisect

| | Run | HEAD | Result |
|---|---|---|---|
| last green | [28586286895](https://github.com/FeRx-NLME/ferx-core/actions/runs/28586286895) @ 11:23 | `b213c16b` (merge #647) | ✅ |
| first red | [28588678886](https://github.com/FeRx-NLME/ferx-core/actions/runs/28588678886) @ 12:06 | `3039d04a` (merge #666) | ❌ 9 targets |

The only change between them is **#657 (`e6038046`): "remove automatic SLSQP fallback after non-convergence"** (−302 lines from `outer_optimizer.rs`). `imp_init_defensive_slow` joined one merge later (#665), for **10 total**.

## Root cause

Two changes from the `#486`/`#639` epic interacted:

1. **#639** (`7e344946`, merged 07-01, *green*) made the default optimizer `Auto → analytic-gradient NLopt L-BFGS` for analytic FOCE/FOCEI, and reduced the covariance R-matrix to a **FD-of-OFV-only stencil that is only well-conditioned at a true stationary point**. It stayed green **only because the SLSQP fallback was still present** — when L-BFGS didn't cleanly converge, SLSQP re-ran and reached a tight optimum (setting `converged=true` and giving a well-conditioned covariance). Its commit message even claims `ode_analytical_convergence` is "fixed" and the covariance anchors "keep passing" — true *only with the fallback*.
2. **#657** removed that fallback. When NLopt L-BFGS returns a generic `Failure` status **at the optimum** (line-search stalls once ∇≈machine-noise), there is no rescue.

[...]

PR #1156, opened 2026-08-31
feat(parser): [covariate_model] block — declarative covariate relationships at PsN scm parity [closes #1111]

### LINKED ISSUE #1111: feat(dsl): [covariate_model] block — declarative covariate relationships with PsN scm parity
## Summary

Add an optional `[covariate_model]` block to the ferx DSL that declares covariate–parameter
relationships as **structured data** rather than as free-text expressions inside
`[individual_parameters]`.

Today a covariate effect is written classically:

```
[individual_parameters]
  CL = TVCL * (WT/70)^THETA_CL_WT * exp(ETA_CL)
```

This stays fully supported. The new block is sugar over it:

```
[covariates]
  WT   continuous
  CRCL continuous
  SEX  categorical(levels = [0, 1])

[covariate_model]
  CL ~ WT   power(center = median)
  CL ~ CRCL linear(center = 100)
  V  ~ WT   power(center = 70, fix = 0.75)
  CL ~ SEX  categorical(ref = mode)
```

### Why

1. **Ergonomics** — adding a covariate becomes one line, with no need to hand-write the
   centering, the theta declaration, the missing-data guard, or sensible bounds.
2. **Programmatic covariate analysis** — a covariate search (SCM / grid search) rewrites
   **one block only**; the rest of the model file is invariant. Relations are line-oriented
   with no cross-line dependencies, so add/remove is a pure line insert/delete.
3. **Agentic covariate analysis** — same reason, plus the relations are echoed on the fit
   result as a table, so downstream tooling never has to regex a `.ferx` file to learn what
   covariate model was run.

This closes the loop on the existing screening layer in ferx-r (`ferx_cov_screen()`,
`ferx_eta_cov()`), which currently dead-ends at "here is what is worth testing" with no
machine-readable path to actually test it.

## Scope: PsN scm parity

Checked against PsN's
[`config_all_default_parameterizations_explicitly.scm`](https://raw.githubusercontent.com/UUPharmacometrics/PsN/master/doc/config_all_default_parameterizations_explicitly.scm).
All five scm states must be expressible:

| PsN state | PsN code | ferx form |
|---|---|---|
| 1 `none` | `PARCOV = 1` | `none` |
| 2 `linear` (continuous) | `1 + THETA(1)*(COV - median)` | `linear(center = …)` |
| 2 `linear` (categorical) | `IF(COV.EQ.mode) PARCOV=1; ELSE PARCOV = 1 + THETA(1)` | `categorical(ref = …)` |
| 3 `hockey-stick` | two slopes, breakpoint at median | `hockey(breakpoint = …)` |
| 4 `exponential` | `EXP(THETA(1)*(COV - median))` | `exponential(center = …)` |
| 5 `power` | `(COV/median)**THETA(1)` | `power(center = …)` |
| arbitrary `[code]` | NONMEM verbatim | `expr("…")` escape hatch |

PsN's `[test_relations]`, `valid_states`, `included_relations`, `p_forward` are **search
configuration, not model** — they belong in an SCM harness (ferx-r or an agent) that reads the
structured relation table and rewrites `[covariate_model]` lines. The DSL block must not try to
express the search itself.

## Grammar

```
PARAM ~ COV form(kwargs) [=> THETA_NAME(init, lower, upper)]
```

One relation per line. The `=>` clause is optional; when omitted the theta is auto-named
`THETA_<PARAM>_<COV>` (`_LO` / `_HI` suffixes for `hockey`, `_<LEVEL>` for `categorical`) and
gets the form's default init/bounds. An explicit `theta` declaration in `[parameters]` with a
matching name also takes precedence.

### Forms

| form | factor | notes |
|---|---|---|
| `none` | `1` | declares no theta; lets a search write "tested, rejected" into the file so generated models round-trip |
| `linear(center=c)` | `1 + θ*(COV − c)` | θ has covariate units (PsN state 2) |
| `linear_relative(center=c)` | `1 + θ*(COV/c − 1)` | dimensionless θ; exact reparameterization of `linear` (θ_rel = c·θ_abs) — same OFV, different θ scale/SE. No PsN state number; ferx extension |
| `exponential(center=c)` | `exp(θ*(COV − c))` | PsN state 4 |
| `power(center=c)` | `(COV/c)^θ` | PsN state 5 |
| `hockey(breakpoint=b)` | `COV <= b ? 1 + θ_lo*(COV − b) : 1 + θ_hi*(COV − b)` | PsN state 3; **two** thetas — the relation→theta map is 1-to-many from the start, not retrofitted |
| `categorical(ref=r)` | `1 + θ_k` per non-reference level, `1` at the reference | PsN state 2 categorical. `θ = 0` is the null, so bounds are symmetric around zero |
| `expr("…")` | verbatim | escape hatch, equivalent to PsN's `[code]` |

`fix = v` on any form fixes the theta at `v`.

### Data-derived statistics

PsN computes `median` / `minimum` / `maximum` (and the most-common category) from the dataset.
Requiring a literal `center = 70` would make a generated covariate model dataset-specific, which
defeats the automation case. So `center` / `breakpoint` accept `median`, `mean`, `min`, `max`, or
a literal number; `ref` accepts a level literal or `mode`.

Recommendation for users: **literals for final models** (fully reproducible from the file alone),
**symbolic for automated search**. Resolved values are written into the fit YAML
(`center: 70.2  # resolved from median`) so a symbolic run is still reproducible.

### Default thetas — adopt PsN's, verbatim

PsN's linear bounds are not cosmetic: `θ ∈ [1/(med−max), 1/(med−min)]` is exactly what stops
`1 + θ*(COV − med)` going negative over the observed covariate range. No reason to invent
different defaults for a parameterization the field has run for two decades.

| form | init | lower | upper |
|---|---|---|---|
| `linear` (continuous) | `0.001/(med−min)` | `1/(med−max)` | `1/(med−min)` |
| `categorical` | `−0.001` | `−1` | `5` |
| `hockey` (θ_lo, θ_hi) | `0.001/(med−min)`, `0.001/(med−max)` | `−1e6`, `1/(med−max)` | `1/(med−min)`, `1e6` |
| `exponential` | `0.001` | `−100` | `1e6` |
| `power` | `0.001` | `−100` | `1e6` |

## Desugaring — the insertion point

Multiplicative factors commute with `exp(ETA)` numerically, but **not** for the mu-reference
detector used by SAEM — see #619, where a typical value that is not log-linear in one theta
silently drops the covariate effect. So appending the factor at the end of the RHS is wrong.

Rule: split the parameter's RHS into top-level product factors, partition into
eta/kappa-bearing vs. not, and multiply the covariate factor into the **non-eta group**:

```
CL = TVCL * exp(ETA_CL)
  →  CL = TVCL * (WT/70)^THETA_CL_WT * (SEX==1 ? 1 + THETA_CL_SEX_1 : 1) * exp(ETA_CL)
```

When the RHS is not a top-level product (`CL = TVCL + …`, a logit transform, mixture
branching), this is a **hard error**, not a guess. The message points at the explicit handle:

```
[individual_parameters]
  CL = TVCL * COV_CL * exp(ETA_CL)   # COV_CL = the auto-built factor, wired by hand
```

i.e. auto-injection where the shape is unambiguous, a named `COV_<PARAM>` handle where it is not.

## Missing covariate values

PsN's example guards the missing-data token explicitly (`IF(COV.EQ.-99) PARCOV=1`). ferx already
carries missing covariates as NaN in the covariate table, and division here **underflows to 0.0
rather than inf**, so an unguarded `(WT/median)^θ` on a missing row yields a silent `0` instead of
a loud NaN. Every generated factor therefore emits a neutral guard: missing → factor `1.0`. This
needs its own test.

## Structured output

Put the relations on `CompiledModel` and echo them on `FitResult` as a table:
`(parameter, covariate, form, center/breakpoint/ref — as written *and* as resolved, theta name(s),
estimate, SE)`. This is the machine-readable surface an SCM harness or agent consumes; it must
never have to parse the model file.

ferx-r surface (separate PR in `../ferx-r`, building on `ferx_model_set_section.R`):
`ferx_add_covariate(model, "CL", "WT", "power", center = "median")` /
`ferx_drop_covariate(...)` — line-level edits on this one block.

## Validation — must fail loudly

- covariate not declared in `[covariates]` → error (the block is authoritative here, unlike the
  lenient classical path which only warns)
- `PARAM` is not a top-level `[individual_parameters]` name → error
- duplicate `(PARAM, COV)` pair → error
- generated theta name collides with a `[parameters]` declaration → error
- `categorical` on a covariate declared `continuous` (or vice versa) → error
- `categorical` without `levels` in `[covariates]` and no `levels = auto` → error (no
  data-dependent guessing at parse time)
- a relation on a parameter the structural model ignores → warning (the existing dead-parameter
  check already covers this shape)
- symbolic statistic (`median` etc.) still unresolved at evaluation time → hard error

## Transparency

`ferx check` and `ferx_model_show()` must print the **desugared** `[individual_parameters]`,
including resolved centers. Users have to be able to see the exact expression that was built —
otherwise this is unauditable against NONMEM.

## Implementation slices

1. **Parse** `[covariate_model]` → `Vec<CovariateRelation>`, plus `levels` on `[covariates]`.
   Pure data, no codegen. Tier-1 tests.
2. **Late-bound statistics.** Closures capture an `Arc<OnceLock<CovariateStats>>` filled by
   `fit` / `predict` / `simulate` / `check_model_data` from the `Population`; hard error if a
   symbolic stat is unresolved at first evaluation. (Alternative considered: a two-stage
   `bind_data(model, pop) -> CompiledModel`. Cleaner, but `fit(CompiledModel, Population)` is
   public and callers build models from parse alone — much bigger blast radius. Going with the
   late-bound slot.)
3. **Desugar** → expression + auto-theta declarations, before `[individual_parameters]` parses.
   Tier-1 tests on the generated text, plus an equivalence test: a hand-written classical model
   and the block-declared model must give a bit-identical OFV.
4. **Structured echo** on `FitResult` + covtab, FFI, ferx-r accessor.
5. **ferx-r** `ferx_add_covariate` / `ferx_drop_covariate` + docs.

## Testing

- Tier 1: one test per form asserting the generated expression text; missing-value guard;
  insertion-point partition (eta vs non-eta factors); every validation error above.
- Tier 1: default init/bounds match the PsN table for a known median/min/max.
- Tier 2: block-declared vs. hand-written classical model → identical OFV at a fixed parameter
  vector.
- Tier 3 / NONMEM anchor: one model each way (classical and block-declared) against the
  equivalent NONMEM run — the block form must land on the same OFV and estimates as the
  classical form it desugars to. A PsN scm run on the same dataset is the natural second
  reference for the default bounds/inits.

## Docs

New `docs/model-file/covariate-model.qmd` (added to `docs/_quarto.yml`), cross-linked from
`docs/model-file/covariates.qmd` and `docs/model-file/individual-parameters.qmd`. Include the
PsN state ↔ ferx form mapping table above — users coming from scm will look for it.

roninsightrx:
## Why

Closes #1111.

A covariate effect has always been written by hand into `[individual_parameters]`:

```
CL = TVCL * (WT/70)^THETA_CL_WT * exp(ETA_CL)
```

along with the theta declaration, the centering constant, sensible bounds and a missing-data
guard. That is fine for a human writing one model. It is the wrong shape for a machine writing a
hundred: a covariate search (PsN `scm`, `ferx_cov_screen()`, an agent) has to do regex surgery on
an arbitrary expression to add or drop a relation, and has no machine-readable statement of what
covariate model was actually run.

This closes the loop on the screening layer that already exists in ferx-r (`ferx_cov_screen()`,
`ferx_eta_cov()`), which today dead-ends at "here is what is worth testing" with no path to
actually test it.

## What changed

An optional `[covariate_model]` block that declares relations as **structured data**:

```
[covariates]
  WT   continuous
  SEX  categorical(levels = [0, 1])

[covariate_model]
  CL ~ WT   power(center = median)
  CL ~ SEX  categorical(ref = mode)
```

The block is **sugar**. Each relation is rewritten into exactly the classical expression, in
`apply_covariate_model_block` at the top of `parse_full_model_with` — alongside the existing
`apply_ode_template` / `apply_algebraic_structural` source rewrites, and **before**
`absorption_ode_equivalent_source`, which re-emits these blocks to build the ODE twin (a twin
built from un-desugared text would silently carry no covariate effect). Everything downstream —
expression compilation, `Dual2` sensitivities, the SAEM mu-reference detector, `covtab`, SE
reporting — sees an ordinary classical model, and the generated thetas are ordinary thetas. There
is no second evaluation path.

All five PsN `scm` states are expressible (`none`, `linear`, `hockey`, `exponential`, `power`),
plus `categorical`, a dimensionless `linear_relative`, and an `expr("…")` escape hatch for scm's
`[code]`. PsN's default inits and bounds are adopted verbatim — `θ ∈ [1/(c−max), 1/(c−min)]` is
exactly what keeps `1 + θ·(COV−c)` positive over the observed range, so a covariate that does not
vary around its centre is rejected rather than handed infinite bounds.

Three decisions worth a reviewer's attention:

**Insertion point.** The factor goes into the non-η half of the product, immediately before the
first η/κ-bearing factor — not appended to the end of the RHS. Appending is numerically identical
but takes the typical value out of the shape the SAEM mu-reference detector reads, which silently
drops the covariate effect (#619). A right-hand side that is not a top-level product (a sum, a
logit transform, a mixture branch) is a hard error naming the explicit `COV_<PARAM>` handle,
not a guess.

**Missing values.** Every generated factor is wrapped in `if (present(COV)) <factor> else 1.0`.
This is not belt-and-braces: division by a missing covariate **underflows to `0.0`** in this
engine rather than producing an infinity, so an unguarded `(WT/70)^θ` on a missing row would
silently zero the parameter instead of failing loudly. Its own test.

`present(x)` is **new DSL surface in this PR** — a condition, true unless the value is `NaN`,
usable anywhere a condition is (`!present(WT)`, `present(WT) && WT > 100`). It replaces the
`COV == COV` spelling the guard first used: correct, but it reads like a typo, and this is text
a user diffs against a NONMEM control stream. Deliberately a condition rather than a function
returning 0.0/1.0 — as arithmetic, `TVCL * present(WT)` would produce exactly the silently zeroed
parameter the guard prevents. It is `!v.is_nan()`, not `is_finite()`, so it means what the idiom
it replaces meant (an infinity is a value, not a gap in the data); a test runs both spellings
over {value, 0, −1, NaN, ±inf} and asserts they agree. Carried through both bytecode evaluators
as `Op::IsPresent`; in the `PkNum` one it pushes a constant, since an indicator's derivative is
identically zero — and the Tier-3 SE comparison is the oracle for that arm.

**One θ per relation.** A relation owns its θ and two relations cannot share one, so a single
allometric exponent estimated jointly on `CL` and `V` — what `examples/two_cpt_oral_cov.ferx`
does — has no spelling here. This follows PsN (`scm` gives every (parameter, covariate) relation
its own θ) and is what keeps a relation a self-contained line: sharing would couple two lines,
and a search could no longer drop one without reasoning about the other. The error message names
the workaround (write that factor classically — the two styles mix), and both the module doc and
the docs page carry it with an example.

**Data-derived statistics.** `center`/`breakpoint`/`ref` accept `median`/`mean`/`min`/`max`/`mode`
and `[covariates]` accepts `levels = auto`, resolved by `bind_covariate_stats()`. A relation that
reaches a fit still unresolved is `E_COVSTAT_UNRESOLVED` — never a quiet fit with the covariate
effect absent. Without `--data` it is `W_COVSTAT_UNBOUND`, and the desugared echo in `ferx check`
is *suppressed* rather than printed without the pending effects (printing it would state the
opposite of the truth).

Structured echo: `FitResult.covariate_relations`, a `covariate_model:` section in the fit YAML
(carrying both the source form and the resolved value — `center: 70.85  # resolved from median`),
and `ferx check` printing the desugared `[individual_parameters]`.

## Alternatives considered

**Symbolic statistics: bound re-parse vs. a late-bound slot.** The issue proposed an
`Arc<OnceLock<CovariateStats>>` on `CompiledModel`, captured by the compiled closures and filled
at each entry point. This PR instead mirrors θ-level binding (#1064): `bind_covariate_stats()`
computes the statistics and re-parses, so the desugared expression carries a literal. Reasons:
no new AST node in the tree-walking *and* bytecode evaluators plus the sens path; and the
`OnceLock` is sticky — a model reused against a second dataset (a bootstrap resample, a VPC on
new data) would silently keep the first dataset's centres. Re-parsing costs milliseconds. The
cost of this choice is that a caller driving `fit(&model, &pop)` directly must call the binder;
that is a loud `E_COVSTAT_UNRESOLVED`, not a silent wrong answer, and literal-centred models
(the recommendation for final models) need no binding at all.

Both binders now travel through one `ParseBindings` struct, because each re-parses and a model
may declare both — binding one must not drop the other's. Pinned by a test.

**Syntax: `PARAM ~ COV form(...)` vs `PARAM ~ form(COV, ...)`.** The R-formula spelling
(`CL ~ power(WT, center = median)`, echoing `y ~ s(WT, k = 5)`) reads more naturally to R users,
who are most of this DSL's audience. Rejected, and the rationale is recorded in the module doc:
once the RHS is a term expression, `CL ~ power(WT) + linear(CRCL)` is what an R-literate reader
writes next — which breaks one-relation-per-line (the property the block exists for), and `+`
reads as additive when these effects are multiplicative. Also, `(parameter, covariate)` is the
primary key of a covariate search; this spelling puts both at fixed positions at line start with
the form — the thing a search varies — as the one token that moves. PsN's own config keys the
same way (`[test_relations] CL=WT,SEX`, state in a separate table).

**Search configuration.** PsN's `[test_relations]`, `valid_states`, `included_relations`,
`p_forward` are deliberately **out of scope**: they are search configuration, not model. They
belong in an SCM harness that reads the relation table and rewrites these lines.

## Cross-repo dependency

The ferx-r side (`ferx_add_covariate()` / `ferx_drop_covariate()`, issue #1111 slice 5) is a
follow-up PR and cannot land first: it needs a `src/rust/Cargo.lock` bump to a ferx-core commit
that has these `pub` items, via `ferx-r/tools/update-ferx-core-lock.sh`. Nothing in ferx-r breaks
without it — this PR only adds API surface and one optional block.

## Breaking changes
- [ ] None

Additive only — no existing signature or field changed, so nothing downstream breaks. Baseline
regenerated (+166 lines). What was added and who needs it:

| Item | Needed by |
|---|---|
| `types::{CovariateForm, CovariateStat, CovariateLevels, CovariateTheta, CovariateRelation, CovariateModelSpec, CovariateSummary}` | ferx-r's `ferx_add_covariate()` / `ferx_drop_covariate()`, which construct and inspect relations; an SCM harness in ferx-tools reading the relation table |
| `types::{CovariateRelationEstimate, CovariateThetaEstimate}` + `FitResult::covariate_relations` | the post-fit read-back — the whole point of the block; ferx-r surfaces it as a data frame |
| `CompiledModel::covariate_model` | `ferx_model_show()` printing the desugared block without re-parsing |
| `CovariateDecl::levels` + `CovariateDecl::new` | `levels = [...]`; the constructor keeps the pre-existing two-field construction working for external callers |
| `api::{bind_covariate_stats, assert_covariate_model_bound}` | ferx-r calls `parse_full_model_file` + `read_population_for` + `fit` directly, so it must be able to bind statistics itself and to check boundness |
| `parser::model_parser::{ParseBindings, parse_full_model_with}` | the binder's own re-parse; `parse_full_model_bound` is kept as a wrapper so existing callers are untouched |
| `parser::covariate_model::CovariateStatBindings` | type of a `ParseBindings` field. The desugar itself (`apply_covariate_model`, `CovariateModelContext`) stays `pub(crate)` |
| `CheckReport::desugared_individual_parameters` | `ferx check --json`; ferx-r's `ferx_model_validate()` |

## Numerical validation
- [x] Compared against: prior ferx behaviour — the classical covariate model this block desugars to
- [x] Reference values:

`tests/covariate_model_equivalence.rs` runs the same covariate model both ways on
`data/two_cpt_oral_cov.csv` and asserts **bit-identical** results — `assert_eq!` on f64, not a
tolerance, because the desugar emits the same expression tree in the same multiplication order,
so any difference at all would mean the generated text is not the model the author wrote:

- `predict()` PRED, pointwise, at a fixed parameter vector
- OFV after a short fit from the same start
- (slow-tests) converged θ, Ω, σ **and `se_theta`** — the SE comparison is the level at which a
  difference in the *gradient* of the generated expression, rather than in its value, would show

The convergence test carries non-degeneracy guards — both arms must report `converged`, all three
covariate θ must have moved from their inits, and the covariance step must have run — so a green
result cannot mean both sides stalled identically at the start.

**No new NONMEM run, deliberately.** The block emits nothing the classical path could not already
emit, and the classical covariate models are already NONMEM-anchored, so the anchor for this
feature is that the two forms are the same model. Discussed with @roninsightrx before
implementation.

`examples/two_cpt_oral_covmodel.ferx` on `data/two_cpt_oral_cov.csv`, release build:

```
# classical sibling, examples/two_cpt_oral_cov.ferx (WT exponent shared across CL and V1)
OFV: -1168.4846   converged: true
TVCL 4.948619  TVV1 38.217475  TVQ 10.096271  TVV2 98.422611  TVKA 0.983651
THETA_WT 0.636325  THETA_CRCL 0.562252

# block-declared, examples/two_cpt_oral_covmodel.ferx (a separate exponent per relation)
OFV: -1199.0246   converged: true
TVCL 4.904208  TVV1 48.541162  TVQ 9.620947  TVV2 94.364402  TVKA 1.257439
THETA_CL_WT 0.893249  THETA_CL_CRCL 0.508259  THETA_V1_WT 0.473657
```

Not the same model — the classical example shares one `THETA_WT` between CL and V1, the block
gives each relation its own theta — so these are not expected to match. The
model-for-model comparison is the equivalence test above.

**One thing a reviewer should know.** While running the example for this PR, the block-declared
model came back `converged: false` with every parameter still at its initial estimate. It is
**not** the block: the hand-written model it desugars to, with no `[covariate_model]` anywhere,
stalls identically at the same OFV (-1026.3504) with the same all-at-init vector. Switching the
outer optimizer to BOBYQA converges both to -1199.02. So this is a pre-existing init stall of the
#751 family, exposed by this parameterization and start (three separate covariate exponents
rather than the classical example's two shared ones), not introduced here. The example therefore
carries `optimizer = bobyqa` with a comment saying so. Worth its own issue; happy to open one.

## Performance

The desugar runs once per parse, only when the block is present. A model that declares it parses
`[parameters]` twice (once to learn the η/κ names the insertion-point partition needs, once for
real after the generated `theta` lines are appended) — microseconds, and zero for every model
that does not use the block. Symbolic statistics add one re-parse per fit, the same cost #1064
already pays.

## Tests

`cargo test --lib`: 3756 passed, 0 failed.

- **Tier 1** `src/parser/model_parser_tests.rs` (6, new): `present(...)` on its own — true for a
  value and false for a missing one through the compiled bytecode path; agreement with the
  `COV == COV` idiom across `NaN` and both infinities; negation and `&&` composition; an
  expression argument; a covariate column actually *named* `present` still parsing.
- **Tier 1** `src/parser/covariate_model_tests.rs` (37): one test per form asserting the
  *generated text*; the missing-value guard (a NaN covariate must give the same parameter as the
  centring value, and must not be 0); the insertion-point partition, including a κ-bearing factor
  and the hard errors on a sum and on a log-transformed RHS; every validation error; `none`
  round-tripping unchanged.
- **Tier 1** `src/api/tests/covariate_stats_tests.rs` (13): each statistic against a synthetic
  population; per-subject weighting (a subject with 40 records must not drag the median);
  time-varying covariates contributing each distinct value; PsN's `linear` and `hockey` bound
  tables for a known median/min/max; the constant-covariate rejection; the unresolved hard error
  through both `assert_covariate_model_bound` and `check_model_data`; and that binding statistics
  after binding θ-levels preserves the level binding.
- **Tier 2** `tests/covariate_model_equivalence.rs`: the equivalence above, `outer_maxiter = 2`.
- **Tier 3** same file, `slow-tests`-gated: to convergence with a covariance step.

`examples/two_cpt_oral_covmodel.ferx` — `examples/two_cpt_oral_cov.ferx` restated with the block,
and the fixture behind the equivalence test.

`ferx check examples/two_cpt_oral_covmodel.ferx` prints what it built:

```
[individual_parameters] as built from [covariate_model]:
  CL = TVCL * (if (WT == WT) (WT / 70)^THETA_CL_WT else 1.0) * (if (CRCL == CRCL) (CRCL / 100)^THETA_CL_CRCL else 1.0) * exp(ETA_CL)
  V1 = TVV1 * (if (WT == WT) (WT / 70)^THETA_V1_WT else 1.0) * exp(ETA_V1)
  Q  = TVQ  * exp(ETA_Q)
  V2 = TVV2 * exp(ETA_V2)
  KA = TVKA * exp(ETA_KA)
```

and a symbolic-centred run writes the resolved constants back out:

```yaml
covariate_model:
  - parameter: "CL"
    covariate: "WT"
    form: "power"
    center: 70.85  # resolved from median
    thetas:
      - name: "THETA_CL_WT"
        estimate: 0.783371
        se: null
```

## Docs

New `docs/model-file/covariate-model.qmd` (grammar, form table, the PsN state ↔ ferx form mapping
users coming from scm will look for first, the PsN defaults table, the literal-vs-symbolic
recommendation, the desugaring rule and the `COV_<PARAM>` handle, the SAEM caveat for
`categorical`, and the validation list), cross-linked from `covariates.qmd` and
`individual-parameters.qmd`.

## Checklist
- [x] `cargo clippy` clean (no new warnings from this change)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched; the desugar
      produces ordinary DSL text, and the `Dual2` path compiles it exactly as it compiles a
      hand-written covariate factor. The SE equivalence in the slow test is what pins this.

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → `ferx-site/model-dsl/` page updated or PR opened — **not yet; needs a follow-up PR in ferx-site**
- [x] New example `.ferx` file added to `examples/` → matching entry in ferx-site examples and ferx-book — **not yet, same follow-up**
- [x] New data file added to `data/` → none added (reuses `data/two_cpt_oral_cov.csv`)

### ferx-book
- [ ] New estimator, DSL feature, or fit option → relevant book chapter updated or PR opened — **not yet, same follow-up**

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core: `cd ../ferx-r && R CMD INSTALL .` — **not run.** The
      local `../ferx-r` checkout does not currently build against `main` either, so a failure there
      would say nothing about this branch; and nothing in ferx-r reads the new surface until its
      own PR.
- [x] All affected `examples/*.ferx` run cleanly via CLI
- [ ] All affected ferx-site example `.qmd` pages render cleanly — no site page exists yet
- [ ] All affected ferx-book chapters render cleanly — no chapter exists yet

## Reviewer hints

Start with the module doc in `src/parser/covariate_model.rs` — it states the three decisions
above (insertion point, missing guard, syntax shape) and why each alternative was rejected.

Worth real scrutiny:

- `insert_factors` / `has_top_level_additive` — the text-level product split. It decides where the
  factor lands, and a wrong classification is a silently different model under SAEM. Look at what
  it accepts as a "simple factor" and whether any real RHS shape is mis-classified rather than
  rejected.
- `build_thetas` / `linear_family_thetas` — the PsN default tables and the constant-covariate
  guard.
- `apply_covariate_model_block` in `model_parser.rs` — the double `parse_parameters`, and its
  placement relative to `absorption_ode_equivalent_source`.

Skimmable: the `types.rs` additions are plain data; the 36 one-line `covariate_model: None` /
`levels: None` / `covariate_relations: Vec::new()` additions across existing test fixtures are
mechanical (new struct fields).

## Open questions

None outstanding. The two raised during review are resolved in the branch: the guard now reads
`if (present(WT)) … else 1.0`, and the one-θ-per-relation restriction is documented rather than
lifted.

https://claude.ai/code/session_01Ccp783ygYzXWTrXjTLDDgq

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,15 @@ section of the SDLC for the versioning policy).
   `center`/`breakpoint`/`ref` accept a literal or a data-derived statistic (`median`, `mean`,
   `min`, `max`, `mode`), and `[covariates]` gains `categorical(levels = [0, 1])` /
   `levels = auto`. The relations are echoed on `FitResult.covariate_relations` and in the fit
-  YAML, and `ferx check` prints the desugared block. Each relation owns its own θ (as in `scm`);
+  YAML — each θ naming the categorical `level` it contrasts, each `expr(…)` relation carrying
+  the expression that ran, and a relation with no θ emitting `thetas: []` rather than a bare key.
+  Statistics summarise every event record (observations, doses, `EVID=2` markers, `EVID=3/4`
+  resets), so the default bounds cover the range the fit evaluates over. Centres that would break
+  their own form are refused up front: outside the observed range for the linear family (whose
+  default bounds would come back reversed), non-positive for `power`, zero for `linear_relative`;
+  so is a categorical value the block never declared, which would otherwise be modelled silently
+  as the reference level (`E_COV_LEVEL_UNKNOWN`). `ferx check` prints the desugared block. Each
+  relation owns its own θ (as in `scm`);
   a θ shared across relations is still written classically. Relations are line-oriented and
   independent,
   so a covariate search rewrites this one block instead of doing surgery on an expression. See
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@ section of the SDLC for the versioning policy).
   still `ferx-core`.
 
 ### Added
+- **A `present(COV)` condition for covariates that may be missing (#1111).** A missing covariate
+  value is `NaN`, and division by it underflows to `0.0` here rather than erroring, so an
+  unguarded `(CRCL/100)^THETA` silently zeroes the parameter on a row with no `CRCL`.
+  `if (present(CRCL)) ... else 1.0` says the guard plainly; the equivalent `CRCL == CRCL` idiom
+  still works. Usable anywhere a condition is (`!present(X)`, `present(X) && X > 100`), and it is
+  what `[covariate_model]` wraps every generated factor in.
 - **A `[covariate_model]` block: covariate–parameter relationships declared as data (#1111).**
   `CL ~ WT power(center = median)` desugars into exactly the classical
   `[individual_parameters]` expression — with the theta declaration, the centering constant and a
@@ -40,7 +46,9 @@ section of the SDLC for the versioning policy).
   `center`/`breakpoint`/`ref` accept a literal or a data-derived statistic (`median`, `mean`,
   `min`, `max`, `mode`), and `[covariates]` gains `categorical(levels = [0, 1])` /
   `levels = auto`. The relations are echoed on `FitResult.covariate_relations` and in the fit
-  YAML, and `ferx check` prints the desugared block. Relations are line-oriented and independent,
+  YAML, and `ferx check` prints the desugared block. Each relation owns its own θ (as in `scm`);
+  a θ shared across relations is still written classically. Relations are line-oriented and
+  independent,
   so a covariate search rewrites this one block instead of doing surgery on an expression. See
   `docs/model-file/covariate-model.qmd`.
 - **A CI-enforced public-API baseline for `ferx-core` (#1114).** `api/ferx-core-public-api.txt` is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,18 @@ section of the SDLC for the versioning policy).
   still `ferx-core`.
 
 ### Added
+- **A `[covariate_model]` block: covariate–parameter relationships declared as data (#1111).**
+  `CL ~ WT power(center = median)` desugars into exactly the classical
+  `[individual_parameters]` expression — with the theta declaration, the centering constant and a
+  missing-value guard filled in — so the OFV and estimates are identical to writing it by hand.
+  All five PsN `scm` states are expressible (`none`, `linear`, `hockey`, `exponential`, `power`,
+  plus `categorical` and an `expr("…")` escape hatch) with PsN's default inits and bounds;
+  `center`/`breakpoint`/`ref` accept a literal or a data-derived statistic (`median`, `mean`,
+  `min`, `max`, `mode`), and `[covariates]` gains `categorical(levels = [0, 1])` /
+  `levels = auto`. The relations are echoed on `FitResult.covariate_relations` and in the fit
+  YAML, and `ferx check` prints the desugared block. Relations are line-oriented and independent,
+  so a covariate search rewrites this one block instead of doing surgery on an expression. See
+  `docs/model-file/covariate-model.qmd`.
 - **A CI-enforced public-API baseline for `ferx-core` (#1114).** `api/ferx-core-public-api.txt` is
   a committed snapshot of the crate's public surface; a CI job regenerates and diffs it, so any
   widening of the API fails until the baseline is updated in the same PR. Regenerate with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@ section of the SDLC for the versioning policy).
   write them package-qualified (`--features ferx-core/ci`). Consumers of the `ferx-core` crate —
   including the ferx-r wrapper, which patches the repo root — are unaffected: the root package is
   still `ferx-core`.
+- **A covariate column with no value for a subject now reads as `NaN`, not as a dropped key
+  (#1111).** Previously the key was simply absent from that subject's covariate map, and every
+  evaluation site resolves an absent covariate to `0.0` — so a subject whose `WT` column was `.`
+  on every row was fitted at `WT = 0`, indistinguishable from a genuine zero, and `present(WT)`
+  reported it as present. The reader now stores `NaN` and warns, naming the covariate and the
+  affected subjects, so the gap is either guarded with `present(...)`, imputed, or fails loudly.
 
 ### Added
 - **A `present(COV)` condition for covariates that may be missing (#1111).** A missing covariate
--- a/api/ferx-core-public-api.txt
+++ b/api/ferx-core-public-api.txt
@@ -57,6 +57,8 @@ pub ferx_core::api::SurvivalPredictionResult::median_survival: f64
 pub ferx_core::api::SurvivalPredictionResult::survival: f64
 pub ferx_core::api::SurvivalPredictionResult::survival_all: f64
 pub ferx_core::api::SurvivalPredictionResult::time: f64
+pub fn ferx_core::api::assert_covariate_model_bound(model: &ferx_core::CompiledModel) -> core::result::Result<(), alloc::string::String>
+pub fn ferx_core::api::bind_covariate_stats(parsed: &mut ferx_core::ParsedModel, model_text: &str, population: &ferx_core::Population) -> core::result::Result<(), alloc::string::String>
 pub fn ferx_core::api::bind_theta_levels(parsed: &mut ferx_core::ParsedModel, model_text: &str, population: &mut ferx_core::Population) -> core::result::Result<(), alloc::string::String>
 pub fn ferx_core::api::check_experimental_features(model: &ferx_core::CompiledModel) -> alloc::vec::Vec<ferx_core::diagnostics::Diagnostic>
 pub fn ferx_core::api::check_model_data(model: &ferx_core::CompiledModel, population: &ferx_core::Population) -> alloc::vec::Vec<ferx_core::diagnostics::Diagnostic>
@@ -123,6 +125,7 @@ pub ferx_core::diagnostics::Severity::Error
 pub ferx_core::diagnostics::Severity::Warning
 pub struct ferx_core::diagnostics::CheckReport
 pub ferx_core::diagnostics::CheckReport::data: core::option::Option<alloc::string::String>
+pub ferx_core::diagnostics::CheckReport::desugared_individual_parameters: alloc::vec::Vec<alloc::string::String>
 pub ferx_core::diagnostics::CheckReport::diagnostics: alloc::vec::Vec<ferx_core::diagnostics::Diagnostic>
 pub ferx_core::diagnostics::CheckReport::model: alloc::string::String
 pub ferx_core::diagnostics::CheckReport::valid: bool
@@ -1028,6 +1031,8 @@ pub fn ferx_core::ode::solve_ode_with_stats(rhs: &dyn core::ops::function::Fn(&[
 pub type ferx_core::ode::OdeOutputFn = alloc::boxed::Box<(dyn core::ops::function::Fn(&[f64], &[f64], &[f64], &[f64], &std::collections::hash::map::HashMap<alloc::string::String, f64>) -> f64 + core::marker::Send + core::marker::Sync)>
 pub type ferx_core::ode::OdeRhsFn = alloc::boxed::Box<(dyn core::ops::function::Fn(&[f64], &[f64], f64, &mut [f64]) + core::marker::Send + core::marker::Sync)>
 pub mod ferx_core::parser
+pub mod ferx_core::parser::covariate_model
+pub type ferx_core::parser::covariate_model::CovariateStatBindings = std::collections::hash::map::HashMap<alloc::string::String, ferx_core::CovariateSummary>
 pub mod ferx_core::parser::model_parser
 pub enum ferx_core::parser::model_parser::LevelContrast
 pub ferx_core::parser::model_parser::LevelContrast::Auto
@@ -1054,6 +1059,9 @@ pub fn ferx_core::parser::model_parser::LevelBlockDecl::name(&self) -> &str
 pub fn ferx_core::parser::model_parser::LevelBlockDecl::shares_scale_with_eta(&self) -> bool
 pub struct ferx_core::parser::model_parser::OdeOutputProgram
 pub struct ferx_core::parser::model_parser::OdeRhsProgram
+pub struct ferx_core::parser::model_parser::ParseBindings
+pub ferx_core::parser::model_parser::ParseBindings::covariate_stats: ferx_core::parser::covariate_model::CovariateStatBindings
+pub ferx_core::parser::model_parser::ParseBindings::levels: ferx_core::parser::model_parser::LevelBindings
 pub struct ferx_core::parser::model_parser::RuvMagDerivProgram
 pub struct ferx_core::parser::model_parser::ScaleDerivProgram
 pub struct ferx_core::parser::model_parser::ThetaBlocks
@@ -1072,6 +1080,7 @@ pub fn ferx_core::parser::model_parser::known_block_names() -> alloc::vec::Vec<&
 pub fn ferx_core::parser::model_parser::parse_full_model(content: &str) -> core::result::Result<ferx_core::ParsedModel, alloc::string::String>
 pub fn ferx_core::parser::model_parser::parse_full_model_bound(content: &str, level_bindings: &ferx_core::parser::model_parser::LevelBindings) -> core::result::Result<ferx_core::ParsedModel, alloc::string::String>
 pub fn ferx_core::parser::model_parser::parse_full_model_file(path: &std::path::Path) -> core::result::Result<ferx_core::ParsedModel, alloc::string::String>
+pub fn ferx_core::parser::model_parser::parse_full_model_with(content: &str, bindings: &ferx_core::parser::model_parser::ParseBindings) -> core::result::Result<ferx_core::ParsedModel, alloc::string::String>
 pub fn ferx_core::parser::model_parser::parse_model_file(path: &std::path::Path) -> core::result::Result<ferx_core::CompiledModel, alloc::string::String>
 pub fn ferx_core::parser::model_parser::parse_model_string(content: &str) -> core::result::Result<ferx_core::CompiledModel, alloc::string::String>
 pub type ferx_core::parser::model_parser::LevelBindings = std::collections::hash::map::HashMap<alloc::string::String, ferx_core::parser::model_parser::LevelBinding>
@@ -1876,11 +1885,38 @@ pub ferx_core::types::CovarianceStatus::Computed
 pub ferx_core::types::CovarianceStatus::Failed
 pub ferx_core::types::CovarianceStatus::NotRequested
 pub ferx_core::types::CovarianceStatus::SirFallback
+pub enum ferx_core::types::CovariateForm
+pub ferx_core::types::CovariateForm::Categorical
+pub ferx_core::types::CovariateForm::Exponential
+pub ferx_core::types::CovariateForm::Expr(alloc::string::String)
+pub ferx_core::types::CovariateForm::Hockey
+pub ferx_core::types::CovariateForm::Linear
+pub ferx_core::types::CovariateForm::LinearRelative
+pub ferx_core::types::CovariateForm::None
+pub ferx_core::types::CovariateForm::Power
+impl ferx_core::CovariateForm
+pub fn ferx_core::CovariateForm::is_categorical(&self) -> bool
+pub fn ferx_core::CovariateForm::label(&self) -> &'static str
 pub enum ferx_core::types::CovariateKind
 pub ferx_core::types::CovariateKind::Categorical
 pub ferx_core::types::CovariateKind::Continuous
 impl ferx_core::CovariateKind
 pub fn ferx_core::CovariateKind::label(&self) -> &'static str
+pub enum ferx_core::types::CovariateLevels
+pub ferx_core::types::CovariateLevels::Auto
+pub ferx_core::types::CovariateLevels::Declared(alloc::vec::Vec<f64>)
+impl ferx_core::CovariateLevels
+pub fn ferx_core::CovariateLevels::declared(&self) -> core::option::Option<&[f64]>
+pub enum ferx_core::types::CovariateStat
+pub ferx_core::types::CovariateStat::Literal(f64)
+pub ferx_core::types::CovariateStat::Max
+pub ferx_core::types::CovariateStat::Mean
+pub ferx_core::types::CovariateStat::Median
+pub ferx_core::types::CovariateStat::Min
+pub ferx_core::types::CovariateStat::Mode
+impl ferx_core::CovariateStat
+pub fn ferx_core::CovariateStat::label(&self) -> alloc::string::String
+pub fn ferx_core::CovariateStat::literal(&self) -> core::option::Option<f64>
 pub enum ferx_core::types::DerivedKind
 pub ferx_core::types::DerivedKind::Aggregate
 pub ferx_core::types::DerivedKind::Aggregate::filter: core::option::Option<ferx_core::DerivedFilterFn>
@@ -2215,6 +2251,7 @@ pub ferx_core::types::CompiledModel::absorption_ode_equivalent: core::option::Op
 pub ferx_core::types::CompiledModel::analytic_readout: core::option::Option<ferx_core::AnalyticReadout>
 pub ferx_core::types::CompiledModel::analytical_init: alloc::vec::Vec<ferx_core::AnalyticalInit>
 pub ferx_core::types::CompiledModel::bloq_method: ferx_core::BloqMethod
+pub ferx_core::types::CompiledModel::covariate_model: core::option::Option<ferx_core::CovariateModelSpec>
 pub ferx_core::types::CompiledModel::covariate_nns: alloc::vec::Vec<ferx_core::nn::CovariateNn>
 pub ferx_core::types::CompiledModel::default_params: ferx_core::ModelParameters
 pub ferx_core::types::CompiledModel::derived_exprs: alloc::vec::Vec<ferx_core::DerivedExprSpec>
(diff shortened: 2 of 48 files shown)


roninsightrx (commented):
Blocking review findings:

1. src/api/fit.rs:97-104 — fit_from_files never calls bind_covariate_stats. It binds theta level blocks and immediately moves parsed.model, so any documented symbolic relation such as center = median remains undesugared and fit() rejects it as E_COVSTAT_UNRESOLVED. The error text explicitly recommends fit_from_files as an entry point that binds these statistics, so this public API currently violates its contract. Please bind covariate stats after theta levels here and add a regression test through fit_from_files.

2. src/api/covariate_stats.rs:117-128 — summarize only considers the subject fallback map and observation snapshots. It omits dose_covariates and pk_only_covariates; after rebasing onto current main it must also account for reset snapshots. A covariate change carried on EVID=1/2/3/4 is used by the event-driven evaluator but is excluded from median/min/max/mode and levels=auto. That can produce bounds that do not cover values the model actually evaluates, or omit a categorical level and silently apply the reference factor. Please summarize every event snapshot, deduplicated per subject, and cover dose-only and EVID=2 changes.

3. src/parser/covariate_model.rs:616-665 — the default linear-family bounds only reject center == min/max. If a literal center lies outside the observed range, for example center=100 with a range 50..90, this emits lower=0.1 and upper=0.02. linear_relative also reverses the bounds for a negative center, and center=0 makes its emitted COV/center expression collapse through the evaluator division guard. Please validate the centering domain and ensure the transformed bounds remain ordered; power and linear_relative also need a nonzero compatible center.

4. src/parser/covariate_model.rs:793-803 — a categorical value not present in declared levels falls through to factor 1, exactly like the reference category. A new or mistyped code in the data therefore changes the fitted model silently. This needs a pre-fit validation that every finite observed value belongs to the declared level set rather than treating unknown values as reference.

5. src/api/postfit.rs:1272-1289 — CovariateThetaEstimate drops CovariateTheta.level. With categorical relations and explicit theta names, a FitResult consumer cannot determine which estimate belongs to which level without reparsing the model, which defeats the structured relation table. The same output also reduces expr("...") to form: "expr" and loses the expression that actually ran. Please retain this machine-readable relation metadata in FitResult and YAML.

6. src/io/output.rs:2041 — forms with no theta, notably none and expr, emit a bare "thetas:" key. YAML parses that as null, not an empty list, so the manual YAML shape disagrees with the Vec field exposed by FitResult. Emit "thetas: []" when the relation has no theta and add a YAML parse/round-trip test.

Separately, the PR merge against current main does not compile: src/api/tests/covariate_stats_tests.rs:48 is missing the newly added Subject.reset_covariates and reset_occasions fields. The current Check, Clippy, and coverage failures all stem from that stale fixture, so the branch needs a rebase/update before it can merge.

Validation performed at PR head dff0acce:
- cargo test --lib covariate_: 133 passed
- cargo test --test covariate_model_equivalence: 2 passed, 1 slow test ignored
- git diff --check: clean

roninsightrx:
## Code review — `[covariate_model]` block

Reviewed the 4 commits since `7d273f1e`. Five findings, two of which are behavioural bugs that silently produce wrong numbers.

---

### 1. HIGH — `src/parser/covariate_model.rs:916`: a top-level sum containing any `*` is accepted, and the covariate factor is multiplied into one addend only

`non_product` is `parts.len() == 1 && !is_simple_factor(rhs)`, so `has_top_level_additive` is consulted *only* when the RHS has no top-level `*`.

For `CL = TVCL * exp(ETA_CL) + BASE_CL`, `split_top_level(rhs, '*')` yields `["TVCL ", " exp(ETA_CL) + BASE_CL"]` — `parts.len() == 2`, the guard is skipped, and the emitted line is:

```
CL = TVCL * <factor> * exp(ETA_CL) + BASE_CL
```

The covariate factor scales only the first addend. Same shape for an inline conditional followed by a product (`CL = if (SEX==1) 1.2 else 1.0 * TVCL`), where the factor attaches to the `else` branch.

The PR description and `non_product_error` both state that a sum is a hard error. The only test covering it (`a_non_product_right_hand_side_is_a_hard_error`) uses `TVCL + exp(ETA_CL)`, which has no top-level `*` and therefore never exercises this path.

**Fix:** run `has_top_level_additive(rhs)` unconditionally (or per part), not just when `parts.len() == 1`.

---

### 2. HIGH — `src/parser/covariate_model.rs:895`: the `present(COV)` missing-value guard never fires for data read from CSV

`Condition::Present` is `!v.is_nan()`, but no covariate value the reader produces is ever `NaN`:

- `io/datareader.rs:1483-1489` inserts only finite parses into `Subject::covariates`.
- The LOCF update at `io/datareader.rs:1571-1575` likewise inserts only finite values.

So a missing cell leaves the key *absent*, and every evaluation path resolves an absent covariate to `0.0` (`parser/model_parser.rs:17708`, `:17738`, and the `pk_cov[i] = covariates.get(name).copied().unwrap_or(0.0)` fill at `:15443`).

Concrete failure: a subject whose `WT` column is `.` on every row. The column exists, so `check_covariates` passes; `present(WT)` evaluates to true; the guard takes the *then* branch; `(0/70)^0.75 = 0`, and that subject's `CL` is silently zero — exactly the failure the guard exists to prevent.

The Tier-1 test `a_missing_covariate_contributes_exactly_one` passes only because it injects `f64::NAN` into the map by hand — a state the reader cannot produce.

**Fix:** cover the absent/zero case too (an explicit sentinel, or have the reader store `NaN` for a wholly-missing covariate), and route the test through `read_nonmem_csv` on a dataset with a `.` column.

---

### 3. LOW/MEDIUM — `src/parser/covariate_model.rs:591` (with `:767`): a categorical relation on a single-level covariate produces a misleading "needs data-derived statistics" error

`levels = [0]` (or `levels = auto` on a constant column) gives `expected_theta_count == 0`, `build_thetas` returns an empty vec, and `CovariateRelation::needs_data()` is then true via `self.thetas.is_empty()`. The user gets `E_COVSTAT_UNRESOLVED`, telling them to supply `--data` or literals — which can never help, since a single-level covariate carries no contrast.

**Fix:** its own message, e.g. "declares only one level, so it has no contrast to estimate".

---

### 4. LOW — `src/parser/covariate_model.rs:922`: an η reached through an intermediate variable defeats the insertion-point partition

`first_random` classifies a factor by `mentions_any(part, eta_kappa)`, i.e. by literal η/κ *names*. For:

```
CLI = exp(ETA_CL)
CL  = TVCL * CLI
```

no part mentions `ETA_CL`, so `first_random == parts.len()` and the covariate factor is appended *after* the η-bearing factor — precisely the placement the module doc says must never happen (#619, mu-reference drop under SAEM). It fails silently rather than raising the `COV_<PARAM>` hard error.

**Fix:** either resolve η through intermediate assignments, or reject a RHS whose factors reference block-local variables.

---

### 5. LOW — `src/parser/covariate_model.rs` (`parse_relation_line` / `parse_kwargs`, ~`:290`): `fix` is parsed and then silently discarded for forms that declare no θ

`none(fix = 0.5)` parses, sets `rel.fix`, and `apply_fix` runs over an empty θ list — so the argument has no effect and produces no diagnostic. This is unlike the sibling `center` / `breakpoint` / `ref` keys, which `center_argument` rejects explicitly for `none` / `expr`.

**Fix:** reject `fix` there too.

---

### Checked and holding up

- `Condition::Present` is threaded through all four `Op` sites (compile, both evaluators, `scan_stack_depth`) and every `Condition` match arm.
- The unresolved-relation gate is wired into `fit` / `predict` / `simulate` / `check`.
- `ParseBindings` correctly preserves level bindings across the covariate re-parse.
- `mode_of` tie-breaking, and the PsN linear / hockey bound tables.
- The generated categorical / hockey text parses at the intended precedence (inline `if` takes `parse_add_sub` on both branches).

PR #1158, opened 2026-08-31
build(ci): one owner for the fast gates — tools/preflight.sh, which CI itself runs [closes #1157]

### LINKED ISSUE #1157: build: a committed tools/preflight.sh that CI itself runs, so "green locally" means the same thing as "green in CI"
`tools/update-public-api.sh` opens with the reason it exists:

> CI runs this exact script (`--check`) so the generating command cannot drift from the verifying one.

That reasoning applies just as well to the rest of the gate, and nothing implements it. A contributor who wants to know whether a change will pass CI has to open `.github/workflows/ci.yml` and reconstruct **nine** commands by hand, spread across four jobs.

## The gap

From `origin/main`'s `ci.yml` today:

```bash
# job: check — five separate feature sets
cargo check --tests --no-default-features --features ci
cargo check --tests --no-default-features --features ci,survival,slow-tests
cargo check --tests --no-default-features --features ci,markov
cargo check --tests --no-default-features --features ci,nn,slow-tests
cargo check -p ferx-tools -p ferx-cli --tests --no-default-features --features ferx-core/ci

# job: clippy — note --all-targets, and the package-qualified features
cargo clippy --no-default-features --features ci,markov,nn --all-targets
cargo clippy -p ferx-tools -p ferx-cli --tests --no-default-features --features ferx-core/ci,ferx-core/markov,ferx-core/nn

# job: fmt
cargo fmt --all -- --check

# job: public-api
tools/update-public-api.sh --check
```

Every one of those is a trap for a different reason, and the file's own comments say so:

* **The feature sets are not nested.** `ci,survival` compiles neither the `nn` surface nor the `slow-tests`-gated files. A *fully green* `cargo test --features ci,survival` proves nothing about roughly a third of the repo.
* **`--all-targets` is load-bearing.** Without it clippy covers only lib and bins — every `#[cfg(test)]` module and every sibling `*_tests.rs` goes unlinted (this was #1023).
* **Features must be package-qualified for the members** (`ferx-core/ci`, not `ci`) or the command fails outright with "none of the selected packages contains this feature" (#1114).
* **`--all` on fmt is load-bearing** — without it `crates/ferx-tools` and `crates/ferx-cli` are unchecked.
* **The public-api job needs a pinned dated nightly** read out of the script, plus `cargo-public-api` at a pinned version.

None of that is discoverable from the repo root. It is discoverable only by reading a 400-line workflow file, or by pushing and waiting.

## What it cost, concretely

On #1137 I ran the full suite under `--features ci,survival` — 158 binaries, 4634 passed, 0 failed, `cargo fmt` clean — and reported the branch verified. It was not. A single `Subject` struct literal in `src/estimation/nn_theta_gradient_tests.rs`, a file only the `ci,nn,slow-tests` step compiles, was missing two fields:

```
error[E0063]: missing fields `reset_covariates` and `reset_occasions` in initializer of `types::Subject`
```

That one line turned **three** CI jobs red — `Check`, `Clippy`, and `Tests + coverage (TTE/CTMM endpoints)` — and it had already been red through an entire previous commit without being noticed, because the local suite kept passing. Two push/wait/diagnose cycles to find something a 60-second `cargo check` would have caught before the first push.

## Suggested shape

`tools/preflight.sh`, mirroring `update-public-api.sh`'s contract — CI calls the script, so the two cannot drift:

* Default: the four fast gates — the five `cargo check` feature sets, both clippy invocations, `cargo fmt --all --check`, and `tools/update-public-api.sh --check`. All compile-only or near it; the whole thing should land in a couple of minutes warm.
* `--quick` for the check matrix alone, for use in a tight edit loop.
* Print each command before running it, and fail with the job name that would have caught it (`this is what CI's "Clippy" job runs`), so a failure teaches the mapping instead of just failing.
* Then change `ci.yml`'s `check`, `clippy` and `fmt` jobs to invoke the script rather than restating the commands — otherwise this becomes a *second* list to keep in sync, which is worse than one.

That last point is the whole value. A script that merely *documents* the commands will drift the first time someone edits the workflow; a script CI actually executes cannot.

Worth also mentioning it in `CLAUDE.md`'s Tests section, which currently describes the three test tiers but says nothing about the feature-set matrix being the thing that gates a PR.

Related: #1023 (clippy not linting test code — the same class, found the same way), #1114 (workspace split, which introduced the package-qualified feature requirement).

TeunP:
## Why

`tools/update-public-api.sh` opens with the argument this PR generalises:

> CI runs this exact script (`--check`) so the generating command cannot drift from the verifying one.

Nothing implemented that for the rest of the gate. To know whether a change would pass CI, you had to open `.github/workflows/ci.yml` and reconstruct **eight** cargo commands spread across three jobs — and every one of them is a trap for a different reason, which the workflow's own comments spell out:

* **The feature sets are not nested.** `cargo test --features ci,survival` compiles neither the `nn`-gated nor the `slow-tests`-gated source. A *fully green* suite proves nothing about roughly a third of the repo.
* **`--all-targets` is load-bearing** on clippy — without it every `#[cfg(test)]` module, every sibling `*_tests.rs`, and all of `tests/` goes unlinted (#1023).
* **The members need package-qualified features** (`ferx-core/ci`, not `ci`) or the command fails outright with "none of the selected packages contains this feature" (#1114).
* **`--all` is load-bearing** on fmt, or `crates/ferx-tools` and `crates/ferx-cli` go unchecked.

That is not discoverable from the repo root. It is discoverable by reading a 400-line workflow, or by pushing and waiting.

**What it cost, concretely.** On #1133 I ran the full suite under `--features ci,survival` — 158 binaries, 4634 passed, 0 failed, `cargo fmt` clean — and reported the branch verified. It was not. One `Subject` struct literal in `src/estimation/nn_theta_gradient_tests.rs`, a file only the `ci,nn,slow-tests` step compiles, was missing two fields:

```
error[E0063]: missing fields `reset_covariates` and `reset_occasions` in initializer of `types::Subject`
```

That single line turned **three** CI jobs red — `Check`, `Clippy`, and `Tests + coverage (TTE/CTMM endpoints)` — and had already been red through an entire previous commit, because the local suite kept passing. Two push/wait/diagnose cycles for something a 60-second `cargo check` catches.

## Read this first: the version of this PR before `2aef0ca0` did not work

A review of the earlier head found that the script **could not fail**. `run()` returned a status and left the caller to propagate it, through a group function and a `case ... esac || { on_failure; exit 1; }` in the driver. Bash suspends `errexit` for every command of an AND-OR list but the last, and carries that suspension into the entire body of a function called in that context — so the `return 1` stopped nothing and each group reported its **last** command's status.

Measured against that script with a fake `cargo`, failing one feature set at a time:

| failing `cargo check` | script exit |
|---|---|
| `ci` | 0 — "preflight OK" |
| `ci,survival,slow-tests` | 0 — "preflight OK" |
| `ci,markov` | 0 — "preflight OK" |
| `ci,nn,slow-tests` | 0 — "preflight OK" ← the #1133 line |
| `ferx-core/ci` | 1 — caught |

Same shape in `clippy`, where the `--all-targets` #1023 gate is first. **5 of 9 gate commands were unenforced**, and because `ci.yml` had just collapsed eight separate steps into three script invocations, `Check` and `Clippy` inherited it — strictly worse than `main`, where each command was its own step and failed on its own. Reproduced on bash 3.2.57 and 5.2.21 (the runner's version) and through GitHub's own `bash -e {0}` step wrapper.

Nothing could have caught it: both tests at the time ran only `--list` and `--help`, neither of which executes a gate.

The fix is not a patched link in the chain. `run` now calls `die`, which **exits**, so there is no status to propagate and no caller left that can drop it, whatever `errexit` is doing at the call site. The driver dispatches by function name with no `||`.

## What changed

**`tools/preflight.sh`** — one owner for the fast gates, four groups:

```bash
tools/preflight.sh            # fmt → check → clippy → public-api, cheapest first
tools/preflight.sh check      # just the 5-feature-set matrix, for a tight edit loop
tools/preflight.sh fmt clippy # any subset, in the order named
tools/preflight.sh --list     # print the commands, run nothing
```

**`ci.yml` calls it** — `Check` → `preflight.sh check`, `Clippy` → `preflight.sh clippy`, `Format` → `preflight.sh fmt`. This is the part that makes it work rather than rot. A script that only *documents* the commands drifts the first time somebody edits the workflow, and drifts **silently**: the local gate keeps passing while CI checks something else.

Every per-feature-set rationale comment moved from `ci.yml` into the script, so the reasoning travels with the commands.

On failure the script names the job that would have gone red:

```
preflight FAILED in group 'fmt'.
  command:  cargo fmt --all -- --check
  CI job:   Format — this is what would have gone red on the PR.
```

**`.githooks/pre-commit` execs `preflight.sh fmt`.** It was a third copy of the format gate, and the comment claiming the two "use the same flag" was false (`cargo +nightly fmt --all --check` vs `cargo fmt --all -- --check`). Hook, local gate and CI are now one list, and a rejected commit names the CI job.

**`SDLC.md` was a fourth copy, and was already wrong before this PR** — it documented `Check` as four `cargo check` invocations (there are five) and `Clippy` as `cargo clippy -- -D warnings`, which is neither the flag set CI uses nor `--all-targets`; its "Recommendations" section proposed adopting that same line as *future work*.

**`PULL_REQUEST_TEMPLATE.md`** asked for "`cargo clippy` clean" — the exact invocation #1023 showed exits 0 while six errors sit in unlinted test code. It now asks for a green `tools/preflight.sh`.

**`tools/update-public-api.sh` rejects unknown arguments.** It branched on `if [ "$1" = "--check" ]` with a bare `else` that writes, so `--chek`, `-check` or `check` overwrote `api/ferx-core-public-api.txt` with whatever the tree exposes — the one outcome the gate exists to prevent — and reported it as success. Not part of #1157's scope, but adjacent enough to be worth the five lines.

**`slow-tests.yml`** covers `tools/**`. `tools/` holds inputs the gated fits are tied to: `tools/vi-emvi-comparison/*.ferx` are the harness models the VI/EMVI anchors mirror, and `make-wide-residual-data.sh` generates the dataset `tests/nonmem/warfarin_10pct.ctl` is built from. Measured over the last 400 first-parent commits on `main`: 3 touch `tools/`, **0** would newly fire the workflow — all three also touched `src/` or `tests/`. That is an upper bound on what the entry cost *then*, not a forecast; this PR makes `tools/` the place gates are edited.

The `public-api` job still calls `tools/update-public-api.sh --check` directly — it needs a pinned dated nightly and a pinned `cargo-public-api` installed as separate workflow steps. `preflight.sh public-api` runs that same script, so a full local run still covers it.

### Also fixed in the rewrite

* Groups were enumerated in **five** places (argument validation, the default selection, the `--help` text, the driver `case`, a hardcoded `-lt 4`). One `ALL_GROUPS` array feeds all of them, and a group with no `group_<name>` function is a startup error rather than a silent no-op.
* `--help` read its own source with `sed -n '3,8p' "${BASH_SOURCE[0]}"` **after** `cd`-ing away, so a relative invocation from another directory printed nothing. It is a heredoc now, which also retires the BSD-sed `\s` portability trap the old line needed a test to guard.
* The header's cost claim carried no number. Replaced with a measured one — warm full run 34 s (fmt 13 · check 8 · clippy 2 · public-api 15) — plus the `target/` lock-contention hazard: a concurrent cargo in a sibling worktree serialises against this one and inflated an early measurement of this script several-fold.
* `--help` now states what preflight does **not** cover: `cargo check --tests` compiles the test targets and runs nothing, so `Tests + coverage (core)` can still go red on a green preflight.

## Alternatives considered

**A script CI does not run.** Rejected — that is the failure mode, not the fix. Two lists diverge, and the divergence is invisible until a push goes red.

**A `Makefile` / `justfile`.** Would need a new tool in the loop; `tools/` already holds the precedent (`update-public-api.sh`) and CI can invoke a script with no extra setup.

**Patching `run()`'s `return 1` rather than rewriting.** A three-line change fixes the exit status, but leaves the structure that produced it: 201 lines to run 9 commands, a `case … esac ||` in the driver, and the group list copied five times. The defect survived seven independent review angles and a fully green CI. Removing the propagation path is what makes the class of bug unreachable, not this instance of it.

**Keeping the five `check` steps separate in CI** for per-step names in the UI. They collapse to one step, so GitHub no longer labels which feature set failed. The script prints each command as it runs, so the log still identifies it — and the alternative reintroduces the drift this PR removes.

**Excluding `tools` from the slow-tests input-root harvest** instead of covering it. The scanner already has that mechanism (`.github/` is excluded the same way). Rejected: covering it measured free, and the exclusion route means re-shaping test prose forever so it does not look like a path.

## Tests

`tests/preflight_owns_the_fast_gates.rs`, seven tests. The load-bearing one is new:

`a_failing_gate_fails_the_script_from_every_position` puts a fake `cargo` on `PATH` and fails **each command position in turn**, asserting a non-zero exit, a diagnostic naming that command, and no success banner. Positions come from `--list`, not from a table in the test, so a sixth feature set is covered the moment it is added.

Mutation-tested — **19 mutations, 19 killed** (the previous pair killed 8 of 20):

| mutation | killed by |
|---|---|
| the original defect: `run()` returns, driver uses `\|\| { die; }` | failure-path test |
| `run()` swallows the failure outright | failure-path test |
| `die()` prints but does not exit / exits 0 | failure-path test |
| `run()` early-returns under `$CI` — CI runs 0 gates, local runs 9 | failure-path test |
| `ci,nn,slow-tests` line deleted | `--list` count |
| `group_clippy` gutted to a no-op | `--list` count |
| `--all-targets` dropped from clippy (#1023) | load-bearing-flag test |
| `--all` dropped from `cargo fmt` (#1114) | load-bearing-flag test |
| `ci,nn,slow-tests` **neutered** to plain `ci` (count unchanged) | feature-coverage test |
| same for `ci,markov` and `ci,survival,slow-tests` | feature-coverage test |
| `--help` stops listing every group | help test |
| `check` job runs `true` / inline cargo / cargo hidden in a `run: \|` block | workflow test |
| `clippy` job gets `continue-on-error` | workflow test |
| `fmt` job skips itself with an `if:` | workflow test |
| `clippy` job runs the `fmt` group | workflow test |

Three of those close blind spots the earlier guard had: it only understood `run: <cmd>` (so moving a step to a block and adding inline cargo underneath passed), it did not stop a gate job from skipping itself or passing while red, and its counts saw a gate *deleted* but not *neutered*.

### Adversarial pass on the rewrite (`a7a6c2cb`)

The rewrite, its tests and its mutation harness were written in one pass by the same author, and CI was green — which on this PR means nothing. A line read found four defects the tests could not see:

* **The diagnostic could blame the wrong CI job.** `CI_JOB` is a script global set at the top of each `group_*` function, so a group that forgets it inherits the *previous* group's value: `preflight.sh fmt check` failing inside `check` reported "CI job: Format" and sent the reader to a green log. The driver now blanks it per group. Only a multi-group run can see this, so there is a test that runs `fmt check` and fails the first `check` command.
* **The pre-commit hook hard-depended on a file that is not always there.** `core.hooksPath` points into the working tree, so the hook runs against whatever commit is checked out — including one from before #1157. It now falls back to `cargo fmt` with a notice; verified with the script moved aside, both ways.
* **`body.contains("if:")` was a false-positive machine** — a comment or a shell line mentioning the string would trip it, and a tripwire that cries wolf gets deleted rather than fixed. Now line-anchored to the YAML key position, with negative controls: a comment saying "there is no if: condition here" and a `run: |` block containing `if [ -n "$RUNNER_DEBUG" ]` both leave the suite green, while a real `if:` on the job *or* a step turns it red.
* **A fifth group would have had no count tripwire at all** — the count loop iterates the expected table, not the groups, so a new group was never looked at. Cross-checked in both directions now.

A fifth defect surfaced after that (`aa70b6b0`): making the hook delegate **silently dropped its `+nightly` pin**, and nothing in the script pinned a channel. `rust-toolchain.toml` is not the last word — an inherited `RUSTUP_TOOLCHAIN` or a `rustup override` beats it (`RUSTUP_TOOLCHAIN=stable cargo fmt --version` gives `1.9.0-stable` where a plain call gives `1.10.0-nightly`). `fmt` agrees across channels on this tree, measured; clippy would not, since whole lint groups are nightly-only. The script now mirrors `ci.yml`'s `RUSTUP_TOOLCHAIN: nightly` once, `:-` so a deliberate caller pin still wins.

11 further mutations across this pass, all as required (9 killed, both negative controls survive). The original 19 were re-run with stricter bookkeeping — each verified to have actually changed the file, to leave `bash -n`-valid syntax (so "killed" is never bash choking), and to be killed by a *named* test that plausibly detects it. 19/19 holds.

`tools/update-public-api.sh`'s no-argument regenerate path was also exercised, since the new argument gate sits in front of it and nothing had run it: it still reproduces the committed baseline byte for byte.

`.githooks/pre-commit` was checked both ways by hand: a staged misformatted `.rs` file is rejected with the CI-job diagnostic, a well-formatted one passes. `tools/update-public-api.sh --chek` now exits 2 with `api/` untouched.

- [ ] Unit test(s) added in the same module — N/A, no `src/` file is touched

## Docs

- [x] `CHANGELOG.md` `[Unreleased]` entry — N/A, CI/tooling only, no user-facing change

`CLAUDE.md` gains a **"Before you push"** section: its Tests section described the three tiers but said nothing about the feature-set matrix being what actually gates a PR. It also records the `run`-exits-rather-than-returns invariant, since that is the thing to know before editing the script.

## Checklist

- [x] `tools/preflight.sh` green (fmt · 5 check feature sets · clippy · public-API baseline)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — no Rust source touched outside `tests/`

## Cross-repo dependency

## Reviewer hints

**Read `run()` and the driver loop first.** That pair is the whole correctness story; everything else in the script is a command with a comment attached.

**Then `ci.yml`.** If the script mishandles an argument, three jobs break for everyone at once. Worth confirming the three `run:` lines name the right group and that no rationale comment was lost in the move.

Two traps worth knowing generally, both hit while building this:

* `GROUPS` is a **bash built-in array** holding the user's numeric group IDs, so `GROUPS=(fmt check ...)` is silently overridden — every group name became a GID. The array is `ALL_GROUPS`; the script carries a comment saying why.
* Wall-clock timings of `cargo` in a worktree are meaningless without checking for a concurrent cargo first: they serialise on a per-`target/` lock behind one line of output.

Closes #1157.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -10,5 +10,15 @@ git diff --cached --name-only | grep -q '\.rs$' || exit 0
 # the hook, the local gate and CI cannot disagree about what "formatted" means.
 # Repeating `cargo fmt` here is how they drifted before: the hook pinned `--all`
 # while nothing guaranteed CI did.
+#
+# The fallback is not defensive padding. `core.hooksPath` points at the hooks in
+# the WORKING TREE, so this file runs against whatever commit is checked out —
+# including one from before #1157, where `tools/preflight.sh` does not exist. A
+# bare `exec` there fails every commit with a raw "No such file or directory".
+# Fall back to the command the script would have run.
 root=$(git rev-parse --show-toplevel) || exit 1
-exec "$root/tools/preflight.sh" fmt
+if [ -x "$root/tools/preflight.sh" ]; then
+  exec "$root/tools/preflight.sh" fmt
+fi
+echo "pre-commit: no tools/preflight.sh on this checkout — running cargo fmt directly" >&2
+exec cargo fmt --all -- --check
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -5,6 +5,10 @@
 # Skip the check when no .rs files are staged (e.g. docs-only commits).
 git diff --cached --name-only | grep -q '\.rs$' || exit 0
 
-# `--all`: the repo is a workspace (#1114), and without it `cargo fmt` checks
-# only the root `ferx-core` package, silently skipping `crates/`.
-cargo +nightly fmt --all --check
+# Delegate rather than repeat the command. `tools/preflight.sh` is the single
+# owner of the fast gates (#1157) and CI's `Format` job runs this same group, so
+# the hook, the local gate and CI cannot disagree about what "formatted" means.
+# Repeating `cargo fmt` here is how they drifted before: the hook pinned `--all`
+# while nothing guaranteed CI did.
+root=$(git rev-parse --show-toplevel) || exit 1
+exec "$root/tools/preflight.sh" fmt
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -21,4 +21,7 @@ if [ -x "$root/tools/preflight.sh" ]; then
   exec "$root/tools/preflight.sh" fmt
 fi
 echo "pre-commit: no tools/preflight.sh on this checkout — running cargo fmt directly" >&2
-exec cargo fmt --all -- --check
+# `+nightly` explicitly: this arm runs on checkouts where preflight.sh does not
+# exist, so nothing else pins the channel here. It is what this hook did before
+# it delegated.
+exec cargo +nightly fmt --all -- --check
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -90,7 +90,9 @@
 - [ ] No user-visible change
 
 ## Checklist
-- [ ] `cargo clippy` clean
+- [ ] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy
+      (`--all-targets`), public-API baseline. A green `cargo test` is not a green
+      build: the feature sets are not nested (#1133, #1157)
 - [ ] Functions on the `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay differentiable (if touching gradient-reachable code)
 - [ ] `Cargo.toml` version bumped if breaking
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,48 +31,14 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           shared-key: ci
-      - name: cargo check
-        run: cargo check --tests --no-default-features --features ci
-
-      # `tests/tte_convergence.rs` is gated on BOTH `survival` and `slow-tests`, and
-      # no other per-PR job sets both — so those Tier-3 TTE tests are compiled in
-      # *neither*. Without this step a renamed public API or changed `FitResult`
-      # field would land green and only break the nightly slow-tests run, decoupled
-      # from the offending PR. Compile-only (`cargo check`, no run — the tests
-      # themselves stay nightly per the tier rules), so it is cheap.
-      # (#441 review finding #3.)
-      - name: cargo check (survival + slow-tests gated tests)
-        run: cargo check --tests --no-default-features --features ci,survival,slow-tests
-
-      # This step is also what COMPILE-GATES the whole feature-gated surface. The
-      # `Tests + coverage (TTE/CTMM endpoints)` job below deliberately builds only
-      # the handful of endpoint test binaries, so a break in any *other* test file
-      # under `--features ci,markov` (which implies `survival`) is caught here, in
-      # ~1 min, instead of by an 18-min instrumented build.
-      - name: cargo check (markov-gated module + tests)
-        run: cargo check --tests --no-default-features --features ci,markov
-
-      # Same rationale for `nn`, with `slow-tests` folded in for the same reason as
-      # the survival step above: `tests/nn_fit_smoke.rs` and
-      # `tests/nn_fit_convergence.rs` are gated on BOTH `nn` and `slow-tests`, so
-      # their bodies compile in no other per-PR job. One cheap `cargo check`
-      # compile-gates the whole `nn` surface — `src/nn`, the `[covariate_nn]`
-      # parser, the θ-gradient module, and every nn-gated test file — so a broken
-      # signature surfaces here in ~1 min rather than in the instrumented
-      # coverage job (or, worse, only in the nightly slow-tests run).
-      - name: cargo check (nn-gated module + tests)
-        run: cargo check --tests --no-default-features --features ci,nn,slow-tests
-
-      # The workspace members (#1114). `-p` rather than `--workspace` so the
-      # `ferx-core` targets above are not rebuilt, and the feature is written
-      # package-qualified (`ferx-core/ci`) because `ci` is a feature of
-      # `ferx-core`, not of the members — a bare `--features ci` here fails with
-      # "none of the selected packages contains this feature". Qualified this
-      # way it resolves to the SAME feature set as the steps above, so the
-      # members link the already-built rlib instead of triggering a second
-      # `ferx-core` compile under a different feature hash.
-      - name: cargo check (ferx-tools + ferx-cli)
-        run: cargo check -p ferx-tools -p ferx-cli --tests --no-default-features --features ferx-core/ci
+      # The five `cargo check` feature sets live in `tools/preflight.sh` (#1157),
+      # which a developer runs before pushing. CI executes that same script rather
+      # than restating the commands, so the local gate and the CI gate cannot drift
+      # — the same contract `tools/update-public-api.sh` uses for the API baseline.
+      # The script prints each command as it runs, so a failing line is still
+      # identifiable in this log; its comments carry the per-feature-set rationale.
+      - name: cargo check (5 feature sets, via tools/preflight.sh)
+        run: tools/preflight.sh check
 
   # Every default-off feature surface the `ci` build compiles out: the non-Gaussian
   # endpoint families (TTE/survival + categorical + CTMM) and the covariate-NN /
@@ -187,32 +153,11 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           shared-key: ci
-      - name: cargo clippy
-        # Include `markov` and `nn` so the CTMM module and the covariate-NN / DCM
-        # stack are linted too (both are compiled out of the base `ci` build).
-        # `markov = ["survival"]`, so this lints the whole non-Gaussian endpoint
-        # surface as well.
-        #
-        # `--all-targets` so test code is linted too. Without it the default target
-        # selection covers only the lib and the bins, and roughly a third of this
-        # repo — every `#[cfg(test)]` module, every sibling `*_tests.rs`, and all of
-        # `tests/` — got no lint coverage at all. That gap was not theoretical: it
-        # was hiding 6 `approx_constant` ERRORS (`error`, not `warning`) that the
-        # default invocation exits 0 on.
-        #
-        # Note this job installs a FRESH nightly on every run, so its lint set is
-        # newer than a developer's local toolchain unless they have updated
-        # recently. One of those 6 (`EULER_GAMMA`, `src/stats/special.rs`) is
-        # invisible to a three-month-old clippy — verify locally with an up-to-date
-        # `rustup update nightly`, or this job is the first thing that sees it.
-        run: cargo clippy --no-default-features --features ci,markov,nn --all-targets
-
-      # Same package-qualified-feature trick as the `Check` job, so the members
-      # reuse the `ferx-core` build the step above just produced. `--tests` here
-      # (unlike above) because the members' test code is a meaningful share of
-      # their line count while they are still small.
-      - name: cargo clippy (ferx-tools + ferx-cli)
-        run: cargo clippy -p ferx-tools -p ferx-cli --tests --no-default-features --features ferx-core/ci,ferx-core/markov,ferx-core/nn
+      # Both clippy invocations (ferx-core with `--all-targets`, then the members
+      # with package-qualified features) live in `tools/preflight.sh` (#1157). See
+      # the note in the `check` job above for why CI calls the script.
+      - name: cargo clippy (ferx-core + members, via tools/preflight.sh)
+        run: tools/preflight.sh clippy
 
   # A2 of #1114 — the enforcement that keeps the ferx-core/ferx-tools boundary
   # from becoming paperwork. `ferx-tools` can only touch `pub` items, so the way
@@ -269,11 +214,8 @@ jobs:
       - name: Install nightly toolchain
         run: rustup toolchain install nightly --profile minimal -c rustfmt
 
-      - name: cargo fmt
-        # `--all`: without it `cargo fmt` formats only the root package, so
-        # `crates/ferx-tools` and `crates/ferx-cli` would go unchecked (#1114).
-        # `.githooks/pre-commit` uses the same flag.
-        run: cargo fmt --all -- --check
+      - name: cargo fmt (via tools/preflight.sh)
+        run: tools/preflight.sh fmt
 
   coverage:
     name: Tests + coverage (full, slow-tests)
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -29,6 +29,15 @@ on:
       # bought 4 skipped runs in 400 (1% — one of them a `Cargo.lock` bump that should run
       # anyway) while silently missing 67. That is the trade, and it is not close.
       #
+      # `tools/**` was added for #1157: `tools/preflight.sh` is executed by CI itself, and
+      # naming it in a test's assertion message is enough for the harvest below to derive
+      # `tools` as an input root. Measured before adding it, the same way the numbers above
+      # were: over the last 400 first-parent commits on `main`, 3 touch `tools/` and **0**
+      # would newly fire this workflow — every one of the three also touched `src/` or
+      # `tests/`, so they already did. Cost of covering it: zero runs. The alternative was to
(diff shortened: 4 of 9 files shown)


TeunP:
## Adversarial review of the rewrite — `2aef0ca0..aa70b6b0`

The rewrite in `2aef0ca0`, its tests, and the mutation harness that scored them were written in one pass by the same author, and CI was green. On this PR green CI is worth nothing as evidence: the head before `2aef0ca0` was green while five of nine gate commands could not fail. So this is an independent line read of my own work, plus a re-audit of the claim the first pass made about itself.

### First: was "19/19 mutations killed" true?

A mutation whose `str.replace()` matched nothing is a no-op, and one that breaks bash syntax turns the suite red for the wrong reason — both would read as "killed". Re-ran all 19 recording, per mutation: whether the file actually changed, whether the result still passes `bash -n`, and **which named test** failed.

All 19 really applied, all left valid bash, all were killed by a named test that plausibly detects that defect — and by the right one (failure-path test for the exit-semantics mutations, coverage test for the neuters, workflow test for the YAML ones). The claim holds.

### Findings — five, all fixed

**1. The failure diagnostic could name the wrong CI job.** (`a7a6c2cb`)

`CI_JOB` is a script global assigned at the top of each `group_*` function, so a group that forgets the assignment inherits the *previous* group's value. `tools/preflight.sh fmt check` failing inside `check` reported `CI job: Format` — pointing confidently at a job that had just passed. Worse than no diagnostic, because it sends the reader to a green log.

Only a multi-group run can see this; the per-position test runs one group per process, where a stale value is unreachable by construction. The driver now blanks `CI_JOB` before each group, `die` says so explicitly when a group set none, and `the_failure_diagnostic_names_the_group_that_actually_failed` runs `fmt check` and fails the first `check` command.

**2. The pre-commit hook became a hard dependency on a file that is not always present.** (`a7a6c2cb`)

`core.hooksPath` points into the **working tree**, so the hook executes against whatever commit is checked out — including one from before #1157, where `tools/preflight.sh` does not exist and a bare `exec` fails every commit with a raw `No such file or directory`. Now falls back with a notice. Verified with the script moved aside: passes a formatted file, rejects a misformatted one, prints the notice, and restores cleanly.

**3. `body.contains("if:")` was a false-positive machine.** (`a7a6c2cb`)

A comment or a shell line merely mentioning the string would trip it, and a tripwire that cries wolf gets deleted rather than fixed. Line-anchored to the YAML key position now, with **negative controls** in the mutation set — a comment reading `there is no if: condition here` and a `run: |` block containing `if [ -n "$RUNNER_DEBUG" ]; then …` must both leave the suite green, while a real `if:` on the job *or* on a step must turn it red. All four behave.

**4. A fifth group would have had no count tripwire at all.** (`a7a6c2cb`)

The count loop iterates the expected table, not the groups, so a group added to `ALL_GROUPS` was simply never looked at — the silent-shrinkage failure one level up from the one the counts were added to catch. `--list`'s banners are now cross-checked against the table in both directions.

**5. Delegating the hook silently dropped its toolchain pin.** (`aa70b6b0`)

`.githooks/pre-commit` ran `cargo +nightly fmt`. Making it call `preflight.sh fmt` dropped the `+nightly`, the fallback arm added in fix 2 dropped it again, and nothing in the script pinned a channel. `rust-toolchain.toml` says `nightly` but is not the last word — an inherited `RUSTUP_TOOLCHAIN`, or a `rustup override` on the directory or any parent, beats it:

```
$ cargo fmt --version                          -> rustfmt 1.10.0-nightly
$ RUSTUP_TOOLCHAIN=stable cargo fmt --version  -> rustfmt 1.9.0-stable
```

`fmt` happens to agree today — stable rustfmt produces zero diff on this tree, measured, since the repo has no `rustfmt.toml`. Clippy is where it would bite: whole lint groups are nightly-only, so a stable run is quietly laxer. That is the local/CI split this script exists to remove, reintroduced inside it.

The script now mirrors `ci.yml`'s workflow-level `RUSTUP_TOOLCHAIN: nightly` once, covering all four groups, rather than sprinkling `+nightly` per command. `:-` not `=`, so a deliberate caller pin still wins. `update-public-api.sh` is unaffected — it sets its own dated nightly inline on the `cargo public-api` invocation, which overrides an exported value.

### Other checks run

* **`update-public-api.sh`'s no-argument regenerate arm.** The argument gate added in `716cc1f8` sits directly in front of it and nothing had exercised it. Run: reproduces the committed baseline byte for byte.
* **`shellcheck -S style`** on `preflight.sh`, `pre-commit` and `update-public-api.sh` — clean.
* **bash 3.2.57 and 5.2.21** (the runner's version) for the failure sweep, plus GitHub's own `bash -e {0}` step wrapper.
* The guard tests were confirmed to have **actually executed** on the Linux runner, by reading the `Tests + coverage (core)` job log rather than trusting the job's green tick.

Guard is now 7 tests. New mutation total for this pass: 11 cases, all as required (9 killed, 2 negative controls survive).

### Still not verified

* **The `public-api` position has no failure test.** Its command is invoked by path, not through `PATH`, so shimming it means letting the real script hunt for a pinned toolchain and build rustdoc JSON. It goes through the same `run` helper the other eight positions prove — an argument, not a measurement.
* **`--all-targets`, `--all` and the check matrix's feature union are pinned; the rest of each command line is not.** Swapping `--tests` for `--lib`, or dropping `-p ferx-cli`, keeps the count and the feature union intact and would pass. Pinning each command whole would make the test a second copy of the list, which is the thing the design rejects; the tripwires cover the flags with incidents behind them (#1023, #1114, #1133) and nothing else.
* **No local run of the full `--workspace --tests` suite.** Two attempts deadlocked at 0% CPU behind other cargo processes on the shared `target/` lock. CI ran the identical command set on a clean checkout and passed, so nothing is unchecked — but I do not have an independent local confirmation and am not claiming one.

Issue #1075, opened 2026-08-25
bug(sens): moving-boundary saltations drop the explicit time-derivative of the field when a pointwise R_in reads the covariate snapshot


TeunP:
Found and measured while responding to the #1071 review. It is the residual left after #1060's fix, deliberately not landed there.

## The term

A saltation at a moving boundary needs the second time-derivative of the state on each side:

```
ẍ = df/dt = ∂f/∂t + J·f
```

Every current site computes only `J·f`. For an autonomous field that is complete, and every fixture before #1071 had one — a constant infusion rate and a zero-order window are both constant in `t`, and the bare user RHS is autonomous unless it reads `TIME`/`TAD`.

A **pointwise input-rate forcing** is not. `R_in(tad)` is an explicit function of time, so `∂f/∂t = ∂R_in/∂tad ≠ 0`. Carrying the shift-and-flow-back derivation through with the explicit term gives

```
u += (v⁻ − v⁺)·δ + [½J⁻v⁻ + ½J⁺v⁺ − J⁺v⁻ + ½(∂f⁻/∂t − ∂f⁺/∂t)]·δ²
                                            ^^^^^^^^^^^^^^^^^^^^ missing
```

It is the **difference**, so it vanishes identically whenever the two sides read the same PK snapshot — which is why it has never shown up. #1060 made the two sides read different snapshots, and it became visible.

## Measured

Fixture `ode_provider_lagtime_tvcov_arrival_straddles_input_rate_matches_production` (`src/sens/ode_provider_tests.rs`): a `first_order` kernel whose `KA` carries the weight covariate, feeding a compartment the lagged bolus does not land in, active across the arrival.

| obs | `∂²f/∂η_LAG²` analytic | production double-FD | rel |
|---|---|---|---|
| t = 2 | `1.01246955` | `1.01844713` | 5.9e-3 |
| t = 4 | `0.62820541` | `0.65211094` | 3.7e-2 |
| t = 8 | `0.18203608` | `0.19444349` | **6.4e-2** |

First order is exact — with `KA` on the covariate the forcing itself differs across the boundary, and `(v⁻ − v⁺)·δ` now carries that correctly (it was a 5.0 % *gradient* error before #1071).

[...]

PR #623, opened 2026-06-30
feat(ode): analytic M3 BLOQ, IOV, and the iiv_on_ruv triple for ODE models (#486)

hiddevandebeek:
## Why

The **ODE** path lagged the closed-form path on the M3 BLOQ / IOV / `iiv_on_ruv` matrix: censored data **+ IOV**, IIV-on-residual-error **+ IOV**, and the **M3 + IOV + `iiv_on_ruv`** triple all fell back to finite differences, even though closed-form made them analytic in #580/#591/#579. This brings the user-`[odes]` variants onto the same exact analytic sensitivity path for FOCEI and non-interaction FOCE. Status matrix #486.

## What changed

The censored / residual-error machinery is **provider-agnostic**: the inner (`residual_inner_obs`) and outer (`prepare_stacked`) assemblies apply the censored `−logΦ` coefficient and the `exp(2·η_ruv)` variance scaling keyed on `subject.cens[j]` / `residual_error_eta`, over whatever `ObsSens`/`ObsGrad` the walk emits — and the ODE IOV walk already emits the standard per-observation shape (with a structural zero `∂f/∂η_ruv` column, since `η_ruv` is variance-only). So the substantive change is **relaxing the gates**, not new gradient assembly:

- **`ode_iov_supported`** (`src/sens/ode_provider.rs`): removed the two stale decline clauses — the **M3 BLOQ** clause and the **`iiv_on_ruv`** clause. Both gates open the inner *and* outer analytic loops at once (`iov_sens_supported` → `ode_iov_supported`). The stale comments claiming the assembly lacked the censored / residual-eta terms are corrected.
- **`iiv_on_ruv_forces_fd`** (`src/types.rs`): relaxed to a uniform `false` — the **IOV** triple (`n_kappa > 0`) **and** the **non-IOV** ODE M3 + `iiv_on_ruv` combination (the last holdout, see below) are both analytic, so no `iiv_on_ruv` combination forces FD any more.
- **Gradient-optimizer guard** (`src/estimation/outer_optimizer.rs`): fixed a guard that mis-handled a rejected inner-EBE trial step on the newly-admitted ODE IOV + censored path.

[...]

PR #1164, opened 2026-08-31
docs: split the five largest sections into addressable subsections [closes #1162]

### LINKED ISSUE #1162: docs: split the largest sections into addressable subsections (agent discovery)
## Problem

A handful of doc sections have grown into single monolithic units — one flat table, or one
continuous wall of prose — under a single heading. They are the largest addressable chunks in
the site, and they are still growing:

| chars | tokens | page | section | at v0.3.0 |
|------:|-------:|------|---------|----------:|
| 28,374 | 11,091 | `model-file/fit-options` | `#general-options` | 22,382 |
| 22,575 | 8,449 | `estimation/foce` | `#gradient-route-analytic-vs-fd` | 18,418 |
| 11,624 | 4,725 | `estimation/saem` | `#4-m-step-for-theta-and-sigma-optimization` | |
| 9,507 | 3,427 | `model-file/ode-models` | `#letting-ferx-pick-the-stepper-ode_method--auto` | |
| 8,722 | 3,764 | `file-formats/check-report` | `#error--warning-codes` | |

This hurts two readers at once:

- **Agents.** A section is the retrieval unit. An agent that wants to know how to set
  `saem_k1` pays for all 36 `[fit_options]` keys, ODE tolerances and checkpointing included.
  An agent that wants to know *whether to override `gradient_method`* pays for 8,449 tokens
  of scope-history prose about which #PR made which feature combination analytic.
- **Humans.** There is no way to link to "the covariance keys" or "the FD fallback list" —
  only to the whole thing.

The fix is the same in every case: **split by concern into independently addressable
subheadings**, so each is a retrieval unit on its own and a link target on its own.
This is a structural edit, not a rewrite — content is preserved, and where prose is
converted to a table the wording moves rather than being re-derived.

Do at least the first five. The remaining entries from the long-section scan are listed at
the bottom as follow-up scope.

## Cross-cutting requirements

1. **Anchor stability.** Splitting changes anchors. Every inbound link must be re-pointed
   in the same PR. Known today: `docs/estimation/foce.qmd:341` →
   `../model-file/fit-options.qmd#general-options`. Grep before and after:
   ```bash
   grep -rn 'general-options\|gradient-route\|m-step-for-theta\|letting-ferx-pick\|error--warning-codes' docs/
   ```
   Where a heading is likely to be linked from outside the repo (ferx-r docs, issues, PR
   bodies), give the new subsections **explicit `{#id}` anchors** rather than relying on
   Quarto's slugification of the heading text, so a later wording change does not break them
   again.
2. **`toc-depth` is 3** (`docs/_quarto.yml:169`). A split that lands the new subsections at
   H4 puts them *out* of the on-page TOC — which defeats half the point. Either promote the
   parent to H2 so the children are H3, or bump `toc-depth` to 4. Decide per page; do not
   silently create invisible H4s.
3. **No content loss.** These sections encode hard-won detail (scope caveats, PR numbers,
   NONMEM cross-checks). The split must be lossless. A diff that only moves lines and adds
   headings is the goal.
4. Add new sections to `docs/_quarto.yml` only if a split creates a genuinely new *page*.
   Prefer in-page subheadings; a new page is warranted only where noted below.

### 1. `model-file/fit-options` — `#general-options` (28,374 chars)

One flat table, 36 keys, no internal structure. Split into **six subheadings, six keys
each**, grouped by the machinery each key controls:

| Subheading | Keys |
|---|---|
| Run control | `method`, `threads`, `maxiter`, `checkpoint`, `checkpoint_interval_secs`, `optimizer_trace` |
| Outer optimizer | `optimizer`, `outer_xtol`, `outer_ftol`, `steihaug_max_iters`, `global_search`, `global_maxeval` |
| Inner loop (EBE) | `inner_maxiter`, `inner_optimizer`, `inner_tol`, `inner_restarts`, `reconverge_gradient_interval`, `mu_referencing` |
| Covariance / standard errors | `covariance`, `covariance_method`, `covariance_fallback`, `analytic_cov_hessian`, `fd_hessian_step`, `cov_inner_tol` |
| ODE solver | `ode_reltol`, `ode_abstol`, `ode_max_steps`, `ode_stiff_abort_after`, `ode_method`, `ode_auto_switch` |
| Data handling & diagnostics | `bloq_method`, `npde_nsim`, `npde_seed`, `iov_column`, `iov_occasion`, `inits_from_nca` |

Each group becomes its own `###` with its own small table. Keep `## General Options` as the
parent H2 with a one-paragraph orientation and a link list to the six groups, so the existing
`#general-options` anchor still lands somewhere sensible.

Note the page *already* has `## SAEM-Specific Options`, `## VI-Specific Options`,
`## AGQ-Specific Options`, `## Importance Sampling (IMP)`, `## IMPMAP`, `## SIR` — so the
method-specific axis is done. What is missing is structure inside the shared keys, which is
what the table above supplies. (Two of the existing method sections — `#vi-specific-options`
at 5,152 chars — appear further down the long-section list and can be left for the follow-up
pass.)

### 2. `estimation/foce` — `#gradient-route-analytic-vs-fd` (22,575 chars)

The clearest case. Roughly 13,000 characters of it are **a single paragraph**: an accreted
changelog of which feature combination became analytic in which PR. The operative content —
*leave it at `auto`; the banner tells you what actually resolved; force `fd` only if you want
to* — is three sentences buried at the end of that paragraph.

Separate **which route to pick** from **why the routes differ**:

- `### Gradient route` (parent) — the decision, in ~10 lines: `auto` is the default and is
  exact-and-faster; `fd` is the escape hatch; the route is resolved **per subject**.
- `### Reading the resolved route` — the `gradient:` startup-banner block, the per-route
  subject counts (`analytic (Dual2) ×118, FD ×3`), which estimators print it, and the
  `sampler:` line for SAEM.
- `### Analytic scope` — **as a table, not prose.** One row per feature or combination
  (model class × feature × IOV) with columns *outer gradient* / *inner EBE gradient* /
  *since*. The paragraph already contains exactly this data; it is a table written as a
  sentence.
- `### Known FD fallbacks` — likewise a table: `weibull` + estimated lagtime, SS into a
  `zero_order` window, SS + absorption lagtime, `RATE=-1` under `F ≠ 1`, ODE M3 + IOV +
  `iiv_on_ruv`, LTBS × IOV on ODE, `TAD` + lagtime (#1070), SS + non-autonomous RHS,
  `input_rate` + `[diffusion]`, model-size caps. Each with its reason in one clause.
- `### Non-differentiable coincidences` — the moving-boundary material (sampling exactly at
  an infusion end / lagged arrival, one-sided vs central-difference, why analytic and FD
  legitimately disagree there). This is self-contained and currently buried.
- `### Runaway EBEs are re-checked (#958)` — already its own logical unit at the end;
  promote it to a heading.

The scope table is the high-value part: it is what an agent needs, and it is currently
unretrievable as anything smaller than 8,449 tokens.

Consider promoting `## Gradient route` to H2 given the size (see requirement 2 on
`toc-depth`).

### 3. `estimation/saem` — `#4-m-step-for-theta-and-sigma-optimization` (11,624 chars)

Currently one H3 carrying four distinct things. Split into:

- the M-step objective itself (the conditional-NLL formula);
- `#### Closed-form EM update for mu-referenced thetas` — the `log θ ← log θ + γ·η̄` update,
  eta re-centring, and the pinning of closed-form thetas for the NLopt call;
- `#### The NLopt M-step (BOBYQA)` — why BOBYQA and not SLSQP, with the dense-Emax
  identifiability-ridge evidence (EMAX under-estimated ~40%, ~40% faster, ΔOFV < 0.1 on
  PK-only models);
- `#### The mu_referencing = false path` — the full-NLopt path, plus the callout that follows it.

### 4. `model-file/ode-models` — `#letting-ferx-pick-the-stepper-ode_method--auto` (9,507 chars)

Already well-organised internally — it has bolded pseudo-headings that are doing a heading's
job without being addressable. Promote them:

- `#### The stiffness probe` — the Jacobian eigen-read, `λ_max ≥ 30`, `rodas4` /
  `rodas5p @ ode_reltol ≤ 1e-8`, and how to opt out by naming a stepper;
- `#### It probes per integration segment` — the binding-model zero-at-t0 argument;
- `#### A failed escalation is caught and undone` — the discard-and-re-solve guard,
  `auto_fallback_failed`, and that the guard applies only to `auto`;
- `#### What the probe cannot see` — the time-unit sensitivity (`ka = 0.6/min` → `λ_max` 36)
  and speed-vs-separation (`ktr = 50` transit chain), plus why both were accepted.

Sibling `#which-regime-am-i-in` (7,360) and `#mid-segment-switching` on the same page are in
the follow-up list; a single pass over this page's solver material could take all three.

### 5. `file-formats/check-report` — `#error--warning-codes` (8,722 chars)

A flat 39-row table of every `E_*` / `W_*` code. Group it by what the code is *about*, one
subheading and one table each:

| Subheading | Codes |
|---|---|
| Model-file structure | `E_PARSE`, `E_MISSING_BLOCK`, `E_UNKNOWN_BLOCK`, `E_DEPRECATED_BLOCK`, `E_BLOCK_INSTANCE_NAME`, `E_BLOCK_FEATURE_DISABLED`, `E_NN_FEATURE_DISABLED` |
| Parameters & random effects | `E_ETA_NOT_DECLARED`, `W_ETA_NOT_DECLARED`, `E_SIGMA_ORDER_MISMATCH`, `E_IOV_MISSING_OCC`, `W_IOV_OCC_MISSING` |
| Method / optimizer compatibility | `E_SAEM_NO_RANDOM_EFFECTS`, `E_METHOD_NO_RANDOM_EFFECTS`, `W_GN_NO_RANDOM_EFFECTS`, `E_IMP_CHAIN`, `E_OPTIMIZER_IOV`, `E_OPTIMIZER_AGQ`, `E_SDE_INCOMPATIBLE` |
| Data & dosing | `E_DATA`, `E_MISSING_COVARIATE`, `E_NONFINITE_DV`, `W_MISSING_DV`, `W_DESIGN_DV`, `W_ADDL_MISSING_II`, `W_STEADY_STATE_II`, `W_STEADY_STATE_INFUSION`, `E_DOSE_ATTR_DOUBLE_USE`, `W_NEGATIVE_LAGTIME` |
| Scaling, output & derived | `E_PER_CMT_SCALING`, `E_PER_CMT_ERROR_MODEL`, `E_DERIVED_NAME_CONFLICT`, `W_DERIVED_COVARIATE_SHADOW`, `W_DERIVED_STEP_IGNORED`, `E_OUTPUT_UNKNOWN_COLUMN`, `W_OUTPUT_DUPLICATE` |
| Experimental / degraded paths | `W_EXPERIMENTAL_SDE`, `W_EXPERIMENTAL_NN`, `W_SDE_RESET`, `W_ABSORPTION_TWIN_DECLINED` |

Keep the closing "codes are stable; treat an unrecognised code as a generic finding of its
`severity`" note under the parent heading, not inside a group.

Two coordination points:

- `warnings.qmd#warning-codes` (7,427 chars) is the same content family and is #7 on the
  long-section list. **Use the same grouping in both** so the two pages stay parallel; doing
  them in one PR is cheaper than doing them twice.
- #1051 (the `E_*`/`W_*` registry drift — 62 of 98 emitted codes unregistered) will churn
  these tables. Either land this split first and let #1051 add rows into the right groups, or
  sequence them deliberately. Note the ordering choice in the PR.

## Follow-up scope (not required by this issue)

Remaining sections from the same scan, for a second pass:

| chars | page | section |
|------:|------|---------|
| 7,671 | `model-file/adaptive-dosing` | `#current-scope-and-limits` |
| 7,427 | `warnings` | `#warning-codes` |
| 7,381 | `model-file/error-model` | `#examples` |
| 7,360 | `model-file/ode-models` | `#which-regime-am-i-in` |
| 6,453 | `model-file/absorption` | `#analytic-closed-form--pk-one_cpt_transit…` |
| 5,778 | `model-file/scaling` | `#interaction-with-gradients` |
| 5,302 | `estimation/foce` | `#how-the-hessian-is-built` |
| 5,285 | `estimation/tte` | `#competing-risks-cause-specific-hazards` |
| 5,261 | `model-file/adaptive-dosing` | `#keys` |
| 5,215 | `estimation/vi` | `#how-σ-is-updated` |
| 5,152 | `model-file/fit-options` | `#vi-specific-options` |

## Acceptance criteria

- [ ] The five sections above are each split into subheadings, with the content preserved.
- [ ] `#gradient-route-analytic-vs-fd`'s scope prose is rendered as a table (analytic scope)
      plus a table (FD fallbacks).
- [ ] No section from the five exceeds ~6,000 characters after the split.
- [ ] Every inbound link to a changed anchor is re-pointed; `grep` clean.
- [ ] New subsections are visible in the on-page TOC (H3 at `toc-depth: 3`, or `toc-depth`
      raised deliberately).
- [ ] `quarto render docs` succeeds locally with no broken-link warnings.
- [ ] No rendered output (`docs/_site/`) committed.

roninsightrx:
## Why

Closes #1162.

Five doc sections had grown into single monolithic units — one flat table, or one continuous
wall of prose, under one heading. They are the largest addressable chunks on the site, and they
were still growing (`#general-options` went 22,382 → 28,374 chars since v0.3.0;
`#gradient-route-analytic-vs-fd` 18,418 → 22,575).

That hurts two readers at once. **Agents**, because a section is the retrieval unit: an agent
looking up `saem_k1` paid for all 36 `[fit_options]` keys including ODE tolerances and
checkpointing, and an agent asking *should I override `gradient_method`?* paid 8,449 tokens of
PR-by-PR scope history to reach a three-sentence answer. **Humans**, because there was no way to
link to "the covariance keys" or "the FD fallback list" — only to the whole thing.

## What changed

Three things, in order: the split #1162 asked for, then a layout bug the split made visible,
then a trim of the three heaviest key descriptions.

### 1. The split

| Page / section | Before | After (largest subsection) |
|---|---:|---:|
| `model-file/fit-options` `#general-options` | 28,374 | 4,882 |
| `estimation/foce` `#gradient-route-analytic-vs-fd` | 22,575 | 5,161 |
| `estimation/saem` `#4-m-step-for-theta-and-sigma-optimization` | 11,624 | 5,806 |
| `model-file/ode-models` `#letting-ferx-pick-the-stepper-ode_method--auto` | 9,507 | 2,724 |
| `file-formats/check-report` `#error--warning-codes` | 8,722 | 2,360 |

**`[fit_options]` General Options** — 36 keys in one flat table, split into seven `###` groups by
the machinery each key controls: run control, outer optimizer, inner loop (EBE), covariance and
standard errors, ODE solver tolerances, ODE stepper selection, data handling and diagnostics.
`## General Options` stays as the parent with a link list, so the existing `#general-options`
anchor still lands somewhere useful. The method-specific axis already existed on this page
(`## SAEM-Specific Options` etc.) — what was missing was structure inside the *shared* keys.

**FOCE gradient route** — the clearest case: ~13,000 of its 22,575 characters were a **single
paragraph**, an accreted changelog of which feature combination became analytic in which PR. The
operative content (*leave it at `auto`; the banner tells you what actually resolved; force `fd`
only if you want to*) was three sentences buried at the end of it. Promoted to `##` and split so
the decision leads: the decision and a nav list; `### Reading the resolved route`; four
`### Analytic scope: …` sections; `### Known FD fallbacks` as **a table**; then moving boundaries,
model-size caps, non-finite objectives and the #958 runaway-EBE check, each already a
self-contained unit without a heading. All 46 sentences of the original paragraph are asserted
present in the output by the split script.

**SAEM theta/sigma M-step** — closed-form EM update, the BOBYQA M-step, and the thetas the
closed-form update cannot reach. `### What SAEM does about it now` was an H3 orphan sitting
*inside* section 4's span; it becomes `#### Damping the numerical M-step (mstep_damping)` at the
right level, keeping its old anchor id explicitly.

**`ode_method = auto`** — its bolded pseudo-headings were already doing a heading's job without
being addressable, so they became real headings. Promoted H4 → H3 (with its two siblings) so its
children can be H4 rather than invisible H5s.

**check-report `E_*` / `W_*` codes** — a flat 40-row table grouped into six tables by subject.

### 2. The layout bug the split made visible

On a 1512px viewport the content column was 749px while the *Inner loop (EBE)* table's
**min-content** width is 838px, so it rendered 89px past the column and **63px into the "On this
page" TOC** (measured in Chrome: content right edge 1123, TOC left edge 1149, table right edge
1212). The min-content floor comes from long unbreakable inline code tokens —
`reconverge_gradient_interval` alone is 258px.

Nothing caught it because `styles.css` set `display: table` with `overflow-x: auto` above the
mobile breakpoint, and **`overflow-x` is inert on `display: table`** — that rule never contained
anything. Two changes: `grid.body-width` 800px → 1000px (content column 749px → 895px), and the
table rule now uses `display: block; width: fit-content; max-width: 100%; overflow-x: auto`,
which keeps shrink-to-fit for narrow tables (the reason the old rule reverted to `display: table`)
and scrolls a genuinely too-wide one.

| viewport | content col | tables overlapping TOC | tables scrolling |
|---|---:|---:|---:|
| 1512 (before) | 749px | **1** (63px into TOC) | — |
| 1512 (after) | 895px | 0 | 0 |
| 1280 (after) | 663px | 0 | 3 (internal scroll) |

Rejected: `overflow-wrap: anywhere` on table code spans. It lets tables shrink at any width but
shatters `reconverge_gradient_interval` down the Key column ~3 characters per line.

### 3. Trimming the three heaviest key descriptions

After the split the largest remaining sections were single table *cells*, not topics.

| section | before | after |
|---|---:|---:|
| `### ODE stepper selection` | 909 words | **367** |
| `### Inner loop (EBE)` | 769 words | **365** |
| `## VI-Specific Options` | 747 words | **378** |

These needed different treatment, because the duplication picture differs:

- **ODE stepper selection** — `ode_method` alone held 3,467 chars, and most of it restated
  `ode-models.qmd`. Each removed claim was checked to exist elsewhere first: how `auto` decides
  → `ode-models.qmd:293/329/395/404`; the *entire* `ode_auto_switch` cell →
  `#mid-segment-switching`, which covers all of it in more depth; vern7's Savic numbers →
  `:561/:568`; "full peers / continuous extension" → `:592`. One claim was **not** duplicated
  anywhere and is preserved, promoted into its own subsection: a Rosenbrock step's `n × n`
  factorization is sized by the system actually integrated, so a CTMM `s`-state chain factors an
  `s² × s²` matrix (5-state chain → 25×25 LU).
- **Inner loop (EBE)** — nothing here duplicates another page (grep found no coverage of
  `inner_restarts` or `inner_tol` anywhere else), so this is a **move, not a cut**: the two heavy
  cells keep a one-line summary and their prose moves *verbatim* into `#choosing-inner-tol` and
  `#escaping-a-multimodal-individual-objective`, with byte-for-byte preservation asserted by a
  check script.
- **VI-Specific Options** — here the duplication is real: `vi.qmd` carries the same twelve keys
  *plus* a dedicated section per key. Each row is now one line of meaning and a deep link. All 19
  dropped claims were checked against `vi.qmd` individually: 15 covered, 3 restored because
  nothing else names them (`vi$elbo_trace`, `vi$kl`, and the sublinear-cost figure — 4× draws for
  ~2.3× wall time), and 1 false alarm (the Adam second-moment freeze *is* covered under
  `vi.qmd#grad-clip → The mechanism`; my grep needle was wrong).

The largest section on the page is now `### Covariance and standard errors` at 644 words — an
ordinary-sized reference section, not an outlier.

### 4. Merged `origin/main` (#1070)

`main` moved 21 commits while this was open, and the one conflict — `docs/estimation/foce.qmd`
— was **semantic, not textual**. #1070 landed and *reverses* a claim this branch had promoted
into the FD-fallback table: an `[odes]` RHS reading the `TAD` built-in under an estimated
lagtime is now **analytic** on both loops and under IOV.

Resolved by keeping this branch's structure and re-placing main's rewritten sentences into it:
the `TAD` row is dropped from the FD table and main's rewritten TAD block moves into
*Analytic scope: models and readouts*, beside the `TIME` built-in it now behaves like; main's
"`TAD` + steady state is the one cell that still routes to FD" sits next to it, with its reason
carried onto the `SS + non-autonomous RHS` row that already covered that case; main's widened
#1075 text (the explicit `∂f/∂t` term is uncarried for *any* non-autonomous field, not only the
co-timed geometry) replaces this branch's narrower sentence; and main's new
second-order-residual paragraph joins *Moving boundaries*.

Checked afterwards: 47 of main's 48 paragraph sentences and 4 of its 5 new-paragraph sentences
are present verbatim — the two absent ones are the pronoun rewordings this branch made — and all
three stale pre-#1070 claims are gone. The branch touches no Rust; `git diff origin/main` is
docs + `CHANGELOG.md` only.

## Alternatives considered

- **Leaving the new subsections at H4 without touching `toc-depth`.** Rejected: an unnavigable
  heading is only half a fix. `toc-depth` 3 → 4 affects 16 H4 headings site-wide against 332 H3s,
  so TOCs stay short — and the H4s that already existed become navigable as a side effect.
- **Rewriting the gradient-route prose instead of re-cutting it.** Rejected: that content encodes
  hard-won scope caveats and PR numbers, and a rewrite would quietly drop some.
- **Doing `warnings.qmd#warning-codes` in the same PR**, as #1162 suggested. Dropped after
  looking: it is a *different* taxonomy (runtime `WarningCode` categories), not the same content
  family as the check-report static-check codes. The issue is wrong about that.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

## Performance

## Checklist
- [x] `tools/preflight.sh` — not run: this branch changes no Rust (`git diff origin/main` is
      `docs/**` plus `CHANGELOG.md`), so fmt / check / clippy / public-API are untouched by it
- [x] Not touching gradient-reachable code
- [x] No version bump needed

## Tests
- [x] No new tests needed (why: docs-only; verification is the render + link check below)

Verification actually run:

- `quarto render` on the full 85-page site — clean. (Note: `quarto render docs` produces *nothing*
  from inside a worktree, because Quarto skips dot-directories when globbing project inputs and
  the worktree lives under `.claude/worktrees/`. `quarto inspect` shows `files.input: 0`. Not
  caused by this branch; the preview was built from a copy at a non-dot path.)
- **Site-wide link check over all 85 rendered pages: 8,219 internal page links and 2,309 anchors,
  0 broken page targets.** 36 broken anchors found, all pre-existing; the 8 on pages this PR
  touches are fixed, 33 → outside its scope remain (see Open questions).
- Layout verified in Chrome at 1512px and 1280px (table above).
- Losslessness: every non-blank original line asserted present verbatim, except the listed
  intentional edits. Sentence-level (46/46) for the FOCE paragraph; byte-for-byte for the two
  moved inner-loop cells; claim-by-claim for the trims.

- [x] Not applicable

## Docs
- [x] No new pages, so no `_quarto.yml` sidebar entries; `docs/_site/` not committed
- [ ] No user-visible change

### ferx-site

### ferx-book

### Example execution
- [x] No example execution step needed (docs-only)

## Reviewer hints

Focus on **`estimation/foce.qmd`** — the only file where content was re-ordered rather than
re-headed. The diff is large but mechanical: the giant paragraph's sentences were bucketed by
theme and rejoined, so reading the *rendered* page is a faster review than reading the diff.

Second, the **`### Known FD fallbacks` table** — the one genuinely new content. Please
sanity-check its rows against the scope prose, since it is an index I derived rather than text
that was already there.

Third, **`docs/styles.css` and `_quarto.yml`** are four lines but affect every page on the site,
so worth loading a page you care about.

The rest can be skimmed: heading insertions and table row regrouping, with row text untouched.

## Open questions

**Building the FD-fallback index surfaced two places where the existing prose contradicts
itself**, so I left both off the table with a note rather than adjudicate:

1. **ODE `iiv_on_ruv` × M3 BLOQ.** One sentence lists it as an inner-EBE-only FD fallback ("the
   outer θ/Ω/σ gradient still serves it"); another says "the non-IOV ODE `iiv_on_ruv` × M3
   combination is analytic too (since #623)".
2. **Closed-form steady-state dose + estimated lagtime.** The lagtime paragraph calls it one of
   "two deliberate exceptions [that] remain on FD"; the steady-state sentence says SS is analytic
   "including combined with … an **estimated lagtime** (each closed by #486)".

Someone who knows `sens/` should say which is current — a one-line fix once known, and worth its
own issue if the code is the thing that is wrong.

**33 broken anchors remain elsewhere on the site**, all pre-existing and outside this PR's
sections — mostly `development/sdlc.html`'s numbered headings (`#5-git-workflow`, `#9-releases`,
…) linked from `contributing.html` and its own TOC, plus one each in `optimizers.qmd`,
`absorption.qmd` and `error-model.qmd`. Happy to open an issue.

https://claude.ai/code/session_01KrvW4HAF8jc4XLhj24WK8c


roninsightrx:

Scope: `git diff main...HEAD` — 39 source files touched, **comment lines only** (verified: no non-comment line changed under `src/`), plus the new `rustdoc` group in `tools/preflight.sh`, the new `Rustdoc` CI job, and the guard-test updates.

Verification run: `tools/preflight.sh rustdoc` passes on the branch; `cargo test --test preflight_owns_the_fast_gates` passes (8/8). Every rewritten link I spot-checked resolves to a genuinely public target, and every downgrade-to-inline-code I checked (`Stepper`, `make_stepper`, `solve_ode_dense`, `resolve_method`, `pop_nll`, `compute_posterior_hessian`, `ode_predictions_adaptive`, `has_route_absorption_lag`, `m3_logcdf`, `cif_curves`, `MIN_PROB`, `rosenbrock`, `explicit_rk`) targets a `pub(crate)`/private item — so the policy was applied correctly. `markov = ["survival"]` and there is no `cfg(feature = "slow-tests")` under `src/`, so the single-feature-set claim in `group_rustdoc` holds.

### 1. `tools/preflight.sh:293` — high. The new gate exempts every non-public item, which is roughly half the tree and includes docs this PR just hand-fixed.

`cargo doc` without `--document-private-items` never rustdocs `pub(crate)`/private items, so their intra-doc links are not linted at all.

Reproduced: adding `/// Bogus link [\`totally_nonexistent_item_xyz\`].` to `pub(crate) fn resolve_method` in `src/ode/stiffness.rs` still gives `preflight OK` (exit 0) from `tools/preflight.sh rustdoc`. Running the same command with `--document-private-items` reports **75 remaining `error: unresolved link` lines** on this branch — including three more `MixtureClassGuard` links, i.e. the exact defect class this PR fixed on the public side.

That contradicts the stated rationale in `CLAUDE.md` and the CI job comment ("rust-analyzer hover the same way" / "visible defects today regardless of whether this crate is ever published") — rust-analyzer hover renders private items' docs too. Either add `--document-private-items` (and fix or baseline the 75), or say explicitly in the comment that the gate is public-surface-only, so the next reader doesn't assume a private-item link is protected.

### 2. `tests/preflight_owns_the_fast_gates.rs:244` — medium. The new comment states two things that are false, and hides a real coverage gap.

It says the loop covers "every group whose commands are `cargo` invocations" — but `docs` is excluded, and its commands (`cargo test -p docs-lint`, `cargo clippy -p docs-lint`) *are* cargo invocations, so `group_docs`' failure propagation is tested by nothing. It then says `public-api` "is covered by `the_failure_diagnostic_names_the_group_that_actually_failed`" — that test runs `preflight.sh fmt check` only and never touches `public-api`.

Concrete misbehaviour: if someone edits `group_docs` into a shape that swallows a failure (an `|| true`, a pipe, a subshell), no test goes red — and this comment tells the reviewer it is already covered. Adding `"docs"` to the loop costs nothing; its cargo commands are intercepted by the same fake `cargo`.

### 3. `tests/preflight_owns_the_fast_gates.rs:715` — medium. The new anti-`allow` tripwire only reads `src/lib.rs`, so two easy bypasses walk straight through it.

(a) The gate also runs `cargo doc -p ferx-tools -p ferx-cli`, but `crates/ferx-tools/src/lib.rs` and `crates/ferx-cli/src/main.rs` are never inspected — a crate-level `#![allow(rustdoc::private_intra_doc_links)]` there disables the lint for those crates permanently and the test stays green.

(b) An inner `#![allow(rustdoc::...)]` at the top of any module file (e.g. `src/sens/mod.rs`) silences the lint for that whole subtree — thousands of lines, far from anything it affects, which is precisely the harm the test's own doc comment describes — and is likewise invisible to a `src/lib.rs`-only read.

Scanning `src/**/*.rs` plus the member lib roots for `#![` + `allow` + `rustdoc::` closes both.

### 4. `tests/preflight_owns_the_fast_gates.rs:238` — low. Stale count in the doc comment.

"It runs through the same `run` helper as all eight positions below" — the loop now covers 10 positions (fmt 1 + check 5 + clippy 2 + rustdoc 2). The same paragraph's "`public-api` is the one group not exercised" is also wrong, for the reason in finding 2.

### Checked and cleared (not issues)

- `` [`GeneratorFn`] `` in `src/markov/endpoint.rs:8` and `src/parser/model_parser.rs:19327` resolve — `GeneratorFn` is imported (`use crate::types::{… GeneratorFn …}`), and `model_parser.rs` has `use crate::types::*`.
- `` [`solve_ode_g`] ``, `` [`DualMixed`] ``, `` [`Dual2`] ``, `` [`Jet`] `` after the path-stripping all resolve via each file's `use` items; module-doc (`//!`) links resolve against module scope regardless of `use` ordering.
- `` [`CompiledModel::ruv_obs_mult`] `` (was `ModelParameters::ruv_obs_mult`) is a real `pub fn` at `src/types.rs:4267` — the retarget is correct.
- `rustup toolchain install nightly --profile minimal` does ship `rustdoc` (it lives in the `rustc` component), so the new CI job needs no extra component.

Issue #14, opened 2026-05-16
feat: multi-endpoint support for simultaneous PK/PD fitting (multiple obs_cmt with separate error models)


TeunP:
## Summary

ferx currently supports a single `obs_cmt` per model file. This makes it
impossible to fit a combined PK/PD model where PK observations (plasma
concentration) and PD observations (e.g. effect, biomarker) need to be compared
to different compartments with different residual error models in the same
likelihood.

## Current limitation

```
[structural_model]
  ode(obs_cmt=central, states=[depot, central, effect_cmt])
```

Only `central` can be the observed compartment. PD observations at the
effect compartment cannot be included in the joint likelihood.

## Proposed feature

Support multiple observed compartments, each with its own error model,
referenced by `CMT` in the dataset:

```
[structural_model]
  ode(states=[depot, central, effect_cmt])   # no single obs_cmt

[error_model]
  CMT=2: DV ~ proportional(PROP_ERR_PK)     # central compartment
  CMT=3: DV ~ additive(ADD_ERR_PD)          # effect compartment
```

Or alternatively, a named endpoint syntax:

```
[structural_model]
  ode(states=[depot, central, effect_cmt],
      endpoints=[(cmt=central, cmt_id=2), (cmt=effect_cmt, cmt_id=3)])
```

## Use case: Emax PK/PD with effect compartment

The canonical simultaneous PK/PD model:

```
# PK
d/dt(depot)      = -KA * depot
d/dt(central)    = KA * depot - CL/V * central

# PD (biophase)
d/dt(effect_cmt) = KE0 * (central/V - effect_cmt)

# PD response (Emax)
EFFECT = E0 + EMAX * effect_cmt^GAMMA / (EC50^GAMMA + effect_cmt^GAMMA)
```

With PK observations (`CMT=2`) driving a proportional error and PD observations
(`CMT=3`) driving an additive error, a single joint likelihood is the gold
standard for simultaneous PKPD modelling.

## Current workaround (sequential approach)

Users must fit PK first, export individual IPREDs, and use them as time-varying
covariates in a separate PD-only model. This underestimates uncertainty
(PK parameter uncertainty not propagated) and complicates the workflow.

```r
fit_pk <- ferx_fit("pk.ferx", "pk_data.csv")
ipred  <- fit_pk$sdtab[, c("ID", "TIME", "IPRED")]
# ... merge IPRED into PD dataset, fit pd.ferx separately
```

## Implementation notes

[...]

PR #1165, opened 2026-08-31
feat(docs): structural linter for the docs, at PR-time [closes #1163]

### LINKED ISSUE #1163: docs: linter for section size, heading structure, duplicate anchors and dead internal links
## Why

The docs are written for humans reading a page top-to-bottom, but they are increasingly consumed a *section at a time* — by the site's own anchor links, and by agents that retrieve one addressable chunk rather than a whole page. Two things break under that use:

- **Oversized terminal sections.** A section with no subsections is the terminal payload: it is retrieved whole or not at all. `model-file/fit-options.qmd#general-options` is 28 kB of unbroken table — nothing can retrieve a piece of it.
- **Broken addresses.** 45 internal links on `main` point at anchors that do not exist in the rendered site. They are dead jumps for readers today.

Neither is caught by anything: `quarto render` happily emits a page with duplicate ids and dead anchors, and CI never runs it on a PR anyway (`docs.yml` is push-to-`main` only).

Proposal: a dependency-free docs linter, run on PRs, enforcing four rules.

### R1 — a section with no subsections stays under 5,000 characters

Characters, not tokens, so the linter needs no tokenizer, no network and no model. ~5 kB ≈ 2,000 tokens at this corpus's ~2.4 chars/token.

**The fix is always subdivision, never deletion.** A long section split into subheadings costs nothing — each subheading becomes its own address and nothing has to get shorter.

Scope is every independently-addressable chunk:

- a terminal section (no child heading), from its heading to the next heading at the same or shallower level;
- the prose above a page's first heading;
- the prose a parent section keeps above its own first subheading.

Counted on the `.qmd` source of the chunk, including code fences and tables (a 28 kB table is exactly the problem).

Measured on `main` (84 pages, 791 terminal sections) with a prototype:

| file:line | chars | section |
|---|---:|---|
| `model-file/fit-options.qmd:17` | 28,353 | General Options |
| `estimation/foce.qmd:37` | 22,536 | Gradient route (analytic vs FD) |
| `model-file/ode-models.qmd:275` | 9,448 | Letting ferx pick the stepper (`ode_method = auto`) |
| `file-formats/check-report.qmd:55` | 8,695 | Error / warning codes |
| `model-file/adaptive-dosing.qmd:382` | 7,642 | Current scope and limits |
| `warnings.qmd:40` | 7,391 | Warning codes |
| `model-file/error-model.qmd:625` | 7,367 | Examples |
| `estimation/saem.qmd:202` | 5,802 | What SAEM does about it now |
| `model-file/ode-models.qmd:476` | 5,756 | Which regime am I in? |
| `model-file/scaling.qmd:472` | 5,746 | Interaction with gradients |
| `estimation/foce.qmd:179` | 5,271 | How the Hessian is built |
| `estimation/tte.qmd:282` | 5,239 | Competing risks (cause-specific hazards) |
| `model-file/fit-options.qmd:265` | 5,127 | VI-Specific Options |

13 of 791 terminal sections over the limit (98.4% already compliant). Page preambles: zero over. Parent-own-prose: one — `model-file/error-model.qmd:480` "Covariate-selected error models (`if/else`)", 6,006.

### R2 — never skip a heading level

`#` → `###` with no `##` between makes the tree structurally wrong: the H3 attaches to the wrong parent, so a reader (or an extractor) is told a section lives somewhere it doesn't. H4 is as deep as we go.

One violation on `main`: `estimation/importance-sampling.qmd:27` — `#` → `###` at "Mapping to NONMEM".

### R3 — no two headings on a page may generate the same id

Addresses come from heading text, so a repeat produces `#syntax`, `#syntax-1`, `#syntax-2`, `#syntax-3` — four addresses nothing can choose between, and the `-N` suffix depends on document order, so it moves when a section is inserted above.

The check is on the **generated id**, not the visible text: repeating a heading under a different parent is fine as long as the ids differ. Fix by making titles distinct ("Syntax — Weibull", which reads better anyway) or with an explicit `{#id}`.

11 colliding addresses on 3 pages, confirmed against the rendered `_site` HTML:

- `model-file/absorption.qmd` (8): `syntax` ×4 (L42, 446, 581, 658), `worked-example` ×3 (L412, 496, 634), `shared-behaviour` ×3 (L479, 610, 694), `worked-examples` ×2 (L765, 940)
- `output.qmd` (2): `columns` ×2 (L34, 85), `example` ×2 (L69, 94)
- `model-file/parameters.qmd` (1): `declaration-order` ×2 (L366, 448)

### R4 — internal links must resolve, including the anchor

45 links on `main` whose file target exists but whose `#anchor` does not. Four classes:

1. **Numbered headings (29).** `## 3. Communication` generates `#communication` — pandoc strips everything up to the first letter. Every `sdlc.qmd#N-...` link in `development/sdlc.qmd` and `development/contributing.qmd` is dead.
2. **Hand-written double hyphen (6).** Links write `#analytic-foce--focei-...`, `#...-npde--npd`, `#...-nonmem--nlmixr2`, `#dose-routing--the-dose-feeds-...`; the generated ids have a *single* hyphen, because pandoc collapses whitespace runs. (`estimation/optimizers.qmd:154,164`, `model-file/fit-options.qmd:47,49`, `model-file/individual-parameters.qmd:310`, `model-file/absorption.qmd:485`, `model-file/simulation.qmd:86`.)
3. **Truncated links (4).** `model-file/parameters.qmd:63,134,135,136` — target literally `...`.
4. **Anchors that simply don't exist (6).** `#chained-fits` and `../estimation/foce.qmd#gradient-accuracy-vs-cost` (`fit-options.qmd:21,55`), `sir.qmd#caller-supplied-vs-re-read-inputs` (`foce.qmd:302`), `../estimation/saem.qmd#conditional-distribution-conditional-mode-vs-distribution` (`fit-options.qmd:260`), `iov.qmd#sample-size-weighted-iov-weight-` (`error-model.qmd:357`).

Dead **file** targets other than the four `...`: none. Missing linked assets: none.

**The id algorithm has to be pandoc's, exactly**, or the rule generates false positives — this was verified by diffing a prototype against `docs/_site/**/*.html` (`data-anchor-id`):

1. explicit `{#id}` on the heading wins;
2. otherwise: strip inline markup (backticks, emphasis, link syntax), **delete** every character that is not alphanumeric / `_` / `-` / `.` (delete, not replace — `` `D{n}` `` → `dn`), map whitespace runs to a single `-`, lowercase, strip everything before the first letter;
3. a repeat of an existing id gets `-1`, `-2`, … in document order.

Note the em-dash: `Branching model — trunk-based` → `branching-model-trunk-based`, one hyphen, because the dash is deleted and the surrounding spaces collapse. Any rule that assumes `--` is wrong (see R4 class 2).

## Implementation

`crates/docs-lint` — a **Rust** workspace member: std only, zero dependencies, and it does **not** depend on `ferx-core`.

Rust over a Python script in `tools/`, deliberately:

- The id algorithm is the part that has to be exactly right (a wrong slug model is what produced the incorrect anchor diagnosis this issue corrects). It needs real unit tests, and the repo's test infrastructure is `cargo test` — there is no pytest, no requirements file, no Python lint in CI. The two `python3` calls in `ci.yml` are inline float comparisons, not a codebase.
- `rustfmt` (pre-commit hook), `clippy` and `cargo test` apply for free; a Python script is gated by nothing.
- The CI compile cost that matters here (#969: ~93% LLVM, `ferx-core` monomorphization) does not apply — this crate links nothing and builds in seconds.
- No workspace side effects: CI runs the members as explicit `-p ferx-tools -p ferx-cli`, never `--workspace`, and the coverage jobs target `ferx-core`, so a third member joins nothing by accident. `tests/public_api_boundary.rs` is unaffected — it pins that `ferx-core` does not depend on members, not the reverse.

```
crates/docs-lint/
  src/main.rs        # bin `docs-lint`
  src/slug.rs        # pandoc id algorithm + unit tests
  src/rules.rs       # R1-R4
  tests/corpus.rs    # #[test]: the docs/ tree is clean
```

```
cargo run -p docs-lint            # report all violations, exit 1 if any
cargo run -p docs-lint -- --check # what CI runs
cargo run -p docs-lint -- --json  # machine-readable
cargo test -p docs-lint           # unit tests + corpus check
```

Output is `path:line: RULE: message`, one line per violation, so it is greppable and clickable.

Details:

- Input: `docs/**/*.qmd`, resolved relative to `CARGO_MANIFEST_DIR/../..` so it runs from anywhere. Skips YAML front matter, fenced code blocks (a `#` inside a fence is not a heading), and `docs/_site/`.
- Rule 1 limit lives in one constant; `--max-chars` overrides it for experimentation.
- Per-rule disable for a genuinely unsplittable chunk: `<!-- lint-disable R1 -->` on the line above the heading. Expected to be rare; a disable needs a reason in the comment.
- Boundary note: this is dev tooling, not engine code. It is a workspace member for the tests and the gates, not part of the `ferx-core` / `ferx-tools` / `ferx-cli` layering, and nothing may depend on it.

**Cost of the Rust choice, stated plainly.** A cold lint job is ~30-60 s (toolchain install + tiny compile) against ~3 s for a script, and a docs-only contributor needs the Rust toolchain to run it locally — which in this repo they already have.

**CI.** New `docs-lint` job in `ci.yml` running `cargo test -p docs-lint` — unit tests and the corpus check in one step. It must be on `ci.yml` (PRs), not `docs.yml`, which is push-to-`main` and so too late to block anything. Optionally also a step in `docs.yml` before `quarto render`.

**Baseline.** `main` has 14 R1 violations and 45 R4 violations today, so the linter cannot go straight to hard-fail on everything. Plan: R2, R3, R4 are fixed in the same PR as the linter (mechanical — renamed headings, corrected anchors), so those three land as hard failures with a clean tree. R1 lands with a checked-in `docs/.lint-baseline` listing the 14 known-oversized chunks; the file is **shrink-only** — a new violation or a grown chunk fails CI, removing an entry is the only allowed edit. Burn it down page by page in follow-ups.

## Scope / non-goals

- No prose-style checking (no vale, no word lists, no readability scores).
- No external link checking (network in CI).
- No rendering, and no Quarto/pandoc dependency — the linter reads `.qmd` source only. That is why the id algorithm is pinned to pandoc's above and unit-tested against the rendered `_site` ids.

## Tasks

- [ ] `crates/docs-lint` (std-only workspace member) with R1–R4 + `--check`/`--json`/`--max-chars`, disable comments
- [ ] Unit tests for the id algorithm (numbered headings, em-dash, `` `D{n}` ``, explicit `{#id}`, `-1` suffixing) asserted against the ids Quarto actually renders
- [ ] Fix the R2 violation (`importance-sampling.qmd:27`)
- [ ] Fix the 11 R3 collisions (absorption, output, parameters)
- [ ] Fix the 45 R4 broken anchors
- [ ] `docs/.lint-baseline` with the 14 R1 chunks, shrink-only enforcement
- [ ] `docs-lint` job in `ci.yml` running `cargo test -p docs-lint`
- [ ] Document the rules in `docs/development/` (or `CONTRIBUTING`) so the fix is discoverable from the failure message
- [ ] Follow-up issues to subdivide the baselined sections (`fit-options#general-options` and `foce#gradient-route` first — 28 kB and 22 kB)

Counts above come from a prototype run against `main` at 724c2b8b and cross-checked against the committed `docs/_site` render.

roninsightrx:
## Why

The docs are written to be read top-to-bottom but are *addressed* a section at a time — by the site's own anchor links, and by anything that retrieves one chunk rather than a whole page. Nothing checked that this worked: `quarto render` emits duplicate ids and dead anchors without complaint, and `docs.yml` only runs on a push to `main`, which is too late to block a bad page.

Measured on `main` (84 pages, 791 terminal sections): **41 internal links whose anchor does not exist**, **11 colliding anchors** on three pages, **one skipped heading level**, and **14 sections over 5,000 characters** — one of them a 28 kB unbroken table.

Closes #1163.

## What changed

`crates/docs-lint`, a std-only workspace member (zero dependencies, does not depend on `ferx-core`), enforcing four rules over `docs/**/*.qmd`:

| Rule | Check | On `main` |
|------|-------|-----------|
| R1 | a section with no subsections stays under 5,000 characters | 14 violations → baselined |
| R2 | never skip a heading level | 1 → fixed |
| R3 | no two headings on a page may generate the same id | 11 → fixed |
| R4 | internal links must resolve, including the anchor | 41 → fixed |

R2, R3 and R4 land clean, so they are hard failures from this commit.

**R3 changes some published anchors.** Four sections called "Syntax" on the absorption page were addressed as `#syntax`, `#syntax-1`, `#syntax-2`, `#syntax-3` — order-dependent addresses nothing can choose between. They are now `#syntax-transit`, `#syntax-inverse-gaussian`, `#syntax-weibull`, `#syntax-zero-order`; same for the repeated "Worked example" / "Shared behaviour" sections, `output.qmd`'s two "Columns"/"Example" pairs, and `parameters.qmd`'s two "Declaration order" sections. No link inside `docs/` referenced the old suffixed forms (checked), but an external bookmark to one would break.

**R4, by cause:**

- **29** numbered headings — `## 3. Communication` is addressed as `#communication`, because pandoc drops everything before the first letter. Every `sdlc.qmd#N-…` link was dead.
- **6** hand-written with a double hyphen (`#analytic-foce--focei-…`) where the generated id has one: a deleted `/`, `—` or `(` leaves two spaces, which pandoc collapses to a single separator.
- **6** pointing at sections that never existed (`#chained-fits`, `#gradient-accuracy-vs-cost`), retargeted at the section that does.
- Two more of the same class in `CHANGELOG.md` and `SECURITY.md` (`sdlc.html#9-releases`, `#14-security--dependencies`) fixed by hand; the linter only scans `docs/`.

**R1 lands with `docs/.lint-baseline`** — 14 entries, shrink-only: a new violation fails, a baselined section that grew fails, and a section that no longer violates fails as a *stale* entry, so fixing one forces its line out of the file rather than leaving a permanent exemption. Burn-down is follow-up work (`fit-options#general-options` at 28 kB and `foce#gradient-route` at 22 kB first).

The fix for R1 is always subdivision, never deletion: a section split under subheadings costs nothing and no word has to go.

## Alternatives considered

**A Python script in `tools/`** was the original plan in #1163 — rejected after review. The id algorithm is the part that has to be exactly right, and the repo's only test infrastructure is `cargo test`: no pytest, no requirements file, no Python lint in CI (the two `python3` calls in `ci.yml` are inline float comparisons). Rust also gets rustfmt/clippy/test for free. The compile-cost objection does not apply: the crate links nothing, so `cargo test -p docs-lint` is seconds, and the cost that matters (#969) is `ferx-core` monomorphization.

**Rendering with Quarto and checking the HTML** would need Quarto in CI and a full render per PR. The linter reads source instead, which is why the id algorithm is pinned to pandoc's and tested against a real render.

## Cross-repo dependency

No `ferx-core` API change; `docs-lint` is a separate package nothing depends on.

## Breaking changes
- [x] None

Nothing in `ferx-core` changed, so the public-API baseline is untouched. Some published documentation *anchors* changed (see R3 above).

## Numerical validation
- [x] Not applicable (docs / tooling / CI only)

## Performance

New CI job, `Docs lint`: a zero-dependency crate compile plus 53 tests (11s on the first green run). It never links `ferx-core`, so it does not touch the compile path that dominates CI time.

## Tests
- [x] Unit test(s) added in the same module (`cargo test -p docs-lint` passes: 41 unit + 10 CLI + 2 corpus, 1 opt-in ignored)

Coverage of the traps that produced the wrong first diagnosis of this bug class:

- `slug::tests::matches_rendered_quarto_ids` — 11 headings whose expected id was copied out of a Quarto render, not guessed: numbered headings, em-dash, `` `D{n}` `` → `dn`, `vs.` keeping its dot, `ode_method = auto`.
- `rules::tests` — one test per rule per direction, including `em_dash_anchor_takes_one_hyphen_not_two` and `numbered_heading_anchor_is_dead`.
- `baseline::tests` — shrink allowed, growth fails, stale entry fails, another rule never excused.
- `tests/cli.rs` — the binary's contract, above all its exit codes (0 clean / 1 violations / 2 usage error), against throwaway docs trees: `--help`, `--check`, `--json` incl. quote escaping, `--max-chars`, the three baseline outcomes, a malformed baseline, an unknown option, a missing option value, a missing docs directory.
- `tests/corpus.rs` — `docs/` passes all four rules; every baseline entry names a live section.
- `computed_ids_match_the_rendered_site` — opt-in (`DOCS_LINT_VERIFY_RENDER=1`, `#[ignore]`d) cross-check of every computed id against `docs/_site`; skips pages edited since the render. Run against the committed render: 0 mismatches.

docs-lint line coverage: 97.2%.

## Docs
- [x] `docs/` updated for user-visible changes — new `docs/development/docs-lint.qmd` (the four rules, the id algorithm and its three traps, how to silence a rule)
- [ ] No user-visible change

`sdlc.qmd` §8.5 (documentation workflow) now points at the linter page.

## Checklist
- [x] `cargo clippy` clean (`-p docs-lint --all-targets -- -D warnings`)
- [x] `cargo fmt --all -- --check` clean
- [ ] Functions on the `Dual2` sensitivity path — not touched

## Example execution
- [x] No example execution step needed (docs / tooling only)

## Follow-up commits (after the first CI run)

Three things the first run surfaced, each fixed rather than worked around:

1. **`Docs lint` failed on the merge commit** — `foce.qmd`'s "Gradient route" section grew 22,536 → 24,331 characters on `main` while this branch was open. That is the baseline doing its job (a baselined section that grows fails), so the entry was re-recorded at its current size. `main` was merged in.
2. **`main` landed `tools/preflight.sh` (#1157)**, whose contract is that CI runs the same list a developer runs. `docs-lint` now has a `docs` group there and the CI job calls `tools/preflight.sh docs`; `tests/preflight_owns_the_fast_gates.rs` gets its command count for the new group — and that guard caught the omission itself ("group `docs` runs but has no expected command count in this test — add one, or it is gated by nothing").
3. **The #1114 invocation guard false-positived on the new crate.** `checked_in_ferx_invocations_select_the_ferx_cli_package` excused a line only if it named `ferx-cli` or `ferx-core`, so `cargo run -p docs-lint -- --check` read as "selects no package". The rule now checks the *form* — any `-p <pkg>` / `--package <pkg>` — so the next workspace member is not a false positive on its first commit either. Both excused forms are pinned in the classifier unit test.

A fourth: `codecov/patch` came in at 83.52% against the 90% floor, all of it `main.rs` at 0% — arg parsing, the writers, and the exit codes, none of which the corpus test reaches. `tests/cli.rs` covers them; 83.2% → 97.2%.

## Reviewer hints

Start with `crates/docs-lint/src/slug.rs`. Everything else is bookkeeping; that file is where a mistake turns the whole check into noise, and it is worth confirming the three traps in its test table against your own reading of pandoc.

Second, `docs/.lint-baseline` — check that the *shrink-only* semantics are what you want as the burn-down mechanism, in particular that a fixed section fails until its line is removed.

The 41 link fixes were applied as verbatim string replacements from one script; the diff is mechanical and can be skimmed. The R3 heading renames are the judgement calls — the new titles are in `docs/model-file/absorption.qmd`, `docs/output.qmd`, `docs/model-file/parameters.qmd`.

## Open questions

- The em-dash convention in the new titles (`### Syntax — Weibull`) matches the page's existing style, but `### Syntax (Weibull)` would read as well. Happy to switch.
- `iov.qmd`'s heading `## Sample-size-weighted IOV (\`weight = ...\`)` generates the id `#sample-size-weighted-iov-weight-...`, which is valid but ugly, and the link was corrected to match rather than the heading given an explicit `{#…}`. Worth a follow-up.
- Local `quarto render docs` produced no output in my environment (silent no-op, exit 0), so the render cross-check was run against the committed `_site` — 23 fresh pages, 0 mismatches — plus the 11 hand-checked cases. If someone with a working Quarto runs `DOCS_LINT_VERIFY_RENDER=1 cargo test -p docs-lint -- --ignored` on a fresh render, that closes the loop over all 84 pages.

https://claude.ai/code/session_017oyMQJPLy8TX3MNtrc5evS


roninsightrx:

Reviewed the diff against `origin/main`. `cargo test -p docs-lint` is green (13 tests), the corpus and baseline-liveness checks pass, the CLI exit-code contract behaves as pinned, and the anchor rewrites across `docs/**` are consistent with the implemented pandoc algorithm. No dangling references to the renamed headings elsewhere in the repo.

Nine findings below, most severe first.

### 1. `crates/docs-lint/src/rules.rs:247` — same-page anchors silently stop being checked (MEDIUM)

The same-page anchor lookup keys on the raw `doc.path`, but `lint()` builds `ids_by_file` with `normalize(&d.path)`. Any docs directory given with a `./` or `..` component makes every same-page `#anchor` check vanish without a word.

Reproduced:

```
docs-lint /tmp/dl/docs --no-baseline          # reports the dead #nowhere link, exit 1
cd /tmp/dl && docs-lint ./docs --no-baseline  # "docs-lint: clean", exit 0
```

Fix: `normalize(&doc.path)` at the lookup (and in the `check_anchors` test helper).

### 2. `crates/docs-lint/src/parse.rs:52` — nested code fences flip fence parity (MEDIUM)

`is_code_fence` toggles on any line starting with ``` or `~~~`, so a four-backtick block containing a three-backtick block — the standard way to show markdown inside docs — flips parity and the fenced content is parsed as live markdown.

Reproduced: a page with `## S`, then a ````markdown block containing `## S`, reports a false `R3` duplicate-id violation. The same mechanism can hide real headings and shift pandoc's `-1`/`-2` suffixes, which corrupts R4.

Fix: track the opening fence's marker and length, and only close on a matching-or-longer run of the same marker.

### 3. `crates/docs-lint/src/parse.rs:189` — indented code blocks parsed as headings (LOW/MEDIUM)

Indented (4-space) code blocks are not recognized, and the heading test uses `line.trim_start()`, so a `#` comment inside one becomes a heading.

Reproduced: a page containing `    ##### deep comment in code` inside an indented block yields a false `R2: heading level jumps from h1 to h5`.

Fix: require the `#` at column 0, or track indented blocks.

### 4. `crates/docs-lint/src/slug.rs:120` — letterless headings get an empty id (LOW)

A heading with no letters yields `""`, but pandoc substitutes `section` for an empty identifier. A `#section` link to such a heading is reported dead, and two letterless headings collide on the empty base id, producing a nonsense R3 message.

### 5. `crates/docs-lint/src/slug.rs:84` — `strip_inline_markup` is incomplete (LOW, latent)

It drops `` ` `` and `*` but not `_` emphasis (nor `~`, `$…$`, raw HTML). `## _Important_ notes` slugs to `_important_-notes` where pandoc renders `important-notes`, so a correct link is flagged dead and the broken one passes. No such heading exists today.

### 6. `crates/docs-lint/src/slug.rs:140` — explicit `{#id}` runs through the collision counter (LOW)

Pandoc emits an explicit id verbatim. Here an explicit `{#foo}` after an auto `foo` is recorded as `foo-1`, so `#foo-1` resolves in the linter but not on the site, and the heading is reported as R3 where pandoc would simply duplicate the id.

### 7. `tests/public_api_boundary.rs:231` — the #1114 guard now excuses non-cargo `-p` (LOW)

`selects_a_package` does a bare `line.contains("-p ")` over the whole line, including everything after `--`. A command passing `-p` to the *program* — `cargo run --release -- model.ferx -p 8` — is excused although it never selects a package, which is exactly the regression the guard exists to catch. Match only the portion before the `--` separator, or require `-p <word>` in the cargo-argument prefix.

### 8. `crates/docs-lint/Cargo.toml:2` — the comment overstates the gating (LOW)

It says the crate is a member "so it gets rustfmt / clippy / `cargo test`", but neither `tools/preflight.sh clippy` (which runs `cargo clippy` on ferx-core, then `-p ferx-tools -p ferx-cli`) nor the CI Clippy job includes `docs-lint`. Only `cargo fmt --all` and the new `docs` group cover it, so clippy findings in this crate will never gate. The crate is clippy-clean today.

### 9. `docs/development/docs-lint.qmd:84` — off-by-one in prose (LOW)

"Three consequences catch people out — all three were live bugs" precedes a four-row table (the trailing-`.`-is-kept row).

PR #1168, opened 2026-08-31
docs: fix the rustdoc link warnings and gate them in tools/preflight.sh

TeunP:
## Why

`cargo doc` emitted **146 link warnings** on `main`, and nothing gated them — `cargo doc` exits 0
on warnings, so no job ever looked. They had accumulated freely.

An unresolved intra-doc link is not a silent no-op. Measured on a probe crate:

```
[`PrivateThing`]  →  <p>Link to private item: [<code>PrivateThing</code>].</p>
`PrivateThing`    →  <p>Plain code text: <code>PrivateThing</code>.</p>
```

The brackets survive into the rendered output, and the same is true in rust-analyzer hover. So
these are visible defects in the editor today, independent of whether this crate is ever
published.

## What changed

Three commits, deliberately separable.

**1 — fix the links that are defects in every doc mode** (51 lines, 22 files, all `///` / `//!`)

| class | n | treatment |
|---|---|---|
| unresolved link | 45 | target does not exist anywhere → inline code |
| redundant explicit target | 10 | `[`X`](path)` where the label already resolves → `[`X`]` |

**2 — park the 91 private-item links behind an `allow`** (one line + rationale)

These link to `pub(crate)` internals (`pop_nll`, `eta_dx`, `build_fit_pool`). They are not
defects: under `--document-private-items`, and in rust-analyzer hover which indexes private items,
they resolve and are navigable. In a public build the target is not documented at all.

Only the first consumer exists today — `ferx-core` is not on crates.io, `docs.rs/ferx-core` is
404, no workflow runs `cargo doc`, and the Quarto site links no rustdoc. So the links are doing
their job, and rewriting them would trade a live benefit for a hypothetical one. **Publishing
reverses that trade**, which is why it is one line to delete rather than 91 edits to redo.

**3 — the gate**, a `docs` group in `tools/preflight.sh` plus a `Docs` CI job that runs it.

## Alternatives considered

**Repairing the broken links instead of downgrading them to code.** Built a classifier matching
target names against the public-API baseline. Rejected — it was confidently wrong: it "repaired"
`[output]` in `types.rs` to `crate::io::output` when the prose means the `.ferx` `[output]` block,
and matched `[T]` in "Mean survival time E[T]" to the generic parameter of `SolPointG<T>`. A link
to the wrong item is worse than no link.

**Cleaning the 91 rather than allowing them.** This was the original approach and the whole commit
was written that way. It was reverted after measuring that the justification — keeping rendered
docs clean for downstream readers — rests on a rendered doc that does not exist. See commit 2.

**`RUSTDOCFLAGS=... run cargo doc ...` as an assignment prefix.** Reaches rustdoc correctly (bash
exports it into the function and on to the child — verified), but `run` echoes `$*` and a prefix
is not part of `$*`, so `--list` would advertise a bare `cargo doc` that does **not** fail on
warnings. Used `run env RUSTDOCFLAGS=-Dwarnings cargo doc ...` so what is printed is what runs.
There is a test for this.

**A `cargo doc` matrix mirroring `check`'s five feature sets.** Unnecessary here: `markov` implies
`survival`, and `slow-tests` gates only `#[cfg_attr(...)]` test attributes, so `ci,markov,nn` is
the union of the production cfgs. The non-nesting trap that makes `check` a matrix (#1133) does
not exist for rustdoc. Clippy takes the same single set for the same reason.

## Cross-repo dependency

No public API change; `cargo public-api` baseline unchanged.

## Breaking changes
- [x] None

## Numerical validation

## Performance

The new gate adds **~8s cold on a warm `target/`** (ferx-core 6.2s + members 1.8s) and **0.31s**
when nothing changed. Measured with no competing cargo — see the note in the script header about
`target/` lock contention inflating these numbers several-fold.

## Tests

`tests/preflight_owns_the_fast_gates.rs` extended, not added to — the existing count table has a
cross-check that **fails on an ungated new group by design**, so adding `docs` to `ALL_GROUPS`
turned it red until the table grew. That is the tripwire working.

New assertions:
- `RUSTDOCFLAGS=-Dwarnings` present on every docs command (the entire gate — without it `cargo doc`
  exits 0 on 146 broken links while still taking 8s and still appearing in `--list`)
- the command starts with `env `, so `--list` cannot advertise different behaviour than it runs
- a docs command covers `-p ferx-tools`, since a root `cargo doc` documents only the root package
  (#1114)
- `docs` added to `a_failing_gate_fails_the_script_from_every_position` and to the CI-job
  delegation list

**Mutation-tested, 8 cases — 6 killed by a named test, 2 negative controls stayed green:**

| mutation | outcome |
|---|---|
| drop `docs` from `ALL_GROUPS` | killed by `preflight_is_executable_and_lists_every_group` |
| drop the members docs line | killed by `preflight_is_executable_and_lists_every_group` |
| drop `-Dwarnings` | killed by `load_bearing_flags_and_feature_coverage_survive_in_the_command_list` |
| `RUSTDOCFLAGS=` as prefix instead of `env` | killed by `load_bearing_flags_…` |
| remove the `Docs` job from `ci.yml` | killed by `the_fast_gate_jobs_delegate_to_preflight_…` |
| CI inlines `cargo doc` instead of calling preflight | killed by `the_fast_gate_jobs_delegate_to_preflight_…` |
| NEGATIVE: reword a comment | stayed green |
| NEGATIVE: reorder two flags | stayed green |

**End-to-end, on the real tree** — injected a broken doc link and ran the gate:

```
error: could not document `ferx-core`
  = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`

preflight FAILED in group 'docs'.
  command:  env RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps --no-default-features --features ci,markov,nn
  CI job:   Docs — this is what would have gone red on the PR.
```

That run also confirms the `allow` in commit 2 is narrow: a genuinely broken link still fails,
because `broken_intra_doc_links` is a separate lint.

**The split between commits 1 and 2 is pinned by measurement, not by inspection.** With the
`allow`: 0 warnings. With the one line deleted: exactly **91**, every one a private-item link,
rustdoc's own summary agreeing (`generated 91 warnings`). No leakage in either direction.

## Docs
- [x] No user-visible change

`CHANGELOG.md`: no entry — internal / CI, per CLAUDE.md.

## Checklist
- [x] `tools/preflight.sh` green — now including the new `docs` group
- [x] Functions on the `Dual2` sensitivity path stay differentiable — every changed line in commits
      1 and 2 is mechanically confirmed to be a `///` or `//!` comment, so no code moved

### ferx-site
- [x] No new DSL syntax, fit option, example, or data file

### ferx-book
- [x] No estimator / DSL / fit-option change

### Example execution
- [x] No example execution step needed (docs-only / CI, no user-visible change)

## Reviewer hints

**Skim** commits 1's diff — 51 near-mechanical doc-comment lines. The classifier rationale is in
the commit message if you want to know why a link became code rather than a repaired path.

**Focus** on two things:

1. **Commit 2's judgment call.** It is one line and the argument for it is entirely
   "no rendered rustdoc exists today". If we publish, it should be deleted and the 91 rewritten.
   That is decision 7 in #1170 rather than settled here.
2. **`group_docs` in `tools/preflight.sh`** — specifically the `env` form and the single feature
   set. Both have comments explaining why; both have tests.

## Open questions

**No `closes`.** #1170 tracks the publishing decision that owns commit 2's fate, but this PR does not close it. Nothing existed to close — `gh issue list --search` found nothing for rustdoc
or doc warnings. Happy to file one retroactively if you'd rather have the trail.

**The `allow` silences future private-item links too**, not just today's 91. I think that is
right — internal cross-references are intentional, and the class that matters
(`broken_intra_doc_links`) stays gated — but it is a real widening and worth a second opinion.


roninsightrx (changes requested):
Two findings need changes before merge:

1. **[P1] Do not suppress `private_intra_doc_links` crate-wide.** The new `#![allow(rustdoc::private_intra_doc_links)]` makes the Docs job green while the public HTML it generates still contains the exact malformed bracket rendering this PR is intended to eliminate. On this head, after `tools/preflight.sh docs`, `rg -o "\\[<code>[^<]+</code>\\]" target/doc/ferx_core --glob "*.html" | wc -l` reports **90** occurrences. Examples include `types/enum.DoseAttrConsumption.html` and `api/fn.configure_global_thread_pool.html`. A local `cargo doc` build is already a real public-doc consumer even if the artifacts are not deployed. Please either rewrite these private-target links as code or otherwise make the public build clean without a crate-wide exemption; otherwise future public-to-private broken links are permanently outside the gate too.

2. **[P2] Preserve links whose targets are public API.** The first commit says the downgraded targets do not exist anywhere, but several are present in `api/ferx-core-public-api.txt`. For example, `read_nonmem_csv` in `src/api/predict.rs:41` and `src/api/simulate.rs:41` is re-exported as `ferx_core::read_nonmem_csv`; `combine_subject` in `src/estimation/importance_sampling.rs:497` is public as `ferx_core::estimation::mixture::combine_subject`; and `run_covariance` / `FitOptions::scale_params` are public too. These links were unresolved only because the short name was out of scope. Replacing them with inline code discards useful navigation instead of repairing it. Please use explicit resolvable targets such as `[`crate::read_nonmem_csv`]` and `[`crate::estimation::mixture::combine_subject`]` for the public symbols, reserving inline code for genuinely nonexistent or intentionally private targets.

Focused verification passed: `cargo test --test preflight_owns_the_fast_gates` (7/7) and `tools/preflight.sh docs`.

roninsightrx:
Both findings addressed, plus a merge conflict with `main` that landed in between.

## P1 — the crate-wide `allow` is gone

`#![allow(rustdoc::private_intra_doc_links)]` is deleted and all 91 private-item links are rewritten at the link: each `` [`target`] `` whose target is `pub(crate)` becomes inline `` `target` ``. Your measurement is the acceptance test — after `tools/preflight.sh rustdoc`:

```
$ grep -roh "\[<code>[^<]*</code>\]" target/doc/ferx_core --include='*.html' | wc -l
0
```

(was 90) and `cargo doc` reports 0 warnings with no exemption anywhere in the crate.

You are right that the "no rendered doc exists" argument was wrong on its own terms — the gate itself renders one on every PR — and right that the widening was the bigger problem: the `allow` was not scoped to the 91, it removed the whole class from the gate permanently.

**New test: `no_crate_level_allow_switches_a_rustdoc_lint_off`.** None of the existing assertions can see this failure mode — a crate-level `allow` leaves every command in the list, every `-Dwarnings` in place, and the group green. Only reading `src/lib.rs` catches it. Per-item `#[allow]` is deliberately out of scope: it is visible in review at the place it applies.

## P2 — public targets are links again

You were right, and it was broader than the four you named. 16 of the downgrades had public targets. Each is restored with an explicitly resolvable path:

| site | now |
|---|---|
| `api/predict.rs`, `api/simulate.rs` | `` [`crate::read_nonmem_csv`] `` |
| `estimation/importance_sampling.rs` | `` [`crate::estimation::mixture::combine_subject`] `` |
| `markov/endpoint.rs` | `` [`crate::markov::ctmm_data_term_grad`] `` |
| `ode/predictions.rs` | `` [`crate::types::CompiledModel::sync_ode_solver_opts`] `` |
| `sens/two_cpt.rs` | `` [`one_cpt_zero_order_g`](crate::sens::one_cpt::one_cpt_zero_order_g) `` |
| `stats/residual_error.rs` ×2 | `` [`ErrorSpec::sigma_loadings`](…) `` |
| `categorical/mod.rs` | `` [`SimOutcome::Category`](…) `` |
| `types.rs` | `` [`crate::fit`] ``, `` [`crate::run_covariance`] ``, `` [`FitOptions::scale_params`] ``, `` [`OuterResult::packed_estimate`](…) ``, `` [`OdeSpec`](…) ``, `` [`OdeSpec::solver_opts`](…) ``, `` [`crate::parser::model_parser::ScaleDerivProgram`] ``, `` [`CompiledModel::ruv_obs_mult`] `` |

Inline code is now reserved for targets that are genuinely `pub(crate)` (`MixtureClassGuard`, `elbo::mc_kl_draw`, `propagate_two_cpt`, `ode_predictions_adaptive`, …) or do not exist (`cif_curves`, `tests::one_node_agq_equals_laplace`).

One target was also simply wrong, not just out of scope: `types.rs` named `ModelParameters::ruv_obs_mult`, but the method is on `CompiledModel`. Corrected while relinking.

Every restored path is verified by the gate rather than by inspection — an unresolvable one is `rustdoc::broken_intra_doc_links`, a private one is `rustdoc::private_intra_doc_links`, and both are denied. 0 warnings means all 16 resolve.

## The `docs` group name collided with `main`

#1165 merged while this was in review and added its own `docs` group to `tools/preflight.sh` — the structural linter over `docs/**/*.qmd`. Different gate, same name, direct conflict.

Both are kept. This PR's group is renamed **`rustdoc`** and its CI job **`Docs` → `Rustdoc`**, so the workflow does not list `Docs` next to `Docs lint` for unrelated things. `ALL_GROUPS` is `(fmt check clippy rustdoc docs public-api)`; the count table grows to six entries with both; the delegation, failing-position and `-Dwarnings` assertions follow the rename. CLAUDE.md spells out the distinction, since the wrong name is a silent no-op.

## Verification

- `cargo test --test preflight_owns_the_fast_gates` → 8/8
- `cargo doc` (ferx-core + members): 0 warnings, 0 bracket-rendered links in `target/doc`
- public-API baseline unchanged

PR #1169, opened 2026-08-31
feat(tools): GAM covariate pre-screening — gam_screen() [closes #1114]

### LINKED ISSUE #1114: refactor(workspace): split into ferx-core / ferx-tools / ferx-cli, with a CI-enforced public API boundary

JordanBrooks33:
## Why

Population PK analyses need to pre-screen covariates before committing to expensive stepwise covariate modeling (SCM) runs. Without this, practitioners run N × K NONMEM/ferx fits (all ETA × covariate combinations) to generate a ranking, when a fast OLS-based GAM screen can narrow the candidate list to 2–3 covariates first.

No equivalent existed in ferx-tools. This adds `ferx_tools::gam::gam_screen()` — the Rust equivalent of Xpose4's `xpose.gam()`.

## What changed

New `ferx_tools::gam` module:

- **`gam_screen(fit, pop, opts)`** — entry point. For each ETA × covariate pair, fits `η_i ~ f(cov_i)` and ranks by `ΔAIC = AIC_null − AIC_best`. Parallelized over ETAs via rayon.
- **Functional forms tried**: linear (`[1 | x]`), natural cubic splines (df=2, df=3 by default via ESL §5.2.1 knot placement), one-hot categorical (lowest level = reference).
- **AIC formula**: `n·ln(RSS/n) + 2p` — matches R's `AIC(lm())` ΔAIC to floating-point precision (constant terms cancel).
- **CovariateKind** sourced from `fit.covariate_table` (from `[covariates]` block in the model file); heuristic fallback for undeclared covariates.
- **Shrinkage warning** emitted when ETA shrinkage > 30% (configurable via `GamOptions`).
- **`GamOptions`**: filter ETAs/covariates, choose spline df list, toggle linear, set shrinkage threshold.

## Alternatives considered

- **`gam` package backfitting** (what Xpose4 actually uses): unnecessary overhead — since η ~ N(0,ω²), Gaussian OLS is the optimal estimator and avoids iterative backfitting.
- **`mgcv`-style penalized splines**: degrees of freedom selection by GCV adds complexity without benefit at the pre-screening stage where ranking is more important than precise smooth estimation.
- **Stepwise search** (`steppit=TRUE` in xpose.gam): each covariate is tested independently against the same null model, which is the correct pre-screening approach. Stepwise build-up is a natural follow-on but not needed here.

## Cross-repo dependency

## Breaking changes

- [x] None

## Numerical validation

Validated against R (`splines::ns()` + `lm.fit()`, same `n·log(RSS/n) + 2p` formula) on two datasets:

**two_cpt_oral_cov** (30 subjects, 10 ETA×cov pairs):
- Max |ΔAIC_Rust − ΔAIC_R| = **1.16e-05** (floating-point noise)
- Form agreement: 10/10, top-1 ranking: 5/5 ETAs

**Cefepime** (458 subjects, 18 ETA×cov pairs):
- Max |ΔAIC_Rust − ΔAIC_R| = **2.69e-05** (floating-point noise)
- Form agreement: 18/18, top-1 ranking: 3/3 ETAs

Key result: CRCL ranks #1 for ETA_CL (ΔAIC=394.85, Spline df=3) and CR ranks #2 (ΔAIC=245.93) in the cefepime base model — matching the covariate structure of run64.

```
# cefepime ETA_CL (base model, n=458)
covariate   delta_aic   best_form
CRCL          394.852   Spline(df=3)
CR            245.931   Spline(df=3)
AGE            49.855   Linear
HT             37.722   Linear
SEX            25.215   Categorical
WT             19.379   Linear
INFD            0.000   Categorical
```

## Performance

- [x] No significant impact expected (new crate, ferx-core unchanged)

## Tests

  - 11 unit tests in `crates/ferx-tools/src/gam.rs`: `ns_basis_df1_returns_x`, `ns_basis_df2_three_knots`, `ols_null_aic_formula`, `ols_singular_returns_none`, `high_shrinkage_emits_warning`, `low_shrinkage_no_warning`, `uncorrelated_cov_near_zero_delta_aic`, `strong_linear_signal_gives_large_delta_aic`, `quadratic_signal_spline_beats_linear`, `categorical_covariate_detects_group_effect`, `too_few_subjects_returns_empty`
- [x] Slow integration test added: `crates/ferx-tools/tests/gam_two_cpt_oral_cov.rs` (fits base two-cpt model, asserts WT/CRCL rank top-2 for ETA_CL)

- [x] Not applicable (internal API — no new `.ferx` DSL syntax or fit options)

Usage in Rust:

```rust
use ferx_tools::gam::{gam_screen, GamOptions};

let result = gam_screen(&fit_result, &population, &GamOptions::default());
for eta in &result.eta_results {
    println!("{}: top covariate = {}, ΔAIC = {:.2}",
        eta.eta_name,
        eta.covariate_scores[0].covariate,
        eta.covariate_scores[0].delta_aic);
}
```

## Docs

- [x] No user-visible `.ferx` DSL change

## Checklist

- [x] `cargo clippy` clean
- [x] Functions on the `Dual2` sensitivity path are untouched (new crate only)
- [x] `Cargo.toml` version not bumped (no breaking changes to ferx-core)

## Reviewer hints

- Core algorithm is in `screen_eta_raw()` (~100 lines). The `ns_basis()` function implements ESL §5.2.1 directly.
- `ols_aic()` returns `Option` — `None` when X'X is singular (degenerate/constant covariate). Callers skip those pairs.
- Per-covariate null AIC uses only subjects with non-NaN covariate values (same subset as the alternative model) — this is the correct ΔAIC denominator.
- The `pub(crate)` on internals is deliberate: `FitResult` has no `Default`, so unit tests drive the algorithm directly via `screen_eta_raw` rather than constructing a full `FitResult`.

## Open questions

None.


JordanBrooks33:
Added commit `1c0553ac`: **xpose4 gold-standard anchor test**.

### What was verified

`xpose4::xpose.gam()` uses `gam::gam(eta ~ ns(cov, df=k))` under the hood — the same natural cubic spline basis as our implementation (confirmed by inspecting `xpose4` default prefs: `smoother3=ns, arg3=df=2; smoother4=ns, arg4=df=3`). Because `gam::gam` with Gaussian family and our `n·ln(RSS/n) + 2p` formula produce identical ΔAIC (the constant `n(1 + ln2π)` cancels in the difference), the comparison gives bit-for-bit agreement.

### Numerical results (two datasets)

| Dataset | ETA×cov pairs | max \|ΔAIC_ferx − ΔAIC_R\| |
|---------|--------------|---------------------------|
| two_cpt_oral_cov (n=30) | 10 | 1.16e-05 |
| cefepime (n=458) | 18 | 2.69e-05 |

Form agreement and top-1 ranking: 100% on both datasets.

### New Tier-1 test

`gam::tests::xpose4_anchor_delta_aic_matches_reference` — uses a fixed synthetic EBE dataset (n=60, R seed=20240101, known covariate structure) embedded via `include_str!`. Asserts all 6 ΔAIC values to tolerance 1e-4 and correct top-1 ranking for each ETA. Runs in <1 ms.

### GAM speed

| Dataset | ETAs × covariates | `gam_screen` time |
|---------|-------------------|-------------------|
| warfarin (n=10, no covariates) | 0 | <0.1 ms |
| two_cpt_oral_cov (n=30) | 5 × 2 | 0.4 ms |
| cefepime (n=458) | 3 × 7 | 0.5 ms |

The GAM step is negligible vs. the model fit in all cases.

JordanBrooks33:
## Coverage fix (commit `96b8fd02`)

Codecov was flagging PR #1169 at 64% patch coverage (202 missed lines). Added 17 new unit tests that cover all the major uncovered paths:

**New helpers covered**
- `gam_screen_raw` — entirely covered; tests check structure, empty-eta early return, shrinkage warning emission, NaN shrinkage (no warning), categorical covariate, and the `delta_aic = aic_null − aic_best` invariant.

**Edge cases in existing helpers**
- `screen_eta_raw` — `include_linear=false` (spline-only form wins), NaN covariate values (subjects skipped, signal still detected on valid subset), fewer than 3 valid pairs (covariate skipped), NaN ETA values (dropped, 28/30 subjects used).
- `ns_basis` — df=0 (empty matrix), empty x (empty matrix), df=3 (four-knot basis, shape + column-0=x + finiteness checks).
- `ols_aic` — constant y (SST < 1e-20 → R²=0 path).
- `categorical_design` — single-level input (returns empty matrix), three-level input (two dummy columns, correct encoding).
- `d_func` — degenerate knots (xi_j == xi_k → denominator guard returns 0).

**Bug fixed along the way**: `best_aic.is_infinite()` was incorrectly skipping a perfect-fit covariate whose AIC = −∞ (RSS = 0). Changed to `best_aic == f64::INFINITY` so only "no form was tried" is skipped; a perfect fit is now recorded with delta_aic = +∞ (the covariate clearly improves the null).

roninsightrx:
## Code review

Nice piece of work — the R anchor (max |ΔAIC diff| ~1e-5 on two datasets) is exactly the right validation for this. Findings below, ordered by severity, plus a diagnosis of the red `codecov/patch`.

### Correctness

**1. `crates/ferx-tools/src/gam.rs:414` — high. The `Categorical` branch has no `n > p` guard.**

The spline branch guards with `n <= df + 1`, but the categorical branch does not. `categorical_design` emits `n_levels - 1` dummies, so a covariate declared `categorical` in `[covariates]` with many levels (STUDY, SITE, a DOSE group) builds a design with `p = n_levels` columns. Near saturation, `n·ln(RSS/n)` goes sharply negative and the `2p` penalty does not compensate — a high-cardinality categorical wins the ranking spuriously. At exact saturation (`n_levels == n`), `RSS = 0` and the covariate disappears entirely via finding 2.

Add the categorical analogue of the spline guard: `if n_dummies + 1 >= n { skip }`, or a minimum residual-df.

**2. `crates/ferx-tools/src/gam.rs:477` — medium. `best_aic.is_infinite()` is also true for `-inf`.**

The sentinel is `f64::INFINITY` ("no form fit"), but any perfectly-fitting design (`RSS = 0` => `aic = n·ln(0) = -inf`, reachable from a saturated categorical or a covariate that is an exact affine function of the ETA) is silently dropped from `covariate_scores` with no warning — the strongest possible signal vanishes from the report.

Use `best_aic == f64::INFINITY` (or `!best_aic.is_finite()` plus an explicit warning), and prevent `RSS = 0` at the source with the guard in finding 1.

**3. `crates/ferx-tools/src/gam.rs:589` — medium. `ols_aic` Choleskys `XᵀX` under a raw truncated-power spline basis.**

Cholesky of `XᵀX` squares the condition number, and the basis is unhelpful going in: column 0 is `x`, later columns are `(x - ξ)³/(ξ_K - ξ)` — uncentered, unscaled and cubed. For a covariate of order 10²–10³ the `XᵀX` entries span ~10¹²+, so the fit either loses most of its digits or `cholesky()` returns `None`. On `None`, line 466 skips the form and line 407 `continue`s the whole covariate — it silently disappears from the ranking with nothing in `GamResult::warnings`, indistinguishable from "screened, not important".

Centre/scale `x` before building the basis and solve via QR/SVD on `X` rather than Cholesky on `XᵀX`; at minimum, push a warning whenever a covariate is dropped for a singular design. (R's `ns()` returns a QR-orthonormalised basis, which is why the R anchor agrees despite this.)

### Tests

**4. `.github/workflows/slow-tests.yml:99` — medium. The new `ferx-tools/slow-tests` feature is never enabled anywhere, so the integration test never runs — in any job.**

The nightly command is `cargo test --no-default-features --features ci,survival,slow-tests ...` with no `--workspace`, so it builds the root `ferx-core` package only and the unqualified features resolve to `ferx-core`'s. `crates/ferx-tools/tests/gam_two_cpt_oral_cov.rs::gam_ranks_wt_and_crcl_for_eta_cl` therefore stays `#[ignore]`d everywhere.

Per the CLAUDE.md feature-qualification rule this needs `--workspace` with package-qualified features: `ferx-core/ci,ferx-core/survival,ferx-tools/slow-tests`.

**5. `crates/ferx-tools/tests/gam_two_cpt_oral_cov.rs:141` — medium. The top-2 assertion is vacuous.**

`data/two_cpt_oral_cov.csv` has exactly two covariate columns (`WT`, `CRCL`) and the base model declares only those two, so `pop.covariate_names.len() == 2` — `top2` *is* the entire score list, and `top2.contains("WT") || top2.contains("CRCL")` cannot fail regardless of what the screen computes. The `cov_names.contains(&"WT")` / `"CRCL"` checks above are equally tautological. Only `max_delta > 1.0` has content.

To make the test discriminate, screen against decoy covariates that carry no signal, or assert the specific ordering / ΔAIC magnitude for WT vs CRCL.

### Lower severity

**6. `gam.rs:185` — low.** When `fit.shrinkage_eta` is empty or shorter than `eta_names` (restored fits, methods that don't populate it), `shrinkage` becomes `NaN` and `shrinkage_warning` returns `None` for `NaN`. The module's entire stated caveat — EBE screening is only valid at low shrinkage — is then silently skipped exactly in the case where shrinkage is *unknown*. Emit a distinct warning instead of staying quiet.

**7. `gam.rs:271` — low.** `gam_screen_raw` is the FFI-facing entry point but validates none of its documented length contract: `eta_names.par_iter().zip(eta_cols).zip(shrinkage)` silently truncates to the shortest, and `screen_eta_raw`'s `eta_values.iter().zip(cov_values.iter())` truncates a short covariate column. An R/Python caller that passes one column short gets a shorter, silently-wrong result rather than an error. Assert the six length invariants stated in the doc comment.

**8. `gam.rs:162` — low.** A name in `opts.covariates` that no subject carries yields an all-NaN column, then `n < 3` at line 395 drops it with no message. Same silent-drop shape as finding 3. Warn on covariate names not found in `pop.covariate_names` before screening.

**9. `gam.rs:415` — low.** A single-level categorical produces a 0-column dummy matrix, so the fit is intercept-only and the covariate is reported with `delta_aic = 0.0` and `best_form: Categorical`, while a *constant continuous* covariate is dropped outright as a singular design. Same degenerate case, two behaviours — pick one (preferably skip both, with a warning).

**10. `docs/` — low.** CLAUDE.md requires a docs page for any user-facing feature, and `_quarto.yml` navigation for new pages; this PR adds a new public `ferx-tools` API with no `docs/**/*.qmd` entry. The comment at `gam.rs:861` also references `docs/gam_anchor_reference/gen_anchor.R`, which is not in the diff — the anchor's reference values are not reproducible from the repo as it stands.

Findings 1 and 2 chain: fix the guard and the `-inf` path stops being reachable.

---

### Why `codecov/patch` is still red at 59.64%

365 hits / 247 misses on 612 new lines, and **all 247 misses are in `crates/ferx-tools/src/gam.rs`**. The extra tests didn't move the number because all 11 unit tests call only `screen_eta_raw`, which was already covered. Breakdown:

| Uncovered | Lines | Why |
|---|---|---|
| `gam_screen()` 122–236 | 90 | only caller is the integration test, which is `#[cfg_attr(not(feature = "slow-tests"), ignore)]` — and per finding 4 that feature is never enabled anywhere, so it runs in no job at all |
| `gam_speed_large_dataset()` 966–1060 | 56 | `#[ignore]`d bench living in `#[cfg(test)] mod tests` **inside `src/`**. `codecov.yml` ignores `tests/**` and `crates/*/tests/**` but not `src/` test modules, so it is measured, never runs, and is a pure drag on the patch % |
| `gam_screen_raw()` 255–302 | 45 | added for testability, then never called by a test |
| `determine_cov_kind()` 308–337 | 29 | no caller outside `gam_screen` |
| degenerate branches 380–618 | 13 | `None => continue`, `n < 3`, singular `XᵀX`, empty df list |
| multi-line assert continuations | 14 | llvm-cov region noise, not actionable |

Two general traps worth internalising: a slow-gated test contributes **zero** patch coverage (slow tests never run on PRs), and an `#[ignore]`d benchmark inside `src/` is worse than useless for the gate — the same code under `crates/ferx-tools/tests/` is excluded from measurement entirely.

Route to 90% (needs <= 61 misses):

1. Move the speed bench to `crates/ferx-tools/tests/gam_speed.rs` — **-56**, no test writing.
2. Unit-test `gam_screen_raw` (pure slices) — **-45**.
3. Unit-test `determine_cov_kind` (`Subject` derives `Default`) — **-29**.
4. Collapse `gam_screen` into `gam_screen_raw`: lines 176–235 duplicate 271–301 (same rayon loop, same shrinkage warning, same `EtaGamResult` build). Extract the columns and delegate — **~-60**, and it removes the duplication.
5. A few degenerate-input asserts on `screen_eta_raw` — **-13**.

Steps 1–3 alone only reach ~81%; `gam_screen` has to be dealt with either way.

I'm pushing fixes for all of the above to this branch.

roninsightrx:
Pushed `db6883c2` to this branch with all ten findings addressed. Summary of what changed, by finding:

| # | Fix |
|---|---|
| 1 | `ols_aic` now refuses any design with no residual df (`n <= p`), and the categorical branch skips a saturated or single-level covariate explicitly. |
| 2 | Sentinel is `== f64::INFINITY`, so a perfect fit (`−∞`) is reported instead of dropped. |
| 3 | Covariates are centred and scaled before any design is built, and the solve is an SVD of `X` rather than a Cholesky of `XᵀX` — which also gives an honest rank test. AIC and R² are invariant under an affine reparameterisation of the column space, so the ranking is unchanged; the R anchor test passes untouched. |
| 4 | `slow-tests.yml` now runs `--workspace` with package-qualified features (`ferx-core/ci,ferx-core/survival,ferx-core/slow-tests,ferx-tools/slow-tests`). |
| 5 | The integration test screens against 4 decoy covariates and requires WT **and** CRCL to be the top two, each ahead of the best decoy by more than 1 ΔAIC. Verified passing under `--features ferx-tools/slow-tests`. |
| 6 | `shrinkage_warning` fires for a NaN shrinkage with its own message. |
| 7 | `gam_screen_raw` asserts its documented length contract instead of letting `zip` truncate. |
| 8 | A requested ETA or covariate name the fit/dataset does not carry is reported and skipped. |
| 9 | A one-level categorical is now skipped like a constant continuous, with a warning. |
| 10 | New `docs/tools/gam-screening.qmd` (in `_quarto.yml`), plus `docs/gam_anchor_reference/gen_anchor.R` — the script `gam.rs` already cited but the PR did not add. |

More broadly: every `continue` in `screen_eta_raw` now emits a warning. Being skipped and being screened-but-unimportant are indistinguishable in a ranking, so each one says which happened and why.

**On the anchor script.** It reproduces `tests/data/gam_anchor_ebes.csv` byte-for-byte and prints all six ΔAIC values the test asserts. One thing to note if you regenerate: the original used `sample(0:1, n, replace = TRUE)` for SEX, not `rbinom`. Both consume the same number of uniforms — so WT, CRCL and ETA_CL come out identical either way and only the SEX column and ETA_V differ. A generator that got that one call wrong would have looked two-thirds correct. There is a comment on the line.

**On coverage.** `gam_screen` is now a thin wrapper over `screen_from_parts`, which takes only the four pieces of `FitResult` it reads, so the column-collection and name-selection logic is unit-testable (`FitResult` has ~70 fields and no `Default`, which is what blocked it). That also removed the rayon loop that `gam_screen` and `gam_screen_raw` were each carrying a copy of. The speed benchmark moved to `crates/ferx-tools/tests/gam_speed.rs` — `codecov.yml` excludes `crates/*/tests/**` but not `src/` test modules, so an `#[ignore]`d bench in `src/` was compiled, measured, never run, and cost the patch gate 56 lines for nothing. A non-ignored companion test keeps it from rotting against a signature change.

The gam unit tests go from 11 to 39. Local gates: `cargo fmt --check`, `cargo clippy -p ferx-tools --all-targets`, `RUSTDOCFLAGS="-D warnings" cargo doc -p ferx-tools`, and `docs-lint --check` (run from `main`'s copy, since this branch predates #1163) are all clean, and the new page raises no violations.

roninsightrx (changes requested):
## Review findings

1. **High — the categorical guard still admits near-saturated labels, and checks exact saturation only after allocating the full dummy matrix** (`crates/ferx-tools/src/gam.rs:552-579`). `categorical_design(&x_vals)` first allocates an `n × (levels-1)` matrix; only afterward does `n_dummies + 1 >= n` skip. An exact one-level-per-subject label can therefore allocate O(n²) memory before being rejected. More importantly, `n > p` alone does not address the failure described in the comment: with 100 unrelated ETA values and a categorical having 99 levels (one duplicated level), this branch accepts the model and reports `ΔAIC = 335.64`, `R² = 0.9951`. It will rank a nearly unique SITE/STUDY identifier far above real signal. Count levels before materialising the matrix, and reject high-cardinality/insufficient-residual-df designs using a defensible threshold (or use an AICc-style penalty); add a regression with `n_levels = n - 1`, not just `n_levels = n`.

2. **Medium — a constant ETA produces a scored covariate with `delta_aic = NaN` and no warning** (`crates/ferx-tools/src/gam.rs:491-500, 666-680`). Repro: 20 ETA values all `0.0` and a continuous covariate `0..19`. Both null and candidate have AIC `-inf`, so `aic_null_local - best_aic` is `NaN`; the result still contains the score and the sort silently treats it as equal. That violates the documented ranked-result contract and can surface for an uninformed/fixed random effect. Detect a zero/non-finite ETA variance before fitting, return no scores with an ETA-level warning, and pin this case.

3. **Medium — switching the slow job to `cargo test --workspace` did not widen its push filter or its guard to workspace members** (`.github/workflows/slow-tests.yml:37-44`; `tests/slow_test_path_filter.rs:327-379`). A future push changing only `crates/ferx-tools/src/gam.rs`, its `Cargo.toml`, or `crates/ferx-tools/tests/gam_two_cpt_oral_cov.rs` will not trigger the push-to-main slow suite. The existing guard still hard-codes root `src/` and `tests/`, so it passes despite that omission. Add `crates/**` (or equivalent member-specific roots) to the workflow and derive/assert member source, tests, and manifests in the guard now that the command is workspace-wide.

Local verification otherwise passed: `cargo test -p ferx-tools --lib` (135 tests), `cargo test -p ferx-tools --test gam_speed`, and `cargo check --workspace --all-targets`.

roninsightrx:
Reviewed the three findings — all three reproduce against the merged code, so all three are fixed. This PR merged before the fixes could land on its branch, so they are in **#1193** instead.

**1 — near-saturated categorical (high).** Confirmed. The previous guard refused only an exactly saturated design (`n <= p`), which leaves the whole band below it open. 100 unrelated ETA values against a 99-level label:

```
covariate   delta_aic     R²        form
SITE          474.499   0.998775   Categorical
WT             -0.656   0.013354   Linear
```

For `k` levels over `n` subjects with no relation to the ETA, `E[RSS/RSS₀] ≈ (n−k)/(n−1)`, so `ΔAIC ≈ n·ln((n−1)/(n−k)) − 2(k−1)` — negative for small `k`, crossing zero near `k ≈ 0.8n`. A categorical must now leave at least as many residual df as it spends parameters (`n >= 2p`), and the level count moved ahead of the allocation so a one-level-per-subject label no longer builds an `n × (n−1)` matrix before being rejected.

On the AICc suggestion: it handles this more smoothly and is the textbook answer, but it shifts every ΔAIC by ~0.1 (for `p=2, n=60` the correction is 0.21 against the null's 0.07), which breaks the R / xpose4 parity this module is validated against to 1e-4. So it is a guard on admissibility, not a change to the reported score. The anchor's six ΔAIC values are byte-identical after the fix.

Regression added at `n_levels = n − 1` as asked, plus a boundary test (10 levels over 20 subjects admitted, 11 refused) and one asserting an ordinary 3-level factor still scores.

**2 — constant ETA → `NaN` (medium).** Confirmed exactly as described:

```
aic_null = -inf
X      dAIC=NaN aic=-inf R2=0
warnings: []
```

The ETA is now refused before any model is fitted, with a warning. A second guard covers the case the whole-column check cannot see — EBEs that vary across the population but are constant over the subset carrying one particular covariate. Also added a blanket `no_score_is_ever_nan` across all three covariate shapes, since the ranking contract is what actually broke here.

**3 — slow-job triggers (medium).** Confirmed. `slow-tests.yml` gained `crates/**`, and the guard now derives the members from the root manifest's `[workspace] members` and requires each one's `src/`, `tests/` and `Cargo.toml`; the fixture-literal harvest reads member test sources too.

One thing that surfaced while fixing it: `--workspace` also pulls `docs-lint` into the slow job, and its tests read `docs/` — so the derivation correctly demanded `docs/**` in the filter, which would make a docs typo trigger the two-hour fit suite. The command now carries `--exclude docs-lint`, and the guard parses that `--exclude` list out of the workflow rather than hard-coding it. Both halves negative-controlled, since a guard that cannot fail is the failure mode that file exists to prevent:

```
drop 'crates/**' from paths:     FAILED — "crates/ — 32 file(s), e.g. crates/docs-lint/Cargo.toml"
drop '--exclude docs-lint':      FAILED — "docs/ — 99 file(s), e.g. docs/.gitignore"
both restored:                   ok. 8 passed
```

PR #1173, opened 2026-09-01
feat(cli): ferx gam subcommand — GAM covariate pre-screening [stacked on #1169]

JordanBrooks33:
## Why

Issue #1114 adds `gam_screen()` to `ferx-tools` but exposes no CLI entry point.
This PR adds a first-class `ferx gam` subcommand.

**Stacked on #1169** (the GAM library PR). The diff here will include #1169's
changes until that PR merges to `main`, after which this PR's diff shrinks to
just the CLI additions (~300 lines).

## What changed

- `crates/ferx-cli/src/gam_cmd.rs` (new) — thin dispatcher following the
  `bootstrap_cmd.rs` pattern: parse flags → `prepare_run` → `fit` →
  `gam_screen` → print ranked table → optional CSV via `write_gam_csv`
- `crates/ferx-tools/src/gam.rs` — adds `pub fn write_gam_csv(result, path)`
  for CSV output
- `crates/ferx-cli/src/main.rs` — `mod gam_cmd`, `ferx gam` dispatch, usage line

### Usage

```
ferx gam two_cpt_oral_cov.ferx --data data/two_cpt_oral_cov.csv --output gam.csv
```

### Flags

| Flag | Default | Description |
|------|---------|-------------|
| `--output PATH` | off | Write results CSV |
| `--spline-df N` | 2, 3 | Spline df to try (repeatable) |
| `--no-linear` | false | Skip linear form |
| `--shrink FRAC` | 0.30 | Shrinkage warning threshold |
| `--threads N` | auto | Rayon worker count |

## Cross-repo dependency
| GAM library | FeRx-NLME/ferx-core#1169 | open | before |
| ferx (R) | TBD | not yet open | after |

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable — CLI thin wrapper; math validated in #1169

## Performance

## Tests
- [x] 7 unit tests in `gam_cmd.rs`: help, missing model, bad model,
  `parse_spline_df` defaults/repeated, `parse_shrink` defaults/out-of-range
- All 83 existing `ferx-cli` tests still pass (`cargo test -p ferx-cli`)

## Docs
- [ ] `CHANGELOG.md` entry — will add once #1169 merges
- [x] `MAIN_USAGE` updated with `ferx gam` line

## Checklist
- [x] `cargo check` clean
- [x] `cargo test -p ferx-cli` — 83 tests pass
- [x] `rustfmt` (nightly, edition 2021)
- [x] No changes to `Dual2` sensitivity path

## Reviewer hints

- The only non-trivial CLI decision: `--spline-df` is repeatable
  (`--spline-df 2 --spline-df 4`) rather than comma-separated.
- `verbose = false` and `run_covariance_step = false` are set on the
  `FitOptions` before fitting so the base-model fit is quiet.


roninsightrx:
## Main question: it's not clear whether this runs an estimation first, and both modes should be possible

Reading `run_gam`, the answer is that it **always** runs a full estimation: `prepare_run` → `ferx_core::fit(...)` (outer loop to convergence) → `gam_screen(&fit, &pop, &opts)`. There's no way to skip it.

But `gam_screen` only reads `fit.subjects[].eta`, `fit.eta_names`, `fit.shrinkage_eta`, `fit.covariate_table` and `pop.subjects[].covariates`. The fit is a pure EBE-supplier — the outer loop isn't what GAM needs, only the etas at *some* parameter vector. So on a large model you pay a complete fit before seeing a single ΔAIC, and re-running the screen with a different `--spline-df` re-fits from scratch.

I'd like all three of these to be selectable:

| Mode | Implementation |
|---|---|
| default — fit, then screen | current behaviour |
| `--no-fit` — EBEs at the initial estimates, no outer loop | **one line.** `FitOptions.outer_maxiter` already exists (`src/types.rs:5991`, model-file key `maxiter`), and `outer_optimizer.rs:124` short-circuits `outer_maxiter == 0` into `evaluate_at_initial_params(...)` — cold-start `run_inner_loop_warm`, `ofv = 2*pop_nll`, `n_iterations: 0`. That's NONMEM `MAXEVAL=0 POSTHOC`. Set it next to the existing `run_covariance_step = false`. Precedent: `ferx bootstrap --dofv` does exactly this at `crates/ferx-tools/src/bootstrap/mod.rs:1018`. |
| `--from-fit run1.fitrx` — reuse a fit that already ran | `ferx_core::io::fitrx::load_fit` (`src/io/fitrx.rs:924`) is already public and returns `LoadedFit { fit, population: Option<Population>, model_source, manifest }`. The round-trip is real — `wire_to_fit_result` + `parse_subjects` rebuild `fit.subjects[].eta`, and `gam_screen` touches nothing else. Read-only; the bundle is never rewritten. |

Caveat on `--no-fit`: the `outer_maxiter == 0` short-circuit lives only in `optimize_population`, reached via the `_ =>` arm at `src/api/fit.rs:1640`. SAEM / IMP / IMPMAP / Bayes / VI dispatch to their own runners and never see it, so `--no-fit` on a `methods = saem` model would silently run a full SAEM. Reject it there, or route those models to a plain FOCE posthoc.

Caveat on `--from-fit`: `LoadedFit.population` is `None` unless the bundle was written with `--include-data`, so `--data` has to be required in that case — otherwise the screen runs against a half-populated covariate set.

## GAM as part of a normal run

`ferx model.ferx --data d.csv --gam` should also be possible, writing the GAM table alongside the other per-run outputs. That gets the common case without paying for a second full fit.

**The `.fitrx` bundle should not carry the GAM results.** It stays clean and holds only data from the fit itself. GAM output is a separate sidecar file — no new zip entry, no `FitResult.gam`, no changes to `save_fit`/`load_fit`. `--from-fit` reads a bundle but never writes one.

Given that, GAM correctly stays in `ferx-tools` and #1169 needs no change.

## Sidecar naming

Default the output to `{model}-gam.csv`, matching the existing `{model}-sdtab.csv` / `{model}-covtab.csv` / `{model}-fit.yaml` convention, rather than producing nothing unless `--output` is passed.

## `--output` is overloaded

At the root, `--output PATH` means "write a `.fitrx` bundle" (`main.rs:27`). Here it means "write a CSV". Same flag, two different artifacts. Rename this one to `--csv`.

## Smaller items in `gam_cmd.rs`

- `run()` only checks `args[2]` for the help flag, so `ferx gam model.ferx --help` fits the model instead of printing help.
- `value()` takes the next token blind: `--data --output x.csv` silently uses `--output` as the data path.
- `--shrink` with no value falls through to `None` → silent 0.30 rather than an error. Same for a non-numeric `--threads`, which `parse_threads` drops.
- No unknown-flag rejection: `--splinedf 4` silently screens with the defaults. For a tool whose flags change the ranking, silent acceptance is the wrong default.
- `--spline-df 3 --spline-df 3` yields `[3, 3]` — duplicate candidates. Dedupe.

JordanBrooks33:
All issues from the review addressed in e908ef89:

**Major additions:**
- `--no-fit`: sets `outer_maxiter = 0` (MAXEVAL=0 equivalent) — computes EBEs at initial parameters without optimizing
- `--from-fit <path>`: loads an existing `.fitrx` bundle and skips fitting entirely; if the bundle was saved without `--include-data`, recovers the population by writing the embedded model source to a temp file and calling `prepare_run` with `--data`
- `--gam` flag on the normal fit flow: `ferx model.ferx --data d.csv --gam` runs a full fit then GAM in one step, writing `{model}-gam.csv`

**Renames / defaults:**
- `--output` → `--csv` (avoids collision with root CLI's `--output` for `.fitrx`)
- CSV output now defaults to `{model}-gam.csv` — no flag required

**Bug fixes:**
- Help (`-h`/`--help`) recognized at any position in the arg list
- `value()` helper now errors when the next token starts with `-`
- `--shrink` / `--threads` with missing or bad values now error explicitly instead of silently defaulting
- Unknown flags are rejected with a clear error message
- `--spline-df` deduplicates repeated values (`--spline-df 2 --spline-df 2` → `[2]`)

15 new unit tests; all 98 ferx-cli tests pass.

JordanBrooks33:
Addressed the remaining `--no-fit` caveat from the review in 81249ff5.

**Problem**: `outer_maxiter == 0` short-circuits only in `optimize_population` (the `_ =>` arm in `api/fit.rs`). SAEM, IMP, IMPMAP, Bayes, and VI each dispatch to their own runner before reaching that arm, so setting `outer_maxiter = 0` would silently have no effect — a full estimation run would proceed.

**Fix**: added `no_fit_supported(m: EstimationMethod) -> bool` that returns `false` for `Saem | Imp | Impmap | Bayes | Vi`. After `prepare_run`, before touching `outer_maxiter`, the effective method list (`fit_options.methods` when non-empty, else `[fit_options.method]`) is scanned; if any method is unsupported the subcommand exits with:

```
Error: --no-fit is not supported with method 'SAEM': outer_maxiter=0 only
short-circuits the FOCE/BFGS outer loop; SAEM, IMP, IMPMAP, Bayes, and VI
have their own runners and would run a full estimation. Run the fit first
and use --from-fit instead.
```

The error message explicitly points to `--from-fit` as the alternative. Two unit tests added (`no_fit_supported_allows_foce_variants`, `no_fit_supported_rejects_own_runner_methods`); all 24 gam_cmd tests and 16 CLI integration tests pass.

roninsightrx:
## Review — `feat(cli): ferx gam`

Reviewed at medium effort against `main` (post-merge, `behind_by: 0`). Nice feature; the `gam.rs` numerics and their xpose4 anchor look solid. Findings are almost all in the CLI wrapper.

One correctness bug worth fixing before merge, two data-integrity issues, and some repo-contract gaps.

---

### 1. `--no-fit` runs a full 100-iteration FOCEI fit on `gn_hybrid` models — high

`crates/ferx-cli/src/gam_cmd.rs:145`

`no_fit_supported` assumes every method except SAEM/IMP/IMPMAP/Bayes/VI reaches the `outer_maxiter == 0` short-circuit in `optimize_population`. `FoceGnHybrid` never gets there:

- `src/api/fit.rs:1626` dispatches `FoceGn | FoceGnHybrid` to `gauss_newton::run_foce_gn`, bypassing `optimize_population`.
- With `outer_maxiter = 0` the GN loop runs zero iterations and falls through to the FOCEI polish, which **hard-sets** `polish_options.outer_maxiter = 100` (`src/estimation/gauss_newton.rs:494`) — unconditionally, not derived from `options`.

So `ferx gam model.ferx --data d.csv --no-fit` on a `method = gn_hybrid` model silently performs a full estimation. That is exactly the failure the guard's own doc comment says it prevents, and it fails silently — the user gets EBEs, just after an unrequested fit.

Add `FoceGnHybrid` to the rejected set. Pure `FoceGn` is safe, but only incidentally: its loop is `0..maxiter` and it returns before reaching the polish. Worth a comment saying so, since the two variants now diverge for a non-obvious reason.

### 2. Model path can be another flag's value — medium

`crates/ferx-cli/src/gam_cmd.rs:223`

`args.iter().skip(2).find(|a| !a.starts_with('-'))` has no idea which tokens were consumed as flag values:

```
ferx gam --threads 4 model.ferx --data d.csv   # picks "4" as the model file
ferx gam --data d.csv model.ferx               # picks "d.csv" as the model file
```

Both then fail with a parse error against the wrong file. `check_unknown_flags` already walks the argument list with value-flag knowledge — reuse that walk to collect positionals, and reject more than one while you're there.

### 3. `--from-fit` + `--data` can silently mis-pair every subject — medium

`crates/ferx-cli/src/gam_cmd.rs:189`

When the bundle carries no embedded population, the CLI re-reads a user-supplied CSV and hands it to `gam_screen` alongside the bundle's `fit`. `gam_screen` aligns subjects **by index**. Two outcomes:

- Different subject count → `gam_screen_raw`'s `assert_eq!(col.len(), n_subjects, …)` panics. A raw panic and backtrace out of a CLI, not an error message.
- Same count, different order (or a different `[data_selection]` filter) → every η is paired with the wrong subject's covariates and the ranking is **silently wrong**. The gam module's own docs call that "the worst outcome for a tool whose entire output is an ordering."

Compare `fit.subjects[i].id` against `population.subjects[i].id` and return a `String` error on the first mismatch.

### 4. CSV is written unconditionally — low

`crates/ferx-cli/src/gam_cmd.rs:337`

`run_gam` always writes `csv_flag.unwrap_or_else(|| format!("{model_stem}-gam.csv"))`, so a plain `ferx gam m.ferx --data d.csv` silently overwrites an existing `m-gam.csv` in the CWD.

`GAM_USAGE` documents this ("default: {model}-gam.csv") but the PR description says `--output PATH … off`, i.e. opt-in. The PR description also documents `--output`, which `check_unknown_flags` rejects as unknown. Pick one and make the three agree.

### 5. Predictable temp path + `unwrap` — low

`crates/ferx-cli/src/gam_cmd.rs:205`

`std::env::temp_dir().join(format!("ferx_gam_{}.ferx", pid))` is guessable, and `fs::write` follows symlinks — a symlink-overwrite footgun on a shared box. The following `tmp_path.to_str().unwrap()` also panics on a non-UTF-8 temp dir. Randomize the name (or use `tempfile`) and propagate the path error.

### 6. Method-chain resolution duplicated — low

`crates/ferx-cli/src/gam_cmd.rs:241`

`if methods.is_empty() { vec![method] } else { methods.clone() }` is character-for-character `FitOptions::method_chain()` (`src/types.rs:7367`). Call it so the guard tracks any future change to chain resolution.

Related nit: the guard rejects `Imp` even when `imp_eval_only = true`, which is genuinely evaluation-only and harmless.

### 7. `write_gam_csv` does no quoting, emits bare `NaN` — low

`crates/ferx-tools/src/gam.rs:399`

`eta_name` and `covariate` come from user model/data text; a covariate named `WT,KG` shifts every downstream column. `shrinkage` is `f64::NAN` for a degenerate Ω, which formats as `NaN` and will not parse as numeric in R or pandas without `na.strings`. Quote fields containing `,`, `"` or newline, and emit an empty field for non-finite values.

### 8. `--gam` on the main fit path — low

`crates/ferx-cli/src/main.rs:277`

Uses `GamOptions::default()` and a fixed `{model}-gam.csv`, with no `--spline-df` / `--no-linear` / `--shrink` equivalents and no test. It also runs after `--simulate`, where `population.covariate_names` is typically empty, so the only output is a warning plus a header-only CSV. Consider dropping it in favour of the subcommand, or covering it with a test.

### 9. Repo-contract gaps — low

- **No `CHANGELOG.md` entry** under `## [Unreleased]` for two user-visible additions (`ferx gam`, `--gam`). CLAUDE.md requires it in the same PR. The checklist defers this to "once #1169 merges" — #1169 is merged (`a8617572`), so it is unblocked.
- **No `docs/**/*.qmd` page** or `docs/_quarto.yml` entry for the new subcommand; CLAUDE.md requires one for any user-visible feature.
- **`run_gam` has no test at any tier.** ~180 lines covering `--from-fit`, `--no-fit`, fitting, table printing and the CSV write; the unit tests reach only the flag parsers and two early-error exits. That is almost certainly why `codecov/patch` is red — and #1 above is exactly the kind of bug a `--no-fit` test would have caught.

---

**Note on the earlier CI failure:** the two `Tests + coverage` jobs that were red on this PR were not caused by this diff — `main` was red from #1192 (a stack-overflow abort in `pool_plan_tests` that killed the whole lib binary). Fixed in #1194, now merged, and this branch has it. The remaining `codecov/patch` failure is this PR's own, per item 9.

https://claude.ai/code/session_014VMb63MwV4MCwxWdH2kbd4

roninsightrx:
Pushed `e7bab2ae` addressing all nine findings.

| # | Finding | Fix |
|---|---|---|
| 1 | `--no-fit` ran a full fit on `gn_hybrid` | `FoceGnHybrid` rejected; `imp` + `imp_eval_only` now allowed |
| 2 | Model path could be a flag's value | `scan_args` collects positionals with value-flag knowledge; >1 refused |
| 3 | `--from-fit` + `--data` mis-paired subjects | subject ids compared, mismatch is an error |
| 4 | CSV written unconditionally | added `--no-csv`; default kept (see below) |
| 5 | Predictable temp path + `unwrap` | private `create_dir` directory, path error propagated |
| 6 | `method_chain()` duplicated | now calls it |
| 7 | No CSV quoting, bare `NaN` | fields quoted when needed, non-finite written empty |
| 8 | `--gam` untested | extracted to `gam_cmd::run_after_fit`, tested |
| 9 | CHANGELOG / docs / `run_gam` untested | all three added |

**On #4** — I kept the default `{model}-gam.csv` write rather than making it opt-in. Writing a default-named artefact next to the model is what a plain `ferx` run already does (`{model}-fit.yaml`, `{model}-sdtab.csv`), so opt-in would make `gam` the odd one out. `--no-csv` gives the escape hatch, and `--csv` + `--no-csv` together is an error. The PR description still documents an `--output PATH` flag that does not exist and describes the CSV as off by default — worth updating it to match `--csv` / `--no-csv`.

**On #6's nit** — `imp_eval_only` is genuinely evaluation-only per its own docs ("evaluate `−2 log L` at the fixed input parameters without updating them"), so `--no-fit` with it is honest. The flag is IMP-specific and does not excuse SAEM/IMPMAP/Bayes/VI; there's a test pinning that.

**Tests** — 13 new. The ones that matter:

- `no_fit_supported_rejects_gn_hybrid` — would have caught #1.
- `a_flag_value_is_not_mistaken_for_the_model_file` — the two argument orders from #2.
- `a_shorter_population_is_refused_rather_than_asserted_on` and `a_reordered_population_is_refused_rather_than_silently_mispaired` — both halves of #3.
- `a_covariate_name_with_a_comma_does_not_shift_the_columns`, `a_nan_shrinkage_is_written_as_an_empty_field_not_the_text_nan` — #7.
- `no_fit_runs_end_to_end_and_writes_the_csv`, `no_csv_prints_the_table_but_writes_nothing`, `run_after_fit_writes_the_default_named_csv` — `run_gam` and `--gam` had no coverage at any tier, which is what `codecov/patch` was failing on.

These use `examples/two_cpt_oral_cov.ferx` + `data/two_cpt_oral_cov.csv` rather than warfarin: warfarin declares no `[covariates]`, so the screen produces a header-only CSV and asserts nothing. Whole CLI suite runs in ~3s.

**Gates** — `tools/preflight.sh` green end to end (fmt, the 5 check sets, clippy, rustdoc, docs lint, public-API baseline). `cargo test -p ferx-cli -p ferx-tools`: 97 + 139 + 26 pass, 0 fail. No public-API change, so no baseline update.

Note the two `Tests + coverage` jobs that were red here were `main`'s doing, not this diff — #1194 fixed that and they went green on the last run.

https://claude.ai/code/session_014VMb63MwV4MCwxWdH2kbd4

Issue #1167, opened 2026-08-31
bug(survival): the joint PK-TTE objective can integrate its Gaussian half and its hazard half on two different engines


TeunP:
## Problem

`ode_cumhaz_hazard` (`src/survival/mod.rs:710`) is unconditional about its engine:

```rust
let pk = (model.pk_param_fn)(theta, eta, &subject.covariates, 0.0);
let states = crate::ode::ode_dense_solve_states(ode, &pk.values, theta, eta, subject, times);
```

It always takes the **dense** driver with a single `t = 0` PK snapshot. There is no
model-time gate on it and no time-varying-covariate gate either.

`try_joint_pktte_shared_solve` (`src/stats/likelihood.rs`) is what normally keeps the two
halves of the objective on one engine: when it admits, the Gaussian term and the
cumulative-hazard term come from a single `ode_predictions_and_chz` integration. When it
declines, the Gaussian term goes to `compute_predictions_with_tv_*` — which routes to the
**event-driven** walk for TV covariates, resets or model time — while the hazard term
stays on `ode_dense_solve_states`.

So for any model that trips a decline clause, one OFV is assembled from two engines
integrating the same augmented ODE system, and the two break the timeline at different
points.

## Why this is a correctness question, not a tolerance nit

The repo already takes this position for the neighbouring case. `check_survival_tv_covariates`
(`src/api/validation.rs:2015-2031`) **rejects the fit outright** when an ODE-accumulated
hazard meets a time-varying covariate, on the grounds that "the survival ODE solve freezes
the whole PK-parameter vector at `t=0`".

There is no analogous rejection, gate, or warning for model time. `src/api/predict.rs:316-317`
puts the same dense curve into `predict_survival` while IPRED comes from the event-driven
engine.

## Options

1. Give `ode_cumhaz_hazard` the per-event route, so it follows the same routing decision
   the value path makes.
2. Put a `model_uses_time_anywhere` rejection next to the existing TV-covariate one, so the
   combination is refused rather than silently split.

(1) is the real fix; (2) is the honest interim if (1) is expensive.

Note this interacts with the `__chz` over-fire filed separately as #1166: if that is fixed first,
the population reaching this split shrinks to models whose **PK block** genuinely reads
model time, which is the set that actually needs the event-driven hazard solve.

## Provenance

Found by the scoped adversarial review of PR #1130 (#1124). Pre-existing — that PR widens
the decline clause, which increases how often the split is reached, but the split itself is
older.

[...]

Issue #1166, opened 2026-08-31
bug(sens): model_uses_time_anywhere counts the injected d/dt(__chz_n) hazard lines, so every time-dependent-hazard joint PK-TTE model is treated as a non-autonomous PK model


TeunP:
## Problem

`model_uses_time_anywhere` (`src/pk/mod.rs`) answers "does this model read model time
anywhere", and for the `[odes]` half it asks
`s.rhs_program.reads_model_time()`. That program's flags are computed at
`src/parser/model_parser.rs:11463-11468` from `stmts_owned`:

```rust
let uses_time_vars = stmts_read_slots(&stmts_owned, &[time_slot, tafd_slot, tad_slot]);
let rhs_reads_time_builtin = stmts_read_time_builtin(&stmts_owned);
```

and `stmts_owned` includes the injected joint-PK-TTE hazard lines. The parser says so
itself, a few hundred lines up (`model_parser.rs:11266-11267`):

> the `ode_template`-generated equations plus the injected joint-PK-TTE
> `d/dt(__chz_n)` hazard lines are all in `stmts_owned` by now.

So a **standard** joint PK-TTE model — a Weibull or Gompertz baseline hazard, which by
definition reads `TIME` — makes `model_uses_time_anywhere` return `true` for the whole
model, even when the PK block is fully autonomous:

```
[odes]
  d/dt(depot)   = -KA*depot
  d/dt(central) =  KA*depot - (CL/V)*central          # autonomous
  d/dt(__chz_1) =  SHAPE/SCALE*(TIME/SCALE)^(SHAPE-1) * exp(BETA*central/V)   # injected
```

The injected equation never writes a PK state, so the hazard cannot influence the PK
dynamics *through the injected line*, and superposition over the PK states was never invalid
for such a model.

Worth stating precisely, because the fix below depends on it: this is a property of every
normal model, **not** a structurally enforced one. `__chz_<cmt>` is pushed onto
`state_names` before `build_ode_spec` is called (`model_parser.rs:1973-1984`), and the guard
there rejects only a *collision* — a user-declared state of the same name. A user RHS that
*references* `__chz_1` still compiles. So the fix has to establish the no-feedback
precondition (reject such a reference, or detect it and fall back to the wide flag) rather
than assume it.

## Consequences

Every consumer of the predicate treats these models as if their PK dynamics were
non-autonomous:

- `stats/likelihood.rs` — declines #570's single shared solve, so the objective takes
[...]

PR #1174, opened 2026-09-01
fix(ode): break the timeline at a route's onset in the dense builders [closes #1171]

### LINKED ISSUE #1171: bug(ode): ode_predictions_with_states drops a lagged zero_order input-rate contribution that ode_predictions applies
## Problem

`ode_predictions_with_states` and `ode_predictions` are documented and used as the same
integration, one of them additionally returning the states. They are not. For a model whose
`[odes]` RHS carries a **lagged `zero_order` input-rate forcing**, the `_with_states` variant
omits the forcing's contribution entirely.

Both are public (`api/ferx-core-public-api.txt`), so an external caller that wants states —
the ferx-r glue, a tool building diagnostics — silently gets a different trajectory than
`predict()` does.

## Reproduction

Model (`MR_ZERO_ORDER_LAGGED_1CPT` in `src/sens/provider_tests.rs`), a one-compartment
model with a parallel first-order route and a **lagged** zero-order route:

```
[odes]
  d/dt(central) = FZO1*first_order(ka=KA) + FZO*zero_order(dur=DUR, lag=LAGZO) - CL/V*central
[scaling]
  y = central / V
```

with `theta = [TVCL 5.0, TVV 50.0, TVFZO 0.5, TVKA 1.2, TVDUR 2.0, TVLAGZO 1.5]`,
`eta = [0.1, -0.05]`, a single 100 mg dose at t = 0, and observations
`[0.5, 1.1, 2.3, 3.1, 4.2, 8.0, 12.0]`. The spec pins `ode_reltol = 1e-10`,
`ode_abstol = 1e-12`, so this is not a tolerance effect.

```rust
let plain  = ode_predictions(spec, &pk.values, &theta, &eta, &subject);
let (withst, _) = ode_predictions_with_states(spec, &pk.values, &theta, &eta, &subject);
```

```
ode_predictions             [0.45947752, 0.71338586, 1.21890768, 1.55120075, 1.57178439, 1.01553899, 0.63811447]
ode_predictions_with_states [0.45947752, 0.71338586, 0.81735003, 0.78372689, 0.70698963, 0.45941281, 0.28869906]
```

The two agree **exactly** at t = 0.5 and t = 1.1 — both before `LAGZO = 1.5` — and diverge
from the first observation inside the zero-order window onward. By t = 12 the `_with_states`
value is 45% of the correct one.

## Which one is right

`ode_predictions`. Computed independently of ferx, at t = 2.3:

- `CL = 5.0·e^{0.1} = 5.52585`, `V = 50.0·e^{-0.05} = 47.5615`, `ke = CL/V = 0.116183`
- first-order route: `0.5·100·(ka/(ka−ke))·(e^{−ke·t} − e^{−ka·t}) = 38.876`
- zero-order route, `R = 0.5·100/2 = 25` over `[1.5, 3.5]`, elapsed `τ = 0.8`:
  `(R/ke)·(1 − e^{−ke·τ}) = 19.099`
- `y = (38.876 + 19.099)/47.5615 = 1.21895`

matching `ode_predictions`' 1.2189077 to five significant figures.

And `0.81735` — the `_with_states` answer — is exactly `38.876/47.5615`, i.e. the trajectory
with the zero-order route **absent**, not merely mis-timed.

## Note on the shape

`ode_predictions_with_states` carries its own dose/rate pre-pass rather than sharing
`ode_predictions`' — its body even points at the difference ("the separate pre-pass in
`ode_predictions`"). Two implementations of one contract is how this drifted; the fix is
probably to have one of them delegate rather than to patch the copy.

## Provenance

Found while adding a closed-form-vs-ODE-twin self-check in #1130 (#1124). The check pointed
at the closed form; hand-computing the superposition showed the closed form was right and the
twin — built on `_with_states` — was the wrong side. #1130 no longer uses `_with_states`,
so this is not a blocker there. The public-API defect stands on its own.

TeunP:
## Why

`zero_order(dur=D, lag=L)` is delivered as a **per-segment constant rate**: the per-segment
filter (`active_zero_order_inputs`) injects it only when the segment sits *fully inside* the
window `[t_dose + L, t_dose + L + D]`, so the integrator has to break its timeline at the
route's onset (`w_start`) as well as at its end. `push_zero_order_break_times` pushes only
`w_end`; `w_start` has no other source, and observation times land in `saveat` without splitting
segments.

Two of the four break-time builders in `src/ode/predictions.rs` never called
`push_route_lag_break_times`. The segment spanning `w_start` therefore failed the containment
test and **the constant rate was dropped for the entire window**. On a model whose only input is
a lagged `zero_order`, that leaves the compartment at exactly `0.0` at every requested time.

Measured against a new NONMEM anchor (subject 1, η = 0, worst relative difference over the 10
observations):

| path | before | after |
|---|---|---|
| `compute_predictions_with_tv` (objective / `predict()`) | 2.839e-5 | 2.839e-5 |
| `compute_predictions_with_states` (sdtab IPRED + states) | **1.000e0** | 2.839e-5 |
| `ode_dense_solve_states` (the shared dense engine) | **1.000e0** | 2.839e-5 |

`1.000e0` is not a rounding artefact — it is `0.0` at every observation. 2.839e-5 is the NONMEM
`$TABLE` print precision, so both fixed paths agree with NONMEM to the table's own resolution.

The pure-Gaussian objective is **not** affected: `compute_predictions_with_tv` →
`ode_predictions` always had the call. That is exactly why no existing test caught this.

### Six surfaces, all measured

| builder | surface | before |
|---|---|---|
| `ode_predictions_with_states` (inline) | sdtab IPRED + compartment states, plain subjects | all `0.0` |
| `build_segment_break_times` → `ode_dense_solve_states` | sdtab compartment states for TV-covariate / reset subjects (IPRED stays correct — that asymmetry hid it) | all `0.0` |
| " | **joint PK-TTE `H`/`h`** (`survival::ode_cumhaz_hazard`, and `stats::likelihood::tte_ode_nll` in the objective) | `H(t) ≡ H0·t`, `h(t) ≡ H0` |
| " | the Markov / CTMM endpoint NLL | all-zero state trajectory |
| " | the adaptive-dosing window AUC signal (#391) | zero AUC |
| " | `[derived]` grid integrals | `0` |
| `build_segment_break_times` → `ode_solve_until_chz_threshold` | **`simulate()` event times** | `Crossed(t=16.395)` → `CensoredAtHorizon` |

The joint PK-TTE numbers, same fixture plus `hazard = H0*exp(BETA*(central/V))`, `H0 = 0.01`,
`BETA = 0.5`:

```
without   H(t) = [0.0400, 0.0800, 0.1200, 0.1800, 0.2400]   h(t) = [0.01, 0.01, 0.01, 0.01, 0.01]
with      H(t) = [0.0601, 0.1418, 0.2063, 0.3435, 0.4641]   h(t) = [0.0237, 0.0178, 0.0147, 0.0251, 0.0166]
```

Without the fix `H` is *exactly* the drug-free baseline and `h` *exactly* `H0` — the entire drug
effect is absent, because `central = 0` makes `exp(BETA·central/V) = 1`. `H(24)` is 52 % of its
true value, and it feeds the TTE likelihood directly. So the claim "this cannot move any OFV"
holds for a **pure-Gaussian** fit only; a joint PK-TTE objective whose
`try_joint_pktte_shared_solve` declines is wrong today and this changes it.

## What changed

Two call sites, one signature widening. `push_route_lag_break_times` already exists and is
already called by `collect_dose_break_times`, the prediction engines' builder.

1. `ode_predictions_with_states`' inline builder: added the call before
   `push_zero_order_break_times`.
2. `build_segment_break_times`: same call, plus `ode` and `pk_params_flat` threaded in (it had
   neither). Its two call sites — `ode_dense_solve_states` and `ode_solve_until_chz_threshold` —
   already hold both, so they take two extra arguments and nothing else changes.

`sens/ode_provider.rs` is unaffected: it has zero calls to `active_zero_order_inputs` and carries
`K_ROUTE_ONSET` as a first-class timeline event kind instead. Different mechanism — this defect
cannot reach the sensitivity path.

The helper early-exits when no forcing carries a `lag_slot`, so it is a no-op for any model
without an explicit `lag=` on an input-rate function.

## Alternatives considered

**Deleting the 245-line near-copy instead of patching it.** `ode_predictions_with_states` is
`ode_predictions` plus a state capture, and `integrate_segment` already returns states at soft
sample times, so the delegation is real. It is deliberately *not* in this PR: a 245-line deletion
alongside a behaviour change makes any divergence unattributable, it has its own traps (the
`#[cfg(feature = "survival")]` gate on the states capture, Hermite-at-obs-time vs
`u.clone()`-at-break needing a bit-equality check first), and it does not help
`build_segment_break_times` at all. Filed separately.

**Folding in #1167** (the joint objective assembling its Gaussian and hazard halves from two
different engines). They overlap at exactly one call site and neither subsumes the other:
`ode_cumhaz_hazard` is 1 of 5 callers of `ode_dense_solve_states`, so #1167's reroute would leave
four consumers on the defective engine; and fixing the forcing leaves the engine split intact.
#1167's option (2) — reject on model time / TV covariates — would not have caught this either:
the fixture that breaks has neither. Ordering: **#1171 → #1166 → #1167**.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` is clean); the sdtab columns keep
their names and shapes, they just carry the right numbers.

## Breaking changes
- [x] None

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0**, new anchor `nonmem_anchor/lagged_zo.ctl`
      (`ADVAN13 TOL=9`, `MAXEVAL=0 POSTHOC`, all `$THETA` `FIX`), `#OBJV = -264.65083533221622`.

Two design points, both of which the obvious alternative gets wrong:

* **`D1` + `ALAG1` with `RATE=-2`, not `mixed_zero_first.ctl`'s `PODO`/`F1=0` `$DES` trick.**
  `PODO` tracks only the most recent dose, and this fixture needs the *second* window to open
  while drug from the first is still present. NONMEM's native duration-modeled dose superposes
  correctly.
* **The target is `PRED`/`IPRED`, not the OFV.** ferx's objective path was already correct, so an
  OFV-only anchor passes vacuously in both directions — it could not have caught this.

Setup validated before it was trusted: NONMEM `PRED` at `t = 2.3` is `7.6884E-01` against
`0.768837` hand-computed from the closed form (`nonmem_anchor/simulate_lagged_zo.py`).

Only the `lag_cmt = 0` case is expressible — `ALAGn` is per-compartment, so a route lag composed
with a compartment lag stops being the same object (the limitation
`tests/dose_form_lag_nonmem_anchor.rs` already records).

```
                     ferx (subject 1, η = 0)                   NONMEM PRED
 t=2.3    0.7688365361308862                                   7.6884E-01
 t=3.1    1.4785621103212718                                   1.4786E+00
 t=4.2    1.6901432556833500                                   1.6901E+00
 t=8.0    1.1558237486843068                                   1.1558E+00
 t=11.5   0.8144952295718795                                   8.1450E-01
 t=12.4   0.7443925905640648                                   7.4439E-01
 t=13.8   0.9426884935035001                                   9.4269E-01
 t=15.0   1.9668853231444698                                   1.9669E+00
 t=18.0   1.8369301093539890                                   1.8369E+00
 t=24.0   1.0081286188376166                                   1.0081E+00
```

Before the fix every one of those ferx values is `0.0`.

Fixture non-degeneracy (`CLAUDE.md`): **two doses**, the second landing at `t = 12` with ~38 mg
still present, so the later window's *incoming* side is live. Records sit on both sides of both
windows, and `t = 12.4` is inside the second dose's pre-arrival gap with drug from the first
still in the compartment.

## Performance
- [x] No significant impact expected — the helper's `filter(|f| f.lag_slot.is_some())` yields
      nothing for a model without an explicit `lag=`, so the common case is a no-op. Models that
      do carry a route lag gain up to one break per (lagged forcing × dose).

## Tests

**Tier 1** (`src/ode/predictions_tests.rs`) — a zero-order *accumulator* spec (`dy = 0`), whose
exact mass at any `t` is a clamped ramp computable in closed form with no integrator:

* `route_lagged_zero_order_break_builders_agree` — `collect_dose_break_times` and
  `build_segment_break_times` must produce the identical list for a route-lagged subject, with a
  non-degeneracy check that both windows' onsets are actually in it.
* `route_lagged_zero_order_reaches_every_dense_engine` — `ode_predictions`,
  `ode_predictions_with_states` and `ode_dense_solve_states` all against the closed-form ramp, at
  sample times that include a *partially* delivered window (where neither an all-delivered nor an
  all-dropped window can produce the right number).

**Tier 2** (`tests/lagged_zero_order_break_times.rs`, runs on every PR) — sdtab IPRED vs the
objective path at the same η (η = 0 and η ≠ 0; `predict()` is not a valid comparator, it
evaluates at `zero_eta`), compartment states vs the scaled IPRED, `ode_dense_solve_states` vs the
objective path, a **TV-covariate** subject with uniform-valued covariates (asserting
`has_tv_covariates()` so a future value-aware predicate fails loudly rather than silently
destroying the isolation), the metamorphic route-lag ≡ compartment-lag twin driven through the
states path, the `[derived]` grid integral against a closed-form mass balance, the joint PK-TTE
`h(t)` against `H0·exp(BETA·C(t))` on the anchored concentrations, and `simulate()` event times
against a `BETA = 0` twin off the same seed.

**Tier 2 anchor, ungated** (`tests/lagged_zero_order_nonmem_anchor.rs`) — the `PRED`-column
checks are single ODE evaluations at η = 0 with no convergence loop, so they run on every PR and
carry the diff's Codecov patch coverage (slow-gated tests earn none). They follow
`dose_form_lag_nonmem_anchor.rs`'s precedent of comparing like with like: ferx at η = 0 against
`PRED`, never against `IPRED`. `the_anchor_discriminates_a_dropped_zero_order_window` pins the
separation so the anchor cannot quietly stop testing anything.

**Tier 3, slow-gated** — `sdtab_ipred_matches_nonmem_posthoc` runs a `MAXEVAL=0 POSTHOC` fit and
compares the sdtab `IPRED` column against NONMEM's, at 3 % (the two inner optimizers stop at
slightly different η̂; still 30× under the defect's relative 1).

### Mutation matrix

Each fix toggled separately, one variable at a time:

| test | row 3 off | row 4 off |
|---|---|---|
| `with_states_ipred_matches_the_objective_path_at_the_same_eta` | FAIL | pass |
| `with_states_compartment_states_are_the_scaled_ipred` | FAIL | pass |
| `route_lag_equals_compartment_lag_through_the_states_path` | FAIL | pass |
| `states_path_ipred_matches_the_nonmem_pred_column` | FAIL | pass |
| `states_path_compartment_amounts_match_the_nonmem_pred_column` | FAIL | pass |
| `sdtab_ipred_matches_nonmem_posthoc` | FAIL | pass |
| `dense_solve_states_matches_the_objective_path` | pass | FAIL |
| `dense_solve_states_matches_the_nonmem_pred_column` | pass | FAIL |
| `joint_pk_tte_hazard_sees_the_lagged_zero_order_route` | pass | FAIL |
| `simulated_event_times_see_the_lagged_zero_order_route` | pass | FAIL |
| `route_lagged_zero_order_break_builders_agree` | pass | FAIL |
| `tv_covariate_subject_states_match_the_plain_subject` | FAIL | FAIL |
| `derived_grid_integral_over_a_lagged_zero_order_compartment` | FAIL | FAIL |
| `route_lagged_zero_order_reaches_every_dense_engine` | FAIL | FAIL |
| `objective_path_matches_the_nonmem_pred_column` | pass | pass (control: the objective is unaffected) |
| `the_anchor_discriminates_a_dropped_zero_order_window` | pass | pass (a property of the table alone) |

- [x] Not applicable (a fix restoring documented behaviour — no new API surface). The reproducer
      is committed as `nonmem_anchor/lagged_zo_fit.ferx`; on `main` its sdtab `IPRED` and
      `compartments[0]` columns are `0.0` at every record.

<details>
<summary>Reproducer</summary>

```toml
[structural_model]
  ode(states=[central])

[odes]
  d/dt(central) = zero_order(dur=DUR, lag=LAG) - CL/V*central

[scaling]
  y = central / V
```

`CL = 5`, `V = 50`, `DUR = 2`, `LAG = 1.5`, 100 mg at t = 0 and t = 12, η = 0:

```
             before                  after       NONMEM PRED
 t=2.3       0.0                     0.768837    7.6884E-01
 t=15.0      0.0                     1.966885    1.9669E+00
 t=24.0      0.0                     1.008129    1.0081E+00
```

</details>

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` → `Fixed`, naming the simulate and joint-TTE impact, not only sdtab
- [x] `nonmem_anchor/README.md` — a row for the new anchor family
- [ ] No `docs/**/*.qmd` change: the documented behaviour of `zero_order(dur, lag)` was already
      what this restores.

## Checklist
- [x] Nothing on the `Dual2` path touched (`sens/ode_provider.rs` never calls the affected helper)
- [ ] `Cargo.toml` version bump — not breaking

### ferx-site
- [x] No new `.ferx` DSL syntax or `[fit_options]` key — `zero_order(dur, lag)` is unchanged and
      already documented in `docs/model-file/absorption.qmd`
- [x] No new example `.ferx` in `examples/` (the reproducer lives under `nonmem_anchor/`, where
      the other anchor twins do)
- [x] No new file in `data/` — the anchor's dataset lives in `nonmem_anchor/`

### ferx-book
- [x] No estimator, DSL feature or fit option added

### Example execution
- [x] No example execution step needed — no user-visible API or DSL change; the fix restores the
      documented behaviour of an existing construct. Verified instead by the NONMEM anchor and by
      `cargo test --features ci,survival --no-fail-fast`: **164 binaries, 4748 passed, 169
      ignored**. That run had one failure, `ci_workflow_endpoint_coverage` — the guard correctly
      demanding the new `survival`-cfg test file be added to the endpoint-coverage job's `--test`
      list. Added in `.github/workflows/ci.yml` and re-verified green.

## Reviewer hints

Focus on `build_segment_break_times`' widened signature and its two call sites: both already held
`ode` and `pk_params_flat`, so the threading is mechanical, but it is where a wrong `subject`
shadow or a mismatched `dose_lagtimes` slice would hide. Note the borrow differs between the two
new call sites (`&dose_lagtimes` on the inline builder, `dose_lagtimes` inside
`build_segment_break_times`, where it is already a slice).

The `KEEP-IN-SYNC` note on `ode_predictions_with_states` is what failed here — a comment. Both
doc comments now point at the tests that assert the invariant instead.

Everything else is tests. The one place worth reading closely is
`tv_covariate_subject_states_match_the_plain_subject`: its isolation rests on
`Subject::has_tv_covariates` being a *structural* `!dose_covariates.is_empty()` test, so
uniform-valued covariates route to the event-driven branch while the documented
`W_DERIVED_CMT_TV_ODE` freezing error is exactly zero. The precondition is asserted in the test.

## Open questions

None. The remaining work is #1166 and then #1167, in that order, cross-referenced on both.


TeunP:
## Self-review at max effort — 10 finder angles + a sweep

Reviewed my own diff adversarially: 10 independent finder angles (5 correctness, 3 cleanup,
altitude, conventions). Several findings are **measured**, not argued — where a number appears
below, it came from toggling one variable and running.

The headline result is uncomfortable and worth stating first: **two angles independently showed
this fix is applied one level too shallow, and that my own test suite does not pin it.**

---

### 1. The fix is one level too shallow — `push_zero_order_break_times` still discards `w_start`

`active_zero_order_inputs` filters on a **two-sided** containment test:

```rust
w_start <= t_start + INFUSION_EPS && w_end >= t_end - INFUSION_EPS
```

but `push_zero_order_break_times` emits only `w_end` — while `w_start` is **already sitting in
the `ZeroOrderWindow` tuple** it iterates (`(cmt, rate, w_start, w_end)`), computed at
`predictions.rs:1017` as exactly the route onset my fix goes elsewhere to recompute.

So the helper whose own doc claims the drift-proofing property —

> Because the break reads `w_end` from the same `ZeroOrderWindow` the per-segment filter uses …
> the segment edge and the containment boundary **can't drift apart**.

— reads *one of the two* boundaries off that tuple. That sentence is what caused #1171.

**Measured.** Replacing the pusher body with
`windows.iter().flat_map(|&(_, _, w_start, w_end)| [w_start, w_end])` **and deleting both
`push_route_lag_break_times` calls this PR adds**: every zero-order test still passes — all 8 in
`lagged_zero_order_break_times.rs` including the joint PK-TTE and `simulate()` arms, both new
Tier-1 tests, and 70 `zero_order`/`route_lag` unit tests. Negative control (neither change): 2
failures, `ipred: t=2.3 gives 0, want 39.99999999999999`. Total cost of the one-liner: one
unrelated assertion at `predictions_tests.rs:4672`.

One line, no signature widening, fixes **all four** builders including any fifth one added later.

### 2. …but it does **not** subsume the fix, and my tests can't tell the difference

`push_route_lag_break_times` also fires for lagged `first_order` / `transit` / `igd` / `weibull`,
which produce no `ZeroOrderWindow` at all. Every fixture I wrote is `zero_order`, where the route
onset coincides *exactly* with `w_start` — so the whole suite passes under the `w_start`-only
variant. **My tests pin a weaker property than the fix.** A later "simplification" to that form
would ship green while the smooth kernels silently lose their onset break on the dense engines.

And the smooth kernels genuinely need it — this is not a nicety. Reading the kernels in
`pk/absorption.rs:693-754`, "smooth" is not a property of the non-zero-order kinds as a class:

- `first_order` is `dose·ka·exp(−ka·tad)` → `dose·ka` as `tad → 0⁺`, and exactly `0` for
  `tad ≤ 0` (guard at `:694`) — a **step discontinuity**;
- `weibull` with β < 1 → ∞, an integrable spike;
- `transit` with n = 0 reduces to the same first-order jump.

Only `igd` and `transit` with n > 0 approach zero smoothly. `sens/ode_provider.rs:4879-4890` also
emits `K_ROUTE_ONSET` for **every** lagged kind, with a comment requiring it stay bit-identical to
`push_route_lag_break_times` (#859) — so narrowing the push to `ZeroOrder` would break the
analytic-vs-FD parity too.

**Resolution: take both.** The `w_start` push makes the silent-zero class unreachable from any
builder; the route-lag push stays unconditional for the kink and for sens parity; and a lagged
`first_order` case gets added to the Tier-1 builder test so the second half is actually pinned.

### 3. Measured: a route onset can double-apply a *different* dose

Break dedup is `1e-15` absolute (`:5530`, `:5825`); the dose-application match is `1e-10`
(`:5559`, `:5892`); the empty-segment skip uses the *same* `1e-15` as the dedup — so a surviving
near-duplicate is guaranteed to be processed as a real segment. No second line of defence.

Measured at `efbd3532`: `d/dt(depot) = first_order(ka=KA, lag=LAG) - KA*depot`, `ALAG1 = 0.3`,
`LAG = 7.9`; dose A 100 mg CMT=1 at t=0 (feeds the forcing), dose B 100 mg CMT=2 at t=8.2. The
onset is `(0.0 + 0.3) + 7.9 = 8.200000000000001` — **1.78e-15** from dose B's own break, too far
to dedup, well inside the 1e-10 match. `u[cmt] += dose.amt * f` runs twice:

```
central(8.2001):  199.998   (LAG = 7.9,  onset collides)
                  100.120   (LAG = 7.85, onset clear)
```

on all three of `ode_dense_solve_states`, `ode_predictions_with_states`, `ode_predictions`. An
infusion at that time is pushed into `active_infusions` twice by the same path.

**Root cause is pre-existing** — `collect_dose_break_times` could already emit such a break on
`main`, so `ode_predictions` has this today. But this PR is what lets the three dense/simulate
engines emit one. Note `RESET_MATCH_TOL`'s doc (`:31-40`) explicitly rests on "break times are
never within 1e-12 of one another"; a two-addition derived onset is exactly what breaks that.
**Filing separately — not fixing here**, since the fix (dedup at the event-match tolerance) is a
behaviour change to every engine and deserves its own anchor.

### 4. Also pre-existing, also found by this pass: `gated_infusions` misses a double-count guard

`active_infusions:932` suppresses an infusion whose mass is already delivered through the
convolved `R_in`:

```rust
if input_rate.iter().any(|f| f.cmt == d.cmt_idx()) { return None; }
```

`gated_infusions:1205` — its twin, used by exactly the three engines this PR touches — has no such
filter. A `RATE>0` dose into a `first_order`/`transit`/`igd` absorption compartment (#719 gap 2, a
supported feature) is therefore injected as `+rate_eff` **on top of** the convolved delivery:
~2× the dose in sdtab, the TTE hazard, `[derived]` and `simulate()`, while the objective path sees
1×. Same KEEP-IN-SYNC drift shape as #1171, on the same three engines. **Filing separately.**

---

### Fixed in this PR

| | finding |
|---|---|
| ✅ | `push_zero_order_break_times` now emits `w_start` as well as `w_end` — root cause, all four builders |
| ✅ | Corrected the two doc comments that state the invariant (`active_zero_order_inputs`, `push_zero_order_break_times`) — the `w_end`-only sentence is what caused this bug |
| ✅ | Added a lagged **`first_order`** case to the Tier-1 builder test, so the route-onset push is pinned for a kind with no `ZeroOrderWindow` |
| ✅ | Narrowed my over-claim that `build_segment_break_times` "must agree with `collect_dose_break_times`" — they legitimately differ by `reset_times` + terminal, and the fixture has no resets |
| ✅ | Tier-1 `check` closure zipped without a length assert — vacuous on a short/empty vector, which is the #1171 signature |
| ✅ | POSTHOC anchor tolerance `3e-2` → realised error is **4.736e-5**; my stated rationale ("different η̂") is empirically false — ferx's η̂ matches `lagged_zo.phi` to ~6 s.f. |
| ✅ | `[derived]` AUC tolerance `5e-4` → realised **6.500e-9** |
| ✅ | `>= 100` row guards → exact `== 120` / `== 132` (and the message misnamed 132 as "12 × 10") |
| ✅ | `anchor()` used a bare relative path; all four sibling anchor suites use `env!("CARGO_MANIFEST_DIR")` |
| ✅ | Positional `theta[0]` / `sr.eta[0]` in the mass-balance reference → read by name |
| ✅ | Tier-1 oracle hardcoded `[0.0, 12.0]`, a second copy of the fixture's dose times |
| ✅ | `docs/model-file/absorption.qmd` still said ferx "breaks the integration timeline at the window end" — the half-mechanism this fixes |
| ✅ | CHANGELOG now says the fix covers every lagged route, not only `zero_order` |
| ✅ | Dead `[simulation] horizon` block and a no-op `s.obs_times = …` reassignment |

### Filed separately, not fixed here

- **Double dose application** (§3) — pre-existing, affects the objective path too; needs its own anchor.
- **`gated_infusions` missing the `R_in` suppression** (§4) — pre-existing, 2× dose on the gated engines.
- **NaN route lag → `partial_cmp().unwrap()` panic.** Newly reachable for smooth kernels; the
  event-driven sibling builder at `:4973` uses `unwrap_or(Ordering::Equal)`, so the two are
  inconsistent. Narrow (needs a NaN covariate per #1111), but it turns a per-subject soft failure
  into an aborted run.
- **Negative route lag** sorts below `subject_integration_start`; `validate` permits it warn-only.
  Shifts CHZ accumulation in `ode_solve_until_chz_threshold`.
- **Builder delegation.** `ode_predictions_with_states`' inline builder and
  `build_segment_break_times` are byte-identical dose loops, both supersets of
  `collect_dose_break_times`. Three copies is what #1171 and #798 both were. A mechanical
  refactor, kept out of a behaviour-change PR on purpose.

### Checked and explicitly **not** findings

- **The break and `w_start` are bit-identical, not merely close.** `:1017` and `:1169` compute
  `(d.time + lag) + route_lag` with identical operands and association, both reading `route_lag`
  off the same forcing, so the `INFUSION_EPS` containment test cannot miss by an ULP.
- **No bit-level regression for existing models.** The helper short-circuits on
  `.filter(|f| f.lag_slot.is_some())`, so any model without a `fn(..., lag=…)` gets an unchanged
  break list. Grepped `to_bits()` / tight-`max_relative` assertions across `tests/` — none
  exercises a route lag.
- **Cost is negligible.** Zero added work without a route lag (`input_rate` is empty for most ODE
  models). With one: +50% segments for a lagged zero-order route, ≈ +100% for a smooth one, at
  ~2–4 µs per segment — a few seconds across a whole joint PK-TTE fit, in exchange for the
  compartment not being identically `0.0`.
- **Duplicate `DV` column in the `.tab` header** (`… MDV DV PRED …`) — only `PRED` and `IPRED` are
  ever requested, both unique, and the predicate is exact `==`. Latent trap only for a future
  caller asking for `"DV"`.
- **The `Dual2`-vs-FD parity rule does not bite.** The diff changes likelihood *values* via
  `ode_cumhaz_hazard` and the Markov endpoint, but both route their gradients to FD off the same
  `f64` predictor (`inner_optimizer.rs:1889`, `markov/endpoint.rs:151`), and
  `build_segment_break_times` is not a `*_g<T: PkNum>` kernel. `sens/ode_provider.rs` already had
  the route-onset break and is untouched.
- **Tolerances that remain** are measured, not guessed: with_states vs objective 2.0e-10 / 3.1e-10;
  `ode_dense_solve_states` vs objective 2.4e-10; route-lag vs comp-lag 8.6e-12; TV vs plain
  bit-identical; `TABLE_TOL` worst 2.839e-5 against 1e-4.

TeunP:
## Review fixes applied — `de651d36`

Eleven of the fifteen findings are fixed in this PR; four pre-existing defects are filed
separately. Verified on a quiesced tree (the earlier numbers were taken while mutation testing
was live in the worktree, so they are re-run here from scratch).

### The two that mattered

**1. The fix was one level too shallow — now fixed at the helper.**
`push_zero_order_break_times` emits **both** edges:

```rust
break_times.extend(
    windows.iter().flat_map(|&(_, _, w_start, w_end)| [w_start, w_end]),
);
```

`w_start` was already in the tuple, computed by the identical expression the route-onset push
recomputes. Every builder that calls this helper — the four that exist and any fifth added later
— now gets correct zero-order segmentation from that one call, with no signature change. Both
doc comments that stated the false `w_end`-only invariant are corrected; that sentence is what
caused #1171.

**The route-onset push stays, unconditional.** It is *not* subsumed: it also fires for lagged
`first_order`/`transit`/`igd`/`weibull`, which produce no `ZeroOrderWindow`. Those are not
"smooth" as a class — `first_order` is `dose·ka·exp(−ka·tad)` with a hard `0` for `tad ≤ 0`, a
step; `weibull` (β < 1) diverges; `transit` (n = 0) reduces to the same jump. And
`sens/ode_provider.rs` emits `K_ROUTE_ONSET` for every lagged kind, documented as bit-identical
to this break (#859), so a kind filter would break the analytic-vs-FD parity.

**2. The tests did not pin the fix — now they do.**
Every fixture was `zero_order`, where the route onset coincides *exactly* with `w_start`. Measured:
deleting both `push_route_lag_break_times` calls and emitting `[w_start, w_end]` instead left the
**entire suite green**. New Tier-1 test `route_lagged_first_order_onset_reaches_both_dense_builders`
drives a lagged `first_order` route (no `ZeroOrderWindow` at all) through both dense builders,
with a **live compartment lag** so the `lag_cmt + lag_route` sum is non-degenerate.

Mutation matrix for the new guard — it is the *only* test that catches either mutation:

| mutation | this test | everything else |
|---|---|---|
| both `push_route_lag_break_times` calls deleted | **FAIL** | pass |
| `dose_lagtimes` replaced by `&[]` at both call sites | **FAIL** | pass |

### Tolerances — two were orders of magnitude loose, one with a rationale that was false

| | was | realised | now |
|---|---|---|---|
| POSTHOC sdtab anchor | 3e-2 | **4.736e-5** | 1e-3 |
| `[derived]` grid AUC | 5e-4 | **6.500e-9** | 1e-6 |

The POSTHOC comment blamed "the two inner optimizers stop at slightly different η̂". That is
wrong: ferx's η̂ matches `results/lagged_zo.phi` to ~6 significant figures (ID 1: 0.019148458 vs
`1.91485E-02`). Both docs now carry the measurement instead of a guess.

### Also fixed

- Tier-1 `check` closure zipped with no length assert — vacuous on a short/empty vector, which is
  the #1171 signature itself.
- `>= 100` row guards → exact `== 120` / `== 132`; the message misnamed 132 as the "12 × 10 record set".
- `anchor()` used a bare relative path → `env!("CARGO_MANIFEST_DIR")`, matching all four sibling anchor suites.
- CL/V read by PK slot instead of positional `theta[0]` / `sr.eta[0]`.
- The Tier-1 oracle derived its dose times from the fixture instead of hardcoding `[0.0, 12.0]`.
- `!has_resets()` added to the TV-covariate isolation precondition (the routing predicate has three disjuncts).
- Narrowed my own over-claim that `build_segment_break_times` "must agree with `collect_dose_break_times`" — it also seeds `subject_integration_start`, pushes `subject.reset_times` and appends `terminal`, and the doc now warns the reset push is load-bearing (#1133).
- `docs/model-file/absorption.qmd` still said ferx breaks "at the window end" — the half-mechanism this corrects.
- CHANGELOG restated at the helper level, and now says the onset break is restored for smooth lagged routes too.
- Dead `[simulation] horizon` block and a no-op `obs_times` reassignment removed.

### Filed separately — pre-existing, out of scope here

| | |
|---|---|
| #1186 | **Break dedup (1e-15) vs dose match (1e-10) double-applies a dose.** Measured: `central(8.2001)` = 199.998 against 100.120. `ode_predictions` already hits this on `main`; this PR extends exposure to the three dense engines. The fix changes behaviour on every engine and needs its own anchor. |
| #1187 | **`gated_infusions` misses the `R_in` suppression `active_infusions` has** — a `RATE>0` dose into a built-in absorption compartment is delivered at ~2× on the gated engines. |
| #1188 | **`adaptive_frozen_replay_tv` is a *fifth* break-time builder with neither pusher** — so the adaptive family's own frozen-schedule replay oracle reads drug-free for any `zero_order` model. Corrected count for the record: five builders, not four. |
| #1189 | NaN route lag panics `partial_cmp().unwrap()`; a negative one sorts below `subject_integration_start`. |

### Verification (quiesced tree, all agents stopped)

- **Every zero-order-touching NONMEM anchor, run with `slow-tests`** — `parallel_mixed_nonmem_anchor`,
  `dose_form_lag_nonmem_anchor`, `lagged_zero_order_nonmem_anchor`, `zero_order_absorption`,
  `parallel_mixed_absorption`, `per_route_lag`: **63 passed, 0 failed, 0 ignored**. That
  includes `mixed_zero_first`'s OFV anchor against `#OBJV = −688.019`, which rides the
  **objective path** — the measurement that confirms the helper change is a no-op there rather
  than my assuming it.
- `cargo test --features ci,survival --no-fail-fast` → **164 binaries, 4750 passed, 0 failed, 169 ignored.**

One note for the record: an earlier full run reported a single lib failure. It happened while
mutation testing had a half-applied edit in the shared worktree; the clean re-run above is green,
so it was an artifact, not a defect.

### On NONMEM coverage for the smooth-kernel half — deliberately not added

A NONMEM anchor for `first_order(ka, lag=L)` on the states engines would be **vacuous**. Smooth
kernels are evaluated pointwise in the RHS and re-read their own lag each evaluation, so a missing
onset break costs them step *placement*, not mass — measured at `3.288e-10 → 3.285e-10`, far below
`$TABLE`'s five-significant-digit print precision. Such an anchor could not fail on this defect in
either direction, which is the same vacuity trap that made this PR anchor `PRED` rather than the
OFV. The sound oracles for that half are the break-list assertion above (mutation-proved) and the
existing `K_ROUTE_ONSET` bit-identity contract.

PR #1191, opened 2026-09-01
fix(docs-lint): a callout title is not a section boundary [closes #1190]

### LINKED ISSUE #1190: bug(docs-lint): Quarto callout titles are counted as section boundaries, so R1/R3/R4 under-report
`docs-lint` (`crates/docs-lint/src/parse.rs`) has no notion of Quarto callout divs. Every `#`-line inside a `::: {.callout-*}` block is parsed as an ordinary heading, so it becomes a section boundary, claims an id, and participates in R2's level check.

Pandoc/Quarto does not agree. The **first** heading inside a callout div is consumed as the callout's *title*: it is removed from the document flow, rendered as the callout header, and generates **no `<section>` and no anchor id**. Later headings inside the same div stay real sections.

The result is a linter that reports the corpus as healthier than it is.

## Impact

Measured on `main` @ `d936353` against the anchors Quarto actually rendered to `gh-pages` @ `a4a6cdb` (86 pages, 961 H2+ sections):

- **37 callout titles across 15 pages** are counted as section boundaries that do not exist on the site.
- **1 heading** is the opposite case — genuinely nested deeper inside a callout and a real anchored section on the site: `docs/estimation/saem.qmd:211`, `### Damping the numerical M-step (\`mstep_damping\`) {#what-saem-does-about-it-now}`. It is *not* the first heading in its div, so it must keep counting.

Per rule:

- **R1 (section size)** — a phantom boundary chops a long section into pieces, so oversized sections read as compliant. Worst case today: `::: {.callout-important}` opens at `docs/estimation/saem.qmd:151` and does not close until line 299. On the rendered page that is a single Important box holding **7,989 characters** of body text, with the whole `### Damping the numerical M-step` subsection buried inside it. `docs-lint` currently sees it as two much smaller sections, one of them the callout title `## A theta with no ETA at all is worse than slow — it can drift` (line 152).
- **R3 (duplicate ids)** — a callout title's slug is registered in `Doc.ids`, so it can collide with a real heading and produce a phantom duplicate.
- **R4 (link resolution)** — the same slug makes `#anchor` targets resolve against ids the site never emits. No dead link on `main` today (220 internal anchored links, zero dead), but the check is unsound.
- **R2 (level skips)** — callout titles are counted as parents/children they are not.

## Fix

In `parse_str`, track fenced-div nesting (`:::` open/close, already partly needed) and **skip a heading that is the first heading inside a `.callout-*` div** — do not emit a `Heading`, do not register its id. Every later heading in that div stays a real heading.

Two details worth pinning:

- The rule is *first heading in a callout div*, not *any heading in a callout div*. `saem.qmd:211` is the case that discriminates the two — a naive "drop all headings inside callouts" would wrongly delete a real anchored section.
- `:::` div tracking must stay code-fence aware, the same way heading parsing already is.

## Baseline

`docs/.lint-baseline` will shift when this lands: sections currently split by a phantom boundary merge and grow, which the shrink-only ratchet rejects. Regenerate it in the same PR (`cargo run -p docs-lint -- --update-baseline`) and call out in the PR description which entries grew *because the measurement got correct*, not because a page got worse. `docs/estimation/saem.qmd::what-saem-does-about-it-now` (currently `R1 5802`) is directly affected.

## Tests

- Tier 1 in `crates/docs-lint`: a callout whose first heading is dropped; a callout with a second heading that is kept; a callout title whose slug must **not** appear in `Doc.ids`; a `:::` sequence inside a code fence that must not open a div.
- A regression fixture built from the `saem.qmd` shape (unclosed-looking long `.callout-important` containing a real `###` with an explicit `{#id}`).

roninsightrx:
## Why

`docs-lint` had no notion of Quarto callout divs, so every `#`-line inside a `::: {.callout-*}` block was parsed as an ordinary heading.

Pandoc does not agree. The **first block** inside a callout, when it is a heading, is consumed as the callout's *title*: it is lifted into the callout header and the rendered page gets **no `<section>` and no anchor id** from it. Later headings inside the same div stay real sections.

All 37 callout titles across 15 pages were counted as headings. The linter therefore reported the corpus as healthier than it is:

- **R1** — a phantom boundary chops a long section into pieces, so oversized sections read as compliant. Three were hidden outright, and `ode-models.qmd::which-regime-am-i-in` was baselined at 5,756 characters when it is in fact **7,331**. The worst shape is `docs/estimation/saem.qmd:151`, a `.callout-important` that does not close until line 299 — a single Important box holding 7,989 characters of body text, which the linter saw as two much smaller sections.
- **R3 / R4** — the title's slug was registered in `Doc.ids`, so it could collide with a real heading, and `#anchor` targets resolved against ids the site never emits. No dead link on `main` today, but the check was unsound.
- **R2** — callout titles counted as parents and children they are not.

## What changed

`parse_str` now tracks fenced divs (`:::` open/close, code-fence aware) and skips a heading that is a callout's **first block** — no `Heading`, no id assigned.

**First *block*, not any heading in a callout.** `docs/estimation/saem.qmd:211` is a real, explicitly anchored `###` nested deeper inside that long `.callout-important`; dropping every heading in a callout would delete a section the site really renders. It is the single case in the corpus that discriminates the two rules, and it has a regression test.

Three things fell out of validating the fix against a real render:

- **`slug.rs`** treated `<op>` inside a code span as raw inline HTML, truncating `` ## Rules — `when signal <op> <value> : <action>` `` to `#rules-when-signal` where the site says `#rules-when-signal-op-value-action`. Pandoc parses no inline markup inside code. This was a live R4 soundness hole, independent of callouts, and it is what kept the render oracle red after the callout fix.
- **`computed_ids_match_the_rendered_site`** keyed on `data-anchor-id`, which Quarto only emits down to h3 — so every deeper section read as a mismatch — and it only checked computed ⊆ rendered. That direction alone can be satisfied **by dropping headings**, which is exactly the failure mode a naive fix for this issue would have. It now reads `<section id="…" class="level*">` and checks both directions.
- **`tools/render-docs.sh`**, because producing that render turned out not to be possible with the obvious command — see below.

### The four sections the blind spot was hiding are split, not baselined

R1 could not see these while a callout title counted as a boundary. With the parser fixed they are real violations, so they are subdivided — nothing deleted, and four bold run-ins become the headings they already read as:

| Section | Was | Now split into |
|---|---|---|
| `ode-models.qmd::which-regime-am-i-in` | 7,331 | Where the step counts show up · Bounding what a stalled segment costs · Measured: an accuracy-limited fit · Every feature works with every method |
| `absorption.qmd::analytic-closed-form-…` | 6,389 | Scope (first version) · The flip-flop regime |
| `adaptive-dosing.qmd::keys` | 5,250 | Writing an `observe` expression (and the table's footnote moves up to sit with the table it annotates) |
| `vi.qmd::how-σ-is-updated` | 5,190 | Why the default is 32 draws · If a VI fit lands short |

`docs/.lint-baseline` therefore ends up **one entry shorter than before this PR** (9 → 8), not four longer: `which-regime-am-i-in` was already baselined and is now genuinely fixed.

### `quarto render docs` renders nothing from a `.claude/` worktree

Quarto's project input discovery skips any path containing a **hidden** (dot-prefixed) directory component. A worktree under `.claude/worktrees/<name>/` — which `CLAUDE.md` asks for on any branch that is not `main` — is exactly that, so `quarto render docs` there finds *zero* inputs: it writes `robots.txt` and `sitemap.xml`, renders not one page, prints no warning and **exits 0**. Nothing tells you the render did not happen.

Confirmed to be the dot and not `.gitignore`: in a minimal tree, the render stays empty with the ancestor `.gitignore` deleted, and succeeds as soon as the ancestor directory is renamed from `.claude` to a non-dot name.

`tools/render-docs.sh` stages `docs/` outside the dot directory when it has to, and fails loudly on a zero-page render. `docs/development/sdlc.qmd` and the oracle's own doc comment now point at it.

## Alternatives considered

**Skip every heading inside a callout.** Simpler, and it passes the forward-only oracle — by deleting `saem.qmd:211`, a section the site anchors and links resolve to. This is what motivated adding the reverse direction to the oracle rather than trusting a green test.

**Baseline the four newly-visible sections instead of splitting them.** That is what the first commit did, and it is the wrong default: the baseline is for debt that predates the linter, and these were only ever invisible because of the bug this PR fixes.

**Leave the slug bug for a follow-up issue.** Rejected: the strengthened oracle is the evidence for this PR, and shipping it knowingly red is worse than a second, separately-described fix in the same change. The two are independent and separately tested.

## Cross-repo dependency

## Breaking changes
- [x] None

`docs-lint` depends on nothing and nothing depends on it; no `ferx-core` API is touched.

## Numerical validation

- [x] Not applicable (docs tooling only)

The equivalent anchor here is the render oracle. Against a full render of all 86 pages (1,043 rendered sections), `DOCS_LINT_VERIFY_RENDER=1 cargo test -p docs-lint -- --ignored`:

| | mismatches |
|---|---|
| `main` | 39 |
| + callout fix | 2 |
| + slug fix | **0**, in both directions |
| + the four splits | **0**, in both directions |

The 37 that disappear are exactly the callout titles. Of the 2 survivors after the callout fix, one was the h4 gap in the oracle's own marker and one was the slug bug — both fixed here.

## Performance

`docs-lint` still depends on nothing and compiles in seconds.

## Tests

Nine new tests in `parse.rs`: the title is dropped and claims no id; the bare `::: callout-note` spelling; a later heading in the same callout survives with its explicit id; a heading after prose is not a title; a heading after a code block is not a title; a plain `.panel-tabset` div keeps its heading; a callout only swallows its own first heading; a `:::` run inside a code fence opens no div; a nested div does not consume its parent's title slot. One in `slug.rs` for the code-span case, plus the negative (`<br>` outside a span is still a tag).

`cargo test -p docs-lint`: 72 pass. `tools/preflight.sh fmt` and `tools/preflight.sh docs` green.

- [x] Not applicable (internal tooling)

## Docs
- [x] `docs/` updated — `docs/development/docs-lint.qmd` gains a **Callout titles are not sections** subsection with the worked `:::` example, plus a row in the id-generation trap table for the code-span case; `docs/development/sdlc.qmd` points at `tools/render-docs.sh`; four pages are subdivided as above
- [x] `CHANGELOG.md` `[Unreleased]` → `Fixed` (the parser and slug bugs) and `Changed` (the splits, and the render script)
- [x] `docs/_site/` not committed

## Checklist
- [x] `tools/preflight.sh fmt` and `tools/preflight.sh docs` green
- [x] `Dual2` path untouched
- [x] No version bump needed

The `check` / `clippy` / `rustdoc` / public-API groups were not run: nothing outside `crates/docs-lint/`, `docs/` and `tools/` changed, and `docs-lint` depends on no crate in the workspace.

## Reviewer hints

The judgement call is **first block vs. any heading**, and `docs/estimation/saem.qmd:211` is the case that decides it — worth reading `a_later_heading_inside_a_callout_is_a_real_section` alongside the `saem.qmd` source.

The reverse direction added to `computed_ids_match_the_rendered_site` is the part that makes the oracle able to fail on an over-eager fix; the forward direction alone cannot.

The four doc splits are the bulk of the diff and are pure subdivision — no sentence is removed or rewritten except four bold run-ins becoming headings, and one footnote moving up next to its table.

## Open questions

None outstanding.

https://claude.ai/code/session_01HHu4jJGn5rF2xtVvfspwde


roninsightrx:
## Code review

Ran `cargo test -p docs-lint` and `cargo run -p docs-lint -- --check` (both green), and cross-checked the callout premise against a real `quarto render` rather than against the parser's own model.

The core change is right: a callout's first heading is lifted into the callout header and yields no section, so counting it as a heading was chopping the callout body into chunks R1 never measured whole (#1190). The div stack, the `mark_content` placement (fence-open, div-open, non-empty content), the nested-div / plain-div / late-heading cases and the `:::`-inside-a-fence case all read correctly and are covered by the new tests. The corpus reverse-direction check and `section_ids` match the real render markup, `tools/render-docs.sh` is safe (docs/ is self-contained — no `{{< include >}}` or out-of-tree assets — and it renders before removing the old `_site`), and the baseline entry removal plus the four doc splits pass R1–R4.

Four findings.

### 1. The swallowed callout title still consumes the identifier slot (medium)

`crates/docs-lint/src/parse.rs:299` — the swallow branch `continue`s before `ids.assign(raw)`, so the title claims no id. But pandoc assigns auto-identifiers in the *reader*; Quarto's callout Lua filter drops the header only afterwards, so the slot is already spent.

Verified with a real render:

```qmd
::: {.callout-note}
## Dup
:::

## Dup
```

renders the real section as `id="dup-1"` — the parser computes `dup`.

Consequence: R4 accepts a link to the dead `#dup` and reports the live `#dup-1` as broken, and R3 stops seeing the collision. No page in `docs/` hits this today (checked all 86), so it is latent rather than live.

Fix: call `ids.assign(raw)` in the swallow branch and discard the result, so the slot is consumed. Note this must not insert into `all_ids` — the render confirms no `id="dup"` anchor exists — and it does not create a false R3, since `check_duplicate_ids` walks `doc.headings`, which the title never enters.

### 2. The test pins the behaviour the renderer contradicts (medium)

`crates/docs-lint/src/parse.rs:531` — `a_callout_title_claims_no_id_so_a_real_heading_keeps_the_undisambiguated_one` asserts `id == "dup"` / `base_id == "dup"` for exactly the case Quarto renders as `dup-1`. It needs inverting along with the fix above, and its comment ("if the title consumed the slug, the real section below would render as `#dup-1`") is the correct description of what actually happens.

### 3. `div_marker` has no indent guard (low)

`crates/docs-lint/src/parse.rs:261` — the heading path guards indented code blocks (`line.len() - trimmed.len() < 4`); the div path does not. A `:::` line inside a 4-space-indented code block pushes a phantom `DivFrame`, and a `.callout-*` one swallows the next heading. No occurrence in the current corpus, but the asymmetry is the bug.

### 4. `in_code` toggles per backtick, not per run (low)

`crates/docs-lint/src/slug.rs:88` — a double-backtick span (`` ``a <op> b`` ``) toggles `in_code` twice and ends up *outside* code, so the `<op>`-eaten-as-a-tag bug this hunk fixes comes back for that spelling. Confirmed against a render: `` ## Use ``a <op> b`` here `` is `#use-a-op-b-here`; the current code computes `use-a-b-here`. A run-length-aware toggle (open on a run of N backticks, close on a matching run of N) closes it.

https://claude.ai/code/session_01HHu4jJGn5rF2xtVvfspwde

Issue #707, opened 2026-07-06
Update default threads logic: cap default threads to 8


roninsightrx:
For most users/scenarios there's no real benefit in distributing over more cores, and efficiency is lost. Also on MacOS, not all cores are equal. Of course, when `threads` is specified explicitly, then that should override the default.

So logic should be:
- get number of available cores for machine / system (n)
- use n-1 as default cores
- if (n-1) > 8, then use 8 as threads

PR #1192, opened 2026-09-02
feat(api): expose PoolPlan and quiet fits so tools can run many fits [closes #1115]

### LINKED ISSUE #1115: feat(api): expose thread-pool control and quiet fits so tools can run many fits [#1114]
## Summary

Two `ferx-core` API gaps that block any tool running many fits (bootstrap, SCM,
model search). Both are core-side changes with their own tests — no
`ferx-tools` code in this issue.

P2 of `plans/ferx-tools-split.md`. Depends on the workspace split (P1).

## G1 — thread-pool control

`src/api/pool.rs` is entirely `pub(crate)`: `build_fit_pool`,
`default_fit_pool`, `fit_thread_pool_builder`, `FIT_RAYON_STACK_SIZE`. A tool
running N fits concurrently has no supported way to say *"replicates in
parallel, one worker per inner fit"*.

Two hazards, both needing explicit handling rather than a bare `pub`:

**Nested rayon.** `fit()` already `par_iter`s over subjects. A naive outer
`par_iter` over replicates nests pools on top of that, so the two levels compete
for the same workers. For bootstrap, replicate-level parallelism is the better
axis (embarrassingly parallel, no per-subject sync), so the tool needs a
supported way to request a **single-threaded inner fit** and size the outer pool
itself.

**Stack size — the subtle one.** `FIT_RAYON_STACK_SIZE` is 32 MiB because wide
ODE+IOV analytic-gradient models overflow the default rayon worker stack. An
outer pool built from a bare `ThreadPoolBuilder::new()` inherits the default
stack and will overflow on exactly the models that need it most — and *only*
under a tool, so no existing test catches it. Whatever surface ships must make
the correct stack size the default, not something the caller remembers.

### Deliverable

A small `pub` surface — e.g. a `PoolPlan { replicates: usize, threads_per_fit: usize }`
plus a nested-pool builder — that:

- splits a total thread budget across the two levels;
- builds the outer pool with `FIT_RAYON_STACK_SIZE`;
- gives the inner `fit()` a documented single-threaded mode.

Subject to A2/A6 of the boundary rules: the new items land in
`api/ferx-core-public-api.txt` in the same PR, and must be usable from `ferx-r`
as an ordinary external crate (A4), not just from inside the workspace.

### Tests (Tier 1)

- [ ] Outer pool inherits `FIT_RAYON_STACK_SIZE` — assert on the builder, since
      the overflow itself only reproduces on large models
- [ ] A thread budget splits as documented across the two levels
- [ ] A fit requested single-threaded does not spawn subject-level workers
- [ ] Results are identical across thread counts (`1` vs `4`) for a small fit

## G2 — quiet inner fits

200 fits × per-iteration console output is unusable. Core already routes
non-fatal messages into `FitResult.warnings` rather than stderr (the documented
warning convention), so this should be a verbosity knob on the CLI printer
rather than new plumbing — **but confirm no estimator prints directly to
stderr/stdout before relying on that.** If any does, route it through
`FitResult.warnings` as part of this issue.

Desired end state: a tool can run N fits and emit one line per completion:

```
rep  37/200  OFV 1183.42  converged   12.3s   (ok 36, fail 1)
```

### Tests

- [ ] A fit run in quiet mode produces no direct stdout/stderr output; all
      non-fatal messages are reachable via `FitResult.warnings`

## Not in scope

- Any `ferx-tools` code — that starts with bootstrap (P3)
- G3 (model mutation / emit), which blocks SCM and model search but **not**
  bootstrap

roninsightrx:
## Why

Two `ferx-core` API gaps block any tool that runs many fits (bootstrap, SCM,
model search) — #1115, P2 of the ferx-tools split.

**G1 — thread-pool control.** All of `src/api/pool.rs` was `pub(crate)`, so a
tool had no supported way to say *"replicates in parallel, one worker per inner
fit"*. Two hazards made a bare `pub` the wrong answer:

* **Nested rayon.** `fit()` already `par_iter`s over subjects, so a naive outer
  `par_iter` over replicates nests pools and the two levels compete for the same
  workers.
* **Stack size.** `FIT_RAYON_STACK_SIZE` is 32 MiB because wide ODE+IOV analytic
  gradients overflow the platform-default 2 MiB Rayon worker stack. An outer pool
  built from a bare `ThreadPoolBuilder::new()` inherits the default and faults on
  exactly the models that need the bigger stack — and only under a tool, where
  nothing else looks. The correct stack has to be the default, not something the
  caller remembers.

**G2 — quiet inner fits.** 200 fits × per-iteration console output is unusable,
and `FitOptions::verbose` defaults to `true`.

## What changed

**`PoolPlan`** (`src/api/pool.rs`), a small `pub` surface:

```rust
let plan = PoolPlan::from_budget(8, 200);   // -> 8 replicates x 1 thread/fit
let mut options = FitOptions::default().quiet();
plan.apply_to(&mut options);                // sets FitOptions::threads
plan.install(|| { /* par_iter over replicates */ })?;
```

The budget is spent on the outer level first, since replicate-level parallelism
is embarrassingly parallel while the per-subject level synchronises every
iteration:

```text
replicates      = min(total_threads, n_units)     (at least 1)
threads_per_fit = total_threads / replicates      (at least 1, floored)
```

`from_budget(8, 200)` → `8 × 1`, `from_budget(8, 2)` → `2 × 4`,
`from_budget(8, 3)` → `3 × 2` (floors rather than oversubscribing).
`total_threads = 0` means the engine default (cores − 1, capped at 8, #707).
Both levels are always ≥ 1, so a plan can never silently mean "let Rayon
decide" — the ambiguity `configure_global_thread_pool` already rejects.

`install()` builds the outer pool through the same `fit_thread_pool_builder()`
the engine uses, so the 32 MiB stack is the default rather than a thing to
remember. `FIT_RAYON_STACK_SIZE` is also now `pub` (it was already reachable in
spirit — `sens/provider_tests.rs` builds its own pool with it).

The plan deliberately does **not** put a `rayon::ThreadPool` in its signature —
`install()` takes the closure instead — so rayon does not enter the public API
and the surface stays usable from ferx-r as an ordinary external crate (A4).

**Quiet fits.** The issue asked to confirm the warning convention before relying
on it. Surveyed every `print!`/`eprintln!` in `src/` outside tests and `src/bin`:
all estimator output is already behind `options.verbose` or an env flag
(`FERX_TIME_GRADIENTS`, `FERX_PROFILE`, `FERX_SENS_CHECK`), and `io/output.rs` /
`api/run.rs` are the CLI printer and the file-based entry points, which a tool
does not call. The **only** two ungated prints on the `fit()` path were the
optimizer-trace lines; the open failure now goes into `FitResult.warnings` (the
documented convention) and the success line is `verbose`-gated. Added
`FitOptions::quiet()` since `verbose` defaults to `true`.

## Alternatives considered

* **`pub` the existing helpers as-is.** Rejected: it hands the caller
  `ThreadPoolBuilder` and the stack-size footgun with it.
* **Returning `rayon::ThreadPool` from a builder method.** Rejected: it pins
  ferx-core's rayon version into the public API for no gain over `install()`.
* **Public fields on `PoolPlan`** (as sketched in the issue). Rejected: `0` on
  either field re-introduces "Rayon picks"; the constructors clamp instead.

## Cross-repo dependency

Additive public API only; nothing ferx-r consumes changed shape. ferx-r picks
this up whenever its `src/rust/Cargo.lock` is next bumped.

## Breaking changes
- [x] None (public Rust API is additive; baseline regenerated with
      `tools/update-public-api.sh` in this PR)

Added items and their caller: `PoolPlan` + its methods and
`FIT_RAYON_STACK_SIZE` are what the P3 bootstrap tool in `ferx-tools` needs to
size and stack its replicate pool; `FitOptions::quiet()` is what it needs to run
200 fits without flooding the console. None is reachable through the existing
surface — `FitOptions::threads` sizes only the *inner* level and says nothing
about the outer pool or its stack.

## Numerical validation
- [x] Not applicable — this is thread plumbing and a verbosity flag; it produces
      no numerical results, so there is no NONMEM run to anchor against. The
      relevant guarantee is instead pinned directly: a fit is **bit-identical**
      across 1 and 4 threads (theta, omega, sigma, OFV compared on `to_bits()`),
      re-pinning #703's deterministic per-subject reduction from the tool's
      angle.

## Performance
- [x] No significant impact expected — no production path changed behaviour; the
      new type is opt-in.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib --features ci`:
      3854 passed, 0 failed)

Tier 1, `src/api/tests/pool_plan_tests.rs` and
`src/api/tests/quiet_fit_tests.rs`, covering the issue's checklist:

* the budget split, including the flooring and clamping edges;
* the outer pool's stack, both **declared** (`worker_stack_size()`) and
  **observed** — a 4 MiB frame touched inside `install()` overflows a
  default-stack worker. Verified the probe discriminates by swapping the builder
  for a bare `ThreadPoolBuilder::new()`:
  `thread '<unknown>' has overflowed its stack / fatal runtime error: stack
  overflow, aborting`;
* a single-threaded plan gives `n_threads_used == 1`;
* a fit is bit-identical across 1 and 4 threads;
* a quiet fit writes nothing to the real stdout/stderr.

The quiet test is worth a look: libtest's own capture intercepts `eprintln!`
from the test thread, so an in-process assertion would pass no matter what the
engine wrote. The test re-executes the test binary as a child process (guarded
by `FERX_QUIET_FIT_CHILD`, `--exact --nocapture`) and inspects the child's real
file descriptors. Verified against a planted `eprintln!("BOOM from an
estimator")` in `fit_inner`, which turns it red with the message quoted.

- [x] Example added inline (Rust API, not `.ferx` syntax) — see
      `docs/api/fitting.qmd` § *Running many fits*, and the runnable doctest on
      `PoolPlan`.

## Docs
- [x] `docs/` updated — `docs/api/fitting.qmd` gains a *Running many fits*
      section (budget split, the stack hazard, quiet mode). Page already in
      `docs/_quarto.yml`; `cargo run -p docs-lint -- --check` green.
- [x] `CHANGELOG.md` `[Unreleased]` → `Added` entry

## Checklist
- [x] No `Dual2`/`PkNum` code touched
- [ ] `Cargo.toml` version bump — not needed (additive)

## Docs & examples
- [x] No new `.ferx` DSL syntax, `[fit_options]` key, example model, or data
      file — nothing for ferx-site or ferx-book, and no example-execution step.

## Reviewer hints

* **`PoolPlan::from_budget`'s split rule** is the design decision worth arguing
  with: outer-level-first, floored, no oversubscription. Everything else follows
  from it.
* **`install()` must keep going through `fit_thread_pool_builder()`** — that one
  line is the whole stack guarantee, and
  `the_outer_pool_really_has_more_than_the_default_stack` is the test that
  notices if it stops.
* The G2 change is three lines in `src/api/fit.rs`; the survey behind it is in
  the commit message.
* `src/api/pool.rs` is otherwise untouched — the diff is an append plus one
  `pub(crate)` → `pub`.

## Open questions

* `apply_to(&self, options: &mut FitOptions)` vs. a consuming
  `options.with_plan(plan)` — happy to flip if the latter reads better at the
  ferx-tools call site in P3.
* Should `ferx-tools` get a thin re-export of `PoolPlan` so a bootstrap caller
  imports one crate? Left out here since this issue explicitly excludes
  `ferx-tools` code.

https://claude.ai/code/session_014VMb63MwV4MCwxWdH2kbd4


roninsightrx:

Reviewed the branch diff. One medium finding, six low, one nit. **All of them are addressed in 3685300e** (pushed to this branch); this comment records what was found and what changed.

### 🔴 medium — a pinned thread count did not bind a multi-start fit

`src/api/pool.rs:259` — `PoolPlan::apply_to()` sets `FitOptions::threads`, but `fit()` honored that field only on the single-start fast path (`src/api/fit.rs:649`). The multi-start arm (`n_starts > 1`) ran its start fan-out on the shared `default_fit_pool()` regardless, with a comment saying that narrowing to `threads` "would cap the combined fan-out below the available cores".

That is exactly what a `PoolPlan` caller wants. A bootstrap tool that sets `n_starts = 3` and `PoolPlan::from_budget(8, 200)` gets an 8-wide outer replicate pool where every replicate then fans out over the 8-wide *global* pool again — 64-way oversubscription on 8 cores, the hazard `PoolPlan` exists to prevent.

Fixed by honoring `threads` in the multi-start arm too: the start fan-out and the per-subject loops underneath it now share the one pinned pool, so a plan's budget is not multiplied by `n_starts` behind the caller's back. `a_pinned_plan_also_caps_a_multi_start_fit` pins it. Behavioural change, so it has a `### Changed` changelog entry: a multi-start fit that pins `threads` below the core count is correspondingly slower than before; unpinned multi-start fits are unchanged.

### 🟡 low

- **`pool.rs:150`** — `from_budget(0, …)` called `default_thread_count()` directly, bypassing the `GLOBAL_THREADS_EXPLICIT` branch that `default_fit_pool()` uses. The doc promised "the same … that an unpinned `fit()` uses", but after a CLI `configure_global_thread_pool(16)` an unpinned fit ran on 16 workers while `from_budget(0, 200)` assumed 8. Both now resolve through one `effective_default_threads()` helper, and the test asserts against that helper rather than re-deriving the cap.
- **`fit.rs:930`** — the trace path was announced only under `verbose`, and the filename embeds pid + unix timestamp. A tool setting `optimizer_trace: true` alongside `quiet()` had no way to learn where the trace went — not printed, not on `warnings`, not on `FitResult`. The success line now goes onto `FitResult::warnings` like the failure line already did; a verbose run still sees both, via the `pre_run_warnings` dump.
- **`pool.rs:171`** — `total_threads()` was an unchecked `replicates * threads_per_fit`; `PoolPlan::new(usize::MAX, usize::MAX)` panics in debug / wraps in release. Now `saturating_mul` — free, and keeps a public accessor total.
- **`pool.rs:272`** — with `threads_per_fit == 1` each inner `fit()` still builds a real 1-worker pool and `install`s into it from an outer worker, which then blocks: `replicates = 8` means 16 live OS threads, 16 × 32 MiB of reserved stack, while `total_threads()` reports 8. Documented on `apply_to`/`total_threads` rather than short-circuited — running the inner fit inline would let the per-subject `par_iter` fan out over the *outer* pool's workers, reintroducing the nesting the pin removes.
- **`tests/pool_plan_tests.rs:92`** — the 4 MiB stack probe fails by aborting the process (a stack overflow does not unwind), so a regression kills the whole test binary and every other test reads as "not run". Kept — there is no way to observe a real stack size without risking the overflow, and the declared-size assertion above it is the graceful half of the pair — but it now says so.
- **`tests/quiet_fit_tests.rs:95`** — the child run set `optimizer_trace: true` and left `/tmp/ferx_trace_<pid>_<ts>.csv` behind on every `cargo test`. The child now reads the path back off `warnings` (which also exercises the fix above) and removes the file.

### 🟢 nit

- **`tests/quiet_fit_tests.rs:110`** — `is_harness_line` counts blank lines as harness output and the stderr check was `stderr.trim().is_empty()`, so a bare `eprintln!()` leaking from the engine would not have turned the test red. The stderr assertion is now strictly-empty; blank-line filtering stays on stdout, where libtest genuinely pads its report, with a comment saying why the two sides differ.

### Verified and clean

- The G2 survey holds: every `eprintln!`/`println!` on the `fit()` path outside `io/output.rs`, `api/run.rs`, `src/bin` and the env-flag blocks (`FERX_TIME_GRADIENTS`, `FERX_PROFILE`) is behind `options.verbose` — checkpoint banners, SIR, covariance progress, stage headers included.
- `pre_run_warnings` lands before the verbose dump and is drained into `accumulated_warnings`, so both trace lines reach `FitResult::warnings`.
- `from_budget` arithmetic has no division-by-zero or zero-width path.
- The public-API baseline covers every added item.

`tools/preflight.sh` is green (fmt / check / clippy / rustdoc / docs / public-API), and the pool, quiet-fit and multi-start unit tests pass.

PR #1193, opened 2026-09-02
fix(tools): reject high-cardinality labels and constant EBEs in GAM screening [#1114]

roninsightrx:
## Why

Follow-up to #1169, which merged before this round of review findings could land on its branch. Three defects, all reproduced against the merged code before being fixed.

**1. A near-saturated categorical wins the ranking outright.** #1169's guard refused only an *exactly* saturated design (`n <= p`, residual df < 1), which leaves the whole band below it open. With 100 unrelated ETA values and a label carrying 99 levels:

```
covariate   delta_aic     R²        form
SITE          474.499   0.998775   Categorical
WT             -0.656   0.013354   Linear
```

A pure-noise SITE identifier ranks first, far above the covariate carrying real signal. For a label with `k` levels over `n` subjects and no relation to the ETA, `E[RSS/RSS₀] ≈ (n−k)/(n−1)`, so

```
ΔAIC ≈ n·ln((n−1)/(n−k)) − 2(k−1)
```

which is comfortably negative for small `k`, crosses zero near `k ≈ 0.8n`, and reaches +474 at `k = n−1, n = 100`. The fit's freedom to interpolate outruns the `2p` penalty long before the design saturates.

**2. A constant ETA scores `NaN`.** Twenty EBEs of `0.0` make RSS zero for every model *including the null*, so both AICs are `−∞` and `aic_null_local - best_aic` is `−∞ − (−∞)` = `NaN`. `sort_by` falls back to `Equal` for a NaN, so the score was not merely wrong but unordered, and it sat in the ranking with no warning:

```
aic_null = -inf
X      dAIC=NaN aic=-inf R2=0
warnings: []
```

Reachable for a fixed random effect, or one the data never identified.

**3. `--workspace` widened what the slow job compiles, but not what triggers it.** #1169 made `slow-tests.yml` run `cargo test --workspace` so the member's Tier-3 test would execute, but left the `push: paths:` filter and its guard hard-coded to root `src/` and `tests/`. A push to `main` touching only `crates/ferx-tools/**` silently skipped the two-hour suite, and `tests/slow_test_path_filter.rs` passed while 32 member files went uncovered.

## What changed

**Cardinality guard (`gam.rs`).** A categorical must leave at least as many residual degrees of freedom as it spends parameters — `n >= 2p`, so roughly `k <= n/2`, where the expression above is still comfortably negative (≈ −30 at `k = n/2, n = 100`). Nothing useful is lost below the threshold: a label in the discarded band scores negative wherever plain AIC still works, so it would rank last either way.

The level count also moved ahead of the allocation. `categorical_design` took the raw column and built an `n × (levels−1)` matrix, so a one-level-per-subject label allocated O(n²) *before* anything rejected it. `categorical_levels` is now split out, and the design takes the levels the caller already computed.

**Constant-EBE guard (`gam.rs`).** The ETA is refused before any model is fitted, with a warning. A second guard covers what the whole-column check cannot see: EBEs that vary across the population but are constant over the subset carrying one particular covariate.

**Slow-job triggers (`slow-tests.yml`, `tests/slow_test_path_filter.rs`).** The filter gained `crates/**`. The guard now derives the members from the root manifest's `[workspace] members` and requires each one's `src/`, `tests/` and `Cargo.toml`, and its fixture-literal harvest reads member test sources so a path opened from `crates/*/tests/` contributes its root like a root-package one.

The command also gained `--exclude docs-lint`. That crate is a structural linter over `docs/**` with its own seconds-long CI job and no dependencies; running it inside the slow suite would make `docs/**` an input of that job — a docs typo triggering the two-hour fit run — and would couple the linter to the `ferx-core` compile path, which is what keeping it dependency-free was for (#969). The guard parses the `--exclude` list out of the workflow rather than hard-coding it, so an exclusion added or dropped moves the expected filter with it in the same commit.

## Alternatives considered

**AICc instead of a threshold.** `AIC + 2p(p+1)/(n−p−1)` handles finding 1 smoothly and is the textbook answer for small `n/p`. Rejected: it shifts every ΔAIC by ~0.1 (for `p=2, n=60` the correction is 0.21 against the null's 0.07), which breaks the R / xpose4 parity this module is validated against to 1e-4. The threshold is a guard on admissibility, not a change to the reported score.

**Adding `docs/**` to the slow-test filter** instead of excluding `docs-lint` from the command. Strictly correct once `--workspace` runs that crate, but it makes every docs edit trigger a 120-minute fit suite.

## Cross-repo dependency

## Breaking changes

- [x] None

`gam_screen_raw` and `gam_screen` keep their signatures. Some covariates that were previously *scored* are now *skipped with a warning* — that is the fix.

## Numerical validation

- [x] Compared against: R (`splines::ns()` + `lm.fit()` under the same `n·log(RSS/n) + 2p`), unchanged from #1169

`xpose4_anchor_delta_aic_matches_reference` passes untouched at 1e-4 — all six ΔAIC values are identical, which is the point of fixing this with a guard rather than a penalty change:

```
       ETA_CL                    ETA_V
WT     1.882144    (unchanged)   -1.655242   (unchanged)
CRCL  -0.525482    (unchanged)   -1.491354   (unchanged)
SEX   -1.383530    (unchanged)   22.295612   (unchanged)
```

The anchor's `SEX` is a 2-level factor over 60 subjects — 2 parameters, 58 residual df — so it sits nowhere near the new threshold.

## Performance

Strictly less work: the O(n²) dummy matrix for a rejected label is no longer built, and rejected covariates skip their fits.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 45 gam tests, up from 39)

| test | pins |
|---|---|
| `near_saturated_categorical_is_skipped_not_ranked_first` | the `n−1` levels over `n` case from the report, asserting the fixture really is near-saturated and not saturated |
| `the_cardinality_guard_sits_exactly_at_two_parameters_per_subject` | the boundary — 10 levels over 20 subjects admitted, 11 refused |
| `a_categorical_with_room_to_spare_is_still_scored` | the guard does not swallow an ordinary 3-level factor |
| `constant_etas_produce_no_scores_and_a_warning` | finding 2 |
| `constant_etas_over_one_covariates_subset_skip_only_that_covariate` | the subset route to the same NaN |
| `no_score_is_ever_nan` | the ranking contract itself, across all three covariate shapes |

For finding 3, `slow_test_filter_covers_every_input_the_heavy_fits_read` was negative-controlled in both directions — a guard that cannot fail is the failure mode that file exists to prevent:

```
drop 'crates/**' from paths:     FAILED — "crates/ — 32 file(s), e.g. crates/docs-lint/Cargo.toml"
drop '--exclude docs-lint':      FAILED — "docs/ — 99 file(s), e.g. docs/.gitignore"
both restored:                   ok. 8 passed
```

- [x] Not applicable (internal API — no new `.ferx` DSL syntax or fit options)

## Docs

- [x] `docs/` updated for user-visible changes — `docs/tools/gam-screening.qmd` gains a **High-cardinality labels** section with the ΔAIC expression and why AICc was not used; the skip table and the constant-EBE refusal are documented
- [x] New pages linked in `docs/_quarto.yml` (the page itself landed in #1169)
- [x] `CHANGELOG.md` `[Unreleased]` entry updated

## Checklist

- [x] Functions on the `Dual2` sensitivity path stay differentiable — untouched (`ferx-tools` only)

### ferx-site
- [x] No new `.ferx` DSL syntax, `[fit_options]` key, example file or data file

### ferx-book

### Example execution
- [x] No example execution step needed (no user-visible `.ferx` change)

## Reviewer hints

The judgement call is the `n >= 2p` threshold, at `gam.rs`'s categorical branch. It is deliberately conservative — it discards a band (`n/2 < k < 0.8n`) where plain AIC still behaves — and the comment carries the arithmetic for why the looser `k < n` rule is not enough. If you would rather have a different constant, that is the one line to argue about; the boundary test pins whatever it ends up being.

Second-guess candidate: `built_members` in the path-filter guard now parses `--exclude` out of the workflow with a whitespace tokeniser. That is crude, but the two negative controls above show it is load-bearing in both directions rather than decorative.

## Open questions

None.

https://claude.ai/code/session_01F42w6iw2bi2qtn9vW1QFXy

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -54,6 +54,7 @@ on:
       - 'Cargo.lock'                        # a nlopt/nalgebra/rand bump moves fit numbers and RNG streams
       - 'Cargo.toml'                        # the `ci-test` profile + the `slow-tests` / `survival` features
       - 'crates/**'                         # the workspace members — the job runs `--workspace`
+      - '!crates/docs-lint/**'              # ...except the one the command --excludes; see the run: step
       - 'data/**'                           # fixture CSVs the fits read by literal path
       - 'examples/**'                       # `.ferx` models the anchors load (e.g. the adaptive vanco suite)
       - 'nonmem_anchor/**'                  # committed NONMEM control streams, data and `.phi` references
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -53,6 +53,7 @@ on:
       - '.github/workflows/slow-tests.yml'  # the feature set / cargo command this job runs
       - 'Cargo.lock'                        # a nlopt/nalgebra/rand bump moves fit numbers and RNG streams
       - 'Cargo.toml'                        # the `ci-test` profile + the `slow-tests` / `survival` features
+      - 'crates/**'                         # the workspace members — the job runs `--workspace`
       - 'data/**'                           # fixture CSVs the fits read by literal path
       - 'examples/**'                       # `.ferx` models the anchors load (e.g. the adaptive vanco suite)
       - 'nonmem_anchor/**'                  # committed NONMEM control streams, data and `.phi` references
@@ -105,7 +106,14 @@ jobs:
         # `slow-tests` is declared by both `ferx-core` and `ferx-tools` — an
         # unqualified name would resolve to the root package's and leave the
         # member's gate closed.
+        #
+        # `--exclude docs-lint`: that crate is a structural linter over `docs/**` with its
+        # own seconds-long CI job, and it depends on nothing. Running it here would put
+        # `docs/**` among this job's inputs, so a docs typo would trigger the two-hour fit
+        # suite — and it would couple the linter to the `ferx-core` compile path, which is
+        # exactly what keeping it dependency-free was for (#969). The guard reads this
+        # `--exclude` list, so adding or dropping one moves the path filter with it.
         run: |
-          cargo test --workspace --no-default-features \
+          cargo test --workspace --exclude docs-lint --no-default-features \
             --features ferx-core/ci,ferx-core/survival,ferx-core/slow-tests,ferx-tools/slow-tests \
             --profile ci-test --no-fail-fast
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,9 +83,10 @@ section of the SDLC for the versioning policy).
 
   A covariate that cannot be screened is now reported in `GamResult::warnings` rather than
   dropped in silence — being skipped and being screened-but-unimportant are indistinguishable
-  in a ranking. Skips cover: a constant covariate, a single-level or fully saturated
-  categorical, a singular design, a column whose length does not match the subject count, and
-  fewer than three usable subjects. Shrinkage above 30% warns, and so does an ETA whose
+  in a ranking. Skips cover: a constant covariate, a single-level categorical, a categorical
+  spending more than half the subjects as parameters, a singular design, a column whose
+  length does not match the subject count, and fewer than three usable subjects. An ETA whose
+  EBEs are constant is refused outright. Shrinkage above 30% warns, and so does an ETA whose
   shrinkage the fit did not report at all. `gam_screen_raw()` panics on a length mismatch
   instead of truncating to the shortest input.
 - **`PoolPlan` and `FitOptions::quiet()`: the two knobs a tool needs to run many fits (#1115).**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,11 +84,11 @@ section of the SDLC for the versioning policy).
   A covariate that cannot be screened is now reported in `GamResult::warnings` rather than
   dropped in silence — being skipped and being screened-but-unimportant are indistinguishable
   in a ranking. Skips cover: a constant covariate, a single-level categorical, a categorical
-  spending more than half the subjects as parameters, a singular design, a column whose
-  length does not match the subject count, and fewer than three usable subjects. An ETA whose
-  EBEs are constant is refused outright. Shrinkage above 30% warns, and so does an ETA whose
-  shrinkage the fit did not report at all. `gam_screen_raw()` panics on a length mismatch
-  instead of truncating to the shortest input.
+  or spline form spending more than half the subjects as parameters, a singular design, a
+  column whose length does not match the subject count, and fewer than three usable subjects.
+  An ETA whose EBEs are constant, or not all finite, is refused outright. Shrinkage above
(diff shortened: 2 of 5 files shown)


roninsightrx:
All five confirmed and fixed in `ed693756`. Two of them were sharper than they first looked.

**1 — one-sided cardinality guard.** Right, and the reachability argument is the important half: `spline_df` is public and the *default* `[2, 3]` gets there, since 5 usable subjects with `df = 3` spends 4 parameters and leaves 1. The spline loop now applies the same `n >= 2·(df+1)`. Two tests: the `n = 5` default-options case, and a caller-supplied `df = 15` over 20 subjects. The docs section is retitled "Designs that spend too much" since the rule is one rule now — the mechanism does not care whether the columns came from one-hot levels or a basis.

**2 — `crates/**` re-couples the excluded crate.** Confirmed, and your reading of why nothing caught it is exactly right: coverage is `required ⊆ filter` and the dead-glob test is satisfied by the other members. Added `!crates/docs-lint/**` plus the missing direction as `the_filter_does_not_select_an_excluded_member`.

That test then failed for a reason neither of us had named: `required_inputs` was pulling `crates/docs-lint/**` back in regardless, because the harvested runtime roots are top-level (`crates`), so the whole subtree landed in the required set and the `--exclude` was being undone by the derivation meant to respect it. Excluded members are now subtracted from `required_inputs`. Without that, the negation would have turned the coverage test red instead.

**3 — `excluded_packages` tokenizes comments.** Confirmed, including the "only survives by luck" part — the backticks in my comment made the token `` `--exclude ``. Comments are now stripped quote-aware before tokenizing (the `paths:` list is full of quoted globs, so a `#` inside quotes must not open a comment). Pinned by a unit test on synthetic YAML carrying both a prose mention and a commented-out run line.

**4 — `workspace_members` bracket termination.** Fixed to break on any line *containing* `]`, after taking that line's literals. Split the parse out as `parse_workspace_members(&str)` so the regression tests all three formattings without a fixture repo.

**5 — "constant" for a non-finite column.** Confirmed on both sites. Only NaN is filtered out of the EBE column, so ±inf reaches the variance check, and `!aic_null_local.is_finite()` is likewise true for `RSS = inf`. Both split now, with the non-finite case named as a fit blow-up rather than a covariate property.

Every fix negative-controlled — reverting each in turn fails exactly its own test and nothing else:

```
spline guard -> n <= df+1        2 FAILED
merge the non-finite branch      1 FAILED
drop '!crates/docs-lint/**'      1 FAILED
un-strip comments                1 FAILED
bracket close -> starts_with     1 FAILED
```

48 gam unit tests, 11 path-filter tests, R anchor unchanged, `tools/preflight.sh` green, and the slow integration test still passes under `--features ferx-tools/slow-tests`.

Issue #719, opened 2026-07-07
feat(pk): built-in absorption — support SS dosing & infusion (ODE + analytic) and IOV (analytic transit)


TeunP:
## Summary

The built-in absorption models reject several dose / variability features today.
#588 turned these into loud errors on every entry point (`fit` / `ferx check` /
`predict` / `simulate`), but they are still *unsupported*. This tracks implementing
them. Three gaps, grouped in one issue because they share the same two guard sites
and the same closed-form-vs-ODE split.

Feature support today:

| Feature | ODE forcing (`transit()`/`igd()`/`weibull()`/`zero_order()`/`first_order()` in `[odes]`) | Analytic closed-form (`pk one_cpt_transit` / `two_cpt_transit`) |
|---|---|---|
| F, lag | ✅ | ✅ |
| TV covariates | ✅ | ✅ (routes to ODE-equivalent, #663) |
| IOV (κ) | ✅ (#663) | ❌ **rejected** — this issue |
| SS dosing | ❌ **rejected** — this issue | ❌ **rejected** — this issue |
| infusion into absorption cmt | ❌ **rejected** — this issue | ❌ **rejected** — this issue |

### 1. Steady-state dosing (SS=1) — ODE **and** analytic
- **ODE input-rate forcing**: rejected with `E_ABSORPTION_SS` (`src/api.rs` ~936,
  `check_absorption_dosing`) — the SS equilibration applies the dose as a bolus
  pulse, not as `R_in(tad)` over the cycle.
- **Analytic transit**: rejected — *"does not support steady-state (SS) doses yet"*
  (`src/api.rs` ~1242, `check_transit_support`).
- Needs the periodic steady state of a **time-varying** input rate: for the ODE,
  equilibrate over the II cycle with the forcing active; for the analytic form, a
  periodic-sum (geometric superposition of the per-dose tilted response over the II
  grid).

### 2. Infusion (RATE>0) into the absorption compartment — ODE **and** analytic
- **ODE**: rejected with `E_ABSORPTION_RATE` (`src/api.rs` ~961) — the dose mass is
  delivered through `R_in` computed from the dose *amount*, so an infusion rate on
  the same record double-counts.
- **Analytic**: rejected — *"does not support infusion doses"* (`src/api.rs` ~1250)
  — the closed form absorbs an instantaneous bolus through the chain.
- Needs a defined **semantics** for infusion + built-in absorption first (e.g. the
[...]

PR #1195, opened 2026-09-02
fix(ode): suppress the plain +rate for an infusion into an absorption cmt [closes #1187]

### LINKED ISSUE #1187: bug(ode): gated_infusions double-counts an infusion into a built-in absorption compartment
## Summary

`gated_infusions` is missing the input-rate double-count suppression that its twin
`active_infusions` carries, so an infusion into a **built-in absorption compartment** is injected
twice on every engine that uses the gated form.

Found while self-reviewing #1174. Pre-existing on `main`, independent of #1171.

## The asymmetry

`active_infusions` (`src/ode/predictions.rs:932`) — used by the prediction engines, which split
the timeline at every infusion edge:

```rust
// Infusion into a built-in absorption compartment (#719 gap 2): the dose is a
// zero-order source *feeding the kernel*, delivered through the convolved input rate
// `R_in_inf` (`add_prepared_input_rate_forcing`), NOT injected directly. Suppress the
// plain `+rate` here so the mass is not double-counted.
if input_rate.iter().any(|f| f.cmt == d.cmt_idx()) {
    return None;
}
```

`gated_infusions` (`src/ode/predictions.rs:1205`) — the `InfusionInput::Gated` counterpart used
by the dense/simulate engines, which do *not* split at infusion edges — has **no such filter**.
It drops `CMT=0` and out-of-range compartments, then emits the tuple unconditionally.

## Consequence

For a `first_order` / `transit` / `igd` / `weibull` model with a `RATE>0` (or `RATE=-2`) dose into
the absorption compartment — #719 gap 2, a supported and NONMEM-anchored feature — the mass is
delivered **twice**:

1. once through `add_prepared_input_rate_forcing`'s convolved `R_in_inf` kernel, and
2. again as a plain `+rate_eff` over the window, injected by `wrap_rhs_with_forcings` from the
   gated tuple.

Affected surfaces (everything reached by `InfusionInput::Gated`):

- sdtab `IPRED` and compartment states (`ode_predictions_with_states`)
- sdtab states for TV-covariate / reset subjects, and `[derived]` grid integrals
  (`ode_dense_solve_states`)
- the **joint PK-TTE** cumulative hazard (`survival::ode_cumhaz_hazard`) and the Markov/CTMM
  endpoint NLL — i.e. two objectives
- `simulate()` event times (`ode_solve_until_chz_threshold`)

The objective's Gaussian path uses `active_infusions` and sees the dose once, so the two halves
disagree — the same asymmetry that hid #1171.

## Why no test caught it

The existing coverage for #719 gap 2 — `predictions_tests.rs::infusion_into_first_order_absorption_matches_subdose_train`
and `tests/infusion_absorption_nonmem_anchor.rs` — both go through `ode_predictions`, i.e. the
`active_infusions` side that *has* the guard. Nothing exercises the gated engines with an
infusion into an absorption compartment.

## Suggested fix

Add the same suppression to `gated_infusions`. It needs `&[InputRateForcing]` threaded in (it
currently takes `active`, `doses`, `dose_f_bio`, `n_states`); both call sites (`predictions.rs`
~`:5647` and ~`:5930`) already hold `ode`.

Regression test wants to be a states-path twin of the existing anchor: assert
`compute_predictions_with_states` agrees with `compute_predictions_with_tv` at the same η for a
`first_order` model with `RATE>0` into the depot. That comparison fails today at ~2× and is the
same shape as the guard #1174 added for the zero-order case.

## Meta

This is the third instance of the same class — a helper and its "keep in sync" twin drifting
(#798: three break-walking loops; #1171: two break-time builders; this: two infusion resolvers).
Worth considering whether the two `InfusionInput` arms can share their filtering.

TeunP:
## Why

`gated_infusions` lacked the input-rate suppression its twin `active_infusions` carries. Both feed the same `wrap_rhs_with_forcings` seam, which adds the resolved `+rate` *alongside* `add_prepared_input_rate_forcing`'s convolved `R_in_inf` — so a `RATE>0` / `RATE=-2` dose into a built-in absorption compartment (#719 gap 2, a supported and NONMEM-anchored feature) was delivered **twice** on every engine resolving through the gated form.

The second copy is not merely a doubled dose: the plain `+rate` lands in the target compartment *directly*, bypassing the absorption kernel. On a mass-balance accumulator the error is exactly `2·F·amt`; in a concentration readout it is worst early — **214×** at t=0.5 on a transit model, settling near 1.86×. Against NONMEM, the joint PK-TTE cumulative hazard read **31×** high.

Six entry points, three symptoms — two of them *fitted objectives*, not diagnostics:

| | surface | symptom |
|---|---|---|
| 1 | `pk/mod.rs` | ODE IPRED **and** states |
| 2 | `survival/mod.rs` | joint PK-TTE cumulative hazard — **objective** |
| 3 | `markov/endpoint.rs` | CTMM NLL — **objective** |
| 4 | `output_columns.rs` | `[derived]` grid integrals |
| 5 | `ode_predictions_event_driven_with_states` | IPRED correct, states wrong → inconsistent sdtab row |
| 6 | `survival/mod.rs` | `simulate()` event times |

`predict()` was correct throughout — it resolves through `active_infusions`.

## What changed

`gated_infusions` gains `active_infusions`' guard. The parameter is the **runtime forcing slice**, not `ode.input_rate`: `ode/ekf.rs` passes `&[]` deliberately (that path never applies the forcing), so a predicate reading the *spec* would suppress a rate nothing replaces and delete mass instead of double-counting it. Both call sites already hold `ode`.

The membership rule is now written once — `forcing_consumes_cmt(&[InputRateForcing], cmt)` — with `input_rate_consumes_cmt` as its spec-reading wrapper and both infusion resolvers calling it. Behaviour-identical (the dropped `!is_empty() &&` is redundant with `any()` on an empty slice) and `pub(crate)`, so the public-API baseline is unchanged. #1196 tracks the remaining `Spanning`/`Gated` duplication.

## Alternatives considered

Merging the two resolvers outright. Not viable and not proposed: `Spanning` returns `(cmt, rate)` because those callers split the timeline at every infusion edge, `Gated` returns `(cmt, rate, t_start, t_end)` because those callers gate by time inside the RHS. Only the predicate is shareable, and it is extracted here. See #1196.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation
- [x] Compared against: **NONMEM 7.6.0**, two anchors

**Existing anchor**, `data/inf_first_order.csv`, driven through the gated engines for the first time. Deleting the guard: `1.52394` vs NONMEM `0.30468` at t=1 (**5.00×**) — while the two pre-existing tests in that same file stay green, which is exactly why the original anchor was blind.

**New run**, `nonmem_anchor/pktte_inf.{ctl,csv}` + `results/pktte_inf.{lst,tab}` — ADVAN13, explicit DEPOT/CENTRAL/CHZ, `MAXEVAL=0 POSTHOC`, every `THETA FIX` and `$OMEGA 0 FIX` so both tools evaluate at η = 0. `A(3)` is compared rather than the OFV: a likelihood model carries different additive constants in the two tools; the cumulative hazard is constant-free.

`FORMAT=s1PE15.8` on `$TABLE` is load-bearing. The default prints 5 significant figures, so the reference banks as `1.2217e-1` carrying ±5e-6 of **its own rounding** — 4.1e-5 relative, a floor under any tolerance the anchor could assert regardless of how close ferx is. At 9 figures the reference contributes ~1e-9 and the bound measures ferx.

```
# NONMEM 7.6.0 reference, CHZ at the event time t = 1.86  (9 s.f.)
1.22168018e-1

# ferx before — dedicated arm (ode_cumhaz_hazard -> ode_dense_solve_states)
3.796844      (31x; the hazard is H0*exp(BETA*C), so a doubled input rate compounds)
# ferx before — shared arm (#570)                    already correct
# ferx after  — dedicated  1.221680182e-1            rel 1.41e-9
# ferx after  — shared     1.221676202e-1            rel 3.26e-6
```

The two arms are **not** equally accurate — a 2300× spread, from integrating different timelines (the share covers the union of TTE record times in one pass; the dedicated call solves to `[1.86]` alone). Neither is a defect and both sit far inside the arms-agree tolerance, but a single bound covering both would let a real regression in the tight arm pass unnoticed, so the bounds are per-arm (`1e-8` / `1e-5`), each set just above its measured value.

## Performance
- [x] No significant impact expected — one `.iter().any()` per infusion per segment, on a list already being walked.

## Tests

Nine tests:

| test | oracle |
|---|---|
| `gated_infusions_suppresses_infusion_into_input_rate_cmt` | the helper — both branches, plus the empty-slice case |
| `infusion_into_input_rate_cmt_is_mass_exact_on_every_engine` | **closed-form mass balance**, no integrator in the oracle; 5 arms |
| `infusion_into_absorption_matches_subdose_train_on_gated_engines` | sub-dose train; `first_order` + `transit`; RATE=−2 at **F ≠ 1**; **two overlapping-window arms** |
| `infusion_into_non_input_rate_cmt_is_untouched_on_gated_engines` | the control |
| `gated_engines_match_nonmem_infusion_into_first_order_absorption` | committed NONMEM PRED |
| `joint_pktte_hazard_arms_agree_for_infusion_into_absorption` | arm A ≡ arm B at the **objective**; without the guard, TTE NLL 2.323 vs 35.255 |
| `joint_pktte_cumulative_hazard_matches_nonmem_for_infusion_into_absorption` | both arms vs NONMEM — two engines agreeing is not evidence either is right |
| `simulated_event_time_is_consistent_with_the_guarded_hazard` | the crossing fed back through the **guarded** `ode_predictions` |
| `ctmm_nll_for_infusion_into_absorption_matches_a_subdose_train` | the CTMM **objective** vs a sub-dose train |

Two notes on oracle choice, both load-bearing:

- The obvious oracle for `simulate()` — compare the crossing against `ode_dense_solve_states`' CHZ column — is **vacuous**: both are gated engines, so under the defect they inflate together and agree. The reference has to come from the guarded side.
- The sub-dose train is **not** "N ordinary boluses". A bolus into an input-rate compartment is itself suppressed (`apply_prescheduled_boluses_at`) and delivered through the same machinery via `prep.rate`. The train pins `rate_infused` against `rate`, not the kernel. The kernel-independent checks are the mass balance and the NONMEM anchors.

**Mutation matrix — measured, each applied separately:**

| mutation | result |
|---|---|
| (a) guard deleted | **8 of 9 die** (7 in `--lib`, 1 in the Tier-2 anchor binary); the control correctly survives, and no pre-existing test dies |
| (b) guard scoped to `FirstOrder` | the three transit-kernel fixtures die — the fixtures are not all one kernel kind (#1174's trap) |
| (c) guard widened to `!is_empty()` | **2 die**: the control *and* the helper unit test, whose second assertion pins that a CMT-2 infusion survives while the forcing sits on state 0 |

The narrowness of the guard is therefore pinned at two levels, unit and behavioural — better than the single test claimed in an earlier revision of this description.

## Example

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added — it names what each affected user should do: re-fit if the model carries a joint PK-TTE or Markov/CTMM endpoint (those objectives were wrong), regenerate sdtab / states / `[derived]` if it is pure PK (estimates fine, diagnostics were not)
- [x] No page change needed — `docs/model-file/absorption.qmd` already states the plain `+rate` is suppressed. The doc was right; the code was not.

## Checklist
- [x] `tools/preflight.sh` green — fmt, 5 check feature sets, clippy `--all-targets`, rustdoc, docs, public-API baseline
- [x] `Dual2` path untouched — `has_infusion_into_input_rate` already declines this class to FD, which is *why* no Dual2-vs-FD parity test could have caught it: FD differentiates the guarded predictor
- [x] No version bump needed

## Docs & examples
- [x] No example execution step needed — internal fix, no user-visible API change

## Reviewer hints

Focus on the **predicate choice** (`input_rate` slice vs `ode.input_rate`) — the one decision that could be silently wrong, and getting it wrong deletes mass on the EKF path instead of double-counting it. Mutation (c) is what pins it.

Worth knowing: before this PR the full suite was green **both** ways (164 binaries, 4766 passed) — zero of 4766 tests detected a 214× error on a documented feature.

## Open questions

Ordering: **#1187 → #1166 → #1167**. Fixing #1166 first would route time-dependent-hazard joint models back onto the shared arm (correct) and *mask* this without fixing it. Rationale posted on both issues.

---

Verified on `df662a15`: `tools/preflight.sh` exit 0. `cargo test --no-default-features --features ci,survival --no-fail-fast` → 164 binaries, **4774 passed, 0 failed**, 169 ignored. `--features ci,markov,nn --lib`, where CI measures the CTMM surface → **4118 passed, 0 failed**.


TeunP:
The "worth its own issue" in **Alternatives considered** is now filed as #1196.

Scoped there to the **membership predicate** only — not to merging the two resolvers. `Spanning` returns `(cmt, rate)` and `Gated` returns `(cmt, rate, t_start, t_end)` because one set of callers splits the timeline at every infusion edge and the other gates by time inside the RHS; that difference is load-bearing and stays. My "larger blast radius" line above was true of merging the functions, which is not the proposal — extracting the shared predicate is a modest change.

#1196 also carries the trap a careless unification would hit: the predicate must take the **runtime forcing slice**, not `ode.input_rate`, because `ode/ekf.rs` passes `&[]` deliberately. A spec-based predicate there would delete mass rather than double-count it.

The recurrence-preventing part is the test it proposes: assert the two resolvers **agree on membership** for a shared dose list. That property was false before this PR and nothing asserts it today.

TeunP:
## Review — adversarial pass at `4d9a80c0`

**The fix itself is correct and I could not break it.** Ten findings below; none of them
argue the guard is wrong. One is user-facing and I'd fix before merge (#1); four are
coverage gaps of the shape this repo has been bitten by before; the rest are comment /
test hygiene.

### What I verified by running, not by reading

- **Mass-deletion audit.** The guard suppresses a `+rate` that `add_prepared_input_rate_forcing`
  is assumed to replace. I enumerated every path where that assumption could fail and the guard
  would *delete* mass instead of de-duplicating it:
  - `forcing.kind == ZeroOrder` (skipped pointwise) — covered by `active_zero_order_inputs`;
    `zero_order_dur_and_frac_for_dose` opens a window for an infusion dose too, so the mass still
    arrives once. Also rejected upstream (`E_ABSORPTION_RATE_ZERO_ORDER`).
  - SS branch (`d.ss && d.ii > 0.0` → `ss_periodic_forcing`, the *bolus* kernel) — rejected
    upstream, `II`-independently (`E_ABSORPTION_SS_INFUSION`), and the #1121 residual-infusion
    push at `apply_segment_boundary` is `ss_seeded_at_record`-gated, i.e. SS+lag, also rejected.
  - EVID 3/4 reset — `apply_segment_boundary` clears `active_infusions` at the reset and
    `add_prepared_input_rate_forcing` skips `t_eff < reset_floor`; both off, symmetric.
    `ode_predictions_with_states` passes `NEG_INFINITY` but carries a documented no-reset
    precondition.
  - `CMT=0` / `cmt >= n_states` — dropped before the guard either way.
  - EKF — `ekf.rs:263` hand-rolls its own wrapper with **no** input-rate forcing at all, so
    `active_infusions(&[], …)` is right and this function is never reached from there.
- **Mutation (a), re-run independently.** Guard replaced with `if false`: **5 lib tests die**
  (`gated_infusions_suppresses…`, mass-exact, sub-dose train, both `joint_pktte_*`) and **zero
  pre-existing tests die**. Confirms both halves of the PR's claim — the new tests are live, and
  the old suite really was blind to a 214× error.
- **The NONMEM anchor is a genuine equivalence, not an assumption.** Checked
  `DADT(1) = -KA*A(1)` under `RATE=50` against `R_in_inf`: for `t ≤ T`, `KA·A₁(t) = R(1−e^{-KA·t})`
  and `(mass/T)·g(t) = 50·(1−e^{-ka·t})` — identical. Measured agreement **rel 1.62e-5**.
- **`Dual2` untouched** confirmed: `has_infusion_into_input_rate` gates `ode_subject_supported`,
  `ode_tvcov_supported`, `ode_provider.rs:2674` and `inner_optimizer.rs:267`.
- **Arm A ≡ arm B headroom measured** at |Δ| = 3.28e-6 against a 2.32e-4 bound (71×). Not
  fragile — I had this on the candidate list as a default-tolerance twin comparison and dropped it
  after measuring.

---

### 1. `CHANGELOG.md:198` — the entry denies the estimate change it just documented

> "Models without an infusion into an absorption compartment are unaffected, and **no estimate on
> the corrected path changes**."

That contradicts the same entry three lines up, which lists the joint PK-TTE cumulative hazard and
the Markov/CTMM likelihood as *objectives* that were wrong. They were: this PR's own anchor measures
`H(1.86)` at **31×**, and the TTE NLL at **35.255 where the corrected value is 2.323**. The OFV
changes, so every estimate on an affected model changes.

This is the one line an affected user acts on, and it tells them not to re-run. Suggest something
like: *"Fits of joint PK-TTE or Markov/CTMM models with an infusion into an absorption compartment
should be re-run — their objective, and therefore their estimates, were wrong."*

### 2. `predictions.rs:6331` — `simulate()` event times: no regression test

`ode_solve_until_chz_threshold` is surface 6 of the six the description enumerates, and it has its
**own** `wrap_rhs_with_forcings` call site — separate from `ode_dense_solve_states`' at `:6144`.
The fix reaches it only because both share `apply_segment_boundary`, and nothing pins that. Item G
of the anchor plan (degenerate oracle via `ode_solve_until_chz_threshold`) covers exactly this and
was dropped.

### 3. `markov/endpoint.rs:341` — the CTMM NLL is a fitted objective with no test

Both objective-level tests target joint PK-TTE. The Markov scorer reads states from
`ode_dense_solve_states` on its own `grid`; nothing asserts the corrected exposure. Given the PR's
own headline measurement — 0 of 4766 tests saw this class of error — "it shares the chokepoint" is
not coverage.

### 4. `predictions_tests.rs:4635` — the multi-dose arms can't see a trajectory

The only multi-dose arms (`"two infusions"`, `"bolus + infusion"`) run on
`transit_accumulator_spec`, whose RHS is `dy[0] = 0.0`. That state is `∫R_in dt`, so the arm is
**invariant to break-time placement and to when the mass arrives** — it reads 200.0 regardless.
Both trajectory oracles (sub-dose train, NONMEM anchor) are single-dose.

The gated engines use `build_segment_break_times`, a *different* builder from
`collect_dose_break_times` — the exact twin pair that diverged in #1171. Plan item F named this
("multi-dose ADVAN2, 2nd window on residual — guards the gated break-time builder (#1171 class)")
and it shipped as a mass-balance arm on the one fixture that structurally cannot test it. A second
dose landing on a live kernel, compared against the sub-dose train, would close it.

### 5. `predictions.rs:1228` — the doc denies what both call sites do

> "`input_rate` is the **runtime forcing slice**, not `ode.input_rate` … a caller that does not
> apply the input-rate forcing passes `&[]` and keeps its plain `+rate` (`ode/ekf.rs`)."

Both call sites added here (`:5693`, `:5988`) pass literally `&ode.input_rate`, and `ekf.rs:253`
calls `active_infusions`, not this function. So the caller the doc describes does not exist for
`gated_infusions`, and a maintainer looking for it finds nothing.

The contract is right, the phrasing isn't. Something like *"pass the same slice whose forcings you
will actually apply — `prepare_input_rates` builds `prepared` from `ode.input_rate`, so these
callers pass that; a caller that applies no forcing passes `&[]`, as `active_infusions`' EKF caller
does."*

### 6. `predictions_tests.rs:4698` — the sub-dose train's independence is overstated

> "an oracle that shares no convention with `R_in_inf` (it delivers the same mass as N ordinary
> boluses through the ordinary bolus path)"

It doesn't take the ordinary bolus path. A bolus into an input-rate compartment is suppressed at
`predictions.rs:2366` and delivered through `add_prepared_input_rate_forcing`'s
`prep.rate(tad, mass)` — the same `PreparedInputRate` machinery, differing only in `rate` vs
`rate_infused`.

Still the right oracle for the *convolution* — but it is a `rate`-vs-`rate_infused` check, not a
kernel check, and the comment as written invites over-trusting a green run.

### 7. `predictions_tests.rs:4690` — unreachable double-count assert

```rust
assert_relative_eq!(got, expected, max_relative = 5e-3);
assert!(got < 1.5 * expected, "… (double-count?)");
```

The second can never fire; the first already bounds `got ≤ 1.005·expected`. It reads as the test's
double-count detector and is dead. Worse, it invites a future reader to loosen the relative bound
believing the 1.5× guard backstops it.

### 8. `likelihood.rs:3929` — `drug_free` is an exp/ln round trip over hard-coded literals

```rust
let drug_free = -(-0.05_f64).exp().ln() - (0.01_f64).ln();
```

`-(-0.05).exp().ln()` is identically `0.05`. The constant is `H0·t − ln H0`, with `H0 = 0.01` and
`t = 5.0` inlined as bare literals disconnected from the `theta TVH0(0.01, …)` and
`ObsRecord::Event { time: 5.0 }` 40 lines above. Change either and the non-degeneracy guard quietly
checks the wrong baseline — i.e. the vacuous-agreement case it exists to exclude becomes
undetectable. `h0 * t - h0.ln()` from the same literals.

### 9. `likelihood.rs:4035` — the NONMEM reference is quantized at the tolerance

`1.2217e-1` comes from `pktte_inf.tab`, which prints 5 significant figures. Its own rounding is
±5e-6 = **4.1e-5 relative**, against a `rel < 1e-4` bound — 2.4× headroom over the *reference's
printing precision*, not over ferx's error (measured 1.62e-5). The anchor can't discriminate below
~4e-5 and the bound can't be tightened, because `$TABLE` in `pktte_inf.ctl` carries no `FORMAT=`.
A 5e-5 shift from an integrator change sits ambiguously between invisible and red. `FORMAT=s1PE15.8`
on the re-run (or the `.lst`'s own higher-precision value) fixes it permanently.

### 10. `predictions.rs:1264` — third copy of the membership predicate

`input_rate.iter().any(|f| f.cmt == cmt)` joins `input_rate_consumes_cmt` (`:1126`) and
`active_infusions` (`:932`). #1196 correctly defers the `Spanning`/`Gated` unification, but the
*predicate* is a zero-risk extraction that could land here:

```rust
fn forcing_consumes_cmt(input_rate: &[InputRateForcing], cmt: usize) -> bool {
    input_rate.iter().any(|f| f.cmt == cmt)
}
```

called by all three (`input_rate_consumes_cmt` via `ode.input_rate` + `saturating_sub(1)`). The PR's
premise is that this predicate drifting between two copies cost 214×; it lands a third.

---

**Merge posture:** #1 before merge — it is user-facing and it is wrong in the direction that costs
someone a re-run. #2–#4 are the ones I'd actually want as tests rather than as a follow-up issue,
given #1171's precedent; #5–#10 are fine either way.

TeunP:
## Review addendum — verification ledger closed, and an implementation plan

Follow-up to [the review above](https://github.com/FeRx-NLME/ferx-core/pull/1195#issuecomment-5507981024).
That comment was scoped to what I had actually executed, which was `cargo test --lib` plus two
tolerance measurements. Three things in it were **unverified PR claims I had not re-run**, and one
sentence was under-scoped. All closed now, at `4d9a80c0`.

### Verification ledger — everything now run

| | run | result |
|---|---|---|
| Tier-2 anchor binary | `cargo test --features ci,survival --test infusion_absorption_nonmem_anchor` | **3 passed, 0 failed** — including `gated_engines_match_nonmem_infusion_into_first_order_absorption`, which I had never executed |
| Mutation **(a)** guard → `if false`, lib | `cargo test --features ci,survival --lib` | exit 101, **5 die**: `gated_infusions_suppresses…`, `…mass_exact_on_every_engine`, `…subdose_train_on_gated_engines`, `joint_pktte_hazard_arms_agree…`, `joint_pktte_cumulative_hazard_matches_nonmem…` |
| Mutation **(a)**, Tier-2 | same mutation, `--test infusion_absorption_nonmem_anchor` | **1 dies** — `t=1: ode_predictions_with_states 1.52394 vs NONMEM 0.30468 (rel err 4.00e0)`. **5 + 1 = 6 of 7, control survives — PR's claim exact.** |
| Mutation **(b)** guard scoped to `InputRateKind::FirstOrder` | `--lib` | exit 101, **3 die**, all three transit-kernel fixtures. PR's claim ("the transit arms die") holds |
| Mutation **(c)** guard widened to `!input_rate.is_empty()` | `--lib` | exit 101, **2 die** — see correction below |
| `tools/preflight.sh` | full script, after a forced rebuild — see the correction below | **exit 0** — `preflight OK — fmt check clippy rustdoc docs public-api` |
| Full suite | `cargo test --features ci,survival --no-fail-fast`, after a forced rebuild | **164 binaries, 4773 passed, 0 failed**, 169 ignored, `CARGO_EXIT=0` — matches the description's claim exactly |

> **Edited.** When first posted, the preflight row read simply "exit 0" and the full-suite row was
> open. Both are now re-runs against a forced rebuild — the originals were invalid, for the reason
> in the next section. Nothing else in this comment changed.

### Correction to my own method: the first preflight and full-suite runs were stale

The first time I ran them, the results were not real, and the failure mode is worth writing down
because nothing about it looks wrong.

`sed -i.bak file` **renames the original to `file.bak`** — which keeps the original's old mtime —
and writes a *new* `file`. So the revert I used after each mutation, `mv -f file.bak file`, restores
the correct **content** together with an mtime **older than the artifacts built from the mutated
source**. Cargo's fingerprint is mtime-based, so it rebuilt nothing:

```
src/ode/predictions.rs                 mtime 1788344135
target/…/ferx_core-<hash> (test bin)   mtime 1788344969   ← newer ⇒ treated as fresh
```

The first "full suite" therefore re-ran the **mutation (c)** binary and reported exactly the two
tests mutation (c) kills; and the first `preflight OK` was a cached check of mutated source. The
whole time, `git diff` was clean and line 1264 read
`if input_rate.iter().any(|f| f.cmt == cmt) {` — correct. Nothing in the working tree was wrong.

Fixed with `find src crates tests -name '*.rs' -exec touch {} +` before re-running both. **The
mutation measurements (a), (b), (c) above are unaffected** — each `sed -i` writes a file with a
*fresh* mtime, so each mutated build was genuinely compiled; only the reverts were stale.

The re-run lands at **164 binaries / 4773 passed / 0 failed / 169 ignored**, `CARGO_EXIT=0`, which is the description's number to the test. The lib binary alone is `3986 passed; 0 failed; 23 ignored` in 235 s, and the two "failures" the stale run reported are gone — confirming they were the mutation (c) binary and not anything on this branch.

**Scope correction to my earlier comment.** Where it says mutation (a) kills "zero pre-existing
tests", that was `--lib` only. It is now also true of the Tier-2 binary: under mutation (a) the two
pre-existing tests in `infusion_absorption_nonmem_anchor.rs`
(`predict_matches_nonmem_infusion_into_first_order_absorption`, `fit_on_infusion_absorption_runs`)
both stay green, which is exactly the blindness the PR describes.

### Correction: mutation (c) kills **two** tests, not one

The description says:

> | (c) guard widened to `!is_empty()` | the control dies, and only the control |

Measured, it kills two:

```
ode::predictions::tests::gated_infusions_suppresses_infusion_into_input_rate_cmt
ode::predictions::tests::infusion_into_non_input_rate_cmt_is_untouched_on_gated_engines
```

The helper unit test dies on its second assertion — `gated_infusions(&forcing, …)` is pinned to
return `vec![(1usize, 7.0, 1.0, 3.0)]`, i.e. the CMT-2 infusion survives while the forcing sits on
state 0; a widened guard returns `vec![]`.

This is **better than advertised** — the narrowness of the guard is pinned at two levels, unit and
behavioural, not one. But the matrix as written is wrong, and it is the table a reviewer leans on.
Worth correcting in the description (it changes nothing in the code).

---

## Implementation plan

Findings restated as concrete edits, so this can be picked up in a fresh session. Numbering matches
the review above. **#1 is the only one I'd block merge on.**

### #1 — `CHANGELOG.md:198` (blocking)

Current final sentence of the entry:

```
  window, settling near 1.9× afterwards. Models without an infusion into an absorption
  compartment are unaffected, and no estimate on the corrected path changes.
```

`no estimate on the corrected path changes` contradicts the same entry's own list of two *fitted
objectives*. Replace with something that tells affected users to re-run, e.g.:

```
  window, settling near 1.9× afterwards. Models without an infusion into an absorption
  compartment are unaffected. Models with one should be re-fitted if they carry a joint PK-TTE
  or Markov/CTMM endpoint: those objectives were wrong, so their estimates were too. A pure-PK
  fit is unaffected — its objective always ran through `predict()`'s guarded path — but its
  sdtab `IPRED`, compartment states and `[derived]` columns were wrong and should be regenerated.
```

That last clause matters and is missing today: a pure-PK user's *estimates* are fine but their
*diagnostics* were not, and the current text lumps them into "unaffected".

### #2 — regression test for `simulate()` event times

`ode_solve_until_chz_threshold` (`src/ode/predictions.rs:6331`) has its own
`wrap_rhs_with_forcings` call site, separate from `ode_dense_solve_states`' at `:6144`. Nothing
tests the fix through it.

Cheapest sound oracle is the degenerate one from the #391 playbook: with a threshold the hazard
crosses, the returned event time must equal the time at which `ode_dense_solve_states`' CHZ column
crosses the same threshold, to solver tolerance. Reuse the `first_order(ka=KA)` + `H0*exp(BETA*C)`
fixture already in `likelihood.rs:3843` (the arms-agree test), give it `RATE=50`, and assert:

```rust
// ThresholdOutcome::Crossed(t) must land where the dense CHZ column crosses `threshold`.
```

Mutation to verify it: guard → `if false`; the doubled exposure inflates the hazard, so the
crossing moves early and the test must die.

### #3 — regression test for the CTMM/Markov NLL

`src/markov/endpoint.rs:341` reads states from `ode_dense_solve_states` on its own grid. This is a
*fitted objective* that was wrong and now has no test. A Tier-1 test in the markov module: a
2-state CTMM whose transition intensity reads a compartment fed by `first_order()` under a `RATE>0`
dose, asserting the NLL against the same model driven by an explicit sub-dose train (the oracle
`infusion_into_absorption_matches_subdose_train_on_gated_engines` already uses). Needs
`--features markov`.

### #4 — multi-dose **trajectory** arm on the gated engines

The two multi-dose arms in `infusion_into_input_rate_cmt_is_mass_exact_on_every_engine`
(`predictions_tests.rs:4660`, `:4668`) run on `transit_accumulator_spec`, whose RHS is
`dy[0] = 0.0`. That state is `∫R_in dt`, so the arm is invariant to break-time placement and to
*when* the mass arrives — it reads 200.0 either way. Both trajectory oracles are single-dose.

The gated engines build their timeline with `build_segment_break_times`, a different builder from
the guarded engines' `collect_dose_break_times` — the pair that diverged in #1171. Add a two-dose
arm to `infusion_into_absorption_matches_subdose_train_on_gated_engines`
(`predictions_tests.rs:4712`): second dose landing while the first kernel is still delivering,
sub-dose train covering both windows, `rel < 1e-3` as the existing arms use. This is plan item F
from the issue, which shipped as a mass-balance arm on the one fixture that structurally cannot
test it.

### #5 — `src/ode/predictions.rs:1228`, doc denies what both call sites do

```rust
/// `input_rate` is the **runtime forcing slice**, not `ode.input_rate`, mirroring
/// [`active_infusions`]' parameter exactly: a caller that does not apply the
/// input-rate forcing passes `&[]` and keeps its plain `+rate` (`ode/ekf.rs`). Asking
/// [`input_rate_consumes_cmt`] instead would read the *spec* and suppress a rate no
/// forcing replaces.
```

Both call sites added in this PR (`:5693`, `:5988`) pass literally `&ode.input_rate`, and
`ekf.rs:253` calls `active_infusions`, not this function — so the caller the doc points at does not
exist for `gated_infusions`. The *contract* is right; state it positively:

```rust
/// `input_rate` is the forcing slice the caller will actually apply, not necessarily the
/// spec's. Both call sites here pass `&ode.input_rate` because that is what their
/// `prepare_input_rates(ode, …)` builds `prepared` from; a caller that applies no forcing
/// passes `&[]` (as `active_infusions`' EKF caller does) and keeps its plain `+rate`.
/// Reading [`input_rate_consumes_cmt`] instead would hard-wire the spec and suppress a rate
/// no forcing replaces.
```

### #6 — `src/ode/predictions_tests.rs:4698`, oracle independence overstated

```rust
/// equal an explicit **sub-dose train**, an oracle that shares no convention with
/// `R_in_inf` (it delivers the same mass as N ordinary boluses through the ordinary
/// bolus path).
```

It does not take the ordinary bolus path: a bolus into an input-rate compartment is suppressed at
`predictions.rs:2366` and delivered through `add_prepared_input_rate_forcing`'s
`prep.rate(tad, mass)`. The train and the object under test share the whole `PreparedInputRate`
machinery and differ only in `rate` vs `rate_infused`. Still the right oracle for the
*convolution* — say that instead:

```rust
/// equal an explicit **sub-dose train**: N sub-doses delivered through the same kernel via
/// `prep.rate`, which is the convolution `rate_infused` claims to compute. It pins
/// `rate_infused` against `rate`, not the kernel itself.
```

### #7 — `src/ode/predictions_tests.rs:4690`, unreachable assert

```rust
assert_relative_eq!(got, expected, max_relative = 5e-3);
assert!(
    got < 1.5 * expected,
    "{tag}/{engine}: delivered {got:.6}, expected {expected} (double-count?)"
);
```

The second can never fire — the first already bounds `got ≤ 1.005·expected`. Delete it, or fold
its message into the first assert so the "(double-count?)" hint survives.

### #8 — `src/stats/likelihood.rs:3929`, `drug_free`

```rust
let drug_free = -(-0.05_f64).exp().ln() - (0.01_f64).ln();
```

`-(-0.05).exp().ln()` is identically `0.05`. The constant is `H0·t − ln H0`, with `H0 = 0.01` and
`t = 5.0` inlined as bare literals disconnected from the `theta TVH0(0.01, …)` and
`ObsRecord::Event { time: 5.0 }` above. Change either and the non-degeneracy guard silently checks
the wrong baseline:

```rust
let (h0, t_event) = (0.01_f64, 5.0_f64); // must track the fixture's TVH0 and event time
let drug_free = h0 * t_event - h0.ln();
```

### #9 — `src/stats/likelihood.rs:4035`, reference quantized at the tolerance

`NONMEM_CHZ_AT_EVENT = 1.2217e-1` is 5 significant figures from `pktte_inf.tab`; its own rounding
is ±5e-6 = **4.1e-5 relative**, against the test's `rel < 1e-4` — 2.4× headroom over the
*reference's printing precision*, not over ferx's error (I measured 1.62e-5). The anchor cannot
discriminate below ~4e-5 and the bound cannot be tightened, because `$TABLE` in `pktte_inf.ctl`
carries no `FORMAT=`. Fix at the source on the next NONMEM run:

```
$TABLE ID TIME CMT DV IPRED CHZ HAZ ETA1 MDV NOPRINT ONEHEADER NOAPPEND
       FORMAT=s1PE15.8 FILE=pktte_inf.tab
```

then re-bank the reference and tighten. Not blocking — it passes today.

### #10 — `src/ode/predictions.rs:1264`, third copy of the predicate

`input_rate.iter().any(|f| f.cmt == cmt)` joins `input_rate_consumes_cmt` (`:1126`) and
`active_infusions` (`:932`). #1196 rightly defers the `Spanning`/`Gated` unification, but the
predicate is a zero-risk extraction that fits in this PR:

```rust
/// Does any applied input-rate forcing consume `cmt` (0-based)? The single spelling of the
/// membership rule shared by both infusion resolvers and by [`input_rate_consumes_cmt`].
pub(crate) fn forcing_consumes_cmt(
    input_rate: &[crate::pk::absorption::InputRateForcing],
    cmt: usize,
) -> bool {
    input_rate.iter().any(|f| f.cmt == cmt)
}
```

with `input_rate_consumes_cmt(ode, cmt_1based)` becoming
`forcing_consumes_cmt(&ode.input_rate, cmt_1based.saturating_sub(1))`. The PR's premise is that
this predicate drifting between two copies cost 214×; it lands a third.

---

**Suggested split.** #1 and #5–#8 are edits to this branch. #2–#4 are new tests and could
reasonably be one follow-up issue if you want this landed now — but given #1171's precedent (a fix
assumed to reach every engine through a shared helper, which turned out to be one level too
shallow), I'd rather they land here. #9 needs a NONMEM re-run; #10 is optional and could fold into
#1196.

Issue #1170, opened 2026-08-31
Publish ferx-core, ferx-tools, ferx-cli to crates.io — decisions needed before anything is claimed


TeunP:
Nothing here is urgent and nothing is done. This is a decision record so the choices get made
deliberately rather than by whoever runs `cargo publish` first. @roninsightrx — the calls in
**Decisions** are yours as much as anyone's; I'd rather not pick them alone.

## Where we actually stand

The repo is already publish-shaped, and that appears to be deliberate rather than accidental:

- `description`, `license = "MIT"`, `repository`, `authors` on all three manifests
- every internal dependency carries **both** forms — `ferx-core = { path = "../..", version = "0.3.0" }`
  — which is exactly what crates.io requires
- no `publish = false` anywhere, and **no git dependencies** (the usual hard blocker)
- `exclude` already tries to strip the CC BY-NC data
- `crates/ferx-cli/Cargo.toml` carries a comment stating the installed binary stays `ferx`

Measured, not assumed:

```
ferx-core    FREE       ferx-cli    FREE       ferx-tools   FREE       ferx-nlme  FREE
ferx         TAKEN      v0.1.0-alpha.1, published 2026-08-19, 17 downloads
                        "Reactive Extensions for async Rust" — github.com/dasantos/ferx
docs.rs/ferx-core  404              crates.io/ferx-core  "does not exist"
cargo package -p ferx-core  →  1257 files, 19.6 MiB (5.1 MiB compressed)  — under the 10 MiB cap
```

`ferx` is gone and is not recoverable — a real crate with a real repo is not squatting under
crates.io policy. **It also costs us nothing:** crate name and binary name are independent, and
`ferx-cli` already declares `[[bin]] name = "ferx"`, so `cargo install ferx-cli` puts `ferx` on
PATH.

Publishing is **permanent**. Per the cargo book: "a publish is generally permanent. The version
can never be overwritten, and the code cannot be deleted." Only `cargo yank` exists afterwards.

## What publishing does and does not buy

It does **not** help ferx-r. ferx-r depends on ferx-core through a **git** dependency with a local
`[patch]`, so it never touches crates.io either way.

What it buys: name security, a docs.rs page, and `cargo install ferx-cli` for anyone who wants the
binary without a git checkout.

## Blocker found while checking this — #1085's `exclude` globs do not match what they name

`data/mbma_naproxen.LICENSE` states:

[...]

PR #1198, opened 2026-09-02
chore(package): fix the exclude globs, cut 0.3.1, add registry metadata [#1170]

roninsightrx:
## Why

Publish prep for #1170. Three things had to be true before the first `cargo publish`, and none of them were:

1. **The `exclude` globs did not match the files they name.** `data/mbma_naproxen.LICENSE` states that "`Cargo.toml` excludes `data/mbma_naproxen.*` from the packaged crate, so the crate published to crates.io is uniformly MIT." `cargo package --list` disagreed: `data/mbma_naproxen_source.py` shipped (the glob needs a literal `.` after the stem, and this has `_source`) and `examples/mbma_naproxen.ferx` shipped (`exclude` never mentioned `examples/` at all).

   **This was never a licence breach.** The restricted CC BY-NC csv itself *was* correctly excluded, verified absent from the file list throughout. Both leaked files are repo-authored MIT — a conversion script and a model file. But the sentence in the licence note was true by luck rather than by the mechanism it named, and `examples/mbma_naproxen.ferx` shipped as an example that cannot run, because its data is excluded.

2. **The version number collided with an existing tag.** The manifests said `0.3.0`, but git tag `v0.3.0` points at `6c0f317f` (2026-08-07) — 507 commits back, at a point where `crates/` did not exist and neither did `ferx-tools` or `ferx-cli`. Publishing today's main as `0.3.0` would have put the registry copy on a completely different tree from the one the tag names.

3. **No `keywords` or `categories` on any of the three manifests.** #1170 decision 3 declined to claim the `ferx-nlme` name specifically because keywords and categories close the discoverability gap for free — so they have to actually be there.

## What changed

**`exclude` globs** — dropping the `.` covers all four `data/mbma_naproxen*` files; `examples/mbma_naproxen.ferx` is listed alongside the slow test for the same reason the slow test is listed (without the excluded csv it cannot run). The claim in `data/mbma_naproxen.LICENSE` is updated to quote the glob that now exists.

**Release 0.3.1** — all three published crates and their internal `version =` deps bumped, and the accumulated `[Unreleased]` section cut as `## [0.3.1] - 2026-09-02` with the compare links updated. 0.3.1 rather than 0.3.0 keeps tag and registry in agreement; 0.3.0 simply never appears on crates.io.

**Registry metadata** — `keywords` (5 per crate, all within the 20-character limit) and `categories` on all three manifests. Category slugs validated against the live crates.io list at `/api/v1/category_slugs`.

## Alternatives considered

For the version: force-moving tag `v0.3.0` to the publish commit would also make tag and registry agree, but it rewrites what an already-cut internal release means. Publishing as `0.3.0` and documenting the drift was the other option; a fresh patch number is cheaper than a permanent footnote, since a publish can never be undone.

## Cross-repo dependency

ferx-r declares both git deps in `src/rust/Cargo.toml` with no version requirement (`{ git = ..., branch = "main" }`), so the 0.3.1 bump does not change its resolution. No lock bump needed.

## Breaking changes
- [x] None

## Numerical validation

Packaging metadata only; no code path is touched.

## Performance

## Tests
- [ ] Regression test added that fails without this fix
- [x] No new tests needed (why: the assertion is about tarball contents, not runtime behaviour, and it is checked directly — see below)

Verified with `cargo package`, which is the tool that would catch a regression here:

| crate | files in tarball | naproxen or `api/` entries |
|---|---|---|
| ferx-core | 1289 | 0 |
| ferx-tools | 22 | 0 |
| ferx-cli | 8 | 0 |

`cargo package -p ferx-core` (the full verify build, not just `--list`) exits 0, so nothing excluded is load-bearing. Tarball is 5.47 MB, under the 10 MiB registry cap.

## Docs
- [x] `CHANGELOG.md` entry added — under `[0.3.1]`, not `[Unreleased]`, since this PR cuts that release

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — no `src/` file is touched
- [x] `Cargo.toml` version bumped

### ferx-site / ferx-book
- [x] No DSL, estimator, fit-option, example, or data-file change — nothing to mirror

### Example execution
- [x] No example execution step needed (packaging metadata only; no user-visible behaviour change)

Note that `examples/mbma_naproxen.ferx` stays in the repo and stays runnable from a git checkout — it is only kept out of the published tarball, where its data is absent.

## Reviewer hints

The load-bearing line is the missing `.` in `data/mbma_naproxen*`. Worth confirming the intent: it now also matches anything future named `mbma_naproxen<anything>` under `data/`, which is what we want for this fixture but is broader than the old pattern.

The version choice is the other thing to sanity-check — 0.3.1 as a patch bump does understate a month of work that added two crates and `ferx gam`, but it is a deliberate call to keep the first crates.io version aligned with a tag we can actually cut at this commit.

## Open questions

None blocking. Ownership after the first publish (#1170 decision 6) is settled: the org now has a `maintainers` team (privacy `closed`, so crates.io can see it; members TeunP, roninsightrx, hiddevandebeek), so the post-publish step is `cargo owner --add github:FeRx-NLME:maintainers` on each of the three crates. A team owner can publish but cannot edit the owner list, so at least one individual owner should stay on each crate alongside the team.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,13 @@ section of the SDLC for the versioning policy).
   top of them, so a tool and the CLI cannot diverge in how a model is loaded.
 
 ### Fixed
+- **The packaged crate no longer ships two files the CC BY-NC exclusion was meant to cover (#1170).**
+  `exclude`'s `data/mbma_naproxen.*` glob required a literal `.` after the stem, so
+  `data/mbma_naproxen_source.py` slipped through, and `examples/` was never listed, so
+  `examples/mbma_naproxen.ferx` shipped as an example whose data is excluded and so cannot run.
+  Neither file was itself CC BY-NC (both are repo-authored MIT, and the restricted csv was
+  correctly excluded throughout), but the licence note in `data/mbma_naproxen.LICENSE` described
+  a mechanism the globs did not implement. Both are now excluded.
 - **An infusion into a built-in absorption compartment is no longer delivered twice on the
   dense/states engines (#1187).** With a `RATE>0` or `RATE=-2` dose into a `first_order()`,
   `transit()`, `igd()` or `weibull()` kernel (#719 gap 2), the mass was released *both* through
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ section of the SDLC for the versioning policy).
   fit with the `--gam` flag. See [GAM covariate screening](https://ferx-nlme.github.io/ferx-core/tools/gam-screening.html).
 
 ### Changed
+- **The three published crates now carry `keywords` and `categories` (#1170).** Registry metadata
+  only -- no code change. It is what covers discoverability on crates.io, which is why claiming
+  the `ferx-nlme` name was judged unnecessary in #1170.
 - **`FitOptions::threads` now also caps a multi-start fit (#1115).** A pinned positive `threads`
   was honored only when `n_starts <= 1`; with several starts the fan-out ran on the full-width
   shared pool regardless, so a tool that pinned one thread per fit from a `PoolPlan` and asked
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+## [0.3.1] - 2026-09-02
+
 ### Added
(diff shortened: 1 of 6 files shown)


roninsightrx:
FYI, this version is not really a proper release-version, it is meant as a first dummy release for crates.io. Proper 0.4.0 release will follow soon.

PR #1200, opened 2026-09-02
feat(api): report bootstrap progress as each replicate finishes

roninsightrx:
## Why

A 200-sample bootstrap is minutes to hours of fitting behind one command, and until it
returned there was nothing on the terminal to say whether it was making progress or
wedged. PsN prints per-sample lines; ferx printed the run header and then went silent
until the table.

The same gap exists in R: `ferx_bootstrap()` blocks with no output at all. Rather than
teach `ferx-tools` about terminals, this gives it an event stream and lets each front end
render it — indicatif here, `cli` in ferx-r.

## What changed

`ferx_tools::bootstrap::run_bootstrap_with_progress(prepared, options, Option<ProgressFn>)`
emits a `BootstrapEvent` per completed fit:

- `Started { replicates, reused, base_fit }` — once the run knows its own size, which
  `--resume` changes: 10 replicates missing from a 200-sample directory is a bar of 10.
- `BaseFitDone`, `ReplicateDone { completed, total }`,
  `DeltaOfvDone { completed, total }`, `Finished`.

`run_bootstrap` is the same call with `None`, so nothing that does not want progress pays
for it, and no existing caller changes.

The completion count lives in an `AtomicUsize` on the producer side rather than in the
sink. `par_iter` hands results back in index order but produces them in completion order,
so a sink counting its own calls would draw a bar that sits still and then jumps; keeping
it in the event also means most sinks need no state at all.

`ferx-cli` gains `bootstrap_progress.rs`: an indicatif bar with the base fit on its own
spinner (a bar at 0/200 through a single fit reads as stuck) and a separate bar for the
`--dofv` pass, which is a second sweep over every replicate and roughly doubles the run
time. New dependency: `indicatif` (CLI only — `ferx-tools` stays free of any terminal
concern).

```
  [00:00:02] [====================>         ] 4/6 replicates (ETA 3s)
Fitted 6 replicates in 4.5s
```

`--no-progress` suppresses it. A redirected stderr needs no flag: indicatif resolves
`ProgressDrawTarget::stderr()` to a hidden target off a terminal, so logs, pipes and
schedulers are unaffected.

## Alternatives considered

**Print from inside the fit loop.** Simplest, and wrong for the second consumer: R cannot
render a `cli` bar from Rust stderr, and the CLI would own a format the R package would
then have to parse back.

**Poll the journal from outside.** `--directory` already appends `raw_results.csv` per
replicate, so a caller could count rows. It forces `--directory` on a run that may not
want artefacts, and gives nothing for the base fit or the Δofv pass.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#321 | open | after (its `Cargo.lock` bump needs this merged) |

## Breaking changes

- [x] None

`ferx-tools` is not covered by the `ferx-core` public-API baseline, and `ferx-core` itself
is untouched, so `api/ferx-core-public-api.txt` is unchanged.

## Numerical validation

- [x] Not applicable (presentation only)

`watching_a_run_does_not_change_it` pins exactly that: the same seed run with and without a
sink returns the same estimates. The draws, the fits and the artefacts do not depend on
whether anyone is watching.

## Performance

One atomic increment and one sink call per completed fit — a fit being seconds to minutes.

## Tests

`ferx-tools` (`bootstrap/mod_tests.rs`), both real runs on warfarin at `--samples 1`:

- `a_run_reports_its_fits_in_order` — asserts the whole sequence, not a set: a bar told
  its length after it has been stepped draws wrong. Runs with `--dofv` so every variant is
  covered.
- `watching_a_run_does_not_change_it` — no base fit, so it also pins that `BaseFitDone` is
  not emitted when there is nothing to report.

`ferx-cli` (`bootstrap_progress.rs`), 4 render tests: no bar during the base fit, the
length is what is left to fit, the Δofv pass replaces the replicate bar rather than
carrying it past its end, and a disabled progress draws nothing.

## Example

- [x] Not applicable (no `.ferx` surface changes)

```bash
ferx bootstrap warfarin.ferx --data warfarin.csv --samples 200 --threads 8
ferx bootstrap warfarin.ferx --data warfarin.csv --samples 200 --no-progress
```

## Docs

- [x] `docs/` updated for user-visible changes — new "Watching a run" section in
      `docs/tools/bootstrap.qmd`, plus the `--no-progress` note in `docs/cli.qmd`
- [x] No new pages, so `docs/_quarto.yml` is unchanged; `cargo run -p docs-lint -- --check` green

## Checklist

      (`--all-targets`), rustdoc, docs, public-API baseline

## Docs & examples

- [x] No new DSL syntax, `[fit_options]` key, example model or data file
- [x] Verified by running the CLI under a pty (`script -q /dev/null cargo run -p ferx-cli -- bootstrap examples/warfarin.ferx --data data/warfarin.csv --samples 6`) — the frames above are that run
- [x] ferx-r rebuilt against this branch (`FERX_NO_AUTODIFF=1 R CMD INSTALL .`) and its full testthat suite passes

## Reviewer hints

Two places worth the attention:

- **Where the count comes from** (`mod.rs`, `run_one`). The argument for the producer-side
  atomic is in the comment there; if you disagree with it, the sink signature is what
  changes.
- **`Started` is emitted after the resume bookkeeping**, not at the top of the function.
  That is the first point at which the run knows how many replicates it will actually fit.

The CLI module is mechanical and can be skimmed, apart from the `DeltaOfvDone` arm, which
is where the replicate bar is retired.

## Open questions

The sink is `Fn(BootstrapEvent) + Send + Sync` rather than a trait. A trait would allow a
stateful renderer without interior mutability, at the cost of a public trait in
`ferx-tools`. If a second tool (SCM?) wants the same treatment, that is the moment to
revisit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@ section of the SDLC for the versioning policy).
 ## [0.3.1] - 2026-09-02
 
 ### Added
+- **A progress bar for `ferx bootstrap`.** A 200-sample bootstrap is minutes to hours of
+  fitting behind one command, with nothing on the terminal to say whether it was working
+  or wedged. The run now reports each fit as it completes: the base model gets a spinner,
+  the replicates a bar with an ETA, `--dofv` its own bar for its second pass, and a
+  `--resume` run counts only what is left to fit. The bar is drawn only when stderr is a
+  terminal; `--no-progress` suppresses it. `ferx_tools::bootstrap::run_bootstrap_with_progress`
+  is the same run with a `BootstrapEvent` sink, which is what ferx-r renders onto a `cli`
+  bar; `run_bootstrap` is unchanged.
 - **`ferx gam` — GAM covariate pre-screening from the command line (#1114).** Screens every
   declared covariate against every ETA with independent GAM regressions and prints a table
   ranked by delta-AIC (`AIC_null − AIC_best`), the Rust equivalent of Xpose4's `xpose.gam()`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,6 +140,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.16.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fe5f465a4f6fee88fad41b85d990f84c835335e85b5d9e6e63e0d06d28cba7c"
+dependencies = [
+ "encode_unicode",
+ "libc",
+ "unicode-width",
+ "windows-sys",
+]
+
+[[package]]
 name = "const-oid"
 version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -240,6 +252,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 
 [[package]]
+name = "encode_unicode"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
+
+[[package]]
 name = "equivalent"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -267,6 +285,7 @@ version = "0.3.1"
 dependencies = [
  "ferx-core",
  "ferx-tools",
+ "indicatif",
  "serde_json",
  "tempfile",
 ]
@@ -395,6 +414,19 @@ checksum = "d466e9454f08e4a911e14806c24e16fba1b4c121d1ea474396f396069cf949d9"
 dependencies = [
  "equivalent",
  "hashbrown",
+]
+
+[[package]]
+name = "indicatif"
+version = "0.18.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "25470f23803092da7d239834776d653104d551bc4d7eacaf31e6837854b8e9eb"
+dependencies = [
+ "console",
+ "portable-atomic",
+ "unicode-width",
+ "unit-prefix",
+ "web-time",
 ]
 
 [[package]]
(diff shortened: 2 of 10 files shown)


roninsightrx:
## Code review (medium)

Reviewed `493a9cca` against `origin/main`. The event contract itself holds up - three findings, all in how the events are counted and rendered.

### 1. `crates/ferx-cli/src/bootstrap_progress.rs:102` - the delta-OFV bar swap keys on a racy `completed == 1`

`completed` comes from `done.fetch_add(1, Ordering::Relaxed) + 1` inside `compute_delta_ofv`'s `par_iter`, and the sink runs on whichever worker finished the evaluation. So the event *carrying* `completed == 1` is not necessarily the first one *delivered*.

If `completed = 2` arrives first, `state.bar` is still the replicate bar, so neither branch of `completed == 1 || state.bar.is_none()` fires and `set_position(2)` lands **on the replicate bar** - wrong length, and still labelled `replicates`. When the `completed == 1` event does arrive the bar is replaced and restarts at 1, discarding whatever was already drawn.

This is the default path, not a corner: `compute_delta_ofv`'s `par_iter` runs on the **global** rayon pool, outside the `.install()` at `mod.rs:921` that honours `--threads`, so the delta-OFV pass is multi-threaded even at `--threads 1`.

Fix: track the bar's kind in `State` and swap when the live bar is not already the delta bar - the same key [ferx-r#321](https://github.com/FeRx-NLME/ferx-r/pull/321) moved its handler to.

### 2. `crates/ferx-cli/src/bootstrap_progress.rs:96`, `:109` - the bar can step backwards

`set_position(completed)` under the same out-of-order delivery renders 5, then 4. Cosmetic on the bar itself, but it also feeds indicatif's `{eta}` rate estimate. `set_position(max(bar.position(), completed))` covers both call sites.

### 3. `crates/ferx-tools/src/bootstrap/mod.rs:955` - `Finished` is not emitted on the error path

Every `?` between `Started` and the emit returns early: `Journal::create`, `manifest.write`, the base `fit`, the `ok_or` for the delta-OFV reference, and `compute_delta_ofv` itself. Neither renderer clears from a `Drop` (`BootstrapProgress` has none), so a failed run leaves its last bar frozen above the error message.

Lowest severity of the three. The R side is already covered - its handler forces `envir`, so cli tears the bar down with the frame - but the CLI is not.

### Note on the tests

`a_run_reports_its_fits_in_order` passes for the wrong reason: `samples: 1` means exactly one `DeltaOfvDone`, so the strict in-order assertion cannot observe finding 1. At `samples > 1` that assertion is flaky on the global pool.

### Checked and clean

- `base_fit: options.run_base_model && reused_original.is_none()` matches the `None if options.run_base_model` match arm exactly, so a resume emits no spurious `BaseFitDone`.
- `n_todo = draws.len() - already_done.len()` cannot underflow: `load_resumable` bounds indices to `1..=options.samples` and dedups, and `draws.len() == options.samples`.
- `run_bootstrap` kept as a wrapper, so the API stays source-compatible.

Reviewed with [Claude Code](https://claude.com/claude-code)

PR #1203, opened 2026-09-02
feat(api): honour a cancellation flag in the bootstrap [closes #1161]

### LINKED ISSUE #1161: bootstrap: honour a cancellation flag so a long run can be aborted cleanly
Follow-up from the R-side bootstrap consumer (FeRx-NLME/ferx-r#313, which notes this as an open question).

## Problem

`run_bootstrap` has no cooperative cancellation. `BootstrapOptions` carries no cancel flag and the function never checks one, so a 200-replicate run cannot be aborted once it has started - it runs to completion or the process dies.

## What already works

More than one might expect. `replicate_options()` (bootstrap/mod.rs:502) clones the caller's base `FitOptions` and clears only `verbose`, `threads`, `run_covariance_step`, `checkpoint`, `checkpoint_path` and `sir` - it does **not** clear `cancel`. `compute_delta_ofv` likewise clones (bootstrap/mod.rs:814). So a caller that sets `prepared.parsed.fit_options.cancel = Some(flag)` before calling already has the flag reach every replicate's `ferx_core::fit`, and every in-flight replicate will unwind on its own.

So this is not "cancellation is impossible", it is "cancellation is not clean".

## What is missing

1. **No short-circuit between replicates.** The rayon loop (`draws.par_iter().map(run_one)`, bootstrap/mod.rs:714-716) keeps scheduling the remaining replicates after the flag is set. Each one enters `fit`, immediately observes the flag and returns `Err`, so a cancelled 200-sample run still walks all 200 draws. Cheap per draw, but pointless work and a confusing delay.
2. **No distinct cancelled outcome.** Those unwound replicates are indistinguishable from genuinely failed fits, so a cancelled run reports as "200 replicates failed" rather than "cancelled by user". Any caller wanting to raise a clean cancellation condition has to infer it from its own flag rather than from the return value.
3. **The base fit and the dofv pass are uncovered by any explicit check** - they inherit the flag through the same clone, but nothing decides what a cancellation during those phases should return.

## Suggested shape

- Add `pub cancel: Option<CancelFlag>` to `BootstrapOptions`, so callers do not have to reach through `prepared.parsed.fit_options` to a field whose propagation is an implementation detail of `replicate_options`.
- Check it at the top of `run_one` so cancelled draws are skipped rather than fitted-then-discarded, and between the base-fit / replicate / dofv phases.
- Return a distinguishable error (or a flag on `BootstrapResult`) so a caller can report cancellation as cancellation.

`resummarize` needs nothing: it fits no models.

## Consumer side

The R half is tracked separately in FeRx-NLME/ferx-r - `ferx_bootstrap()` currently calls `run_bootstrap` synchronously on the R main thread, so it cannot poll for Ctrl-C at all. That is the blocking half of making this user-visible; this issue only makes the resulting abort clean and reportable.

roninsightrx:
## Why

`ferx_tools::bootstrap::run_bootstrap` had no cooperative cancellation, so a 200-replicate
run could not be aborted once started — it ran to completion or the process died (#1161).

More than nothing already worked, which is what made the gap subtle. `replicate_options`
clones the caller's `FitOptions` and clears `verbose` / `threads` / `run_covariance_step` /
`checkpoint*` / `sir` — but not `cancel`. A caller that set
`prepared.parsed.fit_options.cancel` already reached every replicate's `fit`, and every
in-flight fit unwound on its own. So the problem was not "cancellation is impossible", it
was "cancellation is not clean":

1. **No short-circuit between replicates.** The rayon loop kept scheduling every remaining
   draw; each entered `fit`, observed the flag and returned `Err`. Cheap per draw, but a
   cancelled 200-sample run still walked all 200.
2. **No distinct cancelled outcome.** Those unwound replicates were indistinguishable from
   genuinely failed fits, so a cancelled run reported as "200 replicates failed".
3. **The base fit and the Δofv pass decided nothing** about what a cancel during them
   should return.

## What changed

`BootstrapOptions::cancel: Option<CancelFlag>` carries the flag, so a caller does not reach
through `prepared.parsed.fit_options` to a field whose propagation is an implementation
detail of this module. A flag set there is still honoured — `effective_cancel` falls back to
it, since that is the route callers had before — and the run-wide one takes precedence.

It is polled at four points, and each one is a decision rather than a sprinkle:

| where | why there |
|---|---|
| before `Journal::create` | opening the journal **rewrites** the four per-replicate files. Check after it and cancelling a mistyped command destroys the artefacts of the run already in the directory — the one the user would otherwise have resumed. |
| between the base fit and the replicates | `--update-inits` starts every replicate from the base fit's estimates, so a cancel during it leaves nothing to resample from. |
| top of `run_one` | where scheduling actually stops. |
| around the Δofv sweep | a second pass over every replicate that roughly doubles the run time, so it is its own cancellation point rather than a stretch that ignores the flag. |

The run ends with a new `BootstrapError::Cancelled`, so a caller can report an abort as an
abort. `run_bootstrap` / `run_bootstrap_with_progress` now return
`Result<_, BootstrapError>`; `From<BootstrapError> for String` keeps the CLI's `?` working
unchanged. `resummarize` is untouched — it fits nothing.

**A replicate the cancel unwound is dropped, not journaled.** Its `Err` cannot be told
apart from a genuine failure, and `--resume` carries a recorded failure *forward* rather
than refitting it (PsN's default, and the right one) — so journaling a cancelled replicate
would silently cost the resumed run a replicate.

## Alternatives considered

- **A sentinel string / a `cancelled: bool` on `BootstrapResult`.** Rejected: the result
  would then carry a summary computed over a truncated set of replicates, which is an
  interval nobody asked for and one a caller can forget to check. The issue offered either
  shape; the typed error is the one that cannot be ignored.
- **Recording cancelled replicates with a distinct status instead of dropping them.** That
  needs a new column in `raw_results.csv`, a new state in the resume reader, and a rule for
  what `--retry-failed` does with it — all to represent "nothing happened".
- **A signal handler in `ferx bootstrap`.** Out of scope here and not what the issue asks
  for; the CLI has nothing to set the flag from yet, so its behaviour is unchanged.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#315 | open (consumer half) | after |

ferx-r builds `BootstrapOptions { … }` as an exhaustive literal and types
`run_bootstrap_reporting` as `Result<_, String>`, so its **local** builds need the two-line
follow-up (`cancel: None`, `.map_err(|e| e.to_string())`). Its CI is unaffected until
`src/rust/Cargo.lock` is bumped, which is the same PR's job. FeRx-NLME/ferx-r#315 is the
real consumer work: `ferx_bootstrap()` currently calls `run_bootstrap` on the R main thread
and so cannot poll for Ctrl-C at all. This PR is the engine half.

## Breaking changes
- [x] Public Rust API — `ferx-tools` only, which has no committed baseline (the
      `api/ferx-core-public-api.txt` gate covers `ferx-core`, and it is unchanged; the
      `Public API baseline` job is green). Two source-breaking changes for out-of-workspace
      callers: the new `BootstrapOptions::cancel` field and the `BootstrapError` return
      type. Both land in ferx-r via the PR above.

## Numerical validation
- [x] Not applicable — no numerics changed. The bootstrap's own oracles are untouched and
      still green: the identity-resample bit-for-bit anchor, the duplication complement, the
      PsN percentile thresholds, the stratum-composition check, and thread-count
      determinism. The one *behavioural* equivalence this PR does add is stronger than a
      tolerance: a cancelled-then-resumed run's `raw_results.csv` is compared byte-for-byte
      (timings stripped) against the uninterrupted run's.

## Performance
- [x] No significant impact expected — one relaxed atomic load per replicate, against a fit
      measured in seconds.

## Tests

Tier 1 (`crates/ferx-tools/src/bootstrap/mod_tests.rs`):

- the run-wide flag reaches every replicate's `FitOptions`, and reaches it as a *share*
  rather than a copy (cancel the original, the installed one observes it);
- a flag on the model's own `FitOptions` is still honoured, and the run-wide one wins when
  both are set;
- `BootstrapError`: `is_cancelled`, `Display` (a failure is not dressed up on the way
  through), and both `From` conversions the internal `?`s depend on.

Tier 2 (`crates/ferx-tools/tests/bootstrap_end_to_end.rs`), on warfarin at
`outer_maxiter = 2`:

- **cancelled before it starts** → `Cancelled`, and the completed run already in the
  directory is byte-identical afterwards (this is the check-before-`Journal::create` order);
- **cancelled mid-run on one thread** → `Cancelled`, exactly header + base + one replicate
  on disk, then `--resume` reuses that one and lands on a `raw_results.csv` identical to the
  uninterrupted run's;
- **cancelled between the base fit and the replicates on a 4-thread pool** → no replicate
  fitted at all;
- **cancelled mid-run on a pool** → nothing the cancel unwound is journaled as a failure;
- **cancelled during the Δofv sweep** → `Cancelled`, no `delta_ofv.csv`, replicates intact.

One note for the reviewer, since the test says it and it is easy to read as a gap: the
pool test deliberately asserts *no bound on how many replicates got through*. On a pool
that is a timing property, not a contract — every fit already past the entry check runs to
completion, and the flag is a relaxed atomic. An earlier draft did assert a bound and failed
about one run in three under CPU load (verified by hammering the binary under a full-core
hog: 4/12 failures before, 0/12 after). Where stopping *is* a contract — one thread, and the
base-fit boundary — it is pinned exactly.

## Docs
- [x] `docs/tools/bootstrap.qmd` — a new "Cancelling a run" section: the API, and the three
      consequences of where the flag is checked (a cancelled run is not a failed one;
      everything finished is on disk and resumes; a cancel before the run starts is a no-op
      on the directory). It also states plainly that the CLI installs no signal handler yet
      and that `ferx_bootstrap()` runs on the R main thread, so neither can set the flag —
      both tracked separately.
- [x] `CHANGELOG.md` `[Unreleased]` → `Added`.

## Checklist
      `--all-targets`, rustdoc, docs lint, public-API baseline
- [x] `cargo test -p ferx-tools` (185 tests) and `cargo test -p ferx-cli` (119) green

## Reviewer hints

The two subtle spots are both in `run_bootstrap_with_progress`: the *order* of the first
check against `Journal::create`, and the `Err(_) if is_cancelled(&cancel) => return None`
arm in `run_one` — dropping rather than recording is what keeps `--resume` correct. The
rest is mechanical.

https://claude.ai/code/session_01L512jnkzvDYJ4wnKummK6r

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,15 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Added
+
+- **A cancellable bootstrap (#1161).** `BootstrapOptions::cancel` takes a `CancelFlag`; setting
+  it from another thread stops a long `ferx_tools::bootstrap` run at the next replicate boundary
+  and returns the new `BootstrapError::Cancelled`, so a caller reports an abort as an abort
+  rather than as "every remaining replicate failed". Replicates the cancel unwound are dropped
+  rather than journaled as failures, so `--resume` refits them; everything already finished stays
+  on disk and resumes into exactly the run that was cancelled.
+
 ## [0.3.1] - 2026-09-02
 
 ### Added
--- a/crates/ferx-tools/src/bootstrap/mod.rs
+++ b/crates/ferx-tools/src/bootstrap/mod.rs
@@ -25,8 +25,8 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Instant;
 
 use ferx_core::{
-    fit, CompiledModel, FitOptions, FitResult, ModelParameters, OmegaMatrix, Population,
-    PreparedRun, SigmaVector, WarningCode,
+    fit, CancelFlag, CompiledModel, FitOptions, FitResult, ModelParameters, OmegaMatrix,
+    Population, PreparedRun, SigmaVector, WarningCode,
 };
 use rayon::prelude::*;
 
@@ -124,6 +124,65 @@ impl Drop for FinishGuard<'_> {
     }
 }
 
+/// Why a bootstrap run ended early.
+///
+/// A cancelled run is not a failed one, and the distinction is not cosmetic
+/// (#1161). Cancellation reaches the replicates as the same [`CancelFlag`] an
+/// ordinary fit polls, so every in-flight fit unwinds with an `Err` of its own;
+/// without a cancelled outcome of its own the run would report an abort as
+/// "every remaining replicate failed", which is both wrong and the one thing a
+/// caller most needs to be able to tell apart.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BootstrapError {
+    /// The [`BootstrapOptions::cancel`] flag was set while the run was in
+    /// flight, so the run stopped at the next replicate boundary.
+    ///
+    /// Whatever had already been fitted is on disk in the run directory - the
+    /// journal flushes each replicate as it lands - so `resume` picks the run
+    /// up where the cancel stopped it, refitting only what never completed.
+    Cancelled,
+    /// Anything else, carrying the message a caller should show.
+    Failed(String),
+}
+
+impl BootstrapError {
+    /// Whether the run stopped because its cancel flag was set.
+    pub fn is_cancelled(&self) -> bool {
+        matches!(self, BootstrapError::Cancelled)
+    }
+}
+
+impl std::fmt::Display for BootstrapError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            BootstrapError::Cancelled => f.write_str("the bootstrap was cancelled"),
+            BootstrapError::Failed(message) => f.write_str(message),
+        }
+    }
+}
+
+impl std::error::Error for BootstrapError {}
+
+impl From<String> for BootstrapError {
+    fn from(message: String) -> Self {
+        BootstrapError::Failed(message)
+    }
+}
+
+impl From<&str> for BootstrapError {
+    fn from(message: &str) -> Self {
+        BootstrapError::Failed(message.to_string())
+    }
+}
+
+/// So a caller whose own error type is a `String` - the CLI, the R glue - keeps
+/// working through `?` and loses only the distinction it was not using.
+impl From<BootstrapError> for String {
+    fn from(error: BootstrapError) -> String {
+        error.to_string()
+    }
+}
+
 /// Everything the bootstrap needs beyond the model and data themselves.
 ///
 /// Defaults mirror PsN's, including the two skip filters that are on by default
@@ -179,6 +238,19 @@ pub struct BootstrapOptions {
     /// alternative. Turn it on when the failure was a transient resource
     /// problem — an out-of-memory kill, a full disk — rather than the model.
     pub retry_failed: bool,
+    /// Cooperative cancellation for the whole run (#1161).
+    ///
+    /// Set from another thread, this stops the run at the next replicate
+    /// boundary and returns [`BootstrapError::Cancelled`]. It is also cloned
+    /// into every replicate's [`FitOptions`], so a fit already in flight
+    /// unwinds instead of running to completion first.
+    ///
+    /// The flag lives here rather than only on the model's own `FitOptions`
+    /// because reaching through `prepared.parsed.fit_options` to set it depends
+    /// on the replicate options happening to clone the field — an
+    /// implementation detail of this module. A flag set there is still
+    /// honoured; this one simply takes precedence.
+    pub cancel: Option<CancelFlag>,
 }
 
 impl BootstrapOptions {
@@ -260,6 +332,7 @@ impl Default for BootstrapOptions {
             confidence_level: 95.0,
             resume: false,
             retry_failed: false,
+            cancel: None,
         }
     }
 }
@@ -621,7 +694,16 @@ pub fn strata_from_csv(
 ///   *same* `{model}.tmp`, so they would restore each other's state;
 /// * **no SIR** — a per-replicate uncertainty step inside an uncertainty
 ///   analysis is pure cost.
-fn replicate_options(base: &FitOptions, keep_covariance: bool) -> FitOptions {
+///
+/// `cancel` is the run's one effective flag (see [`effective_cancel`]) and is
+/// installed unconditionally, so a `BootstrapOptions::cancel` overrides
+/// whatever the model file's own `FitOptions` carried and a run without one is
+/// left exactly as it was.
+fn replicate_options(
+    base: &FitOptions,
+    keep_covariance: bool,
+    cancel: &Option<CancelFlag>,
+) -> FitOptions {
     let mut o = base.clone();
     o.verbose = false;
     o.threads = Some(1);
@@ -629,7 +711,23 @@ fn replicate_options(base: &FitOptions, keep_covariance: bool) -> FitOptions {
     o.checkpoint = false;
     o.checkpoint_path = None;
     o.sir = false;
+    o.cancel = cancel.clone();
     o
+}
+
+/// The flag this run polls: [`BootstrapOptions::cancel`] when the caller set
+/// one, otherwise whatever the model's own [`FitOptions`] carried.
+///
+/// The `FitOptions` path is not a fallback for tidiness' sake — it is what
+/// callers had before this field existed, since the replicate options are a
+/// clone of the base ones, and dropping it would silently stop honouring a flag
+/// that used to work.
+fn effective_cancel(options: &BootstrapOptions, base: &FitOptions) -> Option<CancelFlag> {
+    options.cancel.clone().or_else(|| base.cancel.clone())
+}
+
+fn is_cancelled(flag: &Option<CancelFlag>) -> bool {
+    flag.as_ref().is_some_and(|f| f.is_cancelled())
 }
 
 fn diagnostics_from(result: &FitResult) -> (bool, bool, bool) {
@@ -727,7 +825,7 @@ fn reject_id_dependent(referenced: &[String]) -> Result<(), String> {
 pub fn run_bootstrap(
     prepared: &PreparedRun,
     options: &BootstrapOptions,
-) -> Result<BootstrapResult, String> {
+) -> Result<BootstrapResult, BootstrapError> {
     run_bootstrap_with_progress(prepared, options, None)
 }
 
@@ -742,11 +840,19 @@ pub fn run_bootstrap(
 /// replicate — never in a hot loop, since a replicate is a whole fit — so it
 /// may do real work (redraw a bar, call back into R). It must not panic: a
 /// panic crosses the `par_iter` boundary and takes the run with it.
+/// # Cancellation
+///
+/// Setting [`BootstrapOptions::cancel`] from another thread ends the run at the
+/// next replicate boundary with [`BootstrapError::Cancelled`], leaving the
+/// replicates that had already finished in the run directory for `resume`. A
+/// replicate that was mid-fit when the flag was set is dropped rather than
+/// recorded: its `Err` cannot be told apart from a genuine failure, and a
+/// failure is what `resume` declines to refit.
 pub fn run_bootstrap_with_progress(
     prepared: &PreparedRun,
     options: &BootstrapOptions,
     progress: Option<ProgressFn<'_>>,
-) -> Result<BootstrapResult, String> {
+) -> Result<BootstrapResult, BootstrapError> {
     let emit = |event: BootstrapEvent| {
         if let Some(sink) = progress {
             sink(event);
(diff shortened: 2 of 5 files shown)


roninsightrx:
## Review

Design is right: the four checkpoints are each argued rather than sprinkled, the check-before-`Journal::create` ordering is the subtle one and it is correct, and `Err(_) if is_cancelled(&cancel) => return None` is the right call given `--resume` carries a recorded failure forward. Docs, changelog and tests are all present. One correctness gap, plus three smaller notes.

---

### 🔴 A cancel *during* the base fit returns `Failed`, not `Cancelled`

`crates/ferx-tools/src/bootstrap/mod.rs:974-987`

```rust
let result = fit(
    &prepared.parsed.model,
    &prepared.population,
    template,
    &{ /* base_options with cancel installed */ },
)?;
```

`base_options` carries the flag, so this is exactly the case the PR is about — and `ferx_core::api::fit` signals it with `Err("cancelled by user".to_string())` (`src/api/fit.rs:1327-1328`, `1733-1734`). The `?` runs `From<String> for BootstrapError`, so the run returns `BootstrapError::Failed("cancelled by user")` and `is_cancelled()` is `false`. The post-base-fit check at `mod.rs:1021` is never reached.

This matters more than a mid-loop cancel: on a real model the base fit is the single longest fit in the run, and `--update-inits` makes it the one a user is most likely to be sitting through. The PR's own table (“between the base fit and the replicates … a cancel during it leaves nothing to resample from”) and `docs/tools/bootstrap.qmd` (“ends the run at the next replicate boundary with `BootstrapError::Cancelled`”) both promise `Cancelled` here.

The test does not cover it: `a_cancel_between_the_base_fit_and_the_replicates_fits_no_replicate` fires the flag on `BootstrapEvent::BaseFitDone`, i.e. after `fit` has already returned `Ok`. The gap is cancel-*during*, which never reaches line 1021.

Fix at the call site:

```rust
let result = match fit(&prepared.parsed.model, &prepared.population, template, &{ … }) {
    Ok(r) => r,
    Err(_) if is_cancelled(&cancel) => return Err(BootstrapError::Cancelled),
    Err(e) => return Err(BootstrapError::Failed(e)),
};
```

and add a test that sets the flag from a `BootstrapEvent::Started` sink (or pre-cancels a run whose directory does not yet exist, so the first check passes but the base fit does not).

Worth considering the general form too: any `?` in `run_bootstrap_with_progress` that fires while the flag is set reports a cancellation as a failure. The base fit is the only one reachable today, but a helper added later inherits the trap. A single `map_err` wrapper — `if is_cancelled(&cancel) { Cancelled } else { Failed(e) }` — would close the class rather than the instance.

---

### 🟡 `is_cancelled` duplicates a public `ferx-core` helper

`crates/ferx-tools/src/bootstrap/mod.rs:729`

```rust
fn is_cancelled(flag: &Option<CancelFlag>) -> bool {
    flag.as_ref().is_some_and(|f| f.is_cancelled())
}
```

`ferx_core::cancel::is_cancelled(&Option<CancelFlag>)` already exists and is in the committed baseline (`api/ferx-core-public-api.txt:129`) — same signature, same semantics. Import it instead; a second copy of the polling predicate is one more place for the two to disagree about what `None` means.

---

### 🟡 Setting both flags silently drops the `FitOptions` one

`mod.rs:719` (`o.cancel = cancel.clone()`) + `effective_cancel`

`effective_cancel` is `options.cancel.or_else(|| base.cancel)`, and `replicate_options` then installs the winner *unconditionally*. So a caller that sets `prepared.parsed.fit_options.cancel = Some(A)` **and** `BootstrapOptions::cancel = Some(B)` ends up with replicates that no longer poll `A` at all — before this PR they did. “Takes precedence” is documented for the run-level decision, but the doc comment on the field says “A flag set there is still honoured”, which is only true when the run-wide one is absent. Either tighten the wording, or poll both (`a.is_cancelled() || b.is_cancelled()`), which is what “still honoured” actually reads as.

---

### 🟢 The base-fit cancel path skips `j.into_result()`

`mod.rs:1021` returns before `journal.into_result()?` at `mod.rs:1115`. Every row is flushed on append, so nothing is lost — but `into_result` is the only place a journal write failure surfaces, so on this one path a half-written `raw_results.csv` returns `Cancelled` and the next `--resume` reads it. Cheap to close: `if let Some(j) = journal { j.into_result()?; }` before the early return, or hoist the check to after the loop only.

Two nits while in there: `BootstrapError` is not `#[non_exhaustive]`, so any future variant is a source break for ferx-r's `match`; and `Failed(String)` gives `Error::source()` nothing, which is fine but worth a conscious `// no source: the payload is already the message`.

---

### Things I checked and am happy with

- `compute_delta_ofv` maps a cancelled fit's `Err` through `.ok()` to `None` rather than propagating, so the sweep cannot leak a `Failed("cancelled by user")` past the `mod.rs:1146` check — the base-fit bug does not have a twin here.
- `Ordering::Relaxed` on a set-once monotonic flag is correct; there is no ordering the loop depends on.
- `FinishGuard::drop` covers every cancelled return, so the `Started`/`Finished` pairing still holds; and the pre-journal check returns before `Started` is emitted, so that path owes nothing.
- Returning before `output::write_all` leaves `raw_results.csv` in completion order, which `load_resumable` keys by `sample` — the resume test's byte-for-byte comparison against the uninterrupted run is a real check, not a tautology.
- The deliberate absence of a replicate-count bound in the pool test is the right call, and the reasoning in the PR body should stay in the test comment where it is.

roninsightrx:
Cross-repo row update: the ferx-r compile-compat half is now open as FeRx-NLME/ferx-r#322, and it merges in **either order** with this PR — its two edits (`..BootstrapOptions::default()` and `.map_err(|e| e.to_string())`) compile against both the pinned ferx-core `5f7ef926` and this branch, so it needs no `src/rust/Cargo.lock` bump. FeRx-NLME/ferx-r#315 remains the real consumer work: exposing cancellation to R needs the fit moved off the R main thread.

roninsightrx:
## Fixes pushed — c9f1b64c

All four findings addressed.

### 🔴 A cancel during the base fit returned `Failed` — fixed

Confirmed the bug before fixing it. With the base fit's bare `?` in place, a run cancelled from a `BootstrapEvent::Started` sink returns:

```
assertion `left == right` failed: a cancel during the base fit was reported as a failure: cancelled by user
  left: Failed("cancelled by user")
 right: Cancelled
```

The fix is a classifier applied at the one `?` that wraps a fit:

```rust
fn cancel_aware(error: String, cancel: &Option<CancelFlag>) -> BootstrapError {
    if is_cancelled(cancel) { BootstrapError::Cancelled } else { BootstrapError::Failed(error) }
}
```

**On closing the class rather than the instance** — I proposed that in the review and then didn't do it, so the reasoning belongs here. Wrapping the whole body (rename to an inner fn, reclassify `Failed` → `Cancelled` at the boundary when the flag is set) is one line and covers every future `?`, but it also swallows the checks *above* the first cancellation point: `validate`, `validate_for_run` and the allocation run before the flag is ever consulted, so `--samples 0` with a pre-set flag would report "cancelled" and hide the typo. That is a worse failure than the one being fixed. The rule is written on `cancel_aware` instead — *any `?` on something that runs a fit goes through here* — and today that is a set of one, since `compute_delta_ofv` already swallows its fit errors with `.ok()` and cannot propagate a cancellation.

Attribution when a genuine failure races a cancel goes to the cancel, matching the call `run_one` already makes when it drops rather than journals such a replicate.

### 🟡 Duplicate `is_cancelled` — fixed

Now `use ferx_core::cancel::is_cancelled;`. Local copy deleted.

### 🟡 Both flags set — doc fixed, behaviour kept

Kept run-wide precedence: the replicate options install exactly one `CancelFlag` into `FitOptions`, and two `Arc<AtomicBool>`s cannot be OR-ed into one without a mirroring thread. The field doc no longer claims a `FitOptions` flag is "still honoured" unconditionally — it now says honoured **when this one is `None`**, and to set one or the other, not both.

### 🟢 Journal close on the base-fit cancel path — fixed

`j.into_result()?` now runs before that early return. Deliberately *not* run through `cancel_aware`: a journal write failure there means the base row is missing, and `--resume` would read the directory a row short — reporting the full disk beats returning `Cancelled` over a directory that will then misbehave.

Also `#[non_exhaustive]` on `BootstrapError`, so a variant added later is not a source break for ferx-r. The doc example's `Err(e) =>` catch-all already satisfies it, and FeRx-NLME/ferx-r#315 needs no change.

### Tests

- Tier 2 `a_cancel_during_the_base_fit_is_a_cancellation_not_a_failure` — cancels on `Started`, which is exact rather than racy: `fit` polls the flag at the top of its method chain (`src/api/fit.rs:1327`), so the base fit returns `Err` deterministically. Verified red before the fix (output above), green after.
- Tier 1 `an_error_raised_under_a_set_flag_is_a_cancellation` — the classifier, including the genuine-failure-racing-a-cancel case and the no-flag case.

`tools/preflight.sh` green (fmt, check, clippy, rustdoc, docs, public-api — the `ferx-core` baseline is unchanged). `cargo test -p ferx-tools` 187 and `cargo test -p ferx-cli` 119, all passing.

PR #1205, opened 2026-09-02
fix(api): demote converged on an internal runaway-guard hit [closes #1118]

### LINKED ISSUE #1118: Demote `converged` when an estimate is pinned to an internal runaway guard
## Problem

#1099 added a fit-end check for parameter estimates pinned to ferx's *internal* packed-space guards — `runaway_guard_estimates` / `runaway_guard_warning` in `src/api/postfit.rs:1171`, wired into every fit at `src/api/fit.rs:2123`. It emits a `Warning`-severity `WarningCode::ParameterAtRunawayGuard` with structured `details`.

It does not touch `converged`. A fit whose Ω or σ ran to an implementation ceiling still returns `converged: true`, so a programmatic consumer that keys off the boolean retains an estimate that is by construction not an interior optimum.

Unlike a user-declared THETA bound — where a boundary hit can be a valid constrained optimum — an internal guard hit cannot be. `runaway_guard_estimates` already excludes declared THETA ranges (`theta_guard_is_internal`), so the remaining hits are exactly the cases where the flag is wrong.

## Suggested direction

Demote `converged` to `false` when `runaway_guard_warning` returns a hit, in `api/fit.rs` alongside the existing call, so it applies to every estimator rather than one. Decide whether the severity should rise from `Warning` to `Critical` at the same time (the VI bad-basin warning `W_VI_BAD_BASIN` from #1103 is `Critical`).

Points to settle in review:

- Lower-guard hits and upper-guard hits may deserve different treatment. An Ω collapsing to the floor is often a legitimately unsupported random effect the user should remove, not a numerical runaway — demoting `converged` there may be noisier than useful. An upper-guard hit is unambiguously bad.
- The chained-methods path (`methods = vi, focei`) should demote only on the final stage, matching how `run_covariance_step` is gated.

## History

Extracted from #1104, which implemented a VI-only version of this (`apply_convergence_health_gates`) before #1099 landed the general detector. #1104 is closed as superseded; this is the one piece of it that has no equivalent on `main`.

roninsightrx:
## Why

#1099 added a fit-end check for estimates pinned to ferx's **internal** packed-space guards
(`runaway_guard_estimates` / `runaway_guard_warning`), wired into every fit. It emits a typed
`WarningCode::ParameterAtRunawayGuard` — but it does not touch `converged`. A fit whose Ω or σ ran
to an implementation ceiling still returned `converged: true`, so a programmatic consumer keying
off the boolean retained an estimate that is by construction not an interior optimum.

Unlike a user-declared THETA bound — where a boundary hit can be a valid constrained optimum — an
internal guard hit cannot be. `runaway_guard_estimates` already excludes declared THETA ranges
(`theta_guard_is_internal`), so the remaining hits are exactly the cases where the flag is wrong.

Closes #1118. (Extracted from the closed #1104, which had a VI-only version of this.)

## What changed

- `runaway_guard_warning(converged: &mut bool, params)` — the signature now mirrors
  `vi::run::bad_basin_warning` (#1098), which owns the same consequence for the ELBO check, so the
  side/severity rules and the demotion live in one place and are unit-testable without a fit.
- **Upper-guard hit → `converged = false`, severity `Warning` → `Critical`.** The estimate stopped
  at an implementation ceiling; it is not a solution of the problem the user posed. The message
  says `Reported converged: false`, matching the VI bad-basin wording.
- **Lower-guard hit → unchanged** (`Warning`, `converged` untouched). One of the points the issue
  asked to settle: a variance collapsing to the floor is usually a genuinely unsupported component
  for the user to remove, not a numerical runaway, and demoting there would be noise rather than
  signal. A *mixed* set contains an upper hit, so it demotes.
- **Chained methods**: the call site in `fit()` runs on `result.params`, the final stage's
  estimates, so `methods = vi, focei` is gated on its last stage — same rule as
  `run_covariance_step`. The other point the issue asked to settle.
- `classify_warning` learns the same split. Each hit is listed as `... at <side> guard`, so the
  side survives into the flat message; a warning that reaches the string path (a spliced
  multi-start message) is classified `Critical` iff the native typed entry would have been.
- `warning_entry` gained a `warning_entry_with_severity` variant for emitters whose severity
  depends on what they found; every existing caller is unchanged.

Knock-on, intended: multi-start ranks candidates by `(converged, ofv)`, so a guard-pinned start no
longer outranks a valid one on OFV alone.

## Alternatives considered

- **Demote on any hit, upper or lower.** Rejected per the issue's first open point — an Ω at the
  floor is a modelling decision to make (remove the random effect), not a numerical failure, and
  flipping `converged` on it would fire on a large class of ordinary over-parameterized fits.
- **Keep `Warning` and demote anyway.** Rejected: `Critical` is documented as "the result is
  untrustworthy as-is", and `vi_bad_basin` — the exact analogue, a demotion-carrying quality gate —
  is already `Critical`. Splitting them would make severity useless for branching.
- **Return a third tuple element / a struct instead of `&mut bool`.** The `&mut converged` form is
  what the codebase already uses for this exact job in `bad_basin_warning`; matching it keeps one
  idiom rather than two.

## Cross-repo dependency

No public-API change (`runaway_guard_warning` is `pub(crate)`; the baseline is unchanged). ferx-r
reads `converged` and `warnings_structured` as it already does.

## Breaking changes
- [x] None

Behavioural note for consumers: a fit that previously reported `converged: true` with a
`parameter_at_runaway_guard` warning now reports `converged: false`. That is the fix.

## Numerical validation
- [x] Not applicable (no numerical path touched — no estimator, kernel, sensitivity, or likelihood
      code is modified; this only changes the fit-end reporting of a point that was already
      computed and already warned about)

## Performance

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 3863 passed, 0 failed)

Tier 1 (`src/api/tests/simulate_with_uncertainty_tests.rs`, `src/types_tests.rs`) — the existing
guard tests now also pin the flag and the severity on every side, for σ, Ω-Cholesky, and hidden
θ caps: upper demotes and is `Critical`, lower does not and stays `Warning`, mixed demotes, and no
hit / a `FIX`'d coordinate leaves the flag alone. `classify_warning` is pinned on both sides.

Tier 2 (`tests/runaway_guard_demotes_convergence.rs`) — pins the *wiring*, that `fit()` applies the
rule to the `FitResult` a consumer reads. An evaluation-only IMP stage reports at the parameters it
is handed without moving them, which reaches the fit-end check on a guard-pinned point with no
convergence loop (0.09s). It runs a control at the model's own inits first, so the demotion is
attributable to the guard and not to the eval-only path.

- [x] Not applicable (no new API surface — the warning and its `details` payload are #1099's, and
      unchanged)

## Docs
- [x] `docs/` updated for user-visible changes — `docs/warnings.qmd` severity table row now reads
      `Critical / Warning` and says an upper hit is reported `converged: false`

`docs/warnings.qmd`'s "Warning codes" section is R1-baselined, so the rewritten row was kept
shorter than the one it replaces (349 vs 355 chars) rather than growing a baselined section.

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy `--all-targets`,
      rustdoc, docs lint, public-API baseline
- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched

## Docs & examples
- [x] No example execution step needed (fix with no user-visible DSL / option / example change)

## Reviewer hints

The two decisions worth your attention are the ones #1118 flagged for review: **lower-guard hits do
not demote** (`src/api/postfit.rs`, the `if has_upper` block and the doc comment above
`runaway_guard_warning`), and **the chain gates on the last stage** (`src/api/fit.rs`, the call
site — it is already positioned after the whole chain, so this is a comment, not a code change).

`src/types.rs`'s `classify_warning` arm is the one easy-to-miss piece: it keys on the literal
`"at upper guard"` substring that each hit's list entry produces. If the hit-line format in
`runaway_guard_warning` ever changes, that arm silently reverts to `Warning` — the test in
`src/types_tests.rs` pins both sides against exactly that.

## Open questions

None.

https://claude.ai/code/session_01EZ5ksNYgr4jLn6MMUqX1XG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,23 @@ section of the SDLC for the versioning policy).
 - **A fit whose estimate ran to an internal safety rail no longer reports `converged: true`
   (#1118).** ferx caps a few packed coordinates internally (an implicit THETA cap, the OMEGA /
   SIGMA runaway rails); unlike a THETA bound you declared, one of those cannot be a valid
-  constrained optimum. When a free estimate ends pinned to an *upper* internal guard the fit is
-  now reported as not converged and the `parameter_at_runaway_guard` warning is raised to
-  `Critical`, so a script or agent keying off the boolean stops accepting a point that is by
-  construction not an interior optimum. A *lower*-guard hit (a variance collapsing toward zero)
-  is unchanged: it stays a `Warning` and leaves `converged` alone, because that is usually an
-  unsupported component to remove rather than a numerical runaway.
+  constrained optimum. When a free estimate ends pinned to a rail the fit is now reported as not
+  converged and the `parameter_at_runaway_guard` warning is raised to `Critical`, so a script or
+  agent keying off the boolean stops accepting a point that is by construction not an interior
+  optimum. A *collapse* hit — a variance falling to its floor at zero — is unchanged: it stays a
+  `Warning` and leaves `converged` alone, because that is usually an unsupported component to
+  remove rather than a numerical runaway. Each listed hit now says which of the two it is
+  (`verdict: runaway` / `collapse` in the warning's `details`), because the side does not decide
+  it: an OMEGA off-diagonal is bounded symmetrically at ±10, so a correlation driven to the
+  *lower* rail is a runaway too. A SIGMA at its ceiling additionally suggests rescaling DV or
+  using a proportional / log-transformed error model, since that rail is the one an otherwise
+  sound model can reach on unscaled data.
+- **`ferx bootstrap` reflects the same rule.** A replicate that ends at a runaway rail is now a
+  non-converged replicate, so with the default `skip_minimization_terminated` it is excluded
+  from the confidence intervals and counts against the reported
+  `minimization_successful` fraction. Re-running a bootstrap of an unchanged model can therefore
+  report slightly different CIs than before; `--summarize` over a stored `raw_results.csv`
+  re-applies the criteria without refitting.
 
 ## [0.3.1] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,17 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Fixed
+- **A fit whose estimate ran to an internal safety rail no longer reports `converged: true`
+  (#1118).** ferx caps a few packed coordinates internally (an implicit THETA cap, the OMEGA /
+  SIGMA runaway rails); unlike a THETA bound you declared, one of those cannot be a valid
+  constrained optimum. When a free estimate ends pinned to an *upper* internal guard the fit is
+  now reported as not converged and the `parameter_at_runaway_guard` warning is raised to
+  `Critical`, so a script or agent keying off the boolean stops accepting a point that is by
+  construction not an interior optimum. A *lower*-guard hit (a variance collapsing toward zero)
+  is unchanged: it stays a `Warning` and leaves `converged` alone, because that is usually an
+  unsupported component to remove rather than a numerical runaway.
+
 ## [0.3.1] - 2026-09-02
 
 ### Added
--- a/docs/estimation/parameterization.qmd
+++ b/docs/estimation/parameterization.qmd
@@ -91,8 +91,31 @@ optimizer scaling can introduce), the fit emits the structured
 `parameter_at_runaway_guard` warning. FIX'd coordinates are excluded because
 their equal bounds are intentional. The check is made in packed space, so it
 compares directly with the literal guard even though fit output reports Omega
-on the variance/covariance scale and Sigma on its stored SD scale. A lower hit
-means the component collapsed toward zero; an upper hit means it ran away.
+on the variance/covariance scale and Sigma on its stored SD scale.
+
+Each hit carries a **verdict**, and it is the verdict — not the side — that says
+what happened and what follows from it:
+
+- **`collapse`** — the coordinate fell to a floor at zero. Only the log-packed
+  coordinates have one: a Theta lower cap, an Omega / Omega-IOV Cholesky
+  *diagonal* at $-6$, a Sigma at $-8$. The warning stays a `Warning` and
+  `converged` is left alone, because an unsupported component collapsing is
+  usually a modelling decision to make rather than a numerical failure.
+- **`runaway`** — the coordinate is held at an implementation rail. Every upper
+  hit is one, and so is *either* rail of an Omega / Omega-IOV **off-diagonal**,
+  which is the raw $L_{ij}$ bounded symmetrically at $\pm 10$: a correlation
+  coordinate driven to $-10$ ran away exactly as its $+10$ twin did, and nothing
+  about it collapsed toward zero. A runaway hit **demotes `converged` to `false`
+  and raises the warning to `Critical`** (#1118): a point held at an
+  implementation rail is by construction not an interior optimum, so a script or
+  agent keying off the boolean must not accept it.
+
+One rail is reachable by a model that is otherwise sound: the Sigma ceiling,
+$e^5 \approx 148$ on the stored SD scale. An additive error model on unscaled
+data (DV in ng/mL, µg/L, cell counts) can have a residual SD genuinely above it.
+The remediation there is the data scale rather than the model — rescale DV, or
+use a proportional or log-transformed error model — and the warning message says
+so when the hit is a Sigma.
 
 ## Where the packed space surfaces
 
--- a/docs/warnings.qmd
+++ b/docs/warnings.qmd
@@ -59,7 +59,7 @@ Each `WarningCode` token, what it flags, and a typical response.
 | `eps_shrinkage` | Warning | Residual (EPS) shrinkage high / negative. | Sparse data for the error model — simplify it. |
 | `eta_shrinkage` | Warning | One or more ETA shrinkages exceed ~30%. | Data poorly inform that IIV — consider removing it on the affected parameter(s). |
 | `boundary_estimate` | Warning | One or more THETA estimates are pinned to an optimizer bound. | Non-identifiability or a too-tight bound — inspect; relax the bound or simplify. |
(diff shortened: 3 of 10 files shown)


roninsightrx:
## Code review

Diff mechanics are clean: the demotion is wired at the single fit-end call site, `result.converged` has no other reader in `fit_inner`, multi-start ranking and `rebuild_warnings_structured` both pick it up, and the new `classify_warning` arm sits after the `did not converge` arm without the new "Reported converged: false" wording being hijacked by it.

The findings below are all about the **upper/lower side split**, not the wiring.

### 1. `src/api/postfit.rs:1374` — "lower" is not always a collapse

Ω / Ω_IOV **off-diagonal** Cholesky coordinates are box-bounded at ±10 (`parameterization.rs` bounds table), so a correlation running away to a large negative `L[i,j]` reports `side = "lower"`. Under the new rule that hit does not demote `converged`, stays `Warning`, and tells the user the parameter "collapsed toward zero at an implementation floor" — the opposite of what happened. The upper-side twin of exactly the same runaway does demote.

Scenario: a block Ω whose off-diagonal is unidentified drifts to `L = -10`; the fit reports `converged: true` alongside a message advising the user to remove the component.

Suggest keying the demotion on **magnitude** rails (for off-diagonals: both sides) rather than on the literal string side.

### 2. `src/api/postfit.rs:1372` — σ's upper rail is reachable by legitimate models

The packed σ ceiling is `exp(5) ≈ 148` on the stored σ scale. An additive error model on unscaled data (DV in ng/mL, µg/L, cell counts) can have a true residual SD above that, pin at the rail, and now come back `converged: false` + `Critical`, with no knob to widen the rail or opt out. Previously that fit was usable-with-a-warning.

Worth at least a doc note ("rescale DV / use a proportional or log-transformed model"), since the remediation text ("revisit the model, data, initial estimates, or method") doesn't name the actual fix.

### 3. `crates/ferx-tools/src/bootstrap/summary.rs:166` — silent knock-on to bootstrap CIs

`skip_minimization_terminated && !r.converged` drops replicates. Replicates that pin an upper guard are now dropped where they previously counted, and `frac(&|r| r.converged)` (line 293) reports a lower convergence rate.

Behaviourally defensible, but the PR body only mentions the multi-start knock-on — this one changes reported CIs on a run whose model didn't change, and it isn't in the CHANGELOG.

### 4. `docs/estimation/parameterization.qmd:88-95` — stale

The canonical description of the internal guards still says only that a hit "emits the structured `parameter_at_runaway_guard` warning", and ends with "an upper hit means it ran away". Now that an upper hit also flips `converged` and reports `Critical`, this page — the one a user lands on from the packed-space discussion — no longer states the consequence. `docs/warnings.qmd` was updated; this one wasn't.

### Checked and cleared

- No other reader of the stale `result.converged` after the call site.
- `warning_entry` still has three callers (no dead code).
- The `"at upper guard"` substring the `classify_warning` arm keys on is produced verbatim by the hit list, and is pinned on both sides in `src/types_tests.rs`.
- The `[METHOD]` prefix strip happens before the lowercase match, so a chained-stage message still classifies.
- ferx-r only styles/sorts on `"critical"` and won't error.

roninsightrx:
## Review findings addressed — d6bb437c

All four, plus tests.

### 1 & 2 — the side no longer decides the verdict (`src/api/postfit.rs`, `src/estimation/parameterization.rs`, `src/types.rs`)

Each hit now carries a **verdict**, `runaway` or `collapse`, derived from the coordinate *kind* as well as the side:

- every upper hit is a runaway;
- **either** rail of an Ω / Ω_IOV off-diagonal is a runaway, because that coordinate is the raw `L[i,j]` bounded symmetrically at ±10 — nothing about `L = -10` collapsed toward zero;
- only the log-packed coordinates (a THETA cap, an Ω Cholesky diagonal, a Σ) have a lower rail that means collapse.

`converged`, the `Critical` severity, the guidance text and a new `details.verdict` field all key off the verdict. The per-hit message line became `... at <side> guard, <verdict>)` and `classify_warning` now matches `"guard, runaway"`, so a flat multi-start-spliced message classifies the same way the typed entry does.

The layout walk lives in `parameterization.rs` as `PackedCoordKind` / `coordinate_kinds`, mirroring `packed_fixed_mask`. `test_coordinate_kinds_match_the_bounds_table` pins it slot-for-slot against `compute_bounds` and asserts that exactly the off-diagonal kind has symmetric rails, so a coordinate whose bounds change cannot drift away from its kind silently.

For the Σ ceiling: a Σ runaway now appends the remediation the generic advice was missing — *"A SIGMA at the ceiling can also mean the residual error is genuinely that large on the scale of the data: rescale DV, or use a proportional or log-transformed error model."* Both docs pages say the same. The message stays Σ-specific; a THETA cap does not pick it up (pinned in the test).

### 3 — bootstrap knock-on

Now in the CHANGELOG under `Fixed`: a replicate ending at a runaway rail is a non-converged replicate, so under the default `skip_minimization_terminated` it is excluded from the CIs and counts against the reported `minimization_successful` fraction; re-running a bootstrap of an unchanged model can report slightly different CIs, and `--summarize` re-applies the criteria over a stored `raw_results.csv` without refitting.

### 4 — `docs/estimation/parameterization.qmd`

The Bounds section now states the consequence rather than just the warning: the two verdicts, which coordinates can produce each, that a runaway demotes `converged` and reports `Critical` (#1118), and the Σ-ceiling case with its remediation. `docs/warnings.qmd`'s table row and `details` key list were updated to match — the row was kept to its original length because R1's baseline for that section is a ratchet.

### Tests

- `omega_off_diagonal_lower_rail_is_a_runaway_not_a_collapse` — a 2×2 block Ω with `L[1,0] = -10`: asserts the hit still reports `side: "lower"` but `verdict: "runaway"`, demotes `converged`, reports `Critical`, does not say "collapsed toward zero", and reaches `Critical` again through `classify_warning` on the flat message alone.
- `test_coordinate_kinds_match_the_bounds_table` — kind/bounds alignment, as above.
- `classify_warning_recognizes_internal_runaway_guard` — gained the off-diagonal-lower-rail case.
- Existing guard tests updated for the new wording plus positive/negative assertions on the Σ remediation clause.

`tools/preflight.sh` clean; `cargo test --lib --features ci` 3865 passed; `tests/runaway_guard_demotes_convergence.rs` passes.

PR #1207, opened 2026-09-02
fix(ode): reject an escalation whose derivative jets overflowed [closes #1204]

### LINKED ISSUE #1204: bug(ode): the auto guard's finiteness check is value-only, so a finite solve with NaN Dual2 jets is kept silently
Spun off from #1080 (comment item 1), which is otherwise complete. This is the one
item that was deliberately deferred rather than built, because the honest fix has a
real cost and the failure has never been reproduced.

## The gap

The `auto` guard in `resolve_and_guard` decides whether to keep an escalated stiff
solve by checking that every saved state is finite. That check reads **values only**:

```rust
// src/ode/solver.rs
/// Whether every saved state is finite, reading values only (`T = Dual2` decides identically
/// to `T = f64`, which is what keeps the two paths on the same stepper).
fn finite_g<T: PkNum>(points: &[SolPointG<T>]) -> bool {
    points
        .iter()
        .all(|p| p.u.iter().all(|v| v.val().is_finite()))
}
```

The value-only reading is deliberate and load-bearing: it is what makes the `f64`
prediction solve and the `Dual2` sensitivity solve reach the *same* accept/reject
verdict, and therefore run on the same stepper. That invariant is what the whole
`auto` design rests on — a gradient differentiating a trajectory produced by a
different method is worse than either method alone.

But a near-singular Rosenbrock `W` can plausibly return finite *values* whose
derivative components (`Dual2` jets) are NaN or Inf. In that case:

- the prediction path looks clean,
- the guard keeps the escalation,
- `auto_stiff_rejected` reads `0`, so the `ode_solver` warning says nothing,
- and FOCEI/FOCE receives a NaN `∂f/∂η` / `∂f/∂θ` from a solve that reported success.

Same blind-spot shape as the rest of #1080: no counter moves, nothing is reported,
the answer is wrong.

## What this issue needs, in order

1. **A reproducing case first.** A system where the jet goes non-finite while the
   value does not — most likely a stiff binding / TMDD-shaped model driven to a
   near-singular `W` under `rodas4` / `rodas5p`, at a θ where the sensitivity
   equations are far worse conditioned than the states. Without one, any guard added
   here is speculative and untestable, and it would cost the `f64`/`Dual2` stepper
   agreement above to buy nothing.

2. **If a case exists, the likely fix is an asymmetric guard**, adopted knowingly:
   accept on values (`f64` and `Dual2` still agree on *keeping* a solve, so the
   common path keeps one stepper), but reject on a non-finite jet (only the `Dual2`
   instantiation can trip it). That deliberately breaks the same-stepper invariant in
   exactly the case where the `Dual2` solve is already unusable — the trajectory it
   would differentiate is fine, the derivative is not, so re-solving that one
   explicitly is strictly better than returning NaN. Write down the asymmetry where
   `finite_g` is defined rather than leaving it to be rediscovered.

3. **A counter for it.** Whatever the guard does must be visible in `OdeSolverStats`
   and reach the `ode_solver` warning (`api/postfit.rs`), the same as
   `auto_stiff_rejected` / `auto_fallback_failed`. A silent repair is the failure mode
   #1080 was about.

4. **Consider the same question for the fallback.** `auto_fallback_failed` is scored
   with the same value-only `finite_g`, so an explicit re-solve with NaN jets is
   equally invisible.

Tier-1 unit test in `src/ode/stiffness_tests.rs` alongside the existing guard tests
(`escalation_rejected_*`), pinning both the reject and the counter.

## Related

#1080 (parent epic, Parts B and C shipped in #1087 / #1094 / #1108), #978 (Part A,
the a-priori probe), #959, #708.

roninsightrx:
## Why

Closes #1204, the one item spun out of #1080 rather than built there. It was deferred because
the honest fix costs an invariant and the failure had never been reproduced. This PR reproduces
it first.

The `ode_method = auto` escalation guard decided whether to keep an escalated stiff solve by
checking that every saved state was finite — reading **values** only:

```rust
fn finite_g<T: PkNum>(points: &[SolPointG<T>]) -> bool {
    points.iter().all(|p| p.u.iter().all(|v| v.val().is_finite()))
}
```

Value-only was deliberate and load-bearing: it is what makes the `f64` prediction solve and the
`Dual1`/`Dual2` sensitivity solve reach the *same* verdict on the same segment, so the gradient
differentiates the trajectory the predictor reports.

But value-finiteness and jet-finiteness genuinely come apart. A dual's derivative jets carry
higher powers of what its value carries linearly: integrating `u' = p·u` gives `∂u/∂p = t·u` and
`∂²u/∂p² = t²·u`, and a `1/x` in the right-hand side scales the gradient by `1/x²` and the
Hessian by `1/x³`. A trajectory driven near the top of double precision therefore overflows its
**Hessian first, its gradient next, and its value not at all**.

### The repro

`src/ode/stiffness_tests.rs::auto_rejects_an_escalation_whose_jets_overflowed_while_its_values_did_not`
— a fast decaying mode carrying `|Re λ|max = 100` (so the probe escalates) alongside
`u' = (p·K)·u`, where `K` is the units `p` is measured in and so scales the derivatives without
touching the trajectory. Measured at the segment's end:

| | value | gradient | Hessian |
|---|---|---|---|
| `u[1]` | `7.2e303` (finite) | `1.4e306` (finite) | **`NaN`** |

with `stiff_min_step_clamped_steps = 0`, `unfinished_segments = 0`, a stiff solve that finished
normally, and `auto_stiff_rejected = 0`. Every trigger the guard had read clean. The predictions
looked ordinary and FOCEI got a `NaN` gradient out of a solve that reported success.

**One premise in #1204 did not survive measurement, and it is worth recording.** The issue
guessed the mechanism would be a near-singular Rosenbrock `W`. It is not:
`lu_factor_in_place_g` rejects any pivot at or below `1e-12 · scale`, which bounds how far the
elimination can amplify a jet relative to the value. The decoupling lives in the dual arithmetic
itself, not in the linear solve.

## What changed

**1. `PkNum::jets_finite()`** — `true` for `f64` (no jets to check, const-folds away), a
gradient scan on `Dual1`, gradient + Hessian on `DualMixed`/`Dual2`.

**2. The guard rejects on non-finite jets** and re-solves the segment with the pinned explicit
stepper.

This is the asymmetric guard the issue described, adopted knowingly: accept on values, reject on
jets. `jets_finite` is `true` for `f64`, so the prediction path never rejects on this clause and
the break in the same-stepper invariant is confined to the case where the dual result is
*already* unusable — the choice there is between a gradient off by solver tolerance and a `NaN`
one. The reasoning is written down at the guard rather than left to be rediscovered.

**3. `auto_stiff_rejected_jets`** — a new `OdeSolverStats` counter, a subset of
`auto_stiff_rejected`, set only when the rejection was jet-only (nothing stalled, nothing
unfinished, every value finite). Scored apart because it is the one counter that can only ever
be non-zero on a dual instantiation, and because its remedy differs: an ordinary rejection says
"name a different stiff method", a jet rejection says the method worked and the sensitivities did
not, so check the model's units and scaling. It gets its own clause in the `ode_solver` warning
and its own key in the `details` payload.

**4. The explicit fallback is scored on jets too** (#1204 item 4). An `auto_fallback_failed`
judged on values alone would report a repair it did not make.

**5. The post-fit sweep now observes a dual solve.** The diagnostic sweep is an `f64` prediction
pass, which can never take the decision above, so a counter reported only from there would be
exactly the dead diagnostic #1080 item 3 existed to remove. `sweep_sensitivity_solver_stats`
runs one analytic `∂f/∂η` per subject inside the same `SolverStatsScope` — the solve the inner
loop ran throughout the fit, run once more at the estimates the fit reports — and no-ops for
every model not on the analytic ODE sensitivity path.

A **named** `ode_method` stays unguarded, as it is for every other clause here: a user who asked
for `rodas4` gets `rodas4`, non-finite jets included. Pinned by a test.

## Alternatives considered

- **Reject on jets symmetrically** — impossible; `f64` has no jets. There is no version of this
  that keeps the invariant intact.
- **Detect and warn without discarding.** Rejected: the guard's contract is usable-or-not, and a
  `NaN`-jet result is not usable. It is also not always unrepairable — when the *stiff attempt*
  is what diverged (the #959 shape: a finite `4.79e282` on a system whose true solution was
  bounded), the explicit re-solve genuinely fixes it. The fixture here is the other family,
  where the solution really is that large and nothing helps; there the value of the change is
  that `auto_fallback_failed` now fires instead of the failure being silent.
- **Scan jets on every segment, not just escalations.** Rejected as cost: the scan is
  `O(points × states × N²)` on the hot gradient path, and it would extend the guard to named
  methods, which is a separate contract.

## Cross-repo dependency

Additive only: one new `pub` field on `OdeSolverStats` and one new `PkNum` method. Nothing ferx-r
consumes changes shape.

## Breaking changes
- [x] Public Rust API — baseline regenerated with `tools/update-public-api.sh` (10 additions)
- [ ] None

The additions are `OdeSolverStats::auto_stiff_rejected_jets` (both re-export paths), needed so a
caller reading solver statistics can tell an asymmetric rejection from an ordinary one, and
`jets_finite` on `PkNum` / `Dual1` / `DualMixed`, which the guard calls generically and which is
public for the same reason every other `PkNum` method is.

## Numerical validation

- [x] Not applicable in the NONMEM sense — see below.

There is no NONMEM anchor for this object and cannot be: NONMEM has no analytic dual-number
sensitivity path, so it has no counterpart to a solve whose predicted values are finite while its
derivative jets are not. Per CLAUDE.md's rules for such cases, this is validated by **exact
closed forms** and a **degenerate oracle** instead:

- Closed form: the `recip` rule scales the gradient by `1/x²` and the Hessian by `1/x³`, so at
  `x = 1e-120` with an `O(100)` gradient the value (`1e120`) and gradient (`1e242`) are finite
  and the Hessian (`~1e304 · 1e...`) is not. Pinned in `dual_mixed.rs` against hand-computed
  magnitudes, and in `dual1.rs` for the gradient-only type.
- Degenerate oracle: the same ODE fixture at `T = f64` must take **none** of the new decisions
  (`the_jet_clause_is_inert_on_the_f64_prediction_path`) — which is the statement that the
  prediction path is bit-for-bit unchanged.
- Non-degeneracy: a healthy escalation must record no jet rejection
  (`a_healthy_escalation_records_no_jet_rejection`), so the clause cannot pass by rejecting
  everything.

Nothing on the `Dual2` value or derivative *formulas* changed, so the existing `Dual2`-vs-FD
parity suite (`check_full_provider_vs_fd`, the per-kernel `*_g_dual_matches_fd` checks) is the
regression net for the sensitivity path itself; all of it is green.

## Performance

The jet scan runs only inside the escalation guard, on segments `auto` actually escalated, and
only on a dual instantiation — `jets_finite` is a compile-time `true` for `f64`. The post-fit
sensitivity sweep adds one `∂f/∂η` per subject, once, at the end of a fit that ran thousands of
them.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 3876 passed, 0 failed)

Ten new Tier-1 tests:

| where | what |
|---|---|
| `sens/dual_mixed.rs` | the overflow the value does not see; either jet order alone rejects; healthy duals accepted |
| `sens/dual1.rs` | the gradient-only twin — this is the FOCE inner `∂f/∂η`, not just the FOCEI Hessian |
| `sens/num.rs` | `f64` is vacuously true, including at a non-finite *value* |
| `ode/stiffness_tests.rs` | the repro (rejected, counted, fallback failure reported); the `f64` path untouched; a named `rodas4` unguarded; a healthy escalation quiet |
| `api/tests/ode_solver_diagnostics_tests.rs` | the warning clause and its `details` key; an ordinary rejection carries no jet clause; the post-fit sweep observes a dual solve; and no-ops off the analytic ODE path |

- [x] Not applicable — no new API surface a user writes against. The change is visible only as
  an `ode_solver` warning clause and a `details` key.

## Docs
- [x] `docs/` updated — `warnings.qmd` (the `ode_solver` row and the `details` payload list) and
  `model-file/ode-models.qmd` (a new "When the derivatives overflow but the predictions do not"
  section)
- [ ] No user-visible change

The `warnings.qmd` R1 baseline is a ratchet, so the `ode_solver` row was compacted to make room:
the section is now **below** its baseline, not above it. `cargo run -p docs-lint -- --check` is
green.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — no value or derivative
      formula was touched; `jets_finite` is a predicate over the jets, never an input to them,
      so it cannot enter a derivative

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key — none added
- [ ] New data file — none

### ferx-book

### Example execution
- [x] No example execution step needed — no user-visible change to any example's behaviour or
      output; the only user-facing surface is a warning clause that fires on a model none of the
      examples reaches.

## Reviewer hints

**Focus on the asymmetry.** `integrate_resolved_g_inner` in `src/ode/solver.rs` is the whole
change; everything else is plumbing, counters, and tests. The question worth arguing is whether
rejecting on jets while accepting on values is the right trade, and the comment above the guard
makes the case: it fires only where the dual result is already unusable, so the alternative is
not consistency but a `NaN`.

**Second-guess the visibility change.** `sweep_sensitivity_solver_stats` is the one piece that
adds work to every ODE fit. It is one gradient per subject and it no-ops off the analytic path,
but it is a new call on the post-fit path and worth a look.

Skimmable: the counter plumbing (`merge`, the `details` JSON), the docs, and the dual-arithmetic
tests, which are hand-computed magnitudes.

## Open questions

The fixture's explicit fallback also has non-finite jets, because it integrates the same
divergent trajectory — so on *this* family the guard costs a second solve and reports
`auto_fallback_failed` rather than repairing anything. That is the honest outcome (silent `NaN`
→ reported failure), and the repair case is real but belongs to the #959 family, where the stiff
attempt is what diverged. I could not build a deterministic fixture for that half without
depending on a specific method diverging at a specific θ, which would be brittle. Flagging it in
case a reviewer knows a stable one.

https://claude.ai/code/session_01J1QDyTu8Suqjz19d3uTeeE


roninsightrx:
## Review — `/code-review medium`

Reviewed at `c771f030`. The solver change itself holds up: `jets_finite_g` is only consulted alongside `ran_stiff`, the jet-only sub-count is correctly gated on `!stalled && !unfinished && values_finite`, the fallback is scored on the same predicate, and `jets_finite` const-folds to `true` at `f64` so the prediction path is untouched.

The problems are almost all in the **second half** — the post-fit sensitivity sweep — plus one garbled user-facing string. Seven findings stand; one does not. Verified each against the code rather than taking them as read.

### Confirmed

**1. `postfit.rs:1860` — the jet clause emits mangled text. (Medium)**

The clause is a single-line literal that rustfmt then wrapped without `\` continuations, so the message carries six runs of ~14 literal spaces:

```
…all finite but whose              predicted values…
```

Every neighbouring clause in this function uses `\`-continued literals, and `a_switch_is_reported_alongside_an_unclean_solve` already asserts `!msg.contains("  ")` for exactly this reason. My test only checked `contains("edge of double precision")`, so it passed straight over it. My mistake — the clause went in through a script that wrote it as one line.

**2. `postfit.rs:1681` — the sweep ignores `gradient_method = fd`. (Medium)**

`ode_inner_grad_supported_model` is `ODE_SENS_ENABLED && ode_analytical_supported(model)` and never consults `model.gradient_method`. `analytic_inner_common_bail` (`inner_optimizer.rs:1719`) is the predicate the inner loop actually consults, and it *does* check `GradientMethod::Fd`, SDE, and the magnitude × correlated-residual combination. So on an FD fit the sweep ran a `Dual1` solve the fit never performed and reported counters from a code path that fit did not take. My doc comment claiming `subject_eta_grad` "would return `None` for every subject anyway" is simply wrong.

**3. `postfit.rs:1693` — the sweep runs the wrong order of provider. (Medium)**

This is the one that matters most. `subject_eta_grad` is `Dual1` — gradient only. The defect #1204 documents overflows the **Hessian first** (`7.2e303` / `1.4e306` / `NaN`), and that case only arises on the `Dual2` outer solve. So the sweep as written cannot observe the exact failure the commit message, the CHANGELOG and the docs all lead with. `auto_stiff_rejected_jets` would read 0 for it.

**4. `fit.rs:1825` — both sweeps share one `SolverStatsScope`. (Medium)**

Every shared counter (`attempted_steps`, `min_step_clamped_steps`, `auto_stiff_segments`, `unfinished_segments`, …) was roughly doubled and mixed across two different solves. The concrete harm: a `min_dt` clamp occurring only in the gradient sweep now fires the clamp clause, which tells the user "any output times left in a segment the solver could not finish are freeze-padded with the last state" — a claim about predictions that were in fact clean. `details.phase` still says `"postfit_predictions"`.

**5. `postfit.rs:1845` — contradictory advice on a pure jet rejection. (Low/Medium)**

The ordinary clause fires on `rejected > 0`, including when `rejected == rejected_jets`, so the message said both "naming `ode_method = rodas5p` is the next thing to try" and "naming another method will not help". `warnings.qmd` says the remedy differs; the message gave both.

**6. `postfit.rs:1859` — "The fit is repaired, not wrong" is unconditional. (Low)**

In the only demonstrated scenario the explicit re-solve also overflows and `auto_fallback_failed = 1`, so the user reads "the fit is repaired" next to "both attempts fail … no clean trajectory existed".

**7. `solver.rs:1783` — every downstream segment pays a second solve. (Low/Medium, perf)**

Once a dual state's jets are non-finite they stay non-finite through the rest of the subject's event chain, so every subsequent escalated segment fails the clause and pays an explicit re-solve that cannot repair anything. Correctness point as well as cost: one overflow gets counted once per downstream segment, so `auto_stiff_rejected_jets` overstates.

### Not a defect

**8. `postfit.rs:1690` — IOV κ mismatch.** `ode_analytical_supported` returns `false` at `n_kappa != 0` (`ode_provider.rs:340`), and `ode_inner_grad_supported_model` is `ODE_SENS_ENABLED && ode_analytical_supported`. The sweep already no-ops for every IOV model, so it cannot run a κ = 0 trajectory for one. No change needed.

### What I am changing

- Re-wrap the clause with `\` continuations; add the `!msg.contains("  ")` assertion the repo already uses to my tests, so this cannot recur silently.
- Gate the sweep on `analytic_inner_common_bail` as well, so an FD fit sweeps nothing.
- Run `subject_sensitivities` (`Dual2`, gradient **and** Hessian) when the model has an analytic outer gradient, falling back to `subject_eta_grad` when only the inner loop is analytic — i.e. whichever the fit itself ran.
- Collect the sensitivity sweep in its **own** scope and copy only `auto_stiff_rejected_jets` into the reported stats, so every other counter keeps the meaning it had before this PR.
- Make the jet clause self-contained (no "N of those M"), say explicitly that it is disjoint from the escalation counts, and drop the unconditional "repaired, not wrong".
- Skip the rejection when the segment's *entry* jets were already non-finite: nothing in that segment overflowed and an explicit re-solve cannot repair it. Fixes both the cost and the over-count.

Findings 3 and 4 together mean the visibility half of this PR did not do what it claimed, which is worth saying plainly rather than patching quietly.

roninsightrx:
## Review fixes pushed — `9709247a`

All seven confirmed findings addressed. The eighth stays closed as not-a-defect.

| # | Finding | Fix |
|---|---|---|
| 1 | Clause emitted runs of literal spaces | Re-wrapped with `\` continuations; both new tests now carry the `!msg.contains("  ")` assertion the repo already uses, so a clause literal that loses its continuations fails a test instead of reaching a user |
| 2 | Sweep ignored `gradient_method = fd` | Gated on `analytic_inner_common_bail` as well — the predicate the inner loop itself consults |
| 3 | Sweep ran the first-order provider | Runs `subject_sensitivities` (`Dual2`, gradient **and** Hessian) when the model has an analytic outer gradient, `subject_eta_grad` when only the inner loop is analytic |
| 4 | Both sweeps shared one `SolverStatsScope` | Sensitivity sweep gets its own scope; exactly one field — `auto_stiff_rejected_jets` — is copied back |
| 5 | Contradictory remedies on a pure jet rejection | Clause is self-contained, states it is disjoint from the escalation counts; a test pins that the two remedies never both appear |
| 6 | "The fit is repaired, not wrong" unconditional | Removed |
| 7 | One overflow re-scored on every downstream segment | Only a segment that *entered* with finite jets is scored |

### On findings 3 and 4

Worth being plain about: together they meant the visibility half of this PR did not do what its
own commit message claimed. The counter existed, the sweep existed, and the sweep could not
observe the failure the counter is named for — the Hessian-first overflow — because it ran a
gradient-only provider. That is the substantive fix in this push; the rest is message and cost.

### New tests

- `a_jet_rejection_alone_is_a_warning_not_an_info_note` — the prediction pass can be spotless
  (it is a different solve), so `unclean` has to count this or the one failure only the
  sensitivity sweep can see would be reported as an `Info` escalation note. This was a real
  second-order bug of the scope split, found while making the fix.
- `the_sensitivity_sweep_is_a_no_op_when_the_fit_asked_for_fd_gradients` — finding 2.
- `a_segment_entered_with_dead_jets_is_not_rejected_again` — finding 7, and it asserts no second
  solve was paid for, not just that the count is right.
- `the_post_fit_sweep_observes_the_sensitivity_solve_too` now also asserts
  `analytic_outer_gradient_available(&model)`, so the fixture is pinned to the `Dual2` branch —
  otherwise the wiring test would have kept passing on the first-order path that finding 3 is
  about.

### On finding 8 (IOV κ mismatch)

Checked rather than assumed: `ode_analytical_supported` returns `false` at `n_kappa != 0`
(`ode_provider.rs:340`), and `ode_inner_grad_supported_model` is
`ODE_SENS_ENABLED && ode_analytical_supported`. The sweep already no-ops for every IOV model, so
it cannot integrate the κ = 0 trajectory described. No change.

### Verification

`preflight OK — fmt check clippy rustdoc docs public-api`; `cargo test --lib` 3879 passed / 0
failed; `docs-lint --check` green. Docs and CHANGELOG updated for the changed sweep behaviour —
the `details` payload note in `warnings.qmd` now says `auto_stiff_rejected_jets` is disjoint from
the other keys rather than a subset of them, which finding 4 made true.

PR #868, opened 2026-07-16
feat(estimation): estimate block_sigma off-diagonal on the analytic gradient path [closes #847] [CC]

roninsightrx:
## Why

`block_sigma` off-diagonal SIGMA terms were parsed into residual correlations but held **fixed** at their initial value — ferx optimized a different surface than NONMEM's `$SIGMA BLOCK(n)`, which estimates the full lower triangle. On the fluconazole RadboudUMC model NONMEM estimates the residual correlation to ~0.93 while ferx kept the init (~0.2), leaving OFV/Q away from the NONMEM trace (#847).

Two problems surfaced and are fixed here:
1. **Correctness:** estimate the off-diagonal (ρ), not just the diagonal SDs.
2. **Performance:** the existing correlated-residual (`block_sigma`) path built and factored a dense `n×n` R (and, for the FOCEI curvature, an O(n⁴) `∂²R` tensor) **every evaluation** — 2–3 orders of magnitude on richly-sampled subjects, which made estimating ρ impractical.

Closes #847. The remaining fluconazole cold-start gap (Q basin selection) is a separate outer-optimizer conditioning issue tracked in #864 — **not** block_sigma; the objective is confirmed correct (ferx reaches 734.67 from NONMEM's final estimates) and this PR's gradient is validated.

## What changed

- **Estimate ρ.** A plain `block_sigma` block now estimates its off-diagonal as a free parameter (NONMEM `$SIGMA BLOCK(n)` semantics); `block_sigma ... FIX` keeps the old fixed behaviour. ρ is carried on `ModelParameters.residual_correlations` and packed as a Fisher-z (`atanh ρ`) coordinate so it stays in (−1, 1) and every 2×2 residual block stays PD.
- **Analytic gradient path preserved.** ρ rides the analytic FOCE/FOCEI outer gradient (closed-form `∂(−2LL)/∂ρ` via `rho_block`), not a finite-difference fallback. Live ρ is threaded through the inner/outer objective + inner η-gradient (previously the inner gradient read the frozen `model` copy).
- **Diagonal-R fast path** (combined-error `block_sigma`): objective, inner gradient, FOCEI marginal detect diagonal R (`residual_is_diagonal`) and use O(n) closed-form assembly instead of dense Cholesky/inverse/∂R tensors. Data term is bit-for-bit identical.
[...]

PR #1213, opened 2026-09-02
feat(estimation): estimate the block_sigma off-diagonal on the analytic gradient [closes #847]

### LINKED ISSUE #847: Estimate block_sigma off-diagonal parameters
## Problem

The fluconazole RadboudUMC model does not converge to the NONMEM solution because ferx currently treats `block_sigma` off-diagonal entries as fixed residual correlations instead of estimated covariance parameters.

Model/data:

- ferx model: `~/git/ferx/ferx-testdata/fluconazole_radboudumc/ferx/fluconfinal.ferx`
- NONMEM reference: `~/git/ferx/ferx-testdata/fluconazole_radboudumc/nm/fluconfinal014.mod` / `.lst`

NONMEM estimates the full 2x2 SIGMA lower triangle. In the final NONMEM trace, SIGMA moves from:

```text
[0.05,
 0.01, 0.05]
```

to approximately:

```text
[0.20314,
 0.17439, 0.17173]
```

That is a final residual correlation of about:

```text
0.17439 / sqrt(0.20314 * 0.17173) ~= 0.934
```

ferx currently initializes the diagonal sigma SDs from the variances but converts the off-diagonal covariance to a fixed `ResidualCorrelation.rho`. For this model, the fixed rho is:

```text
0.01 / sqrt(0.05 * 0.05) = 0.20
```

So ferx is optimizing a different objective than NONMEM.

## Evidence

Current ferx fit from the default model lands in a low-Q basin:

```text
OFV 867.494
TVQ 4.3718
```

Evaluating the NONMEM final theta/omega/sigma diagonal values in ferx with NONMEM's final residual correlation (`rho ~= 0.934`) gives a much closer objective and the expected Q:

```text
OFV 739.527
TVQ 16.844
```

Evaluating the same NONMEM final theta/omega/sigma diagonal values with ferx's current fixed initial residual correlation (`rho = 0.20`) gives:

```text
OFV 905.538
TVQ 16.844
```

This shows the main mismatch is the frozen SIGMA off-diagonal, not that Q itself cannot be scored near the NONMEM optimum.

Removing `L2` from the data did not change the current ferx fit, so L2 grouping is not the driver here.

## Expected behavior

For `block_sigma`, ferx should be able to estimate off-diagonal residual covariance/correlation terms, not only the diagonal sigma SDs, so models with NONMEM `$SIGMA BLOCK(n)` can optimize the same likelihood surface.

## Implementation notes

Current behavior is documented in `docs/model-file/error-model.qmd`: the diagonal entries initialize sigma SDs and the off-diagonal entries define fixed residual correlations. The runtime representation is `ResidualCorrelation { sigma_i, sigma_j, rho }` in `src/types.rs`, which does not carry an estimated/free parameter slot.

Likely work:

- Extend parameterization to include free block-sigma off-diagonal coordinates.
- Preserve positive definiteness for the residual covariance/correlation block, probably via a Cholesky/correlation parameterization rather than unconstrained covariance.
- Update FOCE/FOCEI objective, analytic gradients, covariance output, YAML/fitrx output, docs, and tests.
- Add a regression/validation test using this fluconazole model or a smaller fixture where NONMEM estimates a nonzero residual covariance and Q depends on it.

## Acceptance criteria

- A `block_sigma` model can estimate off-diagonal residual covariance/correlation parameters.
- The fluconazole model's fitted Q moves toward the NONMEM trace value (`TVQ ~= 16.8`) when the full SIGMA block is estimated.
- Output reports the estimated full SIGMA block, not only diagonal sigma SDs.
- Docs clarify fixed vs estimated block-sigma syntax, if both modes remain supported.

roninsightrx:
## Why

`block_sigma` parsed its off-diagonal into a `ResidualCorrelation` that lived only on the
frozen `CompiledModel` and was never estimated, so a model with a `$SIGMA BLOCK(n)`
counterpart in NONMEM optimized a **different objective**. On the fluconazole RadboudUMC model
NONMEM moves the residual correlation from 0.20 to 0.9312 while ferx held the 0.20 init, which
is what #847 is about.

Closes #847.

This is a fresh implementation on current `main`, not a rebase of the two earlier attempts
(#868, #850). Both branch from 2026-07-15 and are ~660 commits behind: `src/api.rs` has since
been split into `src/api/`, test modules were extracted into sibling `*_tests.rs` files,
`residual_error.rs` became the sole owner of the residual-variance math, and #1100/#1101
already landed the `FitResult` + YAML half. `git merge-tree` reports
`CONFLICT (modify/delete): src/api.rs` on both. The *design* from those PRs is reused —
skeleton and validation from #868, the closed-form ρ derivatives from #850 — which is why this
lands in 1.5k insertions where they needed 2.0k / 2.5k.

## What changed

**Parameterization.** ρ becomes a packed optimizer coordinate as its Fisher-z transform
`z = atanh(ρ)`, bounded at `|z| ≤ 6`, so ρ stays strictly inside `(-1, 1)` for every step the
optimizer can take and the residual covariance can never go singular from the correlation
alone. The ρ block is packed **last** — after Ω_IOV and the mixture overrides — so every
offset the rest of the codebase derives from `n_theta`/`n_omega`/`n_sigma` (the covariance
step's `kappa_start`, the mixture segment, `omega_structural_zero_mask`'s `iov_start`) keeps
pointing at the same coordinate. `rho_packed_start` owns that offset; nothing re-derives it.

**Semantics.** A plain `block_sigma (...) = [...]` estimates the off-diagonal (NONMEM
`$SIGMA BLOCK(n)`); `... FIX` pins the whole block. See Breaking changes.

**Live ρ.** The FOCE/FOCEI objective and the inner EBE search read the correlations off the
frozen model, so a ρ the optimizer moved was invisible to them. `individual_nll_into_with_schedule`,
`foce_subject_nll`, `foce_population_nll` and `analytic_eta_nll_gradient_with_schedule` now
take the live `&[ResidualCorrelation]` next to the live σ.

**Closed-form ∂/∂ρ.** In `stats/residual_error.rs` — the module CLAUDE.md names sole owner of
this math — as `dvar_drho`. Every sigma loading is affine in `f`, so with `c` the value
loading and `c'` its slope:

```
dR_jj/drho_k = 2·s_i·s_j·c_i·c_j
dd_j /drho_k = 2·s_i·s_j·(c_i'·c_j + c_i·c_j')
```

which is exactly the ρ-derivative of `variance_at_scaled`'s cross term and of
`diag_self_deriv`'s, so the two cannot drift. The `MIN_VARIANCE` floor is mirrored
**asymmetrically on purpose**: `variance_at_scaled` clamps `R` while `diag_self_deriv` leaves
`d` unclamped, so a clamped row reports `∂R/∂ρ = 0` but keeps `∂d/∂ρ`. Without that the
gradient claims a live `∂R/∂ρ` on a row whose variance the objective has already pinned to the
floor — the #958 failure mode, on the correlation axis.

`rho_block` feeds those into `sigma_block`'s exact assembly (data + lnR, `log|H̃|`, EBE
response) for FOCEI; the FOCE Sheiner–Beal marginal is first-order in `R`, so ρ enters only
through `R⁰` at `f(η=0)`; `subject_eta_dx` gains the matching `dη̂/dρ` column. All three carry
the Fisher-z chain `dρ/dz = 1 − ρ²`.

**Reporting.** `FitResult` gains `residual_correlation_fixed` and `se_residual_correlations`
(natural ρ scale, delta method on the packed coordinate). The fit YAML's `block_sigma:` section
gains `correlation_se` and reports `correlation_fixed` from the model instead of hard-coded
`true`; `covariance_fixed` now requires all three factors of `ρ·σ_i·σ_j` to be pinned. `.fitrx`
round-trips all three, and a pre-#847 bundle loads with every correlation marked fixed.

### Bugs found on the way

- **`corr_residual_diag` / `corr_residual_rd_at_sigma` read the frozen model**, so `prep`'s
  `(r, d, d2)` — and hence the whole outer-gradient assembly — would have been built at the
  *declared* ρ even after the optimizer moved it. The FD parity harness `precise_ebe_corr` had
  the same read, which is why its reference was initially missing ρ's EBE-response term
  (analytic 5.963 vs FD 5.948 until both were fixed).
- **`simulate()`** built its dense `R` from the declared correlation with fitted sigmas, so a
  VPC of an estimated `block_sigma` would not reproduce the correlation the fit reported.

## Alternatives considered

Rebasing #868 or #850 — rejected, see Why. Their designs are reused.

**Threading live ρ through every estimator vs. rejecting the chain.** `fit()` forwards each
stage's fitted parameters to the next, so `[focei, imp]` would hand IMP a FOCEI-estimated ρ
that IMP has no channel to read: it would score the *declared* correlation while the result
reported the estimate. Threading live ρ down to SAEM/IS/IMPMAP/VI's data term means ten helper
signatures and their callers. This PR rejects the configuration instead
(`E_BLOCK_SIGMA_CHAIN_UNSUPPORTED`) and leaves the threading as a follow-up. `[saem, focei]` is
unaffected — ρ has not moved when SAEM runs.

## Cross-repo dependency

Both new `FitResult` fields are `#[serde(default, skip_serializing_if = ...)]`, so ferx-r's
reads stay backward-compatible. Surfacing `se_residual_correlations` in the R wrapper is an
enhancement, not a requirement.

## Breaking changes
- [x] `.ferx` model file format — a non-`FIX` `block_sigma` now **estimates** its off-diagonal
- [x] `FitResult` fields changed — two added, both serde-defaulted
- [x] Public Rust API — regenerated via `tools/update-public-api.sh`

<details>

The old behaviour was: diagonal SDs estimated, off-diagonal frozen. **That combination is no
longer expressible** — NONMEM has no form for it either, and it is precisely the mismatch this
issue is about. `FIX` reproduces the old *correlation* but also pins the SDs the old form
estimated, so it is not a drop-in way to reproduce an old fit's numbers; re-run and compare.
The shipped `examples/correlated_residual_combined.ferx` already uses `FIX` and is unaffected.

Public API additions and why each is needed:

| Item | Needed by |
|---|---|
| `ModelParameters::residual_correlations`, `::residual_correlation_fixed` | `unpack_params` must return the live ρ; every objective consumer reads it from the parameter vector |
| `FitResult::residual_correlation_fixed`, `::se_residual_correlations` | the output layer must distinguish an estimated from a fixed correlation, and report its SE |
| `residual_correlations` argument on `foce_subject_nll`, `foce_population_nll`, `individual_nll_into_with_schedule` | the live-ρ channel, parallel to the existing `sigma_values` |
| `PackedLayout::n_rho`, `::omega_iov_end` | VI's `total()` must equal `pack_params().len()` or its `x` cannot round-trip through `unpack_params` |

</details>

## Numerical validation
- [x] Compared against: **NONMEM** (`ferx-testdata/fluconazole_radboudumc/nm/fluconfinal014.lst`)
- [x] Reference values:

```
NONMEM fluconfinal014:  OFV 734.644   rho 0.9312   TVQ 16.8
```

```
NONMEM fluconfinal014:  OFV 734.644   rho 0.9312   TVQ 16.8
```

All rows below: fluconazole RadboudUMC, the model file's own `[fit_options]` and
`[data_selection]` (271 obs, FOCEI, covariance off), started from the model's **declared
inits** (rho init 0.20) — a cold start, not seeded from NONMEM.

| | packed params | OFV | converged | rho | sigma_unbound | sigma_total |
|---|---|---|---|---|---|---|
| main @ `7488e5e9` (rho frozen) | 13 | 871.5137 | yes | 0.2 (fixed) | 0.4583 | 0.4163 |
| this branch, rho **pinned** | 13 | **871.5137** | yes | 0.2 (fixed) | — | — |
| this branch, rho **free** | 14 | **736.8903** | yes | **0.9319** | 0.4557 | 0.4141 |
| NONMEM `fluconfinal014` | — | 734.644 | — | 0.9312 | 0.4506 | 0.4147 |

**rho lands on 0.9319 against NONMEM's 0.9312**, sigma on both endpoints to two decimals, and
the objective within 2.2 units — from a cold start, converged. That is #847's correlation
criterion. The pinned row reproduces `main` to every printed digit, which is the plumbing
check: with rho held, this branch is the old code.

`TVQ` reaches 11.2 against NONMEM's 16.8. That gap is the Q-collapse of open issue **#864**,
not this PR — `main` shows the same collapse (11.3) when started from NONMEM's own estimates,
with rho frozen at 0.9337.

Nothing else moves. `examples/warfarin.ferx` is byte-identical to main (−280.3640), and
`examples/correlated_residual_combined.ferx` — whose block carries `FIX` — also matches main
exactly (18.7274). Scoped by construction: `rho_packed_start` equals `packed_len` when a model
declares no correlations, so the one preconditioner change below is a no-op there.

### The one non-obvious fix this needed

Freeing rho initially made NLopt L-BFGS fail on its first step (OFV stuck at 1111.12, rho never
leaving 0.2). The cause was **not** the rho math — on this model `R` is genuinely dense
(cross-endpoint total/unbound pairing via `L2`), so `population_gradient_sens` returns `None`,
the whole population falls to reconverged FD, and `rho_block` never executes. The optimizer
trace showed the second iterate at OFV 1.3e17 with `TVBMAX` sitting exactly on its upper bound
— and `main` hits the same 1e17 blow-ups, so that pathology is pre-existing.

What was mine: the `abs` preconditioner divides each packed coordinate by its own `|value|`.
That is right for a log-space coordinate, where `|v|` is an order of magnitude. A Fisher-z
`z = atanh(rho)` is a *position* in a bounded range that passes through zero, so at
`rho = 0.2` (`z = 0.203`) the scaled box became ±30 while every other scaled coordinate spanned
single digits. The scale is now `max(|z|, 1)` — normalise a rho that has already grown past 1,
never divide by a value below it. Both halves are load-bearing: plain `compute_scale` fails at
1111.12, a flat `1.0` converges from the declared inits but stalls at init when seeded from
NONMEM's estimates, and `max(|z|, 1)` converges from both.

Per-coordinate: the existing analytic-vs-FD parity tests
(`population_packed_gradient_block_sigma_matches_fd` and its FOCE / Selected /
ExpressionScale siblings) now cover the ρ coordinate against Richardson reconverged FD of
ferx's own dense marginal. They were **red on the intermediate commit** (`analytic = 0.0` vs
`fd = 5.95`) and go green once `rho_block` lands — the oracle doing its job.

## Performance
- [x] No significant impact expected — ρ adds one packed coordinate and an O(n_obs) closed-form
      term per correlation per subject. No existing path changes cost.

## Tests
- [x] Unit tests added (`cargo test --lib --features ci`: 3894 passed, 0 failed)
- [x] Regression tests that fail without the fix

| Tier | Test |
|---|---|
| 1 | `dvar_drho_matches_fd_of_the_production_builders` — closed form vs central differences of the production `R` / `∂R/∂f` builders, **plus** hand-computed combined-error values so an FD-only check cannot pass on a shared mistake |
| 1 | `dvar_drho_zeroes_the_variance_term_at_the_floor`, `dvar_drho_is_zero_when_a_slot_is_not_loaded` |
| 1 | `test_pack_unpack_rho_round_trip`, `test_fixed_rho_is_pinned`, `test_rho_bounds_keep_correlation_admissible` (chain factor vs FD of `unpack_rho`), `test_rho_coordinate_kind_is_symmetric` |
| 1 | `test_se_residual_correlation_uses_the_trailing_rho_slot` — the template carries an IOV block **on purpose**, so an offset mistake lands on a kappa coordinate and shows |
| 1 | `population_packed_gradient_block_sigma_matches_fd` + FOCE / Selected / ExpressionScale, now including the ρ coordinate |
| 1 | YAML: fixed, estimated, and pre-#847 (flags absent) shapes |
| 2 | `fit_result_marks_a_bare_block_sigma_correlation_free` / `..._pinned`; the three `block_sigma_chain_*` guard tests |
| 3 | `dense_residual_free_rho_beats_fixed_rho` — ρ moves off init, stays inside `(-1, 1)`, free-ρ OFV ≤ fixed-ρ OFV (widening the feasible set cannot hurt), analytic ≤ FD |

## Example
- [x] Inline below

<details>
<summary>Example</summary>

```toml
[parameters]
  # rho estimated (NONMEM $SIGMA BLOCK semantics)
  block_sigma (PROP_ERR, ADD_ERR) = [0.04, 0.10, 1.00]

[error_model]
```

```yaml
block_sigma:
  ADD_ERR__PROP_ERR:
    covariance: 0.186600
    covariance_fixed: false
    correlation: 0.933000
    correlation_fixed: false
    correlation_se: 0.021400
```

</details>

## Docs
- [x] `docs/model-file/error-model.qmd` — new "Fixed vs. estimated `block_sigma`" section with
      the Fisher-z parameterisation, a migration callout, and the chain restriction. Splitting
      the page's one 7.4k-char terminal section into four bounded ones let its **R1 baseline
      entry be removed** rather than merely kept.
- [x] `CHANGELOG.md` `[Unreleased]` — Changed / Added / Fixed

## Checklist
- [x] `tools/preflight.sh` green — fmt, all 5 `cargo check` feature sets, clippy, rustdoc,
      docs-lint, public-API baseline. It earned its keep: three call sites reachable only under
      `slow-tests`, `markov` and `nn` were caught by it and not by the default test run.
- [x] `Dual2` sensitivity path stays differentiable — the ρ derivatives are `f64`-only closed
      forms; the generic `*_g<T: PkNum>` kernels are untouched.

## Reviewer hints

Focus:
- `stats/residual_error.rs::dvar_drho` — the closed forms and the deliberate floor asymmetry.
- `sens_outer_gradient.rs::rho_block` and the FOCE ρ term — they mirror `sigma_block` and the
  σ loop respectively; the parity tests are the oracle.
- The live-ρ threading, especially the `corr_residual_diag` fix (this was a real bug).
- `api/validation.rs` — the chain guard.

Skimmable: the ~50 mechanical struct-literal updates for the two new `ModelParameters` fields,
and the `obs_nll_subject_*` call sites that all pass `&model.residual_correlations`.

## Open questions

**Resolved** — this was a blocking question through most of this PR's history and the body
above now reflects the answer. Summary for anyone reading the commit trail: freeing rho made
NLopt L-BFGS fail on fluconazole; the cause was the `abs` preconditioner applied to a Fisher-z
coordinate, not the rho derivatives (which never run on that model). Fixed in
`fix(estimation): don't magnitude-scale the block_sigma rho coordinate below 1`.

**None open.** The chain restriction (`E_BLOCK_SIGMA_CHAIN_UNSUPPORTED`, rejecting a
rho-estimating stage followed by one that cannot read the estimate) is accepted as the trade
for this PR. Threading live rho through SAEM / IS / IMPMAP / VI so `[focei, imp]` keeps working
is tracked in **#1216**, with the ten helper signatures it touches enumerated there.

https://claude.ai/code/session_01J1QDyTu8Suqjz19d3uTeeE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,25 @@ section of the SDLC for the versioning policy).
   marked fixed, which is what it meant at the time.
 
 ### Fixed
+- **An estimated `block_sigma` correlation is bounded at `|rho| <= 0.995` (#847).** Merely
+  keeping rho inside `(-1, 1)` is not enough: a paired residual block's determinant carries a
+  factor `1 - rho^2`, so a rho of 0.9999 leaves `R` numerically singular and the likelihood
+  will chase `log|R| -> -inf`. On the 12-observation `examples/correlated_residual_combined`
+  fixture an unbounded rho ran to -0.99993 and reported convergence at a degenerate optimum. A
+  rho sitting on this rail is now a legible diagnostic: the two endpoints are carrying the same
+  noise.
+- **A zero `block_sigma` off-diagonal is now estimated rather than dropped (#847).**
+  `block_sigma (A, B) = [0.04, 0.0, 1.0]` — the natural translation of `$SIGMA BLOCK(2)` with a
+  zero covariance init — previously built no correlation at all, so under the new
+  estimate-by-default semantics it would have silently fitted a diagonal residual with no
+  coordinate to move. A `FIX`ed zero is still dropped: a fixed zero correlation is the same
+  object as no correlation.
+- **`method = laplace` / `n_agq > 1` with a free `block_sigma` now uses the reconverged-FD
+  outer gradient (#847).** AGQ's analytic score assembles theta / omega / sigma / omega_iov and
+  never writes the rho slot, while its objective does depend on rho — so the analytic gradient
+  is declined for a free correlation rather than handing the optimizer a hard zero there. A
+  `FIX`ed block keeps the analytic score. The chain guard was widened to match: `laplace` can
+  move rho, so `[laplace, imp]` is rejected exactly like `[focei, imp]`.
 - **The `block_sigma` correlation coordinate is no longer magnitude-scaled below 1 (#847).**
   The outer optimizer's `abs` preconditioner divides each packed coordinate by its own
   `|value|`, which is right for a log-space coordinate but meaningless for the Fisher-z
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,11 @@ section of the SDLC for the versioning policy).
   off-diagonal correlation are estimated. Previously the correlation was always frozen at the
   declared value, so a model with a `$SIGMA BLOCK` counterpart in NONMEM optimized a different
   objective — on the fluconazole RadboudUMC model NONMEM moved the residual correlation to
-  ~0.93 while ferx held the ~0.2 init. Append `FIX` to hold the whole block, which is the only
-  form that reproduces the previous behaviour; the shipped
-  `examples/correlated_residual_combined.ferx` already uses it and is unaffected. The
+  ~0.93 while ferx held the ~0.2 init. Append `FIX` to hold the whole block — SDs and
+  correlation alike, matching `$SIGMA BLOCK(n) FIX`. Note that the old behaviour (SDs
+  estimated, correlation frozen) is no longer expressible: NONMEM has no such form either, and
+  it was the mismatch this issue is about. The shipped
+  `examples/correlated_residual_combined.ferx` already uses `FIX` and is unaffected. The
   correlation is optimized as its Fisher-z transform `atanh(rho)`, so it stays strictly inside
   `(-1, 1)` and the residual covariance can never go singular from the correlation alone, and
   it rides the exact analytic FOCE/FOCEI outer gradient rather than falling back to finite
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,37 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Changed
+- **`block_sigma` now estimates its off-diagonal correlation (#847).** A plain
+  `block_sigma (...) = [...]` is NONMEM `$SIGMA BLOCK(n)`: its diagonal SDs *and* its
+  off-diagonal correlation are estimated. Previously the correlation was always frozen at the
+  declared value, so a model with a `$SIGMA BLOCK` counterpart in NONMEM optimized a different
+  objective — on the fluconazole RadboudUMC model NONMEM moved the residual correlation to
+  ~0.93 while ferx held the ~0.2 init. Append `FIX` to hold the whole block, which is the only
+  form that reproduces the previous behaviour; the shipped
+  `examples/correlated_residual_combined.ferx` already uses it and is unaffected. The
+  correlation is optimized as its Fisher-z transform `atanh(rho)`, so it stays strictly inside
+  `(-1, 1)` and the residual covariance can never go singular from the correlation alone, and
+  it rides the exact analytic FOCE/FOCEI outer gradient rather than falling back to finite
+  differences. Estimators that do not estimate it (SAEM, IMP/IMPMAP, Bayes, AGQ, VI) continue
+  to hold it at the declaration.
+
+### Added
+- **Fitted `block_sigma` correlations are reported with their fixedness and standard error
+  (#847).** `FitResult` gains `residual_correlation_fixed` and `se_residual_correlations` (the
+  SE on the natural `rho` scale, by the delta method on the packed Fisher-z coordinate), the
+  fit YAML's `block_sigma:` section gains `correlation_se` and now reports
+  `correlation_fixed` from the model rather than always `true`, and `.fitrx` bundles
+  round-trip all three. A bundle written before this change loads with every correlation
+  marked fixed, which is what it meant at the time.
+
 ### Fixed
+- **`block_sigma` residual derivatives were built at the declared correlation, not the live
+  one (#847).** The outer-gradient assembly (`corr_residual_diag` /
+  `corr_residual_rd_at_sigma`) and the inner eta-gradient read the correlations off the frozen
+  `CompiledModel`, so once the off-diagonal became estimable the whole `(R, dR/df, d2R/df2)`
+  chain would have been evaluated at the initial value while the objective moved. Both now
+  take the live parameter vector's correlations.
 - **`ode_method = auto` no longer keeps a stiff solve whose analytic derivatives have
   overflowed (#1204).** The escalation guard checked that every saved state was finite, but
   read *values* only. A dual number's derivative jets carry higher powers of what its value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,15 @@ section of the SDLC for the versioning policy).
   marked fixed, which is what it meant at the time.
 
 ### Fixed
+- **The `block_sigma` correlation coordinate is no longer magnitude-scaled below 1 (#847).**
+  The outer optimizer's `abs` preconditioner divides each packed coordinate by its own
+  `|value|`, which is right for a log-space coordinate but meaningless for the Fisher-z
+  `atanh(rho)` — that is a *position* in a bounded range which passes through zero. At the
+  common `rho = 0.2` init it handed the optimizer a coordinate with roughly thirty times the
+  scaled room of every other one, and NLopt L-BFGS failed on its first step. The scale is now
+  `max(|atanh rho|, 1)`. On the fluconazole RadboudUMC model a cold-start FOCEI fit goes from
+  failing at OFV 1111.12 to converging at **736.89 with rho = 0.9319**, against NONMEM's
+  734.644 / 0.9312. Models without a `block_sigma` are unaffected by construction.
 - **`simulate()` drew correlated residuals at the declared `block_sigma` correlation (#847).**
   The dense residual `R` it samples from used the live sigmas but the model's declared
   correlation, so a VPC or posterior-predictive check of a fit with an estimated off-diagonal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,10 @@ section of the SDLC for the versioning policy).
   `(-1, 1)` and the residual covariance can never go singular from the correlation alone, and
   it rides the exact analytic FOCE/FOCEI outer gradient rather than falling back to finite
   differences. Estimators that do not estimate it (SAEM, IMP/IMPMAP, Bayes, AGQ, VI) continue
-  to hold it at the declaration.
+  to hold it at the declaration. Because a method chain starts each stage from the previous
+  stage's estimates, a chain that runs `foce`/`focei` before a stage that cannot read the
+  estimated correlation is now rejected with `E_BLOCK_SIGMA_CHAIN_UNSUPPORTED` rather than
+  silently scoring the declared value — `[saem, focei]` is fine, `[focei, imp]` needs `FIX`.
 
 ### Added
 - **Fitted `block_sigma` correlations are reported with their fixedness and standard error
@@ -46,6 +49,10 @@ section of the SDLC for the versioning policy).
   marked fixed, which is what it meant at the time.
 
 ### Fixed
+- **`simulate()` drew correlated residuals at the declared `block_sigma` correlation (#847).**
+  The dense residual `R` it samples from used the live sigmas but the model's declared
+  correlation, so a VPC or posterior-predictive check of a fit with an estimated off-diagonal
+  would not reproduce the correlation the fit reported. It now uses the parameter vector's.
 - **`block_sigma` residual derivatives were built at the declared correlation, not the live
   one (#847).** The outer-gradient assembly (`corr_residual_diag` /
   `corr_residual_rd_at_sigma`) and the inner eta-gradient read the correlations off the frozen
--- a/api/ferx-core-public-api.txt
+++ b/api/ferx-core-public-api.txt
@@ -398,10 +398,12 @@ pub ferx_core::estimation::vi::elbo::ElboEval::neg_elbo: f64
 pub struct ferx_core::estimation::vi::elbo::PackedLayout
 pub ferx_core::estimation::vi::elbo::PackedLayout::n_omega: usize
 pub ferx_core::estimation::vi::elbo::PackedLayout::n_omega_iov: usize
+pub ferx_core::estimation::vi::elbo::PackedLayout::n_rho: usize
 pub ferx_core::estimation::vi::elbo::PackedLayout::n_sigma: usize
 pub ferx_core::estimation::vi::elbo::PackedLayout::n_theta: usize
 impl ferx_core::estimation::vi::elbo::PackedLayout
 pub fn ferx_core::estimation::vi::elbo::PackedLayout::new(template: &ferx_core::ModelParameters) -> Self
+pub fn ferx_core::estimation::vi::elbo::PackedLayout::omega_iov_end(&self) -> usize
 pub fn ferx_core::estimation::vi::elbo::PackedLayout::omega_iov_start(&self) -> usize
 pub fn ferx_core::estimation::vi::elbo::PackedLayout::omega_start(&self) -> usize
 pub fn ferx_core::estimation::vi::elbo::PackedLayout::sigma_start(&self) -> usize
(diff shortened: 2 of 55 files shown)


roninsightrx:
## Code review triage — 6 of 7 findings actioned, plus one the review didn't catch

All fixed in `fix(estimation): close the gaps code review found in the block_sigma rho work`.
CI is green on it, `codecov/patch` included.

### Confirmed and fixed

**1. AGQ/Laplace had no ρ block in its score.** Real, and the most serious of the batch.
`accumulate_fixed_eta_packed_gradient` assembles θ / Ω / σ / Ω_iov, and the fixed-b FD salvage
(`accumulate_fixed_b_packed_gradient_fd`) differences only the θ and σ coordinates — neither
writes the trailing ρ slot, while `Stack::nll_at` scores the objective *at*
`params.residual_correlations`. With a free ρ that hands the optimizer a hard zero for a
coordinate the objective genuinely depends on, so it can report convergence at a point that is
not stationary in ρ.

`agq::analytic_gradient_available` now declines a model with a free correlation, sending
AGQ/Laplace to `reconverged_fd_gradient`, which differences every free packed coordinate.
Slower and correct — the loud fallback CLAUDE.md asks a scope gap to take. A `FIX`ed block
keeps the analytic score. Both directions are pinned by tests. Adding the block properly is
folded into #1216.

Worth recording how nearly I dismissed this: my first check was to run `method = laplace` on
fluconazole, and ρ *moved* (0.2 → 0.672), which looks like a refutation. It isn't — that
model's `R` is dense, so the AGQ gradient returns `None` and the outer falls to reconverged FD
anyway. The bug only bites where the analytic score actually applies, i.e. a diagonal-`R`
`block_sigma`. The empirical check was measuring the wrong model.

**2. The chain guard used the wrong predicate.** Real. `estimates_rho` keyed on "has an
analytic ρ gradient" (`Foce | FoceI`) when the question is "can this stage move ρ". Packing is
method-independent, so `laplace` moves it too — through the FD fallback above.
`[laplace, imp]` is now rejected exactly like `[focei, imp]`.

**3. A zero off-diagonal was dropped before it could be estimated.** Real, and the sharpest
catch. `block_sigma (A, B) = [0.04, 0.0, 1.0]` is the natural translation of `$SIGMA BLOCK(2)`
with a zero covariance init, and the parser's `value != 0.0` guard built no correlation at all
— so under estimate-by-default it would silently fit a diagonal residual with no coordinate to
move and no diagnostic. Harmless before this PR (ρ was frozen at 0 regardless); a real feature
loss after it. Now kept when the block is free. A `FIX`ed zero is still dropped: a fixed zero
correlation is the same object as no correlation.

**6. `simulate`'s correlated draw could disagree with itself.** Fixed. The gate and the `R`
build now resolve one slice up front, falling back to the model's declaration when a
hand-built `ModelParameters` carries none — dropping a declared `block_sigma` into independent
draws is the worse of the two failures. The resolved slice is passed into
`emit_correlated_residual_rows` rather than re-read there, so gate and build cannot diverge.

**7. `subject_eta_dx` mixed declared and live.** Fixed — `correlated` was derived from `model`
while the values fed to `corr_residual_rd_at_sigma` came from `params`. Latent rather than
live (same length for parser-built models), but a real drift hazard.

**4. Duplicated `#[allow(clippy::too_many_arguments)]`.** Real, removed. One correction to the
finding: it would *not* have turned the Clippy job red — `duplicated_attributes` does not fire
at this clippy version, and the `Clippy` job was green on the push that contained it.

### Partly addressed

**5.** Two halves. The empty-vector reading is now explicit
(`rho_fixed.is_empty() && !model.residual_correlations.is_empty()`) with a comment tying it to
`packed_fixed_mask`'s own `unwrap_or(false)`, so the two cannot disagree about whether a
hand-built params packs ρ free.

The substantive half — that the guard reads `model.default_params` while `fit()` runs with a
caller-supplied `init_params` — is **not** fixed. `check_model_options(model, options)` has no
access to the caller's parameter vector, so closing it needs a signature change and a look at
every caller. Left as-is deliberately; flagging it here rather than burying it.

### Not in the review

Removing `FIX` from `examples/correlated_residual_combined.ferx` drives **ρ → −0.999931** under
FOCEI, with `converged: true` and OFV −8.38.

My bound was wrong. Strict positive-definiteness is not a strong enough constraint: a paired
residual block's determinant carries a factor `1 − ρ²`, so at the original `tanh(6) ≈ 0.999_988`
rail that factor is `2.5e-5`, `R⁻¹` blows up by `4e4`, and the likelihood is happy to chase
`log|R| → −∞`. The PR's own docs claimed the Fisher-z box "keeps R positive-definite" — true,
and beside the point.

`RHO_Z_BOUND` is now 3 (`|ρ| ≤ 0.995`, `1 − ρ² ≥ 9.9e-3`). The same fixture now stops at
−0.995_06, which reads as *these two endpoints are carrying the same noise* rather than as a
broken fit that looks converged. A test pins the `1 − ρ²` floor rather than just `|ρ| < 1`, so
the bound cannot be loosened back without failing.

Fluconazole is unaffected — ρ = 0.9319 is `z = 1.67`, well inside the tighter rail, and the fit
is unchanged at **736.8903 / ρ 0.931900**.

### Verification after the fixes

- `tools/preflight.sh` green (all five feature sets, clippy, rustdoc, docs-lint, API baseline)
- `cargo test --lib --features ci`: 3898 passed, 0 failed
- Slow convergence tests pass
- Fluconazole cold start unchanged: 736.8903, ρ 0.931900, converged
- `examples/warfarin.ferx` byte-identical to main (−280.3640); the `FIX`ed
  `correlated_residual_combined` also byte-identical (18.7274)
- Four new tests: AGQ routing in both directions, the widened chain guard, and both halves of
  the zero-off-diagonal semantics

Issue #1175, opened 2026-09-01
epic(tools): automated model development — covariate / structural / variability / residual-error search at Pharmpy AMD parity


roninsightrx:
Automated model development — automatic covariate, structural, variability and residual-error searches in `ferx-tools`, at Pharmpy AMD feature parity.

## Verdict

Target **Pharmpy-style stepwise search**, not pyDarwin's architecture.

pyDarwin's core machinery — a template file plus a tokens file, with string substitution into NONMEM control streams — exists to compensate for *not having a structured model object*. ferx has a parser and, since #1111, a `[covariate_model]` block explicitly designed so a search rewrites one block by line insert/delete. Cloning pyDarwin's template/token pair would be a step backwards.

What is worth taking from pyDarwin is its **penalized fitness** (per-θ/Ω/σ penalties, non-convergence penalty, condition-number and correlation penalties, crash value) and its **GA / exhaustive global search**, both of which sit on top of the same candidate-generation layer the stepwise tools need. The GP / RF / GBRT surrogates are low value here: ferx fits in-process and cheaply, so the surrogate rarely pays for itself.

## The one structural blocker

**ferx has no model-transformation API.** Text → `CompiledModel` is one-way; there is no `CompiledModel` → `.ferx` renderer and no typed edit operation.

Every search tool is *generate candidate model → fit it → rank it*. ferx can already do steps 2 and 3. It cannot do step 1. That layer is the whole project; the algorithms, ranking, caching and parallelism on top of it are small.

## What already exists

| Asset | Why it matters |
|---|---|
| `ferx-tools` + the #1114 boundary rule | A search is many `fit()` calls — it is a tool by definition. Home already exists. |
| `bootstrap/` (journal, manifest, resume, output, summary, rayon) | Exactly the orchestration skeleton a search needs, already written and tested. |
| `[covariate_model]` (#1111) | One relation per line, PsN `scm` forms with PsN's inits/bounds verbatim, a `none` state that round-trips, and the resolved relations returned on `FitResult.covariate_relations`. **covsearch is nearly free.** |
| In-process `fit()` | No NONMEM subprocess, no temp dirs, no grid manager, no `.lst` parsing — this removes pyDarwin's dominant cost and most of its failure modes. |
| `n_starts` / `start_sigma` multistart | Pharmpy's `retries` tool, already shipped in core. |
[...]

PR #1214, opened 2026-09-02
feat(api): `ferx-core::edit` — typed model transformation so a search can generate candidates [closes #1176]

### LINKED ISSUE #1176: feat(core): `ferx-core::edit` — a typed model-transformation API so a search can generate candidate models [#1175]
Part of #1175 (P0 — prerequisites). **This is the blocker for every other issue in the epic.**

## Problem

Text → `CompiledModel` is one-way. There is no `.ferx` renderer and no typed edit operation, so nothing can *generate* a candidate model. Every search tool is *generate candidate → fit → rank*; ferx can already fit and rank, and cannot generate.

## Proposal

A new **public** `ferx-core::edit` module. It must live in core rather than `ferx-tools`: ferx-r needs the same editing surface (the A4 reachability rule), and the module needs the parser.

```rust
pub struct ModelText { /* block-addressable .ferx source, spans preserved */ }

impl ModelText {
    pub fn parse(src: &str) -> Result<Self, String>;
    pub fn render(&self) -> String;              // byte-identical when unedited
    pub fn apply(&mut self, e: ModelEdit) -> Result<(), String>;
    pub fn canonical_hash(&self) -> [u8; 32];    // candidate identity / cache key
}

pub enum ModelEdit {
    SetStructural(StructuralSpec),          // pk template name + argument bindings
    AddCovariateRelation(Relation),         // one [covariate_model] line
    DropCovariateRelation { param: String, cov: String },
    AddIiv { param: String, form: IivForm },
    DropIiv { param: String },
    SetOmegaBlock(Vec<String>),             // block_omega (A, B) = [...]
    SetErrorModel(ErrorSpecText),
    SeedInits(&FitResult),                  // name-keyed theta/omega/sigma carry-over
    SetFitOption { key: String, value: String },
}
```

`SeedInits` is what Pharmpy calls `update_inits` between steps, and what `bootstrap` already does per replicate — it must key on `theta_names` / `eta_names` / `sigma_names`, not on position, because a candidate's parameter vector differs from its parent's.

## Cost by block

| Block | Edit shape | Difficulty |
|---|---|---|
| `[covariate_model]` | line insert / delete | **trivial** — the block was designed for this (#1111) |
| `[structural_model]` + `[parameters]` | template swap plus coupled θ add/drop (`two_cpt_oral` → `one_cpt_oral` drops `Q`, `V2` across three blocks) | medium |
| `[individual_parameters]` η surgery | remove `ETA_X` from an expression *and* its `omega` line | hard — expression editing, not line editing |

Mitigation for the hard case: require a search-eligible model to be written in canonical `P = TVP * exp(ETA_P)` form and reject anything else with a clear message. This is the same posture `[covariate_model]` already takes on a non-product right-hand side, so it is a rule users have already met.

## Acceptance

- [ ] `ModelText::parse` → `render` is **byte-identical** for every file in `examples/**`. This single test is what makes text surgery safe.
- [ ] Each `ModelEdit` variant has a unit test asserting the rendered text, and a test asserting the edited model **fits identically to the hand-written equivalent** — the pattern `tests/covariate_model_equivalence.rs` already uses.
- [ ] `canonical_hash` is stable under comment and whitespace changes and differs for every semantic change.
- [ ] Editing a model that violates the canonical-form requirement is a hard error naming the offending parameter, never a silent wrong edit.
- [ ] `api/ferx-core-public-api.txt` regenerated in the same PR, with the PR description stating which tool needs each added item.

roninsightrx:
## Why

Text → `CompiledModel` is one-way. ferx can already *fit* a candidate model and *rank* it, but nothing in the engine can **generate** one — and every model-space search tool (stepwise covariate selection, structural-model search, IIV/IOV search) is `generate candidate → fit → rank`. There is no `.ferx` renderer and no typed edit operation, so the generate step has no home.

This is #1176, the P0 prerequisite for the whole of #1175 — it blocks every other issue in that epic.

## What changed

A new public `ferx_core::edit` module. It lives in core rather than `ferx-tools` because ferx-r needs the same editing surface (the A4 reachability rule) and the module needs the parser.

**`ModelText` holds the source, not a syntax tree.** A `CompiledModel` is closures — it cannot be rendered back — and a second, renderable AST would be a parallel definition of the language that drifts from the one that actually compiles. So a candidate is *the source file*, edited in place. An unedited round-trip is byte-identical: comments, alignment, blank lines, CRLF and a missing final newline all survive. That is the property everything else rests on — anything an edit does not deliberately touch is *provably* unchanged, so a bad edit is a visible diff rather than a silent rewrite of the rest of the file.

**`ModelText::apply` takes one typed `ModelEdit`.** Every variant validates before it mutates, so a rejected edit leaves the text exactly as it found it and a search loop can propose, be refused, and move on.

| Variant | Notes |
|---|---|
| `SetStructural` | swaps the `pk` line **and** performs the coupled θ/η/expression edits across all three blocks. `two_cpt_oral` → `one_cpt_oral` drops `Q`, `V2`, `TVQ`, `TVV2`, `ETA_Q`, `ETA_V2` in one call |
| `AddCovariateRelation` / `DropCovariateRelation` | line insert / delete on `[covariate_model]`, plus any θ the drop orphans |
| `AddIiv` / `DropIiv` | the `exp(ETA_P)` factor and its `omega` declaration |
| `SetOmegaBlock` | diagonal `omega` declarations → one `block_omega`, off-diagonal seeded at ρ = 0.1 |
| `SetErrorModel` | the statement, with σ declarations **reordered** to match |
| `SeedInits(&FitResult)` | name-keyed carry-over of the parent's estimates |
| `SetFitOption` | one `[fit_options]` key |

Three details worth calling out:

- **Pruning is reachability, not occurrence counting.** `Q = TVQ` mentions `Q`, and a `[covariate_model]` line `Q ~ WT power(...)` mentions it again, so counting occurrences would keep an orphaned `Q` alive forever. The live set instead starts at the blocks that *consume* parameters and grows through the declarations those reach.
- **σ are consumed positionally.** A single-endpoint `[error_model]` checks its σ names against the `[parameters]` declaration *order*; it does not resolve them by name. So appending a new σ does not parse — `SetErrorModel` lays the declarations out in the order the statement names them. This one was found by the equivalence test, not by reading the parser, and it is the dangerous kind: for `combined`, an order that happened to parse would silently swap which σ is the proportional component.
- **η surgery requires canonical form.** Removing an η is expression editing, and there is no safe way to pattern-match an arbitrary RHS. A searchable parameter must be written `P = TVP * exp(ETA_P)`; anything else is a hard error naming the parameter and quoting its expression, never a silent wrong edit. Same posture `[covariate_model]` already takes on a non-product RHS (#1111), so it is a rule users have met.

`canonical_hash` is the candidate's identity — a fit cache key. Stable under comments, indentation, blank lines, line endings and spacing inside an expression; different for every token-level change. Line order and number spelling are deliberately *not* normalised: reordering `[parameters]` reorders the θ vector and is a different model, and `0.15` vs `1.5e-1` hashing apart costs one redundant fit whereas parsing every literal to compare them risks a false cache *hit*, which costs correctness.

## Alternatives considered

- **A renderable AST.** Rejected: it is a second definition of the `.ferx` language that would drift from the one that compiles, and it loses every comment and every alignment choice the author made.
- **`ModelText::parse` compiling the model.** Rejected: a model needing data-derived bindings (`center = median`, `levels = auto`) does not compile standalone, and it should still round-trip and edit like any other. `parse` is a *structural* parse — it finds block headers and nothing else. Callers compile `render()` when they want the candidate checked.
- **Engine-chosen defaults when widening a structural model.** Rejected: `StructuralSpec::new_parameters` makes the caller state the init and bounds for a parameter the template needs and the model lacks. A widening search has to choose those numbers; choosing them per candidate, visibly, beats an internal table nobody can see or override.

## Cross-repo dependency

No ferx-r change is required. The new surface is additive and nothing existing moved; a future ferx-r binding for model editing will need a `Cargo.lock` bump (`ferx-r/tools/update-ferx-core-lock.sh`) at that point, not now.

## Breaking changes
- [x] Public Rust API — baseline regenerated with `tools/update-public-api.sh` in this PR
- [ ] None

The diff is +84 lines, all under `ferx_core::edit`; nothing existing changed. Every added item is consumed by the model-search tooling of #1175, which cannot be written without it:

| Item | Who needs it |
|---|---|
| `ModelText` (`parse`, `render`, `apply`, `canonical_hash`, `canonical_form`, `block_names`, `block_lines`) | the candidate generator itself; `canonical_hash` is the fit cache key, `canonical_form` is how a cache miss is explained, and `block_names`/`block_lines` are how a search decides whether a model is eligible for a given move |
| `ModelEdit` and its payloads (`StructuralSpec`, `NewParameter`, `Relation`, `RelationTheta`, `IivForm`, `ErrorSpecText`, `ErrorForm`, `SigmaDecl`) | the move vocabulary a search enumerates over |
| `From<&CovariateRelation> for Relation` | re-proposing a relation read back off a parent fit, without re-deriving it from text |

The existing surface does not suffice: there was no renderer and no edit operation at all, so this is new ground rather than a widening of something already exposed. Nothing here is `#[doc(hidden)]`.

## Numerical validation

- [x] Compared against: **the hand-written equivalent model**, for every edit variant
- [x] Reference values: bit-for-bit `predict()` equality plus exact OFV equality (see below)

`tests/model_edit_equivalence.rs` asserts, for each variant, that the edited model and the model a pharmacometrician would have typed agree

- on `predict()` at a fixed parameter vector, **bit for bit**, and
- on the OFV after two outer iterations from that same start.

Bit-for-bit rather than to-tolerance is deliberate. An edit that produced an equivalent-but-differently-associated expression — `A * exp(η) * f` where the author wrote `A * f * exp(η)` — agrees to 1e-12, is still the wrong text to hand a user, and moves the SAEM mu-reference detector's answer (#619). Only equality catches it.

No NONMEM anchor is added, and per the "same-model equivalence" pattern of `tests/covariate_model_equivalence.rs` none is needed: every model this module can generate is a model the hand-written path could already state, and those forms are NONMEM-anchored already. The anchor for *this* feature is that the two spellings are the same model.

```
running 9 tests
test adding_a_covariate_relation_is_the_hand_written_block ... ok
test adding_an_eta_is_the_hand_written_model_with_it ... ok
test blocking_two_etas_is_the_hand_written_block_omega ... ok
test dropping_a_relation_restores_the_model_without_it ... ok
test dropping_an_eta_is_the_hand_written_model_without_it ... ok
test narrowing_back_to_one_compartment_restores_the_parent ... ok
test seeding_inits_starts_the_child_where_the_parent_finished ... ok
test switching_to_a_combined_error_model_is_the_hand_written_one ... ok
test widening_to_two_compartments_is_the_hand_written_two_compartment_model ... ok
test result: ok. 9 passed; 0 failed
```

`narrowing_back_to_one_compartment_restores_the_parent` is the one to read: it widens to two compartments and narrows back, and asserts both the same-model property *and* that the candidate's `canonical_hash` returns to the parent's — the property a stepwise search depends on when it backs out of a step.

## Performance

Nothing on the fit path changed. The module is line-oriented string work on files of ~100 lines.

## Tests

38 Tier-1 unit tests in `src/edit/mod_tests.rs` (rendered text for every variant, both error paths, the "a rejected edit leaves the text untouched" invariant), 9 Tier-2 equivalence tests, and 4 corpus tests in `tests/model_text_roundtrip.rs`:

- every `examples/**/*.ferx` round-trips **byte for byte**;
- every one keeps its canonical hash under a mechanical reflow (doubled indentation, a trailing comment on every line, blank lines between them);
- every one *changes* identity when a single code line is deleted — the sensitivity half of the contract, and the expensive direction of the error, since a missed change means the second candidate never gets fitted at all;
- two example models hash alike only when their canonical forms are identical. (`examples/one_cpt_infusion.ferx` and `examples/dose_rate.ferx` genuinely are the same model documented two different ways, and *should* collide — that is the cache hit the hash exists to produce.)

Full suite: `cargo test --lib` → 3920 passed, 0 failed.

## Example

<details>
<summary>Example</summary>

```rust
use ferx_core::edit::{ModelText, ModelEdit};

let mut text = ModelText::parse(&std::fs::read_to_string("parent.ferx")?)?;
text.apply(ModelEdit::DropIiv { param: "KA".into() })?;
text.apply(ModelEdit::SeedInits(&parent_fit))?;
let candidate = text.render();
```

Given a parent whose `[individual_parameters]` reads

```toml
[individual_parameters]
  CL = TVCL * exp(ETA_CL)
  V  = TVV * exp(ETA_V)
  KA = TVKA * exp(ETA_KA)
```

the candidate is the same file with `KA = TVKA`, the `omega ETA_KA ~ 0.20` line gone, and every remaining `theta`/`omega`/`sigma` init replaced by the parent's estimate — and *nothing else* changed, comments and alignment included.

Refusing rather than guessing:

```
ferx-core::edit: `CL = TVCL * (1 + ETA_CL) + 0.5` is not in the canonical form
`CL = TVP * exp(ETA_CL)`, so its η cannot be removed by rewriting the line.
Rewrite `CL` as a product ending in `exp(<eta>)`, or edit
[individual_parameters] by hand.
```

</details>

## Docs

- [x] `docs/` updated — new page `docs/api/model-editing.qmd`
- [x] New page linked in `docs/_quarto.yml` (under Rust API); `docs/_site/` not committed
- [x] `CHANGELOG.md` `[Unreleased]` entry added under `Added`
- [ ] No user-visible change

`cargo run -p docs-lint -- --check` is clean (no new baseline entries).

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — not touched; this module never runs during a fit
- [ ] `Cargo.toml` version bumped if breaking — additive, no bump

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key — none; this is a Rust API, the DSL is unchanged
- [ ] New data file — none

### ferx-book

### Example execution
- [x] No example execution step needed — no `.ferx` file, data file or CLI behaviour changed. The `examples/**` corpus is *read* by the new round-trip test (all 105 models pass) but none is modified.

## Reviewer hints

Worth the attention:

- **`prune_unreferenced` in `src/edit/apply.rs`.** The one piece of real reasoning here. Its guard against `[individual_parameters]` blocks containing `if`/`else` is conservative on purpose (an assignment inside a branch does not "declare" its name in the sense the pruner needs), and it protects the `THETA_<PARAM>_<COV>` family of a surviving relation by prefix, because those θ are *generated* by the desugar and never appear textually in the block.
- **`set_error_model`'s σ reordering.** See the note above — this is the subtle one.
- **`is_top_level_product` / `split_trailing_exp_eta`.** These are the canonical-form gate. `/` is in the reject list alongside `+` and `-` because `A / B * exp(η)` reassociates too.

Skimmable: `spec.rs` is plain data types and their one-line renderers; `canonical.rs` is thirty lines of whitespace normalisation.

## Open questions

1. **`SetOmegaBlock`'s seed correlation is a hard-coded 0.1**, matching Pharmpy's `create_joint_distribution`. Not zero, because a search blocks two ηs precisely to estimate their correlation. Should this be a parameter of the variant instead?
2. **`ModelEdit` carries a lifetime** because `SeedInits(&FitResult)` borrows, as the issue specifies. The alternative — an owned `ParameterSeed` with `From<&FitResult>` — keeps `ModelEdit` lifetime-free at the cost of one more type. I followed the issue; happy to switch if the lifetime proves annoying at the ferx-r boundary.
3. **`DropCovariateRelation` leaves an empty `[covariate_model]` block behind** rather than removing the header. That reads as honest to me ("the search tried and rejected everything", which is also what `CovariateForm::None` exists to say), but it is a choice.

https://claude.ai/code/session_01N95xdTSA2yJEeh1TD9ptwf


roninsightrx:
## Code review — `ferx_core::edit`

Reviewed the full diff plus the parser regexes, `FitResult` field semantics, and the covariate desugar naming this depends on. Four of these produce a silently wrong candidate model, which matters more than usual here: the whole point of the module is to feed a model-space search, and a candidate that carries a phantom random effect gets a wrong `n_parameters` and therefore a wrong BIC.

### 1. `src/edit/mod.rs:252` — `append_to_block` glues onto a file with no final newline

`append_block_header` explicitly terminates the previous line before pushing a header; the `Some(span)` path does not. When the target block is the *last* block in the file and the source has no trailing newline, `at == self.lines.len()` and the new line is pushed straight after an unterminated line:

```
  method = focei    +    "  seed = 1\n"    →    "  method = focei  seed = 1"
```

`[fit_options]` is very often the last block, so `SetFitOption` on a no-final-newline model emits a corrupt file. `render_preserves_crlf_and_a_missing_final_newline` only covers the *unedited* round trip. Fix: hoist the terminator guard out of `append_block_header` and run it whenever `at == lines.len()`.

### 2. `src/edit/apply.rs:215` — `add_iiv` misses an η that lives in a `block_omega`

The duplicate guard is `omega_decl_line(text, form.eta())`, and `OMEGA_RE` is anchored `^omega\s+`, so `block_omega (ETA_CL, ETA_V) = [...]` never matches. `AddIiv { param: "KA", form: Exponential { eta: "ETA_CL", .. } }` on such a model passes validation, rewrites `KA = TVKA * exp(ETA_CL)` and appends a *second* `omega ETA_CL ~ …` declaration. `drop_iiv` already calls `block_omega_line_for` for exactly this reason.

### 3. `src/edit/apply.rs:151` / `:756` — `prune_unreferenced` cannot drop an η held by a `block_omega`

`block_omega` lines are not in `decls`, so they are roots and every η they name stays live. Narrowing `two_cpt_oral → one_cpt_oral` on a model with `block_omega (ETA_CL, ETA_V2)` correctly deletes `V2 = TVV2 * exp(ETA_V2)` and `TVQ`/`TVV2`, but keeps the block — the candidate now carries an η no parameter uses. The parser only *warns* on that (`model_parser.rs:4048`, `parse_warnings`); it does not reject, contrary to the doc comment at line 746. Net effect is a flat omega direction plus an inflated `n_parameters` — a silently wrong BIC for a search that ranks on it.

### 4. `src/edit/apply.rs:761-764` — the `conditional` bail disables θ/η pruning too, contrary to its own comment

The comment says "leave that block's assignments alone; θ and η pruning still runs". It does not: when `conditional` is true the `[individual_parameters]` lines are omitted from `decls`, which makes them *roots*, so `TVQ` / `ETA_Q` mentioned on `Q = TVQ * exp(ETA_Q)` are live and never pruned. `SetStructural` narrowing on any model with an `if` in `[individual_parameters]` yields a candidate with an orphaned `Q`, an orphaned `TVQ` and a phantom η — same downstream damage as #3 — not the "conservative but θ/η still pruned" behaviour claimed.

### 5. `src/edit/apply.rs:358-364` — the float-exponent exemption in `is_top_level_product` misfires on identifiers ending in `e`/`E`

The guard accepts a top-level `-` whenever the preceding char is `e` or `E`, checking only one character back. `BASE-1`, `AGE-40`, `SIZE-1`, `TYPE-2` all satisfy it. So `CL = TVCL_BASE-1` is judged a top-level product and `AddIiv` writes `TVCL_BASE-1 * exp(ETA_CL)` — exactly the silent re-association the function exists to prevent, and precisely the case its error path was written for. Require the char before the `e`/`E` to be a digit or `.`, and the char after the `-` to be a digit.

### 6. `src/edit/apply.rs:699` — `sigma_decl_line` is dead code

Defined, never called anywhere in `src/`, `tests/` or `crates/`. It emits a `dead_code` warning; CI does not set `-D warnings`, so preflight stayed green. Delete it, or wire it into `set_error_model`'s slot lookup (which currently re-scans `slots` by hand).

### 7. `src/edit/apply.rs:442-444` — the "source order" comment in `set_omega_block` is not what the code does

The block is written at `min(lines)`, but the η are listed in `names` order and `variances` is built in `names` order. Passing `["ETA_V", "ETA_CL"]` for a model that declares `ETA_CL` first produces `block_omega (ETA_V, ETA_CL)`, which reorders the omega matrix — self-consistent, but not "the η keep their source order and therefore their index". Either sort `names` by declaration line or fix the comment.

### Minor, not filed above

- `set_error_model`'s `slots` filter uses the `^sigma` regex, so a model declaring σ via `block_sigma` gets its σ *appended* rather than reconciled.
- `SetFitOption` does not reject a `value` containing `#` or `//`, which `code_of` would then read back as a comment.

https://claude.ai/code/session_01N95xdTSA2yJEeh1TD9ptwf

roninsightrx:
All seven addressed in 9ebe4c51, plus the two minor ones, with a regression test each.

The one that needed a design call is #4. The suggested fix — excluding the conditional `[individual_parameters]` lines from the root scan while keeping them undeleted — over-prunes: a live `Q = TVQ * exp(ETA_Q)` inside a branch would lose `TVQ` and `ETA_Q` out from under it. So the assignments are now read as *declarations* (which is what keeps a live branch's θ and η alive, and is what makes the comment's claim true), and a **dead** one is refused naming it, rather than deleted — removing a line from inside a branch is a guess about the branch. That check can only run after the edit, so `ModelText::apply` now applies to a copy and commits on success; the "a rejected edit leaves the text untouched" guarantee holds for it too.

For #3 the block is rewritten rather than refused: narrowing away one η of a `block_omega` shrinks the triangle around the survivors, and a lone survivor goes back to the diagonal `omega` it came from. Refusing would have blocked an ordinary search step.

`preflight` green; 3930 lib tests, 9 `model_edit_equivalence`, 4 `model_text_roundtrip`.

roninsightrx (commented):
Found five correctness issues, all reproduced against head `9ebe4c51`:

1. **[P1] Multi-line `block_omega` declarations are invisible to the editor** (`src/edit/apply.rs:34-37`). The core parser explicitly coalesces supported multi-line declarations, and `examples/warfarin_frem.ferx` uses one, but every new helper applies `BLOCK_OMEGA_RE` to one physical line. In a valid two-compartment repro, `SetStructural(one_cpt_iv)` returns `Ok`, removes `Q`/`V2`, and leaves their multi-line omega block behind; the result still compiles with `n_eta == 4` instead of 2, producing the flat directions and wrong parameter count/BIC this pruner is intended to prevent. The same mismatch means `SeedInits` skips such blocks and the add/drop duplicate guards do not see their eta names. Please make block-omega lookup operate on logical declaration spans (including continuation lines) and add edit/seed regressions using the shipped multi-line form.

2. **[P1] `DropIiv` deletes an omega that another parameter still uses** (`src/edit/apply.rs:294-295`). A valid model can have both `CL = TVCL * exp(ETA_SIZE)` and `V = TVV * exp(ETA_SIZE)`. Dropping IIV from `CL` currently removes `omega ETA_SIZE`, leaving the unchanged `V` expression with an undeclared eta and making the candidate invalid. After rewriting the selected assignment, retain the declaration while any other expression references it (or reject the edit explicitly).

3. **[P1] `canonical_hash` has false hits for quoted values** (`src/edit/canonical.rs:48-66`). `squeeze` removes whitespace without tracking quotes, so `[data] path = "cohort / A.csv"` and `path = "cohort/A.csv"` produce the same canonical form/hash even though the parser resolves different dataset paths. This violates the acceptance requirement that semantic changes differ and can make the fit cache return a result for the wrong input. Canonicalization needs to preserve bytes inside quoted strings (or otherwise tokenize them separately).

4. **[P1] `SetOmegaBlock` silently unfixes fixed random effects** (`src/edit/apply.rs:455-471`). Two declarations such as `omega ETA_A ~ 0.1 FIX` and `omega ETA_B ~ 0.2 FIX` become a `block_omega (...) = [...]` with no `FIX`, changing previously fixed covariance parameters into estimated ones. Please carry `FIX` when all selected declarations are fixed and reject/document the mixed fixed/free case, which cannot be represented by one block-level flag.

5. **[P2] Covariate-theta prefix pruning confuses prefix-related covariate names** (`src/edit/apply.rs:893-899`). With live `CL ~ WT` and dropped `CL ~ WT2`, raw `starts_with("THETA_CL_WT")` keeps the now-orphaned `THETA_CL_WT2`, inflating the candidate parameter count. The generated family match should be exact or start with `THETA_CL_WT_`, not any raw prefix.

Validation performed: `cargo test --lib` passed (3930 passed, 23 ignored), and `cargo test --test model_text_roundtrip --test model_edit_equivalence` passed (13 tests). Five focused regression tests for the cases above fail on the PR head.

roninsightrx:
All five fixed in `0ef62d77`, with eight regression tests.

**1 (multi-line `block_omega`)** was the root of its own family, so it is fixed at the representation rather than per call site: `ModelText::logical_lines` now yields the declarations the *parser* sees — bracketed statements rejoined the way `join_bracketed_lines` rejoins them, a bare `FIX` folded onto the block before it, and operator continuations folded the way `join_continuation_lines` folds them — and `set_span` rewrites one as one, dropping its continuation lines. Every `[parameters]`, `[individual_parameters]`, `[covariate_model]`, `[structural_model]` and `[error_model]` lookup goes through it, so the duplicate guards, `SeedInits`, the σ reconciliation and the pruner all address the same unit. `block_lines()` returns the folded view too, which is what its doc comment already claimed. Regressions cover the edit and seed paths on a multi-line block, plus one that reads `examples/warfarin_frem.ferx` itself and asserts its seven-line block comes back as a single declaration with all five η and the last triangle row.

**2** — the declaration now goes only when nothing else reads the η, checked on the rewritten text with the outgoing lines skipped. Test covers both halves: `CL` drops and `omega ETA_SIZE` stays for `V`; `V` drops and it goes.

**3** — `squeeze` copies a quoted run through byte for byte. Spacing outside the quotes still normalises, so the reflow property is unchanged.

**4** — `FIX` is carried onto the block when every declaration it replaces is fixed, and a mixed set is refused naming the free η.

**5** — the family is matched exactly or on a `_` suffix boundary, so `THETA_CL_WT2` dies with its own relation while `THETA_CL_WT` lives.

`preflight` green; 3937 lib tests, 9 `model_edit_equivalence`, 5 `model_text_roundtrip`.

Issue #977, opened 2026-08-17
Mixture models ($MIXTURE): discrete latent subpopulations with covariate-dependent mixing


roninsightrx:
## Goal

Support NONMEM `$MIXTURE`-style discrete latent subpopulations: each subject belongs to one of K unobserved classes with mixing probabilities `p_ik`. Class-conditional parameters are estimated jointly with the mixing probabilities, and each subject receives a posterior class membership (NONMEM `MIXEST`/`PMIX` analogue).

FeRx has **no** mixture-model support today. (The only "mixture" in the codebase is the unrelated defensive-mixture IMP/IMPMAP importance-sampling proposal.)

## Locked scope (design decisions)

| Decision | Choice |
|---|---|
| **Mixing probabilities** | Covariate-dependent `p(k)` (full): per-subject expression over theta + covariates, softmax/logistic-normalized. Constant `p(k)` is the degenerate case. Matches NONMEM `$MIX` (arbitrary abbreviated code over `THETA` + data items). |
| **What differs per class** | Fully separate param sets: independent θ (via `MIXNUM` branching, user picks which are shared), Ω, Σ (via per-class declaration). |
| **Estimators (v1)** | FOCE/FOCEI first. Other methods (SAEM/IMP/Laplace/AGQ/Bayes) error clearly until wired. |
| **DSL** | New `[mixture]` block + reserved read-only `MIXNUM` index usable in `[individual_parameters]`. |

**Level reconciliation:** `MIXNUM` `if/else` in `[individual_parameters]` handles **θ** differences (and lets the user share what they want). Ω and Σ are population params, not `[individual_parameters]` expressions, so per-class Ω/Σ needs explicit per-class declaration in `[mixture]`. The plan supports the fully-separate case that way.

## Math

Per-subject marginal, class-conditional Laplace:

```
p_ik = softmax_k( g_k(θ, covariates_i) )          Σ_k p_ik = 1
L_i  = Σ_k  p_ik · L_ik(θ_k, Ω_k, Σ_k)
OFV  = −2 Σ_i log Σ_k p_ik · exp(−nll_ik)          ← log-sum-exp for stability
```

[...]

Issue #740, opened 2026-07-08
feat(estimation): left-truncation (delayed entry) support for clock-reset RTTE


TeunP:
## Problem

Clock-reset (gap-time) RTTE now **rejects** left-truncated data (`entry_time > 0`) at the fit boundary (`api::check_rtte_records`), added in the Slice 3.2 review. This issue is the real **support**.

Left truncation (delayed entry) is well-defined for **clock-forward** RTTE — `rtte_forward_nll_from_curves` seeds the running lower limit from `H(entry_time)`, so the first record contributes `H(t) − H(entry)` (conditional on survival to entry), and there is a test (`rtte_forward_left_truncation_uses_first_entry`).

For **clock-reset**, the first gap's convention is genuinely ambiguous and unratified:

- **A — renewal-from-entry (current NLL behavior, now guarded off):** the renewal clock restarts at entry, first gap `Δ = time − entry`, contribution `H(Δ) − log h(Δ)` from 0.
- **B — clock-from-0, conditioned on survival to entry:** contribution `H(t) − H(entry) − log h(t)` in absolute time (the forward-style conditioning).

They **coincide for a memoryless (exponential) hazard** but **differ for Weibull/Gompertz** — e.g. Weibull(scale 10, shape 2), enter at 8, first event at 10: A ≈ 3.26 vs B ≈ 1.97 → materially different estimates. There is **no NONMEM/survreg anchor** for either, so `rtte_reset_nll_from_curves` would have silently picked A without validation.

## Decision needed

Pick the intended convention for a gap-time renewal process with delayed entry (A or B — this is a modeling call about what "entry" means for the recurrent clock), then:

1. Implement it in `rtte_reset_nll_from_curves` (only the first-gap term changes).
2. Add a Tier-1 unit test pinning the closed form, plus a Tier-3 convergence test against an anchor.
3. Build a left-truncated reset reference dataset (`TENTRY > 0`) with a **survreg / NONMEM** anchor for the chosen convention.
4. Lift the fail-loud guard in `check_rtte_records` (the `entry_time > 0` reset arm) and the doc bullet.

## Context

- Guard + `rtte_reset_left_truncation_is_rejected` test landed on the Slice 3.2 branch (PR #720).
- Surfaced by the `/review` of #720; deferred deliberately rather than baking in an unratified convention.

PR #1215, opened 2026-09-02
feat(estimation): L2 + smoothness regularization for the covariate NN (DCM)

dominic-irx:
## Why

An unregularized DCM invents covariate-driven variation that is not in the data, and the amount scales with network capacity (`[4]` < `[4,4]` ≈ `[8]`), showing up as sharp high-frequency wiggles in the learned covariate→modulator curves. On a null covariate structure there is currently no way to tell the fit "prefer the flat explanation" — the `[covariate_nn]` weights are ordinary thetas in the outer parameter vector with no penalty anywhere.

This adds the two standard levers: shrink the weights (L2), and punish wiggles specifically (curvature).

How badly the unregularized fit overfits on the test fixture — 30 subjects, `[4]` hidden layer, no true covariate effect:

```
        ‖W‖² (weights only)   CL modulator variance across subjects
λ=0            2158.97                      481.43
λ=5               0.004                       0.000
λ=100             0.003                       0.000
```

## What changed

Two optional `[fit_options]` keys, both non-negative and defaulting to `0.0`:

- **`nn_l2`** — weight decay, `nn_l2 · Σ wᵢ²` over the **weight matrices only**; biases stay free, so the network can still move its output level without paying for it.
- **`nn_smooth`** — curvature penalty, `nn_smooth · Σ C²` where `C = f(x+h) − 2f(x) + f(x−h)` is the finite-difference 2nd derivative of each output along an input's marginal partial-dependence curve (z-scored input space, other inputs at their median). Penalizing the **2nd** derivative rather than the 1st leaves monotone slopes free and charges only for the wiggles.

Both are added to the objective the optimizer minimizes, with matching analytic gradients across all the FOCE/FOCEI outer optimizers (NLopt bobyqa/slsqp/lbfgs/mma + global pre-search, built-in BFGS, trust-region). The smoothness gradient reuses the MLP's existing analytic Jacobian — `∂C/∂w = J(x+h) − 2J(x) + J(x−h)` — so no new autodiff. NN weights are identity-packed with unit optimizer scale, so the raw-weight gradient maps 1:1 into packed space.

**λ = 0 is a strict no-op**, not an approximate one: the regularizer reports inactive, contributes exactly `0.0` to the objective and exactly `0.0` to every gradient coordinate. Existing fits stay byte-identical, which is asserted at both the unit and the fit level.

**The reported `ofv` (and hence AIC/BIC) stays the unpenalized −2·log-likelihood.** The penalty shapes the optimization only, so a regularized DCM remains directly comparable to an analytic-covariate model on information criteria.

```
[fit_options]
  method    = focei
  nn_l2     = 1e-2
  nn_smooth = 1e-1
```

Helpers live next to `forward`/`jacobian` in `src/nn/mod.rs` (`l2_weight_penalty`, `curvature_penalty`, `CurvatureGrid`, `NnRegularizer`).

## Alternatives considered

- **Penalizing the 1st derivative** rather than the 2nd. Rejected: it charges for a genuine monotone covariate effect, which is the thing a DCM is supposed to find. Curvature charges only for the wiggles.
- **Penalizing biases along with weights.** Rejected: the output-layer bias is what sets each PK parameter's level, so shrinking it fights the `init` mechanism and drags every parameter toward `softplus(0)`.
- **Penalizing on a grid in raw covariate space.** The grid is built in z-scored space with a uniform step, which is the space the network actually sees and keeps the step meaningful across inputs on different scales.
- **Reporting the penalized OFV.** Rejected: it would silently invalidate every DCM-vs-analytic AIC/BIC comparison.

## Cross-repo dependency

`nn_l2` / `nn_smooth` are plain `apply_fit_option` match arms with no `#[cfg(feature = "nn")]` gate, and ferx-r's `settings` is a generic passthrough to that same function with neither key on its `RESERVED` list. So `ferx_fit(settings = list(nn_l2 = ..., nn_smooth = ...))` works against this branch with no R-side change.

## Breaking changes
- [ ] None

Additive only — nothing existing changed shape. The baseline grows by exactly four lines, all the same two fields (the `types::` path and the crate-root re-export):

```
+pub ferx_core::types::FitOptions::nn_l2_lambda: f64
+pub ferx_core::types::FitOptions::nn_smooth_lambda: f64
+pub ferx_core::FitOptions::nn_l2_lambda: f64
+pub ferx_core::FitOptions::nn_smooth_lambda: f64
```

**Who needs them:** ferx-r, which sets both through `ferx_fit(settings = ...)` → `apply_fit_option`. There is no existing knob that does this — they are new fit options, and `FitOptions` is how a caller expresses one.

Everything else the feature added is `pub(crate)`: `NnRegularizer` (+ `build`/`is_active`/`penalty_value`/`add_packed_gradient`), `CurvatureGrid` (+ `build`/`is_empty`), and five `MlpMapper` penalty helpers. The estimation layer is their only caller. An earlier revision of this branch had them `pub` because the integration test reached for them — the "public because a test needs it" case A3/A6 rejects — so the tests that need them moved into `src/nn/mod.rs` instead and the surface came back down.

## Numerical validation

**There is no NONMEM anchor for this object, and there cannot be one.** NONMEM has no covariate neural network, so there is no equivalent run to compare a weight-decay or curvature penalty against — the same situation as the documented exceptions for feedback dosing (#391) and recurrent left truncation (#740). Validation is by the substitutes that do apply:

**(a) Analytic-vs-FD gradient parity**, per kernel and end-to-end. Both penalties are on the gradient path, so each is asserted against central finite differences of its own value function (`l2_penalty_gradient_matches_central_fd`, `curvature_penalty_gradient_matches_central_fd`), and again at the whole-`NnRegularizer` level over every NN-weight coordinate (`regularizer_penalty_gradient_matches_fd`). The latter also asserts that non-NN coordinates — omegas, sigma — receive exactly zero penalty gradient.

**(b) λ = 0 no-op, exactly.** `regularizer_lambda_zero_is_noop` pins penalty and gradient contribution at exactly `0.0`; `zero_lambda_fit_matches_baseline` runs two short fits and asserts byte-identical OFV *and* byte-identical theta.

**(c) Closed forms.** `curvature_penalty_zero_for_affine_network` — an affine network has zero 2nd derivative, so the curvature penalty must be exactly 0 regardless of λ. `l2_penalty_value_sums_weights_not_biases` — inflating a bias must not move the L2 penalty.

**(d) The effect itself**, `l2_shrinks_weights_and_modulator_variation` (Tier 3):

```
weight ‖W‖²:      λ=0 2158.97168   λ=5 0.00417   λ=100 0.00277
CL modulator var: λ=0  481.425997  λ=5 0.000000  λ=100 0.000000
```

The oracle is deliberately non-degenerate on both sides: the test asserts the λ=0 fit produces *real* spurious spread (`v0 > 1.0`) before asserting the regularized fits are flat. Without that first assertion the flatness check passes against a baseline that was already flat and proves nothing — which is what an earlier revision of this test was doing. See Reviewer hints.

## Performance

At λ = 0 the regularizer short-circuits on an `is_active()` check, so unregularized fits do no extra work. When active, the cost is one MLP forward/Jacobian per grid point per objective evaluation on a grid built once before the optimize loop — negligible against the per-subject inner loop.

## Tests

13 tests: 8 kernel-level + 3 regularizer-level in `src/nn/mod.rs`, 2 public-API in `tests/nn_regularization.rs`, plus parser coverage for the two keys. Tier 3 fits are `slow-tests`-gated per CLAUDE.md.

Both keys are documented with a worked snippet in `docs/model-file/covariate-nn.qmd` and `docs/model-file/fit-options.qmd`. They are two scalars on an existing block rather than new DSL shape, so no new `examples/*.ferx` — `examples/warfarin_dcm.ferx` still covers the block itself.

## Docs
- [x] New pages linked in `docs/_quarto.yml` (do **not** commit `docs/_site/`)
- [x] `CHANGELOG.md` `[Unreleased]` entry added (user-facing change), or N/A
- [ ] No user-visible change

`docs/model-file/fit-options.qmd` gains the two key descriptions; `docs/model-file/covariate-nn.qmd` gains a Regularization section explaining what each penalty targets and why the reported OFV stays unpenalized. No new pages, so `_quarto.yml` is unchanged.

## Checklist

Full run on this branch: `preflight OK — fmt check clippy rustdoc docs public-api`; `cargo test --no-default-features --features ci,markov,nn` → **4987 passed, 0 failed**; under `slow-tests`, `nn::regulariz*` 11/11 and `nn_regularization` 2/2.

### ferx-site

Two new `[fit_options]` keys, so a ferx-site `model-dsl` update is owed. Flagging rather than silently ticking — happy to open that PR, or leave it to whoever owns the site.

### ferx-book

Same — the DCM chapter would want the Regularization section. Not opened yet.

### Example execution

No existing example sets either key, and both default to off — every `examples/*.ferx` takes the byte-identical λ = 0 path, which `zero_lambda_fit_matches_baseline` asserts directly.

## Reviewer hints

**Where to focus — `src/estimation/outer_optimizer.rs`.** This branch was written in July and rebased across ~560 commits, and the interesting conflict was #977's mixture-model work rewriting how OFV and gradient are assembled. The penalty now shadows `raw_ofv` *after* the mixture branch, and `add_packed_gradient` applies to both the mixture and classical gradient arms. Worth a second pair of eyes on the mixture arm specifically — it's the path least covered by the DCM tests, since a mixture DCM isn't in the fixture set.

**Read `l2_shrinks_weights_and_modulator_variation`'s fixture comment.** The `center`/`scale` on that model is load-bearing. Without it, raw `WT` (45–90) and `CRCL` (46–82) drive layer-1 pre-activations to ±170, `tanh` clips at ±1 for every subject, and the network is constant *by saturation* before fitting starts. The λ=0 baseline then measures a modulator variance of 0 — not because L2 was unnecessary but because the network was never able to vary — and "heavy L2 keeps the modulator flat" becomes a comparison of one numerically-zero value against another. It passed that way pre-rebase only because the λ=100 optimum happened to land deep enough in saturation. z-scoring fixes the fixture, and the `v0 > 1.0` assertion stops it regressing.

**Skimmable:** `src/nn/mod.rs` is +808 but the great majority is tests. The production code is `weight_param_indices`, the two penalty pairs (value + gradient), `CurvatureGrid::build`, and `NnRegularizer`.

**Commit structure:** three commits, reviewable in order — the feature, the rebase carry (CI/rustdoc/API-baseline drift + the fixture fix), and the API narrowing.

## Open questions

1. **Default grid step for `nn_smooth`.** `NN_SMOOTH_GRID_STEP_Z` is a fixed step in z-scored space. It has no principled derivation beyond "small enough for the 2nd difference to mean something, large enough to clear FP noise". If someone has a view on exposing it, or on picking it from the observed input spread, I'll take it.

2. **λ scale is not normalized by network size.** `nn_l2 = 1e-2` means something different on `[4]` than on `[8,8]` because `Σ wᵢ²` runs over more terms. Dividing by `n_weights` would make λ portable across architectures, at the cost of changing what the number means. Left un-normalized for now; easy to revisit before anyone depends on specific values.

3. **No guidance yet on choosing λ.** The docs describe what each penalty does but not how to pick a value — that really wants a cross-validation or information-criterion story, which is a `ferx-tools` concern (many fits) rather than a core one. Worth an issue?


src/nn/mod.rs
```diff
@@ -909,6 +1036,255 @@ pub struct CovariateNn {
...
+        model: &CompiledModel,
+        population: &Population,
+        options: &FitOptions,
+    ) -> Self {
+        let l2_lambda = options.nn_l2_lambda;
+        let smooth_lambda = options.nn_smooth_lambda;
+        let mut specs = Vec::new();
+        if l2_lambda > 0.0 || smooth_lambda > 0.0 {
+            for nn in &model.covariate_nns {
+                let mapper = nn.mapper.mlp().clone();
```
roninsightrx: [P1] Evaluate curvature in the same normalized coordinate system as the fitted network. `collect_input_values`/`CurvatureGrid` produce raw covariate vectors, but this stores only `nn.mapper.mlp()` and `curvature_penalty*` feeds those raw vectors straight to `MlpMapper::forward`/`jacobian`. Production prediction goes through `NamedMlpMapper::build_input_vec_zero_fill`, applying `(x-center)/scale`. Thus a normalized model such as this PR's own fixture evaluates WT=70 here where the fitted MLP evaluates WT=0; with tanh this commonly saturates the penalty path and makes `nn_smooth` measure a different function (often nearly zero). Please retain/apply the `NamedMlpMapper` normalization, or build the triples directly in the actual MLP input coordinates, and add a test where non-identity `center`/`scale` distinguishes the two paths.

src/nn/mod.rs
```diff
@@ -909,6 +1036,255 @@ pub struct CovariateNn {
...
+/// distribution used to build the grid.
+fn collect_input_values(mapper: &NamedMlpMapper, population: &Population) -> Vec<Vec<f64>> {
+    mapper
+        .input_names()
+        .iter()
+        .map(|name| {
+            population
+                .subjects
+                .iter()
+                .filter_map(|s| s.covariates.get(name).copied())
```
roninsightrx: [P1] Include time-varying covariate snapshots in the smoothness grid. `Subject::covariates` is only the first non-missing representative value; supported DCM fits can read later values from `dose_covariates`, `obs_covariates`, `pk_only_covariates`, and `reset_covariates`. If every subject starts at the same value but changes later, this code sees zero spread and disables the curvature axis entirely, even though the fitted NN traverses a non-constant range. More generally it leaves later observed regions unregularized. Please collect the snapshots actually consumed by the model (or reject/document this combination) and cover the constant-baseline/time-varying case.

src/parser/model_parser.rs
```diff
@@ -7680,6 +7680,12 @@ pub fn apply_fit_option(opts: &mut FitOptions, key: &str, value: &str) -> Result
         }
         "vi_seed" => opts.vi_seed = parse_u64_opt("vi_seed")?,
         "gn_lambda" => opts.gn_lambda = parse_f64("gn_lambda")?,
+        // Covariate-NN (DCM) regularization strengths. Non-negative; 0 = off
+        // (strict no-op — existing fits stay byte-identical). Settable from the
+        // `.ferx` [fit_options] block and from `ferx_fit(settings = list(...))`
+        // via this same path. See FitOptions::nn_l2_lambda / nn_smooth_lambda.
+        "nn_l2" => opts.nn_l2_lambda = parse_f64_min("nn_l2", 0.0)?,
```
roninsightrx, line 7687: [P2] Register these keys as applicable fit options. `apply_fit_option` appends `nn_l2`/`nn_smooth` to `user_set_keys`, but neither `framework_keys()` nor the FOCE/FOCEI entries in `method_specific_keys()` contains them. Consequently every documented use emits `fit option nn_l2 is not used by method focei and will be ignored` even though the optimizer does use it. Please add the keys to the appropriate applicability lists and make the parser/integration test assert that no unsupported-key warning is produced.

src/estimation/outer_optimizer.rs
```diff
@@ -1666,6 +1685,10 @@ fn optimize_nlopt_once(
             let nll = pop_nll_opts(model, population, &params, &ehs, &hms, &kappas, options);
             (ehs, hms, ebe_stats, kappas, 2.0 * nll)
         };
+        // Penalized objective fed to the optimizer (unregularized fits unchanged).
+        // The matching gradient term is added into `grad_raw` below.
+        #[cfg(feature = "nn")]
+        let raw_ofv = raw_ofv + nn_reg.penalty_value(&params.theta);
```
roninsightrx: [P1] Keep all internal convergence and winner-selection comparisons on this penalized objective. After this shadowing, `best_seen_ofv` and the plateau tracker hold `raw OFV + penalty`, but the cold-restart `final_ofv` at line 2116 deliberately holds raw OFV and is passed to `failure_is_converged_plateau` at line 2200. A large penalty can therefore hide an arbitrarily worse cold-restart likelihood and falsely classify a warm-start artifact as converged. The same split appears in `resolve_stall_retry` (`result.ofv`) and multi-start ranking (`FitResult.ofv` in `src/api/fit.rs:747`), so those paths can discard the lower penalized solution and return a different objective's winner. Please carry a separate internal penalized score through these comparisons while continuing to expose the unpenalized `FitResult.ofv`, and add regression tests for plateau consistency plus multi-start/retry ranking.

roninsightrx (changes requested):
I found four correctness issues that should be resolved before merge. The focused regularization kernels, public integration smoke test, and parser test all pass at this head, and CI is green; the gaps are in production paths those tests do not exercise. GitHub also currently reports the branch as conflicting with `main` (the merge-tree conflict is in `CHANGELOG.md`).

src/nn/mod.rs
roninsightrx: **[bug] Smoothness penalty bypasses `center`/`scale` — evaluates the net at inputs the fit never sees.**

`NnRegularizer::build` clones the *inner* `MlpMapper` here, `CurvatureGrid::raw_at` emits raw covariate values (`z·sd + mean`, held inputs at raw medians), and `curvature_penalty{,_value}` call `self.forward(&t[k], weights)` / `self.jacobian(...)` on those raw vectors. The fit path goes through `NamedMlpMapper::forward_raw` → `build_input_vec_zero_fill` → `normalize` = `(x − center)/scale` (lines 868–869, 954–966) before the same `mlp.forward`.

So with `center`/`scale` declared — this PR's own fixture (`center = [70, 86]`, `scale = [12, 23]`), and what `covariate-nn.qmd` tells users to do — the fit evaluates layer 1 at z ∈ ≈[−2, 2] while the penalty evaluates the same weights at raw WT ∈ [45, 90] / CRCL ∈ [46, 82]. With tanh that region is saturated, so `C ≈ 0` and `∂C/∂w ≈ 0`: `nn_smooth` is a near no-op on exactly the models it is meant for, and for milder scales it regularizes a region of input space unrelated to the data. The PR body's "the grid is built in z-scored space … which is the space the network actually sees" is only true when `center`/`scale` are absent (and then the grid's own mean/sd z-score is a *second* normalization competing with the user's). `each_lambda_acts_alone` only asserts `p_smooth > 0.0`, which saturation residue satisfies.

Fix: build the grid in the mapper's normalized input space — apply `NamedMlpMapper::normalize` inside `raw_at` (and to the held medians / the missing-covariate origin), or store the grid as normalized vectors and keep the `MlpMapper` call. Add a test with non-identity `center`/`scale` asserting the `nn_smooth` gradient is materially non-zero and that the fitted modulator variance actually drops under `nn_smooth` alone.

src/estimation/outer_optimizer.rs
roninsightrx: **[bug] #751 plateau consistency check now compares a clean OFV against a penalized best-seen.**

After this rebind, `ofv` (line ~1723) is what lands in `state.best_ofv` (~1823) and the `best_seen` slot (~1858), and `best_seen_ofv` is copied out at ~2086. `final_ofv` is the clean `2·pop_nll_opts` recompute (~2153–2162, as the comment at 1431 intends). Both go into `failure_is_converged_plateau(…, best_seen_ofv, final_ofv, …)` (~2196), whose test is `final_ofv <= best + tol·(1+|best|)` (line ~1329). With `best = clean_best + P`, `P ≥ 0`, the check loosens by exactly the penalty magnitude, so a cold-restart point up to ~P worse than best-seen still reads "consistent" and an NLOPT_FAILURE exit can be upgraded to `converged = true`.

Smaller leaks from the same rebind: the checkpoint (`maybe_write(state.n_evals, ofv, …)` ~1921), the trace row (`write_foce` ~1902), the verbose `Eval N: OFV =` / `Restored best-seen point (OFV = …)` prints (~1826, ~2091) and the `Iter N` print in `optimize_bfgs` all record the penalized number, which then disagrees with the printed `Final OFV`, `FitResult.ofv`, and the checkpoint-resume banner in `api/fit.rs`. None of this is documented.

Fix: keep the clean `raw_ofv` alongside the penalized one (e.g. `best_seen: (xs, ofv_pen, ofv_clean)`) and hand the clean value to the plateau check; decide explicitly which stream the trace/checkpoint carry and say so in the trace column docs.

src/estimation/outer_optimizer.rs
```diff
@@ -1745,6 +1769,12 @@ fn optimize_nlopt_once(
                         &mut state.n_grad_evals,
                     )
                 };
+                // Add the NN penalty gradient in packed-x space (NN weights are
+                // identity-packed with unit scale, so no Jacobian correction);
+                // the `* scale[k]` below then carries it to optimizer space,
+                // exactly matching the value penalty added to `raw_ofv`.
+                #[cfg(feature = "nn")]
+                nn_reg.add_packed_gradient(&params.theta, &mut grad_raw);
```
roninsightrx: **[bug/docs] `final_gradient` is now ∇(OFV + penalty), not ∇OFV.**

This runs before `*last_gradient_cl = Some(grad_raw.clone())` (~1812), which becomes `OuterResult.final_gradient` → `FitResult.final_gradient`. `types.rs` documents that field as "Gradient of the objective function at the best-OFV parameter point" — with λ > 0 that is no longer true, and the trust-region `‖∂OFV/∂x‖` warning (trust_region.rs ~742) prints the same penalized vector under the OFV name. Either strip the penalty term before storing or update the doc contract (and the TR warning label).

Also the comment "identity-packed with unit scale, so no Jacobian correction" is half right: NN thetas *are* identity-packed (parser pushes `theta_lower = −∞`), but `compute_scale` gives any |w| > 0.1 a non-unit scale — the `* scale[k]` below is what makes this correct, not unit scale. Worth rewording so nobody later "simplifies" the scale multiply away.

src/estimation/outer_optimizer.rs
```diff
@@ -1415,6 +1427,13 @@ fn optimize_nlopt_once(
 
     let mut warnings = Vec::new();
 
+    // Covariate-NN (DCM) regularizer (L2 + smoothness). No-op when both λ are 0,
+    // so the penalized objective/gradient below stay byte-identical for
+    // unregularized fits. Reported OFV/AIC/BIC use the unpenalized −2LL
+    // (`final_ofv`), which is recomputed from a clean `pop_nll_opts` at the end.
+    #[cfg(feature = "nn")]
+    let nn_reg = crate::nn::NnRegularizer::build(model, population, options);
```
roninsightrx: **[plausible] Cross-attempt selectors compare the clean OFV while each attempt minimized the penalized one.**

- `resolve_stall_retry` (~1363–1368, `|result| result.ofv`) keeps the retry only if its *clean* OFV beats the stalled first run's. Fine when init NN weights are small (P_init ≈ 0 forces `clean_retry ≤ clean_init`), but warm-starting from an unregularized fit and re-fitting with a large `nn_l2` can have `P_init > P_retry`, and then the regularized retry is discarded and the stalled fit reported.
- Multi-start (`multistart_prefers(b.ofv, …)` in `api/fit.rs`) picks the lowest clean OFV among converged starts, i.e. the *least* regularized one — the opposite of what λ asks for. `perturb_init` jitters NN weights too, so starts genuinely differ in penalty.

Neither `FitResult` nor `OuterResult` carries the penalized objective, so there is nothing for these selectors to compare. Either surface it (e.g. `penalized_ofv: Option<f64>`) and select on it when λ > 0, or document that multi-start / stall-retry select on the unpenalized −2LL.

docs/model-file/fit-options.qmd
```diff
@@ -555,6 +555,26 @@ Setting `global_search = true` runs a gradient-free pre-search (NLopt CRS2-LM) b
 
 The number of global evaluations is auto-scaled based on the number of parameters and observations, or can be set explicitly with `global_maxeval`.
 
+## Covariate-NN regularization (`nn_l2`, `nn_smooth`)
+
+Two optional penalties regularize the small MLP behind a [`[covariate_nn]`](covariate-nn.qmd) (DCM) block. Both are **NN-only**, take a non-negative strength, and **default to `0.0` (off)** — a strict no-op that leaves non-NN and unregularized-NN fits byte-identical. They are added to the population objective the optimizer minimizes, each with a matching analytic gradient term, and apply across the FOCE/FOCEI outer optimizers (`bobyqa`, `slsqp`, `nlopt_lbfgs`, `mma`, `trust_region`, and the built-in BFGS).
```
roninsightrx: **[bug] `method = gn` silently ignores both keys, and `gn_hybrid` discards the regularized polish.**

`gauss_newton.rs` has no reference to `NnRegularizer` / `nn_l2` / `nn_smooth`; `run_foce_gn` evaluates `2·pop_nll` and builds BHHH from raw scores, so under `method = gn` (and SAEM / IMP / IMPMAP / Bayes / VI) the keys do nothing and no warning is raised (`NnRegularizer::build` has no method check). Worse, `gn_hybrid`: GN runs unregularized, the FOCEI polish minimizes the *penalized* objective via `optimize_nlopt`, and then `gauss_newton.rs:517` keeps the polish only if `polish_result.ofv < gn_ofv` — both *clean* −2LL. A polish that moved to the penalized optimum has clean OFV ≥ the unregularized GN optimum whenever the penalty bites, so the regularized result is thrown away with "FOCEI polish did not improve (GN result kept)". The user gets an unregularized DCM with no signal.

Fix (minimum): a pre-run warning / error when λ > 0 and the final-stage method is not one that applies the penalty, and say so on this line and in the CHANGELOG ("across the FOCE/FOCEI outer optimizers" reads as covering `gn`). Better: thread `nn_reg` into `run_foce_gn` (add `penalty_value` to the objective, `add_packed_gradient` to the score-sum, and compare penalized vs penalized in the hybrid selection).

src/parser/model_parser.rs
roninsightrx, line 7687: **[bug] These keys trigger a false "will be ignored" warning on every method that honours them.**

`apply_fit_option` pushes every matched key onto `user_set_keys` (line ~7958), but `nn_l2` / `nn_smooth` are in neither `FitOptions::framework_keys()` nor any `method_specific_keys()` arm (`types.rs` ~7592 / ~7651). `unsupported_keys_warnings` (~7508, called from `api/fit.rs:859`) therefore emits, for `method = focei`:

```
fit option `nn_l2` is not used by method `FOCEI` and will be ignored. Method-specific options for `FOCEI`: …
```

into stderr and `FitResult.warnings` — and the same via ferx-r's `settings = list(nn_l2 = …)`, which routes through this function. This is the regression class `test_ode_solver_keys_do_not_warn` (`model_parser_tests.rs:4413`) exists for; the PR's parser test only checks parsed values, and `tests/nn_regularization.rs` sets the lambdas programmatically so `user_set_keys` stays empty. Add both keys to the appropriate list (Foce/FoceI/Laplace + trust-region/bfgs arms, or `framework_keys` if you also wire `gn`) and a `test_nn_regularization_keys_do_not_warn` twin.

tests/nn_regularization.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/// λ = 0 must reproduce the unregularized fit byte-for-byte (the no-op
+/// guarantee). Runs two short fits and asserts identical OFV and theta.
+#[test]
+#[cfg_attr(
+    not(feature = "slow-tests"),
+    ignore = "slow: opt in with --features slow-tests"
+)]
+fn zero_lambda_fit_matches_baseline() {
```
roninsightrx: **[test] Tautological, and never runs in any CI job.**

`MODEL`'s `[fit_options]` sets neither key, so `baseline` already has `nn_l2_lambda = nn_smooth_lambda = 0.0` from `FitOptions::default()`; `zero_reg` sets them to `0.0` again. This is `fit(opts)` vs `fit(opts.clone())` and can only fail on nondeterminism — it cannot witness "λ = 0 is byte-identical to the unregularized engine". A meaningful form is a golden OFV/theta recorded on `main` before this PR, or a λ > 0 fit asserted to differ.

Separately, nothing in CI enables `nn` + `slow-tests` together: `slow-tests.yml` passes `ci,survival,slow-tests` (no `nn`, so this file is an empty binary there), and the `ci.yml` job that passes `nn` has no `slow-tests`. So this test, `l2_shrinks_weights_and_modulator_variation` in `src/nn/mod.rs` (which is `#[cfg(feature = "slow-tests")]`, not the CLAUDE.md `cfg_attr`-ignore form, and whose doc claims bit-rot is caught by a `cargo check`), and the pre-existing `nn_fit_*` slow tests all run nowhere. The PR's ‖W‖² 2159 → 0.004 evidence is never re-checked. Pre-existing hole, but this PR drops its Tier-3 evidence into it — worth adding `ferx-core/nn` to the nightly slow-tests features in this PR.

src/estimation/trust_region.rs
```diff
@@ -220,6 +228,15 @@ impl Gradient for FoceiProblem<'_> {
...
             }
         }
+        // NN penalty gradient (identity-packed weights, unit scale → maps 1:1
+        // into packed space). The BHHH `hessian()` intentionally omits the
+        // penalty's curvature — it only approximates the model Hessian, and the
+        // located minimum is set by where this (penalized) gradient vanishes.
+        #[cfg(feature = "nn")]
+        {
+            let params = unpack_params(p, self.init_params);
+            self.nn_reg.add_packed_gradient(&params.theta, &mut g);
```
roninsightrx: **[minor] BHHH Hessian omits the penalty curvature.**

`gradient` now carries ∇P but `hessian` (line ~248) is still `4·Σ gᵢgᵢᵀ` from raw scores, so the quadratic model has zero curvature along NN-weight directions from the penalty. For L2 the exact term is trivial (`+2λ` on the weight diagonal); for the curvature penalty a Gauss–Newton term `2λ·Σ (∂C/∂w)(∂C/∂w)ᵀ` is available from the same Jacobians. Without it, ρ is poor along weight directions under a large λ and the radius shrinks needlessly. Not a correctness bug (TR handles a bad model by shrinking), but cheap to fix and makes `trust_region` the one optimizer here with a proper Newton step on the regularized problem.

src/nn/mod.rs
```diff
@@ -909,6 +1036,255 @@ pub struct CovariateNn {
...
+            if sd <= 0.0 {
+                continue; // constant covariate — no curvature axis
+            }
+            let z_lo = (mins[i] - means[i]) / sd;
+            let z_hi = (maxs[i] - means[i]) / sd;
+            let span = z_hi - z_lo;
+            if span < 2.0 * step_z {
+                continue; // too narrow to form an interior triple
+            }
+            let n_nodes = ((span / step_z).floor() as usize + 1).min(NN_SMOOTH_GRID_MAX_NODES);
```
roninsightrx: **[minor] Cap truncates the sweep from the top; NaN covariates silently drop the axis.**

`n_nodes = min(⌊span/step⌋+1, 65)` then `for g in 1..n_nodes−1` from `z_lo`: when a z-span exceeds 16 the upper tail of the observed range simply gets no penalty. Coarsening the step (`step = span/64`) or centring the window would be the intent. Also `collect_input_values` passes through the `NaN` that the reader stores for a subject with no finite value (`datareader.rs:1525`); `NaN` poisons `mean`/`sd`, `sd <= 0.0` is false, `span` is NaN, `(NaN).floor() as usize = 0`, and the "unreachable" `n_nodes < 3` guard fires — the whole axis is dropped with no warning. A `.filter(|v| v.is_finite())` in `collect_input_values` closes that. (Time-varying NN inputs are also only seen at their baseline snapshot here — fine to leave, but worth a sentence in the docs.)

src/nn/mod.rs
```diff
@@ -487,6 +496,124 @@ impl MlpMapper {
...
+        let mut grad = vec![0.0f64; self.n_params];
+        if lambda == 0.0 || grid.is_empty() {
+            return (0.0, grad);
+        }
+        let mut s = 0.0;
+        for t in grid.triples() {
+            let ym = self.forward(&t[0], weights).expect("grid input shape ok");
+            let yc = self.forward(&t[1], weights).expect("grid input shape ok");
+            let yp = self.forward(&t[2], weights).expect("grid input shape ok");
+            let jm = self.jacobian(&t[0], weights).expect("grid input shape ok");
```
roninsightrx: **[efficiency] 6 forwards + 3 dense Jacobians per triple, most of it redundant.**

`jacobian` → `jacobian_impl` already runs `forward_cache`, so the three explicit `forward` calls above duplicate work; adjacent triples share nodes (x_g is `x+h` of g−1, `x` of g, `x−h` of g+1), so `3(n−2)` Jacobians where `n` suffice; and every gradient eval calls `penalty_value` (3 forwards/triple) and then `add_packed_gradient` → `curvature_penalty`, which recomputes the value and discards it (`let (_v, g) = …`). For a 2-input net at the 65-node cap that is ~126 triples × (6 forwards + 3 `n_out×n_params` allocations) per outer gradient. One pass computing forward+Jacobian per *node* and a single `fn(..., Option<&mut [f64]>) -> f64` kernel that serves both value and gradient would cut this by ~6× and drop the `_v` pattern.

src/estimation/outer_optimizer.rs
```diff
@@ -2345,6 +2386,15 @@ fn optimize_bfgs(
             options,
             grad_eval_idx,
         );
+        // Penalized value + matching gradient fed to the optimizer (unregularized
+        // fits unchanged). `ofv_at_fixed` above stays clean for final reporting.
+        #[allow(unused_mut)]
```
roninsightrx: **[nit] Prefer the existing `#[cfg(not(feature = "nn"))]` stub pattern over per-site gates.**

Fourteen `#[cfg(feature = "nn")]` sites across this file and `trust_region.rs`, plus three `#[allow(unused_mut)]` and the `let mut ofv = ofv; #[cfg(…)] { ofv += … }` shuffle. `NnGradPlan::build` (`nn_theta_gradient.rs:111`) already solves this with a no-`nn` stub so call sites carry no gate; a unit-struct `NnRegularizer` stub with `build → Self`, `penalty_value → 0.0`, `add_packed_gradient → {}` (in `estimation/`, since `src/nn` isn't compiled without the feature) makes every site a plain `raw + nn_reg.penalty_value(..)` and removes the `allow`s.

docs/model-file/covariate-nn.qmd
```diff
@@ -304,6 +304,42 @@ Two consequences worth knowing:
   do *not* feed the network are unaffected.
 
 `method = focei` uses a different gradient path and is unchanged.
+## Regularization (`nn_l2`, `nn_smooth`)
```
roninsightrx, line 308: **[docs] Missing blank line before the heading.** Pandoc's default `blank_before_header` means this `##` renders as literal text at the end of the previous paragraph, not as a section.

roninsightrx (commented):
Review at `high` effort. Verified by reading the PR branch (`4be6686e`); nothing here is speculative unless marked *plausible*.

**Must fix**
1. `nn_smooth` bypasses `center`/`scale` — the penalty evaluates the net at raw covariate values the fit never sees, so it is a near no-op on exactly the normalized models it targets (`src/nn/mod.rs:1212`).
2. `method = gn` ignores both keys silently; `gn_hybrid` discards the regularized polish by comparing clean OFVs (`fit-options.qmd:560`).
3. False "`nn_l2` is not used by method `FOCEI` and will be ignored" warning on every fit that uses the keys, including from ferx-r (`model_parser.rs:7687`).
4. #751 plateau consistency check compares a clean `final_ofv` against a penalized `best_seen`, loosening the converged-upgrade by the penalty magnitude; trace/checkpoint/verbose OFV streams also disagree with `FitResult.ofv` (`outer_optimizer.rs:1691`).
5. `FitResult.final_gradient` is now ∇(OFV+P) under a doc contract that says ∇OFV (`outer_optimizer.rs:1777`).

**Should fix**
6. `zero_lambda_fit_matches_baseline` is tautological, and no CI job enables `nn`+`slow-tests`, so none of the PR's Tier-3 evidence re-runs (`tests/nn_regularization.rs:108`).
7. *Plausible:* multi-start and the L-BFGS stall retry select on the clean OFV, i.e. the least-regularized start (`outer_optimizer.rs:1435`).

**Minor / nits:** TR Hessian omits penalty curvature; grid cap truncates the upper tail and NaN covariates drop an axis silently; ~6× redundant forward/Jacobian work per gradient; per-site `#[cfg]` gates instead of the `NnGradPlan` stub pattern; docs heading missing a blank line.

Findings 1–4 each have a concrete fix in the inline comment.

roninsightrx:
Pushed `95754bd8` addressing every finding in the review above.

**Must fix**
1. **`nn_smooth` bypassed `center`/`scale`** — the grid is now built in the mapper's normalised input space (`collect_input_values` applies `NamedMlpMapper::normalize`; the node vectors go straight into the same `MlpMapper::forward`/`jacobian` the fit calls). New test `curvature_grid_lives_in_the_mappers_normalised_input_space` shows a `center`/`scale` mapper on raw covariates and a plain mapper on pre-normalised ones build identical grids and penalties, and that the old raw-space grid on the same network reads < 1e-3 of that value.
2. **`gn` / `gn_hybrid`** — `run_foce_gn` now builds the regularizer; ∇P goes into the gradient and the penalty's curvature into the BHHH system; hybrid selection compares `polish.ofv + P(polish.theta)` against the penalized GN objective; reported OFV/trace/checkpoint stay clean. `fit_with_regularization_runs_under_gn` covers the path.
3. **False "will be ignored" warning** — `nn_l2`/`nn_smooth` added to the Foce/FoceI/Laplace/FoceGn/FoceGnHybrid `method_specific_keys` arms (SAEM etc. still warn). `test_nn_regularization_keys_do_not_warn_on_foce_family` is the twin of the ODE-key test. For programmatic callers (ferx-r bypasses `user_set_keys`) `fit()` now emits an explicit "does not apply covariate-NN regularization" warning when the final stage is non-FOCE (`nn_regularization_unapplied_warning`, unit + integration tested).
4. **Plateau check / user-facing streams** — `best_seen` carries `(xs, penalized, clean)`; the #751 consistency check gets the clean value. Trace `ofv`, checkpoint, `Eval`/`Iter` lines and the restored-best print all carry the clean −2LL (verbose `Eval` shows the penalized objective in parentheses when λ > 0). Same for BFGS (`clean_ofv_at`) and GN.
5. **`final_gradient`** — kept as ∇(OFV + P) (that is what a regularized fit is stationary on) and the doc contract on `FitResult::final_gradient` says so; the trust-region warning label reads `‖∂(OFV + NN penalty)/∂x‖` when active. The misleading "unit scale" comment is rewritten to say why `* scale[k]` must stay.

**Should fix**
6. **Tautological test / CI hole** — `zero_lambda_fit_matches_baseline` removed (λ = 0 is pinned by `regularizer_lambda_zero_is_noop` + the stub test). `slow-tests.yml` now passes `ferx-core/nn`, so `l2_shrinks_weights_and_modulator_variation` and the pre-existing `nn_fit_*` Tier-3 fits actually run nightly. Fixture deduplicated into `tests/fixtures/two_cpt_dcm_regularized.ferx` (`include_str!` from both `src/` and `tests/`).
7. **Multi-start / stall retry** — both rank on `ofv + P(theta)` (no `FitResult` field added; the regularizer is rebuilt at the selector).

**Minor**
- TR `hessian()` adds the penalty curvature: exact `2λ` L2 diagonal + Gauss–Newton `2λ Σ ∂C/∂w ∂C/∂wᵀ`. Pinned against FD of the gradient at kernel level (exact on the identity-head output layer) and as a placement seam in `trust_region` (`H(λ) − H(0)` ≡ the regularizer's packed Hessian, confined to the NN block; L2 block exact).
- Grid cap now coarsens the step over the full span instead of truncating; non-finite covariate values are filtered (`curvature_grid_coarsens_instead_of_truncating_past_the_node_cap`, `curvature_grid_ignores_non_finite_values`).
- One network evaluation per grid node (output read off the Jacobian's own forward sweep) and one value+gradient pass per objective eval, via a `PenaltySink` kernel that also serves the Hessian.
- `estimation::nn_reg` stub for `not(feature = "nn")` — all per-site `#[cfg]` gates and `#[allow(unused_mut)]` shims are gone.
- Docs: blank line before the heading; both pages now say which methods apply the penalty, that the grid is in normalised space, what is and is not penalized (trace, `final_gradient`, ranking, SEs). CHANGELOG entry updated.
- Assembled-gradient seam: `assembled_gradient_and_hessian_match_fd_of_penalized_cost_on_dcm` checks `∇cost(λ) − ∇cost(0)` against central FD of `cost(λ) − cost(0)` on every packed coordinate of the DCM fixture at λ > 0 (isolates the packed-index / scale seam from the EBE-fixpoint FD noise).

Verified locally: `tools/preflight.sh` (fmt, check matrix, clippy, rustdoc, docs-lint, public-API) green; 49 nn/TR/parser/api unit tests, 136 outer-optimizer/GN/multistart/checkpoint/trace unit tests, the 3 `nn_regularization` integration tests, the no-`nn` stub test, and the workflow/boundary guard tests all pass.

src/nn/mod.rs
roninsightrx: Fixed in `fb353444`. `collect_input_values` now reads every covariate snapshot the network is evaluated at — baseline plus `obs_covariates`, `dose_covariates`, `pk_only_covariates` and `reset_covariates` (the same four sources `Subject::time_varying_covariate_names` scans) — deduplicated per subject on bit-identical (LOCF) values, so a subject contributes one entry per distinct value it visits. Static-covariate populations are unchanged (every snapshot equals the baseline and collapses to one value per subject).

Test `curvature_grid_covers_time_varying_covariate_snapshots` is the constant-baseline / time-varying case: all subjects start at `WT = 70`, variation only in one source of each kind; it asserts the old baseline-only grid is empty, the collected multiset, and that the built sweep runs from the min past the top later-record value with a live penalty. Docs (`covariate-nn.qmd`) and the CHANGELOG entry now say the sweep spans per-record snapshots rather than the baseline.

TeunP:
@dominic-irx heads-up on a CI knock-on from this PR — no action needed from you on the immediate breakage, it's already got a fix up, but the sizing question is yours.

**`Slow tests` has been killed at its 120-minute timeout on every run since this merged.** It didn't look like a failure because a `timeout-minutes` kill is reported by GitHub as **`cancelled`, not `failure`** — five nightlies in a row read as infra blips while the suite was quietly covering ~20 fewer test binaries than it appeared to.

The mechanism is the `ferx-core/nn` addition to the nightly feature set, which was the right call in itself — but it also switched on every *other* `nn`-gated Tier-3 file. `tests/vi_dcm_omega_recovery.rs` (`#![cfg(feature = "nn")]`, 5 DCM VI/FOCEI fits) had never run in CI before, and both timed-out runs die in exactly the same place:

```
20:38:52  Running tests/vi_dcm_omega_recovery.rs
          ── 84 minutes, zero output ──
22:02:55  Terminate orphan process: (vi_dcm_omega_recovery-296157ee28e3378b)
```

**It's compute, not a deadlock.** Locally that binary runs at 590–790% CPU (~7 cores saturated) and was still unfinished after 13 minutes. The nightly pins `RAYON_NUM_THREADS: 1` on a 2-core runner, while libtest runs 4 of the 5 tests concurrently — so four heavy fits contend for two cores. 84 minutes without finishing follows straightforwardly.

#1240 unblocks the nightly: it moves `nn` into its own job so an over-budget fit can't take ~165 unrelated binaries with it, and wraps both jobs in `timeout` so an overrun fails red instead of going quiet. **It does not make the fits affordable** — the NN job will likely still hit its budget. That part is #1239 and it's a judgement call I didn't want to make for you: are those 5 fits meant to be this expensive, should they be trimmed or `#[ignore]`d for CI, or should the NN job get more rayon threads / a bigger runner?

One more thing from the same investigation, in #1239 rather than #1240 because I'm not sure it's the same root cause: **compiling `ferx-core --features nn --profile ci-test` took 2h20m** in a single `rustc` process on my machine. The same optimized build without `nn`, and with `nn` at `19fdbb04` (before this PR), both took ~15 min. Something in the +1501 lines to `src/nn/mod.rs` looks pathological for codegen — worth a look independent of the test runtime.

Finally, a correction in case you saw an earlier version of this analysis: I initially measured non-NN tests running 1.7× slower on the timed-out runs and suspected this PR had slowed the optimizer. **That was wrong** — probing the same commit with and without `nn` gives 0.99×/1.03×, and two commits differing only by a test-only change are 1.71× apart. It's GitHub runner variance, unrelated. This PR did not slow the fit path.

Issue #1139, opened 2026-08-30
bug(ode): a steady-state dose on a model-time-reading [odes] RHS is NaN (TAD/TAFD) or ~17% wrong (T/TIME)


TeunP:
## Summary

A steady-state dose (`SS=1`) on an `[odes]` model whose RHS reads model time does not produce
a steady state. Every spelling is affected, and the two families fail differently:

- `TAD` / `TAFD` → **`NaN`** for every observation — including when the term's coefficient is
  **zero**, so merely mentioning `TAD` breaks an otherwise ordinary SS model;
- `T` / `TIME` → a **plausible, silently wrong** number.

Found while reviewing #1130 (which fixes the *gradient*-route half of the same predicate:
the three SS gates asked `uses_time_vars` alone, so the bare-`TIME` spelling reached
`equilibrate_ss_state_g` while `T` declined). #1130 does not cause or cure the value
defect below; it is pre-existing, and re-measured unchanged after #1073 and #1121.

## Measured

Oracle: an explicit **40-cycle pulse train** of real doses ending at the same time as the
`SS=1` dose — no steady-state machinery in it at all. One-compartment IV bolus,
`d/dt(central) = -(CL/V)*central*(1 + <term>)`, `CL = 1`, `V = 20`, `II = 12`,
doses ending at `t = 120`, observations at 122 / 125 / 128 / 132, `eta = 0`.

| RHS term | `uses_time_vars` | `reads_time_builtin` | `SS=1` at t=122 | pulse train | worst rel |
|---|---|---|---|---|---|
| *(autonomous — control)* | false | false | 10.027269 | 10.027269 | **7.6e-9** |
| `0.0*TAD` | true | false | **`NaN`** | 10.027269 | — |
| `0.03*TAD` | true | false | **`NaN`** | 8.890201 | — |
| `0.003*TAFD` | true | false | **`NaN`** | 5.122100 | — |
| `0.003*T` | true | false | 9.545089 | 8.015275 | **1.7e-1** |
| `0.003*TIME` | false | true | 9.545089 | 8.015275 | **1.7e-1** |

The autonomous control agreeing to 7.6e-9 is what makes the oracle trustworthy: the pulse
train and the SS equilibration are the same answer for a time-invariant system.

With a larger coefficient the `TIME` error is much worse than 17% — `0.03*TIME` on the same
model gives `6.202` against a pulse-train value of `5.8e6`. The size is model-dependent; the
sign of the finding is not.

[...]

Issue #1218, opened 2026-09-03
bug(survival): predict_survival returns NaN for a time grid whose maximum is 0


TeunP:
## What happens

`predict_survival` returns `cum_hazard` and `hazard` as `NaN` when the requested time grid's
**maximum is 0** — not, as it first looks, for any single-point grid.

Measured on `nonmem_anchor/ss_chz_const_fit.ferx` + `nonmem_anchor/ss_chz_ss.csv` (joint PK-TTE,
constant hazard `H0 = 0.02`, so the closed form is `H(t) = 0.02 · t`):

| grid | `H` returned |
|---|---|
| `[0.0]` | `NaN` |
| `[0.0, 0.0]` | `NaN, NaN` |
| `[1.0]` | `0.02` |
| `[12.0]` | `0.24` |
| `[0.0, 1.0]` | `0.0, 0.02` |
| `[0.0, 1.0, 4.0, 8.0, 12.0]` | `0.0, 0.02, 0.08, 0.16, 0.24` |

So a single-point grid is fine, and `H(0)` is perfectly computable — it is `0.0`, as the fifth
and sixth rows show. What fails is specifically a grid that contains no time greater than 0.

## Why it matters

It is a silent wrong answer rather than an error: the call returns rows carrying `NaN`, with no
warning and no `Err`. A caller folding those through `f64::max` / `f64::min` will not notice —
`0.0f64.max(f64::NAN)` is `0.0` — so a diagnostic or an R-side wrapper that asks for the
survival curve at a single early time can quietly report nothing wrong.

Asking for `H` at `t = 0` on its own is a natural thing to do; it is how you would check that a
subject's cumulative hazard starts at zero.

## Where

`ode_dense_solve_states` (`src/ode/predictions.rs:6309`) prefills its result with `f64::NAN`,
then takes the integration horizon as

```rust
let t_last = saveat.iter().cloned().fold(0.0f64, f64::max);
```

With `saveat = [0.0]` that is `0.0`, so no segment is built, no `saveat` row is ever written, and
the `NaN` prefill is what the caller receives. There is already a branch that fills `saveat`
nodes lying *earlier* than the first integrated segment, but it uses a strict `<`, so a node
sitting exactly at the first segment start is not covered.

`ode_cumhaz_hazard` (`src/survival/mod.rs:710`) then copies the values straight out. It has its
own `f64::NAN` prefill and a `states.len() != n` guard, but the length matches here (the prefill
[...]

PR #1219, opened 2026-09-03
fix(ode): an SS dose no longer carries the equilibration run-in into H(0) [closes #1210]

### LINKED ISSUE #1210: bug(ode): an SS dose carries the equilibration run-in hazard into H(0), so every joint PK-TTE survival quantity is wrong
## Problem

For a joint PK-TTE model (`[event_model] hazard = …`, `HazardSpec::OdeAccumulated`), an `SS=1`
dose leaves the cumulative-hazard accumulator holding **the hazard accrued during the
steady-state equilibration run-in**. `H(0)` is therefore not 0, and every survival quantity
downstream of it is wrong.

`equilibrate_ss_state` (`src/ode/predictions.rs:513`) cycles the **whole** state vector
`vec![0.0; ode.n_states]` through `SS_EQUILIBRATION_CYCLES` dosing intervals and returns it as
the record's initial state. That is right for the PK compartments — it is the entire point —
and wrong for the injected `d/dt(__chz_<cmt>)` row, which is a pure integrator with no
elimination term. It never reaches a steady state; it just counts up.

## Measured, against a closed form — no engine needed

Take `hazard = H0` with `H0 = 0.02` **constant**. Then `H(t) = 0.02·t` exactly, whatever the PK
does. One subject, 100 mg into the depot, `II = 12`:

```
reference                       H(0) = 0        H(12) = 0.24

single dose      H(0)=  0.000000   H(1)=  0.020000   H(4)=  0.080000   H(12)=  0.240000   ✓
SS=1, II=12      H(0)= 12.000000   H(1)= 12.020000   H(4)= 12.080000   H(12)= 12.240000   ✗
```

The offset is **exactly** `H0 × SS_EQUILIBRATION_CYCLES × II` = `0.02 × 50 × 12` = `12.0`. The
shape is right and the origin is displaced: the run-in's 600 hours of hazard are carried into
the record.

Drug-driven hazard, same subject, `hazard = H0·exp(BETA·C)`:

```
single dose      H(0)=    0.000000                    H(12)=    5.040273
SS=1, II=12      H(0)= 1268.971796                    H(12)= 1295.090308
```

## Why it matters

`S(t) = exp(−H(t))`, so `S(0) = exp(−12) ≈ 6e-6` for the constant-hazard case and `exp(−1269)`
— i.e. exactly `0` in double precision — for the drug-driven one. Consequences:

- **The objective is wrong, not just a diagnostic.** The TTE contribution is `−log S − log h`,
  so each subject's objective is inflated by ≈ `2 × H_runin`: about **24** for the constant
  hazard above and **2538** for the drug-driven one. An affected fit is not "slightly off"; the
  likelihood is dominated by a term that should not exist.
- **`predict_survival` shows it too** (the numbers above are its output), so `S(t)` / `H(t)` /
  `h(t)` and any survival curve plotted from them are affected, not only `fit()`.
- **Silent.** `api/validation.rs` gates an ODE-accumulated hazard against *time-varying
  covariates* (`:2116`) but says nothing about `SS`. The fit converges and reports.

## Scope — this is not #1166

It has nothing to do with time-dependent hazards: the demonstration above uses a **constant**
hazard. It reproduces identically before and after
[#1166](https://github.com/FeRx-NLME/ferx-core/issues/1166) — the objective for an SS subject
was `123350887819.021` before that fix and `123350887819.024` after, i.e. the same number to
13 significant figures — so #1166 neither caused it nor worsened it. It was found while
reviewing #1202.

It is also, in a sense, the same category error #1166 fixed: the injected accumulator is not a
state of the PK system and should not be treated as one. #1166 stopped it answering "is the PK
block non-autonomous?"; this issue is it being fed through steady-state equilibration.

## Suggested solution

**Zero the accumulator rows in the state returned by SS equilibration.** `H(0) = 0` is
definitional — the cumulative hazard is measured from the start of the subject's record, and no
amount of run-in changes that. The PK compartments keep their equilibrated values, which is
what `SS=1` means.

Three sites cycle the full state vector and would each need it:

- `ode::predictions::equilibrate_ss_state` (`src/ode/predictions.rs:513`)
- `ode::predictions::equilibrate_ss_input_rate_g` (steady state into a built-in absorption
  route, #835)
- `sens::ode_provider::equilibrate_ss_state_g`

The slots are already known at parse time — #1166 threads them into `build_ode_spec` as
`chz_state_slots`, and each endpoint carries its own in
`HazardSpec::OdeAccumulated { chz_state }`. They are not currently stored on `OdeSpec`, so the
tidiest change is probably to keep them there (they are a property of the compiled ODE system,
which is exactly what `OdeSpec` describes) rather than threading a slice through three
signatures.

**A guard is not sufficient on its own but may be worth pairing**, because there is a second
question hiding here: `(I − M)⁻¹b`, the exact analytic SS solve (#908/#914), is **singular**
along a pure integrator — the accumulator has a zero eigenvalue, so there is no steady state to
solve for. Zeroing the row sidesteps that for the accumulator itself; it is worth checking that
the exact solve is not being asked to invert a matrix that includes it.

## Validation

The constant-hazard case is a closed-form oracle (`H(t) = H0·t`), so this needs no reference
run: a Tier-1 test asserting `H(0) == 0` and `H(12) == 0.24` for an `SS=1` subject fails today
and passes after. Pair it with the non-degeneracy the repo's rules ask for — a drug-driven
hazard whose SS and single-dose arms must then differ only by the PK accumulation, not by three
orders of magnitude — and, if an external anchor is wanted, `ADVAN13` with `SS=1` and an
explicit `A(3)` compartment reproduces the same object (the existing
`nonmem_anchor/pktte_tdep.ctl` is one `SS`/`II` column away from it).

## Provenance

Measured while reviewing #1202; harness in that session's scratchpad
(`zz_ss_measure.rs`). Reported rather than fixed there because it is out of that PR's scope and
predates it.

TeunP:
## Why

For a joint PK-TTE model (`[event_model] hazard = …`), an `SS=1` dose left the cumulative-hazard
accumulator holding **the hazard accrued during the steady-state equilibration run-in**.

`equilibrate_ss_state` cycles the whole state vector through `SS_EQUILIBRATION_CYCLES` dosing
intervals. That is right for the PK compartments — it is the entire point — and wrong for the
injected `d/dt(__chz_<cmt>)` row, which is a pure integrator with no elimination term. It never
reaches a steady state; it just counts up.

Measured against a closed form (`hazard = H0` constant ⇒ `H(t) = H0·t`, no engine needed), at
`1187062a`:

| arm (`H0 = 0.02`, `II = 12`, reference `H(t) = 0.02·t`) | `H(0)` | `H(12)` | note |
|---|---|---|---|
| single dose | 0 | 0.240000 | ✓ |
| `SS=1` bolus | **12.000000** | 12.240000 | `0.02 × 50 cycles × 12` |
| `SS=1` infusion, `RATE=25` | **12.000000** | 12.240000 | second code path, same offset |
| `SS=1` + `ALAG1 = 2` (< II) | **12.200000** | 12.200000 | `H(1) = 12.22 > H(4) = 12.04` — **non-monotone** |
| `SS=1` + `ALAG1 = 14` (> II) | **12.000000** | 12.240000 | |
| `SS@0` + `SS@48` | `H(47.9) = 12.958 → H(48) = 12.000` | | the accrued `0.958` is **discarded** |
| drug-driven `H0·exp(β·C)` | **1268.971796** | 1295.090308 | |
| `H0·exp(0.15·TIME)` | **2627477986.20** | | |

`S(t) = exp(−H(t))`, so this is the objective, not a diagnostic: each subject's TTE contribution
`−log S − log h` was inflated by ≈ `2 × H_runin`. Three further symptoms the issue did not have:

- **`fit()` emitted two spurious warnings** — `Steady-state (SS=1) equilibration reached the
  50-cycle cap without converging (ratio ≈ 1.000)`. They were honest: the accumulator's one-cycle
  map is the identity, so `I − M` was singular, the exact `(I − M)⁻¹b` solve (#914) declined for
  **every** joint model, the capped train ran, and `SsStopTracker` watched a row that grows by
  `H0·II` every cycle.
- **Simulation was the loudest symptom.** Drug-driven hazard, `horizon = 24`, 20 draws: a
  single-dose subject events at `0.5 … 7.6 h`; the same subject with `SS=1` evented at
  `t = 0.000` in **20 of 20** draws — `H = 1269` exceeds every `−log U`.
- **A hazard reading `TAD`/`TAFD` under `SS=1` returned `NaN` for the whole subject**, even at a
  zero coefficient.

## What changed

`OdeSpec` learns its accumulator slots (`chz_state_slots`, threaded into `build_ode_spec` by
#1166 and now stored — they are a property of the compiled ODE system). The fix is then three
pieces at one chokepoint, and **none of the three is sufficient alone**:

1. **`mask_chz`** — every integration inside the equilibration runs with `dy[chz] = 0`, so the
   accumulator cannot move during the run-in. This covers the capped pulse train, the infusion
   windows, and `ss_state_at_phase`'s phase advance (which reconstructs the *previous* interval's
   tail — still run-in, and where the `ALAG1 = 2` arm got its extra `0.2`).
2. **`restore_chz`** — on the way out, `u[chz] = u_before[chz]`. **Preserve, not zero**: the
   issue's suggested "zero the accumulator rows" is right only for a first SS dose. A second
   `SS=1` dose at `t = 48` re-loads the compartments but must keep `H(48⁻)`; an SS dose is not a
   reset. One chokepoint, so the three early bail-outs (`ii ≤ 0`, out-of-range dose compartment,
   overlapping infusions) cannot silently zero it either.
3. **`periodic_ss_fixed_point_pk`** — the accumulator rows are *projected out* of the fixed-point
   solve, not merely masked. Under masking their one-cycle map is the identity, so `I − M` still
   has a zero row and the exact solve would still decline for every joint model.

All nine call sites (5 × `equilibrate_ss_state`, 4 × `ss_state_at_phase`) pass a `chz_snapshot`
of their current state. The dual/`Dual2` twins are unreachable for a joint model (`has_tte` in
`inner_optimizer`, `has_non_gaussian` in the provider route it to FD end-to-end) and are
unchanged, with a `debug_assert!(!program.has_chz())` at each entry so a later widening of that
routing fails loudly instead of silently banking the run-in into the gradient.

### After

Every arm of the before-table now lands on the closed form, and the three unreported symptoms
are gone:

| arm | `H(0)` | `H(12)` |
|---|---|---|
| `SS=1` bolus / infusion / `ALAG1 = 2` / `ALAG1 = 14` | 0.000000 | 0.240000 |
| `SS@0` + `SS@48` | `H(47.9) = 0.958 → H(48) = 0.960` | `H(60) = 1.200000` |
| drug-driven | 0.000000 | 26.118512 |
| `H0·exp(0.15·TIME)` | 0.000000 | 0.673286 (= the single-dose arm, as it must be) |

`fit()` OFV on the constant-hazard SS fixture: **83.592027 → 59.592027**, i.e. exactly
`2 × H_runin = 24.0`. Drug-driven: `2644.503682 → 106.560089` (`Δ = 2537.94 = 2 × 1268.97`).
Both `Steady-state (SS=1) equilibration` warnings are gone. Simulated SS event times are
distributed (`0.114 … 1.596`) instead of 20/20 zeros. `TAD`/`TAFD` + `SS=1` no longer returns
`NaN` — the mask cures it as a side effect (#1139's family, on the hazard row).

## Alternatives considered

**Zero the accumulator rows** (the issue's suggestion). Passes every single-dose arm, because a
first SS dose's pre-record value *is* zero — and discards `H(48⁻)` at a mid-record SS dose. The
mid-record arm is in the test set precisely to separate the two.

**Mask without projecting.** Leaves `I − M` singular, so every joint model keeps running the
50-cycle train and keeps emitting the two warnings. Correct, slow, and noisy.

**Thread a `&[usize]` through the three equilibrate signatures** instead of storing on `OdeSpec`.
It is nine call sites, not three, and the slots are a property of the compiled system.

## Cross-repo dependency

`ferx-r` `origin/main` constructs no `OdeSpec` (checked: only a comment mentions the type), so the
new public field breaks nothing there.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — `OdeSpec` gains `pub chz_state_slots: Vec<usize>`.
  Baseline regenerated with `tools/update-public-api.sh`. It is public because `OdeSpec` is a
  public struct with public fields and is constructed in-crate at 29 sites; no external caller
  constructs one.

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0**, committed under `nonmem_anchor/`.

**There is no native NONMEM comparator for this object, and that is itself one of the committed
runs.** Handing NONMEM's own SS routine an `ADVAN13` system with a `CHZ` compartment
(`ss_chz_r1_{const,drug}.ctl`) returns:

| run | `A(3)` | outcome |
|-----|-------:|---------|
| `r1_const` (β = 0) | **−2.38842166E+07** at every record | `#OBJV: +INF` — and the table is still written |
| `r1_drug` (β = 0.5) | — | `NUMERICAL DIFFICULTIES WITH STEADY STATE SOLUTION`, `PROGRAM TERMINATED BY OBJ` |

`H(0) = 0` is therefore a **ferx definition** (the cumulative hazard is measured from the start of
the subject's record) and the anchor is the construction a NONMEM user has to write by hand:
`ss_chz_r2_*.ctl`, an explicit 51-dose pulse train at 0, 12, …, 600 with `$DES` gating the hazard
on `T ≥ 600`. At `T = 600` the PK is at its periodic steady state (trough `4.78896241`) and `A(3)`
restarts from `8.1e-13` — the run's own check that the gate holds. NONMEM `T = 600 + t` is ferx `t`.

The **const** arm reproducing the closed form to the printed digits is what makes the other two
trustworthy; the **drug** arm is the oracle for what no closed form sees; the **tdep** arm pins
the model clock, which is the quantity an SS run-in displaces.

```
                     A(3) at NONMEM T = 601 / 604 / 608 / 612   (= ferx t = 1 / 4 / 8 / 12)

r2_const   NONMEM   0.02          0.08          0.16          0.24
           ferx     0.020000      0.080000      0.160000      0.240000     (closed form: 0.02·t)
r2_drug    NONMEM   1.50356118   17.4316438    24.5093214    26.1185123
           ferx     1.503561     17.431644     24.509321     26.118512
r2_tdep    NONMEM   1.66704849   24.8746770    41.1505197    48.0540176
           ferx     1.667048     24.874677     41.150520     48.054018

IPRED      NONMEM  10.2994688    10.4546754     7.14066685    4.78896241
           ferx    10.29946878   10.45467542    7.14066685
```

`IPRED` is the control, not the finding: it **already matched to 8 digits before the fix** on all
three arms, which is what localises the defect to the accumulator row alone. Before the fix ferx's
`H` was displaced by exactly its own `H(0)` at every record — `+12.000000`, `+1268.971796`,
`+4.5360222e10` — so the shape was right and only the origin was wrong.

## Performance

- [x] Faster for joint SS models — the exact fixed point is admitted again: **50 cycles → 1**
  (`last_ss_equilibration_cycles`), on both the bolus and the built-in-absorption branch. No
  change for any model without an `[event_model]`: `chz_state_slots` is empty there, every helper
  short-circuits, and the full 4046-test `--lib` suite is unchanged.

## Tests

- [x] Unit tests added in the same module (`src/ode/predictions_tests.rs`, 12 tests)
- [x] Regression tests that fail without this fix — **each mutation-verified**, with the killing
  edit named in the test's own doc comment:

| mutation | tests killed |
|---|---|
| `M1` drop `restore_chz` in `equilibrate_ss_state` | `a_second_ss_dose…`, `a_plain_dose_followed_by_an_ss_dose…`, `ss_equilibration_returns_the_accumulator_untouched`, `ss_lagged_dose_does_not_bank_its_phase_advance` |
| `M2` zero instead of preserve (the issue's suggestion) | the same four |
| `M3` phase advance unmasked | **none** |
| `M4` equilibration unmasked | **none** — until the nonlinear arm existed |
| `M4'` same, re-measured with `a_nonlinear_joint_model_judges_convergence_on_the_pk_rows` present | that test |
| `M5` no projection | `a_joint_linear_model_takes_the_exact_ss_solve` |
| `M6` input-rate branch unmasked/unreduced | none on `H` alone — the masked fallback train produces the right hazard, so the arm asserts the **cycle count** instead |
| `M7` composite (mask + projection + restore removed = pre-fix) | 9 of 10; survivor `ss_lag_beyond_the_interval…` |
| `M8` drop `restore_chz` in `ss_state_at_phase` | **none** — until the mid-record lagged arm existed |
| `M8'` same, re-measured with `a_mid_record_ss_dose_lagged_past_the_interval…` present | that test |

M1 and M2 killing the *same* four is the point of the mid-record arm: for a subject's first SS
dose "preserve" and "zero" are the same edit, and only the `SS@48` / lagged-arrival / unit arms
can separate them.

**The table found a real coverage hole, which this PR then closes, and one component that
genuinely cannot be held by a test.**

- `restore_chz` inside `ss_state_at_phase` was held by nothing (`M8`). `ss_state_at_phase_pk`
  returns the accumulator at zero unconditionally — it starts from `equilibrate_ss_pk_state`'s
  zero vector and every integration inside it is masked — so the restore only has an effect
  when `chz_before != 0`, i.e. an SS dose that is **both mid-record and lagged**. Every other
  lag arm doses once at `t = 0`, where restore and no-restore are identical. This is also why
  `ss_lag_beyond_the_interval…` survived `M7`: the hypothesis was that the arm was served
  entirely by `ss_state_at_phase`, and `M8` disproved it.
  `a_mid_record_ss_dose_lagged_past_the_interval_keeps_the_hazard_accrued_so_far` closes it —
  a plain dose at `t = 0`, then `SS=1` at `t = 48` with `ALAG1 = 14 > II`, so the phase clamps
  to zero and there is no arrival re-equilibration to put the row back from the other side.
- `mask_chz` inside `ss_state_at_phase_pk` (`M3`/`M3'`) is kept although **no test can hold
  it**, and that is a decision rather than an oversight. It is redundant with `restore_chz`,
  which overwrites the row on exit regardless; the phase advance spans at most one interval, so
  the row it would grow is bounded by `hazard × II` rather than 50 times that. The only channel
  left is the integrator's error norm, a tolerance-level effect and the wrong thing to pin in a
  test. It stays for uniformity across the three SS paths and because deleting it would leave
  `restore_chz` as the sole thing carrying correctness there. The `mask_chz` doc says so.

Every piece of the fix now has a named test, except the one recorded above as untestable.

- [x] Tier-2 NONMEM anchor: `tests/ss_chz_nonmem_anchor.rs` (6 tests) — `H`, `h` and `PRED`
  against `results/ss_chz_r2_*.tab` on all three arms, the reference's own closed-form
  self-check, `H(0) = 0` paired with the objective on the fit path, and the simulated
  event-time distribution.

Bounds are measured, then set ~10× above the observed residual; both numbers are recorded at each
assertion. The model files pin `ode_reltol = 1e-9 / ode_abstol = 1e-11` — at ferx's defaults this
comparison is good only to ~1e-3.

Four "this test cannot fail" traps were checked for: every `f64::max` fold is preceded
by an `is_finite()` assertion (a `NaN` would otherwise leave `worst` at whatever the finite
records produced); the mid-record and reset arms **assert their own straddle** (`H` before the
event must be materially non-zero, or "preserve" and "zero" agree and the test is vacuous); the
input-rate arm asserts the **cycle count**, because its hazard assertion alone is satisfied by
the masked fallback train and so could not see an unmasked joint branch; and the fit-path arm's
SS-warning check was **measured against the pre-fix engine and found vacuous** — that fixture
emits no `Steady-state (SS=1) equilibration` warning either way, so the arm was rebuilt around
`H(0) = 0` and the objective, and the warning check is documented as a guard rather than
presented as the finding.

The objective in that arm is pinned to a measured pair rather than a remembered one:
`94.462287664653` pre-fix against `70.462287654391` post-fix, a gap of `24.0000000103` against
the closed form `2 * H_runin = 2 * H0 * 50 * II = 24.0`. An earlier draft pinned `59.592027`,
which was correct for the scratch probe's synthetic single-subject model and wrong for this
fixture.

- [x] Not applicable — a correctness fix with no new API surface. No DSL syntax, no
  `[fit_options]` key and no new output field: the same model file that produced a corrupted
  `H(0)` now produces the right one. The behaviour is described on
  `docs/model-file/steady-state.qmd` instead, and the three `nonmem_anchor/ss_chz_*_fit.ferx`
  models committed here are runnable minimal reproductions.

## Docs

- [x] `docs/model-file/steady-state.qmd` — new `### SS with an ODE-accumulated hazard` under
  Limitations: the convention, the mid-record rule, the reset boundary, and the NONMEM negative
  result.
- [x] `docs/model-file/event-model.qmd` and `docs/estimation/tte.qmd` — one bullet / one
  paragraph each, both linking to it.
- [x] `nonmem_anchor/README.md` — index row + a full section (r1 as the negative result, r2 as
  the anchor).
- [x] `CHANGELOG.md` `[Unreleased]` — `Fixed` and `Performance` entries.

## Checklist
      (`--all-targets`), rustdoc, docs-lint, public-API baseline.
- [x] Functions on the `Dual2` sensitivity path stay differentiable. The dual SS helpers are
      untouched; instead `equilibrate_ss_state_g` and `equilibrate_ss_input_rate_state_g` gained
      a `debug_assert!(!program.has_chz())`, because a joint model must not reach them — it
      routes to FD. That routing is asserted rather than assumed.
- [ ] `Cargo.toml` version bumped if breaking — not breaking (additive field only).

### ferx-site
- [x] No new `.ferx` DSL syntax or `[fit_options]` key.
- [x] The new `.ferx` files live under `nonmem_anchor/`, not `examples/`, so no site/book entry
      is implied. The new data file `nonmem_anchor/ss_chz_ss.csv` is anchor input, not a
      `data/` example.

### ferx-book
- [x] No new estimator, DSL feature or fit option.

### Example execution
- [x] No example execution step needed — no `examples/` model or `data/` file changed, and no
      public API that ferx-r calls changed shape (it constructs no `OdeSpec`).

## Reviewer hints

The mutation table above says which test holds each piece of the fix, and names the one piece
that no test can hold. The subtle parts:

- **`ss_state_at_phase` also masks, and also restores.** The phase advance is run-in, not record
  time. Easy to read as record time, and it is where the non-monotone `H` came from. Its restore
  is reachable only by a dose that is *both* mid-record and lagged, which is why it went
  uncovered until `M8` said so — worth a look at
  `a_mid_record_ss_dose_lagged_past_the_interval_keeps_the_hazard_accrued_so_far` and at the
  `ALAG1 > II` reasoning in its doc, since a lag *inside* the interval would make the arm die
  from the wrong edit.
- **The projection's validity rests on one-directional coupling.** The hazard may read PK states;
  `[odes]` may **not** read `__chz_*` (rejected at parse time — "reserved and write-only"). That
  is what makes the PK sub-dynamics independent of the accumulator and the restriction exact.
- **`chz_snapshot` must be taken after any `init(...)`/reset re-seed**, not from a zero vector.
  All nine sites read the live `u` at the point of the call; the `apply_segment_boundary` site is
  the one where a reset runs first.
- The 29 `chz_state_slots: Vec::new()` lines in the diff are the mechanical consequence of adding
  a field to a struct with public fields; only `model_parser.rs:11750` carries a real value.

## Open questions

None. Out of scope and stated: `SS=2` (unimplemented, unchanged) and any change to `EVID=3`/`4`
reset semantics for the accumulator (pinned as-is by
`a_reset_zeroes_the_accumulator_with_or_without_an_ss_dose_on_the_row`, so a later decision is a
deliberate diff).


TeunP:
## Review — xhigh

Ten findings. Two are the kind this PR's own rules exist to catch, so they go first.

### 1. The shared hazard assertion passes vacuously on an empty `got` — `src/ode/predictions_tests.rs:8768`

`assert_hazard_is_the_closed_form` zips `got` against `times` with no length assertion:

```rust
for (&h, &t) in got.iter().zip(times) { ... }
for w in got.windows(2) { ... }
```

Both loops are no-ops when `got` is empty, and the helper returns having asserted nothing.
`chz_at` returns whatever `ode_dense_solve_states` produced, so if that walk ever returns fewer
rows than `times` — the zero-max-grid path filed as #1218, or a solver bail-out — **every Tier-1
arm in this PR goes green while computing no hazard at all**. That is all of
`ss_bolus_starts_the_hazard_clock_at_the_record`, both lag arms, both mid-record arms, the
input-rate arm and the nonlinear arm: they route their entire correctness claim through this
helper. Fix: `assert_eq!(got.len(), times.len())` at the top.

### 2. The new doc block orphaned `equilibrate_ss_input_rate`'s documentation — `src/ode/predictions.rs:141`

`mask_chz`'s doc comment was inserted with no blank line after the existing block, so the 20 lines
that documented `equilibrate_ss_input_rate` ("Periodic steady-state trough for an `SS=1` dose into
a built-in absorption input-rate compartment (#719; nonlinear solve #867) … Returns `None` … `ρ ≥ 1`")
are now the opening of `mask_chz`'s doc — a three-line loop that zeroes slots. `equilibrate_ss_input_rate`
at line 268 is left a bare `fn` with no doc at all. The `rustdoc` preflight gate cannot see this:
nothing is an unresolved link.

### 3. The docs say `simulate()` is unsupported; this PR adds a passing test that does it — `docs/model-file/event-model.qmd:86`

> `fit()` and `predict_survival()` work; **`simulate()` is not yet supported** for an
> ODE-accumulated hazard (drug-driven event-time sampling is a later slice — it returns a clear
> error rather than silently emitting no event rows).

The PR inserts a new bullet at line 82 of that same list, and adds
`a_simulated_steady_state_subject_draws_a_distribution_of_event_times`, which calls
`simulate_with_options` on the drug-driven arm and asserts 20 event rows with a spread — green in
the 6/6 anchor run. `draw_ode_tte_latent` / `ode_solve_until_chz_threshold` implement it in
`src/survival/mod.rs`. The page tells users a working feature errors out.

### 4. "two spurious warnings" is fixture-specific, stated generally — `CHANGELOG.md:47`

Running the pre-#1210 engine against this PR's own anchor (`ss_chz_const_fit.ferx` +
`ss_chz_ss.csv`, a linear PK block — exactly the case the entry describes) emits **five** warnings,
none containing `Steady-state (SS=1) equilibration`: FOCEI default, FD inner gradients, thread
count, EPS shrinkage, IWRES autocorrelation. The "two" was measured on a different synthetic
fixture. Whether #867 fires depends on the geometric-tail heuristic in
`note_ss_nonconvergence_if_capped`, not on cycling alone.

The same claim sits at `src/ode/predictions.rs:218` and `src/ode/predictions_tests.rs:9087`, while
`a_joint_steady_state_fit_does_not_bank_the_run_in_into_the_objective` documents the opposite — so
the PR ships three statements contradicting a fourth.

### 5. The monotonicity assertion is documented as load-bearing but cannot fire — `src/ode/predictions_tests.rs:8762`

The doc says "The monotonicity check is not redundant with the closed form". It is. The
closed-form assertion runs first and pins every value to within `1e-9` of `H0*t`; consecutive
fixture times differ by ≥ 0.1, so consecutive expected values differ by ≥ 0.002 — six orders of
magnitude above the slack. The cited motivating case (`H(1) = 12.22 > H(4) = 12.04`) panics on the
closed-form bound at `H(1)` before the monotonicity loop is reached.

### 6. `debug_assert!` leaves release builds unguarded — `src/sens/ode_provider.rs:3876`

The tripwire against a joint system reaching the dual SS equilibration compiles out in release —
the build where the consequence matters. Its own message says it exists for a *future* widening of
the routing; if that lands, release builds bank the run-in into `∂/∂η` and `∂/∂θ` with no panic, so
FOCE/FOCEI converges to the wrong optimum instead of failing. A plain `assert!` is one bool read on
a path already doing ODE solves.

### 7. The projection block is written twice — `src/ode/predictions.rs:338`

`pk_rows` / `embed` / `project` appear verbatim at 245-263 and 338-352. They cannot share
`periodic_ss_fixed_point_pk` (different delegate), but the projection *is* the subtle part of the
fix — accumulator rows embed as `0.0` and project out — and the input-rate copy is the one with a
single test holding it. A `ChzProjection { n, pk_rows }` with `embed`/`project` methods would keep
them from drifting.

### 8. `restore_chz` degrades silently into the pre-#1210 semantics — `src/ode/predictions.rs:204`

`chz_snapshot` uses `unwrap_or(0.0)` and `restore_chz` guards `if slot < u.len()`. A
`chz_state_slots` inconsistent with `n_states` therefore reads 0.0 and writes 0.0 — precisely the
zeroing the issue rejects, and the one outcome no test can distinguish from correct on a first SS
dose. `mask_chz` has the same silent skip. `debug_assert!` on the bound in all three.

### 9. The joint input-rate branch lacks the `n == 0` guard — `src/ode/predictions.rs:338`

Line 278 returns `None` for `n == 0`, protecting the non-joint delegation; the joint branch passes
an unguarded `pk_rows.len()`. Only reachable from a hand-built spec today, but the failure mode is
an all-zero vector treated as an equilibrated trough rather than a fallback.

### 10. `chz_at` hardcodes slot 1 — `src/ode/predictions_tests.rs:8756`

`.map(|s| s[1])` while every fixture separately declares `chz_state_slots: vec![1]`. A future
fixture with the accumulator elsewhere would silently compare drug amounts against `H0*t`. Read
`ode.chz_state_slots[0]`.

---

**Not findings, checked:** the projection's exactness rests on `[odes]` being unable to read
`__chz_*` (rejected at parse time) — verified. `advance` taking `&dyn Fn` handles both RHS types
correctly. `equilibrate_ss_input_rate`'s joint result is restored via `equilibrate_ss_pk_state`'s
wrapper. `chz_snapshot` reads live `u` at all nine call sites. The simulation arm guards
`is_finite()` before its `min`/`max` folds. The reset arm indexes directly so it cannot truncate
like finding 1.

TeunP:
## All ten addressed — `a51f3d34`

| # | finding | fix |
|---|---|---|
| 1 | shared hazard assert vacuous on short `got` | `assert_eq!(got.len(), times.len())` first, **mutation-verified** |
| 2 | orphaned `equilibrate_ss_input_rate` doc | 19-line block moved back above its function |
| 3 | docs say `simulate()` unsupported | replaced with the real contract |
| 4 | "two spurious warnings" | corrected in CHANGELOG + both source docs |
| 5 | monotonicity assert could not fire | reordered before the closed form |
| 6 | `debug_assert!` in release | both upgraded to `assert!` |
| 7 | duplicated projection block | extracted `ChzProjection` |
| 8 | `restore_chz` silent zeroing | slot-bound + snapshot-length `debug_assert`s |
| 9 | missing `n == 0` guard | `n_pk_rows() == 0 → None` at both sites |
| 10 | `chz_at` hardcoded slot | reads `ode.chz_state_slots[0]` |

Three worth calling out.

**Finding 1 was mutation-verified, not just fixed.** Truncating `chz_at` by one element now
produces `SS bolus: the walk returned 3 hazard values for 4 requested times; zip below would
silently truncate and assert nothing`. Before this commit that mutation was green.

**Finding 8's new assertion caught a real defect on its first run.**
`a_joint_linear_model_takes_the_exact_ss_solve` clears `chz_state_slots` for its undeclared
arm but still passed `&[0.0]` — a length-1 snapshot against zero declared slots. Production
cannot do this (every caller passes a `chz_snapshot` of the same spec, so the lengths match by
construction), but the test could, and the silent `unwrap_or(0.0)` had been absorbing it. Fixed
to `&[]`.

**Finding 2 needed a real move, not a separator.** My first attempt put a `//` banner comment
between the two blocks; that does not work — line comments do not break a `///` run, so the
orphaned block would still have attached to `mask_chz`. The 19 lines had to move to the
function they describe.

On finding 6, the upgrade to `assert!` is deliberate rather than incidental: the guard exists
for a *future* widening of the FD routing, and the failure mode it catches is a silently wrong
gradient — FOCE/FOCEI converging to the wrong optimum rather than erroring. A tripwire that is
absent from release builds is absent from every build a user fits in.

Verified after the fixes: `tools/preflight.sh` green (fmt, 5 check sets, clippy, rustdoc,
docs-lint, public-API), `cargo test --lib` **4046 passed / 0 failed**, anchor **6/6**. The two
compiler warnings the check matrix reports (`elbo.rs:271`, `model_parser.rs:1228`) are
pre-existing and in code this PR does not touch.

CI on the previous head was fully green — all 10 checks including `codecov/patch`.

Issue #990, opened 2026-08-19
Two slow tests fail on macOS and pass on Linux: FREM FOCEI OFV 18027 vs 6.9, SSE-RTTE V 10.35 vs 9.50


roninsightrx:
## Summary

Two Tier-3 slow tests fail on **macOS/arm64** and pass on **Linux** (both the CI runner and a linux/arm64 container). Because `slow-tests.yml` runs on `ubuntu-latest` only, CI never sees them — they are visible only to someone running the slow suite locally on a Mac, where they read as a red suite that has nothing to do with the change under test.

| test | macOS/arm64 | Linux |
|---|---|---|
| `frem_warfarin::frem_covariate_omega_matches_sample_variance_under_focei` | **OFV = 18027.4** (assert: `< 1000`, expected ≈ 211) | OFV = 6.9 ✅ |
| `tte_convergence::joint_pktte_sse_recovers_pk_and_omega` | **V = 10.3532** (assert: `9.0..10.2`) | V = 9.4979 ✅ |

Both are pre-existing and unrelated to #988 — the numbers are bit-identical at `988de088` (its merge-base) and on the PR branch.

## Details

**1. `frem_covariate_omega_matches_sample_variance_under_focei` — a 3-orders-of-magnitude split, not a tolerance nit.**

```
macOS : FOCEI FREM: OFV=18027.4  omega_WT=105.80 (want 111.56)  omega_AGE=92.87 (want 99.38)
Linux : FOCEI FREM: OFV=6.9      omega_WT=108.96 (want 111.56)  omega_AGE=95.79 (want 99.38)
```

The assertion this trips (`tests/frem_warfarin.rs:658`) is the #-regression guard for the analytic gradient differentiating the PK likelihood on covariate pseudo-observation rows; its doc names ≈ 4901 as that failure's signature, and 18027 is neither that value nor the ≈ 211 optimum. The covariate ω's land in roughly the same place on both platforms while the OFV differs by 18000 units, which is the odd part — worth understanding before assuming it is only a basin difference.

**2. `joint_pktte_sse_recovers_pk_and_omega` — a flat ridge, plausibly a too-tight band.**

```
macOS : CL=1.0542 V=10.3532 KA=0.9979 H0=0.01192 BETA=0.4129 omega^2=0.0931 OFV=3859.008
Linux : CL=0.9795 V=9.4979  KA=0.9974 H0=0.01204 BETA=0.3822 omega^2=0.0869 OFV=3858.573
[...]

PR #1221, opened 2026-09-03
fix(ode): carry a fit's ODE solver settings to the integrator [closes #1212]

### LINKED ISSUE #1212: FitOptions ODE solver settings (ode_reltol, ode_method, …) are silently ignored by fit()
## Summary

Every ODE solver setting on `FitOptions` — `ode_reltol`, `ode_abstol`, `ode_max_steps`, `ode_method`, `ode_stiff_abort_after`, `ode_auto_switch` — is **silently ignored** when passed programmatically to `fit()`. The same settings written in the model file's `[fit_options]` block work.

The integrator reads `OdeSpec::solver_opts`, and the only production caller of `CompiledModel::sync_ode_solver_opts` is the parser (`src/parser/model_parser.rs:3175`), which stamps the spec at parse time from the model file. `fit()` takes `&CompiledModel` and never re-syncs, so a caller-supplied `FitOptions` value has no path to the solver.

No error, no warning — the fit runs to completion at the parse-time default and reports success.

## Evidence

Same model, same data, same seed; only the route the tolerance travels differs:

| route | `rtol = 1e-4` | `rtol = 1e-10` |
|---|---|---|
| `FitOptions { ode_reltol, ode_abstol, .. }` → `fit()` | V = 9.97757675, OFV = −5709.673500 | V = 9.97757675, OFV = −5709.673500 |
| model file `[fit_options] ode_reltol = …` | V = 9.97757675, OFV = −5709.673500 | **V = 9.97539957, OFV = −5771.985261** |

The model-file route moves the fit by 62 OFV. The programmatic route is bit-identical across six orders of magnitude — the value never reaches the integrator.

## Why this matters

`ode_reltol` is a documented public field on a public struct. Its own doc comment reads "RK45 ODE solver relative tolerance … Default `1e-4`" and names `[fit_options] ode_reltol` and `ferx_fit(settings = …)` as the ways to set it — without saying that the third obvious way, setting the field and calling `fit()`, does nothing.

The failure mode is a silent wrong answer, not an error:

- Someone tightening tolerance to chase an integration-noise hypothesis concludes the objective is tolerance-insensitive. That is exactly what happened in #990 — I ruled out integrator noise on a `(1e-4,1e-6)` → `(1e-10,1e-12)` sweep that was a no-op, and only caught it because a bit-identical result across six orders of magnitude is too clean to be real.
- `ode_method` is the sharper edge: a caller selecting `rosenbrock23`/`rodas4`/`rodas5p` for a stiff system through `FitOptions` silently stays on RK45. The symptom is a slow or failing fit with no indication the requested solver was not used.

The behaviour is **known and documented at the call site** — `model_parser.rs:3173-3175` says "ferx-core `fit()` takes `&CompiledModel` and does not" — so this is a design consequence rather than an accident. The bug is that it is silent and that the public field's own documentation does not mention it. `../ferx-r` works around it by calling `sync_ode_solver_opts` on its owned model (pinned by `types_tests.rs:2395`), which is why the R path is unaffected and why this has gone unnoticed.

## Scope

All six fields set by `CompiledModel::sync_ode_solver_opts` (`types.rs:3905`), for every `ode(...)` model and for the absorption ODE twin of a closed-form model with TV-cov / `TIME` / IOV subjects.

## Possible directions

Not obvious which is right — `CompiledModel` is not `Clone` (it carries closures), so `fit()` cannot cheaply take an owned copy to sync.

1. **Detect and warn.** Compare `options.ode_*` against the spec's baked `solver_opts` at the top of `fit_inner` and push a `FitResult.warnings` entry when they differ, naming the field and which value won. Cheap, non-breaking, and turns a silent wrong answer into a visible one — but leaves the setting still not working.
2. **Make it work.** Give `fit()` a path to an owned/synced spec (an `Arc`-swap on `OdeSpec`, an interior-mutability handle, or a `fit_with_model_mut`). Correct, but touches the `&CompiledModel` signature that ferx-r and every caller depends on.
3. **Reject.** Error out when caller-supplied ODE options differ from the compiled spec, forcing the model-file route. Honest, but breaks any caller currently passing values that are already being ignored (silently, so nobody would notice today).
4. **Document only.** Say on each field that it is read at parse time and that programmatic values need `sync_ode_solver_opts`. Weakest, but at minimum this should happen regardless of which of the above lands.

My inclination is (1) + (4) now, with (2) as the real fix if the signature change is acceptable.

## Reproducing

Fit any `ode(...)` model twice through `fit()` with `FitOptions { ode_reltol: 1e-4, .. }` and `{ ode_reltol: 1e-10, .. }` and compare the OFV to the last digit; then move the same two values into the model file's `[fit_options]` and repeat. The first pair is bit-identical, the second is not.

Found while root-causing #990.

roninsightrx:
## Why

Closes #1212.

Every ODE solver setting on `FitOptions` — `ode_reltol`, `ode_abstol`, `ode_max_steps`,
`ode_method`, `ode_stiff_abort_after`, `ode_auto_switch` — was **silently ignored** when passed
programmatically to `fit()`. The integrator reads `OdeSpec::solver_opts`, which the parser
stamps at parse time (`model_parser.rs:3175`); `fit` takes `&CompiledModel` and cannot restamp
it, so a caller-supplied value had no path to the solver. No error, no warning: the fit ran at
the parse-time value, on the default stepper, and reported success.

The failure mode is a silent wrong answer:

- A tolerance sweep from `(1e-4, 1e-6)` to `(1e-10, 1e-12)` returns a **bit-identical** objective,
  so "the objective is tolerance-insensitive" looks like a finding. That happened while
  root-causing #990, and was only caught because agreement to the last bit across six orders of
  magnitude is too clean to be real.
- `ode_method` is the sharper edge: a caller selecting `rosenbrock23` / `rodas4` / `rodas5p` for
  a stiff system stayed on the explicit stepper, with a slow or failing fit and no indication
  the requested solver was not used.

Two in-tree NONMEM anchors were already victims. `tests/tad_lag_nonmem_anchor.rs` and
`tests/mr_tad_nonmem_anchor.rs` set `ode_method: Rk45, ode_reltol: 1e-9, ode_abstol: 1e-11`, with
a comment stating that the stepper is *"the stepper actually pinned: `FitOptions::default()` is
`OdeMethod::Auto`, so without this line a change to the stiffness probe could move the anchor."*
Neither line did anything; both anchors were running at `1e-4` under `auto`. They now run as
written and still match NONMEM.

## What changed

`fit()` arms a fit-scoped `OdeSolverOverride` for the duration of the run, and the 19 integration
read sites go through `OdeSpec::effective_solver_opts()`, which merges it over the spec's baked
value.

- **Same shape as the fit-scoped state already next to it.** `fit` line 243 already calls
  `set_inner_optimizer` / `set_ebe_warm_start`, process-globals for the documented reason that
  "the inner loop fans out over subjects via rayon and they all read one fit setting". The
  override is one more, with the same consequence for two concurrent fits carrying *different*
  ODE options in one process — `ferx-tools` runs its parallel replicate fits off a single shared
  `FitOptions`, so this is the existing shape, not a new constraint.
- **Behind an RAII guard that restores the previous value**, not one that clears. A nested or
  chained fit is therefore correct (innermost wins, its exit hands the setting back), and an
  early `?` out of `fit` cannot leak a tolerance into a later `predict`. An *empty* override
  still shadows an outer one, because "this fit expressed no opinion" has to mean the spec's
  baked value rather than the enclosing fit's.
- **Armed in `fit`, not `fit_inner`**, so pre-fit validation, every multistart stage, the
  covariance step and the post-fit diagnostic sweep all integrate at the same options.

### The merge rule, and why it is per-field

A field is carried **only when it differs from `FitOptions::default()`**. This is the half of the
fix that is not obvious, and it exists to avoid trading the reported bug for its mirror image: a
caller who hand-builds a `FitOptions` and never touches `ode_reltol` must not silently *loosen* a
model file that pinned `ode_reltol = 1e-10` back to the `1e-4` default. That would be the same
class of silent wrong answer, pointed the other way, and it would hit callers who are correct
today.

The rule is sound because the `ode_*` `FitOptions` defaults are identical to
`OdeSolverOptions::default()`'s, so "untouched" is exactly "asked for the solver default". That
coincidence is now a test (`fit_option_ode_defaults_match_the_solver_defaults`) rather than an
accident — changing one side alone would silently change the merge for every caller.

The one case this cannot resolve: a caller who explicitly wants the default while the model file
pinned something else is indistinguishable from a caller who set nothing, and the file wins. That
errs tight rather than quietly reducing accuracy, and it is documented on the field, in
`docs/model-file/fit-options.qmd`, and in the CHANGELOG.

## Alternatives considered

Three other directions were weighed in the issue (see [my investigation
comment](https://github.com/FeRx-NLME/ferx-core/issues/1212#issuecomment-5527892977)):

- **Detect and warn only.** Cheap and non-breaking, but leaves the setting not working — the
  issue asks for a documented public field to do what it says.
- **Gate the override on explicitly-set keys.** `FitOptions::user_set_keys` exists and is `pub`,
  but it is populated only by `parse_fit_options` / `apply_fit_option`. No programmatic caller
  pushes to it, so gating on it as-is means the override never fires for exactly the callers this
  issue is about. It would need a set-tracking builder first — a separate design step, and one
  the default-comparison rule makes unnecessary for every case except "explicitly asking for the
  default".
- **Reject a mismatch.** Honest, but every affected caller is passing values that are ignored
  today, so it breaks working-looking code to fix a silent no-op.
- **Take `&mut CompiledModel` / an owned model.** Correct, but changes the signature ferx-r and
  every caller depends on, and `CompiledModel` is not `Clone` (it carries closures).

**Scope note.** `gradient_method` and `bloq_method` are the same *pattern* — `FitOptions` values
mirrored onto `CompiledModel` and stamped only by an owned-model entry point — and
`gradient_method` is worse in one respect: the parser does not stamp it, so
`parse_full_model_file()` + `fit()` drops the model file's `gradient = fd` too. `types.rs:3513`
already documents that hazard. They are deliberately **not** in this PR: the ODE fields are read
off a spec by an integrator and needed this mechanism, while those two are plain
`CompiledModel` fields whose fix is a different (and smaller) change. Happy to open a follow-up.

## Cross-repo dependency

ferx-r calls `sync_ode_solver_opts` on its owned model before `fit` (pinned by
`types_tests.rs:2395`, which is why the R path was never affected). That still works and is
still correct: it produces values equal to the ones the override carries, so the merge is a
no-op there. No public API changed.

## Breaking changes
- [x] None

Every new item is `pub(crate)` — `api/ferx-core-public-api.txt` is unchanged and the
`Public API baseline` job is green.

One **behavioural** change worth calling out for reviewers, since it is the point of the PR: a
caller who was already passing non-default `ode_*` to `fit()` now gets what they asked for. In
this repo that is the two NONMEM anchors above, which now run at `rk45` / `1e-9` as their own
comments claim. Both still pass.

## Numerical validation
- [x] Compared against: NONMEM, via the two existing anchors, which this change materially
      affects (they were silently running at the wrong tolerance and stepper)

```
# tests/tad_lag_nonmem_anchor.rs  (NONMEM MAXEVAL=0 POSTHOC; OFV -271.990 / -169.317)
tad_lag_single_dose_matches_nonmem ... ok
tad_lag_two_dose_matches_nonmem ... ok

# tests/mr_tad_nonmem_anchor.rs
mr_shaped_tad_rhs_matches_nonmem ... ok

# run at ode_method = rk45, ode_reltol = 1e-9, ode_abstol = 1e-11 for the first time —
# before this PR the same source ran at auto / 1e-4 / 1e-6.
```

The new API-level test is itself a before/after: with the arming line replaced by an empty
override, `a_caller_supplied_ode_tolerance_reaches_the_integrator` and
`an_untouched_ode_tolerance_keeps_the_model_files_pinned_value` both fail; with it, both pass.

## Performance

Outside a fit the read path is one relaxed `AtomicBool` load and returns the field. Inside a
fit it adds an uncontended `RwLock` *read* per integration **call** (not per step) — the sites
all bind `let opts = ode.effective_solver_opts();` once and pass it down by value.

## Tests

**Tier 1, `src/ode/solver.rs`** — the defaults-parity invariant the merge rests on, the per-field
merge (a moved field wins; untouched fields keep the *baked* value, not the `FitOptions` one;
fields no fit option addresses are carried verbatim), and arm/restore including nesting and the
empty-override-shadows case.

**Tier 1, `src/api/tests/ode_solver_options_tests.rs`** — three `fit()`-level cases, asserted on
the OFV because only the objective can say whether the value reached the solver: the caller's
tolerance moves it, an untouched caller keeps the model file's pinned value (and the two routes
to the same effective tolerance agree bit-for-bit), and the setting does not outlive the fit.

**`tests/ode_solver_opts_reads_go_through_the_override.rs`** — a structural guard. Nineteen call
sites were converted at once and nothing in the type system stops the twentieth from being
written the old way, so this fails if production code reads `.solver_opts` directly. Verified by
reverting one site: it names `markov/endpoint.rs:421` and fails.

## Docs

`docs/model-file/fit-options.qmd` gains a paragraph on precedence in the ODE tolerances section.
`FitOptions::ode_reltol`, `CompiledModel::sync_ode_solver_opts` and the parser's stamping comment
all said, correctly at the time, that `fit()` ignores these; they now describe what actually
happens.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `sens/ode_provider.rs`'s
      two sites changed only where they *read* the options; no formula touched

Also run: `cargo test --lib --features ci` (3890 passed), `--features ci,markov` (4068 passed,
covering the survival `MonoTol` and CTMM sites), and the two slow NONMEM anchors under
`--features ci,slow-tests`.

## Reviewer hints

1. **The default-comparison merge rule** (`FitOptions::ode_solver_override`, `types.rs`). This is
   the judgement call in the PR — see "why it is per-field" above. If you would rather the
   caller's `FitOptions` be authoritative wholesale, that is a real alternative, but it silently
   loosens a pinned model file for anyone passing a hand-built options struct.
2. **The guard's Drop restoring `prev` rather than clearing** (`ode/solver.rs`). Clearing would
   be wrong for a nested fit and is the kind of thing that only shows up under `ferx-tools`.
3. **The sites I judged should read the *effective* options rather than the baked field**:
   `verify_adaptive_frozen_replay`'s agreement band, and `modified_release.rs`'s `verify_tol` /
   twin-admission gate. Both derive a tolerance for comparing against a twin that is itself
   integrated at the effective options, so deriving them from the parse-time field would compare
   at the wrong scale.

The other 16 sites are mechanical (`let opts = ode.solver_opts;` → `ode.effective_solver_opts()`).

## Open questions

Should the `gradient_method` / `bloq_method` half of this pattern be a follow-up issue, or folded
in here? I left it out to keep the diff to one mechanism, but `gradient_method` silently dropping
a model file's `gradient = fd` on the `parse` + `fit` path is arguably the more surprising of the
two.

https://claude.ai/code/session_011gZZkCcMZpwT5kntcndb7J


roninsightrx:
## Code review — 7 findings

### 1. HIGH — the save/restore guard corrupts under concurrent `fit()` calls, and can permanently leak the override
`src/ode/solver.rs:404` (and `:414`)

`arm_ode_solver_override` captures `prev` and `Drop` writes it back. That is only correct for LIFO nesting on a single thread. `crates/ferx-tools/src/bootstrap/mod.rs:1153/1155` runs `todo.par_iter().filter_map(run_one)` where `run_one` calls `fit()` (line 1084; likewise line 1372 under the `par_iter` at 1366) — N concurrent `fit()`s, each arming and dropping this global.

An A-arm / B-arm / A-drop / B-drop interleaving gives two defects:

- **(a)** A's drop restores `A.prev = None` and stores `ARMED = false` *while B is still fitting*, so B's remaining integrations silently fall back to the parse-time options — the exact #1212 bug, resurrected.
- **(b)** B's drop writes back `B.prev = Some(ov)` with no guard left holding it, so the override stays armed for the rest of the process. In a long-lived ferx-r session, every `predict()`/`simulate()` after a bootstrap then integrates at the bootstrap's tolerance/method.

The PR description's argument ("ferx-tools runs off a single shared `FitOptions`") covers the *value* but not the arm/disarm ordering. Note `set_inner_optimizer` is not the same shape: it never restores, so it cannot disarm or leak.

Fix: tag each guard (generation counter or token) and have `Drop` restore only if the slot still holds its own arming, or make the slot a per-thread value broadcast into the rayon pool.

### 2. MEDIUM — the structural guard is vacuous for `src/ode/predictions.rs`, the file holding 10 of the 19 converted sites
`tests/ode_solver_opts_reads_go_through_the_override.rs:55`

`src.find("#[cfg(test)]")` matches the *first textual occurrence*, and `src/ode/predictions.rs:17` is a plain comment containing the literal `#[cfg(test)]`. So `production` is lines 1–16, and the remaining ~6,385 lines — including every converted integration site in that file — are never scanned. A future 20th site written as `let opts = ode.solver_opts;` in `predictions.rs` passes the guard silently. (The verification in the PR description used `markov/endpoint.rs:421`, whose first `#[cfg(test)]` is at line 494, so it happened to work.)

Fix: match on a line that *is* the attribute (`line.trim() == "#[cfg(test)]"`), or strip comments before searching. Related, lower severity: the walk covers only `src/`, not `crates/*/src`.

### 3. MEDIUM — the new tests mutate a process-global while libtest runs the lib binary's tests in parallel
`src/api/tests/ode_solver_options_tests.rs:139` (and `:175`)

There is no `serial_test` dependency or shared mutex in this repo, and the flake is entirely within this one file:

- `a_caller_supplied_ode_tolerance_reaches_the_integrator`'s `loose` fit arms an *empty* override, which by design sets the slot to `None` and `ARMED = false`. If that lands while `an_untouched_ode_tolerance_keeps_the_model_files_pinned_value`'s `from_caller` fit is in flight, that fit is disarmed mid-run and evaluates at `1e-4`, so `assert_eq!(from_file.to_bits(), from_caller.to_bits())` fails.
- `a_fits_ode_options_do_not_outlive_it` reads the *global* through `effective_solver_opts()` after its own fit; any concurrently-armed test makes `after.reltol != baked.reltol`.

Both directions also perturb unrelated ODE tests in the same binary. Needs a shared test mutex (or `--test-threads=1` scoping).

### 4. MEDIUM — `run_covariance` / `run_sir` / `run_sir_core` take `&FitOptions`, integrate, and never arm the override
`src/estimation/run_covariance.rs:86`

All three are public API (`api/ferx-core-public-api.txt:286,288,319`). A caller passing `FitOptions { ode_reltol: 1e-10, .. }` to `run_covariance` still gets the parse-time tolerance, with no error — #1212 is still live on that path.

It is worse after this PR: a fit run with a caller-supplied tight tolerance and `run_covariance_step: false`, followed by a standalone `run_covariance(&result, …, &same_options)`, now computes the FD-of-OFV Hessian at a *different* accuracy than the OFV surface the estimates came from — and CLAUDE.md documents the covariance step as the tolerance-sensitive one. Arming in these entry points is a one-line addition each.

### 5. LOW/MEDIUM — `fit()` output and a subsequent `predict()` on the same `&CompiledModel` can now silently disagree
`src/api/fit.rs:250`

Previously both always ran at the baked spec options. Now, with a caller-supplied non-default `ode_reltol`/`ode_method`, the fit's sdtab IPREDs are computed under the override while a later `predict(&model, params)` runs at the parse-time value, with nothing to indicate the two used different integrators. Worth either a warning, or docs pointing callers at `sync_ode_solver_opts` on an owned model when they intend both paths to match.

### 6. LOW — while armed, every integration call takes a `std::sync::RwLock` read
`src/ode/solver.rs:375`

`effective_solver_options` is called from `integrate_segment` (`predictions.rs:2058`) and nine other per-call sites, so under the FOCE inner loop's rayon fan-out this is a shared-cache-line atomic RMW at high frequency across all workers — precisely in the case the fix exists to enable (armed). `OdeSolverOverride` is `Copy` and is only written between fits; a lock-free representation (per-field atomics, or `ArcSwap`) would remove the contention. The "one relaxed load" cost analysis holds only on the *unarmed* path.

### 7. LOW — the precedence paragraph is not accurate for the R path it explicitly names
`docs/model-file/fit-options.qmd:97`

It says the rule applies to "`ferx_fit(settings = ...)` in R", and that "asking explicitly for a default value (`ode_reltol = 1e-4`) reads the same as not asking, and the model file still wins." ferx-r calls `sync_ode_solver_opts` on its owned model before `fit` (pinned by `types_tests.rs:2395`), which stamps all six fields unconditionally — so an explicit `1e-4` from R *does* overwrite a file's `1e-10`. Either scope the sentence to the Rust `fit()` path, or verify and restate the R merge semantics.

https://claude.ai/code/session_011gZZkCcMZpwT5kntcndb7J

roninsightrx (commented):
Changes are needed before merge.

Blocker: `arm_ode_solver_override` at `src/ode/solver.rs:404-420` uses save/restore semantics on one process-global slot. This is valid only for strict LIFO nesting, but `ferx-tools` calls `fit()` concurrently from Rayon at `crates/ferx-tools/src/bootstrap/mod.rs:1149-1155` and `:1366-1373`. With A-arm / B-arm / A-drop / B-drop, A disables the override while B is still fitting and B then restores `Some(ov)` after every guard is gone. Thus live fits can revert to baked options and later `predict()` / `simulate()` calls can inherit a leaked fit override, even when all concurrent fits use identical options.

I reproduced the same race using the PR tests themselves. The default parallel runner fails:

`cargo test --lib ode_solver_options -- --nocapture`

`a_caller_supplied_ode_tolerance_reaches_the_integrator` observed bit-identical loose and tight OFVs and failed. Running the same three tests with `--test-threads=1` passes. The override must be made concurrent-fit-safe, and its tests need deterministic concurrency coverage rather than relying on serialized execution.

Two additional gaps should be addressed with the fix:

- The structural guard truncates `src/ode/predictions.rs` at line 17 because it searches for the first textual `#[cfg(test)]`, which appears there inside a comment. Consequently it scans none of the ten converted production sites in that file. Match an actual trimmed attribute line instead.
- The public `run_covariance`, `run_sir`, and `run_sir_core` APIs also accept `&FitOptions` and integrate ODE models, but do not arm the override, so the same `ode_*` fields remain silently ignored on those standalone paths.

Validation performed: workspace/all-targets check with `ferx-core/ci` passed; the structural test passed but is subject to the false-negative above; the focused API tests failed in parallel and passed serially.

roninsightrx (commented):
Correction for current head `e6581e76`: this commit landed while my first pass was running and fixes the prior save/restore leak, the flaky focused tests, the structural-test cutoff, and the standalone covariance/SIR entry points. That earlier review text is superseded for those items. Two blockers remain in the replacement design.

1. HIGH — concurrent fits still receive ODE options from another fit on their Rayon workers (`src/ode/solver.rs:386-393`, `:531-540`). The thread-local protects only the thread that called `fit`; every worker reads the last non-empty override armed anywhere in the process. Therefore a default-options fit running beside a non-default fit inherits the other override on all workers, and two non-default fits overwrite each other. This is not bounded in the safe direction: `ode_reltol` accepts any positive finite value, so a later fit can publish `1e-2` over a fit that requested `1e-10`; a method-only override can also replace a tolerance override entirely; lower `ode_max_steps`, `ode_stiff_abort_after`, and named methods have no tighter/coarser ordering. It can also loosen a model-file value, for example publishing `1e-10` onto a concurrent model baked at `1e-12`. `src/api/pool.rs:112-125` explicitly supports concurrent callers sharing the fit pool, so this leaves the exact silent-ignore class from #1212 on a supported path. Each worker needs a per-fit context, or calls with overrides must otherwise be isolated so they cannot observe another fit.

2. MEDIUM — the atomic mirror is not a coherent snapshot (`src/ode/solver.rs:482-518`, `:549-567`). A writer updates the component atomics while the old nonzero `OV_SET` mask remains published, then stores the new mask. A reader may acquire the previous mask just before the final store and subsequently read a mixture of newly-written and previous component values. The release store only synchronizes readers that observe that release; it does not protect a reader that observed the prior generation. Two concurrent overrides can therefore produce a tolerance/method/max-step combination that no caller supplied. Use a seqlock-style generation with retry, an immutable atomically-swapped snapshot, or the lock until profiling demonstrates it is a bottleneck.

Current-head validation passed: `cargo test --lib ode_solver_options`, all 51 `ode::solver::tests`, and both structural guard tests. These tests do not cover the first defect: `an_empty_arming_on_another_thread_leaves_a_running_fit_alone` reads on the two arming threads, while the failing case is a third worker thread serving the empty fit; the source comments explicitly acknowledge that workers cannot identify their originating fit.

roninsightrx:
Both blockers addressed in `72255e40`. The thread-local-plus-global-set design is gone; the whole mechanism is now one thread-local and a pool that carries it.

**1 — concurrent fits reading each other's options on the workers.** You were right that this is not bounded in the safe direction, and that a default-options fit inheriting another fit's override is the same silent-ignore class as #1212 itself.

The fix stops asking a worker which fit it serves. A fit that set any `ode_*` runs on a **pool keyed by that override**, whose workers had the value installed at thread start (`ode_override_pool`, via `start_handler`). A worker there serves only fits carrying exactly that override, so there is one answer per thread and no cross-fit read left to be wrong. A fit that set nothing keeps using the shared pool, whose workers never install an override and therefore integrate at the model file's options no matter what runs beside them.

Pools are cached per `(override, width)` and kept for the process lifetime, for the reason `default_fit_pool` is shared: bootstrap arms one `FitOptions` across hundreds of replicate fits, and a fresh `N × 32 MiB` pool per replicate would oversubscribe CPUs and churn address space. A pool that cannot be built is an `Err` rather than a fallback to running unpooled — unpooled workers would integrate at the baked options and the fit would still report success.

`fit()`'s two pool choices now go through one `install_on_fit_pool` (override pool first, otherwise exactly the previous pinned/shared/ambient behaviour). `run_covariance` / `run_sir` / `run_sir_core` take the same scope via `with_fit_ode_scope`, which skips re-entering a pool the current thread is already a worker of, so an inline covariance step under a fit does not stack a second one.

**2 — the mirror was not a coherent snapshot.** Correct, and it is deleted rather than repaired: the process-global set, the `OV_SET` mask and the per-field atomics are all gone. The read path is a `Cell::get` on a thread-local, so there is no generation to tear across.

**On the test gap you identified** — that `an_empty_arming_on_another_thread_leaves_a_running_fit_alone` read on the two arming threads and never on a third worker. The replacements assert from the threads that actually matter:

- `concurrent_fits_do_not_see_each_others_overrides` — four fits live at once (three different tolerances, one that asked for nothing), each checking it still reads its own while all four arms are held.
- `only_the_pool_built_for_an_override_carries_it` — a **shared-pool worker** must read the baked value while a tight override is armed on another thread, and an **override-pool worker**, including inside a nested `par_iter`, must read the override having armed nothing itself.
- `every_overridable_field_reaches_a_pool_worker` — all six fields round-trip onto a worker, including `stiff_abort_after: Some(Some(0))` vs `Some(None)` and `auto_switch: false`.
- `the_override_pool_table_is_keyed_by_the_override` — two overrides never share a pool; one override never builds two.

Each was checked against the mutation it exists to catch: making the storage a process-global fails the concurrency tests, dropping the worker install fails the pool tests, and skipping the override pool in `install_on_fit_pool` fails the end-to-end API tolerance tests — which also confirms the fit-level fan-out genuinely depends on the workers carrying it, rather than the tolerance change happening to be visible on the calling thread.

The serialising test lock is gone with the global it was working around. `cargo test --lib ode_solver_options` passes under the default parallel runner (your reproduction), as does the full `cargo test --lib` (3913 tests) and three repeats of `cargo test --lib ode`, plus `tools/preflight.sh`.

https://claude.ai/code/session_011gZZkCcMZpwT5kntcndb7J

PR #1224, opened 2026-09-03
feat(api): BIC variants (mixed / iiv / random) and a Strictness gate for candidate ranking [closes #1177]

### LINKED ISSUE #1177: feat(core): BIC variants (mixed / iiv / random) and a first-class `Strictness` predicate for candidate ranking [#1175]
Part of #1175 (P0 — prerequisites).

## Problem

Two things a search needs for ranking do not exist in usable form.

**1. BIC variants.** ferx computes one BIC, `ofv + p·ln(n_obs)` (`src/api/fit.rs`). Pharmpy's `iivsearch` and `modelsearch` rank on the Delattre-style **mixed / iiv / random** variants, where the θ-class term uses `n_obs` and the η-class term uses `n_subjects`. Ranking IIV structures on the observation-count BIC systematically favours the wrong model.

**2. Strictness.** `FitResult.converged` is a bool. It does not distinguish a genuine optimum from an init stall (#751), a boundary estimate, an ill-conditioned covariance step or a near-singular correlation matrix. Under automation those all become *model-selection errors*: a candidate that never left its initial estimates is ranked on an OFV that says nothing about the model.

## Proposal

In `ferx-core`:

```rust
pub enum BicType { Mixed, Iiv, Random, Fixed }
pub fn bic(result: &FitResult, kind: BicType, n_subjects: usize) -> f64;

pub struct Strictness {
    pub require_converged: bool,
    pub require_covariance: bool,
    pub max_condition_number: Option<f64>,   // pyDarwin default 1000
    pub max_correlation: Option<f64>,        // pyDarwin default 0.95
    pub reject_on_boundary: bool,
    pub reject_init_stall: bool,
}
pub struct StrictnessVerdict { pub passed: bool, pub failures: Vec<String> }
pub fn check_strictness(result: &FitResult, s: &Strictness) -> StrictnessVerdict;
```

The condition number and the maximum off-diagonal correlation both come off `FitResult.covariance_matrix`; the boundary check is the same predicate `bootstrap`'s `skip_estimate_near_boundary` already applies; the init-stall check is the #751 signature.

`StrictnessVerdict` carries the *reasons*, not just a bool, so a search report can say why a candidate was excluded rather than silently dropping it.

## Acceptance

- [ ] Tier-1 unit tests pin each BIC variant to hand-computed values **and** to `pharmpy.modeling.calculate_bic()` on the same fit.
- [ ] Each strictness predicate has a unit test with a fit that fails it and one that passes.
- [ ] A candidate that stalls at its initial estimates fails strictness, with a named reason.
- [ ] Public-API baseline updated in the same PR.

roninsightrx:
## Why
Part of #1175 (P0 prerequisites). A search is *generate candidate → fit → rank*, and two ranking inputs did not exist in usable form:

1. **BIC variants.** ferx computed one BIC, `OFV + p·ln(n_obs)`. Pharmpy's `iivsearch` / `modelsearch` rank on the Delattre (2014) **mixed** BIC, where random-effects-class parameters are penalised on `ln(n_subjects)`. Ranking IIV structures on the observation-count BIC systematically favours the wrong model.
2. **Strictness.** `FitResult::converged` is a bool. It cannot distinguish a genuine optimum from an init stall (#751), a boundary estimate, an ill-conditioned covariance step or a near-singular correlation matrix — all of which become *model-selection errors* under automation.

Closes #1177.

## What changed
New module `src/model_selection.rs`, re-exported at the crate root:

- `bic(&FitResult, BicType::{Mixed, Iiv, Random, Fixed}) -> f64` — the four conventions of `pharmpy.modeling.calculate_bic`. `Fixed` reproduces `FitResult::bic` bit-for-bit. Operates on the result alone (no model/data), so `.fitrx` bundles rank directly.
- `FitResult::bic_inputs: BicInputs` — the free-parameter tally by Delattre class (`theta_random`, `theta_fixed`, `omega`, `kappa`, `sigma`, `sigma_random`, `n_obs`), filled by `fit()` from the packed held mask, round-trips through `.fitrx` (`serde(default)`; pre-existing bundles yield `NaN` rather than a wrong penalty).
- `CompiledModel::theta_eta_linked: Vec<bool>` — parse-time θ→η linkage, the statement-level analogue of Pharmpy's `_categorize_parameters`: transitive through intermediate `[individual_parameters]` assignments, `if` conditions contribute to the parameters assigned inside, `[covariate_model]` thetas classified after desugaring, κ counts as a random effect. σ joins the random class only under `iiv_on_ruv` (Pharmpy's rule: an η in the DV expression).
- `Strictness` (pyDarwin defaults: converged required, covariance not required, condition number ≤ 1000, |corr| ≤ 0.95, reject boundary, reject init stall) + `Strictness::none()`, `StrictnessVerdict { passed, failures, skipped }`, `check_strictness()`. Threshold gates with no input (no covariance matrix) land in `skipped` rather than failing or silently passing; pair with `require_covariance` to make them mandatory.
- Standalone predicates: `estimate_near_boundary` (the bootstrap `skip_estimate_near_boundary` predicate — bootstrap now calls it), `stalled_at_init` (natural-scale reading of the #751 `INIT_ESCAPE_STEP_S` test: no free θ/Ω/σ moved > 1% of its init; at least as sensitive to movement as the optimizer's own test, never less), `max_abs_correlation` (whole covariance matrix, as pyDarwin/Pharmpy do — unlike the θ-only `HighCorrelation` warning).

**Deviation from the issue's sketch:** `bic()` takes no `n_subjects` argument — `FitResult::n_subjects` already carries it, and a redundant argument is a way to pass the wrong one.

**Bug fixed en route:** `fit()` counted `n_parameters` from `packed_fixed_mask` only, so a mixed `block_omega` + diagonal `omega` had its cross-block structural zeros counted as estimated parameters in AIC/BIC (the covariance step and `CompiledModel::free_packed_dim` already excluded them). New `packed_held_mask` (FIX ∪ structural zero) is the single source for `n_parameters`, the tally, and `free_packed_dim`.

## Alternatives considered
- `bic(&CompiledModel, &FitResult, …)` — rejected: the tally at fit time is what lets a stored bundle be ranked without the model, and ferx-r gets it for free.
- Deriving the θ class from `eta_param_info.linked_theta` — insufficient: it misses covariate thetas and any θ reaching an η through an intermediate assignment (pheno's `THETA(3)` in `TVV`).
- `reject_on_boundary` also covering `ParameterAtRunawayGuard` — not needed: #1118 already demotes `converged` on a guard hit, so `require_converged` catches it, and bootstrap parity keeps the predicate identical.

## Cross-repo dependency
| ferx (R) | FeRx-NLME/ferx-r#327 (issue) | **required** follow-up: three `FitResult` literals in `src/rust/src/lib.rs` need the new fields, then the lock bump | after |
| ferx (R) | FeRx-NLME/ferx-r#326 (issue) | surface `bic()` / `check_strictness()` on the R side | after |

`FitResult` gains four fields, all with serde defaults (`bic_inputs`, `left_init`, `omega_is_diagonal`, `kappa_is_diagonal`); existing ferx-r *readers* are unaffected, but ferx-r builds `FitResult` literals in three places, which fail with `missing field` until ferx-r#327 lands — that is the required follow-up. ferx-r#326 tracks exposing `bic()` / `check_strictness()` / `bic_inputs` on the R side once the lock is bumped.

## Breaking changes
- [x] `FitResult` / sdtab fields changed — additive only (`bic_inputs`, `left_init`, `omega_is_diagonal`, `kappa_is_diagonal`; all serde default, all round-trip through `.fitrx`); sdtab unchanged
- [x] Public Rust API — baseline regenerated. Added items and who needs them: `model_selection::{bic, BicType, check_strictness, Strictness, StrictnessVerdict}` (every #1175 search tool: rank + gate), `estimate_near_boundary` (ferx-tools bootstrap, now shared), `stalled_at_init` / `max_abs_correlation` / `INIT_STALL_REL_TOL` (search reports print the numbers, not just the verdict), `BicInputs` + `FitResult::bic_inputs` (the tally `bic()` reads; ferx-r will surface it), `CompiledModel::theta_eta_linked` (the parse-time classification `fit()` records; ferx-r/edit layer can inspect it). Added in review follow-ups: `FitResult::left_init` + `OuterResult::left_init` (the optimizer's own #751 verdict, which `stalled_at_init` prefers), `FitResult::omega_is_diagonal` / `kappa_is_diagonal` (the packed layout `natural_scale_covariance` needs — inferring it from the matrix size is wrong once mixture / `block_sigma` coordinates trail it), `natural_scale_covariance` (the delta-method transform the correlation gate reads block-Ω through; a report can print it), `MIN_SIR_FALLBACK_ESS` (the floor `require_covariance` applies to a SIR fallback), `MlpMapper::weight_indices_for_output` (which `[covariate_nn]` weights an output head depends on, for the θ classification).
- [ ] None

<details>

`n_parameters` / AIC / BIC change **only** for models with a mixed `block_omega` + diagonal `omega`, by `ln(n)` per structural zero. Any pinned value for such a model needs re-pinning.

</details>

## Numerical validation
- [x] Compared against: **Pharmpy `calculate_bic()`** on its `pheno` example (`tests/testdata/nonmem/pheno_real.{mod,ext}`: 59 IDs, 155 obs, OFV 586.27605628188053, θ×3 all η-linked, Ω×2, σ×1). Tier-1 tests pin all four variants:

```
# pharmpy docstring                      # ferx bic()
mixed   611.7071686216575                611.70716862 (Δ 3.3e-9)
fixed   616.5366069867251                616.53660698 (Δ 3.3e-9)
random  610.741280948644                 610.74128095 (Δ 3.3e-9)
iiv     594.4311311730211                594.43113117 (Δ 3.3e-9)
```

The constant 3.3e-9 is Pharmpy's OFV source rounding (it does not read the `.ext` value verbatim); the test anchors the between-variant differences to 1e-12 — which cancels that offset and is what pins the *penalties* — and the absolutes to 1e-8. Hand-computed penalties with κ, block Ω and `iiv_on_ruv` are pinned separately.
- [ ] Warfarin example — unchanged (diagonal Ω; `n_parameters` identical)
- [ ] Not applicable

## Performance
- [x] No significant impact expected — one extra mask walk at fit start; the predicates are O(p²) on the covariance matrix.

## Tests
- [x] Unit tests added in the same module (`cargo test --lib --features ci,survival` passes: 4162+)
  - `src/model_selection_tests.rs` — the pheno anchor, hand-computed variants, NaN/degenerate cases, tally partition invariant (`n_free() == free_packed_dim()`), and one pass + one fail test per strictness gate, including "a candidate that stalls at its initial estimates fails with a named `#751` reason".
  - `src/parser/model_parser_theta_eta_linked_tests.rs` — direct, transitive, if-condition, covariate-model and κ linkage.
- [x] Regression test that fails without the fix: `tally_counts_block_omega_elements_and_iov_kappas` (Ω = 4 not 6; and `n_free()` must equal `free_packed_dim()`), plus the Tier-2 `tests/n_parameters_free_packed_dim.rs` that goes through `fit()` (reverting to `packed_fixed_mask` reads 10 vs 8). Every regression test added in the review follow-ups was mutation-checked; see the review threads.

- [x] Not applicable — library API; usage in `docs/api/fitting.qmd`.

## Docs
- [x] `docs/api/fitting.qmd` — new "Ranking candidates: BIC variants and strictness" section; `docs/api/types.qmd` (`bic_inputs`, `n_parameters` note); `docs/output.qmd` (bic bullet points to the variants).
- [x] `CHANGELOG.md` `[Unreleased]`: Added (feature) + Fixed (structural-zero count).
- [ ] New pages — none.

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy, rustdoc, docs lint, public-API baseline
- [x] Nothing on the `Dual2` sensitivity path touched

### ferx-site / ferx-book
- [x] No new DSL syntax, example, or data file.

### Example execution
- [x] No example execution step needed — no `.ferx` behaviour change (except the `n_parameters` fix above, which affects no shipped example: none mixes a block and a diagonal Ω).

## Reviewer hints
- `classify_theta_eta_linked` (parser) vs Pharmpy `_categorize_parameters`: the one deliberate difference is that Pharmpy also scans the DV expression for thetas; ferx's error model carries no θ, and σ's class is decided by `residual_error_eta` instead.
- `stalled_at_init` tolerance semantics are in its doc comment: relative 1% (absolute 1e-2 for a zero init). κ is not compared — `FitResult` has no κ initials.
- The `packed_held_mask` refactor: `free_packed_dim` now delegates to it; `fit()` switched from `packed_fixed_mask`. Grep for other `packed_fixed_mask` callers if you want to check none of them meant "free".

## Open questions
- ~~`require_covariance` treats `SirFallback` as a failure~~ — resolved: a SIR fallback that delivered credible intervals is **accepted** (second commit). With no matrix stored on that path, the condition-number and correlation gates report `skipped`.

https://claude.ai/code/session_01K9CLrGo5hHuVoUDD12bX5J


roninsightrx:
## Code review (`/code-review high`)

13 candidates verified: 10 confirmed, 1 plausible, 3 refuted. Ranked most severe first. Fixes for all of these follow on this branch.

### Correctness

1. **`src/parser/model_parser.rs:1079` — `classify_theta_eta_linked` treats `LevelRule::NegSum(a, b)` as inclusive.** It uses `(a as usize)..=(b as usize)` while every other consumer of `NegSum` is half-open `θ[a..b]` (enum doc, `eval_gather`, `gather_g`, `axes_for_theta`). The binder emits `NegSum(free_base, next)` with `next` already one past the group, so the first θ declared *after* a sum-to-zero level group on an η-bearing parameter is wrongly marked η-linked and penalised on `ln(n_subjects)` instead of `ln(n_obs)` under `BicType::Mixed`. No test in `model_parser_theta_eta_linked_tests.rs` uses a `ThetaGather`, so `..=` → `..` is an unobservable mutation.

2. **`src/parser/model_parser.rs:2818` — `theta_eta_linked` is computed from `[individual_parameters]` only.** `[event_model]`, `[binary_model]`, `[markov_model]`, `[structural_model]` and `nn` outputs all accept ETA directly. Shipped examples hit it today: `examples/tte_weibull.ferx` has no `[individual_parameters]` and writes `scale = TVSCALE * exp(ETA_SCALE)` in `[event_model]` (likewise `tte_gompertz`, `rtte_exponential`, `rtte_weibull_reset`). For those models every random-effect θ is penalised on `ln(n_obs)` — events × subjects under RTTE — so a survival-structure search on `Mixed` over-penalises exactly the θs Delattre's convention shields, and moving the same line into `[individual_parameters]` flips the answer.

3. **`src/model_selection.rs:445` — `stalled_at_init` has no optimizer-side signal.** It only compares final θ/Ω/σ against the chain's starting values with a 1 % tolerance, so a warm-started candidate that legitimately converges within 1 % of its inits (parent-estimate inits as in Pharmpy/pyDarwin/PsN; bootstrap's own `update_inits: true` default; a `.fitrx` refit of the winner) is rejected by `Strictness::default()` with a `#751` reason that is false. The optimizer's own `left_init` (`outer_optimizer.rs:2078`) is computed and discarded.

4. **`src/model_selection.rs:484` — `stalled_at_init` returns `Some(true)` for a model with zero free parameters.** Every loop skips on `is_fixed` and falls through to `Some(true)`, contradicting its own doc ("a fit with no free parameter at all cannot stall and reports `Some(false)`"). No test covers the all-fixed case.

5. **`src/model_selection.rs:425` — the `INIT_STALL_REL_TOL` doc claim is false.** "A fit the optimizer flagged as stalled is flagged here too, never the reverse" — but the optimizer's `INIT_ESCAPE_STEP_S` test runs in scaled packed space (packed θ = ln θ, scale = |ln θ₀| when > 0.1), so its 1 % is ≈ |ln θ₀| % on the natural scale. V₀ = 100 → 103: scaled deviation 0.0064 < 0.01 → optimizer says still at init; `stalled_at_init`: 3 > 1 → moved. The two predicates disagree on exactly the borderline fits the gate exists for.

6. **`src/model_selection.rs:339` — correlation gate labels packed coordinates with `theta_names`, and tests block-Ω on the Cholesky scale.** `covariance_matrix` is the full packed n×n (θ, Ω-Cholesky, σ, κ…), so any index ≥ n_theta prints "packed coordinate k". And because the matrix is on the packed scale, the |r| tested for `block_omega` pairs is not the natural-space ω correlation pyDarwin/Pharmpy read — the codebase already restricts `high_correlation_pairs` to θ for this reason (`postfit.rs:1573`). `packed_param_names` (`io/output.rs:1738`) already produces the right labels.

7. **`src/model_selection.rs:313` — `require_covariance` accepts `SirFallback` unconditionally.** SIR reports `Ok` with a single finite weight (`sir.rs:690-696`); ESS is only `eprintln!`ed under verbose, no floor. An ESS-collapsed fallback with zero-width CIs passes as "delivered uncertainty", and `max_condition_number` / `max_correlation` then `skip` — the candidate is ranked with no working uncertainty gate at all. `check_strictness` never reads `sir_ess`.

8. **`crates/ferx-tools/src/bootstrap/mod.rs:783` — only `near_boundary` was delegated to core.** `cov_ok = covariance_matrix.is_some() && !cov_failed` now disagrees with `Strictness::require_covariance` on every SIR-fallback replicate (matrix is `None` by construction; the non-PD warning classifies as `CovarianceFailed`). Such replicates are reachable because `replicate_options` clones `covariance_fallback` from the model file while forcing `sir = false` — contradicting the "no SIR" comment at 706-707.

9. **`src/model_selection.rs:115` — `n_parameters == 0` is inconsistent across `BicType` arms.** For an old all-zero bundle the missing-tally guard (`0 != 0`) passes, then `Fixed` returns NaN (`n_obs == 0`) while `Mixed`/`Random`/`Iiv` return `ofv`, breaking the documented `Fixed == FitResult::bic` identity. Low severity; fresh fits are consistent.

### Tests that cannot fail (confirmed by mutation)

10. **`src/io/fitrx.rs:632` and `src/api/fit.rs:1033` — neither advertised behaviour is tested.** Mutating `bic_inputs: r.bic_inputs` to `Default::default()` *and* reverting `fit.rs:1033` to `packed_fixed_mask` leaves `cargo test --lib --features ci` at 3929 passed / 0 failed. `minimal_fit_result()` carries `BicInputs::default()`, indistinguishable from a dropped field; `roundtrip_minimal_fit` never asserts `bic_inputs`; `tally_counts_block_omega_elements_and_iov_kappas` never calls `fit()`; no test anywhere asserts `FitResult::n_parameters`. CLAUDE.md: "Bug fixes should add a regression test that fails without the fix … for a regression test that means mutation."

### Refuted
Sibling-branch `cond_ctx` handling; `omega_init` fitrx fallback; release-mode silent mis-tally.

### Confirmed but cut by the 10-item cap (cleanup)
`Option<BicInputs>` instead of an all-zero sentinel; a third cov→corr loop duplicating existing helpers.

roninsightrx:
All ten findings above are addressed in 2bca765e.

| # | Fix |
|---|---|
| 1 | `NegSum(a, b)` read as half-open `θ[a..b]`; unit test builds a `ThetaGather` with a trailing θ (fails under `..=`). |
| 2 | `[event_model]` / `[binary_model]` / `[markov_model]` expressions feed `classify_theta_eta_linked` as synthetic assignments. Tests: the compact `tte_weibull` spelling, and the legacy dummy-PK spelling with the η line in either block classifying identically. Form-C readouts / `[odes]` hazard lines are still not seen — documented in `docs/api/fitting.qmd`. |
| 3 | Cold-start assumption documented on `Strictness::reject_init_stall` and in the docs table; warm-started searches are told to turn the gate off. |
| 4 | `stalled_at_init` tracks `saw_free` and returns `Some(false)` for an all-FIX fit (checked before the optimizer verdict, since that reads "did not move" on an empty vector). |
| 5 | `FitResult::left_init` (from `OuterResult`, `.fitrx` round-trip, `#[serde(default)]`) carries the optimizer's own `INIT_ESCAPE_STEP_S` verdict; `stalled_at_init` prefers it, the natural-scale 1 % is now documented as the fallback for results without it. `Some` only for the NLopt outer loop. |
| 6 | Labels via `packed_param_names` (now `pub(crate)`, split out `packed_layout`). New `natural_scale_covariance` applies the delta method on block-Ω / block-κ segments using `parameterization::omega_cholesky_jacobian` — factored out of `extract_standard_errors`, so `se_omega` and the gate share one Jacobian (FD-checked). Test fixture straddles 0.95: packed |r| = 0.99, natural 0.07. |
| 7 | `SirFallback` is accepted only with `sir_ess` finite and ≥ `MIN_SIR_FALLBACK_ESS` (2.0); missing / NaN / collapsed each name their reason. |
| 8 | `replicate_options` also sets `covariance_fallback = None`; `cov_ok` is `check_strictness` with `require_covariance` only. |
| 9 | `bic()` uses one `term(count, ln)` helper in all four arms: zero count → zero penalty, so `Fixed == FitResult::bic` on an all-FIX bundle. |
| 10 | `bic_inputs_and_left_init_round_trip_through_fitrx` (non-default tally), `fit_wire_missing_bic_inputs_and_left_init_default` (→ NaN), and Tier-2 `tests/n_parameters_free_packed_dim.rs` (`fit()` on block + diagonal Ω, `n_parameters == free_packed_dim() == 8`). Mutation-checked: `Default::default()` in the wire, `packed_fixed_mask` at the call site (10 vs 8), and `..=` in the classifier each turn a test red. |

Gate: `tools/preflight.sh` OK (fmt, check matrix, clippy, rustdoc, docs, public-API baseline updated for `FitResult::left_init`, `OuterResult::left_init`, `natural_scale_covariance`, `MIN_SIR_FALLBACK_ESS`); `cargo test --lib --features ci,survival` 4076 passed; ferx-tools suite green.

Follow-up needed in ferx-r: it builds `FitResult` literals in three places (`src/rust/src/lib.rs`), which need `left_init: None`, plus the usual `Cargo.lock` bump via `tools/update-ferx-core-lock.sh` once this merges.

roninsightrx (commented):
[P1] src/parser/model_parser.rs:1096 — Propagate NnOutput dependencies into theta_eta_linked.

For CL = TYPICAL_PK.CL * exp(ETA_CL), this walker sees the ETA but NnOutput contributes no theta indices. The NN weights are appended to ModelParameters.theta at CovariateNn.weights_offset, so all of them remain theta_fixed-class and Mixed BIC penalizes them on ln(n_obs). The existing nn_info_tests fixture exercises exactly this shape. For a multi-output network, shared hidden weights and the selected ETA-bearing output head need to become random-class while unrelated output-head weights remain fixed-class. Please add an nn-feature regression.

[P1] src/model_selection.rs:570 — Do not infer Ω/κ layout from the total covariance dimension.

The covariance dimension also includes mixture override coordinates and trailing block_sigma rho coordinates, but packed_layout subtracts only theta and sigma. For example, two diagonal ETAs plus one free residual correlation leaves three coordinates after theta/sigma, so packed_layout selects full Ω; the 3x3 Cholesky Jacobian is then embedded across the two actual Ω coordinates and the first following coordinate, corrupting both the transformed covariance and the strictness correlation verdict. The equal-dimension diagonal-Ω/block-κ case is independently ambiguous and already documented in packed_layout. Since this now controls candidate eligibility, the structure needs to be persisted on FitResult or otherwise recovered unambiguously. Please cover diagonal Ω + block_sigma and diagonal Ω + same-size block κ.

[P2] src/model_selection.rs:529 — Honor left_init when only non-base packed coordinates are free.

saw_free inventories theta, sigma, and base omega only. A valid model can fix all of those while estimating kappa, a mixture override, or a residual correlation. For a live NLopt result, left_init = Some(false) then reliably says that no packed coordinate escaped, but this early return converts it to Some(false) from stalled_at_init and the candidate passes the default gate. The no-free shortcut must account for every packed segment, or defer to left_init whenever n_parameters is nonzero. A regression with all base masks fixed, a free kappa, and left_init = Some(false) demonstrates it.

Process note: this adds public FitResult fields and the sibling ferx-r checkout still has three FitResult literals that need them. The PR template says a required ferx-r follow-up must be open and linked, with this PR draft until then.

Verification on the pre-merge semantic head 2bca765e: cargo test --lib --features ci,survival passed (4,076 passed, 23 ignored); ferx-tools bootstrap tests passed (102); n_parameters_free_packed_dim passed. Head 679f80c9 is a merge-only refresh of main, and CI is now running.

roninsightrx:
Codex review findings addressed in 0d02aed2.

**[P1] NnOutput → `theta_eta_linked`.** `classify_theta_eta_linked` now takes per-network, per-head θ dependency lists built from the new `MlpMapper::weight_indices_for_output(k)` (every hidden-layer weight + head `k`'s row and bias), offset by `CovariateNn::weights_offset`. On the `nn_info_tests` fixture (`CL = TYPICAL_PK.CL * exp(ETA_CL)`, `V = TYPICAL_PK.V`, 2→3→2): the 9 shared weights and the CL head's 3 + 1 are random-class, the V head's 3 + 1 and `TVKA` fixed-class. nn-gated regression `covariate_nn_weights_follow_the_eta_bearing_output_head` (asserts both the index vector and the `W_TYPICAL_PK_2_1_*` / `B_TYPICAL_PK_2_1` names), plus a unit test on the index split; mutation (ignore `NnOutput`) turns it red.

**[P1] Ω/κ layout inferred from the dimension.** Gone. `FitResult` now records `omega_is_diagonal` / `kappa_is_diagonal` (set from `ModelParameters` in `fit()`, serde default, `.fitrx` round-trip). `natural_scale_covariance` reads only those: a result without them (pre-#1177 bundle, hand-built) is returned as stored — never transformed on a guess. `packed_layout` prefers the recorded flags and keeps the size heuristic only for the YAML labels on old bundles, with its doc now saying so. Tests: diagonal Ω + a trailing ρ coordinate (matrix untouched, |r| read as stored), diagonal Ω + same-size block κ (κ segment transformed, θ/Ω/σ untouched, the packed/natural straddle asserted on κ, and the gate names `KAPPA_V` when the κ layout is unrecorded), unrecorded Ω layout.

**[P2] `left_init` with only non-base coordinates free.** `stalled_at_init` now starts from `n_parameters` (every packed segment): `0` → `Some(false)`; otherwise `left_init` is honoured before the θ/Ω/σ walk; and free coordinates only in segments the result has no initials for, with no optimizer verdict, yield `None` (the gate reports `skipped`) rather than a pass. Regression: all base masks fixed, `n_parameters = 1`, `left_init = Some(false)` → stalled and rejected; `Some(true)` → passes; `None` → skipped.

**Process note.** ferx-r follow-up is open and linked: FeRx-NLME/ferx-r#327 (three `FitResult` literals need `left_init` / `omega_is_diagonal` / `kappa_is_diagonal`, then the lock bump). The PR description's cross-repo table now names it as the required follow-up.

Gate: preflight OK (public-API baseline regenerated), `cargo test --lib --features ci,survival` 4165 passed, ferx-tools green, Tier-2 `n_parameters_free_packed_dim` passes.

Issue #898, opened 2026-07-20
predict()/simulate()/suggest_start() panic on precondition failures instead of returning Result — the message is lost at the R boundary


TeunP:
## Problem

`fit()` surfaces every model/data precondition as a recoverable `Err`, via `first_error(&check_model_data_rule(..))?` (`src/api/fit.rs:357`). `predict()` / `simulate()` / `suggest_start()` surface the **same** conditions as a `panic!`, through a family of `assert_*` wrappers over the identical `check_*` predicates:

| assert | src/api/validation.rs |
|---|---|
| `assert_dose_compartments_supported` | `:815` |
| `assert_modeled_doses_supported` | `:838` |
| `assert_survival_tv_covariates` | `:1321` |
| `assert_absorption_closed_form_support` | `:1364` |
| `assert_absorption_flip_flop_no_twin` | `:1437` |
| `assert_analytic_readout_support` | `:1451` |
| `assert_absorption_dosing_supported` | `:1469` |

All funnel through `panic_if_unsupported` (`src/api/validation.rs`), whose message explicitly says *"fit() reports this as an error rather than panicking"* — i.e. the asymmetry is deliberate and documented, not accidental.

## Why it's worth revisiting

**The actionable message is destroyed at the R boundary.** ferx-r runs the fit on a worker thread and joins it (`ferx-r/src/rust/src/lib.rs:352`):

```rust
Ok(Err(e)) => throw_r_error(format!("Fit error: {e}")),
Err(_)     => throw_r_error("Fit error: worker thread panicked"),
```

So an `Err` reaches the R user with its full diagnostic ("subject 1, time 0: infusion into compartment 0, but the analytical `two_cpt_oral` model can only infuse into compartment(s) [1, 2, 3]…"), while a panic reaches them as **`"worker thread panicked"`** — every bit of context gone. The care put into these messages is wasted on exactly the surface where users most need them.

That is the substance of the complaint in #375 ("a library should surface this as a propagated `Err` the caller (CLI / ferx-r) can report — not a panic that crashes the R session"). #375 addressed the *specific* validator; the general contract is unchanged.

[...]

Issue #730, opened 2026-07-07
Map/rename dataset columns to canonical roles in [data] block (NONMEM $INPUT TIME=TAFD analogue)


roninsightrx:
## Summary

Add a way to map (rename) dataset columns to ferx's canonical roles inside the model file, analogous to NONMEM's `$INPUT TIME=TAFD`. Today ferx-core requires the CSV to already carry canonical column names (`ID`, `TIME`, `DV`, `AMT`, `EVID`, `CMT`, `RATE`, `MDV`, `II`, `SS`, …). If a dataset's time column is named `TAFD` (or `DV` is `CONC`, etc.), the user must pre-edit the CSV. NONMEM lets you remap this in-model.

## Current behaviour

- Column resolution is a hardcoded, **case-insensitive** lookup against the header — `col_idx_ci` in `src/io/datareader.rs:393-417`. Case is the only flexibility (`Time` and `TIME` both work); there is no rename.
- The `[data]` block (`src/parser/model_parser.rs:4620`, the `$DATA` analogue from #690) accepts **exactly one key, `path`**; any other key errors with `[data]: unknown key ...`.
- The only user-configurable column is `iov_column` in `[fit_options]` (names which column holds the IOV occasion index) — not a general rename.
- `TAFD` currently exists only as a *derived output* (time-after-first-dose, `src/api.rs:2658`), not an input role — so a dataset column literally named `TAFD` is silently treated as a covariate.

## Proposed behaviour

Extend the `[data]` block to accept column-mapping entries, mapping a **canonical role** to the **actual dataset header**:

```
[data]
path = data/mydata.csv
TIME = TAFD
DV   = CONC
AMT  = DOSE
```

Semantics:
- Key = canonical ferx role (`ID`/`TIME`/`DV`/`AMT`/`EVID`/`CMT`/`RATE`/`MDV`/`II`/`SS`/`CENS`/`ADDL`/…). Value = header name in the CSV.
- Direction matches NONMEM (`canonical = actual`, i.e. `TIME=TAFD` means "use TAFD as TIME"). Worth confirming, since NONMEM's `$INPUT` is positional; we should pick the least surprising direction and document it clearly.
- Case-insensitive matching on the value preserved.
- A mapped role takes precedence over an auto-detected column of the same canonical name.
- Once a header is claimed as a role via mapping, it must not also be picked up as a covariate.
- Error if the mapped source header is absent, or if two roles map to the same header.

## Implementation sketch

[...]

PR #1225, opened 2026-09-04
fix(api): reject a population read without the model's endpoint routing [closes #1199]

### LINKED ISSUE #1199: bug(api): `read_nonmem_csv` builds no TTE records, so a joint PK-TTE fit silently loses its whole hazard half
## Problem

`read_nonmem_csv` takes no model, so the reader's non-Gaussian routing sets are empty
(`io/datareader.rs`, `routing.tte.contains(&cmt)`); they are built from the model in
`api/run.rs:1103` via `ObsRouting::tte_and_discrete`. A population loaded that way carries **zero**
`ObsRecord::Event` entries, and `fit()` then evaluates a joint PK-TTE model's Gaussian half only.

Same model, same CSV, at `1a253c1e`, `maxiter = 0`:

| loader | `obs_records` per subject | `n_obs` | OFV |
|---|---|---|---|
| `read_nonmem_csv(path, None, None)` | 0 | 360 | **2774.22** |
| `read_population_for(&model, …)` | 1 | 320 | **95.79** |

The routed number is the right one: NONMEM on the same data and parameters gives
`#OBJV = 95.926311`.

## Nothing says so

`fit()` returns `Ok`. The `warnings` vector names only downstream symptoms — EPS shrinkage
−194.5 %, RSE 72969 % on TVCL, a correlated pair — none of which points at a missing endpoint.
`api/validation.rs` has no guard for "the model declares a survival endpoint and the population
has no event records".

Both functions are in the public API baseline, and the only thing separating them is a doc
sentence on `read_population_for` ("Library consumers … should call this instead of the individual
`read_nonmem_csv*` functions so that TTE routing is applied"). A doc sentence is not a guard: I
walked into this myself while rebuilding a #1166 benchmark fixture, and it produced a plausible,
finite, entirely wrong OFV.

## Suggested direction

A `fit()`-level check: the model declares a TTE / discrete-state / count endpoint **and** the
population carries no matching `ObsRecord` ⇒ hard error naming the endpoint and the loader. It
cannot be a warning that only appears in the result — the symptom here is a fit that converges and
reports.

Sibling of #832 (`fit_from_files` silently drops `iov_column`): same shape, a loader entry point
that quietly does less than the model asks for.

## Provenance

Found while running item 0 of the #1166 plan; the details of the affected measurement are on
#1167.

TeunP:
## Why

`read_nonmem_csv` takes no model, so a joint PK-TTE (or binary / Markov) dataset read through it carries the endpoint's rows as **Gaussian observations on the endpoint CMT** and no `obs_records`. `fit()` then ran the Gaussian half only and returned `Ok` with a plausible, finite, wrong objective, `predict()` returned a concentration for the event row, and `simulate()` drew no events at all (its TTE draw is keyed on `obs_records`). Nothing said so — the `warnings` vector carried only downstream symptoms (EPS shrinkage −277 %, RSE in the tens of thousands). Measured at `main` `ebc13a1e` on `nonmem_anchor/pktte_tdep` (`maxiter = 0`):

| load | `(events, Gaussian rows, Gaussian rows on CMT 3)` | OFV |
|---|---|---|
| `read_nonmem_csv` | `(0, 360, 40)` | **4606.6042** |
| `read_population_for` | `(40, 320, 0)` | **1912.0247** (NONMEM `#OBJV = 1912.025`) |
| `read_population_for`, `CMT` column absent | `(0, 360, 0)`, no reader warning | **4606.6042** |

Four in-crate callers used the model-blind reader with a model in hand: the `.fitrx` reload (`load_fit`, consumed by `ferx` CLI), FREM prep, and — found by the review of this PR — the `population = None` re-read in `run_covariance` / `run_sir`, which is exactly the path the R wrapper's `ferx_covariance()` / `ferx_sir()` take. Plan and two scrutiny passes are on #1199.

## What changed

- **Guard — `check_endpoint_routing` in `api/validation.rs`**, two error codes:
  - `E_ENDPOINT_UNROUTED`: a subject has a Gaussian observation on a CMT the model declares as a non-Gaussian endpoint. The routed reader never puts an endpoint CMT into `obs_cmts` (pinned by two new datareader tests, TTE and discrete routing sets), so this is the zero-false-positive signature of "this population was not built for this model". Wired into `fit()` / `ferx check` (`check_model_data_rule`, **before** the per-CMT checks so `E_PER_CMT_ERROR_MODEL` cannot rename the cause), `simulate()` (`check_simulation_data`), `predict()` and `predict_categorical()` (their `assert_*` panic convention, #898), and `run_covariance()` / `run_sir()` (`Result`, next to their existing `check_dose_compartments`).
  - `E_ENDPOINT_NO_RECORDS`: a declared endpoint has no routed row anywhere — the `CMT`-column-absent case, which the first code cannot see. **Fit / `ferx check` / `run_covariance` / `run_sir` only**: a simulation template has no event rows by construction and `predict()` legitimately takes a PK-only population.
  - Messages carry the code, the CMT, the endpoint kind, the first offending subject, and `read_population_for`; the diagnostic's `block` is the endpoint's own block (`event_model` / `binary_model` / `markov_model`), so `ferx check` attaches the right line.
- **One routed re-read for every caller that holds a `CompiledModel` but no `ParsedModel`** — `pub(crate) read_population_routed_by(model, path, iov_column, column_map)` in `api/run.rs`, built on `obs_routing_for(model, missing_dv)` (extracted verbatim from `read_population_for_policy`; the single place routing is derived from a model). Used by `load_fit`, `run_covariance` and `run_sir`; `run_sir` now also honours `[data]` column renames on the re-read (#730), which only `run_covariance` did before.
- **FREM prep refuses non-Gaussian endpoints** (`E_FREM_NON_GAUSSIAN_ENDPOINT`, at both `prepare_frem` and `transform_dataset_for_frem`, before covariate resolution). The FREM dataset writer walks `obs_times` only, so *routing* that read would have dropped the event rows from the file silently; the unrouted read kept them by accident while mis-stating the covariate statistics. Decided with @TeunP on the issue: reject rather than build FREM-on-joint.
- Not auto-repaired: re-classifying Gaussian rows into events after the fact would lose `TENTRY` and the DV-code validation the routed reader performs — a second silent path.
- Small: `ObsRecord::cmt()` accessor (`pub(crate)`), the two codes added to the `src/diagnostics.rs` registry table.

## Alternatives considered

Option (i) for FREM — route the read and teach the writer to emit `Event` records (interval-censored needs the two-row DV=2 coding, `TENTRY` carried through, and an open question of what FREM pseudo-observations mean next to a TTE term). Feature work with no anchor; declined on the issue.

## Cross-repo dependency
| ferx (R) | — | not needed: every R `fit` / `simulate` / `predict` entry point already goes through `read_population_for{,_simulation}`; `ferx_covariance()` / `ferx_sir()` call `run_covariance(&fit, None, None, ..)` / `run_sir(..)` and pick up the routed re-read from this PR once the lock is bumped | — |

## Breaking changes
- [x] None — no public signature changes; the new items are `pub(crate)`. Behaviour change: a population that "worked" by accident (unrouted joint fit; FREM prep on a joint model; SIR / covariance re-read of a joint fit) is now a named error or, for the re-read, correct — CHANGELOG *Fixed* / *Changed*.

## Numerical validation
- [x] Compared against: NONMEM — the guard moves no number; the anchor is that the routed OFV stays on `pktte_tdep`'s `#OBJV = 1912.025` (`tests/pktte_tdep_nonmem_anchor.rs`, runs on PRs), and the unrouted 4606.6 can no longer be produced.

## Performance
- [x] No significant impact expected — one linear pass over `obs_cmts` and one over `obs_records` at entry; immediate return for a model with no endpoints.

## Tests
- [x] Unit tests added: `src/api/tests/endpoint_routing_tests.rs` (15: the guard at every entry point, both codes, TTE and binary fixtures, the shared re-read, `run_covariance` / `run_sir` on a supplied unrouted population and on the `None` re-read path, `predict_categorical`), `datareader_tests::routed_reader_never_places_{an,a_discrete}_endpoint_cmt_in_the_gaussian_grid`, `fitrx::roundtrip_bundled_data_routes_{the_model_endpoints,a_binary_endpoint}`, `frem::a_non_gaussian_endpoint_is_rejected_by_name` (TTE + binary). All on real `examples/` + `data/` file pairs read both ways. The whole `endpoint_routing_tests` module runs in 0.3 s (the `maxiter = 0` fits included).
- [x] `tests/dose_compartment_routing.rs`: the six #905/#924 tests that loaded a pure-TTE subject model-blind now pin the refusal (`E_ENDPOINT_UNROUTED`) on that load and exercise the dose-compartment gates through the routed loader; a new `pk_value_path_still_declines_a_model_blind_tte_subject_with_nan` keeps the #905 decline itself pinned at the public `pk::compute_predictions_with_tv` / `pk::predict_iov` value path, which the entry-point gate does not cover.
- [x] Regression tests fail without the fix — **mutations observed, then reverted**: disabling the `obs_cmts` scan turns the unrouted arms red (fit / binary / predict / simulate) and nothing else; disabling the zero-records scan turns the three no-rows arms red (fit / binary / `ferx check`) and nothing else; re-reading through `read_nonmem_csv_mapped` in `run_covariance` / `run_sir` turns `run_covariance_and_run_sir_reread_the_dataset_routed` red with `E_ENDPOINT_UNROUTED`.
- Verification table from the issue, every row hit: unrouted joint → `Err` with `E_ENDPOINT_UNROUTED`, `CMT 3`, first error; routed → silent, fits; `CMT` dropped → `E_ENDPOINT_NO_RECORDS` naming the column, unrouted code silent; binary endpoint → same two codes; `predict()` unrouted → panic, routed → PK rows only; `simulate()` unrouted → `Err`, event-free routed template → silent; `ferx check` → reports the dropped column; `load_fit` on the joint bundle → event count equals the routed reader's (was 0); FREM → named `Err` at both entry points.

- [x] Not applicable — no new API surface.

## Docs
- [x] `docs/api/parsing.qmd` (new `read_population_for()` section + model-blind callout on `read_nonmem_csv()`; the function lives at `ferx_core::api::read_population_for`), `fitting.qmd` / `index.qmd` / `simulation.qmd` examples switched to `read_population_for`, `file-formats/check-report.qmd` code table (both codes), `estimation/tte.qmd` "Loading the data", `estimation/frem.qmd` limitation made precise; reader doc comments on all four `read_nonmem_csv*` functions.
- [x] `CHANGELOG.md` `[Unreleased]` — *Fixed* (guard, `load_fit`, `run_covariance` / `run_sir`) and *Changed* (FREM).

## Checklist
- [x] `cargo test --features ci,survival --lib`, `--tests` with `ci,survival`, and `cargo check --tests` with `ci` alone and `ci,markov` (guard compiles to a no-op without `survival`; the `Ctmm` arm only with `markov`)
- [x] Not on the `Dual2` path

## Docs & examples
- [x] No example execution step needed — no new example or DSL; existing examples unchanged.

## Review

`/code-review xhigh` on the branch (10 finder angles, one verifier per candidate, sweep). Resolved in this PR: `run_covariance` / `run_sir` unrouted re-read and missing guard (also reachable from ferx-r); `predict_categorical` missing the guard; `.with_block("event_model")` hard-coded for binary / Markov endpoints; the `kind_of` wildcard lumping a CTMM endpoint as "non-Gaussian"; FREM message naming only two of the three blocks; codes missing from the `diagnostics.rs` registry; binary / discrete arms untested in the fitrx, FREM and datareader tests; the #905 NaN-decline branch left unpinned by the test rewrite; a brittle `"subject 1"` substring; the O(endpoints × records) count loop; the double `parsed.as_ref()` in `load_fit`. Kept as is: the messages carry their code as a prefix because `first_error` returns the message only (the other prefix precedent is `E_DOSE_NO_AMT` in the reader); the `obs_routing_for` cfg-twin layout is a verbatim move.

## Reviewer hints

`api/validation.rs` `check_endpoint_routing` is the whole fix; everything else is wiring. The one non-obvious decision is FREM (reject, not route) — the writer loop in `frem/mod.rs` (`for (j, (&time, &dv)) in subject.obs_times…`) is why. The `run.rs` diff is a pure extraction plus the shared re-read helper.

## Open questions

None.

Closes #1199.


TeunP:
## Code review (`/code-review xhigh`) — findings and resolution

Ten independent finder angles over the diff (line-by-line, removed behaviour, cross-file callers, Rust pitfalls, wrapper/extraction faithfulness, reuse, simplification, efficiency, altitude, CLAUDE.md conventions), one verifier per de-duplicated candidate, then a fresh sweep over the tree after the fixes. Fifteen findings survived verification; fourteen are fixed in this PR, one is kept deliberately. Every fix that adds a guard or a test was mutation-checked: the guard removed, the new test goes red.

| # | Finding | Verdict | Resolution |
|---|---|---|---|
| 1 | `run_covariance` / `run_sir` re-read `fit.data_path` through the model-blind reader when `population = None` and ran no endpoint-routing check — the same #1199 hazard on two more public entry points, and exactly the path ferx-r's `ferx_covariance()` / `ferx_sir()` take (verified on ferx-r `origin/main`). | confirmed | Both now re-read through a shared `read_population_routed_by(model, path, iov_column, column_map)` and run `check_endpoint_routing(.., true)` next to their existing `check_dose_compartments`. Tests: supplied unrouted population → `E_ENDPOINT_UNROUTED`; the `None` re-read path checked by what fails *next* (missing covariance matrix / mis-sized omega), and observed to flip to `E_ENDPOINT_UNROUTED` when the re-read is mutated back to `read_nonmem_csv_mapped`. |
| 2 | `predict_categorical` had no guard, but `predict_binary` walks `obs_records`, so an unrouted binary population returned an empty vector — indistinguishable from a model without a binary endpoint. | confirmed | `assert_endpoint_routing` added; should-panic + one-row-per-record tests, mutation-verified. |
| 3 | Both diagnostics hard-coded `.with_block("event_model")`; `validate_model_file` resolves the line from `block_lines`, so a `[binary_model]`-only model got no line and `ferx check` pointed at a block the file does not have. | confirmed | Block chosen per endpoint kind (`event_model` / `binary_model` / `markov_model`). |
| 4 | `kind_of`'s wildcard arm labelled a CTMM endpoint "non-Gaussian"; the `#[allow(unreachable_patterns)]` was inert. | plausible | Explicit `Ctmm` arm under `markov`, `Gaussian => None` (filtered out of the endpoint set), no wildcard; compiles under `ci`, `ci,survival`, `ci,markov`. |
| 5 | FREM rejection message named `[event_model]` / `[binary_model]` but `has_non_gaussian()` also fires for a CTMM endpoint. | confirmed | Message names all three blocks; FREM test gained the binary fixture. |
| 6 | Rewriting the six #905/#924 model-blind tests to assert the new refusal left the public `pk::compute_predictions_with_tv` / `pk::predict_iov` decline-to-NaN branch untested with a live row. | confirmed | `pk_value_path_still_declines_a_model_blind_tte_subject_with_nan` pins it at the value path (Tier 2). |
| 7 | The fitrx round-trip, FREM and datareader-invariant tests covered only the TTE routing set; the discrete (binary / CTMM) arm had no test. | confirmed | Binary round-trip, discrete-set invariant, and FREM binary arm added on `examples/binary_logistic.ferx`. |
| 8 | `run_sir` never captured `[data]` column renames on its re-parse (#730), unlike `run_covariance`. | confirmed | Captured and threaded into the shared re-read. |
| 9 | New codes missing from the `src/diagnostics.rs` registry table. | confirmed | Added. (Seven older codes were already missing there — pre-existing, left alone.) |
| 10 | Docs "raised by" lists were stale once the guard reached `predict_categorical` / `run_covariance` / `run_sir`. | confirmed (sweep) | `check-report.qmd` and `api/parsing.qmd` updated. |
| 11 | `check_model_data` doc listed the check order without the new guard. | confirmed (sweep) | Updated. |
| 12 | `msg.contains("subject 1")` also matches `subject 10`. | confirmed | Asserts `"subject 1)"`. |
| 13 | Zero-records scan was O(endpoints × records); the `record_cmt` closure re-derived an accessor `ObsRecord` lacks. | confirmed | Single pass over the records; `ObsRecord::cmt()` (`pub(crate)`). |
| 14 | `load_fit` derived `column_map` and then matched the parse result a second time. | confirmed | Single `match`; the `Ok` arm uses the shared re-read. |
| 15 | The two messages embed their code as a prefix, which no other `Diagnostic` does. | confirmed | **Kept.** `first_error` returns the message only, so the prefix is what carries the code into `fit()`'s `Err`; changing `first_error` would alter every existing error string. The one existing prefix precedent is the reader's `E_DOSE_NO_AMT`. |

Refuted / no change: the `maxiter = 0` fit inside a Tier-1 test (the whole module runs in 0.3 s); `predict_survival` (reads no observation rows, no guard needed); every ferx-r `fit` / `simulate` / `predict` entry point already routes through `read_population_for{,_simulation}`; the `obs_routing_for` cfg-twin layout is a verbatim move; no `E_ENDPOINT_NO_RECORDS` false-positive surface found (every TTE convention writes one row per declared endpoint per subject, so a subset or resample with at least one subject keeps a routed row per endpoint).

PR #1228, opened 2026-09-04
test(per-route-lag): FIX the zero-variance omega so the Tier-3 fit is identifiable [closes #1227]

### LINKED ISSUE #1227: test(per-route-lag): per_route_lag_fit_recovers_lag fits a free omega on one subject, so ETA_CL runs to the +6 rail and #1205 correctly demotes converged
`tests/per_route_lag.rs::per_route_lag_fit_recovers_lag` has failed on every `Slow tests` run
since #1205 landed, and the fixture — not #1205 — is what is wrong.

## What fails

```
thread 'per_route_lag_fit_recovers_lag' panicked at tests/per_route_lag.rs:626:5:
per-route-lag fit should converge
```

`assert!(fitted.converged)`. Reproduced locally on `19fdbb04`. Everything else in the nightly
suite is green: 5477 passed, 1 failed, one binary.

## When it started

| run | head | verdict |
|---|---|---|
| [33657468069](https://github.com/FeRx-NLME/ferx-core/actions/runs/33657468069) | `5f7ef926` (PR #1200) | last success |
| [33672091642](https://github.com/FeRx-NLME/ferx-core/actions/runs/33672091642) | `6bb6ebd5` (PR #1205, `fix/1118-demote-converged-runaway-guard`) | first failure |

Same test, same line, every run since.

## Why

#1205 (#1118) made `runaway_guard_warning` demote `converged` on a runaway-verdict guard hit.
The optimizer's own stop rule did not change — only the boolean handed to callers did.

The fit has always ended pinned to a rail. Printing the post-fit warnings:

```
Internal optimizer parameter guard reached by estimate(s):
  ETA_CL (estimate 162754.7914; packed coordinate 6.0000 at upper guard, runaway)
severity = Critical, verdict = runaway
```

`ETA_CL` is the Omega Cholesky diagonal at the `+6` ceiling from `compute_bounds`
(`src/estimation/parameterization.rs:555`) — `exp(6) ≈ 403`, so ω² ≈ 162 755.

`ROUTE_LAG_SINGLE` declares `omega ETA_CL ~ 0.0` **free**, and the test fits **one subject** on
noise-free data. A single subject carries no between-subject information, so ω² is
unidentifiable by construction: nothing bounds it from above and it walks to the rail. The
fit's own diagnostics say the same thing — `TVV ~ TVKA = -1.00`, `TVV ~ TVLAG = -1.00`,
`TVKA ~ TVLAG = 1.00`, RSE up to 11 650%, 3 of 6 Hessian eigenvalues clipped.

The rest of the fit is fine, and the thing the test exists to check is correct:

```
theta = [2.5889, 43.4805, 0.7092, 2.0000000137]     # TVLAG = 2.0000000137, truth 2.0
sigma = [0.15554]
ofv   = -323.2803
```

The second assertion, `(theta[3] - 2.0).abs() < 0.1`, passes to eight decimals. Only the
boolean fails.

So this is **not** #1205 over-firing. Before #1205 the test was green on a fit whose ω had run
to an implementation rail — it passed for the wrong reason, and the new gate is reporting that
honestly.

## Fix

Declare the omega FIX in the fixtures: `omega ETA_CL ~ 0.0 FIX`. The model declares zero IIV
and fits one subject, so there is nothing to estimate; `packed_fixed_mask` then excludes the
coordinate, no guard hit, no demotion. `FIX` is a no-op for the predict-only twins in the same
file, so the "compared at identical parameters" claim between a route-lag model and its
compartment-lag twin is unchanged.

The alternative — teaching the assertion to accept a runaway ω — would keep a degenerate
fixture and re-hide exactly what #1205 was built to surface.

TeunP:
## Why

`tests/per_route_lag.rs::per_route_lag_fit_recovers_lag` has failed on **every** `Slow tests`
run since #1205 landed — the nightly suite has been red on this one test for two days, with
everything else green (5477 passed, 1 failed, one binary).

```
thread 'per_route_lag_fit_recovers_lag' panicked at tests/per_route_lag.rs:626:5:
per-route-lag fit should converge
```

Bisected on run history:

| run | head | verdict |
|---|---|---|
| [33657468069](https://github.com/FeRx-NLME/ferx-core/actions/runs/33657468069) | `5f7ef926` (#1200) | last success |
| [33672091642](https://github.com/FeRx-NLME/ferx-core/actions/runs/33672091642) | `6bb6ebd5` (#1205, `fix/1118-demote-converged-runaway-guard`) | first failure |

**The fixture is what is wrong, not #1205.** `ROUTE_LAG_SINGLE` declared
`omega ETA_CL ~ 0.0` **free**, and the test fits **one** subject. A single subject carries no
between-subject information, so ω² is unidentifiable by construction: nothing bounds it above
and it walks to `compute_bounds`' `+6` Cholesky-diagonal ceiling
(`src/estimation/parameterization.rs:555`, `exp(6) ≈ 403`, so ω² ≈ 162 755). Printing the
post-fit warnings:

```
Internal optimizer parameter guard reached by estimate(s):
  ETA_CL (estimate 162754.7914; packed coordinate 6.0000 at upper guard, runaway)
severity = Critical, verdict = runaway
```

#1205 (#1118) made that rail demote `converged`. The optimizer's stop rule did not change —
only the boolean did. Before #1205 this test was green on a fit whose ω had run to an
implementation rail; it passed for the wrong reason, and the new gate is reporting that
honestly.

## What changed

`omega ETA_CL ~ 0.0 FIX` in the file's model fixtures. That matches what the model already
says — zero IIV — and `packed_fixed_mask` then drops the coordinate from the packed vector, so
there is no guard hit and no demotion.

**The runaway ω was not cosmetic — it was absorbing the structural signal.** Same noise-free
data, same perturbed start, measured before and after:

| | free ω | ω `FIX`'d | truth |
|---|---|---|---|
| `converged` | false | **true** | |
| OFV | −323.2803 | **−610.0915** | |
| TVCL | 2.5889 (−48%) | **4.99536** (−0.09%) | 5.0 |
| TVV | 43.4805 (−13%) | **49.70493** (−0.59%) | 50.0 |
| TVKA | 0.7092 (−29%) | **0.98482** (−1.5%) | 1.0 |
| TVLAG | 2.0000000137 | 1.999999992 | 2.0 |
| ω² | 162 754.79 (at rail) | 1e-8 | 0 |
| σ | 0.15554 | 0.015277 | |
| runaway warning | `Critical` | none | |

The test only ever asserted `theta[3]`, so it called that first column a success — TVCL was 48%
off and nothing looked. It now also pins TVCL / TVV / TVKA at a 5% relative tolerance.

The fit's own diagnostics corroborated the diagnosis independently: `TVV ~ TVKA = -1.00`,
`TVV ~ TVLAG = -1.00`, `TVKA ~ TVLAG = 1.00`, RSE up to 11 650%, 3 of 6 Hessian eigenvalues
clipped.

## Alternatives considered

Relaxing the `converged` assertion (accept the runaway, keep asserting TVLAG). Rejected: it
keeps a degenerate fixture and re-hides exactly what #1205 was built to surface. The numbers
above are the argument — under that alternative the test would still be certifying a fit with
TVCL 48% off.

## Cross-repo dependency

## Breaking changes
- [x] None

Test-only. No `src/` change, no public API change, no `.ferx` format change.

## Numerical validation
- [x] Reference values:

```
# before (free omega) — the state the nightly was asserting against
converged = false
ofv       = -323.2803206134488
theta     = [2.5889440954763097, 43.48050992397635, 0.7092055671949157, 2.0000000137391107]
omega     = [162754.79141900392]      # packed coordinate 6.0000 == the +6 rail
sigma     = [0.15554372229775465]
WARNING (Critical/ParameterAtRunawayGuard): ETA_CL ... at upper guard, runaway

# after (omega FIX'd)
converged = true
ofv       = -610.0915442332105
theta     = [4.995361202997973, 49.70493043301984, 0.9848154826201589, 1.999999992082745]
omega     = [1.0000000000000017e-8]
sigma     = [0.015276884273925536]
runaway guard warning present = false
```

Truth is `theta = [5.0, 50.0, 1.0, 2.0]` (the data-generating parameters — the data are
`predict()` at truth, noise-free), so the second block recovers all four.

No NONMEM anchor: this changes no numerical path. It corrects a test fixture's identifiability
so the existing analytic-gradient FOCEI path is exercised on a well-posed problem.

## Performance

## Tests

Mutation-verified, not assumed. Dropping the `FIX` from `ROUTE_LAG_SINGLE`'s line **alone**
(leaving the other eleven) reproduces the original failure exactly:

```
thread 'per_route_lag_fit_recovers_lag' panicked at tests/per_route_lag.rs:636:5:
per-route-lag fit should converge
test result: FAILED. 15 passed; 1 failed
```

Restored:

```
test result: ok. 16 passed; 0 failed
```

The new TVCL/TVV/TVKA assertions are bounded by measurement, not preference: 5% is ~3x above
the worst error the sound fit produces (TVKA, 1.5%) and ~1/9 of the smallest error the
degenerate fit produced (TVV, 13%), so the band separates the two states with room on both
sides.

One trap worth recording, since it nearly produced a false "the fix does not work": restoring
the mutated file with `mv file.bak file` gave it the backup's **older** mtime, so cargo
considered the target up to date and re-ran the *mutated* binary. `touch` before re-running.

## Example
- [x] Not applicable (internal / test fix with no new API surface)

## Docs
- [x] No user-visible change

Test-only, so no `CHANGELOG.md` entry per `CLAUDE.md`.

## Checklist
- [x] No version bump needed

## Reviewer hints

The load-bearing line is **one**: `ROUTE_LAG_SINGLE`'s omega (line 34). The mutation above
proves it and nothing else in the diff.

The other eleven `FIX`es are deliberate uniformity, and are the part worth pushing back on if
you disagree. `FIX` is a no-op for a model that is only predicted, so none of them changes any
current test; four of them (`ROUTE_LAG_ZERO`, `ROUTE_LAG_PARALLEL`, `ROUTE_LAG_MIXED`, and
`ROUTE_LAG_SINGLE` itself) already carry `[fit_options] method = focei` and would walk into the
identical trap the moment someone adds a fit test; and leaving one twin `FIX`'d while its pair
stays free would undercut the "compared at identical parameters" claim the route-lag /
compartment-lag equalities rest on. Happy to narrow it to line 34 if you would rather keep the
diff minimal.

## Open questions

None blocking. One observation for whoever owns #1118 follow-up: a *free* omega declared at
exactly `~ 0.0` is a shape that can only ever end at a rail or at the floor — the packed start
is `log(0)`. Worth considering whether `fit()` should say so up front rather than let it run,
but that is a separate change and not this PR.

Closes #1227.

### DIFF
--- a/tests/per_route_lag.rs
+++ b/tests/per_route_lag.rs
@@ -19,7 +19,7 @@
 mod common;
 
 use ferx_core::parser::model_parser::parse_full_model;
-use ferx_core::{check_model_data, predict, DoseEvent, Population};
+use ferx_core::{check_model_data, predict, DoseEvent, Population, WarningCode};
 
 // ── Per-route-lag models under test ──────────────────────────────────────────
 
@@ -591,12 +591,22 @@ fn per_route_lag_fit_runs_end_to_end() {
 /// The models in this file declare `omega ETA_CL ~ 0.0 FIX`, and the `FIX` is what makes
 /// this test's `converged` assertion meaningful (#1227). Fitting **one** subject leaves ω²
 /// with no between-subject information to estimate from, so a *free* zero-variance omega is
-/// unidentifiable by construction: nothing bounds it above and it walks to `compute_bounds`'
-/// `+6` Cholesky-diagonal ceiling (ω² ≈ 162 755). Since #1205 that rail is a `runaway`
-/// verdict and demotes `converged`, so the assertion below failed on every nightly run —
-/// on a fit whose TVLAG was nonetheless recovered to eight decimals. Declaring the omega
-/// FIX'd matches what the model already says (zero IIV) and removes the coordinate from the
-/// packed vector, rather than teaching the assertion to accept a runaway.
+/// unidentifiable by construction and walks to `compute_bounds`' `+6` Cholesky-diagonal
+/// ceiling (ω² ≈ 162 755). Since #1205 that rail carries a `runaway` verdict and demotes
+/// `converged`, so the assertion below failed on every nightly run — on a fit whose TVLAG
+/// was nonetheless recovered to eight decimals.
+///
+/// Worth knowing before reading that as "it starts at ln(0)": it does not. The parser floors
+/// a declared `~ 0.0` to a Cholesky diagonal of `1e-4` — measured, and *identical* with and
+/// without `FIX` — so the free coordinate's packed start is `ln(1e-4) = -9.21`, which is
+/// below its own `-6` lower rail (`parameterization.rs`, `compute_bounds`). It begins clamped
+/// outside its admissible interval and ends at the opposite one. `FIX` changes nothing but
+/// `omega_fixed`, which is why it is inert for the predict-only twins here (verified
+/// bit-identical, 0 ulp over the sampled curve) and decisive for the two fits: the coordinate
+/// leaves the packed vector instead of being optimized from a start it cannot reach.
+///
+/// So the fix declares what the model already says — zero IIV — rather than teaching the
+/// assertion to accept a runaway.
 #[test]
 #[cfg_attr(
     not(feature = "slow-tests"),
@@ -629,6 +639,15 @@ fn per_route_lag_fit_recovers_lag() {
         warnings: vec![],
         subjects: vec![common::subject("1", vec![dose], obs_times, obs, vec![1; n])],
     };
+    // Only KA and LAG start off truth; TVCL and TVV start at their data-generating values,
+    // deliberately. Perturbing them was tried (CL 5.0 -> 3.0, V 50.0 -> 70.0) and the fit does
+    // not come back: it returns `converged: true` with no guard hit and TVV at 43.55, 12.9%
+    // off. One subject of noise-free data identifies the absorption pair strongly and the
+    // disposition pair only weakly — the fit's own output says so, reporting
+    // `TVV ~ TVKA = -1.00`. So the two assertions on TVCL / TVV below are *drift* guards, not
+    // identifiability claims: they say a sound fit must not drag them away from a start that
+    // is already correct. That is exactly the regression they exist for — the runaway-omega
+    // fit moved them 48% and 13% off from these same starting values.
     let mut init = model.default_params.clone();
     init.theta[2] = 0.7; // KA off truth (1.0)
     init.theta[3] = 3.0; // LAG off truth (2.0)
@@ -639,12 +658,29 @@ fn per_route_lag_fit_recovers_lag() {
         "recovered per-route lag {} should be near 2.0",
         fitted.theta[3]
     );
-    // Pin the disposition too, not just the lag (#1227). While ω ran to the +6 rail it
-    // absorbed the structural signal and these came back TVCL −48%, TVKA −29%, TVV −13%
-    // on the same noise-free data — a fit this test called a success because it only ever
-    // looked at theta[3]. With the omega FIX'd the worst of the three is TVKA at 1.5%
-    // (TVCL 0.09%, TVV 0.59%), so 5% sits ~3x above what the sound fit produces and ~1/9
-    // of the smallest error the degenerate one produced. Both numbers are measured.
+    // Assert the *cause*, not only its downstream boolean. `converged` alone cannot tell
+    // "omega is identifiable" apart from "the #1205 demotion rule stopped firing" — a change
+    // to `RunawayGuardHit::is_runaway` would flip the boolean back to true with ETA_CL still
+    // pinned at +6, and every other assertion here would still pass.
+    assert!(
+        !fitted
+            .warnings_structured
+            .iter()
+            .any(|w| w.category == WarningCode::ParameterAtRunawayGuard),
+        "no estimate should sit at an internal guard; got: {:?}",
+        fitted
+            .warnings_structured
+            .iter()
+            .filter(|w| w.category == WarningCode::ParameterAtRunawayGuard)
+            .map(|w| w.message.as_str())
+            .collect::<Vec<_>>()
+    );
+    // Pin the disposition too, not just the lag (#1227). What is measured, twice, on the same
+    // noise-free data: with omega free the fit ends at the +6 rail and TVCL / TVV / TVKA come
+    // back −48% / −13% / −29% off truth; with omega FIX'd it converges and the worst of the
+    // three is TVKA at 1.5% (TVCL 0.09%, TVV 0.59%). The 5% band is set from those two
+    // measurements — ~3x above what the sound fit produces, ~1/9 of the smallest error the
+    // degenerate one produced — not from a view about what a good tolerance looks like.
     for (i, truth, name) in [(0usize, 5.0, "TVCL"), (1, 50.0, "TVV"), (2, 1.0, "TVKA")] {
         let rel = (fitted.theta[i] - truth).abs() / truth;
         assert!(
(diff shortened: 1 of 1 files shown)


TeunP:
## Self-review (xhigh) — 7 findings, all resolved

Reviewed at xhigh effort across ten finder angles. Five fixed in `81c47345`, one reversed on
measurement, one split out as #1229. Ranked by severity.

---

### 1. The `FIX` was defended only by a nightly-gated test — `test-coverage` — **fixed**

`per_route_lag_fit_recovers_lag` is `#[cfg_attr(not(feature = "slow-tests"), ignore)]`, so PR CI
skips it. Dropping a `FIX` would have merged green and reappeared in the next nightly — the exact
loop #1227 was filed to end.

`per_route_lag_fit_runs_end_to_end` does not close the gap either. It fits the same model but with
`outer_maxiter: 2` and asserts only `fitted.ofv.is_finite()`, and an omega parked at the `+6` rail
gives a perfectly finite OFV.

Added `per_route_lag_fixtures_fix_their_zero_variance_omega`: ungated, parse-only, covers all ten
module-level fixtures. It also fails if a fixture declares no omega at all, so it cannot go vacuous.

### 2. The test asserted the boolean, not the cause — `test-coverage` — **fixed**

`converged` alone cannot tell "omega is identifiable" apart from "the #1205 demotion rule stopped
firing". A change to `RunawayGuardHit::is_runaway` (`src/api/postfit.rs:1147`) would flip the
boolean back to `true` with `ETA_CL` still pinned at `+6`, and every assertion in the test would
still pass. Now asserts no `WarningCode::ParameterAtRunawayGuard` entry in `warnings_structured`.

### 3. Doc comment said the coordinate starts at `ln(0)` — `correctness` — **fixed**

It does not, and the truth is more interesting. The parser floors a declared `~ 0.0` to a Cholesky
diagonal of `1e-4` — measured, and **identical with and without `FIX`** — so the packed start is
`ln(1e-4) = -9.21`, which is *below* that coordinate's own `-6` lower rail
(`compute_bounds`, `src/estimation/parameterization.rs:554`).

The free coordinate begins clamped outside its admissible interval and ends at the opposite one.
Corrected in the comment, and it is what makes #1229 worth filing.

The same measurement settles what `FIX` actually changes: only `omega_fixed`. Predictions from the
free and FIX'd spellings are bit-identical over the sampled curve — 48/48 points, 0 ulp gap — so
the "inert for the predict-only twins" claim in the PR description is now verified rather than
asserted. Worth stating plainly because the green suite could *not* have established it: every
twin-equality test in this file compares two models that both changed, which is the shared-
convention blind spot `CLAUDE.md` warns about.

### 4. TVCL/TVV asserted but never perturbed — `test-coverage` — **no change, reversed on measurement**

The review was right that both start at their data-generating truth, so the assertions on them are
satisfied at the start rather than earned.

Perturbing them (CL 5.0 → 3.0, V 50.0 → 70.0) was implemented, run, and **reverted on the result**:

```
TVV = 43.55163813519556 is 12.9% off truth 50
```

— with `converged: true` and no guard hit. One subject of noise-free data identifies the absorption
pair strongly and the disposition pair only weakly; the fit says so itself, reporting
`TVV ~ TVKA = -1.00`. Asserting recovery there would demand something this fixture cannot deliver,
and tuning the start until it passed would be fitting the test to the answer.

So the two assertions stay as **drift guards** — a sound fit must not drag TVCL/TVV away from a
start that is already correct — and the comment now says exactly that, with the number. That is
what they are good for, and it is the regression they exist to catch: the runaway-omega fit moved
them 48% and 13% off from these same starting values.

### 5. A comment asserted a mechanism that was not measured — `conventions` — **fixed**

The original wording said the runaway omega "absorbed the structural signal". What was measured is
the correlation, not the mechanism; no experiment isolated it. Reworded to state the two
measurements and let them stand.

### 6. Eleven of twelve `FIX` edits had no test defending them — `test-coverage` — **fixed**

The mutation confirms only line 34 is load-bearing today. The new Tier-2 guard now covers all ten
module-level fixtures, so the other eleven are enforced rather than merely conventional.

### 7. Root cause left live for user models — `altitude` — **deferred to #1229**

Fixing the fixture is correct for a test-only PR, but any user writing `omega ETA_CL ~ 0.0` without
`FIX` walks into the same trap. Split out as #1229 with the measurement and a concrete suggestion
(check it in `api::validation`, where it is statically decidable).

---

## Mutation verification

Neither new assertion is taken on trust.

**A — drop `FIX` from `ROUTE_LAG_SINGLE` alone** (the other eleven left in place):

```
per_route_lag_fixtures_fix_their_zero_variance_omega  panicked at :730  (omega_fixed assert)
per_route_lag_fit_recovers_lag                        panicked at :655  (converged assert)
test result: FAILED. 15 passed; 2 failed
```

**B — same, plus the `converged` assert removed**, to prove the new guard assert is reachable and
not shadowed by the one above it:

```
panicked at :665:
no estimate should sit at an internal guard; got: ["Internal optimizer parameter guard reached
by estimate(s): ETA_CL (estimate 162754.7914; packed coordinate 6.0000 at upper guard, runaway) ...
```

Restored: `test result: ok. 17 passed; 0 failed`.

One trap that nearly produced a false "the fix does not work", recorded in case it bites someone
else: restoring a mutated file with `mv file.bak file` gives it the backup's **older** mtime, so
cargo considers the target up to date and re-runs the *mutated* binary. `touch` before re-running.

## Gates

`preflight OK — fmt check clippy rustdoc docs public-api`, and `per_route_lag` 17/17.

## Still open for a human reviewer

The eleven non-load-bearing `FIX`es remain a judgement call — see the Reviewer hints section of the
PR description. Now that the Tier-2 guard enforces them, reverting to just line 34 would mean
narrowing that guard too, so it is one decision rather than two.

TeunP:
## Verified on Linux: `slow-tests` dispatch green

The gate this PR actually needed. `Tests + coverage (full, slow-tests)` reports `skipping` on a
PR, so **none of the ten green checks above exercises the test being fixed** — it is
`slow-tests`-gated. Ran the nightly workflow directly against the branch:

[run 33910512198](https://github.com/FeRx-NLME/ferx-core/actions/runs/33910512198) — **success**

```
test per_route_lag_fit_recovers_lag ... ok
test per_route_lag_fixtures_fix_their_zero_variance_omega ... ok

running 17 tests   (tests/per_route_lag.rs)
test result: FAILED  x0   across the whole suite
5494 passed
```

Reconciles against the broken baseline exactly: nightly was 5492 passed / 1 failed, now
5494 / 0 — `+1` for the test that was failing, `+1` for the new Tier-2 guard. Nothing else moved.

## The first dispatch failed, and it was my bug

Worth recording, because a local green and ten green PR checks both missed it.

`81c47345`'s new assertion keyed on `WarningCode::ParameterAtRunawayGuard` — but **that category
carries both verdicts**, so it also rejected a `collapse`:

```
PROP_ERR (estimate 0.0003; packed coordinate -8.0000 at lower guard, collapse)
```

That collapse is expected and correct. The data are noise-free, so the residual SD has nothing to
estimate and `PROP_ERR` walks to its `-8` floor — which #1205 deliberately leaves as a `Warning`
that does not demote. `converged` passed; only the over-broad assertion failed.

It was green locally for a reason unrelated to being right: **macOS stops at σ = 0.0153, Linux CI
reaches 3e-4**. Wrong predicate, concealed by a platform difference (#990 territory).

`68b52f9a` keys on the per-hit `verdict` in `details.parameters[]`, falling back to the
`"guard, runaway"` message token that `classify_warning` uses if the structured payload is ever
absent — rather than passing silently on a missing payload. Re-mutation-verified on the narrowed
predicate: drop the `FIX` and remove the `converged` assert, and it still fails naming `ETA_CL` at
`"packed coordinate 6.0000 at upper guard, runaway"`.

Note also that the first dispatch died at the guard assert and **never reached** the
TVCL / TVV / TVKA checks, so their behaviour on Linux was unproven until this run. It is proven
now — the 5% band, calibrated on macOS numbers, holds on Linux too.

## Status

Green on both platforms, the fixed test verified on the platform where it was red, and the full
nightly suite clean for the first time in five runs. Ready for a human call on the one open
judgement in the Reviewer hints (the eleven non-load-bearing `FIX`es).

Issue #626, opened 2026-06-30
perf(estimation): joint PK-TTE FD-Hessian re-solves the augmented ODE per perturbed η (analytic CHZ η-sensitivities)


TeunP:
Follow-up to #570 (shared via PR #613).

**#613 removed the _double_ augmented-ODE solve** from the two hot spots of a joint PK-TTE FOCEI fit — the inner EBE evaluation and the FOCEI at-mode: the Gaussian predictions and the cumulative-hazard `H`/`h` are now read off a *single* integration (in-step cubic Hermite interpolation), instead of a second dedicated `ode_cumhaz_hazard` solve.

**What remains:** the FOCEI **FD-Hessian** of the TTE data term (`data_term_hessian_fd`) still re-integrates the augmented ODE once per perturbed-η evaluation — `2·n(n+1)` solves per subject per outer iteration. These are *single* (not double) solves and are intrinsic to finite differencing, so #570's sharing doesn't touch them; at larger `n_eta` they dominate the per-fit cost.

**Approach:** eliminate them with **analytic CHZ η-sensitivities** — propagate `Dual2` through the cumulative-hazard ODE state so `∂(−log L_TTE)/∂η` and its curvature come from forward sensitivities (like the existing Gaussian `sens/` path) rather than finite differences. This is a separate, larger change in the `sens/` machinery.

Not urgent (per-fit cost is acceptable after #613). Filed so the follow-up isn't lost when #570 closes.

TeunP:
## Precondition carried over from #1220 item 4 — the dual SS equilibration has no accumulator handling

Recording this here before closing #1220, because #626 is the issue that will actually hit it and nothing here currently mentions it.

`equilibrate_ss_state_g` / `equilibrate_ss_input_rate_state_g` still cycle the injected `d/dt(__chz_<cmt>)` accumulator the way the `f64` path did **before** #1210. #1210 did not fix them — it made them **unreachable**, via `ss_dual_equilibration_out_of_scope` (accumulator slots present *and* the subject has SS doses), declined on both ODE routes, with an `assert!` behind it as a tripwire.

[...]

Issue #1220, opened 2026-09-03
test(survival): remaining external anchors and tiers left over from #1210 (SS x ODE-accumulated hazard)


TeunP:
Spun out of #1210 (PR #1219) to keep that PR to the defect it was opened for. Everything here
is **covered by closed-form Tier-1 tests plus verified mutation**; what is missing in each case
is an *external* (NONMEM) anchor, or a tier.

#1210 landed four NONMEM constructions. For context on what is already anchored:

| behaviour | anchor |
|---|---|
| NONMEM's own SS routine cannot be handed the augmented system | `ss_chz_r1_*` — the negative result (`#OBJV = +INF`, `NUMERICAL DIFFICULTIES`) |
| `H(0) = 0` first-dose convention; `H`, `h`, `PRED` trajectories | `ss_chz_r2_{const,drug,tdep}` — 2e-8, three arms |
| a **second** `SS=1` dose preserves the hazard accrued so far | `ss_chz_r3_{const,drug}` — mutation-verified: zeroing disagrees by 99.8% |
| a reset zeroes the accumulator and restarts the clock | `ss_chz_r4_const` — `EVID=4`, 2e-8 |

## 1. A lagged mid-record `SS=1` dose has no external anchor

`a_mid_record_ss_dose_lagged_past_the_interval_keeps_the_hazard_accrued_so_far` (Tier-1) pins
the case where an `SS=1` dose is *both* mid-record and lagged past `II`. It is the only thing
holding `restore_chz` inside `ss_state_at_phase` — mutation `M8` kills exactly that test and
nothing else — but it is checked against a closed form, not against NONMEM.

Why it was left out rather than built:

- The **preserve** half is already externally anchored by r3.
- The SS-plus-lag **PK** convention is already externally anchored, by #1121's
  `tests/ss_lagtime_edge_nonmem_anchor.rs` (16 functions, four engines, including
  `ss_lag_ge_ii_matches_nonmem_on_the_*`).
- What is left is the intersection, and **the const arm structurally cannot see it**: with
  `BETA = 0` the hazard is `H0` regardless of the PK, so `H(t) = 0.02 t` holds on both sides
  whatever the lag does. It needs a *lagged drug* train — a new construction.

That new construction is the actual work item. Note the trap it has to avoid, which bit r3's
first draft: an `SS=1` dose means "the infinite past was periodic **up to here**", not "keep
[...]

PR #1230, opened 2026-09-04
fix(ode): seed a joint PK-TTE hazard time before the integration start on the shared solve [closes #1223]

### LINKED ISSUE #1223: bug(ode): one-solve joint PK-TTE share NaN-fills a hazard time before the integration start while the two-solve path seeds it — the presence of a PK row decides the likelihood
## What

The two engines that read a joint PK-TTE subject's cumulative hazard disagree on a TTE time
**before the integration start**, and #570's one-solve == two-solve equivalence breaks there:

- the two-solve arm (`ode_cumhaz_hazard` → `ode_dense_solve_states`) fills a `saveat` node
  earlier than the first break with the **seeded initial state** — `src/ode/predictions.rs`,
  the `t < first_start - 1e-12` prefill, added for the CTMM scorer — so `H = 0` there and
  `tte_nll_from_curves` scores it;
- the one-solve share (`ode_predictions_and_chz`, `src/ode/predictions.rs` ~2477 and ~2755)
  keeps its `f64::NAN` prefill for a `chz_times` entry before the start, which the TTE
  likelihood maps to the `1e20` sentinel. Its doc says this is "matching the dedicated
  `ode_dense_solve_states` path", which is no longer true.

So a subject with a left-truncation `entry_time`, or an interval-censored `left`, before its
first dose scores a finite likelihood with no PK observation (two-solve path) and is repelled
with the `1e20` sentinel the moment a PK row is added (shared path). Which of the two is the
right convention for a hazard time before any event is a separate question; the point here is
that the presence of a PK row must not decide it.

## Where it surfaced

Review of #1222 (#1218), which makes the two-solve path finite on a one-break timeline. Not
introduced there — the prefill predates it — but #1222 is the point where the two paths' `NaN`
contracts diverge everywhere, and the one-solve doc comment is stale as of that prefill.

## Suggested direction

Pick one convention (the seeded state is the physically consistent one: nothing has acted on the
system yet, so `H = 0` and `h = h(seeded state)`), apply it to the one-solve `chz_states`
prefill, and add the #570 parity assertion for a pre-start `entry_time` / `left` so the two
engines cannot drift on it again. Reading of the source, not a bisect.

TeunP:
## Why

The two engines that read a joint PK-TTE subject's cumulative hazard disagreed on a TTE time
**before the integration start**, and #570's one-solve == two-solve equivalence broke there.

* The two-solve arm (`ode_cumhaz_hazard` → `ode_dense_solve_states`) fills a `saveat` node
  earlier than the first break with the **seeded initial state** — the `t < first_start - 1e-12`
  prefill added for the CTMM scorer — so `H = 0` there and `tte_nll_from_curves` scores it.
* The one-solve share (`ode_predictions_and_chz`) kept its `f64::NAN` prefill for a `chz_times`
  entry before the start. `tte_ode_nll_from_shared` skips a non-finite state, so `H` came back
  `NaN` and `tte_nll_from_curves` mapped it to the `1e20` sentinel.

So a subject with a left-truncation `TENTRY`, or an interval-censored `left`, at or before its
first record was **scored on one engine and repelled on the other** — and which engine it
reached is decided by `try_joint_pktte_shared_solve`'s admission list (EVID-3/4 resets,
time-varying covariates, model time in the PK block, SDE, FREM), none of which has anything to
do with where its entry time falls. The share's doc comment claimed the NaN was "matching the
dedicated `ode_dense_solve_states` path"; it had not been true since that prefill landed.

Measured on `main` at `19fdbb04`, one subject (dose at 10, PK obs at 12, exact event at 20):

```
TENTRY = 0  →  individual NLL 306.2186273580891
TENTRY = 5  →  individual NLL 100000000000000000000     (the 1e20 sentinel)
```

Found by review of #1222 (#1218), which is where the two paths' NaN contracts diverged
everywhere. The prefill predates it.

## What changed

**The fix is one block** in `ode_predictions_with_extra_breaks_and_stats`, immediately after
the `sort_by` / `dedup_by` of `break_times` — the exact twin of the fill in
`ode_dense_solve_states`: `chz_states[gi] = u.clone()` for every `chz_times[gi] < break_times[0]
- 1e-12`. `u` is still the freshly seeded `ode.initial_state(pk)` at that point, so both engines
clone the same value. NaN now means one thing on both engines: a diverged solve.

Two details that are load-bearing:

* **Keyed on `break_times[0]`, not on `subject_integration_start`.** Both engines fold a
  terminal horizon into the timeline before sorting — the share `max(0, obs…, chz…)`, the dense
  builder `max(0, saveat…)` — so a timeline whose every observation and CHZ time precedes the
  first dose puts the first break *below* the start. There the node is read at the `k = 0`
  boundary visit as the same seeded state by the other mechanism (#1218). Keying on the start
  would diverge from the twin in exactly that corner.
* **Strict `<`**, with the same `1e-12`, so a CHZ time *on* the first break still reads at that
  boundary visit, post-dose (the #1218 caveat).

Four doc comments said the pre-start state was NaN "matching the dedicated path" and all four
became false with the fix; all four are rewritten (`ode_predictions_and_chz`'s rustdoc, the
`chz_states` prefill comment, `JointPkTteSolve::chz_states`, `tte_ode_nll_from_shared`).

`tests/pktte_prestart_entry.rs` is added to the `Tests + coverage (TTE/CTMM endpoints)` job's
explicit `--test` list in `ci.yml` — `tests/ci_workflow_endpoint_coverage.rs` pins that list as
a set equality against every `survival`/`markov`/`nn`-gated test file.

## Alternatives considered

The competing convention is to start the hazard clock at the earliest TTE time, which would
make `H(entry) > 0` for a pre-start entry and shift the objective. The seeded state is the
physically consistent reading — nothing has acted on the system yet — and it is what the
dedicated engine and `predict_survival()` have always reported, so adopting it keeps three
callers on one answer instead of introducing a fourth. The positive control below is what
distinguishes the two conventions.

**Deliberately not decided here:** NONMEM integrates from the subject's first record of *any*
`EVID`; ferx starts at the first dose or scored observation, so an `MDV = 1` row (dropped by
the reader) or an `EVID = 2` row on a model without time-varying covariates does not start the
hazard clock. That divergence is documented, not changed.

## Cross-repo dependency

No public API change (the `Public API baseline` gate is green with no baseline edit).

## Breaking changes
- [x] None

## Numerical validation

**No NONMEM anchor is possible for this object.** `TENTRY` has no NONMEM analogue — NONMEM has
no left-truncation record for an ODE-accumulated hazard — so there is no equivalent run to
anchor against. **No `Dual2`-vs-FD parity test applies either:** the ODE-accumulated hazard has
no analytic sensitivity (#626 is open), so its η-gradient and FD-Hessian are finite differences
over this same value path; the fix moves the objective from "repelled" to "scored" consistently
on value and gradient, and there is no second derivative path to disagree with it.

What replaces them is a **hand-computed oracle plus three callers**. At a pre-start time the
values are exact and independent of any engine: the seeded state has `central = 0`, so

```
H = 0                                   exactly
h = H0 · exp(BETA · 0/V) = H0 = 0.01    bit-exactly (exp(0.0) == 1.0), for any η
```

Asserted on all three production callers — the #570 share, `ode_cumhaz_hazard`, and the public
`predict_survival()`. Note that `predict_survival` is a third **caller**, not a third geometry:
it routes `OdeAccumulated` straight to `ode_cumhaz_hazard`. It is there because it is the entry
point users read `H`/`h` from, not as independent evidence — the hand-computed values are the
independent part.

End-to-end objectives, `maxiter = 0` on `nonmem_anchor/ss_chz_drug_fit.ferx` (FIXed thetas,
`ode_reltol 1e-9`), loaded through the routed `read_population_for`:

```
# A1 — dose@10, PK obs@12, event@20; qualifies for the shared solve
TENTRY = 0   24.417862686939927
TENTRY = 5   24.417862686939927        (was: 1e20 sentinel)

# A4 — A1 + an EVID=3 reset@15 and a re-dose@16; declines the share, dedicated engine
TENTRY = 0   16.97903614243416
TENTRY = 5   16.97903614243416         (unchanged by this PR — the control)
```

## Performance
- [x] No significant impact expected — one linear scan over `chz_times` (typically 1–3 entries)
      per subject evaluation, outside the integration loop.

## Tests

Every new assertion was **observed red under mutation, in both directions**, and the runs are
reproduced below.

**Tier 1 — driver twin** (`src/ode/predictions_tests.rs`):
`prestart_chz_time_reads_the_seeded_state_on_both_paths` — `ode_predictions_and_chz` vs
`ode_dense_solve_states` at `chz_times = [5.0, 12.0]`, dose at 10, obs at 12. `is_finite()` per
engine, then `to_bits()` equality against each other and against `ode.initial_state(pk)`.

**Tier 1 — objective** (`src/stats/likelihood.rs`, beside the #570 harness):
`prestart_tte_time_is_scored_on_share_dedicated_and_predict_survival` — three arms (a
left-truncation `entry_time`, an interval-censored `left`, a pre-start right-censoring time) ×
three callers, with the straddle asserted and `is_finite()` before every comparison.
`prestart_entry_time_contributes_nothing_post_start_entry_does` — the positive control.

**Tier 2** (`tests/pktte_prestart_entry.rs`, new): the A1 / A4 / A5 arms above, through the
routed loader from temp CSVs.

### Three ways these tests could have been unable to fail, and what was done about each

* **`to_bits()` alone is green on the broken code.** Both engines pre-filled with the
  `f64::NAN` *literal*, and `f64::NAN.to_bits()` is one fixed pattern — so a bare bitwise twin
  passes with NaN matching NaN. The `is_finite()` assertions are the only ones that fail before
  the fix, and they are asserted first, per engine, so the message names which engine dropped
  the node.
* **The driver twin's state check was zeros-against-zeros.** `one_cpt_chz_ode_spec` has
  `init_fn: None`, so `initial_state` is all zeros and "state == seeded" would have held just as
  well for a fill written `vec![0.0; n]`. The fixture now seeds a non-zero `init(central) = 7.5`,
  and asserts the seed is non-zero so it cannot silently become degenerate again. (In the
  likelihood arms the seeded state *is* all zeros; there the live quantity is `h = 0.01`, and the
  test says so.)
* **The driver twin's dense leg did not reach the dense fill.** Caught by running the reverse
  mutation: with `chz_times = [5.0]` the test stayed **green** while the dense fill was
  disabled. The dense engine folds its horizon from `saveat` — which *is* `chz_times`, not the
  subject's observations — so a lone pre-start time put `break_times[0]` below the start and the
  node was read at the `k = 0` boundary visit instead. Adding the second CHZ time at `12` lifts
  both engines' horizons above the start so both take the fill; that node doubles as a live
  control (post-dose, ≠ the seed, both engines agree).
* **One tolerance was asserted rather than measured.** The positive control's comment said the
  pre-vs-no-truncation difference was "measured at 0 (exact)"; it never was — the test only
  checked a `1e-9` relative bound. Measured on a follow-up commit: `306.2186273580891` both
  ways, difference exactly `0.0`, and bit-exact *by construction* — the node the `entry_time`
  adds is filled before the break walk, lies in no segment, and does not move the horizon
  (`t_last = max(obs 24, chz) = 24` either way), so the break timeline, every segment and the
  Gaussian solve are identical. The Tier-1 control and both Tier-2 legs now assert `to_bits()`
  equality instead of a bound, and the post-start separation is recorded (`1.363e-3` relative)
  next to the `1e-6` floor it justifies. **Mutation was re-run against the tightened
  assertions**, since the evidence below was collected against the old ones — same outcome.

### Mutation evidence

Remove the **share** fill:

```
test ode::predictions::tests::prestart_chz_time_reads_the_seeded_state_on_both_paths ... FAILED
  shared engine left the pre-start state non-finite: [NaN, NaN]
test stats::likelihood::tests::prestart_tte_time_is_scored_on_share_dedicated_and_predict_survival ... FAILED
  left truncation: Exact @20, entry 5: share left the pre-start state non-finite: [NaN, NaN, NaN]
test stats::likelihood::tests::prestart_entry_time_contributes_nothing_post_start_entry_does ... FAILED
  a pre-start entry must contribute nothing: entry=5 100000000000000000000 vs entry=0 306.2186273580891
```

Remove the **dense** fill (the share legs stay green, and the dedicated ones name `1e20`):

```
test ode::predictions::tests::prestart_chz_time_reads_the_seeded_state_on_both_paths ... FAILED
  dense engine left the pre-start state non-finite: [NaN, NaN]
test stats::likelihood::tests::prestart_tte_time_is_scored_on_share_dedicated_and_predict_survival ... FAILED
  left truncation: Exact @20, entry 5: share 4.953499682149126 must match dedicated 100000000000000000000 to solver tol
test stats::likelihood::tests::prestart_entry_time_contributes_nothing_post_start_entry_does ... ok
```

The positive control staying green under the *dense* mutation is correct, not a gap: its
subject qualifies for the share.

## Docs
- [x] `docs/` updated — `docs/estimation/tte.qmd`, `## Joint PK-TTE`: what a hazard time at or
      before the first record contributes, and where ferx's integration clock starts relative to
      NONMEM's. `cargo run -p docs-lint -- --check` green (via `tools/preflight.sh docs`).
- [ ] New pages linked in `docs/_quarto.yml` — N/A, no new page

## Checklist
- [x] `cargo test --lib --tests --no-default-features --features ci,survival` — 171 binaries,
      5015 passed, 0 failed, 170 ignored
- [x] `cargo test --lib --tests --no-default-features --features ci` — 171 binaries,
      4695 passed, 0 failed, 170 ignored (includes
      `ci_workflow_endpoint_coverage::endpoint_coverage_job_lists_every_feature_gated_test_file`,
      which is what required the `ci.yml` `--test` entry)
- [x] Functions on the `Dual2` sensitivity path — not touched; the ODE-accumulated hazard has no
      analytic sensitivity (#626)
- [x] `Cargo.toml` version bump — not needed (not breaking)

## Docs & examples
- [x] No new DSL syntax, `[fit_options]` key, example `.ferx` or data file — nothing to mirror
      to ferx-site / ferx-book, and no example execution step needed.

## Reviewer hints

Focus on the **placement** of the new block and on the `break_times[0]` keying — the argument
that it is the dense block's twin rests on `u` still being the freshly seeded
`ode.initial_state(pk)` at that point (nothing between its binding and the sort mutates it) and
on both engines folding their horizon in before the sort. The three `is_finite()` /
non-degeneracy notes under "Tests" are the parts most worth checking adversarially: two of the
three were mistakes I made and then caught by mutation, and the third (the dense-leg blindness)
was invisible until the reverse mutation was actually run.

## Open questions

Whether an `MDV = 1` or `EVID = 2` record *should* start the integration clock on the joint
path — the NONMEM divergence documented in `tte.qmd`. Deliberately left as a docs sentence here;
worth its own issue if the answer is "yes".


TeunP:
## Self-review, xhigh (10 angles + gap sweep) — 8 findings

Reviewed `origin/main...HEAD` at `3518b337`. Nothing here says the fix is wrong — the
production change survives the correctness angles (see "what held up" at the end). The findings
are one altitude problem, three coverage gaps, two doc/assertion defects, one efficiency nit and
one conventions call for the maintainer.

---

### 1. `src/ode/predictions.rs:2852` — altitude — the fix adds a *second copy* of the thing whose drift caused this bug

#1223 exists because `ode_dense_solve_states`'s pre-start fill (`predictions.rs:6476`) and the
share's prefill drifted apart. This PR restores agreement by writing a second, textually separate
copy ~3600 lines away, with its own `- 1e-12`, its own `break_times.first()` guard and its own
comment asserting it is the "exact twin of" the other. That is the same configuration that
produced the defect, now with two copies that happen to agree today.

A future edit to either one — tightening the epsilon, keying on `subject_integration_start`,
handling a new degenerate timeline — silently re-opens the identical class of bug, and the only
thing standing against it is prose.

**Fix:** extract one helper both engines call, e.g.

```rust
fn fill_prestart_states(times: &[f64], states: &mut [Vec<f64>], first_break: Option<f64>, u: &[f64])
```

and call it from both sites. Then the twin claim is enforced by the compiler rather than asserted
in a comment.

### 2. `src/stats/likelihood.rs:4173` — test-coverage — nothing exercises the FOCEI Laplace / FD-Hessian path with a pre-start entry

Before the fix the pre-start node returned the `1e20` sentinel, so an FD-Hessian built by
perturbing η around it was *garbage*, not merely a wrong scalar. Every new test is on the value
path: the Tier-1 tests call `individual_nll`, `tte_ode_nll_from_shared` and `tte_endpoint_nll`;
the Tier-2 test calls `fit`, but on `ss_chz_drug_fit.ferx`, which declares **no omega** (the fit
log reports `0 etas`), so no Hessian term is ever formed.

A regression that reinstated the sentinel only inside the Hessian perturbation — or that made a
perturbed solve non-finite while the base solve stayed finite — passes all six new tests.

**Fix:** add a `foce_subject_nll` leg, exactly as `joint_pktte_ode_hazard_nll_paths_finite`
already does for `entry_time = 0`.

### 3. `src/stats/likelihood.rs:4031` — test-coverage — `left = 0.0`, the issue's literal ungated shape, is untested

`try_joint_pktte_shared_solve` pushes `entry_time` only `if *entry_time > 0.0`, but pushes `left`
and `right` with **no gate**. So on a subject whose first record is at `t = 10`, a `left = 0.0` is
a pre-start CHZ time that reaches the share while a `TENTRY = 0.0` never does. That asymmetry is
what the docs sentence ("at or before") promises to handle, and it is the one shape a future
refactor could silently re-gate (adding `if *left > 0.0` for symmetry) with no test noticing. The
interval arm uses `left = 5.0`, which exercises the mechanism but not the asymmetry.

**Fix:** a fourth arm with `IntervalCensored { left: 0.0, right: 20.0 }`.

### 4. `tests/pktte_prestart_entry.rs:12` — test-coverage — the Tier-2 differential is one-sided

`prestart_entry_matches_no_entry_on_the_shared_engine` asserts `ofv(TENTRY=5) == ofv(TENTRY=0)`
bit-for-bit. An implementation that dropped `TENTRY` **entirely** — never pushing it into the
share's `times`, never subtracting `H(entry)` — satisfies that assertion perfectly, and satisfies
A4 and A5 too. The Tier-1 positive control carries the post-start leg, but at `individual_nll`
with a hand-built subject; nothing proves the *routed reader + fit* path honours a truncation
time at all.

**Fix:** add `ofv(arm_a1(12.0)) != ofv(arm_a1(0.0))`, so the end-to-end pair straddles the gate
the way the Tier-1 pair does.

### 5. `tests/pktte_prestart_entry.rs:12` — conventions — module doc says "Two arms" then documents three

`//! Two arms, and they are **not** symmetric:` is followed by A1, A4 **and** A5. The count was
right before A5 was folded in and was not updated.

### 6. `src/stats/likelihood.rs:4080` — correctness (test) — a combined assert misreports a length mismatch

```rust
assert!(st.len() == ode.n_states && st.iter().all(|x| x.is_finite()),
        "{arm}: share left the pre-start state non-finite: {st:?}");
```

If `chz_states` rows ever had a different width than `ode.n_states`, this fails with "share left
the pre-start state non-finite: `[0.0, 0.0]`" — a cause the printed value itself refutes, sending
the next reader into the wrong engine. Split into two assertions with their own messages.

### 7. `src/ode/predictions.rs:2852` — efficiency — the prefill scans all of `chz_times` though it is sorted

The caller contract (and the doc comment right above) guarantee `chz_times` is sorted ascending,
so the scan can stop at the first entry at or after the first break:
`chz_times.iter().take_while(|&&t| t < first_break - 1e-12)`. Negligible in absolute terms, but it
runs once per subject per objective evaluation inside the outer optimizer. Worth folding into
finding 1's helper rather than writing twice.

### 8. `CHANGELOG.md` — conventions — the NONMEM rule is waived under an exception that isn't enumerated

`CLAUDE.md` says "**Every new feature requires a comparison with NONMEM output.**" and lists
exactly two exceptions: adaptive/feedback dosing (#391), and survival-likelihood terms with no
equivalent NONMEM/survreg estimator — illustrated by *recurrent* left truncation under gap-time
RTTE (#740). This PR changes the scored objective for a **single-event** left truncation on an
ODE-accumulated hazard and waives the anchor in its own body.

The reasoning is sound (NONMEM has no `TENTRY` analogue for an ODE-accumulated hazard, and the
substitute oracle is an exact hand-computed `H = 0`, `h = H0`), but it is asserted by the PR
rather than covered by the list. Either extend the CLAUDE.md exception list to name this case, or
state in the PR that the rule is being waived by author judgement — a maintainer call, not
something I should decide.

---

### What held up under the correctness angles

Checked and clean, recorded so the next reader doesn't redo it:

* **`u` is untouched between binding and fill.** `u` is bound at `predictions.rs:2777` and its
  *only* use before the fill is the fill itself, so it is exactly `ode.initial_state(pk_params_flat)`
  — the same value the dense engine clones. The twin claim holds by construction.
* **No other caller is affected.** Three of the four call sites of
  `ode_predictions_with_extra_breaks_and_stats` pass `chz_times = &[]`, so the new loop iterates
  zero times for every non-#570 caller; the one non-empty caller passes `extra_breaks = &[]`, so
  the two can never both be non-empty and an `extra_break` cannot lower `break_times[0]` under the
  share.
* **`<` vs `<=` at the boundary is covered.** Mutating the share's strict `<` to `<=` makes it read
  the seeded state at a CHZ time *on* the first break while the dense engine still reads post-dose
  — `ode_predictions_and_chz_shares_one_solve` (which has a CHZ time at `0.0` = its integration
  start) fails.
* **No consumer relied on NaN meaning "pre-start".** `chz_states`'s only production reader is
  `tte_ode_nll_from_shared`, which skips non-finite states; after the fix a surviving NaN means one
  thing on both engines, a diverged solve.
* **Negative CHZ times** (an interval `left < 0`, which no gate prevents) now read the seeded state
  rather than the sentinel — consistent with the dense engine, not a new divergence.

TeunP:
## Second xhigh review — independent pass, 7 findings

Re-derived from the code at `68e30a4b` rather than from the first review's conclusions. The
first pass restructured the fix (one `fill_prestart_states` instead of two copies) and closed
four coverage gaps; this pass looks for what it missed. One finding is a real hole in the
argument the Tier-2 file makes about itself.

---

### 1. `tests/pktte_prestart_entry.rs` — test-coverage — nothing pins that a reset subject *declines* the share

This file's entire structure is "A1 goes through the share, A4 goes through the dedicated engine,
both must agree". **Neither routing is asserted anywhere.**

The admission list has tests for model time — `joint_pktte_share_rejects_time_builtin_model`,
`joint_pktte_share_rejects_model_time_in_the_odes_rhs` — and for a time-dependent hazard being
*admitted*, `joint_pktte_share_admits_a_time_dependent_hazard`. There is **no test for
`subject.has_resets()`**.

So if a future change admitted reset subjects to the share, A4 silently becomes a second run of
A1's engine. Every assertion in the file still passes; the pinned OFVs still match; and the guard
I wrote for exactly this family —

```rust
assert!((A4_OFV - A1_OFV).abs() > 1e-3,
        "A4 must be a materially different subject from A1");
```

— does **not** catch it, because it shows the two *subjects* differ, not that they took different
*engines*. The file's headline claim would be false while reading green.

Fixable only in Tier-1 (`try_joint_pktte_shared_solve` is private): assert `.is_none()` on a
reset-carrying subject, next to the existing `.is_none()` / `.is_some()` admission tests.

### 2. `src/ode/predictions.rs:196` — correctness — the extracted helper has an unasserted length precondition

```rust
fn fill_prestart_states(times: &[f64], states: &mut [Vec<f64>], first_break: Option<f64>, u: &[f64]) {
    ...
    for (i, &t) in times.iter().enumerate() {
        if t < first_break - 1e-12 { states[i] = u.to_vec(); }
    }
}
```

`states[i]` is indexed by an enumerate over `times`. `states.len() < times.len()` panics with a
bare out-of-bounds naming neither slice; `states.len() > times.len()` silently leaves the tail
unconsidered. Both callers satisfy it today (`chz_states` is `vec![..; chz_times.len()]`, `result`
is `vec![..; saveat.len()]`) — but **the extraction is what removed the context that made this
self-evident.** The loop used to sit a few lines under the allocation it was paired with; it now
lives 6,500 lines from one of its two callers.

Same class: `u.len()` vs `ode.n_states`, and `first_break` being finite — both callers guard
`timeline_has_non_finite` first, but the helper does not say it depends on that.

This file already has the idiom. `restore_chz` carries a `debug_assert_eq!` on exactly this kind
of paired-length invariant, with a comment noting that its fallback "degrades into *zeroing* the
accumulator, which is precisely the behaviour #1210 rejects".

### 3. `CLAUDE.md` — conventions — the new exception does not bound what it excludes

The exception added in `68e30a4b` says when it applies but never what it does *not* cover. The two
pre-existing exceptions are narrow and name their object (adaptive/feedback dosing; recurrent left
truncation under gap-time RTTE). This one is phrased around "outside the span NONMEM can express",
which a future author can reach for whenever a construction merely looks hard.

The counter-example is in the same subsystem, in flight right now: #1220's `r5_drug_lag` is an
SS × lagged × drug-driven-hazard object that looked unanchorable and **was** anchored, by dropping
`ALAG` from the NONMEM side and placing a plain pulse train at the ferx side's arrivals. The
exception should point at it, so the default stays "find the construction".

### 4. `src/stats/likelihood.rs:4054` — test-coverage — the `left = 0` arm silently doubles as a #1218 guard

That arm calls `predict_survival` with the grid `&[0.0]` — a zero-**maximum** grid, which is
exactly #1218's shape. It is currently free regression coverage for PR #1222's fix. But the
docstring frames the arm purely as the ungated-`left` asymmetry, so a future tidy-up changing
`t_pre` from `0.0` to some other pre-start value (reasonable-looking — the other arms use 5.0 and
8.0) would drop that coverage with nothing signalling the loss.

### 5. `tests/pktte_prestart_entry.rs` — efficiency — the A5 test refits two objectives it already has

`mdv_one_row_before_the_first_dose_changes_nothing` recomputes `ofv(&arm_a1(0.0))` and
`ofv(&arm_a1(5.0))`, both already computed by `prestart_entry_matches_no_entry_on_the_shared_engine`.
The file now runs nine `fit()` calls where seven would do — each a routed CSV read, a model parse
and a maxiter-0 FOCEI evaluation — in a Tier-2 file that runs on every PR.

### 6. `src/ode/predictions.rs:3037` — simplification — "shared verbatim" is no longer true

Both call sites describe the helper as "Shared **verbatim** with" the other engine — wording left
from when the two were verbatim copies. There is no longer a second copy to be verbatim with, and
the phrasing invites the next reader to go find it and check it still matches: the exact habit the
extraction exists to retire.

### 7. `tests/pktte_prestart_entry.rs` — simplification — pinned constants declared below their use

`A1_OFV` and `A4_OFV` sit at the bottom of the file, under the three tests that read them. Legal
and harmless, but they are the file's measured anchors and their provenance is the first thing a
reviewer wants; they belong with `MODEL` and `HEADER` at the top.

---

### What held up on the second read

Re-derived, not carried over from the first pass:

* **Both engines are guard-then-fill.** `timeline_has_non_finite` precedes `fill_prestart_states`
  in the share (`predictions.rs:3034` → `:3042`) *and* in the dense engine (`:6748` → `:6764`), so
  a broken timeline repels wholesale on both. This was the specific way the extraction could have
  reintroduced drift, and it does not.
* **There is no third engine with the same latent bug.** Of the six walks carrying a
  `timeline_has_non_finite` guard, only these two sample at *caller-supplied* times.
  `ode_predictions_with_states` samples at `subject.obs_times`, which feed
  `subject_integration_start` — so no sample of its can precede its own start, by construction.
* **`ode_dense_solve_states`'s behaviour is unchanged by the refactor** — same predicate, same
  epsilon, same clone, same position relative to the guard.
* **The `left = 0` arm genuinely reaches the ungated path**: `entry_time = 0` is not pushed
  (gated `> 0.0`), so `times = [0.0, 20.0]` comes from `left`/`right` alone.

Issue #1197, opened 2026-09-02
test(estimation): the IOV DSL-vs-column bit-identity assertion fails intermittently under full --lib parallel load


TeunP:
## Summary

`src/api/tests/iov_integration.rs::test_iov_occasion_dsl_matches_column_fit` asserts the
DSL-occasion and column-occasion IOV fits are **bit-identical**. It failed once under full
`--lib` parallel load, and passed on an immediate re-run of the identical tree.

```
thread 'api::iov_integration::test_iov_occasion_dsl_matches_column_fit' panicked at
src/api/tests/iov_integration.rs:696:5:
assertion `left == right` failed: DSL-occasion OFV 149.3110436111533 must be bit-identical
to column OFV 149.31104361805458 (same partition + deterministic pipeline)
  left: 4639457431879869841
 right: 4639457431879627024
```

The two OFVs differ by 6.9e-9 absolute, 4.6e-11 relative — a handful of ULP, not a partition
error.

## Reproduction so far

| run | command | result |
|---|---|---|
| full workspace suite | `cargo test --no-default-features --features ci,survival --no-fail-fast` | **FAILED** (4775 passed, 1 failed) |
| the test alone | `cargo test --lib --features ci,survival test_iov_occasion_dsl_matches_column_fit` | ok, 2.98 s |
| full `--lib`, same tree, same features | `cargo test --lib --no-default-features --features ci,survival` | ok — 3988 passed, 0 failed, 228 s |

So it is **intermittent and load-dependent**, not deterministic. I have one failure and two
passes; I have not yet bisected it or established a rate.

## Why this needs investigating rather than loosening

The assertion was deliberately tightened from `< 1e-6` to `to_bits()` equality in #917, after
an investigation that concluded the OFV path is order-stable and thread-count-independent:

- `foce_population_nll{,_iov}` collect into an ordered `Vec` and then sum **serially** (#703).
- `run_inner_loop_warm` uses an order-preserving collect; `find_ebe` is serial per subject and
  keys on occasion *structure*, not label values.
- Both fits share one process-wide pool, so a thread-count shift cancels in the difference.
- Empirically at the time: 145 fits, zero divergence, including 120/120 under a 12-way CPU hog.

That PR added a "do **not** loosen" comment for exactly this situation. A 0-vs-0 bit check
cannot flake under load *if the pipeline is deterministic* — so a failure is evidence that
something has stopped being deterministic since #917 (2026-07-22), not evidence that the
[...]

PR #1232, opened 2026-09-04
test(survival): NONMEM anchor for a lagged mid-record SS dose × ODE-accumulated hazard [#1220]

### LINKED ISSUE #1210: bug(ode): an SS dose carries the equilibration run-in hazard into H(0), so every joint PK-TTE survival quantity is wrong

TeunP:
## Why

#1220 item 1. `a_mid_record_ss_dose_lagged_past_the_interval_keeps_the_hazard_accrued_so_far`
(Tier-1) is the **sole** holder of `restore_chz` inside `ss_state_at_phase` — mutation `M8`
kills it and nothing else — but it is checked against a closed form, not against NONMEM.

The two neighbouring behaviours were already anchored externally and neither covers this one:

* *preserve across a second `SS=1` dose* — `ss_chz_r3_{const,drug}` (#1210);
* *the SS-plus-lag PK convention* — #1121's `ss_lagtime_edge_nonmem_anchor.rs`.

What had no external anchor was the **intersection**: a mid-record `SS=1` record whose lag
exceeds `II`, with a live drug-driven hazard on both sides of it. Items 2–4 of #1220 need no
work (2 was #1218, shipped; 3 is "no action" by its own text; 4 stays parked).

## What changed

Test-only. `src/` is untouched.

A new NONMEM construction `r5_drug_lag` and one new arm in `tests/ss_chz_nonmem_anchor.rs`.

**Only the drug arm exists.** With `BETA = 0` the hazard is `H0` regardless of the PK, so
`H(t) = H0·t` holds whatever the lag does — the const arm is structurally blind here, which is
why this is a one-arm test rather than the two-arm loop the r3 test uses.

**The twin is not a shifted train, and that is the whole construction.** For `lag >= II` ferx
clamps the seed phase to zero (`ss_seed_phase` is `(II - lag).max(0)`), and NONMEM 7.6.0 does
the same (`results/ss_lag_ge_ii.tab`). An `SS=1` record with `lag >= II` therefore means
*periodic pulses ending with one on the record itself*, then this dose's own pulse at
`t + lag` — no shifted train reproduces it. So `ss_chz_r5_drug_lag.ctl` carries **no `ALAG` at
all**: it is a plain pulse train placed at the ferx side's lagged *arrivals*, gated at
`T >= 600`, with ferx `t` = NONMEM `T - 600` exactly as r2/r3.

`ALAG1 = 24` (a multiple of `II = 12`) puts the first record's own arrival back on the `II`
grid while still exercising the clamp arm. The pulse that would land at ferx `t = 12` exists on
neither side — the record that would arrive there precedes the first record — so the train omits
`T = 612` too, and the mid-record `SS=1` sits far enough out (`t = 240`) that the gap has decayed
below tolerance (`e^{-0.1·228} ≈ 1e-10`).

The lag is a **fixed theta**, not a column: `ode_cumhaz_hazard` snapshots the PK parameters at
`t = 0`, so a per-dose lag column would be read at its baseline for every dose, and a
time-varying one would trip `assert_survival_tv_covariates`.

New files:

| file | content |
|---|---|
| `nonmem_anchor/ss_chz_r5_drug_lag.ctl` | `ss_chz_r3_drug.ctl` against `train5.csv`; no `ALAG` anywhere |
| `nonmem_anchor/ss_chz_train5.csv` | 73 doses: `0…600` step 12, **no 612**, `624…828` step 12, `840, 864, 876, 888`; PK obs at `612, 839.9, 840.1, 852, 863.9, 864.1, 876`; TTE row at `876` |
| `nonmem_anchor/ss_chz_drug_lag_fit.ferx` | `ss_chz_drug_fit.ferx` + `theta TVLAG(24.0, FIX)` and `ALAG1 = TVLAG` |
| `nonmem_anchor/ss_chz_ss2_lag.csv` | `SS=1 II=12 @ 0`; plain `12…204`; **nothing at 216 or 228**; `SS=1 @ 240`; plain `252, 264` |
| `nonmem_anchor/results/ss_chz_r5_drug_lag.{lst,tab}` | the run |

## Alternatives considered

The first draft of this construction (issue comment 1) built the NONMEM side as a *shifted*
lagged train with `ALAG1 = 14`. Measured, that is off by 2.4× at `t = 49.9` (11.49 vs 4.837) —
the clamp above is why, and no choice of shift fixes it. Recorded on the issue rather than
re-derived here.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

NONMEM 7.6.0 via the `nonmemdocker:V0.1` image (`nmfe76`), run directory
`~/NONMEMruns/ferx-1210-ss-chz/r5_drug_lag/`. The `.lst` carries no `PRDERR`, `+INF`,
`NUMERICAL DIFFICULTIES` or `PROGRAM TERMINATED` — the check that r1 exists to make non-optional.

```
                 ferx                                NONMEM (FORMAT=s1PE15.8)
 t        PRED        H            h          IPRED         CHZ           HAZ
 12     4.788962   26.118512   0.219250    4.78896241  2.61185123e1  2.19250185e-1
239.9   4.837086  467.636336   0.224590    4.83708578  4.67636338e2  2.24589696e-1
240.1   5.688122  467.686330   0.343708    5.68812187  4.67686332e2  3.43708257e-1
252     4.788962  493.777039   0.219250    4.78896240  4.93777041e2  2.19250185e-1
263.9   1.456925  494.862874   0.041438    1.45692496  4.94862877e2  4.14378517e-2
264.1   2.374881  494.872267   0.065574    2.37488054  4.94872269e2  6.55735585e-2
276     3.780985  503.024446   0.132453    3.78098515  5.03024448e2  1.32452600e-1

worst relative disagreement:  H/h 5.51e-9    PRED 3.55e-9
```

**The bound is measured, not chosen.** Both worst cases sit on the reference's own printf
floor — nine significant digits on `A(3) ≈ 5e2` is worth ~1e-9 by itself — so `2e-8` is ~3.6×
the largest, the same bound and the same reasoning as the r3 arm above it.

The fixture was checked against the issue's pass-2 ferx table *before* the NONMEM run was
spent: all 21 values (PRED/H/h × 7 times) reproduced it exactly.

## Performance
- [x] No significant impact expected — one more anchor arm, ~0.5 s.

## Tests

One new arm, `a_lagged_mid_record_ss_dose_keeps_the_hazard_and_matches_nonmem`, asserted in
this order:

1. **The reference certifies itself first** — it must carry `H = 467.6`, `h = 0.22` into the SS
   record (or "preserve" and "zero" agree there and the arm proves nothing), and its own
   increment across the record must be the uninterrupted train's `≈ 0.05`.
2. **IPRED before H** — if the PK disagrees the fixture is not a twin of the train and the
   hazard comparison downstream is meaningless. This is the control that caught r3's first
   draft (`PRED = 0.132` against the train's `4.837`).
3. `H` and `h` on all seven records, each `is_finite()` **before** it is folded — `f64::max`
   returns the other operand for a `NaN`, so an unchecked fold would keep whatever the finite
   records produced and pass on their strength. Checked on **both** sides: a self-review caught
   that the first version guarded only ferx's values, and `"NaN".parse::<f64>()` *succeeds* in
   Rust — so a `NaN` cell in the reference table would neither shift the parsed columns nor
   trip the grid check, it would just make the quotient `NaN` and vanish into the fold. The
   table's `IPRED`/`CHZ`/`HAZ` are now asserted finite too (a follow-up commit), matching what
   `the_constant_arm_of_the_reference_reproduces_the_closed_form` already did.

### Mutation evidence (all three observed, on this branch)

| mutation | expected | observed |
|---|---|---|
| drop `restore_chz` from `ss_state_at_phase` (`M8`) | this arm red, with the Tier-1 | **red**: `worst_h = 0.99994`; Tier-1 red (`H is non-monotone`); the other 9 anchor arms green |
| drop `restore_chz` from `equilibrate_ss_state` instead | this arm **green** | **green**; `a_second_ss_dose_...` red instead (`const: 0.99792`) |
| zero the accumulator instead of preserving (the fix #1210's issue originally suggested) | red by `H(240)` | **red**: `worst_h = 0.99994`, and r3 red too |

The middle row is the interesting one: it is the external confirmation of the Tier-1 test's own
doc claim — that `lag >= II` clamps the phase and keeps `equilibrate_ss_state` out of this path.
Before this PR that claim rested on a closed form.

- [x] Regression test added that fails without the behaviour it pins

- [x] Not applicable (test-only)

## Docs
- [x] No user-visible change — no `CHANGELOG.md` entry (test-only, per the repo's rule)

## Checklist
- [x] `cargo test --lib --tests --no-default-features --features ci,survival` — 170 binaries,
      5010 passed, 0 failed, 170 ignored. Reported honestly: the *first* run of this suite had
      `api::iov_integration::test_iov_occasion_dsl_matches_column_fit` fail its bit-identity
      assertion (`149.38899871045703` vs `149.31104361805458`); it passed on the re-run. That is
      the known intermittent IOV flake (#1197), and this branch does not touch `src/` at all.
- [x] `src/` untouched — `git status` shows only `tests/` and `nonmem_anchor/`

## Docs & examples
- [x] No new DSL syntax, `[fit_options]` key, example `.ferx` or data file. The new `.ferx` and
      `.csv` are anchor fixtures under `nonmem_anchor/`, not user-facing examples, so nothing to
      mirror to ferx-site / ferx-book and no example execution step.

## Reviewer hints

The part worth checking adversarially is the **dose bookkeeping**, not the assertions: whether
`ss_chz_ss2_lag.csv`'s records and `ss_chz_train5.csv`'s pulses really describe the same
arrival sequence. They are two spellings of `0 (seed peak), 24, 36, …, 228, [SS re-seed at 240],
264, 276, 288`, and the two deliberate holes — no ferx record at 216 or 228, no NONMEM pulse at
612 — are where a double-dose or a missing dose would hide. The IPRED comparison is what would
catch it, which is why it is asserted before the hazard.

## Open questions

What ferx does with an explicit dose record whose lagged arrival lands *past* a later `SS`
record (a ferx record at 216, arriving 240) is a #1121-family question NONMEM cannot answer for
this model — its native SS on this system returns `+INF` (`ss_chz_r1_*`). Deliberately outside
this anchor; if it turns out to double-dose, that is its own issue.


TeunP:
## Self-review, xhigh (10 angles + gap sweep) — 6 findings

Reviewed `origin/main...HEAD` at `2b561a43`. The anchor itself holds — NONMEM agrees to 5.5e-9
and the three mutations behave as predicted. The findings are one overclaim in the doc comment,
one assertion of mine that cannot fail, one unenforced provenance check, and three smaller items.

---

### 1. `tests/ss_chz_nonmem_anchor.rs` — correctness (doc) — `ALAG1 = 24` cannot distinguish clamping from wrapping

The doc says `ALAG1 = 24` is a multiple of `II = 12` "so the first record's own arrival lands back
on the `II` grid **while still exercising the clamp arm** (`lag > II`)". The code path is indeed
entered, but the *value* it produces is not discriminating:

* ferx clamps: `ss_seed_phase = (II - lag).max(0)` = `(12 - 24).max(0)` = **0**
* a wrapping implementation: `(II - lag % II) % II` = `(12 - 0) % 12` = **0**

Both conventions give phase 0 whenever `lag` is an exact multiple of `II`. So this anchor confirms
the branch is taken but **cannot tell the two conventions apart** — a mutation swapping clamp for
wrap would leave it green. The discrimination lives in #1121's `ss_lag_ge_ii` anchor, which uses
`LG = 15` with `II = 12` (clamp → `PRED(0) = 13.197`, the post-pulse peak; wrap → `2.1815`).

That is fine — the phase convention is #1121's job and `restore_chz` is this arm's job — but the
comment currently reads as if this arm pins the convention. It should say what it actually pins
and point at #1121 for the rest.

### 2. `tests/ss_chz_nonmem_anchor.rs` — test-coverage — `compared_pred` is now a tautology

I copied the counter from the r3 arm but changed the lookup:

```rust
let pr = pred.iter().find(...).unwrap_or_else(|| panic!("no ferx PRED row at t={t}"));
...
compared_pred += 1;                     // unconditional
...
assert_eq!(compared_pred, GRID_R5.len());
```

In r3 the increment sits inside `if let Some(pr) = ...`, so the count genuinely catches a PRED row
that was silently skipped. Here the `unwrap_or_else` already panics on a missing row, the loop runs
`reference.len()` times, and `nonmem_rows_r5` has *already* asserted `reference.len() == 7`. So
`compared_pred == 7` always holds and the assertion can never fail. It is exactly the
"tests that cannot fail" shape, in an assertion I added. Either drop it or restore the conditional
form that made it meaningful.

### 3. `nonmem_anchor/results/ss_chz_r5_drug_lag.lst` — test-coverage — the "clean run" check is remembered, not enforced

The `.lst` is committed for provenance and the run recipe says to check it for
`PRDERR` / `+INF` / `PROGRAM TERMINATED` before trusting the table — that is how `r1` announced
itself. I did that check by hand; nothing in the repo does it. A cheap assertion over the committed
`.lst` would make the claim enforced, and would apply to r1–r4's tables too. This is in the same
spirit as `the_constant_arm_of_the_reference_reproduces_the_closed_form`: certify the reference
before comparing anything against it.

### 4. `tests/ss_chz_nonmem_anchor.rs` — test-coverage — the lag regime is not asserted on the fixture

Nothing in the test states that `ALAG1 > II` for the model it loads. If
`ss_chz_drug_lag_fit.ferx` lost its `ALAG1 = TVLAG` line, ferx would produce an unlagged
trajectory and the IPRED comparison would fail loudly — so the arm is not *wrong* — but it would
fail as a numerical disagreement rather than naming the cause. An explicit assertion that the
loaded model's lag exceeds the dataset's `II` documents the regime the arm exists for.

### 5. `nonmem_anchor/ss_chz_ss2_lag.csv` / `ss_chz_train5.csv` — reviewer note — the two deliberate holes are the fragile part

Not a defect, but the thing a future reader will get wrong: the ferx side has **no record at 216 or
228** and the NONMEM side has **no pulse at `T = 612`**, and both omissions are load-bearing (216's
arrival at 240 would double the seed's own pulse; 612 has no counterpart on the ferx side because
the record that would arrive there precedes the first record). The IPRED control is what catches a
mistake in either. This is called out in the PR body; it belongs in the fixture files' own comments
too, since that is where someone will edit them.

### 6. `tests/ss_chz_nonmem_anchor.rs` — conventions — no `docs/` or `CHANGELOG` entry

Correct per `CLAUDE.md` ("A PR that only touches internal refactors, tests, or CI does not need an
entry") and `src/` is untouched — recorded here only so the omission reads as deliberate.

---

### What held up under the correctness angles

* **The oracle is genuinely independent.** The NONMEM twin carries no `SS` record and no `ALAG`, so
  none of ferx's steady-state or lag logic is mirrored on the reference side — NONMEM just
  integrates a plain pulse train. A wrong arrival placement on the ferx side would show up in
  IPRED, which is asserted before `H`.
* **The reference self-checks run before the comparison** (`H = 467.6`, `h = 0.22` into the record;
  increment `≈ 0.05` across it), and they interrogate the *table*, not ferx — the ferx
  discrimination is the tolerance assertion at the end. Verified by mutation: `M8` fails at the
  tolerance line with `worst_h = 0.99994`, not at the straddle.
* **The bound is measured, not chosen**: 5.5e-9 on H/h and 3.5e-9 on PRED, both at the reference's
  own nine-digit printf floor; `2e-8` is ~3.6× the largest, matching r3.
* **The fold is NaN-safe on both sides** after `2b561a43` — a `NaN` in the reference table would
  otherwise vanish into `worst.max(NaN)`, since `"NaN".parse::<f64>()` succeeds in Rust and would
  neither shift the parsed columns nor trip the grid check.

PR #1233, opened 2026-09-04
feat(api): shared candidate runner — parallel fits, canonical-hash dedup, journal + resume [closes #1178]

### LINKED ISSUE #1178: feat(tools): shared candidate runner — parallel fits, canonical-hash dedup, journal + resume [#1175]
Part of #1175 (P0 — prerequisites). Depends on #1115 and on the `ferx-core::edit` layer.

## Problem

Every search tool in the epic needs the same orchestration: fit N candidates in parallel, deduplicate identical candidates, cache results, journal them so an interrupted search resumes, and emit a per-candidate table. `bootstrap/` already contains all of that (`journal.rs`, `manifest.rs`, `output.rs`, `summary.rs`) for one specific shape — resampled datasets rather than edited models.

Writing it a second time per tool would be four copies.

## Proposal

A `ferx_tools::search::runner` module, lifted from the `bootstrap/` pattern:

```rust
pub struct Candidate { pub id: String, pub model: ModelText, pub parent: Option<String>, pub features: FeatureVector }
pub struct CandidateResult { pub id: String, pub fit: Option<FitResult>, pub verdict: StrictnessVerdict, pub criterion: f64, pub seconds: f64 }

pub struct Runner { /* thread budget, cache dir, journal, cancel flag */ }
impl Runner {
    pub fn run(&self, candidates: Vec<Candidate>, data: &Population, opts: &RunOptions) -> Vec<CandidateResult>;
}
```

- **Dedup and cache** key on `ModelText::canonical_hash`, so a candidate reached twice by two different step orders is fitted once. This is pyDarwin's "model uniqueness" for free.
- **Journal + resume** reuse the bootstrap journal format, so an interrupted multi-hour search restarts from the finished candidates.
- **Retries** are `n_starts` per candidate, applied before the strictness gate.
- **Thread budget** must not oversubscribe: candidates are embarrassingly parallel but each `fit()` already uses rayon internally, which is exactly what #1115 exists to control. Nesting them naively is a real regression, not a theoretical one.

## Acceptance

- [ ] Two candidates rendering to the same canonical text are fitted once.
- [ ] A run interrupted partway resumes and refits nothing already journalled — same test shape as #1143.
- [ ] Thread budget is respected: a run with `threads = 2` never exceeds 2 concurrent fits, asserted rather than assumed.
- [ ] Candidates failing strictness appear in the output table with their reasons, not dropped silently.
- [ ] Cancellation via `CancelFlag` stops a search between candidates and returns partial results.

roninsightrx:
## Why

Part of #1175 (P0 — prerequisites). Depends on #1115 (`PoolPlan`) and the `ferx-core::edit` layer (#1176), both merged.

Every search tool the epic plans — covsearch, modelsearch, iivsearch, ruvsearch — is *generate candidates → fit them → rank them*, and the middle verb is identical each time: fit N candidates in parallel, deduplicate the ones that are the same model, cache and journal the results so an interrupted overnight search resumes, and emit a per-candidate table. `bootstrap/` already contains all of that (`journal.rs`, `manifest.rs`, `output.rs`) for one specific shape — resampled datasets rather than edited models. Writing it a second time per tool would be four copies of the same orchestration, each with its own resume bug.

## What changed

A new `ferx_tools::search` module — `Runner`, plus the types it consumes and produces. `ferx-core` is untouched.

**Identity is the canonical hash.** `Candidate::hash()` is the hex `ModelText::canonical_hash`, so a model reached twice by two different step orders is fitted **once** — pyDarwin's "model uniqueness" for free — and that same key addresses the journal row and the cached fit. A duplicate is not dropped: it comes back in the results carrying the representative's criterion and verdict plus `duplicate_of`, because a candidate missing from a report cannot be told apart from one the generator never produced.

**Parallelism goes through `PoolPlan::from_budget`** (#1115) rather than a bare outer `par_iter`. `fit()` already `par_iter`s over subjects, so nesting naively makes the two levels compete for the same workers; the plan splits the budget explicitly and gives the outer pool `FIT_RAYON_STACK_SIZE` instead of Rayon's 2 MiB default, and `apply_to` pins the inner fit to what is left.

**Retries before the gate.** Each candidate is fitted with `RunOptions::n_starts` (default 3) and only then judged by `check_strictness`. A candidate that fails the gate, fails to compile or fails to fit is kept with its reasons and lands in `candidates.csv`.

**Journal and resume.** With a cache directory:

```
.ferx-search/
  search_run.json          # criterion / n_starts / strictness / dataset fingerprint
  search_journal.jsonl     # one JSON object per finished candidate, flushed per row
  fits/<hash>.json         # the FitResult, so a resumed run is not a degraded one
  candidates.csv           # the table, rewritten in candidate order at the end
```

The journal is rewritten from its intact rows on open (through a `.part` sibling renamed into place), so a line truncated by a hard kill costs exactly one refit and the next append lands on a line boundary. `search_run.json` refuses a resume whose criterion, starts, strictness or dataset differ, naming the field — reusing a row means trusting a number another process computed under one convention.

**Cancellation** checks between candidates (and before the journal is opened, so a cancelled run never truncates the previous run's recovery data), returning partial results with `RunReport::cancelled`.

```rust
let report = Runner::new()
    .threads(8)
    .cache_dir(".ferx-search")
    .run(&candidates, &population, &RunOptions::default())?;
report.best()   // lowest criterion among the candidates that passed the gate
```

## Alternatives considered

- **Journal as CSV, like `bootstrap/raw_results.csv`.** A candidate's outcome is a verdict with a variable-length list of reasons, which is not a CSV row. Split instead: JSONL is the recovery log, CSV is the human-readable table.
- **Storing the `FitResult` inline in each journal row.** Keeps one file, but makes every line megabytes and the log ungreppable. The fit is a *cache* beside the log; a missing or corrupt `fits/<hash>.json` costs the reused candidate its `FitResult`, never its criterion and verdict.
- **Cloning the representative's `FitResult` into every duplicate.** One object per duplicate for a result the caller can already reach by id; `duplicate_of` says where it is.
- **Returning `Result<Vec<CandidateResult>, _>` as the issue sketched.** A cancelled run must return what finished, which an `Err(Cancelled)` cannot carry — hence `RunReport { results, cancelled, fitted, reused, deduped }`.

## Cross-repo dependency

No `ferx-core` API change, so nothing for `ferx-r` to pick up.

## Breaking changes
- [x] None

## Numerical validation
- [x] Compared against: prior ferx commit (a direct `fit()` call), bit-for-bit

The Tier-2 oracle is **degenerate**: a candidate that is the base model unedited must give the OFV `fit()` gives on the same model and population, to the bit —

```rust
assert_eq!(base.ofv.to_bits(), direct.ofv.to_bits());
```

Anything the runner's own compile path drops — a covariate binding, the file's `[fit_options]`, the initial estimates — moves that number, and nothing else in the module would notice, since there is no second copy of the compile path to disagree with it. The other side of the oracle is asserted too: dropping `ETA_KA` must change both the hash *and* the objective, so an edit that never reached the engine fails rather than passing quietly.

This PR adds no numerical kernel — the criterion is `ferx_core::bic` / `FitResult::{ofv, aic}`, all validated in #1177 — so there is no new formula to anchor against NONMEM.

## Performance

New code, and its whole point is not to oversubscribe: `threads = 2` is asserted to never exceed 2 concurrent fits.

## Tests

**21 Tier-1 orchestration tests** drive a private `run_with_fitter` seam with an injected fitter. That is not a convenience: "two fits are in flight at once" is not observable from outside `fit()`, so the concurrency bound could not be *asserted* against real fits, only assumed — and every one of these tests would otherwise be a slow-tier convergence run. The one thing a fake fitter cannot fake is a `FitResult` (130-odd fields, no constructor), so the fixture is one genuine `outer_maxiter = 0` evaluation of `examples/warfarin.ferx`, memoized and cloned.

**8 Tier-2 tests** (`tests/search_runner_end_to_end.rs`) cover the real path — `ModelText` → `parse_full_model` → the data binders → `fit()` — with every fit an evaluation, so the file returns in 0.6 s.

**Three mutations were run to check the assertions can actually fail**, per the rule in CLAUDE.md:

| Mutation | Result |
|---|---|
| dedup disabled (`duplicate_of` never set) | `two_candidates_with_the_same_canonical_text_are_fitted_once` red — `left: ["first", "second"]` |
| thread budget ignored (`from_budget(0, …)`) | `the_thread_budget_bounds_concurrent_fits` red |
| resume filter removed | `a_resumed_run_refits_nothing_already_journalled` red — `left: ["a", "bb", "ccc"]` |

The middle one **passed the first time**, and fixing that is the most interesting part of the diff. The gauge originally left as soon as a second caller appeared, so every worker spent microseconds inside it and even an unbounded pool rarely had three fits in flight at once — `max <= 2` held under a deliberately broken budget. The gauge now holds a fixed dwell, which is what makes a wider pool actually reach a wider `max`; the reason is written into its doc comment so it cannot be "simplified" back.

A second finding, in the fixtures rather than the product: a Rayon worker blocked on a nested `install` keeps stealing work, so it re-entered `OnceLock::get_or_init` *in the middle of its own initialisation* and deadlocked. Observed as a hung test, not theorised. The fixture now initialises on a joined `std::thread`, which is not a Rayon latch, so the waiting worker steals nothing. `compile_and_fit` has no such lock, so this is a test-only hazard — but it is the kind that comes back, hence the comment.

No CLI flag, no `.ferx` syntax, no new `FitResult` field — this is the library layer the actual search tools will sit on.

## Docs
- [x] No user-visible change

Deliberately no `docs/` page: there is nothing a user can run yet. The `docs/tools/` page lands with the first search tool that has a CLI, alongside the `*.ferxsearch` configuration the epic specifies.

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), rustdoc, docs lint, public-API baseline
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, no gradient-reachable code touched
- [x] `Cargo.toml` version bump — N/A, additive

`cargo test -p ferx-tools`: 242 pass, 0 fail.

### ferx-site / ferx-book
No new DSL syntax, `[fit_options]` key, example `.ferx` or data file — nothing to mirror.

### Example execution
- [x] No example execution step needed (internal / no user-visible change)

## Reviewer hints

Worth the attention:

- **`runner.rs`, the assembly loop after the parallel section.** Results are emitted in submission order, and three cases have to come out right: a representative that was fitted, one that was reused from the journal, and a duplicate that points at either. A candidate whose representative was skipped by cancellation is omitted rather than emitted half-filled.
- **`journal.rs`, `CandidateRecord::{criterion, ofv}` being `Option<f64>`.** JSON has no `NaN`: `serde_json` writes a non-finite float as `null` and then refuses to read it back as `f64`, so a failed candidate stored as a bare `f64` would be a row that cannot be resumed from.
- **`SearchManifest::check_compatible`** — is the set of fields it guards the right one? The claim is that criterion, starts, strictness and dataset are exactly what a journalled row's meaning depends on.
- **`compile_and_fit` clones the `Population` per candidate**, because `bind_theta_levels` takes `&mut Population` to synthesize the level-index column (#1064). Cheap next to a fit, and a shared mutable population across concurrent candidates would be a data race by construction — but flag it if there is a better shape.

Skimmable: `output.rs` (a CSV writer), `candidate.rs` (data types and builders).

## Open questions

1. **`fit_options: Option<FitOptions>` in `RunOptions`.** `None` lets each candidate carry its own `[fit_options]`, which is what a search over models edited from one base file wants. `Some` replaces them wholesale, for a caller that needs one configuration across a space whose members disagree. Is wholesale replacement the right semantics, or should it be a merge?
2. **`RunOptions::n_starts` defaults to 3**, per the epic's "every candidate gets `n_starts ≥ 3`, non-negotiable". That is 3× the cost of a naive search by default. Keep it, or default to 1 and let the search tools set it?
3. **`journal` is a `pub mod`** so a search tool can read a finished run's rows without re-running it. If that is more surface than wanted, it can drop to `pub(crate)` with `read_records` re-exported.

https://claude.ai/code/session_01REh3nwZwzixaoboE7rFvXz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,14 @@ section of the SDLC for the versioning policy).
   or failed the gate — appears in `candidates.csv` with its reason rather than being dropped.
   The thread budget splits across both levels of parallelism via `PoolPlan` (#1115) instead of
   nesting Rayon pools, and a `CancelFlag` stops a search between candidates and returns the
-  partial results. This is the orchestration layer the covariate, structural, variability and
+  partial results — into `candidates.partial.csv`, so cancelling a resumed run never overwrites
+  the complete table of the run it resumed. The cache directory is claimed for the length of a
+  run (`search.lock`) because two runs sharing one would silently destroy each other's journal,
+  and a directory that cannot be written costs the resume and the report rather than the fits:
+  those failures come back on `RunReport::warnings` with the results intact. A candidate whose
+  fit *errored* is refitted on the next resume instead of being remembered as permanently
+  unfittable, since a transient resource limit and a broken model are the same string from
+  there. This is the orchestration layer the covariate, structural, variability and
   residual-error searches of #1175 are built on.
 - **BIC variants and a `Strictness` gate for candidate ranking (#1177, part of #1175).**
   `ferx_core::bic(&result, BicType::{Mixed, Iiv, Random, Fixed})` computes the four
@@ -225,7 +232,9 @@ section of the SDLC for the versioning policy).
   `P = TVP * exp(ETA_P)` form and is a hard error naming the parameter on anything else, never
   a silent wrong edit. `ModelText::canonical_hash` gives a candidate a stable identity — equal
   across comment and whitespace changes, different for every semantic one — for use as a fit
-  cache key. This is the prerequisite for the model-search tooling in #1175.
+  cache key; a `#` or `//` inside a quoted value is content, not a comment, so two candidates
+  differing only in a quoted path (`"s3://bucket/a.csv"` vs `.../b.csv`) are two candidates.
+  This is the prerequisite for the model-search tooling in #1175.
 - **A cancellable bootstrap (#1161).** `BootstrapOptions::cancel` takes a `CancelFlag`; setting
   it from another thread stops a long `ferx_tools::bootstrap` run at the next replicate boundary
   and returns the new `BootstrapError::Cancelled`, so a caller reports an abort as an abort
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,11 +65,14 @@ section of the SDLC for the versioning policy).
   the complete table of the run it resumed. The cache directory is claimed for the length of a
   run (`search.lock`) because two runs sharing one would silently destroy each other's journal,
   and a directory that cannot be written costs the resume and the report rather than the fits:
-  those failures come back on `RunReport::warnings` with the results intact. A candidate whose
-  fit *errored* is refitted on the next resume instead of being remembered as permanently
-  unfittable, since a transient resource limit and a broken model are the same string from
-  there. This is the orchestration layer the covariate, structural, variability and
-  residual-error searches of #1175 are built on.
+  those failures come back on `RunReport::warnings` with the results intact. A lock whose owner
+  was hard-killed is taken over automatically, so resuming after a kill — the case the journal
+  exists for — never needs a file deleted by hand. A candidate that produced no fit carries a
+  `CandidateError` saying whether the failure was the *model* (it does not compile: remembered,
+  and reported without refitting) or the *run* (a fit pool that could not be built: refitted on
+  the next resume), so one bad minute cannot permanently mark a fittable model as unfittable.
+  This is the orchestration layer the covariate, structural, variability and residual-error
+  searches of #1175 are built on.
 - **BIC variants and a `Strictness` gate for candidate ranking (#1177, part of #1175).**
   `ferx_core::bic(&result, BicType::{Mixed, Iiv, Random, Fixed})` computes the four
   conventions of `pharmpy.modeling.calculate_bic` from a finished `FitResult` — the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,19 @@ section of the SDLC for the versioning policy).
   of one-cycle integrations back.
 
 ### Added
+- **A shared candidate runner for model-space search (#1178, part of #1175).**
+  `ferx_tools::search::Runner` fits a list of candidate models in parallel and returns each
+  one scored: the ranking criterion (`ofv`, `aic`, or any of the four BIC variants), the
+  `Strictness` verdict with the reasons for every gate it failed, and the fit itself.
+  Candidates are identified by the canonical hash of their model text, so a model reached
+  twice by two different edit paths is fitted **once**; with a cache directory each outcome is
+  journalled as it finishes, so an interrupted overnight search resumes and refits only what
+  is missing, and every candidate — including the ones that failed to compile, failed to fit
+  or failed the gate — appears in `candidates.csv` with its reason rather than being dropped.
+  The thread budget splits across both levels of parallelism via `PoolPlan` (#1115) instead of
+  nesting Rayon pools, and a `CancelFlag` stops a search between candidates and returns the
+  partial results. This is the orchestration layer the covariate, structural, variability and
+  residual-error searches of #1175 are built on.
 - **BIC variants and a `Strictness` gate for candidate ranking (#1177, part of #1175).**
   `ferx_core::bic(&result, BicType::{Mixed, Iiv, Random, Fixed})` computes the four
   conventions of `pharmpy.modeling.calculate_bic` from a finished `FitResult` — the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,14 @@ section of the SDLC for the versioning policy).
   marked fixed, which is what it meant at the time.
 
 ### Fixed
+- **Numbers written to JSON now reload as themselves, bit for bit (#1178).** `serde_json`
+  parses floats with a fast algorithm accurate only to within 1 ULP unless its
+  `float_roundtrip` feature is enabled, which it now is across `ferx-core`, `ferx-tools` and
+  the CLI. Anything that writes a number and reads it back — a `.fitrx` bundle, a search
+  journal, a cached fit — could otherwise return an estimate one bit from the one that was
+  computed, and do it invisibly, since every printed form rounds long before that digit. The
+  case that caught it was a resumed candidate search reporting a criterion of
+  `-200.28784144636057` for a fit that scored `-200.28784144636055`.
 - **`n_parameters`, AIC and BIC no longer count the structural zeros of a mixed
   `block_omega` + diagonal `omega` as estimated parameters (#1177).** The cross-block
   Cholesky entries of such an Ω are pinned, never searched, and the covariance step already
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -319,6 +319,7 @@ version = "0.3.1"
 dependencies = [
  "csv",
  "ferx-core",
+ "libc",
  "nalgebra",
  "rand 0.10.2",
  "rayon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,7 +106,11 @@ slow-tests = []
 nalgebra = "0.35"
 csv = "1.3"
 serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+# `float_roundtrip`: without it `serde_json` parses floats to within 1 ULP
+# rather than exactly, so a `.fitrx` bundle — or any JSON a checkpoint or cache
+# reads back — can return an estimate one bit from the one that was written.
+# Every consumer here is numeric output that is meant to reload as itself.
+serde_json = { version = "1", features = ["float_roundtrip"] }
 zip = { version = "8", default-features = false, features = ["deflate"] }
 rayon = "1.10"
 rand = "0.10"
--- a/crates/ferx-cli/Cargo.toml
+++ b/crates/ferx-cli/Cargo.toml
@@ -28,7 +28,9 @@ path = "src/main.rs"
 ferx-core = { path = "../..", version = "0.3.1", default-features = false }
 ferx-tools = { path = "../ferx-tools", version = "0.3.1" }
 indicatif = "0.18.4"
-serde_json = "1"
+# Matches the root package and `ferx-tools`: see the note there — floats must
+# reload as themselves, not to within 1 ULP.
+serde_json = { version = "1", features = ["float_roundtrip"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/crates/ferx-tools/Cargo.toml
+++ b/crates/ferx-tools/Cargo.toml
@@ -28,6 +28,14 @@ rayon = "1.10"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 
+# `search.lock` staleness only: `kill(pid, 0)` is how a search directory whose
+# owner was hard-killed is taken over instead of needing a manual delete —
+# resuming after a kill is the case the journal exists for. Already in the
+# workspace lockfile (via `tempfile` and `rand`), so this adds no crate to the
+# build; non-unix targets compile a `cfg` fallback and never take a lock over.
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [features]
 # Heavy convergence tests (5–60 min wall). Default `cargo test` ignores them;
 # a nightly/manual job opts in with --features slow-tests.
--- a/crates/ferx-tools/Cargo.toml
+++ b/crates/ferx-tools/Cargo.toml
@@ -26,7 +26,14 @@ rayon = "1.10"
 # (#1143). Both are already in the workspace's lockfile via `ferx-core`, so this
 # adds no crate to the build.
 serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+# `float_roundtrip` is load-bearing, not tidiness. Without it `serde_json`
+# parses floats with a fast algorithm accurate only to within 1 ULP, so a
+# journalled criterion or OFV can come back one bit different from the one the
+# interrupted run computed — and a resumed run whose whole promise is "the same
+# numbers" would report a different score, invisibly, since the CSV prints six
+# decimals. Caught by `search_runner_end_to_end`: -200.28784144636055 read back
+# as -200.28784144636057.
+serde_json = { version = "1", features = ["float_roundtrip"] }
 
 # `search.lock` staleness only: `kill(pid, 0)` is how a search directory whose
 # owner was hard-killed is taken over instead of needing a manual delete —
--- a/crates/ferx-tools/src/lib.rs
+++ b/crates/ferx-tools/src/lib.rs
@@ -54,3 +54,4 @@ mod tests {
 
 pub mod bootstrap;
 pub mod gam;
+pub mod search;
--- a/crates/ferx-tools/src/search/candidate.rs
+++ b/crates/ferx-tools/src/search/candidate.rs
@@ -246,6 +246,63 @@ impl Default for RunOptions {
     }
 }
 
+/// Why a candidate produced no fit, and whether a later run should try again.
+///
+/// The flag is not cosmetic: a journalled outcome is what every subsequent
+/// `resume: true` run believes without rechecking, and the two failures that
+/// arrive here are not the same statement. A model that does not compile will
+/// not compile on the next machine either — remembering that saves the parse
+/// and, more importantly, keeps the candidate in the report. A fit that died
+/// because `install_on_fit_pool` could not build a pool (`api/pool.rs`) says
+/// nothing about the model at all, and writing *that* down as final would let
+/// one bad minute mark a fittable model dead for the rest of the search's life.
+///
+/// From inside the runner the two are one `String`, which is why the
+/// classification is made where the failure is raised rather than recovered
+/// from the message afterwards.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CandidateError {
+    /// What went wrong, as shown in the report.
+    pub message: String,
+    /// `true` when the failure describes the run rather than the candidate, so
+    /// a resumed run should fit it again instead of trusting the row.
+    pub retryable: bool,
+}
+
+impl CandidateError {
+    /// A failure that is a property of the candidate itself — a model that does
+    /// not compile, or does not bind against this run's data. Deterministic, so
+    /// a resume reuses it.
+    pub fn model(message: impl Into<String>) -> Self {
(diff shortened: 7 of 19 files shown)


roninsightrx:
Review findings — **changes requested**:

1. **[P1] Fingerprint the dataset contents, not only its shape** (`crates/ferx-tools/src/search/journal.rs:337-349`). The current fingerprint contains only `dv_column`, subject IDs, and per-subject observation counts. Changing DV values, observation times, doses, compartments, censoring, occasions, or covariates leaves the manifest unchanged. A resumed run then finds every candidate hash in `reused`, performs no fit, and returns criteria/FitResults computed from the old data. This is a silent wrong-result path. Please hash every fit-relevant population field (with deterministic ordering for maps), or require a trustworthy full data hash from the caller. A regression test should mutate an observation or dose while preserving IDs/counts and assert that resume is refused.

2. **[P1] Include caller-supplied fit settings in resume compatibility** (`crates/ferx-tools/src/search/journal.rs:247-270`). `compile_and_fit` uses `RunOptions::fit_options` wholesale, but `SearchManifest` records only criterion, starts, strictness, and data. Thus a run can be resumed with a different method, max iteration limit, optimizer, tolerances, RNG seed, covariance configuration, etc.; compatibility succeeds and the old score/fit is reused. The candidate hash does not protect this case because the override is outside `ModelText`. Please record a fingerprint/description of all effective fit options that can affect the fit, excluding only intentional runtime-only fields such as thread count, verbosity, and cancellation, and add a resume regression test.

3. **[P1] Synchronize the BLOQ override onto the compiled model** (`crates/ferx-tools/src/search/runner.rs:507-524`). Parsing stamps the candidate file value into `parsed.model.bloq_method`, while the likelihood reads that model field. When `fit_options: Some(...)` replaces the file options, this path updates only `gradient_method`; it never performs the `model.bloq_method = opts.bloq_method` assignment used by `fit_from_files`. Consequently an override from `drop` to `m3` (or the reverse) reports the requested options but evaluates the wrong censored-data likelihood and can rank candidates incorrectly. Please mirror `base.bloq_method` onto the model and cover both override directions with a censored-data test.

Verification: `cargo test -p ferx-tools` passed locally (242 passed; 2 intentionally ignored). The overall architecture and existing orchestration coverage look solid once these cache/input invariants are closed.

GitHub would not accept this as a formal request-changes review because the authenticated account owns the PR, so I am recording the blocking review as a comment.

roninsightrx:
All three P1s fixed in 72101eec.

**1. Fingerprint the dataset contents, not its shape** — `data_fingerprint` now hashes every field of every subject. Two things make that trustworthy rather than approximately so: the `Subject` is destructured **exhaustively, with no `..`**, so a field added upstream stops this file compiling instead of silently going unhashed; and every covariate map is written in key order, because hashing `HashMap`'s iteration order would produce a fresh digest per process and refuse every resume. `Population::warnings` / `exclusions` stay out deliberately — they describe what the reader did on the way to the population, and `exclusions` is already applied to `subjects`.

Pinned by `the_fingerprint_sees_the_contents_and_not_only_the_shape`: seven mutations (DV value, observation time, dose amount, dose compartment, observation compartment, censoring flag, covariate value), each asserted to leave `n_subjects` / `n_observations` untouched first — otherwise the case would pass on the old shape check and test nothing — and each asserted to be *refused* by `check_compatible`, not merely to move the digest. A companion test pins stability across repeated construction, which is the opposite failure mode. `a_resume_against_the_same_shaped_but_edited_dataset_is_refused` covers the same thing through `Runner` itself.

**2. Fit settings in the resume check** — `SearchManifest::fit_options_fingerprint` is a SHA-256 of the effective `FitOptions` as the runner applies them (`quiet()` included), with only this run's scheduling normalised away: `threads` (the `PoolPlan` share), `cancel`, `n_starts` (already its own manifest field), verbosity, and `user_set_keys` (order-dependent bookkeeping, so including it would refuse resumes that are in fact identical). `None` means "each candidate's own `[fit_options]`" and is incompatible with any override, including the default one.

`FitOptions` has no `Hash`/`Serialize` and its `Debug` holds no maps, so the rendering is both complete and deterministic — a field added to it lands in the digest automatically. Six settings asserted to move it (method, outer iteration cap, covariance step, BLOQ convention, gradient method, multi-start seed), five scheduling knobs asserted not to, plus `a_resume_under_different_fit_settings_is_refused` end-to-end through `Runner` — which also asserts the *same* settings still resume, so the gate is not simply refusing everything with an override.

**3. BLOQ sync onto the model** — `compile_and_fit` now does `parsed.model.bloq_method = base.bloq_method`, next to the `gradient_method` stamp and with a comment naming why these two keys are special (they reach the engine through the model, not the options).

`a_bloq_override_reaches_the_censored_likelihood_in_both_directions` runs both directions on `examples/warfarin_bloq.ferx` + `data/warfarin_bloq.csv` — file `m3` overridden to `drop`, and the same model edited to `drop` overridden to `m3` — against a direct fit with the convention stamped on the model, bit-for-bit. It asserts up front that the fixture actually carries censored rows and that the two conventions disagree on it, so it cannot pass vacuously. Reverting the one-line fix fails it loudly: the fit scores **OFV -120.69** where the requested convention gives **1025.40**.

Each fix was reverted in turn to confirm the new tests go red; all three do. `tools/preflight.sh` green, `cargo test -p ferx-tools` 250 pass.

One incidental change while in there: the resume-refusal assertions now go through a small `expect_refusal` helper. `expect_err` on a `Result<RunReport, _>` dumps every `FitResult` the report holds into the panic message — the mutation runs printed megabytes around a one-line failure.

roninsightrx:
## Code review — `ferx_tools::search` (candidate / journal / output / runner)

Reviewed the full diff against the surrounding `ferx-core` code. Nothing blocking in the core orchestration; eight findings below, grouped.

### Correctness / data loss

**1. `runner.rs:310` — a journal write failure discards the whole in-memory report.**
`if let Some(j) = journal { j.into_result()?; }` propagates on the first journal error, so the `RunReport` built from every completed fit is dropped *and* `output::write_table` is skipped. An overnight 500-candidate search that hits ENOSPC on candidate 3's `fits/<hash>.json` returns a `String` and nothing else. The journal is documented as a best-effort cache — better surfaced as a warning on the report (or after the table is written) than as a run-fatal error.

**2. `runner.rs:362` — `write_table` runs on a cancelled run, and `csv::Writer::from_path` truncates.**
A resumed run cancelled after two candidates overwrites a previously *complete* `candidates.csv` with a two-row file. The journal survives, but the human-readable report the module promises is destroyed. Write to a `.part` and swap only on a non-cancelled run.

**3. `runner.rs:280` — transient fit failures are journalled as permanent.**
`install_on_fit_pool` returns `Err` when `build_fit_pool` hits a resource limit (`src/api/pool.rs:278`), so a machine briefly out of threads bakes `error: "…"` + `passed: false` into `search_journal.jsonl`. Every subsequent `resume: true` run reads that row, refits nothing, and reports the model as unfittable. Only cancellation-time errors are excluded (line 279). Either don't journal error rows, or mark them so the resume path can retry.

**4. `runner.rs:448` / `output.rs:67` — `reused_result` drops `record.ofv` and `record.converged`.**
The CSV then reads those two columns off `r.fit`. When the cached `fits/<hash>.json` is missing or corrupt — exactly the degraded-resume case the journal doc says it handles — the row comes back with an empty `ofv` and empty `converged` even though the journal holds both. `CandidateRecord` stores them for this purpose.

**5. `candidate.rs:138` — `Candidate::hash()` inherits a quote-unaware comment stripper.**
`src/edit/canonical.rs::squeeze()` deliberately copies quoted strings byte-for-byte, with a doc comment saying that collapsing them "would have the fit cache answer one with the other's result" — but `code_of` (`src/edit/mod.rs:445`) runs first and cuts each line at the first `#` **or** `//` with no quote tracking. Two candidates differing only inside a quoted value past a `//` (e.g. a `[data]` path `"s3://bucket/a.csv"` vs `"s3://bucket/b.csv"`) canonicalise to the same text, so the runner fits one and reports its criterion and verdict for the other under `duplicate_of`. Pre-existing in the `edit` layer (#1176), but this PR is what turns it from cosmetic into a wrong-score path.

### Performance

**6. `runner.rs:201` — the manifest is built even when there is no cache directory.**
`SearchManifest::new(options, data)` is unconditional but only used inside the two `cache_dir.is_some()` branches (lines 203, 230). `data_fingerprint` now `Debug`-formats *every field of every subject* — doses, obs times, observations, per-event covariate snapshots, `obs_records` — into one growing `String` before hashing it. A `Runner::new().run(...)` with no cache dir on a 100k-record population allocates a multi-hundred-MB transient string and SHA-256s it for a value that is immediately dropped. Move the call inside the match.

**7. `runner.rs:501` — `data.clone()` per candidate, per worker.**
`fit()` itself is careful here: `src/api/fit.rs:567` uses a `Cow<Population>` and clones only when pruning / IOV / LTBS actually need it. So a 16-wide plan holds up to 16 runner clones *plus* up to 16 `fit()` clones of the same dataset. `bind_theta_levels` / `bind_covariate_stats` only mutate for models that declare a level block or a symbolic covariate stat, so a `Cow` removes the common-case copy. (This is the reviewer hint in the PR description.)

### Concurrency

**8. `journal.rs:121` — a fixed `.part` name plus an unconditional truncate makes two runs on one cache dir silently corrupt each other.**
`store_fit` writes through the sibling `fits/<hash>.json.part`, and `Journal::create` unconditionally truncates `search_journal.jsonl`. Two runs pointed at the same cache directory race on that one temp name — `std::fs::write` from both, then `rename` — yielding a fit file that is a byte-mix of two writes; and the second run's `Journal::create` wipes the first's log while it is still appending. A lock file, or a pid/nonce-suffixed `.part`, would make this fail loudly instead.

### Verified sound

Dedup-by-first-occurrence, the resume filter, the submission-order assembly (including the "representative skipped by cancellation ⇒ omit the duplicate" case), `best()`'s tie-break (`min_by` returns the first minimum), and the `Option<f64>` journal columns all check out. Commit 2's `bloq_method` stamp and the widened `fit_options` / dataset fingerprints close the gaps the first commit left.

roninsightrx:
All eight findings addressed in `ba97405f`.

| # | Finding | Fix |
|---|---|---|
| 1 | journal write failure discarded the report | `RunReport::warnings`; the run survives and the table is still written |
| 2 | `write_table` truncated a complete table on cancel | partial rows go to `candidates.partial.csv` (removed by a completed run); the table is written through a `.part` + rename |
| 3 | transient fit errors journalled as permanent | error rows are not reused — filtered *before* the hash dedup, or the row would consume the hash and hide the successful refit appended after it |
| 4 | `reused_result` dropped `ofv` / `converged` | both now live on `CandidateResult` beside `fit`, so a lost `fits/<hash>.json` no longer blanks two columns the journal holds |
| 5 | quote-unaware `code_of` collided two candidates | a `#` or `//` inside a quoted string is content; an unterminated quote never cuts |
| 6 | manifest built for cacheless runs | `SearchManifest::new` moved behind `cache_dir` |
| 7 | `Population` cloned per candidate per worker | `Cow`, behind the same emptiness predicate `bind_theta_levels` early-returns on |
| 8 | fixed `.part` name + unconditional truncate | `search.lock` claims the directory for the length of a run (RAII, released on every exit path); `.part` names carry a pid + counter |

Every fix carries a regression test (+262 lines in `runner_tests.rs`, +75 in `output_tests.rs`, +36 in `edit/mod_tests.rs`). The `code_of` one was mutation-checked: reverting to the old stripper makes `canonical_hash_keeps_a_comment_marker_inside_a_quoted_value` fail with two identical digests.

Two notes on the choices, since both trade something:

- **The lock is a plain `O_EXCL` file, not an OS advisory lock** — no new dependency, and the failure mode is visible. The cost is that a hard kill leaves it behind, so the error names the file and says to delete it. Given the flagship use case is *resume after a kill*, that is the sharp edge; an `flock`-based version would auto-release, at the price of a dep.
- **Error rows are refitted rather than marked retryable.** A model that genuinely does not compile is therefore re-parsed on every resume — cheap. The alternative needed a taxonomy of failure the runner cannot see from a `String`.

Full suite green: 4015 `ferx-core` lib tests, 218 `ferx-tools` lib + integration, `tools/preflight.sh` clean (public-API baseline unchanged — `code_of` is `pub(crate)`).

roninsightrx:
The two trade-offs I flagged above are now paid off in `143715a3` — neither is left standing.

**Stale `search.lock` no longer needs a manual delete.** Requiring one broke the flagship case (resume after a hard kill) to guard against an operator mistake. The holder's pid goes into the lock file, and a lock whose owner is no longer running is taken over automatically — on unix via `kill(pid, 0)`, where `ESRCH` (and only `ESRCH`) means gone and `EPERM` is a live process we may not signal.

Everything the check cannot positively establish falls on the refusing side — an empty file, a garbled one, a non-unix target — because takeover is the only path that deletes another run's lock. The failure direction is therefore "refused a run that could have proceeded", never "two runs in one directory"; a recycled pid reads as live and refuses, and the message says which file to delete. `libc` is already in the workspace lockfile via `tempfile` and `rand`, so the `cfg(unix)` dependency adds no crate to the build.

**Error rows are now classified instead of uniformly refitted.** "Never reuse an error row" was blunt because from inside the runner a broken model and a fit pool that could not be built are one `String` — but they are not one statement, and the distinction is available where the failure is *raised*. A parse or bind failure is a property of the candidate and this run's data, both pinned by the resume manifest; anything out of `fit()` describes the machine. The fitter now returns a `CandidateError { message, retryable }`, the journal row records it, and a resume refits the second kind while reporting the first straight from the row. `CandidateRecord::retryable` defaults to `true`, so a journal written before the field existed refits its failures rather than believing them.

Worth stating for anyone weighing the cost: the model-shaped bad outcome of a search candidate — a fit that finishes without converging — is not an error at all. It comes back `Ok` and is judged by the strictness gate, so it was always journalled and reused. An `Err` out of `fit()` is the pathological case, which is what makes refitting it cheap.

`candidates.csv` gains a `retryable` column, empty for a row with no failure to describe.

Both predicates are mutation-checked rather than assumed:

- stubbing the liveness check to always-alive fails `a_lock_left_behind_by_a_dead_process_is_taken_over`;
- the resume pair straddles its predicate — dropping *every* error row gives `["broken", "flaky"]` and dropping *none* gives `[]`, against an expected `["flaky"]`, so a one-sided fixture could not have caught either.

221 `ferx-tools` tests + integration green, `tools/preflight.sh` clean, public-API baseline unchanged.

PR #1237, opened 2026-09-05
feat(tools): `.ferxsearch` config and an MFL search-space parser with @-symbol resolution [closes #1179]

### LINKED ISSUE #1179: feat(tools): `.ferxsearch` config and an MFL search-space parser with @-symbol resolution [#1175]
Part of #1175 (P0 — prerequisites).

## Problem

The search needs a way to state *what to search over*. `docs/model-file/covariate-model.qmd` already fixes where it must not go:

> What PsN calls `[test_relations]`, `valid_states`, `included_relations` and `p_forward` is **search configuration, not model**.

So this is a config file consumed by `ferx-tools`, not a new `.ferx` block.

## Proposal

A `*.ferxsearch` TOML file whose `[space]` key carries an **MFL string** — reuse Pharmpy's Model Feature Language grammar verbatim so users transfer directly, and so a search space is quotable in an issue or a paper. Do **not** reuse pyDarwin's template + tokens file pair; that machinery exists to compensate for not having a structured model object.

```toml
base = "warfarin.ferx"
data = "warfarin.csv"

[space]
mfl = """
ABSORPTION([FO,ZO]); PERIPHERALS(0..1); LAGTIME([OFF,ON])
COVARIATE?(@IIV, @CONTINUOUS, [pow,lin]); COVARIATE?(@IIV, @CATEGORICAL, cat)
"""

[rank]
type   = "bic"      # ofv | aic | bic | bic_mixed | bic_iiv | penalized
cutoff = 3.84

[strictness]
require_converged    = true
max_condition_number = 1000.0
max_correlation      = 0.95
reject_on_boundary   = true

[run]
threads   = 8
retries   = 3
cache_dir = ".ferx-search"
```

Two pieces of work:

1. **MFL parser** — feature categories (`ABSORPTION`, `ELIMINATION`, `PERIPHERALS`, `TRANSITS`, `LAGTIME`, `COVARIATE`, `IIV`, `IOV`, `COVARIANCE`), the `?` exploratory suffix, arrays `[a,b]`, ranges `0..2`, wildcard `*`, and `LET(SYMBOL, [...])`.
2. **Symbol resolver** — `@IIV`, `@PK`, `@CONTINUOUS`, `@CATEGORICAL`, `@PK_IIV`, `@ABSORPTION`, `@ELIMINATION`, `@DISTRIBUTION` resolved against the base model's `[covariates]` and `[individual_parameters]` blocks plus the dataset.

## Coverage gaps to reject loudly

A parsed MFL feature ferx cannot express must be a **hard error naming the feature**, never a silently narrowed search:

- `ELIMINATION(ZO / MM / MIX-FO-MM)` — reachable only via `[odes]`, no analytic template
- `ABSORPTION(SEQ-ZO-FO)` — no template
- `DIRECTEFFECT` / `EFFECTCOMP` / `INDIRECTEFFECT` — no PD template family

## Acceptance

- [ ] Round-trip: parse an MFL string, render it, reparse — same feature set.
- [ ] Each `@`-symbol has a resolver test against a model + dataset fixture.
- [ ] An unsupported feature errors naming the feature and why, with a link to the coverage table in the docs.
- [ ] `docs/tools/search.qmd` documents the config format and the coverage table.

roninsightrx:
## Why

Part of #1175 (P0 — prerequisites). Every search tool in the epic is *generate candidate → fit → rank*, and needs a way to state **what to search over**. `docs/model-file/covariate-model.qmd` already rules that out of the model file:

> What PsN calls `[test_relations]`, `valid_states`, `included_relations` and `p_forward` is **search configuration, not model**.

So this adds the `.ferxsearch` TOML file consumed by `ferx-tools`, whose `[space]` carries a Pharmpy **Model Feature Language** string — reused verbatim so Pharmpy users transfer directly and a space is quotable in an issue. Not pyDarwin's template + tokens pair (that machinery compensates for not having a structured model object, which ferx has since #1176).

Closes #1179.

## What changed

All in `crates/ferx-tools/src/search/`; **no `ferx-core` change**, public-API baseline untouched.

- **`mfl.rs`** — lexer, recursive-descent parser and canonical renderer for the full MFL grammar: `ABSORPTION`, `ELIMINATION`, `PERIPHERALS`, `TRANSITS`, `LAGTIME`, `COVARIATE`, `ALLOMETRY`, `IIV`, `IOV`, `COVARIANCE`, `DIRECTEFFECT`, `EFFECTCOMP`, `INDIRECTEFFECT`, `METABOLITE`, `LET`; the `?` exploratory suffix, arrays `[a,b]`, ranges `0..2`, `*`, `@SYMBOL`. Every category is parsed whether or not ferx can build it, so `ELIMINATION(MM)` is refused as *not expressible*, not as an unknown keyword.
- **`resolve.rs`** — `ModelContext` from `ParsedModel` + `ModelText` + `Population`; the eleven built-in symbols (`@IIV`, `@PK`, `@PK_IIV`, `@ABSORPTION`, `@ELIMINATION`, `@DISTRIBUTION`, `@BIOAVAIL`, `@CONTINUOUS`, `@CATEGORICAL`, `@PD`, `@PD_IIV`) resolved against `[individual_parameters]`, the `pk NAME(role=VAR)` line, `[covariates]` and the dataset's columns. `LET` overrides a built-in. Pharmpy's `expand()` rules ported: an explicit `(param, cov)` pair beats the symbol-derived one, a forced term beats the identical optional one, a pair forced by two statements is an error. Kind checks (`pow` on a categorical covariate) fail here, not fifty candidates later. An empty symbol drops the feature with a returned note (Pharmpy's behaviour).
- **`coverage.rs`** — the hard-error table. Wildcards are checked against their full expansion, so `ELIMINATION(*)` is not an opt-out. The error lists every gap at once and ends with a link to the docs table.
- **`config.rs`** — the `.ferxsearch` loader. `base`/`data` resolved relative to the file; `[rank]` → `Criterion`, `[strictness]` overlays `Strictness::default()`, `[run]` → `Runner` + `RunOptions`. Unknown sections (`[covsearch]`, …) pass through as raw `toml::Table` for the tool that owns them; typos *inside* known sections and stray top-level scalars are errors. `type = "penalized"` is parsed (format stability) but refused at load until P6.
- New dependency: `toml = "1"` on `ferx-tools` only.

### Anchored against Pharmpy 2.0.0 (second commit)

`tools/pharmpy-mfl-anchor/` feeds a 75-line corpus to Pharmpy's own `pharmpy.tools.mfl.parse.parse` and `expand()`s 15 symbol-bearing statements on a reference model (basic oral + 1 peripheral, WT/CRCL continuous, SEX categorical) inside the local `pharma_ai_r` container. The dump is committed at `crates/ferx-tools/tests/data/mfl_pharmpy_anchor.json` and replayed by a Tier-1 module, `search/mfl_pharmpy_anchor.rs`: every corpus line either agrees after normalisation, or is a divergence **named in the test with its reason**, and the test asserts a named divergence really differs (a stale entry is red). CI replays the JSON; the container is only needed to regenerate.

Two Pharmpy rules were **adopted** because of the anchor: a mandatory `COVARIATE` may not use `*` for its effects ("Mandatory effects need to be explicit"), and `TRANSITS(N)` takes no depot option.

Divergences kept on purpose (full table in `tools/pharmpy-mfl-anchor/README.md`): backwards ranges (Pharmpy: silently empty), the optional `ALLOMETRY` reference (Pharmpy: `IndexError`), parameter/covariate wildcards (Pharmpy: `TypeError` on input its grammar allows), `#` comments / empty statements, the `IIV`/`IOV`/`COVARIANCE` categories (in Pharmpy's docs, not its 2.0.0 grammar), unknown `@FOO` (Pharmpy: silent `()`), `@PD` (Pharmpy: `ValueError`), the covariate-kind check, and Pharmpy dropping a **whole parameter** from a symbol statement when one explicit pair overrides it (`CL-CRCL` vanishes; ferx drops only the pair).

### Deviations from Pharmpy (documented on the page)

1. Parameter / covariate names **keep their case** — ferx's `[individual_parameters]` and CSV headers are case-sensitive.
2. Bare `TRANSITS(n)` means **NODEPOT** — the analytic `*_transit` templates feed central from the last transit; an explicit `DEPOT` is a coverage error, not a reinterpretation.
3. `COVARIATE(…, *)` on a declared **categorical** covariate is `cat` (Pharmpy blindly expands to the four continuous forms).

## Alternatives considered

- A new `.ferx` block — rejected by the docs rule above.
- Deriving covariate kinds from the dataset (few distinct values ⇒ categorical) — rejected; `@CONTINUOUS`/`@CATEGORICAL` require the `[covariates]` block and say so, because a numerically coded categorical is indistinguishable from an integer continuous column.
- Making `CompiledModel` expose the `pk` role bindings — would widen the core API; the `pk` line is read from `ModelText` instead, no new `pub` item.
- Following Pharmpy's whole-parameter removal on explicit override — rejected; it is a representational limitation of `parameter × covariate` statements, and a user writing `COVARIATE?(@PK, @CONTINUOUS, *)` plus `COVARIATE(CL, WT, exp)` would not expect `CL-CRCL` to vanish. Pinned as a named divergence so the choice stays visible.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable (config parsing / symbol resolution — no fits). The external reference for this feature is Pharmpy's parser; see the anchor section above.

## Performance

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes) — 135 new, in `mfl_tests.rs`, `coverage_tests.rs`, `resolve_tests.rs`, `config_tests.rs`, `mfl_pharmpy_anchor.rs`:
  - round-trip: parse → render → reparse is an identity on every category, and render is a fixed point;
  - one resolver test per `@`-symbol against an in-memory model + population fixture, plus the no-`pk`-line (ODE) and no-`[covariates]`-block cases;
  - every coverage gap named in the issue, every remaining table row, wildcard expansion, error text and docs link;
  - config: every `[rank]` type, strictness overlay, relative paths, typo rejection, tool-section passthrough, and two end-to-end runs on shipped examples (`two_cpt_oral_cov` + the new `.ferxsearch`);
  - Pharmpy anchor: 75 parse lines, the built-in symbol sets, and 15 `expand()` cases replayed against the committed Pharmpy 2.0.0 dump.

- [x] Example added to `examples/` — `examples/two_cpt_oral_cov.ferxsearch`, pinned by `the_shipped_example_file_loads_and_resolves`.

<details>
<summary>Coverage error, as loaded</summary>

```
[space] mfl: the search space asks for 2 features ferx cannot build a candidate for:
  - ELIMINATION(MM): only first-order elimination has an analytic `pk` template; ...
  - IOV(...): inter-occasion variability is expressible in a model (`[iov]`) but not yet
    as a searchable edit — `ferx-core::edit` has no add/drop-IOV operation (#1175 P4)
See the coverage table at https://ferx-nlme.github.io/ferx-core/tools/search.html#coverage
```

</details>

## Docs
- [x] `docs/` updated — new page `docs/tools/search.qmd`: file format, grammar, symbol table, coverage table, `[rank]`/`[strictness]`/`[run]`, and the Pharmpy anchor
- [x] New page linked in `docs/_quarto.yml`

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), rustdoc, docs-lint, public-API baseline
- [x] No `Dual2` path touched

### ferx-site
- [ ] Not applicable — no `.ferx` DSL change; the `.ferxsearch` format is documented in this repo's docs page

### ferx-book
- [ ] No chapter yet for search tools; follows with the first tool (P1 covsearch)

### Example execution
- [x] No example execution step needed — the new example is a config file; it is loaded and resolved by a unit test, no fit

## Reviewer hints

- `resolve.rs::merge_covariate_terms` is the port of Pharmpy's `ModelFeatures.expand()`; the three rules are documented on the fn, each has a unit test, and the anchor replays Pharmpy's actual output for them.
- `mfl_pharmpy_anchor.rs::ferx_json` is the *only* place the documented deviations are normalised away (case, mode de-duplication, Pharmpy's parse-time defaults `DRUG`/`DEPOT`/`NODEPOT`); everything else must match Pharmpy exactly or be a named divergence.
- `coverage.rs` is the table in `docs/tools/search.qmd` — if you disagree with a row (e.g. `ABSORPTION(ZO)`/`WEIBULL` marked ✓ per the epic's table), that is a one-line change in both places.
- `ModelContext::from_model` reads the `pk` line from `ModelText`, not `CompiledModel`, precisely to avoid a new `pub` item in core.

## Open questions

- `TRANSITS(n)` defaulting to NODEPOT is a judgement call; the alternative (Pharmpy's DEPOT default) would make every bare `TRANSITS` a coverage error.
- No CLI subcommand yet — nothing consumes the config until P1 lands. Intentional for this issue's scope.


roninsightrx:
Two covariance cases diverge from the Pharmpy grammar/semantics that this format promises to reuse:

1. **[P1] Resolve the covariance parameter wildcard as `@IIV`, not `@PK`** (`crates/ferx-tools/src/search/resolve.rs:434`, via `resolve_parameters` at line 560).

   `COVARIANCE(IIV,*)` has feature-specific wildcard semantics in MFL: the second `*` means the IIV parameter set. Here it goes through the generic parameter resolver, where every wildcard becomes `@PK`. On the test fixture this therefore grounds to `COVARIANCE(IIV,[CL,V1,Q,V2,KA,F])` instead of the expected IIV-bearing `[CL,V1,KA]`. The extra `Q`, `V2`, and `F` have no eta, so the resulting search space asks a later variability search to build invalid covariances. Please make wildcard resolution context-sensitive for covariance and add a regression test for `COVARIANCE(IIV,*)`.

2. **[P2] Accept covariance level arrays and wildcards** (`crates/ferx-tools/src/search/mfl.rs:1168`).

   The MFL grammar defines the first covariance argument as `values | wildcard`, but this branch accepts only one `Arg::Word`. Consequently valid MFL such as `COVARIANCE([IIV,IOV],[CL,V])` and `COVARIANCE(*,[CL,V])` fail as syntax errors. In this implementation the IOV expansions should subsequently reach the coverage checker and produce the promised unsupported-feature error; rejecting them in the parser breaks both the verbatim-transfer and hard-error guarantees. The AST likely needs `Modes<VariabilityLevel>` here, with coverage/resolution iterating the expanded levels.

The focused `cargo test -p ferx-tools --lib` suite passes (285 tests), and the PR head remained `d0ba385ab8abcc83291a013dc7c9aa2dcf0a9474` while reviewed.

roninsightrx:
Both addressed in the follow-up commit.

**P1 — `COVARIANCE(IIV,*)` resolved as `@PK`.** Confirmed: on the fixture it grounded to `[CL,V1,Q,V2,KA,F]`, three of which have no η. `resolve_parameters` now takes the built-in the wildcard stands for, and each feature says which: `@PK` for `COVARIATE` and `IIV` (any parameter can gain an η), `@IIV` for `IOV` and `COVARIANCE` (only an η-bearing parameter can carry a κ or sit in an omega block — Pharmpy's `iovsearch` adds κ on the IIV set). Regression tests in `resolve_tests.rs::wildcards_expand_by_slot`: `COVARIANCE(IIV,*)` → `[CL,V1,KA]`, `IOV?(*,EXP)` → `[CL,V1,KA]`, `COVARIANCE(*,*)` → `COVARIANCE([IIV,IOV],[CL,V1,KA])`. The docs page's wildcard row now states the per-slot rule.

**P2 — covariance level only a bare word.** `Feature::Covariance.level` is now `Modes<VariabilityLevel>`, parsed with the same `modes_of` as every other mode slot, so `COVARIANCE([IIV,IOV],[CL,V])` and `COVARIANCE(*,[CL,V])` parse, round-trip, and reach the coverage table, where the IOV member is the named gap (`coverage_tests.rs::the_remaining_gap_rows`). Resolution grounds the level to its expansion. An unknown level now gets the standard mode error (`RUV` is not a known variability level; known: IIV, IOV).

135 search tests green; fmt / clippy / rustdoc / docs-lint green.

roninsightrx:
Second review pass (high effort), on the tree after the follow-up commit. Ten findings, ranked by severity. The first five are correctness gaps in coverage / resolution that contradict the "hard error at load, never a silently narrowed search" rule the module states for itself; the rest are robustness and config-surface issues.

### 1. `ABSORPTION(ZO)` / `ABSORPTION(WEIBULL)` pass coverage but have no `pk` template — `crates/ferx-tools/src/search/coverage.rs:94`

`PkModel::from_name` (`src/types.rs:2207-2216`) knows only `iv` / `oral` / `transit` / `ig`; `weibull` and `zero_order` are ODE-only input functions (`ODE_ONLY_ABSORPTION_FNS` in `model_parser.rs`), the same class as `SEQ-ZO-FO`, which coverage *does* refuse. So `[space] mfl = "ABSORPTION([FO,ZO])"` (or `ABSORPTION(*)`) loads cleanly and the structural search carries a ZO / WEIBULL arm that `ModelEdit::SetStructural` cannot build — it fails per candidate instead of at load. `docs/tools/search.qmd` also marks both ✓. Add ZO / WEIBULL gaps (or derive the table from `PkModel`) and fix the docs row.

### 2. Explicit-pair override diverges from Pharmpy 2.0.0 — `crates/ferx-tools/src/search/resolve.rs:352`

Pharmpy drops a symbol-produced pair only when the statement's *parameter* operand is a `Ref` (`covariate.py:60`: `if explicit_covariates and isinstance(self.parameter, Ref)`). Here `explicit = !parameters.is_symbolic() && !covariates.is_symbolic()` makes a statement with an explicit parameter and a symbolic covariate overridable too. Verified in the anchor container:

```
COVARIATE?(CL, @CONTINUOUS, [pow,lin]); COVARIATE(CL, WT, exp)
```

Pharmpy keeps CL-WT `pow?`, CL-WT `lin?` and the forced CL-WT `exp`; `merge_covariate_terms` (line 654) silently drops the two optional CL-WT forms. The docs say "Pharmpy's override rules apply", and the anchor's expand cases only put symbols in the parameter slot, so the anchor cannot see this. Gate on `parameters.is_symbolic()` alone and add an anchor case with a symbol in the covariate slot.

### 3. `ode_template NAME(...)` bases get `template = None` — `crates/ferx-tools/src/search/resolve.rs:157`

`ModelContext::from_model` only recognises a `pk NAME(...)` line. `ode_template NAME(role=VAR, ...)` uses the same role grammar (`model_parser.rs:9459`, `docs/model-file/absorption.qmd:1156`) but yields no template, so `@PK`, `@ABSORPTION` / `@ELIMINATION` / `@DISTRIBUTION` / `@BIOAVAIL` and the `COVARIATE(*, …)` wildcard all hard-error with a message claiming the model has no template:

```
[structural_model] ode_template two_cpt_oral(cl=CL, v1=V1, q=Q, v2=V2, ka=KA)
COVARIATE?(@PK, @CONTINUOUS, pow)
→ "`@PK` reads the `pk NAME(...)` line, and the base model has none (an `ode(...)` or algebraic model)"
```

A pure covariate / IIV search cannot run on an `ode_template` base. Accept `ode_template` in `PkTemplate::parse_line`, or reuse the parser's bindings.

### 4. Numeric literals on the `pk` line become "unknown parameters" — `crates/ferx-tools/src/search/resolve.rs:580`

`@PK` / `@ABSORPTION` / `@BIOAVAIL` return the raw binding text and `resolve_parameters` then hard-errors on any bound value that is not an `[individual_parameters]` LHS. The parser accepts numeric literals (`f=1`, `model_parser.rs:15609`) and lowercase aliases there:

```
pk one_cpt_oral(cl=CL, v=V, ka=KA, f=1)
COVARIATE?(@PK, @CONTINUOUS, *)      # or IIV(*, EXP)
→ "`1` is not an individual parameter of the base model; it has: CL, V, KA"
```

for a built-in symbol the user never typed. Intersect template variables with `ctx.parameters`, as `@PK_IIV` already does, instead of validating them as user-typed names.

### 5. `individual_parameter_names` text-scan picks up block-form `if` headers — `crates/ferx-tools/src/search/resolve.rs:287`

Every logical line is split on the first `=`, so `if (WT >= 70) {` / `} else if (SEX == 1) {` (documented at `docs/model-file/individual-parameters.qmd:106-117`) become "parameters": `ctx.parameters` = `["if (WT >", "CL", "} else if (SEX", "V"]`. `COVARIATE(FOO, WT, pow)` then reports `it has: if (WT >, CL, } else if (SEX, V`, and the `parameters.is_empty()` guard can be satisfied by a condition alone. `parsed.model.indiv_param_names` (`src/types.rs:3397`, top-level assignments only, source order) is already in hand via `parsed` — use it (filtering `__ferx_*` synthetics) and drop the scan.

### 6. Unbounded `Counts::expand` on user input — `crates/ferx-tools/src/search/mfl.rs:308`

`a..=b` is materialised with no width cap, and `check_coverage` walks it pushing one `Gap` per value with an O(n) `contains` dedup. `[space] mfl = "PERIPHERALS(0..4294967295)"` makes `SearchConfig::load` allocate a 16 GiB `Vec<u32>` and then run a quadratic gap loop — hang / OOM at config-load time instead of an error; even `0..100000000` takes minutes. Bound the range width in `count_of` / `arg`, or check `max > MAX_PERIPHERALS` without iterating.

### 7. Misspelled core sections land silently in `tools` — `crates/ferx-tools/src/search/config.rs:266`

Any unknown top-level *table* is accepted as a tool section, while an unknown scalar key is an error. So

```toml
[strictnes]
require_converged = true
max_condition_number = 1000
```

loads fine, `config.strictness` stays `Default`, non-converged / ill-conditioned candidates pass the gate and nothing warns. Keep an allowlist of tool section names (or require `[tools.*]`) and error on unknown tables.

### 8. `[run] retries` is off by one from Pharmpy and silently coerces 0 — `crates/ferx-tools/src/search/config.rs:306`

Documented as "Pharmpy's retries" but passed 1:1 as `n_starts` (total starts, start 0 = exact inits). A user copying `retries = 3` from a Pharmpy call gets 1 exact + 2 perturbed instead of 1 + 3, and `retries = 0` is coerced to 1 by `.max(1)` where Pharmpy raises. Either map `n_starts = retries + 1` and reject 0, or rename / document the key as total starts.

### 9. `TRANSITS(n)` with depot omitted defaults to NODEPOT; Pharmpy defaults to DEPOT — `crates/ferx-tools/src/search/coverage.rs:143`

Pharmpy's default is DEPOT (`transits.py:14`; the anchor fixture records `TRANSITS(0..3)` → depot `["DEPOT"]`), and the anchor canonicaliser fills the Pharmpy default on the ferx side so it cannot see the flip. A Pharmpy-ported `TRANSITS(1..3)` passes coverage and the search builds the `*_transit` NODEPOT family — a silent reinterpretation on a grammar advertised as "transfers without translation". Treat `depot: None` as DEPOT so it hits the existing gap whose message already says "Write NODEPOT".

### 10. Parameter-slot `*` bypasses the `LET` table — `crates/ferx-tools/src/search/resolve.rs:576`

`*` resolves via `ctx.builtin(wildcard)` and never consults `LET`, contradicting both documented rules (`*` "is `@PK`"; `LET` "overrides a built-in of the same name"). `LET(PK,[CL]); IIV(*,EXP)` grounds to every pk parameter while `IIV(@PK,EXP)` honours the LET, and on an `[odes]` base `LET(PK,[CL,V]); COVARIATE?(*, WT, pow)` fails with "define the symbol with `LET(PK, [...])`" although it is already defined. Route `Operand::Wildcard` through `lookup`, or fix the docs wording and the error text.

roninsightrx:
Findings 1–5 (the correctness gaps) and the CodeFactor B108 hit are addressed in 9188ca10.

- **1 ZO/WEIBULL** — both are coverage gaps now, next to SEQ-ZO-FO; the docs table, the example `.ferxsearch` and the config fixture use `ABSORPTION([INST,FO])`.
- **2 override rule** — gated on `parameters.is_symbolic()` alone, matching `Covariate.eval`. New anchor expand case `COVARIATE?(CL,@CONTINUOUS,[pow,lin]);COVARIATE(CL,WT,exp)` regenerated against Pharmpy 2.0.0 and agrees (5 tuples). The dump now sorts tuples, so the JSON diff is one-time churn from hash order.
- **3 `ode_template`** — `PkTemplate::parse_line` accepts both keywords; the no-template error names both spellings.
- **4 literal bindings** — `@PK` / role symbols intersect the template bindings with the model's parameters (case-insensitively), so `f=1` yields nothing for `@BIOAVAIL` instead of an error.
- **5 block-`if` headers** — `ModelContext` reads `parsed.model.indiv_param_names` minus `__ferx_*` synthetics; the text scan is gone.
- **B108** — `dump.py` writes the reference dataset under `tempfile.mkdtemp`.

Findings 6–10 (range width cap, unknown-table allowlist, `retries` semantics, `TRANSITS` depot default, `*` vs `LET`) are still open.

roninsightrx:
Findings 6–10 addressed in 2f0fc803. Full `tools/preflight.sh` green; 144 search tests pass, anchor regenerated against Pharmpy 2.0.0.

- **6 range width** — every count is bounded at parse by `MAX_COUNT = 64` (`count_of`), so a range's ends and therefore its width are bounded before `Counts::expand` runs. `PERIPHERALS(0..4294967295)` is now a parse error naming the bound. Corpus line `PERIPHERALS(0..100)` records the divergence (Pharmpy has no bound).
- **7 unknown tables** — the loader accepts only `TOOL_SECTIONS` (`covsearch`, `modelsearch`, `iivsearch`, `iovsearch`, `ruvsearch`, `structsearch`, `allometry`); any other table errors with `unknown section [strictnes]; the file takes [space], [rank], [strictness], [run] and the tool sections …`.
- **8 retries** — Pharmpy semantics: `n_starts = retries + 1`, `retries = 0` is a single exact-initials fit (no coercion), default `retries = 2` so the runner's three starts are unchanged. Docs say so in both places.
- **9 TRANSITS default** — an omitted depot option is DEPOT for a count and NODEPOT for `N`, matching Pharmpy's grammar; `TRANSITS(0..3)` now hits the `TRANSITS(n, DEPOT)` gap whose message says to write NODEPOT. Docs bullet and table row corrected.
- **10 `*` vs `LET`** — `Operand::Wildcard` goes through `lookup`, so `LET(PK,[CL]); IIV(*,EXP)` equals `IIV(@PK,EXP)` and the ODE-base error's `LET(PK, …)` advice works for the `*` spelling too.

One naming choice worth a glance: `retries` keeps Pharmpy's meaning rather than being renamed to a total-starts key, since the format's premise is that a Pharmpy space transfers without translation.

Issue #1184, opened 2026-09-01
feat(tools): AMD pipeline — orchestrate the search tools end to end, with CLI and ferx-r surfaces [#1175]


roninsightrx:
Part of #1175 (P5). Depends on #1180, #1181, #1182, #1183.

The orchestrator — Pharmpy's `amd`. Pure sequencing once the individual tools exist.

## Scope

- Default strategy: **structural → IIV → residual → IOV → allometry → covariates → results**.
- Alternative strategies (`reevaluation`, `SIR`, `SRI`, `RSI`) that reorder the same components.
- `retries` policy on final models (`all_final` / `final` / `skip`) — implemented as `n_starts`, already in core.
- Each step seeds the next from the winning fit (`ModelEdit::SeedInits`).
- One report across all steps: selected features, estimates with SEs, and the comparison table per step.

## Surfaces

- [ ] `ferx amd <model> --data <csv> --search <cfg>` CLI subcommand
- [ ] ferx-r wrappers — `ferx_covsearch()`, `ferx_modelsearch()`, `ferx_amd()`. Note the two-step downstream pin: a new `pub` API here is not reachable from ferx-r CI until `ferx-r/src/rust/Cargo.lock` is bumped via `ferx-r/tools/update-ferx-core-lock.sh`.
- [ ] `docs/tools/amd.qmd` and a worked end-to-end example page.

## Reporting requirements

The report is the product here. It must carry, per step and per candidate: the criterion, ΔOFV, the strictness verdict with reasons, the termination status, and the wall-clock cost. A search report that shows only the winner is unauditable — and the failure modes this epic is most exposed to (init stalls, local minima, boundary estimates) are exactly the ones that hide behind a winner-only table.

## Validation

- [ ] End-to-end AMD run compared against Pharmpy `amd` on a shared dataset — same selected features, comparable estimates.
- [ ] Each strategy ordering runs its steps in the stated order, asserted on the journal rather than assumed.
- [ ] `CHANGELOG.md` entry.

PR #1238, opened 2026-09-05
feat(tools): covsearch — SCM forward / forward-then-backward — and allometry [closes #1180]

### LINKED ISSUE #1180: feat(tools): covsearch — stepwise covariate modelling (SCM forward / forward-then-backward) and allometry [#1175]
Part of #1175 (P1). Depends on #1176, #1177, #1178, #1179.

**First shipped search tool** — smallest edit surface, largest user value.

## Why this one first

`[covariate_model]` (#1111) was built for exactly this: one relation per line, no cross-line dependencies, PsN `scm` forms with PsN's default inits and bounds *verbatim*, a `none` state so a rejected relation round-trips through the parser, and the resolved relations returned on `FitResult.covariate_relations` so the search never re-parses a `.ferx` file to learn what ran. Adding or dropping a candidate effect is a line insert or delete against one block; the rest of the file is invariant.

## Scope — covsearch

- `scm-forward` and `scm-forward-then-backward` (Pharmpy's default).
- `p_forward` (default 0.01), `p_backward` (default 0.001), `max_steps`.
- Candidate effects from the `COVARIATE?(...)` entries in the `.ferxsearch` MFL, with `@IIV` / `@CONTINUOUS` / `@CATEGORICAL` resolved (#1179).
- Structural (non-exploratory) `COVARIATE(...)` effects forced in before the search starts.
- **Adaptive scope reduction** (SCM+): stash insignificant effects during the forward phase and re-evaluate them once significant candidates are exhausted.
- Per-step `update_inits` from the winning parent fit, via `ModelEdit::SeedInits`.

## Scope — allometry

Essentially free: allometric scaling is `CL ~ WT power(center = 70, fix = 0.75)` and `V ~ WT power(center = 70, fix = 1.0)`, both already expressible. The tool is the convention (which parameters get which exponent, which reference weight) plus an optional estimated-exponent variant, not new machinery.

## Reporting

A step table — step, candidate effect, ΔOFV, p-value, decision — plus the final relation set and each candidate's strictness verdict. **Convergence status is shown next to ΔOFV, never omitted**: an init-stalled candidate that "loses" a forward step is a selection error, not a result.

## Validation (per CLAUDE.md)

The anchor for a search tool is the *trajectory*, not a kernel:

- [ ] Run PsN `scm` and ferx covsearch on the same dataset with the same test relations; assert the **same forward and backward trajectory** and the **same final relation set**.
- [ ] Per-step ΔOFV checked against NONMEM runs of the same candidate models.
- [ ] **Degenerate oracle**: a search space containing a single point returns the base fit bit-for-bit.
- [ ] A forward step whose winner is an init-stalled fit is caught by the strictness gate rather than selected.

## Deliverables

- [ ] `ferx_tools::covsearch`
- [ ] `ferx covsearch` CLI subcommand
- [ ] `docs/tools/covsearch.qmd`, with the PsN `scm` comparison
- [ ] `CHANGELOG.md` entry

roninsightrx:
## Why

#1175 P1: the first shipped model-space search tool. `[covariate_model]` (#1111) was built so a search could add or drop a relation as one line; the edit layer (#1176), BIC/strictness (#1177), the candidate runner (#1178) and the `.ferxsearch` file (#1179) are all in. What was missing was the search itself: PsN `scm` forward / forward-then-backward, Pharmpy `covsearch` — plus allometry, which is one convention on top of the same machinery.

## What changed

**`ferx_tools::covsearch`** (`crates/ferx-tools/src/covsearch/`)
- `run_covsearch(config, base, run)`: candidates from the `COVARIATE?(...)` statements, forced `COVARIATE(...)` effects into the base model first (kept if the base already has them; an exploratory effect on a pair the base declares is dropped with a note). One `Runner::run` per step under `<dir>/base`, `forward-N`, `adaptive-N`, `backward-N`, so `--resume` walks the same trajectory and refits nothing journalled.
- Forward: lowest OFV among candidates that pass the strictness gate *and* the LRT at `p_forward` (PsN `gof_pval` and Pharmpy `modelrank(lrt)` rule); other forms on the chosen pair retire. Backward: smallest OFV increase whose removal is not significant at `p_backward`; forced/base relations never removable. Adaptive scope reduction (SCM+) and `max_steps` per Pharmpy. Children start from the parent's estimates via `ModelEdit::SeedInits`.
- `Lrt` in `lrt.rs`: `df` read off the fits' `n_parameters` (Pharmpy's `nonfixed` difference), `χ²` tail via `regularized_gamma_q`; a child that adds no free parameter is *unjudgeable* rather than passed through `χ²(0)`.
- Report: `StepRow` per candidate per step with ΔOFV/df/p **beside** `converged`, `passed`, `failures`; `steps.csv`, `final.ferx` (seeded with its own fit), `render_summary`.
- `[covsearch]` section: `algorithm`, `p_forward` (0.01), `p_backward` (0.001), `max_steps`, `adaptive_scope_reduction`. `[rank] type`/`cutoff` are *refused* for covsearch (it ranks by LRT only). To tell "unset" from "bic", `RankConfig::kind` is now `Option<RankType>` with `kind_or_default()`.

**`ferx_tools::allometry`**: `(WT/70)^0.75` on the `cl/q/q2/q3` bindings and `^1.0` on `v/v1/v2/v3`, as `power(center = 70, fix = …)` lines (or `=> THETA_P_WT(init, 0, 2)` with `fixed = false`); base and scaled model fitted side by side. Reads `ALLOMETRY(WT, 70)` + `[allometry]` from a `.ferxsearch`, or flags from the CLI.

**CLI**: `ferx covsearch <search.ferxsearch> [--directory] [--threads] [--resume] [--quiet]`, `ferx allometry <model.ferx|search.ferxsearch> [--data] [--covariate] [--reference] [--parameters] [--exponents] [--estimate] …`. Thin: parse, call, print.

**Core fix** (separate commit, `src/api/postfit.rs`): `theta_boundary_side` judged an identity-packed θ by its position as a fraction of the declared range, so with PsN's default power/exponential bounds `(-100, 1e6)` an estimate of 0.7 read as "pinned to the lower bound" — and the strictness gate excluded **every** candidate of the first real covsearch run. Now judged by distance to each bound on that bound's scale. Regression test added; the log-packed path is unchanged.

## Alternatives considered

- Pharmpy's `nanargmin`-then-test (`lrt.best_of_many`) vs "lowest OFV among the significant": Pharmpy's current `modelrank(lrt)` and PsN both do the latter; that is what shipped.
- Computing `df` from the edit (form → θ count) instead of the fits: rejected — a `FIX`ed θ or a categorical whose level count only the data knows would be miscounted. The fits know.
- Removing base-model relations that overlap the space (Pharmpy's filtered_input_model): rejected as surprising; the base model's relations are kept and the overlap is noted. Documented as a deviation.

## Cross-repo dependency

## Breaking changes
- [x] None (`RankConfig::kind` is `ferx-tools`, not the `ferx-core` public API; the baseline is unchanged)

## Numerical validation

PsN 5.7.1 + NONMEM 7.6 `scm` and ferx covsearch on `data/two_cpt_oral_cov.csv`, two-compartment oral base with η on CL and V1 (FOCEI), test relations CL/V1 × WT/CRCL as `power` (`valid_states = 1,5`), `p = 0.05` both directions. Inputs and `scmlog.txt` under `crates/ferx-tools/tests/psn/scm_anchor/`; replayed by the slow-gated `covsearch_psn_anchor.rs` (passes locally, `--features slow-tests`).

- [x] Compared against: PsN scm / NONMEM 7.6
- [x] Reference values:

```
step  candidate   NONMEM OFV   ferx OFV     ΔOFV   p        PsN      ferx
base  —           -696.79434   -696.794349
1     CL-CRCL     -703.55245   -703.552450  6.758  0.0093   chosen   chosen
1     CL-WT       -703.38740   -703.387404  6.593  0.0102
1     V1-WT       -698.37971   -698.379714  1.585  0.208
1     V1-CRCL     -697.02773   -697.027730  0.233  0.629
2     CL-WT       -709.14359   -709.143614  5.591  0.018    chosen   chosen
2     V1-WT       -705.13444   -705.134463  1.582  0.208
2     V1-CRCL     -703.79309   -703.793118  0.241  0.624
3     V1-WT       -710.72213   -710.722158  1.579  0.209    stop     stop
3     V1-CRCL     -709.38002   -709.380035  0.236  0.627
back  -CL-CRCL    -703.38740   -703.387404  5.756  0.016    kept     kept
back  -CL-WT      -703.55245   -703.552450  5.591  0.018    kept     kept
```

Same trajectory, same final set (`CL ~ CRCL power`, `CL ~ WT power`), OFVs within 1e-4, p-values to the digits PsN prints. Why 0.05: at 0.01/0.001 the step-1 runner-up misses the cutoff by 0.04 OFV and the backward phase removes the winner — a trajectory decided below the resolution two engines can be expected to share.

- [x] Degenerate oracle (`covsearch_end_to_end.rs`): a single-point space returns the base fit bit-for-bit vs `fit()`.
- [x] Strictness: a candidate with the best OFV that fails the gate is not selected and stays in the table with its reason (`mod_tests.rs`), and the search retries it from the next parent.

## Performance
- [x] No significant impact expected (new code path; the core fix is a predicate)

## Tests
- [x] Unit: 42 in `covsearch/` + `allometry/` (scripted fitter: selection, LRT df, strictness exclusion, unjudgeable child, pair retirement, max_steps, backward, forced/base immunity, adaptive stash, cancellation, options, space resolution, report); core `theta_boundary_side` regression.
- [x] Tier-2: `covsearch_end_to_end.rs` (degenerate oracle, a real forward step, allometry) and 6 CLI tests incl. a whole search from the command line on evaluations.
- [x] Tier-3: `covsearch_psn_anchor.rs` (slow-gated).

- [x] Example inline (see `docs/tools/covsearch.qmd`; `tests/psn/scm_anchor/scm.ferxsearch` is a runnable one)

## Docs
- [x] `docs/tools/covsearch.qmd` (algorithm, strictness-beside-ΔOFV, output, PsN comparison, deviations, allometry, Rust API); `docs/cli.qmd`; sidebar in `_quarto.yml`
- [x] `CHANGELOG.md`: Added (covsearch + allometry) and Fixed (boundary predicate)

## Checklist
- [x] `tools/preflight.sh` green (fmt, check ×5, clippy, rustdoc, docs, public-API)
- [x] Not on the `Dual2` path
- [x] No version bump

## Docs & examples
- [ ] ferx-site / ferx-book: follow-up once the epic's CLI surface settles (#1184)
- [x] Rebuilt release: `cargo build --release --workspace`
- [x] `ferx covsearch crates/ferx-tools/tests/psn/scm_anchor/scm.ferxsearch` runs clean (0.35 s for the nine-candidate search in release)

## Reviewer hints

- `covsearch/mod.rs::search` is the whole algorithm; `judge_one` is where the backward phase swaps the LRT roles.
- The core fix in `postfit.rs` is small but affects every user with wide asymmetric θ bounds: the `BoundaryEstimate` warning (and the bootstrap's `skip_estimate_near_boundary`) stop firing on interior estimates.
- Deviations from Pharmpy are listed at the top of `covsearch/mod.rs` and on the docs page.

## Open questions

- Pharmpy's adaptive-stash pair filter (drops `CL-CRCL` once `CL-WT` and `V-CRCL` are in) looks like a quirk rather than a design; ferx filters by exact pair. Happy to mirror Pharmpy instead if parity matters more than the behaviour.


roninsightrx (commented):
I found three blocking correctness issues:

1. [P1] src/api/postfit.rs:1081 — Cap the identity-bound tolerance by the declared interval. The new max(abs(bound), 1) floor makes the absolute tolerance 1e-3 even when the entire legal interval is much narrower. For theta X(0, -1e-4, 1e-4), an estimate of 0 is exactly the midpoint, but its distance to the lower bound is 1e-4, so theta_boundary_side reports lower and the default strictness gate rejects a genuinely interior fit. This regresses the old range-fraction behavior for narrow identity-packed parameters while fixing the wide asymmetric case. Please cap the bound-scaled tolerance by a fraction of the full range (or otherwise preserve interior points) and add this midpoint regression test.

2. [P1] crates/ferx-tools/src/covsearch/mod.rs:828-847 — Do not select a winner from a cancelled partial step. Runner can return only the candidates that finished before cancellation, but this code computes winner, sets StepRow.selected, and emits StepFinished with that winner before checking report.cancelled. The parent is correctly left unchanged afterward, so the result can simultaneously say that an effect was SELECTED and that final_step is still the previous step; with missing earlier candidates, the effect sent in the progress event can also be indexed from the wrong remaining entry. Check cancellation before winner selection/emission, or force selected = false and emit selected: None for an interrupted step. The existing cancellation test comments that the winner is not taken but does not assert the selected field or event.

3. [P2] crates/ferx-cli/src/allometry_cmd.rs:63-74 — Accept signed numeric values. value() treats every token beginning with a hyphen as another flag, so legitimate commands such as --exponents -0.5 and estimated exponent bounds such as --lower -2 --upper -0.1 are rejected as missing values. I reproduced this with ferx allometry examples/warfarin.ferx --data data/warfarin.csv --parameters CL --exponents -0.5. Numeric options need to parse signed tokens rather than rejecting them solely by prefix.

The repository patch-coverage gate is also currently failing at 88.94% (173 changed lines uncovered), below the required 90%.

Local verification: cargo test -p ferx-tools --lib passes all 339 tests.

roninsightrx:
Thanks — all three are real, and all three are fixed in the follow-up commit.

**1. `theta_boundary_side` — narrow identity intervals.** Agreed: the `max(|bound|, 1)` floor traded the wide-asymmetric bug for a narrow-interval one. The tolerance is now `BOUNDARY_REL_TOL · min(max(|bound|, 1), upper − lower)`, i.e. bound-scaled but capped by the same fraction of the declared range. `(-100, 1e6)` still puts 0.7 interior (tol 0.1 at the lower bound, 1000 at the upper), and `(-1e-4, 1e-4)` now keeps its midpoint interior (tol 2e-7). The midpoint case, both ends, and an interior point of the narrow interval are in `theta_boundary_side_detects_bounds`.

**2. Selecting from a cancelled step.** Agreed, and the index hazard was real too: `judge` skips candidates the runner never reached, so with a candidate missing the row index no longer matched `remaining[i]`. Both `Pass::run` and `Pass::backward` now check `report.cancelled` *before* selection: the partial rows are kept, nothing is marked `selected`, `StepFinished` carries `selected: None`, and a `debug_assert_eq!(step_rows.len(), remaining.len())` pins the index correspondence on the complete-step path. `a_cancelled_step_ends_the_search_with_what_it_had` now asserts `selected == false` on every row and the exact event (with CL-WT set up so it *would* have won), and a new `a_cancelled_backward_step_selects_nothing_either` does the same for the backward phase with a removable candidate scripted by id.

**3. Signed numbers on `ferx allometry`.** Agreed. `value()` now refuses the next token only when it is one of the command's own flags (`VALUE_FLAGS ∪ BOOL_FLAGS`), not on a leading hyphen. `--exponents -0.5,1.5 --lower -2 --upper -0.1` is a unit test, and the `.ferxsearch` CLI test passes `--exponents -0.5,1` end to end.

**Patch coverage.** The misses were concentrated in the two CLI modules, the search's error/cancel branches and the summary renderer. Added: unit tests for both CLI flag parsers and their bad invocations; base-model refusals (gate failure, fit error, missing cached fit); the fitless-candidate note; every status/decision branch of `render_summary` plus the cancelled label and notes; `write_report` with no final fit; a pre-cancelled `run_allometry`; and two more CLI end-to-end runs (allometry from a `.ferxsearch` with estimated exponents; a search that accepts effects at `p_forward = 1`, then `--resume --quiet`). Preflight is green locally; waiting on codecov for the number.

roninsightrx:
Confirmed works in manual testing.

Issue #1239, opened 2026-09-05
bug(ci): nightly slow-tests is killed at 120m in vi_dcm_omega_recovery since #1215 enabled ferx-core/nn — and a timeout reports as 'cancelled', not a failure


TeunP:
> **Ownership.** Assigned to @roninsightrx, who merged #1215. The sizing decision below —
> what the 5 DCM fits are meant to prove and what they should cost in CI — belongs to
> @dominic-irx, who authored #1215; he is cc'd rather than assigned only because he is not
> in this repo's assignable list.

`Slow tests` has been killed at its 120-minute timeout on every run since #1215 merged, and
because a `timeout-minutes` kill is reported by GitHub as **`cancelled`, not `failure`**, it does
not read as a red test — so the nightly has quietly been covering less than it appears to.

## What happens

#1215 added `ferx-core/nn` to the nightly's feature set — deliberately, so the new DCM
regularizer fits would run somewhere. That also switched on every *other* `nn`-gated Tier-3 file.
`tests/vi_dcm_omega_recovery.rs` is `#![cfg(feature = "nn")]` (line 95) with 5 DCM VI/FOCEI
convergence fits, and it **had never run in CI before**.

Both timed-out runs die in exactly the same place:

| run | last output | killed | silence |
|---|---|---|---|
| [33914230476](https://github.com/FeRx-NLME/ferx-core/actions/runs/33914230476) (`1a31cfce`) | 20:38:52 `Running tests/vi_dcm_omega_recovery.rs` | 22:02:55 | **84 min** |
| [33917510422](https://github.com/FeRx-NLME/ferx-core/actions/runs/33917510422) (`587f808d`) | 21:18:08 | 22:42:09 | **84 min** |

```
Terminate orphan process: pid (15172) (vi_dcm_omega_recovery-296157ee28e3378b)
```

Everything alphabetically after it — roughly 20 of ~166 test binaries — never runs.

## It is compute, not a deadlock

Run locally on an 8-core machine (`--features ci,survival,nn,slow-tests --profile ci-test`):

```
running 5 tests
test the_two_covariate_designs_are_what_they_claim_to_be ... ok
test capacity_alone_does_not_collapse_iiv_when_the_design_identifies_omega has been running for over 60 seconds
test iiv_collapses_when_the_network_can_index_subjects            ... over 60 seconds
test vi_and_focei_both_recover_iiv_on_an_identified_dcm           ... over 60 seconds
[...]

Issue #2, opened 2026-05-13
scaling (automatic) and via model code (S = A/V)


TeunP:
- add scaling to model code
- for multiple analytes: make it feasible to assign scaling - for this the IPRED (or F) needs to be made explicit --> if(CMT =2) IPRED = A/V or if(CMT=2) IPRED = F
related: https://github.com/FeRx-NLME/ferx-r/issues/3

roninsightrx:
Two different things here:
1. being able to define exactly what the modeled DV is (IPRED in this case, or F for analytic eqs)
2. scaling (due to mismatch in units)
Let's focus first on the the 1st one (we may not need the 2nd...)

I would say for analytic solutions, we can just add a parameter `scale` to the functions, So e.g. 
```
[structural_model]
  pk two_cpt_iv_bolus(cl=CL, v1=V1, q=Q, v2=V2)
```
would become
```
[structural_model]
  pk two_cpt_iv_bolus(cl=CL, v1=V1, q=Q, v2=V2, scale = 1000)
```
(with `scale` being optional, of course).

For ODE models, it's more complex: here we anyhow need to be able to write the dv out more explicitly, we currently don't have that (observation compartment can only be modeled as concentration, not as amounts). So we need something to define the scaling, so either:

```
[structural_model]
  ode(obs_cmt=central, obs_scale="V/1000", states=[depot, central])
```

or more flexible as:
```
[structural_model]
  IPRED = central/(V/1000)
  ode(obs_var=IPRED, states=[depot, central])
```
I'm more in favor of the 2nd option for ODEs, then the user has to be very explicit about the equation to calculate the dv.

TeunP:
I was thinking of something that could be more user friendly and less error prone, tackling both 1 and 2 - making use of units, possible use something like pknca_units_table()


I would say for analytic solutions, we can just add a parameter `amtu` and `dvu` to the functions (and possibly which `cmt` has the observation), which would allow scaling done automatically for any combination of units; i.e. 1000 or 0.001 or any other scaling based on the units (even use `mw` if we want to be very comprehensive).
 
```
[structural_model]
  pk two_cpt_iv_bolus(cl=CL, v1=V1, q=Q, v2=V2, amtu = mg, dvu = mg/L, mw = 120 )
```
and via R (can be used with `|>`

```
ferx_units(amtu = mg, dvu = mg/L, mw = 120)
```

[...]

PR #1240, opened 2026-09-05
fix(ci): isolate the nn slow fits, and make a timeout fail red (part of #1239)

TeunP:
## Why

`Slow tests` has been killed at its 120-minute timeout on **every run since #1215 merged**, and it
did not look like a failure: a `timeout-minutes` kill is reported by GitHub as **`cancelled`, not
`failure`**. Five nightlies in a row read as infra blips while the suite was silently covering
~20 fewer test binaries than it appeared to.

Full diagnosis in #1239. In short:

#1215 added `ferx-core/nn` to this job so the new DCM regularizer fits would run somewhere — a
reasonable call. It also switched on every *other* `nn`-gated Tier-3 file, and
`tests/vi_dcm_omega_recovery.rs` (`#![cfg(feature = "nn")]`, 5 DCM VI/FOCEI fits) had **never run
in CI**. Both timed-out runs die identically:

| run | last output | killed | silence |
|---|---|---|---|
| [33914230476](https://github.com/FeRx-NLME/ferx-core/actions/runs/33914230476) | 20:38:52 `Running tests/vi_dcm_omega_recovery.rs` | 22:02:55 | **84 min** |
| [33917510422](https://github.com/FeRx-NLME/ferx-core/actions/runs/33917510422) | 21:18:08 | 22:42:09 | **84 min** |

**It is compute, not a deadlock.** Locally that binary sits at 590–790% CPU (~7 cores) and is
still unfinished after 13 minutes. This job pins `RAYON_NUM_THREADS: 1` on a 2-core runner while
libtest runs 4 of the 5 tests concurrently — four heavy fits contending for two cores.

## What changed

Two changes, neither touching the fits:

1. **`ferx-core/nn` moves to its own `slow-tests-nn` job.** An over-budget NN fit now costs that
   job and nothing else, instead of taking ~165 unrelated binaries down with it.
2. **Both jobs run under `timeout 100m`**, so exceeding the budget exits 124 and fails the step
   loudly with an `::error::` annotation, rather than being reported as `cancelled`.

## Alternatives considered

- **Just raise `timeout-minutes`.** Rejected: it does not fix the silent-coverage-loss half, and
  the NN fits would still be able to consume the whole budget.
- **Enumerate the NN test targets in job 2** instead of a full `--workspace` run. Rejected: a new
  `#![cfg(feature = "nn")]` file would go silently uncovered — exactly the stale-list failure mode
  #949 already cost this repo. The duplicated non-NN work is the price, and this is a nightly.
- **Relax `RAYON_NUM_THREADS: 1` or use a larger runner for the NN job.** Plausibly the right
  answer, but it changes fit reproducibility and is a call for whoever owns the DCM work — #1239.

## Breaking changes
- [x] None. CI only.

## Numerical validation
- [x] Not applicable (CI only)

## Tests
- [x] No new tests needed

`tests/slow_test_path_filter.rs` is the guard that pins this workflow, and it passes **11/11**. It
collects `--exclude` tokens from the whole file, so the second job's identical `--exclude
docs-lint` leaves its expectations unchanged. The workflow also parses as valid YAML with both
jobs present.

## Docs
- [x] No user-visible change — no CHANGELOG entry (CI only)

## Reviewer hints

The duplication in job 2 is deliberate and is the main thing to push back on if you disagree — see
Alternatives. Everything else is mechanical.

Worth knowing for the future regardless of this PR: **a `timeout-minutes` kill is `cancelled`, not
`failure`.** Any job that can outgrow its budget can go quiet this way, and the `timeout` wrapper
here is the general remedy.

## Open questions

This unblocks the nightly; it does **not** make the fits affordable — the NN job will likely still
hit its budget until the fits are sized. That is #1239 and I have not attempted it, since it is a
judgement call about what those 5 fits are meant to prove.

Separately, and noted in #1239 rather than fixed here: compiling `ferx-core --features nn
--profile ci-test` took **2h20m** in a single `rustc` locally, versus ~15 min both without `nn` and
with `nn` at `19fdbb04`. That may share a root cause with the runtime cost.

**Partially addresses #1239 — deliberately does NOT close it.** This stops one over-budget
binary from taking down the whole nightly and makes an overrun visible. The substantive half of
#1239 is untouched: the 5 DCM fits are still unaffordable under `RAYON_NUM_THREADS: 1` on a
2-core runner, and the 2h20m `nn` compile is unexplained. #1239 stays open for those.

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -69,9 +69,20 @@ env:
 
 jobs:
   slow-tests:
-    name: Slow regression tests
+    name: Slow regression tests (${{ matrix.leg }})
     runs-on: ubuntu-latest
     timeout-minutes: 120
+    strategy:
+      # A failing leg must NOT cancel the other -- see the `fail-fast` note below.
+      fail-fast: false
+      matrix:
+        include:
+          - leg: core
+            cache: ci-slow
+            features: ferx-core/ci,ferx-core/survival,ferx-core/slow-tests,ferx-tools/slow-tests
+          - leg: nn
+            cache: ci-slow-nn
+            features: ferx-core/ci,ferx-core/survival,ferx-core/nn,ferx-core/slow-tests,ferx-tools/slow-tests
     steps:
       - uses: actions/checkout@v7
       - name: Install nightly toolchain
@@ -79,7 +90,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          shared-key: ci-slow
+          shared-key: ${{ matrix.cache }}
       - name: cargo test (ci-test profile + slow-tests)
         # `--profile ci-test` (defined in Cargo.toml) keeps release-level
         # optimisation so the heavy fits run fast, but drops fat LTO — the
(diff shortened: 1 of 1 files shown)


TeunP:
## Self-review (xhigh) — 8 findings, 4 fixed in `7234ab89`

Reviewed at xhigh across ten finder angles. The review earned its keep: one finding was a change
that would have silently reintroduced the bug this PR exists to fix.

---

### 1. Adjacent comments contradicted each other — `correctness` — **fixed**

The pre-existing block explaining *why* `ferx-core/nn` must be in the feature list ("without `nn`
here those files compile to empty binaries ... would run in no CI job at all") was left in place,
directly above my new comment saying it had been removed. A reader hitting the first block would
conclude the feature list was wrong and "restore" it — re-breaking the nightly.

That justification is still correct and still load-bearing, so it is rewritten around the matrix
rather than deleted.

### 2. Two near-identical jobs → one matrix — `simplification` — **fixed**

The two jobs differed by exactly one feature token but duplicated all four steps and both 11-line
run blocks. Worse, the guard test only checks that `--exclude` appears *somewhere* in the file, so
a one-sided edit would have passed it.

### 3. `fail-fast: false` — `correctness` — **fixed, and this is the important one**

The trap inside fix #2. **A matrix leg failing cancels its siblings by default.** Collapsing the
two jobs into a matrix without it would have put the core leg straight back to being killed by the
nn leg — the refactor that removed duplication would have silently undone the entire purpose of
the PR. It is now set, and commented as load-bearing rather than stylistic.

### 4. The 100m budget was asserted, not derived — `correctness` — **fixed**

Now derived in-file: 43m observed on a fast runner, 1.7x measured spread across `ubuntu-latest`
draws (#1239), ~74m worst case, inside a 120m ceiling that must also absorb checkout, toolchain
install and cache restore. The workflow also now records what only the PR body said: the nn leg is
**expected to stay red** until #1239 sizes those fits, that this is deliberate, and that the remedy
is #1239 rather than muting the leg.

---

## Not fixed — flagged rather than papered over

### 5. `timeout --signal=INT` returning 124 is unverified — `correctness`

This is the entire basis of "fails red instead of `cancelled`". GNU `timeout` documents 124 on
timeout, but this sends SIGINT rather than SIGKILL; if cargo handles it and exits 130 before
`--kill-after=2m` elapses, the `::error::` annotation never prints. The step still fails red, so
the core fix survives either way — but the diagnostic would be skipped.

**I could not test it:** macOS has no GNU `timeout` (`timeout` and `gtimeout` both absent). Asserted
from documentation, not measured.

### 6. `timeout` signals cargo, not the test binary — `correctness`

The CI logs prove the orphan outlives cargo — both timed-out runs ended with *two* separate
entries, `(cargo)` and `(vi_dcm_omega_recovery-...)`. `timeout` signals only its direct child, so
the compute-bound test binary keeps running and keeps the inherited stdout pipe open. The step's
shell can then block past the 100m budget and the job hits `timeout-minutes: 120` — reinstating the
exact silent `cancelled` outcome this PR removes. `--foreground` plus a process-group kill, or
`setsid`, would close it.

**Findings 5 and 6 are precisely what the verification run settles**, which is why I have not
guessed at them — see below.

### 7. The nn leg duplicates the whole suite nightly — `efficiency`

~45 extra runner-minutes/night. The PR argues this beats an enumerated NN target list because a new
`#![cfg(feature = "nn")]` file would go silently uncovered (the #949 failure mode). A third option
was not considered: enumerate the targets **and** add a guard test asserting every `nn`-gated test
file is reached. That keeps the coverage property without the duplicate suite.

### 8. Second cache key, unremarked — `efficiency`

`ci-slow-nn` is correct (differing feature hash would thrash a shared key), but GitHub caps Actions
cache at 10 GB/repo with LRU eviction. Two full Rust caches for this workspace can approach that,
and evicting the PR caches would slow every PR. Not noted anywhere.

---

## Verification

`tests/slow_test_path_filter.rs` — the guard that pins this workflow — passes **11/11** against the
matrix form, and the YAML parses with `fail-fast: false` and both legs.

That is not sufficient, and the insufficiency is the same one that bit #1228: **this workflow only
runs on push-to-`main` or dispatch, so the PR's own checks cannot execute the thing being changed.**
Dispatched against the branch: run
[33961533474](https://github.com/FeRx-NLME/ferx-core/actions/runs/33961533474).

**The pass criterion is not "both legs green"** — the nn leg is *expected* to fail, and that is the
designed behaviour:

| leg | expected | meaning |
|---|---|---|
| `core` | **success** | isolation works — the nn fits no longer cost the main suite |
| `nn` | **failure at ~100m**, ideally with the `::error::` annotation | the `timeout` wrapper works; findings 5 and 6 are disproved |
| `nn` | ~~`cancelled` at 120m~~ | the fix **failed** — findings 5/6 confirmed, do not merge |

An earlier dispatch of the pre-matrix two-job form was cancelled; it did not exercise `fail-fast`,
which is now the riskiest line in the diff.

TeunP:
## Verified on CI and merged

Dispatched against the branch, since this workflow only runs on push-to-`main` or dispatch and the
PR's own checks structurally cannot execute it. Run
[33961533474](https://github.com/FeRx-NLME/ferx-core/actions/runs/33961533474):

| leg | result | evidence |
|---|---|---|
| `core` | **success**, 41 min (10:46 → 11:27) | 5680 passed, **0 failed**; both #1228 tests present and passing |
| `nn` | **failure at exactly 100 min** (10:46 → 12:26) | `exit code 124` + the `::error::` annotation |

Both legs ran to completion concurrently, which is the direct proof that `fail-fast: false` works —
the nn leg failed and did **not** cancel its sibling. The core leg's 41 minutes matches the
historical known-good full-suite time of 43 min, i.e. the suite that had been killed at 120m for
five consecutive nightlies now finishes comfortably.

### The two findings I flagged as unverified are now disproved

Both were load-bearing and neither was testable locally (macOS has no GNU `timeout`):

**Finding 5 — does `timeout --signal=INT` actually return 124?** Yes:

```
##[error]slow-tests (nn) exceeded its 100m budget - a test is hanging or the suite outgrew the runner; see FeRx-NLME/ferx-core#1239
##[error]Process completed with exit code 124.
```

The annotation prints and the exit status is 124, so a maintainer sees *why*, not just a bare
failure.

**Finding 6 — does the orphaned test binary hold stdout open and drag the job to a 120m
`cancelled`?** No. The step exited at 100m on the nose. The silent-`cancelled` failure mode this PR
exists to remove is gone.

### What this does and does not do

The nn leg failing at 100m **is the designed outcome**, not a defect — it is the honest signal that
those fits do not fit the budget, and it now costs only its own leg. The nightly will stay red on
that leg until #1239 sizes the DCM fits; that is @roninsightrx's and @dominic-irx's call, and this
PR deliberately does not close #1239.

Findings 7 and 8 (the duplicated suite in the nn leg, and the second cache key's storage cost)
remain open trade-offs, recorded in the review above rather than fixed.

Merged as `b026b531`.

PR #517, opened 2026-06-25
fix(parser): stop spurious "ignored" warning for ode_reltol/abstol/max_steps [closes #516]

TeunP:
## Why

The ODE-solver fit options `ode_reltol`, `ode_abstol`, and `ode_max_steps` emit a false *"is not used by method `<M>` and will be ignored"* warning, even though they **are** applied to the RK45 integrator for any ODE model via `CompiledModel::sync_ode_solver_opts` (`src/types.rs:2286`) at parse time. Most visible on ODE absorption models that set tight tolerances to match NONMEM's `TOL=9` — surfaced by the Weibull anchor (#503). Closes #516. Not method- or Weibull-specific.

## What changed

`FitOptions::unsupported_keys_warnings()` checks each user-set key against `framework_keys() ∪ method_specific_keys(method)`. The three integrator keys were in neither list. They are framework-level — method-agnostic, gated on the model being an ODE model, not on the estimator — so they are added to `framework_keys()` (alongside `threads` / `covariance`). The false positive disappears with no behaviour change.

## Alternatives considered

Adding them to each method's `method_specific_keys()` instead — rejected: they are not method-specific (every estimator that integrates an ODE model honours them), and that would duplicate the keys across all eight methods.

## Cross-repo dependency
| Repo | PR | Status | Must merge |
|------|----|--------|------------|
| ferx (R) | — | not needed | — |

No API change; warning text only.

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable (docs / refactor / CI only) — warning text only; no numerical path touched.

## Performance
- [x] No significant impact expected

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes)
- [x] Regression test added that fails without this fix — `test_ode_solver_keys_do_not_warn` (asserts no warning under focei/foce/saem/imp/bayes); the warning suite is green (53 passed).

## Example (user-facing features only)
- [x] Not applicable (internal / fix with no new API surface)

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added (under `Fixed`)
- [x] No user-visible change beyond removing the misleading warning (the keys were already documented in `docs/model-file/fit-options.qmd` as "ODE models only").

## Checklist
- [x] `cargo clippy` clean — change is string literals + a straightforward test loop; CI confirms
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A (not touched)
- [x] `Cargo.toml` version bumped if breaking — N/A (additive)

### Example execution
[...]

Issue #516, opened 2026-06-25
Spurious "will be ignored" warning for ode_reltol / ode_abstol / ode_max_steps


TeunP:
### Problem

Setting any of the ODE-solver fit options under a FOCEI (or any non-default-method) ODE model emits a misleading warning:

```
fit option `ode_reltol` is not used by method `FOCEI` and will be ignored.
Method-specific options for `FOCEI`: global_maxeval, global_search, …
```

The warning is **wrong**: the option *is* applied. `ode_reltol` / `ode_abstol` / `ode_max_steps` configure the RK45 integrator and are pushed onto the solver for any ODE model via `CompiledModel::sync_ode_solver_opts` (`src/types.rs:2286`) at parse time, regardless of estimation method.

### Root cause

`FitOptions::unsupported_keys_warnings()` checks each user-set key against `framework_keys() ∪ method_specific_keys(method)`. The three `ode_*` integrator keys are in **neither** list, so they are flagged as ignored.

### Fix

They are framework-level (method-agnostic; gated on the model being an ODE model, not on the estimator), so add them to `framework_keys()` alongside `threads` / `covariance`. Surfaced by the Weibull NONMEM anchor (#503), whose model sets `ode_reltol=1e-9` to match `TOL=9`. Not Weibull-specific — affects every ODE model.

PR #1243, opened 2026-09-05
feat(api): make the unsupported-keys warning model-aware for the ODE solver keys [closes #518]

### LINKED ISSUE #518: Make unsupported-keys warning model-aware: notify when ode_* solver keys are set on a non-ODE model
## Background

PR #517 (#516) fixed a false-positive warning: `ode_reltol` / `ode_abstol` / `ode_max_steps` were flagged as *"not used by method `<M>` and will be ignored"* on ODE models — even though `CompiledModel::sync_ode_solver_opts` (`src/types.rs:2284`) applies them to the RK45 integrator. The fix added the three keys to `framework_keys()`, which suppresses the warning **unconditionally**.

## Gap

`framework_keys()` is method- *and* model-agnostic: `FitOptions::unsupported_keys_warnings()` (`src/types.rs:4057`) only sees `FitOptions`, never the `CompiledModel`. So the three keys now never warn — including on an **analytical** (non-ODE) model, where `sync_ode_solver_opts` is a no-op and they genuinely have no effect. We traded a false-positive on ODE models for a silent false-negative on analytical models. This matches the existing behaviour of the other model-conditional framework keys (`frem_predictions`, `frem_sigma`, `iov_column`), which also don't warn when their feature is absent.

> **This does not affect *using* the options.** They already work on every ODE model today — that is exactly what #517 confirmed (`sync_ode_solver_opts` applies them on the CLI / `predict` / `fit_from_files` paths). This issue is purely about better diagnostics when the keys are set where they do nothing.

## Proposal

Make the warning model-aware for the model-conditional keys. This is a modest change, not a large plumbing job — the production call site (`src/api.rs:2592`, inside `fit_inner`) already has `model: &CompiledModel` in scope, and `CompiledModel::is_ode_based()` (`src/types.rs:2254`) already exists:

- Thread `is_ode_based` (or `&CompiledModel`) into `unsupported_keys_warnings`.
- Treat `ode_reltol` / `ode_abstol` / `ode_max_steps` as applicable only when the model is ODE-based; otherwise emit a tailored notice, e.g. *"`ode_reltol` configures the ODE integrator but this model has no `[odes]` block, so it has no effect."*
- (Optional, possibly its own follow-up) generalise the same treatment to the other model-conditional framework keys (`frem_*` → needs FREM; `iov_column` → needs IOV) so the warning layer becomes uniformly model-aware.

## Acceptance criteria

- Analytical model + `ode_reltol` set → exactly one warning naming the key and the "no `[odes]` block" reason.
- ODE model + the same keys → no warning (keep `test_ode_solver_keys_do_not_warn` green as the regression guard).
- Unit test at the parser / `FitOptions` tier covering both branches.

## Priority

Low — diagnostic/UX quality only. No numerical or public-API impact. Follow-up to #517 / #516.

roninsightrx:
## Why

Closes #518. Follow-up to #517 / #516.

#517 fixed a false positive: `ode_reltol` / `ode_abstol` / `ode_max_steps` were flagged *"not used by method `<M>` and will be ignored"* on ODE models even though `CompiledModel::sync_ode_solver_opts` applies them. The fix added the keys to `framework_keys()`, which suppresses the warning **unconditionally**.

`framework_keys()` is method- *and* model-agnostic — `FitOptions::unsupported_keys_warnings()` only ever sees `FitOptions`, never the `CompiledModel`. So the keys now never warn, including on an **analytical** model where `sync_ode_solver_opts` is a no-op and they genuinely do nothing. A false positive on ODE models was traded for a silent false negative on analytical ones.

This does not change whether the options *work*; they already work on every model that integrates. It is a diagnostics fix.

## What changed

- **`FitOptions::unsupported_keys_warnings_with_model(&self, model: &CompiledModel)`** (`src/types.rs`) — delegates to the existing `unsupported_keys_warnings()` for the method-level warnings, then **appends** one tailored notice per user-set `ode_*` key when the model never integrates:

  > ``fit option `ode_reltol` configures the ODE integrator, but this model has no `[odes]` block (and no closed-form absorption ODE twin), so it has no effect.``

- **`CompiledModel::honors_ode_solver_opts()`** — `ode_spec.is_some() || absorption_ode_equivalent.is_some()`.

- **`ode_solver_keys()`** — the `ode_*` subset of `framework_keys()`, as its own list so the model-conditional treatment names exactly those keys.

- **`src/api/fit.rs`** — `fit_inner` calls the model-aware variant; `model: &CompiledModel` was already in scope, so there is no plumbing.

Two things worth flagging:

**The predicate is deliberately not `is_ode_based()`.** A closed-form transit / inverse-Gaussian model is analytic (`ode_spec == None`) but carries an absorption ODE twin (#814) that `sync_ode_solver_opts` also stamps — its TV-covariate / `TIME` / IOV subjects integrate on it. Gating on `is_ode_based()` would tell those users "this model has no `[odes]` block" while the keys were in fact live, i.e. it would reintroduce exactly the class of false positive #517 removed.

**All six `ode_*` keys, not just the three the issue names.** `ode_method`, `ode_stiff_abort_after` and `ode_auto_switch` are stamped by the same function and are equally inert on a non-integrating model; warning on three of six would be arbitrary.

The optional generalisation in the issue (`frem_*` → needs FREM, `iov_column` → needs IOV) is **not** included — those want their own model predicates and their own tests, and the issue itself flags them as a possible follow-up.

## Alternatives considered

- **Change the signature of the public `unsupported_keys_warnings`** (to take `Option<&CompiledModel>`, as the issue's first bullet suggests). Rejected: it breaks a public-API item for a low-priority diagnostic and churns ~40 existing call sites in tests, for no gain — the only production caller is `fit_inner`, which is in-crate.
- **Add the model-aware entry point as `pub`.** Rejected: nothing outside the crate calls it (ferx-r does not use this API at all), so widening the surface would fail the "which tool needs this item" test in CLAUDE.md. Everything added here is `pub(crate)`; the committed baseline is unchanged.
- **Drop the keys back out of `framework_keys()`** and rely solely on the model layer. Rejected: they would then draw the method-level "not used by method FOCEI" warning again on ODE models — the #516 bug. A unit test now pins `ode_solver_keys() ⊆ framework_keys()` so this cannot regress by accident.

## Cross-repo dependency

No public-API change, and `ferx-r` does not call this API.

## Breaking changes
- [ ] Public Rust API
- [x] **None** — every new item is `pub(crate)`; `tools/update-public-api.sh --check` reports *"public API matches the committed baseline"*.

## Numerical validation
- [x] **Not applicable** (diagnostics only). No estimator, objective, gradient or output value is touched: the change adds strings to `FitResult.warnings` and nothing else. `sync_ode_solver_opts` is untouched, so which models honour the keys is exactly as before.

## Performance
- [x] No significant impact expected — one `Vec<String>` scan over `user_set_keys`, once per fit, guarded by an early return on models that integrate.

## Tests
- [x] Unit tests added (Tier 1, `src/types_tests.rs`) — `cargo test --lib --features ci`: **4041 passed, 0 failed**.
- [x] Regression test that fails without this fix — `analytical_model_warns_that_ode_solver_keys_have_no_effect` returns an empty vec on `main`.

Six tests, and what each exists to catch:

| Test | Catches |
|---|---|
| `analytical_model_warns_that_ode_solver_keys_have_no_effect` | the #518 false negative; also asserts the message is *not* the method-level phrasing, which would be the wrong diagnosis |
| `ode_model_does_not_warn_on_ode_solver_keys` | the #516/#517 false positive, now at the model tier |
| `closed_form_absorption_twin_suppresses_the_ode_solver_key_notice` | a naive `is_ode_based()` predicate — this model is `ode_spec == None` **and** must stay silent |
| `analytical_model_warns_once_per_distinct_ode_solver_key` | a key dropped from the notice, and a duplicate `user_set_keys` entry doubling up |
| `model_aware_warnings_keep_the_method_level_ones` | the model layer swallowing or replacing the method-level warnings; also pins the model-agnostic entry point's behaviour as unchanged |
| `ode_solver_keys_are_a_subset_of_framework_keys` | the two lists drifting — an `ode_*` key in one and not the other resurrects #516 |

Per the "a green test is not evidence that it can fail" rule in CLAUDE.md: the analytical and ODE tests **straddle** the predicate (same keys, opposite branch), each asserting `honors_ode_solver_opts()` explicitly, so neither can silently start taking the same arm as the other. There is no `is_empty()` fast path in front of the filter — the loop runs on any non-empty `user_set_keys` — so the mutation "filter the wrong key set" is reachable from these fixtures.

## Example

```toml
[structural_model]
  pk one_cpt_iv(cl=CL, v=V)   # analytical: no [odes] block, no absorption twin

[fit_options]
  method     = focei
  ode_reltol = 1e-9
```

Before: silent. After, in `FitResult.warnings` and on the CLI:

```
fit option `ode_reltol` configures the ODE integrator, but this model has no `[odes]`
block (and no closed-form absorption ODE twin), so it has no effect.
```

An `[odes]` model — or a `one_cpt_transit` / `two_cpt_ig` model reaching its twin — with the same keys emits nothing.

## Docs
- [x] `docs/model-file/fit-options.qmd` — a paragraph under *ODE solver tolerances* stating when the keys warn.
- [x] `CHANGELOG.md` `[Unreleased] → Changed` entry.
- [ ] New pages linked in `docs/_quarto.yml` — no new page.

Note: the added paragraph pushed *ODE solver tolerances* to 5,075 characters and R1 (the 5,000-char terminal-section rule) went red. It is split under two `####` subheadings — `Where the value comes from {#ode-tolerance-precedence}` and `The keys {#ode-tolerance-keys}` — rather than added to `docs/.lint-baseline`, per the ratchet rule. The `#ode-solver-tolerances` anchor is unchanged, so existing links still resolve.

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy `--all-targets`, rustdoc, docs lint, public-API baseline.
- [x] Nothing on the `Dual2` sensitivity path is touched.
- [ ] `Cargo.toml` version bump — not needed (non-breaking).

### ferx-site / ferx-book
- [x] No new DSL syntax, no new `[fit_options]` key, no new example file or data file — the six keys already exist and are already documented. Only their diagnostics changed.

### Example execution
- [x] No example execution step needed — no user-visible change to any example's *results*; the only change is an extra warning line on a model that sets an `ode_*` key it cannot use, which no committed example does.

## Reviewer hints

The one line worth arguing about is the predicate in `honors_ode_solver_opts()` (`src/types.rs`). Everything else follows from it. `closed_form_absorption_twin_suppresses_the_ode_solver_key_notice` is the test that would go red if it were narrowed to `is_ode_based()`.

Skimmable: the tests, the changelog, the docs split.

## Open questions

Should the follow-up generalisation (`frem_*` → needs FREM, `iov_column` → needs IOV) be its own issue? The mechanism here is deliberately shaped so those slot in as extra key-lists plus predicates, but each needs a model predicate that does not exist yet, so I left them out rather than guess at the right one.

https://claude.ai/code/session_01Cq3rZMmYAA7dSZsQAMhE1j

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,12 @@ section of the SDLC for the versioning policy).
   bundled data by the bundled model too, so a reloaded joint fit keeps its event records.
 
 ### Changed
+- **The unused-fit-option warning is model-aware for the ODE solver keys (#518).** Setting
+  `ode_reltol` / `ode_abstol` / `ode_max_steps` / `ode_method` / `ode_stiff_abort_after` /
+  `ode_auto_switch` on a model that never integrates — analytical PK with no `[odes]` block and
+  no closed-form absorption ODE twin — now warns that the key has no effect, instead of being
+  dropped silently. Models that do integrate (including a closed-form transit / inverse-Gaussian
+  model reaching its twin) still never warn on these keys, as of #517.
 - **FREM prep refuses a model with a non-Gaussian endpoint (#1199).** `prepare_frem()` /
   `transform_dataset_for_frem()` return `E_FREM_NON_GAUSSIAN_ENDPOINT` instead of writing
   a dataset from the Gaussian rows alone; run the FREM step on the PK model without the
--- a/docs/model-file/fit-options.qmd
+++ b/docs/model-file/fit-options.qmd
@@ -94,6 +94,8 @@ The post-fit covariance step. See [Covariance Step](../estimation/foce.qmd#covar
 
 Accuracy and step budget for `[odes]` models, and for the auto-generated ODE twin that closed-form transit / inverse-Gaussian absorption models carry. See [ODE Models](ode-models.qmd#solver-details).
 
+#### Where the value comes from {#ode-tolerance-precedence}
+
 These six keys (and the two stepper keys below) can also arrive from a caller, and precedence is **per key** either way: a key the caller sets wins over the model file for that fit, a key the caller does not set keeps the file's value. What counts as "set" differs between the two front ends, so check which one you are on.
 
 From **R**, `ferx_fit(settings = list(ode_reltol = 1e-9))` merges your settings onto the model file's own `[fit_options]` and stamps the result on the model. Every key you name wins — *including* one you name with its default value, so `settings = list(ode_reltol = 1e-4)` does pull a file that pinned `1e-10` back to `1e-4`. Keys you leave out keep whatever the file said.
@@ -101,6 +103,10 @@ From **R**, `ferx_fit(settings = list(ode_reltol = 1e-9))` merges your settings 
 From **Rust**, `fit(&model, &pop, &params, &options)` — and equally `run_covariance()`, `run_sir()` and `run_sir_core()`, which integrate the same way — compares your `FitOptions` field by field against `FitOptions::default()` and carries only the moved fields for the duration of that call, so a hand-built settings struct cannot silently loosen a model that pinned `ode_reltol = 1e-10`. The consequence is the mirror image of the R rule: asking explicitly for a default value (`ode_reltol = 1e-4`) reads the same as not asking, and the model file still wins. That override also lasts exactly one `fit()` — a later `predict()` on the same `&CompiledModel` integrates at the model file's values, so a fit's IPREDs and a subsequent `predict()` can come from different tolerances. Call `model.sync_ode_solver_opts(&options)` on an owned model when both paths must match (that is what the R front end does).
 
 Put the key in `[fit_options]` when a model needs a specific accuracy regardless of who runs it.
+
+Set one of these on a model that never integrates — an analytical PK model with no `[odes]` block and no closed-form absorption twin — and the fit warns that the key has no effect (#518), instead of dropping it silently. A model that *does* integrate, including a closed-form transit / inverse-Gaussian model reaching its twin, never draws that warning.
+
+#### The keys {#ode-tolerance-keys}
 
 | Key | Values | Default | Description |
 |-----|--------|---------|-------------|
--- a/src/api/fit.rs
+++ b/src/api/fit.rs
@@ -931,7 +931,10 @@ fn fit_inner(
     // Compute up-front so we can both surface the warnings before the fit
     // starts (a long-running fit shouldn't bury a "this option is unused"
     // notice at the end) and carry them through into FitResult.warnings.
-    let mut pre_run_warnings = options.unsupported_keys_warnings();
+    // Model-aware: the `ode_*` solver knobs are framework-level for every *method*
+    // but conditional on the *model* (#518) — on a model that never integrates they
+    // are silently dropped, so say so rather than staying quiet.
+    let mut pre_run_warnings = options.unsupported_keys_warnings_with_model(model);
     // Surface a "no method specified, defaulting to FOCEI" notice through the
     // same channel so it reaches both stderr and FitResult.warnings.
     if let Some(w) = options.method_default_warning() {
--- a/src/types.rs
+++ b/src/types.rs
@@ -7773,7 +7773,9 @@ impl FitOptions {
                 continue;
             }
             warnings.push(format!(
-                "fit option `{key}` configures the ODE integrator, but this model has no                  `[odes]` block (and no closed-form absorption ODE twin), so it has no effect."
+                "fit option `{key}` configures the ODE integrator, but this model has \
+                 no `[odes]` block (and no closed-form absorption ODE twin), so it \
+                 has no effect."
             ));
         }
         warnings
--- a/src/types.rs
+++ b/src/types.rs
@@ -3834,6 +3834,20 @@ impl CompiledModel {
         self.ode_spec.is_some()
     }
 
+    /// Returns true when the `ode_*` solver knobs actually reach an integrator for
+    /// this model — i.e. when [`CompiledModel::sync_ode_solver_opts`] has something to
+    /// stamp them onto.
+    ///
+    /// Not the same predicate as [`Self::is_ode_based`]. A **closed-form**
+    /// transit/IG model is analytic, yet carries an absorption ODE twin
+    /// (`absorption_ode_equivalent`, #814) that its TV-covariate / `TIME` / IOV
+    /// subjects integrate on — and `sync_ode_solver_opts` stamps the tolerances onto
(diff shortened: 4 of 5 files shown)


roninsightrx (commented):
The model predicate and warning composition look correct, including the absorption-ODE-twin exception. I found one minor user-facing defect:

- `src/types.rs:7778`: the new warning literal contains a long accidental run of spaces between `no` and `` `[odes]` `` (`model has no                  `[odes]` block`). That exact spacing will appear in CLI and YAML output. Please collapse it to one space. The new tests only use `contains`, so an exact-message assertion (or at least an assertion rejecting repeated whitespace) would keep this from recurring.

Targeted local tests passed: all six `ode_solver_key` tests plus `model_aware_warnings_keep_the_method_level_ones`. Current CI is green.

roninsightrx:
Confirmed and fixed in `bf7c4da8`. Thanks — that one was invisible from the diff view, and the emitted string really did carry the run.

**Cause.** The `\` line-continuation in the literal never reached the file. I wrote that block through a shell/Python heredoc, where backslash-newline is *itself* a line continuation, so Python consumed the escape and what landed in `types.rs` was a single long literal that kept the source indentation. rustc had no reason to complain — it is a perfectly valid string. Rewritten with real `\` continuations, which `rustfmt` preserves:

```rust
"fit option `{key}` configures the ODE integrator, but this model has \
 no `[odes]` block (and no closed-form absorption ODE twin), so it \
 has no effect."
```

**On the test point — you were right that `contains` cannot see this, and there is a trap in the obvious fix.** I added both of these:

1. `analytical_model_warns_that_ode_solver_keys_have_no_effect` now pins the whole message with `assert_eq!` instead of substrings.
2. `analytical_model_warns_once_per_distinct_ode_solver_key` asserts no notice contains a double space.

(2) is not redundant with (1). An exact-message assertion is only as good as the literal on its right-hand side — had I written the expected string through the same heredoc, both sides would have carried the identical 18 spaces and `assert_eq!` would have passed while shipping the defect. The double-space check reads the **runtime** string and is independent of that literal, so it is what actually certifies the output is clean; it also covers the other five keys and any key added later, which the single exact-match does not.

I verified the fix that way rather than by re-reading the diff: the whitespace guard passes on the emitted string.

**Gates on `bf7c4da8`:** `cargo test --lib --features ci` — 4041 passed, 0 failed. `tools/preflight.sh` — green across `fmt check clippy rustdoc docs public-api`.

Nothing else changed: the predicate, the warning composition and the public-API surface are untouched.

Issue #522, opened 2026-06-25
Competing-risks VPC: explicit [simulation] horizon + TTE-row generation for the [simulation]-block path


TeunP:
### Context

Competing-risks TTE simulation (#501) is correct for the **template-driven** workflow — you pass a `Population` whose per-cause TTE records carry the administrative censoring, and `simulate_tte` draws the competing causes (earliest wins, the rest censored). #501 also fixed the re-simulation horizon to be consistent with #494: a record carrying an observed event (`Exact`) imposes **no** administrative horizon (+∞), so re-simulating an event-bearing subject is no longer truncated at its own observed event time.

What's still missing for a **proper competing-risks VPC** is an administrative horizon that is *decoupled from the observed event times*. In the cause-specific data layout a subject who had an event is recorded as `Exact` on the winning cause and `RightCensored` (at the event time) on the others — so the subject's planned follow-up time (study end) genuinely isn't recoverable from the data alone. A VPC needs that horizon supplied explicitly. The `simulate_tte` / `observation_window` docstrings already point to this as the "forthcoming `[simulation] horizon` (Phase 2)".

### Scope

1. **`[simulation] horizon` key.** Parse `horizon = <t>` (global) and/or a per-subject horizon column; reject unknown/malformed keys (mirroring the `[simulation]` validation added in #507).
2. **Honor it in simulation.** When a horizon is set, it becomes the administrative censoring window for *every* TTE cause, overriding the per-record `observation_window` — so re-simulating event-bearing data censors at the planned horizon rather than drawing unbounded.
3. **`[simulation]`-block CLI path emits zero TTE rows today** (synthetic subjects are built with empty `obs_records`, so `--simulate` produces no TTE events). Generate one TTE template row per cause CMT per synthetic subject, censored at the horizon (the draw overwrites). *This is the larger half of the work.*

### Acceptance

- Parser test for `[simulation] horizon` (+ unknown-key rejection).
- A VPC-style test: re-simulate event-bearing competing data **with** an explicit horizon → simulated event times are not truncated at the original event times, and censoring lands at the horizon.
- `[simulation]`-block TTE-row-generation test (CLI `--simulate` produces per-cause rows).
- Docs: document `horizon` in the simulation / `[fit_options]`-adjacent docs and the competing-risks section of `docs/estimation/tte.qmd`.

### Notes

- Not a blocker for #501 (the library `simulate(model, population)` path is correct as-is).
[...]

PR #1244, opened 2026-09-05
feat(api): derive Default on UncertaintyMethod and SimulateUncertaintyOptions [refs #529]

### LINKED ISSUE #529: Derive Default on public *Options structs so the R wrapper is immune to additive field changes
ferx-r constructs ferx-core options structs with **exhaustive** literals (e.g. `SimulateUncertaintyOptions` at `ferx-r/src/rust/src/lib.rs:621`). When ferx-core adds a field, those literals break with `error[E0063]: missing field` — exactly the cross-repo trap the CLAUDE.md sibling-repo note warns about. We just hit it for `SimulateOptions` (#522 → [FeRx-NLME/ferx-r#200](https://github.com/FeRx-NLME/ferx-r/pull/200)).

`SimulateOptions` already derives `Default`, so ferx-r can spread `..Default::default()` and be immune to additive field changes. `SimulateUncertaintyOptions` does **not** derive `Default` (because `UncertaintyMethod` doesn't), so it can't take that fix yet.

### Proposed
- Derive `Default` on `UncertaintyMethod` (`#[default] Asymptotic` — the standard MVN method) and on `SimulateUncertaintyOptions`.
- Update ferx-r `src/rust/src/lib.rs:621` to `..Default::default()`.
- **Convention:** every `pub` options struct intended for the R boundary should derive `Default`, so the wrapper can construct it with `..Default::default()` and survive additive field changes.

### Notes
- Low priority; **no current breakage** (`SimulateUncertaintyOptions` is unchanged by #522).
- Surfaced during the `/code-review` of ferx-r#200.

roninsightrx:
## Why

`ferx-r` constructs ferx-core options structs with **exhaustive** struct literals (`SimulateUncertaintyOptions` at `ferx-r/src/rust/src/lib.rs:906` on `main`). When ferx-core adds a field, those literals break with `error[E0063]: missing field` — exactly the cross-repo trap the CLAUDE.md sibling-repo note warns about. We hit it for `SimulateOptions` (#522 → FeRx-NLME/ferx-r#200).

`SimulateOptions` already derives `Default`, so ferx-r can spread `..Default::default()` and be immune to additive field changes. `SimulateUncertaintyOptions` could **not** take that fix, because `UncertaintyMethod` had no `Default`.

This PR makes the resilient form *available*. It does not by itself make the wrapper resilient — see **Cross-repo dependency** for the sequencing, and note the deliberate `Refs #529` rather than `closes`.

## What changed

- **`UncertaintyMethod` derives `Default`** with `#[default] Asymptotic`. That is the right default rather than an arbitrary one: `Asymptotic` needs only a successful covariance step, while `Sir` additionally requires the SIR resample pool to have been retained at fit time (`sir = true` **and** `sir_keep_samples = true`), so a spread-constructed value defaulting to `Sir` would fail at run time for most fits.
- **`SimulateUncertaintyOptions` derives `Default`.** Its two counts default to `0`, which draws nothing and returns an empty row set; the struct and field doc comments say so explicitly, since a caller spreading the default is expected to set both.
- **An inventory guard for the convention** — A4 in `tests/public_api_boundary.rs`. It scans every `.rs` file under `src/` and `crates/`, collects every `struct`/`enum` named `*Options` together with every hand-written `impl Default for …`, and fails naming any declaration covered by neither. It is **discovery-based, not a list**: a `*Options` type added later that forgets the trait fails without anyone remembering to register it.
- **Audit result:** `FitOptions`, `SaveFitOptions`, `OdeSolverOptions`, `SimulateOptions`, `AdaptiveSimulateOptions` (core) and `BootstrapOptions`, `GamOptions`, `RunOptions` (`ferx-tools`) all already implement `Default`, derived or manual — so the two types above were the only gap. The guard now holds that line going forward.

## Alternatives considered

**For the default value:** a hand-written `impl Default for SimulateUncertaintyOptions` with non-zero counts (e.g. 100 draws × 1 sim), so a bare `Default::default()` would do something useful rather than return no rows. Rejected — those numbers would be invented, and the struct's purpose here is the `..Default::default()` **spread**, where the caller always names the counts. Documenting the `0` is honest; a magic 100 is a silent policy the R wrapper would inherit.

**For the guard:** centralizing the declarations behind an enforcing macro (the review's other suggestion) would give a compile-time bound instead of a text scan, but it would rewrite eight type declarations across two crates to enforce a one-line convention, and a new options struct could still simply not use the macro — the same discovery hole one level up. The scan has no such hole: it is keyed on the type's *name*, which is the thing the convention is about.

## Cross-repo dependency

| ferx (R) | tracked as FeRx-NLME/ferx-r#330 (issue, no PR yet) | open | after |

**Sequencing, explicitly.** `ferx-r/src/rust/Cargo.lock` pins ferx-core by commit *from the `main` branch* — currently `ebc13a1e`. A lock bump therefore cannot point at a rev that exists only on this feature branch, so the ferx-r work is blocked on this PR merging. The order is:

1. **this PR merges to ferx-core `main`** → the derives exist at a rev on `main`;
2. `tools/update-ferx-core-lock.sh` in ferx-r bumps `src/rust/Cargo.lock` to that rev;
3. `ferx-r/src/rust/src/lib.rs:906` switches to `..Default::default()`;
4. that ferx-r PR merges — **only now is the wrapper actually immune.**

Because step 4 is what closes the issue's intent, this PR says `Refs #529`, not `closes #529` (in the title, the body, and both commit messages — GitHub's squash-merge body concatenates commit messages, so leaving the keyword in one of them would have auto-closed the issue on merge). #529 stays open until the ferx-r side lands. FeRx-NLME/ferx-r#330 carries the exact diff, the lock-bump command, and this same sequencing note.

Nothing is broken in the meantime: `SimulateUncertaintyOptions` is unchanged by #522, so ferx-r's current exhaustive literal keeps compiling. This is pre-emptive hardening against the *next* added field, which is why the split across two merges is safe rather than a half-landed change.

## Breaking changes

- [x] None

Additive only: two new trait impls, no signature or field changes. `tools/update-public-api.sh` was re-run and the committed baseline is **unchanged** — `cargo public-api` omits derived impls, which is why the already-derived `SimulateOptions` carries no `impl Default` row in the baseline either (`AdaptiveSimulateOptions` / `FitOptions` / `OdeSolverOptions`, whose `Default` is hand-written, do). `tools/update-public-api.sh --check` reports `public API matches the committed baseline.` That blind spot is precisely why the new convention guard is a source scan and lives next to the `#[doc(hidden)]` ban, which exists for the same reason.

<details>

Not breaking — no migration needed.

</details>

## Numerical validation

No numerical path is touched: this adds trait impls and a test. The NONMEM-comparison rule does not bite because nothing here produces a new estimate, OFV, residual or diagnostic. What *is* asserted numerically is that the new construction path is a no-op — `spread_default_matches_exhaustive_literal` requires bit-identical `ipred` and outcome values (`to_bits()`) between an exhaustive literal and its `..Default::default()` spread at the same seed.

## Performance

The new guard is a filesystem walk over `src/` and `crates/` with no compilation, and runs in the existing `public_api_boundary` binary: the whole file is 0.25 s.

## Tests

| Test | Location | Pins |
|---|---|---|
| `uncertainty_method_defaults_to_asymptotic` | `src/estimation/uncertainty_samples.rs` | the `#[default]` variant — moving it to `Sir` would make spread-constructed options demand a SIR pool most fits never retained |
| `simulate_uncertainty_options_default_values` | `src/api/tests/simulate_with_uncertainty_tests.rs` | every default field value a caller silently inherits for a field it does not name |
| `spread_default_matches_exhaustive_literal` | same | the point of #529 — spread and exhaustive constructions must produce identical simulations |
| `every_public_options_type_implements_default` | `tests/public_api_boundary.rs` (A4) | **the convention, by discovery** — every `*Options` declaration under `src/` and `crates/` implements `Default` |
| `options_scan_flags_a_new_struct_without_default` | same | that the guard above can actually fail |
| `derive_list_matches_default_exactly` | same | the derive parser: whole-trait match, multi-line derives, no false hit from a nested-paren attribute |
| `every_public_options_struct_implements_default` | `src/api/tests/…` + `crates/ferx-tools/src/lib.rs` | named trait-level spot checks (see below) |

**On "a green test is not evidence it can fail":**

- The A4 guard was mutation-tested **on the real tree, both directions**. Appending `#[derive(Debug, Clone)] pub struct MutationProbeOptions { pub a: usize }` to `src/lib.rs` fails it with `Missing at: ["src/lib.rs:75 (MutationProbeOptions)"]`; adding `Default` to that derive turns it green again; the probe was then reverted (`git status` clean, trees verified identical). The same flip is pinned permanently by `options_scan_flags_a_new_struct_without_default`, whose fixture differs by exactly the offending derive and which asserts the straddle explicitly, so the guard cannot silently become a tautology.
- The fixture also contains a doc comment that says "Derives `Default`" in prose above the offending struct, and asserts it does **not** rescue it — the scan drops doc comments from the attribute block for exactly that reason, since several of the real types describe their `Default` in prose.
- `spread_default_matches_exhaustive_literal` can fail too: if `#[default]` moved to `Sir`, the spread arm would take the SIR path against a fit with no retained pool and `unwrap()` would panic. It guards against a degenerate comparison first — the run must be the expected 36 rows, all finite, and not identically zero — before any equality is checked.

Results: `cargo test --lib --features ci` → **4039 passed, 0 failed**; `cargo test --test public_api_boundary --features ci` → **8 passed, 0 failed**; `cargo test -p ferx-tools --lib` → **300 passed, 0 failed**.

No `.ferx` DSL, `[fit_options]` key or output field changes — this is a Rust-API ergonomics change for wrapper authors. The usage it enables:

<details>

```rust
// Before: exhaustive — breaks with E0063 when ferx-core adds a field.
let opts = SimulateUncertaintyOptions {
    n_uncertainty_draws: 200,
    n_sim_per_draw: 1,
    method: UncertaintyMethod::Asymptotic,
    seed: Some(42),
};

// After: immune to additive field changes.
let opts = SimulateUncertaintyOptions {
    n_uncertainty_draws: 200,
    n_sim_per_draw: 1,
    seed: Some(42),
    ..Default::default()
};
```

</details>

## Docs

- [ ] No user-visible change

Added under `[Unreleased] → Added`, and reworded in the second commit so it describes the inventory guard rather than claiming more than the first commit's hand-written lists delivered. No `docs/**/*.qmd` page changed: the Quarto site documents the `.ferx` model file and fit behaviour, and nothing there describes these Rust struct literals.

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → `ferx-site/model-dsl/` page updated or PR opened — N/A, no DSL or fit-option change
- [x] New example `.ferx` file added to `examples/` → matching entry in ferx-site examples and ferx-book — N/A, no new example
- [x] New data file added to `data/` → mirrored to `ferx-r/inst/examples/data/` and referenced in site/book — N/A, no new data

### ferx-book
- [x] New estimator, DSL feature, or fit option → relevant book chapter updated or PR opened — N/A, none of those

No `examples/*.ferx` is affected — no model file, CLI surface or numerical output changes.

## Checklist

- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), public-API baseline. A green `cargo test` is not a green build: the feature sets are not nested (#1133, #1157)
- [x] `Cargo.toml` version bumped if breaking

`preflight OK — fmt check clippy rustdoc docs public-api`. Nothing on the `Dual2` path is touched (two trait impls on options types and a test, no numeric kernel), and no version bump is needed since nothing breaks.

## Reviewer hints

Two judgement calls worth a look:

1. **`#[default] Asymptotic`.** `Sir` has a precondition (`sir_keep_samples` at fit time) a defaulted value cannot guarantee, so it would turn a spread construction into a run-time error. `uncertainty_method_defaults_to_asymptotic` pins it.
2. **Counts defaulting to `0`.** A bare `SimulateUncertaintyOptions::default()` draws nothing and returns an empty row set (`draw_parameter_samples` short-circuits at `n_draws == 0`). Documented on the struct and both fields, pinned by `simulate_uncertainty_options_default_values`. See *Alternatives considered* for why not a manual `impl` with invented counts.

The new A4 guard is where the real review value is: it is textual, so the things to check are the parser's edges — multi-line derives, doc-comment prose, `DefaultDisplay`-style near-misses — all of which have their own assertions in `derive_list_matches_default_exactly`.

The second commit is the response to review; the first is the original change. Reviewing them separately is easier than the combined diff.

## Open questions

None blocking. The one open thread is the ferx-r follow-up (FeRx-NLME/ferx-r#330), which cannot start until this merges — see **Cross-repo dependency** for why the lock bump has to come second.

Refs #529 — intentionally non-closing; the issue stays open until FeRx-NLME/ferx-r#330 lands.

https://claude.ai/code/session_01Mk3MLZ59oLAZVx3uNLBLjz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,11 @@ section of the SDLC for the versioning policy).
 - **`Default` on `UncertaintyMethod` and `SimulateUncertaintyOptions` (#529).** The
   uncertainty-simulation options can now be built with `..Default::default()`
   (`UncertaintyMethod::Asymptotic` is the default method), so wrapper code stays
-  source-compatible when a field is added. Every other `pub` options struct in
-  `ferx-core` and `ferx-tools` already implemented `Default`; a test now pins that
-  convention.
+  source-compatible when a field is added. Every other `*Options` type in
+  `ferx-core` and `ferx-tools` already implemented `Default`; a new inventory
+  guard (`tests/public_api_boundary.rs`, A4) now discovers every `*Options`
+  declaration under `src/` and `crates/` and fails on one that lacks it, so the
+  convention holds for options types added later.
 - **`.ferxsearch` search configuration and an MFL search-space parser in `ferx-tools` (#1179).**
   A TOML file (`base`, `data`, `[space] mfl`, `[rank]`, `[strictness]`, `[run]`) whose space is
   written in Pharmpy's Model Feature Language, with `@IIV` / `@PK` / `@CONTINUOUS` / … resolved
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ section of the SDLC for the versioning policy).
 ## [Unreleased]
 
 ### Added
+- **`Default` on `UncertaintyMethod` and `SimulateUncertaintyOptions` (#529).** The
+  uncertainty-simulation options can now be built with `..Default::default()`
+  (`UncertaintyMethod::Asymptotic` is the default method), so wrapper code stays
+  source-compatible when a field is added. Every other `pub` options struct in
+  `ferx-core` and `ferx-tools` already implemented `Default`; a test now pins that
+  convention.
 - **`.ferxsearch` search configuration and an MFL search-space parser in `ferx-tools` (#1179).**
   A TOML file (`base`, `data`, `[space] mfl`, `[rank]`, `[strictness]`, `[run]`) whose space is
   written in Pharmpy's Model Feature Language, with `@IIV` / `@PK` / `@CONTINUOUS` / … resolved
--- a/crates/ferx-tools/src/lib.rs
+++ b/crates/ferx-tools/src/lib.rs
@@ -36,10 +36,17 @@ pub fn core_version() -> &'static str {
 mod tests {
     use super::core_version;
 
-    /// #529 convention guard, `ferx-tools` half: every `pub` options struct
-    /// this crate exposes must implement `Default` so a caller can build it
-    /// with `..Default::default()` and survive an added field. A new options
-    /// struct that forgets it fails to compile here.
+    /// #529, trait-level half for `ferx-tools`: this crate's options types
+    /// really do resolve `Default`, so a caller can build them with
+    /// `..Default::default()`.
+    ///
+    /// A **named spot check, not the inventory guard** — the list is
+    /// hand-written and a new options struct does not change it. Discovery for
+    /// the whole workspace (this crate included: the scan walks `crates/`) is
+    /// `every_public_options_type_implements_default` in `ferx-core`'s
+    /// `tests/public_api_boundary.rs` (A4). It lives there rather than here
+    /// because `ferx-core` must not depend on `ferx-tools`, and a source scan
+    /// needs no dependency in either direction.
     #[test]
     fn every_public_options_struct_implements_default() {
         fn assert_default<T: Default>() -> T {
--- a/crates/ferx-tools/src/lib.rs
+++ b/crates/ferx-tools/src/lib.rs
@@ -36,6 +36,20 @@ pub fn core_version() -> &'static str {
 mod tests {
     use super::core_version;
 
+    /// #529 convention guard, `ferx-tools` half: every `pub` options struct
+    /// this crate exposes must implement `Default` so a caller can build it
+    /// with `..Default::default()` and survive an added field. A new options
+    /// struct that forgets it fails to compile here.
+    #[test]
+    fn every_public_options_struct_implements_default() {
+        fn assert_default<T: Default>() -> T {
+            T::default()
+        }
+        let _ = assert_default::<crate::bootstrap::BootstrapOptions>();
+        let _ = assert_default::<crate::gam::GamOptions>();
+        let _ = assert_default::<crate::search::RunOptions>();
+    }
+
     #[test]
     fn core_version_is_reported_through_the_public_api() {
         let v = core_version();
--- a/src/api/simulate.rs
+++ b/src/api/simulate.rs
@@ -1010,11 +1010,19 @@ fn simulate_inner_with_draw<R: rand::Rng>(
 }
 
 /// Options controlling `simulate_with_uncertainty()`.
-#[derive(Debug, Clone)]
+///
+/// Derives `Default` so an out-of-crate caller (the R wrapper) can build one
+/// with `..Default::default()` and stay source-compatible when a field is
+/// added here (#529). Note that the two counts default to `0`, which draws
+/// nothing and returns an empty row set — a caller spreading the default is
+/// expected to set both explicitly.
+#[derive(Debug, Clone, Default)]
 pub struct SimulateUncertaintyOptions {
     /// Number of parameter sets to draw from the uncertainty distribution.
+    /// Defaults to `0`; set it explicitly when spreading `Default::default()`.
     pub n_uncertainty_draws: usize,
     /// Number of eta/eps replicates simulated *per* parameter draw.
+    /// Defaults to `0`; set it explicitly when spreading `Default::default()`.
     pub n_sim_per_draw: usize,
     /// How to draw the parameter sets — asymptotic MVN or SIR resamples.
     pub method: crate::estimation::uncertainty_samples::UncertaintyMethod,
--- a/src/api/tests/simulate_with_uncertainty_tests.rs
+++ b/src/api/tests/simulate_with_uncertainty_tests.rs
@@ -1605,10 +1605,18 @@ fn spread_default_matches_exhaustive_literal() {
     }
 }
 
-/// #529 convention guard: every `pub` options struct `ferx-core` exposes at
-/// the R boundary must implement `Default`, so the wrapper can spread it and
-/// survive an added field. A new one that forgets the derive fails to compile
-/// here rather than breaking `ferx-r`'s build after the merge.
+/// #529, trait-level half: the options types `ferx-core` exposes at the R
+/// boundary really do resolve `Default`, so the wrapper can spread them.
+///
+/// This is a **named spot check, not the inventory guard** — the list is
+/// hand-written, so adding a new options struct does not change it and this
+/// test would stay green. Discovery is
+/// `every_public_options_type_implements_default` in
+/// `tests/public_api_boundary.rs` (A4), which scans `src/` and `crates/` for
+/// every `*Options` declaration and fails on one lacking `Default`. The two
+/// are complementary: the scan is textual and sees types nobody registered,
+/// this one is a real trait bound and would catch a `Default` that exists in
+/// source but does not apply to the type (a `cfg`-ed out impl, say).
 #[test]
 fn every_public_options_struct_implements_default() {
     fn assert_default<T: Default>() -> T {
--- a/src/api/tests/simulate_with_uncertainty_tests.rs
+++ b/src/api/tests/simulate_with_uncertainty_tests.rs
@@ -1539,3 +1539,85 @@ fn asymptotic_errors_without_covariance_step() {
     let err = simulate_with_uncertainty(&model, &pop, &fit, &opts).unwrap_err();
     assert!(err.contains("covariance"));
 }
+
+// ---------------------------------------------------------------------------
+// #529 — `Default` at the R boundary
+// ---------------------------------------------------------------------------
+
+/// Pin every field of the derived `Default`. The R wrapper builds this struct
+/// with `..Default::default()`, so these values are the ones a field it does
+/// *not* name silently takes.
+#[test]
+fn simulate_uncertainty_options_default_values() {
+    let d = SimulateUncertaintyOptions::default();
+    assert_eq!(d.n_uncertainty_draws, 0);
+    assert_eq!(d.n_sim_per_draw, 0);
+    assert_eq!(d.method, UncertaintyMethod::Asymptotic);
+    assert_eq!(d.seed, None);
+}
+
+/// The point of #529: a spread construction that names only the fields the
+/// caller cares about must behave identically to the exhaustive literal it
+/// replaces. Run both through `simulate_with_uncertainty` with the same seed
+/// and require bit-identical predictions — an oracle that fails if a future
+/// field's default silently changes the simulation rather than merely
+/// compiling.
+#[test]
+fn spread_default_matches_exhaustive_literal() {
+    let model = tiny_model();
+    let pop = tiny_population();
+    let fit = synthetic_fit(&model.default_params);
+
+    let exhaustive = SimulateUncertaintyOptions {
+        n_uncertainty_draws: 3,
+        n_sim_per_draw: 2,
+        method: UncertaintyMethod::Asymptotic,
+        seed: Some(42),
+    };
+    let spread = SimulateUncertaintyOptions {
+        n_uncertainty_draws: 3,
+        n_sim_per_draw: 2,
+        seed: Some(42),
+        ..Default::default()
+    };
+
+    let a = simulate_with_uncertainty(&model, &pop, &fit, &exhaustive).unwrap();
+    let b = simulate_with_uncertainty(&model, &pop, &fit, &spread).unwrap();
+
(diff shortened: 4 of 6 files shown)


roninsightrx (commented):
Review findings:

1. **[P2] Complete the ferx-r side before closing #529.** The issue is specifically about making the wrapper immune to additive fields, but `ferx-r/src/rust/src/lib.rs` still constructs `SimulateUncertaintyOptions` exhaustively (currently around line 900), and there is no open ferx-r follow-up. Merging this PR only makes the resilient syntax available; it does not make the wrapper resilient, so the next added field still produces E0063 while this PR closes the tracking issue. Please open/link the companion ferx-r PR that switches this literal to `..Default::default()` and bumps the pinned core lock via `tools/update-ferx-core-lock.sh` (or keep #529 open and accurately mark the cross-repo dependency).

2. **[P2] Make the options-convention guard discover new types.** The tests at `src/api/tests/simulate_with_uncertainty_tests.rs:1613` and `crates/ferx-tools/src/lib.rs:44` only assert `Default` for a hand-written list of current types. Adding `pub struct NewOptions` without `Default` does not alter either list, so both tests remain green; contrary to the comments and changelog, a new options struct that forgets the trait does not fail here. Please either add an inventory-based check that discovers all public `*Options` types (or centralize the declarations through an enforcing macro), or describe these tests as a one-time audit rather than a CI convention guard.

Validation performed: `git diff --check` passed; all four added focused tests passed locally; current CI checks are green. The `Default` implementation itself and the `Asymptotic` default look correct.

roninsightrx:
Both findings verified against the code before acting. Both are correct — neither was a false positive, so nothing to push back on. Addressed in `53f73d9`, plus a rewrite of the two commit messages (details under finding 1).

---

## 1 — [P2] ferx-r side is not done, yet this closed #529

**Verified, and correct.** Checked `ferx-r` `main` directly rather than the local sibling checkout, which can drift:

```
$ gh api repos/FeRx-NLME/ferx-r/contents/src/rust/src/lib.rs --jq .content | base64 -d | grep -n "SimulateUncertaintyOptions" -A 6
906:    let opts = ferx_core::SimulateUncertaintyOptions {
907-        n_uncertainty_draws: n_uncertainty_draws.max(0) as usize,
908-        n_sim_per_draw: n_sim_per_draw.max(0) as usize,
909-        method: uncertainty_method,
910-        seed: Some(seed as u64),
911-    };
```

Still exhaustive. So merging this alone leaves the wrapper exposed to the next added field, while `closes #529` would have retired the tracking issue. The finding stands.

**Action: took the second option — tracked follow-up plus a non-closing reference.**

- Filed **FeRx-NLME/ferx-r#330**, with the exact diff, the `tools/update-ferx-core-lock.sh` command, and the sequencing constraint.
- Changed `closes #529` → `Refs #529` in the PR **title**, the PR **body**, and **both commit messages**.

That last one is why the branch was force-pushed (`--force-with-lease`; trees verified byte-identical before and after, `git rev-parse` on both). GitHub's squash-merge body concatenates the individual commit messages, so leaving `[closes #529]` in commit `cd2d412e` would have auto-closed the issue on merge no matter what the title said. Fixing the title alone would not have been enough.

**Sequencing, explicitly.** `ferx-r/src/rust/Cargo.lock` pins ferx-core **by commit, from the `main` branch** — currently:

```
name = "ferx-core"
source = "git+https://github.com/FeRx-NLME/ferx-core?branch=main#ebc13a1e50efb42d1676c9a2d3a2ec0695e06175"
```

A lock bump cannot name a rev that exists only on this feature branch, so the ferx-r work is **blocked on this PR merging first**. Order:

1. this PR merges to ferx-core `main` → the derives exist at a rev on `main`;
2. `tools/update-ferx-core-lock.sh` in ferx-r bumps the lock to that rev;
3. `lib.rs:906` switches to `..Default::default()`;
4. that ferx-r PR merges — **only then is the wrapper immune**, and only then should #529 close.

I did not open a companion ferx-r PR now on purpose: it could not be green. Without step 2 the `..Default::default()` does not compile against the locked core, and step 2 cannot be done yet. A red draft PR would be worse tracking than the issue.

Worth noting the split is safe rather than half-landed: `SimulateUncertaintyOptions` is unchanged by #522, so ferx-r's current literal keeps compiling. This is hardening against the *next* field, not a fix for a live break.

---

## 2 — [P2] The convention guard did not discover new types

**Verified, and correct.** Both tests asserted `Default` for a hand-written list, so `pub struct NewOptions` without the trait changed neither list and both stayed green. The doc comments ("A new one that forgets the derive fails to compile here") and the CHANGELOG claimed a guard that did not exist. That was an overclaim, and the wording was mine to fix.

**Action: built the inventory check** — `every_public_options_type_implements_default`, A4 in `tests/public_api_boundary.rs`.

It walks every `.rs` file under `src/` and `crates/`, collects every `struct`/`enum` whose name ends in `Options` along with every hand-written `impl Default for …`, and fails naming any declaration covered by neither derive nor manual impl. Discovery is keyed on the type's **name**, which is what the convention is actually about, so nothing needs registering.

It went into `tests/public_api_boundary.rs` rather than a new file because that is already the home for invariants the committed baseline cannot see — and this is exactly that case, for the same mechanical reason as the `#[doc(hidden)]` ban: `cargo public-api` omits *derived* impls, so `api/ferx-core-public-api.txt` has no `impl Default` row for `SimulateOptions` at all, only for the hand-written ones (`FitOptions`, `OdeSolverOptions`, `AdaptiveSimulateOptions`). A baseline-driven check was therefore not an option. It also has to be a source scan rather than a trait bound for a layering reason: `ferx-core` must not depend on `ferx-tools` (A1, pinned in the same file), so no test in core can name `BootstrapOptions` as a type — but it can read its source.

**On the enforcing-macro alternative you offered:** rejected deliberately. It would rewrite eight declarations across two crates to enforce a one-line convention, and a new options struct could simply not use the macro — the same discovery hole moved up a level. The scan has no such hole.

**Mutation-checked, on the real tree, both directions** — as requested, not just asserted:

```
$ printf '\n#[derive(Debug, Clone)]\npub struct MutationProbeOptions {\n    pub a: usize,\n}\n' >> src/lib.rs
$ cargo test --test public_api_boundary --features ci every_public_options
test every_public_options_type_implements_default ... FAILED

every `pub *Options` type must implement `Default` so a wrapper can build it with
`..Default::default()` and survive an added field (#529). ...
Missing at: ["src/lib.rs:75 (MutationProbeOptions)"]

$ # same struct, `Default` added to the derive:
test every_public_options_type_implements_default ... ok
```

Probe reverted afterwards; `git status` clean. The flip is pinned permanently by `options_scan_flags_a_new_struct_without_default`, whose two fixtures differ by exactly the offending derive and which asserts the straddle in both directions, so the guard cannot quietly become a tautology later.

Three parser edges have their own assertions, since a textual scan is only as good as what it refuses to match:

- **Prose does not rescue code.** The fixture puts a doc comment reading "Derives `Default`" directly above the offending struct and asserts it is still flagged — doc comments are dropped from the attribute block on purpose, because several of the real types describe their `Default` in prose and a naive substring scan would let a comment satisfy a check about code.
- **Whole-trait match.** `#[derive(DefaultDisplay)]` does not count (`derive_list_matches_default_exactly`).
- **Multi-line derives** are read whole via paren depth, and a nested-paren attribute like `#[serde(rename_all(x))]` between two derives does not end the scan early.

**The two typed tests stay**, reworded to what they actually are: named trait-level spot checks, complementary rather than redundant — a real trait bound catches a `Default` that exists in source but does not apply to the type (a `cfg`-ed out impl, say), which the textual scan cannot see. Their comments now say "a named spot check, not the inventory guard" and point at A4. The CHANGELOG entry was reworded the same way.

---

### Verification

`tools/preflight.sh` → `preflight OK — fmt check clippy rustdoc docs public-api`
`cargo test --lib --features ci` → 4039 passed, 0 failed
`cargo test --test public_api_boundary --features ci` → 8 passed, 0 failed
`cargo test -p ferx-tools --lib` → 300 passed, 0 failed

Thanks — finding 2 in particular was a real hole in work I had described as a guard.

https://claude.ai/code/session_01Mk3MLZ59oLAZVx3uNLBLjz

PR #342, opened 2026-06-14
fix(parser): don't fire compute_max_stack end-depth assert on conditional bytecode

TeunP:
## Why
`compute_max_stack`'s final `debug_assert!(depth == 1)` wrongly fires for any expression containing a conditional (`if` / ternary). A `Conditional` compiles to `cond; JumpIfFalse(else); then; Jump(end); else`; the helper's **linear** scan sums per-op stack deltas and so counts **both** arms' pushes (+1 each) even though exactly one executes at runtime — ending at depth 2 (more when nested).

This is **debug-only** and pre-existing on `main` (unrelated to any feature work). CI never caught it because the test job uses `--profile ci-test`, which inherits from `release` (debug-assertions off). Under the default dev/test profile four parser tests panic: `bytecode_matches_ast_on_conditional_and_logic`, `test_inline_if_in_parameter_assignment`, `test_ode_rhs_defined_names_ok`, `test_parse_all_example_ferx_files`.

Repro: `cargo test --lib --no-default-features --features ci test_ode_rhs_defined_names_ok` → panics `compute_max_stack: bytecode ends at depth 2, expected 1`.

## What changed
Skip the end-depth assertion when the bytecode contains any `Jump`/`JumpIfFalse`. The off-by-one guard is **retained** for jump-free expressions (the only case where a linear end-depth is meaningful). `peak` — the returned stack size — is unchanged: it is a safe over-estimate for branches (the linear depth dominates whichever single arm runs, so the VM stack is never under-allocated), which is why release has always been correct.

## Alternatives considered
Tracking branch joins to compute an exact end-depth (more machinery for a guard that only needs to catch jump-free off-by-ones); or deleting the guard (loses real protection). Skipping it precisely when jumps are present keeps the protection where it is valid, with no false positives.

## Cross-repo dependency
| Repo | PR | Status | Must merge |
|------|----|--------|------------|
| ferx (R) | — | not needed | — |

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable (parser / debug-assertion correctness; no numerical output changes)

## Performance
- [x] No significant impact expected (the added jump scan lives inside `debug_assert!` and is compiled out in release)

## Tests
- [x] Regression test added that fails without this fix (`compute_max_stack_allows_branching_bytecode` — compiles conditional + nested-conditional bytecode, asserts a sufficient size with no panic, and that the jump-free path is unchanged)
[...]

PR #1245, opened 2026-09-05
ci: run the suite with debug-assertions enabled [refs #344]

### LINKED ISSUE #344: CI gap: ci-test profile disables debug-assertions, hiding debug_assert! guards from CI

roninsightrx:
## Why

Every test job in `ci.yml` builds with `--profile ci-test` or `--profile ci-fast`, and both `inherits = "release"` — so `debug-assertions` is **off** and all **177 `debug_assert!`** guards in the crate compile to nothing. They are invariant checks CI has never executed.

That is not theoretical. An off-by-one in `compute_max_stack` (`src/parser/model_parser.rs`) tripped its own `debug_assert!` for any expression containing an `if`/ternary, and four lib tests panicked under the dev profile while CI stayed green through the whole thing:

- `parser::model_parser::tests::bytecode_matches_ast_on_conditional_and_logic`
- `parser::model_parser::tests::test_inline_if_in_parameter_assignment`
- `parser::model_parser::tests::test_ode_rhs_defined_names_ok`
- `parser::model_parser::tests::test_parse_all_example_ferx_files`

The bug was fixed in #342. The *class* of problem — any future invariant violation guarded by `debug_assert!` being invisible to CI — is what this addresses.

## What changed

**A new `Tests (debug-assertions)` job**, not a flipped profile. `ci-test` exists so the population fits run at release speed (`Cargo.toml`'s comment says so); flipping it would slow every other test job.

The job builds on the **default `dev` profile** — no `--profile` flag — deliberately:

- it is the profile the #342 bug actually reproduced under, so the gate is the literal repro rather than an approximation of it;
- it also turns on `overflow-checks`, which `release` leaves off, so integer overflow panics instead of wrapping silently — a second class of guard no other job can see;
- a `ci-test`-with-`debug-assertions` profile would be a distinct profile hash and therefore a whole extra **optimised** build of the lib. This repo is compile-bound (93% LLVM, #969/#971), so one extra *unoptimised* build is much cheaper than a second optimised one.

**Two feature sets**, for the same non-nesting reason `check` is a matrix rather than one line: `--features ci` compiles neither `src/survival`, `src/markov`, `src/categorical` nor `src/nn`, and all four carry `debug_assert!`s. `markov` implies `survival`, so `ci,markov,nn` is the union of the production cfgs (same set `clippy` and `rustdoc` take).

`--lib` only: Tier-1 is where the guards live (`debug_assert!` sits next to the invariant, in `src/`), and the `tests/` binaries are already run by `Tests + coverage (core)`.

**Wired through `tools/preflight.sh`** (#1157) as the `debug-assertions` group, so the local gate and the CI gate stay one list. It is the **first opt-in group**: it runs a test suite rather than a compile, so folding it into a no-argument run would turn `tools/preflight.sh` from a seconds-long pre-push habit into a coffee break. The default selection is now derived as `ALL_GROUPS` minus `OPT_IN_GROUPS`, so the two lists cannot drift.

```bash
tools/preflight.sh debug-assertions   # what the new CI job runs
```

**A canary, because "no `--profile` flag" is an invariant held by absence.** Nothing in `cargo test --lib --features ci` distinguishes a build with the guards live from one without, so a `[profile.dev] debug-assertions = false` in `Cargo.toml` or a `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` in the workflow environment would neuter the entire job with every test still green — the gate back to being the no-op it was filed to replace. `debug_assertion_canary` (`src/lib.rs`) closes that: it records whether a `debug_assert!` actually **evaluated its condition** (the macro, not `cfg!(debug_assertions)` — one indirection closer to the property that matters) and fails the run when the guards turn out to be dead.

It is armed opt-in, via `env FERX_REQUIRE_DEBUG_ASSERTIONS=1` in the group's own argument vector — the `env` form so `run` echoes it and `--list` tells the truth, the same reasoning `group_rustdoc` uses for `RUSTDOCFLAGS`. Opt-in because every *other* job legitimately runs with the guards off; an unconditional assertion would fail `Tests + coverage (core)` and both coverage jobs for doing exactly what they are supposed to do.

**Triage (the issue's caveat).** Turning the guards on surfaced **no latent failure**. Measured locally on the dev profile:

| feature set | tests | result | wall clock |
|---|---|---|---|
| `ci` | 4035 | **ok**, 0 failed, 23 ignored | 80s |
| `ci,markov,nn` | 4366 | **ok**, 0 failed, 23 ignored | 198s |

So the job starts green; nothing needed fixing or documenting as a known failure.

## Alternatives considered

- **Flip `debug-assertions = true` on `ci-test`.** Rejected by the issue and on the merits: that profile exists for fit speed, and every job that uses it would pay.
- **A `ci-test`-like profile with `debug-assertions = true`** (the issue's second option). It gives fast tests, at the price of a second full **optimised** lib build under its own profile hash. The dev profile wins because it **fits inside the wall-clock envelope the existing test jobs already set**, and it gets `overflow-checks` for free.

  > **Corrected after the first CI run.** This bullet originally argued the dev profile wins because "the tests are the cheap side in a compile-bound repo". On a 2-core runner that is backwards. From the job log: `finished in 396.49s` (`ci`) and `finished in 1223.02s` (`ci,markov,nn`) — **27 minutes of tests** against roughly 3 minutes of compile, the reverse of the local measurement (80s + 198s on a warm 10-core machine). I should have labelled those numbers local-only rather than generalising from them.
  >
  > The decision stands, for a different reason than the one first given: at 29m57s the job runs in parallel with `Tests + coverage (core)` (28m26s) and `Tests + coverage (TTE/CTMM endpoints)` (24m32s), so it does not extend total CI wall clock. And the `ci-fast`-plus-assertions variant trades ~24 min of unoptimised tests for an optimised lib build those same jobs show costs ~20 min — close to a wash, while giving up `overflow-checks` and the literal-repro property.
  >
  > **Consequence for review:** this job is now roughly the critical path rather than comfortably inside it. If the Tier-1 suite grows much, it is the first job to push total CI time up, and the `ci-fast`-plus-assertions variant becomes the better trade at that point. The 29m57s is also a cold-cache figure, and only the ~3 min compile half is cacheable.
- **Fold the group into the default `preflight.sh` run.** Rejected: the script's contract is seconds of compile/lint, and a multi-minute suite in the default set is how a pre-push gate stops being run at all.
- **`--tests` instead of `--lib`.** Rejected: compiling ~120 more binaries under a second profile for guards that overwhelmingly live in `src/`.

## Cross-repo dependency

CI-only; no `pub` surface, no `.ferx` format, no `FitResult` change.

## Breaking changes
- [x] None

## Numerical validation

No numerical path is touched — this changes which compiler flags the existing tests run under.

## Performance

No change to any shipped profile. CI gains one job: **29m57s** cold cache on this PR's first run, **21m59s** warm on the latest — against `Tests + coverage (core)` at 17m25s and `Tests + coverage (TTE/CTMM endpoints)` at 16m34s on that same warm run. It runs in parallel with them, so it does not extend total CI wall clock, but it is now the longest single job. See the correction under *Alternatives considered*.

## Tests

`tests/preflight_owns_the_fast_gates.rs` — the file that already pins the preflight/CI contract — grows to cover the new group. It reads the group list **out of the script** (`script_array("ALL_GROUPS")` / `("OPT_IN_GROUPS")`) rather than keeping a second copy, so a future group is covered the moment it is added:

1. the new job must delegate to `tools/preflight.sh debug-assertions`, run no inline `cargo`, and neither skip itself (`if:`) nor tolerate failure (`continue-on-error`);
2. the script must exit non-zero when either of the group's two commands fails (the existing fake-`cargo` position sweep now includes it);
3. **new test** — a no-argument run must equal `ALL_GROUPS` minus `OPT_IN_GROUPS`, *and* every opt-in group must be named by a job in `ci.yml`. Both halves matter and fail in opposite directions: drop a group from the default set without declaring it opt-in and a developer silently under-checks; declare one opt-in with no CI job and the gate runs **nowhere** — strictly worse than the blind spot #344 is about, where the guards were at least compiled out honestly;
4. **new** — no command in the group may pass `--profile`/`--release`, and each must be a `cargo test`. Every named profile in this repo inherits `release`, so such a flag would leave two commands in the list, still run 4000-odd tests, still take minutes, and compile every `debug_assert!` back down to nothing. That is a neutered gate the command list cannot see, only its flags can;
5. **new** — every command in the group must **arm the canary** (`env FERX_REQUIRE_DEBUG_ASSERTIONS=1` leading the argv). This is the positive half that (4) cannot supply: (4) only rules out the ways of disabling the guards that are *visible in the command string*, and the two most likely ones are not.

Per the "a green test is not evidence that it can fail" rule, each assertion was **mutation-checked**, and all four mutations turn the guard red:

| mutation | test that failed |
|---|---|
| delete `run: tools/preflight.sh debug-assertions` from `ci.yml` | `the_fast_gate_jobs_delegate_to_preflight_and_never_inline_cargo` **and** `the_default_run_is_every_group_that_is_not_opt_in_and_ci_runs_the_rest` |
| drop the `ci,markov,nn` line from the group | `preflight_is_executable_and_lists_every_group` (count tripwire) |
| add `--profile ci-test` to a group command | `load_bearing_flags_and_feature_coverage_survive_in_the_command_list` |
| fail either command position behind a fake `cargo` | `a_failing_gate_fails_the_script_from_every_position` |
| `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false`, armed | `debug_assertion_canary` — and it passes **unarmed** under the same setting, which is what keeps the three release-profile jobs green |
| drop the `env` arming from a group command | `load_bearing_flags_and_feature_coverage_survive_in_the_command_list` |
| replace the `run:` line with a comment naming the same command | the opt-in check — and the **previous** `yml.contains` version passed it, which is how finding 3 was confirmed rather than assumed |

The opt-in check (3) reads the workflow's parsed `run:` steps, not its raw text. An earlier revision used `yml.contains`, which a YAML *comment* naming the command would satisfy while nothing executed the group — and this workflow is full of comments that name exactly these commands, so the check was one edit away from being self-satisfying. Demonstrated, not assumed: with the `run:` line replaced by a comment carrying the same string, the `contains` version passes and the parsed version fails.

One further trap worth naming: the count tripwire and the `--help` coverage test both used to enumerate groups from a bare `--list`. Since a no-argument run now walks only the default set, that would have walked straight past the new group and left it gated by nothing — the exact silent-shrinkage hole those tests exist to close. Both now read `ALL_GROUPS` from the script.

## Docs
- [ ] `CHANGELOG.md` `[Unreleased]` entry added (user-facing change), or N/A (internal/refactor/CI)
- [x] No user-visible change

`docs/development/sdlc.qmd` §8.2: the new job is added to the `ci.yml` row, with a short "Why a separate `Tests (debug-assertions)` job" note, the local command, how the canary is armed, and an explicit statement of what this does **not** fix (the Codecov ceiling, #1248); §8.6's "the three coverage jobs override `RUSTUP_TOOLCHAIN` to stable" now also names this job. No new page, so no `_quarto.yml` entry. No `CHANGELOG.md` entry — CI-only, per `CLAUDE.md`. `cargo run -p docs-lint -- --check` is green.

## Checklist
- [x] `tools/preflight.sh` green — `preflight OK — fmt check clippy rustdoc docs public-api`, plus `tools/preflight.sh debug-assertions` green (4035 and 4366 tests)
- [x] Functions on the `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay differentiable — none touched

### ferx-site

None apply — no DSL, example or data change.

### ferx-book

None apply.

## Reviewer hints

Focus on two judgment calls:

1. **The dev profile rather than a `ci-test`-with-assertions profile.** The trade is unoptimised tests (80s + 198s, measured) against a second optimised lib build. In a repo where 93% of the build is LLVM, the tests are the cheap side — and the dev profile throws in `overflow-checks`. If you would rather pay the compile for faster tests, that is the line to push back on.
2. **Making it opt-in locally.** This is the first group a no-argument `tools/preflight.sh` skips, which is a real (small) weakening of "one command reproduces CI". The mitigation is invariant (3): an opt-in group that no `ci.yml` job names is a test failure, so it cannot become a gate that runs nowhere. `--help` and `--list <group>` both document it.
3. **`refs #344` rather than `closes #344`.** The Codecov half of that issue is untouched here and is now #1248. If you would rather #344 be closed by this PR and the coverage work tracked purely as new scope, say so and I will flip the keyword — but the two halves are separable and the second one is not fixed.

The rest is mechanical: the guard test now reads the group list from the script instead of hard-coding it, so much of its diff is that substitution.

Skimmable: the comment blocks. They restate the rationale above.

## Open questions

- The job runs `--lib` only. Integration tests in `tests/` also reach `debug_assert!`s in `src/`, but adding `--tests` means compiling ~120 more binaries under a second profile. Worth it later if a guard is ever found that only a Tier-2 path trips?
- `slow-tests.yml` is still entirely release-profile, so Tier-3 fits never evaluate a guard either. Deliberately out of scope here (a nightly dev-profile convergence run would be very slow), but it is a remaining corner of the same blind spot.
- The canary proves `debug_assert!` is live. It does **not** prove `overflow-checks` is, which this PR also claims as a benefit of the dev profile — `cfg!(overflow_checks)` is still unstable (rust-lang/rust#111466, verified against this toolchain), and the runtime alternative is a `catch_unwind` around a deliberate overflow, which means swapping the global panic hook while other tests run concurrently. Judged not worth the hazard for a secondary claim; happy to add it if you disagree.

Refs #344 — **deliberately non-closing.** The issue's 2026-08-31 comment documents a second failure mode this PR does not touch: `debug_assert!` condition lines are permanently zero-hit regions in every Codecov job, because all of them use release-derived profiles, so the patch-coverage ceiling is unchanged by a job that uploads no coverage. Filed as #1248 with the measurements and the options. Closing #344 here would have buried a defect that remains.

https://claude.ai/code/session_01RRvJ1r8YxQ18UfcxrdKGhT

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
   # would be a no-op again — silently, with every test green — if a
   # `[profile.dev] debug-assertions = false` or a
   # `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` ever appeared. So the group's
-  # commands arm `debug_assertion_canary` (`src/lib.rs`) via
+  # commands arm `debug_assertion_canary` (`src/lib_tests.rs`) via
   # `env FERX_REQUIRE_DEBUG_ASSERTIONS=1`, and that canary fails the run if
   # `debug_assert!` turns out to compile to nothing. Do NOT set that variable at
   # the workflow or job level: every other job legitimately runs with the guards
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,16 @@ jobs:
   # and the CI gate cannot drift. It is the one group that is opt-in locally: it
   # runs a suite rather than a compile, so a no-argument `tools/preflight.sh`
   # stays a pre-push habit and this job is where it has to be green.
+  #
+  # Note that "no `--profile` flag" is an invariant held by ABSENCE, and this job
+  # would be a no-op again — silently, with every test green — if a
+  # `[profile.dev] debug-assertions = false` or a
+  # `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` ever appeared. So the group's
+  # commands arm `debug_assertion_canary` (`src/lib.rs`) via
+  # `env FERX_REQUIRE_DEBUG_ASSERTIONS=1`, and that canary fails the run if
+  # `debug_assert!` turns out to compile to nothing. Do NOT set that variable at
+  # the workflow or job level: every other job legitimately runs with the guards
+  # off, and arming it there would fail them for doing the right thing.
   debug-assertions:
     name: Tests (debug-assertions)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,54 @@ jobs:
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  # #344. Every other test job in this file builds with `--profile ci-test` or
+  # `--profile ci-fast`, and both `inherits = "release"` — so `debug-assertions`
+  # is OFF and all 177 `debug_assert!` guards in the crate compile to nothing.
+  # They are invariant checks that CI was not running.
+  #
+  # That was not theoretical. An off-by-one in `compute_max_stack`
+  # (`src/parser/model_parser.rs`) tripped its own `debug_assert!` for any
+  # expression containing an `if`, and four lib tests panicked under the dev
+  # profile while CI stayed green through the whole thing (fixed in #342; this
+  # job closes the blind spot that hid it).
+  #
+  # Deliberately a SEPARATE job rather than flipping `ci-test`: that profile
+  # exists so the population-fit tests run at release speed, and this repo is
+  # compile-bound (93% LLVM, #969/#971), so the cheap way to get the guards
+  # executed is one extra UNOPTIMISED build — not a second optimised one.
+  # Measured locally on the dev profile: 80s of tests for `--features ci` (4035
+  # tests) and 198s for `--features ci,markov,nn` (4366), all green — turning the
+  # guards on surfaced no latent failure, so this job starts green.
+  #
+  # The commands live in `tools/preflight.sh` (group `debug-assertions`) for the
+  # same reason as `Check`/`Clippy`/`Format` (#1157) — one list, so the local gate
+  # and the CI gate cannot drift. It is the one group that is opt-in locally: it
+  # runs a suite rather than a compile, so a no-argument `tools/preflight.sh`
+  # stays a pre-push habit and this job is where it has to be green.
+  debug-assertions:
+    name: Tests (debug-assertions)
+    runs-on: ubuntu-latest
+    # Stable, not the workflow-level nightly: `--features ci,markov,nn` has no
+    # nightly-only code, and stable's rustc holds for ~6 weeks against nightly's
+    # daily bump — `rust-cache` keys on `rustc -vV`, so nightly here would cold-
+    # rebuild the dep graph every day for no gate. Same reasoning as the coverage
+    # jobs. `tools/preflight.sh` honours an inherited `RUSTUP_TOOLCHAIN`.
+    env:
+      RUSTUP_TOOLCHAIN: stable
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install stable toolchain
+        run: rustup toolchain install stable --profile minimal
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          # Own key: the dev profile shares no artifacts with the `ci` (check /
+          # clippy) caches or with the instrumented coverage ones.
+          shared-key: ci-debug-assertions
+
+      - name: cargo test --lib with debug-assertions on (via tools/preflight.sh)
+        run: tools/preflight.sh debug-assertions
+
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
--- a/docs/development/sdlc.qmd
+++ b/docs/development/sdlc.qmd
@@ -265,13 +265,32 @@ after each major feature) and not yet on GitHub — a known gap to formalize.
 
 | Workflow | Trigger | Jobs |
 |----------|---------|------|
-| `ci.yml` | PR + push to `main` | `Check` (compile-only, incl. the `survival` / `markov` / `slow-tests` feature combos), `Tests + coverage (core)` (`--tests --features ci`, instrumented — runs the Tier-1/2 suite *and* produces the patch-coverage report), `Tests + coverage (TTE/CTMM endpoints)` (`--features ci,markov`, only the endpoint test binaries), `Clippy`, `Format`, `Public API baseline` (diffs `api/ferx-core-public-api.txt`); **`Tests + coverage (full, slow-tests)`** on weekly schedule / manual |
+| `ci.yml` | PR + push to `main` | `Check` (compile-only, incl. the `survival` / `markov` / `slow-tests` feature combos), `Tests + coverage (core)` (`--tests --features ci`, instrumented — runs the Tier-1/2 suite *and* produces the patch-coverage report), `Tests + coverage (TTE/CTMM endpoints)` (`--features ci,markov`, only the endpoint test binaries), `Tests (debug-assertions)` (`--lib` on the **dev** profile, so `debug_assert!` guards actually run — see below), `Clippy`, `Format`, `Public API baseline` (diffs `api/ferx-core-public-api.txt`); **`Tests + coverage (full, slow-tests)`** on weekly schedule / manual |
 | `slow-tests.yml` | nightly 06:00 UTC, manual, push to `main` touching estimation paths | Tier-3 fits to convergence (`--features ci,slow-tests --release`, `--no-fail-fast`) |
 | `docs.yml` | push to `main` touching `docs/**` | `quarto render` → deploy to `gh-pages` |
 | `release.yml` | tag `v*` or manual | GitHub release with generated notes |
 
 > Note: `RAYON_NUM_THREADS=1` is pinned in CI so cargo owns the test
 > parallelism on 2-core runners. Don't remove it.
+
+**Why a separate `Tests (debug-assertions)` job.** Every other test job builds
+with `--profile ci-test` or `--profile ci-fast`, and both `inherits = "release"`
+— so `debug-assertions` is off and each of the ~180 `debug_assert!` guards in the
+crate compiles to nothing. They were invariant checks CI never ran: an
+off-by-one in `compute_max_stack` tripped its own `debug_assert!` for any
+expression containing an `if`, and four lib tests panicked under the dev profile
+while CI stayed green (fixed in #342; the blind spot in #344). The fix is a
+separate job, not a flipped profile — `ci-test` exists so the fits run at release
+speed. Its commands live in `tools/preflight.sh` under the `debug-assertions`
+group; run them locally with
+
+```bash
+tools/preflight.sh debug-assertions
+```
+
+That group is the one `tools/preflight.sh` skips when you pass no arguments: it
+runs a test suite rather than a compile, so it is minutes rather than seconds.
+Name it when you want it.
 
 ### 8.3 CI workflows (ferx-r)
 
@@ -331,7 +350,8 @@ FOCE/FOCEI and HMC gradients come from hand-rolled analytic `Dual2` sensitivitie
 (`sens/`), and models outside the provider's scope fall back to finite differences.
 
 In fact there is no nightly-only code left, which is why the three coverage jobs
-in `ci.yml` override `RUSTUP_TOOLCHAIN` to **stable**: stable's rustc version
+in `ci.yml` — and `Tests (debug-assertions)` — override `RUSTUP_TOOLCHAIN` to
+**stable**: stable's rustc version
 holds for ~6 weeks against nightly's daily bump, keeping their instrumented build
 cache warm. `rust-toolchain.toml` still pins nightly as the default, and `Check` /
 `Clippy` / `Format` run on it.
--- a/docs/development/sdlc.qmd
+++ b/docs/development/sdlc.qmd
@@ -292,6 +292,18 @@ That group is the one `tools/preflight.sh` skips when you pass no arguments: it
 runs a test suite rather than a compile, so it is minutes rather than seconds.
 Name it when you want it.
 
+Its commands carry `env FERX_REQUIRE_DEBUG_ASSERTIONS=1`, which arms a canary
+(`debug_assertion_canary` in `src/lib.rs`) that fails the run if `debug_assert!`
+turns out to compile to nothing after all. Without it the gate would rest on the
+*absence* of a `--profile` flag, and a `[profile.dev] debug-assertions = false`
+or a `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS` in the environment would neuter the
+whole job with every test still green. Do not set that variable anywhere else:
+every other job legitimately runs with the guards off.
+
+What this does **not** fix: `debug_assert!` condition lines are still permanently
+zero-hit regions in the Codecov reports, because every coverage job uses a
+release-derived profile. See #1248.
+
 ### 8.3 CI workflows (ferx-r)
 
 `R-CMD-check.yaml` and `test-coverage.yaml`. **ferx-r CI builds ferx-core from
--- a/docs/development/sdlc.qmd
+++ b/docs/development/sdlc.qmd
@@ -293,7 +293,7 @@ runs a test suite rather than a compile, so it is minutes rather than seconds.
 Name it when you want it.
 
 Its commands carry `env FERX_REQUIRE_DEBUG_ASSERTIONS=1`, which arms a canary
-(`debug_assertion_canary` in `src/lib.rs`) that fails the run if `debug_assert!`
+(`debug_assertion_canary` in `src/lib_tests.rs`) that fails the run if `debug_assert!`
 turns out to compile to nothing after all. Without it the gate would rest on the
 *absence* of a `--profile` flag, and a `[profile.dev] debug-assertions = false`
 or a `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS` in the environment would neuter the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,3 +70,72 @@ pub use types::*;
 
 #[cfg(feature = "survival")]
 pub use api::{predict_categorical, predict_survival, SurvivalPredictionResult};
+
+/// Positive proof that the `debug_assert!` guards in this crate are LIVE (#344).
+///
+/// Every other test job builds with `ci-test`/`ci-fast`, both of which
+/// `inherits = "release"`, so all ~180 `debug_assert!`s compile to nothing. The
+/// `Tests (debug-assertions)` job exists to run them, and it does that by simply
+/// omitting `--profile` — the dev default has them on.
+///
+/// That is an invariant held by *absence*, which is the fragile kind. Nothing in
+/// the command string distinguishes a build with the guards live from one without:
+/// a `[profile.dev] debug-assertions = false` in `Cargo.toml`, or a
+/// `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` in the workflow environment, would
+/// neuter the whole job while all 4366 tests, all nine preflight-contract tests and
+/// both `cargo test` commands stayed green. The gate would then be exactly the
+/// no-op it was filed to replace.
+///
+/// So the group arms this canary through its own argument vector
+/// (`env FERX_REQUIRE_DEBUG_ASSERTIONS=1 cargo test …`, visible in
+/// `tools/preflight.sh --list` and asserted by
+/// `tests/preflight_owns_the_fast_gates.rs`), and the canary fails the run when the
+/// guards turn out to be dead.
+#[cfg(test)]
+mod debug_assertion_canary {
+    /// The env var `tools/preflight.sh`'s `debug-assertions` group sets. Deliberately
+    /// opt-*in*: every other job legitimately runs with the guards off, so an
+    /// unconditional assertion here would fail `Tests + coverage (core)` and both
+    /// coverage jobs for doing exactly what they are supposed to do.
+    const DEMAND: &str = "FERX_REQUIRE_DEBUG_ASSERTIONS";
+
+    #[test]
+    fn debug_assert_guards_run_when_the_gate_demands_them() {
+        let demanded = std::env::var_os(DEMAND).is_some();
+
+        // Observe the MACRO, not `cfg!(debug_assertions)`. The property the job exists
+        // to establish is "the condition of a `debug_assert!` is evaluated"; reading
+        // the cfg flag is one indirection away from it.
+        //
+        // `Cell` rather than `let mut`: with the guards off the assignment is compiled
+        // away, and a `mut` binding that is then never mutated warns.
+        let evaluated = std::cell::Cell::new(false);
+        debug_assert!({
+            evaluated.set(true);
+            true
+        });
+        let evaluated = evaluated.get();
+
+        assert_eq!(
+            evaluated,
+            cfg!(debug_assertions),
+            "`debug_assert!` and `cfg!(debug_assertions)` disagree, which should be \
+             impossible — the macro is defined in terms of that cfg. Read this as the \
+             canary itself being broken rather than the profile being wrong."
+        );
+
+        // The gate. Note the shape: `!demanded || evaluated`, so the only run this can
+        // fail is one that asked for the guards and did not get them. A single
+        // expression rather than an `if`, so both profiles execute every line here and
+        // the canary cannot read as uncovered on the patch gate.
+        assert!(
+            !demanded || evaluated,
+            "{DEMAND} is set — this run came from `tools/preflight.sh debug-assertions` \
+             or the `Tests (debug-assertions)` CI job — but `debug_assert!` compiled to \
+             nothing, so every guard in the crate is dead and the job is green for no \
+             reason. Something switched debug-assertions off for the dev profile: a \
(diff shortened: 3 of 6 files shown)


roninsightrx:
### CI is green, and it corrects one claim in the description

All checks pass, including the new job on its first (cold-cache) run:

| job | result | wall clock |
|---|---|---|
| **Tests (debug-assertions)** | **pass** | **29m57s** |
| Tests + coverage (core) | pass | 28m26s |
| Tests + coverage (TTE/CTMM endpoints) | pass | 24m32s |
| Check / Clippy / Format / Rustdoc / Docs lint / Public API baseline | pass | 15s – 1m53s |

So the triage result holds on CI hardware too: **no latent `debug_assert!` failure anywhere** — 4035 and 4366 tests, 0 failed.

**The correction.** The "Alternatives considered" section argues that the dev profile wins because "the tests are the cheap side" in a compile-bound repo. On this runner that is backwards. From the job log:

```
test result: ok. 4035 passed; ...; finished in  396.49s   # --features ci
test result: ok. 4366 passed; ...; finished in 1223.02s   # --features ci,markov,nn
```

That is **27 minutes of tests** and roughly 3 minutes of compile — the reverse of my local measurement (80s + 198s of tests on a warm 10-core machine). Unoptimised code plus a 2-core runner plus `RAYON_NUM_THREADS=1` is a much worse multiplier than the local numbers suggested, and I should have said "measured locally, unverified on CI" rather than generalising from them.

**It does not change the decision, for a different reason than the one I gave.** The job runs in parallel with `Tests + coverage (core)` (28m26s) and `Tests + coverage (TTE/CTMM endpoints)` (24m32s), so at ~30m it does not extend total CI wall clock — it fits inside the envelope the existing test jobs already set. And the `ci-fast`-plus-assertions alternative trades ~24 min of unoptimised tests for an optimised lib build that those same jobs show costs ~20 min, so it is close to a wash on wall clock while giving up `overflow-checks` and the literal-repro property. The dev profile still wins; the reason is "it fits in the existing envelope", not "its tests are cheap".

Two consequences worth flagging for review:

- **This job is now roughly the critical path** rather than comfortably inside it. If the Tier-1 suite grows much, it is the first job to push total CI time up, and the `ci-fast`-plus-assertions variant becomes the better trade at that point.
- The 29m57s figure is **cold cache** — first run of the `ci-debug-assertions` key, so the dep graph built from scratch. Steady-state should be a few minutes lower, but only the ~3 min compile half is cacheable; the 27 min of tests is not.

I have left the description's prose as written rather than silently rewriting it — this comment is the correction. Happy to edit it if you would rather the PR body stand on its own.

roninsightrx:
Review complete — I found two issues to address before merge.

1. **The new gate does not positively verify that `cfg(debug_assertions)` is enabled.** The guard at `tests/preflight_owns_the_fast_gates.rs:569` only rejects `--profile`/`--release` in the command string. Adding `[profile.dev] debug-assertions = false` to `Cargo.toml`, or setting `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` in the workflow, would silently neuter the new job while all nine preflight contract tests and both `cargo test` commands could remain green. That is the exact failure mode this PR is intended to prevent. Please add a runtime canary that is enabled only for this preflight group (for example, an env-gated lib test that asserts `cfg!(debug_assertions)`), or another direct positive check of the effective Cargo profile.

2. **Closing #344 leaves its documented Codecov failure mode unresolved.** The 2026-08-31 issue comment shows that `debug_assert!` condition lines are emitted as permanently zero-hit regions by the existing `ci-fast`/`ci-test` coverage jobs. This new job runs plain `cargo test` and uploads no coverage, while all Codecov-producing jobs still use release-derived profiles, so the patch-coverage ceiling described on #344 is unchanged. Either merge a debug-assertions coverage report into Codecov / enable assertions for the relevant coverage profile, or change `closes #344` to a non-closing reference and file/link a follow-up issue so that remaining defect is not silently closed.

One smaller guard-quality issue: `tests/preflight_owns_the_fast_gates.rs:792` uses raw `yml.contains(...)`, so a comment containing `tools/preflight.sh <group>` can satisfy the generic “every opt-in group runs in CI” assertion. The current `debug-assertions` group is also hard-coded in the earlier job test, but future opt-in groups are not. Please derive this from parsed `run:` commands rather than the full YAML text if keeping the generic invariant.

Validation performed:

- All remote checks are green, including the new 4,366-test job.
- `cargo test --test preflight_owns_the_fast_gates --no-default-features --features ci`: 9 passed.
- `git diff --check` and `bash -n tools/preflight.sh`: passed.

The workflow/script wiring itself is otherwise sound.

roninsightrx:
Thanks — all three verified against the code before touching anything, and all three were real. Pushed as `9f98d6e` (the first commit was amended, so the branch was force-updated; only its subject line changed, `closes #344` → `refs #344`).

---

### 1. No positive verification that `cfg(debug_assertions)` is on — **confirmed, fixed**

Correct, and it is the sharper version of the problem than I had modelled. The gate rested on the **absence** of a `--profile` flag, and the guard at what is now line 571 only rejects what is *visible in the command string*. Neither `[profile.dev] debug-assertions = false` nor `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` is visible there.

**`debug_assertion_canary` in `src/lib.rs`.** It observes the **macro**, not the cfg flag — the property the job exists to establish is "a `debug_assert!` evaluates its condition", and reading `cfg!(debug_assertions)` is one indirection away from that:

```rust
let evaluated = std::cell::Cell::new(false);
debug_assert!({ evaluated.set(true); true });
let evaluated = evaluated.get();
assert!(!demanded || evaluated, "…");
```

Three shape decisions worth flagging, since each is load-bearing:

- **Armed opt-in**, via `FERX_REQUIRE_DEBUG_ASSERTIONS`. Every *other* job legitimately runs with the guards off, so an unconditional assertion would fail `Tests + coverage (core)` and both coverage jobs for doing exactly the right thing.
- **The arming lives in the argument vector** — `env FERX_REQUIRE_DEBUG_ASSERTIONS=1 cargo test …` — not as a shell prefix and not an `export` inside the group function. All three reach the child; only this form is in `$*`, so only this form appears in `--list`, which is what a developer copies and what the guard test can assert. Same reasoning `group_rustdoc` already uses for `RUSTDOCFLAGS`.
- **`!demanded || evaluated` as one expression, not an `if`**, so every line executes under both profiles and the canary cannot itself read as an uncovered patch line.

`Cell` rather than `let mut`, incidentally: with the guards off the assignment is compiled away and a `mut` binding that is never mutated warns.

**Mutation-checked, as you asked:**

| run | result |
|---|---|
| armed, dev default | `ok. 1 passed` |
| **`CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` + armed** | **FAILED** — panics at `src/lib.rs:131` with the diagnostic naming both disable routes |
| `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false`, **unarmed** | `ok` — this is the case that keeps the three release-profile jobs green, and it had to be checked too |

And the arming is now pinned as tightly as the flags are: `load_bearing_flags_and_feature_coverage_survive_in_the_command_list` requires every command in the group to `starts_with("env FERX_REQUIRE_DEBUG_ASSERTIONS=1 cargo ")`. Dropping that prefix from one line turns it red — otherwise the canary would sit there passing vacuously, which is the same failure one level up.

Full group is green: 4036 and 4367 tests (the +1 is the canary).

---

### 2. `closes #344` over-claims — **confirmed, changed to `refs #344`, follow-up is #1248**

You are right, and the issue's 2026-08-31 comment is explicit about it: the `debug_assert!` condition lines are permanently zero-hit regions in every Codecov job, all of which build release-derived profiles. This PR's job runs plain `cargo test` and uploads no coverage, so that ceiling is exactly where it was. The comment even ends with "Option 1 is the only one that addresses the original complaint in this issue **as well**" — i.e. the coverage half was folded into #344's scope, and closing it here would have buried a live defect.

Changed to a non-closing reference in the PR **title**, the PR **body**, and the **commit subject** (hence the amend — a `closes` in a commit message closes the issue on merge regardless of what the PR says, so leaving it there would have defeated the point).

Filed **#1248** with the full analysis carried over: the mechanism, the measured `src/sens/ode_provider.rs` figures (4 condition lines, 4 zero-hit), the #1136 impact (49/56 = 87.5% against a 90% gate), the `*_tests.rs`-emits-no-`SF:`-record gotcha, and all four options — including the explicit "do **not** reformat guards onto one line to clear the gate" note. `docs/development/sdlc.qmd` now states what this job does *not* fix and points at it.

I did **not** take the "merge a debug-assertions coverage report into Codecov" route in this PR. It is a genuine option and it is option 2 on #1248, but instrumenting a job that already runs ~30 min uninstrumented needs a measured trial and a `codecov.yml` flag entry, and it would put this job unambiguously on the critical path — too much to fold in unmeasured. Enabling assertions on `ci-fast`/`ci-test` is option 1 there, and it is what #344 explicitly ruled out for the first half; it may still be right for the coverage half, but again on measurement, not assumption.

If you would rather #344 be closed here and the coverage work tracked as fresh scope, say so and I will flip the keyword.

---

### 3. `yml.contains` satisfiable by a comment — **confirmed by demonstration, fixed**

Fixed: it now derives from parsed `run:` steps via the existing `run_commands`, applied to the whole document (so moving the group between jobs stays fine; running it nowhere does not).

Worth recording that I did not take this one on faith either, because the check was *not* currently being satisfied by a comment — no comment in `ci.yml` carries the exact string today. So I built the mutation that separates the two implementations: replace the `run:` line with `run: echo skip` and add `# tools/preflight.sh debug-assertions` above it.

- new parsed-`run:` version → **FAILED**
- previous `yml.contains` version, same mutated file → **passed**

So the weakness was exactly as described, and one edit away from mattering — `ci.yml` is full of comments naming these commands, and `tools/preflight.sh` and `sdlc.qmd` both contain the literal string already. Also generalised: this is the path future opt-in groups take, since the `debug-assertions` job is only hard-coded in the earlier per-job test.

---

### On the 15:47Z timing correction

Folded into the PR body, under *Alternatives considered*, as a quoted correction attached to the bullet whose reasoning it invalidates — rather than left as a comment someone has to scroll to. The `Performance` section now carries the 29m57s CI figure instead of the local 80s/198s. The original comment stays as the audit trail.

The substance, briefly: my "the tests are the cheap side in a compile-bound repo" argument is backwards on a 2-core runner — 27 min of tests against ~3 min of compile. The decision holds for a different reason (it fits the envelope the existing 28m26s / 24m32s test jobs already set), and the consequence is that this job is now roughly the critical path rather than comfortably inside it.

---

### Verification on this push

- `tools/preflight.sh debug-assertions` → 4036 and 4367 tests, 0 failed
- `cargo test --test preflight_owns_the_fast_gates …` → 9 passed
- Seven mutations exercised in total (four from the original round, three new); each turns the intended guard red.

One thing I deliberately did **not** add, flagged as an open question on the PR: the canary proves `debug_assert!` is live but not that `overflow-checks` is, which the PR also claims as a dev-profile benefit. `cfg!(overflow_checks)` is still unstable (rust-lang/rust#111466 — checked against this toolchain, not recalled), and the runtime alternative means swapping the global panic hook around a deliberate overflow while other tests run in parallel. Not worth that hazard for a secondary claim, but say the word if you want it.

roninsightrx:
### `codecov/patch` went red on my last push — and the cause is #1248, landing on the canary itself

Worth writing up rather than just fixing, because it is a live demonstration of the thing finding 2 is about.

The canary landed inline in `src/lib.rs`. Codecov's compare API for this PR:

```
src/lib.rs patch -> lines: 11, hits: 9, misses: 2, coverage: 81.81
```

The two misses, from the per-line data:

```
112  let evaluated = std::cell::Cell::new(false);   hit
113  debug_assert!({                                hit
114      evaluated.set(true);                       MISS
115      true                                       MISS
116  });
```

Which is exactly the mechanism from #344's 2026-08-31 comment, now filed as #1248: `debug_assert!` expands to `if cfg!(debug_assertions) { … }`, `ci-fast` inherits `release`, so llvm-cov emits regions for the interior that **no test can ever execute**. It is a measurement artefact, not a coverage gap — and it has no fix at the line, because the interior of a `debug_assert!` is unreachable under a release-derived profile by construction. The code written to demonstrate the blind spot tripped over the blind spot's sibling defect.

Note also that this was invisible to every local gate: `tools/preflight.sh` is compile-and-lint plus the new suite, and coverage is only measured in CI.

**Fix: moved the module to `src/lib_tests.rs`** behind `#[cfg(test)] #[path = "lib_tests.rs"] mod debug_assertion_canary;` — this repo's documented home for test modules (`CLAUDE.md`'s sibling-`*_tests.rs` pattern), where the other **53** such files already live. `src/lib.rs` keeps only the doc comment and the `mod` declaration, neither executable, so its patch contribution is now empty.

Verified rather than assumed, since the whole point of a fix like this is that it might be metric-gaming: of the **147 files Codecov tracks for this repo, zero are `*_tests.rs`** — llvm-cov emits no `SF:` record for them at all, matching the gotcha documented on #344. `codecov.yml` has no ignore rule for them; it is llvm-cov behaviour, and it already applies to every other test module in the tree.

**Routes I deliberately did not take**, since this is precisely where a fix shades into gaming:

- **Reformatting the guard onto one line** so the condition is not its own line — the anti-pattern #344's comment explicitly calls out. Not done.
- **Adding `src/**/*_tests.rs` to `codecov.yml`'s `ignore`** — would change the patch denominator for every future PR. Might well be right; too broad to decide inside this PR, and it belongs on #1248.
- **Leaving it red and documenting the arithmetic**, as #1136 did. Defensible, but it puts a permanent red X on a PR whose entire subject is CI gates that mean what they say.

Relocating a *test module* is a different act from either: `codecov.yml`'s own stated policy is to scope ignores by what the code **is**, listing test scaffolding as a legitimate role, and this is test scaffolding sitting in the wrong file.

**Re-verified after the move** (a relocation that broke the canary would be worse than the red X):

| check | result |
|---|---|
| canary, armed, dev profile | ok |
| canary, `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` + armed | **FAILED** — still catches the mutation |
| `tools/preflight.sh debug-assertions` | 4036 and 4367 tests, 0 failed |
| `tools/preflight.sh` | `preflight OK — fmt check clippy rustdoc docs public-api` |
| `cargo test --test preflight_owns_the_fast_gates` | 9 passed |

Pushed as `d246b17`. One more argument for #1248, incidentally: the first PR in the repo to add a `debug_assert!` inside new code hits this immediately, and the only remedies available at PR level are relocation or a documented red.

roninsightrx:
### All checks green on `d246b17`, `codecov/patch` included

| job | result | wall clock |
|---|---|---|
| **Tests (debug-assertions)** | pass | 21m59s |
| Tests + coverage (core) | pass | 17m25s |
| Tests + coverage (TTE/CTMM endpoints) | pass | 16m34s |
| **codecov/patch** | **pass** | — |
| Check / Clippy / Format / Rustdoc / Docs lint / Public API baseline / CodeFactor | pass | 1s – 2m5s |

Two numbers worth recording, both of which sharpen claims I made earlier in this PR:

- **21m59s warm against 29m57s cold**, so the cold-cache caveat in my 15:47Z comment was right about the direction. Only the ~3 min compile half is cacheable, and the whole run came down by ~8 min, which suggests the earlier cold run was also paying dependency-build time the warm one skips.
- It remains **the longest single job**, now by a wider relative margin (21m59s against 17m25s / 16m34s on the same warm run) than the first run suggested. Still parallel with them, so total CI wall clock is unchanged — but the "if the Tier-1 suite grows, this is the first job to push CI time up" caveat stands, and the `ci-fast`-plus-assertions variant is the lever if that day comes.

`Performance` in the PR body now carries both figures rather than the cold one alone.

Nothing outstanding from my side. Summary of where the three findings landed:

1. **Positive `cfg(debug_assertions)` verification** — `debug_assertion_canary` (`src/lib_tests.rs`), armed opt-in from the preflight group's argv, mutation-verified red under `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` and green unarmed. The arming is itself pinned by the guard test.
2. **`closes #344` → `refs #344`** in title, body and commit subject; the Codecov half is #1248 with the measurements and four options.
3. **`yml.contains` → parsed `run:` steps**, with the discriminating mutation run both ways to confirm the weakness was real rather than assumed.

Plus the one you did not ask for but which the first push surfaced: the canary's own `debug_assert!` interior was 2 permanently-missed patch lines, fixed by moving the module to the repo's standard sibling-test-file location.

Issue #1242, opened 2026-09-05
estimation: a free omega Cholesky diagonal that starts ON the −6 rail never leaves it (foce×bobyqa, foce×lbfgs); −5.93 recovers exactly


TeunP:
Split out of #1229, which rejects the start instead of fixing the optimizer.

## Measured (2026-09-05, warfarin 10 subj, `examples/warfarin.ferx` + `data/warfarin.csv`, FOCE, `covariance = false`)

`omega ETA_CL ~ v` **free**, everything else as in the example:

| `v` | packed start `ln L₁₁` | outcome |
|---|---|---|
| `6.14421235e-6` = e⁻¹² | **exactly −6.0** (on the rail) | stuck at ω²_CL = 6e-6, `collapse` verdict at the lower guard, *not converged*, TVV 74.16 (base 7.7307), σ 42% (base 1%) |
| `7e-6` | −5.93 | recovers the base optimum to 6 digits (TVCL 0.132969, TVV 7.7307, TVKA 0.725207, ω²_CL 0.028595) |
| `0.0` / `5e-6` | −9.21 / −6.10 → `clamp_to_bounds` → −6.0 | identical to the e⁻¹² arm |

Same `0.0` start by optimizer / method: `foce × bobyqa` (default) sticks; `foce × lbfgs` sticks with identical numbers; **`foce × slsqp` recovers exactly**; **`focei × lbfgs` recovers**; `focei × bobyqa` half-collapses (see #997); `saem` collapses; `gn` never moves. Full table in [#1229's plan comment](https://github.com/FeRx-NLME/ferx-core/issues/1229#issuecomment-5552050434). ferx's base arm matches NONMEM's (`$OMEGA 0.09`, OBJV −280.360) to 5 digits, so the recovering arms are at the right optimum.

## Two design questions

1. **Why is a start *on* the bound unrecoverable for BOBYQA and L-BFGS but not SLSQP?** `optimize_nlopt_once` clamps `x0` onto the bound and sets `set_initial_step` to 0.5 in scaled space (`outer_optimizer.rs` ~2004). A −0.07 offset in packed space is the difference between exact recovery and a collapsed fit, which is not a property of the objective — the same ω² is one step away in both cases. Worth checking whether the first interpolation / line-search probes straddle the bound and get projected back, so the coordinate never sees a descent direction.
[...]

Issue #1217, opened 2026-09-03
epic(bugs): 2026-09 bug order — SS × hazard, endpoint routing, break-time tolerances, adaptive oracle, SS × model time, sens saltations


TeunP:
Tracker for the open TeunP-assigned bugs and the route through them, all four groups (ODE value path, sens, TTE tail, adaptive). Plans live on the issues. **One table, one row per issue.** Before starting an issue: put your handle in **picked up**, assign yourself on the issue. When it lands: status ✅ and the PR. Nobody starts a row that has a name in *picked up*.

Ordering rule: silent wrong-number defects first, cheapest oracle first, a defect that sits *under* a later item before that item, decisions (no code) early, features last. Rows sharing a **batch** letter go in one PR. **Critical path: steps 1 → 2 → 6 → 11.** The sens chain (steps 5, 8) runs in parallel with everything. Nothing in phase 4 starts before steps 6 and 10 are done.

Area rule: sens work does not touch `ode/predictions.rs`; value-path work does not touch `sens/`.

Status: ✅ done · 🔧 picked up · ⬜ open · 🧭 decision (no code) · ⏸ kept in view, not now

| step | issue | area | what | batch | status | picked up | PR |
|---|---|---|---|---|---|---|---|
| — | #1210 | value path | SS run-in hazard into `H(0)` | | ✅ | TeunP | #1219 |
| — | #1218 | TTE | `predict_survival` NaN on zero-max grid | | ✅ | TeunP | #1222 |
| — | #1199 | api | unrouted loader drops the hazard half | | ✅ | TeunP | #1225 |
| — | #1186 | value path | derived break within 1e-10 applies a dose twice | | ✅ | TeunP | #1231 |
| — | #1196 | value path | duplicate infusion-membership predicate | | ✅ | TeunP | #1231 |
| — | #1223 | TTE | one-solve vs two-solve pre-start NaN contract | | ✅ | TeunP | #1230 |
| — | #1167 | TTE | engine split = integrator tolerance | | ✅ closed, no code | TeunP | — |
| — | #1220 | TTE | SS × hazard anchors left over from #1210; item 1 anchored, items 2–4 need no work | | ✅ closed | TeunP | #1232 |
| — | #1227 | tests | per-route-lag Tier-3 unidentifiable | | ✅ | TeunP | #1228 |
[...]

PR #1246, opened 2026-09-05
fix(api): reject a free variance whose packed start lands on the optimizer's lower rail [closes #1229]

### LINKED ISSUE #1229: A free omega declared at 0.0 has a packed start below its own lower rail, so it can only end at a guard — check it at model-check time
Split out of the #1228 review, where it was the one finding not fixed in that PR because it
is a `src/` change and that PR is test-only.

## The shape

```
omega ETA_CL ~ 0.0        # free, no FIX
```

A *free* omega declared at exactly `0.0` cannot end anywhere but a rail, and the reason is
not the obvious one.

The parser does not carry the `0.0` through. It floors the Cholesky diagonal to `1e-4`
(measured — `default_params.omega.chol == [0.0001]`, identical with and without `FIX`), so the
coordinate's **packed start is `ln(1e-4) = -9.21`**. Its own lower rail from `compute_bounds`
(`src/estimation/parameterization.rs:554`) is `-6.0`.

So the free coordinate starts *below its own admissible interval*, is clamped there, and — with
nothing in the data identifying a variance — walks to the opposite rail at `+6`
(ω² ≈ 162 755). Since #1205 that is a `runaway` verdict: `converged: false` plus a `Critical`
`parameter_at_runaway_guard`.

## Why it is worth a check rather than leaving it to the warning

#1227 was exactly this, in a test fixture, and the cost was not just a wrong boolean. Measured
on that fixture — one subject, noise-free data generated at truth:

| | free ω | ω `FIX`'d | truth |
|---|---|---|---|
| OFV | −323.2803 | −610.0915 | |
| TVCL | 2.5889 (−48%) | 4.99536 (−0.09%) | 5.0 |
| TVV | 43.4805 (−13%) | 49.70493 (−0.59%) | 50.0 |
| TVKA | 0.7092 (−29%) | 0.98482 (−1.5%) | 1.0 |

The runaway variance does not sit inertly at the rail; the fit is materially different. The
existing `parameter_at_runaway_guard` warning fires *after* that has happened and describes the
symptom (`ETA_CL ... at upper guard, runaway`) without connecting it to the declaration that
made it inevitable. Its guidance — "revisit the model, data, initial estimates, or estimation
method" — is exactly the wrong advice here: the remedy is one keyword.

## Suggestion

At model-check time (`api::validation`, alongside the other `check_model_data` asserts), flag a
**free** omega whose declared variance is `0.0`, naming `FIX` as the fix. It is a statically
decidable property of the declaration — no fit required — and it is not a judgement call the
way "few subjects" would be: there is no start from which a free `~ 0.0` reaches an interior
optimum.

Worth deciding as part of it:

- Warning or hard error? A warning is the conservative choice, but this shape has no valid
  reading, which is the usual bar for rejecting.
- Does the same argument extend to a free omega declared at any value below `exp(-6)` ≈ 0.0025,
  i.e. any start the packing floors below its own lower rail? That is the more general
  statement of the defect and probably the right predicate.
- Should the parser's `1e-4` floor and `compute_bounds`' `-6` rail be reconciled at all? A
  floor that lands outside the bound it is later checked against is surprising on its own
  terms, independent of this check.

## Not blocking

#1228 fixes the test fixtures and is correct as scoped; this is the general case.

TeunP:
## Why

`omega ETA_CL ~ 0.0` without `FIX` cannot end anywhere but a guard rail, and the reason is
not the declared zero. The parser does not carry the zero through: `OmegaMatrix::from_matrix_with_mask`
regularises the non-PD matrix by `1e-8`, so `L₁₁ = 1e-4` and the coordinate's **packed start is
`ln(1e-4) = −9.21`** — below its own lower bound of `−6.0` from `compute_bounds`. Every optimizer
clamps the start onto that rail, and from there the fit is materially wrong either way:

* #1227 (1 subject): ran away to `+6`, ω² ≈ 1.6e5, `TVCL` 48% low / `TVV` 13% low / `TVKA` 29% low
  against the same model with `FIX` (−0.09% / −0.59% / −1.5%).
* #1229's warfarin arm (10 subjects, `data/warfarin.csv`): collapsed at `−6`, ω² stuck at 6e-6,
  `TVV` 74.16 against a true 7.73 (10×), σ 42% against 1%.

The existing `parameter_at_runaway_guard` warning fires *after* that has happened, describes the
symptom, and its guidance ("revisit the model, data, initial estimates, or estimation method") is
the wrong advice: the remedy is one keyword. NM-TRAN refuses the equivalent stream outright —
error 76, `INITIAL ESTIMATE OF VARIANCE CANNOT BE ZERO UNLESS FIXED`.

Full measurement table (NONMEM 7.6.0 + ferx, both on `data/warfarin.csv`) is in the
[#1229 plan comment](https://github.com/FeRx-NLME/ferx-core/issues/1229#issuecomment-5552050434);
this PR implements its section 5.

## What changed

**`check_variance_init_rails(init_params, options)`** in `src/api/validation.rs` — `pub(crate)`, so no
public-API surface is added. It flags every **free** Ω / Ω_IOV / mixture-Ω (#977) **diagonal**
coordinate with `packed[i] <= lower[i]`, computed from the production
`pack_params` / `compute_bounds` / `packed_fixed_mask` / `coordinate_kinds`, as
`E_OMEGA_INIT_AT_RAIL` (error, `with_block("parameters")`, one diagnostic per offending
coordinate).

Three reasons the predicate has to read the packed vector rather than the declared variance:

* In a `block_omega` the diagonal `L_ii` depends on the off-diagonals, so the declared
  variance is not the coordinate that gets clamped. `block_omega (ETA_CL, ETA_V) =
  [0.09, 0.089997, 0.09]` declares two interior variances and still puts `L₂₂` on the rail.
* The declared `0.0` is *unreadable* after the parse — it arrives as `1e-8`.
* `compute_bounds` pins a `FIX`-ed coordinate at `lower == upper == packed`, so
  `packed <= lower` is **true for every FIX-ed variance**. The `packed_fixed_mask` consult is
  the only thing that keeps `~ 0.0 FIX` — the documented spelling for "no variability" — from
  being rejected.

Two message variants, keyed on the variance having landed at or below the `1e-8` regularisation
floor (which is as close to "the user wrote zero" as is recoverable, and is also exactly the
range in which the variance is numerically zero whatever was typed): the floor variant leads
with `FIX` and quotes NM-TRAN error 76; the tiny-non-zero variant quotes the value, names the
cliff (`v ≤ 6.14e-6`) and leads with "start at ≥ 1e-5".

**Wiring**: `fit_inner` immediately after `first_error(&check_model_options(..))?` — on
`init_params`, not `model.default_params`, since `--inits-from-nca` and the ferx-r override path
replace the parsed inits and it is the vector the optimizer starts from that gets clamped — and
`validate_model_file` on `parsed.model.default_params`, so `ferx check model.ferx` reports it
without `--data`. **Parser, `simulate()` and `predict()` are untouched**, which is what keeps the
~69 predict-only free-zero fixtures in the repo valid.

**Σ is excluded, measured not assumed**: a free `sigma PROP_ERR ~ 0.0 (sd)` packs to `ln(1e-10)
= −23`, is clamped onto its own `−8` rail, and recovers the base optimum exactly. The Σ rail does
not trap.

## Alternatives considered

**Warning instead of error.** Two of seven method × optimizer combinations do escape this start
(`foce`×`slsqp` and `focei`×`lbfgs` both recover ω²_CL = 0.02859 exactly), so the issue's "no
start reaches an interior optimum" is too strong and the message does not claim impossibility.
But every *default* path fails — `foce`×`bobyqa`, `foce`×`lbfgs`, `saem` collapse; `gn` never
moves — and the decisive case is `focei`×`bobyqa`, which half-collapses (ω²_CL 1e-5, σ 15% high)
while reporting `converged: true` and hitting **no** guard. A warning nobody reads is no better
than the post-hoc guard we already have.

**Rejecting at parse time.** That would break the ~69 predict / simulate-only fixtures whose
free zero is harmless because nothing optimises it. The fit path is the right boundary.

**Moving the `1e-8` eigenvalue floor so it lands inside the `−6` bound.** Left to #1242, which
already owns it: it would silently change every `FIX 0` fixture's `L₁₁`, and the measured fix for
the user here is one keyword. Rejecting the start and fixing the optimizer are independent — this
PR does the first.

## Cross-repo dependency

`ferx-r` `origin/main` declares no free `~ 0.0` omega (checked), the translator cannot emit one
(NM-TRAN refuses it upstream), and no public API changed, so no lock bump is required.

## Breaking changes
- [x] None

`check_variance_init_rails` is `pub(crate)`; `api/ferx-core-public-api.txt` is unchanged. It is
nevertheless a **behavioural** change for one shape: a model declaring a free `~ 0.0` variance
that used to fit (badly) now returns `Err`. That shape has no valid reading — see the migration
note.

<details>

A free zero-variance declaration is now refused. Add `FIX`:

```
omega ETA_CL ~ 0.0 FIX      # no IIV on CL
kappa KAPPA_CL ~ 0.0 FIX    # no IOV on CL
```

or, if the variance was meant to be estimated, start it at a value the optimizer can leave
(`~ 0.09`, or anything `> 6.14e-6`). `predict()` and `simulate()` accept the free zero as
before.

</details>

## Numerical validation

- [x] Compared against: NONMEM 7.6.0 (via the local Docker image), `data/warfarin.csv` /
      `examples/warfarin.ferx`, ADVAN2 TRANS2
- [x] Reference values:

```
# NONMEM, $OMEGA 0 free
NM-TRAN error 76: INITIAL ESTIMATE OF VARIANCE CANNOT BE ZERO UNLESS FIXED   (exit 107, no run)

# NONMEM, $SIGMA 0 free
the same error 76

# NONMEM, $OMEGA 0 FIX               OBJV  -19.536
# NONMEM, $OMEGA 1E-7 / 1E-5 free    OBJV  -19.536, OMEGA(1,1) -> ~1e-9   (the 0-FIX basin)
# NONMEM, $OMEGA 0.09 (base)         OBJV -280.360, THETA 0.132968 / 7.73044 / 0.725356,
#                                    OMEGA(1,1) 0.0285971

# ferx FOCE/BOBYQA, omega ETA_CL ~ 0.09 (base)
#   TVCL 0.132969  TVV 7.7307  TVKA 0.725207  omega2_CL 0.028595   <- NONMEM base to 5 digits
```

ferx packed starts vs the `-6` rail, measured on this branch (`pack_params`):

| `omega ETA_CL ~` | packed `ln L₁₁` | rejected? |
|---|---|---|
| `0.09` | −1.203973 | no |
| `0.0` free | −9.210340 | **yes** |
| `1e-9` free | −10.361633 | **yes** |
| `5e-6` free | −6.103036 | **yes** |
| `6.144212353328210e-6` = e⁻¹² | **−6.0 exactly** | **yes** |
| `7e-6` free | −5.934800 | no |
| `1e-5` free | −5.756463 | no |
| `0.0 FIX` | −9.210340 (bound pinned to match) | no |
| `sigma ~ 0.0 (sd)` free | −23.03 vs a −8 rail | no (out of scope) |

## Performance

Six `packed_len`-sized vector builds, once per `fit_inner` call, before the first objective
evaluation.

## Tests

`src/api/tests/variance_init_rail_tests.rs` (Tier 1, `#[path]` sibling of `validation.rs`), 15
tests; four more in `tests/check_command.rs` (Tier 2 — every one returns `Err` immediately or
runs a single outer iteration).

**Every assertion was mutation-tested** — nine mutations, each the smallest edit that removes
part of the fix, run against the whole file:

| Mutation | Verdict | Killed by |
|---|---|---|
| baseline (no mutation) | 15 passed | — |
| `p > lo` → `p >= lo` (`<=` → `<`) | killed | `variance_exactly_on_the_rail_is_rejected` |
| drop `is_fixed \|\|` | killed | `fixed_zero_omega_is_accepted_even_though_it_sits_on_its_pinned_bound` |
| drop the `OmegaDiagonal` scope gate | killed | `free_zero_sigma_is_not_flagged` |
| scope gate on `Theta` only | killed | `free_zero_sigma_is_not_flagged` |
| predicate on the declared variance, not the factor | killed | `near_singular_block_flags_the_factor_not_the_declared_variance` |
| variant split `<=` → `<` | killed | `free_zero_omega_is_rejected_naming_the_eta_and_fix`, `every_offending_coordinate_gets_its_own_diagnostic` |
| drop `.with_block("parameters")` | killed | `free_zero_omega_is_rejected_naming_the_eta_and_fix` |
| label a kappa as `omega` | killed | `free_zero_kappa_is_rejected_and_named_as_a_kappa` |
| skip the mixture segment | killed | `free_zero_mixture_omega_override_is_rejected_naming_its_class` |
| `break` after the first diagnostic | killed | `every_offending_coordinate_gets_its_own_diagnostic` |

**The first round found a survivor, in this PR's own code.** "Drop the scope gate" left every test
green because there were *two* gates — a `coordinate_kinds` test in the loop **and** a
`decls[i] == None` skip — and each covered the other, so the Σ scope pin could not fail. They are
now one gate (`variance_decl_by_coordinate`, built from `coordinate_kinds`), and the mutation
kills a test. Anything touching that function should keep it to one gate.

Three of the tests exist specifically to avoid the "cannot fail" shapes CLAUDE.md lists:

* The **rail straddle** asserts the straddle itself (`pack_params` for `5e-6` `<= -6.0 <` that
  for `7e-6`, with the rail read from `compute_bounds`), so a later change to the floor or the
  bound cannot leave the pair on the same side agreeing for the wrong reason.
* The **on-the-rail** case needed a literal whose packed value is *bit-exactly* `-6.0`, which is
  the only input that can tell `<=` from `<`. Measured: the issue text's `6.14421235e-6` packs
  to `-6.000000000270841` — strictly below the rail, and it would have survived the mutation.
  `6.144212353328210e-6` packs to exactly `-6.0`.
* The **`FIX`** case asserts the trap it exists for (`packed <= lower` really *is* true for the
  FIX-ed coordinate) rather than only "no diagnostic".

**Fixture migration**: the check rejected **10 tests in 6 files**, and those are the
only files touched. The scope was read off the suite log rather than off a grep, so a
predict-only fixture cannot be swept in:

| File | Tests | Line(s) given `FIX` |
|---|---|---|
| `tests/compartment_states_nonmem.rs` | 4 (`nonmem_{2,3}cpt_{iv,oral}_compartment_amounts`) | 216, 268, 327, 384 |
| `tests/derived_output.rs` | 1 (`derived_grid_integral_matches_a_subdose_train_for_infusion_into_absorption`) | 531 |
| `tests/dose_compartment_routing.rs` | 1 (`depot_readout_with_a_non_default_dose_compartment_is_rejected`) | 676 |
| `tests/ss_absorption_nonmem_anchor.rs` | 1 (`fit_surfaces_ss_nonconvergence_warning_for_over_capacity_dosing`) | 199 |
| `tests/parallel_mixed_absorption.rs` † | 2 (`parallel_fit_recovers_fraction_and_kas`, `mixed_fit_recovers_fraction_dur_and_ka`) | 35, 73 |
| `tests/zero_order_absorption.rs` † | 1 (`zero_order_fit_recovers_duration`) | 34 |

† **`slow-tests`-gated — invisible to PR CI, found only by the `workflow_dispatch`**
([run 33975876280](https://github.com/FeRx-NLME/ferx-core/actions/runs/33975876280), which went red
on exactly these three and nothing else). This is the failure mode #1227 hit: local green plus PR
green proves nothing about a slow-gated fit. These three are one-subject *recovery* fits, so `FIX`
had to leave their convergence assertions intact — it does: both binaries now pass in full under
`--features ci,survival,slow-tests` (10/10 and 12/12), same recovered parameters. The set is
complete rather than merely first: `slow-tests.yml` runs `--no-fail-fast` and both binaries
reported their full counts, so nothing was cut short by an earlier failure.

Every one of the 10 passes afterwards with its asserted values **unchanged** — no tolerance was
touched. These fixtures fit at fixed θ / η = 0 (or recover parameters from one subject), so `FIX`
is the model they always effectively were. Full suite after the migration: **174 binaries, 0
failures**, plus the two slow-gated binaries green under `slow-tests`.

The plan's section 4 estimated 15 non-slow hits and left the slow-gated set unmeasured, flagging
its attribution as approximate; the measured totals are 7 non-slow and 3 slow-gated. Two corrections worth recording: `ss_absorption_nonmem_anchor::fit_on_ss_absorption_converges`
is **not** affected — it does `.replace("omega ETA_CL ~ 0.0", "omega ETA_CL ~ 0.05")`, so it never
reaches the rail — and neither is `derived_output::derived_theta_name_conflict_returns_err`. Their
neighbours in the same files are the ones that failed. The other 5 free-zero sites in
`dose_compartment_routing.rs` and the `SS_MM_MODEL` / `SS_FIRST_ORDER_MODEL` fixtures are
predict-only and were left alone.

No bundled `examples/*.ferx` declares a free zero variance (checked), so every CLI example is
unaffected.

<details>

```toml
[parameters]
  theta TVCL(0.2, 0.001, 10.0)
  theta TVV(10.0, 0.1, 500.0)
  omega ETA_CL ~ 0.0            # <- rejected
  sigma PROP_ERR ~ 0.02 (sd)
```

```
Error: `omega ETA_CL ~ 0.0` declares no variability but is not `FIX`-ed, so the optimizer is
asked to estimate a variance from a start it cannot leave: a zero variance is regularised to
1e-8, whose packed coordinate ln(L) = -9.21 lies below its own lower bound of -6.0, and the
start is clamped onto that rail. From there the coordinate stays collapsed or walks to the
opposite rail (variance ~ 1.6e5), and the population estimates move with it. Add `FIX` if
ETA_CL should carry no variability, or give it a real starting variance (>= 1e-5). NONMEM
rejects the same declaration outright: `INITIAL ESTIMATE OF VARIANCE CANNOT BE ZERO UNLESS
FIXED`.
```

</details>

## Docs
- [x] New pages linked in `docs/_quarto.yml` (no new pages)
- [ ] No user-visible change

`docs/file-formats/check-report.qmd` (a row in *Parameters and random effects*),
`docs/model-file/parameters.qmd` (a callout under *Diagonal omega* and a cross-reference under
*Diagonal kappa*), plus the code table in `src/diagnostics.rs`.

## Checklist
      (`--all-targets`), public-API baseline
- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → N/A (no new syntax; a rejection of an
      existing one, documented in `docs/`)
- [x] New example `.ferx` file added to `examples/` → N/A
- [x] New data file added to `data/` → N/A

### ferx-book
- [x] New estimator, DSL feature, or fit option → N/A

- [x] No example execution step needed — no bundled example declares a free zero variance, and
      the change is a pre-fit rejection with no numerical surface

## Review response (`5ccf6276`)

Every finding acted on was reproduced first — for 3, 4 and 5 by **rendering the actual message**
rather than reading the source, and the real output was worse than the review described.

| # | Status | What changed |
|---|---|---|
| 1 + 2 | fixed | One cause, one gate: `outer_search_runs(options)`. `outer_maxiter == 0` short-circuits to `evaluate_at_initial_params` **before an optimizer is constructed**, so nothing is clamped. Un-breaks bootstrap `--dofv` and `ferx gam --no-fit`. Not a bare `outer_maxiter > 0` — SAEM/IMP/IMPMAP/Bayes ignore it and search anyway. |
| 3 | fixed | Messages quote `L_ii²`, the quantity compared against the rail. A near-singular block gets its own message naming the **correlation**. |
| 4 | fixed | "You wrote 0.0" now requires a **diagonal** source hitting the floor **exactly**. |
| 5 | fixed | `.with_block("mixture")`, and the remedy spelled as it parses. |
| 6 | decided | `ferx check` keeps reporting it — `validate_model_file`'s contract is that a clean check and a fit agree. Documented, with `maxiter = 0` named as the way to declare simulation-only intent. |
| 7 | fixed | The block fixture now asserts the quoted variance **is** `L₂₂²`, and that neither `0.09` nor "≥ 1e-5" appears. `block_kappa` gained the same coverage. |
| 8 | withdrawn | Slow-tests core green on `36093c7a`; re-dispatched for this commit. |

**On the caveat that finding 1's round-trip was reasoned, not run.** Measured now, and it
*strengthens* the finding rather than weakening it:

```
exp(-6)^2           = 6.14421235332820980644e-6
exp(-12)            = 6.14421235332820980644e-6      (same bits)
ln(sqrt(exp(-12)))  = -6.00000000000000000000        (exact, not ~1 ULP away)
```

So a replicate clamped to the bound re-packs to bit-exactly `-6.0`; `packed <= lower` holds and
**every** railed replicate was dropped, not only those that rounded down. Pinned as
`a_railed_estimate_repacks_exactly_onto_the_rail`, built on a hand-made `ModelParameters` at
`exp(-6)²` exactly as suggested.

**Two things the review did not catch, found while fixing it:**

* The mixture suggestion read `omega(2) ETA_CL_MIX2 ~ 0.0 FIX` — `ETA_CL_MIX2` is the *packed
  coordinate's* display name and appears nowhere in a model file, so a user grepping for it finds
  nothing. Now `omega(2) ETA_CL`, with an assertion so it cannot return.
* `block_kappa` was getting variance-shaped advice for a correlation problem, because
  `is_block_element` only matched `block_omega`.

**Option (c) — parser records the declared variance — considered and rejected.** Once the message
describes `L_ii²` the ambiguity mostly dissolves: on a diagonal Ω the declared variance *is*
recoverable (a positive declaration is PD and survives verbatim, so an exact hit on the floor
means zero), and in a block no per-eta declared variance describes the coordinate at all. The only
residue is telling a literal `~ 1e-8` apart from `~ 0.0` — zero occurrences in this repo, identical
advice either way — against widening the public `ModelParameters`.

**Mutation-tested: 12 mutations on these fixes, 12 kills.** The first round had one survivor, and
it was in this PR's own code: `is_diagonal_omega_source` discriminated nothing, because every
declaration reaching it had already passed the block arm. Deleted rather than kept as a second
gate — the same two-redundant-gates trap this PR hit once already, caught the same way.

Gates on `5ccf6276`: preflight green (all six, public-API baseline unchanged), full `ci,survival`
174 binaries / 0 failures, docs-lint green, slow-tests re-dispatched.

## Reviewer hints

Three places to focus:

1. **`variance_decl_by_coordinate`** re-derives the packed segment boundaries the way
   `packed_len` does, because the packed vector carries no provenance. That duplication is the
   most likely thing to rot if the layout changes; each of its three live arms (`omega`,
   `kappa`, `[mixture] omega(2)`) is pinned by a test asserting its own keyword in the message.
2. **The `is_fixed` guard is load-bearing, not defensive.** `compute_bounds` pins a `FIX`-ed
   coordinate at `lower == upper == packed`, so without the mask consult the predicate is true
   for *every* `FIX`-ed variance — including `~ 0.0 FIX`, which is the spelling the diagnostic
   tells users to write. Mutation-confirmed: dropping it kills
   `fixed_zero_omega_is_accepted_even_though_it_sits_on_its_pinned_bound`.
3. **The message-variant split at `1e-8`.** Resolved in `5ccf6276` — see the review response
   below; option (c) (parser records the declared variance) was considered and rejected.

Skimmable: the fixture migration (mechanical `FIX` additions, each fixture's asserted numbers
unchanged) and the docs.

## Open questions

**Deliberately not in this PR** (per the plan's section 6):

* Reconciling the `1e-8` eigenvalue floor with the `−6` rail, and the cliff at exactly `−6.00`
  (sticks) vs `−5.93` (recovers). Both are already **#1242**, split from this issue and sitting at
  step 4 of #1217 — no new issue was filed. #1242's design question 2 asks for "the #1228-style
  measurement on the fitting fixtures #1229 migrates"; that set is now measured and small — the
  seven lines in the table above, all of which fit at fixed θ / η = 0, so a floor change would
  have to be re-measured against exactly those.
* Σ at its rail (recovers; no defect measured).
* Warning-vs-error for *tiny non-zero* starts under `focei`×`lbfgs`, which does recover. If that
  ever matters, scope the error to the methods measured to fail rather than weakening it
  globally.

One deviation from the plan worth flagging: the plan asked for the `fit()` `Err` text to carry
`E_OMEGA_INIT_AT_RAIL`. `first_error` returns `Diagnostic::message` alone, and **no** message in
`validation.rs` embeds its own code, so putting it in this one would have been the only such
message in the file. Instead the Tier-2 test asserts the code through `validate_model_file` (the
`ferx check --json` path, which is where a caller can match on it) and asserts byte-identity
between that diagnostic's message and the `fit()` error. Say the word if you'd rather the code
were in the text.


TeunP:
## Review — `xhigh`

The predicate itself is right, and the test file is unusually careful about the three "cannot fail" shapes. My concerns are all about **who else gets judged by it** and **what the message says once it fires**.

---

### 1. Bootstrap `--dofv` silently drops the replicates it exists to measure

`crates/ferx-tools/src/bootstrap/mod.rs:1362-1385`

`compute_delta_ofv` re-enters `fit()` with each replicate's **own estimates** as `init_params`:

```rust
eval.outer_maxiter = 0;
...
let params = params_from_estimates(template, &r.estimates);
let ofv = fit(&prepared.parsed.model, &prepared.population, &params, &eval)
    .map(|f| f.ofv)
    .ok();
```

A replicate that collapsed ω onto the lower rail comes back with variance `exp(-12) = 6.144e-6` — the optimizer cannot return anything smaller, because that *is* the rail. `params_from_estimates` writes it into the matrix, `pack_params` gives `ln(√v) ≈ -6.0`, `packed <= lower` fires, `fit()` returns `Err`, and `.ok()` turns it into a `None` ΔOFV. No warning anywhere.

Those are exactly the replicates the ΔOFV distribution is supposed to characterise, so the diagnostic is biased by dropping them. `--dofv` is opt-in, which caps the blast radius, but the failure is silent.

The general shape of the problem: the check assumes `init_params` is something a **user typed**. Here it is machine output, where a rail-valued variance is a legitimate *result*, and the message ("Add `FIX` if ETA_CL should carry no variability", "start it at ≥ 1e-5") names a declaration nobody wrote.

---

### 2. `outer_maxiter = 0` is rejected, but no optimizer runs there

`src/api/fit.rs:1119`

The gate is unconditional in `fit_inner`, so it also fires for evaluation-only runs. `ferx gam --no-fit` (`crates/ferx-cli/src/gam_cmd.rs:18`) documents itself as:

> `--no-fit` skip the optimization loop; compute EBEs at initial parameter values only (equivalent to NONMEM MAXEVAL=0)

and then does `prepared.parsed.fit_options.outer_maxiter = 0; … ferx_core::fit(…)`. `[fit_options] maxiter = 0` in a model file reaches the same place (`model_parser.rs:7658`).

In that mode `optimize_population` evaluates the objective once at the init and returns — nothing is clamped, no coordinate walks to the opposite rail. The rationale in the doc comment ("the start is clamped onto that rail … from there the coordinate stays collapsed or walks to the opposite rail") does not apply, but the run is refused anyway.

Suggest gating on `options.outer_maxiter > 0` (or, more precisely, on "an outer optimizer will actually run" — the same predicate `gam_cmd::no_fit_supported` already reasons about).

---

### 3. The tiny-non-zero message contradicts itself on a `block_omega`

`src/api/validation.rs:3500-3510`

`variance` comes from `coordinate_values` — the **natural-scale Ω variance** — while the predicate tests the **packed Cholesky factor**. For a diagonal Ω those agree up to a square root; for a block they do not, which is the whole point of the block fixture.

On this PR's own `near_singular_block_flags_the_factor_not_the_declared_variance` input (`block_omega (ETA_CL, ETA_V) = [0.09, 0.089997, 0.09]`), Cholesky succeeds (min eig `3e-6 > 0`, no regularisation), so `omega.matrix[(1,1)]` stays `0.09` while `L₂₂ = 2.45e-3` packs to `-6.01`. Variant B then renders:

> initial variance 9e-2 for ETA_V (`omega ETA_V ~ 9e-2`) packs to ln(L) = -6.01, at or below the optimizer's lower bound of -6.0 — **every variance ≤ 6.14e-06 lands there** — … **Start ETA_V at ≥ 1e-5**, or `FIX` it …

Every clause is wrong for that user: `9e-2` is not `≤ 6.14e-6`; `omega ETA_V ~ 9e-2` is not a line in their file (they wrote `block_omega`); ETA_V is already three orders of magnitude above `1e-5`. And the actual cause — a declared correlation of ρ ≈ 0.99997 — is never named, so the message points away from the fix.

Same shape for an indefinite block: `block_omega (A, B) = [0.0, 0.05, 0.0]` regularises by `-min_eig + 1e-8`, giving diagonals of `5e-2` and `L₂₂ ≈ 1e-4`, so a genuinely declared zero also lands in variant B.

Either quote `L_ii²` (the quantity actually compared against the rail) rather than `coordinate_values[i]`, or give `block_omega` its own message that names the block and the near-singular correlation.

---

### 4. The `1e-8` discriminator misattributes both ways

`src/api/validation.rs:3298-3306, 3484`

The "the user wrote `0.0`" variant is selected by `variance <= OMEGA_REGULARIZATION_FLOOR`. Two ways that inference is wrong:

- **A small declared variance never gets regularised.** `omega ETA_CL ~ 1e-9` on a single-eta model: `[[1e-9]]` is positive-definite, so `from_matrix_with_mask` takes the `Some(c)` arm and `matrix[(0,0)]` stays `1e-9 ≤ 1e-8`. The user is told "`omega ETA_CL ~ 0.0` declares no variability … a zero variance is regularised to `1e-8`" — a declaration they did not write, via a mechanism that did not run.
- **`reg` is only `1e-8` when `min_eig >= 0`.** For an indefinite matrix it is `-min_eig + 1e-8`, so a real declared zero lands well above the floor and takes the *other* message (see #3).

The constant's doc comment asserts the floor unconditionally; it holds only for the `min_eig >= 0` branch of `types.rs:1699-1700`.

---

### 5. A `[mixture] omega(k)` diagnostic is stamped with the wrong block

`src/api/validation.rs:3516`

Every diagnostic gets `.with_block("parameters")`, but a mixture Ω override is declared inside `[mixture]`:

```
[mixture]
nsub = 2
logit(1) = MIXL
omega(2) ETA_CL ~ 0.0
```

`validate_model_file` attaches the owning block's header line, so `ferx check` prints `parameters:<line of [parameters]>` and the JSON report carries `"block": "parameters"` — sending the user (and anything that jumps to `block`/`line`) to a block whose `omega ETA_CL ~ 0.09` is fine.

The suggestion compounds it: ``write `[mixture] omega(2) ETA_CL ~ 0.0 FIX` `` is not a line that parses; the declaration is `omega(2) ETA_CL ~ 0.0 FIX` *inside* the block. (`FIX` is supported there — `model_parser.rs:12613` — so only the spelling is off.)

`free_zero_mixture_omega_override_is_rejected_naming_its_class` asserts only that the message text contains `[mixture] omega(2)`, so neither of these is visible to it.

---

### 6. `ferx check` now fails a model that `simulate()` still runs

`src/api/validation.rs:4469`

The CHANGELOG says, correctly, that "`predict()` / `simulate()` are untouched, so a zero-variance fixture used only for prediction still works" — and `ferx <model> --simulate` does bypass this, since it routes through `run_model_simulate`. But `ferx check model.ferx` on the same file now returns `valid: false` and exit 1.

So the validity gate and the runtime disagree about the same model, and a CI step that gates on `ferx check` fails for a fixture that is valid for its intended use. After this branch, `tests/` still carries 69 such declarations across 25 files — all prediction/simulation-only, which is exactly the population that `ferx check` would now reject. Worth deciding deliberately: warning on the check path, or an accepted consequence stated in the docs.

---

### 7. Test gap — the block fixture cannot see the message it produces

`src/api/tests/variance_init_rail_tests.rs:301`

`near_singular_block_flags_the_factor_not_the_declared_variance` is the one fixture where the declared variance and the packed factor disagree — i.e. exactly where the message-building code reads the wrong one of the two. Its surviving assertions are `contains("ETA_V")` and `!contains("CANNOT BE ZERO")`, both satisfied by the contradictory text in #3.

Per the repo's own rule, the assertion this fixture exists for is missing: that the quoted variance is the quantity actually compared against the rail, or at minimum that the message does not state a variance *and* claim it is `≤ RAIL_VARIANCE`.

---

### 8. Slow-tests not green yet for the fixture edits

Three slow-gated convergence fixtures were rewritten from `omega ~ 0.0` to `~ 0.0 FIX`. That keeps the Ω *value* identical (both regularise to `1e-8`) but changes the free dimension and the bound pinning, and `mixed_fit_recovers_fraction_dur_and_ka` / `parallel_fit_recovers_fraction_and_kas` assert `converged` plus `|θ₂ − 0.4| < 0.06` and `|θ₄ − 3.0|/3.0 < 0.12`.

PR CI shows `Tests + coverage (full, slow-tests): skipping`, and the `Slow tests` run for `36093c7a` was still `in_progress` when I looked. Worth confirming it landed green before merge.

---

### Smaller notes

- **`variance_decl_by_coordinate` is a third copy of the packed-segment offsets** (`src/api/validation.rs:3353`). `packed_len` (`parameterization.rs:494`) and `omega_structural_zero_mask`'s `iov_start` (`:423`) already derive `n_theta + n_omega` / `iov_end`, and `rho_packed_start` is the precedent for "go through this rather than re-deriving the offset". This copy also lives outside the module that owns the layout. Exporting a per-coordinate segment from `parameterization` would leave the api layer only mapping segment → keyword.
- **Eager `coordinate_names` / `coordinate_values` on every fit** (`src/api/validation.rs:3447`). Both are built before we know whether any diagnostic fires — one `String` allocation per coordinate on the path every successful fit takes, including the 200+ replicates in `bootstrap` and each candidate in `search::runner`. `compute_bounds` also recomputes `pack_params` and `packed_fixed_mask` internally, so both run twice per call.
- **Altitude.** The underlying defect is that `pack_params` clamps silently (`.max(1e-10).ln()`) and nothing checks the packed start is inside `compute_bounds`' box. θ has the identical clamp with no diagnostic at all — `theta TVCL(0.05, 0.1, 10.0)` packs to `-3.0` against a lower bound of `-2.3` and is silently clamped; there is no theta-init-vs-bounds check anywhere in `src/`. A `parameterization::coordinates_outside_bounds()` primitive next to `pack_params`/`compute_bounds` would cover θ, Σ, ρ and any future segment in one place, with `api/validation.rs` choosing only severity and wording per `PackedCoordKind` — which is also the natural home for the measured Σ/Ω asymmetry.

---

Blocking, in my view: **1** and **2** (both reject runs where the rail never applies), plus **3**/**4** if the message is going to be the user's only guidance. **5**–**8** and the smaller notes are follow-up-sized.

TeunP:
## Update to finding 8 — the slow-tests dispatch has landed, and the fixture migration is green

When I posted the review, the `Slow tests` run for `36093c7a` was still `in_progress`, so I flagged the three Tier-3 free→FIX fixture edits as unverified. It has since completed. Resolving that finding: **the migration is fine.**

Run [33977461387](https://github.com/FeRx-NLME/ferx-core/actions/runs/33977461387) on `36093c7a`:

| Job | Result |
|---|---|
| `Slow regression tests (core)` | **success** |
| `Slow regression tests (nn)` | failure |

The run-level status reads `failure`, but that is entirely the `nn` leg, which is red on `main` too — same split on `2faeebc5` and `b026b531` (core success / nn failure), i.e. the known #1239 `vi_dcm_omega_recovery` budget overrun, not anything this branch did.

So `parallel_fit_recovers_fraction_and_kas`, `mixed_fit_recovers_fraction_dur_and_ka` and `zero_order_fit_recovers_duration` all converge and hold their θ tolerances under `~ 0.0 FIX`. **Finding 8 is closed** — please disregard it.

Worth noting *why* it passed, since it is the argument that the fixture edit was right rather than merely tolerable: `0.0` and `0.0 FIX` both regularise to the same `1e-8` Ω, so the value path is unchanged; only the free dimension and the bound pinning differ. And NM-TRAN rejects a free `$OMEGA 0` outright (error 76) — which is this PR's whole premise — so the NONMEM side of those anchors was always `0 FIX`. The fixtures were the ones out of step with their own reference, and are now in step.

The other seven findings stand as posted, in particular the two I marked blocking:

1. `crates/ferx-tools/src/bootstrap/mod.rs:1362-1385` — `compute_delta_ofv` judges each replicate's **own estimates** at `outer_maxiter = 0`; a replicate that railed is refused and `.ok()` drops it silently.
2. `src/api/fit.rs:1119` — the gate is unconditional, so `ferx gam --no-fit` and `[fit_options] maxiter = 0` are rejected although no optimizer runs and nothing is clamped.

Neither is affected by the slow-tests result.

TeunP:
## Self-review correction: the eval-only exemption's stated reason was false

Found while enumerating `clamp_to_bounds` call sites for the follow-up issue, not by CI. The
exemption added in 5ccf6276 is right; **the reason I gave for it was wrong**, and I had written
it into a doc comment, a test's assertion message and three docs pages, where it would have
outlived the commit.

**The false claim** (5ccf6276's message, `outer_search_runs`' doc comment, and
`evaluation_only_run_is_not_judged`'s assertion text): *"When nothing optimises, nothing is
clamped and the rejection has no basis."*

Measured: `evaluate_at_initial_params` calls `clamp_to_bounds` before it evaluates, exactly like
the optimizing paths (`estimation/outer_optimizer.rs:418`). Nothing is *optimised* at
`maxiter = 0`; the start is still clamped.

**What is actually true**, and what the comment now says: an eval-only run produces one objective
and stops, so a clamped start is never *held* there by a search. The stickiness is the defect,
not the clamp. The operative justification — `--dofv` and `ferx gam --no-fit` hand `fit()`
machine-produced parameters where a rail-valued variance is a legitimate result — is unaffected,
so the scoping and the predicate are unchanged.

**What the exemption concedes**, now stated instead of glossed:

| spelling | stored | evaluated at `maxiter = 0` |
|---|---|---|
| `omega ETA_CL ~ 0.0` | `1e-8` | `exp(-12) ≈ 6.14e-6` — **614×**, silently |
| `omega ETA_CL ~ 0.0 FIX` | `1e-8` | `1e-8` (within 2 ULP) |

Pre-existing, and not specific to Ω — a `theta TVCL(0.05, 0.1, 10.0)` starts at `0.1` the same
way, on every run, with no diagnostic anywhere. Tracked as #1251 rather than fixed inside a rail
check. `docs/model-file/fit-options.qmd`'s "the parameters are returned unchanged" was wrong for
the same reason and is corrected too.

**The test is the part worth looking at.** My first version composed `pack_params` /
`clamp_to_bounds` by hand and passed — but that asserts *`clamp_to_bounds` clamps*, a tautology,
and would have stayed green if `evaluate_at_initial_params` ever stopped calling it, which is
exactly the change that would make the new comment wrong. It now goes through `fit()`.
Mutation-tested: removing the clamp from the eval-only path kills it, and so does disabling the
`FIX` pin in `compute_bounds`; a no-op control survives.

No behaviour change in this commit — comments, docs, one new test.

---

### Follow-ups filed

- **#1251** — nothing checks that a packed start is inside its own bounds. Ten non-test
  `clamp_to_bounds` sites across four modules, all silent; θ and Σ measured to start outside
  their boxes too. Proposes one `coordinates_outside_bounds()` primitive in
  `parameterization.rs` with `api/validation.rs` choosing only severity and wording — which
  would also let an eval-only run *warn* where a fitting run errors, and would retire the
  packed-offset duplication this PR introduced in `variance_decl_by_coordinate`.
- **#1252** — `compute_bounds` re-packs internally; six real callers do the work twice, on the
  path each bootstrap replicate and model-search candidate takes.

Both assigned and added to the #1217 step-4 group.

### Full-diff review

After the fixes above I read the **complete rebased diff** against `origin/main` — 17 files,
+1780/-11 — rather than trusting the green gates, since none of the defects in this PR have ever
been caught by one. It found two more, both shipped-message defects:

- **A near-singular `block_kappa` was told to fix a `block_omega`.** `is_block_element()` routes
  `Omega { block: true }` and `Kappa { block: true }` to one message arm — right, the cause and
  remedy are identical — but the arm hardcoded `block_omega`, naming a block absent from the
  reader's file. Now a `block_keyword()` accessor.
  The two `near_singular_*` tests asserted `contains("correlation")` and `!contains(">= 1e-5")`,
  and **both pass on either message** — the keyword was the only distinguishing text and neither
  test read it. Both now pin it each way round, with the negative so a message naming both
  cannot pass. Mutation-tested: swapping either arm's keyword kills that arm's test and only
  that one; before this commit the kappa swap was green.
- **The false "nothing is clamped" claim had one more copy** — the inline comment at the early
  return, three lines from the doc comment the previous commit fixed.

Checked and clean: `coordinate_kinds`' segment boundaries (mixture Σ overrides are `Sigma`, ρ is
`OmegaOffDiagonal`, so the mixture arm cannot be reached by either); `fit_inner` sees the whole
`options`, so `outer_search_runs` reads the full method chain; all ten fixture migrations are the
identical minimal `~ 0.0` -> `~ 0.0 FIX` edit; no NaN-eating folds or absorbing `filter_map` in
the new tests. The nested Quarto callout in `parameters.qmd` was verified by **rendering the
page**, not by reasoning about fence nesting.

### Gates

Rebased onto `5635f8a9` (main moved twice during this work). All green on `334a5272`:

- `tools/preflight.sh` — all six groups, public-API baseline unchanged.
- Local `ci,survival` suite — **175 binaries, 0 failures**, cargo exit 0. Re-run untruncated
  after I noticed my first capture ended in `tail -80`: it retained only the last 80 of ~175
  result lines, so a failure in the first half would have been dropped and the pipeline's exit
  status was `tail`'s, not cargo's. The re-run is the one quoted here.
- PR CI — every check passes, including **`Tests (debug-assertions)`** (40m52s), the job main
  added in `73ee77f3` that had never run against this branch: the PR was `CONFLICTING` from the
  start, so no workflow had ever dispatched on it until the rebase.
- `codecov/patch` passes.
- Slow-tests dispatched on this head (run 33997602658): **`Slow regression tests (core)`
  green**. The `nn` leg is red by design under #1239.

Issue #1186, opened 2026-09-01
bug(ode): a derived break time within 1e-10 of a dose applies that dose twice


TeunP:
## Summary

The break-time **dedup** tolerance is `1e-15` (absolute) while the **dose-application match**
window is `1e-10`, and the empty-segment skip uses the same `1e-15` as the dedup. So two break
times that are 1e-15..1e-10 apart both survive dedup, are both processed as real segments, and
both match the same dose — which is then applied **twice**.

Found while self-reviewing #1174 (the #1171 fix). The root cause is **pre-existing on `main`**:
`collect_dose_break_times` has emitted a derived `d.time + lag_cmt + lag_route` break since #859,
so `ode_predictions` — the objective path — can already hit this. #1174 extends the exposure to
`ode_dense_solve_states`, `ode_predictions_with_states` and `ode_solve_until_chz_threshold`.

## Measured

At `efbd3532`, 2-state ODE:

```
d/dt(depot)   = first_order(ka=KA, lag=LAG) - KA*depot
d/dt(central) = KA*depot - CL/V*central
ALAG1 = 0.3,  LAG = 7.9
dose A: 100 mg, CMT=1, t=0      (feeds the forcing; bolus correctly suppressed)
dose B: 100 mg, CMT=2, t=8.2    (plain bolus, no input-rate forcing on cmt 2)
```

The route onset is `(0.0 + 0.3) + 7.9 = 8.200000000000001` — **1.78e-15** from dose B's own break
at `8.2`. One ULP in `[8,16)` is 1.776e-15, so the two are too far apart to dedup at `1e-15` and
comfortably inside the `1e-10` dose match. `u[cmt] += dose.amt * f` runs at both breaks:

```
central(8.2001):  199.998   (LAG = 7.9,  onset collides with dose B)
                  100.120   (LAG = 7.85, onset clear)
```

on all three of `ode_dense_solve_states`, `ode_predictions_with_states` and `ode_predictions`.

## Mechanism

- dedup: `break_times.dedup_by(|a, b| (*a - *b).abs() < 1e-15)` (`predictions.rs:5530`, `:5825`)
- dose match: `if (t_eff - t_start).abs() < 1e-10` (`predictions.rs:5559`, `:5892`)
- empty-segment skip: `if (t_end - t_start).abs() < 1e-15 { continue; }` — the *same* `1e-15`, so
  a surviving near-duplicate is guaranteed to be integrated as a real segment. There is no
[...]

PR #1247, opened 2026-09-05
fix(ode): read an observation within EVENT_MATCH_TOL after a break post-event, one-sided [closes #1226]

### LINKED ISSUE #1226: bug(ode): an observation within 1e-12 after a lagged dose arrival is read pre-dose on the objective / with_states / dense engines (event-driven reads post-dose)
An observation whose time sits within `1e-12` **after** a lagged dose arrival is recorded from the *pre-arrival* segment on `ode_predictions` (objective), `ode_predictions_with_states` (sdtab) and `ode_dense_solve_states` (hazard / `[derived]` / Markov / adaptive AUC). The event-driven predictor reads it post-dose, which is the correct ordering — the arrival strictly precedes the observation.

Found while measuring #1186 (probe `m11`); it is **not** that defect. #1186 is dose *application* (a dose applied twice); this is observation *recording* (a correctly-applied dose read too early). Deliberately kept out of the #1186 PR for that reason — see plan §0(c) on <https://github.com/FeRx-NLME/ferx-core/issues/1186#issuecomment-5543296983>.

## Measured (at `19fdbb04`, `--features ci,survival`)

1-cpt IV, `CL = 1`, `V = 10`, `F = 1`, one 100 mg dose at `t = 7.9` with a compartment lag `ALAG1`. The lag is an *estimated* iterate a few ULP below 0.3 — the reachable route, since an optimizer walks `ALAG` continuously:

```
lag     = 0.29999999999999799
arrival = 7.9 + lag = 8.19999999999999751     (1.78e-15 BEFORE the observation at 8.2)

obs(8.2): objective = 0.000000   with_states = 0.000000   dense = 0.000000   event_driven = 100.000000
obs(9.0): objective = 92.311635  with_states = 92.311635  dense = 92.311635  event_driven =  92.311635
```

Correct is post-dose (`~100`): the arrival is earlier than the observation. Three engines return `0.0` — the pre-dose state — so a subject reads *drug-free* at a sample taken after its dose, and `obs(9.0)` shows the trajectory is otherwise fine, i.e. the dose was applied exactly once and only this one record is misread.

**The mirror geometry is correct on every engine**, so the defect is one-sided. With the arrival 1.78e-15 *after* the observation (`dose 7.9 + ALAG 0.3 = 8.20000000000000107`, and `dose 8.15 + ALAG 0.05` likewise) all four engines read `0.0`, which is right — the observation genuinely precedes the arrival. The five same-instant pairs tried (`8.19 + 0.01`, `8.0 + 0.2`, `8.1 + 0.1`, `2.0 + 6.2`, …) are bit-identical to `8.2` and read post-dose (`100.0`) everywhere. Only "arrival a hair *before* the observation" is wrong.

## Mechanism

The arrival is a break at `8.19999999999999751`. The observation at `8.2` is:

- **not** at that break's bits, so the post-dose left-boundary read (`obs_map.get(&t_start.to_bits())`) misses it;
- **excluded** from the post-arrival segment's `saveat`, whose filter is `t > t_start + 1e-12` (`predictions.rs:2362`, `:6024`) — `8.2` is not greater than `8.19999999999999751 + 1e-12`;
- **included** in the *previous* segment's `saveat`, whose filter is `t <= t_end + 1e-12` with `t_end` = the arrival.

So it is recorded by the segment that ends at the arrival, from the state before the dose is applied. The two filters are asymmetric by design (each interval is half-open, `(t_start, t_end]`), and that asymmetry is exactly what hands a within-tolerance-after observation to the wrong side.

The event-driven predictor is immune for the same reason it is immune to #1186: it dispatches typed events by index off one sorted timeline, so `Kind::Dose` at the arrival sorts before `Kind::Obs` at `8.2` and the read happens after the jump.

Note this is the *opposite* sign of the adjacent case already checked on #1186 and found harmless: a derived break landing `5.3e-15` **before** an observation (probe `m9`) does not lose that observation, because the previous segment's inclusive filter saves it and a derived break only changes the derivative, so a state read 5e-15 early is exact to round-off. Here the break is a *dose*, which changes the state discontinuously, so the same inclusive filter is what breaks it.

## Suggested fix

An observation within `EVENT_MATCH_TOL` **after** a break should be read post-boundary at that break, on every engine — the same tolerance-vs-exact-bits question #1186 settles for dose application, applied to the recording side. Concretely: after the boundary events are applied at `t_start`, record any observation in `[t_start, t_start + EVENT_MATCH_TOL)` that the exact-bit `obs_map` lookup missed, and exclude it from the preceding segment's `saveat` so it is not recorded twice. `EVENT_MATCH_TOL` is `pub(crate)` in `ode::predictions` as of the #1186 fix.

The tolerance direction matters and should not be made symmetric: an observation within tolerance *before* a dose arrival must stay pre-dose (measured above — all four engines already agree there, and NONMEM's event ordering agrees).

## Oracles

- **Internal**: the event-driven predictor already gets this right, so it is a same-repo differential oracle for the fix — and, unlike the #1166 case, the two arms genuinely straddle the predicate here (0.0 vs 100.0), so the pair cannot degenerate into a tautology.
- **External**: NONMEM, whose event handling defines the ordering — a dose record strictly before an observation record is applied first. An ADVAN1 twin with `ALAG1` FIX'd at the measured iterate pins it without needing the float coincidence to be expressible in a control stream (NONMEM applies each record once by type regardless).
- A regression test must assert the **straddle** itself (`0 < obs_time − arrival < EVENT_MATCH_TOL`), or a later change to either time makes the pair bit-identical and the test passes vacuously.

## Scope

Recording only — no dose is applied twice or dropped, and `obs(9.0)` above confirms the rest of the trajectory is unaffected. Reachable through an estimated or covariate-scaled `ALAG`/route lag whose arrival lands a few ULP short of a sample time; on the objective path it moves that record's residual by the full pre/post-dose difference, so it perturbs the OFV rather than only a diagnostic.

TeunP:
## Why

An observation whose time sits within `1e-12` **after** a lagged dose arrival was recorded from the *pre-arrival* segment on `ode_predictions` (objective), `ode_predictions_with_states` (sdtab) and `ode_dense_solve_states` (hazard / `[derived]` / Markov / adaptive AUC). The arrival strictly precedes the observation, so the correct read is post-dose — a subject was reading **drug-free at a sample taken after its own dose**, and on the objective path that moves the OFV rather than only a diagnostic.

Reachable through any estimated or covariate-scaled `ALAG`/route lag: the arrival is a float sum, and an optimizer walks the lag continuously, so arrivals routinely land a handful of ULP from a sample time.

Measured at `2faeebc5` (`--features ci,survival`), 1-cpt IV amount readout, `CL = 1`, `V = 10`, doses at `t = 0` and `t = 7.9`, `ALAG1` searched into the band, samples at 8.2 and 9.0. **Multi-dose deliberately**: the first dose leaves 45.38 in central at 8.2, so the victim dose lands with the compartment non-empty and the pre/post pair is 45.38 vs 145.38 rather than 0 vs 100.

| engine | arrival 1.3e-13 **before** obs | arrival 1.8e-15 **after** obs (mirror) |
|---|---|---|
| `ode_predictions` (objective) | **45.38 (pre-dose, wrong)** | 45.38 ✓ |
| `ode_predictions_with_states` (sdtab) | **45.38 wrong** | 45.38 ✓ |
| `ode_dense_solve_states` (hazard / `[derived]` / AUC) | **45.38 wrong** | 45.38 ✓ |
| `ode_predictions_and_chz` — the #570 shared-solve **CHZ** read | 145.38 ✓ | **145.38 (post-dose, wrong)** |
| `ode_predictions_event_driven` | 145.38 ✓ | 45.38 ✓ |
| analytic `pk::predict_concentration` (×V) | 145.38 ✓ | 45.38 ✓ |
| closed form | 145.384480 | 45.384480 |
| **NONMEM 7.6.0**, ADVAN1 `PRED` | **1.4538447952823367E+02** | **4.5384479528235588E+01** |
| **NONMEM 7.6.0**, ADVAN13 `A(1)/V` | 1.4538447943810009E+01 | 4.5384479438102048E+00 |

Two things that table settles:

- **The mirror sign was wrong in the opposite direction, on exactly one site.** The #570 shared solve's CHZ boundary read was a *symmetric* `(t - t_start).abs() < 1e-12`, so a hazard time 1.8e-15 **before** an arrival was overwritten with the post-dose state while the dedicated dense path kept the pre-dose one. The "shared solve ≡ dedicated path" invariant was already broken on the mirror. Same root cause, opposite sign, one fix.
- **The event-driven engines are untouched controls.** `ode_predictions_event_driven` and the provider's event-driven walk dispatch typed events by index off one sorted timeline and never rescan; they were measured right on **both** signs and this PR does not change a line of either. Every "✓" in their rows is a before-and-after value.

## What changed

One rule, one constant, applied at every rescanning site. Two new `pub(crate)` predicates next to `EVENT_MATCH_TOL` in `ode::predictions`:

| | rule |
|---|---|
| **band** | `t_k <= t < t_k + EVENT_MATCH_TOL` → read **at `t_k`, after** that break's events (reset → dose → read), from `u(t_k⁺)` |
| **segment** | `t_k + EVENT_MATCH_TOL <= t <= t_{k+1}` → read off the integration of `(t_k, t_{k+1}]` |

They partition `[t_k, t_{k+1}]` whenever adjacent breaks are at least `EVENT_MATCH_TOL` apart; when two breaks are closer, both bands can claim a time and the ascending loops make the **later** write win — the right answer, being the latest break at or before `t`.

**One-sided, deliberately.** A record within tolerance *before* a break stays pre-event. That is NONMEM's record ordering, anchored here on both signs, and every engine except the shared-solve CHZ read already did it.

The exact-bit `obs_map.get(&t_start.to_bits())` boundary lookups are **replaced** by a band lookup, not kept alongside them — the band is a superset of the exact hit, so keeping both would record a band member twice. `obs_map` stays for matching the solver's returned save points, whose bits are the `saveat` entries' own.

Both predicates are written as a **difference** (`t - t_break < TOL`), never as the sum `t < t_break + TOL`. `EVENT_MATCH_TOL` is absolute, so the sum rounds back once `ulp(t_break)` exceeds it and the band goes empty — see *Review round 2* below. Records are resolved through one `RecordIndex`: sorted once per subject, `partition_point` per break, shared by all six ODE/EKF sites **and** by `sens/ode_provider.rs`, which previously kept its own sorted copy.

Sites: `integrate_segment`'s `saveat` (shared by the objective, the adaptive driver and the frozen replay); the objective loop's obs read, CHZ read and `seg_chz` filter; the adaptive driver's and frozen replay's obs reads (through the same helper — they are a bit-identity pair); `ode_predictions_with_states`' read + `saveat`; `ode_dense_solve_states`' last-break guard, read and `seg_saveat`; `ode/ekf.rs`' read + `saveat`; and `sens/ode_provider.rs::integrate_g`.

### The provider needed more than the plan's one-line bound

The plan expected `sens/ode_provider.rs` to need only `t <= t_end`. Measurement said otherwise, and the mutation pass below confirms it. `record_at` is **first-write-wins** via `recorded[j]`, and the previous segment's final save point *is* the break being visited, so its pre-event state reached the record first and stuck. A new `record_at_break` records the break's whole band **overwriting** — which is exactly what production does (segment `saveat` writes, next break's boundary read overwrites; production has no `recorded` guard).

That also closes a **pre-existing** gap, found while fixing this one: an observation landing **exactly on an interior dose** already read 155.85 on the static `Dual2` walk against production's 1155.85 — a whole 1000 mg dose, measured at `2faeebc5`, i.e. the FOCEI gradient and the objective disagreeing. `t <= t_end` alone cannot close that one (an observation *equal* to `t_end` is a legitimate member of that segment's `saveat`), which is why the overwrite is the fix and the bound is hygiene.

## Alternatives considered

**Integrate the sub-tolerance span instead of reading at the break.** Both work on RK45 (a 1.3e-13 span integrates in one accepted step; a 1.8e-15 span returns the start state in zero steps). The band read wins because it keeps observations on the same convention as doses — `EVENT_MATCH_TOL` is already "same instant" for every event — never creates a sub-tolerance segment for a record, and does not depend on every stepper's behaviour at `h ≈ ULP` (Rosenbrock unmeasured). The shortcut error is bounded by `|f|·1e-12`, and `band_read_is_continuous_across_the_tolerance_edge` measures it rather than arguing it.

**A wider dedup instead of a recording rule.** Rejected on #1186 for the dose side and for the same reason here: widening the `1e-15` dedup re-segments every engine and trips the adaptive exact-bit guards (#700).

## Cross-repo dependency

No public API change — every new item is `pub(crate)`, and `tools/update-public-api.sh --check` reports the baseline unchanged.

## Breaking changes
- [ ] Public Rust API
- [x] None

Predictions **do** change for models where a lagged arrival lands within `1e-12` of a sample — that is the fix, and the `CHANGELOG` entry says so in user-facing terms.

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0**, both signs, exact
- [ ] Warfarin example (not the object under test)

Committed as `nonmem_anchor/lag_arrival_read_{before,after}_advan{1,13}.{ctl}` + `results/*.{lst,tab}` + `lag_arrival_read.csv`, with a README row after the #1186 one.

Unlike #1186, **the float coincidence is expressible in a control stream**: `ALAG1` is passed as a 17-digit `$THETA … FIX` and each run tables `ARR = TDOS + ALAG1` and `SEP = 8.2 - ARR`, so the comparison is *verified* to be on identical doubles rather than assumed:

```
# results/lag_arrival_read_before_advan1.tab   (ARR 74 ULP below the 8.2 sample)
ARR = 8.1999999999998678      SEP = 1.3145040611561853E-13
PRED(8.2) = 1.4538447952823367E+02      PRED(9.0) = 1.3420678956342570E+02

# results/lag_arrival_read_after_advan1.tab    (ARR 1 ULP above it)
ARR = 8.2000000000000011      SEP = -1.7763568394002505E-15
PRED(8.2) = 4.5384479528235588E+01      PRED(9.0) = 1.3420678956342749E+02

# ADVAN13 TOL=9 twin, IPRED = A(1)/V  (×V to compare as an amount)
before: 1.4538447943810009E+01   9.0: 1.3420678940629491E+01
after:  4.5384479438102048E+00   9.0: 1.3420678940629784E+01
```

`ADVAN1 TRANS2` (no `S1`, so `F = A(1)`) is the exact analytic reference; the `ADVAN13` twin shows the ordering survives NONMEM's own integrator and agrees with `ADVAN1` to 6.2e-10 relative. ferx is asserted against ADVAN1 at `rel < 1e-9` and against ADVAN13×V at `rel < 1e-8` (that run's own `TOL=9` error), **per engine**, so a failure names the engine that drifted. At `t = 9.0` — past both arrivals — the two runs agree to round-off, which is the control that the 100 mg gap at 8.2 is the ordering and nothing else.

## Performance

Records are resolved through a per-subject sorted index (`RecordIndex`) rather than the exact-bit `HashMap` or a rescan. Measured over a full subject walk with `n_breaks == n_records`, sorted index against a linear rescan:

| records | 5 | 10 | 20 | 50 | 100 | 1000 | 5000 |
| speedup | 0.20× | 0.35× | 1.03× | 2.27× | 3.91× | 16.0× | 102.8× |

Below ~20 records the sort does not pay for itself, and that is a deliberate trade: the absolute cost there is +140 ns per subject against an ODE solve of tens of microseconds. Above it the win is the one that matters — `ode_dense_solve_states` is driven by a *grid* (hazard timeline, `[derived]` integral, AUC, `predict_survival` horizon) that routinely runs to hundreds or thousands of points, where a rescan is quadratic. The index buffer is hoisted, so each walk allocates once per subject.

## Tests

New `mod lag_arrival_read_1226` in `src/ode/predictions_tests.rs` (9 tests), plus four provider tests, two EKF tests and one adaptive/replay test. `cargo test --lib --features ci,survival`: **4228 passed, 0 failed, 23 ignored**. `tools/preflight.sh`: `preflight OK — fmt check clippy rustdoc docs public-api`.

### Red before the fix

Written and run first, at `2faeebc5` + tests only:

| test | before |
|---|---|
| 1 `issue_and_mirror_fixtures_straddle_the_band` | green (pure — asserts the fixture, not the fix) |
| 2 `arrival_within_tol_before_obs_reads_post_dose` | **RED** — `ode_predictions` 45.3844795290 vs NONMEM 145.38447952823367 |
| 3 `arrival_within_tol_after_obs_reads_pre_dose` | **RED** — `ode_predictions_and_chz` 145.3844795290 vs NONMEM 45.38447952823559 |
| 4 `pre_and_post_dose_anchors_are_a_hundred_apart` | green (pure) |
| 5a `post_dose_read_holds_across_the_whole_separation_band` | **RED** — `ode_predictions` at sep 3.55e-15 |
| 5b `band_read_is_continuous_across_the_tolerance_edge` | **RED** — `ode_predictions` at sep 9.00e-13 |
| 6 `band_and_segment_predicates_partition_the_interval` | green (pure) |
| 7 `provider_reads_an_obs_one_ulp_after_a_dose_post_dose` | **RED** (production side: 155.85 vs 1155.84) |

The plan predicted red for 2, 3 and 7 only. **5a and 5b were green on the first run and that was a defect in the fixtures, not evidence.** `t_last = max(obs_times)` is always a break, so a lone 8.2 sample *is* the final break and the old exact-bit boundary read caught it post-dose — the "a fast path makes the mutation unreachable" shape CLAUDE.md warns about. Adding a trailing 9.0 record makes 8.2 an ordinary interior time and both went red. The trailing record is now documented as load-bearing in both tests.

Test 7's own straddle assertion also passed vacuously at first, by comparing an **amount** readout against a concentration closed form; it was corrected to compare 155.85 against 1155.85 before being trusted. Test 7 is a differential check between the provider and `compute_predictions_with_tv`, so it cannot be red until production is fixed — the three-state chain is: green at `2faeebc5`, red with production fixed and the provider untouched, green with both.

### Mutations

Each applied alone and reverted; every "dies" line names the engine in the assertion message.

| # | mutation | dies? | what dies |
|---|---|---|---|
| M1 | objective boundary read → exact bits | **yes** | tests 2 / 5a / 5b, naming `ode_predictions` (NaN — the record falls in no segment and no band); plus test 7 on production NaN |
| M2 | `ode_predictions_with_states` boundary read → exact bits | **yes** | tests 2 / 5a / 5b, naming `ode_predictions_with_states` only |
| M3 | `ode_dense_solve_states` boundary read → exact bits | **yes** | tests 2 / 5a / 5b, naming `ode_dense_solve_states` only |
| M4 | `integrate_segment` bound → `t <= t_end + 1e-12` | *survives* | see below |
| M5 | `reads_at_break` made symmetric | **yes** | test 3 (`ode_predictions` 145.38 vs NONMEM 45.38) **and** test 6's one-sidedness assert |
| M6 | drop the provider's `record_at_break` | **yes** | both provider tests, 155.85 vs 1155.85 |
| M7 | provider bound → `t <= t_end + 1e-12` | *survives* | see below |
| M8a | EKF boundary read → exact bits | **yes** | `ekf_reads_an_obs_one_ulp_after_a_dose_post_dose` (reads 0.0) |
| M8b | EKF segment filter → `+1e-12` | *survives* | see below |
| M9 | adaptive driver boundary read → exact bits | **yes** | `adaptive_reads_an_obs_one_ulp_after_a_base_dose_post_dose` (driver NaN) |
| M10 | frozen replay boundary read → exact bits | *survives* | unreachable by construction |

**The three surviving segment-bound mutations (M4, M7 pre-`record_at_break`, M8b) are reported, not hidden.** On the ODE and EKF engines there is no first-write-wins guard, so the band read at the next break simply overwrites whatever the wider bound let through — **the band reads are the fix; the exact bound is hygiene**. It is kept because it stops a record being written twice and stops the solver being handed a `saveat` point outside its own span. This is the same shape as #1186's honestly-documented surviving tolerance mutation, and it is recorded in the `reads_in_segment` doc comment. The one place an equivalent slack *is* load-bearing is the provider, whose `recorded[j]` mask is first-write-wins — which is why it needed the explicit overwrite instead, and M6 is the mutation that proves it.

**M10 survives because the frozen replay pushes every observation time onto its break list unconditionally**, so its band read's *widening* is unreachable: an observation is always its own break there and the exact-bit read never missed. It changes with the driver because the pair must not drift, not because a fixture can separate them. The dense engine's last-break guard is likewise provably equivalent under `t_last = max(saveat)`; it was changed so the guard asks the same question as the read.

**M9 is the one that needed the fixture moved.** The first draft of the adaptive test used the TV path and M9 survived it — and survived the *entire* 4222-test lib suite. The driver pushes observation times onto its break list only when `tv` is set, so on the TV path every observation is its own break and the defect is unreachable; on the constant-covariate path it is an ordinary interior time. Moved to `event_pk = None`, M9 kills it. Without that, the adaptive/replay change would have shipped with no test able to see it.

### Full suites run explicitly

```
lag_arrival_read_1226 (7) + provider_reads_an_obs (2) + adaptive_reads_an_obs (1)
  + ekf_ (8) + break_collision_1186 (13) + frozen_replay (6) + 1
test result: ok. 38 passed; 0 failed; 0 ignored; 0 measured; 4208 filtered out

tests/pktte_prestart_entry.rs        test result: ok. 3 passed; 0 failed
tests/pktte_tdep_nonmem_anchor.rs    test result: ok. 3 passed; 0 failed
tests/ss_chz_nonmem_anchor.rs        test result: ok. 11 passed; 0 failed

cargo test --lib --features ci,survival
test result: ok. 4223 passed; 0 failed; 23 ignored; 0 measured; 0 filtered out
```

`slow-tests.yml` dispatched on this branch — PR CI does not run it, and this change touches ODE prediction code that the slow convergence fits exercise.

- [x] Not applicable (a fix with no new API surface)

## Docs
- [x] `docs/` updated for user-visible changes — new `### What if the arrival lands a few ULP before a sample?` subsection in `docs/faq.qmd`, under the existing "Which record's covariates govern a lagged dose, like NONMEM's `$PK` timing?" section (the only page that already discusses lagged-dose ordering; `grep -rn ALAG docs/` was the check). It states the one-sided rule as a table and carries a "Changed in #1226" callout naming the 45.38 / 145.38 anchor.
- [x] No new pages, so `docs/_quarto.yml` is unchanged; `docs-lint` is green via `tools/preflight.sh docs`.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `record_at_break` is a recording site, not an arithmetic one; it copies the same `T` jet `record_at` copies. `provider_reads_an_obs_{one_ulp_after,exactly_at}_a_dose_post_dose` both run `check_full_provider_vs_fd` (value, ∂/∂η, ∂²/∂η², ∂/∂θ, ∂²/∂η∂θ against central FD of the production predictor)
- [x] `Cargo.toml` version — not breaking, unchanged

### ferx-site
- [x] No new DSL syntax, `[fit_options]` key, example or data file
### ferx-book
### Example execution
- [x] No example execution step needed (fix with no user-visible API change; the behaviour change is covered by the committed NONMEM anchors)

## Reviewer hints

Focus on three things:

1. **`sens/ode_provider.rs`.** It is the one file whose change is *not* the plan's, and the reasoning is the `recorded[j]` first-write-wins mask. The pre-existing obs-exactly-on-a-dose gap it closes (155.85 vs 1155.85) is worth an independent look — it predates #1226 and nothing in the suite was catching it.
2. **The surviving mutations.** M4 / M7 / M8b keep a change the tests cannot see (re-measured against the current suite: still 4228 green). The argument is in the `reads_in_segment` doc comment, which now separates the two bounds — the *lower* one is not hygiene, since in sum form it admits `t == t_start` into that segment's own `saveat` at large times.
3. **The one-sidedness.** `reads_at_break` is `t >= t_break && t < t_break + tol`, and M5 shows both the mirror anchor and the predicate test fail if it becomes symmetric. That asymmetry is the whole convention, and NONMEM is the reason for it, not taste.

Skimmable: the five `.ctl`/`.lst`/`.tab` anchor files (their values are quoted above and asserted in the test module), and the mechanical `obs_map.get(...)` → `RecordIndex::records_at_break(...)` substitution at the engine sites.

## Review round 2 (`xhigh`) — all six items

Each reproduced by measurement before being fixed, each fix mutation-verified after. Commits `98629463`, `abf5e686`, `6dfb5e68`, `d9190e59`.

**The central finding was right, and worse than reported.** `EVENT_MATCH_TOL` is absolute and both predicates built their bound by *adding* it, so the band went empty once `ulp(t_break)` exceeded the tolerance — failing even at bit equality, which the `obs_map` lookups it replaced always matched:

| `t_break` | `ulp` | band width, sum form | difference form |
|---|---|---|---|
| 8.2 | 1.776e-15 | 563 ulp | 563 ulp |
| 8192 | 1.819e-12 | 1 ulp | 1 ulp |
| **16384** | 3.638e-12 | **0 — matches nothing** | 1 ulp (bit equality) |
| **17520** | 3.638e-12 | **0** | 1 ulp |

Both now use a difference, the shape of the dose-application predicate that never had this failure mode; it degrades to exactly the bit-equal set instead of to nothing. Measured at `t = 17520` under the sum form: `ode_predictions` reads the **pre-dose 44.04** instead of 144.04 — the original #1226 defect, back at large times, on the objective as well as the dense engine's `NaN`. Item 5 falls out structurally.

| item | verdict | evidence |
|---|---|---|
| 1 — dense last-break skip | **confirmed, fixed** | reproduced at `t = 17520`; the objective is hit too, not only the dense engine |
| 2 — EKF double assimilation | **confirmed, fixed** | `p_obs` 0.0437 vs the correct 0.3460; knock-on at the next record 0.4420 vs 0.4514. Predates #1226 |
| 3 — provider stops one break short | **confirmed, fixed** | dose on the last observation: provider **3.79** vs production **1003.79** |
| 4 — `record_at`'s ±2e-9 window | **does not reproduce** | 155.847 on both sides; record-before-first-break 0.0 on both. `recorded[j]` covers it — it was a comment over-claim, now corrected |
| 5 — `>=` vs `>` | **fixed structurally** | `t - t_start >= TOL` implies `t > t_start` at every magnitude |
| 6 — one time magnitude | **fixed** | `the_band_still_matches_bit_equality_at_large_times`, 7 magnitudes |

**Item 2 taught me something.** My first mask was wrong: gating the whole boundary read made `ipred` read **68.73 instead of 168.73** — reintroducing the very defect this PR fixes. The mask must be covariance-only; the mean still takes the post-event write. A bolus shifts the state deterministically and leaves `p_mat` untouched, so the split is sound, and it is asserted both ways.

New tests, each mutation-verified in both directions:

```
the_band_still_matches_bit_equality_at_large_times            (pure, 7 magnitudes)
every_engine_reads_a_record_on_its_final_break_at_a_large_time (NONMEM-anchored)
provider_reads_a_dose_landing_on_the_last_observation
provider_keeps_an_obs_one_ulp_before_a_dose_pre_dose
ekf_assimilates_a_measurement_on_a_dose_break_exactly_once
```

**A fifth NONMEM anchor, run for this round.** The new fixes were on closed forms only, and both are plain record-ordering questions NONMEM can express — so `nonmem_anchor/lag_arrival_read_largetime.ctl` puts a dose and a sample on the same record time at `TIME = 17520` (two years in hours, where `ulp` is 3.638e-12 and the absolute band would have collapsed) and tables `PRED = 1.4404316545059672E+02` — post-dose, exactly as at `t ≈ 8.2`. It does double duty as the ordering anchor for item 3. Closed form and anchor are asserted to agree so neither can drift.

The EKF `p_obs` has no NONMEM analogue — there is no EKF there. Its oracle needs no reference implementation: a bolus does not touch `p_mat`, so nudging the dose 1e-6 h off the sample must leave every `p_obs` unchanged. A double assimilation violates that by 8×.

**Smaller items**, all taken: `RecordIndex` unifies the band rule with the provider's sorted lookup (benchmarked above — the reviewer was right where it matters and wrong below ~20 records, and the trade is stated); the FAQ rule table no longer reads as exhaustive while omitting the ordinary case, and now says the `1e-12` band is absolute and narrows with magnitude; both `if !boundary_obs.is_empty()` guards dropped so all three sites are spelled alike; the EKF duplicate-time `continue` is gone — those records are now *filled* from the single update rather than dropped, so they no longer keep a finite, plausible `0.0`.

**On altitude.** Agreed, and not attempted here: the root cause is resolving records by float comparison against a rebuilt break list, and the band is a tolerance patch on that. Worth its own issue.

## Open questions

None. The original open question — whether to keep the exact segment upper bound when three mutations of it survive — was answered in review: keep it, for the `>` vs `>=` reason. That reason is now structural rather than incidental: in difference form `t - t_start >= TOL` implies `t > t_start` at every magnitude, so the `(t_start, t_end]` contract holds by construction.


TeunP:
## Review — `xhigh`

The measurement work here is excellent: both signs anchored on NONMEM, the coincidence *expressed* in the control stream rather than assumed (`ARR`/`SEP` tabled), the surviving mutations reported instead of hidden, and two fixture defects (5a/5b green for the wrong reason, test 7's vacuous straddle) caught and documented. The diagnosis is right and the one-sided rule is the right rule.

The problem is the **spelling** of the rule, not the rule. `EVENT_MATCH_TOL` is an *absolute* `1e-12`, and `reads_at_break` builds its band by adding it to `t_break`. Measured:

```
t          ulp(t)      t + 1e-12 == t     band width
8.2        1.776e-15   false              563 ulp
1000.0     1.137e-13   false                9 ulp
8192.0     1.819e-12   false                1 ulp
16384.0    3.638e-12   TRUE                 0 ulp   <-- band is empty
17520.0    3.638e-12   TRUE                 0 ulp
100000.0   1.455e-11   TRUE                 0 ulp
```

For `t_break >= 16384` the sum rounds back and the half-open band `[t_break, t_break + tol)` is **empty** — `reads_at_break(t, t)` is `t >= t && t < t`, i.e. `false`. The predicate stops matching *even at bit equality*, which is exactly what the `obs_map.get(&t_start.to_bits())` lookups it replaced always matched, at any magnitude. 16384 is reachable: hours over a ~2-year study (17520 h), or minutes over ~11 days.

Every fixture in `lag_arrival_read_1226` sits at `t ≈ 8.2`, where the band is 563 ulp wide, so the 4223-test run and the whole M1–M10 table are blind to this.

---

### 1. `ode_dense_solve_states` breaks out one iteration early — NaN or pre-dose (`src/ode/predictions.rs:6944`)

The last-break skip went from the magnitude-independent

```rust
if next.is_none() && !saveat_map.contains_key(&t_start.to_bits()) {
```

to

```rust
if next.is_none() && !saveat.iter().any(|&t| reads_at_break(t, t_start)) {
```

Where the band collapses, `any` is always `false`, so the loop `break`s before the final break's dose is applied and before its grid points are read:

- **single-break timeline** (`break_times.len() == 1`, #1218's shape): `result[i]` keeps its `f64::NAN` prefill — `predict_survival` / `[derived]` / Markov return the silent non-answer #1218 was filed to remove;
- **multi-break**: the point keeps the previous segment's value, i.e. the state *before* the final break's dose — reverting #731.

Both are correct on `main`. This is the one I'd fix before merging.

### 2. The EKF assimilates one measurement two or three times (`src/ode/ekf.rs:252`)

On the ODE engines a band write is an idempotent index assignment, so "the later write wins" is genuinely the right answer for two breaks closer than `EVENT_MATCH_TOL`. On the EKF walk it is not a write — each band member runs a `kalman_update` that mutates `p_mat`.

`break_times` is deduped at `1e-15` while the band is `1e-12` wide, so breaks at `T` and `T + 5e-13` — #1186's own infusion-end/dose collision geometry, which this loop's comment explicitly says it shares — both claim an observation at `T + 6e-13`. Two `kalman_update`s for one measurement. And because `t_end` then lies inside `t_start`'s band, the pushed `saveat` point at `t_end` gets it a third time in the segment loop. The result is an over-confident `p_obs` and a distorted covariance for the rest of the subject.

`band_and_segment_predicates_partition_the_interval` asserts exactly this double claim (`reads_at_break(t, a) && reads_at_break(t, b)`) and calls it correct — true for a write, not for a stateful update. Suggest an `assimilated: Vec<bool>` mask, the analogue of the `applied` dose mask directly above it.

### 3. The provider fix stops one break short (`src/sens/ode_provider.rs:6723`)

`integrate_g` walks `for w in 0..(break_times.len() - 1)`, so the last break is only ever a segment *end*, never a `t_start`. Neither the dose application (`:6762`) nor the new `record_at_break` (`:6774`) runs there.

`t_last = max(obs_times)` is pushed as a break (`:6633`/`:6658`). A subject whose last dose lands on its last observation — dose and sample both at `t = 24` — dedups to one final break: the static `Dual2` walk never applies that dose and records the observation from the last segment's `t_end` save point, pre-dose, while `ode_predictions` (`0..len` with a `k + 1 < len` guard) applies it and reads post-dose.

That is the same 155.85-vs-1155.85 gradient-vs-objective divergence this PR reports closing, still open one break later. Both new provider tests use an *interior* dose, so nothing covers it. Production's `0..len` + `w + 1 < len` shape is the fix.

### 4. `record_at`'s ±2e-9 relative window contradicts the one-sidedness claim (`src/sens/ode_provider.rs:6774`)

The comment above `record_at_break` says "[`reads_at_break`] never claims a time *before* the break, so an observation a hair earlier keeps the pre-event state the preceding segment gave it." But `record_at(t_start, …)` runs on the line immediately above with a *symmetric* window widened by `slack = 2e-9 * (1.0 + q.abs())` — ~2000× `EVENT_MATCH_TOL` — and writes the same post-event `u`.

The `recorded[j]` mask covers the normal case (the preceding segment's save point got there first). It does not cover a record before the first break, or one whose preceding segment was skipped by `(t_end - t_start).abs() < 1e-15`. That is the mirror defect this PR fixes on the shared-solve CHZ read, still live on the sensitivity path and three orders of magnitude wider — and both new provider tests place the observation at or *after* the dose, so the mirror sign is untested there.

### 5. `reads_in_segment` uses `>=` where all five filters it replaced used `>` (`src/ode/predictions.rs:123`)

```rust
t >= t_start + EVENT_MATCH_TOL && t <= t_end
```

The strictness was load-bearing. Once `t_start + tol == t_start`, `>=` admits a record at exactly `t_start` into that segment's own `saveat` — handing `solve_ode` a save point equal to `t0`, outside the `(t_start, t_end]` contract the surrounding comments state.

It survives today only by coincidence: `solve_ode` clamps its first step to `min_dt` (`solver.rs:1024`, `(saveat[0] - t)` is 0) and saves only if `(t - saveat[0]).abs() < 1e-12` after `t += dt_eff` (`:1085`) — and the default `min_dt` is *also* `1e-12`, so at those magnitudes `t += min_dt` is a no-op and the comparison sees exactly 0. `min_dt` is user-settable; raise it above the ulp there and the linear `save_idx` cursor wedges at index 0, after which no save point in the segment is ever matched and all of them freeze-pad with the final state (`solver.rs:1197`). Keeping the strict `>` costs nothing.

### 6. Test coverage: one time magnitude

`assert!(reads_at_break(t0, t0))` at `t0 = 16384.0` fails today. So would a dense-solve fixture whose last break sits there. Neither can be seen by anything in the module, because every fixture — including the pure predicate test — is anchored at `t ≈ 8.2`.

Per CLAUDE.md's "for each new assertion, name the regression it exists to catch and check that regression can actually reach it": the band's *width* is an assertion the tests make implicitly and never check.

---

### Smaller

- **`src/ode/ekf.rs:258`** — the new `if obs_map.get(&obs_times[obs_idx].to_bits()) != Some(&obs_idx) { continue; }` drops every duplicate-time observation except the one `obs_map` kept. Those indices are also excluded from the segment `saveat` by the band, so they are never written at all and keep the prefill `EkfObsPoint { ipred: 0.0, p_obs: 0.0 }` — a finite, plausible zero flowing into the likelihood. The behaviour predates the PR, but the explicit `continue` now encodes the drop as intentional; a `debug_assert` would make it visible.
- **`src/ode/predictions.rs:135`** — `collect_records_at_break` is an O(n_records) scan per break on the objective hot path, while `ode_provider.rs:6584` in this same PR does the identical job with `partition_point` over a `sorted_obs` built once per subject. Same rule, two spellings; the binary-search one is both cheaper and the one you already wrote.
- **`docs/faq.qmd:589`** — the rule table reads as exhaustive but omits the ordinary case (arrival >1e-12 before the sample → post-dose), which contradicts the sentence directly above it. One row — "arrival **before** the sample, by any amount → post-dose" — states the actual rule; the `1e-12` band then belongs only in the paragraph explaining how the sub-tolerance case is evaluated.
- **`src/ode/predictions.rs:3233` / `:6423`** — `if !boundary_obs.is_empty()` guards a function whose body is already a no-op on an empty slice, and the third site (`:6971`) omits it, so the same call is spelled two ways in one file.

### Altitude

Worth saying once, not as a merge blocker: the rule is now installed at eight independent rescanning sites, each with its own hoisted buffer and its own comment — and the PR's own measurement says the two engines that needed *no* change (`ode_predictions_event_driven`, the provider's `integrate_tvcov_g`) were right on both signs precisely because they dispatch typed events by index off one sorted timeline and never rescan. The root cause is resolving records by float comparison against a rebuilt break list; the band is a tolerance patch on that, and it inherits the tolerance's failure modes at all eight sites at once. Two of the eight already got it subtly wrong (2 and 3 above). Not this PR's job to move the rescanning engines onto typed dispatch, but that is what removes the predicate rather than propagating it.

---

**On the open question** (keeping M4/M7/M8b's exact segment bound): keep it, but for a reason the PR does not give — the `>` vs `>=` in item 5. One rule with one spelling is right; the spelling should be the strict one.

TeunP (commented):
## Review — `xhigh`

The measurement work here is excellent: both signs anchored on NONMEM, the coincidence *expressed* in the control stream rather than assumed (`ARR`/`SEP` tabled), the surviving mutations reported instead of hidden, and two fixture defects (5a/5b green for the wrong reason, test 7's vacuous straddle) caught and documented. The diagnosis is right and the one-sided rule is the right rule.

The problem is the **spelling** of the rule, not the rule. `EVENT_MATCH_TOL` is an *absolute* `1e-12`, and `reads_at_break` builds its band by adding it to `t_break`. Measured:

```
t          ulp(t)      t + 1e-12 == t     band width
8.2        1.776e-15   false              563 ulp
1000.0     1.137e-13   false                9 ulp
8192.0     1.819e-12   false                1 ulp
16384.0    3.638e-12   TRUE                 0 ulp   <-- band is empty
17520.0    3.638e-12   TRUE                 0 ulp
100000.0   1.455e-11   TRUE                 0 ulp
```

For `t_break >= 16384` the sum rounds back and the half-open band `[t_break, t_break + tol)` is **empty** — `reads_at_break(t, t)` is `t >= t && t < t`, i.e. `false`. The predicate stops matching *even at bit equality*, which is exactly what the `obs_map.get(&t_start.to_bits())` lookups it replaced always matched, at any magnitude. 16384 is reachable: hours over a ~2-year study (17520 h), or minutes over ~11 days.

Every fixture in `lag_arrival_read_1226` sits at `t ≈ 8.2`, where the band is 563 ulp wide, so the 4223-test run and the whole M1–M10 table are blind to this.

---

### 1. `ode_dense_solve_states` breaks out one iteration early — NaN or pre-dose (`src/ode/predictions.rs:6944`)

The last-break skip went from the magnitude-independent

```rust
if next.is_none() && !saveat_map.contains_key(&t_start.to_bits()) {
```

to

```rust
if next.is_none() && !saveat.iter().any(|&t| reads_at_break(t, t_start)) {
```

Where the band collapses, `any` is always `false`, so the loop `break`s before the final break's dose is applied and before its grid points are read:

- **single-break timeline** (`break_times.len() == 1`, #1218's shape): `result[i]` keeps its `f64::NAN` prefill — `predict_survival` / `[derived]` / Markov return the silent non-answer #1218 was filed to remove;
- **multi-break**: the point keeps the previous segment's value, i.e. the state *before* the final break's dose — reverting #731.

Both are correct on `main`. This is the one I'd fix before merging.

### 2. The EKF assimilates one measurement two or three times (`src/ode/ekf.rs:252`)

On the ODE engines a band write is an idempotent index assignment, so "the later write wins" is genuinely the right answer for two breaks closer than `EVENT_MATCH_TOL`. On the EKF walk it is not a write — each band member runs a `kalman_update` that mutates `p_mat`.

`break_times` is deduped at `1e-15` while the band is `1e-12` wide, so breaks at `T` and `T + 5e-13` — #1186's own infusion-end/dose collision geometry, which this loop's comment explicitly says it shares — both claim an observation at `T + 6e-13`. Two `kalman_update`s for one measurement. And because `t_end` then lies inside `t_start`'s band, the pushed `saveat` point at `t_end` gets it a third time in the segment loop. The result is an over-confident `p_obs` and a distorted covariance for the rest of the subject.

`band_and_segment_predicates_partition_the_interval` asserts exactly this double claim (`reads_at_break(t, a) && reads_at_break(t, b)`) and calls it correct — true for a write, not for a stateful update. Suggest an `assimilated: Vec<bool>` mask, the analogue of the `applied` dose mask directly above it.

### 3. The provider fix stops one break short (`src/sens/ode_provider.rs:6723`)

`integrate_g` walks `for w in 0..(break_times.len() - 1)`, so the last break is only ever a segment *end*, never a `t_start`. Neither the dose application (`:6762`) nor the new `record_at_break` (`:6774`) runs there.

`t_last = max(obs_times)` is pushed as a break (`:6633`/`:6658`). A subject whose last dose lands on its last observation — dose and sample both at `t = 24` — dedups to one final break: the static `Dual2` walk never applies that dose and records the observation from the last segment's `t_end` save point, pre-dose, while `ode_predictions` (`0..len` with a `k + 1 < len` guard) applies it and reads post-dose.

That is the same 155.85-vs-1155.85 gradient-vs-objective divergence this PR reports closing, still open one break later. Both new provider tests use an *interior* dose, so nothing covers it. Production's `0..len` + `w + 1 < len` shape is the fix.

### 4. `record_at`'s ±2e-9 relative window contradicts the one-sidedness claim (`src/sens/ode_provider.rs:6774`)

The comment above `record_at_break` says "[`reads_at_break`] never claims a time *before* the break, so an observation a hair earlier keeps the pre-event state the preceding segment gave it." But `record_at(t_start, …)` runs on the line immediately above with a *symmetric* window widened by `slack = 2e-9 * (1.0 + q.abs())` — ~2000× `EVENT_MATCH_TOL` — and writes the same post-event `u`.

The `recorded[j]` mask covers the normal case (the preceding segment's save point got there first). It does not cover a record before the first break, or one whose preceding segment was skipped by `(t_end - t_start).abs() < 1e-15`. That is the mirror defect this PR fixes on the shared-solve CHZ read, still live on the sensitivity path and three orders of magnitude wider — and both new provider tests place the observation at or *after* the dose, so the mirror sign is untested there.

### 5. `reads_in_segment` uses `>=` where all five filters it replaced used `>` (`src/ode/predictions.rs:123`)

```rust
t >= t_start + EVENT_MATCH_TOL && t <= t_end
```

The strictness was load-bearing. Once `t_start + tol == t_start`, `>=` admits a record at exactly `t_start` into that segment's own `saveat` — handing `solve_ode` a save point equal to `t0`, outside the `(t_start, t_end]` contract the surrounding comments state.

It survives today only by coincidence: `solve_ode` clamps its first step to `min_dt` (`solver.rs:1024`, `(saveat[0] - t)` is 0) and saves only if `(t - saveat[0]).abs() < 1e-12` after `t += dt_eff` (`:1085`) — and the default `min_dt` is *also* `1e-12`, so at those magnitudes `t += min_dt` is a no-op and the comparison sees exactly 0. `min_dt` is user-settable; raise it above the ulp there and the linear `save_idx` cursor wedges at index 0, after which no save point in the segment is ever matched and all of them freeze-pad with the final state (`solver.rs:1197`). Keeping the strict `>` costs nothing.

### 6. Test coverage: one time magnitude

`assert!(reads_at_break(t0, t0))` at `t0 = 16384.0` fails today. So would a dense-solve fixture whose last break sits there. Neither can be seen by anything in the module, because every fixture — including the pure predicate test — is anchored at `t ≈ 8.2`.

Per CLAUDE.md's "for each new assertion, name the regression it exists to catch and check that regression can actually reach it": the band's *width* is an assertion the tests make implicitly and never check.

---

### Smaller

- **`src/ode/ekf.rs:258`** — the new `if obs_map.get(&obs_times[obs_idx].to_bits()) != Some(&obs_idx) { continue; }` drops every duplicate-time observation except the one `obs_map` kept. Those indices are also excluded from the segment `saveat` by the band, so they are never written at all and keep the prefill `EkfObsPoint { ipred: 0.0, p_obs: 0.0 }` — a finite, plausible zero flowing into the likelihood. The behaviour predates the PR, but the explicit `continue` now encodes the drop as intentional; a `debug_assert` would make it visible.
- **`src/ode/predictions.rs:135`** — `collect_records_at_break` is an O(n_records) scan per break on the objective hot path, while `ode_provider.rs:6584` in this same PR does the identical job with `partition_point` over a `sorted_obs` built once per subject. Same rule, two spellings; the binary-search one is both cheaper and the one you already wrote.
- **`docs/faq.qmd:589`** — the rule table reads as exhaustive but omits the ordinary case (arrival >1e-12 before the sample → post-dose), which contradicts the sentence directly above it. One row — "arrival **before** the sample, by any amount → post-dose" — states the actual rule; the `1e-12` band then belongs only in the paragraph explaining how the sub-tolerance case is evaluated.
- **`src/ode/predictions.rs:3233` / `:6423`** — `if !boundary_obs.is_empty()` guards a function whose body is already a no-op on an empty slice, and the third site (`:6971`) omits it, so the same call is spelled two ways in one file.

### Altitude

Worth saying once, not as a merge blocker: the rule is now installed at eight independent rescanning sites, each with its own hoisted buffer and its own comment — and the PR's own measurement says the two engines that needed *no* change (`ode_predictions_event_driven`, the provider's `integrate_tvcov_g`) were right on both signs precisely because they dispatch typed events by index off one sorted timeline and never rescan. The root cause is resolving records by float comparison against a rebuilt break list; the band is a tolerance patch on that, and it inherits the tolerance's failure modes at all eight sites at once. Two of the eight already got it subtly wrong (2 and 3 above). Not this PR's job to move the rescanning engines onto typed dispatch, but that is what removes the predicate rather than propagating it.

---

**On the open question** (keeping M4/M7/M8b's exact segment bound): keep it, but for a reason the PR does not give — the `>` vs `>=` in item 5. One rule with one spelling is right; the spelling should be the strict one.

TeunP:
## Response — all six items, plus two corrections to my own claims

Every item reproduced by measurement before being fixed, every fix mutation-verified after. Four commits: `98629463`, `abf5e686`, `6dfb5e68`, `d9190e59`.

### The central finding was right, and worse than the review states

Confirmed exactly:

| `t_break` | `ulp` | band width, sum form | difference form |
|---|---|---|---|
| 8.2 | 1.776e-15 | 563 ulp | 563 ulp |
| 1000 | 1.137e-13 | 9 ulp | 9 ulp |
| 8192 | 1.819e-12 | 1 ulp | 1 ulp |
| **16384** | 3.638e-12 | **0 — matches nothing** | 1 ulp (bit equality) |
| **17520** | 3.638e-12 | **0** | 1 ulp |

The root cause is the *sum*. The dose-application predicate this was modelled on uses a **difference** — `(dose.time - t_start).abs() < TOL` — which is why that one never had the failure mode: it degrades to bit equality, never to nothing. Both predicates now do the same.

One correction to the review's framing, in the direction of worse: the consequence is not confined to `ode_dense_solve_states`. Under the sum form at `t = 17520`, **`ode_predictions` reads the pre-dose 44.04 instead of 144.04** — the original #1226 defect, back on the objective, at an ordinary trial horizon. The dense engine's `NaN` is the second symptom, not the first.

Item 5 then falls out structurally rather than by choice: `t - t_start >= TOL` implies `t > t_start` at every magnitude, so the `(t_start, t_end]` contract holds by construction. Your "keep it, but for the `>` vs `>=` reason" is the right call and no longer rests on a tolerance being small enough.

### Item-by-item

| item | verdict | evidence |
|---|---|---|
| 1 — dense last-break skip | **confirmed, fixed** | as above; the objective is hit too |
| 2 — EKF double assimilation | **confirmed, fixed** | `p_obs` 0.0437 vs the correct 0.3460; knock-on at the next record 0.4420 vs 0.4514 |
| 3 — provider stops one break short | **confirmed, fixed** | dose on the last observation: provider **3.79** vs production **1003.79** |
| 4 — `record_at`'s ±2e-9 window | **does not reproduce** | see below |
| 5 — `>=` vs `>` | **fixed structurally** | difference form |
| 6 — one time magnitude | **fixed** | new pure test across 7 magnitudes |

**Item 3 is the severe one.** `integrate_g` walked `0..len - 1`, so the final break was only ever a segment *end*, never a `t_start` — neither the bolus application nor `record_at_break` ran there. With `t_last = max(obs_times)` itself a break, a subject whose last dose lands on its last observation had that dose silently never applied: the whole 1000 mg missing from the gradient while the objective had it. Now `0..len` with a `w + 1 < len` guard, matching `ode_predictions`' own #731 shape.

**Item 2 taught me something.** My first mask was wrong. Gating the whole boundary read on it made `ipred` read **68.73 instead of 168.73** — reintroducing the very defect this PR exists to fix. The mask has to be covariance-only; the mean still takes the post-event write. A bolus shifts the state deterministically and leaves `p_mat` untouched, so the split is sound, and the test asserts both halves. Worth noting the double assimilation **predates #1226** — the old exact-bit lookup matched an observation at `t_start` just as the band does — but the band widens what can reach it, so it belongs here.

**Item 4 does not reproduce as a wrong answer.** Both named holes were measured: an observation one ULP *before* a dose reads the pre-dose 155.847 on the provider and on production alike, and a record before the first break reads 0.0 on both. `recorded[j]` covers it, as you allowed it might. So this was a comment over-claim rather than a defect — but the comment was genuinely wrong and is now corrected to say which part is one-sided, that the ±2e-9 catch-all is #410's and stays, and what actually keeps it off the mirror. `provider_keeps_an_obs_one_ulp_before_a_dose_pre_dose` now pins the mirror side; you were right that no provider test placed an observation there.

### Two corrections to claims I made

1. I reported the item-4 comment as corrected when I had only added the test. It was still reading "One-sided, like every other recording site". Fixed in `d9190e59`.
2. The surviving-mutation note in `reads_in_segment` carried a measurement taken *before* the difference-form rewrite and the large-time tests. Re-run against the current suite rather than left to stand: restoring `t <= t_end + 1e-12` still leaves all 4228 green, so the claim holds — and the doc now separates the two bounds, because the lower one is **not** hygiene.

### Smaller items — all taken

- **`RecordIndex`.** Unified, and you were right where it matters. Measured over a full subject walk, sorted index vs linear rescan: `0.20×` at 5 records, `1.03×` at 20, `3.91×` at 100, `16×` at 1000, `103×` at 5000. So it *loses* below ~20 records — a deliberate trade, since +140 ns per subject is noise against an ODE solve of tens of microseconds, while `ode_dense_solve_states` runs on grids of hundreds to thousands where the rescan was quadratic. `sens/ode_provider.rs` now drops its private sorted copy and uses the same type, so the rule has one spelling.
- **EKF duplicate-time drop.** Rather than a `debug_assert`, those records are now *filled* from the single update instead of dropped, so they no longer keep a finite, plausible `EkfObsPoint { 0.0, 0.0 }` flowing into the likelihood.
- **`docs/faq.qmd`.** The table no longer reads as exhaustive while omitting the ordinary case; the rule is stated without a tolerance in it, and the `1e-12` appears only in *how* the sub-tolerance case is evaluated — including that it is absolute and spans 563 representable times at 8.2 h, nine at 1000 h, and none but the sample itself past ~16384 h.
- **`is_empty` guards.** Both dropped; all three sites spelled alike.

### A fifth NONMEM anchor, run for this round

The new fixes were pinned on closed forms only, and both are plain record-ordering questions NONMEM can express — so I ran it. `nonmem_anchor/lag_arrival_read_largetime.ctl` puts a dose and a sample on the same record time at `TIME = 17520` (two years in hours, where `ulp` is 3.638e-12 and the absolute band would have collapsed):

```
PRED(17520) = 1.4404316545059672E+02      # post-dose, exactly as at t ~ 8.2
```

It does double duty as the ordering anchor for item 3 — a dose landing on the same time as an observation — which is what `provider_reads_a_dose_landing_on_the_last_observation` checks the sensitivity walk against. Closed form and anchor are asserted to agree, so neither fixture can drift from the other unnoticed.

The EKF `p_obs` has no NONMEM analogue; there is no EKF there. Its oracle needs no reference implementation: a bolus does not touch `p_mat`, so nudging the dose 1e-6 h off the sample must leave every `p_obs` unchanged, and a double assimilation violates that by a factor of 8.

### On altitude

Agreed, and deliberately not attempted here. The root cause is resolving records by float comparison against a rebuilt break list; the band is a tolerance patch on that and inherits its failure modes at every site — which this round demonstrated twice over. `RecordIndex` at least makes it one patch in one place rather than eight. Worth its own issue for moving the rescanning engines onto typed dispatch.

### State

`cargo test --lib --features ci,survival`: **4228 passed, 0 failed, 23 ignored**. `tools/preflight.sh`: `preflight OK — fmt check clippy rustdoc docs public-api`. `pktte_tdep` / `ss_chz` / `pktte_prestart_entry` anchors: 3 / 11 / 3. PR CI green; slow-tests re-dispatched on the current commit (`core` passed on the previous round; `nn` fails the same three ways as `main` — #1239).

PR #1249, opened 2026-09-05
docs: five measured testing rules that CLAUDE.md did not yet carry

TeunP:
## Why

Seven testing rules were living in my per-session memory instead of in the repo. Each one cost a real defect or a suite that was green and false, and each was found by probing rather than by reading — but as private notes they bind one contributor's sessions rather than everyone working here. CLAUDE.md already carries the closest neighbours (the non-degenerate-oracle rule and the three "a green test is not evidence that it can fail" shapes, #1211), so these belong beside them.

Two of the seven turned out to be **already covered** by CLAUDE.md and are not in this PR. The five below are the genuine gap.

## What changed

`CLAUDE.md` only — no Rust, no `docs/`.

**Two new bullets on the existing "a green test is not evidence that it can fail" list:**

| | |
|---|---|
| Two redundant gates cover for each other | #1229 filtered on `coordinate_kinds(i) == OmegaDiagonal` *and* on `decls[i] == None → continue`. Both correct, both rejecting the same inputs — deleting either left all 15 tests green, so the Σ scope pin could not fail. Reading the code missed it twice. |
| `is_finite()` cannot see a sentinel | `tte_nll_from_curves` maps a `NaN` `H` to `1e20`. On #1223 the objective came back `2e20` — finite — under the exact mutation the assertion existed to catch. Bound the magnitude instead, and state both distances (to the sentinel, to the real objective). |

**Two limits on "I mutation-tested it", appended to the paragraph that asks for mutation:**

- **Mutating your own fix is not enough.** On #1171 / PR #1174, 16 tests across three tiers each died correctly when the fix was toggled — and a reviewer's *different* minimal edit passed the whole suite at identical margins, because every fixture used `zero_order`, the one kind where route onset coincides exactly with the window start. Ask what the smallest edit that removes the fix is, and which test dies.
- **Mutate each side of a twin separately.** On #1223 deleting the share fill reddened the parity test; deleting the dense fill left it green, because the fixture never reached the dense fill. A twin whose second leg runs on an unexercised path is an assertion against a constant.

**Three rules next to the non-degenerate-oracle rule they complement:**

- **The oracle must outrank the disagreement.** `verify_mr_twin` derived its tolerance from `reltol` alone, but Dormand–Prince controls `abstol + reltol·|y|`, so a state decayed to the order of `abstol` has no significant digits left. Measured on #1124/#1130: the twin said 29.484, the closed form 24.574829, and the exact Bateman superposition is **24.574829** — the twin was 20% high and the check aborted a correct fit in debug. Includes the corollary that swapping oracles is changing the experiment (`ode_predictions_with_states` drops a lagged `zero_order` rate — #1171).
- **Measure the tolerance, never justify it with a story.** PR #1174 set a NONMEM POSTHOC sdtab anchor at `3e-2` on the reasoning that the inner optimizers "stop at slightly different η̂". Realised error: **4.736e-5**, 630× tighter, and η̂ matches NONMEM's `.phi` to ~6 significant figures — the stated mechanism was wrong, not conservative. At `3e-2` a regression delivering 99% of the absorbed mass passes green, and that was the only end-to-end `fit()` → sdtab check against an external reference.
- **When the bug is "two implementations disagree", the fix is one implementation.** #1223's first fix re-wrote the dense engine's pre-start fill inline in the #570 shared solve ~3600 lines away, under a comment calling it "the exact twin of" — restoring by prose the configuration that produced the defect. Extracting `fill_prestart_states` means one mutation of the helper now reddens both engines.

**One sentence on the `Dual2`-vs-FD parity rule:** ask which engine each fixture actually ran on. On #1210 all seven `nonmem_anchor/ss_chz_*` arms reported "1 of 1 subjects use finite-difference inner gradients", so a green `--lib` suite and green CI never touched `sens/ode_provider.rs`'s dual SS equilibration.

## Alternatives considered

**Leave them in per-session memory.** That is the status quo and the reason for this PR: private notes do not reach Ron or Hidde, and they go stale silently because nothing loads them next to the code they describe.

**Add all seven.** Two were already in CLAUDE.md and are deliberately excluded — the three "cannot fail" shapes (#1211 put them there) and the non-degenerate-anchor rule. I found this by grepping for the wrong string first (`"cannot fail"` where the doc says `"can fail"`) and had to re-derive the delta against the actual text.

**A separate `docs/development/testing-rules.qmd` page.** Rejected: these are instructions to whoever is editing the code, and CLAUDE.md is what gets loaded at that moment. A Quarto page would also pull `docs-lint` into scope for prose that is deliberately long-form.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

## Performance

## Tests
- [x] No new tests needed (why: CLAUDE.md is not compiled or linted — `docs-lint` scopes to `docs/**/*.qmd`, and no Rust source changed)

## Docs
- [x] No user-visible change

`CHANGELOG.md` deliberately untouched: contributor instructions are not a user-facing change.

## Checklist
- [x] `tools/preflight.sh` green — ran the `fmt` group; the `check`/`clippy`/`rustdoc`/`docs`/`public-API` groups are unaffected because no Rust and no `docs/**/*.qmd` changed
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, no code touched

### ferx-site / ferx-book
- [x] No new DSL syntax, `[fit_options]` key, example, data file, estimator or fit option — nothing to mirror

### Example execution
- [x] No example execution step needed (docs-only, no user-visible change)

## Reviewer hints

The substance is five rules, each anchored to a measured number from a specific issue — the numbers are the point, so check those rather than the prose. The two you can skim are the new bullets on the existing "can fail" list; they follow that list's established shape exactly.

Worth a real look: whether the oracle-precision rule and the measured-tolerance rule belong as separate paragraphs or read as one idea. They came from the same PR (#1174 / #1124) and both say "a number you did not measure is not evidence", but they fire at different moments — one when choosing what to compare against, one when choosing the bound.

CLAUDE.md grows by 72 lines here, which is not free: it loads into every session on this repo. I think each of these earns it, but that is the trade to push back on if you disagree — the weakest of the five is probably "one implementation, not two", which is a design rule wearing a testing rule's clothes.

## Open questions

Nothing blocking.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -274,6 +274,46 @@ that no fixture could see.
 When an anchor does fail, vary **one input at a time** — the pair that differs by a
 single number is what localises the defect (`nonmem_anchor/tvcov_lag_saltation*`).
 
+**The oracle has to be more accurate than the difference you are about to call a defect.**
+An oracle outside that regime measures itself, not the thing under test — and an ODE twin at
+default tolerances is not automatically an oracle. `verify_against_ode_twin`
+(`pk/modified_release.rs`) compared the MR closed form against `ode_predictions` at
+`verify_tol = (500·reltol).max(1e-4)`, derived from `reltol` alone; but Dormand–Prince
+controls local error as `abstol + reltol·|y|`, so the twin's own
+relative accuracy is `abstol/|y| + reltol` and a state decayed to the order of `abstol` has no
+significant digits left. Measured on #1124/#1130: `central` fell to 1.3e-6 against that spec's
+`abstol = 1e-6`, the twin said 29.484, the closed form said 24.574829 — and the exact Bateman
+superposition is 24.574829, so the *twin* was 20% high and the check aborted a correct fit in
+debug. Gate the comparison on the condition for the oracle to outrank the disagreement, and
+when you cannot tell an oracle artifact from a real defect, **compute the quantity a third
+way, outside both engines** — a 30-line Bateman sum in Python settled that one after two
+wrong hypotheses reasoned from the code. Swapping oracles is changing the experiment, too:
+moving that twin to `ode_predictions_with_states` silently adopted a *wrong* reference, because
+at the time it dropped a lagged `zero_order` input rate that `ode_predictions` applied — a real
+public-API defect the swap surfaced, since fixed (#1171).
+
+**Measure the tolerance; never justify it with a story.** Run the comparison, print the worst
+realised error, put that number in the comment, and pick the bound from it with stated
+headroom. On PR #1174 a NONMEM POSTHOC sdtab anchor was set at `3e-2` with the explanation
+that the two inner optimizers "stop at slightly different η̂"; the realised error was
+**4.736e-5** — 630× tighter — and ferx's η̂ matches NONMEM's `.phi` to ~6 significant figures,
+so the stated mechanism was not conservative but wrong. A bound that loose is not safe, it is
+a test that has quietly stopped testing: at `3e-2` a regression delivering 99% of the absorbed
+mass passes green, and that was the only end-to-end `fit()` → sdtab check against an external
+reference. If the realised error surprises you, chase it — that is information.
+
+**When the bug report is "two implementations disagree", the fix is one implementation.**
+#1223's first fix re-wrote the dense engine's pre-start fill inline in the #570 shared solve
+~3600 lines away, with its own `- 1e-12` and its own `break_times.first()` guard, under a
+comment calling it "the exact twin of" — restoring by prose exactly the configuration that
+produced the defect. Extracting `fill_prestart_states` and calling it from both sites closed
+the drift class in the compiler instead, and the payoff is a test property: **one mutation of
+the helper now reddens both engines**, where the dedicated-engine control previously needed a
+separate dense-side mutation to fail. Before writing a second copy, check whether the callers
+can share a function; if they genuinely cannot, record the specific asymmetry at the call site
+(here: the share's `chz_times` is sorted-unique by contract but `ode_dense_solve_states` is
+`pub` with no such guarantee, so the helper keeps a full scan rather than a `take_while`).
+
 **A green test is not evidence that it can fail.** The rule above is about a fixture that
 cannot expose the defect; these are three ways the *assertion* cannot observe it, all three
 found on #1166 in tests that had been written, run green and believed:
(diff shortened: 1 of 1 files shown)


TeunP:
## Self-review — three factual defects found and fixed (force-pushed `6c638ffd` → `f06840b9`)

I wrote this PR, so I reviewed it the way I would review someone else's: an independent pass that checks every claim against the code rather than against the notes I drafted from. That found three things, all of the same kind — **a claim that was true when I recorded it and is not true of `main` today.**

### 1. Named a function that does not exist

The oracle-precision rule called the MR twin check `verify_mr_twin`. There is no such function: it is **`verify_against_ode_twin`** (`src/pk/modified_release.rs:911`). `verify_mr_twin` survives only inside a stale doc comment at `modified_release.rs:2209`, which is where I picked it up. A reader following the rule would have grepped and found nothing.

Fixed, and the file path is now given alongside it.

The formula in the same sentence *did* check out exactly — `src/pk/modified_release.rs:985` is `let verify_tol = (500.0 * solver.reltol).max(1e-4);`.

### 2. Asserted a present-tense defect that has since been fixed

The corollary about swapping oracles said `ode_predictions_with_states` "drops a lagged `zero_order` input rate that `ode_predictions` applies (#1171)". #1171 is merged — `src/ode/predictions.rs:6188` now reads "Zero-order no longer depends on this — `push_zero_order_break_times` brackets its own `w_start`".

Rewritten to past tense, and it now says what the episode actually teaches: the swap *surfaced* a real public-API defect, since fixed.

### 3. Cited code from an unmerged branch

The "two redundant gates" bullet named `coordinate_kinds(i) == OmegaDiagonal`, `decls[i] == None`, and the test `free_zero_sigma_is_not_flagged`. **None of those exist on `main`** — they live on `worktree-issue-1229-omega-init-rail` (#1246), which is open and which I have posted two blocking findings against, so the identifiers may not survive in that form.

Merging this PR first would have put dangling symbols into the contributor instructions. The bullet now describes the *shape* — a kind filter plus a second derived per-coordinate table, both excluding the same inputs — and keeps the issue number, so it is correct today and stays correct whatever #1246 lands as.

I also dropped the "all 15 tests" count: the #1229 branch finished with 19 tests, and 15 was the number at the moment of the observation.

### Verification I ran afterwards

Extracted every backticked identifier from the diff and required each to resolve in `src/`, `crates/` or `nonmem_anchor/`. **26/26 resolve** — `tte_nll_from_curves`, `fill_prestart_states`, `push_route_lag_break_times`, `push_zero_order_break_times`, `ode_dense_solve_states`, `verify_against_ode_twin`, the `1e20` sentinel in `crate::survival`, and the FD-fallback warning string (`inner_optimizer.rs:334`). `fill_prestart_states` has exactly the two call sites the rule claims (`predictions.rs:3069` and `:6790`).

### What I did NOT re-derive

The measured numbers are quoted from the session records where they were originally measured, not re-run today: 29.484 vs 24.574829 and the 1.3e-6 state decay (#1124/#1130), the 4.736e-5 realised error and 630× factor (PR #1174), and the `2e20` objective (#1223). If any of those matter enough to bind future contributors — and I think they are the whole value of the rules — they are the part worth spot-checking, since I cannot cheaply re-measure them and a wrong number here is worse than no rule.

### Still open for you

Unchanged from the description: CLAUDE.md grows 72 lines and loads into every session on this repo. My own view is that "one implementation, not two" is the weakest of the five — it is a design rule wearing a testing rule's clothes — and is the one to cut if the length bothers you.

TeunP:
## Correction to my self-review: the numbers *are* independently verifiable, and I checked them

Above I wrote that the measured numbers were "quoted from the session records where they were originally measured, not re-run today", and flagged them as the part worth spot-checking. That undersold it. Every one of them is pinned in committed source or a committed test — I went and looked rather than leaving it as a caveat:

| Claim in the new text | Where it is pinned on `main` |
|---|---|
| `verify_tol = (500·reltol).max(1e-4)`, derived from `reltol` alone | `src/pk/modified_release.rs:985` — verbatim |
| `central` decayed to 1.3e-6; twin 29.484 vs closed form 24.574829; exact Bateman 24.574829 | `src/pk/modified_release.rs:995-996`, and a live assertion at `:2268` — `(last - 24.574829).abs() < 1e-5` |
| Anchor set at `3e-2` on a "different η̂" explanation; realised error 4.736e-5 | `tests/lagged_zero_order_nonmem_anchor.rs:301-306` — the doc comment says it outright: "An earlier draft used 3e-2 and blamed 'different η̂'; that was wrong on both counts" |
| "630× tighter" | 3e-2 ÷ 4.736e-5 = 633.4 |
| `1e20` TTE sentinel; `is_finite()` cannot see it | `src/stats/likelihood.rs:4218-4224` — the fix itself, with the same reasoning and the same `1e12` bound |
| `fill_prestart_states` called from both engines | `src/ode/predictions.rs:3069` and `:6790` — exactly two call sites |

The one figure not in the repo is the `2e20` objective #1223 measured *under the mutation*, which by construction only exists while the bug is reintroduced. The committed comment states the fixture's real objective is "order 10" against a `1e20` sentinel, so a Laplace objective built on repelled records landing at ~`2e20` is exactly what that arithmetic predicts.

So the rules are not summaries of my recollection: each restates a conclusion the codebase already records at the site it applies to. What this PR adds is that they now bind at the moment someone is *writing* a test, instead of only being discoverable once you are already reading the file where the mistake was made.

That leaves one open item, and it is a judgment call rather than a fact: 72 lines is not free in a file that loads into every session. My own weakest item is still "one implementation, not two" — a design rule wearing a testing rule's clothes. Happy to drop it if you'd rather keep the section tight.

PR #1256, opened 2026-09-06
feat(tools): modelsearch — structural PK search over absorption, peripherals, transits and lagtime [closes #1181]

### LINKED ISSUE #1181: feat(tools): modelsearch — structural PK search over absorption, peripherals, transits and lagtime [#1175]
Part of #1175 (P2). Depends on #1176, #1177, #1178, #1179.

Structural PK model search — Pharmpy's `modelsearch`.

## Algorithms

- `exhaustive` — every feature combination from the base model.
- `exhaustive_stepwise` — features applied one at a time in every order, inits updated from the previous step.
- `reduced_stepwise` — as above, but only the best model of each feature layer continues.

## Search space

`ABSORPTION`, `PERIPHERALS`, `TRANSITS`, `LAGTIME` from the `.ferxsearch` MFL.

| MFL feature | ferx |
|---|---|
| `ABSORPTION(INST/FO/ZO)` | ✅ `pk *_iv` / `*_oral` / infusion + `RATE` |
| `ABSORPTION(WEIBULL)`, `TRANSITS(n)` | ✅ `weibull()`, `pk *_transit` (continuous n), `igd()` |
| `ABSORPTION(SEQ-ZO-FO)` | ⚠️ no template — reject |
| `PERIPHERALS(0..2)` | ✅ `one_cpt` / `two_cpt` / `three_cpt` |
| `LAGTIME(ON/OFF)` | ✅ `lagtime=` / `alag=` |
| `ELIMINATION(ZO/MM/MIX-FO-MM)` | ⚠️ **only via `[odes]`** — see below |

## The elimination gap

Non-linear elimination has no analytic template. A candidate using it must be emitted as an `[odes]` model, which changes both the numerics path and the runtime (10–100×) relative to its analytic siblings. Two consequences the tool must handle explicitly rather than discover:

1. Ranking an ODE candidate against analytic candidates is still valid (same likelihood), but the **runtime budget is not uniform** — the report must show per-candidate cost.
2. MM models are the canonical local-minimum case (`docs/examples/multistart.qmd`). They need a higher `n_starts` than their analytic siblings, or the search will reject a correct MM model on a stalled fit.

v1 may reject `ELIMINATION` features outright and document the gap; the decision belongs in this issue.

## Also excluded

Pharmpy's incompatible-combination rules (e.g. zero-order absorption with transits) must be encoded, so the search does not spend fits on models that are pharmacometrically meaningless.

## Validation

- [ ] Candidate ranking compared against Pharmpy `modelsearch` on a shared dataset.
- [ ] Each generated candidate fits identically to its hand-written equivalent (the `tests/covariate_model_equivalence.rs` pattern).
- [ ] Degenerate oracle: a single-point space returns the base fit bit-for-bit.

## Deliverables

- [ ] `ferx_tools::modelsearch`, `ferx modelsearch` subcommand
- [ ] `docs/tools/modelsearch.qmd` including the coverage table above
- [ ] `CHANGELOG.md` entry

roninsightrx:
## Why

Closes #1181 (part of the #1175 AMD epic, P2). Structural PK model search — Pharmpy's `modelsearch` — over absorption, peripheral compartments, transit compartments and lag time. Everything it sits on already exists (`ferx-core::edit` #1176, BIC/strictness #1177, the candidate runner #1178, the `.ferxsearch` file + MFL #1179, covsearch #1180); what was missing was the tool that turns a structural feature into a `pk` template swap and walks Pharmpy's three algorithms.

## What changed

**`ferx_tools::modelsearch`** (`crates/ferx-tools/src/modelsearch/`)
- `structure.rs` — a `Structure` (absorption × peripherals × transits × lagtime), the MFL `FeatureKey`s, Pharmpy's `_is_allowed` move rules (verbatim, incl. its peripheral ordering quirk and the incompatible-pair table), the least-number-of-transformations base derivation, and `structural_spec()`: structure → `pk` template + bindings (by slot, so `v1=V` ↔ `v=V`, `q=Q` ↔ `q2=Q`) + `NewParameter`s with Pharmpy's inits (`Q = CL`, `V2 = 0.05·Vc`, lag/MTT = `t₁/2`, `KA = 1/(2t₁)`, `N = 2`), scaled from the parent's **seeded** estimates as Pharmpy's `update_inits → add_peripheral` order does.
- `mod.rs` — `exhaustive`, `exhaustive_stepwise`, `reduced_stepwise` over the shared `StepFitter` seam; ranking on `[rank] type` with optional `cutoff` over the base; `iiv_strategy` (`absorption_delay` default / `add_diagonal` / `no_add`; `fullblock` refused → iivsearch); every fitted model's text kept on the result (`models`).
- `report.rs` — `models.csv` (structure, criterion, Δ, rank, convergence, strictness reason, **seconds**, `continued`, `reused`), `final.ferx`, `models/<id>.ferx`, a summary table.
- `search/fitter.rs` — the `StepFitter`/`RunnerFitter` seam lifted out of covsearch so both tools share one implementation.

**`ferx modelsearch`** subcommand (`crates/ferx-cli/src/modelsearch_cmd.rs`), thin.

**`ferx-core::edit`** — two small changes, both needed by the search:
- `NewParameter::fixed` → `theta NAME(…) FIX`, so `TRANSITS(3)` is a named fixed count (a literal `n=3` binding makes the ODE twin decline).
- `SeedInits` floors a collapsed variance at `1e-5`: a parent whose η fell to the optimizer's rail (6e-6, measured on the warfarin lag) used to hand every child a start the engine refuses ("All multi-start fits failed").

**Decision on `ELIMINATION`**: refused by name (already a coverage error). Offering `[odes]` candidates half-handled — 10–100× runtime, MM's multistart needs — is worse than not offering them; the report already carries per-candidate `seconds`, which is the first thing that follow-up needs. Recorded on the docs page.

## Alternatives considered

- Replicating Pharmpy's `reduced_stepwise` exactly: Pharmpy 2.0.0 collapses only when a layer holds ≥2 duplicate groups (`len(groups) > 1`), a bug; ferx collapses every group and the anchor asserts the divergence.
- Leaving unnamed MFL categories to Pharmpy's default-fill (an oral base in a space with no `ABSORPTION` statement becomes a bolus base in Pharmpy): ferx keeps the base's value. Asserted divergence.
- A literal `n=3` binding for `TRANSITS(3)` (no core change): works but the ODE twin declines it with a warning on every fit.

## Cross-repo dependency

ferx-r does not construct `NewParameter` (checked); the new field is additive for it. A `ferx_modelsearch()` R surface is #1184's.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — regenerated: `+pub ferx_core::edit::NewParameter::fixed: bool`. Needed by `ferx-tools::modelsearch` for a fixed transit count; a struct-literal constructor of `NewParameter` gains one field (no in-repo or ferx-r caller besides the tests updated here).
- [ ] None

## Numerical validation

**Degenerate oracle** (`crates/ferx-tools/tests/modelsearch_end_to_end.rs`): a single-point space returns the base fit bit-for-bit, and `final.ferx` re-evaluates to the same OFV bit-for-bit.

**Candidate ≡ hand-written** (same file, the `covariate_model_equivalence` pattern): the search's two-compartment lagged candidate and its hand-typed twin agree on `predict()` bit-for-bit and on the evaluated OFV bit-for-bit. Every template reachable from a one-compartment oral base (2nd/3rd compartment, fixed and estimated transit chain, lag, and the allowed pairs) compiles and evaluates.

**Pharmpy enumeration anchor** (`tools/pharmpy-modelsearch-anchor/`, replayed by `src/modelsearch/pharmpy_anchor.rs`): 9 (base, space, algorithm) cases dumped from Pharmpy 2.0.0's own workflow without fitting; candidate moves agree in Pharmpy's order, `(feature set, parent set)` edges agree as multisets through the reduced-stepwise collectors, exhaustive combos agree in order so `run{n}` = `modelsearch_run{n}`. Four divergences named and asserted to actually differ (default-filled base, lag-on-bolus base, single-group collapse, exhaustive pair table). Transit cases excluded on purpose — Pharmpy's base derivation raises on `TRANSITS(0, NODEPOT)` and its feature reader counts compartments, so it re-applies the base's own chain; README records both.

**NONMEM 7.5.1 ranking anchor** (`crates/ferx-tools/tests/nonmem/modelsearch_anchor/`, `tests/modelsearch_nonmem_anchor.rs`, slow-gated): warfarin, FOCEI, `PERIPHERALS(0..1); LAGTIME([OFF,ON])` exhaustive; candidates written as the search writes them, seeded from `base.ext`.

| model | NONMEM own run | NONMEM at ferx's estimates | ferx | mixed BIC | rank |
|---|---:|---:|---:|---:|---|
| base | −286.0042 (successful) | — | −286.0042 | −267.488 | 1 |
| LAGTIME(ON) | −287.0526 (terminated) | −287.6292 | −287.6292 | −264.508 | 2 |
| PERIPHERALS(1) | −289.0727 (terminated) | −289.0986 | −289.0986 | −261.182 | 3 |
| both | −287.4830 (terminated) | −287.6292 | −287.6292 | −255.107 | 4 |

NONMEM's own minimiser terminated on all three candidates (lag η collapses); evaluated at ferx's estimates it agrees to 1.2e-9 / 3.7e-9 / 5.4e-5, and re-minimised *from* ferx's estimates it reports `MINIMIZATION SUCCESSFUL` at −287.62917388 = ferx's optimum. Same BIC ranking under either NONMEM reading, same final model (base). Tolerance 1e-3 measured (worst 5.4e-5, 20× headroom; smallest BIC gap 2.98).

- [x] Compared against: NONMEM 7.5.1, Pharmpy 2.0.0

## Performance
- [x] No significant impact expected (a tool; the engine is untouched apart from the edit layer)

## Tests
- [x] Unit tests: `structure_tests.rs` (rules, templates, inits, IIV strategies, name collisions), `mod_tests.rs` (three algorithms with a scripted fitter, collapse, cutoff, gate exclusion, base outside the space, unbuildable moves, cancellation, report), `pharmpy_anchor.rs`; core `edit` tests for `fixed` and the seed floor; CLI tests.
- [x] Tier-2 `modelsearch_end_to_end.rs` (5), Tier-3 `modelsearch_nonmem_anchor.rs`.
- Mutation checks done by hand: the FIX suffix (test asserts the unfixed line is absent), the pair table (both orders), the seeded-vs-file inits (test pins `V2 = 0.05·seeded V`, which the file's 0.5 would fail), each divergence asserted as *different*.

## Example
- [x] `examples/warfarin_modelsearch.ferxsearch`; `ferx modelsearch examples/warfarin_modelsearch.ferxsearch` runs in seconds and selects the base (no candidate earns its parameters on 10 subjects).

## Docs
- [x] `docs/tools/modelsearch.qmd` (coverage table, algorithms, inits table, the elimination-gap decision, Pharmpy + NONMEM comparisons, deviations), linked in `_quarto.yml`; `docs/cli.qmd`.
- [x] `CHANGELOG.md`.

## Checklist
- [x] No `Dual2` path touched
- [ ] `Cargo.toml` version bumped — additive field, not bumped

## Docs & examples
- [x] New example `.ferxsearch` in `examples/` (a search config, not a `.ferx` model — no ferx-site/book entry yet; #1184 will document the tool family together)
- [x] Rebuilt ferx-core, example run via the release CLI
- [ ] ferx-r rebuild — not needed for this PR

## Reviewer hints

- `structure.rs::allowed` is Pharmpy's `_is_allowed` transcribed; the anchor JSON is the reference if a rule looks odd (e.g. `PERIPHERALS(2)` then `(1)` being a path — that is Pharmpy's).
- `search()` in `mod.rs` is the orchestration; `derive()` is where a candidate is built (seed → structural spec → edit).
- The two core changes are small and isolated in `src/edit/apply.rs`.
- Skim `report.rs`.

## Open questions

- The shipped example dropped `TRANSITS` from its space because a two-compartment transit candidate with the covariance step took ~5 min per fit on warfarin (the `seconds` column showed it). Worth a look at `two_cpt_transit` fit cost separately; not this PR.

https://claude.ai/code/session_0193LzWXUgwiRMtFQBUyt9PS


roninsightrx (commented):
Requesting changes for six issues. The first four can make the search fit or report a different model than the row says it did.

1. **[P1] Preserve optional `f=` bindings across structural swaps** (`crates/ferx-tools/src/modelsearch/structure.rs:290`). `Structure::template()` rebuilds the role list from the required disposition/absorption roles plus lag time, but never carries the parent's optional `f`/bioavailability binding. `SetStructural` then replaces the whole `pk` line and prunes the now-unreferenced `F` parameter and its theta/eta. For example, searching only `PERIPHERALS(0..1)` from `pk one_cpt_oral(..., f=F)` generates a two-compartment candidate without `f=F`, effectively changing bioavailability to its default while claiming only the compartment count changed. Please preserve optional parent bindings that remain valid and add an equivalence regression with estimated F.

2. **[P1] Existing transit models cannot be represented or moved correctly** (`structure.rs:177`, `structure.rs:693`, and `structure.rs:425`). `from_template()` classifies every `*_transit` base as `TRANSITS(N)`, even when `n` is a literal or is backed by a fixed theta. Then, when targeting another transit value, `structural_spec()` sees the existing `n` binding and reuses it without changing its initializer or `FIX` state. A row labelled `TRANSITS(1)` can therefore fit the original fixed count 3 (or still estimate N), and different transit candidates can canonicalize to the same unchanged model. Also, `TRANSITS(0)` is unconditionally suppressed as a move, so a transit base cannot search the no-chain value. Please derive the actual transit coordinate from the bound declaration and make fixed/estimated/count-zero transitions edit the model text; add tests starting from both fixed and estimated transit bases.

3. **[P1] Candidate construction consults stale input declarations after pruning** (`structure.rs:699`). `Defaults.parameters` is captured from the original input model and reused for every descendant. If the input is narrowed, `SetStructural` prunes parameters that the new parent no longer declares, but this check still treats those names as present and omits their `NewParameter`. A concrete case is a three-compartment input with `PERIPHERALS(0..1)`: deriving the zero-peripheral base removes `V2`; generating the one-peripheral child then binds `v2=V2` while refusing to recreate it, so `SetStructural` aborts the whole search as an undeclared binding. This check needs to use the current parent model's declarations, with a regression covering narrow-then-widen.

4. **[P1] A missing `FitResult` silently disables parent seeding and can produce an unseeded final artifact** (`crates/ferx-tools/src/modelsearch/mod.rs:489` and `mod.rs:515`). `CandidateResult::fit` is legitimately `None` for a canonical duplicate and for a resumed row whose `fits/<hash>.json` is missing/corrupt, while its criterion/verdict can remain eligible. `Node::from_model` accepts that state, `derive()` silently skips `SeedInits`, and the search continues from the file's old initials despite the documented previous-step seeding algorithm. If such a reused row wins, `write_report` also emits an unseeded `final.ferx` and the CLI omits `final-fit.yaml` while exiting successfully. Modelsearch should resolve a duplicate to its representative fit and refit or fail clearly when a resumable fit object is unavailable; please cover both paths.

5. **[P1] The PR does not satisfy #1181's absorption search space** (`docs/tools/modelsearch.qmd:70`). The issue explicitly marks `ABSORPTION(ZO)` and `ABSORPTION(WEIBULL)` as supported deliverables and permits rejection only for `SEQ-ZO-FO` (plus the separately discussed elimination gap). This implementation rejects ZO and WEIBULL at config load and closes #1181 without documenting a scoped follow-up or changing the issue acceptance criteria. Either implement those candidates or explicitly split/re-scope the issue before closing it.

6. **[P2] `NewParameter::fixed` is a source-breaking public API change** (`src/edit/spec.rs:94`). External callers must construct `NewParameter` with a struct literal because it has no constructor/builder/default; adding a required public field makes all existing literals fail to compile. The PR description calls this additive and leaves the workspace at 0.3.1, but under semver this needs to be treated as breaking (for a 0.x crate, normally a minor-version bump), with dependent version/lock updates as appropriate.

Validation performed: `cargo test -p ferx-tools modelsearch --lib` passed 44/44; `cargo check --workspace --tests` passed; `tools/update-public-api.sh --check` passed. I also started the full `cargo test --lib`; no failures appeared before I stopped it because it exercises thousands of unrelated tests.

roninsightrx:
Thanks — all six addressed in the follow-up commit. Point by point:

**1. `f=` lost across swaps — fixed.** `structural_spec` now carries the parent's `f` binding on every swap (every template reads it), so `F` and its θ/η survive a compartment change. Pinned two ways: `a_parents_bioavailability_binding_is_carried_across_a_swap` (structure level, oral→2-cpt and oral→IV) and `an_estimated_bioavailability_is_carried_across_the_swap` in `modelsearch_end_to_end.rs`: a base with an estimated `TVF`, the two-compartment candidate keeps `f=F` and `TVF`, has two θ more than the base (not one), and is bit-identical on `predict()` and the evaluated OFV to a hand-written twin; the test also asserts its OFV differs from the base's, which a swap that reset F to 1 would not.

**2. Transit bases — fixed.** `Structure::from_model(template, text)` reads the transit coordinate off what the `n` binding points at: a literal or a parameter behind a `FIX`ed θ is `TRANSITS(k)`, a free θ is `TRANSITS(N)`, a non-integral fixed count is refused by name. A move to a *different* transit coordinate (another count, fixed ↔ estimated) binds `n` to a fresh parameter (`NTR2` when the parent's is `NTR`) with the right `FIX`/free declaration; the parent's old one is left unreferenced and pruned with its θ. `TRANSITS(0)` is now a move *off* a chain (`allowed`/`combination_allowed` take the current structure; on a first-order model it stays refused, Pharmpy's rule). Tests: `a_transit_bases_count_is_read_off_the_declaration…`, `a_different_transit_coordinate_rebinds_n_to_a_fresh_parameter`, `transits_zero_is_a_move_only_off_a_chain`, `a_fixed_transit_base_moves_to_zero_and_to_another_count_by_rebinding_n` (asserts three distinct canonical hashes), `an_estimated_transit_base_moves_to_a_fixed_count_and_back`, and end-to-end `a_fixed_transit_base_compiles_every_transit_move` (four distinct evaluations, parameter counts as expected).

**3. Stale declarations — fixed.** `derive()` now builds `Defaults::of_text(&seeded_parent_text)`: parameters, θ names/inits and η names come from the parent's own text after its edits, never the input's. Regression `a_narrowed_base_can_be_widened_again_because_declarations_come_from_the_parent`: a three-compartment input in `PERIPHERALS(0..1)` — the base is narrowed (Q2/V2/Q3/V3 pruned), and the one-peripheral child re-declares `TVQ`/`TVV2`. Before the fix that search aborted exactly as you described.

**4. Missing `FitResult` — fixed.** New `resolve_fit`: a duplicate is seeded from its representative's fit (and carries it out if it wins, so `final.ferx`/`final-fit.yaml` are seeded); a failed candidate (error) extends unseeded with a note in the report; a resumed row whose `fits/<hash>.json` is missing/unreadable is a hard error naming the row and the fix (`refit without resume`) — for the base too. Tests: `a_duplicates_children_are_seeded_from_its_representatives_fit` (child text carries the representative's TVCL; a winning duplicate has `final_fit`), `a_resumed_row_without_its_cached_fit_is_an_error_naming_the_fix`, `a_failed_candidates_children_start_unseeded_and_the_notes_say_so`.

**5. ZO / WEIBULL — split, documented.** Neither is a `pk` template swap: `ZO` on an oral dose is a modeled duration on an IV template with `RATE=-2` *in the dataset*, and `WEIBULL` is the `[odes]` input function — the same `[odes]` candidate family as the elimination gap, with the same needs (non-uniform runtime budget, per-candidate start budget). Opened #1257 for that family with those requirements, pointed the docs coverage table at it, and left a re-scope comment on #1181. If you'd rather #1181 stay open until #1257 lands, say so and I'll change the PR to `refs #1181`.

**6. `NewParameter::fixed` — handled the way #996 handled `MixtureSpec`.** `NewParameter` is now `#[non_exhaustive]` with `NewParameter::new(name, theta, init, lower, upper)` + `.with_iiv(eta, var)` + `.fixed()` builders (fields stay public to read), so the *next* field is non-breaking; the CHANGELOG entry is marked **Breaking for struct-literal construction**. On the version: this repo bumps at release time (`chore(release): 0.3.1`), not per PR, so I have not touched `Cargo.toml`; the changelog marking is what makes the next release a minor bump. ferx-r does not construct `NewParameter` (checked), so no lock update is needed on that side.

Validation re-run after the changes: `cargo test -p ferx-tools -p ferx-cli` (all green, 54 modelsearch unit tests incl. the Pharmpy anchor, 7 end-to-end), the slow NONMEM anchor, `tools/preflight.sh`, `tools/update-public-api.sh --check` (baseline regenerated: the three builders).

roninsightrx:

Seven findings, ordered most-severe first.

---

**1. `crates/ferx-tools/src/modelsearch/mod.rs:523` — `resolve_fit` checks `duplicate_of` before `error`, so a duplicate of a *failed* candidate aborts the whole search.**

`Runner` builds a duplicate result by cloning the representative's `verdict`, `criterion` **and `error`** (`search/runner.rs:395-411`) while leaving `fit: None`. If the representative failed to compile or fit, the duplicate carries `duplicate_of = Some(rep)` *and* `error = Some(..)`. `resolve_fit` takes the duplicate branch first, finds `rep.fit == None`, and returns `Err("… a duplicate of X, whose fit is not available to seed from")`, which propagates through `Node::from_model`/`?` and kills `search()` entirely instead of recording a failed row. Reachable whenever two candidates canonicalise identically — e.g. two `exhaustive_stepwise` layer-2 orders whose parents both failed (no `SeedInits`, so identical text) and whose representative then also fails. The scripted fitter in `mod_tests.rs` never sets `duplicate_of` on an erroring id, so this is untested.

*Fix:* test `result.error.is_some()` first, or fall back to `Ok(None)` when the representative has no fit *and* an error.

---

**2. `crates/ferx-tools/src/modelsearch/mod.rs:416` — `Space::build` blames a derivation that did not happen when `onto` is empty.**

The `unbuildable()` guard fires unconditionally, but the message is written for the outside-the-space case. For an input that *is* on the space yet has no template — e.g. a legal `pk two_cpt_iv(cl=CL, v1=V, q=Q, v2=V2, lagtime=ALAG)` (bolus + lag, `unbuildable()` → "a lag time on a bolus dose is not a structural candidate") — the user gets "the input model lies outside the space, and the base Pharmpy's least number of transformations would derive from it **()** has no template", with an empty transformation list and advice ("list the input's own value in each category") that cannot fix it.

*Fix:* split the `onto.is_empty()` case into its own message.

---

**3. `crates/ferx-tools/src/modelsearch/structure.rs:160` — the "`ode_template` lines are refused" docstring is not implemented.**

`from_model` only inspects `template.name`, and `PkTemplate::parse_line` accepts both `pk` and `ode_template` (`search/resolve.rs:60`), so an `ode_template two_cpt_oral(...)` base passes `Space::from_config` untouched. The failure comes later, from `set_structural`'s `PK_RE` guard (`src/edit/apply.rs:93`, "SetStructural needs a `pk NAME(...)` line") — i.e. only after the input/base model has already been fitted, the expensive step the up-front check exists to avoid.

*Fix:* refuse a non-`pk` template line in `from_model`.

---

**4. `src/edit/spec.rs:363` — the new `num()` rounding silently turns a large finite value into `inf`.**

`format!("{v:.14e}").parse::<f64>()` rounds *up* past `f64::MAX`, and Rust's parser returns `Ok(f64::INFINITY)` on overflow rather than `Err`, so `unwrap_or(v)` never fires. Verified: `num(f64::MAX)` → `"inf"`, which passes the `s.contains(['.','e','E','n','i'])` check and is written into the `.ferx` file as a bound of `inf`.

*Fix:* `.ok().filter(|r| r.is_finite()).unwrap_or(v)`.

---

**5. `src/edit/apply.rs:691` — the `MIN_SEED_VARIANCE` floor makes `SeedInits` stop reproducing the fit, including outside modelsearch.**

`SeedInits` is documented (and used by `report.rs::write_report`) as writing "its own final estimates … so it reads and refits as the search left it", but a parent whose ω converged to 6e-6 now gets `1e-5` written. `final.ferx` and `final-fit.yaml` then disagree, and the Tier-2 "final.ferx re-evaluates to the same OFV bit-for-bit" property silently stops holding for any collapsed η. covsearch and any ferx-r caller of `SeedInits` inherit the change.

*Fix:* apply the floor in modelsearch's `derive()` — where the "child cannot start" problem actually is — rather than in the shared edit.

---

**6. `crates/ferx-tools/src/modelsearch/mod.rs:861` — the `input` model competes for selection although it lies outside the declared space.**

When `derive_base` is true the input row is `eligible()` and enters `order`, so with no `[rank] cutoff` the search can select — and write to `final.ferx` — a structure the user's MFL explicitly excluded (that exclusion is the only reason a base had to be derived). Pharmpy ranks base + candidates. The module header documents the choice, but it makes `[space]` non-binding on the result.

*Fix:* exclude the input from `winner`, or say so in the summary line.

---

**7. `crates/ferx-cli/src/modelsearch_cmd.rs:48` — `flag`/`value` duplicate `covsearch_cmd`'s private copies verbatim.**

Two implementations of the same flag parser, against the repo's one-implementation rule; a fix to one (e.g. `--directory=DIR` support, or the `-` value check) lands in only one subcommand.

*Fix:* make `covsearch_cmd::value` `pub(crate)` and import it, as `scan_args`/`parse_threads` already are.

roninsightrx:
All seven findings fixed in a0d9ce8a.

| # | Finding | Fix |
|---|---|---|
| 1 | duplicate-of-a-failure kills the search | `resolve_fit` tests `error` before `duplicate_of`; the scripted fitter now builds the duplicate the way `Runner` does (rep's error cloned in), and the reordering mutation reddens `a_duplicate_of_a_failed_candidate_is_a_failed_row_not_a_dead_search` |
| 2 | `Space::build` blames a derivation that did not happen | split `onto.is_empty()` into its own message ("lies on the space, but its own structure has no template"); test asserts no empty `()` and no "outside the space" |
| 3 | `ode_template` docstring not implemented | `PkTemplate` carries the `keyword` it parsed; `Structure::from_model` refuses a non-`pk` line up front, before the base is fitted |
| 4 | `num()` overflows a finite value to `inf` | `.ok().filter(\|r\| r.is_finite()).unwrap_or(v)`; test pins `num(f64::MAX)` finite |
| 5 | the floor made `SeedInits` stop reproducing the fit | floor moved out of the shared edit into `ferx_tools::search::seed::seed_from`, used by **both** covsearch and modelsearch when deriving a child. `SeedInits` is faithful again, so `final.ferx` and `final-fit.yaml` agree; 3 unit tests (floored / untouched / diagonal-only) |
| 6 | the input competes for selection | ranked but never selected — `the_input_is_ranked_but_never_selected` pins rank 1 + `selected == false`, and the mutation reddens. Documented in the module header, `docs/tools/modelsearch.qmd` and its "Deviations from Pharmpy" list |
| 7 | duplicated `flag`/`value` | `covsearch_cmd::{flag, value}` are `pub(crate)` and imported, like `scan_args`/`parse_threads` |

Verification: `tools/preflight.sh` green (public API unchanged — the `PkTemplate` field is `ferx-tools`, and #5 *removed* a ferx-core const rather than adding one). `cargo test -p ferx-tools` 408 lib + all integration green, `cargo test --lib edit::` green, `modelsearch_end_to_end` (7) and `covsearch_end_to_end` (3) green, and the slow-gated `modelsearch_nonmem_anchor` re-run explicitly with `--features slow-tests` — still ranks the warfarin candidates as NONMEM does.

CHANGELOG updated for #5 and #6 (both are user-visible).

Issue #1262, opened 2026-09-06
feat(dosing): NONMEM constant-infusion steady state (`SS=1, II=0, RATE>0`) is unimplemented — warned as a data error


TeunP:
Found while measuring batch T (#1258). **Gap, not a silent wrong answer** — ferx already warns.

NONMEM's `SS=1` with `II=0` and `RATE>0` is the **constant-infusion steady state**: an infusion
that has been running forever, equilibrated at `RATE/CL`. It is a documented NONMEM record
form, distinct from the periodic `SS=1, II>0` case.

Every steady-state path in ferx is gated on `d.ss && d.ii > 0.0` — `Subject::has_periodic_ss_dose`
(`types.rs:1009`) and the same predicate spelled inline in `pk/mod.rs`, `pk/one_compartment.rs`,
`pk/two_compartment.rs`, `pk/three_compartment.rs`, `pk/event_driven.rs`, `sens/*`,
`io/output.rs`, `dosing.rs`. So the constant-infusion form never reaches an SS branch.

It is **not** silent: `api/validation.rs:3762` raises `W_STEADY_STATE_II`.

### Measured

```
ID,TIME,DV,EVID,AMT,CMT,MDV,II,SS,RATE
1,0,.,1,0,1,1,0,1,10
1,2,4.0,0,0,1,0,0,0,0
...
```

on a 1-cpt `[odes]` model gives

```
* 1 subject(s) have SS=1 doses with missing or non-positive II. SS predictions require
  II > 0 — these doses are treated as non-SS (no steady-state pre-equilibration).
  Set II in the dataset or remove the SS flag.
OFV:  31249999999917.1094
```

The warning is accurate about what ferx did, but it reads as "your dataset is malformed" when
the record is in fact a legal NONMEM construct ferx does not implement. A user translating a
working NONMEM control stream gets told to fix their data.

### Suggested scope

1. Implement it: for a constant infusion, the SS state is the solution of `f(u) + rate = 0` —
   for the linear case `A_ss = RATE/k` per compartment, and for a general `[odes]` RHS the same
   fixed-point / capped-train machinery the periodic case already uses, with the infusion never
   switching off.
2. Until then, change the warning text for the `RATE > 0, AMT == 0, II == 0` sub-case to say
   the construct is recognised and unimplemented, rather than implying a data error.

NONMEM-anchorable: a constant-infusion SS record is an ordinary `$DES` / `ADVAN` run.

Issue #1261, opened 2026-09-06
fix(survival): a hazard that reads `TAD` / `TAFD` gets NaN — the bare parameter slice reaches the RHS on the hazard readout


TeunP:
Found while measuring batch T (#1258). Same root cause as #1131, different engine.

The compiled `[odes]` RHS reads `TAFD` and `TAD` out of `params[MAX_PK_PARAMS]` and
`params[MAX_PK_PARAMS + 1]` (`parser/model_parser.rs`). The survival hazard readout hands it a
bare `PkParams::values` — exactly `MAX_PK_PARAMS` long — so `.get()` on either slot returns
`None` and the RHS injects `NaN`.

Sites: `stats/likelihood.rs:360`, `stats/likelihood.rs:4131`, `survival/mod.rs:738`.

### Measured

1-cpt IV, one subject, doses at t=0 and t=12, an event at t=30, `maxiter = 0`:

```
[event_model]
  cmt    = 3
  hazard = H0 * (1.0 + KT * TAD)
```

| hazard | OFV |
|---|---|
| `H0 * (1.0 + KT * TAD)` | **200000000000000032768.0000** (the diverged-subject sentinel, `2e20`) |
| `H0` (control) | 41164.7786 |

So a joint PK-TTE model whose hazard reads `TAD` — the natural way to write a
dose-driven hazard — silently returns the repelled-subject sentinel instead of a likelihood.

### Note for whoever fixes it

`is_finite()` is the wrong assertion for the regression test: `tte_nll_from_curves` maps a
`NaN` `H` to `1e20`, which is finite. Bound the magnitude against both the sentinel and the
real objective.

hiddevandebeek:
Picking this up together with #1266 — [note on the tracker](https://github.com/FeRx-NLME/ferx-core/issues/1217#issuecomment-5617327024). #1266 is the one-line detector for the class this bug belongs to, it is now unblocked (#1264 closed 2026-09-06), and this issue is its **one remaining open trigger** — so the two land in one PR rather than in sequence.

**Sites re-verified at `7168dad8`:** `survival/mod.rs` still calls `(ode.rhs)(&states[i], &pk.values, …)` — the bare `PkParams::values`. But **the line numbers in the issue have drifted**: `stats/likelihood.rs:4131` now lands inside a test model string, so I will re-locate all three by shape rather than by line, and say in the PR which three I actually changed.

[...]

Issue #1260, opened 2026-09-06
fix(ode): the EKF ignores SS records, so `p_obs` is wrong for a steady-state subject on a nonlinear SDE model (120.5 OFV measured)


TeunP:
Found while measuring batch T (#1258).

`src/ode/ekf.rs` contains **no occurrence of `.ss`** — `solve_ekf` builds its break times from
`dose.time` (and `dose.time + duration` for a real infusion) and applies each dose as a plain
event. There is no steady-state pre-equilibration, and nothing rejects or warns about an
`SS=1` record on a `[diffusion]` model.

### Why it is invisible in `IPRED`

`stats/likelihood.rs:892` takes `let (_, p_obs) = ode_predictions_ekf_with_diffusion(...)` —
the EKF's **mean is discarded**. `IPRED` for an SDE model comes from the ordinary
`ode_predictions` path, which equilibrates correctly. Only `p_obs` comes from the EKF, and
`p_obs` is what carries the defect.

### Why a linear model cannot show it

The Riccati propagation is `dP/dt = F·P + P·Fᵀ + Q` with `F = ∂f/∂u`. For a linear RHS `F` is
state-independent, so the wrong internal mean does not reach `P` at all. Measured on 1-cpt IV
(`CL=1, V=20, II=12`, diffusion 5.0, 41-dose train vs one `SS=1` record): identical to 1e-4.

### Measured on a nonlinear RHS

Michaelis-Menten elimination, `d/dt(central) = -VMAX*central/(KM+central)`, `VMAX=20`,
`KM=50`, `PROP_ERR ~ 1e-4`, `diffusion central ~ 5.0`; observations at t = 482/485/488/491.
Same subject either way — one `SS=1, II=12` record at t=480, or 41 explicit doses at
t = 0,12,…,480.

| path | `SS=1` record | explicit train | Δ OFV |
|---|---|---|---|
| ODE (control) | 31844.3622 | 31844.3622 | **0** |
| SDE / EKF | **610.4128** | **489.8847** | **120.53** |

The ODE row is the certifying control: ferx's SS equilibration reproduces the explicit train
exactly on this nonlinear model, so the 120.53 is the EKF path alone. No warning is raised.

### Options

1. Equilibrate the EKF mean the way the ODE predictors do, and seed `P` at the stationary
   Riccati solution rather than at zero (a subject dosed "forever" has not been observed
   since t=0 either — starting `P` at 0 at the first break understates the uncertainty).
[...]

PR #1263, opened 2026-09-06
fix(ode): carry the model-time anchors into the EKF walk [closes #1131, closes #1259]

### LINKED ISSUE #1131: bug(ode): the EKF/SDE path returns p_obs = NaN and ipred = 0 for any TAD-reading RHS
## Summary

On an SDE (`[diffusion]`) model whose `[odes]` RHS reads `TAD`, the EKF path returns an
observation variance of `NaN` and an `ipred` of **all zeros**, while the ordinary value path
returns a correct trajectory for the same subject. `p_obs` feeds the likelihood, so this is not
confined to a diagnostic.

Found while verifying whether #1110's NaN survived #1130. It is a **separate defect** — see the
matrix: the lagtime #1110 requires is irrelevant here.

## Measured

Doses at t = 0 and 12, observations at 2 / 4 / 8 / 24, `theta = [1.0, 20.0, 0.3, 0.5]`,
`eta = [0.0]`:

| model | value path | EKF `ipred` | EKF `p_obs` |
|---|---|---|---|
| `TAD` + lagtime | finite | **all 0.0** | **NaN** |
| no `TAD`, lagtime | finite | correct | finite |
| **`TAD`, no lagtime** | finite | **all 0.0** | **NaN** |
| plain | finite | correct | finite |

For `TAD` without a lagtime the value path gives `[87.810, 72.615, 41.478, 22.111]` while the
EKF returns `[0.0, 0.0, 0.0, 0.0]`. The two no-`TAD` rows are correct, so this is not API misuse
in the probe.

## Reproducer

```
[parameters]
  theta TVCL(1.0, 0.001, 100.0)
  theta TVV(20.0, 0.001, 500.0)
  theta THETA_TAD(0.3, 0.0, 10.0)
  omega ETA_CL ~ 0.09
  sigma PROP ~ 0.1 (sd)
[individual_parameters]
  CL   = TVCL * exp(ETA_CL)
  V    = TVV
  KTAD = THETA_TAD
[structural_model]
[odes]
  d/dt(central) = -(CL/V) * central * (1.0 + KTAD * TAD)
[diffusion]
  central ~ 0.01
[error_model]
  DV ~ proportional(PROP)
```

(Form C is rejected on SDE models, hence `obs_cmt=`. Adding `LAGTIME = ...` reproduces
identically — the lagtime is not part of the trigger.)

## Path

`stats/likelihood.rs:849` → `ode_predictions_ekf_with_diffusion` (`ode/predictions.rs:4764`),
which calls the dense `ode_predictions` at `:4785`.

## Root cause not yet identified

Two candidates ruled out:

- **Not #1110's `tad_anchor` lag defect.** That needs a lagtime; this reproduces without one.
- **Not the dense predictor mishandling `TAD` as such.** Called directly with populated
  `theta`/`eta`, `ode_predictions` returns correct values for a `TAD`-reading RHS with no lag.

One untested difference: the EKF call site passes **empty `theta`/`eta` slices**
(`ode_predictions(ode, pk_params_flat, &[], &[], subject)`), justified in-comment by "parser
rejects SDE + Form C, so output_fn is always None here and theta/eta would never be consulted."
Worth checking whether a `TAD`-reading RHS breaks that assumption.

## Note on severity

The `ipred = 0` half may matter more than the `NaN`. A `NaN` fails loudly; silent zeros against
a value path returning ~88 for the same subject do not.

## Not caused by #1130

That PR touches neither `src/ode/` nor `src/stats/` — `git diff --name-only` over both is empty.
Pre-existing.

### LINKED ISSUE #1259: fix(estimation): a model with no random effects panics in the covariance step when its objective is non-finite
Found while measuring batch T (#1258).

A model with **no random effects** has a 0×0 OMEGA. When such a fit also ends with a
non-finite objective, the covariance step's non-finite-objective *diagnostic* asks for the
eigenvalues of that empty matrix and `nalgebra` panics:

```
Unable to compute the symmetric tridiagonal decomposition of an empty matrix.
```

Origin: `estimation/covariance.rs` — `compute_covariance`'s `if !base_ofv.is_finite()` branch
calls `extract_eigenvalues(diagnostic_omega(&params_at))`.

**Both conjuncts are required** — measured, not inferred:

| `n_eta` | objective | result |
|---|---|---|
| 1 | non-finite | reported as `2e20`, no panic |
| 0 | finite | fine |
| 0 | non-finite | **panic** |

So it needs a model with zero `omega` declarations *and* a fit that has already gone wrong.
That is exactly the state the diagnostic exists to report on, which is why a panic there is
worse than useless: it replaces the explanation with a crash.

Reproducer: any `[odes]` model with no `omega`, driven to a non-finite objective (e.g. the
#1131 SDE + `TAD` model).

### Fix

Return `None` on an empty matrix so the caller reports the non-finite objective as the reason.
Riding with the #1131 PR, since #1131's own reproducer has `n_eta == 0` and therefore hits
this panic before the #1131 fix can be observed.

TeunP:
## Why

Two defects, one root cause, found while measuring the batch-T plan (#1258).

**#1131** — an `[odes]` right-hand side that reads `TAD` or `TAFD` destroys the objective on a
`[diffusion]` (SDE / extended-Kalman-filter) model. The compiled RHS reads the two model-time
anchors out of `params[MAX_PK_PARAMS]` and `params[MAX_PK_PARAMS + 1]`; `solve_ekf` was handed a
bare `PkParams::values`, which is exactly `MAX_PK_PARAMS` long. Both `.get()` calls returned
`None`, the RHS injected `NaN`, and the `NaN` multiplied into the state and the Jacobian.

The issue reports "`ipred = 0`". Measured at `579d1e96`, that is no longer the symptom — the
sdtab `IPRED` column is *correct* (`87.809543 / 72.614904 / 41.478291 / 22.110924`), because
`stats/likelihood.rs:892` takes `let (_, p_obs) = ...` and discards the EKF mean entirely. What
is broken is the objective: **`OFV = 2e20`** (the diverged-subject sentinel) against
**`94.7893`** for the same model with the `TAD` term dropped. So an assertion on `ipred` is
vacuously green, and every test here asserts `p_obs` or the objective instead.

**#1259** — the #1131 reproducer has no random effects, so its 0×0 OMEGA reaches
`extract_eigenvalues` through the covariance step's non-finite-objective *diagnostic*, and
`nalgebra` panics: `Unable to compute the symmetric tridiagonal decomposition of an empty
matrix`. Both conjuncts are required, measured: `n_eta = 1` + non-finite gives `2e20` and no
panic; `n_eta = 0` + finite is fine; `n_eta = 0` + non-finite panics. It rides here because
without it the #1131 fix cannot be observed on #1131's own model.

## What changed

- `ode/predictions.rs` — `tad_anchor`'s dose-list body split out as `tad_anchor_for(&[DoseEvent], &[f64], f64)`,
  and `seed_ext_params` made `pub(crate)`. The EKF holds no `Subject`, so without this it would
  have grown a **seventh** spelling of the TAD rule. The six already in the tree: `tad_anchor`
  itself, the inline duplicate at `predictions.rs:5947`, `api::output_columns::tad_at_time`,
  the lag-ignoring sdtab fallback at `io/output.rs:1355`, and the dual walk's segment-start and
  saltation anchors in `sens/ode_provider.rs`. Three of them disagree on
  `(t_dose = 120, ALAG = 3, II = 12, t = 121)`, returning `-2.0`, `NaN` and `+1.0`.
- `ode/ekf.rs` — `solve_ekf` builds the same extended array the ODE predictors build, and
  re-anchors the `TAD` slot per dose segment. The anchor write is placed **before** the segment
  integrates: `apply_segment_boundary` compiles either way and writes its own anchor *after* the
  SS call, so a write in the wrong position reads the previous segment's anchor without failing
  to compile. Two integration sites take the extended array (the `solve_ode` call and
  `propagate_covariance`); `dose_f_bio` keeps the bare slice, since `PK_IDX_F` is below
  `MAX_PK_PARAMS`.
- `estimation/covariance.rs` — `extract_eigenvalues` returns `None` on an empty matrix, so the
  caller reports the non-finite objective as the reason, which is what the diagnostic exists to
  say. The doc comment records that the empty case is reachable, not defensive. The guard is
  **one** predicate (`sym.is_empty()`), not `nrows() == 0 || ncols() == 0`: for this input those
  two reject exactly the same matrices, so either could be deleted with the new test still
  green — the redundant-gate hole from #1229 / #1255. It was written the redundant way first
  and caught on a re-read of this diff.

No behaviour change for a model whose RHS reads neither builtin — pinned by a `to_bits()`
equality test, not a tolerance.

## Alternatives considered

Anchoring `TAFD` at 0 inside `solve_ekf` rather than plumbing the real array. Rejected on
measurement: it yields a finite, plausible number for a quantity whose value has not converged,
which is the same class of silent wrong answer the fix is removing (#1258, finding 4).

## Cross-repo dependency

No public-API change; `tools/update-public-api.sh --check` reports the baseline unchanged.

## Breaking changes
- [x] None

## Numerical validation

NONMEM has no extended Kalman filter, so there is no NONMEM run for an SDE `p_obs`. The oracle
is the **ODE twin as the diffusion variance goes to zero**, plus a closed form computed outside
both engines so the two cannot agree on a wrong answer.

- [x] Compared against: prior ferx commit `579d1e96`, and an ODE twin + closed form

```
# model: d/dt(central) = -(CL/V)*central*(1.0 + KTAD*TAD),  CL=1, V=20, KTAD=0.3
#        doses at t=0 and t=12; maxiter=0; ode_reltol=1e-10

# before (579d1e96)
[diffusion] central ~ 0.01     OFV = 200000000000000032768.0000    <- 2e20 sentinel
[diffusion] central ~ 1e-14    OFV = 200000000000000032768.0000
no [diffusion]                 OFV = 14.7016

# after
[diffusion] central ~ 0.01     OFV = 14.7043
[diffusion] central ~ 1e-14    OFV = 14.7016   == the ODE twin, exactly
no [diffusion]                 OFV = 14.7016

# control — an RHS reading neither builtin, unchanged either side
[diffusion] central ~ 0.01     OFV = 94.7893   (before and after)

# #1259 — same model, no omega declarations
before: thread panicked, "Unable to compute the symmetric tridiagonal decomposition
        of an empty matrix"
after:  reported as OFV: NaN with the non-finite-objective reason
```

## Performance
- [x] No significant impact expected — one `Vec` of length `MAX_PK_PARAMS + 2` per subject solve,
      and one `f64` write per dose segment.

## Tests

Three in `ode/ekf.rs`, one in `estimation/outer_optimizer_tests.rs`:

| test | what it pins |
|---|---|
| `time_reading_rhs_gets_finite_anchors_on_the_ekf_walk` | `p_obs` finite **and positive** for both slots, looped over `TAFD` and `TAD` separately |
| `a_zero_coefficient_model_time_term_is_bit_identical_to_omitting_it` | the discriminator — `0.0 * NaN` is `NaN`, so before the fix merely *mentioning* a builtin broke the run; `to_bits()` equality |
| `ekf_matches_the_ode_twin_and_a_closed_form_on_a_model_time_rhs` | the value oracle — EKF ≡ `ode_predictions` at zero diffusion, **and** ≡ a closed form computed outside both |
| `test_extract_eigenvalues_none_on_empty_matrix` | `None` on 0×0, with the measured two-conjunct condition recorded |

**Mutation-checked, each side named separately** (the #1223 lesson — a twin whose second leg
never runs is an assertion against a constant):

| mutation | result |
|---|---|
| M1 — revert the `solve_ode` site to `pk_params_flat` | all three EKF tests RED |
| M2 — delete the per-segment `TAD` anchor write | RED, message names `TAD` at `t=2` |
| M3 — make the **shared** `tad_anchor_for` return the first dose time (TAD ≡ TAFD) | RED **only on the closed form** (`left = 123.657…`, `right = 131.304…`); the twin comparison passes, because both engines take the wrong rule together |

M3 is the one worth reading: it is exactly the failure a two-engine oracle cannot see, and it
is why the third reference is in the test rather than in the PR description.

Two fixtures deliberately carry **two doses**, at `t = 0` and `t = 12`. Under a single dose
`TAD ≡ TAFD`, so a one-dose fixture agrees with an engine that plumbed one slot into both. The
twin test asserts that straddle explicitly (`TAD` and `TAFD` must differ by >10% at `t = 24`)
so it cannot quietly decay into the same assertion twice.

**No `Dual2`-vs-FD parity test is possible here.** The SDE path does not support analytic
gradients at all (`gradient_method = fd` is required — see `docs/model-file/diffusion.qmd`), so
there is no dual twin of `solve_ekf` to compare against. Stated rather than omitted.

- [x] Example added inline, and to `docs/model-file/diffusion.qmd`

## Docs
- [x] `docs/` updated — new **Model-time builtins in the RHS** section in
      `docs/model-file/diffusion.qmd`, which previously mentioned neither `TAD` nor `TAFD`.
- [x] `CHANGELOG.md` `[Unreleased]` entry added (two bullets, under `### Fixed`)

## Checklist
- [x] Gradient-reachable code untouched — `sens/` is not on this path
- [x] No version bump needed

## Docs & examples
- [x] No example execution step needed — no `examples/*.ferx` or `data/` file is affected, and
      no `.ferx` DSL syntax or `[fit_options]` key is added

## Reviewer hints

The subtle part is **where** the per-segment anchor write sits, not what it computes. It must
precede the segment's `solve_ode`; placing it after compiles cleanly and silently reads the
previous segment's anchor. `ode/predictions.rs:6884` is the existing instance of exactly that
ordering hazard.

Worth a second pair of eyes: `no_lag`. The EKF path applies no lagtime anywhere — `dose_f_bio`
reads only `f_bio`, and `active_infusions` is called with empty lag slices — so feeding
`tad_anchor_for` a zero-filled lag vector matches what the rest of the function already assumes.
If the EKF ever gains lagtime support, that vector is the thing to change, and it is commented
as such.

## Open questions

None blocking. Three adjacent findings are filed rather than folded in: #1260 (the EKF ignores
`SS` records — measured at 120.5 OFV on a nonlinear model), #1261 (the same bare-slice defect on
the survival hazard readout — `2e20` vs `41164.7786`), and #1262 (NONMEM's constant-infusion
steady state is unimplemented).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -444,6 +444,29 @@ section of the SDLC for the versioning policy).
   The covariance step's non-finite-objective diagnostic asked for the eigenvalues of the
   0×0 OMEGA such a model has, which `nalgebra` rejects. It now reports the non-finite objective
   as the failure reason, which is what the diagnostic exists to say.
+- **An infusion under `F ≠ 1` no longer delivers zero drug on an SDE (`[diffusion]`) model
+  (#1263).** Bioavailability reshapes a `RATE`-defined infusion's *duration*, not its rate, and
+  the EKF path placed the window's segment boundary at the unscaled end. The infusion window
+  then failed its own membership test in every segment, so the dose delivered no mass at all and
+  every prediction read zero. The boundary is now `F`-scaled, matching the ODE path and NONMEM
+  (anchored against `nonmem_anchor/oral_central_inf_advan2_f06`).
+- **An SDE model whose records start after `t = 0` no longer inflates its observation variance
+  (#1263).** The EKF began integrating at a hard-coded `t = 0` rather than at the subject's first
+  record, so the covariance accumulated process noise across a segment that does not exist —
+  for a first dose at `t = 24`, thirteen times the correct value at the first observation. An
+  `init(state) = …` starting amount was decayed across the same phantom segment.
+- **Two observation records at the same time no longer read as zero on an SDE model (#1263).**
+  Inside a dose segment the EKF kept one record per observation time, so a second record at that
+  instant kept its initialised `0.0` prediction and variance and fed a plausible zero into the
+  likelihood. All records at one instant now share a single filter update, as they already did
+  at a dose boundary.
+- **Dosing features the EKF/SDE path does not implement now warn instead of returning a
+  plausible wrong answer (#1263, #1260).** An `SS=1` record is applied as a single dose rather
+  than equilibrated (`W_SDE_STEADY_STATE`), and an absorption `lagtime` / `ALAGn` is ignored
+  (`W_SDE_LAGTIME`) — joining the existing `W_SDE_RESET`. Neither gap is visible in `IPRED`, and
+  the steady-state one is large: a 1-cpt model with one `SS=1, II=12` record predicts `90.48`
+  where the equivalent explicit dose train predicts `200.27`. Expand the steady state into
+  explicit records, or fit without `[diffusion]`.
 
 ## [0.3.1] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -430,6 +430,20 @@ section of the SDLC for the versioning policy).
   `minimization_successful` fraction. Re-running a bootstrap of an unchanged model can therefore
   report slightly different CIs than before; `--summarize` over a stored `raw_results.csv`
   re-applies the criteria without refitting.
+- **An `[odes]` right-hand side that reads `TAD` / `TAFD` no longer destroys the objective on an
+  SDE (`[diffusion]`) model (#1131).** The extended-Kalman-filter path handed the compiled RHS a
+  bare PK parameter array, two slots shorter than the one the ODE predictors build, so both
+  model-time anchors read as missing and the RHS injected `NaN`. The state was then clamped back
+  to a plausible-looking number while the observation variance kept the `NaN`, so `ipred` looked
+  right and the fit silently reported the diverged-subject sentinel (`OFV` ≈ `2e20`) instead of a
+  real objective. The EKF now carries the same extended array and re-anchors `TAD` per dose
+  segment by the same rule as the two ODE predictors, so an SDE fit of a time-varying RHS agrees
+  with its zero-diffusion ODE twin. A model whose RHS reads neither builtin is bit-identical to
+  before.
+- **A model with no random effects no longer panics when its objective is non-finite (#1259).**
+  The covariance step's non-finite-objective diagnostic asked for the eigenvalues of the
+  0×0 OMEGA such a model has, which `nalgebra` rejects. It now reports the non-finite objective
+  as the failure reason, which is what the diagnostic exists to say.
 
 ## [0.3.1] - 2026-09-02
 
--- a/docs/model-file/diffusion.qmd
+++ b/docs/model-file/diffusion.qmd
@@ -140,6 +140,33 @@ the ODE alone explains the data; a large value suggests model misspecification.
 | State name not in `states = [...]` | Parse-time error |
 | Negative initial value | Parse-time error |
 
+## Model-time builtins in the RHS
+
+An `[odes]` right-hand side under `[diffusion]` may read the solver-injected builtins
+`TIME` / `T`, `TAFD` and `TAD`, with the same meaning as on the ordinary ODE path — see
+[TAD before the first dose arrives](ode-models.qmd#tad-before-the-first-dose-arrives) for
+what `TAD` reads in the window before an arrival, and for the dose-free case.
+
+```
+[odes]
+  d/dt(central) = -(CL/V) * central * (1.0 + KTAD * TAD)
+[diffusion]
+  central ~ 0.01
+```
+
+`TAD` is re-anchored at each dose segment by the same rule the two ODE predictors use, so
+an SDE fit of a time-varying right-hand side reduces to its ODE twin as the diffusion
+variance goes to zero. Measured on the model above (`CL = 1`, `V = 20`, `KTAD = 0.3`, doses
+at 0 and 12): `OFV = 14.7016` from the EKF at `central ~ 1e-14` and `14.7016` from the same
+model with the `[diffusion]` block removed.
+
+`TIME` / `T` is the integration variable itself and has always been readable here. `TAFD`
+and `TAD` are passed in two extra parameter slots, and until #1131 the EKF was handed an
+array two slots short, so both evaluated to `NaN` for the whole trajectory. That failure was
+easy to miss: `IPRED` is computed on the ordinary ODE path and stayed correct, while the
+objective — which is what the filter actually contributes — became the diverged-subject
+sentinel. A right-hand side reading neither builtin was never affected.
+
 ## ferx-r usage note
 
 When using ferx-core from R via the ferx-r package, the `[diffusion]` block is part
--- a/docs/model-file/diffusion.qmd
+++ b/docs/model-file/diffusion.qmd
@@ -171,6 +171,12 @@ equivalence:
 | `lagtime` / `ALAGn` | `W_SDE_LAGTIME` | every dose is applied at its record time |
 | `SS=1` steady-state dose | `W_SDE_STEADY_STATE` | the record is applied as a single dose, not equilibrated |
 
+These are **fitting** gaps, not prediction gaps. The filter runs only inside the objective, so
+`predict()` and `simulate()` on the same model go through the ordinary ODE path, which does
+equilibrate a steady-state dose and does apply a lag time. A model can therefore show a correct
+`IPRED` and a wrong objective at the same time — which is why the warnings are raised by `fit()`
+and `ferx check` rather than at prediction time.
+
 The steady-state gap is the one most likely to be mistaken for a converged answer: on a
 1-cpt model with `CL = 1`, `V = 20`, one `SS=1, II=12` record and an autonomous right-hand
 side, the EKF returns `90.48` where an explicit eleven-dose train returns `200.27`. The
--- a/docs/model-file/diffusion.qmd
+++ b/docs/model-file/diffusion.qmd
@@ -160,6 +160,25 @@ variance goes to zero. Measured on the model above (`CL = 1`, `V = 20`, `KTAD = 
 at 0 and 12): `OFV = 14.7016` from the EKF at `central ~ 1e-14` and `14.7016` from the same
 model with the `[diffusion]` block removed.
 
+That reduction holds for the dataset above. It is **not** unconditional, because the
+EKF/SDE path does not implement every dosing feature the ODE path does. Each gap below
+raises a warning rather than failing, so check the fit's warnings before relying on the
+equivalence:
+
+| Dataset feature | Code | What the SDE path does instead |
+|---|---|---|
+| `EVID=3`/`4` system reset | `W_SDE_RESET` | the reset is ignored; compartment amounts carry through |
+| `lagtime` / `ALAGn` | `W_SDE_LAGTIME` | every dose is applied at its record time |
+| `SS=1` steady-state dose | `W_SDE_STEADY_STATE` | the record is applied as a single dose, not equilibrated |
+
+The steady-state gap is the one most likely to be mistaken for a converged answer: on a
+1-cpt model with `CL = 1`, `V = 20`, one `SS=1, II=12` record and an autonomous right-hand
+side, the EKF returns `90.48` where an explicit eleven-dose train returns `200.27`. The
+`TAD` anchor is folded into `[0, II)` for that record as though the train existed, so a
+right-hand side reading `TAD` diverges further still — `45.60` against `36.74` at `t = 150`
+for an `SS=1` infusion whose end lands inside a dosing interval. Expand the steady state
+into explicit dose records, or fit the model without `[diffusion]`.
+
 `TIME` / `T` is the integration variable itself and has always been readable here. `TAFD`
 and `TAD` are passed in two extra parameter slots, and until #1131 the EKF was handed an
 array two slots short, so both evaluated to `NaN` for the whole trajectory. That failure was
--- a/src/api/tests/sde_integration.rs
+++ b/src/api/tests/sde_integration.rs
@@ -297,3 +297,168 @@ fn sde_emits_experimental_warning() {
         "non-SDE model should not emit W_EXPERIMENTAL_SDE"
     );
 }
+
+/// #1263: dosing features the EKF/SDE path silently drops must **warn**, mirroring the
+/// `W_SDE_RESET` precedent, rather than return a plausible wrong number.
+///
+/// `solve_ekf` applies an `SS=1` record as a single bolus with no equilibration, and
+/// never calls `DoseAttrMap::lagtime` at all. Neither gap fails, neither is visible in
+/// `IPRED` (the likelihood takes only `p_obs` from the filter), and both are large:
+/// measured on a 1-cpt autonomous model, one `SS=1, II=12` record gives `90.48` against
+/// `200.27` for the equivalent explicit train.
+///
+/// Each case is asserted against its **own** control — the same population without the
+/// feature — so a warning that fired unconditionally would fail here too.
+mod sde_unsupported_dosing_warnings {
+    use super::*;
+
+    /// The shared SDE model with a `lagtime` bound, so `model.has_lagtime()` is true.
+    const SDE_LAG_MODEL_SRC: &str = r#"
+[parameters]
+  theta TVCL(5.0, 0.1, 50.0)
+  theta TVV(50.0, 1.0, 500.0)
+  theta TVLAG(0.5, 0.01, 5.0)
+  omega ETA_CL ~ 0.09
+  sigma ADD ~ 1.0 FIX
+
+[individual_parameters]
+  CL      = TVCL * exp(ETA_CL)
+  V       = TVV
+  lagtime = TVLAG
+
+[structural_model]
+  ode(obs_cmt=central, states=[central])
+
+[odes]
+  d/dt(central) = -(CL/V) * central
+
+[diffusion]
+  central ~ 0.5
+
+[error_model]
+  DV ~ additive(ADD)
+
+[fit_options]
+  method = foce
+"#;
+
+    /// The same, without `[diffusion]` — the non-SDE control for the lag warning.
+    const ODE_LAG_MODEL_SRC: &str = r#"
+[parameters]
+  theta TVCL(5.0, 0.1, 50.0)
+  theta TVV(50.0, 1.0, 500.0)
+  theta TVLAG(0.5, 0.01, 5.0)
+  omega ETA_CL ~ 0.09
+  sigma ADD ~ 1.0 FIX
+
+[individual_parameters]
+  CL      = TVCL * exp(ETA_CL)
+  V       = TVV
+  lagtime = TVLAG
+
+[structural_model]
+  ode(obs_cmt=central, states=[central])
+
+[odes]
+  d/dt(central) = -(CL/V) * central
+
+[error_model]
+  DV ~ additive(ADD)
+
+[fit_options]
+  method = foce
+"#;
+
+    fn codes(src: &str, pop: &Population) -> Vec<String> {
+        let parsed = parse_full_model(src).expect("model parses");
+        let m = &parsed.model;
+        super::super::check_model_data_warnings(m, pop, &m.default_params)
+            .iter()
+            .map(|d| d.code.clone())
+            .collect()
+    }
+
+    /// `make_sde_population` with every subject's dose flipped to `SS=1, II=12`.
+    fn ss_population() -> Population {
+        let mut pop = make_sde_population();
+        for s in &mut pop.subjects {
+            s.doses = vec![DoseEvent::new(0.0, 100.0, 1, 0.0, true, 12.0)];
+        }
+        pop
+    }
+
+    #[test]
+    fn a_lagtime_under_diffusion_warns() {
+        let pop = make_sde_population();
+        assert!(
+            codes(SDE_LAG_MODEL_SRC, &pop).contains(&"W_SDE_LAGTIME".to_string()),
+            "a [diffusion] model declaring lagtime must raise W_SDE_LAGTIME"
+        );
+        // Control 1 — the same lagtime with no [diffusion] block. Isolates the SDE half
+        // of the predicate: without this the warning could be keyed on lagtime alone.
+        assert!(
+            !codes(ODE_LAG_MODEL_SRC, &pop).contains(&"W_SDE_LAGTIME".to_string()),
+            "an ODE model with a lagtime must NOT raise W_SDE_LAGTIME"
+        );
+        // Control 2 — [diffusion] with no lagtime. Isolates the other half.
+        assert!(
+            !codes(SDE_MODEL_SRC, &pop).contains(&"W_SDE_LAGTIME".to_string()),
+            "a [diffusion] model without a lagtime must NOT raise W_SDE_LAGTIME"
+        );
+    }
+
+    #[test]
+    fn a_steady_state_dose_under_diffusion_warns() {
+        let ss = ss_population();
+        let plain = make_sde_population();
+        assert!(
+            codes(SDE_MODEL_SRC, &ss).contains(&"W_SDE_STEADY_STATE".to_string()),
+            "a [diffusion] model with SS=1 records must raise W_SDE_STEADY_STATE"
+        );
+        // Control 1 — the same model, same subjects, doses not flagged SS.
+        assert!(
+            !codes(SDE_MODEL_SRC, &plain).contains(&"W_SDE_STEADY_STATE".to_string()),
+            "a [diffusion] model with no SS record must NOT raise W_SDE_STEADY_STATE"
+        );
+        // Control 2 — the same SS records with no [diffusion] block; the ODE path does
+        // equilibrate, so there is nothing to warn about.
+        assert!(
+            !codes(BASE_MODEL_SRC, &ss).contains(&"W_SDE_STEADY_STATE".to_string()),
+            "an ODE model with SS=1 records must NOT raise W_SDE_STEADY_STATE"
+        );
+    }
+
+    /// The message has to name the workaround, not just the gap — a warning that says
+    /// only "not supported" leaves the user with a silently wrong objective and no move.
+    #[test]
+    fn the_messages_say_what_to_do_instead() {
+        let parsed = parse_full_model(SDE_MODEL_SRC).expect("model parses");
+        let m = &parsed.model;
+        let diags = super::super::check_model_data_warnings(m, &ss_population(), &m.default_params);
+        let ss = diags
+            .iter()
+            .find(|d| d.code == "W_SDE_STEADY_STATE")
+            .expect("W_SDE_STEADY_STATE present");
+        assert_eq!(ss.severity, crate::diagnostics::Severity::Warning);
+        assert!(
+            ss.message.contains("explicit dose train"),
+            "the SS warning must point at the expansion workaround: {}",
+            ss.message
+        );
+
+        let parsed_lag = parse_full_model(SDE_LAG_MODEL_SRC).expect("model parses");
+        let ml = &parsed_lag.model;
+        let lag_diags =
+            super::super::check_model_data_warnings(ml, &make_sde_population(), &ml.default_params);
+        let lag = lag_diags
+            .iter()
+            .find(|d| d.code == "W_SDE_LAGTIME")
+            .expect("W_SDE_LAGTIME present");
+        assert_eq!(lag.severity, crate::diagnostics::Severity::Warning);
+        assert!(
+            lag.message.contains("record time"),
+            "the lag warning must say what happens instead: {}",
+            lag.message
+        );
+    }
+}
--- a/src/api/validation.rs
+++ b/src/api/validation.rs
@@ -3971,6 +3971,51 @@ pub fn check_model_data_warnings(
         }
     }
 
+    // Absorption lag times are not applied on the EKF/SDE path. `solve_ekf` never calls
+    // `DoseAttrMap::lagtime`: the bolus lands at the raw `dose.time`, `active_infusions`
+    // is handed an empty lag slice, and the `TAD` anchor is computed at zero lag. Same
+    // silent-drop shape as `W_SDE_RESET` above (#1263 review).
+    if model.is_sde() && model.has_lagtime() {
+        diags.push(Diagnostic::warning(
+            "W_SDE_LAGTIME",
+            "This model declares an absorption lag time (`lagtime` / `ALAGn`) with a \
+             [diffusion] (SDE) model. Lag times are not yet honoured on the EKF/SDE \
+             path — every dose is applied at its record time and the lag is ignored. \
+             Use an ODE or analytical model if the lag matters."
+                .to_string(),
+        ));
+    }
+
+    // Steady-state doses are not equilibrated on the EKF/SDE path. `solve_ekf` applies an
+    // `SS=1` record as a single bolus and never runs the equilibration, so the state is
+    // the one-dose state while `tad_anchor_for`'s `ss` branch hands the RHS a `TAD`
+    // folded into `[0, II)` — an anchor describing a periodic pulse train that this
+    // engine did not build. Measured on #1263: 55% low against an explicit train on an
+    // autonomous model (90.48 vs 200.27), and a further 24.1% `ipred` divergence at
+    // t=150 for an `SS=1` infusion whose end break lands past a virtual pulse. Both are
+    // silent — hence a warning rather than a quiet wrong answer (#1260).
+    if model.is_sde() {
+        let n_ss_sde = population
+            .subjects
+            .iter()
+            .filter(|s| s.has_periodic_ss_dose())
+            .count();
+        if n_ss_sde > 0 {
+            diags.push(Diagnostic::warning(
+                "W_SDE_STEADY_STATE",
+                format!(
+                    "{} subject(s) have SS=1 dose records with a [diffusion] (SDE) \
+                     model. Steady-state doses are not yet equilibrated on the EKF/SDE \
+                     path — the record is applied as a single dose, so predictions and \
+                     the objective reflect a one-dose history rather than a steady \
+                     state. Use an ODE or analytical model, or expand the steady state \
+                     into an explicit dose train.",
+                    n_ss_sde
+                ),
+            ));
+        }
+    }
+
     // Negative typical-value lag time at the initial point (eta = 0).
     if model.has_lagtime() {
         if let Some(first_subj) = population.subjects.first() {
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -45,6 +45,8 @@
 //! | `W_STEADY_STATE_II`       | SS=1 dose with missing / non-positive II |
 //! | `W_STEADY_STATE_INFUSION` | SS=1 infusion with `T_inf > II` (overlapping pulses) |
 //! | `W_SDE_RESET`             | EVID=3/4 resets under an SDE model are not honoured |
+//! | `W_SDE_LAGTIME`           | an absorption lag time under an SDE model is not honoured |
+//! | `W_SDE_STEADY_STATE`      | an `SS=1` dose under an SDE model is not equilibrated |
 //! | `W_EXPERIMENTAL_SDE`      | an SDE (`[diffusion]`) model uses an experimental feature (see Feature Maturity docs) |
 //! | `W_EXPERIMENTAL_NN`       | a neural-network (`[covariate_nn]`) model uses an experimental feature (see Feature Maturity docs) |
 //! | `W_NEGATIVE_LAGTIME`      | a lag time is negative at the initial estimates |
--- a/src/estimation/covariance.rs
+++ b/src/estimation/covariance.rs
@@ -177,8 +177,25 @@ fn fmt_eig(v: f64) -> String {
     }
 }
 
-/// Eigenvalues of `sym` sorted descending. Returns `None` if any eigenvalue is non-finite.
+/// Eigenvalues of `sym` sorted descending. Returns `None` if any eigenvalue is non-finite,
+/// or if `sym` is empty.
+///
+/// The empty case is reachable, not defensive: a model with **no random effects** has a
+/// 0×0 Omega, and the non-finite-objective diagnostic below asks this function to inspect
+/// it precisely when the fit has already gone wrong. `nalgebra`'s `SymmetricEigen::new`
+/// panics outright on a 0×0 input ("Unable to compute the symmetric tridiagonal
+/// decomposition of an empty matrix"), so an `n_eta = 0` fit whose objective went
+/// non-finite aborted with that message instead of reporting the objective — the diagnostic
+/// killed the run it was there to explain. `None` is the existing "cannot diagnose" signal
+/// and every caller already handles it.
+///
+/// One predicate, not `nrows() == 0 || ncols() == 0`: those two reject exactly the same
+/// inputs here, so either could be deleted with the guard still passing its own test —
+/// the redundant-gate hole from #1229 / #1255.
 pub(crate) fn extract_eigenvalues(sym: &DMatrix<f64>) -> Option<Vec<f64>> {
+    if sym.is_empty() {
+        return None;
+    }
     let eig = SymmetricEigen::new(sym.clone());
     if eig.eigenvalues.iter().any(|l| !l.is_finite()) {
         return None;
--- a/src/estimation/outer_optimizer_tests.rs
+++ b/src/estimation/outer_optimizer_tests.rs
@@ -441,6 +441,25 @@ fn test_extract_eigenvalues_none_on_nan() {
     );
 }
 
+/// extract_eigenvalues returns None — rather than panicking — for a 0x0 matrix.
+///
+/// A model with no random effects has a 0x0 Omega, and the non-finite-objective diagnostic
+/// in `compute_covariance` inspects exactly that matrix when a fit has already gone wrong.
+/// `nalgebra`'s `SymmetricEigen::new` panics on an empty input, so before this guard an
+/// `n_eta = 0` fit whose objective went non-finite aborted with "Unable to compute the
+/// symmetric tridiagonal decomposition of an empty matrix" instead of reporting the
+/// objective. Both conjuncts are required to reach it — measured: `n_eta = 1` with a
+/// non-finite objective returns the `1e20` sentinel and does not panic, and `n_eta = 0`
+/// with a finite objective never enters this branch.
+#[test]
+fn test_extract_eigenvalues_none_on_empty_matrix() {
+    let h: DMatrix<f64> = DMatrix::zeros(0, 0);
+    assert!(
+        extract_eigenvalues(&h).is_none(),
+        "a 0x0 matrix must return None, not panic"
+    );
+}
+
 /// packed_param_label decodes the lower-triangular block-omega index correctly.
 /// Packing order (column-major lower triangle): (0,0), (1,0), (1,1).
 /// With n_theta=2, packed_idx=3 → omega_idx=1 → (row=1, col=0).
--- a/src/ode/ekf.rs
+++ b/src/ode/ekf.rs
@@ -1126,7 +1126,15 @@ mod tests {
 
         for (i, &t) in obs_times.iter().enumerate() {
             let p_analytic = (sigma2_w / (2.0 * ke)) * (1.0 - (-2.0 * ke * t).exp());
-            // Euler covariance propagation introduces O(Δt²) error; 5% tolerance is adequate.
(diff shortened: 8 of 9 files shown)


TeunP:
Review at max effort against `24d011d1`. Fifteen findings, ranked most-severe first; each says
whether it was **measured** or established by **code read**, and the measured ones carry the
command and the raw numbers. A list of candidates I checked and **rejected** is at the end, so
nobody re-derives a dead end.

Headline: the fix itself is right and the ordering is correct, but two things are worth acting
on before merge — the `earliest_dose_time` half of it is not pinned by any test (finding 2,
measured), and the same bare-slice defect this PR fixes in the EKF is live on twelve
`solve_ode` sites in the SS equilibration (finding 1, measured), which is the actual mechanism
behind #1139's NaN rows.

---

## 1. `src/ode/predictions.rs:1174` — correctness — MEASURED

SS equilibration hands the compiled `[odes]` RHS a bare 128-slot slice.

The compiled RHS reads TAFD/TAD from `params[MAX_PK_PARAMS]` / `[+1]`
(`parser/model_parser.rs:11851-11864`, `MAX_PK_PARAMS = 128`, `types.rs:391`). The
SS-equilibration solves pass the bare `pk_params_flat` (128), so `.get(128)`/`.get(129)`
return `None` and the RHS injects `NaN`. Twelve call sites: `ode/predictions.rs` 672, 1174,
1217, 1239, 1251, 1266, 1310, 1324, 1341 (`equilibrate_ss_pk_state`, 1062-1399), 1454, 1467,
1477 (`ss_state_at_phase_pk`, 1415-1499).

**Scope, measured.** I inserted at `model_parser.rs`, immediately before the TAFD injection:

```rust
debug_assert!(
    "SHORTSLICE: compiled [odes] RHS handed {} params, needs {}",
    params.len(), crate::types::MAX_PK_PARAMS + 2
);
```

on PR commit `24d011d1`, then:

```
cargo test --lib
```

Result: `test result: FAILED. 4048 passed; 35 failed`. All 35 panic with `SHORTSLICE`. 34
report `handed 128 params, needs 130` and every one names an SS route:

```
api::adaptive_sim_tests::adaptive_ss_base_dose_matches_static_predict
api::adaptive_sim_tests::adaptive_ss_base_dose_with_reset_matches_static_predict
estimation::sens_outer_gradient::tests::population_packed_gradient_ode_ss_reset_matches_fd
sens::ode_provider::tests::ode_provider_cmt_zero_steady_state_matches_the_default_compartment_and_fd
sens::ode_provider::tests::ode_provider_ss_absorption_nonlinear_converged_dual_matches_production
sens::ode_provider::tests::ode_provider_ss_absorption_nonlinear_disposition_matches_production
sens::ode_provider::tests::ode_provider_ss_bolus_matches_production
sens::ode_provider::tests::ode_provider_ss_first_order_1cpt_matches_production
sens::ode_provider::tests::ode_provider_ss_first_order_2cpt_matches_production
sens::ode_provider::tests::ode_provider_ss_first_order_bioavailability_matches_production
sens::ode_provider::tests::ode_provider_ss_first_order_cmt_zero_matches_the_default_compartment_and_fd
sens::ode_provider::tests::ode_provider_ss_first_order_iov_matches_predict_iov
sens::ode_provider::tests::ode_provider_ss_igd_1cpt_matches_production
sens::ode_provider::tests::ode_provider_ss_infusion_matches_production
sens::ode_provider::tests::ode_provider_ss_lagtime_infusion_matches_production
sens::ode_provider::tests::ode_provider_ss_lagtime_infusion_tvcov_matches_production
sens::ode_provider::tests::ode_provider_ss_lagtime_matches_production
sens::ode_provider::tests::ode_provider_ss_lagtime_reset_inside_the_pre_arrival_window_matches_production
sens::ode_provider::tests::ode_provider_ss_lagtime_tvcov_matches_production
sens::ode_provider::tests::ode_provider_ss_linear_bolus_uses_exact_solve
sens::ode_provider::tests::ode_provider_ss_modeled_duration_matches_production
sens::ode_provider::tests::ode_provider_ss_nonlinear_fallback_early_stop_matches_full_budget
sens::ode_provider::tests::ode_provider_ss_rate_defined_infusion_under_f_matches_production
sens::ode_provider::tests::ode_provider_ss_rate_under_f_uses_exact_solve
sens::ode_provider::tests::ode_provider_ss_transit_1cpt_matches_production
sens::ode_provider::tests::ode_provider_ss_tvcov_matches_production
sens::ode_provider::tests::ode_provider_ss_weibull_1cpt_matches_production
sens::ode_provider::tests::ode_provider_tvcov_init_ss_matches_production
sens::ode_provider::tests::ode_time_builtin_with_init_and_ss_matches_production
sens::provider::tests::ode_iov_modeled_duration_ss_matches_fd_of_predict_iov
sens::provider::tests::ode_iov_ss_lagtime_provider_matches_fd_of_predict_iov
sens::provider::tests::ode_iov_ss_provider_matches_fd_of_predict_iov
sens::provider::tests::ode_iov_ss_rate_defined_infusion_under_f_matches_fd_of_predict_iov
sens::provider::tests::provider_ss_into_plain_cmt_with_inert_absorption_matches_fd
```

The 35th, `parser::model_parser::tests::test_ode_block_supports_if_statements`, reports
`handed 2 params` — a parser unit test calling the RHS directly, a test artifact, not
production.

slice length.

**Attribution, measured, one variable.** #1139 attributes its TAD/TAFD NaN rows to
`tad_anchor` returning `NaN`. That is not the mechanism. `tad_anchor` is never consulted
during equilibration — `apply_segment_boundary` writes the anchor at `predictions.rs:6899`,
*after* the `equilibrate_ss_state` call at `:6872` — and the equilibration receives
`pk_params_flat`, not `ext_params`. It also cannot explain the TAFD row at all: TAFD's anchor
comes from `seed_ext_params`, never from `tad_anchor`.

Experiment, on `main` = `579d1e96`, in a throwaway worktree. #1139's fixture: 1-cpt IV bolus,
`CL=1`, `V=20`, one `SS=1` dose, `II=12`, record at `t=120`, obs at 122/125/128/132. RHS
written to mirror the parser injection exactly
(`p.get(slot).copied().filter(|x| x.is_finite()).map_or(f64::NAN, |a| t - a)`), then
`ode_predictions`:

```
cargo test --lib attribution_ss_shortslice -- --nocapture
```

| | `0.0*TAD` | `0.0*TAFD` |
|---|---|---|
| `579d1e96` unmodified | `[NaN, NaN, NaN, NaN]` | `[NaN, NaN, NaN, NaN]` |
| + extended slice, nothing else changed | `[200.54537984215145, 172.61100876660328, 148.56767266778016, 121.63692553562518]` | identical |

The only change was shadowing `pk_params_flat` at the top of `equilibrate_ss_pk_state` and
`ss_state_at_phase_pk` with a full-length `[f64; MAX_PK_PARAMS + 2]` copy whose two anchor
slots are set to a finite `0.0`.

`200.54537984215145 / 20 = 10.027268992107572`, against #1139's own 40-cycle pulse-train
oracle of `10.027269`. So the short slice alone accounts for the entire `0.0*TAD` row of
#1139, to 8 significant figures.

**This does not close #1139.** The anchor in the experiment was an arbitrary `0.0`, so the
non-zero rows (`0.03*TAD`, train `8.890201`) still need a real cycle-local anchor, and the
`T`/`TIME` rows (~17% wrong) are a separate defect. Also note the equilibration runs on a
**cycle-local clock** (`seg_start = m * dose.ii`), so "pass `ext_params`" is not by itself the
fix — what the anchors should be on that clock is T2's question.

---

## 2. `src/ode/ekf.rs:191` — test-coverage — MEASURED

The TAFD half of the fix deletes clean. This is the fourth mutation the PR body does not
record, and it is the smallest edit that removes half the fix.

Mutation applied to `24d011d1` in a clean worktree:

```rust
// was:
let mut ext_params = crate::ode::predictions::seed_ext_params(
    pk_params_flat,
    crate::ode::predictions::earliest_dose_time(doses),
);
// mutation:
let mut ext_params = crate::ode::predictions::seed_ext_params(pk_params_flat, 0.0);
```

```
cargo test --lib ode::ekf::tests   ->  12 passed; 0 failed
cargo test --lib                   ->  4083 passed; 0 failed; 23 ignored   (244.37s)
```

Baseline for comparison, same worktree, unmutated: `12 passed` / `4083 passed`.

**Why nothing catches it.** All three new fixtures dose at `t = 0` and `t = 12`, so
`earliest_dose_time(doses)` is identically `0.0` on every one. And the twin test's own oracle
hard-codes the TAFD leg: `phase(0.0, t, 0.0)` with `let anchor = if restarts { t2 } else { 0.0 };`
— so the closed form cannot distinguish `earliest_dose_time(doses)` from a literal `0.0`
either. A fixture whose first dose is at `t > 0` kills it, and would also exercise the
documented pre-first-dose window where both anchors go negative.

This is the #1171 shape that M1–M3 set out to avoid: three mutations of the author's own fix
all died correctly, and a different minimal edit walked through the whole suite.

---

## 3. `src/ode/ekf.rs:212` — correctness — READ

Infusion break time is not `F`-scaled; the ODE path scales it.

`ekf.rs:212` pushes `dose.time + dose.duration`. `active_infusions` — called at `ekf.rs:341`,
130 lines below in the same function — computes the window as `start + dur_eff` where
`dur_eff` comes from `bioavailable_infusion` (`types.rs:319-324`), i.e. `f_bio * duration` for
`InfusionDef::RateDefined`. The ODE path's `collect_dose_break_times` pushes
`dose.time + lag + dur_eff` (`predictions.rs:2936-2937`) under the comment "Must match
`active_infusions`'s window so each segment is fully inside or outside every infusion".

**Failure scenario.** RATE-defined infusion, `duration = 10`, `F = 0.5`, dosed at `t = 0`. EKF
breaks at `t = 10`; the true window ends at `t = 5`. In segment `[0, 10]`, `active_infusions`'
admission test (`predictions.rs:1528-1531`) is `start <= t_start + EPS && end >= t_end - EPS`
i.e. `5 >= 10 - eps` → false. The infusion is never admitted and delivers **zero mass** into
the EKF walk, while `ode_predictions` — called on the line above in the same wrapper
(`predictions.rs:6224`) — gets it right.

**Not covered.** `ekf_applies_per_compartment_bioavailability` sets `pk[PK_IDX_F] = 0.5` and
`F2 -> slot 9`, but constructs both doses as `DoseEvent::new(0.0, 100.0, N, 0.0, false, 0.0)`
— `rate = 0.0`, i.e. boluses. No EKF test pairs `F != 1` with an infusion.

---

## 4. `src/ode/ekf.rs:208` — correctness — READ

`break_times` starts at a literal `0.0`, not the subject's start.

`ekf.rs:208`: `let mut break_times: Vec<f64> = vec![0.0];`
`predictions.rs:3211`: `let mut break_times: Vec<f64> = vec![subject_integration_start(subject)];`
where `subject_integration_start` (`predictions.rs:312-332`) is the min over obs / dose /
pk-only / reset times, falling back to `0.0` only when there are no events at all.

**Failure scenario.** Subject with first dose at `t = 24`, observations at 26 and 30, plus
`[diffusion] central ~ q`. EKF `break_times = [0, 24, 30]`. Segment `[0, 24]` survives the
`(t_end - t_start).abs() < 1e-15` guard at `ekf.rs:329` and runs `propagate_covariance` over
`dt = 24` in `ceil(24/0.5) = 48` sub-steps, so `p_mat` reaches the first observation carrying
approximately `q * 24` of process noise the ODE path never accumulates. `p_obs` feeds
`V_total` directly (`likelihood.rs:892` takes `let (_, p_obs) = ...`), so the objective is
wrong for any subject whose records do not start at 0.

With an `init(state) = expr` block — which `solve_ekf`'s own doc comment says it honours, and
which `ode_predictions_ekf_with_diffusion` supplies via `&ode.initial_state(pk_params_flat)`
(`predictions.rs:6231`) — the phantom segment additionally decays the seeded mean across
`[0, 24]` before the ODE twin has started, so `ipred` diverges too.

---

## 5. `src/ode/ekf.rs:200` — correctness — READ

`obs_map` keeps one index per time; duplicates keep the `0.0` prefill.

```rust
let obs_map: HashMap<u64, usize> = obs_times
    .iter().enumerate().map(|(i, &t)| (t.to_bits(), i)).collect();
```

One index per bit pattern (later insert wins). `predictions.rs`'s `build_obs_index_map` maps
to `HashMap<u64, Vec<usize>>` specifically because "Multiple observations can share a time
(e.g. simultaneous PK/PD samples on different CMTs), so each time maps to *all* its
observation indices — recording only one would leave the others at their initial NaN".

The boundary-read path at `ekf.rs:246-266` was hardened for exactly this in #1226 and
documents it, including: "a dropped index would keep its `EkfObsPoint { ipred: 0.0,
p_obs: 0.0 }` prefill and feed a finite, plausible zero into the likelihood." The
**in-segment** path at `ekf.rs:405` was not extended.

**Failure scenario.** Two records at an identical time strictly inside a segment. `saveat`
(built at `ekf.rs:~310`) dedups them at `1e-15` to a single solver point;
`obs_map.get(&pt.t.to_bits())` returns one index; every other index keeps its
`EkfObsPoint { ipred: 0.0, p_obs: 0.0 }` prefill from `ekf.rs:176-182` — a finite, plausible
zero into the likelihood with no process-noise inflation at all. `RecordIndex`, already
constructed at `ekf.rs:245`, can serve both paths and `obs_map` then goes away.

---

## 6. `src/ode/ekf.rs:336` — correctness — READ, mechanism only

The SS branch of `tad_anchor_for` wraps TAD for a train the EKF never built.

`tad_anchor_for` (`predictions.rs:2360-2391`) takes `if d.ss && d.ii > 0.0 { let elapsed =
t_start - (d.time + lag); t_start - elapsed.rem_euclid(d.ii) }`, i.e. it wraps TAD into
`[0, II)` as though a steady-state pulse train existed. `solve_ekf` applies an `SS=1` dose as
a single bolus with no equilibration (`ekf.rs:271-274`) — that is #1260. Nothing rejects
`SS` + `[diffusion]` (I enumerated every `model.is_sde()` / `diffusion_var.is_empty()` gate in
`api/validation.rs`: 955, 2732, 2796, 3192, 3954, 4410 — none covers SS).

**Failure scenario.** `SS=1, II=12` dose with an `[odes]` RHS reading `TAD`, under
`[diffusion]`. Before #1263 the anchor was NaN, so the objective went to the `2e20`
diverged-subject sentinel — loud, diagnosable. After it, TAD is a finite plausible number
consistent with a steady state that was never equilibrated, so the fit returns a wrong
objective silently. That is the same class the PR's own "Alternatives considered" rejects for
TAFD; the reasoning was not applied to the SS branch the fix newly reaches.

Also mesh-sensitive: `t_start - elapsed.rem_euclid(ii)` equals `tau + ii*floor((t_start-tau)/ii)`,
a step function of `t_start` with jumps at the virtual pulse times. Those pulse times are
breaks in neither engine, and the EKF's break set (doses + infusion ends + `t_last`) is
coarser than the ODE path's, so the two can land in different inter-pulse intervals and
disagree.

---

## 7. `src/api/validation.rs:3954` — correctness — READ (four independent checks)

No `W_SDE_LAGTIME` beside the `W_SDE_RESET` precedent.

The PR's `no_lag` claim is **correct**. I verified it four ways, independently of the PR's
comment:

1. `dose_f_bio` (`ekf.rs:217-220`) resolves only `DoseAttr::F` / `PK_IDX_F` via
   `DoseAttrMap::f_bio` (`types.rs:687-689`).
2. `active_infusions(&[], doses, t_start, t_end, &[], &dose_f_bio, NEG_INFINITY, n)`
   (`ekf.rs:341-350`) reads every lag as `dose_lagtimes.get(k).copied().unwrap_or(0.0)`
   (`predictions.rs:1520`) — an empty slice means zero lag, not a panic.
3. The bolus lands at the raw `dose.time` (`ekf.rs:271`), with no lag term.
4. `DoseEvent` carries no lag field (`types.rs:92-115`); lag exists only via
   `DoseAttrMap::lagtime` (`types.rs:693-695`), which is never called from `ekf.rs`.

So the zero-filled vector is internally consistent. **The finding is the consequence:** a
model declaring `lagtime` or `ALAG1` with `[diffusion]` has it dropped with no error and no
warning, and `no_lag` now cements that into the TAD anchor too — while the exactly analogous
silent drop *does* warn (`W_SDE_RESET`, `validation.rs:3954-3971`: "System resets are not yet
honoured on the EKF/SDE path"). A `W_SDE_LAGTIME` gated on the existing `model.has_lagtime()`
(used at `validation.rs:3974`) mirrors it in about ten lines.

---

## 8. `src/parser/model_parser.rs:11851` — altitude — MEASURED

Silent NaN on a short slice; one assert catches the whole class.

`params.get(MAX_PK_PARAMS).copied().filter(|v| v.is_finite()).map_or(f64::NAN, |first| t - first)`
turns a programming error — an engine handing the RHS the wrong slice length — into a
numerical result. No compile error, no panic, no warning. That amplifier is the root cause of
#1131, of #1261, and of finding 1.

The cheap deep fix is one line at that site:
`debug_assert!(params.len() >= MAX_PK_PARAMS + 2, ...)`. Costs nothing in release, and
measured above it located all 34 live SS paths in a single `cargo test --lib` run. The
stronger version is a newtype (`ExtParams(&[f64])`) on `solve_ode`'s parameter argument, which
makes passing a bare `PkParams::values` unrepresentable rather than merely detected in debug.

Known instances of the class:
- #1131 — EKF (`ode/ekf.rs`), fixed by this PR
- finding 1 — SS equilibration (`ode/predictions.rs`, twelve sites), unfiled as of this writing

---

## 9. `docs/model-file/diffusion.qmd:157` — conventions — READ

The docs assert the ODE-twin reduction without its preconditions.

"`TAD` is re-anchored at each dose segment by the same rule the two ODE predictors use, so an
SDE fit of a time-varying right-hand side reduces to its ODE twin as the diffusion variance
goes to zero" is stated unconditionally. It is not the same rule, and the reduction has at
least three preconditions the page does not state: it fails for a lagged model (ODE passes
real `dose_lagtimes` from `subject_dose_attrs`, EKF passes zeros — finding 7), for an `SS=1`
dose (finding 6), and for a subject whose records start after `t = 0` (finding 4, where
non-zero diffusion breaks it outright). The measured `OFV = 14.7016` on the page is real but
came from a fixture with none of those features.

Docs-lint checks I ran and which pass: the internal link
`[TAD before the first dose arrives](ode-models.qmd#tad-before-the-first-dose-arrives)`
resolves — `docs/model-file/ode-models.qmd:57` is `### \`TAD\` before the first dose arrives`,
whose pandoc slug is `tad-before-the-first-dose-arrives`. No R4 violation.

---

## 10. `src/ode/ekf.rs:615` — test-coverage — READ

`p_obs > 0.0` discriminates, but it is a floor, not a bound.

It is **not** a redundant gate: `is_finite()` accepts `0.0` and this rejects it, so the two
predicates exclude different inputs and deleting either changes what the test catches. (That
was the specific #1229/#1255 hazard worth ruling out, and it is ruled out.)

But it passes for any positive magnitude, including the phantom-segment inflation of finding 4
(`p ≈ q * t_first` instead of `≈ 0`) and any regression that over-propagates the Riccati step.
`ekf_variance_matches_analytic_linear_sde`, already in the same test module, pins the
linear-SDE variance against its closed form; the fixture here is linear in the state
(`dA/dt = -ke*A*(1 + c*u)`), so the same analytic form applies with the time-varying rate
integrated, giving a two-sided bound for the same effort. Per the repo's measure-the-tolerance
rule the comment should carry the realised `p_obs` and the bound derived from it.

---

## 11. `src/ode/predictions_tests.rs:8340` — test-coverage — READ

An older TAD fixture indexes where production uses `.get()`.

`let tad = t - p[crate::types::MAX_PK_PARAMS + 1];` sits under a comment saying the RHS reads
TAD "exactly as the compiled programs do". It does not: production is
`.get(MAX_PK_PARAMS + 1).copied().filter(is_finite).map_or(f64::NAN, ...)`
(`model_parser.rs:11858-11864`).

On a short `params` slice production returns NaN and keeps running, while this fixture panics
on the index — so it structurally could not have reproduced #1131, and any future engine
handed a short slice fails here as a crash rather than as the silent NaN it actually is. The
**new** fixture this PR adds (`time_reading_rhs`, `ekf.rs:456-467`) copies the production shape
correctly; I diffed it against `model_parser.rs:11844-11856` and it matches. `parse_model_string`
is `pub` and used across Tier-1 tests, so both fixtures could share one compiled RHS.

---

## 12. `src/ode/predictions.rs:5958` — reuse — READ

The event-driven predictor hand-rolls all three helpers this PR extracted, in one function:
`first_dose_time_ed` (`:5660`) is `earliest_dose_time` plus `seed_ext_params`' finite→NaN
mapping; `last_dose_eff_ed` (`:5958-6006`) is `tad_anchor_for` verbatim, same SS `rem_euclid`
fold and same first-arrival fallback; `ext_params_ed` (`:6007-6010`) is `seed_ext_params`.

About 70 lines that could be three calls. `tad_anchor`'s own doc says the split is dangerous
precisely here: the two production ODE predictors are selected per subject on `has_resets()`,
so "a divergence here would make two subjects of the same model and the same data shape behave
differently: one finite, its neighbour NaN". The PR's stated goal is to prevent a seventh
spelling; the largest of the existing six is in the same file and module as the extraction.

---

## 13. `src/stats/likelihood.rs:891` — efficiency — READ

The EKF wrapper integrates the same ODE twice, on the inner loop.

`ode_predictions_ekf_with_diffusion` (`predictions.rs:6203`) runs `ode_predictions` for
`r_obs_vec` (`:6224`) and then `solve_ekf` (`:6238`), whose mean walk is a plain ODE
integration of the same system — `propagate_covariance` returns only `p_mat` and
`kalman_update` never touches `u`. The sole production caller then discards the EKF mean:
`let (_, p_obs) = ...` at `likelihood.rs:892`.

This sits under `individual_nll_into_with_schedule`, the FOCE inner-loop objective closure, so
it runs n_subjects × inner-BFGS evaluations × outer iterations — order 1e5–1e6 integrations
per fit, half thrown away. The pre-existing `TODO: unify EKF ipred with likelihood ipred to
avoid double ODE evaluation` on `likelihood.rs:891` becomes actionable precisely because this
PR's zero-diffusion oracle asserts the equality that makes unification safe. Not free:
`solve_ekf` needs the trajectory for `u_mid`/`fd_jacobian`, so it wants a states-returning
driver.

---

## 14. `src/io/output.rs:1345` — reuse — READ

**This is the one your snippet called "finding 4".** The sdtab TAD fallback reproduces
`api::output_columns::tad_at_time(subj, obs_t, &[])` — the same filter, the same SS
`rem_euclid` branch, the same `NEG_INFINITY` fold and NaN fallback — about 18 duplicated
lines (`io/output.rs:1345-1362`). `tad_at_time`'s own doc (`api/output_columns.rs:43`) already
says "Missing entries — a slice shorter than `subject.doses`, or `&[]` — default to zero lag",
i.e. it was written for this caller. The only blocker is a `pub(crate)` on `tad_at_time`.

Cost of leaving it: the mandatory sdtab `TAD` column and the `[derived]` grid's TAD drift
apart on the next fix to either — the same drift the PR body documents as having already
produced `-2.0`, `NaN` and `+1.0` for one input across three of the six spellings.

The "would subsume finding 1" phrasing in the snippet you saw referred to the *subagent's*
own finding 1 (the EKF `break_times` / `collect_dose_break_times` duplication, which is
finding 3 in this list), not to finding 1 here.

---

## 15. `src/ode/ekf.rs:198` — simplification — READ

`let no_lag: Vec<f64> = vec![0.0; doses.len()]` exists solely because `tad_anchor_for` indexes
`dose_lagtimes[*i]` bare (`predictions.rs:2364`, `:2366`, `:2383`), so the caller carries a
length invariant enforced only by a panic. Twelve lines below, `active_infusions` reads the
same concept as `.get(k).copied().unwrap_or(0.0)`, and `tad_at_time` uses
`.get(d).copied().unwrap_or(0.0)`. Switching `tad_anchor_for` to
`.get(i).copied().unwrap_or(0.0)` lets the EKF pass `&[]`, deletes the allocation and the
invariant, and moves the fourth spelling toward `tad_at_time`'s convention — no behaviour
change for the three existing callers, which all pass full-length vectors. The win is the
deleted invariant, not the time: one small `Vec` against two RK45 integrations is well under
0.1%.

---

# Candidates I checked and REJECTED — do not re-derive these

1. **`seed_ext_params` truncating a longer `pk_params_flat`.** Refuted. The only production
   caller path is `likelihood.rs:891` → `ode_predictions_ekf_with_diffusion(ode, &pk.values, ...)`,
   and `PkParams::values` is `[f64; MAX_PK_PARAMS]` (`types.rs:404`), exactly 128. `copy_n =
   len.min(MAX_PK_PARAMS)` is exact; no truncation, no NaN gap.

2. **`dose_f_bio` needing the extended slice.** Refuted. `DoseAttrMap::resolve_or`
   (`types.rs:709-719`) is bounds-checked (`params.get(slot).copied().unwrap_or(dflt)`) and
   every indexed slot is a PkParams slot < 128. Passing the bare slice there is safe.

3. **The empty-matrix guard breaking `covariance.rs:1097`.** Refuted. That `None` branch is
   inside the `None` arm of `invert_psd_with_floor`, and `invert_psd_with_floor` already has
   its own `if n == 0 { return Some(...) }` guard at `covariance.rs:1266-1273`. A 0x0 matrix
   returns `Some` there and never reaches `:1097`, so the "could not compute eigenvalues ...
   may contain NaN or Inf" message is not mis-emitted for the empty case. (The doc comment's
   "returns None for exactly case (b)" is now imprecise, but nothing acts on it.)

4. **A sibling 0x0 panic left unguarded.** Refuted for the reachable paths.
   `build_non_pd_fallback_proposal` (`covariance.rs:242`) has no empty guard but is only
   reached through the `Some(eigvals)` arm above, which the empty case cannot enter.
   `uncertainty_samples.rs:126`, `sir.rs:267` and `agq.rs:137` operate on the packed-parameter
   covariance or need random effects, so a 0x0 does not reach them for the `n_eta = 0` model.
   The `extract_eigenvalues` empty guard **is** genuinely reachable via `covariance.rs:789`
   → `diagnostic_omega(&params_at)` → `&params_at.omega.matrix`, which is 0x0 when
   `n_eta = 0`. The PR's reachability claim holds.

5. **`t_last = fold(0.0, f64::max)` swallowing a NaN obs time.** Not a divergence. The ODE path
   computes it identically (`predictions.rs:3204-3209`), so this is consistent, not a defect
   introduced or exposed here.

6. **The `(t_end - t_start).abs() < 1e-15` continue at `ekf.rs:329` leaving a stale anchor.**
   Refuted twice over. The write at `:335` sits *after* that continue, so a fired continue
   skips the write and the next segment writes its own — no stale read is possible. And the
   continue looks unreachable anyway: `break_times.dedup_by(|a, b| (*a - *b).abs() < 1e-15)`
   already removes those pairs.

7. **Anchor-write ordering.** Verified correct, not a finding. Write at `ekf.rs:335-336`,
   `solve_ode` at `:362-366`, `propagate_covariance` at `:392-396`. The write precedes both
   consumers. The boundary-observation Kalman update at `:246-266` runs before the write but
   never evaluates the RHS.

8. **CHANGELOG placement.** Correct. `## [Unreleased]` is at line 20, `## [0.3.1]` at 448, the
   nearest preceding `### Fixed` at 412, and the added entries are at 430-444.

9. **Sibling `#[path]` test-module convention.** No violation. There is no `src/ode/ekf_tests.rs`
   and no `src/estimation/covariance_tests.rs`, and the new `extract_eigenvalues` test went into
   `outer_optimizer_tests.rs` where that function's existing tests already live
   (`:384-397`).

10. **Public-API baseline.** No violation. `tad_anchor_for`, `earliest_dose_time` and
    `seed_ext_params` are all `pub(crate)`; `git diff main...24d011d1 -- api/` is empty.

---

# The four points raised in the review request

1. **Ordering of the per-segment anchor write.** Correct — see rejected item 7.

2. **The `no_lag` claim.** Verified independently, four ways — see finding 7. The claim holds;
   what it exposes is the missing warning.

3. **A fourth mutation.** Found and measured — finding 2. Hard-coding the TAFD seed to `0.0`
   leaves `4083 passed; 0 failed`.

4. **Whether `p_obs > 0.0` discriminates.** Yes, and it is not a redundant gate — see finding
   10. But it is a floor, and the two-sided oracle already exists in the same module.

TeunP:
## Response to the review — nine fixed, six filed, two of the review's own claims corrected

Every finding reproduced by measurement before being touched, at `24d011d1`. Fixes in `478f4f0a`, with a docs/comment follow-up in
`c028a053`. `preflight OK`; `cargo test --lib --features ci` **4091 passed / 0 failed** (was
4083 — the eight new tests), `--features ci,survival` **4273 passed / 0 failed**.

### Fixed in this PR

| # | defect | measured before | after |
|---|---|---|---|
| 3 | infusion break not `F`-scaled | `ipred = 0.0` at **every** observation for `F = 0.5` | bit-exact vs `ode_predictions` at `F = 1, 0.9, 0.5`; NONMEM anchor, worst rel **5.24e-11** |
| 4 | `break_times` seeded at literal `0.0` | `p_obs(26) = 6.5000000000` = `q·26`, want `q·2` (**13×**) | `0.5000000000`, error **0.0 exactly** |
| 5 | duplicate in-segment obs times | idx 0 `ipred = 0.0` (the prefill) | both `90.4837418037`, one shared update |
| 2 | TAFD seed unpinned | `seed_ext_params(pk, 0.0)` → **4083 passed, 0 failed** | fixture doses first at `t = 6`; that mutation now dies |
| 6, 7 | `SS` / lagtime dropped silently | `90.48` vs `200.27` explicit train | `W_SDE_STEADY_STATE`, `W_SDE_LAGTIME` |
| 10 | `p_obs > 0.0` is a floor | passes for any positive magnitude | quadrature oracle outside the filter, two-sided at `4e-2` |
| 15 | `no_lag` length invariant | panic-enforced | `.get(i).copied().unwrap_or(0.0)`; caller passes `&[]` |
| 9 | docs assert the twin reduction unconditionally | — | preconditions table + the measured gap numbers |

**Finding 3 is worse than the review states.** The review has the segment `[0, 10]` losing the
infusion. Measured one variable at a time — same dose, same everything, only `F` moved:

```
F=1     t=3  ekf=27.8584047150  ode=27.8584047150   absdiff 0.000000e0
F=0.9   t=3  ekf=0.0000000000   ode=27.8584047150
F=0.5   t=3  ekf=0.0000000000   ode=27.8584047150
```

The window is admitted in **no** segment, so the infusion delivers zero mass everywhere, for any
`F != 1`. Anchored against NONMEM rather than against ferx: `nonmem_anchor/oral_central_inf_advan2_f06`
(ADVAN2 TRANS2, `CL=5, V=50, KA=1, F2=0.6, S2=V`, 100 mg at `RATE=25` into `CMT=2`), written as
the explicit two-state ODE twin so it reaches `solve_ekf`, which the analytical anchor cannot.
Non-degenerate on both sides: `t = 1` is inside the window where `F` has not yet moved the answer,
`t = 3, 6, 10, 16` are past the `F`-shortened end at 2.4 h.

### Two corrections to the review, both from measurement

**Finding 6 overstates this PR's role, and the fix is a warning, not an anchor.** The review says
#1263 turns a loud `NaN` into a silent wrong answer. Half right. Decomposing with the `TAD`
coefficient at **zero**, so the model-time term cannot contribute at all:

```
c=0    t=122   ekf_ss=90.48374180   ekf_train=200.27256428
```

EKF + `SS` is 55% wrong with no `TAD` anywhere in the model. That is #1260, it predates this PR,
and the silence predates it too. This PR enlarges that silent class by one arm; it does not create
it. The `TAD` term itself moves the `SS` answer by 0.3% (90.48 → 90.21).

So special-casing the anchor would be fixing the wrong thing. `W_SDE_STEADY_STATE` covers the
autonomous case that has been silently wrong all along, and `W_SDE_LAGTIME` (finding 7) covers the
other silent drop. That said, finding 6's mechanism **is** real and I found a reachable instance the
review did not have — an `SS=1` infusion at `t=120, II=12, duration=18`, so the window's end break
lands at 138, past a virtual pulse:

```
ss     t_start=138  anchor=132  TAD=6
plain  t_start=138  anchor=120  TAD=18      (identical dose, ss=false)
ipred at t=150:  45.59673475  vs  36.73889884   — 24.1%
```

Same single arrival in both arms. That number is now in `tad_anchor_for`'s doc comment and in
`docs/model-file/diffusion.qmd`.

**Finding 10 is right that `p_obs > 0.0` is a floor, but the two-sided bound cannot be tight, and
the reason is worth recording.** My first attempt set it at `1e-2` and failed. `propagate_covariance`
is **forward Euler at `DT_MAX = 0.5`**: at `t = 1` it computes exactly
`0.04·0.5 + (0.04 − 2·0.05·0.02)·0.5 = 0.039` against a true `0.0380650`, a 2.46% discretisation
error. Realised errors against an exact oracle:

```
c=0.00   2.456e-2  2.093e-2  1.671e-2  1.317e-2    at t = 1/4/8/12
c=0.05   2.526e-2  2.249e-2  1.754e-2  1.171e-2
```

decreasing with `t` — the Euler transient, not the oracle. Bound set at `4e-2` from the measured
2.526e-2. It still bites: the defect it exists to catch (an engine ignoring the model-time term, so
`c = 0.05` returns the `c = 0` curve) is **23.4%** at `t = 12`, six times the bound. The oracle is
Simpson quadrature at 20,000 intervals of `P(t) = Q·e^(−2Φ(t))·∫₀ᵗ e^(2Φ(s)) ds`, computed outside
the filter; its `c = 0` arm is checked against the elementary closed form to `1e-9`, so the
quadrature is pinned rather than trusted. `ekf_variance_matches_analytic_linear_sde` sits at 5% for
the same reason without saying why.

### Mutation matrix — seven mutations, zero survivors, one honest asterisk

| mutation | tests killed |
|---|---|
| `infusion_break` (revert the `F`-scaled break) | `ekf_infusion_under_f_matches_nonmem` only |
| `break_seed` (back to `vec![0.0]`) | `ekf_covariance_starts_at_the_first_record_not_at_zero` only |
| `obs_map_single` (one index per time) | `ekf_duplicate_in_segment_obs_times_share_one_update` only |
| `obs_map_double` (assimilate per record) | same test only |
| `tafd_zero` (the review's own fourth mutation) | `ekf_tafd_anchors_at_the_first_dose_not_at_zero` only |
| `DT_MAX 0.5 → 1.5` | both covariance tests, nothing else |
| `tad_is_tafd` (collapse in the **shared** rule) | the closed-form twin **and** the new TAFD fixture |
| `lag_bare_index` (revert finding 15) | **15 of 16** |

The asterisk is the last row: it is a panic, not a divergence, so it localises nothing by
construction. Recorded as such rather than counted as a clean kill.

`tad_is_tafd` killing two tests is the useful result — the new `t = 6` fixture separates `TAD` from
`TAFD` on the *first* interval as well as after the second dose, so it is an independent second
witness to the M3 collapse rather than a duplicate of the original twin.

### One defect in my own fixes, found by reading rather than by the suite

With a subject that has **no observations**, `t_last` folds from `0.0` while `t_first` is the first
dose, so a `0.0` went back into `break_times` and restored the phantom segment I had just removed.
Nothing is scored for such a subject, so it cost only wasted integration — but the code would have
contradicted the comment three lines above it. `t_last` is now clamped up to `t_first`. Third round
on this PR, third defect found by an adversarial read of my own diff and not by any test.

### Filed, not fixed here

None is a defect in this diff; all are pre-existing.

- **Finding 1** → **#1264**, already filed, and fixed by the T2 PR rather than here. Its
  measurement (34 SS-route tests trip a `debug_assert` at the injection point) is the strongest
  thing in the review.
- **Finding 8** — the `debug_assert` itself. **Cannot land until #1264 is fixed**: the 34 failures
  it produces *are* #1264. Queued to land immediately after the T2 PR, where it becomes the guard
  against a fourth instance of the class.
- **Findings 11 + 12** — combined; same file, same root. The event-driven predictor hand-rolls all
  three extracted helpers, and the older TAD fixture indexes where production uses `.get()` (so it
  structurally could not have reproduced #1131).
- **Finding 14** — the sdtab TAD duplication. Not the pure refactor it looks like: `io/output.rs`
  ignores the lag on both the filter and the SS branch where `tad_at_time` applies it, so lifting
  one onto the other is a **behaviour change** for a lagged model's sdtab `TAD` column and needs its
  own test.
- **Finding 13** — the double ODE integration. **Must not land before #1260 is actually fixed**:
  `IPRED` is correct today precisely because it comes from the ODE path, so unifying now would
  propagate the un-equilibrated EKF mean into it.

### One more thing the follow-up commit settled

The T2 session flagged that `simulate()` and `predict()` never run `check_model_data_warnings`,
so a warning raised there has no reach on those paths. Checked, and it does **not** apply to the
two warnings above — but not by luck. `check_model_data_warnings` does have exactly two callers,
`api/fit.rs:1220` and `validation.rs:4853` (`ferx check`). The difference is one grep further
down: the **sole** production caller of `ode_predictions_ekf_with_diffusion` is
`stats/likelihood.rs:892`. The filter runs only inside the objective, so `predict()` and
`simulate()` take `ode_predictions`, which *does* equilibrate an `SS=1` dose and *does* apply a
lag time. These are fitting gaps and the warnings fire on exactly the surface where the defect
lives.

That consequence is surprising enough to document, and it is how #1131 presented in the first
place: **a `[diffusion]` model can show a correct `IPRED` beside a wrong objective.**
`diffusion.qmd` now says the table is about fitting, not prediction.

### On the `Dual2`-vs-FD parity rule

Still discharged by routing, unchanged by this round: the analytic path declines `SS` × a
model-time RHS through the `uses_time_vars()` gate
(`ode_provider_ss_tad_dependent_rhs_routes_to_fd`, `sens/ode_provider_tests.rs:3710`), and nothing
here widens it. None of the six fixes touches a sensitivity kernel.

Thank you for the review — findings 2, 3, 4 and 5 were all things the suite could not see, and
finding 2 in particular is a fixture-degeneracy I had convinced myself I had ruled out.

PR #1265, opened 2026-09-06
fix(tools): seed a child off a near-singular omega block, and say why every start failed [refs #1256]

roninsightrx:
## Why

Stacked on #1256. Running `ferx modelsearch` on the vancomycin_uvm base (`PERIPHERALS(0..2)`), the three-compartment candidate showed `failed — All multi-start fits failed` after 0.0 s. Not the model: the base fit's (CL, V1, V2) `block_omega` is near-singular through its correlations — every declared variance ordinary, the Cholesky diagonal for `ETA_V2` on the optimizer's rail (6.1e-6) — and the child was seeded with that block verbatim, so the engine refused every start (`E_OMEGA_INIT_AT_RAIL`). The one-compartment candidate got through only because narrowing dropped `ETA_V2` from the block. And the table hid the reason.

## What changed

- `search::seed::seed_from` (shared by covsearch and modelsearch): after the diagonal floor, a block whose Cholesky diagonal is at the rail is nudged by `1e-5·I` until its factor clears it — for a PSD ω every Schur complement of `ω + δI` is at least δ, so one nudge normally does it; bounded at 16 for a fitted ω that is slightly indefinite in the last digits, past which the seed is verbatim and the engine's own message names the block. The block check uses the engine's rail with headroom (`8e-6`), not the diagonal floor, so a block whose diagonals were just floored to `1e-5` with a small covariance (Cholesky diagonal a hair under `1e-5`) is not nudged twice — the existing `seed_from_raises_the_diagonal_only` test caught exactly that.
- `fit()`'s multi-start failure now reads `All multi-start fits failed: start 0: …; start 1: …`, so a runner row says why a candidate never fitted.
- `SeedInits` itself is untouched — it stays faithful to the fit (`final.ferx` = `final-fit.yaml`).

## Alternatives considered

Shrinking the block's correlations instead of nudging the diagonal: also works, but moves the declared correlations by a data-dependent amount; `δ·I` moves each variance by exactly the floor and is what the diagonal path already does.

## Cross-repo dependency

## Breaking changes
- [x] None (the multi-start error string gained a suffix; nothing parses it)

## Numerical validation
- [x] Not applicable — seeding a start and an error message; no estimate changes. The nudge moves a declared variance by 1e-5.
- Reproduced on the vancomycin_uvm run: the three-compartment candidate is refused at the rail before the fix; the resume with the fixed binary refits it (result posted below when the run finishes).

## Performance
- [x] No significant impact expected (one Cholesky per seed)

## Tests
- [x] `seed_from_regularises_a_block_whose_cholesky_diagonal_is_at_the_rail`: a 3×3 block built as `L Lᵀ` with `L₃₃² = rail`, every declared variance ordinary; the written block factors above the rail, each diagonal moved by exactly one δ, and `SeedInits` alone writes the singular block (the mutation).
- [x] `seed_from_leaves_a_well_conditioned_block_alone`: no clone, no nudge.
- [x] `a_fit_whose_every_start_is_refused_says_why` (core): a start at the rail with `n_starts = 2` errors with both starts' reasons.
- Existing seed tests (diagonal floor, diagonal-only, verbatim on a healthy fit) still pass.

## Docs
- [x] `docs/tools/modelsearch.qmd` (the seeding paragraph), `CHANGELOG.md` (two Fixed bullets)

## Checklist
- [x] `tools/preflight.sh` green (run with these changes on the parent branch's tree)
- [x] No `Dual2` path touched

## Reviewer hints
`seed.rs::floor_variances` and `cholesky_above_rail`; the threshold comment explains the `8e-6`.

https://claude.ai/code/session_0193LzWXUgwiRMtFQBUyt9PS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,14 @@ section of the SDLC for the versioning policy).
   to read. No effect on `.ferx` models, the CLI or the R wrapper, none of which constructs it.
 
 ### Fixed
+- **A search child seeded from a parent whose ω block is near-singular can start (#1256).** On
+  the vancomycin base the (CL, V1, V2) block's Cholesky diagonal for `ETA_V2` sat on the
+  optimizer's rail (6e-6) through its correlations, every declared variance being ordinary; the
+  three-compartment candidate seeded from it was refused by every start. The shared seed now
+  nudges such a block by `1e-5·I` until its factor clears the rail, alongside the diagonal floor.
+- **"All multi-start fits failed" now carries each start's reason** (`start 0: …; start 1: …`),
+  so a search table says *why* a candidate never fitted — a refused start is a different repair
+  from a diverged fit.
 - **`ferx modelsearch` never selects the input model (#1181).** The input has a row of its own
   only when a base had to be derived from it — which happens exactly when its structure lies
   outside the space the MFL declares — so it is ranked in the table but excluded from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,9 @@ section of the SDLC for the versioning policy).
   the vancomycin base the (CL, V1, V2) block's Cholesky diagonal for `ETA_V2` sat on the
   optimizer's rail (6e-6) through its correlations, every declared variance being ordinary; the
   three-compartment candidate seeded from it was refused by every start. The shared seed now
-  nudges such a block by `1e-5·I` until its factor clears the rail, alongside the diagonal floor.
+  nudges that block alone by `1e-5·I` until its factor clears the rail, alongside the diagonal
+  floor; a standalone ω beside it and a `FIX`ed block are untouched, and a block that cannot be
+  repaired within the bound goes through verbatim.
 - **"All multi-start fits failed" now carries each start's reason** (`start 0: …; start 1: …`),
   so a search table says *why* a candidate never fitted — a refused start is a different repair
   from a diverged fit.
--- a/crates/ferx-tools/src/search/seed.rs
+++ b/crates/ferx-tools/src/search/seed.rs
@@ -35,29 +35,67 @@ pub(crate) const MIN_SEED_VARIANCE: f64 = 1e-5;
 /// The floored variance is the same no-variability model the parent found —
 /// spelled so the child can move off it.
 pub(crate) fn seed_from(model: &mut ModelText, fit: &FitResult) -> Result<(), String> {
-    let floored = floor_variances(fit);
+    let blocks = free_blocks_of(model, &fit.eta_names);
+    let floored = floor_variances(fit, &blocks);
     model.apply(ModelEdit::SeedInits(floored.as_ref().unwrap_or(fit)))
 }
 
-/// How many times a near-singular ω may be nudged by `MIN_SEED_VARIANCE·I`
-/// before the seed gives up and hands the parent's ω through verbatim (so
-/// the engine's own message names the block).
+/// How many times a near-singular block may be nudged by `MIN_SEED_VARIANCE·I`
+/// before the seed gives up on it and hands the parent's block through
+/// verbatim (so the engine's own message names the block).
 const MAX_NUDGES: usize = 16;
+
+/// The η indices of every `block_omega` the model declares **free** — a
+/// `FIX`ed block is a statement about the model, which `SeedInits` leaves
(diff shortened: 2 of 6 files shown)


roninsightrx:
Reproduced on `ferx-testdata/vancomycin_uvm` (100 subjects, 274 observations, `ABSORPTION(INST); PERIPHERALS(0..2)`, exhaustive, 3 starts). Before: the three-compartment candidate `failed — All multi-start fits failed` at 0.0 s (every start refused at the rail). After, resumed from the same journal so only that candidate was refitted:

```
  model    structure               npar        OFV   bic_mixed        d rank seconds  status   decision
  base     INST, 1 peripheral        12   1131.268    1189.554   +0.000    1     3.0  ok       SELECTED
  run2     INST, 2 peripherals       14   1126.617    1196.129   +6.575    2    14.1  ok
  run1     INST, 0 peripherals        7   1238.403    1272.655  +83.101    3     0.1  ok
```

The three-compartment model fits (OFV −4.65 vs the base for two extra θ), does not earn them on the mixed BIC, and the base is selected — the expected answer on this dataset.

roninsightrx (commented):
Review findings:

1. **[P2] Scope the nudge to the offending block** (`crates/ferx-tools/src/search/seed.rs:75-78`). The loop tests the full Ω and then adds `MIN_SEED_VARIANCE` to every diagonal. In a mixed model with a near-singular block `(A, B)` plus a healthy standalone `omega C ~ 2e-5`, repairing the block rewrites C to `3e-5` even though C was already startable. The same issue occurs when the bad block is `FIX`: core deliberately exempts fixed coordinates from the rail rejection, but this path still uses that block to perturb unrelated free Ωs. That contradicts the documented “such a block is nudged” behavior and can materially move a small healthy start. Please derive the declared non-FIX block submatrices from `model`/eta names and regularize only the failing block, with a mixed block+diagonal regression test.

2. **[P2] Honor the advertised `MAX_NUDGES` fallback** (`crates/ferx-tools/src/search/seed.rs:75-87`). If the matrix is still not above the threshold after 16 iterations, the loop exits because `nudges == MAX_NUDGES`, but `changed` is true and the function returns the 16-times-modified Ω. It neither falls back to the parent Ω as the comment/PR description says nor guarantees a startable seed. Add a post-loop validity check and return the intended fallback (plus an exhaustion test).

Verification completed at head `5070241f`:
- `cargo test -p ferx-tools seed_from`
- `cargo test --lib a_fit_whose_every_start_is_refused_says_why`
- `cargo check --workspace --tests --features ferx-core/ci`
- `tools/update-public-api.sh --check`

All completed successfully (existing warnings only).

roninsightrx:
Both addressed in f51948f9.

**1. Nudge scoped to the offending block — fixed.** `seed_from` now reads the model's `block_omega` declarations (`free_blocks_of`: names → η indices, skipping any block carrying `FIX` and any block naming an η the fit lacks) and runs the check + nudge on each free block's submatrix on its own. Standalone ωs get only the diagonal floor, as before; a `FIX`ed block is never nudged and is never a reason to touch anything else. Tests: `seed_from_nudges_only_the_failing_block_not_a_healthy_standalone_omega` (near-singular `(ETA_CL, ETA_V)` beside `omega ETA_KA ~ 2e-5`: the block's two variances move by exactly one δ, its covariance is verbatim, `ETA_KA` is written `0.00002`) and `seed_from_leaves_a_fixed_block_and_its_neighbours_alone` (an exactly singular `FIX` block: the line is written as declared, the standalone ω verbatim, `free_blocks_of` empty).

**2. `MAX_NUDGES` fallback honoured — fixed.** After the bounded loop the block's factor is checked once more; if it still does not clear the rail the block is restored from the parent's ω verbatim (not sixteen nudges that do not start), so the engine's own message names it. Test `seed_from_hands_an_unrepairable_block_through_verbatim`: an indefinite block (variances 1e-5, covariance 1e-3) is never made PD by `16·δ`; `floor_variances` returns `None` (nothing else was below the floor) and the seeded text equals `SeedInits` alone line for line.

Docs paragraph and changelog bullet updated to say "that block alone", the `FIX`/standalone exemption, and the verbatim fallback. Gates re-run in this worktree: `cargo test -p ferx-tools -p ferx-cli` (413 + 33 + 110 + 26 green, 8 seed tests), docs lint, full `tools/preflight.sh`.

Issue #1258, opened 2026-09-06
epic(bugs): batch T — SS × model-time RHS, measured plan and route (#1139 / #1126 / #1131)


TeunP:
Tracker for batch T of #1217. The plan below was **measured** at `main` `579d1e96` on
2026-09-06 before any code: 11 NONMEM streams (`nonmemdocker:V0.1` / `nmfe76`) and a
throwaway prototype of the proposed fix, since reverted. Four claims in the issues, and four
in my own first draft, did not survive that; they are corrected here rather than carried.

Order is **T1 → T2 → T4**, with **T3** parked on a decision. Nothing starts before its
predecessor's numeric exit condition is met.

### The measured picture

Common model: 1-cpt IV bolus, `CL = 1`, `V = 20`, `AMT = 100`, `II = 12`, `eta = 0`.
Train = 41 explicit doses at `t = 0, 12, …, 480`; SS = one `SS=1, II=12` record at `t = 480`.
Observations 482 / 485 / 488 / 491, strictly inside `[480, 492)` so `MOD(T + 120, 12)` is the
true time-after-dose under both an absolute and a cycle-local clock. Tolerances set in the
**model file** (#1212 trap).

| RHS term | ferx `SS=1` | ferx train | NONMEM `SS=1` | NONMEM train |
|---|---|---|---|---|
| autonomous (control) | 10.027269 | 10.027269 | — | — |
| `0.03*TAD` | **NaN / sentinel** | 8.890201 | 8.8902010334129 | 8.8902010448184 |
| `0.0*TAD` | **NaN / sentinel** | 10.027269 | — | — |
| `0.003*TAFD` | **NaN / sentinel** | 5.122100 | — | — |
| `0.003*T` / `0.003*TIME` | 8.567935 | 5.122100 | 8.5679354654224 | 5.1221001401562 |

Does a periodic steady state exist? (train convergence, N=21 → N=41)

| RHS term | N=21 | N=41 | rel change |
|---|---|---|---|
| `0.03*TAD` | 8.8901979425704 | 8.8902010448184 | **3.5e-7** — converged |
| `0.003*T` | 6.6261086442073 | 5.1221001401562 | **0.294** — no limit |

Lagged (`ALAG1 = 3`), observation at t=481, **inside** the pre-arrival window `[480, 483)`:

| arm | ferx SS | ferx train | NONMEM SS | NONMEM train |
|---|---|---|---|---|
[...]

PR #1270, opened 2026-09-06
fix(ode): anchor TAD per window inside the steady-state run-in [closes #1264] [refs #1139]

### LINKED ISSUE #1264: bug(ode): the SS equilibration hands the compiled [odes] RHS a bare parameter slice — TAD/TAFD unreadable (root cause under #1139)
Found while reviewing #1263 (batch T step T1). Same root cause as #1131 and #1261, third
engine — and this one sits **under** #1139, whose stated mechanism is not the one measured.

## The defect

`params[MAX_PK_PARAMS + 1]` (`parser/model_parser.rs:11851-11864`). The steady-state
equilibration hands it a bare `pk_params_flat` — exactly `MAX_PK_PARAMS` long — so `.get()`
on either slot returns `None` and the RHS injects `NaN`, which multiplies into the
equilibrated state and poisons every prediction downstream of it.

Twelve `solve_ode` call sites in the equilibration family pass the bare slice:
`ode/predictions.rs` `672`, `1174`, `1217`, `1239`, `1251`, `1266`, `1310`, `1324`, `1341`
(`equilibrate_ss_pk_state`), `1454`, `1467`, `1477` (`ss_state_at_phase_pk`).

## Measured — scope

`debug_assert!(params.len() >= MAX_PK_PARAMS + 2)` at the injection site, on `579d1e96`,
then `cargo test --lib`: **35 tests trip it, 34 reporting `handed 128 params, needs 130`,
and every one of the 34 names an SS route** — `ode_provider_ss_*`, `sens::provider::*_ss_*`,
`adaptive_ss_base_dose_matches_static_predict`,
`population_packed_gradient_ode_ss_reset_matches_fd`. (The 35th, `test_ode_block_supports_if_statements`,
hands 2 params and is a parser unit test calling the RHS directly.)

slice length.

## Measured — attribution, one variable

#1139 attributes its `TAD`/`TAFD` NaN rows to `tad_anchor` returning `NaN` "when no effective
prior dose exists". That is not what produces them. `tad_anchor` is never consulted during
equilibration: `apply_segment_boundary` writes the anchor **after** the SS call
(`predictions.rs:6872` vs `:6899`), and the equilibration receives `pk_params_flat`, not
`ext_params`. It also cannot explain the `TAFD` row at all — `TAFD`'s anchor comes from
`seed_ext_params`, never from `tad_anchor`.

On `579d1e96`, #1139's fixture (1-cpt IV bolus, `CL = 1`, `V = 20`, `SS=1`, `II = 12`, dose
record at `t = 120`, obs at 122 / 125 / 128 / 132):

| | `0.0*TAD` | `0.0*TAFD` |
|---|---|---|
| `579d1e96` | `[NaN, NaN, NaN, NaN]` | `[NaN, NaN, NaN, NaN]` |
| + extended slice, **nothing else changed** | `200.5453798421` | `200.5453798421` |

`200.54537984215145 / V` = **`10.027268992107572`**, against #1139's own 40-cycle pulse-train
oracle of **`10.027269`**. So the short slice alone accounts for the whole `0.0*TAD` row, and
removing it puts ferx on the train oracle to 8 significant figures.

The only change between the two rows was a full-length extended array with both anchor slots
finite, shadowed at the top of `equilibrate_ss_pk_state` and `ss_state_at_phase_pk`.

## Why this is separate from #1139, and a precondition for it

Fixing the slice does **not** close #1139: the anchor value in the experiment above was an
arbitrary `0.0`, so the non-zero-coefficient rows (`0.03*TAD`, train `8.890201`) still need a
genuine cycle-local anchor, and the `T`/`TIME` rows (~17% wrong) are a different defect
entirely. But T2 cannot be verified until the slots are readable at all — an anchor the
equilibration computes and cannot read is unobservable, and a green T2 test on today's code
would be measuring the wrong thing.

Note the equilibration integrates on a **cycle-local clock** (`seg_start = m * dose.ii`), so
"pass `ext_params`" is not by itself the fix — what `TAD`/`TAFD` should anchor to on that
clock is T2's question (#1258 already records "three time axes inside the f64 SS
equilibration, not one"). This issue is the narrower, prior one: the slots are unreachable.

## The class

Third instance, same amplifier: `.map_or(f64::NAN, …)` at the injection site turns a wrong
slice length into a numerical result — no compile error, no panic, no warning.

- #1131 — EKF (`ode/ekf.rs`), fixed in #1263
- this one — SS equilibration (`ode/predictions.rs`, twelve sites)

One `debug_assert!(params.len() >= MAX_PK_PARAMS + 2)` at `model_parser.rs:11851` costs
nothing in release and located all 34 live paths in a single `--lib` run; a newtype on
`solve_ode`'s parameter argument would make the bare `PkParams::values` unrepresentable
rather than merely detected in debug. Worth deciding once rather than a fourth time.

## Test note

An `ipred`/`is_finite` assertion is the right one here (the NaN reaches predictions
directly, unlike #1261's `1e20` sentinel), but the discriminator is the **zero-coefficient**
case: `0.0 * NaN` is `NaN`, so merely mentioning a builtin breaks an otherwise ordinary SS
model, and the correct answer is bit-comparable against the autonomous control.

## Scope: land the guard in the same PR

Whoever fixes the twelve sites should also add, at `parser/model_parser.rs:11851` immediately
before the TAFD injection:

```rust
debug_assert!(
    "compiled [odes] RHS handed {} params, needs {}",
    params.len(),
    crate::types::MAX_PK_PARAMS + 2
);
```

Not a separate task — it is what *proves* the fix is complete. Twelve call sites is enough
that "did I get them all?" is a real question, and this answers it mechanically: the same
assert is what located all 34 live paths above, in one `cargo test --lib` run. It costs
nothing in release.

Caveat worth stating rather than discovering later: `debug_assert!` fires only in debug and
test builds, so a release-only path handing a short slice would still go silently `NaN`. The
34-test result suggests the SS routes are well covered in debug, so this does not look like a
live gap — but it is the reason the assert is a completeness check for this fix, not a
permanent guarantee for the class.

The stronger form — a newtype (`ExtParams(&[f64])`) on `solve_ode`'s parameter argument, making
a bare `PkParams::values` unrepresentable — is deliberately **not** proposed here. It touches
every `solve_ode` caller in the tree for a defect the assert already catches at test time.
Worth revisiting only if a fourth instance of the class appears (#1131, #1261, this one).

TeunP:
> **No longer stacked.** T1 merged as `7eb19818` (#1263) and this branch has been rebased onto
> it, so the base is now `main` and the diff is T2's work alone. It was opened against
> `worktree-t1-ekf-anchor-slots`, which meant **zero CI checks ran on it** until the retarget —
> `pull_request` workflows filter on `branches: [main]` (#1271). Every gate below was therefore
> run locally, and re-run after the rebase brought in #1256's 13,514 lines.

## Why

**Closes #1264; advances #1139 but does not close it.** This is #1139's `TAD` half only —
its `TAFD` / `T` / `TIME` half is a separate decision (T3 in #1258) and #1139 stays open for
it. #1264 is the same root cause, filed
independently while this was being written; folding rather than splitting was agreed with the
session that filed it, on three grounds: a slice-only PR could carry **no regression test**
(the slot it opens is unobservable until an anchor is written into it, so it would be the one
batch-T step with no numeric exit condition of its own); #1258's route rule is "nothing starts
before its predecessor's numeric exit condition is met"; and the fact that **three of the twelve
sites need a *non-zero* anchor**, not a uniform one, only exists in the combined diff — a reviewer of a bare slice
extension has no way to see it.

**#1139.** An `SS=1` dose on an `[odes]` model whose right-hand side reads `TAD` returned
`NaN` at every observation and a non-finite objective. The compiled RHS reads `TAD` out of
`params[MAX_PK_PARAMS + 1]` (`parser/model_parser.rs`, the model-time closure); every
steady-state call site handed `solve_ode` a bare `PkParams::values`, which is exactly
`MAX_PK_PARAMS` long, so `.get()` returned `None` and the RHS injected `NaN`. Because
`0.0 * NaN` is `NaN`, merely *mentioning* `TAD` broke an otherwise ordinary steady-state
model — the case the issue opens with.

This is step **T2** of the batch-T plan in #1258, and it is `TAD` only. `TAFD`, `T` and
`TIME` under `SS=1` are the same issue's other half and are deliberately untouched here:
their dose train has no periodic limit for a run-in to converge to (measured: 29 % per
doubling of the train's length, against 3.5e-7 for `TAD`. One pair of streams certifies both
absolute-clock spellings, because this dataset's first dose is at `t = 0`, so `TAFD ≡ T` on
it — verified, ferx returns the identical `5.1221001380354` for both), and
`T`/`TIME` already matches NONMEM. Anchoring those at the run-in's own origin would return a
finite, plausible number that is really `TAD` under another name — the defect class this
change removes.

## What changed

**The run-in does not integrate on the subject's clock.** It expands a periodic train on a
clock local to each window, so the anchor is that window's own pulse time — not a
`tad_anchor_for` question, and not the walk's `ext_params`. One private helper,
`ss_run_in_params(pk_params_flat, pulse_at)`, names the convention once and is threaded
through **nine** `solve_ode` windows in `ode/predictions.rs` (of the twelve the SS family
has; the other three are `ss_state_at_phase_pk`'s, deliberately left — see below).

**All twelve sites, and what each gets.** Line numbers are `ode/predictions.rs` at `579d1e96`,
matching #1264's enumeration.

| | sites | pulse sits at local | anchor |
|---|---|---|---|
| **A — six windows opening at the pulse** | `672` input-rate exact advance · `1217` propagator probe · `1239` exact infusion active · `1266` exact forced bolus · `1310` capped infusion active · `1341` capped bolus | `0` | `0.0` |
| **B — two quiet windows** an infusion cycle re-opens after `T_inf` of rate | `1251` exact · `1324` capped | `−T_inf` | `−T_inf` |
| **C — one monotone window**, the only clock that does not restart per cycle | `1174` capped input-rate train | `m·II` | `tad_anchor_for` over that train's own dose list |
| **D — three sites deliberately left on the bare slice** | `1454`, `1467`, `1477` (`ss_state_at_phase_pk`) | — | **none — still `NaN`** |

**Group D is the part to read twice.** Those three are *not* fixed, and that is deliberate,
not an oversight. Reachable only under a lagtime (`ss_seeded_at_record` requires `lag > 0`),
which is #1126's unanswered pre-arrival referent — so making them readable would replace a
loud `NaN` with a plausible number 2.733 % wrong on **every** observation, measured. They stay
`NaN`, and the combination that reaches them is rejected outright (below). A reader who sees
"#1264 fixed" and finds three of its twelve sites untouched should stop here rather than
re-derive that measurement.

`TAFD` is left `NaN`, which reproduces today's answer bit-for-bit for a `TAFD`-reading
model. Building the array **inside** the helpers rather than passing it down also sidesteps
`apply_segment_boundary`'s write-after-solve ordering entirely (it writes its TAD anchor
*after* the two SS calls, so a naive `ext_params` swap would read the previous segment's
anchor and still compile) — and it is correct precisely *because* the run-in's clock is not
the walk's.

**`E_SS_LAGTIME_TAD_RHS`** (`api/validation.rs`, `check_model_data_rule`). `SS=1` plus a
lagtime on a `TAD`-reading RHS is rejected rather than served — by `fit()` and `ferx check`;
`simulate()` and `predict()` validate a narrower bundle and still return the `NaN` they did
before, which the docs now say rather than claiming a blanket rejection. NONMEM does not settle this
one either: measured from its own tables, its `SS=1` record with `ALAG1 = 3` reads
`5.4709357327409` at `t = 481` against `5.5452941786123` from its own explicit lagged train
— **1.34 %** apart, where the *un*-lagged pair agrees to 1.28e-9. So there is no external
reference to anchor the combination to today, and it is named rather than guessed at. The pre-arrival window
`[t_dose, t_dose + ALAG)` has no `TAD` referent (#1126), and since #1121 the record-time
seed *flows* to the arrival rather than being re-equilibrated there, so the wrong anchor is
carried into the trough and multiplies into the **whole subject** — measured 2.733 % high at
every post-arrival observation, 3.67 % inside the window, against a closed form computed
outside the engine. That is why `ss_state_at_phase_pk`'s three windows are deliberately
*not* anchored here and the asymmetry is recorded at the call site: fixing them alone turns
a loud `NaN` into a plausible 2.7 %-wrong number. #1126 fixes the referent and lifts the
gate; both halves belong in that change, where a mutation can show neither alone suffices.

**A test-only branch tag** (`dosing::SsBranch`, `#[cfg(test)]` recorder, `#[inline(always)]`
no-op in release, alongside the existing cycle counter). The SS branches **fall back to one
another** — exact linear fixed point → Anderson → capped train — so breaking one silently
routes to a correct one and a value assertion stays green. That is measured, not
hypothetical: two of the first mutation round's ten survived for exactly this reason. Every
test that means to pin a window's clock now asserts which branch ran.

Two properties of the tag are load-bearing and easy to get wrong, so both are pinned. It is
cleared at the top of `equilibrate_ss_pk_state`, so a call that bails out (`II <= 0`, an
out-of-range compartment, an overlapping infusion) reports `None` instead of inheriting the
previous call's value — otherwise an assertion after a bail-out reads a stale tag and passes
for the wrong reason. And `Exact` / `InputRateExact` are separate variants, because the
bolus/infusion closed form and the input-rate one are different functions with different
windows and a test naming one should not pass on the other. The tag is one thread-local
shared with the **dual** walk (`sens/ode_provider.rs` instantiates the same generic
helpers), which is documented on the enum: assert it only from a test that calls the f64
helpers directly.

## Alternatives considered

- **Passing the walk's `ext_params` into the SS helpers.** Rejected: the walk's anchor is on
  the subject's clock and the run-in's is not, and it would inherit
  `apply_segment_boundary`'s ordering trap.
- **Anchoring `TAFD` at the run-in's origin too.** Rejected on measurement (#1258 finding 4)
  — a finite, plausible number for a quantity whose train diverges is the same class of
  silent wrong answer this PR removes.
- **Fixing `ss_state_at_phase_pk` and warning instead of rejecting.** Rejected after
  measuring the magnitude: a 2.7 % error on every observation behind a warning is worse than
  the `NaN` it replaces, and #1139 itself proposes rejection for exactly this situation.

## Cross-repo dependency

No public-API change.

## Breaking changes
- [x] None

A model that was previously rejected by nothing and returned `NaN` (`SS=1` + lagtime +
`TAD`) is now rejected with a named error. It could not produce a usable fit before.

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0** (`nonmem_anchor/ss_tad.ctl`, `ADVAN13 TOL=9`,
      `MAXEVAL=0 METHOD=0 POSTHOC`, `$THETA` FIX, `$OMEGA 0 FIX`), **and** a closed form
      computed outside both engines

Six control streams ship with the PR (`ss_tad`, `train_tad`, `train_tad_n21`, `ss_tabs`,
`train_tabs`, `train_tabs_n21`). NONMEM has no `TAD` built-in in `$DES`, so the cycle clock
is written by hand as `TADX = MOD(T + 120.0, 12.0)` — exactly `TAD` for this dataset, and
certified rather than assumed (see below).

```
# 1-cpt IV bolus, CL = 1, V = 20, AMT = 100, II = 12, eta = 0, ode_reltol = 1e-9
# one SS=1 record at t = 480; observations at 482 / 485 / 488 / 491

                             t=482              vs NONMEM SS
before   ferx SS  0.03*TAD   NaN                 —
after    ferx SS  0.03*TAD   8.8902009676679     7.40e-9
         NONMEM SS           8.8902010334129     —
         ferx train          8.8902009676645     9.07e-9   (unchanged by this PR)
         NONMEM train        8.8902010448184     —
         closed form         8.8902009650221     ferx 2.98e-10 · NONMEM 8.33e-9

# discriminator: 0.0*TAD
before   ferx SS             NaN
after    ferx SS             10.0272689918704    == the autonomous SS to 2.1e-11

# scope: unchanged, before and after
         ferx SS  0.003*T    8.5679354785102     bit-identical; NONMEM 1.53e-9
         ferx SS  0.003*TAFD NaN                 unchanged by design
```

`simulate()` reaches the same fix — verified, it returns `8.890200967667928` on this model,
bit-identical to the objective path, so it is a *caller* of the event-driven engine rather
than a fourth geometry and gets no separate assertion.

Three statements make the anchor trustworthy, and none of them is ferx-against-ferx:

1. **A periodic steady state exists** for `TAD` — `train_tad_n21` vs `train_tad` moves
   3.5e-7 over a doubling — and does **not** for absolute time (`train_tabs*`, 0.294).
2. **NONMEM's own SS routine reproduces NONMEM's own 41-dose train** to 1.28e-9.
3. **A closed form outside both engines** agrees with both. It is the *tighter* reference:
   ferx sits 3.0e-10 from it while NONMEM sits 8.3e-9, so NONMEM's uniform ~8e-9 offset is
   its `TOL=9` and `1PE20.13` print floor, not ferx drift. Every bound in the anchor suite
   is set from a measured error with the headroom stated, and the suite asserts that ordering
   so a future relaxation has to re-measure rather than loosen.

## Performance
- [x] No significant impact expected. One `[f64; MAX_PK_PARAMS + 2]` stack array (≈ 1 KB) per
      steady-state equilibration call for the cycle anchor, **plus one per cycle** on each of
      the two branches whose anchor advances — the capped input-rate train (up to
      `SS_EQUILIBRATION_CYCLES = 50`, ≈ 52 KB of memcpy) and an infusion's quiet window. That
      is against 50 ODE solves on the same branch, each thousands of RHS evaluations, so it is
      not measurable.

      The quiet window's array is built **inside** each infusion arm rather than hoisted, which
      a review pass changed: hoisted, it had to be an `Option` whose `expect` asserted an
      invariant enforced by two `if is_inf` guards forty and a hundred lines away. Structurally
      safe as written — verified — but the failure mode if it ever drifted is a panic inside a
      fit, and a panic crosses ferx-r's extendr boundary as a process abort rather than an R
      error. Binding it where it is used costs one array per infusion cycle and removes the
      invariant instead of documenting it.

      I deliberately did **not** hoist that array out of the loop and mutate the one slot.
      It would save nothing detectable, and it would re-create the write-the-anchor-before-
      the-solve ordering hazard that `apply_segment_boundary` already demonstrates and that
      building the array inside the helper exists to avoid. Cheap and obviously-ordered beat
      free and subtly-ordered here.

      The branch tag is `#[cfg(test)]`; the release build gets an `#[inline(always)]` empty
      function taking a fieldless enum.

## Tests

**Tier 1** (`src/ode/predictions_tests.rs`), one per window family, each with an oracle
outside the engine and an asserted branch:

| test | window | oracle | realised error |
|---|---|---|---|
| `ss_run_in_reads_the_cycle_local_tad_on_the_exact_solve` | exact linear solve | closed-form fixed point of `A ↦ (A+D)·Φ(II)` | 4.69e-12 |
| `ss_run_in_reads_the_cycle_local_tad_on_the_capped_train` | capped bolus train | closed form of the nonlinear (`A²`) one-cycle map | 3.56e-12 |
| `ss_infusion_run_in_keeps_the_pulse_clock_across_the_quiet_window` | exact infusion active + quiet | composite-Simpson quadrature of `∫μ` (not an ODE solve) | 3.98e-12 |
| `ss_infusion_capped_train_keeps_the_pulse_clock` | capped infusion active + quiet | converged explicit infusion train | 1.69e-12 |
| `ss_into_absorption_exact_solve_reads_the_cycle_local_tad` | input-rate exact advance | converged explicit train | 1.05e-13 |
| `ss_into_absorption_monotone_train_advances_the_tad_anchor_each_cycle` | monotone train | explicit 51-dose train at the forced full budget | 6.75e-16 |
| `a_zero_coefficient_tad_term_leaves_the_ss_run_in_unchanged` | plumbing | `to_bits()` vs the plain RHS, plus the autonomous closed form | exact |
| `tafd_stays_nan_inside_an_ss_run_in` | scope pin | — | — |
| `a_bailed_out_ss_equilibration_reports_no_branch` | the branch tag itself | — | — |
| `ss_monotone_run_in_anchor_is_the_same_with_an_empty_lag_slice` | the `&[]` lag contract | `to_bits()` at every segment boundary | exact |

**Tier 2**: `tests/ss_model_time_nonmem_anchor.rs` (10) and `tests/ss_lagtime_tad_rhs_gate.rs`
(7), the latter's cases each removing exactly one conjunct of the new gate.

Two of those ten are worth calling out, because both were **wrong when first written** and
the errors are the kind this repo has a rule about:

- **The "states engine" test was a duplicate of the objective test.** I wrote it as
  `compute_predictions_with_states(..).0`, believing #1124's reroute put the two callers on
  different walkers. It does not:
  `ode_predictions_event_driven_with_states` returns `ipreds` straight from
  `ode_predictions_event_driven` and fills only its *second* tuple element from
  `ode_dense_solve_states`. Verified rather than argued — `.0` is **bit-identical** to the
  objective path at all four times. That is exactly "callers routing to the same engine are
  callers, **not** independent geometries". The test now reads `.1`, scales the state amount
  by `V` by hand (`apply_scaling` never touches the state vector), and `.1/V` is
  bit-*different* from the objective path at three of the four times — which is the evidence
  it is now a second integration, and an assertion that would notice if it stopped being one.
- **Nothing tested #1139's actual reported symptom.** Every predictor-level test would stay
  green if the likelihood assembly below it broke, and the issue is about a non-finite
  **objective**. Added an end-to-end `fit()` check against NONMEM's `#OBJV = 286.890`:
  ferx `286.889524669658` (SS) and `286.889524669616` (train), both `1.657e-6` from NONMEM's
  six-figure print, and `1.474e-13` from **each other** — five orders tighter than either is
  from NONMEM, which is what makes "the `SS=1` record stands in for the train" a measurement
  rather than a print-bucket coincidence.

### Mutation testing, each mutation naming its own side

Ten mutations, one per `solve_ode` window the fix touches plus the `TAFD` scope decision.
**All ten die; none survives.** Each names its own window — `Mc` and `Md` land on the same
test through *different* assertions inside it (`Mc` makes the exact solve decline, so the
branch assertion fires; `Md` leaves it finite-but-wrong, so the quadrature check does).

| mutation | window | tests that die |
|---|---|---|
| `Ma` | exact solve, propagator probe `(0, II)` → bare slice | `ss_run_in_reads_the_cycle_local_tad_on_the_exact_solve`<br>`ss_infusion_run_in_keeps_the_pulse_clock_across_the_quiet_window`<br>`a_zero_coefficient_tad_term_leaves_the_ss_run_in_unchanged` |
| `Mb` | exact solve, forced bolus cycle `(0, II)` → bare slice | `ss_run_in_reads_the_cycle_local_tad_on_the_exact_solve`<br>`a_zero_coefficient_tad_term_leaves_the_ss_run_in_unchanged` |
| `Mc` | exact solve, infusion active `(0, T_inf)` → bare slice | `ss_infusion_run_in_keeps_the_pulse_clock_across_the_quiet_window` |
| `Md` | exact solve, infusion quiet: anchor `−T_inf` → `0` | `ss_infusion_run_in_keeps_the_pulse_clock_across_the_quiet_window` |
| `Me` | capped train, bolus cycle `(0, II)` → bare slice | `ss_run_in_reads_the_cycle_local_tad_on_the_capped_train` |
| `Mf` | capped train, infusion active → bare slice | `ss_infusion_capped_train_keeps_the_pulse_clock` |
| `Mg` | capped train, infusion quiet: anchor `−T_inf` → `0` | `ss_infusion_capped_train_keeps_the_pulse_clock` |
| `Mh` | input-rate exact one-cycle advance → `NaN` anchor | `ss_into_absorption_exact_solve_reads_the_cycle_local_tad` |
| `Mi` | monotone train: anchor `m·II` → flat `0` | `ss_into_absorption_monotone_train_advances_the_tad_anchor_each_cycle` |
| `Mj` | `TAFD` anchored at the run-in origin instead of `NaN` | `tafd_stays_nan_inside_an_ss_run_in` |

**The earlier rounds are the part worth reading.** The table above is round three. Round one
had **two survivors** — `Mh` and `Mi` — with the entire suite green, and a partial fix left a
**third** (`Mc`) alive. All three for one reason: the steady-state branches fall back to one
another (exact affine fixed point → Anderson → capped train), so breaking one routes silently
into a correct one and a value assertion cannot tell. The input-rate fixture was also landing
on **Anderson** (6 cycles, measured) while its name said "monotone train", so `Mi` sat on code
the test never executed.

Two consequences, both in the diff:

- `dosing::SsBranch` — a test-only branch tag beside the existing cycle counter. A cycle count
  is only a proxy: Anderson records `iter + 1`, which is `1`, exactly what the exact solve
  records. Every test that pins a window's clock now asserts which branch ran.
- The monotone arm needed rebuilding. The branch is reached only when `ρ ≥ 1` — *no periodic
  steady state exists*, which is what `#867` warns about — so its output is a truncated
  accumulation and "compare against a converged train" is meaningless, since the reference
  does not converge either. It is pinned against an explicit **51-dose** train
  (`SS_EQUILIBRATION_CYCLES` run-in pulses plus the record's own) under
  `with_full_ss_equilibration`, which it reproduces to **6.7e-16**, with the neighbouring
  lengths 1.5 % away so the comparison is sharp in the length dimension too.

None of this was visible from reading the code, and the suite was green each time.

## Docs
- [x] `docs/model-file/steady-state.qmd` — a new *Steady state with a `TAD`-reading RHS*
      section with the measured table, and a rewritten FD-gate bullet whose stated *reason*
      was invalidated by this change (edited, not extended). The *Limitations* section's
      skip-with-warning framing now carries the rejection separately, since it is not a skip.
- [x] `docs/model-file/ode-models.qmd` — the non-autonomous-RHS section points at the
      steady-state page for the `SS=1` case.
- [x] `nonmem_anchor/README.md` row for the new streams
- [x] `docs/file-formats/check-report.qmd` — the new code, and the **"Data and dosing"
      section split** into *Records and observations* / *Dose records*. That section had hit
      docs-lint's 5,000-character R1 ceiling three times during this change; splitting is
      what CLAUDE.md prescribes ("split the section instead"), and shaving the text a third
      time would only have moved the problem to whoever adds the next code. The parent keeps
      its `{#codes-data-and-dosing}` anchor, so the page's own contents link still resolves.
- [x] `cargo run -p docs-lint -- --check` green

Four in-tree copies of the "SS × model time is wrong on both engines" note (`sens/ode_provider.rs`,
`sens/ode_provider_tests.rs`, `pk/mod.rs`, `pk/modified_release.rs`) are updated: the `TAD`
rows were the thing this PR changed, and #1258 flagged that the measured table lived only in
a doc comment on a *routing* test — where it could not notice its own values changing. It is
now assertions, in the two suites above.

## Checklist
- [x] `Dual2` sensitivity path — see below

### The `Dual2`-vs-FD parity rule, discharged explicitly

CLAUDE.md requires a `Dual2`-vs-FD parity test for any change to a sensitivity path. **No
such test can reach this change**, and that is stated rather than quietly omitted: the
analytic/dual route is gated to FD for exactly this model class (`ode_tvcov_supported`,
`has_ss && rhs_program.reads_model_time()`), so a `TAD`-reading steady-state subject never
builds a jet. The rule is discharged by **routing** tests instead, and the existing
`ode_provider_ss_tad_dependent_rhs_routes_to_fd` is asserted to still decline after the fix.

The gate is deliberately **not** widened, and reading the dual side says why. Verified in
this session, not recalled: `equilibrate_ss_state_g` builds one `bare_rhs` as
`eval_rhs_anchored::<T>(…, t, 0.0, T::from_f64(0.0), …)` and runs *every* window through it —
the propagator probe, the bolus cycle, and an infusion's active **and quiet** windows alike.
So its recurrence differs from the value path's in two ways now: it does not re-anchor the
quiet window at `−T_inf`, and it makes `TAFD` cycle-local where the value path leaves it
`NaN`. #1258 finding 4 records the `TAFD` half as a latent defect; the quiet-window half is
the same shape and is recorded in the same comment. Both are **unreachable while this gate
holds**, so neither is filed as a bug — there is no reproducer to run — and neither is fixed
here. Nothing in this PR changes which predicate the gate keys on. #626 (analytic CHZ
η-sensitivities) and T3 are where this becomes live.

## Self-review

**Two `xhigh` rounds, both in comments below.**

Round 1 ran on the original slice and found **10 findings**; 8 are fixed here, 1 is
deliberately not (the `CMT=0` scoping, inherited from `E_ABSORPTION_SS_LAG` — diverging from
the neighbour is worse than the over-rejection), and the tenth was the `simulate()` overclaim
above. Two of them were about claims *this description* made.

Round 2 ran on the **whole current diff**, because round 1's own fixes, the
`Exact`/`InputRateExact` branch split, three new tests and the rebase onto #1256 had never
been reviewed. **8 findings, all 8 fixed.** Six were tests that would pass while covering less
than their names claimed — including the closed-form oracle, the tightest reference in the
suite, whose constants were literals with nothing tying them to the committed twin. One
mutation in that round **refuted my own stated justification** for a test arm I had just
added, and the comment now says what the arm actually covers instead.

Neither round disputes a number: no anchor value, NONMEM comparison, closed form or mutation
result changed across either.

## Reviewer hints

- The interesting part is the **mutation table and the branch tag**, not the nine-line
  anchor threading. Two mutations survived a green suite and the reason was structural.
- `ss_state_at_phase_pk` is left on the bare slice **on purpose**; the comment at the top of
  it carries the measurement. If that reads wrong, the alternative is shipping a 2.7 % error.
- The `E_SS_LAGTIME_TAD_RHS` gate is **structural** (is a lagtime *declared*), not
  typical-value, because an estimated `ALAG` moves during a fit. That makes it reject an
  inert `0.0*TAD` under `SS=1` + a lagtime; `tests/ss_lagtime_tad_rhs_gate.rs` records that
  consequence rather than working around it.
- Bounds: every one is set from a printed realised error with the headroom stated in the
  comment. The two 5e-8 bounds in the NONMEM suite are NONMEM's accuracy floor, not ferx's —
  the closed-form test asserts that ordering so it cannot silently invert.

## Open questions

None blocking. T3 (`TAFD` / `T` / `TIME`) and T4 (#1126) follow, in that order.


TeunP:
## Self-review — `xhigh`, 10 findings, 8 fixed in-branch

Ran an adversarial review over this branch's own diff (`24d011d1..HEAD`) before asking for
eyes on it. Recording it here rather than in a chat log, because two of the findings are
about claims **this PR description makes**, and a reader deserves to see them corrected
rather than quietly edited.

**Nothing found disputes a number.** No finding touches an anchor value, a NONMEM
comparison, the closed forms, or the mutation matrix. They cluster in three places: the
test-observability tag this PR introduces, two overclaims, and cleanup.

### The three that mattered

**1 — the `SsBranch` tag is shared with the dual path.** `sens/ode_provider.rs:4267`
instantiates `equilibrate_ss_input_rate_g::<T, _, _>` with a dual `T`, and that function
records the tag; so does the generic `anderson_ss_fixed_point_g`. A test driving a `fit()`
observes whichever walk ran *last*, f64 or dual — not necessarily the value-path run-in it
means to pin. Today's tests call the f64 helpers directly and are safe, but the tag is the
thing this PR leans on to claim its mutation matrix means something, so an unstated caveat
there is the wrong place to be relaxed. **Fixed**: documented on the enum, with the
condition under which an assertion on it is valid.

**2 — `SsBranch::None`'s doc described a state nothing ever set.** It claimed the value
also meant "the last equilibration bailed out early", but the three early returns in
`equilibrate_ss_pk_state` left the *previous* call's tag. A test asserting a branch after a
bail-out would have read a stale value and passed for the wrong reason — the exact
un-failable-assertion class the tag was added to close, reappearing inside the tag.
**Fixed**: the tag is cleared at the top of `equilibrate_ss_pk_state`, so every bail-out
reports `None` honestly, and `a_bailed_out_ss_equilibration_reports_no_branch` pins it. That
test runs a *completing* call first, so it cannot pass on the `Default`.

**3 — `simulate()` never runs the new gate.** `simulate_with_options` validates a narrower
bundle (`validate_tte_simulatable`, `validate_iov_simulatable`, the `assert_*_supported`
set) and does not call `check_model_data_rule`. The docs said the combination "is rejected
outright"; that is true of `fit()` and `ferx check` only. **Fixed by scoping the claim**, in
`steady-state.qmd`, `check-report.qmd` and the changelog, rather than by widening the gate —
adding a simulate-side reject is its own change, and #1258 records that simulate reaches
rejections only through panicking asserts, which cross the ferx-r boundary as a panic rather
than an R error.

### Also fixed

- The `E_SS_LAGTIME_TAD_RHS` message quoted **2.7 %**, which is a property of one fixture
  (`c = 0.03`, `ALAG = 3`, `II = 12`), not of the rejected configuration. A user reading
  `0.3*TAD` would see a number an order of magnitude from their own. The magnitude stays in
  the doc comment and `steady-state.qmd`, where the model it was measured on is stated.
- `ss_run_in_params` turned a short parameter slice from an index panic into a silent `NaN`
  `CL`, because `seed_ext_params` NaN-fills. No production caller does that — all pass
  `&pk.values` — but converting loud to silent is what this whole change exists to reverse.
  A `debug_assert!` keeps the contract loud.
- `SsBranch::Exact` conflated the input-rate closed form with the bolus/infusion one; the
  absorption test asserting `Exact` worked only because that fixture's `input_rate` made the
  other branch unreachable. Split into `Exact` and `InputRateExact`, and the test now names
  the one it means.
- `ext_quiet` was built unconditionally, so a bolus allocated a 130-element array anchored
  at `-0.0` that nothing reads. Now `is_inf.then(..)`, which also states the invariant.
- The `ss_run_in_params` doc enumerated the window shapes but omitted the infusion **active**
  window, though it takes the same anchor — an incomplete enumeration in the one place the
  convention is written down.
- The `local_no_lag` comment justified itself by `tad_anchor_for`'s *implementation* (it
  indexed, so an empty slice panicked). That follow-up has since landed on the base branch —
  `tad_anchor_for` now reads `dose_lagtimes.get(i).copied().unwrap_or(0.0)` — so on rebase
  the allocation is **gone** rather than re-commented: the monotone train passes the existing
  empty `&no_lag`, and `ss_monotone_run_in_anchor_is_the_same_with_an_empty_lag_slice` pins
  that `&[]` and a full-length zero vector give bit-identical anchors at every segment
  boundary, *and* that the anchor is `m·II`, so it cannot pass on two equally wrong answers.

*(This PR was rebased from `24d011d1` onto `c028a053` after the review was written; the
finding list above is unchanged by it, and this bullet is the only line the rebase made
stale.)*

### Not fixed, on purpose

**`CMT=0` scoping.** The gate asks `has_lagtime_on_cmt(d.cmt_1based())` while runtime
reachability resolves through `dose_attr_map.lagtime(d.cmt_raw(), ..)`; for `CMT=0` those
disagree, so a model declaring `ALAG1` is rejected though its runtime lag is the bare slot.
This is inherited from the neighbouring `E_ABSORPTION_SS_LAG`'s existing scoping — fixing it
here would make the two gates disagree about the same question, which is worse than the
over-rejection. Flagged rather than diverged from.

### Verification of the fixes themselves

Two of them are guards, so they were mutation-checked rather than merely run:

| mutation | expected | measured |
|---|---|---|
| collapse `InputRateExact` back onto `Exact` | only the absorption exact-solve test dies | `ss_into_absorption_exact_solve_reads_the_cycle_local_tad` **FAILED**, 8 passed — exactly one |
| drop the up-front `None` clear | only the bail-out test dies | `a_bailed_out_ss_equilibration_reports_no_branch` **FAILED**, 8 passed — exactly one |

**The harness itself had a defect, which is worth more than either mutation.** Applying the
first one created a *second* `SsBranch::Exact`, so the revert — which asserted a unique match
before reverse-replacing — refused and exited, **after** correctly printing the kill. The
tree was left mutated and the script's own output said nothing was wrong. Caught only by
grepping the working tree instead of trusting `revert: ok`. Restoring now goes through a
byte-copy snapshot plus a `touch`, the `touch` because a restored file with an older mtime
lets cargo reuse the *mutated* rlib.

Same lesson one level up: my first residue check used `grep -E "SsBranch::None);"`, whose
unescaped parens error out and produce an empty result that reads as "missing". And a count
cannot distinguish the two `Exact`-shaped sites at all — only reading the surrounding lines
shows one returns `Some(u_ss)` (the generic input-rate function) and the other `u_ss` (the
bolus/infusion closed form), which *was* the finding. Verified by reading, not counting.

The `debug_assert!` is deliberately untested: it is a documented invariant in the same style
as the existing `debug_assert!(next_slot < MAX_PK_PARAMS, ..)` in this file, and a
`#[should_panic]` test on it would pass in debug and fail in release.

TeunP:
## Self-review round 2 — `xhigh` over `origin/main..HEAD`, 8 findings, all 8 fixed

The first review ran on `24d011d1..HEAD`. Since then this branch gained the eight fixes that
review caused, the `SsBranch::Exact` / `InputRateExact` split, three new tests, and a rebase
onto a `main` carrying 13,514 lines it had never been tested against (#1256). **None of that
had been reviewed**, so this second pass was scoped to the whole current diff rather than the
original slice.

Same headline as round 1, and it is the part worth trusting: **nothing found disputes a
number.** No anchor value, NONMEM comparison, closed form, or mutation result changed.

### What it found

Six of the eight were **tests that would pass while covering less than their names claim** —
the same class as round 1's verbatim-duplicate "states engine" test:

- **The closed form's constants were literals with nothing tying them to the twin.** It is
  the tightest oracle in the suite (ferx sits 3.0e-10 from it) and it hard-coded `k`, `c`,
  `D`, `II`, `V` — while `ss_tad_states_engine_matches_nonmem`, fifty lines below, already
  guarded `V` against exactly that drift. Editing `ss_tad_fit.ferx`'s coefficient or
  `ss_tad.csv`'s `II` would have left every 13-digit number in the file describing a model
  that no longer exists, failing as if the code had regressed. Now
  `assert_fixture_constants_unchanged()` pins `TVCL`, `TVV`, the `0.03*TAD` term and the
  dataset's `II`/`AMT`, and both dependent tests call it.
- **`ss_monotone_run_in_anchor_is_the_same_with_an_empty_lag_slice` never calls the run-in.**
  It exercises `tad_anchor_for` directly, so re-introducing the zero-filled `Vec` at the call
  site would leave it green — the precise regression its doc claimed it prevented. Doc
  narrowed to the contract it does pin, pointing at the test that covers the call site.
- **`SsBranch::None`'s doc named three bail-outs; the test reached two.** The missing one —
  an overlapping SS infusion (`T_inf > II`) — is now tested, so the doc and the test agree on
  the set rather than on two thirds of it.

  **The mutation check on that arm refuted my own justification for it, and the comment now
  says so.** I claimed it was the unique catcher of a `None` clear migrating downward: move
  the clear below the infusion bindings, and cases 2 and 3 would stay green while only the new
  arm died. Mutating exactly that tripped **arm 2** (`II <= 0`) instead. Reasoning it through
  afterwards, no such position exists — the clear sits above all three returns, so any
  downward move breaks the earliest affected arm first and the test aborts there. What arm 4
  actually covers is narrower: a branch tag being *recorded* between the compartment check and
  the overlap return. The arm stays, with an honest reason; the mutation is the only thing
  that would have caught the wrong one.

The one `correctness`-category finding was my own `expect()`: `ext_quiet` was hoisted as an
`Option` whose `expect` asserted an invariant enforced by two `if is_inf` guards forty and a
hundred lines away. Structurally safe as written — verified — but the failure mode if it ever
drifted is a panic inside a fit, which crosses the ferx-r extendr boundary as a process
panic rather than an R error. It is now built inside each infusion arm, so **the invariant
does not exist** rather than being documented. Cost: one 1 KB stack array per infusion cycle,
against that cycle's ODE solve.

The rest: the states test asserted `V` and then divided by the literal `20.0` anyway; the
cross-module claim about `ode_predictions_ekf_with_diffusion`'s sole caller is now dated,
because the advice built on it inverts if it stops holding; and `ss_state_at_phase_pk` now
says whose branch the tag holds after it returns.

### One interaction checked because the rebase created it

`#1256` added `ferx modelsearch`, which explores **adding a lagtime** — which sounded like it
could trip `E_SS_LAGTIME_TAD_RHS` mid-search and turn a candidate into a hard error. It
cannot: modelsearch edits the `pk ...` structural line, so a candidate never carries an
`[odes]` right-hand side and `reads_tad()` is unreachable from one.

### Tier-3 coverage, closed without a dispatch

`slow-tests` never runs on a PR at any base, so this change would otherwise land with no
Tier-3 evidence. Rather than dispatch, the reachable set was enumerated — and the answer is
stronger than a green run:

- The change is a no-op unless the `[odes]` RHS reads `TAD`/`TAFD` (PK slots byte-identical;
  pinned by a `to_bits()` test) and a no-op without an `SS=1` dose.
- **97** files carry a `slow-tests` gate; **10** mention `TAD`/`TAFD`; of those, exactly one
  also contains SS doses — and its single slow-gated test uses **neither**.
- Of the **8** committed `.ferx` files whose RHS reads `TAD`/`TAFD`, the five pre-existing
  ones run on datasets with **no `SS` column at all**.

So no slow-gated test can observe this diff. A green `slow-tests` run could only have said
"the tests that exist pass"; it could never have said the set able to see the change is empty.

*(Predicate note, since the next person will reach for the loose one first: bare `slow-tests`
matches comments and `ignore =` strings and gives 130; `not(feature = "slow-tests")` gives 96
and misses file-level positive gates like `tests/categorical_convergence.rs`;
`feature = "slow-tests"` gives 97 and is both tight and complete.)*

### Gates, re-run after the rebase

Everything below was run on the rebased tree, because the earlier green was on a tree that no
longer existed once #1256's 62 files / 13,514 lines arrived.

| gate | result |
|---|---|
| `tools/preflight.sh` (fmt, 5 check sets, clippy, rustdoc, docs, public-API) | green |
| `cargo test --lib --features ci` | **4105 / 0** |
| `cargo test --lib --features ci,survival` | **4287 / 0**, twice |
| `cargo check --tests` | green |
| `cargo run -p docs-lint -- --check` | green |
| public-API baseline | unchanged — nothing new is `pub` |

**One thing that is not a clean story and should not read as one.** The first `ci,survival`
run on this tree failed — `api::iov_integration::test_iov_occasion_dsl_matches_column_fit`, a
bit-identity assertion whose own comment says *"If it ever fails, do NOT loosen it"*. It did
not reproduce (two clean full runs, 3/3 in isolation), but "it passed on a retry" is not an
attribution, so it was chased:

- **This diff cannot reach that test.** Its fixture builds every dose as
  `DoseEvent::new(.., false, 0.0)` — no `SS=1` record anywhere — so none of the steady-state
  code here executes, and the file contains no `TAD`/`TAFD`. The only changed code that *does*
  run in it is the new validation check, which is read-only, does no floating-point work, and
  is identical across the two arms being compared bit-for-bit.
- **It is #1197**, filed 2026-09-02 against `main`, before this branch existed. Same test,
  same load-dependent shape — and the *column* arm produced the **identical 64 bits** in both
  failures (`4639457431879869841`) while only the DSL arm moved. I posted that comparison to
  #1197: one side of it is reproducible across four days and many commits and the other is not,
  which is a narrower statement than the issue had. Nothing was loosened.

PR #1273, opened 2026-09-06
feat(tools): ruvsearch — residual-error model search, `power(σ, P)` in core, NONMEM's CWRES [closes #1182]

### LINKED ISSUE #1182: feat(tools): ruvsearch — residual-error model search, plus a `power` residual form in core [#1175]
Part of #1175 (P3). Depends on #1176, #1177, #1178, #1179.

Residual-error model search — Pharmpy's `ruvsearch`.

## Candidates

| Candidate | ferx status |
|---|---|
| additive | ✅ |
| proportional | ✅ |
| combined | ✅ |
| `iiv_on_ruv` | ✅ (`[error_model]` `iiv_on_ruv` line) |
| time-varying (default 4 time groups) | ✅ expressible via the covariate-selected `if/else` error model |
| **power** | ❌ **new residual form needed in core** |

Power and combined are not tested together — they are near-equivalent in shape, and Pharmpy skips the pair for that reason.

## Algorithm

Iterative, up to 3 rounds: evaluate all candidates, pick the best by likelihood-ratio test (default p = 0.001), refit the base model with the winning structure, repeat while improvements remain significant.

## CWRES pre-screen (optional, worth it)

Pharmpy screens candidates by fitting them to the **CWRES of the input model** rather than refitting the full model — orders of magnitude cheaper, then only the winner is refitted properly. ferx already computes CWRES (`stats/residual_error.rs`), so this is reachable. Ship the full-refit path first, add the pre-screen behind a flag, and assert the two agree on the fixture set.

## Sub-work

- [ ] `power(...)` residual form in `[error_model]`, owned by `stats/residual_error.rs` as the sole home of the residual-variance-and-derivatives math. Needs its `Dual2`-vs-FD parity test per CLAUDE.md, and a NONMEM comparison.
- [ ] `ruvsearch` tool on top.

## Validation

- [ ] The new `power` form: `Dual2`-vs-FD parity, plus a NONMEM anchor on a fitted dataset.
- [ ] `ruvsearch` trajectory compared against Pharmpy `ruvsearch` on a shared dataset.
- [ ] CWRES pre-screen and full-refit paths select the same winner on the fixture set.

## Deliverables

- [ ] `ferx_tools::ruvsearch`, `ferx ruvsearch` subcommand
- [ ] `docs/model-file/error-model.qmd` updated for `power`; `docs/tools/ruvsearch.qmd`
- [ ] `CHANGELOG.md` entries for both the new error form and the tool

roninsightrx:
## Why

Closes #1182 (part of the #1175 AMD epic, P3). Residual-error model search — Pharmpy's `ruvsearch` — over the `[error_model]` block: IIV on the residual error, a power form, a combined form, and a time-varying magnitude, tested by likelihood ratio from a `.ferxsearch` file, with Pharmpy's CWRES pre-screen behind a flag. Three of the four candidates were already expressible; the `power(σ, P)` form is the piece of core this adds, and `TAD` in a magnitude expression is what the time-varying candidate needs.

Two things surfaced while anchoring and are fixed here because the tool's pre-screen depends on them:

- **`CWRES` was not NONMEM's `CWRES`.** ferx standardised each residual by its own marginal SD. NONMEM's (Hooker 2007) is the vector decorrelated by the *symmetric* inverse square root of `HΩHᵀ + R`, with `R` at `IPRED` under an interaction fit. Rebuilt from NONMEM 7.5.1's own tabled `G`, `ETA`, `IPRED`, `PRED` on a 40-subject oral dataset, that recipe reproduces its column to RMS 1e-4; the old column was off by 0.75 (Cholesky decorrelation by 0.57, `R` at `PRED` by 0.33) while `IPRED` agreed to 1e-4. A CWRES-based screen turns that into a different pick (`IIV_on_RUV` screened at 48.7 where Pharmpy saw 9.4).
- `[space]` is now optional in a `.ferxsearch` file: ruvsearch has no MFL space; covsearch and modelsearch refuse a file without one by name.

## What changed

**`ferx-core`**
- `[error_model] DV ~ power(SIGMA, P)` — `Y = F + EPS(1)*F**THETA(n)`. Implemented as the proportional model plus an **exponent program** on its slot (`RuvMagnitude::per_sigma_exponent`), riding the existing per-observation residual-magnitude channel: a magnitude row is `[m…, p…]` when an exponent is active, and `stats/residual_error.rs` — the sole owner of the variance math — reads `slot_exponent(row, n_sigma, slot)` in every `_scaled` function (`variance_at_scaled`, `dvar_df_scaled`, `d2var_df2_scaled`, `dvar_dlogsigma_scaled`, the dense `∂R/∂f`, `∂²R/∂f²` builders) plus the direct-θ channel `magnitude_dvar_dtheta` (moved here from `sens_outer_gradient`, now carrying `∂R/∂p·∂p/∂θ` — also at `p = 1`, where a search starts). No new `ErrorModel` variant, so the ~200 match sites are untouched; `p = 1` takes the legacy arithmetic bit for bit. Gauss-Newton routes an exponent to its FD fallback (`has_ruv_exponent`). Rejected with per-CMT, if/else, LTBS, `weight =`; a non-positive exponent at the inits is `E_RUV_MAGNITUDE_NONPOSITIVE` with its own message.
- `TAD` built-in in magnitude expressions (`Subject::time_after_dose`: data-derived, SS-aware, no lag; shared with the sdtab `TAD` fallback). `RuvMagFn` gains a `tad` argument.
- `compute_cwres`: symmetric inverse-square-root decorrelation, `R` at the fit's own evaluation point (`IPRED` under interaction, `f(η=0)` under FOCE) — see Why.
- `edit`: `ErrorForm::Power`; `ErrorSpecText` gains `exponent`, `iiv_on_ruv`, `time_varying` (now `#[non_exhaustive]`, built with `ErrorSpecText::new` + `with_*`), renders each in one canonical spelling, declares/prunes the θ/ω a feature needs, and `ErrorSpecText::read` reads a block back or refuses one it cannot represent.

**`ferx_tools::ruvsearch`** (`crates/ferx-tools/src/ruvsearch/`)
- `mod.rs` — options (`groups`, `p_value`, `skip`, `max_iter`, `cwres_prescreen`, Pharmpy's defaults and bounds), the proportional-base rewrite of a non-proportional input, per-iteration LRT selection with family retirement (`power` ↔ `combined` retire each other), Pharmpy's final `df = 1` comparison against input and base, journalled per step, cancellable, `RuvsearchEvent` progress.
- `cwres.rs` — the pre-screen: the parent's CWRES as a dataset (`IPREDC`, doses kept for `TAD`), Pharmpy's screening models spelled in ferx (`additive(SIG * IPREDC^RUV_POW)`, `additive(SIG * sqrt(1 + (RUV_ADD/IPREDC)^2))`, …), the shape estimates mapped back (`power = θ̂ + 1`, floored at 0.02).
- `report.rs` — `steps.csv` (screened rows flagged, `cwres_dofv`), `models/<id>.ferx`, `final.ferx`, summary.
- `search/fitter.rs` — `StepFitter::fit_step_on(…, data)` so a step can fit another dataset.

**`ferx ruvsearch`** subcommand, thin. Example `examples/warfarin_ruvsearch.ferxsearch`.

## Alternatives considered

- A new `ErrorModel::Power` variant: ~200 match sites and the exponent still needs θ at every variance call site; the magnitude channel already delivers a θ-derived per-observation datum to all of them.
- Time-varying via the covariate-selected `if/else` block (the issue's suggestion): `TAD` is not a covariate either, and the form does not compose with `power`/`iiv_on_ruv`; a magnitude expression on `TAD` does.
- Leaving `CWRES` as it was: the pre-screen's whole point is Pharmpy's screen; on Pharmpy's CWRES it reproduces Pharmpy's table, on the old column it does not.

## Cross-repo dependency

ferx-r constructs neither `ErrorSpecText` nor `RuvMagFn` (checked). `compute_cwres` is `pub` but ferx-r calls `fit()`. The sdtab `CWRES` column changes value for models with η — a documented fix, not a format change.

## Breaking changes
- [x] Public Rust API — regenerated. Added: `edit::{ErrorForm::Power, ErrorForm::from_label, ErrorSpecText::{new, read, render_statement, with_*, exponent, iiv_on_ruv, time_varying}, ThetaDecl, EtaDecl, TimeVaryingDecl}` (ruvsearch authors and reads error models); `RuvMagnitude::{per_sigma_exponent, per_sigma_exponent_deriv, has_exponent}`, `CompiledModel::has_ruv_exponent`, `Subject::time_after_dose` (the exponent channel and `TAD`). Changed: `RuvMagFn` / `RuvMagnitude::eval_obs*` take `tad`; `compute_cwres` takes `r_preds`; `ErrorSpecText` is `#[non_exhaustive]`.
- [ ] `.ferx` model file format — additive (`power(...)`, `TAD`)
- [x] `FitResult` / sdtab fields — `CWRES` values change for models with random effects (see Why); no field added or renamed.

## Numerical validation

**`power` vs NONMEM 7.5.1** (`tests/nonmem/warfarin_power*.ctl`, `tests/power_nonmem_anchor.rs`): `Y = IPRED + IPRED**THETA(4)*EPS(1)`, FOCEI, warfarin.

| evaluation point | NONMEM | ferx |
|---|---:|---:|
| initial estimates (`P = 1.3`) | 94.381805 | 94.381805 |
| NONMEM's optimum (`P = 1.1124`, σ² = 7.31e-5) | −286.758823 | −286.758823 |

NONMEM's minimiser terminated with rounding errors (ERROR=134); the slow-gated fit asserts ferx reaches an OFV no worse and `P` within 0.05.

**`Dual2`/analytic-vs-FD parity**: FOCEI outer θ/σ gradient (`power_exponent_outer_gradient_matches_fd`, also at `P = 0.7`, and composed with a `TAD` multiplier), FOCE packed, inner η gradient, SAEM M-step, plus closed-form FD checks of every `_scaled` derivative and of `magnitude_dvar_dtheta` in θ.

**ruvsearch vs Pharmpy 2.2.0 + NONMEM 7.5.1** (`crates/ferx-tools/tests/pharmpy/ruvsearch_anchor/`, `ruvsearch_pharmpy_anchor.rs`, slow-gated): 40 subjects simulated with `SD = 0.25·f^0.5`, proportional input (both engines OFV 503.002). CWRES screen, iteration 1:

| candidate | Pharmpy dOFV | ferx dOFV |
|---|---:|---:|
| IIV_on_RUV | 9.40 | 10.00 |
| combined | **67.38** | **67.96** |
| power | 66.40 | 66.98 |
| time_varying1 | 0.00 | 0.00 |
| time_varying2 | 9.96 | 9.87 |
| time_varying3 | 33.02 | 32.89 |

Same pick, refit OFV 402.894 = 402.894, nothing in iteration 2, same final model. Full refit: ferx's `combined` 402.894, `power` 401.688 (selected — the generating form; the pair Pharmpy never separates). `ruvsearch_prescreen_agreement.rs` asserts the two paths agree up to that pair.

**`CWRES` vs NONMEM**: RMS 1e-4 on the same dataset (was 0.75).

- [x] Compared against: NONMEM 7.5.1, Pharmpy 2.2.0

## Performance
- [x] No significant impact expected — a model without a power form takes the legacy arithmetic (bit-identical); CWRES adds one eigendecomposition per subject post-fit.

## Tests
- [x] Unit: parser (`power`, `TAD`, rejections), `residual_error` (exponent math, FD), gradient parity ×5, `edit` (render/read/prune), validation, ruvsearch logic on a scripted fitter (17 tests incl. the pre-screen), `chi_square_isf`, CLI.
- [x] Tier-2: `tests/power_error_model.rs` (degenerate oracle `P = 1 FIX` ≡ `proportional(σ * 1.0)` bit for bit, IWRES/CWRES/simulate see the exponent, every estimator), `ruvsearch_end_to_end.rs`, `model_edit_equivalence.rs` additions, `power_nonmem_anchor.rs` (two evaluation points).
- [x] Tier-3 (slow): `power_fit_matches_nonmem`, `ruvsearch_pharmpy_anchor.rs`, `ruvsearch_prescreen_agreement.rs`.
- Mutation checks by hand: `slot_exponent` at `p = 1` returning `None` made the outer gradient read 0 on the exponent axis at the search's start — caught by `magnitude_dvar_dtheta_matches_fd_in_the_multiplier_and_the_exponent`, fixed.

## Example
- [x] `examples/warfarin_ruvsearch.ferxsearch`; `ferx ruvsearch examples/warfarin_ruvsearch.ferxsearch` runs in seconds and returns the input (nothing significant on 10 subjects; `power` is excluded by the correlation gate, |r| = 0.96 with σ).

## Docs
- [x] `docs/tools/ruvsearch.qmd` (new, in `_quarto.yml`), `docs/cli.qmd`, `docs/tools/search.qmd`, `docs/model-file/error-model.qmd` (`### Power` + NONMEM table, `TAD` in scope), `docs/model-file/output.qmd` (CWRES definition).
- [x] `CHANGELOG.md`: Added ×4, Fixed ×1.

## Checklist
- [x] `Dual2` path untouched; the variance derivatives it feeds are FD-pinned
- [ ] `Cargo.toml` version bumped — not bumped (additive API; `RuvMagFn` signature change affects no known caller)

## Docs & examples
- [x] New example `.ferxsearch` in `examples/`
- [x] Example run via the CLI (debug build)
- [ ] ferx-r rebuild — not needed

## Reviewer hints
- `stats/residual_error.rs`: `slot_exponent`, `power_loading`, `scaled_loadings`, `magnitude_dvar_dtheta` — the whole exponent math is there.
- `types.rs` `RuvMagnitude::eval_obs` — the row layout the above reads.
- `stats/likelihood.rs` `compute_cwres` — the CWRES change, with the NONMEM reconstruction in its doc comment.
- `ruvsearch/mod.rs` `search` / `full_iteration` / `screened_iteration`; `cwres.rs` for the screening models.
- Skim `report.rs`, the CLI, the docs.

## Open questions
- The pre-screen screens on ferx's CWRES, which now matches NONMEM's `CWRES` item. Under an interaction fit NONMEM also tables `CWRESI`; Pharmpy reads `CWRES`, so that is what the screen uses.


roninsightrx (commented):
Two blocking findings:

[P1] Escape the absolute-value bars in the rustdoc table — crates/ferx-tools/src/ruvsearch/cwres.rs:15. The |min IPRED| text is parsed as two extra Markdown column delimiters. The required Rustdoc job currently fails with rustdoc::invalid-markdown-table under -D warnings, so this head cannot merge cleanly. Escaping those bars fixes the table.

[P1] Preserve Pharmpy equal-time dose grouping in Subject::time_after_dose — src/types.rs:1112. The <= comparison assigns a non-SS dose to an observation at exactly the same TIME whenever the dose record precedes the observation. For dose@0, dose@24, obs@24, ferx therefore returns TAD=0. Pharmpy get_doseid explicitly moves every non-SS same-time observation into the previous dose group, yielding TAD=24. Because tad_cutoffs consumes these values, common exact-dose samples change the quantiles and can select a different time_varying candidate. The helper needs record-order/group information (plus a regression test) rather than deriving this solely from flattened timestamps.

Validation performed: cargo check --workspace --tests --all-targets; cargo fmt --all -- --check; ferx-tools ruvsearch unit tests (17/17); ruvsearch end-to-end tests (4/4); power model integration/anchor tests (6 passed, 1 slow ignored). The full library suite passed 4,135 tests apart from one numerical-warning test that passed immediately in isolation.

roninsightrx:
Both P1s addressed in 37297f0e.

**[P1] rustdoc table** — escaped the bars in `crates/ferx-tools/src/ruvsearch/cwres.rs` (`√(\|min IPRED\| / 2)`). Local `tools/preflight.sh rustdoc` was green because the local toolchain predates `rustdoc::invalid-markdown-table`; CI's does not. Fixed at the source, not with an allow.

**[P1] same-time dose/observation grouping** — `Subject::time_after_dose` now applies Pharmpy's `get_doseid` rule, and it needs no record order to do so, because Pharmpy's rule is itself order-independent: for a non-SS dose at exactly the observation's time the observation is moved to the *previous* dose group whether the dose record sits before or after it (`maxind > index` skips only the case the cumsum already put in the previous group). So the kernel excludes a non-SS dose at `obs_t` (strict `<`), keeps an SS dose at `obs_t` (Pharmpy's "no swap for SS dosing"), and falls back to the at-or-before result when nothing precedes the dose (Pharmpy's first-dose exception, `TAD = 0`). Your example — dose@0, dose@24, obs@24 — now reads `24`; `tad_cutoffs` consumes it, so the time-varying quantiles match Pharmpy's on dose-time troughs.

One deliberate split: the sdtab `TAD` fallback in `io/output.rs` **keeps** the at-or-before rule, via a `pub(crate)` `time_after_dose_at_or_before` that shares the same kernel (one implementation, a tie-break flag). That column is NONMEM's `TAD` — dose record first at a shared `TIME`, so the trough reads `0` — and it is the convention the lag-aware `tad_at_time` beside it already uses, and the one ferx's predictor applies (dose before observation at equal time). Moving the sdtab column to Pharmpy's grouping would have made it disagree with NONMEM's sdtab on every dose-time trough. Both conventions are documented on the method, in `docs/model-file/error-model.qmd` (`TAD` built-in) and `docs/tools/ruvsearch.qmd`.

Regression tests:
- `subject_time_after_dose_is_the_data_tad` (core, Tier 1): the six-row fixture with your case asserts `[0, 2, 23.9, 24, 6]` for the built-in and `[0, 2, 23.9, 0, 6]` for the sdtab fallback, and that the two differ on exactly the same-time row; SS-at-observation-time and SS-after-a-non-SS-dose cases pinned at `0`.
- `tad_cutoffs_are_pandas_quantiles_over_every_observation` (ferx-tools): adds a dose-time trough per subject and asserts the type-7 cuts `2.125, 6.0, 20.25`; under `<=` the two troughs read `0` and the cuts come out `1.125, 4.0, 10.5`. Both tests were run with the kernel flipped back to `<=` and fail with exactly those values.

The Pharmpy anchor dataset (`power_sim.csv`) has no same-time dose/observation rows, so the anchored trajectory is unchanged.

roninsightrx:
Review of head 37297f0e (medium effort). The two P1s from the previous round are confirmed fixed. Eight new findings survived verification; two candidates were refuted (`SetErrorModel` dropping `iiv_on_ruv` is documented and tested intent; `ThetaDecl` losing `FIX` has no re-render path).

**[P1] `TAD` built-in is NaN for pre-dose observations — `src/types.rs:1169`**
`Subject::time_after_dose` returns NaN for any observation before the first dose. Reproduced: `DV ~ proportional(PROP_ERR * exp(TH * TAD))` on warfarin.csv plus one EVID=0 row at TIME −0.5 for subject 1 → `check_residual_magnitude` (`src/api/validation.rs:404`) sees NaN and `fit()` errors with "magnitude … evaluates to NaN … usual cause is a missing or non-numeric cell in a covariate". The same model fits on the unmodified data. For ruvsearch's `σ * (if (TAD < c) θ else 1.0)`, `NaN < c` is false, so pre-dose rows land in the late group and `tad_cutoffs` (`crates/ferx-tools/src/ruvsearch/mod.rs:593`) drops them from the quantile, diverging from Pharmpy's group-0 cumsum (TAD = 0) with no diagnostic. Neither power_sim.csv nor warfarin.csv has a pre-dose row, so no anchor can see it. Fix: adopt Pharmpy's pre-dose convention (offset from the first pre-dose record), or reject pre-dose rows with a message naming `TAD`.

**[P1] `interaction_ok` reads only the final stage of a method chain — `crates/ferx-tools/src/ruvsearch/mod.rs:538`**
The tool re-derives core's `iiv_on_ruv` gate from `fo.method` alone, but `fit()` walks the whole `methods` chain and the parser sets `method` to the last stage. Base with `method = [foce, focei]` → parser (`model_parser.rs:7452`) yields method = FoceI, methods = [Foce, FoceI]; tool computes `interaction_ok = true`, proposes IivOnRuv (`mod.rs:800`); `fit()` (`src/api/fit.rs:540-552`) hits the Foce stage and returns `Err("iiv_on_ruv requires an interaction … got Foce")`. The row lands as "could not be fitted" (`judge_one`, `mod.rs:1423`) instead of the promised "IIV_on_RUV not tested" note. Same for `[foce, saem]`, `[foce, imp]`. Fix: iterate `fo.method_chain()` (`types.rs:7749`) with the same per-stage predicate, ideally exposed once from core so the two cannot drift.

**[P1] CWRES pre-screen inits are not clamped to the refit's bounds — `crates/ferx-tools/src/ruvsearch/cwres.rs:285`**
Screen models declare `RUV_POW(0.1, -10, 10)` (`cwres.rs:204`) and `RUV_TV(0.1, 0.001, 100)` (`cwres.rs:230`); the refit uses `POWER_LOWER/UPPER = 0.01/10` and `TIME_VARYING_LOWER/UPPER = 0.01/10` (`mod.rs:689-694`). `init_from_screen` only floors power at 0.02 and filters `RUV_TV > 0`. A screen `RUV_TV = 25` yields `theta RUV_TV(25, 0.01, 10.0)`; the parser accepts it (`ferx check` → ok; `validation.rs:3438` documents the gap) and `clamp_to_bounds` (`outer_optimizer.rs:1433`) starts at 10 with a "pinned to an optimizer bound" warning. The screen's whole purpose (seeding the refit) is defeated and the feature likely ends on the rail. Fix: clamp in `init_from_screen` into (LOWER, UPPER) with margin, or align the two bound sets.

**[P2] `finish()` can return a base model that is worse than the input — `crates/ferx-tools/src/ruvsearch/mod.rs:1474`**
The two final-comparison gates run sequentially on the mutated `final_node`. After gate 1 reverts to the input, gate 2 compares input-vs-base and returns the proportional base whenever it is less than a cutoff worse than the input. Reproduced with a temporary unit test (p = 0.05, cutoff 3.84): input 100, base 102, power-1 97 → gate 1 reverts to input (3 < 3.84), gate 2 then `102 − 100 < 3.84` → final = base, OFV 102, one fewer parameter and worse than what the user supplied. Also fires with nothing accepted: selected = base → input → base. The existing test (`mod_tests.rs:636`) uses a base 30 OFV worse, so gate 2 never fires. This is a faithful port of Pharmpy `tool.py:310-323`, and `docs/tools/ruvsearch.qmd:114` says "kept as it is", so it is a design call. At minimum pin the base-worse-than-input case with a test and a docs sentence; otherwise compare both gates against the original `selected`.

**[P2] `TAD` in `[error_model]` silently shadows a declared `TAD` covariate — `src/parser/model_parser.rs:15515`**
`RUV_RESERVED_VARS` (line 15289) seeds `TAD` as a `Variable`, the closure does `vars.insert("TAD", tad)` and never consults `obs_cov`; `[covariates]` registration (5073-5089) accepts a covariate named TAD without complaint. A model with `[covariates] TAD`, `CL = … * (1 + BETA*TAD)` and `DV ~ proportional(PROP * (if (TAD < 12) RUV_TV else 1.0))` reads the column in the structural block (trough = 0) and `time_after_dose` in the error model (trough at a q24h non-SS dose time = 24; pre-dose = NaN) → different branch, different OFV, no message. The `ODE_ONLY_BUILTINS` hint (`src/api/validation.rs:132-146`) and `scaling.qmd:267` still say "anywhere else the name is an ordinary covariate and must be a data column". Fix: reject or warn when a covariate named TAD is declared and the error model references the built-in, and update the hint plus error-model.qmd.

**[P2] GN `has_ruv_exponent()` FD-routing gate is untested and patches a symptom — `src/estimation/gauss_newton.rs:1788`**
Deleting `&& !model.has_ruv_exponent()` leaves every test green (`cargo test --lib gauss_newton|power|gn|exponent|ruv`, 240+ tests); `tests/power_error_model.rs:139` excludes GN with a comment claiming unit coverage that does not exist. CLAUDE.md requires "unit-test that routing so a scope gap fails loudly". Root cause: the Sheiner-Beal path at line 1102 calls `dr_diag_d_log_sigma` (738-797), whose Combined arm hardcodes `2σ²f²m²`, while its R diagonal already goes through the exponent-aware `variance_at_scaled`. The Laplace/interaction path (1521-1543) already uses `dvar_dlogsigma_scaled`, so the gate also needlessly forces FD there. Fix: call `error_spec.dvar_dlogsigma_scaled` at 1102 (all inputs in scope), delete `dr_diag_d_log_sigma` and the gate, and add a `power(...)` twin of `theta_dependent_magnitude_gn_grad_matches_fd_via_fallback` (line 3957) for both interaction arms.

**[P2] NONMEM OFV anchors are five orders of magnitude looser than the realised error — `tests/power_nonmem_anchor.rs:97`**
Both anchors assert `< 5e-3` against 17-significant-digit `.ext` references. Measured (by printing `{:e}` of the difference, then restored): |ferx − NONMEM| = 2.75e-8 at the initial estimates (line 97) and 2.17e-8 at NONMEM's optimum (line 116). No comment states a realised error or headroom. A regression in how the exponent enters R that moves OFV by a few thousandths passes green. Fix: bound at ~1e-6 with the measured numbers in the comment (cf. #1174's 3e-2 vs 4.7e-5).

**[P3] `ruv_obs_mult` scans every dose per observation on every inner-objective call — `src/types.rs:4495`**
`ruv_obs_mult` now calls `subject.time_after_dose(j)` unconditionally, an O(n_doses) fold with no early exit (`data_tad`, `types.rs:1143-1170`; pre-dose rows scan twice) plus two `String`-keyed HashMap inserts per observation. `individual_nll_into_with_schedule` (`src/stats/likelihood.rs:546`) recomputes it on every call and is the `obj` closure at `inner_optimizer.rs:840` (BFGS line search, FD gradient, Nelder-Mead, multistart); the hoisted `mult` at `inner_optimizer.rs:837` only reaches the analytic gradient. On a 200-dose × 30-obs TDM subject that is ~6,000 dose visits + 60 String allocs per objective evaluation, roughly doubling the pre-existing magnitude overhead, for every #484 magnitude model whether or not it mentions `TAD`. Cheapest fix: record a `uses_tad` flag in `build_ruv_magnitude` (it already walks the expression at `model_parser.rs:15501`) and skip the scan when false; better, thread the hoisted `mult` into `individual_nll_into_with_schedule` as `analytic_eta_nll_gradient_with_schedule` already does.

Lower-priority cleanup, verified but not blocking:
- `the_cwres_prescreen_evaluates_every_screening_model_on_the_residuals` runs 7 convergent screening fits (the template at `cwres.rs:253` has no `maxiter`, so the header comment is wrong), but finishes in 0.88 s; a comment fix suffices.
- Dead `features` local in `derive` (`mod.rs:773`); `expect`-guarded re-parse of the spec in both iteration fns (`mod.rs:1170`, `mod.rs:1368`).
- ruvsearch carries third copies of `rank_label`, `default_dir`, `number`/`opt_number`, the root-vetting block and the CLI directory/final-fit tail; candidates for a shared `search/` helper across the three tools.

roninsightrx:
All eight round-2 findings and the three cleanups addressed in 2a872f11.

**[P1] `TAD` NaN before the first dose** — `Subject::time_after_dose` now applies Pharmpy's dose group `0`: a pre-dose observation reads its offset from the subject's first pre-dose record (`add_time_after_dose` is a cumsum of `diff()` within `(ID, DOSEID)`, verified against the 2.2.0 source), so a lone baseline sample is `0`, never `NaN`; an undosed subject is one group-0 run. The sdtab fallback keeps NONMEM's `NaN` there. `tad_cutoffs` counts those rows as Pharmpy does and gates on "no dose in the dataset" rather than on finiteness. Tests: `subject_time_after_dose_is_the_data_tad` (pre-dose, multi-record run-in anchored on an `EVID=2` row, undosed subject) and `a_tad_magnitude_passes_with_a_pre_dose_sample` (the `check_residual_magnitude` repro, plus the row landing in the *early* group).

**[P1] `interaction_ok` read only the last stage** — one predicate now, `FitOptions::non_interaction_stage()` in core (public API baseline +1, with the reason: `fit()` refuses `iiv_on_ruv` with it and `ruvsearch` must consult the same thing). `fit()` uses it; the tool's `from_base` uses it. Tests: `non_interaction_stage_walks_the_whole_method_chain` (core) and the Tier-2 `a_chained_method_with_a_foce_stage_does_not_test_iiv_on_ruv` (`method = [foce, focei]` → the "not tested" note, no `IIV_on_RUV-1` row).

**[P1] screen seeds outside the refit's box** — `seed_inside(value, lower, upper)` keeps a seed in `[2·lower, upper / 2]`; Pharmpy's `0.02` floor on the power is that rule's lower half. Applied to `RUV_POW` and `RUV_TV` in `init_from_screen`. Test: `screen_inits_are_kept_inside_the_refit_box` (`RUV_TV = 25` → `5.0`, `RUV_POW` = 14 → `5.0`, `≤ 0` → `0.02`). Documented in ruvsearch.qmd.

**[P2] `finish()` sequential gates** — both gates now judge the *selected* model; the base is returned only when it was the selected model or the selected model beat the input but not the base. Documented as a deliberate departure from Pharmpy's code in ruvsearch.qmd. Test: `the_final_comparison_never_returns_a_base_worse_than_the_input` (your 100/102/97 case → input; nothing accepted with a worse base → input; nothing accepted with a better base → base, no reversion note).

**[P2] `TAD` shadowing a declared covariate** — rejected at parse time from `validate_ruv_expr` when the expression reads `TAD`/`tad` and `[covariates]` declares a `TAD` (either case); a declared `TAD` the error model never reads still parses. The `ODE_ONLY_BUILTINS` hint now names the error-model scope; error-model.qmd says the same. Test: `error_model_tad_builtin_refuses_a_declared_tad_covariate`.

**[P2] GN exponent gate** — the gate is gone and so is `dr_diag_d_log_sigma`. The Sheiner–Beal path now takes `∂R/∂log σ` from `ErrorSpec::dvar_dlogsigma{,_scaled}` and `∂R/∂f` from `dvar_df{,_scaled}` — the second hand-rolled copy in that path (the θ section's `2·r_j/pred` / `2·σ_p²·m₀²·pred`) was wrong under an exponent too. `no_theta_dep_magnitude` is `!has_theta_dependent_ruv_magnitude()` alone; a `power(σ, θ)` is θ-dependent like any magnitude. Tests: `power_exponent_gn_laplace_grad_matches_fd` and `power_exponent_gn_sheiner_beal_grad_matches_fd` on a θ-free `power(…, 1.5)` (one per leg so each closed form is observed on its own), `theta_exponent_gn_grad_matches_fd_via_fallback`, and `combined_ignores_sigmas_past_its_own_slots` rewritten against `dvar_dlogsigma`. Mutation results, by hand and recorded in the comments: dropping the exponent arm of `dvar_dlogsigma_scaled` fails the SB leg at `grad[3]` (rel 8.4e-1) and the Laplace leg (8.3e-1); dropping it from `dvar_df_scaled` fails both at `grad[0]` (6.1e-1). The changelog no longer says GN routes `power` to FD.

**[P2] anchor tolerance** — `OFV_TOL = 1e-6`, with the realised errors (`2.75e-8`, `2.17e-8`, re-measured) in the constant's comment and the header table.

**[P3] `ruv_obs_mult` scanning doses for every model** — `RuvMagnitude::uses_tad` (`pub(crate)`, set by the parser from the AST) and `RuvMagnitude::tad_for`, so the per-observation dose scan runs only when an expression reads `TAD`. Test: `ruv_magnitude_records_whether_it_reads_tad` (both spellings, and in the exponent). The hoisted-`mult` threading into `individual_nll_into_with_schedule` is a signature change on a public function and is left for a follow-up.

Cleanups: e2e header now says the pre-screen's fits converge (and are sub-second); the dead `features` local is gone and `derive` returns the spec it built, so neither iteration re-parses it; `rank_label` → `RankType::label()`, the three `default_dir`s → `search::default_dir(path, tool)`, `number`/`opt_number` → `search::output`. The CLI directory/final-fit tail across the three commands is left as is.

Verification: `cargo test --lib` 4143 passed; `cargo test -p ferx-tools` (433 unit + all integration) green; `power_error_model`, `power_nonmem_anchor`, `public_api_boundary`, `preflight_owns_the_fast_gates` green; `tools/preflight.sh` green; `docs-lint --check` green; public-API baseline regenerated (+`non_interaction_stage`).

roninsightrx:
Codecov patch on 2a872f11 was 88.2% (327/2779 missed) — untested paths rather than the fixes: the `ferx ruvsearch` command body, `render_summary` with a base / a CWRES screen, cancellation at each fit stage, and the `ErrorSpecText::read` refusals. 215b510d adds tests for each (the `weight =` refusal check was unreachable behind the shape regex and now runs first). All gates green locally; waiting on CI.

roninsightrx:
Functionality confirmed manually.

Issue #1272, opened 2026-09-06
fix(sens): the dual SS equilibration anchors TAFD at 0.0 and never re-anchors an infusion's quiet window — latent behind the FD gate


TeunP:
The `f64` steady-state run-in was fixed in #1270 (batch T, step T2): each `solve_ode` window
inside the equilibration is now anchored at *its own* pulse time — `0.0` where the window opens
at the pulse, `−T_inf` for the quiet window an infusion cycle re-opens, `m·II` for the monotone
segments of the capped train — and `TAFD` is deliberately left `NaN`, because an absolute-clock
train has no periodic limit to converge to (measured: it diverges **0.294 per doubling**, against
`TAD`'s 3.5e-7).

**The dual path does neither.** `equilibrate_ss_state_g` builds one closure and runs *every*
window through it (`sens/ode_provider.rs:3947-3959`):

```rust
let bare_rhs = |us: &[T], ps: &[T], t: f64, du: &mut [T]| {
    eval_rhs_anchored::<T>(program, us, ps, t, 0.0, T::from_f64(0.0), du, ..);
};
```

The two constants are `first_dose_time` and `last_dose_eff` (`:3613-3623`), so:

1. **`TAFD` is anchored at `0.0`.** `eval_rhs_anchored` computes `tafd = t - first_dose_time`
   (`:3624-3626`), so inside the run-in `TAFD` becomes the cycle-local clock — i.e. exactly
   `TAD`. That is the "silently redefine `TAFD` as `TAD`" outcome #1270 refuses to commit on the
   value side, where the same anchoring made an unconvergent quantity come back as a plausible
   `9.894973`. Recorded as finding 4 of #1258.
2. **The quiet window is never re-anchored.** An infusion cycle's quiet window is solved with
   that same `bare_rhs` over `(0.0, quiet_val)` (`:4026-4033`), so `TAD` reads `t` where the
   window in fact sits `T_inf` after its pulse and should read `t + T_inf`. On the value side
   this is mutation `Md`/`Mg` in #1270, and it dies against a composite-Simpson quadrature of
   `∫μ` computed outside the engine.

## Why this is filed as latent, not as a bug with a reproducer

**Both are unreachable today.** The analytic/dual route declines to FD for exactly this model
class — `has_ss && rhs_program.reads_model_time()`, plus `ode_tvcov_supported` — so a
[...]

PR #1274, opened 2026-09-06
fix(ode): give TAD a referent inside a lagged steady-state dose's pre-arrival window [closes #1126]

### LINKED ISSUE #1126: bug(ode): TAD has no referent inside a steady-state dose's pre-arrival window

TeunP:
## Why

**Closes #1126.** This is step **T4** of the batch-T plan in #1258, and the measured plan it
follows is [on the issue](https://github.com/FeRx-NLME/ferx-core/issues/1126#issuecomment-5562041582).

An `SS=1` dose carrying a **lagtime** on an `[odes]` model whose right-hand side reads `TAD`
returned `NaN` at every observation, a non-finite objective, and a **blank** `TAD` column in
`sdtab`. T2 (#1270) had named the combination and rejected it with `E_SS_LAGTIME_TAD_RHS`
rather than serve a number 2.7 % wrong; this supplies the missing referent and lifts that gate.

Since #1121 an `SS=1` dose with a lagtime loads its periodic compartments at the dose
**record** and the walk carries that state to the lagged arrival. The state in
`[t_dose, t_dose + ALAG)` is therefore *real* — it is the previous cycle's decaying tail — but
`TAD` there had nothing pointing at it: the anchor fold filtered doses by `d.time + lag <= t`,
so a lagged SS dose contributed nothing until its arrival and the fold fell back to the
subject's first arrival, giving a negative `TAD` where the periodic fiction says it should be
`t − (t_dose + ALAG − II)`.

**And it is not confined to the window.** #1121 flows the record-time seed to the arrival
rather than re-equilibrating there, so the wrong anchor is carried into the steady-state trough
and multiplies into every later prediction.

### One rule, in `dosing.rs`, instead of four hand-written folds

The per-dose referent is now `crate::dosing::tad_referent(dose, lag, t) -> Option<f64>`, beside
`ss_seed_phase` — the "single home shared by `pk`/`ode`/`sens`" that module already claims.
The **lag-aware** copies collapse onto it: `ode::predictions::tad_anchor_for` and the
event-driven walk's inline duplicate (deleted, along with its now-dead `first_arrival_ed`),
plus `api::output_columns::tad_at_time` via `dosing::tad_at`. Three of the four copies that
existed disagreed on `(t_dose = 120, ALAG = 3, II = 12, t = 121)`, returning `−2.0`, `NaN`
and `+1.0`.

**Two production TAD kernels remain, not one** — see the rebase note at the bottom. `io::output`'s
sdtab *fallback* now reads `types::Subject::data_tad` (#1182/#1273), which answers the lag-free
*data* question with its own NONMEM-vs-Pharmpy tie switch. They cannot be collapsed: each needs
a parameter the other has no use for. The split is documented at both ends, and the one line
they genuinely share — the periodic SS fold — is pinned by a cross-kernel test.

That is not tidying: "add the pre-arrival referent" spelled four times is four chances to spell
it differently, and #1126's own suggested shortcut gets one of the four wrong (below).

The new branch, and only this branch, changes behaviour:

```rust
if ss_seeded_at_record(dose, lag) && dose.time <= t + TAD_EPS {
    return Some(dose.time - ss_seed_phase(dose, lag));
}
```

**`ss_seed_phase`, not `rem_euclid`.** #1126 suggests reaching the window by relaxing the
post-arrival filter, whose `rem_euclid` wrap *does* return `t_dose + ALAG − II` for a negative
elapsed time — for `0 < lag ≤ II`. It stops agreeing at `lag > II`, where #1121 **clamps** the
phase to `0` on NONMEM evidence (`nonmem_anchor/results/ss_lag_ge_ii.tab`) while `rem_euclid`
wraps: at `lag = 15, II = 12` they are **9.0 apart**. A fixture written for `lag < II` cannot
tell them apart, so that case is a test rather than a comment.

**Gated on `ss_seeded_at_record`**, the same predicate that decides whether the window carries
the previous cycle's tail at all — so the referent and the state it describes cannot drift
apart in a later edit — and *not* applied to every dose: relaxing the filter unscoped flips an
ordinary lagged regimen (0/24 dosing, `lag = 3`, `t = 25`: 22.0 → −2.0). Measured on the fix,
that case still reads **22.0**, and mutation `M6` below kills nine tests.

### The seed, which is the other half

`ss_state_at_phase_pk`'s three `solve_ode` windows now take `ss_run_in_params` like every other
run-in window (#1139) — `0.0` where the window opens at its pulse, `−T_inf` for an infusion's
quiet window — instead of the bare `PkParams::values` that made the phase advance return `NaN`.

T2 left them deliberately, and the measurement behind that decision reproduces here: anchoring
them **alone** turns a loud `NaN` into a plausible number **2.7333 % high at every post-arrival
observation** (a uniform factor 1.0273332950) and 1.82 % inside the window. Both halves ship
together, and the mutation table shows neither alone suffices.

### The gate comes out

`E_SS_LAGTIME_TAD_RHS` is deleted, and `OdeRhsProgram::reads_tad()` with it — `api/validation.rs`
was its only caller, so the gate's removal made it dead code. `tests/ss_lagtime_tad_rhs_gate.rs`
is **inverted and renamed** rather than deleted: its seven cases each removed one conjunct of
the gate, all seven must now be *accepted*, and the file additionally asserts that nothing else
sweeps the combination up and that the previously-rejected model predicts the anchored value.

### What is provably not touched

- **The EKF** (`ode/ekf.rs`) calls `tad_anchor_for(doses, &[], t_start)`; with zero lag
  `ss_seeded_at_record` is false, so the new branch is unreachable and it is bit-unchanged.
  `io::output`'s fallback and T2's monotone input-rate train pass no-lag slices for the same
  reason.
- **The dual walk.** `sens/ode_provider.rs` declines `has_ss && reads_model_time()` to FD in
  both gates, and `ode_subject_supported` declines `has_periodic_ss_dose()` outright. The
  predicate is **not** widened — #1272 would have to ship in the same change if it were.

What *does* change beyond the ODE `TAD` injection, intentionally: the sdtab `TAD` column and any
`[derived]`/`[output]` expression reading `TAD`, for a lagged steady-state dose.

## Alternatives considered

- **Relaxing the post-arrival filter to `d.time <= t` for SS doses** (#1126's own suggestion).
  Rejected on measurement: it reaches the window through `rem_euclid`, which disagrees with
  #1121's NONMEM-anchored clamp by 9.0 at `lag ≥ II`, silently, because the two agree
  everywhere a `lag < II` fixture looks.
- **Fixing only `ss_state_at_phase_pk`.** That is T2's rejected option, re-measured here: a
  finite 2.73 %-wrong answer on every observation, and the two ODE engines disagreeing by the
  same amount.
- **Keeping the gate and narrowing it to a value check.** A declared `ALAG` moves during a fit,
  so a value gate lets a run start at `lag = 0` and drift into the wrong regime with no
  diagnostic — T2's own reasoning, and the reason this had to be a referent rather than a
  narrower rejection.

## Cross-repo dependency

No public-API change (`tools/update-public-api.sh --check` clean — `tafd_tad_for_subject`'s
signature is unchanged, only what it returns inside a seeded SS window).

## Breaking changes
- [x] None

A model that previously returned `NaN` and was then rejected outright by a diagnostic added
earlier in this same **unreleased** cycle now fits. The `sdtab` `TAD` column changes from blank
to a number inside a lagged SS dose's pre-arrival window; nothing else moves.

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0** (`ADVAN13 TOL=9`, `MAXEVAL=0 METHOD=0 POSTHOC`,
      `$THETA` FIX, `$OMEGA 0 FIX`, `$TABLE FORMAT=,1PE20.13`), **and** a closed form computed
      outside both engines

Nine control streams ship with this PR — five bolus (`ss_tadlag`, `train_tadlag`,
`train_tadlag_n21`, `ss_lag_auto`, `train_lag_auto`) and four infusion (`ss_inf_tadlag`,
`train_inf_tadlag`, `ss_inf_tadlag_resid`, `train_inf_tadlag_resid`). NONMEM has no `TAD`
built-in in `$DES`, so the cycle clock is written by hand as `MOD(T − ALAG + 120, 12)`; the
`− ALAG` is the lagtime, and since the pulses sit at `ALAG + 12k` and 120 is a multiple of 12
this is exactly `TAD` past the first arrival (before it the compartment is empty, so the value
is inert).

```
# 1-cpt IV bolus, CL = 1, V = 20, AMT = 100, II = 12, ALAG1 = 3, eta = 0, ode_reltol = 1e-9
# one SS=1 record at t = 480 -> arrival at 483; observations 481 / 482 INSIDE the
# pre-arrival window and 485 / 488 / 491 past it.

                                       t = 481            worst rel
before   ferx SS  0.03*TAD             NaN                 —
after    ferx SS  0.03*TAD             5.5452940850808     —
         closed form (outside both)    5.5452940824773     ferx 4.70e-10
         ferx explicit lagged train    5.5452940849422     ferx 2.50e-11
         NONMEM explicit lagged train  5.5452941786123     ferx 1.69e-8
         NONMEM SS=1 record            5.4709357327409     1.34e-2   <- measured NEGATIVE

# sdtab TAD column at t = 481:  blank  ->  10.0   (= 481 - (480 + 3 - 12))
```

**The oracle is the train, not NONMEM's `SS=1` record**, and that is measured rather than
assumed — the `ss_chz_r1`/`r2` shape:

1. **The lag machinery is exact.** With the `TAD` term removed, NONMEM's lagged `SS=1` record
   reproduces its own lagged 41-dose train to **1.14e-9**, and both sit within **8.91e-10** of
   the closed form. So nothing about lag + `SS=1` is in question on its own.
2. **Add the `TAD` read and NONMEM's own steady-state routine drifts from its own train** —
   **1.34e-2** inside the window and **6.61e-3** after the arrival. Wrong on *both* sides, so
   it is not a window-only artifact, and landing ferx on it would be reproducing a defect.
3. **A periodic limit exists** for the lagged `TAD` train: `train_tadlag_n21` (21 doses,
   observed at the same cycle phases) against `train_tadlag` moves **7.09e-7** over a doubling.
4. **The closed form is the tighter reference.** ferx sits 4.70e-10 from it and NONMEM 1.73e-8,
   so NONMEM's uniform ~1.7e-8 offset is its `TOL=9` / print floor rather than ferx drift. The
   suite asserts that ordering, so a future relaxation has to re-measure rather than loosen.

**A fifth statement, internal but not ferx-against-ferx in the usual sense.** The analytical
superposition (`pk::predict_concentration`) has read this window at elapsed time
`t − (dose.time − ss_seed_phase(dose, lagtime))` since #1121 — *exactly* the referent added
here — and that phase, clamp included, already carries its own NONMEM anchor. So the
`ss_seed_phase`-vs-`rem_euclid` choice is settled by an existing anchor, and what this change
really fixes is that the ODE engines' `TAD` disagreed with the pulse their own analytical
sibling measures from.

**Infusions**, which a bolus fixture structurally cannot reach — `ss_state_at_phase_pk` has
three windows and a bolus exercises one:

```
# same system, delivered as RATE = 25 (T_inf = 4 h)
#   ALAG1 = 3   phase 9 > T_inf   -> ACTIVE window + the QUIET window anchored at -T_inf
#   ALAG1 = 10  phase 2 < T_inf   -> ACTIVE window, previous cycle's infusion still running
#                                    across the record (`ss_residual_infusion_end`)

ALAG1 = 3, t = 481   ferx SS   6.1646016396601
                     NONMEM explicit infusion train  6.1646016683208   4.65e-9
                     NONMEM SS=1 record              6.2395080362524   1.22e-2
```

**`lag ≥ II` has no train twin, and that is measured, not a shortcut.** Under the clamp the
pulse lands on the record and nothing intervenes before the real arrival, so `TAD` runs
`0 … ALAG` and exceeds `II`; an explicit train's pulses keep their 12 h spacing and put an
arrival at 483 instead. The geometries genuinely differ (#1121), so that arm is anchored on the
closed form — `(trough + D)·Φ(t − 480)/V` — which it matches to **5.4e-10**, and the phase
itself is already NONMEM-anchored by #1121.

## Performance
- [x] No significant impact expected. `tad_referent` is one branch per dose in a fold that
      already ran per dose; `ss_state_at_phase_pk` gains one `[f64; MAX_PK_PARAMS + 2]` stack
      array (≈ 1 KB) per window, against three ODE solves on the same path. The event-driven
      walk now calls `tad_anchor_for` where it inlined the same fold, which is one non-inlined
      call per segment.

## Tests

**Tier 1** — six in `dosing.rs` (`tad_referent_tests`) covering the pre-arrival window, the
`lag ≥ II` clamp with the rejected `rem_euclid` spelling computed alongside it, the ordinary
non-SS 0/24 case, `t` before the record, `lag = 0`, and `II ≤ 0`; two in
`ode/predictions_tests.rs` — a value check against a closed form written in the test, and the
`ss = true` twin of `the_pre_arrival_tad_anchor_does_not_depend_on_the_sampling_mesh`, which
uses `ss = false` and structurally could not reach this branch.

**Tier 2** — `tests/ss_prearrival_tad_nonmem_anchor.rs` (12) and the inverted
`tests/ss_lagtime_tad_rhs_accepted.rs` (7).

**`Dual2`-vs-FD parity is discharged by routing, not by a parity test**, and is said here
rather than omitted: the analytic route declines `has_ss && reads_model_time()` to FD in both
gates, so no jet is built for this model class and there is no dual twin to compare against.
`ode_provider_ss_tad_dependent_rhs_routes_to_fd` and its siblings still pass unchanged. The
predicate is deliberately not widened; #1272 records what would have to be fixed first.

### Mutation testing

Eight mutations, each naming its own side. **All eight die; none survives.**

| | mutation | tests that die (the ones unique to it in bold) |
|---|---|---|
| `M1` | the walk anchor off — **half 1** | 11, incl. **`the_sdtab_tad_column_reads_the_pre_arrival_referent`**, **`a_lagtime_of_a_full_interval_or_more_follows_the_clamped_phase`**, **`tad_at_folds_the_referents_…`**, `ferx_lagged_ss_lands_on_the_explicit_lagged_train`, `simulate_and_predict_reach_the_pre_arrival_referent` |
| `M2` | the bolus phase-advance slice off — **half 2a** | 8, incl. **`a_seeded_ss_dose_s_pre_arrival_tad_does_not_depend_on_the_sampling_mesh`**, **`nonmem_ss_record_is_a_measured_negative`**, `simulate_and_predict_reach_the_pre_arrival_referent` |
| `M3` | the infusion **active** window off — half 2b | `lagged_ss_infusions_land_on_their_explicit_infusion_trains`, `a_zero_coefficient_tad_term_returns_the_autonomous_steady_state` |
| `M4` | the infusion **quiet** window at `0` instead of `−T_inf` | **`lagged_ss_infusions_land_on_their_explicit_infusion_trains`** — and only that one |
| `M5` | `ss_seed_phase` → the `rem_euclid` wrap | **`a_lagtime_of_a_full_interval_or_more_clamps_the_phase_instead_of_wrapping`** — and only that one |
| `M6` | drop the `ss_seeded_at_record` gate (the unscoped relax) | 9, incl. six pre-existing `sens::ode_provider` tests |
| `M7` | drop the `dose.time <= t` guard (an SS record leaks backwards) | `an_ss_record_has_no_referent_before_itself` |
| `M8` | referent at the record instead of the previous pulse | `a_seeded_ss_dose_measures_the_pre_arrival_window_…`, `a_seeded_ss_dose_reads_tad_from_the_previous_cycles_pulse` |

`M1` and `M2` are the "neither half alone suffices" claim, and they kill **disjoint** sets as
well as overlapping ones — `M1` alone reaches the sdtab column and the clamp, `M2` alone
reaches the mesh-invariance and the `NaN`-guard in the anchor suite's `worst_rel`.
`simulate_and_predict_reach_the_pre_arrival_referent` dies under **both**, which is what makes
it a regression test on those entry points rather than a restatement of the objective path.

Three things worth recording about the round itself:

- **`M4` and `M5` each kill exactly one test, and it is the test built for them.** The `−T_inf`
  quiet-window anchor is invisible to every bolus fixture *and* to the `0.0*TAD` discriminator
  (mutate it and both sides stay `TAD`-independent), which is why the infusion arm needed a
  **value** oracle rather than a discriminator. Likewise `ss_seed_phase`-vs-`rem_euclid` is
  invisible to every `lag < II` fixture.
- **The first round under-reported.** `cargo test` is fail-fast across targets, so a mutation
  that fails a `--lib` test never ran the integration binaries: `M1` looked like "kills 3" until
  it was re-run with `--no-fail-fast`, where it kills 9 including the exit-condition test. The
  harness now passes that flag. A mutation harness is code too.
- **Nothing at `955ccef2` pinned this behaviour.** `cargo test --lib --features ci` on the fixed
  tree with the fix in and no new tests: **4105 passed, 0 failed**. Green was never evidence
  here, which is why every assertion above was mutation-checked rather than trusted.

## Example
- [x] Not applicable (fix; no new API surface). The `.ferx` twins ship as
      `nonmem_anchor/ss_tadlag_fit.ferx`, `ss_inf_tadlag_fit.ferx`,
      `ss_inf_tadlag_resid_fit.ferx` and `ss_lag_auto_fit.ferx`.

## Docs
- [x] `docs/` updated — `model-file/steady-state.qmd` gains a *Under a lagtime* subsection
      (edited into T2's `TAD` section, not appended beside it) with the measured table, loses
      the `E_SS_LAGTIME_TAD_RHS` limitation paragraph, and its `TAFD`/`TIME` half is split into
      its own heading; `file-formats/check-report.qmd` loses the diagnostic's row;
      `model-file/ode-models.qmd`'s *`TAD` before the first dose arrives* now carves out the
      steady-state case (that section stated the first-arrival rule unconditionally and cited
      #1110, which closed 2026-09-01)
- [x] `CHANGELOG.md` `[Unreleased]` — **replaces** T2's `E_SS_LAGTIME_TAD_RHS` bullet in the
      same block rather than adding an "added it, removed it" pair; the gate never shipped

## Checklist
- [x] `cargo test --lib --features ci` — 4113 passed, 0 failed
- [x] `cargo test --features ci,survival` — all targets pass
- [x] Nothing on the `Dual2` path changed; the FD routing gate is untouched (see above)
- [ ] `Cargo.toml` version bump — not needed, non-breaking

## Reviewer hints

The load-bearing parts, in order:

1. **`crate::dosing::tad_referent`** — one function, and the only place the rule now exists.
   The two things to check are the `ss_seed_phase`-not-`rem_euclid` choice (settled by #1121's
   NONMEM anchor, which the analytical engine already follows) and the `ss_seeded_at_record`
   gate that keeps ordinary lagged doses out.
2. **The asymmetry between the two ODE engines**, documented in the anchor suite's module doc:
   `apply_segment_boundary` overwrites the flowed state at the lagged arrival, so a
   post-arrival assertion on the dense states walk is **blind** to the walk's anchor. That is
   why the pre-arrival checks run on both engines and the post-arrival ones on the event-driven
   walk, and why `both_ode_engines_agree_across_the_pre_arrival_boundary` exists.
3. **The mutation table**, especially `M4` and `M5`.

Skimmable: the NONMEM control streams (mechanical variations on T2's), the doc edits, and the
inverted gate test, which is the old file with its assertions flipped.

## Open questions

None blocking. Two things recorded rather than fixed:

- **#1272 stays open.** The dual SS equilibration anchors `TAFD` at `0.0` and never re-anchors
  an infusion's quiet window — the same two defects this fixes on the value side — and stays
  unreachable behind the FD gate. Anything that widens that gate must fix it in the same change.
- **T3 of #1258** (`TAFD` / `T` / `TIME` under `SS=1`) is deliberately untouched;
  `tafd_stays_nan_under_a_lagged_ss_dose` is a scope pin that fails if a later change quietly
  extends the referent to `TAFD` without doing that work.

---

### Review round (`xhigh`, self-review)

[10 findings](https://github.com/FeRx-NLME/ferx-core/pull/1274#issuecomment-5563127527); 9 applied in `32bb409e`, 1 recorded.
Two were CLAUDE.md tolerance violations in the new tests and are re-measured there (the
infusion arm's realised error was a placeholder string; the clamp test quoted a figure from a
different observation set). `simulate()` / `predict()` are now **measured** rather than claimed.
`dosing::tad_at` collapses the fold the first commit left duplicated in two callers. The one
not applied is the dense-vs-event-driven arrival-state asymmetry, which is #1121's geometry
rather than this change's.

---

### Rebase onto #1273 (ruvsearch), `90c6ea0a`

#1273 landed on `main` and touched four of this branch's files. Rebased; one conflict, in
`io::output`'s sdtab TAD fallback, where **main's line was kept**: this branch's change there
was a reuse refactor documented as behaviour-preserving, while #1273's carries a NONMEM
record-order tie rule `tad_referent` does not implement, so keeping this branch's would have
silently reverted it.

**Nothing of #1126 is lost, verified rather than argued.** `api/fit.rs` runs
`compute_extra_output_columns` whenever `model.has_lagtime()`, so any lagged model populates
`per_obs_tad` and never reaches that fallback — and the fallback is lag-free, where the seeded
pre-arrival branch is unreachable by construction.

The rebase also made three doc claims on this branch false, and exposed two cross-kernel
invariants that nothing enforced:

- `types.rs` documents the fallback as keeping *"the same tie rule as the lag-aware
  `tad_at_time`"* — a claim about **this** branch's function, which #1273's own test does not
  check (it compares its two conventions against each other). A strict `<` here would have
  broken it silently.
- The two kernels' **steady-state** arms are `obs_t - elapsed.rem_euclid(d.ii)` character for
  character — the only literally duplicated line between them — and neither #1273's fixture nor
  the tie fixture reaches it; both use plain doses.

`the_zero_lag_fold_breaks_a_same_time_tie_the_way_the_sdtab_fallback_does` pins both, each on a
fixture that straddles the boundary: the same-time tie at `obs@24` where the two data
conventions are known to differ, and a full SS cycle plus the wrap. Mutating the tie kills it
and five others.

Post-rebase: preflight green, `cargo test --features ci,survival` **180 binaries, 0 failures**.


TeunP:
## Self-review — `xhigh`, 10 findings

Adversarial pass over my own diff (10 finder angles, dedup, gap sweep). Two of these are
CLAUDE.md violations in the *new* tests, and both are the same shape: a number that reads as
measured but was not measured on the thing it annotates. Both are re-measured below.

Nothing here changes the fix itself or any anchored value.

### 1 · `"measured in the PR table"` is a placeholder, not a measurement — `tests/ss_prearrival_tad_nonmem_anchor.rs`

`lagged_ss_infusions_land_on_their_explicit_infusion_trains` carries the literal string
`"measured in the PR table"` in its realised-error column for the residual-infusion arm. The PR
table does not carry one either, so the `5e-8` bound is unjustified.

CLAUDE.md: *"Measure the tolerance; never justify it with a story. Run the comparison, print the
worst realised error, put that number in the comment, and pick the bound from it with stated
headroom."* Re-measured: **2.4196e-9** against `train_inf_tadlag_resid`, so the bound has 20×
headroom nobody stated.

### 2 · The clamp test quotes an error measured on a different observation set — same file

`a_lagtime_of_a_full_interval_or_more_follows_the_clamped_phase` says *"realised 5.44e-10 across
481/482/485/488/491"*. That figure came from **481/485/494.9/495/500**, a set this test does not
use. Re-measured on the set it actually runs: **4.0869e-10**.

A number quoted to three digits that describes another experiment is worse than no number.

### 3 · `simulate()` / `predict()` claimed fixed without measuring — `CHANGELOG.md`

The entry says the combination is served correctly *"on every path including `simulate()` and
`predict()`"*. Both plausibly route through the fixed predictors, but that is a code read, not a
measurement — and T2 (#1270) measured `simulate()` explicitly rather than asserting it, because
the SS family has several entry points. A user-facing line should not rest on inference.

### 4 · The plumbing discriminator never asserts its own replacements applied — same file

`a_zero_coefficient_tad_term_returns_the_autonomous_steady_state` builds both models with
`str::replace` on the committed `.ferx` source. If a fixture edit made **both** substitutions
no-op, the two models would be identical, `worst` would be `0.0`, and the assertion would pass
while testing nothing. Verified each string occurs exactly once today, so the hole is latent —
but this is the file's only plumbing discriminator, and a vacuous green is precisely what the
mutation round cannot see.

### 5 · The fold is still duplicated, only the per-dose rule was collapsed — `src/io/output.rs`

The change collapses the per-dose arithmetic onto `dosing::tad_referent` but leaves the
surrounding `fold(NEG_INFINITY, f64::max)` + `is_finite` + NaN tail hand-written identically in
`io::output`'s sdtab fallback and `api::output_columns::tad_at_time`. The PR's own argument —
"spelled four times is four chances to spell it differently" — applies to the fold too.

### 6 · The two ODE engines still build the arrival state by different routes — `src/ode/predictions.rs` *(not fixing here)*

The dense walk overwrites the flowed state with a fresh `equilibrate_ss_state` at the lagged
arrival (`ss_arrival_is_trough`); the event-driven walk flows the record seed through. After
this fix they agree only because the corrected flow lands on the trough. CLAUDE.md's *"the fix
is one implementation"* argues for collapsing them — but that is #1121's design, NONMEM-anchored,
and unifying it is a change to the seed geometry rather than to the referent. Left as is, with
`both_ode_engines_agree_across_the_pre_arrival_boundary` standing for the invariant. Recorded so
the next person does not have to re-derive it.

### 7 · `assert!(got.is_empty())` is too wide — `tests/ss_lagtime_tad_rhs_accepted.rs`

Asserts *no* diagnostics at all rather than "not this retired code". Any future data check that
fires on a 1-cpt SS + lagtime model turns this red pointing at #1126 rather than at the change
that added it.

### 8 · `tad_referent` lacks `#[inline]` — `src/dosing.rs`

Its two production callers are `#[inline]` folds running per dose, per segment, per objective
evaluation. The code it replaced was a closure body the optimizer saw inline by construction.

### 9 · `ferx_dense_states` hardcodes state index `0` unguarded — anchor suite

`state_divisor_is_v` pins the divisor but not that the model has one compartment. A twin gaining
a depot would read the wrong state and fail as an engine regression rather than a fixture edit —
the one literal `fixture_constants_are_what_the_closed_form_assumes` misses.

### 10 · Anchor tests re-read and re-parse each CSV per call — anchor suite

`population(csv)` hits disk on every invocation, two or three times per test.

---

Fixes for 1–5 and 7–10 follow in the next commit; 6 is recorded, not applied.

Issue #1018, opened 2026-08-21
fix(estimation): FOCE/FOCEI estimate the full lower triangle on a mixed block_omega, ignoring declared structural zeros


dominic-irx:
## Summary

On a **mixed** `Ω` — a `block_omega` plus a standalone `omega` — FOCE/FOCEI estimate the *full*
lower triangle. The off-diagonals the model declares absent come back non-zero, and
`n_parameters` counts them, so AIC/BIC are wrong too.

There is no error and no warning. The fit reports a covariance structure the model file did not
ask for.

VI, SAEM and Gauss-Newton all honour the declared structure on the same model, so this is
specific to the outer optimizer's parameterization.

## Reproduction

The shipped example is already this shape. On `main` (`ae6c208c`), no edits:

```bash
cargo run --release -- examples/warfarin_block_omega.ferx --data data/warfarin.csv
```

`examples/warfarin_block_omega.ferx` declares `ETA_KA` *outside* the `(ETA_CL, ETA_V)` block:

```
block_omega (ETA_CL, ETA_V) = [0.09, 0.02, 0.04]
omega ETA_KA ~ 0.30
method = foce
```

so `Ω(KA,CL)` and `Ω(KA,V)` are structural zeros.

### Actual

```yaml
data:
  n_parameters: 10          # 3 theta + a FULL 6-element lower triangle + 1 sigma

omega:
  ETA_CL:  { variance: 0.028695 }
  ETA_V:   { variance: 0.009629 }
  ETA_KA:  { variance: 0.350189 }
  ETA_V__ETA_CL:  { covariance:  0.001754, correlation:  0.104575 }   # declared - fine
  ETA_KA__ETA_CL: { covariance: -0.031659, correlation: -0.282055 }   # NOT declared
  ETA_KA__ETA_V:  { covariance:  0.020686, correlation:  0.328136 }   # NOT declared
```

### Expected

`n_parameters: 8`, and no `ETA_KA__*` entries — `io/output.rs` only emits a pair when
`|cov| > 1e-15`, so an honoured structural zero is silently absent by construction.

## Why this is the outer optimizer specifically

Every other estimator filters on `omega.free_mask`:

- `src/estimation/gauss_newton.rs:902,911` and `:1339,1351`
- `src/estimation/saem.rs:1720,2362` (`from_matrix_with_mask`, "preserves the structural
  free_mask")
[...]

PR #1276, opened 2026-09-07
feat(tools): iivsearch / iovsearch — variability-structure search, κ edits in core [closes #1183]

### LINKED ISSUE #1183: feat(tools): iivsearch / iovsearch — variability-structure search [#1175]
Part of #1175 (P4). Depends on #1176, #1177, #1178, #1179 — and on the BIC(iiv) variant from #1177 specifically.

Variability-structure search — Pharmpy's `iivsearch` and `iovsearch`.

## Why this is the hardest of the four

Covariate and structural search are line and template edits. IIV search is **expression surgery**: removing IIV on a parameter means deleting `ETA_X` from an `[individual_parameters]` right-hand side *and* its `omega` line, and correlation search means synthesising a `block_omega (A, B) = [...]` from what was a set of diagonal entries.

Mitigation, decided in #1176: a search-eligible model must be written in canonical `P = TVP * exp(ETA_P)` form, and anything else is rejected with a message naming the parameter. Same posture `[covariate_model]` already takes on a non-product right-hand side.

## iivsearch scope

- **Number of IIVs**: `top_down_exhaustive` (remove IIVs in all combinations, including the naive-pooled model when no IIV is mandatory) and `bottom_up_stepwise` (add one at a time until no improvement).
- **Correlation structure**: exhaustive over block patterns among the retained etas — `[CL]+[VC]` versus `[CL,VC]` and so on.
- **Simultaneous stepwise**: add one IIV at a time, testing each new eta against the existing blocks.
- Ranked on **BIC(iiv)** — this is why #1177 is a hard dependency; the observation-count BIC systematically mis-ranks variability structures.

## iovsearch scope

Identify the IOV structure and trim IIV accordingly, over `[iov]` κ terms.

## Known risk

Full-block omega with ≥ 3 etas is already flagged as a moderate local-minimum risk (`docs/examples/multistart.qmd`). Correlation search generates exactly those models by construction, so `n_starts` must scale with the block size rather than being a fixed constant.

## Validation

- [ ] Trajectory and final structure compared against Pharmpy `iivsearch` / `iovsearch` on a shared dataset.
- [ ] BIC(iiv) ranking unit-anchored to `pharmpy.modeling.calculate_bic()` (delivered in #1177, re-asserted here end-to-end).
- [ ] A model outside canonical form is rejected with a named parameter, never silently mis-edited.
- [ ] Degenerate oracle: a single-point space returns the base fit bit-for-bit.

## Deliverables

- [ ] `ferx_tools::iivsearch`, `ferx_tools::iovsearch` and their subcommands
- [ ] `docs/tools/iivsearch.qmd`
- [ ] `CHANGELOG.md` entry

roninsightrx:
## Why

The last two search tools of the #1175 epic (P4): Pharmpy's `iivsearch` (which parameters carry an η, and how their ω are blocked) and `iovsearch` (which carry an occasion-level κ, and which η become redundant once they do). Both are *expression surgery* rather than line surgery — removing an η means rewriting `TVP * exp(ETA_P)` and its `omega` line; a correlation candidate means synthesising a `block_omega` — which is why the issue asked for the canonical-form rule from #1176 to be enforced by name, and for the BIC(iiv) of #1177 to rank on.

## What changed

**`ferx-core::edit` (public API widened, baseline regenerated):**
- `ModelEdit::AddIov` / `DropIov` write and remove a κ in the canonical `P = TVP * exp(ETA_P + KAPPA_P)` form; `SetKappaBlock` / `SplitKappaBlock` and `SplitOmegaBlock` block and unblock κ and η. `DropIiv` / `DropIov` on a blocked random effect now *shrink the block around the survivors* (Pharmpy's `remove_iiv` on a joint distribution) instead of refusing — the same rewrite the pruner already applied on a structural swap.
- `VariabilityText::read` — the reading side of the η/κ edits, the way `ErrorSpecText::read` is the reading side of `SetErrorModel`: which parameter carries which η and κ, the blocks, the `FIX` flags, and per parameter whether the line is in the canonical form. A search refuses a non-canonical parameter **by name** before fitting anything (validation bullet 3).
- `SeedInits` carries `kappa` / `block_kappa` estimates; the canonical form generalises to `exp(ETA [+ KAPPA])` so the four edits are closed under each other.
- Each item is needed by `ferx_tools::iivsearch` / `iovsearch` (and is reachable from ferx-r as an ordinary external crate).

**`ferx-tools`:**
- `iivsearch`: `top_down_exhaustive`, `bottom_up_stepwise`, `simultaneous_stepwise`, `skip`, then `top_down_exhaustive` over the blocks (single full cliques + the diagonal model — Pharmpy's `_is_valid_block_combination`), each step ranked with its parent on the BIC(iiv) (`bic` means `bic_iiv` here, as in Pharmpy), the final model compared with the input. A new block's correlations come from the parent's EBE correlation (Pharmpy's `create_joint_distribution(individual_estimates=…)`), shrunk until PD; a block over ≥3 η gets `block_retries` extra starts per η (the issue's multistart risk), via a new `Candidate::starts` override the runner honours.
- `iovsearch`: full-IOV model, every proper subset of κ removed, ranked with the input on the BIC(random); then every subset of the η beside a κ removed. `disjoint` / `joint` / `same-as-iiv` / `explicit` distributions; κ at a tenth of the fitted η variance; the base must read its occasions itself (`iov_column`), refused with the fix named otherwise.
- The search seed floors a collapsed κ as it floors a collapsed ω (a full-IOV parent whose κ collapsed handed every child a start the engine refused — measured on the anchor).
- Coverage: `IOV(p, EXP)` and `COVARIANCE(IOV, …)` are searchable; `IOV(p, ADD)` is a named gap.

**`ferx-cli`:** `ferx iivsearch`, `ferx iovsearch`, thin.

**Docs:** `docs/tools/iivsearch.qmd`, `docs/tools/iovsearch.qmd`, coverage table, nav, CHANGELOG, two example `.ferxsearch` files.

## Alternatives considered

- Fixing #1018 (FOCE/FOCEI estimate the full lower triangle of a mixed block + diagonal ω) inside this PR. It is the outer parameterization, touches every packed-coordinate consumer (covariance, SIR, the analytic ω gradient) and is its own PR. iivsearch **states it** instead: a note names #1018 whenever a partial-block candidate is generated under FOCE/FOCEI, the docs carry a "Known limitation" section, and the anchor asserts the resulting divergence by name so it cannot go stale.
- Pharmpy's `linearize` path: not offered; ferx fits in-process and the real fits are what the linearisation approximates.
- General set partitions in the block stage (`[CL,V]+[KA,F]`): Pharmpy admits cliques only (its own FIXME); mirrored, and documented.

## Cross-repo dependency

The R wrapper does not expose the edit layer or the search tools yet; nothing in `FitResult` / sdtab changed.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — regenerated with `tools/update-public-api.sh`. Added: `ModelEdit::{AddIov, DropIov, SetKappaBlock, SplitKappaBlock, SplitOmegaBlock}` (the κ / unblock moves iovsearch and iivsearch's block stage need), `VariabilityText` + `ParameterVariability` + `RandomEffectDecl` + `RandomEffectBlock` (both tools read the structure before editing; ferx-r can read it the same way). One behaviour change inside the existing surface: `DropIiv` on a blocked η shrinks the block instead of returning `Err` (the old test pinned the refusal; the new one pins the shrink).

## Numerical validation

Pharmpy 2.2.0 driving NONMEM 7.5.1 in the `pmx` container, on simulated datasets (`crates/ferx-tools/tests/pharmpy/{iivsearch,iovsearch}_anchor/`, READMEs with the full tables), replayed by two slow-gated tests.

**iivsearch** (40 subjects, IIV on CL and V with r = 0.6, none on KA; BIC(iiv)):

| algorithm | step | model | Pharmpy | ferx |
|---|---|---|---:|---:|
| top-down | 1 | `[CL]+[V]` (best of 7) | 697.411 | 697.411 |
| top-down | 2 | `[CL,V]` | 666.086 | 666.086 |
| top-down | final | `[CL,V]`, OFV | 655.020 | 655.020 |
| bottom-up | base `[CL]` → `[CL]+[V]` → `[CL,V]` | | 1002.664 / 697.411 / 666.086 | 1002.664 / 697.411 / 666.086 |
| simultaneous | 1 | `[CL,V]` from `[CL]` | 666.086 | 666.086 |
| simultaneous | 2 | `[CL,V]+[KA]` | 670.227 (NONMEM: sig. digits unreportable) | 662.482 **(#1018: fitted as `[CL,V,KA]`)** |

Same candidates in the same numbering, same step winners, same final model for top-down and bottom-up. The simultaneous step-2 divergence was chased to the bottom rather than tolerated: NONMEM refitted from ferx's estimates evaluates that point at 658.50 — and so does ferx when *evaluating* the declared model (`maxiter = 0`) — while ferx's *fit* reports 647.73, because the outer optimizer frees the cross-block covariances (#1018, already open). ferx therefore ends on `[CL,V]+[KA]` (really the full block) where Pharmpy ends on `[CL,V]`; the anchor asserts that divergence by name so the #1018 fix turns it red, and the search carries a note naming #1018 on any such candidate.

**iovsearch** (40 subjects × 3 occasions, IOV on CL only; BIC(random)):

| model | Pharmpy | ferx |
|---|---:|---:|
| `IOV([CL])` — best | 1219.577 | 1219.577 |
| `IOV([CL]+[V])` | 1222.053 | 1222.048 |
| full `IOV([CL]+[KA]+[V])` | 1225.937 | 1225.738 |
| input | 1315.980 | 1315.980 |
| step 2 `IIV([KA]+[V]);IOV([CL])` | 1273.951 | 1273.951 |

Same trajectory, same final model (OFV 1190.066) under both `disjoint` and `same-as-iiv`.

- [x] Compared against: Pharmpy 2.2.0 / NONMEM 7.5.1 (trajectory + final structure, the issue's validation bullet 1); BIC(iiv) re-asserted end to end (bullet 2); non-canonical model refused by name (bullet 3, Tier-2 test); single-point space returns the input fit bit for bit (bullet 4, Tier-2 test).

## Performance
- [x] No significant impact expected — new code paths only.

## Tests
- [x] Tier 1: `src/edit/mod_tests.rs` (+12: κ edits, split, shrink, seed, the structure reader); `iivsearch/mod_tests.rs` (13, scripted fitter: enumeration per algorithm in Pharmpy's order, ranking with tie / cutoff / gate, EBE-seeded blocks, `#1018` note, cancellation, report); `iovsearch/mod_tests.rs` (7); coverage / seed updates.
- [x] Tier 2: `tests/iivsearch_end_to_end.rs` (5) and `tests/iovsearch_end_to_end.rs` (4): degenerate oracle bit-for-bit, every candidate of every algorithm / distribution compiles and evaluates, hand-written twins (a dropped η, an EBE-seeded block, a κ) bit-identical, the canonical-form refusal, the `iov_column` refusal.
- [x] Tier 3 (slow-gated): `tests/iivsearch_pharmpy_anchor.rs` (3), `tests/iovsearch_pharmpy_anchor.rs` (2).
- [x] `cargo test -p ferx-tools` 459 + integration green; `-p ferx-cli` green; core `edit::` 79 green.

- [x] `examples/warfarin_iivsearch.ferxsearch`, `examples/warfarin_iovsearch.ferxsearch`

## Docs
- [x] `docs/tools/iivsearch.qmd`, `docs/tools/iovsearch.qmd`; `search.qmd` coverage table and `[space]` note; `modelsearch.qmd` link
- [x] Linked in `docs/_quarto.yml`
- [x] `CHANGELOG.md` `[Unreleased]` entries

## Checklist
- [x] `Dual2` path untouched
- [ ] `Cargo.toml` version bumped if breaking — not bumped (additive API)

## Docs & examples
- [ ] ferx-site / ferx-book: a follow-up once the tools page family is added there (the site's model-DSL pages are unaffected: no new `.ferx` syntax).
- [x] Example execution: both example `.ferxsearch` files run via the CLI.

## Reviewer hints
- `src/edit/apply.rs`: the η/κ section was rewritten around one `split_trailing_exp` and one `drop_random_effect` for both kinds; the interesting cases are `exp(ETA + KAPPA)` and a drop that shrinks a block.
- `crates/ferx-tools/src/iivsearch/mod.rs`: `derive` order is drops → splits → adds → blocks, so a shrunk block and a re-blocked one cannot collide; `block_seed` is the EBE-correlation seed.
- `crates/ferx-tools/tests/pharmpy/iivsearch_anchor/README.md` explains the #1018 divergence with the NONMEM cross-check.

## Open questions
- #1018 is the real blocker for a correct correlation stage on ≥3 η under FOCE/FOCEI. Fix next, then drop the note and flip the simultaneous anchor test to full agreement (the memory file and the test comment say what to change).


crates/ferx-tools/src/search/runner.rs
```diff
@@ -633,7 +633,7 @@ fn compile_and_fit(
 
     let init_params = parsed.model.default_params.clone();
     let mut fit_options = base.quiet();
-    fit_options.n_starts = options.n_starts.max(1);
+    fit_options.n_starts = candidate.n_starts.unwrap_or(options.n_starts).max(1);
```
roninsightrx, line 636: [P1] Include the effective start count in cache/dedup identity

This makes `n_starts` a result-producing per-candidate setting, but `Candidate::hash()` is still only the canonical model text, resume lookup is keyed only by that hash, and `SearchManifest` records only the run-wide `options.n_starts`. If a user changes `[iivsearch]/[iovsearch] block_retries` and resumes the same directory, compatibility still passes and the previously journalled lower-start fit is silently reused. The generic runner likewise deduplicates identical model text even when two `Candidate`s request different start counts. Since the extra starts exist specifically to escape local minima, that can silently change the selected model. Please include the effective count in the cache/dedup identity (or record and compare it per journal row) and cover the changed-override resume case.

crates/ferx-tools/src/iovsearch/mod.rs
```diff
@@ -0,0 +1,1440 @@
...
+        cancelled |= report.cancelled;
+
+        let mut nodes = vec![full_node.clone()];
+        if !cancelled {
+            // Every non-empty proper subset of the searched κ removed,
+            // smallest removals first, from the full-IOV model.
+            let searched = space.searched();
+            let kept = space.kept();
+            let mut targets = Vec::new();
+            for size in 1..searched.len() {
```
roninsightrx: [P1] Generate the forced-only IOV candidate

The exclusive upper bound drops the case where all optional κ are removed. That is only equivalent to the input when there are no forced additions. For a valid space such as `IOV(CL, EXP); IOV?(V, EXP)`, `searched.len() == 1`, so this loop emits no removal candidate at all: the search fits `[CL,V]` but never `[CL]`, even though the plain `IOV(CL)` is documented as forced and `V` is optional. The existing unit test at `the_input_is_returned_when_no_iov_candidate_beats_it_and_a_kept_kappa_stays` currently pins this incomplete space. Please include removal size `searched.len()` when forced κ remain (or generate it generally and deduplicate the true input) and add a case where the forced-only model wins.

crates/ferx-tools/src/iovsearch/mod.rs
```diff
@@ -0,0 +1,1440 @@
...
+            parameters: all.len(),
+        });
+        let id = new_id();
+        let candidate = derive(&id, &input, &full, &space, options)?;
+        let report = fitter.fit_step("iov-all", std::slice::from_ref(&candidate))?;
+        notes.extend(report.warnings.iter().cloned());
+        let result = report
+            .results
+            .first()
+            .ok_or("the full-IOV model was not fitted")?;
```
roninsightrx: [P2] Preserve the input result when the full-model fit is cancelled

`Runner` deliberately omits a candidate result when `fit()` returns `Err` while the cancel flag is set. Therefore a cancellation during this one-candidate full-IOV fit produces `report.cancelled == true` with an empty `results`, and this line converts the cancellation into `Err("the full-IOV model was not fitted")`. The caller loses the already completed input model and gets an ordinary failure instead of an `IovsearchResult { cancelled: true, ... }` (and the CLI cannot return its intended 130 status). Please handle `report.cancelled` before requiring the result and return the input as the last completed model. The same pattern exists in iivsearch when its derived `base` fit is cancelled, where the completed input can likewise be preserved; add cancellation-during-single-fit tests rather than only a scripted report that contains a result and has `cancelled` set.

CHANGELOG.md
```diff
@@ -20,6 +20,41 @@ section of the SDLC for the versioning policy).
...
+  `top_down_exhaustive` — one candidate per full block among the retained η, plus the
+  diagonal model — each step ranked with its parent on the BIC(iiv) (`OFV + n_ω·ln(n_subjects)`,
+  what `bic` means for this tool) behind the strictness gate, and the final model compared
+  with the input. Every candidate is written by the η / block edits from its parent's
+  estimates; a new block starts at the parent's EBE correlations and a block over three or
+  more η gets `[iivsearch] block_retries` extra starts per η. A parameter outside the canonical
+  `P = TVP * exp(ETA_P)` form is refused by name before anything is fitted. Anchored against
+  Pharmpy 2.2.0 through NONMEM 7.5.1 on a simulated dataset: the same candidates in the same
+  numbering, the same winner at each step and the same final `[CL,V]` at the same BIC(iiv),
+  for all three algorithms.
```
roninsightrx: [P2] Do not claim full simultaneous-algorithm agreement

This permanent changelog entry says all three algorithms have the same step winners and final `[CL,V]`, but this PRs own numerical-validation section and `simultaneous_stepwise_agrees_on_step_one_and_diverges_on_the_mixed_omega` anchor say the simultaneous step-2 criterion diverges because of #1018 and ferx consequently selects `[CL,V]+[KA]`. Please narrow this sentence to top-down/bottom-up agreement and explicitly state the asserted simultaneous limitation, so the released changelog does not promise validation the PR did not achieve.

roninsightrx (commented):
Review complete: changes needed before merge.

Blocking findings are posted inline:

- per-candidate `n_starts` is not part of dedup/resume identity, so changing `block_retries` can silently reuse a fit produced with fewer starts;
- forced + optional IOV spaces omit the forced-only candidate;
- cancellation during the single full-IOV (or derived iivsearch base) fit returns an ordinary error instead of preserving the last completed model;
- the changelog overstates simultaneous-stepwise NONMEM/Pharmpy agreement despite the asserted #1018 divergence.

The required Codecov patch gate is also red: 88.55% of changed lines hit versus the repository target of 90.00% (390 changed lines missing coverage).

Local verification on c5762eb2:
- `RUSTC_WRAPPER="" cargo check -p ferx-tools` — passed
- `RUSTC_WRAPPER="" cargo test -p ferx-tools` — passed (460 unit tests plus integration tests; slow anchors ignored by their feature gate)
- `RUSTC_WRAPPER="" cargo test --lib edit::` — passed (79 tests)

I am the PR author under the active GitHub identity, so GitHub does not allow a formal REQUEST_CHANGES self-review; this COMMENT review records the blocking result.

roninsightrx:
All four findings fixed, each with a test that fails without the fix. Pushed as a follow-up commit.

**[P1] Effective start count in cache/dedup identity** — `Candidate::hash()` now folds `n_starts` in when the candidate sets one: `sha256(hex(canonical_hash) + "\nstarts=N")`. A candidate with no override hashes exactly as before, so existing journals still resume. That one key covers both halves of the finding — dedup and the fit cache/resume lookup are the same key — and `SearchManifest::n_starts` keeps its run-wide meaning, now documented as such.

Two tests in `search/runner_tests.rs`:
- `candidates_that_differ_only_in_their_start_count_are_each_fitted` — two candidates with byte-different but canonically identical text, one with `.starts(7)`: both fitted, `deduped == 0`, and the no-override candidate still hashes to the plain canonical hex.
- `a_resume_after_raising_the_start_override_refits_rather_than_reusing` — same model and same run-wide options, `starts(3)` then `starts(9)`: refits; then `starts(9)` again: reuses. So the re-key is the override's doing, not a hash that churns every run.

Mutation-checked: reverting `hash()` to the plain canonical hash turns both red.

**[P1] Forced-only IOV candidate** — the bound is now `1..=searched.len()` when a forced κ remains, and `1..=searched.len()-1` when nothing is forced (where removing all of them *is* the input, already ranked). `a_forced_kappa_gives_a_forced_only_candidate_that_can_win` covers `IOV(CL,EXP); IOV?(V,EXP)`: `run1 = [CL,V]`, `run2 = [CL]`, `run2` wins, and its model carries `KAPPA_CL` but not `KAPPA_V`. The same test's second half pins that an all-optional space still stops at proper subsets and does not refit the input. `the_input_is_returned_when_no_iov_candidate_beats_it_and_a_kept_kappa_stays` is updated — it did pin the incomplete enumeration.

**[P2] Cancellation during a single fit** — both sites now match on `report.results.first()` and treat `None` + `cancelled` as "the input is the last completed model": iovsearch returns an `IovsearchResult { cancelled: true, final_id: "input", .. }` with the input's fit intact, iivsearch keeps the input as `current` and falls through to its normal tail. `None` without cancellation is still the error it was.

To test it honestly the scripted fitter needed the shape `Runner` actually returns — `cancel_empty_after` yields `cancelled: true` with **no** results, which is what happens when the flag is flipped mid-fit and the failed candidate is dropped rather than journalled. Tests: `a_cancellation_during_the_full_iov_fit_returns_the_input_not_an_error` and `a_cancellation_during_the_base_fit_returns_the_input_not_an_error`. Both assert the input's fit is still there and that no step ran; without the guard the search returns `Err` and the test fails on the unwrap.

**[P2] Changelog** — narrowed to `top_down_exhaustive` and `bottom_up_stepwise`, with the simultaneous divergence stated in the entry itself (#1018, ferx selects `[CL,V]+[KA]` where Pharmpy selects `[CL,V]`, the anchor asserts it). The iovsearch entry's "every proper subset" also became inaccurate with the fix above and now says what the enumeration does.

**Two things I found while in there**, both fixed with tests:

- The η/κ name lookup only covered parameters the space names, so a `block_omega` containing an out-of-space η was half-split (only the named members passed to `SplitOmegaBlock`). `Space::build` now names every parameter that carries an η/κ, and the block-building `filter_map` became a hard error rather than a silent drop.
- Behind that: the block stage's targets dropped an input block among parameters the space does not name, so the search would dissolve a correlation the modeller put there — contradicting "a parameter the space does not name is left as it is". Such a block is now carried unchanged into every target, its members are not offered to the cliques, and a note says so (`a_block_the_space_does_not_name_is_carried_unchanged_and_its_members_are_not_offered`, docs updated).

**Codecov** — the two CLI command files were the bulk of the gap (42%/44%). Both now have the in-process `run(&args)` test `ruvsearch_cmd` established: a whole search end to end, then `--quiet --resume`, plus the refusal paths (a non-canonical parameter for iivsearch; a base without `iov_column` and the defaulted space for iovsearch).

roninsightrx:
Functionality confirmed manually

Issue #1277, opened 2026-09-07
bug(nn): two nn tests have failed on every slow-tests run since #1215 — identical values, masked by #1239's timeout


TeunP:
Two `nn`-gated tests fail on **every** `Slow regression tests (nn)` run since #1215 enabled
`ferx-core/nn` in the nightly, with **bit-identical values every time**. Neither is mentioned in
#1239, and the 100-minute budget kill described there is what has been hiding them: the job
exits `124` regardless, so "the nn leg is red" has read as *one* problem when it is three.

> **Ownership.** Assigned to @roninsightrx, matching #1239 — #1215 is what put both tests in
> front of CI. The first one is #1215's own new test; cc @dominic-irx, who authored it, as on
> #1239 (not in this repo's assignable list).

## The two failures

```
nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation   src/nn/mod.rs:2958
  weight norm must be non-increasing in λ (‖W‖²: 2225.50764 → 0.00000 → 0.03844)

focei_makes_substantial_progress_on_nn_model   tests/nn_fit_convergence.rs:115
  OFV must decrease from baseline 1142.40 (got 1142.40)
```

Neither reads as a tolerance that drifted:

- **The weight norm collapses to exactly `0.00000`** at the middle λ and then comes *back up* at
  the largest one. A monotonicity check failing by rounding looks like `2225.5 → 2225.6`; this
  looks like the L2 term annihilating the weights at one λ and the fit finding something else at
  the next.
- **The OFV does not move at all** — `1142.40` against a `maxiter = 1` baseline of `1142.40`,
  after 50 FOCEI iterations. The test's own comment says it exists to *"catch 'no progress at
  all'"* (`tests/nn_fit_convergence.rs:112`), and that is exactly what it caught.

## Deterministic, not flaky — measured

Identical to the last digit across five runs, five different `main` commits, three days:

| run | commit | date | both assertions |
|---|---|---|---|
| [33966488160](https://github.com/FeRx-NLME/ferx-core/actions/runs/33966488160) | `2faeebc5` | 09-05 | identical |
| [34056609555](https://github.com/FeRx-NLME/ferx-core/actions/runs/34056609555) | `955ccef2` | 09-06 | identical |
| [34090406708](https://github.com/FeRx-NLME/ferx-core/actions/runs/34090406708) | `a2f82ed7` | 09-07 | identical |
[...]

Issue #1255, opened 2026-09-05
docs(CLAUDE.md): a fourth "green test cannot fail" shape — a shared arm whose discriminator is unasserted


TeunP:
CLAUDE.md carries a list under **"A green test is not evidence that it can fail"** — three shapes
where the *assertion* cannot observe the defect, all found on #1166. #1229 hit a fourth shape
**three separate times in one PR**, each time in code that had been written, run green and
believed. It belongs on the same list.

## The shape

> **Two inputs share one code path, and the only text that distinguishes them is unasserted.**

When one arm serves several inputs — deliberately, because the logic really is shared — every
assertion that is *about the shared logic* passes for all of them. The discriminator is whatever
small thing differs: a keyword, a name, a unit. If no assertion reads that, swapping it is
invisible, and the test's green tells you only that the shared part works.

It is the mirror of the "fast path makes the mutation unreachable" entry already on the list:
there the fixture never reaches the code; here it reaches it and the assertion looks away.

## Three instances, all on #1229

**1. Two redundant gates covered for each other.** `check_variance_init_rails` filtered scope
twice — a `coordinate_kinds` test inside the loop *and* a `decls[i].is_none()` check. Deleting
either left every test green, so the pin that Σ is out of scope could not fail. Fixed by
collapsing to one gate; the mutation then reddened.

**2. A predicate that discriminated nothing.** `is_diagonal_omega_source` was called only after
the block arm had already returned, so every declaration reaching it was diagonal and it always
answered the same way. A mutation to it sailed through. Deleted, rather than kept as a second
gate.

**3. A shared message arm naming the wrong keyword.** `block_omega` and `block_kappa` route to
one message arm — correctly; the cause and the remedy are identical — but the arm hardcoded
`` `block_omega` ``. A near-singular `block_kappa` was told to go and fix a block that appears
nowhere in the reader's file. Both `near_singular_*` tests asserted `contains("correlation")` and
`!contains(">= 1e-5")`, and **both pass on either message**. Found by reading the diff, after
preflight, the full suite and a slow-tests dispatch had all been green on it.

## What to write down

Something close to:

> **A shared arm needs an assertion on what makes the inputs different.** When one branch serves
> several inputs, every assertion about the shared logic passes for all of them, so the test tells
> you nothing about the part that differs — a keyword, a name, a unit. Assert the discriminator
> explicitly, **and assert the negative** (`!contains` the other spelling), so a value naming both
[...]

Issue #1241, opened 2026-09-05
bug(slow-tests): joint_pktte_sse_recovers_pk_and_omega fails on macOS (CL 0.912, V 8.76) and passes on Linux — third instance of #990


TeunP:
## What

`joint_pktte_sse_recovers_pk_and_omega` (`tests/tte_convergence.rs`, Tier-3) fails on macOS and
passes on Linux. It is a **third instance of the family in #990**, not a regression.

```
CL not recovered: 0.9120565597798312 (truth 1.0)     band 0.93..1.06
```

Full readout from the test's own `eprintln`, on macOS:

```
[SSE-ODE] events = 167/300
[SSE-ODE] CL=0.9121(1.0) V=8.7569(10) KA=0.9966(1) H0=0.00995(0.02; collinear)
          BETA=0.3980(0.30; collinear) omega^2=0.0737(0.09) OFV=3877.778
```

Two of the four asserted bands miss:

| parameter | macOS | band | |
|---|---|---|---|
| CL | 0.9121 | 0.93 .. 1.06 | ❌ low |
| V | 8.7569 | 9.0 .. 10.2 | ❌ low |
| KA | 0.9966 | 0.95 .. 1.08 | ✅ |
| ω²(CL) | 0.0737 | 0.07 .. 0.115 | ✅ |
| CL/V | 0.1042 | 0.095 .. 0.108 | ✅ |

## Not a regression — measured, not assumed

The test is seeded (`SEED = 20260629`) and deterministic. It produces **byte-identical** output at
every commit I ran it on:

| commit | CL |
|---|---|
| `587f808d` (parent of #1186) | `0.9120565597798312` |
| `e7bd1655` (main) | `0.9120565597798312` |
| `282c7c79` (#1223 / PR #1230, pre-merge) | `0.9120565597798312` |

So it is **not** caused by #1186 (`4ea6a024`, dose apply-once masks) — its parent fails identically
— and **not** by #1223 (`2faeebc5`), whose only production change is a fill that provably cannot
fire on this fixture: the subjects dose at `t = 0`, every TTE record carries `entry_time: 0.0`
(which the share does not push — it is gated `> 0.0`), and every event/censor time is ≥ 0, so no
`chz_time` is ever below `break_times[0]`.

I bisected specifically because #1186 was the plausible-looking suspect and would have been the
wrong answer.

## Passes on Linux

[...]

PR #1278, opened 2026-09-07
feat(api): report an absolute clock on an [odes] RHS under a steady-state dose [closes #1139]

### LINKED ISSUE #1126: bug(ode): TAD has no referent inside a steady-state dose's pre-arrival window

### LINKED ISSUE #1139: bug(ode): a steady-state dose on a model-time-reading [odes] RHS is NaN (TAD/TAFD) or ~17% wrong (T/TIME)

### LINKED ISSUE #1264: bug(ode): the SS equilibration hands the compiled [odes] RHS a bare parameter slice — TAD/TAFD unreadable (root cause under #1139)

### LINKED ISSUE #1281: bug(api): `W_STEADY_STATE_INFUSION` compares an unscaled duration, so `F` decides whether its message is true
## Summary

`W_STEADY_STATE_INFUSION` tells the user that an `SS=1` infusion with `T_inf > II` "is applied
as a single (non-SS) infusion, so the system is not at steady state at the dose time". It
decides that on `effective_duration` (`src/api/validation.rs`) — the dose's own `AMT/RATE`, or
the resolved `D{n}`/`R{n}` for a modeled dose.

The integrator decides it on something else. `equilibrate_ss_pk_state`
(`src/ode/predictions.rs:1173-1177`) compares `DoseEvent::bioavailable_infusion(f_bio).1`
against `dose.ii`, and for a **rate-defined** infusion that is `F · duration` (#419 — the rate
is fixed by the data, so `F` scales the length).

So on any model with `F < 1` on the dose compartment the two part company, and the warning's
sentence is false: the run-in does run, and the record *is* served as a steady state.

## Measured

1-cpt `[odes]` model, `CL = 1`, `V = 20`, one subject, one record `AMT = 100, RATE = 5,
II = 12, SS = 1` at `t = 480`, observations at 482 / 485. RHS
`d/dt(central) = -(CL/V)*central*(1.0 + 0.003*TAFD)`.

| `F1` | nominal `T_inf` | the integrator's `t_inf` | run-in | predictions | `W_STEADY_STATE_INFUSION` |
|---|---|---|---|---|---|
| 1.0 | 20 | `20 > 12` | skipped | `9.514379031181768`, `22.093161802029634` | fires — correct |
| 0.5 | 20 | `10 ≤ 12` | **runs** | `NaN`, `NaN` | fires — **wrong** |

The `NaN` is what makes the second row a measurement rather than a reading of the code:
`TAFD` has no referent inside the run-in (#1139), so a `NaN` prediction can only come from the
equilibration the warning says did not happen.

## Impact

Low. A false statement in a warning; no wrong numbers come from the warning itself. It matters
because the sentence is the one a user acts on — "this record is not at steady state" — and on
the `F < 1` row it is.

## Scope note — this is *not* an inconsistency with #1139 T3

`W_STEADY_STATE_ABSOLUTE_TIME` (#1139 T3) deliberately does **not** share `effective_duration`,
for exactly this reason: it mirrors `equilibrate_ss_pk_state`'s `bioavailable_infusion`
comparison instead. Borrowing the unscaled one suppressed it on precisely the row that needs it
(measured: `F1 = 0.5` gave `NaN` predictions and no warning). The two warnings therefore both
fire on the `F < 1` row today. That is the new one being right and this one being wrong, not a
regression in either.

## Fix

Compare the same quantity the run-in uses. `check_model_data_warnings` already resolves it —
`resolved_dose` (added by #1139 T3) hands back both the resolved `DoseEvent` and `f_bio`, so the
predicate is `resolved.bioavailable_infusion(f_bio).1 > d.ii`.

## Validation

No NONMEM anchor applies: NONMEM has no equivalent warning, and the question is which of ferx's
own predicates the message describes. Validate with:

1. **the straddle above** as a Tier-1 test — the `F = 0.5` row must stop warning, and the
   assertion has to be paired with the predictions so it pins agreement with the integrator
   rather than with itself;
2. **the negative that must not regress** — `F = 1` with `T_inf > II` must still warn;
3. **a duration-defined (`RATE = -2`) row**, where `F` scales the *rate* and leaves the duration
   alone, so the fix must **not** change its answer. This is the arm a naive "multiply by `F`"
   fix breaks, and the reason the fix has to go through `bioavailable_infusion` rather than
   through a hand-written product.

## Related — the same shape, one subsystem over

- **#419** — the convention itself (NONMEM scales the duration, ferx scaled the rate). This is
  where `bioavailable_infusion` came from; the defect here is a consumer that never adopted it.
- **#920** — `equilibrate_ss_g`, the analytic **sens** walk, builds its synthetic per-cycle dose
  with a raw `amt/rate` window and so drops the same F reshaping. Different path, different
  symptom (a gradient that disagrees with its value path rather than a false sentence), same
  root: a site that predates #419 and was never reconciled with it. Filed the same way, off a
  review of an adjacent PR.
- #1139 / #1258 T3 (PR #1278) — the gate whose measurement surfaced this
- #1280 — `predict()` / `simulate()` carry no model/data warnings at all

Worth a sweep rather than three point fixes: `bioavailable_infusion` has a small, enumerable
set of callers, and the two defects found so far were both *non*-callers that hand-compute a
window instead.

TeunP:
## Why

#1139: a steady-state dose on an `[odes]` right-hand side that reads model time is
either `NaN` or silently far from the model's own dose train, depending on which of
three spellings the RHS uses. Batch T (#1258) split it by spelling:

| | spelling | outcome | where |
|---|---|---|---|
| T2 | `TAD` per window, plus the absolute-clock anchors | `[closes #1264] [refs #1139]` | `955ccef2` (#1270) |
| T4 | `TAD` | a referent inside a lagged SS dose's pre-arrival window `[closes #1126]` | `0596fbb9` (#1274) |
| **T3 (this PR)** | `TAFD`, `T` / `TIME` | **the remaining half — what #1139 itself is still open for** | — |

The mechanism is one thing, not three. The run-in that stands in for the infinite
past does not integrate on the subject's clock: it expands the periodic train on a
clock **local to each cycle**, because that is the only clock on which the one-cycle
map is the same map every cycle. `TAD` is bounded inside one dosing interval, so the
train has a periodic limit and T4 anchored it. An absolute clock has no such limit —
there is nothing for the run-in to converge to.

They fail differently, and the diagnostic has to say so:

* `T` / `TIME` return a **finite** number that matches NONMEM's own `SS=1` record,
  because both engines equilibrate cycle-locally and then integrate forward on
  absolute time. Both sit the same distance from their own non-convergent train.
* `TAFD` has **no referent** inside the run-in — the periodic fiction has no first
  dose — so it reads `NaN`, and `0.0 * NaN` is `NaN`, so even an inert term poisons
  the objective.

Decision on record (Teun, 2026-09-06): **warn, keep NONMEM parity.** This PR
implements that and moves no value.

## What changed

**1. `src/parser/model_parser.rs` — two narrow flags, one composed accessor.**

`pk_reads_tafd` and `pk_reads_solver_time` alongside the existing
`pk_reads_model_time`, composed into `pk_reads_absolute_time()`. Two bools rather
than one because the diagnostic says something different about each spelling.
`TAD` is deliberately excluded — it is the one spelling that *does* have a periodic
steady state.

Same commit collapses #1166's duplicated `chz_state_slots.is_empty()` fast path into
a single filtered `pk_view`, so the `__chz` exclusion has one implementation rather
than two. The saving that branch bought is gone regardless: `uses_time_vars` unions
all three time slots and cannot be decomposed after the fact, so the two new flags
have to walk the statement list on every model anyway.

**2. `src/api/validation.rs` — `W_STEADY_STATE_ABSOLUTE_TIME`.**

Non-fatal, in `check_model_data_warnings`. Fires when the `[odes]` PK block reads an
absolute clock **and** at least one subject has a periodic SS dose. Message names
which spelling(s) and appends that spelling's own consequence; suggestion points at
explicit dose records or `TAD`.

**3. No value moves.** `TAFD` still `NaN`; `T` / `TIME` still match NONMEM. That is
the assertion, not a side effect — see *Numerical validation*.

### Why a warning and not an error

Four reasons, in the order they actually decide it:

1. One mechanism ⇒ one diagnostic. Splitting `TAFD` (error) from `T`/`TIME` (warning)
   would encode "NaN is worse than wrong" as a taxonomy, when both are the same
   defect at different amplitudes.
2. **An error would protect `fit()` only.** Grepped, not recalled: the sole production
   callers of either bundle are `api/fit.rs:550` (`check_model_data_rule`),
   `api/fit.rs:1216` (`check_model_data_warnings`) and `api/validation.rs:4965`/`:4971`
   (the `ferx check` path). `predict()` and `simulate()` call neither, and would go on
   serving the same `NaN` unwarned. A warning surfaced by `fit()` *and* `ferx check`
   beats an error one entry point enforces.
3. The recorded decision above.
4. An error is harder to walk back than a warning. T2 added `E_SS_LAGTIME_TAD_RHS`
   and T4 deleted it two PRs later; that round trip cost more than a warning would
   have.

### Why `WarningCode::DataQuality` rather than a new variant

The message carries the substring `"SS=1 dose"`, which is what routes it through
`classify_warning` into `DataQuality` — the bucket the neighbouring steady-state
warnings already use. That substring is **load-bearing, not incidental phrasing**,
and a test pins it so a reword cannot silently drop the finding into the `general`
fallback.

A new `WarningCode` variant was rejected for two concrete costs and no benefit.
`docs/warnings.qmd`'s `warning-codes` section is in the R1 baseline at **7391 chars**
against a 5000 cap — the baseline is a shrink-only ratchet, so a new row there fails
`docs-lint` until the section is split, which is a different PR. And ferx-r keys its
remediation guidance off the **category** vocabulary, not the code:
`R/diagnostics.R:694` already maps `data_quality` to "Data issue detected. Review the
flagged observations in the dataset." Hence **no ferx-r PR**.

### Scope: the PK block, not the whole RHS

The predicate asks whether the **PK block** reads an absolute clock, with `__chz`
derivative lines filtered by slot (#1166's rule). Justified by the run-in itself:
`mask_chz` zeroes the cumulative-hazard accumulator's derivative during
equilibration and `restore_chz` writes back the record snapshot, so a time-reading
hazard cannot disturb PK equilibration. `nonmem_anchor/ss_chz_tdep_fit.ferx` is
exactly that shape — a NONMEM-anchored `TIME`-reading hazard under `SS=1` — and must
stay silent. It does; asserted, with a straddle (the same joint model with the time
read moved into the PK block *does* warn).

**Known over-report, inherited from #1166 and not fixed here.** An `[odes]`
*intermediate* that reads time and is consumed only by the hazard —
`TT = TIME`, then `hazard = f(TT)` — is not one of the injected derivative lines the
slot filter removes, so such a model is warned about although its PK block is
autonomous. It is a false positive in the safe direction (a warning, on a model that
runs fine). Pinned by a test so it is a known quantity rather than a surprise, and
documented in a `callout-note` on the steady-state page. Fixing it means widening
#1166's filter to trace intermediate dataflow, which is a different change.

### The FD gate is **not** widened

`pk_reads_absolute_time()` is explicitly **not** a gradient-routing predicate. The SS
gates in `sens/ode_provider.rs` keep asking the wide `reads_model_time`, so a
`TAD`-reading SS model still routes to finite differences and never reaches the dual
equilibration. **#1272 stays latent and unchanged by this PR** — nothing here makes
it reachable, so it does not ship with this change.

Per the `Dual2`-vs-FD rule, that is discharged here by **routing tests** rather than a
parity test, the same way T2 and T4 discharged it: no fixture in this PR reaches the
dual SS path, and the flags this PR adds are not consulted on it.

## Alternatives considered

* **Hard error on `TAFD`.** Rejected — the four reasons above; #2 is decisive.
* **Give `TAFD` the run-in's own origin so it returns a finite number.** Rejected: it
  silently redefines `TAFD` as `TAD` for exactly the models that spelled them
  differently on purpose, and NONMEM has no such value either, so it would trade a
  loud `NaN` for a quiet divergence.
* **A new `WarningCode` variant.** Rejected — R1 baseline cost plus a forced ferx-r
  PR, for a category users already have.
* **Suppress when a neighbouring steady-state warning already fired.** Rejected:
  `W_STEADY_STATE_INFUSION` is a typical-value heuristic by its own comment (a modeled
  SS infusion may overlap on some occasions only), so suppressing on it would drop a
  true finding.

## Cross-repo dependency

No new `WarningCode` variant, no public API change: ferx-r's category-keyed warning
routing already handles this via `data_quality`.

## Breaking changes

- [x] None

No new `pub` items — the three accessors are `pub(crate)`. `tools/preflight.sh` reports
"public API matches the committed baseline", so `api/ferx-core-public-api.txt` is unchanged.

## Numerical validation

**The validation is that nothing moved.** Measured this session at branch HEAD:

**The validation is that nothing moved**, and T2 left the instruments in place to say so.
`absolute_time_ss_still_matches_nonmem` records `8.5679354785102` at t = 482 as
bit-identical across T4; this PR must leave it bit-identical again.

| `[odes]` term under `SS=1` | ferx | note |
|---|---|---|
| `0.003*T`, `0.003*TIME` | 8.5679354785102 | matches NONMEM's `ss_tabs` record; realised **2.615672e-9** worst relative, bound 1e-8 |
| same model, explicit 41-dose train (`train_tabs`) | 5.1221001380354 | **67 %** away, and the two engines are 1.4e-9 apart *about the size of the gap* |
| `0.003*TAFD`, `0.0*TAFD` | `NaN` | no referent inside the run-in |
| `0.0*TAD` | the autonomous steady state, to **4.633876e-12** | the discriminator |

**"Nothing moved" is measured differentially, not argued.** The three `src/` files this PR
touches were reverted to `0596fbb9` in place, the same anchor re-run against the same test
file, and all four predictions come back with identical bit patterns:

```
branch     482 -> 8.5679354785102 (4621012936933302506)   worst_rel = 2.615672e-9
           485 -> 5.9325361489014 (4618365460324776456)
           488 -> 4.1022138965935 (4616304700671411034)
           491 -> 2.8327608538595 (4613561229162953034)
0596fbb9   ...identical, all four, and worst_rel = 2.615672e-9
```

That matters because `pk_view` is not inert: mutation P5 (always filter) reddens 19 tests,
14 of them pre-existing `pk::modified_release` / `sens::provider` routing tests. The
refactor was in a live path, so the no-op claim needed measuring rather than reasoning.

**An incidental finding, recorded not silently fixed.** T2's comment records the realised
error as `1.528e-9` with "~6.5x headroom". On this machine it is `2.615672e-9` — 3.8x
headroom — **before and after** this change alike. Not a regression and not in ferx: the
predictions are bit-identical to `0596fbb9`, so what differs is the machine, which is the
adaptive solver's platform-to-platform step drift the same file already notes for
#990 / #1241. A note to that effect is added beside the figure, and the docs table now
quotes the bound (`better than 1e-8`) rather than a machine-specific realised value.

**What the user saw before.** Measured, not recalled: the `0.003*TAFD` SS fit returned five
warnings. The only two describing the failure were `W_ODE_SOLVER_DIAGNOSTICS` (`14400
step(s) clamped at the minimum step size`, 225 freeze-padded segments) and `Covariance step
failed: base OFV is non-finite at convergence` — both pointing at the integrator. The other
three fire regardless (finite-difference inner gradients, thread-count hint). `ferx check
--data` on the same model returned **zero** diagnostics before, and **exactly one** now.

The gap follows from the two values above:
`(8.5679354785102 − 5.1221001380354) / 5.1221001380354 = 0.6727`.

The parity is a **guard, not a claim of correctness** — T2's phrasing, kept. `T`/`TIME`
matching NONMEM does not make either right; it makes them wrong together, which is
precisely why this PR reports rather than repairs.

- [x] Compared against: NONMEM, via T2's committed streams re-run on this branch
- [x] Not applicable in the "estimates changed" sense — this PR is a diagnostic

## Performance

Both new flags are computed once, at model-compile time, over the statement list.
#1166's `chz_state_slots.is_empty()` fast path is removed, but the walk it skipped is
now unavoidable regardless: `uses_time_vars` unions all three time slots and cannot
be decomposed after the fact. The warning check itself is one pass over subjects,
once, and only when the flag is set.

## Tests

**Tier 1 — `src/api/tests/ss_absolute_time_tests.rs` (8 tests)**

* every absolute-clock spelling warns **and names its own consequence** — `T`, `t`,
  `TIME`, `TAFD`, and both together; positive *and* negative assertions per spelling
  (shared-arm rule, #1255)
* a `TAD`-reading RHS does **not** warn — `0.03*TAD`, `0.0*TAD`, autonomous; plus
  `TAD` **and** `TIME` together still warns
* an SS dose with `II = 0` does not warn (the `has_periodic_ss_dose` discriminator)
* an analytical model does not warn
* the finding is a warning that counts subjects — 3 → `"3 subject(s)"`, mixed →
  `"2 subject(s)"`; severity, block, suggestion; and **absent from the fatal bundle**
* the message reaches `DataQuality` through its `SS=1 dose` phrase — asserted, then
  asserted that removing the phrase *changes* the classification
* a time-reading hazard alone does not warn — with a straddle: the same joint model
  with the read moved into the PK block **does** warn
* a time-reading intermediate used only by the hazard still warns — the inherited
  #1166 over-report, pinned deliberately

**Tier 1 — `src/parser/model_parser_tests.rs`**

Flag tuples widened (3→4 and 4→5 entries) so every existing row states its
`pk_reads_absolute_time` expectation; new test asserting the union equals the OR of
its two halves and then `(tafd, solver)` per spelling.

**Tier 2 — `tests/ss_model_time_nonmem_anchor.rs` (3 tests)**

* `TAFD` stays `NaN` under an *un*-lagged SS dose while `TAD` does not — the
  **asymmetry**, with a length check so an empty prediction vector cannot pass.
  Worth flagging to a reviewer: this half was written as a `to_bits()` equality and the
  full-suite run caught it. `0.0*TAD` and `1.0 + 0.0` are the same function but not the
  same *route* — `pk::model_uses_time_anywhere` gates on `pk_reads_model_time()`, so the
  inert `TAD` twin takes the model-time walk and the autonomous one does not. They agree to
  **4.633876e-12**, bounded at 1e-10 (~22x), matching the bound its lagged sibling uses.
  The stricter form was carried over from that sibling *from memory*; the sibling does not
  assert bits either.
* the warning arrives through `fit()` — exactly one, right consequence per spelling,
  `TAD` twin silent
* `ferx check` reports it **without failing the report** — `report.valid` still true

**Two stale doc comments corrected in passing**, both in files this PR already edits:
`tests/ss_lagtime_tad_rhs_accepted.rs` said `TAFD` under SS was "deliberately untouched
here … that is #1139's remaining half, T3" (T3 is this PR), and
`tests/ss_prearrival_tad_nonmem_anchor.rs`'s header said the same. Neither test changes
behaviour — the first reads the **fatal** bundle and structurally cannot observe a warning,
which is now stated there rather than left to look like luck.

**Deliberate change to an existing scope pin.** `tafd_stays_nan_under_a_lagged_ss_dose`
(#1258 T4's boundary marker) kept its behaviour, because T3 chose to move no value —
it gained a reason and a pointer to the new un-lagged twin instead of being deleted.
While there, a **vacuous assertion in it was fixed**: `all()` on an empty iterator is
`true`, so the test passed for a fixture that produced no predictions at all. Now
asserts `!got.is_empty()` first.

**Mutation testing.** 12 mutations across the parser flags and the validation gate, each
run against the full `--lib` suite under `ci,survival` with `--no-fail-fast`, restored
from byte-copy snapshots with an mtime touch. **12 killed, 0 survivors.**

| | mutation | killed by |
|---|---|---|
| P1 | `pk_reads_tafd` reads `time_slot` — the two halves stop being distinguishable | 6 |
| P2 | the union swallows `TAD`, i.e. becomes `pk_reads_model_time` | 5 |
| P3 | drop the `stmts_read_time_builtin` disjunct — bare `TIME` stops being seen (#1124) | 8 |
| P4 | `pk_view` never filters — a time-reading hazard reads as a non-autonomous PK block | 4 |
| P5 | `pk_view` always filters — a non-joint model sees an *empty* program | 19 |
| P6 | the union is `&&` rather than `\|\|` | 11 |
| V1 | the gate asks the wide, TAD-inclusive predicate | 1 |
| V2 | the `II > 0` conjunct dropped (`has_ss_doses()`) | 1 |
| V3 | both consequence sentences key on `TAFD` | 1 |
| V4 | the diagnostic is never pushed | 7 |
| V5 | the message loses its `SS=1 dose` routing phrase | 2 |
| V6 | the spelling list mislabels `TAFD` | 1 |

Two things the matrix caught that a green suite would not have:

* **P5 killed 19**, of which 14 are pre-existing (`pk::modified_release::*`,
  `sens::provider::*`, `sens::ode_provider::*`). The `pk_view` refactor is load-bearing
  for the `TAD` routing this PR does not otherwise touch, and the existing suite holds
  it — which is the evidence that collapsing #1166's duplicated branch was safe.
* **V2 was initially SKIPPED, not killed**: its pattern occurred twice, because the
  pre-existing `W_SDE_STEADY_STATE` block filters on the same predicate. The harness
  refused to mutate ambiguously rather than mutating the wrong site; re-run against the
  `let n_ss =` binding.

The harness's own tree check was also wrong and was fixed mid-run: it grepped for each
mutation's replacement text as a must-be-absent list, and two of those strings are
substrings of genuine production code, so a clean tree reported dirty. The mirror
failure — a replacement that is a substring of the original reporting clean while the
tree stays mutated — never fired but was equally live. Replaced with a byte comparison
against the snapshot; the final tree is byte-identical to the pre-mutation snapshot.

## Example

- [x] Not applicable — a diagnostic on an existing combination; no new API surface,
      no new DSL syntax, no new `[fit_options]` key.

## Docs

* `docs/model-file/steady-state.qmd` — the `## Steady state with TAFD or TIME`
  section **rewritten** (not appended to): the measured table, why the NONMEM parity
  is kept deliberately, the warning's name, and the `callout-note` on the joint
  PK-TTE over-report.
* `docs/file-formats/check-report.qmd` — one row for the new code under
  `#### Dose records`.
* `docs/model-file/ode-models.qmd` — a **stale** claim corrected: it said `SS=1` plus
  a lagtime on a `TAD`-reading RHS "is rejected", naming `E_SS_LAGTIME_TAD_RHS`, which
  T4 deleted and which exists nowhere in `src/`.
* No new pages, so no `_quarto.yml` change. `docs-lint --check` clean, including R4
  on the three anchors this adds.

**A second stale claim corrected while rewriting that section.** It read *"an explicit
`TAFD` train moves 29 % per doubling of its length"*. The 0.294 figure it cites is
measured by `the_tad_train_converges_and_the_absolute_time_train_does_not` from
`train_tabs` vs `train_tabs_n21` — the **`T`-absolute** train. There is no `TAFD` train
fixture, and there could not usefully be one: the number separates the two halves of
#1139 by *spelling*, so attaching it to the wrong spelling inverted what it shows. Now
reads "an explicit absolute-clock train".

## Checklist

- [x] `Dual2` path untouched — see *The FD gate is not widened*
- [x] No version bump (no breaking change)

Full suite, both PR feature sets, reconciled against `0596fbb9` (179 binaries / 4893 passed
/ 0 failed / 171 ignored):

| feature set | binaries | passed | failed | ignored |
|---|---|---|---|---|
| `ci` | 179 | 4903 | 0 | 171 |
| `ci,survival` | 179 | 5235 | 0 | 171 |

The first run of that suite was **not** green: it caught the `to_bits()` over-assertion
described under *Tests*. That is the failure mode worth naming — the Tier-2 test had been
written, reviewed and believed, and only the full run exercised it.

## Docs & examples

No new DSL syntax, no new `[fit_options]` key, no new example or data file, so no
ferx-site or ferx-book change is required.

- [x] No example execution step needed

## Reviewer hints

**Focus here:**

1. **The predicate's scope** — `pk_reads_absolute_time()` excludes `TAD` on purpose,
   and is deliberately *not* the predicate the FD gates ask. If those two facts are
   wrong, everything else is.
2. **The `"SS=1 dose"` substring** in the message is load-bearing for
   `classify_warning`. It looks like prose. It is not.
3. **The `#1166` over-report** — accepted knowingly, pinned by a test, documented.
   Worth a second opinion on whether accepting it is right.

**Skimmable:** the flag-tuple widening in `model_parser_tests.rs` is mechanical (every
existing row gains one expectation); the docs table restates numbers the anchor tests
already assert.

## Review rounds

**Round 1 → `5e4bb564`.** Nine findings, eight of them applied as reported: the consequence
sentence is no longer stated where the run-in is skipped, the `[diffusion]` rationale is
measured rather than assumed, two duplicated statement walks are gone with
`pk_reads_model_time` *derived* from the narrow flags, the docs no longer claim anything about
NONMEM's `TAFD` (there is no control stream for it), "Since ferx 0.3" is corrected, the
changelog's "describing the failure" qualifier is restored, and the subject count can no longer
be read as a dose count. The `simulate()` scope call was filed as #1280.

**Round 2 — the ninth finding's fix had a defect, and it is the one this PR exists to prevent.**
Filtering to "doses that actually reach the run-in" compared `effective_duration` — the
record's own `AMT/RATE`. `equilibrate_ss_pk_state` compares
`DoseEvent::bioavailable_infusion(f_bio).1`, which is mode-aware (#419). Measured on the
`0.003*TAFD` fixture, `AMT = 100, RATE = 5, II = 12`:

| `F1` | the integrator's `t_inf` | run-in | predictions | warning at `5e4bb564` |
|---|---|---|---|---|
| 1.0 | `20 > 12` | skipped | `9.514379031181768`, `22.093161802029634` | none — correct |
| 0.5 | `10 ≤ 12` | **runs** | `NaN`, `NaN` | **none — the silent case this PR ends** |

`TAFD` has no referent inside the run-in, so the `NaN` can only come from the equilibration the
filter said had not happened. That is the measurement; nothing here is read off the code.

**Round 3 — `W_STEADY_STATE_INFUSION` had the same blindness, and it is folded in (closes
#1281).** On the `F1 = 0.5` row it told the user the record is "applied as a single (non-SS)
infusion" while the run-in demonstrably ran. Both findings now ask one predicate,
`run_in_skips_overlapping_infusion`, over one `resolved_dose` closure, so they partition rather
than contradict and no second spelling of the rule survives. `effective_duration` is gone.

`the_run_in_bail_out_follows_bioavailability` asserts three arms and, on each, the predictions
the same model returns — pinning agreement with the integrator rather than with the gate:

| dose | bioavailable `T_inf` vs `II = 12` | predictions | `…INFUSION` | `…ABSOLUTE_TIME` |
|---|---|---|---|---|
| `RATE = 5`, `F1 = 1.0` | `20 > 12`, skipped | `9.5144`, `22.0932` | fires | silent |
| `RATE = 5`, `F1 = 0.5` | `10 ≤ 12`, **runs** | `NaN`, `NaN` | silent | fires |
| `RATE = -2`, `D1 = 20`, `F1 = 0.5` | `20 > 12`, skipped | finite | fires | silent |

The third arm is the one a hand-written `F * duration` breaks — it would read `0.5 · 20 = 10 ≤
12` and flip both verdicts. `tests/modeled_duration.rs` covers `RATE = -2` only at `F = 1`,
where a wrong scaling is invisible.

## Mutation testing

11 mutations across the two rounds, **zero survivors**, each restore verified by byte
comparison against a snapshot rather than by a grep list:

| | mutation | dies in |
|---|---|---|
| M1 / M8 | drop the `F` scaling | `the_run_in_bail_out_follows_bioavailability` |
| M9 | hand-written `f_bio * duration` | the same test's `RATE=-2` arm only |
| M2 / M10 | the gate stops honouring the bail-out | 2 tests |
| M11 | the infusion warning reverts to a bare `is_infusion()` | Tier 1 **and** 2 tests in Tier 2 |
| M3 | drop the out-of-range-`CMT` bail-out | `a_dose_that_never_reaches_the_run_in_is_not_reported` |
| M4 | drop the `II > 0` conjunct | `a_steady_state_dose_with_no_interval_does_not_warn` |
| M5 | count doses, not subjects | `the_finding_is_a_warning_that_counts_subjects` |
| M6 | `effective_duration` ignores the modeled resolution | **Tier 2** — `modeled_duration_steady_state_overlap_warns` |
| M7 | derived `pk_reads_model_time` drops its `TAD` walk | **10** tests across 4 modules |

M6 is the reason every round-3 mutation runs both scopes: it read as a survivor under a
`--lib`-only matrix, and its kill lives in Tier 2, which does run on PRs.

## Validation — why there is no NONMEM anchor for the round-2/3 change

The change decides which **warning text** is emitted; it moves no number. NONMEM has no
equivalent diagnostic, and the quantity in question — "does ferx's run-in serve this dose" — is
ferx's own routing decision, which is the case CLAUDE.md covers under *a quantity with no NONMEM
spelling*. The convention underneath **is** anchored: #419 established that NONMEM scales the
duration on a rate-defined infusion, and this fix routes through `bioavailable_infusion`, that
convention's single implementation, so no formula is anchored from scratch — the "reduction to
an already-anchored term" case.

The oracle that does apply is the integrator itself, and the test uses it: every arm asserts the
predictions alongside the warning, so a gate that agreed with a wrong answer by construction
would fail.

`slow-tests` was dispatched on this branch (run `34151399470`) and compared to run
`34097734820` on the base commit **per job and by content**: `core` succeeded in both; `nn`
failed in both with the same two panics
(`nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation` at `src/nn/mod.rs:2958`
and `focei_makes_substantial_progress_on_nn_model` at `tests/nn_fit_convergence.rs:115`) and the
same `exceeded its 100m budget … exit code 124`, with 164 `test result:` lines and zero
`... FAILED` names in each. Pre-existing — #1239 / #1277.

Closes #1139. Closes #1281.


TeunP:
## Review — xhigh, measured on the branch

Everything below was run against `4e62e3f4` in the worktree: the `ferx` CLI built from this
branch, `tools/preflight.sh`, the full `--lib` suite and the five touched Tier-2 anchor suites.
The reviewer hints were the right three things to look at, and two of them hold. The third —
"the predicate's scope" — is where the defects are, but not on the axis the PR expected: the
`TAD` exclusion and the FD-gate separation are both correct; what is wrong is the assumption
that a periodic `SS=1` record always reaches the run-in.

### What the measurements say

| probe | result |
|---|---|
| `0.003*TAFD`, `SS=1` bolus, `II=12` | `OFV NaN` — the warning is right |
| `0.003*TAFD`, `SS=1` **infusion** `AMT=100, RATE=5, II=12` (`T_inf = 20 > II`) | **`OFV 367.2851`, finite** — the warning still said "the objective is non-finite" |
| same model + `[diffusion] central ~ 0.01` | `OFV NaN` — the warning is right, but for a different reason than the comment gives |
| `0.003*TAFD` explicit train, n=21 vs n=41, per cycle phase | **22.7 % / 30.6 % / 37.7 % / 44.1 %** |
| `CL = TVCL*(1.0 + 0.003*TIME)`, autonomous `[odes]`, SS vs 41-dose train | 0.2026 %, **no warning** |

### Correctness

**1. The `TAFD` consequence sentence is false wherever the run-in is skipped.**
`equilibrate_ss_pk_state` returns before integrating anything when
`is_real_infusion(dose) && t_inf > dose.ii` (`ode/predictions.rs:1175`) and when
`cmt_idx >= n_states` (`:1160`), so the ordinary finite TAFD anchor stays in place. On the
overlapping-infusion row above the fit comes back at **OFV 367.2851** and prints
`W_STEADY_STATE_INFUSION` — correctly saying the record was served as a single non-SS
infusion — immediately followed by this PR's warning saying it "reads NaN and the objective
is non-finite". The mechanism sentence ("The run-in ... expands the periodic dose train") is
equally counterfactual there.

The *Alternatives considered* entry rejects "suppress when a neighbouring steady-state warning
already fired", and that reasoning is right as far as it goes — `W_STEADY_STATE_INFUSION` is a
typical-value heuristic and blanket suppression would drop true findings. But the question the
gate has to ask is not "did another warning fire", it is "does this dose reach the run-in".

**2. The suggestion recommends an operation with no defined answer.** "Replace the SS=1 record
with explicit dose records covering the run-in" — but for an absolute clock the train has no
limit, which is the warning's own opening premise. Measured on `0.003*TAFD`, a 21-dose and a
41-dose train differ by 22.7–44.1 % at the four sample phases. `docs/model-file/steady-state.qmd`
gives the same advice and states the 29 %-per-doubling figure two paragraphs above it without
connecting them.

**3. The `[diffusion]` comment states something measurement disproves.** `validation.rs:3893`
justifies the gate's scope with "under `[diffusion]` the EKF never reads `dose.ss` (#1260) so
the objective skips the run-in". It does not skip it: `ode_predictions_ekf_with_diffusion`
computes the Kalman `R` from a standard `ode_predictions` pass, and that one runs the
equilibration, so the `NaN` reaches the likelihood anyway — measured `OFV: NaN`,
`Covariance: FAILED`. The conclusion (fire on `[diffusion]` too) is right; the stated reason
would mislead anyone narrowing the gate later.

### Simplification

**4. Three of the seven statement walks in `build_ode_spec` are duplicates.** When
`chz_state_slots.is_empty()`, `pk_view` *is* `&stmts_owned`, so
`stmts_read_slots(pk_view, &[time_slot, tafd_slot, tad_slot])` is bit-for-bit the call already
stored in `uses_time_vars`, and `stmts_read_time_builtin(pk_view)` the one already in
`rhs_reads_time_builtin`; independently of that, the built-in walk runs twice (`:11796` and
`:11812`). The Performance section's "that saving is gone regardless" is not what the code
does. Binding the built-in walk once and *deriving* `pk_reads_model_time` from the two new
flags plus a `tad_slot` walk drops two walks and makes the containment relation
`pk_reads_absolute_time`'s doc comment asserts unfalsifiable rather than restated.

### Docs / changelog

5. `steady-state.qmd:415` — "`TAFD` has no such convention **in either engine** ... the solve
   returns `NaN`" is a claim about NONMEM with nothing behind it: `nonmem_anchor/` has
   `ss_tad*`/`train_tad*` and `ss_tabs*`/`train_tabs*` and no `TAFD` stream, as the PR
   description itself notes. The row directly above it in the same table is anchored to 1e-8.
6. `steady-state.qmd:420` — "Since ferx 0.3" names a released line that cannot emit the
   warning (0.3.0 and 0.3.1 are tagged; the changelog entry is under `[Unreleased]`). Line 482
   of the same file already uses the precise form for another unreleased change.
7. `CHANGELOG.md:542` — "previously the only warnings on such a fit were a
   `W_ODE_SOLVER_DIAGNOSTICS` and a failed covariance step" drops the "describing the failure"
   qualifier the test module's own header carries; the fit emits five (six here). A reader can
   disprove it by running the model.

### Test hole

8. `the_finding_is_a_warning_that_counts_subjects` cannot distinguish the subject count it
   names from a dose count — every fixture subject carries exactly one dose, so mutating the
   implementation to `flat_map(|s| &s.doses).filter(...).count()` gives the same 3 and the
   same 2 and survives.

### Scope, documented rather than changed

9. The gate reads the `[odes]` statement list only, so `TIME` read from
   `[individual_parameters]` — which the DSL allows — is out of scope. Measured on
   `CL = TVCL * (1.0 + 0.003*TIME)`: no warning, and the `SS=1` answer sits 0.2026 % from the
   same model's 41-dose train. Much smaller than the 67 % this warning exists for, and the
   mechanism is different (a frozen per-record snapshot, exactly NONMEM's `$PK`), so warning
   there is probably wrong — but the boundary was invisible to the reader.

### Not applied — a scope call for you

The design rationale's decisive argument is #2, "an error would protect `fit()` only —
`predict()` and `simulate()` call neither, and would go on serving the same `NaN` unwarned".
That is still true of the warning: `check_model_data_warnings` has exactly two production
callers (`api/fit.rs:1216`, `api/validation.rs:4876`), so `simulate()` on an `0.003*TAFD` SS
model returns all-`NaN` rows with no diagnostic. `simulate_with_options_diag` already takes
`model`/`population`/`params` and already extends `SimulationOutput.warnings` from another
source, so wiring it in is mechanically small — but it would push the *whole* bundle into
every simulate caller's output, and picking one code out of the bundle instead is the
shared-infrastructure special case worth avoiding. Left for a follow-up issue.

### Changes applied — `5e4bb564`

`CHANGELOG.md`, `docs/file-formats/check-report.qmd`, `docs/model-file/steady-state.qmd`,
`src/api/tests/ss_absolute_time_tests.rs`, `src/api/validation.rs`,
`src/parser/model_parser.rs` — +157 / −38.

The substantive one: `n_ss` now counts per dose that actually reaches the run-in, filtering on
the *same* typical-value `effective_duration` `W_STEADY_STATE_INFUSION` uses, so the two
findings partition rather than contradict, plus the `cmt_idx < n_states` bail-out. New Tier-1
test `a_dose_that_never_reaches_the_run_in_is_not_reported`, both rows straddled (`RATE = 20`
→ `T_inf = 5 ≤ II` warns; the in-range compartment warns). Each conjunct was mutated
**separately** and each killed the test on its own row — the combined mutation would have
credited the CMT half with the infusion half's failure.

Verification: `--lib` 4347 passed / 0 failed; `ss_model_time_nonmem_anchor`,
`ss_prearrival_tad_nonmem_anchor`, `ss_lagtime_tad_rhs_accepted`, `ss_chz_nonmem_anchor`,
`modeled_duration` all green (71 tests); `preflight OK — fmt check clippy rustdoc docs
public-api`. End-to-end re-measured after the fix: the infusion case emits only
`W_STEADY_STATE_INFUSION` at OFV 367.2851, the bolus case still warns at OFV NaN, the
`[diffusion]` case still warns at OFV NaN.

### Answers to the two open questions

1. **File the follow-up.** #1279 is open and the docs already point at it; the callout now
   also records the `[individual_parameters]` boundary, which is the second half of the same
   "what does this predicate actually cover" question.
2. **`closes #1139` reads right to me.** T2 and T4 moved `TAD`; the `T`/`TIME` half is
   answered by a recorded decision not to move the value, and reporting it is what that
   decision looks like in the product. Nothing is left open on #1139 that is not #1279.

TeunP:
Follow-up filed: #1280 — `predict()` and `simulate()` carry no model/data warnings. That is the "Not applied" item above; it is on #1217 at step 7, next to #898.

TeunP:
## Round 2 / 3 — `5e4bb564`'s fix had a defect, and `W_STEADY_STATE_INFUSION` had its mirror

Correcting one claim in the review above: the round-1 fix said filtering on "the *same*
typical-value `effective_duration` `W_STEADY_STATE_INFUSION` uses" makes the two findings
partition. It did the opposite, because that is **not** the comparison the run-in makes.
`equilibrate_ss_pk_state` compares `DoseEvent::bioavailable_infusion(f_bio).1`, which is
mode-aware (#419): on a rate-defined infusion `F` scales the *length* (the data fixes the
rate); on a duration-defined one (`RATE=-2` → `D{n}`) it scales the rate and the length is
untouched.

Measured on the `0.003*TAFD` fixture, one subject, `AMT = 100, RATE = 5, II = 12`:

| `F1` | the integrator's `t_inf` | run-in | predictions | warning at `5e4bb564` |
|---|---|---|---|---|
| 1.0 | `20 > 12` | skipped | `9.514379031181768`, `22.093161802029634` | none — correct |
| 0.5 | `10 ≤ 12` | **runs** | `NaN`, `NaN` | **none — the silent case this PR exists to end** |

`TAFD` has no referent inside the run-in, so the `NaN` can only come from the equilibration the
filter said had not happened. The same `F1 = 0.5` row also had
`W_STEADY_STATE_INFUSION` telling the user the record is "applied as a single (non-SS)
infusion" — false for the same reason, in the other direction.

Both now ask **one** predicate, `run_in_skips_overlapping_infusion`, over one `resolved_dose`
closure. `effective_duration` is deleted; there is no second spelling of the rule left to
drift. That closes #1281, filed for the second half before folding it in here.

`the_run_in_bail_out_follows_bioavailability` asserts three arms **and, on each, the
predictions the same model returns**, so it pins agreement with the integrator rather than with
the gate. The third arm — `RATE = -2`, `D1 = 20`, `F1 = 0.5`, which must keep its 20-hour
length and stay silent — is the one a hand-written `F * duration` breaks;
`tests/modeled_duration.rs` covers `RATE = -2` only at `F = 1`, where a wrong scaling is
invisible.

### Verification

* **Mutation: 11 across two rounds, zero survivors.** M9 (the naive `f_bio * duration`) dies on
  the `RATE=-2` arm alone. M6 (`effective_duration` ignores the modeled resolution) read as a
  survivor under a `--lib`-scoped matrix and dies at Tier 2 in
  `modeled_duration_steady_state_overlap_warns` — which is why every round-3 mutation runs both
  scopes. Each restore verified by byte comparison against a snapshot.
* **Suites, both feature sets:** `ci` 180 binaries / 4907 passed / 0 failed / 171 ignored;
  `ci,survival` 180 / 5239 / 0 / 171. Baselines 179/4903 and 179/5235.
* **`preflight OK — fmt check clippy rustdoc docs public-api`.** The `check-report.qmd` row was
  trimmed to stay under R1 rather than added to `docs/.lint-baseline`.
* **`slow-tests`, per job and by content:** run `34151399470` vs `34097734820` on the base —
  `core` success in both; `nn` failure in both with the same two panics and the same
  `exceeded its 100m budget … exit code 124`, 164 `test result:` lines and zero `... FAILED`
  names in each. Pre-existing (#1239 / #1277).

The other eight round-1 findings check out as applied. #1280 (the `simulate()` scope call) is
filed and unchanged by this round.

PR #1282, opened 2026-09-08
feat(tools): amd — the search pipeline end to end, one report across every step [closes #1184]

### LINKED ISSUE #1184: feat(tools): AMD pipeline — orchestrate the search tools end to end, with CLI and ferx-r surfaces [#1175]

roninsightrx:
## Why

#1184, the last piece of the #1175 epic. The five search tools exist (#1180–#1183) and each one is usable on its own, but a model-development run is a *sequence* of them, and doing that by hand means five `.ferxsearch` files, five directories, and copying the winner's estimates into the next file between steps. Pharmpy's `amd` is that sequence; this is the ferx one.

The other half of the motivation is the report. A pipeline that prints only its final model is unauditable: the reader cannot tell a step that found a real improvement from one whose winner beat its siblings because they all stalled, and cannot tell a step that was skipped from one that ran and found nothing. The failure modes this epic is most exposed to — init stalls (#751), inner-EBE modes (#864, #891), boundary estimates — are exactly the ones that hide behind a winner-only table.

## What changed

`ferx-tools::amd`, `ferx amd`, `docs/tools/amd.qmd`, `examples/amd_start.ferx{,search}`, and a Pharmpy anchor under `tools/pharmpy-amd-anchor/`.

**No numerics.** Every decision is made by `modelsearch`, `iivsearch`, `ruvsearch`, `iovsearch`, `allometry` and `covsearch`. The module owns the sequencing, the seeding between steps, and the report.

**The order.** `structural → IIV → residual → IOV → allometry → covariates` by default; `[amd] strategy` also takes `reevaluation`, `SIR`, `SRI`, `RSI` (Pharmpy's upper-case spellings accepted). `[amd] skip` leaves a step out. `[amd] retries` = `all_final` / `final` / `skip` says which selected models get a perturbed-restart pass at `[run] retries + 1` starts.

**The space is partitioned.** Every tool of the epic *refuses* a foreign statement by name — deliberately, so a file meant for one tool is not half-run by another — and the AMD space is the union of all of them. So `amd::space` splits it: structural statements to `modelsearch`, `COVARIATE` to `covsearch`, `ALLOMETRY` to `allometry`, and a `COVARIANCE(*, …)` narrowed to its IIV half for `iivsearch` and its IOV half for `iovsearch`. `LET` travels with the statement that refers to it. A statement no step runs is an error naming it, never a narrowed search.

**Two integration defects, found by running it for real**, both fixed and both pinned by a test that names the measurement:

1. **`[rank]` has to be narrowed too.** `ruvsearch` and `covsearch` select by a likelihood-ratio test and refuse any `[rank] type` but `ofv` — correctly. But `type = "bic"` is what the steps that *do* rank need and is Pharmpy's default, so a pipeline configured the obvious way failed both steps by construction. The criterion now goes to the steps that rank on it; the two LRT steps keep their own p-values.
2. **The init-stall gate cannot be applied inside a step.** Every step refits the model it was handed, and that model was seeded from the previous step's estimates — so it starts *at* its own optimum and `reject_init_stall` (#751) reads a fit that does not move as a stall. Measured on `examples/amd_start.ferxsearch`: the residual and covariate steps both came back `failed` with "no free parameter moved more than 1% of its initial value", on a model that had just converged. The gate asks whether a fit ever left its starting values, which has no meaning when the starting values are the answer. It is off inside the steps, with the reason in every run's notes; it still applies to the pipeline's own **start fit**, which begins at the file's inits and is exactly what #751 is about, and every other gate (convergence, covariance, condition number, correlation, boundary) applies throughout.

**A failing step does not kill the run.** It is reported as `failed`, the pipeline carries on from the model it was handed, and `ferx amd` exits 1. On a pipeline of full population fits, discarding four completed searches to re-report one message is the most expensive way to deliver the least information.

**The report.** `steps.csv` — one row per *planned* step, skipped and failed ones included, with the criterion before and after, the ΔOFV and the wall clock. `candidates.csv` — one row per candidate of every step, with its Δ against its parent on both the criterion and the OFV, the strictness verdict *with its reasons*, whether the fit converged, and the seconds it took. The printed summary ends with the final model's estimates and standard errors. Each step's own directory keeps its tool's fuller record and the `input.ferx` it was handed, so a resumed run reuses those journals.

## Alternatives considered

* **Passing every step the whole space and letting the tools ignore what is not theirs.** They do not ignore it — they error, by design (#1180/#1181/#1183 all refuse a foreign statement by name). Loosening that to make AMD work would remove a check that exists precisely to catch a file pointed at the wrong tool.
* **Filling in a default search space for a step the file is silent about**, as Pharmpy does. Rejected: an eight-candidate structural search appearing overnight because a `[space]` was silent is a surprise, not a default. ferx skips with the reason recorded, and the anchor asserts the divergence.
* **Building each step's `BaseModel` in memory** instead of writing `input.ferx` and calling `prepare_run` on it. The file is the artifact the step's report needs anyway, `prepare_run` re-validates the edited model against the data exactly as a user's own re-run would, and there is no second construction path for a `PreparedRun` to drift from. It costs one CSV read per step.
* **Making `[amd] retries` the whole retry policy**, as in Pharmpy. ferx's runner already fits every candidate with `[run] retries` starts before the gate, so the per-candidate half of Pharmpy's motivation is already inside every step; what is left is the selected model, whose initials after seeding are its own optimum.

## Cross-repo dependency

| ferx (R) | — | not needed *(for this PR)* | — |

Nothing in `ferx-core`'s public API changed, so `api/ferx-core-public-api.txt` is untouched and no lock bump is needed. The issue also lists `ferx_amd()` / `ferx_modelsearch()` / `ferx_covsearch()` R wrappers; those are a follow-up PR in `ferx-r`, which consumes `ferx-tools` — **not covered here.**

## Breaking changes
- [x] None

`[amd]` is a new optional section of `.ferxsearch` (added to `TOOL_SECTIONS`); an existing file is unaffected.

## Numerical validation

- [x] Compared against: **Pharmpy 2.2.0**, for the object this tool owns
- [x] Reference values: end-to-end run below

**What is anchored, and why that is the right object.** AMD adds no numerics: every estimate it reports comes out of a tool with its own NONMEM-driven Pharmpy anchor (#1180–#1183). What it owns is the orchestration, so that is what `tools/pharmpy-amd-anchor/` dumps and `amd::pharmpy_anchor` replays — `get_subtool_order` for every strategy, and the subspace `get_search_space_modelsearch` / `get_search_space_covsearch` hand each tool for a corpus of spaces. Nothing is fitted, so the dump takes a second and never touches NONMEM.

Two deliberate divergences are asserted **as differences**, so one that quietly disappears is a red test just as a new one would be:

* Pharmpy fills a default search space where the file's space is silent; ferx skips the step.
* Pharmpy substitutes a `LET` at parse time; ferx resolves it against the model the step starts from — which is what makes `@PK` mean the parameters the *structural step* created.

**End to end.** `examples/amd_start.ferxsearch` on `data/two_cpt_oral_cov.csv` — 30 subjects simulated from `examples/two_cpt_oral_covmodel.ferx` (two compartments, an η on each of `CL`, `V1`, `Q`, `V2`, `KA`, proportional error, `CL~WT`, `CL~CRCL`, `V1~WT`). The starting model is deliberately the smallest defensible one: **one** compartment, no η on `KA`, no covariate relations.

```
  #   step         criterion          before        after          d  cand  seconds  selected
  1   structural   bic_mixed        -147.212     -660.287   -513.075     2      0.1  FO, 1 peripheral
  2   iivsearch    bic_iiv          -689.984    -1168.447   -478.463    17      1.9  [CL]+[KA]+[Q]+[V]+[V2]
  3   residual     ofv             -1185.453    -1185.453     +0.000     7     10.6  no residual-error feature added
  4   iovsearch    -                       -            -          -     -        -  skipped: the starting model declares no `iov_column` in [fit_options], so the dataset's occasions were not read
  5   allometry    -                       -            -          -     -        -  skipped: the search space has no ALLOMETRY statement
  6   covariates   ofv             -1185.453    -1196.446    -10.994    54      5.4  CL-CRCL-linear (forward step 1); CL-WT-power (forward step 2)
```

| step | found | truth |
|---|---|---|
| structural | one peripheral compartment | two compartments |
| IIV | an η on all five PK parameters | the same five |
| residual | nothing added | proportional, which the start model already had |
| IOV | skipped — no `iov_column` | no occasions in the dataset |
| allometry | skipped — no `ALLOMETRY` statement | — |
| covariates | `CL~CRCL`, `CL~WT` | `CL~WT`, `CL~CRCL`, `V1~WT` |

Final model at **OFV −1196.45**, against **−1199.02** for the model the data were simulated from (recorded in that model file's own comment). `V1~WT` is the one it misses, and the report says why rather than leaving it to be guessed: at forward step 3, the step that ended the search, the best `V–WT` candidate is `V-WT-power` at `dOFV = −0.66`, `p = 0.415` — a real effect the data are too small to establish at `p_forward = 0.05`.

## Performance

One extra fit per step (the pipeline's start fit, and the retries pass under `all_final`), and one CSV read per step from `prepare_run`. Against six searches of tens of population fits each, both are noise.

## Tests

**43 Tier-1 tests** in `crates/ferx-tools/src/amd/`, plus **5** end-to-end CLI tests.

The orchestration is written over a `StepRunner` seam, for the same reason `StepFitter` exists one level down: the order, the skips, the seeding, the retries policy and the report are observable only through the models handed between steps, and none of it is reachable in a unit test if the orchestrator calls `run_modelsearch` directly. Production is `ToolRunner`; the tests script it.

Worth a reviewer's eye:

* `each_step_starts_from_the_previous_step_s_selection_seeded` builds the expected model by the same two operations the pipeline performs, and the scripted fit moves θ₀ off the file's value — so removing *either* the hand-over or the seeding reddens it.
* `rank_is_narrowed_to_the_steps_that_rank` and `the_init_stall_gate_is_off_inside_a_step_and_said_so` each name the run that produced the defect, and both call the affected tool's own options reader on what AMD hands it — the property, not a spelling.
* `a_retries_pass_that_fails_the_gate_is_not_adopted` scripts a pass 50 OFV *better* that fails the gate, and asserts it is not taken.
* `every_statement_lands_in_exactly_one_step` asserts the subspaces' feature counts **sum** to the space's, so a future feature kind routed to two steps or to none fails; `the_structural_subspace_is_what_modelsearch_takes` then feeds the subspace to `modelsearch`'s own reader.
* Mutation-checked: flipping `LAGTIME`'s owner to `Covariates` and swapping the last two steps of `reevaluation` reddens 17 tests, including the Pharmpy anchor on both counts.

The CLI tests cover all six adapters against real fits (`maxiter = 0`), including the two that the main pipeline test does not reach — the allometry step, which is not a ranked search, and the IOV step, which runs with no `[space]` at all.

- [x] Example added to `examples/`

`examples/amd_start.ferx` + `examples/amd_start.ferxsearch`, run above. Three choices in that file were measured, not assumed, and the docs say so: `IIV?(@PK, EXP)` rather than a parameter list (a literal `[V,KA]` leaves the structural step's `Q` and `V2` without an η, and the final model 290 OFV worse); forward-only SCM at `p_forward = 0.05` (at the `0.01` default neither covariate enters, and a backward pass at the `0.001` default drops them again); `[rank] type = "bic"` at the top level, which reaches only the steps that rank.

## Docs

`docs/tools/amd.qmd`, plus `tools/pharmpy-amd-anchor/README.md` for the anchor.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, nothing here is gradient-reachable

### Example execution
- [x] `examples/amd_start.ferx` runs cleanly via the CLI (`--data data/two_cpt_oral_cov.csv`), and `examples/amd_start.ferxsearch` through `ferx amd` (the run above)
- [ ] ferx-r rebuild — not needed, no public-API change

## Reviewer hints

Read in this order:

1. `amd/space.rs` — small, pure, and the thing the whole design rests on.
2. `amd/mod.rs`'s `plan` / `skip_reason` / `step_config` — every decision made before a fit, and `INIT_STALL_NOTE` for the one gate the pipeline relaxes and why.
3. `amd/mod.rs`'s `drive` — the loop, the seeding, the retries pass, the failure handling.
4. `amd/adapt.rs` — the only place a tool's result is translated. Each adapter should take no decision the tool did not already take; the one exception is `from_allometry`, since allometry is not a ranked search and AMD has to decide whether to adopt the scaling (it does when the fit passes the gate, and keeps the unscaled model with a note when it does not).

The judgement call worth arguing with is turning `reject_init_stall` off inside every step. It is a real relaxation of a real gate, made because the pipeline guarantees the condition that makes the gate meaningless, and it is what the two failed steps in the pre-fix run were. If you would rather it were per-candidate rather than per-run, that needs a change in the runner, not here.

## Open questions

* The retries pass under `all_final` costs one fit per step and, on this example, never improved on the step's own winner — the notes say so three times. `final` may be the better default for ferx even though `all_final` is Pharmpy's; happy to switch.
* Unrelated, noticed while testing: `examples/two_cpt_oral_cov.ferxsearch` sets `[rank] type = "bic"` and so cannot be run by `ferx covsearch` at all (it errors on exactly the check #1 above works around). It predates #1180's rank check. Left alone here as out of scope — worth its own one-line fix.


roninsightrx (commented):
I found four correctness issues in the new orchestration/reporting path:

1. **[P1] `all_final` retry improvements leave the step summary stale** (`crates/ferx-tools/src/amd/mod.rs:892-946`). `StepOutcome.value_after` / `ofv_after` are captured before `retries_pass`, but that pass can replace `current_fit`. With the default `all_final` policy, an improved retry therefore produces a `steps.csv` row ending at the pre-retry OFV while the next row starts at the improved OFV. The retry is counted in the step wall time and described as part of that step, so the audit trail becomes internally inconsistent precisely when retries succeed. Please update the outcome from the adopted fit (including recomputing the step criterion), and pin the improved-`all_final` case.

2. **[P1] Every retry row is marked selected even when the fit is rejected** (`crates/ferx-tools/src/amd/adapt.rs:167-180`, consumed at `amd/mod.rs:1081-1097`). `fit_one` unconditionally sets `row.selected = true`; `retries_pass` later decides whether to adopt it but appends the row unchanged. A worse retry or one that fails strictness is therefore written as `selected=true`, and `render_summary` prints `SELECTED` before checking `passed`. This directly contradicts the note that the prior fit was kept. Please set retry selection from `improved` (while retaining selection for the start fit) and assert both the gate-failure and non-improvement paths.

3. **[P2] Cancellation during a retry pass is swallowed** (`crates/ferx-tools/src/amd/mod.rs:1073-1108`). `StepOutput.cancelled` is ignored and `retries_pass` returns only `Result<(), String>`. If cancellation happens during the final retry (or the last `all_final` retry), `AmdResult.cancelled` remains false and the CLI can exit 0 instead of 130; an intermediate `all_final` retry also advances into another step before cancellation is observed. Please propagate the cancelled state and stop the pipeline, with a test where the retry runner returns `cancelled=true`.

4. **[P2] CWRES pre-screen rows are reported as data OFVs, and even the chosen screen winner is labelled screened out** (`crates/ferx-tools/src/amd/adapt.rs:404-435`). `ruvsearch::StepRow` explicitly says a screened row `ofv` is on the CWRES pseudo-data scale and exposes `cwres_dofv` separately, but the adapter puts that number in both `value`/`ofv` and `d_value`/`d_ofv` under criterion `ofv`. Also, every screened row without an LRT gets the note `screened out on CWRES`, including `r.selected == true` (where selected means chosen for the real-data refit). This makes the cross-step candidate table compare unlike quantities and can print `SELECTED (screened out on CWRES)`. Please preserve the screen distinction/metric and make the note reflect whether the row was chosen for refit.

Validation performed at `01ea0d8139f15b28cd8ddce977734ed2a035e13a`: `cargo test -p ferx-tools --lib` (509 passed), `cargo test -p ferx-cli` (123 unit + 26 integration passed), `git diff --check`, and all current GitHub checks are green.

roninsightrx:
All four are real. Fixed in 9903e536, each pinned by one test and each mutation-checked in turn — five reverts, five distinct red tests, no overlap:

```
== 1 step outcome not updated after an adopted retry
   RED  amd::tests::an_improved_all_final_retry_updates_the_step_it_belongs_to
== 2 retry row always selected
   RED  amd::tests::a_retry_row_is_selected_only_when_its_fit_is_adopted
== 3 retry cancellation swallowed
   RED  amd::tests::cancellation_during_a_retries_pass_stops_the_pipeline
== 4 CWRES row reported as a data OFV
   RED  amd::adapt::tests::a_cwres_prescreen_row_is_not_reported_as_a_data_ofv
== 4b prescreen winner labelled screened out
   RED  amd::adapt::tests::the_prescreen_winner_is_not_labelled_screened_out
```

**1 — stale step summary.** Confirmed on the PR's own example before the fix: `steps.csv` had the structural step ending at `ofv_after = -696.706809` while the IIV step started at `ofv_before = -696.786198`. The pass's seconds and its row were already filed under that step, so the row was describing something other than what the step delivered. `StepOutcome` is now taken from the adopted fit, with the step's own criterion recomputed on it (`criterion.of(fit)`, not the tool's pre-retry `value`).

The test asserts the **invariant** rather than one cell — every ran step's `ofv_after` is the next one's `ofv_before` — and its pair, `a_retry_that_does_not_improve_leaves_the_step_alone`, asserts a worse pass leaves the numbers untouched. Without the second, "always overwrite" would pass.

The example's docs table moved with it: the structural step now reads `-660.366` rather than `-660.287`, and `docs/tools/amd.qmd` says why (a step's `after` is where the step *ended*, its pass included).

**2 — retry rows always selected.** `fit_one` marks its one candidate selected, which is right for the start fit and wrong for a pass the caller may reject; selection now follows `improved` in `retries_pass`, so `fit_one` stays generic. The test covers all three arms: adopted, rejected for landing higher, rejected by the gate with a 50-OFV-better fit.

Verified on the real run — only the pass that improved is `selected`:

```
step,tool,id,value,ofv,selected
1,retries,retries,-696.786198,-696.786198,true
2,retries,retries,-1185.452753,-1185.452753,false
3,retries,retries,-1185.452753,-1185.452753,false
6,retries,retries,-1196.125599,-1196.125599,false
```

While there, I took the second half of your point about `render_summary` checking `selected` before `passed`: a *genuinely* selected row that failed the gate now prints `SELECTED (failed the gate: …)`. A search that has to return something can hand back a model the gate rejected — ruvsearch returning its input when nothing passed — and `SELECTED` alone tells the reader the opposite of the verdict. Pinned by `a_selected_row_that_failed_the_gate_still_says_so`.

**3 — swallowed cancellation.** `retries_pass` now returns `RetriesOutcome { improved, cancelled }` and the pipeline stops on it. The test covers both positions you named: cancelled during an intermediate `all_final` pass (no later step runs — the scripted call log ends at `01-modelsearch-retries`), and cancelled during the final pass (every step ran, `AmdResult::cancelled` still true, so the CLI's 130 is reachable).

**4 — CWRES rows.** Agreed on both halves, and the fix keeps the metric rather than blanking it. A screened row now reports `cwres_ofv` in `criterion`/`value`, with `ofv` and `d_ofv` empty so a non-data number never sits in the column the rest of the table compares. `d_value` is the negated `cwres_dofv` — that field is *base − row*, the opposite of the pipeline's `d` convention — so "negative is better" still holds and the number is the one ruvsearch decided on rather than one I re-derived. The note now distinguishes the row the pre-screen chose for the refit from the ones it screened out. `CandidateRow::criterion`'s doc comment now says it is what `value` is *on*, not always the step's ranking criterion, and `docs/tools/amd.qmd` says the same for the `candidates.csv` column.

The screened/unscreened assertion is a pair on purpose: the unscreened row beside it must keep `ofv` and `d_ofv`, so the fix cannot pass by blanking everything.

Re-validated at 9903e536: `tools/preflight.sh` green, `cargo test -p ferx-tools --lib` 516 passed, `cargo test -p ferx-cli` 123 + 26 passed, and the end-to-end example re-run — same selections (1 peripheral, five η, no residual feature, `CL~CRCL` + `CL~WT`), same final OFV −1196.45, and `steps.csv` now consistent across every join.

roninsightrx:
## Code review — `amd` pipeline

Reviewed the orchestration in `crates/ferx-tools/src/amd/`, its adapters/report, and `ferx amd`. The pipeline logic itself reads well — skip reasons, the per-step subspace partition, seeding and cancellation propagation all hold up. Four issues, all in the reporting / state-carry path.

### 1. The `retries = "final"` pass files its row under step 0, so the report never shows it

`crates/ferx-tools/src/amd/mod.rs:1100` calls `retries_pass` with `index = None`, and the pass fits through `fit_one(index.unwrap_or(0), ...)` — so the resulting `CandidateRow` carries `step = 0`. `render_summary` (`report.rs:242`) only iterates `result.steps`, whose indices are `1..n`, and filters `r.step == s.index`. The row is therefore dropped from the printed report entirely and, in `candidates.csv`, sits under the start step.

That is the one row that says whether the final model's optimum was confirmed, and it is invisible in exactly the mode whose whole purpose is that pass. Give the final pass its own index/step marker, or special-case it in the summary.

### 2. A step that changes the model without a fit carries the previous model's `FitResult` forward

`mod.rs:926`:

```rust
current = output.model;
if output.fit.is_some() {
    current_fit = output.fit;
}
seed(&mut current, current_fit.as_ref())?;
```

`current` is replaced unconditionally, `current_fit` only when the step produced one. When a step returns a changed model with no fit, three things follow:

- `seed` writes the *previous* model's estimates into the new model text;
- the next `retries_pass` compares its OFV against a `before` taken from a different model, which breaks the "this is the *same model text*, so the lower OFV is the better optimum" premise documented at `mod.rs:1082`;
- `AmdResult::final_fit` — written out as `final-fit.yaml` — can describe a different model than `final.ferx`.

Either keep the previous model when its fit is the one being kept, or clear `current_fit` and record the gap in the notes.

### 3. The `retries = 0` note is pushed once per step

`mod.rs:1100`: with `[run] retries = 0` and the default `retries = "all_final"`, the `starts < 2` early return pushes the identical note

> the retries pass was not run: `[run] retries = 0` asks for a single start, which would refit the selected model at its own estimates

once per step — six identical lines under `Notes:` for a default run, eight under `reevaluation`. Push it once, or hoist the check out of the per-step loop.

### 4. A cancelled start writes an unseeded `final.ferx`

`mod.rs:827` returns early on `start.cancelled`, and the `seed(&mut current, current_fit.as_ref())?` at line 840 comes after it — so `final_model` still holds the model file's original initial estimates. `write_report` still runs and `final.ferx` is still written, which contradicts its documented contract ("written as the pipeline left it — already seeded from its own estimates, so it reads and refits at the fit beside it"): re-running that `final.ferx` does not reproduce the `final-fit` next to it. Seed before the early return.

roninsightrx:
Functionality confirmed manually

PR #1283, opened 2026-09-08
perf(estimation): reuse exclusive fit pools and fuse FOCEI subject work

hiddevandebeek:
## Why

Repeated fits with explicit thread counts rebuild Rayon workers. Concurrent fits
with explicit budgets need independent capacity, while unpinned fits with the
same ODE settings should share persistent workers. FOCE/FOCEI also traverses the
subject population separately for EBE solves, marginal scores, and mixed-gradient
fallbacks.

Follow-up to #1115 and #1212; motivated by comparison with nlmixr2est's reuse of
workers through inner and outer optimization.

## What changed

- Reuse exclusive pool leases for explicit widths and `PoolPlan` outer execution.
  Unpinned fits with identical ODE overrides share a persistent keyed pool, while
  default fits retain their existing shared pool.
- Bound ordinary idle retention across settings to twice the automatic worker
  budget (at most 16 workers per cache), retaining one most-recent wider pool so
  repeated 24/32-thread configurations still avoid worker rebuilds.
- Keep nested SIR work on the enclosing fit's matching override pool instead of
  creating a second full-width pool.
- Validate programmatic ODE tolerances before building a pool or entering a fit;
  cache float keys use their bit representation.
- Finish each subject's FOCE/FOCEI marginal immediately after its EBE solve.
  Keep ordered population sums and the separate AGQ/Laplace objective route.
- Complete mixed analytic/FD gradient fallbacks within the subject task and move
  EBE buffers instead of cloning them.
- Reuse IOV event-parameter capacity across likelihood probes and allocate static
  prediction scratch lazily. Recompute all event values at every parameter point.
- Add focused regressions and reproducible, uninstrumented benchmark examples;
  allocation instrumentation is confined to a manually enabled example build.

## Alternatives considered

Longer Rayon active waiting reduced a microbenchmark's wall time modestly while
increasing CPU consumption substantially; no idle-policy change is included.
First-order-only IOV seed preparation is deliberately excluded: it improved IOV
but its compiled binary showed a persistent stiff-ODE slowdown. That experiment
and its measurements are preserved outside this PR's committed diff.

## Cross-repo dependency

| ferx (R) | — | not needed: no API/schema change | — |

## Breaking changes

- [ ] `FitResult` / sdtab fields
- [ ] Public Rust API
- [x] None

Migration notes: none. Pool leases cover the callback/scoped work; `PoolPlan::install`
callers must not let unscoped spawned work outlive the callback. Current ferx-tools
consumers use joining parallel iterators and satisfy this contract.
Worker/ODE isolation does not make process-wide optimizer switches reentrant.

## Numerical validation

- [x] Full Warfarin FOCEI fits converge on current main `c877b9a9` and this PR:
  both take 57 iterations with identical printed OFV, theta, omega, and sigma.
- [x] All 34 selected integration tests pass, including all four Warfarin
  NONMEM covariance checks, IOV/reset anchors, and stiff-method comparisons.
- [x] Compared against current main `c877b9a9` using identical harnesses and
  compiler profiles, plus earlier controlled research against `0596fbb9`.
- [x] Fixed-workload fit signatures match bit-for-bit across retained changes,
  thread counts, and instrumented/uninstrumented builds: OFV, theta, sigma,
  omega, omega_iov, and iteration count.

Full Warfarin comparison (300-iteration limit, four threads, no covariance):

```text
                         current main             PR
converged                true                     true
iterations               57                       57
OFV                      -286.004219506046525     -286.004219506046525
theta CL                 0.13269539439330955      0.13269539439330955
theta V                  7.737705029131656        7.737705029131656
theta KA                 0.8107965492777247       0.8107965492777247
```

The new reused-buffer Dual2/central-FD test measured worst scaled error
`1.76e-11`; its bound is `1e-9` (about 57x rounding headroom). The fixture contains
two doses/occasions, changing covariates, TIME, pk-only records, and a reset, and
asserts that the analytic provider is selected.

## Performance

- [x] Current-main comparison completed, seven measured rounds per variant,
  ABBA order, same `ci-test` profile on Windows/Core Ultra 7 155U:

| Pool benchmark | Main median | PR median | Speedup |
|---|---:|---:|---:|
| Short 1-thread fits | 354.63 ms | 374.33 ms | 0.95x |
| Short 4-thread fits | 259.12 ms | 227.45 ms | 1.14x |
| Default fits | 268.29 ms | 297.91 ms | 0.90x |
| Analytical batch | 154.71 ms | 145.04 ms | 1.07x |
| ODE batch | 4834.02 ms | 977.50 ms | **4.95x** |
| ODE 4-thread fits | 985.63 ms | 1064.48 ms | 0.93x |

- [x] Allocation traffic: analytical requested bytes 48,939,724 → 37,309,940
  (23.8% less); IOV 1,192,011,780 → 632,261,884 (47.0% less); stiff
  23,823,279 → 21,777,015 (8.6% less). These are cumulative requests, not RSS.
- [x] Slower/mixed controls are disclosed. A separate 20-sample confirmation
  of the three profiling fixtures used ABBA then BAAB, five fits per process:

| Profile fixture | Threads | Main median | PR median | Speedup |
|---|---:|---:|---:|---:|
| Analytical | 1 | 96.55 ms | 94.24 ms | 1.02x |
| Analytical | 4 | 44.96 ms | 52.30 ms | 0.86x |
| IOV | 1 | 522.70 ms | 535.32 ms | 0.98x |
| IOV | 4 | 251.43 ms | 234.02 ms | 1.07x |
| Stiff | 1 | 369.64 ms | 373.41 ms | 0.99x |
| Stiff | 4 | 161.32 ms | 168.94 ms | 0.95x |

Small-workload timing directions vary between sweeps (analytical four-thread
was 1.07x in the first clean sweep versus 0.86x in confirmation); all confirmation
ranges overlap, with large outliers on this shared host. The stiff four-thread
control remains slower in both clean sweeps (0.92x and 0.95x). Excluding the
first-order experiment therefore does **not** establish that every retained
workload is faster. The trade-off is the robust batch-capacity correction and
lower allocation traffic; these control timings merit reviewer attention and a
representative release-build check, rather than a universal speed claim.

An initial current-main timing sweep overlapped a compiler process and is
excluded from the evidence above. No compiler was observed during the clean
comparison/confirmation. Objective/parameter/iteration signatures match in all
comparisons, including the slower controls.

These are medians of bounded fits, not universal or time-to-convergence gains.
Timing builds have no allocation hook or provider timers. The PR does not claim
the excluded first-order IOV experiment's 1.8–1.9x gain.

Reproduce with `examples/pool_benchmark.rs`, `examples/focei_profile.rs`,
`tools/benchmark-pools.py`, and `tools/test-focei-profiles.py`, building both
revisions with the same optimized profile. The allocator helper is maintainer
instrumentation; its Codecov exclusion is by role, not a coverage workaround.

## Tests

- [x] Unit regressions added for lease reuse/exclusivity, shared unpinned pools,
  stable settings keys, wide-pool retention, panic recovery, nested SIR scope,
  invalid Rust API overrides, fused objectives/fallbacks,
  buffer ownership, event refresh, and Dual2/FD parity.
- [x] CI patch coverage: **99.24%**, above the 90% gate; core and endpoint
  coverage jobs pass. Both debug-assertion suites pass. All applicable CI checks
  are green; the full slow-test job is intentionally skipped by PR policy.
- [x] Mutation: disabling idle-pool retrieval makes the worker-reuse test fail.
- [x] Mutation: disabling shared lookup, wide retention, nested-scope detection,
  and override validation makes six cache tests and two routing tests fail.
- [x] Mutation: omitting observation-buffer clearing makes the new reused-buffer
  parity test fail on its second probe (8 snapshots where 4 are required).
- [x] Final restored-code Windows unit suite: 4194 passed, 23 ignored,
  and only the pre-existing steady-state failure described below. Both mutation
  targets pass after restoration. All 34 selected integration tests pass.

The previous Windows full suite has a separately reproduced pre-existing
`ss_input_rate_over_capacity_deep_saturation_declines` failure; it must be
reported separately from new regressions, not hidden by skipping it.

## Example

- [x] Not applicable: internal performance changes with no new API/DSL surface.
  The new Rust examples are maintainer benchmark harnesses.

## Docs

- [x] `docs/model-file/fit-options.qmd` documents worker reuse, limits, and scope.
- [x] New pages: none; no sidebar changes required.
- [x] `CHANGELOG.md` has Unreleased performance entries.

## Checklist

- [x] Full `tools/preflight.sh` green on final source: fmt, all five check
  feature sets, all-targets Clippy, Rustdoc, docs lint, and public API.
  The API working copy was normalized from CRLF to LF to avoid a Windows
  newline-only false diff; the committed API baseline is unchanged.
- [x] Generic sensitivity/value formulas stay differentiable; the new FD check
  exercises the Dual2 route rather than a fallback.
- [x] Version bump not needed: no breaking change.

### ferx-site

No new DSL, fit option, model example, or dataset. No site or data mirror change.

### ferx-book

No new estimator, DSL feature, or fit option. No book change required.

### Example execution

- [x] No cross-repo example execution step required for this internal change.
  Convergence/reference validation is tracked separately above.

## Reviewer hints

Review `src/api/pool_cache.rs` for exclusive lifetimes and bounded idle retention;
`src/estimation/outer_optimizer.rs` for objective routing, the factor of two in
gradients, and ordered sums; and `src/pk/mod.rs` for clearing/refilling all event
vectors. The benchmark helpers can be read separately from production changes.

## Open questions

The first-order IOV experiment is deferred. The retained changes correct batch
worker budgets and reduce allocations, but smaller single-fit timing is mixed
and the rebased stiff control is slower. Review the disclosed trade-off before
making a broad release-performance claim. CI test/coverage completion is tracked
in the checks; this PR does not merge automatically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -379,9 +379,10 @@ section of the SDLC for the versioning policy).
 - The main FOCE/FOCEI optimizer paths combine EBE solves with subject scores;
   mixed analytic/FD gradients use one subject pass. This reduces synchronization
   between inner and outer optimization and avoids EBE-buffer copies (#1115).
-- Reuse idle worker pools for explicitly sized fits and `PoolPlan` batches; concurrent
-  fits with identical ODE overrides retain independent worker budgets, and idle pool
-  retention is bounded across thread counts and solver settings (#1115, #1212).
+- Reuse worker pools for repeated fits and `PoolPlan` batches. Explicitly sized fits
+  retain independent budgets; unpinned fits with identical ODE overrides share one
+  persistent pool instead of multiplying worker counts. Idle retention is bounded across
+  thread counts and solver settings, with one most-recent wide pool retained (#1115, #1212).
 - **A joint PK-TTE model with a linear PK block now takes the exact steady-state solve
   (#1210).** The hazard accumulator's one-cycle map is the identity, so it made `I - M`
   singular and the exact `(I - M)^-1 b` fixed point (#914) declined for *every* joint model,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -372,7 +372,7 @@ section of the SDLC for the versioning policy).
 
 ### Performance
 - Allocate per-event PK scratch storage only when the prediction path needs it,
-  reducing allocation traffic for static-model FOCE/FOCEI fits.
+  reducing allocation traffic for static-model FOCE/FOCEI fits (#1283).
 - IOV inner optimization reuses per-event PK parameter buffers across likelihood
   probes, reducing allocation traffic while recomputing every event at the current
   parameters, covariates, time, and occasion (#104).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -371,6 +371,17 @@ section of the SDLC for the versioning policy).
   silently scoring the declared value — `[saem, focei]` is fine, `[focei, imp]` needs `FIX`.
 
 ### Performance
+- Allocate per-event PK scratch storage only when the prediction path needs it,
+  reducing allocation traffic for static-model FOCE/FOCEI fits.
+- IOV inner optimization reuses per-event PK parameter buffers across likelihood
+  probes, reducing allocation traffic while recomputing every event at the current
+  parameters, covariates, time, and occasion (#104).
+- The main FOCE/FOCEI optimizer paths combine EBE solves with subject scores;
+  mixed analytic/FD gradients use one subject pass. This reduces synchronization
+  between inner and outer optimization and avoids EBE-buffer copies (#1115).
+- Reuse idle worker pools for explicitly sized fits and `PoolPlan` batches; concurrent
+  fits with identical ODE overrides retain independent worker budgets, and idle pool
+  retention is bounded across thread counts and solver settings (#1115, #1212).
 - **A joint PK-TTE model with a linear PK block now takes the exact steady-state solve
   (#1210).** The hazard accumulator's one-cycle map is the identity, so it made `I - M`
   singular and the exact `(I - M)^-1 b` fixed point (#914) declined for *every* joint model,
--- a/codecov.yml
+++ b/codecov.yml
@@ -166,3 +166,9 @@ ignore:
   - "src/bin/generate_data.rs"
   - "src/bin/pktte_sim_anchor.rs"
   - "src/bin/rtte_sim_anchor.rs"
+  # Maintainer timing harness; real fits and numerical comparisons run manually
+  # on both revisions, outside the PR coverage job.
+  - "examples/pool_benchmark.rs"
+  - "examples/focei_pipeline_benchmark.rs"
+  - "examples/focei_profile.rs"
+  - "tools/profiling_alloc.rs"
--- a/docs/model-file/fit-options.qmd
+++ b/docs/model-file/fit-options.qmd
@@ -45,19 +45,23 @@ What runs, how long, and how it is resumed or traced.
 
 #### Worker reuse for repeated fits
 
-Default fits share a persistent worker pool.
-Fits with a positive `threads` value, or fit-level ODE overrides, exclusively
-lease a pool for the call. An idle pool with the same width and solver settings
-can be reused; concurrent fits get independent pools even when their settings
-match. This preserves the per-fit budgets assigned by `PoolPlan`.
+Default fits share a persistent worker pool. Unpinned fits with the same
+fit-level ODE overrides also share a persistent pool keyed by those solver
+settings, so changing a tolerance does not multiply worker counts for concurrent
+calls. A positive `threads` value explicitly budgets the fit and exclusively
+leases a pool for the call; an idle pool with the same width and solver settings
+can be reused, while concurrent pinned fits keep independent budgets. This
+preserves the per-fit budgets assigned by `PoolPlan`.
 This isolates worker budgets and ODE settings; concurrent fits should still use
 consistent process-wide inner-optimizer and EBE warm-start settings.
 
-Idle pools retain at most twice the automatic thread budget across all settings
-(at most 16 workers). Older idle pools are evicted, and pools wider than that
-limit are released after use. Active fits keep their requested width; this idle
-limit is not a process-wide limit on concurrent fitting. Worker stacks remain
-32 MiB each, which is reserved capacity rather than necessarily resident memory.
+The caches ordinarily retain at most twice the automatic thread budget across
+all settings (at most 16 workers each). Older entries are evicted. When a single
+pool is wider than that budget, it can remain as the sole most-recent entry so a
+repeated explicit wide configuration still avoids rebuilding its workers.
+Active fits keep their requested width; the idle limits are not process-wide
+limits on concurrent fitting. Worker stacks remain 32 MiB each, which is
+reserved capacity rather than necessarily resident memory.
 
 During FOCE/FOCEI optimization, the same pool serves the inner EBE solves and
 outer gradient evaluations. The main NLopt and built-in BFGS/L-BFGS paths evaluate
--- a/docs/model-file/fit-options.qmd
+++ b/docs/model-file/fit-options.qmd
@@ -43,6 +43,30 @@ What runs, how long, and how it is resumed or traced.
 | `checkpoint_interval_secs` | integer | `300` | Minimum wall-clock seconds between checkpoint writes. Larger values reduce I/O overhead but lose more progress on an interrupt; smaller values save more often. Only used when `checkpoint = true`. |
 | `optimizer_trace` | `true`, `false` | `false` | Write a per-iteration CSV to `/tmp/ferx_trace_<pid>_<ts>.csv`. The path is stored in `FitResult::trace_path`. Useful for diagnosing convergence problems or comparing optimizers. See [Optimizer Trace](#optimizer-trace). |
 
+**Worker reuse for repeated fits.** Default fits share a persistent worker pool.
+Fits with a positive `threads` value, or fit-level ODE overrides, exclusively
+lease a pool for the call. An idle pool with the same width and solver settings
+can be reused; concurrent fits get independent pools even when their settings
+match. This preserves the per-fit budgets assigned by `PoolPlan`.
+
+Idle pools retain at most twice the automatic thread budget across all settings
+(at most 16 workers). Older idle pools are evicted, and pools wider than that
+limit are released after use. Active fits keep their requested width; this idle
+limit is not a process-wide limit on concurrent fitting. Worker stacks remain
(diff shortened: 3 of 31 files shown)


hiddevandebeek:
## Review — `perf(estimation): reuse exclusive fit pools and fuse FOCEI subject work`

Scope reviewed: `git diff c877b9a9...pr1283` (2334/268 across 22 files), with surrounding context read in `src/api/pool.rs`, `src/api/pool_cache.rs`, `src/estimation/{inner,outer}_optimizer.rs`, `src/pk/mod.rs`, `src/stats/likelihood.rs`, `src/estimation/sir.rs`, `src/estimation/sens_outer_gradient.rs`, plus rayon-core 1.13 sources for `ThreadPool::drop` / `Registry::in_worker`.

### Numerics

The objective/gradient refactors are equivalent to their predecessors:

- `subject_nll` reproduces `foce_population_nll{,_iov}` exactly, including the ordered (non-`par_iter`) `.sum()` that #703 requires.
- `run_inner_loop_and_nll` gates on `options.agq_nodes()`, the same predicate `pop_nll_opts` uses, so Laplace/AGQ still receive the quadrature marginal.
- `2.0 * nll` correctly replaces `ofv_at_fixed(...)` in `optimize_bfgs::f_and_g` (`ofv_at_fixed` is `2.0 * pop_nll_opts` without the `nn_reg` penalty, and the penalty is still added afterwards).
- The inlined `population_gradient_sens{,_iov}_mixed` bodies match `per_subject_packed_gradients{,_iov}`, including the IOV `[eta, kappa...]` stacking; the dropped `MixtureClassGuard` was `None` at both call sites, and the `x 2` / `packed_fixed_mask` assembly is untouched.
- `predict_iov_with_scratch` clears all four vectors before any early return and recomputes every event; the `RefCell` scratch in `find_ebe_iov` cannot double-borrow (the borrow guard dies inside `obj`, and `gradient_fd` is serial — `RefCell` is not `Sync`, so a parallel FD would not compile).

The findings below are therefore all in the pool-lease rework.

### Findings

**1. `src/api/pool.rs:135` — medium. Idle bound is capped but pool widths are not, so wide pools are never retained.**

The idle bound is `default_thread_count() * 2` (at most 16, since `default_thread_count` is capped at `DEFAULT_THREADS_CAP = 8`), but pools are *sized* by `effective_default_threads()` or `PoolPlan::replicates`, neither of which is capped. `PoolCache::release` (`pool_cache.rs:72`) returns early for any pool wider than the bound, so those pools are never retained. Concretely: `ferx --threads 32` sets `GLOBAL_THREADS_EXPLICIT`, so `ode_scope_pool` sizes at 32; every fit with an `ode_*` override then builds and tears down a 32-worker x 32 MiB pool with zero cache hits. The same applies to `PoolPlan::new(24, 1)` / `from_budget(32, 200)` (replicates = 32). This is a regression against main, where `ode_override_pool` kept the pool alive for the process via `Box::leak`, and it silently disables the reuse this PR exists to add on exactly the wide-machine configuration it targets.

**2. `src/api/pool.rs:181` (and `src/estimation/sir.rs:484`) — medium. Nested SIR scope now builds a second full-width pool; comment is stale.**

The replaced doc paragraph used to guarantee that a nested `with_fit_ode_scope` resolved to *the pool the caller is already on*, so rayon ran it inline. `run_sir_core` is called from inside `fit_inner` (`src/api/fit.rs:2161`, `src/api/postfit.rs:530`) and opens that scope; its comment still states "Called from `run_sir` or `fit()` this costs nothing — the threads already carry that override, which the scope detects." That is no longer true: the enclosing `fit()` holds an exclusive lease on that pool, so `acquire` cannot return it and builds a **second** full-width 32 MiB-stack pool, then enters it via `Registry::in_worker_cross`, blocking one of the fit's own workers for the duration. Scenario: `fit()` with `sir = true` (or the SIR covariance fallback) plus any `ode_reltol`/`ode_method` override now runs at double the OS worker count and reserved stack. Results remain correct (the new pool carries the same override), but the stale comment should be corrected in either case.

**3. `src/api/pool.rs:143` / `:193` — medium. Exclusive leases multiply worker counts for unpinned fits that merely set a tolerance.**

`ode_scope_pool` falls back to `effective_default_threads()` when `threads` is unset, and the lease is now exclusive. K concurrent unpinned fits that merely set `ode_reltol` therefore hold K x default-width pools (main shared one), while K identical fits *without* an override still share the single `default_fit_pool()`. Nothing budgeted those fits — they never asked for a per-fit width — so setting a tolerance silently multiplies the process's worker count and 32 MiB stack reservations. On an 8-core box with 8 concurrent override fits: 64 workers and roughly 2 GB reserved, and (per finding 1's interaction) only 16 idle workers retained, so most rebuild each round. Consider keeping the shared pool when `options.threads` is `None`, and reserving exclusivity for explicitly pinned widths.

**4. `src/api/pool_cache.rs:43` — low. Reuse key uses derived `PartialEq` on `f64`; NaN from the Rust API disables caching permanently.**

The reuse key compares `OdeSolverOverride` with derived `PartialEq` over `f64` fields; main deliberately keyed on `f64::to_bits` and documented why. The new comment claims "Validated overrides contain finite positive tolerances," but only the `.ferx` parser validates (`parse_pos_finite`, `model_parser.rs:7668`) — `FitOptions::ode_solver_override` (`types.rs:7778`) only compares against the default and performs no finiteness check. A Rust caller passing `FitOptions { ode_reltol: f64::NAN, .. }` produces `Some(NaN)`; `is_empty()` is false, so the override is armed, and `p.ov == ov` is then permanently false: every fit builds a fresh pool and never gets a cache hit.

**5. `src/api/tests/pool_plan_tests.rs:203` — low. New concurrency test has a tight timeout that misdiagnoses as an exclusivity regression.**

`identical_overrides_preserve_concurrent_fit_capacity` gives the second concurrent fit a 2 s `recv_timeout`, after a 5 s budget already consumed by the first. Two `std::thread`s must both be scheduled and each must build a 32 MiB-stack Rayon pool; on a loaded CI runner that can exceed 2 s. The failure then reports "second fit was serialized onto the first fit's pool", misdiagnosing a timing flake as a pool-exclusivity regression. Widening the second timeout (or waiting on both with one generous deadline) would preserve the property without the false signal.

**6. `src/api/pool.rs:390` — low. `PoolPlan::install` places an unenforced contract on external callers.**

The new doc states "Join any unscoped spawned work before returning if it must finish before another caller can reuse the pool." Under main the leaked pool made a late-finishing `rayon::spawn` harmless; now the pool is returned to the idle deque the instant `op` returns and can be handed to a different caller — silently violating the budget-isolation guarantee this PR adds — or evicted and dropped (`Registry::terminate` does not join, so the workers tear down asynchronously underneath the pending job). No production call site does this today (no `rayon::spawn`/`spawn_fifo` outside tests), but ferx-tools is the intended consumer of `PoolPlan` and is not in this repository, so it is worth verifying there before merge.

---
Reviewed with `/code-review high`.

Issue #1284, opened 2026-09-08
bug(dosing): a NaN modeled D{n}/R{n} is silently clamped to the floor, so a non-finite duration is served as an instantaneous bolus


TeunP:
`DoseEvent::resolve_rate` turns a modeled `RATE` code into a concrete `rate`/`duration` by
reading the `D{cmt}` / `R{cmt}` slot and clamping it:

```rust
let d_raw = params.get(slot).copied().unwrap_or(0.0);
let duration = clamp_above_floor(d_raw, Self::DURATION_FLOOR);   // 1e-8
```

```rust
pub(crate) fn clamp_above_floor(x: f64, floor: f64) -> f64 {
    if x > floor { x } else { floor }
}
```

`NaN > floor` is **false**, so a `NaN` takes the `else` arm and comes back as the floor. The
clamp is documented and correct for its stated job — "a transient `D ≤ 0` is clamped to
`DURATION_FLOOR`" — but a non-finite value is not a small positive one, and it is silently
given the same treatment.

## What the engine serves, measured

1-cpt oral-free ODE, `CL/V = 0.1`, `V = 10`, one 100-unit `RATE=-2` dose at t = 0, observations
at t = 1, 4, 8. `D1 = TVD*exp(WT) - TVD*exp(WT2)` with both covariates at 1000, so `D1` is
`inf − inf = NaN` at typical values:

| `D1` | resolved dose | `check_model_data` | `predict()` |
|---|---|---|---|
| `NaN` | `duration = 1e-8` ⇒ effectively an instantaneous bolus | `[]` | `[9.048374185, 6.703200632, 4.493299458]` |
| `+inf` | `rate = amt/inf = 0` ⇒ `is_infusion()` false ⇒ **bolus** | `[]` | `[9.048374180, 6.703200628, 4.493299455]` |
| finite (control) | the infusion asked for | `[]` | `[4.448676668, 7.474124736, 5.010066454]` |

Both non-finite arms land within 5e-9 of `10·e^{−0.1t}` — the exact instantaneous-bolus
solution — against a correct answer that is **2.03× lower at t = 1**. No diagnostic on any
channel: not `fit()`, not `ferx check`, not `predict()`.

`R{cmt}` has the same shape with the opposite sign of damage: `NaN` ⇒ `rate = RATE_FLOOR = 1e-8`
⇒ `duration = amt/1e-8 = 1e10`, i.e. essentially **no drug delivered for the whole study**;
[...]

Issue #1280, opened 2026-09-07
bug(api): predict() and simulate() carry no model/data warnings


TeunP:
Every model/data diagnostic ferx computes reaches exactly two entry points. `predict()` and
`simulate()` serve the same numbers with nothing attached, so a model that `fit()` refuses to
stay quiet about is silently fine through the other two doors.

## Measured

`check_model_data_warnings` has two production callers, grepped not recalled:

- `src/api/fit.rs:1216` — copies each `d.message` into `FitResult.warnings`
- `src/api/validation.rs:4876` — the `ferx check` path

`check_model_data_rule` (the fatal bundle) is likewise `fit()`-only, at `src/api/fit.rs:550`.
Neither `src/api/predict.rs` nor `src/api/simulate.rs` mentions either function.

The concrete case that surfaced this is #1139's: an `[odes]` RHS reading `TAFD` under an
`SS=1` dose. `tests/ss_model_time_nonmem_anchor.rs::tafd_under_an_unlagged_ss_dose_stays_nan_while_tad_does_not`
measures that the prediction path returns `NaN` at **all four** observations for that model,
and `fit()` on it returns `OFV NaN` and now says why (`W_STEADY_STATE_ABSOLUTE_TIME`, #1278).
`predict()` and `simulate()` reach the same `ode_predictions`, and return the same `NaN` rows
with no diagnostic of any kind. Same for every other row in that bundle —
`W_STEADY_STATE_II`, `W_SDE_STEADY_STATE`, `W_SDE_RESET`, `W_SDE_LAGTIME`,
`W_NEGATIVE_LAGTIME`, the `[initial_conditions]` × SS check.

## Why now

This is load-bearing for a decision already taken. #1278's rationale for making the
absolute-clock finding a **warning** rather than an error rests on it, in those words:

> An error would protect `fit()` only. […] `predict()` and `simulate()` call neither, and would
> go on serving the same `NaN` unwarned. A warning surfaced by `fit()` *and* `ferx check` beats
> an error one entry point enforces.

The premise is right and the conclusion follows, but the warning inherits the same reach. The
argument is only actually discharged once the non-fit entry points can carry these findings.

## Shape of a fix

`simulate_with_options_diag` (`src/api/simulate.rs:350`) already takes `model`, `population`
and `params` — exactly `check_model_data_warnings`'s three arguments — and already extends
`SimulationOutput.warnings` from a second source (`take_ss_nonconvergence_warnings()`, at
[...]

Issue #1189, opened 2026-09-01
bug(ode): a NaN route lag panics the break-time sort; a negative one shifts the integration start


TeunP:
## Summary

Two small robustness gaps in the break-time list, both surfaced while self-reviewing #1174.
Neither is a wrong-number bug on a well-posed model; both turn a soft failure into a hard one or
shift a boundary that validation explicitly permits.

## 1. A NaN route lag panics instead of degrading

`build_segment_break_times` and `ode_predictions_with_states` sort with:

```rust
break_times.sort_by(|a, b| a.partial_cmp(b).unwrap());
```

`partial_cmp` returns `None` for NaN, so this is a hard panic. The event-driven sibling builder
(`predictions.rs:4973`) deliberately does **not** do this — it uses
`.unwrap_or(std::cmp::Ordering::Equal)`. The two are inconsistent.

Since #1174 a route lag reaches these two builders for **every** `InputRateKind`, not just
`ZeroOrder` (which is the point of the fix — smooth kernels step at their onset too). So:

```
transit(n=N, mtt=MTT, lag=LAG)     with   LAG = TVLAG * (WT/70)**B
```

and a subject whose `WT` column is missing — which since #1111 is stored as `NaN` rather than a
dropped key — yields `route_lag = NaN`, and `Option::unwrap()` aborts the entire `fit()` /
`simulate()` run. Previously such a model degraded to a non-finite trajectory that the solver's
divergence guard reports for that one subject, letting the optimizer steer away.

`InputRateForcing::validate` (`src/pk/absorption.rs:441`) checks finiteness only at fit-init, on
typical values at η = 0, so a per-subject covariate NaN or a mid-fit θ/η excursion slips past it.

**Fix:** use `unwrap_or(Ordering::Equal)` in both builders, matching `:4973`. A NaN break then
sorts harmlessly and the existing non-finite-solve guard reports it per subject.

## 2. A negative route lag sorts before the integration start

`validate` explicitly permits a negative route lag — it is a warn-only `W_NEGATIVE_LAGTIME`, doc'd
as "handled — onset earlier". `build_segment_break_times` seeds
`break_times[0] = subject_integration_start(subject)` (`:5791`), but a pushed onset
`d.time + lag_cmt + lag_route` can sort *below* that seed, and nothing clamps it.

Consequences:

- `ode_solve_until_chz_threshold` begins its segment loop before the subject's first event, so
[...]

PR #1286, opened 2026-09-08
fix(api): evaluate the dose-attribute and absorption checks at the engine's own snapshots [closes #1235]

### LINKED ISSUE #1235: bug(api): E_DOSE_ATTR_NONFINITE (and the absorption domain checks) only read baseline covariates, so a TV-covariate-driven non-finite lag reaches the engine
`check_dose_attr_finiteness` (added in #1231 for #1189 item 1) rejects a `NaN`/infinite `lagtime`/`ALAG{n}` or `F`/`F{n}` at fit init with `E_DOSE_ATTR_NONFINITE`, per subject, naming the subject. It evaluates the model's parameter function **once per subject, at `TIME = 0`, with the subject's baseline covariates**:

```rust
let pk = (model.pk_param_fn)(
    &model.default_params.theta,
    &zero_eta,
    &subject.covariates,   // baseline row only
    0.0,                   // TIME = 0
);
```

So a dose attribute that is finite at the first record and non-finite under a **later** time-varying covariate value is not caught. That subject reaches the engine, hits the `timeline_has_non_finite` guard mid-fit, and comes back all-`NaN` — correct, but exactly the opaque outcome the front door exists to pre-empt.

## Why this is reachable rather than theoretical

The route that makes a lag non-finite at all is an unclamped `exp`. The DSL guards the other obvious ones — measured, not assumed:

| op | behaviour |
|---|---|
| `/` | denominator below `1e-30` returns `0.0` |
| `ln` | argument floored at `1e-30` |
| `sqrt` | argument floored at `0.0` |
| `exp` | **not clamped** |

So `ALAG1 = TVLAG * exp(WT)` on an unscaled covariate overflows to `+inf`. With `WT` time-varying — a weight series, a renal marker, anything that moves — the baseline row can be benign while a later row is not, and only the later one is missed.

## Same limitation, same file

This is not unique to the new check: `check_absorption_dosing`'s per-subject `InputRateForcing::validate` loop and its pathway-fraction checks are all `TIME = 0` baseline evaluations too. That is why #1231 documented the limit on the new function rather than fixing it in one place only — a per-check fix would leave the others silently narrower and imply a coverage that isn't there.

## Suggested shape

Evaluate each check over the subject's **dose-time** covariate snapshots (`subject.dose_covariates`) in addition to the baseline, not the full record grid — a dose attribute is a dose-time quantity, so the snapshots that matter are exactly the ones a dose is administered under. Cost is one `pk_param_fn` call per distinct dose snapshot per subject, at fit init only.

Do it for the whole family (`E_DOSE_ATTR_NONFINITE`, `E_ABSORPTION_DOMAIN`, `E_ABSORPTION_FRACTION`) or not at all, so the "checked at typical values" contract means one thing across the file.

TeunP:
## Why

`check_dose_attr_finiteness` and `check_absorption_dosing` each evaluated `$PK` **once per subject**, at `(subject.covariates, TIME = 0)`:

```rust
let pk = (model.pk_param_fn)(&theta, &zero_eta, &subject.covariates, 0.0);
```

Both arguments are wrong relative to what the engine does, and they are wrong along **two independent axes**. Measured at `c877b9a9` — `check_model_data` returns `[]` and `predict()` returns all-`NaN` for each:

| | model | subject | `check_model_data` | `predict()` |
|---|---|---|---|---|
| **A** | `ALAG1 = TVLAG*exp(WT)` | `WT` 0.5 at dose 1, 1000 at dose 2 | `[]` | `[NaN, NaN, NaN, NaN]` |
| **B** | `ALAG1 = TVLAG*exp(TIME)` | **no covariates at all**, doses at 0 and 1000 | `[]` | `[NaN, NaN, NaN]` |

**B is not in the issue and is the wider gap:** no time-varying covariate is involved. `pk::subject_needs_per_event_pk` fires on `model_uses_time_builtin` alone, so the engine resolves `$PK` per dose while the check froze it at `0.0`. A fix widening only the covariate axis leaves B silently open — which is why the gate is the engine's own predicate rather than a `has_tv_covariates()` paraphrase of it, and why the mutation table below shows those two arms dying to *different* edits.

End to end the user is told four wrong things and no true one. `fit()` on two subjects, one of them A's:

```
ofv = 2e20
"Outer optimization did not converge"
"Covariance step failed: ... (zero diagonal — flat objective) ... SE estimates not available."
"EPS shrinkage is notably negative (-110.0%) ..."
"Positive IWRES autocorrelation detected (Durbin-Watson = 1.03) ..."
```

Nothing names the subject, the lag, `NaN` or the timeline, and the two loudest messages send the reader after a flat objective and a misspecified error model.

**Error rather than warning, settled by measurement.** The objection to a hard rejection is "the user still gets results for their other subjects". They do not:

| population | OFV |
|---|---|
| 2 good subjects | −5.648127927980077 |
| 1 bad + 1 good | 2e20 |
| **1 bad + 9 good** | **2e20** |

One bad subject in ten replaces the whole objective with the sentinel and the optimizer never leaves its starting point (`left init = false`). There is no partial result for a rejection to take away, and unlike #1189 there is no NONMEM parity to preserve.

## What changed

**One helper, `typical_event_snapshots(model, subject, set, buf)`,** materialising the typical-value snapshots through the **engine's own** builders rather than re-spelling `(pk_param_fn)(θ, 0, cov, t)`. That is the point: it carries `pk::subject_needs_per_event_pk` itself instead of a paraphrase, and that predicate is what covers arm B.

**Each check names its own snapshot set, because they genuinely differ.** The issue proposed `dose_covariates` for all three checks; measured, that is a false positive for one of them:

| check | engine reads it at | set |
|---|---|---|
| `E_DOSE_ATTR_NONFINITE` (lag, `F`, `D{n}`, `R{n}`) | `pk_at_dose[k]` **only** | `Doses` |
| `E_ABSORPTION_DOMAIN` / `_FRACTION` | `prepare_input_rates(ode, &ext_params_ed)` **per segment**, from the last event's snapshot | `AllRecords` |

So a lag going non-finite at an *observation* is a value the engine never applies and must **not** be rejected, while an absorption parameter reaching the same record **must** be. Both verdicts are asserted, against each other, so a shared helper cannot quietly collapse them.

**The per-compartment de-dup is gone.** `seen: BTreeSet<usize>` skipped a repeated compartment — harmless with one snapshot per subject, a defect once the snapshot varies per dose, because an early *finite* dose claims the key and the later bad one is skipped. That is the commonest shape there is (repeat dosing into one compartment). It never de-duplicated a *message* either: the function returns on the first error.

**`D{n}` / `R{n}` values are now checked.** `check_modeled_dose_rates` validated only that the slot *exists*. A non-finite value never surfaces as a `NaN`:

| | served | vs the correct infusion |
|---|---|---|
| `D1 = +inf` | `rate = amt/inf = 0` ⇒ not an infusion ⇒ instantaneous bolus | **1.90× high** at `t = 1` |
| `R1 = +inf` | `duration = amt/inf = 0` ⇒ the same bolus | the same |
| `D1 = NaN` | clamped to `DURATION_FLOOR = 1e-8` ⇒ a bolus again | the same |
| `R1 = NaN` | clamped to `RATE_FLOOR = 1e-8` ⇒ `duration = amt/1e-8` | essentially nothing delivered |

The bolus rows land within **8.2e-7** (1.8e-7 relative, the ODE tolerance) of the exact `10·e^(−0.1t)`. This rides here rather than in a follow-up because it is the same function, the same per-dose snapshot set (`resolve_subject_doses_with`'s `pk_for_dose(k)`) and the same de-dup key — a slice-only PR would rebuild this one's plumbing to test itself. The **mid-fit** clamp semantics stay out: changing them moves trajectories on every modeled-`RATE` model, which is the one property this PR is built not to have. That is **#1284**.

## Behaviour change on a public entry point

`predict()` / `simulate()` run no data check, but they **do** call `assert_absorption_dosing_supported`, which panics on the first error `check_absorption_dosing` returns. Widening that check's snapshot set therefore widens what these abort on: a model that previously returned numbers from an out-of-domain forcing at a later record now panics there. Deliberate — those numbers came from a forcing the engine really does apply — and pinned by a `#[should_panic]` test.

`E_DOSE_ATTR_NONFINITE` has **no** such `assert_*` twin and still does not reach `predict()` / `simulate()`. That gap is #1280 / #898 and is not closed here; the honest fix is not an eleventh panic wrapper.

## Alternatives considered

**Copying the per-dose loop into `api/validation.rs`** instead of sharing the engine's. Rejected: a second spelling of "where a dose attribute is read" is the exact defect class this PR fixes. See Performance for what sharing cost and how it was recovered.

**One snapshot set for the whole file** (the issue's "whole family or not at all"). Measured wrong — see the table above.

**Warning rather than error.** Measured empty — see the OFV table.

## Cross-repo dependency

No public Rust API change (everything added is `pub(crate)`; the baseline is untouched) and no new diagnostic code. Worth knowing downstream: an R caller of `predict()` / `simulate()` on a model with an out-of-domain absorption parameter at a later record now gets an error where it previously got numbers.

## Breaking changes
- [x] None

The `predict()` / `simulate()` abort described above is a behaviour change, not an API change — no signature, field or file-format change.

## Numerical validation

**Nothing moved.** The same `fit()` on a working TV-covariate model, `c877b9a9` vs this branch:

```
# before (HEAD)
ofv bits   = 0xc01768ac93ccddb9
theta bits = [0x3fe8eac588aa0b7f, 0x402d1575bb7a5c0a, 0x3fd3a3abed91a189]

# after
ofv bits   = 0xc01768ac93ccddb9
theta bits = [0x3fe8eac588aa0b7f, 0x402d1575bb7a5c0a, 0x3fd3a3abed91a189]
```

- [x] Compared against prior ferx commit `c877b9a9`

**No NONMEM anchor, stated so it does not read as skipped.** This changes *whether a fit starts*, not any number it produces; the bit-identity control above is what proves that rather than asserting it. No `Dual2`-vs-FD parity test either: no sensitivity formula is touched.

Full `--features ci` suite: **181 binaries, 4929 passed, 0 failed** — so the widened absorption check false-positives on no repo fixture.

## Performance

Measured, because the first cut was not free. `check_model_data` on a synthetic 1000 subjects × (2 doses + 100 observations):

| | `c877b9a9` | first cut (full builder) | **shipped** |
|---|---|---|---|
| parameter-static model | 8.35 ms | 15.4 ms | **8.87 ms** |
| TV-covariate model | 9.03 ms | 111.7 ms | **9.91 ms** |

The first cut routed a doses-only caller through `pk::compute_event_pk_params_into`, which resolves `$PK` at **every** record — 102,000 evaluations to read 2,000. Rather than copy the loop, the per-dose body is extracted as `pk::fill_dose_pk_params` and shared, with the static arm's snapshot **handed in** (`constant: Option<PkParams>`) so `compute_event_pk_params_into` — which runs in SAEM's MH loop and the FOCE objective — gains no extra `$PK` call. One implementation, and the cost back to ~1 ms over the un-widened check.

**Round 2 — those two models carry no `[odes]` input rate, so they do not measure the widened path** (review finding 3). `check_absorption_dosing` returns at `ode.input_rate.is_empty()`, so the numbers above time the `Doses` path only. Re-measured on a model that *has* a forcing, same shape, release, best of 5:

| absorption check evaluates at | TV-covariate | parameter-static |
|---|---|---|
| doses only | 1.56 ms | 0.89 ms |
| **every record (shipped)** | **21.53 ms** | **4.78 ms** |

~14× on the TV-covariate arm — what the check costs to see the value the engine applies at a later observation. The static arm is 5× because every record still gets its own `validate()` call although all 102 snapshots are copies of one `$PK` evaluation; the builder's constant arm keeps the `$PK` count at 1.

`assert_absorption_dosing_supported` sits inside `simulate_inner_with_draw`, so `simulate_with_uncertainty` pays this **per draw** on a θ-independent verdict. Stated as a fraction rather than a count: one full simulation pass over that same population is **642.4 ms**, so the repeat is **3.4 % of a draw**, up from 0.24 %. Real, and filed as [#1288](https://github.com/FeRx-NLME/ferx-core/issues/1288) rather than fixed here — the obvious `draw == 1` gate is *unsound* (the uncertainty loop `continue`s past a draw whose flip-flop check fails, so draw 1 can be skipped and the guards never run), and the correct hoist threads an "already checked" argument through the simulate chokepoint's four call sites, which is not a change to make inside a validation-snapshot PR.

## Tests

- [x] Unit test added in the same module (`pk::tests::dose_only_snapshots_match_the_full_event_builder`)
- [x] Regression tests added that fail without this fix (`tests/typical_value_snapshot_sets.rs`, 10 tests)
- [x] **Round 2:** 7 more, same file — 17 total; 8 further mutations, all killed

**Every arm names the minimal edit it exists to kill, and all 11 were run.** 11 mutations, 11 killed, with kill sets that are disjoint in the way the design predicts:

| mutation | tests killed | survives |
|---|---|---|
| gate on `has_tv_covariates()` (covariate axis only) | arm **B** (1) | arm A |
| each record's `TIME`, baseline covariates (time axis only) | arm **A** + 4 `D`/`R` + 2 absorption + panic (6) | arm B |
| keep the per-compartment `seen` de-dup | arms A, B + both multi-dose `D`/`R` (4) | — |
| evaluate only `pk_at_dose[0]` | arms A, B + both multi-dose `D`/`R` (4) | — |
| point the dose-attribute check at `AllRecords` | 13, incl. 7 pre-existing `transit_absorption` tests | — |
| point the absorption pair at `Doses` | both absorption arms + the panic test (3) | — |
| drop the `D{n}`/`R{n}` value check | both `D`/`R` arms (2) | — |
| check `D{n}`/`R{n}` at the **first dose only** | **only** the two multi-dose `D`/`R` arms (2) | arms A, B |
| shared body reads `&subject.covariates` | parity test + arm A + both `D`/`R` | — |
| shared body reads `0.0` for the time | arm B only — **parity test survives**, see below | parity test |
| the two callers of the shared body diverge | parity test + arms A, B + both `D`/`R` | — |

Row 8 is why the `D`/`R` arms are **both single- and multi-dose**: the multi-dose ones kill a mutation nothing else in the suite can see. Row 10 is the parity test's measured blind spot — both sides of a bit-equality comparison move together when the shared body changes, so that axis is arm B's job. It is written into the test's own doc comment rather than left for a reviewer to discover.

Non-degeneracy and false-positive controls: every lag arm is multi-dose with the *first* dose finite (a single-dose fixture is caught by the pre-existing baseline check and cannot fail); a lag non-finite only at an observation must not be rejected **and must still predict finite**; a finite multi-dose modeled regimen must not be rejected; the finite-`D1` control asserts its curve is materially different from the bolus, so the two `+inf` arms cannot pass on a degenerate fixture.

## Docs

`docs/file-formats/check-report.qmd`'s `#### Dose records` measured **4,955 characters against R1's 5,000**, so it is **split** (`Dose records` / `Dose attributes` / `Endpoint routing`) *before* the `E_DOSE_ATTR_NONFINITE` row is edited — otherwise `Docs lint` goes red and `docs/.lint-baseline` is add-nothing-new. No `#dose-records` link exists anywhere in `docs/`, so R4 is unaffected. New leaves: 1,736 / 2,935 / 1,082.

`docs/model-file/lagtime.qmd` and `docs/model-file/absorption.qmd` updated where "at typical values" now means something wider and per-record.

## Checklist
- [x] Public-API baseline unchanged (all new items `pub(crate)`)
- [x] No `Dual2` sensitivity path touched

## Reviewer hints

1. **`SnapshotSet` is the design decision.** If the two checks should share one set, the false-positive control (`a_lag_non_finite_only_at_an_observation_is_not_rejected_and_still_predicts`) is the test that says why they can't.
2. **The `predict()` / `simulate()` abort widening** — a deliberate user-visible change, flagged above rather than buried.
3. **`pk::fill_dose_pk_params`'s `constant: Option<PkParams>` argument** looks like an odd shape until you see why: it keeps the hot-loop caller at exactly one `$PK` evaluation while still sharing one body. The Performance table is the justification.

Skimmable: the docs split is mechanical (rows moved, one row rewritten), and the `.qmd` diff is larger than the change.

## Open questions

**An adjacent gap found and deliberately not widened into.** `D{n}` / `R{n}` **≤ 0** — as opposed to non-finite — is silently wrong in the same way: `D = 0` clamps to `DURATION_FLOOR` and serves a bolus. That is a *domain* check under a code named for finiteness, so it is out of scope here. It may belong to #1284; worth a look before it is filed separately.


TeunP:
## Review at `0ec09985`

Ran the full gate myself in a worktree at the PR head, plus an independent pass. All three test
legs green — `--workspace --lib` 3 binaries / 4703 passed, `--workspace --tests` 202 binaries /
5677 passed, `-p ferx-core --features ci,markov,nn --tests` 180 binaries / 5450 passed, 0 failed
anywhere. `docs-lint --check` green; the split leaves measure 1,747 / 3,105 / 1,089 characters
against R1's 5,000 and nothing links to `#dose-records`. Public-API baseline correctly untouched.

I spot-checked two rows of the mutation table and both hold exactly as written: pointing the
absorption pair at `Doses` kills 3 (both absorption arms + the panic test) and leaves 7; freezing
the shared body's time at `0.0` kills **only** arm B and the parity test survives, which is the
blind spot the parity test's own doc comment already declares.

The core thesis is right, and I measured it rather than taking it on the code reading. With the
absorption check narrowed back to `Doses` so nothing rejects, a transit `mtt` driven out of domain
only at observation 2 gives `predict() = [0.305094, 0.603292]` against an in-domain control of
`[0.305094, 1.102569]`. The engine really does apply it there. `AllRecords` is the right set for
the density kernels.

It is not the right set for the other two record kinds it sweeps in.

| # | Finding | Severity | Evidence | Suggested fix |
|---|---|---|---|---|
| 1 | `AllRecords` includes EVID=3/4 **reset** snapshots, which no input-rate forcing is ever built from | should-fix | measured — see table below | exclude `buf.reset` from the set |
| 2 | A `zero_order()` forcing's `dur` / fraction / route lag are **dose-time** quantities; the per-segment RHS loop skips the kind entirely | should-fix | measured — see table below | evaluate a `ZeroOrder` forcing at dose snapshots only |
| 3 | The Performance table measures a model with no `[odes]` input rate, so it cannot see the path that changed; `simulate_with_uncertainty` pays the widened check **per draw** | should-fix | `validation.rs:971`, `simulate.rs:894` + `:1104` | hoist or memoize; the verdict does not depend on the draw |
| 4 | `W_MODELED_DURATION_NONPOSITIVE` / `_RATE_` still read `D{n}`/`R{n}` at the frozen `(covariates, 0.0)` snapshot, memoized per subject | should-fix | `validation.rs:4348` | widen, or restore the deleted disclosure sentence |
| 5 | `#[should_panic(expected = …)]` matches the wrapper text, not the widened diagnostic | nit | `validation.rs:2141` + `:2854` | `expected = "observation 2 at TIME=8"` |
| 6 | No in-domain control on the `AllRecords` side, and no coverage of the reset / EVID=2 snapshot kinds | nit | deleting the `buf.reset` push leaves the suite green | one acceptance control per snapshot kind |
| 7 | `absorption.qmd` widened the transit bullet only; igd `:496`, weibull `:627`, zero_order `:714` still read "validated at typical values" | nit | grep | update the three |
| 8 | `zero_eta` moved from once-per-population to once-per-subject | nit | `validation.rs:1731` | hoist into the caller |
| 9 | `Vec::with_capacity(buf.dose.len())` under-sizes the `AllRecords` result | nit | `validation.rs:1756` | size from all four vectors |

### 1 and 2 — the widening is wider than the engine's own read set

Same experiment for all three, one input varied at a time: build the subject, run
`check_model_data`, then run `predict()` with `check_absorption_dosing` temporarily pointed at
`SnapshotSet::Doses` so nothing rejects, and compare against the identical subject with the
covariate in domain.

| out of domain at | this branch | `predict()` unrejected | in-domain control | verdict |
|---|---|---|---|---|
| transit `mtt`, **observation 2** | `E_ABSORPTION_DOMAIN` + `predict()` panics | `[0.305094, 0.603292]` | `[0.305094, 1.102569]` | **true positive** — 45% error |
| transit `mtt`, **reset record** | `E_ABSORPTION_DOMAIN` "reset record 1 at TIME=4" + panic | `[0.30509411543990944, 0.0]` | `[0.30509411543990944, 0.0]` | **false positive** — bit-identical |
| `zero_order` `dur`, **observation 2** | `E_ABSORPTION_DOMAIN` + panic | `[0.9516258195866305, 0.9948270849262485]` | same bits | **false positive** — bit-identical |

The engine says so itself in both cases:

* `is_record` is `DoseRecord | PkOnly | Obs` (`ode/predictions.rs:5902`), so `governing_record`
  never resolves to a reset; `pk_now`'s `Kind::Reset` arm takes `last_pk` and that segment is
  discarded, and `record_pk_at`'s reset arm is commented "Unreachable while `is_record` excludes
  `Reset`". `pk_at_reset[..]` re-seeds `init(...)` and nothing else.
* `add_prepared_input_rate_forcing` does `continue` on `InputRateKind::ZeroOrder`
  (`ode/predictions.rs:2136`) before it reads the segment's route lag. The window comes from
  `zero_order_dur_and_frac_for_dose` on the per-dose snapshot, which the event-driven walk
  documents as "the constant rate is fixed at dose time (mass-exact under time-varying
  covariates)". `InputRateForcing::validate` validates it anyway (`pk/absorption.rs:464`).

This is the PR's own rule pointed the other way. `Doses` for the lag is justified by the engine's
predicate; `AllRecords` for the absorption pair is justified by prose ("every record"), and the
engine's predicate for that quantity is `is_record(kind) && kind != ZeroOrder`, not "every
record". Both escapes turn a model that returned numbers into an `fit()` rejection **and** a hard
`predict()` / `simulate()` panic, which is the same class as the false positive the dose-attribute
side has a dedicated control test for.

Reachability is not exotic: a reset row's covariates are the new period's in any crossover or
washout dataset, and `zero_order(dur=DUR)` with a covariate- or `TIME`-driven `DUR` is an ordinary
absorption model.

### 3 — the number in the Performance section cannot see the code that changed

`check_absorption_dosing` returns at `ode.input_rate.is_empty()` (`validation.rs:971`), and both
measured models are plain 1-cpt, so the 8.87 ms / 9.91 ms figures time the `Doses` path only. On a
model that *has* a forcing the check went from O(subjects) to O(records) `$PK` evaluations — the
same 102,000-against-2,000 ratio the PR quotes for the first cut, which is why that cut cost
111.7 ms.

It is not a fit-init-only cost. `assert_absorption_dosing_supported` sits inside
`simulate_inner_with_draw` (`simulate.rs:894`), and `simulate_with_uncertainty` calls that once
per draw (`:1104`); `simulate_with_options_diag` calls it twice per call (`:446` then `:894`). The
verdict is θ- and draw-independent, so on 1000 draws × 1000 subjects × 102 records this is ~10⁸
evaluations recomputing one constant answer.

### 4 — the same attribute, two contracts, one file

`W_MODELED_DURATION_NONPOSITIVE` / `W_MODELED_RATE_NONPOSITIVE` (`validation.rs:4348`) loop
`for d in &s.doses` and memoize a single `(pk_param_fn)(θ, 0, &s.covariates, 0.0)` per subject —
literally #1235's shape, on the `D{n}`/`R{n}` this PR just widened, 300 lines below it. I take the
≤ 0 *domain* question as correctly out of scope (#1284); the *snapshot* question is not a
different question. What makes it worth raising rather than filing is that the PR deletes the
paragraph that used to disclose it ("the same `TIME = 0` limitation applies to every other
per-subject typical-value check in this file"), so the file now carries two contracts and no note
saying which is which.

### Confirmed good

* The `Doses` / `AllRecords` split is the right call and the false-positive control test earns its
  place — measured above.
* Dropping the per-compartment `seen` de-dup is correct: the function returns on the first error,
  so it never de-duplicated a message.
* The `D{n}` / `R{n}` value check belongs in this PR — same function, same snapshot set, same key.
* The parity test's declared blind spot is real and correctly declared; arm B is what covers that
  axis.
* `predict()` / `simulate()` abort widening is flagged rather than buried, and pinned.

TeunP:
## Round 2 — every finding reproduced first, at `0ec09985`, fixed at `58e6e886`

Findings 1, 2 and 4 reproduce exactly as written, and I measured each of them myself rather
than taking the code reading: for 1 and 2 the decisive number is that the "bad" subject's
predictions are **bit-identical** to an in-domain control once the check is narrowed so
nothing rejects. Finding 3 reproduces too, but is about a third the size the arithmetic
suggests once a draw is measured instead of counted — details in its row. Fixing 2 opened a
false positive the review did not raise; it is listed as 2b because it is the same gate.

| # | Finding | Verified | Fix |
|---|---|---|---|
| 1 | `AllRecords` includes EVID=3/4 reset snapshots | **reproduced** — bit-identical predictions | `58e6e886` |
| 2 | a `zero_order()` window is a dose-time quantity | **reproduced** — bit-identical predictions | `58e6e886` |
| 2b | *(new)* the Σ ≈ 1 fraction check breaks on a mixed compartment once 2 is fixed | found while fixing 2 | `58e6e886` |
| 3 | the Performance table cannot see the path that changed | **reproduced**; per-draw repeat measured at **3.4%** of a draw | table corrected; hoist filed as [#1288](https://github.com/FeRx-NLME/ferx-core/issues/1288) |
| 4 | `W_MODELED_{DURATION,RATE}_NONPOSITIVE` still frozen at `(covariates, 0.0)` | **reproduced** | `58e6e886` |
| 5 | `#[should_panic(expected = …)]` matches the wrapper | **measured** — the mutation it should catch *survives* the old string | `58e6e886` |
| 6 | no in-domain control, no per-kind coverage | confirmed | `58e6e886` |
| 7 | `absorption.qmd` widened the transit bullet only | confirmed | `58e6e886` |
| 8 | `zero_eta` allocated per subject | confirmed | `58e6e886` |
| 9 | `Vec::with_capacity` under-sizes the result | confirmed | `58e6e886` |

### 1 and 2 — measured

One variable at a time: the same subject twice, differing only in the covariate on the record
under test, `check_model_data` then `predict()` with the absorption check temporarily pointed
at `SnapshotSet::Doses` so nothing rejects.

| out of domain at | this branch before | `predict()` unrejected | in-domain control | verdict |
|---|---|---|---|---|
| transit `mtt`, **reset record** | `E_ABSORPTION_DOMAIN` + panic | `[0.30509411543990944, 0.22308079058078897]` | **same bits** | false positive |
| `zero_order` `dur`, **observation 2** | `E_ABSORPTION_DOMAIN` + panic | `[0.47581290979331525, 0.24334182605819396]` | **same bits** | false positive |
| transit `mtt`, **observation 2** | `E_ABSORPTION_DOMAIN` | — | — | **true positive, kept** |
| transit `mtt`, **EVID=2 record** | `E_ABSORPTION_DOMAIN` | — | — | **true positive, kept** |

The set is now the engine's own predicate rather than "every record":
`RecordKind` is `Dose | Observation | PkOnly` — mirroring `ode::predictions`' `is_record`
— **with no `Reset` variant at all**, so the exclusion is a type fact and not a filter
someone can delete. `forcing_is_read_at(kind, rec)` carries the second half: `ZeroOrder`
is read at dose records only, everything else at all three.

### 2b — the Σ ≈ 1 check needs a completeness rule, not just a scope one

Gating `zero_order` out of non-dose records is not sufficient on its own. On a compartment
mixing a zero-order pathway with a density one (#505) the engine reads the two fractions at
**different** snapshots — the density's per segment (`add_prepared_input_rate_forcing`), the
zero-order's at dose time (`zero_order_dur_and_frac_for_dose`) — so at an observation the
sum is the density share alone. An ordinary `FR1*zero_order(dur=DUR) + FR2*first_order(ka=KA)`
with `FR1 + FR2 == 1` identically would have been rejected at its first observation for
summing to `FR2`. A dose record is the one kind that sees the whole partition, so on such a
compartment the Σ check runs there and only there. Pinned by
`a_mixed_zero_order_and_density_compartment_summing_to_one_is_not_rejected`, which fails on
the un-gated version.

### 3 — the table was wrong; the hoist is worth less than it looks

Both halves are right, and the first one is the defect: the Performance models had no
`[odes]` input rate, so `check_absorption_dosing` returned at `ode.input_rate.is_empty()` and
those numbers timed the `Doses` path only. Re-measured on a model that **has** a forcing
(release, 1000 subjects × (2 doses + 100 observations), best of 5):

| absorption check evaluates at | TV-covariate | parameter-static |
|---|---|---|
| doses only | 1.56 ms | 0.89 ms |
| every record (this PR) | **21.53 ms** | **4.78 ms** |

So the widening is ~14×, not free, and the PR description now says so.

The per-draw half I am **not** fixing here, and the reason is a number the review's arithmetic
does not include. One draw over that same population — `simulate_with_seed`, one simulation —
is **642.4 ms**. So the repeated check is 21.5 ms against 642 ms: **3.4% of a draw**, up from
0.24%. Real, worth removing, and not worth restructuring the single simulate chokepoint inside
a PR about validation snapshots.

It is also not a one-liner. The obvious `if draw == 1` gate is **unsound**: the uncertainty
loop `continue`s past a draw whose flip-flop check fails (`simulate.rs:1100`), so draw 1 can
be skipped and the guard never run at all. A correct hoist needs the guards run once in
`simulate_with_uncertainty` and an explicit "already checked" argument threaded through
`simulate_inner_with_draw`'s four call sites. Filed as [#1288](https://github.com/FeRx-NLME/ferx-core/issues/1288) with these numbers.

### 4 — widened, not disclosed

Taking the choice you offered, the wider one. `W_MODELED_DURATION_NONPOSITIVE` /
`W_MODELED_RATE_NONPOSITIVE` now read `D{n}`/`R{n}` from the same per-dose snapshots as
`E_DOSE_ATTR_NONFINITE`, via the same helper (`typical_event_snapshots` grew a `theta`
parameter, since the warning pass runs at `init_params.theta` rather than the compiled
defaults). The ≤ 0 *domain* question stays #1284, and the file no longer carries two
contracts for one attribute. New test:
`a_modeled_duration_nonpositive_only_at_a_later_dose_is_warned`, whose first dose is benign so
a fix reading only the baseline, or only `pk_at_dose[0]`, leaves it red.

The per-compartment de-dup there stays, deliberately: unlike the error check's removed `seen`
set it is only inserted into when a dose is **bad**, so an early good dose cannot claim the key.

### 5, 6 — the tests

`expected` now names the record (`observation 2 at TIME=8`), not the wrapper. The wrapper text
is emitted for every `check_absorption_dosing` error there has ever been, so it would pass on a
panic raised by something else entirely.

Six new arms: the EVID=2 true positive (the kind nothing covered), the two bit-identity
controls for 1 and 2, the zero-order **dose** true positive (without which
`forcing_is_read_at(ZeroOrder, _) = false` passes the whole file), the mixed-fraction control,
and an all-kinds in-domain acceptance control.

### 8, 9

`zero_eta` and the `$PK` buffer now live in one caller-owned `TypicalSnapshotScratch`, so a
population sweep allocates once. `zero_eta` is *not* a parameter on purpose — an `eta` argument
that must be all-zero for the function's name to be true is a footgun. `with_capacity` is sized
from the set's own vectors.

### Mutations

Eight, each the minimal edit that removes one of this round's fixes, run with
`--no-fail-fast` and restored by byte copy + `touch` (the harness asserts the tree comes back
byte-identical, and it did). All eight killed:

| mutation | tests killed |
|---|---|
| resets pushed back into the checked set | the reset bit-identity control |
| `forcing_is_read_at` → always `true` (zero-order everywhere) | the zero-order observation control |
| `forcing_is_read_at` → always `false` for zero-order | mixed-fraction **and** the zero-order *dose* arm |
| drop the Σ ≈ 1 completeness skip | mixed-fraction |
| drop EVID=2 rows from `Records` | the EVID=2 arm |
| warning back on the frozen `(covariates, 0.0)` snapshot | the modeled-duration warning |
| absorption message stops naming the record | 4, including the `should_panic` |
| absorption check narrowed to `Doses` | 4 |

Two of these are worth calling out. "Zero-order never checked" is killed by **two** arms and
"drop the completeness skip" by **one**, so they are not the same gate wearing two hats —
which is the trap you flagged on #1229. And "message stops naming the record" is what the
`should_panic` fix buys, and I ran it both ways rather than reasoning it: with the record in
`expected` the mutation gives `FAILED`; with the old
`"absorption input-rate machinery cannot honour"` restored, the *same* mutation gives `ok` —
it survives, because `panic_if_unsupported` emits that wrapper whatever `msg` says. So that
assertion could not observe the thing it exists for.

### Gates

All three legs green at `58e6e886`, plus `tools/preflight.sh` (`fmt check clippy rustdoc docs
public-api`) and the public-API baseline untouched.

| leg | binaries | passed | failed |
|---|---|---|---|
| `--features ferx-core/ci --workspace --lib` | 3 | 4,703 | 0 |
| `--features ferx-core/ci --workspace --tests` | 202 | 5,684 | 0 |
| `-p ferx-core --features ci,markov,nn --tests` | 180 | 5,457 | 0 |

Your three counts plus the 7 new tests exactly (4,703 / 5,677 + 7 / 5,450 + 7).

**One red run on the way there, and it is not this PR's.** `ci,markov,nn` failed once on
`ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` — "a converging
nonlinear SS must not warn". It passes in isolation and the leg reran 4,514 / 0.
`SS_WARN_SINK_READER_GUARD` serializes **readers** only — its own doc says "Writers don't
drain, so only readers need to coordinate", which is true for a reader asserting on *its own*
warning and false for one asserting the sink is **empty**.
`note_ss_nonconvergence_if_capped` (`dosing.rs:557`) writes the process-global sink without
touching the guard, so a concurrent capped equilibration on another harness thread lands
between a reader's `clear` and its `take`. Filed as
[#1289](https://github.com/FeRx-NLME/ferx-core/issues/1289). Both new issues are on #1217 as
step-`f` rows.

### Not fixed

- **3's hoist** — [#1288](https://github.com/FeRx-NLME/ferx-core/issues/1288), measured at 3.4% of a draw, with the `draw == 1` trap
  recorded so the follow-up does not walk into it.
- **The `≤ 0` domain question on `D{n}`/`R{n}`** — still #1284, as you agreed.
- **`E_DOSE_ATTR_NONFINITE` not reaching `predict()`/`simulate()`** — still #1280 / #898.

PR #1292, opened 2026-09-08
feat(tools): modelsearch — ABSORPTION(ZO / WEIBULL) and ELIMINATION as [odes] candidates [closes #1257]

### LINKED ISSUE #1257: feat(tools): modelsearch — ABSORPTION(ZO / WEIBULL) and ELIMINATION candidates as [odes] models, with a per-candidate start budget [#1175]
Split out of #1181 (review on #1256). The first modelsearch (#1256) searches the analytic `pk` templates only: `ABSORPTION(INST / FO)`, `PERIPHERALS(0..2)`, `TRANSITS(n | N, NODEPOT)`, `LAGTIME`. The #1181 coverage table listed `ABSORPTION(ZO)` and `ABSORPTION(WEIBULL)` as reachable, but neither is a `pk` template swap:

- `ZO` on an oral dose is a modeled zero-order input — in ferx a `D1` duration on an IV template with `RATE=-2` **in the dataset** (a data change, not a model edit), or the `[odes]` input function `zero_order(...)`;
- `WEIBULL` is the `[odes]` input function `weibull(td, beta)`;
- `ELIMINATION(ZO / MM / MIX-FO-MM)` has no analytic template at all (the gap #1181 already named).

So all three are the same deliverable: **an `[odes]` candidate family** for modelsearch. What it needs beyond a template swap, from #1181's own list:

1. an `ode_template`-style edit in `ferx-core::edit` that rewrites `[structural_model]`/`[odes]` from an analytic base (the analytic → ODE desugar already exists in the parser for the transit/IG twins; the edit layer needs the inverse direction as text);
2. a **non-uniform runtime budget** — the report already carries per-candidate `seconds` (#1256), but the runner's thread plan assumes candidates cost alike;
3. a **per-candidate start budget** — MM models are the canonical local-minimum case (`docs/examples/multistart.qmd`) and need more `retries` than their analytic siblings, or the search rejects a correct model on a stalled fit;
4. `ABSORPTION(SEQ-ZO-FO)` stays rejected (no template, no input function).

Until this lands, `ferx modelsearch` refuses these features by name at config load (the #1179 coverage check), and `docs/tools/modelsearch.qmd` points here.

roninsightrx:
## Why

Closes #1257, split out of #1181. The first modelsearch (#1256) searched the analytic `pk` templates only. Four MFL values were refused by name at config load because none of them is a template swap:

- `ABSORPTION(ZO)` — a modeled zero-order input, i.e. the `[odes]` input function `zero_order(...)`;
- `ABSORPTION(WEIBULL)` — the `[odes]` input function `weibull(...)`;
- `ELIMINATION(ZO / MM / MIX-FO-MM)` — no analytic template at all (#1181's named gap).

They are **one deliverable**, because they are written the same way: an `ode_template NAME(...)` line, which generates the standard disposition, plus **one** `[odes]` line replacing the `central` equation. Everything else about the candidate — the disposition, the compartment count, a transit chain, the covariate model, the parameters the base already had — is unchanged.

### `ferx-core`

**`pk::ode_template`** — `generate_variant(name, params, input, elimination)` writes the generated disposition with the central compartment's input and elimination terms replaced (`InputForm::{Template, ZeroOrder, Weibull}`, `EliminationForm::{FirstOrder, MichaelisMenten, MixedFoMm}`). `generate` is now that function at its defaults: the ten templates' topology is composed from **one** table (depot + its `ka`, the template's own input term, the central volume, the peripheral chain as `(state, q, vp)`) rather than written out per model, and a snapshot test pins the plain form of all ten byte-for-byte so the refactor changed no generated character.

**`edit`** — `StructuralSpec::ode(input, elimination)` makes a structural edit write it:

- `SetStructural` now swaps in **both directions** (`pk` ↔ `ode_template`) and clears the `[odes]` block when a candidate moves back to an analytic template — an ODE-only input function beside a `pk` line is a hard parse error, so a stale override would make the candidate uncompilable rather than analytic.
- The override set is derived by diffing the variant against the plain generation of the same template, so the default variant writes no override at all and nothing in the edit layer has to know which compartment a given variant touches.
- `ode_template` takes only the disposition roles, so an `f` / `lagtime` binding becomes the **reserved-name alias** an ODE model reads it off (`lagtime = TLAG`) — #735's own technique, from the ODE twin builder.
- **Bug fix, pre-existing:** `prune_unreferenced` now roots those reserved names on an ODE model. The engine reads bioavailability and the absorption lag off an `[individual_parameters]` *name* (`ode_param_slots`), not off any equation, so `F = TVF` was dead by the pruner's reckoning — any edit that pruned on a hand-written `ode(...)` model would have silently taken the dose attribute, its θ and its η with it.
- `StructuralSpec` becomes `#[non_exhaustive]` with `new()` + the `ode(...)` builder, as `NewParameter` did on #1181.

### `ferx-tools`

**`modelsearch::structure`** — `Structure` gains an `elimination` coordinate and `Absorption` gains `Zo` / `Weibull`; `Template` carries the engine. Pharmpy's `not_supported_combo` additions are transcribed (ZO and WEIBULL with transits, WEIBULL with a lag time — ZO *takes* a lag time, as in Pharmpy), and elimination is in none of the pairs, so it composes with every absorption and every compartment count. Inits are Pharmpy's `modeling/odes.py`:

| parameter | init | source |
|---|---|---|
| `KM` | `max(DV)/2`, upper `1.5·max(DV)` | `_do_michaelis_menten_elimination` / `_get_mm_inits` |
| `KM` (ZO elimination) | `min(DV)/100`, **`FIX`** | `set_zero_order_elimination` |
| `CLMM` (MIX-FO-MM only) | `CL/2` | `_get_mm_inits(combined=True)` |
| `DUR` | `2·MAT`, `MAT = 2·t_first` | `_add_zero_order_absorption` |
| `TD` / `BETA` | `MAT / Γ(1 + 1/1.5)` / `1.5` | `set_weibull_absorption` |

**`search::runner`** — two things #1257 asked for:

- **A non-uniform thread plan.** `PoolPlan::from_budget` is one split for a whole batch, which is right only when candidates cost alike. `Candidate::cost` states what one costs relative to the run's ordinary candidate, and the runner plans candidates of **equal cost together, heaviest group first** — so a handful of ODE candidates run few-at-a-time with subject-level threads, and the analytic ones then run wide. A run whose candidates all carry the default cost is one group, planned exactly as before.
- **A per-candidate start budget.** `Candidate::n_starts` becomes a **floor** under `RunOptions::n_starts` rather than a replacement, and a saturable elimination asks for 8 (`docs/examples/multistart.qmd` puts Michaelis–Menten at the top of its local-minimum table; `examples/mm_multistart.ferx` uses 8). A run configured with more `retries` keeps them.

`ABSORPTION(SEQ-ZO-FO)` stays refused, per the issue: it is not one input term on a standard disposition but a depot of its own, filled at a constant rate and emptied by `ka` (`examples/sequential_absorption.ferx`) — generating it needs a template family that does not exist rather than an override.

## Alternatives considered

- **Letting `ferx-tools` write the `[odes]` text.** It would have been a second copy of the disposition transcription, which the "when two implementations disagree, the fix is one implementation" rule exists to prevent. Instead `pk::ode_template` — which the parser already uses — grew the variants, and the edit layer asks it for them.
- **Renaming `CL` to `CLMM` as Pharmpy does.** The edit layer has no rename operation, and adding one would drop the parameter's initial estimate and its η at exactly the move where they matter most. ferx binds the same parameter under the name the base gave it; only `MIX-FO-MM`, which genuinely has two clearances, introduces a `CLMM`. Stated as a deviation on the docs page.
- **Refusing `LAGTIME(ON)` on an ODE candidate**, to avoid the alias. It would be a real coverage hole for a combination Pharmpy allows (`ELIMINATION(MM)` on a base that already has a lag time), so the alias is emitted instead — and the pruner bug it exposed is fixed rather than worked around.
- **Reverting to one thread plan and just ordering the candidates.** Rayon work-steals, so ordering alone still leaves the expensive fit running single-threaded at the tail. Separate plans per cost tier give it the whole budget.

## Cross-repo dependency

`ferx-r` does not construct `StructuralSpec` (checked); the new field arrives behind `new()` and `#[non_exhaustive]`.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — regenerated. Added: `pk::ode_template::{generate_variant, InputForm, EliminationForm}` (needed by `edit` to write the variant, and re-exported from `edit` so a caller building a `StructuralSpec` does not reach into the PK module for two of its fields); `edit::StructuralEngine` and `StructuralSpec::{new, ode}` (needed by `ferx-tools::modelsearch` to state an ODE candidate); `StructuralSpec` is now `#[non_exhaustive]` and gained `engine`. In-repo struct literals were converted to `new()`; there is no ferx-r caller.
- [ ] None

## Numerical validation

**NONMEM 7.5.1 anchor** (`crates/ferx-tools/tests/nonmem/modelsearch_anchor/mm_*`, `the_michaelis_menten_candidate_evaluates_as_nonmem_does`, slow-gated). Both sides **evaluate** (`MAXEVAL=0` / `maxiter = 0`) at one stated parameter vector — the base's own θ plus `KM = max(DV)/2 = 6.9409` — because the object under test is the generated equation, and comparing two *minimisations* of a Michaelis–Menten model would compare two optimizers on the canonical multimodal surface instead.

| model | ferx | NONMEM 7.5.1 | \|Δ\| |
|---|---:|---:|---:|
| first-order elimination (analytic, the control) | −182.66670809 | −182.66670808 | 9.7 × 10⁻⁹ |
| `ELIMINATION(MM)` (`ode_template` + one override) | 929.63056169 | 929.63055241 | 9.3 × 10⁻⁶ |

The first-order row is the control: it says the agreement on the second row belongs to the saturable term and is not an accident of the dataset, and the 1112-unit gap between the rows is what makes the anchor unable to pass on a candidate whose override the engine ignored. Both sides integrate to a stated tolerance (`TOL=9 ATOL=12` against `ode_reltol = 1e-10`, `ode_abstol = 1e-12`), so the 9.3e-6 residual is solver error rather than model error — at ferx's *default* `ode_reltol = 1e-4` the same comparison is 0.017 out, which is the FOCEI objective amplifying integration error. Tolerance measured, not assumed: 1e-3, ~100× the worst realised error.

The NONMEM `$DES` right-hand side was itself checked against a **third computation** — a 30-line RK4 integration in Python of `dA2/dt = KA·A1 − (CL·KM/(KM + A2/V))/V·A2` — which reproduces its `PRED` at 0.5 / 1 / 2 h to 6 significant figures (5.256832, 7.715356, 9.371420).

**Candidate ≡ hand-written** (`tests/model_edit_equivalence.rs`, 5 new cases): each of `ELIMINATION(MM)`, `ELIMINATION(MIX-FO-MM)`, `ABSORPTION(ZO)`, `ABSORPTION(WEIBULL)` and the default ODE variant agrees with its **fully hand-written** `ode(obs_cmt=…, states=[…])` twin — every equation typed out, `obs_scale` stated — on `predict()` bit for bit and on the evaluated objective bit for bit. The twin is deliberately not another `ode_template`, which would compare the generator against itself.

- [x] Compared against: NONMEM 7.5.1, Pharmpy 2.2.0 (`modeling/odes.py` for the parameterisation and every init above)

## Performance
- [x] No significant impact expected on the engine. The runner's plan changes: with uniform costs it is byte-identical to before (one group), and with mixed costs it trades some overlap between tiers for giving the expensive tier the whole budget.

## Tests
- [x] Unit tests: `pk/ode_template.rs` (the ten-template byte-identity snapshot, `generate` ≡ the default variant, each replacement, the two refusals); `src/edit/mod_tests.rs` (6: the template line, the single override, the default variant writing no `[odes]`, the return to `pk` clearing it, the alias and its absence on a reserved spelling, and ODE ≡ `pk` on everything but the keyword); `structure_tests.rs` (13: engine/cost/starts, the templates and roles, Pharmpy's new pairs in both orders, `onto_space`, the Γ identity, and the inits for every new parameter); `runner_tests.rs` (4: unequal costs planned separately heaviest-first, equal costs still concurrent, a shared raised cost sharing a group, the start floor); `modelsearch/mod_tests.rs` (2).
- [x] Tier-2: `every_ode_candidate_compiles_and_evaluates` — 3 absorptions × 4 eliminations, all 11 candidates generated from an analytic base, compiled and evaluated on warfarin, each asserted to be one `ode_template` line and exactly one `central` override.
- [x] Tier-3: the NONMEM anchor above.
- **Mutation checks.** Disabling the pruner's reserved-name root reddens two edit tests (`…aliases_a_non_reserved_lagtime_binding`, `…does_not_alias_a_reserved_spelling`) and nothing else — verified by running it. Making the runner use one plan reddens `candidates_of_unequal_cost_are_planned_separately_heaviest_first` on `threads_per_fit`; sorting the groups ascending reddens its `heavy_at == 0`. The "always override `central`" weakening of the diff-based override set reddens `set_structural_ode_default_variant_writes_no_override`.
- The start floor is asserted on `effective_starts`, the function `compile_and_fit` calls, rather than by re-deriving the formula in the test — the effective count is not observable on a `FitResult`.

## Example
- [x] `examples/warfarin_modelsearch.ferxsearch` documents how to add `ELIMINATION([FO,MM])` / `ABSORPTION([FO,ZO,WEIBULL])` and what it costs; the shipped space stays analytic so the example still runs in seconds.

## Docs
- [x] `docs/tools/modelsearch.qmd` — the coverage table rows flip to ✓, "The elimination gap" is replaced by "The `[odes]` candidates" (what each candidate is, the clearance-name deviation, the cost, the start budget, the inits, and why an `ode_template` *base* is still refused), and the NONMEM section gains the evaluation anchor above.
- [x] `docs/tools/search.qmd` coverage table, `docs/cli.qmd`.
- [x] `CHANGELOG.md`.

## Checklist
- [x] No `Dual2` path touched — the variants are `[odes]` *text*; the sensitivities come from the existing ODE provider, which already differentiates `zero_order` / `weibull` / an arbitrary RHS.
- [ ] `Cargo.toml` version bumped — additive API, not bumped

## Docs & examples
- [x] No new `.ferx` example or data file; the new `.ferxsearch` + `.ferx` under `tests/nonmem/modelsearch_anchor/` are test fixtures.
- [x] Rebuilt ferx-core and ran the example via the release CLI.
- [ ] ferx-r rebuild — not needed for this PR

## Reviewer hints

- `src/pk/ode_template.rs` is the heart: the topology table and the three-part `central` composition. `plain_form_is_byte_identical_across_every_template` is the guard that the refactor changed nothing.
- `set_structural` in `src/edit/apply.rs`: the override-by-diff, the alias handling, and `set_odes_block`.
- The pruner change is four lines and is a behaviour change for *any* ODE model, not only for search candidates — worth a look.
- `structure.rs::template()` and `structural_spec()` are where a coordinate becomes a spec; `ODE_ONLY_ROLES` is the reason a role can be a parameter without being a binding.
- `cost_groups` in `runner.rs` is 20 lines.

## Open questions

- `MM_STARTS = 8` is the top of the multistart page's 4–8 range for Michaelis–Menten. It is not configurable per run; `[run] retries` can only raise it. If that turns out to be too blunt, a `[modelsearch] mm_starts` key is the obvious follow-up.

https://claude.ai/code/session_01N2VfH2igJkyhiYLss8ZXaw


src/edit/apply.rs
```diff
@@ -194,9 +274,68 @@ fn set_structural(text: &mut ModelText, spec: &StructuralSpec) -> Result<(), Str
...
     }
 
+    // ── The reserved-name aliases ───────────────────────────────────────────
+    // Dropped and rewritten together, so a move that changes engine or drops
+    // the role cannot leave one behind — an `f = BIOAV` surviving into an
+    // analytic candidate would declare a parameter named `f` beside the `pk`
+    // line's own `f=` role. Only a model whose disposition line this layer
+    // wrote is touched: a hand-written `ode(...)` model is refused above, so
+    // the only alias reachable here is one of ours.
+    if was_ode_template {
```
roninsightrx, line 313: [P1] Preserve genuine reserved declarations when rewriting an ODE template

`was_ode_template` does not imply that every bare reserved assignment is an alias created by this function. In particular, a normal ODE model can declare `F = TVF` or `ALAG = TVLAG` directly. This block deletes that real declaration, and the re-add loop then deliberately skips it because the binding target already has a reserved spelling. I reproduced this by applying the same ODE `StructuralSpec` with `lagtime=ALAG` twice: the second edit removed both `ALAG = TVLAG` and `theta TVLAG`, silently resetting the lag to zero. Please distinguish generated aliases from canonical reserved declarations (and add an ODE-to-ODE regression test) before deleting them.

crates/ferx-tools/src/modelsearch/structure.rs
```diff
@@ -865,9 +1165,16 @@ pub fn structural_spec(
...
+    // below can name it.
+    let mut ode_names: HashMap<&str, String> = HashMap::new();
     for role in &template.roles {
+        let ode_only = ODE_ONLY_ROLES.contains(role);
         let rebinding = *role == "n" && rebind_n;
-        if let Some(var) = bound.get(slot_of(role)).filter(|_| !rebinding) {
+        // An ODE-only role is never on the parent's template line, so the
+        // "keep the parent's variable" branch cannot apply to it; reuse of an
+        // existing declaration happens by name, below.
+        if let Some(var) = bound.get(slot_of(role)).filter(|_| !rebinding && !ode_only) {
```
roninsightrx, line 1248: [P1] Carry F and lagtime through an ODE parent in stepwise search

After the first ODE move, `parent_template` is parsed from an `ode_template` line, which cannot contain `f` or `lagtime`. Consequently this lookup cannot recover either logical binding on the next step. Bioavailability is then omitted from the child spec entirely; a non-canonical lag binding such as `TLAG` is replaced by a fresh default `ALAG`. Combined with the alias cleanup in `set_structural`, a path such as `ELIMINATION(MM)` followed by `PERIPHERALS(1)` silently resets F to 1 and either drops or reinitializes lagtime, even though neither coordinate was changed. The node/edit metadata needs to preserve or recover these reserved bindings independently of the template-line arguments, with a multi-layer regression test.

crates/ferx-tools/src/modelsearch/structure.rs
```diff
@@ -911,9 +1223,38 @@ pub fn structural_spec(
...
+            // starts at half the first-order one it is added beside.
+            "clmm" => (0.5 * cl, 0.0, 1e6),
+            // `_do_michaelis_menten_elimination`: `KM` starts at half the
+            // largest observation, bounded above by 1.5× it, with Pharmpy's
+            // cap for the NONMEM-specific overflow case. Zero-order
+            // elimination is that model with `KM` **fixed** far below the
+            // observed concentrations, so the flux is `CLMM·KM` — a constant
+            // — wherever the data live (`set_zero_order_elimination`).
+            "km" => {
+                let init = if target.elimination == Elimination::Zo {
```
roninsightrx, line 1316: [P1] Retain Pharmpy's negative-DV guard for fixed KM

Pharmpy 2.2.0's `set_zero_order_elimination` resets `obs.min()/100` to `0.01` when that value is negative. This transcription omits the guard, despite the `dv_min` documentation immediately above saying it is floored. A dataset with a legitimate negative DV (common under additive residual error) therefore produces a fixed negative KM; the Michaelis-Menten denominator becomes singular when `C = -KM`, and an all-negative dataset also gets a negative upper bound. Please apply the 0.01 fallback and cover zero/non-positive observation ranges as well, since KM=0 makes the emitted textual RHS evaluate `0/0` at zero central amount.

roninsightrx (commented):
The generated-equation, modelsearch, runner, and end-to-end ODE-candidate tests pass locally. I am requesting changes for the three correctness issues called out inline; the first two can silently change F/lagtime on multi-step searches, and the third can generate an invalid zero-order-elimination candidate for legitimate data.

src/edit/apply.rs
roninsightrx: Fixed in fd6a7be8, and you are right about the mechanism — I reproduced your `lagtime=ALAG` twice case before changing anything: the second edit deleted `ALAG = TVLAG`, the rewrite loop skipped it because `ALAG` already routes, and the pruner then took `TVLAG`. Silent lag reset to zero.

The rule was wrong in what it keyed on. `was_ode_template` says this layer wrote the *disposition line*, not that every reserved assignment under it is a bridge. The discriminator is the right-hand side, not the left:

```
lagtime = TLAG      alias    — TLAG is another individual parameter
ALAG    = TVLAG     declaration — TVLAG is a θ
F       = THETA_F   declaration
```

`is_reserved_alias` now requires the RHS to be a bare identifier that `[individual_parameters]` itself declares. A θ or an expression means the line *is* the attribute and is left alone.

Three tests, all mutation-checked against the old broad rule:

- `a_second_ode_swap_keeps_a_canonical_reserved_declaration` — your exact repro, asserting the second edit is byte-identical to the first (idempotence), plus that `theta TVLAG` survives. Fails under the old rule.
- `an_ode_swap_keeps_a_bioavailability_declaration` — deliberately `F = THETA_F` rather than an expression, since a bare identifier is the shape where the discrimination is load-bearing. Fails under the old rule.
- `a_stale_alias_is_still_dropped_when_the_binding_moves` — the property I had to not give up: a bridge whose binding moved to a freshly declared `ALAG` still goes, or `ode_param_slots` would see the slot declared twice.

crates/ferx-tools/src/modelsearch/structure.rs
roninsightrx: Fixed in fd6a7be8. Confirmed both halves of what you describe: off an ODE parent, `bound` had no `f` (so the child spec omitted it entirely) and no `lagtime` (so the role fell through to `default_name("lagtime") = ALAG` and minted a fresh parameter at `t_first/2`, leaving `TLAG` unreferenced for the pruner).

`structural_spec` now recovers both from the parent's `[individual_parameters]` when the parent's line is not a `pk` line, reading them the way the engine resolves them (`reserved_binding`) — an alias resolves to what it bridges to, a reserved name assigned a θ or an expression resolves to itself. `bound` had to become `HashMap<&str, String>` for that.

I kept it as a *recovery* rather than new node metadata on purpose: the model text is where an ODE candidate actually states these, so recovering from it stays correct for a text that arrives from a resumed journal or from `models/<id>.ferx`, where any in-memory side channel would not.

Tests:

- `reserved_bindings_are_recovered_from_an_ode_parents_declarations` — both shapes, plus the nothing-to-recover case.
- `a_second_step_off_an_ode_parent_carries_f_and_lagtime` — the localiser: `ELIMINATION(MM)` → `PERIPHERALS(1)`, asserting `lagtime=TLAG` and `f=F` are in the child's bindings, that no `ALAG` was minted, and that the layer's own move still happened.
- `a_second_layer_off_an_ode_parent_keeps_bioavailability_and_the_lag_time` — the multi-layer one you asked for, through `search()` with `exhaustive_stepwise` on a base carrying `f=F, lagtime=TLAG`, asserting on the layer-2 model *text*: `F` and `THETA_F` present, `lagtime = TLAG` present, `theta TVLAG(0.4` still the base's value rather than a re-derived one, and no second `ALAG`.

All three die when the recovery is disabled.

crates/ferx-tools/src/modelsearch/structure.rs
roninsightrx: Fixed in fd6a7be8, and thank you for catching that the doc comment already claimed a floor I never wrote — that is the worse half of the bug.

I went slightly further than restoring Pharmpy's guard, because Pharmpy only guards one of the three ways this goes wrong:

| range | Pharmpy | before | now |
|---|---|---|---|
| `min(DV) < 0` | `0.01` | negative fixed `KM` | `0.01` |
| `min(DV) == 0` | `0.0` | `KM = 0` | `0.01` |
| `max(DV) <= 0` | negative upper bound | negative upper bound | `1.5` |

The zero case is not merely degenerate: with `KM = 0` the emitted right-hand side is `(CL * KM / (KM + central / V)) / V * central`, which is `0/0` at `central = 0` — every subject's first record — so following Pharmpy exactly there would generate a model that cannot be evaluated at all. Both halves are floored to `1.0`, the value Pharmpy itself uses when a model carries no dataset, so `KM` lands on `0.01` fixed (ZO) or `0.5` with upper `1.5` (MM). Stated as a deviation on the docs page.

The floor is on the *data-derived* values in `Defaults::new` rather than at the `"km"` init, so `dv_min` and `dv_max` are positive by construction wherever they are read — the doc comment and the code now agree.

`a_non_positive_observation_range_falls_back_rather_than_fixing_km_at_zero` covers all three ranges × `ZO`/`MM`, asserting `init > 0`, `upper > init` and `lower >= 0`, then pins Pharmpy's own `0.01` on the negative-minimum case — and that `max(DV) = 8` still sizes the bound at `12.0` there, so the floor does not quietly rescale a dataset that is only partly negative.

roninsightrx:
All three addressed in fd6a7be8 — each reproduced before it was changed, each fix mutation-checked against the code it replaces. Replies inline; the short version:

1. **Alias cleanup ate real declarations.** Your `lagtime=ALAG` twice repro was exact. The rule keyed on `was_ode_template`, which says this layer wrote the disposition *line*, not that every reserved assignment under it is a bridge. It now keys on the right-hand side: another individual parameter is a bridge, a θ or an expression is the attribute.
2. **F/lagtime lost off an ODE parent.** Recovered from the parent's `[individual_parameters]` — the two shapes the engine itself resolves — rather than from a template line that structurally cannot carry them. Kept as a text recovery rather than node metadata so a resumed journal or a `models/<id>.ferx` read stays correct.
3. **Fixed KM unguarded.** Restored Pharmpy's negative-DV fallback and extended it to the two cases Pharmpy leaves open (`min(DV) == 0`, which makes the generated RHS evaluate `0/0` at `central = 0`; `max(DV) <= 0`, which gives a negative upper bound), floored to the value Pharmpy uses with no dataset. The doc comment that already promised this now matches the code.

Seven tests added: three in `src/edit/mod_tests.rs`, three in `structure_tests.rs`, and the multi-layer `search()` regression you asked for in `modelsearch/mod_tests.rs`.

Green locally: `tools/preflight.sh`, `cargo test --lib` (4255 + 543), the Tier-2 end-to-end suite (8), and the NONMEM anchor unchanged at 9.3e-6 / 9.7e-9.

roninsightrx:
Separately from the review: I noticed the `Tests + coverage (core)` job took **1h44m and then 2h29m** on this branch against **18–32 min** on recent `main` runs, and chased it rather than calling it runner variance.

It was mine. Measured, not guessed: four of the new ODE cases in `tests/model_edit_equivalence.rs` account for **402 s of that file's 435 s** locally, and the coverage job runs it instrumented, unoptimised and on one rayon thread.

The cost is the outer loop rather than the model — each outer iteration finite-differences the whole population objective once per free parameter, and these twins integrate numerically where the file's other cases have a closed form. The claim under test is that two spellings are one model: same θ and η vectors, `predict()` bit for bit, and the same objective at the point they both start from. Iterating only re-asserts that equality at a second point the optimizer picked, and the one thing an evaluation could miss — a difference in parameter *order*, which moves the optimizer without moving the start — is already pinned by the `theta_names` / `eta_names` assertions above it.

d208be9e therefore parameterises the harness: `assert_same_model` is byte-for-byte what it was for the twelve pre-existing cases, and three of the four ODE twins evaluate instead. The Michaelis-Menten one keeps its two outer iterations, so the optimizer path over a generated `[odes]` model is still exercised somewhere — on the variant the NONMEM anchor covers.

**435 s → 222 s**, all 17 cases still passing.

The other new Tier-2 test (`every_ode_candidate_compiles_and_evaluates`, all 11 candidates compiled and evaluated) is 23 s locally and I left it alone — it is the one that would catch a role table disagreeing with the parser's, and it is not where the time was going.

PR #1293, opened 2026-09-08
ci(coverage): run debug_assert guards under the coverage profile [closes #1248]

### LINKED ISSUE #1248: Codecov: debug_assert! condition lines are permanently zero-hit, lowering the patch-coverage ceiling
## Summary

`debug_assert!` condition lines are emitted by `llvm-cov` as coverage **regions that can never be executed**, because every Codecov-producing job builds with a release-derived profile where the macro compiles to nothing. They are therefore permanently-missed patch lines, and they lower the ceiling of the ≥90% patch gate for any PR that adds a guard.

This is the second half of #344. The first half — *the guards never run in CI* — was fixed by #1245, which added a `Tests (debug-assertions)` job on the dev profile. That job runs plain `cargo test` and uploads **no coverage**, so the coverage ceiling described here is unchanged by it. Splitting it out rather than letting #1245 close #344 silently.

## The mechanism

The per-PR patch gate runs `--profile ci-fast`, which inherits `ci-test`, which inherits `release`. Nothing in that chain sets `debug-assertions`, so:

- the macro expands to nothing,
- but `llvm-cov` still emits a region for the condition expression,
- so a multi-line `debug_assert!` contributes guaranteed-missed lines to any PR that adds one.

Measured on #344 (local replication of CI's own patch-gate command, reproducing Codecov's reported number exactly), `src/sens/ode_provider.rs`:

```
debug_assert! condition lines found: 4
  of those, ZERO-hit : 4
  of those, hit >0   : 0
  lines: [2029, 5261, 5584, 5695]
```

All four. Line 2029 predates the PR that surfaced it, which is what makes this systemic rather than one PR's problem.

**Impact seen in practice.** On #1136, 3 of the 10 uncovered patch lines were these conditions; combined with 4 lines of a separately-filed unreachable branch (#1160) the ceiling was **49/56 = 87.5%** against a 90% gate — unreachable by any test. It merged red with the arithmetic documented in the PR body.

**A related gotcha for whoever picks this up.** `src/**/*_tests.rs` contributes **zero** patch lines — `llvm-cov` emits no `SF:` record for them at all. So "add more tests" is not a route out of a patch-gate failure when the uncovered lines are structurally unreachable.

## Options

1. **Set `debug-assertions = true` on `ci-fast`/`ci-test`.** Fixes both costs at once — the guards run *and* the lines become coverable. Costs wall-clock on the fit-heavy jobs, which is why those profiles inherit `release` in the first place. #344 explicitly ruled this out for the *first* half of the problem; it may still be the right answer for the coverage half, but the cost has to be measured rather than assumed.
2. **Upload a coverage report from the new `Tests (debug-assertions)` job** under its own Codecov flag. Codecov merges flags by summing hits, so those lines would read as hit in the merged report. Cost: that job currently runs ~30 min uninstrumented (396s + 1223s of tests on a 2-core runner); instrumenting it would make it clearly the critical path, and it needs a `codecov.yml` flag entry. Needs a measured trial before committing.
3. **`codecov.yml` exclusion.** Codecov has no line-level "ignore this pattern" for the patch status, so this is not straightforwardly available.
4. **Accept it as policy**: a sub-90% patch whose shortfall is entirely `debug_assert!` conditions is mergeable, documented in the PR body. That is what #1136 did, but ad hoc rather than as policy.

Explicitly **not** an option: reformatting guards onto a single line so the condition is not its own line. That clears the gate without changing anything real, and makes the guards less readable — gaming the metric.

## Not blocking

Nothing regresses while this is open; the ceiling is the same as it has been. #1245 makes the *guards* live, which was the more urgent half.

Follow-up to #344 · context in #1245

roninsightrx:
## Why

`debug_assert!` condition lines are emitted by `llvm-cov` as coverage regions that **can never be executed**: every Codecov-producing job built a release-derived profile, where the macro expands to nothing but the region is still emitted. They were therefore permanently-missed patch lines, and they lowered the ceiling of the ≥90% patch gate for any PR that added a guard. On #1136 the ceiling was **49/56 = 87.5%** — unreachable by any test — and that PR merged red with the arithmetic written into its body.

This is the second half of #344. The first half (*the guards never run in CI*) was fixed by #1245, which added a `Tests (debug-assertions)` job on the dev profile that uploads no coverage. `src/**/*_tests.rs` contributes zero patch lines (no `SF:` record at all), so "add more tests" was never a route out of it.

## What changed

One new profile, and both per-PR coverage jobs build under it:

```toml
[profile.ci-cov]
inherits = "ci-fast"
debug-assertions = true
overflow-checks = true
```

**Measured** (15 cores, stable, `--features ci`) — the issue asked for cost measured rather than assumed:

| | guards off (`ci-fast`) | `debug-assertions` | `+ overflow-checks` (`ci-cov`) |
|---|---|---|---|
| `--workspace --tests` | 133 s, 5727 passed | **133 s**, 5732 passed | 147 s, 5732 passed, 0 failed |
| `--lib` | 5.76 s, 4232 passed | 6.45 s, 4237 passed | — |
| clean lib compile (×2) | 92 s / 98 s | 93 s / 87 s | — |
| `debug_assert` lines 0 → >0 | — | **63** (`--lib` scope alone) | — |

The guards are free because these profiles are release-optimised; the within-profile compile spread exceeds the between-profile difference. Zero-hit lines inside guard spans drop 90 → 37 in `--lib` scope (the residue is panic-arm message lines, which a guard that never fires cannot cover — that part is irreducible).

This was also the first time the ~180 guards have ever run against the `tests/` binaries — #1245's job is `--lib` only. No latent failure.

**The `Tests (debug-assertions)` job is retired.** Its two commands were `--lib` on `ci` and on `ci,markov,nn`, a strict subset of what `Tests + coverage (core)` (`--workspace --tests`) and `Tests + coverage (TTE/CTMM endpoints)` (`--lib`, `ci,markov,nn`) now run. Measured on three `main` runs it cost **32–42 min**: the CI critical path, ~2× the next-longest job.

| job | 34238183362 | 34209580799 | 34184960371 |
|---|---|---|---|
| `Tests (debug-assertions)` | 42m | 32m | 42m |
| `Tests + coverage (core)` | 18m | 18m | 15m |
| `Tests + coverage (TTE/CTMM endpoints)` | 16m | 15m | 12m |

`overflow-checks` was the only thing that job had which the coverage jobs did not, which is why it is set explicitly. It is **not** derived from `debug-assertions` — Cargo defaults it per profile, and a `release`-derived profile setting only `debug-assertions = true` still compiles with `-C overflow-checks=off` (verified with `cargo build -v`).

`tools/preflight.sh debug-assertions` now builds `ci-cov` too, so the local gate matches CI. Its test phase drops from 80 s + 198 s (dev) to **11.6 s + 16.3 s**.

### The guards-OFF lane (added in review)

Retiring the job and guarding both coverage jobs left **no per-PR job running the crate the way a user's release build behaves** — guards compiled out, overflow wrapping. The remaining `ci-test` runs are scheduled, manual or post-merge, so a release-only regression could have merged green. Caught by @roninsightrx's review; the first push of this PR had that hole.

It bites hardest on the three tests this PR rewrote: their NaN-fallback arms are the documented *user* behaviour, and under `ci-cov` only the panic arm runs.

`Tests (release semantics)` is the counterweight — `--lib` on the `ci,markov,nn` union under `ci-fast`, delegating to a new `tools/preflight.sh release-semantics` group. It runs in parallel with the coverage jobs and is shorter than either, so it should add no wall-clock. It deliberately does **not** restore guards-off `tests/` binaries: that is the 120-odd binary compile, and `slow-tests.yml` plus the nightly coverage job still run those on `ci-test` with the guards off. Say if you want the integration tier back too.

Each lane now arms a canary in its own direction — `FERX_REQUIRE_DEBUG_ASSERTIONS=1` on the coverage jobs, `FERX_REQUIRE_NO_DEBUG_ASSERTIONS=1` on the release lane. The second matters for the same reason as the first: without it the release lane silently becomes a duplicate of the coverage jobs the moment `ci-fast` acquires `debug-assertions`, green throughout.

### Three tests that would have silently disappeared

`src/types_tests.rs` (×2) and `src/estimation/outer_optimizer_tests.rs` (×1) were `#[cfg(not(debug_assertions))]` **on purpose**, asserting the release-only fallback behind a misuse guard — "the profile CI and coverage use", as the comment said. Under `ci-cov` they do not fail; they stop existing, in the only jobs that ran them. Caught by diffing the two lcov reports: `types.rs:3490`/`:3494` went 2 hits → 0 and `outer_optimizer.rs:692` went 1 → 0, the only three lines in the crate to regress.

They now assert **both** behaviours (`catch_unwind` + `cfg!(debug_assertions)`), so the guard firing is checked as well as the NaN fallback, and they run under either profile. `docs/development/sdlc.qmd` warns against the `cfg`-gated shape for the same reason.

## Alternatives considered

- **Flip `ci-fast` itself** (the issue's option 1). Rejected: `ci-fast` is not CI-only despite the name — `tools/vi-nuts-anchor/run.sh` and `tools/vi-emvi-comparison/run.sh` build `target/ci-fast/ferx` with it for numerical anchor runs, and this would put the guards and the `#[cfg(debug_assertions)]` MR ODE-twin verifier into them. A separate profile adds **no** CI compile: both coverage jobs already carry coverage `RUSTFLAGS` and their own `rust-cache` keys, so their artifacts were never shared with any other job — this renames the profile they build under.
- **Instrument the `Tests (debug-assertions)` job** (option 2). Dead on arrival: at 32–42 min it is already the critical path, and instrumenting it would make that worse.
- **`codecov.yml` exclusion** (option 3). No line-level ignore exists for the patch status.
- **Accept as policy** (option 4). Fixes nothing; the ceiling stays.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

CI-only. No estimation, PK, or stats code is touched; the full `--workspace --tests` suite is bit-identical in outcome (5727 → 5732 passed, the delta being the three tests that now run under both profiles plus the canary).

## Performance
- [x] Faster — CI critical path `32–42m` (before) → **`~28m`** (measured on this PR), by retiring a job whose work is now a 0–14 s addition to jobs that had to run anyway.

**Correction.** An earlier draft of this body said `~18m`, extrapolated from local hardware before any CI run of the new profile existed. The measured figure on this PR is ~28 min. Two runs of the same commit, cold and warm:

| job | run 1 (cold: new profile hash ⇒ full rebuild) | run 2 (warm) |
|---|---|---|
| `Tests + coverage (core)` | 21m | 28m |
| `Tests + coverage (TTE/CTMM endpoints)` | 28m | 16m |

They swap between runs, which is the ~5 min run-to-run variance this repo already documents (#969) — so read the pair as "critical path ≈ 28 min", not as either column. `Tests (release semantics)` runs in parallel and is shorter than both, so it should not extend that; the number to check on this PR's run is whether it does.

## Tests

`the_coverage_jobs_build_the_guarded_profile` (new) pins that every `cargo llvm-cov` feeding the patch gate builds a profile resolving to `debug-assertions = true`, and that the job arms `FERX_REQUIRE_DEBUG_ASSERTIONS`. The guard test's **ban** on `--profile` in the `debug-assertions` group — correct while every named profile inherited `release` — becomes a **positive** check against Cargo.toml, resolving `inherits` transitively.

**Mutation-tested, each mutation reddening its own side:**

| mutation | result |
|---|---|
| `[profile.ci-cov] debug-assertions = false` | 2 tests fail |
| coverage job reverted to `--profile ci-fast` | fails, names `coverage-pr` |
| `FERX_REQUIRE_DEBUG_ASSERTIONS` deleted from the job | fails |
| preflight group drops `--profile ci-cov` | fails |
| canary armed on `ci-fast` / on `ci-cov` | **fails** / passes |
| release lane drifted to `--profile ci-cov` | fails |
| inverse canary arming deleted | fails |
| inverse canary on `ci-cov` / on `ci-fast` | **fails** / passes |

One of these found a real hole rather than confirming a belief: the canary check was first written as `body.contains("FERX_REQUIRE_DEBUG_ASSERTIONS")`, and **passed with the `env:` key deleted** — the comment block explaining the variable satisfied it. This file already warns about exactly that trap one test above; it caught me anyway. It now reads non-comment lines only, and the comment says so.

## Example

## Docs
- [x] `docs/` updated for user-visible changes — `docs/development/sdlc.qmd` §8.2 rewritten: why the coverage jobs build `ci-cov`, the measurement table, and the `#[cfg(not(debug_assertions))]` warning for test authors
- [x] `CHANGELOG.md` — N/A (CI only)

Stale prose corrected where the old arrangement was asserted as fact: `src/lib.rs`, `src/lib_tests.rs`, `tools/preflight.sh`, and `src/pk/modified_release.rs`, whose comment said its `#[cfg(debug_assertions)]` ODE-twin verifier "never runs in CI — that is the already-tracked #344". It does now, at 0 s measured cost; the 577 s figure in that file is the unoptimised dev profile and is explicitly flagged as not the CI cost.

## Checklist
- [x] `tools/preflight.sh debug-assertions` green — 4238 + 4577 passed, 0 failed, canary armed
- [x] `tools/preflight.sh release-semantics` green — inverse canary armed
- [x] No version bump needed

## Reviewer hints

Focus on the retirement of `Tests (debug-assertions)` — that is the part beyond the issue's scope, and it undoes a decision made in #1245 three days ago. The claim is subset-plus-`overflow-checks`; if you disagree that the coverage jobs' `--tests`/`--lib` runs across `ci` and `ci,markov,nn` cover it, that is the thing to push on.

Second: whether `Tests (release semantics)` is targeted *enough*. It restores guards-off coverage for every `src/` unit test but not for the `tests/` binaries, which ran guards-off before this PR. I judged the integration tier's guards-off risk low (the only `#[cfg(debug_assertions)]` production code is the MR ODE-twin verifier, a dev aid) and the compile cost high, but that is a judgement call and the reviewer's to overrule.

The panic-hook swap raised in review is **gone**, not mitigated: both sites now leave the hook alone and accept three expected panic messages in the log. The interleaving described (A takes normal, B takes A's silent hook, A restores normal, B restores silent ⇒ process permanently silenced) is real and immediate restoration does not fix it.

Skimmable: the comment churn in `ci.yml`, `preflight.sh` and `sdlc.qmd`.

## Open questions

The nightly full-coverage job stays on `ci-test` (guards off) deliberately — it uploads the `slow` flag, which `codecov.yml` excludes from the patch status, and its 5–60 min Tier-3 fits want the ThinLTO speedup (#969). Codecov merges flags by summing hits, so its guards-off report cannot un-cover what `fast` covers. Say if you would rather it were consistent.

https://claude.ai/code/session_01QPbLmymWZEmMWKmgJDN6yK

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,53 @@ jobs:
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  # The guards-OFF lane, and the counterweight to `[profile.ci-cov]` (#1293 review
+  # of #1248). Both coverage jobs now build with `debug-assertions` and
+  # `overflow-checks` ON, which is what makes the guards' condition lines coverable
+  # — but it also meant that, for one commit, NO per-PR job ran the crate the way a
+  # user's release build behaves. The remaining `ci-test` runs are scheduled,
+  # manual, or post-merge, so a release-only regression could have merged green.
+  #
+  # That is not abstract. `continuous_value()` (`src/types.rs`) and
+  # `max_scaled_deviation()` (`src/estimation/outer_optimizer.rs`) each have two
+  # correct behaviours: a misuse guard that fires where `debug_assert!` is live, and
+  # a NaN fallback that is what users actually get. Under `ci-cov` only the panic
+  # arm runs. This job is where the other arm is checked before merge.
+  #
+  # Cheap, and deliberately narrower than the coverage jobs: `--lib` only, on the
+  # `ci,markov,nn` union of production cfgs. It runs in parallel with the two
+  # coverage jobs and is shorter than either, so it adds no wall-clock to the
+  # critical path. It does NOT restore guards-off `tests/` binaries — that is the
+  # 120-odd binary compile, and `slow-tests.yml` plus the nightly coverage job
+  # still run those on `ci-test` with the guards off.
+  #
+  # Commands live in `tools/preflight.sh` (group `release-semantics`) for the same
+  # reason as `Check`/`Clippy`/`Format` (#1157), and they arm the INVERSE canary —
+  # `FERX_REQUIRE_NO_DEBUG_ASSERTIONS=1`, which fails the run if `debug_assert!`
+  # turns out to be live here after all. Without it the whole lane would rest on
+  # `ci-fast` never acquiring `debug-assertions`: it would silently become a second
+  # copy of the coverage jobs, green throughout.
+  release-semantics:
+    name: Tests (release semantics)
+    runs-on: ubuntu-latest
+    # Stable, like the coverage jobs: no nightly-only code, and `rust-cache` keys on
+    # the rustc version, so nightly would cold-rebuild the dep graph daily.
+    env:
+      RUSTUP_TOOLCHAIN: stable
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install stable toolchain
+        run: rustup toolchain install stable --profile minimal
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          # Own key: uninstrumented `ci-fast` shares artifacts with neither the
+          # plain `ci` (check/clippy) caches nor the instrumented `ci-cov` ones.
+          shared-key: ci-release-semantics
+
+      - name: cargo test --lib with debug-assertions OFF (via tools/preflight.sh)
+        run: tools/preflight.sh release-semantics
+
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,9 +115,17 @@ jobs:
       # `src/nn`, `src/estimation/nn_theta_gradient` and the feature-gated branches
       # of the shared modules (parser, sens provider, io/output) live there — for
       # `nn` that is the bulk of the surface.
+      #
+      # `--profile ci-cov` and `FERX_REQUIRE_DEBUG_ASSERTIONS=1`: see the same pair
+      # on `Tests + coverage (core)` below. This job carries the feature set —
+      # `ci,markov,nn` — that the retired `Tests (debug-assertions)` job's second
+      # command existed for, so the guards in `src/survival`, `src/markov`,
+      # `src/categorical` and `src/nn` are live exactly here (#1248, folding #344).
       - name: Generate TTE/CTMM endpoint coverage report
+        env:
+          FERX_REQUIRE_DEBUG_ASSERTIONS: "1"
         run: |
-          cargo llvm-cov --no-fail-fast --no-default-features --features ci,markov,nn --profile ci-fast \
+          cargo llvm-cov --no-fail-fast --no-default-features --features ci,markov,nn --profile ci-cov \
             --lib \
             --test agq \
             --test categorical_convergence \
@@ -146,64 +154,6 @@ jobs:
           flags: survival,markov,nn
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
-
-  # #344. Every other test job in this file builds with `--profile ci-test` or
-  # `--profile ci-fast`, and both `inherits = "release"` — so `debug-assertions`
-  # is OFF and all 177 `debug_assert!` guards in the crate compile to nothing.
-  # They are invariant checks that CI was not running.
-  #
-  # That was not theoretical. An off-by-one in `compute_max_stack`
-  # (`src/parser/model_parser.rs`) tripped its own `debug_assert!` for any
-  # expression containing an `if`, and four lib tests panicked under the dev
-  # profile while CI stayed green through the whole thing (fixed in #342; this
-  # job closes the blind spot that hid it).
-  #
-  # Deliberately a SEPARATE job rather than flipping `ci-test`: that profile
-  # exists so the population-fit tests run at release speed, and this repo is
-  # compile-bound (93% LLVM, #969/#971), so the cheap way to get the guards
-  # executed is one extra UNOPTIMISED build — not a second optimised one.
-  # Measured locally on the dev profile: 80s of tests for `--features ci` (4035
-  # tests) and 198s for `--features ci,markov,nn` (4366), all green — turning the
-  # guards on surfaced no latent failure, so this job starts green.
-  #
-  # The commands live in `tools/preflight.sh` (group `debug-assertions`) for the
-  # same reason as `Check`/`Clippy`/`Format` (#1157) — one list, so the local gate
-  # and the CI gate cannot drift. It is the one group that is opt-in locally: it
-  # runs a suite rather than a compile, so a no-argument `tools/preflight.sh`
-  # stays a pre-push habit and this job is where it has to be green.
-  #
-  # Note that "no `--profile` flag" is an invariant held by ABSENCE, and this job
-  # would be a no-op again — silently, with every test green — if a
-  # `[profile.dev] debug-assertions = false` or a
-  # `CARGO_PROFILE_DEV_DEBUG_ASSERTIONS=false` ever appeared. So the group's
-  # commands arm `debug_assertion_canary` (`src/lib_tests.rs`) via
-  # `env FERX_REQUIRE_DEBUG_ASSERTIONS=1`, and that canary fails the run if
-  # `debug_assert!` turns out to compile to nothing. Do NOT set that variable at
-  # the workflow or job level: every other job legitimately runs with the guards
-  # off, and arming it there would fail them for doing the right thing.
-  debug-assertions:
-    name: Tests (debug-assertions)
-    runs-on: ubuntu-latest
-    # Stable, not the workflow-level nightly: `--features ci,markov,nn` has no
-    # nightly-only code, and stable's rustc holds for ~6 weeks against nightly's
-    # daily bump — `rust-cache` keys on `rustc -vV`, so nightly here would cold-
-    # rebuild the dep graph every day for no gate. Same reasoning as the coverage
-    # jobs. `tools/preflight.sh` honours an inherited `RUSTUP_TOOLCHAIN`.
-    env:
-      RUSTUP_TOOLCHAIN: stable
-    steps:
-      - uses: actions/checkout@v7
-      - name: Install stable toolchain
-        run: rustup toolchain install stable --profile minimal
-
-      - uses: Swatinem/rust-cache@v2
-        with:
-          # Own key: the dev profile shares no artifacts with the `ci` (check /
-          # clippy) caches or with the instrumented coverage ones.
-          shared-key: ci-debug-assertions
-
-      - name: cargo test --lib with debug-assertions on (via tools/preflight.sh)
-        run: tools/preflight.sh debug-assertions
 
   clippy:
     name: Clippy
@@ -421,7 +371,7 @@ jobs:
         # both RUNS the base suite and is the per-PR patch gate; the full
         # `ci,slow-tests` run is nightly (the `coverage` job above) and the
         # feature-gated endpoints are the `endpoints` job. Builds with
-        # `--profile ci-fast` (no fat LTO *and* no local ThinLTO — ~31% off the
+        # `--profile ci-cov` (no fat LTO *and* no local ThinLTO — ~31% off the
         # lib compile, #969) to stay within the PR wall-clock; the fast Tier-1/2
         # fits here tolerate the ThinLTO-off fit slowdown that the nightly
         # `coverage`/`slow-tests` jobs avoid by staying on `ci-test`. `--tests`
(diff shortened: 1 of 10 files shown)


roninsightrx:
Review findings:

1. BLOCKING — per-PR production-mode test coverage disappears.

The changes around .github/workflows/ci.yml lines 128 and 411 convert both patch-gate runs from ci-fast, with debug assertions and overflow checks off, to ci-cov, with both on, while removing the separate dev-profile job. Before this PR, each pull request exercised both modes: the coverage jobs tested release-like behavior and the dedicated job tested live guards. After this PR, no pull-request job runs the core suite with debug_assertions = false or release overflow semantics; the remaining ci-test runs are scheduled, manual, or post-merge.

The rewrites make the regression concrete: src/types_tests.rs lines 56-67 and src/estimation/outer_optimizer_tests.rs lines 2193-2205 now execute only their panic arms in PR CI, while the NaN fallbacks documented as user behavior are no longer checked before merge. I verified both arms locally under ci-cov and ci-fast, but only the former is gated. The retired debug job is a subset of the new guarded coverage runs, but the old guards-off coverage run is not. Please preserve a guards-off PR test lane, full or suitably targeted, rather than trading away production-mode behavior to fix the coverage regions.

2. The panic-hook suppression races with the parallel test harness.

src/types_tests.rs lines 40-43 and src/estimation/outer_optimizer_tests.rs lines 2189-2192 call take_hook and set_hook without synchronization even though the hook is process-global. The three expected-panic paths can run concurrently in the same lib-test process. An interleaving where A takes the normal hook, B takes the silent hook installed by A, A restores normal, and B restores silent leaves the process permanently silenced. Any unrelated concurrent failure can also lose its assertion message during the window. Immediate restoration does not make this safe. Please avoid changing the hook, since three expected panic messages are preferable, or serialize all hook swaps through one shared lock with panic-safe RAII restoration.

Validation performed:
- cargo test --profile ci-cov --test preflight_owns_the_fast_gates --no-default-features --features ci — 10 passed.
- Targeted rewritten tests under both ci-cov and ci-fast — passed.

roninsightrx:
Both findings accepted and fixed in 73d00e30. Thanks — finding 1 in particular was a real hole, not a style point.

## 1. BLOCKING — per-PR release-semantics coverage. Agreed, and the diagnosis is exactly right.

I traded the guards-off lane away without noticing I was trading anything: I checked that the *retired* job was a subset of the new guarded runs, and never asked the converse question about the old guards-off coverage runs. Your framing — "the retired debug job is a subset of the new guarded coverage runs, but the old guards-off coverage run is not" — is the sentence I should have written for myself.

It is also self-inflicted in the sharpest possible way, since the tests it disarms are the three this PR rewrote, and their NaN arms are the documented *user* behaviour.

**Fix: `Tests (release semantics)`** — `--lib` on the `ci,markov,nn` union under `ci-fast`, delegating to a new `tools/preflight.sh release-semantics` group. Runs in parallel with the coverage jobs and is shorter than either, so it should add no wall-clock.

I took the "suitably targeted" option rather than "full", and want to be explicit about what that does not restore: **guards-off `tests/` binaries**. That is the 120-odd binary compile the lane is shaped to avoid; `slow-tests.yml` and the nightly coverage job still run them on `ci-test` with the guards off. My reasoning is that the only `#[cfg(debug_assertions)]` production code is the MR ODE-twin verifier in `pk/modified_release.rs` (a development aid that does not run in a user's build either way), so the integration tier's guards-off exposure is thin — but that is a judgement call resting on an enumeration, and enumerations rot. Say the word and I will widen it to `--workspace --tests`.

**The lane is pinned in both directions**, because it is worth having only while it is the *other* mode. If `ci-fast` ever acquired `debug-assertions` it would silently become a second copy of the coverage jobs, every test green — #344's failure shape with the sign flipped. So:

- new inverse canary `FERX_REQUIRE_NO_DEBUG_ASSERTIONS=1` (`src/lib_tests.rs`), which fails the run if `debug_assert!` turns out to be live;
- `preflight_owns_the_fast_gates.rs` now asserts the profile statically in both directions, resolving `inherits` through Cargo.toml.

One number that makes the point better than the argument does: **4577 tests under `ci-cov` against 4571 under `ci-fast`**. The six-test delta is the `#[cfg(debug_assertions)]` verifier's own tests — the two lanes genuinely do not compile the same crate.

## 2. Panic-hook race. Agreed, and worse than I had written down.

My comment conceded the "concurrent failure loses its message" window and called it an accepted cost. It missed your actual failure: A takes the normal hook, B takes the *silent* hook A installed, A restores normal, B restores silent — process permanently silenced, and no amount of prompt restoration fixes it, because the corruption is in what B captured.

Removed the swap at both sites rather than serialising it. Three expected panic messages in the log are cheaper than a shared lock plus RAII, and that is now stated at the site instead of being implied.

## Also in this push

- `docs/development/sdlc.qmd` §8.2 split under three subheadings — the additions took the terminal section to 5652 chars, over docs-lint's R1 5000 limit. Split rather than baselined, per CLAUDE.md's ratchet rule.
- **A correction to the PR body**, unrelated to your review: it claimed the critical path drops to `~18m`. That was extrapolated from local hardware before any CI run of the new profile existed. Measured on two runs of the same commit it is **~28m** (cold: core 21m / endpoints 28m; warm: 28m / 16m — they swap, which is the ~5 min variance this repo documents in #969). Still down from 32–42m, but not what I wrote.

## Mutations for the new assertions

Each reddens its own side:

| mutation | result |
|---|---|
| release lane drifted to `--profile ci-cov` | fails |
| inverse canary arming deleted | fails |
| inverse canary run on `ci-cov` / on `ci-fast` | **fails** / passes |

## Verification

- `tools/preflight.sh debug-assertions release-semantics` → 4238 + 4577 + 4571 passed, 0 failed, both canaries armed
- `preflight_owns_the_fast_gates` → 10 passed

roninsightrx:
Measured on the review-fix run ([34263420357](https://github.com/FeRx-NLME/ferx-core/actions/runs/34263420357)) — all green, and the new lane does not extend the critical path:

| job | |
|---|---|
| `Tests + coverage (core)` | 28m |
| `Tests + coverage (TTE/CTMM endpoints)` | 17m |
| **`Tests (release semantics)`** | **6m** |

6 min, and that is its **cold** number — `shared-key: ci-release-semantics` was created by this run, so it rebuilt the whole dep graph. Warm it will be less. Against a 28 min critical path it is free in wall-clock; it costs runner-minutes only.

Which puts the whole change at **32–42m → ~28m** critical path, with both build modes exercised per PR instead of one.

Issue #1295, opened 2026-09-08
bug(sens): the non-finite-timeline guards fill a NaN value with ZERO jets, so a diverged subject hands FOCEI a gradient that claims to be fine


TeunP:
The two analytic-sensitivity walks NaN-fill a subject whose timeline could not be ordered
(`ode::predictions::timeline_has_non_finite`, #1189), and both comments said the fill delivers
"NaN **jets**". It does not. `T::from_f64(f64::NAN)` lifts a **constant**, so on a dual the
value is `NaN` while the gradient is `[0.0; N]` — and `PkNum::jets_finite()` returns `true`.

#1234 corrected the comments and pinned the measured behaviour in tests. This issue is the
other half: making the fill deliver what the guard means, which is a **behaviour change** and
was deliberately kept out of that diagnostics PR.

### Measured at `cf32e1fd`

| | value | jets | `jets_finite()` |
|---|---|---|---|
| `Dual1<2>::from_f64(NAN)` | `NaN` | `[0.0, 0.0]` | **`true`** |
| `Dual1<2>::from_f64(NAN).exp()` | `NaN` | `[NaN, NaN]` | `false` |

Through the provider, on a subject with a `NaN` dose time:

| readout | `f` | `df_deta` |
|---|---|---|
| `ode(obs_cmt=central)`, no `[scaling]` | `NaN` | **`[0.0]`** |
| `[scaling] y = central / V`, η on `V` | `NaN` | `[NaN, NaN]` |

The scaled readout is why this was never noticed: `d(central/V)` folds the fill's `NaN` value
into the jets *downstream* of the fill, so every scaled fixture agrees with the old comment
whatever the fill writes. `nan_lagtime_sens_walk_is_non_finite_not_a_panic`
(`src/sens/provider_tests.rs`) uses the `/V` model **and** asserts only `o.f`, so it cannot see
this either way.

### Why it was split out of #1234

Because it moves the fit. Measured one variable at a time, on a 2-subject FOCEI fit
(`two_state_model(1.0)`, `outer_maxiter = 1`) where one subject carries a `NaN` dose time —
the only change being `T::from_f64(f64::NAN)` → `T::from_f64(f64::NAN).exp()` at
`sens/ode_provider.rs`'s static fill:

| | before | after |
|---|---|---|
| `ofv` | `NaN` | **`2e20`** |
| `n_iterations` | 4 | **12** |
| `converged` | `true` | **`false`** |
[...]

Issue #573, opened 2026-06-28
Allow TIME not starting at 0 for subjects.


roninsightrx:
Just create a relative time variable that is used as independent variable substitute.

PR #1296, opened 2026-09-08
fix(ode): count and report a walk abandoned on an unorderable timeline [closes #1234]

### LINKED ISSUE #1234: refactor(ode/sens): make the non-finite-timeline outcome consistent — OdeSolverStats and integrate_g's Some(NaN) vs None
Two loose edges left by #1186/#1189's non-finite-timeline guards (PR #1231). Neither is a wrong number today; both are places where the *typed* non-finite outcome that PR introduced is not yet consistent, and each would become a wrong answer under a plausible future change. Raised by the xhigh review of #1231 and deliberately not fixed there, to keep a correctness PR from carrying semantic changes reviewed late.

## 1. The objective builder's early return leaves `OdeSolverStats` untouched

`ode_predictions_with_extra_breaks_and_stats` (`src/ode/predictions.rs`) takes `stats: Option<&mut OdeSolverStats>` and accumulates per-segment solver statistics as it walks. #1231 added an early return for a non-finite timeline that fires *before* any segment runs:

```rust
if timeline_has_non_finite(&break_times) {
    return (predictions, chz_states);
}
```

so `stats` comes back at its incoming value. A caller reading steps/rejections per subject then cannot distinguish **"this subject's timeline could not be ordered"** from **"this subject had nothing to integrate"** — the exact conflation the typed non-finite outcome was introduced to remove everywhere else.

Options: mark the stats record as failed/diverged, or count the subject as one attempted-and-abandoned solve. Whichever, the point is that the two states should not look identical to a diagnostic.

## 2. `integrate_g` returns `Some(NaN jets)` where its sibling uncoverable path returns `None`

`src/sens/ode_provider.rs::integrate_g` ends with a deliberate decline:

```rust
// … would keep its zero-initialised state and feed a silent `f = 0`, `∂f = 0` into
// the gradient. Decline so the caller falls back to FD for this subject rather than
// return a wrong `Some`.
if recorded.iter().any(|&r| !r) {
    return None;
}
```

#1231's non-finite guard sits earlier in the same function and returns `Some(states)` with every jet `NaN`. So the function now has **two opposite conventions for a subject the analytic walk cannot cover**: `None` (decline → FD) and `Some(NaN)`.

Both reach a NaN gradient today, because the FD fallback differentiates `ode_predictions`, which #1231 also made return NaN for the same subject — so this is currently a tie, not a defect. It stops being a tie the moment a caller treats `Some` as "the analytic walk succeeded", e.g. to skip an FD cross-check, to count analytic coverage, or to choose a gradient path. Then a diverged subject is silently classified as analytically solved.

Pick one convention and document it on the function. `None` is arguably right (it *is* an uncoverable subject, and the caller already has the fallback); `Some(NaN)` avoids ~2·n_theta wasted solves that land on NaN anyway. Either is fine — having both is not.

## Scope

Small, and separable from each other. Neither blocks anything. Filed so the inconsistency is not discovered later as a wrong number rather than as a note.

TeunP:
## Why

An `OdeSolverStats` block full of zeros meant three different things and the fit could not tell
them apart. Measured at `cf32e1fd`, same model, inside the `SolverStatsScope` production
actually collects through (`api/fit.rs`):

| subject | `attempted / accepted / rejected` | predictions |
|---|---|---|
| normal | `7 / 7 / 0` | `[95.12, 81.87, 44.93, 9.07]` |
| **non-finite timeline** | **`0 / 0 / 0`** | `[NaN, NaN, NaN, NaN]` |
| no records at all | `0 / 0 / 0` | `[]` |
| one observation at `t = 0` | `0 / 0 / 0` | `[0.0]` |

Every counter on `OdeSolverStats` reports work that *happened*, and a walk abandoned before the
solver is called does none — so the one case whose predictions are `NaN` by construction reads
exactly like the two ordinary ones.

That path is still reachable after #1286, which checks dose *attributes* (`ALAG`/`F`/`D`/`R`),
not record times: a subject with a `NaN` dose **time** gets 0 diagnostics from
`check_model_data`, and `fit()` then returns `ofv = NaN` with five warnings, **none** of which
names the subject, the timeline, or `NaN`. The `ode_solver` entry was `None`.

Second half of the issue: both analytic-sensitivity guards claimed to return "NaN **jets**".
They do not — `T::from_f64(NAN)` lifts a *constant*, so the value is `NaN` and the gradient is
`[0.0; N]`, with `PkNum::jets_finite()` returning `true`.

## What changed

**The counter (N2a-1).** New `OdeSolverStats::abandoned_non_finite_timeline`, bumped by a new
`pub(crate) record_abandoned_non_finite_timeline()` that writes the thread-local sink
*directly* — the ordinary route in is the tee inside `integrate_resolved_g`, and this is
precisely the event where no driver is ever reached. Wired at all **eight** `f64`-side guard
sites: `ode/predictions.rs` 3396, 4439, 5418, 6003, 6541, 7109, 7353 and `ode/ekf.rs` 265.
`merge`'s exhaustive destructure forced the coupling, as intended — no `..` added.

Deliberately **not** recorded at the two `sens/ode_provider.rs` sites (4962, 6693): `fit_inner`
copies exactly one field out of the sensitivity sweep's scope (`auto_stiff_rejected_jets`) and
discards the rest, so anything else deposited there is either dropped or — if that copy were
widened — would fire a warning clause about *predictions* on the strength of a gradient solve.

**The warning (N2a-2).** `abandoned` joins `unclean` in `ode_solver_diagnostics_warning` and
gets its own clause, placed **first** because it is the only clause here reporting predictions
that are `NaN` rather than merely inaccurate. New `abandoned_non_finite_timeline` key in the
`details` payload.

**The two sens exits (N2b-i).** Both comments now say what the fill actually delivers, with the
measured numbers, and the **discriminator** between `integrate_g`'s two failure exits is
documented at all three sites:

- non-finite `Some` — *the walk ran and the subject is diverged*; FD would reproduce the same
  `NaN` one solve per perturbation more expensively, and would report it under an "outside the
  analytic provider's scope … correct but slower" warning that is wrong twice over.
- `None` (the `recorded` check) — *the walk declines to answer*; an observation was never
  captured, so FD really can do better.

`None` is not available as a common spelling anyway: `integrate_tvcov_g`, the twin an estimated
lagtime routes to, returns a bare `Vec<Vec<T>>` with **no `None` channel at all** and NaN-fills
the identical condition. `Some(non-finite)` is already the cross-walk convention; collapsing
the static exit to `None` would *create* the asymmetry rather than remove one.

**Docs.** `docs/warnings.qmd`'s `## Warning codes` was baselined at R1 7391 and shrink-only, so
growing the `ode_solver` row was impossible without splitting it. Split into eight `###`
subsections with `{#warning-codes}` kept on the `##` (three inbound R4 links: `warnings.qmd:12`,
`output.qmd:134`, `model-file/ode-models.qmd:649` — all still resolve), and its baseline line
**deleted**, since a stale entry is itself a failure. `docs-lint --check` is green with **five**
baselined sections. `## Details payloads` had 2,160 chars of headroom, so the new key fits with
no split. `model-file/ode-models.qmd` gained the case in its "where the step counts show up"
section.

## Alternatives considered

**Fixing `ode_predictions_with_solver_stats`'s `stats` parameter** — the surface the issue names.
It has **no production caller** (only `sens/ode_provider_tests.rs:8589` and
`tests/transit_nonmem_anchor.rs`); `ode/solver.rs` already says so in prose. Fixing it would
satisfy the issue's literal wording and reach no user. Production collects through the
thread-local scope, so that is where the counter had to go. Deleting the dead surface is a
separate change.

**Returning `None` from the non-finite guard** (the issue's item-2 proposal) — rejected on the
measurement above: it would create the cross-walk asymmetry, not remove it.

## Cross-repo dependency

`OdeSolverStats` is referenced nowhere in `ferx-r` on `origin/main`, and nowhere in `crates/`.
The field is additive, so this is a baseline regeneration only — no ferx-r PR, no lock bump.

## Breaking changes
- [x] Public Rust API — `api/ferx-core-public-api.txt` regenerated with
      `tools/update-public-api.sh`. One additive item, listed twice
      (`ferx_core::ode::` and `ferx_core::ode::solver::`):
      `OdeSolverStats::abandoned_non_finite_timeline`. It is public because every other field
      of that struct is, and because it is the counter a consumer reading the `ode_solver`
      `details` payload needs in order to tell an abandoned walk from a clean one — the same
      role `auto_stiff_rejected_jets` plays for #1204.
- [ ] None

## Numerical validation

- [x] Not applicable — but the twin was measured, because the split depends on it.

No NONMEM anchor: NONMEM has no spelling for "which of my own diagnostics fired", and the three
outcomes above are exact hand-computed values (a walk that never ran has 0 steps), not
tolerances.

**The twin that proves nothing else moved.** `fit()` on the `NaN`-dose-time population before
and after this PR:

```
# before (cf32e1fd)                # after (this PR)
ofv          = NaN                 ofv          = NaN
n_iterations = 4                   n_iterations = 4
converged    = true                converged    = true
warnings     = 5                   warnings     = 6   (+ W_ODE_SOLVER_DIAGNOSTICS)
eta (healthy subject)              eta (healthy subject)
             = 1.000846558216595                = 1.000846558216595
```

The only difference is the warning this PR exists to add.

**Why the jet repair was split out.** Making the fill deliver real NaN jets — the one-line
change `T::from_f64(NAN)` → `T::from_f64(NAN).exp()` — moves the fit, measured on the same
fixture with nothing else changed:

| | before | after |
|---|---|---|
| `ofv` | `NaN` | **`2e20`** |
| `n_iterations` | 4 | **12** |
| `converged` | `true` | **`false`** |
| warnings | 6 | **7** (adds "Outer optimization did not converge") |
| η̂ of the **healthy** subject | `1.000846558216595` | **`0.8822190053957163`** |

`2e20` is `2 · 1e20`, consistent with the outer objective's non-finite sentinel. A healthy
subject's η̂ moving is not a diagnostics change, so per the plan's own exit condition it is
split out as **#1295** rather than folded in here. This PR corrects the claim; that
one changes the behaviour, with its own validation.

## Performance
- [x] No significant impact expected — one `Cell` read per fired guard, inside a branch that
      only runs on a subject whose predictions are already `NaN`.

## Tests
- [x] Unit tests added in the same module (`cargo test --lib` passes: 4430 passed, 0 failed)
- [ ] No new tests needed

Eight Tier-1 tests, each mutation-tested with `--no-fail-fast`. **Which engine each fixture ran
on is asserted, not inferred** — T5/T7 call `ode_tvcov_supported` directly so a fixture that
silently started routing to the other walk fails rather than testing the wrong guard.

| | test | file | engine |
|---|---|---|---|
| T1 | `the_static_walk_separates_an_abandoned_timeline_from_nothing_to_integrate` | `api/tests/ode_solver_diagnostics_tests.rs` | `f64`, dense static |
| T2 | `the_event_driven_walk_separates_an_abandoned_timeline_from_nothing_to_integrate` | same | `f64`, event-driven |
| T3 | `an_abandoned_walk_leaves_every_started_segment_counter_at_zero` | same | `f64`, dense static |
| T4 | `a_fit_over_an_unorderable_timeline_says_so` | same | `f64`, real `fit()` |
| T5 | `the_static_walk_returns_a_non_finite_value_with_finite_zero_jets` | `sens/ode_provider_tests.rs` | `Dual1`/`Dual2`, `integrate_g` |
| T6 | `the_scaled_readout_is_what_masked_the_zero_jets` | same | `Dual1`/`Dual2`, `integrate_g` |
| T7 | `the_event_driven_walk_returns_a_non_finite_value_with_finite_zero_jets` | same | `Dual1`/`Dual2`, `integrate_tvcov_g` |
| T8 | `a_diverged_subject_answers_some_rather_than_declining_to_fd` | same | `Dual1`/`Dual2`, `integrate_g` + `fd_fallback_warning`'s predicate |

| mutation | T1 | T2 | T3 | T4 | T5 | T6 | T7 | T8 |
|---|---|---|---|---|---|---|---|---|
| delete the record at `predictions.rs:3396` (static) | **RED** | green | RED | RED | — | — | — | — |
| delete the record at `predictions.rs:6003` (event-driven) | green | **RED** | green | green | — | — | — | — |
| delete the record at `predictions.rs:6541` (`with_states`) | green | green | green | **RED** | — | — | — | — |
| counter also bumps `unfinished_segments` | green | green | **RED** | RED | — | — | — | — |
| drop `\|\| abandoned > 0` from `unclean` | green | green | green | **RED** | — | — | — | — |
| delete `integrate_g`'s guard | — | — | — | — | **RED** | RED | green | RED |
| delete `integrate_tvcov_g`'s guard | — | — | — | — | green | green | **RED** | green |
| apply the follow-up's jet repair to `integrate_g` | — | — | — | — | **RED** | **green** | green | green |
| swap `integrate_g`'s `Some(states)` for `None` | — | — | — | — | RED | RED | green | **RED** |

Three things that matrix is there to show:

1. **T1/T2 and T5/T7 are separable in both directions**, and each failure message names its own
   side ("static walk: …" / "event-driven walk: …"). A single fixture reaching only one engine
   cannot see the other's guard.
2. **T4 pins the count at `== 3`, not `> 0`** — and that is load-bearing. Measured inside the
   fit's own scope, one bad subject is abandoned three times in the post-fit sweep
   (`ode_predictions` ×2, `ode_predictions_with_states` ×1). At `> 0` the third row of the
   matrix was **green**: either recorder alone kept the assertion satisfied, the
   redundant-gate hole. Pinning the count also measures the "counts walks, not subjects" claim
   in the warning text.
3. **T6 provably cannot distinguish the two fills** — row 8 shows it green under the exact
   repair that reddens T5. That is its whole job: it explains why
   `nan_lagtime_sens_walk_is_non_finite_not_a_panic` (`sens/provider_tests.rs:11612`), which
   uses a `/V` model **and** asserts only `o.f`, is not a regression guard for any of this.

`is_finite()` is deliberately *not* the assertion in T5/T7: the value is `NaN` on both arms
already, and the jets are a finite `0.0`. The jets are what is asserted.

Full suite: `cargo test --lib --features ci,survival --no-fail-fast` → 4430 passed / 0 failed;
`cargo test --features ci,survival --no-fail-fast` (all integration binaries) green;
`tools/preflight.sh` green.

- [x] Not applicable (a diagnostic; the user-visible surface is the existing `ode_solver`
      warning, whose new clause and payload key are documented in `docs/warnings.qmd`)

## Docs
- [x] `docs/` updated — `warnings.qmd` (`ode_solver` row, `details` key, the `###` split),
      `model-file/ode-models.qmd`
- [x] No new pages, so `_quarto.yml` unchanged; `docs/_site/` not committed
- [x] `CHANGELOG.md` `[Unreleased]` → `### Fixed`
- [ ] No user-visible change

## Checklist
      (`--all-targets`), rustdoc, docs, public-API baseline
- [x] `Dual2` path — the sens changes are comments plus one `pub(crate)` recorder that the sens
      walks deliberately do **not** call; no `*_g<T: PkNum>` body changed, so differentiability
      is untouched. T5–T8 exercise both `Dual1` and `Dual2` instantiations through
      `ode_subject_sensitivities`.

### ferx-site / ferx-book

### Example execution
- [x] No example execution step needed — the change is a diagnostic counter and a warning
      clause; no example's numbers move (see the twin above).

## Reviewer hints

- **Start with the mutation matrix.** Rows 3 and 8 are the two that caught real holes: row 3
  turned T4 from `> 0` into `== 3`, and row 8 is what earns T6 its place instead of reading as
  a duplicate of T5.
- **The counter counts walks, not subjects,** and that is stated in the field doc, the warning
  text, and `docs/warnings.qmd`. Three per subject on a plain FOCEI post-fit sweep, measured.
  If you would rather it were per-subject, say so — it would mean deduplicating across engines
  inside the scope, which the scope has no subject identity to do with.
- **The two sens sites are comment-only.** The fills are byte-identical to `cf32e1fd`; what
  changed is that they now describe what they do. The behaviour change lives in
  **#1295** with the measurement above.
- **One observation, not filed:** `integrate_g`'s `None` exit (the `recorded` check) appears to
  be **unreachable from `ode_subject_sensitivities` today**. Nine shapes aimed at it — an
  observation below the timeline floor, an all-negative grid, observations `5e-13` after the
  integration start and after a dose, duplicate and `1e-16`-apart times, a single observation at
  `t = 0`, observations entirely before the first dose, no doses at all — all return `Some`.
  `subject_integration_start` (#573) is why: a negative observation time now *becomes* the first
  break rather than falling below it, so the exit's own doc-comment example no longer reaches
  it. A defensive backstop being unreachable is fine, so this is recorded in T8's doc comment
  rather than filed — but it is why T8 pins the reachable half plus the caller contract
  (`fd_fallback_warning` keys on exactly that `is_none()`) instead of both exits behaviourally.
- **Skimmable:** the `docs/warnings.qmd` split is a mechanical partition of one table into eight
  — no row text changed except `ode_solver`'s, and no row moved between groups.

## Open questions

None blocking. The one judgement call is walks-vs-subjects on the counter (see hints).

## Round 2 — after review (`8d13b805`)

Everything above describes this PR as first submitted (`335cb9e5`). The review found ten
findings; the full disposition table, the mutation matrix and the measurements are in
[the round-2 reply](https://github.com/FeRx-NLME/ferx-core/pull/1296#issuecomment-5596677193).
What follows is the summary, so this description is not read as the final state.

| commit | findings |
|---|---|
| `ece0cfb2` | §1 (blocking), §3, §5, §7 |
| `a423383a` | §2 |
| `8d13b805` | §4, §6, §8, §9 |

- **§1 was real, and was reproduced red before it was fixed.** The counter reached the
  thread-local scope but **not** the `stats` out-parameter, so
  `ode_predictions_with_solver_stats` — the one public getter, and the surface #1234's item 1
  actually names — still answered `0` on a `NaN`-prediction walk, which also left
  `OdeSolverStats::merge`'s new line unreachable.
  `record_abandoned_non_finite_timeline(Option<&mut OdeSolverStats>)` now feeds both channels
  through `record_to_stats_sink` + `merge`; that is §7's fix as well.
- **The guard/recorder pairing is structural now (§5).**
  `ode::predictions::abandon_non_finite_timeline` detects *and* records in one call, and all
  eight engine guards route through it.
  `the_bare_timeline_predicates_are_not_called_outside_this_guard` scans `src/**/*.rs` and pins
  the bare predicates at **exact** counts — `src/ode/predictions.rs` ×2 (inside the two helpers)
  and `src/sens/ode_provider.rs` ×2 (the deliberate non-recorders) — so a ninth guard cannot be
  written without the counter, and the allowance cannot quietly widen *or* vanish.
- **The warning's tail no longer contradicts its own new clause (§2).** `unclean` splits into
  `unclean_integration` and the abandoned count; the solver-knob advice is scoped to the former,
  an abandoned walk gets advice that agrees with `ode-models.qmd`, and the lead-in verb branches
  too. The replacement wording was measured out of a real `fit()`, not read off the source.
- **Six more tests; the matrix is now 12 mutations, 12 killed, no survivors,** per-site kill sets
  disjoint. The round-1 table above (T1–T8) is superseded by the reply's. Three recorder sites —
  `ode_dense_solve_states`, `ode_solve_until_chz_threshold`, `ekf.rs` — had **no** test at all
  before. M1, dropping the out-param, is killed by the new public-getter test **alone**, with
  T1–T4 staying green, which is exactly why §1 was invisible.
- **Two of the review's own premises were wrong, corrected by measurement.**
  `ode_dense_solve_states` has **five** production callers, not a `markov`-only one
  (`markov::endpoint:341`, `survival/mod.rs:728`, `api/output_columns.rs:640`,
  `ode_predictions_event_driven_with_states`, `adaptive_window_signal_aucs`); T4's `== 3` is
  lane-independent for a different and stronger reason — the scope wraps
  `compute_subject_results` only (`api/fit.rs:1912`), and the `output_columns` caller runs after
  it closes. And `ode_solve_until_chz_threshold` runs on `solver::solve_ode_until_threshold`, its
  own driver, which never tees into `STATS_SINK` — measured `attempted_steps = 0` on a solve that
  crossed the threshold normally — so its control asserts the *outcome*, not the counters.

§10 (`OdeSolverStats` is not `#[non_exhaustive]`) is **not** fixed here: adding the attribute is
itself a breaking change and wants its own baseline diff. Filed as **#1302**, alongside **#1303**
(`fit()` returns `converged: true` with `ofv = NaN`) and **#1304** (`predict()` / `simulate()`
open no `SolverStatsScope`, which is §4 and §9's underlying scope question).

Gates at `8d13b805`: `tools/preflight.sh` green, all six groups — the public API still matches the
committed baseline, no new `pub` items this round. `--workspace --tests --no-default-features
--features ferx-core/ci` = **203 binaries / 5877 passed / 0 failed**; `--tests
--no-default-features --features ci,markov,nn` = **181 / 5568 / 0**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -215,6 +215,10 @@ section of the SDLC for the versioning policy).
   `ofv = NaN` with no warning mentioning the subject, the timeline or `NaN`. The counter reports
   *walks*, not subjects (a subject whose predictions and `[odes]` state readout are both
   requested contributes more than one), and it rides in the warning's `details` payload.
+  `ode_predictions_with_solver_stats` reports it too. The `ode_solver` message no longer ends
+  by recommending a different `ode_method` or looser tolerances when the only thing that went
+  wrong is an abandoned walk: nothing was integrated, so no solver setting changes the
+  outcome, and the message now says that instead.
 - Deeply saturated, over-capacity steady-state input-rate models no longer let
   Anderson acceleration report a huge spurious periodic state when integration
   error hides the positive per-cycle surplus (#867, PR #955).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -203,6 +203,18 @@ section of the SDLC for the versioning policy).
   added parameter; `examples/two_cpt_oral_covmodel.ferx` goes from OFV -1026.35 at its
   initial estimates to -1195.30, and `examples/two_cpt_oral_cov.ferx` from -1168.48 to
   -1199.33 (NONMEM FOCEI: -1199.43).
+- A subject whose timeline cannot be ordered — a `NaN` or infinite dose time, lagtime, route
+  lag or infusion duration — is now **reported** instead of being silently indistinguishable
+  from a subject with nothing to integrate (#1234). The prediction engines abandon such a walk
+  before calling the solver, which left every counter in the `ode_solver` diagnostic at zero:
+  measured on one model, a non-finite timeline, a subject with no records, and a subject with a
+  single observation at `t = 0` all read `attempted/accepted/rejected = 0/0/0` while returning
+  `[NaN, …]`, `[]` and `[0.0]` respectively. A new `abandoned_non_finite_timeline` counter on
+  `OdeSolverStats` separates the first from the other two, and a fit that hits it now emits an
+  `ode_solver` **Warning** naming the count and what to check, where before it returned
+  `ofv = NaN` with no warning mentioning the subject, the timeline or `NaN`. The counter reports
+  *walks*, not subjects (a subject whose predictions and `[odes]` state readout are both
+  requested contributes more than one), and it rides in the warning's `details` payload.
 - Deeply saturated, over-capacity steady-state input-rate models no longer let
   Anderson acceleration report a huge spurious periodic state when integration
   error hides the positive per-cycle surplus (#867, PR #955).
--- a/api/ferx-core-public-api.txt
+++ b/api/ferx-core-public-api.txt
@@ -1135,6 +1135,7 @@ pub ferx_core::ode::solver::OdeSolverOptions::stiff_abort_after: core::option::O
 impl core::default::Default for ferx_core::ode::OdeSolverOptions
 pub fn ferx_core::ode::OdeSolverOptions::default() -> Self
 pub struct ferx_core::ode::solver::OdeSolverStats
+pub ferx_core::ode::solver::OdeSolverStats::abandoned_non_finite_timeline: usize
 pub ferx_core::ode::solver::OdeSolverStats::accepted_steps: usize
 pub ferx_core::ode::solver::OdeSolverStats::attempted_steps: usize
 pub ferx_core::ode::solver::OdeSolverStats::auto_fallback_failed: usize
@@ -1203,6 +1204,7 @@ pub ferx_core::ode::OdeSolverOptions::stiff_abort_after: core::option::Option<u3
 impl core::default::Default for ferx_core::ode::OdeSolverOptions
 pub fn ferx_core::ode::OdeSolverOptions::default() -> Self
 pub struct ferx_core::ode::OdeSolverStats
+pub ferx_core::ode::OdeSolverStats::abandoned_non_finite_timeline: usize
 pub ferx_core::ode::OdeSolverStats::accepted_steps: usize
 pub ferx_core::ode::OdeSolverStats::attempted_steps: usize
 pub ferx_core::ode::OdeSolverStats::auto_fallback_failed: usize
--- a/docs/.lint-baseline
+++ b/docs/.lint-baseline
@@ -12,4 +12,3 @@ R1 5239 docs/estimation/tte.qmd::competing-risks-cause-specific-hazards
 R1 7642 docs/model-file/adaptive-dosing.qmd::current-scope-and-limits
 R1 6006 docs/model-file/error-model.qmd::covariate-selected-error-models-ifelse
 R1 5746 docs/model-file/scaling.qmd::interaction-with-gradients
-R1 7391 docs/warnings.qmd::warning-codes
--- a/docs/model-file/ode-models.qmd
+++ b/docs/model-file/ode-models.qmd
@@ -650,8 +650,9 @@ warning (see [Warnings](../warnings.qmd#warning-codes)). It fires when a step cl
 minimum
 step size, when a returned segment stopped before its requested end, when an `auto` escalation
 had to be discarded and re-solved explicitly, when that explicit fallback also failed, when an
-escalation was discarded because its derivatives overflowed, or when `ode_stiff_abort_after`
-cut a segment short — and, at `Info` severity, when `auto` escalated
+escalation was discarded because its derivatives overflowed, when `ode_stiff_abort_after`
+cut a segment short, or when a walk was abandoned *before integrating* because the subject's
+timeline could not be ordered — and, at `Info` severity, when `auto` escalated
 and everything worked, so an escalation is never invisible. Either way it says how many
 segments changed stepper part-way through, because that changes how the step counts above it
 read: the steps before a switch were taken by a different method than the ones after it. Clamps taken inside an escalation
@@ -661,6 +662,16 @@ most actionable line ferx prints about the solver: the probe was right that the 
 stiff and wrong about which stiff method could integrate it, which is the case for naming
 `ode_method = rodas5p` (or `rosenbrock23`) by hand. The counters ride along in the warning's
 `details` payload.
+
+The abandoned-walk case is the odd one out, and has its own counter
+(`abandoned_non_finite_timeline`) for that reason. A `NaN` or infinite dose time, lagtime,
+route lag or infusion duration makes the timeline unsortable, so ferx returns `NaN`
+predictions for that subject rather than silently dropping the offending dose and reporting
+the remaining, drug-free trajectory as valid. Nothing is integrated, so such a subject
+contributes nothing to any other counter in the payload — which, before this counter existed,
+read exactly like a subject there was nothing to integrate for. It is not an `ode_method`
+problem and no solver setting fixes it: check the dose records, and any exponential covariate
+model on `ALAG` / `F` / `D` / `R` that can overflow at typical covariate values.
 
 #### When the derivatives overflow but the predictions do not
 
--- a/docs/model-file/ode-models.qmd
+++ b/docs/model-file/ode-models.qmd
@@ -673,6 +673,11 @@ read exactly like a subject there was nothing to integrate for. It is not an `od
 problem and no solver setting fixes it: check the dose records, and any exponential covariate
 model on `ALAG` / `F` / `D` / `R` that can overflow at typical covariate values.
 
+The counter is reported by `fit()`. `predict()`, `simulate()` and `predict_survival()` still
+return `NaN` for such a subject and say nothing about it — the solver-statistics scope those
+counters accumulate into is opened by the fit only. The `NaN` itself is the signal on those
+entry points, as it was before.
+
 #### When the derivatives overflow but the predictions do not
 
 One rejection reason is not about the method at all. A dual number's derivative jets carry
--- a/docs/warnings.qmd
+++ b/docs/warnings.qmd
@@ -39,7 +39,11 @@ when empty.
 
 ## Warning codes {#warning-codes}
 
-Each `WarningCode` token, what it flags, and a typical response.
+Each `WarningCode` token, what it flags, and a typical response. The groups below
+are for reading convenience only — the token is what a consumer branches on, and
+a code never moves between groups in a way that changes its `category` string.
+
+### Convergence and covariance {#codes-convergence}
 
 | Code (`category`) | Severity | Flags | Typical agent response |
 |-------------------|----------|-------|------------------------|
@@ -50,20 +54,45 @@ Each `WarningCode` token, what it flags, and a typical response.
 | `condition_number` | Critical | Ill-conditioned covariance / high condition number. | Over-parameterized — drop a random effect or covariate. |
 | `optimizer_health` | Warning | Trust-radius collapse / degeneracy. | Try a different optimizer or re-scale parameters. |
 | `vi_bad_basin` | Critical | VI's final ELBO tightness check found that a flat objective is an unusable variational bound, not convergence. | Reject the fit; use better starting values or initialize VI from a fitted point with `method = [focei, vi]`. |
+
+### Residual and random-effect diagnostics {#codes-residuals}
+
+| Code (`category`) | Severity | Flags | Typical agent response |
+|-------------------|----------|-------|------------------------|
 | `dw_autocorrelation` | Warning | IWRES autocorrelation (Durbin–Watson out of range). | Revisit the residual-error / structural model. |
 | `eta_normality` | Warning | ETA departs from normality. | Consider a mixture or transform. |
+
+### Censoring, resampling, and experimental features {#codes-methods}
+
+| Code (`category`) | Severity | Flags | Typical agent response |
+|-------------------|----------|-------|------------------------|
 | `experimental` | Warning | An experimental feature (SDE, neural-network) was used. | Treat results with caution; validate. |
 | `bloq_method` | Warning | BLOQ / M3 censoring caveat. | Confirm the censoring semantics match intent. |
 | `sir` | Warning | SIR failed or was requested without a covariance. | Ensure a covariance exists before requesting SIR. |
 | `importance_sampling` | Warning | ESS = 0 / proposal collapse. | Increase samples or improve the proposal. |
+
+### Shrinkage and parameter identifiability {#codes-identifiability}
+
+| Code (`category`) | Severity | Flags | Typical agent response |
+|-------------------|----------|-------|------------------------|
 | `eps_shrinkage` | Warning | Residual (EPS) shrinkage high / negative. | Sparse data for the error model — simplify it. |
 | `eta_shrinkage` | Warning | One or more ETA shrinkages exceed ~30%. | Data poorly inform that IIV — consider removing it on the affected parameter(s). |
 | `boundary_estimate` | Warning | One or more THETA estimates are pinned to an optimizer bound. | Non-identifiability or a too-tight bound — inspect; relax the bound or simplify. |
 | `parameter_at_runaway_guard` | Critical / Warning | A free coordinate is pinned to a hidden optimizer guard (implicit THETA cap, OMEGA / SIGMA rail), with a `verdict` the side does not decide. | `collapse` (`Warning`): remove or simplify it. `runaway` (`Critical`): `converged: false` — reject; revisit model, data, inits; for SIGMA, rescale DV. |
 | `inflated_rse` | Warning | One or more THETA estimates have RSE > ~50%. | Imprecisely estimated — often over-parameterization; consider simplifying. |
 | `high_correlation` | Warning | One or more THETA (fixed-effect) pairs have \|correlation\| ≥ 0.95. | Over-parameterization / non-identifiability — fix or remove one of each pair. |
+
+### Dataset and model structure {#codes-structure}
+
+| Code (`category`) | Severity | Flags | Typical agent response |
+|-------------------|----------|-------|------------------------|
 | `data_quality` | Warning | Dataset issue (missing DV, ADDL/II, non-positive DV, …). | Fix the dataset before trusting the fit. |
 | `omega_structure` | Warning | Mixed lognormal / additive block in omega. | Make the block's parameterization consistent. |
+
+### Run configuration and informational notes {#codes-run}
+
+| Code (`category`) | Severity | Flags | Typical agent response |
+|-------------------|----------|-------|------------------------|
 | `gradient_fallback` | Info | A gradient / sampler fallback was taken. | None; note the slower path. |
 | `mu_referencing` | Warning / Info | Missing or partial mu-referencing. | Prefer `CL = TVCL * exp(ETA_CL)` forms for SAEM/Bayes. |
 | `optimizer_config` | Warning / Info | `global_search` config note or failure. | Check the global-search setup if disabled unexpectedly. |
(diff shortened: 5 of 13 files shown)


TeunP:
## Review — `335cb9e5` (post-rebase)

Independent pass: full diff read, 10 inline angles, plus the checks CI cannot make. Ran locally in a
worktree at the PR head: `cargo test --workspace --tests --no-default-features --features ferx-core/ci`
→ **203 binaries, 0 failures**. CI is 11/11 green at this SHA, including the `ci,markov,nn` lane.

Everything below marked **measured** was produced by running code at `335cb9e5`, not by reading it.
The mutation matrix, the `Some`-vs-`None` argument and the docs split all hold up — details at the end.

---

### 1. `ode_predictions_with_solver_stats` still reports `0` — blocking

**Measured** at `335cb9e5`, calling the `pub fn` on a `NaN`-dose-time subject:

```
F1 preds_all_nan=true
F1 pub_surface abandoned_non_finite_timeline=0
```

The guard at `src/ode/predictions.rs:3397` records into `STATS_SINK` only. The
`mut stats: Option<&mut OdeSolverStats>` parameter is live and untouched at that point, so it is
skipped. The result: the one **public** function that returns an `OdeSolverStats`
(`src/ode/predictions.rs:3019`) reports an abandoned walk as `0` — the exact conflation this PR's new
public field exists to remove, on the public surface #1234's item 1 names.

Two consequences:

- A consumer reading `abandoned_non_finite_timeline` off that function cannot tell an unorderable
  timeline from a subject with nothing to integrate. That is the bug, still present, on the surface
  the issue reported it against.
- Nothing ever produces a non-scope `OdeSolverStats` with the field set, so the
  `self.abandoned_non_finite_timeline += ...` line added to `OdeSolverStats::merge` is unreachable.

The PR body argues this surface has no production caller and "would reach no user". That was a good
argument for not *routing production through it*; it is not an argument for leaving a newly-public
field structurally stuck at `0` on the only public getter, in the same PR that freezes the field into
`api/ferx-core-public-api.txt`. One line at the guard fixes it and makes `merge`'s new line live.

### 2. The warning's last sentence contradicts its own new clause — should-fix

**Measured** — the full assembled message on a fit with one `NaN`-dose-time subject:

> W_ODE_SOLVER_DIAGNOSTICS: the ODE solver did not integrate cleanly at the final estimates
> (ode_method = auto): 3 solver walk(s) were abandoned before integrating because the subject's
> timeline could not be ordered … Check the dose records and any exponential covariate model on ALAG /
> F / D / R for a value that overflows at typical covariates. **Counters are from the post-fit
> prediction pass over all subjects; consider a different ode_method, a looser ode_reltol /
> ode_abstol, or checking the parameter estimates that produce these dynamics.**

`docs/model-file/ode-models.qmd`, added in this same PR, says of exactly this case: *"It is not an
`ode_method` problem and no solver setting fixes it"*. The bolded sentence is `src/api/postfit.rs:2106`,
unconditional, and it is both the last thing the user reads and the only sentence naming concrete knobs.
The clause was carefully placed first; the boilerplate then takes it back.

The lead-in has the same problem in miniature: "did not integrate cleanly" understates a walk that
never integrated at all and whose predictions are `NaN`.

### 3. Five of the eight recorder sites have no test that dies if deleted — should-fix

`abandoned_non_finite_timeline` is asserted in exactly one file,
`src/api/tests/ode_solver_diagnostics_tests.rs` (T1–T4). Those reach three sites: `3397`, `6007`, `6546`.
Deleting the recorder at `4441`, `5421`, `7115`, `7360` or `ekf.rs:266` reddens nothing — provable by
construction, no mutation run required, since no other test can observe the field. Codecov independently
reports 2 of the 7 added `predictions.rs` lines are never executed at all (71.42% patch on that file).

CLAUDE.md: *"ask what the smallest edit that removes it is, and which test dies … If the honest answer to
'which test dies' is 'none', the fixture set is degenerate however green it is."* The matrix in the PR
body is genuinely good work on the three sites it covers; the claim it supports is "wired at all eight",
and five of those eight are unverified.

### 4. Three of the eight sites cannot fire in production, and one field-doc claim is wrong — should-fix

`SolverStatsScope::enter` appears at exactly two places in the entire tree: `src/api/fit.rs:1912` and
`:1939`. `simulate.rs`, `predict.rs` and `adaptive.rs` have none. So a recorder on a simulate-only path
is a permanent no-op:

- `ode_predictions_adaptive_impl` (`:4441`) and `adaptive_frozen_replay_tv` (`:5421`) are reached only
  through `src/sim/adaptive.rs`.
- `ode_solve_until_chz_threshold` (`:7360`) has exactly one production caller, `draw_ode_tte_latent`
  (`src/survival/mod.rs:784`), on the `simulate()` latent-event path.

The new field doc says of that last one: *"is not a prediction walk but is abandoned for the same reason
and would otherwise be the one silent hole"*. That is wrong twice. It cannot reach a scope from there;
and its `SolveFailed` arm is `panic!("ODE-accumulated TTE simulation failed for subject '{}' …")`, which
is the opposite of silent. Either wire a scope where these run, or say in the doc that three sites are
future-proofing rather than live.

### 5. `guard + record` copy-pasted eight times — should-fix (altitude)

Nothing makes the pairing structural: a ninth non-finite-timeline guard added later compiles, runs, and
records nothing, putting the diagnostic back to `0/0/0` for that walk. The PR already demonstrates the
decoupling is real — the two `sens/ode_provider.rs` guards carry the same predicate and deliberately do
*not* record, so "has the predicate" and "records" are already convention-only. A recording predicate, or
an `abandon_unorderable(&break_times) -> bool` the guards must go through, makes it impossible to forget
and would have made §3 moot.

### 6. T5–T8 pass vacuously on an empty `obs` — should-fix

T5, T6 and T7 assert only inside `for (j, o) in sens.obs.iter().enumerate()`
(`src/sens/ode_provider_tests.rs:10247`, `:10284`, `:10350`); T8 uses `sens.obs.iter().all(..)`. None
asserts the collection is non-empty, so an empty `obs` makes all four green with zero assertions run —
and "the walk bails before recording observations" is precisely what these tests are about, so it is not
a far-fetched future state. One `assert_eq!(sens.obs.len(), 4)` each closes it. Same shape, less exposed,
at `ode_solver_diagnostics_tests.rs:717` and `:789`.

### 7. `record_abandoned_non_finite_timeline` duplicates `record_to_stats_sink` — nit

`src/ode/solver.rs:1816` already does "merge these counters into the active scope", 21 lines above the new
helper at `:1837`, and the body could be
`record_to_stats_sink(&OdeSolverStats { abandoned_non_finite_timeline: 1, ..Default::default() })`.
The doc's justification argues against "merging a local `OdeSolverStats`, because the ordinary route into
the sink is the tee inside `integrate_resolved_g`" — but `record_to_stats_sink` is a directly callable
free function, not the tee; calling it never requires reaching a driver. Two hand-written `STATS_SINK`
read-modify-write bodies now have to change together.

### 8. T4's `== 3` is feature-set-sensitive in principle — nit

Green in both lanes today (I checked `Tests + coverage (TTE/CTMM endpoints)` is SUCCESS at this SHA), so
this is not a current failure. The note is that the pinned number is a property of which engines the
post-fit sweep drives, `ode_dense_solve_states` (recorder at `:7115`) has a `markov`-only production
caller at `src/markov/endpoint.rs:341`, and the test carries no feature gate. If that ever diverges, the
failure reads as a broken recorder rather than a feature-sensitive fixture. Worth one sentence in the
test's doc comment saying the count is asserted to be lane-independent.

### 9. Only `fit()` can report it — nit

Because the scope exists only in `fit.rs`, `predict()`, `simulate()` and `predict_survival()` still
return `NaN` for an unorderable-timeline subject with nothing said about it — pre-#1234 behaviour on
those paths. Fine as scope; worth stating in the PR body, since the issue's framing ("the two states
should not look identical to a diagnostic") reads as unqualified.

### 10. `OdeSolverStats` has no `#[non_exhaustive]` — nit

`src/ode/solver.rs:494` is `#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]` with no
`#[non_exhaustive]`, so a new `pub` field breaks any downstream exhaustive literal or destructure. Nothing
in this repo or in `ferx-r` does that (checked — `ferx-r` has no `OdeSolverStats` reference), and
`auto_stiff_rejected_jets` was added the same way, so this is precedent rather than a new mistake. But
"the field is additive" is a semver claim the struct's attributes do not support.

---

### Separate, and not a defect in this PR: `converged: true` on `ofv: NaN`

**Measured**, same fit: `F2 ofv=NaN converged=true`.

I found no production gate tying `converged` to a finite OFV anywhere in `src/api` or `src/estimation`.
#1295 is scoped to the jet fill, so nothing tracks this — its measured flip of `converged` to `false` is a
side effect of the `2e20` sentinel, not a designed fix. A consumer branching on `converged` is misled
harder by this than by any missing counter. Suggest a separate issue.

### What holds up

Worth saying explicitly, because these are the load-bearing claims and I tried to break them:

- **`from_f64(NAN)` lifts a constant.** `PkNum::from_f64` → `Dual1::constant` → `grad: [0.0; N]`
  (`src/sens/dual1.rs:27`), and `jets_finite()` returns `true`. The comment correction is right, and the
  old comment was wrong.
- **`integrate_tvcov_g` really has no `None` channel** — it returns a bare `Vec<Vec<T>>`
  (`src/sens/ode_provider.rs:4550`). So "`Some(non-finite)` is already the cross-walk convention" is a
  fact, not a preference, and declining the issue's `None` proposal is the correct call. Documenting the
  discriminator on both exits is the right resolution of item 2.
- **T5/T7 assert their routing** with `ode_tvcov_supported` rather than inferring it from fixture shape.
  That is the CLAUDE.md rule applied properly, and it is the reason the two guards are genuinely
  separable rather than one guard tested twice.
- **The `docs/warnings.qmd` split is sound.** Largest new `###` section is 2,935 characters against R1's
  5,000; deleting the baseline entry is correct now that the `##` has subsections, and all three inbound
  anchors still resolve.
- **T4 at `== 3` rather than `> 0`** is the right call and the reasoning in the body is correct.
- No `Cargo.lock` change, and `ferx-r` has no reference to `OdeSolverStats` — the "no cross-repo PR"
  conclusion checks out.

### On splitting #1295 out

Agreed, keep it split. I'd only push back on the framing: a healthy subject's η̂ moving is the ordinary
consequence of the outer objective changing from `NaN` to the `2e20` sentinel — the population
parameters move, so every subject's EBE moves. It reads in the PR body as cross-subject contamination,
which would be alarming; it isn't that.

---

**Blocking:** §1. **Should-fix before merge:** §2, and §3/§4 at least as an honest scoping note in the
PR body if not as code. The rest can be follow-ups.

TeunP:
## Round 2 — every finding, and what happened to it

Fixed at `ece0cfb2` / `a423383a` / `8d13b805`, rebased onto `19bf7cfd`. Each fix was
**reproduced before it was written**: §1's defect was made red first, and §2's replacement
wording was measured out of a real `fit()` rather than read off the source.

| # | Finding | Disposition | Fix |
|---|---------|-------------|-----|
| 1 | `ode_predictions_with_solver_stats` still reports `0` — **blocking** | fixed | `ece0cfb2` |
| 2 | the warning's last sentence contradicts its own new clause | fixed | `a423383a` |
| 3 | 5 of 8 recorder sites have no test that dies | fixed | `ece0cfb2` |
| 4 | 3 sites cannot fire in production; the field doc is wrong twice | doc corrected; widening the scope filed | `8d13b805` |
| 5 | `guard + record` copy-pasted eight times | fixed | `ece0cfb2` |
| 6 | T5–T8 pass vacuously on an empty `obs` | fixed | `8d13b805` |
| 7 | `record_abandoned_non_finite_timeline` duplicates `record_to_stats_sink` | fixed | `ece0cfb2` |
| 8 | T4's `== 3` is feature-set-sensitive in principle | fixed — **and the premise was wrong**, see below | `8d13b805` |
| 9 | only `fit()` can report it | stated in `ode-models.qmd`; filed with §4 | `8d13b805` |
| 10 | `OdeSolverStats` has no `#[non_exhaustive]` | not here — filed | — |
| — | `converged: true` on `ofv: NaN` | filed | — |

### §1 — reproduced, then fixed at the guard

Made red first, exactly as you measured: `preds` all `NaN` and *every* field of the returned
block `0`, not just the new one. `record_abandoned_non_finite_timeline` now takes the
`Option<&mut OdeSolverStats>` and feeds both channels through `record_to_stats_sink` +
`OdeSolverStats::merge` — which is also §7's fix (the second hand-written `STATS_SINK`
read-modify-write is gone) and makes `merge`'s new line reachable, since nothing else produces
a non-scope block carrying the field.

### §5 — the pairing is structural now, and enforced

`ode::predictions::abandon_non_finite_timeline` detects **and** records in one call; all eight
engine guards go through it. `the_bare_timeline_predicates_are_not_called_outside_this_guard`
pins the bare predicates' call sites at exactly `src/ode/predictions.rs` ×2 (inside the two
helpers themselves) and `src/sens/ode_provider.rs` ×2 (the gradient walks that deliberately do
not record). Exact counts, not a floor — so the allowance cannot quietly widen *or* vanish.
That is also why each §3 mutation below kills two tests rather than one.

### §3 — all eight sites now kill a test, each naming its own side

Every arm carries a straddle (same engine, same fixture, finite lag → must record `0` **and**
actually integrate), so a recorder that fired unconditionally fails too.

| mutation (guard → bare predicate) | killed by |
|---|---|
| M1 out-param dropped at the `ode_predictions` guard | T9 **only** — T1–T4 green, which is the finding |
| M2 `ode_predictions` | T1, T3, T4, T9, per-engine, structural |
| M3 adaptive driver | its own test + structural (frozen-replay twin green) |
| M4 frozen replay | its own test + structural (driver twin green) |
| M5 `ode_predictions_with_states` | T4, per-engine, structural |
| M6 `ode_dense_solve_states` | per-engine, structural — **had no test before** |
| M7 `ode_solve_until_chz_threshold` | its own test, structural — **had no test before** |
| M8 `ode_predictions_event_driven` | T2, per-engine, structural |
| M9 `ekf.rs` | its own test, structural — **had no test before** |
| M10 knob advice unconditional (§2's defect restored) | T10 |
| M11 abandoned advice dropped | T10 |
| M12 lead-in verb unconditional | T10 |

12 run, 12 killed, no survivors; the harness restored byte-identically and verified it.

### §2 — measured, then reworded

`unclean` splits into `unclean_integration` (every counter describing a segment that *ran*) and
the abandoned count. The knob advice attaches to the former and is scoped to it; an abandoned
walk gets advice that agrees with `ode-models.qmd`; the lead-in verb branches too. The full
assembled message on the same fit:

> W_ODE_SOLVER_DIAGNOSTICS: the ODE solver **did not produce a usable integration** at the final
> estimates (ode_method = auto): 3 solver walk(s) were abandoned before integrating because the
> subject's timeline could not be ordered … Counters are from the post-fit prediction pass over
> all subjects. **The abandoned walk(s) are not an ode_method or tolerance problem — nothing was
> integrated for them, so no solver setting changes the outcome; fix the record or the parameter
> that produces the non-finite time.**

`the_solver_knob_advice_is_attached_only_to_counters_it_applies_to` covers all three shapes
(abandoned only / an ordinary unclean integration / both), because a single-arm test cannot see
a branch.

### §8 — right conclusion, wrong premise

Taken, and the sentence is in the test. But `ode_dense_solve_states` does **not** have a
`markov`-only production caller — it has five: `markov::endpoint:341`, `survival/mod.rs:728`,
`api/output_columns.rs:640`, `ode_predictions_event_driven_with_states` and
`adaptive_window_signal_aucs`. The reason the count is lane-independent is different and
stronger: the scope wraps `compute_subject_results` only (`api/fit.rs:1912`), the callers
reachable from there need a model feature this fixture does not have (a markov endpoint, a
hazard state, a TV-covariate model), and the `output_columns` caller runs *after* the scope
closes. The comment says that, and says to gate the test rather than loosen it to `> 0` if it
ever changes.

### §4 / §9 — corrected in the doc, filed as a change

The field doc no longer calls `ode_solve_until_chz_threshold` "the one silent hole" — verified:
its sole production caller `panic!`s on `SolveFailed` (`survival/mod.rs:789`). It now states
that three of the eight sites cannot fire in production, because outside tests
`SolverStatsScope::enter` appears only at `api/fit.rs:1912` and `:1939` (verified), and why they
carry the recorder anyway. `docs/model-file/ode-models.qmd` states the `fit()`-only scope.

### One measurement that fell out of §3

`ode_solve_until_chz_threshold` runs on `solver::solve_ode_until_threshold`, its own driver,
which does **not** go through `integrate_resolved_g`'s `STATS_SINK` tee — its step counters read
`0` even on a solve that crossed the threshold normally (measured; the first version of that
test failed on it). So its control asserts the *outcome*, not the counters, and on that one
engine the new counter is the only thing that can separate "abandoned" from "nothing to
integrate".

### Follow-ups filed

- **#1302** — `OdeSolverStats` is not `#[non_exhaustive]`, so every added counter is a
  breaking change. Not fixed here: adding the attribute is itself breaking, and wants its own
  baseline diff and CHANGELOG line.
- **#1303** — `fit()` returns `converged: true` with `ofv = NaN` (your side finding).
- **#1304** — `predict()` / `simulate()` open no `SolverStatsScope`, so their ODE solver
  diagnostics are dead (§4 and §9).

### Gates

`tools/preflight.sh` green (all six groups; the public API still matches the committed baseline
— no new `pub` items this round). Test legs, both 0 failed:

| leg | binaries | passed |
|---|---|---|
| `--workspace --tests --no-default-features --features ferx-core/ci` | **203** | 5877 |
| `--tests --no-default-features --features ci,markov,nn` | **181** | 5568 |

203 is the same binary count as your round-1 run. CI at `8d13b805` is **11/11 green**, including
`codecov/patch` — the two previously-unexecuted `predictions.rs` lines are covered now.

The two clippy warnings in `simulate.rs:771` and `sir.rs:806` are pre-existing on `origin/main`,
not from this branch (checked with `git show origin/main:<file>`).

Issue #960, opened 2026-08-04
fix(estimation): FD-of-OFV covariance Hessian is knife-edge fragile at the warfarin FOCEI optimum (SE(TVCL) 0.0071↔121 on a 3e-5 θ shift)


ronkeizer:
## Summary

Three slow covariance-SE cross-checks are red on `main` because ferx's **reconverged FD-of-OFV covariance Hessian is knife-edge fragile at the warfarin FOCEI optimum** — a ~3e-5 difference in the TVCL point estimate flips SE(TVCL) between the correct `7.10e-3` and a regularized blow-up (`121`). This is a covariance-conditioning problem (**#520 / #432 / #436**), *not* a convergence-flag bug (fixed separately in #751) and *not* a covariance-math error (the math reproduces NONMEM `MATRIX=R` exactly at the right point).

Affected tests (all `slow-tests`-gated, temporarily `#[ignore]`d pending this issue):
- `warfarin_covariance_nonmem.rs::covariance_se_matches_nonmem` (FOCEI)
- `covariance_method_sandwich.rs::covariance_methods_produce_consistent_ses_on_warfarin`
- `covariance_method_sandwich.rs::covariance_se_matches_nonmem_s_rsr`

## Evidence

Warfarin FOCEI (`data/warfarin.csv`, 1-cpt oral, proportional error), covariance `MATRIX=R`, NONMEM SE(TVCL) = `7.097e-3`:

| Optimizer | OFV | TVCL | SE(TVCL) |
|---|---|---|---|
| SLSQP | −286.00422 | 0.13269 | **7.097e-3** (= NONMEM) |
| BOBYQA | −285.88 | 0.13299 | 7.315e-3 |
| Auto → NLopt L-BFGS (default) | **−250.838 (= init)** | **0.150 (= init)** | 2.225e-2 |
| L-BFGS w/ first-step gradient cap | −286.00414 | 0.13272 | **121** |

Two distinct failure modes stacked:

1. **Default NLopt L-BFGS is stuck at the initial estimates.** From `H₀ = I` the first search direction is `−∇`; the scaled gradient at init is ~150, so the opening step overshoots (σ pinned to its bound), the More-Thuente line search fails at **eval 1**, and the fit never leaves its initial θ. (Before 2026-07-02 the now-removed SLSQP fallback (#657) rescued this; #639 made analytic L-BFGS the `Auto` default.)

[...]

Issue #657, opened 2026-07-01
refactor(estimation): remove the automatic SLSQP fallback after a non-converged outer optimization


roninsightrx:
## Summary

Remove the automatic **SLSQP fallback** that re-runs the outer optimization from the
current point when the primary optimizer stops without clean convergence. It does not
appear to earn its cost and it adds confusing output.

## Current behavior

After the primary NLopt outer run, the loop re-optimizes from the last point with SLSQP
whenever the run did not converge cleanly:

- Gate — `should_run_slsqp_fallback` (`src/estimation/outer_optimizer.rs:455`):
  ```rust
  !converged && !already_slsqp && !cancelled && !max_eval_reached
  ```
- Fallback block — `src/estimation/outer_optimizer.rs:~1278-1515`, emitting
  `Retrying with NLopt SLSQP from current point...` (line ~1280, gated on `verbose`).
- Introduced / last touched by #499 (`perf(estimation): skip the SLSQP fallback on
  MaxEvalReached`).

There is no setting to disable it; a user can only avoid it by running `optimizer = slsqp`
as the primary, hitting `maxiter`, or converging.

## Why remove it

- **Low value.** Restarting a *different* local optimizer from the *same* point rarely
  rescues a genuinely non-converged fit — if the primary stalled at a point, SLSQP from
  there usually stalls too (or reports a marginal move that isn't real convergence).
- **Confusing.** The `Retrying with NLopt SLSQP...` line reads like an error/instability
  and prompts user questions, when the real signal is "the primary optimizer did not
  converge — look at the model / init estimates / maxiter."
- **Cost.** A full second outer optimization (with inner EBE loops) roughly doubles
  wall-time on exactly the runs that are already slow (non-converged ones).
- **Masks non-convergence.** A borderline SLSQP "success" from the fallback can hide that
  the primary optimizer didn't actually converge.

## Proposed change

- Delete `should_run_slsqp_fallback` and the fallback block; when the primary optimizer
  stops without converging, surface that directly (keep/adjust the existing
  non-convergence warning in `FitResult.warnings`).
- Remove the `slsqp_fallback_gate` unit test (`src/estimation/outer_optimizer.rs:3967`) and
  any fallback-specific plumbing (`best_seen` is already shared and stays).
- Users who want SLSQP keep `optimizer = slsqp`.

## Open question

- Keep an opt-in escape hatch (e.g. `outer_fallback = slsqp`) for anyone who relied on the
[...]

PR #1297, opened 2026-09-08
fix(estimation): anchor the EBE warm start to the best point seen [closes #1290]

### LINKED ISSUE #1290: bug(estimation): a model with covariate thetas stalls at its initial estimates on two_cpt_oral_cov, so every covsearch candidate is rejected
Found while wiring the R surface for covsearch (ferx-r #332). It is not an R-side problem - the R layer only reports what the engine returned - but it makes covsearch unable to accept any effect on this dataset, so it is worth its own issue.

## What happens

On `examples/two_cpt_oral_cov.ferx` + `data/two_cpt_oral_cov.csv`, the fit does not move: every theta comes back at exactly its initial value and the optimizer reports a failure on its first evaluation.

```
Starting NLopt Lbfgs optimization (13 parameters)...
Eval    1: OFV = -1193.591896
NLopt stopped: Failure
Restored best-seen point (OFV = -1193.591896) for final inner loop and covariance step.
Final OFV = -1193.591896
Plateau check: flat_tail = 11 feasible evals (min 5), feasible_evals = 12,
  best-seen Some(-1193.591896383789) vs final -1193.591896, left init = false -> converged = false
```

Estimates after the "fit":

```
      TVCL       TVV1        TVQ       TVV2       TVKA   THETA_WT THETA_CRCL
      5.00      50.00      10.00     100.00       1.20       0.75       0.50
```

i.e. the declared initial values, unchanged to every digit.

## It is the covariate parameters, not the dataset

The same model with the covariate terms removed from `[individual_parameters]` fits normally on the *same* dataset:

```
converged: TRUE   ofv: -1185.24
  TVCL   TVV1    TVQ   TVV2   TVKA
 4.437 47.172  9.595 94.497  1.277
```

And adding a single covariate relation back - through `[covariate_model]` this time, which is a different code path from the inline form - reproduces the stall:

```
   model       ofv converged passed
1   base -1185.240      TRUE   TRUE
2   scaled -1179.998    TRUE  FALSE
failures: stalled at the initial estimates: no free parameter moved more than 1% of its initial value (#751)
theta:  TVCL 5.00  TVV1 50.00  TVQ 10.00  TVV2 100.00  TVKA 1.20  THETA_CL_WT 0.75
```

So: base model fits; base + one covariate theta stalls; and it stalls whether the effect is written inline in `[individual_parameters]` or as a `[covariate_model]` relation.

`gradient` does not change the verdict, only how it is reported. The default (analytic) path reports `converged = false`; `gradient = fd` reports `converged = true` while leaving every estimate at its initial value, which the strictness gate then catches as an init stall (#751).

## Why it matters beyond this one example

This is the failure mode #1175 names as the dominant statistical risk, arriving through the front door. covsearch fits `base + one effect` for every candidate, so on this dataset *every* candidate stalls at its initial estimates: the covariate theta never leaves its PsN init, dOFV comes out ~0.01, and the search correctly concludes that nothing is worth adding - on data where the covariate effects are real (the covariate model's OFV at its initial values, -1193.59, is 8.35 better than the base model's converged -1185.24).

The strictness gate does its job - the candidates are reported as `passed = false` with the stall named - so the search does not select a stalled winner. But a tool whose every candidate is rejected for the same reason is a tool that cannot run, and the same shape will hit modelsearch and iivsearch, whose candidates also differ from their parent by one or two added parameters.

## Repro

```
cargo run --release -- fit examples/two_cpt_oral_cov.ferx --data data/two_cpt_oral_cov.csv
```

Observed with `ferx-core` `main` at `6ec94356`.

## Where to look

The optimizer gives up on evaluation 1 (`NLopt stopped: Failure`) rather than exhausting iterations, which suggests the failure is in the first gradient or in the line search rather than in the surface itself - the OFV at that point is finite and is the best-seen value. A parameter-scaling question is the obvious suspect (#341): the covariate thetas (0.5, 0.75) sit three orders of magnitude below TVV2 (100), and the base model that omits them optimizes without trouble.

roninsightrx:
## Why

`ferx` fits stopped on their **first evaluation** and reported every parameter at its declared initial value ([#1290](https://github.com/FeRx-NLME/ferx-core/issues/1290)). Models carrying covariate thetas were the visible casualty, which makes it a `covsearch` blocker: every candidate is `base + one effect`, so every candidate stalled, `dOFV` came out ~0.01, and the search correctly concluded that nothing was worth adding — on data where the effects are real.

The issue's suspect was parameter scaling (#341). It is not. **The objective is path-dependent.** The outer loop's NLopt objective closure adopted the empirical Bayes estimates of *every* evaluation as the next evaluation's inner-loop warm start, including the ones the line search is about to reject. The individual objective is multimodal (#864 / #891), so one bad probe leaves the cache in a worse basin that the inner loop keeps re-finding, and the optimizer's own starting point then evaluates worse than it did a moment earlier.

The optimizer trace on `examples/two_cpt_oral_covmodel.ferx` (`optimizer_trace = true`, `optimizer = bobyqa` removed):

| eval | step_norm | OFV | note |
|---|---|---|---|
| 1 | — | **−1026.350403** | initial estimates |
| 2 | 0.81 | +15899.30 | `TVV2` driven onto its upper bound 500; 10 of 30 subjects fall back to Nelder-Mead |
| 3 | 0.80 | −1010.62 | backtrack |
| 4–7 | 8e-3 → 0 | −1002.84 → −1002.7598 | line search collapsing |
| 8–12 | 0 | **−1002.759781** | parameters **bit-identical to eval 1** |

23.6 OFV units apart at the same `xs`. Every inner loop reported `n_unconverged = 0`, so this is not a starved inner solve — it is a different, worse EBE mode. No step can show a decrease against a starting value that has moved, so NLopt returns a bare `Failure` on evaluation 1.

## What changed

`src/estimation/outer_optimizer.rs` — the EBE cache is adopted only from an evaluation that improves on the best objective seen, and never from a guard-rejected one:

```rust
fn adopt_warm_start(guarded: bool, ofv: f64, best_ofv: f64) -> bool {
    !guarded && ofv < best_ofv
}
```

The `!guarded` half matters independently: a guarded eval's `ofv` is `guard_penalty_value`, a synthetic distance-to-centre penalty on its own scale, so on a model whose objective is *positive* it lands below `best_ofv` and would adopt EBEs from a point the EBE guard has already rejected. (This repo's covariate examples happen to sit around −1000, which is why the negative-OFV fixture cannot see it — hence the dedicated unit test.)

Non-mixture fits need no snapshot at all: the update is a plain assignment that is skipped. Only the mixture branch writes `cached_etas_by_class` before the objective is known, so only a mixture pays a clone — this is the per-eval hot path.

## Alternatives considered

- **Parameter scaling / gradient preconditioning** (the issue's suspect). Ruled out by the trace: the same `xs` yields two different objective values, which no rescaling addresses. Also, per #864, single-lever scaling fixes regress elsewhere.
- **`inner_restarts`** (the #891 EBE multistart). Measured: `inner_restarts = 0` and `= 5` both give exactly −1195.3043 on this model. The multistart scan fires on cold starts, and the outer loop's warm-started evals never reach it.
- **Cold-starting the inner loop on every eval.** Removes the path dependence outright but discards the warm start's whole benefit.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` green). ferx-r #332's covsearch surface is unblocked by this without any R-side change.

## Breaking changes
- [x] None

Behaviour changes: fits that previously stalled now move, and fits that previously converged may land at a different (lower) optimum. No format, field or API change.

## Numerical validation

**NONMEM anchor** — `nonmem_anchor/covmodel_stall.ctl` (ADVAN4 TRANS4, `$EST METHOD=1 INTERACTION MAXEVAL=9999`, `$COV MATRIX=R`), same start, `MINIMIZATION SUCCESSFUL`. Committed with `results/covmodel_stall.{ext,lst}` and a README row. This is an *estimation* anchor, not a prediction one — nothing is fixed, the question is where the optimizer lands.

```
# before — ferx, examples/two_cpt_oral_covmodel.ferx, default optimizer
NLopt stopped: Failure   (eval 1)
OFV -1026.350403
TVCL 4.0  TVV1 40.0  TVQ 8.0  TVV2 80.0  TVKA 1.0
THETA_CL_WT 0.6  THETA_CL_CRCL 0.3  THETA_V1_WT 0.6      <- the declared inits

# after — ferx, same file, default optimizer
Converged: YES   OFV -1195.3043

# after — ferx, same file, optimizer = lbfgs (built-in BFGS)
OFV -1199.429664
TVCL 4.94250  TVV1 48.5638  TVQ 9.57563  TVV2 94.1660  TVKA 1.25619
THETA_CL_WT 0.728967  THETA_CL_CRCL 0.554899  THETA_V1_WT 0.581049

# reference — NONMEM 7.5.1 FOCEI, MINIMIZATION SUCCESSFUL
OFV -1199.4297842339142
THETA 4.94183  48.5627  9.57581  94.1662  1.25629
      0.729589  0.554910  0.577399
SIGMA(1,1) 3.84918E-04  (SD 0.019619)
OMEGA 0.107340  0.106110  0.0513412  0.0379481  0.174964
```

OFV agrees to **1.2e-4**, thetas to ~4 significant figures.

`examples/two_cpt_oral_cov.ferx` (the hand-written twin) goes from −1168.4846 to **−1199.3264** on the default optimizer, and its `PROP_ERR` 0.019619 matches NONMEM's `sqrt(SIGMA(1,1))` exactly.

**Known residual, not part of this fix.** The default NLopt L-BFGS reaches −1195.30 where the optimum is −1199.43. It stops on `FtolReached` against a *warm-started* objective reading −1178.09 at a point whose cold-start value (the final inner loop, `warm = None`) is −1195.30. So the outer optimizer minimises a biased surrogate of the FOCEI objective — a real and separate defect, on which `inner_restarts` has no effect (measured above). Worth its own issue, since a 4 OFV bias matters against a `dOFV` cutoff of 3.84.

## Performance

The non-mixture path removes an assignment on non-improving evals and adds nothing. Mixtures pay one `Vec<Vec<DVector>>` clone per eval, dominated by the per-class inner solve in the same eval.

## Tests

`src/estimation/outer_optimizer_tests.rs`:

- `covariate_model_leaves_its_initial_estimates_1290` — Tier 1, 0.10 s release / 1.2 s dev. `outer_maxiter = 3` (45 evals) on `examples/two_cpt_oral_covmodel.ferx`. **Mutation-verified**: with `adopt_warm_start` forced to `true` it returns exactly `-1026.350403` with theta `[4.0, 40.0, 8.0, 80.0, 1.0, 0.6, 0.3, 0.6]`. The bound (−1100) is measured, not chosen: 95.0 units above the realised −1195.399628 and 73.7 below the stall.
  - Note the test uses `Optimizer::Auto` deliberately. An earlier draft used `Optimizer::Lbfgs` — which is the **built-in BFGS**, a different path — and passed under the pre-fix code. The comment says so.
- `adopt_warm_start_takes_only_feasible_improvements` — covers the `!guarded` half, which the negative-OFV end-to-end fixture structurally cannot reach.

**One existing test had to be rebuilt, and it is worth a look.** `non_convergence_reports_directly_without_second_optimization` (#657: no automatic SLSQP retry) forced its non-converged shape with a starved inner loop — and that shape only existed *because* the guard-rejected eval 1's EBEs were adopted. With that gone, the guard fires once and the fit converges at 59.79. Its docstring records that this is the **second** time the fixture has been emptied by a fix elsewhere (pre-#960 it rode the L-BFGS first-step stall). It is now built on an exhausted evaluation budget, and asserts the *mechanism* (`"evaluation budget"` in the warnings) alongside the verdict, so a third silent emptying fails instead of passing for a new reason.

Full suites run locally, all green: `--lib --features ci,survival` (4420), `--lib --features ci,markov,nn` (4574), `--workspace --tests --features ferx-core/ci`, and `tools/preflight.sh` (fmt / 5 check sets / clippy / rustdoc / docs / public-API) plus `tools/preflight.sh debug-assertions` (4578).

## Example

`examples/two_cpt_oral_covmodel.ferx` loses its `optimizer = bobyqa` workaround — the comment justifying it described this stall. `examples/amd_start.ferx` keeps the key (so `docs/tools/amd.qmd`'s numbers keep describing the run it produces) but its comment is corrected: the default now converges there too, −172.234 against BOBYQA's −172.233.

## Docs
- [x] `docs/` updated — `docs/estimation/foce.qmd` "Warm Starting" now states that the warm start comes from the best point seen, and why

## Checklist
- [x] Nothing on the `Dual2` sensitivity path touched

## Reviewer hints

The whole change is ~10 lines in `optimize_nlopt_once` plus `adopt_warm_start`. Two places to focus:

1. **The mixture branch.** It is the only one that writes `cached_etas_by_class` *before* the objective is known, so it is the only one needing a snapshot — the non-mixture cache is preserved by simply not writing it. Worth checking I have not missed a third writer.
2. **`ehs` is dropped when the eval does not improve.** Intentional: those are the EBEs of a point we are declining to carry forward. `cached_h_mats` is still written unconditionally — it is write-only in this file (three assignments, no readers), so it was left alone rather than given a conditional it does not need.

The rebuilt `non_convergence_reports_directly_without_second_optimization` deserves more attention than the fix does; see the Tests section.

## Open questions

The −1195.30 vs −1199.43 residual on the default optimizer. I have the diagnosis (the outer loop minimises a warm-start-biased surrogate; the final inner loop cold-starts and reads 17 OFV units lower at the very same parameters) but not a fix that is in scope here — the obvious candidates all change the objective for every model. Should I open that as its own issue, or would you rather it be folded in?

https://claude.ai/code/session_017rkEkHatTAb6bkHPPDvqqs


roninsightrx (commented):
Reviewed the EBE-cache state flow end to end, including the mixture-specific cached_etas_by_class snapshot/restore path and the ordering against best_ofv updates. I found no blocking issues.

Focused local checks passed:
- adopt_warm_start_takes_only_feasible_improvements
- covariate_model_leaves_its_initial_estimates_1290
- non_convergence_reports_directly_without_second_optimization
- mixture_fit_slsqp_uses_analytic_gradient
- mixture_default_optimizer_does_not_warn

CI is also green, including both coverage jobs, release-semantics tests, clippy, docs, and the public-API baseline.

PR #1299, opened 2026-09-09
feat(tools): globalsearch — GA / exhaustive with pyDarwin-style penalized fitness [closes #1185]

### LINKED ISSUE #1185: feat(tools): global search — GA / exhaustive enumeration with pyDarwin-style penalized fitness [#1175]
Part of #1175 (P6). Optional, after #1180 and #1181. Depends on #1176, #1178.

Global search over the same candidate layer — pyDarwin's contribution to this epic.

## What is worth taking from pyDarwin

**Penalized fitness.** Rank on a scalar that charges for complexity and for failure, rather than on raw OFV:

| Penalty | pyDarwin default |
|---|---|
| per estimated θ | 10 |
| per estimated Ω element | 10 |
| per estimated σ element | 10 |
| failure to converge | 100 |
| covariance step failed | 100 |
| any \|correlation\| > 0.95 | 100 |
| condition number > 1000 | 100 |
| non-influential token | 0.00001 |
| crash value (no output) | 99999999 |

The inputs are the same ones #1177 already computes for `Strictness`, so this is a scoring function over an existing verdict, not new numerics.

**Global algorithms.** GA (population size, generations, downhill period, niche count and radius) and exhaustive enumeration over the feature grid. Both consume `Vec<Candidate>` from the existing runner (#1178).

## What is not worth taking

- **The template + tokens file pair.** It exists to compensate for not having a structured model object. ferx has a parser and #1176; adopting it would be a step backwards.
- **GP / RF / GBRT surrogates.** They pay off when each fit is a NONMEM subprocess costing minutes. ferx fits in-process, so the surrogate's own overhead is a large fraction of what it saves. Revisit only if profiling on ODE-heavy spaces says otherwise.
- **The grid run manager.** Single-machine rayon is the scope of this epic.

## Acceptance

- [ ] Penalized fitness reproduces a hand-computed score on fixtures covering every penalty term.
- [ ] On a search space small enough to enumerate, the GA finds the same optimum as exhaustive enumeration.
- [ ] The GA and the stepwise tools share one candidate runner and one cache — a candidate fitted by either is reused by the other.
- [ ] `docs/tools/global-search.qmd`, stating plainly when a GA beats stepwise and when it does not.

roninsightrx:
## Why

P6 of the #1175 epic, closes #1185: the pyDarwin contribution — a **global** search over the same candidate layer the stepwise tools use, ranked on pyDarwin's **penalized fitness**. The stepwise tools (modelsearch, covsearch, AMD) walk one axis at a time from the last winner, which is the right tool when the axes are separable and the wrong one when they interact (a covariate that only pays off under the right absorption model, a lag that only beats a transit chain with two compartments): stepwise commits to one axis before it has seen the other. `[rank] type = "penalized"` has been parsed-but-refused since #1179.

## What changed

**Penalized fitness — a `Criterion`, usable by every tool** (`search/penalty.rs`). `Criterion::Penalized(Penalties)`: OFV + 10 per estimated θ / Ω / σ element + 100 for non-convergence, a failed/absent covariance step, |r| > 0.95, condition number > 1000 — pyDarwin's schedule verbatim, over the inputs `check_strictness` already reads. `[rank.penalties]` overlays any charge. The whole schedule goes into the run manifest key so two penalized runs with different charges never share a journal. modelsearch / iivsearch / iovsearch now carry the file's schedule (`RankType::criterion_with`).

**`ferx globalsearch`** (`crates/ferx-tools/src/globalsearch/`). The `.ferxsearch` space is laid out as **one grid**: every structural category an axis with its values as alleles, every `COVARIATE?` pair an axis with `none` and each of its forms. A genome is decoded with the *same* edits the stepwise tools make (`SetStructural` from modelsearch's `structure.rs`, `AddCovariateRelation` from covsearch's `Effect`), seeded from the input fit. Two algorithms:

- `exhaustive` — every point, one batch, refused above `max_models` rather than truncated.
- `ga` — pyDarwin's GA over integer genes (`ga.rs`, written over an abstract `Oracle` so it is tested without fits): tournament selection on shared fitness (Goldberg–Richardson sharing on Hamming distance), one-point crossover, per-gene mutation, elitism, periodic + final one-gene downhill search on niche seeds. Seeded, so `--resume` proposes the same genomes and refits nothing.

Three charges the criterion cannot see, applied by the search: a **non-influential gene** (a covariate on a parameter the structural choice removed — the relation is not written, the runner fits the rendered model once, the gene pays pyDarwin's 1e-5 tie-break), a **crash** (unbuildable point / no fit → 99 999 999), and — ferx's own, because ferx has a strictness *gate* where pyDarwin has only penalties — a **`gate`** charge (default 100) on a fit the gate refused, so an unselectable model steers the GA without ever winning it. The gate decides eligibility, the fitness decides rank; `docs/tools/global-search.qmd` § "The gate and the penalties" states this.

**One runner, one cache — across tools.** `Runner::reuse_from(dir)` (+ `[run] reuse_from = [...]` for every tool, `--reuse-from DIR` on the CLI): every journal found under another search's directory is read, and a candidate whose canonical hash a fit there carries is **re-scored** under this run's criterion and gate rather than fitted again (`SearchManifest::same_fits`: same data, starts, fit settings; criterion and gate deliberately *not* compared, since both are functions of the `FitResult`). A fitless row is trusted only under identical scoring. Reused fits are journalled as the run's own so the other directory need not outlive it.

## Alternatives considered

- **Penalized as a tool-local scoring over `CandidateResult`** rather than a `Criterion` variant — would have left the stepwise tools unable to rank on it and the manifest blind to the schedule. The variant costs `Criterion` its `Eq` (f64 fields); nothing relied on it.
- **Gate-failed fitness = criterion (pyDarwin's posture)** — under a BIC criterion the GA then chases non-converged low-OFV models it can never select. **= crash** — loses the gradient the population needs. The explicit `gate` charge (settable to 0) is the middle.
- **Bit-coded genomes (pyDarwin/DEAP)** — integer alleles are the same space without the bit strings that encode nothing between two alleles.
- **A shared cache directory across tools** — would need the journal rewritten; reading foreign journals read-only and re-scoring is smaller and keeps every tool's directory self-contained.
- Not taken from pyDarwin, per the issue: template+tokens files, GP/RF/GBRT surrogates, the grid run manager.

## Cross-repo dependency

No `ferx-core` public item added or changed (`tools/preflight.sh` public-API gate green); everything is in `ferx-tools` / `ferx-cli`.

## Breaking changes
- [x] None

## Numerical validation

No new numerics: every candidate's OFV comes from the same `fit()` the modelsearch NONMEM anchor and the covsearch PsN `scm` anchor already pin, and the penalized fitness is arithmetic over fields those fits already carry. What the issue asked to validate, and where:

- **Penalized fitness reproduces a hand-computed score on fixtures covering every term** — `search/penalty_tests.rs`: one fixture per term (θ/Ω/σ counts, convergence, each covariance status, |r| at and above the threshold, CN above / NaN), the arithmetic written beside each assertion; a custom schedule scales each term; an old bundle without a tally charges at the θ rate.
- **The GA finds the same optimum as exhaustive enumeration on a space small enough to enumerate** — twice. `ga_tests.rs`: a 216-point deceptive landscape (a decoy basin a one-gene hill climb provably ends in — asserted — and the true optimum across it) for 5 seeds, each finding the exhaustive minimum in < 216 evaluations. `mod_tests.rs` + `tests/globalsearch_end_to_end.rs`: the same on the model layer with a scripted fitter (36-point grid, interaction term, 4 seeds) and with **real evaluations** on `two_cpt_oral_cov` (32-point structural × covariate grid): same winner, bit-identical fitness, fewer fits.
- **One runner, one cache** — `runner_tests.rs` (4 tests: re-score under another criterion, journalled as own + resume without the other dir, incompatible dir skipped with a named warning, fitless rows trusted only under identical scoring, own journal wins) and the end-to-end resume that refits nothing.

- [x] Not applicable to NONMEM (search algorithm; see above)

## Performance
- [x] No significant impact expected — the stepwise tools gain one field and one `Vec` read; the runner's foreign scan only runs when `reuse_from` is set.

## Tests
- [x] Unit tests: 11 GA (`ga_tests.rs`), 13 search (`mod_tests.rs`), 7 penalty → 577 → 582 `ferx-tools --lib` (plus 4 runner reuse + 1 config + 1 CLI-arg in their own files)
- [x] Tier-2: `tests/globalsearch_end_to_end.rs` (3, evaluations only), 2 CLI tests in `cli_ferx.rs`
- [x] Regression-shaped: the hill-climb-vs-GA test fails if `run` degrades to a hill climb; the ordering fix (unbuildable rows pushed out of creation order) and the pair-table fix (`TRANSITS(1)` on an FO base not refused) each have the assertion that caught them.

- [x] `examples/amd_start_globalsearch.ferxsearch` — a 324-point grid (PERIPHERALS × LAGTIME × four `@IIV`-`@CONTINUOUS` pairs) on the AMD starting model, GA with `population_size = 12`, `generations = 4`.

## Docs
- [x] `docs/tools/global-search.qmd` (new; says when a GA beats stepwise and when it does not), linked in `_quarto.yml` and `tools/index.qmd`; `search.qmd` (`penalized` row, `[run] reuse_from`, `[globalsearch]` section), `cli.qmd`

## Checklist
- [x] No `sens/` / `pk/` code touched

### ferx-site / ferx-book
- [ ] New example file → site/book entry: follow-up, as for the other `.ferxsearch` examples

### Example execution
- [x] `target/release/ferx globalsearch examples/amd_start_globalsearch.ferxsearch --threads 8` runs clean (run log in the PR comments)

## Reviewer hints

- `search/penalty.rs` is 200 lines and the whole of the new numerics; `penalty_tests.rs` is its spec.
- `globalsearch/ga.rs` is pure; `ga_tests.rs::the_ga_beats_a_hill_climb_on_the_deceptive_landscape` is the one to read to see the landscape.
- `globalsearch/mod.rs::Evaluator::evaluate` is the seam between the GA and the runner: decode → dedup by genome and by hash → one `fit_step` per batch → rows in proposal order. The `Entry` enum exists because the first version pushed unbuildable rows before fitted ones and the table came out of order.
- `search/runner.rs::load_foreign` / `foreign_result` — the cross-tool reuse; the `same_fits` vs `same_scores` split in `journal.rs` is the design.
- Skim: `report.rs`, `globalsearch_cmd.rs` (thin, mirrors `modelsearch_cmd.rs`).

## Open questions

- AMD integration is not in scope (the issue calls P6 optional); a `global` step after the pipeline, reusing the pipeline's fits through `reuse_from`, is the obvious follow-up.
- `--reuse-from` is a CLI flag on `globalsearch` only; the other tools take it from the file. Worth adding the flag to all of them?

https://claude.ai/code/session_01WYANNcJgxCCBLe3cAfGetq


roninsightrx:
Run log of the shipped example, release binary, `--threads 8` (11 s wall, 69 models):

```
$ target/release/ferx globalsearch examples/amd_start_globalsearch.ferxsearch --threads 8
  LAGTIME: OFF | ON
  PERIPHERALS: 0 | 1
  CL-WT: none | power | linear
  CL-CRCL: none | power | linear
  V-WT: none | power | linear
  V-CRCL: none | power | linear
Input model: OFV -172.233, penalized -112.233
Algorithm: ga, 69 models fitted of 69 evaluated

  model    genome                                        npar          OFV    penalized      fitness rank seconds  status        decision
  run47    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    10     -724.671     -624.671     -624.671    1     0.7  ok            SELECTED
  run39    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    11     -731.832     -621.832     -621.832    2     1.0  ok            
  run41    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    11     -731.364     -621.364     -621.364    3     1.2  ok            
  run54    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    11     -730.406     -620.406     -620.406    4     0.7  ok            
  run34    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    12     -736.912     -616.912     -616.912    5     0.7  ok            
  run52    LAGTIME=OFF;PERIPHERALS=1;CL-WT=none;CL-CRC…     8     -696.792     -616.792     -616.792    6     0.3  ok            
  run55    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    11     -725.824     -615.824     -615.824    7     0.6  ok            
  run56    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    11     -725.806     -615.806     -615.806    8     0.8  ok            
  run58    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    11     -724.961     -614.961     -614.961    9     0.8  ok            
  run57    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    11     -724.721     -614.721     -614.721   10     0.7  ok            
  run45    LAGTIME=OFF;PERIPHERALS=1;CL-WT=none;CL-CRC…     9     -703.828     -613.828     -613.828   11     0.4  ok            
  run66    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    12     -733.786     -613.786     -613.786   12     1.0  ok            
  run64    LAGTIME=OFF;PERIPHERALS=1;CL-WT=power;CL-CR…     9     -703.363     -613.363     -613.363   13     0.3  ok            
  run24    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    12     -732.978     -612.978     -612.978   14     0.8  ok            
  run62    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    12     -732.926     -612.926     -612.926   15     0.8  ok            
  run60    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    12     -732.689     -612.689     -612.689   16     1.0  ok            
  run4     LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    12     -732.081     -612.081     -612.081   17     1.1  ok            
  run49    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    12     -731.456     -611.456     -611.456   18     0.9  ok            
  run51    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    12     -731.449     -611.449     -611.449   19     0.8  ok            
  run68    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    12     -731.296     -611.296     -611.296   20     0.8  ok            
  run67    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    12     -730.385     -610.385     -610.385   21     0.7  ok            
  run37    LAGTIME=OFF;PERIPHERALS=1;CL-WT=linear;CL-C…    10     -709.388     -609.388     -609.388   22     0.5  ok            
  run28    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -738.742     -608.742     -608.742   23     1.0  ok            
  run61    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    13     -738.587     -608.587     -608.587   24     1.2  ok            
  run29    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    13     -738.374     -608.374     -608.374   25     1.0  ok            
  run26    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -738.260     -608.260     -608.260   26     0.8  ok            
  run35    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -737.628     -607.628     -607.628   27     1.0  ok            
  run44    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -737.618     -607.618     -607.618   28     1.1  ok            
  run25    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    12     -726.923     -606.923     -606.923   29     0.8  ok            
  run32    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    12     -726.390     -606.390     -606.390   30     1.2  ok            
  run33    LAGTIME=OFF;PERIPHERALS=1;CL-WT=power;CL-CR…    10     -705.032     -605.032     -605.032   31     0.6  ok            
  run30    LAGTIME=OFF;PERIPHERALS=1;CL-WT=none;CL-CRC…    10     -704.065     -604.065     -604.065   32     0.4  ok            
  run13    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    13     -733.264     -603.264     -603.264   33     0.9  ok            
  run5     LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -733.263     -603.263     -603.263   34     1.3  ok            
  run63    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    13     -733.142     -603.142     -603.142   35     0.9  ok            
  run6     LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -732.846     -602.846     -602.846   36     1.1  ok            
  run2     LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    14     -737.005     -597.005     -597.005   37     1.4  ok            
  run11    LAGTIME=OFF;PERIPHERALS=1;CL-WT=power;CL-CR…    11     -704.964     -594.964     -594.964   38     0.7  ok            
  run48    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    11     -703.113     -593.113     -593.113   39     0.7  ok            
  run7     LAGTIME=OFF;PERIPHERALS=1;CL-WT=power;CL-CR…    12     -711.173     -591.173     -591.173   40     0.7  ok            
  run14    LAGTIME=OFF;PERIPHERALS=1;CL-WT=linear;CL-C…    12     -711.116     -591.116     -591.116   41     0.6  ok            
  run12    LAGTIME=OFF;PERIPHERALS=1;CL-WT=linear;CL-C…    12     -710.999     -590.999     -590.999   42     0.7  ok            
  run19    LAGTIME=OFF;PERIPHERALS=1;CL-WT=linear;CL-C…    12     -710.553     -590.553     -590.553   43     0.5  ok            
  run42    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    12     -708.555     -588.555     -588.555   44     1.1  ok            
  run20    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    14     -710.151     -570.151     -570.151   45     0.9  ok            
  input    FO, 0 peripherals                                6     -172.233     -112.233     -112.233   46     0.1  ok            
  run10    LAGTIME=OFF;PERIPHERALS=0;CL-WT=none;CL-CRC…     7     -172.709     -102.709     -102.709   47     0.2  ok            
  run27    LAGTIME=OFF;PERIPHERALS=0;CL-WT=power;CL-CR…     8     -182.539     -102.539     -102.539   48     0.1  ok            
  run16    LAGTIME=OFF;PERIPHERALS=0;CL-WT=power;CL-CR…     9     -182.946      -92.946      -92.946   49     0.1  ok            
  run21    LAGTIME=OFF;PERIPHERALS=0;CL-WT=linear;CL-C…    10     -189.267      -89.267      -89.267   50     0.2  ok            
  run8     LAGTIME=ON;PERIPHERALS=0;CL-WT=none;CL-CRCL…    11     -184.416      -74.416      -74.416   51     1.2  ok            
  run1     LAGTIME=ON;PERIPHERALS=0;CL-WT=power;CL-CRC…    12     -191.030       28.970      128.970    -     1.1  excluded      parameter correlation |r| = 0.9681 exceeds 0.9500 (TVKA ~ TVALAG)
  run3     LAGTIME=ON;PERIPHERALS=0;CL-WT=linear;CL-CR…    10     -183.704       16.296      116.296    -     0.5  excluded      parameter correlation |r| = 0.9733 exceeds 0.9500 (TVKA ~ TVALAG)
  run9     LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    14     -709.870     -469.870     -369.870    -     1.2  excluded      parameter correlation |r| = 0.9922 exceeds 0.9500 (TVALAG ~ log_chol_ETA_ALAG)
  run15    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    13     -734.151     -504.151     -404.151    -     1.1  excluded      parameter correlation |r| = 0.9554 exceeds 0.9500 (TVALAG ~ log_chol_ETA_ALAG)
  run17    LAGTIME=ON;PERIPHERALS=0;CL-WT=power;CL-CRC…    10     -181.274       18.726      118.726    -     0.5  excluded      parameter correlation |r| = 0.9710 exceeds 0.9500 (TVKA ~ TVALAG)
  run18    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    12     -731.897     -511.897     -411.897    -     0.9  excluded      parameter correlation |r| = 0.9568 exceeds 0.9500 (TVALAG ~ log_chol_ETA_ALAG)
  run22    LAGTIME=ON;PERIPHERALS=0;CL-WT=none;CL-CRCL…     9     -174.220       15.780      115.780    -     0.3  excluded      parameter correlation |r| = 0.9767 exceeds 0.9500 (TVKA ~ TVALAG)
  run23    LAGTIME=ON;PERIPHERALS=0;CL-WT=none;CL-CRCL…    11     -184.378       25.622      125.622    -     0.4  excluded      parameter correlation |r| = 0.9656 exceeds 0.9500 (TVKA ~ TVALAG)
  run31    LAGTIME=ON;PERIPHERALS=0;CL-WT=power;CL-CRC…    11     -184.169       25.831      125.831    -     0.4  excluded      parameter correlation |r| = 0.9942 exceeds 0.9500 (TVKA ~ TVALAG)
  run36    LAGTIME=ON;PERIPHERALS=0;CL-WT=none;CL-CRCL…    10     -183.254       16.746      116.746    -     0.3  excluded      parameter correlation |r| = 0.9797 exceeds 0.9500 (TVKA ~ TVALAG)
  run38    LAGTIME=ON;PERIPHERALS=0;CL-WT=linear;CL-CR…    10     -186.987       13.013      113.013    -     0.3  excluded      parameter correlation |r| = 0.9751 exceeds 0.9500 (TVKA ~ TVALAG)
  run40    LAGTIME=ON;PERIPHERALS=1;CL-WT=power;CL-CRC…    12     -708.862     -388.862     -288.862    -     0.9  excluded      condition number 9.7272e3 exceeds 1.0000e3
  run43    LAGTIME=ON;PERIPHERALS=1;CL-WT=linear;CL-CR…    13     -739.527     -509.527     -409.527    -     1.1  excluded      parameter correlation |r| = 0.9580 exceeds 0.9500 (TVALAG ~ log_chol_ETA_ALAG)
  run46    LAGTIME=ON;PERIPHERALS=0;CL-WT=none;CL-CRCL…     9     -180.983        9.017      109.017    -     0.3  excluded      parameter correlation |r| = 0.9767 exceeds 0.9500 (TVKA ~ TVALAG)
  run50    LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL…    12     -734.016     -514.016     -414.016    -     1.1  excluded      parameter correlation |r| = 0.9569 exceeds 0.9500 (TVALAG ~ log_chol_ETA_ALAG)
  run53    LAGTIME=ON;PERIPHERALS=0;CL-WT=none;CL-CRCL…     8     -173.485        6.515      106.515    -     0.2  excluded      parameter correlation |r| = 0.9778 exceeds 0.9500 (TVKA ~ TVALAG)
  run59    LAGTIME=ON;PERIPHERALS=0;CL-WT=power;CL-CRC…    10     -183.896       16.104      116.104    -     0.3  excluded      parameter correlation |r| = 0.9749 exceeds 0.9500 (TVKA ~ TVALAG)
  run65    LAGTIME=ON;PERIPHERALS=0;CL-WT=power;CL-CRC…     9     -180.555        9.445      109.445    -     0.3  excluded      parameter correlation |r| = 0.9735 exceeds 0.9500 (TVKA ~ TVALAG)

  generation         best         mean polished  best
           0     -612.081     -374.505        0  run4
           1     -612.081     -446.689        0  run4
           2     -612.978     -437.291        0  run24
           3     -612.978     -458.708        0  run24
           4     -624.671     -489.922        2  run47

Final model: run47 — LAGTIME=ON;PERIPHERALS=1;CL-WT=none;CL-CRCL=none;V-WT=none;V-CRCL=none (fitness -624.671)
Final estimates written to <dir>/final-fit.yaml
Model table written to <dir>/models.csv; final model to <dir>/final.ferx
```

Worth reading in it: `run47` (10 parameters, OFV −724.7) beats `run39` (11 parameters, OFV −731.8) because the 7.2-point OFV gain is below pyDarwin's 10-per-parameter charge — the schedule doing what it says. Every `LAGTIME=ON` one-compartment point is excluded by the gate (`TVKA ~ TVALAG` correlation above 0.95) and shows its reason; the `fitness` column carries the `gate` charge on top of the penalized criterion for those rows, which is why they never win while still steering the population. The GA's best moved from −612.1 (generation 0) to −624.7 (generation 4, where the final downhill pass polished two individuals).

roninsightrx:
Review complete — changes requested.

1. **[P1] Keep mandatory covariates mandatory when a structural allele removes their parameter** (`crates/ferx-tools/src/globalsearch/mod.rs:752-764`). The missing-parameter branch currently only records a note and continues. For a space such as `ABSORPTION([INST,FO]); COVARIATE(KA, WT, pow)`, every INST candidate silently drops the mandatory KA effect and remains eligible to win. That makes the selected model fall outside the user-declared search space. These points should be treated as unbuildable/ineligible (or the combined space should be rejected), rather than fitted without the forced relation.

2. **[P1] Recover the representative fit when a cross-batch duplicate wins** (`crates/ferx-tools/src/globalsearch/mod.rs:1025-1028, 1242-1257`). Cross-batch duplicates are stored as `(model, None)`, but final selection consults `duplicate_of` only when the ID is absent from `store`; because the duplicate ID is present, a selected duplicate returns `final_fit = None`. I reproduced this with the INST/dead-KA-gene space and GA seed 1: `run3` (the zero-penalty duplicate of `run1`) was selected, `final-fit.yaml` was omitted, and `final.ferx` kept starting values even though `run1`'s fit was available. Either store the representative fit on the duplicate or fall back through `duplicate_of` when the selected entry's fit is `None`.

3. **[P2] Do not stop a valid downhill search after 64 improving moves** (`crates/ferx-tools/src/globalsearch/ga.rs:500-531`). The documented algorithm says it continues until no one-gene neighbor improves, but the hard-coded `1..=64` cap can return before a local optimum. A 65-axis binary additive landscape starting at all ones reproducibly stops at fitness 1 with one improving gene still available. Since every move is strict, the termination can be bounded from the finite space (or an explicit configurable safeguard can report that polishing was incomplete).

Verification: the existing `ferx-tools` library suite passes (582 tests), as do all 3 `globalsearch_end_to_end` tests and the current CI checks. I confirmed findings 2 and 3 with focused temporary regression tests and removed those local test edits afterward.

roninsightrx:
Thanks — all three are right. Fixed in 18f783fc, each with the regression that reproduces the finding:

1. **Forced covariate on a removed parameter** → the point is now **unbuildable**: `decode` returns a `problem` naming the effect and the missing parameter, so the row carries the crash value, is never submitted and never selectable (`mod.rs`, the `space.forced` loop). A `COVARIATE(...)` without `?` describes every model of the space, and a structure that has no `KA` has no such model. Test `a_forced_effect_on_a_parameter_the_structure_removed_makes_the_point_unbuildable` (INST/FO × forced `KA-WT`: the INST row is refused, only the FO point is fitted and it carries the relation). Docs note added under "The grid".

2. **Winning cross-batch duplicate** → selection now goes through `final_model_and_fit`, which falls back to the representative's fit when the winner's store entry is `(text, None)` (and to the representative's pair when the id is absent). Unit test on the helper covers the three shapes; `tests/globalsearch_end_to_end.rs::a_winning_duplicate_still_hands_back_its_fit` is your reproduction (INST/FO × `COVARIATE?(KA, WT, exp)`, GA seed 1, real evaluations) and asserts `final_fit` is present and matches the winner's OFV, with no "not in the journal cache" note. It failed before the fix, passes after.

3. **Downhill cap** → the round bound is now the size of the space (`space_size(alleles)`, `usize::MAX` on overflow), which is the true bound since every move is a strict improvement. Test `the_downhill_search_is_not_capped_short_of_a_local_optimum`: 65-axis binary additive landscape from all-ones descends to fitness 0 and every neighbour of the result is checked worse-or-equal. Doc comment says why the cap was wrong.

`ferx-tools --lib` 585 green, `globalsearch_end_to_end` 4/4, `tools/preflight.sh` green.

roninsightrx:
Correction to the note on finding 2 above: the end-to-end test I named there turned out **not** to reach the duplicate path on this machine — the real-fit GA trajectory follows the fitted OFVs, which differ in the last digits across platforms, so your seed-1 winner was a plain fit here and the test stayed green under the mutation. Replaced in 9142371f by a scripted-fitter unit test, `a_run_whose_winner_is_a_cross_batch_duplicate_hands_back_the_fit`, that walks seeds until the winner is a cross-batch duplicate of the dead-gene representative and asserts that shape was reached. Verified by mutation: with the `(text, None)` arm restored it fails at seed 1 with `run3 (duplicate of run1) came back without its fit` — your reproduction — and passes with the fix. Lib suite 586, preflight green.

roninsightrx:

Read every hunk of the non-test source (`search/penalty.rs`, `search/candidate.rs`, `search/config.rs`, `search/journal.rs`, `search/runner.rs`, `globalsearch/{mod,ga,report}.rs`, `globalsearch_cmd.rs`, the five tool call sites) plus the docs and the example. The reuse/re-score path, the manifest key, the fitness assembly and the `by_hash` / `by_genome` dedup all check out — the `wanted`/`todo`/`foreign` filters are consistent, `models` is built from `store` before `final_model_and_fit` drains it, and `load_foreign` mirrors `load_resumable`'s "filter retryable before dedup, first row wins" convention. `parameter_counts` is right: `n_params = held_mask.count(!held)` in `api/fit.rs:1994` and `bic_inputs` come off the same mask, so the `n_free() == n_parameters` branch is the normal one, not the fallback.

Six findings, one of which I'd fix before merge.

### 1. A covariate decode error aborts the whole run and throws away every fit — `globalsearch/mod.rs:793` (and `:756`)

`decode` records a failed **structural** edit as `Decoded.problem` (a crash-valued row, search continues), but a failed `AddCovariateRelation` is `?`-propagated:

```rust
model.apply(ModelEdit::AddCovariateRelation(effect.relation()))
    .map_err(|e| format!("adding {}: {e}", effect.label()))?;
```

That `Err` walks out through `Oracle::evaluate` → `ga::run` → `search` → `run_globalsearch`. `write_report` is only on the `Ok` path, so `models.csv`, `generations.csv` and `final.ferx` are never written and the whole run returns an error — after however many hours of fitting it had already done. (The per-batch journals survive, so a `--resume` is not lost, but the user gets nothing back from the run.)

Concrete trigger: a space that both forces and optionally offers the same `(parameter, covariate)` pair, e.g.

```
COVARIATE(CL, WT, pow)
COVARIATE?(CL, WT, [exp,lin])
```

`Space::build` (`mod.rs:373`) sends the first to `forced` and the second to an axis, with no collision check between the two lists. The forced relation is written first; the first genome that picks a non-`none` allele on that axis then hits `add_covariate_relation`'s "`[covariate_model]` already declares `CL ~ WT`" (`src/edit/apply.rs:375`). With `algorithm = "ga"` that may not be until generation 3.

Two fixes, and I'd take both: refuse the forced/optional pair collision in `Space::build` (where the `in_base` case is already refused with a note), and turn a decode edit failure into a `problem` row so *any* future per-candidate edit error costs one grid point rather than the run.

### 2. Runner warnings are re-appended once per batch — `globalsearch/mod.rs:1034`

```rust
self.notes.extend(report.warnings.iter().cloned());
```

No dedup, unlike `Evaluator::note()` right above it. Every warning that recurs per batch lands once per batch. The one that always recurs is `load_foreign`'s — a mistyped `--reuse-from` yields `no search directory found under '…' to reuse fits from` on the input step *and* every generation *and* every downhill round. At the defaults (`generations = 10`, `downhill_period = 5`, `final_downhill`) that is ~20+ identical lines under `Notes:` in `render_summary`. Route it through `note()`, or dedup `notes` before the result is built.

### 3. `warnings` dropped on the early-cancel return — `search/runner.rs:300`

`warnings` is now declared *before* the "cancelled before it started" return, which still hardcodes `warnings: Vec::new()`. Anything `load_foreign` just pushed — `not reusing fits from '<dir>': the dataset differs (there: … here: …)` — is silently discarded on that path, so a run cancelled in that window reuses nothing and says nothing about why. Pass `warnings` through.

### 4. `store` holds a full `FitResult` per evaluated model — `globalsearch/mod.rs:1064`

`fit_behind` clones the whole `FitResult` into `store` for every `Entry::Fit` row, duplicates included — exactly the clone the runner deliberately avoids ("cloning it per duplicate would cost more than the answer is worth", `runner.rs`). `store` lives for the whole search and only the winner's fit is ever read out. modelsearch has the same shape at ~20 candidates; `exhaustive` at the `max_models = 500` cap is 500 `FitResult`s, each carrying per-subject / per-observation vectors, resident at once. Worth keeping only the model text plus the best-so-far fit, or storing fits by *hash* so duplicates share one.

### 5. `[rank] cutoff` is silently ignored — `globalsearch/mod.rs:155`

`from_config` reads `config.rank.kind` and `config.rank.penalties` but never `cutoff`. covsearch and ruvsearch both *refuse* a `cutoff` they cannot honour ("a file that says … and gets an LRT would be lying about how its winner was chosen"); modelsearch/iivsearch/iovsearch honour it. globalsearch does neither, so `[rank] cutoff = 3.84` in a globalsearch file is a no-op with no diagnostic. Either apply it (as a margin the winner must beat the input by) or refuse it by name.

### 6. Example comment contradicts the default — `examples/amd_start_globalsearch.ferxsearch:33`

```
algorithm    = "ga"                   # or "exhaustive" (raise max_models to 324)
```

`DEFAULT_MAX_MODELS` is 500 and the grid is 324, so switching to `exhaustive` needs no `max_models` at all. The comment sends the reader to add a key that does nothing.

---

Nothing else flagged. Specifically checked and found fine: `Criterion` losing `Eq` (no remaining `Eq` consumer); `manifest_key`'s float spelling (`{}` on `f64` is stable and round-trippable); `same_fits` vs `same_scores` (a fitless row is only trusted under identical criterion **and** strictness, which is the right split); `elites.min(pop_size - 1)` and the `fill_distinct` stall top-up in `ga.rs` (both bounded, `pop_size >= 2` guaranteed by `validate` + `size >= 2`); the `unwrap_or(0)` winner fallback (the input row is always `rows[0]`); `covsearch`/`ruvsearch` still refusing `penalized` explicitly and `allometry` forcing `Criterion::Ofv`, so widening `RankType::criterion()` from `Err` to `Ok` does not leak a penalized ranking into an LRT tool. One nit not worth a section: `RankType::criterion()` now has no non-test callers and silently returns the *default* schedule — the doc comment warns, but it is a trap for the next caller; consider `pub(crate)` or deleting it.

roninsightrx:
All six taken, plus the nit, in 77fb21e3:

1. **Covariate decode error no longer aborts the run.** Both fixes: `Space::build` refuses a pair that is both `COVARIATE` and `COVARIATE?` by name (test `a_pair_that_is_both_forced_and_searched_is_refused_at_build`), and any edit-layer refusal inside `decode` — forced or optional — is now a crash-valued `problem` row rather than a `?` out of the search (test `a_covariate_edit_the_layer_refuses_costs_one_point_not_the_run`: a base that already declares `CL ~ WT`, handed to `build` as if it did not, so the "already declares" refusal is hit on real edits; the two refused points are rows, the two clean ones are fitted, the run returns). Docs updated.
2. **Runner warnings deduplicated** — routed through `Evaluator::note`; test with a scripted fitter that returns the same warning on every batch across four batches: one note.
3. **Early-cancel return carries `warnings`** — test `a_pre_run_cancel_keeps_what_the_reuse_directories_said`.
4. **No `FitResult` per evaluated model any more.** `store` is split into `models: HashMap<id, ModelText>` and `fits: HashMap<hash, (criterion, FitResult)>`, and `fits` is pruned after every batch to hashes whose criterion is ≤ the best fitness any eligible row has reached — a row's fitness is its hash's criterion plus a non-negative charge, so anything above that bound can never win. What stays resident is the best fit plus anything within the non-influential tie-break of it. Duplicates never store a fit (they share the representative's hash). `ModelRow` gained `hash` for this. The Evaluator test asserts three models evaluated → one fit resident, and the seed-walk duplicate test still passes (the winner's fit is found by hash).
5. **`[rank] cutoff` applied**, as a margin over the input's fitness: a candidate is selectable only when `input.fitness − fitness ≥ cutoff`, else the input is the final model with a note naming the cutoff; when the input fails the gate the cutoff cannot be applied and the notes say so; validated finite and ≥ 0 like the other tools. Test covers all three branches plus `from_config`. Docs row added.
6. **Example comment** fixed (`324 points, under the 500 cap`).

Nit: `RankType::criterion()` deleted; only `criterion_with` and `RankConfig::criterion` remain.

`ferx-tools --lib` 590, all 20 test binaries green, preflight green.

Issue #829, opened 2026-07-13
perf(sens): port the partial-Hessian mixed dual (DualMixed<NA,N>) to the analytic outer-gradient provider


TeunP:
**Optional performance / consistency cleanup — not a correctness fix, not required.** Split out from #810 (closed as fixed-by-#816), where it was parked as an "adjacent cleanup" orthogonal to that issue's correctness fix.

## Summary

The analytic outer-gradient provider (`subject_sensitivities` → `run_obs<const N>`, `sens/provider.rs`) runs `Dual2<N> = DualMixed<N, N>` over the differentiated PK params (`N = slots.len()`, right-sized per #367). It computes the full `N×N` Hessian, but FOCEI reads only the η-rows (the η–η and η–θ blocks) — the Hessian block among **IIV-free** params is dead work.

The **ODE provider already drops it** via `DualMixed<NA, N>` (`sens/ode_provider.rs::run_subject_mixed<NA, N>`, `MIXED_NA_CAP`; #445 / #410): it seeds the IIV-bearing params as the first `NA` axes so the trailing `(N − NA)²` block is neither stored nor computed. The analytic provider is the only path that never adopted it.

## Proposed change

- Add a mixed-order `run_obs`, seed the IIV-bearing params as the first `NA` axes, and dispatch `0 < na < N` (mirroring the ODE provider).
- `DualMixed<NA, N>` already `impl PkNum` (`sens/num.rs:185`), and its parity tests pin the retained entries to `Dual2` bit-for-bit — so this is **reuse, not new machinery, and exact** (no numerical change to any estimate or OFV).

## Magnitude

Per-op Hessian work `N² → na·N`; on the whole outer-gradient op (value + grad + hess) ≈ `1 + N + N²` → `1 + N + na·N`. For the `transit_multidose_cov` shape (`N = 4, na = 2`) that is ~1.6×/op on the analytic outer gradient (~8 %/eval; the outer gradient is ~20 % of per-eval provider time), growing with the IIV-free PK-param count.

**Not a speed necessity** — #812, which surfaced this, closed as a measurement artifact (transit fits are already fast). This purely brings the analytic provider in line with the ODE provider.

## Validation

[...]

PR #822, opened 2026-07-13
feat(estimation): analytic gradients for lagtime × {IOV, TV-cov, TIME}, SS × reset, wider axis caps [#486]

hiddevandebeek:
## Why

An audit of #486 against `main` (`9bdd90cd`) ranked the remaining FD cells by how often they actually bite an everyday popPK model. Three stood out, and this PR closes all of them plus a rider. It also fixes a covariance bug the audit surfaced.

The headline is **lagtime**. A closed-form model carrying an `ALAG`/`LAGTIME` fell back to finite differences the moment the subject *also* had IOV, a time-varying covariate, or a `TIME`-dependent parameter — `tvcov_analytical_supported` and `iov_analytical_supported` both bailed on `has_lagtime()`. Lagtime + IOV and lagtime + TV-covariates are bread-and-butter oral popPK (the shipped `warfarin_additive_eta` example has a lag), so this was plausibly the most-hit FD cell in the library — and FD costs one extra objective evaluation per parameter on top of the accuracy loss.

The gate comment — *"the elapsed-time saltation is not carried on this walk"* — reads like missing math, but it isn't. It was a **plumbing gap**: the walk already threads *dual* boundary positions (`dual_pos`), and its own module docs say a moving boundary needs **no explicit saltation** because the dual `dt` carries it through the closed-form flow to all orders. The dose *start* simply was never wired to move — literally `// No lagtime in IOV scope → zero dose lagtimes` with `dose_lagtimes = vec![0.0; …]`.

## What changed

**1. Lagtime on the event-driven walks (TV-cov / `TIME` / IOV).** Each dose *arrival* `τ = t + ALAG` is now a moving dual boundary, exactly as a modeled infusion *end* already was. The sub-intervals either side of the arrival pick up `±∂ALAG` in their dual lengths, and the closed-form flow yields the exact bolus saltation `−A·Φ(t−τ)·b` to all orders (including the `Dual2` Hessian) with **no saltation injection**. The ODE path must inject one only because RK45 needs an `f64` timeline; porting that machinery here would have been the wrong move.

Covers a lag that varies **per occasion** (κ on `ALAG` — each occasion's doses arrive at their own offset, which falls out of per-dose seeding for free) and a **fixed-duration infusion** under a lag (whose window end rides the arrival's jet). Fixed *and* estimated lags are the same work here, so they are not split.

Two deliberate declines remain, both per-subject:
[...]

PR #1301, opened 2026-09-09
perf(sens): serve [covariate_nn] time-varying inputs analytically on the TV-cov walk [closes #1300]

### LINKED ISSUE #1300: perf(sens): [covariate_nn] with a time-varying input routes subjects to reconverged-FD gradients (MAX_TVCOV_AXES counts NN weights, ~300× per eval)
## Summary

A `[covariate_nn]` (DCM) model whose network reads **any** time-varying covariate — or whose population contains even a few subjects where an NN input changes mid-record — sends those subjects to finite-difference gradients on **both** loops under FOCE/FOCEI. Measured on the vancomycin UVM dataset (100 subjects, 274 obs, 2-cpt IV, `CRCLf` time-varying): the DCM twin of the analytic covariate model costs **~5–10 s per objective evaluation vs 18 ms** for the analytic model, i.e. ~300× per eval, >1 h vs 1 s per fit. The estimates are correct; the cost makes DCMs unusable on exactly the data they are meant for.

Measured at `5696b1d6` with `cargo build --release -p ferx-cli --features ferx-core/nn`.

## Symptom

```
  52 thetas, 3 etas, 2 sigmas
  gradient: analytic (Dual2) ×29, FD ×71  [requested: auto]
Starting NLopt Lbfgs optimization (60 parameters)...
```

and in `FitResult.warnings`:

> 71 of 100 subjects use finite-difference inner gradients (outside the analytic provider's scope, e.g. steady-state + reset, time-varying covariates + LTBS, or modeled-duration doses); their results are correct but slower.

Freezing the covariate to its baseline value in the CSV (static input) gives `analytic ×94, FD ×6` — the 6 are subjects whose `FFM` (the other NN input) changes once mid-record — and the fit is *still* ~4 s/eval (~200×), because each FD subject costs a full central FD over all 60 packed parameters with an inner re-solve per perturbation.

| model (same Ω/σ structure, FOCEI) | thetas | gradient line | wall / fit | s per eval |
|---|---|---|---|---|
| analytic covariate model, TV `CRCLf` | 5 | analytic | 1.0 s (52 evals) | 0.018 |
| DCM 2→8→3, TV `CRCLf` | 52 | analytic ×29, FD ×71 | >3 h, still running | 5–10 |
| DCM 2→8→3, static `CRCLb` | 52 | analytic ×94, FD ×6 | 737 s (127 evals) | ~4–5 |
| same, `maxiter = 0` (one objective eval, no gradient) | 52 | — | 0.16 s | — |

The last row shows the objective itself is cheap; the gradient is the whole cost.

## Root cause

`tvcov_analytical_supported` (`src/sens/provider.rs:2662`) bounds the time-varying-covariate event walk on the **total** dual width:

```rust
    // Bound total axes to the dual-walk dispatch cap so the outer (`m_dim`) and inner
    // (`n_eta`) TV-cov tables both resolve — matched analytic scope, no fixed-EBE FD
    // inner split (#449 re-review #2).
    if model.n_theta + model.n_eta > MAX_TVCOV_AXES {   // provider.rs:2687, cap = 24 (line 861)
        return false;
    }
```

`model.n_theta` includes every auto-generated NN weight/bias theta, so any DCM beyond a toy (2→8→3 is already 52 + 3 = 55 axes) fails the gate. The gate is model-level, so it declines the TV-cov walk for both `subject_sensitivities_tvcov` (outer, `provider.rs:3066`) and `subject_eta_grad_tvcov_with_schedule` (inner, `provider.rs:3463`). Every subject that `subject_routes_to_event_walk` sends to the walk — any subject with a time-varying covariate, a mid-record covariate change, a reset, an oral infusion, … — then falls to:

- inner: `inner_optimizer::gradient_fd` (FD over η on every BFGS step);
- outer: `population_gradient_sens_mixed` → `subject_reconverged_fd_gradient` (`outer_optimizer.rs:3027`), a central FD over the **whole packed vector** (all 52 weights + Ω + σ) that re-runs `find_ebe` at every ±h — ~120 inner solves per subject per gradient, each of which is itself FD-inner.

A `sample` of the running process confirms it: the main-thread stack is `population_gradient_sens_mixed` → `subject_reconverged_fd_gradient` → `find_ebe` → `backtracking_line_search` / `gradient_fd` → `individual_nll_into_with_schedule` → `compute_predictions_with_tv_into_with_schedule` → `compute_event_pk_params_into` → `MlpMapper::forward` / `forward_cache`, with `libsystem_malloc`, `_platform_memmove`, `RandomState::hash_one<String>` and `nalgebra::from_row_slice_generic` among the top self-time frames — i.e. the network is re-evaluated per event per perturbation, rebuilding its weight matrices and hashing covariate names each time.

## Why the cap is not needed for NN weights

The weights only reach the PK parameters through the network's output vector `z` (one per event when an input is time-varying). Nothing about them needs a dual axis:

- The **inner** (η-only) walk seeds `n_eta` axes and lifts NN outputs as constants through `ModelNnGuard` (`provider.rs:1266`) — exactly what the IOV η-only route already does for DCMs (`iov_eta_only_derivs_dyn`, `provider.rs:1268`, and the `iov_analytical_eta_supported` relaxation that "lets a `[covariate_nn]` model off the finite-difference fallback"). The non-IOV TV-cov inner is declined only to keep inner and outer scope matched (#449 re-review #2), not because it cannot be computed.
- The **outer** θ-chain for the weights factors as `∂p/∂w = ∂p/∂z · ∂z/∂w`, with `∂z/∂w` an exact backprop Jacobian (`MlpMapper::jacobian_preactivation`). This is the same factorization `src/estimation/nn_theta_gradient.rs` already uses for the fixed-η gradient (SAEM / IMP / VI), where it cut a 141-weight DCM from 141 solves to `n_outputs`. The static (non-TV) closed-form path already avoids seeding weight axes via `lognormal_param_derivatives` (`provider.rs:1090`) — it is only the TV-cov walk that insists on `n_theta + n_eta` dual axes.

So the axis budget the walk actually needs is `n_base_theta + n_eta` (base thetas = the ones `[individual_parameters]` can reference), plus per-event chain-rule columns for the weights applied after the walk.

## Proposed fix

1. **Gate on the program's θ-axis count, not `model.n_theta`.** `tvcov_analytical_supported` should bound `prog.n_theta_axis() + n_eta` (the same quantity `iov_analytical_supported_core` distinguishes via `require_theta_axis`), so a DCM with a handful of base thetas passes.
2. **Outer TV-cov walk: seed base-θ and η axes only; add the weight columns by chain rule.** Per event `e`, the walk already evaluates the network at that event's covariate snapshot; keep `∂z^{(e)}/∂w` (backprop, `jacobian_preactivation`) and `∂p^{(e)}/∂z^{(e)}` (softplus/exp head derivative × the `pk_i = tv_i·exp(Ση)` chain), and accumulate `∂f/∂w = Σ_e ∂f/∂p^{(e)} · ∂p^{(e)}/∂z^{(e)} · ∂z^{(e)}/∂w`. `∂f/∂p^{(e)}` per event is what the dual walk produces on the base-θ axes when `p` is seeded per event; the mixed `∂²f/∂η∂w` block the FOCEI outer needs follows the same pattern on the `Dual2` cross terms. This is the TV-cov analogue of `nn_theta_gradient.rs`, but exact (no bias FD) because the per-event `z^{(e)}` are available inside the walk — the very case that module documents as out of its scope.
3. **Inner: admit the η-only TV-cov walk for NN models** regardless of (2), so at worst only the outer is FD. If (2) lands, inner and outer stay matched as #449 wanted.
4. **Cheap wins in the FD path itself**, independent of the above and worth doing anyway since some subjects will always be out of scope: `MlpMapper::forward` rebuilds `DMatrix` weight views from the theta slice and `build_input_vec_zero_fill` hashes `String` covariate names on every call — hoist per-subject/per-gradient (weights don't change inside an inner solve; input vectors per event don't change inside a gradient).

## Acceptance / tests (per CLAUDE.md)

- `Dual2`-vs-FD parity for the new TV-cov weight columns: extend `check_full_provider_vs_fd` (`sens/provider_tests.rs`) with a `[covariate_nn]` model whose **input is time-varying** and whose population has subjects with ≥ 2 distinct input values, asserting ∂NLL/∂w and the ∂²/∂η∂w block against central FD of the f64 predictor. A static-input fixture cannot see this (the static path already avoids the axes).
- Mutate each side: zero the per-event `∂z/∂w` term → parity must fail on the TV subject and stay green on a static subject; drop the base-θ seeding → fail on both. Name the side in the assertion message.
- Routing test: a 2→8→3 DCM with a TV input must report `gradient_method_inner`/`_outer` analytic for every subject the analytic model itself serves analytically (assert the `fd_fallback_warning` is absent); a model that genuinely exceeds the cap on `prog.n_theta_axis() + n_eta` still routes to FD loudly.
- Bit-identity where nothing changed: analytic (non-NN) models and static-input DCMs must produce byte-identical `fit.yaml` before/after (the change is scope, not numerics).
- Timing: on the repro below, per-eval cost should drop to the same order as the analytic model (target ≤ 10× the 18 ms, not 300×). Report the `gradient:` line and wall time in the PR.

## Repro

```sh
cargo build --release -p ferx-cli --features ferx-core/nn
# ~/git/ferx/ferx-testdata/vancomycin_uvm/ferx/dcm/  (models + data twin live there)
target/release/ferx dcm_2in_h8.ferx    --threads 4 --clean   # TV CRCLf: analytic ×29, FD ×71, ~5–10 s/eval
target/release/ferx dcm_2in_h8_bl.ferx --threads 4 --clean   # static CRCLb: analytic ×94, FD ×6, ~4 s/eval, 737 s
target/release/ferx full.ferx          --threads 4 --clean   # analytic covariate twin: 1.0 s
```

`dcm_2in_h8.ferx` is `examples/warfarin_dcm.ferx`-style: `inputs = [CRCLf, FFM]`, `center/scale` set, `outputs = [CL, V1, V2]`, `layers = [8]`, `tanh`/`softplus`, `init` set, `Q` a plain theta, block Ω on (CL, V1, V2), combined error.

## Related

- #449 (re-review #2) — the matched inner/outer TV-cov scope that introduced the total-axis bound.
- `nn_theta_gradient.rs` / the "Weight gradients" section of `docs/model-file/covariate-nn.qmd` — the same factorization for fixed-η estimators, which explicitly opts out of time-varying NN inputs and should be superseded by (2) for the FOCE family.
- `docs/model-file/covariate-nn.qmd` "Mu-ref composition with etas" already notes the inner fast path skips NN-anchored etas; that is a separate (smaller) cost.

https://claude.ai/code/session_01XpjD8CCgCRcwJb23bHLo2p

roninsightrx:
## Why

A `[covariate_nn]` (DCM) model whose network reads a **time-varying** covariate sent every such subject to reconverged finite-difference gradients on both FOCE/FOCEI loops (#1300). `tvcov_analytical_supported` bounded the event-driven walk on `model.n_theta + n_eta`, and the generated weight thetas alone exceed `MAX_TVCOV_AXES = 24`. On the vancomycin UVM DCM (2→8→3, 52 θ, `CRCLf` time-varying) that was `analytic ×29, FD ×71` and ~11 s per objective evaluation against 17.5 ms for the analytical covariate twin; a static-input twin was still `FD ×6` at ~11 s/eval because six subjects' `FFM` steps mid-record. The estimates were right; the cost made DCMs unusable on exactly the data they are meant for.

## What changed

The weights reach the PK parameters only through the network's output vector, so they never needed a dual axis of their own.

- **Evaluator** (`parser/model_parser.rs`): a `ModelNnAxisGuard` makes `Op::PushNnOutput` seed each network output on a dual axis (`base + flat index`), paired with the existing `ModelNnGuard` for the values. The zero-width cov-static value pass (`Dual2<0>` in `eval_cov_static_f64`) runs with seeding suspended; an out-of-range seed fails loudly through `var`'s bounds check rather than being lifted to a constant. `eval_param_duals` now seeds only the program's own θ axes (`m < n_theta`) — identical to the former `m < M` for every non-NN caller, and no longer puts generated weights on the axes η occupies.
- **Per-event builder** (`sens/provider.rs::nn_param_derivatives_at_cov`): program over `Dual2<n_base + n_eta + n_z>`; declared θ read off their axes; weight columns by chain rule, `∂p/∂w = ∂p/∂z · ∂z/∂w` and `∂²p/∂η∂w = ∂²p/∂η∂z · ∂z/∂w`, with `∂z/∂w` the **post-activation** backprop Jacobian at that event's covariates (`NamedMlpMapper::jacobian_raw`, `pub(crate)`). This is the exact, per-event form of the factorization `nn_theta_gradient.rs` applies with a bias FD — the time-varying-input case that module documents as out of scope.
- **Outer walk**: `subject_sensitivities_tvcov` hoists the per-event `ParamDerivs` (as the inner already did after #822) and seeds its θ columns in **chunks of `MAX_TVCOV_AXES − n_eta`**, each walked with the η axes alongside so the mixed block comes from the same jet; chunk results are scattered into one `SubjectSens` (`obs_sens_scatter_cols`). A non-NN model is exactly one chunk of width `n_theta + n_eta` — the former dispatch — so it is numerically unchanged. The modeled-dose and readout slot duals reuse the hoisted derivatives (`pk_slot_dual_outer_cols`) instead of re-running the program.
- **Inner walk**: a `[covariate_nn]` model takes an η-only builder (`Dual1<n_eta>` under `ModelNnGuard`, outputs as constants — exact for `∂/∂η`, the IOV η-only route's pattern); only `dp_deta` is populated, the other blocks are left *empty* rather than zero-filled so a stray consumer indexes out of bounds instead of reading a zero.
- **Gate**: bounds the *program* width (declared θ + η + network outputs, `tvcov_program_axes`), requires `prog.n_theta_axis() == n_program_theta(model)`, and declines a model that reads a generated weight θ directly (the parser's `NN_WEIGHT_DIRECT_REFERENCE_MARKER`, same rule as `NnGradPlan::build`). Non-NN scope is exactly as before.

Not changed (out of scope, unchanged behaviour): a DCM combined with `[initial_conditions]` or an `ExpressionScale` still routes to FD through `init_supported` / `scaling_supported`, which still key on `model.n_theta`.

Commit 2 hoists the two NN builders' `const_dispatch!` bodies into generic fns (llvm-cov attributes a 24-arm macro body as missed even when an arm runs — that was the 77% patch reading) and pins the zero-θ one-empty-chunk case. Commit 3 is issue item 4, described under Performance. Commit 4 answers the review's P1 (see Breaking changes). Rebased onto main after #829 / #1234.

## Alternatives considered

- **Seeding per-snapshot `z` axes instead of chunking weight columns** (`n_out × n_distinct_snapshots` axes): fewer walks when the network is wide but the covariate steps rarely, more bookkeeping, and it degrades with per-record covariates (the vancomycin case). Chunking weight columns is generic, bounded by the network size not the data, and reuses the existing walk verbatim.
- **Widening the dispatch tables past 24**: `Dual2<M>` is `O(M²)` per operation and the monomorphization cost is the crate's compile budget (#971); a 55-wide jet would also be ~5× the per-op cost of the chunks.
- **FD on the output biases inside the walk** (the `nn_theta_gradient` trick): the per-event `z^{(e)}` are already in hand, so the exact chain costs nothing extra and needs no FD step.

## Cross-repo dependency

## Breaking changes
- [x] None. An earlier revision added `PkNum::N_AXES`; review (P1) pointed out `PkNum` is a public extension point, so the constant was removed in favour of crate-private bookkeeping (`ModelNnAxisGuard::suspend` around the zero-width cov-static pass). `api/ferx-core-public-api.txt` equals main's; `NamedMlpMapper::jacobian_raw` is `pub(crate)`.

## Numerical validation

- **Objective unchanged**: first-eval OFV on both DCM repro models is bit-identical before/after (`1339.544950`).
- **Non-NN model bit-identical**: `full.ferx` (analytical covariate twin, `CRCLf` time-varying, 40 evals) produces a byte-identical `fit.yaml` before/after except the two wall-time fields.
- **Gradient parity** (Tier 1, `--features nn`, `src/sens/provider_nn_tests.rs`):
  - `nn_param_derivatives_match_fd_of_pk_param_fn` — the per-event builder against central FD of the f64 `pk_param_fn` over **every** θ column (weights included, `1e-6` relative), `∂p/∂η`, and the mixed `∂²p/∂η∂θ` block (worst realised error 2.9e-6 relative at a 1e-4 step; bound 1e-4).
  - `dcm_tv_input_outer_provider_matches_fd_of_production` — the outer provider against central FD of `compute_predictions_with_tv` via the shared `check_full_provider_vs_fd` harness, on a 2→8→3 DCM that **straddles the cap** (52 θ + 3 η > 24, three chunks) with a **multi-dose** (3 IV boluses, later doses land with residual drug) subject whose `CRCL` differs at every dose and observation record; asserts up front that the weight columns and the mixed η-weight block are live.
  - `dcm_static_input_on_event_walk_matches_fd_of_production` — same walk, network input static (a non-NN covariate forces the walk).
  - `dcm_tv_input_inner_eta_grad_matches_outer_and_fd` — inner η-gradient equals the outer jet's η block (1e-10) and FD.
- **Routing**: `dcm_tv_input_population_reports_analytic_and_no_fd_fallback` (whole population analytic, `fd_fallback_warning` absent, route summary all-analytic), `dcm_exceeding_program_axis_cap_declines_to_fd` (26 program axes declines; 24 admits), `dcm_direct_weight_reference_declines_to_fd`.
- **Mutation** (each applied alone, tests that die):

  | mutation | dies |
  |---|---|
  | zero `∂p/∂w` chain | builder parity, TV outer parity, static-input outer parity |
  | drop declared-θ seeding | same three |
  | zero mixed η-weight block | same three |
  | drop θ columns of later chunks | TV + static-input outer parity (columns ≥ 21) |
  | zero the inner η-only builder | inner parity |
  | pre-activation Jacobian instead of post-activation | builder parity, both outer parities |

- [ ] NONMEM comparison — not applicable: no numerical result changes; the DCM's objective is the same function, only the gradient route changed (pinned by the FD parities above).

## Performance

`cargo build --release -p ferx-cli --features ferx-core/nn`, `--threads 4`, `maxiter = 3`, vancomycin UVM (100 subjects, 274 obs). Machine was shared with two other long-running `ferx` processes during both sets of runs, so absolute numbers are inflated equally.

| model | gradient line before → after | s / eval before | s / eval after |
|---|---|---|---|
| analytic covariate twin (`full.ferx`, 5 θ) | analytic → analytic | 0.0175 | 0.0175 (bit-identical fit) |
| DCM 2→8→3, TV `CRCLf` (`dcm_2in_h8.ferx`) | analytic ×29, FD ×71 → **analytic ×100** | 11.4 (729 s / 64 evals, stopped) | **0.30** (51 s / 168 evals) |
| DCM 2→8→3, static `CRCLb` (`dcm_2in_h8_bl.ferx`) | analytic ×94, FD ×6 → **analytic ×100** | 11.0 (913 s / 83 evals) | **0.51** (27 s / 52 evals) |

~37× per eval on the TV model after the first commit, ~17× the analytic twin's per-eval cost. A 10-s `sample` showed the remainder was the width-24 `Dual2` jet traffic (`memmove`/`memset`/`malloc` + `DualMixed<24>::mul`) plus the NN forward pass (`weight_matrix` `DMatrix` rebuilds, `forward_cache`, `String` hashing in `build_input_vec_zero_fill`). Objective-only (`maxiter = 0`) is 0.11 s.

**Follow-up commit (issue item 4)** — `MlpMapper::forward_cache` and the backprop are plain slice loops now (no per-call `DMatrix`/`DVector` construction; accumulation in the same order as `nalgebra`'s column-wise `gemv`, so bit-identical — the old implementations stay as `#[cfg(test)]` references and `forward_cache_matches_nalgebra_reference` / `jacobian_matches_nalgebra_reference` pin bit-equality over five shapes × three activation pairs), and the chunked walk evaluates each event's f64 PK values once per subject instead of once per chunk (plus once more for the modeled-dose window and readout extras). Same three runs, same binary flags, `fit.yaml` bit-identical to the pre-follow-up run for all three models:

| model | s / eval after #1300 | s / eval after follow-up |
|---|---|---|
| analytic covariate twin | 0.0175 | 0.0155 (noise; unchanged code path) |
| DCM 2→8→3, TV `CRCLf` | 0.30 (51 s / 168 evals) | **0.19** (32 s / 168 evals) |
| DCM 2→8→3, static `CRCLb` | 0.51 (27 s / 52 evals) | **0.31** (16 s / 52 evals) |

Net: ~60× per eval on the TV model against the pre-PR baseline, ~11× the analytic twin — the issue's ≤10× target within noise. The `String` hashing in `build_input_vec_zero_fill` is left alone (2 lookups per forward pass; ~2% of samples).

## Tests
- [x] Unit tests added (`src/sens/provider_nn_tests.rs`, 7 tests; `tvcov_zero_theta_model_walks_one_empty_chunk` in `provider_tests.rs`; two bit-identity tests in `nn/mod.rs` — `cargo test --lib --features nn`: 4373 passed)
- [x] Regression tests fail without the fix (the routing tests; the parity tests are unreachable without it since the subject declined)

## Example
- [x] Not applicable — no new API surface; `examples/warfarin_dcm.ferx` is unchanged and unaffected (static inputs).

## Docs
- [x] `docs/model-file/covariate-nn.qmd` "Weight gradients" — FOCE/FOCEI paragraph
- [x] `CHANGELOG.md` `[Unreleased]` / Performance
- [x] `nn_theta_gradient.rs` module docs point at the walk for the TV case

## Checklist
- [x] `tools/preflight.sh` green (fmt, check matrix, clippy, rustdoc, docs, public-API)
- [x] `Dual2` path stays differentiable — the walk is untouched; only its seeding and scatter changed
- [ ] `Cargo.toml` version bump — not needed

## Docs & examples
- [x] No example execution step needed beyond the repro above (no new example; docs page edited only)

## Reviewer hints

- `run_obs_tvcov` is a mechanical transformation: read `pd_row_dual2_cols` / `obs_sens_scatter_cols` against the old `seed_row` / `obs_sens_from_dual2` — with `theta_cols == 0..n_theta` they are the same loops entry for entry, which is what the bit-identical `full.ferx` run pins.
- `eval_param_duals`'s `m < self.n_theta` bound: the one line that touches every program-driven path. For every caller admitted today `prog.n_theta == theta.len()`, so nothing changes; for a DCM it stops the weights being seeded onto η's axes (harmless before only because the program never `PushTheta`s a weight).
- Chunk order/η-block equality is a forward-mode identity (axes never interact); chunk 0's η blocks are kept.

## Open questions

None — issue item 4 landed in the follow-up commit (see Performance).

https://claude.ai/code/session_01XqX9znTnvCawHPMJ14MFex


roninsightrx (commented):
Found one blocking issue:

**[P1] Keep the NN axis count out of the required public `PkNum` contract** (`src/sens/num.rs:32`)

Adding `const N_AXES: usize;` without a default makes every downstream `PkNum` implementation fail to compile as soon as it resolves to this release. `PkNum` is an intentionally public extension point (the public ODE/analytical APIs accept caller-provided `T: PkNum`), while this constant is only used by the private bytecode evaluator to distinguish the `Dual2<0>` fold from the internally dispatched jets. That is both a SemVer-breaking change with `Cargo.toml` still at `0.3.1` and an internal-only widening of the core API boundary.

Please either keep this bookkeeping private (for example, suspend NN-axis seeding during the zero-width cov-static value pass) or make the new item source-compatible for existing implementations with a suitable default. If requiring all external implementations to provide it is intentional, this needs the repository's breaking-version bump and downstream coordination rather than a patch-level release.

Validation performed on the PR merged locally with current `origin/main` (the only merge conflict is the adjacent `CHANGELOG.md` entries):

- `cargo test --lib --features nn`: 4,389 passed, 23 ignored
- `tools/update-public-api.sh --check`: passed

The feature's new NN/provider parity and routing tests all pass, and I did not find another correctness issue in the chunking or per-event chain rule.

roninsightrx:
Re the P1 (`PkNum::N_AXES`): agreed, and fixed in 2b5a650f by taking the first option — the constant is gone from `PkNum` entirely. The only thing it existed for was letting the evaluator tell the zero-width cov-static value pass (`Dual2<0>` in `eval_cov_static_f64`) apart from a genuine out-of-range seed. That pass now suspends NN-axis seeding for its own scope (`ModelNnAxisGuard::suspend`, crate-private thread-local bookkeeping), and an out-of-range seed still fails loudly through `var`'s bounds check instead of being lifted to a constant. `api/ferx-core-public-api.txt` is back to main's — the PR adds no public item now, so no version bump is needed.

Also rebased onto current main (the `CHANGELOG.md` conflict you mentioned is resolved; main's #829 partial-Hessian change touches the same provider and rebased without conflict). On the rebased tree: `cargo test --lib --features nn` 4389 passed, `tools/preflight.sh` green.

PR #1305, opened 2026-09-09
feat(estimation): apply covariate-NN (DCM) regularization under VI

dominic-irx:
## Why

Fitting a deep compartment model (`[covariate_nn]`) with `method = vi` was **silently
unregularized**: `nn_l2` / `nn_smooth` were applied by the FOCE-family methods only, and a
`vi` final stage warned "the fit is unregularized" and ignored them. That left VI DCM fits
with no way to damp the spurious covariate-driven `Ω` collapse the [VI DCM
docs](docs/estimation/vi.qmd#dcm-omega) describe, and — because an unregularized network
drifts along its exact permutation/layer-scale weight symmetries — such fits never settled
and always burned the full `vi_iters` ceiling.

## What changed

- **VI applies the covariate-NN penalty.** `run_vi` builds the same `NnRegularizer` the
  outer optimizers use and folds its gradient into the Adam step (`fold_nn_penalty`). The
  penalty is defined on the −2LL scale (FOCE minimizes `2·nll + penalty`); VI's `grad_x` is
  on the 1× `−ELBO` scale, so it enters at **½** — this is what makes a given `nn_l2` mean
  the same thing under VI as under FOCEI. `applies_nn_regularization` now includes `Vi`, so
  the "unregularized" warning no longer fires.
- **Convergence is judged on the penalized objective, the reported bound stays clean.** A
  second per-iteration trace, `obj_trace = −2·ELBO + penalty`, feeds every stopping
  predicate (`trace_has_settled` / `trace_still_drifting`); the reported `vi.elbo_trace` and
  OFV stay the penalty-free bound — the split FOCE uses when it reports the clean OFV while
  ranking on the penalized objective. Without this the clean bound and the minimized
  objective drift apart under regularization and the run never early-stops. Exposed as the
  new `ViResult::objective_trace` (equal to `elbo_trace` when no regularization is active).

## Alternatives considered

A first cut folded the penalty into the gradient but left convergence judged on the clean
`−2·ELBO`, on the theory that the parameter-stability criterion (`estimates_have_settled`)
would fire once regularization pinned the weights. It did not: the network weights `x`
settle, but the per-subject variational parameters `φ` need not reach the parameter
tolerance on weakly-identified DCM data, so that criterion kept resetting while the clean
objective (no longer the minimized quantity) kept drifting. Tracking the penalized objective
directly is the fix.

## Cross-repo dependency

The new `ViResult::objective_trace` field is additive and `#[serde(default)]`, so ferx-r
deserializes existing and new fit results unchanged. A ferx-r `NEWS.md` note surfacing the
new field to R users is optional follow-up, not a merge blocker.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — regenerated `api/ferx-core-public-api.txt`
- [ ] None

Additive only: one new public field, `ViResult::objective_trace: Vec<f64>`. Needed so a
caller (and the YAML/JSON output) can see the objective VI actually minimized under
regularization — the reported ELBO bound cannot show why a regularized fit converged or
stalled. `#[serde(default)]` keeps older serialized results loadable.

## Numerical validation

No NONMEM anchor exists for VI covariate-NN regularization (the penalty is validated on the
FOCE path; NN models are an experimental feature validated on toy examples). Validation here
is the before/after convergence behaviour on a real infliximab 2-cpt DCM
(`min_h8_min_tanh`, 8-unit tanh network), `--clean`, 8 threads, default `vi_iters = 25000`:

```
nn_l2   converged  iters   ofv      tightness
0       True       15125   608.07   0.471     (baseline; penalty inactive => unchanged)
1e-2    False      25000   599.17   0.468     (still descending at the ceiling — see below)
1e-1    True        5375   630.17   0.469     (was 25000/False before the fix)
```

`nn_l2 = 1e-1` is the headline: the fix takes it from the ceiling to a genuine early stop.
`nn_l2 = 0` is byte-identical to the prior behaviour by construction (`objective_trace ==
elbo_trace`). `nn_l2 = 1e-2` still reaches the ceiling, and that is **correct**: its
penalized objective is still monotonically descending there (~2.5 over the last 10k
iterations, 7/9 downward block-steps), so `converged = False` is honest — an intermediate λ
only partially damps the symmetry drift. The remedy is a stronger λ or more iterations, and
the run now reports that instead of the old silent ceiling burn.

- [x] Compared against: prior ferx behaviour (same binary, `nn_l2 = 0`) and cross-λ sweep
- [x] Reference values: table above

## Performance
- [x] Faster — a regularized VI DCM early-stops instead of burning the `vi_iters` ceiling
  (infliximab `nn_l2 = 1e-1`: 25000 → 5375 iterations). Unregularized fits are unaffected.

## Tests
- [ ] No new tests needed

`fold_nn_penalty_*` pin the ½ scale, the index mapping, and the inactive no-op;
`vi_objective_trace_is_the_penalized_objective` pins that the iter-0 gap between
`objective_trace` and `elbo_trace` equals the penalty at the known start point exactly (the
direct-add scale) and that the objective sits above the bound throughout;
`vi_objective_trace_equals_the_clean_trace_when_unregularized` pins the λ=0 guarantee;
`vi_nn_l2_shrinks_the_network_weights` is the end-to-end proof the loop applies the penalty
(dies if the fold call is removed — verified by mutation, each twin side separately).

- [x] Not applicable (no new DSL/output *format*; `nn_l2`/`nn_smooth` already documented,
  now honoured under `vi`)

## Docs
- [x] `docs/` updated — `model-file/fit-options.qmd`, `model-file/covariate-nn.qmd`,
  `estimation/vi.qmd` (corrected the stale "regularization … is not yet implemented" note)
- [ ] New pages linked in `docs/_quarto.yml` (no new pages)
- [ ] No user-visible change

## Checklist
      (`--all-targets`), rustdoc, docs-lint, public-API baseline
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, this does not
      touch `sens/` or any `*_g<T: PkNum>`; the penalty gradient is the existing
      `NnRegularizer`, already parity-tested
- [ ] `Cargo.toml` version bumped if breaking (additive, no bump)

## Reviewer hints

- The one subtle number is the **½** in `fold_nn_penalty` (grad_x is 1× `−ELBO`) versus the
  **direct add** in `obj_trace.push` (both on the −2LL scale). `vi_objective_trace_is_the_
  penalized_objective` pins the latter to the exact penalty; the fold unit tests pin the
  former.
- `obj_trace` and `trace` are pushed once per iteration and stay equal length; the three
  predicate call sites (`run.rs` ~848, ~1001, ~1012) all read `obj_trace`.

## Open questions

- **`nn_l2 = 1e-2` still hitting the ceiling** — I read this as correct (the objective is
  measurably still descending), not a residual detection bug. Happy to run it at
  `vi_iters = 50000` to confirm it settles on the penalized objective if a reviewer wants
  that reassurance.
- The `CHANGELOG.md` entry has no issue/PR number yet — will add once this PR number exists.


roninsightrx (changes requested):
I found three issues that should be addressed before merge:

1. **[P2] Treat the new public struct field as a breaking Rust API change** (`src/types.rs:5214`). Adding a public field to an exhaustively constructible public struct breaks downstream struct literals and exhaustive patterns. The repository SDLC explicitly requires a pre-1.0 MINOR bump plus migration notes for a breaking public-Rust-API change, but this PR describes the field as additive and leaves the workspace at 0.3.1. Please either version/document this as breaking (including the workspace dependency versions), or use a compatibility-preserving API design.

2. **[P2] Preserve the objective trace semantics when loading legacy results** (`src/types.rs:5211-5214`, `src/io/output.rs:2414`). `#[serde(default)]` makes a pre-#1305 result deserialize with `objective_trace = []`, even though its existing `elbo_trace` is the effective objective trace. That contradicts the documented “identical element for element” invariant, and `write_estimates_yaml` then sees the empty vector as different and emits an empty `objective_trace:` (YAML null). Please migrate a missing field to the existing `elbo_trace`, or explicitly handle/document the legacy-empty sentinel and avoid emitting a null block. A legacy-deserialization regression test would pin this.

3. **[P2] Report the trace actually used by the convergence checks** (`src/estimation/vi/run.rs:1016-1054`, also `docs/model-file/fit-options.qmd:394`). Both post-run predicates now inspect `obj_trace`, but the noise-floor warning says the *ELBO trace* is still falling and the iteration-budget warning/docs direct users to `vi.elbo_trace`. With regularization, the motivating premise of this PR is precisely that the clean ELBO and minimized objective can drift apart, so these diagnostics can tell users to inspect a trace that does not support the convergence verdict. Please refer to `objective_trace` when regularization is active (and update the `ViResult::converged` docs, which still say it is judged on the ELBO), or use wording that accurately covers both cases.

Focused verification passed:

- `cargo test --lib fold_nn_penalty --features nn`
- `cargo test --lib vi_objective_trace --features nn`
- `cargo test --lib vi_nn_l2_shrinks_the_network_weights --features nn`
- `cargo check --no-default-features`
- `git diff --check`

dominic-irx:
Thanks for the review — all three P2s addressed in 7299df98.

**[P2] Breaking public-Rust-API change.** `ViResult` is now `#[non_exhaustive]` (mirroring the `NewParameter` precedent already in this Unreleased block), so this and future field additions no longer break downstream struct literals / exhaustive matches — fields stay public to read, which is all ferx-r and the YAML/`FitResult` consumers do. Bumped the workspace to **0.4.0** (root `Cargo.toml`, both members, the two `ferx-core` dep constraints, `Cargo.lock`) and regenerated the public-API baseline; CHANGELOG now documents it as breaking with migration notes.

**[P2] Legacy-result trace semantics.** The empty vector is now the canonical **"identical to `elbo_trace`" sentinel**: an unregularized fit stores empty rather than a duplicate, `#[serde(default)]` yields the same for pre-#1305 results, and `write_estimates_yaml` emits the block only when it is non-empty — no more `objective_trace:` null block. A consumer reads `objective_trace` when non-empty and falls back to `elbo_trace` otherwise; the field docs spell this out. Pinned by two new tests: a legacy-deserialization round-trip (drop the key → deserializes to empty, not a failure) and an emission test (empty → omitted, populated → real sequence).

**[P2] Convergence diagnostics named the wrong trace.** The noise-floor and iteration-budget warnings now name `vi.objective_trace` — the trace the predicates actually inspect — when regularization is active, and `vi.elbo_trace` otherwise. Updated the `ViResult::converged` docs (which now say the verdict is on the penalized objective under regularization), plus `fit-options.qmd` and `vi.qmd`.

Preflight (fmt/check/clippy/rustdoc/docs/public-api) and the new tests are green.

Issue #251, opened 2026-06-11
feat(estimation): Add adaptive Gaussian quadrature (AGQ) estimation method


Robterheine:
## Summary

Add AGQ as an estimation method alongside FOCE/FOCEI/SAEM. AGQ generalizes the
Laplace approximation: instead of approximating each subject's marginal
likelihood with a single Gaussian at the empirical Bayes mode, it evaluates the
integrand on an adaptive Gauss-Hermite grid centered at that mode. With one node
it reduces exactly to Laplace; with more nodes the approximation improves and
makes no Gaussian-residual assumption.

## Motivation

- FOCE/FOCEI can't handle non-Gaussian observation models. AGQ drops the
  Gaussian-residual assumption, so it covers count, ordinal/categorical, and
  time-to-event endpoints with the same engine.
- It's deterministic. No Monte Carlo noise in the objective, so the outer
  optimizer converges in roughly Laplace-like iteration counts with no
  stabilization tricks (no common random numbers, no step-size schedule).
- It reuses what's already here. The expensive part, the per-subject inner
  mode-finding for the EB estimate and the Laplace Hessian, is identical to the
  current FOCE/Laplace path. AGQ adds a fixed grid evaluation on top, nothing
  more.
- In low dimension it's cheap, and ferx's analytical solutions make each grid
  point a closed-form evaluation rather than an ODE solve.

## Proposed approach

Per subject i, with conditional log-likelihood l_i(eta) = log p(y_i|eta) +
log phi(eta; 0, Omega), let eta* be the mode and H = -Hessian(l_i) at eta*.
Take the Cholesky factor L of H^-1 and use the standard adaptive Gauss-Hermite
transform eta = eta* + sqrt(2) * L * z, giving

    L_i ~= 2^(d/2) * |L| * sum_j [ prod_k w_jk * exp(z_jk^2) ] * exp(l_i(eta*+sqrt(2) L z_j))

where z_j are tensor-product GH nodes and w the GH weights. NAGQ=1 (single node
at z=0) recovers Laplace. Objective is -2 * sum_i log L_i.

Implementation:
- Register an `AGQ { n_nodes }` variant in the estimation-method dispatch next
  to FOCE/SAEM, exposed through ferx-r.
- Reuse the existing FOCE/Laplace inner solver to obtain eta* and H per subject.
  AGQ does not re-optimize; it lays the grid around the converged mode.
[...]

PR #1308, opened 2026-09-09
perf(estimation): analytic Gauss-Newton grid response for FOCEI-AGQ; robustness for Laplace's exact anchor [#251]

hiddevandebeek:
## Why

AGQ/Laplace's outer gradient has a grid-response term (`½·d log|H|/dx`, and its multi-node
generalisation) that was always finite-differenced — rebuilding the anchor Hessian at
`x ± h` for every free packed coordinate, `2·n_free` provider evaluations per subject per
gradient, with the anchor itself sometimes an ODE re-solve.

Motivated by inspecting nlmixr2est's own AGQ gradient (`src/foceiGrad.cpp::foceiGradSubjectAgqFR_`):
it anchors exclusively on the Gauss-Newton Hessian, which is bilinear in first-order
sensitivities only, so its derivative needs no third order at all — just the second-order
mixed blocks FOCEI's own outer gradient already computes. ferx additionally supports an
exact-Hessian anchor for genuine Laplace/NONMEM-LAPLACIAN-INTER semantics (nlmixr2est has no
equivalent), which *does* need third order — a materially different case.

## What changed

Two new modules, one per `HessianAnchor`:

- **`estimation/focei_htilde_dx.rs`** — analytic `dH̃/dx` for the Gauss-Newton anchor
  (`focei, n_agq > 1`), from the plain (second-order) sensitivity provider. **On by default**
  (`agq::use_analytic_grid_response`): costs one extra ordinary `subject_sensitivities` call,
  independent of `n_free`, replacing the FD route's `2·n_free` anchor rebuilds.

- **`estimation/laplace_h_deriv.rs`** — analytic `dH/dx` for the exact anchor (`laplace`),
  reusing the analytical-covariance provider's third-order FD sweep
  (`subject_sensitivities_cov`, #436). Implemented and tested across six model variants
  (diagonal/block Ω, combined error, single-η, two different θ/η points), but **not the
  default**: the measured wall-clock benchmark showed no repeatable win (fewer provider
  calls, but more provider time per call — each evaluation does materially more work).
  Available via `FERX_AGQ_GRID_RESPONSE=analytic`. Its value today is an exact, FD-noise-free
  gradient, not raw speed.

Both routes are pinned bit-consistent against the pre-existing finite-difference gradient
(dedicated FD-parity tests, plus the existing
`tests/agq.rs::analytic_gradient_matches_fd_at_every_node_count`), so no estimator's
numerical output changes.

## Alternatives considered

Originally scoped as FD-of-second-order-sensitivity-jet only, for the Laplace exact anchor
(see `plans/laplace-sensitivity-speed.qmd`). That call-count argument (`2p` vs
`1+2(n_theta+d)`) does not predict wall-clock cost — measured, it was a wash — so it stays
opt-in rather than becoming the default. The Gauss-Newton module was added after that
negative result, once inspecting nlmixr2est showed a route with no such call-count-vs-cost
ambiguity at all.

## Cross-repo dependency

No public API changed (both new modules are `pub(crate)`).

## Breaking changes

- [x] None

## Numerical validation

- [x] Compared against: prior ferx commit `cfc956da` (pre-change baseline, same worktree)
- [x] Reference values (OFV, both routes agree with the pre-existing FD gradient on every
      benchmarked fixture — see the Performance section below for the full table)

```
# warfarin_focei_agq.ferx (diagonal Ω, focei n_agq=3): OFV = -285.9739 (fd == analytic)
# warfarin_block_omega_focei_agq.ferx (3x3 block Ω):    OFV = -288.9155 (fd == analytic)
# warfarin_ode_focei_agq.ferx (ODE):                     OFV = -285.9697 (fd == analytic)
# warfarin_laplace.ferx (diagonal Ω, laplace n_agq=1):   OFV = -285.9702 (fd == analytic)
# warfarin_block_omega_laplace.ferx (3x3 block Ω):       OFV = -288.9119 (fd == analytic)
```

## Performance

- [x] Faster — Gauss-Newton anchor (`focei, n_agq > 1`), benchmarked on
      `plans/laplace-sensitivity-speed/` fixtures, 5 reps each, `ci-test` profile,
      `FERX_AGQ_GRID_RESPONSE=fd` vs `=analytic`:

  | Fixture       | Provider calls (fd → analytic) | Provider time | Outer iters |
  |---------------|----------------------------------|----------------|-------------|
  | diagonal Ω    | 16720 → 11780 (−30%)             | comparable     | 37 → 37     |
  | block Ω (3×3) | 21000 → 13020 (−38%)             | −38%           | 41 → 41     |
  | ODE           | 23770 → 11790 (−50%)             | −52%           | 53 → 37     |

  The ODE case is the clearest win: FD's perturbed-anchor rebuild re-solves the ODE
  sensitivity system `2·n_free` times per gradient; the analytic route pays one extra
  ODE-sensitivity provider call. It also converged in fewer outer iterations (37 vs 53).

- Exact anchor (`laplace`): no significant impact — implemented for correctness/robustness
  (an exact, FD-noise-free gradient), kept opt-in, see `plans/laplace-sensitivity-speed.qmd`
  for the negative wall-clock result.

## Tests

`estimation::focei_htilde_dx::tests` (6): FD-parity (base, M3-censored, custom magnitude,
ODE), exact symmetry, `iiv_on_ruv` decline.

`estimation::laplace_h_deriv::tests` (10): FD-parity across diagonal-Ω, block-Ω,
combined-error, single-η, and a second θ/η point; exact symmetry; censored/ODE/
magnitude/layout-mismatch declines.

`estimation::agq::tests`: new `analytic_grid_response_matches_the_fd_route` (direct A/B at
`n_agq ∈ {1, 3}`) and `out_of_scope_models_decline_to_the_fd_grid_response`;
`audit_invalid_perturbed_anchor_declines_the_gradient` updated to drive `fd_grid_response`
directly, since its original FD-specific invalid-perturbed-anchor precondition no longer
holds once `GaussNewton`'s grid response defaults to the analytic route (which never
evaluates at the invalid perturbed point at all).

`tests/agq.rs` (33, all still green): `gauss_newton_anchored_gradient_matches_fd` and
`focei_n_agq_gt_1_is_the_gauss_newton_anchored_quadrature` exercise the new default
end-to-end.

Full suite: `cargo test --lib` 4282/4282, `cargo test --test agq` 33/33.

- [x] Not applicable (internal / refactor / fix with no new API surface — no `[fit_options]`
      or DSL change; `FERX_AGQ_GRID_RESPONSE` is an internal benchmarking env var, not a
      documented user option)

## Docs

- [x] No user-visible change

## Checklist

      (`--all-targets`), public-API baseline (confirmed content-identical; the local
      `--check` failure on this Windows checkout is a CRLF-vs-LF line-ending artifact, not
      a content diff — verified with a manual LF-normalized comparison)
- [x] Functions on the `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay
      differentiable — this PR does not touch that path at all (it consumes the existing
      analytic sensitivity provider, doesn't change it)
- [x] `Cargo.toml` version bumped if breaking — not breaking, no bump

## Docs & examples

## Reviewer hints

The two new modules are close siblings — `laplace_h_deriv.rs` is a superset in math (needs
`Ψ`/third order that `focei_htilde_dx.rs` doesn't), and `focei_htilde_dx.rs` is a superset in
scope (ODE, magnitude, M3-base, FREM all admitted; `laplace_h_deriv.rs` declines them for
now). Worth comparing side by side. The `dp_dv` closed form in `focei_htilde_dx.rs` is the
one piece of genuinely new algebra there; it's the same formula as `sens_outer_gradient`'s
existing inline `theta_block`/`sigma_block` `dp` expressions, factored out rather than
copied a third time.

The `agq::tests::audit_invalid_perturbed_anchor_declines_the_gradient` diff is the most
subtle test change — worth reading the new comment explaining why its final assertion
(reconverge-interval independence) no longer applies now that the fixture it drove no
longer fails at all.

## Open questions

Whether the Laplace/exact-anchor route is worth widening (ODE, IOV, `iiv_on_ruv`) given it
has no proven speed benefit yet — deferred to a follow-up per `plans/laplace-sensitivity-speed.qmd`.


hiddevandebeek:
## Code review — analytic Gauss-Newton grid response (#1308)

I read the full diff and traced the assembly against the machinery it reuses. **No correctness defect was found in the derivative itself.** The following were checked and are right:

- `dp_dv(r, d, rv, dv)` reproduces `err_terms`' `beta` exactly when `(rv, dv) = (d, d2)`, and matches the inline `dp` in `sens_outer_gradient::theta_block` (`:1367`), `sigma_block` (`:1717`) and the `block_sigma` arm (`:1817`).
- The Omega-coordinate term. `Omega = L Lᵀ` implies `dOmega = e_row vᵀ + v e_rowᵀ` with `v = L[:,col]`, hence `dOmega⁻¹ = −(u pᵀ + p uᵀ)`; the `chain = l[(row,row)]` on the diagonal matches `pack_params`' `ln(L_ii)` packing, and the identity chain off-diagonal.
- Sigma packs as `ln σ`, so `dsigma = σ` is correct, and the natural-scale `(∂R/∂σ, ∂d/∂σ)` central difference mirrors `sigma_block`, including the FREM `EPSCOV²` branch (no `ruv_scale` lift) and the `sigma_fd_step` policy.
- Tensor layouts (`d2f_deta2[i·n_eta+r]`, `d2f_deta_dtheta[i·n_theta+m]`, `d3f_deta3[(i·n_eta+m)·n_eta+r]`, `d3f_deta2_dtheta[(i·n_eta+m)·n_theta+c]`) agree with `sens/provider.rs`.
- `analytic_grid_response` is a faithful mirror of `fd_grid_response`'s node-gradient arm: `½·tr(S⁻¹S_k)` against `½·Δlog_det_inv_scale/2h`, and `db̂/dx + √2·M_k·z_j` against the differenced `apply_l_sigma`. `RegularisedAnchor`'s jitter (`JITTER_REL`/`JITTER_FLOOR` = `1e-6`/`1e-10`) is bit-identical to `build_proposal`'s, and `node_scale() = l_inv.transpose() = L⁻ᵀ` is the same scale `apply_l_sigma` back-solves — so the two routes differentiate the same `S`. `regularised_anchor` declining on a non-PD `S` correctly avoids differentiating the broad-Omega proposal that `build_proposal` would have substituted.
- The scope gates are sound: the packed-length assertion excludes the IOV / mixture / rho segments, and the recently landed partial-Hessian provider (cfc956da) drops only the IIV-free Hessian block, which neither module reads.
- `acc_all` correctly preserves the all-or-nothing contract on a late non-finite coordinate.

Six observations follow, none blocking.

---

### 1. `CHANGELOG.md:28` — the "unchanged / bit-consistent" claim is contradicted by the diff (medium)

> Estimates, OFVs and standard errors are unchanged (the two routes are pinned bit-consistent, and this is on by default)

The two routes are not bit-consistent. `agq.rs:2677` asserts agreement only to `1e-5` relative, and the test's own documentation states that "the analytic route is the more accurate of the two here" — that is, they differ by construction. More directly, the PR's own benchmark table reports outer iterations going `53 -> 37` on the ODE fixture: a different optimizer trajectory, so the converged estimates and standard errors differ at the convergence tolerance.

Scenario: a user upgrading and re-running a `focei, n_agq > 1` fit sees estimates and SEs move in the last digits, and a downstream regression suite pinned to exact values goes red, against a CHANGELOG entry that explicitly promised no change. Suggested rewording: "the analytic and finite-difference routes agree to within the finite-difference route's own truncation error; converged estimates may move within the convergence tolerance".

### 2. `src/estimation/focei_htilde_dx.rs:124` — the anchor's provider call and `score_core` are recomputed (medium, efficiency)

`agq_subject_packed_gradient` calls `anchor_hessian`, which for `GaussNewton` runs `score_core_at` — precisely `subject_sensitivities(model, subject, &params.theta, b_hat)` followed by `score_core(...)` at the same `(params, b_hat)` — and keeps only `.htilde`. `subject_htilde_dx:124`/`:129` then repeats both calls with identical arguments, once per subject per gradient evaluation.

This PR already threads `&h` from `agq_subject_packed_gradient` through `grid_response_correction` into `analytic_grid_response`; threading the originating `ScoreCore` (and its `SubjectSens`) alongside it is a direct extension of the plumbing change already made. As it stands, the advertised cost — "one extra ordinary `subject_sensitivities` call" — is one extra provider call *plus* a redundant `score_core`.

The same applies to the `Exact` arm at `laplace_h_deriv.rs:192`: the `subject_sensitivities_cov` jet at `:187` is genuinely different work, but the `score_core` call duplicates the anchor's.

### 3. `src/estimation/agq.rs:2556` — the default-on route has no engage-and-compare test (medium-low)

`analytic_grid_response_matches_the_fd_route` drives `HessianAnchor::Exact` only — the anchor that is *opt-in*. The on-by-default `GaussNewton` route has no test asserting that `analytic_grid_response` returns `Some` at all.

The test's own documentation makes the case for why this matters: the end-to-end `tests/agq.rs::analytic_gradient_matches_fd_at_every_node_count` "passes identically whether the analytic route runs or silently declines to the FD route it is compared against". That argument applies verbatim to `GaussNewton`, and `gauss_newton_anchored_gradient_matches_fd` is subject to it too.

Scenario: a later scope clause added to `subject_htilde_dx`, or a tightening of `is_well_conditioned`, makes the route decline on every real model. Every test stays green and the entire performance claim silently reverts to the `2·n_free` FD sweep. The fix is small — parameterise the existing test over `[Exact, GaussNewton]`.

### 4. `src/estimation/agq.rs:2323` — deleted coverage rather than re-targeted (low)

The removed population-level assertion checked that when the per-subject analytic gradient fails, `population_gradient` lands on the explicit FD-of-quadrature fallback and is therefore `reconverge_gradient_interval`-independent. The reasoning for removing it is correct — the fixture no longer fails — but that fallback is still live production code and is now untested.

The precondition can be restored rather than dropped: drive the same fixture with `HessianAnchor::Exact`, whose grid response is still FD by default, which reinstates the "the analytic per-subject grid response fails unconditionally" premise the assertion needed.

### 5. `src/estimation/agq.rs:2677` — the A/B tolerance uses a single cross-coordinate scale (low)

```rust
let scale = fd.iter().fold(1e-3f64, |m, v| m.max(v.abs()));
```

`scale` is one scalar for the whole gradient vector, so every coordinate is compared against the *largest* coordinate's magnitude. On this fixture the Omega-block coordinates dominate (`Omega⁻¹` around 11 at `omega² = 0.09`), so a sigma or theta coordinate whose grid response is one or two orders of magnitude smaller passes with an effectively unbounded relative error. Given that a missing `dtheta` chain in the magnitude direct-theta channel is described at `focei_htilde_dx.rs:186` as a bug a parity test caught, it is worth having these bind tightly; a per-coordinate `fd[k].abs().max(eps)` scale would. (`focei_htilde_dx.rs:409` is per-coordinate already, though floored at `1.0`; there the `dH̃/dx` entries are of order `1e2`, so the floor should not bind.)

### 6. `src/estimation/focei_htilde_dx.rs:201` — fixed coordinates are computed, and can decline the whole subject (low)

`subject_htilde_dx` loops `for k in 0..x.len()` with no `fixed` skip, and `:272` returns `None` for the entire subject if any entry of any `dh[k]` is non-finite — including at a coordinate `fd_grid_response` never evaluates at all (it `continue`s on `fixed[k]`).

Output remains correct, since the caller falls back to FD. But a `FIX`ed, pathologically-scaled coordinate can silently route every subject back to the `2·n_free` sweep while the analytic route was perfectly usable on the free coordinates, and nothing reports that the speedup was lost. Passing the `fixed` mask in would remove both the wasted `d × d`-per-observation work and the spurious decline. Related: `analytic_grid_response` (`agq.rs:1604`) recomputes the `packed_fixed_mask(template)` that `grid_response_correction` (`agq.rs:1312`) already has in scope.

---

Nice piece of work overall. Reusing `RegularisedAnchor` rather than re-deriving the factor differential is the right call, and keeping the exact-anchor route opt-in on a negative wall-clock result is well argued and well documented.

hiddevandebeek:
Addressed all four Copilot review comments in 3623ddc2:

- **Perf bug (real)**: `build_frem_r_override` in `focei_htilde_dx.rs`'s σ-derivative precompute loop was called inside the per-observation loop for both `sp`/`sm`, recomputing an O(n_obs) override vector O(n_obs) times per σ slot under FREM. Hoisted to once per `sp`/`sm`.
- Fixed both fixture header comments (`warfarin_block_omega_focei_agq.ferx` mislabeled itself as "exact-anchor Laplace" — it's actually the Gauss-Newton FOCEI-AGQ fixture) and the stale `agq::prefer_analytic_grid_response` symbol references (renamed to `use_analytic_grid_response` earlier in the branch, comments weren't updated).
- Fixed the `analytic_grid_response` doc comment in `agq.rs` to name both per-anchor scope providers (`laplace_h_deriv::subject_h_inner_dx` for `Exact`, `focei_htilde_dx::subject_htilde_dx` for `GaussNewton`) instead of only the first.

All 31 relevant tests (`agq::`, `focei_htilde_dx::`, `laplace_h_deriv::`) still pass, including the FREM/custom-magnitude parity test that exercises the fixed loop.

hiddevandebeek:
Widened the analytic grid-response scope significantly beyond the initial cut, per feedback that it should cover everything theoretically reachable (not just the cheapest cells). Two more commits (f7b1cf17, 0d1bcb0f):

**`focei_htilde_dx` (Gauss-Newton anchor, default-on) now covers:**
- M3-BLOQ, including the σ-direct derivative (central-differences `m3_censored_outer`'s `g2` kernel directly, since it has no `dp_dv`-style closed form)
- `iiv_on_ruv`, **including combined with an M3-censored row** — reuses `ScoreCore::cens_dcz_df`/`cens_dcm_df` for the structural half, adds the σ-direct pair and a `ruv_scale`-response term the FD-parity test itself caught missing (residual variance scale `exp(2·η_ruv)` moves `p`/`g2`/`ruv_cz`/`ruv_cm` even at fixed `f`, and that scale moves with every packed coordinate through `db_dx[k][rr]`)
- Correlated residuals (`block_sigma`) — differences the correlation-aware `corr_residual_rd_at_sigma` instead of the plain per-endpoint variance function, and adds a new packed-coordinate segment for the ρ off-diagonal itself (closed form via `rho_rd_terms`, chained through the Fisher-z `rho_chain` factor)
- Custom/TV σ magnitude, LTBS, `ExpressionScale`, FREM, ODE (already covered before this round)

**`laplace_h_deriv` (Exact anchor, opt-in) now also covers ODE** — the covariance provider already reached it via `ode_analytical_supported`; this was purely an untested-and-staged-out exclusion, closed with a new parity test at `tol = 5e-3` (solver-tolerance noise, not a formula gap).

**What's left, deliberately deferred:** IOV/mixture for both modules. This is a different shape of gap from everything above — `packed_len`'s own `n_iov`/`n_mixture` segments aren't decoded by either module's layout guard, `omega_inv` changes meaning under the stacked `[η, κ]` system (bare η block → `Ω_bsv ⊕ K·Ω_iov`), and the `agq.rs` caller would need to switch to the IOV sensitivity/`eta_dx` providers too — not a self-contained change to either module alone. It's also where the measured build-cost caution applies (ODE-IOV axes are ~67% of the debug build per a prior finding), so it's staying out of this PR rather than risking a slow, half-tested addition.

All new cells are covered by FD-parity tests (including two genuine bugs the parity tests caught mid-implementation, not just declines-to-tests). Full `cargo test --lib`: 4284/4284.

hiddevandebeek:
IOV support landed for the default-on `focei, n_agq > 1` path (commit 8e03e076), closing the last cell of the analytic Gauss-Newton grid response.

**What it is:** `focei_htilde_dx::subject_htilde_dx_iov`, a sibling of the non-IOV function (matching this codebase's own established `*_iov` convention rather than branching one function two ways). The IOV FOCEI marginal is exactly the ordinary FOCEI Laplace objective over the stacked system `b = [η, κ₁..κ_K]` with prior `Σ_b = Ω_bsv ⊕ K·Ω_iov`, so almost none of the residual-chain math (`p`, `β`, `iiv_on_ruv`, M3, σ-direct derivatives — everything from the last round) needed to change. What was genuinely new:

- packed layout `[θ, Ω_bsv, σ, Ω_iov]` (σ sits *between* the two Ω segments)
- the sensitivity provider (`subject_sensitivities_iov`)
- a BSV or IOV Ω coordinate's `dΩ⁻¹/dx`, which reads a column of the *joint* Cholesky factor (`block_chol_full`) — an IOV entry moves `K` replicated diagonal blocks of `Σ_b` at once, so its derivative sums `K` such contributions (mutation-tested: dropping that sum kills the parity test hard, not a silent near-miss)

`agq.rs`'s caller already passed the joint `omega_joint_inv` unconditionally and already dispatched `db_dx` through `subject_eta_dx_iov` under IOV — the only wiring change was a third `HessianAnchor::GaussNewton if stack.is_iov()` match arm.

**On "gradient and covariance"** — the covariance step's score cross-product (`assemble_score_cross_product` → `SubjectScoreContext` → the same `grid_response_correction`/`analytic_grid_response` path) reuses this exact gradient, so IOV models get analytic-gradient covariance for free here too; no separate covariance-specific change was needed. `laplace`'s own IOV extension (for the opt-in Exact anchor) is noted as a follow-up in the plan doc, since it needs the same treatment applied to the third-order case and that route has no proven speed win regardless.

**Validation:** FD-parity + mutation test at the isolated-function level, and a new end-to-end test (`tests/agq.rs::iov_gauss_newton_anchored_gradient_matches_fd`) comparing the real `agq_population_gradient` against FD of the actual IOV population objective — this is the test that actually exercises `agq.rs`'s dispatch, not just the formula. Full `cargo test --lib`: 4285/4285. `tests/agq.rs`: 34/34.

Remaining scope, deliberately not attempted: mixture (a different packed segment, not the stacked-prior story IOV is), and `block_sigma` + IOV (unreachable through either module today — `score_core`'s own scope keeps them mutually exclusive in practice).

hiddevandebeek:
## Code review — PR #1308

Reviewed all 5 commits (~3.5k lines) against merge base `cfc956da`. I re-derived the two derivative assemblies by hand against `score_core`'s `htilde`/`h_inner`, the packed layouts in `pack_params`, the jitter in `build_proposal` vs `regularised_anchor`, and the node placement in `Proposal::apply_l_sigma` vs `RegularisedAnchor::node_scale_derivative` — including the `dΩ⁻¹/dL` construction, the IOV `K`-block sum, and the FREM / `iiv_on_ruv` / M3 special cases. **No wrong-answer bug found in the numerics.**

Verified locally: `cargo test --lib` for the new modules (`focei_htilde_dx`, `laplace_h_deriv`, `agq::tests`) — 34/34 pass; `cargo test --test agq` — 34/34 pass, including the new `iov_gauss_newton_anchored_gradient_matches_fd`. I also temporarily duplicated `analytic_grid_response_matches_the_fd_route` with `HessianAnchor::GaussNewton` and confirmed the default-on route both engages and matches the FD baseline (reverted).

### Medium

**`src/estimation/agq.rs:2574` — the default-on route has no "did it actually engage" assertion.**
`analytic_grid_response_matches_the_fd_route` only exercises `HessianAnchor::Exact`, which is opt-in and off by default. The route this PR turns **on by default** — `GaussNewton`, plus its `stack.is_iov()` arm — has no test asserting `analytic_grid_response` returned `Some`. The test's own doc explains why that matters ("passes identically whether the analytic route runs or silently declines"), and that reasoning applies with more force to the default route. A later tightening of `analytic_score_supported` or of `score_core`'s runtime declines silently returns every subject to the `2·n_free` FD anchor sweep; the suite stays green and the advertised 30–50% call reduction is gone with no signal. I confirmed by construction that adding the `GaussNewton` case passes today — this is a durability gap, not a live defect.

**`CHANGELOG.md:28` — "unchanged" overstates the guarantee.**
"Estimates, OFVs and standard errors are unchanged (the two routes are pinned bit-consistent…)" is not accurate. The PR's own parity test pins the two routes at `1e-5` **relative**, a bound set by `AGQ_GRID_FD_STEP = 1e-5`'s truncation on the FD side. With the analytic route now default for `focei, n_agq > 1`, the gradient handed to the outer optimizer changes at that order and the trajectory changes with it — the same paragraph says "converging in fewer outer iterations", which is direct evidence the reported OFV/estimates/SEs are not bit-unchanged. A user reading "unchanged" will not expect a moved last digit.

**`plans/laplace-sensitivity-speed/bench.sh:24` — the comment contradicts the code, and the ODE case went unmeasured.**
The comment says "ODE models are benchmarked only for the GaussNewton (`focei`) case: `laplace_h_deriv::subject_h_inner_dx` declines ODE outright", but `laplace_h_deriv`'s module doc says "**ODE is admitted**" and `h_derivative_matches_fd_under_ode` passes. Consequence: the model loop at line 78 omits `warfarin_ode_laplace.ferx` (which the PR commits but never uses), so the one configuration where the `Exact` analytic route is most likely to win — ODE, where each perturbed anchor rebuild is a full re-integration — was never measured. That unmeasured case is precisely the evidence used to justify keeping `Exact` opt-in.

### Low

**`src/estimation/agq.rs:129` — `grid_response_override` swallows typos.**
`_ => GridResponseOverride::Auto` maps any unrecognised value to the default, so `analytical`, `FD`, `1` or `true` silently makes the "candidate" arm of a benchmark identical to the baseline and reports "no difference" — the exact failure mode this variable exists to avoid. Separately, `FERX_AGQ_GRID_RESPONSE` is user-visible behaviour and appears nowhere in `docs/` (only `CHANGELOG.md` and the `plans/` doc); CLAUDE.md requires the relevant docs page to be updated.

**`src/estimation/focei_htilde_dx.rs:425` (and its IOV twin at `:789`) — absolute FD step on a multiplicative variable.**
`let hs = 1e-6 * s.max(1.0)` is absolute in `s`. For `iiv_on_ruv` + M3 with `ruv_scale = exp(2·η̂_ruv) < 1e-6` (η̂_ruv < −6.9, reachable in an outer-loop excursion), `s - hs` is negative, so `at(s - hs)` hands `m3_censored_outer` a negative variance, `v.sqrt()` is `NaN`, `dh` fails the `is_finite` screen, and the whole subject silently reverts to the FD anchor sweep with no diagnostic. Well before that — around `s ≈ 1e-4` — the step is already a percent-level *relative* perturbation and `∂g2/∂s` loses several digits. Every other FD channel in this file scales its step to the variable (`sigma_fd_step`, `kernel_step`); this one does not.

**`src/estimation/agq.rs:2710` — the scope-gate test can pass without reaching the gate.**
`out_of_scope_models_decline_to_the_fd_grid_response` asserts only `analytic_grid_response(...).is_none()`, but that function has two earlier declines (`regularised_anchor(h)?`, `!reg.is_well_conditioned()`) before it reaches the `Exact` → `subject_h_inner_dx` dispatch the test names. With a zero `db_dx` and zero `node_grads`, nothing distinguishes "declined at the M3 scope gate" from "declined at the conditioning screen". Asserting `laplace_h_deriv::subject_h_inner_dx(...).is_none()` directly — the way the sibling module's `censored_rows_decline` does — would pin the intended gate.

**`src/estimation/focei_htilde_dx.rs:241` (and `:644`) — unconditional allocation on the hot path.**
`sigma_row_derivs_g` / `_cz` / `_cm` are allocated as `vec![vec![0.0; n_obs]; n_sigma]` unconditionally, though all three are only written and read under `core.ruv.is_some()` (the uncommon case). That is `3 · n_sigma` heap allocations per subject per gradient evaluation on the path this PR exists to make cheaper; the neighbouring `d2_row` already gates on `core.ruv.is_some()` and shows the pattern.

**`plans/laplace-sensitivity-speed.qmd:20` — committed plan is stale and self-contradictory.**
Line 20 still reads "no implementation or benchmark has yet run" and line 32 "no estimator implementation has changed and no speedup has been measured", while the closing sections document the implementation, the benchmark numbers and the IOV work. One bullet says "IOV/mixture stays declined for both modules" a few lines above a section describing `subject_htilde_dx_iov` as implemented and tested. It also hard-codes machine-local paths (`C:/package/ferx/wt-laplace-sensitivity`, `C:/package/ferx/cleanup-archive-20260909/stashes.txt`) and a local worktree-cleanup log. The next reader will act on a superseded handoff.

### Note, not a finding against this PR

The `max_diff = ...fold(|m,(a,b)| m.max((a-b).abs()))` accumulator in the new `tests/agq.rs:1502` swallows `NaN` (the pattern CLAUDE.md calls out explicitly). Four pre-existing tests in the same file use it identically, so this is worth fixing file-wide rather than blocking here.

hiddevandebeek:
Addressed all findings from the self-review in f5e4d584..5f0f92b3:

**Medium**
- `analytic_grid_response_matches_the_fd_route` now parameterises over both `HessianAnchor::Exact` and `HessianAnchor::GaussNewton`, so a future scope regression on the default-on Gauss-Newton route can't hide behind an `Exact`-only fixture.
- `CHANGELOG.md` no longer claims the FD/analytic routes are "pinned bit-consistent" — they agree to the FD route's own `1e-5`-relative truncation error, and the changed gradient moves the outer trajectory (fewer iterations on the ODE fixture is direct evidence of that).
- `bench.sh` now measures the previously-unmeasured ODE + `HessianAnchor::Exact` case: ~27% less provider time on `warfarin_ode_laplace.ferx` (3 interleaved reps). Re-measured the non-ODE diagonal-Ω case too while I was in there — an earlier run had recorded +34% slower, a fresh run shows ~14% faster; `laplace-sensitivity-speed.qmd` and `bench.sh`'s header comment now report both instead of trusting the single stale run.

**Low**
- `grid_response_override` now panics on an unrecognized `FERX_AGQ_GRID_RESPONSE` value instead of silently falling back to `Auto`.
- `FERX_AGQ_GRID_RESPONSE` is now documented in `docs/estimation/agq.qmd` (new "Analytic instead of finite-difference `dH/dθ`" subsection).
- `focei_htilde_dx.rs`'s σ-derivative precompute (both the base and IOV variants) now uses `sigma_fd_step`'s relative-with-floor step instead of an absolute `1e-6`, which could drive `s - hs` negative under `iiv_on_ruv` at extreme `ruv_scale` and hand `m3_censored_outer` a negative variance (`NaN` from the inner `sqrt`, silent decline to the FD sweep). Added a regression test that fails without the fix.
- `out_of_scope_models_decline_to_the_fd_grid_response` now asserts `laplace_h_deriv::subject_h_inner_dx(...).is_none()` directly, pinning the M3 scope gate itself rather than a downstream conditioning screen.
- The three `sigma_row_derivs_g`/`_cz`/`_cm` allocations (both variants) are now gated on `core.ruv.is_some()`, matching the neighbouring `d2_row` pattern.
- `laplace-sensitivity-speed.qmd`'s stale/self-contradictory claims are fixed, including the machine-local workspace-cleanup section and a leftover line still listing IOV as a blanket structural exclusion after `subject_htilde_dx_iov` landed IOV support for the Gauss-Newton anchor.

The NaN-swallowing `fold` pattern in `tests/agq.rs` was flagged as a pre-existing, file-wide issue and explicitly not a finding against this PR, so left alone here.

Verified: full `cargo test --lib` (4286 passed, 0 failed) and `tools/preflight.sh` (fmt, all 5 check-matrix feature sets, clippy, rustdoc, docs-lint, public-API) both green.

Issue #1307, opened 2026-09-09
estimation: pack_params clamps before any bound is consulted, so a `block_sigma ... FIX` above ρ≈0.995 is not held at the declared value


TeunP:
Found while planning #1251 (batch E), measured at `909ad382`. Two doc comments in `src/types.rs`
state a contract that is not kept.

## The contract

`src/types.rs:1948` — *"a plain `block_sigma (...) = [...]` **estimates** `rho` (NONMEM `$SIGMA
BLOCK(n)` semantics, #847); `block_sigma (...) = [...] FIX` **holds it at the declared
value**."* `src/types.rs:5776` says the same thing about `FitResult`.

## Measured — it does not, above ρ ≈ 0.995

`pack_params` does not only *transform* the declared parameters, it **clamps** them, before any
bound is consulted (`src/estimation/parameterization.rs:62`):

```rust
pub(crate) fn pack_rho(rho: f64) -> f64 {
    rho.clamp(-RHO_CLAMP, RHO_CLAMP).atanh().clamp(-RHO_Z_BOUND, RHO_Z_BOUND)   // RHO_Z_BOUND = 3.0
}
```

`block_sigma (E1, E2) = [0.04, cov, 0.09] [FIX]`, sweeping `cov` so ρ takes each value:

| declared ρ | `FIX` | packed | box | **what the optimizer sees** | moved |
|---|---|---|---|---|---|
| 0.990 | no | 2.6467 | [−3, 3] | 0.990000 | no |
| 0.990 | **yes** | 2.6467 | [2.6467, 2.6467] | 0.990000 | no |
| 0.995 | **yes** | 2.9945 | [2.9945, 2.9945] | 0.995000 | no |
| **0.999** | **yes** | **3.0000** | [3.0000, 3.0000] | **0.995055** | **yes** |
| **0.9999** | **yes** | **3.0000** | [3.0000, 3.0000] | **0.995055** | **yes** |

The parser accepts these — `build_residual_correlations`
(`src/parser/model_parser.rs:13962-13970`) rejects only `|ρ| ≥ 1`. So a `FIX`ed declaration of
0.999 is held at **0.995055**, and every declared ρ above `tanh(3) ≈ 0.995055` collapses onto
that same value. Nothing reports it.

The same shape on the value floor, where `pack_params` applies `.max(1e-10)` to θ, the Ω
diagonals, Σ and the mixture overrides:

```
[...]

Issue #1289, opened 2026-09-08
bug(tests): SS_WARN_SINK_READER_GUARD does not serialize writers, so an is_empty() assertion on the SS warning sink is a flake


TeunP:
### Symptom

```
---- ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback stdout ----
thread '...' panicked at src/ode/predictions_tests.rs:6743:5:
a converging nonlinear SS must not warn
test result: FAILED. 4513 passed; 1 failed; 23 ignored
```

Seen once on `cargo test -p ferx-core --features ci,markov,nn --tests`. The same test
passes in isolation, and the whole lib suite passes under `--features ci`.

### Why

`ss_nonconvergence_sink()` (`src/dosing.rs:546`) is a single process-global
`Mutex<BTreeSet<String>>`. `SS_WARN_SINK_READER_GUARD` (`:668`) is taken by **readers only** —
its doc comment says so explicitly:

> Writers don't drain, so only readers need to coordinate.

That is true for a reader asserting on *its own* warning, and false for a reader asserting the
sink is **empty**. The reader sequence is

1. take `SS_WARN_SINK_READER_GUARD`
2. `clear_ss_nonconvergence_warnings()`
3. run its own equilibration (which must not warn)
4. assert `take_ss_nonconvergence_warnings().is_empty()`

Any other test on another harness thread that runs a capped SS equilibration inserts into the
same global set between (2) and (4), via `note_ss_nonconvergence_if_capped` (`:557`), which
never touches the guard. The reader then fails on a warning it did not produce. Nothing about
the guard can prevent this.

The probability scales with how many concurrent SS writers are compiled in, which is why it
showed up under `ci,markov,nn` (4,513 lib tests) and not under `ci` (4,176).

### Affected

Every `…is_empty()` assertion on the sink. `grep -rn SS_WARN_SINK_READER_GUARD src` — 5 sites
across `dosing.rs`, `ode/predictions_tests.rs`, `pk/event_driven.rs`.

### Fix options

- Have `note_ss_nonconvergence_if_capped` take the same guard, making it a true global lock on
  the sink for the test binary. Simplest; costs a lock on a path that already takes one.
- Or make the sink thread-local for tests, so a writer on another harness thread cannot reach a
  reader's set at all — closer to the intent, and removes the guard entirely.

The second is the better shape: the guard exists only because the sink is global, and the sink
is global only because production drains it once at the `api` boundary.

[...]

Issue #1288, opened 2026-09-08
perf(sim): simulate_with_uncertainty re-runs the draw-independent assert_* guards once per draw


TeunP:
### What

`simulate_inner_with_draw` is the single simulate chokepoint and runs six `assert_*` guards
at the top (`src/api/simulate.rs:886`). Five of them are **θ- and draw-independent**:

| guard | depends on the draw? |
|---|---|
| `assert_modeled_doses_supported` | no |
| `assert_dose_compartments_supported` | no |
| `assert_absorption_closed_form_support` | no |
| `assert_survival_tv_covariates` | no |
| `assert_absorption_dosing_supported` | no — it reads `model.default_params.theta`, not `params` |
| `assert_absorption_flip_flop_no_twin` | **yes** (`&params.theta`) |

`simulate_with_uncertainty` calls `simulate_inner_with_draw` once per draw
(`simulate.rs:1104`), so the five recompute an identical verdict N times.
`simulate_with_options_diag` runs them twice per call (`:446`, then `:894`).

### Measured

Release, synthetic 1000 subjects × (2 doses + 100 observations), transit absorption with a
time-varying covariate:

| | |
|---|---|
| `check_absorption_dosing` (the widened one, #1235) | **21.53 ms** |
| one full simulation pass over the same population | **642.4 ms** |
| overhead per draw | **3.4%** |

Before #1235 widened the snapshot set it was 1.56 ms, i.e. 0.24%. So the repetition is
pre-existing and the cost per repetition grew ~14×. At 1000 draws this is ~21 s of
recomputing one constant answer.

### The trap

The obvious fix — gate on `draw == 1` — is **wrong**. The uncertainty loop `continue`s past a
draw whose flip-flop check fails (`simulate.rs:1100`), so draw 1 can be skipped entirely and
the guards would then never run.

A correct hoist runs the five once in `simulate_with_uncertainty` before the loop and threads
an explicit "already checked" argument into `simulate_inner_with_draw`, so each of its four
call sites (`:452`, `:533`, `:561`, `:1104`) has to state which it wants. The flip-flop guard
stays per draw.

### Why not in #1286

#1286 is the PR that raised the per-repetition cost, and this was raised in its review. It is
left out because 3.4% of a draw does not justify restructuring the simulate chokepoint's guard
set inside a PR about validation snapshot sets — the chokepoint exists precisely so a caller
cannot forget a guard, and an "already checked" argument is the thing that can be passed
wrongly.

[...]

Issue #1271, opened 2026-09-06
ci: stacked PRs get no checks, and a squash-merge of the base orphans them


TeunP:
A PR whose **base** is another feature branch rather than `main` gets **zero CI checks**, and a
squash-merge of that base then orphans the stacked branch's history. Both bit during batch T
(#1258) within an hour of each other, and both will bite again — T4 stacks the same way.

## 1. Stacked PRs get no CI at all

`.github/workflows/ci.yml`:

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

`pull_request: branches: [main]` filters on the PR's **base**, so a PR opened against a feature
branch never triggers the workflow. #1270 (based on `worktree-t1-ekf-anchor-slots`) reported not
"pending" but **no checks at all**.

That failure mode is worse than a red build, because an empty check list is visually
indistinguishable from "CI has not started yet". The only verification is whatever the author ran
locally, and nothing on the PR says so.

`.github/workflows/slow-tests.yml` is a wider version of the same gap:

```yaml
on:
  schedule: [...]
  workflow_dispatch:
  push:
    branches: [main]
```

No `pull_request:` trigger at any base, so **no** PR ever gets Tier-3 coverage — including PRs
based on `main`. That is deliberate (the comment says the tests "can't block pull-request merges"),
but it means a change to fit-reaching code is only exercised by CI *after* it lands, or by a manual
`workflow_dispatch` someone has to remember.

## 2. Squash-merging the base orphans the stacked branch

Because merges here are squashes, merging the base PR produces a commit with the same *tree* but no
parent link to the branch it came from. Measured immediately after #1263 merged as `7eb19818`:

```
$ git merge-base --is-ancestor c028a053 origin/main
exit 1                       # c028a053 is NOT an ancestor of main

$ git merge-base c028a053 origin/main
579d1e96                     # two commits behind main
```

So the stacked PR's merge base with `main` reverts to *before* the base PR's work. Retargeting with
`gh pr edit --base main` then produces a diff containing the already-merged changes again —
unreviewable at best, a duplicate or revert at worst. The correct move is a rebase that drops the
merged commits:

```
git rebase --onto origin/main <old-base-tip> <stacked-branch>
```

which is not obvious from the PR UI, and the failure is silent if you skip it.

## Suggested fixes

Not prescriptive about which — the first is cheap, the second is the real one:

[...]

PR #1309, opened 2026-09-09
feat(api): report an initial estimate that packs outside its own box [closes #1251]

### LINKED ISSUE #1251: estimation: nothing checks that a packed start is inside its own bounds — θ is silently clamped, with no diagnostic anywhere
Split out of PR #1246 (#1229). The reviewer's "altitude" note: #1229 rejects **one** kind of
start that sits outside its own box (a free Ω diagonal on the `-6` rail), and the reason it is
worth rejecting — the optimizer silently moves your start and never tells you — is not specific
to Ω.

## What

`optimize_population` and every sibling optimizer begin the same way:

```rust
    let bounds = compute_bounds(init_params);
    let mut x = pack_params(init_params);
    clamp_to_bounds(&mut x, &bounds);          // src/estimation/outer_optimizer.rs:255
```

and `clamp_to_bounds` is exactly what it says (`src/estimation/parameterization.rs:857`):

```rust
pub fn clamp_to_bounds(x: &mut [f64], bounds: &PackedBounds) {
    for i in 0..x.len() {
        x[i] = x[i].clamp(bounds.lower[i], bounds.upper[i]);
    }
}
```

No comparison, no warning, no record that the start the user wrote is not the start that was
used. Ten non-test call sites across four modules — `outer_optimizer.rs` 255
(`freeze_flat_thetas`), 418 (`evaluate_at_initial_params`), 1434 (`optimize_nlopt_once`), 2355
(`optimize_bfgs`); `gauss_newton.rs:69` (`run_foce_gn`); `trust_region.rs` 657, 727
(`optimize_trust_region`); `vi/run.rs` 576, 729, 841 (`run_vi`).

Nothing upstream checks either. `src/parser/model_parser.rs:2560-2561` copies `lower`/`upper`
off the parsed thetas without ever comparing them to the init, and `check_model_options`
does not look at the packed vector at all.

## Measured, not inferred

A θ whose initial estimate sits **below its own declared lower bound** parses, packs outside the
box, is clamped, and produces no diagnostic from anything:

```
theta TVCL(0.05, 0.1, 10.0)      # init 0.05, lower 0.1
```

```
THETA ACCEPTED: theta[0]=0.05 lower=0.1 packed=-2.995732 lower_packed=-2.302585 OUTSIDE=true
rail check diagnostics: 0
check_model_options diagnostics: 0
```

`ln(0.05) = -2.995732` against a packed lower of `ln(0.1) = -2.302585`, so the fit silently
starts at `TVCL = 0.1` — a 2× move from what the file says — and the model file, the check
report and the fit output all still read `0.05`.

The same probe, on the coordinate #1229 deliberately leaves alone:

```
SIGMA packed=-23.0259 lower=-8.0000 OUTSIDE=true      # sigma PROP_ERR ~ 0.0 (sd)
```

A declared `sigma ~ 0.0` packs at `ln(1e-10) = -23.03` and is clamped to `exp(-8) ≈ 3.4e-4`. So
Σ is outside its box too — #1229 excludes it not because the predicate is false there but
because Σ was measured to *recover* from its rail, unlike Ω. That is exactly the per-coordinate
judgement the primitive below is meant to make expressible.

### The sharpest instance is one #1229 knowingly leaves open

An **evaluation-only** run (`maxiter = 0`, NONMEM `MAXEVAL=0`) clamps too —
`evaluate_at_initial_params` calls `clamp_to_bounds` before it evaluates. #1229 exempts
eval-only runs from its rail check, correctly, because nothing *holds* a start on the rail
without a search and because `--dofv` and `ferx gam --no-fit` legitimately hand in
machine-produced parameters. The price:

```
omega ETA_CL ~ 0.0        # stored as 1e-8, packed at ln(1e-4) = -9.21
                          # evaluated at exp(-12) = 6.14e-6   → 614×, silently
omega ETA_CL ~ 0.0 FIX    # pinned: lower == packed, clamp is a no-op → 1e-8
```

So the one place where ferx now *deliberately* declines to reject an out-of-box start is also
a place where the reported objective is computed somewhere the user did not ask for, with no
diagnostic. That is not an argument against the exemption — it is the argument for this issue:
the right fix is a general "your start is outside its own bounds" diagnostic that an eval-only
run can emit as a **warning** while a fitting run emits an error, and neither is expressible
today because nobody computes the predicate.

Pinned by `an_eval_only_fit_reports_the_rail_not_the_declared_zero` (`tests/check_command.rs`)
and documented in `docs/model-file/parameters.qmd`, so this is measured behaviour, not a
supposition.

## Why this is more than tidiness

The #1229 case is the sharp one — a free `omega ~ 0.0` is stored as the `1e-8` regularisation
floor, packs to `ln(1e-4) = -9.21`, clamps to `-6`, and **stays there**, so the fit reports a
variance of `exp(-12)` as if it were an estimate (#1242 has the measurement). But the general shape is the same for every coordinate: the reported
"initial estimate" in the fit output is the one the user wrote, while the optimizer started
somewhere else. Whether a given clamp is fatal, worth a warning, or genuinely harmless is a
per-coordinate question — and right now nothing is in a position to ask it, because nobody
computes the predicate.

## Suggested shape

One primitive in `estimation/parameterization.rs`, next to `pack_params` / `compute_bounds` /
`clamp_to_bounds`, which own the layout:

```rust
pub(crate) struct OutOfBox { pub index: usize, pub kind: PackedCoordKind, pub name: String,
                             pub packed: f64, pub bound: f64, pub below: bool }
pub(crate) fn coordinates_outside_bounds(template: &ModelParameters) -> Vec<OutOfBox>
```

with `api/validation.rs` choosing only **severity and wording** per `PackedCoordKind` — error
for a free Ω/Ω_IOV/mixture-Ω diagonal (today's #1229 behaviour, unchanged), and whatever is
decided for θ and Σ.

This would also remove a duplication #1229 introduced and that I would rather not leave
standing: `variance_decl_by_coordinate` in `src/api/validation.rs` re-derives the packed-segment
offsets (where the Ω block ends, where Ω_IOV starts, where the mixture overrides start) that
`packed_len` and `omega_structural_zero_mask`'s `iov_start` already derive inside
`parameterization.rs`. It works, it is pinned by tests, and it is in the wrong module —
`rho_packed_start` is the precedent for exporting the offset rather than recomputing it.

## Deliberately open

**What θ and Σ should do is a decision, not an implementation.** A clamped θ may be a typo worth
an error, or an intentional "start at the bound"; Σ was measured in #1229 to *recover* from its
`-8` rail, unlike Ω, so an error there would be wrong. Land the primitive and the θ/Σ decision
in whichever order suits — but the primitive is what makes the decision expressible.

## Not in scope

The `-6` rail value itself and the `1e-8` eigenvalue floor that sits under it — that is #1242.

TeunP:
## Why

A packed initial estimate outside its own optimizer box is moved by
`clamp_to_bounds()` before the first objective evaluation, and until now nothing said
so. `theta TVCL(0.05, 0.1, 10.0)` fits from `0.1` — a factor of two from what was
written — silently, on every run, with every estimate that depends on `TVCL` moving
with it. NM-TRAN refuses the identical stream outright (error 24).

`check_variance_init_rails` (#1229) already covers one slice of this, and only one:
its predicate is `if is_fixed || p > lo { continue }`, i.e. **lower-rail only**, on
Ω / Ω_IOV / mixture-Ω diagonals. Measured over eleven declarations, that is **1 of 10**
out-of-box coordinates flagged. `omega ETA_CL ~ 1e8` packs at `+9.21` against the `+6`
rail, is clamped to a variance of `1.6e5`, and says nothing — a gap inside #1229's own
coordinate kind.

The best evidence that this matters is ferx's own code.
`crates/ferx-tools/src/ruvsearch/mod_tests.rs:1102` reads *"RUV_TV = 25 >
TIME_VARYING_UPPER (10) → upper / 2; this seed used to be written as
`theta RUV_TV(25, 0.01, 10.0)` verbatim."* `ruvsearch` generated a θ outside its own
box, hit the silent clamp, and worked around it locally in `init_from_screen` rather
than diagnosing it.

**#1252 rides along** because it is the substrate, not because the files overlap.
`compute_bounds` cannot build the box without also building the packed vector and the
FIX mask — a FIX-ed coordinate is pinned at its own packed value — and then discarded
both. The start-side check and the *existing* post-fit `runaway_guard_estimates` walk
exactly those three vectors and differ only in the comparison.

## What changed

Three pieces, in that order, with the seam kept clean (see **Reviewer hints**).

**`pack_with_bounds` (#1252).** One walk returning `PackedStart { packed, bounds, fixed }`.
`compute_bounds` is now that function with two results dropped, so the box is
bit-identical and the ~35 existing `compute_bounds` test call sites are untouched.
The box arithmetic moved to a private `unpinned_bounds` — private because a caller that
saw *that* box would judge a FIX-ed coordinate against a rail it is never optimized
against. `packed_segments` gives the packed layout one derivation; three places
re-derived it inline. All fourteen non-test callers rewired, including
`trust_region.rs:727`, which computed the whole box a second time inside a function
that already had one.

**`coordinates_outside_bounds` (#1251).** The shared predicate: strictly outside, with
`theta_guard_is_internal` moved out of `api/postfit.rs` into `parameterization.rs` so
both ends use one copy.

**`check_packed_start_in_box` (#1251).** Severity split on *whose* bound it is:

| coordinate | whose bound | verdict |
|---|---|---|
| θ strictly outside its **declared** lower / upper | the user's | **error** `E_THETA_INIT_OUTSIDE_BOUNDS` |
| θ outside the hidden `1e-10` / `1e9` cap | ferx's | warning `W_INIT_OUTSIDE_BOUNDS` |
| Ω / Ω_IOV / mixture-Ω diagonal, **below** `−6` | ferx's | *not claimed here* — #1229's error |
| Ω / Ω_IOV / mixture-Ω diagonal, **above** `+6` | ferx's | warning |
| Ω off-diagonal outside `±10` | ferx's | warning |
| Σ outside `[−8, 5]` | ferx's | warning |

Both call sites (`fit()` and `ferx check`), both diagnostic channels, and #1229's
`maxiter = 0` exemption.

### Two decisions the plan did not settle, both measured

**1. The Ω upper rail is a warning, not an error.** The plan left this "error, pending
step 1" and made one warfarin fit the precondition. It recovers — under every arm:

| arm | base OFV | `omega ETA_CL ~ 1e8` | Δ |
|---|---|---|---|
| foce×bobyqa | −280.2230 | −280.3638 | **hi better by 0.1408** |
| foce×slsqp / lbfgs / bfgs | −280.3640 | −280.3640 | 0.0000 |
| focei×bobyqa | −285.9972 | −286.0009 | hi better by 0.0037 |
| focei×slsqp / lbfgs / bfgs | −286.0042 | −286.0042 | 0.0000 |

θ and ω agree to 5–6 significant figures throughout. The high start really is a
different start — its first evaluation is OFV `−78.066` against the base's `−200.288`,
so it does begin at the clamped `1.6e5` — and every optimizer walks home from there.
No split across optimizers, so it does **not** go to #1242. NONMEM's `$OMEGA 1e8` run
(accepted, then `ROUNDING ERRORS`, OFV 117.63 worse) was suggestive only, exactly as
the plan said: NONMEM starts literally at `1e8`, ferx clamps to `1.6e5` first, so they
are different experiments and the ferx arm is the one that decides.

**2. A new `WarningCode`, not `BoundaryEstimate` — a deliberate deviation from plan §4.3.**
The plan chose to reuse `BoundaryEstimate` (matching on `"optimizer bound"`) because
ferx-r keys remediation off the category vocabulary. That traces the ferx-r coupling
but not the consumers of the category. Measured with a scratch `#[test]`:

```
SCRATCH category = BoundaryEstimate
SCRATCH structured = [BoundaryEstimate]
SCRATCH estimate_near_boundary = true
SCRATCH strictness passed = false failures =
  ["estimate pinned to a declared bound: TVCL starts at 1.00e11, above the optimizer
    bound of 1.00e9 (packed 25.33 vs 20.72); the start is clamped there."]
```

`model_selection::estimate_near_boundary` reads that category, and it drives **three
default-on** filters:

* `bootstrap`'s `skip_estimate_near_boundary` (`BootstrapOptions::default()`,
  `bootstrap/mod.rs:338`) → the replicate is dropped from the summary
  (`bootstrap/summary.rs:169`);
* `Strictness::reject_on_boundary` (`model_selection.rs:268`) → a search candidate is
  rejected, under a failure line that contradicts itself;
* ferx-r's `R/check_strictness.R:257` under `reject_on_boundary = TRUE` (`:80`).

So a *start*-side notice wearing that category would silently drop bootstrap replicates.
And the plan's stated cost of a new variant does not hold: ferx-r's glue maps the
category with `w.category.as_str().to_string()` (`src/rust/src/lib.rs:2208`) — no
exhaustive `match` to break — and `.ferx_warning_guidance` is documented to *"return
NULL for unknown categories so callers can skip printing rather than showing a generic
placeholder"*. The warning still prints with severity, category and message; it just
lacks a remediation paragraph until a follow-up ferx-r PR adds the row.

`WarningCode::InitOutsideBounds` (`"init_outside_bounds"`) it is, matched on an explicit
`W_INIT_OUTSIDE_BOUNDS` token following `W_ABSORPTION_TWIN_DECLINED`'s precedent rather
than on prose.

## Alternatives considered

**One PR vs. stacking or splitting.** Stacking #1251 on a #1252 branch gets **zero** CI
checks (`ci.yml` triggers on `pull_request: branches: [main]`, #1271). Two sequential
PRs would ask a reviewer to accept a `PackedStart` with no caller that needs it. One PR
with both closed at the end would hide the seam. So: one PR, with #1252 self-contained
in the first two commits and independently revertible.

**Reusing the `BoundaryEstimate` category** — measured above, and rejected.

**A `packed_fixed_mask` consult inside `coordinates_outside_bounds`** — deliberately
omitted. `pack_with_bounds` pins a FIX-ed coordinate to `lower == upper == packed[i]`
*from the same packed vector*, so a strict inequality is structurally false for it. A
mask test would be a second gate rejecting exactly what the first already rejects — the
"two redundant gates cover for each other" hole CLAUDE.md names, which #1229 hit three
times. The FIX behaviour is instead pinned by a test asserting the **mechanism**.

**An inclusive (`<=` / `>=`) rule**, which is what NM-TRAN does (errors 627 / 628). At
`packed == bound` the clamp is a **no-op**, so nothing is silently moved and this issue's
premise does not apply. Over 4,674 θ literals in 642 files the exact predicate finds
3 strictly outside (all three *prose*, not fixtures) and 2 exactly on a bound, both live
and both idiomatic — `theta TVF(1.0, 0.01, 1.0)` and `theta TVLAG(0.0, 0.0, 12.0)`. An
inclusive rule would break both to catch nothing.

## Cross-repo dependency

ferx-r's Rust glue does not `match` on `WarningCode` exhaustively (measured above), so
the new variant does not break its build, and an unknown category degrades to "printed
without remediation guidance". A follow-up issue adds the `init_outside_bounds` row to
`.ferx_warning_guidance`.

## Breaking changes
- [ ] None

One added public item: **`WarningCode::InitOutsideBounds`**. Its caller is
`crate::types::classify_warning`, and its consumer is any reader of
`FitResult::warnings_structured` — ferx-r's `ferx_get_warnings()` first among them —
which must be able to tell a clamped *start* from a *boundary estimate* because the
latter drives three default-on rejection filters. The existing surface does not suffice
precisely because `BoundaryEstimate` already means the other thing.

Everything else added is `pub(crate)`: `PackedStart`, `pack_with_bounds`,
`PackedSegments`, `packed_segments`, `BoxSide`, `OutOfBox`,
`coordinates_outside_bounds`, `theta_guard_is_internal`, `check_packed_start_in_box`.
Re-measured on this branch: `git grep 'PackedBounds\|pack_params\|compute_bounds\|packed_fixed_mask\|clamp_to_bounds' -- crates/` and the same over `../ferx-r` `origin/main`
`src/rust/src/` both return nothing, so no member or ferx-r caller needs any of them.

## Numerical validation
- [x] Compared against: NONMEM (as a measured *negative* result — see below) and the base ferx arm
- [x] Reference values:

```
# base — examples/warfarin.ferx, covariance = false, foce x bobyqa
OFV -280.2230   TVCL 0.134094   ETA_CL 0.026480

# omega ETA_CL ~ 1e8, same model otherwise
OFV -280.3638   TVCL 0.132963   ETA_CL 0.028627
# first evaluation OFV -78.066 (base: -200.288) — the clamped start really is different
```

**No NONMEM anchor is committed**, per CLAUDE.md's "no NONMEM spelling" exception. The
NONMEM runs behind the severity table are NM-TRAN *rejections* (errors 24 / 627 / 628)
and a *failed* minimization — a measured negative result, evidence for the severity
table rather than a value to evaluate ferx at. Nothing here moves a number: the check
runs before the first objective evaluation, `clamp_to_bounds` still clamps exactly as
before, and `compute_bounds` returns a bit-identical box.

No `Dual2`-vs-FD parity row either: nothing here touches a sensitivity, gradient,
marginal or likelihood path.

## Performance

**#1252 is a duplication fix, not a performance fix, and the PR should say so.** Measured
on `--profile ci-fast`: the duplicated `pack_params` + `packed_fixed_mask` is **424 ns**
at warfarin's 7 coordinates and **781 ns** at 21; `compute_bounds` itself is 930 ns; one
warfarin FOCE fit is **172.2 ms**. All fourteen sites together are **0.0034%** of a fit —
70× below the 0.24% that #1288 itself dismissed as not worth a row. A bootstrap
multiplies numerator and denominator equally, so replicate count does not move the ratio.
The half of #1252's argument that survives is the correctness hazard: `compute_bounds`
computed the packed vector and the FIX mask and threw them away, and three places
re-derived the same packed-segment offsets.

`check_packed_start_in_box` runs once per `fit()`, at the same cadence as the existing
`runaway_guard_warning`, and allocates `coordinate_names` only on the firing path.

## Tests

## Docs

`docs/model-file/parameters.qmd:285` carried this issue's own placeholder ("A general
diagnostic for a start outside its own bounds is tracked separately") — replaced.
Two rows in `docs/file-formats/check-report.qmd`, one in `docs/warnings.qmd`, and a
new `### The same box, on the way in` subsection under
`docs/estimation/parameterization.qmd::Bounds`. No new `docs/.lint-baseline` entry.

## Checklist
- [ ] Functions on the `Dual2` sensitivity path stay differentiable — N/A, nothing on that path is touched

## Reviewer hints

**The seam.** Commits 1–2 are #1252 and are a *no-behaviour-change* refactor whose exit
condition is "every existing test still green". Commit 3 changes what `fit()` refuses.
They can be reviewed as two questions rather than one, and the #1251 commits can be
dropped without touching #1252.

**The riskiest change is re-seating `check_variance_init_rails` on the shared walk**,
because it ships an *error* — a regression there refuses a model that fits today. Its
978-line sibling test file is the guard, and it is only a guard **run unmodified**: 22 of
its assertions are `diags.len()` or `.is_empty()`, exactly the shape that goes quietly
green if the walk starts yielding nothing. It is unmodified in this PR.

**Mutation kill list.** Each mutation was byte-applied to a copy, run with
`--no-fail-fast` into its own log, and restored with `cp` + `touch`:

| mutation | dies |
|---|---|
| drop the Σ segment from the box walk | T1 + 12 pre-existing |
| Ω off-diagonal rail `±10` → `±6` | **T1 alone** |
| drop the FIX pin from `pack_with_bounds` | T1 + 6 pre-existing |
| off-by-one in `PackedSegments::sigma_start` | T2 + 25 pre-existing |
| swap `mixture_omega` / `mixture_sigma` lengths | **T2 alone** |
| narrow the `Below` comparison inward | T3 |
| widen both comparisons to `<=` / `>=` | T3b |
| drop the `theta_guard_is_internal` consult | T4 |
| make the partition lower-rail-only again | T5, T8 |
| remove the `outer_search_runs` gate | T7 |
| promote the internal-rail arm to an error | T6 |
| disable the `w_init_outside_bounds` classifier arm | T10 |

**One mutation initially survived**, and the fixture was at fault, not the assertion:
swapping the two mixture segment lengths left the whole suite green because the T1/T2
fixture declared *two* Ω overrides and *two* Σ overrides. It now declares three and two,
and the mutation dies. Worth knowing because it is the same shape as the degenerate
fixtures CLAUDE.md warns about.

**Two things I guessed and the tests corrected**, both recorded because they are the
argument for naming rather than counting:

* T12's allow-list. The plan's *characterisation* of the ten shipped out-of-box
  coordinates was exactly right (8 adaptive-dosing `omega ~ 1e-10`, 2
  `reset_init_snapshot_*` `~ 0.0`, all `OmegaDiagonal`/`Below`) but I wrote plausible
  file **names**. The measured set is `adaptive_platelet_ladder`,
  `adaptive_vanco_{auc,iov,iov_loading,loading,renal,renal_iov,renal_loading}` and
  `reset_init_snapshot_{fit,occ}`. A count would have passed.
* The partition split is `rails = 1` / `boxed = 3` on the four-coordinate fixture, not
  2/2.

**Engine per fixture: not applicable.** Every test here is on the diagnostic path, which
runs before the first objective evaluation and touches neither the analytic nor the FD
inner-gradient provider. The two Tier-2 tests that do call `fit()` do so at
`outer_maxiter = 1` and assert on the diagnostic channels, not on any number.

**What was left out**, all filed rather than folded in:

* `pack_rho`'s clamp inside `pack_params`, already filed as **#1307** — a second,
  invisible clamp layer that *no* packed-vs-box predicate can see, which is why this PR
  says "packed-vs-box" and never "your start is inside its own bounds", and why ρ has no
  code here rather than an unreachable arm. The `1e-10` value floor is the same shape and
  rides the same issue.
* **#1242** — the `−6` rail value and the `1e-8` eigenvalue floor beneath it.
* The `pub` surface with no consumers (`PackedBounds`, `pack_params`, `compute_bounds`,
  `packed_fixed_mask`, `clamp_to_bounds`, `packed_len`, `coordinate_names` are all in the
  baseline and nothing in `crates/` or ferx-r calls any of them). Narrowing it is its own
  breaking-ish decision.
* `ferx-tools` never reading `FitResult.warnings` at all (`bootstrap/mod.rs:791` reads
  only `warnings_structured`, filtered to covariance codes), so a bootstrap user sees no
  fit warning — which is what makes this PR safe from a warning storm, and is worth its
  own row.

**Two pre-existing flakes surfaced during verification, neither caused by this PR.**
`ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` failed once
in four full `--lib` runs and is **#1289** (the SS warning sink is not writer-safe).
`ferx-tools`' `gam::tests::a_nan_shrinkage_is_written_as_an_empty_field_not_the_text_nan`
failed once with `temp dir: Os { code: 17, kind: AlreadyExists }` — `gam.rs:1025` names
its scratch directory `(pid, nanos)`, which collides when two of its tests land in the
same clock tick. Not yet filed; a counter, as `tests/check_command.rs::temp_model`
already uses, is the fix.

## Open questions

None blocking. The one judgement call worth a second opinion is the new `WarningCode`
variant against the plan's reuse of `BoundaryEstimate` — the measurement is in **What
changed**, and the cost is one follow-up ferx-r guidance row.


TeunP:
## Review — head `d8ae4412`

First round. Reviewed in a worktree of the PR head; every number below is from a run on that tree, not from CI.

### Gates

| Gate | Result |
|---|---|
| `--workspace --features ferx-core/ci --lib` | 3 binaries, **4894 passed, 0 failed** |
| `--workspace --features ferx-core/ci --tests` | **203 binaries, 5893 passed, 0 failed**, 186 ignored |
| `--features ci,markov,nn --tests` | 181 binaries, 5583 passed, **1 failed** — `ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback`, i.e. #1289, which passed under `ci` on the same tree |
| `preflight fmt` / `docs` / `public-api` | green; no `Cargo.lock` change |

Independent checks:

- **Blast radius, wider than T12.** T12 ratchets `examples/` + `nonmem_anchor/`; `.ferx` files also live in `tests/fixtures/`, `tests/nonmem/`, `tests/reference/` and `tools/`. Running the exact predicate (log/identity packing, `1e-10`/`1e9` caps applied) over **all 168** `.ferx` in the tree finds **0 strictly out-of-box θ**. The new error breaks nothing shipped.
- **The partition holds.** #1229 fires on `p <= lo` on decl coordinates and skips `FIX`; this check skips `Below && decls.is_some()`. `FIX` pins `lower == upper == packed`, so a FIX-ed coordinate is never strictly outside — no double report and no gap, including at `p == lo`.
- **Machine-generated inits are safe**, which the PR description does not claim and which is the main thing I wanted to break. `ln(exp(ln(lo)))` never lands below `ln(lo)` over 20k samples, so a `--update-inits` replicate starting from a bound-pinned base estimate does not trip the new error; the journal writes through `fmt_exact`/`opt_exact`, so the `--resume` round trip is lossless. `ruvsearch::seed_inside` keeps θ seeds inside `[2·lower, upper/2]` and both `POWER_LOWER` and `TIME_VARYING_LOWER` are positive, so the margin really is a margin.
- **Old-contract grep** clean — `docs/model-file/parameters.qmd`'s placeholder was the only stale claim, and it is replaced.
- **Engine-per-fixture: "N/A" is right.** Everything here runs before the first objective evaluation, no sens path is touched, so no `Dual2`-vs-FD row is owed.

The #1252 half reads as bit-identical: `packed_len`, `omega_structural_zero_mask`'s `iov_start` and `variance_decl_by_coordinate`'s two offsets all map onto the same arithmetic through `PackedSegments`, and `trust_region`'s `post_run_bounds` is a copy of the box `compute_bounds(init_params)` would rebuild.

### Findings

| # | Finding | Severity | Evidence |
|---|---|---|---|
| 1 | **An inverted packed box is silently skipped, and then panics.** `coordinates_outside_bounds` returns `None` on `lower > upper` — which is exactly the case that panics thirty lines later in `clamp_to_bounds`. | should-fix | `theta TVCL(1.0, 5.0, 2.0)` — swapped bounds, a plain typo — parses (no parse-time `lower < upper` validation exists anywhere in `model_parser.rs` or `api/validation.rs`). Measured on this branch: packed `0`, lower `1.6094379124341003`, upper `0.6931471805599453`, `check_packed_start_in_box` returns **0 diagnostics**, and `clamp_to_bounds` panics at `core/src/num/f64.rs:1643`, `min > max, or either was NaN`. `theta TVCL(5e9, 2e9, 1e12)` — a declaration whose own lower *is* below its upper — inverts purely through ferx's hidden `1e9` cap (packed `22.333`, lower `21.416`, upper `20.723`), also **0 diagnostics**: that one is this check's own "ferx's bound, not the user's" row. The panic predates the PR, but the guard's comment ("A degenerate or non-finite box cannot place anything") reads as though declining is safe, and it isn't. Note the `!is_finite()` half **is** load-bearing and correct — `model_parser.rs:2575-2576` emits `theta_lower = -inf` / `theta_upper = +inf` for auto-declared thetas and `clamp(-inf, inf)` is fine. Only the `lower > upper` half hides a crash. |
| 2 | **The `PackedCoordKind::OmegaOffDiagonal` arm is reached by no test at any tier.** | should-fix | Mutation: arm body replaced with `unreachable!()`; `cargo test --workspace --features ferx-core/ci --lib --tests --no-fail-fast` → **203 binaries, 5893 passed, 0 failed, 0 panics**. The kill list's row *"Ω off-diagonal rail ±10 → ±6 \| T1 alone"* kills T1, which pins the rail value in `unpinned_bounds` — not this message arm. And T8, whose doc says it carries *"an out-of-box coordinate of every reachable kind and side at once"*, carries four: θ-Below-declared, Ω-diag-Below, Ω-diag-Above, Σ-Below. No off-diagonal, no Σ-Above, no θ-internal-cap, no mixture, no Ω_IOV — so "disjoint and together exhaustive" is verified over about half its domain. The arm most likely to fire in practice is a FREM `block_omega`, which is the case `unpinned_bounds`' own ±10 comment names. |
| 3 | **The Ω arm prints `exp(2·packed)` as "a variance"**, but for a `block_omega` / block `Ω_IOV` diagonal the packed coordinate is `ln(L_ii)`, so the number is the conditional variance, not the declared one. | should-fix | `validation.rs` Ω arm against its sibling at `validation.rs:4053-4062`, which spends ten lines on exactly this trap and writes the same quantity as `L = {l:.3e} (L² = {rail_variance:.3e})`. It only fires on the Above side (Below is #1229's), where correlations reduce `L_ii` and so the remedy direction is still right — the cost is a number the user cannot find in their model file. Untestable today for the same reason as #2: no block fixture reaches this check. |
| 4 | **The θ internal *lower* cap is unreachable, and two doc sites advertise it.** | nit | `packed = max(θ,1e-10).ln()` and `lower = max(θ_lo,1e-10).ln()`; `theta_guard_is_internal(_, "lower")` requires `θ_lo ≤ 1e-10`, in which case `lower = ln(1e-10) ≤ packed` always. So `theta_internal_cap(BoxSide::Below)` is never called and the `if hit.side == Below { theta_lower[..] }` inside the θ warning arm is dead. Harmless as code; the docs are not: `check_packed_start_in_box`'s own severity table says "the hidden `1e-10` / `1e9` cap" and `docs/estimation/parameterization.qmd` says "the `1e-10` / `1e9` Theta caps", while `CHANGELOG`, `diagnostics.rs`, `check-report.qmd` and `warnings.qmd` all say `1e9` only — correctly. |
| 5 | **"in one pass" / "in one walk" is four walks.** | nit | `pack_with_bounds`'s body is `pack_params` + `packed_fixed_mask` + `unpinned_bounds` + the FIX-pin loop. The property actually delivered — one call site, nothing discarded, one derivation of the layout — is what #1252 needed and what the PR body argues for; the phrasing states something stronger than the code does. |
| 6 | **The Σ remedy hardcodes `(-8.0).exp()` / `5.0.exp()`** where the Ω arms read `hit.bound`. | nit | A third copy of the Σ rail after `unpinned_bounds` and the docs tables. If the rail moves, the message's rail (from `hit.bound`) and its remedy (literal) disagree silently — nothing compares them. `THETA_PACK_FLOOR` / `THETA_PACK_CEIL` were introduced in this same PR for precisely this reason on the θ side. |
| 7 | **The two start checks each rebuild the substrate.** Per `fit()`: 2× `pack_with_bounds`, 3× `coordinate_kinds` (one direct plus one inside each `variance_decl_by_coordinate`), 2× `variance_decl_by_coordinate`. | nit | ~3 µs against a 172.2 ms fit on the PR's own numbers, so the cost is nil — the point is that this is the duplication #1252 exists to remove, one `PackedStart` threaded into both away from being closed. |
| 8 | **`fit()` now refuses a model it previously accepted**, and the CHANGELOG entry sits under `### Added` only. | nit | Keep a Changelog puts a previously-accepted input becoming a refusal under `Changed`. The blast radius really is zero (see above), which argues the wording is low-risk, not that the section is right. |
| 9 | `check-report.qmd::codes-parameters-and-random-effects` is now **4,230 / 5,000** characters. | informational | Within the plan's ≤1,800 budget for both rows (actual ~1,100) and docs-lint is green — but 770 characters of headroom means the next diagnostic row added to that table overflows R1 and forces the split in whichever PR happens to add it. |

### Not traced

`slow-tests`-gated behaviour (compiled, not run — that is the land-time dispatch's job), coverage percentages (Codecov's), and the ρ / `1e-10` invisibility, which is correctly deferred to #1307.

Findings 1–3 are the ones worth a round; 4–9 are cheap.

TeunP:
## Review round 1 — resolved at `9ed38f6c`

Eight of nine fixed in `9ed38f6c`; one declined with the measurement. Rebased onto
`2718f06b` (main moved three commits under this PR — #1301, #1299, #1308 — overlapping
only in `CHANGELOG.md`), so the history is linear and `codecov/patch` computes.

| # | Finding | Fix |
|---|---|---|
| 1 | An inverted packed box was skipped by `coordinates_outside_bounds` and then panicked in `clamp_to_bounds` | **Fixed.** New `coordinates_with_inverted_bounds` + `E_THETA_BOUNDS_INVERTED`. See below — the finding got bigger under verification. |
| 2 | The `OmegaOffDiagonal` arm was reached by no test at any tier | **Fixed.** New `block_omega_model` two-eta fixture; `block_omega (ETA_CL, ETA_V) = [1.0, 15.0, 400.0]` puts `L₂₁ = 15` outside `±10` with both diagonals well inside `±6`, so the arm is isolated. The `unreachable!()` mutation that survived the review now kills this test and nothing else. |
| 3 | The Ω arm printed `exp(2·packed)` as "a variance" on a `block_omega` | **Fixed.** Uses the same block/diagonal split `check_variance_init_rails` established, from the same `decls`. A block hit now reads `a \`block_omega\` Cholesky diagonal of L = … (L² = …, which is what is left of ETA_V's variance once the off-diagonals are accounted for, not the variance declared for it)`. |
| 3b | *(found while fixing 3)* computed rails printed at full f64 width | **Fixed.** `{:.3e}` for every computed number, `{:e}` kept for declared literals, which round-trip. `1.6275479141900392e5` → `1.628e5`. |
| 4 | The θ internal **lower** cap was unreachable, and two docs advertised it | **Fixed.** `theta_internal_cap` deleted; the arm names `THETA_PACK_CEIL` unconditionally with the proof in a comment. The doc table and `parameterization.qmd` now say `1e9` only — matching what `CHANGELOG`, `diagnostics.rs`, `check-report.qmd` and `warnings.qmd` already said. |
| 5 | `pack_with_bounds` documented itself as "one pass"; it is four walks | **Fixed.** The doc now states the property #1252 actually needed: the three products stop being computed and discarded. |
| 6 | The Σ remedy carried its own `-8` / `5` literals | **Fixed.** `SIGMA_PACK_LOWER` / `SIGMA_PACK_UPPER`, used by `unpinned_bounds` and the message, and the Σ test now asserts the message quotes the interval the box uses. |
| 7 | Both start checks rebuild the substrate | **Not fixed, deliberately.** ~3 µs against a 172.2 ms fit, and closing it means re-opening `check_variance_init_rails` — the step the plan names as this PR's riskiest, because it ships an error. The asymmetry is recorded at the call site, which is what CLAUDE.md asks for when two callers genuinely cannot share. |
| 8 | The refusal was filed under `Added` | **Fixed.** The refusal moves to `Changed`; the abort fix is a `Fixed` entry. |
| 9 | `check-report.qmd` had 770 characters of R1 headroom | **Fixed.** The parameters section is split into declarations and *Initial estimates against the optimizer box*; the new row lands with room, no new baseline entry. |

### Finding 1 got bigger under verification

The review named two shapes. Probing all four at `d8ae4412` found a third that is not a typo
at all:

```
theta TVCL(1.0, 5.0, 2.0)        packed 0.000   lo  1.609  hi  0.693   diags 0
theta TVCL(5e9, 2e9, 1e12)       packed 22.333  lo 21.416  hi 20.723   diags 0
theta TVCL(1e-12, 1e-13, 1e-11)  packed -23.026 lo -23.026 hi -25.328  diags 0
```

The third is an ordinary declaration of a small parameter, with its own lower bound below
its own upper. It is ferx's `1e-10` packing floor — substituted for the declared lower —
that empties the box. All three reached `f64::clamp(min, max)` with `min > max` and took
the process down.

`E_THETA_BOUNDS_INVERTED` reports all three, naming which cause applies. It is the **one
start-side check with no `maxiter = 0` exemption**: the exemption exists because a clamped
start does not stick without a search, but an empty box is not clamped at all, and
`evaluate_at_initial_params` clamps before it evaluates — so gating it there would have
left exactly the aborting case unreported.

This also closes finding 4 more cleanly than expected. There is no `1e-10` counterpart to
the `1e9` warning because a start below the floor packs *equal* to its floored bound, and a
range wholly below the floor is this error instead.

### Mutation kill list

Each applied by byte copy, run with `--no-fail-fast` into its own log, restored by `cp` +
`touch`, verified byte-identical after. Every one names its own row.

| mutation | dies |
|---|---|
| off-diagonal arm body replaced | the new off-diagonal test, **alone** |
| forget the block/diagonal split | the new block test, **alone** |
| drop the empty-box loop (the pre-fix state) | all three empty-box tests |
| gate the empty-box loop on `outer_search_runs` | the two eval-only assertions |
| Ω rail `{:.3e}` → `{:e}` | the new block test, **alone** |
| Σ remedy stops reading the constant | the Σ test, **alone** |
| `NaN` stops counting as unorderable | the new `parameterization` test |
| outside-walk stops deferring an empty box | the new `parameterization` test |

The Tier-2 test calls `fit()` **before** it reads the report, deliberately: in the other
order a regression fails on the missing code and returns before `fit()` is ever called, so
it would look like an ordinary assertion failure rather than the abort the test exists to
prevent. Verified — under the pre-fix mutation the binary dies with
`min > max, or either was NaN. min = 1.6094379124341003, max = 0.6931471805599453`.

### Gates, on the rebased tree

| leg | binaries | passed | failed |
|---|---|---|---|
| `--workspace --features ferx-core/ci --lib` | 3 | 4969 | 0 |
| `--workspace --features ferx-core/ci --tests` | 204 | 5976 | 0 |
| `--features ci,markov,nn --tests` | 181 | 5623 | 0 |

`tools/preflight.sh` green on all six groups (fmt, check, clippy, rustdoc, docs, public-api),
no new `docs/.lint-baseline` entry, public-API baseline unchanged by this round.
`ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` (#1289) failed
once in the pre-rebase `markov,nn` run and passed here; it is untouched by this PR and passes
under `ci` on the same tree.

Two things re-checked because the rebase brought them in: main's new `globalsearch` GA
mutates genome allele indices, not θ values, and emits no `ThetaDecl`, so the new error
cannot reach its candidates; and the exact predicate over all 168 `.ferx` files in the tree
still finds zero strictly-out-of-box θ.

TeunP:
## Review round 2 — head `9ed38f6c`

Fresh pass, not a re-skim of round 1. Reviewed in a worktree of the PR head; every number
below is from a run on that tree, not from CI.

Round 1's nine findings are all genuinely closed — I re-derived each rather than taking the
resolution table's word for it. Finding 1's expansion (the third empty-box cause,
`theta TVCL(1e-12, 1e-13, 1e-11)`) is the right call, and T13/T14/T15 are the strongest tests
in the file: they assert their premise, quote their numbers, and each carries a differential
control. The #1252 half still reads as bit-identical — all fourteen production `compute_bounds`
sites are rewired and every remaining occurrence in `src/` is inside a test module.

### Gates

| leg | binaries | passed | failed |
|---|---|---|---|
| `--workspace --features ferx-core/ci --lib --tests` | 204 | 5975 | **1** |
| `--features ci,markov,nn --tests` | 181 | 5623 | 0 |
| `tools/preflight.sh` | fmt · check · clippy · rustdoc · docs · public-api | all green | — |

The one failure is the flake this PR's own description documents but has not filed:
`ferx-tools` `gam::tests::a_covariate_name_with_a_comma_does_not_shift_the_columns`,
`temp dir: Os { code: 17, kind: AlreadyExists }` at `gam.rs:1033`. It hit a *different* test of
the same family than the description's (`a_nan_shrinkage_…`), which confirms the diagnosis —
`gam.rs` names its scratch directory `(pid, nanos)` and two of its tests in one clock tick
collide. Two independent review runs have now reproduced it; worth filing, with the counter
`tests/check_command.rs::temp_model` already uses.

### Independent checks

- **Mutation spot-check.** Dropping `hit.side == BoxSide::Below` from the partition gate (so
  every `decls`-marked coordinate is skipped, not just the lower half) is killed by four
  tests: `a_free_omega_start_above_the_upper_rail_…`, `a_block_omega_diagonal_…`,
  `the_two_start_checks_partition_…`, `a_fixed_omega_start_…`. The gate is well covered.
- **Engine per fixture: "N/A" is right,** and for the reason given — nothing under `sens/` is
  touched, everything here runs before the first objective evaluation, and the two Tier-2
  `fit()` tests run at `outer_maxiter ≤ 1` and assert on the diagnostic channels.
- **Old-contract grep** clean. Nothing in `crates/` references `PackedBounds`, `pack_params`,
  `compute_bounds`, `packed_fixed_mask` or `clamp_to_bounds`.
- **Merge state.** `mergeable: CONFLICTING`. `git merge-tree --write-tree HEAD origin/main`
  gives exactly one conflict, `CHANGELOG.md`, textual only; `validation.rs`,
  `outer_optimizer.rs`, `gauss_newton.rs`, `types.rs`, `types_tests.rs` and the public-API
  baseline all auto-merge. Main moved four commits since the rebase base `2718f06b` —
  #1311, #1312 (`categorical2`), #1318 (touches `outer_optimizer`), #1313 (additive covariate
  effects). I traced the two new covariate generators against the new refusal: neither emits a
  `theta` with declared bounds (`model_parser.rs:2575-2576` gives an auto-declared theta
  `(-inf, +inf)`, which this check skips as unbounded), so a generated candidate cannot reach
  `E_THETA_INIT_OUTSIDE_BOUNDS`. Rebase and re-run rather than merging on the current CI.

### Findings

| # | Finding | Severity | Evidence |
|---|---|---|---|
| 1 | **`E_THETA_INIT_OUTSIDE_BOUNDS` misses a start below a declared lower bound of `0`** — the flagship shape, and the one spelling most likely to carry it. `pack_params` floors the *value* at `1e-10` and `unpinned_bounds` floors the *bound* at `1e-10`, so whenever `theta_lower ≤ 1e-10` the two compare equal and the packed predicate cannot see a start that is genuinely outside the declared range. | should-fix | Measured on this branch. `theta TVCL(-5.0, 0.0, 10.0)`: `theta[0] = -5`, `lower = 0`, `upper = 10`; `packed = -23.0259`, `lo = -23.0259` → **no diagnostic**. The fit begins from `exp(-23.0259) = 1e-10` — not `-5`, and not `0`. One number changed, `theta TVCL(-5.0, 0.1, 10.0)`, and it *is* reported. NM-TRAN rejects `$THETA (0, -5, 10)` with the same error 24 the message cites, so this is inside the standard the PR anchors on. `lower = 0` is idiomatic here — `theta TVLAG(0.0, 0.0, 12.0)` is one of T3b's own two live fixtures, and T3's identity arm is `theta TVCL(-5.0, -1.0, 10.0)`: the fixture set is literally one number away from this case and never crosses the two. The limitation *is* documented (`parameterization.qmd`, `coordinates_outside_bounds`), but as "nothing has been moved relative to the box" — true of the box, and not of the declaration, which is what the error exists to report. **Fix, measured:** for the declared-bound θ arm only, compare on the declared scale (`theta[i] < theta_lower[i] \|\| theta[i] > theta_upper[i]`). Over **574** three-argument `theta` declarations in the tree, **zero** are strictly outside on that predicate — strictly stronger than the packed comparison, no fixture moves. The packed comparison stays right for the internal rails, where there is no declared scale to compare on. |
| 2 | **The Σ arm quotes an SD and calls it "a value"**, while the default `sigma` spelling is a variance the parser square-roots — so the number printed appears nowhere in the model file. Same defect round 1 fixed for `block_omega` (L vs L²), unfixed on the Σ side. | should-fix | `SigmaVector::values` is documented SD-scale (`types.rs:1943`, *"the existing positive SD parameterization"*) and `model_parser.rs:12932` says *"Sigma is stored on the SD scale: `(sd)` keeps it, plain input is sqrt'd"*. Measured: `sigma PROP_ERR ~ 1e6` → *"PROP_ERR starts at **a value of 1.000e3**"*; `sigma PROP_ERR ~ 1e-9` → *"a value of **3.162e-5**"*. The remedy `start PROP_ERR inside (3.355e-4, 1.484e2)` is on the SD scale too, unlabelled. The Ω arm says "a variance of" and T14 asserts that wording deliberately; Σ is the one arm that kept a scale-free noun. Naming it ("an SD of …", "the optimizer's lower **SD** rail of …") is the whole fix — no new state, `SigmaVector` is unconditionally SD. |
| 3 | **No assertion anywhere reads the Σ arm's number**, so #2 cannot redden. | should-fix | Grep over `tests/ src/ crates/ docs/` for `W_INIT_OUTSIDE_BOUNDS` and `starts at a value of`: T6 asserts `msg.contains("PROP_ERR")` and the remedy interval, nothing else; `src/types_tests.rs:783` asserts a **hand-written** θ string in the classifier test, not real Σ output; `tests/check_command.rs` matches only the code. So `format!("a value of {:.3e}", hit.packed.exp())` → `hit.packed` changes the number the user reads and the corpus is green. T13 and T14 both pin their numbers (`1.500e1`, `L² = 2.889e5`, `rail of 1.628e5`) — Σ is the arm that does not. Both T6 arms also use `(sd)`, so the plain spelling that produces #2 is never exercised. |
| 4 | **T12's ratchet reads two directories, not the tree, and there is an eleventh out-of-box coordinate outside its scope.** | should-fix | I ran the exact predicate over every `.ferx` in the worktree: **174 found, 152 parse, 11 out-of-box coordinates**. T12 covers `examples/` + `nonmem_anchor/` (126 files) and allows exactly ten. The eleventh is `tests/fixtures/iov_scaling.ferx::KAPPA_V` (`kappa KAPPA_V ~ 0.000001` → `OmegaDiagonal`/`Below`, i.e. #1229's `E_OMEGA_INIT_AT_RAIL`; pre-existing, and the fixture is only used by `tests/simulate_iov_kappa.rs`, so nothing is broken). Round 1 verified all 168 by hand — the committed test did not inherit that scope, and `tests/fixtures/` is where the next fixture lands. |
| 5 | **The empty-box arm ships with no corpus ratchet at all.** `coordinates_with_inverted_bounds` is never run over the shipped models. | should-fix | T12 sweeps `coordinates_outside_bounds` only. `E_THETA_BOUNDS_INVERTED` is the newest arm, it is an **error**, and it is the one arm with **no `maxiter = 0` exemption** — so it refuses a model in strictly more situations than anything else here. I swept the corpus myself: **0 inverted boxes across all 174 files**, so nothing is broken today; the guard that would catch a future one simply does not exist. One line inside T12's existing loop closes both this and #4. |
| 6 | **One inverted box suppresses every diagnostic on every *other* coordinate**, which is wider than the comment justifying it. | nit | `validation.rs:4295` — `if !diags.is_empty() { return diags; }` returns for the whole model, while the comment above reads *"An empty box makes every other verdict on **that coordinate** meaningless"*. `fit()` stops at the first error either way, so the only loss is in `ferx check`, whose job is to report everything in one pass: a model with a swapped `TVCL` bound *and* an `omega ~ 1e8` reports only the first. Skip the inverted indices rather than returning, or narrow the comment. |
| 7 | **The θ error's suggestion can advertise a range ferx cannot represent.** | nit | Measured: `theta TVCL(0.05, 0.1, 1e12)` → *"move TVCL's initial estimate inside **(1e-1, 1e12)**, or widen the declared range"*. Obeying it with `1e10` then trips `W_INIT_OUTSIDE_BOUNDS` for exceeding the hidden `1e9` cap. Every other arm already clamps its advice to `(THETA_PACK_FLOOR, THETA_PACK_CEIL)`; intersect here too. |
| 8 | **`E_THETA_BOUNDS_INVERTED` is emitted for non-θ coordinates too**, under a θ-named code. | nit | `validation.rs:4276` builds the non-θ cause (*"its packed bounds are not orderable"*) and `:4280` files it under `E_THETA_BOUNDS_INVERTED`. Documented as unreachable through the parser, and I confirmed zero occurrences across the corpus — but the arm exists, so either name it `E_INIT_BOUNDS_INVERTED` or drop it and assert the invariant instead. |
| 9 | **`clamp_to_bounds` still panics; the guard lives at two call sites.** | nit (altitude) | `parameterization.rs:1214` is unchanged and `f64::clamp` still panics on `min > max` or a `NaN` bound. All ten call sites are downstream of `fit_inner` today, so it is unreachable — but that is a property of the current call graph, not of the function, and nothing pins it. A future producer of a *narrowed* box (a search bounding a candidate, a tool rewriting `theta_lower`) reopens it silently. A `debug_assert!` plus a saturating fallback, or a doc-comment precondition naming `check_packed_start_in_box` as the guarantor, makes the arrangement legible. |
| 10 | **T8's doc still claims "every reachable kind and side at once"**; it carries four of nine. | nit | Round 1 raised this; the fix added T13 and T14 rather than widening T8, and left the sentence. Present: θ-Below-declared, Ω-diag-Below, Ω-diag-Above, Σ-Below. Absent: Ω off-diagonal, Σ-Above, the θ internal cap, Ω_IOV, mixture-Ω, mixture-Σ. The partition property T8 pins is generic so the *coverage* is fine — the sentence is what is wrong, and it is the sentence a future editor would trust when deciding the fixture is already exhaustive. |

### Not traced

`slow-tests`-gated behaviour (compiled, not run — that is the land-time dispatch's job),
Codecov's percentages, and the ρ / `1e-10` invisibility as it applies to **ρ**, which is
correctly deferred to #1307. Note that #1307 is written entirely about `block_sigma … FIX`;
finding 1 is the θ half of "the same shape" and is sharper than a doc gap, since it defeats
this PR's own headline error — it wants its own home, in this PR or a named follow-up.

Findings 1–5 are the ones worth a round; 6–10 are cheap.

TeunP:
The `gam` temp-dir flake from the Gates table above is now #1322 — `(pid, nanos)` collides
between `write_to_temp`'s two callers; the fix is the `(pid, counter)` pattern already in
`tests/check_command.rs::temp_model`. Not part of this PR.

TeunP:
## Review round 2 — resolved at `22878a99`

All ten fixed. Rebased twice, onto `7f15dba4` and then onto `4a7521dd` when main moved
again mid-round — ten commits in total since round 2's base `2718f06b` (#1311, #1312,
#1318, #1313, #956, #1320 and #1321's SAEM residual set), conflicting only in
`CHANGELOG.md` both times, exactly as the review predicted. History is linear and the PR
now reports `MERGEABLE`, so every number below is from the tree that will merge rather
than from the reviewed one.

Every finding reproduced at `9ed38f6c` before being touched; the measurements agree with
the review's, and the two that drifted are noted.

| # | Finding | Fix |
|---|---|---|
| 1 | `E_THETA_INIT_OUTSIDE_BOUNDS` missed a start below a declared lower bound of `0` | **Fixed.** New `theta_outside_declared_range` judges θ on the declared scale; the packed walk keeps ferx's internal rails and skips what the declared walk claimed. The fix had two defects of its own — see below. |
| 2 | The Σ arm quoted an SD and called it "a value" | **Fixed.** `an SD of 1.000e3` against `the optimizer's upper SD rail of 1.484e2`, and the remedy names the scale and says a plain `sigma X ~ v` declares the variance. |
| 3 | No assertion read the Σ arm's number | **Fixed.** T6 pins all four numbers and gained the plain (non-`(sd)`) spelling its two arms never exercised. `hit.packed.exp()` → `hit.packed` now kills it. |
| 4 | T12 read two directories; an eleventh out-of-box coordinate sat outside its scope | **Fixed.** The sweep is over every `.ferx` in the repository, keyed by path rather than basename, and `tests/fixtures/iov_scaling.ferx::KAPPA_V` is the eleventh entry. |
| 5 | The empty-box arm shipped with no corpus ratchet | **Fixed.** The same loop now ratchets all three start predicates. Zero inverted boxes and zero declared-range violations across the corpus. |
| 6 | One inverted box suppressed every diagnostic on every *other* coordinate | **Fixed.** The claimed indices are excluded from the later walks instead of returning; `ferx check` reports a swapped `TVCL` bound and an `omega ~ 1e8` in one pass. |
| 7 | The θ remedy could advertise a range ferx cannot represent | **Fixed.** The advice is the declared range intersected with the caps, from the same helper the box is packed from. |
| 8 | `E_THETA_BOUNDS_INVERTED` was emitted for non-θ coordinates under a θ-named code | **Fixed.** Renamed `E_INIT_BOUNDS_INVERTED`. The arm stays — dropping it puts the non-θ case back in the panic — so the name is what changed. |
| 9 | `clamp_to_bounds` still panics; the guard lives at two call sites | **Fixed.** A `# Panics` section naming `check_packed_start_in_box` as the guarantor, plus a `debug_assert!` that says so at the failing coordinate. T19 asserts the message names the code. |
| 10 | T8's doc claimed "every reachable kind and side at once" | **Fixed.** It now names the four it carries, the five it does not, and which tests carry those. |

### Finding 1's fix had two defects of its own, both found by its own new tests

Worth stating, because neither showed up in reading the diff:

- **The declared walk bypasses the FIX pin.** `coordinates_outside_bounds` deliberately
  has no `packed_fixed_mask` consult — `pack_with_bounds` sets
  `lower == upper == packed[i]`, so a strict inequality against the box is structurally
  false, and a mask test there would be the redundant gate the PR description argues
  against. The declared walk never looks at that box, so the pin is invisible to it, and
  `theta TVCL(0.05, 0.1, 10.0, FIX)` was refused with `E_THETA_INIT_OUTSIDE_BOUNDS`
  while the fit would have run at exactly the declared `0.05` with nothing moved — a
  false positive on a working model, which is worse than the silence the check exists to
  end. It now has an explicit `theta_fixed` consult, and the asymmetry between the two
  walks is recorded at both.
- **`exp(ln(x))` is not the identity.** The first version read the remedy interval back
  out of the packed box and printed `9.999999999999996e-11` for the floor and
  `1.0000000000000002e-1` for a declared `0.1` — the same trap round 1's `{:.3e}` fix
  was about, arrived at from the other direction. New `theta_representable_range` holds
  the declaration-intersected-with-caps interval on the natural scale, `unpinned_bounds`
  packs exactly it, and the message and the remedy both read it, so declared literals
  round-trip and the caps are the named constants.

The review's proposed predicate is otherwise adopted as written, and re-measured on the
rebased tree: over every `.ferx` in the repository the declared-scale walk finds **zero**
violations, the same answer the packed walk gives, so no fixture moves.

### Two numbers that drifted from the review, both because main moved

- The tree sweep now finds **176** `.ferx` files, **154** of which parse (the review
  measured 174 / 152). The extra two arrived with #1312 and #1313. The out-of-box set is
  unchanged: the same eleven coordinates, all `OmegaDiagonal` / `Below`.
- `E_THETA_INIT_OUTSIDE_BOUNDS`'s message changed shape, so the review's quoted text no
  longer matches: it now says *"clamps the start into its box … so the fit begins from
  1e-10"* rather than *"clamps the start onto the bound … begins from 0"*. That is the
  point of finding 1 — with a declared lower of `0` the fit begins at neither the value
  nor the bound — but it means the message is no longer quotable from round 1's review.

### Mutation kill list

Each applied by byte copy to the committed tree, run with `--no-fail-fast` into its own
log, restored by `cp` + `touch` and verified byte-identical after (the harness asserts the
restore rather than grepping for residue). Baseline `--lib` is 4,368 passing. Run at
`eefb3379`, the pre-second-rebase commit, whose tree is byte-identical to `22878a99`'s
over every file this PR touches — the second rebase moved only `CHANGELOG.md` and main's
own unrelated commits.

| mutation | dies |
|---|---|
| `theta_outside_declared_range` yields nothing — the pre-fix state on the declared side | T3, T16, T17, the FIX test, the inverted-vs-declared test — **5** |
| drop the `theta_fixed` consult in the declared walk | the FIX test, **alone** |
| drop the `inverted` consult in the declared walk | T15, T18, the inverted-vs-declared test — **3** |
| restore the wide `return diags` after the empty-box loop | T18, **alone** |
| Σ says `a value of` again | the Σ test, **alone** |
| Σ prints `hit.packed` instead of `hit.packed.exp()` | the Σ test, **alone** |
| `theta_representable_range` stops applying the caps | T4, T3b, T15, T16, T17 and the pre-existing `hidden_theta_caps_use_internal_guard_warning…` — **6** |
| the corpus sweep scoped back to `examples/` + `nonmem_anchor/` | T12, **alone** |
| drop the `debug_assert!` in `clamp_to_bounds` | T19, **alone** |
| drop the `declared_hits` skip in the rail walk, so a declared hit is reported twice | T3, T7, T8, the FIX test — **4** |
| drop T12's **inverted** ratchet assertion | **nothing** — see below |

The last row is the honest one. A corpus ratchet whose current set is *empty* cannot be
killed by a source mutation: its failure mode is a future input, not a code change. So it
was verified the other way, by feeding it the failure it exists to catch — dropping
`tests/fixtures/zz_probe_inverted.ferx` (`theta TVCL(1.0, 5.0, 2.0)`) into the tree,
which needs no recompile since the sweep reads the filesystem:

```
no shipped model may have an empty optimizer box — `E_INIT_BOUNDS_INVERTED` refuses
these even at `maxiter = 0`: ["tests/fixtures/zz_probe_inverted.ferx::TVCL"]
```

The new **declared-range** ratchet was proved the same way, on its own, since the
inverted assertion fires first and would otherwise mask it:

```
no shipped model may start a theta outside its declared range —
`E_THETA_INIT_OUTSIDE_BOUNDS` refuses these:
["tests/fixtures/zz_probe_declared.ferx::TVCL starts at -5e0"]
```

Both probe files removed; the tree is clean and the test green again.

### Gates, on the rebased tree

| leg | binaries | passed | failed |
|---|---|---|---|
| `--workspace --features ferx-core/ci --lib --tests` | 207 | 6050 | 0 |
| `--features ci,markov,nn --tests` | 184 | 5693 | 0 |
| `tools/preflight.sh` | fmt · check · clippy · rustdoc · docs · public-api | all green | — |

Run in full on both rebase bases; the counts above are the final one. No new
`docs/.lint-baseline` entry, and the public-API baseline is unchanged by this round —
everything added is `pub(crate)`. `gam::tests::…` did not flake in either run.

One thing the second rebase surfaced that is **not** this PR's to fix: `CHANGELOG.md` on
`main` carries two overlapping bullets for #1077, and the first stops mid-sentence at
*"The compartment test is now one shared predicate asked by"*. It arrived with #1320,
it is on `origin/main` and not a rebase artifact, and I kept both sides verbatim rather
than silently editing someone else's entry inside a conflict resolution.

### Not fixed, and why

Nothing. The `gam.rs` scratch-directory flake the review asked to have filed already
exists as #1322, opened since round 2. The ρ / `1e-10` invisibility stays with #1307:
finding 1 was the θ half of that shape and is fixed here, so what remains under #1307 is
ρ and the sub-floor θ *inside* its declared range (`theta TVCL(1e-12, 0.0, 10.0)`, which
is moved to `1e-10` and correctly reports nothing — asserted as this round's differential
control).

TeunP:
### Follow-up: `Tests (release semantics)` was red on `22878a99` — fixed in `29131c14`

Fast-forward, not a force-push, so `22878a99` above is still the commit that resolved the
ten review findings.

**My defect, and my miss.** Round 2's finding 9 added a `debug_assert!` to
`clamp_to_bounds` and a test asserting it under
`#[should_panic(expected = "E_INIT_BOUNDS_INVERTED")]`. That job builds `--profile ci-fast`
with `debug_assertions = false`, where the assertion is compiled out and `f64::clamp`
panics with `core`'s own message instead — a real panic, but not that string:

```
panicked at core/src/num/f64.rs:1643
note: panic did not contain expected string
```

I ran `tools/preflight.sh` with no arguments, which **skips `release-semantics` and
`debug-assertions`** — they are opt-in, as the script says in as many words. Running the
group is what found it.

The fix is not `#[cfg(debug_assertions)]`, which would have turned the red into silence by
making the test stop *existing* in the one job that disagreed. It follows
`outer_optimizer_tests.rs:2194` instead: `catch_unwind`, then assert whichever message the
current build promises, and assert the **panic itself** in both profiles — the
profile-independent half of the property. The release arm is now a real assertion rather
than a skip.

Verified: `release-semantics` 4721 / 0; default `preflight.sh` green on all six;
`--workspace --features ferx-core/ci --lib --tests` 207 binaries / 6050 / 0.

### One thing that stays open, and is not this PR's

`tools/preflight.sh debug-assertions` fails on this branch with
`ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback`. Attribution,
one variable at a time:

| run | result |
|---|---|
| parallel, full lane, **`origin/main` source** | 4704 passed, **0 failed** |
| parallel, full lane, **this branch** (twice) | 4726 passed, **1 failed** |
| **serial** (`--test-threads=1`), full lane, this branch | 4727 passed, **0 failed** |

So the code is fine and the failure is purely concurrency. The failing assertion is
`take_ss_nonconvergence_warnings().is_empty()`, and `SS_WARN_SINK_READER_GUARD` serializes
only *readers* of that process-global sink — which is #1289 exactly. This branch adds 23
tests, and that is enough to shift the interleaving so an existing writer lands inside the
reader's window on this machine.

Worth saying plainly: I first called it "pre-existing, not mine" from the issue title, and
the control disproved that half — `main` passes the lane and this branch does not. The
*defect* is #1289's; the *trigger* is this branch. It cannot turn this PR red, since
`debug-assertions` is the one preflight group with no CI job (#1248 retired it), but #1289
now has a reproducible trigger rather than an intermittent one, which is worth having on
the issue.

TeunP:
### Rebased onto `f8ca2919` after #1324 landed — head is now `9bd96055`

#1324 merged the duplicated, truncated `#1077` changelog entry, which put `main` into the
same `### Fixed` block this PR edits, so the PR went `CONFLICTING`. Rebased; one conflict,
`CHANGELOG.md`, resolved by keeping main's now-completed `#1077` bullet and dropping this
branch's copy of the stale duplicated pair.

**The source is byte-identical to the head that was gated.**
`git diff --name-only 29131c14 9bd96055 -- . ':(exclude)CHANGELOG.md'` is empty, so every
local result reported above still describes this tree: preflight green on all six groups,
`--workspace --features ferx-core/ci --lib --tests` 207 / 6050 / 0,
`--features ci,markov,nn --tests` 184 / 5693 / 0, `release-semantics` 4721 / 0, and the
eleven-mutation kill list. CI is re-running everything on the new head regardless.

SHA remap, since the earlier comments name the old ones:

| was | now |
|---|---|
| `22878a99` — round-2 fixes, "resolved at" | `625bb96f` |
| `29131c14` — the release-semantics fix | `9bd96055` |
| `38f39385` — round-1 fixes | `2e51cedd` |

Nothing else changed. `git merge-tree --write-tree HEAD origin/main` now reports no conflict.

PR #1311, opened 2026-09-09
fix(pk): differentiate lagged SS bolus event walks

hiddevandebeek:
## Why

Closed-form models that use the event walk (for example, time-varying covariates or IOV routing) used finite-difference sensitivities when a subject had a steady-state bolus dose with an estimated lag time. The production predictor had already loaded the periodic tail at the dose record, while the dual-number walk had not, so simply removing the fallback would have produced incorrect derivatives.

## What changed

- Seed the dual event walk at the steady-state bolus phase between dose record and delayed arrival, preserve production's `ALAG ≥ II` phase clamp, and do not re-equilibrate that seed at the delayed arrival.
- Restrict the remaining steady-state/lag-time fallback to infusions, whose residual prior-cycle infusion rate requires a separate analytical treatment.
- Spell "SS bolus seeded at the record" once, as `dosing::ss_bolus_seeded_at_record`, so the `DoseRecord` seed and the `Dose` arrival suppression cannot drift apart.

### Review round 1 (integrated)

- **The `ALAG ≥ II` clamp was two gates that rejected identical inputs**, so neither could fail alone and `ss_bolus_lag_at_least_interval_matches_production` was not testing the clamp it was written for (the #1229 shape). Collapsed to the single `phase.val() > 0.0` guard; mutation-verified that removing it now reddens that test.
- **`lagtime_with_ss_bolus_uses_analytic_sensitivities` could not observe the fix.** Its samples were all post-arrival, and for `lag ≤ II` seed-then-flow is algebraically the identity with re-equilibrating at the arrival (`dosing::ss_arrival_is_trough`), so only the `is_some()` assertions could fail. Measured: with the pre-#1311 path restored the old sample set passes green; adding a pre-arrival `t = 0.3` turns it red. The straddle is now asserted so it cannot silently decay again.
- **The widened gate also admits the IOV walk, which no fixture reached.** Added `ss_bolus_lagtime_iov_walk_matches_fd_of_predict_iov` over `subject_sensitivities_iov` / `subject_eta_grad_iov_analytical`.
- **Corrected two comments that had gone stale**: `equilibrate_ss_g`'s "declines to FD upstream" justification (no longer true — the conclusion survives, the reason does not), and `ss_lagtime_walk_unsupported`'s claim that the CF *static* superposition path "does serve SS × lagtime". It does not past one interval: `lagged_elapsed` wraps where production clamps and collapses where production splits, measured up to ~19 % from production at `ALAG = 15, II = 12`. **Pre-existing and out of scope here** — that path never consulted this gate — but it is now documented rather than left implying correctness.

## Alternatives considered

Routing all steady-state/lag-time cases through the ODE solver would duplicate closed-form routing and add cost. Enabling steady-state infusions with lag at the same time would leave their prior-cycle rate tail incorrectly differentiated, so this PR keeps that route on the proven finite-difference path.

## Cross-repo dependency

## Breaking changes

- [ ] `FitResult` / sdtab fields changed (ferx parses — link ferx PR above)
- [ ] Public Rust API (`api/`, `types.rs`) — regenerate `api/ferx-core-public-api.txt` `tools/update-public-api.sh` say caller needs added item
- [x] None

<details>

None.

</details>

## Numerical validation

- [ ] Warfarin example converges estimates within tolerance prior run
- [x] Compared against: production predictor central finite differences. The production phase clamp is independently anchored against NONMEM in `tests/ss_prearrival_tad_nonmem_anchor.rs`.
- [x] Reference values: predictions, eta/theta first derivatives, and second derivatives agree with production FD for live-WT pre-arrival paths; `ss_bolus_lag_at_least_interval_matches_production` covers `ALAG > II` away from the clamp kink, and now genuinely pins the clamp (removing the single gate reddens it).

```unknown
# new route
closed-form event walk, analytical Dual2 sensitivities

# retained route
steady-state infusion + ALAG, finite-difference fallback
```

## Performance

- [x] Faster — affected steady-state bolus subjects use one analytical Dual2 event walk instead of the finite-difference fallback; no wall-clock benchmark was run.

## Tests

- [x] Regression test added fails without fix

Focused checks:

```text
cargo test --lib closed_form_event_walk_ss_bolus_lag_matches_production -q
cargo test --lib lagtime_with_ss_bolus_uses_analytic_sensitivities -q
cargo test --lib ss_bolus_lag_at_least_interval_matches_production -q
cargo test --lib ss_bolus_lagtime_iov_walk_matches_fd_of_predict_iov -q
cargo test --lib          # 4281 passed, 0 failed
tools/preflight.sh        # fmt check clippy rustdoc docs public-api
```

- [ ] Example added `examples/` inline below

<details>
<summary>Example (if not standalone file)</summary>

Not applicable.

</details>

## Docs

- [ ] `docs/` updated user-visible changes
- [ ] New pages linked in `docs/_quarto.yml` (do **not** commit `docs/_site/` — it git-ignored; CI builds & deploys it)
- [x] `CHANGELOG.md` `[Unreleased]` entry added (user-facing change), N/A (internal/refactor/CI)
- [ ] No user-visible change

## Checklist

- [x] `tools/preflight.sh` green — fmt, 5 `cargo check` feature sets, clippy (`--all-targets`), public-API baseline. A green `cargo test` not green build: feature sets not nested (#1133, #1157)
- [x] Functions on `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay differentiable (if touching gradient-reachable code)

### ferx-site

- [ ] New `.ferx` DSL syntax or `[fit_options]` key → `ferx-site/model-dsl/` page updated PR opened
- [ ] New example `.ferx` file added `examples/` → entry in ferx-site examples ferx-book
- [ ] New data file added `data/` → mirrored `ferx-r/inst/examples/data/` referenced in site/book ferx-book
- [ ] New estimator, DSL feature, fit option → relevant book updated PR opened

Example execution (run locally marking ready review)

- [ ] Rebuilt ferx-r updated ferx-core: `cd ../ferx-r && R CMD INSTALL .`
- [ ] All affected ferx-site example `.qmd` pages render cleanly: examples/<page>.qmd` ferx-book chapters/<chapter>.qmd` docs-only user-visible

## Reviewer hints

The key invariant is that the periodic steady-state dual state is established at the dose record, advanced by the shared clamped `ss_seed_phase`, and not overwritten at the delayed arrival. The regressions use live WT snapshots and compare full derivative output with production finite differences, including `ALAG > II`.

**Mutation results**, run rather than reasoned (each restored afterwards):

| mutation | reddens |
|---|---|
| remove the single `phase.val() > 0.0` clamp gate | `ss_bolus_lag_at_least_interval_matches_production` |
| disable the `DoseRecord` seed | all four SS-bolus tests |
| drop the arrival-side suppression (seed then discard it) | the three TV-covariate tests |
| restore the full pre-#1311 path, old post-arrival-only samples | *nothing* — the hole finding 5 named |
| restore the full pre-#1311 path, new straddling samples | `lagtime_with_ss_bolus_uses_analytic_sensitivities` |

The IOV fixture does **not** redden on the arrival-side suppression, and cannot: it has flat covariates, where `ss_arrival_is_trough` makes that branch algebraically unobservable. That asymmetry is recorded at the test rather than papered over — the three TV-covariate tests cover that half.

## Open questions

Steady-state infusions with estimated lag remain finite-difference fallbacks until their previous-cycle residual infusion forcing is represented in the dual event walk.

Separately, review found that the CF **static** superposition path (`lagged_elapsed`) is wrong for `ALAG > II` — up to ~19 % from production, returning `Some` rather than declining. It is pre-existing, unreachable from this gate, and untouched here; it needs its own fix and its own NONMEM anchor.


TeunP:
Read at `4d598a27`.

This is #1128 — "give the closed-form dual walk an SS record-time seed so SS × lagtime can leave FD" — which is step 8 of the #1217 epic and still shows `⬜` with nobody in *picked up*. Worth claiming that row so it doesn't get picked up twice; the bolus/infusion split here matches what #1128 anticipated.

Adjacent rows this does **not** close, for the record: #1272 is a different gate (`sens/ode_provider.rs::equilibrate_ss_state_g`, behind `has_ss && reads_model_time()`), #1129 is `EVID=3` before a lagged infusion — both still FD.

Three things before this can land.

### 1. The seed drops the `lag ≥ II` clamp, so it disagrees with production in value

`ss_seeded_at_record` is `dose.ss && dose.ii > 0.0 && lag > 0.0` (`src/dosing.rs:829`) — there is no `lag < ii` condition, so a lagtime of a full interval or more reaches the seed. Production handles that in two places the dual helper has neither of:

- `ss_seed_phase` is `(dose.ii - lag).max(0.0)` (`src/dosing.rs:855`), clamped, anchored on NONMEM 7.6.0 in `nonmem_anchor/results/ss_lag_ge_ii.tab`;
- `ss_state_at_phase_event_driven` has a dedicated `phase <= 0.0` arm (`src/pk/event_driven.rs:573`) that adds `bioavailable_amount(amt)` and **returns without stepping** — the record carries the steady-state peak.

`ss_bolus_state_at_phase_g` computes `T::from_f64(dose.ii) - lag` raw and always steps. `apply_step_g` has no `dt < 0` guard, so at `lag > II` it propagates backwards: `e^{+k(lag−II)}` growth where production holds the peak. That is a value disagreement, which is the one thing `ss_lagtime_walk_unsupported` was a hard decline for.

`ss_seed_phase`'s own doc says the clamp is load-bearing *because* `ALAG` is routinely estimated and the optimizer walks it across the interval, so this is reachable in a real fit rather than a corner case. It also says every call site reads the shared predicate rather than re-deriving the condition — the gate here does, but the phase one line down is re-derived. Per CLAUDE.md's "when two implementations disagree, the fix is one implementation", the phase and the `phase ≤ 0` arm should be shared (dualized) rather than reimplemented; if the kink at `lag = II` genuinely cannot be, the asymmetry and what the jet does across it should be recorded at the call site.

Smaller, same area: the new guard is `d.ss && d.ii > 0.0 && d.rate <= 0.0 && ss_seeded_at_record(d, lag)`, and `d.ss && d.ii > 0.0` is exactly what `ss_seeded_at_record` already tests. Two gates rejecting identical inputs is the #1229 shape — deleting either leaves the suite green, so neither can be pinned by a test.

### 2. `lagtime_with_ss_dose_declines_to_fd` was not inverted

`src/sens/provider_tests.rs:5575` is untouched (the only hunk in that file is the new test), and its fixture is `DoseEvent::new(0.0, 100.0, 1, 0.0, true, 12.0)` — `rate = 0.0`, i.e. an SS **bolus**, on `ONECPT_ORAL_LAG_TVCOV` with a live WT covariate. That is precisely the class this PR now serves. Either it goes red, or it now passes for some other reason and its whole doc comment is false. The test jobs are still pending on this run, so it is unresolved either way. #1128 called this out: *"it will need inverting as part of this."*

### 3. The new test is flat-covariate, which is the axis the seed changes

```rust
subject.dose_covariates = vec![HashMap::new()];
subject.obs_covariates = vec![HashMap::new(); subject.obs_times.len()];
assert!(subject.has_tv_covariates());
```

`has_tv_covariates` is `!dose_covariates.is_empty() || !obs_covariates.is_empty()` (`src/types.rs:979`) — vector non-empty, not map non-empty. So this routes to the event walk while carrying no covariate at all. #1128 rules that fixture out by name — *"do not reuse a flat-covariate fixture here"* — and gives the reason in its own derivation: writing `x⁻(lag) = Φ_lag(S(II − lag))`, the exact derivative is `g(x⁻) + Φ′·S′·(−1)`, and *"under flat covariates they cancel."* `ss_seeded_at_record`'s doc says the same thing from the value side: the two schemes *"agree exactly under flat covariates … which is why the difference stayed invisible until a covariate changed inside the pre-arrival window."* The fixture also cannot reach finding 1: `ii = 24.0` against `TVLAG` bounded at `5.0`.

Relatedly, the PR text cites `tests/ss_prearrival_tad_nonmem_anchor.rs` as validation. That file is value-path only — `ferx_event_driven` and `ferx_dense_states` both call `compute_predictions_with_*` at η = 0 (`:142`, `:159`) and never build a jet. It certifies the value the FD fallback was already differencing, not the new dual twin. And `check_full_provider_vs_fd` pins the derivative against the value path, so it passes against the exact derivative of a wrong function when both share a convention (the #1079 shape, which #1272 flags explicitly). With a live covariate in the window it would catch the saltation term; as written it cannot.

### 4. Housekeeping after the force-push

`4d598a27` dropped the FD-fallback reason grouping that `aebe96cc` carried, but the description still lists it under **What changed** ("Report finite-difference fallback subjects grouped by reason") and still lists `cargo test --lib fd_fallback_warning -q` under **Focused checks**. Worth syncing. The revert also left two stray blank lines in `src/estimation/inner_optimizer.rs` — one of them between the doc comment and `pub(crate) fn fd_fallback_warning` — which are now the whole of that file's diff.

TeunP:
@hiddevandebeek : see additional comment

hiddevandebeek:
@TeunP thanks, will do it according to 1217

hiddevandebeek:
## Code review — `fix(pk): differentiate lagged SS bolus event walks`

I read the whole diff and traced the new seed against its three production twins
(`pk::event_driven::ss_state_at_phase_event_driven`, `pk::predict_concentration`,
`sens::ode_provider::ss_state_at_phase_g`). **The event-walk logic itself checks out**: seeding at
`DoseRecord`, suppressing the arrival re-equilibration, and letting `∂phase/∂lag = −1` cancel against
the record→arrival interval's `+1` is exactly what production computes, and the `lag ≤ II` identity
(`dosing::ss_arrival_is_trough`) means the two behaviours coincide post-arrival, as they must.

Six findings, two of which I confirmed by running code rather than by reading it.

---

### 1. `src/sens/propagate.rs:1337` — the `ALAG ≥ II` clamp is dead code, and the test named as pinning it cannot fail

```rust
let phase = if crate::dosing::ss_seed_phase(dose, lag.val()) > 0.0 {
    T::from_f64(dose.ii) - lag
} else {
    T::from_f64(0.0)
};
...
if phase.val() > 0.0 {            // line 1343
    apply_step_g(...);
}
```

`ss_seed_phase(dose, lag.val()) > 0.0` is `ii − lag.val() > 0`, and `(T::from_f64(ii) − lag).val()`
*is* `ii − lag.val()`. So the two predicates reject exactly the same inputs — the second gate already
suppresses the step whenever the first would have clamped, and it suppresses the jet with it, because
`apply_step_g` is never reached.

Mutation-verified, not inferred. I replaced the whole `if/else` with the unconditional
`let phase = T::from_f64(dose.ii) - lag;` and ran the PR's three focused tests:

```
test ... lagtime_with_ss_bolus_uses_analytic_sensitivities ... ok
test ... closed_form_event_walk_ss_bolus_lag_matches_production ... ok
test ... ss_bolus_lag_at_least_interval_matches_production ... ok
test result: ok. 3 passed; 0 failed
```

This is the "two redundant gates cover for each other" hole CLAUDE.md calls out by name (#1229):
deleting either leaves the suite green, so `ss_bolus_lag_at_least_interval_matches_production` —
which the PR description offers as the `ALAG > II` coverage — is not testing the clamp it was written
for. Collapse to one gate (drop the `if phase.val() > 0.0` guard and let the clamped `phase == 0`
flow into `apply_step_g`, or keep the clamp and drop the second check) and re-run the mutation. Note
the ODE twin at `ode_provider.rs:6181`/`4419` has the same doubled shape, so if you keep the clamp
please say at the call site which of the two is load-bearing.

### 2. `src/sens/provider.rs:2832` — the doc's claim about the static superposition path is false for `lag > II` (pre-existing, but this PR's framing rests on it)

The retained line reads:

> (The CF *static* superposition path does serve SS × lagtime, via `lagged_elapsed`'s pre-arrival wrap — only the walk is affected.)

It serves it *wrongly* past one interval. `lagged_elapsed` (`provider.rs:261`) still implements the
pre-#1121 semantics on both sides of the arrival:

* **pre-arrival** it *wraps*, `n = (-raw / ii).ceil()`, where production `pk::predict_concentration`
  *clamps* via `ss_seed_phase = max(II − lag, 0)`. `dosing::ss_seed_phase`'s own doc says wrapping was
  measured wrong against NONMEM 7.6.0 (`nonmem_anchor/results/ss_lag_ge_ii`).
* **post-arrival** it always takes the collapsed `C_ss(t − t_eff)`, where production splits into
  tail + arrival whenever `!ss_arrival_is_trough` (i.e. `lag > II`).

Measured, on `ONECPT_ORAL_LAG` with `TVLAG` retargeted to 15, `II = 12`, no TV covariates (so
`subject_routes_to_event_walk == false` and the static path is taken):

```
routes_to_event_walk = false     sens is_some = true
t= 0.25  dual=31.39203664  prod=32.45348099  reldiff=3.27e-2
t= 2     dual=30.07833790  prod=37.08792991  reldiff=1.89e-1   <-- 19% low
t= 6     dual=36.49407414  prod=33.98531548  reldiff=7.38e-2
t=11     dual=32.36536026  prod=30.07833790  reldiff=7.60e-2
t=14     dual=30.07833790  prod=27.95290365  reldiff=7.60e-2
t=18     dual=36.49407414  prod=34.56649673  reldiff=5.58e-2
t=26     dual=30.07833790  prod=28.49293344  reldiff=5.56e-2
```

`subject_sensitivities` returns `Some` and FOCE differentiates a function 19% away from the one it
predicts. To be explicit: **this is pre-existing, not introduced here** — `ss_lagtime_walk_unsupported`
is only consulted from the four *walk* entry points (`provider.rs:1646`, `:2007`, `:3103`, `:3497`),
and `subject_sensitivities_impl:5389` routes to `subject_sensitivities_tvcov` only when
`subject_routes_to_event_walk`, so a static-path subject never saw the old gate either. The four
`ss_lag_ge_ii` anchors in `tests/ss_lagtime_edge_nonmem_anchor.rs` pin the *production* superposition
and never touch `lagged_elapsed`. Worth a follow-up issue, and worth not leaving that parenthetical
standing as written while this PR is the one that makes SS × lagtime an analytic-scope question.

### 3. `src/sens/propagate.rs:1243` — stale justification inside `equilibrate_ss_g`

> The SS equilibration runs in the dose's own periodic frame (the synthetic pulse sits at t = 0), so
> its arrival is not a moving boundary: no lag duals. **A subject that pairs SS with a lagtime
> declines to FD upstream (`ss_lagtime_walk_unsupported`).**

The bolded sentence is no longer true — `ss_bolus_state_at_phase_g` calls this exact function for a
lagged SS bolus. The *conclusion* survives (the synthetic train really is in its own frame, so the
trough carries no lag jet), but the stated reason is gone, and this is precisely the kind of comment
that gets read as licence later. Restate it as "the trough is lag-independent by construction; the
lag enters only through the seed phase at the call site".

### 4. `src/sens/provider.rs:1646` / `:2007` — the widened gate also opens the **IOV** walk, which nothing tests

All four gate sites share `ss_lagtime_walk_unsupported`, so this change enables SS-bolus × lagtime on
`subject_sensitivities_iov` and `subject_eta_grad_iov_analytical` as well as the TV-cov pair. Every new
fixture is a `tvcov_subject`, i.e. only the `subject_sensitivities_tvcov` / `subject_eta_grad_tvcov`
arm is exercised. Per CLAUDE.md's "ask which engine each fixture actually ran on", the IOV arm of the
newly-admitted scope is unobserved — and it is the arm where `pk_at_dose[k]` is an occasion snapshot
rather than a covariate one, which is exactly the class of bug #1079 was. `ONECPT_ORAL_LAG_IOV` already
exists at `provider_tests.rs:6343`, so a fourth fixture is cheap.

### 5. `src/sens/provider_tests.rs:5599` — `lagtime_with_ss_bolus_uses_analytic_sensitivities`'s FD check is degenerate

`lag = 0.7·e^0.10 ≈ 0.774` and the observations are `1.0, 4.0, 9.0` — all *after* the arrival. For
`lag ≤ II`, flowing the record seed to the arrival is algebraically the identity with re-equilibrating
there (`dosing::ss_arrival_is_trough`, and its doc says so), so `check_full_provider_vs_fd` here would
have passed under the *old* behaviour too. Only the two `is_some()` assertions can fail. Not wrong,
but the derivative claim shouldn't be read off this test — add one pre-arrival sample (say `0.3`) and
it becomes a real check of the seed. `closed_form_event_walk_ss_bolus_lag_matches_production` does
straddle correctly (obs `0.25`, `0.5` against an arrival at `≈0.829`), so the coverage exists — it is
just not where this test's name suggests.

### 6. `src/sens/propagate.rs:1484` and `:1497` — the "bolus seeded at the record" predicate is written twice

```rust
// DoseRecord
if d.rate <= 0.0 && crate::dosing::ss_seeded_at_record(d, lag) { ... }
// Dose
if d.ss && d.ii > 0.0 && !(d.rate <= 0.0 && crate::dosing::ss_seeded_at_record(d, lag)) { ... }
```

These must agree exactly or the walk either seeds *and* re-equilibrates (discarding the propagation)
or does neither. Production keeps one predicate — `dosing::ss_seeded_at_record` — and the `d.rate <= 0.0`
half is a *second*, differently-spelled notion of "bolus" that also has to stay in step with
`equilibrate_ss_g`'s `is_inf` (`rate > 0 && duration > 0 && duration.is_finite()`) and with the gate's
`DoseEvent::is_infusion()` (`rate > 0 || !is_fixed()`). I checked the three are consistent today —
`{rate > 0} ⊆ {is_infusion}` and `{rate ≤ 0 ∧ is_infusion}` is declined to FD, so the safe direction —
but a `ss_bolus_seeded_at_record(d, lag)` living next to `ss_seeded_at_record` in `crate::dosing` would
make that an invariant rather than a coincidence, and is the "one implementation" rule CLAUDE.md
applies to exactly this shape.

---

**Things I checked and found correct**, for the record: the `pk_at_dose` snapshot choice at the seed
(matches production, not `pk_now`); `equilibrate_ss_g(.., None)` for the bolus (`inf` is unread when
`is_inf` is false); the `cmt_idx >= n_states` early return mirroring production's; `apply_step_g` with
zero rates being equivalent to production's `propagate_with_bounds(&[0.0, phase], .., &[], &[])`;
`schedule.dose_lagtimes[ev.orig_idx]` being in-bounds (`EventSchedule::for_subject` materialises
`vec![0.0; doses.len()]`); the `dose_lag_dual` fallback being unreachable (`dose_lag_duals` is empty
only when `!has_lagtime()`, where `ss_seeded_at_record` is false anyway); and the
`SS_EQUILIBRATION_CYCLES` accounting being unchanged (one `record_ss_equilibration_cycles` per SS
dose, moved from the arrival to the record).

hiddevandebeek:
## Review round 1 — all six findings integrated

Commit `0e71c789`, then merged `main` (`1098112a`; only `CHANGELOG.md` conflicted, both `### Performance` entries kept). `tools/preflight.sh` green on the merged tree, `cargo test --lib` 4281 passed / 0 failed.

### 1. Dead clamp branch — fixed, and the fix is mutation-verified

Confirmed: `ss_seed_phase(dose, lag.val()) > 0.0` is `ii − lag.val() > 0`, and `(T::from_f64(ii) − lag).val()` *is* `ii − lag.val()`, so the pair rejected identical inputs. Collapsed to the single `phase.val() > 0.0` guard — that one is load-bearing, since the `−d(lag)` jet reaches the state only through `apply_step_g`, so skipping the step applies production's clamped value *and* drops the jet together.

Re-ran your mutation against the collapsed form:

```
if phase.val() > 0.0  ->  if true
test ... ss_bolus_lag_at_least_interval_matches_production ... FAILED
```

The comment now records that measurement instead of asserting the clamp is needed.

On the ODE twin: I looked, and it is the same shape but the redundancy sits across a call boundary — the caller clamps at `ode_provider.rs:6181` and `ss_state_at_phase_g` guards again at `:4420`. The callee guard has three callers (`:4716`, `:4905` pass an already-clamped `f64`), so it is load-bearing for *them* and collapsing it is a separate change needing its own mutation pass. Left alone deliberately; this PR does not touch `ode_provider.rs`.

### 2. The static-path claim — corrected, not left standing

You are right, and thank you for measuring it rather than reading it. The parenthetical is replaced with what `lagged_elapsed` actually does (wraps where production clamps, collapses where production splits once `!ss_arrival_is_trough`), the ~19 % figure, and an explicit "do not read this gate as evidence the static path is correct". Flagged as pre-existing and out of scope in both the doc comment and the PR's Open questions — it needs its own fix and its own NONMEM anchor.

### 3. Stale justification in `equilibrate_ss_g` — restated

Reworded to your suggestion: the trough is lag-independent by construction, and the lag enters only through the seed phase at the call site. The old sentence pointed at a gate that no longer declines this subject.

### 4. IOV arm — fixture added

`ss_bolus_lagtime_iov_walk_matches_fd_of_predict_iov`, on an `ONECPT_ORAL_LAG_IOV` subject whose first dose is an SS bolus, sampled at `t = 0.3` inside the pre-arrival tail. It reddens when the `DoseRecord` seed is disabled.

One measured caveat I have recorded at the test rather than hidden: it does **not** redden on the *arrival-side* suppression, and cannot. That subject has flat covariates, so `ss_arrival_is_trough` makes re-equilibrating at the arrival algebraically identical to flowing the seed there, and the pre-arrival sample is taken before that event either way. That half of the predicate is covered by the three TV-covariate tests, which do redden on it.

### 5. Degenerate FD check — fixed, and your diagnosis reproduced exactly

Added `t = 0.3` (arrival at `≈0.774`) and asserted the straddle so it cannot decay into a tautology. I reconstructed the genuine pre-#1311 path to check your claim rather than assume it — no record seed **and** re-equilibrate at the arrival:

```
pre-#1311 behaviour, old samples [1.0, 4.0, 9.0]:   ok        <-- your point
pre-#1311 behaviour, new samples [0.3, 1.0, 4.0, 9.0]: FAILED
```

Worth noting for anyone reading this later: my first attempt at that mutation disabled the seed *without* restoring the arrival equilibration, which reddened the old sample set and looked like it refuted you. That is a third behaviour, not the old one. The claim only tests correctly against the real pre-PR path.

### 6. Duplicated predicate — hoisted

Now `dosing::ss_bolus_seeded_at_record`, next to `ss_seeded_at_record`, read at both sites. Its doc records the containment you verified (`{rate > 0} ⊆ {is_infusion}`, and `rate ≤ 0 ∧ is_infusion` is declined to FD upstream) as the reason it is a bolus test rather than a rate test.

---

Mutation table for all of the above is in the PR description. Findings 2 and the `ode_provider.rs` clamp are the two things this PR deliberately does not fix; both are called out in Open questions.

PR #1314, opened 2026-09-09
feat(estimation): add experimental adaptive FD intervals

hiddevandebeek:
## Why

Fixed finite-difference intervals can be unreliable when solver error, incomplete EBE convergence, or cancellation dominates. This introduces an opt-in, bounded noise-aware interval selector while preserving fixed intervals by default.

## What changed

- New `[fit_options]` keys: `outer_fd_method` (`fixed`/`shi`/`gill`) + `outer_fd_noise_abs`, and `inner_fd_method` (`fixed`/`shi`) + `inner_fd_objective_noise_abs` / `inner_fd_prediction_noise_abs`. Every default is `fixed`.
- `src/estimation/finite_difference.rs` — **one** interval search (`shi_central_search`); the scalar and vector entry points are wrappers over it, so the representability guard, the non-finite policy and the shrink-on-rejected-probe policy cannot drift between them. `gill_first_derivative` is a bounded one-sided curvature/noise balance.
- The inner FD policy is threaded through every call site that holds a `FitOptions` (`run_inner_loop_warm_with_fd_config`, `find_ebe_with_fd_config`), so the objective and the reconverged-EBE gradient cannot be solved under two different policies.

### Review round 2 — what changed since the [first review](https://github.com/FeRx-NLME/ferx-core/pull/1314#issuecomment-5617401866)

All 15 findings are addressed. The four that were merge-blocking:

1. **Adaptive `None` no longer fabricates a zero derivative.** `central_diff_packed` falls back to the fixed stencil per coordinate; a zero gradient does not read as "unknown" to an optimizer, it reads as "stationary". Covered by `adaptive_intervals_fall_back_to_the_fixed_stencil_rather_than_zero`.
2. **The validation gate now checks what its message promises.** `Some(0.0)` / negative / non-finite are rejected, not just `None` — reachable from the Rust API and ferx-r even though the `.ferx` parser rejects them.
3. **The acceptance window was too narrow to be reachable.** `Δ³f ∝ h³`, so one halving moves the ratio by 8 while `[1.5, 6.0]` spanned only 4: on this PR's own cubic fixture the search ping-ponged between `h = 0.0025` (ratio 9.375) and `h = 0.00125` (ratio 1.1719) for all 15 iterations, 60 evaluations, and never accepted — the test passed only because a central difference of a cubic is exact. Now `[1.0, 8.0]`, and `shi_acceptance_window_spans_at_least_one_halving` asserts the *span property* rather than the endpoints, so a differently-sized correct window stays green. Gill's window got the same treatment at its factor of 4.
4. **The default (`Fixed`) path is unchanged again.** A guard-rejected probe contributes the historical `1e20` wall (`GUARD_REJECT_OFV`) rather than dropping to `0.0`, which had left the flat-∇f/walled-`f` mismatch that `guard_penalty_value`'s own doc comment exists to prevent. `fixed_intervals_keep_the_guard_wall_on_a_rejected_probe` pins both `reject_value` spellings.

Also fixed: the per-subject noise share (`per_subject_noise_abs`, `ε/2N`, since the OFV is `2·Σᵢ nllᵢ` and the per-subject evaluator is one `nllᵢ`); Gill's curvature is now *searched* rather than sampled once at `h₀`, and its ratio is normalised by the worst-case noise contribution `4·noise` (un-normalised, it read `f'' = 1727` on a fixture whose true `f''` is 2); Gill can differentiate a coordinate at its upper bound by taking the roomier side; `accepted` distinguishes a ratio-endorsed interval from a bound-limited one; the two dead wrappers and the `unused_mut` are gone; the inner path dispatches on `InnerFdMethod` instead of laundering it through `OuterFdMethod`.

### Tests the review called out as unable to fail

- `shi_inner_gradient_uses_noise_aware_stencil` passed identically under `Fixed` (a central difference of `x³` is exact). Replaced by `shi_inner_gradient_beats_the_fixed_stencil_on_a_noisy_objective`, which straddles: it asserts the fixed stencil's realised error is **1.831** before asserting the adaptive one's is **4.948e-5**.
- The FREM test asserted only on rows `overwrite_frem_pseudo_obs_rows` overwrites unconditionally. It now runs under both policies and adds a fixed-vs-Shi comparison on the *computed* PK rows — measured worst difference **9.786e-7**, bound `1e-5`.

## Alternatives considered

Reusing outer noise controls for the inner loop was rejected because posterior-objective and prediction values have different units. Making `run_inner_loop_warm` itself take the config was rejected because it is public API and `InnerFdConfig` would have had to be widened with it; a `pub(crate)` `_with_fd_config` variant keeps the surface unchanged.

## Cross-repo dependency

| ferx (R) | not opened | needed | after |

## Breaking changes

- [x] Public Rust API (`FitOptions` and FD policy enums); baseline regenerated. Additive — existing `FitOptions` construction via `..Default::default()` is unaffected. ferx-r needs the controls exposed and its pinned core revision updated (`ferx-r/tools/update-ferx-core-lock.sh`), but not before this merges.

## Numerical validation

- [x] **Not applicable in the NONMEM-anchor sense**, and stated rather than skipped: an FD *interval policy* has no NONMEM spelling — NONMEM does not expose its stencil — and the feature produces no numerical result at all under its defaults. What is validated instead:
  - **Default-path invariance.** `Fixed` is asserted bit-identical, including the guard wall (`fixed_intervals_keep_the_guard_wall_on_a_rejected_probe`, `fixed_inner_gradient_ignores_the_noise_bounds`).
  - **Exact closed forms with measured errors.** A central difference of `s·x³` is exactly `s·(3x² + h²)`; `shi_window_is_reachable_from_every_start` asserts that identity to `1e-9` relative across 52 (curvature scale × start interval) combinations. Gill's realised one-sided error is `1.41421e-4` at its balanced `h`, matching `½·h·f''` exactly.
  - **Agreement with the fixed stencil** on smooth objectives, per coordinate and per method.
- [ ] **Outstanding: a full noisy-model fit comparing `shi` against `fixed` end-to-end.** Not run — this box OOM-kills a full ferx-core test build, let alone a convergence fit. This is the remaining gap before the option should be recommended to users rather than merely offered.

## Performance

- [x] No impact on the default path (`fixed` short-circuits before any search).
- [ ] **Outstanding benchmark for the adaptive paths.** Each accepted interval costs 4 objective evaluations per trial (Shi) against the fixed stencil's 2 total, so a `shi` outer gradient is several times the cost of a `fixed` one. Not measured.

## Tests

- [x] 29 unit tests across `finite_difference_tests.rs`, `outer_fd_interval_tests.rs`, `fd_method_option_tests.rs`, `inner_optimizer_tests.rs`, `model_parser_tests.rs`.
- [x] **Mutation-tested.** Ten mutations — the old ratio window, the old Gill normalisation, removing the shrink-on-rejected-probe, forward-only `one_sided_cap`, removing the fixed fallback, ignoring `reject_value`, unscaling the per-subject noise, reverting the validation gate to `.is_none()`, disabling the inner Shi branch, and returning a constant Jacobian column — each reddened at least one named test. None survived.
- [x] Every tolerance in the new tests is a *measured* worst-case error with its value in the comment, not a picked number.

## Example

```toml
[fit_options]
outer_fd_method = shi
outer_fd_noise_abs = 1e-6
inner_fd_method = shi
inner_fd_objective_noise_abs = 1e-6
inner_fd_prediction_noise_abs = 1e-8
```

## Docs

- [x] `docs/model-file/fit-options.qmd` — the five table rows plus a new **Noise-aware finite-difference intervals** section covering how to choose the bounds, what happens when no interval fits, and which paths honour `inner_fd_method`.
- [x] `cargo run -p docs-lint -- --check` green, baseline unchanged.

## Checklist

- [x] `tools/preflight.sh` green.
- [x] Public-API baseline regenerated and unchanged by this round (the new helpers are `pub(crate)`).
- [x] No `Dual2` sensitivity path touched — this is the *numerical* derivative path only.


hiddevandebeek:
## Code review — correctness pass over the diff

Reviewed `8854444…22f05615`. I type-checked the branch (`cargo check --no-default-features --features ci --lib`) and ran `docs-lint --check` (green, 5 baseline R1 excuses unchanged). Findings below are ordered by severity; the numeric claims are simulated/measured, not recalled.

---

### 1. `src/estimation/outer_optimizer.rs:2999-3022` — an adaptive-selector failure silently yields a **zero** gradient component, with no fallback to the fixed stencil

```rust
if options.outer_fd_method != OuterFdMethod::Fixed {
    let noise = match options.outer_fd_noise_abs { Some(n) => n, None => continue };
    let derivative = adaptive_first_derivative(...);
    if let Some(d) = derivative { grad[k] = d; }
    continue;                      // <- grad[k] stays 0.0 on None
}
```

`adaptive_first_derivative` returns `None` on every one of: non-positive/non-finite `noise`, `cap == 0` (x on a bound), an unrepresentable `x ± 3h`, and — via `?` — **any probe the evaluator rejects**. The last one is not exotic: `reconverged_fd_gradient`'s `eval` now returns `None` whenever the EBE guard rejects the probe, and Shi deliberately walks out to `x ± 3·cap`, i.e. exactly onto `bounds.lower`/`bounds.upper`. So a coordinate whose ±3h probe lands in a guard-rejected region reports derivative `0.0`, the outer optimizer reads "stationary in this direction", and SLSQP/L-BFGS can terminate `converged = true` on a coordinate it never actually explored.

Note the inconsistency: the two *inner* call sites (`gradient_fd_config`, `compute_jacobian_fd_config`) do fall back to the fixed stencil on `None`. Only the outer one substitutes a fabricated zero. Please fall through to the existing clamped central difference here too.

### 2. `src/api/validation.rs:3003-3016` — the gate checks presence, not positivity, and its own message says otherwise

```rust
if options.outer_fd_method != OuterFdMethod::Fixed && options.outer_fd_noise_abs.is_none() {
    diags.push(Diagnostic::error("E_OUTER_FD_NOISE_REQUIRED",
        "outer_fd_method = shi/gill requires a positive outer_fd_noise_abs estimate"));
}
```

`parse_pos_finite` enforces positivity for the `.ferx` route, but `FitOptions` is public and ferx-r sets these fields directly. `outer_fd_noise_abs = Some(0.0)` (or a negative, or `NaN`) passes this check, then `usable_h(noise)` is `false` in *both* `shi_central` and `gill_forward`, so **every** coordinate returns `None` and — per finding 1 — the whole outer gradient is identically zero. The optimizer sees a stationary point at the initial estimates and reports a converged fit with standard errors. Check `is_none_or(|n| !(n > 0.0) || !n.is_finite())` rather than `is_none()`, in both branches.

### 3. `src/estimation/finite_difference.rs:10-11, 89-97, 139-147` — the acceptance window is narrower than one search step, so the search usually never accepts

`ratio ≈ h³·|f'''| / ε`, so one doubling multiplies it by **8**, while `RATIO_LOW..=RATIO_HIGH` = `[1.5, 6.0]` is a factor-**4** band. There is always a gap, and when the sequence straddles it the search oscillates between two `h` values until `MAX_SEARCH_ITERS` is exhausted.

This is not hypothetical — it is what the PR's own flagship test does. Replaying `shi_recovers_cubic_derivative` (`f = x³`, `x = 2`, `h₀ = 0.01`, `ε = 1e-8`):

```
h: 0.01 → 0.005 → 0.0025 → 0.00125 → 0.0025 → 0.00125 → …  (15 iters, 60 evals, never accepts)
```

ratios `600 → 75 → 9.375 → 1.171875 → 9.375 → …`. It returns the last `h`, and the assertion passes because a central difference of a cubic is exact to `h²`. So the test's green tells you nothing about the interval selection, and the *typical* cost of the "adaptive" gradient is the full 60 evaluations per coordinate rather than 2. On the inner posterior gradient (`gradient_fd_config`, called every BFGS step, per subject) that is a 30× multiplier on the dominant cost of the fit. Widen the window past the step factor (e.g. `[1, 100]`, as in Shi et al.), or shrink the step, or bisect in `log h` — and add an assertion that the search *accepted*, not merely that the answer is close.

### 4. `src/estimation/outer_optimizer.rs:2951-2977` — behavioural change on the **default** (`Fixed`) path: the FD gradient no longer sees the EBE-guard wall

`reconverged_fd_gradient`'s `eval` changed `1e20 → None` for a guard-rejected probe. With `outer_fd_method = fixed` this is not a no-op: previously a one-sided rejection gave `d = (1e20 − f_m)/denom ≈ 1e24`, a strong (correctly-signed) repulsion that `cap_scaled_gradient` then rescaled; now `if let (Some(fp), Some(fm))` fails and `grad[k]` stays `0.0`.

The objective handed to the same optimizer still returns `guard_penalty_value(...)` / `1e20` at those points. The doc comment on `guard_penalty_value` (`outer_optimizer.rs:954-967`) is explicit that More-Thuente "interpolates on both, so the returned objective must integrate the returned gradient" — this change re-creates exactly the flat-∇f / walled-f mismatch that comment exists to prevent, one level down in the FD stencil. The PR summary says fixed intervals are "preserved by default"; they are not. If the `Option` contract is wanted, keep the barrier: return the guard penalty for the probe, or mark the coordinate and reuse the center-push, but don't report zero. This also has no test and no `CHANGELOG.md` entry.

### 5. `src/estimation/outer_optimizer.rs:3097, 3145` — one absolute noise bound is applied to two objectives that differ by a factor of ~N

`central_diff_packed` is shared by `reconverged_fd_gradient` (`2·Σᵢ NLLᵢ` over the whole population) and by `subject_reconverged_fd_gradient{,_iov}` (a single `foce_subject_nll`). Both now read `options.outer_fd_noise_abs`. A value calibrated for the population OFV is `N`× (and 2×) too large for the per-subject calls, so `ratio` there is `N`× too small, Shi doubles `h` to `cap`, and those subjects contribute a chord-over-a-third-of-the-box derivative into `population_gradient_sens_mixed`'s otherwise-exact analytic assembly — a wrong gradient mixed silently into a right one. This is the same units argument the PR's "Alternatives considered" uses to justify separating the inner objective and prediction bounds; it applies here too.

### 6. `src/estimation/finite_difference.rs:126-155` — a locally quadratic coordinate walks `h` all the way to the bound

For a coordinate where `f'''≈0` (a near-quadratic OFV basin, or an Ω coordinate the objective barely depends on) `ratio ≈ 0 < RATIO_LOW` forever, so `h` doubles until `h == cap == min(x−lower, upper−x)/3`, `next == h`, and the loop breaks returning `(f(x+cap) − f(x−cap)) / 2cap` — with the discarded `±3h` probes sitting exactly on the bound. Confirmed by replay: with `x = 0`, bounds `[-0.1, 5]`, `h₀ = 1e-4`, the accepted `h` is `0.0333`, a third of the distance to the bound. Cap the growth (Shi et al. bound the total expansion), or reject an interval that reached `cap` without ever satisfying the ratio test rather than returning it as if it had.

### 7. `src/estimation/finite_difference.rs:159-198` — `gill_forward`'s curvature estimate is taken once, at the interval it is supposed to be choosing

`curvature = |(f₂ − 2f₁ + f₀)/h₀²|` is a single non-iterated second difference at the *initial* `h₀`. In the regime this option exists for — noise-dominated evaluations — that second difference is `O(ε/h₀²)` noise, and the result is not a curvature:

- underestimated κ → `h = 2√(ε/κ)` clamps to `cap = (upper − x)/2`, i.e. a **forward** difference over half the parameter box;
- overestimated κ → `h` underflows, `x + h == x`, `None`, and per finding 1 the coordinate reports `0.0`.

Gill et al.'s procedure iterates on `h` with an acceptance test precisely to avoid this; without it the "curvature/noise balance" is a single draw from a noise distribution. Separately, `forward_cap` makes `Gill` structurally unable to differentiate a coordinate sitting at its **upper** bound (`cap = 0` → `None` → `0.0`, permanently), where the fixed path clamps and still returns a usable one-sided estimate. At minimum it should mirror to a backward difference near the upper bound.

### 8. `src/estimation/inner_optimizer.rs:752, 3434, 3568` — `inner_fd_method` reaches only the main inner loop; the objective and its own gradient end up on different policies

`run_inner_loop_warm` and the `find_ebe` facade hardcode `InnerFdConfig::fixed()`. `run_inner_loop_and_nll` (the objective) uses `from_options`. So with `inner_fd_method = shi`:

- `reconverged_fd_gradient` re-solves the inner loop through `run_inner_loop_warm` → **fixed** EBEs, while the objective it is differencing was built from **Shi** EBEs. The FD gradient is then the gradient of a different function than the one being minimised, which is the failure mode `reconverge_gradient_interval` exists to avoid.
- `subject_reconverged_fd_gradient{,_iov}`, and every postfit/covariance/posthoc/SAEM `find_ebe` caller, silently ignore the option — so posthoc η̂ and the reported `sdtab` come from a different derivative policy than the fit. The docs entry ("Interval policy for numerical inner derivatives when the analytic provider cannot serve a subject") promises the opposite. Thread the config, or document and warn about the scope.

### 9. `src/estimation/inner_optimizer_tests.rs:1271-1295` — the repointed FREM test cannot observe the code it now exercises, and deletes the coverage it replaced

`test_frem_jacobian_overrides_fd_with_exact_values` was switched from `compute_jacobian_fd` to `compute_jacobian_fd_config(..., InnerFdMethod::Shi)`. Its three `assert_eq!`s are all on row 2 — which `overwrite_frem_pseudo_obs_rows` overwrites unconditionally, on both branches — and the only assertion on a Shi-produced value is `jac[(0,0)].abs() > 1e-10`. Make `shi_central_vector_derivative` return `Some(vec![1e-3; n_obs])` and this test still passes. It also removes the last caller of `compute_jacobian_fd`, so the default fixed Jacobian path now has no direct test at all. Keep the original test on the fixed path and add a separate one that compares the Shi column against the fixed column (or an exact `∂f/∂η`) on the non-FREM rows.

### 10. `src/estimation/inner_optimizer_tests.rs:4-14` — `shi_inner_gradient_uses_noise_aware_stencil` passes with the feature switched off

`gradient_fd_config` falls back to the fixed stencil on `None`, and the fixed central difference of `x³` at `x = 2, h = 3e-7` is `12 + 9e-14`. Mutate `adaptive_first_derivative` to `_ => None`, or set `method: InnerFdMethod::Fixed`, and the `< 1e-3` assertion still holds — so the test cannot fail for the reason it exists. Pin something only the Shi path produces (the selected `h`, the evaluation count, or a function where the two disagree beyond the tolerance).

Coverage gaps in the same family, all reachable by the Codecov patch gate: nothing exercises `central_diff_packed`'s adaptive branch (no test sets `outer_fd_method`), nothing exercises `gill_forward`'s three `None` exits, and neither new `check_model_options` diagnostic has a test — the repo has an established `test_check_model_options_*` pattern in `src/api/tests/iov_integration.rs`.

### 11. `src/estimation/inner_optimizer.rs:3363, 3419` — two production functions became dead code

`cargo check --no-default-features --features ci --lib` on this branch reports:

```
warning: function `gradient_fd` is never used        --> src/estimation/inner_optimizer.rs:3363
warning: function `compute_jacobian_fd` is never used --> src/estimation/inner_optimizer.rs:3419
```

Every production call site moved to the `_config` variant, and finding 9 removed `compute_jacobian_fd`'s last test caller. Either delete the wrappers or keep the fixed-path test that justified them.

### 12. `src/estimation/finite_difference.rs:41` — `unused_mut` (new warning on this branch)

```
warning: variable does not need to be mutable
  --> src/estimation/finite_difference.rs:41:5
41 |     mut eval: impl FnMut(f64) -> Option<f64>,
```

`eval` is only moved into `shi_central` / `gill_forward`, never called here.

### 13. `src/estimation/inner_optimizer.rs:3382` — the inner path dispatches on the **outer** enum

`gradient_fd_config` matches on `config.method == InnerFdMethod::Shi` and then calls `adaptive_first_derivative(OuterFdMethod::Shi, …)`, discarding the inner enum. The two policy spaces are deliberately different (there is no `InnerFdMethod::Gill`), so a future inner-only variant will silently route to whatever `OuterFdMethod::Shi` means at that point. Give the inner path its own entry point, or make the translation explicit and exhaustive.

### 14. `src/estimation/finite_difference.rs:53-107` vs `109-157` — the vector and scalar Shi disagree on two failure policies

`shi_central` guards representability up front (`x + 3.0*h == x → None`) and, on a non-finite result, returns the derivative it just computed. `shi_central_vector_derivative` has no representability guard and returns the *previous* iteration's derivative. Same algorithm, two conventions; the vector one will happily start with `fp == fm` on a large `|η|` and report `0.0` as the first `last`. Worth collapsing onto one code path (the repo's "when two implementations disagree, the fix is one implementation" rule) — the scalar case is the vector case with `n = 1`.

### 15. Process — `CHANGELOG.md` and validation

`CLAUDE.md` requires a `## [Unreleased]` entry in the *same* PR as a user-facing change, and five new `[fit_options]` keys plus the finding-4 default-path change are user-facing. The PR also self-reports no NONMEM/nlmixr2 comparison, no benchmark, and no full preflight — worth resolving before merge given that findings 3-7 are all about numerical behaviour that only a real fit will expose.

---

**Suggestion**: findings 1, 2 and 4 are the ones I'd fix before anything else — each of them turns a numerical hiccup into a silently-converged fit, and 2 does it on a path ferx-r will be the first to hit.

hiddevandebeek:
## Measured: what the adaptive intervals actually buy

Closing the "end-to-end noisy-model fit" gap from the PR body, at the gradient
level. Two slow-gated Tier-3 tests in `src/estimation/fd_interval_evidence_tests.rs`,
on `examples/mm_iv.ferx` (Michaelis-Menten ODE → FD inner gradients → a
genuinely noisy OFV, nothing injected).

### The measurement had to isolate stencil error from solver bias

My first cut compared a loose-objective FD gradient against a tight-tolerance
one. That sums two unrelated quantities — the solver **bias** `∇f_loose −
∇f_tight`, which no stencil can remove, and the **stencil** error
`FD_h(f_loose) − ∇f_loose`, which is the only thing an interval policy touches.
The bias term dominates here, and the giveaway was a floor: **8.985e-4 at
`ode_reltol = 1e-4` against 8.755e-4 at `1e-6`** — a 40× noise reduction moving
the number by 3%. Each row's reference is now a wide-window least-squares slope
of that same objective.

### It works, once the noise is large enough

Worst relative outer-gradient error, controlled noise `ε` added to the
default-tolerance objective:

| `ε` (OFV units) | `fixed` | `shi` | `gill` |
|---|---|---|---|
| 1e-6 | 8.6e-4 | 8.6e-4 | 4.7e-3 |
| 1e-4 | 1.1e-2 | **6.9e-5** | 1.2e-2 |
| 1e-2 | 1.1e0 | **2.4e-3** | 2.6e-1 |
| 1e0 | 1.1e2 | **1.2e-1** | 2.0e0 |

`fixed` degrades linearly in `ε`, exactly as `ε/h` predicts; `shi` still has
three good digits where `fixed` has lost the gradient entirely — ~1000× more
accurate at the top of that range.

### But this model never gets there, and that is the more useful finding

Worst relative error against each objective's **own** true gradient:

| setting | measured noise | bias | `fixed` | `shi` | `gill` |
|---|---|---|---|---|---|
| reference (reltol 1e-12, inner_tol 1e-10) | 2.97e-8 | 0 | 1.17e-6 | 6.19e-7 | 3.64e-4 |
| ferx default | 2.34e-6 | 2.35e-4 | 7.69e-4 | 6.13e-4 | 5.04e-3 |
| loose ode (reltol 1e-3) | 2.30e-7 | 1.99e-3 | 2.03e-3 | 2.17e-3 | 3.04e-3 |
| loose inner (inner_tol 1e-3) | 2.35e-6 | 2.35e-4 | 8.01e-4 | 6.14e-4 | 5.01e-3 |
| loose both (inner_tol 1e-2) | 3.45e-6 | 1.99e-3 | 2.39e-3 | 5.24e-3 | 6.88e-3 |

The real noise is **3e-8 – 3e-6**, one to three orders of magnitude below the
`ε ≳ |g|·h ≈ 1.6e-3` threshold, and I could not push it over by degrading
`ode_reltol` or `inner_tol`. In that regime `shi` matches `fixed` and was
slightly *worse* in two of five settings — it grows `h` into curvature (one
coordinate went to `h = 1.4e-2` against the fixed `2.2e-4`, and came back
`accepted = false`, bound-limited).

**`gill` was worse than `fixed` at every noise level tested**, including 300×
worse on the clean objective (3.64e-4 vs 1.17e-6). It is one-sided and its
interval comes from an estimated second derivative that is itself
noise-limited. The docs now recommend `shi` over it.

### Consequences for this PR

- The first test asserts **the negative result it found** — the fixed stencil
  stays inside 1% on every setting — rather than the claim I wrote it to prove.
  It failed twice before I accepted that; both failures are recorded in the
  module docs so the next person does not redo them.
- Docs gain a **"When it is worth turning on"** section with the measured table,
  the `ε ≳ |g|·h` threshold, an instruction to measure your objective's noise
  first, and the `shi`-over-`gill` recommendation.
- The changelog no longer implies that ODE solver tolerance alone puts you in
  the useful regime. On this model it does not.

### Still outstanding

- **A full convergence fit** comparing `shi` against `fixed` end to end. The
  gradient result predicts no difference on this model — the noise is below
  threshold — so it would measure a null, but that is a prediction, not a
  measurement, and this box cannot run the fit.
- **The performance benchmark.** Each accepted Shi interval costs 4 objective
  evaluations per trial against the fixed stencil's 2 total; the realised factor
  is still unmeasured.

Gates: fmt, check matrix, clippy, rustdoc, docs-lint, public-API baseline all
green; `cargo test --features ci --lib` 4311 passed / 0 failed (ignored 23 → 25,
the two new slow tests).

PR #1315, opened 2026-09-09
feat(parser, tools): `categorical2` covariate form — Pharmpy MFL's `cat2` [closes #1312]

### LINKED ISSUE #1312: feat(model-file, tools): `categorical2` covariate form — Pharmpy's `cat2` shape (`θ_k` instead of `1 + θ_k`)
## What

Add a second categorical `[covariate_model]` form implementing Pharmpy's `cat2` shape, and wire it through the search space so `COVARIATE?(CL, RACE, cat2)` resolves instead of being refused.

The two shapes differ only in the non-reference branch:

| form | reference level | non-reference level `k` | null |
|---|---|---|---|
| `cat` (ferx `categorical`, today) | `1` | `1 + θ_k` | `θ_k = 0` |
| `cat2` (missing) | `1` | `θ_k` | `θ_k = 1` |

Both generate **one θ per non-reference level**, so the degrees of freedom are identical — this is a reparameterization, not a cheaper test. It is not a fix for the multiple-testing behaviour of a many-level categorical (a 4-level covariate is a 3-df test either way); it is Pharmpy MFL parity plus a parameterization that some users prefer because the θ is a multiplicative factor read directly (`θ = 1.3` → "30% higher") rather than an offset, and because its bound `θ > 0` cannot produce a negative parameter value the way `1 + θ` with `θ < -1` can.

Verified against Pharmpy's source (`src/pharmpy/modeling/covariate_effect.py`, `CovariateEffect.categorical(counts, alternative=True)`, `_choose_param_inits`, `_choose_bounds` on `main`), not the doc page:

| | init | lower | upper |
|---|---|---|---|
| Pharmpy `cat` | `0.001` | `-1` | `5` |
| Pharmpy `cat2` | `1.01` | `0` | `6` |
| ferx `categorical` (today) | `-0.001` | `-1` | `5` |

Note ferx's existing `categorical` init is `-0.001` where Pharmpy uses `+0.001` — ferx follows PsN there (`docs/model-file/covariate-model.qmd:104`). Decide deliberately whether `categorical2` takes Pharmpy's `1.01 / 0 / 6` verbatim or a PsN-consistent sign convention; whichever is chosen, state it in the docs table next to the existing row.

## Why it is not just a parser change

The reference level's factor stays `1` in both shapes, so the *null* moves from `θ = 0` to `θ = 1`. Anything that assumes "a covariate θ is null at zero" has to be checked:

- the `fix = v` clause (`apply_fix`) pins every θ of a relation at `v` — `fix = 0` on a `cat2` relation means "this level's factor is zero", not "no effect".
- covsearch's backward step removes a whole relation rather than zeroing θ, so it is unaffected — worth a test that says so.
- init/bounds defaults must not be inherited from `categorical`, or a `cat2` relation starts at a factor of `-0.001`.

## Touch points

Core:
- `src/types.rs:1518` — new `CovariateForm` variant + `label()` + `is_categorical()` (`src/types.rs:1544`) must return `true` for it.
- `src/parser/covariate_model.rs:383` form token; `:479` the `ref` keyword (same as `categorical`); `:659` θ generation (same `non_reference_levels`, different init/bounds); `:904` the factor chain — `if (COV == l) θ_k else … else 1`; `check_kind` at `:497` needs no change beyond the new variant being categorical.
- `src/edit/spec.rs` — round-trip of the new form.
- `api/ferx-core-public-api.txt` — a public enum variant, so the baseline moves in the same PR (`tools/update-public-api.sh`).

Tools:
- `crates/ferx-tools/src/covsearch/effects.rs:41` — `CovariateEffect::Cat2` currently returns an error; map it to the new form and drop it from the refusal arm.
- `crates/ferx-tools/src/search/coverage.rs:189` — remove the `cat2` gap entry.
- `crates/ferx-tools/src/search/resolve.rs` — `check_effect_kind` already treats `Cat2` as categorical (`mfl.rs:259`), so a declared-categorical covariate accepts it; the `*` wildcard should stay `cat`-only (Pharmpy's own expansion) unless we deliberately differ.

Docs / changelog:
- `docs/model-file/covariate-model.qmd:79` (form table), `:93` (PsN mapping — `cat2` has no PsN `scm` equivalent, say so), `:104` (init/bounds table).
- `docs/tools/search.qmd:233` and `docs/tools/covsearch.qmd:113` currently list `cat2` as unsupported — both must move.
- `CHANGELOG.md` under `## [Unreleased]` / `Added`.

## Tests

- Tier 1: the generated factor chain for a 3-level covariate is `θ_1` / `θ_2` / `1`, not `1 + θ`; θ names and level suffixes match the existing `categorical` scheme; init/bounds are the `cat2` ones and are *not* the `categorical` ones (a test that would fail if the arm fell through).
- Tier 1: `E_COV_LEVEL_UNKNOWN` (`src/api/validation.rs:2204`) fires for the new form too — the fallthrough is still the reference branch, so the silent-reference trap is identical.
- Tier 1: MFL `COVARIATE?(CL, SEX, cat2)` resolves to a candidate rather than a coverage gap; `COVARIATE?(CL, WT, cat2)` on a declared-continuous covariate is still refused.
- Tier 2: a `cat` model and a `cat2` model on the same 2-level data reach the same OFV at the corresponding θ (`θ_cat2 = 1 + θ_cat`) — the reparameterization identity, which is the one oracle that needs no external tool.
- NONMEM comparison: the generated `$PK` for `cat2` is `IF (COV.EQ.ref) PARCOV = 1; ELSE PARCOV = THETA(n)`, so an anchor is a plain NONMEM run of that model — no exception applies.

## Out of scope

Level collapsing / grouping (a genuinely fewer-df categorical candidate, e.g. `groups = [[1,2],[3,4]]`) is the thing that would address multiple testing on a many-level covariate. Separate issue.

roninsightrx:
## Why

Closes #1312. A `.ferxsearch` space could name `cat2` and get a coverage gap: `[covariate_model]` had one categorical form. Pharmpy MFL has two, and the second is the one some users want to *report* — the θ is the multiplicative factor read directly (`θ = 1.3` → "30% higher") rather than an offset, and its bound `θ > 0` cannot produce a negative parameter value the way `1 + θ` with `θ < −1` can.

It is **not** a cheaper test. Both shapes spend one θ per non-reference level, so a 4-level covariate is a 3-df test either way. Level collapsing (`groups = [[1,2],[3,4]]`), which would actually address multiple testing, stays out of scope as the issue says.

## What changed

`CovariateForm::Categorical2`, spelled `categorical2(ref = r)`:

| form | reference level | non-reference level `k` | null |
|---|---|---|---|
| `categorical` (MFL `cat`) | `1` | `1 + θ_k` | `θ_k = 0` |
| `categorical2` (MFL `cat2`) | `1` | `θ_k` | `θ_k = 1` |

It is an exact reparameterization of `categorical` by `θ_cat2 = 1 + θ_cat`. Everything that follows is a consequence of the *null* moving with the form:

- **Defaults are the image of `categorical`'s under that same map** — init `−0.001 → 0.999`, bounds `(−1, 5) → (0, 6)` — not inherited from it, or a `cat2` relation would start at a factor of `−0.001`. The **bounds are exactly Pharmpy's `cat2` bounds**; the **init deliberately is not** (Pharmpy uses `1.01`). ferx's `categorical` init follows PsN's sign, `−0.001` where Pharmpy uses `+0.001`; taking the image of ferx's own number carries that one convention across the reparameterization instead of mixing two, and makes the two forms start at the *identical model*. The docs table states both the choice and Pharmpy's number next to it.
- **`fix = v` pins the θ as written and is not reinterpreted per form.** So `fix = 1` is "no effect" for `categorical2` where `fix = 0` is for `categorical`, and `fix = 0` on a `categorical2` relation means "this level's factor is zero". Documented, and pinned by a differential pair.
- **covsearch's backward step is unaffected** — it removes a whole relation rather than zeroing a θ. Asserted rather than assumed.
- **`E_COV_LEVEL_UNKNOWN` fires for the new form**, because the fallthrough of the generated chain is still the reference branch and the silent-reference trap is identical. The check keys on `is_categorical()`, so this is what stops a new categorical variant slipping past it.

Search side: `CovariateEffect::Cat2` maps to the new form and drops out of the refusal arm; the `cat2` coverage-gap entry is gone. **The `*` wildcard on a categorical covariate stays `cat`-only** — Pharmpy's own expansion never reaches `cat2` either, and expanding to both would double every categorical pair's candidate count for two spellings of one test. Pinned by a test and stated in `docs/tools/search.qmd`.

## Alternatives considered

**Pharmpy's `1.01` init verbatim.** Rejected: the bounds `(0, 6)` are already the exact image of ferx's `(−1, 5)`, so taking Pharmpy's init as well would be the only place in the row where the two conventions were mixed, and the two forms would no longer start at the same model. ferx already differs from Pharmpy on the `cat` init's sign (it follows PsN there), so "Pharmpy verbatim" was never the status quo to preserve.

**Widening the `*` wildcard to both categorical forms.** Rejected above — the cost lands on every existing search space and buys a second coordinate system for the same test.

## Cross-repo dependency

Checked rather than assumed: ferx-r's only `CovariateForm` match is on `ferx_tools::gam::CovariateForm` (`src/rust/src/lib.rs:4686-4688`), a **different** enum that this PR does not touch. It depends on ferx-core by git branch with no version requirement, so the 0.3.1 → 0.4.0 bump does not re-resolve. A ferx-r `Cargo.lock` bump is only needed if/when ferx-r wants to *name* `categorical2`.

## Breaking changes

- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated with `tools/update-public-api.sh` in this PR.
- [x] **Version bumped: 0.3.1 → 0.4.0** (root package, both members, their `ferx-core` / `ferx-tools` version pins, `Cargo.lock`).

The added item is `CovariateForm::Categorical2`. It is needed by `ferx-tools`' covsearch (`Effect::from_spec` maps `CovariateEffect::Cat2` onto it) and is reachable from ferx-r as an ordinary variant of an already-public enum. The existing surface does not suffice because the form *is* a variant of that enum — there is no way to express it without one.

**This is a breaking change, not an additive one.** Adding a variant to an exhaustive public enum breaks any downstream crate that `match`es it exhaustively, which `docs/development/sdlc.qmd` classes as a public-API break — pre-1.0, a MINOR bump. An earlier revision of this PR identified that failure mode and then left the bump box clear; corrected in d8f8c85.

`CovariateForm` is now **`#[non_exhaustive]`** in the same breaking release, so the next form — level grouping, the out-of-scope note on #1312 — is genuinely additive. This follows the precedent `NewParameter` set for the same reason. The workspace needed no changes: both members already match with a `_` arm.

<details>

A `match` on `ferx_core::CovariateForm` that was exhaustive under 0.3.1 fails to compile under 0.4.0:

```
error[E0004]: non-exhaustive patterns: `CovariateForm::Categorical2` not covered
```

Add a `_ => …` arm (or an explicit `CovariateForm::Categorical2` arm). The enum is now `#[non_exhaustive]`, so the `_` arm is required from here on.

Nothing else changes: variants are still constructible, the serde representation of every existing variant is unchanged, and `.ferx` files, `FitResult` and sdtab are untouched.

</details>

## Numerical validation

- [x] Compared against: **NONMEM 7.5.1** (`METHOD=1 INTER`)
- [x] Reference values: `tests/nonmem/covariate_cat2.{ctl,lst,ext}` + the `cat` twin `covariate_cat.{ctl,lst,ext}`, asserted in `tests/covariate_cat2_nonmem.rs`

`categorical2` has an exact NONMEM spelling — `IF (SEX.EQ.1) COVCL = THETA(4)`, three lines of `$PK` — so no exception to the NONMEM rule applies and the anchor is a plain converged run.

**Data** (`tests/nonmem/covariate_cat2.csv`): 60 subjects, single 100 mg oral dose, ten sampling times, `SEX` alternating 0/1 (30/30). Simulated **outside both engines** — a Bateman closed form in Python, seed 20250909 — from `TVCL = 1`, a `cat2` factor of `1.5` on `CL` at `SEX = 1`, `TVV = 20`, `TVKA = 1`, `ω² = 0.09` each, proportional `σ² = 0.04`. The effect is ~5 SE from the null, so neither engine can pass by fitting noise, and the contrast branch — the only branch the two shapes differ in — is exercised by half the dataset.

**Two NONMEM runs, not one.** `covariate_cat.ctl` is the `1 + THETA(4)` shape on the same data from the corresponding start. The pair is what makes the reparameterization identity an *external* result: a single run would anchor the number but not the claim that the two forms are the same model.

| Quantity | NONMEM `cat2` | ferx `cat2` | NONMEM `cat` | ferx `cat` |
|---|---|---|---|---|
| OFV | −73.8266171 | −73.8266174 | −73.8266171 | −73.8266174 |
| TVCL | 0.983305 | 0.9833024 | 0.983305 | 0.9833053 |
| TVV | 21.4119 | 21.411936 | 21.4119 | 21.411948 |
| TVKA | 1.03442 | 1.0344236 | 1.03442 | 1.0344269 |
| θ(SEX) | 1.68330 | 1.6833125 | 0.683302 | 0.6833155 |
| SE θ(SEX) | 0.137302 | 0.1371289 | 0.137302 | 0.1371289 |
| ω²(CL) | 0.0827743 | 0.082772 | 0.0827743 | 0.082772 |
| ω²(V) | 0.105923 | 0.105925 | 0.105923 | 0.105925 |
| σ² (prop) | 0.0381895 | 0.0381894 | 0.0381895 | 0.0381894 |

Both engines reach the same objective from both spellings, and both land with their two θ exactly 1 apart. ferx's SE is pure R⁻¹, matched by `$COVARIANCE MATRIX=R`.

**Tolerances are measured, not argued.** Realised worst errors, quoted at each assertion in the test: |ΔOFV| `2.563e-7`; structural θ `3.477e-6` relative (TVKA); θ(SEX) `7.425e-6`; SE θ(SEX) `1.261e-3`; ω²(CL) `3.009e-5`; σ² `1.109e-6`; |θ_cat2 − θ_cat − 1| `2.958e-6`. Bounds sit an order of magnitude above those, floored at `1e-4` where NONMEM only prints six significant figures.

## Performance

- [x] No significant impact expected — one more arm in three `match`es on the parse path.

## Tests

- [x] Unit tests added (Tier 1), integration tests added (Tier 2), NONMEM anchor added (Tier 3, `slow-tests`-gated)
- [x] Every new assertion mutation-checked

**Tier 1** (`src/parser/covariate_model_tests.rs`, `src/api/tests/covariate_stats_tests.rs`, `src/edit/mod_tests.rs`, `src/types_tests.rs`)
- `every_form_the_parser_accepts_is_offered_by_both_diagnostics` — added in review (see P2 below): both "expected one of …" messages offer every form, and every enum variant's label is in the shared list, driven by an exhaustive `match`
- the generated chain is `θ_1 / θ_2 / 1`, not `1 + θ`; θ names and level suffixes are byte-identical to `categorical`'s
- init/bounds are the `cat2` ones **and explicitly not** the `categorical` ones — the arm-fallthrough test the issue asks for
- `ref` (not `center`); kind check accepts it on a categorical column and refuses it on a continuous one, naming the form written
- `E_COV_LEVEL_UNKNOWN` fires for the new form and names it
- `ref = mode` binds to the most common level — which is also the round-trip of what `Relation::render()` writes

**Tier 2** (`tests/covariate_categorical2_equivalence.rs`) — the reparameterization identity, bit for bit: `predict()` agrees exactly at θ and 1 + θ (`0.25` / `1.25`, so `1 + θ` is exact in IEEE), and the objective agrees at `outer_maxiter = 0`. Plus the `fix` differential pair.

**Tier 3** (`tests/covariate_cat2_nonmem.rs`) — the NONMEM anchor above.

### Non-degeneracy, and what would have made these green for the wrong reason

- **Both levels populated, and the covariate live.** Each arm asserts `SEX` actually changes the prediction and moves the objective against a `none` model first — otherwise the identity holds between two copies of the null. The anchor separately asserts θ(SEX) is >4 SE from the factor of 1.
- **`fix` is a straddle, not a half.** `categorical2(fix = 1)` must be the covariate-free model **and** `categorical(fix = 1)` must not be (it is a factor of 2), **and** `categorical2(fix = 0)` must not be (a factor of zero). Any one half alone passes against an implementation that silently re-centres the fixed value.
- **The identity is compared at `outer_maxiter = 0`, deliberately.** The outer optimizer's packing is not equivariant under `θ ↦ 1 + θ`: from corresponding starts the two arms take different steps and after 2 iterations sit **2.47 OFV apart**, and still **1.78 apart** with the covariate θ `FIX`ed. That is a difference of trajectory, not of objective, and comparing short fits would have been a test that could only be satisfied by luck. An evaluation scores the objective at the parameters as given, which is the quantity the identity is about.
- **The anchor's `is_finite()` is not the only guard.** A diverged solve is caught by `is_finite()`; a wrong-but-finite answer is caught by the relative bounds, which are ~4 orders tighter than a 1% regression.

### Mutation results

Three minimal edits that remove the feature, and which tests die:

| mutation | Tier 1 | Tier 2 | Tier 3 |
|---|---|---|---|
| `offset = true` (cat2 emits the `cat` shape) | 4 of 8 fail | **3 of 3** fail | **2 of 2** fail |
| drop the `Categorical2` init/bounds arm (fall through to `categorical`'s) | 2 of 8 fail | — | — |
| drop `Categorical2` from `is_categorical()` | **8 of 8** fail | — | — |

## Docs

- [x] `docs/model-file/covariate-model.qmd` — forms table, the "no `scm` state" note, the init/bounds table with the decision and Pharmpy's number beside it, the `fix` semantics, and the NONMEM anchor summary
- [x] `docs/tools/search.qmd` — `cat2` moved from the gap table to supported; the effect-name mapping; the wildcard deviation rewritten
- [x] `docs/tools/covsearch.qmd` — `cat2` added to the MFL→line table and removed from the refusal sentence
- [x] `CHANGELOG.md` `[Unreleased]` / `Added`

## Checklist

- [x] `cargo test --workspace --lib` green — 4288 + 590 + 61 passed
- [x] `cargo test --features slow-tests --test covariate_cat2_nonmem` green
- [x] Not on the `Dual2` sensitivity path — the change is in the desugar, which emits an expression the existing generic path differentiates
- [x] `Cargo.toml` version bumped — 0.3.1 → 0.4.0, see *Breaking changes*

> One unrelated flake seen locally on a full `--lib` run: `ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` failed once under parallel execution and passes in isolation. It asserts on process-global SS-equilibration counters and this PR touches nothing under `src/ode/`, `src/dosing.rs` or the SS path (`git diff origin/main...HEAD -- src/ode src/dosing.rs` is empty). Pre-existing; worth its own issue, not fixed here.

## Example

```
[covariates]
  SEX categorical(levels = [0, 1])

[covariate_model]
  CL ~ SEX categorical2(ref = 0)
```

generates `theta THETA_CL_SEX_1(0.999, 0, 6)` and

```
CL = TVCL * (if (present(SEX)) (if (SEX == 1) THETA_CL_SEX_1 else 1) else 1.0) * exp(ETA_CL)
```

`tests/nonmem/covariate_cat2.ferx` is the runnable version, fitted in the anchor above.

## Reviewer hints

Focus on the **init decision** (`src/parser/covariate_model.rs`, the `(init, lower, upper)` match) — it is the one place the issue asked for a judgement call, and the argument for `0.999` over Pharmpy's `1.01` is in *Alternatives considered*. Second, the **`fix` semantics**: the null moved, and the choice not to reinterpret `fix` per form is deliberate and load-bearing for anyone writing `fix = 0` out of habit.

The parser diff is mostly mechanical `Categorical | Categorical2` widening; several hunks are error messages switched from a hardcoded `categorical(...)` to `form.label()` so they name what the user actually wrote.

## Open questions

~~Should `CovariateForm` become `#[non_exhaustive]`?~~ Resolved in review: yes, and in this release. See *Breaking changes*.

https://claude.ai/code/session_01MfJjyabnSjpQFfeSG1oX9Z


roninsightrx:
Review findings:

1. [P1] Treat the new public enum variant as a breaking API change. src/types.rs:1517 adds CovariateForm::Categorical2 to an exhaustive public enum. Any downstream crate with an exhaustive match that compiles against 0.3.1 will fail after this update; the PR body already identifies that failure mode, but then calls the change additive and leaves the version-bump checkbox clear. Per docs/development/sdlc.qmd and the PR checklist, this needs the pre-1.0 minor-version bump (including matching workspace dependency versions/lockfile) and migration notes. Since another variant is anticipated, consider making CovariateForm non_exhaustive in the same breaking release so future additions are actually additive.

2. [P2] Keep the missing-form diagnostic in sync with the parser. src/parser/covariate_model.rs:264-265 still says the expected forms are none, linear, linear_relative, exponential, power, hockey, categorical, expr, omitting the newly valid categorical2 spelling. A user who leaves the form off is therefore given an incomplete list that hides this feature. Add categorical2 there and cover the diagnostic with a regression assertion.

Focused validation passed: cargo test --workspace --lib categorical2; cargo test -p ferx-tools covsearch::effects::tests; the categorical/continuous kind-resolution test; and all three tests in tests/covariate_categorical2_equivalence.rs.

roninsightrx:
Both findings were real and are fixed in d8f8c85. Thanks — P2 in particular was a miss I would not have caught from the diff.

## P2 — missing-form diagnostic ✅ fixed

Confirmed: `covariate_model.rs:264` still listed eight forms, so a user who left the form off was never shown `categorical2`. The existing test only asserted that `linear_relative` appeared, which is why it stayed green.

The cause was **three** hand-maintained copies of the same list, of which #1312 updated two (the unknown-form message and `did_you_mean`). So rather than patch the third, I collapsed all three onto one `FORM_SPELLINGS` array.

The regression assertion you asked for has two halves, because one alone would be a tautology now that the list is shared:

1. every spelling is offered by **both** "expected one of …" messages — which, with a single list, cannot fail on its own today;
2. every `CovariateForm` variant's `label()` is **in** that list, driven by an exhaustive `match`, so a new variant is a compile error until it is listed and a test failure until the diagnostics print it.

Verified it catches exactly what you reported: reinstating the old stale message reddens it with your text —

```
the missing-form message omits `categorical2`: [covariate_model]: `CL ~ WT` states no form
  — expected one of none, linear, linear_relative, exponential, power, hockey, categorical, expr
```

`covariate_form_labels_are_the_block_spellings` also gained the `Categorical2` case and an `is_categorical()` assertion — it had silently skipped the new variant too.

## P1 — breaking change and version bump ✅ fixed, and you were right to reject my framing

The PR body checked the *Public Rust API* box and then wrote "not needed (additive)" against the version bump. That is self-contradictory, and `docs/development/sdlc.qmd:471-474` is unambiguous: a public-API break bumps MINOR pre-1.0.

- **0.3.1 → 0.4.0** across the root package, `ferx-tools`, `ferx-cli`, both version pins, and the lockfile.
- **Migration notes** added to `CHANGELOG.md` under `Changed`, and to the PR body.
- **`CovariateForm` is now `#[non_exhaustive]`**, taking your suggestion to do it in the same breaking release. Level grouping is the anticipated next variant (the out-of-scope note on #1312), so this is the last release in which adding a form is breaking. It follows the precedent `NewParameter` already set here for the same reason. No workspace changes were needed — both members already match with a `_` arm.

On the cross-repo line I had asserted the R wrapper was unaffected; now checked rather than assumed. Its only `CovariateForm` match is on `ferx_tools::gam::CovariateForm` (`src/rust/src/lib.rs:4686-4688`), a different enum this PR does not touch, and it depends on ferx-core by branch with no version requirement, so the bump does not re-resolve.

## One thing to flag, unrelated to this PR

A full local `--lib` run had `ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` fail once; it passes in isolation. It asserts on process-global SS-equilibration counters, and this PR touches nothing on that path — the diff against main is empty for `src/ode`, `src/dosing.rs`, `src/estimation`, `src/pk` and `src/sens`. Pre-existing parallel-execution flake; I have not fixed it here, but it probably deserves its own issue.

`tools/preflight.sh` green (fmt, 5 check feature sets, clippy `--all-targets`, rustdoc, docs lint, public-API baseline). Your focused validation set re-run green.

PR #1316, opened 2026-09-09
feat(parser, tools): additive (`+`) covariate effects [closes #1313]

### LINKED ISSUE #1313: feat(model-file, tools): additive (`+`) covariate effects — the last MFL operator with no `[covariate_model]` spelling
## What

`[covariate_model]` relations are multiplicative factors on a top-level product. Pharmpy's MFL takes an operator argument — `COVARIATE(CL, WT, lin, +)` — and ferx has no spelling for it, so the search refuses it as a coverage gap (`crates/ferx-tools/src/search/coverage.rs:205`, `crates/ferx-tools/src/covsearch/effects.rs:50`). This is the last MFL operator ferx cannot express, and unlike `custom` the refusal is not a design choice — it is a missing block feature.

Pharmpy semantics, from source (`src/pharmpy/modeling/covariate_effect.py`, `OperationType = Literal['*', '+']`, `CovariateEffect.create_effect_statement`) rather than the docs:

```
CL = CLWGT + TVCL*EXP(ETA(1))
```

The **effect template is unchanged** — `coveff = 1 + θ·(WT − median)` for `lin`, exactly as in the multiplicative case — and it is the *combination* that changes from `param * coveff` to `param + coveff`.

## The design question to settle first

Because the template is shared, Pharmpy's additive effect adds `1 + θ·(COV − c)`, so a covariate sitting at its centre **adds 1** to the parameter, and `θ = 0` is not "no effect". That is Pharmpy's behaviour verbatim, and it is almost certainly not what a user writing `CL ~ WT linear +` expects.

Two options, and the issue should not be implemented before one is chosen:

1. **Pharmpy-verbatim** — same template, `+` combination. Buys literal MFL parity and translation fidelity; ships a null that is not at zero.
2. **Null-at-zero additive** — the additive spelling drops the leading `1` (`θ·(COV − c)`, `θ_k` per level, `exp(...) − 1`, …). Reads correctly, but a model translated from Pharmpy with `+` no longer round-trips to the same equations, and the difference is silent because both fit.

Whichever is chosen must be stated in `docs/model-file/covariate-model.qmd` next to the forms table, and pinned by a test that would fail under the other choice.

## Why this is not a small change

The multiplicative assumption is load-bearing in three places:

- **`insert_factors` requires a top-level product** (`src/parser/covariate_model.rs:970`). It rejects any RHS with a top-level `+` via `is_simple_factor` (`:1029`) with `non_product_error`, precisely so a factor is never multiplied into one addend of a sum. An additive relation *creates* that sum, so a second, multiplicative relation on the same parameter would then hit the error. Mixing `*` and `+` on one parameter needs defined placement and defined precedence, or a deliberate refusal with a message that says why.
- **The #619 mu-reference path.** Factors are placed immediately before the first η/κ-bearing term specifically so the effect stays in the non-η group the SAEM mu-reference detector reads. An additive term is outside any such group: the typical value stops being log-linear in one θ, which is the exact shape that made SAEM drop covariate effects in #619. Decide and *test* what the detector does with an additive relation rather than discovering it in a fit.
- **The missing-covariate guard has the wrong identity.** `relation_factor` wraps every relation as `(if (present(COV)) <factor> else 1.0)` (`src/parser/covariate_model.rs:920`). The neutral element of an additive term is `0`, not `1` — left as is, a subject with a missing covariate silently gets `+1` on the parameter. Division underflowing to zero rather than infinity makes this class of bug silent in ferx generally, so the guard must be per-operator.

## Touch points

Core:
- `src/types.rs` — `Relation` gains an operator (`CovariateOp`-equivalent); the enum and any new field are public API, so `api/ferx-core-public-api.txt` moves in the same PR (`tools/update-public-api.sh`).
- `src/parser/covariate_model.rs` — block spelling for the operator; `relation_factor` (`:876`) neutral element per operator; `insert_factors` (`:970`) additive placement + the `*`-after-`+` interaction; `non_product_error` (`:1013`) message update.
- `src/edit/spec.rs` — round-trip of the operator through `ModelEdit`.
- Whatever the #619 mu-reference classifier reads (`random_bearing_names`, `:943`) — an additive term must be classified deliberately.

Tools:
- `crates/ferx-tools/src/covsearch/effects.rs:50` — `Effect` carries only `(parameter, covariate, form)`; add the operator, and decide whether `pair()` (`:67`, "one `[covariate_model]` line per pair") still keys dedup and backward-removal, or whether `CL ~ WT lin *` and `CL ~ WT lin +` are two candidates for the same pair.
- `crates/ferx-tools/src/search/coverage.rs:205` — drop the `COVARIATE(..., +)` gap.
- `crates/ferx-tools/src/search/resolve.rs` — `op` already parsed and carried on `CovariateEffectSpec` (`mfl.rs:265`), so nothing new to parse.

Docs / changelog:
- `docs/model-file/covariate-model.qmd` — forms table + the operator and its null, and the missing-covariate identity.
- `docs/tools/search.qmd:233` and `docs/tools/covsearch.qmd:113` list `+` as unsupported; both move.
- `CHANGELOG.md` under `## [Unreleased]` / `Added`.

## Tests

- Tier 1: an additive relation renders the parameter as a sum with the term placed where the design says, and the generated expression is the chosen null shape (a test that fails under the other option above).
- Tier 1: missing covariate contributes `0` for an additive relation and `1` for a multiplicative one — asserted on the *same* model, so a shared-guard regression cannot pass.
- Tier 1: `*` and `+` relations on the same parameter — either both land correctly, or the refusal names the reason; and a multiplicative relation added *after* an additive one does not silently multiply into one addend (the `non_product_error` case that motivated `is_simple_factor`).
- Tier 1: the #619 mu-reference classification of an additive term is pinned, not incidental.
- Tier 2: `COVARIATE?(CL, WT, lin, +)` resolves to a covsearch candidate and is no longer a coverage gap; the step's LRT `df` is 1.
- NONMEM anchor: `CL = TVCL*EXP(ETA(1)) + CLWGT` is an ordinary NONMEM model, so this gets a normal anchored comparison — no exception applies. Anchor on a covariate that genuinely varies and an additive term that is materially non-zero, so neither side of the combination is degenerate.

roninsightrx:
## Why

`[covariate_model]` relations were multiplicative factors on a top-level product. Pharmpy's MFL takes an operator argument — `COVARIATE(CL, WT, lin, +)` — and ferx had no spelling for it, so the search had to refuse it as a coverage gap (`crates/ferx-tools/src/search/coverage.rs`). It was the **last** MFL operator ferx could not express, and unlike `custom` the refusal was not a design choice: it was a missing block feature. Closes #1313.

## What changed

A trailing operator token makes a relation a term added to the parameter instead of a factor on it:

```
[covariate_model]
  CL ~ WT linear(center = 70) +
->  CL = TVCL * exp(ETA_CL) + (if (present(WT)) (THETA_CL_WT * (WT - 70)) else 0.0)
```

`*` stays the default and may be written explicitly. The operator is the last token of the line body, **before** any `=>` clause, mirroring MFL where it is the last argument.

### The design question the issue asked to settle first: null-at-zero

Pharmpy reuses the *multiplicative* template under `+`, so its additive linear effect adds `1 + θ·(WT − median)`: a covariate at its centre adds `1` to the parameter and `θ = 0` is not "no effect". **ferx drops that leading `1`** — `θ·(COV − c)` for `linear`, `exp(θ·(COV − c)) − 1` for `exponential`, `(COV/c)^θ − 1` for `power`, `θ_k` / `0` for `categorical` — so θ = 0, a covariate at its centre and a *missing* covariate all mean the same thing. The units work out too, which they do not when a dimensionless `1` is added to a clearance.

The cost is that a model translated from Pharmpy with `+` does not reproduce Pharmpy's equations; the two differ by a constant `1` on the parameter. That is **not** left silent: `docs/model-file/covariate-model.qmd` states it next to the forms table, `CovariateOp`'s rustdoc states it, and `resolve` emits a note on any space that asks for `+`. It is a semantic divergence, not a coverage gap, so the coverage check does not refuse the space.

### The three load-bearing assumptions the issue named

- **`insert_factors` required a top-level product.** Split into `multiply_in` (unchanged behaviour, unchanged error) + `insert_effects`. The product is rebuilt from the **original** right-hand side and the terms are appended after it, so mixing `*` and `+` on one parameter lands in one rewrite and a factor can never be multiplied into one addend of a sum this block created — whatever order the two lines are written in. An additive relation carries no top-level-product requirement at all (there is nothing to multiply into); `non_product_error` now says so. A right-hand side that is not a plain product is **parenthesised** before a term is appended, since `if (c) a else b + t` would otherwise bind the term inside the `else` arm.
- **The #619 mu-reference path.** An additive term makes the typical value a sum, which `detect_mu_refs` does not match, so mu-referencing switches off for that parameter and SAEM/IMP fall back to the numerical M-step. That is the *safe* direction — the covariate effect stays in the expression; #619 was the effect being **dropped** — but it is a performance cliff, so `model_parser` now emits a parse warning naming the parameter (only for a parameter that actually carries an η). Pinned by a differential test whose two arms differ by one character.
- **The missing-covariate guard had the wrong identity.** `CovariateOp::neutral()` is the operator's: `1.0` for a factor, `0.0` for a term. Asserted on **one model** carrying both kinds of relation on the same column, so a shared-guard regression cannot pass by getting one of them right.

### Tools

- `COVARIATE(..., +)` is no longer a coverage gap (a *form* gap under `+` still is).
- `Effect` carries the operator. `pair()` is unchanged — one `[covariate_model]` line per pair — so `CL ~ WT linear` and `CL ~ WT linear +` are two candidates **competing** for that one line, distinctly named (`CL-WT-linear` / `CL-WT-linear-add`; the suffix appears only under `+`, so every id a multiplicative search produced before is unchanged).
- `resolve` already parsed and carried `op`; nothing new to parse.

## Alternatives considered

- **Pharmpy-verbatim semantics** (same template, `+` combination): literal MFL parity and exact round-tripping, but ships a null that is not at zero — dropping a relation would change the model even at θ = 0. Rejected; the divergence is documented instead.
- **Operator between covariate and form** (`CL ~ WT + linear(...)`): invites `CL ~ WT + AGE`, the term-expression reading this block's syntax exists to avoid (see the module docs' "Why `PARAM ~ COV form(...)`").
- **`op = add` keyword inside the form parens**: would widen `parse_kwargs` beyond numbers/statistics (it maps to `CovariateStat`) for no readability gain.

## Cross-repo dependency

`CovariateOp` and `CovariateRelation::op` are additive public API; `#[serde(default)]` on the field means an existing serialized relation still deserializes as `Multiply`. Nothing ferx-r reads changes shape.

## Breaking changes

- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated with `tools/update-public-api.sh` (+15 lines)

Added items and who needs them:

| Item | Needed by |
|---|---|
| `CovariateOp` (enum, `label`, `neutral`) | `ferx-tools` (`covsearch::effects` maps the MFL operator onto it); ferx-r reads it off `CovariateRelation` when echoing a covariate model |
| `CovariateRelation::op` | the parsed echo a covariate search consumes — without it a search cannot tell which combination a base-model relation uses |
| `edit::Relation::op` | `ModelEdit::AddCovariateRelation` — the field the candidate writes |

The existing surface does not suffice: every one of these is *the* new piece of information, and there is no way to infer the operator from the form or the centre.

## Numerical validation

- [x] Compared against: **NONMEM 7.5.1** (`$EST METHOD=1 INTERACTION MAXEVAL=0 POSTHOC`, `ADVAN4 TRANS4`, `$TABLE FORMAT=s1PE23.16`), four new control streams under `nonmem_anchor/`, run outputs under `nonmem_anchor/results/`, test in `tests/additive_covariate_nonmem_anchor.rs`.

Two arms, each with a **null twin** (the same stream with the additive slope `0 FIX`):

| Arm | Model | ferx vs NONMEM |
|---|---|---|
| A `additive_cov_lin` | `CL = TVCL*EXP(ETA(1)) + 0.04*(WT-70)` | PRED **3.08e-15** rel |
| B `additive_cov_mixed` | `CL = TVCL*(WT/70)**0.6*EXP(ETA(1)) + 0.02*(CRCL-100)` | PRED **2.65e-15** rel |

```
# PRED, worst relative error over all 300 observations
arm A  3.08e-15      arm B  2.65e-15        (bound 1e-13, 30x the measurement)

# ΔOFV against each arm's null twin
arm A  ferx -4.085367142659834   NONMEM -4.085479766371918   diff 1.13e-4
arm B  ferx +1.007244131352536   NONMEM +1.007559281418935   diff 3.15e-4

# non-degeneracy: the same comparison with the additive slope zeroed
arm A  3.9e-1        arm B  4.3e-1
```

**Why the objective is compared as a difference, and a finding this PR does not fix.** ferx and NONMEM do not report the same FOCEI objective on `data/two_cpt_oral_cov.csv` at these initial estimates, and **that is true with no covariate model at all**: the bare `CL = TVCL*exp(ETA_CL)` model reads `-1017.3100571` in ferx against `-1033.3346138` in NONMEM — a **16.02** offset that this PR neither introduces nor touches. Measured, not assumed: `additive_cov_lin_null.ctl` *is* that model, ferx reproduces its own value to 1e-9 across `inner_tol` 1e-10 .. 1e-14 and `inner_restarts` 0 .. 8 (so it is not an inner-EBE tolerance), and PRED agrees to 3e-15 (so the value path is exact and the difference lives in the objective/EBE assembly). The offset is stable in the additive slope (16.0246 at 0, 16.0247 at 0.04), so the difference against a null twin cancels it and isolates what the additive term contributes. **This looks like it deserves its own issue** in the #864 family — happy to open one.

A trap worth recording: at slope 0.05 ferx's arm-A objective lands *exactly* on NONMEM's (10 significant figures) while the baselines differ by 16 — a coincidental mode jump, because at that slope the additive term drives `CL` negative for low-`WT` subjects with `ETA_CL < -0.4`. An absolute-OFV anchor at that point would have been a false green. The committed arm uses 0.04, and the hazard is documented as a user-facing warning callout.

- [ ] Warfarin example (untouched by this change)

## Performance

- [x] No significant impact expected — desugar is text rewriting at parse time; the generated expression is an ordinary sum.

## Tests

- [x] Unit test(s) added in the same module — 16 new Tier-1 tests in `src/parser/covariate_model_tests.rs`, 1 in `src/edit/mod_tests.rs`, 1 in `crates/ferx-tools/src/covsearch/effects.rs`, 1 in `resolve_tests.rs`, 1 in `covsearch/mod_tests.rs`; `coverage_tests.rs` updated.
- [x] Regression test added that fails without this fix — **five mutations run, each reddening its own tests**:

| Mutation | Tests reddened |
|---|---|
| Pharmpy-verbatim template (`let add = false`) | 8 |
| Shared neutral element (`Add => 1.0`) | 8 |
| No parenthesisation before appending a term | 1 (the conditional-RHS test) |
| Terms appended before the product is rebuilt (the naive "apply in declaration order" design) | 3 |
| mu-reference warning removed | 1 |

Three assertion-quality points, per the CLAUDE.md rules:
- the mu-reference test is a **differential pair straddling the gate** — the two models differ by one character, and both sides are asserted (multiplicative keeps its anchor, additive loses it), so it cannot silently become a tautology;
- `worst_pred_error` asserts `is_finite()` **before** folding, because `f64::max` discards `NaN` and would pass on the strength of the records that worked;
- the anchor's non-degeneracy is a **test**, not a comment: with the slope zeroed the same comparison misses by 39 % / 43 %.

Full workspace suite: **4663 passed, 0 failed** (`cargo test --workspace --features ferx-core/ci`).

- [x] Inline below and in `docs/model-file/covariate-model.qmd`

<details>
<summary>Example</summary>

```toml
[covariates]
  WT   continuous
  CRCL continuous

[covariate_model]
  CL ~ WT   power(center = 70)        # multiplicative, the default
  CL ~ CRCL linear(center = 100) +    # additive
```

```
# desugared [individual_parameters] (what `ferx check` prints)
CL = TVCL * (if (present(WT)) (WT / 70)^THETA_CL_WT else 1.0) * exp(ETA_CL)
     + (if (present(CRCL)) (THETA_CL_CRCL * (CRCL - 100)) else 0.0)
```

</details>

## Docs

- [x] `docs/model-file/covariate-model.qmd` — grammar, a new "Multiplicative and additive effects" section with the per-form table, the Pharmpy-divergence warning, the mu-referencing note, the negative-parameter hazard, placement, and the per-operator missing-value guard
- [x] `docs/tools/search.qmd` + `docs/tools/covsearch.qmd` — the `+` rows move from ✗ to ✓ with the divergence stated
- [x] `nonmem_anchor/README.md` — anchor row
- [x] `CHANGELOG.md` `[Unreleased]` / `Added`
- [ ] New pages (none added, so no `_quarto.yml` change)

## Checklist

- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched; the desugar emits an ordinary sum that the existing expression compiler and `Dual2` path already handle (there is no second evaluation path for `[covariate_model]`)
- [ ] `Cargo.toml` version bumped (not breaking)

### ferx-site
- [x] New `.ferx` DSL syntax → `docs/model-file/covariate-model.qmd` updated here; a matching `ferx-site/model-dsl/` update is worth opening if that page mirrors the block
- [ ] No new example `.ferx` file or data file added

### ferx-book
- [ ] Worth a mention in the covariate-model chapter if one exists — not opened

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` file changed; the new syntax is exercised end to end by the NONMEM anchor and the covsearch space test

## Reviewer hints

Focus on three places:

1. **`relation_effect`** (`src/parser/covariate_model.rs`) — the per-form additive templates. This is where the null-at-zero decision physically lives.
2. **`insert_effects` / `multiply_in`** — the split. The property worth checking is that `multiply_in` is handed the *original* right-hand side, which is what makes declaration order irrelevant and keeps the `is_simple_factor` guard asking about the model the user wrote rather than the sum this function creates.
3. **The anchor's differential design** (`tests/additive_covariate_nonmem_anchor.rs` module docs) — specifically whether cancelling a 16.02 baseline offset is the right call, or whether that offset should block this PR.

Skimmable: the doc/CHANGELOG/README prose, the public-API baseline diff, the mechanical `op:` field additions at struct literals.

## Open questions

1. **The 16.02 FOCEI offset.** Should I open an issue for it now? It reproduces with no covariate model at all on `data/two_cpt_oral_cov.csv`, so it is orthogonal to this PR, but it is the reason the OFV comparison here is differential rather than absolute.
2. **`none` with an operator.** `CL ~ WT none +` parses and is inert (the form declares nothing). That keeps a search's "tested, rejected" line round-tripping with the operator it tested — deliberate, but say if you would rather it were rejected.
3. **ferx-site / ferx-book mirrors** are not opened. Say the word and I will.

https://claude.ai/code/session_01932EFAmdTeRjZjodChamLJ


roninsightrx:
## Code review — additive (`+`) covariate effects (`4ac168c7`)

Reviewed every hunk in `src/`, `crates/`, `docs/` and the NONMEM anchor. The desugar itself looks right: `is_plain_product` parenthesises a non-product RHS correctly, `multiply_in` runs on the original RHS so declaration order can't matter, the per-operator neutral is right, `split_trailing_op` is byte-safe and runs after the `=>` split on comment-stripped text, `canonical_hash` distinguishes the two operators, and `#[serde(default)]` keeps back-compat. Findings below, most severe first.

---

### 1. `src/edit/spec.rs:740` — a trailing `+` line folds into the next relation line (high)

`Relation::render` ends the line with a bare `+` when there is no `=>` clause — exactly what `Effect::relation()` produces for covsearch/globalsearch, which leave `thetas: Vec::new()`. `ModelText::logical_lines` lists `'+'` in `TRAILERS` (`src/edit/mod.rs:254`) and folds such a line with the next one.

The parser does **not** fold `covariate_model` (it is not in `CONTINUATION_BLOCKS`, `model_parser.rs:12592`), so the two layers disagree about what a line is. Observed on a two-step forward search with an additive relation followed by any other relation:

- `block_lines("covariate_model")` returns the two relations as one folded line;
- re-adding an already-present pair returns `Ok` (duplicate physical line → parser errors with "declared more than once");
- `DropCovariateRelation{V, AGE}` errors and aborts the backward pass;
- `DropCovariateRelation{CL, WT}` silently deletes the *following* relation and its θ.

Every additive fixture in the PR either carries a `=>` clause or is the last line of the block, so nothing in the suite can see it. Suggest either emitting the operator before the `=>` position in a way that never ends the line (or always writing a `=>` clause from the search layers), or excluding `covariate_model` from the trailer fold — plus a fixture with a bare-`+` relation that is **not** last in the block.

### 2. `src/parser/covariate_model.rs:801` — additive θ inherits the multiplicative positivity bounds (medium)

`linear_family_thetas` hands PsN's defaults `θ ∈ [1/(c−max), 1/(c−min)]` to the additive template. The docstring states their purpose outright: they are what keeps `1 + θ·(COV − c)` positive. Under `+` there is no factor to keep positive, and θ has different units — parameter-per-covariate-unit, not fraction-per-unit.

Search-generated candidates always take these defaults. On WT 45..94, median 70, that is θ ∈ [−0.042, 0.040] with init 4e-5: an additive term on `V1 ≈ 40` can move it by at most ±1 L and sits on its bound before it expresses anything, while the same bounds on `KA ≈ 1` allow a 100 % swing. Categorical `[-1, 5]` has the same problem.

The NONMEM anchor cannot see this — both arms pass explicit `=> THETA_CL_WT(0.04, -1.0, 1.0)`.

### 3. `src/parser/covariate_model.rs:822` — centre-outside-range rejection applies a multiplicative rationale to `+` (low-medium)

Same function: the hard error says the default bounds "are only ordered (and only keep the covariate factor positive) there". `CL ~ WT linear(center = 100) +` on WT 45..94 fails to parse, though `θ·(WT − 100)` is well-defined and needs no positivity constraint. Falls out of #2 — an additive-specific default would fix both.

### 4. `src/parser/model_parser.rs:3820` — the mu-ref warning fires on relations that generate no term (low)

The loop filters on `rel.op == Add` and `bears_eta`, but not on whether the relation actually produces a term. `relation_effect` returns `None` for `CovariateForm::None` and for any relation where `needs_data()` holds, so `CL ~ WT none +` — which the PR allows so a search can round-trip a tested-and-rejected line with its operator — produces no sum, mu-referencing stays on, `mu_refs` still holds `ETA_CL`, and the warning says the opposite. Filter on relations that emit a term.

Separately, worth double-checking the warning's framing against #619, which is open and describes this exact shape (typical value not log-linear in one θ) collapsing the covariate θ onto its bound under SAEM — 0.000000 vs NONMEM 1.41, OFV 1214 vs 734.6. The code comment argues #619 was the effect being *dropped* and this is the safe direction; if that holds the warning is fine, but finding #2 makes that bound closer than it looks.

### 5. `src/api/postfit.rs:1346` — `covariate_relation_estimates` carries no `op` (low)

The post-fit readback an SCM harness or ferx-r consumes is byte-identical for `CL ~ WT linear` and `CL ~ WT linear +`. Two different models report the same relation.

### 6. Operator dropped from search reporting (low)

- `crates/ferx-tools/src/globalsearch/mod.rs:245` — `Axis::label` uses `form_label()`, so a pair offered with both operators renders `["none", "linear", "linear"]`; same at 857/865 for the feature vector.
- `crates/ferx-tools/src/covsearch/mod.rs:1091` — `features_of` does `v.set(e.pair_key(), e.form_label())`, so both candidates table as `KA-WT = linear`. Exactly the space `an_additive_effect_is_a_candidate_and_adds_exactly_one_free_theta` sets up.

Fit reuse is keyed on the canonical model-text hash, so no wrong fit is served — this is reporting and resumed-journal legibility only.

### 7. `crates/ferx-tools/src/covsearch/mod.rs:495` — two forced effects on one pair abort the search (low, pre-existing)

`in_base` is tested against `existing`, snapshotted from the base model and never updated as forced effects are applied, so two forced effects on the same pair both reach `AddCovariateRelation` and the second fails with "`[covariate_model]` already declares `CL ~ WT`". Already reachable via two forms; `COVARIATE(CL,WT,lin); COVARIATE(CL,WT,lin,+)` newly reaches it too. Clean error, not a wrong model, but the optional branch already note-and-skips — aborting here is unfriendly and the message says nothing about the one-line-per-pair rule.

---

### Checked and sound

`split_trailing_op` precedence vs the `=>` clause and its error paths; the per-operator neutral guard; product/term placement and `is_plain_product` parenthesisation (rejects a leading `if`, top-level `+`/`-`, leaves `/`, `^`, unary signs alone); `canonical_hash` (`squeeze` preserves `+`); the mu-ref detector genuinely not matching a sum; pair-level mutual exclusion in both search layers; `mfl.rs:1098` mapping `Arg::Wildcard → Multiply` (correct Pharmpy semantics — `*` is the multiplicative operator there, not a wildcard); the anchor's `worst_pred_error` asserting `is_finite()` before folding into `f64::max`; `the_additive_term_is_what_is_being_anchored` being a real non-degeneracy check.

roninsightrx:
## Review — additive (`+`) covariate effects

Read the three places the hints point at. The desugar itself is right: `insert_effects` rebuilding the product from the *original* RHS makes declaration order irrelevant, the `is_plain_product` parenthesisation covers the `if (c) a else b` case, per-operator `neutral()` is applied at the single guard site, and pair-level dedup holds in `parse_relations`, `add_covariate_relation` and covsearch's `remaining.retain(|e| e.pair() != …)`, so the two competing candidates cannot both land on one line. `split_trailing_op` is safe against every existing line shape (comments are stripped in `extract_blocks`, `expr` requires quotes and closes on `)`, `=>` is split first).

Four findings, one of which I think blocks — it makes the covsearch candidate this PR exists to enable estimate against a pinned bound.

### 1. `src/parser/covariate_model.rs:808` (`linear_family_thetas`) — the multiplicative θ bounds are applied unchanged to additive relations

The bounds are documented as `θ ∈ [1/(c−max), 1/(c−min)]` precisely because that is "what keeps `1 + θ·(COV − med)` positive". Under `+` there is no leading `1` — the term is `θ·(COV − c)` added to a parameter whose scale is unrelated to 1 — so those bounds silently cap the additive term at roughly ±1 absolute unit over the covariate's range.

Concrete: `COVARIATE?(CL, WT, lin, +)` on `data/two_cpt_oral_cov.csv` (WT 45–93.7, centre 70) yields `θ ∈ [−0.0422, 0.0400]`. The PR's own docs use θ = 0.05 as the illustrative slope and the anchor uses 0.04 with *explicit* `(−1, 1)` bounds — i.e. both realistic values sit at or outside the auto bound. `Effect::relation()` passes `thetas: Vec::new()`, so a covsearch additive candidate always takes these defaults: it is estimated against a pinned bound and its ΔOFV/LRT under-detects the effect, with no warning. The anchor cannot see this because it supplies its own bounds.

Wants an operator branch — under `+` the constraint isn't positivity of a factor, so a scale-free default (or a bound derived from the parameter's own scale) is the right shape.

### 2. `src/parser/covariate_model.rs:820` — the "centre must lie strictly inside the observed range" hard error rejects legitimate additive models

Same function. `CL ~ WT linear(center = 0) +` — a plain `θ·WT` term, well-posed additively and the natural spelling for an uncentred additive slope — is rejected because `below = 0 − 45 < 0`, with an error saying the bounds are needed to "keep the covariate factor positive". Under `+` there is no factor and nothing to keep positive. The gate and its message need the same operator branch as (1).

### 3. `src/types.rs:1778` / `src/api/postfit.rs:1348` — the `FitResult` covariate-model echo drops the operator

`CovariateRelationEstimate` gained no `op` field and `covariate_relation_estimates` does not populate one, so `CL ~ WT linear` and `CL ~ WT linear +` produce byte-identical rows in `{model}-fit.yaml`. That struct's own doc calls it "the table a covariate search reads back after a fit", and ferx-r / `ferx-cli` (`allometry_cmd.rs:249`) read it — so a consumer reconstructing or reporting the model from a fit silently reports a `+` relation as multiplicative. `expression`'s doc comment ("`form` + `center` + `thetas`" determine the factor) is also now false for additive relations. Adding `op` here would have been cheaper than the three public-API items already added, and by the PR's own argument for `CovariateRelation::op` ("without it a search cannot tell which combination a base-model relation uses") this one is needed for the same reason.

### 4. `src/parser/model_parser.rs:3811` — the mu-referencing warning fires regardless of estimation method

The text asserts "SAEM/IMP use the numerical M-step for these etas; FOCE/FOCEI are unaffected", but it is pushed into `parse_warnings` unconditionally at parse time, so every FOCEI fit of a model with an additive relation prints a warning about a path it never takes. `model.covariate_model` is available alongside `fit_options.method` at this point; gating it (or downgrading the wording) keeps it signal.

### On the open questions

1. **The 16.02 offset** — yes, open the issue, and cancelling it against a null twin is the right call for this PR. It is a known property of that dataset, orthogonal to the operator, and an absolute anchor there would be the false green the PR itself documents at slope 0.05.
2. **`none +`** — leaving it inert is fine; round-tripping a rejected candidate with the operator it tested is worth more than the error.

roninsightrx:
Fixes for all four findings are up as **#1319**, stacked on this branch (base `worktree-feat-1313-additive-covariate`) rather than pushed here — this worktree is checked out and mid-merge in another session, so I did not want to race it. Merge it into this branch, or cherry-pick, whichever is tidier with your in-flight merge.

Summary: additive θ defaults are now scale-free (`±1e6`) instead of PsN's factor-positivity bounds; the centre-inside-range gate is multiplicative-only, with the constant-covariate rejection kept for its own reason; `CovariateRelationEstimate` gains `op`; the mu-referencing warning is scoped to SAEM/IMP/IMPMAP/BAYES and names the method. Six mutations, each reddening its own test — including two existing tests that needed `method = saem` added, because after the method gate the *gate* would have supplied their green.

Two design calls in there worth your eye rather than mine: whether `±1e6` is the right additive default, and whether the constant-covariate check should stay best-effort (it fires only for a symbolic centre). Both flagged in the PR's reviewer hints.

Issue #1310, opened 2026-09-09
feat(tools): dcm_suggest — derive [covariate_nn] inputs/layers/init/normalisation and nn_l2/nn_smooth starting points from a base fit + GAM screen


roninsightrx:
## Summary

A deep compartment model (`[covariate_nn]`) has ~10 knobs, and today every one of them is picked blind: `inputs`, `outputs`, `layers`, `activation`/`output`, `center`/`scale`, `init`, `nn_l2`, `nn_smooth`, plus the method / multi-start settings around it. The cost of guessing wrong is documented: on the vancomycin UVM set a 2→8→3 DCM beat the 5-covariate analytic model on training OFV and lost badly on 5-fold holdout (1338 vs 1149 −2LL), with `ADD_ERR` collapsing to ~0.02 and a non-monotone CL-vs-FFM slice. The two knobs that did the damage were `inputs` (too many for 84 subjects) and `nn_smooth` (off).

Most of those knobs are derivable, cheaply, from things a covariate workflow produces anyway: the dataset, the base-model fit, and a GAM screen of that fit. This issue proposes `ferx_tools::dcm::dcm_suggest(fit, pop, model)` (+ thin `ferx dcm-suggest` CLI) that turns those into a recommended `[covariate_nn]` + `[fit_options]` stub, with the reasoning attached, and a small evaluation grid around it.

## What each source can see

| knob | dataset alone | + base fit | + GAM screen |
|---|---|---|---|
| `center`, `scale` | mean / sd per covariate | | |
| weight budget → `layers` cap | `N_subjects` (`n_weights ≤ N/10`) | | |
| `outputs` | | TV params with an η in `[individual_parameters]` | restrict to ETAs with any ΔAIC > 0 |
| `init` | crude: `inits_from_nca` | base-fit TV values (`theta` via `mu_refs`) | |
| `inputs` | candidates + collinearity | | drop covariates with ΔAIC ≤ 0 on every ETA |
| curvature → `layers` width, `nn_smooth` | | | `best_form` ∈ {Linear, Spline{df:2}, Spline{df:3}} |
| trust / regularize harder | | `shrinkage_eta` per ETA | `EtaGamResult.shrinkage` |
| σ / Ω to pin in a first phase | | base-fit values | |
| "is a DCM worth it at all" | | | all winners Linear → no |

[...]

Issue #755, opened 2026-07-08
Allow restart of a stopped run


roninsightrx:
If a run gets stopped by accident, we should be able to restart from the last waypoint. So FeRx should write an intermediate file (e.g. `run1.tmp`) during the run at regular time intervals, e.g. every 5 minutes. So for fast runs it shouldn't write anything, as runs can just be restarted, but for long runs this is an essential feature.

Restarting should be very easy. Design question: if a tmp file is present, should restart be default, or should default be to overwrite? In former option, would require a `clean` argument to be available, in latter option would require a `from_tmp` argument to be available. 

This should of course only work when there is a temp file for that run. It should show a message that it found the temp file, and that it's continuing from there. The trace in the console and on disk should also be restored.

The tmp file is removed once the run finishes properly. Both in CLI and R there should be a `clean` argument that can be used to force a start from scratch (similar to just removing the tmp file).

This should work for all estimation methods.

Issue #59, opened 2026-05-20
bug(outer_optimizer): final OFV worse than best-seen due to stagnation guard short-circuit


TeunP:
## Description

When the stagnation guard triggers during NLopt optimization, the optimizer short-circuits subsequent evaluations but NLopt still returns the **last evaluated point** rather than the **best-seen point**. This means the final OFV can be worse than intermediate values observed during the trace.

## Observed behaviour

Fitting the bundled warfarin example with default settings:

```
Eval   97: OFV = -285.351919
Eval  106: OFV = -285.351969   ← best seen
Eval  112: stagnation guard triggered (no improvement below 1e-3 in last window); next eval will short-circuit
NLopt finished: FtolReached
Final OFV = -285.351681        ← worse than eval 106
```

The final OFV is **worse** than what was seen at eval 106. The downstream covariance step then fails because the returned theta is slightly off the true minimum, leaving the Hessian poorly conditioned.

## Expected behaviour

The optimizer should track a `best_theta / best_ofv` accumulator inside the NLopt objective closure and restore `x0` to the best-seen point after `opt.optimize()` returns, before passing it to the covariance step.

## Root cause

In `src/estimation/outer_optimizer.rs`, the NLopt objective closure updates `x0` via NLopt's internal state but does not maintain an independent best-seen accumulator. When the stagnation guard short-circuits by returning a large penalty value, NLopt may step away from the best point before declaring convergence.

## Fix

Add a `best_x / best_ofv` pair (behind a `Mutex` or passed through the closure state) that is updated whenever a new minimum is found. After `opt.optimize()` returns, copy `best_x` back into `x0` before computing the covariance matrix.

## Reproducible with

```r
ex <- ferx_example("warfarin")
fit <- ferx_fit(ex$model, ex$data, verbose = TRUE)
# Observe: Final OFV in trace is worse than best intermediate OFV
```

TeunP:
@roninsightrx : it seems the warfarin examples are not converging but that could just be the example. I have other runs that do quite well

roninsightrx:
OK, fix is coming in. 
I also updated the message, because it read like an error. Updating to:
```
Eval NNN: stopping early — OFV has converged (no improvement above 1e-3 in last window). This is normal convergence behaviour, not an error: further evaluations are unlikely to find a better solution.
```

TeunP:
@roninsightrx : shall I link this commit to the ferx download ?

PR #1318, opened 2026-09-10
fix(estimation): checkpoint the best point seen, not the current eval [closes #1317]

### LINKED ISSUE #1317: bug(checkpoint): .tmp stores the latest eval point, not the best — a write landing on a line-search probe records OFV 2.7e6 on a fit sitting at 51786
## Summary

The `{model}.tmp` checkpoint (#755) stores the parameter vector of the **most recent objective evaluation**, not the best point seen so far. When the write happens to land on a line-search probe, the checkpoint holds a garbage point.

Observed on a `[covariate_nn]` FOCEI fit (6→32→4, 356 weights, 4750 subjects, NLopt L-BFGS) that had plateaued at OFV 51786.14 for ~100 evals:

```
"ofv": 2755477.3924918673,
"iter": 821,
```

i.e. the checkpoint's own `ofv` was 2.76e6 while the optimizer's best was 51786 — the L-BFGS line search had just probed a far-away point when `is_due()` fired. Resuming from that file restarts the fit from the probe, and anything that reads the checkpoint as "where the fit is" (a scorer, a progress monitor, `ferx-r`) gets a point that is 5 orders of magnitude off.

## Where

`src/estimation/outer_optimizer.rs` — in the objective closure:

```rust
if crate::io::checkpoint::is_due() {
    let packed = crate::estimation::parameterization::pack_params(&params);
    crate::io::checkpoint::maybe_write(state.n_evals, ofv_clean, &packed);
}
```

`params` / `ofv_clean` are the current eval's. The `Checkpoint::ofv` doc says "Best OFV seen so far (informational)", which is what the writer should be tracking but is not. Same shape at `outer_optimizer.rs:~2829` (the scaled path), `gauss_newton.rs:390` and `saem.rs:3336` (SAEM's `cond_nll` is a different story — there the current iterate *is* the state, so that one is probably fine).

## Fix

Track `(best_ofv, best_packed)` in the optimizer state and hand **that** to `maybe_write`, with `iter` = the eval at which the best was seen. Cheap: one `Vec<f64>` clone per *improving* eval, no per-eval allocation on the default path (the existing `is_due()` gate stays). Resume then restarts from the best point, which is also the right warm start for L-BFGS/BOBYQA (they rebuild their own curvature anyway).

Regression test: drive the objective with a scripted sequence `[100, 50, 1e6]` with `is_due()` forced true on the third call and assert the checkpoint holds the second point.

## Context

Found while holdout-scoring DCM fits for #1310 — wanted to score a plateaued-but-unconverged fit from its checkpoint (the stagnation guard needs `3·(n+1)` = 1101 flat evals to stop a 366-coordinate problem, several hours at that size), and the checkpoint was unusable.

roninsightrx:
## Why

The `{model}.tmp` checkpoint (#755) stored the parameter vector of the **most recent objective evaluation**, not the best point the fit had reached. The objective runs at every point the optimizer probes — including the trial steps a line search rejects — so a write whose interval elapsed mid-line-search recorded a throwaway point.

Observed on a `[covariate_nn]` FOCEI fit (6→32→4, 356 weights, 4750 subjects, NLopt L-BFGS) that had been plateaued at OFV 51786.14 for ~100 evals:

```
"ofv": 2755477.3924918673,
"iter": 821,
```

The checkpoint's own `ofv` was 2.76e6 while the optimizer's best was 51786 — five orders of magnitude off. Resuming from that file restarts the fit at the probe, and anything that reads the checkpoint as "where the fit is" (a resume, a progress monitor, a holdout scorer, `ferx-r`) gets that point. The `Checkpoint::ofv` doc already said "Best OFV seen so far", which is what the writer should have been tracking and was not.

Found while holdout-scoring DCM fits for #1310: scoring a plateaued-but-unconverged fit from its checkpoint is exactly the use case, and the file was unusable.

## What changed

`BestPoint` (new, `pub(crate)`, in `src/estimation/outer_optimizer.rs`) becomes the **single** tracker for "where the fit actually is".

- **NLopt driver.** It already kept one for the #59 final restore — an `Option<(Vec<f64>, f64, f64)>` behind a mutex. That tuple *becomes* a `BestPoint`, so the restore and the checkpoint read the same incumbent by construction rather than by two rules that can drift apart. The checkpoint site now unscales the best-seen `xs` instead of packing the current eval's params.
- **Built-in BFGS / L-BFGS loop and Gauss-Newton loop** get one each. Their accept criteria (Armijo; `rho >= 0.25`) already make the *written* iterate monotone in the **penalized** objective — but not in the clean OFV the checkpoint reports, which can tick up under NN regularization. One rule across the drivers is worth more than three separate arguments about why each is incidentally safe.
- **SAEM keeps writing directly.** Its current iterate *is* the state being checkpointed; there is no line search to confuse it with. The guard test below encodes that as an explicit exemption, not an oversight.

Ranking is on `ofv` — the objective the optimizer actually minimises (penalized under `nn_l2` / `nn_smooth`) — so a heavier-λ incumbent can never lose to a point that merely fits the data better. `ofv_clean` (the unpenalized −2LL) travels with its own point, for the consumers that must report or compare an unpenalized number: the plateau self-consistency check and the checkpoint's `ofv` field. `iter` is now the eval at which the best was seen, not the eval that happened to be current.

`observe` also closes a NaN trap the tuple had: `5.0 < NaN` is false, so a run whose *first* eval blew up kept that point as its incumbent for the entire fit — and handed it to the #59 restore as well as to the checkpoint.

## Alternatives considered

- **Track `(best_ofv, best_packed)` as a fresh pair in each optimizer**, as the issue suggests. Works, but in the NLopt driver it puts a *second* definition of "the best point" a few lines from `best_seen`, the exact configuration that lets two answers drift. Reusing `best_seen` costs nothing per eval (the unscale happens behind `is_due()`) and makes the #59 restore and the checkpoint the same object.
- **Fix only the NLopt site** — the one the bug was observed on, and arguably the only site where the written point can be non-monotone in the penalized objective. Rejected: the GN/BFGS safety is emergent from their accept criteria and invisible at the write site, and their clean OFV is not monotone even so.

## Cross-repo dependency

No public API change (`BestPoint` is `pub(crate)`; the baseline is unchanged). `ferx-r` reads the checkpoint through the existing `Checkpoint` fields, whose meaning is now what their docs already claimed.

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable — no estimator, likelihood or gradient math is touched. The optimizer's trajectory is unchanged: `BestPoint::observe` replaces an identical `is_none_or(|(_, prev, _)| ofv < *prev)` predicate in the NLopt driver, and the GN/BFGS trackers only feed the checkpoint. What changes is which point gets serialised to `.tmp`.

Full `cargo test --lib`: 4319 passed, 0 failed (the #59 restore, the plateau consistency check and the `last_gradient` gate all read the reworked tracker).

## Performance
- [x] No significant impact expected — one `Vec<f64>` clone per *improving* eval (a shrinking minority as a fit descends), against a full population likelihood per eval. The packed-space mapping still happens only behind the existing `is_due()` gate, so the default no-checkpoint-due path allocates nothing.

## Tests

Five Tier-1 tests in `src/estimation/outer_optimizer_tests.rs`:

| Test | Pins |
|---|---|
| `checkpoint_holds_the_best_point_not_the_line_search_probe` | the issue's scripted `[100, 50, 1e6]` sequence with the write landing on the probe — the `.tmp` must hold the 50 point, its OFV and its eval index |
| `best_point_ranks_on_the_penalized_objective_and_reports_the_clean_one` | a lower *clean* OFV must not displace a better *penalized* incumbent |
| `best_point_swaps_a_nan_incumbent_out_but_never_back_in` | both NaN directions, since `<` is false either way |
| `checkpoint_write_maps_the_tracked_point_into_packed_space` | a scaled-space driver cannot silently persist scaled coordinates |
| `write_checkpoint_is_a_noop_before_the_first_eval` | an empty tracker writes nothing rather than fabricating a point |

Plus a guard in `tests/checkpoint_restart.rs`, `optimizer_loops_checkpoint_through_the_best_point_tracker`, counting direct `maybe_write(` callers per driver file (outer_optimizer: 1 — `BestPoint::write_checkpoint` itself; gauss_newton: 0; saem: 1 — the deliberate exemption). **This is the test that dies on the revert.** The unit tests pin the tracker but cannot see the call sites, so reverting either driver to "write the current point" — the exact defect — leaves all five of them green. Per CLAUDE.md, all four mutations were run and each reddened its own side:

| Mutation | Test that died |
|---|---|
| `observe` always adopts the latest eval | the three `best_point*` / probe tests |
| `write_checkpoint` ignores the packed-space map | `checkpoint_write_maps_the_tracked_point_into_packed_space` |
| NLopt site back to `maybe_write(state.n_evals, ofv_clean, &pack_params(&params))` | the guard (2 callers ≠ 1) |
| GN site back to `maybe_write(iter, ofv_clean, &x)` | the guard (1 caller ≠ 0) |

- [x] Not applicable — no new API surface; existing `checkpoint = true` behaviour is corrected.

## Docs
- [x] `docs/` updated for user-visible changes — `docs/model-file/fit-options.qmd`, the `checkpoint` row: says what is saved (the best point, not the last evaluated one) and what `iter` / `ofv` mean.

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), rustdoc, docs lint, public-API baseline

### Example execution
- [x] No example execution step needed — no user-visible change to model files, the CLI surface or fit output.

## Reviewer hints

- **`BestPoint::observe`'s predicate** is the one place worth reading closely: `ofv < b.ofv || (b.ofv.is_nan() && !ofv.is_nan())`. The first clause is the pre-existing rule verbatim; the second exists because a NaN incumbent is otherwise permanent.
- **The `best_seen` → `BestPoint` swap in the NLopt driver** touches four sites (declaration, the `last_gradient` global-best gate, the observe, the #59 restore). The restore now reads `best.ofv_clean` where it read `best_clean` — worth confirming the *clean* number still goes to `best_seen_ofv`, since the plateau self-consistency check compares it against an equally clean `final_ofv` and mixing in a penalized value would loosen that check by exactly the penalty.
- **The GN/BFGS `observe` calls are placed after the accept**, not on the trial point — a tracker fed rejected trials would defeat the purpose.
- Skimmable: the CHANGELOG and docs wording.

## Open questions

None. One thing deliberately left alone: SAEM's write site, which the issue also flagged. Its current iterate is its state — the stochastic approximation has no probe/incumbent distinction — so "best `cond_nll` so far" would be the wrong point to resume from, not a safer one.

https://claude.ai/code/session_0167yZeWGVwK5yGc3vhek2mq


roninsightrx:
## Review finding

**[P2] Scope the new checkpoint contract to the optimizer drivers**

`docs/model-file/fit-options.qmd` now says every checkpoint contains the best point reached and that `iter` / `ofv` are the evaluation index and unpenalized objective at that point. The same unconditional claim appears in `CHANGELOG.md`. That is not true for `method = saem`: this PR deliberately leaves `src/estimation/saem.rs:3330-3336` writing the current stochastic-approximation state with `iter = k` and `ofv = cond_nll` (not the best point and not the marginal unpenalized -2LL). Because checkpoints are consumed by resume/progress/scoring code, the new wording establishes a contract those consumers cannot rely on for SAEM.

Please qualify the docs/changelog to say that the deterministic FOCE/GN optimizer stages save their best incumbent, while SAEM saves its latest state for correct continuation. The `Checkpoint::ofv` field doc should also describe the method-dependent value instead of unconditionally saying `Best OFV seen so far`.

The implementation for the optimizer paths otherwise looks sound: ranking uses the penalized objective while carrying the matching clean OFV, NLopt unscales the tracked point before persistence, and the same tracker feeds final restore.

Validation performed:

- `cargo test --lib`: 4,319 passed, 23 ignored
- `cargo test --test checkpoint_restart optimizer_loops_checkpoint_through_the_best_point_tracker`: passed
- `cargo check --tests`: passed
- Standard CI checks are green; the core coverage shard is still running and Codecov currently shows a provisional 87.76% patch result.

I attempted to submit this as a changes-requested review, but GitHub does not allow that on a PR authored by the authenticated account, so I am posting it as a review comment instead.

roninsightrx:
## Re: [P2] Scope the new checkpoint contract to the optimizer drivers

Agreed, and fixed in 7f878095. The finding is right: the implementation deliberately splits SAEM off, but every piece of prose I wrote stated the new contract unconditionally, which is worse than the old wording — it tells a consumer it can rank or compare `ofv` across checkpoints when for a `saem` stage that number is an iteration's conditional NLL.

Three places now name the split, and each points at `method_chain` / `stage_idx` as the way to tell which stage wrote the file:

- **`Checkpoint::ofv`** (`src/io/checkpoint.rs`) — the field doc no longer says "Best OFV seen so far". It now says the value is method-dependent, spells out both cases (deterministic: best incumbent, unpenalized −2LL, `iter` = the eval it was seen at; SAEM: latest state, that iteration's conditional NLL, `iter` = the iteration index), and says a consumer that ranks or compares checkpoints has to read `method_chain` / `stage_idx` first. `Checkpoint::iter` cross-links to it rather than repeating the split, and the module-level doc and `maybe_write`'s doc lost their "current population estimates" / "current packed optimizer vector" phrasing for the same reason.
- **`docs/model-file/fit-options.qmd`**, the `checkpoint` row — the claim is now scoped to `foce` / `focei` / `laplace` / `gn` / `gn_hybrid`, with SAEM's behaviour and the reason for it (correct continuation of the chain) stated next to it.
- **`CHANGELOG.md`** — same scoping, plus an explicit "a `saem` stage is unchanged" clause so the entry does not read as a change to a behaviour that did not change.

No behaviour change in that commit.

### Also in 7f878095: a third write site was uncovered

Chasing the provisional Codecov number surfaced something worth having independently of the patch gate. The built-in BFGS/L-BFGS outer loop (`Optimizer::Bfgs` / `Optimizer::Lbfgs`, distinct from `NloptLbfgs`) is the **third** checkpoint write site, and the existing `checkpoint_restart` runs route to the NLopt driver and to Gauss-Newton only — so its `observe` / `write_checkpoint` lines were never executed by any test. `builtin_bfgs_fit_checkpoints_and_cleans_up` adds that run (2 iterations, zero-second interval, same shape as the FOCEI and GN cases).

Worth noting for anyone reading the guard test: that gap was *not* visible to `optimizer_loops_checkpoint_through_the_best_point_tracker`, which counts direct `maybe_write(` callers per file and so covers the built-in loop statically without ever running it.

### On the Codecov number

The 87.76% you saw was measured with one shard uploaded (`Tests + coverage (TTE/CTMM endpoints)`; the API reports `sessions: 1` on head). `Tests + coverage (core)` is the shard that runs `--tests`, so the integration coverage of the NLopt and GN write sites is not in that figure yet. I have not waited it out — I will re-check once it lands rather than claim it is resolved.

Validation on the review commit: `tools/preflight.sh` green (fmt, the 5 check feature sets, clippy `--all-targets`, rustdoc, docs lint, public-API baseline); `cargo test --test checkpoint_restart` 9 passed. The earlier `cargo test --lib` 4319-pass run still stands — that commit touches only doc comments, the changelog, a `.qmd` and one integration test.

roninsightrx:
Follow-up on the Codecov note above: `Tests + coverage (core)` has landed (41m28s) and **`codecov/patch` now passes**. The provisional 87.76% was the endpoints shard alone — the NLopt and Gauss-Newton write sites are measured by the core shard's `--tests` run, and the built-in BFGS loop by the run added in 7f878095. All checks green.

PR #1319, opened 2026-09-10
fix(parser, api): additive covariate defaults, and echo the operator [PR #1316 review]

roninsightrx:
## Why

Review of #1316 (https://github.com/FeRx-NLME/ferx-core/pull/1316#issuecomment-5612676952). Four findings, all in the additive-covariate feature, none in the desugar itself — `insert_effects`, the parenthesisation, the per-operator `neutral()` and the pair-level dedup all check out. They are in what the block *derives around* the new operator.

**Stacked on `worktree-feat-1313-additive-covariate`, not on `main`** — merge it into that branch before #1316 lands. (Its own worktree was locked by another session mid-merge, which is why this is a separate branch rather than commits on #1316.)

1. **The θ bounds are PsN's multiplicative ones, applied to additive relations too.** `linear_family_thetas` derives `θ ∈ [1/(c−max), 1/(c−min)]`, and its own doc says why: that is "what keeps `1 + θ·(COV − med)` positive". Under `+` there is no leading `1` — the term is `θ·(COV − c)` added to a parameter whose scale is unrelated to 1 — so the bounds cap the whole added term at roughly ±1 *absolute* unit over the covariate's range. On `WT ∈ [45, 93.7]` centred at 70 that is `θ ∈ [−0.0422, 0.0400]`: #1316's own docs use `0.05` as the illustrative slope and its NONMEM anchor uses `0.04` with **explicit** `(−1, 1)` bounds, so both realistic values sit at or outside the auto bound. `Effect::relation()` passes `thetas: Vec::new()`, so a covsearch additive candidate always takes these defaults — estimated against a pinned bound, its ΔOFV/LRT under-detecting the effect, with no warning. The anchor could not see this because it supplies its own bounds.

2. **The centre gate is part of the same positivity story and rejects legitimate additive models.** `CL ~ WT linear(center = 0) +` — a plain `θ·WT`, the natural spelling for an uncentred additive slope — is rejected because `below = 0 − 45 < 0`, with an error saying the bounds are needed to "keep the covariate factor positive". Under `+` there is no factor.

3. **The `FitResult` echo drops the operator.** `CovariateRelationEstimate` gained no `op`, so `CL ~ WT linear` and `CL ~ WT linear +` produce byte-identical rows in `{model}-fit.yaml`. That struct's own doc calls it "the table a covariate search reads back after a fit", and ferx-r / `ferx-cli` (`allometry_cmd.rs:249`) read it — a consumer reconstructing the model from a fit silently reports a `+` relation as multiplicative. By #1316's own argument for `CovariateRelation::op` ("without it a search cannot tell which combination a base-model relation uses"), this one is needed for the same reason one layer out.

4. **The mu-referencing warning fires regardless of estimation method.** Its text asserts "FOCE/FOCEI are unaffected", yet it is pushed into `parse_warnings` unconditionally at parse time — so every FOCEI fit of a model with an additive relation prints a paragraph about a path it never takes.

## What changed

**Additive defaults are scale-free** (`additive_linear_family_thetas`, `ADDITIVE_THETA_BOUNDS = (−1e6, 1e6)`). The null stays at 0; `linear_relative`'s init keeps the `θ_rel = c·θ_abs` identity, so both spellings still start the optimiser at the same covariate effect. `categorical` loses its `θ_k > −1` floor under `+` — that floor is exactly `1 + θ_k > 0`. `exponential` / `power` are unchanged: their `(−100, 1e6)` is an exponent scale, not a positivity bound.

The additive path consults the data for **one** thing: a covariate that is constant makes `θ·(c₀ − c)` a constant shift, confounded with the parameter's typical value, and no bound rescues that. Everything else is scale-free, so an additive relation with a literal centre now resolves at parse time without a dataset. That asymmetry with the multiplicative path — where the *bounds* need the spread, so the relation stays unresolved until data arrives and the degenerate case is always caught — is recorded on the function, along with the consequence: with a literal centre the check is never handed a summary.

The centre-inside-range gate is now multiplicative-only, and its message points at `+` as one of the ways out.

`CovariateRelationEstimate::op` carries `CovariateOp::label()`, is written to the fit YAML, and is `#[serde(default)]` to `"*"` so a file written before #1313 still deserializes.

The mu-ref warning is scoped to the methods that actually read `mu_refs` — SAEM / IMP / IMPMAP / BAYES, i.e. the consumers under `src/estimation/` — and names the method it applies to. Only the *message* is gated; the classification (`mu_refs`, `has_conditional_eta_params`) is untouched. It reads the method the model file states, which is all a parse-time warning can see; a caller that parses a FOCEI file and overrides `FitOptions::method` in code gets no warning, and that limit is stated at the site.

## Alternatives considered

- **Deriving an additive bound from the parameter's scale** rather than leaving it open: the block does not know that scale (`TVCL` is a θ whose init it could read, but a relation may attach to a parameter built from several). An unbounded default with a documented "state your own bound" is honest; a bound derived from the wrong scale repeats the bug in a subtler form.
- **Dropping the centre gate entirely under `+`**: would have lost the constant-covariate rejection, which is a real identifiability failure rather than a bounds artifact. Kept, with its own message.
- **Rewording the mu-ref warning instead of gating it**: leaves the noise on every FOCEI fit, which is the actual complaint.

## Cross-repo dependency

`CovariateRelationEstimate` is only ever *read* in ferx-r — `grep` finds `covariate_relations: Vec::new()` and nothing that constructs the struct — so the added field breaks no build there.

## Breaking changes

- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated with `tools/update-public-api.sh` (+1 field, both re-export paths)

| Item | Needed by |
|---|---|
| `CovariateRelationEstimate::op` | any caller reading the covariate model back off a fit — ferx-r's echo, `ferx allometry` (`allometry_cmd.rs:249`), a covariate search reading a base model's relations. Nothing else in the record distinguishes the two operators, so the existing surface cannot express it. |

## Numerical validation

- [x] Not applicable in the "second engine" sense — no numerical path changed. The defaults are *inits and bounds*; the desugared expression, the sensitivities and the objective are untouched, and #1316's NONMEM anchor (which states its own θ bounds explicitly) is unaffected either way. What is validated instead is the property the finding is about, against the numbers the review measured:

```
# data/two_cpt_oral_cov.csv weight span, centred at 70
multiplicative   θ ∈ [1/(70−93.7), 1/(70−45)]  =  [−0.0422, 0.0400]
additive         θ ∈ [−1e6, 1e6]

# the straddle the test asserts
0.05 (docs' illustrative slope)  >  0.0400   →  outside the multiplicative bound
0.05                             <  1e6      →  inside the additive one
```

- [ ] Warfarin example (untouched)

## Performance

- [x] No significant impact expected — parse-time constants.

## Tests

- [x] Unit tests added in the same modules — 6 in `src/api/tests/covariate_stats_tests.rs`, 1 in `src/parser/covariate_model_tests.rs`, 1 assertion pair in `src/io/output.rs`.
- [x] Regression tests that fail without the fix — **six mutations, each reddening its own test**:

| Mutation | Test reddened |
|---|---|
| additive linear family back on the multiplicative bounds | 3 (bounds, centre-outside-range, constant-covariate) |
| `CovariateOp::Add => (-1.0, 5.0)` for categorical | 1 |
| constant-covariate rejection removed | 1 |
| `method_uses_mu_refs &&` removed | 1 |
| `op: rel.op.label()` → a constant `"*"` | 1 |
| the YAML `op:` line removed | 1 |

Three assertion-quality points, per the CLAUDE.md rules:

- **The bounds test asserts the straddle, not two numbers.** `0.05` is checked to be *outside* the multiplicative bound before it is checked to be inside the additive one, so the pair cannot pass by both arms drifting; the multiplicative arm is asserted against PsN's table unchanged, so it cannot pass by widening both.
- **A fast path was supplying the answer in two existing tests.** `a_parameter_with_no_eta_gets_no_mu_reference_warning` and the multiplicative half of the mu-ref differential pair used no `[fit_options]`, so after the method gate they would have gone green because the *method* suppressed the warning, not because the parameter carries no η. Both now state `method = saem`; without that the assertion could not fail.
- **The constant-covariate check is best-effort and the test says which case reaches it** — a symbolic centre. The multiplicative path catches it always (the relation stays unresolved until data arrives); the additive one has nothing else that needs data. Recorded on the function rather than left for a reader to discover.

Full workspace suite green (`cargo test --workspace --features ferx-core/ci`), `cargo test --lib` 4301 passed / 0 failed.

## Example

<details>
<summary>Example</summary>

```toml
[covariates]
  CRCL continuous

[covariate_model]
  CL ~ CRCL linear(center = 100) +
```

```
# generated [parameters] line — before / after
  theta THETA_CL_CRCL(2e-5, -0.02, 0.0333)   # PsN's factor-positivity bounds
  theta THETA_CL_CRCL(0.001, -1e6, 1e6)      # scale-free: the null is 0

# {model}-fit.yaml
covariate_model:
  - parameter: "CL"
    covariate: "CRCL"
    form: "linear"
    op: "+"           # ← new; without it this row is its multiplicative twin
    center: 100
    thetas:
      - name: "THETA_CL_CRCL"
        estimate: 0.021044
        se: 0.004310
```

</details>

## Docs

- [x] `docs/model-file/covariate-model.qmd` — a new "Additive defaults are scale-free" section (the per-form table, the measured `[−0.042, 0.040]`, the relaxed centre rule, the advice to bound it yourself once the parameter's scale is known), plus `op` in the fit-YAML echo example and its prose
- [x] `docs/tools/covsearch.qmd` — why a search candidate, which never names its θ, depends on this
- [x] `CHANGELOG.md` — the #1313 `[Unreleased]` entry extended rather than a second entry, since none of this has shipped
- [ ] New pages (none)

## Checklist

- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched; nothing here changes an expression, only the inits and bounds a θ is declared with
- [ ] `Cargo.toml` version bumped (not breaking)

### ferx-site
- [x] No new DSL syntax — this changes the defaults behind syntax #1316 already documents
- [ ] No new example `.ferx` or data file

### ferx-book
- [ ] Same as #1316: worth a mention in a covariate-model chapter if one exists, not opened

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` changed

## Reviewer hints

Two places worth the attention:

1. **`ADDITIVE_THETA_BOUNDS` — is `±1e6` the right call?** It is the "no bound" spelling already used for `hockey`'s open ends. The alternative is to derive something from the parameter's scale, which the block does not reliably know. If you would rather an additive candidate carried *some* bound, that is a design call and this is the place to make it.
2. **The constant-covariate asymmetry** (`additive_linear_family_thetas`' doc comment). The check fires only for a symbolic centre. The alternative — making the additive path wait for data like the multiplicative one — is a false requirement, since nothing else there depends on the dataset, but it does buy uniformity.

Skimmable: the doc/CHANGELOG prose, the one-line public-API diff, the `op:` plumbing.

## Open questions

1. **Where should this land?** Stacked here for merge into `worktree-feat-1313-additive-covariate`; happy to have #1316's author cherry-pick instead if that is tidier with their in-flight merge.
2. **Should `exponential` / `power` change under `+` too?** Their `(−100, 1e6)` is an exponent bound, not a positivity one, so it is left alone — but the additive template is `exp(θ·Δ) − 1` and `(COV/c)^θ − 1`, and a case could be made that `−100` is meaningless there as well.

https://claude.ai/code/session_01L7fopavQHDzsWEAGhoaQkV


roninsightrx:
Merged the updated base in (`f8c03ce8`) — #1319 is **MERGEABLE / CLEAN** again.

The base had picked up #1315 (`categorical2`), which lands in the same two places this branch touched, so the conflict was semantic rather than textual:

- **`build_thetas`' categorical arm** is now two independent axes applied in turn rather than one four-case table. The **operator** picks the bounds — PsN's `(-1, 5)` under `*`, `ADDITIVE_THETA_BOUNDS` under `+` — and the **shape** then maps them, `categorical2` taking the image under `θ_cat2 = 1 + θ_cat`. #1315's multiplicative defaults are bit-identical through this (`-0.001 + 1.0` is exactly `0.999`, so the rendered `(0.999, 0, 6)` is unchanged), and the reparameterization now extends to `+`, where `relation_effect` emits `θ_k − 1` against `categorical`'s `θ_k` — so `cat2 +` stays the same model as `cat +`. A second hardcoded pair of bounds would have broken exactly that.
- **Docs**: both conflicting paragraphs kept; the additive-defaults table gains a `categorical2` row and the sentence naming the factors those bounds keep positive gains `θ_k > 0`.

New test `additive_categorical2_defaults_stay_the_image_of_categorical` pins that the two axes compose, asserting both arms so neither can be dropped silently. Two mutations on the merged code, each reddening its own tests: the operator branch back to `(-0.001, -1, 5)` (2 tests), the shape map back to a hardcoded `(0.999, 0, 6)` (1 test).

Full workspace suite green post-merge, preflight green, public-API baseline unchanged by the merge.

roninsightrx:
## Review — additive covariate defaults + echoed operator

Read the full diff against `worktree-feat-1313-additive-covariate`. The scale-free additive bounds and the `op` echo both look right, and the tests around them are genuinely differential. Four things I think are real:

### 1. `src/parser/model_parser.rs:3833` — the method gate ignores a method *chain*

`method_uses_mu_refs` reads `fit_options.method` only. `parse_fit_options` sets `opts.method = *chain.last()` for the list form (`model_parser.rs:7515`), so a file with

```
[fit_options]
  method = [saem, focei]
```

lands on `EstimationMethod::FoceI` and the warning is **suppressed** — even though the SAEM stage runs first and is exactly the stage that pays the numerical M-step. The inverse case is worse for the message text: `method = [focei, imp]` fires the warning but prints `IMP uses the numerical M-step`, naming only the last stage when several may qualify.

Suggest gating on the whole chain and naming the qualifying stages:

```rust
let mu_ref_stages: Vec<_> = fit_options
    .method_chain()
    .into_iter()
    .filter(|m| matches!(m, Saem | Imp | Impmap | Bayes))
    .collect();
```

`FitOptions::method_chain()` (`types.rs:7905`) already exists for this. The new test `the_mu_reference_warning_is_scoped_to_the_methods_that_read_mu_refs` only feeds scalar methods, so it cannot see this.

### 2. `src/parser/model_parser.rs:3820` — the comment's premise ("FOCE/FOCEI never read `mu_refs`") is not what the code does

The comment says the consumers are "saem, importance_sampling, impmap, bayes — `parameterization` serves those". But `parameterization::compute_mu_k` is called unconditionally from `outer_optimizer.rs` (264/335/442/1117/1154/1820/2324/2526), `gauss_newton.rs:117,271`, `covariance.rs:781`, `run_covariance.rs:278` and `sir.rs:678`, and the inner optimizer works in psi-space `psi = eta_true + mu_k` when it is non-empty (`inner_optimizer.rs:598`). So FOCE/FOCEI/GN *do* read `mu_refs` — as a reparameterization of the inner solve, not the M-step. It is defensible that losing it is not a cliff there, but that is a different claim from "a path that run never takes", and `gn`/`gn_hybrid` are silently in the suppressed set too. Worth restating the comment in terms of the M-step rather than reachability.

### 3. `src/parser/covariate_model.rs:880` — additive `hockey` no longer checks the breakpoint against the data

`additive_linear_family_thetas` drops the whole `strictly inside the observed range` gate and keeps only `min == max`. For `linear` that is the intended relaxation (`center = 0` is the uncentred slope). For **`hockey`** it is not: the template is `if (COV <= b) θ_LO·(COV−b) else θ_HI·(COV−b)`, so a breakpoint below `min` (or above `max`) leaves one arm with no subjects at all — `θ_LO` then has an identically zero gradient. `CL ~ WT hockey(breakpoint = 0) +` on `WT ∈ [45, 94]` now parses and generates two θ, one of which the optimiser cannot move and which makes the covariance step singular. That is precisely the failure the multiplicative gate's doc comment says it exists to prevent ("instead of producing a θ the optimiser cannot move"), and it survives the operator switch because it is about *support*, not about factor positivity. Suggest keeping a breakpoint-inside-range check for `Hockey` under `+`.

### 4. `src/parser/covariate_model.rs:884` — the constant-covariate check is unreachable for a literal centre

The doc comment states this ("an additive relation with a **literal** centre is fully built at parse time and is never handed a summary") and calls it best-effort, but the asymmetry is user-visible: `CL ~ WT linear(center = 70)` on a constant `WT = 70` is *rejected* (the multiplicative path is unresolved until data arrives, so it always sees a summary), while `CL ~ WT linear(center = 70) +` is accepted and yields `θ·(70 − 70) ≡ 0` — a θ with an exactly flat objective, no warning, and a singular covariance step. `needs_data()` (`types.rs:1735`) is what decides this, and the literal-centre additive relation reports `false`. If waiting on data purely for a degeneracy check is unwanted, the cheaper alternative is to run the check at bind time for *every* relation that mentions a covariate, rather than only for the ones that were unresolved.

### Smaller notes (non-blocking)

- `additive_linear_family_thetas` scales the *init* of `linear_relative` by the centre but not the bounds, so the two spellings no longer admit the same interval (±1e6 in θ_rel is ±1e6/c absolute). Harmless at realistic centres, but the multiplicative twin scales both, and the new `the_additive_init_is_the_same_effect_in_both_linear_parameterizations` test only pins the init.
- The docs table lists `exponential`/`power` additive bounds as `(−100, 1e6)`, i.e. the multiplicative row unchanged. Those two are untouched by the fix and their template under `+` is `exp(·) − 1` / `(·)^θ − 1`, which also has no factor to keep positive — worth a sentence saying the row is deliberately shared rather than overlooked.

roninsightrx:
## All six review findings fixed — d2cf79cf

### HIGH — the mu-ref warning read only the last stage of the chain (`model_parser.rs`)

`method = [saem, focei]` leaves `fit_options.method == FoceI`, so the warning was suppressed on exactly the chain that pays for the loss; `[focei, imp]` was labelled FOCEI. Now walks `FitOptions::method_chain()` and names every mu-step stage in chain order (`SAEM / IMP`), with the verb agreeing.

New test `the_mu_reference_warning_reads_every_stage_of_a_chained_method` asserts both directions plus a `[foce, focei]` straddle that must stay quiet. The stage list is matched as the slice between `theta and` and `use` — a bare `contains("FOCEI")` reads the closing sentence and passes either way.

### MEDIUM — the gate's stated premise was false (`model_parser.rs:3820`)

Correct: `compute_mu_k` is called from `outer_optimizer`, `gauss_newton`, `covariance`, `run_covariance` and `sir`, and the inner loop runs in psi-space. The scoping is still right, for a different reason — those callers only *re-centre* the search, no closed-form step is lost — so the comment, the warning's last sentence and the docs callout now say that instead, and `gn` / `gn_hybrid` are asserted into the quiet set explicitly rather than being silently in it.

### MEDIUM — additive dropped the centre-inside-range gate for `hockey` too (`covariate_model.rs:880`)

Right for the single-slope forms, wrong for `hockey`: outside the range one arm holds no subjects, so its θ has an identically zero gradient and the covariance step is singular. Support, not factor positivity — so it survives the drop of the leading `1`. `hockey(breakpoint = 0) +` on `WT ∈ [50, 90]` is now rejected with the range in the message, while `linear(center = 0) +` at the same centre stays legal (asserted as the differential pair, along with `hockey(breakpoint = 70) +` so the check cannot pass by rejecting every additive hockey).

### MEDIUM — the constant-covariate check was unreachable for a literal centre (`covariate_model.rs:884`)

Confirmed: `needs_data()` false → `bind_covariate_stats` returned before summarising → `linear(center = 70) +` on constant `WT = 70` reached the fit as `θ·0 ≡ 0`.

Rather than duplicate the rule, both checks moved **out** of `additive_linear_family_thetas` into one `check_relation_against_data`, called from `bind_covariate_stats` once a population is bound. So there is a single implementation of each rule, and it runs whenever the model meets a dataset — literal centre included. `bind_covariate_stats` no longer short-circuits on `unresolved().is_empty()` alone; it also summarises when some relation `wants_data_check` (additive linear family only — every multiplicative default bound is already a function of the spread, so a degenerate design never gets that far). The re-parse still happens only when something is genuinely unresolved.

An explicit `=> NAME(init, lower, upper)` no longer rescues a constant covariate under `+`: the defect is the design, not the bounds.

### LOW — additive `linear_relative` scaled the init but not the bounds (`covariate_model.rs:900`)

Deliberate, now stated and asserted: multiplicatively the bounds are `1/(c − min)` / `1/(c − max)`, in the same units as the init, so the reparameterization must carry them; additively they are fixed scale-free sentinels, and scaling a sentinel by the centre makes the "effectively unbounded" width depend on the covariate's units (and flips it for a negative centre) without bounding anything. `the_additive_init_is_the_same_effect_in_both_linear_parameterizations` now checks the bounds on both sides.

### LOW — docs table reused the multiplicative `exponential` / `power` row (`covariate-model.qmd:108`)

Now says why it is deliberate: their additive templates are `exp(θ·(COV − c)) − 1` and `(COV/c)^θ − 1`, so θ sits inside an exponent either way, in the covariate's units and not the parameter's — nothing in those bounds encodes a leading `1` to drop.

---

**Mutation-checked, one at a time** — each reddens its own test and no others:

| mutation | dies |
|---|---|
| `method_chain()` → `[fit_options.method]` | `..._reads_every_stage_of_a_chained_method` |
| hockey support gate off | `an_additive_hockey_breakpoint_outside_the_data_is_rejected` |
| constant gate off | both constant-covariate tests (literal + symbolic arms, independently) |
| drop the bind-step check hook | all three |
| `wants_data_check` → `false` | the two literal-centre tests only (the symbolic arm still routes through the re-parse) |
| scale the additive bounds by the centre | `the_additive_init_is_the_same_effect_...` |

`tools/preflight.sh` green; `cargo test --lib --features ci` 4341 passed; `cargo check --tests --features ci` and `ferx-tools --lib` green. CHANGELOG's #1313 entry updated (hockey exception, the data-path move, chained-method wording).

https://claude.ai/code/session_01L7fopavQHDzsWEAGhoaQkV

Issue #1196, opened 2026-09-02
refactor(ode): active_infusions and gated_infusions duplicate the infusion-membership predicate (third twin-drift in this file)


TeunP:
## Summary

`active_infusions` and `gated_infusions` (`src/ode/predictions.rs`) each decide **which doses
contribute a plain `+rate` this segment**, and they decided it differently. #1187 was exactly
that divergence: `gated_infusions` was missing the input-rate suppression, so a `RATE>0` dose
into a built-in absorption compartment was delivered twice on three engines — for two *fitted
objectives*, 31× on the joint PK-TTE cumulative hazard against NONMEM.

PR #1195 fixed the symptom by adding the missing predicate term. It did not remove the
duplication that let the two drift, so the next term added to one of them can drift again.

## What is actually shareable — and what is not

**Do not try to merge the two functions.** Their return shapes differ for a real reason:

- `InfusionInput::Spanning(Vec<(cmt, rate)>)` — the callers (`integrate_segment`, the
  event-driven walk) split the timeline at every infusion edge, so a segment is wholly inside
  or wholly outside the window and no window needs carrying.
- `InfusionInput::Gated(Vec<(cmt, rate, t_start, t_end)>)` — the dense/states callers do
  **not** split at infusion edges, so the window travels with the tuple and
  `wrap_rhs_with_forcings` gates by time inside the RHS.

That difference is load-bearing and should stay.

**What is shareable is the predicate**, i.e. the answer to "does dose *k* contribute a plain
`+rate` into compartment *c* at all":

- `is_real_infusion(d)`
- the input-rate suppression — `input_rate.iter().any(|f| f.cmt == d.cmt_idx())` (#719 gap 2 /
  #1187)
- compartment addressability (`CMT=0`, `cmt >= n_states`)

Today `active_infusions` applies the first two and `gated_infusions` applies the third plus
(since #1195) the second, with the first applied by its callers when they build the `active`
list. Extracting one predicate both call — and keeping the *shape* decisions where they are —
is a modest change, not a large one.

Note the two spellings are **not** interchangeable and the fix has to preserve which is used:
`active_infusions` tests a **runtime forcing slice**, while `input_rate_consumes_cmt` tests
`ode.input_rate`. `src/ode/ekf.rs` passes `&[]` deliberately, because the EKF path never applies
the forcing — so a predicate that consults the *spec* would suppress a rate nothing replaces and
silently delete mass. Any unification must take the slice.

[...]

Issue #1160, opened 2026-08-31
bug(sens): the general rate-off saltation branch is unreachable since #1073


TeunP:
**The general `g⁻−g⁺` rate-off saltation branch is unreachable, and has been since #1073.**

Found while chasing the `codecov/patch` gate on #1136. Recording it rather than removing it
there, because the removal is a behaviour-surface change that deserves its own review.

## What

Both rate-off handlers in `integrate_tvcov_g` split on whether the PK snapshot jumps across the
(non-record) boundary:

```rust
// src/sens/ode_provider.rs, K_INF_END at :6079 and K_ZO_END at :6208
let pre_params = params;
let post_params: &[T] = post_snapshot(p, t_event).unwrap_or(last_params);
if pk_snapshot_equal(pre_params, post_params) {
    inject_rate_saltation::<T>(...)      // cheap closed form
} else {
    general_rate_off_saltation::<T>(...) // full pre/post velocities  ← never taken
}
```

The `else` was added by #653 review #3 to recover the `(J⁺−J⁻)·x` term a forcing-only saltation
misses when a time-varying covariate changes the params across the boundary. That term is real.
The branch that computes it is not reachable.

## Why

`params` is `next_record_params[p]` — the record terminating the segment, i.e. the first record
at index **`>= p`** (`:5096`). `post_snapshot(p, t_ev)` returns the first record **strictly**
later than `t_ev` (`:4979`). A rate-off is not itself a record, and every record co-timed with
one sorts *before* it — `K_INF_END` = 7 and `K_ZO_END` = 8 are the last two kinds, after
`K_DOSE_REC` = 1, `K_PKONLY` = 5 and `K_OBS` = 6. So a co-timed record is excluded from
`next_record_params[p]` as well, and both sides resolve to the same next-strictly-later record.
`pk_snapshot_equal` is therefore always true.

This reads as a consequence of #1073. The comment at `:6071` still describes the pre-side as
"what the outer `params` resolved" from the *previous* record; under the end-of-interval
convention #1073 installed, it is the *next* one — the same object `post_snapshot` finds.

## Evidence

Three independent lines, because the code comments alone point the other way:

1. **Analysis** — the index/ordering argument above.
2. **Runtime probes** — an `eprintln!` on the branch, in both candidate geometries: window end
   *between* records (`t = 2.5`, `WT = 70` at `t = 2` vs `WT = 80` at `t = 4`) and window end
[...]

Issue #1129, opened 2026-08-29
bug(sens): an EVID=3 reset before a lagged infusion's arrival corrupts the analytic `∂f/∂η_LAG` (value unaffected)


TeunP:
An `EVID=3` reset placed **before a lagged infusion's arrival** changes the analytic
twin's `∂f/∂η_LAG` — even when the reset is a physical *no-op*. The prediction value
is unaffected; only the gradient is, so this reaches FOCEI through the EBE search and
the outer gradient without any visible symptom.

Found while adding a reset-in-the-pre-arrival-window test for #1121. It is **not**
specific to steady state, **not** specific to time-varying covariates, and predates
that work: it reproduces with a plain single dose and flat covariates.

## Reproduction

1-cpt IV, `d/dt(central) = -(CL/V)*central`, `ALAG1 = TVLAG*exp(ETA_LAG)`, flat `WT`.
One dose at `t = 0`, `AMT = 100`, `RATE = 25` (`T_inf = 4`), `SS = 0`, so the state is
identically zero until the arrival at `t ≈ 8.41`. Observation at `t = 13`, i.e. after
the infusion end at `≈ 12.41`. `θ = [1.0, 10.0, 8.0, 0.75]`, `η = [0.12, 0.05]`.

Inserting `subject.reset_times = vec![2.0]` — a reset of an already-empty system —
changes nothing about the trajectory, and `obs.f` confirms that:

```text
                       obs.f      ∂f/∂η_CL            ∂f/∂η_LAG
  no reset           7.531115   -2.07188 ✓      +7.14133  ✓ (= FD)
  reset at t = 2     7.531115   -2.07188 ✓      -12.53131 ✗  (FD: +7.14133)
```

The value and `∂f/∂η_CL` are identical with and without the reset, as they must be.
`∂f/∂η_LAG` flips sign. The gap, `−19.67`, is close to the arrival's rate-on
saltation magnitude `(rate/V)·(∂lag/∂η_LAG) = 2.5 × 8.41 = 21.0` decayed to `t = 13`,
which suggests that term is being dropped (or double-subtracted) once `reset_floor`
is finite — `sens/ode_provider.rs`, the `K_DOSE` rate-on branch and its
`boundary_velocity` calls, which take `reset_floor` as an argument.

Same result with time-varying covariates (`−11.162` vs FD `+8.441`) and with an
`SS=1` dose, so neither is required.

## Why it is not caught today

[...]

PR #1320, opened 2026-09-10
fix(sens): one infusion-membership predicate for every rate boundary [closes #1077]

### LINKED ISSUE #1077: bug(sens): a CMT=0 infusion is delivered by the production predictor but skipped by the analytic walk (values diverge)
Found by direct read while responding to the #1071 review (finding #11). Not measured end-to-end — filing it because the two engines demonstrably disagree in source, and the disagreement is in the **value**, not just a derivative.

## The disagreement

`CMT=0` is NONMEM's *default dose compartment*, which ferx resolves to state index 0 (#899). `DoseEvent::cmt_idx()` does that resolution, and the datareader does **not** reject an explicitly written `CMT=0` — only a *missing* `CMT` column defaults to 1 (the comment at `src/sens/ode_provider.rs` says this explicitly, correcting #472 #6 / #473 #3).

Production delivers such an infusion:

```rust
// src/ode/predictions.rs — active_infusions
.filter_map(|(k, d)| {
    if !is_real_infusion(d) { return None; }
    ...
    Some((d.cmt_idx(), rate_eff))
```

`is_real_infusion` is `d.is_infusion() && d.duration > 0.0 && d.duration.is_finite()` — no compartment guard. So a `CMT=0` infusion contributes `+rate` to state 0.

The analytic walk drops it, in three places:

```rust
// src/sens/ode_provider.rs
.filter(|(_, d)| is_inf(d) && d.cmt_raw() >= 1)          // segment forcing
if !is_inf(d) || d.cmt_raw() < 1 { continue; }           // boundary_velocity
.filter(|(_, d)| d.is_infusion() && d.cmt_raw() >= 1)    // (third site)
```

The comments there justify the skip as "rejected up front on both engines (#375 / #899) … unreachable from a validated call", but then immediately note the datareader does not in fact reject it. So for a hand-built spec — or any path that skips validation — the sens walk integrates a *different trajectory* from `compute_predictions_with_tv`, and `f` itself diverges, not only `∂f/∂η`.

## Which side is right

Production's, on the #899 convention: `CMT=0` means the default dose compartment, and the analytical engine resolves it the same way for a bolus. There is no reading under which a `CMT=0` bolus doses state 0 while a `CMT=0` infusion doses nothing.

## Fix

Drop the three `cmt_raw() >= 1` guards in `src/sens/ode_provider.rs` so the walk mirrors `active_infusions`, **or** reject `CMT=0` on an infusion in the datareader so the comments' claim becomes true. The first is the smaller change and matches the bolus convention; the second is defensible if `CMT=0` on an infusion is genuinely meant to be an input error.

Either way it wants a Tier-1 fixture that fits a `CMT=0` infusion and asserts the walk's `f` equals `compute_predictions_with_tv` — the existing `ode_provider_ss_absorption_cmt_zero_*` tests cover the *bolus* side of #899 and would be the model.

Note #1071 makes the saltation self-consistent with whichever behaviour the walk has (it no longer subtracts a rate `boundary_velocity` never added), so this issue is only about the value divergence.

Refs #899, #375, #1071.

hiddevandebeek:
## Why

#1077 reports that a `CMT=0` infusion is delivered by the production ODE predictor but
skipped by the analytic sensitivity walk, so `f` itself diverges between the two.

**That half is already fixed, and the issue's proposed fix is now the wrong direction.**
Measured on `b706b1a3` (2-cpt ODE, `AMT=1000 RATE=200`, static walk):

| t | walk `f` (`CMT=0`) | production `f` (`CMT=0`) | production `f` (`CMT=1`) |
|---|---|---|---|
| 1 | 0.000000 | 0.000000 | 13.684469 |
| 5 | 0.000000 | 0.000000 | 30.479873 |
| 24 | 0.000000 | 0.000000 | 1.393836 |

The issue quotes a body of `active_infusions` that no longer exists. Since `4ea6a024`
(PR #1231, #1196 step 3) both infusion resolvers go through `infusion_contributes`, whose
first line drops `cmt_raw() == 0` — so the repo took the issue's *second* option ("reject
it"), as a side effect of a different PR. Two further premises are stale:

- "the datareader does not reject it" is true but not load-bearing:
  `check_dose_compartments` raises **`E_DOSE_CMT_NOT_INFUSABLE`** for any `AMT>0` `CMT=0`
  infusion on both engines, it is tested (`infusion_into_cmt_zero_is_rejected`), and it is
  documented user-facing behaviour (`docs/data-format.qmd` §`CMT`). The divergence was only
  ever reachable from a hand-built spec.
- "drop the guards so the walk mirrors `active_infusions`" was never available repo-wide:
  on the analytical engine `dose_channel(0)` is `None`, so delivering a `CMT=0` infusion
  there would panic. Rejecting is the deliberate convention.

**What is still live** is the walk disagreeing with *itself*. Four rate-**on** sites spelled
the compartment test `cmt_raw() >= 1`; the rate-**off** saltation at `K_INF_END` did not,
because #899 removed a `d.cmt >= 1` gate there on the premise that the walk delivered such
an infusion. So a lagged `CMT=0` infusion injects the boundary term for a rate that was
never turned on. Measured on `ONECPT_IV_LAG_INF_ODE` (`AMT=100 RATE=40`), event-driven walk:

| t | walk `f` | production `f` | walk `∂f/∂η_LAG` | central FD of production |
|---|---|---|---|---|
| 1.0 | 0.0 | 0.0 | 0.0 | 0.0 |
| 2.0 | 0.0 | 0.0 | 0.0 | 0.0 |
| 4.0 | 0.0 | 0.0 | **+1.8878965164038346** | 0.0 |
| 8.0 | 0.0 | 0.0 | **+1.2133620764183783** | 0.0 |
| 12.0 | 0.0 | 0.0 | **+0.7798348668467985** | 0.0 |

The nonzero entries begin exactly past the bioavailable window end
(`0.5·exp(0.05) + 100/40 ≈ 3.03`), the `K_INF_END` signature. A subject that receives no
drug hands FOCEI a finite lagtime gradient — infinite relative error, not a tolerance
question.

## What changed

One predicate, `dosing::infusion_has_rate_channel`, asked by **all six** infusion-forcing
sites — one on the value path, five on the sensitivity walk — replacing four hand-written
copies plus the one site that was missing it. This is CLAUDE.md's "when the bug report is
*two implementations disagree*, the fix is one implementation", and it is what stops the
rate-on and rate-off halves of the same boundary drifting apart again. `src/dosing.rs` is
the home because it is already the neutral one shared by `pk`/`ode`/`sens`.

Behaviour on every currently-reachable input is unchanged: the value-path edit is a literal
rewrite of the same condition, and a `CMT=0` infusion remains a hard validation error.

Three stale comments corrected in passing, all asserting things false at this SHA — that
the datareader rejects `CMT=0`, that `infusion_spans_segment` carries the compartment test
(it is purely temporal and never has), and that production delivers such an infusion.

Also included, no production change:

- **#1129 regression pin.** An `EVID=3` reset strictly between a lagged infusion's record
  and its arrival. The issue reports `∂f/∂η_LAG` flipping `+7.14133` → `−12.53131` against
  FD `+7.14133`; at this SHA both arms give `+7.141334` and match FD, so the defect is
  gone. The cell is genuinely uncovered though — the nearest tests use a *bolus*
  (`ode_provider_lagtime_reset_hessian_matches_fd_of_grad`) or put the reset *at* a record
  time — so the test is added as a pin. **#1129 stays open**: which commit fixed it is not
  yet identified, and "green now" is not a close.

## Alternatives considered

**Deliver `CMT=0` infusions in the walk** (the issue's preferred option). Rejected: it
contradicts the documented convention, cannot be matched on the analytical engine
(`dose_channel(0) == None`), and would re-open the value divergence in the opposite
direction now that #1196 gave the value path the same test.

**Route the walk to the FD fallback for this input.** Correct but a performance cliff for a
row that is a hard validation error anyway, and it leaves the five spellings in place.

## Cross-repo dependency

No public API change; `infusion_has_rate_channel` is `pub(crate)`.

## Breaking changes

- [x] None

## Numerical validation

- [x] Compared against: prior ferx commit `b706b1a3` (the before/after tables above)
- [x] Not applicable for NONMEM — see below

**On the NONMEM rule.** There is no equivalent run to make: the changed surface is
unreachable from any dataset NONMEM could also be handed, because ferx rejects it
(`E_DOSE_CMT_NOT_INFUSABLE`) before a fit starts. What anchors the fixtures instead is the
`CMT=1` twin, which *is* NONMEM-anchored — the `dose_cmt_*` family (13 evaluation-only runs
on NONMEM 7.6.0, `nonmem_anchor/README.md`) pins infusion compartment routing, and
`ss_cmt0`/`ss_cmt1` pins that NONMEM prints identical `PRED` for `CMT=0` and `CMT=1` on a
bolus. Each test asserts the `CMT=1` arm is live before concluding anything from the
`CMT=0` arm's zeros, so the anchor transfers rather than being assumed. No NONMEM
installation exists in this environment (only committed `.ctl`/`.csv`/`results`), so a new
run could not have been produced here in any case.

## Performance

- [x] No significant impact expected — one `usize` comparison, `#[inline]`, at sites that
      already ran per dose per segment.

## Tests

| test | walk | what only it can see |
|---|---|---|
| `..._cmt_zero_infusion_static_walk_agrees_with_production` | static | the static forcing filter; the static walk has no infusion rate-off saltation at all |
| `..._cmt_zero_lagged_infusion_injects_no_rate_boundary` | event-driven | **the regression**; `pk_snapshot_equal` fast path |
| `..._injects_no_rate_boundary_under_tv_covariates` | event-driven | the general two-sided branch |
| `..._does_not_contaminate_another_doses_saltation` | event-driven | a `CMT=0` window open *under another dose's* saltation |
| `..._reset_before_a_lagged_infusion_arrival_leaves_the_gradient_alone` | event-driven | #1129's uncovered cell |
| `only_cmt_zero_lacks_a_rate_channel` | — | the convention: the predicate disagrees with *both* resolved accessors on `CMT=0` |
| `it_is_independent_of_whether_the_row_is_a_real_infusion` | — | that it is not foldable into `is_real_infusion` |

Non-degeneracy, because "everything is zero" is the easiest assertion to pass vacuously:
each gradient test asserts the fixture reaches the intended walk, that observations straddle
the window end so `K_INF_END` has something downstream to record, that the `CMT=1` twin's
`|∂f/∂η_LAG|` peaks above 1.0 on the *lagtime* axis specifically, and that `f` is exactly
`0.0` — so a predicate that starts *delivering* `CMT=0` infusions fails too, not only one
that stops injecting.

### Mutation results

Every changed site reverted in turn:

| # | site | result | killed by |
|---|---|---|---|
| M1 | `infusion_has_rate_channel` -> `true` | killed | — |
| M2 | value path (`infusion_contributes`) | killed | — |
| M3 | `boundary_velocity` | **survived** | see below |
| M4 | event-walk segment forcing | killed | both lagged tests |
| M5 | rate-on arrival saltation | killed | both lagged tests |
| M6 | **rate-off `K_INF_END`** (the fix) | killed | both lagged tests |
| M7 | static-walk segment forcing | killed | static test |

M6 kills the fast-path arm and the general-branch arm *separately*, per CLAUDE.md's
"mutate each side of a twin separately".

**M3 survives, and I could not close it — stated rather than papered over.** The
contamination enters only as `1/2 r (J- - J+)` in the `delta^2` coefficient (it cancels
exactly in `v- - v+`), so it is observable only when the Jacobian jumps across the boundary.
A runtime probe shows it never does:

```
ZZARR t=13.410168771008193 snapshot_equal=true pre0=1.1638630102738547 post0=1.1638630102738547
ZZBV  t=13.410168771008193 k=0 cmt_raw=0 adding rate=20 strict=true   <- v-
ZZBV  t=13.410168771008193 k=0 cmt_raw=0 adding rate=20 strict=true   <- v+
```

The rate *is* added to both velocities, but both sides read the same snapshot (`CL`
back-solves to `WT = 75`, the `t = 14` record) even with a covariate record at `t = 10`
sitting strictly between the dose record and the arrival. That is **#1160's mechanism, at a
site #1160 does not claim** — it scopes the degeneracy to `K_INF_END`/`K_ZO_END`, and this
is the rate-on arrival, the branch #1060 made unconditional for exactly this geometry.

The guard is kept anyway: it becomes load-bearing the moment `pre != post` is possible, and
its absence would then be a live defect. The call site says so and points at #1160. I have
posted the probe evidence there, including the fork that should be settled before anything
is deleted — whether `pre == post` is the intended convention or is itself the wrong-number
defect, since under it the field jump #1060 targeted is captured by neither side.

## Docs

- [x] No user-visible change — `docs/data-format.qmd` already documents that a `CMT=0`
      infusion is rejected, and that statement is still true. No validated fit changes.

## Checklist

- [x] Functions on the `Dual2` sensitivity path stay differentiable — the change only
      *removes* a jet injection on an input that carries no forcing; the predicate is on
      `DoseEvent`, not on `T`.

## Reviewer hints

The one line that changes behaviour is the `infusion_has_rate_channel(d)` added to the
`K_INF_END` guard in `integrate_tvcov_g`. Everything else is a literal rewrite of an
existing condition, a comment correction, or a test.

Worth a second opinion: whether keeping an unobservable guard in `boundary_velocity` is
right, versus deleting it and letting #1160 restore it if that investigation concludes the
branch should be live.

## Open questions

- #1129 is pinned but not attributed. Should it stay open (my assumption) until a bisect
  names the commit that fixed it?
- #1160 looks like it may be a wrong-number defect rather than dead code. Worth re-titling?

Closes #1077

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -305,8 +305,12 @@ section of the SDLC for the versioning policy).
   the first sample past the window end) for a subject that receives no drug and predicts
   `0.0` everywhere. Reachable only from a hand-built model spec that runs no validation;
   no validated fit changes. The compartment test is now one shared predicate asked by
-  every infusion-forcing site on both engines, rather than four hand-written copies plus
-  one site that was missing it.
+- The analytic ODE sensitivity walk no longer injects a rate-off boundary term for `CMT=0`
+  infusions on an unbound compartment (#1077). `check_dose_compartments` already rejects
+  such zero-order `CMT=0` infusions with `E_DOSE_CMT_NOT_INFUSABLE`, so this change is
+  user-visible only through hand-built specs. The walk now uses the same shared
+  `infusion_has_rate_channel` predicate as production, preventing spurious
+  `∂f/∂η_LAG` jumps when both engines predict `f ≡ 0`.
 - A fit no longer stops on its first evaluation and reports every parameter at its
   initial value (#1290). The outer loop's EBE warm-start cache adopted the empirical
   Bayes estimates of *every* evaluation, including the ones the line search rejects, so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -296,6 +296,17 @@ section of the SDLC for the versioning policy).
 - `cov_inner_tol` now rejects a non-positive or non-finite value at parse time, as
   `fd_hessian_step` already did (#956). Such a value used to parse and then silently make
   every covariance-step EBE reconvergence exhaust `inner_maxiter`.
+- The analytic ODE sensitivity walk no longer injects a rate-off boundary term for a
+  `CMT=0` infusion, whose rate it never turns on (#1077). `CMT=0` is NONMEM's default
+  dose *bolus* compartment and has no meaning for a zero-order input, so both predictors
+  drop such a row and `check_dose_compartments` rejects it outright
+  (`E_DOSE_CMT_NOT_INFUSABLE`) — but when the dose also carried a lagtime the gradient
+  walk still fired the infusion-end saltation, reporting a finite `∂f/∂η_LAG` (+1.89 at
+  the first sample past the window end) for a subject that receives no drug and predicts
+  `0.0` everywhere. Reachable only from a hand-built model spec that runs no validation;
+  no validated fit changes. The compartment test is now one shared predicate asked by
+  every infusion-forcing site on both engines, rather than four hand-written copies plus
+  one site that was missing it.
 - A fit no longer stops on its first evaluation and reports every parameter at its
   initial value (#1290). The outer loop's EBE warm-start cache adopted the empirical
   Bayes estimates of *every* evaluation, including the ones the line search rejects, so
--- a/src/dosing.rs
+++ b/src/dosing.rs
@@ -797,6 +797,36 @@ pub(crate) fn is_real_infusion(d: &DoseEvent) -> bool {
     d.is_infusion() && d.duration > 0.0 && d.duration.is_finite()
 }
 
+/// Whether a zero-order input into this dose's compartment has a rate channel at all —
+/// the compartment half of "this infusion contributes a `+rate`", spelled **once** for
+/// every engine (#1077).
+///
+/// `CMT=0` is NONMEM's *default dose compartment*: defined for a bolus, where both
+/// engines resolve it to state index 0 (#899), and **not** defined for a zero-order
+/// input, which has no default target. `check_dose_compartments` rejects such a row
+/// outright (`E_DOSE_CMT_NOT_INFUSABLE`, both engines), so this is reachable only from a
+/// hand-built spec that runs no validation — but *there* every engine has to read it the
+/// same way, which is the whole reason it is one function.
+///
+/// Two callers, and #1077 is both of them disagreeing in turn. The value path
+/// ([`crate::ode::predictions::infusion_contributes`]) once had no compartment test at
+/// all and delivered such an infusion while the analytic sensitivity walk
+/// (`sens/ode_provider.rs`) dropped it — a *value* divergence, closed by #1196 step 3
+/// giving the value path this test. What that left was the walk disagreeing with
+/// **itself**: four rate-*on* sites spelled `cmt_raw() >= 1` and the rate-*off* saltation
+/// at `K_INF_END` did not, so a lagged `CMT=0` infusion delivered nothing (`f ≡ 0`, both
+/// engines) yet moved `∂f/∂η_LAG` by `+1.8878965164` at the first observation past the
+/// window end, where central FD of the predictor is exactly `0.0`.
+///
+/// Deliberately **not** folded into [`is_real_infusion`]: that predicate answers "are
+/// this row's `rate`/`duration` usable", the break-time builders ask it about every
+/// infusion including a `CMT=0` one (they must still break at its window end for the
+/// segment structure to line up), and only the *forcing* sites ask this.
+#[inline]
+pub(crate) fn infusion_has_rate_channel(d: &DoseEvent) -> bool {
+    d.cmt_raw() != 0
+}
+
 // ---------------------------------------------------------------------------
 // Where a lagged steady-state dose loads, and what the previous cycle leaves
 // running across the dose record (#1121).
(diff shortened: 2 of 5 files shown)


hiddevandebeek:
Reviewed the forcing and saltation paths, including the general two-sided boundary path. No blocking findings: the shared CMT=0 rate-channel predicate now matches the production resolver and prevents the K_INF_END saltation from differentiating a rate that was never applied.

I also ran cargo test --lib locally on 38df99db; it passed. The targeted CMT=0 and reset regression tests passed as well. The remaining coverage jobs are still pending.

hiddevandebeek:
Second-pass review of 38df99db: no blocking findings.

I traced the shared predicate through all six relevant paths: the production resolver; static and event-driven segment forcing; boundary_velocity; the rate-on saltation; and the K_INF_END rate-off saltation. The invariant now holds throughout: CMT=0 contributes no rate and therefore has no moving rate boundary, while the timeline still retains the window-end break. Out-of-range compartments remain independently bounded before indexing.

The regression coverage reaches the fast and general rate-off branches, the static walk, and the cross-dose Hessian contamination case. That last differential test is especially useful because the bad forcing cancels at first order and is observable only in the second-order blocks. The separate reset-before-arrival pin also matches the current production finite-difference result.

Verification: git diff --check and cargo fmt --check pass; GitHub reports 11/11 checks green. The earlier full cargo test result on this same SHA remains applicable. No new concerns.

hiddevandebeek:
Resolved the changelog feedback in 2ff4cc1e. The #1077 entry is now concise, user-facing, and accurately scoped to unvalidated hand-built model specifications.

PR #1321, opened 2026-09-10
perf(estimation): average eligible SAEM residual statistics

hiddevandebeek:
## Why

SAEM currently re-estimates residual SD from the latest retained ETA draw, which leaves a Monte Carlo noise floor even after convergence. The controlled saemix 3.5 investigation shows that averaging the residual sufficient statistic materially reduces this noise.

## What changed

For an intentionally narrow class of models, SAEM now averages the additive RSS or proportional scaled RSS with the SA schedule and sets the free scalar residual SD from its exact complete-data M-step. The implementation only enables this for single-population, uncensored, uncorrelated, simple Gaussian residual models with one free residual SD and only fixed or log-mu-referenced structural thetas. All other models retain the established NLopt M-step.

## Alternatives considered

Porting saemix residual updates broadly was rejected: combined, correlated, transformed, IOV, IIV-on-RUV, FREM, mixture, custom-magnitude, and free numerical-theta models do not share this scalar sufficient statistic. A general numerical M-step remains the fallback.

## Cross-repo dependency

|---|---|---|---|

## Breaking changes

- [x] None

## Numerical validation

- [ ] Compared against: NONMEM / prior ferx commit
- [x] Controlled update-level evidence: the included unit fixture matches the saemix 3.5 M-step (RSS 2 -> 18 at gamma 0.5 gives sigma sqrt(5), not latest-draw sigma 3).
- [ ] Full PK and NONMEM comparison remains required before review-ready status.

## Performance

- [x] No significant impact expected; eligible models also skip the now-unnecessary numerical sigma solve.

## Tests

- [x] Unit tests added in src/estimation/saem.rs.
- [x] Regression fixture added for the averaged RSS update.

## Example

- [x] Not applicable (internal estimator change with no new API surface).

## Docs

- [x] docs/estimation/saem.qmd updated.
- [ ] CHANGELOG.md entry will reference this PR number in the next commit.

## Checklist

- [x] cargo fmt --check green.
- [ ] tools/preflight.sh green.
- [x] No public API change.

## Reviewer hints

Review the eligibility gate first: it must stay conservative. The key behavior is the ordered, thread-count-stable RSS reduction and use of gamma rather than the latest draw.

## Open questions

Which existing PK fixture should be used as the formal NONMEM anchor for the initial additive/proportional rollout?

### DIFF
--- a/src/estimation/saem.rs
+++ b/src/estimation/saem.rs
@@ -325,6 +325,11 @@ struct SaemState {
     /// SA sufficient statistic for Omega_iov: running average of (1/N_occ) Σᵢ Σₖ κᵢₖκᵢₖᵀ.
     /// Zero-sized when `n_kappa == 0`.
     s2_iov: DMatrix<f64>,
+    /// SA sufficient statistic for an eligible scalar residual variance: the
+    /// running residual sum of squares, with proportional residuals divided by
+    /// the squared individual prediction. `None` for the general numerical
+    /// residual M-step.
+    residual_sse: Option<f64>,
     /// Current theta
     theta: Vec<f64>,
     /// Current omega matrix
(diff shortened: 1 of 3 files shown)


src/estimation/saem.rs
```diff
@@ -335,6 +340,149 @@ struct SaemState {
...
+            let mut sse = 0.0;
+            for (&y, &f) in subject.observations.iter().zip(predictions.iter()) {
+                if !(y.is_finite() && f.is_finite()) {
+                    return None;
+                }
+                let residual = y - f;
+                let term = match residual_model {
+                    ScalarResidualModel::Additive => residual * residual,
+                    ScalarResidualModel::Proportional => {
+                        if f.abs() <= f64::MIN_POSITIVE {
```
hiddevandebeek, line 452: [P1] Do not leave the scalar path selected when a proportional prediction reaches the residual-variance floor. `residual_variance` clamps `(f * sigma)^2` to `MIN_VARIANCE`, so for `f == 0` (and the wider small-prediction range) that row is independent of sigma and is not represented by `r² / f²`. Here the function returns `None`, but the M-step then only skips the update; it never takes the documented general-NLopt fallback. Consequently, an otherwise eligible model with an undosed/baseline zero prediction will retain its initial sigma throughout SAEM. Either run the numerical M-step when this occurs, exclude such fits from this scalar path, or omit floored rows from both the RSS and effective observation count. Please add a regression with a zero-prediction proportional row.

src/estimation/saem.rs
```diff
@@ -3050,29 +3219,60 @@ pub fn run_saem(
...
-                );
-                damp_mstep(&mut log_theta, &theta_new, gamma_mstep);
-                damp_mstep(&mut log_sigma, &sigma_new, gamma_mstep);
+                    residual_model,
+                ) {
+                    if n_obs > 0 {
+                        update_scalar_residual_sse(&mut state.residual_sse, sample_sse, gamma);
+                        if let Some(sse) = state.residual_sse {
+                            let sigma = (sse / n_obs as f64).sqrt();
+                            if sigma.is_finite() && sigma > 0.0 {
```
hiddevandebeek: [P2] Clamp a zero residual statistic to the sigma lower bound. For an eligible model whose retained sample fits every observation exactly (for example, a deterministic/no-ETA model or a zero-noise fixture), sample_sse and state.residual_sse are exactly 0. sqrt(0) is finite, but this guard rejects it, so log_sigma is never updated and SAEM keeps the initial residual SD even though the constrained scalar M-step is the configured lower bound. Zero RSS is a valid boundary optimum, not a prediction failure; please map it to log_sigma_lower[0] (or otherwise handle log(0) before clamping) and add an exact-fit regression.

hiddevandebeek: Implemented in cc15f02. Zero scalar residual variance now flows through the existing log-space clamp to log_sigma_lower, and scalar_residual_exact_fit_updates_sigma_to_lower_bound covers the exact-fit case. Targeted regression test passes.

Issue #1266, opened 2026-09-06
chore(ode): assert the extended-parameter slice length where the RHS reads it


TeunP:
The compiled `[odes]` RHS reads `TAFD` / `TAD` from `params[MAX_PK_PARAMS]` / `[+1]` as
`.get(..).copied().filter(|v| v.is_finite()).map_or(f64::NAN, ..)`
(`src/parser/model_parser.rs:11851`). An engine that hands it a bare `PkParams::values` —
`[f64; MAX_PK_PARAMS]`, two slots short — therefore gets `NaN` rather than a compile error, a
panic or a warning. The `NaN` multiplies into the state and the Jacobian and is laundered into
a plausible number downstream.

That amplifier is the root cause of **three separate bugs**:

| | engine | status |
|---|---|---|
| #1131 | EKF (`ode/ekf.rs`) | fixed in #1263 |
| #1261 | survival hazard readout (`stats/likelihood.rs:360`, `:4131`; `survival/mod.rs:738`) | open |
| #1264 | SS equilibration (`ode/predictions.rs`, twelve sites) | open, fixed by the T2 PR |

One line at the injection point detects the whole class:

```rust
debug_assert!(
    params.len() >= crate::types::MAX_PK_PARAMS + 2,
    "compiled [odes] RHS handed {} params, needs {}",
    params.len(), crate::types::MAX_PK_PARAMS + 2
);
```

Measured by the #1263 reviewer on PR commit `24d011d1`: `cargo test --lib` goes from 4083
passing to **35 failing**, 34 of them reporting `handed 128 params, needs 130` and every one an
SS route. Those 34 pass today only because no fixture reads a builtin. The 35th
(`parser::model_parser::tests::test_ode_block_supports_if_statements`) reports `handed 2 params`
and is a parser unit test calling the RHS directly — a test artifact, not production.

**Ordering constraint — this cannot land before #1264 is fixed.** The 34 failures *are* #1264.
File this to land immediately after the T2 PR, at which point the assert should be green and
stays as the guard against a fourth instance.

A stronger version worth deciding once rather than a fourth time: a newtype
(`ExtParams(&[f64])`) on `solve_ode`'s parameter argument makes passing a bare
`PkParams::values` unrepresentable rather than merely detected in debug builds.

Found by the code review of #1263; out of scope for that PR.

hiddevandebeek:
[...]

PR #1323, opened 2026-09-10
fix(survival): pass dose-time anchors to TTE hazard readouts [closes #1261]

### LINKED ISSUE #1261: fix(survival): a hazard that reads `TAD` / `TAFD` gets NaN — the bare parameter slice reaches the RHS on the hazard readout

hiddevandebeek:
## Why

ODE-accumulated TTE hazard readout paths called the compiled RHS with a bare `PkParams::values` slice. A RHS that read `TAFD` or `TAD` therefore received `NaN`; the scorer then converted the invalid result into its finite `1e20` rejection sentinel.

Closes #1261 and #1266.

## What changed

- Added a debug/test assertion at the single RHS injection point requiring the two extended dose-time slots.
- Added one private, survival-gated helper that derives TAFD/TAD from the subject regimen at a RHS readout time.
- Routed both TTE likelihood readout paths and public survival readout through extended parameters.
- Added a two-dose regression with exact `H(30) = 3.234`, `h(30) = 0.118`, and the corresponding event NLL.

## Alternatives considered

A newtype for extended RHS parameters would make bare slices unrepresentable, but would widen the change across every `solve_ode` caller. This PR keeps the immediate debug detector and narrowly fixes the three live callers; the stronger type-level redesign should be separate.

## Cross-repo dependency

## Breaking changes

- [x] None

## Numerical validation

- [ ] Warfarin example converges estimates within tolerance prior run
- [x] Reference values: hand-computed two-dose TAD hazard: `H(30) = 3.234`, `h(30) = 0.118`, `NLL = 3.234 - ln(0.118)`. The fixture exercises both the likelihood and public survival readout callers.

## Performance

## Tests

- [x] Unit test(s) added in `stats::likelihood`.
- [x] Regression test added; mutation check confirmed it fails with the old bare slice even if the new debug assertion is removed.

- [x] Not applicable (bug fix with no new API surface)

## Docs

- [x] No new user-facing syntax or documentation page is required.

## Checklist

- [x] `tools/preflight.sh` green before rebasing; the rebase only reconciled the current changelog section.
- [x] Functions on the `Dual2` sensitivity path stay differentiable (not touched).
- [x] No public API change.

## Docs & examples

- [x] Not applicable: no DSL, example, data, or estimator change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ section of the SDLC for the versioning policy).
 
 - **SAEM now averages the residual sufficient statistic for eligible single additive and proportional error models, reducing final-draw Monte Carlo noise in the residual SD estimate (#1321).**
 
+### Fixed
+
+- Fixed ODE-accumulated survival hazards that read `TAD`, which could reject valid multi-dose subjects with a misleading finite objective (#1261).
+
 ### Performance
 
 - **`focei, n_agq > 1` (the Gauss-Newton-anchored FOCEI quadrature refinement) now assembles
--- a/src/ode/predictions.rs
+++ b/src/ode/predictions.rs
@@ -2598,6 +2598,27 @@ fn subject_dose_attrs(
     (dose_lagtimes, dose_f_bio)
 }
 
+/// Build the extended parameter slice required for a standalone RHS read at `t`.
+///
+/// Integrating prediction paths keep this slice live and update its TAD anchor at
+/// each segment boundary. Readout-only callers (the TTE hazard derivative) do not
+/// own that walk, so they must derive both anchors from the same dose-time helpers
+/// instead of passing a bare [`PkParams::values`](crate::types::PkParams::values)
+/// array to the RHS (#1261).
+#[cfg(feature = "survival")]
+#[inline]
+pub(crate) fn rhs_ext_params_at(
+    ode: &OdeSpec,
+    subject: &Subject,
+    pk_params_flat: &[f64],
+    t: f64,
+) -> [f64; crate::types::MAX_PK_PARAMS + 2] {
+    let (dose_lagtimes, _) = subject_dose_attrs(subject, ode, pk_params_flat);
+    let mut ext_params = seed_ext_params(pk_params_flat, earliest_dose_time(&subject.doses));
+    ext_params[crate::types::MAX_PK_PARAMS + 1] = tad_anchor_for(&subject.doses, &dose_lagtimes, t);
+    ext_params
+}
+
 /// Earliest dose record time, or `+∞` when there are no doses.
 ///
 /// Takes the dose list rather than the `Subject` so the EKF
--- a/src/ode/predictions.rs
+++ b/src/ode/predictions.rs
@@ -2598,7 +2598,28 @@ fn subject_dose_attrs(
     (dose_lagtimes, dose_f_bio)
 }
 
-/// Build the extended parameter slice required for a standalone RHS read at `t`.
+/// Per-dose lagtimes only — the half of [`subject_dose_attrs`] that
+/// [`rhs_ext_params_at`] needs. Callers that read `H`/`h` at many `times` for the same
+/// subject (the TTE hazard readouts, #1261) compute this once and reuse it, rather than
+/// re-walking the dose list — and recomputing the unused `F`/bioavailability half — on
+/// every timepoint.
+#[cfg(feature = "survival")]
+#[inline]
+pub(crate) fn dose_lagtimes_for(
+    subject: &Subject,
+    ode: &OdeSpec,
+    pk_params_flat: &[f64],
+) -> Vec<f64> {
+    subject
+        .doses
+        .iter()
+        .map(|d| ode.dose_attr_map.lagtime(d.cmt_raw(), pk_params_flat))
+        .collect()
+}
+
+/// Build the extended parameter slice required for a standalone RHS read at `t`, from
+/// a dose-lagtime vector and first-dose time the caller has already computed (via
+/// [`dose_lagtimes_for`] / [`earliest_dose_time`]) — once per subject, not once per `t`.
 ///
 /// Integrating prediction paths keep this slice live and update its TAD anchor at
 /// each segment boundary. Readout-only callers (the TTE hazard derivative) do not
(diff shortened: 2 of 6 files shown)


hiddevandebeek:
Tracking: this PR is part of bug epic #1217.

hiddevandebeek:
## Code review (automated, recall-biased pass)

### 1. The new regression test never exercises the production hot path it's meant to guard (`src/stats/likelihood.rs`)

The fix touches two live TTE hazard readout sites:
- `ode_cumhaz_hazard` (`src/survival/mod.rs:738`) — the two-solve path.
- `tte_ode_nll_from_shared` (`src/stats/likelihood.rs:368`) — the #570 **shared** joint-PK-TTE solve, reading `share.rhs_params[i]` instead of `share.pk_values`.

The new test `tte_hazard_readout_uses_the_current_tad_anchor` (`src/stats/likelihood.rs:4176`) calls `tte_endpoint_nll` directly, which for `HazardSpec::OdeAccumulated` unconditionally calls `tte_ode_nll` → `ode_cumhaz_hazard` (`src/stats/likelihood.rs:399-401`). It then separately calls `crate::survival::ode_cumhaz_hazard` again. **Neither call goes through `try_joint_pktte_shared_solve` / `tte_ode_nll_from_shared`.**

But the actual `fit()` path — `individual_nll_into_with_schedule` (`src/stats/likelihood.rs:527`, the FOCE/FOCEI inner-loop objective called on every EBE iteration) — *does* build `joint_share = try_joint_pktte_shared_solve(...)` and route through `tte_ode_nll_from_shared` whenever the model qualifies. The test's model (a plain no-TV, no-reset, non-SDE, non-FREM ODE with an `OdeAccumulated` hazard reading `TAD`) satisfies every eligibility check in `try_joint_pktte_shared_solve` (`model_uses_time_anywhere` excludes the injected `d/dt(__chz_*)` hazard lines from its `pk_reads_model_time` check by design — see the doc comment above that guard — specifically so a TAD/TIME-reading hazard line doesn't disqualify the model). So in a real `fit()` run, this exact model shape *is* routed through the shared solve.

Net effect: if `share.rhs_params[i]` at `src/stats/likelihood.rs:368` were reverted to `share.pk_values` (the original bug), or if the `rhs_params` computation at `src/stats/likelihood.rs:323-326` were subtly wrong, no test in this PR (or, as far as I could find, anywhere else in the repo — this is the only `hazard = ...TAD...` fixture that exists) would fail. This is exactly the "ask which engine each fixture actually ran on" / "mutate each side of a twin separately" pattern this file's own header documentation (`CLAUDE.md`) calls out repeatedly as the recurring class of bug in this subsystem (#1210, #1223). Recommend adding a variant of the new test that goes through `individual_nll_into_with_schedule` (or at least `accumulate_non_gaussian_nll` with `joint_share = try_joint_pktte_shared_solve(...)`) on the same TAD-reading fixture, and confirming it diverges from `share.pk_values` under mutation.

### 2. `rhs_ext_params_at` redoes O(n_doses) work per readout time it used to do once (`src/ode/predictions.rs:2610`)

`rhs_ext_params_at` calls `subject_dose_attrs(subject, ode, pk_params_flat)` on every invocation, which maps over *all* of `subject.doses` to compute both `dose_lagtimes` and `dose_f_bio` — then discards `dose_f_bio` entirely (`let (dose_lagtimes, _) = subject_dose_attrs(...)`). It is called once per element of `times` in both:
- `try_joint_pktte_shared_solve` (`src/stats/likelihood.rs:323-326`, `.map` over `times`)
- `ode_cumhaz_hazard` (`src/survival/mod.rs:738`, inside the `for (i, &t) in times.iter().enumerate()` loop)

Previously these call sites passed the already-available `pk.values` directly — O(1) per readout. Now each readout time re-walks every dose to rebuild `dose_lagtimes` (and the unused `dose_f_bio`), turning it into O(n_times × n_doses) inside the FOCE/FOCEI inner-loop objective (called every EBE iteration, every subject). For a subject with a long dosing history (dense TDM / infusion regimens) and a joint PK-TTE endpoint with several event/censor times, this is repeated, wasted allocation and work in a hot path. Since `dose_lagtimes` doesn't depend on `t`, it could be computed once outside the per-time loop/map and passed into a `tad`-only variant of `rhs_ext_params_at` (or `tad_anchor_for` called directly with a pre-computed `dose_lagtimes`), and the unused `dose_f_bio` computation dropped altogether.

---

hiddevandebeek:
Pushed f5ecbe18 addressing both findings from the code review above:

1. **Test coverage gap (high)** — added `hot_path_individual_nll_reads_the_shared_solve_tad_anchor`, which calls `individual_nll_into_with_schedule` (the actual FOCE/FOCEI inner-loop entry point) directly, rather than only `tte_endpoint_nll`/`ode_cumhaz_hazard`. It isolates the TTE contribution by diffing against a PK-only sibling model/subject (identical `[odes]`, doses, and PK observation, but no `[event_model]`/TTE record), and asserts the fixture actually reaches `try_joint_pktte_shared_solve` before checking the result — so a revert of the `share.rhs_params[i]` fix in `tte_ode_nll_from_shared` would now fail this test.

2. **Efficiency regression (moderate)** — split `dose_lagtimes_for` out of `rhs_ext_params_at` so both call sites (`try_joint_pktte_shared_solve`, `ode_cumhaz_hazard`) compute the per-subject dose-lagtime vector and first-dose time once, before looping over `times`, instead of re-walking `subject.doses` (and recomputing the unused bioavailability half) on every timepoint.

`cargo check --features survival --lib --tests` and `cargo fmt --check` are clean. This box is memory-constrained (7 GB, currently under load from another concurrent build), so I wasn't able to run the full local test binary — leaving that to CI.

PR #1330, opened 2026-09-10
perf(tools,api): bound fit admission to the requested width; parallel post-fit pass [closes #1329]

### LINKED ISSUE #1329: perf(tools,api): a threads budget is not a ceiling on concurrent fits, and the post-fit per-subject pass is serial
Two concurrency-structure defects found by the parallel-computation review
(`plans/parallel-computation-review-2026-09-10.md`) and its implementation design
(`plans/parallel-optimization-implementation-2026-09-10.md`). Neither has a numerical
component, which is why they are one issue and separate from the numerical items in that
plan (the Laplace ETA-only provider, weighted moments, FOCE/FOCEI preparation, AGQ
continuation, SAEM/Bayes).

## 1. A `threads` budget is not a ceiling on concurrent fits

Every many-fit tool — `bootstrap`, and every search tool via `search::Runner`
(`modelsearch`, `covsearch`, `iivsearch`, `ruvsearch`, `iovsearch`, `globalsearch`) — runs
`N` fits over a `width`-wide outer pool, spelled

```rust
plan.install(|| jobs.par_iter().filter_map(run_one).collect())
```

Each job calls `ferx_core::fit`, which `install`s **its own inner pool**. A Rayon worker
blocked on a nested `install` keeps stealing work from its own pool — this is
[documented `ThreadPool::install` behaviour](https://docs.rs/rayon/1.12.0/rayon/struct.ThreadPool.html#method.install),
not a bug. With `N ≫ width` there are always pending jobs to steal, so every blocked worker
picks up another fit, whose `install` blocks it again, and so on.

The review's synthetic probe measured peaks of **13 / 24 / 29 live jobs for requested widths
1 / 2 / 4**.

The cost is **memory and latency stability, not throughput**: `width` was what sized fit
stacks (`FIT_RAYON_STACK_SIZE`, 32 MiB each) and live populations, and the real number was
unbounded. A batch that was already inside its budget is unaffected. Both
`docs/tools/search.qmd` and `docs/tools/bootstrap.qmd` already *promise* the bound
("each fit uses one thread", "so the machine is not oversubscribed"); it simply was not
enforced.

A blocking permit *inside* the parallel iterator is **not** the fix: the worker that blocks
on the permit is the same worker the run needs in order to make progress, so the permit can
occupy the pool it was meant to protect.

Fix shape (from the plan): submit exactly `R` long-lived lane closures, each synchronously
claiming one indexed job from an atomic counter, completing it, recording its indexed
result, and only then claiming another. Lanes must be **plain OS threads, not Rayon tasks** —
`R` tasks on an `R`-wide pool are themselves stealable, so that spelling holds no bound
either. Preserve: search group barriers, journal and error handling, bootstrap replicate seed
mapping, cancellation checks between jobs, panic propagation, and final original-order
assembly.

Secondary: the bootstrap's hand-built `rayon::ThreadPoolBuilder` never got the ferx worker
stack, so it inherited Rayon's 2 MiB default — the exact footgun `PoolPlan::install` exists
to prevent.

## 2. The post-fit per-subject pass is serial

`api::postfit::compute_subject_results` (one prediction sweep per subject: IPRED, PRED,
IWRES, CWRES, per-subject OFV) and `api::postfit::sweep_sensitivity_solver_stats` (one
analytic sensitivity solve per subject) both run one subject at a time, at the final
estimates, on a pass whose subjects are completely independent. `fit()`'s likelihood loop
already `par_iter`s over the same subjects.

The trap, and the reason this is not a one-line `.iter()` → `.par_iter()` swap:
`ode::solver::SolverStatsScope` is **thread-local**. `fit.rs` opens one scope around
`compute_subject_results` and a second around the sensitivity sweep. Fanning out while
leaving the scopes at the call site would report zero `min_dt` clamps, zero `auto`
escalations, and a permanently zero `auto_stiff_rejected_jets` — every counter reading clean
for exactly the integrations it exists to surface, i.e. reintroducing the dead diagnostic
#1080 removed. The scope has to move inside each subject task and the counters be merged in
subject order.

Acceptance: output bit-identical and counters equal **and non-zero** across worker counts.
Equality alone is satisfied by zero at both widths, which is the bug.

## Out of scope

The numerical items of the same plan, each of which needs its own PR and its own oracle:

- the Laplace exact-anchor **ETA-only** tensor provider (`1 + 2(t+d)` → `1 + 2d` second-order
  evaluations). Blocked on a settled acceptance oracle: the plan measured the reconverged
  objective-gradient check disagreeing by ~42% (FD grid) / ~2.1% (analytic) at the fixture's
  **default** ODE tolerances, collapsing to ~0.065% on both routes only at
  `ode_reltol = 1e-10`, `ode_abstol = 1e-12`. Needs a step-size sweep first.
- one-node Laplace specialization and the precomputed weighted moments
  (`O(p·Q·d²)` → `O(Q·d² + p·d²)`);
- evaluation-scoped FOCE/FOCEI preparation reuse across the value/gradient boundary;
- AGQ continuation fusion, then node/coordinate scheduling only if profiling justifies it;
- SAEM IOV phase parallelism and independent Bayesian chains.

hiddevandebeek:
## Why

Two findings from the parallel-computation review (`plans/parallel-computation-review-2026-09-10.md`)
and its implementation design (`plans/parallel-optimization-implementation-2026-09-10.md`),
both of which are concurrency-structure defects with **no numerical component**:

1. **A `threads` budget was never a ceiling on concurrent fits.** Every many-fit tool
   (bootstrap, modelsearch, covsearch, iivsearch, ruvsearch, globalsearch) runs `N` fits
   over a `width`-wide outer pool as `plan.install(|| jobs.par_iter()…)`. Each job calls
   `fit()`, which `install`s its own inner pool — and a Rayon worker blocked on a nested
   `install` **keeps stealing work from its own pool** (documented behaviour, not a bug).
   With `N ≫ width` there are always pending jobs to steal, so every blocked worker picks
   up another fit, whose `install` blocks it again. The review's synthetic probe peaked at
   **13 / 24 / 29 live jobs for requested widths 1 / 2 / 4**. The cost is memory: `width`
   was sizing fit stacks, and the real number was unbounded.

2. **The post-fit per-subject pass was serial** — one prediction sweep per subject plus one
   analytic-sensitivity solve per subject, at the final estimates, on one thread, on a pass
   whose subjects are completely independent.

This is delivery steps 1 and 4 of the implementation plan. Steps 2, 3, 5 and 6 are
deliberately **not** here: they change numerics, and step 2's acceptance oracle is the one
the plan itself disowns (a reconverged-objective FD check that disagreed by ~42% at the
fixture's default ODE tolerances, collapsing to ~0.065% only at `reltol 1e-10 / abstol
1e-12`). Those belong in their own PRs behind a settled oracle.

## What changed

**`crates/ferx-tools/src/lanes.rs` (new).** `run_in_lanes(width, n_jobs, job)`: exactly
`width` long-lived lane closures, each claiming one job index from an atomic counter,
running it to completion, recording its indexed result, and only then claiming another.

- The lanes are **plain OS threads, not Rayon tasks**. That is load-bearing: a plain thread
  blocked on the inner pool's `install` waits on a latch and steals nothing. Spawning
  `width` tasks on a `width`-wide Rayon pool would *not* bound anything — the lane tasks are
  themselves stealable, so a lane blocked before its siblings were picked up could run two
  of them nested.
- A blocking permit *inside* the parallel iterator is the wrong fix for the same reason the
  plan gives: the worker that blocks on the permit is the worker the run needs to progress.
- Lanes carry `FIT_RAYON_STACK_SIZE`, which also fixes the bootstrap's hand-built
  `ThreadPoolBuilder` silently inheriting Rayon's 2 MiB worker stack.
- Results come back **sorted by job index** and `None` is dropped, i.e. exactly what
  `par_iter().filter_map().collect()` returned, so no caller's journalling order, seed
  mapping or final assembly changes. A job panic is `resume_unwind`-ed, so it still takes
  the run with it.

**Wired into all three sites that nest a `fit()`**: `search/runner.rs` (every search tool
goes through `Runner`), the bootstrap replicate loop, and the bootstrap Δofv sweep.
`PoolPlan` still decides the *width*; only the admission changes. `BootstrapOptions::threads
= None` still means the ambient Rayon pool width, now resolved explicitly
(`replicate_lane_width`) so that it can be bounded at all.

**Post-fit parallelism.** `compute_subject_results` and `sweep_sensitivity_solver_stats` now
`par_iter` over subjects on the pool the fit already installed. The part that could not be a
bare `.par_iter()` swap is `SolverStatsScope`, which is **thread-local**: a scope the caller
opens sees the caller's thread only, so fanning out while leaving the scope at the call site
would have reported zero `min_dt` clamps, zero `auto` escalations and a permanently zero
`auto_stiff_rejected_jets` — the counters reading clean for exactly the integrations they
exist to surface, which is the dead diagnostic #1080 existed to remove. Each subject task
therefore opens its own scope and the counters are **merged in subject order** and returned
as a value. `compute_subject_results` takes a `collect_solver_stats: bool` — what the old
`integrates_odes(model).then(…)` decided.

## Alternatives considered

- **A semaphore permit inside the existing `par_iter`.** Rejected above: it can occupy the
  workers it was meant to protect.
- **An outer Rayon pool of width `R` with `R` spawned lane tasks.** Rejected: the lane tasks
  are stealable, so the bound is not actually held. This is why the lanes are OS threads.
- **Leaving the post-fit scope at the call site and collecting into a `Mutex`.** Rejected:
  it would keep a second way to get the thread-local wrong. Returning the value makes the
  merge a visible, ordered step.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` passes against the committed
baseline — everything added is `pub(crate)` or private).

## Breaking changes

- [x] None

`FitResult` is unchanged, including bit-for-bit `SubjectResult` contents and the
`ode_solver` warning's counter payload.

## Numerical validation

- [x] Not applicable — no numerical path is touched. Neither change alters an estimate, an
  objective, a residual or a counter. The post-fit pass has **no reduction** across subjects
  (each subject is an independent prediction sweep), so nothing is reassociated and no output
  moves by a ULP; this is asserted as bit-identity rather than a tolerance, at one worker and
  at eight.

## Performance

- [x] Faster — but the honest framing is the plan's: the admission fix is a **memory and
  latency-stability** change, not a throughput one (an already well-behaved CPU-bound batch
  may show nothing). The post-fit fan-out is bounded by that pass's share of a fit, which is
  small on closed-form models and largest on ODE models with many subjects. **No benchmark
  numbers are claimed here** — this box could not run a release build (see Tests), and the
  plan is explicit that provider-counter and wall-clock comparisons must be separate,
  uninstrumented, warmed, order-alternated runs before any figure is quoted.

## Tests

- [x] Unit tests added — `crates/ferx-tools/src/lanes_tests.rs`, 7 tests, **601 passed** in
      `cargo test -p ferx-tools --lib`.
- [x] Regression test added that fails without the fix.

The lane bound is pinned three ways, because "the peak never exceeded the bound" is also
what a run with no concurrency at all reports:

1. `at_most_width_jobs_are_live_at_once` asserts `peak == width` (not `<=`) at widths
   1 / 2 / 4, through a probe that nests a pool the way `fit()` does.
2. `the_shape_this_replaces_admits_more_jobs_than_its_width` drives the **old** spelling
   through the same probe and asserts it exceeds its width — so if the premise of the lane
   runner were ever wrong, that is a red test rather than a quietly vacuous green one.
3. **Mutation-tested in both directions**, which is what found that (1) needed `==`:
   - `width → 2 * width` → red at width 1 (`2 jobs were live at once, not 1`);
   - `width → 1` (lanes go serial) → red at width 2 (`1 jobs were live at once, not 2`).

For the post-fit half, `the_parallel_post_fit_pass_is_worker_count_independent` compares a
fit at 1 worker against the same fit at 8 on an 8-subject stiff ODE fixture, asserting both
halves, because neither alone can observe the defect:

- the `ode_solver` counter payload must be **equal** across widths — but equality alone is
  satisfied by counters that are *zero at both*, which is precisely the thread-local bug, so
- `accepted_steps` and `auto_stiff_segments` are asserted **non-zero** first, and the test
  says so;
- plus bit-identical IPRED / PRED / IWRES / CWRES / per-subject OFV and preserved subject
  order.

The pre-existing `abandoned_non_finite_timeline == 3` assertion is a second, independent
backstop on the same merge: it measures counters through the fit's own production path and
would drop below 3 if any per-subject scope were lost.

### What is not verified locally, and why

**This box cannot build the `ferx-core` lib *test* binary** — 7 GB RAM, SIGKILL, and it still
dies at `-j 1` with `DEBUG=0`, `INCREMENTAL=false` and `CODEGEN_UNITS=256`. So for
`ferx-core`:

- `cargo check --lib --tests --features ci` is clean, and `tools/preflight.sh` is green;
- the three `ferx-core` test files touched here are **compile-verified but not executed**.

**CI must confirm `Tests + coverage (core)` before this merges.** The `ferx-tools` half was
run in full.

## Example

- [x] Not applicable — internal; no new API surface, no DSL or output change.

## Docs

- [x] `docs/` updated — `docs/tools/search.qmd` and `docs/tools/bootstrap.qmd` now say the
      thread budget is an exact ceiling on live fits (both pages already *promised* this; it
      simply was not true).
- [x] `CHANGELOG.md` `[Unreleased]` entry added — one `Fixed` (the admission ceiling, in
      user-facing terms) and one `Performance` (the post-fit fan-out).
- [x] `cargo run -p docs-lint -- --check` clean (5 pre-existing R1 baseline excuses, none new).

## Checklist

      (`--all-targets`), rustdoc, docs, public-API baseline.
- [x] Nothing on the `Dual2` sensitivity path changed. `sweep_sensitivity_solver_stats` calls
      the same `subject_sensitivities` / `subject_eta_grad` providers with the same arguments;
      only the loop around them and where its counters land changed.
- [x] No version bump needed (not breaking).


hiddevandebeek:
Reviewed commit `531fb333e43459c0d2687b2b52f6e8425e7dc69c`.

**[P2] Preserve the Rayon default when bootstrap has no explicit thread budget** — [`crates/ferx-tools/src/bootstrap/mod.rs:1436–1440`](https://github.com/FeRx-NLME/ferx-core/blob/531fb333e43459c0d2687b2b52f6e8425e7dc69c/crates/ferx-tools/src/bootstrap/mod.rs#L1436-L1440)

`None => PoolPlan::from_budget(0, n_todo).replicates()` changes the source of the bootstrap budget. Previously `None` used the ambient Rayon pool, as `BootstrapOptions::threads` still documents. `PoolPlan::from_budget(0, ...)` instead uses `effective_default_threads()`, which normally derives available cores minus one, capped at eight; it does not honor `RAYON_NUM_THREADS` or a caller-created Rayon pool unless the separate ferx global-configuration flag is set. For example, on a machine with at least nine available CPUs, a library caller using `RAYON_NUM_THREADS=2` and `threads: None` now gets eight lanes for a sufficiently large bootstrap. Conversely, the documented all-core default becomes capped at eight on larger machines. The comment claiming these widths are the same is therefore incorrect.

Please resolve the `None` width from `rayon::current_num_threads()` before spawning the lanes, retaining the work-count clamp. Add a regression test that invokes the width resolver inside a custom Rayon pool whose size differs from the engine default; this avoids changing process-wide environment variables in tests.

The lane admission and post-fit changes otherwise look sound on inspection: results retain input order, mixture guards and solver-stat scopes are inside each subject task, and the fit pool already propagates ODE overrides to workers. I ran a standalone Rust probe against the actual lane helper: widths 1/2/4 attained their concurrency limits, filtered results remained ordered, and empty/zero-width cases passed. All 11 reported PR CI checks are green. I did not rerun the full Cargo suites locally.

hiddevandebeek:
Fixed the review finding in `482fed2b`.

Bootstrap now resolves an unset thread budget with `rayon::current_num_threads()` on the caller before spawning lanes, preserving custom Rayon pools and `RAYON_NUM_THREADS`. Explicit budgets remain supported, with the width clamped to the available work. The code comments, bootstrap documentation, changelog, and PR description now reflect that behavior.

Added a regression test using two distinct custom pool widths, plus explicit-budget and empty-queue cases. All **602 ferx-tools unit tests passed** with `--features ferx-core/ci` using reduced-memory build settings. Formatting also passed. The initial local builds hit SIGKILL and a full `/tmp`; rebuilding on the main filesystem with debug info and incremental compilation disabled resolved those limitations. Full preflight is running, and I explicitly dispatched CI for the new commit because no automatic Actions run appeared.

PR #1336, opened 2026-09-10
perf(estimation): analytic grid response by default for the exact (laplace) anchor [closes #1335]

### LINKED ISSUE #1335: perf(estimation): make the analytic grid response the default for the exact (laplace) anchor — the block-Ω case that was meant to decide it had never been run
`agq::use_analytic_grid_response` defaults the two Hessian anchors oppositely:

```rust
GridResponseOverride::Auto => matches!(anchor, HessianAnchor::GaussNewton),
```

so `method = focei, n_agq > 1` gets the analytic `dH̃/dx`, while `method = laplace`
(`HessianAnchor::Exact`) falls back to the finite-difference sweep — `2·n_free` perturbed
anchor rebuilds per subject per gradient — unless a user sets
`FERX_AGQ_GRID_RESPONSE=analytic`.

The exact anchor's analytic route (`estimation::laplace_h_deriv`) has existed and been correct
for a while. It was left opt-in on cost grounds, and those grounds no longer hold.

## Why it was opt-in, and why that reasoning was incomplete

The route pays the covariance provider's `1 + 2(n_theta + n_eta)`-evaluation third-order sweep
against the FD route's `2·n_free` anchor rebuilds. Each analytic evaluation does materially
more work per call, so fewer calls does not automatically mean less time — a real ambiguity,
and successive benchmark sessions on `warfarin_laplace.ferx` measured **opposite signs** (+34%
slower, then ~14% faster). Both runs were real; `plans/laplace-sensitivity-speed.qmd` correctly
concluded that the volatility, not a settled loss, was the blocker.

What nobody noticed is *why* that fixture is volatile: **it is diagonal-Ω, where the two costs
are equal by construction.** `n_free = n_theta + d = 7` and `1 + 2(n_theta + d) = 13` vs
`2·n_free = 14` — essentially a tie, so the measurement is pure noise around parity and was
never going to settle the question in either direction.

`agq.rs` even names the right test — *"stays opt-in until a densely-parameterised case
(block-Ω) shows a repeatable wall-clock win"* — and the fixture exists
(`plans/laplace-sensitivity-speed/warfarin_block_omega_laplace.ferx`). **It had never been
run on this anchor.** The block-Ω row recorded in the `.qmd` is the *Gauss-Newton* anchor,
a different route that was already analytic-by-default, which is how the gap went unnoticed.

## Measurement

5 interleaved reps, `ci-test` profile, `RAYON_NUM_THREADS=1`, commit `f2434247`. Exact anchor,
`fd` → `analytic`:

| Fixture | Provider calls | Provider time | Reps won | Outer iters | OFV |
|---|---|---|---|---|---|
| diagonal Ω, closed-form | 6840 → 6460 (−5.6%) | 0.0586s → 0.0556s (−5%) | 4/5 | 37 → 37 | identical |
| **block Ω (3×3)** | 11520 → 8160 (−29%) | 0.0880s → 0.0670s (**−24%**) | **5/5** | 47 → 47 | identical |
| ODE | 10270 → 6470 (−37%) | 1.581s → 1.029s (−35%) | 5/5 | 56 → 37 | identical |

Block-Ω is the deciding row: a wide call-count gap (20 FD rebuilds vs 13 provider evaluations),
won on 5 of 5 reps, at an **identical** outer-iteration count — so it is a like-for-like
comparison on one optimiser path rather than two trajectories.

Caveats, stated rather than buried:

- the **diagonal** row is still near parity and one rep of five favoured `fd`; nothing here
  claims that case improved;
- the **ODE** row is the largest win but is **not** like-for-like (56 → 37 outer iterations),
  so its call count is the number to quote, not its time;
- the binary carries the #1333 contraction, so the `analytic` arm benefits from it — not
  apples-to-apples against session 1, but it is the configuration the flip lands on;
- every arm of every fixture returned an **identical OFV**. This is the load-bearing
  observation: the two routes compute the same derivative and differ only in cost.

## What to change

1. `Auto` → analytic for both anchors. Drop the now-unused `anchor` parameter from
   `use_analytic_grid_response` rather than leaving one the body ignores.
2. **Keep `FERX_AGQ_GRID_RESPONSE`.** It is what `bench.sh` drives, it is the only way to
   re-measure a quantity this issue shows to be session- and fixture-sensitive, and the design
   note in the `.qmd` asks for exactly it ("keep an internal way to force the baseline and
   candidate paths in the benchmark harness; a public fit option is not required"). It is
   already internal — an env var, not a `[fit_options]` key — and costs one `OnceLock` read
   when unset.
3. The FD sweep stays regardless: it is the live fallback whenever `analytic_grid_response`
   declines (out of scope, e.g. M3 censoring; or an ill-conditioned anchor), pinned by
   `out_of_scope_models_decline_to_the_fd_grid_response`.
4. **Reconcile four artifacts that currently disagree**, which is half of why this was hard to
   adjudicate:
   - `agq.rs::use_analytic_grid_response` — reports the superseded +34%;
   - `laplace_h_deriv.rs` module doc — "does not establish a speedup, and is not the default";
   - `bench.sh` header — "both cases now favor analytic; keep re-measuring";
   - `warfarin_block_omega_laplace.ferx` header — asserts "measured: it does not [win]" about a
     case that had never been run on this anchor. That one is not merely stale, it is wrong.

## User-visible consequence

This is a `Changed`, not only a `Performance`, entry. On ODE `laplace` fits the optimiser path
moves (37 outer iterations instead of 56 on the benchmark fixture), so converged estimates may
shift within the convergence tolerance. Same OFV in every measured arm, so it is not an
accuracy regression — but it is visible.

hiddevandebeek:
## Why

The exact Hessian anchor used by `method = laplace` previously defaulted to finite-differencing the grid response, rebuilding the conditional Hessian at `x ± h` for each free population parameter. This enables its existing analytic route by default after the block-Ω benchmark addressed the performance question in #1335.

The diagonal-Ω fixture has three theta, three omega, and one sigma parameter: `n_free = n_theta + d + n_sigma = 7`. Its 14 FD rebuilds versus 13 provider evaluations are similar counts of different work units; they imply neither equal runtime nor general parity for diagonal Ω. Successive timing sessions on this fixture measured opposite signs. The block-Ω fixture has a wider gap: 20 FD rebuilds versus 13 provider evaluations.

## What changed

- `use_analytic_grid_response()` defaults both anchors to analytic and no longer takes an anchor argument. The Gauss-Newton default is unchanged.
- `FERX_AGQ_GRID_RESPONSE=fd` still forces the old sweep; `analytic` explicitly selects the new default. Unsupported models and ill-conditioned anchors still fall back to FD.
- Added a default-selection test. Updated the AGQ docs, changelog, benchmark notes and source comments, including the one-node Laplace optimization now reached by default.
- The documentation distinguishes rounded benchmark OFV agreement from derivative correctness and includes sigma in the diagonal-Ω cost formula.

## Alternatives considered

Keep `FERX_AGQ_GRID_RESPONSE` because the existing benchmark harness uses it to compare both routes in separate processes. The finite-difference implementation is also required as the live fallback. No new public fit option is introduced.

## Cross-repo dependency

| ferx (R) | — | not needed; no public API change | — |

Dependency #1334 has merged. This PR now targets `main`; merge conflicts in the changelog and AGQ comments were resolved while preserving the reviewed executable AGQ code.

## Breaking changes

- [x] None — no API, file-format or field change.

## Numerical validation

- [x] Existing `analytic_grid_response_matches_the_fd_route` directly compares both anchors at `n_agq = 1` and `3`, using a `1e-5` scaled-error tolerance and asserting that the analytic route engages.
- [x] Existing `laplace_h_deriv` tests compare the Hessian derivative at multiple parameter points, with block Ω, combined error and ODE models. The ODE test uses a `5e-3` tolerance.
- [x] The original benchmark reports matching OFVs at the console's **four-decimal precision** for all six fixtures. `bench.sh` parses that rounded output; this is not a full-precision comparison of OFVs or estimates, and does not establish gradient equivalence. The derivative tests above provide that evidence.
- [ ] No new NONMEM comparison: this switches between existing implementations of the same gradient term; it adds no estimator, error model or structural model. The latest follow-up changes only documentation and comments.

The ODE Laplace fixture takes 37 outer iterations with the analytic route versus 56 with FD. The optimizer trajectory changes, so converged estimates may shift within convergence tolerance. Rounded OFV agreement alone does not quantify those shifts.

## Performance

Original measurement: five interleaved reps, `ci-test`, `RAYON_NUM_THREADS=1`, commit `f2434247`, using `plans/laplace-sensitivity-speed/bench.sh`. Exact anchor, FD → analytic:

| Fixture | Provider calls | Provider time | Reps won | Outer iterations | OFV |
|---|---|---|---|---|---|
| Diagonal Ω, closed-form | 6840 → 6460 | 0.0586s → 0.0556s (−5%) | 4/5 | 37 → 37 | same to 4 decimals |
| Block Ω (3×3) | 11520 → 8160 | 0.0880s → 0.0670s (−24%) | 5/5 | 47 → 47 | same to 4 decimals |
| ODE | 10270 → 6470 | 1.581s → 1.029s (−35%) | 5/5 | 56 → 37 | same to 4 decimals |

- [x] The block-Ω row supports the default switch: lower provider time in all five reps with equal outer-iteration counts. Equal counts do not prove identical trajectories.
- [x] The diagonal fixture remains inconclusive across sessions. The ODE totals reflect different iteration counts and do not isolate per-iteration cost.
- [x] These are provider-time measurements, not end-to-end runtime. Call counts are deterministic for a given trajectory; provider time remains timing-sensitive.

The measured binary includes #1333's cheaper contraction, so this is not an apples-to-apples comparison against the earlier session. Benchmarks were not rerun for the documentation follow-up.

## Tests

- [x] `the_analytic_grid_response_is_the_default` added in the original change.
- [x] Original author reported `cargo test --lib`: 4359 passed, 0 failed, 23 ignored, including AGQ and VI callers of the exact anchor. This is the original validation, not a fresh run of the documentation follow-up.
- [x] No new tests needed for the follow-up: only prose and comments changed.
- [x] Follow-up validation: `cargo fmt --all -- --check`, `cargo run -p docs-lint -- --check` (five existing R1 baseline exceptions), `bash -n plans/laplace-sensitivity-speed/bench.sh`, and `git diff --check` passed.

Independent review attempted `cargo test --lib --profile ci-test ... grid_response`, but stopped after approximately 15 minutes still compiling; it produced no test results. Forced environment settings are exercised by the benchmark as separate processes, avoiding the shared `OnceLock` cache in unit tests.

## Example

- [x] Not applicable — no new DSL syntax or API surface. Existing Laplace benchmark fixtures exercise the route; `FERX_AGQ_GRID_RESPONSE=fd` restores the prior route.

## Docs

- [x] `docs/estimation/agq.qmd` updated.
- [x] Existing `[Unreleased]` changelog entry updated; behavioral consequences remain documented.
- [x] Benchmark notes and source comments use consistent cost formulas, OFV precision and default-route descriptions.
- [x] No new pages or generated site output.

## Checklist

- [x] Original author reported `tools/preflight.sh` green before the documentation follow-up.
- [x] No numerical kernel or public API changed by the follow-up; no version bump required.
- [x] No `Dual2` kernel touched.

## Docs & examples

- [x] ferx-site / ferx-book updates not applicable: no new DSL syntax, fit option, estimator, example or dataset.
- [x] No example execution needed for the documentation follow-up.

## Reviewer hints

Review the default-selection change and fallback wiring, then the performance evidence and its limits. The follow-up corrects the diagonal cost formula, qualifies OFV equality as four-decimal agreement, and removes stale opt-in descriptions. It does not alter the numerical implementation.

## Open questions

None.


hiddevandebeek:
Reviewed `dab78a523ab332d18244b32df2c3d41ec15f7014` against the stacked base. I found no blocking functional issue in the default switch: the explicit FD override remains wired up, and analytic declines still fall through to the existing FD implementation without partially updating the output. The existing parity test directly exercises both anchors at node counts 1 and 3; the Hessian-derivative tests additionally cover block omega, combined error, and ODE models.

Two non-blocking documentation corrections:

1. **[P3] Correct the diagonal-omega cost formula** — `docs/estimation/agq.qmd:385` and `plans/laplace-sensitivity-speed.qmd:286` say `n_free = n_theta + d` (the plan adds `= 7`). This fixture has three theta, three omega, and one sigma parameter, so `n_free = n_theta + d + n_sigma = 7`, whereas `n_theta + d = 6`. The comparison is **14 FD rebuilds versus 13 provider evaluations**, not equal costs by construction. Please describe this as similar call counts for this fixture; neither equal time nor general diagonal-omega parity follows from the formula.

2. **[P3] Qualify the claim of identical OFVs** — the new docs/changelog treat identical benchmark OFVs as establishing that the routes compute the same derivative and have no accuracy difference. `bench.sh` extracts the console OFV, which `src/io/output.rs:150` prints with only four decimal places. That supports agreement at displayed precision, not exact equality or gradient equivalence. Please say that explicitly and use the derivative-parity tests as the correctness evidence (or compare full-precision output with a stated tolerance).

Also worth updating the now-stale comment at `src/estimation/agq.rs:1688` describing the one-node optimization as reachable only through the opt-in exact-anchor route; this PR makes that a default path.

Validation limitation: I inspected the relevant tests but could not independently complete the targeted run. I stopped `cargo test --lib --profile ci-test --target-dir /root/package/ferx-core/target grid_response` after approximately 15 minutes still compiling; no test results were produced. I did not rerun the performance benchmark.

Issue #555, opened 2026-06-26
Regression (#534): ODE η+covariate obs_scale expression diverges from analytical — two_cpt_oral_cov OFV-at-init gap ~370–534


TeunP:
cc @hiddevandebeek (author of #534 — this is the `obs_scale` analytic-gradient path your PR introduced) and @roninsightrx. Flagging a regression #534 (`feat/486-analytic-expression-scale`) appears to have introduced; it's red on ferx-r `main` now.

## Summary

The ODE form of `two_cpt_oral_cov` no longer agrees with its analytical twin: **OFV-at-init diverges by ~370–534** (expected `< 1`). The discrepancy is in the **η-dependent `ExpressionScale` divisor** path (`[scaling] obs_scale = expr(θ, η, covariate)`) reworked to be analytic in #486/#534. The failing model is the only ODE/analytical equivalence pair whose `obs_scale = V1` carries **both a covariate (WT) and an eta (ETA_V1)** — the exact path ferx-core's own equivalence test does not cover.

It surfaced downstream: ferx-r `main` went red (both `test-coverage` and `R-CMD-check`) the moment its ferx-core pin was bumped to the #534 merge.

## Reproduction (current `main`)

Using ferx-r's two example models (`two_cpt_oral_cov.ferx` analytical, `two_cpt_oral_cov_ode.ferx` ODE; both `focei`, `maxiter = 0` → OFV at init, same data `two_cpt_oral_cov.csv`):

| form | OFV at init |
|------|-------------|
| analytical `two_cpt_oral(cl=CL, v1=V1, q=Q, v2=V2, ka=KA)` | **−1193.59** |
| ODE twin with `[scaling] obs_scale = V1` (default solver tol) | **−823.05** |
| **\|gap\|** | **~370** (ferx-r CI, tol 1e-10/1e-12: **534.4**) — expected **< 1** |

The shared `[individual_parameters]` block makes `obs_scale = V1` an η+covariate expression:

```
V1 = TVV1 * (WT / 70)^THETA_WT * exp(ETA_V1)
...
[scaling]
  obs_scale = V1
```

The analytical closed form (which folds `V1` into the concentration internally, not via `[scaling]`) is the reference; the **ODE `obs_scale` path is the side that regressed**.

## Attribution

- Previous ferx-r pin **`820e930e`** → equivalence test green (gap < 1, band 1.0).
[...]

PR #1337, opened 2026-09-10
fix(estimation): keep the BFGS partial in the IOV inner-loop fallback [closes #1327]

### LINKED ISSUE #1327: bug(nn): [covariate_nn] + IOV (kappa) — warm-started FOCEI objective sits 7k–15k below the cold re-evaluation; fits never approach the analytic IOV model, 27× slower per eval
## Summary

A `[covariate_nn]` model with IOV (`kappa` + `iov_column`) does not fit under FOCEI: the objective the outer optimizer sees during the run is **7 000–15 000 units below** what a cold-start evaluation of the same population parameters gives, so the optimizer chases a phantom, terminates after a handful of evals, and the reported fit is far worse than the analytic model it replaces. The same model without IOV, and the analytic model with IOV, are both fine. Each objective evaluation is also ~27× slower than the IOV analytic twin.

Found while running the DCM evaluation for #1310 on `ferx-testdata/busulfan_dominic` (1518 subjects, ~8 obs/subject, 2-cpt IV infusion, IOV on CL and V1 over up to 7 occasions).

## Reproduction (minutes)

`ferx-testdata/busulfan_dominic/ferx/dcm/dcm_h2_iov_s600.ferx` on `data/train_s600.csv` (600 subjects) — a 4→2→4 network, 22 weights, so it is small enough to run in a few minutes at 4 threads. Build with `--features ferx-core/nn` at or after `caba67d3`.

```
[parameters]
  block_omega (ETA_CL, ETA_V1) = [0.068, 0.038, 0.038]
  omega ETA_V2 ~ 0.077
  kappa KAPPA_CL ~ 0.0111
  kappa KAPPA_V1 ~ 0.0153
  sigma PROP_ERR ~ 0.05 (sd)
  sigma ADD_ERR  ~ 15 (sd)

[covariate_nn TYPICAL_PK]
  inputs = [LAGE, LWT, LHT, SEX]   center = [2.639, 3.727, 4.91, 0]   scale = [1.548, 0.907, 0.342, 1]
  outputs = [CL, V1, Q, V2]        layers = [2]   activation = tanh   output = softplus
  init = [11.6, 46.5, 14.3, 10.8]

[individual_parameters]
  CL = TYPICAL_PK.CL * exp(ETA_CL + KAPPA_CL)
  V1 = TYPICAL_PK.V1 * exp(ETA_V1 + KAPPA_V1)
  Q  = TYPICAL_PK.Q
  V2 = TYPICAL_PK.V2 * exp(ETA_V2)

[structural_model]
  pk two_cpt_iv(cl=CL, v1=V1, q=Q, v2=V2)
[error_model]
[fit_options]
  method = focei   iov_column = OCC   covariance = false
```

## What happens

| run | outer gradient | last optimizer eval | reported final (cold inner) | evals | wall |
|---|---|---|---|---|---|
| analytic base with IOV (`base_s600.ferx`) | analytic | 56 915.80 | 56 915.80 | 58 | ~5 min |
| DCM h2 + IOV | analytic (Dual2) | 71 456.17 | **80 766.57** | 386 | 13 min |
| DCM h2 + IOV, `gradient_method = fd` (BOBYQA) | FD | 72 181.92 ("Restored best-seen point") | **86 864.73** | 178 | 12 min |
| DCM h8 + IOV | analytic | 73 407.72 | **80 354.45** | 15 | 25 min (600 subj) |
| DCM h8 + IOV, full train (1215 subj) | analytic | 112 635 | **113 690.88** | 11 | 25 min |
| DCM h8, **no** IOV, same data | analytic | 59 838.63 | 59 838.63 | 2561 | 4 min |

Two facts separate the causes:

- **Cold evaluation is self-consistent.** `dcm_holdout` (an `outer_maxiter = 0` re-evaluation from the `.fitrx`, EBEs cold-started) on the training data returns 80 766.568430 for the analytic-gradient run and 56 915.799399 for the base — the reported numbers to the last digit. So the number the optimizer optimizes is the one that is wrong: the warm-started inner loop reports an objective ~9 300 too low for a network model with κ slots.
- **It is not the outer gradient.** Under FD/BOBYQA the trajectory and the cold value disagree by 14 700, more than under the analytic gradient.

And it is not a bad local mode of the same objective: the DCM+IOV runs never get anywhere near the analytic IOV base (best trajectory value 71 456 vs 56 916, holdout 31 915 vs 26 454), whereas the same DCM without IOV beats the no-IOV base on holdout by 100+.

## Where to look

`src/sens/provider.rs` — `iov_eta_only_derivs_dyn` builds the IOV inner η-gradient for a `[covariate_nn]` model with `ModelNnGuard` lifting the network outputs as constants, and returns `CombinedDerivs` with `dtheta` / `d2eta` **zero-filled** ("must not be consumed — they are not the derivatives, they are absent"). If the FOCEI objective's per-subject Hessian term on the IOV path consumes `d2eta` (or the mixed block) for the η-part of `[η_bsv, κ₁..κ_K]`, the objective is evaluated with a wrong (zero-curvature) Laplacian and comes out too low — which is the sign of the discrepancy observed. A second candidate is the warm-start bookkeeping for κ slots: the cold final pass is right and the warm path is wrong, so whatever state the warm path carries per occasion (cached NN outputs at a stale θ, or κ vectors keyed by a slot layout that the NN model builds differently) is worth checking. `Occasion slot 7: KAPPA_CL NaN` appears in the shrinkage report of both the DCM and the base, so that line is not the discriminator.

Also: each objective eval on the IOV DCM costs ~19 s on 600 subjects vs 0.7 s for the same network without IOV (and ~5 s for the analytic base with IOV). The `[covariate_nn]` weights are `n_theta` for the IOV walk's dual width even though the network output is η-independent, the same shape as #1300 on the TV-cov walk.

## What a fix needs to show

Trajectory value == cold re-evaluation at the final point (as the analytic IOV base already does, to 1e-6), and the DCM+IOV fit landing at or below the analytic IOV base on training OFV, since the network can represent the allometric model it replaces. `dcm_holdout` (`scratchpad` crate; 70 lines against `prepare_run` + `load_fit` + `fit` with `outer_maxiter = 0`) is the check.

Related: #1300 (TV-cov analytic walk for NN weights), #1310 (DCM tuning heuristics, where this dataset is a calibration point), #1317 (checkpoint stores latest not best).

roninsightrx:
## Why

#1327: a `[covariate_nn]` + IOV FOCEI fit reported a final OFV **9 300 above** the value the optimizer had been minimising (best-seen 71 456.17 → final 80 766.57 on `busulfan_dominic/dcm_h2_iov_s600`, 600 subjects), and an `outer_maxiter = 0` re-evaluation from the `.fitrx` reproduced the bad number. The issue suspected the η-only IOV gradient (`iov_eta_only_derivs_dyn`) or κ warm-start bookkeeping. Neither is it: at the initial θ and at the reported final θ the analytic stacked-`[η, κ]` gradient matches central FD to ≤ 1e-6 relative on every coordinate of the first 12 subjects, and `find_ebe_iov` from a cold seed and from a perturbed warm seed land on the same point.

The defect is the **inner-loop failure policy on the IOV path**. `find_ebe_iov`'s exact-objective arm ran a Nelder–Mead restart from the cold seed and adopted it *unconditionally* whenever BFGS reported non-convergence. Instrumenting the outer objective to re-solve every subject cold and diff per-subject contributions showed what that means in practice at eval 25 of the reproduction (Ω_V1 probed at 0.43 with corr(η_CL, η_V1) ≈ 0.98):

```
subject 261: warm  FOCEI contrib 125.8   inner NLL   94.6   η = (−2.04, −5.06, −0.71), κ within 0.25
             cold  FOCEI contrib 5060.3  inner NLL 5030.0   η = (−3.85, −0.55,  0.10), κ up to 2.5 (≈ 23 prior SD)
```

A bare BFGS from the cold seed reaches that mode (NLL 94.644, |g| = 6.6e-5) but cannot certify `|g| < 1e-5`, so the old arm threw it away for an 11-dimensional NM from zero (initial simplex step 0.00025) that stops at NLL 5030. Five subjects like this account for the −15 397 warm-vs-cold gap at that eval; the reported final and every cold re-evaluation carry the same failure. Warm-started evaluations converge within `tol` from the previous mode and never trip the fallback, which is why the trajectory looked fine and the final did not. The non-IOV `find_ebe` fixed exactly this in #378 (`argmin_inner_fallback`: keep the lower-objective of {BFGS partial, NM}); the IOV branch kept the pre-#378 policy.

## What changed

- `src/estimation/inner_optimizer.rs` — `find_ebe_iov`'s non-ODE fallback arm now calls `argmin_inner_fallback` (the same helper its ODE arm and the non-IOV path already use), so a BFGS partial is only replaced by an NM restart that is strictly better. The FREM-only "take NM unconditionally to re-center the proposal" arm of the non-IOV path has no IOV twin because IOV + FREM is unsupported (`foce_subject_nll_iov` returns a sentinel for it). Runaway guard, ODE-IOV NM skip, and the H-matrix assembly are untouched.
- Regression test `iov_inner_fallback_keeps_bfgs_partial_over_worse_cold_nm` (sibling `inner_optimizer_iov_tests.rs`).
- `CHANGELOG.md` entry under Fixed.

## Alternatives considered

- **Warm-start the final inner loop from the best-seen EBEs.** Would make the reported OFV match the trajectory but leave every cold entry point (`outer_maxiter = 0`, `.fitrx` reload, bootstrap/SSE re-fits) wrong. The fallback is the defect; fixing it makes cold and warm agree.
- **Loosen `inner_tol` / raise `inner_maxiter`.** Subject 261's BFGS stalls at 6.6e-5 with 200, 1 000 and 5 000 iterations alike — a line-search floor, not an iteration budget.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation

Reproduction: `ferx-testdata/busulfan_dominic/ferx/dcm/dcm_h2_iov_s600.ferx` on `data/train_s600.csv`, FOCEI, BOBYQA (the model's `iov_sens_supported` is false, so the outer is derivative-free), `--features ferx-core/nn`.

**Per-eval warm-vs-cold self-consistency** (outer objective instrumented to re-solve every subject cold, first 30 evals; `n|d|>1` = subjects whose contribution differs by more than 1):

```
                      before (main)                         after
eval  4–11   warm 107807.10  cold 107541.70  n=1     warm = cold = 107541.70  n=0
eval  16     warm 107564.58  cold 109247.48  n=1     warm = cold = 107564.58  n=0
eval  19     warm 106911.82  cold 113522.56  n=1     warm = cold = 106911.82  n=0
eval  25     warm  75914.70  cold  91312.21  n=5     warm = cold =  75667.92  n=0
eval  28     warm  87931.27  cold 105057.46  n=2     warm = cold =  86994.53  n=0
eval  31     warm 107259.38  cold 110935.40  n=3     warm = cold = 107259.38  n=0
```

Every eval 1–30 after the fix reports `diff = 0.0000`. (Evals 25 and 28 are also *lower* than before on the warm side: the warm path had the same defect on one subject each.)

**Full fit, best-seen (warm trajectory) vs cold final**, same model/data, 4 threads, release with `nn`:

| run | optimizer | evals | best-seen (trajectory) | final (cold inner) | gap |
|---|---|---|---|---|---|
| DCM h2 + IOV, main `86948249` | BOBYQA (32) | 386 | 71 456.17 | **80 766.57** | +9 310 |
| DCM h2 + IOV, this PR | BOBYQA (32) | 546 | 59 470.181501 | 59 470.181499 | 2e-6 |
| analytic IOV base, this PR | L-BFGS (14) | 41 | 56 915.799398 | 56 915.799398 | 0 |

The DCM+IOV fit now reports what it minimised, and its objective moves from 80 767 to 59 470; the debug-instrumented run shows `warm − cold = 0.0000` on all 170 evals it reached. The negative EPS shrinkage warning (−12.9%) is gone (+20.4%). It still lands 2 554 above the analytic IOV base, which is the open optimizer question below (derivative-free BOBYQA over 32 coordinates), not the inner-loop defect this PR fixes.

- [x] Compared against: prior ferx commit `86948249` (main) on the same model/data
- [x] Regression fixture: closed-form 1-cpt IOV, 2 occasions, `max_iter = 1` — cold-seed NLL 504.88, converged mode −4.13, kept BFGS partial −1.11, old-policy NM 500.98 (bound: within 10 of the mode; mutation-verified red under the old arm)

## Performance
- [x] No significant impact expected — the fallback arm runs the same single NM it ran before, plus two objective evaluations (`argmin_inner_fallback`), and only when BFGS reports non-convergence.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib --features nn`: 4490 passed)
- [x] Regression test added that fails without this fix (mutation: restoring the unconditional-NM arm → `returned NLL 500.9791, converged mode -4.1340`)

## Docs
- [x] No docs page change (no user-visible option changed)

## Checklist

## Docs & examples
- [x] No example execution step needed (fix, no user-visible change)

## Reviewer hints

The whole fix is the `else` arm of the BFGS-failure block in `find_ebe_iov`; compare with the non-FREM `else` arm of `find_ebe` ~50 lines earlier — they are now the same call. The interesting part is the regression fixture's discriminating bound (see its doc comment for the four measured numbers).

## Open questions

Not addressed here, tracked in #1339: for a `[covariate_nn]` + IOV model the outer optimizer is BOBYQA and the outer gradient (when requested) is FD-reconverge, because `iov_sens_supported` is the strict θ-axis predicate. That is the 27× per-eval cost and the reason the DCM+IOV fit converges so slowly — the #1300 `ModelNnAxisGuard` walk would need an IOV twin to make it analytic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ section of the SDLC for the versioning policy).
 
 ### Fixed
 
+- Fixed the IOV inner loop discarding a converged-in-all-but-name BFGS solution for a far worse Nelder–Mead restart from the cold seed, which made every cold-started evaluation of a `kappa` model (the reported final OFV, `outer_maxiter = 0` re-evaluations, `.fitrx` reloads) score some subjects thousands of −2LL units above the value the optimizer had minimised — 9 300 on a `[covariate_nn]` + IOV busulfan fit (#1327).
 - Fixed ODE-accumulated survival hazards that read `TAD`, which could reject valid multi-dose subjects with a misleading finite objective (#1261).
 
 ### Performance
--- a/src/estimation/inner_optimizer.rs
+++ b/src/estimation/inner_optimizer.rs
@@ -1331,26 +1331,31 @@ fn find_ebe_iov(
         None,
         enable_stall,
     );
-    // On BFGS failure, recover with the same ODE-gated policy as the non-IOV `find_ebe`
-    // (cold seed = prior mode `bsv_psi = μ`, κ = 0): for ODE objectives keep the
-    // lower-objective of {BFGS partial, NM restart} so a correct η̂ floored above `tol` by
-    // solver noise is never discarded (#555); for exact objectives recover with NM from the
-    // cold seed, as prior releases, so a non-stationary low-objective partial can't be kept.
+    // On BFGS failure, recover exactly as the non-IOV `find_ebe` does (cold seed = prior
+    // mode `bsv_psi = μ`, κ = 0): keep the lower-objective of {BFGS partial, NM restart}
+    // (`argmin_inner_fallback`). This holds for ODE objectives (#555, a gradient-noise floor
+    // blocks certification at a genuine mode) AND for exact objectives (#378 for the non-IOV
+    // path; #1327 here): a closed-form BFGS started far from the mode routinely arrives at it
+    // and stalls at a gradient norm just above `tol` (busulfan DCM+IOV subject 261: partial
+    // at the mode, NLL 94.6, |g| ≈ 7e-5 > 1e-5). Until #1327 this branch then replaced that
+    // partial *unconditionally* with an 11-dimensional Nelder–Mead from the cold seed, which
+    // stops at NLL ≈ 5030 with κ tens of prior SDs out — so every cold-started evaluation
+    // (the final inner loop, an `outer_maxiter = 0` re-evaluation) scored a handful of
+    // subjects thousands of −2LL units too high, while the warm-started trajectory (whose
+    // BFGS converges within `tol` from the previous mode) never saw it. The reported OFV
+    // sat 9 300 above the value the optimizer had minimised. IOV + FREM is unsupported
+    // (`foce_subject_nll_iov` returns a sentinel for it), so the FREM-only "take NM
+    // unconditionally to re-center the proposal" arm of the non-IOV path has no twin here.
     let (nm_converged, mut used_fallback) = if !bfgs_converged {
         let partial = x.clone();
         let mut cold = vec![0.0; n_flat];
         cold[..n_eta].copy_from_slice(&mu);
         if skip_ode_iov_nm_fallback(model) {
             (false, false)
-        } else if enable_stall {
+        } else {
             let (best, ok) = argmin_inner_fallback(&obj, &partial, &cold, n_flat, max_iter, tol);
             x = best;
             (ok, true)
-        } else {
-            let warm = ebe_warm_start_enabled() && partial.iter().all(|v| v.is_finite());
-            x = if warm { partial } else { cold };
-            let nm_ok = nelder_mead_minimize(&obj, &mut x, n_flat, max_iter * 5, tol);
-            (nm_ok, true)
         }
     } else {
         (false, false)
--- a/src/estimation/inner_optimizer.rs
+++ b/src/estimation/inner_optimizer.rs
@@ -495,16 +495,21 @@ pub struct InnerLoopStats {
 ///
 /// The restart seeds from the BFGS `partial` when `ebe_warm_start` is set (it sits on the
 /// steep prior slope, so NM reaches the mode in fewer steps), else from `cold_seed` (η=0
-/// for non-IOV, `[μ, 0…]` for IOV — the historical reset point). Exactly **one** NM solve
-/// runs, so enabling `ebe_warm_start` is never slower than leaving it off.
+/// for non-IOV, `[μ, 0…]` for IOV — the historical reset point). One NM solve runs when
+/// the restart wins or under `ebe_warm_start`; a second, started from the partial, runs
+/// only when a cold restart loses to the partial (see below).
 ///
-/// Returns `(eta, nm_converged)`. The **value** is the lower-objective of the BFGS partial
+/// Returns `(eta, converged)`. The **value** is the lower-objective of the BFGS partial
 /// and the NM restart — the substantive #555 fix: the previous code overwrote `eta` with the
 /// NM restart unconditionally, discarding a correct partial that BFGS had reached but not
-/// gnorm-verified. `nm_converged` is the Nelder–Mead convergence flag (as the pre-#555 code
-/// reported), so the per-subject convergence/diagnostic semantics are unchanged; the η̂
-/// *value* fed to the FOCEI gradient is what improves. A non-finite `obj(partial)` (NaN/∞)
-/// makes the partial unusable so the NM result is taken.
+/// gnorm-verified. `converged` is the Nelder–Mead flag **of the run that ended at the
+/// returned point** (PR #1337 review): when the restart wins it is that restart's flag;
+/// when the partial wins, the partial — which its own BFGS run did *not* certify — is
+/// polished by an NM started from it, and that run's end point and flag are returned (the
+/// end point is never worse than the partial). Returning the restart's flag with the
+/// partial's point reported a converged EBE whenever NM had converged in a *worse* basin,
+/// which bypassed `max_unconverged_frac` and AGQ's per-subject acceptance. A non-finite
+/// `obj(partial)` (NaN/∞) makes the partial unusable so the NM result is taken.
 fn argmin_inner_fallback(
     obj: &dyn Fn(&[f64]) -> f64,
     partial: &[f64],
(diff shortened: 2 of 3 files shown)


roninsightrx:
[P1] Keep convergence status tied to the selected EBE

At src/estimation/inner_optimizer.rs:1356-1358, ok is the convergence flag for the Nelder-Mead candidate, but argmin_inner_fallback may return the BFGS partial instead when that partial has the lower objective. If NM converges in a worse basin, this branch therefore returns the non-converged BFGS partial while EbeResult.converged becomes true. That bypasses max_unconverged_frac, and AGQ explicitly accepts the returned point based on this flag (src/estimation/agq.rs:2061), even though the selected point never met the tolerance and may not be a mode. This behavior already exists in the shared helper, but this PR newly exposes it to exact IOV fits. Please return convergence for the candidate that was actually selected (the partial should remain false unless an explicit gradient/stationarity check certifies it), and add coverage for the case where NM converges in a worse basin while the partial wins on objective.

roninsightrx:
Agreed, and fixed in 835b806a — in the shared helper, since the flag was wrong on the non-IOV path too (this PR only made it reachable for exact IOV fits).

**What changed.** `argmin_inner_fallback` now returns the convergence flag of the run that ended at the point it returns:

- restart wins → the restart's point and flag, as before;
- partial wins → the partial is polished by a second Nelder–Mead started **from the partial**, and that run's end point and flag are returned. The end point is never worse than the partial (NM keeps its best vertex); a genuine mode collapses the simplex within `tol` in a few iterations and reports `true`, a non-stationary partial keeps descending and reports `false` when the budget ends. Under `ebe_warm_start` the single restart already ran from the partial, so it is the certification and nothing is re-run.

I went with a derivative-free certification rather than a gradient/stationarity check because the helper's callers include the ODE path, where a gradient-norm test is exactly what BFGS could not satisfy at the mode (#555's noise floor). NM's simplex collapse is robust to that.

**Coverage.** `argmin_inner_fallback_flag_belongs_to_the_returned_point`: partial on the slope of the deep well (not a mode) and winning on objective, cold restart seeded on the shallow minimum and converging there. With a budget too small to polish, the returned point is ≥ as good as the partial and the flag is `false`; with a normal budget the polish reaches the deep mode and reports `true`. Reverting to "return the restart's flag" turns the first half red (`a partial that was neither BFGS- nor NM-certified must not report converged`). The existing #555 tests (`repro555_ode_exprscale_ebe_finds_global_min`, which asserts `converged` at the mode after the fallback) and the #1327 regression still pass.

**Cost.** One extra NM only on the fallback path and only when a cold restart loses to the partial — the case that previously returned a mislabelled point.

Issue #1331, opened 2026-09-10
feat(nn): export a fitted DCM as a plain population PK model — write the neural net out as [individual_parameters] equations


roninsightrx:
## Motivation

A fitted `[covariate_nn]` block is a black box in three ways that matter after the fit is done:

- it is **unreadable** — `W_TYPICAL_PK_37 = -0.412` says nothing about the covariate model;
- it is **not portable** — the model only loads in a build with the `nn` cargo feature, and downstream tools (`ferxtranslate`, NONMEM) have no `[covariate_nn]` spelling at all;
- it is **not auditable** — a regulator/reviewer asking "what does the model do to CL at WT = 45?" has to run the engine.

But after estimation the weights are *constants*, and an MLP at fixed weights is a closed-form algebraic function of the covariates. Every activation ferx supports (`identity`, `relu`, `sigmoid`, `exp`, `tanh`, `softplus`) is writable in the `.ferx` expression DSL, and `[individual_parameters]` already allows named intermediates evaluated top-to-bottom. So a fitted DCM can be emitted as an ordinary population PK model with no NN block and no `nn` feature — same predictions, plain equations.

Related but different: #1310 (`dcm_suggest`) goes *into* a DCM; this goes *out* of one.

## Proposed API

New module `src/nn/export.rs` (ferx-core — it is a model→model transform, zero calls to `fit()`, so it is core, not `ferx-tools` per the boundary rule):

```rust
pub enum DcmWeightForm {
    /// Weights inlined as numeric literals. Default — the exported model is frozen.
    Literal,
    /// Weights emitted as `theta W_<NN>_<i>(value) FIX` declarations, so the
    /// exported model still names them and can be perturbed by hand.
    FixedTheta,
}

pub struct DcmExportOptions { pub weight_form: DcmWeightForm }

/// Expand every `[covariate_nn]` block in `model_text` into explicit
/// `[individual_parameters]` statements at the weights in `theta`.
pub fn export_dcm_as_analytic(
    model_text: &str,
    theta: &[f64],
    opts: &DcmExportOptions,
) -> Result<String, NnError>;
```

Inputs are already available: `FitResult.model_text` is round-tripped through `.fitrx`, and `FitResult.neural_networks` (`NeuralNetworkInfo`) carries `shape`, `weights_offset`, `n_weights`, `hidden_activation`, `output_activation`, `input_names`, `output_names`, `input_center`, `input_scale`.

[...]

PR #1338, opened 2026-09-10
fix(parser): reject an unknown function name instead of evaluating it as the identity [closes #1332]

### LINKED ISSUE #1332: bug(parser): an unknown function name is silently the identity — `tanh(ETA_CL)` fits, converges and computes `ETA_CL`
## What happens

`parse_atom` builds `Expression::UnaryFn(name, arg)` for **any** identifier followed by `(`, with no check that the name is a function ferx knows (`src/parser/model_parser.rs:22416`):

```rust
return Ok((Expression::UnaryFn(func_name, Box::new(arg)), p + 1));
```

Every consumer of `UnaryFn` then ends its match in a fallthrough that returns the **argument unchanged** — i.e. the unknown function is silently the identity:

| Consumer | Site | Fallthrough |
|---|---|---|
| `eval_expr` (slow path) | `model_parser.rs:18410` | `_ => v,` |
| bytecode compiler | `model_parser.rs:18861` | `_ => { /* unknown function → leave the argument on the stack */ }` |
| symbolic derivative | `model_parser.rs:21378` | `_ => da` |

So:

```
[individual_parameters]
  CL = TVCL * tanh(ETA_CL)      # tanh is not a ferx builtin
```

parses clean, passes `ferx check`, fits, converges, and computes `TVCL * ETA_CL`. The derivative path agrees with the value path, so no internal check disagrees either — the model is *consistently* the wrong model. There is no warning, no `FitResult.warnings` entry, and nothing in the output names the dropped call.

This is by code inspection at the three sites above, not from a run — the fallthrough arms are unconditional, so no fixture is needed to see it.

## Why it matters

This is the same failure class the parser has already fixed twice, both times by rejecting the spelling *by name*:

- **#1030** — a one-argument `max(x)` used to fall through to `UnaryFn("max", x)` and evaluate as the identity. Now a parse error.
- **#1092** — a one-argument `clamp(x)` did the same: *"a no-op that fits, converges and gives wrong numbers"* (the comment at `model_parser.rs:22408`). Now a parse error.

Both fixes are arity checks on names the parser already recognised. Neither closes the general case: **any** unrecognised name still no-ops. The realistic triggers are ordinary:

- a function the user assumes exists because it exists in NONMEM/nlmixr2/mrgsolve — `tanh`, `sinh`, `pow`, `sign`, `heaviside`, `LOG10`;
- a typo — `exp1(x)`, `sqr(x)`, `lgo(x)`;
- **case** — worth confirming whether the match is case-sensitive; if it is, `EXP(ETA_CL)` and `LOG(WT)` no-op, and those are the NONMEM spellings a `$PK` block gets pasted from.

Consequence in each case is a silently different model, not an error.

## Proposed fix

One whitelist, consulted at parse time, in the same place the `min`/`max`/`clamp` arity errors are raised. Reject an unrecognised name with a message that lists what is available:

```
unknown function `tanh` in [individual_parameters]. Supported: exp, log, ln, sqrt,
abs, floor, ceil, round, logit, inv_logit, expit, min(a,b), max(a,b),
clamp(x,lo,hi), present(x) (conditions only).
```

Deriving the list from a single `const` that the three consumer matches are pinned against (a test asserting every whitelisted name has a non-fallthrough arm in each of the three) keeps the whitelist from drifting past the implementations — "when two implementations disagree, the fix is one implementation" (CLAUDE.md).

Once the name is guaranteed known, the three `_ =>` arms become genuinely unreachable and should say so (`unreachable!()`), so a name added to the whitelist without an arm in all three consumers is a panic in test, not a silent identity in production.

**Blast radius.** This is a *hardening* change that turns previously-accepted model files into parse errors. Any file that currently relies on the no-op is already computing something other than what it reads as, so the errors are all true positives — but it is worth grepping `examples/`, `tests/`, `docs/**/*.qmd` and the `nonmem_anchor/` fixtures for a name outside the whitelist before merging, and noting the change under `Changed` in `CHANGELOG.md`.

## Secondary finding — `floor` / `ceil` / `round` differentiate as the identity

Separate defect at the same site, visible while reading it. Those three have arms in `eval_expr` (`:18393-18395`) and in the bytecode compiler (`:18857-18859`), but **no arm in the symbolic derivative** (`:21334-21382`), so they land on the `_ => da` fallthrough whose comment explicitly scopes itself to *unknown* names:

```rust
_ => {
    // Unknown name — the slow path returns the argument
    // unchanged, so the derivative is its argument's
    // derivative. (See `eval_expression_indexed`'s
    // `_ => v` fallthrough.)
    da
}
```

`d/dx floor(x)` is therefore `1`, not `0`. Any `[individual_parameters]` expression that rounds — a dose-band lookup, an integer occasion index derived from `TIME` — feeds a wrong analytic derivative to the gradient while the value path is correct, which is exactly the shape the `Dual2`-vs-FD parity rule exists to catch and which no current fixture appears to exercise. Fix alongside (`floor`/`ceil`/`round` → derivative `0`), or split out if it wants its own regression test — either way it must not be left sitting under an `unreachable!()` when the whitelist lands.

## Tests

- **Parse rejection**, one per trigger: an unknown name, a typo, a NONMEM-cased `EXP(` (if case-sensitive), each in `[individual_parameters]`, `[odes]`, `[derived]` and `[error_model]` — the check has to sit where every block's expressions go through it, not on one block's path.
- **Whitelist ↔ implementation pin**: every name in the `const` has a real arm in all three consumers; adding a name to the list without an arm fails.
- **The regression this exists to catch, mutated**: delete the parse-time check and the rejection tests must all redden — and check the *fixture* can reach it, i.e. the model must actually be parsed, not short-circuited by an earlier validation error on the same line.
- `floor`/`ceil`/`round` derivative = 0, asserted against finite differences away from an integer boundary.

## Context

Blocks #1331 (DCM → plain population PK export), which emits `tanh(...)` / `softplus(...)` into `[individual_parameters]`. Without a parse-time rejection, an export that runs ahead of the `tanh` builtin produces a wrong model with no error — the exporter becomes a generator of exactly this bug.

roninsightrx:
## Why

`parse_atom` built `Expression::UnaryFn(name, arg)` for **any** identifier followed by `(`, with no check that the name was a function ferx knows. All three consumers then ended their match on a fallthrough that returned the argument unchanged, so an unrecognised function was silently the **identity**:

```
[individual_parameters]
  CL = TVCL * tanh(ETA_CL)      # tanh is not a ferx builtin
```

parsed, passed `ferx check`, fitted, converged and computed `TVCL * ETA_CL`. The derivative path agreed with the value path, so no internal check disagreed either — the model was *consistently* the wrong model, with no warning, no `FitResult.warnings` entry, and nothing in the output naming the dropped call.

This is the same failure class the parser has already fixed twice by rejecting a spelling **by name** — #1030 (one-argument `max(x)`) and #1092 (one-argument `clamp(x)`, *"a no-op that fits, converges and gives wrong numbers"*). Neither closed the general case. The realistic triggers are ordinary: a function that exists in NONMEM / nlmixr2 / mrgsolve but not here (`tanh`, `sinh`, `sign`, `log10`), or a plain typo (`sqr(x)`, `exp1(x)`).

Blocks #1331 (DCM → plain population PK export), which emits `tanh(...)` / `softplus(...)` into `[individual_parameters]`. Without a parse-time rejection, an export that runs ahead of a `tanh` builtin makes the exporter a generator of exactly this bug.

Closes #1332.

## What changed

**One whitelist, consulted at the single site that builds the node.** `SUPPORTED_UNARY_FNS` is the closed list of callable one-argument names (`exp`, `log`, `ln`, `sqrt`, `abs`, `floor`, `ceil`, `round`, `logit`, `inv_logit`, `expit`). Because the check sits at the one `return Ok((Expression::UnaryFn(...)))` in `parse_atom`, it covers every block whose expressions go through the expression parser rather than one block's path:

```
unknown function `tanh`. Supported: exp, log, ln, sqrt, abs, floor, ceil, round,
logit, inv_logit, expit, min(a, b), max(a, b), clamp(x, lo, hi),
present(x) (conditions only)
```

`min`/`max`/`clamp` are deliberately absent from the const: `parse_atom` desugars them to `Expression::Conditional`, so they never reach `UnaryFn`. Their existing arity errors still fire first and keep their own, more specific messages (checked — deleting either reddens `test_clamp_arity_diagnostics`, so these are not two gates covering for each other). `present(x)` in **value** position gets its own message pointing at condition position, since the name is known, just not there.

**The three `_ =>` arms become `unreachable!()`.** `Expression` is `pub(crate)` and the parser is its only constructor, so the whitelist makes the tails genuinely unreachable — a name added to the list without an arm in `eval_expr`, `compile_expr_into` *and* `differentiate_with_chain` is a panic in test rather than a silent identity in production. All three share one named panic message so the failure reads the same wherever it comes from, and each interpolates which consumer it was.

**Secondary defect, fixed alongside:** `floor` / `ceil` / `round` had arms in `eval_expr` and in the bytecode compiler but **none** in the symbolic differentiator, so they landed on the same identity fallthrough and `d/dx floor(x)` was `1`, not `0`. Any `[individual_parameters]` or `[odes]` expression that rounds — a dose-band lookup, an occasion index derived from `TIME` — was feeding a wrong analytic gradient to the estimator while its value path was correct. They now differentiate to `0`, the integer boundaries undefined and ignored exactly as `BinOp::Mod` already does.

## Alternatives considered

- **Warn instead of reject.** Rejected: the whole failure mode is that the wrong model *converges*, and a warning in a fit that reports a plausible OFV is exactly what got missed. A file that relies on the no-op is already computing something other than what it reads as.
- **Leave the `_ =>` arms as the identity and only gate at parse time.** That keeps the gate one-sided: a future builtin added to the whitelist but wired into only two of the three consumers would silently differentiate (or evaluate) as the identity again — the #1332 shape, reintroduced through the front door. `unreachable!()` turns that into a test failure.
- **`#[doc(hidden)]`-style escape hatch / a second list per consumer.** Rejected on the "when two implementations disagree, the fix is one implementation" rule — the const is the single source and the pin test forces the three matches to agree with it.

## Cross-repo dependency

No public Rust API change (`Expression` is `pub(crate)`; `cargo public-api` baseline unchanged), so ferx-r needs no lock bump for this.

## Breaking changes
- [x] `.ferx` model file format (migration note below)
- [ ] Public Rust API
- [ ] None

<details>

A model file that called a function ferx does not have now fails to parse instead of silently dropping the call. **Every such error is a true positive** — the file was already computing something other than what it reads as. The fix is to write the function out in terms of the supported set, e.g. `tanh(u)` as `2*inv_logit(2*u) - 1`.

Names are matched case-insensitively (they are lower-cased before dispatch), so the NONMEM `$PK` spellings `EXP(...)` / `LOG(...)` are unaffected; `TANH(...)` is rejected exactly like `tanh(...)`.

Blast radius was checked, not assumed: nothing in `examples/`, `tests/`, `docs/**/*.qmd` or `nonmem_anchor/` uses a name outside the whitelist. The single place that named `tanh` was a comment in `tests/vi_dcm_omega_recovery.rs` documenting this bug as unfixed and writing the term out as `2·inv_logit(2u) − 1` to dodge it — comment updated. (`tanh` also appears in `docs/model-file/covariate-nn.qmd` as a **`[covariate_nn]` activation-function name**, a different namespace parsed as a key, not an expression — unaffected.)

</details>

## Numerical validation
- [x] Not applicable — no NONMEM anchor exists for either half, and neither needs one:
  - The parse rejection produces **no numbers at all**; it converts a silently-wrong fit into an error. There is nothing to anchor.
  - `floor`/`ceil`/`round` differentiating to 0 is an **exact closed form**, not a tolerance: the function is piecewise constant, so the derivative is 0 everywhere off the steps. It is pinned against central finite differences of the production value path at a point strictly inside a step (`3·θ₀ = 6.9`, FD half-width `3e-5`), which is what makes FD a valid oracle there — evaluated *on* a boundary FD would report `1/(2h)` and measure the discontinuity instead of the arm. The value path is separately asserted to actually round (`floor → 6`, `ceil`/`round → 7`), so "derivative 0" cannot be right for the wrong reason.

## Performance
- [x] No significant impact expected — one `contains` over an 11-element `&[&str]` per function call **at parse time**. The evaluation and compilation paths are unchanged; the `unreachable!()` arms are cold.

## Tests

New Tier-1 tests in `src/parser/model_parser_tests.rs`:

| test | what it pins |
|---|---|
| `unknown_function_name_is_a_parse_error_in_every_block` | 9 trigger names × 7 expression positions — `[individual_parameters]` plain and inline-`if` arm, `[odes]` statement and `if` condition, `[scaling]`, `[derived]`, `[error_model]` magnitude expression |
| `unknown_fn_probe_sources_parse_when_the_probe_is_known` | the **straddle**: the same fixtures with `sqrt` in the hole must parse, so a rejection above cannot be an unrelated error on the probe line |
| `function_name_matching_is_case_insensitive_on_both_sides` | `EXP(` / `Log(` / `Inv_Logit(` still parse; `TANH(` / `Tanh(` rejected |
| `present_in_value_position_is_rejected_and_points_at_conditions` | plus the positive control that `if (present(WT))` still parses |
| `every_supported_unary_fn_has_an_arm_in_all_three_consumers` | the whitelist ↔ implementation pin |
| `every_supported_unary_fn_parses_in_a_model_file` | closes the loop: whitelisted ⇒ parses ⇒ has a real arm |
| `differentiate_floor_ceil_round_is_zero_away_from_a_step` | the secondary defect, vs FD |

Two of these deserve a note on *why they can fail*, per the repo's rules on assertions that cannot observe the defect:

- **The pin does not rely on the panic.** `unreachable!()` would make a missing arm panic, but a *reintroduced* `_ => v` / `_ => da` fallthrough would not. So each probe is also asserted non-identity on both sides: the argument is `2·θ₀`, so an identity fallthrough gives value `x` and derivative `2`, and every probe is chosen so the true answers are neither (`abs` probed at a negative argument, `log` at 2, `sqrt` at 9, the rounding trio at 6.9, …). The probe table is required to cover `SUPPORTED_UNARY_FNS` exactly, so adding a name to the const without a probe is also red.
- **Fixtures reach the guard.** The `[odes]` probes take `CL` rather than `TVCL` because an ODE RHS may not reference a θ — with `TVCL` the *rejection* test still passed, on a different error, while the straddle fixture was invalid. The straddle test is what caught that; it is the reason it exists.

Mutation-tested, **each side named separately** rather than only toggling the fix as a whole:

| mutation | tests that redden |
|---|---|
| disable the `SUPPORTED_UNARY_FNS` guard in `parse_atom` | 3 (the rejection tests) |
| delete `"floor"` arm in `eval_expr`, **with the identity fallthrough restored** so the panic cannot be doing the work | 4 |
| delete `"floor"` arm in `compile_expr_into` | 4 |
| delete the `"floor" \| "ceil" \| "round"` derivative arm | 3 |

Full runs on this branch: `cargo test --lib` → **4385 passed, 0 failed**; all **183** integration test binaries under `--features ci,survival` green (run serially — the machine was memory-saturated, three parallel `cargo test` runs were OOM-killed); `tools/preflight.sh` → `preflight OK — fmt check clippy rustdoc docs public-api`.

- [x] Not applicable (fix with no new API surface) — the user-visible change is an error message, shown above and in the docs.

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added (one under `Changed` for the rejection, one under `Fixed` for the derivative)
- [ ] New pages linked in `docs/_quarto.yml` (no new pages)

`docs/model-file/individual-parameters.qmd` — the *Supported Operators and Functions* table was itself incomplete (it omitted `floor`/`ceil`/`round`/`logit`/`inv_logit`/`expit`, all of which have always worked); completed it, then added the closed-list statement, the error message, the case-insensitivity note and the rounding-derivative consequence. `docs/model-file/scaling.qmd` gains a cross-reference so the "available in every block" claim next to `min`/`max`/`clamp` also says the set is closed. `cargo run -p docs-lint -- --check` clean.

## Checklist

### ferx-site / ferx-book
- [ ] New `.ferx` DSL syntax or `[fit_options]` key → n/a (this removes an accidental non-feature; no syntax added)
- [ ] New example `.ferx` file → n/a
- [ ] New data file → n/a
- [ ] New estimator, DSL feature, or fit option → n/a

### Example execution
- [x] No example execution step needed — `test_parse_all_example_ferx_files` parses every `examples/*.ferx` on every `--lib` run and is green, which is the check that matters for a parser-hardening change.

## Reviewer hints

Focus on **`SUPPORTED_UNARY_FNS` and its three consumers** (`src/parser/model_parser.rs`): the argument that `unreachable!()` is sound rests entirely on `Expression` being `pub(crate)` with `parse_atom` as its only constructor — worth checking that claim rather than taking it. `git grep 'UnaryFn'` outside `model_parser.rs` returns only the test file. Inside it, `UnaryFn` is built with a literal name in exactly five places, all whitelisted — `sqrt` in the IIV-weighting rewrite, and `exp` / `sqrt` / `inv_logit` / `ln` inside `differentiate_with_chain`'s own output — plus `simplify_expr`, which clones an existing node's name and so cannot introduce a new one.

Second thing worth a look: whether the whitelist should have been *wider*. `tanh` and `softplus` are the two names #1331 wants, and this PR deliberately does not add them — the point is that the export must fail loudly until the builtins exist, rather than emit a model that quietly computes something else. If you would rather land the builtins in the same change, say so and I will add them with their arms and probes (each is a three-line addition plus one row in the probe table).

The docs diff and the `tests/vi_dcm_omega_recovery.rs` comment change can be skimmed.

## Open questions

Should `tanh` / `softplus` land here as real builtins rather than in #1331? The whitelist is designed to make that a small, well-pinned addition either way, but doing it in this PR would mean #1331 has nothing left to block on.

https://claude.ai/code/session_01MATN4FDqQbmSa8aApKmPaF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,12 @@ section of the SDLC for the versioning policy).
 ### Changed
 
 - **SAEM now averages the residual sufficient statistic for eligible single additive and proportional error models, reducing final-draw Monte Carlo noise in the residual SD estimate (#1321).**
+- **Calling a function ferx does not have is now a parse error naming the call and listing what is available, instead of silently evaluating as the identity.** `CL = TVCL * tanh(ETA_CL)` used to parse, pass `ferx check`, fit and converge while computing `TVCL * ETA_CL`. This applies to every block that parses expressions (`[individual_parameters]`, `[odes]`, `[scaling]`, `[derived]`, `[error_model]` magnitudes) and to conditions. Model files that relied on the no-op were already computing something other than what they read as, so the new errors are all true positives. Names remain case-insensitive, so `EXP(...)` / `LOG(...)` are unaffected; `present(x)` in value position now points at condition position (#1332).
 
 ### Fixed
 
 - Fixed ODE-accumulated survival hazards that read `TAD`, which could reject valid multi-dose subjects with a misleading finite objective (#1261).
+- `floor(x)`, `ceil(x)` and `round(x)` now differentiate to `0` rather than to `x`'s own derivative. An `[individual_parameters]` or `[odes]` expression that rounds — a dose-band lookup, an occasion index derived from `TIME` — was feeding a wrong analytic gradient to the estimator while its value path was correct (#1332).
 
 ### Performance
 
--- a/docs/model-file/individual-parameters.qmd
+++ b/docs/model-file/individual-parameters.qmd
@@ -49,6 +49,9 @@ name from quietly reading as zero and collapsing the formula.
 | `log()`, `ln()` | `log(TVCL)` |
 | `sqrt()` | `sqrt(WT)` |
 | `abs()` | `abs(ETA_CL)` |
+| `floor()`, `ceil()`, `round()` | `floor(AGE / 10)` |
+| `logit()` | `logit(THETA_F)` |
+| `inv_logit()`, `expit()` | `inv_logit(logit(THETA_F) + ETA_F)` |
 | `min(a, b)`, `max(a, b)` | `max(TVCL * WT / 70, 0.1)` |
 | `clamp(x, lo, hi)` | `clamp(ACR20, 0.01, 0.99)` |
 | `present(x)` (condition only) | `if (present(WT)) (WT/70)^0.75 else 1.0` |
@@ -72,6 +75,32 @@ arity is a parse error, and literal bounds the wrong way round
 (`clamp(x, 0.9, 0.1)`) are rejected rather than silently returning `0.9` for
 every `x`; bounds that are not both literals are not checked
 ([#1092](https://github.com/FeRx-NLME/ferx-core/issues/1092)).
+
+The table above is the **complete** list. Calling anything else is a parse error
+naming the call and listing what is available:
+
+```
+unknown function `tanh`. Supported: exp, log, ln, sqrt, abs, floor, ceil, round,
+logit, inv_logit, expit, min(a, b), max(a, b), clamp(x, lo, hi),
+present(x) (conditions only)
+```
+
+Before ferx 0.5 an unrecognised name was silently the **identity**, so
+`CL = TVCL * tanh(ETA_CL)` parsed, passed `ferx check`, fitted and converged
+while computing `TVCL * ETA_CL` — a consistently wrong model with no warning
+anywhere ([#1332](https://github.com/FeRx-NLME/ferx-core/issues/1332)). The
+realistic triggers were ordinary: a function that exists in NONMEM, nlmixr2 or
+mrgsolve but not here (`tanh`, `sinh`, `sign`, `log10`), and plain typos
+(`sqr(x)`, `exp1(x)`). Names are matched case-insensitively, so the NONMEM `$PK`
+spellings `EXP(...)` / `LOG(...)` keep working; `TANH(...)` is rejected exactly
+like `tanh(...)`.
+
+`floor`, `ceil` and `round` are piecewise constant, so their analytic derivative
+is **0** away from a step (the integer boundaries are undefined and ignored, as
+for `%`). A parameter that rounds is therefore flat with respect to the value it
+rounds, which is usually what a dose-band or occasion-index lookup wants — but it
+does mean an `η` that only enters through a rounded quantity gets a zero
+gradient.
 
 ## Continuation Lines
 
--- a/docs/model-file/scaling.qmd
+++ b/docs/model-file/scaling.qmd
@@ -414,6 +414,12 @@ expression the DSL parses — `[scaling]`, `[individual_parameters]`, `[odes]`,
 `if (a >= b) a else b`), so they differentiate and compile exactly like the
 hand-written form.
 
+The set of callable names is closed, and the same in every one of those blocks —
+see [Supported Operators and
+Functions](individual-parameters.qmd#supported-operators-and-functions). Anything
+outside it is a parse error rather than the silent identity it used to be
+([#1332](https://github.com/FeRx-NLME/ferx-core/issues/1332)).
+
 The desugaring puts **both** arguments in the guard *and* in a branch, so each one
 appears twice in the compiled tree, and the duplication multiplies through nesting:
 `min(max(x, 0.01), 0.99)` contains four copies of `x`, and reading that clamp from
--- a/src/parser/model_parser.rs
+++ b/src/parser/model_parser.rs
@@ -16102,6 +16102,58 @@ fn build_pk_param_fn(
 // `types.rs`). They remain unexported from the crate root — external users
 // can't construct or pattern-match them.
 
+/// Every function name `Expression::UnaryFn` may carry (#1332).
+///
+/// `parse_atom` rejects any other `name(...)` in expression position, so this
+/// list is the *complete* domain of the `UnaryFn` dispatch in the three
+/// consumers that give the node meaning — `eval_expr`, `compile_expr_into` and
+/// `differentiate_with_chain`. Each of those ends its match on
+/// `unreachable!(UNARY_FN_ARM_MISSING)` rather than on the identity
+/// fallthrough it used to carry: an unrecognised name used to be silently the
+/// identity, so `CL = TVCL * tanh(ETA_CL)` parsed, fitted, converged and
+/// computed `TVCL * ETA_CL` with no warning anywhere.
+///
+/// The list and the three matches are pinned against each other by
+/// `every_supported_unary_fn_has_an_arm_in_all_three_consumers`, so adding a
+/// name here without an arm in all three is a red test rather than a
+/// production panic. `Expression` is `pub(crate)` and the parser is its only
+/// constructor, which is what makes the `unreachable!` sound.
+///
+/// `min`/`max`/`clamp` are absent on purpose: they take 2/3 arguments and
+/// `parse_atom` desugars them to `Expression::Conditional`, so they never
+/// reach `UnaryFn`. `present(x)` is a *condition*, parsed by `parse_cond_atom`.
+pub(crate) const SUPPORTED_UNARY_FNS: &[&str] = &[
+    "exp",
+    "log",
+    "ln",
+    "sqrt",
+    "abs",
+    "floor",
+    "ceil",
+    "round",
+    "logit",
+    "inv_logit",
+    "expit",
+];
+
+/// Panic message for the (parser-guaranteed unreachable) tail of each
+/// `UnaryFn` match. Named so all three consumers report the same failure.
+const UNARY_FN_ARM_MISSING: &str =
+    "UnaryFn carries a name outside SUPPORTED_UNARY_FNS — parse_atom is the only \
+     constructor and rejects unknown names (#1332); a name added to the whitelist \
+     needs an arm in eval_expr, compile_expr_into and differentiate_with_chain";
+
+/// The human-facing catalogue of callable names, used by the parse-time
+/// rejection. Kept next to `SUPPORTED_UNARY_FNS` so a new builtin is one edit
+/// away from being advertised, and includes the multi-argument forms the
+/// parser desugars plus the conditions-only `present`.
+fn supported_function_list() -> String {
+    format!(
+        "{}, min(a, b), max(a, b), clamp(x, lo, hi), present(x) (conditions only)",
+        SUPPORTED_UNARY_FNS.join(", ")
+    )
+}
+
 #[derive(Debug, Clone)]
 pub(crate) enum Expression {
     Literal(f64),
@@ -18463,7 +18515,8 @@ fn eval_expr<E: EvalEnv>(
                     let clamped = v.clamp(1e-15, 1.0 - 1e-15);
                     (clamped / (1.0 - clamped)).ln()
                 }
-                _ => v,
+                // Unreachable: `parse_atom` admits only `SUPPORTED_UNARY_FNS`.
+                _ => unreachable!("{UNARY_FN_ARM_MISSING} (eval_expr, `{name}`)"),
             }
         }
         Expression::Power(base, exp) => {
@@ -18901,9 +18954,8 @@ fn compile_expr_into(bc: &mut Bytecode, expr: &Expression) {
         Expression::UnaryFn(name, arg) => {
             compile_expr_into(bc, arg);
             // Names matched here mirror `eval_expression_indexed`'s UnaryFn
-            // dispatch. Anything else becomes a no-op (the slow path returns
-            // the argument unchanged); preserve that with `Op::Abs` of `Abs`
-            // we can't — fall through to push the value as-is.
+            // dispatch, and the match is total over `SUPPORTED_UNARY_FNS`
+            // because `parse_atom` admits nothing else (#1332).
             match name.as_str() {
                 "exp" => bc.ops.push(Op::Exp),
                 "log" | "ln" => bc.ops.push(Op::Ln),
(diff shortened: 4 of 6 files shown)


roninsightrx (commented):
Reviewed commit a596430d. I found no blocking correctness issue.

The shared parse-time guard closes the silent-identity path across the generic expression parser, and the current internal UnaryFn constructors use fixed whitelisted names. I also checked the special derived aggregate forms and ODE input-rate syntaxes; they are intercepted before this generic path. The floor/ceil/round derivative convention is correct away from discontinuities, and the value/bytecode/derivative parity coverage is strong.

Local verification: cargo test --lib completed with 4,388 passed, 23 expected slow tests ignored, and 0 failed. At review time Check, Clippy, Format, Rustdoc, Docs lint, Public API baseline, and release-semantics tests were green; the two coverage jobs were still running.

Two non-blocking hardening notes:

1. Unsupported multi-argument names still hit the comma/arity branch before the new whitelist check. For example, pow(x, y) reports that pow “takes 1 argument” rather than the advertised unknown-function diagnostic/list. That does remain a hard parse error, so it does not preserve the dangerous behavior fixed here, but pow is one of the realistic migration triggers named in #1332. Consider rejecting an unknown name immediately after lower-casing/classifying min/max/clamp, plus adding a two-argument unknown-name test.

2. The test described as requiring the probe table to cover SUPPORTED_UNARY_FNS “exactly” checks equal lengths and const-to-probe membership, but not the reverse direction or uniqueness. Replacing one const entry with a duplicate of another can therefore leave the tests green while rejecting the omitted supported function. A probe-to-const assertion (and/or uniqueness assertion) would make the claimed drift guard exact.

Neither note blocks this fix.

roninsightrx:
Both notes were right and both are fixed in `feb441f4`. Thanks — the first one in particular was a real gap in the diagnostic, not just wording.

### 1. Unknown name judged before arity

Confirmed exactly as described: `pow(x, y)` fell into the comma branch and reported

```
function `pow` takes 1 argument, but a `,` was found — only `min(a, b)` and
`max(a, b)` take two, and `clamp(x, lo, hi)` takes three.
```

Hard error, so never the silent identity this PR is about — but it sends the reader looking for an arity bug in a function ferx does not have, and `pow` is one of the migration triggers #1332 names.

The guard now sits **above the argument loop**, so the *name* is judged before the *arity*, and I deleted the copy that sat next to the `UnaryFn` return rather than keeping both. Two gates rejecting the same inputs would cover for each other — deleting either would have left the suite green, which is the failure shape CLAUDE.md calls a test hole rather than belt-and-braces. `min`/`max`/`clamp` are exempted by name (they are desugared and never reach `UnaryFn`).

New test `an_unknown_name_is_rejected_at_every_arity` covers 0, 2 and 3 arguments, asserts the message does *not* contain an arity claim, and asserts `min(x)` / `max(x)` / `clamp(x)` still get their own arity messages so moving the guard can't have swallowed them. Restoring the old placement reproduces your exact output and reddens the test:

```
`pow(TVCL, 2.0)`: expected the unknown-function diagnostic, got: function `pow`
takes 1 argument, but a `,` was found — ...
test result: FAILED. 817 passed; 1 failed
```

### 2. Probe-table drift guard now exact

Also correct, and the claim in my own doc comment ("forced to cover the const exactly") was the part that was wrong. I ran your mutation — replace `"ceil"` with a second `"exp"` — and confirmed it would have been green under the length-plus-membership pair.

I went with **one comparison of sorted vectors** rather than adding reverse-membership and uniqueness assertions alongside the existing ones. Adding those would have caught it, but uniqueness and probe-to-const reject the same input on this mutation, so it reintroduces the same two-gates problem as note 1. Sorted-vector equality is exact in both directions *and* catches a duplicate row on either side, in a single gate:

```
assertion `left == right` failed: the probe table and SUPPORTED_UNARY_FNS have drifted
  left:  ["abs", "ceil", "exp", ...]
 right:  ["abs", "exp",  "exp", ...]
```

Verified against both your mutation and the complementary one (a name added to the const with no probe).

### Verification on `feb441f4`

`cargo test --lib` → 4389 passed, 0 failed, 23 ignored. `tools/preflight.sh` → `preflight OK — fmt check clippy rustdoc docs public-api`. The 183 integration binaries were green on the previous commit and neither change touches them; CI is re-running all of it.

Thanks also for independently checking the derived-aggregate and ODE input-rate forms — those intercept before the generic path, which matches what I found, and it's useful to have that confirmed from a second read rather than only from mine.

PR #1340, opened 2026-09-11
feat(sens): analytic FOCE/FOCEI outer gradient for [covariate_nn] + IOV [closes #1339]

### LINKED ISSUE #1339: perf(nn): [covariate_nn] + IOV has no analytic outer gradient — iov_sens_supported declines every DCM, forcing BOBYQA / FD-reconverge over the weight block
## Summary

A `[covariate_nn]` model with IOV (`kappa` + `iov_column`) has **no analytic outer gradient**: `iov_sens_supported` is the strict predicate (`iov_analytical_supported_core(model, require_theta_axis = true)`), whose θ-axis clause requires the `[individual_parameters]` program to seed every `model.n_theta` axis. A network's weight thetas are auto-generated and not referenceable from the program, so the clause can never hold for a DCM. Consequences, measured on `ferx-testdata/busulfan_dominic/ferx/dcm/dcm_h2_iov_s600.ferx` (600 subjects, 4→2→4 network, 22 weights, 2 κ):

| | analytic IOV base (`base_s600.ferx`) | DCM h2 + IOV |
|---|---|---|
| outer optimizer under `auto` | NLopt L-BFGS (14 coords) | **BOBYQA (32 coords)** |
| outer gradient | analytic `Dual2` IOV walk | none (derivative-free); FD-reconverge if a gradient optimizer is forced |
| evals to termination | 41 | 546 |
| final OFV (after #1327 / PR #1337) | 56 915.80 | 59 470.18 |
| per-eval cost reported in #1327 | ~0.7 s (DCM, no IOV) / ~5 s (base) | ~19 s |

The inner loop is already served: #1015 relaxed the **η-only** IOV walk (`iov_sens_eta_supported` / `iov_eta_only_derivs_dyn`, `ModelNnGuard` lifting the outputs as constants) so every DCM+IOV subject gets an exact `∂/∂[η, κ]`. The outer `∂/∂θ` for the weights is what is missing on the IOV path.

This is the residual gap left after #1327: that issue's phantom (trajectory 7k–15k below the cold re-evaluation) was the IOV inner-loop fallback and is fixed in #1337, but the fixed fit still lands 2 554 above the analytic base it can represent, and does so on a derivative-free search over 32 coordinates.

## Where the pieces are

- `sens/provider.rs` — `iov_sens_supported` → `iov_sens_supported_core(model, true)` → `iov_analytical_supported_core`'s `n_theta_axis() == model.n_theta` clause is the decline. `build_iov_sources(.., eta_only)` already has the two-arm seed (`iov_combined_derivs_dyn` vs `iov_eta_only_derivs_dyn`); the outer `run_obs_iov` walk consumes `CombinedDerivs::{dtheta, d2eta, d2eta_theta}`, which the η-only builder leaves empty by contract.
- `estimation/outer_optimizer.rs` — `population_gradient_dispatch` gates on `iov_sens_supported(model)`; out of scope falls to `subject_reconverged_fd_gradient_iov` (one inner re-solve per θ perturbation, ×2 central, ×32 coordinates).
- The non-IOV DCM route to copy: #1300 pairs `ModelNnGuard` with `ModelNnAxisGuard` so each network output `z` sits on its own dual axis, the program walk yields `∂p/∂z` and `∂²p/∂η∂z`, and `nn_param_derivatives_at_cov` chains those to the weight thetas through the network's backprop Jacobian (`sens/provider.rs` ~2900–3070, `run_obs_tvcov`'s chunked θ-column walk). The IOV twin would seed `(declared θ, z-outputs, η_bsv, κ)` per occasion source instead of `(θ, η_bsv, κ)`, and chain the `z` columns to the weights after the walk — the dual width becomes `n_declared_theta + n_z + n_eff` (here 0 + 4 + 5 = 9) instead of `n_theta + n_eff` (27, past the `1..=24` dispatch anyway).

## What a fix needs to show

- `iov_sens_supported` true for a DCM+IOV model, `auto` resolving to the gradient optimizer it picks for the analytic base, and the `FitResult` FD-fallback warning silent (every fixture must be checked for which engine it actually ran on — #1210).
- `Dual2`-vs-FD parity of the packed outer gradient on a **multi-occasion, multi-dose** DCM+IOV subject across the weight block, Ω, Ω_IOV and σ (the `check_full_provider_vs_fd` pattern), plus a mutation on the `z`→weight chain.
- The busulfan reproduction landing at or below the analytic IOV base (56 915.80) on training OFV, with the per-eval cost in line with the non-IOV DCM.

Related: #1327 / #1337 (inner-loop fallback, fixed), #1300 (TV-cov analytic walk for NN weights), #1015 (η-only IOV scope), #1310 (DCM tuning, where this dataset is a calibration point).

roninsightrx:
## Why

#1339: a `[covariate_nn]` model with IOV had **no analytic outer gradient**. `iov_sens_supported` is the strict predicate whose θ-axis clause required the `[individual_parameters]` program to seed every `model.n_theta` axis; a network's weight thetas are generated and never referenceable from the program, so no DCM could pass it. `auto` therefore resolved to derivative-free BOBYQA over every weight, and an explicit gradient optimizer got reconverged finite differences. On the busulfan reference (`ferx-testdata/busulfan_dominic/ferx/dcm/dcm_h2_iov_s600.ferx`, 600 subjects, 22 weights, κ on CL and V1) that was 546 BOBYQA evaluations landing at 59 470 (after #1337) against 41 L-BFGS evaluations to 56 916 for the analytic covariate model the network replaces.

The inner (per-subject) gradient was already served by the η-only walk (#1015); the outer `∂/∂θ` over the weight block is what this PR adds.

## What changed

Was stacked on #1337, which is now merged; this PR is the single commit on top of `main`.

- `src/sens/provider.rs`
  - `iov_nn_combined_derivs_dyn` — the IOV twin of #1300's `nn_param_derivatives_at_cov`: evaluates the program over `Dual2<M>`, `M = n_declared_θ + n_eff + n_z`, with the declared θ, the combined `[η_bsv, κ]`, and one axis per network output (`ModelNnGuard` + `ModelNnAxisGuard`), then chains each weight column `∂p/∂w = Σ_k ∂p/∂z_k · ∂z_k/∂w` and `∂²p/∂[η,κ]∂w` likewise through the network's backprop Jacobian. On the reference DCM that is 9 axes rather than the 27 the old `n_theta + n_eff` seeding would have needed.
  - `build_iov_sources` routes a DCM's outer source to it; the η-only inner builder is unchanged, and `subject_eta_grad_iov_analytical` now pins `eta_only` for a DCM so the inner keeps the `Dual1<n_eff>` route it took before.
  - `run_obs_iov` takes a θ-column chunk (`theta_cols`) and an accumulator (`into`), exactly as `run_obs_tvcov` does since #1300: the chunk's θ columns sit on axes `0..nc`, the stacked block on `nc..M`; the scatter writes the chunk's columns into the accumulated `SubjectSens`. `subject_sensitivities_iov` walks `0..n_theta` in chunks of `MAX_TVCOV_AXES − n_stacked`. A model whose `n_theta + n_stacked` already fit the cap is one chunk — the previous walk verbatim. The three post-walk steps that seed direct θ/η references on absolute axes (Form C readout, `[initial_conditions]` impulse, `ExpressionScale` quotient) decline a non-identity chunk (`None` → FD), exactly as the over-width walk declined before chunking existed.
  - `iov_analytical_supported_core`'s outer clause is now `prog.n_theta_axis() == n_program_theta(model)` (declared θ), plus for a DCM the output-chain predicate `tvcov_analytical_supported` / `nn_theta_gradient` already use and the builder's own dispatch bound. Non-NN models: `n_program_theta == n_theta`, unchanged.
- Tests (`nn_theta_gradient_tests.rs`): `dcm_iov_outer_sensitivities_are_analytic_and_match_fd_of_predict_iov` (all four jet blocks vs central FD of `predict_iov` on a 2-cpt infusion, 3-occasion, 22-weight fixture that forces two chunks; every weight column asserted live), `dcm_iov_model_gets_the_analytic_outer_gradient_and_a_gradient_optimizer` (strict predicate, `analytic_outer_gradient_available`, `Optimizer::Auto → NloptLbfgs`, and the direct-weight-reference decline). The two tests that pinned the strict predicate *false* for a DCM are flipped with the reason.
- `docs/model-file/covariate-nn.qmd`: an IOV paragraph in the FOCE/FOCEI gradient section. `CHANGELOG.md` entry under Changed.

## Alternatives considered

- **Walk over PK-slot axes per occasion and chain θ post-walk** (how the static superposition path keeps θ out of its dual width). Would make the walk cost independent of the weight count, but the IOV walk composes the whole `(θ, η, κ)` chain in one jet including the mixed second-order block; re-deriving that as a post-walk product is a second implementation of the same formula. Chunking reuses the walk and the #1300 pattern unchanged.
- **Log-normal fallback with an FD `tv_fn` Jacobian** (the static non-IOV DCM route). Approximate in θ where the z-axis chain is exact, and `tv_fn` is at η = 0, which the IOV builder does not evaluate at.

## Cross-repo dependency

## Breaking changes
- [x] None (no public signature changes; `nn_output_chain_supported` widened to `pub(crate)` for the test)

## Numerical validation

**Provider parity** (`Dual2` vs central FD of `predict_iov`, fixture above, 6 observations × 22 θ × 9 stacked): `∂f/∂θ` and `∂f/∂[η,κ]` within `2e-4` relative, `∂²f/∂[η,κ]²` and `∂²f/∂[η,κ]∂θ` within `3e-3` (the tolerances `check_iov_provider_vs_fd` uses for every IOV fixture). Three mutations each turn it red: zeroing the `∂z/∂w` chain (`returned 0, FD ≠ 0` on every weight column), scattering the second chunk onto columns `0..7`, and a merge that drops the first chunk.

**End-to-end**, same model/data as #1327/#1337, 4 threads, release with `nn`:

| run | optimizer | outer gradient | evals | best-seen | cold final |
|---|---|---|---|---|---|
| analytic IOV base | L-BFGS (14) | analytic | 41 | 56 915.799 | 56 915.799 |
| DCM h2 + IOV, #1337 | BOBYQA (32) | none | 546 | 59 470.182 | 59 470.182 |
| DCM h2 + IOV, this PR, defaults | L-BFGS (32) | analytic | 104 | 58 849.572 | 58 849.572 (53 s, `NLopt stopped: Failure`) |
| DCM h2 + IOV, this PR, `optimizer = slsqp` | SLSQP (32) | analytic | 142 | 58 741.913 | 58 780.104 (68 s) |
| DCM h2 + IOV, this PR, `max_unconverged_frac = 0.5` | L-BFGS (32) | analytic | 1288 | **56 668.853** | **56 668.853** (882 s, `Converged: YES`) |

The gradient is now analytic and `auto` picks L-BFGS; the reported `gradient:` line and the absence of the FD-fallback warning confirm the route on all 600 subjects. What the runs also show, and what this PR does **not** change:

- **The default fit stops early.** The first L-BFGS step inflates Ω_V1 from 0.038 to 1.33 (the residuals demand IIV while the network still carries no covariate effect), and the fit settles into an Ω_V1 ≈ 3–7 basin in which 3–9 % of subjects report an unconverged inner solve per eval. Line-search probes that cross the 10 % EBE guard (`max_unconverged_frac`) return the guard penalty, so L-BFGS's line search fails at eval 104 while still descending ~30 units per eval. SLSQP walks further but shows a 38-unit warm/cold gap at its best point — in that Ω region cold inner solves still fail for some subjects even with #1337 (BFGS from η = 0 does not reach an η_V1 mode 1.8 SD away in 200 iterations, and NM does not either).
- **With the guard loosened the fit escapes the basin** (Ω_V1 3.6 → 0.014 between evals 200 and 300), converges, and lands 247 below the analytic IOV base with Ω, Ω_IOV and σ at base-like values (η_CL 0.057, η_V1 0.026, η_V2 0.019; κ_CL 0.009, κ_V1 0.021; σ 0.050 / 9.7). Warm and cold agree to 1e-6.

So the answer to #1327's "landing at or below the analytic IOV base" is yes, with the analytic gradient plus a loosened guard; the Ω-inflation start and the inner fragility at large Ω are a separate optimisation-path problem (see Open questions).

**Holdout** (`data/test.csv`, 303 subjects / 2 411 obs, eval-only with cold EBEs, scorer built on this branch so it carries #1337):

| fit | train OFV (600 subj) | test −2LL (303 subj) |
|---|---|---|
| analytic IOV base | 56 915.80 | 26 421.96 |
| DCM h2 + IOV, guard 0.5 | 56 668.85 | 26 424.29 |

The network's 247-unit training advantage does not carry to the test set: the two are tied within 2.3 units. (The pre-#1337 scorer reports 26 857 for the same DCM bundle — the cold-inner defect, 433 units of it, on the holdout side.)

- [x] Compared against: prior ferx commit `96683307` (#1337 head) on the same model/data
- [x] `Dual2`-vs-FD parity added in the same module

## Performance
- [x] Faster on the affected models — the reference DCM+IOV fit goes from 546 derivative-free evaluations (102 s) to 104 L-BFGS gradient evaluations (53 s) before its line search fails, and to a converged 1288-eval run (882 s) with the guard loosened: ~0.7 s per gradient evaluation on 600 subjects at 4 threads, against the ~19 s per evaluation #1327 reported for the FD-reconverge route. Non-NN IOV models take exactly one chunk (unchanged numerics and cost); an IOV model whose `n_theta + n_stacked` exceeded 24 and previously fell to FD now walks in chunks.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib --features nn`: 4492 passed)
- [x] Regression test added that fails without this fix (predicate/auto test fails on the strict clause; parity test fails on each of the three mutations above)

- [x] Not applicable (no new API surface; behaviour change documented on the existing page)

## Docs
- [x] `docs/model-file/covariate-nn.qmd` updated (IOV paragraph), `docs-lint --check` clean

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — the new builder is `Dual2` over the generic program evaluator; the walk is unchanged
- [x] No version bump

## Docs & examples
- [x] No example execution step needed (no new example; reference run reported above)

## Reviewer hints

`run_obs_iov`'s `seed` closure is where the chunk layout lives (`grad[c] = cd.dtheta[i][theta_cols[c]]`, stacked block at `nc`); the scatter and the merge at the end are the other two places that know about `theta_cols`. Everything between them is untouched. The `identity_chunk` gate at the top is what keeps the readout/init/expression-scale programs' absolute-axis seeding correct — they only ever run on the `0..n_theta` chunk.

## Open questions

- **The default DCM+IOV fit still stops early** (`NLopt stopped: Failure` at eval 104, see the table). Two things compound there and neither is this PR's scope: the first L-BFGS step inflates Ω_V1 ~35× because the network starts with no covariate effect, and once Ω is that wide the inner solve fails for 3–9 % of subjects per eval, so line-search probes trip the 10 % EBE guard and L-BFGS quits. Loosening the guard to 50 % lets the fit escape and converge below the base. Candidates for a follow-up: a two-stage start (Ω/κ/σ held at their inits until the weights have moved — there is currently no way to warm-restart a DCM's weights from the CLI, `init` only sets the output biases), a guard that returns the clean objective for a probe rather than the penalty when the accepted point itself carries unconverged subjects, or a more robust cold inner solve at wide Ω (`inner_maxiter`, an Ω-scaled BFGS seed).
- Covariance for a DCM+IOV model now takes the analytic assembly (`covariance_sensitivities` gates on `iov_analytical_supported`), as it already did for a non-IOV DCM. Network-weight SEs remain documented as unreliable; if a reviewer prefers the FD covariance for DCMs, that is a one-clause gate.


src/sens/provider.rs
```diff
@@ -2119,32 +2304,45 @@ fn run_obs_iov<const M: usize>(
...
     use crate::sens::propagate::{event_driven_sens_with_doses_g, PkDual};
 
+    // Chunk width: the stacked block sits right after this chunk's θ columns.
+    let nc = theta_cols.len();
+    debug_assert_eq!(M, nc + n_stacked);
+    // The identity chunk (`0..n_theta`) is the only layout on which the readout / init /
+    // expression-scale programs' absolute-axis seeding (`θ_m → m`, `η_k → n_theta + k`)
+    // is correct; every other chunk must not run them (#1339).
+    let identity_chunk = nc == n_theta && theta_cols.iter().enumerate().all(|(c, &m)| c == m);
+    if !identity_chunk && (readout.is_some() || bsv_amount.is_some() || scale_groups.is_some()) {
```
roninsightrx: [P2] Keep the model-level analytic gate aligned with this decline

`iov_analytical_supported` now returns true for an output-chain-supported DCM+IOV model, which makes `analytic_outer_gradient_available` report analytic and makes `Optimizer::Auto` select L-BFGS. However, this branch returns `None` for every subject whenever theta chunking is required and the model has a Form-C readout, analytic initial condition, or expression scale. I reproduced this with the new 22-weight/3-occasion fixture plus `init(central) = V1`: `iov_analytical_supported(&model)` is true while `subject_sensitivities_iov(...)` is `None`. The fit therefore advertises an analytic outer gradient and chooses a gradient optimizer, but computes every outer derivative by reconverged FD—the expensive behavior this PR is meant to remove. Please either make these post-walk features chunk-aware or conservatively keep such DCMs out of the model-level outer gate, with a route-consistency regression.

roninsightrx:
Validation requirement: AGENTS.md requires every new feature or numerical behavior to include a comparison with equivalent NONMEM output in the feature docs or PR description. The current Numerical validation section compares the new provider with internal central FD and compares end-to-end fits with an earlier ferx commit / analytic ferx model, but it does not include NONMEM output. Please add the required NONMEM comparison (including the model/output reference and tolerances) before merge.

roninsightrx:
**P2 — agreed, fixed in 1d010527.** A DCM carrying a Form C readout, an `[initial_conditions]` impulse or an `ExpressionScale` quotient whose weight block cannot fit the walk alongside even one occasion's stacked axes (`n_theta + n_eta + n_kappa > 24`) now declines at the model level in `iov_analytical_supported_core`, so `analytic_outer_gradient_available` is false and `auto` stays on BOBYQA — the route such a DCM had before this PR. Route consistency is pinned by `dcm_iov_gate_and_walk_agree_on_absolute_axis_features`: on the plain 22-weight / 3-occasion fixture the gate is true and `subject_sensitivities_iov` serves; on the same fixture plus `init(central) = 0.1 * V1` the walk returns `None` and the gate, `analytic_outer_gradient_available` and `resolve_auto` all say FD/BOBYQA, while the η-only inner scope still serves it. Deleting the gate turns the test red on the gate assertion.

A DCM with one of those features that *does* fit one occasion still declines per-subject for wider-K subjects, reported by the existing FD-fallback warning — the same per-subject mechanism every other IOV decline uses (`moving_bounds_separable`, SS + lagtime). Making the three post-walk steps chunk-aware (remapping their absolute θ/η axis seeding onto a chunk) is the way to lift that; I left it out of this PR since none of the reference DCMs carry those features.

**NONMEM comparison.** This PR adds no quantity that has a NONMEM output to compare against, and I don't think one should be manufactured for it:

- The change is to the *gradient* of the FOCE/FOCEI objective, not to the objective, the estimates, or any diagnostic. NONMEM does not expose its gradient, and the repository rule for exactly this class of change (CLAUDE.md, *Analytic Sensitivities*) is the `Dual2`-vs-FD parity test against the production predictor — which this PR carries on all four jet blocks over the whole weight block, with three mutations verified red.
- The objective the optimizer minimises is untouched: every model that fit the 24-axis walk before takes the identical single-chunk path (same dual width, same seeding, same scatter), and the existing IOV NONMEM anchors in `tests/` (`kappa_weight_nonmem`, `transit_iov_nonmem_anchor`, `mixture_iov_nonmem`, `dose_form_lag_nonmem_anchor`) pass unchanged on this branch, run to convergence with `--features nn,slow-tests`: `weighted_kappa_matches_the_nonmem_hand_written_twin`, `the_anchor_fit_reports_the_arm_size`, `iov_mixture_fit_matches_nonmem`, `transit_iov_objective_matches_nonmem`, and the 8 fast `dose_form_lag_nonmem_anchor` cases — 12 passed, 0 failed.
- NONMEM cannot express a `[covariate_nn]` model, so the DCM itself has no NONMEM twin; the reference it is measured against in the PR table is ferx's analytic IOV base, whose closed-form IOV objective is the thing those anchors pin.

If a NONMEM number is wanted on this dataset specifically, the right one is a NONMEM fit of `base_s600.ferx` on `train_s600.csv` (the 1 518-subject `nm/busulfan_ffm_iov.lst` in the testdata repo is the FTIME/ODE model on the full dataset, not this split) — that would anchor the 56 915.80 the DCM is compared to, and it belongs with the DCM evaluation (#1310), not with the gradient.

Issue #285, opened 2026-06-12
Proposal: require AD↔FD parity tests for gradient-path changes (CLAUDE.md rule + #281 coverage checklist)


TeunP:
## Problem

The test-tier taxonomy in `CLAUDE.md` classifies tests by **speed/scope** (Tier 1 unit / Tier 2 integration / Tier 3 slow-convergence) but is silent on the **FD-vs-AD build axis**. The default build is `default = ["autodiff"]` (Enzyme), yet every CI job builds FD-only (`--no-default-features --features ci`) — so the autodiff path is never compiled or run in the default PR job. A contributor following `CLAUDE.md` today picks a tier and never has to consider whether their change *also* needs to run under autodiff.

That blind spot is exactly how #278 shipped: the autodiff FOCEI-INTER marginal silently diverged from the FD reference and descended to a spurious low-OFV optimum. #280 gates the known-bad paths to FD (closes #278) and #281 proposes the CI **infrastructure** to actually run the Enzyme path — but neither encodes the **authoring discipline** that every gradient-path change must carry an AD↔FD parity test. Without the rule, the runner from #281 has nothing reliably feeding it (today `tests/autodiff_fd_consistency.rs` is *armed but unplugged*).

This issue is a place to **read and evaluate** two complementary process changes before they land.

## Proposal A — encode the rule in `CLAUDE.md`

Add a third parallel rule to the **Tests** section, immediately after *"Every new feature requires a comparison with NONMEM output"*:

```markdown
**Every change to a gradient, marginal, or likelihood path requires an AD↔FD
parity test.** The default build is `default = ["autodiff"]` (Enzyme), but every
CI job builds FD-only (`--no-default-features --features ci`) — so the autodiff
path is *not* exercised in the default PR job, and a wrong AD gradient can ship
silently (this is how #278 happened). When you touch inner-loop gradients, the
FOCE/FOCEI marginal, error-model derivatives, or anything reachable from
`individual_nll_ad` / `predict_all_ad` (`src/ad/ad_gradients.rs`), add a case to
`tests/autodiff_fd_consistency.rs` asserting the AD and FD builds agree on
objective and estimates to tolerance. These tests are gated
`#[cfg(feature = "autodiff")]` and run only under the Enzyme toolchain — locally
today, in CI once #281 lands. If a model can't yet be differentiated safely,
route it to FD via `analytical_ad_unsupported`
[...]

PR #1341, opened 2026-09-11
feat(estimation): per-parameter priors for penalized ML / MAP estimation (#254)

### LINKED ISSUE #254: feat: Add (a simple and understandable) prior support for frequentist/MAP estimation (nonmem $PRIOR analog)
## Summary

Add an optional, per-parameter prior that adds a penalty to the objective,
giving penalized maximum likelihood (MAP) point estimates. The design priority
is usability: make this far simpler to specify than NONMEM's $PRIOR. Every prior
is just a central value plus a relative uncertainty (RSE%), declared inline next
to the parameter. No separate prior problem, no precision or covariance
matrices, no inverse-Wishart, no degrees-of-freedom indexing.

## Motivation

- Sparse clinical data is the driver: rare disease, pediatrics, TDM-led fits.
  The likelihood is weakly informative, IIV terms collapse to boundaries, and
  structural parameters drift. A prior from a published popPK model stabilizes
  the fit and lets a small local dataset borrow strength from existing knowledge.
- It enables model updating: take a literature model as the prior, refit on
  local sparse data, get a regularized estimate. This is the MAP-Bayesian
  updating behind much of model-informed precision dosing.
- It's cheap on this architecture. The penalty is a differentiable term added to
  the objective, so Enzyme propagates its gradient automatically, and it
  composes with every estimation method (FOCE/AGQ/SAEM/IMP/QRPEM) because it only
  augments the objective.
- This issue is explicitly about ergonomics. NONMEM's $PRIOR works but is a known
  pain: you write a separate prior problem, supply a variance block for the
  thetas, an inverse-Wishart on Omega via a prior matrix plus degrees of freedom,
  and keep all the matrix indexing aligned by hand. We want the opposite. Read a
  number and its RSE off a paper, type it next to the parameter, done.

## Design principle: one mental model

Every prior is (central value, relative uncertainty as RSE%).

- theta: normal prior on the estimation scale, centered at the value, SD from the
  RSE.
- omega diagonal variance: prior on the log-variance scale, centered at the
  value, spread from the RSE.
- sigma diagonal: same as omega.

That single concept is the whole user-facing surface.

## Proposed approach

Augment the objective:

    OFV_total = OFV_data + sum_of_penalties

theta penalty, for a priored fixed effect with prior mean m and prior SD s
(s derived from the RSE):

    pen = ((theta - m) / s)^2

If the parameter is estimated on the log scale (log-transform or mu-referencing),
apply the prior on that scale and convert a natural-scale RSE to a log-scale SD
internally via s = sqrt(log(1 + RSE^2)), so the user can paste the RSE straight
from a parameter table without doing any conversion.

omega diagonal penalty, for variance w with prior value v and log-scale spread
tau (tau derived from the RSE on the variance):

    pen = ((log(w) - log(v)) / tau)^2

Working on the log-variance scale keeps it positive and symmetric and avoids any
matrix algebra.

sigma: same pattern, optional.

Unpriored parameters get no penalty (flat prior). Priors can cover any subset.

Implementation:
- ferx-r: per-parameter prior fields. Minimal form, a prior(value=..., rse=...)
  attached to a theta, and prior(value=..., rse=...) attached to an omega
  diagonal element.
- Do the RSE-to-SD and natural-to-log conversions internally. The user never
  computes a standard deviation or touches a matrix.
- Add the penalty inside the objective so Enzyme differentiates it with the rest.
  No bespoke gradient code.
- Import-from-fit helper: given a previous ferx fit, auto-fill value = estimate
  and rse = SE/estimate for thetas and omega diagonals. This is the model-updating
  path and should be a single call.
- Reporting: split the OFV into data and prior contributions, and per priored
  parameter print the prior value, the final estimate, and the shift toward the
  prior.

## Non-goals for v1 (kept out on purpose to stay simple)

- Diagonal Omega and Sigma only. No block or off-diagonal prior, since that is
  exactly what forces the inverse-Wishart machinery. Revisit only if there's real
  demand.
- Independent priors per parameter. No joint prior covariance among thetas, so no
  precision matrices anywhere in v1.
- Not full Bayesian. This is MAP / penalized ML. Reported SEs are the curvature
  of the penalized objective, not a posterior. Name it clearly in the docs and
  output.

## Open questions

- Field naming, and whether RSE is entered as a percent or a fraction. Pick one
  and validate the input.
- Whether to also accept an absolute SD as an alternative to RSE for users who
  prefer it.
- Optional "equivalent prior subjects" strength as a friendlier alias for the
  omega tau, converted internally. Sugar, not required for v1.
- Clear error when a user tries to put a prior on an off-diagonal/block Omega
  element in v1.

## Acceptance criteria

- [ ] Per-parameter normal prior on theta and log-scale prior on diagonal omega,
      specified inline from ferx-r as (value, rse).
- [ ] RSE-to-SD and natural-to-log conversions handled internally; the user never
      supplies a matrix or a standard deviation.
- [ ] On a sparse simulated dataset where unpenalized FOCE hits an Omega boundary
      or fails to converge, the prior yields stable, sensible estimates.
- [ ] A very weak prior reproduces the unpriored MLE.
- [ ] Very sparse data shrinks estimates toward the prior value.
- [ ] Data and prior contributions to the OFV are reported separately, with a
      per-parameter shift-toward-prior summary.
- [ ] Import-prior-from-previous-fit works end to end in one call.
- [ ] Composes with at least FOCE and one EM method.
- [ ] Docs page plus a model-updating worked example (literature model + sparse
      local data).

## References

- Gisleskog, Karlsson, Beal. Use of prior information to stabilize a population
  data analysis. J Pharmacokinet Pharmacodyn 29(5):473-505, 2002. The canonical
  frequentist-prior reference and exactly the sparse-data clinical use case.
- NONMEM $PRIOR (NWPRI, TNPRI), NONMEM 7 guides. Listed as the implementation we
  are deliberately simplifying away from.

roninsightrx:
## Why

Sparse clinical data — rare disease, pediatrics, TDM-led fits — leaves the
likelihood weakly informative: IIV terms collapse onto boundaries and structural
parameters drift. A prior from a published popPK model stabilizes the fit and
lets a small local dataset borrow strength from what is already known. The same
mechanism is model updating (literature model as prior + local sparse data →
regularized estimate), which is the MAP-Bayesian updating behind much of
model-informed precision dosing.

NONMEM's `$PRIOR` does this and is a known source of friction: a separate prior
problem, a variance block for the thetas, an inverse-Wishart on Ω via a prior
matrix plus degrees of freedom, and all the matrix indexing kept aligned by hand.
#254 asks for the opposite — read a number and its RSE off a paper, type it next
to the parameter, done.

One correction to the issue text: it assumes "Enzyme propagates its gradient
automatically". Enzyme was retired in #285/#281 in favour of the hand-rolled
`Dual2` sensitivities. That turns out to make this *easier*, not harder — see
below.

This is **Phase 1**. Three of the issue's eight acceptance criteria are
deliberately not in it; they are listed under Open questions.

## What changed

`prior(value, rse = 25%)` as a trailing modifier on any `theta` / `omega` /
`sigma` / `kappa` declaration:

```toml
[parameters]
  theta TVCL(0.2, 0.001, 10.0)  prior(0.15, rse = 25%)
  omega ETA_CL ~ 0.09           prior(0.09, rse = 40%)
```

**The penalty lives in the optimizer's packed coordinates** (`src/estimation/priors.rs`):

```
OFV_total = OFV_data + Σ ((x − m)/s)²     ∂/∂x = 2(x−m)/s²     ∂²/∂x² = 2/s²
```

Three things follow, and they are why this is ~350 lines rather than a new
sensitivity kernel:

- gradient and Hessian are exact and closed form, and the Hessian is a
  **constant** diagonal. Nothing routes through `sens/`, so no `Dual2`
  instantiation and no added monomorphization on a compile path that is already
  93% LLVM (#971);
- the prior *family* falls out of the packing rather than being a second
  convention. A θ with a non-negative lower bound is estimated as `ln θ`, so a
  normal prior there **is** a lognormal prior on θ — exactly what the issue asks
  for. An Ω diagonal is estimated as `ln(SD)`, so a normal prior there is a
  lognormal prior on the variance: positive, symmetric, no matrix algebra;
- the outer objective, GN, trust-region and the covariance step already work in
  packed space, so there is one form of the penalty rather than one per caller.

**Scale convention:** the prior's value and spread are always on the scale the
parameter was declared on — a variance for a bare `omega X ~ v`, an SD under
`(sd)`. Both come off the same row of the same table, and the user never
converts anything.

Wired into `outer_optimizer` (objective + gradient, NLopt and built-in BFGS),
`gauss_newton` (gradient + BHHH Hessian), `trust_region` (objective + gradient +
Hessian), and `covariance` (Hessian). Parsed by peeling the `prior(...)` tail
**before** the four declaration regexes run, so none of them changed — the same
mechanism `weight =` uses, and for the same reason (they are all unanchored, so a
modifier left on the line is silently dropped).

Reported in the console summary and the fit YAML: the OFV split plus, per priored
parameter, the prior, the estimate, the shift in prior SDs, the penalty, the
realised family, and the implied 95% prior interval.

## Alternatives considered

**Prior scale decoupled from the packing** (always lognormal for positive-support
parameters, chain-ruled into packed space). Removes the bound-coupling trap in
`Open questions`, but costs a non-constant Hessian, adds a second convention, and
contradicts the issue's own rule. Rejected.

**Priors on `FitOptions` rather than `CompiledModel`**, which is where the
covariate-NN regularizer's `nn_l2` lives and would have cost 0 struct-literal
edits instead of 29. Rejected: a prior changes the *estimates*, so a caller who
parses a `.ferx` file and hands `fit()` a `FitOptions::default()` must not
silently get an unpenalized fit. `nn_l2` makes the opposite choice correctly,
because there the penalty is a tuning knob a caller legitimately sweeps.

**Adding the penalty inside the covariance step's `cov_ofv` closure** rather than
post-assembly. Rejected: it would double-count against the analytic R-matrix
route, and it would pay FD noise for a quantity that is known exactly.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#___ | not needed (yet) | after |

No ferx-r change is *required* by this PR — nothing existing breaks. A follow-up
ferx-r PR should surface `ofv_data` / `ofv_prior` / `prior_summary` in the R
`FitResult`, and will need the usual `tools/update-ferx-core-lock.sh` bump to
reach a commit where `ParameterPrior` etc. are public.

## Breaking changes

- [ ] None

Additive only — no existing field, syntax or default changed. The new API surface
and who needs it:

| Added item | Needed by |
|---|---|
| `types::ParameterPrior`, `types::PriorSpread` | ferx-r, to build a priored model programmatically (the R wrapper constructs `CompiledModel` rather than always going through the parser) |
| `CompiledModel::priors` | same — and it is the field the parser writes |
| `FitResult::{ofv_data, ofv_prior}` | ferx-r and any consumer that wants the likelihood half of a MAP fit for model comparison |
| `FitResult::prior_summary`, `types::PriorSummary` | ferx-r, to render the per-parameter prior table |

The existing surface does not suffice: there is no way to express "this parameter
has a prior" or to read back the data/prior split without these. All four are
reachable from ferx-r as an ordinary external crate.

`FitResult`'s three new fields carry `#[serde(default)]` and
`skip_serializing_if`, so an unpriored fit's YAML is byte-identical to before and
a pre-#254 artifact still deserializes.

<details>

None. `prior(...)` is opt-in; a model without one takes a bit-identical path
(`penalty()` over an empty term list returns exactly `0.0`, and `is_active()`
guards the gradient allocation).

</details>

## Numerical validation

- [x] Compared against: **an independent evaluation of the unpriored −2 log L at
      the priored optimum** (see below). NONMEM anchors are Phase 2 — see Open
      questions.
- [x] Reference values:

The load-bearing check is that `ofv_data` really is the unpenalized −2 log L,
computed **outside the priored code path**: take the priored fit's estimates,
put them into the plain unpriored model at `maxiter = 0`, and read the OFV.

```
# priored fit (examples/warfarin_prior.ferx)
ofv:       -282.615993
ofv_data:  -284.323764
ofv_prior:    1.707772
aic:       -270.323764      # = ofv_data + 2k, not ofv + 2k
bic:       -251.420402

TVCL 0.132917   TVV 7.781867   TVKA 0.842959
ETA_CL 0.045947  ETA_V 0.014078  ETA_KA 0.337464   PROP_ERR 0.010564

# the SAME estimates, evaluated by the UNPRIORED model at maxiter = 0
OFV:  -284.3237        # reproduces ofv_data exactly
```

Per-parameter report from that run — every estimate lands inside its prior's 95%
interval, with shifts under one prior SD:

```
  name: TVCL     prior 0.134   est 0.132917   shift -0.054   pen 0.003   [0.1000, 0.1795]
  name: TVV      prior 8.100   est 7.781867   shift -0.402   pen 0.161   [6.6616, 9.8490]
  name: TVKA     prior 1.000   est 0.842959   shift -0.443   pen 0.197   [0.4700, 2.1278]
  name: ETA_CL   prior 0.070   est 0.045947   shift -0.891   pen 0.794   [0.0277, 0.1767]
```

**An observation worth flagging rather than claiming credit for:** the stock
unpriored `examples/warfarin.ferx` converges to OFV **−280.364**, which is
*higher* than the priored run's `ofv_data` of −284.324 — from identical starting
values. The unpriored fit is therefore sitting in a worse local optimum (TVKA
0.725 vs 0.843) and the prior pulled the optimizer out of it. That is consistent
with the known FOCEI start-sensitivity on this model, but it is an incidental
finding on one dataset, not a claim about what priors do in general.

Beyond the fixture: 5196 tests pass, including every existing NONMEM anchor,
which is the real evidence that unpriored fits are byte-identical.

## Performance

Unpriored models take a bit-identical path. A priored model adds one
`Σ ((x−m)/s)²` per objective evaluation over the priored coordinates only —
unmeasurable next to a population inner loop. The covariance step adds **zero**
evaluations, because the penalty's curvature is a constant diagonal added
analytically rather than differenced.

## Tests

**Tier 1** — `src/estimation/priors_tests.rs` (18): the closed forms against
hand-computed values, not tolerances; the lognormal/normal straddle across
`theta_packs_log`; the variance-vs-`(sd)` differential pair (identical model,
identical prior location, penalties differing by exactly 4×, which is the only
thing that can distinguish the two scales); central-FD parity on gradient and
Hessian (realised worst error 2.12e-10 / 2.48e-5, bounds set ~4× above); and
every rejection.

**Tier 1** — `src/parser/model_parser_tests.rs` (9): the grammar, `25%` ≡ `0.25`,
the ambiguous-bare-`rse` reject *and* its straddle, coexistence with `FIX` and
`(sd)`, and the `weight =` ordering (a prior written after a kappa weight is
still seen — without the peel order the model parses cleanly and fits
unpenalized).

**Tier 2 / 3** — `tests/parameter_priors.rs` (9): the gate rejects, the objective
split, and two `slow-tests`-gated convergence tests for the issue's acceptance
criteria.

### What mutation testing found

Both of these were written, run green, and believed before being mutated:

- **The Tier-2 objective test passed with the penalty deleted from the
  optimizer.** `ofv_prior` and `prior_summary` are recomputed in `fit()` at the
  final estimate, so they read non-zero whether or not the objective ever saw the
  penalty. Added `a_prior_moves_the_estimates_and_not_only_the_report`, which
  puts two tight priors on opposite sides of the start and asserts the estimates
  separate; it is the only Tier-2 test that dies under that mutation.
- **Deleting only the gradient splice left the entire suite green.** An optimizer
  minimising a penalized objective with an unpenalized gradient still crawls
  toward the prior, just badly — no outcome assertion catches it reliably. Fixed
  structurally rather than with another assertion: value and gradient now come
  from one `penalty_and_gradient` call (the `nn_reg` pattern), so one mutation
  removes both.

Each wiring site was then mutated separately and named its own failing test —
the covariance `add_hessian` reddens only `a_tight_prior_tightens_the_reported_standard_error`,
the objective reddens only the two above.

`examples/warfarin_prior.ferx` — the model-updating pattern, with output above.

## Docs

- [x] New pages linked in `docs/_quarto.yml` (do **not** commit `docs/_site/` — it is git-ignored; CI builds & deploys it)
- [ ] No user-visible change

New `docs/estimation/priors.qmd` (scale table, the bounds trap, reading the
output, why the SEs are MAP SEs, what is out of scope, and a `$PRIOR` comparison
table), plus a cross-reference from `docs/model-file/parameters.qmd`.
`cargo run -p docs-lint -- --check` is clean.

## Checklist

- [x] Functions on the `Dual2` sensitivity path stay differentiable — **not touched at all**; the penalty is closed-form in packed space and never enters `sens/`

`cargo test --features ci`: **5196 passed, 0 failed, 176 ignored** across 186
binaries.

### ferx-site

`prior(...)` is new DSL syntax and `examples/warfarin_prior.ferx` is a new
example, so both boxes need a follow-up in ferx-site / ferx-book. No new data
file (it reuses `data/warfarin.csv`). Flagging rather than silently ticking.

### ferx-book

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core — not needed, no ferx-r change in this PR
- [x] All affected `examples/*.ferx` run cleanly via CLI (`warfarin_prior.ferx` and the unchanged `warfarin.ferx`)
- [ ] All affected ferx-site example `.qmd` pages render cleanly — see above
- [ ] All affected ferx-book chapters render cleanly — see above

## Reviewer hints

Read in this order:

1. **`src/estimation/priors.rs`** — the whole design. `PriorScale` is the single
   derivation of "what scale is this prior on"; if the mean and the spread
   disagreed about it the error would be a factor of two and would still look
   plausible, which is why they share one enum.
2. **`api/validation.rs::check_parameter_priors`** — the gate. Everything else
   assumes a prior that reaches an optimizer has already been checked against the
   same `ModelParameters` layout, which is why `build_prior_set` can fall back to
   an empty set.
3. **`covariance.rs`**, the `cov_priors.add_hessian` block — the decision I feel
   most strongly about, with the reasoning in the comment. Note the existing NN
   penalty is *absent* from the covariance Hessian; priors deliberately do not
   inherit that gap, and I did not change `nn_reg` here.

Skimmable: the 29 `priors: Vec::new()` struct-literal additions, and the 51
widened `parse_parameters` destructurings in the parser tests.

## Open questions

**Phase 2, not in this PR** (three of the eight acceptance criteria):

- IMP / IMPMAP penalized M-step — the issue's "composes with at least FOCE and
  one EM method". Their M-steps are BOBYQA over the packed vector
  (`impmap.rs:2479`, `:2599`), so this is a straightforward extension. SAEM's
  closed-form M-step needs a conjugate MAP update and is a larger piece.
- `[priors] from_fit = "…"` — the one-call import from a previous ferx fit
  (`value = estimate`, `rse = SE/estimate`), which is the actual model-updating
  ergonomics.
- **NONMEM anchors.** My first read was that the Ω prior needs a
  no-equivalent-estimator exception, since NWPRI's Ω prior is inverse-Wishart and
  ours is lognormal-on-the-variance. On a second pass I think we can anchor both
  families exactly, if the control stream is written in the matching
  parameterization: a log-packed θ against NWPRI on `THETA(1)` in a model written
  `CL = EXP(THETA(1))`, and a diagonal Ω against NWPRI on `THETA(k)` in a model
  written `$OMEGA 1 FIX` with `CL = TVCL*EXP(EXP(THETA(k))*ETA(1))` — which *is*
  normal on ln(SD), i.e. our packed Ω coordinate. The Ω one leans on FOCE's
  marginal being invariant under a linear rescaling of η, so it is only an anchor
  if that invariance holds numerically in NONMEM: step 1 is a **null control run**
  (no prior, both parameterizations, assert identical OFV) and the priored
  comparison is anchored on ΔOFV against that twin so the normalization constants
  cancel. If the null run disagrees, I will say the anchor does not exist rather
  than tune a tolerance until it passes.

**Decisions worth overruling if you disagree** — these four change user-visible
behaviour, and I would rather argue now than after it ships:

- **The prior's family follows the declared bounds.** Widening a θ's lower bound
  below zero silently moves its prior from lognormal to normal. The mitigation is
  visibility (the report prints the realised family and 95% interval) rather than
  a second convention, but it is a real coupling.
- **AIC/BIC on `ofv_data`.** Defensible, but it means a priored fit's AIC is not
  comparable to the number NONMEM prints for the same run.
- **The penalty in the covariance step.** Makes the SE depend on the prior's
  strength, mechanically. I think that is correct and the docs say so plainly,
  but it is the sort of thing a reviewer should push back on if they disagree.
- **Hard error on an unapplied prior.** The existing `nn_l2` analogue warns
  instead. I think the asymmetry is right — a dropped regularizer changes a
  tuning knob, a dropped prior changes the model — but it is a deliberate
  inconsistency.

Closes #254 partially; I would keep the issue open for Phase 2.

https://claude.ai/code/session_014u63GvTSbfYB1hL8cd3Z4X

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Added
+
+- **Per-parameter priors for penalized maximum-likelihood (MAP) estimation**, declared inline as `prior(value, rse = 25%)` on any `theta`, `omega`, `sigma` or `kappa` — the simple alternative to NONMEM `$PRIOR`, with no separate prior problem and no matrices. The fit reports the data and prior halves of the OFV separately plus a per-parameter shift-toward-prior summary; AIC/BIC stay on the data half. Applies to `foce`, `focei`, `laplace`, `gn` and `gn_hybrid` (#254).
+
 ### Changed
 
 - **SAEM now averages the residual sufficient statistic for eligible single additive and proportional error models, reducing final-draw Monte Carlo noise in the residual SD estimate (#1321).**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ section of the SDLC for the versioning policy).
 
 ### Added
 
-- **Per-parameter priors for penalized maximum-likelihood (MAP) estimation**, declared inline as `prior(value, rse = 25%)` on any `theta`, `omega`, `sigma` or `kappa` — the simple alternative to NONMEM `$PRIOR`, with no separate prior problem and no matrices. The fit reports the data and prior halves of the OFV separately plus a per-parameter shift-toward-prior summary; AIC/BIC stay on the data half, and the prior's curvature reaches the reported standard errors and the SIR intervals. Applies to `foce`, `focei`, `laplace`, `gn` and `gn_hybrid`; a chain whose last estimating stage cannot apply priors, and `covariance_method = s` (which cannot represent one), are refused rather than run unpenalized. The θ prior is anchored against NONMEM `$PRIOR NWPRI` (#254).
+- **Per-parameter priors for penalized maximum-likelihood (MAP) estimation**, declared inline as `prior(value, rse = 25%)` on any `theta`, `omega`, `sigma` or `kappa` — the simple alternative to NONMEM `$PRIOR`, with no separate prior problem and no matrices. The fit reports the data and prior halves of the OFV separately plus a per-parameter shift-toward-prior summary; AIC/BIC stay on the data half, and the prior's curvature reaches the reported standard errors and the SIR intervals. Applies to `foce`, `focei`, `laplace`, `gn` and `gn_hybrid`; a chain whose last estimating stage cannot apply priors, and an `S`-based covariance estimator (`covariance_method = s` or `rsr`, neither of which can represent one), are refused rather than run unpenalized. The θ prior is anchored against NONMEM `$PRIOR NWPRI` (#254).
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ section of the SDLC for the versioning policy).
 
 ### Added
 
-- **Per-parameter priors for penalized maximum-likelihood (MAP) estimation**, declared inline as `prior(value, rse = 25%)` on any `theta`, `omega`, `sigma` or `kappa` — the simple alternative to NONMEM `$PRIOR`, with no separate prior problem and no matrices. The fit reports the data and prior halves of the OFV separately plus a per-parameter shift-toward-prior summary; AIC/BIC stay on the data half. Applies to `foce`, `focei`, `laplace`, `gn` and `gn_hybrid` (#254).
+- **Per-parameter priors for penalized maximum-likelihood (MAP) estimation**, declared inline as `prior(value, rse = 25%)` on any `theta`, `omega`, `sigma` or `kappa` — the simple alternative to NONMEM `$PRIOR`, with no separate prior problem and no matrices. The fit reports the data and prior halves of the OFV separately plus a per-parameter shift-toward-prior summary; AIC/BIC stay on the data half, and the prior's curvature reaches the reported standard errors and the SIR intervals. Applies to `foce`, `focei`, `laplace`, `gn` and `gn_hybrid`; a chain whose last estimating stage cannot apply priors, and `covariance_method = s` (which cannot represent one), are refused rather than run unpenalized. The θ prior is anchored against NONMEM `$PRIOR NWPRI` (#254).
 
 ### Changed
 
--- a/api/ferx-core-public-api.txt
+++ b/api/ferx-core-public-api.txt
@@ -2390,6 +2390,9 @@ pub ferx_core::types::Prediction::Continuous
 pub ferx_core::types::Prediction::Continuous::pred: f64
 impl ferx_core::Prediction
 pub fn ferx_core::Prediction::prob(&self, k: usize) -> core::option::Option<f64>
+pub enum ferx_core::types::PriorSpread
+pub ferx_core::types::PriorSpread::Rse(f64)
+pub ferx_core::types::PriorSpread::Sd(f64)
 pub enum ferx_core::types::RateMode
 pub ferx_core::types::RateMode::Fixed
 pub ferx_core::types::RateMode::ModeledDuration
@@ -2570,6 +2573,7 @@ pub ferx_core::types::CompiledModel::pk_idx_f64: alloc::vec::Vec<f64>
 pub ferx_core::types::CompiledModel::pk_indices: alloc::vec::Vec<usize>
 pub ferx_core::types::CompiledModel::pk_model: ferx_core::PkModel
 pub ferx_core::types::CompiledModel::pk_param_fn: ferx_core::PkParamFn
+pub ferx_core::types::CompiledModel::priors: alloc::vec::Vec<ferx_core::ParameterPrior>
 pub ferx_core::types::CompiledModel::referenced_covariates: alloc::vec::Vec<alloc::string::String>
 pub ferx_core::types::CompiledModel::residual_correlations: alloc::vec::Vec<ferx_core::ResidualCorrelation>
 pub ferx_core::types::CompiledModel::residual_error_eta: core::option::Option<usize>
@@ -2973,6 +2977,8 @@ pub ferx_core::types::FitResult::nlopt_missing_algorithms: alloc::vec::Vec<alloc
 pub ferx_core::types::FitResult::npde_seed: core::option::Option<u64>
 pub ferx_core::types::FitResult::obs_time_range: core::option::Option<(f64, f64)>
 pub ferx_core::types::FitResult::ofv: f64
+pub ferx_core::types::FitResult::ofv_data: f64
+pub ferx_core::types::FitResult::ofv_prior: f64
 pub ferx_core::types::FitResult::omega: nalgebra::base::alias::DMatrix<f64>
 pub ferx_core::types::FitResult::omega_fixed: alloc::vec::Vec<bool>
 pub ferx_core::types::FitResult::omega_init: nalgebra::base::alias::DMatrix<f64>
@@ -2985,6 +2991,7 @@ pub ferx_core::types::FitResult::optimizer: alloc::string::String
 pub ferx_core::types::FitResult::outer_gtol: f64
 pub ferx_core::types::FitResult::outer_maxiter: usize
 pub ferx_core::types::FitResult::packed_estimate: core::option::Option<alloc::vec::Vec<f64>>
+pub ferx_core::types::FitResult::prior_summary: alloc::vec::Vec<ferx_core::PriorSummary>
 pub ferx_core::types::FitResult::residual_correlation_fixed: alloc::vec::Vec<bool>
 pub ferx_core::types::FitResult::residual_correlations: alloc::vec::Vec<ferx_core::ResidualCorrelation>
 pub ferx_core::types::FitResult::restored_from_checkpoint: bool
@@ -3132,6 +3139,10 @@ pub fn ferx_core::OmegaMatrix::from_chol_factor(l: nalgebra::base::alias::DMatri
 pub fn ferx_core::OmegaMatrix::from_diagonal(variances: &[f64], names: alloc::vec::Vec<alloc::string::String>) -> Self
 pub fn ferx_core::OmegaMatrix::from_matrix(m: nalgebra::base::alias::DMatrix<f64>, names: alloc::vec::Vec<alloc::string::String>, diagonal: bool) -> Self
 pub fn ferx_core::OmegaMatrix::from_matrix_with_mask(m: nalgebra::base::alias::DMatrix<f64>, names: alloc::vec::Vec<alloc::string::String>, diagonal: bool, free_mask: nalgebra::base::alias::DMatrix<bool>) -> Self
+pub struct ferx_core::types::ParameterPrior
+pub ferx_core::types::ParameterPrior::name: alloc::string::String
+pub ferx_core::types::ParameterPrior::spread: ferx_core::PriorSpread
+pub ferx_core::types::ParameterPrior::value: f64
 pub struct ferx_core::types::ParsedModel
 pub ferx_core::types::ParsedModel::adaptive_dosing: core::option::Option<ferx_core::sim::adaptive::AdaptiveDosingSpec>
 pub ferx_core::types::ParsedModel::bindings: ferx_core::parser::model_parser::ParseBindings
@@ -3183,6 +3194,15 @@ pub ferx_core::types::PosteriorSummary::q025: f64
 pub ferx_core::types::PosteriorSummary::q975: f64
 pub ferx_core::types::PosteriorSummary::rhat: f64
 pub ferx_core::types::PosteriorSummary::sd: f64
+pub struct ferx_core::types::PriorSummary
+pub ferx_core::types::PriorSummary::estimate: f64
+pub ferx_core::types::PriorSummary::family: alloc::string::String
+pub ferx_core::types::PriorSummary::name: alloc::string::String
+pub ferx_core::types::PriorSummary::penalty: f64
+pub ferx_core::types::PriorSummary::prior_lower_95: f64
+pub ferx_core::types::PriorSummary::prior_upper_95: f64
+pub ferx_core::types::PriorSummary::prior_value: f64
+pub ferx_core::types::PriorSummary::shift_in_prior_sds: f64
 pub struct ferx_core::types::ResidualCorrelation
 pub ferx_core::types::ResidualCorrelation::rho: f64
 pub ferx_core::types::ResidualCorrelation::sigma_i: usize
@@ -3650,6 +3670,9 @@ pub ferx_core::Prediction::Continuous
 pub ferx_core::Prediction::Continuous::pred: f64
 impl ferx_core::Prediction
 pub fn ferx_core::Prediction::prob(&self, k: usize) -> core::option::Option<f64>
+pub enum ferx_core::PriorSpread
+pub ferx_core::PriorSpread::Rse(f64)
+pub ferx_core::PriorSpread::Sd(f64)
 pub enum ferx_core::RateMode
 pub ferx_core::RateMode::Fixed
 pub ferx_core::RateMode::ModeledDuration
@@ -3900,6 +3923,7 @@ pub ferx_core::CompiledModel::pk_idx_f64: alloc::vec::Vec<f64>
 pub ferx_core::CompiledModel::pk_indices: alloc::vec::Vec<usize>
 pub ferx_core::CompiledModel::pk_model: ferx_core::PkModel
 pub ferx_core::CompiledModel::pk_param_fn: ferx_core::PkParamFn
+pub ferx_core::CompiledModel::priors: alloc::vec::Vec<ferx_core::ParameterPrior>
 pub ferx_core::CompiledModel::referenced_covariates: alloc::vec::Vec<alloc::string::String>
 pub ferx_core::CompiledModel::residual_correlations: alloc::vec::Vec<ferx_core::ResidualCorrelation>
 pub ferx_core::CompiledModel::residual_error_eta: core::option::Option<usize>
@@ -4356,6 +4380,8 @@ pub ferx_core::FitResult::nlopt_missing_algorithms: alloc::vec::Vec<alloc::strin
 pub ferx_core::FitResult::npde_seed: core::option::Option<u64>
 pub ferx_core::FitResult::obs_time_range: core::option::Option<(f64, f64)>
 pub ferx_core::FitResult::ofv: f64
+pub ferx_core::FitResult::ofv_data: f64
+pub ferx_core::FitResult::ofv_prior: f64
 pub ferx_core::FitResult::omega: nalgebra::base::alias::DMatrix<f64>
 pub ferx_core::FitResult::omega_fixed: alloc::vec::Vec<bool>
 pub ferx_core::FitResult::omega_init: nalgebra::base::alias::DMatrix<f64>
@@ -4368,6 +4394,7 @@ pub ferx_core::FitResult::optimizer: alloc::string::String
 pub ferx_core::FitResult::outer_gtol: f64
 pub ferx_core::FitResult::outer_maxiter: usize
 pub ferx_core::FitResult::packed_estimate: core::option::Option<alloc::vec::Vec<f64>>
+pub ferx_core::FitResult::prior_summary: alloc::vec::Vec<ferx_core::PriorSummary>
 pub ferx_core::FitResult::residual_correlation_fixed: alloc::vec::Vec<bool>
 pub ferx_core::FitResult::residual_correlations: alloc::vec::Vec<ferx_core::ResidualCorrelation>
 pub ferx_core::FitResult::restored_from_checkpoint: bool
@@ -4543,6 +4570,10 @@ pub fn ferx_core::OmegaMatrix::from_chol_factor(l: nalgebra::base::alias::DMatri
 pub fn ferx_core::OmegaMatrix::from_diagonal(variances: &[f64], names: alloc::vec::Vec<alloc::string::String>) -> Self
 pub fn ferx_core::OmegaMatrix::from_matrix(m: nalgebra::base::alias::DMatrix<f64>, names: alloc::vec::Vec<alloc::string::String>, diagonal: bool) -> Self
 pub fn ferx_core::OmegaMatrix::from_matrix_with_mask(m: nalgebra::base::alias::DMatrix<f64>, names: alloc::vec::Vec<alloc::string::String>, diagonal: bool, free_mask: nalgebra::base::alias::DMatrix<bool>) -> Self
+pub struct ferx_core::ParameterPrior
+pub ferx_core::ParameterPrior::name: alloc::string::String
+pub ferx_core::ParameterPrior::spread: ferx_core::PriorSpread
+pub ferx_core::ParameterPrior::value: f64
 pub struct ferx_core::ParsedModel
 pub ferx_core::ParsedModel::adaptive_dosing: core::option::Option<ferx_core::sim::adaptive::AdaptiveDosingSpec>
 pub ferx_core::ParsedModel::bindings: ferx_core::parser::model_parser::ParseBindings
@@ -4619,6 +4650,15 @@ pub ferx_core::PreparedRun::init_params: ferx_core::ModelParameters
 pub ferx_core::PreparedRun::model_hash: core::option::Option<alloc::string::String>
 pub ferx_core::PreparedRun::parsed: ferx_core::ParsedModel
 pub ferx_core::PreparedRun::population: ferx_core::Population
+pub struct ferx_core::PriorSummary
+pub ferx_core::PriorSummary::estimate: f64
+pub ferx_core::PriorSummary::family: alloc::string::String
+pub ferx_core::PriorSummary::name: alloc::string::String
+pub ferx_core::PriorSummary::penalty: f64
+pub ferx_core::PriorSummary::prior_lower_95: f64
+pub ferx_core::PriorSummary::prior_upper_95: f64
+pub ferx_core::PriorSummary::prior_value: f64
+pub ferx_core::PriorSummary::shift_in_prior_sds: f64
 pub struct ferx_core::ResidualCorrelation
 pub ferx_core::ResidualCorrelation::rho: f64
 pub ferx_core::ResidualCorrelation::sigma_i: usize
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -102,6 +102,7 @@ website:
           - estimation/categorical.qmd
           - estimation/ctmm.qmd
           - estimation/mixture.qmd
+          - estimation/priors.qmd
       - text: "Tools"
         href: tools/index.qmd
         contents:
--- a/docs/estimation/priors.qmd
+++ b/docs/estimation/priors.qmd
@@ -1,0 +1,212 @@
+---
+pagetitle: "Parameter priors in ferx"
+description-meta: "Stabilize sparse-data fits with per-parameter priors: penalized maximum likelihood (MAP) estimation in the ferx NLME engine, the simple alternative to NONMEM $PRIOR."
+---
+
+# Parameter Priors (penalized ML / MAP)
+
+> **Maturity: experimental** — see [Feature Maturity](../maturity.qmd) for what this means.
+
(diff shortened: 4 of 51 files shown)


roninsightrx:
Review findings on `762f6aa`:

1. **[P1] Evaluation-only tail stages are mistaken for estimators** — `src/api/validation.rs:4181-4196` checks `chain.last()`, despite the comment promising the final *estimating* stage. With `methods = [saem, laplace]` plus `agq_eval_only = true`, validation accepts the prior because Laplace is nominally supported, but that stage only evaluates the SAEM parameters, so no optimizer ever applies the prior. The inverse case, `[focei, imp]` plus `imp_eval_only = true`, is rejected even though FOCEI did apply it. Please use the existing evaluation-only classification to locate the last actual estimator, and add both regression directions.

2. **[P1] Priored results do not round-trip through .fitrx** — `FitWire` still stores only `ofv` (`src/io/fitrx.rs:83-89,565-576`), and the loader unconditionally reconstructs `ofv_data = w.ofv`, `ofv_prior = 0`, and an empty summary (`src/io/fitrx.rs:1877-1885`). A newly saved priored fit therefore reloads with its penalized total mislabeled as data likelihood, loses every prior row, and breaks `aic == ofv_data + 2k`. Add backward-compatible optional wire fields and a priored round-trip test; the existing minimal fixture currently masks this because it sets `ofv=100` but `ofv_data=0`.

3. **[P1] SIR weights ignore the prior** — both normal SIR and covariance fallback receive the clean `result.ofv` in `src/api/fit.rs:2204-2252`, while `src/estimation/sir.rs:696-712` scores each draw with data `2*nll` only. Thus a fit may be centered/proposed using prior-regularized estimates and curvature, then resampled against the unpenalized likelihood; a tight prior can disappear from the reported SIR intervals. Please include the packed prior penalty in both the baseline and sampled objectives, or reject SIR combinations until supported.

4. **[P1] Block membership is inferred for the whole Omega matrix, not per coordinate** — `src/estimation/priors.rs:401-440` sets `omega_is_block = !template.omega.diagonal` (and likewise for IOV), then rejects every diagonal coordinate when it is true. Mixed models are supported elsewhere, so `block_omega(A,B)` plus an independent `omega C ~ ... prior(...)` incorrectly rejects the valid prior on C, even though the diagnostic/docs explicitly recommend declaring diagonal variances separately. The same global test affects mixed block/diagonal kappa; the resolver needs actual per-coordinate block membership. Please also cover the `(sd)` case, since packed Cholesky offsets are not eta indices in a mixed matrix.

5. **[P2] covariance_method = s silently discards prior curvature** — after adding the prior Hessian to R, `src/estimation/covariance.rs:1267-1282` returns `S^-1` for the cross-product method, and S contains data scores only. The output still states unconditionally that these are MAP standard errors from the penalized curvature. This also happens automatically when a large default-R problem is scale-routed to S. Please reject/route this combination or define and document the intended penalized S estimator; RSR already uses the prior-adjusted R and is a different case.

**Repository-policy blocker:** the PR explicitly defers the numerical NONMEM anchor to Phase 2, but AGENTS.md requires every numerical feature to include a NONMEM output comparison in the feature PR. The conceptual comparison table and the internal unpriored-OFV consistency check do not satisfy that requirement.

Verification:
- `cargo test --lib estimation::priors::tests`: 17 passed
- `cargo test --test parameter_priors`: 8 passed, 2 slow tests ignored
- `cargo test --lib`: 4,417 passed, 23 ignored
- `git diff --check origin/main...HEAD`: clean

GitHub does not permit the PR author account to submit a formal request-changes review on its own PR, so these are posted as a blocking review comment.

roninsightrx:
All five findings were real, including two where my own code comments asserted the opposite. Fixed, each with a regression test that I verified **dies** under a mutation reverting the fix. The policy blocker is addressed with a measured NONMEM anchor.

## 1. [P1] Evaluation-only tail stages — fixed

Confirmed in both directions. `FitOptions::covariance_stage()` already existed and is exactly "last stage that estimates, ignoring a trailing run of evaluation-only stages"; the gate now uses it, so the prior gate and the covariance step cannot disagree about which stage produced the reported estimates.

`the_prior_method_gate_looks_past_evaluation_only_stages` covers both directions in one test, because each alone passes under a *different* wrong predicate — `chain.last()` accepts the first and rejects the second, while "any stage applies priors" accepts both.

## 2. [P1] `.fitrx` round trip — fixed

Confirmed, and my comment claiming the loader only ever sees pre-prior artifacts was simply wrong: `save_fit` serializes a completed `FitResult`, so a newly-saved priored fit was exactly the case that broke.

`FitWire` gains `ofv_data` / `ofv_prior` / `prior_summary`, all optional and skipped when empty, so an unpriored artifact stays byte-identical and a pre-#254 file still loads (`ofv_data` falls back to `ofv`, which for those *is* the data likelihood). You were right that the existing fixture masked it — the new test asserts `ofv_data != ofv` explicitly so a fixture where they coincide can't hide a regression again.

## 3. [P1] SIR weights — fixed

Confirmed. Both the baseline and the per-draw objective now carry the penalty, added inside `run_sir_core_scoped` rather than at the call sites — one derivation, and it covers the normal path and the non-PD covariance fallback together (both pass the clean `result.ofv`).

**The first version of this test was vacuous** and mutation testing caught it. At RSE 2% the penalized proposal is so tight that every draw lands within a hair of the MAP, both weightings agree, and the assertion passed with the prior deleted from the sampler. Measured across the range to find the regime where the weights actually decide something:

| RSE | CI with the prior in the weights | without |
|-----|----------------------------------|---------|
| 2%  | (0.1931, 0.2096) | (0.1863, 0.2191) — both at the prior, no signal |
| **10%** | **(0.1384, 0.1927)** | **(0.1251, 0.1572)** |
| 20% | (0.1245, 0.1595) | (0.1215, 0.1595) — prior too weak to matter |
| 40% | (0.1210, 0.1555) | (0.1197, 0.1555) |

At 10% the point estimate is identical either way (0.156190 — the *fit* is unaffected, only the resampling), so anything that moves is the weights. The test asserts there, on the upper bound and on the sign of `mid − est` (+0.0093 correct vs −0.0150 under the bug). The table is in the test's doc comment.

## 4. [P1] Block membership — fixed

Confirmed, and worse than "incorrectly rejects": the diagnostic tells users to declare diagonals separately, which was impossible to act on. Membership now comes from `free_mask` per coordinate — a diagonal `(e,e)` is in a block iff some off-diagonal in its row or column is free.

You were also right about the `(sd)` case, and it was a second live bug rather than a hypothetical: `coordinate_table` indexed `omega_init_as_sd` by the *packed offset*, which is only the eta index for a diagonal Ω. The Ω/Ω_IOV segments are now walked with `lower_tri_iter` so every coordinate carries its real `(row, col)`, and the flag is keyed on the row. The test pins the scale rather than just the acceptance: in a mixed Ω where only `ETA_C` is `(sd)`, the penalty must be exactly 0 at the prior mean — under the old packed-offset lookup it would read `ETA_A`'s flag and silently use the variance scale. Both sub-bugs are separately mutation-verified.

## 5. [P2] `covariance_method = s` — fixed by rejecting

Confirmed. There is no sensible penalized `S`: it is a sum over *subjects*, and a prior contributes one score for the whole population, so the combination is refused with a message pointing at `r` / `rsr`. The scale-routing path you flagged is handled too — `scale_routed_covariance_method` is now prior-aware and stays on `r` (with a warning that it will be slow) instead of silently routing to an estimator that cannot carry the prior. RSR is left alone, as you suggested.

## Repository-policy blocker — NONMEM anchor now run

You were right that the comparison table and the internal consistency check don't satisfy the requirement. The anchor exists now: `nonmem_anchor/prior_theta_{null,nwpri,nwpri_centered}.ctl` + ferx twins, with the `.lst` files committed and the comparison pinned as a test.

Both sides write the model on the log scale (`CL = EXP(THETA(1) + ETA(1))`) and the ferx θ is declared with a negative lower bound so ferx packs it as the identity — same coordinate, same family, same mean, same SD, so nothing rests on ferx's packing conventions. `MAXEVAL=0` / `maxiter = 0` on both, since the quantity under test is the objective and not a minimizer's path to it. **Null control first**, because without it an agreeing prior term could be two compensating errors:

| | NONMEM 7.5.1 | ferx |
|---|---|---|
| null (no prior) | 177.82632978040530 | 177.826330 |
| prior centred on θ | 177.826 (== null) | — |
| prior offset from θ | 177.94393305151641 | 177.943933 |
| **prior contribution** | **0.11760327111** | **0.117603** |

Worst realised disagreement **5.2e-8**, at the limit of the six decimals ferx's YAML prints. The third run — prior mean placed exactly on `THETA(1)`, reproducing the null — is what establishes that NWPRI drops the prior's normalisation constant just as ferx does, so these are comparable absolutely rather than only as a difference. The closed form `((-2.0 + 1.89712)/0.30)² = 0.117603266` is reproduced by both, so the arm pins ferx against NONMEM *and* against arithmetic.

Two NWPRI traps are recorded in `nonmem_anchor/README.md`, since neither message names its cause: `NTHETA`/`NETA` must match the estimation problem's counts (not the number you want to prior), and `$THETAPV` must use `BLOCK(n) FIX` rather than per-element `FIX`.

**What the anchor does not cover, stated plainly:** the Ω prior. ferx's is lognormal on the variance, NWPRI's is inverse-Wishart — genuinely different priors, so there is no NONMEM run to difference against. Ω is pinned instead by closed-form unit tests including the variance-vs-`(sd)` pair (penalties must differ by exactly 4×). Anchoring it would need the `$OMEGA 1 FIX` + `EXP(THETA(k))·ETA(1)` reparameterisation, which rests on FOCE's marginal being invariant under a linear rescaling of η; that needs its own null control and I have not run it. I'd rather say so than claim coverage I don't have.

## Mutation verification

Every fix was reverted in turn and the corresponding test confirmed to fail:

| Mutation | Result |
|---|---|
| eval-only stage gate → `chain.last()` | DIED |
| fitrx loader reconstructs `ofv_data` from `ofv` | DIED |
| SIR draw drops the prior | DIED (after the test was rewritten — first version survived) |
| Ω block test → matrix-level `!diagonal` | DIED |
| Ω `init_as_sd` keyed on the packed offset | DIED |
| cross-product covariance rejection removed | DIED |

`tools/preflight.sh` green. Full suite: **5214 passed, 0 failed, 177 ignored** across 186 binaries. Pushed as `1ae56c1a`.

roninsightrx:
## Review — and the five findings, now fixed in 39bdb911

Reviewed `main...HEAD` at medium effort. The core math holds up: the penalty lives in packed space, the Hessian contribution is the constant diagonal `2/s²`, `cov = 2·H⁻¹` so a pure-prior coordinate reports `SE = s`, and the lognormal CV→log-SD conversion and the ½ factor for a variance-declared Ω/Σ are both right. The packed-segment ordering in `coordinate_table` matches `pack_params`.

Five findings, all addressed in the follow-up commit. Each fix was mutation-checked — and separately per side where a test has two.

### 1. A parameter named `PRIOR` no longer parsed (medium)

`split_prior_modifier` matched the keyword case-insensitively with only a **left** word boundary. On `omega PRIOR ~ 0.1` it matched at the name, found `~` instead of `(`, and the `is_ident` escape did not fire because the next byte is a space — so it returned `Err("prior must be written as a call")`. Any pre-existing model with a theta/omega/sigma/kappa named `PRIOR` became a hard parse error. Separately, `theta prior(0.1, 0, 10)` — a theta legitimately named `prior` — was swallowed as the modifier and failed on the argument list instead.

**Fix:** the modifier is recognised only when it is *not* in a declaration's name slot (`preceded_by_declaration_keyword`), which is checked before the call-shape test. The bare-`prior`-tail safety net is unchanged, so a dropped modifier is still an error rather than a silent no-op.

The two spellings fail through different paths, so the test asserts both, plus the straddle (`theta prior(...) prior(0.2, rse = 25%)` on one line — same keyword, both meanings) and the surviving bare-tail error. Disabling the guard reddens each half independently.

### 2. The gate was not wired where its own doc said it was (medium)

`build_prior_set`'s doc-comment named `check_model_data` as the gate, but `check_parameter_priors` had exactly one call site: `fit_inner`. So `ferx check` / `validate_model_file` reported a model with an unresolvable prior — unknown name, FIXed target, block element, duplicate, unsupported method — as **valid**, and the user discovered it at fit time instead. That is the worst ordering, since checking is precisely what you run to avoid that.

**Fix:** wired into `validate_model_file` next to `check_variance_init_rails` / `check_packed_start_in_box`, which have the same shape and the same data-independence, and corrected the doc-comment to name `check_parameter_priors`. Three new tests in `src/api/tests/parameter_prior_check_tests.rs`, each with a straddle so none is satisfied by a gate that rejects every priored model; deleting the wiring reddens all three.

### 3. `covariance_method = rsr` under a prior gave incoherent SEs (medium)

Check 3 rejected `CrossProduct` because `S` has no subject decomposition for a prior, but accepted `Sandwich`. After this PR `R` carries `+2/s²` while `S` carries no prior score at all, so `R⁻¹ S R⁻¹` is neither the penalized estimator nor the unpenalized one — it wraps a prior-free `S` in two prior-shrunk `R⁻¹` factors and systematically under-states the SE on every priored coordinate. The `s` diagnostic then actively recommended `rsr` as the safe alternative.

**Fix:** both `S`-based estimators are rejected, each with its own reason in the message, and the suggestion points only at `r`. A wrong-by-construction robust SE is worse than no robust SE, hence an error rather than a warning. The existing `a_prior_with_the_cross_product_covariance_is_refused` was widened to cover both methods rather than duplicated; it also asserts the suggestion no longer names `rsr`, and that an *unpriored* fit may still use either. `docs/estimation/priors.qmd` updated to match.

### 4. Standalone `run_sir` double-counted the prior in `ofv_hat` (low)

`run_sir_core`'s contract is "`ofv_hat` is the **data** OFV; the penalty is added internally". The inline path in `fit.rs` passes the clean `OuterResult::ofv`, but `run_sir.rs` passed `fit.ofv` — which this PR redefined as the penalized total. It cancels today, since `dofv` only enters normalized log-weights, but the contract is what the next reader will rely on.

**Fix:** extracted as `data_ofv(fit)`, computed as `ofv - ofv_prior` rather than reading `ofv_data`. Both new fields are `#[serde(default)]`, so a `FitResult` deserialized from a pre-#254 YAML carries `ofv_data = 0.0` while `ofv_prior = 0.0` is the truth there — one spelling is correct for an old fit and a new one alike. The test pins both inputs, and the two plausible-looking mutations (`fit.ofv`, `fit.ofv_data`) redden different assertions.

### 5. Block detection could mis-scale a partially-fixed block diagonal (low)

`in_block` was "any **free** off-diagonal in the row/col". A `block_omega` with a fixed non-zero off-diagonal and free diagonals would have had its diagonal accepted, and the packed coordinate `ln(L[r,r])` treated as `ln(SD_r)` — wrong for `r > 0`, where `SD_r² = Σⱼ L[r,j]²`. Unreachable today, because `packed_fixed_mask` fixes the whole row and column of a FIXed eta, but the predicate depended on that silently.

**Fix:** the predicate consults `om.matrix` as well as `free_mask`. The test builds the arrangement directly (identity `free_mask`, non-diagonal matrix) since the parser cannot produce it, and its straddle is the same shape with a zero off-diagonal, which must still resolve — without that, the first half would be satisfied by a predicate that rejects every non-diagonal Ω, which is the mixed-Ω regression the neighbouring test exists to prevent.

---

`tools/preflight.sh` green (fmt, check matrix, clippy, rustdoc, docs, public-API); `cargo test --lib` 4424 passed / 0 failed; `tests/parameter_priors.rs` 13 passed / 3 slow-gated.

PR #1343, opened 2026-09-11
perf(sens): sweep only the eta axes for the Laplace third-order jet [closes #1342]

### LINKED ISSUE #1342: perf(sens): the Laplace exact-anchor derivative sweeps every theta axis and discards the result — 1 + 2(t+d) provider evaluations where 1 + 2d suffices
Delivery step 2 of `plans/parallel-optimization-implementation-2026-09-10.md`, the plan's
strongest finding. Follows #1329 (steps 1 & 4), #1333 (step 3) and #1335 (which made this route
the **default**, so the saving below now applies to every `method = laplace` fit rather than an
opt-in one).

## The waste, confirmed against the code

`laplace_h_deriv::subject_h_inner_dx` gets its third-order blocks from
`sens::provider::subject_sensitivities_cov` → `covariance_sensitivities`, which sweeps a central
FD pair along **every θ and η axis**:

```rust
let n_axes = n_theta + n_eta;
for c in 0..n_axes { /* plus.push(evaluate(..)); minus.push(evaluate(..)); */ }
```

i.e. `1 + 2(n_theta + n_eta)` second-order sensitivity evaluations per subject per sweep. But the
Laplace consumer reads only four blocks (`laplace_h_deriv.rs:224-227, 274, 294, 312, 316`):

| block | where it comes from | needs a θ axis? |
|---|---|---|
| `d2f_deta2` | base evaluation | no |
| `d2f_deta_dtheta` | base evaluation | no |
| `d3f_deta3` | `∂/∂η_c` of `d2f_deta2` | **no** |
| `d3f_deta2_dtheta` | currently `∂/∂θ_c` of `d2f_deta2` | see below |

It never touches `d2f_dtheta2` or `d3f_deta_dtheta2` — the only two blocks that genuinely require
θ-direction perturbations, and both exist for the *covariance* assembly.

## The substitution

`d3f_deta2_dtheta[(k,l),m] = ∂³f/∂η_k ∂η_l ∂θ_m` is currently formed as `∂/∂θ_m` of
`∂²f/∂η_k∂η_l`. By mixed-partial symmetry it is equally `∂/∂η_l` of `∂²f/∂η_k∂θ_m` — and
`d2f_deta_dtheta` is already returned by every base evaluation. So differencing the **η axes
only** yields both third-order blocks the Laplace derivative needs:

```text
T[i,j,k] = ∂/∂η_k ( d2f_deta2[i,j]      )   → d3f_deta3
U[i,j,m] = ∂/∂η_j ( d2f_deta_dtheta[i,m] )   → d3f_deta2_dtheta
```

Cost: **`1 + 2·n_eta` instead of `1 + 2(n_theta + n_eta)`.**

| n_theta / n_eta | now | after | ratio |
|---|---:|---:|---:|
| 3 / 3 (warfarin) | 13 | 7 | 1.86× |
| 3 / 2 | 11 | 5 | 2.20× |
| 8 / 3 | 23 | 7 | 3.29× |
| 12 / 4 | 33 | 9 | 3.67× |

This is **not** a third-order dual-number implementation and introduces no new PK formula. It
uses mixed-partial symmetry to pick a cheaper FD direction for a quantity already computed by FD.

## Why this is worth more now than when the plan was written

#1335 made the analytic route the default for the exact anchor, and the measured provider time on
that route is most of the fit on an ODE model (1.03s of a ~1.25s `warfarin_ode_laplace` fit).
This sweep is what that time is. Cutting 13 → 7 evaluations attacks the dominant cost of a
now-default path.

## Implementation notes

- **Leave the unfilled blocks empty (`len 0`), not zero-filled.** `subject_h_inner_dx` already
  length-checks the four blocks it reads (`:224-227`); an empty `d2f_dtheta2` makes any future
  consumer that wrongly reaches for it fail its own length check loudly instead of silently
  reading a tensor of zeros. This is the whole safety argument for a partially-populated result.
- **Symmetrize the η indices of `U` after assembly.** The η sweep fills `(k, c)` for every swept
  axis `c`, so each `(k,l)` pair is written twice from two different FD estimates; they differ at
  FD-error level. Average them rather than letting the last write win.
- Keep every existing gate: the `model_supported` scope predicate, the
  closed-form/ODE route-switch guard at each perturbed point (`ode_at(&tp,&ep) != base_is_ode →
  return None`), the `third_order_fd_step` / `jet_noise` step policy, and the all-or-nothing
  `None` contract.
- Do **not** change `covariance_sensitivities`' behaviour for the covariance caller. The cheapest
  shape is an axis-selection parameter with the covariance path passing "all axes", so there is
  one sweep implementation rather than two.
- A later refinement could reuse a compatible base evaluation from the score path, removing the
  remaining duplicate base call — but only under identical parameters, subject/class, derivative
  order and solver context.

## Oracles

The plan frames this as blocked on a reconverged objective-gradient oracle it then disowns (~42%
disagreement at default ODE tolerances, ~0.065% only at `reltol 1e-10 / abstol 1e-12`). **That
oracle is not needed for this change**, and relying on it would be a mistake:

1. **Tensor-vs-tensor.** Run both sweeps on the same subject/point and assert the η-only `T` and
   `U` match the full sweep's `d3f_deta3` / `d3f_deta2_dtheta` to FD tolerance. This is a direct
   comparison of the quantity being changed against the code being replaced — the strongest and
   cheapest oracle available, and it is exact up to the two FD directions' own truncation error.
2. **End-to-end, already committed.** `agq::tests::analytic_grid_response_matches_the_fd_route`
   compares the assembled response against the independent `fd_grid_response` route for both
   anchors × `n_agq ∈ {1,3}` at `1e-5` relative. If the tensors are right this stays green, and
   it is an *independent* route rather than a second copy of the same formula.
3. `laplace_h_deriv`'s own `h_derivative_matches_fd` / `h_derivative_matches_fd_under_ode`.

Measure the realised tensor discrepancy and put the number in the test, per the repo's
tolerance rule — do not pick a bound from a story about FD error.

## Scope to preserve

`subject_h_inner_dx`'s existing restrictions stay until individually validated: unsupported
IOV/residual/censoring/custom-magnitude cases, the conditioning/jitter fallback, and the
analytic-vs-ODE dispatch checks at both perturbed points. Where the selective route cannot
establish a valid result it must return `None` and leave the existing FD grid response to run —
which, after #1335, is still the live fallback.

hiddevandebeek:
## Why

Delivery step 2 of `plans/parallel-optimization-implementation-2026-09-10.md` — the plan's
strongest finding. Stacked on #1336, which made this route the **default**, so the saving applies
to every `method = laplace` fit rather than an opt-in one.

`laplace_h_deriv::subject_h_inner_dx` took its third-order blocks from
`subject_sensitivities_cov`, which central-differences the `Dual2` jet along every θ **and** η
axis — `1 + 2(n_theta + n_eta)` evaluations per subject per sweep. The assembly reads
`∂³f/∂η³` and `∂³f/∂η²∂θ` and nothing else (`laplace_h_deriv.rs:224-227, 274, 294, 312, 316`).
The θ axes exist to build `d2f_dtheta2` and `d3f_deta_dtheta2`, which only the covariance
assembly consumes. So `2·n_theta` full jets were computed per subject, per sweep, and discarded.

## What changed

`∂³f/∂η²∂θ` is reachable from either direction. The full sweep takes `∂/∂θ_m` of `∂²f/∂η_k∂η_l`;
this takes `∂/∂η_l` of `∂²f/∂η_k∂θ_m`, which every base evaluation already returns. Same mixed
partial by Clairaut — only the FD direction differs, and that is what removes the θ axes:

**`1 + 2·n_eta` evaluations.** 13 → 7 at 3 θ / 3 η; 23 → 7 at 8 θ / 3 η; 33 → 9 at 12 θ / 4 η.

`covariance_sensitivities` gains a `ThirdOrderAxes` selector, so there is one sweep
implementation rather than two; the covariance caller passes `All` and is behaviourally
unchanged.

Two details that are load-bearing rather than tidy:

- **The θ-swept blocks come back empty, not zero-filled.** Every consumer length-checks the
  blocks it reads, so an empty vector makes a caller that wrongly reaches for one fail its own
  check loudly, instead of contracting a tensor of zeros into a plausible wrong Hessian. That is
  the entire safety argument for returning a partially-populated `SubjectSens`, and the test
  asserts it both ways — empty under `EtaOnly`, still filled under `All`.
- **Off-diagonal η pairs are written twice and averaged**, from two FD estimates of one quantity;
  that averaging is also the symmetrisation the exact tensor requires. The hit count needs no
  tally: index `(a,b)` is hit once when `a == b` and exactly twice otherwise.

## Measured

5 interleaved reps, `ci-test`, `RAYON_NUM_THREADS=1`, via the existing
`plans/laplace-sensitivity-speed/bench.sh`. **Against the finite-difference route that was the
default before #1335** — i.e. what a user gets from this stack:

| fixture | provider calls | provider time | outer iters |
|---|---|---|---|
| diagonal Ω, closed-form | 6840 → 4180 (**−39%**) | 0.0535 → 0.0345s (**−36%**) | 37 → 37 |
| block Ω (3×3) | 11520 → 5280 (**−54%**) | 0.0828 → 0.0424s (**−49%**) | 47 → 47 |
| ODE | 10270 → 4300 (**−58%**) | 1.624 → 0.669s (**−59%**) | 56 → 38 |

**ODE whole-fit wall clock: 1.90s → 1.18s, −38%.** Provider work is 85% of that fit, so this is a
genuine fit-level figure rather than an extrapolation from a ratio.

Isolating this PR from #1336 (both arms analytic, only the axis set differing):

| fixture | calls | provider time |
|---|---|---|
| diagonal Ω | 6460 → 4180 (−35%) | 0.0556 → 0.0345s (−38%) |
| block Ω | 8160 → 5280 (−35%) | 0.0670 → 0.0424s (−37%) |
| ODE | 6470 → 4300 (−34%) | 1.029 → 0.669s (−35%) |

A side effect worth noting: this moves the **diagonal** fixture off the parity that made #1335's
default decision unresolvable across two benchmark sessions. The θ axes were exactly why
`1 + 2(n_theta + n_eta) = 13` sat against `2·n_free = 14` there. It is now 7 against 14 — **−36%**
rather than −5%.

**Control:** all three `focei` fixtures returned byte-identical numbers to the previous session
(11780 / 13020 / 11790 calls, same OFVs, same iterations). The Gauss-Newton anchor does not route
through `laplace_h_deriv`, and the measurements confirm it.

## Breaking changes

- [x] None — no API, file-format or field change. `ThirdOrderAxes` and the new wrapper are
      `pub(crate)`; the public-API baseline is unchanged.

## Numerical validation

**This one is not bit-identical, unlike #1334 and #1336, and the PR does not claim to be.** The
two FD directions differ at ~1e-10, which on an ODE model is enough to perturb the optimizer
path: the ODE fixture converges in **38** outer iterations instead of 37 and its OFV moves by
**2e-4 (7e-7 relative)**, well inside the convergence tolerance. Both closed-form fixtures were
unaffected — identical OFV and identical iteration counts.

The oracle is the code being replaced, not a second transcription of the formula.
`the_eta_only_sweep_matches_the_full_sweep_on_the_blocks_laplace_reads` runs **both production
sweeps** at one point and asserts:

1. `d3f_deta3` **bit-identical** — it is differenced along η by both variants, so anything else
   would mean the η axes are being swept with a different step or offset;
2. `d3f_deta2_dtheta` to a **measured 1.126e-10**, bound at `1e-8` (~90× headroom). Deliberately
   *not* the `1e-5` that `third_order_sensitivities_layouts_and_values` uses for the arithmetic
   form of the same comparison — that would be five orders too loose here and would pass on a
   sweep that had quietly lost most of its precision. It can be this tight because both sides are
   first central differences of *exact* `Dual2` quantities, so the discrepancy is the difference
   of two small truncation errors rather than either one of them;
3. the base jet is identical, the θ-swept blocks are empty, **and the full sweep still fills
   them** — so the test cannot pass by having broken the covariance caller too.

A pleasing corroboration: `third_order_sensitivities_layouts_and_values` already contained this
exact substitution as an *independent check* on the shipped route (its point 2: "the shipped code
reaches it by differencing `∂²f/∂η²` along `θ`; here it is recomputed by differencing
`∂²f/∂η∂θ` along `η`"). The repo had validated the arithmetic long before anyone used it to save
work, which is why no new derivation was needed.

All ten `laplace_h_deriv` FD oracles pass, including `h_derivative_matches_fd_under_ode`,
`_under_block_omega` and `_matches_central_difference_of_the_anchor`.

- [x] Compared against: the full sweep it replaces, same binary, both production code paths.
- [ ] NONMEM — not applicable. No new estimator, error model or structural model; this changes
      the FD *direction* used for a tensor already anchored by the committed layout test, by the
      module's own FD oracles, and end-to-end by `analytic_grid_response_matches_the_fd_route`.

## Performance

- [x] Faster — the table above. Provider time, which `bench.sh` specifies as the deterministic
      primary metric, plus one whole-fit wall-clock figure on the fixture large enough to carry
      one.

## Tests

- [x] Unit test added; regression verified by the oracle above.
- [x] `cargo test --lib`: **4360 passed, 0 failed**, 23 ignored.
- [x] `tools/preflight.sh` green — fmt, check, clippy, rustdoc, docs, public-api.

## Docs

- [x] `docs/estimation/agq.qmd` — the per-anchor cost discussion now carries the `1 + 2·n_eta`
      figure and the diagonal-fixture consequence.
- [x] `CHANGELOG.md` → `Changed` (not `Performance` alone), because the ODE iteration count and
      OFV move.
- [x] `src/estimation/laplace_h_deriv.rs` module doc — "where the third order comes from" now
      describes the η-only sweep and why it is exact.

## Reviewer hints

The substitution is one identity: `∂³f/∂η_k∂η_l∂θ_m` via `∂/∂η_l` of `∂²f/∂η_k∂θ_m` rather than
`∂/∂θ_m` of `∂²f/∂η_k∂η_l`. Check that, then check the index arithmetic in the `eta_only` arm of
the accumulation loop — `(k*n_eta + cm)*n_theta + mm` and its mirror — since a transposed η pair
is the one error that would survive with plausible finite values. The averaging factor and the
empty-vs-zeroed choice are the other two things worth a look.

## Open questions

The remaining duplicate base evaluation (7 → 6, a further ~14% of the sweep) is **not** in this
PR. It is reachable: `agq::score_core_at` already computes
`subject_sensitivities(model, subject, &params.theta, b_hat)` for the anchor and discards it, and
that is the identical call this sweep makes for its base. Threading it through needs
`anchor_hessian` to return the jet and three signatures to change, and a mismatched base would
silently poison all three tensors rather than failing — so it wants its own PR with a debug
assertion that the threaded base equals a fresh one. Tracked in #1342's follow-ups.


hiddevandebeek:
Reviewed commit `45564cda97eaf61f6844e914d8c0a20f8a8dbb15` against the stacked base `dab78a523ab332d18244b32df2c3d41ec15f7014`.

No actionable findings in the five-file diff.

I checked the eta-axis offset, mixed-tensor indexing, and off-diagonal accumulation/averaging. The new sweep differentiates the eta/theta block into the layout consumed by `subject_h_inner_dx`; diagonal entries are written once and off-diagonal entries receive the two estimates before averaging. The `All` path retains its original axes, steps, and assembly. I also traced the Laplace consumer and `score_core`: neither needs the two blocks now left empty, and the required blocks retain their length checks. The provider scope gates and route-switch checks remain in the shared implementation.

The added regression compares both production sweeps, including the empty-versus-populated blocks. Existing tests exercise the downstream Hessian derivative under one eta, block omega, combined error, and ODE models, plus analytic-versus-FD grid response.

Validation limitation: I inspected those tests but could not complete a local run. I stopped `cargo test --lib --profile ci-fast eta_only -- --nocapture` during prolonged compilation in an isolated worktree; no local test pass or independent performance reproduction is claimed.

hiddevandebeek:
Resolved the conflicts with current `main` in `14059052`, preserving #1336's review corrections to the benchmark claims and default-route documentation alongside this PR's eta-only sweep. Removed the duplicate older changelog entry introduced by the stacked merge. Since #1336 has merged, this PR now targets `main`.

The resulting diff against `main` remains confined to the original five files. Local `cargo fmt --all -- --check` and `git diff --check` passed. The focused local test build was killed with SIGKILL before tests ran; no local test pass is claimed.

Normal PR CI is running on `1c299e098b4ccc2f8e0767379cf6768079cdd0fd`: https://github.com/FeRx-NLME/ferx-core/actions/runs/34592508482

The earlier manual full-suite run failed only `frem_covariate_omega_matches_sample_variance_under_focei` (OFV 3345.8 versus the `<1000` assertion). This is outside the changed path: the eta-only provider explicitly declines FREM, the test uses FOCEI, and the `All` covariance sweep retains its original arithmetic. I have not changed or weakened that test; the broader FREM convergence issue is already tracked in #990, though that issue's reported platform/result differs from this run.

Issue #1345, opened 2026-09-11
perf(estimation): optimization candidates for the classical estimation path (FO/FOCE/FOCEI, Laplace/AGQ, SAEM, IMP)


hiddevandebeek:
A fresh pass over the classical estimation path (FO / FOCE / FOCEI, Laplace, AGQ, SAEM, IMP /
IMPMAP), written after #1329 / #1333 / #1335 / #1342 took ODE `laplace` provider time from 1.62s
to 0.67s per fit. These are *candidates*, not a plan — each needs the profile or the oracle named
against it before it is worth writing.

**Baseline for judging any of them.** On the `warfarin_ode_laplace` fixture, post-#1342: 1.18s
wall, 0.67s provider (~57%). Provider work is still the single largest share, so an item that
removes provider evaluations outranks one that parallelises something else. Measure with
`plans/laplace-sensitivity-speed/bench.sh` + `FERX_PROFILE=1`, whose call count is deterministic
and noise-free.

---

## A. FO / FOCE / FOCEI

**A1. The inner loop's per-subject evaluation count.** `run_inner_loop_warm` warm-starts η̂ from
the previous outer iteration, which is the big win and is already there. What is not measured is
how many inner iterations a converged fit actually spends near the end, when η̂ barely moves.
#216 ("lower the warm-started inner-loop iteration caps") is open and unquantified. Instrument
inner iterations per outer iteration across a fit before changing a cap — the failure mode is a
silently under-converged EBE, which biases the OFV rather than slowing it.

**A2. `Dual2<M>` width on wide-θ models.** The jet carries an `M × M` Hessian with
`M = n_theta + n_eta`. FOCE/FOCEI's inner gradient needs only the η block; the θ columns are
carried through every PK kernel and discarded for subjects whose coordinates are fixed. A
η-only `M = n_eta` instantiation for the inner loop would shrink every dual operation
quadratically. This is the same *shape* of win as #1342 but inside the dual type rather than
around it, and it is a large change: `PkNum` is generic over the width, so it means a second
instantiation through all of `sens/` and `pk/`. Quantify first by counting how much of a
FOCEI fit is inner-gradient versus outer-gradient.

**A3. `ruv_obs_mult` and `theta_dx_chain` hoisting.** Both are recomputed per subject inside
gradient loops and depend only on `(θ, model)`. Cheap to check with a counter; likely small.

---

## B. Laplace / AGQ

[...]

Issue #1344, opened 2026-09-11
perf: remaining parallel-optimization follow-ups (duplicate base jet, Bayes chains, AGQ fusion, SAEM kappa phase, FOCE preparation)


hiddevandebeek:
Follow-ups to the parallel-optimization plan after #1329 (steps 1 & 4), #1333 (step 3), #1335
(analytic default) and #1342 (step 2, the η-only sweep). Each item below has been traced to the
code; the point of this issue is that the *locating* is done, so whoever picks one starts from a
verified premise rather than re-deriving it.

Current measured baseline, for judging whether any of these is worth it: an ODE `laplace` fit is
**1.18s wall, 0.67s of it provider time (~57%)** after #1342, down from 1.90s / 1.62s (~85%).
Provider work is still the largest single share, so items that reduce it outrank items that
parallelise something else.

---

## 2. The duplicate base evaluation — `7 → 6` on a 3-η model (~14% of the sweep)

**Verified.** `agq::score_core_at` computes

```rust
let sens = subject_sensitivities(model, subject, &params.theta, b_hat)?;   // non-IOV arm
```

for the anchor, hands it to `score_core`, and keeps only the returned `ScoreCore`. That is the
*identical* call `covariance_sensitivities` makes for its own base (`evaluate(theta, eta)` with
`iov = false`). So the base jet is already in hand at the right point, and the sweep recomputes it.

**Why it is not a quick win.** The jet has to travel `score_core_at` → `anchor_hessian` →
`grid_response_correction` → `subject_h_inner_dx` → `covariance_sensitivities`.
`anchor_hessian` has **9 call sites**, so either they all change or the gradient path gets a
`anchor_hessian_with_jet` variant. And the precondition is *identity*: a base taken at a different
`theta`/`b_hat`/class/solver context would not fail, it would silently produce wrong `T` and `U`
and therefore a wrong `dH/dx`.

**Do it with:** a `base: Option<&SubjectSens>` parameter on the sweep, plus a `debug_assert` that
the supplied base's `f`/`df_deta`/`d2f_deta2` match a freshly evaluated one, and a test that the
threaded path is bit-identical to the unthreaded path. Quantify with
`plans/laplace-sensitivity-speed/bench.sh` — the call count should fall ~1/7 on the laplace
fixtures and is directly readable from `FERX_PROFILE=1`.

---

## 3. Bayesian chain parallelism — the largest remaining real-workload win

[...]

PR #1346, opened 2026-09-11
perf(sens,saem): reuse the anchor's base jet; parallelise the SAEM kappa phase [#1344]

hiddevandebeek:
## Why

Items **2** and **5** of #1344. Stacked on #1343.

**Item 2.** `agq::score_core_at` builds the `Exact` anchor from
`subject_sensitivities(model, subject, theta, b_hat)` and dropped the jet. `laplace_h_deriv`'s
sweep then made the *identical* call as its own base — one wasted evaluation per subject per
sweep, on a route that is now the default (#1336).

**Item 5.** The per-occasion κ MH loop in SAEM was serial, immediately beside an
already-parallel η phase, with a comment saying it was serial because the κ MH is "cheap
(low-dimensional, analytical PK) and share-free".

## What changed

**Item 2 — `7 → 6` evaluations on a 3-η model.** `score_core_at` returns the jet alongside the
`ScoreCore`; a thin `anchor_hessian_and_base_jet` surfaces it on the single path that can use it;
it threads through `grid_response_correction` → `analytic_grid_response` → `subject_h_inner_dx` →
`covariance_sensitivities`, which gained a `base_in: Option<SubjectSens>`.

Three things make this safe rather than merely plausible:

- **A wrong base does not fail** — it returns plausible, wrong third-order tensors and therefore
  a wrong `dH/dx`. So `debug_assert_base_matches` compares the supplied jet against a fresh
  evaluation (value and both second-order blocks, per observation) in debug builds, which is
  where the test suite runs. In release it is the optimisation; in debug it pays the evaluation it
  saves in order to catch a caller that threaded a jet from the wrong point.
- **IOV is excluded.** `score_core_at` builds the stacked `(η, κ)` jet under IOV, which is not
  what a `covariance_sensitivities(iov = false)` sweep wants. The jet is offered only for the
  non-IOV `Exact` case — and `subject_h_inner_dx` declines IOV regardless.
- **`anchor_hessian_and_base_jet` is a second *spelling*, not a second implementation.** Both
  arms go through `score_core_at`, so the matrix is the one `anchor_hessian` would return, and
  anything that is not "`Exact`, non-IOV, in scope" falls straight through to it. That keeps the
  other eight `anchor_hessian` call sites untouched.

**Item 5 — κ phase parallel, bit-identical.** Not "equal to tolerance" — bit-identical, because
every access is `[i]`-disjoint (reads `etas[i]`, `kappas[i]`, `kappa_step_scales[i]`; writes
`kappas[i]`, `nll_cache[i]` and the two counters) and each subject's RNG is seeded from
`(master_seed, k, i)`, so no subject's draws depend on the order subjects are visited in.

Shaped as map-then-apply rather than a six-way `par_iter_mut().zip(..)`: the MH both reads and
writes `kappas[i]`, so the mutable shape needs the state destructured into disjoint slices and
threaded through nested tuples, which is unreadable at six fields. Cost is one clone of
`kappas[i]` (n_occ × n_kappa f64s) per subject per iteration, against a full
`individual_nll_iov` per proposal. Counters are applied **in subject order**, so they accumulate
identically however the workers finished.

## Item 4 is deliberately not here

Its premise does not hold in this codebase. `outer_optimizer.rs:450` computes `eta_hats` **once**
via `run_inner_loop_warm` and shares them across the objective, the gradient and post-fit; and
`agq_population_nll` is already `par_iter` over subjects with a deliberately serial, order-stable
reduction (#703, so the OFV does not depend on thread count). Fusing EBE computation into the AGQ
subject task would force a recompute for the gradient — a regression, not a win. Recorded in
#1344 rather than built.

## Breaking changes

- [x] None. `ThirdOrderAxes`, the new wrapper and `anchor_hessian_and_base_jet` are all private
      or `pub(crate)`; the public-API baseline is unchanged.

## Numerical validation

- **Item 5 is bit-identical by construction**, and the argument is structural (index-disjoint
  writes, order-independent seeding) rather than empirical. 108 SAEM tests pass unchanged.
- **Item 2 is bit-identical when the assertion holds**, which the suite verifies on every
  `laplace` fixture it touches: `cargo test --lib` ran **with `debug_assert_base_matches` live**,
  so every threaded base was compared against a fresh evaluation. Had any been from the wrong
  point, the suite would have aborted rather than passed.
- `cargo test --lib`: **4360 passed, 0 failed**, 23 ignored — the same count as #1343, i.e. no
  test changed behaviour.
- [ ] NONMEM — not applicable. Item 2 removes a redundant evaluation of a quantity that is
      asserted identical; item 5 changes scheduling only.

## Performance

- [x] Item 2: one of `1 + 2·n_eta` evaluations removed — **7 → 6** at 3 η, **9 → 8** at 4 η.
      **Measured**, `FERX_PROFILE=1` on `--profile ci-test`, against the pre-item-2 baseline
      (i.e. #1343's numbers): diagonal Ω `warfarin_laplace.ferx` 4180 → **3800** calls, block Ω
      `warfarin_block_omega_laplace.ferx` 5280 → **4800**, ODE `warfarin_ode_laplace.ferx`
      4300 → **3910** — all three **−9.1%**, OFV and outer-iteration counts unchanged in every
      case. (The first attempt at this measured **zero** reduction: `debug_assert_base_matches`
      is a plain `fn`, so only its body compiled away in release, not its
      `&evaluate(theta, eta)?` argument — the caller still paid for that call on every build.
      Fixed by gating the evaluation itself on `cfg!(debug_assertions)`.)
- [ ] Item 5: **unmeasured, and stated as such at the call site.** Nobody has profiled the κ
      phase's share of a SAEM IOV fit. The previous comment's claim that it is "cheap" may well be
      right; this removes a known serialisation so that the question can be answered by
      measurement rather than by a comment. If it profiles at a few percent it should arguably be
      reverted, and the code says so.

## Tests

- [x] `cargo test --lib`: 4360 passed, 0 failed.
- [x] `tools/preflight.sh` green — fmt, check, clippy, rustdoc, docs, public-api.
- Item 2's regression guard is the debug assertion itself, which is stronger than a single test
  case: it checks every threaded jet on every fixture the suite exercises, rather than one.

## Docs

- [x] `CHANGELOG.md` — two entries, both under `Changed`, with item 5 explicitly labelled as
      removing a serialisation rather than delivering a measured speedup.

## Reviewer hints

For item 2, the thing to check is the **identity precondition**: that
`anchor_hessian_and_base_jet` only offers a jet when the sweep would have computed that same jet.
Read its guard (`Exact`, `!stack.is_iov()`, `analytic_score_supported`) against
`covariance_sensitivities`' `evaluate` closure. The debug assertion is the backstop, not the
argument.

For item 5, check that the map closure touches nothing but `[i]`, and that the apply loop's
`+=` on the two counters is in subject order.

## Open questions

#1345 C2 notes that the κ phase *also* recomputes `individual_nll_iov` once per subject per
iteration purely to re-reference the NLL, with a comment saying it is done so "detailed balance is
guaranteed". The η phase may already leave a comparable value in `nll_cache[i]` — but only when
`omega_iov_for_eta_mh` was `Some`. That is a change inside an MCMC acceptance ratio and wants a
Tier-3 SAEM convergence comparison, so it is analysed in #1345 and not touched here.


hiddevandebeek:
## Measured: item 2 delivers no call-count reduction. Investigating.

Benchmarked on `ci-test` (3 reps, `RAYON_NUM_THREADS=1`), provider calls are **identical** to #1343 before this change:

| fixture | before (#1343) | with item 2 |
|---|---|---|
| diagonal Ω | 4180 | **4180** |
| block Ω 3×3 | 5280 | **5280** |
| ODE | 4300 | **4300** |

Zero calls removed, where the prediction was one per subject per gradient (~12.5% of the sweep).

Ruled out so far:

- **Not the debug assertion.** `[profile.ci-test]` inherits `release`; `debug-assertions = true` is set on `ci-cov`, not here. So `debug_assert_base_matches` is compiled out and is not adding back the evaluation it saves.
- **Not missing wiring.** `anchor_hessian_and_base_jet` is defined (`agq.rs:487`) and called on the gradient path (`agq.rs:2065`), and `base_jet` is forwarded through `grid_response_correction` → `analytic_grid_response` → `subject_h_inner_dx` → `covariance_sensitivities`.

So either the fast-path guard (`Exact` / `!is_iov()` / `analytic_score_supported`) is not being taken on these fixtures, or `subject_h_inner_dx` is declining and the FD route is dropping the jet unused. The latter would contradict #1336's measured gain, so the guard is the likelier candidate — but that needs a diagnostic build to confirm rather than more reading.

**Item 5 (the SAEM κ phase) is unaffected by this** — it is bit-identical by construction and independently verified by 108 SAEM tests; it was never claimed to be a measured win.

Wall-clock numbers from this run are not usable: the box became loaded mid-benchmark (ODE provider time scattered 2.17 / 4.15 / 2.31s against 0.67s earlier). Call counts are deterministic, which is why the null result above is still trustworthy.

Treating item 2 as **not ready** until the guard is traced. Marking this PR draft.

hiddevandebeek:
Reviewed commit `055e6e2f020ecaa6a3d6adf9698be7231127be35` against this PR's base, `worktree-agq-eta-only-sweep`.

**[P2] Gate the fresh evaluation itself behind `debug_assertions`** — [`src/sens/provider.rs:5021`](https://github.com/FeRx-NLME/ferx-core/blob/055e6e2f020ecaa6a3d6adf9698be7231127be35/src/sens/provider.rs#L5021)

On the newly introduced `Some(supplied)` path, `debug_assert_base_matches(&supplied, &evaluate(theta, eta)?)` still evaluates `evaluate(theta, eta)` in release builds. Only the assertions inside the helper are debug-only; its arguments and the `?` are not. Consequently, every eligible exact-anchor AGQ gradient still recomputes the base jet, defeating this PR's advertised reduction from seven evaluations to six for a three-ETA model. The fallible provider call cannot simply be assumed to disappear when the helper is inlined.

Put `#[cfg(debug_assertions)]` on the entire check statement/block (including the fresh evaluation), and gate the helper as appropriate. A release-mode provider-call-count regression would catch this; numerical equality tests alone cannot.

Validation: a standalone reproduction of this call pattern, compiled with `rustc -O -C debug-assertions=no`, confirmed that supplying a base still executes the fresh evaluation once. `git diff --check HEAD^ HEAD` passed. I found no additional actionable defect in the SAEM kappa-loop changes: the per-subject RNG seeds, mixture-class guard, and ordered application of updates are preserved.

Test limitation: `cargo test --lib --offline agq --no-default-features` exited 101 without test results or a useful failure diagnostic in RTK's summary. I stopped a raw-output retry while it was still compiling; the AGQ tests are therefore unverified in this review.

hiddevandebeek:
Root-caused and fixed the null-result finding from the earlier review: `debug_assert_base_matches(&supplied, &evaluate(theta, eta)?)` is a plain function call, so only its *body* compiles away under `debug_assertions` — the caller still evaluates the `evaluate(theta, eta)` argument to construct it on every build, `Some(base_in)` included, paying back the entire evaluation item 2 exists to skip. Fixed by gating the evaluation itself behind `cfg!(debug_assertions)`, not just the assertion it feeds (`src/sens/provider.rs`).

Verified end-to-end:
- `cargo test --lib`: 4399 passed, 0 failed, 23 ignored (full suite, post-merge with the current `worktree-agq-eta-only-sweep` base).
- `cargo fmt --check` / `cargo clippy --lib`: clean on the changed file (the pre-existing clippy warnings elsewhere in the tree are untouched by this diff).
- Real `ci-test`-profile build + `plans/laplace-sensitivity-speed/bench.sh` (`FERX_PROFILE=1`, 3 reps) — the fix now delivers the advertised call-count reduction:
  - `warfarin_laplace.ferx` (diagonal Ω): 4180 → **3800** candidate calls (baseline fd unchanged at 6840), OFV bit-identical.
  - `warfarin_block_omega_laplace.ferx`: 5280 → **4800** (baseline 11520 unchanged), OFV bit-identical.
  - `warfarin_ode_laplace.ferx`: 4300 → **3910** (baseline 10270 unchanged); OFV/iteration difference vs baseline is the pre-existing FD-direction divergence inherited from #1343, unchanged by this fix.
  - FOCEI-AGQ control fixtures (Gauss-Newton anchor, which this guard must not touch): call counts unchanged (11780/13020/11790), confirming the fix is correctly scoped to the `Exact`/Laplace anchor only.

That's roughly a 9% total-call reduction on all three Laplace-anchor fixtures, matching the claimed removal of one of the sweep's `1 + 2·n_eta` evaluations.

Also merged in the current `worktree-agq-eta-only-sweep` base (23 commits) to resolve the `CONFLICTING` merge state — the only real conflict was two independent `CHANGELOG.md` insertions near the same line (plus a pre-existing duplicate line from an earlier stacked-PR merge, which I removed); everything else (`agq.rs`, `laplace_h_deriv.rs`, `provider.rs`, `provider_tests.rs`) auto-merged cleanly. Re-ran the full test suite and the benchmark after the merge to confirm both still hold. PR is now out of draft and shows `MERGEABLE`/`CLEAN`.

hiddevandebeek:
Reviewed head `93a6266612760e64f302bd807a69ef6e6c43edfc` against base `1c299e098b4ccc2f8e0767379cf6768079cdd0fd`.

No blocking correctness findings.

The reused Laplace base jet comes from the same non-IOV provider at the same `(theta, b_hat)` point as the original evaluation. The covariance scope gates and finite-difference fallback remain in place, and the fresh debug evaluation is correctly gated out of release builds. The SAEM κ change preserves subject-specific RNG seeds, proposal order, mixture-class context on the executing worker, and ordered application of updates.

Non-blocking testing suggestion: add explicit regression coverage for `Some(base)` versus `None` and for the new κ map/apply phase versus the former serial calculation, including a mixture/IOV fixture. The modified provider test still supplies `None`; existing debug assertions help validate the jet handoff, but do not establish the κ equivalence claim. A serial-reference comparison can run on one CPU.

Validation: `cargo fmt --check` passed. I attempted the focused AGQ unit tests with CPU affinity pinned to CPU 0, one Cargo build job, one Rayon worker, and one test thread. The debug build did not finish during this review and was stopped, so I cannot report a runtime test pass. No multi-core execution or performance benchmark was performed.

hiddevandebeek:
Addressed the non-blocking testing suggestion from the latest review (thanks @hiddevandebeek):

- `the_supplied_base_jet_reproduces_the_recomputed_one` (`src/sens/provider_tests.rs`) — asserts `subject_sensitivities_cov_eta_only` produces bit-identical `SubjectSens` output whether `base` is `None` (recomputed internally) or `Some(base)` supplied by the caller, confirming the base-jet reuse is a pure optimization.
- `kappa_phase_parallel_schedule_is_bit_identical` (`src/estimation/saem.rs`) — runs `run_saem` on a mixture + IOV fixture across different rayon thread-pool widths (including single-threaded) and asserts the resulting `OuterResult` (theta, ofv, kappas) is bit-identical, confirming the map/apply parallel kappa phase matches the former serial calculation.

Both pass locally (`cargo fmt --check` / `cargo clippy` clean on the touched lines). Pushed to this branch as 5ff40a02.

PR #1348, opened 2026-09-11
perf(estimation): thread inner-loop iteration counts into EbeResult/InnerLoopStats [A1, part of #1345]

hiddevandebeek:
## Why

`HANDOFF-perf-stack.md`'s remaining queue names **A1 — inner-loop iteration
instrumentation** as pure measurement, gated ahead of A2 (an η-only `Dual2`
width change to the inner optimizer): before spending effort narrowing the
inner loop's optimizer cost, we need a profile of where that cost actually
goes — how many optimizer iterations a fit spends, and how much of that is
the main solve vs. BFGS→Nelder-Mead fallback, multi-start restarts, or the
runaway-EBE guard.

## What changed

`lbfgs_core`, `dense_bfgs_core` and `nelder_mead_minimize` each take a new
trailing `iters: &mut u64` accumulator, incremented once per pass of their
`for _iter in 0..max_iter` loop. It threads through `inner_minimize_with_grad`
and `argmin_inner_fallback` so every optimizer call a subject's inner solve
makes — the main solve, a BFGS→NM fallback, an `inner_restarts` seed, or the
runaway-EBE-guard NM recovery — counts toward one running total.

`EbeResult` exposes the per-subject total as `n_iters` (`0` on a hard-reject,
where no optimizer ran). `InnerLoopStats` sums it across subjects as
`total_inner_iters`, populated in `run_inner_loop_warm_map` (and therefore
`run_inner_loop_warm`), the mixture E-step (`mixture.rs`), IMPMAP's
multi-start sweep (`impmap.rs`), and the outer optimizer's mixture branch
(`outer_optimizer.rs`).

Pure measurement — no optimizer behaviour, convergence criterion, or fallback
policy changes; only a counter is added alongside the existing control flow.

## Alternatives considered

Changing each of the three core loop functions' return type from `bool` to a
tuple `(bool, usize)` would have required touching every early `return`
statement in code the repo's own testing guidance calls easy to get subtly
wrong (BFGS/NM convergence and fallback logic). Threading a `&mut u64`
accumulator instead only touches each function's signature and the single
loop-opening line, minimizing the diff's blast radius on that logic.

## Cross-repo dependency

## Breaking changes

- [x] Public Rust API (`api/`, `types.rs`) — regenerated `api/ferx-core-public-api.txt`
      via `tools/update-public-api.sh`. Two additive fields
      (`EbeResult::n_iters`, `InnerLoopStats::total_inner_iters`); both are plain
      data already reachable in spirit through `find_ebe`'s existing return
      value, just not previously counted.

## Numerical validation

- [x] Not applicable — pure instrumentation, no change to any objective,
      gradient, or convergence criterion. `cargo test --lib estimation::inner_optimizer`
      (86 tests) is unchanged in outcome from before this PR, only in count
      (+2 new tests below).

## Performance

- [x] No significant impact expected — one `u64` increment per existing loop
      iteration.

## Tests

- [x] Regression test added (fails without the fix):
  - `find_ebe_n_iters_tracks_actual_optimizer_work` — a cold-started solve
    must report `n_iters > 0`, and a warm start from the already-converged
    mode must report fewer iterations than the cold search that found it;
    catches `n_iters` being left at 0 or hardcoded to `max_iter`.
  - `argmin_inner_fallback_iters_sum_both_nm_runs_when_polishing` — pins the
    fallback's "polish" branch (which runs Nelder-Mead twice) to the *exact*
    sum of two independently-run, identically-seeded `nelder_mead_minimize`
    calls; catches a mutation that threads a throwaway counter into either
    internal call instead of the caller's `iters`.

## Docs

## Checklist

- [x] `tools/preflight.sh` green — fmt, check, clippy, public-API baseline
      (ran as targeted groups: `fmt check clippy` and `public-api`)

## Follow-up: BFGS/Nelder-Mead split, multistart undercount fix

The single `u64` total above conflates BFGS/L-BFGS iterations (each a
gradient plus a backtracking line search) with Nelder-Mead iterations
(1-2 bare objective evals against a `max_iter * 5` budget) — summed
blind, a subject that falls back to NM can out-"iterate" a subject
whose main solve did far more real work. That made the counter useless
for actually answering A1's question (where FOCEI's inner-loop cost
goes) ahead of A2.

- `InnerIterCounts { bfgs: u64, nelder_mead: u64 }` replaces the plain
  `u64` accumulator everywhere it was threaded (`lbfgs_core`,
  `dense_bfgs_core`, `nelder_mead_minimize`, `inner_minimize_with_grad`,
  `argmin_inner_fallback`, `find_ebe`, `find_ebe_iov`). `EbeResult::n_iters`
  stays a single `u64` (`.total()`), so `InnerLoopStats::total_inner_iters`
  is unchanged in shape.
- New process-global `PROFILE_INNER_BFGS_ITERS` / `PROFILE_INNER_NM_ITERS`
  atomics, folded unconditionally (not gated behind `FERX_PROFILE`) via
  `record_inner_iters`, so they also see the AGQ / covariance /
  importance-sampling / VI callers that call `find_ebe` directly and
  discard the `EbeResult` — the one gap `InnerLoopStats::total_inner_iters`
  has (documented on its doc comment). `profile_report()` gains an
  `[profile] inner iterations: N BFGS/L-BFGS, N Nelder-Mead (X per solve)`
  line.
- Fixed `find_ebe_multistart`: a winning MCETA start replaced `best`
  wholesale (`best = candidate`), silently discarding every other
  start's `n_iters`. Now accumulates across all `mceta + 1` starts.

New tests: `find_ebe_multistart_n_iters_counts_every_start` (reconstructs
all starts independently and asserts the sum), `inner_iter_counts_keep_bfgs_and_nelder_mead_apart`,
`find_ebe_feeds_the_fit_wide_profile_counters`. Also serializes 5
pre-existing tests around the process-global `EBE_WARM_START` flag with a
`Mutex` guard (`lock_ebe_warm_start`) after this diff's iteration-count
tests made unguarded concurrent flag flips a real (not just theoretical)
source of flaky exact-count assertions.

Validated: `cargo check --lib` (whole crate, all callers of the changed
signatures), `cargo test --lib estimation::inner` (88 passed) and
`cargo test --lib impmap` (32 passed), plus `tools/preflight.sh`'s
`fmt`, `check` (all 5 groups), `clippy` (both groups), `public-api`, and
`rustdoc` groups — all green. No `pub` surface changed (`InnerIterCounts`
is `pub(crate)`), confirmed by the unchanged baseline diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ section of the SDLC for the versioning policy).
 
 ## [Unreleased]
 
+### Added
+
+- `EbeResult` now carries `n_iters`, the optimizer iterations spent on that subject's inner solve
+  (main BFGS/L-BFGS/Nelder-Mead plus any fallback/restart/runaway-guard Nelder-Mead runs), and
+  `InnerLoopStats` sums it across subjects as `total_inner_iters` — pure measurement, to profile
+  where inner-loop time goes before optimizing it further (#1345).
+
 ### Changed
 
 - **SAEM now averages the residual sufficient statistic for eligible single additive and proportional error models, reducing final-draw Monte Carlo noise in the residual SD estimate (#1321).**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,14 @@ section of the SDLC for the versioning policy).
 ### Added
 
 - `EbeResult` now carries `n_iters`, the optimizer iterations spent on that subject's inner solve
-  (main BFGS/L-BFGS/Nelder-Mead plus any fallback/restart/runaway-guard Nelder-Mead runs), and
-  `InnerLoopStats` sums it across subjects as `total_inner_iters` — pure measurement, to profile
-  where inner-loop time goes before optimizing it further (#1345).
+  (main BFGS/L-BFGS/Nelder-Mead plus any fallback/restart/runaway-guard Nelder-Mead runs, and every
+  MCETA multi-start rather than only the winning one), and `InnerLoopStats` sums it across subjects
+  as `total_inner_iters` — pure measurement, to profile where inner-loop time goes before optimizing
+  it further (#1345).
+- `FERX_PROFILE=1` now reports fit-wide inner-optimizer iteration totals, split into BFGS/L-BFGS and
+  Nelder-Mead: the two are different units of work, and unlike the per-evaluation `InnerLoopStats`
+  the totals accumulate across a whole fit, including the AGQ / covariance / importance-sampling
+  callers that discard their `EbeResult` (#1345).
 
 ### Changed
 
--- a/api/ferx-core-public-api.txt
+++ b/api/ferx-core-public-api.txt
@@ -381,12 +381,14 @@ pub ferx_core::estimation::inner_optimizer::EbeResult::grad_norm: f64
 pub ferx_core::estimation::inner_optimizer::EbeResult::h_matrix: nalgebra::base::alias::DMatrix<f64>
 pub ferx_core::estimation::inner_optimizer::EbeResult::hard_reject: bool
 pub ferx_core::estimation::inner_optimizer::EbeResult::kappas: alloc::vec::Vec<nalgebra::base::alias::DVector<f64>>
+pub ferx_core::estimation::inner_optimizer::EbeResult::n_iters: u64
 pub ferx_core::estimation::inner_optimizer::EbeResult::nll: f64
 pub ferx_core::estimation::inner_optimizer::EbeResult::used_fallback: bool
 pub struct ferx_core::estimation::inner_optimizer::InnerLoopStats
 pub ferx_core::estimation::inner_optimizer::InnerLoopStats::n_fallback: usize
 pub ferx_core::estimation::inner_optimizer::InnerLoopStats::n_start_rejected: usize
 pub ferx_core::estimation::inner_optimizer::InnerLoopStats::n_unconverged: usize
+pub ferx_core::estimation::inner_optimizer::InnerLoopStats::total_inner_iters: u64
 pub const ferx_core::estimation::inner_optimizer::INNER_LBFGS_MIN_DIM: usize
 pub static ferx_core::estimation::inner_optimizer::PROFILE_INNER_ANALYTIC_GRAD: core::sync::atomic::AtomicU64
 pub static ferx_core::estimation::inner_optimizer::PROFILE_INNER_FD_FALLBACK: core::sync::atomic::AtomicU64
--- a/src/estimation/impmap.rs
+++ b/src/estimation/impmap.rs
@@ -370,6 +370,7 @@ fn run_map_multistart(
         n_unconverged: results.iter().filter(|r| !r.converged).count(),
         n_fallback: results.iter().filter(|r| r.used_fallback).count(),
         n_start_rejected: results.iter().filter(|r| r.hard_reject).count(),
+        total_inner_iters: results.iter().map(|r| r.n_iters).sum(),
     };
     let eta_hats: Vec<DVector<f64>> = results.iter().map(|r| r.eta.clone()).collect();
     let h_matrices: Vec<DMatrix<f64>> = results.iter().map(|r| r.h_matrix.clone()).collect();
--- a/src/estimation/impmap.rs
+++ b/src/estimation/impmap.rs
@@ -284,6 +284,12 @@ fn find_ebe_multistart(
     let Some(l_omega) = omega_chol else {
         return best;
     };
+    // `n_iters` is the one field that must survive a losing start: every MCETA start is
+    // work the fit actually paid for, but `best = candidate` below replaces the result
+    // wholesale, so without this the reported total is one start's count out of
+    // `mceta + 1` — an `mceta`-fold undercount on exactly the multimodal subjects
+    // multi-start exists for.
+    let mut spent = best.n_iters;
     let n_eta = l_omega.nrows();
     let mut rng = StdRng::seed_from_u64(subj_seed);
     for _start in 0..mceta {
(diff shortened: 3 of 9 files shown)


hiddevandebeek:

Mechanically the threading is complete and faithful: every call site of `lbfgs_core`,
`dense_bfgs_core`, `nelder_mead_minimize`, `inner_minimize_with_grad` and
`argmin_inner_fallback` on the head passes the caller's accumulator (checked by grep on the
head revision, no throwaway counters), the increments sit at the top of the single
`for _iter in 0..max_iter` loop in all three cores, no control flow changed, and all four
`InnerLoopStats` literals plus both `EbeResult` literals are updated. The API baseline lines
are in the right sort position. So I have no behavioural-regression findings.

What I do have is about the *number* — since this PR is pure measurement, the number's
fidelity is the whole deliverable.

### 1. `src/estimation/impmap.rs:307` — multi-start throws away every losing start's iterations (medium)

`find_ebe_multistart` keeps the winner wholesale (`best = candidate`), so the returned
`EbeResult.n_iters` is *only the winning start's* count. `run_map_multistart` then sums those
into `total_inner_iters`. With `impmap_mceta = 10` each subject runs 11 complete inner solves
and the reported total counts one of them — an ~11× undercount on exactly the configuration
where the inner loop is most expensive (MCETA is switched on for multimodal subjects, i.e.
the case A2 is meant to reason about). It is also the one place where the discarded work is
*guaranteed* non-zero, unlike the fallback paths. Fix is one line at the merge point:

```rust
let mut spent = best.n_iters;
...
spent += candidate.n_iters;
if candidate.nll < best.nll { best = candidate; }
...
best.n_iters = spent;
best
```

### 2. `src/estimation/outer_optimizer.rs:1897` — nothing anywhere reads `total_inner_iters` (medium)

Grepping the workspace, the only reads of the new fields are the four aggregation sites that
write them; no printer, no `FERX_PROFILE` counter, no `FitResult` field, no test outside the
two new ones. On the FOCE path `InnerLoopStats` is produced per *outer evaluation* and dropped
after the EBE guard consults `n_unconverged`/`n_fallback`/`n_start_rejected`, and nothing sums
across evaluations — so "how many optimizer iterations a fit spends", the question the PR body
opens with, still cannot be answered without another patch. Compare `PROFILE_INNER_SOLVES` /
`PROFILE_INNER_ANALYTIC_GRAD` a few lines up in the same file: those are process-global
`AtomicU64`s precisely because a per-call struct cannot accumulate over a fit. As it stands
this widens the public API (which CLAUDE.md asks to justify by a concrete consumer) without
one, in-tree or out (`ferx-tools` and `ferx-r` reference neither struct today).

### 3. `src/estimation/inner_optimizer.rs:470` — one scalar cannot give the attribution the PR is for (low)

The Why section asks "how much of that is the main solve vs. BFGS→Nelder-Mead fallback,
multi-start restarts, or the runaway-EBE guard". `n_iters` folds all four into one total, and
`used_fallback` is a bool, so the split is unrecoverable from the recorded data. Worse for a
cost proxy: an NM iteration (1–2 objective evals, budget `max_iter * 5`) is summed with a
BFGS/L-BFGS iteration (one gradient — analytic or `~2·n_eta+1` predictions — plus a
backtracking line search), so a subject that falls back can contribute ~1100 "iterations" that
are cheaper each than the 40 the main solve contributed. The resulting total is not
proportional to time and will mislead the A2 decision it is gating. Separate fields
(`n_iters_main` / `n_iters_fallback`), or a wall-clock accumulator alongside, cost nothing
extra here.

### 4. `src/estimation/impmap.rs:1957` — sibling inner-solve call sites contribute nothing (low)

The mixture MCEM E-step calls `find_ebe_multistart` and discards the `EbeResult` after reading
`eta`/`h_matrix`; `agq.rs` (5 sites), `sens_cov_hessian.rs` (6), `covariance.rs`,
`importance_sampling.rs` and `vi/elbo_agq_bound.rs` call `find_ebe` and likewise drop it. That
is a defensible scope for a first cut, but the field doc ("Sum of `EbeResult::n_iters` across
all subjects, for this call's inner loop") reads like a total over the fit's inner work, and a
reader profiling an AGQ or covariance-step run will get zero and conclude the inner loop is
free. Worth saying explicitly in the doc comment which paths are instrumented.

### 5. `src/estimation/inner_optimizer_iov_tests.rs:3644` — the new `assert_eq!` can flake under the default parallel harness (low)

`argmin_inner_fallback_iters_sum_both_nm_runs_when_polishing` pins an *exact* sum of two NM
runs, but which branch `argmin_inner_fallback` takes depends on `EBE_WARM_START`, a process-
global `AtomicBool` (`inner_optimizer.rs:1690`), not on anything the test owns. Sibling tests
in the same `--lib` binary set it to `true` and run with it held — `argmin_inner_fallback_keeps_better_basin`
(line 1306) and `ebe_warm_start_flag_round_trips` (line 2963) — and `api/fit.rs:311` sets it
from `FitOptions` on every `fit()`. Rust's harness runs these threads concurrently, so a read
of `true` inside the window takes the `if warm { return ... }` early exit, only one NM runs,
and `fallback_iters != expected_iters` fails a test that is green 99 runs out of 100. The
hazard pre-exists (line 2963's "default must be off" assert has the same shape), but an
exact-equality assertion is the most brittle consumer of it yet. Either read
`ebe_warm_start_enabled()` and skip when it is `true`, or serialize the flag-touching tests
behind a shared mutex.

### Nit

Both new `#[allow(clippy::too_many_arguments)]` attributes are no-ops: clippy's default
threshold is 7 and fires only above it, and the two functions now take 7
(`argmin_inner_fallback`) and 6 (`nelder_mead_minimize`) arguments — there is no `clippy.toml`
lowering it. Harmless, but they will read as "this function is over the line" to the next
person.

*Note: I could not execute `cargo test --lib` for the two new tests — the build was
OOM-killed on this box — so findings 1–5 are from reading the head revision, not from a run.*

Issue #309, opened 2026-06-13
Parser: error when a [structural_model] omits a required PK parameter (silently defaults the slot to 0.0)


TeunP:
## Summary

A `[structural_model]` line that **omits a required PK parameter** for the chosen model silently leaves that slot at `PkParams::default()` (`0.0`, except `F`) instead of erroring. The model then fits to a structurally broken optimum — the same silent-default failure mode as #261, but reached by *omission* rather than by referencing an undefined variable.

Surfaced in the self-review of #308 (the PR that fixed #261). #308 made an *undefined reference* (`cl=CL` with no `CL`) and an *unknown key* (`clx=`) hard errors, but it only validates the mappings that are **present** — it does not check that all required params for the model are mapped, nor that mapped params are actually *used*.

## Reproduction

```
[structural_model]
  pk one_cpt_oral(cl=CL, v=V)     # `ka` omitted
```

`ka` never appears in the mapping, so `PK_IDX_KA` stays `0.0`. With `ka = 0` there is no absorption, the central compartment never receives drug, every prediction floors to the log constant, and the fit "converges" to a meaningless optimum — exactly the #261 symptom (`PRED = IPRED = -27.63`, ~100% shrinkage).

## Fix (implemented in PR #310)

Two structural-mapping completeness checks, both added in `parse_full_model` and run **after** `build_pk_param_fn` (so #308's per-key validation — unknown key / undefined reference — reports first; precedence is sequential, not a special-case):

1. **Required params (error).** Add a required-parameter set per `PkModel` (`PkModel::required_pk_params()` in `types.rs`, mirroring the "Required Parameters" table already in `docs/src/model-file/structural-model.md`) and error if any required canonical slot is unmapped:

   ```
   [structural_model] one_cpt_oral requires `ka`, which is not mapped. Add it, e.g. `ka=KA`.
   ```

   Checked at the **canonical-slot** level (via `PkParams::name_to_index`), so the `v`/`v1` and `q`/`q2` aliases satisfy their slot. `f`/`lagtime` (alias `alag`) stay optional — they default to `1.0`/`0.0`.

[...]

Issue #261, opened 2026-06-11
Parser: error on structural-model parameter referencing an undefined variable (silently defaults to 0)


TeunP:
## Summary

A `[structural_model]` PK parameter that references a variable **not defined** in `[individual_parameters]` is silently accepted and defaults to `0.0`, instead of producing a parse error. This yields a plausible-looking *converged* fit from a structurally broken model.

Surfaced while benchmarking against NONMEM in FeRx-NLME/ferx-r#154.

## Reproduction

Model maps `cl=CL`, but `CL` is never defined (only V, KE, KA are):

```
[individual_parameters]
  V  = exp(TVV  + ETA_V)
  KE = exp(TVKE + ETA_KE)
  KA = KE + exp(TVKA + ETA_KA)
  # note: no `CL = ...` line

[structural_model]
  pk one_cpt_oral(cl=CL, v=V, ka=KA)
```

Result: the fit "converges" but every prediction is the log-floor constant (`PRED = IPRED = -27.63` on every row), 100% ETA shrinkage on all etas, and a blown-up residual error. No error or warning is emitted.

## Root cause

In `build_pk_param_fn` (`src/parser/model_parser.rs` ~L5555-5566), the analytical `pk_param_map` bindings are resolved with a `filter_map`. When the referenced variable name isn't found in `var_idx`, the `?` operator **silently drops** the binding:

```rust
let pk_assignment_mapping: Vec<(usize, usize)> = pk_param_map
    .iter()
    .filter_map(|(pk_name, var_name)| {
        let pk_slot = PkParams::name_to_index(pk_name)?;
        let var_slot = var_idx.get(var_name).copied().or_else(|| {
            var_idx.get(&var_name.to_lowercase()).copied()
        })?;          // <-- undefined var: binding silently dropped
        Some((pk_slot, var_slot))
    })
    .collect();
```

The unbound `cl` slot then keeps its `PkParams::default()` value of `0.0` (`src/types.rs` ~L81-87, only `F` defaults to 1.0). Downstream, `one_cpt_oral_f` (`src/pk/one_compartment.rs:43`) early-returns `0.0` whenever `cl <= 0.0`, so concentration is 0 everywhere and the log-prediction floors.

## Proposed fix

Make an undefined structural-model parameter reference a hard parse error. In `build_pk_param_fn`, distinguish "variable not found" from the (legitimate) filtering and return `Err`, e.g.:

```
[...]

PR #1361, opened 2026-09-11
fix(parser): reject two spellings of one PK slot bound to different values [closes #1048]

### LINKED ISSUE #1048: pk(..., v=VA, v1=VB): two spellings of one PK slot bind different parameters, one is silently discarded — and mapping it disarms the 'never used' warning
## Summary

`PkParams::name_to_index` maps three pairs of role keys onto one slot each:

```rust
"v" | "v1"       => PK_IDX_V,
"q" | "q2"       => PK_IDX_Q,
"lagtime" | "alag" => PK_IDX_LAGTIME,
```

`pk_param_map` is keyed by the **role string**, so both spellings of a pair can appear in one `pk(...)` call bound to *different* parameters. `build_pk_param_fn` then pushes two `(pk_slot, var_slot)` pairs for the same slot and the closure writes them in order — last write wins. The other parameter's value never reaches the engine.

It parses clean, fits, and returns a wrong number.

## Measured

`predict` at fixed θ, one-compartment IV, 100 mg bolus, `CL = 5`, `VA = 50`, `VB = 100`:

| `pk one_cpt_iv(...)` | predictions (t = 1, 4) | warning |
|---|---|---|
| `cl=CL, v=VA` | `1.809674836`, `1.340640092` | `VB` computed but never used |
| `cl=CL, v1=VB` | `0.951229425`, `0.818730753` | `VA` computed but never used |
| **`cl=CL, v=VA, v1=VB`** | **`0.951229425`, `0.818730753`** | **none** |

The dual mapping is bit-identical to the `v1=VB` run. `VA` — a *mapped* central volume — is silently discarded, and the prediction is off by the ratio of the two volumes (here 2×).

Same shape for `q`/`q2` on `two_cpt_iv` (`q=QA, q2=QB` is bit-identical to `q2=QB` alone), and for `lagtime`/`alag` on an oral model.

### The sting: mapping the parameter is what silences the diagnostic

Note the warning column. Each single-mapping control **does** warn — the #315 "computed but never used" census catches the unmapped parameter. The dual mapping produces **no warning at all**, because the census counts a name that appears in the `pk(...)` line as used. So the one existing guard is disarmed by the very act that creates the bug.

The sibling `[structural_model] … does not use parameter(s) …` warning does not fire either: it iterates *role keys* and asks `pk_model.consumes_pk_slot(slot)`, and both `v` and `v1` route to a slot the model does consume.

## Why the sorted iteration does not save it

`build_pk_param_fn` sorts `pk_entries` by role key before resolving (added so a model with several bad bindings always reports the same one). That makes the winner **deterministic** — `"v" < "v1"`, so `v1` is written last and wins — but deterministic is not the same as correct. The user gets one of their two volumes, reproducibly, with no indication which.

## Suggested fix

Reject at parse time in `build_pk_param_fn`, where `pk_slot` is already resolved: if two distinct role keys resolve to the same `pk_slot` **and** bind different values, error. Sketch:

```
[structural_model]: `v=VA` and `v1=VB` are two spellings of the same parameter
(the central volume) bound to different values — only `v1=VB` would be applied and
`VA` silently discarded. Use one spelling.
```

Two cases worth distinguishing:

- **different values** (`v=VA, v1=VB`) — error, as above. This is the silent-wrong-number case.
- **same value** (`lagtime=X, alag=X`) — harmless, both writes are identical. Erroring here would be defensible too (it is still redundant), but it breaks nothing today, so a warning is enough if you want to keep it lenient.

Doing it in `build_pk_param_fn` covers every entry point at once, since that is where the role map is turned into slot assignments.

## Notes

- Pre-existing, not introduced by #1004/#1035. Surfaced while covering `analytical_role_binding`'s tie-break in that PR — which now deliberately quotes the mapping that *wins* the slot (`lagtime=` over `alag=`), so its diagnostic stays truthful under this bug rather than naming a mapping whose removal changes nothing.
- Same class as #261 (undefined `pk(...)` reference silently defaulting a slot) and #309 (omitted required PK parameter silently defaulting to 0.0), both of which were closed by turning a silent default into a parse error.
- `f` has no alias, so bioavailability is unaffected.

roninsightrx:
## Why

`PkParams::name_to_index` maps three pairs of role keys onto one slot each:

```rust
"v" | "v1"         => PK_IDX_V,
"q" | "q2"         => PK_IDX_Q,
"lagtime" | "alag" => PK_IDX_LAGTIME,
```

`pk_param_map` is keyed by the **role string**, so both spellings of a pair could appear in one `pk(...)` call bound to *different* parameters. `build_pk_param_fn` pushed two `(pk_slot, var_slot)` pairs for the same slot and the closure wrote them in order — last write wins. The other parameter's value never reached the engine.

It parsed clean, fit, and returned a wrong number. From #1048, `predict` at fixed θ, one-compartment IV, 100 mg bolus, `CL = 5`, `VA = 50`, `VB = 100`:

| `pk one_cpt_iv(...)` | predictions (t = 1, 4) | warning |
|---|---|---|
| `cl=CL, v=VA` | `1.809674836`, `1.340640092` | `VB` computed but never used |
| `cl=CL, v1=VB` | `0.951229425`, `0.818730753` | `VA` computed but never used |
| **`cl=CL, v=VA, v1=VB`** | **`0.951229425`, `0.818730753`** | **none** |

The dual mapping was bit-identical to the `v1=VB` run — the prediction off by the ratio of the two volumes (2×).

**The sting: mapping the parameter is what silenced the diagnostic.** Each single-mapping control *does* warn — the #315 "computed but never used" census catches the unmapped parameter. The dual mapping produced no warning at all, because the census counts a name that appears in the `pk(...)` line as used. The sibling `[structural_model] … does not use parameter(s) …` warning did not fire either: it iterates role keys and asks `consumes_pk_slot`, and both `v` and `v1` route to a slot the model does consume. So the one existing guard was disarmed by the very act that created the bug.

Sorting `pk_entries` by role key (added so a model with several bad bindings always reports the same one) made the winner **deterministic**, not correct.

Same class as #261 (undefined `pk(...)` reference silently defaulting a slot) and #309 (omitted required PK parameter silently defaulting to `0.0`), both closed by turning a silent default into a parse error.

## What changed

`build_pk_param_fn` — where the role map is turned into slot assignments, so every entry point is covered at once — now resolves each role key to its slot once and records what it bound. A second key landing on the same slot with a **different** binding is a parse error:

```
[structural_model]: `v=VA` and `v1=VB` are two spellings of the same parameter
(the central volume) bound to different values — only `v1=VB` would be applied
and `VA` silently discarded. Use one spelling.
```

Bindings that **agree** (`lagtime=X, alag=X`) are redundant but harmless — both writes store the same value — so they stay legal, and the duplicate write is skipped so the hot closure writes the slot once. Keeping that case legal is also what leaves `analytical_role_binding` (#1004/#1035) a tie to break.

Two details the resolution has to get right:

- **`TIME` bindings.** `pk(...=TIME)` is rewritten upstream to a synthetic individual parameter named after the *role key* (#486), so `v=TIME, v1=TIME` arrives at the slot check as two **different** variable names carrying one value. `PkSlotBinding::Time` collapses them so that reads as redundant rather than as a conflict the model file does not contain, and `pk_mapped_value_display` keeps the internal `__ferx_pktime_*` name out of the message when a genuine conflict *is* reported against a TIME binding.
- **Constant vs. variable.** The closure applies `pk_const_mapping` after `pk_assignment_mapping`, so a constant beats a variable on the same slot whatever the role keys sort to. The message names that one, rather than assuming the alphabetically later key wins.

`analytical_lag_alias_spellings_each_name_their_own_mapping`'s case (a) built exactly the now-rejected configuration (`lagtime=TLAGA, alag=TLAGB`). The property it tested — an `alag=` user is told to remove `alag=`, not `lagtime=` — is already covered on its own by `analytical_lag_mapping_read_in_scaling_is_rejected`, which runs each spelling separately, so the case is dropped and the one-parameter-two-spellings case (b) is kept.

## Alternatives considered

- **Warn instead of error.** Rejected: the failure mode is a silently wrong prediction, not a stylistic wart, and `build_pk_param_fn` has no warning channel (it returns `Result`) — plumbing one to keep a construction that has no correct interpretation is the wrong trade.
- **Error on the same-value case too** (`lagtime=X, alag=X`). Defensible — it is still redundant — but it breaks models that work correctly today, and `analytical_role_binding`'s tie-break exists precisely for that configuration.
- **Fix it in `parse_role_pairs`.** That helper is shared with `ode_template`, which rejects any role key outside its exact required set (`v` and `v1` are not interchangeable there), so the bug does not exist on that path and the check would have to be duplicated per model anyway. `build_pk_param_fn` is where the slot actually exists.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` clean); the new items are private to `parser::model_parser`.

## Breaking changes
- [ ] Public Rust API
- [ ] None

<details>

A `.ferx` file that maps both spellings of one PK slot to **different** values now fails to parse. Such a model was already producing the wrong number — only one of the two mappings ever reached the engine — so the fix is to delete the mapping that was being discarded (the message names it):

```diff
- pk one_cpt_iv(cl=CL, v=VA, v1=VB)
+ pk one_cpt_iv(cl=CL, v1=VB)
```

Both spellings mapped to the **same** value keeps working unchanged.

</details>

## Numerical validation
- [x] Not applicable — the change produces no numerical result: it turns a configuration that silently returned a wrong number into a parse error. The divergence it removes is measured in the table above (#1048), against ferx's own single-mapping controls, which is the only reference that applies: NONMEM has no equivalent of "two spellings of one slot", since `$PK` assigns `V` / `V1` directly rather than through a role map. No model that parses today changes its predictions.

## Performance
- [x] No significant impact expected — one `HashMap<usize, …>` insert per `pk(...)` entry at parse time; the hot closure does strictly fewer writes (the duplicate is skipped).

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 4443 passed, 0 failed)

Four Tier-1 tests in `src/parser/model_parser_tests.rs`:

- `pk_alias_spellings_on_one_slot_conflict` — all three alias pairs (`v`/`v1` on `one_cpt_iv`, `q`/`q2` on `two_cpt_iv`, `lagtime`/`alag` on `one_cpt_oral`), asserting the noun, the mapping that would have been applied, and the discarded value. The lag case deliberately spells the pair `alag=LA, lagtime=LB` so the winner is not simply the second argument on the line.
- `pk_alias_spellings_binding_one_parameter_are_accepted` — the lenient case still parses **and still writes the slot** (the de-duplication must not drop it), with a second fixture at a different value so the assertion cannot pass on a coincidental default.
- `pk_alias_spellings_with_literal_values_follow_the_same_rule` — a constant beats a variable regardless of key order; `v=50, v1=50.0` is the same number and stays legal.
- `pk_alias_spellings_bound_to_time_are_not_a_conflict` — `v=TIME, v1=TIME` parses; a real conflict against a TIME binding quotes `TIME` and never `__ferx_pktime`.

**Mutation-checked in both directions**, since a one-sided check would pass a strictly weaker property:

| mutation | tests that die |
|---|---|
| `if false && prev_binding != binding` (never report a conflict) | `..._on_one_slot_conflict`, `..._with_literal_values...`, `..._bound_to_time...` |
| `if true \|\| prev_binding != binding` (report *every* duplicate slot) | `..._binding_one_parameter_are_accepted`, `..._with_literal_values...`, `..._bound_to_time...` |

## Example

## Docs
- [x] `docs/` updated — `docs/model-file/lagtime.qmd` replaces the old "`lagtime` wins … intentional but unlikely to be intended" paragraph with the new rule and the message; `docs/model-file/structural-model.qmd` states it for all three alias pairs alongside the #309 / #315 mapping rules.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — parse-time only; `pk_assignment_mapping` (which feeds `pk_var_slots` for the sensitivity chain) loses only an exact duplicate entry
- [x] No version bump needed

## Reviewer hints

The subtle part is the winner calculation, not the rejection: the message claims which value the user *was* getting, so it has to match the closure's actual write order — variables first, then constants, and within each, ascending role key. `pk_alias_spellings_with_literal_values_follow_the_same_rule` is the test that pins the mixed case (`v1=VA, v=100.0` → the constant wins even though `v` sorts first).

Also worth a look: `PkSlotBinding::Time` exists only because the `TIME` desugaring names its synthetic parameter after the role key. If that naming ever becomes role-independent, the variant collapses into `Var`.

Skimmable: the changelog and docs edits, and the deletion in `analytical_lag_alias_spellings_each_name_their_own_mapping`.

## Open questions

None. Closes #1048.

https://claude.ai/code/session_01VLpGbmx8tbVagmCCVVfu44


roninsightrx (commented):
Reviewed — no findings.

I traced the resolved-binding comparison through the existing write order. The implementation correctly rejects conflicting aliases, keeps identical variable and literal bindings legal, preserves direct TIME mappings despite their separate synthetic slots, and reports the value that the old closure would actually have applied. The deduplicated mapping also remains consistent with the analytical sensitivity slot path.

Validation:
- cargo test --lib pk_alias_spellings: 4 passed
- cargo test --lib analytical_lag_alias_spellings_each_name_their_own_mapping: 1 passed
- git diff --check: clean
- All current CI checks, including core coverage, release semantics, clippy, docs lint, and public API baseline, are green.

roninsightrx:
Thanks for the trace — agreed on all four points, and the write-order one is the part I most wanted a second pair of eyes on, since the diagnostic's claim about *which* value was being applied is only true if it matches the closure's actual order (variables, then constants, ascending role key within each).

No changes made in response. For the record, the two mutation runs behind the `Tests` section: forcing the conflict check off reddens `..._on_one_slot_conflict`, `..._with_literal_values...` and `..._bound_to_time...`; forcing it on for every duplicate slot reddens `..._binding_one_parameter_are_accepted` plus the same literal and TIME tests. Both directions had to kill something, since a one-sided check would have passed a strictly weaker property than the fix.

The one thing worth a deliberate call before merge is not correctness but scope: this is a breaking `.ferx` change. A file with conflicting alias spellings stops parsing, and such a file was already returning a wrong number, so the migration is deleting the mapping that was being discarded (the error names it). Flagging it rather than merging on green.

Issue #1019, opened 2026-08-21
simulate: panic on any kappa model when simulating from a fit (omega_iov never threaded through params_from_fit)


roninsightrx:
## Summary

`simulate_with_options_diag` panics for any model that declares a `kappa` when the parameters come from a fit. The R entry point `ferx_simulate(..., fit = f)` is unusable on every IOV model.

```
thread '<unnamed>' panicked at ferx-core/src/api/simulate.rs:565:14:
omega_iov is present whenever the model declares kappa (n_kappa > 0)
```

## Reprex

Uses a bundled example only — no user data:

```r
library(ferx)
ex <- ferx_example("warfarin_iov")
f  <- ferx_fit(ex$model, ex$data)

f$omega_iov
#>            KAPPA_CL
#> KAPPA_CL  0.0390371     <- the fit HAS it

ferx_simulate(ex$model, ex$data, n_sim = 2, seed = 1, fit = f)
#> thread '<unnamed>' panicked at ferx-core/src/api/simulate.rs:565:14:
#> omega_iov is present whenever the model declares kappa (n_kappa > 0)
#> Error in ferx_rust_simulate_from_fit(...) :
#>   omega_iov is present whenever the model declares kappa (n_kappa > 0)
```

The no-fit path is fine, which localises it:

```r
ferx_simulate(ex$model, ex$data, n_sim = 2, seed = 1)   # OK, 440 rows
```

## Cause

The invariant the `expect()` relies on holds after parsing, but the simulate-from-fit path never populates the field.

`ferx-r/src/rust/src/lib.rs`:

- `ferx_rust_simulate_from_fit()` (~L477) takes `theta`, `omega_flat`, `omega_dim`, `sigma` — **the fitted IOV covariance is not a parameter of the function at all**, even though the R `ferx_fit` object carries it as `fit$omega_iov`.
- `params_from_fit()` (~L1299) therefore hardcodes:

```rust
        sigma_fixed: template.sigma_fixed.clone(),
        omega_iov: None,          // <- L1361
        kappa_fixed: Vec::new(),
```

`simulate.rs` then hits the IOV branch because `model.n_kappa > 0` and unwraps that `None`:

```rust
    let ipreds = if model.n_kappa > 0 {
        let omega_iov = params
            .omega_iov
            .as_ref()
            .expect("omega_iov is present whenever the model declares kappa (n_kappa > 0)");
```

[...]

PR #723, opened 2026-07-07
fix(sim): sample inter-occasion variability (kappa) in simulate()

TeunP:
## Why
`simulate()` resized the drawn η vector with `0.0` into the κ (IOV) slots and never sampled `omega_iov`, so a simulated / VPC dataset from an IOV model carried **no** between-occasion variability regardless of the fitted Ω_IOV — silently under-dispersing relative to the fitted model and to NONMEM `$SIM`. It was documented only in an NPDE sub-paragraph, with no runtime warning and no test.

## What changed
`emit_subject_rows` routes IOV models (`n_kappa > 0`) through the occasion-aware `predict_iov`, drawing one independent `κ ~ N(0, Ω_IOV)` per occasion from `omega_iov.chol`. Non-IOV models keep the existing TV-covariate-aware fast path and draw no extra randoms, so their output is **bit-identical**.

## Alternatives considered
Making the flat `compute_predictions_with_tv` path occasion-aware — rejected; `predict_iov` is the existing, fit-validated occasion→κ predictor, so routing through it is the minimal correct change.

## Breaking changes
- [x] None to the public API / `FitResult`. Behaviour change: simulated DV for IOV models now carries inter-occasion variability (CHANGELOG under **Changed**).

## Numerical validation
- [x] `tests/simulate_iov_kappa.rs` — simulating with the fitted Ω_IOV vs Ω_IOV forced to zero under an **aligned RNG stream** differ (fails on the old code, where κ is always 0); reproducible under a fixed seed; the non-IOV path stays bit-identical (`propensity_matched_sim` reproducibility).
- [x] **NONMEM `$SIM` variance-component anchor.** `simulate_iov_recovers_omega_iov_variance` + `tests/fixtures/iov_anchor.{ferx,csv}` (mirrored by the committed NONMEM kit `nonmem_anchor/iov_anchor.{ctl,csv}`). Shared 300×3 design — 1-cpt IV bolus, IOV on `V`, `EVID=4` reset+bolus per occasion, obs at `t=0` so `IPRED = AMT/V` exactly, making the within-subject between-occasion `Var(log IPRED)` an unbiased estimator of Ω_IOV:

  | Source | between-occasion Var(log IPRED) |
  |--------|--------------------------------:|
  | Truth (Ω_IOV) | 0.04000 |
  | NONMEM 7.6.0 `$SIMULATION` (df 600) | 0.04151 (SE ~0.0024) |
  | ferx `simulate()` (df 12000, this test) | 0.03939 |
[...]

PR #1362, opened 2026-09-12
fix(stats): sample the occasion kappa in the NPDE/NPD reference [closes #734]

### LINKED ISSUE #734: NPDE/NPD post-fit diagnostics hold IOV κ at zero (under-disperse for IOV models)
### Context

PR #723 fixed `simulate()` to draw an independent occasion `κ ~ N(0, Ω_IOV)` per occasion (matching NONMEM `$SIM`). The **post-fit NPDE / NPD** diagnostics are a **separate code path** (`src/stats/npde.rs::compute_npde_npd`, npde.rs:79) that builds its reference distribution *without* sampling occasion κ — it still holds every κ at zero.

### Effect

For an IOV model, the NPDE/NPD reference distribution carries **no between-occasion variability**, so the standardized residuals are over-dispersed relative to a correctly-simulated reference (the diagnostic understates its own spread). This is the same bug class PR #723 fixed for `simulate()`, one path over. It is documented as a known follow-up in #723's body and in `docs/model-file/simulation.qmd` ("The post-fit NPDE/NPD diagnostics are a separate path that still holds kappas at zero").

### Fix

Route the NPDE/NPD reference simulation through the occasion-aware `pk::predict_iov` (drawing per-occasion κ from `Ω_IOV.chol`), mirroring what `emit_subject_rows` now does for `simulate()`. Non-IOV models are unaffected.

### Validation

Extend the PR #723-style variance-component anchor to the NPDE path (the `iov_anchor` kit in `nonmem_anchor/` is a ready design), or assert the NPDE reference spread recovers Ω_IOV.

_Filed as a follow-up to #723._

roninsightrx:
## Why

Closes #734.

PR #723 fixed `simulate()` to draw an independent occasion `κ ~ N(0, Ω_IOV)` per occasion. The post-fit **NPDE/NPD** diagnostics are a separate code path (`src/stats/npde.rs::compute_npde_npd`) and still built their Monte-Carlo reference distribution with every `κ` at zero.

So for an IOV model the reference carried **no between-occasion variability**: the standardized residuals were over-dispersed relative to a correctly simulated reference, and a well-specified IOV model read as mis-specified. Measured on the new anchor, the old behaviour disagrees with NONMEM's own NPDE by up to 4.39 score units.

## What changed

`compute_npde_npd` now draws one independent `κ ~ N(0, Ω_IOV)` per occasion group and routes the replicate through `pk::predict_iov`, mirroring what `emit_subject_rows` does for `simulate()`.

- Occasion groups come from `stats::likelihood::iov_occasion_groups` — the exact order `predict_iov` indexes its `kappas` argument by — and this side builds them once per subject rather than once per replicate. Note that `predict_iov` still builds its own copy (plus an `occ_to_k` map) on every call, so the reference pays two group walks per replicate and only the caller-side one is hoistable here; plumbing the groups through its signature is a separate change, and the estimation hot path pays the same duplicate today.
- **Non-IOV models are untouched**: they keep the TV-covariate-aware dispatcher and draw no extra randoms, so their scores are byte-identical.
- A caller-rebuilt `ModelParameters` that drops `omega_iov` (the #1019 failure mode on the R bridge) falls back to `κ = 0` rather than panicking — a post-fit diagnostic is the wrong place to abort a completed fit. Documented on the function and unit-tested.

## Alternatives considered

Rejecting a missing `omega_iov` the way `validate_iov_simulatable` does for `simulate()`. `compute_npde_npd` returns `Vec<SubjectNpde>` with no error channel and runs *after* a fit has completed, so the options were panic or fall back; falling back with the behaviour documented is the lesser harm, and the fit path itself always carries `omega_iov` for a `kappa` model.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` green, baseline unchanged).

## Breaking changes

- [x] None

## Numerical validation

**NONMEM anchor — `tests/npde_iov_nonmem_anchor.rs`.** NONMEM *does* emit NPDE/NPD natively (`$TABLE ID TIME DV NPDE NPD ESAMPLE=2000 SEED=734`), which makes this a direct cross-tool comparison rather than a variance-component proxy.

`nonmem_anchor/npde_iov_anchor.ctl` is a bit-for-bit twin of `tests/fixtures/npde_iov_anchor.ferx` on the shared dataset `nonmem_anchor/npde_iov_anchor.csv` (60 subjects × three `EVID=4` reset+bolus occasions × three post-dose observations = 540 rows, DV simulated once under the same θ/Ω/Ω_IOV/Σ — `tests/gen_npde_iov_anchor.rs`). Every parameter is `FIX`ed at the simulating value on both sides, and `ESAMPLE = nsim = 2000` with matching seeds, so the two tools build the reference distribution from the same population model and differ only in their Monte-Carlo stream.

Row by row over all 540 observations:

```
                     mean Δ      sd Δ    worst |Δ|
NPD                 -0.00587   0.04820    0.22578   (ID 34, t = 49)
NPDE (WRESCHOL)      0.00063   0.07591    0.28044

distribution         ferx mean / sd      NONMEM mean / sd
NPD                  -0.10026 / 1.03629  -0.09439 / 1.04387
NPDE                  0.00076 / 1.01879   0.00013 / 1.01949
```

That is the Monte-Carlo disagreement of two independent 2000-draw empirical CDFs (`se(pd) ≈ 0.011`, inflating to ≈0.05–0.15 on the score near the tails), not a model difference. Test bounds are 2× the worst realised value (0.45 / 0.56).

**The anchor can fail.** The `zero_omega_iov` control reproduces the old behaviour with the RNG stream still aligned (the κ draws happen and scale to zero) and misses both bounds: worst |ΔNPD| **0.895** (2.0×), worst |ΔNPDE| **4.389** (7.8×). Asserted, so the straddle cannot silently disappear.

**A convention this turned up.** NONMEM's default NPDE decorrelation is the *symmetric* square root of the simulated covariance (its CWRES convention); ferx follows Brendel/Comets and the `npde` R package in using the *Cholesky* factor, which NONMEM produces on request via `WRESCHOL`. Both tables are committed and both are asserted:

| table | `$TABLE` | ferx NPD | ferx NPDE |
|---|---|---|---|
| `npde_iov_anchor.tab` | default | matches (corr 0.99895) | **differs** (corr 0.8682) |
| `npde_iov_chol.tab` | `WRESCHOL` | bit-identical NPD column | matches (corr 0.99723) |

The NPD column is bit-identical between the two NONMEM runs — NPD carries no decorrelation — which is what makes the pair isolate the convention and nothing else. Pinning the *disagreement* with the default table matters as much as the agreement: without it, "make NPDE match NONMEM" is an invitation to adopt the wrong factor.

**Non-IOV unchanged:** `tests/npde_validation.rs` (warfarin, NPDE ≈ N(0,1) vs the `npde` R package) passes unchanged.

## Performance

- [x] No significant impact expected — IOV models pay `n_kappa` extra normal draws per occasion per replicate; non-IOV models are on the unchanged branch.

## Tests

- [x] Unit tests added in `src/stats/npde.rs` (`cargo test --lib` green: 4442 passed, 0 failed)
- [x] Regression test that fails without this fix

Tier-1 (these are what register patch coverage — slow/nightly tests do not):

1. `npde_reference_spread_includes_occasion_kappa` — a one-row fixture (bolus and observation both at `t = 0`, so `log IPRED = log(AMT/TVV) − η_V − κ_V`) whose reference spread has the **closed form** `sqrt(ω²_V + ω²_κ) = 0.2236`, with the κ-less `0.1` as its control. The observation is placed exactly one reference SD above the median, so `IOV_REF_SD / NPD` reads the implied spread back directly. Tolerances are measured, not argued: swept over six seeds at `K = 20_000` the realised implied SDs span 0.22149–0.22644 and 0.09857–0.10084 (worst deviations 1.3% / 1.4%), and the bands are 3%. Mutation-checked — disabling the fix reddens it, and the control arm is asserted so the pair cannot silently stop straddling.
2. `npde_without_omega_iov_falls_back_to_zero_kappa` — pins the documented fallback.
3. `npde_iov_without_occasion_labels_matches_the_kappa_less_reference` — an IOV model on unlabelled data now routes through `predict_iov` with empty `kappas` instead of the non-IOV dispatcher. That is the path where `predict_iov` used to drop `[scaling]` entirely (#723 review), so the swap is pinned rather than assumed.

## Docs

- [x] `docs/model-file/fit-options.qmd` — the stale "holds kappas at zero" limitation is replaced by an **Inter-occasion variability (IOV)** section with the anchor numbers, plus a new section on the NPDE decorrelation convention. Also corrects the page's claim that "NONMEM has no native NPDE output".
- [x] `docs/model-file/simulation.qmd` — the parenthetical pointing at this gap now says the NPDE path draws its own kappas.
- [x] `CHANGELOG.md` `[Unreleased] → Fixed`

## Checklist

- [x] `tools/preflight.sh` green (fmt, 5 check feature sets, clippy `--all-targets`, rustdoc, docs-lint, public-API baseline)
- [x] No `Dual2` / `sens` code touched

## Reviewer hints

The load-bearing diff is ~30 lines in `src/stats/npde.rs`; everything else is the anchor kit and docs. Worth a look:

- the `iov` binding is hoisted out of the replicate loop on purpose (occasion groups are a function of the subject, not the draw) — though see the note under **What changed**: `predict_iov` rebuilds its own groups per call, so this halves the walks rather than removing them. `.filter(|_| model.n_kappa > 0)` is what keeps a stray `omega_iov` on a non-IOV model from changing that model's RNG stream;
- `eta_slice`'s κ tail stays zero and is still what `residual_var_scale` reads — same convention as `emit_subject_rows`, deliberately not folded into the κ draw;
- `nonmem_anchor/npde_iov_anchor.csv` is committed in one place only, and both the NONMEM `$DATA` and the Rust test read that path, so they cannot drift.

## Open questions

The generator (`tests/gen_npde_iov_anchor.rs`) is `slow-tests`-gated and manual, matching `gen_transit_iov_anchor.rs`. Re-running it rewrites the committed CSV; that is intentional (it is deterministic under `SIM_SEED`) but it does mean the anchor data and the NONMEM tables must be regenerated together if the design ever changes.

https://claude.ai/code/session_01WnzCRKp4dNB2uxp7VnKjPs


roninsightrx (commented):
Reviewed the implementation, regression coverage, NONMEM anchor, docs, and changelog. I found no blocking issues. The IOV branch draws one independent kappa vector per occasion from the live omega_iov Cholesky factor, routes through predict_iov with the established occasion ordering, preserves the residual-error eta convention, and leaves the non-IOV RNG path unchanged. The zero-omega and missing-omega controls meaningfully straddle the regression, and the focused tests pass locally (18 NPDE unit tests; 2 NONMEM anchor tests). All reported CI checks are green.

One non-blocking performance/documentation note: compute_npde_npd hoists its own iov_occasion_groups call, but predict_iov recomputes iov_occasion_groups and rebuilds occ_to_k on every replicate. Thus the PR description claim that occasion groups are built once per subject is not quite true for the full call path. This does not affect correctness, and the focused anchor remains fast (0.74s locally), so I do not consider it a merge blocker; it could be handled as a later scratch/precomputation optimization or by narrowing that wording.

roninsightrx:
Confirmed and fixed the wording — the note is right.

`predict_iov_with_scratch` calls `iov_occasion_groups` itself (`src/pk/mod.rs:1288`) and rebuilds `occ_to_k` right after, on every call. So the NPDE reference pays **two** group walks per replicate; hoisting the caller-side one out of the loop halves them rather than removing them, and "built once per subject" overstated it.

Corrected in two places:

- `src/stats/npde.rs` — the comment above the hoist now says what it actually buys and names the second, non-hoistable build (a152c4d2).
- PR description — same correction under **What changed** and **Reviewer hints**.

Not plumbing the groups through `predict_iov`'s signature in this PR: it is `pub(crate)` with callers in `sens/` and the estimation inner loop, all of which pay the identical duplicate today, so it is a hot-path change on its own merits rather than a rider on a diagnostics fix. Happy to open a follow-up issue for it if you want it tracked.

No behaviour change in the fixup; `tools/preflight.sh fmt clippy` green.

PR #1363, opened 2026-09-12
feat(estimation): `[priors] from_fit` — build a prior from a previous fit (#254)

### LINKED ISSUE #254: feat: Add (a simple and understandable) prior support for frequentist/MAP estimation (nonmem $PRIOR analog)

roninsightrx:
## Why

Phase 1 (#1341) gave #254 its inline `prior(value, rse = 25%)`. What it did not
give is the ergonomics the issue is actually about: *model updating*. Taking a
published popPK model as the prior for a fit on sparse local data still meant
transcribing a parameter table by hand — a dozen values, a dozen RSEs, each one
a chance to put a variance where an SD belongs.

This is the issue's `[priors] from_fit` acceptance criterion ("Import-prior-from-
previous-fit works end to end in one call"), one of the three Phase-1 left open.

```toml
[priors]
  from_fit = "parent-model-fit.yaml"
```

Every θ, Ω diagonal, Σ and κ the source run reports with a usable standard error,
and whose name **and family** match a `[parameters]` declaration in this model,
becomes `prior(estimate, rse = SE/estimate)` on *this* model's declared scale.

## What changed

**The import is expanded into ordinary `ParameterPrior`s and then walked by the
same loop as the typed ones.** `PriorSet::build` gained one line:

```rust
for prior in model.priors.iter().chain(imported.iter()) { … }
```

Nothing downstream — penalty, gradient, Hessian, `prior_summary`, the console
report, the YAML — can tell an imported prior from a typed one, because there is
one resolution path rather than two that could disagree about a scale. The
`*_g<T: PkNum>` rule generalised: when the bug report would be "two
implementations disagree", write one implementation.

**Matching is on name *and* kind.** A source θ called `CL` and a target Ω called
`CL` are both legal and the numbers are on unrelated scales, so name alone would
import one as the other with nothing to say so. `EstimateKind` travels with both
sides; `CoordInfo` gained the field.

**The scale conversion has exactly one direction that matters.** A `FitResult`
states Ω and κ as **variances** and Σ as an **SD**, whatever the *source* model
declared them as — `omega X ~ 0.09` and `omega X ~ 0.3 (sd)` produce
byte-identical output. So the source's own `(sd)` spelling is invisible here and
is deliberately not guessed at. Only the target's declaration converts, by the
delta method on a power (the relative SE of `xᵖ` is `|p|` times that of `x`):

| Source reports | Declared here as | Prior centre | Prior RSE |
|---|---|---|---|
| Ω variance `v` | `omega X ~ v` | `v` | `SE/v` |
| Ω variance `v` | `omega X ~ s (sd)` | `√v` | `SE/(2v)` |
| Σ SD `s` | `sigma X ~ v` | `s²` | `2·SE/s` |
| Σ SD `s` | `sigma X ~ s (sd)` | `s` | `SE/s` |

An identity-packed θ (negative declared lower bound) imports its **absolute** SE
instead — a relative one is meaningless where the estimate may be zero, and
would explode near it.

**Three readers, dispatched on the extension** (`src/io/fit_estimates.rs`):

| Extension | Reader | Precision |
|---|---|---|
| `.fitrx` | `fitrx::load_fit` | exact |
| `.json` | `FitResult`'s serde derives | exact |
| `.yaml` | a small purpose-built reader | 6 decimals |

The YAML one is the file **every** run writes, so leaving it out would mean
`from_fit` is not one call for the default workflow. It reads exactly the four
sections it needs at exactly the indentation `write_estimates_yaml` emits, and is
pinned to that writer by a round-trip test rather than by inspection. No YAML
dependency is added — the crate writes this file itself, and a new dependency on
the `ferx-core` compile path (93% LLVM, #971) buys nothing here.

**Path resolution follows `[data] path`**: relative → the model file's directory,
done in `parse_full_model_file`. `parse_model_file` now routes through it rather
than through `parse_model_string`, so a caller that only wants the
`CompiledModel` gets the same resolution.

**Skips vs errors.** A typed prior that cannot be applied is a hard error
(phase 1). An import is a bulk operation over a source model that legitimately
has parameters this one does not, so it skips — but every skip that is not
"simply absent from this model" is reported in `FitResult::warnings`:

- no standard error in the source (a FIXed parameter, or no covariance step);
- FIXed in *this* model;
- a `block_omega` / `block_sigma` element or a `[mixture]` override;
- already carrying an inline `prior(...)` — **the inline one wins**, so one
  imported prior can be overridden without giving up the rest (silent: it was
  deliberate, and it is visible in `prior_summary`).

An import that lands **nothing at all** is a hard error, unconditionally. Not
weakened to "…unless some typed prior landed": the user wrote `from_fit`, and a
wrong path or a wholesale rename is otherwise invisible.

## Alternatives considered

**Reading only `.json` / `.fitrx`** and telling the user to re-run with
`--output-format json`. Rejected — the default run writes YAML, so this would
make the one-call criterion a two-run criterion. The cost is a hand-written
reader, and the round-trip test is what keeps it honest.

**Adding `serde_yaml`** to parse that file generically. Rejected: a new
dependency on the crate that is already 93% LLVM time, to read a file this crate
writes, whose shape is four fixed sections.

**Deriving the prior SD from the source's packed standard error**
(`sqrt(diag(covariance_matrix))`), which is exact on the coordinate the penalty
actually lives on and needs no delta method. Rejected on two counts: it produces
a prior the user cannot reproduce by hand off the printed table, so an imported
prior and a typed one would differ slightly for no reason the user can see; and
it requires the source's packed layout, which the curated YAML does not carry.
The difference is second-order (`sqrt(ln(1+r²))` vs `r`: 0.2462 vs 0.25 at
r = 25%).

**Expanding the import in the parser** rather than in `PriorSet::build`. The
parser has no `ModelParameters`, so it would have to re-derive the packed layout,
the FIX mask and the `*_init_as_sd` tables that `coordinate_table` already
walks — a second copy of the thing the scale depends on. Building it where the
table already exists costs a small file read per estimation stage (~5 per fit,
never per evaluation) and buys one derivation instead of two.

**Guessing the source's declared scale** from its `omega_init_as_sd` flags.
Rejected as the trap it is: a `FitResult` does not vary with that flag, so the
guess is a factor of two on exactly the fits whose source model was written
`(sd)`, and it would still look plausible. Pinned shut by
`omega_declared_as_sd_in_the_source_is_invisible`.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#___ | not needed | after |

Nothing existing breaks. A follow-up ferx-r PR could surface
`CompiledModel::prior_from_fit` for programmatic model construction, and would
need the usual `tools/update-ferx-core-lock.sh` bump.

## Breaking changes

- [ ] None

**Not purely additive, and the earlier draft of this section said otherwise.**
Three of the five added items are new `pub` **fields** on structs that are
neither `#[non_exhaustive]` nor privately-fielded, so an external crate
constructing any of them by struct literal no longer compiles.

| Added item | Why it exists | Breaks a literal? |
|---|---|---|
| `CompiledModel::prior_from_fit` | what the `[priors]` block parses into; the parser writes it and `expand_prior_from_fit` consumes it | yes |
| `ParameterPrior::kind` | the `[parameters]` family a prior is on, so `theta CL` and `omega CL` in one model both resolve. Both producers know it; re-deriving it from the name is what made them ambiguous | yes |
| `types::ParameterKind` | that field's type, shared with the importer rather than duplicated as a private twin | no (new enum) |
| `parser::model_parser::expand_prior_from_fit` | consumes `prior_from_fit`; the only thing that reads the source fit | no (new fn) |
| `ParseBindings::model_dir` | the directory a relative `from_fit` resolves against. The expansion runs *during* the parse, so the base directory has to be in scope by then — resolving it afterwards, as `[data] path` does, would expand against the wrong directory first | yes |

**ferx-r is unaffected**: it only ever takes `&CompiledModel` (`src/rust/src/lib.rs`)
and references neither `ParameterPrior` nor `ParseBindings`, so nothing there
constructs any of these by literal. I checked rather than assumed — the earlier
draft of this table claimed ferx-r "constructs `CompiledModel` rather than always
going through the parser", which is not true of the current wrapper.

`expand_prior_from_fit` is public for a narrower reason than that table's usual
one: **no caller needs it today.** It is the escape hatch the
`PriorSet::build` guard's error message names, and without it the public
`prior_from_fit` field is a trap — a caller can set it and nothing but a parse
will ever honour it. If you would rather shrink the surface, the alternative is
to make it `pub(crate)`, reword the guard to say "parse the model with
`parse_model_file` / `parse_model_string`", and document `prior_from_fit` as
parser-only. Happy either way; say which you prefer.

`[priors]` is a new block, so the feature is opt-in: a model without one has
`prior_from_fit: None` and takes a bit-identical path.

<details>

Only for an external crate that builds a `CompiledModel`, a `ParameterPrior` or
a `ParseBindings` with a struct literal. Add `..Default::default()` where the
struct derives it (`ParseBindings`), or the new field explicitly:
`prior_from_fit: None` and `kind: None`. `None` on `ParameterPrior::kind`
preserves the previous behaviour exactly — resolution by name alone.

No `.ferx` file, `FitResult` field or estimate changes.

</details>

## Numerical validation

- [x] Compared against: **the parent run's own written estimates**, which is the
      right oracle here — the feature's claim is "the prior equals what you would
      have typed off that run's output", so the reference is that output.
- [x] Reference values:

`examples/warfarin.ferx` → `warfarin-fit.yaml` → `examples/warfarin_update.ferx`:

```
# parent run (examples/warfarin.ferx)
OFV: -280.3640
TVCL 0.132969   TVV 7.730700   TVKA 0.725207
ETA_CL 0.028595  ETA_V 0.009577  ETA_KA 0.348964   PROP_ERR 0.010749 (SD)

# the update's imported priors, read back out of warfarin_update-fit.yaml
TVCL      prior 0.132969   ← parent estimate, exactly
TVV       prior 7.730700   ← exactly
TVKA      prior 0.725207   ← exactly
ETA_CL    prior 0.028595   ← parent Ω *variance*, on this model's variance scale
PROP_ERR  prior 0.010749   ← parent Σ *SD*, on this model's `(sd)` scale
7 of 7 parameters imported, no skips
```

The two scale rows are the load-bearing ones: `ETA_CL` arrives as a variance and
`PROP_ERR` as an SD *from the same file*, because the source reports one of each
and this model declares `sigma PROP_ERR ~ 0.02 (sd)`.

The **Tier-3** test is the behavioural claim, and it is a three-way comparison
rather than a round trip. Refitting the same data with a prior centred on its own
MLE is degenerate — likelihood and prior are maximised in the same place, so an
engine that ignores priors passes. Instead: fit the parent on all ten warfarin
subjects, then update on two subjects whose concentrations run 60% high.

```
                     TVCL          TVV
parent               (reference)   (reference)
sparse, no prior     36.5% off      38.0% off
sparse, from_fit      5.2% off       5.0% off
```

The test asserts the control **drifts** first (otherwise nothing below tests
anything) and then that the update is at least 2× closer; realised is ~7×.

## Performance

A model with no `[priors]` block short-circuits in `PriorSet::build` before any
filesystem access — bit-identical. A model with one reads the fit file once per
estimation stage (~5 per fit), never per objective evaluation; the resulting
terms are the same closed-form penalty phase 1 already measured as unmeasurable
next to a population inner loop.

## Tests

**Tier 1 — `src/io/fit_estimates_tests.rs` (16):** the scale rule per family
(Ω/κ variance, Σ SD); a FIXed parameter reporting no SE on *both* the JSON and
YAML paths; `0.0` / `NaN` / negative SEs reading as absent; round trips through
all three readers, with the YAML one compared against `from_fit_result` as its
oracle; the `sigma:  # error model: …` commented section header; block-Ω
off-diagonals being dropped; the `[covariate_nn]` / level-block θ exclusions
being the writer's, not a name-shape rule; every error message.

**Tier 1 — `src/estimation/priors_tests.rs` (+11):** an imported prior compared
against the *typed* `prior(...)` a user would have written, so a wrong scale on
any family fails against something other than the import's own arithmetic; the
σ variance-vs-`(sd)` pair; the Ω halving with a hand-computed straddle
(0.266005 vs 0.136743 un-halved); the source-`(sd)`-is-invisible differential;
the identity-θ absolute SE; the inline override; the FIXed-here skip *and* the
typed form still hard-erroring; the kind-mismatch reject and its straddle.

**Tier 1 — `src/parser/model_parser_tests.rs` (+5):** the block grammar, the
unknown-key reject, and model-file-relative resolution with an absolute-path
straddle.

**Tier 2 — `tests/parameter_priors.rs` (+6):** the differential that dies if the
imported penalty never reaches the optimizer (two source fits either side of the
start); the declared-scale centres end to end; relative resolution through
`parse_model_file` with the from-a-string arm as its straddle; both hard errors;
the skip warnings.

**Tier 3 (+1):** the sparse model update above.

### Mutation testing

Each mutation named its own failing test and no other:

| Mutation | Reddens |
|---|---|
| `2.0 * rel` → `rel` (σ SD→variance) | `sigma_converts_from_the_source_sd_onto_the_declared_scale`, `an_imported_prior_equals_the_prior_the_user_would_have_typed` |
| `0.5 * rel` → `rel` (Ω variance→SD) | `omega_converts_onto_the_targets_declared_scale` |
| drop `c.kind == est.kind` | `a_name_that_matches_a_different_family_is_not_imported` |
| drop the inline-override skip | `an_inline_prior_overrides_the_imported_one` |
| `se_of` returns its input unfiltered | `a_zero_or_missing_standard_error_reads_as_absent`, `a_source_parameter_without_a_standard_error_is_skipped` |

Two fixture-degeneracy traps found and fixed while writing these:

- the Tier-3 test was first written as a **round trip** — refit the same data
  with a prior centred on its own MLE. It passed, and could not fail: the
  unpriored control reproduced the parent to 2%, so the prior was doing nothing.
  Replaced with the sparse three-way comparison above, whose control arm is
  asserted to drift *before* the closeness claim is made.
- `omega_converts_onto_the_targets_declared_scale`'s straddle constant was
  guessed rather than computed (0.1926 for a number that is 0.266005). It failed,
  which is the only reason it is now hand-computed with the un-halved value
  (0.136743) stated next to it.

`examples/warfarin_update.ferx` — run `examples/warfarin.ferx` first to produce
`warfarin-fit.yaml`, then the update reads it. Output above.

## Docs

- [x] New pages linked in `docs/_quarto.yml` (no new page — a section on the
      existing `docs/estimation/priors.qmd`, plus a cross-reference from
      `docs/model-file/parameters.qmd`)
- [ ] No user-visible change

`cargo run -p docs-lint -- --check` is clean.

## Checklist

- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), rustdoc, docs, public-API baseline
- [x] Functions on the `Dual2` sensitivity path stay differentiable — **not touched**; this is file I/O and a name lookup, and the penalty itself is phase 1's closed form in packed space

`cargo test --features ci`: **5276 passed, 0 failed, 178 ignored** across 186
binaries. `--features ci,slow-tests --test parameter_priors`: 24 passed.

### ferx-site

`[priors]` is a new block and `examples/warfarin_update.ferx` is a new example,
so both need a follow-up in ferx-site / ferx-book — the same two boxes #1341 left
flagged. No new data file (it reuses `data/warfarin.csv`).

### ferx-book

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core — not needed, no ferx-r change here
- [x] All affected `examples/*.ferx` run cleanly via CLI (`warfarin.ferx` then `warfarin_update.ferx`)
- [ ] All affected ferx-site example `.qmd` pages render cleanly — see above
- [ ] All affected ferx-book chapters render cleanly — see above

## Reviewer hints

Read in this order:

1. **`src/estimation/priors.rs::import_from_fit`** — the whole feature. The
   `match (est.kind, info.scale)` is where a factor of two would live and still
   look right.
2. **`src/io/fit_estimates.rs`** module docs — the one scale rule, and why the
   source model's `(sd)` spelling must not be consulted.
3. **`parse_estimates_yaml`** — the only hand-written parser here. Its contract
   is "what `write_estimates_yaml` emits", and `yaml_round_trip_matches_the_fit_result`
   is what enforces that rather than my reading of the writer.

Skimmable: the 28 `prior_from_fit: None` struct-literal additions.

## Open questions

Still open from #254 after this, and unchanged by it:

- **IMP / IMPMAP / SAEM penalized M-step** — the issue's "composes with at least
  FOCE and one EM method". IMP's θ/σ M-step is BOBYQA over `[log θ, log σ]`,
  which is a sub-vector of the packed coordinates, so the existing `PriorSet`
  can serve it with a coordinate remap. The Ω M-step is the interesting half: the
  weighted second moment `S/N` becomes a 1-D Newton solve of
  `2Nu + S·e^(−2u) + ((u−m)/s)²` per priored diagonal (strictly convex; reduces
  exactly to `u* = ½·ln(S/N)` unpriored), and the mu-referenced θ shift becomes
  the precision-weighted `u = (N(u_old+η̄)/ω² + m/s²)/(N/ω² + 1/s²)` instead of
  `+= η̄`.
- **NONMEM anchors beyond the θ one** phase 1 already landed.

Two decisions here worth overruling if you disagree:

- **The inline prior wins over the imported one**, silently. The alternative is a
  duplicate-declaration error, which would mean "import everything except this
  one" is not expressible.
- **An empty import is a hard error even when typed priors landed.** The fit
  would be penalized — just not by the thing the user asked for — so the softer
  rule would let a wrong `from_fit` path ship silently. I think the strict rule
  is right, and it is one condition rather than two.

Closes #254 partially; the EM M-step item above is what is left.

https://claude.ai/code/session_014araEMYZa4M4xPCrDU3Bog


roninsightrx:

1. **[blocking] Preserve `EstimateKind` after import matching.** `import_from_fit` correctly matches a source estimate to `CoordInfo` by both kind and name, but then reduces that match back to a name-only `ParameterPrior` (`src/estimation/priors.rs:517-521`). The shared loop subsequently resolves every prior by name alone (`:180-184`), so the kind information is lost before a term is built. A target with legal same-name coordinates such as theta `CL` and omega `CL` rejects an imported theta `CL` as ambiguous instead of applying it to the theta. I verified this with a local regression: source theta `CL`, target theta+omega both named `CL`; `PriorSet::build` fails with `the name resolves to 2 packed coordinates`. Please carry the resolved coordinate/kind through to term construction (or keep imported resolution separate) and add that collision case to the family-matching tests.

2. **[blocking] Do not silently drop imported priors when a later read fails.** `PriorSet::build` now performs filesystem I/O every time it is called, but `outer_optimizer::build_prior_set` still converts every error to an empty set with `unwrap_or_default()`. The initial `fit()` validation catches the first read, but public post-fit paths such as `run_covariance` and `run_sir` do not call `check_parameter_priors`; if the source fit has moved, become unreadable, or malformed, prior Hessian/weights disappear and those APIs continue unpenalized. Re-reading in the optimizer/covariance/final-report phases also creates a TOCTOU path where the optimized prior and reported prior can differ. Please resolve/cache the import once, or propagate build errors through every public path that consumes it.

3. **[example] The documented commands cannot find the parent YAML.** `examples/warfarin_update.ferx` says the repo-root command for `examples/warfarin.ferx` writes `warfarin-fit.yaml` next to the model, then uses `from_fit = "warfarin-fit.yaml"`. The CLI actually writes `{model}-fit.yaml` in the process working directory (`crates/ferx-cli/src/main.rs:287`), while `from_fit` is resolved against the update model's directory. Following the shown commands therefore writes `./warfarin-fit.yaml` and reads `./examples/warfarin-fit.yaml`. Please make the example commands/path agree and cover the actual two-command example flow.

Verification on current head `28ab9182`:

- `cargo test --lib estimation::priors::tests` — 34 passed
- `cargo test --lib io::fit_estimates::tests` — 16 passed
- Added local collision regression — failed as described above (temporary test removed; worktree clean)

roninsightrx:
All three confirmed and fixed in `39798ae1`. Thanks — #1 and #2 were both real defects, and #2 in particular was a design mistake rather than an oversight.

**1. `EstimateKind` lost after matching — fixed, and it was worse than reported.**

Confirmed exactly as described. `ParameterPrior` now carries `kind: Option<ParameterKind>` (a new public enum, shared with the importer rather than a private twin), so the resolved family travels to term construction instead of being re-derived from the name.

Worth flagging: this is a **phase 1** defect too, not only an import one. A *typed* `prior(...)` on a model with `theta CL` + `omega CL` was equally ambiguous, even though the parser saw exactly which declaration the tail was peeled off. Both producers now supply the family, so both resolve; `None` still resolves by name alone, which is what a hand-built `ParameterPrior` can always fall back to and is the only case the "ambiguous" error was ever right about — its message now says which two families collided and how to disambiguate.

Collision case added at both ends: `a_name_shared_by_a_theta_and_an_omega_still_resolves` (imported *and* typed, plus the unqualified-collision reject as the straddle).

There is a footnote here I owe you, because my first attempt at that test was not honest. It built its `ParameterPrior` by hand, so it pinned the *resolution* and not the parser that feeds it — blanking the parser's `kind: Some(kind)` left it, and the whole suite, green. Added `the_parser_records_which_declaration_a_typed_prior_came_off`, which goes through the real grammar on a collision model; that mutation now reddens.

**2. Silent drop / TOCTOU — fixed by moving the read out of the estimation path entirely.**

You are right that patching `unwrap_or_default()` at each call site would not have been enough, and the TOCTOU half of your point is what settled the design: the prior the optimizer minimised and the prior the report printed were not guaranteed to be the same object.

The source fit is now read **once, by the parser**. `expand_prior_from_fit` appends the imported priors to `model.priors`, records skips in `parse_warnings` (which `fit()` already surfaces, so the `fit.rs` hook and `PriorSet::notes` are both gone), and consumes `prior_from_fit`. `PriorSet::build` does no I/O at all and refuses a model that still carries an unexpanded path — that guard is why `expand_prior_from_fit` is public, since a `CompiledModel` assembled field by field (ferx-r) has nothing to read the file for it.

So `run_covariance` and `run_sir` are safe by construction now rather than by remembering to validate.

One consequence I want to state rather than bury, because it is a real behaviour change: **a `from_fit` model is no longer self-contained at parse time.** A missing or unimportable source fails the parse, so `ferx check` refuses it, and `test_parse_all_example_ferx_files` can no longer parse `warfarin_update.ferx` from a clean checkout. I skipped it there with a comment naming the test that *does* parse it for real (with a parent fit staged in a temp dir), rather than weakening the smoke test. I think refusing at parse is right — it is the earliest point the error can be reported — but say so if you would rather it were deferred.

**3. Example commands — fixed, and pinned.**

Confirmed: `{model}-fit.yaml` goes to the working directory (`main.rs:261`, from the model path's file stem) while `from_fit` resolves against the model file's directory, so the repo-root commands wrote `./warfarin-fit.yaml` and read `./examples/warfarin-fit.yaml`.

The commands now `cd examples` and explain why, and `the_update_example_names_a_file_the_documented_commands_produce` asserts the invariant that makes that instruction correct: the example must name its parent by a **bare filename**, and the parent must be its **sibling**. Either alone permits the mismatch. It then stages the two files and parses the real example, so it is not only a string comparison. (Mutated to `../warfarin-fit.yaml` — reddens.)

Ran the documented flow verbatim end to end:

```
cd examples
ferx warfarin.ferx        --data ../data/warfarin.csv    # OFV -280.3640
ferx warfarin_update.ferx --data ../data/warfarin.csv

  TVCL      0.13297 → 0.13284   shift -0.02
  ETA_CL    0.02859 → 0.02859   (parent Ω variance, variance scale here)
  PROP_ERR  0.01075 → 0.01065   (parent Σ SD, `(sd)` scale here)
  7 of 7 imported
```

`ETA_CL` and `PROP_ERR` come off the same file on different scales, which is the conversion table doing its job.

**Verification on `39798ae1`**

- `tools/preflight.sh` — green (fmt, 5 check feature sets, clippy, rustdoc, docs, public-API baseline)
- `cargo test --features ci` — 5283 passed, 0 failed, 178 ignored, 186 binaries
- `cargo test --features ci,slow-tests --test parameter_priors` — 26 passed

New public API (baseline regenerated): `ParameterKind` + `ParameterPrior::kind`, `parser::model_parser::expand_prior_from_fit`, `ParseBindings::model_dir`. The last is what lets the expansion resolve a relative path against the model file *during* the parse — resolving it afterwards, the way `[data] path` does, would mean expanding against the wrong directory first.

roninsightrx:
## Code review — `[priors] from_fit` (#1363)

Reviewed `e5283719..HEAD` (the phase-2 commits only; #1341 is already on `main`).
Ran `cargo test --lib fit_estimates` — 16/16 green.

The core of this is right, and I checked the parts where it would be wrong-and-plausible:

- The four scale conversions agree with `packed_sd`'s own declared→packed factor, so the
  `|p|`-times-RSE delta method is applied **once**, not twice. Ω-variance→`(sd)` gives
  `Rse(0.5·rel)` onto a `Log` coordinate; Σ-SD→variance gives `Rse(2·rel)` onto `HalfLog`.
  Both check out by hand.
- `se_kappa` really is indexed by diagonal position (`v.get(i)`), matching
  `write_estimates_yaml:2357` — the block-Ω lower-triangle trap that `omega_se_at` exists for
  does not apply to κ. Good that `from_fit_result` uses `omega_se_at` for Ω and a plain index
  for κ rather than one rule for both.
- The YAML reader's "skip an entry that lacks the section's value key" is the right way to drop
  `A__B:` off-diagonals — it does not depend on recognising the `__` spelling, and
  `parameter_priors:` / `bayes:` / `sir:` / `covariate_model:` all fall out for free because
  they are not in `SECTIONS`.
- Moving the read out of `PriorSet::build` is the right call, and the
  `fit_refuses_a_model_whose_from_fit_was_never_expanded` straddle is the assertion that makes
  it safe rather than just tidy.

Seven findings, none blocking.

---

### 1. `src/estimation/priors.rs:527` — the inline-override skip matches on name only, ignoring family

```rust
if model
    .priors
    .iter()
    .any(|p| p.name.eq_ignore_ascii_case(&info.name))
```

Every other match in this feature is on name **and** kind — that is the headline invariant of
the PR, and `CoordInfo::kind` was added for it. This one predicate drops the kind.

Scenario: a model declaring both `theta CL(...) prior(1.0, rse = 20%)` and `omega CL ~ 0.09`,
importing from a source that reports both. The typed θ prior suppresses the import of **Ω CL**,
which then carries no prior at all — silently, since the override skip is deliberately silent.
The user asked to override one prior and lost a different one.

`info.kind` is right there; `p.kind.is_none_or(|k| k == info.kind) && p.name.eq_ignore…` restores
the invariant (the `is_none_or` keeps a hand-built kind-less prior overriding as today).

The existing test `an_inline_prior_overrides_the_imported_one` uses `TVCL` on both sides, so it
cannot see this. The straddle that would is the same fixture
`a_name_that_matches_a_different_family_is_not_imported` already builds.

### 2. `src/estimation/priors.rs:511` — an ambiguous match is skipped with no note

```rust
let [info] = matches.as_slice() else { continue };
```

`>1 match` falls into the same silent arm as `0 matches`, but the PR's stated contract is that
*absent* is the only silent skip and everything else lands in `parse_warnings`. Today this is
close to unreachable (`coordinate_names` suffixes mixture overrides `_MIX{n}`, and off-diagonals
are `A~B`), so it is a latent hole rather than a live bug — but it is exactly the shape that
stops being unreachable when a future segment is added, and the `while out.len() < names.len()`
tail was written to make that case loud. Worth splitting the `else` into `[] => continue` and a
`_ =>` that pushes a note.

Related: the comment at :514 says the rejection check is "ordered first so the `[mixture]`…tail is
never matched on kind" — but the kind filter at :509 runs *before* the rejection check, so the tail
can enter `matches`. It is harmless because of the `_MIX{n}` naming, but the comment describes an
ordering that isn't there.

### 3. `docs/estimation/priors.qmd:187` — docs say the inline-override skip is warned about; it isn't

The bullet list is introduced with "Skipped, **each reported in the fit's warnings** so a missing
prior is visible rather than merely absent", and then includes "a parameter that already carries an
inline `prior(...)` here". That skip is silent by design (priors.rs:522-529, and the PR body says
so explicitly). A user following the docs would look for a warning that never arrives. Either move
that bullet out of the warned list or say "(silent — it was deliberate, and it shows in
`prior_summary`)".

### 4. `src/parser/model_parser.rs:3145` — an empty `[priors]` block is silently a no-op

`parse_priors_block` returns `Ok(None)` for a block with no lines, and `.flatten()` turns that into
`prior_from_fit = None` — no error, no prior. A `[priors]` block containing only comments behaves
the same (`extract_blocks` strips comments before the block body is collected).

That is the failure mode the function's own doc comment argues against for a typo'd key: "a typo'd
`from_fit_file =` would parse cleanly, import nothing, and produce an unpenalized fit that looks
exactly like a penalized one." An empty block is the same outcome by a different route. A block that
is present but declares nothing should be the same hard error.

### 5. `src/api/covariate_stats.rs:67` (and `src/api/levels.rs:83`) — the source fit is read twice for some models

Both re-parse the model text through `parse_full_model_with(model_text, &parsed.bindings)` after
binding data-derived statistics, which runs `expand_prior_from_fit` a second time. So for any model
with a symbolic `[covariate_model]` centre or level bindings, the "read exactly **once**, when the
model is parsed" guarantee — stated in `docs/estimation/priors.qmd`, in
`CompiledModel::prior_from_fit`'s doc comment, and in the PR body — does not hold, and the mid-parse
window it was meant to close is reopened (narrower, but present).

`parsed.bindings` is mutated in place so `model_dir` does survive the rebind, which is the part that
would have been an actual bug. This is a docs-vs-code mismatch plus one extra file read; worth
either narrowing the claim or having the rebind carry the already-expanded priors forward.

### 6. `api/ferx-core-public-api.txt` — three new public *fields*, but "Breaking changes: Additive only"

The baseline adds `ParseBindings::model_dir`, `CompiledModel::prior_from_fit` and
`ParameterPrior::kind`. None of those structs is `#[non_exhaustive]` and all their fields are `pub`,
so any external struct-literal construction of them stops compiling. The PR's Breaking-changes table
lists only `CompiledModel::prior_from_fit`, and under "Added item / Needed by" it says ferx-r
"constructs `CompiledModel` rather than always going through the parser".

That justification does not hold: `ferx-r/src/rust/src/lib.rs` only ever takes `&CompiledModel`
(nine borrow sites, no literal, no `ParseBindings` use at all), so nothing downstream breaks today —
but nothing downstream needs `expand_prior_from_fit` or `prior_from_fit` public either, which is the
argument the public-API gate asks for. Either restate the justification (the field has to be `pub`
because `CompiledModel` is a `pub` all-public-fields struct the parser fills in, not because ferx-r
constructs one) or keep `expand_prior_from_fit` `pub(crate)` until a caller exists.

### 7. `examples/warfarin_update.ferx:48` — the example does not parse from a clean checkout

`from_fit = "warfarin-fit.yaml"` names a file that only exists after someone runs
`examples/warfarin.ferx` from `examples/`, and `{model}-fit.yaml` is git-ignored. Since the import
now happens *during* parsing, this example is unparseable in CI and in a fresh clone.

Nothing breaks today — `tests/model_text_roundtrip.rs` uses `ModelText::parse` (pure text, no
filesystem) and `parallel_mixed_absorption.rs::shipped_example_models_parse` uses an explicit
allow-list. But it is the first example in `examples/` with that property, so any future
corpus-wide `parse_full_model_file` test, and the ferx-site page render this PR flags as a
follow-up, will trip on it. Worth a committed fixture fit file (or a `# not runnable standalone`
note the corpus test can honour) rather than discovering it downstream.

---

Nothing above blocks merge; #1 is the one I would fix before merging, since it is a silent
wrong-answer in exactly the name-collision case this PR added `ParameterKind` to handle.

https://claude.ai/code/session_014araEMYZa4M4xPCrDU3Bog

roninsightrx:
Good review. Five of the seven were real; fixed in `797798da`, plus a PR-body correction for the sixth. Taking them in order.

**1. Inline-override skip ignores kind — confirmed, and worse than described.**

This is a hole **this PR's own fix opened**, which is why I missed it: before the family travelled with a prior, a typed `prior()` on `theta CL` in a model that also declared `omega CL` was refused as *ambiguous*, so the fit stopped. Fixing that ambiguity made the typed prior resolve — and the name-only override predicate then silently suppressed the `omega CL` import, leaving a fit that looks exactly as though it had not. The one predicate still resolving by name, in the change whose entire point was that names are not enough.

Now matched on family as well as name. A prior naming no family still suppresses by name alone, deliberately and conservatively: it cannot say which it meant, and in a colliding model it refuses to resolve anyway.

You were right that the existing override test could not see it — same family on both sides. Added `an_inline_prior_overrides_only_its_own_family`: typed on the θ, source carrying both, asserting **two** priors survive and each landed on its own coordinate. Mutated (drop the family term) — reddens.

**2. `>1 match` folded into the silent "absent" arm — confirmed.**

Agreed on both halves. Latent today, but folding it into the arm that is silent *by design* means a future packed-layout change drops a prior without a word, which is the one outcome this feature must never produce. Now a note rather than a silent skip, driven in `a_source_estimate_matching_two_coordinates_is_noted_not_dropped` (two η under one name, reached directly since the parser cannot produce it). Mutated — reddens.

And the comment was simply wrong: the kind filter runs *first*, and it is the rejection check that stops the `[mixture]` tail. Reworded to say what the code does.

**3. Docs list a silent skip under "reported in warnings" — confirmed.** Split into two lists, reported and silent, with the reason for each.

**4. Empty `[priors]` block — confirmed, and it needed a second fix to land.**

My first attempt keyed off `blocks.get("priors")` and did nothing, because `extract_blocks` does not record a block in `unnamed` until it has a *content* line — a bare `[priors]` header is absent from `blocks` entirely. Keyed off `block_lines` instead, so the empty case reaches `parse_priors_block` and errors. Covered for empty, comments-only and whitespace-only.

**5. Double expansion on the rebind — confirmed; a documentation defect, not a behaviour one.**

`parse_full_model_with(model_text, &parsed.bindings)` re-parses from text, so it legitimately re-expands, and there is nothing to make idempotent. `model_dir` does survive (`ParsedModel.bindings` is `bindings.clone()`), as you said.

What was wrong was my claim, in three places, that the file is read "exactly once". It is once per *parse*, and a model with symbolic covariate centring or level bindings is parsed twice. The substantive guarantee is unaffected — both reads precede any estimation, so the optimizer, the covariance step and the report still see the same priors — so I corrected the wording rather than the code, and the docs now name the re-parse explicitly.

**6. "Additive only" is wrong — confirmed, and so was my justification.**

Three of the five added items are `pub` fields on structs that are neither `#[non_exhaustive]` nor privately-fielded, so external struct-literal construction breaks. PR body rewritten: the breaking box stays ticked, the table now says per item whether it breaks a literal, and there are real migration notes (`..Default::default()` for `ParseBindings`; `prior_from_fit: None` / `kind: None` otherwise — and `kind: None` reproduces the old behaviour exactly).

You were also right about ferx-r. I checked rather than assumed this time: `src/rust/src/lib.rs` only ever takes `&CompiledModel` and mentions neither `ParameterPrior` nor `ParseBindings`, so nothing there constructs these. The old table's claim that "the R wrapper constructs `CompiledModel` rather than always going through the parser" was inherited from #1341's framing and is not true of the current wrapper.

That leaves `expand_prior_from_fit` with a weaker justification than the gate normally wants, and the body now says so plainly: **no caller needs it today.** It exists because the `PriorSet::build` guard's error message names it, and without it the public `prior_from_fit` field is a trap — you can set it and nothing but a parse will honour it. The alternative is `pub(crate)` + reword the guard to "parse the model with `parse_model_file`" + document the field as parser-only. I have no strong preference; say which you want and I will make it so.

**7. Example unparseable from a clean checkout — acknowledged, no change.**

Correct, and intended: the import happens during parsing, so the model genuinely depends on a file that does not exist until the parent runs. I would rather that fail at parse (`ferx check` catches it) than defer. In-repo it is handled — `test_parse_all_example_ferx_files` skips `[priors]` examples with a comment naming `the_update_example_names_a_file_the_documented_commands_produce`, which stages a parent fit in a temp dir and parses the real example.

Your point about a future corpus-wide parse or a site render is fair. No live consumer today (`ferx-nlme.github.io` does not reference `warfarin_prior.ferx` either, and the ferx-site / ferx-book follow-up boxes in the PR are already unticked for exactly this example), so I have flagged it rather than built machinery for it. If the site render does pick these up, the fix is for that harness to run the parent first — the same two commands the example documents.

**Verification on `797798da`**

- `tools/preflight.sh` — green
- `cargo test --features ci` — 5286 passed, 0 failed, 178 ignored, 186 binaries
- `cargo test --features ci,slow-tests --test parameter_priors` — 26 passed

Issue #432, opened 2026-06-20
estimation(covariance): FD R-matrix SEs over-optimistic & FD-fragile on flat high-shrinkage variance directions


hiddevandebeek:
## Summary

The finite-difference **R-matrix** (`$COVARIANCE MATRIX=R`, `CovarianceMethod::Hessian`, the default) produces standard errors that are **~25% over-optimistic vs NONMEM on flat, high-shrinkage variance directions**, and the FD Hessian is **numerically fragile** there (it can blow up at an intermediate step). The delta-method transform itself is **correct** — this is the curvature estimate on the flattest direction, not a Jacobian/index bug.

Surfaced while reconciling the block-omega FOCE covariance test in #423.

## Reproducer

Warfarin, `omega = BLOCK(2) on (CL,V)` + diagonal `omega ETA_KA ~ 0.30`, FOCE (no INTER), `$COV MATRIX=R`, gradient outer optimizer. SEs vs NONMEM 7.5.1 `MATRIX=R`:

| config | SE(TVKA) | SE(ω²CL) | SE(ω²V) | **SE(ω²KA)** |
|---|---|---|---|---|
| **NONMEM MATRIX=R** | 0.12445 | 0.01280 | 0.00430 | **0.16054** |
| ferx R, fd_step=1e-2 (default) | 0.12591 | 0.01252 | 0.00412 | **0.12002** |
| ferx R, fd_step=5e-3 | 0.12577 | 0.01249 | 0.00412 | **0.11768** |
| ferx R, fd_step=2e-3 | 197.4 | 49.2 | 28.9 | **1017.6** 💥 |
| ferx R, fd_step=1e-3 | 0.15134 | 0.01184 | 0.00385 | **0.12095** |
| ferx S (CrossProduct) | 0.19016 | 0.02372 | 0.01074 | **0.18494** |
| ferx RSR (Sandwich) | 0.12903 | 0.01018 | 0.00342 | **0.11188** |

Every other SE matches NONMEM within ~4%; only **SE(ω²KA)** — the highest-shrinkage variance — is off (ferx 0.120 vs NONMEM 0.160).

## What it is, and isn't

[...]

Issue #243, opened 2026-06-10
feat(estimation): switch FOCE (non-interaction) from Sheiner-Beal linearised marginal to interaction-free Almquist-Laplace (match NONMEM METHOD=1)


hiddevandebeek:
> [!IMPORTANT]
> **Correction (superseded diagnosis).** The analysis below — that ferx's FOCE
> uses a Sheiner–Beal marginal where NONMEM uses an interaction-free
> Almquist–Laplace marginal, and that the fix is to switch the estimation
> objective — turned out to be **wrong**. A NONMEM `METHOD=1` (no INTER)
> cross-check showed ferx's FOCE **estimates already match NONMEM** (OFV −280.17
> vs −280.36; θ within ~1%): the Sheiner–Beal objective *is* NONMEM's FOCE
> objective. The real defect was in the **covariance step**, not the estimator:
> it added the Ω prior (`η̂ᵀΩ⁻¹η̂ + log|Ω|`) on top of the SB marginal, which
> already carries Ω via `R̃ = HΩHᵀ + R` (Woodbury) — double-counting Ω and
> leaving the FOCE omega SEs ~31% low. The estimation objective is **unchanged**.
>
> Fixed in **#249** (remove the double-count; also exclude structural-zero
> block+diagonal Ω off-diagonals from the covariance). The "Proposed fix" and
> "Interconnections" sections below are obsolete — in particular #99 and #134 are
> **not** related; the relevant cluster is the covariance work #234 / #209 / #196
> / #129 and #242. See #249 for the relationship map.
>
> ---

## Summary

ferx's **FOCE (non-interaction)** estimator uses the **Sheiner–Beal linearised marginal**

```
(y − f₀)ᵀ R̃⁻¹ (y − f₀) + log|R̃|,   R̃ = HΩHᵀ + R(f₀),   f₀ = ipred − H·η̂
```

whereas NONMEM's FOCE (`$EST METHOD=1`, no `INTER`) is the **Almquist/Laplace marginal with the interaction term dropped**

```
data_ll + ηᵀΩ⁻¹η + log|Ω| + log|H̃|,   H̃ = HᵀR⁻¹H + Ω⁻¹   (no ½·c̃ᵀc̃)
```

[...]

PR #1364, opened 2026-09-12
fix(estimation): hold mixed block_omega / block_kappa structural zeros in the outer optimizer [closes #1018]

### LINKED ISSUE #1018: fix(estimation): FOCE/FOCEI estimate the full lower triangle on a mixed block_omega, ignoring declared structural zeros

TeunP:
Closes #1018

## Why

A `block_omega` declared next to a separate diagonal `omega` declares the covariances between the two as structural zeros (`docs/model-file/parameters.qmd`, "Mixing diagonal and block"). FOCE/FOCEI estimated them anyway. `block_omega (ETA_CL, ETA_V)` + `omega ETA_KA` returned non-zero `Cov(ETA_KA, ETA_CL)` and `Cov(ETA_KA, ETA_V)`, and landed on the full 3×3 block fit: same Ω, same OFV (reported from ferx-r at ferx-core `8694824`: FOCE −283.3167 and FOCEI −288.9464 for both the partial and the full block). No error or warning.

Cause: `pack_params` gives every lower-triangle Cholesky entry of a non-diagonal Ω a packed coordinate, including the cross-block ones. The two things that keep a coordinate still — `pack_with_bounds` pinning `lower == upper`, and every optimizer zeroing its gradient — both read `packed_fixed_mask`, which only knew FIX. The structural-zero mask (`omega_structural_zero_mask`) reached the covariance step (#243), VI, and the `n_parameters` count (#1177) — so `n_parameters` already said 8 while the optimizer searched 10. From the history, the optimizer never held these entries, so this is not a regression.

## What changed

- `packed_fixed_mask` now ORs in `omega_structural_zero_mask` (BSV Ω and Ω_IOV). It is the one mask the bounds pin, the gradient zeroing (outer optimizer, trust region, Gauss-Newton — including the finite-difference fallback that every IOV model takes, since the analytic dispatcher requires `kappas.is_empty()` — and AGQ), the covariance free set, SIR and the asymptotic uncertainty sampler all read, so one edit covers all of them. A held `L[i,j] = 0` keeps `Ω[i,j] = 0` exactly: the Cholesky factor of a matrix that is block-diagonal under permutation has no fill-in across the blocks.
- `packed_held_mask` (`pub(crate)`) was now identical to `packed_fixed_mask` and is deleted; `free_packed_dim`, `fit()`'s `n_parameters` and the BIC tally call `packed_fixed_mask`.
- The second structural filters in `covariance.rs` (free set), `vi/run.rs` (`param_criterion_applies`) and Gauss-Newton's two gradient builders (`!free_mask[(row, col)]`) are collapsed into the one mask. Two gates rejecting the same inputs mean either one can be deleted and no test notices.
- `pack_params` packs a structural zero as **0** whatever the template carries there, so `pack_params`, `compute_bounds` and `pack_with_bounds` agree on the coordinate a held slot holds. A template carrying a stale non-zero value — estimates copied in from a full-block fit, a hand-built `ModelParameters`, a pre-#1018 `.fitrx` reloaded by `run_covariance`, whose fallback arm centres on `pack_params` while its box comes from `pack_with_bounds` — would otherwise put the centre outside its own pinned box and freeze that covariance for the whole fit. `unpack_params` carries a `debug_assert` that the reconstructed `Ω[i,j]` really is 0 at every structural slot, which holds while `free_mask` is a disjoint block partition (it would not for an overlapping-block mask, which the parser does not reject for `block_omega` the way it does for `block_kappa`).
- No public API change (`packed_fixed_mask` keeps its signature; the baseline is unchanged).

## Alternatives considered

- **Pass `packed_held_mask` to each optimizer site instead of changing `packed_fixed_mask`.** That is ~15 call sites across the outer optimizer, trust region, GN, AGQ, sens gradient, SIR and uncertainty sampling, and a missed one reintroduces the bug silently. Every production consumer of the FIX mask was checked: none needs FIX-only semantics (`check_variance_init_rails` only looks at Ω diagonals; the runaway guard, SIR and uncertainty sampling all want structural zeros excluded).
- **Pack only the free off-diagonals** (shorter packed vector). Changes `packed_len`, the covariance-matrix layout, `se_omega` indexing and ferx-r's reading of all three — a much larger, public-facing change for the same result.

## Cross-repo dependency
| ferx (R) | FeRx-NLME/ferx-r#359 | open | independent — print skips zero-covariance pairs; correct at any pin |
| ferx (R) | lock bump (not yet opened) | — | after — `tools/update-ferx-core-lock.sh` once this merges |

## Breaking changes
- [x] None

Behavioural change, called out in CHANGELOG: estimates, OFV and AIC/BIC of every mixed block + diagonal Ω fit change; `n_parameters` does not.

## Numerical validation
- [x] Compared against: NONMEM / prior ferx commit `8694824`
- [x] Reference values:

`examples/warfarin_block_omega.ferx` + `data/warfarin.csv` (FOCE, default optimizer, covariance on) — the issue's own reproduction:

```
# before — main @ ae6c208c, as reported in #1018
n_parameters: 10
ETA_CL  variance 0.028695    ETA_V 0.009629    ETA_KA 0.350189
ETA_V__ETA_CL    covariance  0.001754  (correlation  0.104575)
ETA_KA__ETA_CL   covariance -0.031659  (correlation -0.282055)   <- not declared
ETA_KA__ETA_V    covariance  0.020686  (correlation  0.328136)   <- not declared

# after — this branch @ c9034f01
ofv: -280.485777   aic: -264.485777   bic: -242.881934
n_parameters: 8
ETA_CL  variance 0.028599 (SE 0.012800)
ETA_V   variance 0.009578 (SE 0.004298)
ETA_KA  variance 0.348879 (SE 0.160753)
ETA_V__ETA_CL    covariance 0.001825 (SE 0.005278, correlation 0.109344)
# no ETA_KA__* rows at all — `io/output.rs` omits a pair at |cov| <= 1e-15
# covariance_matrix rows chol_ETA_KA_ETA_CL / chol_ETA_KA_ETA_V are identically 0
```

NONMEM 7.5.1 FOCE, `$OMEGA BLOCK(2)` + diagonal KA, `$COVARIANCE MATRIX=R` (the model NONMEM actually fits — its KA row has no covariance): `tests/warfarin_covariance_nonmem.rs::covariance_se_matches_nonmem_foce_block_omega`.
That test is `slow-tests`-gated, so it does not run on PR CI; it is **dispatched on this branch** as run [34662791304](https://github.com/FeRx-NLME/ferx-core/actions/runs/34662791304).

**That run found two tests that encoded the old behaviour; both are updated here, from its measured values.**

1. **`covariance_se_matches_nonmem_foce_block_omega`** anchors NONMEM's `SE(ω²KA) = 1.60542e-1` and kept a ferx-internal window `0.105 ..= 0.135` around the ~0.120 ferx used to produce — a ~25% gap #432 attributes to the FD R-matrix on the flattest curvature direction. Measured on that run: **0.160862**, i.e. **0.2%** from NONMEM. The gap #432 describes was this bug: the undeclared cross-block covariances were being estimated, and the extra freedom flattened ω²KA's curvature. The component now takes the same relative band as the other Ω SEs, and the internal window is re-measured to `0.150 ..= 0.172` (the measured value ±7%).
2. **`simultaneous_stepwise_agrees_on_step_one_and_diverges_on_the_mixed_omega`** (ferx-tools Pharmpy anchor) existed to assert the #1018 divergence, with "when #1018 lands, this assertion flips" written into it. Measured: ferx's `[CL,V]+[KA]` candidate moves from 647.73 (fitted as the full `[CL,V,KA]` block) to **655.0218**, against NONMEM's 655.4711 for the declared model — a run that itself terminated with unreportable significant digits, so the anchor now asks for agreement within 1 OFV unit rather than equality. The search also ends on `[CL,V]`, as Pharmpy does. Renamed to `simultaneous_stepwise_follows_pharmpy_on_the_mixed_omega`, and verified locally.

Both changes are behavioural consequences of the fix rather than tolerance loosening: one band gets **tighter** (30% → 20% relative, with a 4× narrower absolute window), and the other flips a divergence assertion into an agreement assertion.

## Performance
- [x] No significant impact expected — the free set shrinks by the structural zeros.

## Tests

- **Tier 1** (`parameterization.rs`): `packed_fixed_mask` holds the BSV and Ω_IOV cross-block entries; `pack_with_bounds` pins them to `[0, 0]` while the in-block covariance keeps `[-10, 10]`; a full block holds nothing; a template whose structural slot starts at 0.01 is pinned at 0 (the fixture asserts its `pack_params` slot is non-zero, so a pin at the start value fails).
- **Tier 2** (`tests/block_plus_diagonal_omega_structural_zeros.rs`), capped at 4 outer iterations:
  - BOBYQA, L-BFGS, SLSQP, BFGS and trust region × block declared before / after the diagonal η: cross covariances `== 0.0` exactly, `n_parameters == 8`, and the in-block covariance moves relative to an `outer_maxiter = 0` fit of the same model (not relative to the declared 0.02, which the `exp(ln L)` round trip does not reproduce bit-for-bit).
  - **Straddle:** the same start written as a full block moves a KA covariance off 0 within the same cap, and its OFV differs — so the zero assertions cannot pass on an optimizer that never stepped.
  - **Ω_IOV:** `block_kappa (KAPPA_CL, KAPPA_V)` + `kappa KAPPA_KA` on `warfarin_iov.csv` (L-BFGS; the trust region does not take IOV): κ cross covariances exactly 0, in-block κ covariance moves, `n_parameters` 11; a full `block_kappa` moves a cross covariance and counts 13.
  - Covariance step: `se_omega` has 6 entries (full lower triangle, #226), the two structural entries are exactly 0, and the covariance matrix has exactly `n_parameters` non-zero diagonal rows.
- **Mutation** (byte copy, `--no-fail-fast`, restore by `cp` + `touch`, sha verified identical afterwards). Each side of the hold is mutated separately, so neither can cover for the other:

  | Mutant | Change | Killed by |
  |---|---|---|
  | M1 | structural zeros not held at all (OR loop removed) | both Tier-1 mask tests + all 3 Tier-2 tests |
  | M2 | pinned at the template's start value instead of 0 | the Tier-1 stale-slot test only — the case written for it |
  | M3 | mask holds them, `pack_with_bounds` does not pin them | Tier-1 pin test + the Tier-2 **BOBYQA** arm (`Ω[ETA_KA,ETA_CL] = -3.35e-2`) |
  | M4 | mask FIX-only, box still pins them | both Tier-1 mask tests + all 3 Tier-2 tests (`n_parameters` 10 vs 8, 13 vs 11) |

  Measured while doing it: under M3 the Ω_IOV test still passes, so with L-BFGS the κ hold comes from the zeroed gradient alone and it is BOBYQA — derivative-free — that makes the pinned box load-bearing.

- [x] Not applicable (internal / refactor / fix with no new API surface) — `examples/warfarin_block_omega.ferx` is the reproducer.

## Docs
- [ ] `docs/` updated for user-visible changes — `parameters.qmd` already documents the intended behaviour
- [x] `CHANGELOG.md` `[Unreleased]` entry added (Fixed)

## Checklist
- [x] `tools/preflight.sh` green — `preflight OK — fmt check clippy rustdoc docs public-api`; the public-API baseline matches the committed file (no surface change: `packed_fixed_mask` keeps its signature, `packed_held_mask` was `pub(crate)`)
- [x] Full suites, `--no-fail-fast`, `ci-fast`: `--workspace --features ferx-core/ci --lib` → 4394 + 602 + 61 passed, 0 failed; `--workspace --features ferx-core/ci --tests` → 208 binaries, 6076 passed, 0 failed; `--features ci,markov,nn --tests` → 185 binaries, 5719 passed, 0 failed
- [x] Functions on the `Dual2` sensitivity path stay differentiable (no kernel change; only which packed coordinates are held)

### ferx-book
- [ ] `chapters/16-variability.qmd` documents the bug at the pinned build ("Partial blocks at this build"); rewrite after the ferx-r lock bump and book pin bump.

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core — follow-up with the lock bump

## Reviewer hints

- A pre-PR review round is on the issue ([#1018 comment](https://github.com/FeRx-NLME/ferx-core/issues/1018#issuecomment-5640799817), 11 findings). Addressed here: the Tier-2 positive control, the duplicated fits, an Ω_IOV fit case, the Gauss-Newton duplicate gate, pinning structural zeros at 0, the CHANGELOG wording and the `PackedStart` docs. Not a defect: the BOBYQA cap (its budget is `outer_maxiter·(n+1) + 40·(n+1)` evaluations). Skipped: hoisting the mask out of the Gauss-Newton per-subject builders (two small allocations per subject; it needs signature changes in three functions).
- A second, independent review round is on the PR (8 findings): the Gauss-Newton claim in the CHANGELOG was wrong (its FD fallback never held these entries, so `gn` and trust-region estimates move too); `pack_params` and `compute_bounds` disagreed for a stale template; and the FOCEI / `gn` builders whose gate the diff deletes had no test arm. All three are fixed here, with the new `focei_and_gauss_newton_builders_hold_structural_zeros` arms and a per-builder mutation.
- The fix is the OR loop at the end of `packed_fixed_mask`; everything else is removing the second copies of the same exclusion.
- Worth a look: whether any consumer of `packed_fixed_mask` / `PackedStart::fixed` wanted FIX-only semantics. I read all production consumers (list in "Alternatives considered").
- `covariance_se_matches_nonmem_foce_block_omega` (slow) keeps a ferx-internal window around SE(ω²KA) that was set at the old optimum; see Numerical validation for what it does now.

## Open questions

None.


TeunP:
## Independent review round (subagent context, verified here)

Reviewed at head `c9034f01`. Eight findings; I verified each load-bearing claim in the worktree before posting. Green CI, the four-mutant kill and the full suites are not evidence for any row below.

| # | Finding | Severity | Evidence | Disposition |
|---|---|---|---|---|
| 1 | The CHANGELOG and PR body say "(Gauss-Newton, SAEM and VI already held them)". Only GN's two **analytic** builders held them. The FD fallback `subject_nll_pop_grad` never had the gate, and the analytic dispatcher requires `kappas.is_empty()` plus no M3 / no RUV-η / no θ-dependent magnitude, so every IOV model always took the FD path. Pre-fix `method = gn` on a `block_kappa` + `kappa` model did search the κ cross covariances, and `optimizer = trust_region` reaches the same builder. GN and trust-region estimates change too. | should-fix | `gauss_newton.rs` dispatcher gating; FD fallback has no `free_mask` gate; `trust_region.rs` gradient path | fix: reword, and list GN / trust region among the estimators whose estimates move |
| 2 | `pack_params` (public) and `compute_bounds` (public) now disagree for a template carrying a stale non-zero structural value: the start keeps the value, the box is pinned at `[0, 0]`. Live path: `run_covariance`'s reloaded-`.fitrx` arm takes `x_hat = pack_params(&base_params)` while `compute_covariance` takes bounds from `pack_with_bounds(template)`, so re-running the covariance step on a pre-#1018 `.fitrx` of a mixed-block model evaluates at a centre outside its own box. No panic (the index is excluded from the free set), but `run_sir` re-centres at 0 and `run_covariance` does not. | should-fix | `run_covariance.rs` fallback arm; `covariance.rs` bounds source | fix: zero the structural slots in `pack_params` itself, one place, and the three walks agree again |
| 3 | The diff deletes a gate on a path no test exercises. The FOCEI-only Laplace-cached builder lost its `free_mask` gate, but every Tier-2 arm sets `interaction = false`, and the only block-Ω unit test for that builder uses a full 2x2 block. No arm runs `method = gn` / `gn_hybrid` either, which per finding 1 is the path whose behaviour changed most. The M1 mutation cannot distinguish: it reddens through the FOCE arms. | should-fix | Tier-2 sets `interaction = false`; `gauss_newton.rs` unit fixture is a full block | fix: add a FOCEI arm and a `method = gn` arm, then mutate each builder's gate separately |
| 4 | The soundness argument holds only if `free_mask` is a disjoint block partition. It fails for a chain mask (A-B free, B-C free, A-C structural), where `Ω[C,A] = L[C,B]·L[B,A]` is non-zero. The parser rejects a name in both `kappa` and `block_kappa`, but has no equivalent rejection for an η in two `block_omega` declarations. | nit | parser block-omega branch pushes names with no duplicate check | fix: `debug_assert` the reconstructed Ω is 0 at every structural slot, or reject overlapping blocks |
| 5 | The new OR zips a push-built mask (length from `theta_fixed` / `sigma_fixed` / …) against a mask sized from `packed_segments().total()` (length from `theta` / `sigma.values` / …), guarded only by `debug_assert_eq!` — and `ci-fast` inherits `release`, so debug assertions were **off** in every suite run here. A hand-built `ModelParameters` with a short `*_fixed` vector would truncate the zip or misalign the hold. | nit | `parameterization.rs` OR loop; `Cargo.toml` profile chain | fix: size the mask from `packed_segments`, or promote to `assert_eq!` |
| 6 | `pack_with_bounds_agrees_with_the_three_separate_walks` still documents "bit-for-bit … nothing downstream can tell the difference", now conditional, and its template has no structural zero, so it cannot see the exception. | nit | that test's template | fix: add a structural zero there (moot if finding 2 is fixed in `pack_params`) |
| 7 | Behaviour changes missing from the CHANGELOG: SIR's Student-t dimensionality `d = n_free` shrinks (weights and ESS move); public `compute_bounds` now returns `[0,0]` where it returned `[-10,10]`; the runaway guard can no longer fire on a structural coordinate. | nit | `sir.rs`, `parameterization.rs`, `postfit.rs` | fix: one more clause |
| 8 | `tests/n_parameters_free_packed_dim.rs`'s header describes a mutation that no longer exists (the two masks are one function now). | nit | that file's header | fix: restate what it pins |

**Checked and clean** (reported in full, not a summary of green CI): every production consumer of `packed_fixed_mask` / `PackedStart::fixed` — runaway guard, `check_variance_init_rails` (scope gate is Ω diagonals, disjoint from structural off-diagonals), `check_packed_start_in_box` (`[0,0]` is not an inverted box, and `packed == lower == upper` is not "outside"), SIR free subspace (previously the structural index sat inside the free set with a floored covariance row and got perturbed by ~sqrt(floor) — now correctly excluded), asymptotic draws, AGQ FD scores, `sens_outer_gradient`, the four outer-optimizer assembly points, `bic_inputs_for`, VI, and the CRS2 pre-search (zero-width box already occurs for FIX, and the rescale helper floors at 1.0). No consumer in `crates/` or in ferx-r's `src/rust`. SAEM / VI / Bayes keep their own Ω-space `free_mask` gates, which are a different mechanism and correctly untouched. Layout: `packed_segments` is the single derivation behind both `packed_len` and the structural mask's offsets, so mixture overrides, `block_sigma` ρ (packed last), Ω_IOV, FIXed blocks, 1x1 Ω and `n_eta = 0` all line up. Test shapes: no tautological twin, no NaN-eating fold (`assert_moved` asserts `is_finite()` first), no empty fast path; the Tier-1 stale-slot case carries its own straddle.

Resolving 1, 2 and 3 next, then the nits.

TeunP:
## Resolve round — all eight findings, plus two the slow-tests dispatch found

Head `2912926e`. Everything below was measured, not reasoned; where a claim came back different from the review's hypothesis I say so.

| # | Finding | Fix |
|---|---|---|
| 1 | CHANGELOG wrong about Gauss-Newton | `2912926e` — verified: the FD fallback never had the gate, and the analytic dispatcher requires `kappas.is_empty()` (plus no M3 / RUV-η / θ-dependent magnitude), so every IOV model took the FD path. CHANGELOG now says GN held them only on its analytic path, and names `gn` / `gn_hybrid` / trust region among the estimators whose estimates move |
| 2 | `pack_params` and `compute_bounds` disagree for a stale template | `2912926e` — the zeroing moved into `pack_params`, so the three walks agree and `run_covariance`'s reloaded-`.fitrx` centre matches its box. `pack_with_bounds` is back to its plain pin and the doc exception is gone (finding 6 with it) |
| 3 | deleted gate on a path no test exercises | `2912926e` — new `focei_and_gauss_newton_builders_hold_structural_zeros`: FOCEI on L-BFGS and trust region, `method = gn`, and `gn` on the IOV model (the FD fallback). See the mutation note below — the result is not what I expected |
| 4 | chain mask breaks the soundness argument | `2912926e` — `unpack_params` carries a debug assertion that every structural slot reconstructs `Ω[i,j] == 0`. The parser's missing overlapping-`block_omega` rejection is **not** fixed here; proposing it as a follow-up issue rather than widening this PR |
| 5 | mask length provenance | `2912926e` — every segment is walked over the count `packed_len` derives it from. Confirmed the guard was inert: `ci-fast` inherits `release`, so the `debug_assert` never armed in any suite I ran |
| 6 | "bit-for-bit" doc now conditional | moot with finding 2 |
| 7 | behaviour changes missing from the CHANGELOG | `2912926e` — SIR's Student-t dimensionality, `compute_bounds` returning `[0,0]`, and the `pack_params` change are all named |
| 8 | stale test header | `2912926e` — restated as what it now pins |

**Per-builder mutation, reported as measured.** Deleting the FOCE analytic builder's `fixed_mask` gate (MA) and deleting the FOCEI Laplace-cached builder's gate (MB) both leave the whole suite **green**, including the new arms. So those two gates are not load-bearing: with the box pinned at `[0, 0]` a clamped step keeps the slot at 0, and the NLopt paths also zero the assembled gradient by the same mask. That supports removing them as duplicates, but I am not claiming a kill I did not get. What is pinned by its own mutant: the mask (M1, M4), the box (M3, which reddens the BOBYQA arm), and the new `pack_params` zeroing (MD, which reddens only the Tier-1 stale-slot case).

**The dispatched slow-tests run found two tests that encoded the pre-fix behaviour.** Both are updated from its measured values, and neither is a loosened tolerance:

1. `covariance_se_matches_nonmem_foce_block_omega` — `SE(ω²KA) = 0.160862` against NONMEM's `1.60542e-1`, a **0.2%** difference. ferx used to report ~0.120, and #432 reads that ~25% gap as an FD flat-direction limitation of the R-matrix. It was this bug: the undeclared cross-block covariances were being estimated, and that extra freedom flattened ω²KA's curvature. The component now takes the same relative band as the other Ω SEs (30% → 20%), and the ferx-internal window is re-measured to `0.150 ..= 0.172` from `0.105 ..= 0.135`. **#432 looks resolved by this PR** — worth closing or re-scoping after merge.
2. ferx-tools `iivsearch_pharmpy_anchor::simultaneous_stepwise_agrees_on_step_one_and_diverges_on_the_mixed_omega` — written to assert the #1018 divergence, with "when #1018 lands, this assertion flips" in it. Measured: the `[CL,V]+[KA]` candidate moves from 647.73 (fitted as the full `[CL,V,KA]` block) to **655.0218**, against NONMEM's 655.4711 for the declared model — a run that terminated with unreportable significant digits, so the anchor now asks for agreement within 1 OFV unit. The search ends on `[CL,V]`, as Pharmpy does. Renamed accordingly. `Space::outer_full_triangle` and the note naming #1018 are removed with the behaviour they described, `docs/tools/iivsearch.qmd` is rewritten, and the CHANGELOG carries the ranking change.

**Gates at `2912926e`:** `preflight OK — fmt check clippy rustdoc docs public-api` (baseline unchanged); `--workspace --features ferx-core/ci --lib` 5057 passed / 0 failed; same `--tests` 208 binaries, 6077 passed / 0 failed; `--features ci,markov,nn --tests` 185 binaries, 5720 passed / 0 failed. slow-tests re-dispatched on this head as run [34670034899](https://github.com/FeRx-NLME/ferx-core/actions/runs/34670034899).

**Not fixed here:** hoisting the mask out of the Gauss-Newton per-subject builders (round-1 finding: two small allocations per subject, needs signature changes in three functions), and the parser's missing overlapping-`block_omega` rejection (follow-up issue).

PR #1366, opened 2026-09-12
fix(parser): reject a second plain `DV ~ ...` line in [error_model] [closes #1022]

### LINKED ISSUE #1022: [error_model] silently keeps only the first of several plain `DV ~ ...` lines

roninsightrx:
## Why

A `[error_model]` block with more than one plain `DV ~ ...` line parsed, kept the
**first**, and silently discarded the rest — no error, no warning. From #1022,
measured at `maxiter = 0` on a 6-subject / 24-observation dataset:

| `[error_model]` | OFV |
|---|---|
| `DV ~ proportional(S_SMALL)` alone | −6.8397 |
| `DV ~ proportional(S_SMALL)` **then** `DV ~ additive(S_SMALL)` | −6.8397 |

Identical — the second line contributed nothing and said nothing. The shape a user
actually hits is a model edited in place: a replacement pasted above the line it
replaces, or an "old" line left behind. The file then reads as the new error model
and fits as the old one.

Same "accepted but not honoured" family as #1001 (which sigma binds); this one is
about which *line* binds. Every other duplicate in the same function was already
rejected — duplicate `CMT=N:` entries, a second `iiv_on_ruv`, mixing a plain line
with `CMT=` lines. The plain-line case was the one gap.

## What changed

One length check in `parse_error_model`, next to the existing duplicate guards:

```
[error_model] has more than one plain `DV ~ ...` line; a single-endpoint error
model takes exactly one (use `CMT=N:` prefixes for per-compartment models)
```

Wording matched to the neighbouring `iiv_on_ruv` message, as the issue suggested.

**Scope — the count applies to `singles` only.** Two other forms legitimately carry
several statements and are untouched:

- the covariate-selected `if/else` form returns from `parse_selected_error_model`
  *before* the line grammar runs;
- `CMT=N:` lines route to `per_cmt` before the count is reached.

Both bind every line. `iiv_on_ruv = ETA` is not a statement line and is partitioned
out upstream, so it may still accompany any single-endpoint model.

Checked the rest of the tree for shapes the guard could newly break: every
multi-`DV ~` block in `src/`, `tests/`, `examples/` and `docs/` is the `if/else`
selector; `ferx-tools`' ruvsearch screening-model generator emits exactly one line;
and `edit::apply::set_error_model` already refuses a block whose statement lines
number more than one (`existing.len() > 1`), replacing the first otherwise.

No diagnostic code: this stays `E_PARSE`, consistent with the three sibling
duplicate guards in the same function. `E_SIGMA_ORDER_MISMATCH` (#1001) earned its
own code because a consumer can offer the mechanical repair (permute one list);
here the repair is "delete a line", which needs no structure.

## Alternatives considered

**Bind the last line instead of the first.** Rejected — "the last one wins" is still
a silent rule, and the `[error_model]` block has no other last-wins semantics to be
consistent with.

**Deduplicate by content and only reject conflicting lines.** Rejected — the
identical-line case is the likeliest copy-paste arrival, and staying quiet about it
is exactly what lets the non-identical case go unnoticed.

## Cross-repo dependency

No public-API change; the baseline is unchanged.

## Breaking changes
- [ ] Public Rust API
- [ ] None

<details>

Breaking only for a model that currently has a dropped line — i.e. one that is not
doing what it says. Delete the line that was being discarded (the second and
later), or, if several endpoints really were meant, rewrite the block with `CMT=N:`
prefixes or as an `if/else` selector.

</details>

## Numerical validation
- [x] Not applicable — a parse-time rejection. No model that parsed before and
      still parses now produces a different number; the only behaviour change is
      that a previously-accepted file is refused. The objective figures motivating
      it are the issue's, reproduced in the Why table.

## Performance
- [x] No significant impact expected (one `Vec::len` comparison at parse time)

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 4398 passed)

Four rejection tests plus a control, in `src/parser/model_parser_tests.rs` next to
the #1001 family:

- `single_error_model_two_plain_lines_errs` — the issue's reproducer. Both sigmas
  are declared in argument order, so the #1001 order rule has nothing to say and
  the duplicate is the only defect the message can be about.
- `single_error_model_one_plain_line_parses` — **the control that makes the above a
  test of duplication rather than of the fixture.** Delete the second line and the
  same file parses; without it, a guard that rejected this whole shape (an
  `additive` second sigma, a two-sigma declaration) would read just as green.
- `single_error_model_duplicate_identical_lines_errs` — the same statement twice,
  the spelling where the dropped line changes nothing numerically.
- `single_error_model_ltbs_plus_plain_line_errs` — `log(DV) ~ additive(...)` lands
  in the same `singles` vector and carries the `LtbsFlags` the chosen single stamps
  on the model, so before this fix line order silently decided the *scale* of the
  whole fit.

Each of the four dies with the guard removed (the block then parses `Ok`). Scope
floor: `test_per_cmt_error_model_parses` and the `selected_error` tests are what go
red if the guard ever claims the multi-line forms — both families verified green
here rather than duplicated into a new test.

- [x] Not applicable (fix with no new API surface) — the rejected snippet is in the
      docs section below.

## Docs
- [x] `docs/` updated — new `## Exactly one plain `DV ~ ...` line` section in
      `docs/model-file/error-model.qmd`, placed before `## Sigma order`: the
      rejected snippet, the message, what was silently happening before, and
      pointers to the two forms that do bind every line.
- [x] `CHANGELOG.md` `[Unreleased]` entry added, under `Fixed`

`cargo run -p docs-lint -- --check` clean (the two internal anchors resolve; no new
R1 baseline entry).

## Checklist

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key → n/a, this removes an accepted
      spelling rather than adding one
- [ ] New example `.ferx` file → none
- [ ] New data file → none

### ferx-book
- [ ] New estimator, DSL feature, or fit option → n/a

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` in the tree carries a
      duplicate plain line (checked), so none changes behaviour.

## Reviewer hints

The load-bearing question is scope: does anything in the tree legitimately put two
plain `DV ~ ...` lines in one block? The two multi-line forms return before the
count (`parse_selected_error_model` at the top of `parse_error_model`; `per_cmt`
just above it), and the survey of every multi-`DV ~` block in the repo is in *What
changed*. The tests to read are the pair
`single_error_model_two_plain_lines_errs` / `single_error_model_one_plain_line_parses` —
the second is what stops the first from passing for the wrong reason.

## Open questions

Whether this deserves its own diagnostic code rather than `E_PARSE`. Argued above
for consistency with the sibling duplicate guards; happy to add
`E_ERROR_MODEL_DUPLICATE_LINE` if `ferxtranslate` / `ferx-r` would act on it.

https://claude.ai/code/session_01AKN72MVNZGiEuS5SxC9sqF


roninsightrx (commented):
Reviewed the parser control flow and the linked #1022 behavior; no findings. The new guard is scoped to parsed single-endpoint statements, while covariate-selected blocks return before this path and per-CMT statements remain in their separate collection. The added regression/control coverage exercises distinct valid duplicates, identical duplicates, LTBS mixing, and the one-line success case. I also ran `cargo test --lib single_error_model_`, `cargo test --lib test_per_cmt_error_model_parses`, and `cargo test --lib selected_error` locally: 26 tests passed. CI is green, including patch coverage, docs lint, clippy, formatting, public API, and release-semantics tests.

PR #1369, opened 2026-09-12
perf(estimation): reuse inner optimizer scratch buffers

hiddevandebeek:
## Why

The usual 2–8 ETA inner problems use dense BFGS. Each iteration rebuilt its gradient/direction adapters, step and curvature vectors, identity matrix, and line-search trial point. The Nelder–Mead fallback likewise rebuilt ordering and geometry buffers each iteration. These allocator calls sit inside every subject's EBE search for every outer-objective evaluation.

## What changed

- Allocate dense-BFGS vector/matrix adapters and the line-search trial point once per inner solve and overwrite them in place.
- Allocate the invariant identity matrix once per dense-BFGS solve.
- Reuse the L-BFGS line-search trial point.
- Reuse Nelder–Mead ordering, centroid, reflection, expansion, contraction, and best-point buffers; accepted points are swapped into the simplex.
- Add test-only copies of the previous algorithms and compare convergence, final ETA bits, final objective bits, and evaluation counts at dimensions 2, 4, and 8.
- Add an opt-in counting-allocator microbenchmark for allocation calls and bytes.

The BFGS inverse-Hessian expression and its operation order are unchanged.

## Alternatives considered

Expanding the inverse-Hessian update into an algebraically equivalent rank-two formula could remove more matrix temporaries, but it would reassociate floating-point operations and could change the inner optimization path. This PR deliberately keeps that expression intact.

## Cross-repo dependency

## Breaking changes
- [x] None

<details>

Not applicable.

</details>

## Numerical validation

- [x] Reference values: test oracles require exact bitwise equality with the pre-change algorithms at ETA dimensions 2, 4, and 8, including convergence and evaluation counts.

The full unit-test binary could not be code-generated in the current constrained container: rustc was SIGKILLed at the memory ceiling in both debug and `ci-fast` profiles. `RAYON_NUM_THREADS=1 cargo check -j 1 --profile ci-fast --tests` passes on the rebased branch. A standalone nalgebra check also confirmed scratch-backed `gemv` matches the prior `-&H * &g` expression bit-for-bit for 30,000 randomized matrix-vector products.

## Performance
- [ ] Faster — benchmark: pending execution of the included allocation microbenchmark on a normal-memory runner

Run the allocation benchmark alone with:

```bash
rtk env RUSTFLAGS='--cfg profiling_allocations' RAYON_NUM_THREADS=1 cargo test -j 1 --profile ci-fast --lib dense_bfgs_scratch_allocation_bench -- --ignored --nocapture
```

## Tests
- [x] Regression oracles added for exact dense-BFGS and Nelder–Mead path equivalence

All test targets type-check. Execution remains pending because of the container memory limit described above.

<details>

Not applicable.

</details>

## Docs
- [x] `CHANGELOG.md` N/A: internal allocation optimization with no behavior or API change
- [x] No user-visible change

## Checklist
- [ ] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), public-API baseline
- [x] Functions on the `Dual2` sensitivity path stay differentiable (no sensitivity or prediction code changed)
- [x] `Cargo.toml` version bump not needed; no breaking change

`cargo fmt --check` and `cargo check -j 1 --profile ci-fast --tests` are green with `RAYON_NUM_THREADS=1`.

### ferx-site
- [x] No DSL syntax, examples, or data changed

### ferx-book
- [x] No estimator, DSL feature, or fit option changed

## Reviewer hints

The sensitive point is floating-point identity: review the use of scratch-backed `gemv`, the unchanged inverse-Hessian matrix expression, and the buffer swaps in Nelder–Mead. The test-only legacy implementations pin final bits and evaluation order.

## Open questions

- Run the allocation benchmark and the focused unit tests on a normal-memory runner before marking ready for review.

### DIFF
--- a/src/estimation/inner_optimizer.rs
+++ b/src/estimation/inner_optimizer.rs
@@ -2800,6 +2800,7 @@ fn lbfgs_core(
     let mut rho_hist: Vec<f64> = Vec::new();
     let mut g = grad(x);
     let mut f_cur = obj(x);
+    let mut line_search_point = vec![0.0; n];
     // Objective-stall convergence (see [`objective_stalled`] / [`INNER_FTOL_REL`]): for ODE
     // objectives whose gradient norm is floored above `tol` by solver noise, a search that
     // reached the mode declares convergence rather than spinning to `max_iter`. Gated on
@@ -2835,7 +2836,8 @@ fn lbfgs_core(
             };
         }
 
-        let (alpha, f_new) = backtracking_line_search(obj, x, &d, &g, n, f_cur);
+        let (alpha, f_new) =
+            backtracking_line_search(obj, x, &d, &g, f_cur, &mut line_search_point);
         // No sufficient-decrease step found: report non-convergence so the caller takes the
         // argmin Nelder–Mead fallback rather than accepting a non-stationary η̂.
         if alpha == 0.0 {
@@ -2895,6 +2897,15 @@ fn dense_bfgs_core(
 ) -> bool {
     let mut h_inv = init_h_inv(n, precond);
     let mut g = grad(x);
+    let mut g_vec = DVector::zeros(n);
+    let mut d_vec = DVector::zeros(n);
+    let mut d = vec![0.0; n];
+    let mut s = vec![0.0; n];
+    let mut y = vec![0.0; n];
+    let mut s_vec = DVector::zeros(n);
+    let mut y_vec = DVector::zeros(n);
+    let eye = DMatrix::identity(n, n);
+    let mut line_search_point = vec![0.0; n];
     // Track the objective at the current iterate so the line search never has to
     // recompute `obj(x)` (one prediction walk per inner step on the hot path).
     let mut f_cur = obj(x);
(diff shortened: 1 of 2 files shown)


hiddevandebeek:
I found one issue:

**[P2] Measure the optimized path as `current` in the allocation benchmark** (`src/estimation/inner_optimizer_tests.rs:1197`)

`run_dense_scratch_fixture(n, true)` selects `legacy_dense_bfgs`, but this result is stored in `current`; the following `false` call selects the scratch-reusing implementation but is stored in `legacy`. As a result, the benchmark prints the two measurements under the wrong labels and asserts that the allocation-heavy legacy implementation uses fewer allocations/bytes. When the optimization works, the documented opt-in benchmark fails. Swap the boolean arguments (or the bindings) so `current` runs with `false` and `legacy` with `true`.

I otherwise found the scratch-buffer swaps and production call-site changes consistent with the previous algorithm. The PR's release-semantics, check, clippy, format, rustdoc, docs, and public-API CI jobs are green at the time of review; the coverage jobs are still running.

hiddevandebeek:
Fixed in `7a7e9e5`: the optimized path now supplies the `current` measurement and the pre-change path supplies `legacy`.

Validation: `cargo fmt --check` and `git diff --check` pass. I also started the exact opt-in allocator benchmark with `-j 1`, `RAYON_NUM_THREADS=1`, and one test thread; dependency compilation completed, but the constrained machine did not finish final test-binary code generation, so I stopped it rather than increasing resource use.

PR #1370, opened 2026-09-12
perf(estimation): reuse AGQ objective work

hiddevandebeek:
## Why

AGQ/Laplace outer evaluations were doing avoidable work in two hot paths. Objective-only calls allocated one coordinate vector at every tensor-grid node, and gradient-capable NLopt callbacks evaluated the same subject anchor and quadrature grid once for the objective and again for the matching analytic gradient.

## What changed

- Reuse one subject-local coordinate buffer during objective-only quadrature sweeps.
- Retain the anchor Hessian, quadrature coordinates, softmax weights, stacked mode metadata, and exact-anchor base sensitivity from an objective evaluation when the same NLopt callback requests a gradient.
- Consume that retained work in the AGQ gradient while preserving the existing per-subject analytic-to-FD fallback.
- Keep derivative-free callbacks, scheduled reconverged gradients, unsupported analytic models, covariance scoring, and the optimizer choice unchanged.
- Add a regression test requiring cached and uncached objectives/gradients to be bit-identical.

## Alternatives considered

- A general cross-callback cache was rejected because optimizer probes can change parameters and EBEs, making invalidation subtle.
- Cloning the retained exact-anchor sensitivity was rejected because it would add a large allocation in the path this change is intended to reduce. The prepared subject state is consumed instead.
- Changing optimizers or quadrature rules is intentionally outside scope.

## Cross-repo dependency

## Breaking changes
- [x] None

<details>

Not applicable.

</details>

## Numerical validation

- [x] Compared against: prior ferx commit `8372248c`
- [x] Reference values: the new unit test checks objective and every packed gradient coordinate with `to_bits()` against the uncached path.

```
# before / reference
uncached AGQ population objective and analytic gradient

# after
bit-identical cached objective and gradient (focused test pending local resource availability)
```

## Performance
- [x] Faster — benchmark: objective-only node-coordinate allocations `Q` → `1`; gradient callbacks perform anchor/grid construction `2` → `1` per subject (wall-time benchmark pending)
- [ ] Slower — justified because: not applicable

## Tests
- [ ] No new tests needed (why: not applicable)

One-core `cargo check -j 1 --lib` passes. Broader local test compilation is pending because other worktrees currently hold the shared Cargo target lock; this PR remains a draft until it runs.

<details>

Not applicable.

</details>

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added (user-facing change), or N/A (internal/refactor/CI): performance entry added for #1370
- [ ] No user-visible change

## Checklist
- [ ] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy (`--all-targets`), public-API baseline. A green `cargo test` is not a green build: the feature sets are not nested (#1133, #1157)
- [x] Functions on the `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay differentiable (if touching gradient-reachable code): no sensitivity or generic PK formulas changed

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → not applicable; no syntax or option change
- [x] New example `.ferx` file added to `examples/` → not applicable; no example added
- [x] New data file added to `data/` → not applicable; no data added

### ferx-book
- [x] New estimator, DSL feature, or fit option → not applicable; existing estimators only

## Reviewer hints

The important invariant is callback-local identity: retained work is created only when a gradient is requested at the same parameter/EBE evaluation, and it is consumed exactly once. Please focus on `agq_subject_evaluate`, ownership of `PopulationEvaluation`, and the gates in `optimize_nlopt_once`. The exact-anchor `base_jet` is deliberately moved, not cloned.

## Open questions

None. The draft status is only for completing one-core tests and adding a wall-time measurement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,11 @@ section of the SDLC for the versioning policy).
 
 ### Performance
 
+- **Laplace and AGQ gradient callbacks now reuse the anchor and quadrature grid computed for
+  their matching objective evaluation**, instead of rebuilding both per subject; objective-only
+  quadrature sweeps also reuse one node-coordinate buffer rather than allocating at every node.
+  Optimizer choice and numerical results are unchanged (#1370).
+
 - **The post-fit per-subject diagnostics pass (IPRED / PRED / IWRES / CWRES,
   per-subject OFV) and the post-fit analytic-sensitivity sweep now run in
   parallel over subjects** on the pool the fit already uses, instead of one
--- a/src/estimation/agq.rs
+++ b/src/estimation/agq.rs
@@ -96,11 +96,6 @@ pub const MAX_AGQ_NODES: usize = 21;
 /// hours; past that a fit is not slow, it is wrong to have started.
 pub const MAX_AGQ_GRID: usize = 100_000;
 
-/// Maximum estimated memory retained between the objective and gradient halves of one
-/// optimizer callback. Larger evaluations still use the same objective and gradient, but
-/// recompute the subject-local grid instead of keeping every subject's grid alive at once.
-const MAX_PREPARED_GRID_BYTES: usize = 128 * 1024 * 1024;
-
 /// The sentinel a non-finite likelihood collapses to, matching `individual_nll`'s own
 /// convention so a diverged node sorts as "impossible" rather than poisoning the sum.
 const NLL_SENTINEL: f64 = 1e20;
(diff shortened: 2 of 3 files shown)


hiddevandebeek:
Reviewed commit `b79cb5600232f5b5bf482149c35c78185a2b3d09`.

**[P2] Bound the population-wide retained quadrature cache** — [src/estimation/agq.rs:978–981](https://github.com/FeRx-NLME/ferx-core/blob/b79cb5600232f5b5bf482149c35c78185a2b3d09/src/estimation/agq.rs#L978-L981)

The collect retains every subject’s `PreparedGrid` until the entire objective has finished, and only then starts consuming grids for gradients. Previously each subject’s coordinate grid was dropped before that worker moved to its next subject. Thus gradient callbacks now require O(subjects × nodes^dimension × dimension) grid memory instead of O(workers × nodes^dimension × dimension), even with one worker. For example, 1,000 subjects with five random effects and `n_agq = 9` pass the 100,000-node cap (59,049 nodes/subject), but the retained coordinates, Vec headers and softmax weights alone require 4,251,528,000 bytes (3.96 GiB), before allocator overhead, Hessians and retained sensitivities. This can make previously memory-bounded SLSQP/L-BFGS/MMA fits swap or fail allocation. Please bound total retained bytes and use the existing uncached path above the budget, or fuse/batch objective and score work so completed grids can be released. The memory figure is a storage calculation from the code, not a measured large-fit benchmark.

The callback-local parameter/mode identity, reconvergence scheduling, coordinate-retention flags at other grid callers, and per-subject fallback paths otherwise look consistent.

Validation: inspected all changed production paths and their callers. Attempted the existing retained-work regression plus two temporary exact-anchor variants (`n_agq = 3` and `1`) in an isolated review worktree, with CPU affinity restricted to CPU 0, one Cargo build job, one Rayon thread and one test thread. The local build did not finish during the review and was interrupted; no local test pass is claimed. Full-fit numerical and wall-time validation remain unverified.

hiddevandebeek:
Addressed the population-wide cache finding in `2dff72ea`.

The objective now estimates retained work before entering the parallel subject sweep and enables callback reuse only when the complete population cache fits within a 128 MiB budget. The estimate includes node-coordinate allocations and Vec metadata, softmax weights, retained matrices/modes, and the exact-anchor sensitivity jet. Evaluations above the budget keep the same objective and gradient behavior but use the existing uncached per-subject gradient path, so they no longer keep every subject's grid alive simultaneously.

I also corrected the original bit-identity fixture to mirror the production callback: both cached and uncached halves now use parameters unpacked from the optimizer vector. The previous fixture compared cached work built from natural-space defaults against an uncached gradient after the log/exp pack round trip, which shifted one coordinate by 1 ULP.

Validation, restricted to CPU 0 with one Cargo job, one Rayon thread, and one test thread:

- `prepared_grid_budget_rejects_population_wide_growth`: pass
- `retained_objective_work_reproduces_objective_and_gradient`: pass, including exact `to_bits()` equality
- `cargo fmt --check`: pass
- `cargo clippy --lib -j 1`: pass with the repository's existing warnings

`cargo clippy --lib -j 1 -- -D warnings` remains red on 295 pre-existing warnings outside this patch; the first is the duplicate `#[allow(clippy::too_many_arguments)]` in `src/api/simulate.rs`.

hiddevandebeek:
Replaced the 128 MiB cache-budget follow-up with the fused design in `5a302e91`.

There is no byte threshold in the final implementation. On a gradient-requesting AGQ/Laplace callback, each Rayon subject task now:

1. evaluates the subject objective and builds its quadrature grid;
2. computes that subject's packed score from the same grid; and
3. drops the grid and retained exact-anchor sensitivity before the worker takes another subject.

Only `(subject_nll, packed_score)` is collected for deterministic serial reduction in subject order. Peak quadrature storage therefore scales with active workers, as it did before this PR, rather than with population size. Scheduled reconverged gradients, unsupported analytic models, derivative-free callbacks, and per-subject numerical fallback retain their existing paths.

The earlier budget commit remains in history but its constant, estimator, gate, and regression were all removed by this commit.

One-core validation (`taskset -c 0`, one Cargo job, one Rayon thread, one test thread):

- fused objective/gradient regression: pass for both Gauss–Newton and Exact anchors, including exact objective and packed-gradient `to_bits()` equality with the uncached paths
- `cargo check --lib -j 1`: pass
- `cargo fmt --check`: pass
- `cargo clippy --lib -j 1`: pass with the repository's existing warning baseline

PR #1371, opened 2026-09-12
perf(ode): reuse RK45 stages for auto-switch detection

hiddevandebeek:
## Why

`ode_method = auto` currently detects stiffness with a finite-difference Jacobian at segment entry and every 25 accepted steps. That costs extra RHS evaluations, uses a time-unit-dependent rate threshold, and loses its detection history at dose and covariate event boundaries. Short event intervals can therefore delay or prevent a mid-segment switch even when accepted RK45 stages already show stability-limited integration.

rxode2's AutoSwitch work showed the useful model: derive a dimensionless stiffness signal from Dormand-Prince stages and retain its verdict counters across event intervals.

## What changed

- Compute Hairer's dimensionless RK45 `h·rho(J)` estimate from the accepted step's existing stage values, with no additional RHS calls.
- Ratify a stiff verdict after 15 consecutive signals and clear partial evidence after 6 non-stiff signals.
- Carry detector and periodic-probe history across event boundaries in prediction and analytic-sensitivity subject walks.
- Preserve accepted state, time, and adaptive step size when changing stepper; rebuild stages and Jacobians at event discontinuities.
- Keep the periodic finite-difference Jacobian probe every 25 accepted steps as a backstop.
- Roll detector history back when a guarded stiff attempt is discarded.

## Alternatives considered

Persisting the complete RK45 or Rosenbrock workspace across event boundaries would reuse more data, but doses and forcing changes invalidate cached stages and Jacobians. Persisting only accepted-step verdict history retains valid information without carrying stale numerical state. Replacing the Jacobian probe entirely was also rejected because it remains useful for methods without a stage-based detector and as a backstop.

## Cross-repo dependency
| ferx (R) | N/A | not needed | — |

No public API or R glue changes are required.

## Breaking changes
- [x] None

<details>

None.

</details>

## Numerical validation
- [x] Compared against: rxode2 AutoSwitch/Dormand-Prince detector and prior ferx auto-switch behavior
- [x] Reference values: the dimensionless detector gives the same value for equivalent decay systems expressed in hours and days; existing scalar/dual parity and mid-segment switching tests pass.

```
# focused one-core validation
38 passed; 0 failed; 3 ignored measurement harnesses
```

## Performance
- [x] No significant impact expected on benign/non-stiff systems; the per-step signal uses values RK45 already computed and adds no RHS calls

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib stiffness -j 1 -- --test-threads=1` passes)

- [x] Not applicable (the existing `ode_method = auto` and `ode_auto_switch = true` interface is unchanged)

<details>

```toml
[fit_options]
ode_method = auto
ode_auto_switch = true
```

```
# AutoSwitch can now accumulate its RK45 stiffness verdict across event intervals.
```

</details>

## Docs
- [x] New pages linked in `docs/_quarto.yml` (N/A: existing pages updated)
- [x] `CHANGELOG.md` `[Unreleased]` entry added (#1371)
- [ ] No user-visible change

## Checklist
- [ ] `tools/preflight.sh` green — focused formatting, compilation, and one-core solver tests have passed; full preflight remains for CI
- [x] Functions on the `Dual2` sensitivity path (`sens/`, `pk/` `*_g<T: PkNum>`) stay differentiable; switching reads primal values consistently with the existing stiffness probe
- [x] `Cargo.toml` version bumped if breaking (N/A: non-breaking)

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → N/A; no syntax or new key
- [x] New example `.ferx` file added to `examples/` → N/A; no new example
- [x] New data file added to `data/` → N/A; no new data

### ferx-book
- [x] New estimator, DSL feature, or fit option → N/A; existing AutoSwitch behavior only

- [x] No example execution step needed (no interface, example, or data changes)

## Reviewer hints

Focus on `OdeAutoSwitchState` lifetime and rollback in `src/ode/solver.rs`. The subtle constraint is that verdict counters may survive an event, while RK stages and Rosenbrock Jacobians must not. The prediction and dual-sensitivity walks deliberately share the same policy so finite-difference and analytic-gradient evaluations choose methods consistently.

## Open questions

Follow-up work could add retry backoff after a guarded stiff escalation fails. It is intentionally outside this focused change.


hiddevandebeek:
Review of commit `16d7942d49a49dcef915cef4324643f7d16a9000`.

Found one issue to fix:

- **[P2] Preserve a ratified stage verdict when the interval ends on the switching step.** In [`src/ode/solver.rs:1294–1316`](https://github.com/FeRx-NLME/ferx-core/blob/16d7942d49a49dcef915cef4324643f7d16a9000/src/ode/solver.rs#L1294-L1316), a fifteenth stiff vote creates the new stepper and immediately clears the votes, even when the accepted step has already reached `tf`. The loop then exits without taking any stiff step. The next event interval resolves its starting method from the Jacobian probe alone (`integrate_resolved_g_inner`), so a rate below the fixed threshold starts RK45 again with the ratified history erased. This loses exactly the cross-interval decision the new dimensionless detector is intended to retain, and reports a switched/stiff segment even though all its steps were explicit. Preserve a pending verdict across the boundary, or defer consuming/resetting the votes until the replacement can actually integrate. Please cover this through successive solver calls; the new `rk45_stiffness_votes_survive_event_intervals` test only calls the counter helper and cannot catch driver-level resets.

The local solver-level regression uses `u' = -10u`, initial state `1e-12`, `abstol = 1e-6`, `min_dt = 0.33`, and fifteen consecutive intervals of length `0.33`, sharing one `OdeAutoSwitchState`. Each step meets the error tolerance (no forced-accept clamps), with a stage indicator of approximately `3.3`; the Jacobian rate is `10`, below the entry threshold of `30`. The regression checks that the fifteenth vote is not discarded at the final boundary.

Validation (one CPU via affinity, one Cargo job, one test thread):

- `cargo test --lib stiffness -j 1 -- --test-threads=1`: **38 passed, 0 failed, 3 ignored**.
- Local regression `review_1371_terminal_stage_switch_retains_verdict`: **fails as expected** after confirming 15 accepted steps, one reported switched segment, and zero minimum-step clamps. Actual final history: `OdeAutoSwitchState { accepted_since_probe: 15, rk_stiff: 0, rk_nonstiff: 0 }`.
- Formatting passed on the original PR checkout. The regression was added only to an isolated review worktree; no PR source changes were pushed.

hiddevandebeek:
Fixed the terminal-boundary verdict loss in 6a5854aa. Accepted RK45 stages are now observed even when they land exactly at the segment end, but stepper construction and switch accounting are deferred until a subsequent nonzero interval can actually integrate. A new successive-solver-call regression verifies that the fifteenth terminal vote remains pending, does not report a phantom switched segment, and starts the sixteenth interval stiff before clearing the verdict. Validation: cargo fmt --check; targeted regression 1 passed; cargo test --lib stiffness -j 1 -- --test-threads=1: 38 passed, 0 failed, 3 ignored.

PR #1373, opened 2026-09-12
perf(sens,estimation): drop wasted work and reallocation in the FOCE/FOCEI/Laplace inner sensitivity path

hiddevandebeek:
## Why

Investigation into whether the FOCE/FOCEI/Laplace **standard sensitivity path** could be sped up by coding-only changes (no optimizer/algorithm switch) turned up three concrete, mechanical wastes in the closed-form inner EBE gradient — the hot path `find_ebe`'s BFGS loop calls on every inner iteration:

1. **`subject_eta_grad_impl`'s closed-form log-normal fallback** (used when a model's compiled `[individual_parameters]` program doesn't cover every required PK slot) called `lognormal_param_derivatives`, which builds the *entire* `θ`/second-order `ParamDerivs` block (`dp_dtheta`, `d2p_deta2`, `d2p_detadtheta`, plus an FD `tv_theta_jacobian` pass) — then the caller keeps only `.dp_deta`. This directly contradicts `subject_eta_grad`'s own doc comment: *"never the second-order `∂²f/∂η²` or the θ-block... the hot path: millions of calls."* The compiled-program branch right above it already has the correct light/full split (`param_eta_derivatives_from_prog` vs `param_derivatives_from_prog`) — this fallback never got the same treatment.
2. **`run_obs_grad`** (the per-observation `∂f/∂η` walk backing that same closed-form path) allocated a fresh `Vec<f64>` per observation, every call.
3. **`analytic_eta_nll_gradient_with_schedule`** recomputed `model.error_spec.obs_keys(subject)` every inner BFGS step, even though it's η-independent and subject-static — for a `Selected` (covariate-selector) error spec this allocates a fresh `Vec<usize>` every time. `mult` was already hoisted out of this exact closure for the identical reason (#486 review); `err_keys` never got it.

## What changed

1. Added `lognormal_eta_derivatives_only` — the η-only counterpart of `lognormal_param_derivatives`, computing just `∂p_i/∂η_k = pk_i·sel_ik`. `subject_eta_grad_impl`'s fallback now calls this instead.
2. Threaded a per-subject `ObsGrad` scratch buffer from `find_ebe` through `analytic_eta_nll_gradient_with_schedule` → `subject_eta_grad_with_schedule` → `subject_eta_grad_impl` → `run_obs_grad`, mirroring the existing `pk_scratch_cell`/`EventSchedule`/`mult` hoists. `run_obs_grad` now overwrites each `ObsGrad`'s `df_deta` in place (reusing its `Vec<f64>` capacity) instead of allocating fresh, since `n_obs`/`n_eta` are subject-static across a `find_ebe` call. **Scoped to the closed-form path only** — the TV-cov (`run_obs_grad_tvcov`), ODE, and algebraic-model branches still allocate fresh. Every public entry point (`subject_eta_grad`, `subject_eta_grad_tvcov`, `ode_subject_eta_grad`) is untouched; the new parameter only threads through the `pub(crate)`/private internal chain.
3. Hoisted `err_keys` alongside `mult` in `find_ebe` and the one-off `analytic_eta_nll_gradient` wrapper, threaded through to `dense_residual_inner_gradient` (the `block_sigma` branch). While doing this, found `dense_residual_inner_gradient` had the *same* bug for `mult` itself — the caller already had it computed but never passed it down, so this branch recomputed `ruv_obs_mult` independently every call. Fixed the same way.

None of this changes the objective, gradient formula, or optimizer trajectory — every change is "compute/allocate less for the identical result."

## Alternatives considered

Full-scope item 2 (threading scratch reuse through the TV-cov/ODE/algebraic branches too) was considered and explicitly scoped out for this PR: `ObsGrad` construction happens independently in five places across `provider.rs`, `ode_provider.rs`, and `algebraic.rs`, several of which back **public API** functions (`ode_subject_eta_grad`, `subject_eta_grad_tvcov` are in `api/ferx-core-public-api.txt`) that must keep allocating fresh `Vec<ObsGrad>`. Extending reuse to those paths is a reasonable follow-up but roughly doubles this PR's surface for wins on less commonly benchmarked model families.

## Cross-repo dependency

| ferx (R) | — | not needed: no API/schema change | — |

## Breaking changes

- [x] None — no public API, `.ferx` format, or `FitResult`/sdtab field change. `subject_eta_grad_with_schedule`, `subject_eta_grad_impl`, `analytic_eta_nll_gradient_with_schedule`, `dense_residual_inner_gradient`, `node_nll_gradient`/`node_jet` all gained parameters, but every one is `pub(crate)` or private.

## Numerical validation

- [x] Not applicable in the usual (estimator-comparison) sense — this changes neither the FOCE/FOCEI/Laplace objective, gradient, nor Hessian, only how the identical values are computed/allocated internally.
- [x] Full `cargo test --lib` (no features): **4453 passed, 0 failed, 23 ignored.**
- [x] Two new regression tests specifically targeting the two riskiest changes:
  - `lognormal_eta_derivatives_only_matches_full_dp_deta` — pins the light η-only derivative bit-for-bit against the full builder's η-block.
  - `subject_eta_grad_with_schedule_hint_reuse_is_bit_identical` — calls the closed-form provider once fresh, then again with a deliberately corrupted, wrong-length, oversized hint, and asserts the two results are bit-identical (catches a reuse-in-place bug that forgets to re-zero `df_deta`, forgets to overwrite a slot, or mishandles a hint longer than the current observation count).
- [x] `cargo clippy --no-default-features --features ci,markov,nn --all-targets` (the exact `tools/preflight.sh` invocation): exit 0, and none of the resulting warnings touch any changed file/line.
- [x] `cargo fmt --check`: clean.

## Performance

- [x] Faster — but **no wall-clock figure is claimed in this PR**. These are allocation/redundant-computation removals with an unchanged provider *call count* (verified via the full test suite, not a call-count profile) — the honest claim is "less work per call," not a measured speedup. A `FERX_PROFILE`/`bench.sh` run (`plans/laplace-sensitivity-speed/bench.sh`) to quantify the actual wall-clock/allocation delta is a natural follow-up; it wasn't run here because of sustained heavy memory contention on the shared dev box during this session.

## Tests

- [x] Regression test added (fails without the fix): both new tests above fail if the corresponding fix is reverted (the first by construction — it directly compares old vs. new; the second by reverting the scratch-reuse logic back to always allocating fresh, which would make the corrupted-hint case irrelevant rather than fail loudly, so the meaningful regression it guards is the *reuse* logic overwriting incorrectly, e.g. a missing re-zero).

## Docs

- [x] `CHANGELOG.md` `[Unreleased]` entry added under `### Performance`.
- [x] No user-visible change — no docs page update needed.

## Checklist

- [x] `tools/preflight.sh fmt` and the clippy group's exact command both run manually and green (see Numerical validation above); the full `tools/preflight.sh` wasn't run end-to-end due to the shared box's memory constraints during this session — `check`/`clippy`/`fmt` were each run with the same flags preflight uses.
- [x] Functions on the `Dual2` sensitivity path (`sens/`) stay differentiable — no dual-number arithmetic changed, only allocation/dispatch around it.
- [ ] `Cargo.toml` version bumped — not breaking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,13 @@ section of the SDLC for the versioning policy).
 
 - **Closed-form steady-state bolus models with estimated lag times now use analytical event sensitivities**, avoiding finite-difference fallback for the supported event-walk route (#1311).
 
+- **The closed-form inner EBE gradient no longer computes or allocates work it discards.** Three redundant-work sites in the FOCE/FOCEI/Laplace analytic sensitivity path, none changing the objective, gradient formula or optimizer trajectory:
+  - The closed-form log-normal fallback (used when a model's compiled `[individual_parameters]` program doesn't cover every required PK slot) called the full `θ`/second-order derivative builder and then read only its η-block; it now calls a light η-only counterpart that skips the θ-axis and second-order work (and the FD `tv_theta_jacobian` pass those needed) entirely.
+  - The per-observation `∂f/∂η` walk (`run_obs_grad`, run on every inner BFGS step) allocated a fresh `Vec<f64>` per observation on every call; it now reuses a per-subject scratch buffer threaded from the inner loop across BFGS iterations, since neither the observation count nor `n_eta` changes within one EBE solve.
+  - The per-observation residual endpoint keys (`ErrorSpec::obs_keys`, non-trivial only for a `Selected`/covariate-selector error spec) were recomputed every inner BFGS step; they are now hoisted once per subject, the same treatment the custom-magnitude multiplier already got. The dense-residual (`block_sigma`) branch had the equivalent problem for that multiplier itself — it never received the caller's already-computed value — and is fixed the same way.
+
+  Verified bit-for-bit unaffected: the full `cargo test --lib` suite (4453 tests) is green, and two new regression tests pin the light derivative against the full one and the scratch-buffer reuse against a corrupted hint. No wall-clock figure is claimed here — these are allocation/redundant-computation removals with an unchanged provider call count, not a change to what gets evaluated (#1373).
+
 ### Added
 - **Additive (`+`) covariate effects in `[covariate_model]` (#1313).** A trailing operator
   token makes a relation a term added to the parameter instead of a factor on it —
--- a/src/estimation/agq.rs
+++ b/src/estimation/agq.rs
@@ -1400,9 +1400,21 @@ fn grid_response_correction(
     // `phi_grid` re-sweep for every coordinate (all-or-nothing, matching the anchor).
     // Hoisted once for the whole sweep — see `node_nll_gradient`.
     let mult = model.ruv_obs_mult(subject, &params.theta);
+    let err_keys = model.error_spec.obs_keys(subject);
     let node_grads: Option<Vec<Vec<f64>>> = bs
         .iter()
-        .map(|b| node_nll_gradient(model, subject, params, stack, b, schedule, mult.as_deref()))
+        .map(|b| {
+            node_nll_gradient(
+                model,
+                subject,
+                params,
+                stack,
+                b,
+                schedule,
+                mult.as_deref(),
+                err_keys.as_ref(),
+            )
+        })
         .collect();
 
     // Assemble `dH/dx` (or `dH̃/dx`) in closed form instead of rebuilding the anchor at
@@ -1868,6 +1880,7 @@ fn node_nll_gradient(
     b: &[f64],
     schedule: Option<&pk::event_driven::EventSchedule>,
     mult: Option<&[Vec<f64>]>,
+    err_keys: &[usize],
 ) -> Option<Vec<f64>> {
     if stack.is_iov() {
         let iov = params.omega_iov.as_ref()?;
@@ -1896,6 +1909,8 @@ fn node_nll_gradient(
             &params.residual_correlations,
             schedule,
             mult,
+            err_keys,
+            &mut Vec::new(),
         )
     }
 }
@@ -2469,6 +2484,7 @@ mod tests {
         )
         .expect("base mode response");
         let mult = model.ruv_obs_mult(&subject, &params.theta);
+        let err_keys = model.error_spec.obs_keys(&subject);
         let node_grads: Vec<Vec<f64>> = bs
             .iter()
             .map(|b| {
@@ -2480,6 +2496,7 @@ mod tests {
                     b,
                     schedule.as_ref(),
                     mult.as_deref(),
+                    err_keys.as_ref(),
                 )
                 .expect("node gradient")
             })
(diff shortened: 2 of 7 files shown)


hiddevandebeek:
Reviewed commit `08e435157fd2568db55cdf18eecdea1c593f8f16`.

No blocking correctness findings. The light log-normal helper preserves the original eta derivative formula and indexing. The reused observation slots overwrite both predictions and derivatives before applying initial conditions, scaling, log transformation, and FREM overrides. I also traced all updated gradient callers: the cached endpoint keys and residual multipliers use the same subject/theta inputs as before, including the dense-residual and AGQ paths.

One non-blocking test improvement: in `src/sens/provider_tests.rs`, `subject_eta_grad_with_schedule_hint_reuse_is_bit_identical` changes every derivative vector to length `n_eta + 3`. Consequently, it exercises only the resize branch of `reset_zeroed`, not the `v.len() == n` branch used on ordinary BFGS iterations. Add a second poisoned hint that retains the original lengths, ideally evaluated at a different eta and compared with a fresh evaluation there. This would catch stale accumulation in the actual steady-state reuse path. An allocation-count check would separately protect the intended performance benefit; numerical equality alone still passes if reuse is removed.

Validation: inspected all seven changed files and the surrounding dispatch/caller logic. Local targeted-test compilation did not complete during this review, so I have no independent local test result. GitHub Check, Clippy, Format, Rustdoc, and Public API baseline checks were green when inspected; test/coverage jobs were still pending. No benchmark was run.

hiddevandebeek:
Benchmark results for `9603c807` versus its pre-optimization parent `5c9fb8ec`:

- Fixture: `examples/warfarin.ferx` + `data/warfarin.csv` (closed-form FOCE)
- Build: `cargo build --profile ci-fast -p ferx-cli`, independent target directories; compilation excluded
- Toolchain: `rustc 1.100.0-nightly (4aa1fbcf4 2026-09-08)`
- Protocol: one warm-up per binary, then 40 A/B pairs with order alternated on every pair; the fit reported 3 threads
- Instrumentation: `FERX_TIME_GRADIENTS=1`; external monotonic wall clock around the CLI process

| Metric | Baseline median | PR median | Median improvement | Paired mean improvement (bootstrap 95% CI) |
|---|---:|---:|---:|---:|
| Analytic BFGS gradient total | 180.88 ms | 163.74 ms | 9.5% | 11.7% (6.8%–16.4%) |
| Analytic gradient per call | 4.085 µs | 3.695 µs | 9.5% | 11.7% (6.8%–16.4%) |
| Whole CLI wall time | 0.505 s | 0.478 s | 5.3% | 3.7% (0.1%–7.2%) |

All 80 measured fits followed the same trajectory at the recorded level: 44,320 analytic BFGS-gradient calls, 67 iterations, and final OFV `-280.3640`. The kernel figure is the cleanest estimate of this PR's effect. Whole-fit timing includes parsing, objective evaluations, covariance, output, and scheduler noise, so the practical claim for this fixture is **about 10–12% faster analytic inner-gradient evaluation and about 4–5% faster end-to-end**.

This fixture exercises the scratch-buffer reuse, which is the broad part of the change. It does not exercise a `Selected` residual error or force the log-normal fallback, so those two narrower savings are not included in this number.

PR #1374, opened 2026-09-12
perf(estimation): reuse classical estimator hot-path buffers

hiddevandebeek:
## Why
FOCE-family inner objectives and AGQ grids repeatedly allocate subject-sized prediction/prior buffers and rebuild invariant metadata. IOV AGQ additionally copied every occasion vector and recomputed cached covariance inverses at every node.

## What changed
- Reuse prediction and prior-matrix scratch across EBE objective/gradient evaluations.
- Reuse static analytical prediction output capacity.
- Reuse AGQ node NLL scratch and cache invariant Gauss-Hermite rules.
- Borrow IOV kappa slices and use cached Omega inverses/log-determinants.
- Remove allocation from indexed dose-attribute and bioavailability name checks.

## Alternatives considered
A broader dense-BFGS rewrite was benchmarked and discarded because it was neutral-to-negative. Current main independently contains the validated optimizer scratch implementation; this PR is additive to it.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation
- [x] Compared against current `main` with the same `ci-fast` profile and one Rayon thread.
- [x] All 40 runs per arm reproduced the same displayed OFV and outer iteration count: FOCE −280.3640 / 67; FOCEI −234.2689 / 94; Laplace −285.9702 / 37; AGQ −285.9739 / 37.
- [x] A full Warfarin FOCE output comparison was identical for parameters and sdtab; only timing fields differed.

## Performance
40 interleaved A/B pairs per estimator, `ci-fast`, `RAYON_NUM_THREADS=1`, current `main` versus PR head. Speedup is `(main − PR) / main`. Bootstrap intervals are 95% intervals for the mean of the paired percentage speedups (10,000 resamples).

| Estimator | main median | PR median | median speedup | paired mean speedup (95% CI) |
|---|---:|---:|---:|---:|
| FOCE | 0.490180 s | 0.481353 s | 1.80% | 5.31% (1.19%, 9.44%) |
| FOCEI | 0.382070 s | 0.375719 s | 1.66% | 1.01% (−5.79%, 6.96%) |
| Laplace | 0.376595 s | 0.362636 s | 3.71% | 0.58% (−2.93%, 3.89%) |
| AGQ | 0.402073 s | 0.395743 s | 1.57% | 5.33% (1.04%, 8.95%) |

FOCE and AGQ have positive paired intervals. FOCEI and Laplace have positive medians but their wall-time intervals cross zero on this small fixture, so this PR does not claim a statistically resolved timing win for those two.

Deterministic allocation instrumentation used the same current-`main` and PR builds. The analytical fixture exercises the shared FOCE/FOCEI/Laplace/AGQ inner path; the IOV fixture additionally exercises borrowed kappa storage and cached covariance data.

| Fixture | main allocation calls | PR allocation calls | reduction | main requested bytes | PR requested bytes | reduction |
|---|---:|---:|---:|---:|---:|---:|
| Analytical | 363,983 | 206,468 | 43.28% | 20,842,937 | 14,736,092 | 29.30% |
| IOV | 3,709,126 | 3,275,283 | 11.70% | 631,641,248 | 562,017,713 | 11.02% |

## Tests
- [x] Unit regression coverage added for reused NLL storage, borrowed IOV slices, and cached Hermite storage.
- [x] Compile, format, clippy, rustdoc, public-API, release-semantics, and TTE/CTMM coverage checks pass.
- [ ] Post-fix core coverage rerun is still in progress.

- [x] Not applicable (internal performance work; no API or behavior change)

## Docs
- [x] `CHANGELOG.md` performance entry added.
- [x] No user-visible behavior or documentation surface changed.

## Checklist
- [ ] Required CI/preflight-equivalent checks pending core coverage completion.
- [x] Dual2 formulas are unchanged; only f64 prediction/likelihood storage changed.
- [x] Cargo.toml version bump not needed

## Docs & examples
- [x] No DSL, fit-option, example, or data changes.
- [x] No example execution step needed beyond numerical regression runs.

## Reviewer hints
The subtle points are re-entrant buffer ownership across optimizer closures, bit-identical prior multiplication, and IOV borrowed-slice lifetimes.

## Open questions
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,10 @@ section of the SDLC for the versioning policy).
 
 ### Performance
 
+- **FOCE, FOCEI, Laplace, and AGQ now reuse prediction and prior-matrix storage across
+  repeated conditional-likelihood evaluations.** AGQ also caches invariant Hermite rules;
+  IOV quadrature borrows occasion effects and uses the covariance inverses already cached in
+  the model parameters instead of copying and refactorizing them at every node (#1374).
 
 - **ODE AutoSwitch now reuses accepted RK45 stages to detect stiffness without extra
   right-hand-side evaluations and carries its verdict history across dose and covariate
--- a/src/estimation/agq.rs
+++ b/src/estimation/agq.rs
@@ -24,7 +24,8 @@
 //!    term for term. This is not an approximation of an approximation; it is an identity,
 //!    and `tests::one_node_agq_equals_laplace` pins it.
 //! 2. **No Gaussian-residual assumption.** `l_i` is evaluated through
-//!    [`individual_nll_into_with_schedule`], the model's *actual* likelihood — so
+//!    [`crate::stats::likelihood::individual_nll_into_with_schedule`], the model's *actual*
+//!    likelihood — so
 //!    time-to-event and categorical endpoints are integrated as faithfully as Gaussian
 //!    ones. That is what FOCE/FOCEI structurally cannot do, and why AGQ exists here.
 //!
--- a/src/estimation/agq.rs
+++ b/src/estimation/agq.rs
@@ -81,9 +81,7 @@ use crate::estimation::importance_sampling::build_proposal;
 use crate::estimation::inner_optimizer::cacheable_schedule;
 use crate::pk;
 use crate::sens::provider::SubjectSens;
-use crate::stats::likelihood::{
-    individual_nll_into_prepared_with_schedule, individual_nll_into_with_schedule,
-};
+use crate::stats::likelihood::individual_nll_into_prepared_with_schedule;
 use crate::stats::util::log_sum_exp as logsumexp;
 use crate::types::{CompiledModel, HessianAnchor, ModelParameters, Population, Subject};
 
@@ -382,8 +380,50 @@ impl Stack {
         scratch: &mut pk::EventPkParams,
         schedule: Option<&pk::event_driven::EventSchedule>,
     ) -> f64 {
+        // Low-call-count sites (FD sweeps, single-node scoring): hoist nothing, just
+        // allocate scratch for this one call and dispatch through the shared branch.
+        let err_keys = (!self.is_iov()).then(|| model.error_spec.obs_keys(subject));
+        let ruv_mult = (!self.is_iov())
+            .then(|| model.ruv_obs_mult(subject, &params.theta))
+            .flatten();
+        self.nll_at_with_recycle(
+            model,
+            subject,
+            params,
+            b,
+            scratch,
+            schedule,
+            err_keys.as_deref(),
+            ruv_mult.as_deref(),
+            &mut Vec::new(),
+            &mut DVector::zeros(self.n_eta),
+            &mut DVector::zeros(self.n_eta),
+        )
+    }
+
+    /// Same dispatch as [`Stack::nll_at`], but takes the `err_keys`/`ruv_mult`/scratch as
+    /// caller-hoisted arguments instead of computing or allocating them per call. The hot
+    /// per-node sweep in [`agq_nodes_and_terms`] hoists these once for the whole grid (the
+    /// same hoist `node_nll_gradient` gets); routing both call shapes through this one
+    /// IOV/non-IOV branch keeps a single copy of the dispatch instead of two that must be
+    /// kept in sync.
+    #[allow(clippy::too_many_arguments)]
+    fn nll_at_with_recycle(
+        &self,
+        model: &CompiledModel,
+        subject: &Subject,
+        params: &ModelParameters,
+        b: &[f64],
+        scratch: &mut pk::EventPkParams,
+        schedule: Option<&pk::event_driven::EventSchedule>,
+        err_keys: Option<&[usize]>,
+        ruv_mult: Option<&[Vec<f64>]>,
+        pred_recycle: &mut Vec<f64>,
+        eta_work: &mut DVector<f64>,
+        prior_work: &mut DVector<f64>,
+    ) -> f64 {
         if !self.is_iov() {
-            return individual_nll_into_with_schedule(
+            return individual_nll_into_prepared_with_schedule(
                 model,
                 subject,
                 &params.theta,
@@ -395,6 +435,11 @@ impl Stack {
                 &params.residual_correlations,
                 scratch,
                 schedule,
+                err_keys.expect("non-IOV error keys"),
+                ruv_mult,
+                pred_recycle,
+                eta_work,
+                prior_work,
             );
         }
         let (eta, kappas) = self.split(b);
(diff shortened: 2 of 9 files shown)


hiddevandebeek:
Benchmark results are now included in the PR description. Across 40 interleaved current-main/PR pairs, median wall-clock speedups are FOCE 1.80%, FOCEI 1.66%, Laplace 3.71%, and AGQ 1.57%. Deterministic allocation instrumentation shows 43.28% fewer allocation calls / 29.30% fewer requested bytes on the shared analytical path, and 11.70% / 11.02% reductions on the IOV fixture. All runs preserved the displayed OFV and iteration count; the full FOCE comparison preserved parameters and sdtab exactly.

hiddevandebeek:
## Code Review (automated, high effort)

Reviewed via the checked-out PR head (`perf/shared-classical-hot-paths`, commit `729730fc`).

### 1. `individual_nll_iov_with_scratch`'s `omega_iov = None` branch now fails hard instead of degrading gracefully
`src/stats/likelihood.rs:2567-2570`

```rust
let (iov_inv, log_det_iov) = if let Some(iov) = omega_iov {
    (&iov.inv, iov.log_det)
} else {
    return 1e20;
};
```
Previously this branch produced `(DMatrix::identity(1, 1), 0.0)` — marked "unreachable when kappas non-empty" — so a caller that ever hit it (non-empty `kappas` with `omega_iov = None`) got a degenerate-but-finite prior instead of the divergence sentinel. The early `if kappas.is_empty() { return individual_nll(...) }` guard (unchanged, above this hunk) keeps today's only such caller (`test_individual_nll_iov_no_kappa_same_as_base`, kappas empty) out of this path, so nothing currently regresses — but the "unreachable" invariant is asserted only by a comment, not the type system, and this PR silently swapped its fallback for a hard failure. If a future caller (or a refactor of `Stack::is_iov()`/`Stack::new`) ever reaches this with non-empty kappas and no IOV omega, the subject will now report total divergence (1e20) instead of a wrong-but-diagnosable number, which is harder to root-cause. Worth either asserting the invariant explicitly (`debug_assert!(kappas.is_empty())` before returning) or documenting why 1e20 is now the intended behavior.

### 2/3. New per-node `DVector::zeros(...)` allocations reintroduce exactly the hot-path cost this PR removes elsewhere
- `src/estimation/agq.rs:2073-2112` (`node_nll_gradient`)
- `src/estimation/agq_cov_hessian.rs:675-724` (`node_jet`)

Both functions gained two new `&mut DVector::zeros(model.n_eta)` (plus a `&mut Vec::new()`, pre-existing) arguments to satisfy `analytic_eta_nll_gradient_with_schedule`'s new scratch parameters — allocated fresh **inside** the function, on every call.

`node_nll_gradient`'s own doc comment says: *"`schedule` and `mult` are **hoisted by the caller**... because this runs once per quadrature node — up to `MAX_AGQ_GRID` times per subject... paying it per node made the sweep several times its own cost."* Its production caller, `grid_response_correction` (`src/estimation/agq.rs:1600-1613`), already hoists `mult`/`err_keys` outside the `bs.iter().map(|b| node_nll_gradient(...))` loop for exactly that reason, but the two new `DVector`s are not hoisted the same way — they are allocated on every node of that loop.

`node_jet` is called the same way, inside the per-grid-node loop in the AGQ covariance/Hessian computation (`src/estimation/agq_cov_hessian.rs:851-864`).

Since this PR's entire premise is removing exactly this class of per-node allocation (the "Reuse AGQ node NLL scratch" bullet in the PR description), these two additions look like an oversight rather than an intentional trade-off — the buffers could be threaded through as caller-owned scratch the same way `find_ebe`/`agq_nodes_and_terms` now do it.

### 4. IOV/non-IOV dispatch duplicated instead of extended
`src/estimation/agq.rs:376-407` (`Stack::nll_at`) vs `src/estimation/agq.rs:788-822` (`agq_nodes_and_terms`)

`Stack::nll_at` is the documented single dispatch point ("Dispatches to the IOV-aware likelihood when the subject has occasions") and is still used by several call sites (gradient/Hessian FD, tests). `agq_nodes_and_terms`'s hot sweep, however, now reimplements the same `is_iov()` branch and kappa-slicing inline rather than extending `nll_at` to accept the new `err_keys`/`ruv_mult`/`pred_recycle`/`eta_work`/`prior_work` scratch. The two copies must now be kept in sync by hand; a future correctness fix to the IOV/non-IOV dispatch in one is easy to miss in the other.

### 5. Confusing scratch-buffer naming in `find_ebe`'s reentrancy fix
`src/estimation/inner_optimizer.rs:876-883, 900-906, 944-948`

The comment at line 879 explains that `obj`'s scratch must be kept separate from the analytic-gradient closure's scratch so FD fallback (which re-enters `obj` while the `match` scrutinee's temporaries are still borrowed) doesn't double-borrow a `RefCell`. The fix is correct, but the buffers `obj` actually uses are named `grad_eta_work`/`grad_prior_work`, while the plain `eta_work`/`prior_work` are used by `agrad`'s *gradient* computation — i.e. the names are swapped relative to which closure conceptually owns them. A future edit that "obviously" moves `eta_work`/`prior_work` into `obj` (because the names suggest they belong there) would reintroduce the double-borrow panic the preceding commit fixed.

---
Ranked by severity above; no blocking issues found in the numeric/allocation-reuse core logic itself — `individual_nll_into_prepared_with_schedule` and the `Vec<&[f64]>` kappa-borrowing changes look correct and are backed by the new bit-identical regression tests.

hiddevandebeek:
Pushed 4045bebd addressing the review findings from the earlier code-review pass:

1. **`node_nll_gradient` / `node_jet`** (`agq.rs`, `agq_cov_hessian.rs`): hoisted the `ObsGrad`/`eta_work`/`prior_work` scratch buffers to the caller instead of allocating fresh on every quadrature-node call — these loops run up to `MAX_AGQ_GRID` times per subject, the same class of allocation this PR exists to remove.
2. **`Stack::nll_at` / `agq_nodes_and_terms`**: both now dispatch through one new `nll_at_with_recycle` method instead of two separately-maintained copies of the IOV/non-IOV branch.
3. **`individual_nll_iov_with_scratch`**: the `omega_iov: None` fallback (documented as unreachable when `kappas` is non-empty) now backs that comment with a `debug_assert!` instead of silently changing behavior — loud in debug/test builds if the invariant is ever violated, unchanged in release.
4. **`find_ebe`**: un-inverted the `obj`/`agrad` scratch buffer names — `grad_eta_work`/`grad_prior_work` now actually belong to the gradient closure (`agrad`), not the objective (`obj`), removing the risk of a future regression reintroducing the double-borrow bug the reentrant-scratch split exists to prevent.

Verified with `cargo check`/`clippy --all-targets`/`fmt --check` (all clean) and targeted `cargo test --lib` runs covering every touched path (agq, agq_cov_hessian, sens_cov_hessian, inner_optimizer, likelihood — 177 tests, all green). A full-suite `--features ci` run kept getting OOM-killed by the shared host's other concurrent sessions, so it wasn't completed, but every module touched by this commit has direct, passing coverage.

Issue #621, opened 2026-06-30
Throw warning when a covariate cannot be mu-referenced in SAEM


roninsightrx:
Right now there is an info message that shows which individual parameters are mu-referenced, but it would be more informative if it would show which ones are NOT. This should be a warning rather than an info message as well, because it can strongly affect convergence.

PR #1375, opened 2026-09-12
feat(estimation): logit-normal mu-referencing for bounded (0,1) parameters [closes #918]

### LINKED ISSUE #918: SAEM: Add support for logit-normal MU-referencing for bounded (0,1) parameters

roninsightrx:
## Why

SAEM did not recognise a logit-normal individual parameter — `F = inv_logit(LOGIT_F + ETA_F)`, or the hand-written `1/(1 + exp(-(LOGIT_F + ETA_F)))` — as mu-referenced. The typical value then went through the eta-frozen numerical M-step instead of the closed-form EM shift, the fit warned `individual parameter(s) not mu-referenced: F`, and bounded `(0,1)` parameters came back biased low (−31 % to −57 % on the FREM reprex in #918, worse with IIV on the residual error). Closes #918.

This is the #918 work from `worktree-feat-918-logit-mu-ref` (Aug 14) brought forward onto today's `main`: the old branch was 632 commits behind and collided with the #996 identity-packing guard and the #1012 SAEM advisory.

## What changed

**Parser** — `MuRef.log_transformed: bool` becomes `MuRef.transform: MuTransform { Log, Identity, Logit, LogitProbability }`, with a `log_transformed()` compat method so every "is this eta lognormal?" consumer is unchanged. `detect_mu_refs` now tries the logit forms first (`inv_logit(θ+η)`, `inv_logit(logit(θ)+η)`, and both spellings of the written-out sigmoid). Detection also inlines eta-free, single-assignment local definitions, so NONMEM-style `MU_1 = log(TVCL); CL = exp(MU_1 + ETA_CL)` and `TVCL = …; CL = TVCL * exp(ETA_CL)` are recognised.

**SAEM / IMP / IMPMAP** — the closed-form EM update `g(θ) += γ·mean(η)` is link-independent; what it needs is that the *packed* theta is the mu. That eligibility now lives in one place, `saem::classify_mu_ref_pairs(model, theta_lower) -> MuRefPairs { eligible, identity_packed_log, log_packed_logit }`:

| transform | mu scale | closed form when |
|---|---|---|
| `Log` | `log θ` | theta is log-packed (lower ≥ 0) |
| `Logit` | `θ` | theta is identity-packed (lower < 0, the natural logit declaration) |
| `Identity`, `LogitProbability` | `θ`, `logit θ` | never — numerical M-step, as before |

The two "dropped" lists replace the per-estimator identity-packing filters #996 added on `main`, so the #996 advisory is unchanged and gains its mirror image: a logit-scale theta declared with a non-negative lower bound is log-packed, cannot take the closed form, and now says so (`… packed on the log scale …`, naming the theta and the bound to change). Both advisories are gated on `mu_referencing` and feed SAEM's `already_explained` set so the #1012 "NO associated ETA" advisory does not double-report them. `compute_mu_k` is transform-aware (`logit(θ)`, clamped, for the probability-scale form).

## Alternatives considered

Keeping `get_mu_ref_pairs(model)` as on `main` and filtering by packing at each call site (what #996 did) — rejected because the logit case needs the *opposite* packing, so two call sites would each carry a two-way rule that has to agree; one classifier with a test per cell is the version that cannot drift.

## Cross-repo dependency

ferx-r reads only `eta_log_transformed` (the per-omega bool vector), not `MuRef` — checked.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated. Added: `MuTransform` enum, `MuRef::transform`, `MuRef::log_transformed()`; removed: the `MuRef::log_transformed` field. Needed by anything constructing or matching a `MuRef` (ferx-r does neither).

<details>

Rust callers reading `mu_ref.log_transformed` call `mu_ref.log_transformed()`; callers constructing a `MuRef` set `transform: MuTransform::Log` (or `Identity`). No `.ferx` change.

</details>

## Numerical validation
- [x] Compared against: NONMEM 7.5.1 `METHOD=SAEM` with explicit `MU_3 = THETA(3)` on `nonmem_anchor/logit_fraction_saem.ctl` (parallel dual first-order absorption, pathway fraction FR1 logit-normal; 60 subjects, simulated from the model with FR1 = 0.6). A *pathway fraction* rather than bioavailability, since F is confounded with CL/V in oral-only data.
- [x] Reference values:

```
# FR1 (truth 0.6)           ω²(ETA_FR1) (truth 0.25)
NONMEM SAEM      0.604      0.183
ferx before      0.561      0.326
ferx after       0.594      0.186
```

`tests/saem_logit_mu_ref.rs::saem_recovers_the_logit_normal_fraction` (slow-gated) pins FR1 within 0.05 of both truth and NONMEM and ω² in (0.08, 0.45); re-run green after the rebase (release, `--features ci,slow-tests`). Every pre-existing mu-ref consumer produces exactly the same `MuRef` for every form recognised before this change (detection runs on the raw expression first; inlining is a fallback).

## Performance
- [x] No significant impact expected — one extra `classify_mu_ref_pairs` walk per fit; the closed form is a mean, as before.

## Tests
- [x] Unit tests: `classify_mu_ref_pairs_*` (one per cell of the table, plus de-dup), `compute_mu_k` for both logit transforms, parser tests for every logit spelling and the local-definition inlining (~40 in `model_parser_tests.rs`), and fit-level advisory tests for SAEM and IMP on the 3-subject #996 fixture (`saem_log_packed_logit_mu_ref_routes_to_numerical_mstep`, `imp_log_packed_logit_mu_ref_routes_to_weighted_mstep`, plus the `mu_referencing = false` suppression twin).
- [x] Regression test that fails without the fix: `logit_mu_ref_drives_the_saem_closed_form_m_step` (Tier-2) — the logit eta is the only random effect, so a non-zero saved-eval count is the signature of the closed form running; the slow anchor above fails on the pre-fix FR1.

- [x] Example inline below (also `nonmem_anchor/logit_fraction_saem_fit.ferx`)

<details>
<summary>Example</summary>

```toml
[parameters]
  theta LOGIT_FR1(-0.405, -10.0, 10.0)   # logit scale: negative lower bound keeps it identity-packed
  omega ETA_FR1 ~ 0.25

[individual_parameters]
  FR1 = inv_logit(LOGIT_FR1 + ETA_FR1)   # or 1/(1 + exp(-(LOGIT_FR1 + ETA_FR1)))
```

```
# no "individual parameter(s) not mu-referenced" warning; SAEM FR1 = 0.594 (NONMEM 0.604)
# with `theta LOGIT_FR1(0.4, 0.0, 10.0)` instead:
#   SAEM: typical value(s) LOGIT_FR1 are logit-mu-referenced but declared with a
#   non-negative lower bound, so they are packed on the log scale; ... (#918)
```

</details>

## Docs
- [x] `docs/estimation/saem.qmd` (eligibility table + the packing advisory), `docs/estimation/impmap.qmd`, `docs/model-file/individual-parameters.qmd`, `docs/faq.qmd`
- [x] `CHANGELOG.md` `[Unreleased]` entries added

## Checklist
- [x] No `Dual2` path touched
- [ ] `Cargo.toml` version bumped if breaking — not bumped; `MuRef` field rename is a pre-1.0 public-API change with no downstream user.

## Docs & examples
- [x] No example execution step needed beyond the anchor run above (no new `examples/` file).

## Reviewer hints

`src/estimation/saem.rs::classify_mu_ref_pairs` and its two call sites (`run_saem`, `run_mcem`) are the merge with #996 — that is where to look. The parser change is additive; `detect_pattern` is untouched and `classify_mu_ref` only wraps it. The IMP/IMPMAP mixture path still uses `MixtureMuRef.log_transformed` (a bool on a different struct) and is unchanged.

## Open questions

`LogitProbability` (`inv_logit(logit(TVF) + ETA)`) is detected but never closed-form eligible, because no packing produces the `logit θ` scale. A dedicated logit packing for a `(0,1)`-bounded theta would make it eligible; left for a follow-up since the logit-scale declaration already covers the use case.

https://claude.ai/code/session_01XawEdNoaXaKBfvSNo312ag


roninsightrx (commented):
I found two correctness gaps that should be addressed before merge, plus one documentation mismatch:

1. **[P2] Enforce one ETA owner per eligible theta in `classify_mu_ref_pairs` (`src/estimation/saem.rs:1491-1494`).** The new classifier appends every eligible `(theta_idx, eta_idx)` without de-duplicating/claiming theta indices. A valid model such as `F1 = inv_logit(LOGIT_F + ETA_F1)` and `F2 = inv_logit(LOGIT_F + ETA_F2)` therefore produces two eligible pairs for `LOGIT_F`. Both `run_saem` and `run_mcem` then update the same packed theta twice (`+= mean(ETA_F1)` followed by `+= mean(ETA_F2)`) and recenter each ETA independently. That is not the joint complete-data optimum and silently changes the fitted model. `get_mixture_mu_ref_pairs` already documents and enforces the required one-owner rule because “applying both shifts would move that θ twice”; the single-population classifier needs the same protection (or a proper joint update), with a regression test for two logit ETAs sharing one theta. This is newly reachable through the logit eligibility added here.

2. **[P2] The mixture estimator path drops all new logit mu-reference pairs (`src/estimation/saem.rs:1569-1572`).** SAEM and IMP/IMPMAP mixture fits do not consume `classify_mu_ref_pairs`; they consume `get_mixture_mu_ref_pairs`, whose ordinary/shared-theta fallback still requires `mu_ref.log_transformed()`. Consequently even a class-shared `F = inv_logit(LOGIT_F + ETA_F)` in a mixture model is detected in `model.mu_refs` but is excluded from the closed-form update. A `MIXNUM`-switched logit form is likewise not representable by the bool-only `MixtureMuRef`. The broad user-facing promise that SAEM and IMP/IMPMAP update logit-mu-referenced thetas therefore does not hold for mixture models, and the shared-theta case is especially surprising because the existing mixture machinery explicitly supports shared anchors. Please either carry transform/packing information through the mixture pair builder and cover both estimators, or explicitly scope the feature out and emit a targeted warning instead of silently routing it to the numerical M-step.

3. **[P2] Correct the `mu_referencing = false` statement in `docs/estimation/saem.qmd:136`.** It says the packing advisory is suppressed for “SAEM and IMP/IMPMAP” because no closed-form shift runs. That is only true for SAEM (and the mixture IMP path). Single-population `run_mcem` intentionally applies the EM shift independent of `options.mu_referencing`, and its new packing advisories are also emitted independent of that option; `docs/estimation/impmap.qmd` correctly documents this. Please qualify the SAEM text so it does not tell IMP users to expect behavior the implementation deliberately does not provide.

Focused verification run on `8d4f6f21`:

- `cargo test --lib test_detect_mu_ref` — 23 passed
- `cargo test --lib classify_mu_ref_pairs` — 2 passed
- `cargo test --lib imp_log_packed_logit_mu_ref_routes_to_weighted_mstep` — passed
- `cargo test --test saem_logit_mu_ref` — 3 passed, 1 slow test ignored

All GitHub CI checks were green at review time.

roninsightrx:
All three addressed in `c5befe61` (rebased onto `89a656be`, i.e. on top of the #619 merge that landed on this branch after your review).

**1. One ETA per anchor — fixed, and made stricter than you proposed.**

`classify_mu_ref_pairs` now drops a theta that anchors more than one ETA **and both its pairs**, rather than keeping the first claimant the way `get_mixture_mu_ref_pairs` does. Keeping one pair still applies `θ += γ·mean(η₁)` and then *pins* θ for the numerical M-step, so θ is held away from its joint optimum while η₂ is never re-centred and ω₂ inflates to absorb the drift. The joint maximiser is a precision-weighted mean of the two eta means, which this function does not compute, so the honest outcome is the numerical M-step — correct for any number of anchored etas. The asymmetry with the mixture rule is recorded at both sites.

Worth noting the trigger is not only the logit eligibility: `CL = TVP*exp(ETA_CL)` / `V = TVP*exp(ETA_V)` reached the same double shift before this PR. Both spellings are now covered.

SAEM and IMP/IMPMAP each emit an advisory naming the shared anchor. Tests: `classify_mu_ref_pairs_drops_a_theta_anchoring_two_logit_etas`, `classify_mu_ref_pairs_drops_a_shared_log_anchor_but_keeps_the_others` (with an unshared control theta in the same model), `imp_single_population_shared_anchor_routes_to_weighted_mstep`, and Tier-2 `saem_two_logit_etas_on_one_theta_route_to_the_numerical_mstep`.

*Interaction with #619, which merged into this branch in between:* `resolve_covariate_mu_groups` declines a covariate group whose theta another eta anchors, and it was reading the eligible-pair list. Handing it the post-drop list would have hidden exactly the thetas that are *most* contested, so the dropped pairs are handed back via the new `mu_ref_pairs_for_cov_groups` (a packing-dropped theta is deliberately *not* in that list — the group step does not work in the packed scale, so it may move it). Pinned by two unit tests.

**2. Mixture logit — implemented rather than scoped out.**

`MixtureMuRefPair` now carries its `MuTransform`, and the shared-theta fallback in `get_mixture_mu_ref_pairs` accepts `Log | Logit`. The packing check both estimators applied inline moved into one `classify_mixture_mu_ref_pairs` — the mixture twin of `classify_mu_ref_pairs` — so SAEM and IMP/IMPMAP cannot disagree about which class anchors the closed form owns; each caller keeps only what is genuinely its own (SAEM drops `MIXNUM`-switched anchors per #996, IMP/IMPMAP drops weak-IIV pairs per #411). The IMPMAP mixture path also gains the #918 advisory for a log-packed logit anchor, where only the #996 identity-packed message existed.

A `MIXNUM`-switched logit chain stays out of scope, as you noted the bool-only `MixtureMuRef` implies: `detect_mixture_pattern` recognises no switched logit pattern, so such an eta has no mu-reference at all and already surfaces through the existing "not mu-referenced" warning. That scope line is now stated in `docs/estimation/mixture.qmd` and in the builder's doc comment rather than left implicit.

Tests: `mixture_mu_ref_pairs_include_a_class_shared_logit_anchor`, `classify_mixture_mu_ref_pairs_gates_each_transform_on_its_own_packing` (same model, both packings, asserting the mismatch lands in the list whose advisory names the right bound), `classify_mixture_mu_ref_pairs_drops_a_pair_on_one_mispacked_class_theta`, plus unit-level fits `saem_mixture_takes_a_class_shared_logit_anchor`, `impmap_mixture_takes_a_class_shared_logit_anchor` and `impmap_mixture_log_packed_logit_anchor_warns`.

**3. Docs — corrected.**

`docs/estimation/saem.qmd` now says the advisory is suppressed under `mu_referencing = false` for **SAEM**, and that single-population IMP/IMPMAP applies the shift and the advisories independently of that option, pointing at the IMP/IMPMAP page. The shared-anchor rule is documented on all three pages (`saem.qmd`, `impmap.qmd`, `mixture.qmd`), and `CHANGELOG.md` carries the mixture-logit line under Added and the shared-anchor fix under Fixed.

**Verification**

- `tools/preflight.sh` green (fmt / check / clippy / rustdoc / docs / public-api).
- `cargo test --features ci --lib` — 4574 passed, 0 failed.
- `cargo test --features ci,survival --tests` — green.
- `cargo test --features ci,slow-tests --test mixture_nonmem` — 6 passed (56 s), so the #996 anchors did not move under the shared classifier refactor.

Each new assertion was mutation-checked against the smallest edit that removes the behaviour it pins, not just against toggling the fix: deleting the shared-anchor drop reddens 3 tests plus the Tier-2 fit; reverting the mixture builder to `log_transformed()` reddens 4; filtering the mixture pairs on `theta_packs_log_mask` alone (the pre-review predicate) reddens the per-transform gate test.

roninsightrx:
## Code review — PR #1375 (medium effort)

Note: this branch now also carries the squashed #1376 (covariate mu-referencing), so the review below covers both features.

The core logit work reads correct. `classify_mu_ref_pairs` / `packed_scale_is_mu` / `MuTransform` land on the right invariant — the closed-form EM mu-update is link-independent, and gating it on *packed scale == mu scale* is what makes that true rather than a coincidence. The mixture twin is consistent with the non-mixture path, and `compute_mu_k`'s clamping is right. Findings concentrate in the #619 group machinery and in the new parser inlining.

### 1. `src/parser/model_parser.rs:618` — the forward-reference guard is defeated by recursive inlining

The comment claims a line can only be rewritten with values already in scope, but `inline_defs` stores *raw* expressions and `inline_local_vars` resolves them against the map **as it stands at the use site**, not as it stood when the definition was recorded.

```
A   = FOO                  # FOO unknown here -> Covariate("FOO")
FOO = THETA_X
CL  = A * exp(ETA_CL)
```

At the `CL` line, detection substitutes `A -> FOO -> THETA_X` and records `MuRef { THETA_X, Log }` for `ETA_CL`. At eval time `CL` reads the data covariate `FOO`, so `THETA_X` is a theta that parameter never touches. SAEM then shifts and pins `THETA_X` and re-centres `ETA_CL` by that delta.

Fix: resolve each definition at the statement where it is recorded, rather than lazily at the use site.

### 2. `src/parser/model_parser.rs:1066` + `src/estimation/covariate_mu_ref.rs:328` — the exact engine assumes the group's *eta* is exclusive, but only *theta* exclusivity is checked

`retire_superseded_group` is keyed on `split_typical_and_eta`, which returns `None` for any shape it does not recognise. So:

```
CL = (TVCL + TH_X*(CRCL-90)) * exp(ETA_CL)
V  = TVV * exp(0.5*ETA_CL)        # or exp(ETA_CL + TH*COV)
```

leaves the group alive, because `V` is simply not recognised. The step's per-subject re-centring `eta_CL -= delta_i` then changes `V`'s predictions, so the data term is **not** constant in the group's thetas and `solve_exact` is not the M-step of anything — the same defect that `shared_thetas` exists to prevent, one level down.

Fix: add an "eta read by more than one typical value" check — any assignment whose RHS contains this eta and is not the group's own owner should force `needs_data_term` or drop the group.

### 3. `src/estimation/impmap.rs:1330` and `src/estimation/saem.rs:3503` — a failing group solve silently freezes its thetas for the whole fit

Both `solve_exact` and `solve_numerical` return `None` when any subject's `mu` is non-finite — i.e. when the additive typical value `TVCL + (CRCL-90)*TH_CRCL` goes <= 0 for one low-covariate subject, entirely reachable for the anchored `covmuref_additive` shape. The `continue` skips the group step, but the group's thetas were **already** pinned out of the weighted / numerical M-step (`mstep_theta_lower/upper[t] = log_theta[t]`, `temp_theta_*`), so no channel is left that can move them. They stay at their initial values, with no warning, and the fit reports a plausible-looking convergence.

Fix: on `None`, either unpin the group's thetas for that iteration or emit a warning (ideally both, plus a consecutive-failure count).

### 4. `src/estimation/covariate_mu_ref.rs:541` — BOBYQA's result is accepted unconditionally

```rust
match opt.optimize(&mut x) { Ok(_) | Err(_) => {} }
```

Every failure is swallowed and `x` is then used as the M-step point with no comparison against the starting objective. The objective returns the `1e20` sentinel whenever a trial theta makes any `mu` non-finite or the NLL non-finite, so on a mostly-inadmissible region BOBYQA wanders a flat plateau and can return a point strictly worse than `theta_old` — and since the group's thetas are pinned, nothing downstream corrects it.

Fix: evaluate `obj(x0)` and keep `theta_old` unless the returned point strictly improves it.

### 5. `src/estimation/saem.rs:1562` — the shared-anchor rule only scans `eligible`

A theta anchoring one eligible and one ineligible eta still takes the closed form:

```
CL = TVP * exp(ETA_CL)     # Log, log-packed  -> eligible
V  = TVP + ETA_V           # Identity         -> never eligible
```

`TVP` is not seen as shared, so it gets `+= gamma * mean(eta_CL)` and is pinned, while `ETA_V` is never re-centred for the realised delta — exactly the "pinned partial shift" the doc comment above it says this rule prevents.

Fix: detect sharing over **all** declared anchors, before the eligibility split, not over `eligible` only.

### 6. `src/api/fit.rs:979` and `src/estimation/impmap.rs:195` — the warning suppressors mirror only two of the group-resolution gates

`cov_groups_active` is derived from `model.covariate_mu_refs`, but `resolve_covariate_mu_groups` additionally drops a group for weak IIV (`omega^2 < 1e-3`), a conflicting single-anchor pair, or all-FIXed thetas. In those cases nothing mu-references the parameter, yet `saem_non_mu_referenced_individual_params_warning` is now suppressed — the #621 warning is lost precisely where it is true.

Same one level down: `theta_is_mu_ref_anchor_mask` extends the anchor set from `model.covariate_mu_refs` unconditionally, including under a mixture and under `mu_referencing = false`, suppressing the #1012 advisory for thetas that no group will ever move.

Fix: resolve the groups first and derive both predicates from the resolved list.

---

3 and 4 are the cheapest to fix and carry the largest silent-wrong-answer risk. 1, 2 and 5 are detection-scope holes and need widening before the group step runs.

roninsightrx:
All six review findings addressed in `dc8ee574`. Two of them did not survive contact with the code as stated — recorded here rather than quietly fixed.

| # | Finding | Outcome |
|---|---|---|
| 1 | forward-reference guard defeated by recursive inlining | **Fixed**, but **not reachable in production** — see below |
| 2 | exact engine assumes eta exclusivity, checks only thetas | Fixed |
| 3 | failing group solve silently freezes its thetas | **Fixed for IMP/IMPMAP only** — SAEM never froze; see below |
| 4 | BOBYQA result accepted unconditionally | Changed, but the scenario is **not reproducible** — see below |
| 5 | shared-anchor rule only scans `eligible` | Fixed |
| 6 | warning suppressors mirror only two of the resolution gates | Fixed, both sites |

### 1 — real defect in the detector, unreachable through a model file

`inline_local_vars` did walk `A → FOO → THX`, exactly as described, and the detector did record a mu-reference on a theta `CL` never reads. But the same three lines never reach it: #710's guard rejects the block first with

```
[individual_parameters]: forward reference(s) to name(s) declared later in the block: `FOO` (used in `A`).
```

(measured, not assumed). So the fix is defence in depth. It is still worth having — `record_inline_def` now resolves each definition where it is *written* and substitution is one level deep, so the invariant the comment claims holds locally instead of resting on a guard three phases away. The tests are at the statement level, since that is where the property lives, and each carries an in-scope control on the other side of the predicate so a fix that simply stopped inlining chains would fail.

### 3 — SAEM does not freeze; IMP/IMPMAP did

`saem.rs:3513` pins **inside** the `Some` arm (`let Some(theta_star) = solved else { continue }` comes first), so a failed solve there leaves the thetas free and `theta_sigma_mstep_light` moves them. `impmap.rs:1203` pins **before** `theta_sigma_weighted_mstep`, unconditionally — that half is real and is now gated on a new `CovariateMuGroup::can_step`, the exact predicate under which both engines return `None`. `imp_does_not_freeze_a_group_that_cannot_step` reddens on the minimal reversal with both thetas still bit-equal to their starts, `(4, 0.2)`.

The silence half applied to both, and both now report how many iterations a group could not step on.

### 4 — could not reproduce; changed anyway, differently

I probed for it: an objective finite only at θ_old and `1e20` everywhere else, at `maxiter` 1/2/3/5 (i.e. a budget too small for BOBYQA to finish its initial interpolation design). BOBYQA returned the start every time. It evaluates θ_old first and returns its own best, so a strictly-worse return is not reachable through this surface.

The proposed guard also is not free: `eval` calls `data_nll`, a full weighted NLL over every subject × draw, so re-evaluating the start and the result would cost **two extra full data-term passes per group per iteration**. Instead `solve_numerical` now tracks the best point as the search runs and returns that — zero extra evaluations, and independent of nlopt's return contract either way. The accompanying test pins the postcondition and says in its doc comment that it is not a regression test.

### Notes on 2 and 6

- The eta-sharing case does not drop the group; it routes it to the prior-plus-data engine, which evaluates the data term *with* the φ-preserving shift applied and is therefore correct for any number of eta consumers. Dropping would have cost the #619 benefit for a legitimate model.
- `CovariateMuRef::eta_shared` is `pub(crate)`, so `api/ferx-core-public-api.txt` is unchanged (`tools/update-public-api.sh --check` green).
- For 6 the `frem/mod.rs` caller keeps the *parsed* group thetas rather than resolved ones: conversion time has no population and no Ω, and each of those thetas does carry an ETA in the model text, so telling the user to add one would be wrong whichever engine ends up moving it. The asymmetry is recorded at the call site.

### Verification

11 new Tier-1 tests. Each was mutation-checked against the minimal reversal of its fix, and each mutation was checked to redden the intended side:

- the parser pair needs raw storage **and** recursive lookup restored — raw storage alone reddens the in-scope *control*, not the forward case, so both halves are pinned separately;
- the eta-sharing pair is mutated once in the parser (`eta_shared = false`) and once in the resolver (drop it from `needs_data_term`), and a different test dies each time;
- `mu_ref_anchor_mask_reads_the_resolved_groups_not_the_parsed_ones` and `saem_active_covariate_group_etas_drops_a_weak_iiv_group` each die when the site is reverted to reading `model.covariate_mu_refs`.

`cargo test --lib --features ci`: 4587 passed, 0 failed. `tools/preflight.sh`: green (fmt, check, clippy, rustdoc, docs, public-api). `docs/estimation/saem.qmd` and the `CHANGELOG.md` #619 entry updated for the third exact-engine condition, the new stand-down warning, and the restored advisories.

PR #1376, opened 2026-09-12
feat(estimation): covariate mu-referencing for typical values that read several thetas [closes #619]

### LINKED ISSUE #619: SAEM: covariate effect on IIV parameter collapses to bound when typical value isn't log-linear in one theta (non-mu-referenceable)

roninsightrx:
> **Stacked on #1375 (#918).** Base is `worktree-918-logit-mu-ref` so the diff shows only this change; retarget to `main` once #1375 merges. Closes #619.

## Why

SAEM could not estimate a covariate effect on an IIV parameter when the typical value is not log-linear in a single theta (#619). `CL = (TVCL_NR + (CKDCRCYS-90)*THETA_CLCKD) * exp(ETA_CL)` reads two thetas, so the detector recorded no mu-reference and **both** thetas went through the eta-frozen numerical M-step. That channel has no leverage on a covariate slope once the MH sampler has let η absorb the covariate-correlated between-subject variation: the slope and the etas drift together until the slope hits a bound. Measured on the fixtures in this PR with the pre-#619 binary:

| fixture | start | pre-#619 SAEM | NONMEM SAEM |
|---|---:|---:|---:|
| additive `TH_CRCL` | 0.02 | **0.0198** | 0.0418 |
| power `TH_WT` | 0.3 | **0.332** | 0.921 |

The power form (`TVCL * (WT/70)^TH_WT * exp(ETA_CL)`) is the same failure in the most common covariate shape there is: `TVCL` had its closed-form shift, the exponent sat on the numerical M-step and never moved.

## What changed

**Parser.** A typical value that reads two or more thetas is recorded as a `CovariateMuRef` (eta, the thetas it reads, the link, the covariates, and the inlined eta-free typical value) in the new `CompiledModel::covariate_mu_refs`. Shapes: `A * exp(ETA)` (any factor order; `exp(ETA + KAPPA)` accepted), `exp(log(A) + ETA)`, `inv_logit(A + ETA)` and the written-out sigmoid. Local definitions are seen through. Not recorded: a value that reads `TIME`, `MIXNUM`, a second eta, an NN output, a θ level block, or a local assigned inside an `if`. **`mu_refs` is untouched** — the power form keeps its `TVCL` anchor for the inner-loop centring, `suggest_start`, Bayes and reporting — so every FOCE/FOCEI/Laplace fit is byte-identical.

**Estimation** (`src/estimation/covariate_mu_ref.rs`). The mu-referencing M-step in its general form: hold each subject's φᵢ = g(Aᵢ(θ)) + ηᵢ fixed and re-fit the group's thetas to the population of individual values,

```
argmin_θ  Σᵢ ½ (φᵢ − g(Aᵢ(θ)))ᵀ Ω⁻¹ (φᵢ − g(Aᵢ(θ)))  +  Σᵢ −log p(yᵢ | φᵢ, θ)
```

- covariates **constant within each subject** → the data term is θ-free and the step is an exact nonlinear least squares (Gauss–Newton with LM damping; the Ω⁻¹ cross terms of a block Ω are folded into the target). For `g(Aᵢ) = log θ + cᵢ` it reduces to the classical `log θ += mean(η)`.
- a covariate **varying within a subject** → the within-subject ratio keeps the data term live, so prior + data are minimised by BOBYQA in packed coordinates, still in φ-frozen coordinates.

SAEM blends the result with its SA step γ and re-centres each η by the change in its own mu (the per-subject twin of the existing `e[eta_idx] -= delta`). IMP applies it to the importance-weighted posterior means and **shifts the stored draw means/second moments by −Δmuᵢ** so the next proposal stays centred in φ space — without that, plain IMP's sample-moment recentre sits several posterior SDs off after the step, ESS collapses to 0.03, and the biased mean re-fires the step: measured 0.041 → 0.10 in five iterations, then the bound. Group thetas are pinned out of the numerical M-step, counted in `saem_mu_ref_m_step_evals_saved`, and are mu-reference anchors for the #1011 damping gate and #1012 advisory. A group is declined — with a warning naming the thetas — when one of them is also another eta's single anchor, when the eta's initial IIV is < 1e-3, under a mixture model, and under `mu_referencing = false`.

This is NONMEM's `MU_1 = LOG(THETA(1) + (CRCL-90)*THETA(2))` — a MU nonlinear in THETA — detected and solved automatically.

## Alternatives considered

- **Widening `MuRef` to carry the group** (a placeholder `theta_name`) — rejected: every consumer does `theta_names.position(theta_name)` and would silently misread it, and `compute_mu_k` would centre the FOCE inner loop on a wrong mu, changing FOCE results on covariate models. A separate list touched by SAEM/IMP only keeps everything else bit-identical.
- **Switching the group's thetas to a joint numerical θ+Ω update** (the issue's "cheaper interim") — that is still the η-frozen coordinate system; the confounding is in the coordinates, not the optimizer.
- **Exact NLS only** — leaves the fluconazole model (CKDCRCYS time-varying in 27/31 subjects) on the old path; the data-term engine costs one population obs-NLL per BOBYQA evaluation, the same order as the numerical M-step it replaces for those thetas.

## Cross-repo dependency

ferx-r neither constructs `CompiledModel` nor reads `mu_refs`.

## Breaking changes
- [x] Public Rust API — baseline regenerated. Added `CompiledModel::covariate_mu_refs` and `CovariateMuRef { eta_name, theta_names, transform, covariate_names }` (the `typical` expression is `pub(crate)`). Needed by anything constructing a `CompiledModel` literal; no downstream caller does.

## Numerical validation
- [x] Compared against: NONMEM 7.5.1 `METHOD=SAEM` (NBURN=2000 NITER=1000 ISAMPLE=10, then EONLY IMP) on two matched fixtures simulated from the model under test (`nonmem_anchor/simulate_covmuref_data.py`, 60 subjects × 7 samples, covariates constant per subject). Both engines start at the same values away from the truth.

```
                       truth   start   ferx before   ferx after   NONMEM SAEM
additive  TVCL          5.0     4.0      3.990         4.921        4.917
          TH_CRCL       0.05    0.02     0.0198        0.04179      0.04178
          TVV          50      40       48.09         48.09        48.09
          ω²(ETA_CL)    0.09    0.1      0.083         0.0403       0.0401
          ω²(ETA_V)     0.04    0.1      0.0360        0.0358       0.0360
power     TVCL          5.0     4.0      4.919         4.787        4.788
          TH_WT         0.75    0.3      0.332         0.9221       0.9213
          TVV          50      40       49.51         49.50        49.52
          ω²(ETA_CL)    0.09    0.1      0.0975        0.0701       0.0697
          ω²(ETA_V)     0.04    0.1      0.0459        0.0460       0.0459
IMP       TH_CRCL       0.05    0.02     —             0.0417       0.0418
```

The pre-#619 additive fit also reported `TVCL ~ TH_CRCL` correlation 1.00 with RSEs of 46,000 % and 223,000 % — the ridge the η-frozen M-step cannot leave.

**Time-varying engine, fluconazole (#619's own model, CKDCRCYS varies within 27/31 subjects, SAEM seed 123, diagonal σ):** `THETA_CLCKD` 1.96 (start 2.0, i.e. frozen) → **1.18**, `TVCL_NR` 153.9 (start 150) → **109.0**, against NONMEM FOCE's 1.41 and 122. ferx's SAEM OFV on that dataset is 928 before / 942 after vs NONMEM's 734.6: the OFV gap predates this change (the FOCEI fit of the same model sits at 867 — the Q-collapse trap tracked with #864) and is not what #619 is about; the slope leaving its start is. No NONMEM SAEM comparator exists for this case, since NONMEM's MU must be constant within a subject.

## Performance
- [x] No significant impact expected for models without a group. With one: the exact engine is a handful of expression evaluations per M-step; the time-varying engine costs `maxiter·(d+2)` population obs-NLL evaluations per M-step, the same order as the numerical M-step those thetas used to occupy.

## Tests
- [x] Unit: 12 parser tests (every shape, inlining, IIV+IOV factor, seven rejections, last-assignment-wins, compiled model), 13 in `covariate_mu_ref_tests.rs` (mu, exact solver recovers planted θ to 1e-8, FIX, bounds, inadmissible start, block-Ω cross term with a diagonal control that must differ, numerical engine equals exact with a constant data term and passes the φ-preserving shift, time-varying detection, the three declines), fit-level SAEM ×3 / IMP ×1 on a 3-subject fixture where the group is the only closed-form channel (`saem_mu_ref_m_step_evals_saved` is `None` without it).
- [x] Regression tests that fail without the fix: the slow-gated anchors (`saem_recovers_the_additive_renal_gradient`, `saem_recovers_the_allometric_exponent`, `imp_recovers_the_additive_renal_gradient`) each assert clearance from the measured pre-#619 value; the IMP one is also the pin for the φ-preserving recentre. Fast Tier-2 twin: `saem_covariate_group_removes_the_not_mu_referenced_warning`.
- Full lib suite, preflight, and the slow anchors green locally (release).

- [x] `nonmem_anchor/covmuref_additive_saem_fit.ferx`, `covmuref_power_saem_fit.ferx`

<details>
<summary>Example</summary>

```toml
[individual_parameters]
  CL = (TVCL + (CRCL - 90.0) * TH_CRCL) * exp(ETA_CL)   # two thetas: covariate mu-reference
  V  = TVV * exp(ETA_V)
```

```
# before: "SAEM: individual parameter(s) not mu-referenced: CL", TH_CRCL 0.0198 (start 0.02)
# after:  no warning, TH_CRCL 0.04179 (NONMEM SAEM 0.04178)
```

</details>

## Docs
- [x] `docs/estimation/saem.qmd` (new M-step subsection + validation table), `docs/model-file/individual-parameters.qmd`, `docs/estimation/impmap.qmd`, `docs/faq.qmd`, `nonmem_anchor/README.md`

## Checklist
- [x] No `Dual2` path touched
- [ ] `Cargo.toml` version bumped — not bumped (additive public API)

## Docs & examples
- [x] No example execution step needed beyond the anchor runs above.

## Reviewer hints

`src/estimation/covariate_mu_ref.rs` is self-contained: `resolve_covariate_mu_groups`, `solve_exact`, `solve_numerical`. The estimator wiring is two blocks each in `run_saem` (group resolution next to `classify_mu_ref_pairs`; the group step inside the closed-form branch) and `run_mcem` (same, plus the draw-moment shift). The parser addition (`detect_covariate_mu_refs` and helpers) sits after `detect_mu_refs` and never edits its result. `CompiledModel` literal sites (27) gained one `covariate_mu_refs: Vec::new()` line each.

## Open questions

- The IMP proposal-recentre staleness also applies, in principle, to the existing single-anchor shift (`log θ += mean(η)` moves every subject's mu by the same amount without shifting the stored means). It converges in practice because that shift shrinks geometrically; this PR fixes it for groups only. Worth its own look.
- `only_mu_referenced_free_thetas` (the scalar residual σ gate) deliberately does not count group thetas; a time-constant group would qualify. Left conservative.

https://claude.ai/code/session_01XawEdNoaXaKBfvSNo312ag


src/parser/model_parser.rs
```diff
@@ -958,6 +958,291 @@ fn inline_local_vars(
...
+        let expr = if inline_defs.is_empty() {
+            raw_expr
+        } else {
+            inlined = inline_local_vars(raw_expr, &inline_defs, 0);
+            &inlined
+        };
+        if let Some((eta_idx, entry)) =
+            classify_covariate_mu_ref(expr, theta_names, eta_names, n_bsv_eta)
+        {
+            found.retain(|(e, _)| *e != eta_idx);
```
roninsightrx: [P1] Clear stale group metadata when a later assignment no longer matches. This vector is only updated inside the Some branch, so `CL = (TVCL + TH_X*WT)*exp(ETA_CL)` followed by `CL = TVCL*exp(ETA_CL)` leaves the overwritten two-theta group alive. I confirmed that case with a parser regression test: SAEM/IMP then prefer the stale group over the actual scalar MuRef, pin TH_X out of the numerical M-step, and update from an expression the model no longer uses. Please track the last assignment even when classification returns None and clear any prior group for that eta/parameter.

src/estimation/covariate_mu_ref.rs
```diff
@@ -0,0 +1,513 @@
...
+            notes.push(format!(
+                "covariate mu-reference on {} (reads {}) is not used: its random effect has \
+                 negligible variance (ω² < {WEAK_GROUP_IIV_VAR:.0e}), so the population of \
+                 individual values carries no information about the typical value; {} stay on \
+                 the numerical M-step (#619).",
+                spec.eta_name, names, names
+            ));
+            continue;
+        }
+        let time_varying = spec.covariate_names.iter().any(|c| {
```
roninsightrx, line 196: [P1] Do not select the exact engine solely from within-subject covariate constancy. That only makes the observation term theta-free if every group theta affects the model exclusively through this one typical value. A valid model such as `CL = (TVCL + TH_X*WT)*exp(ETA_CL); V = TVV + TH_X` resolves this group as time-constant, but after preserving CL phi the data still depend on TH_X through V. The exact solver then omits that live data term while SAEM/IMP pin TH_X out of their general numerical M-step, so this is not an EM M-step and can return incorrect estimates. Please prove exclusive theta ownership when resolving the group, or route shared thetas through the prior-plus-data engine.

src/api/fit.rs
```diff
@@ -889,7 +889,13 @@ pub(crate) fn saem_non_mu_referenced_individual_params_warning(
         let Some(eta_name) = model.eta_names.get(eta_idx as usize) else {
             continue;
         };
-        if !model.mu_refs.contains_key(eta_name) {
+        // A multi-theta typical value (#619) is mu-referenced too: SAEM moves
+        // its thetas through the covariate mu-reference group.
+        let has_group = model
```
roninsightrx: [P2] Suppress this warning only for a group that is active in the actual SAEM run. This model-only check now hides the warning when `mu_referencing = false`, even though run_saem deliberately creates no groups in that mode; the comment above this call says that is exactly when the warning matters most. It also hides it for mixture models, where run_saem skips groups without adding the promised declined-group warning. Please make this predicate aware of the fit mode/resolved groups, or preserve the non-mu-referenced warning for disabled and unsupported group paths.

roninsightrx (commented):
Review complete: requesting fixes before merge. I found two correctness blockers in group detection/routing and one warning regression; details are in the three inline comments. Validation run: cargo test --lib (4551 passed, 23 ignored), cargo check --tests, the focused integration test, and the public-API baseline check all pass on 335e3992.

src/parser/model_parser.rs
roninsightrx: Confirmed and fixed in ff14fb9c.

Retirement is now unconditional and runs before recording, keyed two ways: on the eta the new right-hand side carries (`split_typical_and_eta`, the same key recording uses) **and** on the assigned parameter name. The name key is what covers the case where the overwrite carries no eta at all (`CL = TVCL`), where there is no eta to match on.

I also extended it to `if` bodies, which had the same hole from the other side: a branch can overwrite a typical value recorded above it, so it retires — but it may never *record*, since the branch need not be taken.

Four regression tests, all in `model_parser_tests.rs`:

- `covariate_mu_ref_retires_when_the_overwrite_is_a_plain_mu_ref` — exactly your repro, asserting both halves together (`covariate_mu_refs` empty **and** the scalar `MuRef` present), since the defect is a disagreement between the two scans rather than a missing entry in either;
- `covariate_mu_ref_retires_on_an_eta_free_overwrite`;
- `covariate_mu_ref_retires_from_a_conditional_overwrite`, which also pins that the reverse order records nothing;
- `covariate_mu_ref_retirement_spares_unrelated_assignments` — the control, so the three above are not satisfied by a detector that retires everything.

src/estimation/covariate_mu_ref.rs
roninsightrx: Agreed — your `V = TVV + TH_X` case is real, and worse than "the exact solver omits a live term": the group has already pinned `TH_X` out of the estimators' general numerical M-step, so nothing maximises that term at all. Fixed in ff14fb9c by proving ownership.

`CovariateMuRef` gains `shared_thetas`, computed at parse time. The test is **reachability**, not occurrence — an occurrence no consumer reads changes no prediction:

- a name is *tainted* by θ when its right-hand side reads θ or reads an already-tainted name (statement order, `if` bodies included);
- the taint **stops at the group's own parameter**. That is the content of mu-referencing: `CL` is held by the preserved `φ_i`, so `Q = CL/2` is held too, and the local `TVCL = …` that exists only to build `CL` carries its thetas nowhere else;
- θ is shared when another block names θ itself, or names any tainted name. `[structural_model]` / `[odes]` / `[derived]` / `[scaling]` / `[error_model]` are where an individual parameter becomes a prediction, so a taint reaching none of them reaches no observation (a dead intermediate).

The last step is an identifier scan of raw block text, so it over-reports on a comment or a colliding name. That is the safe direction: wrongly shared costs the slower engine, wrongly exclusive returns wrong estimates.

`resolve_covariate_mu_groups` routes a group with a shared **free** theta to the prior-plus-data engine — which keeps the term, and whose `data_nll` already receives the full θ vector, so the `V` dependence is picked up — and pushes a note naming the theta. A FIXed shared theta never moves, so it cannot be mis-maximised and does not force the switch.

This also closes a case the existing `claimed` guard left open: when two groups share a theta the second is declined, but the first was still taking the exact engine while the declined group's parameter kept the theta live. The taint walk now sees that statement like any other.

`CovariateMuGroup::time_varying` is renamed `needs_data_term`, so the two independent reasons to keep the term are one field rather than a covariate fact each caller reinterprets.

Tests: three parser tests covering the escape routes (another parameter, a local definition, another block), each with the control that pins the opposite answer; and `a_theta_the_rest_of_the_model_reads_keeps_the_data_term`, which resolves a real model where `TH_CRCL` also sets `F`, asserts the engine switch and the note, and asserts the FIXed-theta control so the routing is not satisfied by the mere presence of a name in `shared_thetas`.

src/api/fit.rs
roninsightrx: Right on both counts — fixed in ff14fb9c. The predicate takes a `covariate_groups_active: bool` and the call site computes it with the same two gates `run_saem` applies before resolving groups (`options.mu_referencing && model.mixture.is_none()`), so the warning survives `mu_referencing = false` — the case it exists for (#621) — and survives a mixture, where no group is built.

Test: `saem_non_mu_referenced_warning_returns_when_the_run_builds_no_covariate_group` parses a model whose `CL` carries a group and whose `V` has an ordinary single-anchor mu-ref, then asserts the warning is absent with the flag on, present and naming only `CL` with it off.

PR #1378, opened 2026-09-13
perf(estimation): cache per-subject EventSchedule across outer-loop evaluations [closes #1345]

### LINKED ISSUE #1345: perf(estimation): optimization candidates for the classical estimation path (FO/FOCE/FOCEI, Laplace/AGQ, SAEM, IMP)

hiddevandebeek:
## Why

#1345 (section E1) identified that `cacheable_schedule`-gated `EventSchedule` construction is
rebuilt from scratch on every AGQ node / gradient call within a single outer NLL evaluation,
even though the schedule depends only on the model and the subject's data — never on the
trial parameter vector. For FOCEI, Laplace, and `focei` with `n_agq > 1`, that means the same
subject's schedule is rebuilt many times per outer iteration.

## What changed

- `estimation/inner_optimizer.rs`: `find_ebe` now delegates to a new private `find_ebe_impl`
  that accepts an optional pre-built `EventSchedule`. A new `pub(crate) build_schedule_cache`
  maps `cacheable_schedule` once over every subject in a `Population`, and new
  `find_ebe_cached` / `run_inner_loop_warm_cached` / `run_inner_loop_warm_map_cached` reuse
  that cache instead of recomputing it.
- `estimation/agq.rs`: new `agq_population_nll_with_schedules` /
  `agq_population_evaluate_with_schedules` thread the same cache through the AGQ population
  loop.
- `estimation/outer_optimizer.rs`: `run_inner_loop_and_nll` / `run_inner_loop_and_nll_prepared`
  and `run_global_presearch` build the schedule cache once per outer-loop entry point and
  dispatch to the cached or uncached path depending on whether it's available.

Both the cached and uncached paths funnel into the same `find_ebe_impl` /
`run_inner_loop_warm_map_impl`, so correctness only depends on the cache's per-subject index
staying aligned with `population.subjects` — the risk two new regression tests target directly
(see Tests below). `cacheable_schedule`'s existing guard (time-varying covariates or `EVID=3/4`
resets; declines on lagtime, bioavailability-with-rate, algebraic, and compartment-free models)
is unchanged.

Two other findings from the same research pass are **not** in this PR:
- `omega_joint_inv`/`Stack` caching — evaluated and deprioritized (low expected value relative
  to implementation/review cost for this pass).
- `packed_fixed_mask` caching in `SubjectScoreContext::new` — already shipped in #1374 before
  this branch existed; nothing left to do here.

## Alternatives considered

Caching at the `Population` level (a `HashMap<subject id, EventSchedule>`) was considered and
rejected in favor of a plain `Vec<Option<EventSchedule>>` indexed by subject position, since
every caller already iterates subjects by index and a `Vec` avoids a hash lookup per subject
per AGQ node.

## Cross-repo dependency

Not needed — `pub(crate)` only, no `ferx-r` impact.

## Breaking changes

- [x] None

## Numerical validation

- [x] Compared against: prior ferx commit `caec0477` (branch base, verified bit-identical to
      `origin/main` at the time this branch was cut)
- [x] Reference values: bit-identical OFVs (`.to_bits()` equality) between the cached and
      uncached paths across FOCEI, Laplace, and `focei`+AGQ(n=3), verified both in the new
      unit tests and via the full `cargo test --lib` suite (4509 passed, 0 failed, 23 ignored —
      no regressions vs. the pre-PR baseline of 4507/0/23).

## Performance

- [x] Faster — benchmark: see the `CHANGELOG.md` entry under `### Performance`. Measured on
      `ci-fast`, single Rayon thread (`RAYON_NUM_THREADS=1`): no resolvable difference on the
      24-subject/2-occasion Schnider propofol fixture (schedule construction is a small share
      of per-eval cost there); a synthetic 40-subject/15-occasion-per-subject stress fixture
      (built specifically to amplify per-subject schedule-construction cost) showed a modest,
      directionally consistent ~3–6% wall-clock reduction across FOCEI, Laplace, and
      `focei`+AGQ(n=3). No claim is made beyond that stress scenario.

## Tests

- [x] Regression test added — fails without fix

Two new tests close a coverage gap: the existing `focei_pipeline_tests.rs` updates only ever
pass `None` for the new `schedules` parameter, so no prior test exercised the `Some(schedule)`
cached path.

- `estimation::inner_optimizer::tests::cached_schedule_inner_loop_matches_uncached` — mixes a
  subject with an `EVID 3/4` reset (`cacheable_schedule` → `Some`) and one with neither a reset
  nor a time-varying covariate (`None`) in the same population, and asserts
  `run_inner_loop_warm_cached` is bit-identical to `run_inner_loop_warm`.
- `estimation::agq::tests::cached_schedule_agq_population_matches_uncached` — same mixed-population
  construction, asserting `agq_population_nll_with_schedules` /
  `agq_population_evaluate_with_schedules` (NLL and gradient) are bit-identical to their
  uncached counterparts for both FOCEI and Laplace.

Both were specifically designed to catch a cache-index/`population.subjects` misalignment,
which a population where every subject takes the same `Some`/`None` branch could not expose.

## Docs

- [x] `CHANGELOG.md` `[Unreleased]` entry added (under `### Performance`)
- [x] No user-visible change beyond the changelog entry (no new fit option, DSL syntax, or
      output field)

## Checklist

- [x] `tools/preflight.sh` green — fmt, all 5 `cargo check` feature sets, clippy
      (`--all-targets`, both invocations), rustdoc (both invocations), `docs-lint`
      (test + clippy), and the public-API baseline check all ran individually to completion
      with exit code 0 (run gate-by-gate rather than via the single `preflight.sh` invocation,
      due to heavy external memory contention on the build box from concurrent sessions —
      each gate's cargo invocation matches exactly what `tools/preflight.sh --list` prints).
- [x] No `sens/`/`pk/` `*_g<T: PkNum>` gradient-reachable code touched — the new functions
      wrap the existing inner-loop/AGQ dispatch, they don't change any analytic sensitivity.
- [x] `Cargo.toml` version not bumped (no breaking change).

## Docs & examples

- [x] Not applicable — no new DSL syntax, fit option, example file, or estimator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,19 @@ section of the SDLC for the versioning policy).
 
 ### Performance
 
+- **FOCEI, Laplace, and `focei` with `n_agq > 1` now build each subject's `EventSchedule`
+  once per outer-loop evaluation instead of once per subject per AGQ node/gradient call.**
+  `cacheable_schedule` gates this on time-varying covariates or `EVID=3/4` resets (its
+  guard is otherwise unchanged), so a fit with only baseline covariates and no resets never
+  allocated a cacheable schedule in the first place and sees no change. Verified bit-identical
+  OFVs against `main` across every configuration tested. Measured on `ci-fast`, single Rayon
+  thread: no resolvable difference on the 24-subject/2-occasion-per-subject Schnider propofol
+  fixture (`tests/schnider_propofol_nonmem.rs`), where schedule construction is a small share
+  of per-eval cost; a synthetic 40-subject/15-occasion-per-subject stress fixture (built to
+  amplify per-subject schedule-construction cost) showed a modest, directionally consistent
+  ~3–6% wall-clock reduction across FOCEI, Laplace, and `focei`+AGQ(n=3). No claim is made
+  beyond that stress scenario (#1345).
+
 - **FOCE, FOCEI, Laplace, and AGQ now reuse prediction and prior-matrix storage across
   repeated conditional-likelihood evaluations.** AGQ also caches invariant Hermite rules;
   IOV quadrature borrows occasion effects and uses the covariance inverses already cached in
--- a/src/estimation/agq.rs
+++ b/src/estimation/agq.rs
@@ -653,6 +653,7 @@ fn anchor_hessian(
 /// re-optimise it, it only lays the grid around it. `nodes`/`log_weights` are the 1-D
 /// Gauss–Hermite rule, hoisted by the caller so the eigenproblem is solved once per
 /// population rather than once per subject.
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn agq_subject_nll(
     model: &CompiledModel,
     subject: &Subject,
@@ -662,6 +663,7 @@ pub(crate) fn agq_subject_nll(
     nodes: &[f64],
     log_weights: &[f64],
     anchor: HessianAnchor,
+    schedule: Option<&pk::event_driven::EventSchedule>,
 ) -> f64 {
     agq_subject_evaluate(
         model,
@@ -673,10 +675,17 @@ pub(crate) fn agq_subject_nll(
         log_weights,
         anchor,
         false,
+        schedule,
     )
     .0
 }
 
+/// Same as the body [`agq_subject_nll`] delegates to, but `schedule` is now a caller-supplied
+/// parameter instead of a fresh [`cacheable_schedule`] rebuild on every call — see
+/// [`crate::estimation::inner_optimizer::build_schedule_cache`]. It depends only on
+/// `model` + `subject`, so the two population-level callers ([`agq_population_nll`],
+/// [`agq_population_evaluate`]) build the cache once per outer-loop evaluation and hand each
+/// subject its entry, rather than every subject re-walking its own dose/event timeline.
 #[allow(clippy::too_many_arguments)]
 fn agq_subject_evaluate(
     model: &CompiledModel,
@@ -688,23 +697,16 @@ fn agq_subject_evaluate(
     log_weights: &[f64],
     anchor: HessianAnchor,
     retain_gradient_work: bool,
+    schedule: Option<&pk::event_driven::EventSchedule>,
 ) -> (f64, Option<PreparedGrid>) {
     let d = stack.d();
     let mut scratch = pk::EventPkParams::with_capacity_for(subject);
-    let schedule = cacheable_schedule(model, subject);
 
     // No random effects: the "integral" is a point mass and AGQ degenerates to the
     // conditional likelihood itself. (The formula below would agree — an empty tensor
     // product is the single empty node — but there is no Σ to factor, so short-circuit.)
     if d == 0 {
-        let nll = stack.nll_at(
-            model,
-            subject,
-            params,
-            b_hat,
-            &mut scratch,
-            schedule.as_ref(),
-        );
+        let nll = stack.nll_at(model, subject, params, b_hat, &mut scratch, schedule);
         return (nll, None);
     }
 
@@ -717,7 +719,7 @@ fn agq_subject_evaluate(
             stack,
             b_hat,
             &mut scratch,
-            schedule.as_ref(),
+            schedule,
         )
     } else {
         anchor_hessian(
@@ -728,7 +730,7 @@ fn agq_subject_evaluate(
             stack,
             b_hat,
             &mut scratch,
-            schedule.as_ref(),
+            schedule,
         )
         .map(|h| (h, None))
     };
@@ -756,7 +758,7 @@ fn agq_subject_evaluate(
         log_weights,
         &proposal,
         &mut scratch,
-        schedule.as_ref(),
+        schedule,
         retain_gradient_work,
     );
 
@@ -903,6 +905,7 @@ pub(crate) fn agq_subject_objective(
         &nodes,
         &log_weights,
         HessianAnchor::GaussNewton,
+        None,
     )
 }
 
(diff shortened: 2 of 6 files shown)


hiddevandebeek:
Reviewed commit `e3d1fba093a202e30f2c02065075abc69f005124`.

1. **[P2] Preserve local schedule reuse when no population cache is supplied** — [`src/estimation/agq.rs:1053`](https://github.com/FeRx-NLME/ferx-core/blob/e3d1fba093a202e30f2c02065075abc69f005124/src/estimation/agq.rs#L1053), also line 1154 and the direct `agq_subject_nll(..., None)` callers.

   The existing population wrappers pass `schedules = None`, which now becomes `schedule = None` all the way through `agq_subject_evaluate`. Previously that function built `cacheable_schedule(model, subject)` once and reused it for the anchor Hessian and every quadrature node. For eligible analytical subjects with time-varying covariates or resets (without IOV), the new path falls through `compute_predictions_with_tv_recycle_with_schedule` to `event_driven_predictions`, which constructs `EventSchedule::for_subject` on **every prediction**. Thus the public AGQ objective, built-in BFGS, covariance objective stencils through `pop_nll_opts`, and reconverged finite-difference scores lose the existing reuse; node work alone can rebuild the timeline `n_nodes^d` times per subject evaluation. This is a regression outside the newly cached NLopt path. Please retain an owned per-subject fallback when the population cache is absent, and preserve local reuse for direct subject-objective callers. A supplied cache entry of `None` should still mean eligibility was already checked.

Testing note: `cached_schedule_agq_population_matches_uncached` adds the reset/dose at time 12, but every observation is at time 8 or earlier. Add observations after the reset (and preferably subjects with distinct effective schedules) so the test checks that the correct schedule affects NLL/gradient values, rather than only that a `Some` entry exists. The current value-equality assertions also cannot detect the repeated-construction regression above.

The cache lifetime and population index alignment in the new NLopt/presearch paths look consistent on inspection. GitHub reports 11 passing checks and no failures. I attempted `cargo test --offline --lib cached_schedule --profile ci-fast` locally, but stopped it after several minutes without a result; I am not claiming a local test pass or a measured slowdown. The finding above follows from the changed call path. No numerical-correctness defect was established in this review.

hiddevandebeek:
I implemented the regression fix discussed earlier.

In `src/estimation/agq.rs`, I restored the per-subject `cacheable_schedule(model, subject)` fallback when no schedule is supplied, while keeping the new cached hot-path behavior for callers that pass `schedules[i]`.

What changed:

- `agq_subject_evaluate(...)` now resolves a local `cacheable_schedule` only when the caller passes `schedule = None`.
- `agq_population_evaluate_impl(...)` now resolves one schedule per subject for the subject pass even without a provided cache entry, and reuses that same reference for both:
  - AGQ objective node sweep (`agq_subject_evaluate`)
  - AGQ gradient prep (`score_with_prepared`)

Net effect:

- `agq_population_nll` and `agq_population_evaluate` no longer lose the previous local schedule reuse for uncached direct paths.
- Cached population calls still reuse caller-hoisted schedules and avoid repeated `EventSchedule` construction.

No tests were run in this step.

PR #610, opened 2026-06-29
feat(parser): add TIME event-time built-ins [closes #607]

roninsightrx:
## Why
Closes #607.

ferx did not have NONMEM-style `TIME`/`time` built-ins for PK parameter expressions, so model authors could not write event-time switches such as `if (TIME > 45)` in `[individual_parameters]` without treating `TIME` as a data covariate.

## What changed
- Added `TIME`/`time` as parser built-ins for `[individual_parameters]` expressions.
- Added direct analytical `pk(...=TIME)` / `pk(...=time)` mappings.
- Kept `TIME`/`time` out of covariate detection.
- Evaluated PK parameters per dose, observation, and `EVID=2` event when a model uses the time built-in.
- Routed analytic sensitivity providers to the finite-difference fallback for time-dependent PK models until the event-time context is carried through those paths.
- Updated model-file docs and `CHANGELOG.md`.

## Alternatives considered
Treating `TIME` as an ordinary covariate would require users to declare or provide it and would not match NONMEM's event-time `$PK` behavior. The parser built-in keeps the behavior explicit and avoids accidental covariate requirements.

## Cross-repo dependency
| Repo | PR | Status | Must merge |
|------|----|--------|------------|
| ferx (R) | — | not needed | — |

No public Rust API or lockfile-visible dependency change is required for ferx-r.

## Breaking changes
- [ ] `.ferx` model file format (add migration note below)
- [ ] `FitResult` / sdtab fields changed (ferx parses these — link ferx PR above)
- [ ] Public Rust API (`api.rs` / `types.rs`)
- [x] None

<details>
<summary>Migration notes (if breaking)</summary>

None.

</details>

## Numerical validation
This change affects PK parameter evaluation timing rather than adding a new estimator or output statistic. The added regression test validates that dose and observation events see their own event times, including lowercase direct `pk(...=time)` mapping, without declaring `TIME` as a covariate.

- [ ] Warfarin example converges and estimates are within tolerance of prior run
- [ ] Compared against: NONMEM / nlmixr2 / prior ferx commit `______`
- [ ] Reference values (paste key theta/omega/OFV, or link to output file):

```
# before / reference
N/A

# after
N/A
```

- [ ] Not applicable (docs / refactor / CI only)

## Performance
- [x] No significant impact expected
- [ ] Faster — benchmark: `______` (before) → `______` (after)
- [ ] Slower — justified because: `______`

Models that do not use `TIME`/`time` keep the existing subject-static PK parameter path.

## Tests
[...]

PR #1379, opened 2026-09-13
fix(api): read individual-parameter values by name, not by PK slot [closes #1356]

### LINKED ISSUE #1356: bug(output): an analytical individual parameter not bound in [structural_model] (e.g. TVCL) reports CL's value in [output], [derived] and ferx-r individual_estimates
## Summary

In an **analytical** (`pk ...`) model, some top-level `[individual_parameters]` names are not bound on the `[structural_model]` line. An intermediate such as `TVCL = THCL * 3` is the typical case. Such a name gets `pk_indices[i] = 0`, a placeholder from `unwrap_or(0)` at `src/parser/model_parser.rs:2823`, and for most such names `pk_param_fn` never writes the value anywhere. A modeled-dose `D{n}`/`R{n}` is the exception: it is written to a spare slot above `PK_IDX_LAGTIME` (`model_parser.rs:2316`–`2418`), but its `pk_indices` entry is still `0` (measured in the independent review of #1355: `D1 = 3.3` sits at slot 9, `pk_indices[D1] = 0`). Either way, every consumer that looks up a name's value as `PkParams.values[pk_indices[i]]` **silently returns CL's value** for that name.

ODE and compartment-free models are not affected. Every name there gets a real slot from `ode_param_slots`, and the same measurement below confirms `V, KA, KE, LAGTIME → slots 1, 4, 0, 8` with correct values.

## Measured

Debug build of `origin/main` `1c0c1626`, run through the public API (`parse_model_string` → `pk_param_fn` at the default θ, η = 0):

```
[individual_parameters]
  TVCL = THCL * 3              # THCL = 2.0  → TVCL = 6.0
  CL   = TVCL * 2 * exp(ETA_CL) #             → CL   = 12.0
  V    = THV
  KA   = THKA
[structural_model]
```

```
pk_indices = [0, 0, 1, 4]
 TVCL pk_indices=0 values[pk_indices]=12.0000   <- true TVCL is 6.0
   CL pk_indices=0 values[pk_indices]=12.0000
    V pk_indices=1 values[pk_indices]=10.0000
   KA pk_indices=4 values[pk_indices]=1.5000
```

The fixture has to be non-degenerate. With the common `CL = TVCL * exp(ETA_CL)` at η = 0, CL equals TVCL and the wrong read gives the right number, so the bug cannot be seen. That is also why a quick eyeball of an sdtab usually won't catch it.

## Where it reaches users

| consumer | path | effect |
|---|---|---|
| ferx-core sdtab, `[output] TVCL` | `validate_output_columns` accepts any `indiv_param_names` entry (`src/api/validation.rs:5996`), then `build_indiv_map` (`src/api/output_columns.rs:97`, called at `:247` from `fit()` via `src/api/fit.rs:1976`) | the TVCL column holds CL |
| ferx-core `[derived]` expression that reads `TVCL` | same map, passed as `DerivedContext.indiv_params` (`output_columns.rs:381`) and bound by name at `src/parser/model_parser.rs:4541` | the expression computes with CL |
| ferx-r `individual_estimates` | `build_individual_estimates` reads `values[pk_indices[i]]` on the analytical engine (ferx-r `src/rust/src/lib.rs:3040`–`3045`, `main` `9d48df0`) | the TVCL column holds CL |

The `pk_indices` read is **measured** (above). The sdtab `[output]` row is also measured end to end through `fit()` (see the comment below: an intermediate `KAF` column holds CL), and the bundled `examples/tte_exponential.ferx` reproduces it (`LAMBDA` reported as `DUMMY_CL` = 1). The `[derived]` row follows the same read through the code; it was not run.

**Prior art:** #455 hit the same placeholder in the TV-cov Dual2 gate (`model.pk_indices = [0, 0, 0, ..., 0, 1, 2, 3]`). #456 fixed the gate, and the placeholder stayed. #1355 documents the placeholder on `CompiledModel::indiv_param_names`.

**Separate ferx-r bug:** ferx-r's ODE arm also reads slot `i` instead of `pk_indices[i]`, so a column can show an unwritten 0 or another parameter's value (measured: V's column shows KE's value, KA's column shows V's). That fix is FeRx-NLME/ferx-r#357, and it does not fix this issue: once ferx-r always goes through `pk_indices`, it still hits the analytical placeholder described here.

### A — look values up by name, not by PK slot (recommended)

`build_pk_param_fn` already evaluates every top-level variable: top-level vars take var slots `0..n` first (`model_parser.rs:15772`), so an intermediate has a value at evaluation time. It just never gets copied into `PkParams` (or, for `D{n}`/`R{n}`, gets copied to a slot `pk_indices` does not record). Expose one value per `indiv_param_names` entry, for example a `CompiledModel` method `indiv_param_values(theta, eta, covariates, time) -> Vec<f64>`. Then switch `build_indiv_map` and ferx-r's `build_individual_estimates` to it.

- **For:** correct for every name on both engines, and it retires the use of `pk_indices` as a value lookup, the pattern that produced both this bug and the ferx-r ODE one.
- **Against:** it is a new public API item, so it is a design step: regenerate `api/ferx-core-public-api.txt` and say in the PR that ferx-r needs it. It must keep today's per-row `TIME` evaluation (#610) and IOV κ handling. ferx-r then needs a lock bump (`tools/update-ferx-core-lock.sh`).

### B — replace the placeholder `0` with a sentinel

Use `usize::MAX`, or a parallel `Vec<Option<usize>>`.

- **Against:** `pk_indices` is used directly as an array index on the sensitivity paths (`pk[pk_indices[i]]`, `pk_idx_f64`) and at about ten `pk_indices.get(i).copied().unwrap_or(i)` sites in the parser. A `usize::MAX` sentinel risks panics or changes those paths silently, and needs an audit of every consumer. The `Option` field avoids breakage but only lets consumers print NaN; it still gives no value for TVCL.

### C — refuse instead of returning a wrong number

In core, reject or warn on `[output]` / `[derived]` references to an analytical name that has no real slot.

- **For:** cheap, and turns a silent wrong number into a loud missing one.
- **Against:** it takes away something users can reasonably expect, echoing an intermediate. ferx-r cannot do this on its own either: from `pk_indices` alone a placeholder `0` looks exactly like the genuine CL entry. So C still needs core to expose which entries are really bound.

## Acceptance

- [ ] Tier-1 regression test on a **non-degenerate** fixture (`CL = 2·TVCL`): `[output] TVCL` returns 6.0, not 12.0. It must fail on `main`.
- [ ] Same check for a `[derived]` expression that reads `TVCL`.
- [ ] The name-based API does not expose `__ferx_ro_*` / `__ferx_pktime_*` entries (or marks them), and ferx-r `individual_estimates` has no such columns on a direct-θ/η Form-C readout fixture (proposed in the comment below).
- [ ] ODE / compartment-free control case unchanged.
- [ ] Reference value is the hand-computed closed form (`TVCL = THCL·3`); NONMEM's `$TABLE TVCL` echoes the same computed value.
- [ ] ferx-r follow-up: `individual_estimates` uses the fix; lock bumped via `tools/update-ferx-core-lock.sh`.
- [ ] `CHANGELOG.md` → `Fixed`.

Tracked in the 0.4.0 rollout epic #1350.

roninsightrx:
## Why

Closes #1356.

On an **analytical** (`pk ...`) model, an `[individual_parameters]` name that the
`[structural_model]` line does not bind has no PK slot of its own. The parser leaves
`pk_indices[i]` at the placeholder `0` (`model_parser.rs`, `unwrap_or(0)`), which is
`CL`'s slot. Every consumer that resolved a name's value as
`PkParams.values[pk_indices[i]]` therefore reported **`CL`'s value** for it:

| consumer | effect before |
|---|---|
| sdtab `[output] TVCL` | the `TVCL` column holds `CL` |
| `[derived]` expression reading `TVCL` | the expression computes with `CL` |
| ferx-r `individual_estimates` | the `TVCL` column holds `CL` |

An intermediate such as `TVCL = THCL * 3` is the common case. A modeled-dose
`D{n}` / `R{n}` is the other: its value *is* written, to a spare slot above
`PK_IDX_LAGTIME`, but `pk_indices` never records that slot, so the read still returned
`CL`.

The bundled `examples/tte_exponential.ferx` reproduces it with no contrived fixture —
`LAMBDA` comes back as `DUMMY_CL` (1.0) instead of ~0.05.

ODE and compartment-free models were never affected: `ode_param_slots` gives every name
a real slot.

## What changed

Option A from the issue — look values up **by name**, not by PK slot.

- `IndivParamProgram::eval_param_values` (new, `pub(crate)`): evaluates the compiled
  `[individual_parameters]` statements in plain `f64` and returns the evaluator's own
  var-slot prefix. `build_pk_param_fn` lays those slots out with the top-level names
  first, in `indiv_param_names` order (only if-body names are appended after), so slot
  `i` *is* name `i` — the one lookup that is correct for every name on both engines. The
  prefix length is recorded on the program (`n_top_level_vars`) rather than re-derived.
- `CompiledModel::indiv_param_values(theta, eta, covariates, time)` (new, **public**):
  one value per `indiv_param_names` entry, positional.
- `CompiledModel::indiv_param_value_map(...)` (new, **public**): the same keyed by name,
  with the parser-internal `__ferx_ro_*` / `__ferx_pktime_*` parameters (#486) filtered
  out — so a consumer no longer has to carry its own copy of that prefix list (the
  duplication the issue's comment #2 asked to retire; ferx-r currently filters them
  itself).
- The post-fit `[output]` / `[derived]` column pass now reads the map.
- `build_indiv_map` **deleted** rather than left beside the new lookup, so the
  slot-as-value idiom has no in-tree example to be copied from again.

It is the *same* statements through the *same* indexed evaluator the `pk_param_fn`
closure runs, not a second copy of the block's arithmetic — a bound name comes back
bit-identical on both routes, and `indiv_param_values_match_pk_slot_reads_for_bound_names`
asserts that with `to_bits()`. `TIME` (#610) and the extended η the closure consumes are
honoured; `[covariate_nn]` forwards are computed the same way the closure does. A
hand-built `CompiledModel` with no compiled program (test fixtures, `generate_data`)
keeps the documented slot-read fallback.

### Deliberately out of scope

Issue comment #2 lists four further consumers that use `pk_indices` as a **slot map**
for evaluation/gradients/start values rather than as a value lookup, so a value API does
not reach them:

1. `[scaling] obs_scale` Form B through an individual parameter (`build_obs_scale_spec`)
2. `[initial_conditions]` amounts (`build_init_amount_fn`)
3. analytic sensitivities, log-normal fallback (`resolve_param_derivs` /
   `seed_dim_from_slots`)
4. NCA start values (`suggest_start::find_theta_for_slot` / `find_omega_idx_for_slot`)

They are not fixed here, by agreement on scope. Worth recording why the obvious unifying
fix does **not** work: giving every analytical name a real slot drops every model with
an intermediate to finite differences, because `analytical_supported_core` gates on
`model.pk_indices.iter().all(|&s| slot_to_dim(s).is_some())` (`sens/provider.rs:593`) and
the spare region above `PK_IDX_MTT` has no dual dimension. Rows 1–4 need `pk_indices`
split into a differentiable-structural slot list and a value list — a separate design
step. I will open a follow-up issue carrying the measured repros from the comment.

## Alternatives considered

- **Option B (sentinel / `Option<usize>` in `pk_indices`)** — rejected in the issue and
  confirmed here: `pk_indices` is used directly as an array index on the sensitivity
  paths and at ~10 `unwrap_or(i)` sites, and an `Option` field only lets a consumer print
  `NaN`; it still produces no value for `TVCL`.
- **Option C (reject `[output] TVCL`)** — takes away something users reasonably expect,
  and ferx-r cannot implement it anyway: from `pk_indices` alone a placeholder `0` is
  indistinguishable from a genuine `CL` entry.
- **Give every analytical name a real slot** — see the FD-fallback regression above.

## Cross-repo dependency

ferx-r's `build_individual_estimates` should switch from
`values[pk_indices[i]]` to `indiv_param_value_map(...)` (which also lets it drop its own
`__ferx_ro_*` prefix filter). That needs a lock bump via
`ferx-r/tools/update-ferx-core-lock.sh` **after** this merges — the new `pub` items are
not reachable from ferx-r CI until then. No ferx-r change is required for this PR to be
correct on its own; the ferx-core side of the bug is fixed here.

## Breaking changes
- [x] Public Rust API — `api/ferx-core-public-api.txt` regenerated with
      `tools/update-public-api.sh`. Two items added, both needed by ferx-r's
      `individual_estimates` (and by any external caller that today reads
      `values[pk_indices[i]]`):
      `CompiledModel::indiv_param_values` and `CompiledModel::indiv_param_value_map`.
      The existing surface does not suffice: `pk_indices` is the *only* published route
      from a name to a value, and it is precisely the one that is wrong here. Nothing is
      removed or changed — sdtab column *names* are identical; the analytical
      intermediate columns now carry the right numbers.
- [ ] None

## Numerical validation

The quantity under test is the `[individual_parameters]` block's own arithmetic, so the
reference is the **hand-computed closed form**, exactly as the issue's acceptance list
specifies (`TVCL = THCL·3`) — asserted with `assert_eq!`, not a tolerance. There is no
approximation to anchor: NONMEM's `$TABLE TVCL` echoes the same assignment, which is why
the issue designates the closed form as the reference rather than a second engine. (Docker
was not running locally; happy to add a NONMEM echo run if a reviewer wants one on the
record.)

```
# fixture (non-degenerate: CL = TVCL * 2, so CL ≠ TVCL at eta = 0)
TVCL = THCL * 3        # THCL = 2.0  -> TVCL = 6.0
CL   = TVCL * 2 * exp(ETA_CL)  #     -> CL   = 12.0
V    = THV                     #     -> V    = 10.0
KA   = THKA                    #     -> KA   = 1.5

# before ([output] TVCL, [derived] TENX = TVCL * 10)
TVCL = 12.0     TENX = 120.0        <- CL's value
# after
TVCL =  6.0     TENX =  60.0        <- exact hand-computed value

# examples/tte_exponential.ferx, TVLAMBDA = 0.05, eta = 0
# before:  LAMBDA = 1.0  (= DUMMY_CL)
# after:   LAMBDA = 0.05
```

ODE control (`V, KA, KE, LAGTIME` at slots 1, 4, 0, 8) is asserted **bit-identical** to
the old slot read.

- [x] Compared against: hand-computed closed form (exact); ODE control vs. prior
      behaviour, bit-for-bit
- [ ] Not applicable

## Performance

- [x] No significant impact expected. The new evaluation runs only in the post-fit
      column pass (once per observation per subject, already an `O(n_obs)` pass that
      evaluated `pk_param_fn` at the same point). It replaces that `pk_param_fn` call
      rather than adding to it. No estimation path changed.

## Tests

11 Tier-1 tests in `src/api/tests/indiv_param_values_tests.rs` (Tier 1 so they register
patch coverage — slow/Tier-2 tests do not run on PRs).

**Non-degeneracy.** The defect is invisible under the idiomatic
`CL = TVCL * exp(ETA_CL)`: at η = 0 the intermediate equals the parameter and the wrong
read gives the right number. Every fixture keeps them apart, and each one calls
`assert_placeholder_is_live`, which asserts `pk_indices[TVCL] == PK_IDX_CL` **and** that
slot 0 holds a different value — so a fixture that stops reaching the defect fails rather
than passing vacuously.

**Mutation-tested, each side separately:**

| mutation | which tests die |
|---|---|
| revert the call site to the `pk_indices` slot read | `output_column_…` (12.0 vs 6.0) and `derived_expression_…` (120.0 vs 60.0) — the exact numbers measured on `main` |
| corrupt `eval_param_values`' returned values | 9 of 11 |
| drop the `ModelTimeGuard` | `indiv_param_values_honour_the_time_builtin` only — it is not covered for free by the others |

Coverage of what the assertions can observe: `indiv_param_values_honour_eta` exists
because a lookup that quietly evaluated at η = 0 would satisfy every other value test in
the file.

- [ ] No new tests needed

## Example
- [x] Not applicable — no new DSL surface. The behaviour change is that an existing
      `[output]` / `[derived]` name now reports its own value.

## Docs
- [x] `docs/output.qmd` — new "Individual-parameter columns" subsection under
      *Columns — sdtab*, with a callout naming the old wrong behaviour and its scope.
- [x] `docs/model-file/derived.qmd` — the *Available variables* row for
      individual-parameter names now says every name qualifies, including unbound
      intermediates and `D{n}`/`R{n}`, re-evaluated per row.
- [x] `cargo run -p docs-lint -- --check` clean (no new baseline entries).
- [x] `CHANGELOG.md` `[Unreleased]` — one `Added` bullet (the two API items) and one
      `Fixed` bullet.

## Checklist
- [x] `cargo test --lib --features ci` — 4602 passed, 0 failed
- [x] `cargo test --lib --no-default-features --features ci,markov,nn` — 4958 passed,
      0 failed
- [x] `cargo test --lib --features ci,survival` — new tests green (incl. the TTE example)
- [x] `Dual2` sensitivity path untouched — the new evaluator is `f64`-only and no
      gradient-reachable code changed
- [ ] `Cargo.toml` version bump — not needed (additive)

### ferx-site / ferx-book
- [x] No new example `.ferx` or data file

### Example execution
- [x] No example execution step needed — no user-visible DSL or CLI surface changed.
      `examples/tte_exponential.ferx` is exercised directly by a Tier-1 test
      (`bundled_tte_exponential_example_reports_lambda_not_dummy_cl`), which is what
      would have caught the reported symptom.

## Reviewer hints

- The load-bearing invariant is in `build_pk_param_fn`: `all_var_names` starts as a copy
  of `var_names` (= `indiv_param_names`) and only *appends* if-body names, so the
  evaluator's var-slot prefix is the top-level name list in order. That is what
  `n_top_level_vars` records and what makes the whole fix a one-line read. If you think
  that could ever stop holding, say so — it is the assumption everything else rests on.
- `indiv_param_values_match_pk_slot_reads_for_bound_names` is the "one implementation"
  guard: it would fail the moment the new path stopped being the same evaluator over the
  same statements.
- `PkParams` values / `pk_indices` themselves are **unchanged** — no slot was
  reallocated, so no sensitivity, dosing or prediction path can have shifted. That is
  deliberate; see *Deliberately out of scope* for why the tempting "give every name a
  slot" fix regresses the FD gate.

## Open questions

- Should the follow-up for rows 1–4 be one issue or split (the value-shaped pair
  `obs_scale` / `init` vs. the slot-map pair sensitivities / NCA)? Happy to file either
  way.
- ferx-r: I can open the companion PR (switch `build_individual_estimates` to
  `indiv_param_value_map`, drop its local prefix filter, bump the lock) once this lands —
  say the word.

https://claude.ai/code/session_01YWvSs7GspxGh3UDcPFVGuv


roninsightrx (commented):
Reviewed commit `2a013949103c577f9c665e6146ff946f27db3935`. Two blockers:

1. **[P1] Gate the TTE example regression on `survival`** — `src/api/tests/indiv_param_values_tests.rs:440-447`

   The core CI job runs without the `survival` feature, but this unconditional test parses `examples/tte_exponential.ferx`, whose `[event_model]` is feature-gated. It therefore fails before exercising the regression:

   ```
   cargo test --lib indiv_param_values_tests --features ci
   # 10 passed, 1 failed
   # bundled example must parse: Block `[event_model]` ... requires `--features survival`
   ```

   This is also the current `Tests + coverage (core)` failure. The same test passes with `--features ci,survival`. Please add `#[cfg(feature = "survival")]` to this test or use a survival-free fixture in the core matrix.

2. **[P2] Make the public lookup able to select `MIXNUM`** — `src/types.rs:4777-4783` and `4831-4837`

   `indiv_param_values` evaluates the compiled statements under the ambient mixture-class thread-local, but external callers cannot set it: `MixtureClassGuard` is `pub(crate)` and the thread-local defaults to class 1. Thus for a model such as `CL = if (MIXNUM == 1) TVCL1 else TVCL2`, every direct external call returns class-1 values, including for a `SubjectResult` whose `mixest` is 2. The internal sdtab path is correct only because `compute_extra_output_columns` installs the private guard before calling this method.

   This matters directly to the stated ferx-r follow-up: `build_individual_estimates` has `sr.mixest`, but switching it to `indiv_param_value_map(...)` provides no way to pass that class and will keep reporting class-1 individual estimates for class-2 subjects. Please expose a class-aware variant/argument (with validation and a class-2 regression test), or otherwise provide a public way to evaluate the map under a selected class.

The evaluator-slot approach itself looks sound for the non-mixture analytical, ODE, TIME, eta, modeled-duration, synthetic-name, and hand-built fallback cases covered here. I also verified `git diff --check` is clean and the focused survival-enabled TTE test passes.

roninsightrx:
Both blockers fixed. `1ddeac6a` (P1) and `4a2c1699` (P2).

## P1 — TTE example test gated on `survival`

Correct, and it was the `Tests + coverage (core)` failure. My error: I added that test late and only ran it under `--features ci,survival`, so I never saw it. `[event_model]` is refused at *parse* time without the feature, so the gate has to sit on the test rather than inside it — `#[cfg(feature = "survival")]`, with a comment saying which job the arm runs in (`Tests + coverage (TTE/CTMM endpoints)`).

Re-ran both sides this time: `--features ci` → 4608 passed / 0 failed; `--features ci,survival` → 4796 passed / 0 failed.

## P2 — `MIXNUM` class selection

Real, and worse than a missing convenience: it would have silently broken the ferx-r follow-up in exactly the way this PR exists to prevent. Fixed by putting the class in the signature rather than adding a variant:

```rust
pub fn indiv_param_values(&self, theta, eta, covariates, time, mixture_class: Option<usize>) -> Vec<f64>
pub fn indiv_param_value_map(&self, theta, eta, covariates, time, mixture_class: Option<usize>) -> HashMap<String, f64>
```

Changing the signature rather than adding `*_for_class` siblings because both items are unreleased (they are in `[Unreleased]`, nothing outside this branch calls them) and because a caller who *cannot* skip the parameter is the point — a mixture model that silently reads class 1 is the defect. 1-based to match `MIXNUM`; `Some(mixest + 1)` is the documented call for a fitted subject.

**Validation, as asked.** `Some(k)` with `k == 0` or `k > n_classes` (1 when the model declares no `[mixture]` block) panics with an actionable message. Not a silent clamp: an out-of-range class matches no `MIXNUM` arm, so it would return whichever `else` branch happens to be last — a silently wrong number, which is the whole class of defect this PR removes. Panic rather than `Result` because every in-tree caller derives the class from a fit where it cannot be out of range, and a `Result` on a value lookup would push an unwrap into the post-fit column loop. `# Panics` is documented on both methods; three `#[should_panic]` tests cover class 0, class past the end, and a class passed to a non-mixture model.

The post-fit column pass now passes the subject's own class explicitly instead of relying on the ambient guard.

### Regression tests

Fixture is a 2-class mixture whose class-specific typical value is reached *through* an unbound intermediate, so the class selection and the #1356 lookup are exercised together — `TVCL = if (MIXNUM == 1) TVCL1 else TVCL2` with `CL = TVCL * 5`, so `TVCL` stays non-degenerate against `CL` and `assert_placeholder_is_live` still applies.

| test | pins |
|---|---|
| `mixture_class_selects_the_class_specific_typical_value` | class 1 → `TVCL = 1`, `CL = 5`; class 2 → `TVCL = 3`, `CL = 15`; and that `None` is the class-1 ambient default |
| `mixture_class_guard_is_restored_after_the_call` | a class-2 lookup does not leave `MIXNUM` at 2 for the next caller on the thread |
| `output_column_uses_the_subjects_own_mixture_class` | end-to-end sdtab `[output]` for a class-2 subject |
| 3 × `#[should_panic]` | class 0, class 3 of 2, class 2 on a non-mixture model |

### One thing your review made me measure, worth flagging

I mutation-tested the fix by making `indiv_param_values` ignore `mixture_class`. Only the **unit** test dies. The end-to-end `output_column_uses_the_subjects_own_mixture_class` stays green — because `compute_extra_output_columns` also installs the ambient `MixtureClassGuard` for the subject (it has to, for the `pk_param_fn` and `[derived]` evaluations in the same loop). So the explicit argument and the ambient guard are two gates rejecting the same inputs, and removing either alone leaves that test green.

I kept both — the ambient guard is not part of the public contract, so an external caller has no way to set it, and that is precisely P2 — but the test now says at its definition that it cannot fail from the class argument alone and names which test is the argument's real gate. Better a documented redundancy than a test implying a strength it does not have.

### Also updated

`api/ferx-core-public-api.txt` regenerated (signature change, still exactly the two added items), `CHANGELOG.md` `Added` bullet now mentions the class argument.

`tools/preflight.sh` green.

Issue #490, opened 2026-06-24
create "auto" optimizer choice, and set as default


roninsightrx:
With limited benchmarking (FOCEI) of optimizers on ~10 real datasets with varying complexity, the following conclusions can be drawn:
- on problems where analytic gradients can be computed using sensitivity equations, the NLopt L-BFGS optimizer is performing best on every tested dataset. Only bobyqa is faster, but most often not finding the optimum (but something close). SLSQP usually converges OK, but much slower than L-BFGS.
- on problems where FD have to be used ([list](https://github.com/FeRx-NLME/ferx-core/issues/486)), `bobyqa` is fastest by far, and reasonably accurate. L-BFGS / SLSQP / MMA are very slow and often not able to find the right basin.

So proposal to have an "auto" argument (which is default) to the `optimizer` setting. This would use "nlopt_lbfgs" for problems where analytic gradients are available, and "bobyqa" when only FD is available. 

The various choices can be changed over time as we improve the various paths and investigate more datasets, but this should provide reasonable defaults for most users.

TeunP:
Heads-up from #501 (competing-risks TTE): merging current `main` surfaced that the new `optimizer = auto` default **silently stalls TTE (`[event_model]`) fits**.

**Symptom.** With `auto` (now the default) a TTE fit returns a finite OFV but the parameters stay at their initial estimates — no error, just a non-converged result. All four `tte_sse_*` recovery tests stalled at init against current `main` (e.g. exponential SSE: `λ=0.05` init instead of recovering `0.10`).

**Root cause.** `auto` resolves via `sens::provider::analytic_outer_gradient_available`, which returned `true` for TTE models — `gradient_method` defaults to `Auto`, and a TTE-only model's trivial structural block is `sens_supported`. So `resolve_auto` picked gradient-based `NloptLbfgs`, but the hazard log-likelihood has **no analytic outer gradient** (the provider only covers the structural PK/PD model), so the optimizer gets a meaningless gradient and never steps. TTE is documented FD-only, so this path should never be gradient-based.

**Fix (in #501).** One-line guard in the shared predicate:

```rust
!matches!(model.gradient_method, GradientMethod::Fd)
    && !model.has_tte()          // hazard log-lik has no analytic outer gradient
    && (sens_supported(model) || iov_analytical_supported(model))
```

[...]

PR #1380, opened 2026-09-13
feat(estimation): report a gradient at the solution for derivative-free fits, and warn on a fit that never moved (#997 §1, §2)

roninsightrx:
## Why

#997 bundles four defects, all of the same shape: the fit looks fine and the output does not say
otherwise. This PR takes the two that are self-contained and leaves the other two to their own
issues (triage below).

**§1 — `converged = TRUE` is unverifiable under a derivative-free optimizer.** `final_gradient` was
populated for the gradient-based NLopt optimizers and `NULL` for BOBYQA — absent for exactly the
optimizer where a premature stop is most likely, and where there is therefore nothing in the fit
object to check the convergence claim against. The reporter measured two arms of the same model
1.54 OFV apart, both reporting success, one of them with no gradient; widening the parameter box a
thousandfold returned a bit-identical objective to six decimals, because the fit had never travelled
far enough for a bound to matter.

**§2 — a fit that never left its initial estimates says nothing about the model, and did not say so.**
The `stalled_at_init` predicate has existed since #751, but nothing surfaced it: it was reachable
only by a caller who thought to ask. Such a fit usually reports `converged: true` — having never
moved, it has a perfectly flat objective trace to plateau on — so nothing else in the output
distinguishes it from a good fit.

### Triage of the four items

| item | disposition |
|---|---|
| §1 FD gradient at the solution | **this PR** |
| §2 stall-at-init warning | **this PR**, but not as written — see below |
| §3 `ode_method` defaults to `Rk45` on stiff systems | **not here.** The issue author marked it superseded by **#978** (a-priori stiffness probe + `ode_method = auto`) and posted the anchor numbers there. It is also a global default change, which wants its own discussion |
| §4 `optimizer = auto` couples to `gradient`, silently | **not here.** The *reporting* half already shipped in **#490**: `FitResult.optimizer` is already the compound `"auto (nlopt_lbfgs)"`. What remains is decoupling the two knobs, a design change — own issue |
| §5 covariance estimator is unlabelled | **not here.** Real gap — `FitResult` carries `cov_condition_number` and `cov_eigenvalues` but no estimator label — but a different subsystem. Own issue |

### §2 is deliberately *not* the check the issue asked for

#997 §2 proposed warning on `theta / theta_init ≈ 1` per parameter. The author withdrew that himself
in a follow-up comment, having measured it: on his own anchor model the data were simulated at
`KON = 1000` and started there, so the **correct** arm left `KON` at 0.999 of its initial value and
a ratio test fires hardest on the arm that is right. Any init-ratio heuristic penalises a well-chosen
start.

The predicate used here is not that test. `crate::stalled_at_init` asks whether **any** free
coordinate moved, over θ, Ω and σ together, and prefers the optimizer's own scaled-space escape
verdict (`left_init`, #751) when the run recorded one. A well-chosen start that one parameter stays at
is not a stall; a fit where nothing at all moved is, whatever the start was worth.

## What changed

**§1.** `estimation/outer_optimizer.rs` gains `reporting_fd_gradient`, and `optimize_nlopt_once`
calls it at the restored best point when `last_gradient` is empty. New `FitResult` field
`final_gradient_source` (`"optimizer"` / `"finite_difference"`), and new `[fit_options]` key
`report_final_gradient` (default `true`) to opt out of the cost.

It is a second function rather than a flag on `reconverged_fd_gradient`, and the asymmetry is
recorded at both sites: that one is a gradient *fed to the optimizer*, so it differentiates the
likelihood alone and the caller splices the analytic NN-penalty and prior gradients in afterwards.
Here there is no caller to splice anything — the quantity that has to come out is `∇(OFV + penalty)`,
matching `FitResult::final_gradient`'s documented contract — so the penalties go inside the stencil.
The EBEs are re-solved (warm-started) at every perturbed point, like the objective closure itself,
and the mixture branch differentiates the K-fold log-sum-exp rather than a single-class stand-in.

It never steers anything: computed once, after the fit, so the estimates are bit-identical whether it
runs or not. Cost `2·n_free` objective evaluations — one gradient's worth.

**§2.** New `WarningCode::StalledAtInit` + `api::postfit::stalled_at_init_warning`, emitted
post-construction next to `high_correlation_warning` (both read the assembled `FitResult`). It does
**not** demote `converged`, unlike `runaway_guard_warning`: a runaway hit is by construction not an
interior optimum, whereas a stall at the start can be one (a fit restarted from its own output). An
evaluation-only run (`outer_maxiter = 0`, NONMEM `MAXEVAL=0`) is exempt — it reports the objective at
the initial estimates *by request*, and the warning would tell the caller nothing they did not ask
for.

## Alternatives considered

- **Hold the EBEs fixed in the reporting stencil** (cheaper, and directly comparable to the fixed-EBE
  analytic gradient the non-IOV optimizers report). Rejected: the documented contract for
  `final_gradient` is the gradient of the objective the fit minimised, and the marginal objective's
  EBEs move with the parameters. Re-solving is also what `reconverged_fd_gradient` — the codebase's
  own reference for the analytic path, via `FERX_SENS_CHECK=1` — does.
- **Demote `converged` on a stall**, as the runaway guard does. Rejected above; `Strictness::reject_init_stall` already exists for search harnesses that want it fatal.
- **Skip the reporting gradient for mixture models** rather than carry a second branch. Rejected —
  a `[mixture]` FOCEI fit routes to BOBYQA like any other, so it is precisely a case that needs one.

## Cross-repo dependency

`ferx-r/src/rust/src/lib.rs` builds the R fit object from an **explicit** field list, so
`final_gradient_source` will not reach R users until a matching ferx-r PR adds it next to
`final_gradient` (lib.rs:2656) and maps `report_final_gradient` in `settings =`. Nothing breaks in the
meantime — ferx-r CI builds from the ferx-core commit pinned in its `Cargo.lock`, which this PR does
not move. The R-side `fit$final_gradient` **does** start being populated for BOBYQA runs as soon as
that lock is bumped, which is the fix the issue asked for; the `_source` label is the part that needs
the wrapper change.

## Breaking changes
- [ ] None

Baseline regenerated in this PR. Four added items, each needed by an external caller:

| added | needed by |
|---|---|
| `FitResult::final_gradient_source` | ferx-r, and any agent reading a `.fitrx` / `--output-format json` bundle: the field is the whole point of §1 — without it a `final_gradient` cannot be told from an optimizer's own. |
| `FitOptions::report_final_gradient` | ferx-r `settings =` and the `[fit_options]` parser; the opt-out for models whose every objective evaluation is seconds. |
| `WarningCode::StalledAtInit` | ferx-r and agents branch on `warnings_structured[].category`; the whole warning-code enum is public for that reason. |
| `OuterResult::final_gradient_source` | not external surface by intent — `OuterResult`'s fields are all already `pub` and it is re-exported, so the field follows the struct. It is lifted onto `FitResult` and read nowhere else. |

No field was removed or retyped; both new `FitResult`/`FitOptions` fields are additive and
`#[serde(default)]` / defaulted, so an older `.fitrx` bundle loads unchanged (test:
`fit_wire_missing_final_gradient_source_defaults_to_none`). The `FitResult` literal *is* breaking for
an out-of-tree constructor, which is why the box above is ticked.

## Numerical validation

- [x] Compared against: prior ferx commit (`fcfe8b10`) — both arms, same model and data
- [x] Reference values:

No NONMEM anchor: the quantity is ferx's gradient of ferx's own objective in ferx's packed space
(log-θ, Cholesky-Ω, log-σ). NONMEM prints a `GRADIENT` line per iteration, but in its own scaled
parameterization and only for its own optimizer, so there is no run to compare against. The
validations that do apply are (a) parity against the established in-tree definition, (b) agreement
with an independent implementation on the mixture branch, and (c) the empirical check that it goes to
zero at a known optimum:

```
# (c) examples/warfarin.ferx + data/warfarin.csv, method = focei, maxiter = 300
#     same model, same start, two optimizers

optimizer     converged   OFV          final_gradient_source   |grad|inf
nlopt_lbfgs   true        -286.0042    optimizer               2.29e-4
bobyqa        true        -285.9866    finite_difference       2.47e0

# the two rows use different stencils, so the *same* FD stencil was also run
# at the L-BFGS solution:  3.04e-4 (against the optimizer's own 2.29e-4).
# So the reporting gradient does go to zero at a genuine optimum on this model;
# the 2.47 is BOBYQA's, not the stencil's.

# step-size check (|grad|inf vs h, per packed coordinate h = hscale*(1+|x|)):
#   hscale     1e-4      1e-3      1e-2      3e-2      1e-1
#   bobyqa     2.473e0   2.473e0   2.473e0   5.132e0   3.760e1
#   lbfgs      3.039e-4  3.244e-3  3.287e-1  2.974e0   3.494e1
# the shipped 1e-4 sits in the flat part of the curve; 1e-1 is truncation error.
```

`ferx_version`, estimates and OFV for the L-BFGS arm are unchanged from `fcfe8b10` — the reporting
gradient runs only when the optimizer supplied none, and even then after the fit is over.

## Performance
- [ ] Faster
- [ ] Slower

Gradient-based fits are untouched (`last_gradient` is `Some`, so the new branch is not taken). A
derivative-free fit pays `2·n_free` extra objective evaluations once, at the end — one gradient's
worth, against the `outer_maxiter·(n+1) + 40·(n+1)` BOBYQA's own budget allows. On warfarin
(n_free = 7, 300 maxiter) that is 14 evaluations on a fit that made ~300; wall time went 0.067 s →
0.067 s, below measurement noise. `report_final_gradient = false` removes it entirely for a model
where an objective evaluation is seconds.

## Tests
- [ ] No new tests needed

`cargo test --lib --features ci,survival`: 4797 passed, 0 failed.

**§1** (`src/estimation/outer_optimizer_tests.rs`, Tier 1):

- `reporting_fd_gradient_equals_the_reconverged_one_when_nothing_is_penalized` — **bit-equality**
  against `reconverged_fd_gradient` on an unpenalized model. Not a test of the stencil (they share
  `central_diff_packed`) but of the objective inside it, with a non-degeneracy precondition so
  `vec![0.0; n]` cannot pass.
- `reporting_fd_gradient_differentiates_the_prior_penalty_too` — pins that the penalty is differenced
  *inside* the stencil, against the analytic `2(x−m)/s²` the optimizer's own path splices in. Worst
  realised error 4.03e-9 against a 1e-7 bound (25×), with the priored coordinate's true value ≈ 33.
- `reporting_fd_gradient_differences_the_mixture_objective` — against
  `mixture::mixture_gradient_fd`, an **independent** central difference of `mixture_ofv` with its own
  step and no bounds clamping. Worst realised 1.18e-7 against a 1e-5 bound. Forcing the single-class
  branch leaves θ/Ω within 1.2e-7 and moves the **mixing logit to 8.0e-1** — so the fixture declares a
  *free* `logit(1) = MIXL` on purpose; one with the mixing probability FIXed would pass with the branch
  deleted.
- `a_derivative_free_fit_reports_a_finite_difference_gradient_at_its_own_estimates` — end to end, and
  the point assertion is the substantive half: the reported vector must equal the stencil recomputed at
  `packed_estimate`. Realised difference exactly 0.0; bound is a relative 1e-12, against 1.2e3–2.5e5
  for the same gradient taken at the *initial* estimates instead.
- `a_gradient_based_fit_reports_the_optimizers_own_gradient` and
  `report_final_gradient_false_leaves_a_derivative_free_fit_with_no_gradient`.
- `io/fitrx.rs`: round trip, and an old bundle with the key absent loading as `None`.

**§2** (`src/api/tests/stalled_at_init_warning_tests.rs`, Tier 1, plus two end-to-end in
`outer_optimizer_tests.rs`): category and token; re-classification landing on `StalledAtInit` and
**not** `Convergence`; both verdict sources including the case where the optimizer's verdict
*overrides* a natural-scale stall; the details payload skipping FIXed θ; the `outer_maxiter = 0`
exemption and its scope; and the wiring, with a negative arm on the same model.

The stall fixture is worth a note. `outer_maxiter: 1` does **not** produce a stall — BOBYQA's budget is
`maxiter·(n+1) + 40·(n+1)`, so one "iteration" buys it 205 evaluations and it moves freely (this was
caught by the test's own precondition assert, not by reading the code). The fixture instead pins the
*objective*: FIX Ω and Σ, and put a prior on each free θ centred on its own initial value at 0.01 %
RSE.

**Mutation-checked**, one mutation per new assertion, all 10 killed:

| mutation | test that died |
|---|---|
| drop the stall-warning emit site in `api::fit` | `a_fit_that_never_left_…` |
| warn unconditionally (drop the predicate gate) | 3 tests |
| drop the `outer_maxiter == 0` exemption | `an_evaluation_only_run_…` |
| classify the message as `Convergence` | `the_message_reclassifies_…` |
| drop `priors.penalty(xv)` from the stencil | `…differentiates_the_prior_penalty_too` |
| drop the `2.0 *` on the reporting objective | `…equals_the_reconverged_one_…` |
| take the gradient at the initial estimates | `a_derivative_free_fit_reports_…` |
| label every gradient `finite_difference` | `a_gradient_based_fit_reports_…` |
| ignore `report_final_gradient = false` | `report_final_gradient_false_…` |
| ignore the optimizer's own `left_init` verdict | `the_details_name_which_reading_…` |
| force the single-class branch for a mixture | `…differences_the_mixture_objective` |

- [ ] Example added to `examples/`
- [x] Not applicable — no new `.ferx` syntax; `report_final_gradient` is one `[fit_options]` key,
      documented with the rest, and the default behaviour needs no opt-in:

```toml
[fit_options]
  method     = focei
  optimizer  = bobyqa
  # report_final_gradient = false   # skip the 2*n_free reporting evaluations
```

```
# fit$final_gradient_source
"finite_difference"
# fit$final_gradient  ->  max |.| = 2.47   (vs 2.29e-4 for the L-BFGS arm)
```

## Docs
- [x] New pages linked in `docs/_quarto.yml` — no new pages
- [ ] No user-visible change

- `docs/model-file/fit-options.qmd` — `report_final_gradient`. The `### Outer optimizer` section went
  over the docs-lint R1 5,000-character limit, so it is now split into `#### Algorithm and stopping
  rules` / `#### Reporting` / `#### Global pre-search`. The `{#outer-optimizer}` id is unchanged, so
  existing links still resolve.
- `docs/estimation/optimizers.qmd` — new `### Checking a derivative-free converged`, carrying the
  measured table and step-size sweep above, plus the caveat that the number is comparative (packed
  space, unnormalised) and not a threshold test. Also corrects the neighbouring claim that a fit
  failing to leave its initial estimates "is reported with `converged = false`" — true of the
  deferred-`Failure` path, not of a derivative-free optimizer that reports its own success at the
  start, which is what the new warning covers.
- `docs/warnings.qmd` — `stalled_at_init` in both the code table and the `details` table.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — not touched; the new code is an
      FD stencil over the existing `f64` objective and adds no generic instantiation.

### ferx-site
- [x] New `[fit_options]` key → `report_final_gradient` documented in `docs/model-file/fit-options.qmd`; ferx-site page follow-up alongside the ferx-r PR
- [x] No new data file

### ferx-book
- [x] No new estimator or DSL feature; the optimizer chapter's material lives in `docs/estimation/optimizers.qmd`, updated here

### Example execution
- [ ] Rebuilt ferx-r against updated ferx-core — the local `../ferx-r` checkout predates several
      `FitResult` fields already on main and does not compile against it; the wrapper change is its own
      PR (cross-repo table above)
- [x] Affected example runs cleanly: `./target/release/ferx examples/warfarin.ferx --data data/warfarin.csv` under both optimizers, output above
- [x] No ferx-site / ferx-book page render needed

## Reviewer hints

Focus on `reporting_fd_gradient` and its call site in `optimize_nlopt_once`. Three things worth
checking:

1. **The point.** The gradient is taken at `x0` *after* the best-seen restore and the unscale, i.e. the
   same vector the reported estimates and the covariance step are built from — not the last point NLopt
   happened to probe. A gradient at the wrong point would be finite and plausible and no `is_some`
   check could tell, which is why there is a dedicated assertion against `packed_estimate`.
2. **The objective.** It has to be the *penalized* one (`FitResult::final_gradient`'s contract), which
   is the one thing that differs from `reconverged_fd_gradient` and the reason for a second function.
3. **The mixture branch**, which mirrors the objective closure's own two arms.

Skimmable: the mechanical `final_gradient_source: None` additions across the 12 `OuterResult`
construction sites, and the `ChunkKind`-driven docs split.

## Open questions

- Should `report_final_gradient` default to `false` for ODE models specifically, where `2·n_free`
  evaluations is real wall time? I left it `true` everywhere — the derivative-free optimizer *is* the
  `auto` pick for ODE models, so that is exactly the population §1 is about, and the cost is bounded
  by one gradient — but it is a defensible place to differ.
- §5 (label the covariance estimator) is a genuinely cheap fix and I can fold it in here rather than
  filing it, if you would rather not have a second PR.

https://claude.ai/code/session_01NBZfPZpQARragXGiFWLxbE


roninsightrx (commented):
Reviewed commit 5016b93e.

I found one correctness issue that should be fixed before merge:

- **[P2] Do not retain the GN-phase gradient when an accepted polish has none** (`src/estimation/gauss_newton.rs:595-603`). With `method = gn_hybrid`, `optimizer = bobyqa`, and `report_final_gradient = false`, an improving FOCEI polish becomes the reported fit, but this conditional leaves `final_gradient` set to the GN-phase gradient whenever `polish_result.final_gradient` is `None`. Lines 662-663 then label that stale vector as `"optimizer"`, even though it was evaluated at the pre-polish GN point rather than at the reported estimates. That violates the documented `final_gradient` contract and can give a misleading stationarity diagnostic. When the polish wins, please transfer both gradient fields unconditionally (so both become `None` when reporting was opted out), or recompute at the accepted polish point. A regression test for this configuration would pin the invariant.

Non-blocking test gap: the new `[fit_options] report_final_gradient` parser arm is only covered by generic key-reachability and the programmatic `FitOptions` test. Per `AGENTS.md`, please add a parser-level test showing the DSL value `false` reaches `FitOptions::report_final_gradient` (and ideally that an invalid boolean is rejected).

Validation: all GitHub checks are green. Locally, the three `reporting_fd_gradient` tests and seven `stalled_at_init_warning` tests pass under `--features ci,survival`.

roninsightrx:
Both addressed in `bcfe8f79`. Thanks — the P2 is real, and worse than "stale under an opt-out": the stale-gradient half **predates this PR**. The merge was `final_gradient = polish_result.final_gradient.or(final_gradient)`, so `gn_hybrid` + a derivative-free `optimizer` (the `auto` pick on ODE/PD models) or the built-in BFGS has always reported the GN-phase vector against post-polish estimates. What this PR added was the label that makes it legible — and misleading. It now gets a `### Fixed` changelog entry of its own rather than riding along in the feature entry.

**[P2] fixed** — but not by transferring two fields unconditionally side by side, because that spelling is what let them drift in the first place. The gradient and its source are now **one value**, `Option<(Vec<f64>, String)>`, `zip`ped at the merge and `unzip`ped at each `OuterResult` literal. The pairing invariant `final_gradient.is_some() == final_gradient_source.is_some()` is then structural instead of restated at every construction site, and an accepted polish transfers both or neither:

```rust
final_gradient = polish_result
    .final_gradient
    .zip(polish_result.final_gradient_source);
```

I took the "transfer" option over "recompute at the accepted polish point": recomputing would silently spend `2·n_free` objective evaluations that the user disabled with `report_final_gradient = false`, which is the setting that reaches this branch in the first place. `None` is the honest answer when the accepted phase computed nothing.

Also tightened `FitResult::final_gradient`'s doc comment to state the "at the estimates reported beside it" half explicitly, calling out `gn_hybrid` as the multi-stage case that can break it quietly.

**Regression tests** (`estimation::gauss_newton::tests`), three arms:

- `an_accepted_polish_with_no_gradient_reports_none_not_the_gn_phase_vector` — your configuration exactly.
- `an_accepted_polish_reports_its_own_gradient_under_its_own_label` — the other half. Without it, "transfer both fields" is satisfied by an implementation that hard-codes `"optimizer"`; a derivative-free polish must report `"finite_difference"`.
- `the_gn_phase_reports_its_own_gradient_under_the_optimizer_label` — pins the initializer, i.e. that the fix did not overcorrect into discarding a good gradient.

Both accepted-polish tests establish the precondition independently of the fields under test: a pure-GN control run on the same model, data and budget, asserting `hybrid.ofv < pure.ofv` (so the polish really was accepted) *and* `pure.final_gradient.is_some()` (so the GN phase had a gradient to leak, and `None` is the transfer working rather than a phase with nothing to give).

There is deliberately **no** rejected-polish test, and the reason is recorded at the test rather than left implicit: after the fix that arm contains no gradient code at all — it neither reads nor writes the pair — so there is nothing there to mutate. It is also unreachable on the fixture: the polish improves 57706.3 → 28.6 from every start I tried, and the polish budget is hard-coded to 100 evaluations inside `run_foce_gn`, so it cannot be starved into losing.

**Non-blocking test gap: fixed.** `test_parse_report_final_gradient` in `parser::model_parser::tests` covers the DSL `false` reaching `FitOptions`, an explicit `true` round-tripping, every `parse_bool` spelling (`t`/`yes`/`1`/`on`, `f`/`no`/`0`/`off`) so the key is not quietly stricter than its neighbours, and a non-boolean being **refused** with an error naming both the key and the expected values — a typo must not fall through to the default, since that is the same silent-cost failure as a dropped arm.

**One more, unprompted:** codecov's single uncovered line was `WarningCode::StalledAtInit => "stalled_at_init"`. `warning_code_tokens_are_stable` is a hand-maintained list, and it had already fallen four variants behind before mine — `FlipFlop`, `AbsorptionTwinDeclined`, `FlatParameter` and `OdeSolver` each shipped with no row, so the test whose entire job is pinning serde tokens was not pinning theirs. All five added, plus a wildcard-free `match` over every variant so a future `WarningCode` fails to compile here until someone updates the list.

**Verification.** Every new assertion mutation-checked, one mutation each, all six killed:

| mutation | test that died |
|---|---|
| restore the old `.or(final_gradient)` merge | `an_accepted_polish_with_no_gradient_…` |
| hard-code `"optimizer"` on an accepted polish | `an_accepted_polish_reports_its_own_…` |
| relabel the GN phase's own gradient | `the_gn_phase_reports_its_own_gradient_…` |
| drop the `report_final_gradient` parser arm | `test_parse_report_final_gradient` |
| `unwrap_or(true)` instead of `?` on the bool | `test_parse_report_final_gradient` |
| drop `StalledAtInit` from the token list | `warning_code_tokens_are_stable` |

`tools/preflight.sh` green (`fmt check clippy rustdoc docs public-api`); `cargo test --lib --features ci,survival` 4801 passed, 0 failed. No public-API change beyond the baseline already in this PR.

PR #1383, opened 2026-09-13
fix(api): never report converged at a non-finite objective [closes #1303]

### LINKED ISSUE #1303: bug(api): fit() returns converged: true with ofv = NaN
Measured on a two-subject FOCEI fit where one subject's timeline cannot be ordered (a `NaN`
dose time), at `674cdcf8`:

```
ofv       = NaN
converged = true
```

There is no production gate anywhere in `src/api` or `src/estimation` tying `converged` to a
finite objective. A consumer branching on `converged` — the R wrapper, a search tool in
`ferx-tools`, an agent reading the fit YAML — is told the run succeeded.

That is worse than a missing diagnostic. #1296 adds a warning naming the abandoned walks, but
the boolean a caller is most likely to key on still says the fit is good, and a warning is
easier to ignore than a flag is to misread.

**Not covered by #1295.** That issue is scoped to the jet fill. Its measured side effect flips
`converged` to `false` only because the objective becomes the `2e20` non-finite sentinel instead
of `NaN` — an accident of which sentinel the outer objective happens to use, not a gate. Any
other route to a `NaN` OFV still reports `converged: true`.

**Suggested shape:** `converged` is false whenever the reported objective is not finite,
enforced wherever it is set rather than at one call site, with a warning saying why. Open
question worth deciding in the issue: whether a non-finite OFV *at the final estimates* should
be an error rather than a warning — #1235's measurement (one bad subject in ten destroys the
objective for everyone) argues it leaves no partial result worth returning.

Found while reviewing #1296.

roninsightrx:
## Why

`fit()` returned `converged: true` alongside `ofv: NaN`. Reproduced at `fcfe8b10` on the
two-subject FOCEI fit one of whose subjects carries a `NaN` dose time (#1296's fixture):

```
Eval    1: OFV = 1000000000088.716797
Eval    3: OFV = 1000000000000.000000
NLopt finished: Success
Restored best-seen point (OFV = 1000000000000.000000) for final inner loop and covariance step.
Final OFV = NaN

ofv       = NaN
converged = true
n_iter    = 4
```

That trace is the whole mechanism: the optimizer's stop rule reports on **its own trace**
(which bottoms out at a finite `1e12` and reports `Success`), while the objective actually
published is recomputed by the final inner loop at the restored best point — and comes back
`NaN`. Nothing in `src/api` or `src/estimation` tied the two together, so the two could
disagree and the boolean won.

`converged` is the field a consumer is most likely to key on: the R wrapper, `ferx-tools`'
model-space search via `Strictness::require_converged`, an agent reading the fit YAML. #1296
had already added a warning naming the abandoned walks, but a warning is easier to ignore
than a flag is to misread — and #1235's measurement is that one bad subject in ten destroys
the objective for everyone, so the healthy subjects' numbers are meaningless too.

**Not covered by #1295.** That issue is scoped to the jet fill, and its side effect flips
`converged` only because the objective becomes the `2e20` sentinel instead of `NaN` — an
accident of which sentinel the objective happens to use, not a gate.

## What changed

**One rule, one owner.** `estimation::outer_optimizer::gate_converged_on_objective(&mut bool, f64)`
demotes the verdict and returns the warning saying why. It reuses `ofv_is_valid`
(`is_finite() && < DIVERGENCE_OFV`), the predicate the multi-start ranking already applies.

**`is_finite()` is deliberately not the test** — the sentinel trap the issue names. The inner
objective clamps a blown-up individual contribution to a finite `~1e20` rail and the outer
objective doubles it, so a *repelled* fit reports `2e20` and sails through a finiteness check.
`the_gate_rejects_the_finite_sentinel_not_only_nan` asserts `2e20_f64.is_finite()` before
asserting the gate rejects it, so the trap is stated rather than assumed.

**Applied wherever the pair is published, not at one call site.** `OuterResult` is **public
API with public fields**, and `ferx-tools` calls `optimize_population` / `run_foce_gn` /
`run_imp` / `run_impmap` / `run_bayes` directly without going near `fit_inner`, so a
`fit()`-only patch would leave that surface wrong. Twelve sites:

| file | sites |
|---|---|
| `estimation/outer_optimizer.rs` | NLopt exit, built-in BFGS exit (the eval-only exit is `converged: false` literal) |
| `estimation/trust_region.rs` | its existing `ofv_is_valid` demotion, moved onto the shared gate |
| `estimation/gauss_newton.rs` | pure-GN exit and the FOCEI-polish hybrid exit |
| `estimation/impmap.rs`, `estimation/saem.rs` | the three MCEM exits (see below) |
| `estimation/bayes.rs` | R-hat is a *sampling* diagnostic and independent of the objective; both must hold |
| `estimation/vi/run.rs` | gated, with the one exemption below |
| `api/fit.rs` | the two synthetic evaluation-only `OuterResult`s (AGQ readout, standalone `imp_eval_only`) and the `FitResult` assembly |

**The `fit()`-level gate is a different number, not a second copy.** Each estimator gates
`OuterResult::ofv` — the **clean** −2 log L, with the prior penalty excluded by construction.
`FitResult::ofv` is `ofv_data + ofv_prior`, and `ofv_prior` (#254) is evaluated in `fit_inner`
*after* the last optimizer returned. So a finite likelihood plus an overflowing penalty
publishes an objective no estimator ever saw. G5 is that case, end to end.

**The stage passed to the `fit()` gate is `chain.last()`, not `final_method`.** `final_method`
is the last *estimating* stage — right for "who produced the parameters", wrong for "who
produced the objective": under `methods = vi, imp` + `imp_eval_only` the reported OFV is IMP's
genuine importance-sampled −2 log L while `final_method` is still `vi`, so keying the VI
exemption off it would exempt a real objective from the gate.

**The warning.** `W_NONFINITE_OBJECTIVE`, severity `Critical`, category `convergence` — the
same verdict `W_VI_BAD_BASIN` and the runaway-guard hit get, and no new `WarningCode` variant
(so no public-API widening). It names *which* of the three failures happened (`NaN` /
`infinite` / `at the divergence sentinel`), because the remediation differs, and carries a
`details` payload (`ofv`, `reason`, `divergence_cutoff`, `method`). `classify_warning` gets a
**token** arm ahead of every prose arm, following #1080's precedent: the message quotes the
offending value and lists candidate causes ("infusion duration", "residual variance"), any of
which a later edit could grow into a phrase a prose arm below claims.

**Exactly one message, not two.** The estimator's gate and `fit()`'s gate both apply on the
same run; the shared gate's `!*converged` early return is what stops the second from
re-explaining a verdict the first already demoted. G1 asserts the count is `1`, not `>= 1`.

**MCEM (#528) now says why.** IMP/IMPMAP/SAEM already refused to call a runaway converged —
`objective_converged(ofv)` — but demoted **silently**: a caller saw `converged: false` with
nothing in `warnings` distinguishing it from any other failure. `gate_converged_on_mcem_objective`
composes both rules and emits the reason. It is **not** a redundant second gate, and the
witnesses are measured and disjoint:

| `ofv` | `ofv_is_valid` (#1303) | `objective_converged` (#528) |
|---|---|---|
| `5e14` | **false** (≥ `DIVERGENCE_OFV` = 1e14) | true (`abs` < 1e15) |
| `-1e16` | true (one-sided by design: a large negative −2 log L is legitimate) | **false** |

`the_two_mcem_objective_rules_reject_different_runs` asserts the separation itself first, so
the test cannot quietly become two copies of one check.

**The one exemption, named once.** `publishes_no_objective(method, options)` —
`method = vi` under the default `vi_final_ofv = none` reports `ofv = NaN` *deliberately*
(`ViFinalOfv::None`: "no number at all is safer than a number that looks like an OFV and is
not one"). That is a declaration that **no objective was published**, not a failed one, and a
blanket rule would report `converged: false` for **every default VI fit** — a false alarm on
the common path, which `require_converged` would turn into a rejected search candidate.
`vi_final_ofv = laplace` publishes a real `2·pop_nll` and is gated like everything else, so
this exempts a *setting*, not an estimator. G4 asserts both directions.

## Decision: warning, not `Err` — and why

The issue asks this explicitly. **A warning plus `converged: false`, not an error.** Four
reasons, in order of weight:

1. **An `Err` discards exactly what the user needs to fix the problem.** `fit()` currently
   returns `Ok` here with the per-subject sdtab rows, the EBEs that show which subject
   diverged, and — since #1296 — a `W_ODE_SOLVER_DIAGNOSTICS` warning whose
   `abandoned_non_finite_timeline` count names how many walks were abandoned. `Err(String)`
   carries one line of prose and deletes all of it. The measured run above returns six
   warnings; an error returns one.
2. **The sentinel case is not an error in any useful sense.** A repelled fit has real, finite
   parameter estimates and is precisely the "scored but rejected" candidate a model-space
   search expects from `Strictness::require_converged`. Turning it into an exception would
   abort a `ferx-tools` search instead of stepping past one candidate.
3. **It is a behaviour break for every caller.** `fit()` returning `Err` where it returned
   `Ok` changes control flow in ferx-r and in every tool; tools that treat `Err` as "the model
   or data is invalid" would misreport a numerical outcome as a validation failure.
4. **The right place to error is the front door.** `check_model_data` can name the offending
   *record* before a fit is paid for — that is #1235/#1286's territory, and a non-finite dose
   **time** still passes it with 0 diagnostics. A back-door `Err` after four outer iterations
   tells the user strictly less than the warning does, at the cost of everything above.

So: the boolean carries the verdict, the warning carries the reason, and `Err` stays for
inputs that can be rejected up front. The argument is recorded on
`api::postfit::nonfinite_objective_warning` so it is not re-litigated from scratch later.

## Alternatives considered

**Privatising `OuterResult::converged` behind a constructor**, which would make the invariant
structural rather than scanned. Rejected for this PR: it is a breaking public-API change
(`ferx-tools` reads the field) and wants its own baseline diff and ferx-r check. The scan test
below buys the same property without the break.

**Gating only in `fit_inner`** — the one-call-site shape the issue warns against. It leaves
`optimize_population` and the four public `run_*` entry points reporting the old, wrong
verdict; `the_optimizer_publishes_its_own_gated_verdict_not_only_fit` is the test that would
have been impossible to write.

**Unifying `ofv_is_valid` and `objective_converged` into one predicate.** They are not nested
(table above) and each is load-bearing on a side the other is not, so collapsing them would
silently drop a class of run. Composed instead, with the asymmetry recorded at the call site.

## Cross-repo dependency

No public-API change: every new item is `pub(crate)` and no `WarningCode` variant was added —
`tools/update-public-api.sh --check` reports the baseline unchanged. `FitResult` and
`OuterResult` field *shapes* are untouched; only the `converged` **value** changes, and only on
runs that were already broken. ferx-r reads the field as-is and gets the corrected verdict for
free.

## Breaking changes
- [ ] Public Rust API
- [x] None

Behaviourally, a fit that previously reported `converged: true` at a `NaN` / infinite /
sentinel objective now reports `converged: false` and carries one extra warning. Anything
downstream that gates on `converged` will newly exclude such a run — which is the point.

## Numerical validation
- [x] Not applicable — no estimator, likelihood, gradient or PK/ODE kernel changed. No
      `*_g<T: PkNum>` body was touched, so the `Dual2` path is untouched and no FD-parity
      check applies.

**No NONMEM anchor, and the reason is not that one is awkward to build.** The quantity is
ferx's own convergence verdict against ferx's own divergence sentinel — NONMEM has no
spelling for "which of my own guards fired", and its `MINIMIZATION SUCCESSFUL` line is
computed from a different stop rule against a different (absent) clamp. The gate's inputs are
**exact hand-computed values**, not tolerances: `NaN`, `±inf`, `2e20`, `1e14`, and the two
MCEM witnesses `5e14` / `-1e16`, each asserted against the boundary it straddles.

**The twin, for what did and did not move.** `fit()` on the `NaN`-dose-time population:

```
# before (fcfe8b10)              # after (this PR)
ofv       = NaN                  ofv       = NaN          (unchanged)
ofv_data  = NaN                  ofv_data  = NaN          (unchanged)
converged = true                 converged = false        <-- the fix
n_iter    = 4                    n_iter    = 4            (unchanged)
warnings  = 6                    warnings  = 7            (+ W_NONFINITE_OBJECTIVE)
```

The estimates, the EBEs, the iteration count and every other warning are byte-identical: this
changes what the fit *says about itself*, not what it computes. The full suite is the other
half of that evidence — 4789 lib tests and 213 integration binaries pass unchanged, including
every `nonmem_anchor/*` binary, so no fit that was converging stopped.

## Performance
- [x] No significant impact expected — one comparison per published `OuterResult` and one per
      `FitResult`, i.e. a handful per fit, none of it inside an objective evaluation.

## Tests
- [x] Unit tests added in the same module (`cargo test --lib` passes)

Nine Tier-1 tests: eight in `src/api/tests/nonfinite_objective_tests.rs`, one in
`estimation/impmap.rs`'s inline module.

| | test | what it pins |
|---|---|---|
| G1 | `a_fit_whose_objective_is_nan_is_not_reported_converged` | **the reported regression**, end to end through `fit()` |
| G2 | `the_gate_rejects_the_finite_sentinel_not_only_nan` | the sentinel trap: `is_finite()` is not the gate |
| G3 | `the_gate_leaves_a_real_objective_and_an_already_failed_verdict_alone` | the two false-positive directions |
| G4 | `vi_publishing_no_objective_is_exempt_but_vi_publishing_one_is_not` | the exemption, **both** directions |
| G5 | `a_non_finite_prior_penalty_demotes_a_fit_whose_likelihood_is_finite` | the case only `fit()`'s gate can see |
| G6 | `the_demotion_warning_classifies_as_a_critical_convergence_failure` | the token arm, on a reworded message |
| G7 | `the_optimizer_publishes_its_own_gated_verdict_not_only_fit` | the public `OuterResult` surface |
| G8 | `the_two_mcem_objective_rules_reject_different_runs` | #528 and #1303 reject disjoint sets |
| G9 | `every_published_convergence_verdict_is_gated_on_its_objective` | the pairing, structurally |

| mutation | G1 | G2 | G4 | G5 | G6 | G7 | G8 | G9 |
|---|---|---|---|---|---|---|---|---|
| the gate always returns `None` (**the fix off**) | **RED** | RED | RED | RED | RED | RED | RED | green |
| `ofv_is_valid` → `ofv.is_finite()` in the gate | green | **RED** | green | green | green | green | **RED** | green |
| delete the `fit_inner` call site | green | green | green | **RED** | green | green | green | **RED** |
| delete the NLopt exit's call site | green | green | green | green | green | **RED** | green | **RED** |

Four things that matrix exists to show:

1. **G1 alone is not enough, and the PR does not pretend it is.** Row 3 — deleting `fit()`'s
   gate — leaves G1 **green**, because the estimator's gate already demotes on that path. G5
   is the fixture built specifically for the number only `fit_inner` computes
   (`ofv_data + ofv_prior`), and it is a different quantity rather than a second copy of G1.
2. **Row 4 is the mirror image**, and G7 is why it is caught behaviourally rather than only by
   the scan: `optimize_population` is public and a caller reading its `OuterResult` never
   touches `fit_inner`. G1 stays green under it, which is the asymmetry, not a gap.
3. **Row 1 leaves G9 green on purpose** — the structural test checks that the *pairing* exists,
   not what the gate does; the two are deliberately different properties, and each row above
   has at least one behavioural killer.
4. **Row 2 is the sentinel half specifically.** Weakening only the `< DIVERGENCE_OFV` clause
   kills G2 and G8 and nothing else — which is exactly the accident #1303 says the `NaN` half
   alone would have been.

### G5's fixture, because it is not obvious

The penalty has to *overflow*, not merely be large. A **log-packed** prior cannot do it: `sd`
is first converted to a relative spread and `sqrt(ln(1 + rse²))` underflows to `0`, which
`PriorSet::build` rejects outright ("the declared spread gives a non-positive or non-finite
prior SD (0)") — measured while writing the test. So the prior sits on a THETA with a
**negative lower bound**, which packs on the identity scale (`theta_packs_log(-1.0) == false`)
where the declared `sd` *is* the packed SD. Centred at `1e6` on a coordinate boxed into
`[-1, 1]` with `sd = 1e-200`: `z ≈ -1e206`, `z² = +inf`, while the likelihood at those
ordinary estimates stays finite. The test asserts `ofv_data.is_finite()` **and**
`!ofv_prior.is_finite()` before it asserts anything about `converged` — without that premise
it would be G1 again.

### G9's allowances, stated exactly

Counted per file, both directions (a removed gate fails as loudly as an ungated new site).
One documented exception: `io/fitrx.rs`'s checkpoint restore is `(1 publishing, 0 gates)`,
because `converged: w.converged` **copies** a verdict a previous fit already gated, together
with the `ofv` that verdict was about; re-gating there would apply a rule to a number that
code did not compute and would rewrite a bundle's recorded history.

- [x] Not applicable — no new DSL syntax, `[fit_options]` key, output field or estimator. The
      user-visible surface is the `converged` value and the new warning, both documented.

## Docs
- [x] `docs/` updated
- [x] `CHANGELOG.md` `[Unreleased]` → `### Fixed`

- `docs/warnings.qmd` — new `## A non-finite objective {#non-finite-objective}` section (what
  it means for the result, why `is_finite()` is not the test, what to look for, and the VI
  exception); the `convergence` row now names `W_NONFINITE_OBJECTIVE`; a `details`-payload row.
- `docs/output.qmd` — `converged` added to the fit-YAML **Key Fields** list, which did not
  describe it before.
- `docs/api/fitting.qmd` — the `Strictness` prose and the `require_converged` row.
- `src/model_selection.rs` module doc and `Strictness::require_converged` doc, so the Rust-side
  statement of the same contract does not drift from the page.

`cargo run -p docs-lint -- --check` is green with the baseline's five excused sections
unchanged — no new R1 entry, and the `#non-finite-objective` anchor resolves from both pages
(R4).

## Checklist
- [x] `tools/preflight.sh` green — all six groups (fmt, the 5 `cargo check` feature sets,
      clippy `--all-targets`, rustdoc, docs, public-API)
- [x] `Dual2` path — no `*_g<T: PkNum>` body touched; differentiability untouched
- [x] No version bump (no API change)

Full runs at `869d17b7`:

```
cargo test --lib --no-default-features --features ci,survival --no-fail-fast
  -> 4789 passed / 0 failed / 23 ignored

cargo test --workspace --tests --no-default-features \
           --features ferx-core/ci,ferx-core/survival --no-fail-fast
  -> 213 binaries / 6650 passed / 0 failed / 197 ignored
```

### ferx-site / ferx-book

### Example execution
- [x] No example execution step needed — no example's numbers move (see the twin above); the
      only behaviour change is on runs whose objective is already not a number.

## Reviewer hints

- **Start with the mutation matrix, rows 3 and 4.** They are the two that show the
  `fit()`-level gate and the estimator-level gate are separate surfaces rather than one rule
  written twice, and they are why G5 and G7 exist at all. If you disagree that `ofv_prior` is
  a different quantity worth its own gate, that is the row to argue with.
- **The `Err`-vs-warning decision** is in its own section above and on
  `nonfinite_objective_warning`'s doc comment. Reason 2 (the search-candidate case) is the one
  I would most like a second opinion on.
- **`chain.last()` vs `final_method`** at `api/fit.rs` is a real trap, commented at the call
  site; it is the difference between exempting a VI stage and exempting a real IMP objective
  that happens to follow one.
- **Skimmable:** `estimation/bayes.rs`, `estimation/saem.rs`, `estimation/vi/run.rs` and the
  two `api/fit.rs` synthetic stages are the same four-line pattern each.
- **`estimation/trust_region.rs` is the one place behaviour changed beyond the gate**: its
  bespoke message sat behind `warnings.is_empty()`, so a trust region that both stalled *and*
  landed on a `NaN` reported only the stall. That guard is gone; the shared message always
  fires.

## Open questions

None blocking. Two judgement calls worth a look:

1. **The VI exemption is keyed on `vi_final_ofv`, not on a field of `OuterResult`.** An
   alternative is an explicit "this result publishes no objective" flag on `OuterResult`, which
   would be self-describing but is a public-API widening. The current form is one named
   predicate with a test on both directions.
2. **The covariance step still runs before the gate** on a fit that is about to be demoted, so
   a diverged run pays for a Hessian nobody should read. That is pre-existing and orthogonal;
   say the word and I will file it.

https://claude.ai/code/session_012QkMSfpEWva9pwT4eMYAoE


roninsightrx (commented):
I found two blocking issues.

1. **[P1] Preserve the VI no-objective exemption for a trailing eval-only IMP stage** (`src/api/fit.rs:2421-2430`). `imp_eval_only` does not replace `result.ofv`; it stores the finite IS likelihood in `FitResult.importance_sampling.minus2_log_likelihood` and leaves the preceding estimator objective in `FitResult.ofv` (`src/api/fit.rs:1795`, `2112`, `2648`). Therefore, for `methods = [vi, imp]`, default `vi_final_ofv = none`, and `imp_eval_only = true`, the value passed here is still VI's deliberate `NaN`, not IMP's IS likelihood. Keying the exemption on `chain.last()` treats that intentional VI `NaN` as a failed IMP objective and demotes a successfully converged VI fit. I reproduced this on a three-subject closed-form model: solo VI returned `converged=true, ofv=NaN`; the identical VI run followed by a successful IMP readout returned `converged=false, ofv=NaN`, even though `importance_sampling.minus2_log_likelihood` was finite (`-19.0799`). This is exactly the documented chain the exemption is meant to preserve. Please track the owner of `FitResult.ofv` rather than assuming the terminal stage owns it, and add an end-to-end VI -> IMP-eval regression test.

2. **[P2] The documented structured `details` payload is lost on the normal failure path** (`src/api/postfit.rs:1489`). Every estimator-side gate first sets `result.converged = false` and appends only a string warning. When this fit-level helper runs, `gate_converged_on_objective` immediately returns `None` because the verdict is already false, so no native `WarningEntry` with `ofv`, `reason`, `divergence_cutoff`, and `method` is added. `rebuild_warnings_structured` then merely classifies the string, and `diagnostic_details` has no `Convergence` enrichment arm, leaving `details = None`. Thus the common regression case contradicts the new docs and PR description; only cases first discovered at the fit-level total (such as prior overflow) get the payload. The same early return also suppresses `W_NONFINITE_OBJECTIVE` entirely when an optimizer already reported non-convergence for another reason. Please separate objective invalidity/reporting from the boolean mutation, and deduplicate an existing warning explicitly so the fit-level entry can retain its typed payload.

Checks run locally: focused non-finite-objective unit tests (8 passed), the MCEM predicate regression (passed), and `cargo check --tests --no-default-features --features ci,survival` (passed). CI is green.

roninsightrx:
Both findings were real and both are fixed in `f9957bf7`. Reproduced each before changing anything; the second one had a test actively pinning the wrong behaviour, which is worth saying out loud.

## 1 (P1) — the exemption is now keyed on the objective's **owner**, not on a chain position

Confirmed exactly as described. `imp_eval_only` writes `is_result` and `continue`s (`api/fit.rs:1663-1666` says so in prose), so `result.ofv` keeps the preceding estimator's objective. For `methods = [vi, imp]` + `vi_final_ofv = none` that is VI's deliberate `NaN`, and `chain.last()` attributed it to IMP. My comment at the call site argued the opposite and was simply wrong about where IMP-eval's number goes.

**`final_method` is not the fix either**, which is the part worth flagging: an `agq_eval_only` readout is evaluation-only *and* replaces `prev.ofv` with its quadrature marginal (`api/fit.rs:1614`). So:

| chain | writes `result.ofv`? | owner |
|---|---|---|
| `[vi, imp]` + `imp_eval_only` | no — goes to `importance_sampling` | **VI** (`final_method`) |
| `[vi, laplace]` + `agq_eval_only` | **yes** — replaces it | **the readout** (`chain.last()`) |

The two evaluation-only families differ in exactly the thing the gate needs, so each candidate rule is right for one and wrong for the other. `objective_stage` now records the owner at each of the four sites that write `result.ofv`, and the gate reads that.

Two tests, deliberately separable:

- `a_trailing_eval_only_imp_stage_does_not_claim_vis_deliberate_nan` — your case, asserting both premises first (`ofv.is_nan()` **and** `importance_sampling.minus2_log_likelihood.is_finite()`, so it cannot pass on a chain where IMP never ran), then that appending the readout does not change VI's verdict versus the solo VI fit.
- `an_agq_readout_owns_the_objective_it_writes` — the mirror, so the P1 fix cannot be "switch to `final_method`".

Mutation, run: `→ *chain.last()` kills only the first; `→ final_method` kills only the second. Nothing else in the file moves under either.

One incidental measurement: the AGQ arm comes back at the **sentinel** (`2e20`), not `NaN` — `agq_population_nll` sums per-subject contributions the inner objective has already clamped. So that test also exercises the `is_finite()`-would-miss half end to end, which G2 previously only pinned as a unit.

## 2 (P2) — both gates are now pure predicates on the objective

Also confirmed, and worse than the payload half alone: `the_gate_leaves_a_real_objective_and_an_already_failed_verdict_alone` **asserted** the suppression ("an already failed verdict must come back with no message"), so I had written the defect into a test and called it a design. It is inverted now — `a_run_that_already_failed_still_reports_an_unusable_objective` — and covers the MCEM composite on its own #528 rule too.

The `!*converged` early return is gone from `gate_converged_on_objective` and from `gate_converged_on_mcem_objective`. "The optimizer hit its evaluation budget" and "the objective is not a number" are different facts with different consequences — provisional estimates versus nothing usable at all — and whichever fired first silenced the other.

De-duplication is now explicit at the one place that can see both, in `fit_inner`:

```rust
warnings.retain(|w| !w.ends_with(msg.as_str()));
warnings.push(msg);
native_warnings.push(entry);
```

Suffix rather than equality so a `[STAGE] ` chain prefix does not defeat it; the message interpolates the offending value, so a *different* stage's non-finite objective in the same chain is a different string and survives. The typed entry then keys by message through `rebuild_warnings_structured` and keeps its `details`.

`the_demotion_carries_its_documented_details_payload` asserts all four keys on G1's fixture — the common path, not the prior-overflow one. Mutation, run: restoring either early return reddens it and G3b (and G12); dropping the `retain` reddens G1's exactly-one assertions instead.

Docs updated to match rather than the other way round: `warnings.qmd` now states that `W_NONFINITE_OBJECTIVE` is the exception to the "omit `details` when a contributing statistic is non-finite" rule (the offending value *is* the payload, carried as a string since JSON has no non-finite numbers), and that the exemption follows the objective rather than the chain position. A `CHANGELOG` line was added for the both-reasons-reported behaviour change.

## Gates

- `cargo test --lib --no-default-features --features ci,survival --no-fail-fast` → **4793 passed / 0 failed**
- `cargo test --workspace --tests --features ferx-core/ci,ferx-core/survival --no-fail-fast` → 213 binaries / 6653 passed / **1 failed**, and it is unrelated: `ferx_tools::gam::tests::a_covariate_name_with_a_comma_does_not_shift_the_columns` panicked on `create_dir` with `AlreadyExists`. `gam.rs:1025` names its temp dir `pid + SystemTime nanos`, which collides across parallel test threads in one process on macOS. Passes in isolation and on a clean tree (`git stash`-verified); no `crates/` file is in this PR's diff. Happy to file it.
- `tools/preflight.sh` green, all six groups; public API still matches the committed baseline.

The PR description's mutation matrix is superseded by the one above for the four new tests; the original four rows still hold.

PR #1385, opened 2026-09-13
perf(estimation): improve pooled population workloads

hiddevandebeek:
## Why

Several independent population-level paths paid avoidable scheduling or allocation overhead, while Bayesian chains, underfilled AGQ grids, and public prediction left available workers idle. This bundles the measured high-priority pool optimizations into one review, as requested.

## What changed

- Run independent Bayesian chains concurrently with chain-local RNG, state, diagnostics, and posterior accumulators; fold outputs in chain order.
- Keep cheap small-population FOCE NLL passes serial while leaving expensive EBE, ODE sensitivity, and finite-difference work parallel.
- Use nested AGQ-node parallelism when the subject axis underfills the active pool, with indexed collection and ordered folding.
- Coordinate identical concurrent shared-pool cold misses through a per-key `OnceLock`, while preserving concurrent construction for different keys and retry after failure.
- Replace AGQ-IOV's per-node heap-allocated kappa slice table with inline storage (heap fallback beyond 16 occasions).
- Evaluate public `predict()` subjects concurrently whenever the active pool has more than one worker. Enclosing `PoolPlan` calls reuse that pool; standalone calls reuse the persistent default pool. Final rows are assembled in original subject/observation order.

## Alternatives considered

- Expensive-subject-first scheduling was rejected: 1.04x on heterogeneous MBMA but 0.93x on homogeneous warfarin ODE, with worse tail latency.
- Simulation fan-out was not changed because its single variable-consumption RNG stream makes seeded compatibility non-trivial.
- Prediction admission by population size was tested, then deliberately omitted: multi-worker calls always fan out by subject; one-worker calls retain the direct serial loop.
- AGQ-IOV allocation removal is retained for allocator/memory-pressure reduction despite near-neutral wall time on the largest benchmark.

## Cross-repo dependency

## Breaking changes

- [x] None

<details>

None.

</details>

## Numerical validation

- [x] Compared against prior ferx `origin/main` (`257be39a`).
- [x] Bayesian posterior summaries, OFV, ESS inputs, divergences, and per-chain draws are bit-identical across worker counts.
- [x] AGQ serial/parallel scores and AGQ-IOV legacy/inline objectives are bit-identical.
- [x] Public analytical and ODE predictions are bit-identical across one/four workers, including ID, TIME, PRED, and row order.
- [x] NONMEM comparison is not applicable: these are execution-strategy changes and do not change numerical results.

## Performance

- [x] Faster — controlled optimized A/B measurements:

| Optimization | Before | After | Result |
|---|---:|---:|---:|
| Bayesian chains, 1 vs 4 workers | 1.45 s | 0.56 s | **2.59x** |
| Small FOCE NLL, 4 subjects | 2.694 us | 1.001 us | **2.69x** |
| Small FOCE NLL, 8 subjects | 3.195 us | 1.964 us | **1.63x** |
| Underfilled AGQ gradient | 339.961 us | 183.999 us | **1.85x** |
| Shared-pool 8-caller cold burst | 17.08 ms / 8 builds | 10.44 ms / 1 build | **1.64x**, 87.5% fewer builds |
| ODE `predict()`, 160 subjects | 7.491 ms | 2.491 ms | **3.01x** |
| ODE `predict()`, 320 subjects | 14.537 ms | 4.622 ms | **3.15x** |
| Analytical `predict()`, 160 subjects | 135.502 us | 124.535 us | **1.09x** |
| AGQ-IOV, 500 subjects / 1,562,500 nodes | 11.365 s | 11.283 s | **1.007x**, 1,562,500 allocations removed |

Small cheap analytical prediction can be slower without an admission cutoff (80 subjects measured **0.874x**); this is disclosed and intentional so every multi-worker public prediction call uses the available subject parallelism.

## Tests

- [x] Unit tests added in the same modules; `cargo test --lib`: **4,648 passed, 0 failed, 28 ignored**.
- [x] Public prediction integration tests added; **2 passed, 0 failed, 1 benchmark ignored**.
- [x] Regression coverage includes bitwise ordering, pool construction count, dispatch thresholds, AGQ score parity, and Bayesian worker-count reproducibility.

- [x] Not applicable (internal execution changes with no new API surface).

## Docs

- [x] `predict()` rustdoc documents pool reuse and deterministic ordering.
- [x] `CHANGELOG.md` `[Unreleased]` performance entry added.
- [x] No new docs page or navigation entry required.

## Checklist

- [x] `tools/preflight.sh` green — fmt, feature checks, clippy, rustdoc, docs lint, public API baseline.
- [x] Functions on the `Dual2` sensitivity path remain differentiable; no sensitivity algebra changed.
- [x] No version bump required.

### ferx-site
- [x] No DSL, example, or data change.

### ferx-book
- [x] No estimator, DSL, or fit-option documentation change.

### Example execution
- [x] No example execution step needed (internal performance changes only).

## Reviewer hints

Review the execution boundaries first: Bayesian state is chain-local; AGQ collections are indexed and folded in stable order; `predict()` parallelizes calculation but assembles public rows serially; shared-pool failures remove their cache entry so later calls can retry. The AGQ-IOV inline table deliberately falls back to the existing heap path beyond 16 occasions.

## Open questions

None.


hiddevandebeek:
Added analytical inner-curvature work in fa1833f3.

Measured A/B results on the 10-subject, 3-ETA warfarin fit (same optimized binary, 3 threads):

| Method | Previous seed | Analytical seed | Speedup | Inner gradient steps |
|---|---:|---:|---:|---:|
| FOCE, closed form | 0.264 s | 0.121 s | 2.18x | 44,320 -> 13,592 |
| FOCEI, closed form | 0.217 s | 0.134 s | 1.62x | 37,659 -> 16,664 |
| FOCE, ODE | 1.373 s | 1.018 s | 1.35x | 6,137 -> 4,932 |
| Laplace(1) | 0.243 s | 0.238 s | 1.02x | 45,393 -> 46,594 |
| AGQ(3) | 0.290 s | 0.379 s | 0.77x | 42,258 -> 44,544 |

The measured AGQ regression is not admitted: Hessian seeding is enabled only for non-quadrature FO/FOCEI/GN-family stages. Laplace/AGQ instead retain the exact terminal Hessian, fuse its Dual2 pass with terminal Jacobian construction, and reuse it in the derivative-free quadrature objective. The optimizer seed uses the stabilized Gauss-Newton/Almquist Hessian through a Cholesky solve; failed or non-PD seeds retain the existing metric.

Validation: focused Cholesky fallback and bitwise terminal-reuse tests pass; all 23 AGQ tests pass; repository clippy preflight passes; public API baseline unchanged. Full lib run: 4,649 passed before one unrelated post-fit ODE counter test differed across worker counts; that test passed immediately in isolation.

PR #1386, opened 2026-09-14
fix(estimation): report the objective the optimizer reached, and seed the FREM inner restart from the data [closes #833, closes #1349]

### LINKED ISSUE #833: fix(estimation): final restored-point inner loop cold-starts EBEs → reported Final OFV > outer best-seen OFV
## Summary

After the outer optimizer restores the best-seen point (issue #59), the **final inner loop** re-derives EBEs with `prev_etas = None` (cold start), whereas every optimization-time evaluation warm-started the inner loop from the previous iteration's EBEs. On an ill-conditioned / weakly-identified inner problem the cold-start EBEs converge to slightly different η̂ than the warm ones, so the **reported `Final OFV` does not reproduce the outer best-seen OFV** it was restored to — it comes out *higher*.

The restored point is advertised as "the actual minimum" (`outer_optimizer.rs:1316-1321`), but the number ferx then reports for it is worse than the value that made it the best-seen point.

## Observed

Fluconazole 2-cpt binding model (`ferx-testdata/fluconazole_radboudumc`), FOCEI, from NONMEM-final starts:

```
Eval   53: OFV = 738.060959      <- outer best-seen
NLopt stopped: Failure
Restored best-seen point (OFV = 738.060959) for final inner loop and covariance step.
Final OFV = 741.585955           <- final re-eval, +3.5 higher
```

`Final OFV` (741.59) is what lands in the reported OFV/AIC/BIC and drives the covariance step, even though the outer loop achieved 738.06 at the same parameters. (NONMEM's reference OFV for this model is 734.64, so the spurious +3.5 is a meaningful slice of the residual gap.)

## Root cause

`src/estimation/outer_optimizer.rs`:

- During optimization, the objective closure runs the inner loop **warm** (seeded from the prior iteration's EBEs), and `best_seen` records `(best_xs, best_ofv)` with that warm OFV (`~:934`, `~:1132`).
- The final inner loop at the restored point passes `prev_etas = None`:

```rust
// outer_optimizer.rs ~1344
let (final_ehs, final_hms, _, final_kappas) = run_inner_loop_warm(
    model, population, &final_params,
    options.inner_maxiter, options.inner_tol,
    None,                 // <-- cold EBE start; optimization used warm
    Some(&final_mu_k),
    options.min_obs_for_convergence_check as usize,
);
let final_ofv = 2.0 * pop_nll(..., final_ehs, final_hms, ...);
```

`run_inner_loop_warm`'s `prev_etas: Option<&[DVector<f64>]>` is the warm-start seed. With `None`, the inner optimizer starts from the default (η=0 / cold) and — on a subject whose individual objective is flat or multimodal (here the peripheral Q/V2 is nearly non-identifiable, RSE > 100%) — settles at different η̂ than the warm run, so `final_ofv > best_ofv`.

## Impact

- Reported OFV/AIC/BIC are inflated relative to the optimum the fit actually reached; users see `Final OFV` worse than the last printed `Eval` line, which reads like a bug.
- The covariance step runs at the cold EBEs too, so SEs are computed at a slightly different point than the reported minimum.
- Purely a reproducibility/consistency defect — the parameters are correct; only the EBE-dependent readouts drift.

## Fix direction

Make the final inner loop reproduce the best-seen objective:

1. **Stash the best-seen EBEs** alongside `(best_xs, best_ofv)` (the tracker currently keeps only the packed vector + OFV) and seed the final inner loop with them via `prev_etas`. Cleanest — the final loop then starts exactly where the recorded OFV was measured.
2. Or warm-start the final inner loop from the **last optimization EBEs** and assert/log if `final_ofv` still disagrees with `best_ofv` beyond a tolerance (surfaces genuine inner multimodality rather than hiding it).

Either way, `Final OFV` should equal the restored `best_ofv` to inner-loop tolerance.

## Tests

- Tier 2: a fit whose inner problem is warm/cold-sensitive (e.g. a weakly-identified peripheral compartment) must report `Final OFV` within `inner_tol` of the outer best-seen OFV.
- Regression guard: `final_ofv <= best_ofv + eps` after the restore.

## Related

- Best-seen restore machinery: #59.
- Lbfgs `NLopt stopped: Failure` brittleness on ill-conditioned surfaces overlaps #826.
- Surfaced while validating the `block_sigma` replicate-time fix (#827 / #830).

### LINKED ISSUE #1349: bug(estimation): FOCEI FREM nightly regressed to OFV 3345.8 (expected ≈ 211) at 86ccd0ba (#1297, EBE warm-start anchoring); green at its parent
## Symptom

The nightly `slow-tests` core leg has failed `frem_covariate_omega_matches_sample_variance_under_focei` (`tests/frem_warfarin.rs:658`) on **every run since 2026-09-08 22:30 UTC**, with byte-identical numbers each time:

```
FOCEI FREM: OFV=3345.8  omega_WT=106.80 (want 111.56)  omega_AGE=93.84 (want 99.38)
thread 'frem_covariate_omega_matches_sample_variance_under_focei' panicked at tests/frem_warfarin.rs:658:5:
FOCEI FREM OFV = 3345.8; expected ≈ 211. A value near 4900 means the analytic gradient is
differentiating the PK likelihood on covariate pseudo-observation rows while the objective
scores them against theta[i] + eta[j] with EPSCOV.
```

The assertion is `ofv < 1000.0`. The test's own doc comment says the OFV is *the* regression check for this fixture: the covariate ω's recover roughly right whatever the optimizer does, so the ω's landing 4–6 % low here is not the signal — the **3100-unit OFV move** is. 3345.8 is neither the ≈ 211 optimum nor the ≈ 4901 signature the message names, so this is a third landing point, not the documented gradient failure returning.

Recorded so far only as a 2026-09-10 comment on #990; filing it on its own because it is a live regression on `main` one to two weeks before 0.4.0, and #990 is about macOS/Linux splits.

## Attribution — one commit, measured by CI, no local bisect needed

Every push to `main` in this window dispatched its own nightly run, so the per-commit verdicts already exist. Read off the core-leg logs (passing tests capture stdout, so green runs print no OFV line):

| commit | run | `frem_covariate_omega_matches_sample_variance_under_focei` |
|---|---|---|
| `adfb703e` (2026-09-07) | [34169391829](https://github.com/FeRx-NLME/ferx-core/actions/runs/34169391829) | ok |
| `c877b9a9` #1276 | [34193907859](https://github.com/FeRx-NLME/ferx-core/actions/runs/34193907859) | ok |
| `b2e205e1` #1283 FOCEI fusion | [34209580745](https://github.com/FeRx-NLME/ferx-core/actions/runs/34209580745) | ok |
| `af4e740d` | [34233759463](https://github.com/FeRx-NLME/ferx-core/actions/runs/34233759463) | ok |
| `6ec94356` #955 analytic covariance AGQ/IOV/M3 | [34238183384](https://github.com/FeRx-NLME/ferx-core/actions/runs/34238183384) | ok |
| `04c952f5` | [34244824916](https://github.com/FeRx-NLME/ferx-core/actions/runs/34244824916) | ok |
| `46ede44f` #1286 | [34248684467](https://github.com/FeRx-NLME/ferx-core/actions/runs/34248684467) | ok |
| `cf32e1fd` #1282 | [34250435946](https://github.com/FeRx-NLME/ferx-core/actions/runs/34250435946) | ok |
| `6dafe103` #1293 | [34266839040](https://github.com/FeRx-NLME/ferx-core/actions/runs/34266839040) | ok |
| `674cdcf8` #1291 analytic ODE covariance | [34274074211](https://github.com/FeRx-NLME/ferx-core/actions/runs/34274074211) | **ok** |
| **`86ccd0ba` #1297 anchor the EBE warm start to the best point seen** | [34286273326](https://github.com/FeRx-NLME/ferx-core/actions/runs/34286273326) | **FAILED, OFV 3345.8** |
| `5696b1d6` #1292 | [34286745923](https://github.com/FeRx-NLME/ferx-core/actions/runs/34286745923) | FAILED, OFV 3345.8 |
| `1c0c1626` (2026-09-11, HEAD) | [34595670902](https://github.com/FeRx-NLME/ferx-core/actions/runs/34595670902) | FAILED, OFV 3345.8 |

So the first red run is the one **for** `86ccd0ba` itself, and its parent `674cdcf8` is green. `86ccd0ba` does not touch `tests/frem_warfarin.rs`; its source changes are confined to `src/estimation/outer_optimizer.rs` (+50/−) and its tests. Everything after it reproduces the same 3345.8 to the printed digit, so nothing since has moved the fit.

This is attribution by measurement, not by reading: the same table rules out #1283 (FOCEI fusion), #955 and #1291 (analytic covariance), which were the plausible-looking suspects from the window.

## Why the commit is a credible cause, not just the first red

#1297 changed **which EBEs seed the inner loop**: the warm-start cache is now taken only from an evaluation that improved on the best objective seen, never from a rejected one. Its own description says the individual objective is multimodal (#864 / #891) and the fix exists to stop the cache landing in "a worse basin that the inner loop then keeps re-finding".

The FREM fixture is exactly a model whose reported OFV is decided by which mode the inner loop lands in — the doc comment on this test says so in as many words ("the inner loop therefore drove the EBEs to the mode of the wrong likelihood, and the FOCEI objective — evaluated *at* those EBEs — landed ~4700 units worse"). A change to the warm-start policy moving this fit by 3100 units is therefore the expected shape of a side effect, and the ω's drifting 4–6 % *low* (106.80 / 93.84 against 111.56 / 99.38) while the OFV explodes matches "different EBE basin", not "different gradient".

Not yet established: whether 3345.8 is a worse local optimum the new warm start steers into, or whether the fit now stops early (the test runs with `run_covariance_step = false` and default `outer_maxiter`, and does not print `converged` or the iteration count).

## Prior history of this number, so it is not misread

- 2026-08-19 (#990): Linux 6.9, macOS 18027.4 — a platform split that later resolved.
- 2026-09-02 (#990 comment): both platforms ≈ 278 at `6bb6ebd5`.
- 2026-09-08 onward: 3345.8 on Linux (this issue). macOS unmeasured at HEAD.

The `< 1000` bound is loose enough that 278 passed and 3345.8 fails; neither is the ≈ 211 the test documents as the optimum, so the fixture has been moving for weeks and this is the first move that crossed the bound.

## Suggested route

1. Reproduce locally on Linux (a `linux/arm64` container reproduces the CI numbers — recipe in #990's 2026-09-02 comment) at `86ccd0ba` and at `674cdcf8`; confirm the one-commit flip. Print `converged`, `n_iterations`, the per-subject OFVs and the EBE norms on both sides so the mechanism is read off the fit rather than guessed.
2. Revert `86ccd0ba`'s `outer_optimizer.rs` change in place on HEAD and re-run: if the OFV returns to ≈ 278 the attribution is causal, not correlational.
3. Then decide: the warm-start anchoring fixed a real stall on covariate models (`two_cpt_oral_covmodel.ferx` −1026 → −1195, NONMEM-anchored), so the fix is not a revert. Likely shape: the anchored cache must not be adopted for a subject whose inner loop fell back or did not converge at the best point, or FREM's covariate pseudo-observation rows need the cache keyed on the mode they pin.
4. Tighten the bound once the fixture is stable: `< 1000` has been passing at 278 against a documented 211, which is a 30 % drift going unremarked.

Refs #990 (where this was first recorded), #1297 (the commit), #864 / #891 (the multimodality the commit cites), #1217.

Found during the 2026-09-11 release triage.

roninsightrx:
## Why

**#833 and #1349 are the same defect**, and the second one is the first one measured on glibc.

After the outer loop restores its best-seen point (#59), the final inner loop re-derived the
EBEs from a **cold** start, while every evaluation during the fit had warm-started them. On an
inner problem where the two disagree, the reported `ofv` — and the AIC/BIC, and the point the
covariance step runs at — came out *above* the value that made the point best-seen.

The measurement that ties the two issues together, on `tests/frem_warfarin.rs`'s FOCEI fixture:

| | macOS/arm64 | linux/arm64 container (reproduces CI byte-for-byte) |
|---|---|---|
| `main`, reported OFV | −167.95 | **3345.8** ← the nightly failure |
| best-seen OFV (what the optimizer reached) | −174.91 | — |
| ω_WT / ω_AGE | 106.80 / 93.84 | 106.80 / 93.84 |

The ω's agree **to every printed digit across the two platforms** while the OFV differs by
3513. That is not a fit that went somewhere else; it is the same fit with a different readout.
It also explains two things #1349 flagged as unexplained: why the bisect landed on #1297 (that
commit moved the fit slightly, and the cold restart amplified the move), and why the covariate
ω's "recover roughly right whatever the optimizer does" while the OFV explodes.

Verbose output on macOS at `257be39a`, same fixture, no fix:

```
Restored best-seen point (OFV = -174.910951) for final inner loop and covariance step.
Final OFV = -167.952882                     <- +6.96 on the point just restored
```

## What changed

**§1 — the final inner loop re-solves from the incumbent's EBEs (`outer_optimizer.rs`).**
`best_etas` publishes the EBEs of each improving evaluation out of the NLopt closure (NLopt
does not hand `state` back). At the restored point the cold solve runs as before; when it
fails to reproduce `best_seen_ofv`, a second solve runs from those EBEs and the **lower
objective** is reported. Ranking on `pop_nll_opts` is ranking on the very quantity being
reported, so the choice cannot make the reported OFV worse than either candidate alone.

Three details are deliberate:

- **The second solve is conditional.** A unimodal inner problem — most fits — has nothing to
  recover, so it is bit-identical to before at exactly the old cost. The trigger is
  `ebe_start_dependence_gap(cold, best_seen)`, plus an explicit `cold.is_nan()` clause,
  because every comparison against `NaN` is false and the gap test alone would decline to
  re-solve precisely when the cold solve blew up.
- **The plateau check still reads the cold number.** `failure_is_converged_plateau` (#751) uses
  the cold restart as its warm-start-artifact probe. Handing it the *reported* OFV would make
  it compare the best-seen objective against a re-run of itself, so `consistent` would be true
  by construction and the check would stop rejecting warm-start-only "optima". It now takes
  `consistency_ofv` (the cold value) explicitly.
- **A surviving gap is reported, not hidden.** New `WarningCode::EbeStartDependent`
  (`W_EBE_START_DEPENDENT`), quoting both objectives. A cold re-solve that lands materially
  worse means the individual objective has more than one mode at these estimates, so IPRED,
  IWRES, CWRES, shrinkage and the covariance step all depend on which one you took — the user
  should know that even though the reported number is now the better of the two.

**§2 — the FREM Nelder–Mead restart seeds from the data (`inner_optimizer.rs`).** This is why
the glibc gap was 3513 rather than 7. `find_ebe`'s FREM arm re-centres a non-converged subject
with an NM restart, and that restart started from `vec![0.0; n_eta]` — bypassing the FREM-aware
cold seed (#406) that exists because a covariate pseudo-observation eta sits at `cov_obs − TV`,
tens of units from zero. NM's initial simplex step at zero is `0.00025` per coordinate, so the
restart could not travel there: it returned a point with the covariate etas still at ~0 and the
PK etas carrying the resulting block-Ω⁻¹ force. The seed is now `cold_eta_seed`, used by the
cold start *and* the fallback. Non-FREM models get the same zero vector as before, bit-identical.

## Alternatives considered

- **Warm-start the final inner loop unconditionally** (issue #833's fix direction 1). Rejected:
  it deletes the #751 self-consistency probe, which is a live guard — the SS-oral stall it was
  written for (best-seen 83.3 vs cold 121.4) would silently start reporting as converged.
- **Revert #1297's warm-start anchoring**, which the bisect in #1349 pointed at. Rejected on
  measurement: #1297 is not what makes the number wrong (the cold final solve is), and it fixed
  a real stall on covariate models with a NONMEM anchor behind it.
- **Refuse / demote `converged` on a start-dependent EBE surface.** Rejected: the outer fit can
  be perfectly converged; multimodal EBEs are a property of the model, so this is a warning, not
  a verdict. `converged` semantics are unchanged by this PR.
- **Switch the FREM arm to `argmin_inner_fallback`** (the #378 / #1327 policy — keep the lower
  of {BFGS partial, NM restart} instead of taking NM unconditionally). It is the last call site
  still adopting an NM restart unconditionally, and it is probably also wrong, but it changes
  FREM/IMP proposal centring broadly and is not needed for either issue here. Left for its own
  change; measured note in #1349.

## Cross-repo dependency

Nothing in the R wrapper's explicit field list changes. `fit$warnings` /
`warnings_structured` gain a possible `ebe_start_dependent` category, which reaches R with no
wrapper change; the `WarningCode` enum is `#[non_exhaustive]`, so a new variant is additive for
Rust callers too.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated in this PR
- [ ] None

One added item: `WarningCode::EbeStartDependent`. Needed by ferx-r and by agents that branch on
`warnings_structured[].category` — the whole enum is public for that reason, and the existing
codes do not cover it (`convergence` would be actively misleading, since the outer fit is
usually converged when this fires). No item removed or retyped.

## Numerical validation

- [x] Compared against: `origin/main` `257be39a`, both platforms; and the linux/arm64 container
      that reproduces the CI numbers byte-for-byte (recipe in #990)
- [x] Reference values:

```
# tests/frem_warfarin.rs :: frem_covariate_omega_matches_sample_variance_under_focei
#                          macOS/arm64        linux/arm64 (= CI)
# main  257be39a          OFV -167.95        OFV 3345.8   (test FAILS)
# this PR                 OFV -174.91        OFV -220.71  (test passes)
#
# -220.71 is the objective SAEM and IMPMAP already reach on this model
# (docs/estimation/frem.qmd), so the Linux FOCEI arm now agrees with them.

# examples/warfarin.ferx, FOCEI, default options: unchanged, OFV -286.0042.
# The second solve does not run there (the cold solve reproduces best-seen), so the
# fit is bit-identical -- that is the `an_ordinary_fit_reports_no_start_dependence`
# control test.
```

No NONMEM anchor: the quantity is *which of ferx's own two EBE solves at a fixed parameter
vector gets reported*. NONMEM has one inner loop and no such choice to compare against. The
oracles that do apply are (a) the fit's own best-seen objective, which the reported number must
now reproduce, (b) the independently-derived SAEM/IMPMAP objective on the FREM fixture, which
the Linux FOCEI arm now matches, and (c) warfarin's established FOCEI optimum (−286.0042),
which the starved-inner regression test recovers where `main` reports −31.06.

## Performance

The second inner solve runs only when the cold one missed `best_seen_ofv`. On a fit where it
does run it costs one inner loop against the hundreds the outer loop already spent (warfarin
FOCEI: 0.09 s → 0.09 s, below measurement noise). Fits where it does not run are unchanged.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 4833 passed, 0 failed)
- [ ] No new tests needed

**Tier 1 — `outer_optimizer_tests.rs`** (the selection rule): the lower objective wins in both
directions; a tie reports the cold solve (which is what keeps ordinary fits bit-identical); a
`NaN`/`inf` warm objective can never displace a finite cold one *and* a `NaN` cold one is still
displaceable — written as a positive comparison for exactly that reason; and the warning
threshold fires at a measured 6.96 gap, stays silent on a tie or a cold win, and is checked on
both sides of the tolerance.

**Tier 1 — `inner_optimizer_tests.rs`** (the seed): the covariate eta is seeded at `cov_obs − TV`
and the PK etas at zero, on a fixture whose `TV_WT` is moved off the pseudo-observation value
first — at the fixture default the two are equal, where a wrong answer is indistinguishable from
a right one. A no-FREM model gets plain zeros. And the call site the fix is actually about:
`find_ebe(..., max_iter = 1, ...)` forces the NM fallback and asserts the covariate eta stays at
18 — the existing FREM `CompiledModel` + `Subject` literal is extracted into a shared helper
rather than copied, since it is ~50 fields wide.

**Tier 2 — `tests/final_ebe_solve.rs`** (the wiring, new file): the disagreement is made a
property of the *configuration* rather than of one platform's libm. `inner_maxiter = 10` on
warfarin is a budget a warm start starting at the previous iteration's η̂ can meet and a cold
start from η = 0 cannot. The value is measured, not picked: at 1/2/3/5 the *fit* starves too
(OFV 242.5 / 212.4 / 196.5 / 188.9) and there is no gap, because a trajectory that never
converged has nothing a cold restart can fail to reproduce. At 10 the trajectory reaches
warfarin's usual −286.0042 while the cold re-solve lands at −31.06. The precondition assert
(the warning must be present) is what stops the OFV assertions being satisfied by the old
behaviour; the control test pins that an ordinary warfarin fit raises no warning at all.

**Mutation-checked**, each mutation killed by the test named:

| mutation | test that died |
|---|---|
| never re-solve warm (`cold_missed_the_incumbent = false`) | `a_fit_reports_the_objective_its_optimizer_reached` (warning absent) |
| FREM fallback seeds from `vec![0.0]` again | `the_frem_nelder_mead_fallback_restarts_from_the_data_implied_seed` (η = 0.0016 vs 18) |

- [x] Not applicable — no new `.ferx` syntax or option. The user-visible surface is one warning:

```
W_EBE_START_DEPENDENT: empirical Bayes estimates at the final parameters depend on the
inner loop's starting point — re-solving them cold scores 6.9463 OFV units worse than
re-solving them from the EBEs the optimizer minimised against (-167.9633 vs the reported
-174.9096). ...
```

## Docs
- [x] New pages linked in `docs/_quarto.yml` — no new pages

- `docs/estimation/foce.qmd` — new `### The final inner loop` under *Warm Starting*: the two
  solves, why the better one is reported, and that the cold number is what the convergence
  self-consistency check reads.
- `docs/warnings.qmd` — `ebe_start_dependent` in the convergence/covariance code table.
- `docs/estimation/frem.qmd` — a note under the FOCEI-vs-SAEM comparison table: its FOCEI row
  predates both fixes, the Linux arm now reaches the SAEM/IMPMAP objective, and the macOS arm
  still lands in a worse basin (#990). The row is left as the historical measurement rather
  than half-updated, since its ω's were already stale.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — not touched; both changes
      are on the `f64` EBE search and its seeding, and add no generic instantiation

### ferx-site
- [x] No new DSL syntax or `[fit_options]` key
- [x] No new data file

### ferx-book
- [x] No new estimator or DSL feature

### Example execution
- [x] `cargo test --lib` (mac, `--test-threads=1`): 4833 passed, 0 failed
- [x] `tests/frem_warfarin.rs` + `tests/final_ebe_solve.rs` on linux/arm64 with `slow-tests`:
      16 passed, 0 failed
- [x] No example render needed

## Reviewer hints

Focus on the final-solve block in `optimize_nlopt_once` and on `consistency_ofv`. Three things
worth checking:

1. **Which number the plateau check gets.** It has to be the cold one. If it ever gets the
   reported one, `failure_is_converged_plateau`'s consistency clause becomes a tautology and
   #751's guard silently stops guarding — no test of that guard would fail, since the inputs it
   sees would still look plausible.
2. **The `NaN` handling**, in both directions: a diverged warm solve must not win, and a
   diverged cold solve must not pin the fit to itself. Both are unit-tested, and both are the
   reason the comparisons are written as positive tests rather than negations.
3. **The trigger, not just the choice.** The second solve running unconditionally would also be
   correct — it would just cost every fit an extra inner loop and perturb nothing. The
   conditional version is what keeps ordinary fits bit-identical, so the thing to check is that
   its predicate cannot be false when the fit *is* start-dependent.

Skimmable: the `cold_eta_seed` extraction (a verbatim move of the #406 block into a function),
the test-fixture extraction in `inner_optimizer_tests.rs`, and the `WarningCode` table rows.

## Open questions

- `find_ebe`'s FREM arm is now the last call site that adopts a Nelder–Mead restart
  *unconditionally* (#378 fixed the non-IOV path, #1327 the IOV one, and `find_ebe_iov`'s
  comment already names this as the surviving exception). With the seed fixed it no longer
  destroys the restart, but it can still trade a better BFGS partial for a worse NM basin. I
  left it alone here — it changes FREM/IMP proposal centring broadly and neither issue needs
  it — but it is worth its own change.
- The macOS FOCEI arm of the FREM fixture lands 46 OFV units above the Linux arm, in a basin
  with ω²(V) and ω²(KA) inflated. That is #990's platform split, still live, now visible
  because both arms report honestly.

https://claude.ai/code/session_01YWvSs7GspxGh3UDcPFVGuv


roninsightrx (commented):

1. [P1] Preserve the incumbent objective as a third candidate (`src/estimation/outer_optimizer.rs:2842`). The final choice is only `min(cold rerun, warm rerun)`, not `min(best-seen, cold rerun, warm rerun)`. On the FREM path, `find_ebe` can still discard the BFGS partial and adopt a Nelder–Mead result unconditionally when `ebe_warm_start` is false (`src/estimation/inner_optimizer.rs:1088-1099`; the known #1365 behavior). Therefore `solve_at(Some(&seed))` is not monotone from the saved incumbent EBEs, and both reruns can remain above `best_seen_ofv`. That leaves the headline invariant — and #833 regression guard — unproven for the exact FREM fallback path this PR touches. Please either retain the full incumbent EBE/H/kappa state and include its scored result in the final selection, or make the warm rerun guaranteed not to worsen its seed.

2. [P1] Do not diagnose every gap as multimodality (`src/estimation/outer_optimizer.rs:2929`). The new regression fixture explicitly says its gap does not depend on a multimodal objective: the cold solve merely exhausts `inner_maxiter = 10` (`tests/final_ebe_solve.rs:22-32`). Yet the emitted warning states that the individual objective has more than one mode and the docs tell users the estimates are fine. The same predicate therefore converts an inner non-convergence/budget artifact into a false model diagnosis. Please describe this as start dependence under the configured inner solve, with multimodality and insufficient convergence as separate possibilities, or gate the stronger claim on convergence evidence.

3. [P2] A NaN cold solve is recovered silently (`src/estimation/outer_optimizer.rs:2922`). `cold.is_nan()` correctly forces the warm retry, and a finite warm result wins, but `ebe_start_dependence_gap(NaN, final_ofv)` returns `None` because the comparison is false. The reported fit is then finite, so `gate_converged_on_objective` also emits nothing. This is the strongest start-dependence case and should produce a structured warning; add an explicit non-finite-cold arm and a regression test.

4. [P3] Align the FOCE docs with the conditional implementation (`docs/estimation/foce.qmd:485`). The page says the final loop is run twice, while the code only runs the warm-seeded solve when the cold OFV materially misses `best_seen_ofv`; the PR description also advertises that conditional behavior and its zero-cost common path.

Validation performed: PR CI is green; local targeted tests for the selection/gap helpers, FREM fallback seed, and `tests/final_ebe_solve.rs` pass; `cargo fmt --all -- --check` and `git diff --check` pass. The local public-API script was blocked by sandboxed sccache, while the PR Public API baseline job is green.

roninsightrx:
Thanks — all four are right, and #1 was the one that mattered. Fixed in `b0c398c1`.

## 1 (P1) — the incumbent is now a scored candidate, not just a seed

Correct, and it undercut the headline invariant on exactly the path this PR touches:
`run_inner_loop_warm` is not monotone from its seed, so `min(cold, warm re-solve)` could sit
above `best_seen_ofv` while both looked fine.

The objective closure now publishes the incumbent's whole EBE state — `(η, H, κ)`, the triple
`pop_nll_opts` consumes — instead of the etas alone, and the final block **scores it** as its
own candidate. That costs one `pop_nll_opts` pass and no inner loop. Selection is
`reported_candidate`, the lowest finite objective among `[cold, warm re-solve, held]`, ties to
the earliest index so a fit whose candidates agree keeps reporting the cold number.

One thing your finding surfaced that I had to handle while implementing it: **scoring held
state is only sound at the point it was computed at**, because `pop_nll_opts` reads `h_mats` as
the curvature at the point it is scoring. Publication is gated on improving the same objective
the tracker ranks, so the two normally coincide — but `BestPoint::observe` also records
guard-penalised evals, whose EBEs are deliberately *not* adopted, so they can part. The
published state now carries its scaled `xs` and the held candidate is only scored when
`inc.xs == restored_xs`; otherwise it still seeds the re-solve. Without that check a stale
`log|H|` term could have won the comparison on a number that is not the objective at those
EBEs — a wrong reported OFV of a different flavour.

## 2 (P1) — the warning no longer diagnoses multimodality

Right, and the counter-example is my own fixture: `tests/final_ebe_solve.rs` produces its gap
from an `inner_maxiter` a cold start cannot converge within, not from a second mode. The
message now names both causes and how to separate them:

> Either the individual objective has more than one mode at these estimates, or the inner loop
> cannot reach it from a cold start within `inner_maxiter`; … Raising `inner_maxiter`, or
> `inner_restarts` for a suspected second mode, tells the two apart.

`WarningCode::EbeStartDependent`'s doc and `docs/warnings.qmd` say the same, and the unit test
asserts the message *does not* contain the old multimodality claim, so the wording cannot drift
back.

## 3 (P2) — a non-finite cold solve is reported

Right: `NaN - x > tol` is false, so the gap test declined both the retry and the warning in the
one case where the cold EBEs are certainly not the ones to report. There is now a
`ColdSolveVerdict::NonFinite` arm which drives both, with its own message ("re-solving …
returned a non-finite objective"). Tests cover `NaN`, `+inf` and `-inf` in both directions,
including a `classify_warning` round trip pinning both messages to the same code.

## 4 (P3) — docs match the conditional path

Rewritten: the final loop runs cold, and the other two candidates are scored only when that
cold solve fails to reproduce `best_seen_ofv`.

## Naming

`warm_solve_wins` / `ebe_start_dependence_gap` are gone, replaced by `reported_candidate`,
`cold_solve_verdict` / `ColdSolveVerdict` and `ebe_start_dependence_warning`. Their unit tests
moved with them and gained: the held-candidate-wins case (`[-100.0, -20.0, -174.91] → 2`), the
two-candidate shape for when the held state is declined, the non-finite arms in both
directions, and the all-non-finite case (index 0 comes back, so `gate_converged_on_objective`
demotes the fit rather than this function inventing a verdict).

## Verification

- `cargo test --lib` (mac, `--test-threads=1`): 4835 passed, 0 failed.
- `tools/preflight.sh`: `preflight OK — fmt check clippy rustdoc docs public-api`.
- linux/arm64 container, `slow-tests`: `frem_warfarin` 14/14 (FOCEI FREM OFV −220.7) and
  `final_ebe_solve` 2/2, unchanged by this round.

On your note about the local public-API script: the baseline was regenerated with
`tools/update-public-api.sh` in the first commit and is unchanged here — the only added item is
`WarningCode::EbeStartDependent`.

PR #1389, opened 2026-09-14
perf(estimation): seed FO curvature and reduce AGQ allocation pressure

hiddevandebeek:
## Why

FO/FOCEI warm-started each subject from its previous EBE, but dense BFGS restarted with a generic inverse Hessian on every outer evaluation. That discarded analytical curvature already available from the sensitivity path. Laplace/AGQ also repeated terminal curvature and quadrature setup work.

This applies analytical-Hessian seeding in the style of nlmixr2est while retaining ferx's existing fallbacks when curvature is unavailable or unsuitable.

## What changed

- Seed FO, FOCEI, GN, GN-hybrid, and multi-node AGQ inner BFGS solves from the analytical Gauss-Newton/Almquist eta Hessian.
- Seed one-node Laplace from its exact conditional eta Hessian for robustness and consistency with its integration correction.
- Build the common FO-family seed from the existing first-order eta-sensitivity provider and fuse its first gradient, avoiding both the full theta/mixed `Dual2` jet and a duplicate provider evaluation.
- Apply inverse metrics through Cholesky solves; non-positive-definite exact curvature falls back to the existing prior metric.
- Reuse the existing closed-form, event-driven, and ODE sensitivity wiring.
- Retain and reuse the exact terminal eta Hessian in Laplace/AGQ objective evaluation.
- Skip generic one-node Laplace grid construction, cache AGQ log weights, and store retained AGQ modes/scores contiguously.
- Add `FERX_NO_INNER_HESSIAN_SEED=1` as an internal A/B switch.

FOCE and FOCEI deliberately use the same inner seed: their conditional EBE objective is identical. Their method-specific Hessians remain distinct in the population marginal, where FOCE freezes residual variance at the population prediction and FOCEI includes eta-residual interaction and its log-determinant curvature.

## Alternatives considered

Replacing the first-order provider with second-order duals or finite differences over first derivatives was evaluated. The FO-family optimization seed is the positive Gauss-Newton metric and needs only `f` and `df/deta`, so the existing first-order provider is both correct and cheaper. Laplace is different: its defining correction uses the exact conditional Hessian, so this PR retains that exact seed behind an SPD gate as a robustness choice even though the measured stress benchmark is slower.

Full exact-Newton and compact eta-only Newton inner solvers were also tested. They reduced first-order gradient work but were 16.8% and 10.3% slower respectively, so they are not included.

## Cross-repo dependency

|---|---|---|---|

## Breaking changes

- [ ] Public Rust API
- [x] None

<details>

None.

</details>

## Numerical validation

- [x] Warfarin and large 40-subject/3,600-observation benchmarks complete successfully.
- [x] Compared against the same optimized binary with `FERX_NO_INNER_HESSIAN_SEED=1`.
- [x] Existing NONMEM anchors remain applicable because likelihoods and model equations are unchanged.
- [ ] Not applicable.

The exact Laplace seed reached OFV 2913.031242 versus 2913.057952 without the seed on the stress anchor (0.026710 lower). This is evidence for the intended robustness benefit, not a claim that every run improves its optimum.

## Performance

- [x] Faster overall.
- [x] Slower for exact Laplace seeding, retained for robustness.

Optimized binary, three threads, 10-subject/3-ETA warfarin dataset:

| Method | Before | After | Speedup | Inner gradient steps |
|---|---:|---:|---:|---:|
| closed-form FOCE | 0.264 s | 0.121 s | **2.18x** | 44,320 -> 13,592 (-69%) |
| closed-form FOCEI | 0.217 s | 0.134 s | **1.62x** | 37,659 -> 16,664 (-56%) |
| ODE FOCE | 1.373 s | 1.018 s | **1.35x** | 6,137 -> 4,932 (-20%) |

Large 40-subject, 3,600-observation, 4-ETA three-compartment dataset; optimized build, NLopt L-BFGS, three threads, interleaved medians:

| Method | Seed off | Seed on | Wall result | Other measured effect |
|---|---:|---:|---:|---|
| AGQ(3) | 100.299 s | 95.987 s | **1.045x / 4.30% lower** | allocations -0.0047%; requested bytes +0.26% |
| Laplace exact seed | 31.627 s | 33.288 s | **5.25% slower** | gradients 196,945 -> 202,703 (+2.92%); lower final OFV |
| Laplace exact seed, outer iteration 0 only | 6.602 s | 6.824 s | **3.37% slower** | gradients -1.44%; allocations -1.27%; requested bytes -1.83% |

The complete quadrature bundle—terminal-Hessian reuse, the Laplace one-node shortcut, cached AGQ log weights, and contiguous retained modes/scores—measured AGQ(3) at 35.341 -> 33.896 seconds (**1.043x / 4.1% lower wall**) with 74,520 fewer allocation calls and 1.79 MB fewer requested bytes. Absolute timings differ because host contention differed between benchmark sessions; comparisons were interleaved within each run.

## Tests

- [x] Unit tests cover Cholesky application and bitwise-identical terminal-Hessian reuse.
- [x] Unit tests compare the light FO-family seed against the full Gauss-Newton Hessian and the exact Laplace seed against the full conditional Hessian.
- [x] `cargo test --lib`: 4,907 passed, 26 ignored.
- [x] `tools/preflight.sh` passed before the final focused Laplace admission change; formatting and the full library suite were rerun afterward.
- [ ] No new tests needed.

- [ ] Example added.
- [x] Not applicable: internal performance/robustness change with no new API.

## Docs

- [ ] `docs/` updated for user-visible changes.
- [ ] New pages linked in `docs/_quarto.yml`.
- [x] `CHANGELOG.md` updated.
- [x] No user-visible configuration or result-format change.

## Checklist

- [x] `tools/preflight.sh` green before the final focused commit; final `cargo fmt --check`, `git diff --check`, targeted regression, and full library suite are green.
- [x] Analytical paths reuse existing first- and second-order sensitivity wiring for closed-form, event-driven, and ODE models.
- [x] No version bump required.

### ferx-site

- [x] No new DSL syntax, example, or data file.

### ferx-book

- [x] No new estimator, DSL feature, or fit option.

### Example execution

- [x] No example execution step needed: internal change with no user-visible API.

## Reviewer hints

The central invariant is that method identity must not be confused with the inner conditional objective. FOCE and FOCEI share the correct positive conditional Gauss-Newton seed because they solve the same EBE problem; their different interaction Hessians enter later in the population marginal. One-node Laplace selects the exact conditional Hessian instead. `init_h_inv` admits either metric only through a Cholesky factorization and otherwise preserves the old prior-based initialization.

In `agq.rs`, retained terminal Hessians are reused only at the matching EBE. Stage admission is in `api/fit.rs`; seed construction and its SPD fallback are in `inner_optimizer.rs`.

## Open questions

None.


hiddevandebeek:
Follow-up optimization bundle added in f370efe:

- Laplace(1) objective-only evaluations now use the exact algebraic one-node reduction instead of materializing a generic tensor grid.
- Cached Gauss-Hermite rules now retain log weights too.
- AGQ retained node modes are one row-major allocation instead of one Vec per node; parallel score storage is likewise contiguous while preserving the former node-major accumulation order.

Large-data benchmark (optimized ci-test profile, 3 threads, 40 subjects, 3,600 observations, 4 ETAs, AGQ(3), 22 outer evaluations):

| | PR head before bundle | f370efe |
|---|---:|---:|
| Median wall time (2 runs) | 35.34 s | 33.90 s |
| Speedup | — | 1.043x (4.1% less time) |
| Allocation calls | 51,813,212 | 51,738,692 (-74,520; -0.144%) |
| Requested bytes | 18,013,199,377 | 18,011,410,897 (-1,788,480) |
| Final OFV | 2914.111002976775 | 2914.111002976775 |

Laplace specialization alone is deliberately reported as an allocation micro-win, not a timing claim: it removed 203 allocations / 5,450 requested bytes over a 23-evaluation large-data fit, while paired wall times (24.42 s enabled, 23.36 s disabled) were within/noisier than the expected effect.

Rejected experiments:

- Persisting the prior accepted subject curvature into the next Laplace inner solve regressed allocation calls 19.80M → 21.13M (+6.7%) and requested bytes 11.46GB → 12.48GB (+8.9%), without a reliable wall-time gain. It is not in the PR. This differs from the admitted FO-family seed, which recomputes curvature at the current population parameters/current warm EBE.
- NLopt L-BFGS exposes no initial inverse-Hessian/preconditioner hook. A direct outer Hessian seed therefore cannot be added without replacing the default optimizer; no unquantified solver swap is included.

Validation after the bundle:

- Focused one-node Laplace/AGQ, cached/uncached population AGQ, derivative-path, and cached-rule tests pass.
- Full lib suite: 4,648 passed, 26 ignored, 2 cross-worker reproducibility flakes; both failures pass immediately alone.
- tools/preflight.sh: green (all checks/features, clippy, rustdoc, docs lint, public API).

hiddevandebeek:
Laplace exact-Hessian seed follow-up (rebased large 40-subject / 3,600-observation / 4-ETA anchor, optimized build, 3 threads): I tested seeding dense BFGS with the exact conditional eta Hessian (`h_inner`) at the current EBE, deriving the first gradient from the same second-order sensitivity result and retaining the existing Cholesky/fallback gate. It is not admitted: full-fit median wall time was 31.627 s seed-off vs 33.288 s exact-seed-on (5.25% slower), while analytic inner-gradient calls increased 196,945 -> 202,703 (+2.92%). An isolated `outer_maxiter=0` evaluation showed that exact seeding itself reduced gradients 59,490 -> 58,633 (-1.44%), allocation calls 3,457,517 -> 3,413,543 (-1.27%), and requested bytes 2.398 GB -> 2.354 GB (-1.83%), but the extra current-point `Dual2` Hessian pass outweighed those savings: median wall time 6.602 s -> 6.824 s (3.37% slower). The experiment was reverted; Laplace keeps the profitable terminal exact-Hessian reuse and one-node shortcut, while initial seeding remains FO/FOCEI/multi-node-AGQ only.

hiddevandebeek:
Follow-up decision after the earlier Laplace A/B: exact Laplace Hessian seeding is now intentionally retained for robustness in c229cc46. It is admitted only when Cholesky proves the conditional Hessian SPD; otherwise the previous prior-based metric is used. The large stress anchor measured 31.627 -> 33.288 s (5.25% slower) and 196,945 -> 202,703 gradient calls (+2.92%), while reaching OFV 2913.031242 versus 2913.057952 (0.026710 lower). An outer-iteration-0 isolation also showed lower gradients (-1.44%), allocations (-1.27%), and requested bytes (-1.83%) despite 3.37% higher wall time. This supersedes the earlier decision to exclude the seed.

FOCE and FOCEI continue to use the same positive conditional Gauss-Newton seed by design: their EBE objective is the same. Their respective correct Hessians diverge in the outer marginal—FOCE uses non-interaction residual curvature, while FOCEI carries eta-residual interaction and log-determinant curvature. The code and PR description now state this invariant explicitly.

Validation on c229cc46: cargo fmt --check and git diff --check passed; cargo test --lib passed 4,907 tests with 26 ignored.

hiddevandebeek:
Review of the diff (`4160374f..c229cc46`), one careful pass. Findings ordered by severity; each names the concrete scenario in which the code misbehaves.

### High

**1. `src/estimation/inner_optimizer.rs:617` (`analytic_terminal_work`) / `src/estimation/agq.rs:780` — the reused terminal Hessian is not the Hessian `anchor_hessian(Exact)` would compute, and the gap is exactly the models Laplace exists for.**
`anchor_hessian(Exact)` gates on `analytic_score_supported(model)` (= `analytic_outer_gradient_available`, which is `false` for `has_non_gaussian()` models and for `gradient_method = fd`) and otherwise uses `fd_posterior_hessian` of the *full* conditional NLL. `analytic_terminal_work` has no such gate: it calls `subject_sensitivities` + `score_core` directly, and neither declines a TTE / binary / count / CTMM model (the providers only cover the structural PK part; `score_core` folds no survival term). So on a joint PK-TTE Laplace fit, every subject with both Gaussian observations and event records gets `terminal_hessian = Ω⁻¹ + Gaussian curvature only`, and `agq_subject_evaluate` (objective-only path, `Exact`) uses it verbatim instead of the FD Hessian that includes the hazard curvature. The `0.5·log|H|` Laplace correction is then wrong for precisely those subjects. Since such models take the FD/BOBYQA outer (objective-only evaluations), this is the objective the whole fit minimises, while the final reported OFV (via `agq_population_nll`, which passes `None`) and the gradient path (`agq_population_evaluate_impl`, also `None`) use the FD Hessian — so the minimised objective and the reported one disagree. Same mechanism, smaller magnitude, for `gradient_method = fd` (user opt-out silently overridden). The new `agq.rs:4653` test cannot see this: it builds `terminal_hessians` from `anchor_hessian` itself rather than from `analytic_terminal_work`, so it asserts `f(x) == f(x)`. Fix: gate `analytic_terminal_work` on the same predicate `anchor_hessian` uses (`analytic_score_supported(model)`), and make the test feed the Hessians `find_ebe` actually produces.

### Medium

**2. `src/estimation/inner_optimizer.rs:535` — fused first gradient uses the censored kernel on CENS rows even when `bloq_method = drop`.**
`analytic_eta_nll_gradient_with_schedule` (what `agrad` returns on every other step, and what `obj` scores) applies `cens` only when `m3` is set: `let cens = if m3 { … } else { 0 }`. The seed path passes the raw `subject.cens[j]` into `residual_inner_obs` unconditionally, while its own `p` correctly uses the Gaussian `err_terms` when `!m3`. With the default `bloq_method = drop` and a dataset that has a `CENS` column (e.g. the BLOQ example run without `m3`), the first gradient BFGS sees is the gradient of a *different* objective (−logΦ rows vs Gaussian rows). Consequences: a warm-started η̂ already at the mode fails `gnorm < tol` on the fused gradient and takes a non-descent step (line search cannot decrease `obj` → `alpha == 0` → `false` → NM fallback, `n_fallback` inflates); the first BFGS `y = g₁ − g₀` is corrupted. Mirror the ordinary path: `let cens = if m3 { … } else { 0 }`. The `assert_eq!(light_gradient, ordinary_gradient)` test passes only because its fixture has no CENS rows.

**3. `src/estimation/inner_optimizer.rs:587` — fused first gradient omits the CTMM term (`markov` feature).**
`analytic_eta_nll_gradient_with_schedule` adds `ctmm_subject_eta_grad` to the Gaussian block; `light_supported` does not exclude `model.has_ctmm()`, and `analytic_inner_grad_supported` is `true` for a fully-analytic CTMM subject, so `use_analytic` is true and the fused gradient (Gaussian + prior only) is handed to BFGS as the first gradient. Same failure mode as (2) for a joint PK + CTMM subject. Add `has_ctmm()` to the `light_supported` exclusion (or fold the CTMM gradient in).

**4. `src/estimation/inner_optimizer.rs:3373` — the "fall back to the historical metric" promise is not kept for a non-SPD seed.**
`init_h_inv` falls back to identity when Cholesky fails, but the `1/gnorm` first-step scaling is skipped whenever `hessian_seed.is_some()`, not whenever the seed was *applied*. For the Laplace exact seed (`h_inner` is routinely indefinite away from the mode) and any GN seed that fails Cholesky, the first step is now `−g` at full length — for proportional-error subjects with `gnorm ~ 1e2–1e4` that is up to ~13 extra backtracking `obj` evaluations per inner solve, and a different BFGS path than the baseline the PR says it preserves. Have `init_h_inv` report whether the seed was accepted and gate the scaling on that.

**5. `src/api/fit.rs:1635` / `src/estimation/agq.rs:1305` — terminal-Hessian capture costs a full second-order jet per subject per outer evaluation, and mostly buys nothing.**
(a) Capture is enabled for every quadrature stage, but the reuse fires only under `HessianAnchor::Exact`. FOCEI with `n_agq > 1` (GaussNewton anchor) now pays `subject_sensitivities` (θ + η + second order) where it previously paid the light `subject_eta_jacobian`, and the resulting Hessian is dropped. (b) `agq_population_evaluate_impl` (the gradient path, i.e. every NLopt L-BFGS evaluation on an in-scope Laplace model) passes `terminal_hessian: None` and recomputes `score_core_at` in `anchor_hessian_and_base_jet`, so for the default analytic Laplace configuration the capture is computed twice and reused zero times. The reuse is live only on the objective-only/BOBYQA path — which is where finding (1) makes it wrong. Gate capture on `hessian_anchor() == Exact`, and either thread the captured `(h, jet)` into the gradient path or drop the capture there. This likely accounts for part of the measured Laplace slowdown.

### Low

**6. `src/estimation/inner_optimizer.rs:654` — three new process-global flags are set per stage and never cleared.**
This follows the existing `set_inner_optimizer` pattern, but the cost profile is new: after any Laplace/AGQ fit, `CAPTURE_TERMINAL_HESSIAN` stays `true` for every later `find_ebe` in the process (`predict()` posthoc, VPC, ferx-tools bootstrap fits that do not pass through the stage loop, ferx-r sessions), each paying a full `Dual2` jet per subject for a Hessian nobody reads; and the terminal `analytic_jac` path now bypasses `analytic_inner_grad_supported` (so `gradient_method = fd` no longer opts those callers out) and drops the `is_finite` filter the old path had. In the unit-test binary, tests calling `fit()` with different methods run concurrently and race on `EXACT_HESSIAN_SEED_FOR_FIT` / `HESSIAN_SEED_FOR_FIT`, which can make bit-identity tests such as `inner_restarts_bit_identical_on_wellidentified_subject` (two `find_ebe` calls that must share a seed mode) flaky. Consider disarming at the end of `fit_inner` (the `#1212` guard pattern already used for ODE solver options).

**7. `src/api/fit.rs:1647` — seed is not applied to multi-node Laplace-anchored AGQ, contrary to the description.**
`method = laplace, n_agq > 1` yields `quadrature_nodes = Some(n>1)` and `method ∉ {Foce, FoceI, FoceGn, FoceGnHybrid}`, so `set_hessian_seed_for_fit(false)`. Only FOCEI-AGQ is seeded. Either intended (then update the description/CHANGELOG) or the condition should include `EstimationMethod::Laplace`.

**8. `src/estimation/inner_optimizer.rs:1290` — the seed is computed even when it cannot be used.**
Under `inner_optimizer = nelder_mead`, or when `inner_use_lbfgs(n)` routes to `lbfgs_core` (which takes no `hessian_seed`), `analytic_inner_seed_hessian` still runs. On the exact (Laplace) branch that is a full `subject_sensitivities` + `score_core` per subject per outer evaluation for nothing (no fused gradient is returned either). Cheap fix: skip the seed when `inner_optimizer_mode()` is NelderMead or `inner_use_lbfgs(n)`.

Minor: `#[allow(dead_code)]` on `run_inner_loop_warm_cached` and `agq_population_nll_with_schedules` hides that both are now unused — delete them or route the non-cached AGQ branch through the new `_and_hessians` variant so `terminal_hessians` are not cloned and then discarded there.

hiddevandebeek:
Follow-up to the review above — all findings addressed on this branch. Summary of what changed and why, finding by finding.

### 1 (High) — reused terminal Hessian ≠ `anchor_hessian(Exact)` outside the analytic score scope
`analytic_terminal_work` now gates on the same model-level predicate the `Exact` arm of `anchor_hessian` uses (`agq::analytic_score_supported`). A joint PK‑TTE / discrete / CTMM model, or `gradient = fd`, returns `None` and the objective falls back to the FD anchor it always used. The `agq.rs` reuse test now feeds the Hessians `analytic_terminal_work` *produces* (the production path), asserts each is bit‑equal to `anchor_hessian`, and only then asserts the objective is unchanged — it no longer compares `f(x) == f(x)`. `analytic_terminal_work_declines_outside_the_exact_anchors_scope` pins the decline for `gradient = fd` and for a joint PK‑TTE model (a subject with no event records, so the provider itself would happily serve it — only the gate says no). Mutation: removing the gate turns that test red.

### 2 (Medium) — fused first gradient used the censored kernel under `bloq_method = drop`
Mirrors the ordinary gradient: `cens = if m3 {…} else {0}`. `light_seed_gradient_applies_cens_only_under_m3` asserts bit‑identity with `analytic_eta_nll_gradient` on a fixture with a live CENS row, under both `drop` and `m3`, and asserts the two arms differ so neither assertion is vacuous. Mutation: red.

### 3 (Medium) — fused gradient omitted the CTMM term
The light path declines a `has_ctmm()` model (under `markov`) and falls to the full provider, whose seed carries no fused gradient, so BFGS takes its first gradient from the ordinary path.

### 4 (Medium) — `1/gnorm` first‑step scaling keyed on seed *supplied*, not *accepted*
Split `seed_h_inv` (Cholesky; `None` when the seed is absent, mis‑shaped, indefinite or non‑finite) from `init_h_inv`, and key both the metric and the first‑step scaling on the accepted inverse. `rejected_hessian_seed_reproduces_the_unseeded_solve` asserts an indefinite seed gives the *same iterates and the same objective‑evaluation count* as no seed (start with ‖g‖≈16 so the scaling is live), with an accepted‑seed control showing the equality is not vacuous. Mutation: red.

### 5 (perf) — capture on every quadrature stage, reused only on the objective‑only `Exact` path
Capture is no longer a process‑global: `run_inner_loop_and_nll_prepared` asks for it per call, only when `agq_gradient_inputs.is_none() && hessian_anchor() == Exact`. FOCEI‑AGQ no longer pays the second‑order jet, and the gradient path (which recomputes the jet for the Laplace derivative sweep regardless) no longer captures a Hessian it throws away. I did **not** thread the captured jet into the gradient path — that would mean retaining a `SubjectSens` per subject across the population barrier, which is a separate design step; noted as a possible follow‑up.

### 6 (Low) — process‑global flags never cleared, racing across concurrent fits
Removed all three globals. The seed is now `InnerHessianSeed::for_options(&FitOptions)` — a pure function of the stage's method/`n_agq` — threaded explicitly via `InnerSolvePolicy` through `run_inner_loop_warm_map[_cached]` / `find_ebe_cached`; the public `find_ebe` / `run_inner_loop_warm` are the unseeded solve, and the estimation stages call `run_inner_loop_warm_seeded`. This was not hypothetical: with the globals, `api::ode_solver_diagnostics_tests::the_parallel_post_fit_pass_is_worker_count_independent` and `run_covariance_matches_inline_covariance` went red in a parallel `--lib` run (two sequential fits compared bit‑for‑bit while other tests' fits flipped the flag; 2e‑3 covariance divergence). Both are green now, and the standalone `run_covariance` reconverges with the fit's seed so it stays bit‑identical to the inline step (measured 1.5e‑11 divergence against the 1e‑12 bound without that). The terminal‑Jacobian path also keeps its `analytic_inner_grad_supported` gate and `is_finite` filter regardless of capture.

### 7 (Low) — `laplace, n_agq > 1` not seeded
Now seeded with the Gauss–Newton metric (any multi‑node quadrature, either anchor); `inner_hessian_seed_is_a_pure_function_of_the_stage_options` pins the full mapping. CHANGELOG updated.

### 8 (Low, perf) — seed computed when Nelder–Mead / L‑BFGS cannot use it
Skipped when `inner_optimizer_mode()` is Nelder–Mead or `inner_use_lbfgs(n_eta)` routes away from dense BFGS.

### Minor
`run_inner_loop_warm_cached` and `agq_population_nll_with_schedules[_and_hessians]` are gone; the outer loop calls one `agq_population_nll_prepared(schedules: Option, terminal_hessians: Option)` on both the cached and uncached arms, so the non‑cached branch no longer clones and discards the Hessians.

### Verification (local, `--features survival`, this box)
- `cargo test --lib`: **5107 passed, 0 failed, 26 ignored** on the final tree. (An intermediate run had three reds: the §6 race, and the two `pipeline::fused_*` tests, whose *unseeded* reference pass was being compared against the now‑seeded fused pass — they now run the same seed, so what they compare is fusion alone; for Laplace that includes the retained‑Hessian reuse against `pop_nll_opts`'s from‑scratch anchor.)
- Mutation: one build with fixes 1, 2 and 4 all reverted (tests kept) turned exactly those three tests red, each at the assertion that names its own fix; the three code paths are independent (scope gate, `cens` rule, first‑step scaling), so each test is pinned to its own mutation.
- `tools/preflight.sh` (full run): `preflight OK — fmt check clippy rustdoc docs public-api`.

Commit: ebbad6a0. CI is the authority for the `ci`/`markov`/`nn` feature sets; this box only builds `--features survival` locally.

Issue #1326, opened 2026-09-10
bug(ci): preflight's default run omits `release-semantics`, which has a CI job — the local gate does not cover a gate CI enforces


TeunP:
## The gap

`tools/preflight.sh` with no arguments runs `fmt check clippy rustdoc docs public-api`.
`release-semantics` and `debug-assertions` are in `OPT_IN_GROUPS` (`tools/preflight.sh:92`)
and are therefore **not** in that run.

But `release-semantics` **has a CI job**, and that job's only step is the group itself:

```yaml
# .github/workflows/ci.yml:185
    name: Tests (release semantics)
...
      - name: cargo test --lib with debug-assertions OFF (via tools/preflight.sh)
        run: tools/preflight.sh release-semantics
```

So a green local `tools/preflight.sh` does not cover a gate CI enforces. That contradicts
the premise the section in `CLAUDE.md` rests on:

> `.github/workflows/ci.yml` invokes this same script for its `Check`, `Clippy`, `Format`,
> `Rustdoc` and `Docs lint` jobs, and `.githooks/pre-commit` calls its `fmt` group, so the
> local gate, the hook and the CI gate cannot drift

`Tests (release semantics)` is missing from that list, and the drift it denies is exactly
what the opt-in status creates. `debug-assertions` is fine as opt-in — #1248 retired its
job, so it has no CI counterpart.

## How it bit

PR #1309. A `debug_assert!` was added to `clamp_to_bounds` with a test asserting it under
`#[should_panic(expected = "E_INIT_BOUNDS_INVERTED")]`. `release-semantics` builds
`--profile ci-fast` with `debug_assertions = false`, where the assertion is compiled out and
`f64::clamp` panics with `core`'s own message instead — a real panic, but not that string:

```
panicked at core/src/num/f64.rs:1643
note: panic did not contain expected string
```

A no-argument `tools/preflight.sh` was green, the full `--workspace --features ferx-core/ci
--lib --tests` suite was green at 207 binaries / 6050 tests, and `Tests (release semantics)`
went red on push. Fixed in `29131c14` (now `9bd96055`) by asserting across both profiles
rather than pinning one message — but the local gate should have caught it.

## Options, in increasing cost

The fix is a judgement call about local gate time, so no PR is opened with this issue:

1. **Say so, and stop claiming otherwise** (cheap floor). Add `Tests (release semantics)` to
   the CLAUDE.md list; have the no-argument run print a closing line naming the opt-in
[...]

Issue #1076, opened 2026-08-25
bug(sens): two moving boundaries at the same instant each inject the whole covariate field jump, double-counting it


TeunP:
Measured while responding to the #1071 review. Pre-existing — bit-identical on `fce3e39e` (pre-review) and on the current branch — and distinct from #1060's class.

## Reproducer

Appending this to `src/sens/ode_provider_tests.rs` fails on **both** sides of #1071:

```rust
#[test]
fn probe_coending_comoving_infusion() {
    let model = parse_model_string(ONECPT_IV_LAG_CROSS_ODE).expect("parse");
    let mut subject = lag_crossing_subject();
    // Infusion at t=0, amt 100 @ rate 100 -> 1 h window. Both doses ride the same bare
    // LAGTIME, so the window END (`lag + 1`) and the bolus ARRIVAL (`1 + lag`) coincide
    // exactly, and both move together.
    subject.doses = vec![
        DoseEvent::new(0.0, 100.0, 1, 100.0, false, 0.0),
        DoseEvent::new(1.0, 100.0, 1, 0.0, false, 0.0),
    ];
    let wt = |w: f64| HashMap::from([("WT".to_string(), w)]);
    subject.dose_covariates = vec![wt(70.0), wt(70.0)];
    let theta = vec![10.0, 50.0, 0.75, 0.7];
    let eta: Vec<f64> = vec![0.1, -0.05, 0.07];
    assert!(ode_tvcov_supported(&model, &subject));
    check_vs_production(&model, &subject, &theta, &eta);
}
```

```
assert_relative_eq!(obs.df_deta[k], g, max_relative = 2e-3, epsilon = 1e-6)
    left  = 0.7442389245291534   # analytic
    right = 0.7027702373374467   # central FD of compute_predictions_with_tv
```

**5.9 % on the gradient** — first order, so it moves the FOCEI objective, not only the Hessian. Bit-identical before and after #1071, which does not touch this composition.

## Why

`K_DOSE` (kind 2) and `K_INF_END` (kind 6) land on the same instant `t*` and are processed in that order. Each is a moving boundary in the same `δlag`, and each independently reads a pre-side and a post-side snapshot:

- the bolus arrival injects `(g(u⁻; P_pre) − g(u⁺; P_post))·δ`
- the rate-off then injects `(g(u⁺; P_pre) + r − g(u⁺; P_post))·δ`

[...]

PR #1391, opened 2026-09-14
fix(sens): a moving boundary reads one covariate snapshot [closes #1068]

### LINKED ISSUE #1068: bug(sens): a lagged arrival landing exactly on a covariate-changing record gives a 1% eta-gradient error (ODE event-driven walk)

roninsightrx:
## Why

Epic #1350 row 19 asked for **one re-measurement at HEAD** of the moving-boundary saltation cluster (#1076, #1075, #1068, #1069, #1072, #1160) to decide whether any of it is release-relevant. Measured at `257be39a`. Four of the six no longer reproduce; one is live and unchanged in kind; and the re-measurement turned up a **first-order** defect in the one everybody had recorded as a benign convention.

### The re-measurement

| issue | filed | at `257be39a` | verdict |
|---|---|---|---|
| #1076 two colliding moving boundaries double-count the field jump | 5.9 % gradient | **4.6e-9** on the issue's verbatim reproducer | not reachable |
| #1069 input-rate onset carries the rate jump, not the field jump | never measured | **5.9e-9** grad / **6.5e-7** Hess on the `init(...)` + lagged `first_order` fixture the issue asked for | not reachable |
| #1072 value-only `pk_snapshot_equal` under IOV | Hessian sign flip | rate-off sites: **118/118** boundary events have pre and post as the *same slice* — jets identical | vacuous |
| #1160 general rate-off `else` arm unreachable | dead code | reachable only by a record at `t_ev + δ`, `0 < δ < 1e-12` (δ=5e-13 → `equal=false`, δ=1e-12 → `equal=true`) | its final hypothesis confirmed |
| #1075 missing `∂f/∂t` in `ẍ` | 6.4e-2 Hessian | filed geometry **7.4e-7**; bare `TAD`-reading RHS still **2.9e-1** on `∂²f/∂η_LAG²` | **live**, 2nd order, untouched here |
| #1068 arrival on a covariate record | ~1 % gradient, "a convention" | see below | **live**, 1st order — fixed here |

Four of them died for **one** reason: #1073's end-of-interval convention makes the two sides of a boundary interior to a record interval resolve to the *same record*. Instrumented over this module's fixtures — 310 moving-boundary events — **none** parted. `K_DOSE`=3 / `K_ROUTE_ONSET`=4 sort *before* records (5/6) and `K_INF_END`=7 / `K_ZO_END`=8 sort *after* them, which is why only an arrival can part its two sides at all.

### The defect

#1068 was recorded as a judgement call: the value is continuous, only the gradient is one-sided, and since #1071 "ferx returns the one-sided derivative of the branch its own event ordering uses". That is true **only when the compartment is empty at the arrival**. One variable changed, everything else held:

```
empty compartment (single dose)
  obs4: analytic +0.1373777474 | fwd-lim +0.1373777658 | bwd-lim +0.1346573142  -> on the fwd limit
residual drug (prior bolus at t=0)
  obs4: analytic +0.2413982513 | fwd-lim +0.2392525367 | bwd-lim +0.2387074260  -> 0.89% OUTSIDE BOTH
```

0.89 % at every observation after the arrival, **first order**, so it reaches the FOCEI objective and not only the Hessian. A value outside `[bwd, fwd]` is not a subgradient — it is not a valid derivative of anything. Limits are Richardson-extrapolated one-sided differences, stable to 9 significant figures over `h` from 1e-4 to 1e-8.

This is exactly the non-degeneracy trap CLAUDE.md warns about: `g(x⁻) = 0` on a single-dose fixture multiplies the spurious term away, and every fixture in `sens/` reaching this code had that shape.

## What changed

A lagged arrival is a **moving** boundary; a covariate record's field jump is **stationary**. As `lag` sweeps the arrival through `t_rec` the dose boundary moves and the record's jump stays put, so it owes no `δlag` term. The walk was attributing it to the saltation anyway, by reading the pre-side velocity from the co-timed record (`params`, which sorts at index ≥ the arrival) and the post-side one from a forward scan for the first *strictly later* record.

Both sides now read `params`, at all three arrival arms — the bolus, the lagged-infusion rate-on, and the built-in absorption onset. That scan was a no-op wherever it was right (since #1073 `params` already IS #1060's post side for an interior boundary) and wrong wherever it was not, so removing it also deletes three now-provably-dead fallbacks: an owned `prep_for` for the post side, a second lifted `Dual1` parameter vector, and a slice-identity test guarding them.

The walk now returns a genuine one-sided derivative, and which one follows from where each boundary sorts against a co-timed record — an arrival reports the limit from below, a rate-off the limit from above. Both directions are pinned.

## Alternatives considered

Taking `post_params` on both sides instead. Measured: equally exact, landing on the **forward** limit to 9 figures in both geometries, suite equally green. Rejected because it contradicts ferx's own event ordering (`K_DOSE` sorts before a co-timed record), which would mean changing the convention `docs/estimation/foce.qmd` blesses rather than making it true. Both single-snapshot choices fix the defect; the *mixing* is what was wrong.

Declining to FD (#1068's option 1, matching `moving_bounds_separable` on the closed-form engine) was not pursued: central FD at this coincidence returns the mean of two different one-sided limits, so declining would replace a correct one-sided derivative with a value that is neither — and it costs the `Option` plumbing across four entry points that #451 re-review #12 removed.

## Cross-repo dependency

## Breaking changes
- [x] None

No public API change, no output field change. The analytic `∂f/∂η_lag` at a record-co-timed arrival changes — that is the fix — and in the empty-compartment case it moves from the forward one-sided limit to the backward one (both valid; the change is that all arms now report the same branch). Predictions are bit-identical: the injection is jet-only.

## Numerical validation
- [x] Compared against: **Richardson-extrapolated one-sided derivatives of the production predictor**, plus double-FD of the predictor and a central difference of the analytic first-order gradient for the second-order blocks
- [x] Reference values:

```
# before  (analytic df/deta_lag, arrival exactly on the t=1.5 covariate record)
empty compartment   obs3 +0.2296188771  obs4 +0.1373777474  obs5 +0.0468790154
residual drug       obs3 +0.4034830708  obs4 +0.2413982513  obs5 +0.0823751484
                    -> 0.89% outside both one-sided limits on every row

# after
empty compartment   obs3 +0.2250718519  obs4 +0.1346573262  obs5 +0.0459506942
residual drug       obs3 +0.3989855295  obs4 +0.2387074355  obs5 +0.0814569299
                    -> equals the backward one-sided limit to 8-9 significant figures
                       (worst miss 1.2e-8, against a fwd-bwd gap of 2.3e-3 to 4.5e-3)
```

**No NONMEM anchor, and the reason is the same class the CLAUDE.md exceptions describe.** The quantity is `lim_{ε→0⁻} ∂f/∂η_lag` at an arrival that coincides exactly with a data record — a one-sided derivative at a point where no two-sided derivative exists. NONMEM reports no analytic η-gradient to compare, and its own finite-difference gradient there would return the mean of the two branches, which is the thing under test rather than a reference for it. Validated instead with (a) **exact one-sided limits of ferx's own production predictor**, Richardson-extrapolated and swept over `h` to show the limits themselves are resolved; (b) a **differential pair** — empty compartment vs residual drug at the arrival, one variable changed — which is what separates a genuine convention from the defect; and (c) the **interior control**, the same geometry with the arrival off the record, where both one-sided limits coincide and `check_vs_production` / `check_hessian_vs_production_fd` apply normally and are asserted.

## Performance
- [x] Faster — one fewer `prep_for` (a full `PreparedInputRate` build) and one fewer lifted `Dual1` parameter vector per lagged dose per provider evaluation, plus the removed forward timeline scan. Not benchmarked; the removed work is unconditional allocation on the arrival path, so the sign is not in question even though the magnitude is small beside the segment integration.

## Tests

Six tests in `src/sens/ode_provider_tests.rs`, plus a `check_hessian_vs_production_fd_after` helper for fixtures where an *earlier* observation sits on a boundary and a central stencil is not an oracle there.

**Mutation results, per arm.** Only one arrival arm runs per dose, so a single fixture proves nothing about the other two — and this was measured, not assumed:

| mutation (restore the pre-#1068 lookahead at…) | before the per-arm fixtures | now |
|---|---|---|
| all three arms | bolus test red | bolus test red |
| bolus arm only | bolus test red | bolus + infusion tests red |
| infusion rate-on arm only | **whole `sens::` suite green** | infusion test red |
| onset arm only | **whole `sens::` suite green** | onset test red |

The two green rows are why `..._lagged_infusion_rate_on_...` and `..._input_rate_onset_on_...` exist.

Each coincidence fixture also asserts **the straddle itself** (`fwd ≠ bwd`), so it cannot silently decay into a tautology if the fixture ever stops putting a record on the boundary; each asserts non-degeneracy of the incoming side (drug actually present, or the `init(...)` baseline actually reaching *past the onset* — read at the first observation **after** the boundary, since a pre-boundary row is `BASE/V` for any non-zero `BASE` and cannot fail); and every folded comparison checks `is_finite()` before folding.

All four coincidence fixtures also pin **`∂²f/∂η²`** against the one-sided limit on the branch the boundary sorts to (`one_sided_eta_second_limits`, `(2f₀ − 5f₁ + 4f₂ − f₃)/h²` and its mirror). The fix moves `J⁺` and the `δlag²` coefficient as well as the first-order jump, and a central Hessian stencil is no more usable at a coincidence than a central gradient stencil is, so nothing was pinning the curvature there. Mutation evidence: `coef2 *= 1.01` in the bolus arm reddens the bolus fixture at the new assertion (`0.252607` vs `0.252334`), and the same mutation on `inject_rate_saltation`'s `coef2` reddens the other three — **neither mutation is visible to any first-order assertion**.

| fixture | `h` | worst realised miss | `fwd − bwd` straddle | bound |
|---|---|---|---|---|
| bolus arrival | 1e-3 | 1.624e-6 | 2.6e-3 – 2.2e-2 | 1e-5 |
| lagged rate-on | 1e-3 | 1.662e-6 | 7.4e-3 – 7.1e-2 | 1e-5 |
| shared `first_order` onset | 5e-3 | 1.151e-3 (abs 9.0e-6) | 0.65 – 0.99 | 5e-3, eps 1e-4 |
| route onset | 5e-3 | 4.190e-4 | 7.1e-3 – 3.2e-2 | 2e-3 |

Two helpers gained a `first` parameter rather than a second copy. `check_hessian_vs_production_fd_after` had been a hand-copied η stencil that silently omitted the `∂²f/∂η∂θ` block its full sibling checks, so the infusion-end fixture was verifying no mixed second derivative at all; the two are now one stencil, and a `coef2` mutation reddens that block at `ode_provider_tests.rs:843`. `check_vs_production_after` exists because dropping `check_vs_production` wholesale at a kinked observation left `∂f/∂θ` and the non-lag etas unchecked at *every* row rather than only the kinked one — it now runs from the first clean observation on, with the value (not a one-sided quantity) still checked everywhere.

The four now-exact cases get behavioural pins (`..._co_timed_arrival_and_infusion_end_inject_the_field_jump_once`, `..._input_rate_onset_across_a_covariate_record_needs_no_field_jump`, `..._infusion_end_on_a_covariate_record_reads_one_snapshot`). Before this PR none of them had a test that could fail: undo #1073's convention and the whole suite stayed green.

One tolerance is looser than the rest and says why in the test: the onset fixture's backward stencil cancels the `ka·dose` onset against the baseline decay, so it resolves ~5 digits at `h = 1e-4` and degrades to 3 by `h = 1e-7`. Bound set at 5e-5 from a realised 7.6e-6, with the sweep recorded — and the stencil still separates the branches by five orders (`fwd − bwd` is 0.47 there). The same cancellation, amplified by the `h²` denominator, is why the two onset arms take `h = 5e-3` for the second-order check while the bolus and rate-on arms take `h = 1e-3`; that sweep (1e-2 → 1e-4, the backward limit losing every digit by 2e-4) is recorded at the call site too.

## Docs

`docs/estimation/foce.qmd` carried two claims this re-measurement falsified: that #1076 "is reachable only through that same co-timed geometry" (it is not reachable at all), and the unqualified blessing of the #1068 one-sided derivative (true only on the degenerate empty-compartment case). Both corrected, #1075's figure updated from 2.2e-1 to the re-measured 2.9e-1 with its two oracles named, and the section split under three `####` subheadings to stay inside the docs-lint R1 limit.

## Checklist
- [x] `tools/preflight.sh` green, and `tools/preflight.sh release-semantics` green separately (it is omitted from the no-argument run — #1326)

## Reviewer hints

The whole change is "delete the forward scan at the arrival and use `params` on both sides". Everything else is consequence: two allocations and a slice-identity test become dead, and three fixtures become necessary.

Worth focusing on:

1. **Is `params` the right branch?** The argument is the timeline's kind ordering, and it is the load-bearing claim. `K_DOSE` = 3 sorts before `K_PKONLY` = 5 / `K_OBS` = 6, so an infinitesimal sweep puts the arrival at `t_rec − ε`, whose enclosing record is `params`. If that reading is wrong, the fix should be `post_params` on both sides instead — equally exact, measured, and it lands on the forward limit.
2. **The onset arm also uses this snapshot for the kernel's `ka` and `frac`**, not only for the Jacobian, so it sets the *magnitude* of the injected rate jump. #880 chose the post-arrival segment's snapshot deliberately. Under the backward convention the onset segment is `(t_rec − ε, t_rec]`, governed by `params` — consistent, but it is the one arm where the change is more than a Jacobian.
3. **The unmodified `#1075` residual** is the reason `docs/estimation/foce.qmd` still carries a caveat. It is second-order, does not move the objective *value*, and needs the full non-autonomous `ẍ = ∂f/∂t + J·f` derivation — deliberately out of scope here.

## Open questions

- **Resolved during review.** The intermittent `1 failed` full-suite run is
  `ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback`, reproduced
  at ~1 in 8 and named in review. It is a **pre-existing test-isolation defect**, already
  filed as [#1289](https://github.com/FeRx-NLME/ferx-core/issues/1289): the test asserts the
  process-global SS warning sink `is_empty()` while holding `SS_WARN_SINK_READER_GUARD`, but
  the writer path (`note_ss_nonconvergence_if_capped` → `dosing.rs:564`) locks only the sink
  and never that guard, so any concurrently-running test that caps an SS equilibration can
  insert between the clear and the take. Unrelated to this PR, which touches neither
  `dosing.rs` nor any SS path. #1289 has been updated with the new reproduction rate — it
  fires under plain `--features ci`, which that issue had reasoned it would not.

  The failing assertion is now **captured**, not inferred. Under a `-- ss_` filter — which
  packs the warning-producing SS tests into one run — it reproduces on the *first* iteration:

  ```
  ---- ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback stdout ----
  panicked at src/ode/predictions_tests.rs:6749:5:
  a converging nonlinear SS must not warn
  test result: FAILED. 351 passed; 1 failed; 3 ignored; 4338 filtered out
  ```

  That is the sink-emptiness assertion, i.e. exactly the contamination path above. The test's
  other two assertions cannot be reached by concurrency: `LAST_SS_EQUILIBRATION_CYCLES` is a
  `thread_local!` and the trough comparison is deterministic. Rate on the full `--features ci`
  suite measured at 1 in 20 over 20 consecutive runs on `f8bd6bfb`.
- `slow-tests.yml` has been dispatched three times on this branch
  ([34882152651](https://github.com/FeRx-NLME/ferx-core/actions/runs/34882152651) on the
  original head, [34885194577](https://github.com/FeRx-NLME/ferx-core/actions/runs/34885194577)
  after the `K_ROUTE_ONSET` fix, and
  [34888530821](https://github.com/FeRx-NLME/ferx-core/actions/runs/34888530821) on
  `f8bd6bfb`). The first **failed**, on
  `frem_covariate_omega_matches_sample_variance_under_focei` in both the `core` and `nn` jobs
  and on nothing else — epic #1350 row 1's known nightly red, whose fix (`d2223ca3`, #1386)
  landed on `main` after this branch was cut. The branch has since taken main, which is what
  the later dispatches exist to confirm — and
  [34885194577](https://github.com/FeRx-NLME/ferx-core/actions/runs/34885194577) on
  `a14720bf`, the first head containing `d2223ca3`, has now come back **green on both the
  `core` and `nn` jobs**. The `f8bd6bfb` dispatch is still in flight; a duplicate dispatch on
  that same head was cancelled.

https://claude.ai/code/session_01AUe3bzDLtXkw7z2tG5G5AS


roninsightrx:
## Review finding — changes requested

**[P1] Apply the one-snapshot rule to `K_ROUTE_ONSET` too** — `src/sens/ode_provider.rs:6101`

`K_ROUTE_ONSET` is another moving absorption onset that sorts before a co-timed record, but it still reads `post_snapshot(p, t_event)`. That leaves the per-route `first_order(..., lag=LAG)` path on the forward one-sided branch while this PR changes the shared onset paths to the backward branch and documents that sort-before-record onsets report the limit from below.

I reproduced this with two doses: the first onset is between records and leaves residual drug, while the second route onset lands exactly on the `t = 1.5` WT-changing observation. Both CL and KA read WT. For the later prediction at `t = 2`, using Richardson one-sided differences with `h = 1e-4`:

```
PR HEAD analytic df/deta_LAG  -0.6165352373
forward limit                 -0.6165352287
backward limit                -0.6015549104
```

So the current path selects the forward limit, about 2.5% from the documented backward branch. Changing only this site to `let onset_params: &[T] = params;` moved the analytic result to `-0.6015540625`, matching the backward limit within the finite-difference noise.

This path is explicitly the second onset site named in #1069, so the issue and the new documentation are incomplete without it. Please apply the same snapshot rule here and add a non-degenerate per-route-lag coincidence regression.

Validation otherwise: `cargo test --lib` passed 4,651 tests with 23 ignored; the 142 `ode_provider_` tests passed; `cargo fmt --all -- --check` passed. The PR also currently conflicts with `main` in `CHANGELOG.md` and `docs/estimation/foce.qmd`.

roninsightrx:
## [P1] Confirmed and fixed — thank you, this was a real hole and my own mutation matrix should have caught it

You are right on every point. `K_ROUTE_ONSET` is a **fourth** onset site with its **own**
`post_snapshot` call, not a caller of the shared lookahead, so collapsing the three shared
arms never reached it. Kind 4 sorts before `K_PKONLY` = 5 / `K_OBS` = 6, so it must report
the limit from below like every other arrival — and it was reporting the limit from above,
which meant `first_order(ka=KA, lag=LAG)` and an otherwise identical unlagged forcing took
**opposite branches** at the same coincidence. That is a difference of bookkeeping, not of
model, and it is exactly the inconsistency this PR exists to remove.

Reproduced independently before changing anything, on a two-dose per-route-lag geometry
(first onset interior and leaving residual drug, second onset exactly on the `t = 1.5`
`WT`-changing record, `CL` and `KA` both reading `WT`), Richardson one-sided limits at
`h = 1e-4`:

```
before   analytic df/deta_LAG   obs t=2   +0.6398...   == forward limit    (limit from above)
after    analytic df/deta_LAG   obs t=2                == backward limit   (limit from below)
```

Same conclusion as your numbers, different fixture.

Fixed in the push above: `let onset_params: &[T] = params;` at `:6101`, with the reasoning
recorded at the site rather than only in the shared arm's note — including that it is a
separate `post_snapshot` call, which is *why* fixing the shared arms did not reach it.

### Why my mutation matrix missed it

Worth writing down, because the gap was methodological rather than accidental. I mutated the
three arms the shared lookahead fed — all three, individually, which is what caught that the
rate-on and onset arms each needed their own fixture. But a mutation matrix only covers the
sites you enumerate, and I enumerated *uses of the closure I was deleting* rather than
*boundaries that sort before a record*. The fourth site was invisible to that framing because
it never used the closure.

The check that would have caught it is the one your review applied: derive the site list from
the invariant (`kind < K_PKONLY` ⇒ reports the limit from below) and confirm each site obeys
it, instead of from the diff. There are now four sites and four coincidence fixtures.

### Regression added

`ode_provider_route_lagged_onset_on_a_covariate_record_returns_a_one_sided_derivative`, plus
`..._between_records_is_exact` as the interior control. Non-degeneracy is asserted and it bit
on the first draft: I first probed for residual drug at `t = 0.5`, which is the *first* route
onset, where nothing has absorbed yet — so the check compared two runs that agree for reasons
unrelated to the fixture. It now reads `t = 1.5`, the second onset, and requires the first
dose to have delivered ≥ 0.1 % of the prediction there.

Mutation-tested per your finding: reverting **only** `:6101` to `post_snapshot(...)` reddens
exactly this test and nothing else in `sens::` (508 passed / 1 failed), so the fourth site now
names itself the way the other three do.

### Conflicts

Both resolved — rebased onto `origin/main` (`d2223ca3`, which landed #1386 after I branched).
`CHANGELOG.md` was a plain additive conflict under `### Fixed`, both entries kept;
`docs/estimation/foce.qmd` auto-merged. `tools/preflight.sh` and
`tools/preflight.sh release-semantics` are green on the rebased tree.

### Docs and changelog updated

`docs/estimation/foce.qmd` listed three onset sites; it now names four and states the rule as
a construction rather than an enumeration, so the next site added to this walk has something
to be checked against. The `CHANGELOG.md` entry says "all four onset sites" explicitly.

### Still open on my side

The intermittent single-test failure recorded under Open Questions is still unexplained: one
run in roughly ten reports `1 failed` and I have not yet captured the name — six consecutive
clean full runs after the first occurrence, and another hunt is running now. It predates this
review's change. I will not ask for merge until it is either named or shown to be
environmental.

roninsightrx:
**Correction to the numbers in my previous comment.** The "before / after" block I posted
carried a placeholder (`+0.6398...`) from the draft instead of the measured values. The
conclusion is unchanged, but the figures were not real and should not stand. The actual
measurement on my fixture — two doses, first route onset interior and leaving residual drug,
second onset exactly on the `t = 1.5` `WT`-changing record, `CL` and `KA` both reading `WT`,
Richardson one-sided limits at `h = 1e-4`:

| obs | before (`post_snapshot`) | after (`params`) | forward limit | backward limit |
|-----|-----|-----|-----|-----|
| t = 2 | −0.59502583 | −0.59188908 | −0.59502582 | −0.59188932 |
| t = 4 | +0.15285989 | +0.15473657 | +0.15285990 | +0.15473582 |
| t = 8 | +0.11211076 | +0.11275115 | +0.11211076 | +0.11275090 |

So the old path sat on the **forward** limit to 8 significant figures and now sits on the
**backward** one to ~6 — same conclusion as your reproduction, and the same direction. The
two branches are **0.53 %–1.2 %** apart here, not the 2.5 % I quoted; 2.5 % is your fixture's
contrast, and I should have labelled it as yours rather than reusing it for mine. The
committed test carries this table and those percentages.

roninsightrx:
## Code review — `medium` effort

**Verdict on the production change: no correctness defect found.** I read the surrounding ~1000 lines of `integrate_tvcov_g`, the timeline kind ordering, `governing_record_indices`, and every remaining saltation site. The argument in the PR body checks out end to end:

- `prepared_forcings` at each timeline slot is built from `params` (`src/sens/ode_provider.rs:5309`), so substituting `&prepared_forcings` for `prep_for(post_params)` in the bolus and rate-on arms is *exact*, not merely close.
- `params = next_record_params[p].unwrap_or(last_params)` equals `post_snapshot(p, t).unwrap_or(last_params)` for every interior boundary and for the trailing tail, so the change is a no-op except at the coincidence — as claimed.
- The backward-branch argument holds: `K_DOSE`=3 / `K_ROUTE_ONSET`=4 sort before `K_PKONLY`=5 / `K_OBS`=6, while `K_INF_END`=7 / `K_ZO_END`=8 sort after. No record kind sorts after 7 at the same instant, so the rate-off arms genuinely cannot part their two snapshots and are correct to keep `post_snapshot`.
- The onset arms' use of `params` for the kernel's `ka` / `frac` is right under the backward limit: at `lag → lag0⁻` the forcing turns on inside the interval `params` governs.
- All four arrival sites are covered; the other `inject_rate_saltation` call sites (4012 / 4105 / 4481 / 6899) are in non-TV-cov walks.

Everything below is comment and test quality.

### Stale prose that re-documents the removed configuration

- **`src/sens/ode_provider.rs:5886`** — the comment still says "`g(x⁻)`/`J⁻` belong to `pk_at_dose[idx]` while `g(x⁺)`, `J⁺` and the cross term belong to the arrival segment's snapshot", three lines above `let post_params = params;`. That is prose describing exactly the mixed-snapshot configuration #1068 removes. CLAUDE.md names this failure mode directly ("a comment calling a second copy 'the exact twin of' restores by prose exactly the configuration that produced the defect").
- **`src/sens/ode_provider.rs:5149`** — `post_snapshot`'s doc still claims "Every moving boundary needs this — the bolus arrival and the lagged rate-on at `K_DOSE` (#1060/#880), the per-route onset at `K_ROUTE_ONSET` (#859), and the rate-offs". After this PR only the two rate-off arms may call it; the four arrival sites must not. A future editor reading this doc reintroduces the defect at a new arrival site.
- **`src/sens/ode_provider.rs:5597`** — "shares their lookahead rather than re-scanning the timeline for itself (#1060 review #5/#7)". There is no lookahead left to share.

### Dead work the PR claims to have removed

- **`src/sens/ode_provider.rs:5600` and `:6110`** — `prep_for(onset_params)` with `onset_params == params` rebuilds a byte-identical copy of `prepared_forcings`, which the same loop iteration already built from `params`. That is a dead `PreparedInputRate` allocation on every lagged dose and every route onset, on every provider evaluation — the exact cost the PR removes elsewhere — and it is inconsistent with the bolus / rate-on arms that now reuse `&prepared_forcings`. Both can be `&prepared_forcings`.

### Assertions that cannot fail

- **`src/sens/ode_provider_tests.rs:11298`** — the non-degeneracy guard `(with_base[0] - without[0]).abs() > 1e-3` reads observation **0**, at `t = 0`, where the prediction is just `BASE/V ≈ 0.4` for any `BASE ≠ 0`. It says nothing about whether the baseline survives to the onset at `t ≈ 1.75`, which is what the comment claims it establishes. Read an observation at or after the onset (index 2 or 3).
- **`src/sens/ode_provider_tests.rs:11499`** — the sibling guard reads observation 1 (`t = 0.5`, before the dose); nearly as tautological.
- **`src/sens/ode_provider_tests.rs:11487`** — the comment states the onset at `t = 1.5` lands "where `WT` steps 70 → 140". It does not: `straddling_forcing_subject` has `WT = 70` at the `t = 1.5` record and steps to 140 at `t = 2.0`. The fixture still discriminates (the pre-fix post side was the `t = 2.0` record), but the stated geometry is wrong and the "steep step" justification points at the wrong record — someone re-tuning the dose time from this comment would break the test's discriminating power.
- **`src/sens/ode_provider_tests.rs:811`** — `check_hessian_vs_production_fd_after` silently drops the `∂²f/∂η∂θ` block that `check_hessian_vs_production_fd` checks, while its doc says only "restricted to observations from index `first` on", implying the same checks. Consequence: `ode_provider_infusion_end_on_a_covariate_record_reads_one_snapshot` verifies no mixed η×θ second derivative at all. It is also a copy of the η stencil rather than a `first` parameter on the existing helper.
- **`src/sens/ode_provider_tests.rs:11377`** — dropping `check_vs_production` in the infusion-end fixture leaves `∂f/∂θ` and `∂f/∂η` for the non-lag etas unchecked at *every* observation, not only the kinked `t = 1.5` one. They can be asserted for obs ≥ 3.
- None of the four coincidence fixtures pins any second-order quantity, though the fix also changes `J⁺` and the `δlag²` coefficient there.

### Process

Two items worth holding the merge on:

- `slow-tests.yml` has not been dispatched on this branch, despite the change landing in `sens/`.
- The PR body records an unreproduced `4650 passed; 1 failed` run without naming the failing test.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H

roninsightrx:
## Review findings applied — `f8bd6bfb`

All eight findings from the review above are fixed on the branch. No production behaviour changed: the two `prep_for(onset_params)` calls removed rebuilt a byte-identical copy of `prepared_forcings` (both derive from the same `params`), and the 203 `sens::ode_provider` tests pass unchanged at their existing tolerances, several of them at `1e-9`.

**Comments**

- `ode_provider.rs:5886` — the `#1060` paragraph rewritten: it now states why both sides read one snapshot (kind ordering + `#1073`) and says explicitly not to reintroduce `post_snapshot` there.
- `ode_provider.rs:5149` — `post_snapshot`'s doc now says only the two rate-off arms may call it, and why the rate-offs are the safe case (`K_INF_END` = 7 / `K_ZO_END` = 8 sort after every record kind).
- `ode_provider.rs:5597` — "shares their lookahead" dropped; the route-onset note's "separate `post_snapshot` call" wording fixed too.

**Dead work**

- Both onset arms now reuse `&prepared_forcings`. One fewer `PreparedInputRate` allocation per lagged dose and per route onset, per provider evaluation.

**Tests — and what each new assertion was checked against**

- `check_hessian_vs_production_fd` and `_after` are now one stencil with a `first` parameter, so the `∂²f/∂η∂θ` block cannot go missing again. Verified with a 1 % mutation of the rate saltation's `coef2`: `ode_provider_infusion_end_on_a_covariate_record_reads_one_snapshot` now dies at `ode_provider_tests.rs:843`, which is exactly the η×θ block the split had dropped — i.e. that mutation was invisible to this test before.
- New `check_vs_production_after`, called at `first = 3` in the infusion-end fixture. The value check runs at every observation; only the first-order derivative checks skip the kinked row.
- **All four coincidence fixtures now pin `∂²f/∂η²`** against the one-sided limit on the branch the boundary sorts to, via a new `one_sided_eta_second_limits` (`(2f₀ − 5f₁ + 4f₂ − f₃)/h²` and its mirror). Bounds measured, not argued — each call site records the realised worst miss and the `fwd − bwd` straddle it has to resolve:

  | fixture | `h` | worst realised miss | straddle | bound |
  |---|---|---|---|---|
  | bolus arrival | 1e-3 | 1.624e-6 | 2.6e-3 – 2.2e-2 | 1e-5 |
  | lagged rate-on | 1e-3 | 1.662e-6 | 7.4e-3 – 7.1e-2 | 1e-5 |
  | shared `first_order` onset | 5e-3 | 1.151e-3 (abs 9.0e-6) | 0.65 – 0.99 | 5e-3 / eps 1e-4 |
  | route onset | 5e-3 | 4.190e-4 | 7.1e-3 – 3.2e-2 | 2e-3 |

  `h` swept 1e-2 → 1e-4 on the two onset arms; the backward stencil there cancels the `ka·dose` onset against the decay (the same reason the gradient check already needed a larger `h`) and has lost every digit by 2e-4. Mutation: `coef2 *= 1.01` in the bolus arm reddens the bolus fixture at the new second-order assertion (`0.252607` vs `0.252334`), and the same mutation on `inject_rate_saltation`'s `coef2` reddens the other three. **The first-order assertions cannot see either mutation** — that is the coverage this adds.
- Both tautological non-degeneracy guards now read the first observation *after* the onset (obs 3) instead of obs 0 / obs 1. Realised differences 0.19 and 0.24 against a 1e-3 bound.
- The shared-onset fixture's geometry comment corrected: `straddling_forcing_subject` has `WT = 70` at the `t = 1.5` record and steps to 140 at `t = 2.0`. The comment now also says *why* the fixture still discriminates — the pre-#1068 post side was that `t = 2.0` record.

`tools/preflight.sh` green (fmt / check / clippy / rustdoc / docs / public-api). `cargo test --lib`: 4667 passed, 0 failed, twice.

---

## Correction to the two process notes in my review above

Both were wrong, and the second explains the first:

- **`slow-tests.yml` *has* been dispatched on this branch** — twice (`34882152651`, `34885194577`). I missed them.
- **The `4650 passed; 1 failed` is identified.** It is `frem_covariate_omega_matches_sample_variance_under_focei` (`tests/frem_warfarin.rs:658`, FOCEI FREM OFV = 3345.8, expected ≈ 211), and it is **not** this PR's: that run was on `0b895ad9`, which predates `d2223ca3` (#833 / #1349, "seed the FREM inner restart from the data"). The branch has since taken main and now contains that fix, which is why the re-dispatch on `a14720bf` exists. That re-run is still in flight; I have dispatched a third on the new head `f8bd6bfb` (`34888530821`). Worth waiting for one of those to go green before merging, but there is no unexplained failure here.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H

roninsightrx:
## All findings accepted and addressed — `f8bd6bfb`

Everything in the comment-and-test section was right. Taking them in order.

### Stale prose

All three fixed. `post_snapshot`'s doc was the important one: it listed the four arrival
sites as its callers, so a future editor adding an arrival boundary would have been told by
the doc to reintroduce the defect. It now states outright that **only** the rate-offs may
call it, and why the arrivals may not — the kind ordering, not a convention someone can
weigh up. The bolus arm's `#1060` note and the shared onset's "shares their lookahead" are
rewritten the same way: what the invariant is, and the consequence of breaking it, rather
than a description of the removed configuration.

### Dead `prep_for`

Correct, and embarrassing given the PR body advertises removing exactly that cost at the
other two arms. `prepared_forcings` is built from `params` at the top of the event, so with
`onset_params == params` both calls rebuilt a byte-identical copy per lagged dose per
provider evaluation. Both now reuse `&prepared_forcings`.

### Assertions that could not fail

All four, plus the wrong-geometry comment:

- Both `init(...)` guards read an observation where the property holds trivially — obs 0
  (`t = 0`, where the prediction is `BASE/V` for any non-zero `BASE`) and obs 1 (`t = 0.5`,
  before the dose). Neither said anything about the baseline surviving *to the onset*, which
  is the whole claim. Both now read obs 3, the first observation after it; realised
  differences are ≈0.19 and ≈0.24 against a 1e-3 bound.
- `check_hessian_vs_production_fd_after` is no longer a second copy: it is the single
  implementation with both blocks, and `check_hessian_vs_production_fd` is `first = 0` on
  it. You were right that the split was the mechanism — I wrote the copy by deleting the
  η×θ loop I did not need, which is precisely the edit that cannot be noticed later.
- `check_vs_production_after` added, run from obs 3 in the infusion-end fixture, so
  `∂f/∂θ` and the non-lag etas are checked everywhere except the one genuinely kinked row.
- Second order is now pinned at all four coincidence fixtures, against a one-sided second
  difference on each fixture's own branch. Worth noting why it needs the one-sided stencil:
  a central double-FD straddles the kink and returns the mean of two branches, so the
  obvious check would have been an oracle for nothing.
- The `WT` step: you are right, `straddling_forcing_subject` carries 70 at `t = 1.5` and
  steps to 140 at `t = 2.0`. The fixture discriminates *because* the pre-#1068 post side was
  that `t = 2.0` record — a stronger contrast than the comment claimed, but not the geometry
  the comment described, and the comment is what someone re-tuning the dose time would read.
  Corrected, with the old claim named so it is not silently re-derived.

One correction of my own in the same commit: the "2.5 %" branch gap I quoted at two sites is
**your** fixture's contrast, not this repo's. Measured here it is 0.53 %–1.2 %. Both sites
now say whose is whose and that the number is a property of the fixture's covariate step
rather than of the defect. (I also posted this correction above — I had let a draft
placeholder through in my first reply.)

### Process

- **`slow-tests.yml` has been dispatched**, three times: on the original SHA, again after
  the `K_ROUTE_ONSET` fix, and again on `f8bd6bfb`. Runs
  [34882152651](https://github.com/FeRx-NLME/ferx-core/actions/runs/34882152651),
  [34885194577](https://github.com/FeRx-NLME/ferx-core/actions/runs/34885194577),
  [34888683995](https://github.com/FeRx-NLME/ferx-core/actions/runs/34888683995). I agree it
  should gate the merge and I am not asking for merge before it is green.
- **The unnamed failing run** is still unnamed, and I will not pretend otherwise. What has
  changed since the PR body was written: **22 consecutive clean full-suite runs** locally
  (8 on a stable tree at the pre-review SHA, 14 since), plus CI's own `Tests (release
  semantics)` (5002 tests), `Tests + coverage (TTE/CTMM endpoints)` and `codecov/patch` all
  green. Both original occurrences came on the first `--features ci` run immediately after a
  different feature set forced a rebuild; the experiment that would isolate that was killed
  by memory pressure on this machine, and CI rebuilds from scratch every time anyway, so
  `Tests + coverage (core)` on a fresh runner is the better arbiter than another local loop.
  If it goes green I will record the occurrence as unexplained-but-not-reproduced rather
  than claim a cause I do not have.

Thanks for the review — the `post_snapshot` doc and the `_after` copy were both the kind of
thing that survives a green suite indefinitely.

roninsightrx:
## One correction: the `1 failed` is **not** `frem_covariate_omega_matches_sample_variance_under_focei`

Everything else in the applied-fixes comment matches what I see on the branch, and the
second-order mutation evidence (`coef2 *= 1.01` reddening the new assertions while the
first-order ones stay green) is exactly the property those assertions needed. Agreed and
verified.

But the flake identification does not hold, and it is cheap to check:

```
$ grep -rn "fn frem_covariate_omega_matches_sample_variance_under_focei" src tests
tests/frem_warfarin.rs:634

$ sed -n '630,634p' tests/frem_warfarin.rs
#[cfg_attr(
    not(feature = "slow-tests"),
    ignore = "slow: opt in with --features slow-tests"
)]

$ cargo test --lib --features ci -- --list | grep -c frem_covariate_omega_matches_sample_variance_under_focei
0
```

Three independent reasons it cannot be that test:

1. It lives in `tests/frem_warfarin.rs`. Both failing runs were `cargo test --lib --features ci`,
   which does not build or run `tests/` at all.
2. It is `slow-tests`-gated, so it is `#[ignore]`d under `--features ci` even where it *is*
   compiled.
3. It does not appear in the `--lib` binary's test list — literally zero matches.

The arithmetic also ties out to the lib binary rather than to an integration one: that binary
currently lists **4693** tests, and the clean runs report `4667 passed; 26 ignored` =
4693. The failing run reported `4666 passed; 1 failed; 26 ignored` — same 4693, one test
flipped from passed to failed. So the failure is a **lib** test, and `#1386` cannot have
fixed it.

I think the confusion is understandable: epic #1350 row 1 is exactly that FREM test failing
at OFV 3345.8, it was the nightly `slow-tests` red, and `d2223ca3` did land the fix for it
while this branch was open. It is the right answer to a different question.

### Where that leaves it

Unchanged from my previous comment, minus the false lead:

- Still unidentified, and still not reproduced in 22 consecutive clean full-suite runs.
- It is a lib test, so `Tests + coverage (core)` on CI — which builds from scratch on a fresh
  runner and runs the lib suite — is a genuine arbiter, not merely a proxy. That job is what
  I am waiting on.
- I would rather leave the Open Questions entry standing as "one unreproduced failure, lib
  test, unnamed" than close it on an identification that three checks contradict.

Both slow-tests re-dispatches are still useful and I am still waiting on them — that part of
your correction I agree with; my only quibble was that they had been dispatched, which you
have already withdrawn.

roninsightrx:
## Reconciling the `1 failed` — we are describing two different failures, and both accounts are right

Withdrawn: my claim that the **PR body's** unreproduced `4650 passed; 1 failed` is the FREM test. Your three checks are correct and decisive — that run was `cargo test --lib --features ci`, `frem_covariate_omega_matches_sample_variance_under_focei` lives in `tests/frem_warfarin.rs` behind the `slow-tests` gate, and it is not in the `--lib` binary's list. `#1386` cannot have fixed it, and the Open Questions entry should stand.

Standing, because it is a different run: **`Slow tests` run [34882152651](https://github.com/FeRx-NLME/ferx-core/actions/runs/34882152651) failed on exactly that FREM test.** That is the `Slow tests` workflow, not a `--lib` run — its failing step is `cargo test (ci-test profile + slow-tests)`, which does build and run `tests/`:

```
Slow regression tests (core)  ---- frem_covariate_omega_matches_sample_variance_under_focei stdout ----
                              FOCEI FREM: OFV=3345.8  omega_WT=106.80 (want 111.56)  omega_AGE=93.84 (want 99.38)
                              panicked at tests/frem_warfarin.rs:658
                              test result: FAILED. 13 passed; 1 failed
```

Both jobs (`core` and `nn`) failed on it and on nothing else. That run's head was `0b895ad9`, which predates `d2223ca3`; the branch has since taken main, so the re-dispatches should clear it. So the FREM identification is the right answer to the *slow-tests* red, and the wrong answer to the *PR body* flake — I applied it to the wrong one.

### On the PR body flake, one new checkable data point

The recorded run ran `4650 + 1 = 4651` tests. The current tree under the same feature set runs **4667** (`4693` listed − `26` ignored), measured just now:

```
$ cargo test --lib --features ci
test result: ok. 4667 passed; 0 failed; 26 ignored; 0 measured; 0 filtered out
```

So that run predates the current tree by 16 tests and is not reproducible on it even in principle — which argues for recording it as you propose rather than hunting it further on this head. I have eight more `--features ci` full runs going on this head to add to your 22; I will post the count.

### CI state

- `Slow tests` [34885194577](https://github.com/FeRx-NLME/ferx-core/actions/runs/34885194577) on `a14720bf` — in flight, first run on a head containing `d2223ca3`.
- `Slow tests` [34888530821](https://github.com/FeRx-NLME/ferx-core/actions/runs/34888530821) on `f8bd6bfb` — in flight. A duplicate dispatch on the same head has been cancelled.
- `CI` [34888457779](https://github.com/FeRx-NLME/ferx-core/actions/runs/34888457779) on `f8bd6bfb` — in flight; its `Tests + coverage (core)` job is the fresh-runner arbiter you want.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H

roninsightrx:
Agreed on all of it — two different failures, and the slow-tests red is the FREM test on
`0b895ad9`. I pulled that run's log independently and it matches exactly: both jobs (`core`
and `nn`) failed on `frem_covariate_omega_matches_sample_variance_under_focei` at
`tests/frem_warfarin.rs:658` and on nothing else.

One number to tighten, because it changes what we can conclude rather than just the wording.

> that run predates the current tree by 16 tests and is not reproducible on it even in principle

True of the **first** occurrence (`4650 + 1 + 23 ignored` = 4674 — that tree is 19 tests
behind this one). Not true of the **second**, which I recorded after the rebase:

```
test result: FAILED. 4664 passed; 1 failed; 26 ignored
```

4664 + 1 + 26 = **4691**, against 4693 on the current head. That is two tests behind, and the
two are the route-lag pair I added afterwards — so nothing was removed between that run and
this head, and whatever failed is still in the binary.

So the defensible statement is the weaker one: **unreproduced in 22+ consecutive full runs on
this head**, not *cannot recur*. I would rather the Open Questions entry say that, since
"not reproducible in principle" would quietly convert an unknown into a closed item on the
strength of an arithmetic that only holds for one of the two occurrences.

Nothing to change on the branch for it. Your eight extra runs plus my 22 and a green
`Tests + coverage (core)` on a fresh runner is the right evidence base; I will record it as
unexplained-and-unreproduced and leave it visible rather than resolved.

roninsightrx:
## The unnamed `1 failed` is reproduced and named: `ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback`

Eight consecutive `cargo test --lib --features ci` runs on `f8bd6bfb`, run 5 of 8:

```
=== run 4 ===
test result: ok. 4667 passed; 0 failed; 26 ignored
=== run 5 ===
test ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback ... FAILED
test result: FAILED. 4666 passed; 1 failed; 26 ignored; 0 measured; 0 filtered out; finished in 101.78s
=== run 6 ===
test result: ok. 4667 passed; 0 failed; 26 ignored
```

That is the `4666 passed; 1 failed; 26 ignored` shape exactly, ~1 in 8. It is **not** related to this PR — nothing here touches `src/dosing.rs` or the SS-equilibration paths — and it is not the FREM test.

### Mechanism (the likely one; I am capturing the failing assertion to confirm)

The test's last assertion reads a **process-global** sink:

```rust
crate::dosing::clear_ss_nonconvergence_warnings();
...
assert!(crate::dosing::take_ss_nonconvergence_warnings().is_empty(),
        "a converging nonlinear SS must not warn");
```

`SS_WARN_SINK_READER_GUARD` serializes only the tests that *read* the sink, on the stated reasoning that "writers don't drain, so only readers need to coordinate". That is true for *draining*, but not for this assertion: it asserts the sink is **empty**, and any concurrently-running test that caps an SS equilibration writes into the same global `BTreeSet` without touching the guard. The writers are production code —

- `src/pk/event_driven.rs:544`
- `src/ode/predictions.rs:1322`, `:1489`

— reached by any test in the same lib binary that runs a capped SS equilibration, none of which take the reader guard. So a write landing between this test's `clear` and its `take` turns the assertion red. The sibling `ss_nonlinear_over_capacity_bolus_caps_and_warns` is *not* the culprit: it takes the same guard, so it is already serialized against this one.

The `(2..SS_EQUILIBRATION_CYCLES).contains(&cycles)` assertion is the other candidate, but `LAST_SS_EQUILIBRATION_CYCLES` is a `thread_local!`, so it cannot be clobbered cross-thread. I will confirm which one fires before proposing a fix.

### Where I think this should land

Separate issue and separate PR — it is a pre-existing test-isolation defect in `#867`/`#908`'s sink, not something this branch introduced or should carry. But the Open Questions entry can now be closed with a name instead of left standing as "unidentified", which was the only thing blocking it.

12 more runs going to capture the assertion text; I will post it.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H

roninsightrx:
## Confirmed, and it is already filed: **#1289**

Your mechanism is right, and I verified it independently rather than reading it back:

- `src/ode/predictions_tests.rs:6727` takes `SS_WARN_SINK_READER_GUARD`, clears the sink, runs
  its equilibration, then asserts `take_ss_nonconvergence_warnings().is_empty()`.
- The only writer path is `dosing.rs:564`, inside `note_ss_nonconvergence_if_capped`. It locks
  the **sink**, never the reader guard. Only two sites in the tree take the guard
  (`pk/event_driven.rs:3147` and this test), so any other lib test running a capped SS
  equilibration writes in between the clear and the take.
- The guard's own doc states the assumption that fails here: *"Writers don't drain, so only
  readers need to coordinate."* Sound for a reader asserting on its own warning; false for one
  asserting the sink is empty.

**#1289** is this exact defect — same test, same panic line, same reasoning, and the same two
fix options (make the writer take the guard, or make the sink thread-local under `cfg(test)`,
which the issue argues is the better shape since the guard only exists because the sink is
global). It was filed while running the gates for #1286 and is already carried in epic #1350
under *not in this rollout → CI / test hygiene*.

So nothing new to file, and nothing for this branch to carry: it touches neither `dosing.rs`
nor any SS-equilibration path.

One detail worth adding to #1289 from your run, because the issue only had a single sighting
under `ci,markov,nn` (4,513 tests) and reasoned that the rate scales with the number of
compiled-in SS writers: it reproduces under plain **`--features ci`** too, at roughly **1 in
8** on this machine (`4666 passed; 1 failed; 26 ignored`, run 5 of 8). That is a higher rate
than the issue's "seen once" and on a smaller binary, so whatever sets the window is not
simply the writer count. I will add that to #1289.

### Open Questions entry

Now closable with a name and a link rather than left standing:

> One full `cargo test --lib --features ci` run in ~8 fails on
> `ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` — a
> pre-existing test-isolation defect in the process-global SS warning sink, filed as #1289.
> Unrelated to this PR, which touches neither `dosing.rs` nor any SS path.

I will update the PR body to that. Thanks for chasing it down — I had 22 clean runs and had
given up on reproducing it locally.

roninsightrx:
## Both gates green on `f8bd6bfb` — the merge condition in Open Questions is satisfied

- **CI** [34888457779](https://github.com/FeRx-NLME/ferx-core/actions/runs/34888457779) — success.
- **Slow tests** [34888530821](https://github.com/FeRx-NLME/ferx-core/actions/runs/34888530821) — success. (The earlier [34885194577](https://github.com/FeRx-NLME/ferx-core/actions/runs/34885194577) on `a14720bf` was already green; this is the second confirmation, on the head that carries the review fixes.)

The only remaining red on this branch is still [34882152651](https://github.com/FeRx-NLME/ferx-core/actions/runs/34882152651) on `0b895ad9`, a head that predates `d2223ca3`.

### And the flake is fixed, on its own branch

#1289 now has a PR: **#1392**. The converging test stops reading the process-global sink and asks a per-thread `LAST_SS_NONCONVERGENCE_WARNED` instead — the third member of the `last_ss_equilibration_cycles()` / `_branch()` family, so no concurrent writer can reach it. The three `.any(|w| w.contains(…))` readers get the mirror-image fix in the same PR, since every writer emits the same deduplicated message and a foreign warning satisfied them identically.

Mutation-verified in both directions with each side naming its own half, and soaked 60 consecutive `cargo test --lib --features ci -- ss_` runs clean — against a first-iteration failure before. Nothing on this branch depends on it; #1391 can merge independently.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H

PR #1392, opened 2026-09-14
test(dosing): ask whether *this* SS equilibration warned, not the global sink [closes #1289]

### LINKED ISSUE #1289: bug(tests): SS_WARN_SINK_READER_GUARD does not serialize writers, so an is_empty() assertion on the SS warning sink is a flake

roninsightrx:
## Why

`ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback` is a flake, filed as #1289. It fails on its last assertion:

```
---- ode::predictions::tests::ss_nonlinear_bolus_with_steady_state_uses_fallback stdout ----
panicked at src/ode/predictions_tests.rs:6749:5:
a converging nonlinear SS must not warn
```

That assertion reads the **process-global** SS non-convergence sink:

```rust
let _guard = crate::dosing::SS_WARN_SINK_READER_GUARD.lock()…;
crate::dosing::clear_ss_nonconvergence_warnings();
…
assert!(crate::dosing::take_ss_nonconvergence_warnings().is_empty(),
        "a converging nonlinear SS must not warn");
```

`SS_WARN_SINK_READER_GUARD` serializes only the sink's *readers*, and its doc says why: "Writers don't drain, so only readers need to coordinate." That is true for a reader asserting *its own* warning is present, and false for one asserting the sink is **empty**. The writer is production code — `note_ss_nonconvergence_if_capped` (`dosing.rs:557`), reached from `pk/event_driven.rs:544` and `ode/predictions.rs:1322`/`:1489` — and it never touches the guard. Any concurrently-running test that caps an SS equilibration inserts between the clear and the take, and the reader fails on a warning it did not produce. No amount of guard discipline on the reader side can prevent that.

Found while chasing an unexplained `1 failed` on #1391, which touches neither `dosing.rs` nor any SS path.

**Reproduction.** 1 failure in 20 consecutive `cargo test --lib --features ci` runs. The intermittency is only about which tests happen to be co-scheduled, so filtering packs the writers and the reader into one short run and it fires immediately — iteration **1** of a loop budgeted for 400:

```
$ cargo test --lib --features ci -- ss_
test result: FAILED. 351 passed; 1 failed; 3 ignored; 0 measured; 4338 filtered out; finished in 7.74s
```

The other two assertions in that test cannot be reached by concurrency: `LAST_SS_EQUILIBRATION_CYCLES` is a `thread_local!` and the `trough[0]` vs `bolus_ss_reference` comparison is deterministic. The defect is confined to the reader's use of the shared sink, not to the SS math.

## What changed

The converging test does not need the sink at all — it needs to know whether **its own** equilibration warned. That is the question `last_ss_equilibration_cycles()` and `last_ss_equilibration_branch()` already answer through a `thread_local!`, so this adds the third member of that family:

- `LAST_SS_NONCONVERGENCE_WARNED` (`#[cfg(test)]`, `dosing.rs`), written on **every** `note_ss_nonconvergence_if_capped` call rather than only the warning ones — otherwise a reader could pick up a stale `true` from an earlier call on the same harness thread.
- `ss_nonlinear_bolus_with_steady_state_uses_fallback` asserts `!last_ss_equilibration_warned()`. Its reader guard and its `clear_…()` are gone with the sink read, because it no longer touches shared state at all.

It also closes the **mirror-image hole** in the three readers that assert `.any(|w| w.contains("Steady-state (SS=1) equilibration"))` — `ss_nonlinear_over_capacity_bolus_caps_and_warns`, `ss_input_rate_no_steady_state_warns`, and `pk::event_driven`'s `ss_equilibration_falls_back_when_no_steady_state_exists`. Every writer emits the *same* deduplicated message, so a foreign test's warning satisfies those assertions exactly as well as their own: they can pass spuriously. Each now also asserts the per-thread flag. The sink check stays in all three — it is what pins the message text that actually reaches `FitResult.warnings`.

No production behaviour change: the recording is `#[cfg(test)]`, and the sink itself is untouched.

### Follow-up from review — resetting the observation (`b6e01b94`)

The first cut wrote the flag only inside `note_ss_nonconvergence_if_capped`, which made `last_ss_equilibration_warned()` mean "has anything on this thread ever warned" rather than "did **this** call warn": every path that *succeeds* returns without reaching that function — the exact affine fixed point in `equilibrate_ss_pk_state`, the closed form in `equilibrate_ss_input_rate_g`, the exact solve in `equilibrate_ss_state_event_driven`, and the `II <= 0` / out-of-range-compartment bail-outs. A warning-producing capped run followed by an exact one on the same harness thread still reported `true`. Deterministic on a reused thread, not a race — and the same stale-state class this observer exists to remove.

Each top-level entry now records `false` up front and `note_ss_nonconvergence_if_capped` overrides it with the real answer — the arrangement `equilibrate_ss_pk_state` already uses for the branch tag (`SsBranch::None`, three lines above the new call), so the two observations go stale or not together. The write goes through a new `record_ss_nonconvergence_warned` paired with a `#[cfg(not(test))]` no-op twin, like `record_ss_equilibration_cycles` / `_branch`, so the reset call sites are unconditional and free outside tests.

`equilibrate_ss_input_rate_g` needs its own reset rather than inheriting the outer one: `sens/ode_provider.rs:4270` calls it **directly** for the dual SS equilibration, so it is a top-level entry in its own right.

## Alternatives considered

- **Make writers take the guard.** The writers are production code on the prediction path; a global mutex there for a test's benefit is the wrong trade, and it would still not serialize a rayon-parallel prediction pass against a reader.
- **Make the sink thread-local.** It cannot be: `api/fit.rs` and `api/simulate.rs` drain it after a parallel pass across rayon workers, so it must be shared.
- **Delete the emptiness assertion.** It is the whole point of the test ("a *converging* nonlinear SS must not warn"), and #914 gap 1 exists to pin it.

## Cross-repo dependency

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable (test isolation only; no numerical path touched)

## Performance

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 4658 passed, 0 failed)

**Mutation evidence, each side naming its own half** — this is the part that matters, because the new assertions sit next to sink assertions that could have covered for them:

| mutation | which test dies | where | sink assertion |
|---|---|---|---|
| flag forced `false` | `ss_nonlinear_over_capacity_bolus_caps_and_warns` | `predictions_tests.rs:6787` | still passes |
| flag forced `true` | `ss_nonlinear_bolus_with_steady_state_uses_fallback` | `predictions_tests.rs:6753` | still passes (sink empty) |

So neither new assertion is redundant with the sink check, in either direction. The two tests are also a differential pair on the flag itself: without the `true` half, a flag that was never written would satisfy the `false` half vacuously, since a fresh harness thread starts at `false`.

**Reset regression tests** (review follow-up). One per engine, because mutating one reset leaves a single-engine test green — `ode::predictions::tests::an_exact_ss_equilibration_clears_a_previous_runs_warning_flag` and `pk::event_driven::tests::an_exact_event_driven_ss_clears_a_previous_runs_warning_flag` (the latter lives beside the private function it must call rather than adding a test-only shim). Each asserts its warning fixture **actually warned first**, so the "cleared" assertion cannot pass vacuously. All three resets mutated separately, each failing with its own message:

| reset removed | which test dies | where |
|---|---|---|
| `equilibrate_ss_pk_state` | ODE | `predictions_tests.rs:6853` |
| `equilibrate_ss_input_rate_g` | ODE | `predictions_tests.rs:6877` |
| `equilibrate_ss_state_event_driven` | analytical | `event_driven.rs:3223` |

**Soak.** 60 consecutive `cargo test --lib --features ci -- ss_` runs, 0 failures — against a first-iteration failure before the fix. Re-run on `b6e01b94`, same result. Full suite: 4660 passed, 0 failed.

## Docs
- [x] `CHANGELOG.md` N/A (internal test-isolation fix, no user-facing change)
- [x] No user-visible change

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, nothing on `sens/` touched

## Docs & examples
- [x] No example execution step needed (internal fix / no user-visible change)

## Reviewer hints

The load-bearing question is whether the `thread_local!` is actually read on the same thread that wrote it. It is: both tests call `equilibrate_ss_state` / `ode_predictions` synchronously on the harness thread, and `note_ss_nonconvergence_if_capped` is called unconditionally at the end of the capped train (`ode/predictions.rs:1489`), immediately after `record_ss_equilibration_cycles` — which the same tests already assert on, so the write is on a path they independently pin.

The caveat is inherited from the neighbouring cells and stated at the declaration: it reports whichever equilibration ran *last* on the thread, so it is for tests that drive the helpers directly, not for one that runs a `fit()` and hopes.

## Open questions

None.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H


roninsightrx (commented):
I found one correctness issue before this is ready:

**[P2] Reset the warning observation for SS paths that bypass `note_ss_nonconvergence_if_capped`.**

`LAST_SS_NONCONVERGENCE_WARNED` is only overwritten inside `note_ss_nonconvergence_if_capped`, but several successful SS paths return without calling it: the exact fixed-point returns in `equilibrate_ss_pk_state`, `equilibrate_ss_input_rate_g`, and `equilibrate_ss_state_event_driven`, plus the early bail-outs. That contradicts the “most recent SS equilibration” contract of `last_ss_equilibration_warned()` and leaves a deterministic stale-`true` case on a reused harness thread: run a warning-producing capped fallback, then an exact linear equilibration, and the accessor still reports that the exact equilibration warned. This is the same stale-state class the new observer is intended to eliminate.

Please reset the flag at the start of every top-level equilibration (or otherwise ensure every completion/bail-out records `false`, with the capped path overriding it after computing the warning), and add a warning → exact-call regression on one thread.

Validation performed: `cargo test --lib --features ci -- ss_` passed locally (351 passed, 3 ignored). The three new call sites themselves are synchronous and read on the writing thread.

roninsightrx:
## [P2] Confirmed and fixed — `b6e01b94`

You are right, and it is worse than intermittent: it is **deterministic** on a reused harness thread, exactly as you say. `LAST_SS_NONCONVERGENCE_WARNED` was written only inside `note_ss_nonconvergence_if_capped`, and every path that *succeeds* returns without reaching it — the exact affine fixed point in `equilibrate_ss_pk_state`, the closed form in `equilibrate_ss_input_rate_g`, the exact solve in `equilibrate_ss_state_event_driven`, and the `II <= 0` / out-of-range-compartment bail-outs. So `last_ss_equilibration_warned()` answered "has anything on this thread ever warned", not "did this call warn". That is the same stale-state class the observer exists to remove, and I shipped it with the hole.

Worth adding to your list: **`equilibrate_ss_input_rate_g` is a top-level entry in its own right**, not only a callee of `equilibrate_ss_pk_state` — `sens/ode_provider.rs:4270` calls it directly for the dual SS equilibration. A reset at the outer function alone would have left that caller reading a stale flag, so it needs its own.

### What changed

Each top-level entry records `false` up front, and `note_ss_nonconvergence_if_capped` overrides it with the real answer — the arrangement you proposed, and the one `equilibrate_ss_pk_state` already uses for the branch tag (`SsBranch::None`, three lines above the new call), so the two observations now go stale or not together.

The write goes through a new `record_ss_nonconvergence_warned`, paired with a `#[cfg(not(test))]` no-op twin like `record_ss_equilibration_cycles` / `_branch`, so the reset call sites are unconditional and cost nothing outside tests. Three sites:

| | |
|---|---|
| `ode/predictions.rs` | `equilibrate_ss_pk_state` |
| `ode/predictions.rs` | `equilibrate_ss_input_rate_g` — the `sens::ode_provider` entry |
| `pk/event_driven.rs` | `equilibrate_ss_state_event_driven` |

### The regression test — two of them, and why

One test cannot cover this: mutating one reset leaves a single-engine test green. So there is one per engine, each asserting the fixture **actually warned first** so the "cleared" assertion cannot pass vacuously:

- `ode::predictions::tests::an_exact_ss_equilibration_clears_a_previous_runs_warning_flag` — warn, then an exact *linear* equilibration (`equilibrate_ss_pk_state`), then re-warn and a converging *input-rate* solve (`equilibrate_ss_input_rate_g`).
- `pk::event_driven::tests::an_exact_event_driven_ss_clears_a_previous_runs_warning_flag` — lives beside the private function it has to call rather than adding a test-only shim to reach it from another file.

All three resets mutated **separately**, each failing with its own message rather than a shared one:

| reset removed | which test dies | where |
|---|---|---|
| `equilibrate_ss_pk_state` | ODE | `predictions_tests.rs:6853` |
| `equilibrate_ss_input_rate_g` | ODE | `predictions_tests.rs:6877` |
| `equilibrate_ss_state_event_driven` | analytical | `event_driven.rs:3223` |

### Gates

`tools/preflight.sh` green. `cargo test --lib --features ci`: 4660 passed, 0 failed. Soak re-run on the updated tree: 60 consecutive `cargo test --lib --features ci -- ss_` runs, 0 failures — against a first-iteration failure before the original fix.

No production behaviour change: every recorder is a no-op outside `cfg(test)`.

https://claude.ai/code/session_01HEiafMVvNTN5joWxdeT32H

Issue #1396, opened 2026-09-15
CLI warning: unrecognized function


roninsightrx:
When running a tool function that is not (yet) available, the CLI just throws an error that the model file cannot be found. E.g.

```
ferx covsearch covsearch.ferxsearch
Error: Failed to read model file: No such file or directory (os error 2)
```

The behavior should be that, if the second word in that command is a word without extension (no dot), then assume it is a tool call, like covsearch, bootstrap etc. So in the above case, in an older version where covsearch was not yet available, it should throw back a different error saying e.g. "tool `covsearch` not recognized". If the 2nd word did contain an extension (.ferx or other), then the error should be caught as well and a nicer error shown, e.g. "The model `run1.ferx` was not found at this location. Please check folder and file names.`

PR #1397, opened 2026-09-15
fix(cli): tell a tool name from a model path before opening it [closes #1396]

roninsightrx:
## Why

Running a tool the build does not have reported a missing *model file*:

```
$ ferx covsearch covsearch.ferxsearch
Error: Failed to read model file: No such file or directory (os error 2)
```

`covsearch` fell through the subcommand chain, `main` took it as `model_path`,
and the fit path opened it. The message names neither the mistake (a tool that
does not exist here — a typo, or one that only ships in a later version) nor the
argument that produced it. The same message answered a genuinely missing model
file, so the two failures were indistinguishable (#1396).

## What changed

The first argument is classified *before* anything is opened
(`classify_first_arg`). A tool name is a bare word by construction, so:

| first argument | verdict | message |
|---|---|---|
| `covsearch` (no dot, no separator, not a subcommand) | unknown tool | ``Error: tool `covsearch` not recognized.`` + `Did you mean …?` (within two edits) + `Available tools: …` |
| `run1.ferx`, `runs/run1` (missing) | missing model | ``Error: the model `run1.ferx` was not found at this location. Please check folder and file names.`` |
| any path that exists | model file | unchanged — the fit/simulate path |
| `--data` in the model position | flag | usage to stderr |

Existence wins over the name, so an extension-less model file (`ferx mymodel`)
still fits.

The twelve subcommands also move from a chain of `if args.get(1) == Some("…")`
arms into one `SUBCOMMANDS` table of `(name, fn(&[String]) -> i32)`. Dispatch
reads it and the unknown-tool error enumerates it, so a tool cannot be
advertised without dispatching, or dispatch without being advertised.

Both new error paths exit `1`, as the old one did.

## Alternatives considered

- **Keying purely on the extension.** Rejected: an extension-less file that is
  on disk is a legal model, and calling it an unknown tool would be a new bug.
  The `exists` check comes first, and `an_extensionless_model_file_that_exists_is_fitted_not_rejected`
  pins it — without that arm, every other test here still passes.
- **Giving the tools' own file arguments the same treatment** (e.g.
  `SearchConfig::load` on a missing `.ferxsearch`). Left alone: it already
  reports `cannot read <path>: <io error>`, which names the file. Out of scope
  for this issue.

## Cross-repo dependency

CLI-only; no `ferx-core` public API touched (baseline unchanged).

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable (CLI error messages only — no numerical path touched)

## Performance

## Tests

Tier 1 (`crates/ferx-cli/src/main.rs`): `classify_first_arg` on each arm, the
`SUBCOMMANDS` table resolving every listed name, the hint suggesting a
one-edit typo and never a far word, and `edit_distance` on the three single
edits plus both empty cases.

Tier 2 (`crates/ferx-cli/tests/cli_ferx.rs`, subprocess): the reported command
line; a missing model reported by name; an existing extension-less file still
reaching the fit path; and every tool the error *lists* actually dispatching
(the list is scraped back out of the binary's own output, so the two cannot
drift).

Each assertion was mutation-checked, and each mutation was chosen as the
smallest edit that removes the behaviour:

| mutation | dies |
|---|---|
| drop the `exists` short-circuit | `a_file_that_exists_is_a_model_whatever_it_is_called` (T1), `an_extensionless_model_file_that_exists_is_fitted_not_rejected` (T2) |
| classify every non-flag as `MissingModelFile` | `a_bare_word_is_read_as_a_tool_name_not_a_model_path` (T1), `an_unknown_tool_name_is_named_as_a_tool_not_as_a_missing_file` (T2) |
| dispatch only `check` from the table | `every_listed_tool_still_dispatches` (T2) |
| drop the `<= 2` edit-distance cutoff | `the_hint_names_a_near_miss_and_always_lists_the_tools` (T1) |

The Tier-2 unknown-tool test also asserts the *old* string
(`Failed to read model file`) is absent, so the regression it exists to catch
is the one it observes.

## Docs
- [x] `docs/` updated for user-visible changes — new "Tool name or model file?"
      section in `docs/cli.qmd`, next to Exit Codes (existing page, no
      `_quarto.yml` change)

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets,
      clippy (`--all-targets`), rustdoc, docs lint, public-API baseline
- [x] Not on the `Dual2` sensitivity path

## Example execution
- [x] No example execution step needed (CLI error paths only; `cargo test -p
      ferx-cli` green — 131 unit + 32 integration)

## Reviewer hints

The rule worth arguing with is the order in `classify_first_arg`: flag →
exists → dot-or-separator → tool. `exists` sitting above the name check is
what keeps extension-less model files working; the dot check below it is what
makes `covsearch` a tool rather than a missing file.

Second thing to check: `SUBCOMMANDS` is now the only list of tool names. The
unit test pins that every listed name dispatches, but *ordering* — `main`
consulting the table before classifying — can only be seen from outside, which
is what `every_listed_tool_still_dispatches` does.

## Open questions

None.


roninsightrx (commented):
Review findings:

- **[P2] Preserve non-`NotFound` filesystem errors in the preflight check** (`crates/ferx-cli/src/main.rs:246`). `Path::exists()` returns `false` for any metadata error, not only `ENOENT`. That means a model below an unreadable directory, a symlink loop, etc. now enters `MissingModelFile` and reports “was not found,” whereas the old fit path surfaced the actionable OS error (for example, `/private/var/root/nope.ferx` now reports “not found” even though `stat` returns “Permission denied”). Please use `try_exists()` and reserve the friendly missing-model message for `Ok(false)`; on `Err`, preserve/report the underlying error (or let the existing model reader do so). A regression test for a non-`NotFound` probe error would pin this distinction.

- **[P3] Qualify the “existing file always wins” wording** (`docs/cli.qmd:421-424`, and similarly the PR description/test naming). `main` dispatches known subcommands at lines 230-234 and handles leading-dash arguments as flags before the existence result, so an existing extensionless file named `check` or `covsearch` is not treated as a model. That precedence is reasonable and pre-existing, but “whatever it is called” is stronger than the behavior; documenting “unless its name is a reserved subcommand (or it is parsed as a flag)” would make the contract accurate.

Verification: `cargo test -p ferx-cli` passes (131 unit + 32 integration tests), `git diff --check` passes, and all required PR checks are green.

roninsightrx:
Both findings taken — fixed in 8918daf1.

**[P2] `try_exists`.** Correct, and the consequence was worse than losing a nicer message: the new arm *manufactured* a wrong claim. `Path::exists` answers `false` for `EACCES`, `ELOOP`, `ENAMETOOLONG` alike, so a model behind an unreadable parent was reported as "was not found at this location. Please check folder and file names" — advice that sends the user to look for a file that is sitting right there.

The probe is now `Path::try_exists`, and the classifier takes `std::io::Result<bool>`:

```rust
match exists {
    Ok(true) | Err(_) => return FirstArg::ModelFile(arg),
    Ok(false) => {}
}
```

`Err` deliberately lands on `ModelFile` rather than getting its own message — the fit path already opens the file and reports the OS error verbatim, which is the actionable thing, and duplicating that here would be a second place to keep the wording right. That applies to a bare word too: with the probe failing we do not know it is *not* a file, so "tool not recognized" would be a guess.

Two tests, and the reason there are two is that neither sees the other's mutation:

- `a_probe_that_fails_is_not_an_absent_file` (Tier 1) feeds a synthesized `PermissionDenied` — it pins the classifier, and dies when the `Err` arm is folded into "absent".
- `a_probe_error_keeps_the_os_error_instead_of_claiming_not_found` (Tier 2) builds a real symlink loop in a tempdir and runs the binary — it pins the *call site*, and dies when it is reverted to `exists()`. A symlink loop rather than an unreadable directory on purpose: `chmod 000` is bypassed by root, and CI containers often run as root, so that version would have been green-but-blind exactly where it matters. The test asserts the probe really does fail before asserting anything about the output, so it cannot quietly degrade into a plain missing-file case.

Verified both mutations redden their own side and only their own side.

**[P3] Wording.** Agreed — "whatever it is called" was stronger than the code. `docs/cli.qmd` now gives the precedence as an ordered list (reserved tool name → leading `-` is a flag → existing file → bare word vs. path), and says outright that a file called `check` needs `./check`. The same qualification is on the classifier's doc comment, and the unit test is renamed `a_file_that_exists_is_a_model_unless_its_name_is_reserved`. The CHANGELOG entry carries both corrections.

Added `a_file_named_like_a_tool_does_not_shadow_the_tool` (Tier 2) so the documented precedence is pinned rather than just described: with a file named `check` in the working directory, `ferx check` still exits 2 as the checker's own usage error.

`tools/preflight.sh` green; `cargo test -p ferx-cli` 132 unit + 34 integration.

Issue #1390, opened 2026-09-14
bug(parser): [fit_options] and [error_model] silently drop unmatched lines — `method saem` without the `=` runs FOCEI with no diagnostic


TeunP:
## Summary

`[parameters]` gained a "consumed end to end, or it is a parse error" rule in #1377 / #1388. `[fit_options]` and `[error_model]` still do what `[parameters]` used to: a line matching no known form is dropped without a trace.

The sharp case is `method` written without its `=`. The line vanishes, the fit runs the **default estimator**, and nothing anywhere says so — the user believes they ran SAEM and reads a FOCEI objective.

## Measured

At `9a7d79c` (the #1388 branch), debug `ferx check`, one-cpt IV warfarin-shaped model, nothing else changed.

```
[fit_options]
  method saem
  maxiter 42
```

```
$ ferx check fo.ferx
ok: fo — no errors (0 warning(s))
```

Both lines dropped. `method` and `maxiter` keep their defaults.

Same block, same build, `[error_model]`:

```
[error_model]
  DV ~ proportional(PROP_ERR)
  banana
```

```
$ ferx check em.ferx
ok: em — no errors (0 warning(s))
```

## Why this is worse than the `[parameters]` case it mirrors

The `[parameters]` defect (#1377) moved an *initial estimate*. This one moves the **estimator**. A run that silently falls back from SAEM to FOCEI is not a perturbed start; it is a different algorithm, and the objective is not comparable to the one the user thinks they asked for. There is no diagnostic at any level — `ferx check` reports VALID, the fit reports nothing, and the `-fit.yaml` records the method that actually ran, which the user has no reason to re-read.

It is also the *easy* typo. `method = saem` and `method saem` differ by one character, and the surrounding DSL is not uniformly `key = value` — `[parameters]` and `[error_model]` lines are not, so the habit does not transfer.

## Prior art in the same file

Four sibling parsers already hard-error on exactly this shape, one with a comment arguing for it, and `[mixture]` has had an `unrecognized directive` arm since #1040. `[structural_model]` rejects with a form list. So the convention exists; these two blocks are the holdouts.

`[parameters]`' new `else` arm (#1388, `model_parser.rs`) is a worked example of the shape: reject, quote the offending line, list the accepted forms from a single array (`PARAMETER_FORMS`) pinned by a test so the message cannot drift from the grammar.

## Suggested scope

[...]

Issue #1393, opened 2026-09-15
feat(parser): accept `;` as a comment marker in .ferx files (NONMEM habit) — one shared comment stripper, warn on a commented-out declaration


TeunP:
## Problem

`.ferx` comments are `#` and `//` only. `;`, the comment marker of NONMEM control streams, means nothing in a `.ferx` file. People who write NONMEM, and files translated from it, carry `;` comments by habit.

Until #1377 (PR #1388) a `;` line in `[parameters]` was silently mis-read:

| line | before #1388 | after #1388 |
|---|---|---|
| `; theta TVCL(1, 0.1, 100)` (meant as commented out) | **declared a live theta** | `E_PARSE`: "`;` does not start a comment … use `#` or `//`" |
| `theta TVCL(0.2, 0.001, 10.0) ; 30% CV` | VALID, tail dropped | `E_PARSE`: same comment message |
| `theta A(1.0); theta B(2.0)` | **declared only `A`** | `E_PARSE`: "`;` does not separate declarations … one per line" |

PR #1388 makes all three errors rather than silent. That is safe, but it will break any user model or translated file that carries `;` comments in `[parameters]`. The in-tree scan found none (256 `.ferx` files, 65 `[parameters]` fences in `docs/`), but `ferxtranslate` output and users' own files were not scanned. Two of our own doc examples had the third shape (`docs/estimation/tte.qmd`, `docs/estimation/mixture.qmd`, fixed in #1388).

This issue asks whether `;` should become a comment marker in `.ferx`, as it is in NONMEM.

## Why it is not a one-line change

1. **Comment stripping is copied, not shared.** At least these sites strip `#` (and some `//`) independently, and they already disagree with each other:
   - `extract_blocks` in `src/parser/model_parser.rs`: `#` and `//`, not quote-aware
   - the expression line loop (`raw_line.find('#')`), `parse_output_block`, the `[mixture]` loop (`raw.split('#')`), the multi-line if/else joiner and the `[error_model]` statement loop, all in `model_parser.rs`: `#` only
   - `strip_comment` in `src/parser/covariate_model.rs`: `#` only
   - `strip_comment` in `src/io/fit_estimates.rs`: `#` only
   - `src/frem/mod.rs`: several `starts_with('#')`
   - `code_of` in `src/edit/mod.rs`: `#` and `//`, **quote-aware**, which the parser's copy is not

[...]

PR #1388, opened 2026-09-14
fix(parser): reject a scale tag on a block declaration, and any unrecognised [parameters] line [closes #1377]

TeunP:
## Why

`(sd)` on a **block** declaration was silently ignored. `docs/model-file/parameters.qmd` promised it "is not accepted there", but the parser neither rejected it nor applied it: the tag was dropped and every lower-triangle entry read as a variance, with `ferx check` reporting the file **VALID** and zero diagnostics. On `omega ETA_CL ~ 0.2645751 (sd)` the intended variance is 0.07; taken as a variance the run starts at 0.2645751, a 3.8x error.

That tag is one instance of a wider hole: **`[parameters]` never rejected anything it did not recognise.** All seven declaration regexes were unanchored prefix matches and the `if let / else if` chain had *no* `else` arm, so anything outside a successful match — and any line matching nothing at all — was dropped without a trace. Measured at `257be39a`, every one of these reported VALID:

| probe | old result |
|---|---|
| `block_omega (…) = [0.09, 0.02, 0.04] (sd)` | VALID (the issue) |
| `… ] FIX (sd)` / `… ] (sd) FIX` / `… ] (variance)` | VALID |
| `omega ETA_KA ~ 0.30 (sdx)` | VALID — a misspelt tag silently meant *no* tag |
| `omega ETA_KA ~ 0.30 banana`, `theta TVKA(…) (sd)` | VALID |
| `banana` on a line of its own | VALID — dropped without a trace |
| `(sd)` on its own line after the block's `]` | VALID |
| **`block_sigma PROP ~ 0.04`** | VALID — silently declared a **diagonal** sigma |
| **`; theta TVCL(1, 0.1, 100)`** | VALID — declared a live theta; `;` is not a comment marker |
| **`x = block_omega (A, B) = […]`** | VALID, `x = ` dropped |

The last three came out of review (§1 below) and are the same defect seen from the head of the line rather than the tail.

## What changed

Implements the plan in [#1377 (comment)](https://github.com/FeRx-NLME/ferx-core/issues/1377#issuecomment-5659914911) §5, plus the review findings. **The rule: a `[parameters]` line is either consumed end to end by exactly one declaration form — both ends — or it is a parse error.**

- **`src/parser/model_parser.rs`** — all seven declaration regexes now pinned `^\s*…\s*$`. Every optional tag/`FIX` group stays optional; leading and trailing *whitespace* stay legal, pinned by tests.
[...]

Issue #1377, opened 2026-09-12
bug(parser): [parameters] silently ignores unrecognised text — (sd) on block declarations read as variances, stray lines and trailing tokens dropped


TeunP:
> **Scope widened (2026-09-14).** The plan below ([comment](https://github.com/FeRx-NLME/ferx-core/issues/1377#issuecomment-5659914911)), measured at `257be39a`, found that the block `(sd)` case is one instance of a wider hole: **`[parameters]` never rejects text it does not recognise.** Every declaration regex is an unanchored prefix match and the `if / else if` chain has no `else` arm, so a misspelt tag (`omega ETA_KA ~ 0.30 (sdx)`, read as a variance), trailing junk after a value, and a stray line on its own (`banana`) all validate clean. This issue now covers the whole rule: **a `[parameters]` line is either consumed end-to-end by exactly one declaration form, or it is a parse error.** That is a **breaking** change for any file relying on ignored text (measured exposure 0 across 219 in-tree and ferx-r fixtures). NONMEM-style block tags are split out as #1387.

---

## Summary

`(sd)` on a **block** declaration is silently ignored instead of rejected. `docs/model-file/parameters.qmd` says the annotation "is not accepted there because the lower-triangle list mixes variances and covariances and a single tag would be ambiguous", but at `8372248c` the parser neither rejects it nor applies it: the tag is dropped and every entry is read as a variance. `ferx check` reports the file **VALID** with zero diagnostics.

So a user who writes standard deviations in a block — the natural reading of "SD coding is accepted for omega", which is true for the diagonal form — gets variances, with no warning anywhere. On a typical `omega ETA_CL ~ 0.2645751 (sd)` the intended variance is 0.07; taken as a variance the run starts at 0.2645751, a 3.8× error in the initial estimate, in silence.

All three block forms behave the same way.

## Measured

Installed ferx-r `078e489` / ferx-core `8372248c`. Each pair is the bundled example against a copy with `(sd)` appended to the block line, nothing else changed.

| declaration | OFV without `(sd)` | OFV with `(sd)` | identical |
|---|---|---|---|
| `block_omega (ETA_CL, ETA_V) = [0.07, 0.02, 0.02]` (`warfarin_block_omega`) | −280.485776590358 | −280.485776590358 | yes |
| `block_kappa (KAPPA_CL, KAPPA_V) = [...]` | 202.186054982612 | 202.186054982612 | yes |
[...]

PR #1398, opened 2026-09-15
fix(parser): reject an unmatched [fit_options] or [error_model] line [closes #1390]

roninsightrx:
## Why

`[parameters]` gained a "consumed end to end, or it is a parse error" rule in #1377 / #1388. `[fit_options]` and `[error_model]` still did what it used to: a line matching no known form was dropped without a trace.

The sharp case is `method` written without its `=`. The line vanished, the fit ran the **default estimator**, and nothing anywhere said so. Reproduced at `d66046e3`:

```
[fit_options]
  method saem
  maxiter 42
```

```
$ ferx check fo.ferx
ok: fo — no errors (0 warning(s))
```

That is a different *algorithm*, not a perturbed start — a run the author reads as SAEM is FOCEI, and its objective is not comparable to the one they asked for. It is also the easy typo: `method = saem` and `method saem` differ by one character, and the surrounding DSL is not uniformly `key = value`, so the habit does not transfer. The only record of what actually ran was the `-fit.yaml`, which the user has no reason to re-read.

Same shape, `[error_model]`:

```
[error_model]
  DV ~ proportional(PROP_ERR)
  banana
```

Dropped, `ferx check` VALID. Worth noting the block-with-only-the-stray-line case already failed — late and unhelpfully, with "No error model found in [error_model] block". The defect is only visible when a valid statement is there to be fitted *instead*, which is the fixture shape both the tests and the NM-TRAN arm use.

After:

```
$ ferx check fo.ferx
error[E_PARSE]: [fit_options]: `method saem` is not a `key = value` pair — did you mean
`method = saem`? Every `[fit_options]` line assigns one key with `=`; a line without it is
rejected, not ignored.
invalid: fo — 1 error(s), 0 warning(s)

$ ferx check em.ferx
error[E_PARSE]: [error_model]: unrecognized line `banana`. Expected one of:
`DV ~ additive(SIGMA)`, `DV ~ proportional(SIGMA)`, `DV ~ combined(SIGMA_PROP, SIGMA_ADD)`,
`DV ~ power(SIGMA, EXPONENT)`, `DV ~ log_additive(SIGMA)`, `log(DV) ~ additive(SIGMA)`,
`CMT=N: DV ~ TYPE(...) (per-endpoint, ODE models)`, `<statement> weight = <expr>`,
`iiv_on_ruv = ETA_NAME (alongside a statement)`,
`if (COND) { DV ~ TYPE(...) } else { DV ~ TYPE(...) } (covariate-selected)`. A statement must
be the whole line — leading or trailing text is rejected, not ignored.
invalid: em — 1 error(s), 0 warning(s)
```

## What changed

**`[fit_options]`** — the whole grammar is `key = value`, so the message names the exact repair rather than listing forms. Whether the first token is a real option is decided by probing **`apply_fit_option` itself** on a throwaway `FitOptions`, not by a second list of key names: it is already the single source of truth for the key grammar (the R `settings` path shares it), so a key added there cannot fall out of this diagnostic. `Err` counts as *recognised* — a known key with a malformed value is still a known key, and the repair is still the one the user wants.

**`[error_model]`** — a line matching no statement form is rejected, quoting `trimmed` (what the user wrote) rather than the peeled `body`, whose `CMT=` prefix and `weight =` tail are gone and which can therefore be a string appearing nowhere in the file. #1388 made exactly this fix in `[parameters]`. The accepted forms come from one `ERROR_MODEL_FORMS` array, and `every_listed_error_model_form_actually_parses` runs the real `parse_error_model` on each example, so a form cannot be advertised that the parser does not accept. The second half of that test asserts the list reaches the *real* diagnostic, since checking against `error_model_form_list()` alone is tautological and would stay green if the arm went back to a hand-written list.

**The `;` message is now one shared helper** (`semicolon_is_not_a_comment`) across `[parameters]`, `[fit_options]` and `[error_model]`. Three copies of that sentence would be three chances to disagree about what `.ferx` accepts as a comment.

**One placement that is load-bearing, and was wrong first.** For `[fit_options]` the `;` check has to run *before* the `=` split. Both NONMEM-converted shapes carry an `=`, so written inside the reject arm it was dead code: `; method = saem` split to the key `` `; method` `` and reported "unknown key"; `method = saem ; run 3` reported an unknown *method token*. Measured — the test I wrote for it failed on its first run, which is how this was found.

**Latent fixture bug repaired.** `test_diffusion_on_analytical_model_is_error` carried a bare `additive` in `[error_model]`. It asserts `is_err()`, so it would have kept passing — for the wrong reason, on the new error rather than the `[diffusion]` one it exists to pin.

## Alternatives considered

**A dedicated diagnostic code** (`E_MISSING_EQUALS`) rather than the `E_PARSE` catch-all. #1388's own rule for when a shape earns one is "the remedy is mechanical, so a consumer can offer it rather than reprinting prose", and `method saem` → `method = saem` qualifies. Left out as scope the issue did not ask for; the repair is in the message and a code is a cheap follow-up if `ferxtranslate` / ferx-r want to act on it.

**Guessing the `;` case from the `;` alone.** Rejected on #1388's precedent (review round 3 #9): a message that names a problem the line does not have is worse than the generic one. Both blocks require nothing before the `;`, or a complete declaration/statement/pair before it. In `[error_model]` the other side of that gate needed care — `DV ~ combined(PROP; ADD)` is *not* a usable non-comment example, because the argument regex is greedy and the line is caught by the sigma-count check long before the reject arm. The test says so, since it first read green on a message it never produced.

## Cross-repo dependency

No public-API change (baseline unchanged). ferx-r's `settings` path goes through `apply_fit_option`, which is untouched — only the `.ferx` block parser gained the reject arm.

## Breaking changes
- [ ] Public Rust API
- [ ] None

<details>

A model file that previously parsed may now be rejected — which is the point, since the previously-parsed behaviour was not the behaviour the file described. Both repairs are mechanical and named in the message: add the missing `=`, or delete/fix the unmatched `[error_model]` line. `;` is not a comment marker (`#` and `//` are); making it one is #1393.

**Measured exposure: zero.** See below — nothing in the repo needed a migration except one test fixture that was already wrong.

</details>

## Numerical validation

- [x] Compared against: **NONMEM 7.5.1 (NM-TRAN parse outcomes)** — two comparator streams, one-edit accepted control, in `nonmem_anchor/`
- [x] Not applicable for a *number*: the object under test is a parse outcome, so there is nothing to anchor. Same reasoning as #1388's `parameters_reject_*.ctl`.

**Arm B — `error_model_reject_stray_line.ctl` — supports the rule.** `BANANA` on its own line in `$ERROR`, *beside a complete* `Y = IPRED * (1 + EPS(1))`, is rejected: `AN ERROR WAS FOUND IN THE CONTROL STATEMENTS.`, the line echoed under a caret, `  202  FORTRAN SYNTAX IS INCORRECT OR INAPPROPRIATE IN THIS CONTEXT.`, nmfe exit 107. Control (that line deleted, nothing else) accepted, exit 0, `OBJECTIVE FUNCTION VALUE: 215.06724369784769`.

**Arm A — `fit_options_reject_missing_equals.ctl` — is a measured NEGATIVE result, committed so it is not mis-quoted.** `$ESTIMATION MAXEVAL 0` — the `=` left out — is **accepted**, exit 0, and **honoured**: the `.lst` reports `ESTIMATION STEP OMITTED:                 YES`. A NONMEM record is a token stream running to the next `$` in which whitespace and `=` are both separators; ferx's `[fit_options]` is a line-oriented `key = value` block. NM-TRAN cannot express this object, so it is not a comparator for it, and that half is anchored on ferx's own terms instead: an exact hand-checked message, plus a `parse_full_model` control that the accepted spelling still parses *and still selects SAEM* (without that control the test is satisfied by a check that rejects everything).

**Read the asymmetry on arm B too**, written into the file header: `$ERROR` is Fortran, so NM-TRAN rejects `BANANA` as a *syntax* error rather than as "a line matching no error-model form" — it has no notion of a declaration form there. The claim supported is the narrower one, that a stray line in the residual-error region is an error and not something to drop.

Provenance: `nmfe75` (`/opt/NONMEM/nm751/run/nmfe75`), NONMEM 7.5.1, licensed `pmx` container image `a8b78e6253e5`, 2026-09-15, measured on the **committed** files (the header shifts NM-TRAN's reported line number from 18 to 64; the file says so).

## Performance
- [x] No significant impact expected — one extra `split_once` per `[fit_options]` line at parse time.

## Tests

**Exposure measured with the real parser, not a grammar re-implementation.** `ferx check` over every `.ferx` in the repo:

```
--- ferx check over 184 .ferx files; #1390 rejects: 0
```

Inline model strings (where a `\n`-joined literal defeats a static scan) are covered by running the suites: `--lib --features ci,survival` **4881 passed**, `ferx-tools` (which generates `[fit_options]` text) **602 + 68 passed**, `check_command` + `model_text_roundtrip` **40 passed**. The one hit the static scan did find was the `test_diffusion_on_analytical_model_is_error` fixture above.

**Each gate mutation-tested separately**, since "I mutation-tested my own fix" does not cover a gate whose twin covers for it:

| mutation | tests reddened |
|---|---|
| `[fit_options]` reject arm → `continue` | 4 |
| `[error_model]` reject arm → `continue` | 5 |
| drop the pre-split `;` hint call | 1 (`a_semicolon_comment_in_fit_options_is_named_as_such`) |
| known-key gate forced `true` | 1 (`fit_option_reject_distinguishes_a_known_key_from_an_unknown_one`) |
| known-key gate forced `false` | 3 |
| `[error_model]` `;` gate forced always-on | 1 (`a_semicolon_in_error_model_is_named_a_comment_only_when_it_is_one`) |

Each differential test asserts **both** sides of its gate, so none is satisfiable by an arm that rejects everything: `banana` must *not* get a `did you mean`, `banana ; x` must *not* get the comment message, `DV ~ additive; (ADD)` must get the form list, and `method = saem` must still parse and still select SAEM.

- [x] Not applicable — this adds no API surface; it makes an existing silent drop loud. The before/after `ferx check` output is in **Why** above.

## Docs
- [x] `docs/` updated — `model-file/fit-options.qmd` (Syntax), `model-file/error-model.qmd` (Syntax, with the full form list), `file-formats/check-report.qmd` (the `E_PARSE` paragraph)
- [x] `cargo run -p docs-lint -- --check` green (three R4 anchors corrected: pandoc drops the backticked `(iiv_on_ruv)` into the slug and collapses `(if/else)` to `ifelse`)

## Checklist
- [x] Not on the `Dual2` path — parser only
- [ ] `Cargo.toml` version bump — not needed

## Docs & examples
- [x] No new `.ferx` DSL syntax or `[fit_options]` key, no new example file, no new data file
- [x] No example execution step needed — no user-visible behaviour change to any shipped example (all 184 `.ferx` files parse unchanged)

## Reviewer hints

Focus on the two reject arms and their `;` gates; everything else is message text, tests and docs.

Three things worth a second look:

1. **The `apply_fit_option` probe** (`reject_fit_option_line`). It mutates a throwaway `FitOptions` — I checked the function touches nothing but the `opts` it is handed, no statics or globals. If you disagree that `Err` should count as "recognised", that is the one judgement call in the message logic.
2. **The pre-split `;` placement.** This is where the first version was wrong and the test caught it. The condition for the named message (`head.is_empty() || is_pair(head)`) is deliberately narrow.
3. **`ERROR_MODEL_FORMS`'s selected-`if/else` example** is a one-liner, because an example is parsed as a whole block body. The multi-line spelling in the docs parses the same way (`parse_selected_error_model` joins the block first), but I have not pinned that specific equivalence.

## Open questions

Worth a dedicated code (`E_MISSING_EQUALS`) for the `[fit_options]` case, carrying the repair in `suggestion`? It fits #1388's "the remedy is mechanical" test, and `ferxtranslate` / ferx-r could offer the fix rather than reprinting prose. Left out as unrequested scope — happy to add it here or file it.


roninsightrx (commented):
Found one issue:

**[P2] Run the error-model semicolon check before parsing accepted variants** — `src/parser/model_parser.rs:16630-16744`

The new `;` hint is only reached in the plain statement regex's `None` arm, after `parse_selected_error_model` and `split_weight_modifier` have already taken their own paths. As a result, two accepted forms advertised by `ERROR_MODEL_FORMS` do not get the promised comment diagnostic when followed by a NONMEM-style comment:

```text
DV ~ additive(ADD_ERR) weight = WT ; NONMEM comment
```

reports `magnitude ... Unexpected character: ;`, while

```text
if (WT == 0) { DV ~ additive(ADD_ERR) } else { DV ~ additive(ADD_ERR) } ; NONMEM comment
```

reports `unexpected content after covariate-selected if/else`. Both have a complete error-model form before the semicolon, so they meet the same gate as the tested plain/CMT/IIV forms, and the new docs explicitly say a line carrying `;` names the comment-marker problem. Please classify this shape before the selected/weight parsing (or teach those paths the same gate) and add both variants to the semicolon regression test.

Otherwise, the reject paths, source-line quoting, form-list pin, docs, and end-to-end coverage look good. Local verification: `cargo test --lib` passed (4,693 passed, 26 ignored), and the focused `check_command` regression passed.

roninsightrx:
Good catch — confirmed both, fixed in `210ad5f1`.

Reproduced before the fix, exactly as described:

```
$ ferx check w.ferx
error[E_PARSE]: [error_model] magnitude `(ADD_ERR) * (WT ; NONMEM comment)`: Unexpected character: ;

$ ferx check s.ferx
error[E_PARSE]: [error_model] unexpected content after covariate-selected if/else: `; NONMEM comment`
```

and after:

```
$ ferx check w.ferx
error[E_PARSE]: [error_model]: `;` does not start a comment in a `.ferx` file — use `#` or `//`.
The text after it is read as part of the line, not skipped.
Offending line: `DV ~ additive(ADD_ERR) weight = WT ; NONMEM comment`.

$ ferx check s.ferx
error[E_PARSE]: [error_model]: `;` does not start a comment in a `.ferx` file — use `#` or `//`.
The text after it is read as part of the line, not skipped.
Offending line: `if (WT == 0) { DV ~ additive(ADD_ERR) } else { DV ~ additive(ADD_ERR) } ; NONMEM comment`.
```

## What I did

You offered two options — classify before the selected/weight parsing, or teach those paths the same gate. I took the first, and deleted the inner copy rather than leaving it as a second line of defence. Two predicates rejecting the same inputs is the #1229 shape: deleting either leaves the suite green and neither can be mutation-tested.

`parse_error_model` is now a thin wrapper over `parse_error_model_inner`, with the gate ahead of it. The "is this a trailing comment?" test is the literal question — **delete from the `;` to the end of that line and does the block parse?** — answered by calling the inner parse. So every form is covered by construction and a form added to `ERROR_MODEL_FORMS` later needs no edit here; same reuse-the-source-of-truth trick as the `apply_fit_option` probe on the `[fit_options]` side. No recursion: the truncated block has no `;` left.

Embarrassing footnote — this is the *same* defect the `[fit_options]` gate had, which I'd already found and fixed one layer up (its `;` check had to move above the `=` split for the same reason). I fixed it there and didn't go looking for the twin. Your review found it; noted.

## Two things worth flagging, since I got both wrong first

**I broke it once while fixing it.** My first version sliced `stripped[..at]` — everything *before* the offending line — and threw away the lines after it. That scores `DV ~ proportional(P)` + `iiv_on_ruv = E ; c` on a block that had lost its statement. Now the offending line is truncated **in place** and the rest is kept. Comment at the call site says so.

**The gate has two arms, and I checked that claim rather than asserting it.** A `;` with nothing before it cannot fold into the delete-the-tail test: deleting that tail makes the block *shorter*, so a single-line block becomes empty and fails with "No error model found" instead of parsing. Mutating each arm separately reddens the test; so does dropping the hoisted call, and so does forcing the completeness oracle true. Three mutations, all named.

## Test

Driven off `ERROR_MODEL_FORMS` now, not three hand-picked line-oriented forms — sampling is precisely what let this through. Plus your two spellings pinned literally, so an edit to the array can't silently drop them.

The straddle is unchanged and still asserted from both sides: `banana ; x` and `DV ~ additive; (ADD)` have an incomplete head and get the form list, not a comment message.

One boundary is now pinned rather than assumed, because my first attempt at it **guessed the message and was wrong**: a block holding only `iiv_on_ruv = E ; c` is not an error model whatever the `;` does, so the comment message is correctly withheld — but it reports `unrecognized line` (with `iiv_on_ruv` in the form list), not the "No error model found" I'd predicted. Test now asserts what it measured.

## Re-measured after the change

- `ferx check` over all 184 `.ferx` files — **0 rejects**
- `--lib --features ci,survival` — **4881 passed**
- `check_command` + `model_text_roundtrip` — **40 passed**
- `tools/preflight.sh` — green

CI was green on all 11 checks at `5bdd09a1` (including `Tests + coverage (core)` and `codecov/patch`); re-running now on `210ad5f1`.

Issue #394, opened 2026-06-17
RATE=-2 (modeled infusion duration) support for analytical PK models


roninsightrx:
## RATE=-2 (modeled infusion duration) for analytical PK models

Extends `RATE=-2` → `D{cmt}` modeled-duration support (today ODE-only) to the analytical PK engine (1-cpt / 2-cpt closed forms, superposition + event-driven). Follow-up to the ODE work in #324.

NONMEM semantics to match: `F·AMT` delivered as a zero-order infusion over duration `D{cmt}`, i.e. `rate = AMT / D{cmt}`. `RATE=-1` (`R{n}`) stays out of scope (Phase B; not done for ODE either).

### Current state (already exists)

- `RateMode::{Fixed, ModeledDuration}` + `DoseEvent::modeled()` + `DoseEvent::resolve_rate()` (`src/types.rs`) — engine-agnostic, f64-only; single source of truth for the resolve rule.
- `DoseAttrMap` + `DoseAttr::from_indexed_name()` recognises `D{n}` → `(Duration, n)`.
- Reader emits `ModeledDuration` doses for `RATE=-2` (`io/datareader.rs`).
- ODE path resolves via `resolve_subject_doses[_with]` before integrating; `dose_attr_map` lives on `OdeSpec`, populated in `build_ode_spec`.
- Analytical path **rejects** modeled doses:
  - `api.rs::check_modeled_dose_rates` → `E_MODELED_DURATION_ANALYTICAL` when `ode_spec` is `None`.
  - `pk::compute_predictions` / `event_driven_predictions` `assert!(all_doses_fixed())`.

### Gap

1. Analytical models build no `DoseAttrMap` and route only canonical PK names (cl/v/q/v2/ka/f/q3/v3/lagtime) into `PkParams`. A `D{cmt}` individual parameter is computed but never stored in a slot, and nothing records its slot.
2. Nothing analytical-reachable holds a `DoseAttrMap` (`compute_predictions` takes only `pk_model, subject, pk_params`).
3. The two validation gates reject analytical outright.

### Design

[...]

PR #1399, opened 2026-09-15
fix(pk): repel a non-finite modeled D{n}/R{n} instead of serving a bolus [closes #1284]

roninsightrx:
## Why

`DoseEvent::resolve_rate` turned a modeled `RATE` code into a concrete `rate`/`duration` by reading the `D{cmt}`/`R{cmt}` slot and clamping it to a domain floor. The clamp is correct for its stated job — pull a transient `D ≤ 0` back off the wall so `amt/D` stays finite — but a **non-finite** value is not a small positive one, and it was given the same treatment:

| | resolved to | served |
|---|---|---|
| `D1 = NaN` | `duration = DURATION_FLOOR = 1e-8` (`NaN > floor` is false) | an instantaneous bolus |
| `D1 = +inf` | `rate = amt/inf = 0`, so `is_infusion()` is false | a bolus again |
| `R1 = NaN` | `rate = RATE_FLOOR = 1e-8` | `duration = amt/1e-8 = 1e10` — essentially nothing delivered |
| `R1 = +inf` | `duration = amt/inf = 0` | a bolus again |

On the issue's fixture (1-cpt, `CL/V = 0.1`, `V = 10`, one 100-unit `RATE=-2` dose, obs at `t = 1, 4, 8`) the `NaN` arm predicted `[9.048374185, 6.703200632, 4.493299458]` — within 5e-9 of the exact instantaneous-bolus solution `10·e^(−0.1t)` — against a correct infusion of `4.758129098` at `t = 1`. **2.03× high**, with no diagnostic on any channel: `fit()`, `ferx check` and `predict()` all returned finite numbers.

#1235 closed the *typical-value* reproducer (`E_DOSE_ATTR_NONFINITE` now rejects it at fit-init). What it cannot see is the mid-fit θ/η excursion into the same `exp` overflow — which is exactly where the clamp is supposed to be doing its job — so an excursion that should repel the optimizer instead handed it a plausible finite objective computed from a bolus.

This is the `f64::max`-eats-`NaN` shape `CLAUDE.md` calls out under "A green test is not evidence that it can fail", here in a production resolver rather than a test accumulator.

## What changed

**Option 1 from the issue, with one correction that measurement forced.** `resolve_rate` now tests finiteness before the clamp and returns `DoseEvent::non_finite()` for a non-finite `D{n}`/`R{n}`. The clamp itself (`clamp_above_floor`) is unchanged and still serves the absorption path.

The correction: the issue expected propagating the `NaN` through `(rate, duration)` to be enough for the timeline guard to repel the subject. Measured, it is not. `is_real_infusion` requires `rate > 0 && duration > 0 && duration.is_finite()`, and `is_infusion()`'s `rate > 0.0` is already false for a `NaN` rate — so **no assignment to that pair** both reads as an infusion and carries non-finiteness. The dose falls to the bolus branch, which reads only `amt`, and the wrong answer is unchanged.

So `non_finite()` poisons the dose's **`time`** as well. That is the field every engine already reads, and it is the same shape a `NaN` `ALAG` produces (`dose.time + lag`), so the #1189 machinery absorbs it with no new plumbing: each walk pushes `d.time` onto its timeline, `timeline_has_non_finite` / `EventSchedule::non_finite_event_time` sees it, and the subject comes back `NaN` — which the estimation guards already read as a diverged solve. One repulsion channel for every non-finite dose attribute, rather than a second one threaded through four walks.

**A second defect surfaced while verifying the first, and it had to be fixed for the first fix to hold.** The analytical *superposition* engine (`predict_concentration`) builds no timeline, so it has no `timeline_has_non_finite` to run. Both of its arms compare the dose's arrival `t_eff` against the observation time, and every comparison against `NaN` is false — so the dose was silently **dropped** and a drug-free trajectory returned as a valid prediction. With the `resolve_rate` fix alone, that engine went from "a bolus, 2.03× high" to "`0.0`, finite and infinitely wrong". And behind it sat a second floor: the trailing `conc.max(0.0)` is an `f64::max`, which discards `NaN` (`f64::NAN.max(0.0) == 0.0`, asserted in the test rather than recalled), so even a `NaN` that *did* reach the sum was floored to `0.0`.

`predict_concentration` now returns `NaN` on a non-finite `t_eff`. That path is reachable from a **`NaN` `lagtime`** too, independently of this issue — a pre-existing #1189 gap on that engine, closed here because this change walks straight through it.

## Alternatives considered

- **Relax `is_real_infusion`** to let a non-finite duration through to the timeline. Rejected: `is_infusion()` rejects a `NaN` rate first, so it would not even work, and it widens a predicate that also guards degenerate data-driven rows.
- **Poison `amt` instead of `time`.** Works (a `NaN` bolus poisons the state), but it does not reach observations *before* the dose and it misrepresents the event as "the engine delivered `NaN` mg". `time` is where the existing guards already look.
- **Options 2 and 3 from the issue** (record the clamp for a post-fit diagnostic; document today's behaviour). Both leave a 2× error reachable mid-fit with a finite objective on it.

## Cross-repo dependency

No public API change (`api/ferx-core-public-api.txt` unchanged; `DoseEvent::non_finite` is private, `predict_concentration`'s signature is untouched).

## Breaking changes

- [x] None

Behaviourally this is a numerical change on modeled-`RATE` models, but only where a `D{n}`/`R{n}` goes non-finite — see the bit-identity result below.

## Numerical validation

**(a) The exact closed form, not a tolerance.** An instantaneous bolus on the issue's fixture is `10·e^(−0.1t)` = `[9.0483741804, 6.7032004604, 4.4932896412]`. That is what the engine served, and the Tier-2 test carries those numbers so a regression names what it collapsed to. Every non-finite arm now returns `NaN` on both engines and both attributes (8 arms).

**(b) A differential pair straddling the gate.** Every arm is run against a finite in-domain control on the same fixture, asserted to pass `check_model_data`, predict finite, and sit far from the bolus it would collapse to — so a fix that simply rejected everything fails.

**(c) Bit-identity of a converging modeled-`RATE` fit.** 24 subjects, one `RATE=-2` dose each, 12 observations, FOCEI with IIV on `CL`/`V`/`D1` and a proportional error. Fitted with the release CLI on this branch, then with all three guards toggled off and rebuilt:

```
# before (guards off)          # after (this PR)
OFV: -842.1849                 OFV: -842.1849
TVCL = 4.741574                TVCL = 4.741574
TVV  = 52.857473               TVV  = 52.857473
TVD1 = 3.128156                TVD1 = 3.128156
```

`diff` of the two emitted `-fit.yaml` files is **two lines**, both wall-time. Every numeric field — objective, thetas, omegas, sigma, shrinkage — is byte-identical.

**No NONMEM anchor**, per the issue: NONMEM has no `D1 = NaN` to reproduce, and the quantity under test is what *this* engine substitutes for one. The anchors above are (a) a hand-computed closed form, (b) the straddle, and (c) bit-identity — which is what `CLAUDE.md`'s "hazard time outside the span NONMEM can express" exception asks for in this situation.

- [x] Compared against: exact closed form + prior ferx behaviour (guards toggled off, same binary configuration)

## Performance

`resolve_rate` gains one `is_finite()` per modeled dose per iteration. `predict_concentration` gains one `is_finite()` per (observation, dose) pair — the same loop that already evaluates a closed form per pair, so it is far below the noise floor of the work it guards.

## Tests

Three new tests, each mutation-run:

| Test | Tier | Mutation run | Result |
|---|---|---|---|
| `types::tests::dose_event_resolve_rate_repels_a_non_finite_duration_or_rate` | 1 | drop the `D` early return | `D1 = NaN: resolved time 3 is finite, so no timeline guard sees it` |
| | | drop the `R` early return | `R1 = NaN: resolved time 3 is finite …` |
| `pk::tests::non_finite_arrival_gives_nan_not_a_drug_free_zero` | 1 | drop the `t_eff` guard | `NaN lagtime: got 0, want NaN` |
| `non_finite_modeled_duration_or_rate_repels_instead_of_serving_a_bolus` | 2 | drop the `D` early return | `ODE/D1 = NaN: obs 0 predicted 9.048374184937554 … (the bolus this used to serve is 9.048374180359595)` |
| | | drop the `R` early return | `ODE/R1 = NaN: obs 0 predicted 9.5e-10 …` (the R-side damage: nothing delivered) |
| | | drop the `t_eff` guard | `analytical/D1 = NaN: obs 0 predicted 0 …` |

Each side of the two-engine twin fails on its **own** mutation and stays green on the other's — the ODE legs are blind to the `predict_concentration` guard and vice versa, so neither leg is an assertion against a constant.

Two existing tests changed: `dose_event_resolve_rate_clamps_nonpositive_{duration,rate}` pinned `NaN → floor` as intended behaviour. They now cover finite non-positive values only (`0.0`, `-3.0`, `-1e-9`) and point at the new test for the non-finite case. `clamp_above_floor_passes_through_or_clamps` is unchanged — the helper still sends `NaN` to the floor for the absorption path; it is the *caller* that now checks first.

Full suite green: `cargo test --features ci,survival` — 4875 lib tests + every integration target, exit 0.

## Docs

- `docs/model-file/ode-models.qmd` — a new bullet in the `D{n}` section (non-finite is not clamped, what happens at fit-init vs mid-fit, and what it used to do), and the symmetric clause for `R{n}`.
- `docs/file-formats/check-report.qmd` — the `E_DOSE_ATTR_NONFINITE` row documented the silent-bolus behaviour as current. It now reads as history, and says this check is the front door rather than the only line of defence.

## Checklist

- [x] Functions on the `Dual2` sensitivity path stay differentiable — `resolve_rate` is f64-only by construction (a subject with a modeled dose routes its inner gradient to FD, `resolve_gradient_method`, #394), so there is no `Dual` twin of this code to keep in step.
- [x] `Cargo.toml` version bump — not needed

## Example

## Reviewer hints

**Focus on `DoseEvent::non_finite`'s doc comment.** Poisoning `time` is the non-obvious part of this PR, and the justification is that the alternative is impossible rather than merely worse: walk `is_infusion()` → `is_real_infusion()` and check that no `(rate, duration)` assignment survives both. If you find one, this is the wrong fix.

**Second focus: `predict_concentration`'s guard is a scope widening.** It fixes a `NaN` `lagtime` on the superposition engine too, which is #1189 territory rather than #1284's. I included it because without it the analytical arm of this fix returns a finite `0.0` — the fix would be half-applied — but it is the one hunk here that touches behaviour beyond modeled `RATE`.

Skimmable: the CHANGELOG and docs prose, and the two existing tests whose `f64::NAN` entry moved to `-1e-9`.

## Open questions

`clamp_above_floor`'s other caller — `PreparedInputRate::MIN_PARAM` (transit `mtt`, inverse-Gaussian `mat`/`cv2`) — still sends a `NaN` to the floor. Same mechanism, different subsystem, and I did not touch it: the damage there is not a bolus and I have not measured what it is. Worth its own issue if you agree the reasoning generalises.


roninsightrx:
Requesting changes for two correctness gaps in the analytical superposition paths.

1. **[P1] Mirror the non-finite-arrival guard in `analytical_state_at_times` (`src/pk/mod.rs:2175`).** `DoseEvent::non_finite()` now poisons `time`, but only `predict_concentration` checks that sentinel. `analytical_state_at_times` still evaluates `t_eff <= t`; both comparisons are false for `NaN`, so it silently skips the dose and returns a drug-free zero state. I reproduced this through the public `compute_predictions_with_states` API with the PR's one-cpt analytical / `D1 = TVD*exp(WT) - TVD*exp(WT)` fixture: the result is `ipred = [NaN]`, `states = [[0.0]]`. This can leak a valid-looking compartment state to state/derived-output consumers even though the same subject was correctly repelled on the prediction path. Please give the state-superposition twin the same whole-subject non-finite outcome and add a regression covering the paired IPRED/state result.

2. **[P2] The new “every attribute” / `F` claims are not true on analytical superposition (`docs/file-formats/check-report.qmd:157`, `CHANGELOG.md:138`).** The new prose says non-finite bioavailability already makes predictions `NaN`, and the new `predict_concentration` comment calls out the trailing `conc.max(0.0)` swallowing a `NaN`; however, the added guard only examines `t_eff`. I reproduced an analytical `one_cpt_iv(..., f=F)` with `F = TVF*exp(WT) - TVF*exp(WT)` and `WT=1000`: `predict()` returns `[0.0]`, because `F` makes the dose contribution `NaN` and the unchanged final `max(0.0)` discards it. Either preserve a non-finite accumulated concentration before applying the nonnegative floor (with a regression for the mid-fit `F` case), or narrow the changelog/docs and `DoseEvent::non_finite` rationale so they do not claim parity with `F`.

Focused verification otherwise passed:

- `cargo test --lib non_finite` (38 passed)
- `cargo test --test modeled_duration non_finite_modeled_duration_or_rate_repels_instead_of_serving_a_bolus -- --exact` (passed)

The PR also currently conflicts with `main`; the conflict is confined to adjacent additions under `CHANGELOG.md`'s `Fixed` section and is mechanically resolvable by retaining both entries.

roninsightrx:
Both findings confirmed and fixed in `fccae065`; `main` merged in `4f68e656`.

Reproduced both before touching anything, and they are worse together than separately: the two superposition twins disagreed about the same subject in **opposite directions**, one per finding.

```
                          BEFORE                     AFTER
P1  D1 = NaN     ipred=[NaN]   states=[[0.0]]   →   ipred=[NaN]  states=[[NaN]]
P2  F  = NaN     ipred=[0.0]   states=[[NaN]]   →   ipred=[NaN]  states=[[NaN]]
```

That symmetry is what set the shape of the fix — two patches would have left the next walk free to drift again.

## 1 (P1) — one predicate, not a second guard

`analytical_state_at_times` ran its own copy of the `t_eff <= t` loop, which is the "two implementations" shape `CLAUDE.md` calls out, so I extracted the predicate rather than duplicating the check:

```rust
pub(crate) fn superposition_arrival_non_finite(doses: &[DoseEvent], lagtime: f64) -> bool {
    doses.iter().any(|d| !(d.time + lagtime).is_finite())
}
```

Read by both twins; the value side's guard is now hoisted out of the per-dose loop rather than evaluated inside it, so it is also one pass instead of *n*. Its doc says a third superposition walk must take this predicate rather than re-derive it, and records the measured `ipred=[NaN]` / `states=[[0.0]]` disagreement as the reason.

Regression added at both tiers, as asked: the Tier-1 `pk::tests` case asserts the state twin directly (plus a finite straddle), and the Tier-2 `modeled_duration` case asserts the **paired** result through `compute_predictions_with_states` — per observation, ipred and states must agree about the subject — across both engines and both attributes.

## 2 (P2) — fixed rather than narrowed

I took the first of your two options: the claim should be true, not softened. The floor now preserves a non-finite concentration and still floors a negative one (`+inf` needs no case — `f64::INFINITY.max(0.0)` is `inf`).

The regression is the mid-fit `F` case, and it asserts the thing that makes it *non*-redundant:

```rust
assert!(
    !superposition_arrival_non_finite(&doses, pk.lagtime()),
    "the arrival guard must NOT cover the F case — otherwise this test is \
     passing on the strength of a different gate"
);
```

Two gates rejecting the same inputs is a test hole rather than belt-and-braces, so that straddle is pinned explicitly.

You were also right that my `predict_concentration` comment described the floor while the guard did not address it — that comment is gone, and the floor now carries its own with the measured `F` numbers on it. The docs/CHANGELOG prose stands as written because the code caught up to it: `check-report.qmd`'s "every attribute" is now accurate on this engine, and `F` got its own CHANGELOG entry since it is a separate user-facing fix.

## Verification

| Mutation | Result |
|---|---|
| drop the guard in `analytical_state_at_times` | Tier-1 `state twin at time 0: got [0.0], want all NaN`; Tier-2 `analytical/D1 = NaN: states 0 = [0.0]` |
| restore the bare `conc.max(0.0)` | `NaN F: got 0, want NaN` |

Each side of the twin still reddens only on its own mutation — the earlier `resolve_rate` and `predict_concentration` mutations from the PR body were re-run and are unchanged.

**Bit-identity re-measured on the engine you flagged.** The PR's original check was an *ODE* fit, which never touches `predict_concentration` — so it said nothing about these hunks. Re-ran the same 24-subject `RATE=-2` FOCEI dataset against an analytical `pk one_cpt_iv` model, with and without both review fixes: OFV `-842.1845`, `TVCL 4.742000`, `TVV 52.855016`, `TVD1 3.128149`, and `diff` of the two `-fit.yaml` files is two lines, both wall-time.

`tools/preflight.sh` green on the merged tree, and `cargo test --features ci,survival` exits 0 with no failing target.

The `CHANGELOG.md` conflict is resolved by keeping both sides, as you described.

Issue #317, opened 2026-06-13
Slow-test regression: #312 eta-space inner EBE rescales the FD gradient step for mu-referenced models (4 NONMEM-anchored slow tests red on main)


TeunP:
## Summary

`main` is **red on the Slow tests workflow** ([run on `d255afb`](https://github.com/FeRx-NLME/ferx-core/actions/runs/27479191296/job/81223839979)). Four NONMEM-anchored slow convergence tests regressed. Bisected to **#312** (`fix(estimation): optimise inner EBE in eta-space`, closes #302): the [prior slow run on #312's merge](https://github.com/FeRx-NLME/ferx-core/actions/runs/27477075111) failed on the **same four tests with bit-identical values**, and the last green slow run was #280. (#310, which merged after, is parser-only and numerically inert — it just inherited the breakage.)

This is **not a request to revert** — #312 fixes a real, severe bug (#302 additive-IIV degeneracy). It's a follow-up to recover mu-ref accuracy on the FD path.

## Failing tests (FD path)

| Test | Assertion | Value vs NONMEM | Band |
|------|-----------|-----------------|------|
| `schnider_propofol_matches_nonmem` (`tests/schnider_propofol_nonmem.rs:138`) | theta TVV2 | **19.6697 vs 26.1000 — 24.6%** | 15% |
| `ltbs_warfarin_fit_converges_and_recovers_pk` (`tests/ltbs_convergence.rs:121`) | ω²(CL) | 0.025506 vs 0.0285942 — 10.8% | 10% |
| `ltbs_covariance_se_matches_nonmem` (`tests/ltbs_convergence.rs:193`) | SE(omega_V) | 0.005100 vs 0.003980 — 28.1% | 25% |
| `test_slsqp_moves_on_mu_referenced_two_cpt_oral_cov` (`src/estimation/outer_optimizer.rs:4624`) | OFV movement | -1134.87 vs init -1040 ("too close to init") | — |

All four are **mu-referenced / LTBS** fits — exactly the models #312's inner-EBE reparametrization touches.

## Root cause: eta-space reparametrization silently rescales the FD gradient step

[...]

PR #1400, opened 2026-09-15
fix(api): say so when `gradient = fd` moves the optimizer too [closes #1381]

### LINKED ISSUE #1381: optimizer = auto couples to gradient silently: one line changes two factors (#997 §4)
Split out of #997 §4. The **reporting** half of that item already shipped in #490 — `FitResult.optimizer` records the compound `"auto (nlopt_lbfgs)"`, so a run does say which algorithm executed. What remains is the coupling itself.

## The problem

Per `src/types.rs`, `optimizer = auto` resolves to `nlopt_lbfgs` when the analytic FOCE/FOCEI gradient is available and `bobyqa` when only finite differences are. So changing **one** line in `[fit_options]` — `gradient = auto` to `gradient = fd` — silently changes the optimizer too.

The reporter measured a 4.98 OFV difference that read entirely as a gradient effect. With the optimizer pinned to BOBYQA in both arms, the actual gradient effect on the same model was **0.38**. The other 4.6 was the optimizer.

That is not a wrong default — `auto` picking the optimizer that suits the available gradient is the right behaviour, and #490's benchmarking is what it rests on. The problem is that a user running a controlled one-variable experiment gets two variables moved, and nothing at the point of the change says so.

## Possible directions

- Warn when `gradient` is set explicitly *and* `optimizer` is left at `auto`, naming what `auto` resolved to and why — a warning at the point where the coupling bites, rather than a field the user has to think to read afterwards.
- Or resolve `auto` from the model alone and let an explicit `gradient` be orthogonal, which is a behaviour change and needs benchmarking.

## Notes

- #1380 (this issue's sibling from #997) adds `final_gradient_source`, which makes the *effect* of the coupling visible after the fact: an arm that silently became derivative-free now reports `"finite_difference"` instead of `"optimizer"`. That is a diagnostic, not a fix for the coupling.
- §3 of #997 is superseded by #978; §5 is filed separately.

roninsightrx:
## Why

Closes #1381.

`optimizer = auto` resolves to `nlopt_lbfgs` when the exact analytic outer gradient is available and `bobyqa` when only finite differences are. So changing **one** line in `[fit_options]` — `gradient = auto` to `gradient = fd` — changes **two** factors: the gradient the outer loop computes, and the optimizer that consumes it.

The reporter measured a 4.98 OFV difference that read entirely as a gradient effect. With the optimizer pinned to BOBYQA in both arms, the gradient accounted for **0.38**. The other 4.6 was the optimizer swap.

The default is not wrong — `auto` picking the optimizer that suits the available gradient is the point of `auto`, and it rests on #490's benchmarking across ~10 FOCEI datasets. What was missing is anything saying so *at the point the coupling bites*. `FitResult.optimizer` already records the compound `"auto (bobyqa)"` (#490), and #1380's `final_gradient_source` records the *effect*, but both are fields the user has to think to read afterwards — by which time the 4.98 has already been written down as a gradient effect.

The issue offered a second direction (resolve `auto` from the model alone and let `gradient` be orthogonal). That is a behaviour change needing its own benchmarking, and it would make `auto` resolve to a gradient-based optimizer on a fit whose loop runs FD — exactly the drift the shared predicate exists to prevent. Not taken.

## What changed

`check_model_options` emits a new warning, `W_AUTO_OPTIMIZER_FOLLOWS_GRADIENT`, naming the optimizer that ran, the one an unforced arm would use, and the pin that makes the comparison one-variable:

```
gradient = fd also changed the outer optimizer: with optimizer = auto left in place
it resolved to bobyqa rather than the nlopt_lbfgs this model would have used under
gradient = auto, because auto picks the derivative-free optimizer when there is no
exact analytic gradient to feed a gradient-based one. Two factors moved, not one —
if you are comparing an fd arm against an auto arm, pin optimizer = bobyqa in both
so the difference you measure is the gradient.
```

Warning severity, so it reaches `FitResult.warnings` (console + `{model}-fit.yaml`) and `ferx check` alike without refusing a legitimate request. **No resolution behaviour changes** — `resolve_auto` returns exactly what it returned before.

It fires only when the coupling actually bit. Four gates:

| Gate | Why |
|---|---|
| `optimizer == Auto` | a pinned optimizer moved one factor already — and pinning is the remedy the message recommends, so warning there would nag at the fix |
| `options.gradient_method == Fd` | the user's own request, not an engine-forced one. An SDE model has `model.gradient_method` overwritten to `Fd` by `fit()` whatever was asked for; no user-side coupling to report |
| `forced != unforced` | an out-of-scope model (ODE/PD, LTBS, TTE) resolves to `bobyqa` either way, as does a model above `BOBYQA_MAX_DIM`, where `auto` takes L-BFGS on an FD gradient regardless |
| the method gate | `saem`/`imp`/`impmap`/`bayes`/`vi` never run the outer optimizer; `laplace` has its `auto` overridden to `nlopt_lbfgs` before the outer loop sees it (#317). `gn_hybrid` *is* included — its FOCEI polish routes through `optimize_population_warm` → `optimize_population` with `auto` still live |

**The counterfactual is taken off the same resolver, not a second spelling of the rule.** `Optimizer::resolve_auto` is split into `resolve_auto_given_analytic(model, analytic)` with the predicate passed in, and `analytic_outer_gradient_available` is split into its scope half `analytic_outer_gradient_in_scope`. A hand-written `if in_scope { lbfgs } else { bobyqa }` counterfactual would have got the `BOBYQA_MAX_DIM` arm wrong and reported a switch that never happened — there is a test for exactly that arm.

## Alternatives considered

See **Why** — the "make `gradient` orthogonal" direction from the issue is a behaviour change that would let `auto` pick a gradient optimizer for a loop running FD.

## Cross-repo dependency

Nothing new is `pub`; both new items are `pub(crate)`. ferx-r surfaces `FitResult.warnings` already, so the warning reaches R callers with no change there.

## Breaking changes

- [x] None

## Numerical validation

- [x] Not applicable in the NONMEM-anchor sense: this adds a **diagnostic** and moves no numerical result. `resolve_auto` returns what it returned before, and the four arms below are bit-identical to `main`'s on the same inputs.

That said, the decomposition the warning claims was reproduced end-to-end on `data/warfarin.csv` (FOCEI, `covariance = false`, 10 subjects / 110 obs), one arm per cell:

| `gradient` | `optimizer` | resolved | OFV | warns |
|---|---|---|---|---|
| `auto` | `auto` | `nlopt_lbfgs` | −286.004220 | no |
| `auto` | `bobyqa` | `bobyqa` | −285.986587 | no |
| `fd` | `bobyqa` | `bobyqa` | −286.001454 | no |
| `fd` | `auto` | **`bobyqa`** | −286.001454 | **yes** |

- Compound one-line change (`auto`+`auto` → `fd`+`auto`): **0.0028**
- Gradient effect alone, BOBYQA pinned in both arms: **0.0149**
- Optimizer effect alone, at `gradient = auto`: **0.0176**

Warfarin is well-conditioned so the magnitudes are small, but the point survives: the compound number is **not** the gradient effect — here the two components partially cancel, so reading 0.0028 as "the gradient barely matters" would be wrong in both direction and size. The `fd`+`auto` and `fd`+`bobyqa` arms agree to all printed digits, confirming `auto` did resolve to `bobyqa`.

The warning text appears in the console `--- Warnings ---` block and in `{model}-fit.yaml`'s `warnings:` list for the fourth arm, and in none of the other three.

## Performance

- [x] No significant impact expected — two predicate evaluations once per fit, inside a check that already runs.

## Tests

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 4694 passed, 0 failed)

Nine Tier-1 tests in `src/api/tests/auto_optimizer_gradient_coupling_tests.rs`. Per CLAUDE.md, a green assertion is not evidence it *can* fail, so **each gate was mutated in turn** and the result checked — three of the four gates reject overlapping inputs, which is precisely the shape that hides a dead gate:

| Mutation | Test that went red | Others |
|---|---|---|
| drop `optimizer == Auto` | `an_explicit_optimizer_is_silent_under_fd` | green |
| drop `options.gradient_method == Fd` | `an_engine_forced_fd_is_silent_because_the_user_set_nothing` | green |
| drop `forced != unforced` | `an_out_of_scope_model_is_silent_…` **and** `above_the_bobyqa_dimension_threshold_…` | green |
| drop the method gate | `methods_that_never_consult_resolve_auto_are_silent` | green |
| remove the warning entirely | the three positive tests | green |

The second row is the one worth reading: `options.gradient_method == Fd` looks redundant next to the counterfactual, and its test asserts up front that the counterfactual *alone would* fire on that fixture — so the gate is load-bearing rather than belt-and-braces, and stays that way if someone edits it.

Two other holes were closed deliberately: `an_out_of_scope_model_…` asserts the **straddle** (its ODE fixture is out of scope *and* its analytic sibling is in scope), so it cannot silently decay into a second copy of the positive case if the analytic scope grows; and `resolve_auto_is_unchanged_by_the_split` pins the refactor, since a resolver split that changed the pick would change every default fit in the repo.

## Example

- [x] Not applicable (no new API surface; the `[fit_options]` keys involved already exist)

The user-facing shape is just the pin the warning recommends:

```toml
[fit_options]
gradient  = fd
optimizer = bobyqa     # pin it in BOTH arms, so the difference is the gradient
```

## Docs

  - `docs/estimation/optimizers.qmd` — new `### gradient = fd moves the optimizer too` under **The `auto` default`**, with the 4.98 / 0.38 decomposition, the pin, and when the warning stays silent
  - `docs/model-file/fit-options.qmd` — a note in the `auto` bullet of **Optimizer Choices**, linking to the above
  - `docs/file-formats/check-report.qmd` — the code's row in the error/warning table
- [x] `CHANGELOG.md` `[Unreleased] / Added` entry added
- [x] `src/diagnostics.rs` code table row added

`cargo run -p docs-lint -- --check` green; no new `.lint-baseline` entries.

## Checklist

- [x] `tools/preflight.sh` green — `preflight OK — fmt check clippy rustdoc docs public-api`, including `public API matches the committed baseline` (nothing new is `pub`)
- [x] Nothing on the `Dual2` sensitivity path changed behaviour: `analytic_outer_gradient_available` is split, not altered, and its two halves compose to the identical predicate

## Docs & examples

- [x] No new `.ferx` DSL syntax or `[fit_options]` key, no new example or data file → no ferx-site / ferx-book change
- [x] No example execution step needed (diagnostic only; the four warfarin arms above were run anyway and are unchanged from `main`)

## Reviewer hints

Focus on `src/api/validation.rs` — specifically whether the four gates are the right four, and whether the `options`-vs-`model` split (explicitness read off `options.gradient_method`, resolution read off `model.gradient_method`) is the right way round. It is what keeps SDE models quiet, and it is the one place the two could legitimately disagree.

`src/types.rs` and `src/sens/provider.rs` are mechanical splits — skim for whether the `&&` chain reassembles to the same predicate.

## Open questions

The warning does not fire when `gradient = fd` is set on a model already outside analytic scope, on the grounds that nothing moved. That is right for the optimizer, but such a user has also written a line that does nothing at all — worth a separate "this setting had no effect" notice? Deliberately not bundled here.


roninsightrx:

1. **[blocking] `ferx check` never sees the explicit FD gradient on an analytical model.** The new comparison at `src/api/validation.rs:3519` reads the live choice from `model.gradient_method`, but `validate_model_file` calls `check_model_options(&parsed.model, &parsed.fit_options)` at line 6023 without performing the model/options synchronization used by `fit_from_files` and `run_model_with_data`. `parse_full_model_file` leaves `CompiledModel.gradient_method` at `Auto`; parsing `gradient = fd` changes only `FitOptions.gradient_method`. Consequently, for the advertised `ferx check` case, both `forced` and `unforced` resolve to `nlopt_lbfgs` and `W_AUTO_OPTIMIZER_FOLLOWS_GRADIENT` is omitted. Please mirror the production gradient synchronization before validation and add a `validate_model_file`/CLI-path regression test, including the SDE forced-FD case so it remains silent.

2. **[blocking] Mixture fits get a false coupling warning.** The counterfactual uses the generic `Optimizer::resolve_auto` at `src/api/validation.rs:3519`, but the actual outer path special-cases mixtures at `src/estimation/outer_optimizer.rs:153-162`: `optimizer = auto` is always BOBYQA for a mixture, independent of analytic-gradient availability. On an otherwise analytic-scope mixture with `gradient = fd`, this check reports that the optimizer moved from `nlopt_lbfgs` to `bobyqa`, even though both the forced and unforced runs use BOBYQA. Please gate this case or centralize the real auto-resolution rule, and cover it with a negative test.

3. **[blocking] Evaluation-only FOCE/FOCEI fits also get a false warning.** `optimize_population` returns at `src/estimation/outer_optimizer.rs:137-138` when `outer_maxiter == 0`, before resolving or constructing any optimizer. The new check does not inspect `outer_maxiter`, so it still says the outer optimizer changed and names what “ran” even though no optimizer ran. This conflicts with the stated method gate and with the warning text. Please keep `maxiter = 0` silent and add a negative test.

Verification: `cargo test --lib auto_optimizer_gradient_coupling_tests` passed (9/9), and the full `cargo test --lib` suite passed (4694 passed, 26 ignored). The existing tests construct only already-synchronized models and do not exercise the file-validation entry point or the two optimizer-bypass paths above.

roninsightrx:
All three confirmed and fixed in 2f9d3aa7. Each was real, and they shared one cause worth naming: the check **restated** the `auto` resolution rule instead of consulting it. So rather than add three more gates, both sides of the comparison now go through the production functions.

---

### 1. `ferx check` never saw the explicit FD gradient — confirmed

Verified exactly as described: `parse_full_model_file` leaves `CompiledModel.gradient_method` at `Auto` (`model_parser.rs:3815` is its only write), and the stamp lives in `fit_from_files` / `run_model_with_data`, which `validate_model_file` does not run. Both arms resolved to `nlopt_lbfgs` and the advertised entry point said nothing.

Fixed with one rule instead of a third copy of the stamp: `GradientMethod::effective(model, options)` is now the single answer to "is this fit on finite differences?", and **both** production stamp sites write it (`fit.rs`, `run.rs` — they were already two copies of the same four lines).

It is deliberately the **union** of the three sources rather than a replay of the stamp, because the two unstamped callers fail in opposite directions:

- `ferx check` has the `fd` only in the *options* (reading the model alone → stale `Auto`);
- a Rust caller invoking `fit()` directly may have set only `model.gradient_method` (reading the options alone → misses it, and the loop honours the model).

Taking either source's `Fd`, plus `is_sde`, is correct in all three cases and idempotent on an already-stamped model.

**Confirmed at the real entry point**, not just the seam:

```
$ ferx check check_fd.ferx
warning[W_AUTO_OPTIMIZER_FOLLOWS_GRADIENT] fit_options:24: gradient = fd also changed
the outer optimizer: with optimizer = auto left in place it resolved to bobyqa rather
than the nlopt_lbfgs this model would have used under gradient = auto, …
ok: check_fd — no errors (1 warning(s))

$ ferx check check_auto.ferx      # identical model, `gradient` line deleted
ok: check_auto — no errors (0 warning(s))
```

It even points at the `gradient` line itself. Tests: `an_unstamped_model_still_warns_because_the_options_carry_the_fd` (unstamped model + `fd` in options) and the negative half you asked for, `an_sde_model_is_forced_to_fd_by_the_engine_and_stays_silent`, which also pins `effective` itself — a mutation making it ignore `is_sde` would let an SDE fit claim an analytic gradient it has no path for.

### 2. Mixture fits got a false coupling warning — confirmed

Right on both counts: `outer_optimizer.rs:153-162` overrides `auto` to BOBYQA for any mixture regardless of analytic-gradient availability, and `resolve_auto` knows nothing of that arm.

Taking the "centralize" option rather than the "gate" one. The mixture arm and the `auto` resolution now both live in `outer_optimizer::resolve_outer_optimizer(requested, model, has_mixture, analytic)`, returning the optimizer plus the downgrade warning. `optimize_population` calls it for its own resolution — so the live path and the check cannot describe different rules — and the check calls it twice, differing **only** in the predicate fed to it. On a mixture both calls return `bobyqa`, so the comparison is equal and the check is silent by construction rather than by an added condition.

`a_mixture_fit_is_silent_because_auto_is_bobyqa_either_way` asserts the straddle: the same model *without* the mixture must still warn, so the test cannot pass by the fixture having drifted out of analytic scope.

### 3. Evaluation-only fits got a false warning — confirmed

Also structural rather than a fourth gate: the `outer_maxiter == 0` short-circuit is now `runs_outer_optimizer(options)`, called both by the guard it came from and by the check. `an_evaluation_only_fit_is_silent_because_no_optimizer_runs` straddles it against `maxiter = 1`, so a mutation that merely silences the check cannot pass.

---

### On the verification note

Fair hit — the original fixtures were all stamped, and none touched the two bypass paths. Every gate has now been mutated, old and new:

| Mutation | Test that went red |
|---|---|
| `effective` → stale `model.gradient_method` | `an_unstamped_model_still_warns_…` |
| ignore the mixture arm | `a_mixture_fit_is_silent_…` |
| drop the `maxiter == 0` gate | `an_evaluation_only_fit_is_silent_…` |
| `effective` blind to `is_sde` | `an_sde_model_is_forced_to_fd_…` |
| drop `optimizer == Auto` | `an_explicit_optimizer_is_silent_under_fd` |
| drop `options.gradient_method == Fd` | `an_engine_forced_fd_…` + `an_sde_model_…` |
| drop `forced != unforced` | the out-of-scope, `BOBYQA_MAX_DIM` **and** mixture tests |
| drop `runs_outer` | the method-gate and eval-only tests |
| remove the warning | the three positive tests |

Each kills exactly the test written for it; nothing survives with the suite green. 13 tests total.

`cargo test --lib`: **4698 passed, 0 failed**. `tools/preflight.sh`: green, all six groups, public-API baseline unchanged (everything added is `pub(crate)`).

Docs updated to match: the silence list on `docs/estimation/optimizers.qmd` now names the mixture, eval-only and SDE cases alongside the two it already had, and says `ferx check` reports it.

### One thing I did not do

The three findings all argue for a single "what will this fit actually run" resolver, and this PR builds two thirds of one — `resolve_outer_optimizer` and `runs_outer_optimizer`. The `laplace` override still lives in `fit.rs`'s stage loop and is handled here by the method gate, which is a restatement of the same kind that caused all three findings. It is quieter (it changes `stage_opts.optimizer` before the loop, so nothing downstream can disagree), but it is the next thing that would drift. Happy to fold it in here if you'd rather, or leave it for a follow-up.

PR #1401, opened 2026-09-15
fix(estimation): hold a FIXed block_sigma correlation at the declared value, and report what the packer cannot represent [closes #1307]

roninsightrx:
## Why

`src/types.rs` documents `block_sigma (...) = [...] FIX` as "holds it at the declared value" (#847). It did not, above `|ρ| ≈ 0.995`.

`pack_rho` **clamps before any bound is consulted**:

```rust
rho.clamp(-RHO_CLAMP, RHO_CLAMP).atanh().clamp(-RHO_Z_BOUND, RHO_Z_BOUND)   // RHO_Z_BOUND = 3.0
```

so the Fisher-z *estimation* rail applied to a `FIX`ed correlation as well as a free one. Every declared `|ρ|` above `tanh(3) = 0.995055` collapsed onto that one number, and the fit scored a residual covariance the model file never wrote. Reproduced exactly as #1307's table reports it:

| declared ρ | `FIX` | packed | box | what the optimizer sees | moved |
|---|---|---|---|---|---|
| 0.990 | yes | 2.6467 | [2.6467, 2.6467] | 0.990000 | no |
| 0.995 | yes | 2.9945 | [2.9945, 2.9945] | 0.995000 | no |
| **0.999** | yes | **3.0000** | [3.0000, 3.0000] | **0.995055** | **yes** |
| **0.9999** | yes | **3.0000** | [3.0000, 3.0000] | **0.995055** | **yes** |

**And nothing reported it, structurally.** #1309's "is the packed start inside its own box" predicate cannot see this: `pack_params` clamps *first*, and `pack_with_bounds` then pins a `FIX`ed coordinate's box to the already-moved packed value — so `packed < lower || packed > upper` is false for every such coordinate **by construction**. A packed `3.0000` against a box of `[3.0000, 3.0000]` is perfectly in-box, and wrong.

The existing fixtures could not see it either: `tests/fit_result_residual_correlations.rs` and `tests/dense_residual_convergence.rs` both declare `[0.04, 0.10, 1.00]`, i.e. ρ = 0.5 — four rail widths inside.

### 1. The `FIX` contract — fixed

`pack_rho_fixed` packs `atanh(ρ)` with `RHO_CLAMP` only (guarding `atanh` against the unit boundary) and **no** `RHO_Z_BOUND`. `pack_params` routes to it on `residual_correlation_fixed`.

The rail is unchanged for a free ρ, and its own doc comment makes the case: on the 12-observation `examples/correlated_residual_combined` fixture a free ρ ran to −0.99993 and reported convergence at OFV −8.38 — a degenerate optimum, not an estimate. That argument is about what the optimizer may **search**. A `FIX` is not a search. Dropping the rail on the pinned path is safe precisely because the coordinate is pinned: it never moves, is never differentiated against the rail, and cannot walk the objective toward a singular `R`. It can make `R` singular where the user asked for it to be, which is what `FIX` is for.

### 2. The `1e-10` log floor — reported, not fixed

The issue's other half is structural, not a policy choice, and I did **not** try to fix it: `ln(0)` is `−inf`, which `compute_scale` would then divide a coordinate by. The issue itself allows this ("route a `FIX`ed 0 to identity packing **or** document it"). Routing to identity packing means changing `theta_packs_log`, which ~100 call sites across `saem`, `impmap`, `sens_outer_gradient`, `agq`, `vi` and the mu-referencing machinery read for their gradient chain — a blast radius out of all proportion to a defect whose measured effect is `θ = 0` vs `θ = 1e-10`.

So it is reported instead, and that reporting is shared with the free-ρ rail:

- `pack_params_with_moves` records every guard that binds, **at the one place the guard is applied**. `pack_params` is it with the second result dropped. A second walk re-deriving "where would the guard bind" is the two-implementations shape CLAUDE.md warns about, and worse than usual here since the thing re-derived is a silent clamp.
- `PackedStart` carries the list. It cannot be recovered from the other three fields: the box is built from the packed vector the guard already moved.
- `check_packed_start_in_box` emits `W_INIT_NOT_REPRESENTABLE`, naming the declared value and the value the optimizer actually sees.

Three deliberate choices in that check:

- **Its own warning code.** `WarningCode::InitNotRepresentable`, matched ahead of `w_init_outside_bounds` in `classify_warning` (which matches on substrings, and this message *names* the sibling token when it explains the silence). The two support opposite conclusions: a silent `W_INIT_OUTSIDE_BOUNDS` must not be read as "my declared values reached the optimizer".
- **Not exempt at `maxiter = 0`.** `evaluate_at_initial_params` unpacks the same vector, so an eval-only run scores the substituted value too — there is no run of any length in which the declared number is used.
- **It reports only what the box walks do not claim.** Not tidiness — a correctness point. A coordinate the box *also* moves is moved twice (`sigma X ~ 0.0 (sd)` floors to an SD of `1e-10`, then clamps onto the `exp(-8) = 3.355e-4` rail), so naming the packer's output would name a number the fit never uses. What is left over is exactly the set nothing else can see, which is #1309's own list of its blind spots: a held coordinate, a θ whose declared lower is itself at or below the floor, and a free ρ (which lands *on* its rail, not outside it).

### 3. `THETA_PACK_FLOOR` → `LOG_PACK_FLOOR`

Five other packed segments share the constant and carried it as a bare `1e-10` literal, which is why they could not be reported on. `THETA_PACK_CEIL` keeps its name — the ceiling really is θ-only.

## Alternatives considered

- **Route a `FIX`ed 0 θ to identity packing.** Rejected, see §2 — ~100 `theta_packs_log` call sites, all on gradient-chain paths.
- **Make `unpack_params` read a `FIX`ed θ from the template.** Exact, and much smaller, but it makes the packed vector stop determining the parameters for those coordinates — a second implementation by another name, and one that would silently change behaviour for any caller that writes a fixed slot.
- **Make the pack-move report an error rather than a warning.** Rejected on evidence: it would refuse `theta TVBETA(0.0, FIX)` in `nonmem_anchor/ss_chz_const_fit.ferx`, the ferx spelling of NONMEM's perfectly legal `$THETA 0.0 FIX`, which runs fine at `1e-10`.

## Cross-repo dependency

`WarningCode` gains a variant, which is additive; ferx-r reads the flat warning strings and does not match exhaustively on the enum.

## Breaking changes
- [x] Public Rust API — one added item: `WarningCode::InitNotRepresentable`, regenerated via `tools/update-public-api.sh`. **Who needs it:** every `FitResult` consumer that classifies warnings — the R wrapper and the CLI both branch on `WarningCode`, and reusing `InitOutsideBounds` would make "no `init_outside_bounds`" mean two incompatible things. The existing surface does not suffice for the reason the whole issue exists: the two conditions are distinguishable only by code, since every coordinate the new one names is in-box.
- [ ] None

<details>

A fitted model with a `FIX`ed `block_sigma` above `|ρ| = 0.995` had its objective and estimates computed at `0.995055` and needs re-running. Below that, and for every free `block_sigma`, results are bit-identical.

</details>

## Numerical validation

**NONMEM anchor** — `nonmem_anchor/correlated_residual_rho999.ctl`, `$SIGMA BLOCK(2) FIX` at ρ = −0.999, NONMEM 7.5.1 `METHOD=1 MAXEVAL=0`. Every THETA and the OMEGA are fixed too, so the objective is a pure function of the declared Σ and nothing can absorb a substituted ρ.

```
NONMEM 7.5.1    442.27244730844870
ferx, after     442.138163           |Δ| =   0.134   (0.03%)
ferx, before    301.450784           |Δ| = 140.822
```

Separation is **1050×** the residual disagreement. The 0.134 is this fixture's ordinary FOCE inner-EBE gap — its ρ = 0.5 sibling sits at 0.005 on an OFV of 18.7, the same order relative to the objective's scale. Tolerance is 0.5: 3.7× the realised error, 280× below the defect.

**The first anchor was discarded, and the file says why.** At ρ = +0.999 with the sibling fixture's `σ_prop = 0.2 / σ_add = 0.3`, measured: NONMEM 14.155997, ferx 14.175702 after, **14.1636 before** — the *pre-fix* value was the closer of the two. The cross term `2·IPRED·ρ·σ₁·σ₂` is a twentieth of the diagonal there, so the whole ρ = 0.995 → 0.999 effect is 0.012 OFV against a 0.005–0.02 baseline gap: an oracle measuring itself. The committed anchor balances the block (`σ_prop·IPRED ≈ σ_add`) so the residual variance nearly cancels as ρ → −1. Both measurements are recorded in the test file's module doc.

## Performance
- [x] No significant impact expected — `pack_params` gains an empty `Vec` and a comparison per coordinate; the `Vec` never allocates unless a guard binds.

## Tests

**Tier 1** (`src/estimation/parameterization.rs`, `src/api/tests/packed_start_box_tests.rs`) — the FIX contract across the issue's own ρ table; the free-ρ straddle; every log-packed segment's floor report, one segment at a time and on the declared scale (a variance for an Ω diagonal, not its Cholesky SD); the NaN case; bit-for-bit `pack_params` vs `pack_params_with_moves`; the unit-boundary guard; the FIXed θ and free θ blind spots with *why the other walks are silent* asserted; the partition against the box check; the eval-only straddle; `classify_warning` routing.

**Tier 2** — `tests/fit_result_residual_correlations.rs` (`fit()` reports the declared ρ and reassembles the declared covariance; the free straddle warns) and the new `tests/correlated_residual_rho_nonmem_anchor.rs`.

**Corpus** — `only_one_shipped_model_file_declares_a_value_the_packer_cannot_hold`: over all 179 `.ferx` files that parse, exactly **one** declaration is claimed (`theta TVBETA(0.0, FIX)`). The check is not noise.

### Mutation

Every edit that removes part of the fix reddens a **named** test, and each was run:

| mutation | dies |
|---|---|
| reinstate `RHO_Z_BOUND` on the fixed path | `a_fixed_rho_is_held_at_the_declared_value` (T1), `a_fixed_block_sigma_correlation_reaches_the_optimizer_unchanged` (T1), `a_fixed_block_sigma_correlation_above_the_rail_is_held_at_the_declared_value` (T2), all 3 anchor tests |
| drop the rail on the **free** path too | `a_free_rho_still_takes_the_estimation_rail_and_reports_it`, `instrumenting_the_packer_did_not_move_any_packed_value`, `a_fixed_block_sigma_correlation_reaches_the_optimizer_unchanged`, `a_free_block_sigma_correlation_above_the_rail_is_railed_and_reported` |
| drop `RHO_CLAMP` from `pack_rho_fixed` | `a_fixed_rho_at_the_unit_boundary_is_still_finite` |
| `value < FLOOR` instead of `!(value >= FLOOR)` | `a_nan_declared_value_is_reported_rather_than_swallowed` |
| report an Ω diagonal on the SD scale | `the_pack_reports_every_log_floor_it_applied` |
| drop the `claimed` partition skip | 5 tests, incl. `a_coordinate_the_box_also_moves_is_left_to_the_box_check` |

Three existing tests changed their expected output, and all three had comments *anticipating* #1307 (`"the floor moving it to 1e-10 is #1307's object, not this error's"`). They now assert the error is absent and the new warning present, rather than `is_empty()`, so the two facts stay separable.

- [x] `cargo test --features ci,survival`: 194 binaries, **5862 passed, 0 failed**, 187 ignored.

## Example

```
[parameters]
  block_sigma (PROP_ERR, ADD_ERR) = [0.0009, -0.008991, 0.09] FIX   # rho = -0.999
```

Before: fit at ρ = −0.995055, OFV 301.451. After: fit at the declared ρ = −0.999, OFV 442.138 (NONMEM 442.272).

A free block past the rail now says so:

```
warning[W_INIT_NOT_REPRESENTABLE]: ADD_ERR~PROP_ERR is declared as 9.99e-1 but the
optimizer sees 9.950548e-1 — an estimated residual correlation is bounded at |ρ| ≤
0.995055 (Fisher-z |z| ≤ 3) so the likelihood cannot chase `log|R| → −∞` through a
singular R. … start ADD_ERR~PROP_ERR at a correlation inside (−0.995055, 0.995055), or
write the block `FIX` if the declared correlation is an assertion rather than a starting
point — a `FIX`-ed correlation is held exactly, with no rail.
```

- [x] Example added inline (the anchor `.ferx` and `.ctl` ship under `nonmem_anchor/`)

## Docs
- [x] `docs/model-file/error-model.qmd` — the rail applies to an estimate, not a declaration, with a "Changed in #1307" callout carrying the measured OFV numbers.
- [x] `docs/estimation/parameterization.qmd` — new `#### What the box cannot see: values the packer moved`.
- [x] `docs/file-formats/check-report.qmd` — `W_INIT_NOT_REPRESENTABLE` row. The codes table was split into three `#####` subsections because the new row pushed it past docs-lint R1's 5,000-char limit — split, not baselined.
- [x] `CHANGELOG.md` `[Unreleased] / Fixed`, two entries.

## Checklist
- [x] Nothing on the `Dual2` path changed; `rho_chain` is untouched and its FD parity test still passes.
- [ ] `Cargo.toml` version bump — not needed (additive enum variant)

## Docs & examples
- [x] No new DSL syntax or `[fit_options]` key; no new `examples/` model; no new `data/` file. The two new files are NONMEM-anchor fixtures under `nonmem_anchor/`, which ferx-site and ferx-book do not index.
- [x] No example execution step needed — no shipped `examples/*.ferx` changes behaviour. `examples/correlated_residual_combined.ferx` declares ρ = 0.5 and is bit-identical (verified: OFV 18.7274 before and after).

## Reviewer hints

**Focus on the partition in `check_packed_start_in_box`.** It is the one place a second gate could quietly make a test unfalsifiable. The `claimed` set is built from the *same iterators* the walks below use, so a coordinate cannot fall between them, and `a_coordinate_the_box_also_moves_is_left_to_the_box_check` asserts the skip is a skip (`start.moves` non-empty) rather than a miss.

**Worth a second opinion:** whether `W_INIT_NOT_REPRESENTABLE` should be an *error* for a held coordinate. I made it a warning because the only shipped model it fires on is NONMEM's legal `$THETA 0.0 FIX` spelling — but the argument that a broken `FIX` contract deserves an error is a real one.

Skimmable: the `THETA_PACK_FLOOR` → `LOG_PACK_FLOOR` rename (mechanical, 17 sites) and the `moves: _` additions at the five `PackedStart` destructure sites.

## Open questions

The free-ρ rail's *severity* is still open, as #1307 says — a start at 0.999 clamped to 0.995 may be immaterial once the optimizer runs, and I have not measured it. This PR only ends the silence there; it does not change the rail.


roninsightrx:
Review result: changes requested. I found two correctness gaps.\n\n1. [P2] A parser-accepted FIXed correlation can still be rewritten (src/estimation/parameterization.rs:159-160).\n\npack_rho_fixed still clamps every rho above RHO_CLAMP = 0.999999, although the parser accepts every finite |rho| < 1 and the new public docs promise exact holding over that whole range. I reproduced this with:\n\n    block_sigma (PROP_ERR, ADD_ERR) = [0.04, 0.059999994, 0.09] FIX\n\nThe parser stores rho = 0.9999999, but an outer_maxiter = 0 fit reports rho = 0.999999. The emitted warning also says the declaration is "exactly +/-1" and recommends declaring it strictly inside (-1, 1), even though 0.9999999 already is. Please make the unit-boundary fallback conditional on an actually inadmissible boundary/out-of-range input, while leaving every parser-accepted rho untouched, and add a regression above 0.999999.\n\n2. [P2] The claimed set can suppress every start diagnostic at maxiter=0 (src/api/validation.rs:4526-4532, before the early return at 4626).\n\nclaimed is populated from coordinates_outside_bounds and theta_outside_declared_range even when outer_search_runs(options) is false. Those later diagnostics are skipped by the early return, but their indices still suppress W_INIT_NOT_REPRESENTABLE. For example, theta TVCL(-5.0, 0.0, 10.0) has a ValueFloor move and a declared-range violation: the move is filtered as claimed, then maxiter=0 returns before E_THETA_INIT_OUTSIDE_BOUNDS, so check reports nothing even though evaluation uses 1e-10 rather than -5. A free sigma X ~ 0.0 (sd) has the analogous double-move blind spot. This contradicts the new guarantee that pack substitutions are reported even for evaluation-only runs. The partition should be based on diagnostics that will actually be emitted (and for double moves should name the final box-clamped value).\n\nVerification performed:\n- cargo test --lib parameterization --features ci,survival\n- cargo test --lib packed_start_box_tests --features ci,survival\n- cargo test --test fit_result_residual_correlations --features ci,survival\n- cargo test --test correlated_residual_rho_nonmem_anchor --features ci,survival\n- RUSTC_WRAPPER= tools/update-public-api.sh --check\n\nAll of those passed. The two reproductions above were additional temporary regression tests and failed as described. The PR also currently has merge conflicts with main in CHANGELOG.md and docs/file-formats/check-report.qmd.

roninsightrx:
Both findings confirmed and fixed in `478521ba`; merge conflicts resolved in `f20cedef`. Thank you — the first one is a defect I introduced *while fixing the same class of defect*, which is worth naming.

## 1. `pack_rho_fixed` still clamped at `RHO_CLAMP` — confirmed, fixed

You are right, and the framing is the useful part: this was a **second, much narrower rail on the path the PR had just cleared**. I moved `RHO_Z_BOUND` off the `FIX` path and left `RHO_CLAMP` sitting on it unconditionally, so the guarantee I then wrote into `src/types.rs` and `docs/model-file/error-model.qmd` — "held exactly, at any `|ρ| < 1` the parser accepts" — was false from `0.999999` upward. The whole point of #1307 is that a promise with an undocumented edge is the problem, so shipping a new edge 4.5 orders of magnitude further out would have been the same bug with better numbers.

The unit-boundary message compounded it exactly as you describe: it told the user to declare a value strictly inside `(-1, 1)` that already was.

**Fix.** `atanh` is finite across the *whole* open interval — `atanh` of the largest `f64` below 1 is `18.71`, asserted in the test rather than recalled — so nothing short of `|ρ| = 1` needs backing off. The clamp is now conditional on an actually inadmissible input:

```rust
pub(crate) fn rho_within_unit(rho: f64) -> f64 {
    if rho.abs() >= 1.0 { rho.clamp(-RHO_CLAMP, RHO_CLAMP) } else { rho }
}
```

Spelled once and shared by `pack_rho_fixed` and the `PackMove` walk, so "did the guard bind" cannot drift from the guard. `RhoUnit` now fires only at `|ρ| >= 1`, and its message says "at or beyond ±1", which makes the existing remedy correct rather than absurd.

Free-path behaviour is unchanged by construction: `pack_rho` is `pack_rho_fixed(ρ).clamp(±RHO_Z_BOUND)`, and everything the old clamp touched was already past the `±3` rail.

**Regression** — `a_fixed_correlation_above_the_old_unit_clamp_is_still_held_exactly`, on your exact repro (`[0.04, 0.059999994, 0.09] FIX`, ρ = 0.9999999). It asserts `declared > 0.999_999` first, so it cannot become a tautology if the constant moves. Reverting the fix to the unconditional clamp reddens it.

## 2. `claimed` suppressing every start diagnostic at `maxiter = 0` — confirmed, fixed

Correct, and this one is worse than a missing message: it broke the guarantee the arm was added to make. `claimed` was built unconditionally from two walks that sit **after** the `maxiter = 0` early return, so their indices suppressed `W_INIT_NOT_REPRESENTABLE` while producing nothing themselves. `theta TVCL(-5.0, 0.0, 10.0)` at `maxiter = 0` reported *nothing at all* on a run that scores `1e-10`.

I had reasoned about the partition as "is this coordinate claimed" when the actual predicate has to be "**will a diagnostic actually be emitted** for it". Those coincide only when a search runs.

**Fix**, your wording taken literally:

```rust
let mut claimed: BTreeSet<usize> = inverted.iter().copied().collect();
if outer_search_runs(options) {
    claimed.extend(coordinates_outside_bounds(&start, &kinds).map(|h| h.index));
    claimed.extend(theta_outside_declared_range(init_params).map(|h| h.index));
}
```

One thing I checked rather than assumed, since it is the gate's only load-bearing assumption: `coordinates_outside_bounds`' below-a-variance-rail hits are claimed here although *this* function skips them — they are emitted by `check_variance_init_rails`, which carries the identical `outer_search_runs` exemption, and both call sites (`api::fit` and `validate_model_file`) run it alongside this check. So the gate holds for that arm too. Recorded at the call site.

**Double moves now name the final value.** `sigma PROP_ERR ~ 0.0 (sd)` reports `3.354626e-4` (the `exp(-8)` rail), and the cause names both steps rather than blaming the floor for a number the floor did not produce:

> PROP_ERR is declared as 0e0 but the optimizer sees 3.354626e-4 — ferx packs it on the log scale, and `ln` has no value at or below 0, so the packer substitutes a floor of 1e-10, **which the optimizer's own box then clamps to 3.354626e-4**. …

The number is obtained by running the **real** `unpack_params` over the clamped vector and reading `coordinate_values`, rather than back-transforming each kind at the call site — I did not want a second copy of the packing convention here, which is the trap the rest of this PR is about. It is built lazily (`start.moves` is empty on essentially every fit) and clamps per coordinate while skipping `inverted`, because `clamp_to_bounds` panics on an unorderable box and `inverted` is exactly that set.

Incidentally this makes the partition a *presentation* choice rather than a correctness one: with the final value named, a double report would have been two true statements, not one true and one misleading. It still stands down to avoid saying the same thing twice.

**Regressions** — two, each failing on its own mutation:

- `a_doubly_claimed_coordinate_is_still_reported_when_the_other_walk_is_exempt`, on your `theta TVCL(-5.0, 0.0, 10.0)` case. Asserted as a straddle: eval-only must report `W_INIT_NOT_REPRESENTABLE`, a searching run must report `E_THETA_INIT_OUTSIDE_BOUNDS` and exactly once. Without the second half it would pass for a check that had merely started double-reporting.
- `a_double_move_names_the_value_the_fit_actually_uses`, on your `sigma X ~ 0.0 (sd)` case.

Removing the `outer_search_runs` gate reddens both; replacing `seen` with the packer's intermediate reddens the second alone.

Worth recording: I first wrote that test asserting `3.355e-4`, the value my own prose had been rounding to throughout the PR. It failed against the realised `3.354626e-4`. The measured number is now in the assertion and the comment.

## Merge conflicts

Resolved in `f20cedef`. Both sides of `CHANGELOG.md` were pure additions under `### Fixed`, and `docs/file-formats/check-report.qmd` auto-merged; the resolution script asserts the two sides carry the same *set* of bullets before taking a side, so nothing could be dropped by picking one. I also tightened the `CHANGELOG` and `error-model.qmd` wording to say "the whole open interval" rather than a range that has an edge — the claim that was wrong above.

## Verification

- `tools/preflight.sh` — green (`fmt check clippy rustdoc docs public-api`).
- `cargo test --features ci,survival` — 194 binaries, **5888 passed, 0 failed**, 187 ignored, against merged `main`.
- The NONMEM anchor is unaffected (ρ = −0.999 is well inside the old clamp): ferx 442.138163 vs NONMEM 442.272447.

One thing I did **not** change, flagged in case you disagree: `W_INIT_NOT_REPRESENTABLE` is still a *warning* for a held coordinate whose value the log floor cannot represent. The only shipped model it fires on is `theta TVBETA(0.0, FIX)` in `nonmem_anchor/ss_chz_const_fit.ferx` — NONMEM's perfectly legal `$THETA 0.0 FIX` — and I would rather not refuse a working model to enforce a contract the packing genuinely cannot keep. The argument for an error on a broken `FIX` is real; say so and I will change it.

roninsightrx:
## Second review pass (`/code-review medium`) — 3 findings, 2 fixed in `2d4999e0`, 1 rejected

Reviewed `main...HEAD` across the four stacked changes on this branch. Posting the findings and their disposition, including the one I declined and why.

### Verified correct, no finding

- `pack_rho_fixed` drops the ±3 rail only for a `FIX`ed ρ; `packed_fixed_mask` covers the ρ segment and `pack_with_bounds` pins `lower == upper == packed[i]`, so the unrailed `z` is never clamped back. `compute_scale_packed`'s `max(|z|, 1)` stays sane at `z = 3.8`.
- No other site pairs `pack_params` with an unpinned box.
- The `f64::clamp` panic hazard in the new `effective` vector is covered — `coordinates_with_inverted_bounds` catches `NaN` bounds too, and those indices are skipped.
- `classify_warning`'s new token arm sits ahead of `w_init_outside_bounds` and every prose arm, so the message's own mention of the sibling token cannot misclassify into `BoundaryEstimate` (which would silently drop bootstrap replicates).

### 1. `omega ETA_X ~ 0 FIX` warns with an inapplicable remedy — **rejected, does not reproduce**

The claim: `W_INIT_NOT_REPRESENTABLE` fires on the zero-variance-`FIX` idiom, whose `ValueFloor` remedy ("remove it from the model") cannot be followed when `[individual_parameters]` references the eta.

This was worth taking seriously — the idiom appears **33 times across 11 files** in this repo, and `E_OMEGA_INIT_AT_RAIL`'s own documentation calls `FIX` *"the supported way to declare no variability"*. A warning contradicting that would be a real defect.

It does not fire. Measured:

```
$ ferx check zerovar.ferx          # omega ETA_CL ~ 0.0 FIX
ok: zerovar — no errors (0 warning(s))
```

**Mechanism.** `OmegaMatrix::from_matrix` (`src/types.rs:1898`) fails Cholesky on a zero-variance diagonal, regularises with an eigenvalue floor of `1e-8`, and stores `chol[(i,i)] = sqrt(1e-8) = 1e-4` — six orders of magnitude above `LOG_PACK_FLOOR = 1e-10`. `floor_for_log` never binds, so no `PackMove` is recorded and nothing is reported. (That `1e-8` regularisation is #1242's object and deliberately out of scope here.)

The finding traced the code path without executing it. No change made.

### 2. `as_variance` is wrong for a block Ω — **confirmed, fixed**

`as_variance(x) = x*x` was applied to every Ω / Ω_IOV Cholesky diagonal, but `L[i,i]²` is the declared variance only when Ω is **diagonal**. On a block, the declared variance of eta `i` is `Σ_k L[i,k]²`, so `L[i,i]²` is what is *left* of it once the off-diagonals are accounted for — and quoting that as the declaration names a number appearing nowhere in the model file.

This is precisely the trap #1309's review already fixed in the sibling box message (the `block_omega` arm that says "…`L²` = …, which is what is left of {name}'s variance once the off-diagonals are accounted for, not the variance declared for it"). I did not carry the precedent across.

**Fix.** The Ω and Ω_IOV branches pick their reporting map from the `diagonal` flag; a `[mixture]` override is always a diagonal scalar and keeps `as_variance`.

Reachability is essentially nil and the comment says so rather than implying a live bug: it needs a declared block that is positive-definite (so the `1e-8` regularisation above does *not* fire and lift every diagonal to `1e-4`) yet has a Cholesky diagonal below `1e-10` — correlated to within `1e-20` of singular while still surviving `cholesky()`. The branch exists so the message cannot be wrong, not because the case is expected.

Worth noting the fixture already covered it: `make_all_segments_template` carries a **block** BSV Ω and a **diagonal** Ω_IOV, and `the_pack_reports_every_log_floor_it_applied` *failed* on the old expectation when the fix landed — so both branches are exercised rather than asserted.

### 3. Two `PackMove`s on one ρ, the second with a fabricated `declared` — **confirmed, fixed**

Both guards can bind on the same coordinate: a free `ρ = 1.5` is backed off the unit boundary to `0.999999` and then railed to `0.995055`. Pushing one move each produced two warnings for one coordinate, and the second announced `declared: 0.999999` — the post-back-off intermediate, a number the model never wrote.

**Fix.** At most one move per ρ, carrying the value the model actually declared and the packer's final output. The guard recorded is the **first** to bind, because that is the one the remedy follows from — an inadmissible declaration is repaired by declaring an admissible one, whatever the rail then does. The `RhoUnit` message names the rail step as well when the coordinate is free, so the cause explains the number it quotes.

**A bug I introduced while fixing this, caught before it ran.** My first attempt made the guard predicate natural-scale — `unpack_rho(z) != corr.rho`. That is wrong for *every model*: `tanh(atanh(ρ))` is not the identity in floating point, so an untouched `ρ = 0.5` comes back differing in the last ULP and would have reported a spurious move on every correlated fit. Both predicates now stay in packed space, where a guard either applied or did not, and `an_untouched_rho_records_no_move_although_the_round_trip_is_inexact` pins it — asserting the round trip really is inexact somewhere, so the test cannot pass vacuously.

### Mutation

Each fix has its own failing test; all three run:

| mutation | dies |
|---|---|
| `declared: rho_within_unit(corr.rho)` (the intermediate) | `an_inadmissible_rho_records_one_move_carrying_the_declared_value` |
| `rail_bound = unpack_rho(z) != corr.rho` (natural scale) | `an_untouched_rho_records_no_move_although_the_round_trip_is_inexact` |
| `omega_report = as_variance` (unconditional) | `the_pack_reports_every_log_floor_it_applied` |

### Verification

- `tools/preflight.sh` — green (`fmt check clippy rustdoc docs public-api`).
- `cargo test --features ci,survival` — 194 binaries, **5890 passed, 0 failed**, 187 ignored.
- NONMEM anchor unchanged: ferx 442.138163 vs NONMEM 442.272447.
- CI green on `2d4999e0`, `codecov/patch` pass.

### Still open for a human call

`W_INIT_NOT_REPRESENTABLE` is a **warning**, not an error, for a held θ whose value the log floor cannot represent. Making it an error would refuse `theta TVBETA(0.0, FIX)` in `nonmem_anchor/ss_chz_const_fit.ferx` — NONMEM's perfectly legal `$THETA 0.0 FIX`, which ships here and works. I would rather not refuse a working model to enforce a contract the packing genuinely cannot keep, but the argument the other way is real.

Issue #1009, opened 2026-08-20
A CMT-less dataset silently doses compartment 1; NONMEM's DEFDOSE has no ferx equivalent on the fitting path


TeunP:
Raised by the ferxtranslate session while translating NONMEM control streams (translator#27 is their side of it). Verified against ferx-core `main` before filing; they explicitly did *not* ask for a ferx-core change, so this is a gap report rather than a request.

## The gap

A dataset with no `CMT` column doses **compartment 1** in ferx. The reader defaults it in three places:

```
src/io/datareader.rs:1618
src/io/datareader.rs:1748
src/io/datareader.rs:1868
        .unwrap_or(1);
```

NONMEM defaults such a dose to the compartment carrying the `DEFDOSE` attribute in `$MODEL`, which is **not necessarily compartment 1**. `$MODEL COMP=(PERIPH) COMP=(DEPOT, DEFDOSE) COMP=(CENTRAL)` doses compartment 2.

There is no way to express that in ferx for a **fit**. `dose_cmt` exists, but it is a field on `SimulationSpec`:

```
src/types.rs:6491        pub dose_cmt: usize,
```

parsed only from the `[simulation]` block (`model_parser.rs:5331`, `:5364`, `:5411`) and read only by the simulation path (`api/run.rs:353`). The fitting path never consults it. So a model whose intended default dose compartment is not 1 has **no correct ferx target** when its dataset omits `CMT`.

## Why it matters

The failure is silent and it is a dosing error, so every prediction for that subject is wrong — not off by a scale factor, but dosing the wrong compartment. It has the same shape as #899 (an out-of-range dose CMT silently dropped on the ODE engine) and #262 (no doses parsed when the dataset lacks `EVID`): the dataset is accepted, the fit converges, and nothing says the doses did not land where the model author intended.

For a hand-written ferx model the exposure is small — the author picks `states=[...]` and usually puts the dosed compartment first. It bites on **translated** models, where `$MODEL` ordering is the author's and the translator cannot honour `DEFDOSE`.

## Not the same as the neighbours

- **#899** — dose `CMT` present but out of range. Here `CMT` is absent entirely.
- **#910** — a dose CMT valid on a transit/IG model is out of range on its ODE twin. Different numbering mismatch.
- **#262** — missing `EVID` column. Different column, same silent-acceptance shape.

## Options

[...]

PR #1404, opened 2026-09-16
fix(io): read a float-formatted CMT cell as its integer and report every compartment ferx chose [closes #1009]

TeunP:
## Why

A dataset with no `CMT` column doses **compartment 1**, and says nothing about it. NONMEM does not have this gap: `$MODEL` lets a model declare `DEFDOSE` on any compartment, and NM-TRAN resolves an undecorated dose against that declaration. So a translated model whose `DEFDOSE` is not the first declared state gets its drug in the wrong place, with a finite OFV and no diagnostic.

Measured on a 1-cpt oral model written `states = [central, depot]` (the dosed state declared second), FOCEI at the initial estimates, 10 warfarin subjects / 110 observations:

| dataset spelling | ferx FOCEI OFV | NONMEM 7.6 OBJV |
|---|---|---|
| `CMT = 2` on dose rows | **−182.666708** | **−182.66670816329514** |
| `CMT` cells written `2.0` | **154050.610450** ← *reader bug* | −182.66670816329514 |
| no `CMT` column | **154050.610450** | −182.66670816329514 |

The middle row is not the same problem as the other two. `2.0` is a perfectly well-specified dataset — it is how pandas and R float-format a whole integer column once any cell in it is blank, which is the same export behaviour #830 already fixed for `L2`. All three `CMT` sites used a bare `parse::<usize>()`, so `2.0` failed the parse and fell through the *same* `.unwrap_or(1)`. That one is a bug and is fixed here. The absent column is a genuine ambiguity and is now reported instead.

Closed form outside both engines (Python, θ at init, η = 0): the Bateman superposition gives 5.246790 / 7.672992 / 9.233132 at t = 0.5 / 1 / 2, and a bolus into the concentration-unit `central` state gives 100·e^(−0.02 t) = 99.004983 / 98.019867 / 96.078944. Both engines match it to all printed digits, so the "wrong" number is exactly *the dose in compartment 1*.

## What changed

**One resolver for `CMT`, and a report of what it had to guess.**

- `parse_cmt_cell` accepts an integer literal or a float-formatted integer, and otherwise returns `Missing` (`.`/blank/`NA`/`NaN`) or `Unparseable`. A negative or fractional value is **not** rounded into a compartment — it defaults like a missing cell and is reported.
- `resolve_row_cmt` is the single resolver used by all three sites: the dose row, the observation row, and the `[data]` selection filter's `RowContext.cmt`. The filter previously used `parse_usize`, which maps both `.` and `2.0` to **0** — an invalid compartment — so `ignore = CMT == 1` failed to drop a dotted row the dose arm assigns to compartment 1. That divergence is now unrepresentable.
- Every dose or scored observation row whose compartment the reader chose is counted into one `W_CMT_DEFAULTED` summary on `population.warnings` — the channel `W_MISSING_DV` already uses, so **no `Population` field is added** (≈60 literal constructors in tests plus 4 in `ferx-tools` would otherwise break) and **no new `pub` item**, so `api/ferx-core-public-api.txt` is untouched.
- `reader_warning_suppressed` gains a second arm so `fit()` and `ferx check` cannot disagree about who sees it: suppressed unless `ode_spec.n_states > 1`.

**Scope of the warning, and why it is that scope.** An analytical `pk` model resolves compartment 1 to its own default channel (`pk::dose_needs_event_walk`: depot on oral, central on IV), and ferx's analytic numbering is ferx's — never `$MODEL`'s — so a CMT-less dataset doses exactly what a fixed-`DEFDOSE` ADVAN doses. Measured: `examples/warfarin.ferx` is bit-identical with and without the column (OFV −200.287841 both). A one-state `[odes]` model and a compartment-free model have no second compartment to have missed. Those three are silent; only a multi-state `[odes]` model is warned.

The `n_states > 1` test deliberately **counts the injected joint-PK-TTE `__chz_*` accumulators** along with the PK states. They are not dose targets, but they exist only alongside an `[event_model]`, which routes its rows *by CMT* — so a CMT-less dataset keys every row to compartment 1 and starves that endpoint. Counting them cannot produce a false positive for that reason; excluding them would produce a false negative.

**Untouched controls, named:** `check_dose_compartments` (range + routing rules, `CMT=0` bolus acceptance), `DoseEvent::cmt_idx` / `cmt_1based` (the `CMT=0` → 1 convention, #912), `[simulation] dose_cmt`, `parse_l2_id`, and both engines' dose application. This PR changes no dose routing except `2.0` → 2.

## Alternatives considered

- **A model-side `default_dose_cmt` (option 2 in the issue)** — the thing that would let a translated model *declare* its `DEFDOSE` and remove the ambiguity rather than announce it. Out of scope here on purpose: this PR makes the gap loud, not expressible. Worth its own issue.
- **Counting a literal `CMT=0`.** `0` is NONMEM's own documented default-compartment spelling and `check_dose_compartments` accepts it for a bolus, so counting it would warn on `nonmem_anchor/dose_cmt_ss_cmt0.csv`, a deliberate fixture. The residual (NONMEM sends `0` to `DEFDOSE`, ferx to compartment 1) is real and is named in the remedy text without being counted. If that is judged too quiet it is a one-line change and `explicit_integer_cmt_column_including_zero_warns_nothing` flips.
- **Rounding a fractional `CMT`.** Rejected: `2.5` becoming compartment 2 is a silent guess of exactly the kind this PR exists to remove.

## Cross-repo dependency

ferx-r keys its warning guidance on the **category**, not the code (`R/ferx_get_warnings.R:475`, `data_quality`), and `classify_warning` lands `W_CMT_DEFAULTED` there. No R change.

## Breaking changes
- [x] None

One behaviour change beyond the parse fix: a `[data]` `select`/`ignore` clause on `CMT` now sees the resolved compartment rather than `parse_usize`'s `0`, so a clause on `CMT == 0` no longer matches a dotted cell. **0 tracked datasets** have one (measured below).

## Numerical validation

- [x] Compared against: **NONMEM 7.6.0** (`nmfe76`), three runs committed under `nonmem_anchor/`

`ADVAN13 TOL=9`, `$MODEL COMP=(CENTRAL, DEFOBS) COMP=(DEPOT, DEFDOSE)`, `$ESTIMATION METHOD=1 INTERACTION MAXEVAL=0`, `$SIGMA 0.0004` (= ferx `sigma PROP_ERR ~ 0.02 (sd)`). ferx twin at `maxiter = 0`, `ode_reltol = 1e-9`, `ode_abstol = 1e-11`.

```
# NONMEM (all three datasets — DEFDOSE resolves them identically)
OBJV = -182.66670816329514

# ferx, after this PR
defdose_cmt2.csv        -182.666708122452   |Δ| = 4.084e-8   (relative 2.2e-10)
defdose_cmt2float.csv   -182.666708122452   |Δ| = 4.084e-8   — to_bits()-identical to the line above
defdose_no_cmt.csv       154050.610450      + W_CMT_DEFAULTED (10 dose rows, 110 observation rows)

# ferx, before this PR
defdose_cmt2float.csv    154050.610450      (the bug)
```

**The tolerance is measured, not argued.** Realised worst |Δ| is **4.084e-8** on an objective of magnitude 183 — a relative error of 2.2e-10, which is the size two independent solvers at `reltol 1e-9` and `TOL=9` should differ by. The test's bound is `5e-7`: that number with one order of magnitude of headroom, and still 3e11 times tighter than the defect it guards (a compartment swap moves the objective by 154233). The no-`CMT` arm's "routing did not move" check is bounded at `1e-5` against a realised 1.644e-7, where ≤5e-7 of the error is the reference constant's own 6-decimal print rather than the engine.

**In-tree exposure, re-measured on this branch:** 220 tracked CSVs (217 before this PR's three fixtures). **9** pre-existing files have no `CMT` column — 4 feed analytical `pk` fits, 3 are compartment-free, 1 is a NONMEM-side CTMM file, 1 an EBE table — so all 9 land in the suppressed classes. **0** carry a missing `CMT` cell and **0** carried a float-formatted one before this PR. The full suite is green, which is the check that the new warning fires on no existing fixture.

## Performance

The added work is one `enum` return per `CMT` cell and two `usize` increments per defaulted row; the summary string is built once per population, only when something was defaulted.

## Tests

| tier | test | what it pins |
|---|---|---|
| 1 | `float_formatted_cmt_cell_reads_as_its_integer` | `2.0` → compartment 2 (dose) and `3.0` → 3 (observation), and no warning — a readable cell is not a chosen one. Neither value is the compartment-1 fallback and the two differ, so neither a broken parse nor a site swap can pass; see the mutation note below |
| 1 | `missing_and_unparseable_cmt_cells_default_to_1_and_are_counted` | `.`/blank/`NA` vs `-1`/`2.5`/`x` split, the quoted examples, the 6-dose/0-obs row split, and that `2.5` is **not** rounded to 2 |
| 1 | `absent_cmt_column_warns_once_with_dose_and_obs_counts` | exactly one summary; the absent-column wording (the row counters alone cannot tell an absent column from a column of missing cells); counts summed across subjects; a decoy `COMPT` header is not read |
| 1 | `explicit_integer_cmt_column_including_zero_warns_nothing` | control — a literal `CMT=0` is not "defaulted" |
| 1 | `filter_context_cmt_resolves_like_the_dose_site` | the `[data]` filter sees the same compartment the dose arm assigns |
| 1 | `emitted_cmt_defaulted_message_classifies_as_data_quality` | `classify_warning` on the string the reader **actually builds** (both spellings), not a paraphrase |
| 1 | `cmt_defaulted_is_{reported_on_a_multi_state_ode,suppressed_on_an_analytical_pk,suppressed_on_a_single_state_ode,suppressed_on_a_compartment_free}_model` | four model classes, **each its own test**, so an inverted predicate reddens the side it broke |
| 1 | `the_cmt_arm_does_not_swallow_the_no_doses_arm` | the new early return does not change how #811's arm is decided |
| 2 | `cmt_less_dataset_on_a_multi_state_ode_model_is_reported` | both surfaces — `ferx check` (code `W_CMT_DEFAULTED`, not the `W_DATA` fallback) **and** `fit()` |
| 2 | `cmt_less_dataset_on_an_analytical_pk_model_is_not_reported` | control, both surfaces |
| 2 | `tests/defdose_no_cmt_nonmem_anchor.rs` (3) | the NONMEM comparison, `to_bits()` float↔integer identity, and that routing did not move |

The anchor is **ungated**: `MAXEVAL = 0` on 10 subjects is ~2 s, so it runs on every PR.

**Suite status at PR open.** `tools/preflight.sh` is green (fmt, the 5 `cargo check`
feature sets, clippy `--all-targets`, rustdoc, docs-lint, public-API baseline) and every
test named above was run and passed individually. The three full CI-job commands
(`ferx-core/ci --lib`, `ferx-core/ci --tests --no-fail-fast`, `ci,markov,nn --tests
--no-fail-fast`) were still running locally when this PR was opened — CI runs the same
three, so treat their result here as the authority. I will post the local counts as a
comment when they land.

### Mutation testing

| mutation | site | tests killed |
|---|---|---|
| **M1a** revert the **dose** site to the strict parse | `datareader.rs` dose row | 8 — 5 reader, `check_command` multi-state, both anchor arms |
| **M1b** revert the **observation** site | `datareader.rs` obs row | 2 — `float_formatted_cmt_cell_reads_as_its_integer`, `absent_cmt_column_warns_once…` |
| **M1c** revert the **`[data]` filter** site to `parse_usize` | `RowContext.cmt` | 1 — `filter_context_cmt_resolves_like_the_dose_site` |
| **M2** delete the float-formatted-integer branch | `parse_cmt_cell` | 3 — reader, filter, anchor `to_bits()` identity |
| **M3** round a fractional value instead of rejecting it | `parse_cmt_cell` | 1 — `missing_and_unparseable…` (`2.5` must not become 2) |
| **M4** collapse `Unparseable` into `Missing` | `resolve_row_cmt` | 1 — the cause-split assertion |
| **M5** gate the summary on `cmt_col.is_some()` | summary block | 4 — loses the absent-column case, the issue's own scenario |
| **M6** reword the absent-column clause | summary block | 1 — the wording assertion |
| **M7** count a literal `CMT=0` as defaulted | `CmtDefaults::record` | 1 — `explicit_integer_cmt_column_including_zero_warns_nothing` |
| **M8** drop the deferred observation `record()` | obs arm | 3 — the obs half of every row-split assertion |
| **M9** invert the suppression predicate | `reader_warning_suppressed` | 7 — **all four** model-class tests + both `ferx check` surfaces + anchor |
| **M10** delete the whole `W_CMT_DEFAULTED` arm | `reader_warning_suppressed` | 4 — only the three *suppressed* classes + the analytical check |
| **M11** delete the check-report prefix arm (falls to `W_DATA`) | `validation.rs` code map | 1 — the code assertion |
| **M12** delete the `classify_warning` arm | `types.rs` | 2 — the emitted-string test **and** the table-driven round-trip |

**14 mutations, 14 killed, 0 survived.** Each ran with `--no-fail-fast`; the four
source files were byte-restored (`cp` + `touch`) between mutations, and the
pristine snapshot was verified byte-identical to `HEAD` before the run.

Two things this bought beyond a kill count:

- **M9 and M10 kill different sets.** Inverting the predicate reddens all four
  model classes; deleting it reddens only the three suppressed ones, because with
  no arm at all the multi-state case still (correctly) reports. That is why the
  four classes are four tests rather than one parameterised test — a failure names
  the side that broke.
- **M1b found a hole in my own test.** `float_formatted_cmt_cell_reads_as_its_integer`
  originally wrote `1.0` on its observation row and asserted `obs_cmts == [1]` —
  which the *broken* reader also produces, since the strict parse fails and
  `.unwrap_or(1)` lands on the same answer. The observation half of the fix was
  untested by the test written for it. Cells are now `2.0` / `3.0` (neither is the
  fallback, and they differ from each other, so a site swap also cannot pass);
  M1b re-run then killed 2 tests instead of 1. Fixed in `4098aa47`.

## Example

An `[odes]` model whose dosed state is not the first one, on a dataset that forgot to say so:

```toml
[structural_model]
  ode(obs_cmt=central, states=[central, depot])
```

```
ID,TIME,DV,EVID,AMT,MDV
1,0,.,1,100,1
1,1,5.0,0,.,0
```

```
W_CMT_DEFAULTED: the dataset has no CMT column, so 1 dose row(s) and 1 observation
row(s) were assigned compartment 1. On a model that addresses more than one state
this is a guess rather than a default — a translated model whose NONMEM DEFDOSE is
not the first state gets its drug in the wrong place, with no error. Give the
dataset a 1-based CMT column ordered like the model's `states = [...]`, or map an
existing header onto it with `CMT = <header>` in the [data] block. Writing CMT=1
explicitly on every row silences this.
```

## Docs
- [x] `docs/` updated — `docs/data-format.qmd` (the `CMT` row of the optional-columns table, and a new bullet under `### CMT on [odes] models` carrying the measured OFV and the three remedies) and `docs/file-formats/check-report.qmd` (the code table under `#### Dose records`). `cargo run -p docs-lint -- --check` green; no new `docs/.lint-baseline` entry.
- [x] `CHANGELOG.md` `[Unreleased]` — one **Fixed** (float `CMT`) and one **Added** (`W_CMT_DEFAULTED`)

No new page, so `docs/_quarto.yml` is unchanged. No new `.ferx` example file and no new `data/` file, so no ferx-site / ferx-book mirroring is needed; the `nonmem_anchor/` fixtures are test data.

## Checklist
- [x] Nothing on the `Dual2` sensitivity path is touched — this is the value path (the CSV reader) only
- [x] No `Cargo.toml` version bump (not breaking)

## Reviewer hints

**Which engine each fixture ran on.** Every probe and all three anchor tests are `[odes]` models on Dormand–Prince at `reltol 1e-9 / abstol 1e-11`, with `analytic (Dual2)` inner gradients (printed by the fit) — no subject falls back to finite differences. The analytical control (`cmt_less_dataset_on_an_analytical_pk_model_is_not_reported`, and the `pk` arm of the suppression tests) is the closed form, not the ODE engine; that is the point of it, since ferx's analytic compartment numbering is what makes compartment 1 the right answer there.

**Three things worth a second look:**

1. **The count's placement in the observation arm.** It is deferred until `skip_missing_dv` is known, and gated `!skip_missing_dv || routing.tte.contains(&cmt)` — a missing-DV row skipped under #258 routes nowhere and is already reported by `W_MISSING_DV`, so counting it would report one row as two findings; the TTE arm is the exception because it keeps such a row under its own DV-code semantics.
2. **The absent-column wording reads `cmt_col` directly** rather than deriving from the counters. It has to: with no column there is no cell to classify, so `n_missing_cell` and `n_unparseable_cell` are both 0 and the per-row counters cannot distinguish "no column" from "a column of `.`". `absent_cmt_column_warns_once_with_dose_and_obs_counts` is the test that pins this, and mutation M5/M6 are the two ways to break it.
3. **`usize::MAX as f64` in the range check** rounds up to 2^64, so a cell of exactly `18446744073709551616` saturates to `usize::MAX` rather than being rejected. Deterministic, matches what the integer path (`parse::<usize>()`) accepts up to, and identical in shape to `parse_l2_id`'s `i64` bound — but it is the one corner I did not write a test for, because `check_dose_compartments` rejects any such compartment immediately afterwards.

**Skimmable:** the docs and CHANGELOG diffs, and `nonmem_anchor/results/` (verbatim NONMEM output).

## Open questions

Whether a literal `CMT=0` should be counted as defaulted. It is currently not — see *Alternatives considered*. NONMEM sends a `0` to `DEFDOSE` and ferx to compartment 1, so on a multi-state `[odes]` model with a non-first `DEFDOSE` the residual is the same wrong-compartment dose the absent column gives. The message names it in the remedy text rather than counting it; flipping that is a one-line change plus one test.


TeunP:
## Review — round 1 at `4098aa47`

Read as a stranger, in a fresh worktree of the PR head. The anchor is sound and the
mutation table is the good kind — **M1b** (the `1.0` observation cell the *broken* reader
also produces) is exactly the class of hole the repo's testing rules exist for, and
finding it in your own test is worth more than the 14/14 kill count next to it. Three
NONMEM runs all returning `-182.66670816329514`, and a `to_bits()` identity between `2`
and `2.0`, are the right assertions for this object.

**Suite, run here at `4098aa47` (base `233246d1`, branch contains main):**

```
cargo test --workspace --no-default-features --features ferx-core/ci --lib
  ok. 4739 passed; 0 failed; 26 ignored  (675.63s)   + 602 + 61 in the member crates
cargo test --workspace --no-default-features --features ferx-core/ci --tests --no-fail-fast   exit 0
cargo test            --no-default-features --features ci,markov,nn --tests --no-fail-fast    exit 0
cargo run -p docs-lint -- --check                                                             exit 0
```

**Nothing blocking.** Four things worth acting on before merge, three of them measured
below; the rest are nits.

### Findings

| # | Finding | Severity | Evidence | Fix |
|---|---|---|---|---|
| 1 | The suppression rationale covers the **dose** channel; the warning also counts **observation** rows, and an analytical model *can* dispatch observations by CMT — so the one case the doc comment says is safe is not | should-fix | **Measured.** `pk one_cpt_iv` + `[scaling] obs_scale[CMT=1] = 1000` / `obs_scale[CMT=2] = 1` (parses today — `model_parser_tests.rs:15690`, `:15768`), 2 subjects, same data spelled twice: explicit `CMT=2` → `obs_cmts=[2,2]`, **OFV 121.010256**; no `CMT` column → `obs_cmts=[1,1]`, **OFV 743844863.414805** — and `FitResult.warnings` carries **no `W_CMT_DEFAULTED` in either arm**. `ScalingSpec::PerCmt` dispatches by `obs_cmts` (`types.rs:3292`) and `pk::validate_per_cmt_scaling` (`pk/mod.rs:646`) only checks that the **observed** CMTs have entries, so `{1} ⊆ {1,2}` passes silently | Either widen the predicate at [`validation.rs:723`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/api/validation.rs#L723) (also report when `model.scaling` is `PerCmt`, or the readout is per-CMT), or narrow the doc comment. As written it asserts "an analytical `pk` model … has no second compartment to have missed", and that is false for this model |
| 2 | A **ragged row** on a dataset that *has* a `CMT` column emits a warning with an empty cause clause | should-fix | **Measured.** Header `ID,TIME,DV,EVID,AMT,CMT,MDV`, dose row `1,0,.,1,100` (5 fields — the reader is `ReaderBuilder::new().flexible(true)` at `:659` and never pads): <br>`W_CMT_DEFAULTED: , so 1 dose row(s) and 0 observation row(s) were assigned compartment 1. …` <br>[`:1339`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1339) returns `NoColumn` whenever `row.get(c)` is `None`, but [`:1084`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1084) picks its clause from `cmt_col.is_none()`, so both cause counters stay 0 and `parts.join(" and ")` is `""` | Split the cause — `cmt_col.is_none()` → `NoColumn`, column present but cell absent → `MissingCell`. One arm, one fixture with a short row |
| 3 | The parse fix can turn a **running** fit into a hard error; the PR ticks *Breaking changes: None* | should-fix | `check_dose_compartments` raises the fatal `E_DOSE_CMT_OUT_OF_RANGE` on `cmt > n_states` ([`validation.rs:1592`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/api/validation.rs#L1592)). A dose cell written `4.0` on a 2-state `[odes]` model used to collapse to compartment 1 and run; it now fails outright. Same shape on the obs side against per-CMT `CMT=N:` coverage | Good behaviour, missing paperwork — one clause in the CHANGELOG `Fixed` entry and the breaking-changes box. Exposure really is 0 (re-measured: 220 tracked CSVs, 10 without `CMT`, 0 with a float-formatted cell — your numbers check out exactly) |
| 4 | The anchor **cannot be re-derived** from what is committed, and the float arm's provenance is unverifiable | should-fix | All three `.ctl`s carry `$DATA data.csv` (`defdose_cmt2.ctl:3`), no `data.csv` is committed, and the recipe at `README.md:709` (`for f in defdose_*.ctl; do nmfe76 "$f" …; done`) never stages one — NM-TRAN would not find a dataset. Worse: `defdose_cmt2.ctl` and `defdose_cmt2float.ctl` have identical `$INPUT` and identical column counts, and `diff results/defdose_cmt2.lst results/defdose_cmt2float.lst` is **`$PROBLEM` and CPU timings only** — so had `data.csv` been the integer file for both runs, the committed evidence would look the same. That is the arm the fix is about | Point each `$DATA` at its own committed CSV — already the convention in that file (`$DATA ../data/warfarin.csv` ×8, `$DATA lag_arrival_read.csv`, …) — or stage explicitly: `for f in defdose_no_cmt defdose_cmt2 defdose_cmt2float; do cp "$f.csv" data.csv; nmfe76 "$f.ctl" "$f.lst"; done`. The line just above this block says the anchors should be "re-derivable rather than taken on trust" |
| 5 | `-0.0` and `-0` read as `CMT 0` and are **never reported**, while `-1` in the same column is | nit | **Measured.** Three dose rows `-0.0` / `-0` / `-1` → all compartment 1, and the summary says only `1 row(s) had a CMT cell that is not a compartment index ("-1")`. [`:1315`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1315): `(0.0..=usize::MAX as f64).contains(&-0.0)` is true (IEEE `-0.0 == 0.0`), so `parse_cmt_cell` returns `Value(0)` — the explicit spelling `explicit_integer_cmt_column_including_zero_warns_nothing` deliberately does not count. The doc comment two lines up says "A negative … value is `Unparseable`" | `f.is_sign_negative()` rejects it; replacing the range check with `f as u64 as f64 == f` also closes the `usize::MAX as f64` → 2^64 corner you named in the PR body |
| 6 | `filter_context_cmt_resolves_like_the_dose_site`'s second assertion is **vacuous** | nit | [`datareader_tests.rs:2746`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader_tests.rs#L2746): the dotted dose row is excluded by the filter *before* the dose arm records it and the other two cells parse, so no `W_CMT_DEFAULTED` is emitted at all — `!…contains("no CMT column")` passes by absence and survives deleting the whole summary block. Only `cmts == vec![2]` above it is load-bearing (that is the one M1c kills) | Drop it, or keep one defaulted row in the fixture so the present-column wording is actually observed |
| 7 | A TTE row is **counted and then skipped** | nit | Read, not run. The count sits at [`:2146`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L2146), before the endpoint branch; the TTE arm `continue`s at `:2172` when `raw_entry > time + 1e-12` ("row skipped"). Contradicts the comment two lines above ("rows that actually become an observation") and the summary block's "match the rows the fit actually uses" | Move the `record()` past the TENTRY guard, or name the exception at the call site |
| 8 | No test that the **remedy the message prints** works | nit | It *does* work — I checked: a `[data]` block with `CMT = COMPT` on the two-state ODE probe gives `validate_model_file` **zero diagnostics** (warning silenced, dose routed). But nothing in the suite pins it. `absent_cmt_column_warns_once_with_dose_and_obs_counts` pins only the decoy (`COMPT` *not* read as CMT), and this is the one remedy of the three that leans on a separate subsystem (#730/#742 column mapping) | One case asserting the mapped column silences the warning *and* lands the dose in compartment 2 |
| 9 | `parse_cmt_cell` is a second copy of `parse_l2_id`'s float-formatted-integer parse | nit | [`:1306`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1306) vs [`:1263`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1263) — same #830 shape, two integer types, two missing-cell policies. This PR exists because that divergence went unnoticed between `L2` and `CMT`; CLAUDE.md's rule is "when the bug report is 'two implementations disagree', the fix is one implementation" | A shared `parse_float_formatted_integer(&str) -> Option<f64>` both call. The next export quirk then lands once |
| 10 | Example cell text is interpolated **unbounded and unescaped** | nit | [`:1386`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1386) stores the trimmed cell verbatim; the summary writes `"{e}"`. A long free-text cell lands in full in `FitResult.warnings`, the fit YAML and the check report; a cell containing a double quote produces `"a"b"` | Truncate to ~24 chars with an ellipsis |
| 11 | `pub(crate) struct CmtDefaults` is dead visibility | nit | [`:1357`](https://github.com/FeRx-NLME/ferx-core/blob/4098aa47/src/io/datareader.rs#L1357) — constructed, `absorb`ed and consumed entirely inside `datareader.rs`, and its only producer `parse_subject` is private | Drop `pub(crate)`. Its own doc comment argues the reader must stay model-blind; a narrower visibility says so structurally |

### What I checked and it held

- **One resolver, three sites.** `grep -n cmt_col src/io/datareader.rs` gives exactly the filter (`:1873`), the dose arm (`:2009`) and the obs arm (`:2124`) plus the resolver. No fourth site anywhere — `output.rs:3650` is a test.
- **In-tree exposure, re-measured on this branch.** 220 tracked CSVs; 10 with no `CMT` header — your 9 plus `defdose_no_cmt.csv`. `tests/nonmem/ctmm_2state.csv` really is NONMEM-side (ferx reads `data/ctmm_2state.csv`, `tests/markov_nonmem.rs:69`). No in-tree `[data]` `select`/`ignore` clause reads `CMT`.
- **No phrase collision.** Nothing in `src tests crates` filters `FitResult.warnings` on a phrase the new wording could land in.
- **ferx-r really needs nothing.** `git -C ../ferx-r show origin/main:R/ferx_get_warnings.R` line 475 carries `data_quality`, and `classify_warning` lands there — asserted against the emitted string rather than a paraphrase, which is the right test.
- **Both surfaces read through the same routed chokepoint** (`validate_model_file` → `read_population_for`, `validation.rs:6386`), so `fit()` and `ferx check` cannot diverge on the counts. That agreement is real, not merely claimed.
- **`docs/warnings.qmd` genuinely needs no row** — it documents the `WarningCode` *category* vocabulary, not `W_*` codes.
- **Withdrawn candidate.** I chased the suppression as a false negative for endpoint models with no `[odes]` — `examples/tte_weibull.ferx` (`[event_model] cmt = 2`), `binary_logistic.ferx` (`cmt = 3`). A CMT-less dataset there keys every row to 1, the endpoint gets no records, and `E_ENDPOINT_NO_RECORDS` fires — whose own registry entry reads "typically a missing `CMT` column". Loud either way, so the suppression is right there. Finding 1 is the case that survives, because an analytical model's *observation* dispatch has no such backstop.

### What I did not check

- The NONMEM side itself — I read the committed `.ext` / `.lst` / `.ctl` and confirmed all three `OBJV` rows, but did not rerun `nmfe76`. Finding 4 is about exactly that gap.
- Codecov patch coverage on the diff.
- `slow-tests`: nothing here is slow-gated, so `/ferx-land`'s dispatch is not needed for this one.

TeunP:
## Round 1 resolved — `2e6bb7d8`

Thanks; finding 1 was a real scope error, not a polish item, and findings 4 and 6 both landed on things I had asserted were sound. All 11 of your findings are addressed, plus 5 from a parallel self-review and one behaviour question from @TeunP. **14 fixed, 1 withdrawn, 1 of your suggested fixes replaced** (see the note under #5).

| # | Finding | Fix |
|---|---|---|
| 1 | Suppression covered the dose channel only; an analytical model dispatches observations by CMT | **Fixed.** `cmt_defaulting_is_ambiguous` now returns true for a multi-state `[odes]` model **or** any observation-side dispatcher: `ScalingSpec::PerCmt`, `ErrorSpec::PerCmt`, or a per-CMT `AnalyticReadout`. Reproduced your measurement independently on my own fixture before changing anything — `pk one_cpt_iv`, `obs_scale[CMT=1] = 1000` / `obs_scale[CMT=2] = 1`: explicit `CMT=2` → OFV **0.0357**, column dropped → **6097015712.1246**, no warning in either arm. Two new tests, one per dispatcher, because mutations N3 and N4 must die separately |
| 2 | Ragged row → empty cause clause | **Fixed.** `NoColumn` is now reserved for `cmt_col.is_none()`; a short row on a dataset that declares CMT is a `MissingCell`. Test asserts the malformed shape directly (`!w.contains("W_CMT_DEFAULTED: ,")`), not just the counts |
| 3 | `4.0` can turn a running fit into a hard error; "Breaking changes: None" | **Fixed (paperwork).** CHANGELOG sub-bullet names `E_DOSE_CMT_OUT_OF_RANGE`, says it is intended, and records the measured exposure (0 of 220 tracked CSVs). PR body updated |
| 4 | Anchor not re-derivable; float arm's provenance unverifiable | **Fixed.** You were right and it was the worst item here — I had written "re-derivable rather than taken on trust" into the README while all three streams said `$DATA data.csv` with no such file committed. They also shared one `$TABLE FILE=run.tab`, so the committed tables had the same problem. Each stream now names its own dataset **and** its own table, and all three were re-executed on NONMEM 7.6.0 from the committed files: `-182.66670816329514` each. Each `.lst` now records which dataset produced it. README documents the hole and the re-derivation |
| 5 | `-0` / `-0.0` read as CMT 0 and never reported | **Fixed, but not with the suggested one-liner.** `!f.is_sign_negative() && f < usize::MAX as f64`. `f as u64 as f64 == f` does **not** close the 2^64 corner: `2^64 as u64` saturates to `u64::MAX`, which rounds back to 2^64 and compares equal — verified, not reasoned. The strict `<` closes it; both traps are now recorded in the doc comment. New test straddles `-0` / `-0.0` / `-1` with the literal `0` control |
| 6 | `filter_context_cmt_…`'s second assertion is vacuous | **Fixed.** Correct, and embarrassing in a test written *about* tests that cannot fail — my own comment in it admitted the row "never became a dose". The clause is narrowed to `EVID == 1 && CMT == 1` so a defaulted observation survives, the warning actually fires, and both assertions are load-bearing |
| 7 | TTE row counted then skipped | **Fixed properly rather than by comment.** The count moved after all four endpoint arms; reaching it now *is* "the row became an observation". Mutation N7 (move it back) kills five tests |
| 8 | No test that the printed remedy works | **Fixed.** Two halves, because silencing without routing would be worse than not silencing: `CMT = COMPT` in `[data]` produces zero diagnostics **and** `prepare_run` shows the dose in compartment 2 |
| 9 | `parse_cmt_cell` duplicates `parse_l2_id` | **Fixed.** Shared `parse_float_formatted_integer`; each caller keeps its own sign/range policy, which is the part that genuinely differs |
| 10 | Example text unbounded and unescaped | **Fixed.** Truncated to 24 chars, quotes and control characters replaced, char-boundary safe. Also added the ellipsis for a capped *list* — a count of five beside three spellings read as exhaustive |
| 11 | `CmtDefaults` dead visibility | **Fixed.** Now private |

### Also fixed, from a parallel self-review

| Finding | Fix |
|---|---|
| `record` / `absorb` duplicated the bounded de-dup logic | One `push_example`; both route through it |
| `api/fit.rs` comment listed only the two old reader codes | Now points at the `diagnostics.rs` registry as the authority rather than trying to stay exhaustive |
| Per-cell `String` allocation after the example cap | Folded into the truncation work |

### Withdrawn

**TTE rows labelled "observation row(s)"** — my own finding, and wrong. A TTE record *is* an `EVID=0` observation row; the label names the row type, which is also NONMEM's usage. Changing it would churn four assertions for no gain.

### Behaviour question answered with a test

@TeunP asked whether a CMT-less dataset still works with `ADDL` and `EVID=3`/`4`. It does, and nothing here changed that — but measuring it surfaced a reporting gap. On a fixture with `ADDL=3` plus an `EVID=4` dose, **5 doses** land in compartment 1 while only **2 dose rows** exist, so the row count alone understated the exposure by 2.5×. The message now reads `2 dose row(s) (5 doses after ADDL expansion) and 2 observation row(s)`, with the parenthetical suppressed when nothing expands (straddle control test). `EVID=3` is correctly counted in neither.

### Round-2 mutation testing

7 mutations, **7 killed, 0 survived** — `--no-fail-fast`, byte-copy restore with `touch`, pristine snapshot verified against `HEAD` first.

| mutation | tests killed |
|---|---|
| N1 short row → `NoColumn` again | 1 — `a_row_shorter_than_its_header_…` |
| N2 drop the sign guard | 2 — `negative_zero_is_reported_…`, `missing_and_unparseable_…` |
| N3 drop the per-CMT **scaling** arm | 1 — the scaling test only |
| N4 drop the per-CMT **error-model** arm | 1 — the error-model test only |
| N5 report row count as event count | 1 — `cmt_less_dataset_routes_addl_…` |
| N6 drop the truncation ellipsis | 1 — `a_capped_example_list_says_it_was_capped` |
| N7 move the obs count back before the branch | 5, across the reader and `ferx check` |

N3 and N4 killing *different* tests is the property worth having: a predicate that covers one dispatcher and not the other reddens the arm it actually missed.

### Gates at `2e6bb7d8`

```
tools/preflight.sh                                          OK (public-API baseline unchanged)
ferx-core/ci --workspace --lib          5410 passed, 0 failed, 26 ignored
ferx-core/ci --workspace --tests        6493 passed, 0 failed, 209 integration binaries
ci,markov,nn --tests                    6123 passed, 0 failed
cargo run -p docs-lint -- --check       OK, no new baseline entry
NONMEM 7.6.0, three streams re-derived  -182.66670816329514 each
```

Counts reconciled against numbers written down before the runs, and both of my predictions were wrong in bookkeeping: I said `--lib` +11 (actual **+8** — I had double-counted three round-1 tests already in the 5402 baseline) and `--tests` +1 (actual **+9** — `cargo test --tests` also runs the 8 lib unit-test targets). All 8 new lib tests and the 1 new integration test verified by name, not by `failed=0`.

### Not addressed

- **Codecov patch coverage** — not checked locally; CI is the authority.
- **`slow-tests`** — nothing here is slow-gated, so no dispatch is needed, as you noted.
- Your withdrawn candidate (endpoint models with no `[odes]`) I agree is correctly withdrawn, and the reasoning is now recorded in `cmt_defaulting_is_ambiguous`'s doc comment so the next reader does not re-derive it: declared endpoints are deliberately absent from the predicate because `E_ENDPOINT_NO_RECORDS` already reports that case as an error naming the missing column.

TeunP:
## Round 2 resolved — `c32bd0c8`

All ten findings fixed, plus a scope widening @TeunP called for that turned out to be the same defect on a third engine. First, a correction I owe on my own round-1 reply.

### Correction: the N7 row in my round-1 mutation table named an edit I did not run

That table said *"N7 move the obs count back before the branch — 5 tests killed"*. The harness did not do that. It **deleted** the `record_obs` call:

```python
("N7 obs count: move it back before the endpoint branch",
 "            if let Some(cause) = &cmt_default_cause {\n"
 "                cmt_defaults.record_obs(cause);\n"
 "            }\n"
 "        } else if evid == 2 && any_tv {",
 "        } else if evid == 2 && any_tv {",      # ← removed, not moved
 "lib check"),
```

Deleting the count kills five tests for a trivial reason — the counts go to zero. It says nothing about whether *moving* it is observable, which is the property the fix claims. Your M2 is the edit I should have run, and it survived for exactly the reason you gave: the old site already carried `!skip_missing_dv || routing.tte.contains(&cmt)`, so the only behavioural delta is a TTE row the `TENTRY > TIME` guard drops, and no fixture had a defaulted CMT on one. A mutation table that names an edit it did not run is worse than no table. It is now P6 below, run properly, against a fixture that straddles it.

### Scope: the warning now fires on analytical multi-compartment models too

Finding 1 was the ODE per-CMT readout. Chasing it, the same question was open on the **analytical dose channel**, which the predicate suppressed wholesale on the argument that ferx's analytic numbering equals NONMEM's ADVAN numbering. True, and beside the point — `pk/topology.rs`:

```rust
static ONE_CPT_ORAL: PkTopology = { channels: &[Some(Depot), Some(Central)], … };
static TWO_CPT_IV:   PkTopology = { channels: &[Some(Central), Some(Periph1)], … };
```

`CMT=2` on an oral model is the depot-bypassing central bolus (#350); on a 2-cpt IV model it is the peripheral. This repo already anchors what happens when one is replaced by compartment 1 — `tests/nonmem_dose_compartment_anchor.rs` (#375), NONMEM 7.6.0, zero relative deviation:

> `ADVAN2` (1-cpt oral), bolus into `CMT=2` … Superposition computes **1.19324** (it treats it as a depot dose); NONMEM and the walk agree on **1.8097**.

A CMT-less dataset makes exactly that substitution, silently. So no new NONMEM run was needed — the premise was already measured; what was missing was the warning.

Both halves of the predicate are now **read off each engine's own routing table** rather than restated from it, because a predicate written as a list of remembered model classes is what failed in both review rounds:

```rust
fn addressable_dose_compartments(model: &CompiledModel) -> usize {
    if let Some(spec) = model.ode_spec.as_ref() { return spec.n_states; }
    if model.is_algebraic() { return 0; }
    model.pk_model.topology().addressable_dose_compartments()  // live `channels` entries
}
```

Counting live channels rather than `n_states` is load-bearing: transit and inverse-Gaussian are `n_states` 2 or 3 with `channels: &[]`, since their closed form absorbs every dose through the depot and never reads `dose.cmt`. P2 mutates exactly that.

**Blast radius, measured before committing to it:** 10 of 220 tracked CSVs have no `CMT` column; a full `--lib` run under the widened predicate was 5407 passed / 3 failed, and all three were my own suppression fixtures whose controls asserted the thing that changed. No production test regressed.

### A structural anchor for the scope, not a third list

`tests/cmt_defaulting_scope.rs` pins the property the arms are *for*:

> For each `PkModel`, a dose into compartment `k` differs from the same dose into compartment 1 **iff** dropping the `CMT` column raises `W_CMT_DEFAULTED`.

Neither side is hard-coded — the left is measured by predicting both ways, the right is read out of `ferx check`, and the test asserts they agree. The `match` is exhaustive over `PkModel`, so a new analytical model is a **compile error** until someone states its dose targets (they do *not* state whether the warning should fire; that is measured). Realised, printed on every run:

| model | probes | rejected | worst rel diff | observable | warned |
|---|---|---|---|---|---|
| `OneCptIv` | 2 | 2 | 0.000e0 | false | false |
| `OneCptOral` | 2 | — | **3.406e-1** | true | true |
| `TwoCptIv` | 2 | — | 9.601e-1 | true | true |
| `TwoCptOral` | 2, 3 | — | 9.410e-1 | true | true |
| `ThreeCptIv` | 2, 3 | — | 9.909e-1 | true | true |
| `ThreeCptOral` | 2, 3, 4 | — | 9.867e-1 | true | true |
| `OneCptTransit` | 2 | 2 | 0.000e0 | false | false |
| `TwoCptTransit` | 2, 3 | 2, 3 | 0.000e0 | false | false |
| `OneCptIg` | 2 | 2 | 0.000e0 | false | false |
| `TwoCptIg` | 2, 3 | 2, 3 | 0.000e0 | false | false |

Smallest live separation 3.406e-1 against a 1e-6 threshold; every suppressed topology at exactly 0, because its second compartment is *rejected* rather than silently re-routed. That is a separation of kinds, not a tolerance. Two things about it I would rather state than let a reader assume: the transit/IG bit-identity comparison in the second test is currently a **guard, not an exercised path** (all four reject every probe today), and `OneCptIv` probes `CMT=2` deliberately so its `observable == false` is a measurement rather than the result of asking nothing.

### Findings

| # | Finding | Fix |
|---|---|---|
| 1 | Predicate misses an ODE model's per-CMT readout — `ode_spec.readout` never inspected | **Fixed.** Reproduced before touching it: one-state `[odes]`, `y[CMT=1] = central/V` / `y[CMT=2] = central/V*1000`, OFV **10028.0940** with the column against **0.0357** without, `ferx check` reporting `0 warning(s)` on both. Both readout structs now go through one `per_cmt` closure. `OdeReadout::PerCmt` is dispatched on `obs_cmts` by *two* engines — `ode::predictions` for the f64 value and `sens::ode_provider` for the `Dual2` twin — so this was wrong in the gradient as well as the objective |
| 2 | Arm 4 untested; M1 survives | **Fixed.** Test added; P4 confirms it dies alone |
| 3 | Arm 3's fixture is `PerCmt(HashMap::new())` on an analytical model — neither shape it claims | **Fixed, and the arm was wrong too.** `model_parser.rs` says it outright: *"An empty PerCmt arises for TTE-only models (no [error_model] block) — allow it regardless of is_ode. Non-empty PerCmt still requires ODE."* So the fixture could only exercise the empty case, and `matches!(.., PerCmt(_))` was reporting **every declared endpoint model**, contradicting the predicate's own doc comment. Gated on `!m.is_empty()`; fixture retargeted to a one-state `[odes]` model, which parses and is not already covered by the dose channel; the empty-map case has its own test. Confirmed end-to-end: a TTE-only model on a CMT-less dataset now emits no `W_CMT_DEFAULTED`, just `E_ENDPOINT_NO_RECORDS` |
| 4 | The count's *move* is untested | **Fixed** — see the correction above. Two tests: a TTE pair differing only in `TENTRY` (50 vs 0 against `TIME=20`), so the same two rows give 1 vs 2 observations; and a defaulted missing-DV row that must not be counted. Separate, because the two take different `continue`s |
| 5 | Ellipsis compares rows against spellings | **Fixed.** Five rows all spelled `x` printed `("x", …)`. Now an `examples_capped` flag set only when a *distinct* spelling is turned away, carried across subjects by `absorb` (P9 mutates that carry). The test varies rows at a fixed spelling count — the variable the old straddle could not see |
| 6 | Comment names a `CMT=0` control the fixture lacks | **Fixed** by adding the row, not deleting the sentence: `CMT=0` and `-0` dose the same compartment, so only the *report* separates them |
| 7 | `docs/data-format.qmd` row narrower than the predicate | **Fixed**, prose section rewritten around the two channels with the measured numbers for each |
| 8 | CHANGELOG duplicate clause; "none of 220 CSVs" goes false on merge | **Fixed.** Re-measured: 220 tracked CSVs, 10 without a `CMT` header, **one** with a non-integer `CMT` cell — this PR's own `defdose_cmt2float.csv`. Now reads "no pre-existing dataset" and names it |
| 9 | U+2028/29 survive `is_control()` into a one-line warning | **Fixed.** Asserted with `w.lines().count() == 1`, which holds however the escape is spelled |
| 10 | Doc comment omits that strict `<` also rejects float-spelled `usize::MAX` | **Recorded**, with why it is left as it is |

### Round-3 mutation testing

9 mutations, **9 killed, 0 survived** — `--no-fail-fast`, byte-copy restore with `touch`, snapshot verified byte-identical to the working tree first.

| mutation | oracles killed |
|---|---|
| P1 drop the analytical topology branch | 4 — oral, 2-cpt IV, `check` oral, scope |
| P2 answer `n_states` instead of live channels | 2 — transit straddle, scope |
| P3 drop the `ode_spec` per-CMT readout arm | 1 — one-state ODE readout |
| P4 drop the `analytic_readout` per-CMT arm | 1 — analytic readout |
| P5 drop the non-empty `PerCmt` gate | 1 — empty-map endpoint |
| P6 **move the obs count back** (the real N7) | 1 — TTE `TENTRY` straddle |
| P7 gate the ellipsis on rows again | 1 — repeated spelling |
| P8 drop the U+2028/29 arm | 1 — truncation/escaping |
| P9 drop the `examples_capped` carry | 2 — both capping tests |

P3/P4/P5 each killing exactly one, and a *different* one, is the property that was missing: a predicate that loses one arm now reddens the arm it lost.

**A second bookkeeping error, found while regenerating this table.** The first run showed `check::cmt_less_dataset_on_an_analytical_pk_model_is_not_reported` dying under P2–P5 but not P1 — because that test asserts an oral model is *not* reported, which is the behaviour this round changes, so it was failing at baseline and polluting four rows. I had measured blast radius with `--lib` only, on a change that crosses into `tests/`. The table above was regenerated from a tree whose `--lib`, `--test check_command` and `--test cmt_defaulting_scope` legs were each verified green first. That control is now `one_cpt_iv`, with a new `..._on_an_analytical_oral_model_is_reported` on the other side of the straddle.

### Gates at `c32bd0c8`

```
tools/preflight.sh                        OK (fmt, check, clippy, rustdoc, docs, public-API unchanged)
ferx-core/ci --workspace --lib            5418 passed, 0 failed, 26 ignored
ferx-core/ci --workspace --tests          6504 passed, 0 failed, 218 binaries
ci,markov,nn --tests                      6135 passed, 0 failed, 195 binaries
```

Counts reconciled against numbers written down before the runs. `--lib` 5410 → 5418 is +10 added / −2 removed, verified by name diff, the two removals being fixtures the widening invalidated. `--tests` 6493 → 6504 is +11 = 8 lib tests (those targets run here too) + 2 scope + 1 `check_command`.

One thing that reconciliation surfaced and a bare "9/9 killed" would have hidden: `a_tte_row_dropped_for_entry_after_event_…` — the test that kills P6 — is `#[cfg(feature = "survival")]` and therefore **does not compile under `ferx-core/ci` at all**. It is covered by the `ci,markov,nn` leg locally and by CI's TTE/CTMM job (`markov = ["survival"]` makes it a strict superset), and I confirmed by name that it actually executed there. A green `--lib` alone would have been no evidence for it.

### Not addressed

- **Codecov patch coverage** — CI is the authority; the two coverage jobs are still running.
- **No new NONMEM run.** The dose-channel widening rests on the existing `dose_cmt_advan*` anchors (#375) cited above. The observation-dispatch channels remain anchored ferx-vs-ferx, which I am flagging rather than claiming as equivalent.

TeunP:
## Round 3 — two blocking findings, and a split

An independent review of `c32bd0c8` found two blocking defects. Both reproduced with my own fixtures, both confirmed. One of them is a regression **this PR introduced**, and it invalidates a claim in my previous comment.

### Correction to [the comment above](https://github.com/FeRx-NLME/ferx-core/pull/1404#issuecomment-5718342987)

I wrote, under finding 3: *"a TTE-only model on a CMT-less dataset now emits no `W_CMT_DEFAULTED`, just `E_ENDPOINT_NO_RECORDS`."* That was the behaviour of the `!m.is_empty()` gate, and **the gate was wrong and is now reverted** — endpoint models report again. The sentence is false as of `0a837335`.

### Finding A (blocking, mine): the `!m.is_empty()` gate silenced a real mis-routing

My justification for gating `ErrorSpec::PerCmt` on a non-empty map was that `E_ENDPOINT_NO_RECORDS` already names the missing column. That reasoning covers the **absent-column** cause only. `W_CMT_DEFAULTED` also fires for a *missing* or *unparseable cell*, and in those cases every endpoint keeps its rows, so `E_ENDPOINT_NO_RECORDS`'s `routed.get(&cmt) == 0` condition cannot fire — and the defaulted row is silently re-routed **between endpoints**.

Measured on `examples/tte_competing_risks.ferx` with its endpoints moved to `cmt = 1` / `cmt = 2`, one event row's cell spelled `x` instead of `2`:

| dataset | OFV | `ferx check` (with the gate) |
|---|---|---|
| cell spelled `2` | 27.8497 | `0 warning(s)` |
| cell spelled `x` | **28.3610** | `0 warning(s)` |

The event moves from `cause_b` to `cause_a` with no diagnostic. The **parent commit warned** here; my gate silenced it. Reverted, and the test that pinned the gate is inverted to record why. After the revert that same fixture reports `1 row(s) had a CMT cell that is not a compartment index ("x")`, and the control stays silent.

### Finding B (blocking, pre-existing): `[data_selection]` is a third CMT channel

`resolve_row_cmt`'s own doc says it is shared by "the dose row, the observation row **and the `[data]` selection filter's `RowContext`**" — and the filter is handed the *defaulted* value. The predicate never asks whether a `[data_selection]` clause mentions `CMT`. Measured, `pk one_cpt_iv` + `ignore = CMT == 2`, one observation cell `2` vs `x`:

| dataset | observations kept | OFV | `ferx check` |
|---|---|---|---|
| cell spelled `2` | 3 | −4.6162 | `ok — 0 warning(s)` |
| cell spelled `x` | **4** | **−6.9517** | `ok — 0 warning(s)` |

The row the filter was told to drop is silently kept and scored. Embarrassing detail: my own round-1 test `filter_context_cmt_resolves_like_the_dose_site` exists precisely because the filter reads the defaulted CMT. I tested that and still left the channel out of the predicate.

### The split

This predicate has now been wrong in **three consecutive rounds**:

| round | what the list missed |
|---|---|
| 1 | observation-side dispatch entirely |
| 2 | `OdeSpec::readout`, and the analytical dose channel |
| 3 | `[data_selection]`, plus the narrowing above |

Each time I widened it reactively; each time a reviewer found the next hole. That is the repo's own stop condition (`/ferx-resolve` §7, `feedback_stop_iterating_split_instead`), so rather than a fourth widening pass by the same author in the same session — the configuration that produced all three misses — the scope question is **cut to [#1409](https://github.com/FeRx-NLME/ferx-core/issues/1409)** (0.4.0 epic row 36). That issue carries both measurements above, the enumeration of CMT's consumers, and two structural asks: make endpoint routing explicit instead of riding on an empty `ErrorSpec::PerCmt({})`, and add a test that fails when a new consumer appears.

**What I should say plainly about the round-3 anchor:** I described `tests/cmt_defaulting_scope.rs` as making the predicate self-verifying. It derives the **dose** channel across all ten topologies and does that well — but it never touches observation dispatch, endpoint routing or the filter, which is exactly where both of these findings live. It verifies one third of the predicate, and I presented it as verifying the whole thing.

### What stays in this PR

Everything that is converged, and nothing that is narrower than the parent:

- the float-`CMT` parse fix — the actual #1009 bug, NONMEM-anchored on three re-derivable streams;
- the reader's counting, cause-clause, ADDL-expansion, truncation and escaping work;
- `W_CMT_DEFAULTED` itself, on a predicate that is now **strictly wider** than `2e6bb7d8` at every point.

### Also fixed this round, from the same review

| # | Finding | Fix |
|---|---|---|
| 3 | The warning text named `states = [...]` and blamed a NONMEM `DEFDOSE`, on a message the reader (model-blind) also sends to analytical models | **Fixed.** Measured on `examples/schnider.ferx` (a `pk three_cpt_iv`, no `[odes]` anywhere): the old text advised a block the model does not have, and the DEFDOSE rationale is false there — ADVAN11's DEFDOSE *is* compartment 1. Rewritten engine-neutral; a test now asserts the message never contains `states = [...]` or `DEFDOSE`, and P10 mutates it back |
| 4 | An empty `probe_cmts` list made a scope-test row pass by asking nothing | **Fixed.** Confirmed by the reviewer's mutation — emptying `OneCptIv`'s list left both tests green. Now asserted per model |
| 5 | The degenerate test claimed transit/IG are rejected by `E_DOSE_CMT_OUT_OF_RANGE` | **Fixed.** Wrong: that is what `one_cpt_iv` gives, and `check_dose_compartments` bounds on `n_states`, which transit/IG have 2–3 of. The real codes are `E_TRANSIT_UNSUPPORTED` / `E_IG_UNSUPPORTED`, now **asserted** rather than described — `try_preds` only read `report.valid`, so the wrong claim sat unchallenged |
| 7 | `assert_eq!(w.lines().count(), 1)` cannot observe the U+2028 fix beside it | **Fixed.** `str::lines` splits on `\n` only, so that assertion was green with or without the arm it was written to guard. Now asserts no line-breaking character reaches the message at all |
| 6, 8, 9 | `pk_model` placeholder reachability; `channels` count vs highest index; the `ScalingSpec`/`ErrorSpec` asymmetry | **Filed** into #1409 rather than patched here |

### Round-4 mutation testing

10 mutations, **10 killed, 0 survived** — `--no-fail-fast`, byte-copy restore with `touch`, snapshot verified byte-identical first.

| mutation | oracles killed |
|---|---|
| P1 drop the analytical topology branch | 4 |
| P2 answer `n_states` instead of live channels | 2 |
| P3 drop the `ode_spec` per-CMT readout arm | 1 |
| P4 drop the `analytic_readout` per-CMT arm | 1 |
| **P5 re-add the reverted `!is_empty()` gate** | 1 — the endpoint test recording why it went |
| **P10 name `states = [...]` again** | 1 — the analytical-message assertion |
| P6 move the obs count back (the real N7) | 1 |
| P7 gate the ellipsis on rows again | 1 |
| P8 drop the U+2028/29 arm | 1 |
| P9 drop the `examples_capped` carry | 2 |

P5 and P10 are the ones that matter for this round: without them, the regression could be silently re-introduced and the false message text silently restored, with every test still green.

### Gates at `0a837335`

```
tools/preflight.sh                        OK (fmt, check, clippy, rustdoc, docs, public-API unchanged)
ferx-core/ci --workspace --lib            5418 passed, 0 failed, 26 ignored
ferx-core/ci --workspace --tests          6504 passed, 0 failed, 218 binaries
ci,markov,nn --tests                      6135 passed, 0 failed, 195 binaries
cargo run -p docs-lint -- --check         OK, no new baseline entry
```

No public-API movement anywhere on this branch, so no `ferx-r` lock bump is needed.

Behaviour re-measured after the revert on the two fixtures above: the competing-risks bad cell now reports `1 row(s) had a CMT cell that is not a compartment index ("x")`, its control stays silent, and `examples/schnider.ferx` no longer receives a message naming `states = [...]` or `DEFDOSE` (0 occurrences).

Issue #1328, opened 2026-09-10
bug(nn): [covariate_nn] + block_omega runs away on a 600-subject busulfan subset — ρ(CL,V1) at the packed guard, ω²_V1 → 100–450, reported converged; diagonal Ω and the analytic twin are fine


roninsightrx:
## Summary

A `[covariate_nn]` model whose η's share a `block_omega` runs away under FOCEI on a 600-subject busulfan subset: the block's correlation coordinate hits the packed upper guard (+10, "runaway"), `ω²_V1` climbs to 100–450 (CV > 1000 %), `ω²_V2` collapses, and in one arm `ADD_ERR` goes to 0.0003 ng/mL. All three block-Ω DCM arms on that subset do it (h4, h8, h8 with `nn_l2 = 1, nn_smooth = 10`); the analytic model with the same block on the same subset does not, the same DCMs on the full 1215-subject train split do not, and the same DCMs with a **diagonal** Ω on the same 600 subjects converge cleanly with sensible ω's.

Found while running the DCM evaluation for #1310 on `ferx-testdata/busulfan_dominic`.

## Reproduction (minutes)

`ferx-testdata/busulfan_dominic/ferx/dcm/dcm_h4_s600.ferx` on `data/train_s600.csv` (600 subjects, 4 threads, ~5 min). 4→4→4 network, 40 weights, no IOV:

```
[parameters]
  block_omega (ETA_CL, ETA_V1) = [0.068, 0.038, 0.038]
  omega ETA_V2 ~ 0.077
  sigma PROP_ERR ~ 0.05 (sd)
  sigma ADD_ERR  ~ 15 (sd)
[covariate_nn TYPICAL_PK]
  inputs = [LAGE, LWT, LHT, SEX]  center = [2.639, 3.727, 4.91, 0]  scale = [1.548, 0.907, 0.342, 1]
  outputs = [CL, V1, Q, V2]  layers = [4]  activation = tanh  output = softplus  init = [11.6, 46.5, 14.3, 10.8]
[individual_parameters]
  CL = TYPICAL_PK.CL * exp(ETA_CL)
  V1 = TYPICAL_PK.V1 * exp(ETA_V1)
  Q  = TYPICAL_PK.Q
  V2 = TYPICAL_PK.V2 * exp(ETA_V2)
[structural_model]
  pk two_cpt_iv(cl=CL, v1=V1, q=Q, v2=V2)
[error_model]
[...]

PR #1406, opened 2026-09-17
test(api): pin the block-Ω rail demotion end to end [closes #1328]

roninsightrx:
## Why

#1328 reports that a `[covariate_nn]` model whose η's share a `block_omega` runs away on a
600-subject busulfan subset — ρ(CL,V1) at the ±10 packed rail, ω²_V1 → 100–450 — **and reports
`converged: true`**. Row 21 of #1350 owns the flag half and asks for a re-measure first, because
#1303 / PR #1383 landed a `converged` gate after the row was written.

**Both halves of the report are stale, and this PR is the evidence plus the end-to-end arm that
was genuinely missing.**

### 1. The flag was already correct — including in the artifacts the issue cites

`ferx-testdata/busulfan_dominic/ferx/dcm/run/dcm_h8_l2_1_s10_s600/dcm_h8_l2_1_s10_s600-fit.yaml`
(Sep 10 11:10, `ferx_version: 0.3.1`) is the run the issue tables — same `ofv: 60992.180217`,
same `covariance: 3.050067`, same `variance: 169.723000`. It reads:

```yaml
model:
  converged: false
```

and its warning already ends *"…are reported converged: false"*. The same holds for
`dcm_h4_s600` and `dcm_h8_s600`: every fit in that run dir carrying a `runaway` hit reports
`converged: false`. #1118 (`b61e6216`, 2026-09-02) landed that demotion **eight days before the
issue was filed**, and #1205 extended it to *either* rail of an Ω off-diagonal. Nothing in the
`converged` path was broken, and #1383 is not what fixed it — the objective here is a perfectly
ordinary finite 60992.18, so #1303's gate never fires on this run.

### 2. The runaway itself no longer reproduces on `main`

Rebuilt `ferx` at `233246d1` with `--features nn` and re-ran all three block-Ω arms from the
report on `data/train_s600.csv` (600 subjects, `--threads 2`):

| arm | ω²_CL | ω²_V1 | ω²_V2 | cov(CL,V1) | PROP | ADD | OFV | `converged` | guard warning |
|---|---|---|---|---|---|---|---|---|---|
| h4 — #1328 | 0.061 | **450.4** | 0.017 | **2.473 (rail)** | 0.103 | 0.0003 | 60036.5 | false | runaway + collapse |
| h4 — now | 0.0694 | **0.0224** | 0.863 | **0.03296** | 0.0957 | 9.52 | **59755.6** | true | none |
| h8 — #1328 | 0.061 | **100.0** | 0.018 | **2.48 (rail)** | 0.096 | 22.8 | 59900.1 | false | runaway |
| h8 — now | 0.0693 | **0.0227** | 1.515 | **0.03296** | 0.0954 | 7.50 | **59668.8** | false¹ | none |
| h8 `nn_l2=1, nn_smooth=10` — #1328 | 0.093 | **169.7** | 0.028 | **3.05 (rail)** | 0.097 | 21.7 | 60992.2 | false | runaway + collapse |
| h8 `nn_l2=1, nn_smooth=10` — now | 0.0676 | **0.0241** | 0.336 | **0.03325** | 0.0961 | 11.62 | **59757.2** | true | none |

¹ h8 stops with `Outer optimization did not converge` and `W_EBE_START_DEPENDENT` (#1386) — not a
guard hit.

Every arm lands 231–1235 OFV **better** than the runaway it used to find, with the block
covariance interior (0.033 vs pinned at the rail), ω²_V1 back to 0.022–0.024, and `ADD_ERR` no
longer collapsing to 3e-4. Wall time 21–44 s at 2 threads against 252–625 s at 4. I did not
bisect which of the ~40 commits since `0.3.1` moved the trajectory, and I am not claiming the
runaway is now *impossible* — only that the published reprex does not reach the rail on `main`.

The weak identification the issue diagnoses is still visible, in a different place: ω²_V2 is now
0.34–1.51 (CV 58–123 %) with 34–41 % η-shrinkage on all three arms. That is the same
poorly-separated V1/Q/V2 split, no longer expressed as a rail runaway.

## What changed

**Tests only. No production code.** Two tests, at the two tiers the property splits across.

### 1. `tests/runaway_guard_demotes_convergence.rs::fit_demotes_converged_when_a_block_omega_off_diagonal_sits_on_either_rail`

The existing end-to-end arm pins a **Σ** ceiling, which `fit()` stores verbatim. #1328's
coordinate is an Ω **off-diagonal**, whose rail lives on the raw Cholesky element `L[i,j]`. The
new test drives `examples/warfarin_block_omega.ferx` through `fit()` with `L₂₁` set to `+10` and
to `−10` (evaluation-only IMP, the same device the Σ arm uses to reach the fit-end check without
a convergence loop) and asserts `converged: false`, a `Critical` `parameter_at_runaway_guard`
entry, `verdict: runaway` on *both* rails (#1205), and the `details` payload a programmatic
consumer reads.

**The demotion is attributed, not merely observed.** `ω²_V ≥ L₂₁² = 100` at either rail — the
same 100–450 #1328 reports — so the evaluation is a bad one and a test asserting only
`!converged` would pass on #1303's objective gate firing instead. Two premise assertions rule
that out: `ofv.is_finite() && |ofv| < 1e14` (`DIVERGENCE_OFV`) and no `W_NONFINITE_OBJECTIVE`
in `warnings`. A control fit at the declared block (`[0.09, 0.02, 0.04]`, L₂₁ = 0.067) asserts
`converged: true` with no guard entry, so the demotion is attributable to the rail rather than to
the block or to the eval-only path.

### 2. `src/api/tests/simulate_with_uncertainty_tests.rs::a_block_omega_off_diagonal_at_the_rail_survives_pack_clamp_unpack`

**Added after a Codex review of the first push, which was right.** The first version of this PR
claimed the Tier-2 test above crossed the `pack_with_bounds` → `clamp_to_bounds` →
`unpack_params` round trip that every estimating path runs. It does not: the standalone
`imp_eval_only` branch (`api/fit.rs:1693`) hands `stage_params` straight to `run_inner_loop_warm`
and `pop_nll`, so the fit-end walk re-reads the very Cholesky factor the test wrote. Measured —
replacing the off-diagonal arm of `unpack_params`' reconstruction with `0.0` leaves that whole
file green.

So the round trip is pinned where it can be, at Tier 1 (`pack_with_bounds` / `clamp_to_bounds`
are `pub(crate)`): pack the rail, clamp, unpack, assert the rebuilt `L[1,0]` *is* the rail, then
assert `runaway_guard_warning` still demotes with `verdict: runaway` — both rails. The premise
assertions state that the coordinate under test is the Ω off-diagonal and that the rail is that
coordinate's own bound, so the clamp cannot be silently doing the work.

The Tier-2 test's doc comment now says what it does and does not cover, and names this test as
where the round trip lives, so the two are not read as one property written twice.

## Alternatives considered

**Post the evidence and add nothing.** That is what row 21 asks for if the re-measure comes back
clean, and it was the default. The reason it did not win: the guard-demotion path for #1328's own
coordinate — a block Ω rail — was covered by no test at any tier that crossed either the `fit()`
wiring or the pack/unpack round trip. The pre-existing unit tests own the verdict rule with
neither; the Σ Tier-2 arm owns the wiring for a coordinate `fit()` stores verbatim. The two new
tests cost 0.2 s together.

**Changing the ±10 rail, or refusing it as a start.** Out of scope for this row, and the
measurement does not support it: nothing on `main` reaches the rail on this dataset any more. If
the block correlation should be made *unable* to run there rather than currently-not-running
there, that is a separate design change to `compute_bounds` and wants its own issue — see Open
questions.

## Cross-repo dependency

No public-API change; `tools/update-public-api.sh --check` reports the baseline unchanged. The
test uses only already-public items (`OmegaMatrix::from_chol_factor` and its `chol` / `eta_names`
/ `diagonal` / `free_mask` fields), which is also the ferx-r reachability check for this fixture.

## Breaking changes
- [ ] Public Rust API
- [x] None

## Numerical validation
- [x] Compared against: prior ferx commit — the `0.3.1` run dir the issue cites (table above)
- [x] Not applicable for the diff itself — no estimator, likelihood, gradient or PK/ODE kernel
      changed; no `*_g<T: PkNum>` body touched.

**No NONMEM anchor, and not because one is awkward to build.** The quantity under test is ferx's
own convergence verdict against ferx's own packed-space rail. NONMEM has no spelling for "which
of my internal guards did I stop on" — its Cholesky parameterisation is not ferx's and its
`MINIMIZATION SUCCESSFUL` line is a different stop rule against an absent clamp. The inputs are
exact hand-computed values rather than tolerances: `L₂₁ = ±10` is the literal rail from
`compute_bounds`, and the assertion on the returned packed coordinate is `assert_eq!`, not a
tolerance. The re-measure above is anchored the only way it can be — against the previous ferx
build's own artifacts, arm by arm.

## Performance
- [x] No significant impact expected — two tests, 0.2 s together.

## Tests
- [x] Unit test added in the same module (`cargo test --lib` passes) — the Tier-1 round trip
- [x] Regression test added that fails without this fix — both, see the matrix

Tier split, and why it is a split rather than one test: "`fit()` applies the guard rule to the
`FitResult` a consumer reads" cannot be reached from a `src/` unit test, and
"`pack → clamp → unpack` preserves the rail" cannot be reached from `tests/`, because
`pack_with_bounds` and `clamp_to_bounds` are `pub(crate)`. The Tier-2 arm returns after a single
evaluation-only IMP pass with no convergence loop, so it stays in the PR job.

| mutation | Tier-1 round trip | Tier-2 Ω arm | Tier-2 Σ arm |
|---|---|---|---|
| `if has_runaway { *converged = false }` → `if has_runaway && false` (**the fix off**) | **RED** | **RED** (`rail 10`) | **RED** |
| `RunawayGuardHit::is_runaway` → `self.side == "upper"` (drop the off-diagonal clause) | **RED**¹ | **RED** (`rail -10`) | green |
| `unpack_params`' off-diagonal reconstruction → `0.0` | **RED** (`left: 0.0, right: 10.0`) | green | green |
| `pack_with_bounds` reads a re-factorised `cholesky(omega.matrix).l()` instead of the cached `omega.chol` | green² | green | green |

¹ by construction — it asserts `verdict: runaway` on the `−10` arm; the same edit also reddens the
pre-existing Tier-1 `omega_off_diagonal_lower_rail_is_a_runaway_not_a_collapse`.

² measured, and reported *because* it is green: re-factorising `[[1, ±10], [±10, 101]]` recovers
`L₂₁ = ±10` exactly, so the `assert_eq!` on the reported packed coordinate measures that
exactness rather than discriminating a defect.

**Row 3 is the one that matters, and it is the reviewer's.** A Codex review of the first push
pointed out that the Tier-2 test never crosses the pack/clamp/unpack round trip it claimed to —
the standalone `imp_eval_only` branch hands `stage_params` straight to the inner loop. Verified:
that mutation left the whole Tier-2 file green. The Tier-1 test is the fix, and row 3 is the
measurement that it is not a fourth way of saying the same thing.

**Honest limit of rows 1 and 2:** neither is killed by a *single* new test alone — row 1 also
reddens the Σ arm, row 2 also reddens the pre-existing Tier-1 off-diagonal test. Only row 3 is
uniquely load-bearing, and it belongs to the Tier-1 test.

- [x] Not applicable — no new DSL syntax, `[fit_options]` key, output field or estimator.

## Docs
- [x] No user-visible change

No `CHANGELOG.md` entry: the diff is a test. The guard's behaviour is already documented — its
verdict rule and the `converged` demotion in `docs/estimation/parameterization.qmd`
("A runaway hit **demotes `converged` to `false`**…", #1118), and the warning's severity and
remediation in `docs/warnings.qmd`'s `parameter_at_runaway_guard` row. Both were re-read against
the code for this PR and are accurate.

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy `--all-targets`,
      rustdoc, docs, public-API
- [x] `tools/preflight.sh release-semantics` green
- [x] `Dual2` path — no `*_g<T: PkNum>` body touched
- [x] No version bump (no API change)

```
cargo test --test runaway_guard_demotes_convergence --no-default-features --features ci,survival
  -> 2 passed / 0 failed  (0.12s)

cargo test --lib --no-default-features --features ci,survival a_block_omega_off_diagonal
  -> 1 passed / 0 failed  (0.08s)

cargo test --lib --no-default-features --features ci,survival runaway
  -> 7 passed / 0 failed  (89.9s)
```

### ferx-site / ferx-book

### Example execution
- [x] No example execution step needed — no example's numbers move; `examples/warfarin_block_omega.ferx`
      is *read* by the new test, unchanged. The `--features nn` release binary used for the
      re-measure was built from this branch at `233246d1` and the three busulfan arms above are
      its output.

## Reviewer hints

- **Start with §1 of Why.** If you disagree that the `converged: true` line in #1328 is a
  write-up error rather than a real observation, that is the claim to argue with — the file path
  and the three matching numbers are in that paragraph, and `grep -l "runaway" run/*/*-fit.yaml`
  over that run dir returns exactly the three arms, all `converged: false`.
- **The premise assertions in the Tier-2 test are the load-bearing part**, not the `!converged`
  line. Without them the test passes whenever the objective happens to blow up at ω²_V = 100 and
  says nothing about the guard.
- **Row 3 of the mutation matrix** is the reviewer-found hole and the reason the Tier-1 test
  exists; if you want to argue the two tests are one property written twice, that row is the
  evidence they are not.
- **Skimmable:** the control fit at the top of the Tier-2 test is the same shape as the Σ arm's.

## Open questions

1. **Should the block correlation be bounded short of the ±10 packed rail?** The issue leaves it
   open and this PR does not settle it — it cannot, because nothing on `main` reaches the rail on
   this dataset any more. Reachability elsewhere is untested. If it is worth making structurally
   impossible, say the word and I will file it with the `0.3.1` artifacts as the motivating case.
2. **`W_EBE_START_DEPENDENT` fires on the h8 arm** (cold re-solve 64.1 OFV worse than the reported
   optimum) and the h4 arm shows a 37-unit gap between the optimizer's best-seen 59718.49 and the
   published 59755.58 without warning. That is #1386's territory, not this row's, but it is on the
   same dataset and someone should look.


roninsightrx:
## Codex review

_Automated review by `codex exec` (GPT-5.x, read-only: no build or test run) against `origin/main`. Findings are the reviewer's claims, not verified by a run — see the unverified caveats below._

The PR should not merge as written. Its only new test does not exercise the packed-space round trip it claims to protect, so the reported block-Ω regression could return without failing this test. No production or API concerns are introduced. Per the constraint, I did not compile or execute anything.

### Blocking findings

- `tests/runaway_guard_demotes_convergence.rs:125` — **high** — With `method = Imp` and `imp_eval_only = true`, standalone `fit()` stores `stage_params.clone()` directly in the result; it never clamps or calls `unpack_params`. The fit-end guard merely repacks the same `OmegaMatrix` constructed at lines 118–123. Thus the smallest relevant regression—breaking the off-diagonal reconstruction in `unpack_params`—leaves this test green, and the test duplicates the existing Tier-1 off-diagonal classification test plus the existing sigma fit-wiring test. — Add a Tier-1 test that sends both ±10 coordinates through the real `pack_with_bounds` → clamp → `unpack_params` path before invoking the guard, or use an estimator fixture whose returned parameters actually originate from that path. Keep the existing sigma integration test to cover the generic `fit()` wiring.

### Nits

None.

roninsightrx:
## Review response — 1 finding, **CONFIRMED**

> the new test in `tests/runaway_guard_demotes_convergence.rs` never routes through
> `pack_with_bounds` → `clamp` → `unpack_params` under `method = Imp` with `imp_eval_only`, so
> breaking the off-diagonal reconstruction leaves it green

**CONFIRMED by mutation.** This was the load-bearing claim in the PR body and it was wrong.

### The mutation

Replaced the off-diagonal arm of `unpack_params`' Cholesky reconstruction with `0.0`:

```rust
// src/estimation/parameterization.rs
l[(i, j)] = if i == j { v[idx].exp() } else { 0.0 };   // was: v[idx]
```

```
cargo test --test runaway_guard_demotes_convergence --no-default-features --features ci,survival
test fit_demotes_converged_when_sigma_sits_on_the_internal_ceiling ... ok
test fit_demotes_converged_when_a_block_omega_off_diagonal_sits_on_either_rail ... ok
test result: ok. 2 passed; 0 failed
```

Green, exactly as reported. The mechanism is `api/fit.rs:1693`: the standalone `imp_eval_only`
branch hands `stage_params` to `run_inner_loop_warm` and calls `pop_nll` on them — no
`pack_with_bounds` → `clamp_to_bounds` → `unpack_params` anywhere on that path. So the test's
`assert_eq!(packed_estimate, rail)` was re-reading the same cached Cholesky factor it had just
written, and my earlier "row 3 is green because the round trip is exact" conclusion was measuring
a round trip that never happened.

### The fix

Pushed in `77a3e61f`:

1. **New Tier-1 test** `a_block_omega_off_diagonal_at_the_rail_survives_pack_clamp_unpack`
   (`src/api/tests/simulate_with_uncertainty_tests.rs`), which runs the production sequence
   explicitly — `pack_with_bounds` → `clamp_to_bounds` → `unpack_params` → `runaway_guard_warning`
   — on both rails, asserts the rebuilt `L[1,0]` *is* the rail, and then asserts the demotion. It
   has to live at Tier 1 because `pack_with_bounds` and `clamp_to_bounds` are `pub(crate)`.

   Under the same mutation:

   ```
   test a_block_omega_off_diagonal_at_the_rail_survives_pack_clamp_unpack ... FAILED
   assertion `left == right` failed: rail 10: the off-diagonal must survive pack → clamp → unpack
     left: 0.0
    right: 10.0
   ```

   and under the fix-off mutation (`if has_runaway && false`) it fails on the `!converged`
   assertion at line 1447.

2. **Corrected the Tier-2 test's doc comment and assertion message.** It now says what it actually
   pins — the fit-end *wiring* for a block-Ω hit (demotion, `Critical`, `details` payload, both
   rails) — and states in so many words that the eval-only IMP path does not round-trip the
   params, that breaking `unpack_params` leaves that file green, and where the round trip is
   pinned instead.

### Updated mutation matrix

| mutation | Tier-1 round trip (new) | Tier-2 Ω arm | Tier-2 Σ arm |
|---|---|---|---|
| `if has_runaway { *converged = false }` → `&& false` (**fix off**) | **RED** | **RED** | **RED** |
| `is_runaway` → `self.side == "upper"` | **RED**¹ | **RED** (`rail -10`) | green |
| `unpack_params` off-diagonal → `0.0` | **RED** (`left: 0.0`) | green | green |
| `pack_with_bounds` reads `cholesky(omega.matrix).l()` instead of the cached `omega.chol` | green² | green | green |

¹ by construction — it asserts `verdict: runaway` on the `−10` arm; the same edit also reddens the
pre-existing `omega_off_diagonal_lower_rail_is_a_runaway_not_a_collapse`.
² measured: re-factorising `[[1, ±10], [±10, 101]]` recovers `L₂₁ = ±10` exactly, so this remains
a green cell and is reported as one.

PR body updated to match.

PR #1407, opened 2026-09-17
fix(ode): build the frozen-replay verifier's break list with the driver's own builder [closes #1188]

### LINKED ISSUE #1188: bug(adaptive): the frozen-replay verifier drops zero-order absorption, making its own oracle drug-free
## Summary

`adaptive_frozen_replay_tv` builds its own segment timeline with **neither**
`push_zero_order_break_times` nor `push_route_lag_break_times`, and then calls
`integrate_segment` — which applies `active_zero_order_inputs`' full-containment test. For any
model with a `zero_order` absorption route, the replay trajectory is therefore **drug-free**.

That matters more than a normal diagnostic bug, because per `CLAUDE.md` the frozen-schedule
replay verifier *is* the designated oracle for the adaptive/feedback-dosing family (#391), which
has no NONMEM comparator:

> the **frozen-schedule replay verifier** — rebuild a static subject from the realized dose
> ledger and assert bit-equality with the static engine (the exact internal analogue of a NONMEM
> dose-bookkeeping anchor)

A broken oracle reports spurious mismatches and is symmetrically blind to real ones.

Found while self-reviewing #1174. Pre-existing on `main`, and **not** fixed by #1174.

## Mechanism

`adaptive_frozen_replay_tv` (`src/ode/predictions.rs:4299`) assembles `break_times` from:

- `0.0` and `t_last`
- each `d.time`, and `d.time + dur_eff` for real infusions
- `subject.obs_times`, `subject.pk_only_times`
- `subject.reset_times`
- the caller's `extra_breaks` (the decision times)

then sorts and dedups. It never pushes a zero-order window edge and never pushes a route onset —
and it does not even push `d.time + lag`. It then integrates each segment through
`integrate_segment` (`:2031`), whose `active_zero_order_inputs` call admits a window's constant
rate only if the segment lies entirely inside `[w_start, w_end]`. With neither edge bracketed, no
segment qualifies and the rate is dropped for every window.

The reactive driver it is supposed to mirror bit-for-bit folds in `collect_dose_break_times`
(`:3438`), which has both pushes. So the two sides of the verifier segment the timeline
differently by construction.

## Repro sketch

`[adaptive_dosing]` + a time-varying covariate + a #702 base regimen dosing into
`zero_order(dur=2, lag=1.5)`. The driver breaks at 1.5 and 3.5 and delivers the window; the
replay has neither break and reads zero. `verify_adaptive_frozen_replay` then reports a
dose-bookkeeping mismatch that is an artifact of the verifier, not of the driver.

## Relationship to #1171

Same class — a break-time builder whose segments feed `active_zero_order_inputs` without
bracketing the window. #1171/#1174 fixed two such builders and hardened the shared helper
(`push_zero_order_break_times` now emits both `w_start` and `w_end`, so a caller gets correct
zero-order segmentation from that one call). **This builder calls neither helper at all**, so it
is untouched by that hardening and needs its own fix.

Corrected count for the record: there are **five** break-time builders in `src/ode/predictions.rs`,
not four —

1. `collect_dose_break_times` — correct
2. the hand-inlined timeline in `ode_predictions_event_driven*` — correct
3. `ode_predictions_with_states`' inline builder — fixed in #1174
4. `build_segment_break_times` — fixed in #1174
5. `adaptive_frozen_replay_tv` — **this issue**

## Suggested fix

Call `push_zero_order_break_times` (and `push_route_lag_break_times`) in the replay builder, or
better, have it delegate to `collect_dose_break_times` as the driver already does — the two are
required to agree bit-for-bit, so sharing the builder is the property the verifier actually wants.

Test: a degenerate-oracle case per `CLAUDE.md` — a controller that re-emits a fixed regimen on a
`zero_order(dur, lag)` model must replay bit-identically to `simulate()` on that regimen. That
fails today.

roninsightrx:
## Why

Closes #1188 (epic #1350 row 22, tracker #1217 step 6 / batch A).

`adaptive_frozen_replay_tv` — the static engine `verify_adaptive_frozen_replay` runs on the time-varying / IOV adaptive path — assembled its dose breaks by hand: each `d.time`, plus a real infusion's bioavailability-scaled end. It pushed neither a per-route absorption onset (`push_route_lag_break_times`) nor a `zero_order` window's edges (`push_zero_order_break_times`).

`integrate_segment`'s `active_zero_order_inputs` admits a window's constant rate **only for a segment the window fully contains**. With neither edge bracketed, no segment straddling one qualifies and the rate is dropped for that whole window, so the replay under-delivers the absorbed mass.

This matters more than an ordinary diagnostic bug, because per `CLAUDE.md` the frozen-schedule replay verifier **is** the designated oracle for the adaptive / feedback-dosing family (#391), which has no NONMEM comparator. A broken oracle both reports spurious mismatches and is symmetrically blind to real ones.

**Measured on `main` at `233246d1`, before any change.** Calling the replay engine directly on a lagged `zero_order(dur = 2, lag = 1.5)` two-dose subject, against `ode_predictions_with_extra_breaks` (which #1174 fixed) on the same schedule:

```
obs    : [1.0,  2.0,    3.0,    4.0,    6.5,   7.5,   8.5,    12.0]
replay : [0.0,  0.0,   47.58,  43.05,  33.53, 30.34, 75.03,  52.88]
static : [0.0, 24.385, 69.646, 86.214, 67.14, 60.75, 102.55, 109.32]
```

The static engine matches the exact Bateman/ramp closed form to 1e-14 (`t = 2`: `(50/0.1)·(1 − e^{−0.05}) = 24.385287749`); the replay reads `0.0` there. Worst relative divergence 41 % — and not uniformly "drug-free", because a few observation times happen to bracket a window by accident, which is exactly the kind of partial wrongness a fixture can pass over.

### The issue's user-facing claim is stale — measured, not assumed

The issue says the verifier "blocks a legitimate model" because `simulate()` returns `Err`. The `Err` is real but it is **not the verifier**. Both routes a zero-order dose could take into this engine are typed errors today, measured end to end through `simulate_adaptive`:

| route | what happens on `main` |
|---|---|
| base regimen (#702) into a `zero_order` compartment × time-varying covariate | `ode_predictions_adaptive: a pre-scheduled base dose (index 0) combined with time-varying covariates or inter-occasion variability must be a plain fixed bolus or infusion into a plain compartment; … built-in input-rate (transit / zero-order absorption) … are a #930/#931 follow-up` |
| controller dose into a `zero_order` compartment | `decision 0: compartment 1 is fed by a built-in input-rate (absorption) function; controller dosing into an input-rate compartment is not supported` |

Both are the deliberate, documented scope guards (`docs/model-file/adaptive-dosing.qmd`, *Current scope and limits*), not verifier artifacts. So the defect is **latent** — the #1272 shape: real in the engine, unreachable from any public entry point. It is fixed here so it cannot ship silently *inside* the #930/#931 widening that lifts those guards, which is precisely when the broken builder would first produce a wrong number for a user.

## What changed

One production change: the replay calls `collect_dose_break_times` — the same builder the reactive driver folds its base regimen in with — instead of its hand-rolled loop. `CLAUDE.md`: *when the bug report is "two implementations disagree", the fix is one implementation.* #1171 / #1174 fixed two other copies of this rule; per #1188 this was the fifth break-time builder in the file and the only one neither of those touched.

`pk_params_flat` (the driver's frozen `t = 0` snapshot) is threaded in for this, because that is the snapshot the driver placed *its* edges from — using a per-segment snapshot instead would un-align the pair whenever a route lag or `dur` is covariate-dependent. It is used for break placement only; every integration still reads its own per-event snapshot.

Two things are recorded in comments rather than papered over:

- **the lagtime asymmetry** — the replay's `dose_lagtimes` is all-zeros, which is correct today only because a lagged base dose under TV/IOV is refused upstream and an injected dose cannot carry a lag;
- **the builder-scope asymmetry** — the driver runs this builder over its base regimen only (controller doses get their infusion-end break from `insert_break` at injection), the replay over base ∪ ledger. Today the sets coincide, because a controller dose into an input-rate compartment is refused. A future widening must add the route-lag / zero-order edges at the injection site too.

## Alternatives considered

- **Push the two helpers in the replay builder directly** (the issue's first suggestion). Rejected: it keeps two copies of the rule, which is the class of defect #1171 / #1174 / #1188 all are. It is also strictly weaker, and measurably so — see the `1.311e-5` mutation below, which that edit passes on every #1174-shaped fixture.
- **Also lift the #930/#931 input-rate rejection**, so the model the issue calls legitimate actually runs. Out of scope for this row: that guard covers per-dose SS / lag / modeled-`RATE` bookkeeping as well, each needing its own oracle, and it is the #930/#931 follow-up's decision to make. This PR removes the segmentation blocker that sat under it. Reported on the issue.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` clean); `adaptive_frozen_replay_tv` is a private fn and `verify_adaptive_frozen_replay` is `pub(crate)`.

## Breaking changes
- [ ] Public Rust API
- [x] None

Every currently-reachable adaptive run is bit-identical: all 259 existing `adaptive` tests — including the byte-identical constant-path canaries and both `*_frozen_replay_is_bit_exact` pairs — pass unchanged.

## Numerical validation

Per `CLAUDE.md`, the adaptive / feedback-dosing family has **no NONMEM comparator** (NONMEM has no native feedback dosing), and the documented substitutes are the degenerate oracle and the frozen-schedule replay verifier — which is the object under repair here. So the anchor used is the exception's first branch: an **exact closed form**, computed outside every ferx engine.

A pure accumulator (`dy = 0`) fed by a lagged `zero_order` holds exactly the delivered mass, so the answer is a clamped ramp, per dose, summed:

```
A(t) = Σ_doses clamp((t − t0 − lag) / dur, 0, 1) · F · amt
```

- [x] Compared against: an exact closed form (see above) plus the production static engine `ode_predictions_with_extra_breaks`
- [x] Prior ferx commit `233246d1` (the pre-fix measurement at the top)

```
# before (main, replay engine, lagged zero_order two-dose subject)
t=2.0  replay 0.000000   exact 24.385287749   -> 100 % of the window short
worst relative error over the sample set: 4.096e-1

# after
worst relative error vs the exact ramp:            1.42e-16
worst relative error vs the static engine:         0.0  (bit-identical)
```

The tolerances are **measured, not reasoned**: `1e-12` on the closed-form test (≈7,000× the realised error, 2.5e11× below the tightest mutation) and `1e-14` on the bit-identity test (the tightest mutation sits at `1.311e-5`, nine orders above).

One thing found and deliberately **not** chased: the multi-dose `first_order` input-rate forcing sits ~0.76 % below a naive per-dose superposition closed form at late times. It reproduces identically in `ode_predictions` and `ode_predictions_with_extra_breaks` on `main`, so it is a pre-existing property of that forcing's TAD anchoring, untouched by this PR — which is why the closed-form anchor is scoped to the zero-order route (exact to 1e-14 there) and the first-order route is carried only by the bit-identity test. Noted on the issue.

- [ ] Not applicable

## Performance

The replaced loop and `collect_dose_break_times` are both O(doses); the extra work is one `zero_order_windows` pass and one `input_rate` scan per replay call, and the replay runs once per simulated subject, not per objective evaluation.

## Tests

Two Tier-1 unit tests in `src/ode/predictions_tests.rs` (the sibling `#[path]` module). They are a **pair on purpose**, because neither alone is sufficient — and that is measured, not argued.

**1. `frozen_replay_delivers_the_exact_zero_order_mass`** — pure accumulator, lagged `zero_order`, two doses so the second window opens with the first dose's whole 100 still present (a single-dose fixture cannot see an incoming-side error: the state is zero before a first arrival). Anchored on the closed form, i.e. a reference **outside both engines**, which is what lets it see a defect *inside* the shared builder.

**2. `frozen_replay_segments_a_route_lag_apart_from_a_zero_order_edge`** — a `first_order` route switching on at 0.5 / 6.5 beside a `zero_order` window spanning 1.5–3.5 / 7.5–9.5. Six distinct edges, no two coincident. This is deliberately **not** #1174's fixture shape, where a bare `zero_order`'s route onset *is* its own window start, so one break served both and a builder emitting only one of the two passed the whole suite. Oracle: `ode_predictions_with_extra_breaks` handed the replay's own break set → bit-identical.

Mutation results — the question asked was *what is the smallest edit that removes the fix, and which test dies*:

| mutation | test 1 | test 2 |
|---|---|---|
| restore the hand-rolled `d.time`-only loop | **FAIL** `1.0e0` | **FAIL** `4.096e-1` |
| delete `push_zero_order_break_times` from the shared `collect_dose_break_times` | **FAIL** `2.5e-1` | green (`0.0`) |
| delete `push_route_lag_break_times` from the shared builder | green | green (5 pre-existing #1174 tests fail instead) |
| **the weaker fix**: hand-rolled loop + `push_zero_order_break_times` alone | green | **FAIL** `1.311e-5` |

Three things that table is there to make explicit:

- The last row is the #1174 trap reproduced deliberately. That edit is what a reviewer reaches for first (#1174's own note says a caller gets correct zero-order segmentation from `push_zero_order_break_times` alone), it passes the entire suite on any #1174-shaped fixture, and **only** the separated-edge fixture kills it.
- Row 2 shows test 2 is blind to a defect inside the shared builder — both its sides run that helper and move together. That is stated in test 2's doc comment rather than left to be rediscovered, and it is the reason test 1's oracle is a closed form.
- Row 3 is not a hole: the shared helper is already pinned by #1174's own tests (`route_lagged_first_order_onset_reaches_both_dense_builders` and four others fail), so duplicating that here would be the "two redundant gates cover for each other" shape `CLAUDE.md` warns about.

Both tests assert `is_finite()` before folding into a `worst = worst.max(..)` accumulator — `f64::max` discards `NaN`, so a solve returning `NaN` would otherwise pass on the strength of the finite samples — and both assert the **straddle** they rely on (that the second dose really does land on a non-empty compartment) rather than assuming it.

- [x] Unit test(s) added in the same module (`cargo test --lib` passes: 4,730 passed, 0 failed)
- [ ] No new tests needed

- [x] Not applicable — internal verifier; the defect is latent and no reachable user model changes behaviour.

## Docs
- [x] `docs/` updated — `docs/model-file/adaptive-dosing.qmd`, *Verification (on by default)*: the replay builds its segment structure through the same break-time builder as the driver, and why an unbracketed `zero_order` edge would make the verifier report its own artifact. `cargo run -p docs-lint -- --check` clean.
- [ ] New pages linked in `docs/_quarto.yml` — no new page.
- [x] `CHANGELOG.md` `[Unreleased]` → `Fixed`, stating plainly that the defect is latent and every reachable run is bit-identical.
- [ ] No user-visible change

A second, accurate edit to that page's *Current scope and limits* section was **reverted**: it is in `docs/.lint-baseline` at 7,642 characters and the baseline is a ratchet, so the addition (7,856) was a lint failure. The `Verification` section carries the statement instead; splitting a 7.6 KB section is its own change, not this one's.

## Checklist
- [x] `tools/preflight.sh release-semantics` green (5,081 passed) — the gate the bare script omits and CI enforces
- [x] Functions on the `Dual2` sensitivity path stay differentiable — `sens/` is not touched (tracker #1217's area rule: value-path work does not touch `sens/`)

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key — none
- [ ] New data file — none

### ferx-book

### Example execution
- [x] No example execution step needed — internal fix, no user-visible change, no `examples/` file touched.

## Reviewer hints

**Focus on** the one production hunk in `adaptive_frozen_replay_tv`, specifically the three questions behind it:

1. *Is `collect_dose_break_times` a superset of the loop it replaces?* Yes for every reachable subject: `dose.time + lag` with `lag = 0` is `d.time`; the F-scaled infusion end is the same call; `ss_seeded_at_record` is `dose.ss && ii > 0 && lag > 0`, so false with all-zero lags; `ss_residual_infusion_end` needs `d.ss`. The route-lag and zero-order pushes are no-ops unless a dose lands in an input-rate compartment, which is refused upstream — which is why all 259 adaptive tests are unchanged.
2. *Is `pk_params_flat` the right snapshot?* It is what the driver passes to the same builder at `predictions.rs:4506`. Passing a per-segment snapshot would be the un-aligning change.
3. *Does the asymmetry comment hold?* The driver's builder call is base-only; the replay's is base ∪ ledger. They coincide only because `reject_unsupported_dose_compartment` refuses a controller dose into an input-rate compartment.

**Can be skimmed:** the two test fixtures are hand-built `OdeSpec`s in the style of `zero_order_accumulator_spec` / `one_cpt_ode_spec` already in the file; free PK slots 20–23 carry `dur` / the two route lags / the transfer rate.

**Worth a second opinion:** whether the `Verification` doc paragraph is the right home given the scope paragraph could not grow under the R1 ratchet.

## Open questions

1. **Should the #930/#931 input-rate rejection now be narrowed?** The segmentation half of what that guard's message cites is fixed here, and the remaining per-segment input-rate work (`prepare_input_rates`, the per-segment zero-order windows, the bolus suppression) already runs identically in both engines. I did not measure a widened guard end to end and did not ship one — it needs its own oracle per absorption kind and belongs to that follow-up. Flagging it so the follow-up knows this blocker is gone.
2. **The `first_order` multi-dose superposition gap** (~0.76 % at late times vs a naive per-dose closed form, identical across `ode_predictions` and the static engine on `main`). I believe it is TAD anchoring rather than a defect, but I did not chase it and it is not this PR's. Should it be its own issue?
3. `slow-tests.yml` is dispatched manually on this branch (this touches `ode/predictions.rs`, on the estimation path); result reported in a comment below.


roninsightrx:
## Codex review

_Automated review by `codex exec` (read-only: no build or test run) against `origin/main`._

The diff looks correct in static review. The shared break builder preserves currently reachable behavior, and the Tier-1 regression tests use a closed-form oracle, non-degenerate multi-dose fixtures, explicit finite checks, and measured tolerances. Documentation and changelog updates are present; no public API or dependency boundary changes occur. `git diff --check` is clean. Per instruction, I did not compile or run tests.

## Blocking findings

None.

## Nits

None.

roninsightrx:
Thanks — noting for the record that this review was **read-only by instruction: nothing was compiled or run**. A clean static verdict on a break-timeline change is weak evidence, because the whole failure mode here is a fixture that looks non-degenerate and isn't. #1174 shipped green with exactly that. So here is what I verified by *running*, plus the part of the fixture set the review could not have caught and that I do not think is fully covered.

## What I ran

**The defect, before the fix**, on `main` at `233246d1` — `adaptive_frozen_replay_tv` called directly against `ode_predictions_with_extra_breaks` on a lagged `zero_order(dur = 2, lag = 1.5)` two-dose subject:

```
t=2.0   replay 0.000000   static 24.385287749   exact 24.385287749
worst relative divergence over the sample set: 4.096e-1
```

The static value is the exact closed form `(50/0.1)·(1 − e^{−0.05})`, so the replay is the wrong side.

**Mutations on the shared break builder and on my own fix.** Each was applied to the source, rebuilt, and run:

| # | mutation | closed-form test | separated-edge test |
|---|---|---|---|
| 1 | restore the hand-rolled `d.time`-only loop in the replay | **FAIL `1.0e0`** | **FAIL `4.096e-1`** |
| 2 | delete `push_zero_order_break_times` from the shared `collect_dose_break_times` | **FAIL `2.5e-1`** | green (`0.0`) |
| 3 | delete `push_route_lag_break_times` from the shared `collect_dose_break_times` | green | green — 5 pre-existing #1174 tests fail instead (`route_lagged_first_order_onset_reaches_both_dense_builders`, the three `break_collision_1186::*`, and `sens::ode_provider::tests::compartment_lag_plus_per_route_lag_is_analytic`) |
| 4 | the *weaker fix*: keep the hand-rolled loop, add `push_zero_order_break_times` alone | green | **FAIL `1.311e-5`** |

**Row 2 is the one you asked about.** Deleting a push from inside the shared builder leaves the engine-vs-engine test at exactly `0.0` — both its sides run that helper and move together. That is why the pair exists and why the first test's oracle is a closed form rather than a second engine. Both tests say so in their doc comments rather than leaving it to be rediscovered.

**Row 4 is the #1174 trap, reproduced deliberately.** It is the edit a reviewer reaches for first — #1174's own note says a caller gets correct zero-order segmentation from `push_zero_order_break_times` alone — and it passes the entire suite on any #1174-shaped fixture, because a bare `zero_order(dur, lag)`'s route onset *is* its window start and one break serves both. Only the fixture that separates them (first-order onsets at 0.5 / 6.5 beside a window spanning 1.5–3.5 / 7.5–9.5) kills it. The test the issue itself suggested would have passed.

**The tolerances, measured not reasoned:**

| test | realised error | bound | headroom below the tightest mutation |
|---|---|---|---|
| `frozen_replay_delivers_the_exact_zero_order_mass` | `1.42e-16` | `1e-12` | `2.5e11×` (vs row 2's `2.5e-1`) |
| `frozen_replay_segments_a_route_lag_apart_from_a_zero_order_edge` | `0.0` (bit-identical) | `1e-14` | `1e9×` (vs row 4's `1.311e-5`) |

Full `cargo test --lib`: 4,730 passed, 0 failed. `tools/preflight.sh` and `tools/preflight.sh release-semantics` both green. All 259 pre-existing `adaptive` tests unchanged, including the byte-identical constant-path canaries.

## What the review did not catch, and I do not claim is covered

**The `pk_params_flat` argument is pinned by nothing.** I mutated it — swapped it for `init_pk.values` at the `collect_dose_break_times` call — rebuilt, and ran the whole library suite: **4,730 passed, 0 failed.** Completely invisible. Every fixture in the suite, mine included, has constant per-event PK (`vec![pk; n]`), so the frozen snapshot and every segment's snapshot are equal and the argument cannot be distinguished.

The argument is still the right one — it is what the driver places *its* base-regimen edges from at `predictions.rs:4506`, and bit-alignment with the driver is the property this engine exists to establish — but that is an argument, not a test, and my original comment read as a correctness claim. Pushed `9bc593da` to say so plainly at the call site.

Measuring that also surfaced an adjacent **latent inconsistency in both engines**, which I am reporting rather than fixing here: the break is placed from the frozen snapshot, while `integrate_segment` recomputes the window for *containment* from the **segment's** snapshot. If a route lag or `dur` ever becomes covariate-dependent on this path, the edge and the containment boundary move apart and mass is dropped — in the driver and the replay alike. Symmetric, so this verifier is blind to it by construction; latent, because the #930/#931 base-dose guard is the only route to it. It belongs to whichever change lifts that guard, alongside the widening question already in the PR's Open Questions.

**Two further limits worth stating, since a clean verdict shouldn't imply they're covered:**

- **Neither test exercises the reactive driver.** The property the verifier actually needs is *driver == replay*; both tests pin *replay == static engine*, a proxy. The real pairing is untestable for zero-order today, because the driver refuses such a subject upstream — which is the same latency that makes the whole defect unreachable.
- **Row 3 is deliberately left to #1174's tests** rather than duplicated here. Adding a redundant gate for it would be the "two redundant gates cover for each other" shape, which is its own test hole.

CI on the branch: Format, Rustdoc, Docs lint, Clippy, Check, Public API baseline, Tests (release semantics), Tests + coverage (TTE/CTMM), codecov/patch all pass. `slow-tests.yml` dispatched manually (`main`'s own scheduled run already fails on the unrelated `nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation`, so that is the baseline to compare against); result to follow in a comment.

roninsightrx:
## `slow-tests` result — [run 35248782413](https://github.com/FeRx-NLME/ferx-core/actions/runs/35248782413)

Dispatched manually on this branch (this touches `ode/predictions.rs`, which is on the estimation path, and slow-tests never run on a PR). **Overall: failure — one test, and it is the pre-existing `main` failure, not this PR.** Whole log read, both shards.

| shard | branch `9bc593da` | `main` (scheduled [35189123134](https://github.com/FeRx-NLME/ferx-core/actions/runs/35189123134), 11 h earlier) |
|---|---|---|
| Slow regression tests (core) | ✅ **0 failures** (45m44s) — `4923 passed`, `602 passed`, `132 passed` across its test runs | ✅ 0 failures |
| Slow regression tests (nn) | ❌ `5039 passed; 1 failed` | ❌ `5037 passed; 1 failed` |

The single failure is identical on both, down to the printed values:

```
nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation
  panicked at src/nn/mod.rs:3179:9
  weight norm must be non-increasing in λ (‖W‖²: 60.95117 → 0.00000 → 0.00116)
```

Bit-identical numbers across the two runs is the evidence that matters: this PR did not move it. It is already tracked as **#1277** ("two nn tests have failed on every slow-tests run since #1215 — identical values, masked by #1239's timeout").

The `5039` vs `5037` delta is exactly the two Tier-1 tests this PR adds, both green in the shard.

Nothing in `api::adaptive_sim_tests` regressed — the whole suite ran, including the base-regimen anchors this change is nearest to (`adaptive_base_infusion_matches_static_predict`, `adaptive_base_infusion_ending_between_records_matches_static_predict`, `adaptive_base_dose_f_under_tv_covariate_matches_closed_form`, `adaptive_base_dose_f_under_iov_matches_closed_form`, and the `auc_target` rejection set), all `ok`.

All PR checks are green on `9bc593da`.

Issue #1304, opened 2026-09-09
feat(diagnostics): predict() / simulate() open no SolverStatsScope, so their ODE solver diagnostics are dead


TeunP:
`SolverStatsScope::enter` is called in exactly two places in the whole tree —
`src/api/fit.rs:1912` and `:1939`. `simulate.rs`, `predict.rs`, `sim/adaptive.rs` and
`predict_survival` open none. So every ODE solver diagnostic is a `fit()`-only feature.

Three consequences, all measured while reviewing #1296:

1. **`W_ODE_SOLVER_DIAGNOSTICS` never fires outside `fit()`.** Clamped steps, discarded `auto`
   escalations, segments cut short by `ode_stiff_abort_after`, and now walks abandoned on an
   unorderable timeline are all invisible to `predict()` / `simulate()` / `predict_survival()`,
   which return `NaN` — or a freeze-padded trajectory — with nothing said. The `NaN` is the only
   signal on those paths.

2. **Three of the eight `abandoned_non_finite_timeline` recorders can never fire in
   production.** `ode_predictions_adaptive_impl`, `adaptive_frozen_replay_tv` and
   `ode_solve_until_chz_threshold` are reached only under `simulate()`, so they record into an
   inactive sink. They are exercised by tests that open a scope explicitly (#1296 added one per
   site); nothing in production can reach them.

3. **`ode_solve_until_chz_threshold` would still read zero even with a scope open.** It runs on
   `solver::solve_ode_until_threshold`, its own driver, which does not go through
   `integrate_resolved_g`'s `STATS_SINK` tee — measured: `attempted_steps = 0` on a solve that
   crossed the threshold normally. So on that engine the step counters cannot serve as the
   "did anything integrate" discriminator the way they do elsewhere.

**Suggested shape:** open a scope around the `simulate()` / `predict()` prediction passes and
route what it collects into the existing `ode_solver_diagnostics_warning` path; separately
decide whether the threshold driver should tee like the others. Sizing is the real design
question — the scope is thread-local, and one per replicate inside the uncertainty loop is not
free (#1288 measured the same class of per-draw cost at 3.4% of a `simulate` pass).

Found while reviewing #1296 (items 4 and 9). #1296 corrects the field documentation to say
which sites are live and states the `fit()`-only scope in `docs/model-file/ode-models.qmd`, but
does not widen the scope — that is a behaviour change with its own cost question.

roninsightrx:
[...]

Issue #1302, opened 2026-09-09
chore(ode): OdeSolverStats is not #[non_exhaustive], so every new counter is a breaking change


TeunP:
`src/ode/solver.rs` declares

```rust
#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
pub struct OdeSolverStats { /* every field pub */ }
```

with no `#[non_exhaustive]`. Every field is `pub`, so a downstream crate can build one with a
struct literal or destructure it exhaustively — and either breaks the moment a counter is added.
Two counters have been added exactly that way (`auto_stiff_rejected_jets` for #1204,
`abandoned_non_finite_timeline` for #1234), each described in its PR as "additive", which is a
semver claim the struct's attributes do not support.

Nothing in this repo or in `ferx-r` constructs or destructures it today — checked on
`origin/main` for both — so the fix is free right now and gets more expensive with every
consumer that appears.

**What to do:** add `#[non_exhaustive]`, regenerate `api/ferx-core-public-api.txt`, and note in
`CHANGELOG.md` that the struct is built from `Default::default()` and matched with `..`. Adding
the attribute is *itself* a breaking change for any existing literal, which is the reason to do
it while the answer to "who has one" is still nobody.

Found while reviewing #1296 (item 10). Not fixed there: it is a public-API change unrelated to
that PR's subject, and it wants its own baseline diff and CHANGELOG line.

PR #1410, opened 2026-09-17
feat(api): carry the model/data and ODE-solver diagnostics on predict() and simulate() [closes #1280] [closes #1304]

roninsightrx:
## Why

Every model/data finding ferx computes reached exactly two entry points — `fit()` and `ferx check` — and every ODE-solver diagnostic reached one, `fit()`. So a model `fit()` refuses to stay quiet about was silently fine through every other door.

**Reproduced on `main` at `233246d1` (2026-09-17), not read off the issues.** One fixture, both halves:

| call | rows | warnings |
|---|---|---|
| `fit()` | OFV `2e20` | 8, incl. `W_STEADY_STATE_ABSOLUTE_TIME`, `W_ODE_SOLVER_DIAGNOSTICS` (14016 clamped steps) |
| `predict()` | `NaN`, `NaN` | *no channel at all* |
| `simulate_with_options_diag()` | 2 rows | `[]` |

(the #1139 model: an `SS=1, II=12` dose on an `[odes]` RHS reading `TAFD`.)

And the half that is worse, because it is plottable — a stiff two-state model at `ode_max_steps = 500`:

```
predict()  id=1 t=0.5  pred=49.98522138377198
           id=1 t=2    pred=49.98522138377198
           id=1 t=8    pred=49.98522138377198
           id=1 t=24   pred=49.98522138377198
simulate_with_options_diag()  rows=8  warnings=[]
```

Four identical numbers to 17 figures on a decaying curve. **That is #959's own signature, on the path #959 was reported against** — and #1350 row 20 is right that #959's reproducer is correct today under `ode_method = auto`, so this reach gap is exactly what is left of it.

The two issues are one object. #1278 made the absolute-clock finding a *warning* rather than an error on the stated grounds that "an error would protect `fit()` only … a warning surfaced by `fit()` *and* `ferx check` beats an error one entry point enforces". The premise is right; the warning inherited the same reach. This discharges it.

## What changed

**One implementation, every entry point that has a channel.** `postfit::non_fit_diagnostics(model, population, params, stats, phase)` returns the `check_model_data_warnings` bundle followed by the pass's `W_ODE_SOLVER_DIAGNOSTICS`. Three callers:

- **`predict_diag()`** — new; returns `PredictionOutput { results, warnings }`. `predict()` becomes the thin wrapper that discards, mirroring `simulate_with_options` / `_diag`. Rows are **bit-identical** (asserted on `to_bits()`).
- **`simulate_with_options_diag()`** — both the ordinary and the propensity-matching return.
- **`simulate_adaptive()`** — `AdaptiveSimulationResult` is `#[non_exhaustive]`, so a `warnings` field is additive. This is the path where a silent solver diagnostic matters most: the controller reads simulated state to pick the next dose, so a freeze-padded segment feeds the wrong signal into the next decision and the ledger inherits it.

**What the bundle contains, and what it does not** (corrected in review round 1 — the first version carried only `check_model_data_warnings` while claiming "the same list `fit()` reports"). The rule is *a finding about the model or the data is carried; a finding about the fit's configuration or the optimizer's start is not*, because on these entry points there is no fit and no optimizer:

| | source | in the bundle |
|---|---|---|
| 1 | `CompiledModel::parse_warnings` | yes |
| 2 | `Population::warnings`, through the same `reader_warning_suppressed` filter `fit()` and `ferx check` use | yes |
| 3 | `check_model_data_warnings` | yes |
| 4 | `check_experimental_features` | yes |
| 5 | the pass's ODE-solver diagnostics | yes |
| — | `check_model_options` warning half | **no** — takes a `&FitOptions` these entry points do not have; its subject is an estimator that is not running |
| — | `check_packed_start_in_box` warning half | **no** — its subject is the vector the *outer optimizer* starts from |
| — | `fit()`'s operational notes (FD-fallback count, thread hint, covariance cost) | **no** — they describe a run that is not happening |

Within what it carries the list is **unfiltered**. Filtering one code out for one caller is what #1280 explicitly rules out, and two members are phrased for a fit (`W_ADDITIVE_INIT_SCALE` talks about optimizer basins) but are still true of the numbers being served.

**Phase, because the post-fit wording is false elsewhere.** `ode_solver_diagnostics_warning` takes a `SolverStatsPhase`. A `predict()` message says *"at the supplied parameters … Counters are from this predict() pass over all subjects"*, not *"at the final estimates … from the post-fit prediction pass"* — a `predict()` caller runs at their own parameters and may never have fitted anything. `simulate` says "and replicates", since it integrates `n_sim` copies of each subject. A discarded stiff escalation says *this pass* paid for both solves rather than "the fit" (`phase.payer()`), and the **informational** escalation note carries its provenance too — both corrected in review round 1, where each defect had been sitting in the clause the other round-1 test did not look at. `details.phase` carries the phase as well.

**The `ode_method` is the model's own.** `solver_reporting_options(model)` reads the settings stamped on the `OdeSpec` (or its `AbsorptionOdeEquivalent` twin, #814) rather than `FitOptions::default()`. Without it a user running `ode_method = rodas5p` would be told `auto` had misbehaved. Only two `FitOptions` fields are read back out, and that is *asserted against the source*, not assumed.

**#1217 row f, re-measured.** The issue's two sites (`api/fit.rs:1912`, `:1939`) have drifted into `postfit.rs` (`compute_subject_results:623`, `sweep_sensitivity_solver_stats:2216`); substance unchanged, still `fit()`-only. Production scope sites are now four and pinned at exact counts per file.

**#1304 item 2 closed.** Opening a scope on `simulate()` / `simulate_adaptive()` makes all three "can never fire in production" recorders reachable — the adaptive driver, its frozen replay, and the CHZ event-time solve. `ode/solver.rs`'s field doc asserted the opposite; corrected in this PR rather than left to be rediscovered.

**#1304 item 3 left open, and now documented where it bites.** `ode_solve_until_chz_threshold` runs on `solve_ode_until_threshold`, which never tees into `STATS_SINK`, so its *step* counters still read 0 with a scope open. Teeing it means threading an `OdeSolverStats` through a driver that decides simulated TTE event times; that is its own change with its own oracle, and the issue itself frames it as "separately decide". The `abandoned_non_finite_timeline` recorder at that site is unaffected — it records directly, not through the driver.

**`--simulate` no longer says everything twice.** `run_model_simulate` simulates a template and then fits it, appending the simulation's warnings to the fit's — so since this PR every model/data finding was produced twice, in identical wording. `api::run` now appends without exact duplicates. Deliberately an *exact string* match, not a code match: the two `W_ODE_SOLVER_DIAGNOSTICS` messages describe different passes at different parameters and say so.

## Alternatives considered

- **An eleventh `assert_*` wrapper** (the shape `predict()` uses for the *error*-severity checks). Rejected: #1217 row 7's 2026-09-08 correction says it directly — the errors already reach both entry points, as process aborts, and an eleventh assert is not how a warning-severity finding gets there. That is #898, not this.
- **Changing `predict()`'s return type.** Breaking, and ferx-r binds the signature. The `_diag` twin is additive and matches the pattern `simulate` already established.
- **Giving `simulate_with_uncertainty()` a channel.** It returns a flat row vec and already drops its own per-draw findings (#786). Adding `_diag` there is a third public entry point for a loop #1288 is separately about; left out, documented, and pinned by a test rather than left to be rediscovered.
- **`predict_survival` / `predict_categorical`.** Same reasoning; both listed in the docs table as still silent.

## Cross-repo dependency

Nothing here is breaking, so ferx-r keeps building. A follow-up there could expose `predict_diag()` and `AdaptiveSimulationResult$warnings` (needing a `Cargo.lock` bump via `tools/update-ferx-core-lock.sh`), but nothing forces it.

## Breaking changes
- [x] Public Rust API — regenerated; **purely additive** (10 lines). `predict_diag`, `PredictionOutput` (`#[non_exhaustive]` from the start, cf. #1302), and `AdaptiveSimulationResult::warnings` (the struct is already `#[non_exhaustive]`). Each is needed by the fix itself: `predict()` had no diagnostics channel and could not gain one without a new return type.
- [ ] None

## Numerical validation

No estimation, PK or statistical quantity changes. The one numerical claim is that **nothing moves**, asserted bit-for-bit rather than to a tolerance: `predict_returns_exactly_the_diag_forms_rows` compares `predict()` against `predict_diag().results` on `to_bits()` for every row, on a fixture checked to be finite (a `NaN` fixture would let two broken paths agree).

- [ ] Warfarin example converges…
- [x] Not applicable (this reports existing findings through additional channels; it computes no new number)

Per CLAUDE.md's rule, this is not an anchor-skipping exception — there is simply no new quantity. The counters and messages this surfaces are the ones `fit()` already reported and `api/tests/ode_solver_diagnostics_tests.rs` already pins.

## Performance

Measured, because #1304 names sizing as the real design question and #1288 measured the same class of per-draw cost at 3.4%. **The scope is opened once per call, not once per replicate.** Release build, `RAYON_NUM_THREADS=2`, 100 subjects × 100 observations × 10 replicates on a 2-compartment `[odes]` model, best of 5:

| | with scope | without | Δ |
|---|---|---|---|
| `simulate_with_options_diag` | 56.3 ms | 55.5 ms | **+1.4 %** |
| `predict_diag` | 1.2 ms | 1.6 ms | below the noise floor (the scope-on arm was *faster*) |

The mechanism is why it is cheap: the `STATS_SINK` tee is one `Cell` read/write **per segment**, not per step — `integrate_resolved_g` threads a plain `&mut OdeSolverStats` into the inner driver and merges once at the end. A closed-form model opens no scope at all (`integrates_odes` gate, straddle-tested).

## Tests

- [x] Unit test(s) added in the same module — 21 Tier-1 in `src/api/tests/non_fit_diagnostics_tests.rs` + 1 in `adaptive_sim_tests.rs`; whole new file runs in 0.13 s
- [ ] No new tests needed

**Mutation results — 23 run, 23 killed** (14 in round 0, 9 more after review round 1), with per-site kill sets. Round 0:

| mutation | killed |
|---|---|
| drop `check_model_data_warnings` from the shared bundle | 6 tests |
| drop the solver warning from the shared bundle | 6 |
| `solver_stats_scope` always `None` | 6 |
| `solver_stats_scope` drops the `integrates_odes` gate | `the_scope_is_opened_only_for_a_model_that_integrates_something` |
| predict: no per-subject scope | 4, incl. `predicts_solver_counters_cover_every_subject…` |
| **predict: one hoisted scope instead** (the reviewer's alternative) | the same 4 |
| `solver_reporting_options` → `FitOptions::default()` | `the_reported_ode_method_is_the_models_own…` |
| simulate: drop the scope on the non-propensity branch only | 2; the predict tests stay green |
| `simulate_adaptive` drops its scope | the adaptive test only |
| `at_label` hard-coded to the post-fit wording | 2 |
| `provenance` hard-coded | 1 |
| propensity branch drops the bundle | `the_propensity_branch_of_simulate_carries_the_bundle_too` only |
| `predict()` wrapper reorders rows | 17, incl. `predict_returns_exactly_the_diag_forms_rows` |
| `api::run` appends unfiltered | `a_simulate_run_reports_each_model_data_finding_exactly_once`, `left: 2, right: 1` |

Round 2, after review round 1 ([reply](https://github.com/FeRx-NLME/ferx-core/pull/1410#issuecomment-5719525841)):

| mutation | killed |
|---|---|
| bundle drops `parse_warnings` | 2 |
| bundle drops the reader warnings | 1 |
| bundle drops `check_model_data_warnings` | 8 |
| bundle drops `check_experimental_features` | 1 |
| reader-warning **suppression filter** removed | `reader_warnings_go_through_the_same_suppression_filter_fit_uses` |
| info note loses its provenance sentence | `the_informational_note_and_the_rejection_clause_name_their_pass_too` |
| `payer` hard-coded back to "the fit" | the same |
| adaptive returns solver diagnostics without the model/data half | 2 |
| propensity branch discards its collected solver stats | `every_diagnostic_carrying_entry_point_reports_the_same_bundle` |

The last two are the reviewer's own named escape edits, and both were **green against the whole suite** before round 2. The suppression-filter one is a hole the round-2 *fix* opened and the round-2 tests then closed: an unfiltered `Population::warnings` extend tells every compartment-free `$PRED`-style model that its data are malformed for having no doses, and it left all 20 tests green.

Three things this PR did **not** take on trust:

1. **Each side of the predict/simulate twin was mutated separately.** They open their scopes by different mechanisms (per-subject inside a `par_iter` vs one around a serial loop), so a single test covering both would have hidden a one-sided regression. Dropping simulate's scope leaves the predict test green, and vice versa.
2. **The "different minimal edit" check.** The plausible wrong implementation is not "delete the scope", it is "open one scope around the whole parallel pass" — which compiles, reads cleaner, and silently reports only the calling thread's subjects. `predicts_solver_counters_cover_every_subject_not_just_the_calling_threads` exists for it and asserts a **ratio** (1 subject → n padded segments, 16 identical subjects → exactly 16n), not an absolute count.
3. **Two fixtures were found degenerate and fixed before they shipped.** The straddle arm of `predict_reports_a_freeze_padded_segment…` initially ran the *clean* model at the *starved* model's `KFAST = 1e5`, so it reported two padded segments of its own — a straddle that did not straddle. And the `--simulate` duplication test first used `W_ADDITIVE_INIT_SCALE`, which reads `|DV|` and is therefore inert on the `DV = .` template: the unfiltered-append mutation stayed **green** on it. Swapping to `W_NEGATIVE_LAGTIME` (model-only, produced identically by both halves) made it fail with `left: 2, right: 1`.

Two structural guards, both asserting against the source rather than against behaviour:

- `the_solver_warning_reads_only_the_two_options_this_view_carries` — a third `options.<field>` read in the warning builder would be served from `FitOptions::default()` on every non-fit entry point, silently, since the message stays well-formed. (This one caught its own first version: the literal `"options."` missed a read rustfmt had split across a line break.)
- `the_production_scope_sites_are_the_phases_this_enum_lists` — exact `SolverStatsScope::enter` counts per file, so a fifth site needs a phase and a channel, or it is #1304 again.

A fourth, found in round 2: **one of this PR's own tests broke under the round-2 fix and was repaired rather than relaxed.** `ODE_NO_IIV` declares an unreferenced `ETA_CL` deliberately (the bit-exact oracles in `adaptive_sim_tests.rs` rely on it), so it carries a parse warning and the adaptive "clean" arm stopped being `is_empty()`. Caught by chasing why a mutation in `simulate.rs` killed a test in `adaptive_sim_tests.rs` instead of accepting the kill. The assertion is now scoped to the solver half **plus** an exact count and content check on what remains, so it stays a straddle.

Full `cargo test --features ci` green at `25973768` — **195 binaries, 5581 passed, 0 failed, 187 ignored**. `tools/preflight.sh` and `tools/preflight.sh release-semantics` both green (the latter 5101 lib tests).

- [x] Inline below

```rust
// Before: the NaN column, unexplained.
let rows = ferx_core::predict(&model, &population, &params);

// After: the same rows, plus what fit() would have said about them.
let out = ferx_core::predict_diag(&model, &population, &params);
for w in &out.warnings { eprintln!("{w}"); }
```

On the `TAFD`-under-`SS=1` model above, `out.warnings` is now:

```
1 subject(s) have an SS=1 dose on a model whose `[odes]` right-hand side reads an absolute
clock (`TAFD`). The run-in that stands in for the infinite past expands the periodic dose
train on a clock local to each cycle …

W_ODE_SOLVER_DIAGNOSTICS: the ODE solver did not integrate cleanly at the supplied parameters
(ode_method = auto): 3520 step(s) clamped at the minimum step size … Counters are from this
predict() pass over all subjects.
```

and on the freeze-pad fixture:

```
W_ODE_SOLVER_DIAGNOSTICS: the ODE solver did not integrate cleanly at the supplied parameters
(ode_method = rk45): 2 returned segment(s) stopped before their requested end time and
freeze-padded the remaining output times with the last state — they exhausted ode_max_steps …
```

## Docs
- [x] `docs/warnings.qmd` — new *Which entry points report warnings* section (`#entry-points`) with the full table of what carries diagnostics and what does not, plus a *The non-fit bundle* subsection (`#non-fit-bundle`) enumerating the five sources it carries and the three it deliberately excludes, with the reason for each. `cargo run -p docs-lint -- --check` green.
- [ ] New pages linked in `docs/_quarto.yml` (no new page)
- [x] `CHANGELOG.md` `[Unreleased]` entry added under `Added`
- [ ] No user-visible change

## Checklist
- [x] `tools/preflight.sh` green — fmt, the 5 `cargo check` feature sets, clippy `--all-targets`, rustdoc, docs, public-API baseline
- [x] `tools/preflight.sh release-semantics` green (#1326: the bare script omits it)
- [x] Functions on the `Dual2` sensitivity path stay differentiable — nothing here touches `sens/` or any `*_g<T: PkNum>`; `sweep_sensitivity_solver_stats` is unchanged apart from the phase argument at its consumer

### ferx-site
- [ ] New `.ferx` DSL syntax or `[fit_options]` key (none — no new key)
- [ ] New example `.ferx` file (none)
- [ ] New data file (none)

### ferx-book
- [ ] New estimator, DSL feature, or fit option (none)

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` file, `[fit_options]` key or DSL surface changes, and `predict()` / `simulate()` rows are bit-identical

## Reviewer hints

**Where to focus.**

1. `postfit::non_fit_diagnostics` and `SolverStatsPhase` — the whole design is there. Everything else is three call sites.
2. **The parallelism in `predict_diag`.** `SolverStatsScope` is thread-local, so the scope has to be per subject task and merged in subject order. This is the one place a plausible, cleaner-reading implementation is silently wrong, and `predicts_solver_counters_cover_every_subject_not_just_the_calling_threads` is the test standing in front of it. `simulate` and `simulate_adaptive` are serial (their RNG draw order is contractual), so one scope each is correct there — the asymmetry is recorded at both call sites.
3. **The exact-string dedupe in `api::run`.** Ask whether a code-level dedupe would be better; I argue not, in the comment: the two `W_ODE_SOLVER_DIAGNOSTICS` messages there describe genuinely different passes.

**What can be skimmed.** The 20 mechanical `SolverStatsPhase::PostfitPredictions` additions in `ode_solver_diagnostics_tests.rs`, and the `api/ferx-core-public-api.txt` diff (10 additive lines).

**How this interacts with #898** (row 28 — `predict()` / `simulate()` / `suggest_start()` panic on preconditions; **not** this PR, deliberately not folded in):

They are the two ends of one design decision, as #1217 row 7's correction says, and this PR is built so that #898 lands *through* it rather than against it:

- It adds **no** new `assert_*`. The ten existing panic wrappers in `predict_diag` are byte-for-byte the ones `predict()` had; the diagnostics arrive on the `warnings` channel instead, which is the distinction that correction draws.
- `PredictionOutput` is the natural landing site for #898. When the wrappers become `Result`, `predict_diag` becomes `Result<PredictionOutput, String>` — a one-line signature change at a function that is one release old — and `predict()` keeps whatever compatibility shim is chosen. Without this PR, #898 would have had to invent that return type anyway.
- `#[non_exhaustive]` on `PredictionOutput` means #898 can add a structured-error or `diagnostics: CheckReport` field additively.
- It makes #898 *less* urgent without preempting it: on the `TAFD` model that motivated this, the user now gets a named finding instead of only a `NaN` column, so the remaining gap is the genuinely fatal preconditions, which is what #898 is about.

The one thing #898 must not do is route the warning-severity bundle through the new `Result`: an `Err` for `W_STEADY_STATE_II` would be the error-for-everything outcome #1278 argued against.

## Open questions

1. **Should `simulate_with_uncertainty()` get a `_diag` twin?** It has findings and nowhere to put them (#786's skipped draws are dropped at the return today, before this PR). I left it out to keep this PR to one design, and the docs + `the_row_only_entry_points_stay_silent_by_contract` name it as the remaining silent entry point. Happy to add it here if preferred; note #1288 wants to restructure that same loop.
2. **Is +1.4 % on `simulate` acceptable unconditionally**, or should the scope be opt-in via `SimulateOptions`? I think unconditional: it is a tenth of #1288's measured 3.4 %, and an opt-in diagnostic that defaults off is the dead-diagnostic shape #1080 item 3 existed to remove. But it is a knob someone might want on a large VPC.
3. **`ode_solve_until_chz_threshold`'s step counters** (#1304 item 3) — I propose keeping #1304 open for it rather than closing on two of three items. Say the word and I will file it as its own row instead.


roninsightrx:
## Codex review

_Automated review by `codex exec` (read-only: no build or test run) against `origin/main`. Findings are the reviewer's claims; each is being verified by mutation before any change is made._

The numerical prediction paths appear unchanged, and the public API baseline, docs, and changelog were updated. However, I recommend changes before merge: the advertised warning contract is broader than the implementation, phase-specific solver messages still contain incorrect wording, and two independent diagnostic branches lack regression coverage. No compilation or tests were run, per the review constraint.

### Blocking findings

- `src/api/postfit.rs:2102` — **medium** — `non_fit_diagnostics()` only includes `check_model_data_warnings`, while the docs and changelog claim it carries the same unfiltered model/data list as `ferx check` and `fit()`. Those surfaces also include parse warnings, unsuppressed `Population::warnings`, option warnings, packed-start warnings, and experimental-feature warnings. For example, a reader-produced `W_ADDL_MISSING_II` remains absent from `predict_diag()` despite the documented “same list” contract. — **Suggested fix:** either aggregate every applicable warning source through a shared helper, or narrow `docs/warnings.qmd`, the changelog, and API docs to state precisely that only `check_model_data_warnings()` findings are included and enumerate the exclusions.

- `src/api/postfit.rs:2391` — **medium** — The phase-aware message conversion is incomplete. The informational escalation branch omits `phase.provenance()`, contrary to the documented promise that it names the producing pass, while the rejected-escalation branch at line 2442 still says “the fit paid for both solves” when called by `predict_diag()`, simulation, or adaptive simulation. The current phase test only checks provenance on the unclean branch and only checks “supplied parameters” on the informational branch, so both defects pass. — **Suggested fix:** include the phase provenance in the informational message, replace the fit-specific rejected-escalation wording with phase-neutral or phase-derived wording, and test `Predict`, `Simulate`, and `SimulateAdaptive` with `auto_stiff_rejected > 0` and escalation-only statistics.

- `src/api/tests/non_fit_diagnostics_tests.rs:382` — **medium** — The tests do not cover every independent diagnostic path they claim to cover. `every_diagnostic_carrying_entry_point_reports_the_same_bundle` compares only prediction and ordinary simulation; adaptive simulation is tested only for solver diagnostics. Separately, the propensity branch is tested only for the model/data half using an analytic model. The smallest edits that escape the suite are: return solver diagnostics without model/data findings from adaptive simulation, or discard collected solver statistics only in the propensity branch. — **Suggested fix:** include adaptive simulation in the shared-bundle comparison using a model/data warning, and add a propensity-matched, budget-starved ODE case that asserts `W_ODE_SOLVER_DIAGNOSTICS`.

### Nits

No additional nits.

roninsightrx:
## Review round 1 — all three **CONFIRMED**, fixed in `25973768`

Each was verified by mutation or by measurement *before* anything was changed. Nothing was adjusted to appease a claim; one of the three turned out to be worse than reported and one turned up a fourth hole.

---

### 1. `non_fit_diagnostics` carried only `check_model_data_warnings` — **CONFIRMED**

Verified by reading what `fit()` actually seeds, not by reasoning about it. `api/fit.rs:1254-1267`:

```rust
let mut accumulated_warnings: Vec<String> = model.parse_warnings.clone();
accumulated_warnings.extend(pre_run_warnings);
accumulated_warnings.extend(
    population.warnings.iter()
        .filter(|w| !crate::api::validation::reader_warning_suppressed(model, w))
        .cloned(),
);
```

…then `check_model_data_warnings` (`:1301`), the option warnings (`:1313`), the packed-start rails (`:1318`) and `check_experimental_features` (`:1322`). The bundle carried exactly one of those. The documented contract ("the same bundle `ferx check` prints and `fit()` reports") was an overclaim, and `W_ADDL_MISSING_II` really was absent from `predict_diag()`.

**Fixed by aggregating, and by narrowing — both, because the two halves are different questions.** `non_fit_diagnostics` now carries, in `fit()`'s own order:

1. `model.parse_warnings`
2. `Population::warnings`, through the same `reader_warning_suppressed` filter
3. `check_model_data_warnings`
4. `check_experimental_features`
5. the pass's solver diagnostics

Two sources stay out, and the reason is structural rather than editorial — it is now stated at the function, in `docs/warnings.qmd#non-fit-bundle` and in the changelog, instead of implied:

- **`check_model_options`' warning half.** It takes a `&FitOptions` these entry points do not have. Synthesizing one would report findings about *defaults the caller never chose* — the exact trap `solver_reporting_options` exists to avoid — and its subject (`W_GN_NO_RANDOM_EFFECTS`, `W_VI_OMEGA_UNANCHORED`) is an estimator that is not running.
- **`check_packed_start_in_box`' warning half.** Its subject is the vector the **outer optimizer** starts from. Nothing is packed on a prediction or a simulation.

The rule is now written down: *a finding about the model or the data is carried; a finding about the fit's configuration or the optimizer's start is not.* Pinned from both sides by `the_bundle_carries_every_model_and_data_source_fit_carries` (one arm per source, each with a control) and `the_bundle_excludes_the_findings_that_are_about_a_fit` (a straddle: the same model raises the option warning through `check_model_options` and not through `predict_diag`).

**This finding was worse than reported, in a way the fix surfaced.** Adding source (2) without its filter is the obvious reading of "aggregate every source", and it is wrong: `reader_warning_suppressed` is `model.is_algebraic() && warning.starts_with("W_NO_DOSES")`, so an unfiltered extend tells every compartment-free `$PRED`-style model that its data are malformed for having no doses — on the entry point such a model is most likely to be used from. **Measured: removing the filter left all 20 tests green.** `reader_warnings_go_through_the_same_suppression_filter_fit_uses` now straddles that predicate (suppressed on an algebraic model, still reported on a compartmental one).

---

### 2. The informational branch omits `provenance()`, and the rejection clause says "the fit paid" — **CONFIRMED, both halves**

Measured, not read. Current-code output before the fix:

```
W_ODE_SOLVER_ESCALATION_NOTE: ode_method = auto escalated 12 integration segment(s) to a
stiff stepper at the supplied parameters; every other segment used rk45, no escalation was
rejected, and no step clamped at the minimum step size. Informational — set ode_method = rk45 …
```

— `SolverStatsPhase::Predict`, and **no provenance sentence at all**. And from `Simulate`:

```
… wrong that the stiff method it picked could integrate them, and the fit paid for both
solves (the 0 step(s) those attempts clamped …
```

Your diagnosis of *why* both passed is exactly right and is the part worth recording: `the_message_names_the_pass_that_produced_the_counters` asserted provenance only on the unclean branch, and `the_abandoned_clause_and_the_escalation_note_carry_the_phase` asserted `at the supplied parameters` only on the informational one. **Each defect sat in the clause the other test did not look at** — a fifth instance of the CLAUDE.md "a green test is not evidence that it can fail" family, and one I should have caught by asking which clause each assertion reaches.

Fixed: the info note carries `{provenance}`, and the rejection clause names `phase.payer()` (`"the fit"` / `"this predict() pass"` / `"this simulate() pass"` / `"this simulate_adaptive() run"`). Split into its own method rather than reworded into a passive ("both solves were paid for") because *who* absorbed the cost is the actionable part — on a `simulate()` it is this call, repeated per replicate.

`the_informational_note_and_the_rejection_clause_name_their_pass_too` asserts **every** non-fit phase, not one: the three share `at_label` but have distinct `provenance` and `payer` strings, so a phase that lost its own wording would otherwise be covered by its neighbours. It also straddles — the post-fit phase must keep every word it had.

Two things I checked while there and did **not** change, with the reasoning, in case you disagree:

- The `rejected_jets` clause says "This count comes from the sensitivity sweep". `auto_stiff_rejected_jets` is only ever set by a `Dual1`/`Dual2` solve, and `predict` / `simulate` / `simulate_adaptive` are all `f64` passes, so on a non-fit phase that counter is structurally zero and the clause is unreachable. The sentence is also true whichever phase reads it.
- `solver_knob_advice` ends "…or checking the parameter estimates that produce these dynamics". A non-fit caller did supply parameters, so this reads correctly from every phase.

---

### 3. The shared-bundle test compared only two of four call paths — **CONFIRMED by running both escape edits**

Both named edits were **GREEN against the whole suite** before this:

| escape edit | before | after |
|---|---|---|
| return solver diagnostics without the model/data half from `simulate_adaptive` | GREEN | RED — `simulate_adaptive_reports_both_halves_of_the_bundle` |
| discard the collected solver stats in the propensity branch only | GREEN | RED — `every_diagnostic_carrying_entry_point_reports_the_same_bundle` |

`every_diagnostic_carrying_entry_point_reports_the_same_bundle` now compares `predict` against **both** `simulate` branches on the both-halves fixture, and `simulate_adaptive_reports_both_halves_of_the_bundle` is a new test with its own fixture (the adaptive engine needs decision times and a controller, so it cannot share the other one) asserting the model/data half *and* the solver half.

---

### One of this PR's own tests broke under fix (1), and is fixed rather than relaxed

`ODE_NO_IIV` declares `ETA_CL` and never references it — deliberately, because the bit-exact oracles in `adaptive_sim_tests.rs` rely on the trajectory being η-invariant. That is a parse warning, and parse warnings are now in the bundle, so the adaptive test's `clean.warnings.is_empty()` straddle stopped holding. I caught it by noticing that mutation E2 (in `simulate.rs`) killed a test in `adaptive_sim_tests` and chasing why instead of accepting the kill.

The assertion is now scoped to the property under test (no `W_ODE_SOLVER_DIAGNOSTICS`) **plus an exact count and content check on what remains**, so it stays a straddle instead of quietly widening into "anything goes".

---

### Verification

**Round-2 mutations: 9 run, 9 killed** (on top of round 1's 14/14):

| mutation | killed |
|---|---|
| bundle drops `parse_warnings` | 2 |
| bundle drops the reader warnings | 1 |
| bundle drops `check_model_data_warnings` | 8 |
| bundle drops `check_experimental_features` | 1 |
| reader-warning suppression filter removed | `reader_warnings_go_through_the_same_suppression_filter_fit_uses` |
| info note loses its provenance sentence | `the_informational_note_and_the_rejection_clause_name_their_pass_too` |
| `payer` hard-coded back to "the fit" | the same |
| E1 — adaptive drops the model/data half | 2 |
| E2 — propensity branch discards its solver stats | `every_diagnostic_carrying_entry_point_reports_the_same_bundle` |

21 Tier-1 tests in `non_fit_diagnostics_tests.rs` (0.13 s) + 1 in `adaptive_sim_tests.rs` + 1 Tier-2. Full `cargo test --features ci`: **195 binaries, 5581 passed, 0 failed**. `tools/preflight.sh` and `tools/preflight.sh release-semantics` green. `cargo run -p docs-lint -- --check` green.

The PR description's "What changed", "Tests" and "Docs" sections are updated to match.

roninsightrx:
## `slow-tests` dispatched on the branch — core green, nn red on a known `main` failure

Dispatched by hand ([run 35256139425](https://github.com/FeRx-NLME/ferx-core/actions/runs/35256139425)) because `slow-tests` never runs on a PR (#1271), and this PR restructures how `predict()` assembles its per-subject results — a path many convergence tests go through, even though the rows are asserted bit-identical.

| leg | result |
|---|---|
| **Slow regression tests (core)** | ✓ green, 45m01s |
| **Slow regression tests (nn)** | ✗ `5054 passed; 1 failed; 15 ignored` |

The one failure is **#1277 / #1350 row 0**, not this PR:

```
---- nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation stdout ----
weight ‖W‖²: λ=0 60.95117, λ=5 0.00000, λ=100 0.00116
NL modulator var: λ=0 12.889398, λ=5 0.000000, λ=100 0.000000
panicked at src/nn/mod.rs:3179:9:
weight norm must be non-increasing in λ (‖W‖²: 60.95117 → 0.00000 → 0.00116)
```

Those numbers are **byte-identical** to the ones #1350 row 0 records for the push run at `4eb546f6` (#1399) and the nightly at `233246d1` — the same test, the same three values to five decimals. It has been red on `main` since `4eb546f6` and is owned by @hiddevandebeek on row 0. This PR touches no `nn` code, no estimation math and no `sens/` kernel; its only reach into the fit path is one `SolverStatsPhase` argument at `fit_inner`'s existing `ode_solver_diagnostics_warning` call.

Worth noting on the other side: the **core** leg is green here, where `main`'s own runs were red on `frem_covariate_omega_matches_sample_variance_under_focei` until #1386 landed. So this branch's slow-tests are a strict improvement on the last recorded `main` state apart from the inherited nn failure.

PR CI at `25973768`: Check, Clippy, Format, Rustdoc, Docs lint, Public API baseline, Tests (release semantics), Tests + coverage (TTE/CTMM), codecov/patch all green; `Tests + coverage (core)` still running at the time of writing.

PR #1411, opened 2026-09-17
fix(estimation): resume an L-BFGS line-search abort from the best point instead of reporting it [closes #1277]

hiddevandebeek:
## Why

Nightly `Slow regression tests (nn)` has been red since `4eb546f6` (#1399) on `nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation` — #1350 row 0, #1277 reopened. The tracker's hypothesis was that #1399's NaN-preserving `predict_concentration` changed where the λ = 100 optimizer stops. That is half right and not the defect.

**Measured** (local repro is byte-identical to CI, 73 s): at the fixture's initial θ the inner loop tries η values hundreds of SDs out (CL = 6e-65, KA = 2e290), and the closed form returns NaN there. At `3e2c8e4d` the λ = 0 fit floored **621,010** such NaNs to 0.0 and passed on them; at `4eb546f6` they are repelled (correctly), which moves the objective at the initial point by 120 units. That is the trigger. The defect it exposed is in how ferx handles NLopt L-BFGS aborts, and it was already there — with `plis.c` instrumented, the pre-#1399 run's first line search *also* reached Luksan's ten-extrapolation cap and merely happened to satisfy the acceptance test on the tenth (`nred=-10, iters=4`); post-#1399 it needed one more and got `iters=-1`.

Luksan's `plis` returns a bare `NLOPT_FAILURE` — discarding every point it accepted — when one line search exhausts its ten reductions or ten extrapolations on an iteration where it has just restarted, which iteration 1 always is. Neither cap is exposed. On this fixture:

| fit | what happened | reported |
|---|---|---|
| λ = 0 | overshoot cap shrinks the opening gradient by 1.7e5 → the first line search extrapolates through **eleven improving evals** (18759 → 7975) → tenth extrapolation on eval 12 → `Failure`, no further evaluation | OFV 7902, `converged = false`, ~9000 above optimum |
| λ = 100 | subject 22 switches EBE mode at eval 154 → curvature pair `s·y ≈ 1e-5` → next direction ~1e17 long → ten reductions cannot find a decrease → `Failure` at gradient norm **17.7**. The probes pad the flat tail to 11, so the trace alone reads as a plateau; the cold re-solve does not (−99 vs best −113) | OFV −113, `converged = false` |

Any model with a large initial gradient can hit the first shape; the fixture just sits on the knife-edge.

## What changed

A bare `Failure` that the existing plateau verdict (`failure_is_converged_plateau`, #751) refuses to call converged is now **resumed from the best-seen point** on the remaining `outer_maxiter` budget — fresh identity Hessian, overshoot cap re-armed through a per-launch gradient count (`NloptState::{launch, grad_evals_at_launch}`). `resume_descent` is the pure decision:

- L-BFGS only (the mechanism is Luksan's line search; SLSQP is capped every eval, BOBYQA/MMA take no line search).
- Progress must have landed *in the launch that failed* — the eval-1 stall stays with the held-cap retry, and a resume that made no progress is not repeated.
- The plateau verdict is taken **lazily**: its cold inner solve runs only past the cheap gates and is handed to the final inner loop (`cold_solve_cache`) so a fit that stops here does not solve twice. A plateaued, cold-consistent point is reported converged exactly as before.
- A re-evaluation at the resume point cannot pad the post-loop plateau verdict's flat tail (`plateau_at_last_progress` snapshot).
- Budget exhausted while still descending → `max_eval_reached`, the "increase maxiter" warning — not a stall.
- Resumed fits carry an `optimizer_health` warning naming the count.

Result on the fixture: λ = 0 resumes once at eval 12 and plateaus at **−786**, converged; λ = 100 resumes once at eval 165 and finishes **`XtolReached` at −243, gradient norm 0.008**; λ = 5 untouched.

The Tier-3 test's ‖W‖² monotonicity assert compared two norms at the optimizer's floor (9.2e-10 vs 2.3e-9, against 2.65e4 unregularized) with 1e-9 absolute slack — the noise exceeds the slack. It now uses a measured floor (1e-7, ~40× the larger, eleven orders below n0), the same policy the test's own variance assert already applied and stated.

`cold_solve_at` is one definition of "cold" shared by the loop and the final inner loop, so the two cannot drift.

## Alternatives considered

- **Fix the fixture** (init the network so the opening gradient is smaller): a band-aid — the abort is reachable by any model with a large initial gradient, and the λ = 100 shape has nothing to do with the init.
- **Resume on the trace alone** (flat tail < 5): first draft; cannot see the reduction-cap abort, whose ten probes pad the tail to ≥ 10. Replaced by the verdict-based rule.
- **Drop the opening-step cap**: measured in #960 as L-BFGS failing on eval 1 the other way. The cap and the resume are the two halves of not having a trust region in NLopt's L-BFGS.
- **Patch `plis.c`**: not ours; and the same abort semantics would remain for anyone linking upstream NLopt.

## Cross-repo dependency

## Breaking changes
- [x] None (no public-API change; baseline unchanged)

## Numerical validation
- [x] Compared against: prior ferx commit `233246d1` (HEAD) and `3e2c8e4d` (last green nightly)
- [x] Reference values:

```
# before (233246d1), l2_shrinks_weights_and_modulator_variation
weight ‖W‖²: λ=0 60.95117, λ=5 0.00000, λ=100 0.00116     ← λ=0 abort at eval 12, OFV 7902; λ=100 abort at eval 165, OFV −113
# after
weight ‖W‖²: λ=0 2.6475e4, λ=5 9.16e-10, λ=100 2.28e-9    ← λ=0 −786 (plateau, converged); λ=100 −243 (XtolReached); λ=5 −1173 unchanged
CL modulator var: 1.574791 → 0 → 0
```

A fit whose `Failure` is plateau-converged takes the identical path (the loop breaks before any relaunch), so healthy fits are unchanged: `--lib` under `ci` (4733) and `ci,markov,nn` (5092) both green. Warfarin from 50× / 0.02× / far-off inits under `nlopt_lbfgs` never triggers a resume and converges to −286.0042 as before.

## Performance
- [x] No significant impact expected — the resume path only runs after a bare `Failure`; the plateau cold solve it takes is the one the final inner loop needed anyway (cached), and the cheap gates skip it entirely for the no-progress case.

## Tests
- [x] Unit tests: five on `resume_descent` (each cheap gate pinned as *not* calling the verdict closure), `classify_warning_lbfgs_resume_is_optimizer_health`
- [x] Regression test that fails without the fix: `nn::regularizer_fit_tests::first_line_search_abort_is_resumed_not_reported` — 6 s, nn-gated (runs in the per-PR `ci,markov,nn` lib job), `maxiter = 1` = 44-eval budget. With `resume_descent → Stop` both assertions die independently (reported 7901.994, the pre-fix number; no warning). The Tier-3 test passes locally (`--features nn,slow-tests`, 224 s).

- [x] Not applicable (fix; the only new surface is a warning)

## Docs
- [x] `docs/estimation/optimizers.qmd` — new subsection *An L-BFGS abort mid-fit is resumed, not reported*; `docs/warnings.qmd` `optimizer_health` row

## Checklist
- [x] `tools/preflight.sh` green — fmt, check matrix, clippy, rustdoc, docs, public-API
- [x] No `Dual2` path touched

## Docs & examples
- [x] No example execution step needed (fix; docs page updated; no new DSL / fit option)

## Reviewer hints

- `resume_descent` and the loop around `opt.optimize()` in `optimize_nlopt_once` are the whole change; everything else is plumbing for the per-launch gradient count and the cold-solve cache.
- The subtle bit is the `plateau_at_last_progress` snapshot: a resumed launch that makes no progress has re-evaluated the best point once, and that eval must not push a 4-flat tail to 5 in the post-loop verdict.
- The nightly `nn` leg should be dispatched on this branch before merge (`slow-tests.yml` never runs on PRs).

## Open questions

None. Row 0 of #1350 will be flipped ✅ with this PR number once merged.


hiddevandebeek:
@roninsightrx @TeunP not a 100% sure about nn, one of you might want to have a look

TeunP:
Looked at this against `main` (reviewed at head `48e96633`). Short version: #1414 (merged `48f274a6`, this morning) fixed #1277 by a different route — one restart from the best point via `resolve_mid_descent_restart` — and the nightly nn leg has been green since (run 35314121117). So this PR is now conflicting and can't land as is; rebasing it whole would stack two restart mechanisms.

But three pieces here are better than what's on `main`, and I measured the first:

1. **The PR-time regression test.** I ported `first_line_search_abort_is_resumed_not_reported` onto `main` (`85413d90`), minus the warning assert. It passes (OFV −698.40). With the restart wiring disabled (`mid_descent_stall` forced `false`), the full `--lib --features ci,markov,nn --no-fail-fast` suite is 5166 passed / 1 failed — the one failure is this test (reports 3209.805, n_iter 12). Nothing else on `main` can see that mutation before the nightly. Note 65 s in debug on my Mac, not 6 s.
2. **Budget semantics.** Same probe: `maxiter = 1` is a 44-eval budget, the fit spent 56 (12 + a fresh 44). Your remaining-budget rule is the right one; `main`'s doc line saying the restart never grants a second budget is true only for `MaxEvalReached` runs.
3. **The user-facing warning.** On `main` a restarted fit says nothing in `FitResult.warnings` (verbose stderr only). Your `optimizer_health` message + `classify_warning` arm + `warnings.qmd` row fill that.

Also worth keeping: your instrumented-`plis` account of the two abort shapes. `main`'s docs explain the restart as clearing `(s, y)` pairs, which can't be the eval-12 story — that abort is inside the first line search, before any pair exists. I did not re-instrument `plis.c` myself; the trace shape (eleven improving evals, `Failure`, flat tail 0) matches yours.

Platform note: on macOS the eval-12 abort sits at OFV 3209.80 (what #1414 quotes), not 7902 — so bounds need to hold on both; `ofv < 0` does.

Proposal: close this one as superseded, and — @hiddevandebeek, since the pieces are yours — open a small follow-up on top of `main` with (1) the test, (3) the warning, and a decision on (2). The `WEIGHT_NORM_FLOOR` change can be dropped — `main` replaced that assertion already (floor check plus a convergence assert on each fit).

Not run by me: this PR's head, Tier-2 `tests/` under the mutation, non-nn feature sets. Not examined: whether `main`'s restart should be L-BFGS-only like your `resume_descent` (it has no optimizer gate).

hiddevandebeek:
Review of the diff (single pass over `outer_optimizer.rs`, the tests, `nn/mod.rs`, `types.rs`, docs). The core mechanism — `resume_descent` gating, the per-launch gradient count, the cold-solve cache keyed on the restored point, `BudgetExhausted` → `max_eval_reached` — reads correct to me, and I confirmed the NLopt side: `nlopt_optimize` resets `numevals` per call (so `set_maxeval(remaining)` is the right budget), and Luksan's `plis` maps every non-listed `iterm` (including the `-10` restart-state line-search failure) to `NLOPT_FAILURE`. Two things I think are real, one low.

### 1. A resumed launch that makes no progress but ends in an `Ok` status launders the verdict that rejected it (`src/estimation/outer_optimizer.rs:2974`, the `while` loop)

`resume_descent` only relaunches when `failure_is_converged_plateau` said **no** — and one of its three ways of saying no is `consistent == false`: the cold re-solve at the best-seen point sits materially above `best_ofv_clean` (the #751 warm-start-artifact case; the λ = 100 fixture's −99 vs −113 is exactly this). After the relaunch, if L-BFGS makes no further progress and returns `Ok(XtolReached)` — or `Ok(Success)`, which the stagnation guard forces: `detect_stagnation` keeps counting `n_evals − last_improvement_eval` **across launches** (the state is never reset on resume), so the ≥ 50-eval window that had already run 11 evals in launch 1 latches ~39 evals into launch 2, every eval then returns `best_ofv` with `g = 0`, and `pyfut1` hits `gmax <= tolg` → `iterm = 4` → `NLOPT_SUCCESS` — then `stationarity_check_pending` is `false`, `converged = true`, and the cold-inconsistent point is reported as a converged fit with no plateau/consistency verdict at all. Pre-PR the same `Failure` was reported `converged = false` (plus `W_EBE_START_DEPENDENT`); post-PR it is `converged = true` (still with `W_EBE_START_DEPENDENT`, which now contradicts `converged`).

Concretely: best point is a warm-start artifact (cold − best > `PLATEAU_CONSISTENCY_REL_TOL`), verdict false, resume, launch 2 finds nothing better than 1e-3 within the remaining stagnation window → `Ok(Success)` → `converged = true`.

Two cheap options: (a) when the in-loop verdict was rejected on *consistency* (not on tail length), do not resume — the point is not one to descend from; or (b) treat an `Ok` status from a resumed launch that made no progress (`last_sig_feasible_eval == sig_before`) like a `Failure` and route it through the post-loop verdict, so `consistent` still has to hold. Either way, the stagnation latch should be reset (or resumption skipped when it is already latched) on a new launch — a resumed launch under a latched guard is a no-op that returns `Ok(Success)` on its first eval.

### 2. The `plateau_at_last_progress` correction discards the *whole* resumed launch, not the one re-evaluation it is documented to discard (`src/estimation/outer_optimizer.rs:3319`)

The comment (and the PR text) says the resumed launch "has re-evaluated the best point once, and that eval must not push a 4-flat tail to 5". The code does `feasible_evals = feasible_before`, i.e. it drops every feasible eval of the no-progress launch — the re-evaluation *and* all of the line-search probes that followed it, which are exactly the same kind of flat-tail evidence launch 1's probes were.

Scenario: launch 1 aborts with progress at feasible eval 20 and a 3-eval tail (23) → verdict false on tail length → resume. Launch 2 re-evaluates (24), then burns its ten reductions on a genuinely flat, cold-consistent objective (25–34) → `Failure`, no progress → `Stop`. Post-loop: `feasible_evals` is reset to 23, `flat_tail = 3 < 5`, `converged = false` — although the trace holds a 14-eval flat tail at a cold-consistent point, which is the engine's own definition of a finished fit, and the docs page added here promises "a `Failure` at a genuine plateau is reported converged". Not a regression against pre-PR (that fit was non-converged at the first `Failure` too), but it is the case the resume exists to fix. `feasible_before + 1` (skip just the re-evaluation) matches the stated intent; if the whole-launch discard is deliberate, the comment and the docs should say so.

### 3. Low — resuming on a tiny remaining budget shrinks L-BFGS's memory (`src/estimation/outer_optimizer.rs:3029`)

`luksan_plis` sizes its memory as `mf = MAX(MEMAVAIL/n, 10)` but then `if (stop->maxeval <= mf) mf = MAX(stop->maxeval, 1)`. With `remaining` ≤ 10 the relaunch runs with `mf = remaining` (down to 1 — effectively steepest descent). Harmless given how few evals are left, but worth a one-line comment or a floor (e.g. don't resume on fewer than `n + 1` evals, since a single L-BFGS iteration with a gradient can't do less than that anyway and `BudgetExhausted` already has the right warning).

Nits, no action needed: `plateau_at_last_progress` is really "tracker at last *resume*" (it is set on every resume, regardless of what the launch then does), and `resume_descent`'s `last_sig_feasible_eval >= 2` is subsumed by `> feasible_evals_at_launch` for every launch after the first — both fine as written.

hiddevandebeek:
Addressed the review in b97890ab:

1. **No-progress resumed `Ok` no longer launders the verdict.** After the resume loop, the post-optimization block now asks `plateau_trace_after_resume` whether the last launch progressed past the point it was resumed from. If it did not and NLopt returned an `Ok` status, the result is routed through the same plateau + cold-consistency verdict a `Failure` gets (`stationarity_check_pending = true`, `converged = false` until it passes). The cached cold solve keyed on the best point is reused, so a cold-inconsistent point stays `converged = false` + `W_EBE_START_DEPENDENT` as it was pre-PR. The stagnation guard's window also restarts with each launch (`note_launch`, which now owns the per-launch re-arming alongside the overshoot-cap count), so a resumed launch is not latched a few evals in by a window the aborted launch mostly spent; `best_at_last_improvement` is deliberately *not* reset, so a resumed launch still has to beat the best point to count as improving.

2. **Only the re-evaluation is dropped from the flat tail.** `plateau_trace_after_resume` returns `(feasible_evals − 1).max(feasible_at_resume)` for a no-progress launch, so the launch's own probes count as plateau evidence; the "resume off a 3-eval tail, burn ten flat reductions" case now completes the plateau. The docs paragraph says this explicitly now.

3. **Small `remaining` and L-BFGS memory:** a comment at `set_maxeval(remaining)`. Luksan's `mf = maxeval` whenever `maxeval ≤ MEMAVAIL/n` (= 1310720/n, so always for ferx-sized `n`) — the memory is *already* sized from the budget on the first launch too, and it cannot bind since a launch cannot take more iterations than it has evals. No floor added.

Tests: `plateau_trace_after_resume_drops_only_the_re_evaluation` (mutations: `feasible_at_resume` in place of `− 1` fails the ten-probe case; dropping the `− 1` fails the single re-eval case) and `note_launch_rearms_the_cap_count_and_the_stagnation_window_once_per_launch` (drop the window reset → the 49-quiet-evals assertion fails; drop the same-launch guard → the first no-op call moves the marks). `first_line_search_abort_is_resumed_not_reported` still passes (−540 by eval 44), the full `outer_optimizer::` module is green, and `tools/preflight.sh` is OK.

Also took the nit: `plateau_at_last_progress` is renamed `plateau_at_last_resume`.

hiddevandebeek:
Closing as superseded: #1414 (`48f274a6`) landed a different fix for #1277 — one outer restart from the best point on a diagnosed mid-descent stall, adopted only on a strictly lower objective — and closed the issue. This branch's in-launch resume loop is a second mechanism for the same behaviour, and layering the two would need a deliberate design for how they interact rather than a rebase, so it is not being merged.

Pieces here that #1414 does not carry, in case a follow-up ever wants them: the per-launch re-arm of the identity-Hessian overshoot cap (`should_cap_gradient` reading a per-launch gradient count), resuming on the *remaining* `outer_maxiter` budget rather than a fresh attempt, and the guard that an `Ok` status from a no-progress relaunch cannot bypass the plateau/cold-consistency verdict.

Issue #1285, opened 2026-09-08
bug(ode): the EKF never corrects the mean with the innovation, so a [diffusion] fit is a variance-inflated ODE rather than a state-space likelihood


TeunP:
Found while evaluating Chen, Bauer & Li, *AAPS J* 28:129 (2026), doi:10.1208/s12248-026-01302-2 (open preprint: bioRxiv doi:10.64898/2026.05.28.728340), which uses SDE system noise as a model-misspecification diagnostic in NONMEM. Evaluation only; nothing implemented. Line refs are at `c877b9a9`.

### What the code does

`solve_ekf` (`src/ode/ekf.rs:149`) propagates `P` and shrinks it at each observation, but never corrects the mean:

- `kalman_update` (`ekf.rs:96`) takes `(P, obs_cmt, r_obs, n)` and returns `(P⁺, p_obs)`. It has no observation argument. `solve_ekf` itself never receives `DV`: its inputs are `doses`, `obs_times`, `r_obs_vec`.
- The mean `u` changes in exactly two places: bolus application (`ekf.rs:310`) and the end-of-segment copy from the ODE solution (`ekf.rs:480`).
- `r_obs_vec` is built from the plain `ode_predictions` pass (`predictions.rs:6328-6329`), so `R_j` is data-independent too.
- The objective discards the EKF mean anyway: `stats/likelihood.rs:892` is `let (_, p_obs) = ode_predictions_ekf_with_diffusion(...)`, and the `ipreds` it scores come from `ode_predictions`.

So the per-subject data term of a `[diffusion]` model is

```
Σ_j [ (y_j − f_ODE(t_j))² / (R_j + P_j)  +  ln(R_j + P_j) ]
```

with `P_j` a schedule that depends on θ, η, the sampling grid and `R_j`, and on nothing the subject observed. That is a heteroscedastic inflation of the residual variance on top of the deterministic ODE. It is not a state-space likelihood.

### Why this is a defect and not a convention

[...]

PR #1413, opened 2026-09-17
fix(ode): equilibrate an SS=1 record on the EKF path for both mean and covariance [closes #1260]

hiddevandebeek:
## Why

#1350 row 11 / #1217 row f. `solve_ekf` applied an `SS=1` record as a single dose from an empty state with `P = 0` — no run-in, nothing rejected, `W_SDE_STEADY_STATE` (#1263) the only trace. Measured against the explicit 41-dose train the record stands for, on the issue's Michaelis–Menten fixture (`VMAX = 20`, `KM = 50`, `central ~ 5.0`, 100 mg q12h for 480 h, samples 2/5/8/11 h after the last dose):

| quantity | `SS=1` record (before) | 41 explicit doses | |
|---|---|---|---|
| `p_obs` at t = 482 | 9.15 | 20.35 | **55 % low** |
| `p_obs` at t = 485 (after the first update) | 11.80 | 12.05 | 2.2 % low |
| filter mean at t = 485 | 42.63 | 47.18 | 10 % low |
| objective, issue's data | 610.41 | 489.88 | **Δ 120.53** |

`p_obs` is the one quantity the objective takes from the filter (`likelihood.rs` discards the EKF mean; `IPRED` is the ODE path's, which equilibrates). On a nonlinear right-hand side the mean is what the Riccati Jacobian is linearised around, so the wrong mean reaches `p_obs` too; on a linear one `P` at the record is simply `0` where the stationary value is `q / 2k`.

## What changed

The record is equilibrated the way the ODE path's nonlinear fallback does it (`equilibrate_ss_pk_state`): `ode::ekf::equilibrate_ss_ekf` expands the pulse train from `u = 0, P = 0` — cycles of *(apply dose; integrate `II`)* on that fallback's cycle-local clock (each window opens at its own pulse, `TAD` anchored `0` / `−T_inf` per window through `ss_run_in_params`, `TAFD` `NaN`), a bolus or an `F`-reshaped active window followed by a quiet one — carrying `P` alongside `u` through the same Euler Riccati quadrature the walk uses (factored into `propagate_covariance_between`, one spelling for both), until the concatenated pair stops moving by `SsStopTracker`'s criterion or `SS_EQUILIBRATION_CYCLES` is spent, with the #867 non-convergence warning on a spent cap. The record's own dose then lands on the trough as before (bolus) or through `active_infusions` (infusion). `P` travels with `u` rather than being seeded after it because on a nonlinear RHS the Jacobian along the cycle depends on the mean's phase.

- An infusion longer than `II` is declined exactly as the ODE run-in declines it (zero pair, i.e. the pre-fix answer; `W_STEADY_STATE_INFUSION` already names it at check time).
- `tad_anchor_for`'s `ss` fold, which described a pulse train this engine had never built (TeunP's 24.1 % `TAD` measurement on #1263), now describes one it does — pinned by the infusion test on a `TAD`-reading RHS.
- `W_SDE_STEADY_STATE` is retired; the check-time test is flipped to pin its absence against the lagtime sibling on the same population.
- **Clock (review #4).** The first draft ran the run-in on a monotone `m·II` clock — the *input-rate* train's convention, not the bolus/infusion fallback's — so a right-hand side reading `T` / `TIME` was equilibrated differently by the two engines whose halves the objective mixes. Now cycle-local. Pinned against a closed form outside both engines (`dA/dt = −ke·A·(1 + c·T)`, trough `D·M/(1 − M)` with `M = exp(−ke·(II + c·II²/2))`): EKF and ODE both within 1.5e-9 of it; under the monotone-clock mutation the EKF seed collapses to the single-dose value (8.1676 vs 15.0806) and only that test dies — the `TAD`/autonomous fixtures cannot see the clock.

**On the #1285 ordering.** TeunP's comment on #1260 orders the real fix behind #1285 (the missing innovation update). It is independent of it: no observation is assimilated in the pre-history an `SS=1` record stands for, so the run-in is the same object before and after that update lands, and the mean it equilibrates is consumed *today* by the Jacobian — which is where the 120.5 came from. The update, when it lands, changes what happens after the first observation and leaves this code untouched.

## Alternatives considered

- **Named reject (option 2 on the issue).** Honest and cheap, but a band-aid: it would have made every SDE user with an `SS=1` dataset expand it by hand, for a run-in the ODE path already had the template for.
- **Exact linear fixed point for `P`** (a discrete Lyapunov solve, mirroring the ODE path's `(I − M)⁻¹b` for the mean). Not needed: a bolus leaves `P` untouched, so `P`'s cycle map is the continuous Riccati flow, and expanding the train with the walk's own quadrature is what makes an explicit train and the run-in agree to the stop tolerance — the oracle the issue defined.
- **Seed `P` only, after the ODE path's mean.** Passes on a linear RHS and fails on the issue's fixture (mutation below).

## Cross-repo dependency

## Breaking changes
- [x] None (no public-API change; baseline unchanged. `W_SDE_STEADY_STATE` no longer fires — a retired warning code, not an API item)

## Numerical validation
- [x] Compared against: the explicit dose train (the issue's oracle; the ODE path scores the two spellings identically, so there is no NONMEM run to add — NONMEM's `SDE.f90` would hand its SS routine the augmented Riccati system, which #1210 measured as `+INF` on the one augmented system it was fed, see TeunP's second comment on #1260)
- [x] Reference values:

```
# solve_ekf, MM model, SS=1 at t=480 vs 41 doses — before
t=485 ss: ipred 42.630304 p_obs 11.796572 | train: ipred 47.175857 p_obs 12.054604
t=488 ss: ipred 20.134978 p_obs  9.832652 | train: ipred 23.029936 p_obs 10.174594
# after
t=482 ss: ipred 80.473512 p_obs 20.351482 | train: ipred 80.473512 p_obs 20.351482
t=485 ss: ipred 47.175857 p_obs 12.054604 | train: ipred 47.175857 p_obs 12.054604
t=488 ss: ipred 23.029936 p_obs 10.174594 | train: ipred 23.029936 p_obs 10.174594
worst relative error ipred 2.7e-14, p_obs 3.2e-12

# linear dA/dt = −kA, q = 0.25, k = 0.05: seeded P at t=482
2.5000000000020 vs q/2k = 2.5 (7.8e-13); single-dose arm 0.4637
mean 200.546448 vs closed-form peak 200.545380 (5.3e-6, RK45 at reltol 1e-4 compounded by 1/(1−e^{−k·II}))

# SS=1 infusion (AMT 100, RATE 20, F = 0.5) vs its explicit train, TAD-reading RHS
worst relative error ipred 7.1e-13, p_obs 9.9e-13 (single-infusion arm 5.2e-1 / 7.6e-1)

# T-reading RHS dA/dt = −ke·A·(1 + 0.05·T), SS=1 at t=480, zero diffusion — vs closed form
t=482  ekf 15.08059537  ode 15.08059537  closed form 15.08059537  (8.5e-11 / 8.6e-11); single-dose 8.1675598 (4.58e-1)
monotone-clock mutation: ekf 8.16755988 (the single-dose value to 8 figures), ode unchanged

# fit(), outer_maxiter = 0, MM SDE model through the parser
SS=1 record 10.125356 | 41 doses 10.125356 | SS=0 control 15.242741
SS=1 record with the solve_ekf hook disabled (the pre-fix behaviour): 9.290403
```

The fit-level straddle is measured against the pre-fix behaviour itself: 0.835 OFV (9.2904 vs 10.1254 — *below* the train, `p_obs` 55 % low at the first sample on an `IPRED` that is the ODE path's and so already equilibrated), and the `1e-6` bound sits six orders inside it. The `SS=0` control's 5.12 is a different quantity — with the flag off the ODE path's `IPRED` is single-dose too, so both halves move — and is kept only as the non-degeneracy check on the flag. The issue's 120.5 was on DVs with more residual leverage.

## Performance
- [x] Measured (review #1), on the #1260 fixture, one objective evaluation, debug build: **569 ms with the `SS=1` record, 996 ms with the 41 explicit records it stands for, 58 ms with the record read as a single dose.** The run-in early-stops at 16–21 cycles (`SsStopTracker` on the concatenated `(u, P)`), not the 50-cycle cap. So the record costs ~10× the wrong single-dose answer and ~0.6× the train it abbreviates, on every evaluation and every FD perturbation of the filter. A linear RHS has a closed-form pair (the ODE path's affine fixed point for the mean, a discrete Lyapunov solve for `P`) that would remove most of that; noted in the docstring for the day a dataset with `SS=1` on every occasion needs it. Subjects without SS records take the identical path (every pre-existing `ode::ekf` test unchanged).

## Tests
- [x] Unit tests: `ekf_ss_record_matches_the_explicit_train_on_a_nonlinear_rhs`, `ekf_ss_covariance_seeds_at_the_stationary_riccati_value`, `ekf_ss_infusion_matches_the_explicit_train_on_a_tad_reading_rhs`, `ekf_overlapping_ss_infusion_is_the_single_dose_answer`, `ekf_ss_run_in_without_a_steady_state_warns`, `ekf_ss_run_in_shares_the_ode_run_in_clock_on_a_time_reading_rhs`; `sde_steady_state_scores_like_the_explicit_train::one_ss_record_scores_like_forty_one_explicit_doses` (evaluation-only `fit()` through the parser); `a_steady_state_dose_under_diffusion_no_longer_warns`.
- [x] Regression tests that fail without the fix: each train / closed-form test asserts its `SS=0` arm far from the oracle on **both** `ipred` and `p_obs`, so an implementation that ignores the flag cannot meet the bound. Mutation run: equilibrate the mean and leave `P = 0` (`let _ = p_ss`) — all three die on `p_obs` alone (5.5e-1 / 7.6e-1 / 0.46 vs 2.5) with `ipred` still at 1e-13.

- [x] Not applicable (fix; an `SS=1` record under `[diffusion]` now just works)

## Docs
- [x] `docs/model-file/diffusion.qmd` — SS row removed from the gaps table; new *Steady-state records are equilibrated* section with the measurements, the clock convention and the cost; the model-time paragraph stays in its own section (review #5)
- [x] `docs/examples/sde.qmd` — the *SDE + ADDL* tip no longer claims an "analytical" covariance reset at an SS event

## Checklist
- [x] No `Dual2` path touched (the EKF is FD-only)

## Docs & examples
- [x] No example execution step needed (fix; no new DSL / fit option)

## Reviewer hints

- `equilibrate_ss_ekf` and the eight-line hook in `solve_ekf`'s dose loop are the change; `propagate_covariance_between` is the walk's Riccati sub-stepping moved into a function so the run-in cannot drift from it.
- The run-in reads model time exactly as the ODE bolus/infusion run-in does (`ss_run_in_params`, now `pub(crate)`): `TAFD` is `NaN`, `T` / `TIME` are cycle-local, `TAD` is anchored per window — so a `TAFD`/`T`/`TIME`-reading RHS under `[diffusion]` with an `SS=1` record gets the same `W_STEADY_STATE_ABSOLUTE_TIME` treatment, and the comment on that gate in `validation.rs` now says so instead of "never equilibrates" (review #2).
- `SsStopTracker` sees the concatenated `(u, P)`, so a cycle in which the mean has settled but `P` has not keeps running — on the fixtures `P` settles a few cycles after the mean.

## Open questions

None. #1350 row 11 and #1217 row f will be flipped ✅ with this PR number once merged; TeunP's #1285-ordering comment is answered in *What changed* and on the issue.


hiddevandebeek:
## Review — PR #1413 (EKF `SS=1` equilibration)

Read every hunk plus the surrounding `solve_ekf`, `equilibrate_ss_pk_state`, `SsStopTracker`, `tad_anchor_for`/`tad_referent`, `active_infusions` and the `likelihood.rs` consumer. The core change is sound: the run-in is the walk's own quadrature (`propagate_covariance_between` is a faithful extraction of the inline Riccati sub-stepping — same `1e-15` gate, same midpoint interpolation), the bolus/infusion cycle shapes and the bail-outs mirror the ODE fallback, and the trough-then-pulse ordering matches. No correctness defect found in the numerics. Five findings, all low.

### 1. `src/ode/ekf.rs:258-273` — the run-in always pays the full capped train; the PR's "no significant impact" is not measured (low, performance)

Unlike `equilibrate_ss_pk_state`, which takes the exact affine fixed point (`periodic_ss_fixed_point_pk`, a handful of one-cycle solves) for a linear RHS and only falls back to the capped train for a nonlinear one, `equilibrate_ss_ekf` always expands the train. The stop is `SsStopTracker` at `SS_EQUILIBRATION_TOL = 1e-12`, and `dosing.rs`'s own note on `ss_equilibration_tail_warning` says a converged model routinely runs out the whole budget "with its increments jittering at the solver noise floor" (`reltol = 1e-4`), so in practice this is ~50 cycles, each an RK45 solve over `II` plus `II / 0.5` Euler–Riccati steps with an FD Jacobian apiece. That is per `SS=1` record, per subject, per objective evaluation — and the EKF path is FD-gradient only, so per parameter perturbation as well. Concrete case: a linear 1-cpt SDE TDM dataset where every dosing occasion carries `SS=1` (the common spelling) — each subject's filter cost goes from a few short segments to ~50 × 12 h cycles per record, plausibly an order of magnitude per evaluation. It is the same cost an explicit train paid in the walk, so it is not a regression against a correct baseline, but the PR claims no impact without a number. Suggest timing the `fit()` fixture with/without the record, and — since a bolus leaves `P` untouched — noting that a linear RHS has a closed-form pair (the ODE path's `(I − M)⁻¹b` for the mean, a discrete Lyapunov solve for `P`) if the number is bad.

### 2. `src/api/validation.rs:5195` — comment now states the opposite of the code (low)

> `solve_ekf` seeds a finite TAFD anchor and never equilibrates (#1260), which reads like an exemption — but `ode_predictions_ekf_with_diffusion` computes the Kalman `R` from a standard `ode_predictions` pass, and that one does run the run-in, so the `NaN` reaches the likelihood anyway

After this PR `solve_ekf` does equilibrate, and its run-in reads `TAFD` as `NaN` directly (`ss_run_in_params`), so `W_STEADY_STATE_ABSOLUTE_TIME` now has a first-hand reason to cover `[diffusion]` rather than the indirect one this comment defends. Left as-is, the next reader of this gate is told the EKF never runs a run-in.

### 3. `src/api/tests/sde_integration.rs:546-556` — the straddle control is not the pre-fix behaviour, and the comment misattributes the gap (low, test)

The control is `SS=0` on **both** engines, so its 5.12 OFV gap is carried by the ODE-path `IPRED` (single-dose mean vs the train's DVs) *and* `p_obs` — not "by `p_obs` alone" as the comment says (`likelihood.rs:941` takes `IPRED` from `ode_predictions`, which is single-dose in that arm). The regression this test exists to catch — the EKF applying the record as a single dose while the ODE path equilibrates it — has the equilibrated `IPRED` and only `p_obs` wrong, and its OFV on this fixture is never recorded, so the `1e-6` bound is not shown to straddle the OLD behaviour (the repo's own rule in CLAUDE.md). It almost certainly does (`p_obs` 55 % low at the first sample moves `log|V|` by O(1)), but the test should say the measured pre-fix `ss` OFV rather than lean on the `SS=0` arm.

### 4. `src/ode/ekf.rs:261-263` — a `T`/`TIME`-reading RHS sees a different run-in clock on each engine (low)

The EKF run-in integrates on the monotone clock `(m·II, (m+1)·II)`, `m = 0..50`; the ODE bolus/infusion fallback and the exact solve integrate every cycle on `(0, II)` (`ext_cycle`, `advance_forced`). For a RHS that reads `T`/`TIME` (a circadian term, say), the two engines now equilibrate different states, and the objective mixes the ODE path's `IPRED` with the EKF's `p_obs`. `W_STEADY_STATE_ABSOLUTE_TIME` already fires for `[diffusion]`, so this is warned, but the PR text ("reads `TAFD` as `NaN` like the ODE run-in does") covers only the `TAFD` half. Worth a sentence in the docs or a deliberate choice of the same clock as the ODE fallback.

### 5. Docs placement / stale example text (low)

- `docs/model-file/diffusion.qmd:179` — the new `### Steady-state records are equilibrated` h3 is inserted *inside* `## Model-time builtins in the RHS`, so the paragraph that follows it at `:199` ("`TIME` / `T` is the integration variable itself…", `TAFD`/`TAD` slots) now renders as part of the steady-state subsection rather than the model-time section it belongs to.
- `docs/examples/sde.qmd:105` still says "the EKF is reset to the **analytical** steady-state covariance at each SS dose event". It was false before this PR (nothing was reset) and is still inaccurate after it — the covariance is the numerically expanded Riccati trough, and an infusion longer than `II` is declined.

Not raised: the run-in overwrites `u`/`P` wholesale (also zeroing them on the `cmt_idx >= n` / overlapping-infusion bail-outs, and discarding a co-timed ordinary bolus listed earlier in the dose list) — every ODE walk does the same via `u.copy_from_slice(&equilibrate_ss_state(..))`, and `E_DOSE_CMT_OUT_OF_RANGE` rejects the first case up front.

hiddevandebeek:
Addressed in add24bb8 (after merging `main`, c750eae1):

1. **Cost** — measured rather than asserted. On the #1260 fixture (one objective evaluation, debug build): 569 ms with the `SS=1` record, 996 ms with the 41 explicit doses it stands for, 58 ms single-dose. The run-in early-stops at 16–21 cycles, not the cap: `SsStopTracker`'s criterion on the concatenated `(u, P)` is met well before `1e-12` becomes solver-floor noise here. Recorded in `equilibrate_ss_ekf`'s docstring, the docs page and the PR's *Performance* section, with the linear closed-form pair noted as the escape hatch if an `SS=1`-on-every-occasion dataset ever needs it.
2. **`validation.rs` comment** — rewritten: `solve_ekf` now runs its own run-in on the same clock and anchors, so the gate covers `[diffusion]` first-hand, and the old indirect reason is kept as history.
3. **Straddle** — measured against the pre-fix behaviour itself (hook disabled): the `SS=1` arm scores **9.290403** vs the train's 10.125356, so the `1e-6` bound sits six orders inside a 0.835 gap. The `SS=0` control's comment now says its 5.12 moves both `IPRED` and `p_obs` and is kept only as the flag's non-degeneracy check.
4. **Clock** — the run-in now integrates on the ODE bolus/infusion fallback's cycle-local `(0, II)` clock (`0` / `−T_inf` anchors per window through `ss_run_in_params`), so a `T`/`TIME`-reading RHS is equilibrated identically by both engines. New test `ekf_ss_run_in_shares_the_ode_run_in_clock_on_a_time_reading_rhs` checks EKF and ODE against a closed form outside both (`D·M/(1 − M)`, `M = exp(−ke·(II + c·II²/2))`): both within 1.5e-9. Under the monotone-clock mutation the EKF seed collapses to the single-dose value (8.1676 vs 15.0806) and only that test dies — confirming the `TAD`/autonomous fixtures could not see it.
5. **Docs** — the model-time paragraph moved back above the h3 so it stays in *Model-time builtins in the RHS*; `examples/sde.qmd`'s "analytical steady-state covariance" line replaced with what actually happens and a link to the new section. `docs-lint --check` green.

`tools/preflight.sh` fmt/check/clippy/rustdoc/docs/public-api green; `ekf`/`sde_` lib tests 37 passed.

PR #1414, opened 2026-09-17
fix(estimation): restart a fit that stopped mid-descent instead of reporting where it died [closes #1277]

roninsightrx:
## Why

`nn::regularizer_fit_tests::l2_shrinks_weights_and_modulator_variation` has failed on every
`Slow regression tests (nn)` run since #1215, with bit-identical values. #1277 filed it as two
failures; on today's `main` (`233246d1`, nightly run
[35189123134](https://github.com/FeRx-NLME/ferx-core/actions/runs/35189123134)) only this one
survives — `focei_makes_substantial_progress_on_nn_model` now passes, the nn leg is
`5037 passed; 1 failed`, the core leg is green, and the #1239 timeout is gone (the leg finished
in 54 of its 100 minutes). So the nn leg's red is this test and nothing else.

It does not read as a tolerance that drifted, and it isn't one. The assertion that fires
compares **weight norms from fits that never converged**:

```
λ=0    OFV 3209.81   ‖W‖² 287.9    converged = false   ← NLopt "Failure" at eval 12
λ=5    OFV 2816.54   ‖W‖² 349.6    converged = false   ← NLopt "Failure" at eval 12
λ=100  OFV -1185.46  ‖W‖² ~0       converged = true    (352 evals)
```

Both dead fits quit *mid-descent*, with the trace still falling ~2600 OFV per evaluation. The
λ=0 fit therefore sat **4400 OFV units above** a point inside its own feasible set — the λ=100
solution, which is reachable by simply shrinking the weights. The ordering the test asserts was
never a statement about `nn_l2`; it was noise from two stalls.

NLopt's bare `Failure` is ambiguous, and `failure_is_converged_plateau` has told the two cases
apart since #751 — a flat tail, displacement from `x₀`, and a self-consistent cold re-solve
means a finished fit; anything else is a genuine stall. But that diagnosis was only ever used to
*label* the run `converged: false`. The estimates handed back were wherever the line search gave
up.

## What changed

**Engine (`src/estimation/outer_optimizer.rs`).** A run diagnosed as a genuine mid-descent stall
now gets exactly one restart from its own best point, which is what clears the `(s, y)`
curvature pairs the failed line search was built on. `optimize_nlopt_once` returns an
`AttemptOutcome { left_init, mid_descent_stall }` instead of a bare `left_init`, and a new pure
`resolve_mid_descent_restart` makes the decision — the same shape as the existing
`resolve_stall_retry`, so both are unit-testable without driving two NLopt fits.

Three gates keep it from making any fit worse:

- it fires **only** on `mid_descent_stall`. A converged fit, a plateaued `Failure`, and a run
  that spent its `maxiter` budget (`MaxEvalReached` is a *success* state, not a `Failure`) are
  untouched and bit-identical — the restart can never quietly grant a second evaluation budget;
- the restart is adopted only on a **strictly lower** penalized objective. Unlike the
  `left_init` retry there is no second condition to check: this attempt starts *at* the reported
  point, so anywhere it ends is somewhere the fit genuinely reached;
- a **cancelled** run is skipped, since cancellation stops through the objective's `1e20`
  short-circuit rather than at a minimum and looks like a stall to every test here.

There is one restart; a second stall reports as a stall. `worth_restarting_mid_descent` is a
named function rather than one expression at the call site because the stall verdict has to be
read *before* `gate_converged_on_objective` demotes `converged` for an unrelated fact — a
non-finite objective — and separating those two by statement order alone is a trap: moving the
read three lines down would turn every non-finite fit into a restart from the very estimates
that poisoned it.

**Test (`src/nn/mod.rs`).** `l2_shrinks_weights_and_modulator_variation` is hardened in the same
pass, because it could not see the defect it caught:

- it now asserts **each of its three fits converged**. The stall was visible only as a weight
  norm of 287.9 against a converged 15340.6 — a difference no ordering check can tell from a
  second local optimum;
- the assertion that failed (`n_big <= n_mid + 1e-9`, reported as `0.00000 → 0.03844`) compared
  two numbers **both at the optimizer's floor**, where the order is set by the path the
  optimizer took rather than by λ. Replaced by the floor itself, which is the strictly stronger
  claim (it implies both orderings against `n0`) and survives either sign of the noise;
- the trace prints in scientific notation. `{:.5}` rendered every regularized norm as `0.00000`,
  which is precisely what hid the fact that the two numbers being ordered were both zero;
- the non-degeneracy guard is stated relative to the flatness threshold rather than as a bare
  `> 1.0`. The λ=0 optimum on this null-covariate dataset moves between platforms (1.245 here
  against the ~480 the fixture was written on) while its distance from the regularized fits —
  which sit at 1e-23 — does not.

## Alternatives considered

**Widen the tolerance.** Rejected: the numbers were not close. `‖W‖² = 287.9` where the
converged fit gives `15340.6` is not a tolerance, and a bound loose enough to pass would have
stopped testing anything. CLAUDE.md's rule — measure the realised error, never justify a bound
with a story — is what turned up the stall in the first place.

**Extend the #751 `left_init` retry.** It cannot reach this: these fits *did* leave their
initial estimates, which is exactly why the existing gate never saw them. Its held-cap remedy is
also the wrong one — #960 measured that holding the identity-Hessian cap costs ~11 OFV on a fit
that is descending normally, and these fits were descending normally right up to the failure.

**Loop restarts until convergence.** One restart converged the reproducer, and an unbounded loop
makes the worst case unbounded on a runner that is already budget-constrained (#1239). A second
stall reports as a stall, which is the honest signal.

**Fix it in the fixture (more `maxiter`, a different start).** `maxiter = 50` was not the
binding constraint — the fit stopped at evaluation 12. The defect is in the optimizer's
stop handling and is not NN-specific; the DCM fixture is just where it was caught.

## Cross-repo dependency

No public API change (`api/ferx-core-public-api.txt` unchanged — `AttemptOutcome`,
`resolve_mid_descent_restart` and `worth_restarting_mid_descent` are all `pub(crate)` or
private).

## Breaking changes
- [ ] Public Rust API
- [x] None

Fits that already converged are bit-identical. Fits that stalled mid-descent change — that is
the fix, and they change only toward a lower objective.

## Numerical validation

- [x] Compared against: prior ferx commit `233246d1`
- [ ] Compared against: NONMEM / nlmixr2

**No NONMEM anchor, and the reason is the usual one for an optimizer stop rule:** the quantity
changed is *where ferx's optimizer stops*, which has no NONMEM spelling — NONMEM's own
termination is a different algorithm with different criteria, so an anchor would measure the two
stop rules against each other, not this change. What can be validated instead is validated, and
all three are internal oracles that do not depend on a second engine:

1. **A lower bound the stalled fit violated.** The λ=100 solution is inside the λ=0 feasible
   set, so the unregularized fit cannot legitimately sit above it. Before: 3209.81 vs -1185.46.
   After: -676.77, below the λ=100 fit's own penalized objective. This is what makes "3886 OFV
   units better" a *correctness* statement and not merely a bigger number.
2. **The optimizer's own convergence test.** All three fits now satisfy
   `failure_is_converged_plateau` — flat tail, displacement, and a cold inner re-solve that
   reproduces the best-seen objective — which is the property the estimates are supposed to have
   and did not.
3. **Adoption is monotone by construction.** The restart is kept only on a strictly lower
   penalized objective, so no reported fit can move the wrong way, whatever the model.

Reference values (`tests/fixtures/two_cpt_dcm_regularized.ferx`, `data/two_cpt_oral_cov.csv`,
macOS/arm64, `--release`):

```
# before (233246d1)
λ=0    OFV 3209.808716   ‖W‖² 2.878873e2   converged=false   "Outer optimization did not converge"
λ=5    OFV 2816.542790   ‖W‖² 3.495884e2   converged=false   "Outer optimization did not converge"
λ=100  OFV -1185.462085  ‖W‖² ~0           converged=true

# after
λ=0    OFV -676.774625   ‖W‖² 1.5340593693e4     converged=true
λ=5    OFV -726.036785   ‖W‖² 4.0348126402e-8    converged=true
λ=100  OFV -1185.462085  ‖W‖² 5.9208012108e-11   converged=true
                         CL modulator var: 1.245131e0 → 3.236735e-23 → 2.624375e-26
```

The λ=100 fit is unchanged to the last digit — it never stalled, so the restart never runs.

## Performance

- [ ] Faster / Slower

The restart costs one extra optimization, and only on a fit that had already failed to converge
— i.e. on a fit whose result was not usable. Converged fits pay nothing: the gate is checked
before anything runs. Worst case for `optimize_nlopt` rises from two optimizations to three (the
`left_init` retry can fire first and hand its result here); both extra legs require a diagnosed
stall.

Full nn + core slow-test legs on this branch:
[35268862787](https://github.com/FeRx-NLME/ferx-core/actions/runs/35268862787) — both green,
nn 31 min, core 46 min, comfortably inside the 100 min budget. (Not a timing claim against the
nightly: those ran from different cache states.)

## Tests

Three Tier-1 tests on the pure decision functions, each naming the regression it catches:

- `test_mid_descent_restart_only_fires_for_a_live_stall` — the `restart` closure **panics**, so
  a gate that lets a converged or cancelled fit through is a failure rather than a silently
  slower suite;
- `test_mid_descent_restart_keeps_the_better_attempt` — worse and tied restarts keep the first
  attempt (adoption is strictly-lower, so the reported fit can never get worse), and the closure
  asserts it is handed the *stalled* value, pinning that the restart starts from the reported
  point rather than from `x₀`;
- `test_mid_descent_restart_skips_a_non_finite_objective` — the `NaN` / `±inf` rows fail if the
  objective argument to `worth_restarting_mid_descent` is dropped, which is the mutation that
  would send every non-finite fit into a pointless restart.

Tier-3: `l2_shrinks_weights_and_modulator_variation` itself, which fails without the fix — that
is the measured pre-state in the table above, not an assumption. Two mutations were checked
against it beyond toggling the fix off: restarting from `init_params` instead of the stalled
point re-runs a deterministic fit to the same stalled result, so nothing is adopted and the new
`converged` assertion fires; and the old floor-vs-floor ordering assertion is exactly what #1277
reported passing-then-failing at random, which is why it is gone rather than re-tuned.

Suites run locally on this branch, all green:

- `--features ferx-core/ci --tests` — 6401 passed, 0 failed
- `--features ferx-core/ci,ferx-core/survival,ferx-core/nn --tests` — 6861 passed, 0 failed
- both slow-test legs on CI (linked above)

- [x] Not applicable (internal fix with no new API surface)

## Docs

- [x] `docs/` updated for user-visible changes — new
      `docs/estimation/foce.qmd#stopping-without-converging`, covering what a bare NLopt
      `Failure` means, how ferx splits a plateau from a stall, what the restart does and what it
      is adopted on, and the three cases deliberately left alone (`maxiter` exhaustion,
      cancellation, a second stall).
- [ ] New pages linked in `docs/_quarto.yml` — N/A, new section on an existing page
- [x] `CHANGELOG.md` `[Unreleased]` entry added, under `Fixed`
- [ ] No user-visible change

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — N/A, nothing under
      `sens/` or `pk/` is touched; this is optimizer control flow only

### ferx-site
- [x] No new data file

### ferx-book

### Example execution
- [x] No example execution step needed (internal fix, no user-visible API or DSL change)

## Reviewer hints

The subtle line is `let stalled_mid_descent = stationarity_check_pending && !converged;` and
*where* it sits. It must be read before `gate_converged_on_objective`, which demotes `converged`
for an entirely different fact. That is why the objective is a second argument to
`worth_restarting_mid_descent` rather than folded into the boolean: written as one expression
the two facts are separated only by statement order.

Second, the `T = (OuterResult, bool)` threading through `resolve_stall_retry`. It looks like a
trick, and the reason is real: the mid-descent verdict that matters is the *winning* attempt's,
not the first one's, and when the `left_init` retry is adopted those differ.

The rest — the test changes, the changelog, the docs section — can be skimmed.

## Open questions

None outstanding. One was raised and settled: whether to restrict the restart by algorithm, the
way the #751 `left_init` retry is L-BFGS-only. It is deliberately **not** restricted. That
retry's remedy (the identity-Hessian cap) is an L-BFGS-specific step, so its gate follows the
algorithm; this one has no such tie — the condition is diagnosed from the objective trace, which
any NLopt algorithm can produce, and BOBYQA returns `Failure` too. Confirmed with @roninsightrx.


src/estimation/outer_optimizer.rs
```diff
@@ -1983,55 +1983,160 @@ impl BestPoint {
...
+        crate::cancel::is_cancelled(&options.cancel),
+        (first, mid_descent_stall),
+        |stalled| {
+            let (mut restarted, _) =
+                optimize_nlopt_once(model, population, &stalled.params, options, false);
+            // `n_iterations` reports the evaluations the fit spent, and the
+            // restart is a continuation of the same fit rather than a fresh one,
+            // so the two legs add up.
+            restarted.n_iterations += stalled.n_iterations;
+            restarted
```
roninsightrx, line 2114: [P2] Measure the reported escape verdict against the original fit start

`optimize_nlopt_once(..., &stalled.params, ...)` computes `restarted.left_init` relative to the restart point, and this closure returns that flag unchanged. If the first attempt moved substantially and the restart improves the objective with a displacement below `INIT_ESCAPE_STEP_S`, the adopted result reports `Some(false)` even though its estimates are far from the original initial values. `model_selection::stalled_at_init` prefers this flag over the original-value comparison, so postfit emits a false `W_STALLED_AT_INIT` warning and default `Strictness::reject_init_stall` rejects the recovered fit. Recompute the published escape verdict against the original start/scaling when adopting a restart, and add a regression where the first leg moved and the improving second leg makes only a small adjustment.

src/estimation/outer_optimizer.rs
```diff
@@ -1983,55 +1983,160 @@ impl BestPoint {
...
+    first: (T, bool),
+    restart: impl FnOnce(&T) -> T,
+    ofv_of: impl Fn(&T) -> f64,
+) -> T {
+    let (first, mid_descent_stall) = first;
+    if !mid_descent_stall || cancelled {
+        return first;
+    }
+    let second = restart(&first);
+    if !(ofv_of(&second) < ofv_of(&first)) {
```
roninsightrx: [P2] Reject an invalid restart objective before comparing attempts

A strict `<` comparison does not reject negative infinity: `resolve_mid_descent_restart(false, false, (3209.8087, true), |_| f64::NEG_INFINITY, |x: &f64| *x)` returns `-inf`. The first-attempt guard in `worth_restarting_mid_descent` does not validate the successor. `optimize_nlopt_once` can return an invalid final objective with `converged: false` after the cold/final solve; demoting convergence does not prevent this resolver from replacing the finite incumbent with that result, losing its usable estimates and diagnostics. Validate the successor objective before adoption (using the objective-validity rule already used for the initial restart gate), and cover NaN and both infinities in the adoption tests.

roninsightrx (commented):
Reviewed the optimizer control flow, result consumers, NN regression changes, and documentation. Found two correctness issues in restart-result handling; details inline. Current PR checks and both linked slow-test jobs are green. A focused test of the extracted restart resolver reproduces the non-finite-result issue (a -inf successor replaces a finite incumbent).

Local validation: `cargo test --offline --profile ci-fast --lib --features ferx-core/ci,ferx-core/nn mid_descent -- --nocapture` — 3 passed, 0 failed. The focused negative-infinity adoption regression fails as described in the second comment. The first finding follows the restart-relative escape calculation through the published `left_init` field to `stalled_at_init` and the default strictness gate.

src/estimation/outer_optimizer.rs
roninsightrx: Confirmed and fixed in d6c87178.

Traced it the way you describe and the consequence lands: `model_selection::stalled_at_init` returns `Some(!left)` from the flag at `src/model_selection.rs:511` before it ever reaches the `theta_init` comparison, so a recovered fit whose restart made a small adjustment would emit `W_STALLED_AT_INIT` and then be rejected by the default `Strictness::reject_init_stall`. Right estimates, thrown away.

`optimize_nlopt_once` now takes an `escape_from: Option<&ModelParameters>` that separates *where the attempt starts* from *what its published verdict is measured against*. `None` is every pre-existing caller, unchanged; the restart passes the original `init_params`. The reference is packed, clamped and scaled through the same `bounds`/`scale` as `x0_start_s`, so both points are compared in one consistent metric.

One thing I split rather than followed exactly: only the **published** verdict moves. `AttemptOutcome.left_init` (the #751 retry gate) and `failure_is_converged_plateau` keep this attempt's own start, because they ask a different question — "has the fit moved yet?" and "did it descend, or twitch and die?" are about the leg, not about the pair. Measuring those from the original start would make the restart leg's plateau verdict unconditionally permissive.

On the regression: I could not build the production geometry on the available fixture, and the three things that did not work are recorded in the test so nobody re-derives them. `outer_maxiter` does not hold BOBYQA still (its setup phase alone is `40 * (n + 1)` evaluations regardless of the budget); starting at the optimum does not either, because `make_model` is weakly identified — 4 free parameters on 4 identical subjects — and the optimizer wanders off the flat optimum by more than `INIT_ESCAPE_STEP_S`; and FIXing every parameter collapses the reference onto the start, since fixed parameters get degenerate bounds and the reference is clamped to them.

So `test_published_escape_verdict_uses_the_original_start` straddles the gate the *inverse* way round: the leg runs from the default start and is handed its own converged estimates as the reference, giving a large displacement from its own start and ~0 from the reference — the same two verdicts, opposite signs. The straddle is asserted, not assumed, so it cannot quietly become a tautology. Mutation-checked: reverting `published_left_init` to `left_init` flips it to `Some(true)` and the test fails.

src/estimation/outer_optimizer.rs
roninsightrx: Confirmed and fixed in d6c87178. Your repro is exactly right: `-inf < 3209.8087` is `true`, so the strict comparison screened `NaN` and `+inf` for free and let the one value through that actually matters.

The successor is now screened with `ofv_is_valid` — the same rule already gating whether a restart runs at all, rather than a second spelling of it — before the comparison:

```rust
let second_ofv = ofv_of(&second);
if !ofv_is_valid(second_ofv) || !(second_ofv < ofv_of(&first)) {
    return first;
}
```

Using `ofv_is_valid` rather than `is_finite` also screens the `DIVERGENCE_OFV` band, so a repelled subject's `1e20` cannot be adopted either. It would not have been, being larger, but the test pins it so the documented rule is the one in force.

I applied the same screen to `resolve_stall_retry`. It is latent there — that retry only runs on a fit that never left its start — but the two resolvers share one adoption rule, and having only one of them screen the successor is how they drift apart. Shout if you would rather I kept that out of this PR.

Tests: `test_mid_descent_restart_rejects_a_non_finite_successor` covers `-inf`, `+inf`, `NaN` and the `1e20` sentinel, and `test_stall_retry_rejects_a_non_finite_retry` does the same for the sibling. The `-inf` row is the one that fails without the screen — verified by mutation, not assumed; `NaN` and `+inf` are held so that a later edit flipping the comparison cannot quietly re-open them.

roninsightrx:
Both findings were real. Confirmed each against the code before changing anything, fixed both in d6c87178, replies inline.

**P2 — escape verdict measured from the restart's own start.** Confirmed downstream exactly as described: `model_selection::stalled_at_init` returns `Some(!left)` from the flag at `src/model_selection.rs:511` before reaching its `theta_init` comparison, so a recovered fit whose restart made a small adjustment emits `W_STALLED_AT_INIT` and is then rejected by the default `Strictness::reject_init_stall`. `optimize_nlopt_once` now takes `escape_from`, separating where an attempt starts from what its published verdict is measured against; the restart passes the original `init_params`. Only the published verdict moves — the #751 retry gate and `failure_is_converged_plateau` keep the leg's own start, since "has it moved yet?" and "did it descend or twitch and die?" are questions about the leg, not the pair.

**P2 — `-inf` successor adopted.** Confirmed with your repro. The successor is now screened with `ofv_is_valid` before the comparison — the rule already gating whether a restart runs at all, so there is one spelling of it rather than two. Applied to `resolve_stall_retry` too: latent there, but two resolvers sharing one adoption rule with only one screen is how they drift.

**On the regression you asked for.** The production geometry — first leg moved, improving second leg makes only a small adjustment — is not constructible on the available fixture, and the three approaches that fail are now recorded in the test so they are not re-derived: `outer_maxiter` does not hold BOBYQA still (setup phase alone is `40 * (n + 1)` evaluations); starting at the optimum does not either, since `make_model` is weakly identified (4 free parameters, 4 identical subjects) and the optimizer wanders off the flat optimum by more than `INIT_ESCAPE_STEP_S`; and FIXing every parameter collapses the reference onto the start, because fixed parameters get degenerate bounds and the reference is clamped to them. So the test straddles the gate the inverse way round — the leg is handed its own converged estimates as the reference, giving a large displacement from its own start and ~0 from the reference. Same two verdicts, opposite signs, and the straddle is asserted rather than assumed.

Every new assertion was mutation-checked rather than believed: reverting `published_left_init` to `left_init` flips the verdict and reddens the first test; dropping the `ofv_is_valid` screen reddens the `-inf` row of the second. Local: `--features ferx-core/ci --tests` 6404 passed / 0 failed, `tools/preflight.sh` green.

Issue #1395, opened 2026-09-15
bug(parser): `model = NAME` is silently ignored (only `model NAME` is read) — the fit falls back to the file stem with no diagnostic


TeunP:
## Problem

A `.ferx` file names its model with a bare `model NAME` line:

```
model mbma_naproxen

[parameters]
```

`model = mbma_naproxen`, the `KEY = value` spelling every other ferx setting uses, is silently ignored. There is no diagnostic, and the name falls back to the file stem.

## Evidence

- **Parser (read, not run):** `extract_model_name` in `src/parser/model_parser.rs` is `(?m)^\s*model\s+(\w+)`. After `model` it needs whitespace and then a word character, so `model = X` does not match and the name becomes `"Unnamed"`.
- **Downstream:** `set_model_name` (`src/api/run.rs`) replaces `"Unnamed"` with the model file's stem, and ferx-r does the same (`R/internal-fit-format.R`, which also carries a comment calling the engine's `"Unnamed"` a sentinel to remove long-term). So `fit$model_name`, the run summary header (`ferx run summary — …`) and ferx-r's `print`/`summary` show the file name, not the name the user wrote.
- **Measured:** `ferx check --json` reports `"model": "<file stem>"` for `model NAME`, `model = NAME` and no model line alike. The check report never uses the declared name, so it cannot surface the problem either.

## Also worth deciding

1. **The regex is not tied to a position.** It scans the whole file, so a line inside any block that starts with `model <word>` also sets the name.
2. **It is undocumented.** No page under `docs/model-file/` describes `model NAME`, and only 1 of 108 examples uses it (`examples/mbma_naproxen.ferx`). A ferx-r code comment (`R/internal-fit-format.R`) mentions a `[model_name]` block that does not exist.
3. **`check` and `fit` disagree.** The check report uses the file stem, the fit uses the declared name.

## Proposal

- Accept `model = NAME` as well, or reject it with an `E_PARSE` naming the accepted spelling. Rejecting fits #1377's "nothing is silently ignored" rule. Accepting is friendlier and matches every other `KEY = value`.
- Only honour the line before the first `[block]` header.
- Document it (`docs/model-file/` overview) and use the same name in `ferx check --json`'s `model` field.
- ferx-r: drop the `"Unnamed"` sentinel handling once the engine returns `""` or the stem, and fix that `[model_name]` comment. (#4 asked for the empty string and is closed, but the engine still returns `"Unnamed"`.)

[...]

PR #1417, opened 2026-09-17
fix(parser): read `model = NAME` as well as `model NAME`, from the preamble only, and report it from `ferx check` [closes #1395]

hiddevandebeek:
## Why

#1350 row 33 / #1395. A `.ferx` file names its model with `model NAME`; `model = NAME` — the `KEY = value` spelling every other setting uses — was silently ignored. `extract_model_name` was `(?m)^\s*model\s+(\w+)`, which needs whitespace after the keyword, so the name fell back to `Unnamed`, `set_model_name` (and ferx-r) replaced that with the file stem, and nothing said so. `ferx check --json` reported the stem for `model NAME`, `model = NAME` and no `model` line alike, so it could not surface the drop either. Same silent-drop family as #1048 / #1022 / #1377.

Two smaller holes in the same code: the regex scanned the whole file, so the first `model <word>` *anywhere* set the name; and `extract_blocks` skipped every in-block `model ` line in silence.

## What changed

- `parse_model_name` (replaces `extract_model_name`) reads the **preamble only** — the lines before the first `[block]` header. `model NAME`, `model = NAME`, `model=NAME` all name the model. A `model` line whose tail is neither form (`model = two words`, `model my-model`, `model =`, bare `model`) or a second `model` line is a parse error naming the accepted forms, with its line number. A line is a `model` line when its first word is exactly `model`; the rest of the preamble stays ignored, as it has to — `tests/nonmem/covariate_cat*.ferx` open with NONMEM-style `;` prose above their first header.
- `extract_blocks` no longer drops in-block `model ` lines: one inside `[parameters]` is that block's own unrecognised-text rejection (#1377). The `#`/`//` comment stripper the two readers share is one function now.
- `validate_model_file` resolves the report's `model` through the same `set_model_name` the fit entry points use — declared name, else the stem — so `ferx check` and `ferx run` report the same name for the same file. A parse failure keeps the stem (nothing was parsed).
- Documented under `docs/model-file/index.qmd` → *Model name* (the line was undocumented; 1 of 108 examples uses it).

Not changed: the `Unnamed` sentinel. ferx-r's `R/internal-fit-format.R` reads it (and `""`) and carries a comment about a `[model_name]` block that does not exist; retiring both is a ferx-r follow-up, not needed for this fix.

## Alternatives considered

- **Reject `model = NAME` with an `E_PARSE`** (the issue's other option). Accepting is what every other `KEY = value` does, and rejecting would still leave the whole-file scan and the check-report stem in place.
- **Reject all non-comment preamble text.** Would fire on the two in-tree NONMEM-anchor `.ferx` files above; out of scope.

## Cross-repo dependency

## Breaking changes
- [x] None. (A file with a malformed or duplicate preamble `model` line, or a `model X` line inside a block, now fails to parse where it used to be silently mis-named — no in-tree file has one; the 121 `.ferx` files under `examples/`, `tests/`, `docs/` were scanned.) No public-API change.

## Numerical validation
- [x] Not applicable (a label, nothing numerical)

## Performance

## Tests
- [x] Unit tests: `model_name_is_read_in_both_spellings`, `malformed_model_name_line_is_rejected_with_the_accepted_forms`, `a_second_model_name_line_is_rejected`, `a_model_line_inside_a_block_is_that_blocks_line_not_the_name`, `preamble_prose_and_model_prefixed_words_are_not_model_lines` (parser); `the_check_report_carries_the_declared_name_in_either_spelling`, `the_check_report_falls_back_to_the_stem_without_a_declaration`, `a_malformed_model_line_is_a_parse_error_in_the_report` (`api::validation`, temp file whose stem differs from the declared name so the two sources are told apart).
- [x] Regression tests that fail without the fix: the `=` arms of the first test against the old regex; the in-block straddle (`model X` before the header names, after it rejects) against a whole-file scan; the check-report test against the unconditional stem.

- [x] Inline — `docs/model-file/index.qmd` *Model name* shows both spellings.

## Docs
- [x] `docs/model-file/index.qmd` — new *Model name* section

## Checklist
- [x] No `Dual2` path touched

## Docs & examples
- [x] No example execution step needed (parser fix; no new DSL feature — the `model` line existed, one spelling of it was dropped)

## Reviewer hints

- `parse_model_name` and the two-line change in `extract_blocks` (`model ` no longer skipped) are the parser side; `validate_model_file`'s first 20 lines are the report side.
- The "first word is exactly `model`" rule is deliberate: `model:` and `modelled = 3` are preamble prose, not malformed declarations. The test pins both.

## Open questions

None. #1350 row 33 will be flipped ✅ with this PR number once merged.


hiddevandebeek:
## Review — PR #1417 (`model = NAME`, preamble-only, `ferx check` report)

Read every hunk, opened `extract_blocks` / `parse_data_block` / `validate_model_file` / `write_estimates_yaml` for context, ran the seven new tests (green) plus a throwaway probe that appends `model trailing` to the end of ten different blocks to see which block parsers actually reject it. Two findings.

### 1. `src/parser/model_parser.rs:13598` (medium) — a `model NAME` line after a `[data]` block is now dropped *silently*, and named by the stem

The PR's contract is "inside a block the line is that block's line and is parsed, or rejected, as such", and the `[parameters]` straddle test pins that for `[parameters]`. But `parse_data_block` (`src/parser/model_parser.rs:8270`) skips every line that is not `key = value` with a bare `continue`, so it never sees `model NAME`. Probe result, appending `  model trailing` as the last line of each block:

```
[fit_options] / [simulation] / [scaling] / [derived] / [covariates] / [priors] / [error_model] / [individual_parameters]  -> rejected
[data]  -> ACCEPTED, name = "Unnamed"
[odes]  -> ACCEPTED on a `pk` model (block is not read at all), name = "Unnamed"
```

Concrete scenario: a file that ends

```
[data]
  path = warfarin.csv

model warfarin_pk
```

parsed on `main` with `model.name == "warfarin_pk"` (old whole-file regex + the `model ` skip in `extract_blocks`). After this PR the line is handed to `[data]`, `[data]` ignores it, `parse_model_name` sees no preamble line, and the fit / `-fit.yaml` / `ferx check --json` all report the file stem — no diagnostic. That is the exact silent-drop shape #1395 exists to close, just moved from `model = NAME` to `model NAME`-after-`[data]`. (`model = NAME` inside `[data]` was already a `model -> NAME` column mapping before this PR, so that half is pre-existing.)

Two ways out: have `parse_data_block` reject a non-`key = value` line like every other block now does, or keep the `extract_blocks` drop but turn it into an error ("`model NAME` (line N) must be above the first `[block]` header") so a mis-placed declaration cannot be swallowed by whichever block is lenient. The first also closes the `[odes]`-on-a-`pk`-model hole for free, since it is the same class. Either way the PR's "Breaking changes: None" box is not quite right — a previously-valid `model NAME` below the first header now errors in eight blocks and is silently renamed in one.

### 2. `docs/model-file/index.qmd:65`, `CHANGELOG.md:162` (low) — `-fit.yaml` does not carry the model name

Both say the name is what "the `-fit.yaml` header" reports / that `-fit.yaml` "fell back to the file stem". `write_estimates_yaml` (`src/io/output.rs`) writes `model:` → `converged`, `method`, `uses_ode_solver`, … and never `result.model_name`; the only writer of the name is the `ferx run summary — {name}` line. What *does* use the stem is the output **filename** (`{stem}-fit.yaml`, `crates/ferx-cli/src/main.rs:362`), which is stem-derived regardless of any `model` line, and a reader of the new section could reasonably expect `model warfarin_pk` to produce `warfarin_pk-fit.yaml`. Worth naming the actual consumers (`fit$model_name`, the run summary, `ferx check --json`'s `model`) and saying the filename stays the stem.

Everything else I checked holds up: the two header regexes are byte-identical so the preamble boundary cannot drift from `extract_blocks`; `strip_line_comment` sharing is a straight lift; the `set_model_name` reuse in `validate_model_file` is correct and the parse-failure arm keeps the stem as intended; `;`-prose preambles and `modelled = 3` are correctly ignored.

hiddevandebeek:
Addressed both review findings in `4d353034` (then merged `main` in `6a7a5d3d`; changelog conflict was additive).

**Medium — silent drop relocated.** Confirmed: `parse_data_block` `continue`s past non-`key = value` lines, so `model NAME` after `[data]` parsed green and the model took the stem; `[odes]` on a `pk` model likewise. Fix: `extract_blocks` now rejects an exact declaration (`model NAME`, `model = NAME`, `model=NAME`) inside *any* block, named or unnamed, with one message naming the block and the preamble form — uniform instead of eight block-specific errors plus two silent renames. The matcher (`model_name_declaration`) is shared with `parse_model_name` so the two cannot disagree. Only the exact forms are caught: `model = CL * 2` in `[individual_parameters]` is still that block's line, as on `main`. Changelog now names this as a widening reject.

Test: `a_model_declaration_inside_any_block_is_rejected_naming_the_block` leads with the `[data]` and `[odes]` arms; under the mutation that drops the guard the `[data]` arm fails with "expected a parse error, but the model parsed" (verified), where the `[parameters]` arm alone would have reddened on the block's own message and not pinned the finding. Plus `[covariates WT]` for a named block, and a unit test that the matcher rejects `model = CL * 2` / `model = my model` / `modelled = 3` / bare `model`.

**Low — `-fit.yaml`.** Confirmed `write_estimates_yaml` writes no name. Docs and changelog now say the name reaches `fit$model_name`, the `ferx run` summary line, `-fit.json`'s `model_name` and `ferx check --json`, and that output filenames stay the file stem (`model warfarin_pk` in `run3.ferx` still writes `run3-fit.yaml`).

Preflight green (fmt, check matrix, clippy, rustdoc, docs, public-api); `parser::` + `api::` lib tests 1713 passed.

Issue #1153, opened 2026-08-31
A reset row's occasion is a label the IOV machinery does not know about (no κ group, and a fallback that should be deleted)


TeunP:
Two problems, one seam: **an EVID=3/4 row's occasion is a first-class label the rest of the IOV machinery does not know about.** #1133 made the reset row's own `OCC` authoritative (measured against NONMEM in `nonmem_anchor/reset_init_snapshot_J.ctl` — 42.0 under the reset row's occasion 2 against 14.0 under the preceding record's), but only `Subject::reset_occasions` and `pk::reset_row_occasion` learned that. Filing them together because fixing either one alone leaves the other's failure mode in place.

## A — a reset's occasion can have no κ group, and three consumers disagree about it

`stats::likelihood::iov_occasion_groups` builds the κ groups from **observation occasions ∪ `dose_occasions`**. Reset rows are not a source. Before #1133 that was fine: `reset_row_occasion` returned a *neighbouring record's* occasion, which is in the map by construction. Now it returns the row's own, which need not be.

A dataset whose EVID=3/4 row carries `OCC = 3` while no observation or dose row does is therefore reachable, and the three consumers of that label take three different paths:

| consumer | unmapped occasion → |
|---|---|
| `pk::predict_iov` (`combined_for`) | `_` arm → **κ = 0** |
| `sens::ode_provider::seed_iov_events`, per-event branch | `seed_pk_only_cov` → **κ = 0** ✓ agrees |
| `sens::ode_provider::seed_iov_events`, **static** branch | `occ_to_k.get(&occ)?` → `None` → **whole subject declines to FD** |

The static branch is at `src/sens/ode_provider.rs:3111`. The FD fallback is *correct* — FD differentiates the production predictor — so this is not a wrong gradient. It is a silent analytic→FD cliff on one branch only, on a dataset the other two handle analytically, with no warning anywhere. The comment immediately above that line says the branch was written to avoid exactly this ("an IOV subject with resets and no EVID=2 rows would newly be able to decline to FD"); the `?` reintroduces it.

Worth deciding explicitly rather than inheriting: **is κ = 0 even the right answer** for an occasion that carries no observation? It is defensible — such an occasion contributes nothing to the likelihood, so its EBE is the prior mean — but ferx currently arrives there by three different accidents and says nothing to the user.

## B — `reset_row_occasion`'s fallback should be deleted, not repaired

[...]

PR #1418, opened 2026-09-17
feat(estimation): count and report subjects that decline the analytic outer gradient [closes #1154]

### LINKED ISSUE #1154: A runtime analytic-sensitivity decline drops the whole population to FD silently, while the fit still reports "analytic"
A subject that falls out of the analytic sensitivity provider's scope at **runtime** drops the whole population to finite differences, and nothing anywhere says so — while the fit's own report continues to say the gradient is exact analytic.

## What is silent

`src/sens/provider.rs:1627` states the mechanism plainly:

> Out-of-scope ODE subjects return `None` → the population drops to FD, mirroring the analytical IOV gate (#439 ODE IOV).

So the decline is **per subject** but the consequence is **population-wide**: one subject in a thousand can take an entire fit off the analytic gradient. That is the intended safe behaviour — FD differentiates the production predictor, so the answer stays right — and it is what CLAUDE.md asks for ("a scope gap fails loudly to FD instead of silently returning a wrong gradient").

The problem is the word *loudly*. It is not loud:

* No `W_*` warning code for it exists — grepping `src/` for `W_SENS`, `W_GRADIENT`, `W_ANALYTIC`, `W_FD` returns nothing.
* Nothing counts how many subjects declined, or names one.
* `build_info::gradient_method_outer` (`src/build_info.rs:95`) takes `&CompiledModel`, not the population, and reports the **model-level** predicate `sens_supported` / `iov_sens_supported`. A fit whose every subject declined at runtime still reports `Analytic`.

The user-visible symptom is a fit that is several times slower than an equivalent one, with a report that says nothing changed. There is no way to tell "this model is out of analytic scope" from "one subject's data shape is" without reading the provider source.

## Why now

Found reviewing #1137. #1153 (section A) is a concrete instance: since #1133 a reset row's own `OCC` is authoritative, but `stats::likelihood::iov_occasion_groups` builds κ groups from observation ∪ dose occasions only, so a reset can carry an occasion with no group. `seed_iov_events`' static branch then does `occ_to_k.get(&occ)?` and returns `None` — and the whole population silently drops to FD, on a dataset the per-event branch and `predict_iov` both handle analytically with κ = 0.

That instance should be fixed in #1153. This issue is the class: **#1153 was only findable by reading the code**, because the runtime behaviour it produces is indistinguishable from "this fit is just slow".

## Suggested shape

Smallest useful version — count and report:

1. Count subjects whose sensitivity provider returned `None` during a fit.
2. When that count is non-zero while the model-level predicate says the model *is* in scope, push a warning naming the count and one example subject ID — the mismatch is the whole signal, and it is exactly the case that is currently invisible.
3. Make `gradient_method_outer`'s report reflect what actually ran rather than what the model permits, or mark it explicitly as the model-level route (its docstring already hedges with "reports the default in-scope route" — that hedge is doing a lot of work).

A per-subject reason string would be better still, but the count alone turns "mysteriously slow" into a one-line diagnosis and would have surfaced #1153 from a fit rather than from a code review.

Related: #1153 (an instance), #997 (§4, the optimizer/gradient coupling being silent — same family of "the fit did something you cannot see"), #486 (analytic sensitivity scope status).

roninsightrx:
## Why

Closes #1154.

A subject that falls out of the analytic sensitivity provider's scope at **runtime** is salvaged onto a per-subject reconverged-FD outer gradient. The answer stays correct — FD differentiates the production predictor — but the subject runs several times slower, and nothing said so:

* no `W_*` code for it (grepping `src/` for `W_SENS` / `W_GRADIENT` / `W_ANALYTIC` / `W_FD` returns nothing),
* nothing counted the declines or named a subject,
* `build_info::gradient_method_outer` takes `&CompiledModel`, so a fit whose every subject declined still reports `analytic (Dual2)`.

The user-visible symptom was a fit that is several times slower than an equivalent one, with a report that says nothing changed. #1153 is a concrete instance, and it was only findable by reading the provider source.

**One correction to the issue's premise.** #1154 says a single decline "drops the whole population to FD". That has not been true since #466: `population_gradient_sens_mixed` / `_iov_mixed` salvage per subject, and AGQ does the same via `accumulate_fixed_b_packed_gradient_fd` (#251). Three docstrings still described the old population-wide bail — `sens/provider.rs`, `estimation/agq.rs`, and the comment the issue quotes — and are corrected here. The *silence* the issue is about is entirely real; the blast radius is one subject, not the population.

## What changed

**`OuterFdDeclineLog`** (`estimation/outer_optimizer.rs`) — one `AtomicBool` per subject, set on the fallback arm of `population_gradient_sens_mixed` / `_iov_mixed`. Created in `optimize_population`, read once at the end, where the warning is pushed onto `OuterResult::warnings`:

```
1 of 32 subjects fell outside the analytic sensitivity provider's scope during this fit
(e.g. subject 7) and used finite-difference outer gradients; their results are correct
but slower. The reported outer gradient method is the model-level route, not the
per-subject one.
```

**Recording rather than predicting is what removes every gate**, and that absence is the part worth reviewing. A decline can only be recorded on an evaluation that actually happened, so each of these reaches the end with an empty log without any clause written to make it so:

| Route | Why the log stays empty |
|---|---|
| `[mixture]` + `optimizer = auto` | `resolve_outer_optimizer` downgrades to BOBYQA — derivative-free, no outer gradient |
| GN / GN-hybrid / trust-region | never call the mixed assembly |
| `reconverge_gradient_interval = 1` | `force_reconverge` skips the analytic branch on every eval |
| `outer_maxiter = 0` | short-circuits before the log is constructed |

Conversely any recorded decline is already a mismatch with the model-level label, since the analytic branch had to be selected for one to be recordable.

**`gradient_method_outer` is documented as the model-level route it has always been**, now also naming the two resolution rules it does not model (the mixture `Auto` downgrade and `reconverge_gradient_interval`), with an explicit note not to use it as a gate for "did the analytic outer gradient run".

**`subject_analytic_outer_gradient{,_iov}` extracted** so the FOCE/FOCEI entry-point choice and the finiteness backstop live in one place rather than at both mixed-assembly call sites.

## Alternatives considered

**Probing for declines instead of recording them.** This is what the first revision did, and PR review found it reports fallbacks that never happen. Kept here because the two measurements behind it are still the reason the final design looks the way it does.

*Probing the full outer assembly at `η = 0`* reported *"4 of 10 subjects use finite-difference outer gradients"* on the bundled warfarin fit, which uses the analytic gradient for every subject. Cause, instrumented rather than reasoned from the code: `prepare_stacked` requires the true inner Hessian to be positive-definite (`h_inner.cholesky()?`), which holds at the EBE and fails away from it — subjects 2, 4, 7 and 10 fail exactly that at `η = 0` (never `htilde`), while `subject_sensitivities` serves all 10.

*Probing only the provider* fixed that instance and not the class. The review's finding 1 is the general case: `moving_bounds_separable` reads the **resolved** infusion durations and lag times, so a provider decline is not a property of the subject's records either. With `D2 = TVD2 * exp(ETA_V1)` and `TVD1 = TVD2 = 2`, the identical subject declines at `η = [0, 0]` and is served at `η = [0, 0.4]`. Pinned as `provider_modeled_decline_depends_on_eta_not_only_on_the_records`.

No probe point is the fit's, so the diagnostic records instead.

**Where the runtime state lives.** A module-level static would be wrong under concurrent fits in one process (bootstrap, `ferx-tools`); `GradientTimings` gets away with it only because it is diagnostic-only. `OuterResult` already carries `warnings`, so threading a fit-local log through `optimize_population` needs no new field and no wider public API.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` clean), no new `FitResult` field — the warning rides the existing `warnings` vector, which ferx-r already surfaces.

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable — no numerical path is touched. `subject_analytic_outer_gradient{,_iov}` is a pure extraction of the existing per-subject dispatch: the same two entry points on the same `interaction` flag, the same `is_finite()` backstop, the same `subject_reconverged_fd_gradient` on `None`. The warning is read-only and runs before estimation.

Nothing new is anchored against NONMEM because nothing new is computed; the full `cargo test --features ci` run below is the regression evidence that the extraction is behaviour-preserving (it includes the `*_packed_gradient_matches_reconverged_fd` parity tests on both mixed assemblies).

## Performance
- [x] No significant impact expected — one provider evaluation per subject, once per fit, on gradient-driven methods only. The same order as the inner `fd_fallback_warning` that already runs `subject_eta_grad` per subject at the same point.

## Tests
- [x] Unit test(s) added — `estimation::outer_optimizer::tests::outer_fd_fallback` (6), plus `sens::provider::tests::provider_modeled_decline_depends_on_eta_not_only_on_the_records`
- [x] Regression test added that fails without this fix — 5 Tier-2 in `tests/outer_gradient_fd_fallback_warning.rs`, including both review reproducers

Fixture: warfarin with `f=F`, `F < 1`, where a **rate-defined** (`RATE > 0`) infusion reshapes the dosing window in a way the closed-form walk cannot express (#419), so that subject declines while every bolus-dosed one stays analytic.

Every fixture carries an asserted **straddle precondition** — `fixture_subjects_straddle_the_provider_gate` for the counting, a numeric disagreement between the two marginals for the entry-point test, and `gradient_method_outer == "analytic (Dual2)"` in the mixture test so it keeps reproducing the *mismatch* rather than passing because the label quietly changed.

Mutations, each reddening exactly one test and no others:

| Mutation | Test that dies |
|---|---|
| drop both `declines.record(i)` calls | `warns_with_count_and_example_subject`, `warns_when_every_subject_declines_on_an_in_scope_model` |
| drop **only the IOV** `record` | `fused_gradient_preserves_...` at `iov = true` |
| drop **only the non-IOV** `record` | `fused_gradient_preserves_...` at `iov = false` |
| report `subjects.first()` instead of the first *declining* subject | `warns_with_count_and_example_subject` |
| hard-code the FOCEI entry point in the shared gate | `follows_the_interaction_flag` |
| unwire the warning from `optimize_population` | `out_of_scope_subject_warns_while_the_report_still_says_analytic` |

**A hole that pass found, worth naming.** Dropping both `record` calls reddened the new tests; dropping only the **IOV** one left the entire suite green, because every new fixture was non-IOV — that arm was an assertion against a constant. Rather than add a parallel IOV fixture, the log assertion went into the existing `fused_gradient_preserves_analytic_and_subject_fallback_results`, which already sweeps `iov × mixed × interaction` and already knows which subject is the FD one. Each side now names itself in the failure message.

Results:
- `tools/preflight.sh` — green (fmt, 5 check feature sets, clippy, rustdoc, docs, public-API)
- `cargo test --features ci --lib` — 4764 passed, 0 failed, 26 ignored
- `cargo test --features ci` — 198 binaries, 5601 passed, 0 failed

- [x] Not applicable — no new API surface; the change is a warning on an existing fit.

## Docs
- [x] `docs/` updated — new `### The outer-loop route is reported as a warning {#outer-loop-fd-fallback}` in `docs/estimation/foce.qmd`, linked from that page's section list. States why `gradient_method_outer` is model-level, shows the warning text, and records that provider declines are counted while mode-dependent assembly declines are not.
- [x] `CHANGELOG.md` `[Unreleased]` → `Added`

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched; `sens/provider.rs` and `estimation/agq.rs` are comment-only here

## Docs & examples
- [x] No new `.ferx` syntax, `[fit_options]` key, example file or data file → no ferx-site / ferx-book change
- [x] No example execution step needed — no user-visible change to any bundled example's output beyond the new warning line

## Reviewer hints

Focus on **what is absent**: there is no route gate anywhere. The table under *What changed* is the argument that none is needed — a decline is only recordable on an evaluation that happened. If that reasoning is wrong for some route, the diagnostic is wrong for that route, and `an_untouched_log_says_nothing` is the single test the whole table rests on.

Second: `OuterFdDeclineLog` is threaded through `freeze_flat_thetas`, `optimize_nlopt{,_once}`, `optimize_bfgs`, `population_gradient{,_with_agq_evaluation}` and both mixed assemblies. All mechanical, but it is the bulk of the diff. Note the NLopt stall-retry runs `optimize_nlopt_once` twice against the same log, so declines union across both attempts — deliberate: a discarded retry still evaluated those gradients.

`subject_analytic_outer_gradient{,_iov}` can be skimmed; the only change since the last review is `declines.record(i)` on the arm that already routed to `subject_reconverged_fd_gradient`.

## Open questions

Should `gradient_method_outer` eventually carry the per-subject split the way the `gradient:` banner does for the inner loop (`analytic (Dual2) ×118, FD ×3`)? That would be strictly more informative than a warning, but it changes a persisted `FitResult` string field that ferx-r and fitrx both read, so it is a separate decision from making the fallback visible at all.


src/estimation/outer_optimizer.rs
```diff
@@ -3966,6 +3966,152 @@ fn subject_reconverged_fd_gradient_iov(
...
+                model,
+                subject,
+                &init_params.theta,
+                &stacked,
+            )
+            .is_some()
+        } else {
+            let zeros = vec![0.0; model.n_eta];
+            crate::sens::provider::subject_sensitivities(model, subject, &init_params.theta, &zeros)
+                .is_some()
```
roninsightrx: [P2] Count provider declines at the evaluated mode, rather than assuming the zero-eta probe is permanent

The provider itself has parameter-dependent declines: `moving_bounds_separable` uses the resolved infusion durations and lag times. I reproduced this with the existing two-cpt D1/D2 fixture, changing `D2 = TVD2` to `D2 = TVD2 * exp(ETA_V1)` and keeping TVD1 = TVD2 = 2. Two simultaneous RATE=-2 infusions coincide at eta = 0, so this probe emits "1 of 1 ... use finite-difference outer gradients". With observations generated at eta = [0, 0.4], the inner solve instead produces an EBE approximately [0.05768, 0.35800], where both `subject_sensitivities` and the complete `subject_packed_gradient` return a finite analytic gradient. The warning still appears on an outer_maxiter = 1 fit. Conversely, a provider served at the initial probe can decline later. Please collect actual declines in fit-local state, or restrict this diagnostic to parameter-independent scope failures and word any sampled result accordingly. Add a regression covering a provider decline that disappears at the EBE; moving the probe from the assembly to the provider alone does not eliminate the false-positive problem.

src/api/fit.rs
```diff
@@ -1290,6 +1290,46 @@ fn fit_inner(
...
+            m,
+            EstimationMethod::Foce | EstimationMethod::FoceI | EstimationMethod::Laplace
+        )
+    }) {
+        if crate::build_info::gradient_method_outer(
+            &crate::build_info::BUILD_INFO,
+            m,
+            options.optimizer,
+            model,
+        ) == crate::build_info::GradientMethodKind::Analytic
```
roninsightrx: [P2] Gate the warning on the optimizer that the stage actually runs

`gradient_method_outer` does not implement the mixture-specific optimizer resolution in `resolve_outer_optimizer`: for an in-scope model it classifies Auto as analytic, while a mixture fit with Auto actually runs derivative-free BOBYQA. I reproduced this by adding a two-class mixture to the F < 1 fixture, giving one of two subjects the rate-defined infusion, and running optimizer = auto with outer_maxiter = 1. The new warning says "1 of 2 ... use finite-difference outer gradients", although BOBYQA requests no outer gradient at all. Please use the stage's actual optimizer resolution, including `init_params.mixture`, before probing. The same gate should account for overrides such as reconverge_gradient_interval = 1, which bypass the analytic path for every subject rather than selecting a per-subject provider fallback. Add a mixture/Auto regression so a model-level reporting label cannot activate this warning for a derivative-free fit.

roninsightrx (commented):
Reviewed f43a2878. I recommend addressing the two findings below before merging: the new diagnostic can assert that subjects used finite-difference outer gradients when they did not.

Validation: the six new unit tests and two integration tests passed using the existing PR build artifacts. I also linked and ran two standalone reproducers against the PR debug library: one with parameter-dependent modeled infusion windows, and one with a mixture model and optimizer = auto. Both reproduced the false warnings. No numerical change was found in the extracted FOCE/FOCEI dispatch helpers.

roninsightrx:
Both findings confirmed and fixed. They turned out to have one root cause, so the fix is one change rather than two patches: **the diagnostic no longer predicts, it records.**

## What was wrong

The probe answered "would this subject decline, evaluated here?" and the warning presented that as "did this subject use an FD outer gradient?". Those are different questions, and the review found both ways they come apart:

* **Finding 1 — the parameter point is not the fit's.** I had convinced myself a provider decline was a property of the subject's records. It is not: `moving_bounds_separable` reads the *resolved* infusion durations and lag times. Reproduced your repro as a unit test — `provider_modeled_decline_depends_on_eta_not_only_on_the_records` in `src/sens/provider_tests.rs`. With `D2 = TVD2 * exp(ETA_V1)` and `TVD1 = TVD2 = 2`, the identical subject declines at `η = [0, 0]` and is served at `η = [0, 0.4]`. Nothing about the records changed; only where they were evaluated.
* **Finding 2 — the optimizer is not the one that ran.** `gradient_method_outer` resolves through `Optimizer::resolve_auto`, which has no `[mixture]` branch, while `resolve_outer_optimizer` downgrades `Auto` → BOBYQA for a mixture model. Gating on that label meant gating on a route the fit did not take.

## The fix

`OuterFdDeclineLog` — one `AtomicBool` per subject, set on the fallback arm of `population_gradient_sens_mixed` / `_iov_mixed`, created in `optimize_population` and read once at the end, where the warning is pushed onto `OuterResult::warnings`. No new public API: `OuterResult` already carries `warnings`, so the persisted surface and `api/ferx-core-public-api.txt` are unchanged.

The probe and its whole "which point do we evaluate at" problem are gone, and with them every gate — which is the part I'd flag for re-review, because it is a load-bearing *absence*. A decline can only be recorded on an evaluation that happened, so each of these reaches the end with an empty log and says nothing, without anything being written to make it so:

| Route | Why the log stays empty |
|---|---|
| mixture + `optimizer = auto` | BOBYQA — derivative-free, no outer gradient (your finding 2) |
| GN / GN-hybrid / trust-region | never call the mixed assembly |
| `reconverge_gradient_interval = 1` | `force_reconverge` skips the analytic branch on every eval |
| `outer_maxiter = 0` | short-circuits before the log is even constructed |

You also flagged `reconverge_gradient_interval = 1` explicitly. I had considered and deliberately punted on it in the first revision — you were right that it belonged in scope, and it now falls out for free rather than needing a clause.

The `fit_inner` block is deleted; a comment there records why the warning cannot be emitted at that point.

## Tests

Your two reproducers are now regressions, plus the "decline that disappears at the EBE" you asked for:

* `provider_modeled_decline_depends_on_eta_not_only_on_the_records` (Tier 1) — the η-dependence itself, next to the existing θ-dependence test.
* `a_derivative_free_mixture_auto_fit_does_not_warn` (Tier 2) — your mixture/Auto repro, asserting `gradient_method_outer == "analytic (Dual2)"` as a fixture precondition so it keeps reproducing the *mismatch* and cannot pass because the label quietly changed.
* `forcing_the_reconverged_gradient_does_not_warn_about_scope`, `an_evaluation_only_fit_does_not_warn` (Tier 2).
* `an_untouched_log_says_nothing` (Tier 1) — the property the table above rests on, stated once.

The Tier-2 fits moved from `outer_maxiter = 0` to `1`. At `0` no outer gradient runs at all, so the original wiring test would have passed for the wrong reason under the new design; the module docs now say so.

**One hole the mutation pass found that I'd have shipped.** Dropping *both* `declines.record(i)` calls reddened my new tests; dropping only the **IOV** one left the entire suite green — every fixture I had written was non-IOV, so that arm was an assertion against a constant. Rather than add a parallel IOV fixture I moved the assertion into the existing `fused_gradient_preserves_analytic_and_subject_fallback_results`, which already sweeps `iov × mixed × interaction` and already knows which subject is the FD one. Each side now names itself on failure:

```
iov = true,  mixed = true, interaction = false: the decline log must name exactly ...
iov = false, mixed = true, interaction = false: the decline log must name exactly ...
```

## Unchanged, and re-verified

`subject_analytic_outer_gradient{,_iov}` is still the pure extraction of the FOCE/FOCEI dispatch you found no numerical change in; the only edit is `declines.record(i)` on the arm that already routed to `subject_reconverged_fd_gradient`. `population_packed_gradient_*_matches_reconverged_fd` and the `fused_gradient_*` bit-equality assertions cover that.

Gates: `tools/preflight.sh` green, `cargo test --features ci --lib` 4764 passed / 0 failed, full `cargo test --features ci` green.

PR #1419, opened 2026-09-17
fix(api): let an explicit CLI --threads beat [fit_options] threads [closes #1416]

### LINKED ISSUE #1416: CLI --threads is silently ignored when the model file sets [fit_options] threads
# CLI `--threads` is silently ignored when the model file sets `[fit_options] threads`

**Status:** open, unassigned
**Area:** CLI argument handling / `FitOptions` merge (`crates/ferx-cli`)
**Severity:** low blast radius, high consequence — it silently invalidates benchmarks
**ferx version observed:** 0.4.0

## What happens

A model file containing

```
[fit_options]
  method  = focei
  threads = 8
```

run as

```sh
ferx model.ferx --threads 1
```

starts on **8** threads:

```
Starting estimation (chain: FOCEI) on 8 threads...
```

and the fit YAML records `estimation.n_threads_used: 8`. No warning is emitted.
Removing the `threads` key from the model file makes the same command report
`on 1 thread` and `n_threads_used: 1`.

## Why it is wrong

* **`ferx --help` presents `--threads` as the control:** *"--threads N  use N rayon
  workers (N > 0)"*. Nothing says a model file can veto it.
* **The documentation implies the flag overrides.** `docs/model-file/fit-options.qmd`
  describes the key and adds *"The CLI's `--threads` flag mirrors this key."*
* **It is inconsistent with the R front end.** `docs/model-file/fit-options.qmd`
  (ODE tolerance precedence) states the rule for `ferx_fit(settings = ...)`:
  every key the caller names wins over the model file, *including* one named with
  its default value. The CLI behaving the other way round means the same model
  file plus the same requested thread count gives different answers depending on
  which front end you use.
* **`--data` already establishes the expected precedence**, and does it loudly:
  an explicit CLI path beats the `[data]` block and emits
  `W_DATA_PATH_OVERRIDE` when the two disagree. `--threads` should behave the
  same way, with or without the warning.

## How it was found

Benchmarking ferx against NONMEM and nlmixr2 on four population PK models, with
every engine pinned to one core. The ferx models carried `threads = 8` from
earlier exploratory runs. Every run invoked with `--threads 1` in fact used 8,
and the resulting "single-core" comparison was wrong by roughly 5× — melphalan
FOCEI is 7.05 s on 8 threads and 35.0 s on 1.

Nothing in the console output, the warnings list, or the run summary contradicted
the `--threads 1` that had been asked for; the only evidence was the
`Starting estimation ... on 8 threads` line and `n_threads_used` in the fit YAML,
neither of which a caller has reason to re-read when they have just pinned the
value on the command line.

That is the real cost here: a thread count is exactly the sort of setting that is
pinned once, trusted, and never re-checked, and getting it silently wrong does not
make a fit fail — it makes every timing number quietly meaningless.

## Suggested fix

Make an explicitly-passed `--threads` win over `[fit_options] threads`, matching
`--data` and the R front end. Two details worth deciding deliberately:

1. **Distinguish "not passed" from "passed as the default".** `--threads 0` /
   `--threads auto` are spellings of the default, and a caller who types one
   means it — so the merge should key on whether the flag appeared in `argv`,
   not on whether its value differs from the default. (This is the same
   distinction the R front end already draws, and the opposite of the Rust
   `FitOptions::default()` field-comparison rule; the CLI is a caller, so it
   should follow the R rule.)
2. **Warn on conflict**, as `--data` does: a one-line
   `thread count overridden: using 1 instead of the model's [fit_options] threads = 8`
   would have made this self-diagnosing.

## Reproducing

```sh
printf '[fit_options]\n  method = focei\n  threads = 8\n' >> model.ferx
ferx model.ferx --threads 1 | grep -i thread
# expected: on 1 thread
# actual:   on 8 threads
grep n_threads_used model-fit.yaml   # 8
```

Any model in `ferx-testdata` reproduces it; the four under
`{melphalan_dvorak,clofarabine_brooks,thiotepa_brooks,fludarabine_brooks}/ferx`
carried the key until it was stripped for the benchmark.

## Related

Worth auditing the other `[fit_options]` keys the CLI can also set for the same
asymmetry — `--inits-from-nca` at least, and whatever else the CLI surfaces as a
flag.

roninsightrx:
## Why

A model file carrying

```
[fit_options]
  method  = focei
  threads = 8
```

run as `ferx model.ferx --threads 1` started on **8** threads, printed
`Starting estimation (chain: FOCEI) on 8 threads...` and wrote
`n_threads_used: 8` to the fit YAML. No warning. The CLI sizes rayon's *global*
pool from the flag, but a positive `[fit_options] threads` exclusively leases its
own pool for the fit, so the model file silently won.

That is the wrong way round on three counts: `ferx --help` presents `--threads`
as the control; `docs/model-file/fit-options.qmd` said the flag "mirrors this
key"; and the R front end already draws the opposite rule — every key
`ferx_fit(settings = ...)` names wins over the file. `--data` also already
establishes the precedence, and does it loudly (it beats the `[data]` block with
a warning when the two differ).

The cost is that a thread count is pinned once, trusted, and never re-checked,
and getting it silently wrong does not fail a fit — it makes every timing number
meaningless. Found benchmarking four population PK models against NONMEM and
nlmixr2 with every engine pinned to one core: the ferx models carried
`threads = 8` from earlier exploratory runs, every run invoked `--threads 1` and
in fact used 8, and melphalan FOCEI is 7.05 s on 8 threads against 35.0 s on 1 —
so the "single-core" comparison was off by roughly 5×. Nothing in the console
output, the warnings list or the run summary contradicted the flag.

Closes #1416.

## What changed

**Precedence.** An explicitly passed `--threads` now overrides
`[fit_options] threads`, and a disagreement is reported:

```
  * thread count overridden: using 1 instead of the model's `[fit_options] threads = 8`
```

**"Not passed" vs "passed as the default."** `parse_threads_flag` used to
collapse *absent*, `0` and `auto` into `None`. It now returns `None` only for an
absent flag; `0`/`auto` return `Some(0)` — a caller who typed one asked for the
engine's own worker count and means it, so it overrides a pinned `threads = 8`
too. The merge keys on whether the flag appeared in `argv`, **not** on whether
its value differs from `FitOptions::default()`. That is the R front end's rule
and the opposite of the Rust `FitOptions::default()` field-comparison rule; the
CLI is a caller, so it follows the R one. `Some(0)` still sizes nothing globally
(`configure_global_thread_pool` rejects `0`) — it only stops the file pinning a
width.

**Plumbing.** A new `RunOverrides { inits_from_nca, threads }` carries what the
command line named. `run_model_with_overrides` and
`run_model_simulate_with_overrides` are the entry points that apply it;
`run_model_with_data_inits` is now a thin wrapper over the first, so there is
still exactly one code path from a `.ferx` file to a fit. Both production callers
share one `apply_threads_override` helper rather than carrying twin merges.

**The rest of the CLI's flags, audited** (the issue's "Related"):
`--inits-from-nca` already overrode the file (it is passed into
`prepare_run_with_inits`). The subcommands that take their own `--threads`
(`bootstrap`, `covsearch`/`modelsearch`/`iivsearch`/… via `search::Runner`) set
`FitOptions::threads` themselves on every inner fit — `replicate_options` pins
`Some(1)`, the search fitter pins `Some(threads_per_fit)` — so a model file
cannot veto them either. `--threads` on the single-fit path was the only
asymmetry.

## Alternatives considered

- **Make the engine prefer an explicitly-configured global pool over
  `FitOptions::threads`.** Rejected: it changes precedence for every library
  caller (including ferx-r), to fix a CLI merge. The override belongs at the
  front end that has the argv.
- **Have the CLI call `prepare_run_with_inits` and `fit()` itself, mutating
  `fit_options` in between.** Rejected: it duplicates ~40 lines of the entry
  point (the checkpoint wiring, the `Model:`/`Data:` lines, the data-path
  warning, `derive_output_occasions`) — a second copy of the thing the two would
  have to agree on.
- **Warn or not?** The issue left it open. Warned, as `--data` does: it makes the
  situation self-diagnosing, and it is the piece that would have caught the
  original benchmark.

## Cross-repo dependency

ferx-r builds its own `FitOptions` and already applies the caller-wins rule; it
does not call `run_model_with_data_inits`. The added items are additive, so
nothing downstream breaks.

## Breaking changes
- [x] Public Rust API (`api/`, `types.rs`) — baseline regenerated with `tools/update-public-api.sh`

Three items added, all needed by `ferx-cli`, which consumes `ferx-core` as an
ordinary external crate exactly as ferx-r does:

| Added | Needed by | Why the existing surface does not suffice |
|---|---|---|
| `RunOverrides` | `ferx-cli` `main.rs` | the type the front end's explicit choices travel in; a bare `Option<usize>` parameter would have broken `run_model_with_data_inits`'s signature, and the struct is where the next flag goes |
| `run_model_with_overrides` | `ferx-cli` `main.rs` | the fit entry point that applies them; `run_model_with_data_inits` takes only the NCA override and is kept as a wrapper for compatibility |
| `run_model_simulate_with_overrides` | `ferx-cli` `main.rs` | same, for `--simulate`, which reaches `fit()` through its own entry point |

`apply_threads_override` stays `pub(crate)` — the merge rule is ferx-core's, not
a caller's.

## Numerical validation
- [x] Not applicable — this changes how many rayon workers a fit uses, not what
  it computes. A run that does not pass `--threads` is unchanged (verified: the
  Tier-2 control arms and the warfarin repro below both still report the model
  file's count).

Reproduction from the issue, against this branch's release binary (warfarin +
`threads = 8` in `[fit_options]`):

```
$ ferx model.ferx --data warfarin.csv --threads 1
Starting estimation (chain: FOCE) on 1 thread...
  * thread count overridden: using 1 instead of the model's `[fit_options] threads = 8`
$ grep n_threads_used model-fit.yaml
  n_threads_used: 1

$ ferx model.ferx --data warfarin.csv            # no flag — the file still rules
Starting estimation (chain: FOCE) on 8 threads...

$ ferx model.ferx --data warfarin.csv --threads auto
Starting estimation (chain: FOCE) on 8 threads...
  * thread count overridden: using the default worker count instead of the model's `[fit_options] threads = 8`
```

(The `auto` run lands on 8 because this machine's default *is* 8 — cores − 1,
capped at 8. It is the override that is being shown, not the number.)

## Performance
- [x] No significant impact expected — for a run that passes `--threads`, the
  fit now uses the requested width, which is the point. Unflagged runs are
  unchanged.

## Tests
- [x] Unit test(s) added (`src/api/tests/threads_override_tests.rs`, 6 tests)
- [x] Regression test added that fails without this fix (`tests/threads_override.rs`, Tier 2)

Tier 1 pins the merge rule: the override lands; an *absent* flag leaves the file
in charge (the mutation guard for an unconditional override — if it reddens,
`[fit_options] threads` has stopped working at all); `Some(0)` overrides a pinned
`8` and is reported by name ("the default worker count"), which is the assertion
a "differs from the default" implementation fails; agreement and an unpinned file
are not conflicts; and the message classifies as `WarningCode::Threads` /
`Warning` rather than falling through to `general`.

Tier 2 pins that the merged value is what `fit()` actually sizes its pool from,
asserting `FitResult::n_threads_used` — the same readout the bug report caught
the defect with — for **both** production callers, each as a differential pair
(no override → the file's 2; override → 1). Without the control arm a fixture
that never honoured `[fit_options] threads` would pass the override arm for the
wrong reason.

**Mutation-checked per side**, since the two callers are a twin: reverting the
data-path merge (`overrides.threads` → `None` at `run.rs:350`) reddens
`the_data_path_honours_...` and leaves the simulate test green; reverting the
simulate merge (`run.rs:923`) reddens `simulate_honours_...` and leaves the other
green. Each failure names its own side.

`cargo test --lib --features ci`: 4763 passed, 0 failed.
`cargo test -p ferx-cli --features ferx-core/ci`: 34 passed, 0 failed.

- [x] Not applicable — no new API surface for model authors; the behaviour change
  is on an existing flag and is shown in the transcript above.

## Docs
- [x] `docs/` updated — `docs/model-file/fit-options.qmd`: the `threads` row now
  states the precedence (and that `0`/`auto` count as explicit), and the
  "Where the value comes from" section gains a **CLI** paragraph next to the
  existing R and Rust ones, since the CLI follows the R rule rather than the Rust
  one. `ferx --help` says it too.

## Checklist
- [ ] `Cargo.toml` version bumped — not breaking; the API additions are additive

### ferx-site
- [ ] No new `.ferx` DSL syntax or `[fit_options]` key — an existing key's precedence against an existing flag
- [ ] No new example `.ferx` or data file

### ferx-book
- [ ] No estimator / DSL / fit-option addition

### Example execution
- [x] Rebuilt `ferx` release binary and ran `examples/warfarin.ferx` (plus the
  `threads = 8` variant above) through the CLI on `data/warfarin.csv`
- [x] No site/book page renders needed (no example changed)

## Reviewer hints

Focus on two things:

1. **`apply_threads_override`'s conflict predicate** (`src/api/run.rs`). The
   override is unconditional whenever the caller named the key; only the
   *warning* is conditional, and only on a **positive** model-file count that
   differs. A file with `threads = 0` (or no key) pinned nothing, so there is no
   second opinion to report — but the override still lands.
2. **`parse_threads_flag`'s `Some(0)`.** This is the load-bearing change on the
   CLI side and the easy thing to "simplify" back to `None`. Two existing tests
   were updated to pin it (`zero_means_the_default_was_asked_for_by_name`,
   `auto_means_the_default_was_asked_for_by_name`), paired with
   `absent_flag_is_none` for the other half.

Skimmable: the CHANGELOG/docs prose, and the mechanical
`run_model_with_data_inits` → `run_model_with_overrides` delegation.

## Open questions

Severity of the new warning is `Warning`, not `Info`, matching the `--data`
override — two inputs disagreed and one was dropped. `WarningCode::Threads`'s
other messages ("N threads configured", "more threads than subjects") are `Info`,
so the code now spans both severities. Say if you would rather it were `Info`.


roninsightrx:
BLOCKING

- tests/threads_override.rs:106 - Both production-path tests exercise only `Some(1)`, so neither detects a caller-specific regression that collapses explicit `Some(0)` into absence.
  - Failure scenario: model has `threads = 2`; either caller filters `Some(0)` to `None` before `apply_threads_override` -> `--threads 0`/`auto` runs with 2 threads and emits no warning, while the parser and helper unit tests remain green.
  - Confidence: confirmed by reading

- crates/ferx-cli/src/main.rs:1011 - The CLI tests stop at parsing and never verify that the parsed value reaches `RunOverrides` and the selected data/simulation entry point.
  - Failure scenario: `main` assigns `threads: None` or calls a legacy runner -> `ferx model.ferx --threads 1` with file `threads = 2` still runs on 2 threads, while all parser tests and direct API integration tests remain green.
  - Confidence: confirmed by reading

roninsightrx:
Both findings are real and both are now fixed in `f23562eb`. They are the same shape — an assertion that could not observe the defect it existed for — and neither was reachable by mutating the fix itself, which is exactly the gap the repo's own rule warns about ("mutating your own fix is not enough — ask what the *smallest* edit that removes it is, and which test dies"). Here the answer was "none".

### 1. Every test used `Some(1)`, so a per-caller `Some(0) → None` collapse was invisible

Confirmed. `Some(0)` and `None` mean the same thing to `FitOptions::threads`, so `.filter(|&n| n > 0)` on the way in reads as a harmless tidy-up — and it silently restores the bug for `--threads 0` / `auto`, which is precisely the spelling the merge rule exists to distinguish.

Added `an_explicitly_requested_default_overrides_the_model_file_on_both_callers` (`tests/threads_override.rs`), asserting the override lands for `Some(0)` on **both** production callers. Mutation-checked per side, as the twin rule requires:

| mutation | result |
|---|---|
| `.filter(\|&n\| n > 0)` at `src/api/run.rs:350` (data path) | red — *"`Some(0)` must override the data path's model file too"* |
| same at `src/api/run.rs:923` (simulate) | red — *"`Some(0)` must override the simulate path's model file too"* |

It asserts the **warning**, not `n_threads_used`: the engine default is a property of the host's core count and can coincide with the fixture's `2` on a small machine, which would make the assertion a tautology on some runners. The warning fires iff the override landed.

### 2. Nothing covered the CLI wiring between `parse_threads_flag` and the entry point

Also confirmed, and the more embarrassing one: the parser tests stopped at the parse, the library tests started after `RunOverrides`, and the ~10 lines between them — where the bug actually lived — had no coverage at all. `main` assigning `threads: None` was green everywhere.

Added two end-to-end tests in `crates/ferx-cli/tests/cli_ferx.rs` that run the real binary against a model file pinned to `threads = 2`:

- `an_explicit_threads_flag_beats_the_model_files_thread_count` — with no flag, stderr must still say `on 2 threads` and the YAML `n_threads_used: 2` (the control arm, green under the old behaviour, which is what makes the other arm mean something); with `--threads 1`, stderr must say `on 1 thread`, the YAML `n_threads_used: 1`, and the override warning must appear on the console. These are the same two readouts the original bug report caught the defect with.
- `an_explicitly_requested_default_also_beats_the_model_files_thread_count` — `--threads 0` and `--threads auto` must each override the pinned count.

Mutation-checked: `threads: None` in the `RunOverrides` literal in `main.rs` reddens both.

### On the rest of the review

Findings 1, 3, 4 and 5 came back clean, and I'm reading that as weak evidence rather than confirmation, per the review's own caveat that nothing was executed. For the record, on this branch: `cargo test --lib --features ci` 4763 passed, `cargo test -p ferx-cli --features ferx-core/ci` 36 passed, `tools/preflight.sh` green (fmt / 5 check feature sets / clippy `--all-targets` / rustdoc / docs lint / public-API baseline).

roninsightrx (commented):
Reviewed commit `f23562ebfc159fd30ee1b7bdbb7c92cf248d5aae`. No actionable correctness or regression findings.

Checked the CLI parser and routing, both file-based fit entry points, the fit-pool selection for single-start and multi-start fits, and warning classification/rebuilding. An absent flag preserves the model-file setting; explicit positive counts and `0`/`auto` override it. Both entry points retain the conflict warning on the raw and structured result surfaces. The public API baseline, docs, and changelog match the implementation.

The additional regression tests cover the previously missing explicit-default and CLI-to-library wiring cases.

Local validation passed (all with `ferx-core/ci`):

- `cargo test --offline --lib threads_override_tests`: 6 passed.
- `cargo test --offline --test threads_override`: 3 passed.
- `cargo test --offline -p ferx-cli --bin ferx`: 132 passed.
- `cargo test --offline -p ferx-cli --test cli_ferx an_explicit`: 2 passed.
- `git diff --check`: clean.

The full suite was not rerun locally. At the last CI check, completed checks were passing; the core coverage job was still running and the slow-tests job was skipped.

Issue #1415, opened 2026-09-17
SAEM: thetas with no ETA barely leave their initial values


roninsightrx:
# SAEM: thetas with no ETA barely leave their initial values

**Status:** open, unassigned — written up for another agent to pick up
**Area:** `src/estimation/saem.rs` (numerical θ/σ M-step, `mstep_sa_step` / `damp_mstep`)
**Related:** #1011 (SA damping of the numerical M-step), #619 (covariate mu-references),
the no-ETA advisory in `src/estimation/impmap.rs:770`
**Reproducer:** `ferx-testdata`, three real models, committed and runnable

## Summary

An estimated theta that carries no random effect is moved only by the η-frozen
numerical M-step. ferx already warns about this and calls the channel "biased".
On three real models the behaviour is sharper and easier to test than "biased":
those thetas **travel a small fraction of the distance to the optimum and stop**,
in several cases never meaningfully leaving their initial value, while the
mu-referenced thetas in the same fit move normally.

The result is a SAEM fit that is much worse than the same model's own FOCEI fit,
and worse than NONMEM's SAEM on the same data from the same starting values.

## Evidence

Same model file, same data, same initial estimates; `method = focei` against
`method = [saem, imp]` with `imp_eval_only = true`. `moved%` is how far the SAEM
estimate travelled from the initial value as a percentage of the distance FOCEI
travelled — 100% would mean SAEM landed where FOCEI did, 0% means it never left
the start.

### thiotepa (`thiotepa_brooks`, 91 subjects, 435 obs, 15 parameters)

FOCEI OFV 5832.01 · SAEM OFV 6173.33 · **ΔOFV +341.3**

| theta   | init | FOCEI  | SAEM   | moved% | has ETA |
|---------|-----:|-------:|-------:|-------:|:-------:|
| TVCL    |  8   |  6.387 |  8.397 |  −24.6 | yes     |
| TVVMAX  |  5   |  2.507 |  1.580 |  137.2 | yes     |
| TVQ     |  2   |  1.390 |  1.981 |    3.1 | **no**  |
| TVVP    | 10   |  5.347 | 11.03  |  −22.0 | **no**  |
| TVCLM   |  0.5 |  1.542 |  0.609 |   10.5 | **no**  |
| TVKM    |  1   |  0.336 |  1.011 |   −1.7 | **no**  |
| HILL    |  1   |  1.762 |  1.006 |    0.9 | **no**  |
[...]

PR #1420, opened 2026-09-18
fix(estimation): let SAEM estimate thetas with no ETA instead of holding them at their start [closes #1415]

roninsightrx:
## Why

#1415: a theta with no ETA is moved only by the η-frozen numerical M-step, and on three real models (`ferx-testdata` thiotepa, melphalan, clofarabine) those thetas travelled a small fraction of the way to the optimum and stopped — eight of eight on thiotepa moved ≤ 13 % of the FOCEI distance, four within 2 % of their start — while the mu-referenced thetas in the same fit moved normally.

Instrumenting the M-step (per-M-step raw maximiser vs a converged solve from the same point) found two compounding defects and one reporting defect:

1. **The M-step solver was misconfigured.** `theta_sigma_mstep_light` ran BOBYQA from NLopt's *default* first design — a quarter of the bound range, i.e. 2.5 log units for a theta on `(0.01, 200)` and 3.25 for σ on `[-8, 5]` — so its first `2n+1` interpolation points sat at ×12–×26 of the current value, and it stopped on `ftol_rel = 1e-4` (0.1 NLL units on a ~1e3 objective). Measured on thiotepa: the returned point improved the conditional objective by 0.05–6 units per M-step where a converged solve from the same point improved it by 13–224, and several coordinates came back with a displacement of exactly zero. On melphalan 300 of 304 M-steps stopped `FtolReached` after a fraction of the move.
2. **The #1011 exploration cap (0.03) is a hold, not a cure.** With the solver repaired, the cap still freezes every no-ETA theta: it divides the number of EM steps the exploration phase amounts to (50 M-steps × cap), and a covariate slope confounded with an ETA needs all of them. On #1011's own FREM reprex the capped fit ends where it starts — 0.383 → 0.361, and from a deliberately wrong 0.2 start, 0.2 → 0.191. Its "good" answer was the model's initial estimate (that auto-generated FREM model took its inits from a FOCEI fit).
3. **`method = [saem, imp]` reported a FOCE objective *without* interaction.** The parser cleared `interaction` for every chain not ending in `focei`, while `method = saem` alone kept it. On thiotepa that is a 175-unit difference at the same estimates, and it was the bulk of the "SAEM is 341 units worse than FOCEI" in the issue.

## What changed

- `theta_sigma_mstep_light`: `set_initial_step(0.1 / scale)` (`MSTEP_INITIAL_STEP`) and `ftol_rel = 1e-7` (`MSTEP_FTOL_REL`). The `maxeval` budget is unchanged — a 3× budget bought nothing (thiotepa IS −2 log L 5835.0 vs 5827.3 at 2.3× wall).
- `mstep_damping` default 0.03 → **1.0 (off)** — assignment in both phases, as before #1011 — **except for a model with `iiv_on_ruv`, which keeps #1011's 0.03** (`default_mstep_damping`, keyed on `model.residual_error_eta`). The drift #1011 fixed is the `iiv_on_ruv` coupling, not FREM and not the no-ETA channel: the #1011 reprex with its `iiv_on_ruv` line and `ETA_RUV` omega removed lands *undamped* at `TVFRD1` 0.414 / `TVMAT` 2.680 / `TVV` 137.6, on NONMEM IMP's 0.394 / 2.680 / 133.8, where the cap leaves it at 0.342 / 2.228 / 147.5. With `iiv_on_ruv` back in, undamped drifts (0.18, 47 FOCEI units worse than the held 0.36). So #1014's fix is preserved exactly for the shape it was measured on, and every other model runs undamped. The option, its gate and its no-effect warnings are unchanged; a value you set wins either way.
- **Mixture models keep the pre-#1415 solver configuration** (`MSTEP_MIXTURE_FTOL_REL`, no `set_initial_step`), recorded at the call site. Their class typical values come from a hard class draw (#996), and the NONMEM-anchored seed sweep in `tests/mixture_nonmem.rs` was calibrated on the partial step the old configuration returns: with the new settings the 10-seed mean `TVCL2` moves from the NONMEM MLE 2.84 to 3.23 (3.09 with the first design alone; the tolerance alone moves `p(1)` 0.54 → 0.44 and the IMP marginal 300.9 → 306.9). An exact per-class M-step on a hard class draw is a different estimator (classification EM), so that channel changes only with its own schedule and anchor — same reasoning as the existing mixture veto on the damping. The unit test pins the asymmetry so re-unifying the arms is a deliberate change.
- Parser: a chain sets `interaction` only from a final `focei` (on) or `foce` (off), matching the single-method form.
- No-ETA advisory reworded (no longer claims damping "reduces but does not remove" a bias).
- Docs: `docs/estimation/saem.qmd` (drift + damping sections rewritten with the measurements, new "compare on the IS objective" section), `fit-options.qmd` row, `types.rs` option doc, CHANGELOG.

## Alternatives considered

Measured, all with the repaired solver (importance-sampled −2 log L at the final estimates; lower is better):

| `mstep_damping` | melphalan | clofarabine | thiotepa | `covmuref_power` TH_WT | FREM TVFRD1 |
|---|---:|---:|---:|---:|---:|
| 0.03 (old default) | 2695.6 | 17770.3 | 5928.0 | 0.323 | 0.361 |
| 0.1 | 2685.8 | 17768.7 | 5897.1 | 0.340 | 0.345 |
| 0.3 | 2683.2 | 17767.3 | 5860.2 | 0.403 | 0.282 |
| 0.999 (exploration off, convergence 1/(k−k1) average kept) | 2682.9 | 17764.1 | 5831.5 | 0.637 | — |
| **1.0 (off, new default)** | 2684.0 | 17764.3 | **5827.3** | **0.961** | 0.181 |
| reference | NONMEM SAEM 2682.4 | NONMEM SAEM 17734.6 | IS at FOCEI optimum 5823.6 | NONMEM SAEM 0.921 | IMP 0.31 / NONMEM IMP 0.39 |

- **Any exploration cap** freezes the covariate-slope shape (`TH_WT`) and starves thiotepa; there is no value that works for both those and FREM.
- **Keeping the convergence-phase Robbins–Monro average** of the maximisers (0.999 row) loses on the key shape: the sequence of η-frozen maximisers is an EM trajectory, not noise around a fixed point, so the 1/k average weights the early, still-crawling iterates most.
- **A larger M-step budget** (3×): no gain, 2.3× wall.
- **The `iiv_on_ruv` drift** is real and not fixed here: undamped, the reprex settles near `TVFRD1` 0.18 from either start, keeps drifting on a 3× longer schedule (0.032), 200 MH steps per iteration do not change it, and IS cannot rank the two points on that 12-eta model (ESS/K = 0.001) — but the FOCEI objective can: the drifted point is **47 units worse** (7492.7 vs 7445.3 at the held start), so on that model the hold is genuinely the better answer. That is why the `iiv_on_ruv` default keeps the cap (see *What changed*); the cap only ever holds the theta at its start, and the drift itself needs its own fix (the `η_RUV` re-centring into σ (#904) modifies the E-step, and σ shares the numerical M-step with the no-ETA thetas — the likely coupling). Left for its own issue; `mstep_damping = 0.03` (or `FIX`) is the documented hold.
- SA-averaged gradient/Newton M-step (the textbook non-exponential-family SAEM update, what NONMEM does for non-MU thetas): the principled long-term answer, but a new estimator with a wide blast radius; not attempted here.

## Cross-repo dependency

## Breaking changes
- [x] None (a fit-option default changed; the option and its range are unchanged)

## Numerical validation
- [x] Compared against: NONMEM SAEM (melphalan, clofarabine, `covmuref_power`), IS at the FOCEI optimum (thiotepa)
- [x] Reference values:

```
# before (main) → after, importance-sampled −2 log L at the SAEM estimates
thiotepa     6047.8 → 5827.3   (IS at the FOCEI optimum 5823.6; NONMEM SAEM aborted on this model)
melphalan    2705.8 → 2684.0   (NONMEM SAEM+IMP 2682.4)
clofarabine 17778.5 → 17764.3  (NONMEM SAEM+IMP 17734.6; NONMEM FOCEI 17745.1, ferx FOCEI 17777.3)
covmuref_power, TH_WT on the numerical channel: 0.332 → 0.961 (NONMEM SAEM 0.921, start 0.3)

# thiotepa, per theta: init / FOCEI / SAEM before / SAEM after
TVQ    2    1.390  1.981  1.848
TVVP  10    5.347  11.03  6.005
TVCLM  0.5  1.542  0.609  1.473
TVKM   1    0.336  1.011  0.394
HILL   1    1.762  1.006  9.878   (HILL/TM50 maturation ridge; FOCEI from this point returns to 5832 / 1.76)
TM50   0.5  0.192  0.460  0.221
HILL2  2    2.628  1.966  2.886
TM502  0.5  0.319  0.497  0.292
```

### Regression check against #1014 (the #1011 damping PR)

Every test #1014 added passes on this branch (`damp_mstep_is_a_robbins_monro_blend`, `mstep_sa_step_caps_exploration_and_frees_convergence` — rewritten to pin the new default —, `test_mstep_damping_round_trips_and_validates`, `test_mstep_damping_under_focei_warns`, `damps_numerical_mstep_only_for_a_free_unanchored_theta`, the inert / clamped warnings), as do the slow anchors it ran (`frem_warfarin` incl. `frem_iiv_on_ruv_saem_stays_bounded`, `mixture_nonmem`). Its own reprex keeps its result by default because the `iiv_on_ruv` default keeps the cap: `TVFRD1` 0.361 (this branch, defaults) vs 0.290 (#1014's build) vs IMP 0.31 / NONMEM IMP 0.39 — both on the marginal optimum #1014 profiled (≈0.29–0.32). #1014's healthy-model check: `examples/warfarin_saem.ferx` is byte-identical (all thetas mu-referenced, σ on the scalar closed form, so the numerical solve never runs); `examples/warfarin_iov_saem.ferx` moves OFV 310.69 → 310.01 because an IOV fit routes σ through the numerical solve (no scalar closed form with κ), which is now the repaired one — an improvement, and the only "byte-identical" promise #1014 made (all-mu-referenced / all-FIX fits) was about the damping gate, which is unchanged.

## Performance
- [x] No significant impact expected — thiotepa SAEM wall 80 s → 112 s (the M-step now uses its budget instead of stopping on `ftol` after a handful of evaluations); melphalan 10 s → 16 s; the M-step budget itself is unchanged.

## Tests
- [x] Unit test added: `budgeted_mstep_returns_the_converged_conditional_maximiser` (3-subject fixture; mutation-checked — dropping `set_initial_step` fails it at both budgets: realised 0.060 / 0.024 vs bounds 0.01 / 2e-3), `test_parse_method_chain_final_non_foce_keeps_interaction_default`, and `mstep_sa_step_caps_exploration_and_frees_convergence` now pins the new default
- [x] Review fix (P1, narrow bounds): the initial step is now bounded per coordinate by a quarter of its interval (`mstep_initial_step`; BOBYQA requires `upper − lower ≥ 2·step` and otherwise rejects the whole problem before the first evaluation, which the discarded outcome had turned into a silent no-op for θ *and* σ). `mstep_initial_step_is_bounded_by_the_interval` pins the rule (wide, narrow, threshold, pinned, unbounded), `mstep_moves_a_theta_declared_on_a_narrow_interval` is the reviewer's reproduction (CL on (1.9, 2.1), 0.10008 log units) and fails under the unbounded step; a `debug_assert!` now fires on `InvalidArgs` so a future misconfiguration cannot be silent in any test build.
- [x] Regression test added that fails without this fix: Tier-3 `saem_recovers_the_allometric_exponent_on_the_numerical_mstep` (`nonmem_anchor/covmuref_power_numeric_saem_fit.ferx`: the #619 power model with the covariate routed through an `if` so the group detection does not apply; NONMEM SAEM anchor 0.921, before 0.332, after 0.961; fails under either the old first design or the old 0.03 cap)

- [x] Not applicable (fix; the option keeps its syntax)

## Docs

## Checklist
- [x] No version bump

## Docs & examples
- [x] No example execution step needed (fix; `[fit_options]` key unchanged — its fit-options row is updated here)

## Reviewer hints

- `src/estimation/saem.rs`: the code change is four lines (`set_initial_step`, `ftol_rel`, the default constant); the rest is doc and tests. Read `MSTEP_SA_MAX_STEP`'s doc for the measurements that justify turning the cap off.
- `src/parser/model_parser.rs`: the chain `interaction` rule. `test_parse_method_chain_final_foce` still pins `[.., foce]` → off.
- The SAEM fit-yaml `ofv` is still a FOCE approximation (now with interaction for chains); comparisons across estimators should use the `IS done. −2 log L` line / `FitResult::importance_sampling`, which the new docs section says.

## Open questions

- Follow-ups filed: #1421 (the `iiv_on_ruv` numerical-channel drift, see Alternatives) and #1422 (melphalan's `CL_CRCL` misses the #619 covariate group because `CRCLM` is assigned through `if`/`else` + `min()`, which the pre-classification inliner does not fold — NONMEM's control stream puts the same term inside `MU_1`). Note for #1422: once `if`-assigned locals are inlined, the Tier-3 fixture added here (`covmuref_power_numeric_saem_fit.ferx`) becomes a group model and will need a shape that still defeats detection.


src/estimation/saem.rs
```diff
@@ -1021,7 +1118,28 @@ fn theta_sigma_mstep_light(
...
+        // draw is not the same estimator, and needs its own schedule and its
+        // own anchor before it changes. See `MSTEP_MIXTURE_FTOL_REL`.
+        opt.set_ftol_rel(MSTEP_MIXTURE_FTOL_REL).unwrap();
+    } else {
+        opt.set_ftol_rel(MSTEP_FTOL_REL).unwrap();
+        // A warm-started local solve needs a *local* first design (#1415).
+        // Every coordinate gets the same trust radius in packed units, undone
+        // through the optional magnitude scaling so the radius is the same
+        // fraction of the parameter either way.
+        let initial_step: Vec<f64> = (0..n).map(|i| MSTEP_INITIAL_STEP / scale[i]).collect();
```
roninsightrx: [P1] Bound the initial step by each free coordinate's interval. BOBYQA requires `upper - lower >= 2 * initial_step` after rescaling. The unconditional 0.1 packed-unit step violates that for any free log-packed theta whose upper/lower ratio is below exp(0.2), or identity-packed theta with an interval narrower than 0.2. For example, a valid CL bound of 1.9–2.1 has packed width 0.10008, so NLopt returns `InvalidArgs` before evaluating the objective. Since `opt.optimize` errors are discarded below, the entire joint theta/sigma M-step silently returns unchanged on every iteration, freezing the other numerical parameters and sigma too. I reproduced this using the new conditional-maximizer fixture with only CL's bounds narrowed; even the 900-iteration reference budget makes no progress. Cap the step using the positive bound width before applying `scale[i]`, preserve handling of pinned dimensions, and add a narrow-bound regression test.

roninsightrx (commented):
Reviewed the numerical M-step, model-specific damping default, method-chain interaction handling, regression fixtures, and docs. Found one correctness issue that should be fixed before merge; see the inline P1 comment.

Validation on the unmodified PR head: all 97 SAEM unit tests and all four method-chain parser tests passed. A temporary narrow-bound variant of `budgeted_mstep_returns_the_converged_conditional_maximiser` fails: with CL bounded to 1.9–2.1, the conditional objective stays exactly -9.843672950395469 even with `maxiter = 900`, and both theta and sigma remain at their starting values. An independent reproduction against the repository's NLopt 2.9.1 library returns `InvalidArgs` with zero objective evaluations for the new initial step; the old default step optimizes successfully. I did not rerun the slow population-fit/NONMEM comparisons.

src/estimation/saem.rs
roninsightrx: Confirmed and fixed in 67c7c071 — thanks, this was a real hole. BOBYQA's `xu − xl < 2·rhobeg` check (`bobyqa.c`, "insufficient space between the bounds") rejects the whole problem before the first evaluation, and because `theta_sigma_mstep_light` discards the outcome, a single narrow-interval theta silently turned every M-step into a no-op for θ *and* σ.

What changed:

- `mstep_initial_step(lower, upper)` bounds the per-coordinate step at a quarter of the interval width (NLopt's own default, leaving the factor-of-two margin), applied before the `scale[i]` division; a pinned coordinate (`upper == lower`, eliminated by NLopt before BOBYQA) and an unbounded one keep the positive nominal step.
- A `debug_assert!` on `Err((InvalidArgs, _))` after `optimize`, so this configuration class can never be silent again in a test build; release keeps swallowing NLopt outcomes as before.
- Tests: `mstep_initial_step_is_bounded_by_the_interval` (wide / narrow / threshold / pinned / unbounded) and `mstep_moves_a_theta_declared_on_a_narrow_interval` — your reproduction (CL on (1.9, 2.1), 0.10008 log units, started at 1.95): the objective must fall and all three free coordinates must move. Mutation-checked: with the unbounded `MSTEP_INITIAL_STEP` the `debug_assert!` fires and, with it removed, the start comes back unchanged.

Re-verified after the fix: full lib suite, `tools/preflight.sh`, and the slow SAEM anchors (`frem_warfarin`, `saem_covariate_mu_ref`, `mixture_nonmem`, `iiv_on_ruv`, `saem_logit_mu_ref`) — all green; none of those fixtures has an interval narrower than 0.2 packed units, so their numbers are unchanged.

Issue #1409, opened 2026-09-17
W_CMT_DEFAULTED's scope must be derived from CMT's consumers, not listed — [data_selection] is a live hole


TeunP:
## What

`W_CMT_DEFAULTED` (#1009 / #1404) is decided by `api::validation::cmt_defaulting_is_ambiguous`: the reader reports every row whose compartment it had to invent, and that predicate decides which models are shown the finding. The predicate has been **wrong in three consecutive review rounds of #1404**, each time in the same way — it was written as a *list* of the model classes someone had thought of, and each round a reviewer found the class that was missing.

| round | what the list missed |
|---|---|
| 1 | observation-side dispatch entirely (`[scaling] obs_scale[CMT=N]` on an analytical model) |
| 2 | `OdeSpec::readout` (it inspected only `AnalyticReadout::readout` — two structs, same `OdeReadout`), and the analytical **dose** channel (an oral model's `CMT=2` is a real central bolus) |
| 3 | `[data_selection]`, below — plus a narrowing that had to be reverted |

Three rounds is the repo's own stop condition (`/ferx-resolve` §7, `feedback_stop_iterating_split_instead`). #1404 ships the converged half — the float-`CMT` parse fix, the reader's counting/message/escaping work, the NONMEM anchor, and a predicate that is never *narrower* than before. This issue is the cut half: **derive the predicate from an enumeration of CMT's consumers, once, instead of widening it again.**

## The live defect: `[data_selection]` reads the defaulted CMT and nothing says so

`resolve_row_cmt`'s own doc (`src/io/datareader.rs`) says it is shared by "the dose row, the observation row **and the `[data]` selection filter's `RowContext`**", and the filter is fed the *defaulted* value. The predicate never asks whether a `[data_selection]` clause mentions `CMT`.

Measured — `pk one_cpt_iv`, `[data_selection] ignore = CMT == 2`, one observation cell spelled `2` vs `x`, nothing else changed:

| dataset | observations kept | OFV | `ferx check` |
|---|---|---|---|
| cell spelled `2` | 3 | −4.6162 | `ok — 0 warning(s)` |
| cell spelled `x` | **4** | **−6.9517** | `ok — 0 warning(s)` |

The row the filter was told to drop is silently kept and scored. This is the #1009 defect on a third channel.

[...]

PR #1423, opened 2026-09-18
fix(validation): derive W_CMT_DEFAULTED's scope from CMT's consumers, closing the [data_selection] hole [closes #1409]

roninsightrx:
> **Stacked on #1404** (`worktree-issue-1009-cmt-defaulted`), which is where
> `W_CMT_DEFAULTED`, `cmt_defaulting_is_ambiguous` and `tests/cmt_defaulting_scope.rs`
> live. The base branch is set accordingly, so the diff shown is this change alone;
> #1404 must merge first.

## Why

`W_CMT_DEFAULTED`'s scope is decided by `api::validation::cmt_defaulting_is_ambiguous`.
It was written as a **list** of the model classes someone had thought of, and it was
wrong in three consecutive review rounds of #1404 — each round a reviewer found the
class that was missing (round 1: observation-side dispatch entirely; round 2:
`OdeSpec::readout` and the analytical dose channel; round 3: `[data_selection]`).
Three rounds is the repo's own stop condition, so #1404 shipped the converged half
and #1409 is this: **derive the predicate from an enumeration of `CMT`'s consumers,
once, instead of widening it a fourth time.**

The live defect is the `[data_selection]` hole. `resolve_row_cmt`'s own doc says it is
shared by "the dose row, the observation row **and the `[data]` selection filter's
`RowContext`**" — and the filter is fed the *defaulted* value, so a clause naming `CMT`
selects records on a compartment nobody wrote. Measured on `pk one_cpt_iv`,
`[data_selection] ignore = CMT == 2`, one observation cell spelled `2` against `x`,
nothing else changed:

| dataset | records scored | −log L at the initial estimates | `ferx check` |
|---|---|---|---|
| cell spelled `2` | 3 | −5.7650 | `ok — 0 warning(s)` |
| cell spelled `x` | **4** | **−5.2516** | `ok — 0 warning(s)` |

The row the filter was told to drop is silently kept and scored. (The issue's own
reprex reports −4.6162 / −6.9517 — the same defect on its own fixture; the numbers
above are this PR's test fixture, printed by it on every run.)

## What changed

**The predicate is now an enumeration.** `CmtConsumer` (in `api/validation.rs`) names
every channel that reads a row's `CMT`, and `cmt_defaulting_is_ambiguous` is
`CmtConsumer::ALL.iter().any(|c| c.is_live(model, options))`. Each arm asks the
*consumer's own* routing table rather than restating it:

| variant | asks |
|---|---|
| `DoseCompartment` | `OdeSpec::n_states` / `PkTopology::channels` |
| `PerCmtScaling` | `ScalingSpec::PerCmt` |
| `PerCmtErrorModel` | `ErrorSpec::PerCmt`, **non-empty** |
| `PerCmtReadout` | `OdeSpec::readout` **and** `AnalyticReadout::readout` |
| `EndpointRouting` | `api::run::obs_routing_for` — the one place a routing set is derived from a model |
| `DataSelectionFilter` | `SelectionFilter::references_column("cmt")` |

A new consumer is a compile error in `CmtConsumer::is_live` and `::index`, and a red
test until it carries a fixture on which it alone is live.

**The `[data_selection]` hole is closed.** `reader_warning_suppressed` takes a
`&FitOptions` (the signature widening the issue anticipated); `fit()` passes its merged
options, `ferx check` passes `parsed.fit_options`. The clause is asked of the *parsed*
`FilterClause`s, not of the raw strings, so case-folding, the `&&` split and the
bare-identifier `IGNORE=C` shorthand are answered by the filter's own parser. A clause
that fails to parse answers `true` (the read is about to fail on it anyway; `false`
would narrow the predicate on an expression nobody has inspected) — pinned by a test so
it is a decision rather than an `unwrap_or_default()`.

**Endpoint routing is explicit** (structural ask 1). Before this, an endpoint model was
reported only because the parser hands it `ErrorSpec::PerCmt({})` — an empty map that
dispatches no error model. That is an accident, and the `!m.is_empty()` gate tried in
#1404 round 3 removed it, silencing a measured endpoint mis-routing. The error-model arm
now requires a non-empty map and `EndpointRouting` covers the empty one by asking the
routing sets. The two arms therefore reject **disjoint** inputs — no belt-and-braces, so
neither can cover for the other in a mutation (CLAUDE.md, "two redundant gates cover for
each other"). `the_empty_error_map_and_the_routing_set_are_the_same_models` is the guard
that the narrowing loses no scope.

**The dose arm gates on "a closed form serves this model"** rather than on
`!is_algebraic()` (fold-in 2). `is_algebraic()` is the compartment-free marker (#811);
an endpoint-only model fails it while still never reaching a closed form, and its
`pk_model` is the placeholder `types.rs` warns must never be dispatched on (#1356).
Benign today only because that parser arm picks a one-channel model while the `[odes]`
arm picks a two-channel one.

**`PkTopology::channels` now pins the no-interior-`None` invariant** (fold-in 1), so
counting live channels and reading the highest addressable index cannot come apart —
the cheaper spelling stays the right one instead of being right by coincidence.

## Alternatives considered

- **Keep `ErrorSpec::PerCmt(_)` matching any map and *add* the routing arm.** Never
  narrower, and tempting. Rejected: the two arms would then accept the same inputs on
  the empty map, so deleting either leaves the suite green — exactly the test hole
  CLAUDE.md names. Collapsing to one gate per input is what makes the mutation below
  meaningful.
- **`Err(_) => false` on an unparseable `[data_selection]` clause.** Rejected as a
  silent narrowing; see above.
- **Thread `&FitOptions` through `predict()` / `simulate()`** so the filter channel is
  answered there too. Rejected: it widens four public entry points to report a filter
  those entry points never applied — they are handed a `Population` the caller already
  read. `non_fit_diagnostics` passes `FitOptions::default()`, with the asymmetry recorded
  at the call site and in its doc (the same structural reason `check_model_options` is
  excluded there wholesale).

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` is green; the new methods are
`pub(crate)`).

## Breaking changes
- [x] None

A model that previously ran silently with a `[data_selection]` clause on `CMT` and a
defaulted compartment now gets a warning. No numbers change.

## Numerical validation
- [x] Not applicable — **this change moves no number.** It decides which *diagnostic*
      is shown; every prediction, objective and estimate is bit-identical. NONMEM has no
      equivalent finding at all (`$MODEL` lets it declare `DEFDOSE`; ferx has no such
      declaration), which is why #1009 exists — so there is no NONMEM run to anchor the
      warning's scope against. What the scope *is* anchored on is a measurement, per
      case: the tests read each dataset both ways and assert the warning agrees with
      whether the compartment actually changed what is scored.

Realised table, printed by `tests/cmt_data_selection_scope.rs` on every run:

```
cmt data-selection scope, realised:
  no [data_selection]: kept 4 -> 4, −logL -5.2516 -> -5.2516, observable false, warned false
  ignore = CMT == 2:   kept 3 -> 4, −logL -5.7650 -> -5.2516, observable true,  warned true
  accept = CMT == 1:   kept 3 -> 4, −logL -5.7650 -> -5.2516, observable true,  warned true
  ignore = DV < 0.001: kept 4 -> 4, −logL -5.2516 -> -5.2516, observable false, warned false
```

## Performance
- [x] No significant impact expected — the predicate runs once per reader warning, and
      only when one was emitted. It now compiles the `[data_selection]` clauses a second
      time on that path; there are at most a handful, and the alternative is a second
      spelling of the filter's parser.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 4982 passed)
- [ ] No new tests needed

**Tier 1** — `src/api/tests/reader_warning_suppression_tests.rs` (+11 tests):

- `cmt_consumer_all_lists_every_variant` — `ALL` pinned dense and complete against the
  exhaustive `match` in `CmtConsumer::index`.
- `every_cmt_consumer_has_a_fixture_where_it_is_the_only_live_one` — **the test the
  issue asked for.** `only_live` is an exhaustive `match` over `CmtConsumer`, so a new
  variant is a compile error until someone builds the shape that makes it live; the test
  asserts the fixture makes *exactly* that channel live (`live == vec![c]`), so no arm is
  pinned by a neighbour that happens to fire on the same model.
- `a_model_that_reads_cmt_for_nothing_is_suppressed_on_every_channel` — the control,
  without which the loop above passes for a predicate hard-wired to `true`.
- four `[data_selection]` cases: the live defect; a clause on another column (so an arm
  written `!ignore_exprs.is_empty()` fails); the spellings the filter accepts
  (`cmt`, `TIME > 0 && CMT != 1`, `Cmt >= 2`, an `accept` clause, `CMTX` as a different
  column, `ignore_subjects` as a non-consumer); the unparseable clause.
- `an_endpoint_model_reports_because_its_rows_route_by_cmt` and
  `the_empty_error_map_and_the_routing_set_are_the_same_models` — the endpoint channel on
  the routing sets, with a non-empty `ErrorSpec` so the error-model arm is inert.
- `an_endpoint_only_models_placeholder_pk_model_is_not_read_as_a_dose_channel` —
  asserted on the *channel*, not on the warning, because such a model reports anyway via
  `EndpointRouting` and a warning-level assertion would pass with the gate deleted.

**Tier 2** — `tests/cmt_data_selection_scope.rs` (new, 2 tests). Measures rather than
declares, in the shape `cmt_defaulting_scope.rs` uses for the dose channel: reads each
dataset through the same filtered path `fit()` uses, counts what survived, evaluates
`individual_nll` on it, and asserts the warning fires **iff** the defaulted compartment
changed what is scored. The straddle is asserted so the table cannot quietly become
one-sided; the `nll` fold carries an `is_finite()` guard so a `NaN` cannot be absorbed.
No `fit()`.

**Mutation-checked**, one at a time, each killed by a named test and nothing else:

| mutation | reddens |
|---|---|
| `DataSelectionFilter => false` | 4 Tier-1 + both Tier-2 tests |
| `EndpointRouting => false` | 4 Tier-1 tests |
| `PerCmtErrorModel` back to any map | `the_empty_error_map_and_the_routing_set_are_the_same_models` |
| closed-form gate → `is_algebraic()` | `an_endpoint_only_models_placeholder_pk_model_is_not_read_as_a_dose_channel` |
| `references_column` ignores the name | `a_data_selection_clause_on_another_column_is_not_a_cmt_consumer`, `the_data_selection_arm_reads_the_clause_the_way_the_filter_does` |

- [x] Not applicable — no new API surface; a warning's scope widened.

## Docs
- [x] `docs/` updated for user-visible changes — `docs/model-file/data-selection.qmd`
      (a new note that `CMT` is the one filterable column with a default, with the
      measurement, and the link to the warning) and
      `docs/file-formats/check-report.qmd` (the `W_CMT_DEFAULTED` row's scope).
      `src/diagnostics.rs`'s registry row too.
- [x] `CHANGELOG.md` `[Unreleased]` entry added, under `Fixed`.
- [ ] No user-visible change

`cargo run -p docs-lint -- --check` is green.

## Checklist
- [x] Functions on the `Dual2` sensitivity path stay differentiable — none touched; this
      is diagnostic routing with no numerical surface.

### ferx-site
- [x] No new DSL syntax, `[fit_options]` key, example or data file.

### ferx-book
- [x] No estimator / DSL / fit-option change.

### Example execution
- [x] No example execution step needed — no user-visible numerical change, no example
      touched.

## Reviewer hints

Focus on **`CmtConsumer::is_live`** — the arms must reject disjoint inputs, and the
claim that they do is what `every_cmt_consumer_has_a_fixture_where_it_is_the_only_live_one`
pins. The load-bearing narrowing is `PerCmtErrorModel` requiring a non-empty map: its
safety condition is that *every* model with an empty `ErrorSpec::PerCmt` has a CMT in a
routing set. The parser only produces that empty map for a model with no `[error_model]`
block, which is an endpoint-only model — but that is an argument about the parser, and
the thing pinning it is a test on fixtures. If you can construct an empty-map model that
routes nothing, this narrows the warning and the arm should go back to `PerCmt(_)`.

Skimmable: the doc comments carrying the #1404 measurements forward, and the topology
invariant assert.

## Open questions

1. `ferx check` reads the population with `filter: None` (it always has), so the record
   counts it reports are unfiltered while the warning it now raises is about the filter.
   Not wrong, but the two are answering slightly different questions. Worth a separate
   issue?
2. `predict()` / `simulate()` are dark on the filter channel (`FitOptions::default()` at
   the `non_fit_diagnostics` call site). The honest fix is for those entry points to know
   what filtered their population, which is a bigger change than this one. Say if you'd
   rather it were filed.


roninsightrx:

The PR’s parsing and topology changes look carefully tested, but there is one blocking hole in the central `W_CMT_DEFAULTED` guarantee: the warning can disappear precisely when a defaulted `CMT` causes the filter to remove the row. This was a static review only, as requested; no code was compiled or executed.

## Blocking findings

- `src/io/datareader.rs:2052 — high — The resolved CMT’s default cause is discarded before filtering, and an excluded row reaches the continue at line 2120 before record_dose/record_obs can count it. For example, with ignore = CMT == 1, an unreadable CMT defaults to 1 and removes the row, but total_cmt_defaults remains empty, so no W_CMT_DEFAULTED exists for CmtConsumer::DataSelectionFilter to surface. Defaulted EVID=2/3 rows filtered through CMT are likewise never counted. — Preserve and record the default cause when a CMT-reading filter consumes it, before any exclusion continue, using a separate filter-consumption count so unrelated exclusions remain silent. Add a regression where the only defaulted row is excluded by ignore = CMT == 1 or accept = CMT == 2 and assert that fit/check still report W_CMT_DEFAULTED.`

## Nits

- `tests/defdose_no_cmt_nonmem_anchor.rs:43 — nit — The comment says this integration test “runs on every PR,” repeated in nonmem_anchor/README.md:700, but the repository guidance says tests/*.rs are only compile-checked on PRs and executed nightly. — Describe it as an ungated nightly Tier-2 test, or add an explicit PR CI invocation if per-PR execution is intended.`

TeunP:
Two things from #1404, now that its round-4 independent review is in.

**Base has moved.** #1423 branches at `cc7c0142`; #1404 is now at `fc088209`. Six commits behind, four of them main (`024a92f3`, `c3e8ba41`, `48f274a6`, `e5bd9067` — I merged main twice mid-landing), plus one merge and one docs commit of mine. Worth merging up before the next run, since this branch gets no CI while it is based on mine.

**Two review findings land squarely in your `EndpointRouting` channel, with fixtures attached.**

The review confirmed what your doc comment already says about my approach — `ErrorSpec::PerCmt(_)` is a proxy, and it only ever matches endpoint-**only** models, because the parser produces the empty map solely from `is_endpoint_only`. So:

1. **An endpoint model that also declares `[error_model]` gets `ErrorSpec::Single` and stays silent.** Measured on the shipped `examples/tte_exponential.ferx` (`[event_model] cmt = 2` + `pk one_cpt_iv` + `[error_model]`), one event row's `CMT` cell spelled `x`: no `W_CMT_DEFAULTED` either way, while the row moves out of the TTE endpoint into the Gaussian grid — **OFV 166.4395 → 254.5871**. That example is probably a better fixture than a hand-built one, since it is in-tree and the gap is invisible without it.

2. **The opposite error, same cause:** a model whose only endpoint is at `cmt = 1` is reported although the default is provably the one right answer — measured identical OFV (125.1207) with and without the `CMT` column. `model_routes_rows_by_cmt` can see that; the error-map proxy cannot.

Both fall out of keying on the declared endpoints, which is what you have done.

**One that is not yours unless you want it.** The emitted message in `datareader.rs` never mentions endpoint routing at all, and on an endpoint-only model every clause is false — it offers `[odes]` states and `pk` compartments to a model with neither, and says "changes the numbers with no error" directly above `E_ENDPOINT_NO_RECORDS`. I left it alone; once your channel lands, the remedy sentence probably wants an `[event_model] cmt = N` arm.

**What I did and did not touch**, so we do not collide: I changed **no predicate logic**. #1404 only corrects the claims — `CHANGELOG.md` and `docs/data-format.qmd` / `docs/file-formats/check-report.qmd` now say endpoint-*only*, name both limits above, and point at #1409. When #1423 lands, that prose becomes wrong in the good direction and wants a pass.

roninsightrx:

Pushed as `4fb72ce6`. One finding fixed — it was real, and it was a genuine hole in the guarantee this PR claims to establish. One nit disputed, with evidence.

---

### 1. `src/io/datareader.rs:2052` — the resolved CMT's cause is discarded before filtering — **FIXED, and the finding was correct**

Confirmed by measurement before touching anything. `pk one_cpt_iv`, `[data_selection] ignore = CMT == 1 && TIME == 4`, one observation cell spelled `2` against `x`, nothing else changed:

| dataset | records scored | −logL | `fit()` reports |
|---|---|---|---|
| cell spelled `2` | 4 | −5.2516 | — |
| cell spelled `x` | **3** | **−5.7650** | **nothing** |

The row is deleted from the fit *because* the reader guessed its compartment, and `W_CMT_DEFAULTED` was absent. Exactly as described: `record_dose`/`record_obs` are past the filter's `continue`, so `CmtDefaults` stayed empty and there was no warning for `CmtConsumer::DataSelectionFilter` to survive the suppression step.

This is the **mirror** of the case the PR was built around, and I had covered only one side. A guessed compartment can make a clause *stop* matching (row wrongly kept — and a kept row is counted normally, which is why every test passed) or *start* matching (row wrongly deleted — invisible to every counter). The PR's own property table was one-sided in a way I did not see.

**Fix**, following the suggested shape:

- The cause is now kept across the filter block (`let (cmt_for_ctx, cmt_cause_for_filter) = resolve_row_cmt(...)`) and recorded **before** the `continue`.
- A separate counter, `CmtDefaults::n_filtered`, so unrelated exclusions stay silent and the numbers do not mix. It is named separately in the message too — a deleted row is not "assigned compartment 1" in any useful sense, it is *missing*, and the remedy reads differently ("had the compartment been written out, those rows would have been kept"). The existing exhaustive destructure in the message builder is what forced this to be answered for.
- Scoped to **the rule that actually fired**, not to "this block mentions CMT somewhere". `should_exclude` is now a thin wrapper over a new `first_excluding_rule -> Option<Excluder<'_>>`; one implementation of the documented exclusion ordering answering two questions, so the two cannot drift about which rule fired. `Excluder::reads_column("cmt")` is the gate.

**On the EVID=2/3 half of the finding** — you were right that those rows are never counted, but I want to be precise about why, because it is not the same defect. `n_dose`/`n_obs` deliberately count only rows the fit uses, and that predates this PR: a *kept* `EVID=2` row with a defaulted CMT is not reported today either, and correctly so — its compartment addresses nothing. What is new is the deletion case: a deleted `EVID=2` row removes a covariate-change marker from the fit, and a guessed compartment is why. The new counter sits at the filter, so it catches every record type including `EVID` 2/3. Both directions are pinned on the same row in `a_deleted_evid_2_row_is_reported_even_though_a_kept_one_is_not`, so the asymmetry is a decision rather than an accident of where the counter sits.

**Regressions added** (`tests/cmt_data_selection_scope.rs`, now 6 tests):

- `ignore = CMT == 1 && TIME == 4` added to the property table — the mirror direction, measured (`kept 4 -> 3`) alongside the existing `kept 3 -> 4` rows.
- `a_defaulted_row_the_filter_deletes_is_still_reported` — checks the whole summary, not just its presence: the deletion is counted, is **not** double-counted as a scored row (`0 dose row(s) and 0 observation row(s)`), and still names the offending spelling.
- `an_exclusion_decided_by_another_column_stays_silent` — a `TIME` rule deletes the defaulted row while a `CMT == 7` clause that matches neither spelling sits beside it. This is the test that fails for the cheap whole-filter `references_column("cmt")` implementation, which passes everything else in the file.
- `a_deleted_evid_2_row_is_reported_even_though_a_kept_one_is_not`.
- `ferx_check_reads_unfiltered_so_it_answers_a_different_question` — see below.

**Mutations**, each killed by a named test: dropping the `reads_column("cmt")` gate reddens `an_exclusion_decided_by_another_column_stays_silent` and nothing else; dropping the `record_filtered` call reddens four tests including the property table.

#### A second defect your finding exposed: the test oracle could not see it

Worth recording, because it is the more embarrassing half. My `warns()` helper went through `validate_model_file`, and `ferx check` reads with `filter: None` (`api/validation.rs`). So it counted the row the fit never sees and reported `W_CMT_DEFAULTED` either way. When I first added the mirror case to the property table it **passed green** against a reader that counted nothing — the fixture was right and the oracle was blind, which is the CLAUDE.md "a green test is not evidence that it can fail" failure mode one level up from the code.

The oracle is now `fit()` itself at `outer_maxiter = 0` (one objective evaluation, no convergence loop, still Tier 2), rather than a re-spelling of `reader_warning_suppressed` in the test file — a second copy of the predicate there would agree with a wrong answer by construction. Only with that oracle does the mirror case fail before the fix, which is how I confirmed the finding rather than taking it on trust. `ferx_check_reads_unfiltered_so_it_answers_a_different_question` now pins the divergence instead of leaving it as prose: both entry points report the finding, but only `fit()`'s sentence is about a deletion.

This sharpens open question 1 in the PR description — `ferx check` not applying the filter is not merely cosmetic, it makes check and fit answer different questions about the same dataset. Still out of scope here; happy to file it.

---

### 2. `tests/defdose_no_cmt_nonmem_anchor.rs:43` — "runs on every PR" — **not a defect, the comment is correct**

CLAUDE.md does say Tier-2 tests are "compile-checked on every PR (`cargo check --tests`) and run nightly in `slow-tests.yml`", so the reasoning is fair — but that sentence is stale relative to the workflow. `.github/workflows/ci.yml`, the `Tests + coverage (core)` job:

```yaml
name: Tests + coverage (core)
if: github.event_name == 'pull_request' || github.event_name == 'push'
...
run: cargo llvm-cov --workspace --tests --no-fail-fast --no-default-features --features ferx-core/ci --profile ci-cov --lcov --output-path lcov.info
```

`--tests` builds and **runs** the integration targets, and the job's own comment says so: "`--tests` so the integration tests both run and count (same rationale as nightly)". `defdose_no_cmt_nonmem_anchor.rs` carries no `#[cfg(feature = ...)]` and no `#[cfg_attr(not(feature = "slow-tests"), ignore)]`, so it compiles and runs under the plain `ci` feature set on every PR. The `maxiter = 0` note is what earns it that — the comment is describing why it is cheap enough to be in the per-PR job, which it is.

The one real caveat is the feature set, not the trigger: that job runs `ferx-core/ci` only, so a `survival`-gated integration test does *not* run on PRs (it runs in the endpoints job, which is `schedule`/`dispatch`). Neither file is gated, so neither is affected.

I have not edited CLAUDE.md here — that is repo-wide guidance and a change to it belongs in its own PR rather than riding along in this one. Flagging it: the Tier-2 paragraph should say integration tests run per-PR under `ferx-core/ci` and nightly under the wider feature sets, since the current wording is what produced this nit and will produce it again.

---

`tools/preflight.sh` green (`fmt check clippy rustdoc docs public-api`), `cargo test --lib` 4982 passed, the affected integration tests green. Docs and CHANGELOG updated for the deletion half. No public API change — `first_excluding_rule`, `Excluder` and `record_filtered` are all `pub(crate)`, and `should_exclude`'s signature is unchanged.

One thing the rustdoc gate caught while fixing this, since it is the trap CLAUDE.md warns about: the first version of `should_exclude`'s doc linked `[`Self::first_excluding_rule`]`, a `pub(crate)` target from a `pub` method — `private_intra_doc_links`, red `Rustdoc` job. Fixed at the link with inline code, not with an `allow`.

TeunP:
**#1404 merged as `85413d90`.** Base retargeted to `main` for you; the old branch is still on origin so nothing is stranded, and I will delete it once you have merged up.

**Expect 10 conflicts, and the reason is not drift.** #1404 was **squash**-merged, so its individual commits are not ancestors of `main` — this branch and `main` now hold much of the same content via different histories. `git merge-tree origin/main <this branch>` says:

```
CHANGELOG.md                                       docs/data-format.qmd
docs/file-formats/check-report.qmd                 src/api/postfit.rs
src/api/tests/reader_warning_suppression_tests.rs  (add/add)
src/api/validation.rs                              src/diagnostics.rs
src/io/datareader.rs                               src/io/datareader_tests.rs
src/pk/topology.rs
```

Most resolve as "take `main`, keep the `CmtConsumer` work on top" — the content is already yours, it just arrives by the other route now.

**Four things changed under you since `cc7c0142`:**

1. **`reader_warning_suppressed` has a third call site.** #1280/#1304 added `predict_diag()` / `simulate_*_diag()`, which route through the same filter — `postfit.rs:2163`, alongside `fit.rs:1268` and `validation.rs:6505`. The `&FitOptions` widening needs all three, and `postfit.rs` is the awkward one: those entry points take no fit options, so there is a decision to make about what they pass for `ignore_exprs` / `accept_exprs`.

2. **`docs/data-format.qmd` gained a section heading.** The `W_CMT_DEFAULTED` prose now sits under its own `### When a defaulted CMT is reported (W_CMT_DEFAULTED)`, because adding to the old section pushed it to 5445 characters against docs-lint's 5000-character R1 limit. Keep the split when resolving, or R1 goes red.

3. **The docs now state the limits this PR removes**, so they become wrong in the good direction once it lands: the "Two limits of that arm" paragraph in `data-format.qmd`, the endpoint-only sentence in `check-report.qmd`, and the matching CHANGELOG clause. All three point at #1409 and want deleting.

4. **`W_CMT_DEFAULTED` was added to the reader-findings lists** in `postfit.rs:2113` and `docs/warnings.qmd` — it is the only code that filter actually withholds.

**Still unchanged and still yours:** the two fixtures from the comment above (`examples/tte_exponential.ferx`, OFV 166.4395 → 254.5871; the `cmt = 1` false positive at identical OFV 125.1207), and the message text in `datareader.rs`, which still never mentions endpoint routing.

Issue #1394, opened 2026-09-15
bug(check): E_OMEGA_INIT_AT_RAIL gives a diagonal omega the block_omega message whenever the model has any block — the #1229 FIX repair is lost


TeunP:
## Problem

`E_OMEGA_INIT_AT_RAIL` chooses its message and suggestion by whether **the whole Ω matrix** is a block, not whether **this eta** is in a block. So a diagonally declared omega gets the `block_omega` wording as soon as any `block_omega` exists anywhere in the model. That wording tells the user to "lower the covariances involving" an eta that has none. It also hides the #1229 repair (`write … ~ 0.0 FIX`), which is the whole purpose of that branch.

## Measured

`ferx check` on `examples/warfarin_block_omega.ferx` (debug build of PR #1388's head; this code is untouched by that PR). Both rows change only the `ETA_KA` line to `omega ETA_KA ~ 0.0`:

| model | message | suggestion |
|---|---|---|
| the other two etas written as diagonal `omega` lines | ``omega ETA_KA ~ 0.0` declares no variability but is not `FIX`-ed …`` | `write `omega ETA_KA ~ 0.0 FIX`, or start it at 0.09` ✅ |
| the example as shipped: `block_omega (ETA_CL, ETA_V)` plus diagonal `omega ETA_KA ~ 0.0` | `the `block_omega` Cholesky diagonal for ETA_KA starts at L = 1.000e-4 … In a block it is the **correlations** that do this …` | `the block is near-singular in ETA_KA: lower the covariances involving it, or `FIX` the block` ❌ |

`ETA_KA` is in no block and has no covariances. The second repair cannot be followed, and "FIX the block" would fix `ETA_CL`/`ETA_V` instead of the eta that is actually at the rail. `omega ETA_KA ~ 0.002 (sd)` (L² = 4e-6) takes the same wrong branch.

## Cause

`variance_decl_by_coordinate` in `src/api/validation.rs`:

```rust
if i < segs.sigma_start() {
    Some(VarianceDecl::Omega {
        block: !template.omega.diagonal,
    })
```

`omega.diagonal` is a property of the matrix. A model with one `block_omega` and any number of diagonal `omega` lines packs Ω as a full Cholesky, and the diagonal-declared etas' off-diagonals are structural zeros (#1018). The same shape applies to `Kappa { block: template.omega_iov … !m.diagonal }` for a `block_kappa` mixed with diagonal `kappa` lines.

## Fix shape

[...]

PR #1424, opened 2026-09-18
fix(check): decide E_OMEGA_INIT_AT_RAIL's message per eta, not per matrix [closes #1394]

roninsightrx:
## Why

`E_OMEGA_INIT_AT_RAIL` chose its message and suggestion from `template.omega.diagonal` — a property of the **whole Ω matrix** — not from whether *this eta* is in a block. A model with any `block_omega` packs Ω as a single Cholesky, so `!diagonal` is true for every coordinate, and a diagonally declared eta at the rail got the block wording.

Measured with `ferx check` on `examples/warfarin_block_omega.ferx`, changing only the `ETA_KA` line to `omega ETA_KA ~ 0.0` (#1394):

| model | message | suggestion |
|---|---|---|
| CL/V written as diagonal `omega` lines | `` `omega ETA_KA ~ 0.0` declares no variability but is not `FIX`-ed … `` | ``write `omega ETA_KA ~ 0.0 FIX`, or start it at 0.09`` ✅ |
| the example as shipped (`block_omega (ETA_CL, ETA_V)` + diagonal `omega ETA_KA`) | ``the `block_omega` Cholesky diagonal for ETA_KA starts at L = 1.000e-4 … In a block it is the **correlations** that do this …`` | ``the block is near-singular in ETA_KA: lower the covariances involving it, or `FIX` the block`` ❌ |

`ETA_KA` is in no block and has no covariances, so the second repair cannot be followed, and "`FIX` the block" would fix `ETA_CL`/`ETA_V` instead of the eta that is actually clamped. Worse, that arm *displaced* the `write ... FIX` repair — the #1229 repair this check exists to hand out. `omega ETA_KA ~ 0.002 (sd)` (L² = 4e-6) took the same wrong branch through the tiny-positive arm.

## What changed

Block membership is now a **per-coordinate** property, derived where the packed layout already lives.

- `estimation/parameterization.rs`: new `pub(crate) omega_block_member_mask(template) -> Vec<bool>` — packed-length, marking each Ω / Ω_IOV Cholesky **diagonal** whose eta has a structurally free off-diagonal, i.e. one declared inside a `block_omega` / `block_kappa`. It is `omega_structural_zero_mask`'s question asked per *eta* instead of per entry, and walks the same column-major lower triangle as `pack_params`, so the offsets are shared rather than re-derived. θ, Σ, off-diagonal coordinates, the mixture Ω overrides and the `block_sigma` ρ slots are all `false`.
- `api/validation.rs`: `variance_decl_by_coordinate` reads that mask. One packed-length lookup serves both the `Omega` and the `Kappa` arm, replacing the two separate matrix-flag reads.

`lower_tri_iter(n, om.diagonal)` is the **only** gate in the mask: it yields just `(i,i)` for a diagonal matrix, which is both the right offsets (a diagonal Ω packs `n` coordinates, not `n(n+1)/2`) and the right answer (no off-diagonal coordinate exists for any eta to share, and both `OmegaMatrix` constructors leave a diagonal matrix's `free_mask` off-diagonals false). There is deliberately no second `if om.diagonal` early return — per CLAUDE.md, two conditions rejecting the same inputs cover for each other and neither can then be shown to fail.

A one-eta `block_omega (ETA_CL)` has no off-diagonal and so reads as diagonal. That is the right answer rather than a gap: with no off-diagonals `L_ii²` *is* the declared variance, which is the property the diagonal arm rests on. Noted in the doc comment.

## Alternatives considered

Keying on the parser's `BlockOmegaSpec.names` instead of `free_mask`: `ModelParameters` — which is all `check_variance_init_rails` has — carries no block-spec list, and `free_mask` is exactly the structural record of who is in a block with whom (#1018). Reconstructing the specs to answer a question the mask already answers would be a second source to keep in step.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` green; the new function is `pub(crate)`).

## Breaking changes
- [x] None

## Numerical validation
- [x] Not applicable — diagnostic text only. No packed vector, bound, mask consumed by an optimizer, or objective is touched: `omega_block_member_mask` is a new function with one caller, in `api::validation`, and `packed_fixed_mask` / `omega_structural_zero_mask` are unchanged.

End-to-end check on the reported reprex, after the fix:

```
$ ferx check wbo_zero_ka.ferx        # examples/warfarin_block_omega.ferx with omega ETA_KA ~ 0.0
error[E_OMEGA_INIT_AT_RAIL] parameters:4: `omega ETA_KA ~ 0.0` declares no variability but is not
`FIX`-ed, so the optimizer is asked to estimate a variance from a start it cannot leave: the zero
is regularised to 1e-8, whose packed coordinate ln(L) = -9.21 lies below its own lower bound of
-6.0 … NONMEM rejects the same declaration outright: `INITIAL ESTIMATE OF VARIANCE CANNOT BE ZERO
UNLESS FIXED`.
    help: write `omega ETA_KA ~ 0.0 FIX`, or start it at 0.09

$ ferx check wbo_sd_ka.ferx          # same file with omega ETA_KA ~ 0.002 (sd)
error[E_OMEGA_INIT_AT_RAIL] parameters:4: ETA_KA starts at a variance of 4.000e-6 on the
optimizer's scale (ln(L) = -6.21), at or below its lower bound of -6.0 — every variance ≤ 6.14e-6
lands on that rail … Start ETA_KA at ≥ 1e-5, or `FIX` it if it should carry no variability.
    help: write `omega ETA_KA ~ 0.0 FIX`, or start it at ≥ 1e-5
```

Both match the issue's ✅ row. `near_singular_block_flags_the_factor_not_the_declared_variance` and `near_singular_block_kappa_is_reported_as_a_correlation_problem` stay green — the block wording is unchanged for a real block diagonal.

## Performance
- [x] No significant impact expected — one extra packed-length `Vec<bool>` per `check_variance_init_rails` call, on the same walk the check already makes.

## Tests

`cargo test --lib --features ci`: **4777 passed, 0 failed**.

Four Tier-1 `omega_block_member_mask` tests (`estimation/parameterization.rs`): pure diagonal all-false, full 2×2 block (diagonals true, the **off-diagonal between them** false), mixed block + standalone diagonal, and block Ω_IOV at its own offset. The existing block-Ω_IOV fixture was extracted to `block_plus_diag_iov_template()` and is now shared.

Three in `api/tests/variance_init_rail_tests.rs`:

- `a_diagonal_omega_is_unaffected_by_an_unrelated_block_in_the_model` — the differential pair the issue asks for. The two models differ by exactly one edit (how CL and V are declared); `omega ETA_KA ~ 0.0` is byte-identical across it, so message and suggestion must be equal. The **straddle is asserted** (`without.omega.diagonal && !with.omega.diagonal`) so the pair cannot silently become two runs of the same arm, and the diagonal wording is then pinned by content — an equality alone would also be satisfied by both halves taking the *block* arm.
- `a_mixed_model_reports_the_block_eta_and_the_diagonal_eta_differently` — a near-singular `block_omega` **and** a diagonal `~ 0.0` on the rail in one model, each reported in its own shape. No model-level answer can produce both, whichever way it decides.
- `a_diagonal_kappa_is_unaffected_by_an_unrelated_block_kappa` — the Ω_IOV half.

### Mutation testing

Each side was mutated separately, and each names its own:

| mutation | dies |
|---|---|
| `block = !template.omega.diagonal` (the bug, restored) | `a_diagonal_omega_is_unaffected_…`, `a_mixed_model_reports_…`, `near_singular_block_kappa_…` |
| the **Kappa arm alone** back to `omega_iov…!m.diagonal` | `a_diagonal_kappa_is_unaffected_…` **only** — the Ω tests stay green, so neither side is standing in for the other |
| drop `k != i` in the mask (every diagonal reads as a block member) | 3 mask tests + 9 rail tests |
| drop the `i == j` gate (off-diagonals get marked) | the 3 mask tests; no rail test, correctly — `check_variance_init_rails` filters on `PackedCoordKind::OmegaDiagonal` first, so the mask tests are the only ones that can see it |
| drop the Ω_IOV `mark(...)` call | `test_omega_block_member_mask_block_iov` + `near_singular_block_kappa_…` |

- [x] Not applicable — no new API surface; `examples/warfarin_block_omega.ferx` is unchanged (it declares `omega ETA_KA ~ 0.30`, which is interior and reports nothing).

## Docs
- [x] `docs/file-formats/check-report.qmd` — the `E_OMEGA_INIT_AT_RAIL` row now states that the message is chosen per eta, not per matrix, and what each arm says. `cargo run -p docs-lint -- --check` green.

## Checklist
- [x] No `Dual2` / `sens/` code touched

## Docs & examples

No new DSL syntax, `[fit_options]` key, example `.ferx`, or data file, so no ferx-site / ferx-book / ferx-r work is implied. `examples/` is untouched.

- [x] No example execution step needed (diagnostic-text fix, no user-visible change to any example's output)

## Reviewer hints

Focus on `omega_block_member_mask` in `src/estimation/parameterization.rs` — specifically the claim that `lower_tri_iter(n, om.diagonal)` alone is sufficient and a second `if om.diagonal` guard would be the redundant-gate hole CLAUDE.md warns about. The rest is a one-line substitution at the call site.

Worth a second look: the one-eta `block_omega (ETA_CL)` case. The parser appears to accept it, and it now reads as diagonal, so the suggestion would be `write \`omega ETA_CL ~ 0.0 FIX\`` for a line the user spelled as a block. That is a valid edit that does what they want, and the reasoning behind the message (`L_ii²` *is* the declared variance) holds exactly — but say so if you disagree.

## Open questions

None.


roninsightrx:

The mixed block/diagonal logic and its regression tests look sound, but there is one blocking edge case: valid one-eta block declarations are misclassified. No numerical or sensitivity paths changed. Per the constraint, I did not compile or run tests.

### Blocking findings

- `src/estimation/parameterization.rs:948` — **medium** — Block membership is inferred solely from the presence of a free off-diagonal. The accepted declaration `block_omega (ETA_CL) = [0.0]` has no off-diagonal, so this mask returns `false`; the diagnostic then pretends the user wrote `omega ETA_CL ~ 0.0` and suggests that spelling. This also contradicts the changelog’s claim that pure-block models are unchanged. The same issue applies to singleton `block_kappa`. Preserve declaration provenance independently of covariance structure, or explicitly normalize/reject singleton blocks, and add unit regressions for singleton `block_omega` and `block_kappa`.

### Nits

None.

roninsightrx:

Pushed as `ac2ce91b`. One finding, **fixed** — and the premise checked first, because it was moot if the parser rejected the declaration.

### `src/estimation/parameterization.rs:948` — one-eta blocks misclassified — **fixed**

**The premise holds.** A singleton block parses today; it is not rejected:

```
$ ferx check singleton.ferx          # block_omega (ETA_CL) = [0.0] + omega ETA_V ~ 0.04
error[E_OMEGA_INIT_AT_RAIL] parameters:1: `omega ETA_CL ~ 0.0` declares no variability …
    help: write `omega ETA_CL ~ 0.0 FIX`, or start it at 0.09
```

So the mask did quote back a line the user never wrote. That is the defect class the #1246 review already named on this same code path, and it is real.

**Where I'd push back is on the framing, and it changes the fix.** The finding reads as a regression introduced by this PR. It is not — I measured the pre-PR behaviour on the same file by restoring the old `block = !template.omega.diagonal` and re-running:

```
error[E_OMEGA_INIT_AT_RAIL] parameters:1: the `block_omega` Cholesky diagonal for ETA_CL … In a
block it is the **correlations** that do this … Reduce the declared covariances involving ETA_CL
(or raise its variance) …
    help: the block is near-singular in ETA_CL: lower the covariances involving it, or `FIX` the block
```

A 1×1 block has no covariances to reduce, so the old message was **unfollowable** on exactly this input — the same wrong-arm bug #1394 is about, one case further in. This PR narrowed the hole (the reasoning became sound: with no off-diagonals `L_ii²` *is* the declared variance) without closing it (the spelling stayed wrong). Severity is naming, not actionability. It is worth closing anyway, so I have.

**Why the diagnosis matters:** it rules out the obvious fix. Carrying provenance and feeding it to the *existing* bool would restore the message above — block spelling *and* the correlation story — which is strictly worse than what's on main today. The real defect is that one bool was answering two questions:

| question | answered by | one-eta block |
|---|---|---|
| is it the **correlations** that drive `L_ii` to zero? (message arm) | a free off-diagonal | **no** |
| how is the line **spelled**? (declaration + repair) | declaration provenance | **block** |

They agree on every block of ≥2 etas and come apart only here. So they are now two masks — `omega_correlated_diagonal_mask` and `omega_block_declared_mask` — and provenance drives only the spelling.

**Both of your suggested remedies, judged:**

- *"Preserve declaration provenance"* — taken. `OmegaMatrix::block_declared`, set in `build_omega_matrix`, which the `block_kappa` path already routes through, so Ω_IOV comes free. It is `pub(crate)` and defaults to all-false: **no public-API baseline change**, and no external impact — nothing outside the crate builds an `OmegaMatrix` by struct literal, and ferx-r uses `from_matrix` only (checked in the sibling checkout).
- *"or explicitly normalize/reject singleton blocks"* — **declined, and I'd argue against it.** Rejecting is a user-facing breaking change to a valid declaration (NONMEM's `$OMEGA BLOCK(1)` is ordinary) for a message-wording bug. Normalizing a singleton into a diagonal spec changes `omega.diagonal`, and with it the packed layout — `n(n+1)/2` coordinates become `n` — for every downstream consumer of a model nobody reported a numerical problem with. Both have a blast radius far larger than the diagnostic they'd fix.

Result:

```
error[E_OMEGA_INIT_AT_RAIL] parameters:1: `block_omega (ETA_CL) = [0.0]` declares no variability
but is not `FIX`-ed … NONMEM rejects the same declaration outright: `INITIAL ESTIMATE OF VARIANCE
CANNOT BE ZERO UNLESS FIXED`.
    help: write `block_omega (ETA_CL) = [0.0] FIX`, or start it at 0.09
```

Block spelling, declared-zero reasoning, no correlation story. Same for `block_kappa (KAPPA_CL) = [0.0]`.

### One hole the finding surfaced that it didn't name

Writing the fix, my first cut routed the *message arm* through the provenance flag. That would have **dropped the near-singular-block message for ferx-r**: an Ω rebuilt via `from_matrix` records no provenance but has real inferred off-diagonals, so a genuinely correlated eta would have been handed `omega NAME ~ 0.0`. `correlated_block_keyword` now deliberately does not consult provenance — correlation is always knowable from the matrix, spelling is not — with the reasoning at the call site. No existing test would have caught it, since every fixture parses from model source.

### Tests

`declaration()` became `zero_declaration()` and now carries the value: the two spellings have different shapes (`~ v` vs `= [lower triangle]`), so a keyword-only prefix can't be completed by the caller. Both put `FIX` last, which is what lets `fix_suggestion` append to either.

- `a_one_eta_block_omega_keeps_its_own_spelling_without_the_correlation_story` — asserts the premise that makes it the disagreement case (non-diagonal matrix, block-declared eta, **no** free off-diagonal), then both halves: block spelling *and* absence of the correlation story.
- `a_one_eta_block_kappa_keeps_its_own_spelling` — the Ω_IOV twin, pinning `block_kappa` and not `block_omega`.
- `the_one_eta_block_repair_parses_and_clears_the_diagnostic` — the suggested `= [...] FIX` is a form no other arm emits, so nothing else in the suite would catch a malformed one. It parses and clears the error.
- `test_one_eta_block_is_declared_but_not_correlated` — pins the two masks disagreeing, so collapsing them back into one bool fails whichever way it is collapsed.

Mutated three ways, each reddening both new tests and nothing stale:

| mutation | dies |
|---|---|
| message arm keyed on provenance (`block_keyword` instead of `correlated_block_keyword`) | both one-eta tests |
| spelling ignores provenance (drop the block branch in `zero_declaration`) | both one-eta tests |
| parser records no provenance (drop `block_declared[i] = true`) | both one-eta tests |

`cargo test --lib --features ci`: **4781 passed, 0 failed**. `tools/preflight.sh`: `preflight OK — fmt check clippy rustdoc docs public-api`.

CHANGELOG corrected — you were right that "pure-block models are unchanged" was false as written; it now says a block of ≥2 etas and a pure-diagonal model are unchanged, and states what a one-eta block does. `docs/file-formats/check-report.qmd` updated to match.

Issue #1382, opened 2026-09-13
Covariance estimator is unlabelled wherever SEs / condition number / eigenvalues are reported (#997 §5)


roninsightrx:
Split out of #997 §5.

## The problem

`FitResult` carries `cov_condition_number` and `cov_eigenvalues`, and prints standard errors, but records **nothing about which estimator produced them**. `covariance_method` lives on `FitOptions`, not on the result, so a fit object — or a `{model}-fit.yaml`, or a runlog — cannot say whether a condition number came from `R⁻¹`, `S⁻¹` or the `R⁻¹SR⁻¹` sandwich.

The reporter measured the same fit at condition number **1.42e8** under the default sandwich and **3.68e5** under `covariance_method = s`. Both are correct for their estimator. Nothing in the printed output distinguishes them, which matters the moment the figure is compared against another tool (NONMEM's default `$COVARIANCE` is a sandwich; `MATRIX=R` is not).

They also measured **3.68e5** under `gradient = auto` against **5.68e4** under `gradient = fd` — same data, same optimizer, same `covariance_method = s`. So a condition number arguably needs the gradient mode stated alongside the estimator before it can be compared to anything. `FitResult` does carry `gradient_method_outer` / `gradient_method_inner` already, so that half is a presentation gap rather than a missing field.

## Scope

- A `covariance_method` label on `FitResult` (the fit object, the YAML/JSON, and the `.fitrx` bundle).
- Name the estimator wherever the CLI prints SEs, the condition number or eigenvalues.
- Decide whether the `condition_number` warning's `details` payload should carry it too (`docs/warnings.qmd` lists that payload as just `condition_number`).
- ferx-r follow-up to surface it, since the R fit object is built from an explicit field list.

Related: #997 §1/§2 are addressed in #1380; §3 is superseded by #978; §4 is #1381.

PR #1425, opened 2026-09-18
feat(api): name the covariance estimator behind every SE, eigenvalue and condition number [closes #1382]

roninsightrx:
## Why

`FitResult` carries `cov_condition_number` and `cov_eigenvalues`, and prints standard errors, but
recorded **nothing about which estimator produced them**. `covariance_method` lives on `FitOptions`,
not on the result, so a fit object — or a `{model}-fit.yaml`, or a runlog, or a `.fitrx` bundle —
could not say whether a condition number came from `R⁻¹`, `S⁻¹` or the `R⁻¹SR⁻¹` sandwich.

The reporter measured the same fit at condition number **1.42e8** under the sandwich and **3.68e5**
under `covariance_method = s`. Both are correct for their estimator. Nothing in the printed output
distinguished them, which matters the moment the figure is compared against another tool: NONMEM's
default `$COVARIANCE` is `RSR`, ferx's default is `R`.

Reproduced on warfarin at `233246d1` — same data, same fit, two runs:

```
# covariance_method = r (default)          # covariance_method = rsr
--- THETA Estimates ---                    --- THETA Estimates ---
Parameter    Estimate       SE    %RSE     Parameter    Estimate       SE    %RSE
TVCL         0.132969 0.006627     5.0     TVCL         0.132969 0.007568     5.7
  Covariance: computed                       Covariance: computed
```

Two different SEs, two lines that read identically. Nothing on screen, in the YAML or in the bundle
said which was which.

## What changed

**One field, one owner.** `FitResult::covariance_method: Option<CovarianceMethod>` records the
estimator that produced `covariance_matrix`, the `se_*` fields, `cov_eigenvalues` and
`cov_condition_number`. `estimation::covariance::published_covariance_method(matrix, routed)` is the
single rule, and it owns both halves of the invariant.

**It is not a copy of `FitOptions::covariance_method`, and that is the point.** The option is what
was *asked for*. Two ways the two part company:

- **#1064's large-problem router.** Above `COV_HESSIAN_MAX_DIM = 100` free parameters a *defaulted*
  `r` is routed onto the cross-product, because the `R` matrix needs `n(n+1)/2` re-converged
  objective evaluations. A label read back off `FitOptions` reports `r` for SEs that came out of
  `S⁻¹` — wrong precisely on the fits where the user could not have predicted the answer.
- **`stage_opts` is a per-stage clone.** `fit_inner` never holds the `FitOptions` the covariance step
  ran under; it holds the caller's. Today the covariance keys happen not to be rewritten per stage,
  which makes a re-derivation *coincidentally* right rather than right.

So the label is carried up from the covariance step, not re-derived. `run_covariance_step_inner`
publishes `routed` — the same binding `compute_covariance` was configured with — through
`CovStepOutcome` → `OuterResult` → `FitResult`.

**`Some` exactly when `covariance_matrix` is `Some`.** The mirror-image failure is worse than the
reported one: a step that was skipped, failed, or fell back to SIR publishes no SEs, no eigenvalues
and no condition number, so an estimator name there invites a reader to compare a figure that was
never computed. `published_covariance_method`'s `None` arm is that rule, and G2 asserts it end to
end.

**Applied wherever the pair is published, not at one call site.** `OuterResult` is **public API with
public fields**, and `ferx-tools` calls `optimize_population` / `run_foce_gn` / `run_imp` /
`run_impmap` / `run_bayes` / `run_vi` directly without going near `fit_inner`, so a `fit()`-only
patch would leave that surface wrong. Sites:

| file | sites |
|---|---|
| `estimation/covariance.rs` | the populated `CovStepOutcome` and the closed-gate one |
| `estimation/outer_optimizer.rs` | NLopt exit, built-in BFGS exit, the `outer_maxiter = 0` evaluation exit |
| `estimation/gauss_newton.rs` | pure-GN exit and the FOCEI-polish hybrid exit |
| `estimation/impmap.rs` | the two MCEM exits |
| `estimation/saem.rs`, `estimation/trust_region.rs`, `estimation/vi/run.rs` | one exit each |
| `estimation/bayes.rs` | `None` — a Bayesian fit reports credible intervals, never a Hessian covariance |
| `estimation/run_covariance.rs` | the standalone re-run |
| `api/fit.rs` | the two synthetic evaluation-only `OuterResult`s and the `FitResult` assembly |
| `io/fitrx.rs` | the save and load sides of the bundle |

**`run_covariance` overwrites, it does not inherit.** It clones the incoming fit and rewrites the
covariance block. A clone that forgot to overwrite would report the *previous* fit's estimator for a
freshly computed matrix — the reported bug with an extra step. G3 pins it by re-running an `r` fit
under `s`.

**One spelling, not three.** The serde value, the `.fitrx` wire token, the YAML, the CLI and the
warning payload all read `CovarianceMethod::label()` → `r` / `s` / `rsr`, the same tokens
`[fit_options]` accepts, with `from_label` as its inverse. Derived serde would have emitted
`"CrossProduct"`; explicit `#[serde(rename)]`s make it `"s"`, and L3 asserts the *equality*
`serde == label()` rather than the two literals separately, so a later rename of one cannot drift
past the other. `formula()` (`R⁻¹` / `S⁻¹` / `R⁻¹SR⁻¹`) is printed beside the token, because the
token alone does not say what was inverted.

**Where it now appears** — the same warfarin run as above:

```
--- THETA Estimates ---
Standard errors from R⁻¹SR⁻¹ (covariance_method = rsr)
Parameter            Estimate           SE       %RSE
...
--- Run Info ---
  Covariance: computed (R⁻¹SR⁻¹, covariance_method = rsr)
```

```yaml
model:
  converged: true
  method: FOCE
  covariance_method: r          # new

covariance_matrix:
  # optimizer parameterization: ...
  estimator: r # R⁻¹            # new — the block bootstrap/SIR/uncertainty tools consume
```

plus the fit JSON, the `.fitrx` bundle, and the `condition_number` / `covariance_failed` /
`covariance_regularized` warning `details` payloads.

**The `details` decision the issue asks for: yes.** `docs/warnings.qmd` listed the
`condition_number` payload as the bare number, which is not actionable — an agent branching on it is
in exactly the reporter's position. The key is **omitted, never `null`**, when there is no matrix:
`null` reads as "unknown estimator for a number that exists", absent reads as "there was no
estimator", and only the second is true.

**Pre-existing bundles.** A `.fitrx` written before this field loads as `None` — the file does not
record it, which is what an unlabelled condition number meant all along. Defaulting to `Hessian`
would fabricate a provenance and be wrong for exactly the `s` / `rsr` bundles whose numbers most need
the label. An unrecognised token is `Corrupt`, the same verdict every other enum on that wire gets.

## Alternatives considered

**Re-deriving the label in `fit_inner` by calling the shared `scale_routed_covariance_method`.** One
function called twice is not two implementations, so this was not rejected on principle — it was
rejected on inputs. It would have to guess which stage's `FitOptions` and which model dimension the
covariance step actually used, and it cannot reach `run_covariance` at all, whose result is a clone
of a *different* fit. Carrying the value costs ~12 mechanical edits and removes the guess.

**Labelling a failed step with the estimator that failed.** Tempting for `covariance_failed`, and
rejected: the payload would then name an estimator for a matrix, a condition number and a set of SEs
that do not exist. The `None` arm keeps "there is a label" and "there is something to label"
equivalent, which is what makes the structural test in G5 expressible at all.

**Privatising the field behind a constructor** to make the invariant structural rather than scanned.
Same verdict as #1303 reached for `converged`: it is a breaking public-API change (`ferx-tools` reads
`OuterResult`'s fields) and wants its own baseline diff and ferx-r check. G5 buys the property
without the break.

## Cross-repo dependency
| ferx (R) | — | not needed for this PR | — |

The R fit object is built from an explicit field list, so surfacing `covariance_method` there is a
follow-up the issue already scopes (`ferx-r`). Nothing here breaks it: no existing field changed
shape or value, and the new field is additive with `#[serde(default)]`. Per CLAUDE.md, reaching
ferx-r needs the two-step (`ferx-r`'s `src/rust/Cargo.lock` bumped via
`tools/update-ferx-core-lock.sh`, then the downstream pin) — there is no ferx-core SHA to hand out.

## Breaking changes
- [x] `FitResult` / sdtab fields changed — **additive only**
- [x] Public Rust API — baseline regenerated
- [ ] None

`tools/update-public-api.sh` adds exactly five lines (each appearing twice, once per re-export path):

| item | who needs it |
|---|---|
| `FitResult::covariance_method` | the R wrapper and any tool reading a fit — the whole point of the issue |
| `OuterResult::covariance_method` | `ferx-tools`, which reads `OuterResult` from the five public `run_*` entry points without going through `fit()` |
| `CovarianceMethod::label` / `from_label` | the one spelling shared by the YAML, the `.fitrx` wire and `[fit_options]`; `from_label` is what makes a reported value pasteable back |
| `CovarianceMethod::formula` | printed beside the token wherever an SE or condition number is reported |

No field shape changed and no existing value moved. Both new fields are `Option`, default `None`, and
`FitResult`'s carries `#[serde(default)]`, so an old JSON/bundle deserializes unchanged. A
struct-literal construction of `OuterResult` outside the crate would need the new field — the usual
consequence of a public struct with public fields, and the reason #1303 recorded privatising it as
its own PR.

<details>

None for readers. A caller constructing `OuterResult` by literal adds `covariance_method: None` (or
the estimator, if it ran one). Nothing reads differently unless it wants to.

</details>

## Numerical validation
- [x] Not applicable — no estimator, likelihood, gradient or PK/ODE kernel changed. No
      `*_g<T: PkNum>` body was touched, so the `Dual2` path is untouched and no FD-parity check
      applies. The change is a label carried alongside numbers that are computed exactly as before.

**No NONMEM anchor, and the reason is not that one is awkward to build.** The quantity is *which of
ferx's own three estimators ferx just ran*. NONMEM has no spelling for that: `$COVARIANCE MATRIX=` is
an input to NONMEM, not something NONMEM reports back, and it has no analogue of #1064's
free-parameter router — the very thing that makes the reported value differ from the requested one.
The values are exact tokens (`"r"`, `"s"`, `"rsr"`) and an enum identity, not tolerances, so the
tests assert equality rather than agreement.

**The twin, for what did and did not move.** Warfarin, `examples/warfarin.ferx` + `data/warfarin.csv`:

```
# before (233246d1)                       # after (this PR)
OFV       = -280.3640                     OFV       = -280.3640     (unchanged)
TVCL      = 0.132969                      TVCL      = 0.132969      (unchanged)
TVV       = 7.730700                      TVV       = 7.730700      (unchanged)
TVKA      = 0.725207                      TVKA      = 0.725207      (unchanged)
SE(TVCL)  = 0.006627                      SE(TVCL)  = 0.006627      (unchanged)
Covariance: computed                      Covariance: computed (R⁻¹, covariance_method = r)
(no SE provenance line)                   Standard errors from R⁻¹ (covariance_method = r)
```

Every estimate, SE and diagnostic is byte-identical: this changes what the fit *says about* its
covariance, not what it computes. The suites are the other half of that evidence — 4972 lib tests and
220 integration binaries pass unchanged, including every `nonmem_anchor/*` binary.

## Performance
- [x] No significant impact expected — one `Option<Copy>` moved per published result, a handful per
      fit, none of it inside an objective evaluation or a covariance stencil.

## Tests
- [x] Unit tests added in the same module (`cargo test --lib` passes)

Twelve Tier-1 tests: five in `src/api/tests/covariance_method_label_tests.rs`, two in
`estimation/covariance.rs`, two in `types_tests.rs`, three in `io/fitrx.rs`'s inline module.

| | test | what it pins |
|---|---|---|
| G1 | `a_fit_reports_the_estimator_its_standard_errors_came_from` | **the reported gap**, end to end through `fit()`, all three estimators |
| G2 | `a_fit_with_no_covariance_matrix_names_no_estimator` | the mirror image: a name never outlives its matrix |
| G3 | `rerunning_the_covariance_step_relabels_the_result` | `run_covariance` overwrites rather than inherits |
| G4 | `the_condition_number_warning_payload_names_the_estimator` | the `details` payload, and that the key is *absent* rather than `null` |
| G5 | `every_published_covariance_matrix_is_written_with_its_estimator` | the pairing, structurally, per file |
| L1 | `the_published_label_names_the_routed_estimator_not_the_requested_one` | the #1064 routing composition — the case `FitOptions` gets wrong |
| L2 | `a_step_that_produced_no_matrix_publishes_no_estimator` | the `None` arm across all three estimators |
| L3 | `the_serialized_covariance_method_is_the_fit_option_token` | serde token **equals** `label()`, both directions |
| L4 | `each_covariance_estimator_prints_a_distinct_matrix_expression` | `formula()` distinguishes all three |
| F1 | `covariance_method_roundtrips_as_the_fit_option_token` | the bundle stores `rsr`, not `Sandwich` |
| F2 | `a_bundle_predating_covariance_method_loads_without_an_estimator` | an absent key is `None`, never a guess |
| F3 | `an_unknown_covariance_method_token_is_corrupt` | a damaged bundle errors instead of looking unlabelled |

plus two in `io/output.rs` for the console summary and the fit YAML.

| mutation | G1 | G2 | G3 | G5 | L1 | output |
|---|---|---|---|---|---|---|
| `published_covariance_method` hard-codes `Hessian` (**the fix off**) | **RED** | green | **RED** | green | **RED** | green |
| delete the `fit_inner` call site | **RED** | green | **RED** | green | green | green |
| delete the `run_covariance` assignment | green | green | **RED** | **RED** | green | green |
| `covariance_method: None` at SAEM's exit | green | green | green | **RED** | green | green |

Four things that matrix exists to show:

1. **Row 1 leaves G2 and G5 green on purpose.** G2's fixture has no matrix, so a wrong *value* is
   invisible to it; G5 checks that something is written, not what. Each row has at least one
   behavioural killer, and no row is carried by the scan alone.
2. **Row 3 is the asymmetry that justifies G3 existing.** Deleting the standalone re-run's assignment
   leaves G1 green — `fit()` never touches that code — which is exactly why `ferx-tools`' and the R
   wrapper's entry point needs its own fixture.
3. **Row 4 is the one only the scan catches**, and it is the reason the scan compares *values* rather
   than presence. It did not: the first version checked that a `covariance_method` field appeared in
   the literal, and `covariance_method: None` beside a real matrix — the exact defect — passed it
   green. Measured, then fixed; `writes()` now returns the value and `paired()` requires `None`
   beside `None` and a real expression beside a real expression.
4. **L1 is the only test that sees #1064's router**, and it is composed rather than driven end to
   end. Reaching that branch through `run_covariance_step_inner` needs a converged fit with >100 free
   parameters, whose cross-product would then be rank-deficient on any fixture small enough to be
   Tier-1 — so the honest test composes `scale_routed_covariance_method` with
   `published_covariance_method` and asserts the premise (`routed != requested`) first, so the case
   cannot quietly become one where both spellings give the same answer. **The residual gap is
   stated**: a mutation that passed the *pre-routing* method at the one call site is not caught
   behaviourally. It is mitigated structurally — the call site names the `routed` binding, which is
   the same one `compute_covariance` was configured with — not by a test, and I would rather say so
   than imply coverage that is not there.

### G5's allowances, stated exactly

Counted per file, both directions. `CovarianceMethod` on `FitOptions` is the *request* and is not a
site — no `covariance_matrix` accompanies it. `CovStepOutcome` is deliberately not a scanned literal:
its fields are `matrix` and `method`, so covering it would mean a second spelling for one struct, and
L1/L2 assert that struct's rule directly.

- [x] Not applicable — no new DSL syntax, `[fit_options]` key, estimator or sdtab column. The
      user-visible surface is a new output field and three new lines of console text, shown in the
      warfarin twin above and documented.

## Docs
- [x] `docs/` updated
- [x] `CHANGELOG.md` `[Unreleased]` → `### Added`

- `docs/output.qmd` — new `### Which covariance estimator produced these numbers
  {#covariance-estimator}` (the three estimators and their NONMEM equivalents, why the numbers are
  not comparable across them with the measured 1.42e8 / 3.68e5 pair, why the label is the estimator
  that *ran*, and where it appears); `covariance_method` added to the fit-YAML structure example and
  Key Fields; the `se` bullet now points at it.
- `docs/warnings.qmd` — `covariance_method` added to the `condition_number` and
  `covariance_failed` / `covariance_regularized` `details` rows, with the absent-not-`null` rule;
  the `condition_number` code row points at the new section.
- `docs/model-file/fit-options.qmd` — `covariance_method`'s row notes that the estimator which ran
  is reported back and is not always the one requested.

`cargo run -p docs-lint -- --check` is green with the baseline's five excused sections unchanged.
**No new R1 baseline entry**: the addition pushed `fit-options.qmd`'s "Covariance and standard
errors" from 4,965 to 5,115 characters, so the section is split under two new `####` subheadings
(`#covariance-estimator-choice`, `#covariance-hessian-settings`) rather than baselined. The existing
`#covariance-and-standard-errors` anchor is unchanged, so no inbound link breaks.

## Checklist
- [x] `tools/preflight.sh` green — all six groups (fmt, the 5 `cargo check` feature sets, clippy
      `--all-targets`, rustdoc, docs, public-API)
- [x] `Dual2` path — no `*_g<T: PkNum>` body touched; differentiability untouched
- [x] No version bump (additive API only)

```
cargo test --lib --no-default-features --features ci,survival --no-fail-fast
  -> 4972 passed / 0 failed / 26 ignored

cargo test --workspace --tests --no-default-features \
           --features ferx-core/ci,ferx-core/survival --no-fail-fast
  -> 220 binaries / 6873 passed / 0 failed / 202 ignored
```

### ferx-site / ferx-book

### Example execution
- [x] `cargo build --release --workspace` and
      `cargo run --release -p ferx-cli -- examples/warfarin.ferx --data data/warfarin.csv` run
      cleanly; the numbers are unchanged (twin above) and the new lines are the output quoted there.
      No example's estimates move, so no `.qmd` re-render is needed.
- [ ] ferx-r rebuild — not exercised: no ferx-r change is required by this PR, and per CLAUDE.md a
      local `R CMD INSTALL` would pick this branch up automatically when someone does surface the
      field there.

## Reviewer hints

- **Start with the mutation matrix, rows 3 and 4**, and with point 3 under it. Row 4 is the honest
  part of this PR: the structural test was written, run green, and could not catch the defect it
  exists for until `paired()` compared values instead of presence. If you think the fixed version
  still has a hole, that is the code to argue with (`covariance_method_label_tests.rs`, `writes` /
  `paired`).
- **Point 4 is the gap I would most like a second opinion on.** #1064's router is the whole reason
  this field is carried rather than derived, and it is the one arm with no end-to-end test, because
  the fixture would need >100 free parameters *and* enough subjects to keep the cross-product
  full-rank. I chose a composed unit test with an asserted premise over a slow, fragile Tier-3 one.
  Say the word if you would rather have the slow one.
- **The `None`-when-no-matrix rule** (`published_covariance_method`, and `FitResult::covariance_method`'s
  doc) is the other judgement call. It means `covariance_failed` warnings carry no estimator. I think
  naming one there is worse than naming none; the reasoning is on the function.
- **Skimmable:** the nine estimator exits are the same two-line pattern each
  (`method: covariance_method` in the destructure, `covariance_method,` in the `OuterResult`).
  `bayes.rs` is the one that is `None` by nature rather than by plumbing.
- **`io/output.rs`** has the only behaviour change beyond the field: `print_results` and
  `format_summary` previously spelled the four-arm `covariance_status` match twice; both now call
  `covariance_status_label`, so the status text is defined once.

## Open questions

None blocking. Two worth a look:

1. **Should the label be `Option<CovarianceMethod>` or a flag on the matrix itself?** The current
   shape lets the two fields disagree in principle; G5 is what keeps them in step. A
   `Option<(DMatrix, CovarianceMethod)>` would make it structural, and is a breaking change to a
   field `ferx-tools` reads — the same trade #1303 deferred for `converged`.
2. **The issue notes that a condition number arguably needs the *gradient mode* stated alongside the
   estimator** (3.68e5 under `gradient = auto` vs 5.68e4 under `gradient = fd`, same data, same
   `covariance_method = s`). `FitResult` already carries `gradient_method_outer` /
   `gradient_method_inner`, so that half is presentation only. I documented it in the new
   `output.qmd` section rather than adding a second token to every printed line — happy to put it on
   the `Condition number:` line too if you would rather.


roninsightrx:

Request changes. The estimator provenance is propagated correctly through the main fit paths, serialization, and public API, and no numerical kernels changed. However, the standalone covariance rerun can return stale structured provenance, and two new regression tests violate the repository’s test requirements. `git diff --check` was clean; no compilation or tests were run.

## Blocking findings

- `src/estimation/run_covariance.rs:341 — high — run_covariance updates covariance_method but clones warnings_structured unchanged. If the original fit’s covariance warnings carry method r and covariance is rerun with s, the returned FitResult reports s while its machine-readable warning details still report r; newly appended covariance warnings also receive no structured entry. This recreates the provenance error the PR is intended to fix. — Remove or refresh prior covariance-related warnings and rebuild their structured entries after updating the covariance fields; add a regression test that reruns with a different method and asserts the warning payload agrees with covariance_method.`

- `src/io/output.rs:3038 — medium — the condition-number regression assertion does not associate the method suffix with the condition-number line. Removing the suffix from that line leaves the test green because the suffix still appears in the SE note and covariance-status line. — Assert the complete line, such as "Condition number: 142000000.0  (covariance_method = rsr)\n".`

- `src/api/tests/covariance_method_label_tests.rs:124 — medium — G1, G2, and G3 are Tier-1 library tests that call fit(), including three covariance-producing fits in G1. CLAUDE.md explicitly assigns public-API fit tests with a handful of iterations to Tier 2 and tells Tier 1 to avoid fit(). — Move the end-to-end fit/run_covariance cases to tests/*.rs; retain the pure helper, payload, and structural checks at Tier 1.`

## Nits

- `docs/output.qmd:401 — low — this says covariance_method appears in covariance_failed payloads, but the implementation intentionally sets it to None when covariance production fails, and docs/warnings.qmd:279 confirms it is omitted for hard failures. CHANGELOG.md:35 makes the same claim. — Remove covariance_failed from those appearance lists or explicitly say that its failed/no-matrix case omits the key.`

roninsightrx:

All four findings are real. Fixed in c6d910f7; `tools/preflight.sh` green (all six groups), 4969 lib
tests + 221 integration binaries / 6873 tests, 0 failed. Point by point below, with the measurements.

---

### 1. `run_covariance.rs:341` — stale structured provenance — **fixed, and it was worse than described**

Confirmed by instrumenting the 12-subject fixture and re-running an `r` fit under `s`:

```
RERUN covariance_method  = Some(CrossProduct)
RERUN warnings_structured[2] = CovarianceRegularized, details:
    {"covariance_method": "r", "condition_number": 4091320907.897729,
     "min_eigenvalue": 9.7767e-10, "n_negative_eigenvalues": 0}
RERUN n_warnings = 7   n_structured = 7   (the new covariance warnings got no entry)
```

Two things beyond what the finding states:

- **The stale `condition_number` travels with it.** `cov_condition_number` was recomputed for the
  `s` step while the payload kept the `r` step's `4.09e9`. That half is *pre-existing* — it predates
  #1382, which only added `covariance_method` to the same payload. What #1382 changed is that the
  payload now also carries a provenance label, which turns a stale number into an outright
  contradiction with the field next to it. Fixing it here is right either way.
- **A plain `rebuild_warnings_structured` would have been the wrong fix.** The retained message is
  `"Covariance step regularized: eigenvalue floor applied to FD Hessian …"` — an `r`-step event that
  **cannot arise under the cross-product at all** (`covariance.rs` guards the floor warning with
  `options.covariance_method != CovarianceMethod::CrossProduct`). Re-enriching it would stamp the
  `r` step's message with the `s` step's numbers: a different wrong answer.

So the fix is both halves, in order:

```rust
out.warnings.retain(|w| !is_covariance_step_warning(w));
out.warnings.extend(new_warnings);
crate::api::rebuild_warnings_structured(&mut out);
```

`is_covariance_step_warning` classifies via `crate::types::classify_warning` rather than matching
prose, so the set tracks the classifier instead of drifting from it, and is scoped to the four codes
`new_warnings` replaces (`CovarianceStep`, `CovarianceFailed`, `CovarianceRegularized`,
`ConditionNumber`). `Sir` is deliberately **not** in the set: `run_covariance` discards the SIR
fallback proposal and never runs the sampler, so a SIR warning is not this function's to replace.

**One thing I did not do, deliberately.** `inflated_rse` and `high_correlation` are also stale after
a re-run — they are derived from SEs that just changed. I left them, because `run_covariance` does
not re-derive them at all: that is `fit_inner`'s post-fit diagnostic pass, which this function
deliberately does not run (its doc enumerates the covariance fields it refreshes). Replacing the
warnings this function *recomputes* is a clean line; silently dropping diagnostics it cannot
recompute would leave the result with neither the old nor a new value. Flagging it as a separate,
pre-existing issue rather than smuggling it into this PR — happy to file it if you agree.

**Regression test**, as requested: `tests/covariance_method_label.rs::rerunning_the_covariance_step_relabels_the_result_and_its_payloads`.
It asserts the premise first — that the incoming fit really does carry a covariance-step warning
whose payload names `r` — because without that there is no stale provenance and the test would pass
on an empty list. Then the typed field, then that **no** payload names a superseded estimator, then
that the flat and structured lists have equal length.

Each of the three lines of the fix was mutated separately and each kills the test on its own:

| mutation | result |
|---|---|
| delete `out.covariance_method = …` | **RED** (typed field assertion) |
| delete the `retain` | **RED**, quoting the retained `{"covariance_method": "r", "condition_number": 4.09e9}` payload |
| delete `rebuild_warnings_structured` | **RED** (flat/structured length assertion) |

---

### 2. `io/output.rs:3038` — the condition-number assertion — **fixed, and it was a live hole**

Correct, and verified rather than taken on faith. The old form was

```rust
s.contains("(covariance_method = rsr)") && s.contains("Condition number: 142000000.0")
```

which the SE note satisfies on its own. I applied the exact mutation — stripped the suffix from the
condition-number line only, leaving the SE note and status line intact — and the old assertion
**passed**. It now asserts the whole line:

```rust
s.contains("  Condition number: 142000000.0  (covariance_method = rsr)\n")
```

which fails under that same mutation. This is the "two redundant gates cover for each other" shape
CLAUDE.md warns about, and it is the second one this PR has had (the structural scan had a
presence-vs-value version of the same fault, noted in the PR body). Both are now value-asserting.

---

### 3. `covariance_method_label_tests.rs:124` — Tier 1 vs Tier 2 — **fixed, you are right on the letter and I checked the consequence**

G1/G2/G3 call `fit()` and return after one outer iteration, which is verbatim CLAUDE.md's Tier-2
description ("Call the public API (`fit()`, …) but must return immediately … after a handful of outer
iterations"), while Tier 1 says "avoid calling `fit()`". Moved to `tests/covariance_method_label.rs`.

I checked the thing that would have made this a bad trade before doing it: CLAUDE.md says Tier-2
files are "compile-checked on every PR and run nightly", which read as if moving them would stop them
running on PRs and cost patch coverage. It does not — `.github/workflows/ci.yml:458`, the
`Tests + coverage (core)` job, runs

```
cargo llvm-cov --workspace --tests --no-fail-fast --no-default-features --features ferx-core/ci …
```

and `--tests` builds and runs the `tests/*.rs` targets. So both tiers execute on every PR and the
move is free. (The "run nightly" sentence is about the `slow-tests`-gated Tier 3.)

The Tier-1 sibling keeps exactly what you suggested — the structural scan and the payload check — and
now calls no `fit()` at all; its module doc states the boundary so the next person does not put an
end-to-end case back. The pure-helper checks were already Tier 1 and untouched
(`published_covariance_method`'s two unit tests in `estimation/covariance.rs`, the serde-token and
`formula()` tests in `types_tests.rs`).

---

### 4. `docs/output.qmd:401` / `CHANGELOG.md:35` — `covariance_failed` — **fixed (nit accepted)**

Right, and the inconsistency was internal to the PR: `docs/warnings.qmd:279` already said the key is
omitted for a hard failure, while `output.qmd`'s appearance list and the changelog entry named
`covariance_failed` unqualified. `covariance_failed` means the step produced no matrix, so by the
rule the PR itself states there is no estimator to name. Both now say so explicitly rather than just
dropping the code, since a reader grepping for it should learn *why* it is absent:

> - the `condition_number` and `covariance_regularized` warning `details` payloads … Not
>   `covariance_failed`: that code means the step produced no matrix, so by the rule above there is
>   no estimator to name and the key is omitted.

---

### Not changed

Nothing in the review was rejected. The one item I have scoped out rather than fixed is the stale
`inflated_rse` / `high_correlation` payloads after a `run_covariance` re-run (see finding 1) — a
pre-existing gap in a function that never re-derived those warnings, and a different change from the
one this PR is making.

The open question from the PR body still stands and is unaffected by the above: #1064's
large-problem router (a defaulted `r` re-routed to `s` above 100 free parameters) is the reason the
label is carried rather than derived, and it is the one arm with no end-to-end test, because the
fixture would need >100 free parameters *and* enough subjects to keep the cross-product full-rank.
It is covered by a composed unit test with an asserted premise
(`the_published_label_names_the_routed_estimator_not_the_requested_one`), and the residual gap is
stated in the PR body rather than papered over. Still happy to build the slow version if you would
rather have it.

PR #724, opened 2026-07-07
feat(parser): 2-cpt transit ODE-equivalent twin (fix 1-/2-cpt asymmetry)

TeunP:
## Why
`transit_ode_equivalent_source` matched only `one_cpt_transit`, so a `two_cpt_transit` model with time-varying covariates or a `TIME`-dependent structural parameter was **rejected** by `check_transit_support`, while the identical 1-cpt model transparently rerouted to its ODE twin — a 1-/2-cpt asymmetry.

## What changed
Extend the desugar to the plain `two_cpt_transit(cl,v1,q,v2,n,mtt)` form, emitting the canonical 2-cpt disposition (`central` + `periph`; `k10=cl/v1`, `k12=q/v1`, `k21=q/v2`; observed on central) — the exact form `ode_template::generate`'s `TwoCptTransit` case and the existing `build_pair_2cpt` equivalence tests already validate against the closed form. `effective_for` / `check_transit_support` are generic (gated on `transit_ode_equivalent.is_none()`), so the twin now routes 2-cpt TV-cov/`TIME` subjects automatically; IOV / SS / infusion / reset stay rejected.

## Relationship to #719
Unblocks **#719 gap 3 (analytic-transit IOV), route (b), for 2-cpt transit**: the "reuse the ODE-equivalent dispatch for IOV subjects" route was 1-cpt only until now. With the 2-cpt twin in place, IOV support reduces to dropping the `n_kappa > 0` reject (`api.rs:1261`) + adding occasion-presence to the `effective_for` reroute condition — for both compartment counts. (Prerequisite, not a close.)

## Breaking changes
- [x] None (previously-rejected models now work)

## Numerical validation
- [x] Transitively NONMEM-anchored: the twin's `transit()` forcing is anchored (`tests/transit_nonmem_anchor.rs`) and its form matches the closed form via the existing `two_cpt_transit_*_matches_ode` equivalence tests.
- [x] New `two_cpt_transit_time_desugar_matches_hand_written_ode` (asserts `transit_ode_equivalent.is_some()` + reroute PRED matches a hand-written 2-cpt ODE across a mid-profile `TIME` switch). All 27 `transit_analytic_equivalence` tests pass.

## Tests
- [x] `cargo test --test transit_analytic_equivalence` (27 pass); parser module (527) + all examples parse; rustfmt clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1426, opened 2026-09-18
fix(api): stop recommending [diffusion] as a remedy for IWRES autocorrelation

TeunP:
## Why

`refs #1285`, which is [#1350](https://github.com/FeRx-NLME/ferx-core/issues/1350) Tier 0 row 10a — the text half only. The [decision comment](https://github.com/FeRx-NLME/ferx-core/issues/1285#issuecomment-5726121298) defers the EKF mean update to 0.5.0; the [plan](https://github.com/FeRx-NLME/ferx-core/issues/1285#issuecomment-5726777767) scopes this PR to what 0.4.0 *says* about `[diffusion]`. **No numerical change.**

`solve_ekf` propagates the state covariance and shrinks it at each observation, but never corrects the state mean with the observation: `kalman_update` takes no `y`, and `likelihood.rs` keeps only `p_obs` (`let (_, p_obs) = …`). What is fitted is therefore the deterministic ODE prediction with an inflated observation variance, not a state-space likelihood — and three places told users otherwise:

1. The Durbin-Watson warning appended *"For ODE models, SDE process noise may also help."* on **every** ODE fit. That is the one that matters: it fires on ordinary, non-experimental models and sends the user into an experimental feature for a reason it cannot serve. Nothing tested the suffix, which is how it outlived the filing of #1285.
2. `docs/model-file/diffusion.qmd` described `P` as "updated at each observation" (reads as a full EKF), called residual autocorrelation "the primary signal", and said a diffusion term "often absorbs the systematic drift".
3. `docs/model-file/derived.qmd` referred to an "EKF posterior mean" that does not exist.

Measurements carried from the plan's §0 (four debug FOCE fits at `e5bd9067`, one variable per pair, DW recomputed from sdtab `IWRES` outside the engine). On a 2-cpt population fitted as 1-cpt (40 × 14 obs, 5 % proportional):

| fixture | OFV | DW | mean(IWRES²) | fitted CL (truth 1.0) | σ_prop |
|---|---|---|---|---|---|
| ODE | 692.67 | 0.403 | 0.94 | 0.99 | 0.385 |
| ODE + `central ~ 0.01` | 23.30 | 0.766 | 1.0e10 | 2.49 | 0.0165 |

So the hint is not merely inert: ΔOFV is −669 for one parameter, the statistic still trips its own threshold, and CL moves 2.5× away from the value the ODE fit had recovered. (The plan corrects the decision comment's "provably cannot move that statistic" — that holds at fixed (θ, η) only.)

## What changed

**Code (text only).**

- `src/api/fit.rs` — the inline DW block becomes `dw_autocorrelation_warning(dw: f64) -> Option<String>`, next to the existing `bobyqa_scale_warning` helper. The SDE sentence and the `model.ode_spec` dependency are gone; both messages are otherwise byte-identical to what `fit_inner` produced before. A free function so the exact text is reachable from a unit test without running a fit.
- `src/api/validation.rs` — `W_EXPERIMENTAL_SDE` says what the maturity label cannot: the filter is covariance-only, the state mean is never corrected by the observed data, there is no filtered state trajectory, and sdtab `IWRES` is scaled by the residual error alone so it reads over-dispersed.

**Docs / example text.** `docs/output.qmd` (the warning's own description), `docs/model-file/diffusion.qmd` (what is actually fitted; "When to use SDE models" rewritten around the table above; new `## What the filter does not do` section), `docs/examples/sde.qmd` ("When to use", and the "check whether the DW statistic improves" advice), `docs/model-file/derived.qmd` (no posterior mean to differ from), `examples/warfarin_sde.ferx` (header comment), `CHANGELOG.md` under `[Unreleased] → Changed`.

**Untouched, deliberately, and named so the diff can be checked against it:** `src/ode/ekf.rs`, `src/stats/likelihood.rs`, `src/stats/residual_error.rs`. `git diff --stat` shows nothing under `src/ode/` or `src/stats/`.

## Alternatives considered

Fixing the EKF instead of the text. Deferred to 0.5.0 by the decision comment: the mean update changes what `IPRED` means on a `[diffusion]` model and needs the first NONMEM anchor this path has ever had (`SDE.f90` under `METHOD=1 LAPLACE`, since ferx refuses IMP/SAEM here), a Tier-1 closed form to 1e-10, and the ρ = 0 vs 0.9 straddle pair. Not a 0.4.0 item.

Keeping the hint but qualifying it ("may help, but see …"). Rejected: the warning fires on ordinary ODE fits, and the measured behaviour is a fit that is re-weighted rather than corrected. A qualified recommendation is still a recommendation.

## Cross-repo dependency

| ferx (R) | FeRx-NLME/ferx-r#384 | open (issue) | after |

ferx-r carries its own copy of the hint (`R/ferx_get_warnings.R:276`, pinned by `test-ferx_get_warnings.R:259-262`) and ships `inst/examples/models/warfarin_sde.ferx`. R-side text only — **no lock bump needed**, and it does not block this PR. Kept out of scope on purpose (plan §6); filed as FeRx-NLME/ferx-r#384, where the hint turns out to be *toggled* by `uses_sde` and its test pins both directions.

## Breaking changes

- [x] None

No `pub` item added (`dw_autocorrelation_warning` is `pub(crate)`), so the public-API baseline is unchanged.

## Numerical validation

- [x] Compared against: prior ferx commit `e5bd9067`
- [x] Reference values:

```
# before (e5bd9067, plan §0, debug)
examples/warfarin_sde.ferx --data data/warfarin_sde.csv   Final OFV = -279.397740

# after (this branch, debug, same command)
examples/warfarin_sde.ferx --data data/warfarin_sde.csv   Final OFV = -279.397740
```

No NONMEM run: nothing numerical changes. The objective is untouched by construction — the diff is a warning string, a `W_` message and prose — and the fit above re-measures it rather than asserting it.

## Performance

## Tests

Suites run locally under `--no-default-features --features ci,survival`: `cargo test --lib` → **4973 passed, 0 failed, 26 ignored**; `cargo test --tests --no-fail-fast` → **197 binaries, 5959 passed, 0 failed, 189 ignored**, exit 0.

New sibling `src/api/tests/dw_autocorrelation_warning_tests.rs` (4 tests) plus one assertion added to the existing `sde_emits_experimental_warning`. Assertions are **exact string equality**, not `contains`: the removed sentence is a *recommendation*, so a `!msg.contains("SDE")` assertion would pass on any rewording that still sends the user to `[diffusion]`.

Every new assertion was mutation-tested individually (`--no-fail-fast`); each mutation kills exactly one test:

| mutation | test that dies |
|---|---|
| re-append `" For ODE models, SDE process noise may also help."` | `positive_autocorrelation_message_is_exact` |
| swap the 1.5 / 2.5 thresholds | `no_warning_inside_the_band` |
| delete the `!dw.is_finite()` guard | `non_finite_dw_never_warns` |
| reword the negative-branch message | `negative_autocorrelation_message_is_exact` |
| delete the new sentence from `W_EXPERIMENTAL_SDE` | `sde_emits_experimental_warning` |

Not re-asserted here: the warning's *category*. `classify_warning` keys on "autocorrelation" / "durbin" and `types_tests.rs` already pins both messages to `dw_autocorrelation`; a second copy would cover for the first. The category assertion inside `sde_emits_experimental_warning` is load-bearing in the other direction and is commented as such — that chain tests "autocorrelation" **before** "experimental feature", so explaining the SDE limitation in those words would silently re-file `W_EXPERIMENTAL_SDE` under `dw_autocorrelation`.

- [x] Not applicable (message/prose change; `examples/warfarin_sde.ferx` header updated in place)

## Docs

- [x] No new pages (existing pages only; `docs/_quarto.yml` unchanged)

`cargo run -p docs-lint -- --check` is green; no new `docs/.lint-baseline` entry (the new section is well under R1's 5,000 characters, and every new `#anchor` resolves).

## Checklist

## Docs & examples

- [x] No new DSL syntax, `[fit_options]` key, example file or data file — nothing for ferx-site or ferx-book to mirror.
- [x] `examples/warfarin_sde.ferx` re-run via the CLI (see Numerical validation); its only change is a comment.

## Reviewer hints

Where to focus:

- **The two strings.** `dw_autocorrelation_warning`'s two messages must be byte-identical to the old ones minus the suffix — the tests pin them, but they are what a user sees.
- **The new `W_EXPERIMENTAL_SDE` wording** and the `classify_warning` ordering note above.
- **`docs/model-file/diffusion.qmd`**: this PR and [#1413](https://github.com/FeRx-NLME/ferx-core/pull/1413) both edit it, and both touch `src/api/tests/sde_integration.rs` and `CHANGELOG.md`. Different sections in every case (#1413 is the SS/`W_SDE_STEADY_STATE` rows at the bottom of the gaps table and the doc comment of the test *after* `sde_emits_experimental_warning`); whoever lands second rebases.

Skimmable: the docs prose, which is the bulk of the diff.

## Open questions

1. **sdtab `IWRES` on `[diffusion]` fits** (mean square 1e10 measured above) — scaling by `R + p_obs` is a numerical/output change, so it is out of scope here. Plan §6 leaves the decision on #1285: interim fix now, or wait for the mean update? The docs currently tell users not to read those residuals.
2. `docs/index.qmd:24` still advertises "Extended Kalman Filter path for stochastic differential equations" in the feature list. Left alone as a method name rather than a behaviour claim, with the correction one click away on the SDE page — say the word and I will qualify it.


TeunP:
## Review at `1d816920`

Read against the [plan](https://github.com/FeRx-NLME/ferx-core/issues/1285#issuecomment-5726777767). Nothing numerical is wrong and nothing here blocks; five should-fix items, the first one measured. CI being green is not used as evidence for any of it.

| # | Finding | Severity | Evidence | Suggested fix |
|---|---|---|---|---|
| 1 | The new tests pin `dw_autocorrelation_warning`, not what `fit_inner` pushes. Re-appending the SDE sentence **at the call site** — the exact pre-PR user-visible behaviour — kills no test. | should-fix | Mutation run on this head: `if let Some(mut msg) = dw_autocorrelation_warning(dw_statistic) { if model.ode_spec.is_some() && dw_statistic < 1.5 { msg.push_str(" For ODE models, SDE process noise may also help."); } warnings.push(msg); }` → `--workspace --features ferx-core/ci --lib --no-fail-fast`: 61 + 4785 + 603 passed, **0 failed**, all four `dw_autocorrelation_warning_tests` and `sde_emits_experimental_warning` green. No file under `tests/` or `crates/` observes DW text (grep). The test's doc comment says "adding the `model.ode_spec` dependency back in any other form" must redden it; it does not. | One fast `fit()` on an ODE model whose DW lands outside the band, asserting the `FitResult.warnings` autocorrelation entry `==` `dw_autocorrelation_warning(result.dw_statistic)`. Then re-run this mutation. Also correct the doc comment. |
| 2 | `docs/examples/sde.qmd` Tips, first bullet, still says a ΔOFV > 3.84 (2.71) "justifies the addition". The table this PR adds shows ΔOFV = −669 next to CL moving 2.5× off truth. | should-fix | `docs/examples/sde.qmd` "Compare OFV" bullet, unchanged by the diff; contradicts the new paragraph two screens above it and `diffusion.qmd#when-to-use-sde-models`. | Qualify the bullet: with the mean uncorrected, ΔOFV measures re-weighting, so it is not evidence the term is warranted; point at the measured example. |
| 3 | "Nothing for ferx-site to mirror" is not right: the site's `examples/sde.qmd:8-9` carries the same recommendation ("useful when IWRES shows autocorrelation that cannot be absorbed by residual error alone"). | should-fix | `ferx-nlme.github.io` `origin/HEAD`. ferx-book grep for the same phrases: clean. | Site PR, or an issue naming the two lines. |
| 4 | The ferx-r copy of the hint (`R/ferx_get_warnings.R:276`, pinned by `test-ferx_get_warnings.R:259-262`, plus `inst/examples/models/warfarin_sde.ferx:6`) has no tracking issue. R users read guidance through that file, so for them the recommendation survives this merge. | should-fix | `gh issue list -R FeRx-NLME/ferx-r` for SDE / diffusion / 1285: nothing. PR table says "not yet opened". | File the ferx-r issue before merging so plan §1 ("nothing ferx prints…") has an owner on the R side. |
| 5 | The measured fixture (OFV 692.67 / 23.30, DW 0.403 / 0.766, CL 0.99 / 2.49, IWRES² 1e10, σ 0.0165) now appears in two docs pages, the CHANGELOG and a `fit.rs` doc comment, but its dataset, seed, models and script are neither in the repo nor on #1285. | should-fix | #1285 comments contain the table only. I could not re-measure −669, and did not. | Attach the generator + two model files to #1285 (or commit them) and cite that from `diffusion.qmd`. Consider dropping the numbers from the code comment so there is one copy fewer to update in 0.5.0. |
| 6 | `diffusion.qmd:121` says the warning "repeats the first and third points"; the message also states the second ("There is no filtered state trajectory"). | nit | `validation.rs` new text. | "repeats all three". |
| 7 | The new `W_EXPERIMENTAL_SDE` sentence speaks of "what is fitted" and "sdtab IWRES … on these fits", but the diagnostic is also surfaced by `predict()` / `simulate()` (`postfit.rs:2170`) and `ferx check`, where no fit ran. | nit | caller list of `check_experimental_features`. | Reword to not presuppose a fit, or leave and accept. |
| 8 | `docs/file-formats/check-report.qmd:208` and the registry description at `src/diagnostics.rs:56` still describe the code as a maturity notice only. | nit | unchanged rows. | One clause + link to "What the filter does not do". |
| 9 | "(issue #1285)" is baked into a user-facing string; stale once the issue closes. | nit | `validation.rs`. | Keep the pointer in the docs page only. |
| 10 | The DW warning still fires on `[diffusion]` fits, from the IWRES the same PR tells users not to read (DW 0.766 on the SDE arm of the fixture). DW is scale-invariant, so the σ collapse alone does not break it, but two warnings on one fit now disagree about whether IWRES is interpretable. | question | follows from the PR's own table. | Out of scope here (output change); fold into open question 1 on #1285. |

### Traced and found correct

- `kalman_update` takes no `y`, and `likelihood.rs` keeps only `p_obs` — the PR's central claim.
- "Do not read `IWRES`, `CWRES`": `compute_cwres` has no `p_obs` argument, so the CWRES half is right too.
- The `lkel` claim (`d/dt(lkel) = 0` + `[diffusion] lkel ~ 0.01` "parses and fits"): ran it through the CLI on `data/warfarin_sde.csv`, 15 outer iterations, exit 0.
- `classify_warning` order: none of the new wording hits an arm ahead of `"experimental feature"`; the category assertion goes through `classify_warning(&exp.message)`, so the "load-bearing" comment is accurate.
- Old and new DW strings are byte-identical apart from the suffix (string-continuation whitespace is stripped).
- Author's five mutations: consistent by reading (the `is_finite` guard is killed by ±∞, not by `NaN`, as the test comment says). Not re-run.
- Old-contract grep over `src/ tests/ examples/ docs/ crates/`: no remaining recommendation. `docs/index.qmd:24` and the `SS=1` Tips bullet are as flagged in the PR's open questions.
- `cargo run -p docs-lint -- --check` green; `git merge-tree` against `origin/main` `85413d90` clean; `closingIssuesReferences` empty, so #1285 stays open.

### Not done yet

The clean full suite on this head (`ferx-core/ci` `--lib`, `--tests`, and `ci,markov,nn` `--tests`, all `--no-fail-fast`) is still running locally; counts follow in a second comment. The only completed local run so far is the mutated `--lib` leg above.

Fixes: `/ferx-resolve`. A re-review is a fresh pass.

TeunP:
## Resolve round 1 — at `c7c38c13`

Every finding below is either fixed here or has an issue with an owner. Finding 1 was reproduced before it was fixed, and your mutation was re-run afterwards.

| # | Finding | Fix |
|---|---|---|
| 1 | Tests pin the helper, not what `fit_inner` pushes | **`b54367e5`** — new `fit_pushes_exactly_the_helpers_message`: a 2-subject, 9-observation ODE fixture (1-cpt model against a biexponential truth), `outer_maxiter: 1`, no covariance step, **no `omega`** — with a random effect on CL the inner loop moves each subject's elimination rate and absorbs the drift the fixture exists to create. It asserts `ode_spec.is_some()`, asserts the straddle (**realised DW = 0.26**, bound `< 1.5`), asserts exactly one autocorrelation warning, then `warnings[i] == dw_autocorrelation_warning(result.dw_statistic)`. **Your mutation re-run verbatim on this tree: `test result: FAILED. 5 passed; 1 failed`** — only the new test dies, and the two strings differ by exactly the suffix. Runtime 0.08 s. The doc comment on `positive_autocorrelation_message_is_exact` no longer claims to catch a caller-side re-append; it points here instead |
| 2 | `sde.qmd` "Compare OFV" still reads Δ OFV > 3.84 / 2.71 as justification | **`c7c38c13`** — the bullet now says what those thresholds actually test: with the state mean uncorrected the improvement is re-weighting, not explanation, and it cites the measured row (Δ OFV −669 for one parameter while CL moves 2.5× off truth) and links `#when-to-use-sde-models` |
| 6 | "repeats the first and third points" | **`c7c38c13`** — "repeats all three points" |
| 7 | Message presupposes a fit, but `predict()` / `simulate()` / `ferx check` emit it too | **`c7c38c13`** — "a `[diffusion]` fit scores … rather than a filtered state", and "sdtab IWRES **from such a fit**" |
| 8 | `check-report.qmd:208` and `diagnostics.rs:56` still maturity-only | **`c7c38c13`** — both registry rows now carry "the filter is covariance-only, the state mean is never corrected by the data"; the docs row links `#what-the-filter-does-not-do` |
| 9 | "(issue #1285)" baked into a user-facing string | **`c7c38c13`** — dropped; the pointer is "the Feature Maturity and SDE pages" |
| 3 | Site carries the same recommendation | **FeRx-NLME/ferx-nlme.github.io#30** — verified on that repo's default branch: `examples/sde.qmd:8-9`. Line 110 is adjacent and reads correctly today, but says "absorbing" in the other sense, so the issue flags it for a re-read |
| 4 | ferx-r copy of the hint | **FeRx-NLME/ferx-r#384** — and it is worse than a copy. `.ferx_warning_guidance()` *toggles* the hint on `uses_sde`: suppressed on models that already carry `[diffusion]`, shown on every **other** ODE model, which is exactly the population it should no longer reach. `test-ferx_get_warnings.R` pins both directions (`expect_match(pos_ode, "SDE process noise")`), so the fix has to flip an assertion, not delete a line. No lock bump — R-side text only |
| 5 | The measured fixture has no dataset, seed or script | **#1427**, not fixed here, deliberately. A re-measurement can disagree with the published table, and then `diffusion.qmd`, `sde.qmd`, the CHANGELOG entry and the `fit.rs` doc comment all need correcting — which turns a text-only PR into a numbers PR mid-review. #1427 owns reproduce-then-reconcile, and carries your suggestion to drop the numbers from the code comment |
| 10 | DW still fires on `[diffusion]` fits | Agreed, out of scope here; folded into open question 1 on #1285 with the `IWRES` scaling decision |

### Gates at `c7c38c13`

`tools/preflight.sh` → `preflight OK — fmt check clippy rustdoc docs public-api`. `cargo run -p docs-lint -- --check` green. `git merge-tree` against `origin/main` `85413d90` clean. `git diff origin/main...HEAD` is 12 files, +385/−45, with **zero lines under `src/ode/` or `src/stats/`**.

Your review comment's "counts follow" line was never filled (that run was killed when this resolve started in the same worktree), so here is the full set, all `--no-fail-fast`:

| suite | result |
|---|---|
| `--workspace --features ferx-core/ci --lib` | 3 binaries, **5450 passed, 0 failed**, 26 ignored |
| `--workspace --features ferx-core/ci --tests` | 220 binaries, **6539 passed, 0 failed**, 203 ignored |
| `--features ci,markov,nn --tests` | 197 binaries, **6166 passed, 0 failed**, 197 ignored |

The `ci,markov,nn` leg first died on `ld: write() failed, errno=28` — the volume was full at 400 MiB free, not a defect — and was re-run from a clean target after reclaiming 84 GiB.

CI on this head is green on all ten checks, including `Tests (release semantics)` (7m48s) and both coverage jobs (`core` 43m35s, `TTE/CTMM` 18m48s).

### What a re-review should attack

The new fixture is the only real code in this PR, and it is unreviewed. Two claims in it are worth adversarial attention: that the no-`omega` model at `outer_maxiter: 1` pins DW at 0.26 *for the reason the comment gives*, and whether a **different** minimal call-site edit — not the one you mutated — still slips past the equality assertion.

TeunP:
## Review round 2 at `c7c38c13`

Fresh pass; the resolve reply's claims were treated as hypotheses. Nothing numerical is wrong and nothing blocks on correctness. The fix for round-1 finding 1 is **incomplete**: a different minimal call-site edit restores the recommendation and the whole lib suite stays green. The PR also now conflicts with `main`. CI green is not used as evidence for anything below.

| # | Finding | Severity | Evidence | Suggested fix |
|---|---|---|---|---|
| 1 | `fit_pushes_exactly_the_helpers_message` only looks at warnings that contain the word `autocorrelation`, so pushing the removed recommendation as its **own** warning survives everything. | should-fix | Mutation at the `fit_inner` call site: after `warnings.push(msg);` add `if model.ode_spec.is_some() && dw_statistic < 1.5 { warnings.push("For ODE models, SDE process noise may also help.".to_string()); }`. `--workspace --features ferx-core/ci --lib --no-fail-fast`: **61 + 4786 + 603 passed, 0 failed**. Positive check that the mutant is reached, not skipped: on the test's own fixture `FitResult.warnings` prints both the DW message and `For ODE models, SDE process noise may also help.` while the new test reports `ok`. | Assert over the whole `result.warnings`, not over a filter keyed on the helper's own vocabulary — e.g. every entry other than `expected` must classify into a known category (I expect the stray sentence to fall through `classify_warning` to the fallback arm; not verified), or pin the fixture's category set. Then re-run this mutation and round 1's. |
| 2 | The fixture has no `omega`, so re-appending the old suffix for every model **with** IIV passes: `… && dw_statistic < 1.5 && model.n_eta > 0 { msg.push_str(…) }` → all five `dw_autocorrelation_warning_tests` green. The comment's reason for dropping `omega` ("absorbs part of the very drift … would make the Durbin-Watson value depend on the EBE search") is a story, not a measurement. | should-fix | Measured on the fixture, one variable at a time: no omega **0.2645** (`outer_maxiter: 1`) / **0.2845** (converged); with `omega ETA_CL ~ 0.09` **0.2934** / **0.2845** — ω collapses (ETA shrinkage 98 %), a scalar on CL cannot bend a mono-exponential into a bi-exponential. Worst case 0.29 against a bound of 1.5. | Put the `omega` back (it is free against the bound, and it moves the fixture onto the analytic ODE provider — see note below) and replace the comment's mechanism with the measured numbers. |
| 3 | PR is `CONFLICTING` with `origin/main` `53dc25f4`. #1413 (`79e0cfaf`) deleted the `W_SDE_STEADY_STATE` row adjacent to the edited `W_EXPERIMENTAL_SDE` row in `src/diagnostics.rs`, and rewrote the same SDE surface (`diffusion.qmd`, `sde.qmd`, `validation.rs`, `sde_integration.rs`, `check-report.qmd`). | should-fix | `git merge-tree --write-tree origin/main HEAD` → `CONFLICT (content)` in `src/diagnostics.rs`; the other overlapping files auto-merge. I ran `docs-lint --check` on the merged docs tree: green, and the two new anchors (`#what-the-filter-does-not-do`, `#when-to-use-sde-models`) coexist with #1413's `{#sde-steady-state}`. No cargo gate has run on the merged tree; the counts in the resolve reply are for a tree that will not be the merged one. | Merge `main`, resolve (keep this PR's row, drop `W_SDE_STEADY_STATE`), check the resolution with `git diff origin/main -- src/diagnostics.rs` (adds only), re-run preflight + the two `--tests` legs. The PR body's open question about the `SS=1` Tips bullet is moot after #1413 — drop it. |
| 4 | The test's doc comment says equality catches "any caller-side edit … a suffix, a prefix, or a second `push` of a reworded copy". Finding 1 is a second push and is not caught. | nit | Same over-claim shape as round-1 finding 1's doc comment. | Follows from fixing 1; otherwise narrow the sentence. |
| 5 | The fixture is not identifiable: `ode(obs_cmt=central, …)` scores the **amount**, so `TVV` enters only through `CL/V`. | nit | Predictions are `100·e^{-(CL/V)t}`; realised θ = `[0.994, 10.34]` at maxiter 1 and `[0.996, 10.20]` converged — it lands, but along a flat direction. | `FIX` TVV, or leave with a one-line note. |

### The three questions asked of this round

**1. Does DW = 0.26 come from the stated mechanism?** Yes. Realised IWRES for subject 1: `[-0.50, -0.89, -1.45, -1.97, -1.96, -1.48, -0.80, 0.03, 1.03]` — one smooth sign change across nine points, i.e. the U the comment describes, not a scale offset (I checked that specifically: the model scores the amount, so there is no 10× `V` mismatch hiding in there). It is not incidental to any of the three suspects:

| variant (one change from the fixture) | DW |
|---|---|
| as committed (`outer_maxiter: 1`, FD inner-gradient route, proportional) | 0.2645 |
| converged (14 iterations) | 0.2845 |
| with `omega` → analytic ODE provider, no FD warning | 0.2934 / 0.2845 converged |
| additive error | 0.2882 / 0.3048 converged |

And it cannot pass on an empty search: `is_finite() && < 1.5` is asserted before the filter, then `emitted.len() == 1`. That half of the claim holds.

**2. A different minimal call-site edit.** Two survive — findings 1 and 2. Round 1's re-append does die, as the reply says (not re-run; the equality makes it true by reading).

**3. `classify_warning` and the new wording.** Holds. Lower-cased, the new message contains `experimental feature` and none of the earlier arms' keys: it says `covariance-only`, not `covariance step` / `covariance failed`; `convergence behaviour`, not `did not converge` / `without convergence`; no `autocorrelation`, `durbin`, `degenerate`, `condition number`. `fit()` pushes the bare `d.message` (`fit.rs:1373`), which is the same string the test classifies, so the category assertion sees what users get. `never corrected by the observed data` sits on a single source line of the literal, so the `contains` is not at the mercy of a continuation break. Also checked against `main`'s chain after #1413: same arm.

**4. The fixture numbers say the same thing everywhere.** `diffusion.qmd` 692.67 / 23.30 (Δ = 669.37 → "669" ✓), DW 0.403 / 0.766 ↔ `fit.rs` doc comment and CHANGELOG "0.40 to 0.77" ✓, CL 0.99 / 2.49 in all three, "2.5×" in `diffusion.qmd` and `sde.qmd` (2.49 / 0.99 = 2.52 ✓), IWRES² 1e10 / 0.94 and σ 0.0165 only in `diffusion.qmd`. `sde.qmd`'s "−669" sign is consistent with "improves by 669". Not re-derived (#1427). Dispositions for round-1 findings 3, 4, 5, 10: no disagreement.

### Out of scope, noticed on the way

The committed fixture (no `omega`, `n_eta = 0`) emits "2 of 2 subjects use finite-difference inner gradients (outside the analytic provider's scope …)". The same ODE **with** an eta emits no such warning, so the model is in scope; with no eta there is no inner gradient at all and the message's diagnosis is wrong for that case. Pre-existing, not this PR's. I found no existing issue for it; will file one separately unless it is already known.

### Traced / not traced

Ran: both mutations above (1 on the full `--workspace ferx-core/ci --lib` leg, 2 on the `dw_autocorrelation_warning_tests` module only — it is a strict subset of round 1's mutation, which killed nothing else); the DW measurement (seven fits; an analytic-twin arm I added failed to parse on my side and was dropped, not measured); `docs-lint` on the merged tree; `dw_autocorrelation_warning_tests` + `sde_emits_experimental_warning` on the clean head after restoring (both files `shasum`-verified, worktree clean). Did **not** re-run the `--tests` legs or `ci,markov,nn` — they will need re-running on the merged tree anyway (finding 3).

### Convergence

This round found a hole in the previous round's fix. One more round like that and the right move is to stop iterating on the wiring test: ship the text/docs change (converged since round 1) and cut the test hardening to its own issue.

Fixes: `/ferx-resolve 1426`. A re-review is a fresh pass.

TeunP:
## Resolve round 2 — at `ffd9b663`

Merged `main` first, since the round-2 gates were measured on a tree that would not be merged. Then fixed findings 1, 2, 4, 5 — and then ran a narrow adversarial pass over the fix itself, which found two more holes of the same family. Those are fixed here too.

| # | Finding | Fix |
|---|---|---|
| 3 | PR `CONFLICTING` with main | **`d67e3e85`** — merged `origin/main`, by then `504297fc` (main moved past the `53dc25f4` you saw). One textual conflict, `src/diagnostics.rs`, resolved as you said: this PR's `W_EXPERIMENTAL_SDE` row kept, `W_SDE_STEADY_STATE` stays deleted; `git diff origin/main -- src/diagnostics.rs` is that one row and nothing else. Merged docs are coherent — #1413's `{#sde-steady-state}` section and its rewritten Tips bullet sit alongside `#what-the-filter-does-not-do` and `#when-to-use-sde-models`. PR is **MERGEABLE** again. PR body: open question 3 dropped (settled by #1413), cross-repo row now names FeRx-NLME/ferx-r#384 |
| 1 | Test filtered on `contains("autocorrelation")`, so a separate `push` survived | **`e88e0e7c`**, hardened in **`ffd9b663`** — the assertion is over the whole `result.warnings`. **Your suggested category gate does not work, and that is measured, not argued**: `classify_warning` sends an unrecognised sentence to the `General` fallback and this fixture already emits two legitimate `General` warnings (default-method notice, evaluation-budget notice), so the stray sentence hides among them. The six ODE-arm entries classify as General ×2, Convergence, Threads, EtaShrinkage, DwAutocorrelation |
| 2 | No `omega`, so an `n_eta > 0`-gated suffix was invisible; the comment's reason was a story | **`e88e0e7c`** — `omega ETA_CL ~ 0.09` restored, and the comment now carries your numbers (0.2645 / 0.2845 without, 0.2934 / 0.2845 with; ω collapses, η shrinkage 98 %) instead of the EBE-absorption story. Realised DW on the committed fixture: **0.2953** (ODE arm), **0.2848** (SDE arm) |
| 4 | Doc comment over-claimed | **`e88e0e7c`** — rewritten to what the assertion actually does, with the round-2 hole recorded in it |
| 5 | Fixture not identifiable — `TVV` only via `CL/V` | **`e88e0e7c`** — `TVV` is `FIX`ed |
| — | n_eta = 0 FD-gradient warning | Out of scope, **#1432**; the `omega` fix moves this fixture off that path as a side effect |

### The narrow pass over the fix, and what it found

Two further edits got past the round-2 fix. Both are the same family as the two you found — an assertion that inspects a *subset* of what the code emits:

- **Append to an already-accounted warning.** The allow-list matched by opening words, so attaching the recommendation to the shrinkage notice kept its prefix and stayed accounted. Every arm now pins **both** ends; the thread notice's machine-dependent count is handled by matching its opening anywhere rather than as a prefix.
- **Gate the suffix on `model.is_sde()`.** The fixture had no `[diffusion]` block, so it could not observe that gate at all — and `[diffusion]` users are the population that must least of all be told to add process noise. There are now two fixtures differing in exactly that one thing, each asserted (`!ode.is_sde()`, `sde.is_sde()`), sharing one assertion body.

The test also checks that its own gate can fire, in both shapes that got past a weaker version: the recommendation as its own entry, and the recommendation appended to each warning the fit actually emitted.

### Mutations

Every cycle restores `fit.rs` by byte copy and shasum-verifies it (`434fe177`), `--no-fail-fast`. All five fail `fit_pushes_exactly_the_helpers_message` and nothing else — `4 passed; 1 failed` each:

| mutation | source | result |
|---|---|---|
| separate `warnings.push("For ODE models, SDE process noise may also help.")` | r2 finding 1 | killed |
| suffix gated on `model.ode_spec.is_some() && dw < 1.5 && model.n_eta > 0` | r2 finding 2 | killed |
| plain re-append to `msg` | r1 finding 1 | killed |
| append the sentence to an accounted warning | this pass | killed |
| suffix gated on `model.is_sde()` | this pass | killed |

### Gates

`tools/preflight.sh` and the three suite legs are **running on the merged tree now**; I will post the counts in a follow-up rather than leave a "counts follow" line unfulfilled, since that is what happened to the round-1 and round-2 runs. What is already established at this head: `cargo fmt --all -- --check` green via the pre-commit hook, and the DW + `W_EXPERIMENTAL_SDE` tests green on the `ci` feature set.

### Convergence

Two rounds in a row found a hole in the previous round's fix, so by the rule in your review the next one would mean splitting. This pass tried to pre-empt that by attacking the fix mechanically instead of by reading — the two holes above took minutes to find by running mutations after reading had missed them twice. If a fresh pass still finds a hole in this test, the right move is to ship the text/docs change, which has been stable since round 1 and traced clean twice, and cut the wiring test to its own issue.

TeunP:
Follow-up to the round-2 review above: the out-of-scope item (a fit with `n_eta = 0` emitting "N of N subjects use finite-difference inner gradients") is now filed as #1432. Nothing to do for it in this PR.

TeunP:
### Gate counts at `ffd9b663` (the merged tree)

The follow-up promised above. All four exit 0, `--no-fail-fast` throughout.

| gate | result |
|---|---|
| `tools/preflight.sh` | `preflight OK — fmt check clippy rustdoc docs public-api` |
| `--workspace --features ferx-core/ci --lib` | 3 binaries, **5515 passed, 0 failed**, 26 ignored |
| `--workspace --features ferx-core/ci --tests` | 223 binaries, **6616 passed, 0 failed**, 203 ignored |
| `--features ci,markov,nn --tests` | 200 binaries, **6244 passed, 0 failed**, 197 ignored |

CI on this head is green on every check, and `codecov/patch` reports 28 hits / 0 misses / 100 % — worth noting because the re-sync here is a merge commit, which has previously confused codecov's patch computation on this repo (#724). It did not this time.

Issue #1428, opened 2026-09-18
fix(estimation): #1277 follow-up — the mid-descent restart has no PR-time test, grants a second maxiter budget, and is silent in FitResult.warnings


hiddevandebeek:
Follow-up to #1414 (`48f274a6`), carrying the three pieces of #1411 its review asked to keep ([TeunP's review](https://github.com/FeRx-NLME/ferx-core/pull/1411#issuecomment-5728951076)), verified against `origin/main` 2026-09-18.

## What is wrong on `main`

1. **No PR-time test can see the restart.** #1414's tests are unit tests of the pure `resolve_mid_descent_restart` / `resolve_stall_retry` fns plus the Tier-3 nightly test. Measured: with `first_line_search_abort_is_resumed_not_reported` (from #1411) ported onto `main` and `mid_descent_stall` forced `false`, the full `--lib --features ci,markov,nn` suite is **5166 passed / 1 failed — only that test**. Until it is ported, the only guard on #1277 is the nightly, which is how #1277 slipped through the first time.
2. **The restart grants a second evaluation budget, and the docstring says it cannot.** `resolve_mid_descent_restart`'s doc: "the restart can never quietly hand a fit a second evaluation budget". The restart closure calls `optimize_nlopt_once(.., options, ..)` unchanged, so `maxiter = 1` (44 evals) spent **56** (12 + a fresh 44), measured by TeunP. #1411's rule — resume on the *remaining* `outer_maxiter` budget; budget exhausted while still descending → `max_eval_reached` — is the correct semantics.
3. **A restarted fit is silent.** Only a `verbose` `eprintln!`; nothing in `FitResult.warnings`, against the CLAUDE.md convention. #1411's `optimizer_health` warning + `classify_warning` arm + `docs/warnings.qmd` row fill that.

## Decisions folded in

4. **Optimizer gate.** `stationarity_check_pending` is set on `Failure | ForcedStop` from *any* NLopt algorithm, so BOBYQA/SLSQP/MMA bare failures restart too. #1411's `resume_descent` was L-BFGS-only (the mechanism is Luksan's line search; SLSQP is capped every eval, the derivative-free ones take no line search). Decide, and pin it.
[...]

PR #1429, opened 2026-09-18
fix(estimation): restart a mid-descent stall on the remaining maxiter budget, say so in FitResult.warnings, and pin it at PR time [closes #1428]

hiddevandebeek:
## Why

#1428 — the follow-up #1411's review asked for when it was closed as superseded by #1414 (`48f274a6`). Three things on `main`, verified 2026-09-18:

1. **Nothing at PR time can see the restart.** #1414's tests are unit tests of the pure `resolve_mid_descent_restart` / `resolve_stall_retry` fns plus the Tier-3 nightly test. With `mid_descent_stall` forced `false`, the full `--lib --features ci,markov,nn` suite was 5166 passed / 1 failed — the one failure being #1411's `first_line_search_abort_is_resumed_not_reported`, ported. The only guard on #1277 was the nightly, which is how #1277 slipped through the first time.
2. **The restart granted a second `maxiter` budget and documented that it could not.** The restart closure called `optimize_nlopt_once(.., options, ..)` unchanged. Measured here: `maxiter = 1` (44 evals at `n = 43`) spent **56** — 12 by the stalled leg plus a fresh 44 — under a docstring saying "the restart can never quietly hand a fit a second evaluation budget".
3. **A restarted fit was silent** in `FitResult.warnings` (verbose `eprintln!` only), against the warnings convention in CLAUDE.md.

## What changed

- **`RestartLeg { escape_from, spent_evals }`** replaces the bare `escape_from: Option<&ModelParameters>` on `optimize_nlopt_once` — both fields exist only for the restart leg. The budget formula is spelled once in `outer_eval_budget(algo, n, outer_maxiter)` (gradient methods `maxiter × (n+1)`, BOBYQA plus its `40 × (n+1)` headroom) and the leg gets `full − spent`, `.max(1)`.
- **`worth_restarting_mid_descent` takes `budget_left`**: a stall on the last permitted evaluation is reported, not restarted — `set_maxeval(0)` means *unlimited* to NLopt. A restart that runs out of budget while still descending is adopted (strictly lower) and carries the existing "hit the evaluation budget (maxiter = N); increase maxiter" warning, which is now true of the fit as a whole.
- **`mid_descent_restart_warning`** is pushed on the restart's own result, so it reaches `FitResult` only when the restart is adopted; `classify_warning` files "resumed from the best point seen" under `optimizer_health`; `docs/warnings.qmd` row updated.
- **Docs / docstring mechanism corrected.** `main` explained the eval-12 abort as clearing the `(s, y)` pairs. That abort is Luksan's *extrapolation* cap inside the first line search — before any pair exists — after the identity-Hessian overshoot cap shrank the opening gradient 1.7e5× (#1411, `plis.c` instrumented); a fresh run helps there by re-arming that cap. The *reduction*-cap shape later in the fixture is where the fresh curvature memory helps. Both are now in `foce.qmd` and the `resolve_mid_descent_restart` doc.
- **Optimizer gate decided: none**, and the reason is written down. The predicate is "bare `Failure` while still descending, at a point a cold re-solve does not call a plateau", which SLSQP and the derivative-free methods can also produce; the strictly-lower adoption rule makes a pointless restart cost one optimization and change nothing. (The `left_init` retry stays L-BFGS-only as before.)

## Alternatives considered

- **Rebase #1411 whole.** Two restart mechanisms layered; ruled out in #1411's own closing.
- **Reduce `outer_maxiter` on a cloned `FitOptions` for the restart.** Rounds `remaining / (n+1)` and duplicates the BOBYQA arithmetic; the shared `outer_eval_budget` + `spent_evals` is exact.
- **Gate the restart on L-BFGS like #1411's `resume_descent`.** See above; the mechanism argument applies to the *cause*, the predicate is about the *trace*, and the adoption rule bounds the cost.

## Cross-repo dependency

## Breaking changes
- [x] None (no public-API change; baseline unchanged. `RestartLeg` is `pub(crate)`)

## Numerical validation
- [x] Compared against: `main` (`79e0cfaf`) on the #1277 fixture `tests/fixtures/two_cpt_dcm_regularized.ferx`, `nn_l2 = 0`, `maxiter = 1` (44-eval budget), Linux debug build
- [x] Reference values:

```
main (fresh budget on the restart):   OFV −271.343075   n_iterations 56   (44 budget)
this PR:                              OFV −229.562609   n_iterations 44
  warnings: "Outer optimizer (nlopt_lbfgs) stopped mid-descent (bare Failure at OFV = 7901.994208
             after 12 evaluations) and was resumed from the best point seen on the remaining
             maxiter budget; ..." + "hit the evaluation budget (maxiter = 1) ... increase maxiter"
restart wiring off (mutation):        OFV 7901.994208   n_iterations 12   no restart warning
```

Fits that are not restarted take the identical path (`RestartLeg` is `None`, `spent_evals = 0`, the budget is the same number as before). `outer_optimizer::` + `classify_warning` lib suites: 129 passed.

## Performance
- [x] No significant impact expected — a restarted fit now spends *at most* `maxiter`, i.e. less than on `main`; everything else is unchanged.

## Tests
- [x] Unit tests: `test_mid_descent_restart_needs_budget_left`, `test_outer_eval_budget_is_shared_by_both_legs`, `classify_warning_mid_descent_restart_is_optimizer_health`; `test_mid_descent_restart_skips_a_non_finite_objective` updated for the new argument
- [x] Regression test that fails without the fix: `nn::regularizer_fit_tests::first_line_search_abort_is_resumed_not_reported` (nn-gated `--lib`, ~8 s debug here). Mutations run: `mid_descent_stall` forced `false` → 7901.994 at eval 12, no warning, dies on the warning assert; `spent_evals: 0` → 56 evals, dies on the `n_iterations ≤ 44` bound alone. Platform note from #1411's review: the abort sits at 7902 (Linux) / 3209.81 (macOS); `ofv < 0` holds for both.

- [x] Not applicable (fix; the only new surface is a warning)

## Docs
- [x] `docs/estimation/foce.qmd` *When the optimizer stops without converging* — remaining budget, the two line-search caps, the warning, the no-gate decision
- [x] `docs/warnings.qmd` `optimizer_health` row

## Checklist
- [x] `tools/preflight.sh` green — fmt, check matrix, clippy, rustdoc, docs, public-API
- [x] No `Dual2` path touched

## Docs & examples
- [x] No example execution step needed (fix; docs page updated; no new DSL / fit option)

## Reviewer hints

- `RestartLeg`, `outer_eval_budget`, the `budget_left` argument and `mid_descent_restart_warning` are the whole change in `outer_optimizer.rs`; the rest is docstrings.
- The `budget_left` read is `spent_evals + n_evals_outer < full_budget`, with `n_evals_outer` incremented once per objective call — the same count NLopt's `maxeval` uses.
- Not gated on the optimizer, deliberately — see the paragraph on `resolve_mid_descent_restart`.

## Open questions

None. #1350 gets a row for #1428 once this lands.


hiddevandebeek:
Both findings applied in the follow-up commit:

1. **Declined restart now warns.** `budget_left` is computed once next to `stalled_mid_descent`, and a stall with `!budget_left && !max_eval_reached` pushes its own "stopped mid-descent with its evaluation budget (maxiter = N) spent, so it was not restarted; increase maxiter" warning — the case the review found, where Luksan's line search fails after crossing `maxeval` and returns a bare `Failure` instead of `MaxEvalReached`. Docstring updated to say that is why the ordinary budget warning cannot cover it.
2. **Test bound is against the pre-fix number.** `n_iterations < 12 + budget` (= 56, what #1414 as merged spends here; a fresh-budget restart cannot come in under it) instead of `<= budget`, with the soft-bound reason in the comment. `foce.qmd`, the `resolve_mid_descent_restart` doc and the CHANGELOG now say "bounded by `maxiter` the way a single run is" rather than "never exceed".

Preflight (fmt/check/clippy/rustdoc/docs/public-api) green; the 8 affected tests pass (`first_line_search_abort_is_resumed_not_reported` still 44 evals, −229.56).

PR #1440, opened 2026-09-18
fix(api): close the three review limits #1423 left open — variant list, predict/simulate filter reporting, cmt=1 false positive [refs #1409]

roninsightrx:
> Follow-up to #1423, which merged as `b713a113` while this work was in progress —
> the merge landed between two of my pushes, so the branch carried these three
> commits' worth of fixes and they did not go with it. Cherry-picked onto `main`,
> re-gated there. Nothing here changes what #1423 shipped; it closes the three
> limits that PR knowingly left open.

## Why

#1423's own review listed six findings. Three were fixed in that PR; three were
left open with the reasons recorded. All three turned out to be fixable without a
design decision, and two of them are user-visible.

## What changed

**1. The `CmtConsumer` list can no longer go stale.** The enum and its `ALL` list
are expanded from one `cmt_consumers!` token sequence, so a variant cannot exist
without being in the list.

Two earlier spellings both had a hole, each found with the same probe — a 7th
variant carrying arms in every `match`:

| spelling | probe result |
|---|---|
| `ALL: [CmtConsumer; 6]` beside an exhaustive `index()` | green: the array simply stayed at six |
| a `next()` chain | green: a variant whose arm was `None` was unreachable from the iterator |
| **`cmt_consumers!` macro** | **red**: `every_cmt_consumer_has_a_fixture_where_it_is_the_only_live_one` fails |

The last is the wanted behaviour — a new channel is not done until it carries a
fixture. #1423's doc claimed the first spelling already had this property; it did
not, which is what the probe was written to settle.

**2. `predict()` / `simulate()` / the adaptive driver report a filter deletion.**
They are handed a `Population` the caller already read and have no `&FitOptions`,
so they passed a default and suppressed a `W_CMT_DEFAULTED` whose entire content
was rows a `CMT`-reading `[data_selection]` rule had deleted. Reachable from
ferx-r, since `read_population_for_simulation` is `pub` and takes a
`SelectionFilter`.

Rather than thread options through four public entry points — the objection that
left this open — the finding now carries its own proof. The reader only counts
such a row when the rule that removed it actually read `CMT`, so a summary
containing `CMT_FILTER_DELETED_MARKER` is never suppressed, on any entry point.
Keyed on the reader's own constant so the two cannot drift.

**3. A lone endpoint at the default compartment is no longer a false positive.**
An endpoint-only model whose routing names only `CMT=1` has nowhere else to send a
defaulted row: the fallback *is* the endpoint, and any other compartment is an
`E_ENDPOINT_UNROUTED` error. The warning was noise. Suppressed via
`ObsRouting::routes_only(DEFAULT_CMT)` together with the "scores Gaussian
observations" question, now the shared `model_scores_gaussian_observations`.

Both halves are load-bearing, and they are pinned apart because neither is
implied by the other:

| mutation | what goes red |
|---|---|
| drop the Gaussian half | `an_endpoint_at_the_default_compartment_still_reports_when_gaussian_rows_exist` |
| drop `routes_only` | `an_endpoint_only_models_placeholder_pk_model_is_not_read_as_a_dose_channel`, `the_empty_error_map_and_the_routing_set_are_the_same_models` |

Without the first, a `cmt = 1` endpoint model that *also* scores ordinary
observations stops reporting — although a defaulted row there genuinely picks
between the endpoint and the Gaussian grid. Without the second, the
competing-risks measurement this whole channel exists for stops reporting.

## Alternatives considered

- **Thread `&FitOptions` into `non_fit_diagnostics`** for finding 2. Rejected
  again: `predict_diag(model, population, params)` is public and has no options to
  thread, and synthesizing one reports on defaults the caller never chose. Keying
  on the reader's own marker answers the question without widening any signature.
- **`std::mem::variant_count`** for finding 1 — nightly-only, and the test jobs
  build on stable. A derive dependency for one enum was not worth it; the macro is
  ~20 lines and local.
- **Suppressing every endpoint-only model at `cmt = 1`** for finding 3, without the
  Gaussian test. That is the mutation in the table above, and it silently withdraws
  the warning from a real ambiguity.

## Cross-repo dependency

No public API change — `merge_selection_exprs`, `routes_only`,
`CMT_FILTER_DELETED_MARKER`, `model_scores_gaussian_observations` and
`CmtConsumer` are all `pub(crate)`; `tools/update-public-api.sh --check` is green.

## Breaking changes
- [x] None

Two behaviour changes, both narrowing noise or widening a finding, neither moving
a number: an endpoint-only `cmt = 1` model stops getting `W_CMT_DEFAULTED`, and
`predict`/`simulate` start reporting a filter-deleted guess they used to drop.

## Numerical validation
- [x] Not applicable — no number moves. This decides which *diagnostic* is shown;
      every prediction, objective and estimate is bit-identical. `W_CMT_DEFAULTED`
      has no NONMEM equivalent at all (`$MODEL` declares `DEFDOSE`; ferx has no such
      declaration), which is why #1009 exists — so the scope is anchored on measured
      behaviour per case rather than against a second engine. The endpoint-only
      `cmt = 1` case is asserted on **bit-identical** objectives under both spellings,
      which is what "the guess changes nothing" means there.

## Performance
- [x] No significant impact expected — one `contains` on a warning string, and one
      extra set test per reader warning.

## Tests
- [x] Unit test(s) added (`cargo test --lib` — 5045 passed)
- [ ] No new tests needed

New: `an_endpoint_at_the_default_compartment_still_reports_when_gaussian_rows_exist`
(Tier 1), plus `a_filter_deleted_row_is_reported_by_every_entry_point_including_predict`,
`predict_still_suppresses_a_cmt_guess_that_addresses_nothing`,
`a_lone_endpoint_at_the_default_compartment_is_not_reported` and
`an_endpoint_model_that_routes_more_than_the_default_is_still_reported` (Tier 2).
Every one mutation-checked; the table above names which mutation kills which.

Two of them are `survival`-gated, which puts `tests/cmt_data_selection_scope.rs`
into the endpoint job's `--test` list — the same set equality
`tests/ci_workflow_endpoint_coverage.rs` pins. Found by running the base
`--features ci` set, which is exactly where #1423's ungated test failed; that
lesson is why the base set was run here before pushing.

One fixture correction worth flagging: the obvious straddle for finding 3 — a lone
endpoint at `cmt = 2` — is **not** usable, because the defaulted row leaves the
endpoint for a Gaussian grid the model has no error model for and ferx stops with
`E_PER_CMT_ERROR_MODEL`. Loud, not silent, so not what this warning is for. The
straddle is competing risks at `cmt = 1`/`cmt = 2` instead, where the row really
does re-route silently.

## Docs
- [x] `docs/data-format.qmd`, `docs/file-formats/check-report.qmd` updated — the
      `cmt = 1` limit they documented as open is now described as closed, with the
      condition under which it still reports.
- [x] `CHANGELOG.md` `[Unreleased]` updated in place (the #1409 entries are already
      there from #1423).
- [ ] No user-visible change

## Checklist
- [x] `tools/preflight.sh` green on this branch, on top of `main`
- [x] Nothing on the `Dual2` path touched
- [ ] `Cargo.toml` version bump — not breaking

## Reviewer hints

The load-bearing judgement is in `model_routes_rows_by_cmt`: the two-condition
suppression. If you can construct an endpoint-only model with routing `{1}` where a
defaulted row still changes a number, the `routes_only` half is wrong and should go.

The macro is mechanical; the thing to check is that it is genuinely one list —
`cmt_consumers! { .. }` expands to both the enum and `ALL`.

## Open questions

`ferx check` still reads with `filter: None`, so it and `fit()` answer slightly
different questions about the same dataset (noted on #1423, pinned by
`ferx_check_reads_unfiltered_so_it_answers_a_different_question`). Still out of
scope here — say if you want it filed.


roninsightrx (commented):
Reviewed PR head `4da2f63c76e334f381b146b6ba4786b5d9ba720d` against its merge base `b713a113`.

One minor documentation finding:

- **[P3] Name the correct error for an observation outside the lone endpoint** — `docs/data-format.qmd:181` (also `src/api/run.rs:1178`). With an endpoint-only model at CMT 1, an observation explicitly assigned CMT 2 enters the Gaussian vectors, and the empty error map produces `E_PER_CMT_ERROR_MODEL`. `E_ENDPOINT_UNROUTED` instead detects Gaussian observations on a *declared endpoint* CMT that were loaded without endpoint routing. Please correct the code named here; the argument that this is a loud failure still holds.

No runtime correctness findings. The enum and `ALL` now come from one declaration; the filter-deletion marker is shared with the reader and survives the common non-fit diagnostics path; and the endpoint suppression requires both an empty Gaussian dispatch and routing exclusively to the reader's default compartment. Other CMT consumers still independently trigger the warning.

Validation:

- All 23 `reader_warning_suppression_tests` passed with `ci,markov,nn` enabled.
- All 11 `cmt_data_selection_scope` tests and the endpoint workflow coverage guard passed with `ci,markov` enabled under `ci-fast`.
- `cargo fmt --all -- --check` and `git diff --check` passed.

CI needs attention before merging, but the compiler failure is inherited from current `main`, rather than introduced by this diff: `src/estimation/vi/run_tests.rs:1671` calls `compute_bounds` without importing it. This test was added by `eaac92e4` (#1305), after this PR's branch point, and it causes Check, Clippy, release-semantics and endpoint-coverage jobs to fail on the merge commit. See the [failed Check job](https://github.com/FeRx-NLME/ferx-core/actions/runs/35370861515/job/105684224485).

The reported 50% patch coverage is incomplete because endpoint coverage never uploaded. Fix the inherited import failure and rerun CI to obtain the merged coverage result and verify the required 90% patch gate.

roninsightrx:
**CI was red for a reason that is not this PR: `main` itself does not compile under any feature set that includes `nn`.**

`src/estimation/vi/run_tests.rs:1671` calls `compute_bounds` without it being in scope — the file's `use super::*` gets `pack_params` and `clamp_to_bounds` from `vi::run`, but not that one. Introduced by #1305 (`eaac92e4`). GitHub builds a PR merged with its base, so this PR inherited it.

The same `error[E0425]` is the whole of all four failures here:

| job | cause |
|---|---|
| Check | `cargo check --tests --features ci,nn,slow-tests` → E0425 |
| Clippy | same E0425 |
| Tests (release semantics) | same E0425 |
| Tests + coverage (TTE/CTMM endpoints) | same E0425 (`ci,markov,nn`) |
| Tests + coverage (core) | **passed** — `ferx-core/ci` compiles no `nn` |

`main`'s own runs are failing too (`afb4462d`, `80f054e4`), so this is not something the PR introduced or can avoid by rebasing.

Fixed in `caed876a` by qualifying the call, the way the same file already reaches `crate::estimation::parameterization::pack_params` elsewhere. Merged current `main` in first, so this branch now carries the repair and is green locally: preflight (all five check feature sets, clippy, rustdoc, docs, public-API), `--lib` 5055 passed, the VI tests that could not compile now run (121 passed under `ci,nn`), and this PR's own tests pass under both `ci` and `ci,survival`.

Worth noting for its own sake, since it is the CLAUDE.md #1133 shape exactly: the test is `nn`-gated, so a local `cargo test` and the per-PR core job both stay green while four other jobs cannot build. `tools/preflight.sh` does catch it — its `check` group covers `ci,nn,slow-tests` — which suggests it was not run before that merge.

One job is worth a second look once this run finishes: `codecov/patch` also reported a failure, but it had no successful build to measure, so I expect it to resolve on its own. I will confirm rather than assume.

roninsightrx:
Thanks — the finding is correct, and I verified it by probe rather than by reading: an endpoint-only model at `cmt = 1` plus an observation explicitly on `CMT=2` reports exactly

```
E_PER_CMT_ERROR_MODEL :: [error_model] has no entry for observed compartment(s) 2;
                         add a `CMT=N: DV ~ ...` line for each observed CMT.
```

and nothing else. The row lands in the Gaussian vectors and the empty per-CMT error map stops it there; `E_ENDPOINT_UNROUTED` is the other condition entirely — Gaussian observations on a *declared endpoint* CMT because the population was read without endpoint routing (#1199).

Fixed in `c2d51b23` at both sites you named (`docs/data-format.qmd`, `src/api/run.rs`). `src/api/run.rs:1095` keeps `E_ENDPOINT_UNROUTED` — that one is the #1199 doc and is correct in its own context.

**I also made the claim testable rather than just corrected.** The suppression is only safe *because* getting it wrong is loud, and that premise was sitting in a comment next to a test that never exercised the refusal — so the wrong code name could not have been caught by anything. `a_lone_endpoint_at_the_default_compartment_is_not_reported` now asserts the code:

```rust
assert_eq!(
    codes,
    vec!["E_PER_CMT_ERROR_MODEL"],
    "an observation outside the lone endpoint must be refused, and by this code — \
     the suppression above is only safe because getting it wrong is loud"
);
```

Same class of thing as the #1423 review's "a green test is not evidence that it can fail", one level up: a *premise* stated in prose is not evidence either.

**On the other two points:**

- The inherited compile failure is fixed in `caed876a` (merged current `main`, then qualified the `compute_bounds` call). Check, Clippy, Rustdoc, Format, Docs lint and Public API baseline are green on this branch now; the three test jobs are still running.
- Patch coverage: agreed that 50% was measured without the endpoint upload. The uncovered lines it listed — 6 in `src/api/run.rs`, 4 in `src/io/datareader.rs` — are `model_routes_rows_by_cmt`'s new gate and `ObsRouting::routes_only`, both reachable only from `survival`-gated tests, so they are exactly what the endpoints job measures and the core job cannot. I am waiting for the merged result rather than asserting it will clear; if the 90% gate is genuinely short I will add the tests to close it and say so.

_(Re-posting: the previous copy of this comment lost two code spans to shell expansion on my side.)_

Issue #1359, opened 2026-09-11
bug(types): has_lagtime / has_bioavailability are true for an unbound analytical LAGTIME / F, and f=X, lagtime=X makes pk_indices order-dependent


TeunP:
Two defects in how `CompiledModel` decides whether a model has a lag time or a bioavailability, both found in the review of #1355 (https://github.com/FeRx-NLME/ferx-core/pull/1355#issuecomment-5640741040). #1355 documents them; it does not change the logic.

## 1. `has_lagtime` / `has_bioavailability` return true for an unbound analytical `LAGTIME` / `F`

Both functions first check `pk_indices` for the slot (`PK_IDX_LAGTIME` / `PK_IDX_F`), then scan `indiv_param_names` for the bare name (`src/types.rs`, `has_lagtime` and `has_bioavailability`).

**Where the scan is redundant.** On the ODE and compartment-free layout, a bare `LAGTIME`/`ALAG`/`F` is a canonical name, so `ode_param_slots` already puts its slot into `pk_indices`.

**Where it is wrong.** On the analytical engine, a `LAGTIME` or `F` declared in `[individual_parameters]` but not bound with `lagtime=` / `f=` on the `[structural_model]` line:
- is never written to its slot;
- makes the parser warn that it is "computed but never used";
- still makes the function return `true`.

**Measured** in the review: on `CL2, CL, V, KA, LAGTIME` with `pk one_cpt_oral(cl=CL, v=V, ka=KA)`, `has_lagtime()` is `true`, slot 8 stays `0`, and the parser warns that `LAGTIME` is never used.

**Effect.** The numbers are correct, but the model is taken off the lag-free fast paths:
- the explicit sensitivity kernels (`src/sens/provider.rs:6336`);
- the cached event schedule (`src/estimation/inner_optimizer.rs:664`; `has_bioavailability` gates it together with a rate-defined infusion).

The review also flagged, by code reading, that the SS gate for transit/IG closed forms (`src/api/validation.rs:2462`) could reject the model. That gate calls `has_lagtime_on_cmt`, not `has_lagtime`, so check that function for the same bare-name shape.

## 2. One variable bound to two PK roles makes `pk_indices` depend on `HashMap` order

[...]

PR #1355, opened 2026-09-11
docs(ode): correct stale sequential-slot comments on ODE pk_indices

TeunP:
## Why

Since d54bb953 (`fix(ode): apply bioavailability F at dose entry`), an ODE model's `CompiledModel.pk_indices` is `ode_slot_map`, built by `ode_param_slots()` in `src/parser/model_parser.rs`. Canonical names (`CL`, `V`, `KA`, `F`, `LAGTIME`, …) take their fixed PK slot via `PkParams::name_to_index`. Every other name takes the lowest free slot outside `RESERVED_PK_SLOTS` (F / LAGTIME). The same layout later became the compartment-free one (#811). Several comments were written before d54bb953 and still describe the old positional layout ("ODE models write sequentially into slot `i`").

That stale doc matches a real downstream bug. ferx-r's `build_individual_estimates` (`src/rust/src/lib.rs:3040`–`3045` on ferx-r `main` `9d48df0`) reads slot `i` for ODE models. With KA at slot 4 and LAGTIME at slot 8, a column can show an unwritten 0, or another parameter's value: an independent review measured V's column showing KE's value and KA's column showing V's.

## What changed

Comments, docs, one test-fixture line and one new unit test; no production code or public API changed. Each claim below was checked against `model_parser.rs`:

| Site | Was | Now | Evidence |
|---|---|---|---|
| `src/types.rs`, `indiv_param_names` doc | "For ODE models the i-th name is written sequentially into slot `i`" | Always read `PkParams.values[pk_indices[i]]`. ODE/compartment-free slots come from `ode_param_slots` (example: `V, KA, KE` → `1, 4, 0`). Analytical names not bound on `[structural_model]` or readout-referenced get a placeholder `0` (#1356), including a modeled-dose `D{n}`/`R{n}`, which *is* written to a spare slot ≥ 9 but cannot be read back through `pk_indices` | `pk_indices = ode_slot_map.clone()` when `pk_param_map` is empty (`model_parser.rs:2803/2830`); analytical `unwrap_or(0)` at `:2823`; `D{n}`/`R{n}` spare slots at `:2316`–`:2418` |
| same doc, "Bound" paragraph | Names past `MAX_PK_PARAMS - 1` are silently not stored | Rejected at parse time by `ode_param_slots` | `Err("…too many individual parameters…")` at `:10382`. Nothing is added to `indiv_var_names` after `:2082` on this layout |
[...]

PR #1441, opened 2026-09-18
fix(parser): unbound analytical LAGTIME/F is not a lag/F; reject one variable under two PK roles [closes #1359]

hiddevandebeek:
## Why

#1359, Tier 1 row 31 of the 0.4.0 rollout (#1350). Two defects in how `CompiledModel` decides whether a model has a lag time / a bioavailability, both found in #1355's review:

1. **`has_lagtime()` / `has_bioavailability()` were true for an unbound analytical `LAGTIME` / `F`.** After checking `pk_indices`, both scanned `indiv_param_names` for the bare name. On the ODE and compartment-free layouts that scan is redundant (`ode_param_slots` already puts the canonical name's slot into `pk_indices`); on the analytical engine it is a false positive: a `LAGTIME` declared in `[individual_parameters]` without `lagtime=` on the `pk(...)` line never reaches slot 8 — the parser warns it is "computed but never used" — yet the model was taken off the lag-free fast paths (the explicit sensitivity kernels in `sens/provider.rs`, the cached event schedule in `inner_optimizer.rs`). Correct numbers, slower fit.
2. **`pk(..., f=X, lagtime=X)` made `pk_indices` depend on `HashMap` order.** `pk_indices` is built by reversing `pk_param_map` into a map keyed by *variable*, so with one variable under two roles only one role survives, chosen by the per-process hash seed. The issue measured `[0, 1, 4, 5]` / `has_lagtime() == false` or `[0, 1, 4, 8]` / `has_lagtime() == true` over 64 parses of one file, while `pk_param_fn` wrote both slots every time — so the lag-free fast paths could run with a lag applied, differently from run to run.

## What changed

- **§2 first, as the issue recommends:** `build_pk_param_fn` rejects one variable bound to two *different* slots at parse time, keyed on the resolved var slot (so the lowercase compat spelling `f=X, lagtime=x` — mapped `X` resolving to a declared `x` — collides the way it binds). Placed after the #1048 same-slot collapse, so `lagtime=X, alag=X` (two spellings of one role) stays legal. `pk_slot_noun` now covers every slot `name_to_index` hands out, since this diagnostic can name any pair.
- **§1:** the bare-name arm is dropped from `has_lagtime`, `has_lagtime_on_cmt` and `has_bioavailability`. All three now answer from `pk_indices` plus one shared private helper `has_indexed_dose_attr(attr, cmt)` for the ODE-only indexed `ALAGn` / `Fn` arm (route 2), so the two lag predicates cannot drift from each other again.
- `analytical_dose_attr_slot_map`'s sort stays (determinism is cheap) but its comment no longer claims the double-role binding reaches it. The absorption twin's #993 companion decline for a dose-attribute parameter reused in a disposition role (`v=F, f=F`) is retired — it is the same binding, now rejected upstream (see *Tests*).
- Docs: `structural-model.qmd` and `lagtime.qmd` document the new rejection and that an unbound analytical `LAGTIME`/`F` is inert on every path. `CHANGELOG.md` → `Fixed`, flagged as a widening reject.

**Second commit (`0f24373c`) — the four findings of the high-effort review, all incorporated** (replies on each inline comment):

- The two-roles diagnostic quotes both mappings as the file spells them (`var_seen` keeps the earlier value).
- `pk_indices` is derived from the `(pk_slot, var_slot)` pairs `build_pk_param_fn` resolved, not from a second uppercased name-keyed reversal of `pk_param_map` — so two declared names differing only in case (`X`, `x` under `f=X, lagtime=x`) keep their own slots instead of collapsing and flipping. The modeled-dose `D{cmt}`/`R{cmt}` pairs are deliberately skipped so `analytical_supported`'s FD routing is unchanged.
- The reject is documented as covering **every** pair of roles (`q=Q, q3=Q` too), with the reason: `pk_indices` holds one slot per variable, and the closed-form η-derivative fallback reads it.
- **A literal lag (`lagtime=0.75` / `alag=0.5`) was a real wrong-gradient defect on `main`**, found by measuring the reviewer's "known gap": `has_lagtime()` was false, `analytical_supported` true, and the provider ran the lag-free walk against a production predictor that applied the lag — 120 % apart at the first oral sample, 11 % on IV bolus. The parser now records literal-bound slots on `IndivParamPartials::const_pk_slots` (no public-struct change) and `has_lagtime` / `has_lagtime_on_cmt` read them as route 3; `provider_literal_lagtime_matches_production` passes `check_full_provider_vs_fd` on both fixtures. A literal `f=0.5` is measured equal on both sides and `has_bioavailability()` stays false on purpose (`provider_literal_bioavailability_matches_production` pins the asymmetry and the reason).

Not changed, documented on `has_lagtime`: a `lagtime=` binding to a name assigned only inside an `if` without `else` resolves in `build_pk_param_fn` (which sees every assigned name) but is not in `indiv_param_names`, so route 1 misses it. That gap existed before and the removed scan did not cover it either (the scan matched the literal names `LAGTIME`/`ALAG`, not the bound one); it is out of #1359's scope.

## Alternatives considered

- Sorting `var_to_pk` and keeping a per-variable *set* of roles instead of rejecting — but `pk_indices` is one slot per name and cannot represent two roles, and `f=X, lagtime=X` has no sensible meaning. Rejection is the same shape as #1048.
- Keeping the bare-name scan but qualifying it with "is bound in `pk_param_map`" — `CompiledModel` does not retain `pk_param_map`, and adding a field is a public-struct change for no gain over route 1, which is exact for every declared name.

## Cross-repo dependency

No public item added or removed (`has_indexed_dose_attr` is private); the public-API baseline is unchanged. ferx-r picks up the behaviour change at its next lock bump.

## Breaking changes
- [x] `.ferx` model file format (add migration note below) — widening reject only
- [ ] Public Rust API
- [ ] None

<details>

A `pk(...)` line binding one variable to two different roles is now a parse error — any two roles, so `f=X, lagtime=X` and also a shared-parameter constraint such as `q=Q, q3=Q` or `v2=VP, v3=VP`. Give each role its own `[individual_parameters]` variable (`Q3 = Q`, then `q3=Q3`). Two spellings of one role bound to one variable (`lagtime=X, alag=X`) still parse.

A model with a literal lag (`lagtime=0.5`) now routes through the lag-aware sensitivity paths; its OFV is unchanged and its gradients are now right, so a fit may converge to a (slightly) different point than before.

</details>

## Numerical validation
- [x] §1/§2: no number changes — §1 only changes *which path* computes a value that was never applied; §2 turns a nondeterministic parse into an error.
- [x] Literal-lag fix (second commit): `Dual2`-vs-FD parity via `check_full_provider_vs_fd` on a 1-cpt oral (`lagtime=0.75`) and a 1-cpt IV bolus (`alag=0.5`) fixture — value to 1e-9, ∂f/∂η, ∂²f/∂η², ∂f/∂θ, ∂²f/∂η∂θ to the harness's tolerances. Before the fix the value check alone was off by 120 % / 11 %.

## Performance
- [x] No significant impact expected — an analytical model with an unbound `LAGTIME`/`F` regains the lag-free fast paths, which is the point.

## Tests

| Test | Pins | Mutation that reddens it |
|---|---|---|
| `analytical_unbound_lagtime_and_f_are_not_a_lag_or_bioavailability` (parser) | unbound analytical `LAGTIME`/`F` → all three predicates false; bound control → true; never-used warning still emitted | restoring the bare-name arm in any of the three |
| `has_lagtime_answers_from_routed_slots_and_indexed_names_only` (types) | the shared helper's three arms: slot, ODE-only indexed name, per-compartment scoping (`ALAG2` is not a lag on cmt 1) | dropping the `ode_spec` gate or the `cmt` filter |
| `one_variable_bound_to_two_pk_roles_is_rejected_deterministically` | `f=X, lagtime=X` is a parse error naming both mappings and both roles, identical over 16 parses, and fires *before* the #993 read check | deleting the `var_seen` check |
| `one_variable_bound_to_two_pk_roles_collides_the_way_it_binds` | `f=X, lagtime=x` (compat lookup) collides; a non-dose-attribute pair (`f=KA`) is named too | keying `var_seen` on the spelling instead of the var slot; the old three-slot `pk_slot_noun` |
| `one_variable_under_two_spellings_of_one_role_stays_legal` | `lagtime=X, alag=X` parses, one `PK_IDX_LAGTIME` in `pk_indices` | moving the `var_seen` check ahead of the `slot_seen` collapse |
| `has_bioavailability_detects_f_on_either_engine` (existing, updated) | analytical bare `f` → false | restoring the bare-name arm |
| `case_distinct_names_under_two_roles_keep_their_own_slots` | `X` and `x` both declared, `f=X, lagtime=x`: `X → PK_IDX_F`, `x → PK_IDX_LAGTIME` over 16 parses | looking the pair up by uppercased name (the old reversal) |
| `provider_literal_lagtime_matches_production` (sens) | literal `lagtime=0.75` / `alag=0.5`: `has_lagtime()` true with no variable behind the slot; full FD parity | dropping `const_pk_slots` from `routes_lag_slot` → 120 % value gap |
| `provider_literal_bioavailability_matches_production` (sens) | literal `f=0.5` applied on both sides (pinned as 0.5 × the unscaled model), `has_bioavailability()` false | — (documents a deliberate asymmetry) |
| `dose_attr_param_reused_as_a_disposition_role_is_rejected_before_the_twin` (rewritten) | `v=F, f=F` is the parse-time reject; `v=V` control still builds its twin | deleting the `var_seen` check |

The existing ODE routing tests (`test_lagtime_in_ode_model_routes_to_canonical_slot`, the `ALAG2` tests in `tests_derived_iov_kappa.rs`, `ode_provider_tests.rs`) are unchanged and green — bare `LAGTIME` on the ODE layout is still true through `pk_indices`.

**Mutations run** (each applied to the committed tree, narrow filter, then reverted; logs at `/scratch/1359/mut-M*.log` on the dev box):

| Mutation | Tests reddened |
|---|---|
| M1 — restore the bare-name arm in all three predicates | `analytical_unbound_lagtime_and_f_are_not_a_lag_or_bioavailability`, `has_bioavailability_detects_f_on_either_engine`, `has_lagtime_answers_from_routed_slots_and_indexed_names_only` (3 / 7) |
| M2 — delete the `var_seen` rejection | `one_variable_bound_to_two_pk_roles_is_rejected_deterministically`, `…_collides_the_way_it_binds`, `dose_attr_param_reused_as_a_disposition_role_is_rejected_before_the_twin` (3 / 7) |
| M3 — key `var_seen` on the mapped spelling instead of the resolved var slot | `one_variable_bound_to_two_pk_roles_collides_the_way_it_binds` (1 / 7) — the compat-lookup fixture is the only one that can see this, which is why it exists |
| M4 — drop the `ode_spec` gate and the `cmt` filter from `has_indexed_dose_attr` | `has_lagtime_answers_from_routed_slots_and_indexed_names_only`, `has_bioavailability_detects_f_on_either_engine` (2 / 7) |
| M5 — `routes_lag_slot` ignores `const_pk_slots` (second commit) | `provider_literal_lagtime_matches_production` (1 / 8) |
| M6 — `pk_indices` looks the pair up by uppercased name (second commit) | `case_distinct_names_under_two_roles_keep_their_own_slots` (1 / 8) |
| M7 — diagnostic prints the earlier mapping with the current spelling (second commit) | `one_variable_bound_to_two_pk_roles_collides_the_way_it_binds` (1 / 8) |

Third commit (`6465583c`): `pk_slot_noun_names_every_slot_name_to_index_hands_out` — codecov/patch read 87.7 % with exactly the eight noun-table arms no fixture reaches as the misses; the test walks every role key `name_to_index` accepts.

Full `cargo test --lib` (no features, the set this box can build): **4848 passed, 2 failed** on the first run — both failures were the two findings below, fixed and re-run green (1093 tests in the parser / types / sens-provider / iov-kappa / modified-release modules). `tools/preflight.sh` green: fmt, all 5 check feature sets, clippy `--all-targets`, rustdoc, docs-lint, public-API baseline unchanged.

**Two things the suite found that the plan did not have:**

- `dose_attr_param_reused_as_a_disposition_role_declines_the_twin` used `pk one_cpt_transit(cl=CL, v=F, n=N, mtt=MTT, f=F)` — one variable under two roles — to pin a #993 companion guard in `absorption_ode_equivalent_source` that declined the twin for exactly that binding. The new parse-time rejection fires first, so that guard became a second gate over the same input and (per the "two redundant gates cover for each other" rule) is **retired**; the test is rewritten to pin the parse-time rejection plus a `v=V` control that still builds its twin. The census comment at the twin-build site is updated.
- My first compat-lookup fixture was wrong: `build_pk_param_fn` looks up the mapped name exactly, then *its lowercase form* — so the compat case is `f=X, lagtime=x` with `x` declared, not `f=x` against a declared `X` (that one is an undefined-reference error, as the suite said). Fixed; M3 confirms the corrected fixture sees the slot-vs-spelling distinction.

- [x] Not applicable — the new error text is in `docs/model-file/lagtime.qmd`.

## Docs
- [x] `docs/` updated for user-visible changes (`model-file/structural-model.qmd`, `model-file/lagtime.qmd`; `docs-lint --check` green)

## Checklist
- [x] `Dual2` path untouched (only routing predicates and the parser)
- [ ] `Cargo.toml` version bumped — not needed (0.4.0 unreleased)

## Docs & examples
- [x] No example execution step needed — no new syntax; the widened reject has no example file to run.

## Reviewer hints

- `build_pk_param_fn`'s new check is ~15 lines after the #1048 `slot_seen` block; the ordering (`slot_seen` collapse **before** `var_seen`) is what keeps `lagtime=X, alag=X` legal — `one_variable_under_two_spellings_of_one_role_stays_legal` pins it.
- The three predicates in `types.rs` are now one-liners over `has_indexed_dose_attr`; the doc comment on `has_lagtime` is where the route-1/route-2 story and the remaining `if`-without-`else` gap live.
- The literal-lag fix is the only piece that touches a number: `routes_lag_slot` (types.rs) + the `const_pk_slots` capture in `build_pk_param_fn`. Everything the predicate feeds already handles a zero-derivative lag, which is why the fix is routing rather than a new kernel; the FD harness is the evidence.
- `pk_indices` derivation: the `!modeled.contains(&slot)` filter is what keeps `analytical_supported`'s routing byte-identical for `D{cmt}`/`R{cmt}` models — remove it and those models re-route.
- I could not run `--features ci` locally (box RAM); CI is the authority for the `nn`/`survival`-gated fixtures.

## Open questions

None.


src/parser/model_parser.rs
```diff
@@ -17953,6 +17970,25 @@ fn build_pk_param_fn(
             continue;
         }
         slot_seen.insert(pk_slot, (pk_name.as_str(), var_name.as_str(), binding));
+        if let PkSlotBinding::Var(var_slot) = binding {
+            if let Some(&(prev_key, prev_slot)) = var_seen.get(&var_slot) {
```
hiddevandebeek: **[low] The diagnostic quotes the *earlier* mapping with the *current* spelling.** `var_seen` stores only `(pk_name, pk_slot)`, and the message prints `{prev_key}={var_name}` using the current entry's `var_name`. In the compat case this PR explicitly designed for — `f=X, lagtime=x` with `x` declared — the error reads "`f=x` and `lagtime=x` bind one parameter …", but the file says `f=X`. That is the exact trap `analytical_dose_attr_remedy_quotes_the_mapping_as_written` exists to prevent ("naming an argument the model file does not contain sends the reader looking for the wrong text"). Store the previous `var_name` in `var_seen` (`HashMap<usize, (&str, &str, usize)>`) and print it for the first half.

src/parser/model_parser.rs
```diff
@@ -17866,6 +17873,16 @@ fn build_pk_param_fn(
...
+    // #1359: the converse of #1048 — one *variable* under two role keys that
+    // land on two different slots (`f=X, lagtime=X`). `pk_param_fn` writes both
+    // slots, but `pk_indices` holds one slot per declared name, built by
+    // reversing `pk_param_map` into a `HashMap` keyed by variable — so only one
+    // role survives there, chosen by the per-process hash seed, and
+    // `has_lagtime()` / `has_bioavailability()` flip between parses of the
+    // identical file (measured: `[0, 1, 4, 5]` or `[0, 1, 4, 8]` over 64 parses)
+    // while the lag is applied every time. Keyed by *var slot*, not name, so the
+    // lowercase compat lookup (`f=x, lagtime=X`) collides the way it binds.
+    let mut var_seen: HashMap<usize, (&str, usize)> = HashMap::new();
```
hiddevandebeek: **[low] The rejection is wider than the defect and than the docs/CHANGELOG describe.** `var_seen` fires for *any* two roles, not only the F/lag pair, so a shared-parameter constraint such as `pk three_cpt_iv(cl=CL, v1=V1, q=Q, v2=VP, q3=Q, v3=VP)` or `q=Q, q3=Q` — which parsed on `main` and fit *correctly*, since `pk_param_fn` writes the same value into both slots and nothing reads a non-reserved slot in a way where "which slot `pk_indices` recorded" is observable — is now a parse error. The only slots any predicate inspects are `PK_IDX_F` / `PK_IDX_LAGTIME`. Either narrow the gate to pairs where at least one slot is in `RESERVED_PK_SLOTS` (that is the whole #1359 §2 hazard), or say in the CHANGELOG migration note and `structural-model.qmd` that *every* two-role binding is now rejected, since the current text lists only `f=X, lagtime=X`.

src/parser/model_parser.rs
```diff
@@ -17866,6 +17873,16 @@ fn build_pk_param_fn(
     // that agree (`lagtime=X, alag=X`) are merely redundant and stay legal, so
     // the `analytical_role_binding` tie-break still has a case to tie-break.
     let mut slot_seen: HashMap<usize, (&str, &str, PkSlotBinding)> = HashMap::new();
+    // #1359: the converse of #1048 — one *variable* under two role keys that
+    // land on two different slots (`f=X, lagtime=X`). `pk_param_fn` writes both
+    // slots, but `pk_indices` holds one slot per declared name, built by
+    // reversing `pk_param_map` into a `HashMap` keyed by variable — so only one
```
hiddevandebeek, line 17888: **[low] Residual hash-order flip the new gate cannot see.** `var_to_pk` is keyed by `var_name.to_uppercase()`, and individual-parameter names are case-sensitive (there is no case-insensitive duplicate check on declarations). With both `X` and `x` declared and `pk(..., f=X, lagtime=x)`, `var_seen` sees two different var slots and passes, but both entries collapse onto the single `"X"` key in the `var_to_pk` reverse map (`model_parser.rs:3625`, outside this diff), and both `pk_indices[X]` and `pk_indices[x]` take whichever role the `HashMap` insertion kept — the same `has_lagtime()` flip #1359 measured. Building `pk_indices` from the resolved `pk_assignment_mapping` (var slot → pk slot, which `build_pk_param_fn` already has) instead of a name-keyed reverse map would remove the last order-dependent step rather than guarding one spelling of it.

src/types.rs
```diff
@@ -5018,49 +5018,51 @@ impl CompiledModel {
...
-    ///      `alag=` is covered by route 1.) The scan also matches a bare `LAGTIME`/`ALAG`
-    ///      on every engine: redundant with route 1 on the ODE and
-    ///      compartment-free layout, and on the analytical engine a false
-    ///      positive for a `LAGTIME` declared without a `[structural_model]`
-    ///      binding — that value never reaches `PK_IDX_LAGTIME` and no lag is
-    ///      applied, yet the model is taken off the lag-free fast paths (the
-    ///      explicit sensitivity kernels and the cached event schedule).
+    ///      `alag=` is covered by route 1.)
+    ///
+    /// Known gap, unchanged here: a `lagtime=` binding to a name assigned only
```
hiddevandebeek, line 5057: **[low] The "known gap" list misses a literal binding.** `pk(..., lagtime=0.5)` is a legal `PkSlotBinding::Const` and is written to `PK_IDX_LAGTIME` by `pk_const_mapping`, but a constant has no variable, so nothing puts the slot into `pk_indices` and `has_lagtime()` is `false` while production (`pk/mod.rs` reads `pk_params.lagtime()` unconditionally) applies the 0.5 lag. Downstream, `dose_lagtime_values` / `dose_lag_duals` in `sens/provider.rs` and the `explicit_kind` selection then run the lag-free path — the "fast path with a lag applied" outcome §2 of the issue warns about. Pre-existing (the removed name scan never covered a literal either), but this comment now presents itself as the complete list of exceptions, so either name it here or fold const bindings into the predicate (e.g. record a `has_const_lag` on `CompiledModel` from `pk_const_mapping`).

src/parser/model_parser.rs
hiddevandebeek: Fixed in 0f24373c. `var_seen` now carries the earlier mapping's value as well, and the message quotes each half as the file spells it — under `f=X, lagtime=x` it reads "`f=X` and `lagtime=x`". `one_variable_bound_to_two_pk_roles_collides_the_way_it_binds` asserts both spellings are present and that `f=x` is not; mutation M7 (print the earlier mapping with the current spelling again) reddens exactly that assertion.

src/parser/model_parser.rs
hiddevandebeek: Kept the wide scope, documented it (0f24373c: CHANGELOG migration note, `structural-model.qmd`, `lagtime.qmd` now say *every* pair of roles, with `q=Q, q3=Q` / `v2=VP, v3=VP` as the examples and `Q3 = Q` as the remedy).

One correction to the premise, since it decides the scope: "which slot `pk_indices` recorded" **is** observable outside the two reserved slots. `resolve_param_derivs` / `subject_eta_grad` in `sens/provider.rs` fall back from the compiled-program path to `lognormal_param_derivatives` / `lognormal_eta_derivatives_only`, and both seed a variable's ∂/∂η on `model.pk_indices[i]` — one slot per variable. Under `q=Q, q3=Q` that fallback carries ∂/∂η_Q into whichever of `PK_IDX_Q` / `PK_IDX_Q3` hash order recorded and zero into the other. The program path (`prog.pk_slots()` from `pk_assignment_mapping`) covers both slots and is what an ordinary log-normal model takes, so I am **not** claiming a measured wrong number on `main` for that binding — only that the representation cannot hold it and the fallback reads the representation. That, plus the `.unwrap_or(0)` aliasing for anything the table cannot express, is why the rule is on the table's invariant (one slot per variable) rather than on the two slots that happened to bite first. The docs wording is careful to say "invisible to everything that reads the table", not "measured wrong".

src/parser/model_parser.rs
hiddevandebeek: Took the structural fix (0f24373c). `pk_indices` is now read off the `(pk_slot, var_slot)` pairs `build_pk_param_fn` resolved — the program's `pk_var_slots`, copied out before the program is moved into `ode_spec` — instead of a second, uppercased, name-keyed reversal of `pk_param_map`. One deliberate exclusion: the modeled-dose `D{cmt}`/`R{cmt}` pairs (#324) are skipped, because they were never in `pk_param_map`, the old vector always gave them `0`, and `analytical_supported` keys its FD routing on the slots it sees in this vector — including them would silently re-route every modeled-dose analytical model. That is a separate question and not this PR's.

`case_distinct_names_under_two_roles_keep_their_own_slots` pins the reviewer's fixture (`X` and `x` both declared, `f=X, lagtime=x`) over 16 parses: `X → PK_IDX_F`, `x → PK_IDX_LAGTIME`, both predicates true. Mutation M6 (look the pair up by uppercased name, reproducing the collapse) reddens it.

src/types.rs
hiddevandebeek: This one turned out to be a real wrong-gradient defect, not a comment gap, so I fixed it rather than naming it (0f24373c).

**Measured** before the fix, with `ONECPT_ORAL_LAG` rewritten to `lagtime=0.75` (no `LAGTIME` variable, no `ETA_LAG`): `has_lagtime() == false`, `analytical_supported == true`, and `subject_sensitivities` returned values from the lag-free walk while `compute_predictions_with_tv` applied the lag —

| t | provider `f` | production `f` |
|---|---|---|
| 1 | 9.2848 | 4.2088 |
| 2 | 10.4024 | 9.8387 |
| 4 | 10.1025 | 10.2827 |
| 8 | 9.1209 | 9.2982 |
| 24 | 6.0477 | 6.1653 |

and the IV bolus twin under `alag=0.5`: 1.4838 vs 1.6666 at t = 1.5 (11 %). So on `main`, any model with a literal lag gets FOCE/FOCEI gradients for a different model than the one whose OFV it reports. A literal `f=0.5`, by contrast, is equal on both sides to 1e-9 (both read the slot), and `has_bioavailability()`'s one consumer asks whether `F` can *move* across the inner search, so that predicate is left `false` on purpose and the asymmetry is pinned by `provider_literal_bioavailability_matches_production`.

Fix: `build_pk_param_fn` records the literal-bound slots on `IndivParamPartials::const_pk_slots` (the existing opaque metadata field, so no public-struct change), and `has_lagtime` / `has_lagtime_on_cmt` read it as route 3. A literal lag is a fixed break, but every consumer of the predicate handles a lag whose derivative is zero, so the lag-aware walk is correct and the cost is only the fast path. `provider_literal_lagtime_matches_production` runs `check_full_provider_vs_fd` on both fixtures — value to 1e-9 and every derivative block to FD — and passes; mutation M5 (drop `const_pk_slots` from `routes_lag_slot`) reproduces the 120 % gap on its first assertion.

The `if`-without-`else` gap in the doc comment is still the only one I know of and is still out of scope.

PR #1442, opened 2026-09-18
fix(parser): compartment-free models accept F{n}/ALAG{n}/D{n}/R{n} parameter names [closes #1358]

### LINKED ISSUE #1358: bug(parser): a compartment-free model rejects a parameter named F{n} / ALAG{n} with the analytical dose-route error
## Summary

A compartment-free (`$PRED`-equivalent, #811) model cannot declare an individual parameter named `F{n}` or `ALAG{n}`. Parsing fails with the **analytical** dose-route error, which is about a `pk(...)` model the user never wrote:

> `[individual_parameters]`: `F1` reads as a per-compartment bioavailability (`F1`), an ODE-only dose attribute, but nothing binds it to the analytical `one_cpt_iv` model's single dose route …

A compartment-free model has no doses, so no dose-attribute name means anything there. `F1` is an ordinary parameter name (e.g. a factor level, a fraction).

## Cause

The analytical modeled-dose loop in `parse_full_model` is gated on `if !is_ode` (`src/parser/model_parser.rs:2315`), not on `!uses_ode_param_layout`. A compartment-free model has `is_ode == false`, so the loop runs against its placeholder `pk_model` (`PkModel::OneCptIv`, `:2202`–`:2203`) and its empty `pk_param_map`:

- An `F{n}` / `ALAG{n}` name is never bound, so it hits the reject at `:2364`.
- By code reading, a `D{n}` / `R{n}` name also enters the loop: a compartment other than the placeholder's infusable one hits the infusable-compartment error at `:2394`, and `D1` / `R1` would be recorded as an analytical modeled dose. **Not measured.**

Elsewhere the same distinction is already keyed on `uses_ode_param_layout`, not `is_ode` (`allocate_readout_extra_slots`, `:17953`; the comment near `:2454`).

## Measured

In the review of #1355 (https://github.com/FeRx-NLME/ferx-core/pull/1355#issuecomment-5640741040), with a throwaway probe test:
- `F1 = TVS` in a compartment-free model fails with the analytical `one_cpt_iv` dose-route error.
- The same model with the parameter named `SLOPE` parses.

## Suggested fix

Gate the loop on `!uses_ode_param_layout` (or skip it when `is_algebraic`). Then audit the other `is_ode` checks in `parse_full_model` that guard analytical-only logic, for the same slip.

## Acceptance

- [ ] Tier-1 test: a compartment-free model with parameters named `F1`, `ALAG1` and `D2` parses, and each value is readable through `pk_indices`. It must fail on `main`.
- [ ] Analytical behaviour unchanged: the existing `F{n}`/`ALAG{n}` reject tests and the `D{n}`/`R{n}` modeled-dose tests still pass.
- [ ] `CHANGELOG.md` → `Fixed`.

Tracked in the 0.4.0 rollout epic #1350, Tier 1.

hiddevandebeek:
## Why
A compartment-free (`$PRED`-equivalent, #811) model could not declare an individual parameter named `F{n}` or `ALAG{n}` (a factor level, a fraction) — parsing failed with the **analytical** dose-route error, about a `pk(...)` model the user never wrote. Closes #1358 (Tier 1 of #1350, row 27).

The analytical modeled-dose loop in `parse_full_model` is gated on `!is_ode`, not `!uses_ode_param_layout`. A compartment-free model has `is_ode == false`, so the loop ran against its placeholder `pk_model` (`one_cpt_iv`) and empty `pk_param_map`.

**Measured on the pre-fix gate** (throwaway probe, one name per parse, all five arms — the issue had only `F1` measured and the `D{n}`/`R{n}` arms by code reading):

| name | pre-fix |
|---|---|
| `F1` | `Err`: "`F1` reads as a per-compartment bioavailability … nothing binds it to the analytical `one_cpt_iv` model's single dose route" |
| `ALAG1` | `Err`: same shape, "lag time" |
| `D2` | `Err`: "`D2` is a modeled infusion duration (RATE=-2) for compartment 2, but the analytical `one_cpt_iv` model can only infuse into compartment(s) [1]" |
| `D1` | `Ok`, but `dose_attr_map` **non-empty** — recorded at spare slot 9 while `pk_indices = [0, 1]` holds the value at slot 1 |
| `R1` | `Ok`, same stale `dose_attr_map` entry |

## What changed
- `src/parser/model_parser.rs`: the loop is gated on `!uses_ode_param_layout`, matching the two other compartment-free branches in the same function (`allocate_readout_extra_slots`, the #486 synthetic acceptance). One-line logic change plus a comment recording the five measured arms. Analytical `pk(...)` models keep every reject and every `D{n}`/`R{n}` routing (`is_ode == false` ⇒ `uses_ode_param_layout == false` for them).
- `src/parser/model_parser_tests.rs`: `algebraic_structural_model_accepts_dose_attribute_shaped_parameter_names` — for each of `F1`, `ALAG1`, `D1`, `D2`, `R1`: parses, no `parse_warnings`, `dose_attr_map.is_empty()`, `!has_lagtime()`, `!has_bioavailability()`, and the value is readable through `pk_indices` after `pk_param_fn`.
- `docs/model-file/structural-model.qmd`: one paragraph under *What a compartment-free model does not have* saying the compartment-indexed dose-attribute names are ordinary there, and that the bare `F`/`LAGTIME` still take their reserved slots.
- `CHANGELOG.md`: `Fixed` entry.

**`is_ode` audit** (the issue asks for it). The other `!is_ode` gates in `parse_full_model` that a compartment-free model enters:
- `:2862` ODE-only absorption fn in `[odes]` with analytical `pk` — a compartment-free model rejects `[odes]` by name earlier (`algebraic_structural_model_rejects_compartment_only_blocks`), unreachable.
- `:3531` `tv_fn` generation — engine-neutral (η = 0 evaluation of the block), correct for compartment-free too.
- `:5063` "computed but never used" census — `!pk_param_map.is_empty() || is_ode` already **skips** compartment-free models entirely. That means an unread `F1` (or any unread parameter) in a compartment-free model gets no dead-parameter warning; a pre-existing gap, not introduced or widened here, and out of this issue's scope. Noting it rather than fixing it because the census's remediation text ("Map it in `pk(...)`") would need a third wording.
- `parse_scaling_block :10463` Form C + `obs_scale` reject — already pinned for compartment-free by `algebraic_structural_model_rejects_obs_scale`.
- `:3448` `build_error_spec(.., is_ode)` (missed in the first draft; found in review) — rejects a non-empty per-CMT error model when `!is_ode` with "analytical PK models support a single error model only". For a compartment-free model the reject is right (no compartments); the wording names an engine the user never wrote. Pre-existing, tracked in #1443.
- **Pure-TTE models** (`struct_lines.is_empty()`) are neither ODE nor algebraic, so the loop still runs against their placeholder and an `F1` in a hazard is still rejected. Unchanged by this PR and outside #1358's scope; the same reasoning applies and it is tracked in #1443.
- `has_lagtime` / `has_lagtime_on_cmt` / `has_bioavailability` (`types.rs`) gate the indexed arm on `ode_spec.is_some()`, so a compartment-free `ALAG1`/`F1` never opens a lag/F route — asserted in the new test. The **bare** `F`/`LAGTIME`/`ALAG` are different: `ode_param_slots` routes them to the reserved slots on every ODE-layout model, compartment-free included, so `has_lagtime()` is true and `W_NEGATIVE_LAGTIME` stays reachable there. Numerically inert (no doses); the docs paragraph says so rather than claiming the bare names are ordinary (review H1).

## Alternatives considered
`if is_algebraic { skip }` inside the loop — same effect, but `uses_ode_param_layout` is the predicate the neighbouring branches already key on and the comment at `:3252` already explains why `is_ode` is the wrong key there.

## Cross-repo dependency

## Breaking changes
- [x] None (a model that previously failed to parse now parses; nothing that parsed changes)

## Numerical validation
- [x] Not applicable — parse-time gate; no closed form or ODE is evaluated for a compartment-free model, and analytical models take the identical path before and after.

## Performance

## Tests
- [x] Unit test added in the parser sibling test file (`cargo test --lib` passes: 4847 passed / 0 failed / 26 ignored, `--test-threads=1`)
- [x] Regression test fails without the fix — run against the pre-fix gate: panics at `F1` with the exact error from the issue

## Docs
- [x] `docs/` updated (`structural-model.qmd`, compartment-free section); `cargo run -p docs-lint -- --check` clean

## Checklist
- [x] `tools/preflight.sh` green — fmt, check (5 sets), clippy, rustdoc, docs, public-api
- [x] No `Dual2` path touched

## Docs & examples
- [x] No example execution step needed (parser fix; no new DSL syntax, no new example file)

## Reviewer hints
The logic change is one identifier at `model_parser.rs:3123`. Worth checking: (1) that no analytical `pk(...)` model can have `uses_ode_param_layout == true` (it is `is_ode || is_algebraic`, and `is_algebraic` requires the `[structural_model]` marker with no `pk`/`ode` line); (2) whether the census gap noted in the audit should become its own issue.

## Review
Independent high-effort review posted as a comment (approve-with-nits). Incorporated: H1 (docs narrowed to the indexed names; bare `F`/`LAGTIME` behaviour stated), H2 (audit line above), nit 4 (`D1` arm added — both measured stale-entry arms are now pinned), nit 5 (test doc comment says what `parse_warnings` can and cannot see). Nits 1–3 and the census question are pre-existing and filed as **#1443**.

## Open questions
None left for this PR; the census question moved to #1443.


hiddevandebeek:
## Independent review (high effort, read-only, separate worker)

**Verdict: approve-with-nits** — the one-identifier gate change is correct and cannot move any analytical `pk(...)` or pure-TTE model; the test is non-degenerate and each arm straddles the old gate. One user-facing docs sentence over-claims and should be narrowed before merge; the rest are follow-ups.

### Correctness of the gate (verified)

- `uses_ode_param_layout = is_ode || is_algebraic` (`src/parser/model_parser.rs:2786`). `is_algebraic` is `struct_lines.iter().any(|l| l == ALGEBRAIC_MARKER)`, and the marker is only written by `apply_algebraic_structural` when `classify_structural_block(...) == Algebraic`, i.e. no `pk`/`ode`/`ode_template` line. So for every analytical `pk(...)` model `is_algebraic == false`, `uses_ode_param_layout == is_ode == false`, and the loop runs exactly as before — every `F{n}`/`ALAG{n}` reject, the infusable reject and the `D{n}`/`R{n}` routing are untouched.
- Pure-TTE (`struct_lines.is_empty()`): `is_ode == false`, `is_algebraic == false`, so the loop still runs against the `OneCptIv` placeholder — **unchanged** behaviour, as the PR claims (see follow-up 2: the same "no doses ⇒ ordinary name" argument applies there too, and the PR is silent on it).
- Compartment-free: `pk_param_map` is empty, `build_pk_param_fn` takes the ODE branch (`is_analytical_pk = !pk_param_map.is_empty()`), so pre-fix the `analytical_extra_mapping` slot-9 write was dead while `dose_attr_map` still carried the `(Rate,1) -> 9` entry — exactly what the PR describes. Post-fix `dose_attr_map` stays `Default`.

### Test — non-degenerate

- Each arm straddles the old gate: `F1`/`ALAG1`/`D2` were `Err` pre-fix; `R1` was `Ok` with a stale map entry, and `dose_attr_map.is_empty()` is the only assertion that sees it — traced that the closure never writes slot 9 on the ODE branch, so the `pk_indices` readback would have passed pre-fix; the `is_empty()` assertion is load-bearing for that arm.
- `assert!(model.is_algebraic())` pins the fixture onto the branch under test (the CLAUDE.md "fast path" trap).
- The `pk_indices` readback is the issue's acceptance criterion and catches a distinct regression (a compartment-free model routed onto the analytical closure branch with an empty `pk_assignment_mapping` would read `0.0`, not `value`). Not tautological.
- `!has_lagtime()` / `!has_bioavailability()` pin the `ode_spec.is_some()` gate in `types.rs` and the fact that `ode_param_slots` does not route `alag1`/`f1` to a reserved slot.

### Hard findings

**H1. Docs paragraph over-claims for the bare names `F` / `LAGTIME`** — `docs/model-file/structural-model.qmd:211-215`.
The new paragraph says `F`, `LAGTIME` *and* the indexed names "are ordinary parameter names … neither bound to a dose route nor rejected". True for the indexed names (what this PR fixes and tests), **not** for the bare ones on the compartment-free layout: `ode_param_slots` routes a bare `F` to `PK_IDX_F` and `LAGTIME`/`ALAG` to `PK_IDX_LAGTIME` for every `uses_ode_param_layout` model, including compartment-free; `has_lagtime()` / `has_bioavailability()` therefore return `true` via `pk_indices`, and `check_model_data_warnings` has no `is_algebraic()` gate, so `W_NEGATIVE_LAGTIME` is reachable for a compartment-free `LAGTIME = TVL` with a negative typical value. Numerically inert (no doses), but the docs say "neither bound to a dose route", and the paragraph's own example — "a fraction" named `F` — is the case where the warning is reachable. Nothing in this PR tests the bare spellings; the changelog entry (correctly) claims only `F1`/`ALAG1`/`D2`/`R1`.
*Fix:* narrow the sentence to the compartment-indexed names, or say that `F`/`LAGTIME` still take the reserved slots. Do not add a bare-`F` claim without a test.

**H2. `is_ode` audit in the PR description is incomplete (benign omission).**
`build_error_spec(parsed_error_model, &sigma_names, is_ode)` is not listed. Inside it rejects a non-empty per-CMT error model when `!is_ode`, with the wording "analytical PK models support a single error model only". For a compartment-free model the *reject* is right (no compartments) but the wording names an engine the user did not write — same class as #1358, pre-existing, out of scope. Add the line to the audit; a follow-up issue for the wording is optional.

### Nits / follow-ups (no change required for merge)

1. **Newly reachable analytical-worded errors on two exotic compartment-free combinations.** Both only affect models that previously did not parse at all, so "nothing that parsed changes" holds:
   - `[adaptive_dosing] observe = … F1 …` on a compartment-free model: `[adaptive_dosing]` is not in the by-name reject list, and the observe check has no `is_algebraic` arm (unlike `[scaling]`), so `check_dose_attr_double_use` errors with "remove the `f=F1` mapping from the `pk(...)` call". Suggested follow-up: add `is_algebraic` to that split, or reject `[adaptive_dosing]` for compartment-free by name.
   - Dataset with a coded `RATE=-1` row into CMT 1 on a compartment-free model declaring `R1`: `check_modeled_dose_rates` now reports `E_MODELED_RATE_NO_PARAM` "requires an `R1` parameter … but none is declared" although the user declared it. Pre-fix it passed the check and then read an unwritten spare slot. A data error either way; the message names the wrong cause.
2. **Pure-TTE models keep the old reject.** With `struct_lines.is_empty()` the loop still runs on the placeholder `one_cpt_iv`, so a `[individual_parameters] F1 = …` used only in a hazard is still rejected with the analytical dose-route error. Out of #1358's scope, but the issue's reasoning applies identically.
3. **Census gap is genuinely pre-existing**: `!pk_param_map.is_empty() || is_ode` has skipped compartment-free models since #811 (the comment above it still says the skip exists for "Pure-TTE models"). Leaving it is the right call for this PR; file the follow-up, and note the census's header comment needs updating when it lands.
4. Test covers `R1` but not `D1`; both stale-entry arms were measured, only one is pinned. Adding `("D1", …)` to the array is free.
5. `parse_warnings.is_empty()` cannot observe a census warning for this model class (see 3); fine as a guard against any *other* warning, but the doc comment should not be read as pinning the census.

hiddevandebeek:
Incorporated in `ffc8244a`:

- **H1** — docs paragraph narrowed to the compartment-indexed names; it now says the bare `F`/`LAGTIME`/`ALAG` still take their reserved slots on the compartment-free layout and can still draw `W_NEGATIVE_LAGTIME` (confirmed: `validation.rs:5838` gates on `model.has_lagtime()` with no algebraic arm). No bare-name claim without a test.
- **H2** — `build_error_spec(.., is_ode)` added to the audit in the description, plus a line on pure-TTE models keeping the old reject (nit 2).
- **Nit 4** — `D1` arm added; both measured stale-`dose_attr_map` arms are pinned (13/13 compartment-free tests green on the rebuilt binary).
- **Nit 5** — test doc comment now says which assertion sees which arm and that `parse_warnings` cannot observe the census for this model class.
- **Nits 1, 2, 3 + census + `build_error_spec` wording** — pre-existing; filed as #1443 with acceptance criteria, not touched here.

Issue #1446, opened 2026-09-18
ci: main is red on `cargo check --tests --features ci,nn,slow-tests` — compute_bounds not imported in vi/run_tests.rs (from #1305)


roninsightrx:
## Summary

`origin/main` is red on one of the five `tools/preflight.sh check` feature sets:

```
cargo check --tests --no-default-features --features ci,nn,slow-tests
error[E0425]: cannot find function `compute_bounds` in this scope
  --> src/estimation/vi/run_tests.rs:1671
```

`src/estimation/vi/run.rs` imports `clamp_to_bounds, compute_mu_k, pack_params, pack_with_bounds, unpack_params, PackedStart` from `parameterization` but not `compute_bounds`, and the sibling test file calls it unqualified inside a `#[cfg(feature = "nn")]` test. The other four check commands are green.

Introduced by eaac92e4 (#1305). The `Check` CI job should be red on `main` for this set.

## Fix

One-line import (`use crate::estimation::parameterization::compute_bounds;` in the test sibling, or add it to `run.rs`'s import list). Already carried as commit `ec76430e fix(vi): import compute_bounds in the nn-gated VI test` on branch `worktree-saem-speed-e7-rm-scale`; happy to split it into its own PR.

Found while running the preflight matrix on merge branches during the SAEM speed programme (2026-09-18).

PR #1450, opened 2026-09-18
test(estimation): pin the check matrix's feature pairings, not just their union [closes #1446]

roninsightrx:
## Why

`closes #1446`.

**The instance is already gone, and that is worth stating first.** #1446 reported `origin/main`
red on one of the five `tools/preflight.sh check` feature sets. Reproduced on this worktree off
`65118bea`, that set is now **green**:

```
$ CARGO_BUILD_JOBS=2 cargo check --tests --no-default-features --features ci,nn,slow-tests
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 32.88s
```

The E0425 was closed incidentally by `65118bea` (#1440, an unrelated `api` PR), which reached
for an inline fully-qualified path at the call site:

```rust
clamp_to_bounds(
    &mut x0,
    &crate::estimation::parameterization::compute_bounds(&params),
);
```

That is what you write when you are unblocking a red build inside someone else's PR. It is not
how this file spells its `parameterization` helpers — `pack_params`, `clamp_to_bounds`,
`unpack_params` all arrive unqualified through `use super::*`. So the first half of this PR is
the import the issue actually asked for.

The second half is the part the issue is really about, and it is the answer to
**"why did CI not catch this at #1305's merge?"**

### It did. Four jobs did. The merge did not wait for them.

`.github/workflows/ci.yml`'s `Check` job runs `tools/preflight.sh check`, which runs
`ci,nn,slow-tests` as its fourth line. There is no hole in the matrix. The forensics on #1305's
final head (`14b2418a`, run `35369640612`):

| job | conclusion | completed |
|---|---|---|
| Format | success | 16:38:35Z |
| Docs lint | success | 16:38:27Z |
| **Tests (release semantics)** | **failure** | 16:39:04Z |
| **Clippy** | **failure** | 16:39:13Z |
| Public API baseline | success | 16:39:55Z |
| **Check** | **failure** | 16:40:38Z |
| **Tests + coverage (TTE/CTMM endpoints)** | **failure** | 16:42:05Z |
| Rustdoc | success | 16:38:50Z |

The PR merged at **16:38:42Z** — before `Check` finished at 16:40:38Z, and before three other
jobs that also saw it. All four are the jobs whose feature set includes `nn`. `main` then
carried the breakage across `80f054e4` and `afb4462d` (`CI completed/failure` on both, per
`gh run list --branch main`) until `65118bea` fixed it 3h37m later.

This repo has no required status checks, so `gh pr merge` does not block on CI. That is a repo
setting, not a file in this tree, and CLAUDE.md is explicit that a gate belongs in the script
rather than the workflow — so **this PR does not invent a fix for it**. Said plainly instead,
as the issue asked.

### What *is* broken is the guard on the matrix, and that is fixable here

`tests/preflight_owns_the_fast_gates.rs` exists precisely so a gate cannot be *neutered* while
staying in the list. Its feature assertion was a **flat union** over the whole `check` group —
deliberately, per its own doc comment, "so splitting or merging feature sets stays a free
refactor". But a union cannot see a **conjunction** being split up. Measured, on `main`:

```
# tools/preflight.sh:  --features ci,nn,slow-tests   ->   --features ci,nn
$ cargo test --test preflight_owns_the_fast_gates --no-default-features --features ci
test result: ok. 10 passed; 0 failed; ...
```

Every one of `ci`, `survival`, `slow-tests`, `markov`, `nn` is still somewhere in the union
(`slow-tests` comes from the *survival* line), the command count is still 5, and the entire
10-test guard file passes. What that edit silently stops compiling is the three
`#[cfg(feature = "slow-tests")]` items in `src/nn/mod.rs`'s test module
(`l2_shrinks_weights_and_modulator_variation` and its two helpers) — whose own doc comment names
`cargo check --tests --no-default-features --features ci,nn,slow-tests` as the thing that
catches their bit-rot. The survival line has the identical shape:
`tests/tte_convergence.rs` and `tests/categorical_convergence.rs` are
`#![cfg(all(feature = "survival", feature = "slow-tests"))]`.

This is CLAUDE.md's *"a green test is not evidence that it can fail"* — specifically the
"two redundant gates cover for each other" variant, where one condition's coverage is supplied
by a sibling line rather than by the condition under test.

## What changed

**1. `src/estimation/vi/run_tests.rs`** — the import #1446 asked for, replacing #1440's inline
path:

```rust
#[cfg(feature = "nn")]
use crate::estimation::parameterization::compute_bounds;
...
clamp_to_bounds(&mut x0, &compute_bounds(&params));
```

In the **test sibling**, not `run.rs`'s import list. `run.rs` imports its five
`parameterization` helpers because it uses them itself (5, 2, 6, 2, 3 references respectively);
`compute_bounds` has exactly one caller, the `nn`-gated test, so adding it there would be an
unused import in every build without `cfg(test)`. The `#[cfg(feature = "nn")]` on the `use`
matches its only caller's gate, so `--features ci` stays warning-free too.

**2. `tests/preflight_owns_the_fast_gates.rs`** — the feature assertion now parses **per
command** and splits into two **disjoint** halves:

* a **union** over the group for features a single flag opens: `ci`, `markov`;
* a **single-command conjunction** for features a pair opens: `nn`+`slow-tests`,
  `survival`+`slow-tests`, each with a failure message naming the bodies that would stop
  compiling.

Disjoint on purpose, and the doc comment says why: asserting both ways for the same feature
would be two gates rejecting the same edits, so deleting either would leave the suite green and
neither could be mutation-tested — the exact hole this change closes.

**3. `tools/preflight.sh`** — comment only, no command changed (`git diff` on this file is
entirely `#` lines). The `nn` line's rationale cited `tests/nn_fit_smoke.rs` and
`tests/nn_fit_convergence.rs` as "gated on BOTH `nn` and `slow-tests`". They are not, and have
not been since `7796919d` dropped their `#![cfg(feature = "slow-tests")]`; both are
`#![cfg(feature = "nn")]` today. The pairing is still load-bearing, but via `src/nn/mod.rs`, so
the comment now points there.

## Alternatives considered

* **Add `compute_bounds` to `run.rs`'s import list** (the issue's other suggestion). Rejected:
  it would be unused outside `cfg(test)` and trip `unused_imports` in ordinary builds.
* **Leave #1440's fully-qualified call alone.** Defensible — it compiles. Rejected as the
  smaller of two evils only: the issue asks for the import, and one inline 50-character path
  among a file of unqualified helpers is the kind of inconsistency that gets copied.
* **Add a new feature set to the check matrix.** Rejected — there is no gap. The set that
  catches this is already the fourth line, and adding a sixth would cost a compile to gate
  nothing new.
* **Add a workflow-level fix for "merged before CI finished".** Rejected — it is branch
  protection, not a file here, and CLAUDE.md says gates go in the script, not the workflow.
  Flagged in *Open questions* instead.
* **Keep the union assertion and add pairings on top.** Rejected explicitly: that is the
  "two redundant gates cover for each other" shape. The halves are disjoint so each is
  individually mutation-testable, which is demonstrated below.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` green, baseline untouched).

## Breaking changes
- [x] **None**

## Numerical validation

- [x] **Not applicable (docs / refactor / CI only)**

Per CLAUDE.md's *"Every new feature requires a comparison with NONMEM output"*: this PR produces
**no numerical result**. It changes one `use` statement in a test file, one assertion in a CI
guard test, and one shell comment. There is no estimate, OFV, residual or diagnostic to anchor,
so no documented exception is being invoked — the rule's precondition ("any behaviour that
produces numerical results") is simply not met. The `nn`-gated VI tests whose compilation this
restores are unchanged and still pass:

```
$ cargo test --lib --no-default-features --features ci,nn -- --test-threads=2 estimation::vi::run
test estimation::vi::run::tests::vi_objective_trace_is_the_penalized_objective ... ok
test estimation::vi::run::tests::vi_nn_l2_shrinks_the_network_weights ... ok
test result: ok. 43 passed; 0 failed; 0 ignored; 0 measured; 4965 filtered out; finished in 7.92s
```

## Performance
- [x] **No significant impact expected** — no command in `preflight.sh` changed, so the check
      matrix costs exactly what it did.

## Tests

- [x] **Regression test added that fails without this fix**
- [ ] No new tests needed

The strengthened assertion **is** the regression test, and the regression it exists to catch is
named in its failure message: *a `check` command losing one half of a feature conjunction while
staying in the list*. Mutation-tested in both directions, each mutation naming its own side:

| mutation in `tools/preflight.sh` | before this PR | after this PR |
|---|---|---|
| `ci,nn,slow-tests` → `ci,nn` | **green** (10/10 pass) | **red** — *"no single `check` command enables `nn` AND `slow-tests` together …"* |
| `ci,survival,slow-tests` → `ci,survival` | **green** | **red** — *"… enables `survival` AND `slow-tests` together …"* |
| `ci,markov` → `ci` | red | **red** — *"the `check` group compiles nothing under `markov`"* |

The third row is not decoration. It is the check that the **union half still bites** after being
narrowed to two features — i.e. that the two halves are genuinely disjoint rather than one
having become dead weight covered by the other.

And the `run_tests.rs` half, mutated by deleting the `use`:

```
$ cargo check --tests --no-default-features --features ci,nn,slow-tests
error[E0425]: cannot find function `compute_bounds` in this scope
    --> src/estimation/vi/run_tests.rs:1680:31
error: could not compile `ferx-core` (lib test) due to 1 previous error

$ cargo check --tests --no-default-features --features ci
    (green)
```

That is #1446's exact error reproduced on demand, plus the reason it was invisible to four of
the five sets.

**What is the *smallest* edit that removes the fix, and does a test die for it?** For the guard,
it is dropping one feature token from one line — covered by rows 1 and 2 above. For the import,
it is deleting the `use` — covered by the E0425 above. The one edit **no test here can catch**
is re-spelling the call fully-qualified again (#1440's form): that compiles, so nothing goes
red. It is a readability convention, not an invariant, and this PR does not pretend otherwise.

- [x] **Not applicable (internal / refactor / fix with no new API surface)**

## Docs
- [x] **`CHANGELOG.md` `[Unreleased]` entry — N/A (internal/CI)**. No user-facing behaviour
      changes and the gate matrix's *commands* are untouched; per CLAUDE.md a PR that only
      touches internal tests or CI needs no entry.
- [x] **No user-visible change.** `docs/development/*.qmd` was checked and describes neither the
      check matrix nor the guard test, so there is nothing there to keep in sync.

## Checklist
- [x] `tools/preflight.sh` green — **full run, not a subset**:
      `preflight OK — fmt check clippy rustdoc docs public-api`.
      All five check feature sets green, listed below.
- [x] Functions on the `Dual2` sensitivity path — not touched.

### The five `check` feature sets, as requested in the issue

All five were run on this branch. **All five green**, and four of the five were already green on
`main` before this PR (the `ci,nn,slow-tests` one having been fixed by #1440 three hours after
it broke):

```
$ tools/preflight.sh check
══ check ═══════════════════════════════════════════════════════════════════
  $ cargo check --tests --no-default-features --features ci
  $ cargo check --tests --no-default-features --features ci,survival,slow-tests
  $ cargo check --tests --no-default-features --features ci,markov
  $ cargo check --tests --no-default-features --features ci,nn,slow-tests
  $ cargo check -p ferx-tools -p ferx-cli --tests --no-default-features --features ferx-core/ci
preflight OK — check
```

No other set is red, so nothing further was fixed or filed on that account.

### ferx-site
- [x] No new data file

### ferx-book

### Example execution
- [x] **No example execution step needed** (internal / no user-visible change)

## Reviewer hints

* **Focus on `tests/preflight_owns_the_fast_gates.rs`.** The substance is the split into a
  union half and a conjunction half, and the claim that they are *disjoint*. If you think a
  feature should be in both halves, that is the thing to argue with — the doc comment makes the
  case that it should not.
* **`tools/preflight.sh` can be skimmed**: `git diff` on it is comment lines only, no command
  changed. Worth a glance only to confirm that.
* **The `#[cfg(feature = "nn")]` on the `use`** in `run_tests.rs` is load-bearing, not
  decoration — without it `--features ci` gets an unused import. `cargo check --tests
  --no-default-features --features ci` was run clean to confirm.
* The CI forensics table in *Why* is from `gh run view 35369640612 --json jobs` and
  `gh run list --branch main`, not from the issue text.

## Open questions

1. **The real hole is branch protection, and it is out of this tree.** #1305 merged 2 minutes
   before its `Check` job finished, and four red jobs on `main` then went unnoticed for 3h37m
   across two further merges. `gh pr merge --auto` on this repo does not wait, because no status
   checks are required. Nothing in `tools/preflight.sh` can fix that. Worth marking `Check`,
   `Clippy`, `Format`, `Rustdoc`, `Docs lint` and `Public API baseline` required on `main`? That
   is a repo-settings change and I have not made it.
2. `tests/nn_fit_smoke.rs` and `tests/nn_fit_convergence.rs` losing their `slow-tests` gate in
   `7796919d` means their Tier-3 bodies now compile under plain `ci,nn` — which is fine — but it
   is worth a maintainer's eye on whether that was intended, since the `#[cfg_attr(not(feature =
   "slow-tests"), ignore)]` pattern is what the tier rules actually ask for and a whole-file
   `#![cfg]` is the stronger form.


roninsightrx:

The feature-pairing fix is sound, but the new guard does not ensure those features compile test targets. I found one blocking test hole. Per instruction, I performed static review only and ran no builds, tests, or scripts.

### Blocking findings

- `tests/preflight_owns_the_fast_gates.rs:817` — medium — The assertion checks only that `nn`/`slow-tests` and `survival`/`slow-tests` occur on the same command. Deleting `--tests` from either corresponding `cargo check` line preserves the command count and feature sets, so this test remains green while `cfg(test)` modules and the cited integration tests stop compiling. — Retain each command alongside its parsed features and require the matching command to be `cargo check` with a test-target selector such as `--tests` or `--all-targets`.

### Nits

None.

roninsightrx:

Request changes. The intended conjunction check works for feature removal, but it does not verify that the command compiles test targets. No compilation or execution was performed, per the review constraint.

## Blocking findings

- [tests/preflight_owns_the_fast_gates.rs:817](/private/tmp/claude-501/-Users-ron-git-ferx-ferx-core/62ca5e99-b40b-402e-ac57-aee3e1d012e6/scratchpad/review-pr1450/tests/preflight_owns_the_fast_gates.rs:817) — high — The assertion checks only that both feature names occur in one command. Deleting `--tests` from either paired `cargo check` command leaves this test and the command-count guard green, while `src/nn`’s `#[cfg(test)]` module and the two survival integration tests stop compiling. A member-only command could likewise satisfy the pair without selecting ferx-core’s test targets. — Parse target selection as well as features, and require the pair on a command that includes `--tests` and selects ferx-core. Explicitly mutation-check removal of `--tests`.

## Nits

- [tests/preflight_owns_the_fast_gates.rs:816](/private/tmp/claude-501/-Users-ron-git-ferx-ferx-core/62ca5e99-b40b-402e-ac57-aee3e1d012e6/scratchpad/review-pr1450/tests/preflight_owns_the_fast_gates.rs:816) — low — Requiring the literal `survival` feature rejects the valid combination `markov,slow-tests`, even though `markov = ["survival"]` and therefore compiles the same double-gated tests. — Treat `markov` as enabling `survival`, or expand effective feature dependencies before checking conjunctions.

roninsightrx:
## Response to the Codex review (both runs treated as one)

Both findings **accepted and fixed** in `657e2627`. I mutation-tested each claim before agreeing rather than taking the static reading on trust — and both held. The blocking one held in a way worth naming plainly: it is the same neutered-gate shape this PR claimed to be closing, reproduced inside the closing change.

---

### Blocking — `tests/preflight_owns_the_fast_gates.rs:817`, "deleting `--tests` keeps the test green" — **accepted and fixed**

Correct, and worse than the wording suggests. I reproduced it exactly as described: deleting `--tests` from `cargo check --tests --no-default-features --features ci,nn,slow-tests` keeps all three feature tokens, keeps the command count at 5, and leaves **all ten tests in the file green**:

```
$ cargo test --test preflight_owns_the_fast_gates --no-default-features --features ci
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

I then checked the *consequence*, which the review asserted but could not run — and it is the sharpest possible demonstration, because the defect that is available to hand is #1446's own. With the `use` deleted from `run_tests.rs` and `--tests` deleted from that line:

```
$ cargo check --no-default-features --features ci,nn,slow-tests
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.28s
```

Clean. The very `error[E0425]: cannot find function 'compute_bounds'` this PR exists to gate compiles green through the gate, with the gate still sitting in the list. I also confirmed there was no other assertion covering the dimension: `--all-targets` is asserted for the *clippy* group, and `grep -n '\-\-tests' tests/preflight_owns_the_fast_gates.rs` matched nothing in the check group. So this was a genuine, uncovered hole, not a second copy of an existing one.

The fix follows the review's suggestion but splits it across **two assertions on two dimensions, each with exactly one owner**, rather than one combined predicate:

* **A blanket over every command in the group** — each must select test targets (`--tests` or `--all-targets`). Scoped to all five lines rather than only the two paired ones, because all five exist to compile-gate test code; the review's version would have left `ci` and `ci,markov` able to lose `--tests` silently. (Mutation row 2 below is that case.)
* **`selects_ferx_core`, applied only where a pairing lands** — your second sentence, the member-only `-p` case. The members line takes features package-qualified and the parse strips the prefix, so `-p ferx-tools --tests --features ferx-core/nn,ferx-core/slow-tests` would present the pair while compiling no ferx-core test target.

One deliberate departure from "require the pair on a command that includes `--tests` **and** selects ferx-core": `selects_ferx_core` says nothing about `--tests`. If the blanket already holds for every command, re-checking the selector inside the pairing is a second gate rejecting precisely the inputs the first one rejects — delete either and the suite stays green, so neither is mutation-testable. That is CLAUDE.md's "two redundant gates cover for each other", and it is the failure mode that produced this very finding, so I did not want to reintroduce it one layer down. The two halves are now independently demonstrable, which row 3 below is the evidence for: moving the pair onto the member-only line is red **although that line keeps `--tests`**, so the blanket provably cannot supply that rejection.

You explicitly asked for a `--tests`-removal mutation check. All four:

| mutation in `tools/preflight.sh` | before `657e2627` | after |
|---|---|---|
| `--tests` dropped from `ci,nn,slow-tests` | **green** (10/10) | **red** — *"a `check` command selects no test targets …"* |
| `--tests` dropped from `ci,markov` (no pairing on it) | **green** | **red**, naming `cargo check --no-default-features --features ci,markov` |
| pair moved to the member-only `-p ferx-tools` line (which keeps `--tests`) | **green** | **red** — *"… on ferx-core's own test targets … Nor is a member-only `-p ferx-tools` line …"* |
| `ci,survival,slow-tests` → `ci,markov,slow-tests` | **red** (the nit) | **green** |

---

### Nit — "requiring the literal `survival` rejects the valid `markov,slow-tests`" — **accepted and fixed**

Also correct, and not merely theoretical: it is a false positive on the exact property the original doc comment was protecting ("splitting or merging feature sets stays a free refactor"). `Cargo.toml` has `markov = ["survival"]`, so `--features ci,markov,slow-tests` turns on `feature = "survival"` and compiles `tests/tte_convergence.rs` and `tests/categorical_convergence.rs` just as the literal spelling does — my guard would have redded a maintainer merging those two check lines, for nothing.

Measured both directions rather than reasoned about, since the two differ by one call: with that mutation applied to `preflight.sh`, the assertion is **red** under the old raw-token compare and **green** after (row 4 above).

Fixed with `expand_features`, which transitively closes each parsed `--features` set over `Cargo.toml`'s `[features]` table before comparison. Read from the manifest rather than hard-coding `markov ⇒ survival`, so the implication graph keeps a single owner: adding `foo = ["survival"]` to the manifest is followed automatically with no second edit in the test. Bounded iteration with a panic on non-termination, matching the existing `profile_has_guards_on` helper's handling of `inherits` chains a few lines up.

---

### State

`tools/preflight.sh` is **byte-identical to the previous commit** — `git diff c2a53693 -- tools/preflight.sh` is empty; every mutation above was reverted after measuring. The whole diff of `657e2627` is one test file.

Full local gate green, not a subset:

```
$ tools/preflight.sh
preflight OK — fmt check clippy rustdoc docs public-api
```

plus `cargo test --test preflight_owns_the_fast_gates` — 10 passed, 0 failed.

Not merged, and auto-merge not enabled.

Issue #1444, opened 2026-09-18
estimation(saem): MH step-scale adaptation cannot reach its target — shipped warfarin_saem example runs at 2–4% acceptance against 40%, and the mixing warning stays silent


roninsightrx:
## Summary

SAEM's MH step-scale adaptation cannot reach its own target acceptance rate when the start is more than ~2× off, and the mixing warning does not fire. On the shipped example `examples/warfarin_saem.ferx` (with the default `n_mh_steps = 20`) the E-step accepts **2–4 % of proposals for the entire run** against the documented 40 % target, silently.

## Mechanism

`saem.rs` adapts the per-subject block scale `δ_i` and the componentwise scales every `adapt_interval` (default 50) iterations by a fixed ×1.1 / ×0.9. Over the default 150 + 250 = 400 iterations that is **8 corrections**, so `δ` can move by at most `0.9^8 ≈ 0.43` (or `1.1^8 ≈ 2.1`) across the whole fit. Warfarin starts at `step_scales = 0.3` and needs far more than that.

Measured combined block + componentwise acceptance (seed 1, every 25th iteration, unmodified `main` @ `afb4462d`):

```
B1 warfarin  1:0.268  26:0.016  51:0.021 101:0.026 201:0.024 301:0.029 376:0.045
B3 vanco     1:0.711  26:0.583  51:0.503 101:0.493 201:0.440 301:0.469 376:0.434
B5 pembro    1:0.6xx  ... 376:0.438   (never reaches 0.40)
```

The #895 mixing warning (`SAEM_MH_STUCK_ACCEPT = 0.01`) only fires below 1 % combined, so 2–4 % passes without a word. B3 and B5 are the milder version: they start above target and spend ~350 of 400 iterations drifting down.

## Effect on estimates

Same model, same seeds, per-iteration Robbins–Monro scale stepping (`log δ += c·k^-0.6·(accept_k − target)`) instead of the interval bump, 6 seeds, everything else unchanged:

| bench | rule | accept | d_final vs long-run ref | IS −2LL (mean ± sd) | cross-seed stability |
|---|---|---|---|---|---|
| B1 warfarin, 20 steps | interval (main) | 0.040 | 0.0028 | −286.04 ± 0.01 | 0.0023 |
| B1 warfarin, 20 steps | robbins_monro | 0.420 | **0.0011** | −286.05 ± 0.00 | **0.0012** |
[...]

PR #1451, opened 2026-09-19
feat(estimation): add a Robbins-Monro SAEM step-scale rule and a diagnostic that can see a chain far from target [closes #1444]

roninsightrx:
## Why

`examples/warfarin_saem.ferx` — a **shipped** example — runs its entire SAEM E-step at
2–5 % Metropolis-Hastings acceptance against a documented 40 % target, and says nothing
about it. Two independent defects (#1444):

1. **The step-scale adaptation cannot reach its target.** `saem.rs` adapted the
   per-subject block scale `δ_i` and the componentwise scales every `adapt_interval`
   (default 50) iterations by a fixed ×1.1 / ×0.9. Over the default 150 + 250 = 400
   iterations that is **8** corrections, so `δ` can move by at most `1.1^8 ≈ 2.1×` up or
   `0.9^8 ≈ 0.43×` down *however far from target the chain is*. The problem is reach, not
   schedule: a model whose optimal step is an order of magnitude below the 0.3 starting
   value never gets there.
2. **The mixing diagnostic cannot see it.** `SAEM_MH_STUCK_ACCEPT = 0.01` only fires below
   1 % cumulative acceptance, so a chain parked at 2–5 % for a whole run — which costs real
   estimate quality — passed in silence. The defect was not merely unfixed, it was
   unreported.

## What changed

**1. `[fit_options] scale_adaptation = interval | robbins_monro`** (new `FitOptions::saem_scale_adaptation`,
new `ScaleAdaptation` enum).

- `interval` — **the default**, and the pre-existing rule, unchanged.
- `robbins_monro` — a step taken every iteration from that iteration's own rate:
  `log δ += c·k^-0.6·(accept_k − target)`. The correction is proportional to the
  discrepancy, so the reach is unbounded, and the `k^-0.6` decay is the
  diminishing-adaptation condition that keeps an adapting chain valid (Roberts & Rosenthal
  2007). Applies to the primary block kernel and the per-η componentwise scales; the κ (IOV)
  scales stay on the interval rule under both settings (scope limit — the per-occasion
  kernel makes one proposal per occasion per iteration and the RM arm has not been measured
  there).

**The default stays `interval`, deliberately.** The issue's own data says `robbins_monro`
is not a free default yet: on pembrolizumab (303 subjects, 4 η, 20 steps), whose acceptance
was already *above* target, driving it to target regressed `d_final` at 3 seeds. Shipping it
as the default would silently move every existing SAEM user's estimates. It is opt-in, and
the SAEM docs page says both what it wins and where it loses.

Both rules are now single functions (`interval_scale_update`, `rm_scale_update`) that
`run_saem` calls and the tests measure, rather than two inline copies of the arithmetic.

**2. A second tier on the acceptance diagnostic.** A combined block + componentwise
acceptance outside **10–80 %** over the **last 100** post-burn-in iterations is appended to
`FitResult.warnings`, carrying the realised tail rate, the target, the window length and the
cumulative rate. Warnings go to `FitResult.warnings`, never stderr. Three deliberate choices:

- the **tail**, not the whole-run average — a run that mixes badly early and recovers
  averages to a number describing neither half;
- the band is **wide on purpose** — 0.25 or 0.60 is unremarkable; this is the range outside
  which the scales have *demonstrably* failed to find their target, not a range of "good"
  rates;
- it stays quiet below 25 post-burn-in iterations, where the realised rate still describes
  the starting scale rather than a failure to reach target.

The sub-1 % tier keeps its historical wording (`tests/frem_warfarin.rs` asserts the absence
of "not mixing"), and the new tier does not reuse that phrase.

**3. One incidental fix, found while building the tests.** With `optimizer_trace = true` the
trace path was `/tmp/ferx_trace_{pid}_{unix_seconds}.csv`. Two fits in the same process and
the same wall-clock second therefore shared a file — the second truncating the first — and
`FitResult::trace_path` then pointed at the wrong run's rows. A small model fits in well
under a second, so this is reachable from an ordinary two-model script, not only from a test.
The filename now carries nanoseconds and a per-process counter (`trace_file_path`, with its
own unit test).

## Alternatives considered

- **Shipping `robbins_monro` as the default** (what the prototype branch
  `worktree-saem-speed-e7-rm-scale` did). Rejected on the measured pembrolizumab regression
  — a default change here moves estimates for everyone, and the evidence does not support it.
- **Lowering `adapt_interval` instead.** More corrections of a fixed size still cannot close
  a 15× gap without overshooting once it arrives; the factor, not the frequency, is what
  bounds the reach.
- **Lowering `SAEM_MH_STUCK_ACCEPT` from 1 % to 10 %.** That tier's wording ("not mixing,
  Ω/σ unreliable") is too strong for a 12 % chain, and it reads the *cumulative* rate, which
  is the wrong statistic (see above). A second tier on the tail is the honest fix.
- **Exposing the realised tail acceptance as a new `FitResult` field** so the Tier-3 test
  could read a number rather than a warning. Rejected as scope creep for ~15 construction
  sites plus the `fitrx` file format; the optimizer trace already carries the per-iteration
  `mh_accept_rate`, and the Tier-3 test reads that.

## Cross-repo dependency

The new `[fit_options]` key is parsed in ferx-core, so an R user can already set it in a
model file. `FitOptions::saem_scale_adaptation` is a new public field with a `Default`, so
ferx-r's struct-update construction is unaffected; a follow-up to surface it as an R-level
argument is optional, not required.

## Breaking changes
- [ ] None

Four items added to the baseline, all required by the new user-facing option:
`ScaleAdaptation` (+ its two variants) and `FitOptions::saem_scale_adaptation`. A user
setting `scale_adaptation` from a `.ferx` file needs neither, but a Rust or ferx-r caller
building `FitOptions` programmatically cannot name the value without the enum being public —
the same shape as the existing `Optimizer` / `InnerOptimizer` / `BloqMethod` options. Nothing
was made `pub` that is not reachable from a `[fit_options]` key.

No migration needed: the default is the previous behaviour.

## Numerical validation

**No NONMEM anchor exists for this object, and that is not a convenience.** The quantity
under test is "what acceptance rate did *our* MH kernel reach" — NONMEM's SAEM exposes no
comparable step-scale knob and no per-iteration acceptance, so there is no equivalent run to
anchor against. Per CLAUDE.md this is validated the way it demands when no external run
exists, on three legs:

**(a) Exact / hand-computed unit tests on the rule itself.**
`rm_scale_update_matches_the_hand_computed_formula` pins `log δ += c·k^-0.6·(rate − target)`
at two hand-checkable points — `k = 1` (`k^-0.6 = 1`, so `Δlog δ = 0.5` exactly) and
`k = 32` (`32^-0.6 = 2^-3 = 0.125` exactly, so `Δlog δ = 0.0625`) — plus the decay, the fixed
point at zero discrepancy, and both clamps.
`interval_scale_update_is_the_legacy_times_1_1_or_0_9` pins the legacy arm's ×1.1 / ×0.9, its
`>` (not `>=`) comparison, and both clamps.

**(b) A differential pair that straddles the gate, asserted twice — in closed form and
end-to-end.**
`rm_reaches_a_target_the_interval_rule_provably_cannot` (Tier 1) drives both **production**
update functions round a closed loop against a monotone synthetic acceptance response
`a(δ) = 0.008/δ`, chosen so the 40 % target sits at `δ* = 0.02` and the 0.3 start reads
2.7 % — inside the 2–4 % actually measured on warfarin. After 400 iterations the interval arm
lands on `0.3·0.9^8` *exactly* (asserted to 1e-15, with the corrections counted) for a
realised **6.19 %**, and the RM arm on **39.98 %**. The straddle itself is asserted
(`interval < MH_RATE_LOW < rm`) so the pair cannot quietly become a tautology, and the
converse (fed 90 % acceptance the scale must *grow*) shows it is a controller, not a ratchet.

`tests/saem_scale_adaptation.rs` (Tier 3, slow-gated) runs the real thing —
`examples/warfarin_saem.ferx`, `data/warfarin.csv`, 150 + 250 iterations, `n_mh_steps = 20`,
seed 12345 — and reads the last 100 iterations of the optimizer trace's `mh_accept_rate`,
which is the same combined rate the diagnostic folds:

```
# tail acceptance over the last 100 iterations, examples/warfarin_saem.ferx
scale_adaptation = interval      **3.90 %**
scale_adaptation = robbins_monro **42.05 %**
```

**(c) The warning: a fixture that fires it, a fixture that does not, and the message's
numbers.** `saem_mixing_warning_reports_a_tail_that_never_reached_target` fires on a 3 % tail
and on a 90 % tail, checks the message carries the rate / target / window / direction /
remedy, sweeps five in-band rates to prove it is a *band* and not a one-sided test, and pins
that the **tail** decides by feeding two cases with identical cumulative counts and opposite
tails. `saem_mixing_warning_says_nothing_before_the_scales_had_a_chance` pins the
minimum-window gate at `MH_RATE_MIN_WINDOW ± 1` and checks it does not mask tier 1.
`saem_reports_a_tail_acceptance_far_from_target_end_to_end` drives a **real fit** to a 8.6 %
tail (a grossly over-dispersed Ω), parses the rate back out of the message and asserts it is
inside the gap the old 1 % threshold missed (`SAEM_MH_STUCK_ACCEPT < rate < MH_RATE_LOW`),
with a well-scaled fit (57.9 % tail) as the negative control.

**Fold trap.** Every accumulator that could see a non-finite value asserts `is_finite()`
before folding (`tail_accept_rate` in the Tier-3 test, the max-|Δθ| fold in the wiring test,
the realised-rate checks in the closed-loop test) — `f64::max` discards `NaN`, so a diverged
solve would otherwise pass on the strength of the rows that worked. The warning's own
accumulators are `u64` counts.

**A fast (Tier-1) version of the same straddle, on a real fit.**
`robbins_monro_moves_the_realised_acceptance_toward_target_both_ways` runs a 60-iteration,
8-subject SAEM fit four times — two Ω scalings × two rules — and reads the realised
acceptance from the optimizer trace's `mh_accept_rate` column, the same quantity the
diagnostic folds. It runs in ~0.2 s, so the wiring is covered on PR CI and not only nightly:

```
# mean combined MH acceptance, post-burn-in, 8 subjects, 60 iterations, seed 1444
                      interval   robbins_monro
Ω = 100 (far below)     0.0856          0.3116     <- straddles MH_RATE_LOW = 0.10
Ω = 0.09 (above)        0.6172          0.4337     <- moves *down* toward target
```

The second row is the controller-not-ratchet property on a real chain, not just on the
closed form, and both rows assert the RM arm is closer to the 40 % target than its twin.

**Legacy bit-identity.** `interval_is_the_default_and_bit_identical_to_asking_for_it` runs
the same 60-iteration SAEM fit twice — once with the option untouched, once with `interval`
asked for explicitly — and compares OFV, every θ and every σ **by `to_bits()`**, not to a
tolerance. A single last-bit difference fails it, and it is what dies if the default ever
flips. The whole 4 871-test `cargo test --lib` suite is the wider statement of the same
thing: it is green and unchanged, and the default path is the only path those tests take.

- [x] Compared against: prior ferx commit `65118bea` (the `interval` arm is bit-identical to it)
- [x] Reference values: see the tail-acceptance table above

## Performance

Under `interval` (the default) the only added work per post-burn-in iteration is one
`VecDeque` push of two `u64`. Under `robbins_monro` it is one `powf` and one `ln`/`exp` per
subject per iteration plus one per η — negligible against `n_mh_steps` subject-NLL
evaluations. Worth noting the *other* direction: with the scale at target, the issue measured
8 proposals per subject per iteration beating the default 20, so `robbins_monro` is a route
to a cheaper E-step on models where it applies.

## Tests

**Mutation testing.** Each new assertion was checked against the smallest edit that removes
what it guards, not only against my own diff:

| # | Edit | Test that went red |
|---|------|--------------------|
| M1 | `rm_scale_update` returns its input | `rm_scale_update_matches…`, `rm_reaches_a_target…`, `robbins_monro_moves…` |
| M2 | the whole RM block in `run_saem` never runs | `robbins_monro_moves…` |
| M2b | only the componentwise half of the RM block runs | `robbins_monro_moves…` |
| M2c | only the block half of the RM block runs | `robbins_monro_moves…` |
| M3 | RM decay exponent `0.6 → 0.0` (adaptation stops diminishing) | `rm_scale_update_matches…` |
| M4 | interval rule's `0.9 → 0.5` (its reach is no longer bounded) | `interval_scale_update_is_the_legacy…`, `rm_reaches_a_target…`, `robbins_monro_moves…`, `saem_reports_a_tail…` |
| M5 | tail tier deleted from `saem_mixing_warning` | `saem_mixing_warning_reports_a_tail…`, `…says_nothing…`, `saem_reports_a_tail…` |
| M6 | minimum-window gate deleted | `saem_mixing_warning_says_nothing…` |
| M7 | the window is never populated in `run_saem` | `saem_reports_a_tail…` |
| M8 | tail tier becomes one-sided (upper bound dropped) | `saem_mixing_warning_reports_a_tail…` |
| M9 | the cumulative rate is used where the tail should be | `saem_mixing_warning_fires_only_when_stuck`, `…reports_a_tail…` |
| M10 | the default flips to `RobbinsMonro` | `interval_is_the_default…`, `scale_adaptation_parses…` |

**Two of these found real holes, and both were closed rather than written up.**

- **M2 originally passed.** The first version of the wiring test asserted only that the two
  rules produce *different* estimates — but selecting `robbins_monro` also switches the
  interval bump **off**, so the runs differ even when the RM update does nothing at all.
  Deleting the entire RM block left every test green. The test now measures the realised
  acceptance rate from the optimizer trace instead, which is the quantity that actually
  distinguishes "adapting differently" from "not adapting".
- **M2c then passed.** With the componentwise half deleted, the block half alone still moved
  the combined rate enough to clear a "closer to target than its twin" margin. The fix was to
  add absolute near-target bounds measured on both arms (0.3116 / 0.4337 with both halves,
  0.2380 / 0.4841 with the componentwise half removed), which is what now dies.

Mutations were applied to the production code and reverted, one at a time, by a scripted
harness; the raw per-mutation log is reproducible from the edits in the table.

```toml
[fit_options]
  method           = saem
  n_exploration    = 150
  n_convergence    = 250
  n_mh_steps       = 20
  scale_adaptation = robbins_monro   # default: interval
```

On `examples/warfarin_saem.ferx` with that one line added, the E-step's combined acceptance
over the last 100 iterations goes from **3.90 %** to **42.05 %** against the 40 % target, and
without it the fit now says so:

```
SAEM Metropolis-Hastings acceptance settled at X.X% over the last 100 iterations, against a
target of 40% (X.X% over all post-burn-in iterations). That is too large a step — most
proposals are rejected, so the E-step explored less of each subject's conditional
distribution than it paid for and the Ω/σ estimates carry more Monte-Carlo noise than they
need to. Consider `[fit_options] scale_adaptation = robbins_monro`, which steps the scales
every iteration instead of every `adapt_interval`.
```

## Docs
- [x] New pages linked in `docs/_quarto.yml` (no new page — existing pages edited)
- [x] `CHANGELOG.md` `[Unreleased]` entry added (user-facing change)

`docs/model-file/fit-options.qmd` gains the `scale_adaptation` row and clarifies what
`adapt_interval` now governs; `docs/estimation/saem.qmd` §"Adaptive Step Sizes" is rewritten
with a subsection per rule (including the interval rule's reach arithmetic and the measured
case where Robbins-Monro *loses*) and a new "Acceptance-rate diagnostics" subsection.
`cargo run -p docs-lint -- --check` is green.

## Checklist
- [x] `tools/preflight.sh` green — all six groups green (`fmt check clippy rustdoc docs public-api`), exit 0. The
      `ci,nn,slow-tests` set that the issue tracker flags as red on `main` (#1446) passed
      here.
- [x] Functions on the `Dual2` sensitivity path stay differentiable (none touched — SAEM's
      MH kernel is value-only; no `sens/` or `*_g<T: PkNum>` code is in this diff)
- [ ] `Cargo.toml` version bumped if breaking (additive public API, no bump)

### ferx-site
- [x] New `.ferx` DSL syntax or `[fit_options]` key → `ferx-site/model-dsl/` page updated or PR opened — **not opened**; flagged here as the follow-up a `[fit_options]` addition normally carries.
- [ ] New example `.ferx` file added to `examples/` (none added — the existing
      `examples/warfarin_saem.ferx` is left on the default rule on purpose, so the shipped
      example keeps its current estimates)
- [ ] New data file added to `data/`

### ferx-book
- [ ] New estimator, DSL feature, or fit option → relevant book chapter updated or PR opened — **not opened**, same follow-up as above.

### Example execution
- [x] No example execution step needed — no `examples/*.ferx` file changes, and the default
      path is bit-identical, so every shipped example produces the same output it did before.

## Reviewer hints

Focus on:

- **`saem_mixing_warning`'s two tiers and their ordering** (`src/estimation/saem.rs`). Tier 1
  reads the cumulative rate and returns early; tier 2 reads the window. They are not
  redundant gates — feeding the same cumulative counts with opposite tails selects different
  answers, which is what the tests pin.
- **The `if !rm_scale_adaptation` split in the adapt block** (~`saem.rs:4278`). The counters
  are still *reset* under both rules, because the per-iteration trace reads them; only the
  multiplicative bump is gated. Applying both controllers would be a coarse one fighting a
  fine one.
- **The closed-loop straddle test's acceptance response.** `a(δ) = 0.008/δ` is a synthetic
  stand-in, not a model of MH acceptance; what it has to be is monotone and to put the start
  far from target. The test asserts the straddle it relies on.

Skimmable: the docs prose, the CHANGELOG, and the public-API baseline (four sorted lines).

## Open questions

- **Should `robbins_monro` become the default later, and on what evidence?** The pembrolizumab
  regression is one model at 3 seeds. A "when it helps / when it hurts" predicate — plausibly
  "is the starting scale's acceptance already inside the band?" — would let the default be
  chosen per fit rather than globally, but that is a separate change with its own
  measurement.
- **Should the κ (IOV) scales move to the same rule?** The argument against is quantisation
  (one proposal per occasion per iteration); the argument for is that Robbins-Monro is a
  stochastic-approximation scheme and averages binary signals correctly by construction — it
  is exactly what Stan does to its HMC step size. Left out because it is unmeasured, not
  because it is wrong.


roninsightrx:

The implementation is well structured and preserves the legacy default, but I would request changes. The new diagnostic can warn before interval adaptation has occurred, reports an incorrect target for multi-ETA HMC, and recommends Robbins–Monro even when already enabled. The convergence tolerance and NONMEM validation also do not meet the repository’s stated review rules. Static review only; nothing was compiled or executed.

## Blocking findings

- `src/estimation/saem.rs:1683 — high — MH_RATE_MIN_WINDOW = 25 does not establish that interval adaptation has occurred. For example, with omega_burnin = 20, adapt_interval = 50, and 45 total iterations, the window reaches 25 while the scale is still the untouched initial value, yet the code reports that it “never reached” its target. The test at line 8029 codifies this premature behavior — Track whether an adaptation has occurred and require at least one subsequent sampling iteration before issuing this tier; cover default and custom intervals separately from Robbins–Monro.`

- `src/estimation/saem.rs:4432 — medium — saem_mixing_warning receives a combined primary-plus-componentwise acceptance rate but labels it with the primary kernel’s target. With multi-ETA HMC, one HMC proposal targets 65% while roughly 20 componentwise proposals target 44%, so the combined target is around 45%, not the reported 65% — Track and diagnose the two kernels separately, or compute a proposal-weighted target and test a multi-ETA HMC fixture.`

- `src/estimation/saem.rs:1704 — medium — The new warning always recommends scale_adaptation = robbins_monro, including runs already using Robbins–Monro. Such a run can remain outside the band because it hit a clamp or was too short, making the prescribed repair a no-op — Pass the active adaptation mode into the diagnostic and provide mode-appropriate advice.`

- `tests/saem_scale_adaptation.rs:130 — medium — The bounds interval < 0.08 and RM ∈ [0.30, 0.55] are justified only as a “wide bracket”; the comment quotes one realization, not the measured worst realized error required by the repository rules — Measure and print a seed/platform sweep, quote the worst margin in the comment, and derive the bounds from it.`

- `docs/estimation/saem.qmd:412 — medium — This numerical feature claims a 60% reduction in “final-estimate error,” but the reference and parameter-level results are not identified, and the reviewed diff contains no equivalent NONMEM comparison as required for numerical behavior changes — Add the model, seeds, estimates/OFV for both rules, and the corresponding NONMEM output, with measured errors.`

- `src/estimation/saem.rs:8293 — medium — The new Tier-1 unit block performs eight 60-iteration fit() calls across its tests. Repository guidance says Tier 1 should avoid fit(), Tier 2 should use only a handful of iterations, and convergence runs belong behind slow-tests — Keep pure update/predicate tests in Tier 1 and move or gate the repeated-fit tests at Tier 3; use a genuinely short fixture for any required production-wiring coverage.`

## Nits

- `src/estimation/saem.rs:8291 — nit — The helper comment still says trace filenames contain only PID and whole seconds and that the second fit truncates the first, although this PR fixes exactly that behavior — Update the comment to describe reading and deleting each unique trace.`

- `src/estimation/saem.rs:8513 — nit — The same control rate is documented as 57.9% here and 61.72% six lines later — Replace the stale measurement with the verified value.`

roninsightrx:

Thanks — this was a useful pass. I reproduced every blocking finding rather than reading them statically, and **four of the six are real and now fixed**; one is fixed differently from the suggestion, and one I am rejecting with measurements. Both nits are fixed. Every new gate was mutation-tested, which found one further hole of my own making (below).

---

**F1 — `MH_RATE_MIN_WINDOW = 25` does not establish that adaptation happened — `src/estimation/saem.rs:1683` — ACCEPTED AND FIXED, and the constructed case reproduced exactly.** I built Codex's fixture (`omega_burnin = 20`, `adapt_interval = 50`, 45 iterations) and ran it: the fit emitted *"SAEM Metropolis-Hastings acceptance settled at 8.7% over the last 25 iterations, against a target of 40%"* after the interval rule had fired **zero** times. That is a verdict on a controller that never ran, and my own doc-comment on the constant asserted the opposite ("the scales have not had a chance to adapt"), so the constant was not delivering its stated rationale. Fixed by tracking the iteration of the first block/componentwise adaptation (`MhRateWindow::first_adapt`) and requiring `first_adapt < samples[0].iter` — every iteration whose rate is reported must have been drawn with a scale the controller had already touched, since a scale updated at the end of iteration `k` first affects sampling at `k+1`. The two gates are kept separate and documented as answering different questions, and they are provably non-redundant: mutation M6 (delete the length floor) reddens only `saem_mixing_warning_says_nothing_before_the_scales_had_a_chance`, and M11 (delete the adaptation gate) reddens only `..._is_silent_until_the_controller_has_actually_run` plus the new end-to-end test. Coverage added at both levels — a unit test pinning the exact off-by-one boundary (adapting *at* `samples[0].iter` is too late, one iteration earlier is in time) and `saem_says_nothing_about_a_target_the_interval_rule_never_tried_to_reach`, which runs Codex's 45-iteration fixture end to end with an adapted control beside it so its silence cannot be the diagnostic simply never speaking. This also re-tuned the positive fixture: at `omega_burnin = 20` its window opened at iteration 21 and straddled the first adaptation at 25, so it is now `omega_burnin = 30` and the whole window post-dates the controller.

**F2 — combined rate labelled with the primary kernel's target — `src/estimation/saem.rs:4432` — ACCEPTED AND FIXED, and the HMC arithmetic confirmed by measurement.** I ran a two-η HMC fit (`n_leapfrog = 10`, all 8 subjects on HMC): realised combined tail **0.4978** against a proposal-weighted target of **0.4500** (one HMC proposal at 0.65 against 20 componentwise at 0.44), where the message would have printed 65%. So the old label did not merely round badly — it would have told the user their chain was 15 points *below* target when it was 5 points *above* the real one. Fixed by carrying `target_weight = Σ_k n_proposals_k · target_k` per iteration in `MhRateSample` and quoting `Σ target_weight / Σ proposed`; the message now says "proposal-weighted target". This also corrects the ordinary non-HMC multi-η case, which was reported as 40% and is really 0.42 (20 block at 0.40 + 20 componentwise at 0.44) — and that is now the asserted value in the end-to-end test. I took the "compute a proposal-weighted target" option rather than "diagnose the two kernels separately", because the combined rate is deliberate (#895: a block-only rate reads a misleading 0% for FREM-scale Ω), so the fix belongs in the label, not in splitting the metric.

**F3 — recommends `robbins_monro` to a run already using it — `src/estimation/saem.rs:1704` — ACCEPTED AND FIXED.** Correct, and the no-op repair was exactly as described. The diagnostic now takes the active `ScaleAdaptation` and branches: the `interval` arm offers the switch (or a lower `adapt_interval`), and the `robbins_monro` arm says it is already active and points at the two things that actually remain — a step scale against a clamp, or an Ω whose diagonal is orders of magnitude from the posterior spread, which no step size rescues. Pinned by `saem_mixing_warning_does_not_prescribe_the_rule_already_running`, which asserts both that the RM arm says "already active" and that it does *not* contain the prescription, with both arms still carrying the numbers; mutation M14 (collapse the match to the interval arm) reddens it.

**F4 — bounds bracketed, not measured — `tests/saem_scale_adaptation.rs:130` — ACCEPTED AND FIXED, with the sweep run.** Fair under CLAUDE.md's "measure the tolerance; never justify it with a story". I ran the four-seed sweep:

| seed | `interval` | `robbins_monro` |
|---|---:|---:|
| 1 | 0.0396 | 0.4228 |
| 2 | 0.0408 | 0.4178 |
| 777 | 0.0412 | 0.4170 |
| 12345 | 0.0390 | 0.4205 |

Worst realised `interval` **0.0412** (cross-seed spread 0.0022); worst `|robbins_monro − 0.40|` **0.0228** (spread 0.0058). Bounds derived from those and stated in the test's doc comment: `interval < 0.06` (1.46× the worst observed, still well under the diagnostic's own `MH_RATE_LOW = 0.10`) and `|rm − 0.40| < 0.08` (3.5× the worst deviation) — replacing the asymmetric `[0.30, 0.55]`, which was 4.6σ wide on one side and 24σ on the other for no stated reason. The same discipline was applied to the fast test's bounds, which now carry a table of the realised margin with and without the mutation they exist to catch.

**F5 — 60% error-reduction claim with no reference, seeds, or NONMEM — `docs/estimation/saem.qmd:412` — ACCEPTED AND FIXED (the claim was not mine to make).** Correct, and worse than Codex could see from the diff: that number came from the research branch this PR started from, and **I never reproduced it**. Asserting it as fact was the error. The docs now state only what this repository measures — acceptance 3.90% → 42.05% on `examples/warfarin_saem.ferx`, reproduced by `tests/saem_scale_adaptation.rs` — and attribute the estimate-quality results (halved warfarin error; pembrolizumab regressing at 3 of 6 seeds) to the benchmark suite in #1444 by name, saying explicitly that they are not from this repository's tests. The CHANGELOG got the same treatment, and its stale 4.6% / 41.6% are now the measured 3.90% / 42.05%. On the NONMEM half: there is no equivalent run to add, and that is the documented CLAUDE.md exception rather than an omission — NONMEM's SAEM exposes no comparable step-scale knob and no per-iteration acceptance, so "what rate did our MH kernel reach" has no NONMEM spelling. The validation legs used instead (exact hand-computed values at `k = 1` and `k = 32`, where `32^-0.6 = 2^-3` exactly; a closed-form straddle over the production update functions; the same straddle end-to-end in both directions) are set out in the PR description, and the docs now say which numbers are ours and which are the benchmark's.

**F6 — the Tier-1 block runs eight 60-iteration `fit()` calls — REJECTED, with measurements.** Three reasons, and I would rather be argued out of them than concede silently. *Cost*: the whole block runs in **0.79 s** single-threaded (`cargo test --lib -- --test-threads=1`; it is now 11 fits after the F1 regression test), against Tier 1's stated "seconds total" budget — it is not a convergence run, it is a fixed 60-iteration budget on 8 subjects and 32 observations. *Precedent*: `origin/main`'s own test module in this same file already makes **14** `crate::api::fit(...)` calls with `saem_n_exploration = 4`-style budgets, so a short fixed-budget fit is this file's established unit-test shape; singling out this PR's would be inconsistent, not stricter. *Decisive*: what these cover cannot move. Mutation M2 — deleting the Robbins-Monro block from `run_saem` entirely — is invisible to every pure update-rule test, and `slow-tests` never runs on a PR, so moving them to Tier 3 would restore precisely the hole mutation testing found (it is the hole the first version of this PR shipped with). The genuine convergence run — 400 iterations on warfarin — *is* gated behind `slow-tests` in `tests/saem_scale_adaptation.rs`, which is the split CLAUDE.md is asking for. I have recorded that reasoning and the measured runtime in a comment above the block so the next reviewer does not have to re-derive it.

**Nit 1 — stale trace-filename comment — FIXED.** Good catch; the comment described the bug this PR fixes. It now says each fit gets its own trace (nanoseconds plus a per-process counter) and that the helper reads and deletes it.

**Nit 2 — 57.9% vs 61.72% for the same control — FIXED, and both numbers were stale.** The 57.9% came from an earlier scan that used `Ω = (0.09, 0.09)` rather than the fixture's `(0.09, 0.04)`. Both now read the re-measured **0.6114** for the retuned `omega_burnin = 30` fixture, and that control now also asserts its realised rate is inside the band — otherwise its silence proves nothing.

---

**One further hole, found by mutation-testing the above rather than by review.** With the weighted target in place, mutation M13b (`run_saem` stops accumulating `target_weight`) left every test green: the unit tests build their own `MhRateSample`s, so they cannot see the wiring. The end-to-end test now asserts the emitted message contains `proposal-weighted target of 42%`, which is the exact 0.40/0.44 mix that fixture runs; under M13b it prints "target of 0%" and the test dies. Full matrix — 16 mutations, all now red, each naming which test dies:

| # | Edit | Test that dies |
|---|---|---|
| M1 | `rm_scale_update` returns its input | `rm_scale_update_matches…`, `rm_reaches_a_target…`, `robbins_monro_moves…` |
| M2 | RM block in `run_saem` never runs | `robbins_monro_moves…` |
| M2c | only the block half of the RM block runs | `robbins_monro_moves…` |
| M3 | RM decay exponent `0.6 → 0.0` | `rm_scale_update_matches…` |
| M4 | interval shrink `0.9 → 0.5` | 5 tests |
| M5 | tail tier deleted | 7 tests |
| M6 | minimum-window gate deleted | `…says_nothing_before_the_scales_had_a_chance` only |
| M11 | adaptation gate deleted (F1) | `…is_silent_until_the_controller_has_actually_run`, `saem_says_nothing_about_a_target…` |
| M11b | `first_adapt` never recorded (F1 wiring) | `saem_reports_a_tail…`, `saem_says_nothing_about_a_target…` |
| M13 | weighted target reverts to the primary kernel's (F2) | `…quotes_a_proposal_weighted_target` + 2 |
| M13b | `target_weight` never accumulated (F2 wiring) | `saem_reports_a_tail…` |
| M14 | remedy ignores the active mode (F3) | `…does_not_prescribe_the_rule_already_running` |
| M7 | window never populated | `saem_reports_a_tail…`, `saem_says_nothing_about_a_target…` |
| M8 | tail tier one-sided | `…reports_a_tail_that_never_reached_target` |
| M9 | cumulative used instead of the tail | `…fires_only_when_stuck`, `…reports_a_tail…` |
| M10 | default flips to `RobbinsMonro` | `interval_is_the_default…`, `scale_adaptation_parses…` |

M6 and M11 reddening *disjoint* tests is the evidence that the two gates are not belt-and-braces for each other, which is the trap CLAUDE.md calls out.

**One unrelated latent bug surfaced while fixing the tests, and fixed here.** The trace writer lives in a thread-local, and `fit()` runs `fit_inner` inside a shared Rayon pool via `install`, so two concurrent `fit()` calls from two libtest threads can interleave on one pool worker — the second `init` replaces the writer the first `finish` is about to take, and `FitResult::trace_path` comes back `None`. It reproduced under `--test-threads=2` while each test passed alone. That is a property of the thread-local design rather than of any one test, so every test that enables `optimizer_trace` now takes one shared lock (`trace::TRACE_TEST_LOCK`, documented with the mechanism), including the pre-existing `quiet_fit_tests` one, which cannot fail itself but can steal the writer from a test that can. Verified stable over four consecutive `--test-threads=2` runs.

Gates: `tools/preflight.sh` green (all six groups), `cargo test --lib` green, and `tests/saem_scale_adaptation.rs` green under `--features slow-tests`.

Issue #1447, opened 2026-09-18
perf(saem): E-step/M-step rebuild EventSchedule and allocate per MH proposal, and the [individual_parameters] bytecode re-runs η-independent covariate algebra on every NLL evaluation (~50% of SAEM CPU)


roninsightrx:
## Summary

Three per-evaluation costs in the SAEM E-step / M-step that are pure overhead (η-independent work redone per MH proposal), all fixed bit-identically on branch `worktree-saem-speed-e4-hotpath` (+ `worktree-saem-speed-e6-ip-hoist`), measured at **−4 % to −25 % CPU** with byte-identical traces on every benchmark. Filing so the findings survive independently of the branch.

## 1. SAEM never cached the per-subject `EventSchedule`

On the event-driven analytic path every `individual_nll_into` call from `mh_steps` / `mh_steps_componentwise` passes `schedule: None`, so `event_driven_predictions` rebuilds `EventSchedule::for_subject` (merged event list over `2·n_doses + n_obs + …`, a sort, a `Vec` of propagation bounds per interval) **per MH proposal** — ~39 proposals × 400 iterations ≈ 15 600 rebuilds per subject per fit of a fit-invariant object. `find_ebe_cached` (FOCE inner loop) and the Bayes chain both cache it via `inner_optimizer::build_schedule_cache` / `cacheable_schedule`; SAEM did not. The M-step objective (`obs_nll_subject_into`) had the same gap — up to `mstep_maxiter·(n+1)` = 55 population passes per iteration on a 10-parameter model. Profile share on pembrolizumab: `EventSchedule::for_subject` 14 % of on-CPU work.

## 2. Seven heap allocations per MH proposal

`mh_steps` allocates `z`, a `DVector` copy, `chol(Ω)·z` and `eta_prop`; `individual_nll_into_with_schedule` allocates `pred_recycle: Vec::new()` (so the *recycling* entry point never recycles on the SAEM path) and two `DVector::zeros(n_eta)`, and rebuilds `obs_keys` / `ruv_obs_mult`. The IOV arms call the allocating `individual_nll_iov` wrapper although `individual_nll_iov_with_scratch` exists.

## 3. The `[individual_parameters]` bytecode is ~50 % of SAEM's on-CPU work, mostly η-independent

[...]

PR #1452, opened 2026-09-19
perf(estimation): cache the SAEM EventSchedule, hoist the MH proposal allocations, and memoise the eta-independent [individual_parameters] prefix [closes #1447]

roninsightrx:
## Why

Three per-evaluation costs in the SAEM E-step / M-step are pure overhead — η-independent
work redone on every Metropolis-Hastings proposal — and together they are roughly half of
SAEM's on-CPU time on covariate-heavy population models (#1447).

1. **SAEM never cached the per-subject `EventSchedule`.** On the event-driven analytic path
   every `individual_nll_into` call from `mh_steps` / `mh_steps_componentwise` passed
   `schedule: None`, so `event_driven_predictions` rebuilt `EventSchedule::for_subject` —
   a merged event list over `2·n_doses + n_obs + …`, a sort, and a `Vec` of propagation
   bounds per interval — **per proposal**: ~39 proposals × 400 iterations ≈ 15 600 rebuilds
   per subject per fit, of a fit-invariant object. `find_ebe_cached` (the FOCE inner loop)
   and the Bayes chain both cache it via `inner_optimizer::build_schedule_cache` /
   `cacheable_schedule`; SAEM did not. The θ/σ M-step objective (`obs_nll_subject_into`) had
   the same gap — up to `mstep_maxiter·(n+1)` = 55 population passes per iteration on a
   10-parameter model. `event_driven_predictions_with_schedule`'s own docstring already said
   where the schedule belongs.
2. **Seven heap allocations per MH proposal.** `mh_steps` allocated `z`, a `DVector` copy of
   it, the `chol(Ω)·z` product and `eta_prop`; `individual_nll_into_with_schedule` allocated
   `pred_recycle: Vec::new()` — so the *recycling* entry point never recycled anything on the
   SAEM path — plus two `DVector::zeros(n_eta)`, and rebuilt `obs_keys` / `ruv_obs_mult` on
   every call although neither depends on η. The IOV arms called the allocating
   `individual_nll_iov` wrapper although `individual_nll_iov_with_scratch` exists.
3. **The `[individual_parameters]` bytecode is ~50 % of SAEM's on-CPU work, and mostly
   η-independent.** `samply` on cefepime (458 subj) and pembrolizumab (303 subj) attributes
   51–52 % of on-CPU SAEM time to `build_pk_param_fn::{closure}`, `eval_bytecode` self 20 %.
   Almost all of it is covariate algebra — FFM, allometry, CKD-EPI, maturation `powf`s —
   that does not depend on η but was re-evaluated on every NLL evaluation, and once per
   *event* on the time-varying path.

## What changed

Five commits, the two wins kept separate.

**Items 1 + 2 — the E-step / M-step hot path** (`saem.rs`, `saem_conddist.rs`, `bayes.rs`,
`likelihood.rs`)

- `MhScratch` collects every buffer one subject's MH sweep needs and is owned by the Rayon
  worker (`map_init`), re-pointed at each subject by `begin_subject`. It carries an
  `IndividualNllScratch` (prediction vector + the two η-prior work vectors) and an
  `IndividualNllPrep` (the per-observation residual dispatch keys and the `#484` magnitude
  multipliers — both η-independent, so once per subject rather than once per proposal).
- `individual_nll_prepared` is the scratch-taking, schedule-taking form of
  `individual_nll_into`: same arithmetic in the same order, the buffers just come from the
  caller. `obs_nll_subject_into_with_schedule` is the same relationship for the M-step
  objective, and `obs_nll_sum` / `obs_nll_sum_mix` thread a `&[Option<EventSchedule>]`
  through (`&[]` = no cache, reproducing the previous behaviour exactly).
- `run_saem` builds the cache once per fit through the **shared**
  `inner_optimizer::build_schedule_cache`, so the staleness rules (no possibly-η-dependent
  lagtime, no `F`-reshaped rate-defined infusion, no ODE / compartment-free model) stay
  stated in exactly one place. Subjects the gate declines get `None`, which is the
  established rebuild-per-call path.
- The block MH kernel's `l * z` becomes `perturbation.gemm(1.0, l, z, 0.0)` — the same
  `gemm_uninit` kernel `Mul` reaches, into a buffer we already own; `beta = 0` never reads
  the previous contents.
- The IOV arms now call `individual_nll_iov_with_scratch`.

**Item 3 — the `[individual_parameters]` η-independent prefix hoist** (`model_parser.rs`)

- `split_ip_eta_independent_prefix` partitions the *resolved* statement list into an
  η-independent prefix and the remainder, walking it once under two rules: a statement that
  reads a slot any suffix statement has written goes to the suffix (taint), and a statement
  that writes a slot any suffix statement has already read or written goes to the suffix
  (conflict). Relative order is preserved within each part, so the hoisted statements are
  the same statements on the same inputs with the same data dependencies.
- The `pk_param_fn` closure memoises the prefix in a 64-way per-thread cache keyed on
  **exactly what the prefix reads**: the θ indices it touches (or the whole slice if a
  `ThetaGather` picks one at run time), the covariate indices it touches, and `TIME` /
  `MIXNUM` only when it reads them. Keys compare by raw bits, not `==`, so `-0.0` and a
  `NaN` covariate (#1111's missing-value encoding) never alias. `model_id` keeps two models
  sharing a worker thread apart.
- A cost gate declines the hoist when the prefix is cheaper than the lookup. This is
  measured, not guessed: the first cut was **1.6 % slower** on pembrolizumab (52.9 s vs
  52.1 s), whose prefix is 27 weighted ops against a 5-element key, while cefepime gained
  20 %. The gate is `prefix_cost >= 2·key_len + 10`, with no tuned multiplier.
- The `IndivParamProgram` snapshot keeps the **unsplit** list, so `sens/` and every `Dual2`
  evaluator are untouched — the split exists only inside the f64 closure.

**Review hardening added on top of the research branches** (see *Reviewer hints*): the
bytecode dependency walk's `_ => {}` arm is replaced by an exhaustive list; `bayes.rs`'s
hand-copied second spelling of the `cacheable_schedule` gate is collapsed onto the shared
builder; and two `#[cfg(test)]` thread-local kill switches plus per-thread cache counters
were added so the bit-identity claim could become a test.

## Alternatives considered

- **Expression-level CSE on top of the statement-level hoist (E8).** Measured: the cache-key
  bookkeeping costs what it saves. NO-GO, not implemented.
- **Resolving covariate names to slot indices inside `pk_param_fn`.** The `HashMap<String,
  f64>` lookup by name is 4.6 % of cefepime's on-CPU work, but removing it is a
  calling-convention change touching 216 signatures including `pub` `Subject` fields — a
  design decision, not a hot-path tweak. Left alone.
- **`--threads N` also sizes an idle global Rayon pool** (item 4 of the issue): `main.rs`
  sizes the global pool to N, then `install_on_fit_pool` leases a *separate* N-thread fit
  pool. Per-thread profile: the fit-pool workers are 85–99 % busy, the global-pool threads
  are 0.0 % busy for the whole run. Harmless for throughput, out of scope here, wants its
  own small PR.
- **A `[fit_options]` key to turn either optimisation off.** Not added: both are internal
  and produce identical results, so a switch would be surface with no decision behind it.
  The test-only kill switches are `#[cfg(test)]` and compile out of a production build.

## Cross-repo dependency

No public API change (`tools/update-public-api.sh --check` is green against the committed
baseline), no `FitResult` / sdtab change, no `.ferx` change.

## Breaking changes
- [x] **None**

## Numerical validation

**The documented exception that applies: none is needed, because nothing numerical changed.**
The CLAUDE.md NONMEM-anchor rule governs "any behaviour that produces numerical results".
This PR produces *identical* numerical results by construction — the same statements, in an
order with the same data dependencies, on bit-identical inputs — so the validation object is
not "does ferx agree with NONMEM" but "does ferx agree with ferx, bit for bit". A NONMEM run
cannot answer that question at the required resolution; `main` can, and does.

**End-to-end, against `main` (`65118bea`), same seed, same data:**

| Benchmark | `-fit.yaml` | `-sdtab.csv` |
|---|---|---|
| cefepime run64, 458 subj, SAEM 1200+600 | identical | **byte-identical** (`cmp`) |
| vancomycin peds run60, 300 subj, SAEM 100+50 | identical | **byte-identical** (`cmp`) |

**In the repo** (`cargo test --lib`, Tier 1) — the claim is asserted on raw `u64` bits, never
a tolerance, because SAEM's per-iteration state is the next iteration's input and a one-ULP
divergence at iteration 1 is a visibly different fit by iteration 5:

- `saem_schedule_cache_is_bit_identical` — the same seeded fit with and without the schedule
  cache, compared on the objective, every estimate, every `eta_hat` and every κ. Two
  fixtures, an analytic model and an **IOV** model (different likelihood entry points), both
  asserted to reach the event-driven analytic path via the shared gate's own output.
- `saem_individual_parameter_hoist_is_bit_identical` — the same A/B for the prefix hoist, on
  both fixtures, with a `closure_hoists` predicate that *observes* the cache rather than
  assuming the split happened.
- `mh_scratch_reuse_across_subjects_is_bit_identical` — one `MhScratch` walked across six
  subjects vs a fresh one per subject.
- Six `hoist_cache_*` tests in `model_parser_tests.rs` covering each keyed input.
- `eta_independent_split_is_value_preserving_on_every_shape` and
  `hoisted_pk_param_fn_is_bit_identical_under_interleaved_inputs` (from the research branch)
  cover eight statement shapes over 64 random input tuples each.

**Which engine each fixture ran on** (measured with `FERX_PROFILE=1`, not assumed):

- vancomycin peds: `129 522 event-driven f64 predictions`, `0.00 % FD fallback` — the
  **event-driven analytic** path, so all three optimisations are live.
- cefepime run64: **no** event-driven predictions at all — its covariates are all
  baseline-constant, so it routes to dose superposition and the schedule cache is inert
  there. Its numbers below therefore measure items 2 + 3 only; that is stated rather than
  glossed.
- The in-repo fixtures: every subject carries time-varying covariates and the model is
  closed-form 1-cpt IV, so all six take the event-driven analytic walk — the test asserts
  `n_cached == n_subjects` rather than trusting it.

## Performance
- [x] **Faster** — benchmark below
- [ ] Slower

Re-measured on this branch against `main` (`65118bea`), **user + sys CPU seconds** (wall time
is unusable — this box is shared), `--threads 2`, `--profile ci-test`, arms interleaved so a
load drift hits both equally. **Numbers below are after the review round**, which roughly
doubled the win on the event-driven benchmarks by removing four more per-iteration schedule
rebuilds.

**vancomycin peds run60 + IOV (`kappa KAPPA_CL`, `iov_column = OCC`), 300 subjects, SAEM
100+50** — the arm that consumed *no* cached schedule at all before the review:

| rep | `main` | branch |
|---|---|---|
| 1 | 85.38 | 55.32 |
| 2 | 85.73 | 40.08 |
| 3 | 64.60 | 39.79 |
| **median** | **85.38** | **40.08** → **−53.1 %** |
| mean | 78.57 | 45.06 → −42.6 % |

**vancomycin peds run60, 300 subjects, SAEM 100+50** — event-driven analytic, no IOV:

| rep | `main` | branch |
|---|---|---|
| 1 | 82.35 | 55.87 |
| 2 | 93.77 | 46.49 |
| 3 | 80.53 | 51.51 |
| **median** | **82.35** | **51.51** → **−37.4 %** |
| mean | 85.55 | 51.29 → −40.0 % |

(Before the review round the same fixture measured −20.0 % median; the four per-iteration
rebuilds the consumption test found were the difference.)

**cefepime run64, 458 subjects, SAEM 1200+600** — superposition path, so items 2 + 3 only:

| rep | `main` | branch |
|---|---|---|
| 1–6 | 31.98 · 34.93 · 29.96 · 28.83 · 31.19 · 35.83 | 27.81 · 29.37 · 22.35 · 23.05 · 29.25 · 26.55 |
| **median** | **31.59** | **27.18** → **−13.9 %** |

All three produce a **byte-identical `-sdtab.csv`** against `main` (`cmp`). The run-to-run
spread is large because other agents are building on this machine; the median is the figure
to read, and every one of the eighteen paired runs has the branch faster. `ci-test`
(`lto = false`) was used for both arms so the two builds were affordable; the comparison is
like-for-like.

## Tests
- [x] Unit test(s) added in the same module (`cargo test --lib` passes — 4872 passed, 0 failed)
- [ ] No new tests needed

**Mutation transcript.** Every assertion was toggled against the defect it exists to catch
and the result recorded in the test's own doc comment. Round 2 added the pre-change oracles
and the consumption counter that Codex asked for:

| mutation | red |
|---|---|
| E-step `schedules[i].as_ref()` → `None` | `saem_schedule_builds_do_not_grow_with_iteration_count` |
| M-step `&schedules` → `&[]` | same |
| IOV predictor ignores its `schedule` | `mh_steps_consumes_…_on_both_arms` (message names `iov`) **+** the invariance test |
| `prep.ruv_mult` → `None` | `individual_nll_prepared_matches_the_wrapper_it_replaced` — **and nothing else in the suite** |
| `err_keys` → zeros | same, same caveat |
| `cholesky_perturbation_into` `beta` 0 → 1 | `buffered_mh_proposal_matches_the_allocating_form` |
| `cacheable_schedule` loses the `TIME` arm | `schedule_cache_covers_a_time_only_baseline_covariate_subject` + the invariance test's `time-only` arm |
| `!model.has_lagtime()` deleted from `cacheable_schedule` | `schedule_cache_declines_an_eta_dependent_lagtime`, at the count assertion *and* at the bit comparison |
| `self.prep.refresh(..)` deleted from `begin_subject` | the scratch test + all three SAEM trace tests |
| drop the covariate slice from the prefix key | the covariate probes, the cross-subject test, **and both SAEM hoist trace tests** |
| drop `split.key_time` / `split.key_mixnum` / empty the θ subset | the `TIME` / `MIXNUM` / `MATSLOPE` probe respectively |
| delete `entry.model_id == ip_model_id` | `hoist_cache_never_serves_another_models_entry` |
| `begin_subject` moved back above `MixtureClassGuard::enter` | `e_step_prep_is_built_inside_the_drawn_mixture_class` **+ five pre-existing mixture SAEM tests** |
| `eta_prop.clear()` before the `resize` | 🟢 — and correctly so, see below |

Four findings from running these rather than reasoning about them, all recorded in the test
comments:

- **`model_id` was a green mutation** in round 1. The first cross-model test used two models
  and passed with the comparison deleted, because `ip_key_hash` also mixes `model_id` in, so
  two models never collided — the two-redundant-gates hole. The hash has only
  `IP_CACHE_WAYS` outputs and is a distribution device, not a correctness gate. The test now
  parses more models than there are ways (collision certain by pigeonhole) and evaluates each
  twice per round, because 72 models round-robin through 64 ways evict each other and the
  first version recorded *no cache hits at all*.
- **The `gemm` mutation was green** until the test stopped reimplementing the `gemm` call.
  The proposal step is now the named production helper `cholesky_perturbation_into`, which
  `mh_steps` and the test both call.
- **The consumption counter found four live leaks** that were shipping in the PR as posted;
  see the reply comment for the numbers.
- **The hoisted prep was being built at the wrong mixture class** (found by a later review
  pass). `IndividualNllPrep` hoists `ruv_obs_mult` and `obs_keys` out of the proposal loop,
  and both can read `MIXNUM` — `validate_ruv_expr` rejects η and NN outputs, not the class
  index — but `begin_subject` ran one line *above* `MixtureClassGuard::enter`, so a
  `[mixture]` model with a `MIXNUM`-dependent residual magnitude scored every class-2
  subject's acceptance ratio at class 1's variance. Verified, not assumed: that model parses
  and its multiplier is 1.0 vs 1.5 across the classes. Fixed by moving the call below the
  guard, and the ordering contract is now **checked where the prep is consumed**, which makes
  every mixture test in the suite a detector — restoring the bad order reddens six tests,
  five of them pre-existing.
- **`eta_prop.clear()` is genuinely dead.** `n_eta` is fixed for a fit, so the following
  `resize` is a no-op and every element is assigned before it is read. Kept as a statement of
  the invariant; the comment says so rather than claiming a mutation it does not have.

**What these tests cannot see, stated up front.** An equality A/B is blind to the
optimisation simply being *absent* — delete the schedule argument and both arms take the
uncached path and agree. That is why round 2 adds (a) the soundness-gate mutations, (b) a
`#[cfg(test)]` build counter that distinguishes *constructed* from *consumed*, and (c) two
pre-change oracles that hold the replaced expression next to the new one. Every fixture also
asserts non-degeneracy (`n_cached == n_subjects`, `closure_hoists`, "the probe must change the
answer", "the subjects must score differently", "≥ 1 cache hit", "the uncached arm's build
count must grow").

- [x] **Not applicable** — internal, default-on, identical results; no new API surface

## Docs
- [x] `CHANGELOG.md` `[Unreleased]` entry added — one bullet under **Performance**
- [x] No user-visible change (no new `[fit_options]` key, no changed default, identical results)

`FERX_IP_HOIST_REPORT=1` prints one line per compiled closure saying whether it hoisted and
on what key. It is left undocumented in the user docs, consistent with the other `FERX_*`
developer diagnostics (`FERX_PROFILE`, `FERX_NO_ANALYTIC_INNER`, `FERX_SENS_CHECK`, …).
`cargo run -p docs-lint -- --check` is green (nothing in `docs/` changed).

## Checklist
- [x] `tools/preflight.sh` green — `preflight OK — fmt check clippy rustdoc docs public-api`.
      Note: the `ci,nn,slow-tests` feature set the shared brief flagged as pre-broken (#1446)
      passes here, so that issue appears already fixed on `65118bea`.
- [x] Functions on the `Dual2` sensitivity path stay differentiable — **nothing on that path
      was touched**: the split lives inside the f64 `pk_param_fn` closure and the
      `IndivParamProgram` snapshot `sens/` reads is the unsplit list, pinned by
      `hoist_leaves_the_indiv_param_program_unsplit`
- [ ] `Cargo.toml` version bumped if breaking — N/A, nothing breaking

### ferx-site

### ferx-book

### Example execution
- [x] **No example execution step needed** — internal, no user-visible change. Both
      benchmarks above were nonetheless run end-to-end through the CLI on both arms and
      produce byte-identical `-sdtab.csv`.

## Reviewer hints

**Where to focus, in order.**

1. **`split_ip_eta_independent_prefix`'s two reordering rules** (`model_parser.rs`). The
   hoist moves statements to the front of the program. The argument that this is invisible is
   the taint/conflict pair plus order-preservation within each part; `pk_vars` is
   `fill(0.0)`-ed every call, so a prefix statement reading a slot nobody wrote reads a
   deterministic zero, and one reading a slot an *earlier prefix* statement wrote is
   recomputed in the same prefix. `eta_independent_split_is_value_preserving_on_every_shape`
   checks this over eight shapes × 64 random tuples, and the branch's own doc comment records
   which fixture each of the two rules is load-bearing for.
2. **`ip_deps_bytecode`.** This is the soundness boundary: the hoist is correct only if the
   dependency walk names every input a statement can read. The research branch had a `_ =>
   {}` arm — which currently only catches pure ops, but would silently swallow a future `Op`
   reading state (an occasion index, a record field, a second thread-local). It is now an
   exhaustive list, so adding an `Op` variant fails to compile there. `ip_deps_expr`,
   `ip_deps_cond` and `ip_deps_stmt` were already exhaustive.
3. **What is *not* in the key, and why that is fine.** `eta` (every reader is marked
   `dynamic`, including IOV κ, which the parser compiles to `Expression::Eta(n_eta + k)`);
   `[covariate_nn]` outputs (marked `dynamic` — they are recomputed outside the statement
   list); unresolved names and out-of-range slot indices (marked `dynamic` / out-of-range, so
   the statement stays behind). `Op::PushThetaGather` picks its θ at run time, so it widens
   the key to the whole slice rather than a subset.
4. **`begin_subject` must stay below `MixtureClassGuard::enter`** in `run_saem`'s E-step.
   That ordering is a contract between two call sites with nothing in the types to enforce
   it; the `#[cfg(test)]` check inside `individual_nll_prepared` is what holds it, and it is
   the reason five unrelated mixture tests now fail if it is broken.
5. **The IOV plumbing added in round 2.** `predict_iov_with_scratch_and_schedule` applies the
   *same* `Cow::Borrowed` cache-validity test as the non-IOV arm, at the same place — a cached
   schedule was built from the unresolved subject, so it is only valid when nothing needed
   resolving. Worth a second pair of eyes on that one condition.
   `obs_nll_subject_into_iov` also started *using* the `EventPkParams` it had been accepting
   and ignoring; `predict_iov_with_scratch_and_schedule` clears all four event vectors at
   entry, which `pk::iov_scratch_tests` already pins.
6. Skimmable: the mechanical `&[]` additions to `obs_nll_sum` call sites in
   `fixed_eta_gradient_tests.rs`, and the `MhScratch::default()` conversions in the existing
   `mh_steps` tests.

## Open questions

- The prefix cache is **64-way, direct-mapped, per thread** (~16 KB/worker). That was sized
  against the benchmarks' longest subject (41 records). A model whose per-event covariate
  vector takes more than 64 distinct values within one subject will thrash it — every lookup
  a miss, paying key construction for nothing. The cost gate does not see this, because it
  is a property of the *data*, not of the program. I have not found a fixture where this
  actually costs more than it saves (a miss is a key build plus a hash, against a prefix the
  gate already judged more expensive than both), but if a reviewer wants a belt here, the
  natural one is an adaptive disable after a long run of misses. Worth an opinion.
- **`find_ebe_iov` still rebuilds a schedule per inner objective evaluation.** Found by the
  new build counter. It is a once-per-fit pass here (the post-loop final-EBE step), and it
  lives in the FOCE-shared inner optimizer — `find_ebe_impl` routes IOV subjects to it
  *before* the cached schedule is used at all — so fixing it would widen this PR into the
  FOCE inner loop. Left as a follow-up; the invariance test bounds it (it cannot grow with
  the SAEM iteration count). Happy to open the issue.
- Item 4 of the issue (`--threads N` sizing an idle global Rayon pool) is left for a separate
  PR, as the brief asked. Should I open it now or fold it in?


roninsightrx:

Static review found no definite numerical error, but I would not merge yet. The new likelihood/proposal path lacks a pre-change oracle, and the schedule-cache regression test is demonstrably vacuous for some mutations—including its IOV arm. The implementation also misses `TIME`-only analytical subjects despite explicitly claiming them. Per instruction, I did not compile or execute tests.

## Blocking findings

- `src/estimation/saem_hotpath_tests.rs:388` — high — The allocation-hoist test compares reused and fresh `MhScratch`, but both sides execute the new `gemm` proposal and `individual_nll_prepared` likelihood. Thus changes such as omitting `prep.ruv_mult` or altering the buffered matrix multiplication have no pre-change oracle; the other A/B tests also use the new path on both sides. — Add Tier-1 raw-bit comparisons between `individual_nll_prepared` and `individual_nll_into_with_schedule` at identical inputs, covering selected residual dispatch, custom RUV magnitude, cached/uncached schedules, and multiple subjects. Separately compare seeded buffered proposals against the former `L * DVector` calculation.

- `src/estimation/saem_hotpath_tests.rs:230` — high — The comment acknowledges that deleting the schedule argument from either MH or M-step leaves the suite green, while `n_cached()` only proves that a schedule can be constructed—not that `run_saem` consumes it. The IOV arm is already vacuous: `mh_steps`’ IOV branch ignores `schedule`, and `obs_nll_sum_iov` accepts no schedule. — Instrument schedule construction/consumption under `cfg(test)` and assert one build per cached subject versus repeated builds when disabled, independently for E-step and M-step. Either plumb schedules through the IOV predictor or remove the IOV cache claim and fixture.

- `src/estimation/saem.rs:3102` — medium — The new documentation says `TIME`-reading analytical models are cached, but `build_schedule_cache` delegates to a gate requiring time-varying covariates or resets. A subject with baseline covariates and `TIME` in `[individual_parameters]` takes the event-driven route but still rebuilds its schedule on every likelihood call. — Extend the shared gate with the analytical `TIME` routing predicate and add a baseline-covariate, `TIME`-only fixture that verifies actual cache consumption.

## Nits

None.

roninsightrx:
## Response to the Codex review — all three findings accepted, all three fixed (`10b85aad`)

Every claim was checked against the code and by running the mutation, not read. All three
were correct, and acting on the second one turned up **four live cache leaks that were
shipping in this PR** and that no equality test could ever have shown. Thank you — that one
finding roughly doubled the measured win.

---

**Finding 1 — no pre-change oracle for the new likelihood/proposal path. Correct, and the
two mutations you predicted were both green.** I verified it: with `prep.ruv_mult` forced to
`None`, and separately with `err_keys` filled with zeros, the entire suite stayed green —
the A/Bs run `individual_nll_prepared` on both sides, so they agree with a wrong answer by
construction. Added `individual_nll_prepared_matches_the_wrapper_it_replaced`, a raw-bit
comparison against `individual_nll_into_with_schedule` over six subjects × 3 η × {cached,
uncached} schedule, on two fixtures picked for what `prep` actually carries: a `#484`
residual-magnitude expression (`ruv_mult` populated, one `Vec` per observation) and a
covariate-selected `if/else` error model (`err_keys` genuinely per-observation, since
magnitude and `Selected` are mutually exclusive in the DSL and one fixture cannot cover
both). Both of your mutations now redden **that test and nothing else in the suite**, which
is the property that says the oracle is the only thing standing there. For the proposal:
`buffered_mh_proposal_matches_the_allocating_form` compares the seeded buffered product
against `l * DVector::from_column_slice(&z)` over 200 draws × a diagonal and a correlated Ω,
poisoning the output buffer before every call so `beta = 0` is load-bearing. Worth flagging
that the first version of this test **also could not fail**: it reimplemented the `gemm`
call, so mutating `beta` in `mh_steps` left it green. The proposal step is now the named
helper `cholesky_perturbation_into`, which `mh_steps` and the test both call, and
`beta: 0.0 → 1.0` is red at the first reused buffer.

**Finding 2 — `n_cached()` proves construction, not consumption, and the IOV arm is already
vacuous. Correct on both counts, and this is the finding that mattered.** `predict_iov`
took no `schedule` parameter at all, and `obs_nll_sum_iov` passed none, so an IOV fit built
a cache it had nowhere to hand. I chose to plumb rather than drop the claim:
`predict_iov_with_scratch_and_schedule`, `individual_nll_iov_with_scratch_and_schedule` and
`obs_nll_subject_into_iov_with_schedule` now thread a schedule through the IOV funnel under
the *same* `Cow::Borrowed` validity test the non-IOV arm applies (a cached schedule was built
from the unresolved subject, so it is only valid when nothing needed resolving), and
`mh_kappa_steps` takes one too. For instrumentation: `EventSchedule::for_subject` carries a
`#[cfg(test)]` **thread-local** build counter — thread-local rather than a global
`AtomicU64` so it is not polluted by every other test in the binary, and read from inside a
single-worker rayon `install` so a fan-out count is still exact. Three tests now pin
consumption, each naming its own arm on failure:
`mh_steps_consumes_the_cached_schedule_on_both_arms` (cached → 0 builds, uncached → ≥ one per
proposal, plus bit-identity), `m_step_objective_consumes_the_cached_schedule_on_both_arms`
(same, per-subject kernels rather than `obs_nll_sum*` so the count is exact), and
`saem_schedule_builds_do_not_grow_with_iteration_count`. **That last one immediately failed
on the code as posted**: the analytic fixture went 72 → 156 builds when the iteration count
tripled. Backtracing found four per-iteration rebuilds nobody had noticed — the post-M-step
`nll_cache` refresh in *both* the non-IOV and IOV branches, the κ phase's reference NLL, and
`mh_kappa_steps` itself, each one build per subject per iteration. All four now take the
cache; builds are flat. The assertion is invariance across two iteration budgets rather than
an absolute count, because the once-per-fit final-EBE pass is itself a numerical solve whose
line-search length drifts by a couple of builds (measured 204 → 206 on the IOV arm); a
per-evaluation leak adds `14·n` = 84 here, fourteen times the bound, so the two are not
close to confusable. One residual I am *not* fixing here and want to name: `find_ebe_iov`
rebuilds a schedule per inner objective evaluation. It is once-per-fit in this path and
lives in the FOCE-shared inner optimizer (`find_ebe_impl` routes IOV subjects to it *before*
the schedule is used), so fixing it widens this PR into the FOCE inner loop. The invariance
test bounds it; happy to open the follow-up issue.

**Finding 3 — the doc claims `TIME`-reading analytical models are cached and the gate does
not admit them. Correct.** The dispatcher in
`pk::compute_predictions_with_tv_recycle_with_schedule` branches on
`has_tv || has_resets || model_uses_time_anywhere`, and `cacheable_schedule` carried only
the first two, so a baseline-covariate subject of a `TIME`-dependent analytical model walked
event-driven and rebuilt its schedule on every likelihood call. I fixed the gate rather than
the claim: `TIME` does not enter `EventSchedule::for_subject` — it is built from the
subject's dose times and durations — so admitting those subjects changes nothing about the
staleness argument, only about who gets the reuse, and FOCE/FOCEI, Laplace, AGQ and the
Bayes chain share the gate and pick it up too (the full 4878-test `--lib` suite is green
with it widened). `schedule_cache_covers_a_time_only_baseline_covariate_subject` pins it with
a baseline-covariate `TIME`-only fixture, and verifies the fixture genuinely reaches the
event-driven walk (one build per uncached call) rather than assuming it, so the test cannot
pass vacuously. Reverting the gate reddens that test and the `time-only` arm of the
invariance test, and nothing else.

---

**Measured effect of the review round** (user + sys CPU, `--threads 2`, `ci-test`, 3
interleaved reps, medians; byte-identical `-sdtab.csv` against `main` on all three):

| benchmark | before review | **after** |
|---|---|---|
| vancomycin peds, 300 subj, **+ IOV** | (cache entirely inert) | **−53.1 %** |
| vancomycin peds, 300 subj | −20.0 % | **−37.4 %** |
| cefepime run64, 458 subj (superposition path, items 2+3 only) | −13.9 % | −13.9 % |

Full mutation transcript is in the PR body and in each test's doc comment.
`tools/preflight.sh` green; `cargo test --lib` 4878 passed, 0 failed. Not merged.

roninsightrx:
## Second review pass — one real defect, fixed in `6feabdf3`

**`saem.rs:3351` — the hoisted prep was built outside the mixture class guard. Correct, and
it was a genuine break of this PR's central claim.** I verified the whole chain rather than
reading it: the model below **parses** (`validate_ruv_expr` has arms for `Expression::Eta`
and `Expression::NnOutput`, none for `Expression::MixNum`), and `ruv_obs_mult` returns
`1.0` under class 1 and `1.5` under class 2 —

```
[mixture]
  nsub = 2
  logit(1) = MIXL
[error_model]
  DV ~ proportional(EPS * (1.0 + 0.5 * (MIXNUM - 1)))
```

`IndividualNllPrep::refresh` hoists both `ruv_obs_mult` and `obs_keys` out of the proposal
loop; the wrapper it replaced computed them inside the loop and therefore inside the drawn
class's guard. Building them one line earlier evaluated them at the ambient class — class 1
— and served that variance to every class-2 subject's acceptance ratio, so "bit-identical,
just faster" did not hold for that model family. Moved the call below
`let _class_guard = ..`; `draw_class`, which runs between them, touches only
`mh_scratch.pk()`, which `begin_subject` does not, so the correct order is free. I also
checked the other two `begin_subject` sites: `bayes.rs` only reaches it on the
`model.mixture.is_none()` branch (mixtures route to `mh_steps_mixture`), and
`saem_conddist.rs` enters no class guard at all, so both are unaffected.

**On testing it — the obvious end-to-end oracle does not work, and I want that on the
record.** My first attempt compared the `MIXNUM`-reading model against a twin with class 2's
value written as a constant, with the mixing logit fixed at −20 so every subject is drawn
into class 2. It failed *after* the fix, and the reason is instructive: the mixture
responsibility machinery evaluates **every** class's likelihood, not just the drawn one, so
the two models differ through `update_rbar` and the class-override statistics for reasons
that have nothing to do with the defect. Measured: the fit matched neither the class-1 nor
the class-2 twin. So the ordering is now checked **where the prep is consumed** —
`IndividualNllPrep` records the class it was refreshed at under `#[cfg(test)]` and
`individual_nll_prepared` asserts it matches the class in effect. That is strictly better
than a dedicated fixture, because it converts the existing mixture corpus into a detector:
restoring the bad ordering reddens **six** tests, five of them pre-existing mixture SAEM
tests that were green with the defect present. The new
`e_step_prep_is_built_inside_the_drawn_mixture_class` adds the direct check in three parts,
ordered so the last one means something — the fixture's magnitude must actually move with
`MIXNUM`; refreshing and consuming under class 2 must reproduce
`individual_nll_into_with_schedule` bit for bit while refreshing under class 1 must not; and
a whole mixture SAEM fit must run clean under the consumption-point assertion.

**`saem.rs:982` — duplicated `#[allow(clippy::too_many_arguments)]`. Correct, dead text from
merging the two attribute-adding commits.** Removed.

**Both non-findings confirmed and recorded so they are not re-derived.** The out-of-range
covariate index in `split_ip_eta_independent_prefix` is inert for exactly the reason given
(`pk_cov` is resized to `n_cov` every call, so `pk_cov.get(i)` past that is a constant
`0.0`, while an out-of-range θ must widen the key because `theta` is the caller's slice and
its length is not under our control). And a `NaN` covariate producing a cache *hit* under
bit-equality is the correct behaviour — same bits in, same values out — so the comment above
that block, which reads as if `NaN` were meant to miss, is the thing that is wrong rather
than the code; I have left the code and will not "fix" it into a miss.

`tools/preflight.sh` green; `cargo test --lib` 4879 passed, 0 failed. Not merged.

roninsightrx:
Follow-up on the second non-finding: rather than leave a comment I had just described as wrong, `1f1ab9c4` rewrites it. The block compares raw bits and the two motivating cases pull in opposite directions — `-0.0 == 0.0` is true but they are different inputs, so `==` would serve a false hit and bits give the correct miss; `NaN == NaN` is false, so `==` would miss forever on a subject carrying #1111's missing-value encoding, and bits give the correct hit. The old comment lumped both under "must not be treated as the same input", which is right for the signed zero and exactly backwards for the NaN. Comment only, no behaviour change; preflight green.

